PR #782, opened 2015-11-12
F/prefix defines

### LINKED ISSUE #745: distribute `config.h`
If somebody want to use syslog-ng's header files in a project then he/she needs syslog-ng's `config.h` file as well. It is produced during the build but `make install` doesn't install it.

The link [1] says that `config.h` needs to be distributed if it affects the interface. Probably it is included in the header files through `syslog-ng.h`.

I see the following solutions:
1. install `config.h`, perhaps renaming it to `syslog-ng-config.h`
1. include `config.h` only in C files, move elsewhere the dependent defines from `syslog-ng.h` (`YYDEBUG` and `PATH_SYSLOGNG`, they are used only from `.c` files), include `config.h` in `.c` files where it is needed

This "feature" is needed to support Rust modules without a locally built syslog-ng.

I think the second approach is better but maybe it's impossible.

@bazsi @lbudai I'd like to hear your opinions.

ihrwein:
Fixes #745 

Note: the `autoconf-archive` package needs to be intalled because it contains the`AX_PREFIX_CONFIG_H` macro.

@juhaszviktor : this change also affects incubator: the VERSION, SOURCE_REVISION macros are also prefixed


bazsi:
This looks good to me, but please tell me that you created this with a
script? In-depth reviewing is nearly impossible, but if it was done with a
script and then it compiles, I would say let's merge it.


-- 
Bazsi

ihrwein:
I used `sed` for prefixing all `HAVE_*` and `ENABLE_*` defines. The ones starting with `PATH_` was done manually, because some of them are defined by configure and some are written by us. I just prefixed the generated ones. The rest was prefixed manually but there's just a handful of them.

ihrwein:
I can add the sed scripts to the commits, but just on Monday. This PR is not so urgent, I don't have to use the master branch.

bazsi:
I don't need the scripts, only the fact that you did them by scripts. Like
I said, I would merge this without further ado.


-- 
Bazsi

Issue #780, opened 2015-11-11
Invalid error check due to signed/unsigned confusion


hannob:
In the function _read_text_file_content_without_trailing_newline in the file
modules/afsocket/transport-unix-socket.c
there is a bug in the error check.

The code sets content_len to the return value of _read_text_file_content. That function returns -1 in case of an error. Then there is a check for (content_len <= 0).

The problem: cotent_len is of type gsize (unsigned), while _read_text_file_content returns gssize (signed). If it returns -1 this can't be stored in conten_len and will underflow, thus the error check will be skipped.

This causes some invalid memory reads which I detected with Address Sanitizer.

To fix this content_len must also be gssize (signed), so it can be -1. This patch (also attached) will fix it:
```
a/modules/afsocket/transport-unix-socket.c	2015-10-27 09:08:53.000000000 +0100
+++ b/modules/afsocket/transport-unix-socket.c	2015-11-11 17:30:21.019247981 +0100
@@ -86,7 +86,7 @@
 static gssize
 _read_text_file_content_without_trailing_newline(const gchar *filename, gchar *buf, gsize buflen)
 {
-  gsize content_len;
+  gssize content_len;
   
   content_len = _read_text_file_content(filename, buf, buflen);
   if (content_len <= 0)
```
[syslog-ng-signed-underflow.txt](https://github.com/balabit/syslog-ng/files/32330/syslog-ng-signed-underflow.txt)

hannob:
Some further Info: This affects both syslog-ng 3.7.2 and the current git code.

The Address Sanitizer stack trace looks like this:
```
==10668==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7f013e792f3e at pc 0x7f013fe1b643 bp 0x7f013e792e90 sp 0x7f013e792e80
READ of size 1 at 0x7f013e792f3e thread T2
    #0 0x7f013fe1b642 in _read_text_file_content_without_trailing_newline modules/afsocket/transport-unix-socket.c:94
    #1 0x7f013fe1b642 in _add_nv_pair_proc_read_unless_unset modules/afsocket/transport-unix-socket.c:109
    #2 0x7f013fe1bc15 in _feed_aux_from_procfs modules/afsocket/transport-unix-socket.c:170
    #3 0x7f013fe1bc15 in _feed_credentials_from_cmsg modules/afsocket/transport-unix-socket.c:195
[...]

PR #783, opened 2015-11-12
Fix an integer underflow.

hannob:
The reason is that the function return value is a -1 on error, however
the variable to store that is unsigned, therefore it is unable to store
the error value and will underflow. The subsequent check will therefore
not catch the error.
To fix this change content_len from gsize (unsigned) to gssigned
(signed).

This was found with Address Sanitizer.

Fixes https://github.com/balabit/syslog-ng/issues/780


dnsjts:
:+1:

bazsi:
patch itself is great, thanks patching it, :+1:

on the other hand could you pls fix the subject of the patch (just add an afsocket: prefix, that should be enough) and a "signed-off-by:" tag at the end. 


Thanks.

hannob:
done that

PR #784, opened 2015-11-13
F/value pairs cleanups

bazsi:
I originally wanted to speed up value-pairs, but the biggest effect was a patch that wasn't thread safe, which I have dropped at the end.

So this series of patches does not really touch value-pairs performance (well at least for my naive testcase, YMMV), but reorganizes the code a bit, extracts a few functions, generally makes the code easier to read and maintain.

I have a few thoughts now how to improve value-pairs performance drastically, but I would do as a separate iteration, after this is integrated, and probably I would start by splitting up this huge file with intermingled internal layers.

Please merge, thanks.

### DIFF
--- a/lib/utf8utils.c
+++ b/lib/utf8utils.c
@@ -55,6 +55,7 @@ append_unsafe_utf8_as_escaped_binary(GString *escaped_string, const gchar *str, 
       switch (uchar)
         {
           case (gunichar) -1:
+          case (gunichar) -2:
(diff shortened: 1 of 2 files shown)


lbudai:
Could you squash/fixup the `!fixup` patch?
After that :+1:

bazsi:
I have squashed in the fixup patch, so if Travis confirms it we can merge it.

Anyone who gets there first :)

PR #785, opened 2015-11-13
F/groupingby parser

bazsi:
This patch series adds a new parser that can perform a group-by operation on incoming messages
as they are received and produce SQL GROUP BY like aggregation.

### DIFF
--- a/lib/template/templates.c
+++ b/lib/template/templates.c
@@ -184,12 +184,44 @@ log_template_format(LogTemplate *self, LogMessage *lm, const LogTemplateOptions 
   log_template_append_format(self, lm, opts, tz, seq_num, context_id, result);
 }
 
+
+/* NOTE: we should completely get rid off the name property of templates,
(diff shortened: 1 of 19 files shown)


faxm0dem:
Is this to be one of N correlation policies in replacement of patterndb's correlation engine?

bazsi:
Well, it is very similar to what db-parser correllation is, and you can use
multiple groupingby() expressioms in the same config.

Instead of coupling it with db-parser rules, you can use ordinary syslog-ng
filters to decide which message get to the same group.

I imagine this to be useable with together with existing parsers (even
db-parser, but csv-parser or kv-parser also works nicely).

faxm0dem:
yes, that's precisely why I asked the question: I thought the plan was to decouple patterndb and correlation in the mid/long term.
I'll try to test this branch ASAP, it's looking like I will definitely use this

PR #787, opened 2015-11-16
F/whitespace fixes

bazsi:
while browsing the code I've found these and fixed them up. 

done by emacs "whitespace-cleanup" function.

I'd merge this soonish as it really has no real code change, just whitespace, let me know if you have objections.


ihrwein:
:+1:

PR #789, opened 2015-11-17
Revert 740 f/fix lexer

### LINKED ISSUE #788: Revert PR #740 on 3.7/master

ihrwein:
Fixes #788


bazsi:
As the current master already has this code, we can just apply this reversion. So I will do that.

PR #791, opened 2015-11-17
F/native parser

ihrwein:
First I thought that the Rust/native bindings should be landed in incubator first, because of the additional dependencies. As the time went on the interface became simpler and simpler. Now, it doesn't have any additional dependency. Nevertheless, this feature is based on the current master but incubator is based on 3.7.

This is just the C part and doesn't need rustc or other Rust related tools to be installed.

The Rust part is splitted into 3 crates (compilation units) and stored in the following git repo: https://github.com/ihrwein/syslog-ng-rs

That repo also needs a thorough review.

### DIFF
--- a/modules/native/native-grammar.ym
+++ b/modules/native/native-grammar.ym
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2015 BalaBit
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+%code top {
+#include "native-parser.h"
+
+}
+
+
+%code {
+
+#include "cfg-parser.h"
+#include "parser.h"
+#include "native-grammar.h"
+#include "syslog-names.h"
+#include "messages.h"
+#include "plugin.h"
+#include "cfg-grammar.h"
+
+#include <string.h>
+
+void native_parser_set_option(LogParser *s, gchar* key, gchar* value);
+LogParser* native_parser_new(GlobalConfig *cfg);
+
+}
+
+%name-prefix "native_"
+
+/* this parameter is needed in order to instruct bison to use a complete
+ * argument list for yylex/yyerror */
+
+%lex-param {CfgLexer *lexer}
+%parse-param {CfgLexer *lexer}
+%parse-param {LogParser **instance}
+%parse-param {gpointer arg}
+
+/* INCLUDE_DECLS */
+
+%token KW_OPTION
+
+%type <ptr> native
+%type <ptr> native_parser
+%type <ptr> native_parser_params
+%%
+
+start
+  : native
+    {
+      *instance = $1;
+      if (yychar != YYEMPTY)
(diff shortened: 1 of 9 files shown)


ihrwein:
Here are two crates which already use these bindings:

https://github.com/ihrwein/dummy-parser
https://github.com/ihrwein/actiondb-parser

You just have to:

1. build and install syslog-ng from this branch
1. include `$prefix/lib/pkgconfig` in `PKG_CONFIG_PATH`
1. clone a repo from above
1. step into the repo and execute `cargo build --release`
1. copy the `.so` file from `target/release` into `$prefix/lib/syslog-ng`
1. use the module from syslog-ng's configuration:
 ```
 @version: 3.7

source s_localhost {
    network(
        ip(
            0.0.0.0
        ),
        port(
            1514
        ),
        transport("tcp")
#        flags(no-parse)
    );
};

log {
    source(
        s_localhost
    );
    parser {
        dummy-rs();
    };
    parser{
        actiondb-rs(
            option("pattern_file", "/home/tibi/install/syslog-ng/etc/loggen.json")
            option("prefix", ".adb")
        );
    };
    destination {
       file("/tmp/parsed.json"
          template("$(format-json --key .adb*)\n")
       );
    };
}; 
 ```
1. you can download the `loggen.json` pattern file from here: https://github.com/ihrwein/actiondb-docker/blob/master/loggen.json
1. update the `pattern_file` parameter in the config

modules/native/native-grammar.ym
```diff
@@ -0,0 +1,114 @@
...
+#include "messages.h"
+#include "plugin.h"
+#include "cfg-grammar.h"
+
+#include <string.h>
+
+__attribute__((visibility("hidden"))) LogParser*
+native_parser_new(GlobalConfig *cfg);
+
+}
```
ihrwein: The hidden attribute isn't needed in this case, I just left it there.

modules/native/native-grammar.ym
```diff
@@ -0,0 +1,114 @@
...
+
+native_parser_options
+  : native_parser_option native_parser_options
+  |
+  ;
+
+native_parser_option
+  : KW_OPTION '(' string string ')'
+    {
+      native_parser_set_option(*instance, $3, $4);
```
ihrwein: I wanted to have something like that but didn't know syslog-ng's grammar system enough. The rust  grammar can consume tokens/characters but I don't know how syslog-ng's grammar will react to that. Can I say that in a `x-parser() { ... }` block the `...` content should be passed to an other grammar?

bazsi: You just call the lex function which returns tokens YYSTYPE). If you can call the lexer implemented in C you will be wired into the grammar system.

In fact this could work with all the other language bindings we have so far.

Also it'd be nice if the proxy would be simpler, can't you fill the LogParser struct directly from rust?

ihrwein: I think one of the reason to use this binding it to simplify writing modules. If the rust grammar is directly hooked into C I think I lose this simplicity. We can wrap the modules in SCLs which can hide the configuration ugliness. I think a way to express a 'verbatim' text and passing it to a rust grammar would be better, but maybe the two approaches can live side by side. First I'd use the SCL and option based one and then we may come up with something smarter.

The `LogParser` struct could be filled by rust but I wanted to keep `LogParser` as an opaque pointer. I tried to minimize the size of the C side and forwarding a function call is the simplest thing I can come up with.

bazsi: The lexer/grammar interface is a single function call that returns a filled
YYSTYPE struct.

The lexer can be invoked using:


      return cfg_lexer_lex(lexer, yylval,
yylloc);

where "lexer" is passed as an argument to the grammar (in C), yylval is the
YYSTYPE instance, and yylloc is the token location; both filled by the
lexer.

Both YYSTYPE (semantic type) and YYLTYPE (location type) are simple structs:

typedef struct YYLTYPE
{
  int first_line;
  int first_column;
  int last_line;
  int last_column;
  CfgIncludeLevel *level;
} YYLTYPE;


typedef struct YYSTYPE
{
  /* one of LL_ types that indicates which field is being used */
  int type;
  union
  {
    gint token;
    gint64 num;
    double fnum;
    char *cptr;
    void *ptr;
    gpointer node;
  };
} YYSTYPE;


But you are right options can be easier for the plugin at least as long as
all parameters are strings.

In more complicated setups it definitely gets way more complicated.


-- 
Bazsi

ihrwein: I'll keep these in mind :)

ihrwein:
@lbudai I fixed your suggestions, but my force push (after rebasing to master) removed the comments.

PR #797, opened 2015-11-23
F/unit test cleanups

bazsi:
This branch resolves all but one memory leaks in the unit tests. The last one is in test_rewrite which triggers a parser error in the grammar, which in turn leaks.

I might resolve that as well, but that's a more complicated change, so I'd like to merge this as it is.


lbudai:
@bazsi: could you squash the fixups?

bazsi:
done and rebased against master.

lib/logproto/logproto-server.c
```diff
@@ -122,12 +122,20 @@ log_proto_server_init(LogProtoServer *self, LogTransport *transport, const LogPr
...
 {
-  if (self->encoding)
-    g_free(self->encoding);
+  GIConv convert;
+
+  g_free(self->encoding);
   self->encoding = g_strdup(encoding);
+
+  /* validate encoding */
+  convert = g_iconv_open("utf8", encoding);
```
bkil-syslogng: It would look nicer if we used `UTF-8` everywhere, and we could introduce a constant for that as well.

bazsi: Hmm, this is already merged. But I don't agree that it matters how we call
utf8. There are aliases for that for a reason.

It doesn't make stuff more readable, reliable or anything. On the other
hand it takes away energy to look after/validate these kind of mistakes we
would make while coding.

bkil-syslogng: In my opinion, we should review merged code just as often as new code, but
that's besides the point.

The correct name of the encoding is "UTF-8" as defined in the standards:

https://www.ietf.org/rfc/rfc2279.txt
http://standards.iso.org/ittf/PubliclyAvailableStandards/index.html
(ISO/IEC 10646)
https://en.wikipedia.org/wiki/UTF-8#External_links

libiconv, for example, won't accept "utf8" as an encoding at all. If we had
extracted this constant earlier, this typo would not have occurred.

http://git.savannah.gnu.org/cgit/libiconv.git/tree/lib/encodings.def?id=48f31c744eb14da699c6ba3fd03f5f232c476bbb#n60

bazsi: I agree that code should be under scrutiny and going back and finding how a
bug could have been prevented is a worthwhile exercise.

However it'd have been nice if you provided this context in the note. I
didn't know that there was a failure of some kind. I still don't get what
the bug is that triggered this comment as without  that I cant judge the
recommended solution.

You surprise me that iconv wouldnt understand "utf8" as an encoding. I use
that pretty regularly and so far it worked just fine, but maybe I was
completely mislead.

I still think that adding a define and using it everywhere will not resolve
the issue, because:

1) then you'll have to remember that there's this constant. and newcomers
won't.

2) in reviews you'd want to flag errors if that  happens, thus everyone
will have to focus on that

3) if you focus on these kind of things you'll not get to the real
important stuff, like layering, interfaces and duplicate implementations

The solution should be a failing in test. And again, can you pls tell me
what failed which triggered this review so we can fix the test instead?

Thanks

lbudai: @bazsi:

Here is the 'internal' 'bug':
```
seems utf-8 encoding does not work.
[2016-05-03T11:43:11.043665] Module loaded and initialized successfully; module='afprog'
Error parsing afprog, unknown encoding UTF-8 in /opt/syslog-ng/etc/syslog-ng.conf at line 6, column 88:
program("/var/testdb_working_dir/90ee8394-0d84-402d-933b-c7997a6dc3e7.sh" encoding(UTF-8));
config:
@version: 3.8
options
{ threaded(yes); }
;
source s_program_37b57ebaaa6a460085d2d35d6f7d56ec
{ program("/var/testdb_working_dir/644aefb0-504c-429f-8022-ce3150b60b75.sh" encoding(UTF-8)); }
;
destination d_file_a3219214c5e541d5a18b0e9ca3eedee2
{ file("/var/testdb_working_dir/81525975-13e5-4a41-8e9e-1bbd7cdc3d25.txt"); }
;
log
{ source(s_program_37b57ebaaa6a460085d2d35d6f7d56ec); destination(d_file_a3219214c5e541d5a18b0e9ca3eedee2); flags(flow-control); }
;
```

@bkil-syslogng:
To be honest, the way you are handle this issue is not the best...
We found something during internal tests and _before_ the validation and tests you are
blaming an already-merged PR (without sharing all the information you have). 

Even if the bug is valid and your fix resolves the issue, we have 
a well defined process that you should follow. When we found 'something' and we validated that this is a bug then we open a corresponding github issue (or, it is also fine if you open the github issue before the validation), the point is that we share _everything_ we know about the issue and when everyone has the chance at being fully informed then we can start the discussion.

bazsi: I think the problem is that we want to open an utf8 to utf8 encoding and
that fails.

lbudai: It could be a problem... (however, I have a SUTC - small&ugly test code and it works when `tocode` and `fromcode` is the same...) Maybe different library versions...

In libiconv, they convert the character encoding names to upper case, so the solution is definitely not what @bkil-syslogng suggested.
Comment from libiconv we have:
`Before calling aliases_lookup, convert the input string to upper case`

bkil-syslogng: I have proposed a recommended improvement in our process that is completely
orthogonal to the bug.

Please move the discussion relating to this bug to its own issue. (My
solution was not making it upper case, we were reading the same code, so
you should know that. And by the way, the problem has been solved (and/or
worked around), so I'm not sure what @lbudai is investigating.) I'd also
thank if we refrained from personal attacks. Comment on the solution, not
on the person.

I still think that we should use atoms whenever feasible. If we move to
UTF-16, should we replace all these strings? I'd actually introduce a layer
for internal conversions instead of these manual calls, so I'm open to all
suggestions.

bkil-syslogng: @bazsi: So theoretically this implementation and using aliases should work
flawlessly, but if I see symmetry breaking, I always ask for the reason,
and we better have one in all cases.

I don't see referring to the encoding as both utf-8 and utf8 in our code
justified in this case, so we should decide which one to use. If a
canonical name is available, we should use that, so it's pretty clear cut
in this case

bkil-syslogng: By the way, we wanted to see where this came from because we were puzzled
whether this never worked in OSE, or it has regressed recently. So the
reference is not a pointless exercise of mine or blaming.

lbudai: @bkil-syslogng : my main problem is the way how you communicated all this staff.
I won't repeat myself (key point: share all the information you have such a case like this and don't let others working on an issue that is already resolved).

Some technical details:
 * glib can use the libiconv or the glibc iconv when it is available

The glibc iconv accepts UTF-8, utf-8, UTF8 and even utf8 but the libiconv implementation
does not accept encoding strings without '-'.

We agreed on that we will open a github issue and @bkil-syslogng will send a fix.
The solution is that we won't use 'utf8' but 'utf-8'...

Issue #796, opened 2015-11-23
template-function is an unknown keyword with `@version: 3.6`


vincentbernat:
When upgrading from 3.6 to 3.7, I get the following error:

```
Error parsing config, root plugin template-function not found in /home/bernat/.virtualenvs/syslog-ng/share/include/scl/cim/template.conf at line 23, column 1:
                                                   included from /home/bernat/.virtualenvs/syslog-ng/etc/scl.conf line 32, column 1
                                                   included from /home/bernat/Documents/exoscale/S03-2015/syslog-ng.conf line 4, column 1

template-function "format-cim" "$(format-json --pair @timestamp='${R_ISODATE}' --pair @message='${MSG}' --key .cim.* --shift 5 --key _* --key .* --replace-prefix .=_ --key *.*)\n";
^^^^^^^^^^^^^^^^^
```

The `template-function` keyword is unknown when using `@version: 3.6`. I didn't find a way to make this part of the configuration active only when `@version` is 3.7 or more recent.

faxm0dem:
This one has become a classic at $WORK
:+1:

bazsi:
Hi,

I've improved my original version (from here:
https://lists.balabit.hu/pipermail/syslog-ng/2015-March/021932.html), which
you can find on this branch:

https://github.com/balabit/syslog-ng/tree/f/allow-new-keywords-with-old-config-versions

Since it's a largish change I would love to see actual, real-world configs
be tested with this patch applied. a "syslog-ng -s" to check syntax should
suffice. Of course careful code review might also prove it is correct.

Feedback appreciated.
@lbudai, @ihrwein


-- 
Bazsi

ihrwein:
Bazsi's mentioned branch has been merged. @vincentbernat could you close the issue, please?

vincentbernat:
Ok!

phemmer:
What version is the fix included in? The linked branch no longer exists, so I cant search for the code to see if it's present.
I'm running syslog-ng 3.7.3 and still seeing this issue.

MrAnno:
Hello @phemmer,

PR #828 should be the one you are looking for. It's not included in 3.7.3.

PR #800, opened 2015-11-24
allow using new keywords in older config versions

bazsi:
The SCL does contain new keywords and doing conditionals there is more than difficult, not to mention that there's really no point.

This patch series tries to push all keywords to external modules and clean up the core keyword set (e.g. the null keyword that historically caused the entire keyword versioning is pushed down to csv and sql modules), which hopefully decreases the chances of a conflict between a keyword and a user identifier.

With all this in place, the patchset also removes keyword versioning entirely, eg. we would accept template-function even if we are running an older config version.

The changes are fragile, as there's a lot of boilerplate in this code and it is difficult to test, so I'd request careful review and if at all possible trying to use the new code with existing configurations.


bazsi:
This should fix #796

faxm0dem:
> With all this in place, the patchset also removes keyword versioning entirely

What user-perspective consequences does this imply?

bazsi:
Sorry for not responding sooner with regards to the user perspective.

There shouldn't be too much of an issue really, before this patch keywords that were introduced later than the "@version" tag  were not recognized as keywords and assumed to be mere identifiers.

This was implemented as at one point SLES had a default syslog-ng configuration that had a "null" destination, and with syslog-ng 3.0 we introduced "null" as a keyword causing parsing failures in SLES:

```
destination null {
};
```
When we introduced the "null" keyword this caused the above configuration to fail.

The use of identifiers in the syslog-ng configuration grammar is not really widespread, they are only used as the names of sources/destinations/filters/parsers/rewrite rules, sometimes in generic options

By pushing most of the keywords to syslog-ng plugins, the collisions they would introduce are most likely avoided, as most new functionality will be pushed into plugins. By the time the parser of the plugin starts we don't often use identifiers.

And the conflict can always be worked around by placing the relevant token in quotes:

```
destination "null" {
};
```

With all this said, I think it is not really a risky change, apart from any implementation mistakes I may have made, but I reviewed stuff a couple of times and I invited the others to do the same.

ihrwein:
:+1:

bazsi:
@ihrwein if you are confident enough please merge it as well, or will @lbudai have a peek as well?

faxm0dem:
thanks @bazsi for the explanation

ihrwein:
@bazsi I'd wait till @lbudai take a look at it. After that I'll merge it.

lbudai:
Hi,

I like this idea and I think, the code is OK.

Before we merge it, we should create an installer and give it to PZoli to run some(*lots* of) tests on it.

lbudai:
I've created a `f/allow-new-keywords-with-old-config-versions-rebased-to-master` branch: there was a conflict in `modules/dbparser/dbparser-parser.c`, I've resolved it (didn't want to force push your branch).

bazsi:
Feel free to force push my branches in this case. A comment would be nice,
but you can do it, do we won't have to bounce branches and PRs around.

lbudai:
@bazsi: ok, then next time we will do it as you suggested.

ihrwein:
I'm closing this as the rebased version is merged :)

Issue #776, opened 2015-11-10
Eliminate compiler warnings and turn on -Werr


ihrwein:
I used the master branch.

```
../modules/afprog/afprog.c: In function 'afprogram_reload_store_item_free':
../modules/afprog/afprog.c:69:3: warning: passing argument 1 of 'log_pipe_unref' from incompatible pointer type [enabled by default]
   log_pipe_unref(reload_info->writer);
   ^
In file included from ../lib/driver.h:29:0,
                 from ../modules/afprog/afprog.h:27,
                 from ../modules/afprog/afprog.c:24:
../lib/logpipe.h:237:6: note: expected 'struct LogPipe *' but argument is of type 'struct LogWriter *'
 void log_pipe_unref(LogPipe *self);
      ^
../modules/afprog/afprog.c: In function 'afprogram_dd_kill_child':
../modules/afprog/afprog.c:329:13: warning: unused variable 'pgid' [-Wunused-variable]
       pid_t pgid;
             ^
../modules/afprog/afprog.c: In function 'afprogram_dd_store_reload_store_item':
../modules/afprog/afprog.c:452:3: warning: passing argument 4 of 'cfg_persist_config_add' from incompatible pointer type [enabled by default]
   cfg_persist_config_add(cfg, afprogram_dd_format_persist_name(self), reload_info, afprogram_reload_store_item_destroy_notify, FALSE);
   ^
In file included from ../lib/logpipe.h:30:0,
                 from ../lib/driver.h:29,
                 from ../modules/afprog/afprog.h:27,
                 from ../modules/afprog/afprog.c:24:
../lib/cfg.h:158:6: note: expected 'GDestroyNotify' but argument is of type 'void (*)(struct AFProgramReloadStoreItem *)'
 void cfg_persist_config_add(GlobalConfig *cfg, gchar *name, gpointer value, GDestroyNotify destroy, gboolean force);

```

To do:
- [x] Eliminate compiler warnings - #779
- [ ] Turn on -Werr #871

bazsi:
I saw these too, and this is pretty unfortunate as these are recent
additions.


-- 
Bazsi

lbudai:
ehh... I think we should do something with these warnings :-(
Maybe Travis/drone.io and -Werr could help ... but we are far from that.
(or create a check-warning target in the makefile)

bazsi:
I think the OSE tree is not that far, we only have a handful of warnings,
[...]

PR #802, opened 2015-11-24
cfg-lex.l: don't report yy_fatal_error() as an unused symbol

bazsi:
The code generated by flex contains yy_fatal_error() which is not used in
the generated code. Insert an artificial reference so that it doesn't
get reported.

### DIFF
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -331,4 +331,4 @@ cfg_lexer_start_block_state(CfgLexer *self, gchar block_boundary[2])
 }
 
 /* avoid warnings of unused symbols */
-gpointer __dummy[] = { yy_top_state };
+gpointer __dummy[] = { yy_top_state, yy_fatal_error };


bazsi:
This one should complete #776

bazsi:
This is a trivial change, one that solves this warning the same way as an earlier warning was solved.

I am merging this now as this has been sitting in the queue for a while now, there are other stuff to review and I don't want to waste review bandwidth on this one.

Let me know if this is something you wouldn't want me to do in the future.

bkil-syslogng:
We might consider reusing the labels easy/medium/hard to classify how much effort a review is projected to take.

For one-liners or other trivialities, we should tag `easy`. If you need to sit down for up to a day, take a deep look, reconsider the higher level design decisions and test the code yourself, you should tag it `hard`. `Medium` should be something in between.

Then anyone having some free time could grab the **oldest** PR (or issue) from the table depending on their free time:
* an `easy` one: if you have 15-30min to spare or are doing some other light activities;
* a `medium` one: if you have about an hour and can concentrate

I propose that we should introduce new backlog items for `hard` reviews, prioritize and schedule them similar to how we do it for other tasks.

ihrwein:
:+1:

bazsi:
it's fine by me. review bottlenecks are a problem in my opinion right now. I can see 30 opened pull requests which is a lot.

PR #810, opened 2015-11-25
Removing directives that split up parts of statements.

flaviommedeiros:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/lib/template/macros.c
+++ b/lib/template/macros.c
@@ -192,6 +192,15 @@ _result_append_value(GString *result, const LogMessage *lm, NVHandle handle, gbo
   result_append(result, str, len, escape);
 }
 
+gboolean sockaddr_inet_check(const LogMessage *msg)
+{
+#if SYSLOG_NG_ENABLE_IPV6
+    return msg->saddr && (g_sockaddr_inet_check(msg->saddr) || g_sockaddr_inet6_check(msg->saddr));
+#else
+    return msg->saddr && g_sockaddr_inet_check(msg->saddr);
+#endif
+}
+
 gboolean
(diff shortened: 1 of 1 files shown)


lib/template/macros.c
```diff
@@ -339,14 +339,15 @@ log_macro_expand(GString *result, gint id, gboolean escape, const LogTemplateOpt
...
 #if SYSLOG_NG_ENABLE_IPV6
-            g_sockaddr_inet6_check(msg->saddr))
+        checker = checker && (g_sockaddr_inet_check(msg->saddr) || g_sockaddr_inet6_check(msg->saddr));
 #else
-            0)
+        checker = checker && g_sockaddr_inet_check(msg->saddr);
 #endif
-           )
+        if(checker)
           {
```
bazsi: I agree with the change in general however, we should use better naming of this variable.

Also checker = msg->saddr completely disregards typing. So apart from the name it should be a Boolean.

And if we are at it, I would try to extract the ipv4 or ipv6 check to a separate function

flaviommedeiros:
Is that what you meant, @bazsi?

bazsi:
This looks great now, now please merge the three patches into one and sign it off so I can merge it.

Thanks.
Bazsi

bazsi:
will you do the rebase or shall I do it?

flaviommedeiros:
@bazsi, please do the rebase. I do not know exactly what to do.

bazsi:
I am closing this in favour of pr #830

PR #813, opened 2015-11-28
tests: made func tests a bit flexible

lmesz:
With this minimal fix in tests, syslog-ng can be installed on arbitary place in a host. This place can be passed via SYSLOG_NG_BINARY and functional test can be executed like before.

### DIFF
--- a/lib/plugin.c
+++ b/lib/plugin.c
@@ -480,6 +480,12 @@ plugin_free_candidate_modules(GlobalConfig *cfg)
   cfg->candidate_plugins = NULL;
 }
 
+const gchar*
+plugin_get_module_path()
+{
+    return module_path;
+}
+
 void
 plugin_list_modules(FILE *out, gboolean verbose)
 {
--- a/tests/functional/globals.py
+++ b/tests/functional/globals.py
@@ -1,13 +1,38 @@
 import os
 
+def is_running_in_build_tree():
+    return 'SYSLOG_NG_BINARY' not in os.environ
+
+def get_module_path_from_binary():
+    syslog_ng_version_info = os.popen("%s --version" % get_syslog_ng_binary(), 'r').read()
+    for line in syslog_ng_version_info.split("\n"):
+        if 'Module-Path' in line:
+            return line.split(" ")[-1]
+    raise Exception("Failed to determine module paths from syslog-ng version")
+
+def format_module_path_for_intree_modules():
+    module_path = ''
+    for (root, dirs, files) in os.walk(os.path.abspath(os.path.join(os.environ['top_builddir'], 'modules'))):
+        module_path += ':'.join(map(lambda x: root + '/' + x + '/.libs', dirs))
+        break
+    return module_path
+
+def get_module_path():
+    if is_running_in_build_tree():
+        return format_module_path_for_intree_modules()
+    return get_module_path_from_binary()
+
+def get_syslog_ng_binary():
+    return os.getenv('SYSLOG_NG_BINARY', '../../syslog-ng/syslog-ng')
+
 def is_premium():
-    version = os.popen('../../syslog-ng/syslog-ng -V', 'r').read()
+    version = os.popen('%s -V' % get_syslog_ng_binary(), 'r').read()
     if version.find('premium-edition') != -1:
         return True
     return False
 
 def has_module(module):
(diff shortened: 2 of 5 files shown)


bazsi:
I am not against such a change, but we originally intended to run the functest in the build environment.  In fact this is nice as the installed version can be tested easily. 

I think the module-path calculation needs to be changed as well though.

lmesz:
Thanks for the fast response and info.
I understand the base concept, and I didn't want to go againt that, but I would like to write func test for one of the module in incubator and didn't want to copy or reimplement the test infrastructure.

And thanks modify the determination of module path.

lmesz:
Hello,

Based on your suggestion changed the module path also.

bazsi:
Almost there. 
It would be better to name get_modules () as get_module_path() or something.

Also, the installation path shouldn't be hard coded in the test script. The syslog-ng binary should tell it to us. It should be added to --version or --module-registry

And last but not least the installed modules should only be used iff we are not building intree. As if I have syslog-ng on my system where I also develop it, it's load the system supplied one instead of my locally compiled one

lmesz:
I made some modifications that you mentioned before. One remained I will try to fix it on tomorrow.

lib/plugin.c
```diff
@@ -481,6 +481,12 @@ plugin_free_candidate_modules(GlobalConfig *cfg)
 }
 
 void
+plugin_get_module_path(FILE *out)
+{
+    fprintf(out, "Module path: %s\n", module_path);
+}
+
```
bazsi: well, if it is called a getter it should be a getter and not printing it.

bazsi: also, the module_path global is pretty sure to be at the wrong location in mainloop.c and should probably be pushed down to plugin.c, but that's a separate issue.

tests/functional/globals.py
```diff
@@ -1,13 +1,27 @@
 import os
 
+def get_builtin_modules():
+    module_path = os.popen("%s --version | grep \"Module Path\" | cut -d ' ' -f 3" % get_syslog_ng_binary(), 'r').read().split(".")
+    return '%s:' % module_path
+
```
bazsi: while reviewing the patches I see two separate interpretation of the term "module_path" which should be separated:

1) it is used for the default module directory as used by "make install"
2) it is used to store the list of directories where modules are searched for.

The 2nd usually includes the first, but probably the name should indicate which one we are working on:

"module_dir" should be used for installation directory
"module_path" should only be used for the colon separated list of directories.

tests/functional/globals.py
```diff
@@ -1,13 +1,27 @@
 import os
 
+def get_builtin_modules():
+    module_path = os.popen("%s --version | grep \"Module Path\" | cut -d ' ' -f 3" % get_syslog_ng_binary(), 'r').read().split(".")
+    return '%s:' % module_path
```
bazsi: We should probably use a dash in the name of "Module-Path", or rather Module-Directory and also a colon to keep the rfc822 style formatting of --version output.

lmesz: Fixed it and changed from "Module Path" to "Module-Directory:"

lib/plugin.c
lmesz: Should I open an issue about that ?

tests/functional/globals.py
lmesz: Okay sorry yes that's true. I fixed the names based on your  instructions. From now on I guess it is clear which functions would do what.

bazsi:
Hi,

I have refined your patches on my branch named lmesz/remove-fix-relative-path-of-syslog-ng-in-func-test

Please let me know if you are fine with the changes and if you are pls modify your pull request.

Cheers,
Bazsi

lmesz:
Thanks for the fixups, I modified my branch based on what you did.

lmesz:
Sorry rebased and fixed how to find module path in version info

bazsi:
The python part looks now ok, but the C part is still missing some of the pieces. If you want I can prepare a branch that contains both your code and mine. What do you think?

lmesz:
Of course it would be good. Thank you.
My goal at the end of the PR is to use this func-test ecosystem to test a module in incubator.
Thanks your help again.

bazsi:
this should be closed without merging once #858 is merged.

bazsi:
I think I am closing this instead to clear up clutter.

PR #816, opened 2015-12-01
SourcePortMacro: Added support for $SOURCEPORT macro

mitzkia:
* With this feature we will be able to parse PROXY protocol v1 messages

### DIFF
--- a/lib/gsockaddr.c
+++ b/lib/gsockaddr.c
@@ -217,7 +217,7 @@ g_sockaddr_inet_format(GSockAddr *addr, gchar *text, gulong n, gint format)
  * Returns: the port in host byte order
  *
  **/
-static guint16
+guint16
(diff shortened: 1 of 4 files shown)


bazsi:
hm.. what's the usecase here? what are those PROXY protocol v1 messages, I don't get it, sorry.

Also implementation wise this doesn't work with ipv6.

mitzkia:
Sorry about that. The proposal for the feature already created, but for now it is only internally available. I will send you an URL for it. Of course after the proposal is properly created it will be open for the community. Thanks for mentioning ipv6, i will complete the patch with it.

bazsi:
Hmm why not copy paste the proposal here, so that others could see it as
well?

mitzkia:
Because the code and the proposal is not yet ready I suggest to close this PR.

PR #818, opened 2015-12-01
syslog-format: fix BSD timestamp parsing

ihrwein:
Fixes the broken master (it's December...)

Signed-off-by: MÓZES Ádám István <adam.mozes@balabit.com>

### DIFF
--- a/modules/syslogformat/syslog-format.c
+++ b/modules/syslogformat/syslog-format.c
@@ -358,7 +358,9 @@ __parse_bsd_timestamp(const guchar **data, gint *length, const GTimeVal *now, st
   gint left = *length;
   const guchar *src = *data;
   time_t now_tv_sec = (time_t) now->tv_sec;
+  struct tm local_time;
   cached_localtime(&now_tv_sec, tm);
+  cached_localtime(&now_tv_sec, &local_time);
 
   if (__is_bsd_pix_or_asa(src, left))
     {
@@ -383,7 +385,7 @@ __parse_bsd_timestamp(const guchar **data, gint *length, const GTimeVal *now, st
 
       *usec = __parse_usec(&src, &left);
 
-      tm->tm_year = determine_year_for_month(tm->tm_mon, tm);
+      tm->tm_year = determine_year_for_month(tm->tm_mon, &local_time);
     }
   else
     {


dnsjts:
This pull request fixes a quite twisted (and incorrect) logic of the BSD year inference.
Next time I shouldn't use the parsed timestamp as the current time for `determine_year_for_month`.

ihrwein:
@lbudai @bazsi This bug breaks the master branch and so every other PRs. Please review it so the other PRs can be rebased.

bazsi:
This looked suspicious even when I copy pasted it to date parser, which has the same logic.

I don't have my computer handy so I really can't review the change in detail, however this was converted to this logic by Adam, and was a very different logic which endured a number of Decembers.

So I would go back to the original solution, time is difficult to get right.

ihrwein:
This was the original code (embedded in syslog-format):
```C
          /* detect if the message is coming from last year. If its
           * month is at least one larger than the current month. This
           * handles both clocks that are in the future, or in the
           * past:
           *   in January we receive a message from December (past) => last year
           *   in January we receive a message from February (future) => same year
           *   in December we receive a message from January (future) => next year
           */
          if (tm.tm_mon > nowtm.tm_mon + 1)
            tm.tm_year--;
          if (tm.tm_mon < nowtm.tm_mon - 1)
            tm.tm_year++;
```

And this is the refactored, currently used code:
```C
/* Determine (guess) the year for the month.
 *
 * It can be used for BSD logs, where year is missing.
 */
gint
determine_year_for_month(gint month, const struct tm *now)
{
  if (month == 11 && now->tm_mon == 0)
    return now->tm_year - 1;
  else if (month == 0 && now->tm_mon == 11)
    return now->tm_year + 1;
  else
    return now->tm_year;
}
```
The bug is in the call site, not in this function. If I remember correctly, there was a bug in the original code as well, but the comments were correct.

I'd keep the new changes, because `syslog-format` was refactored, but write some unit tests for this `determine_year_for_month()` function. What do you think?

bazsi:
There was a change prior to this extraction.

dnsjts:
The wrong behaviour was introduced it in 27259e93d090031c62a3f13ac2d3e2eb5ea694f2.
Here is the wrong sequence of the calls:

1. `cached_localtime(&now_tv_sec, tm);`
2. `scan_bsd_timestamp((const gchar **) &src, &left, tm)`
3. `tm->tm_year = determine_year_for_month(tm->tm_mon, tm);`

The second parameter of the `determine_year_for_month` is the current time. This is no longer true in this sequence.

lbudai:
Hi,

@dnsjts is right, the bug is from a PE refact.
Adam didn't touch the logic, just make it more understandable.

I'd merge this fix as-is, but the main question for me: do we really need this heuristic? :-)

PR #828, opened 2015-12-08
F/allow new keywords with old config versions rebased to master

lbudai:
original PR (#800) created by bazsi:

The SCL does contain new keywords and doing conditionals there is more than difficult, not to mention that there's really no point.

This patch series tries to push all keywords to external modules and clean up the core keyword set (e.g. the null keyword that historically caused the entire keyword versioning is pushed down to csv and sql modules), which hopefully decreases the chances of a conflict between a keyword and a user identifier.

With all this in place, the patchset also removes keyword versioning entirely, eg. we would accept template-function even if we are running an older config version.

The changes are fragile, as there's a lot of boilerplate in this code and it is difficult to test, so I'd request careful review and if at all possible trying to use the new code with existing configurations.


ihrwein:
We may merge this today.

lbudai:
@ihrwein : you can merge it :-)

lbudai:
(and don't forget to close #800 )

PR #823, opened 2015-12-02
F/copyright review 2015 extended

bkil-syslogng:
This is the extended version of the copyright cleanup PR #822. Theoretically all previous occurrences have been replaced with the new convention of Copyright (c) 2015 Balabit, but do comment if you this it has an error. It is enough to merge f/copyright-review-2015-extended, because it also contains f/copyright-review-2015.

Fixes #824

PR #835, opened 2015-12-15
travis: run distcheck and func-tests parallely

ihrwein:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: c
+env:
+  - TEST_SUITE=distcheck
+  - TEST_SUITE=func-test
 install:
   - sudo apt-get update -qq
   - sudo apt-get install -qq curl
@@ -12,7 +15,7 @@ before_script:
 script:
   - unset PYTHON_CFLAGS # HACK
   - ./configure --with-ivykis=internal --prefix=$HOME/install/syslog-ng --enable-pacct --enable-manpages --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl 
-  - make && make distcheck VERBOSE=1 && make func-test VERBOSE=1
+  - make && make $TEST_SUITE VERBOSE=1
 compiler:
   - gcc
   - clang


ihrwein:
Python, don't do this with us!

bkil-syslogng:
I like the idea, however how much do we expect to gain? It starts 4 jobs
per commit instead of 2, and does all parts before the tests twice
(install, autogen, configure, make), which isn't very nice in an energy
efficiency standpoint. I guess it might also increase queuing time as well,
although in your test run, all started up pretty quickly.

This is not to detract you from speeding up CI, but did you look into the
new Travis infrastructure and certain cases of caching options?

ihrwein:
You are right, but I wanted to see how much we can gain with splitting up the testing process into two independent halves. I'll look at the container based infra:tm: as well.

bkil-syslogng:
Nevertheless, I'd still be interested in the build time improvements that this brings us if you can fix the Python issue. Because of linear dependencies, this would be the total latency of a more efficient parallel build anyway, so it'd be good to now its length.

As a side note, I have good success with `make -j` for OSE. Although, I see how it can make it harder to read error messages, that's why I have a longish meta-builder script which boils down to something like make -j || make VERBOSE=1.

Actually I support incremental builds as well, when augmented with proper full build scheduling.

bkil-syslogng:
By the way, although a container-based caching setup would be the fastest, if you find it to be infeasible, we could still receive a free speed benefit by migrating to `TRUSTY BETA`:
https://docs.travis-ci.com/user/ci-environment/

bkil-syslogng:
This should come in handy:
https://github.com/rande/travis-kitchensink/blob/48e5915be37a4141dc59a8fd5c06b807c4abbab8/.travis.yml

The extra dependency which is missing from the repo should be built from source and cached separately, I guess. I guess the other APT stuff should be compatible.

bkil-syslogng:
@ihrwein You can access the Python fix in #837. I'm still experimenting with Container and Trusty builds, you can check out a very WIPpy, constantly rebased status over here:
https://github.com/bkil-syslogng/syslog-ng/pull/5/files

A container indeed starts up in 5-10 seconds, while a VM needs about 2 minutes. Otherwise I couldn't see a speed benefit despite the full dual cores. I haven't done enough fine tuning of caching yet, although my expectations are low, because S3 has very low bandwidth.

ihrwein:
This should pass now. I'm curious about the build & test time.

ihrwein:
func-test: 5-6 mins
distcheck: 7 mins

bkil-syslogng:
@ihrwein I've given it a second thought on the way home. My take is that we can run them in parallel by assigning one to gcc, the other to clang. They should be matched to take a similar amount of time. I can elaborate on why I consider this an elegant and effective solution if you are interested.

ihrwein:
The all-in-one pass is around 8.5 mins.

bkil-syslogng:
Also enabled in #823 in a more efficient way as discussed last month

bkil-syslogng:
#823 has been merged which implements it as follows:
```
if [ "$CC" = "gcc" ]; then make distcheck V=1; else make func-test V=1; fi
```

I.e., it runs on on the gcc thread, and the other on the clang thread. Please double check whether it seems fast enough for you and tell us your opinion about such an optimization. If you are happy with the current form, we can close #835, otherwise please share some more input.

bkil-syslogng:
@ihrwein Is this issue still valid?

ihrwein:
No, it's not valid. I'm closing this.

Issue #808, opened 2015-11-25
OS X ld does not support --whole-archive option, thus breaking the build of 3.7.2 on OS X


douglas-carmichael:
The first issue I dealt with trying to build 3.7.2 on OS X is that ld on OS X does not support --whole-archive.

There is a patch to fix it as part of [MacPorts ticket #43231](https://trac.macports.org/attachment/ticket/43231/patch-configure.ac.diff). Should this be applied to the main repository?

ihrwein:
LGTM. Could you open a PR, please?

bazsi:
Maybe we should start extracting this stuff into m4 macros in the m4
directory, perhaps check if there's OK one good enough in the autoconf
archive.

Our configure script is huge and difficult to navigate.

douglas-carmichael:
@ihrwein How do I create the PR with the patch attached? It seems creating PRs is about comparing *branches.*

ihrwein:
You are right :) The process is as follows:

1.  checkout the current master branch
1. create a new branch, for example `f/fix-whole-archive-on-osx`
1. apply those patches in your local syslog-ng clone
1. commit the changes
1. push the changes into github
1. compare your branch with our master branch

You will see, that the only differences are your new commits.

douglas-carmichael:
@ihrwein I just did all that up until step #5, but got a 403 error trying to push the changes:

`````
remote: Permission to balabit/syslog-ng.git denied to douglas-carmichael.
fatal: unable to access 'https://github.com/balabit/syslog-ng.git/': The requested URL returned error: 403
`````

Should I just send one of you the patch file by email?

ihrwein:
I'd still prefer the pull requestf, because somebody has to do it in the end, and you are almost there :)

You cannot push directly into our repository. The right way is to fork this repo, then push your branch into your forked repo. After that, you can open a PR.

If you need help with the git commands please let me know.

douglas-carmichael:
@ihrwein I'd appreciate a git "cookbook", thanks. :)

ihrwein:
Here are the steps:

1. fork this repo (py pressing the `Fork` button in the upper right corner)
1. step into your local cloned repository (the one where you created you branch)
1. execute `git remote add douglas-carmichael git@github.com:douglas-carmichael/syslog-ng.git`
1. execute `git fetch --all`
1. execute `git push douglas-carmichael f/fix-whole-archive-on-osx`
[...]

Issue #806, opened 2015-11-25
Absence of clock_gettime() on OS X breaks build of lib/timeutils.c on 3.7.2


douglas-carmichael:
System: OS X 10.11, Apple LLVM version 7.0.0 (clang)

After I temporarily removed the alias definitions (reference issue #805) from lib/logmsg.c, I attempted to build and then ran into this error with lib/timeutils.c:
````
  CC       lib/lib_libsyslog_ng_la-timeutils.lo
lib/timeutils.c:309:7: warning: implicit declaration of function 'clock_gettime'
      is invalid in C99 [-Wimplicit-function-declaration]
      clock_gettime(CLOCK_MONOTONIC, &start);
      ^
lib/timeutils.c:309:21: error: use of undeclared identifier 'CLOCK_REALTIME'
      clock_gettime(CLOCK_MONOTONIC, &start);
                    ^
./lib/compat/time.h:32:25: note: expanded from macro 'CLOCK_MONOTONIC'
#define CLOCK_MONOTONIC CLOCK_REALTIME
                        ^
lib/timeutils.c:317:21: error: use of undeclared identifier 'CLOCK_REALTIME'
      clock_gettime(CLOCK_MONOTONIC, &stop);
                    ^
./lib/compat/time.h:32:25: note: expanded from macro 'CLOCK_MONOTONIC'
#define CLOCK_MONOTONIC CLOCK_REALTIME
                        ^
1 warning and 2 errors generated.
`````

I found two links on StackOverflow that provide alternatives for CLOCK_MONOTONIC on OS X.
[clock_gettime alternative in Mac OS X](http://stackoverflow.com/questions/5167269/clock-gettime-alternative-in-mac-os-x)
[Monotonic clock on OSX](http://stackoverflow.com/questions/11680461/monotonic-clock-on-osx)

Would there be an idea as to which alternative would work best for syslog-ng?

**Comment by ihrwein** [comment]
@bkil-syslogng you are in the middle of the compat thing. Could you check this one, please? :)

**Comment by douglas-carmichael** [comment]
@gregory094 Is this issue addressed in your OS X branch (and could your patches, if any, be pushed to master?)

**Comment by gregory094** [comment]
This problem is solved on my OSX branch. What do u mean "be pushed to the master"? What `master`? Why? U can simply get my branch and merge it with `3.8`.

**Comment by jialongl** [comment]
Looks like macOS Sierra 10.12 (16A323) has brought `clock_gettime` back:

```
[...]

PR #840, opened 2015-12-17
OS X support for syslog-ng `3.8`

ngergo:
I've started investigating the issues with `3.8` that prevents `syslog-ng` compiling on OS X platforms. 

Issues solved:
- #806 
- #808 

Please consider it as an **experimental** branch!


kira-syslogng:

presidento:
It is already implemented and on master.

PR #842, opened 2015-12-18
F/format cef extension

bkil-syslogng:
[Review on Reviewable](https://reviewable.io/reviews/balabit/syslog-ng/842)

### DIFF
--- a/lib/value-pairs.c
+++ b/lib/value-pairs.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011-2014 BalaBit IT Ltd, Budapest, Hungary
+ * Copyright (c) 2011-2015 Balabit
  * Copyright (c) 2011-2014 Gergely Nagy <algernon@balabit.hu>
  *
  * This library is free software; you can redistribute it and/or
@@ -979,6 +979,36 @@ vp_cmdline_rekey_verify (const gchar *key, ValuePairsTransformSet *vpts,
   return vpts;
 }
 
+static gboolean
+vp_cmdline_parse_subkeys(const gchar *option_name, const gchar *value,
+                       gpointer data, GError **error)
+{
+  gpointer *args = (gpointer *) data;
+  ValuePairs *vp = (ValuePairs *) args[1];
+  ValuePairsTransformSet *vpts = (ValuePairsTransformSet *) args[2];
+
+  GString *prefix = g_string_new(value);
+  g_string_append_c(prefix, '*');
+  value_pairs_add_glob_pattern(vp, prefix->str, TRUE);
+
+  vp_cmdline_start_key(data, prefix->str);
+
+  vpts = vp_cmdline_rekey_verify(prefix->str, vpts, data);
+  if (!vpts)
+    {
+      g_set_error(error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
+                   "Error parsing value-pairs: --subkeys failed to create key");
+      g_string_free(prefix, TRUE);
+      return FALSE;
+    }
+
+  value_pairs_transform_set_add_func
+    (vpts, value_pairs_new_transform_replace_prefix(value, ""));
+
+  g_string_free(prefix, TRUE);
+  return TRUE;
+}
+
 /* parse a value-pair specification from a command-line like environment */
 static gboolean
 vp_cmdline_parse_scope(const gchar *option_name, const gchar *value,
@@ -1243,6 +1273,8 @@ value_pairs_new_from_cmdline (GlobalConfig *cfg,
       NULL, NULL },
     { "replace", 0, G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_CALLBACK,
       vp_cmdline_parse_rekey_replace_prefix, NULL, NULL },
+    { "subkeys", 0, 0, G_OPTION_ARG_CALLBACK, vp_cmdline_parse_subkeys,
+      NULL, NULL },
     { G_OPTION_REMAINING, 0, 0, G_OPTION_ARG_CALLBACK, vp_cmdline_parse_pair_or_key,
       NULL, NULL },
     { NULL }
--- a/lib/value-pairs.c
+++ b/lib/value-pairs.c
@@ -952,6 +952,33 @@ vp_cmdline_parse_rekey_finish (gpointer data)
   args[3] = NULL;
 }
 
+static void
+vp_cmdline_start_key(gpointer data, const gchar *key)
+{
+  gpointer *args = (gpointer *) data;
+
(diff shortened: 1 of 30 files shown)


lib/value-pairs.c
```diff
@@ -1,6 +1,7 @@
 /*
- * Copyright (c) 2011-2014 BalaBit IT Ltd, Budapest, Hungary
+ * Copyright (c) 2011-2015 Balabit
  * Copyright (c) 2011-2014 Gergely Nagy <algernon@balabit.hu>
+ * Copyright (c) 2015 bkil.hu
```
ihrwein: Please use your name :)

modules/cef/cef-plugin.c
```diff
@@ -0,0 +1,47 @@
...
+gboolean
+cef_plugin_module_init(GlobalConfig *cfg, CfgArgs *args)
+{
+  plugin_register(cfg, cef_plugins, G_N_ELEMENTS(cef_plugins));
+  return TRUE;
+}
+
+const ModuleInfo module_info =
+{
+  .canonical_name = "cef-plugin",
```
ihrwein: Just a nitpick, but I'd use just `cef` here (as its directory name). Also, I'd rename `cef_plugin_module_init` to `cef_module_init`.

modules/cef/format-cef-extension.c
```diff
@@ -0,0 +1,191 @@
...
+  const gchar *char_ptr = str;
+
+  while (*char_ptr)
+    {
+      gunichar uchar = g_utf8_get_char_validated (char_ptr, -1);
+
+      switch (uchar)
+        {
+        case '=':
+          g_string_append (escaped_string, "\\=");
```
ihrwein: I miss raw strings from C :(

modules/cef/format-cef-extension.c
```diff
@@ -0,0 +1,191 @@
...
+}
+
+static gboolean
+tf_cef_append_value(const gchar *name, const gchar *value,
+                     cef_state_t *state)
+{
+  if (state->need_separator)
+    g_string_append_c(state->buffer, ' ');
+
+  g_string_append_escaped(state->buffer, name);
```
ihrwein: The name part has some useful restrictions:
* names are single words,
* contain only alphanumeric characters.

That means, we don't need to escape them as the values. I'd use them as they are, maybe log something into the internal messages if the current character is invalid. Or we can drop these characters, turning the invalid name into a valid one (But what if the name contains only these characters? ...).

modules/cef/format-cef-extension.c
```diff
@@ -0,0 +1,191 @@
...
+{
+  gboolean need_separator;
+  GString *buffer;
+  const LogTemplateOptions *template_options;
+} cef_state_t;
+
+static inline void
+g_string_append_escaped (GString *escaped_string, const char *str)
+{
+  const gchar *char_ptr = str;
```
bazsi: BTW we shouldn't use functions from the glib namespace, it was a bad idea even in the past, and this is a static function anyway so we are not forced to do that

modules/cef/cef-plugin.c
bkil-syslogng: So should we change these in JSON as well?
modules/json/json-plugin.c

modules/cef/format-cef-extension.c
bkil-syslogng: Okay, I'll remove the confusing double escaping for clarity and move the
key checking closer to the escaper.

bkil-syslogng:
I believe that I have addressed all but one of your issues: I'll still need to update grammar based usage tomorrow. I recommend reviewing the new additions by commits.

bkil-syslogng:
I'm considering that the commit involving refactoring gchar* to GString* should be moved to a separate PR and all corner cases should be covered, and every module and implementation should have an extra test to verify that nothing was broken. This probably requires a few more days, but is not essential to this functionality. Of course leaving this out leaves a gapping bug dangling in CEF and every other code, that's why I've originally included it here. What do you think?

@bazsi Yep, good spotting of append_unsafe_utf8_as_escaped_binary_c & co . Actually the required work seemed to escalate without bounds last night, and I wanted to finish this ASAP. This was just a place holder, and the final implementations involves basically copy&pasting the code and adapting to '\0' termination. Please provide some more feedback on your views. Other than this forgotton placeholder, I'm not sure what extra allocation you refer to. Internally, we have everything already stored as a GString, and at some internal boundary, someone started to pass the wrong pointer.

bkil-syslogng:
I will rebase the fixes for the last two commits:
"value-pairs: major changes to expose GString *value instead of gchar*";
"lib/utf8utils.c: bugfix for invalid UTF-8 input …".

I will only rebase earlier things at the very end. Previous comments attached to these two commits will be available at my repo:
https://github.com/bkil-syslogng/syslog-ng/pull/6

bkil-syslogng:
You can now review today's patches.

* replaced all introduced GString* value occurrences with value&value_len.
* lib/utf8utils.c: deduplicated and optimized strlen from \*\_unsafe\_utf8\_to\_escaped\_\\*() (Cool Code of The Day (TM))

bkil-syslogng:
I will start squashing previous commits and rebasing soon, so if you have any specific input regarding the commits, this is a good time to here about them. The original branch and comments will be backup up to my repo, as last time, and I will link them here.

lbudai:
LGTM, :+1:

bkil-syslogng:
Thank you for the reviews. Rebasing and extensions have been finished for this round. Improvements and fixes will be coming according to a newly opened issue and developed separately. We have agreed that this will be the final form which can be merged. No further development is expected. Q.E.D.

bazsi:
can we please merge this? I would like to get my value-pairs stuff merged, but we agreed that this should go in first.

is there anything that holds this back?

bazsi:
btw, can you pls split your patches a bit more in the next series? I try to rebase my stuff against yours, and the largish patches do not help.

Examples:
  * the patch that adds length argument to utf8utils also has the refactoring step
  * the patch that adds the cef module also has the value-pairs changes, which is independent

small and orthogonal patches help review a lot and rebasing.

Thanks.

bkil-syslogng:
Okay, next time I'll ask feedback about the squashing step.

As a side note, I thought your policy was to reduce the number of commits
to a level which still makes sense and generally to hide progress and only
show features. According to this, showing refactoring steps doesn't always
help. It might in some cases, but when I completely replace this small
function I didn't think it would have been meaningful. Anyway, I think we
should eventually use the one in PE because that seems to be faster. Did
you consider that possibility while doing your optimizations?

On Tue, Jan 12, 2016 at 12:13 PM, Balazs Scheidler <notifications@github.com

PR #843, opened 2015-12-21
travis: turn on -Werror for make

### LINKED ISSUE #776: Eliminate compiler warnings and turn on -Werr

### LINKED ISSUE #804: add --disable-werr option
#776 is about to enable -Werr for syslog-ng. Making compilation warning free is done on Ubuntu. I do packaging on other distros with many different releases. Out of experience these give warnings at very different places. Please add a --disable-werr option, so I can compile packages anywhere. I'm happy to provide you with warning messages if I need to enable this option any time.

ihrwein:
Fixes #804 
Fixes #776

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_script:
   - ./autogen.sh
 script:
   - unset PYTHON_CFLAGS # HACK
-  - ./configure --with-ivykis=internal --prefix=$HOME/install/syslog-ng --enable-pacct --enable-manpages --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl 
+  - ./configure CFLAGS=-Werror --with-ivykis=internal --prefix=$HOME/install/syslog-ng --enable-pacct --enable-manpages --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl 
   - make && make distcheck VERBOSE=1 && make func-test VERBOSE=1
 compiler:
   - gcc
--- a/modules/dbparser/pdb-rule.c
+++ b/modules/dbparser/pdb-rule.c
@@ -65,12 +65,14 @@ pdb_rule_set_context_timeout(PDBRule *self, gint timeout)
 void
 pdb_rule_set_context_scope(PDBRule *self, const gchar *scope, GError **error)
 {
-  self->context_scope = correllation_key_lookup_scope(scope);
-  if (self->context_scope < 0)
+  int context_scope = correllation_key_lookup_scope(scope);
+  if (context_scope < 0)
     {
       self->context_scope = RCS_GLOBAL;
       g_set_error(error, 0, 1, "Unknown context scope: %s", scope);
     }
+  else
+    self->context_scope = context_scope;
 }
 
 void
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -22,8 +22,8 @@
  *
  */
 
-#include "python-dest.h"
 #include "python-module.h"
+#include "python-dest.h"
 #include "python-value-pairs.h"
 #include "python-logmsg.h"
 #include "python-helpers.h"
--- a/modules/python/python-grammar.ym
+++ b/modules/python/python-grammar.ym
@@ -23,9 +23,9 @@
 
 %code requires {
 
+#include "python-main.h"
 #include "python-parser.h"
 #include "python-dest.h"
-#include "python-main.h"
 #include "value-pairs.h"
 #include "vptransform.h"
 
--- a/modules/python/python-module.h
+++ b/modules/python/python-module.h
@@ -24,7 +24,7 @@
 #ifndef PYTHON_MODULE_H_INCLUDED
 #define PYTHON_MODULE_H_INCLUDED 1
 
-#include "syslog-ng.h"
 #include <Python.h>
+#include "syslog-ng.h"
 
 #endif
--- a/modules/python/python-parser.c
+++ b/modules/python/python-parser.c
@@ -22,9 +22,10 @@
  *
  */
 
+#include "python-module.h"
+#include "python-grammar.h"
 #include "python-parser.h"
 #include "cfg-parser.h"
-#include "python-grammar.h"
 
 extern int python_debug;
 int python_parse(CfgLexer *lexer, LogDriver **instance, gpointer arg);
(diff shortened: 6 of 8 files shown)


ihrwein:
@bazsi @juhaszviktor could you review this PR?

juhaszviktor:
:+1:

ihrwein:
Update: I added a `-1` substraction to the `strncat` call.

MrAnno:
:+1:

ihrwein:
@lbudai Could you merge this?

bkil-syslogng:
I would +1, if you #undef'ed the macros in conflict. Leaking out of such definitions from headers does usually indicate a bug in my opinion. You could consider #ifdef/#undef if you think this is platform dependent. However, if this is a bug in one of the libraries, we should also report towards them.

bkil-syslogng:
So instead of all the hacky changes in the last commit, I would simply alter python-module.h something like this:
```
#include "syslog-ng.h"

#ifdef _POSIX_C_SOURCE
#include <Python.h>
#else
#include <Python.h>
#undef _POSIX_C_SOURCE
#endif
```

I guess you would need some combinatorics as well if multiple macros are involved, or somehow save the previous value. Those should be given through the compiler and/or command line, and usually not from intermediate code I think.

bkil-syslogng:
An alternative would be to define the said macros at the very beginning of syslog-ng.h if we or our dependencies really take advantage of them. We should then introduce a policy that every file must include syslog-ng.h before any other header file (I guess this is mostly the case at the moment anyhow):

http://pubs.opengroup.org/onlinepubs/007904975/functions/xsh_chap02_02.html

bkil-syslogng:
TL;DR:
```
if test "x$enable_python" = "xyes" -a "x$python_found" = "xyes"; then
  CFLAGS="${CFLAGS} -D_XOPEN_SOURCE=500 -D_POSIX_C_SOURCE=200112L"
fi
```
I have a still little better workaround. The below patch is only a quickie not my final proposal, but at least it could give you some new ideas:
https://github.com/bkil-syslogng/syslog-ng/commit/e8ea0e25002aca26d03b7f225b154237c7859abb

bazsi:
I think we could make an exception with the Python header here, and include
it before any other header. Clearly it was not entirely designed as a
library header. IIRC this problem is also fixed with newer Python versions.
I hope I am not missing anything.

bkil-syslogng:
@bazsi I think we agree that this is a Python bug. I'm not sure which
versions this plagues. What makes me concerned is that certain versions of
this header can leak this macro outside, while others do not. In this case,
it taints our libraries, modules and all of our users to who this
propagates to. This could introduce unwanted behavior changes in
implementation.

What about the principle of least surprise and documenting intention by
code? If we silently place the `#include <Python.h>` to the top and also
alter all includes to which this reaches, we're not documenting a bug to
watch out for in the future.

How about this compromise? We place something like `#ifdef WITH_PYTON /
#include <Python.h>` at the very beginning of syslog-ng.h and introduce a
policy that `#include "syslog-ng.h"` must come first everywhere and must be
present in all files which `#include` anything.

bkil-syslogng:
By the way, I'm still desperate to get some form of `-Werror` merged ASAP. I've uncovered a few embarrassing bugs/typos yesterday that I will submit later today.

@bazsi @ihrwein  If you don't like any of my proposed workarounds, at least put a comment at each affected include directive that must stay put (the ones altered by your commit and maybe others as well), put a big banner on top of each include header which needs to be included specially, just merge that and forget about it. This is my last input, otherwise :+1:

bazsi:
Python is in a satellite module whose only aim is to bring in Python
support.

This shouldn't plague anything outside modules/python, if it is, we have a
bug.

Iirc we had a similar problem in Zorp where Python plays a much more
central role, and at the end we concluded including Python.h first. It
don't remember problems resulting from this.

We already have a policy to include syslog-ng.h first, but I would never
include python.h from syslog-ng.h, that would pollute our global namespace
a lot.

bkil-syslogng:
Now as I checked, it's not the first occasion when we need to adjust macros:
https://github.com/balabit/syslog-ng/blob/master/modules/afsocket/afinet-dest.c#L39
https://github.com/balabit/syslog-ng/blob/master/modules/afsocket/afsocket-grammar.ym#L59

bazsi:
Yeah, libnet is broken as well.

bazsi:
What do you think of this solution:
https://github.com/balabit/syslog-ng/tree/f/turn-on-Werror-bazsi

bkil-syslogng:
@bazsi If the whole stuff still works correctly, I'm very happy with your proposed patch :+1:

bkil-syslogng:
According to my tests, more warning fixes will be needed, otherwise merging this would break master: https://github.com/bkil-syslogng/syslog-ng/pull/12

ihrwein:
I'm closing this in favor of #871

PR #844, opened 2015-12-21
Fix possible deadlock in patterndb

juhaszviktor:
patterndb: the set emit function is possible to callback to the mainloop.

But if the mainloop currently runs the tick_timer of the patterndb,
it causes deadlock, because the timer wait for the patterndb's lock (held by the _pattern_db_process)
and the emit function waits for the mainloop.

Solution is, that unlock the lock, while the emit functions are run


ihrwein:
:+1:

ihrwein:
@juhaszviktor Could you rebase it, please?

PR #852, opened 2015-12-30
F/java modpath cleanups

bazsi:
While reviewing pr #813 by @lmesz I found some mess related to java_module_path (and module_path for that matter, but that'll come in a separate PR).

The problem is that java_module_path variable was made global whereas it doesn't have to be one, so it's better if we don't make it one.

This branch aims to fix that up.


MrAnno:
:+1:

juhaszviktor:
:+1:

Issue #669, opened 2015-09-03
HTTP destination does not log any server side error


mitzkia:
It would be good if syslog-ng can log out server side errors to the console output.

u2yg:
What is the status of this task?

gregory094:
This task is related to #668 and will be fixed with PR #689. We just simply log the  returned status code of the server and a default message to the code as a reason.

ihrwein:
@u2yg it is still an issue

PR #855, opened 2016-01-04
Log HTTP response error codes in HTTP destination

MrAnno:
Fixes: #669 

Signed-off-by: Laszlo Varady <laszlo.varady@balabit.com>

### DIFF
--- a/modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestination.java
+++ b/modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestination.java
@@ -76,13 +76,22 @@ public class HTTPDestination extends TextLogDestination {
 		logger.error("error in writing message.");
 		return false;
             }
+
             responseCode = connection.getResponseCode();
+            if (isHTTPResponseError(responseCode)) {
+                logger.error("HTTP response code error: " + responseCode);
+                return false;
+            }
         } catch (IOException | SecurityException | IllegalStateException e) {
             logger.debug("error in writing message." +
                     (responseCode != 0 ? "Response code is " + responseCode : ""));
             return false;
         }
         return true;
+    }
+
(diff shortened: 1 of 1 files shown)


bazsi:
This looks good to me, I was just wondering if we could make java logs a bit more integrated to syslog-ng internal logging style.

:+1:

ihrwein:
:+1:

lbudai:
@bazsi: you mean that we could use 
this one:
`HTTP response error; error_code ='404' `
instead of this:
`HTTP response code error: 404`
?

It's doable. @MrAnno ?

bazsi:
I am not sure how java generated log messages appear in general. But yeah,
that's what I meant, and just for this log message but all of them.

This is fine, the formatting should change as a whole in a  unified manner.

MrAnno:
Currently, the InternalMessageSender interface enables to send messages with different priorities but without any tags.

`msg_event_suppress_recursions_and_send(msg_event_create(pri, c_str, NULL));`

bazsi:
I consider this an omission of the original implementation which should be
fixed. I wouldn't try to determine priorities when to do it, as there are
many things to improve and this is just one of them.

But java destinations should be just as integrated with syslog-ng as if
they were written in C.

MrAnno:
Oh, I see. I will create a new issue about this.

lbudai:
I'd merge this as-is, then a new PR should fix the java logmessage formatting issue.

bazsi:
I agree.

-- 
Bazsi

ihrwein:
I've merged it but I forget to ask for a rebase. I hope that the master is OK.

lbudai:
we will see... :-)

bkil-syslogng:
The integration protection that I have outlined earlier would have both caught this, and assured us whether this would be safe.

PR #857, opened 2016-01-05
Suspend source if queue is full

juhaszviktor:


bazsi:
Hmm I have some doubts about this implementation that I would like to
discuss, perhaps IRL.

juhaszviktor:
I've updated the pull request, with the store/restore the current window size, what we've discussed IRL.

lbudai:
:+1:

PR #858, opened 2016-01-06
F/remove fix relative path of syslog ng in func test

bazsi:
originallly @lmesz submitted this branch as PR #813 which I've reworked a bit. This contains a combination of his and my work and thus superseeds #813, which should be closed without merging.

The aim of this code is to make it possible to run the functest on a binary that is outside of the build tree by specifying a SYSLOG_NG_BINARY environment variable. e.g. functest would be able to run on an installed binary.

I have added @lmesz's signoff to the last patch as it was missing, but please confirm that it is ok this way (see the contributing guide what that means):

https://github.com/balabit/syslog-ng/blob/master/CONTRIBUTING.md

please merge.


lmesz:
:+1: 
I think it is okay in this way :)

PR #860, opened 2016-01-08
Add debug

jbfuzier:


jbfuzier:
merge

ihrwein:
@jbfuzier You may want to check the `rr` tool: http://rr-project.org/

It can record the execution of an application, then you can replay that in a debugger session as many times as you want. You will get the same addresses, inputs, program flow every time.

jbfuzier:
Thanks that's a great idea !

--
Envoyé depuis Android

PR #861, opened 2016-01-08
Can turn off inheritance of environment variables in program destinations/sources

MrAnno:
Source/destination program used to inherit the entire environment of the parent process.
This pull request introduces a configuration option which can turn off this behavior.

### DIFF
--- a/modules/afprog/afprog-grammar.ym
+++ b/modules/afprog/afprog-grammar.ym
@@ -54,6 +54,7 @@
 
 %token KW_PROGRAM
 %token KW_KEEP_ALIVE
+%token KW_INHERIT_ENVIRONMENT
 
 %type   <ptr> source_afprogram
 %type   <ptr> source_afprogram_params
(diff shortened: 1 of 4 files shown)


bkil-syslogng:
@MrAnno Just a technical recommendation: if you push your commits more often, you don't need to reference them by hash ID in comments. I think they are sorted by commit time, but grouped by pushes. Of course, I don't say that you should always push unfinished code, because that would make internal rebasing and doodling more difficult. However, if you do reference a commit by hash ID, I guess you thought that the given commit will stay until PR finazilation.

lbudai:
could you rebase it to master?

lbudai:
:+1:

bazsi:
Please talk to @fekete-robert about option naming.

Why abbreviate here?

One idea: inherit-environment(yes/no) is more intuitive to me.

lbudai:
I like the inherit-environment name.

I also suggested something like that but the guys said that there is a BI somewhere which use the abbreviated name (or something similar).

I think we can rename it (and update the BI accordingly).

fekete-robert:
inherit-environment sounds good to me.

MrAnno:
I changed the config option name to inherit-environment.

Issue #863, opened 2016-01-11
Undefined symbol when using keep-alive option in program source/destination


MrAnno:
Compiler: gcc (Ubuntu 5.2.1-22ubuntu2) 5.2.1 20151010

If I set up the `keep-alive(yes)` option in afprog source/destination, syslog-ng fails to start with the following error: 

```
syslog-ng: symbol lookup error: .../lib/syslog-ng/libafprog.so: undefined symbol: afprogram_dd_set_keep_alive
```

`afprogram_dd_set_keep_alive` function is marked as inline, therefore, there is no symbol generated for this in the symbol table.

PR #864, opened 2016-01-11
afprog: fix afprogram_dd_set_keep_alive() linkage

MrAnno:
Fixes: #863 


ihrwein:
:+1:

bkil-syslogng:
:+1: Another reason why we should ban hand made micro optimizations like inline. (And of course we should ban non-static inlines) Unless of course the given developer can prove with benchmark that performance will significantly improve on a supported platform and at the same time not decrease on any supported platform.

dnsjts:
:+1:

PR #876, opened 2016-01-20
DNS Cache "recycling" for new threads

jbfuzier:
Here is my proposal of implementation based on this post (https://github.com/balabit/syslog-ng/compare/master...jbfuzier:master#commitcomment-15348408)

One thing I am not sure is that I am keeping a reference to   DNSCacheEntry cache_first; (line455:458) which is a TLS variable and reusing it in the new thread once the cache is "recycled". 

I do not know if this is a problem or not ?


bazsi:
Does the solution work for you?

lib/dnscache.c
```diff
@@ -89,6 +111,15 @@ static gint dns_cache_expire = 3600;
 static gint dns_cache_expire_failed = 60;
 static gint dns_cache_persistent_count = 0;
 static gchar *dns_cache_hosts = NULL;
+pthread_mutex_t lock;
```
bazsi: We should probably get rid of all those TLS variables in favour of using a single pointer to the current DNSCache instance.

That will probably solve the name clashes (the need for the final underscore in the member names) and
also make the code a lot simpler.

lib/dnscache.c
```diff
@@ -89,6 +111,15 @@ static gint dns_cache_expire = 3600;
...
+
+long 
+gettid(void)
+{
+  pid_t tid;
+  tid = syscall(SYS_gettid);
+  return (long) tid;
+}
```
bazsi: AFAIR ivykis exports such a function, so we shouldn't define our own here.

lib/dnscache.c
```diff
@@ -357,26 +405,67 @@ void
...
+      cache_first.prev = NULL;
+      cache_last.prev = &cache_first;
+      cache_last.next = NULL;
+      cache_hosts_mtime = -1;
+      cache_hosts_checktime = 0;
+      persist_first.next = &persist_last;
+      persist_first.prev = NULL;
+      persist_last.prev = &persist_first;
+      persist_last.next = NULL;
+
```
bazsi: by removing the individual TLS variables this would be a lot simpler.

lib/dnscache.c
```diff
@@ -394,4 +483,4 @@ dns_cache_global_deinit(void)
   if (dns_cache_hosts)
     g_free(dns_cache_hosts);
   dns_cache_hosts = NULL;
-}
```
bazsi: here we should free the cached DNSCache instances.

bazsi:
I propose this one to be closed in favour of #947 

@jbfuzier did this solution resolve the issue you were originally facing?
@juhaszviktor I remember you also found this issue independently, do you think #947 would be a solution for you?

jbfuzier:
@bazsi So far I have not managed to compile. Having trouble with AX_PREFIX_CONFIG_H not prefixing as it should :(

PR #879, opened 2016-01-21
Remove autoconf archive dependency

### LINKED ISSUE #878: Remove autoconf-archive dependency

ihrwein:
Fixes #878 

Note: the license of this file is GPL3.


bkil-syslogng:
I think if we handled the file as (part of) a 3rd party external library, mixing `GPL v2` and `GPL v3` would not occur. I.e., I would search for `autoconf-archive` in our bootstrapping stage. If present, use the system provided one. Otherwise use a local mirror which, if you insist to include it inside this same repo, should be well separated, like inside a `/3rd`-party folder. Actually, pushing the relevant file(s) from `autoconf-archive` to a new Git repo and attaching it as a `git submodule` also sounds doable.

lbudai:
please, do not merge this : I'll solve my packaging issue in another way.

PR #880, opened 2016-01-21
Geoip quiet

### LINKED ISSUE #663: geoip: spurious and wrongly routed messages
When the geoip parser fails to match an entry in the database, _e.g._ `127.0.0.1`, it complains with `Invalid database type GeoIP City Edition, Rev 1, expected GeoIP Country Edition`.
- this message is spurious
- and should flow through the internal message queue

ihrwein:
Fixes #663 

Note, that this PR sets the minimal supported geoip version to 1.6.3.

You can find fresh geoip debian packages here: https://launchpad.net/~maxmind/+archive/ubuntu/ppa

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 language: c
 install:
   - sudo apt-get update -qq
-  - sudo apt-get install -qq curl
+  - sudo apt-get install -qq curl python-software-properties
+  - sudo add-apt-repository -y ppa:maxmind/ppa
   - curl http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_12.04/Release.key | sudo apt-key add -
   - echo "deb http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_12.04 ./" | sudo tee --append /etc/apt/sources.list.d/syslog-ng-obs.list
   - sudo apt-get update -qq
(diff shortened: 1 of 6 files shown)


bkil-syslogng:
Please prefix using single underscore instead of double underscores in `__open()`. It was recently mentioned in `F/logmsg serialization` . Otherwise :+1:

bazsi:
why do we need to make this configurable?

ihrwein:
Maybe a system administrator wants to see those messages. Ofc, we can enable silencing by default.

bazsi:
where would they become visible? if they are printed to stderr/stdout those
are usually redirected to /dev/null when syslog-ng is running.


-- 
Bazsi

ihrwein:
I talked with @bazsi IRL and we come up with the following proposal:
* if syslog-ng's stderr isn't a TTY, we use geoip in silenced mode, otherwise we don't make it quiet

faxm0dem:
But they *do* show up in my log path!

ihrwein:
I got confused :) If they are spurious I guess we can drop these messages. Or did you test the branch and it didn't work?

faxm0dem:
I didn't test any branch.
My remark was answering @bazsi's comment:

| where would they become visible? if they are printed to stderr/stdout those are usually redirected to /dev/null when syslog-ng is running

I confirm that in the 3.7.2 release, the geoip messages do appear in the log path, and not on stdio

ihrwein:
I've checked how GEOIP_SILENCE is used: https://github.com/maxmind/geoip-api-c/blob/4f487bf95532e0bba7783d591faff178ab0aa462/libGeoIP/GeoIP.c#L1467

Actually, the `printf` calls are untouched.

Based on this fast I think this PR can be closed...

PR #628, opened 2015-08-10
Makefile: install tools under syslog-ng/tools

ihrwein:
Previously they were installed under tools without the
prefixing syslog-ng directory.

Fixes #587 

Signed-off-by: Tibor Benke <tibor.benke@balabit.com>

Issue #587, opened 2015-07-24
Question: $PREFIX/share/tools/


faxm0dem:
I've noticed that the `make install` procedure installs files in `/usr/local/share/tools`, namely the following:

```
lex-rules.am
cfg-grammar.y
merge-grammar.pl
system-expand
```

While `system-expand` has been useful to me in the past, it's not clear to me why the other 3 should be installed on the target.

Moreover, it would probably make sense to give some context to the full path of those files, to make them easily identifiable as belonging to syslog-ng.

bazsi:
merge-grammar.pl is needed by out-of-tree modules. it is only installed
into syslog-ng indepent location if you install from source, otherwise it
can be customized with configure options --sharedstatedir or something like
that, which should be done by distributions.

-- 
Bazsi

faxm0dem:
Indeed. Don't you think it should be installed into *e.g.* `$PREFIX/share/syslog-ng/tools/` from source?

algernon:
Heh. I just now realized it is $PREFIX/**share/tools**! I agree, using the package name in the path (by default) would make much more sense.

ihrwein:
I created a fix, but bazsi convinced me, that this should be done by package maintainers. @czanik, what do you think?

algernon:
I disagree with Bazsi. Package maintainers can - and already do - work this around. But putting syslog-ng specific files into a directory with a very common name sounds wrong. (Directories commonly used by other software for a specific purpose, eg `/usr/bin` are a different case)

Also, what is the downside of putting it elsewhere? It remains to be findable via pkg-config, which was the proper way to locate these files to begin with.

ihrwein:
If I install syslog-ng with `make install` and I set `--prefix` during configure, I get syslog-ng installed under, say, `$HOME/install/syslog-ng`. IMHO there is no additional value if an additional syslog-ng subfolder is created.

algernon:
Yes, but if you install outside of home, which is a more common case, as end-users rarely install to their home, there is much additional value.

If I install into `$HOME/local`, so I can have `$HOME/local/bin` on PATH, there is still additional value.

bazsi:
First of all, my opinion originates from when we had
sysconfdir='${sysconfdir}/syslog-ng' in our configure script. That was
removed at one point and now you have to explicitly tell the configure
script where your configuration is. And this is what happens from our
debian/rules file. (and probably the rpm spec file as well).

[...]

PR #882, opened 2016-01-22
Clean up configure directory

bazsi:
following up PR #628 that tries to address issue #587.

this PR solves the issue and also cleans up related configure.ac/Makefile.am machinery a bit.

Please note @lbudai, @czanik that once this gets integrated, the --datadir configure option should be removed from debian/rules and rpm specfiles, as it will generate paths like:

/usr/share/syslog-ng/syslog-ng/scl & /usr/share/syslog-ng/syslog-ng/tools

I am not sure where and how we should open tickets about the packaging files to avoid getting lost.

### DIFF
--- a/Mk/Makefile.am
+++ b/Mk/Makefile.am
@@ -3,7 +3,6 @@ EXTRA_DIST	+= Mk/lex-rules.am \
 		Mk/populate-makefiles.sh \
 		Mk/subdir.mk
 
-toolsdir	= $(datadir)/tools
 tools_DATA	= \
 		Mk/lex-rules.am \
 		lib/cfg-grammar.y
--- a/configure.ac
+++ b/configure.ac
@@ -1439,15 +1439,11 @@ AM_CONDITIONAL(ENABLE_JAVA, [test "$enable_java" = "yes"])
 AM_CONDITIONAL(ENABLE_MANPAGES, [test "$enable_manpages" != "no"])
 AM_CONDITIONAL(ENABLE_NATIVE, [test "$enable_native" != "no"])
 
-# substitution into manual pages
-expanded_sysconfdir=[`patheval $sysconfdir | sed -e 's/-/\\\\-/g'`]
-
 AC_SUBST(timezonedir)
 AC_SUBST(pidfiledir)
 AC_SUBST(moduledir)
 AC_SUBST(toolsdir)
 AC_SUBST(scldir)
-AC_SUBST(expanded_sysconfdir)
 AC_SUBST(systemdsystemunitdir)
 AC_SUBST(SYSLOGNG_LINK)
 AC_SUBST(SYSLOGNG_DEPS_LIBS)
--- a/configure.ac
+++ b/configure.ac
@@ -73,8 +73,8 @@ if test "x$exec_prefix" = "xNONE"; then
 fi
 pidfiledir='${localstatedir}'
 moduledir='${exec_prefix}/lib/syslog-ng'
-toolsdir='${datadir}/tools'
-scldir='${datadir}/include/scl'
+toolsdir='${datadir}/syslog-ng/tools'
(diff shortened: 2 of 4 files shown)


lbudai:
The best would be to create a dedicated github project for the OBS packaging files (both rpm and deb) and in that case, issues could be opened.

bazsi:
why don't we merge those into syslog-ng? at least those that we maintain.


-- 
Bazsi

ihrwein:
I'd also prefer a dedicated project for packaging, because syslog-ng's code can be untouched while working on packaking.

czanik:
There are too many variations in packaging, depending on the distribution, which release we use, if we enable only features supported by the base distribution or we also provide missing dependencies in separate package, etc. So, even if we put packaging into git, it should be separate from the syslog-ng package.

bazsi:
I think we are on agreement on the code in this PR. @algernon, @ihrwein both showed support in #587 

The only discussion is about whether to merge packaging or not. That should be handled separately.

So I am merging this now.

czanik:
@bazsi @ihrwein The "xsd" directory is now messed up. If I keep the old configure parameter ( --datadir="%{_datadir}/syslog-ng" ) then all files are installed under /usr/share/syslog-ng/syslog-ng except for the "xsd" directory.
If I remove the trailing "syslog-ng" from it, the most of the files are placed under /usr/share/syslog-ng/ but the xsd directory is placed under /usr/share/xsd
``
abuild@linux-fsdl:/home/abuild/rpmbuild/BUILDROOT/syslog-ng-3.8.0alpha0-1.x86_64/usr/share/syslog-ng> ls *
syslog-ng:
include  tools

xsd:
patterndb-1.xsd  patterndb-2.xsd  patterndb-3.xsd  patterndb-4.xsd
abuild@linux-fsdl:/home/abuild/rpmbuild/BUILDROOT/syslog-ng-3.8.0alpha0-1.x86_64/usr/share/syslog-ng> 
``

bazsi:
This should be fixed up by defining xsddir just like the others.

I am trying to fix that up, thanks for opening the ticket.

czanik:
As discussed with @ihrwein the following problem seems to be related too:

``
linux-fsdl:/etc/syslog-ng # syslog-ng -Fvd
[2016-02-02T15:56:59.233862] Unable to detect fully qualified hostname for localhost, use_fqdn() will use the short hostname;
[2016-02-02T15:56:59.235337] Systemd is detected as the running init system;
[2016-02-02T15:56:59.236420] Starting to read include file; filename='/etc/syslog-ng/scl.conf', depth='1'
[2016-02-02T15:56:59.237012] Global value changed; define='scl-root', value='/usr/share/include/scl'
[2016-02-02T15:56:59.237452] Global value changed; define='include-path', value='/etc/syslog-ng:/usr/share/include'
[2016-02-02T15:56:59.237871] Finishing include; filename='/etc/syslog-ng/scl.conf', depth='1'
[2016-02-02T15:56:59.238301] Reading path for candidate modules; path='/usr/lib64/syslog-ng'
[2016-02-02T15:56:59.238731] Reading shared object for a candidate module; path='/usr/lib64/syslog-ng', fname='afsocket.so', module='afsocket'
[...]
[2016-02-02T15:56:59.261423] Reading shared object for a candidate module; path='/usr/lib64/syslog-ng', fname='graphite.so', module='graphite'
[2016-02-02T15:56:59.261522] Registering candidate plugin; module='graphite', context='template-func', name='graphite_output', preference='0'
Error parsing source, source plugin system not found in /etc/syslog-ng/syslog-ng.conf at line 32, column 2:

	system();
	^^^^^^

syslog-ng documentation: http://www.balabit.com/support/documentation/?product=syslog-ng
mailing list: https://lists.balabit.hu/mailman/listinfo/syslog-
``

PR #884, opened 2016-01-22
Enable -Wshadow compiler flag

bazsi:
this branch resolves a couple of -Wshadow warnings and enables them in travis.


bazsi:
Assuming the travis check succeeds, this one should be ready for merging.

ihrwein:
Travis had some problems this morning, I restarted the builds.

ihrwein:
Could you rebase it to master? Thanks.

lbudai:
There are some issues around `logmsg_serializer`.
@bazsi: if you don't mind, we will fix it on your branch and merge it ASAP.

bazsi:
I don't mind. My branch should be ready, I am just fighting with travis.


-- 
Bazsi

lbudai:
hmmm.. after it passed the -Wshadow, we got a segfault:

`(lt-syslog-ng:9537): GLib-CRITICAL **: g_strsplit: assertion `string != NULL' failed
Segmentation fault (core dumped)`

ihrwein:
The last problem was easily reproducible by executing `syslog-ng -V`. The hard part was to dig this information out (actually, the segfault happened by an `import` statement in python...).

lbudai:
@bazsi: I think with my last patch it will be pass on Travis.
But before we merge, I'd ask you to check it.
( I removed `plugin_global_deinit`, and I'm not sure that I'd call it as `plugin_global_init`. These global inits and deinits a little bit weird: global to who? When syslog-ng started as a logger process it is global, but not when it is started to print some version info/other info about itself. I can live with it, just wanted to point out the issue. )

lbudai:
A rebase is still required (when the changes are accepted, I can do that).

bazsi:
Hi,

The term "global" does not refer to the fact that it is initializing the
global variables (although most often than not this is what they do), but
rather the fact that they initialize the subsystem at startup to perform
action whatever needed to get initialized.

The term "global" was added as a qualification to clearly indicate that an
init function is not related to the object in the given module and also in
a number of cases to avoid clashes. log_msg_init() was at one point such a
global init function (at least as far as I remember this was a concrete
case, but I am sure there were others). log_msg_init() is clearly
associated with the rest of the log_msg_ API, and as can be expected now we
_DO_ have log_msg_ini() that expects a LogMessage *self argument.

value-pairs for example fills a couple of its hashtables at that point.

With all that being said, I agree that we do have an issue in how we
initialize our state, apphook is both a leaf-level API (when it provides
ways to register hooks at various times during operation) and at the same
time it _does_ initialization where it depends on basically everything.
They should probably be split. It also comes back that we need to decrease
coupling (primarily between GlobalConfig and the rest of the code, and then
MainLoop and the rest of the code). Those should set most stuff in order,
but it's not a single iteration away.

The naming of these functions is the least of the problem but I am open to
suggestions.

Also, the proper solution to this problem would be to

1) either call plugin_global_init() just like mainloop_global_init() was
(which probably could be delayed back into apphook, but I am not sure)

2) create a central location for all paths that contain relocatable
references and resolve them there and call that function early, while
getting rid of plugin_global_init() and maybe even mainloop_global_init()
as this is the only thing they do.

The 2nd solution I like much more.

-- 
Bazsi

ihrwein:
I reverted my commit as it's no longer needed. (Three of us touched this branch, so I preferred reverting over force pushing).

lbudai:
For this particular problem (resolve relocatable references in paths) the 2nd solution is what I was also thinking about.

lbudai:
I'll fix the conflict (xsddir fix, PR #911), but after that, please someone merge this :-)

@juhaszviktor : I know that we agreed on that I'll merge diskbuffer _before_ this PR, but
as we have to do some further testing with that, I'll suggest that start with this. 
(I'll fix the possible shadow issues on your branch).

@bazsi: could you look at to the last patch before we merge?

lbudai:
Those variables are actually global variables and we use their addresses in some places... and I didn't want to hide them.

juhaszviktor:
Hmm, yes you are right.
:+1:

bazsi:
the spelling of the variable is mixedCase and not under_scory.

-- 
Bazsi

bkil-syslogng:
SYSLOGDEV-1551 would have caught this issue

lbudai:
@bkil-syslogng : I know what SYSLOGDEV-1551 is (an internal jira ticket which is about using a tool for checking/enforcing some coding-guide rules), but this is not understandable for everyone.

PR #885, opened 2016-01-23
Improve the performance of utf8-utils by using an optimized implementation of strchr()

bazsi:
…for strchr()

This patch adds an optimized version of strchr() that is often used by
$(format-json) to output escaped strings, which improves $(format-json)
performance by about 3 percent.

### DIFF
--- a/lib/str-utils.h
+++ b/lib/str-utils.h
@@ -26,6 +26,7 @@
 #define STR_UTILS_H_INCLUDED 1
 
 #include "syslog-ng.h"
+#include <string.h>
 
 /* functions that should be implemented by GLib but they aren't */
 GString *g_string_assign_len(GString *s, const gchar *val, gint len);
@@ -70,5 +71,35 @@ void g_string_steal(GString *s);
   } while (0)
 
 gchar *__normalize_key(const gchar* buffer);
+
+
+/* This version of strchr() is optimized for cases where the string we are
+ * looking up characters in is often zero or one character in length.  In
+ * those cases we can avoid the strchr() call, which can make a tight loop
+ * doing a lot of strchr() calls a lot faster, especially as this function
+ * is inlined.
+ *
+ * Naming: I originally wanted to use the original function as a prefix
+ * (strchr), however that would potentially pollute the namespace of the
+ * library that we are optimizing.  So I've added an underscore prefix in
+ * order not to clash, with that it is still obvious that it intends to
+ * behave the same as strchr().
+ *
+ * NOTE: don't use this unless strchr() really shows up in your profile.
+ */
+static inline char *
+_strchr_optimized_for_single_char_haystack(const char *str, int c)
+{
+  if (str[0] == c)
+    return (char *) str;
+  if (str[1] == '\0')
+    {
+      if (c != '\0')
+        return NULL;
+      else
+        return (char *) &str[1];
+    }
+  return strchr(str + 1, c);
+}
 
 #endif
--- a/lib/tests/test_str-utils.c
+++ b/lib/tests/test_str-utils.c
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2016 Balabit
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+#include "testutils.h"
+#include "str-utils.h"
+
+/* this macro defines which strchr() implementation we are testing. It is
+ * extracted as a macro in order to make it easy to switch strchr()
+ * implementation, provided there would be more.
+ */
+
+#define strchr_under_test _strchr_optimized_for_single_char_haystack
+
+static void
+assert_strchr_is_null(const gchar *str, int c)
+{
+  assert_null(strchr_under_test(str, c), "expected a NULL return");
+}
+
+static void
+assert_strchr_finds_character_at(const gchar *str, int c, int ofs)
+{
+  char *result = strchr_under_test(str, c);
+
+  assert_not_null(result, "expected a non-NULL return");
+  assert_true(result - str <= strlen(str), "Expected the strchr() return value to point into the input string or the terminating NUL, it points past the NUL");
+  assert_true(result >= str, "Expected the strchr() return value to point into the input string or the terminating NUL, it points before the start of the string");
+  assert_gint((result - str), ofs, "Expected the strchr() return value to point right to the specified offset");
+}
+
+int
+main(int argc G_GNUC_UNUSED, char *argv[] G_GNUC_UNUSED)
+{
+  assert_strchr_is_null("", 'x');
+  assert_strchr_is_null("a", 'x');
+  assert_strchr_is_null("abc", 'x');
+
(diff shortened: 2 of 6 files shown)


lib/tests/test_str-utils.c
```diff
@@ -0,0 +1,38 @@
+#include "testutils.h"
+#include "str-utils.h"
+
+#define strchr _strchr_optimized_for_single_char_haystack
+
+static void
+assert_strchr_is_null(const gchar *str, int c)
+{
+  assert_null(strchr(str, c), "expected a NULL return");
```
juhaszviktor: Why don't use define instead of _strchr_optimized_for_single_char_haystack?

lib/utf8utils.c
```diff
@@ -33,7 +33,7 @@ _is_character_unsafe(gunichar uchar, const gchar *unsafe_chars)
   if (!unsafe_chars)
     return FALSE;
 
-  return strchr(unsafe_chars, (gchar) uchar) != NULL;
+  return _strchr_optimized_for_single_char_haystack(unsafe_chars, (gchar) uchar) != NULL;
```
juhaszviktor: If the _strchr_optimized_for_single_char_haystack is used only from utff8utils.c, why is it in different header file.
To be honest I don't really think, that this function will be used from another source

juhaszviktor: We discussed header/non-header choice with @bkil-syslogng, and out opinion is that this function (_strchr_optimized_for_single_char_haystack) should be in this c file, near to its usage.
Or if you found more opportunities of usage, please update the PR with these, but in case you should rename this function e.g.: strchr_optimized_for_single_char_haystack (without underscore)

bazsi:
with the discussion IRL, I've updated the PR, we agreed with @juhaszviktor and @lbudai to stick to the underscore prefix, but I am happy to discuss it further.

lib/tests/test_str-utils.c
```diff
@@ -0,0 +1,65 @@
...
+  assert_not_null(result, "expected a non-NULL return");
+  assert_true(result - str <= strlen(str), "Expected the strchr() return value to point into the input string");
+  assert_gint((result - str), ofs, "Expected the strchr() return value to point right to the specified offset");
+}
+
+int
+main(int argc G_GNUC_UNUSED, char *argv[] G_GNUC_UNUSED)
+{
+  assert_strchr_is_null("", 'x');
+  assert_strchr_is_null("a", 'x');
```
bkil-syslogng: According to white box testing, you should check all combinations for string lengths of both 0 and >=1. E.g.:
```
assert_strchr_finds_character_at("a", 'a', 0);
```

lib/tests/test_str-utils.c
```diff
@@ -0,0 +1,65 @@
...
+
+  assert_not_null(result, "expected a non-NULL return");
+  assert_true(result - str <= strlen(str), "Expected the strchr() return value to point into the input string");
+  assert_gint((result - str), ofs, "Expected the strchr() return value to point right to the specified offset");
+}
+
+int
+main(int argc G_GNUC_UNUSED, char *argv[] G_GNUC_UNUSED)
+{
+  assert_strchr_is_null("", 'x');
```
bkil-syslogng:
```
assert_strchr_is_null("", '\0');
```

lib/str-utils.h
```diff
@@ -71,4 +72,29 @@ void g_string_steal(GString *s);
...
+ *
+ * Naming: I originally wanted to use the original function as a prefix
+ * (strchr), however that would potentially pollute the namespace of the
+ * library that we are optimizing.  So I've added an underscore prefix in
+ * order not to clash, with that it is still obvious that it intends to
+ * behave the same as strchr().
+ *
+ * NOTE: don't use this unless strchr() really shows up in your profile.
+ */
+static inline char *
```
bkil-syslogng: Why not return `const char*`?

lib/tests/test_str-utils.c
```diff
@@ -0,0 +1,65 @@
...
+  assert_null(strchr_under_test(str, c), "expected a NULL return");
+}
+
+static void
+assert_strchr_finds_character_at(const gchar *str, int c, int ofs)
+{
+  char *result = strchr_under_test(str, c);
+
+  assert_not_null(result, "expected a non-NULL return");
+  assert_true(result - str <= strlen(str), "Expected the strchr() return value to point into the input string");
```
bkil-syslogng:
```
 assert_true(result - str <= strlen(str), "Expected the strchr() return value to point not after the end of the input string");
 assert_true(result >= str, "Expected the strchr() return value to point not before the beginning of the input string");
```
Or something like that.

lib/str-utils.h
bazsi: Because I wanted it to match the strchr() prototype.

lib/tests/test_str-utils.c
bkil-syslogng: Please don't forget to fix this because this tests a bug.

bkil-syslogng:
If it still improves performance after the new fix, you should also consider whether the following call sites could benefit from the same optimizations:
```
modules/basicfuncs/str-funcs.c:tf_sanitize_call()
modules/csvparser/csv-scanner.c
modules/dbparser/radix.c: ... strchr(param, ...)
modules/syslogformat/syslog-format.c
```

bazsi:
-- 
Bazsi

On Wed, Jan 27, 2016 at 2:05 PM, Tamas Nagy <notifications@github.com>
wrote:

> If it still improves performance after the new fix, you should also
> consider whether the following call sites could benefit from the same
> optimizations:
>
> modules/basicfuncs/str-funcs.c:tf_sanitize_call()
>
>
well, this might apply, however in this case there's no default and we
don't know if the user often uses single character escapes. I don't think
we should include it here, until there's a specific use-case that would
need to be sped up.


>
> modules/csvparser/csv-scanner.c
>
>
By default this contains multiple characters in quotes, so the optimization
doesn't really apply.


>
> modules/dbparser/radix.c: ... strchr(param, ...)
>
>
this is the pattern reloading code, not really performance sensitive.


>
> modules/syslogformat/syslog-format.c
>
>
I have added usage of the optimized version here, however this could be
optimized even more:

```
static gint
log_msg_parse_skip_chars_until(LogMessage *self, const guchar **data, gint
*length, const gchar *delims)
{
  const guchar *src = *data;
  gint left = *length;
  gint num_skipped = 0;

  while (left && _strchr_optimized_for_single_char_haystack(delims, *src)
== 0)
    {
      src++;
      left--;
      num_skipped++;
    }
  *data = src;
  *length = left;
  return num_skipped;
}
```

This function can basically be replaced with a single strcspn() call,
however this is only used from an AIX specific branch. ("Message forwarded
for XXXX".)

bkil-syslogng:
>
> > modules/csvparser/csv-scanner.c
> >
> >
> >By default this contains multiple characters in quotes, so the
> optimization
> doesn't really apply.
> >
>
Both scl/apache/apache.conf and scl/solaris/plugin.conf set delimiters to a
single space, same for test_csvparser_perf.c. Same for the admin guide. I
guess you referred to the quotes part, but see the other call site.


> >
> > modules/dbparser/radix.c: ... strchr(param, ...)
> >
> >
> > this is the pattern reloading code, not really performance sensitive.
>
> This is an advantage of optimized reusable primitives: if the
preconditions hold (mostly 0 or 1 long set), you don't need to do detailed
assessment at each call site. But I leave this one up to you..

> modules/syslogformat/syslog-format.c
> >
> >
> I have added usage of the optimized version here, however this could be
> optimized even more:
>
> ```
> static gint
> log_msg_parse_skip_chars_until(LogMessage *self, const guchar **data, gint
> *length, const gchar *delims)
> {
> const guchar *src = *data;
> gint left = *length;
> gint num_skipped = 0;
>
> while (left && _strchr_optimized_for_single_char_haystack(delims, *src)
> == 0)
> {
> src++;
> left--;
> num_skipped++;
> }
> *data = src;
> *length = left;
> return num_skipped;
> }
> ```
>
> This function can basically be replaced with a single strcspn() call,
> however this is only used from an AIX specific branch. ("Message forwarded
> for XXXX".)
>
> Yes, optimization is never finished - it's only good enough for a given
purpose at hand.

lib/tests/test_str-utils.c
bkil-syslogng: The last test case that I feel to be still missing is:
```
assert_strchr_finds_character_at("a", 'a', 0);
```

Otherwise, :+1: from me

I see you have adjusted the other call sites as well.

bazsi:
I have squashed the last missing testcase into the unit tests. If travis passes, I'll merge it.

Let me know if you have an issue with that. We had numerous review iterations already, so this one should be good enough.

lbudai:
I agree :-)

PR #886, opened 2016-01-25
Fix journald program name flapping

### LINKED ISSUE #831: flapping ${PROGRAM} using systemd
I noticed an odd flapping behaviour on one of our servers:

* using `logger -t test` the name of the program changes from one invocation to the other
* `journalctl` always reports the right value (`PROGRAM=test`)
* `messages` (syslog-ng) randomly flaps between `test` and `logger`

```
☠ rpm -q syslog-ng
syslog-ng-3.7.2-2.el7.centos.x86_64
☠ sh /tmp/system-expand
## system() expands to:

channel {
    source {
systemd-journal();
    }; # source
channel {
  channel {
    parser {
      json-parser(prefix('.cim.') marker('@cim:'));
    }};
    flags(final);
  };
  channel { };
};
}; # channel
;
☠ tail -n0 -F /var/log/messages&
☠ logger -t test test
☠ Dec 10 15:38:11 node01 logger[29418]: test
logger -t test test
☠ Dec 10 15:38:14 node01 logger[29420]: test
logger -t test test
☠ Dec 10 15:38:16 node01 test[29421]: test
logger -t test test
☠ Dec 10 15:38:17 node01 logger[29422]: test

☠ journalctl | tail -n4
Dec 10 15:38:11 node01 test[29418]: test
Dec 10 15:38:14 node01 test[29420]: test
Dec 10 15:38:16 node01 test[29421]: test
Dec 10 15:38:17 node01 test[29422]: test
```

ihrwein:
Fixes #831 

The following changes are included in this PR:
* insert all the read key-value pairs under `.journald.`
* if we can map something to a syslog-ng macro, do it as we do it now (like `_HOSTNAME`)
* prefer `SYSLOG_IDENTIFIER` over `_COMM` for setting `PROGRAM`
* make these changes enabled only after `3.8`

After the changes I did some refactoring.

### DIFF
--- a/modules/systemd-journal/journal-reader.c
+++ b/modules/systemd-journal/journal-reader.c
@@ -629,7 +629,7 @@ journal_reader_options_init(JournalReaderOptions *options, GlobalConfig *cfg, co
   if (options->prefix == NULL && !cfg_is_config_version_older(cfg, 0x0308))
     {
       gchar *value = ".journald.";
-      msg_warning("Default value changed for the prefix() option of systemd-journal source in " VERSION_3_8,
+      msg_warning("WARNING: Default value changed for the prefix() option of systemd-journal source in " VERSION_3_8,
                   evt_tag_str("old_value", ""),
                   evt_tag_str("new_value", value),
                   NULL);
--- a/modules/systemd-journal/journal-reader.c
+++ b/modules/systemd-journal/journal-reader.c
@@ -239,16 +239,16 @@ _set_program(JournalReaderOptions *options, LogMessage *msg)
 static void
 _set_message_timestamp(JournalReader *self, LogMessage *msg)
 {
-   guint64 ts;
-   journald_get_realtime_usec(self->journal, &ts);
-   msg->timestamps[LM_TS_STAMP].tv_sec = ts / 1000000;
-   msg->timestamps[LM_TS_STAMP].tv_usec = ts % 1000000;
-   msg->timestamps[LM_TS_STAMP].zone_offset = time_zone_info_get_offset(self->options->recv_time_zone_info, msg->timestamps[LM_TS_STAMP].tv_sec);
-   if (msg->timestamps[LM_TS_STAMP].zone_offset == -1)
-     {
-       msg->timestamps[LM_TS_STAMP].zone_offset = get_local_timezone_ofs(msg->timestamps[LM_TS_STAMP].tv_sec);
-     }
+  guint64 ts;
 
+  journald_get_realtime_usec(self->journal, &ts);
+  msg->timestamps[LM_TS_STAMP].tv_sec = ts / 1000000;
+  msg->timestamps[LM_TS_STAMP].tv_usec = ts % 1000000;
+  msg->timestamps[LM_TS_STAMP].zone_offset = time_zone_info_get_offset(self->options->recv_time_zone_info, msg->timestamps[LM_TS_STAMP].tv_sec);
+  if (msg->timestamps[LM_TS_STAMP].zone_offset == -1)
+    {
+      msg->timestamps[LM_TS_STAMP].zone_offset = get_local_timezone_ofs(msg->timestamps[LM_TS_STAMP].tv_sec);
+    }
 }
 
 static gboolean
--- a/modules/systemd-journal/journal-reader.c
+++ b/modules/systemd-journal/journal-reader.c
@@ -195,18 +195,16 @@ __handle_data(gchar *key, gchar *value, gpointer user_data)
     {
       msg->pri = (msg->pri & ~7) | atoi(value);
     }
+
+  if (!options->prefix)
+    {
+      log_msg_set_value_by_name(msg, key, value, value_len);
+    }
   else
     {
-      if (!options->prefix)
-        {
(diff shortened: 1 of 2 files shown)


modules/systemd-journal/journal-reader.c
```diff
@@ -149,14 +149,8 @@ __reader_wakeup(LogSource *s)
 }
 
 static void
-__handle_data(gchar *key, gchar *value, gpointer user_data)
+__map_key_value_pairs_to_syslog_macros(LogMessage *msg, gchar *key, gchar *value, gssize value_len)
 {
```
bazsi: single underscores is what we agreed upon.

modules/systemd-journal/journal-reader.c
```diff
@@ -193,18 +174,50 @@ __handle_data(gchar *key, gchar *value, gpointer user_data)
...
+  __map_key_value_pairs_to_syslog_macros(msg, key, value, value_len);
+  __fill_message(options, msg, key, value, value_len);
+}
+
+static void
+__set_program(LogMessage *msg)
+{
+  gssize value_length = 0;
+  const gchar *value = log_msg_get_value_by_name(msg, "SYSLOG_IDENTIFIER", &value_length);
+
```
bazsi: this doesn't take prefix into account...

modules/systemd-journal/journal-reader.c
```diff
@@ -597,6 +611,11 @@ journal_reader_options_init(JournalReaderOptions *options, GlobalConfig *cfg, co
   if (options->recv_time_zone_info == NULL)
     options->recv_time_zone_info = time_zone_info_new(options->recv_time_zone);
 
+  if (options->prefix == NULL && cfg_is_config_version_older(cfg, 0x0308))
+    {
+      options->prefix = g_strdup(".journald.");
```
bazsi: please issue an upgrade warning here.

modules/systemd-journal/journal-reader.c
ihrwein: prefix is added by _fill_message().

I think the code is cleaner without the nvhandle-s. I would optimize it further when somebody reports back that it is much slower.

bazsi: but you are getting the value from the message at this point, so if it was
set as "${.prefix.SYSLOG_IDENTIFIER}", you won't get it here.




-- 
Bazsi

ihrwein: Now I see! You are right.

ihrwein:
@juhaszviktor Could you check the failing testcase, please? (I debugged it but found nothing.)

modules/systemd-journal/journal-reader.c
```diff
@@ -193,18 +174,66 @@ __handle_data(gchar *key, gchar *value, gpointer user_data)
...
-          g_free(prefixed_key);
-        }
+      gchar *prefixed_key = g_strdup_printf("%s%s", options->prefix, key);
+      log_msg_set_value_by_name(msg, prefixed_key, value, value_len);
+      g_free(prefixed_key);
+    }
+}
+
+static void
+__handle_data(gchar *key, gchar *value, gpointer user_data)
```
juhaszviktor: Please use single underscore

modules/systemd-journal/journal-reader.c
```diff
@@ -234,6 +263,7 @@ __handle_message(JournalReader *self)
 
   journald_foreach_data(self->journal, __handle_data, args);
   __set_message_timestamp(self, msg);
```
juhaszviktor: I know, that the double underscores don't come from your pencil, but if you modify the file, please correct the underscores

ihrwein:
Based on @juhaszviktor's suggestion I fixed the unit test. The branch is rebased, ready for review.

bazsi:
This looks good to me, there are still some double underscores in the tests, and I see trivial optimizations in case this would be slow, but I think this can go in as it is now.

ihrwein:
@juhaszviktor just a quick glance? :)

bazsi:
I have submitted a pull request against your branch yesterday that contains a couple of fixes yet.

bazsi:
This one: https://github.com/ihrwein/syslog-ng/pull/2

ihrwein:
Thanks. I've merged Bazsi's changes into mine. Should I fixup the commits ?

juhaszviktor:
:+1:

ihrwein:
rebased to master, force pushed

dnsjts:
:+1:

czanik:
Could you backport this fix to 3.7? It would resolve https://bugzilla.suse.com/show_bug.cgi?id=987453

phemmer:
I second the backport to 3.7. At the very least, the `SYSLOG_IDENTIFIER` shouldn't be stripped out. Otherwise I could do something like `rewrite r { set("${.journald.SYSLOG_IDENTIFIER}", value("PROGRAM")) }`. But `${.journald.SYSLOG_IDENTIFIER}` doesn't exist (other journald fields are added, just not this one).

maxenced:
+1 for backport to 3.7 :)

faxm0dem:
this is our workaround for the time being: use `logger -n localhost` which will use `network()` source  instead `system()`

PR #895, opened 2016-02-02
travis.yml: add --keep-going to make

dnsjts:
This switch is added in order to show as many compilation error
as possible, therefore the output of a failed Travis CI build might be
more valuable.

[Review on Reviewable](https://reviewable.io/reviews/balabit/syslog-ng/895)


ihrwein:
:+1:

PR #898, opened 2016-02-03
Fix create-dirs() inheritance in file destinations

MrAnno:
When the global `create-dirs` option was set to `yes`, the local one was ignored.

### DIFF
--- a/modules/affile/affile-common.h
+++ b/modules/affile/affile-common.h
@@ -28,10 +28,10 @@
 
 typedef struct _FileOpenOptions
 {
-  gboolean create_dirs:1,
-           needs_privileges:1,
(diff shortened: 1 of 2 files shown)


dnsjts:
:+1:

bkil-syslogng:
This started my thought process about the following. From the grammar, I could see that you can specify yes/no values either via tokens or via a number. Why don't we check whether the specified value is 0 or 1? (Or the hidden options of -1 and beyond) By the way, we don't we support the tokens true/false, enable/disable as well?

bkil-syslogng:
My above question still stands to be answered, but otherwise, this fix is :+1:

PR #901, opened 2016-02-04
Measure unit and functional test coverage

ihrwein:
This PR implements unit and functional test coverage measurement.

gcc builds measure unit test coverage while clang builds mesure unit and functional test coverage.

Some notes:
- we need a Trusty container as Precise's old lcov/gcov is buggy
- the coveralls-lcov gem uploads the coverage data to coveralls.io (I enabled the service for this repo)
- we need to execute make check again

I sent this PR because I think it's somewhat useful. I'd extract the lcov commands into a shell scipt (like collect-cov.sh) just need to find a good place for it. What if I put all travis related files into a .travis directory?

### DIFF
--- a/.lcovrc
+++ b/.lcovrc
@@ -1,0 +1,4 @@
+geninfo_adjust_src_path = /.libs
+geninfo_compat_libtool = 1
+geninfo_auto_base = 1
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ install:
   - curl http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_12.04/Release.key | sudo apt-key add -
   - echo "deb http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_12.04 ./" | sudo tee --append /etc/apt/sources.list.d/syslog-ng-obs.list
   - sudo apt-get update -qq
-  - sudo apt-get install -qq pkg-config flex bison xsltproc docbook-xsl libevtlog-dev libnet1-dev libglib2.0-dev libdbi0-dev libssl-dev libjson0-dev libwrap0-dev libpcre3-dev libcap-dev libesmtp-dev libgeoip-dev libhiredis-dev sqlite3 libdbd-sqlite3 libriemann-client-dev openjdk-7-jdk gradle-2.2.1 autoconf-archive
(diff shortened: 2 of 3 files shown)


bkil-syslogng:
1. Why is Travis failing?
2. Do we really need the double `make check` invocation?
3. Do you think if it would make sense to create a `make coverage` target of some sort so that our end users could check this locally?

ihrwein:
1. at least one functional test is failing
2. make distcheck cleans up after itself, so we don't see the coverage data. It also builds in a temporary directory
3. actually, we have something like that in collect-cov. we may move the extended coverage.info generation into that shell script and use collect-cov only in travis.

bkil-syslogng:
On Thu, Feb 4, 2016 at 2:24 PM, Tibor Benke <notifications@github.com>
wrote:

>
>    1. at least one functional test is failing
>    2. make distcheck cleans up after itself, so we don't see the coverage
>    data. It also builds in a temporary directory
>
> Could you test whether running `make check` in advance speeds up a `make
distcheck` afterwards? (I guess it should in the ccache variants at least)

What if we only ran it on the faster thread?
 - if [ "$CC" = "gcc" ]; then make distcheck V=1; else make check &&
BACKUP_COVS... && make func-test V=1; fi

By the way, our logs are starting to fill up. What if we started doing
something like this?

make distcheck || make distcheck V=1

Although long term, it might make more sense to kind of filter our running
build logs and only show largers portions of the collected portion in case
of errors (sudo apt-get install syslog-ng?)

ihrwein:
Gcov bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=46266
geninfo bug: https://github.com/linux-test-project/lcov/commit/b87e40e475c560bdc88206df4de6dc8cf094d91f
Lcov bug: https://bugzilla.redhat.com/show_bug.cgi?id=706040

lbudai:
@ihrwein: any update on this?

ihrwein:
@lbudai Nothing new. AFAIK we still need a trusty based image to collect coverage. You can close this PR, we can find it in the history.

bkil-syslogng:
@ihrwein We could either consider adding a new matrix element similar to how it was done for cmake, or we could piggy-back it onto the end of the quick-running cmake one. Or we could always compile dependencies from source.

kira-syslogng:

presidento:
@ihrwein OK, we close it now.

PR #905, opened 2016-02-05
Fix create-dirs() inheritance in file destinations

MrAnno:
This is a backport PR, the original pull request can be found here: #898
> When the global `create-dirs` option was set to `yes`, the local one was ignored.


ihrwein:
:+1:

PR #907, opened 2016-02-05
configure.ac: link against sasl2 lib only when it is available

lbudai:
In mongo-c-driver sasl support is set to auto by default, but
we forced to link against the sasl2 lib.
As mongo-c-driver is able to work without sasl support, syslog-ng
should also be able to work.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -950,7 +950,7 @@ if test "x$with_mongoc" = "xinternal" || test "x$with_mongoc" = "xauto-internal"
 		LIBMONGO_LIBS="\
 			-L\$(top_builddir)/modules/afmongodb/mongo-c-driver \
 			-L\$(top_builddir)/modules/afmongodb/mongo-c-driver/src/libbson \
-			-lmongoc-1.0 -lbson -lsasl2 $OPENSSL_LIBS"
+			-lmongoc-1.0 -lbson $OPENSSL_LIBS"
 		LIBMONGO_CFLAGS="\
 			-I\$(top_srcdir)/modules/afmongodb/mongo-c-driver/src/mongoc \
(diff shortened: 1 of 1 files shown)


ihrwein:
:+1:

bkil-syslogng:
Probably okay, but it would be nice if you could test whether it builds cleanly with and without the lib present on the computer. Just in case the configure/autogen scripts are not 100%.

lbudai:
@bkil-syslogng : I've already done it _before_ sent the PR
(and I'd suggest the same to you when you add a new lib ;-) )

bkil-syslogng:
@lbudai I also did check before I sent my PR. However, because one of our/their scripts picked up my sasl install, it ended up producing linking errors without this.

Actually this PR is also just a workaround. The correct solution would be for `mongo-c-driver` to export all the respective linking flags that they need for their own use. I've now checked in `src/libmongoc-1.0.pc` after configure, and all flags seems to be populated there: `-lsasl2 -lssl -lcrypto   -lrt -L${libdir} -lmongoc-1.0` Maybe it was not correct in the original, pre 1.x branches I've started with, I can't remember.

Could you perhaps suck in all those flags via `configure.ac`?

lbudai:
@bkil-syslogng : sasl2 is not always required by mongo-c-driver, only when you configure it with sasl2 support, or, when it is available on your system and you configure mongo-c-driver with default options; the problem was that you added it as a mandatory dependency to the syslog-ng `configure.ac`.

bkil-syslogng:
@lbudai It doesn't matter. Please address the second paragraph of my comment. The `*.pc` file with either include `-lsasl2` or not, depending on what it finds.

lbudai:
@bkil-syslogng : if you can create a solution, just open a PR then move on ;-)

dnsjts:
:+1:

PR #909, opened 2016-02-08
F/diskq

juhaszviktor:
Disk-buffer: this kind of disk buffer is ported from syslog-ng PE.
    
    This is an inner-destination plugin, make it possible to use a disk based queue.
    There are 2 kind of disk-buffer: reliable and non-reliable.
    
    Non-reliable means, that it uses memory buffers,
    therefore this kind of disk-buffer is not crash safe,
    but in some cases it can be faster than the reliable one
    
    The reliable disk-buffer store every message on the queue in serialized format,
    therefore it is crash-safe, but can be slower than the non-reliable disk buffer.

usage examples:
```
destination d_network1("server"
           disk-buffer(
        	reliable(no)
                qout-size(64)
                disk-buf-size(1000000)
                mem-buf-size(10000)
           )
);

destination d_network2("server2"
           disk-buffer(
                  reliable(yes)
                  disk-buf-size(1000000)
                  mem-buf_length(100000)
           )
);
....
```

### DIFF
--- a/modules/diskq/diskq-options.c
+++ b/modules/diskq/diskq-options.c
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) 2002-2016 Balabit
+ * Copyright (c) 2016 Viktor Juhasz <viktor.juhasz@balabit.com>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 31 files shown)


faxm0dem:
\o/

ihrwein:
@fekete-robert you have to see this :)

fekete-robert:
:dancers:

lbudai:
We need to run some further E2E tests before we merge it.

modules/diskq/diskq-options.c
```diff
@@ -0,0 +1,131 @@
...
+{
+  self->mem_buf_length = mem_buf_length;
+}
+
+void
+disk_queue_options_check_plugin_settings(DiskQueueOptions *self)
+{
+  if (self->reliable)
+    {
+      if (self->mem_buf_length > 0)
```
ihrwein: Shouldn't this be `self->mem_buf_length < 0` (and the other one in this function)?

mitzkia:
The E2E tests are passed for me.

modules/diskq/diskq-options.c
juhaszviktor: No. When we use reliable disk buffer, the mem_buf_length will be ignored, and the mem_buf_size will be used, and vica versa

lbudai:
So, we have a diskbuffer... :-)

Issue #908, opened 2016-02-08
datadir problems in git head due to pr#882


czanik:
@bazsi As I noted at https://github.com/balabit/syslog-ng/pull/882 there were multiple problems introduced with this PR. Right now 3.8 can't be tested, as syslog-ng installs scl files to one location and looks for them at another. Please either fix or revert the pr.

presidento:
@czanik is this issue still valid?

czanik:
It was fixed by https://github.com/balabit/syslog-ng/pull/911
Closing.

PR #911, opened 2016-02-09
make sure all path/dir variables are defined in a single location, by configure.ac

bazsi:
This one should be a partial fix to #908 as it doesn't yet fix xsddir, but it does for scldir and config_includedir

NOTE: YOU CAN GET SEGFAULTS after merging this in a development environment: I've found a segfault in the lexer if two instances of scl/system/plugin.conf is included into the same config. I'm still diagnosing that issue, but I wanted to get this in.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -74,7 +74,9 @@ fi
 pidfiledir='${localstatedir}'
 moduledir='${exec_prefix}/lib/syslog-ng'
 toolsdir='${datadir}/syslog-ng/tools'
-scldir='${datadir}/syslog-ng/include/scl'
+
+config_includedir='${datadir}/syslog-ng/include'
+scldir="${config_includedir}/scl"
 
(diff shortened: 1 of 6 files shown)


bazsi:
So the segfault happens when the plugin.conf is found two times, once in the old and once in the new location both included due to the wildcard import.

This should not happen in production but might if you install syslog-ng from source without uninstalling the old scl files first.

bazsi:
I have now added support for xsddir as pointed out by a comment in PR #882, so this PR should solve all of the problems caused by PR #882.

As there's an actual breakage in current master, merging this should be a priority.

Please review and merge.

@czanik pls try if this solves your problem.

czanik:
@bazsi Thanks, just tested and it seems to have fixed my problems. I have now other problems, but most likely unrelated.

PR #913, opened 2016-02-10
Distribute generated Java related header files

ihrwein:
We distribute the generated grammar files in the dist tarball. However, we also generate Java related header files during the build process, but we don't distribute them.

This PR does the following changes:
* adds four generated header files to the dist tarball
* updates the Makefiles to reflect this change

There is an extra commit here. I changed `allprojects` to `subprojects` because our root project doesn't have code (just subprojects) so the two options are identical, but the latter form (IMHO) is more expressive.

http://stackoverflow.com/questions/12077083/what-is-the-difference-between-allprojects-and-subprojects


lbudai:
Hi,

IMHO until we compiling all java related things with gradle, we should keep that
self-contained. What you are doing here is make a ping-pong game between gradle and make.
Gradle is able to build everything it needs to the java modules.

BUT still there is a problem we want to solve, so I'm happy that you sent this PR ;-)

This area is actually a borderland between java and c: I'd say that if we touch this, then
we should do the followings:
 * move the C-related things under the authority of make (the java binding)
 * use gradle only to build the java-modules

With this step we would be able to
 * release java binding in OBS without hacks
 * release syslog-ng with java support without having gradle(>=2.2)

Future of java plugins is to publish them in a public maven repo.

At the end of this work we can clean up the deb build scripts and integrate it with syslog-ng.

I'd say let's start working on that.

@czanik: I think this is something make you happy :-)

czanik:
+1 (+6x10^23 :-) )

Peter Czanik (CzP) <peter.czanik@balabit.com>
Balabit / syslog-ng upstream
http://czanik.blogs.balabit.com/
https://twitter.com/PCzanik

bazsi:
Hi,

I'd love to keep the java modules within the syslog-ng source tree if at
all possible. why do we want to separate those out? I think the
implementation language of a module doesn't matter to the user and the less
obvious it is that a given module uses a specific programming language the
better.



-- 
Bazsi

czanik:
Right now you can't compile the C part of the Java destination, if you don't have Gradle and the Jar files. What it means in practice, that Java support can't be enabled in distributions:
- Gradle is not included or too old
- even if some of the required Jar files are available in distributions, it's never the version required by syslog-ng
This cut enables the compilation of the Java destination and base modules even in this environment, thus making available the Java destination even in base distro packages. The actual HDFS, Kafka, etc. drivers could be downloaded from a maven repo.

bazsi:
I think we are in agreement. I guess the maven repo can use the binaries
built from the syslog-ng core, can't it?

Bazsi


-- 
Bazsi

lbudai:
Hi,

the source tree is in safe :-) 

This is just about how we are build the java modules and the native language binding.

Currently we are building the native code with gradle : it is working and I think it is fine, but
for package maintainers it is a real s*ck - what we want to do is to make it possible to build java binding without having gradle. After that it is possible to build a  syslog-ng which is ready for work with java modules. Where the java-modules are come from? Well... this is the point when central maven repo come into play. If we release our java modules into a publicly available central then the users can download our modules from that repo and they can use it (even it they don't have gradle in their OS package repo).

Central maven repo:
It is possible to upload artifacts into a publicly available central maven repo (we need to check the details, but as we are an open source project, I think we can do it).
We need to provide a pom file, set the licensing, etc...; AFAIK JNI won't be a problem, as ZMQ also provides their jni lib.

ihrwein:
What do you think about building these files with ant? That should be present on old systems.

lbudai:
Don't really like the idea...
Why would you like to use ant? Just because old systems?  It would be overkill for the language binding (and I'm sure that someone would ask for rewriting the whole java building stuff in ant
and I'm also sure in that we don't want to resolve the dependecies by hand, and don't want to maintain those ugly xml files...)

ihrwein:
Apart from the generated header files, we have to build the class files and syslog-ng-core.jar.

It's possible to define the proper dependency graph in make to build the class files, but it's error prone and hard to maintain. Also, the startup time of  `javac` and `javah` is very long. If we use make we would call these commands many times (there are 10 java files).  I'd like to try ant, because an xml file may be better than a huge manually written makefile.

lbudai:
@ihrwein : it is able to track local dependencies, but not 3rd party deps (no maven repo, maybe there is an extension out there which makes possible to use maven repo in ant). The makefile shouldn't be a huge one, as we are talking about the binding itself, not the modules.
Talk about this IRL.

ihrwein:
@lbudai ok, the makefile is almost ready (build is ok, distcheck fails).

What should be do with java-modules? Generate a maven project and distribute it with the dist tarball?

lbudai:
@ihrwein : As gradle is not required anymore for enabling java support (that will only depends on jdk), we  may should create a `--{enable|disable}-java-modules` option which would depend on gradle (and on java support). And when java modules are enabled, dist tarball should contain the sources.

ihrwein:
I implemented the proposed changes.

Closing this in favor of https://github.com/balabit/syslog-ng/pull/928

PR #918, opened 2016-02-10
Check copyright headers during the execution of make check

bkil-syslogng:
* tests/copyright/policy:
  * Ignore our git submodules
* tests/copyright/check.sh:
  * Reduce default verbosity
* Makefile.am,.travis.yml:
  * Invoke check-copyright on `make check`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ before_script:
   - ./autogen.sh
   - unset PYTHON_CFLAGS # HACK
   - ./configure CFLAGS=-Werror --with-ivykis=internal --prefix=$HOME/install/syslog-ng --enable-pacct --enable-manpages --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl 
-  - if [ "$CC" = "clang" ]; then export COPYRIGHTVERBOSITY=1; make check-copyright; fi
   - make -j || make V=1 --keep-going # to make error messages more readable on error
 script:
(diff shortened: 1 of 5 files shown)


bazsi:
Looks good to me

ihrwein:
:+1:

bkil-syslogng:
@lbudai requested this feature.

@bazsi @ihrwein Note that I'm not completely convinced that this is how it should be done when planning from a higher level perspective. TL;DR:

* `make check`: unit & simple end to end/acceptance test cases
* `make dev-check` (check-quick?): as above, plus contributing safeguards (`-Werror`, static analysis, copyright, VCS, style, simple mem leaks, basic coverage safeguards, ...)
* `make ci` (check-full?): all above, plus performance, fuzz, stability, fault injection, ...

My thought process in detail:
I'd need to review a few project and consider a few use cases, but here is what I figured. After a user compiles & installs a package, she would like to make certain that the software functions as expected on her machine. This is why it's a good idea to distribute `check` tests and run them before usage.

So from this perspective, a user would be generally interested in unit tests and end-to-end functional tests (not our func-tests).

A developer is generally interested in every aspect of operation, however some take longer than others. This is why it is a good idea to separate long running and/or controlled environment tests that include performance, stability, fuzz testing and fault injection along with general unit and functional. The shorter test set would generally include contribution preconditions, like style checks, copyrights, regression, git diversion, simple memory leaks and others I have outlined in a separate issue.

lbudai:
@bkil-syslogng : thanks!

PR #925, opened 2016-02-12
Fix possible deadlock in patterndb

MrAnno:
This is a backport PR, the original pull request can be found here: #844

>patterndb: the set emit function is possible to callback to the mainloop.

>But if the mainloop currently runs the tick_timer of the patterndb,
it causes deadlock, because the timer wait for the patterndb's lock (held by the _pattern_db_process)
and the emit function waits for the mainloop.

>Solution is, that unlock the lock, while the emit functions are run


bkil-syslogng:
:+1:

PR #926, opened 2016-02-12
Fix program driver keep-alive() option linkage

MrAnno:
This is a backport PR, the original pull request can be found here: #864
> Fixes: #863


bkil-syslogng:
:+1:

PR #897, opened 2016-02-03
fix java version check ('test' has '=', not '==' as operator)

schweikert:
configure --enable-java doesn't work for me (syslog-ng 3.7.2):

    checking for JAVA_VERSION... ./configure: 16173: test: 1: unexpected operator

PR #927, opened 2016-02-12
Fix java version check ('test' has '=', not '==' as operator)

MrAnno:
This is a backport PR, the original pull request can be found here: #897

>configure --enable-java doesn't work for me (syslog-ng 3.7.2):

>    `checking for JAVA_VERSION... ./configure: 16173: test: 1: unexpected operator`


bkil-syslogng:
:+1:

lbudai:
@MrAnno : thanks for all the backports!!

PR #928, opened 2016-02-12
Build Java language bindings without Gradle

ihrwein:
This PR removes the dependency between the Java language bindings and Gradle.

The `--enable-java` option means, that the Java language bindings are compiled and `syslog-ng-core.jar` is built.

The new `--enable-java-modules` option enables to build the in-tree Java modules and `syslog-ng-common.jar`. You cannot build the Java modules if `--disable-java` is used.

The following steps are involved for building `syslog-ng-core.jar`:
1. generate class files
2. generate header files based on class files
3. add class files to syslog-ng-core.jar

The class files are built with javac. Some input source files has class files as dependencies (e.g. LogPipe.java), some are leaf files (e.g. InternalMessageSender.java). I checked the dependencies between the files are created the corresponding *_CLASS_DEPS variables. There is no need to
define the dependencies (undefined variables in make expand to the empty string), if a source file is a leaf (without any dependency).

The .SECONDEXPANSION is needed to use these dependency lists based on the stem (the matching part). So if make builds LogPipe.class, the $* expands to `LogPipe`. The $ symbol has to be prefixed with an other $ in second expansions.

The .gradle directory don't have to be removed, as Gradle won't use this directory.

It's worth to check the manual pages of javac and javah.


ihrwein:
If we use the planned gitflow model, we can:
* if we merge to the develop branch we can create a snapshot version of `syslog-ng-core.jar` and upload them to bintray
* if we merge into the master branch, we create a "normal" version of `syslog-ng-core.jar` and upload them to bintray

lbudai:
Thanks! 

:+1: 

( I liked that gradle script... :) )

When Travis is finished, I'll merge this.

@czanik: we can start to modify the build scripts

faxm0dem:
Do I understand correctly that we don't need gradle any more at all?

lbudai:
We need gradle, but only for building java modules.

faxm0dem:
Ack

ihrwein:
rebased, force pushed, propably fixed

ihrwein:
I've asked @czanik to test this branch before merging.

Here are the steps to use this branch:
```
git remote add ihrwein https://github.com/ihrwein/syslog-ng.git
git fetch --all
git checkout build-java-module-without-gradle
```

ihrwein:
I've checked this PR with the DummyTextDestination and it worked :) @czanik also built this branch, I hope he shares his thoughts with us, so we can merge this ASAP.

czanik:
+1 also from me

bkil-syslogng:
I would be happier if you tried it with the other modules as well, but anyway :+1: .

PR #936, opened 2016-02-17
Fix diskq linking

ihrwein:
We experienced a warning during diskq building and `make -j` wasn't stable also. This PR intends to fix these two issues.

```
Warning: Linking the executable modules/diskq/dqtool against the loadable module
libdisk-buffer.so is not portable!
```

### DIFF
--- a/modules/diskq/Makefile.am
+++ b/modules/diskq/Makefile.am
@@ -32,7 +32,7 @@ modules_diskq_libdisk_buffer_la_LIBADD = $(MODULE_DEPS_LIBS)
 modules_diskq_libdisk_buffer_la_LDFLAGS = $(MODULE_LDFLAGS)
 
 modules_diskq_dqtool_SOURCES = modules/diskq/dqtool.c
-modules_diskq_dqtool_LDADD = $(top_builddir)/modules/diskq/libdisk-buffer.la $(top_builddir)/lib/libsyslog-ng.la @BASE_LIBS@ @OPENSSL_LIBS@ @GLIB_LIBS@
+modules_diskq_dqtool_LDADD = $(top_builddir)/modules/diskq/libdisk-buffer.la $(TOOL_DEPS_LIBS) $(top_builddir)/lib/libsyslog-ng.la
 
 modules/diskq modules/diskq/ mod-diskq: modules/diskq/libdisk-buffer.la \
   modules/diskq/dqtool
--- a/modules/diskq/Makefile.am
+++ b/modules/diskq/Makefile.am
@@ -30,6 +30,7 @@ EXTRA_DIST += $(BUILT_SOURCES) modules/diskq/diskq-grammar.ym
 modules_diskq_libdisk_buffer_la_CFLAGS = $(AM_CPPFLAGS) -I$(top_srcdir)/modules/diskq -I$(top_builddir)/modules/diskq -I$(top_srcdir)/lib -I../../lib
 modules_diskq_libdisk_buffer_la_LIBADD = $(MODULE_DEPS_LIBS)
 modules_diskq_libdisk_buffer_la_LDFLAGS = $(MODULE_LDFLAGS)
+modules_diskq_libdisk_buffer_la_DEPENDENCIES= $(MODULE_DEPS_LIBS)
 
 modules_diskq_dqtool_SOURCES = modules/diskq/dqtool.c
 modules_diskq_dqtool_LDADD = $(top_builddir)/modules/diskq/libdisk-buffer.la $(TOOL_DEPS_LIBS) $(top_builddir)/lib/libsyslog-ng.la
--- a/modules/diskq/Makefile.am
+++ b/modules/diskq/Makefile.am
@@ -1,16 +1,10 @@
 bin_PROGRAMS += modules/diskq/dqtool
-
+noinst_LTLIBRARIES += modules/diskq/libsyslog-ng-disk-buffer.la
 module_LTLIBRARIES += modules/diskq/libdisk-buffer.la
 
-modules_diskq_libdisk_buffer_la_SOURCES = \
-  modules/diskq/diskq.c \
-  modules/diskq/diskq.h \
-  modules/diskq/diskq-grammar.y \
+modules_diskq_libsyslog_ng_disk_buffer_la_SOURCES = \
   modules/diskq/diskq-options.h \
   modules/diskq/diskq-options.c \
-  modules/diskq/diskq-parser.c \
-  modules/diskq/diskq-parser.h \
-  modules/diskq/diskq-plugin.c \
   modules/diskq/logqueue-disk.c \
   modules/diskq/logqueue-disk.h \
   modules/diskq/logqueue-disk-non-reliable.c \
@@ -20,6 +14,22 @@ modules_diskq_libdisk_buffer_la_SOURCES = \
   modules/diskq/qdisk.h \
   modules/diskq/qdisk.c
 
+modules_diskq_libsyslog_ng_disk_buffer_la_CPPFLAGS = \
+  $(AM_CPPFLAGS) \
+  -I$(top_srcdir)/modules/diskq
+modules_diskq_libsyslog_ng_disk_buffer_la_LIBADD	=	\
+  $(MODULE_DEPS_LIBS)
+modules_diskq_libsyslog_ng_disk_buffer_la_DEPENDENCIES	=	\
+  $(MODULE_DEPS_LIBS)
+
+modules_diskq_libdisk_buffer_la_SOURCES = \
+  modules/diskq/diskq.c \
+  modules/diskq/diskq.h \
+  modules/diskq/diskq-grammar.y \
+  modules/diskq/diskq-parser.c \
+  modules/diskq/diskq-parser.h \
+  modules/diskq/diskq-plugin.c
+
 BUILT_SOURCES += \
   modules/diskq/diskq-grammar.y \
   modules/diskq/diskq-grammar.c \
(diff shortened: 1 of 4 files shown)


bkil-syslogng:
@ihrwein I've reviewed the changeset, and most seems well. However, leaving out `bin_PROGRAMS` seems suspicious to me. Could you have a look?

bkil-syslogng:
:-1: (I will update this thumb after my next review round, don't take this as a full reject.)

ihrwein:
@bkil-syslogng thanks for the quick review. I think we should split the current diskq modul into two parts:
* the first is a library which provides disk buffering functionality,
* the second is the actual syslog-ng module which embeds this library

dqtool can use the library. This is the model followed by pdbtool and dbparser.

juhaszviktor:
@ihrwein I'm not sure, that this is necessary. I think using -dlpreopen should be enough

bazsi:
IMHO :I wouldn't use -dlpreopen in standalone tools used production. That
trick is good enough for test programs (not nice there either) but not for
prod.


-- 
Bazsi

ihrwein:
The last commit does the splitting, though we still compile some files two times.

modules/diskq/Makefile.am
```diff
@@ -22,15 +14,43 @@ modules_diskq_libdisk_buffer_la_SOURCES = \
...
+  modules/diskq/diskq-plugin.c \
+  $(modules_diskq_libsyslog_ng_disk_buffer_la_SOURCES)
+
+BUILT_SOURCES += \
+  modules/diskq/diskq-grammar.y \
+  modules/diskq/diskq-grammar.c \
+  modules/diskq/diskq-grammar.h
+
+EXTRA_DIST += $(BUILT_SOURCES) modules/diskq/diskq-grammar.ym
+
```
bkil-syslogng: By the way, I haven't seen $(BUILT_SOURCES) on any such right hand side before, so it's also probably an issue to deal with.

`git grep BUILT_SOURCES`

modules/diskq/Makefile.am
```diff
@@ -22,15 +14,43 @@ modules_diskq_libdisk_buffer_la_SOURCES = \
...
+modules_diskq_libdisk_buffer_la_CPPFLAGS = $(AM_CPPFLAGS) -I$(top_srcdir)/modules/diskq -I$(top_builddir)/modules/diskq
+modules_diskq_libdisk_buffer_la_LIBADD = $(MODULE_DEPS_LIBS) $(top_builddir)/modules/diskq/libsyslog-ng-disk-buffer.la
 modules_diskq_libdisk_buffer_la_LDFLAGS = $(MODULE_LDFLAGS)
+modules_diskq_libdisk_buffer_la_DEPENDENCIES = $(MODULE_DEPS_LIBS) $(top_builddir)/modules/diskq/libsyslog-ng-disk-buffer.la
 
 modules_diskq_dqtool_SOURCES = modules/diskq/dqtool.c
-modules_diskq_dqtool_LDADD = $(top_builddir)/modules/diskq/libdisk-buffer.la $(top_builddir)/lib/libsyslog-ng.la @BASE_LIBS@ @OPENSSL_LIBS@ @GLIB_LIBS@
+modules_diskq_dqtool_LDADD = \
+  $(TOOL_DEPS_LIBS) \
+  $(top_builddir)/lib/libsyslog-ng.la \
```
bkil-syslogng: same as ${MODULE_DEPS_LIBS}

modules/diskq/tests/Makefile.am
```diff
@@ -1,6 +1,6 @@
 DISKQ_TEST_CFLAGS = -I$(top_srcdir)/lib -I$(top_srcdir)/libtest -I$(top_srcdir)/modules/diskq @CFLAGS_NOWARN_POINTER_SIGN@
-DISKQ_TEST_LDFLAGS = -dlpreopen $(top_builddir)/modules/syslogformat/libsyslogformat.la -dlpreopen $(top_builddir)/modules/diskq/libdisk-buffer.la
-DISKQ_TEST_LDADD = $(top_builddir)/libtest/libsyslog-ng-test.a $(top_builddir)/modules/diskq/libdisk-buffer.la $(top_builddir)/lib/libsyslog-ng.la @TOOL_DEPS_LIBS@ @OPENSSL_LIBS@
+DISKQ_TEST_LDFLAGS = -dlpreopen $(top_builddir)/modules/syslogformat/libsyslogformat.la
+DISKQ_TEST_LDADD = $(top_builddir)/libtest/libsyslog-ng-test.a $(top_builddir)/modules/diskq/libsyslog-ng-disk-buffer.la $(top_builddir)/lib/libsyslog-ng.la @TOOL_DEPS_LIBS@ @OPENSSL_LIBS@
```
bkil-syslogng: You could use that here again `${MODULE_DEPS_LIBS}` for `DISKQ_TEST_LDADD`
Also consider `DISKQ_TEST_LDFLAGS = ${PREOPEN_SYSLOGFORMAT}`

modules/diskq/Makefile.am
```diff
@@ -22,15 +14,43 @@ modules_diskq_libdisk_buffer_la_SOURCES = \
...
+
+BUILT_SOURCES += \
+  modules/diskq/diskq-grammar.y \
+  modules/diskq/diskq-grammar.c \
+  modules/diskq/diskq-grammar.h
+
+EXTRA_DIST += modules/diskq/diskq-grammar.ym
+
+modules_diskq_libdisk_buffer_la_CPPFLAGS = $(AM_CPPFLAGS) -I$(top_srcdir)/modules/diskq -I$(top_builddir)/modules/diskq
+modules_diskq_libdisk_buffer_la_LIBADD = $(MODULE_DEPS_LIBS) $(top_builddir)/modules/diskq/libsyslog-ng-disk-buffer.la
```
bkil-syslogng: You could consider lifting this into a new variable: `$(top_builddir)/modules/diskq/libsyslog-ng-disk-buffer.la`

bkil-syslogng:
Waiting for a second reviewer: @juhaszviktor, @bazsi (@dnsjts)? Though, of course, anyone is welcome.

modules/diskq/Makefile.am
```diff
@@ -1,18 +1,10 @@
-BUILT_SOURCES += modules/diskq/diskq-grammar.y modules/diskq/diskq-grammar.c modules/diskq/diskq-grammar.h
-EXTRA_DIST += $(BUILT_SOURCES) modules/diskq/diskq-grammar.ym
-
-module_LTLIBRARIES += modules/diskq/libdisk-buffer.la
 bin_PROGRAMS += modules/diskq/dqtool
+lib_LTLIBRARIES += modules/diskq/libsyslog-ng-disk-buffer.la
+module_LTLIBRARIES += modules/diskq/libdisk-buffer.la
```
bazsi: Do we really want to install this library? I think it is enough to create a static lib and link that to all users.

We don't want to deal with lib versioning or anything of that sort IMHO.

modules/diskq/tests/Makefile.am
```diff
@@ -1,6 +1,6 @@
 DISKQ_TEST_CFLAGS = -I$(top_srcdir)/lib -I$(top_srcdir)/libtest -I$(top_srcdir)/modules/diskq @CFLAGS_NOWARN_POINTER_SIGN@
-DISKQ_TEST_LDFLAGS = -dlpreopen $(top_builddir)/modules/syslogformat/libsyslogformat.la -dlpreopen $(top_builddir)/modules/diskq/libdisk-buffer.la
-DISKQ_TEST_LDADD = $(top_builddir)/libtest/libsyslog-ng-test.a $(top_builddir)/modules/diskq/libdisk-buffer.la $(top_builddir)/lib/libsyslog-ng.la @TOOL_DEPS_LIBS@ @OPENSSL_LIBS@
+DISKQ_TEST_LDFLAGS = ${PREOPEN_SYSLOGFORMAT}
+DISKQ_TEST_LDADD = $(top_builddir)/libtest/libsyslog-ng-test.a $(LIBSYSLOG_NG_DISK_BUFFER) $(MODULE_DEPS_LIBS) @TOOL_DEPS_LIBS@ @OPENSSL_LIBS@
 
 modules_diskq_tests_TESTS = \
```
bazsi: Would be nice to get rid off this syslogformat linking into unrelated unit tests.

Another user instead of just using a function to create LogMessage instances without parsing a string.

bazsi: But the last comment is really independent of the main body of the patch.

czanik:
The "make -j" problem seems to be resolved with this patch, but unversioned libraries are not supposed to be put directly in /usr/lib (or /usr/lib64). This is a complaint from the SUSE post build checks:

[  169s] syslog-ng.x86_64: E: devel-file-in-non-devel-package (Badness: 50) /usr/lib64/libsyslog-ng-disk-buffer.so

bkil-syslogng:
@czanik Thank you for the report. I like the referenced tool. Maybe we should connect it to our continuous integration pipeline as well. In general, does it produce any false positives?

czanik:
There are some false positives, but most warnings and errors reported by it are valid. The problem is, that to reach the post build checks the spec file often needs to be modified (like in this case, adding the new library to the files section). So it's not perfect for automation.

modules/diskq/Makefile.am
ihrwein: Then the static lib won.

ihrwein:
Laci changed the lib type to static. Please review the last two commits, after that I'll rebase the branch.

lbudai:
@ihrwein : could you do the rebase?

ihrwein:
rebased, auto squashed, some newly introduced tabs are converted to spaces

PR #941, opened 2016-02-19
F/remove msg error etc null sentinel

bkil-syslogng:
* `msg_error("description", evt_tag_str("when", "before"), NULL);`
* `msg_error("description", evt_tag_str("when", "after"));`

Contains a `sed` moster and some `vararg` magic.

### DIFF
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -340,7 +340,7 @@ afinter_sd_init(LogPipe *s)
 
   if (current_internal_source != NULL)
     {
-      msg_error("Multiple internal() sources were detected, this is not possible", NULL);
+      msg_error("Multiple internal() sources were detected, this is not possible");
       return FALSE;
     }
 
--- a/lib/alarms.c
+++ b/lib/alarms.c
@@ -43,7 +43,7 @@ alarm_set(int timeout)
 {
   if (G_UNLIKELY(alarm_pending))
     {
-      msg_error("Internal error, alarm_set() called while an alarm is still active", NULL);
+      msg_error("Internal error, alarm_set() called while an alarm is still active");
       return;
     }
   alarm(timeout);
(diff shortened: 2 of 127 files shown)


MrAnno:
:+1:

bkil-syslogng:
Rebased to master

bazsi:
I like the change in general, even though I didn't have too much time to
check how it works, and neither did I review the individual call sites.


-- 
Bazsi

bkil-syslogng:
Rebased to master.

Fixed conflicts by dropping the last commit and reapplying the script described in the commit message.

ihrwein:
:+1: (Could you rebase it again? Thanks!)

bkil-syslogng:
@ihrwein Rebased to master. No conflict or omission detected when running the transformation this time.

bazsi:
I am afraid I've just merged code that will not be adjusted for this. I would merge this again and fix up the fallout. The change is trivial, makes sense it's just scattered around.

Do you have a script to run on the "http" destination?

bkil-syslogng:
@bazsi You can re-run the commit message as many times as you'd like. Do take care to remove any extra space if your tools show messages that way. Actually I wrapped a long sed line in the middle, that's the (only) one you need to consider.

I did give this a second thought few days ago, and we may consider saving this under `scripts/` and eventually connect it with our Pull Request examination suite in a way that a bot would notify if you used an extra `,NULL`. It's mostly a matter of whether the `git diff` after the transformation is empty or not. In the long run, we could probably connect a bunch of custom LLVM analysis to catch more important typos/common errors as well.

bkil-syslogng:
@bazsi I ran it on the HTTP module. You can check and merge the result and the script file in #1021.

bazsi:
I don't think we need to add this validation from this point on. We don't need it from now period, it may happen that some as-of-now out-of-tree code contains the extra NULL, but we won't write new ones.

bazsi:
thanks a lot and sorry for breaking things. probably not a good idea to
keep PRs out-of-tree for weeks as the chance of stuff breaking when merging
them in improper order is a lot larger.



-- 
Bazsi

PR #951, opened 2016-02-22
Add "curl" module

falzm:

PR #952, opened 2016-02-24
Add "curl" module

falzm:
Fix previous PR (https://github.com/balabit/syslog-ng/pull/951), successfully compiled against syslog-ng master.

### DIFF
--- a/modules/curl/Makefile.am
+++ b/modules/curl/Makefile.am
@@ -1,3 +1,5 @@
+if ENABLE_CURL
+
 module_LTLIBRARIES      += modules/curl/libcurl.la
 modules_curl_libcurl_la_SOURCES = \
   modules/curl/curl-plugin.h        \
@@ -12,6 +14,11 @@ BUILT_SOURCES       +=      \
   modules/curl/curl-grammar.c       \
   modules/curl/curl-grammar.h
 
+modules/curl modules/curl/ mod-curl: modules/curl/libcurl.la
+else
+modules/curl modules/curl/ mod-curl:
+endif
+
 EXTRA_DIST        +=      \
   modules/curl/curl-grammar.ym
 
@@ -24,6 +31,4 @@ modules_curl_libcurl_la_LIBADD  = $(MODULE_DEPS_LIBS) $(LIBCURL_LIBS)
 modules_curl_libcurl_la_LDFLAGS = $(MODULE_LDFLAGS)
 modules_curl_libcurl_la_DEPENDENCIES= $(MODULE_DEPS_LIBS)
 
-modules/curl modules/curl/ mod-curl: modules/curl/libcurl.la
 .PHONY: modules/curl/ mod-curl
-
--- a/modules/curl/Makefile.am
+++ b/modules/curl/Makefile.am
@@ -24,6 +24,6 @@ modules_curl_libcurl_la_LIBADD  = $(MODULE_DEPS_LIBS) $(LIBCURL_LIBS)
 modules_curl_libcurl_la_LDFLAGS = $(MODULE_LDFLAGS)
 modules_curl_libcurl_la_DEPENDENCIES= $(MODULE_DEPS_LIBS)
 
-modules/curl modules/curl/ mod-curl mod-file: modules/curl/libcurl.la
-.PHONY: modules/curl/ mod-curl mod-file
+modules/curl modules/curl/ mod-curl: modules/curl/libcurl.la
+.PHONY: modules/curl/ mod-curl
 
--- a/modules/curl/Makefile.am
+++ b/modules/curl/Makefile.am
@@ -1,0 +1,29 @@
+module_LTLIBRARIES      += modules/curl/libcurl.la
+modules_curl_libcurl_la_SOURCES = \
+  modules/curl/curl-plugin.h        \
+  modules/curl/curl.c               \
+  modules/curl/curl-grammar.y       \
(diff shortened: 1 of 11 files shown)


lbudai:
Hi,

we are reviewing your PR (thanks for it!!), maybe we will put a lots of review notes here, but we will help you finishing your PR.

falzm:
Sure, looking forward to it.

bazsi:
I am thinking about the name. Curl is pretty widespread but shouldnt we use
"http"?

That would probably be a bit more intuitive.

falzm:
I agree with you and have no problem at all changing the name of this module to "http", but wouldn't it conflict with the "http" Java module?

falzm:
@lbudai I've added the init/deinit methods as found in the Java native driver (see commit c2fb835), but I still get the error at shutdown... :/

MrAnno:
Hello @falzm,

Thank you for your contribution!

>ERROR:lib/stats/stats-cluster.c:160:stats_cluster_untrack_counter: assertion failed: (self && (self->live_mask & (1 << type)) && &self->counters[type] == (*counter))

You are getting this error because you call `g_free(self->url)` in the `_thread_deinit` function, but you also have a reference to the `url` field in `_format_stats_instance`. This function is called by syslog-ng at initialization and shutdown time as well.

I think it would be better if you moved your `g_free()` calls from `_thread_deinit` to a `free_fn` 'virtual method':

```c

static void
curl_dd_free(LogPipe *s)
{
  CurlDestinationDriver *self = (CurlDestinationDriver *)s;
  g_free(self->url);
  /* ... other deallocations */

  log_threaded_dest_driver_free(s);
}

/* ... */

LogDriver *
curl_dd_new(GlobalConfig *cfg)
{
  CurlDestinationDriver *self = g_new0(CurlDestinationDriver, 1);

  /* ... */

  self->super.super.super.super.free_fn = curl_dd_free;

  /* ... */
}
```

And because we need to read the `url` field **before** starting the thread, I would also move this if statement to `curl_dd_init`:
```c
if (!self->url) {
  self->url = g_strdup(CURL_DEFAULT_URL);
}
```

falzm:
Thank you for your hints @MrAnno, I've fixed the issue.

falzm:
Unfortunately I don't have much time at this moment to continue improving this module, would you consider taking it from here? As I'm not a professional developer I'm sure you'll do it much better and quicker than me... ;)

lbudai:
Thanks for your work!
I'm not sure that we will finish the work quicker than you :-)
Any volunteer? [maybe a student who want to participate in GSoC?]

falzm:
Unless I'm mistaken there's not much left to be done according to initial comments, maybe just the configurable `User-Agent` field? If I add this would it be OK to merge it into the main tree?

litterbear:
@lbudai  
I'm a GSoC participant. I want to be the volunteer. 
I'll try to add the configurable User-Agent feature first.
If there are any other things that I can improve, please notify me.
I'm glad to help  ;)

modules/curl/Makefile.am
```diff
@@ -0,0 +1,34 @@
...
+
+modules/curl modules/curl/ mod-curl: modules/curl/libcurl.la
+else
+modules/curl modules/curl/ mod-curl:
+endif
+
+EXTRA_DIST        +=      \
+  modules/curl/curl-grammar.ym
+
+modules_curl_libcurl_la_CPPFLAGS  =     \
```
ihrwein: I'd add this block inside the above `if ENABLE_CURL ... else` block.

modules/curl/curl-parser.c
```diff
@@ -0,0 +1,52 @@
...
+  { "password",     KW_PASSWORD },
+  { "url",          KW_URL },
+  { "headers",      KW_HEADERS },
+  { "body",         KW_BODY },
+  { NULL }
+};
+
+CfgParser curl_parser =
+{
+#if ENABLE_DEBUG
```
ihrwein: Please use `SYSLOG_NG_ENABLE_DEBUG` instead of `ENABLE_DEBUG`.

modules/curl/curl-plugin.h
```diff
@@ -0,0 +1,53 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#ifndef SYSLOG_NG_CURL_CURL_H_INCLUDED
```
ihrwein: Could you rename this define to `CURL_PLUGIN_H_INCLUDED`? That way the define names would be consistent.

modules/curl/curl-plugin.h
```diff
@@ -0,0 +1,53 @@
...
+
+typedef struct
+{
+  LogThrDestDriver super;
+  gchar *curl;
+  gchar *url;
+  gchar *user;
+  gchar *password;
+  GList *headers;
+  gchar user_agent[1024];
```
ihrwein: Could you change this field to be heap allocated? Thanks!

litterbear:
@ihrwein  I committed my modifications here #978

falzm:
Any idea if this module has a chance of being accepted to master branch any time soon?

falzm:
Pull request #978 merged.

PR #954, opened 2016-02-26
Fix systemd-journal source

### LINKED ISSUE #914: 3.8 journal source problem
Journal source does not seem to work in current git head. Relevant parts of debug output:

# syslog-ng -Fvd
[...]
[2016-02-10T09:40:39.391235] Reading shared object for a candidate module; path='/usr/lib64/syslog-ng', fname='sdjournal.so', module='sdjournal'
[2016-02-10T09:40:39.391345] Registering candidate plugin; module='sdjournal', context='source', name='systemd-journal', preference='0'
[...]
[2016-02-10T09:40:39.396031] Can't find systemd-journal on this system;
[2016-02-10T09:40:39.396043] Module initialization failed; module='sdjournal'
[2016-02-10T09:40:39.396049] This module claims to support a plugin, which it didn't register after loading; module='sdjournal', context='source', name='systemd-journal'
Error parsing source, source plugin systemd-journal not found in source confgen system at line 3, column 1:
                                                   included from /etc/syslog-ng/syslog-ng.conf line 32, column 9

systemd-journal
^^^^^^^^^^^^^^^

ihrwein:
Fixes #914 

The root cause was, that we mixed up prefixed defines with unprefixed ones.

I saw the following line in a faulty `syslog-ng-config.h`:

```
#define SYSLOG_NG_SYSTEMD_JOURNAL_MODE JOURNALD_OPTIONS
```
There is a typo there (`OPTIONS` -> `OPTIONAL`), and the value part isn't prefixed with `SYSLOG_NG`. It turned out, that `JOURNALD_OPTIONS` was manually injected into the value by the configure script.

If you check `modules/systemd-journal/journald-subsystem.c:28`, you can see that we use conditional compilation and the condition will be never met (as the injected value doesn't have the `SYSLOG_NG` prefix). 

The fallback case didn't work as `libsystemd-journal.so.0` wasn't present.

Note, that I wasn't able to make `AX_PREFIX_CONFIG_H([$CONFIG_H_NAME], [$DEFINE_PREFIX])` work. The injected `JOURNALD_*` defines are prefixed with `SYSLOG_NG` unconditionally. The source code heavily uses the defines, so I don't think that it's a big problem.


bkil-syslogng:
:+1:

Enabling `gcc -Wundef` and staying with the convention of using `#ifdef` instead of `#if` for conflicting cases would have caught this issue.

MrAnno:
:+1:

PR #962, opened 2016-03-01
Fix disabling MongoDB destination support in syslog-ng 3.8

ihrwein:
You can use --with-mongoc=no configure option to disable
this destination.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -168,7 +168,7 @@ AC_ARG_ENABLE(mongodb,
               ,,enable_mongodb="auto")
 
 AC_ARG_WITH(mongoc,
-              [  --with-mongoc=[system/internal/auto]
+              [  --with-mongoc=[system/internal/auto/no]
                                          Link against the system supplied or the built-in mongo-c-driver library. (default: auto)]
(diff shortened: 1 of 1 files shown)


MrAnno:
:+1:

MrAnno:
```
AC_ARG_WITH(mongoc,
              [  --with-mongoc=[system/internal/auto]
                                         Link against the system supplied or the built-in mongo-c-driver library. (default: auto)]
              ,,with_mongoc="auto")
```
Is it OK that the arg description is not updated with the `no` value?

bkil-syslogng:
:+1: Now the fix looks good to me. I agree that it would be nice to extend the arg description.

Just a side note, don't block on this one. It would be beneficial to refactor so as the recommended way to disable MongoDB could be `--enable-mongodb=no`, because it sounds much more intuitive. In this case, I would update the other arg description instead.

ihrwein:
@MrAnno `no` added

MrAnno:
:+1:

lbudai:
could you rebase it?

ihrwein:
rebased, auto squashed, force pushed

PR #969, opened 2016-03-03
Fix value-pairs tokenizer

### LINKED ISSUE #949: @ in format json is absorbed
I have the following elasticsearch configuration:
destination d_elastic_aa0d413f8dc54bba979e079956b53a6b {
    elasticsearch(client_lib_dir(/var/testdb_working_dir/b6b17447-b836-470e-9a41-29ce43fc3e5e/build/distributions/elastic-libs/lib)
 client_mode("node")
 cluster("elasticsearch-syslog-ng-test-jenkins")
 frac_digits(3)
 index("t1f9401b2ff1e390675a7_testbot")
 port("9300")
 resource("/var/testdb_working_dir/934e1c47-ccd2-444c-8527-6831e3a6dcd4.yml")
 send_time_zone("+09:00")
 server("10.140.32.83")
 template("$(format-json date=${DATE} timezone=${TZ} @timestamp=${ISODATE})")
 type("slng_test_type"));
};

the important part is the: @timestamp=${ISODATE}
Now syslog-ng generates the next message:
Outgoing log entry, json='{"timezone":"+09:00","date":"Feb 25 23:09:07.000","timestamp":"2016-02-25T23:09:07.000+09:00"}'

@ is missing in timestamp field thus the stored value belongs to timestamp field instead of @timestamp.

version:
this is the last commit:
commit 7cca32d6c0a97b19b30fb897e48bc677b6d7f471
Merge: 52f535e d1f7098
Author: Balazs Scheidler <balazs.scheidler@balabit.com>
Date:   Fri Feb 19 20:53:45 2016 +0100

    Merge pull request #938 from bkil-syslogng/f/fix-dup-last_parser-tlsssloptions
    
    F/fix dup last parser tlsssloptions

MrAnno:
Fixes #949.

This revert resolves parsing problems when a token starts with '@'.
In that case, the starting @ symbol disappeared because the `token_start` variable was updated at the beginning of the parser loop.

Also, @dnsjts has a completely rewritten `vp_walker_split_name_to_tokens` function which is much cleaner in my opinion.

### DIFF
--- a/lib/value-pairs/value-pairs.c
+++ b/lib/value-pairs/value-pairs.c
@@ -592,8 +592,8 @@ vp_walker_stack_unwind_all_containers(vp_walk_state_t *state)
   vp_walker_stack_unwind_containers_until(state, NULL);
 }
 
-static gint
-vp_walker_skip_sdata_enterprise_id(const gchar *name, gint pos)
+static const gchar *
+vp_walker_skip_sdata_enterprise_id(const gchar *name)
 {
   /* parse .SDATA.foo@1234.56.678 format, starting with the '@'
      character. Assume that any numbers + dots form part of the
@@ -601,36 +601,41 @@ vp_walker_skip_sdata_enterprise_id(const gchar *name, gint pos)
   do
     {
       /* skip @ or . */
-      pos++;
-      pos += strspn(&name[pos], "0123456789");
+      ++name;
+      name += strspn(name, "0123456789");
     }
-  while (name[pos] == '.' && isdigit(name[pos + 1]));
-  return pos;
+  while (*name == '.' && isdigit(*(name + 1)));
+  return name;
 }
 
 static GPtrArray *
 vp_walker_split_name_to_tokens(vp_walk_state_t *state, const gchar *name)
 {
-  gint pos = 0;
-  gint token_start = 0;
-  size_t name_len = strlen(name);
+  const gchar *token_start = name;
+  const gchar *token_end = name;
 
   GPtrArray *array = g_ptr_array_sized_new(VP_STACK_INITIAL_SIZE);
 
-  do
+  while (*token_end)
     {
-      pos += strcspn(&name[pos], "@.");
-      if (name[pos] == '@')
-        pos = vp_walker_skip_sdata_enterprise_id(name, pos);
-
-      if (name[pos] == '.' || pos == name_len)
+      switch (*token_end)
         {
-          g_ptr_array_add(array, g_strndup(&name[token_start], pos - token_start));
-          pos++;
-          token_start = pos;
+        case '@':
+          token_end = vp_walker_skip_sdata_enterprise_id(token_end);
+          break;
+        case '.':
+          g_ptr_array_add(array, g_strndup(token_start, token_end - token_start));
+          ++token_end;
+          token_start = token_end;
+          break;
+        default:
+          ++token_end;
+          token_end += strcspn(token_end, "@.");
         }
     }
-  while (pos < name_len);
+
+  if (token_start != token_end)
+    g_ptr_array_add(array, g_strndup(token_start, token_end - token_start));
 
   if (array->len == 0)
     {
--- a/lib/value-pairs/value-pairs.c
+++ b/lib/value-pairs/value-pairs.c
@@ -617,7 +617,6 @@ vp_walker_split_name_to_tokens(vp_walk_state_t *state, const gchar *name)
   pos = 0;
   do
     {
-      token_start = pos;
       pos += strcspn(&name[pos], "@.");
       if (name[pos] == '@')
         pos = vp_walker_skip_sdata_enterprise_id(name, pos);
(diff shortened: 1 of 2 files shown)


lbudai:
So, the refactor made by @dnsjts also solves this issue?
Should we merge this or wait for a new PR ?

dnsjts:
@lbudai: Yes, it also solves the issue. IMHO we should update this pull request with my changes, and after that it will be ready to review and merge.
@MrAnno: The mentioned changes can be found here (https://github.com/dnsjts/syslog-ng/commit/1156c3dc68c4057d59b1e8d93bafeef56a0c3137), please review the commits (and commitids), and reset your branch to the mentioned commit.

lbudai:
ok, then I put the blocked label and when you updated, just remove this label.

MrAnno:
Fixups have been applied; rebased to master.

bkil-syslogng:
About the clarification commit.

@dnsjts Actually I would iterate until `'\0'` instead of `strlen()`, because this way we traverse the string twice. I know this was how the original version worked, but boy scout's rule dictates that if you are refactoring something so deeply, you should fix this as well. This would change of course if we want to support binary input and/or already have a length for free along the call chain.

Additionally, I'd change this line:
`pos += strcspn(&name[++pos], "@.");`

Nitpicks:
* I'd skip allocation of the GPtrArray until it is needed instead of always allocating it and then freeing it if unused.
* I'd probably also iterate with pointers instead of indices in this case (although, it should hopefully compile to the same)

lib/value-pairs/value-pairs.c
```diff
@@ -601,34 +601,41 @@ vp_walker_skip_sdata_enterprise_id(const gchar *name, gint pos)
...
         {
-          g_ptr_array_add(array, g_strndup(&name[token_start], pos - token_start));
-          pos++;
+          g_ptr_array_add(array, g_strndup(token_start, token_end - token_start));
+          ++token_end;
+          token_start = token_end;
+        }
+      else
+        {
+          token_end += strcspn(token_end, "@.");
```
bkil-syslogng: The fixes look very good.

However, one of my nitpicks still remains: why wouldn't you like to have a `++token_end` before this `strcspn()`?

MrAnno: I forgot to add this modification, sorry.

bkil-syslogng:
:+1:

lib/value-pairs/value-pairs.c
```diff
@@ -617,20 +617,26 @@ vp_walker_split_name_to_tokens(vp_walk_state_t *state, const gchar *name)
...
         }
+      else
+        {
+          pos += strcspn(&name[pos], "@.");
+        }
     }
-  while (pos < name_len);
+
+  if ((pos - token_start) > 0)
+    g_ptr_array_add(array, g_strndup(&name[token_start], pos - token_start));
```
lbudai: In commit https://github.com/balabit/syslog-ng/pull/969/commits/dd916934a2dc353eb62824f21ab493f16b964ac0
did you add testcase for testing the situation you fixed here?

MrAnno: This commit is only for clarification.
```C
if ((pos - token_start) > 0)
  g_ptr_array_add(array, g_strndup(&name[token_start], pos - token_start));
```
The additional if statement is needed because @dnsjts didn't want to "duplicate" the loop condition (`pos < name_len`) within the loop (like it was before):

```C
do
  {
    /* ... */
    if (name[pos] == '.' || pos == name_len)
    /* ... */
  }
while (pos < name_len);
```

bkil-syslogng: Your solution also sounds more efficient. And from a higher perspective, I
can understand the intention of this code part as well.

dnsjts: The other version would be a bit more complex body of the loop.
I think this one is the lesser of two evils.

dnsjts:
The body of the loop can be rewritten to a simple switch-case statement. Should it be done?

lib/value-pairs/value-pairs.c
```diff
@@ -601,34 +601,42 @@ vp_walker_skip_sdata_enterprise_id(const gchar *name, gint pos)
...
-  do
+  while (*token_end)
     {
-      token_start = pos;
-      pos += strcspn(&name[pos], "@.");
-      if (name[pos] == '@')
-        pos = vp_walker_skip_sdata_enterprise_id(name, pos);
-
-      if (name[pos] == '.' || pos == name_len)
+      if (*token_end == '@')
```
dnsjts: The body of the loop can be rewritten to a simple switch-case statement. Should it be done?

bkil-syslogng: My response to this is that it looks good in both forms to me. It's still
plus one from me.

We're waiting for another viewpoint on this question and/or a plus one
overall, so we can finally merge it.

MrAnno: With switch-case, I think it's more like a state machine.

bkil-syslogng:
@MrAnno State machine looks good.

ihrwein:
:+1: Could you do a rebase? After that, I'd merge it. Thanks!

MrAnno:
Squashed, rebased to master.

PR #976, opened 2016-03-05
F/propagate cflags to tests

### LINKED ISSUE #903: Verify that CFLAGS is propagated down to all of our sources
@MrAnno revealed that some of our sources, for example some of the unit tests do not propagate CFLAGS (correctly). I.e., specifying `-Werror` to `configure` does not stop in case of errors in some of the tests.

TODO: describe issue, solution, etc (TEST_CFLAGS, etc.)

bkil-syslogng:
Fixes #903

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,22 @@ before_script:
   - unset PYTHON_CFLAGS # HACK
   - ./configure CFLAGS=-Werror --with-ivykis=internal --prefix=$HOME/install/syslog-ng --enable-pacct --enable-manpages --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl --enable-extra-warnings
   - if [ "$CC" = "clang" ]; then export COPYRIGHTVERBOSITY=1; make check-copyright; fi
-  - make -j || make V=1 --keep-going # to make error messages more readable on error
+  - . tests/build-log-cflags-propagation.sh;
+    exec_prop_check "make -j V=1";
+    S=$?;
+    if [ "$S" = "42" -o "$S" = "0" ]; then
+      return $S;
+    else
+      make V=1 --keep-going; false;
+    fi # to make error messages more readable on error
 script:
-  - if [ "$CC" = "gcc" ]; then make distcheck V=1; else make func-test V=1; fi
+  - if [ "$CC" = "gcc" ]; then
+      . tests/build-log-cflags-propagation.sh;
+      export DISTCHECK_CONFIGURE_FLAGS="CFLAGS=-Werror --enable-extra-warnings";
+      exec_prop_check "make distcheck V=1 --keep-going";
+    else
+      make func-test V=1;
+    fi
 compiler:
   - gcc
   - clang
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
 before_script:
   - ./autogen.sh
   - unset PYTHON_CFLAGS # HACK
-  - ./configure CFLAGS=-Werror --with-ivykis=internal --prefix=$HOME/install/syslog-ng --enable-pacct --enable-manpages --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl 
+  - ./configure CFLAGS=-Werror --with-ivykis=internal --prefix=$HOME/install/syslog-ng --enable-pacct --enable-manpages --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl --enable-extra-warnings
   - if [ "$CC" = "clang" ]; then export COPYRIGHTVERBOSITY=1; make check-copyright; fi
   - make -j || make V=1 --keep-going # to make error messages more readable on error
 script:
--- a/configure.ac
+++ b/configure.ac
@@ -135,6 +135,9 @@ AC_ARG_WITH([systemdsystemunitdir],
 AC_ARG_ENABLE(debug,
               [  --enable-debug      Enable debugging code.],, enable_debug="no")
 
+AC_ARG_ENABLE(extra-warnings,
+              [  --enable-extra-warnings      Enable extra compiler warnings (recommended).],, enable_extra_warnings="no")
+
 AC_ARG_ENABLE(env-wrapper,
               [  --enable-env-wrapper    Enable wrapper program to set up various environment variables],, enable_env_wrapper=auto)
 
@@ -384,7 +387,7 @@ dnl ***************************************************************************
 dnl Set up CFLAGS
 
 if test "x$ac_compiler_gnu" = "xyes"; then
-        CFLAGS="${CFLAGS} -Wall -Wshadow"
+        CFLAGS="${CFLAGS} -Wall"
         if test "x$enable_debug" = "xyes"; then
                 CFLAGS="${ac_cv_env_CFLAGS_value} -Wall -g"
         fi
@@ -395,6 +398,8 @@ if test "x$ac_compiler_gnu" = "xyes"; then
         if test "x$enable_gcov" = "xyes"; then
                 CFLAGS="${CFLAGS} -fprofile-arcs -ftest-coverage"
         fi
+else
+        enable_extra_warnings="no"
 fi
 CFLAGS="${CFLAGS} -pthread"
 
@@ -1543,6 +1548,7 @@ AM_CONDITIONAL(ENABLE_JAVA, [test "$enable_java" = "yes"])
 AM_CONDITIONAL(ENABLE_JAVA_MODULES, [test "$enable_java_modules" = "yes"])
 AM_CONDITIONAL(ENABLE_MANPAGES, [test "$enable_manpages" != "no"])
 AM_CONDITIONAL(ENABLE_NATIVE, [test "$enable_native" != "no"])
+AM_CONDITIONAL(ENABLE_EXTRA_WARNINGS, [test "$enable_extra_warnings" = "yes"])
 
 AC_SUBST(timezonedir)
 AC_SUBST(pidfiledir)
--- a/tests/build-log-cflags-propagation.sh
+++ b/tests/build-log-cflags-propagation.sh
@@ -1,0 +1,121 @@
+#!/bin/sh
+#############################################################################
+# Copyright (c) 2016 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+exec_prop_check() {
+  local ERROREXIT="42"
+  local CMD="$1"
+  local BUILDLOG="build.$$.log"
+  {
+    eval "$CMD"
+    echo $?
+  } |
+  tee "$BUILDLOG" |
+  reduce_verbosity
+
+  S=`tail -n 1 "$BUILDLOG"`
+  if [ "$S" -eq 0 ]; then
+    build_log_cflags_propagation "$BUILDLOG"
+    S=$?
+    rm "$BUILDLOG"
+    return $S
+  else
+    rm "$BUILDLOG"
+    return $ERROREXIT
+  fi
+}
+
+####
+# private functions:
+
+reduce_verbosity() {
+  grep --line-buffered --invert-match --extended-regexp "^(\
+libtool: (link|relink|install): |\
+depbase=|\
+(test -z|rm|\./lib/merge-grammar.pl|\./doc/mallard2man\.py) |\
+`printf "\t"`?/bin/bash |\
+`printf "\t"`?(gcc|mv) \
+)"
+}
+
+build_log_cflags_propagation() {
+  local BUILDLOG="$1"
+  local NOTPROP="`find_not_prop "$BUILDLOG"`"
+  if [ -z "$NOTPROP" ]; then
+    echo "info: CFLAGS propagation test passed" >&2
+    true
+  else
+    local N=`echo "$NOTPROP" | wc -l`
+    echo "error: -Wshadow/-Werror did not propagate via CFLAGS in $N cases:" >&2
+
+    printf "%s\n" "$NOTPROP" |
+    sed "s~^~> ~" >&2
+    false
+  fi
+}
+
+find_not_prop() {
+  {
+    find_gcc "$@" |
+    ignore_submodule_gcc |
+    grep -vE -- " -Wshadow( |.* )-Werror "
+  } 2>&1
+}
+
+find_gcc() {
+  grep -E "^(\
+libtool: compile: +gcc |\
+`printf "\t"`?/bin/bash \./libtool  --tag=CC   --mode=compile gcc |\
+`printf "\t"`?gcc \
+)" "$@"
+}
+
+ignore_submodule_gcc() {
+  ignore_submodule_gcc_mongo_c_driver "$@" |
+  ignore_submodule_gcc_rabbitmq_c |
+  ignore_submodule_gcc_ivykis
+}
+
+ignore_submodule_gcc_mongo_c_driver() {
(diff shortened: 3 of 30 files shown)


bazsi:
LGTM

:+1:

bkil-syslogng:
Please wait a little bit before merging. I suspect that propagation might be missing at another place and want to look into it. Deadline (timebox): until 2016-03-09T09:00

lbudai:
@bkil-syslogng : so, did you check it?

bkil-syslogng:
Ready for reviewing. My feeling was correct - lots of further omissions have been fixed since the last review round, so please update your thumbs.

tests/build-log-cflags-propagation.sh
```diff
@@ -0,0 +1,117 @@
...
+exec_prop_check() {
+  local ERROREXIT="42"
+  local CMD="$1"
+  local BUILDLOG="build.$$.log"
+  {
+    eval "$CMD"
+    echo $?
+  } |
+  tee "$BUILDLOG" |
+  grep --line-buffered --invert-match --extended-regexp "^(\
```
bkil: It would be nice to extract this `grep` to something like `reduce_verbosity()`.

bkil-syslogng:
Rebased to master, squashed fixups.

Backup save under:
* https://github.com/balabit/syslog-ng/compare/master...bkil-syslogng:t/propagate-cflags-to-tests-bk0?expand=1

.travis.yml
```diff
@@ -12,9 +12,22 @@ before_script:
   - unset PYTHON_CFLAGS # HACK
   - ./configure CFLAGS=-Werror --with-ivykis=internal --prefix=$HOME/install/syslog-ng --enable-pacct --enable-manpages --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl 
   - if [ "$CC" = "clang" ]; then export COPYRIGHTVERBOSITY=1; make check-copyright; fi
-  - make -j || make V=1 --keep-going # to make error messages more readable on error
+  - . tests/build-log-cflags-propagation.sh;
+    exec_prop_check "make -j V=1";
+    S=$?;
+    if [ "$S" = "42" -o "$S" = "0" ]; then
+      return $S;
```
bazsi: what is exit code 42?

.travis.yml
```diff
@@ -12,9 +12,22 @@ before_script:
   - unset PYTHON_CFLAGS # HACK
   - ./configure CFLAGS=-Werror --with-ivykis=internal --prefix=$HOME/install/syslog-ng --enable-pacct --enable-manpages --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl 
   - if [ "$CC" = "clang" ]; then export COPYRIGHTVERBOSITY=1; make check-copyright; fi
-  - make -j || make V=1 --keep-going # to make error messages more readable on error
+  - . tests/build-log-cflags-propagation.sh;
+    exec_prop_check "make -j V=1";
+    S=$?;
```
bazsi: what is exec_prop_check here? I don't see where that is coming from.

configure.ac
```diff
@@ -135,6 +135,9 @@ AC_ARG_WITH([systemdsystemunitdir],
 AC_ARG_ENABLE(debug,
               [  --enable-debug      Enable debugging code.],, enable_debug="no")
 
+AC_ARG_ENABLE(extra-warnings,
+              [  --disable-extra-warnings      Disable extra compiler warnings.],, enable_extra_warnings="yes")
+
 AC_ARG_ENABLE(env-wrapper,
```
bazsi: I don't see why we need this switch, but I doesn't bother me too much either. It defaults to "yes" and I doubt anyone would ever disable this explicitly.

bazsi:
Apart from a couple of questions (which are just things I didn't understand and not things that should be changed), this looks good to me, so we should merge this finally.

configure.ac
bkil-syslogng: We agreed with @dnsjts that we should default this to no, I will send a fixup soon. I want to add dozens of further warnings under this condition. Are you absolutely positive that we should separate `-Wall` from the rest? I would list all of them here (or optionally pass each of them from the configure command line, but that's a bit overkill for our purposes).

.travis.yml
bkil-syslogng: It's coming from the included file: `tests/build-log-cflags-propagation.sh`. It guarantees that all flags are propagated. We should probably extend it with linker flag checking later on

.travis.yml
bkil-syslogng: We retunn 42 if propagation check failed, but the build otherwise succeeded. 0 means both succeeded. Other values signal a build failure. We could use another extremal value if you prefer.

configure.ac
bazsi: Please don't remove -Wall from the default. Having to enable this option
for current and future developers is another thing we would need to look
after. It should work out of the box.

Either of these solutions work for me:
1) have two sets of warnings, one always enabled the other optionally
enabled.
2) have one set, but then enable them all.

Since I don't know the kind of warnings you want to enable option 1. should
be the safer bet. But number two works for me too. If a  warning option
does something I don't agree with we can always remove that later.

My priority is our of the box sane warnings without explicit configure
options.

.travis.yml
bazsi: Not really, I am fine  with 42, a comment would be useful though.

bkil-syslogng:
@bazsi We can discuss this separately because it is a long topic.

TL;DR: A developer *must* see all (whitelisted) warnings. However, a user or a package maintainer wouldn't have much use in them and it could even cause confusion, epsecially if they had set `-Werror` somewhere along the pipeline. Warnings come and go with compiler versions, even the same ones can produce different effects as we've seen in the past.

This may seem pushing it a but too much, but my recommended solution would involve something even more radical.

I agree with @dnsjts that CFLAGS should be left as a realm of the user, not something we would like to play around with. However, as a trade off, it would be worth it to provide different profiles for different scenarios, similar to what I've mentioned in the `make ci`/`quick`/`full` argument in the past. (Participants other then the user would also need to do dynamic analysis on their product (unit/func testing))

Basically, both a developer's and a CI's configure must enable all cheap analysis available (compiler warnings) and preferably enable all modules (cross-module breakage is possible if altering the core). Developers should probably prefer to use memory tracing and debug symbols as well. A CI would enable expensive analysis and full (non-smoke) testing as well. The user profile would implicate none of the above. Perhaps even this profile configuration option would alter the behavior of `make check` to implicate different levels of tests.

So maybe we should introduce `./configure --enable-devel` which would be a shorthand for:
* enable extra warnings
* enable all modules
* enable debug symbols
* run copyright checking on `make check` (hopefully incrementally on changed files only by that time)
* run functional tests on `make check` (in parallel with make check)
* run `make distcheck` after (or besides) `make check`
* run all other advanced future checks on `make check`

It would default to disabled, so users will have trouble free out of the box experience.

I guess most of the above bullet points could have a configure option by itself if someone is only interested in only one or the other.

bazsi:
I am not sure when we can discuss IRL so here are my thoughts.

1) known good and widely used warnings (e.g. Wall) should even be enabled
in user builds, simply because they should not trigger warnings, ever.
Stuff like undeclared functions should be reported, if the other platform
the user compiles on doesn't match the one developers work on.

2) Werror shouldn't be enabled in user builds, we agree on that front.

3) I as a developer want fine-grained control over my test runs and
configure options don't cut it. Why don't we simply use separate make
targets and a combined one to be used by the CI? In case I want a quick
unit test run I want to run make check. If I want the complete suite I run
make CI. I don't want copyright and distcheck in my TDD loops.

4) it's pretty usual to compile production code with O2 and -g and strip
debug symbols into a separate dbg package. And debug symbols help in
diagnosing custom compiled syslog-ng issues without a rebuild.

In traditional UNIX build scripts CFLAGS  is both the realm of the user and
that of the application. That's why there are clear ordering rules so that
the application can be overridden but still provide defaults.

bkil-syslogng:
TL;DR: So all in all, you'd like me to enable -Wall on gcc by default, and
only make enabling of the other extra warnings optional?

Re: @4) I did not imply that we should disable debug symbols on release
builds, I only implied that we should enable them on devel builds.

Re: @3) You are free to use individual options if the default one doesn't
suite you - just as you can enable separate modules instead of
enable-all-modules.

I will probably implement `--enable-devel` in a separate pull request after
some more consideration, because it has many benefits without any drawbacks
to anyone whatsoever. I will consider whether make targets could or should
be altered.

Here are a few flags I have been experimented with. I've POC'ed syslog-ng
until it kind of passed, and the type of changes that were required were
all reasonable, and looked good in code. I've also uncovered a few bugs
while at it, but I still haven't managed to find time to open a PR about
them (we have too many PRs already anyway). We can't enable (most of) these
yet, but ultimately I would like to enforce all of them. This list is
neither double checked nor final - I only copy & pasted it from one of my
sandbox WIP PRs to let you get an idea about the scope of uncertainty:

-Wall -W -Wextra -Wshadow \
  -fno-common -Wdouble-promotion -Wformat=2 -Wcast-align \
  -Wbad-function-cast -Wmissing-prototypes -Wmissing-declarations \
  -Wnested-externs -Wcast-qual -Wwrite-strings -Wfloat-equal \
  -Wabi -Waggregate-return -Winvalid-pch \
  -Wmissing-format-attribute -Wdeprecated-declarations -Woverflow \
  -Wpointer-arith -Wpointer-sign \
  -Wno-unused-parameter -Wno-missing-field-initializers \
  -Wvariadic-macros \
  -Wstrict-prototypes -Wold-style-definition \
  -Wconversion -Wno-error=conversion \
  -Wsign-conversion -Wno-error=sign-conversion \
  -Wredundant-decls -Wno-error=redundant-decls \
  -Wstack-protector -Wno-error=stack-protector \
  -Wdeprecated -Wno-error=deprecated \
  -Wundef -Wno-error=undef


On Thu, Apr 28, 2016 at 10:52 PM, Balazs Scheidler <notifications@github.com
> wrote:

> I am not sure when we can discuss IRL so here are my thoughts.
>
> 1) known good and widely used warnings (e.g. Wall) should even be enabled
> in user builds, simply because they should not trigger warnings, ever.
> Stuff like undeclared functions should be reported, if the other platform
> the user compiles on doesn't match the one developers work on.
>
> 2) Werror shouldn't be enabled in user builds, we agree on that front.
>
> 3) I as a developer want fine-grained control over my test runs and
> configure options don't cut it. Why don't we simply use separate make
> targets and a combined one to be used by the CI? In case I want a quick
> unit test run I want to run make check. If I want the complete suite I run
> make CI. I don't want copyright and distcheck in my TDD loops.
>
> 4) it's pretty usual to compile production code with O2 and -g and strip
> debug symbols into a separate dbg package. And debug symbols help in
> diagnosing custom compiled syslog-ng issues without a rebuild.
>
> In traditional UNIX build scripts CFLAGS is both the realm of the user and
> that of the application. That's why there are clear ordering rules so that
> the application can be overridden but still provide defaults.

bazsi:
On Apr 29, 2016 5:27 PM, "Tamas Nagy" <notifications@github.com> wrote:
>
> TL;DR: So all in all, you'd like me to enable -Wall on gcc by default, and
> only make enabling of the other extra warnings optional?

Yup, -Wall always on. The remainder is up to you.

>
> Re: @4) I did not imply that we should disable debug symbols on release
> builds, I only implied that we should enable them on devel builds.

Yes, but you said that CFLAGS should be user controlled completely or
that's how I understood it. So -g should also be on, just like -O2.

>
> Re: @3) You are free to use individual options if the default one doesn't
> suite you - just as you can enable separate modules instead of
> enable-all-modules.
>
> I will probably implement `--enable-devel` in a separate pull request
after
> some more consideration, because it has many benefits without any
drawbacks
> to anyone whatsoever. I will consider whether make targets could or should
> be altered.

I would like to use enable devel myself, but even with enable devel I want
to be able to run unit tests without copyright checks, and functional test
and coverage analysis. Those all should be available separately.

Then make CI would depend on the individual validation steps.

>
> Here are a few flags I have been experimented with. I've POC'ed syslog-ng
> until it kind of passed, and the type of changes that were required were
> all reasonable, and looked good in code. I've also uncovered a few bugs
> while at it, but I still haven't managed to find time to open a PR about
> them (we have too many PRs already anyway). We can't enable (most of)
these
> yet, but ultimately I would like to enforce all of them. This list is
> neither double checked nor final - I only copy & pasted it from one of my
> sandbox WIP PRs to let you get an idea about the scope of uncertainty:
>
> -Wall -W -Wextra -Wshadow \
> -fno-common -Wdouble-promotion -Wformat=2 -Wcast-align \
> -Wbad-function-cast -Wmissing-prototypes -Wmissing-declarations \
> -Wnested-externs -Wcast-qual -Wwrite-strings -Wfloat-equal \
> -Wabi -Waggregate-return -Winvalid-pch \
> -Wmissing-format-attribute -Wdeprecated-declarations -Woverflow \
> -Wpointer-arith -Wpointer-sign \
> -Wno-unused-parameter -Wno-missing-field-initializers \
> -Wvariadic-macros \
> -Wstrict-prototypes -Wold-style-definition \
> -Wconversion -Wno-error=conversion \
> -Wsign-conversion -Wno-error=sign-conversion \
> -Wredundant-decls -Wno-error=redundant-decls \
> -Wstack-protector -Wno-error=stack-protector \
> -Wdeprecated -Wno-error=deprecated \
> -Wundef -Wno-error=undef
>
>
>
> On Thu, Apr 28, 2016 at 10:52 PM, Balazs Scheidler <
notifications@github.com
> > wrote:
>
> > I am not sure when we can discuss IRL so here are my thoughts.
> >
> > 1) known good and widely used warnings (e.g. Wall) should even be
enabled
> > in user builds, simply because they should not trigger warnings, ever.
> > Stuff like undeclared functions should be reported, if the other
platform
> > the user compiles on doesn't match the one developers work on.
> >
> > 2) Werror shouldn't be enabled in user builds, we agree on that front.
> >
> > 3) I as a developer want fine-grained control over my test runs and
> > configure options don't cut it. Why don't we simply use separate make
> > targets and a combined one to be used by the CI? In case I want a quick
> > unit test run I want to run make check. If I want the complete suite I
run
> > make CI. I don't want copyright and distcheck in my TDD loops.
> >
> > 4) it's pretty usual to compile production code with O2 and -g and strip
> > debug symbols into a separate dbg package. And debug symbols help in
> > diagnosing custom compiled syslog-ng issues without a rebuild.
> >
> > In traditional UNIX build scripts CFLAGS is both the realm of the user
and
> > that of the application. That's why there are clear ordering rules so
that
> > the application can be overridden but still provide defaults.
> > On Apr 28, 2016 6:30 PM, "Tamas Nagy" <notifications@github.com> wrote:
> >
> > > @bazsi <https://github.com/bazsi> We can discuss this separately
because
>
> > > it is a long topic.
> > >
> > > TL;DR: A developer *must* see all (whitelisted) warnings. However, a
user
>
> > > or a package maintainer wouldn't have much use in them and it could
even
> > > cause confusion, epsecially if they had set -Werror somewhere along
the
> > > pipeline. Warnings come and go with compiler versions, even the same
ones
> > > can produce different effects as we've seen in the past.
> > >
> > > This may seem pushing it a but too much, but my recommended solution
> > would
> > > involve something even more radical.
> > >
> > > I agree with @dnsjts <https://github.com/dnsjts> that CFLAGS should be
>
> > > left as a realm of the user, not something we would like to play
around
> > > with. However, as a trade off, it would be worth it to provide
different
> > > profiles for different scenarios, similar to what I've mentioned in
the
> > make
> > > ci/quick/full argument in the past. (Participants other then the user
> > > would also need to do dynamic analysis on their product (unit/func
> > testing))
> > >
> > > Basically, both a developer's and a CI's configure must enable all
cheap
> > > analysis available (compiler warnings) and preferably enable all
modules
> > > (cross-module breakage is possible if altering the core). Developers
> > should
> > > probably prefer to use memory tracing and debug symbols as well. A CI
> > would
> > > enable expensive analysis and full (non-smoke) testing as well. The
user
> > > profile would implicate none of the above. Perhaps even this profile
> > > configuration option would alter the behavior of make check to
implicate
> > > different levels of tests.
> > >
> > > So maybe we should introduce ./configure --enable-devel which would
be a
> > > shorthand for:
> > >
> > > - enable extra warnings
> > > - enable all modules
> > > - enable debug symbols
> > > - run copyright checking on make check (hopefully incrementally on
>
> > > changed files only by that time)
> > > - run functional tests on make check (in parallel with make check)
> > > - run make distcheck after (or besides) make check
> > > - run all other advanced future checks on make check
>
> > >
> > > It would default to disabled, so users will have trouble free out of
the
> > > box experience.
> > >
> > > I guess most of the above bullet points could have a configure option
by

bkil-syslogng:
Moved `-Wall` back to being mandatory, the way it originally was. Fixed newly introduced shadowing in `test_patterndb.c`

Backup saved:
* https://github.com/balabit/syslog-ng/compare/master...bkil-syslogng:t/propagate-cflags-to-tests-bk1?expand=1

Rebased to `master`.

bazsi:
Merging this, great work.

I do have some concerns with regards to the shell scripts that validate whether propagation works. They are pretty far from trivial, and I am not sure we will be able to properly maintain them, once stuff breaks, but anyway it is still a great addition and we'll see what problems it causes when we get there.

Issue #956, opened 2016-02-26
Idea: enable -Wundef and refactor conflicting #if usage to #ifdef


bkil-syslogng:
The compiler can help us detect misconfiguration if we never use undefined symbols. This means that we should always `#define` a configuration option, even if it is `FALSE` (`0`).

An alternative would be to alter some of the condition checks to use `#ifdef` instead. Because in some cases, both of them could look nice, we probably need to either stick to one convention or the other, or start using a prefixing convention to help differentiate between a macro for which a value check is needed vs. an existence check.

Caught one in #954 (and some others in the past)

Kokan:
Closes #2806

PR #977, opened 2016-03-05
Fix defines in diskq

ihrwein:
I saw a warning on OSX and checked the diskq module for further prefixing issues. These two were the only ones.


bkil-syslogng:
:+1: Thank you for these improvements. Note that #956 would have caught these issues.

Issue #970, opened 2016-03-04
Enhancement: add http transport for elasticsearch destination


faxm0dem:
This will have many benefits:
1. many people don't need more performance
2. compatibility will be much improved with respect to the frequent ES API changes
3. decouple syslog-ng and elasticsearch upgrades
4. alleviate the jvm dependency (see next paragraph)

The best part is that it should be easy to implement using the SCL and the new curl destination (or the java http). I'd prefer @falzm's implementaton (see benefit number 4.)

lbudai:
:+1:

falzm:
:+1:

faxm0dem:
this is now implemented in java.
I guess if someone still needs point 4., they should open a new issue

Kokan:
as @faxm0dem it is done, but also probably in the feature curl based solution is going to replace jvm based one.

PR #978, opened 2016-03-06
Improve "curl" module

litterbear:
improve previous PR #952 

I just fix some test bugs so far.
I'll try to add the configurable User-Agent feature in next step..


litterbear:
@lbudai  @falzm 
My improvements will be committed here

litterbear:
Hi,
I've committed the  configurable User-Agent feature.

I use the config below
```
@version: 3.7
@include "scl.conf"
source      s_system { system(); internal();                 };
destination curl_des {
    curl(
        url("http://127.0.0.1:8123")
        user_agent("litter bear  Agent")
    );
};
log                  { source(s_system); destination(curl_des); };
```

Then I tested use this command
```
root@devstack:~/gsoc/debug_server# logger test
```

I verified it with a simple django webserver  I wrote today.
If I comment out `user_agent("litter bear  Agent")`, it will use `syslog-ng 3.8.0alpha0/libcurl 7.35.0` as the default agent. Otherwise it will use the agent I configued
```
syslog-ng 3.8.0alpha0/libcurl 7.35.0
[06/Mar/2016 16:46:48] "POST / HTTP/1.1" 200 0
litter bear  Agent
[06/Mar/2016 16:46:58] "POST / HTTP/1.1" 200 0
```

If there is other things that I can do to improve  "curl" module to make it merged, please let me know.
Very glad to contribute more ;)

modules/curl/curl.c
```diff
@@ -0,0 +1,311 @@
...
+}
+
+void
+curl_dd_set_user_agent(LogDriver *d, const gchar *user_agent)
+{
+  CurlDestinationDriver *self = (CurlDestinationDriver *) d;
+
+  g_free(self->password);
+  g_stpcpy(self->user_agent, user_agent);
+  /* self->password = g_strdup(user_agent); */
```
falzm: I think this isn't necessary.

modules/curl/curl.c
```diff
@@ -0,0 +1,311 @@
...
+  self->password = g_strdup(password);
+}
+
+void
+curl_dd_set_user_agent(LogDriver *d, const gchar *user_agent)
+{
+  CurlDestinationDriver *self = (CurlDestinationDriver *) d;
+
+  g_free(self->password);
+  g_stpcpy(self->user_agent, user_agent);
```
falzm: You should check for the size of the input to avoid writing beyond the self->user_agent `char[1024]` buffer limit – use `g_strlcpy()` instead?

litterbear:
@falzm  
Thanks for pointing these out.   Fixed now   ;P

litterbear:
@lbudai  
Sorry about pressing the  wrong button  "Close and comment". 
I reopened the PR.

modules/curl/curl.c
```diff
@@ -0,0 +1,310 @@
...
+  g_free(self->password);
+  self->password = g_strdup(password);
+}
+
+void
+curl_dd_set_user_agent(LogDriver *d, const gchar *user_agent)
+{
+  CurlDestinationDriver *self = (CurlDestinationDriver *) d;
+
+  g_free(self->password);
```
falzm: Why do you free `self->password` here?

litterbear: @falzm 
It's a mistake ,  fixed now.
Thanks

lbudai:
@litterbear : could you process the review notes for #952  and fix them first?
Then you could rebase the `user_agent` changes.

configure.ac
```diff
@@ -1009,9 +1017,40 @@ if test "x$enable_smtp" != "xno" && test "x$with_libesmtp" != "no"; then
 fi
 
 dnl ***************************************************************************
+dnl libcurl headers/libraries
+dnl ***************************************************************************
+if test "x$enable_curl" != "xno" && test "x$with_libcurl" != "xno"; then
+	libcurl="yes"
+	if test "x$with_libcurl" != "yes" && test "x$with_libcurl" != "x"; then
```
ihrwein: `x` is missing before `yes`

configure.ac
```diff
@@ -1009,9 +1017,40 @@ if test "x$enable_smtp" != "xno" && test "x$with_libesmtp" != "no"; then
...
+	enable_curl=$libcurl
+fi
+
+dnl ***************************************************************************
 dnl libhiredis headers/libraries
 dnl ***************************************************************************
-if test "x$enable_redis" != "xno" && test "x$with_redis" != "no"; then
+if test "x$enable_redis" != "xno" && test "x$with_redis" != "xno"; then
        hiredis="yes"
        if test "x$with_libhiredis" != "yes" && test "x$with_libhiredis" != "x"; then
```
ihrwein: `yes` also needs an x prefix.

litterbear:
I reviewed #952 . Most are fixed by falzm.

I finished items below
1)  make user_agent dynamically allocated
2)  add the prefix x for "yes"   as  @ihrwein mentioned
3)  split func _insert to smaller ones
4) move configures into `if ENABLE_CURL` block
5) replace ENABLE_DEBUG with SYSLOG_NG_ENABLE_DEBUG
6) replace SYSLOG_NG_CURL_CURL_H_INCLUDED


@lbudai    I have some questions below.
1) should I switch the  --disable-curl to --enable-curl ?  https://github.com/balabit/syslog-ng/pull/952#commitcomment-16298893
2) should I  change the name of curl module to "http"? https://github.com/balabit/syslog-ng/pull/952#issuecomment-188448256

faxm0dem:
my two bitcoins for 2): I think we should provide a SCL that provides `http` or `web` destination and wraps to `libcurl` or the java version

bazsi:
Does the java version a support more options? I don't exactly see why we need both if they are functionally equivalent.

faxm0dem:
good point. I have one *possible* reason to keep the java version: if it is required to provide a `client_mode("http")` for the elasticsearch destination (see #970)

lbudai:
@faxm0dem : 
If we want to support Elasticsearch destination via HTTP, then I'd say we should go with the libcurl based HTTP (in that case it will be possible to send logs to an Elasticsearch cluster without java, please note, there is no shield support in this case).

The `client_mode("http")`: the simplest solution would be to add a thin HTTP client to the existing Elasticsearch destination.

faxm0dem:
okay then!

litterbear:
Hi,
I got a problem.
I want to test if  the headers setting will take effect.
The definition of headers setting can be found [here](https://github.com/balabit/syslog-ng/pull/978/files#diff-6d934dff7e82e781ba6e0f0d448eee5bR98)

I refer to the usage of `tags()` in the [document](https://www.balabit.com/sites/default/files/documents/syslog-ng-ose-latest-guides/en/syslog-ng-ose-guide-admin/html/reference-filters.html#example-tags-filtering)
```
source s_tcp {

    network(ip(192.168.1.1) port(1514) tags("tcp", "router"));

};
```
Because I found that it also use string_list to define its usage in `./lib/filter/filter-expr-parser.c`.



Then I use the config below
```
@version: 3.7
@include "scl.conf"
source      s_system { system(); internal(); };
destination http_des {
    http(
        url("http://127.0.0.1:8123")
        user_agent("syslog-ng User Agent")
        headers("X-Syslog-HostAHAHA: asdfasdf", "X-Syslog-HostAHAHA2: asdfasdf", "X-Syslog-HostAHAHA3: asdfasdf")
    );
};
```

The headers I defined(i.e. `"X-Syslog-HostAHAHA: asdfasdf", "X-Syslog-HostAHAHA2: asdfasdf", "X-Syslog-HostAHAHA3: asdfasdf"` ) never appear in the headers.
Then I use gdb to debug [this function](https://github.com/balabit/syslog-ng/pull/978/files#diff-aaa0c54b462885700a85d4ce265086f8R114)

I found that the string in the memory is not what I defined
```
(gdb) l
111       curl_headers = curl_slist_append(curl_headers, header_level);
112
113       header = self->headers;
114       printf("%s\n", header_host);
115       printf("%s\n", header_program);
116       while (header != NULL) {
117         printf("%s\n", (gchar *)header->data);
118         curl_headers = curl_slist_append(curl_headers, (gchar *)header->data);
119         header = g_list_next(header);
120       }
(gdb) p header_program
$5 = "X-Syslog-Program: logger\000\000\000\000\000\000\000"
(gdb) p (gchar*)header->data
$6 = (gchar *) 0x66ce80 "`\316f"
(gdb)
```
It should be `"X-Syslog-HostAHAHA: asdfasdf"` instead of  `0x66ce80 "``\316f"`

Could anyone tell me what's wrong?

Thanks in advance :)

litterbear:
It seems I found the problem, 
I should use the g_list_copy_deep () to  assign the headers.

litterbear:
I rebased the branch and passed the travis-ci's checks.
What else could I do before it is merged?

modules/http/http.c
```diff
@@ -0,0 +1,331 @@
...
+{
+  GList *header = NULL;
+  struct curl_slist *curl_headers = NULL;
+  gchar header_host[128] = {0};
+  gchar header_program[32] = {0};
+  gchar header_facility[32] = {0};
+  gchar header_level[32] = {0};
+
+  g_snprintf(header_host, sizeof(header_host),
+    "X-Syslog-Host: %s", log_msg_get_value(msg, LM_V_HOST, NULL));
```
ihrwein: FYI (don't change the code only for this): https://tools.ietf.org/html/rfc6648

litterbear: @ihrwein  Thanks for pointing it out.
I think the prefix is "X-Syslog-" instead of  just "X-" .  I think it is enough to distinguish itself from other standard parameters.   I can't come up with any scenarios that will cause problem so far.

faxm0dem:
One of the things the java dest provided is templeteable http method. Is this the case here?

falzm:
No and I think it shouldn't be: semantically speaking, `POST` is the only method that makes sense in this context.

faxm0dem:
What about `PUT`?

falzm:
`PUT` is to update an existing resource: in this context, every log message is a new resources, hence `POST`.

faxm0dem:
I agree, but you know one use-case to rule them all…

litterbear:
@faxm0dem  
The user Is always right. 
It supports put method now :)

modules/http/README.md
```diff
@@ -0,0 +1,26 @@
...
+source      s_system { system(); internal(); };
+destination http_des {
+    http(
+        url("http://127.0.0.1:8000")
+        method("PUT")
+        user_agent("syslog-ng User Agent")
+        user("user")
+        password("password")
+        headers("HEADER1: header1", "HEADER2: header2")
+        body("BODY")
```
falzm: Maybe show a templated body as example – e.g. `body("$MESSAGE")` – , to make it look more useful?

litterbear: @falzm  The README.md is updated now

bazsi:
I am merging this as it is now. I still have some white space issues, but I think it is good enough for merging, and I don't want to reopen this discussion about minor stuff.

falzm:
Great, thank you!

lbudai:
@bazsi :
master is broken : licence checker script has some problem with this module.

bkil-syslogng:
@lbudai I'll fix it in #1021 shortly

lbudai:
master is green (both in Travis and drone.io) .

PR #981, opened 2016-03-07
F/libmongo client compatibility over mongo c driver

### LINKED ISSUE #893: Deprecate old configuration parameters in MongoDB destination
Migrating some comments from https://github.com/balabit/syslog-ng/pull/891#issuecomment-178132623:

me:
```
Configuration changes: `servers`, `host`, `port`, `path`, `database`, `username`, `password`, `safe-mode` are removed.  `uri` is added, which accepts a mongodb uri as its parameter (https://docs.mongodb.org/manual/reference/connection-string/).

These changes are effective on the master branch and will be published first in syslog-ng 3.8. @fekete-robert may be also interested in this.
```

@bazsi:
```
Telling the user to migrate to the new options is fine and is the policy we applied in the past. We can even remove options if a long enough grace period is available.

The fact that 3.8 is a major release is not an excuse to break user configurations, our intent is to make upgrades (even accross major versions) as seamless as possible. 

Breaking configurations this way would upset users of that feature and rightly so.

I would consider this a critical bug.
```

bkil-syslogng:
`modules/afmongodb`, `configure.ac`: `ENABLE_LEGACY_MONGODB_OPTIONS`
This re-adds additional support for previous syntax used by libmongo-client
before we started using `mongo-c-driver` and its URI syntax exclusively.

I.e., so additionally to `uri()` and `collection()`, now the following
keywords should work again:

* `servers()`
* `safe_mode()`
* `path()`
* `password()`
* `username()`
* `database()`

Note that these are plainly translated to a connection URI without
much sanity checking or preserving their former semantic meaning.
So various aspects of the MongoDB connection like health checks, retries,
error reporting and synchronicity will still follow the slightly altered
semantics of `mongo-c-driver`.

In the future, theoretically it is enough to revert this single commit
to remove support.
(This is one reason I did not opt for OOP & inheritance to implement this)

Fixes #893

----
I will factor out the system level tests to simple ones and amend this PR with a testing commit.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -192,6 +192,11 @@ AC_ARG_WITH(mongoc,
                                          Link against the system supplied or the built-in mongo-c-driver library. (default: auto)]
               ,,with_mongoc="auto")
 
+AC_ARG_ENABLE(legacy-mongodb-options,
+              [  --enable-legacy-mongodb-options=[yes/no]
+                                         Support libmongo-client non-URI MongoDB options. (default: yes)]
+              ,,enable_legacy_mongodb_options="yes")
+
 AC_ARG_WITH(jsonc,
               [  --with-jsonc=[system/internal/auto/no]
                                          Link against the system supplied or the built-in jsonc library or explicitly disable it. (default:auto)]
@@ -979,6 +984,13 @@ dnl ***************************************************************************
 dnl mongo-c-driver headers/libraries
 dnl ***************************************************************************
 
+if test x"$enable_legacy_mongodb_options" = x"no"; then
+	enable_legacy_mongodb_options_bit=0
+else
+	enable_legacy_mongodb_options_bit=1
+fi
+AC_DEFINE_UNQUOTED(ENABLE_LEGACY_MONGODB_OPTIONS, [$enable_legacy_mongodb_options_bit], [Support libmongo-client non-URI MongoDB options])
+
 if test "x$with_mongoc" = "xauto"; then
   with_mongoc="system"
   PKG_CHECK_MODULES(LIBMONGO, libmongoc >= $LMC_MIN_VERSION,,with_mongoc="auto-internal")
@@ -1549,6 +1561,7 @@ AM_CONDITIONAL(ENABLE_JAVA_MODULES, [test "$enable_java_modules" = "yes"])
 AM_CONDITIONAL(ENABLE_MANPAGES, [test "$enable_manpages" != "no"])
 AM_CONDITIONAL(ENABLE_NATIVE, [test "$enable_native" != "no"])
 AM_CONDITIONAL(ENABLE_EXTRA_WARNINGS, [test "$enable_extra_warnings" = "yes"])
+AM_CONDITIONAL(ENABLE_LEGACY_MONGODB_OPTIONS, [test x"$enable_legacy_mongodb_options" != x"no"])
 
 AC_SUBST(timezonedir)
 AC_SUBST(pidfiledir)
@@ -1641,6 +1654,7 @@ echo "  Linux capability support    : ${has_linux_caps:=no}"
 echo "  Env wrapper support         : ${enable_env_wrapper:=no}"
 echo "  systemd support             : ${enable_systemd:=no} (unit dir: ${systemdsystemunitdir:=none})"
 echo "  systemd-journal support     : ${with_systemd_journal:=no}"
+echo "  libmongo-client options     : ${enable_legacy_mongodb_options}"
 echo " Modules:"
 echo "  Module search path          : ${module_path}"
 echo "  Sun STREAMS support (module): ${enable_sun_streams:=no}"
--- a/libtest/testutils.c
+++ b/libtest/testutils.c
@@ -138,13 +138,18 @@ assert_grabbed_messages_contain_non_fatal(const gchar *pattern, const gchar *err
   print_failure(error_message, args, "no grabbed message contains the pattern=%s", pattern);
   va_end(args);
 
-  fprintf(stderr, "  # Grabbed internal messages follow:\n");
-  for (l = internal_messages; l; l = l->next)
+  if (internal_messages)
     {
-      LogMessage *msg = (LogMessage *) l->data;
-      const gchar *msg_text = log_msg_get_value(msg, LM_V_MESSAGE, NULL);
+      fprintf(stderr, "  # Grabbed internal messages follow:\n");
+      for (l = internal_messages; l; l = l->next)
+        {
+          LogMessage *msg = (LogMessage *) l->data;
+          const gchar *msg_text = log_msg_get_value(msg, LM_V_MESSAGE, NULL);
 
-      fprintf(stderr, "  #\t%s\n", msg_text);
+          fprintf(stderr, "  #\t%s\n", msg_text);
+        }
+    } else {
+        fprintf(stderr, "  # No internal messeges grabbed!\n");
     }
 
   return FALSE;
--- a/libtest/testutils.c
+++ b/libtest/testutils.c
@@ -112,6 +112,26 @@ start_grabbing_messages(void)
 }
 
 void
+display_grabbed_messages(void)
+{
+  GList *l;
+
+  if (internal_messages)
+    {
+      fprintf(stderr, "  # Grabbed internal messages follow:\n");
+      for (l = internal_messages; l; l = l->next)
+        {
+          LogMessage *msg = (LogMessage *) l->data;
+          const gchar *msg_text = log_msg_get_value(msg, LM_V_MESSAGE, NULL);
+
+          fprintf(stderr, "  #\t%s\n", msg_text);
+        }
+    } else {
+        fprintf(stderr, "  # No internal messeges grabbed!\n");
+    }
+}
+
+void
 stop_grabbing_messages(void)
 {
   msg_set_post_func(NULL);
@@ -138,19 +158,7 @@ assert_grabbed_messages_contain_non_fatal(const gchar *pattern, const gchar *err
   print_failure(error_message, args, "no grabbed message contains the pattern=%s", pattern);
   va_end(args);
 
-  if (internal_messages)
-    {
-      fprintf(stderr, "  # Grabbed internal messages follow:\n");
-      for (l = internal_messages; l; l = l->next)
-        {
-          LogMessage *msg = (LogMessage *) l->data;
-          const gchar *msg_text = log_msg_get_value(msg, LM_V_MESSAGE, NULL);
-
-          fprintf(stderr, "  #\t%s\n", msg_text);
-        }
-    } else {
-        fprintf(stderr, "  # No internal messeges grabbed!\n");
-    }
+  display_grabbed_messages();
 
   return FALSE;
 }
--- a/modules/afmongodb/afmongodb-grammar.ym
+++ b/modules/afmongodb/afmongodb-grammar.ym
@@ -47,6 +47,12 @@
 %token KW_MONGODB
 %token KW_URI
 %token KW_COLLECTION
+%token KW_SERVERS
+%token KW_SAFE_MODE
+%token KW_PATH
+%token KW_PASSWORD
+%token KW_USERNAME
+%token KW_DATABASE
 
 %%
 
@@ -72,6 +78,7 @@ afmongodb_option
         {
             afmongodb_dd_set_collection(last_driver, $3); free($3);
         }
+    | afmongodb_legacy_option
     | value_pair_option
         {
             afmongodb_dd_set_value_pairs(last_driver, $1);
@@ -79,6 +86,59 @@ afmongodb_option
     | dest_driver_option
     | threaded_dest_driver_option
     | { last_template_options = afmongodb_dd_get_template_options(last_driver); } template_option
+    ;
+
+afmongodb_legacy_option
+    : KW_IFDEF
+        {
+#if SYSLOG_NG_ENABLE_LEGACY_MONGODB_OPTIONS
+        }
+    | KW_SERVERS '(' string_list ')'
+        {
+            CHECK_ERROR(afmongodb_dd_validate_network_combination(last_driver), @3,
+                         "Can't mix path() & servers()");
+            afmongodb_dd_set_servers(last_driver, $3);
+        }
+    | KW_HOST '(' string ')'
+        {
+            CHECK_ERROR(afmongodb_dd_validate_network_combination(last_driver), @3,
+                         "Can't mix path() & host()");
+            afmongodb_dd_set_host(last_driver, $3);
+            free($3);
+        }
+    | KW_PORT '(' LL_NUMBER ')'
+        {
+            CHECK_ERROR(afmongodb_dd_validate_network_combination(last_driver), @3,
+                         "Can't mix path() & port()");
+            afmongodb_dd_set_port(last_driver, $3);
+        }
+    | KW_PATH '(' string ')'
+        {
+            CHECK_ERROR(afmongodb_dd_validate_socket_combination(last_driver), @3,
+                          "Can't mix path() with host() or server()");
+            afmongodb_dd_set_path(last_driver, $3);
+            free($3);
+        }
+    | KW_DATABASE '(' string ')'
+        {
+            afmongodb_dd_set_database(last_driver, $3); free($3);
+        }
+    | KW_USERNAME '(' string ')'
+        {
+            afmongodb_dd_set_user(last_driver, $3); free($3);
+        }
+    | KW_PASSWORD '(' string ')'
+        {
+            afmongodb_dd_set_password(last_driver, $3); free($3);
+        }
+    | KW_SAFE_MODE '(' yesno ')'
+        {
+	        afmongodb_dd_set_safe_mode(last_driver, !!$3);
+        }
+    | KW_ENDIF
+        {
+#endif /* SYSLOG_NG_ENABLE_LEGACY_MONGODB_OPTIONS */
+        }
     ;
 
 /* INCLUDE_RULES */
--- a/modules/afmongodb/afmongodb-grammar.ym
+++ b/modules/afmongodb/afmongodb-grammar.ym
@@ -51,26 +51,35 @@
 %%
 
 start
-        : LL_CONTEXT_DESTINATION KW_MONGODB
-          {
+    : LL_CONTEXT_DESTINATION KW_MONGODB
+        {
             last_driver = *instance = afmongodb_dd_new(configuration);
-          }
-          '(' afmongodb_options ')'         { YYACCEPT; }
-	;
+        }
+        '(' afmongodb_options ')'         { YYACCEPT; }
+    ;
 
 afmongodb_options
-        : afmongodb_option afmongodb_options
-	|
-	;
+    : afmongodb_option afmongodb_options
+    |
+    ;
 
 afmongodb_option
-	: KW_URI '(' string ')'		{ afmongodb_dd_set_uri(last_driver, $3); free($3); }
-	| KW_COLLECTION '(' string ')'		{ afmongodb_dd_set_collection(last_driver, $3); free($3); }
-	| value_pair_option			{ afmongodb_dd_set_value_pairs(last_driver, $1); }
-	| dest_driver_option
-	| threaded_dest_driver_option
-	| { last_template_options = afmongodb_dd_get_template_options(last_driver); } template_option
-        ;
+    : KW_URI '(' string ')'
+        {
+            afmongodb_dd_set_uri(last_driver, $3); free($3);
+        }
+    | KW_COLLECTION '(' string ')'
+        {
+            afmongodb_dd_set_collection(last_driver, $3); free($3);
+        }
+    | value_pair_option
+        {
+            afmongodb_dd_set_value_pairs(last_driver, $1);
+        }
+    | dest_driver_option
+    | threaded_dest_driver_option
+    | { last_template_options = afmongodb_dd_get_template_options(last_driver); } template_option
+    ;
 
 /* INCLUDE_RULES */
 
--- a/modules/afmongodb/afmongodb-legacy-grammar.c
+++ b/modules/afmongodb/afmongodb-legacy-grammar.c
@@ -1,0 +1,151 @@
+/*
+ * Copyright (c) 2010-2016 Balabit
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "afmongodb-legacy-grammar.h"
+#include "afmongodb-legacy-private.h"
+
+gboolean
+afmongodb_dd_validate_socket_combination(LogDriver *d)
+{
+  MongoDBDestDriver *self = (MongoDBDestDriver *)d;
+
+  if ((self->port != 0 || self->port != MONGO_CONN_LOCAL) && self->address != NULL)
+    return FALSE;
+  if (self->servers)
+    return FALSE;
+
+  return TRUE;
+}
+
+gboolean
+afmongodb_dd_validate_network_combination(LogDriver *d)
+{
+  MongoDBDestDriver *self = (MongoDBDestDriver *)d;
+
+  if (self->port == MONGO_CONN_LOCAL && self->address != NULL)
+    return FALSE;
+
+  return TRUE;
+}
+
+void
+afmongodb_dd_set_user(LogDriver *d, const gchar *user)
+{
+  MongoDBDestDriver *self = (MongoDBDestDriver *)d;
+
+  msg_warning_once("WARNING: Using username() option is deprecated in mongodb driver,"
+      " please use uri() instead");
+
+  g_free(self->user);
+  self->user = g_strdup(user);
+  self->is_legacy = TRUE;
+}
+
+void
+afmongodb_dd_set_password(LogDriver *d, const gchar *password)
+{
+  MongoDBDestDriver *self = (MongoDBDestDriver *)d;
+
+  msg_warning_once("WARNING: Using password() option is deprecated in mongodb driver,"
+      " please use uri() instead");
+
+  g_free(self->password);
+  self->password = g_strdup(password);
+  self->is_legacy = TRUE;
+}
+
+void
+afmongodb_dd_set_host(LogDriver *d, const gchar *host)
+{
+  MongoDBDestDriver *self = (MongoDBDestDriver *)d;
+
+  msg_warning_once(
+      "WARNING: Using host() option is deprecated in mongodb driver, please use uri() instead");
+
+  g_free(self->address);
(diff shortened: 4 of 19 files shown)


bkil-syslogng:
I've neatly squashed all patches and rebased to master. It's ready for reviewing.

As we have raised in #893, `safe_mode` simulation will not try to enforce an extra client side document validation as libmongo-client did, but it will otherwise influence write concerns.

bkil-syslogng:
Note that because of the excessive rebasing, it's easier to go through in commit order (via gitk and the like). GitHub lists them pretty randomly.

bkil-syslogng:
@lbudai Rebased to master.

lbudai:
@bkil-syslogng : Thanks!
The Plan is that we will review it on tomorrow morning, and if everything is fine, we will merge it.

modules/afmongodb/afmongodb-legacy-grammar.c
```diff
@@ -0,0 +1,147 @@
...
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "afmongodb-legacy-grammar.h"
+#include "afmongodb-legacy-private.h"
+
+gboolean
+afmongodb_dd_check_address(LogDriver *d, gboolean local)
```
MrAnno: We should split this function to eliminate the bool parameter. This can violate the Single Responsibility Principle.

modules/afmongodb/afmongodb-legacy-uri.c
```diff
@@ -0,0 +1,289 @@
...
+    return FALSE;
+
+  if (self->uri_str && self->is_legacy)
+    {
+      msg_error("Error: either specify a MongoDB URI (and optional collection) or only legacy options",
+                evt_tag_str("driver", self->super.super.super.id),
+                NULL);
+      return FALSE;
+    }
+  else if (self->is_legacy)
```
lbudai: Maybe this block could be factored out to a private `_build_uri_from_legacy_options(MongoDBDestDriver *self)`.
The `else` branch is not necessary as we return from the `if`.

The final result could be something similar then:

```C
gboolean
afmongodb_dd_create_uri_from_legacy(MongoDBDestDriver *self)
{
  if (!_check_auth_options(self))
    return FALSE;

  if (_is_legacy_options_mixed_with_uri)
    return FALSE;

  return _build_uri_from_legacy_options(self);
}
```

modules/afmongodb/afmongodb.c
```diff
@@ -372,6 +372,10 @@ afmongodb_dd_private_uri_init(LogDriver *d)
 {
   MongoDBDestDriver *self = (MongoDBDestDriver *)d;
 
+  if (!self->uri_str)
+    self->uri_str = g_string_new("mongodb://127.0.0.1:27017/syslog"
+        "?wtimeoutMS=60000&socketTimeoutMS=60000&connectTimeoutMS=60000");
```
lbudai: We could initiate a define.

lbudai:
@bkil-syslogng : thanks for this PR!

modules/afmongodb/afmongodb-legacy-grammar.c
bkil-syslogng: Do you have a recommendation? Would the following pair do it?
* `afmongodb_dd_validate_socket_combination`
* `afmongodb_dd_validate_network_combination`

bkil-syslogng:
Rebased to master, fixups squashed.

The separate fixups in which I've addressed the review questions can be viewed here:
* split afmongodb_dd_check_address() https://github.com/balabit/syslog-ng/commit/232fa27d344fafefcfce7048286baf76af7df3d5
* DEFAULT_URI https://github.com/balabit/syslog-ng/commit/681d1c557612688b244bbeb9a0ba56a9500d3958
* extract _build_uri_from_legacy_options() https://github.com/balabit/syslog-ng/commit/18e387bcd2f41a311fc1e79be1dd3a2e90687269

Full backups saved:
* https://github.com/balabit/syslog-ng/compare/master...bkil-syslogng:t/libmongo-client-compatibility-over-mongo-c-driver-bk15?expand=1
* https://github.com/balabit/syslog-ng/compare/master...bkil-syslogng:t/libmongo-client-compatibility-over-mongo-c-driver-bk14?expand=1
* https://github.com/balabit/syslog-ng/compare/master...bkil-syslogng:t/libmongo-client-compatibility-over-mongo-c-driver-bk13?expand=1

As a side note, could this have caused excessive rebuilding in the past?
https://github.com/bkil-syslogng/syslog-ng/commit/9d9ba49459f035a822b15f46b1d569628575c288

bazsi:
Ive checked out the makefile patch but I don't see how we were compiling
successfully if the path was incorrect until now

bkil-syslogng:
@bazsi Here is the error that was output when I did not reapply this patch: https://travis-ci.org/balabit/syslog-ng/jobs/122841898

bkil-syslogng:
@MrAnno raised whether we should use a single `#if` in the grammar instead of multiple ones.

Here's the patch:
https://github.com/balabit/syslog-ng/commit/2ee211a50454f41a041a83b2161861a23b941bbe

Fixup squashed, rebased to master.

Backup saved:
* https://github.com/balabit/syslog-ng/compare/master...bkil-syslogng:t/libmongo-client-compatibility-over-mongo-c-driver-bk16?expand=1

bkil-syslogng:
@bazsi Probably the dummy rule didn't complain, but the new unit test rule now complains.

MrAnno:
:+1:

bkil-syslogng:
Rebased to master, no content change has been done since bk16.

(No new backup was created because I forgot to squash the fixup last time, so bk16 was the one @MrAnno blessed).

bkil-syslogng:
Fixups squashed, rebased to master.

Now it should pass our new quality assurance suite without errors. You can review the QA related fixups on top of the following backup if you are interested. Some of the fixups are miscellaneous improvements.

* https://github.com/balabit/syslog-ng/compare/master...bkil-syslogng:t/libmongo-client-compatibility-over-mongo-c-driver-bk17?expand=1

bkil-syslogng:
We are waiting for the `:+1:` of @bazsi, as the principal one who has raised the issue of compatibility. Is the kind of change you had in mind?

You did nod in person to acknowledge, but could you please at least review from a high level the kind of test cases that cover mongoDB and the backwards compatibility? The relevant part is about a page long of easy to read code - ideally it shouldn't take more than 5 minutes at this level.

bazsi:
For me the functional compatibility was the important stuff, and you said it was there. And since I didn't know when I could review it, I said you should merge it and not wait for me.

I still hold that it is sometimes more important to progress  with tradeoffs, but since I have a few minutes, I can do a review now.

So here it comes.

configure.ac
```diff
@@ -189,6 +189,11 @@ AC_ARG_WITH(mongoc,
                                          Link against the system supplied or the built-in mongo-c-driver library. (default: auto)]
               ,,with_mongoc="auto")
 
+AC_ARG_ENABLE(legacy-mongodb-options,
+              [  --enable-legacy-mongodb-options=[yes/no]
+                                         Support libmongo-client non-URI MongoDB options. (default: yes)]
+              ,,enable_legacy_mongodb_options="yes")
+
 AC_ARG_WITH(jsonc,
```
bazsi: I am not sure we need this configure option. I understand that we might want to document what portions of the code are related to the compatibility layer, but I don't think we want to make this user controllable.

More options more confusions, and more combinations we need to handle.

libtest/testutils.c
```diff
@@ -138,14 +158,7 @@ assert_grabbed_messages_contain_non_fatal(const gchar *pattern, const gchar *err
...
 
-  fprintf(stderr, "  # Grabbed internal messages follow:\n");
-  for (l = internal_messages; l; l = l->next)
-    {
-      LogMessage *msg = (LogMessage *) l->data;
-      const gchar *msg_text = log_msg_get_value(msg, LM_V_MESSAGE, NULL);
-
-      fprintf(stderr, "  #\t%s\n", msg_text);
-    }
+  display_grabbed_messages();
```
bazsi: Nice one.

modules/afmongodb/afmongodb-grammar.ym
```diff
@@ -66,12 +72,60 @@ afmongodb_options
...
+afmongodb_legacy_option
+    : KW_IFDEF {
+#if SYSLOG_NG_ENABLE_LEGACY_MONGODB_OPTIONS
+}
+	| KW_SERVERS '(' string_list ')'
+          {
+            CHECK_ERROR(afmongodb_dd_validate_network_combination(last_driver), @3, "Can't mix path() & servers()");
+            afmongodb_dd_set_servers(last_driver, $3);
+          }
+    | KW_HOST '(' string ')'
```
bazsi: Nitpick: the indentation doesn't look right here

modules/afmongodb/afmongodb-legacy-grammar.c
```diff
@@ -0,0 +1,151 @@
...
+afmongodb_dd_set_path(LogDriver *d, const gchar *path)
+{
+  MongoDBDestDriver *self = (MongoDBDestDriver *)d;
+
+  msg_warning_once(
+      "WARNING: Using path() option is deprecated in mongodb driver, please use uri() instead");
+
+  g_free(self->address);
+  self->address = g_strdup(path);
+  self->port = MONGO_CONN_LOCAL;
```
bazsi: Is it a good idea to use address for both the UNIX domain path and the name of the server? It doesn't make it very easy to decipher validate_xxx_combination above.

modules/afmongodb/afmongodb-legacy-uri.c
```diff
@@ -0,0 +1,301 @@
...
+_parse_addr(const char *str, char **host, gint *port)
+{
+  if (!host || !port)
+    {
+      msg_debug("Host or port reference should not be NULL");
+      return FALSE;
+    }
+  char *proto_str = g_strdup_printf("mongodb://%s", str);
+  mongoc_uri_t *uri = mongoc_uri_new(proto_str);
+  if (!uri)
```
bazsi: I know that you prefer declaring variables inline, one feedback: long type names with qualifiers clutter up lines, making them more difficult to read.

I know why it is a good idea, local declarations and ask that stuff, but I really had to scroll sideways, and not once during this review.

If only we had type inference, we could use a short statement (like auto as in c++) but without that Iam not sure anymore that it is a good idea.

This case is not the worst offender, the append legacy host function is a lot worse.

modules/afmongodb/afmongodb-legacy-uri.c
```diff
@@ -0,0 +1,301 @@
...
+  g_free(hp);
+}
+
+static gboolean
+_append_host(GList **list, const char *host, gint port)
+{
+  if (!list)
+    return FALSE;
+  MongoDBHostPort *hp = g_new0(MongoDBHostPort, 1);
+  hp->host = g_strdup(host);
```
bazsi: If we have a dedicated struct for host and port pairs, we might want to extract them to a separate constructor function. I am not sure we need mongodb legacy private.h, I would just merge it in if it didn't deserve a constructor either.

modules/afmongodb/afmongodb-legacy-uri.c
```diff
@@ -0,0 +1,301 @@
...
+              return FALSE;
+            }
+          g_free(host);
+
+          port = self->port ? self->port : MONGOC_DEFAULT_PORT;
+          const gchar *address = self->address ? self->address : DEFAULTHOST;
+          gchar *srv = g_strdup_printf("%s:%d", address, port);
+          self->servers = g_list_prepend(self->servers, srv);
+          g_free(self->address);
+          self->address = NULL;
```
bazsi: It bothers me a little that we patch address into servers in a function that should only use the configured data as read only.
This append legacy servers looks a bit hairy

modules/afmongodb/afmongodb-legacy-uri.c
```diff
@@ -0,0 +1,301 @@
...
+        }
+      else
+        {
+          gchar *localhost = g_strdup_printf(DEFAULTHOST ":%d", MONGOC_DEFAULT_PORT);
+          self->servers = g_list_append(NULL, localhost);
+          _append_host(&self->recovery_cache, DEFAULTHOST, MONGOC_DEFAULT_PORT);
+        }
+
+      if (!_parse_addr(g_list_nth_data(self->servers, 0), &self->address, &self->port))
+        {
```
bazsi: So we insert self->address into self->servers and then extract it again?

modules/afmongodb/afmongodb-legacy-uri.c
```diff
@@ -0,0 +1,301 @@
...
+
+static gboolean
+_check_auth_options(MongoDBDestDriver *self)
+{
+  if (self->user || self->password)
+    {
+      if (!self->user || !self->password)
+        {
+          msg_error("Neither the username, nor the password can be empty");
+          return FALSE;
```
bazsi: We check for being null not for being empty. Are those equivalent?

modules/afmongodb/afmongodb-legacy-uri.c
```diff
@@ -0,0 +1,301 @@
...
+
+gboolean
+_build_uri_from_legacy_options(MongoDBDestDriver *self)
+{
+    if (!_append_legacy_servers(self))
+      return FALSE;
+
+    self->uri_str = g_string_new("mongodb://");
+    if (!self->uri_str)
+      return FALSE;
```
bazsi: g_string_new can't be NULL. No need to check for this explicitly

modules/afmongodb/afmongodb-legacy-uri.c
```diff
@@ -0,0 +1,301 @@
...
+
+    if (self->user && self->password)
+      {
+        g_string_append_printf(self->uri_str, "%s:%s@", self->user, self->password);
+      }
+
```
bazsi: Escaping is missing, is there a threat vector here?

modules/afmongodb/afmongodb-legacy-uri.c
```diff
@@ -0,0 +1,301 @@
...
+                             "?wtimeoutMS=%d",
+                             SOCKET_TIMEOUT_FOR_MONGO_CONNECTION_IN_MILLISECS);
+    else
+      g_string_append(self->uri_str, "?w=0&safe=false");
+
+    g_string_append_printf(self->uri_str,
+                           "&socketTimeoutMS=%d&connectTimeoutMS=%d",
+                           SOCKET_TIMEOUT_FOR_MONGO_CONNECTION_IN_MILLISECS,
+                           SOCKET_TIMEOUT_FOR_MONGO_CONNECTION_IN_MILLISECS);
+
```
bazsi: Looking at these options, how they get encoded into an URL, I am not sure it is a good idea to keep the uri the preferred interface for mongo options.

How is it accomplished in other applications?

modules/afmongodb/afmongodb-private.h
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#ifndef AFMONGODB_PRIVATE_H_
+#define AFMONGODB_PRIVATE_H_
+
```
bazsi: This should start with a syslog-ng.h include.

modules/afmongodb/afmongodb-private.h
```diff
@@ -0,0 +1,72 @@
...
+#if SYSLOG_NG_ENABLE_LEGACY_MONGODB_OPTIONS
+  GList *servers;
+  gchar *address;
+  gint port;
+
+  gboolean safe_mode;
+  gchar *user;
+  gchar *password;
+#endif
+
```
bazsi: If we already have a separate layer to format the URL, we might as well really use a separate struct that is embedded here.

modules/afmongodb/afmongodb.c
```diff
@@ -86,8 +60,9 @@ afmongodb_dd_set_uri(LogDriver *d, const gchar *uri)
 {
   MongoDBDestDriver *self = (MongoDBDestDriver *)d;
 
-  g_free(self->uri);
-  self->uri = g_strdup(uri);
+  if (self->uri_str)
+    g_string_free(self->uri_str, TRUE);
+  self->uri_str = g_string_new(uri);
```
bazsi: g_string_assign?

modules/afmongodb/afmongodb.c
```diff
@@ -370,83 +394,128 @@ afmongodb_worker_insert (LogThrDestDriver *s, LogMessage *msg)
...
+  MongoDBDestDriver *self = (MongoDBDestDriver *)d;
 
-  return WORKER_INSERT_RESULT_SUCCESS;
+#if SYSLOG_NG_ENABLE_LEGACY_MONGODB_OPTIONS
+  if (!afmongodb_dd_create_uri_from_legacy(self))
+    return FALSE;
+#endif
+
+  if (!self->uri_str)
+    self->uri_str = g_string_new(DEFAULT_URI);
```
bazsi: What about using the setter?

modules/afmongodb/afmongodb.c
```diff
@@ -370,83 +394,128 @@ afmongodb_worker_insert (LogThrDestDriver *s, LogMessage *msg)
...
-  if (!success && (errno == ENOTCONN))
-    return WORKER_INSERT_RESULT_NOT_CONNECTED;
+  return WORKER_INSERT_RESULT_SUCCESS;
+}
 
-  if (!success)
-    return WORKER_INSERT_RESULT_ERROR;
+gboolean
+afmongodb_dd_private_uri_init(LogDriver *d)
+{
```
bazsi: I don't understand the name, I mean the private qualifier. Why is it private?

modules/afmongodb/afmongodb.c
```diff
@@ -142,26 +146,53 @@ afmongodb_dd_disconnect(LogThrDestDriver *s)
 }
 
 static gboolean
-afmongodb_dd_connect(MongoDBDestDriver *self, gboolean reconnect)
+_dd_connect(MongoDBDestDriver *self, gboolean reconnect)
 {
```
bazsi: I like the shortened names, that's really great.

The DD part of the name here is probably not needed though. That refers to destination driver, and none of the other functions retained those in the name, and is probably not needed here either

modules/afmongodb/afmongodb.c
```diff
@@ -370,83 +394,128 @@ afmongodb_worker_insert (LogThrDestDriver *s, LogMessage *msg)
...
+  /* Always replace a leading dot with an underscore. */
   vpts = value_pairs_transform_set_new(".*");
   value_pairs_transform_set_add_func(vpts, value_pairs_new_transform_replace_prefix(".", "_"));
   value_pairs_add_transforms(self->vp, vpts);
 }
 
 static gboolean
-afmongodb_dd_init(LogPipe *s)
+_logpipe_init(LogPipe *s)
 {
```
bazsi: I wouldn't include class names in virtual methods like logpipe in this case. I understand that init without a qualifier might be too general as a term though. 

In other destinations we either used the unqualified _init as a name or we kept the original fully qualified name. I think both are OK, but I wouldn't start a new convention as you did here

modules/afmongodb/afmongodb.c
```diff
@@ -519,39 +576,42 @@ afmongodb_dd_new(GlobalConfig *cfg)
...
-{
-  .type = LL_CONTEXT_DESTINATION,
-  .name = "mongodb",
-  .parser = &afmongodb_parser,
-};
+      {
+          .type = LL_CONTEXT_DESTINATION,
+          .name = "mongodb",
+          .parser = &afmongodb_parser,
+      };
```
bazsi: Indentation?

modules/afmongodb/tests/test-mongodb-config.c
```diff
@@ -0,0 +1,264 @@
...
+  _expect_uri_in_log("safe_mode_false", "127.0.0.1:27017/syslog" UNSAFEOPTS, "syslog", "messages");
+
+  afmongodb_dd_set_user(mongodb, "user");
+  afmongodb_dd_set_password(mongodb, "password");
+  _expect_uri_in_log("user_password", "user:password@127.0.0.1:27017/syslog" SAFEOPTS, "syslog", "messages");
+
+  afmongodb_dd_set_collection(mongodb, "messages2");
+  afmongodb_dd_set_safe_mode(mongodb, FALSE);
+  _expect_uri_in_log("collection_safe_mode", "127.0.0.1:27017/syslog" UNSAFEOPTS, "syslog", "messages2");
+}
```
bazsi: If the legacy stuff would have its own module we probably would not have to initialize an entire destination for tests.

bazsi:
So I had a number of comments, some of them nitpicks other more material.

I stand by my opinion that this could go in, especially if we are in a hurry and this one prevents the release from happening. Anyway now as I had some material feedback as well I'd like those to be addressed but it's really up to you.

bkil-syslogng:
@bazsi

Thanks for the thorough review!

Some parts of the additions you see here are actually undoing the removal of legacy codes. I've created a temporary internal PR for you so you can see a diff with the new codes. I've addressed your other questions below and in fixups.

https://github.com/bkil-syslogng/syslog-ng/pull/24/files

I feel that I have obeyed the boy scout rule (i.e., you should leave your site cleaner than it was before you arrived) to a reasonable level. When you touch code (or just move it around in my case), you should not be expected to completely rewrite it or do deep refactoring on it, unless your whole changeset itself is related to deep refactoring of related code. My aim was to interface with the old codes, not replace them.

Actually, some of your earlier review notes also concerned legacy codes, but I did not systematically point that out until now and simply fixed up what you asked for.

> I am not sure we need this configure option. I understand that we might want to document what portions of the code are related to the compatibility layer, but I don't think we want to make this user controllable.
>
> More options more confusions, and more combinations we need to handle.

It is an advantage if either a CI, or a much more rarely executed automation can test both states. Two alternatives would be an undocumented environmental variable or code patching - neither of which is pretty.

I would definitely want to quarantine all relevant code via conscious executable documentation. We must remove all of these as soon as feasible. In general, we must have a clear strategy to regularly remove code, because complexity and cruft is only accumulating with time. We must make removal of code in the future excessively easy and desirable, so it must not take discussions and deep investigations to commence. It should also be ideally executable by anyone without any significant prior preparation.

> [modules/afmongodb/afmongodb-grammar.ym]
> Nitpick: the indentation doesn't look right here

Legacy code & last minute refactorings - the file has/had mixed tabs & spaces. I've changed everything to spaces now.

> Is it a good idea to use address for both the UNIX domain path and the name of the server? It doesn't make it very easy to decipher validate_xxx_combination above.

Legacy code.

> I know that you prefer declaring variables inline, one feedback: long type names with qualifiers clutter up lines, making them more difficult to read.
> [...]

My personal aim is to keep lines below about 100 (-110) columns to accommodate pivoting screens, but our guidelines allow for up to 110-120 columns which most of us regularly exceed. GitHub does not mandate scrolling before 120 columns for me in full screen. I could only find a single line beyond 116 columns (128) and it was in `afmongodb-grammar.ym`. However, on your request, I have wrapped all lines to stay under 100 columns. Also, `append_legacy_servers()` is mostly legacy code, as noted below.

For the inline declaration part, it is much cleaner this way in most of the cases in the following aspects:
* being able to better limit scope;
* ensuring proper initialization;
* not having to look for uninitialized variables (`NULL` pointers);
* do away with trivial redundancy of the variable name on initialization;
* seeing the type near usage, not some random distance away.

I still regularly declare separately at the very beginning in specific reasonable cases:
* when the containing block or function is not trivial in length and when we must be cautious about deinitialization at the end;
* when the type and/or the initializer is too complex by itself or has side effects;
* when multiple, logically connected variables need to be initialized and some can not be initialized with others.

> If we have a dedicated struct for host and port pairs, we might want to extract them to a separate constructor function

I've refactored it into `host-list.[ch]`.

> I am not sure we need mongodb legacy private.h, I would just merge it in if it didn't deserve a constructor either.

After refactoring the host-port list, now only a single DRY constant is present there, shared between afmongodb-legacy-grammar and afmongodb-legacy-uri.

> It bothers me a little that we patch address into servers in a function that should only use the configured data as read only.
> This append legacy servers looks a bit hairy
> [...]
> So we insert self->address into self->servers and then extract it again?

Legacy code. Although, I was forced to fix a bug or two in it after extending test coverage.

> We check for being null not for being empty. Are those equivalent?

Legacy code. I have already given thought to such edge case earlier, and it should translate to `:@localhost` which should be reject with an error. Although, I have added test cases for this to assert on the error message.

> g_string_new can't be NULL. No need to check for this explicitly

OK

> Escaping is missing, is there a threat vector here?

They consider it a threat if someone other than the application administrator can influence the behavior of the application or manipulate data of other users. If you could craft a `mongodb()` config snippet at runtime then you could alter where the crafted mongodb destination URI points to using some of these fields, but then again that could be done using directly using the URI.

Could you tell me some specifics about what you had in mind? We could easily implement escaping if you think it carries an advantage.

> Looking at these options, how they get encoded into an URL, I am not sure it is a good idea to keep the uri the preferred interface for mongo options.
>
> How is it accomplished in other applications?

* https://stackoverflow.com/questions/19697180/how-to-use-mongo-uri-in-mongolab-and-heroku-server
* https://www.elastic.co/guide/en/logstash/current/plugins-outputs-mongodb.html
* https://www.mongodb.com/blog/post/using-mongodb-hadoop-spark-part-3-spark-example-key-takeaways

> This should start with a syslog-ng.h include.

OK. (I based on legacy code again, but I've now fixed those as well)

> If we already have a separate layer to format the URL, we might as well really use a separate struct that is embedded here.

Legacy code. I feel that to be over-engineering and can't see the benefit here.

> g_string_assign?

OK, one allocation less on init.

> What about using the setter?

It is not applicable here by itself, the `if` must stay. However, I'd view the below solution more confusing than 'abstract':

```
  if (!self->uri_str)
    afmongodb_dd_set_uri(self, DEFAULT_URI);
```

> I don't understand the name, I mean the private qualifier. Why is it private?

Because it has only been exported for testing reasons. Alternatives seems kludgey, like exporting a `test_private()` unit test function, or including sources in other sources.

> I like the shortened names, that's really great.

> The DD part of the name here is probably not needed though. That refers to destination driver, and none of the other functions retained those in the name, and is probably not needed here either

OK

> I wouldn't include class names in virtual methods like logpipe in this case. I understand that init without a qualifier might be too general as a term though.

> In other destinations we either used the unqualified _init as a name or we kept the original fully qualified name. I think both are OK, but I wouldn't start a new convention as you did here

OK. I will stick with a single underscore variant, though.

> [struct] Indentation?

OK, I will revert this. (This is how Eclipse formatted this file)

> If the legacy stuff would have its own module we probably would not have to initialize an entire destination for tests.

However, we would still need to test both the non-legacy parts and the integration related one similar to this. Simply calling the same infrastructure on the compatibility cases like this is free and also favored by KISS. We need to draw a line somewhere. As explained in the last commit message, my aim was in spirit of YAGNI to not over-engineer something that by design has neither future, nor reuse value. Turning it into OOP has all the drawbacks without none of the benefits.

I feel that adding extra layers back and forth between the grammar, the driver and the legacy would make the whole solution much longer and harder to understand for others as well. I tried to stay as close to the original code as feasible in service of those who already know the previous codes and to introduce as few new bugs as possible.

bazsi:
Hi,

Thanks for considering my suggestions. I agree that we don't want to over
engineer this, but as this code was previously removed it was not clear to
me in the context of the patch what was original code and what is new,
especially as the legacy code was factored out to a separate module.

Anyway, nice work, let's merge this.

bazsi:
Is this ready for merging? some squashing seems to be needed yet

bkil-syslogng:
Squashing is needed indeed, and I've just now started adding some persist
tests, now that we have a specification. Shortlisted.

bkil-syslogng:
Timely rebased to master. Fixups squashed. Persist/stats name tests added:
https://github.com/bkil-syslogng/syslog-ng/commit/129573f5f3389855c5d55887a8188d06e9d530c3
https://github.com/bkil-syslogng/syslog-ng/blob/f/libmongo-client-compatibility-over-mongo-c-driver/modules/afmongodb/tests/test-mongodb-config.c#L137

lbudai:
So can we say that this is the final version that is ready for a final-review?

bkil-syslogng:
Yes. I've timely rebased to master now and also factored out a global
variable from the test.

PR #982, opened 2016-03-08
modules/python: fix `make func-test` and run `make install` in Travis

bkil-syslogng:
I'm expecting that both the Python func-test and the Java one will pass in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ git:
 env: B=autotools
 
 install:
+  - export SLNG_CACHE="$HOME/.local";
+    export SLNG="$HOME/install/syslog-ng";
+    export PATH="$SLNG_CACHE"/bin:$PATH;
+    export PKG_CONFIG_PATH=$SLNG_CACHE/lib/pkgconfig:$SLNG/lib/pkgconfig:$PKG_CONFIG_PATH;
   - curl http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_12.04/Release.key | sudo apt-key add -
   - echo "deb http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_12.04 ./" | sudo tee --append /etc/apt/sources.list.d/syslog-ng-obs.list
   - sudo apt-get update -qq
@@ -31,6 +35,7 @@ install:
       sqlite3
       xsltproc
   - sudo pip install -r requirements.txt
+  - ./scripts/build-dbi.sh || exit $?
 before_script:
   - ./autogen.sh
   - unset PYTHON_CFLAGS # HACK
@@ -47,7 +52,6 @@ before_script:
       --enable-linux-caps
       --disable-sun-streams
       --enable-all-modules
-      --disable-sql
       --enable-pacct
       --enable-manpages
       --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl
--- a/scripts/build-dbi.sh
+++ b/scripts/build-dbi.sh
@@ -1,0 +1,89 @@
+#!/bin/sh -x
+#############################################################################
+# Copyright (c) 2016 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -5,15 +5,47 @@ install:
...
+      --disable-systemd
+      --enable-geoip
+      --enable-riemann
+      --enable-systemd-journal=optional
+      --enable-python
+      --enable-manpages
+      --enable-java
+      --enable-java-modules
+      --enable-native
+      --enable-all-modules
```
lbudai: Why do you need these explicit enables?

configure.ac
```diff
@@ -686,9 +686,14 @@ dnl ***************************************************************************
 dnl libdbi headers/libraries
 dnl ***************************************************************************
 if test "x$enable_sql" = "xyes" || test "x$enable_sql" = "xauto"; then
-	PKG_CHECK_MODULES(LIBDBI, dbi >= $LIBDBI_MIN_VERSION,, with_libdbi="")
+	PKG_CHECK_MODULES(LIBDBI, dbi >= $LIBDBI_MIN_VERSION, with_libdbi="1", with_libdbi="0")
+	if test "$with_libdbi" -eq 0; then
+		dnl if libdbi has no .pc file (e.g., Ubuntu Precise), try it without one
+		AC_CHECK_LIB(dbi, dbi_initialize, LIBDBI_LIBS="-ldbi"; LIBDBI_CFLAGS="-I/usr/include/dbi"; with_libdbi="1")
+	fi
```
lbudai: Could you write a more detailed commit message?

lbudai:
Reverting the libdbi version upgrade patches is not the best, IMHO( as I remember we talked about this : your reason was that in travis we are using the default Ubuntu Precise based environment and because we did not test the afsql driver, we did not know that moving to 0.9.0 libdbi caused a problem: that is not available on Precise). We could provide a newer libdbi package and we could install it in the travis environment (as we do it in case of gradle-2.2.1).

bkil-syslogng:
Yes, and we've also talked about this with Bazsi and others as well. After
I've done some bisecting and reviewing, I've decided that we should
probably stay with a newer libdbi version. I'll probably install it from
source, though I haven't checked whether it is present in a whitelisted PPA
already.

This is one thing I need to fix up, and this is why this PR is in
in-progress status, and not needs-review.

.travis.yml
bkil-syslogng: We need to make explicit that we test all (possible) modules, and this also
documents which ones are untested. The script needs to consider this from a
black box perspective, and it must assume that unspecified modules will be
set to auto by the configure script.

lbudai:
Hi,

even it is not in the need-review state, I can check it :)
( It should not be a problem. )
And I see two issues here: libdbi and the forced enables vs. new module issue.

bazsi:
I wouldn't want to add a 3rd way of compiling dependencies, like you added with the build-dbi.sh script. We now either have it as a system dependency or we ship it as a part of the source tarball. This seems to be a 3rd, more or less travis specific option.

I would prefer that we either:
1) build libdbi using OBS and use that package in travis
2) embed libdbi as a submodule

And my strong preference would be the first option.

bkil-syslogng:
We shall remove the OBS specific route in the future, because there's no
sudo in Container Travis and using that shaves off 1 minute of each build.

I still want to do a quick lookup whether a newer dbi would be available
from an APT addon repo, though. So this is also shortlisted.

bkil-syslogng:
Although there's also rationale in shipping dependencies which are not
common in supported distributions. Note that I also needed to compile
the Riemann connector and systemd the same way to fully support Precise,
see below. At the same time, we should start to remove ones which are
universally available.

https://github.com/bkil-syslogng/syslog-ng/pull/23

bkil-syslogng:
@bazsi 
I've extracted all improvements other than the script to this other one, so it could get merged:
https://github.com/balabit/syslog-ng/pull/1070

bkil-syslogng:
Here is a new compromise that you will probably like regarding handling dependencies.

We should move primarily to Trusty VM (sudo) Travis. If some of the dependencies are not available via the official repo, we can use our own repo. It should be the one which runs distcheck and funct-test. Extra warnings should be enable only on Trusty as well due to its newer gcc.

On Precise, extra warnings and checks should be disabled. We could potentially consider dropping extra dependencies here, and move to container Travis, but this should be a separate phase after the above changes went through.

kira-syslogng:

presidento:
Outdated.

PR #984, opened 2016-03-09
F/file perms extract from cfg

bazsi:
This series extracts a few members from GlobalConfig and uses the FilePermOptions structure instead.

Again a step in decoupling GlobalConfig dependency from various parts of syslog-ng.

Since testing is lacking in the area, I split up the series into really small baby steps that should be possible to validate by reviewing it.

Let's see what travis thinks about it :)

### DIFF
--- a/lib/file-perms.h
+++ b/lib/file-perms.h
@@ -38,11 +38,17 @@ typedef struct _FilePermOptions
 } FilePermOptions;
 
 void file_perm_options_set_file_uid(FilePermOptions *s, const gchar *file_uid);
+void file_perm_options_dont_change_file_uid(FilePermOptions *s);
 void file_perm_options_set_file_gid(FilePermOptions *s, const gchar *file_gid);
+void file_perm_options_dont_change_file_gid(FilePermOptions *s);
 void file_perm_options_set_file_perm(FilePermOptions *s, gint file_perm);
+void file_perm_options_dont_change_file_perm(FilePermOptions *s);
 void file_perm_options_set_dir_uid(FilePermOptions *s, const gchar *dir_uid);
+void file_perm_options_dont_change_dir_uid(FilePermOptions *s);
 void file_perm_options_set_dir_gid(FilePermOptions *s, const gchar *dir_gid);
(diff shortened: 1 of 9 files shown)


bazsi:
Travis seems to have acked it, adding the needs-review label. I have some further work queued, which depends on this one, so a review would be appreciated.

lbudai:
Hi,

`DONTCHANGE`(-2) and `dont_touch`(-1) are a little bit confusing as in the commit message when you initiate the `DONTCHANGE` macro, you call it as a "dont-touch" mode... 
[ can't we use the `DONTCHANGE` instead of `-1` in `dont_touch` ?  ]

bazsi:
Sure, we can, and its better that way.

lib/file-perms.h
```diff
@@ -45,7 +45,8 @@ void file_perm_options_set_dir_gid(FilePermOptions *s, const gchar *dir_gid);
 void file_perm_options_set_dir_perm(FilePermOptions *s, gint dir_perm);
 
 void file_perm_options_defaults(FilePermOptions *self);
-void file_perm_options_init(FilePermOptions *self, GlobalConfig *cfg);
+void file_perm_options_global_defaults(FilePermOptions *self);
+void file_perm_options_init(FilePermOptions *self, FilePermOptions *global_options);
```
bkil-syslogng: I would add a const here for the argument and at some other places as well.

bkil-syslogng:
If I want to nitpick, I'd generally keep the first line of each commit message at most 72 characters long, otherwise GitHub will break it intrusively. gitk still shows it nicely, but anyway I just wanted to let you know for consideration. 

Most projects follow a 50/72 commit message rule, but they generally recommend hard breaking on 72 in any case, so I don't think that 72/72 is an unreasonable rule to follow.

bkil-syslogng:
Other than the message subject improvement and the recommendation that we should strive to use (struct) constness for new codes, I see no issue with this PR. :+1:

bazsi:
I think I have now addressed all review comments, clarified commit message, so this should be ready for merging, once travis acks that.

lbudai:
Hi,

thanks!
Sorry for late reaction.

PR #985, opened 2016-03-10
F/missing py license (Python, Perl and Bash scripts were missing declarations)

bkil-syslogng:
Also enabled automatic checking in the future for the most important non-C kinds (e.g., Java).


modules/python/pylib/syslogng/__init__.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+#############################################################################
```
bazsi: I don't think we need a shebang line here and neither of other modules in the same package.

tests/functional/control.py
```diff
@@ -1,3 +1,25 @@
+#!/usr/bin/env python
+#############################################################################
```
bazsi: no shebang needed either.

bazsi:
Apart from my nitpicks wrt shebang lines, this looks great, thank you.

modules/python/pylib/syslogng/__init__.py
bkil-syslogng: @bazsi 
1. Some of the Python files contained this, some of them did not. I wanted to make it uniform.
2. I recommend using a shebang in each and every script to make file magic consistent on all platforms.

In light of these, do you agree?

bazsi: executable scripts contained this, whereas modules to be imported didn't.
having a shebang line indicates that the file would be executable, which
they are not.

-- 
Bazsi

tests/functional/control.py
bkil-syslogng: @bazsi And let me also add a nitpick of my own. I would generally recommend setting the executable bit on all of the scripts which could be meaningfully executed from the command line.

bkil-syslogng:
@bazsi I've added two new commits to remove redundant shebangs. Please notify me after you have reviewed them, so I can squash the fixup.

bazsi:
ok on my part, thanks for the fixups.

tests/copyright/policy
```diff
@@ -22,6 +22,6 @@ modules/native
  GPLv2+_SSL
 tests
 modules/java/(tools|[^/]*$)
-modules/java-modules/(dummy|elastic|hdfs|http|kafka|[^/]*$)
+modules/java-modules/(dummy|elastic|elastic-v2|hdfs|http|kafka|[^/]*$)
```
lbudai: Hmm... we should add every new module to this file?
That will always be a problem...

lbudai:
@bkil-syslogng : could you squash the patches?

tests/copyright/policy
bkil-syslogng: Yes, we do need to add each one when using the current format. This is because we have both GPL and non-GPL modules in the same folder, and I don't want to add overlapping patterns and row based prioritization, because that can bite us very quickly without ever noticing. As a side note, currently, overlapping declarations produce an error (e.g., the containing folder under GPL and subfolders under LGPL).

After these improvements are merged, this check (and Travis) should always tell you if you forgot to add a module and it should also point out which one. The omission was on my part that I forgot to check `.java` files, so that's why you didn't get a notification about it before.

bkil-syslogng:
Squashed and rebased to master. Backup saved:
https://github.com/bkil-syslogng/syslog-ng/compare/master...bkil-syslogng:t/missing-py-license-bk0?expand=1

Issue #968, opened 2016-03-02
elasticsearch destination fails to index documents when global ES cluster state is not green


faxm0dem:
It seems that the elastic destination will fail to connect (and to index if already connected) if the ES global cluster state is not green (*i.e.* if any even unrelated index is not green).

If I'm right in this assumption, it is a serious limitation: it should be possible to index documents even if the cluster is yellow, or even red (when the red applies to an unrelated-to-syslogng index)

juhaszviktor:
First we wait for green, then wait for yellow. As far as I know if it is red, that means your records won't be indexed. 
Reading the definition of states I wouldn't send logs if the state is red:

[red] (https://www.elastic.co/guide/en/elasticsearch/guide/current/_cluster_health.html)
At least one primary shard (and all of its replicas) are missing. This means that you are missing data: searches will return partial results, and indexing into that shard will return an exception."

faxm0dem:
I disagree. Imagine the following scenario:

1. index `foobar` is red
2. index `syslog-ng` is green (syslog-ng is configured to write there)
3. due to 1. the cluster is red
4. syslog-ng does not join the cluster because of 3.

lbudai:
@faxm0dem : to be honest we think that skipping the health check is dangerous... (maybe we don't really understand your problem, but we trust you and believe you that this is a problem :-) )

So my suggestion would be the following, if it acceptable for you:
 * initiate a new config option in the elasticsearch driver (skip-es-cluster-health-check, or something like similar and sorter... )
 * default value of the new option would be FALSE, you have to set it explicitly in your config

What do you think?

faxm0dem:
Sounds perfectly reasonable to me, thanks!

lbudai:
@faxm0dem : Hi, I've just created a branch, unfortunately I cannot compile/test it right now, but could work. If it works for you then I'll make some further investigation around this topic and contact to you in order to create a final solution (or set this solution as final).

branchname: f/elasticsearch-skip-cluster-health-check

faxm0dem:
I'll test it this afternoon, thanks!

faxm0dem:
`s/logger.warning/logger.warn/`

lbudai:
yep... I did not have a working java environment when I wrote the code :-) 
I'll fix it.

lbudai:
@faxm0dem : fixed

faxm0dem:
LGTM

```
[...]

PR #986, opened 2016-03-11
elasticsearch2: added an option for skipping cluster health check

lbudai:
This patch makes it possible to index documents when the cluster state is
yellow, or even red. This could be only possible when the index you want
to use is writable.

How to use?
Add `skip_cluster_health_check("true")` to your elasticsearch2 destination.

Fixes: #968


pzoleex:
Hi,
I would change something in ESClient.java:
logger.info("connecting to cluster, cluster_name='" + clusterName + "'"); -> go to before any check, its important info

if (options.getSkipClusterHealthCheck()) { -> do not return, just add else branch where it checks the cluster state as earlier
Otherwise we will lose the connected info line at the end of this function that is also important

lbudai:
Actually we cannot be sure that we are connected or not.
It just checking the cluster health check.

But you are right, this is not a clear situation.
I'll modify the log message so I won't need the `else` :-)

lbudai:
@pzoleex : could you check it again?

pzoleex:
ok it's fine for me, however, the method name (connect()) can be misleading (but it was wrong initially so not belongs to these patches)

lbudai:
Could someone merge this PR?
@pzoleex ? @juhaszviktor ?

I sent this PR a month ago...

juhaszviktor:
:+1:

PR #989, opened 2016-03-16
service-management: if previous code was compiled with ENABLE_SYSTEMD,

juhaszviktor:
the systemd service manager will be the current service manager,
even that systemd wasn't active.

Signed-off-by: Juhász Viktor <jviktor@balabit.hu>

EDIT: Backport of #327


bkil-syslogng:
:+1:

Issue #990, opened 2016-03-17
Eating CPU after a big block of kernel messages


mdebb:
When a big amount of kernel log is produced in a very short time (more than 510 lines of kmem_cache dumps in this case), the syslog-ng process can enter into a spin. It drops the rest of the lines, and logs
`syslog-ng[32227]: I/O error occurred while reading; fd='13', error='Broken pipe (32)'` then spins like the following strace output. New kernel messages are not going to be logged:

```
epoll_wait(3, {{EPOLLIN, {u32=16754136, u64=16754136}}}, 7, 10723) = 1
read(5, "\1\0\0\0\0\0\0\0", 8)          = 8
fcntl(13, F_GETFD)                      = 0x1 (flags FD_CLOEXEC)
fcntl(13, F_GETFL)                      = 0x8800 (flags O_RDONLY|O_NONBLOCK|O_LARGEFILE)
setsockopt(13, SOL_SOCKET, SO_OOBINLINE, [1], 4) = -1 ENOTSOCK (Socket operation on non-socket)
epoll_ctl(3, EPOLL_CTL_ADD, 13, {EPOLLIN, {u32=16931120, u64=16931120}}) = 0
epoll_wait(3, {{EPOLLIN|EPOLLERR, {u32=16931120, u64=16931120}}}, 8, 10722) = 1
epoll_ctl(3, EPOLL_CTL_DEL, 13, {0, {u32=16931120, u64=16931120}}) = 0
write(19, "\1\0\0\0\0\0\0\0", 8)        = 8
epoll_wait(3, {{EPOLLIN, {u32=16754136, u64=16754136}}}, 7, 10722) = 1
read(5, "\1\0\0\0\0\0\0\0", 8)          = 8
fcntl(13, F_GETFD)                      = 0x1 (flags FD_CLOEXEC)
fcntl(13, F_GETFL)                      = 0x8800 (flags O_RDONLY|O_NONBLOCK|O_LARGEFILE)
setsockopt(13, SOL_SOCKET, SO_OOBINLINE, [1], 4) = -1 ENOTSOCK (Socket operation on non-socket)
epoll_ctl(3, EPOLL_CTL_ADD, 13, {EPOLLIN, {u32=16931120, u64=16931120}}) = 0
epoll_wait(3, {{EPOLLIN|EPOLLERR, {u32=16931120, u64=16931120}}}, 8, 10722) = 1
[...]

PR #992, opened 2016-03-17
3.7/f/file source reopen on error

bazsi:
This is an attempt at fixing #990

### DIFF
--- a/modules/affile/affile-source.c
+++ b/modules/affile/affile-source.c
@@ -253,13 +253,63 @@ affile_sd_construct_proto(AFFileSourceDriver *self, gint fd)
     }
 }
 
+static void
+affile_sd_reopen_on_notify(LogPipe *s)
+{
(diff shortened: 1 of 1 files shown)


bazsi:
@mdebb has confirmed that the fix seems to work for him, so I forward ported this to current master, which is PR #993

The fix is more like a quickfix and the code in this area could be improved (like merging the open path of the affile driver in case of initial open and subsequent opens), but as it addresses a critical bug I'd suggest to merge it as is, unless there's someone who can take on to improve the code.

lbudai:
@bazsi: I agree, we should merge it as is.

( I have some refactors, but it needs some polishment and I wouldn't block 3.7.3 with this , https://github.com/lbudai/syslog-ng/commit/6a8f3a1ac009b02a74cb6ebdafe30f8e494cb76e )

PR #996, opened 2016-03-22
elasticsearch: Added 2.2 support

pzoleex:
This solution is compatible with elastic 2.1 server too but must compile with elastic 2.2 libs
and use 2.2 libs in syslog-ng


faxm0dem:
yeah I believe that should work :+1:

lbudai:
@pzoleex : Thanks!

PR #997, opened 2016-03-23
Update libmongo-client library to upstream master

MrAnno:
algernon/libmongo-client@7ccdf08e:
> bson: Fix an int size mismatch in _bson_cursor_find()

>_bson_cursor_find() used gint32 for position tracking, while
initializing it to a size_t value. With gcc5 on higher optimisation
levels, this made tests fail. The obvious fix is to use size_t for
position tracking.


bkil-syslogng:
:+1: from me if Travis passes

lbudai:
Thanks!

PR #1003, opened 2016-03-27
Fix finding JNI library

### LINKED ISSUE #1002: jni.h: No such file or directory
While trying to compile current git master:
```
[  120s]   CC       modules/java/native/modules_java_libmod_java_la-java-grammar.lo
[  120s] In file included from modules/java/native/java-parser.h:30:0,
[  120s]                  from modules/java/native/java-grammar.y:27:
[  120s] modules/java/native/java-destination.h:28:17: fatal error: jni.h: No such file or directory
[  120s]  #include <jni.h>
[  120s]                  ^
[  120s] compilation terminated.
[  120s] Makefile:10300: recipe for target 'modules/java/native/modules_java_libmod_java_la-java-grammar.lo' failed
[  120s] make[2]: *** [modules/java/native/modules_java_libmod_java_la-java-grammar.lo] Error 1
[  120s] Makefile:12463: recipe for target 'all-recursive' failed
[  120s] make[1]: *** [all-recursive] Error 1
[  120s] Makefile:5093: recipe for target 'all' failed
[  120s] make: *** [all] Error 2
```
A few days ago it still worked. @MrAnno: you had a few Java related changes for OS X recently. Could that affect also Linux?

MrAnno:
Fixes #1002

The find command does not search in symlink directories by default.
Adding a trailing slash after the path solves the problem.
In this case, the find command starts from the target of the symbolic link, not from the symbolic link itself.

### DIFF
--- a/m4/check_java_support.m4
+++ b/m4/check_java_support.m4
@@ -77,7 +77,7 @@ AC_DEFUN([AX_CHECK_JAVA_VERSION],
       if test -e "$JAVA_HOME_CHECKER"; then
         JNI_HOME=`$JAVA_HOME_CHECKER`
       else
-        JNI_HOME=`echo $JAVAC_BIN | sed "s/\(.*\)[[/]]bin[[/]]java.*/\1/"`
+        JNI_HOME=`echo $JAVAC_BIN | sed "s/\(.*\)[[/]]bin[[/]]java.*/\1\//"`
       fi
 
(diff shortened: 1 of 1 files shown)


bkil-syslogng:
:+1: That's why I don't use forward slashes as `sed` delimiters. Although `find` has an option for symlink traversal ( `-L`), I usually do what you did here, so all is fine.

MrAnno:
There are implementation differences in `find` (the `-H` and `-L` flags) between BSD, OS X and Linux, so I thought this would be the "most portable" way.

PR #1004, opened 2016-03-28
split test_patterndb_rule testcase

bazsi:
This series splits test_patterndb_rule testcase to smaller functions, making the tested functions easier
to understand. This is a preparatory step to implement create-context action, which will be posted in a separate branch.

### DIFF
--- a/modules/dbparser/tests/test_patterndb.c
+++ b/modules/dbparser/tests/test_patterndb.c
@@ -395,6 +395,25 @@ gchar *pdb_ruletest_skeleton = "<patterndb version='3' pub_date='2010-02-22'>\
      </action>\
    </actions>\
   </rule>\
+  <rule provider='test' id='10d' class='violation' context-scope='program' context-id='$PID' context-timeout='60'>\
+   <patterns>\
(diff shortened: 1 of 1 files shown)


lbudai:
LGTM
(just one small note: there is a typo in correl**_l_**ation)

bazsi:
Where's the typo?

bazsi:
ahh. so the verb correlate is with a simple "l".

that's a pretty big issue though as the entire plugin was consistently named as "correllate".

I would say we should merge this as is and then fix the typo with the appropriate aliases left intact. I would do the rename. I was so sure that it was written with two "l" letters.

bazsi:
So I would do the rename in a later patchset, but do that in the entire plugin,  not just the new additions.

Do you agree?

lbudai:
@bazsi : I absolutely agree.

bazsi:
Is this ok for merging? You said it was ok for you, but shall I wait for other reviewers? Can I press the merge button?

thanks

lbudai:
I think it is ok.

PR #1007, opened 2016-04-03
dbparser: add XML validation in test_patterndb

bazsi:
This series of patches improve test_patterndb test program a lot and also contains a number of refactors, in preparation for the new create-context action in patterndb.

It depends on PR #1004, so that should go in first, but I wanted to submit this separately to facilitate review in smaller batches.

### DIFF
--- a/modules/dbparser/pdb-file.c
+++ b/modules/dbparser/pdb-file.c
@@ -45,12 +45,15 @@ pdb_file_detect_version(const gchar *pdbfile, GError **error)
 
   while (fgets(line, sizeof(line), pdb))
     {
-      if (strstr(line, "<patterndb"))
+      gchar *patterndb_tag;
+
+      patterndb_tag = strstr(line, "<patterndb");
+      if (patterndb_tag)
         {
           gchar *version, *start_quote, *end_quote;
 
           /* ok, we do have the patterndb tag, look for the version attribute */
-          version = strstr(line, "version=");
+          version = strstr(patterndb_tag, "version=");
 
           if (!version)
             goto exit;
--- a/modules/dbparser/pdb-file.c
+++ b/modules/dbparser/pdb-file.c
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2002-2016 Balabit
+ * Copyright (c) 1998-2016 Balázs Scheidler
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "pdb-file.h"
+#include "pdb-error.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+
+gint
+pdb_file_detect_version(const gchar *pdbfile, GError **error)
+{
+  FILE *pdb;
+  gchar line[1024];
+  gint result = 0;
+
+  pdb = fopen(pdbfile, "r");
+  if (!pdb)
+    {
+      g_set_error(error, PDB_ERROR, PDB_ERROR_FAILED, "Error opening file %s (%s)", pdbfile, g_strerror(errno));
+      return 0;
+    }
+
+  while (fgets(line, sizeof(line), pdb))
+    {
+      if (strstr(line, "<patterndb"))
+        {
+          gchar *version, *start_quote, *end_quote;
+
+          /* ok, we do have the patterndb tag, look for the version attribute */
+          version = strstr(line, "version=");
+
+          if (!version)
+            goto exit;
+          start_quote = version + 8;
+          end_quote = strchr(start_quote + 1, *start_quote);
+          if (!end_quote)
+            {
+              goto exit;
+            }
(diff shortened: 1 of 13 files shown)


bazsi:
I have now rebased this to master, so this should be the next step of integration in my dbparser test refactor series. Please don't merge anything to dbparser until this series of PRs are in, as this is pretty fragile, moving a lot of code around, thus even a single line changes in dbparser makes rebasing of this pretty difficult.

modules/dbparser/pdb-file.c
```diff
@@ -0,0 +1,165 @@
...
+          *end_quote = 0;
+          result = strtoll(start_quote + 1, NULL, 0);
+          break;
+        }
+    }
+ exit:
+  fclose(pdb);
+  if (!result)
+    {
+      g_set_error(error, PDB_ERROR, PDB_ERROR_FAILED, "Error detecting pdbfile version, <patterndb> version attribute not found or <patterndb> is not on its own line");
```
MrAnno: Is this error message still relevant after `dbparser: improve pdb_file_detect_version in pdb-file`?

bazsi: yup, it still has to be on the same line.

MrAnno: Oh sorry, I misunderstood the message.

MrAnno:
:+1:

bazsi:
Can I merge this with :+1: that @MrAnno posted?

modules/dbparser/pdb-file.c
bkil-syslogng: A bit narrower lines please?

modules/dbparser/pdbtool/pdbtool.c
```diff
@@ -707,20 +707,12 @@ pdbtool_test(int argc, char *argv[])
...
             {
-              fprintf(stderr, "%s: Unable to detect patterndb version, please write the <patterndb> tag on a single line\n", argv[arg_pos]);
-              failed_to_validate = TRUE;
-	      continue;
-            }
-          g_snprintf(cmd, sizeof(cmd), "xmllint --noout --nonet --schema %s/patterndb-%d.xsd %s", get_installation_path_for(PATH_XSDDIR), version, argv[arg_pos]);
-          if (system(cmd) != 0)
-            {
-              fprintf(stderr, "%s: xmllint returned an error, the executed command was: %s", argv[arg_pos], cmd);
+              fprintf(stderr, "%s: error validating pdb file: %s\n", argv[arg_pos], error->message);
```
bkil-syslogng: I liked the error==NULL check above. Want to do that here as well?

modules/dbparser/pdb-file.c
```diff
@@ -76,3 +78,88 @@ pdb_file_detect_version(const gchar *pdbfile, GError **error)
...
+
+  xsd_file = _get_xsd_file(version);
+  if (!is_file_regular(xsd_file))
+    {
+      g_set_error(error, PDB_ERROR, PDB_ERROR_FAILED, "XSD file is not available at %s", xsd_file);
+      g_free(xsd_file);
+      return FALSE;
+    }
+
+  xmllint_cmdline = g_strdup_printf("xmllint --noout --nonet --schema %s %s", xsd_file, filename);
```
bkil-syslogng: Should we use quoting to help support space in installation folder/file names?

modules/dbparser/pdb-file.c
```diff
@@ -76,3 +78,88 @@ pdb_file_detect_version(const gchar *pdbfile, GError **error)
...
+static const gchar *
+_get_xsddir_in_production(void)
+{
+  return get_installation_path_for(SYSLOG_NG_PATH_XSDDIR);
+}
+
+static const gchar *
+_get_xsddir(void)
+{
+  return _get_xsddir_in_build() ? : _get_xsddir_in_production();
```
bkil-syslogng: This works fine, although I would probably have `#ifdef`'ed the whole function definition above and this condition as well.

bkil-syslogng:
I didn't see any major issue with the code. I haven't verified the XML file changes, though. I assume tests work and run correctly. So other than the nitpicks above, :+1:

bazsi:
Thanks for the reviews. I have fixed two nitpicks:

1) too long line in pdb_file_detect_version()
2) and quoting in the invocation of xmllint

Note: that both were copies of code I've moved around and not new code.

I kept the conditionals around get_xsddir() as they were, I think it documents the intent pretty clearly: we use in-tree XSD files in case we are working in tree and installed versions if we are outside. If I was to add a lot of ifdefs there, that would be more difficult to read.

As these were only nitpicks that I mostly fixed, I would continue by merging this, once travis comes back green.

And after that's done, we should continue with PR #1012

modules/dbparser/pdb-file.c
```diff
@@ -77,3 +79,88 @@ pdb_file_detect_version(const gchar *pdbfile, GError **error)
     }
   return result;
 }
+
+static const gchar *
+_get_xsddir_in_build(void)
+{
+#ifdef SYSLOG_NG_ENABLE_DEBUG
+  const gchar *srcdir;
+  static gchar path[256];
```
lbudai: A small note here:
Maybe we could check if `PATH_MAX` is available or not,  if not then we could try to use pathconf (a separate `build_path(base_dir, relative_path)` function could be also added).

(IIRC I have a patch somewhere for determining path max ... maybe in PE).

This is not a blocker issue, it can be merged as-is (and later I'll send a patch).

modules/dbparser/pdb-file.c
```diff
@@ -0,0 +1,166 @@
...
+
+  srcdir = getenv("top_srcdir");
+  if (srcdir)
+    {
+      g_snprintf(path, sizeof(path), "%s/doc/xsd", srcdir);
```
bkil: We will need to check the return value of `g_snprintf()` as well. Would some kind of `g_strdup_printf()` work here instead?

bazsi: Hi,

The problem with allocating the string is that it makes calling this
function a lot more difficult, as the caller needs to free it.

This kind of pattern happens in syslog-ng a lot (e.g. formatting a string I
a static buffer and then returning that), and I would change them to use a
different pattern at once and not individually.

My point is that the caller shouldnt need to bother freeing the string as
that complicatedly call sites a lot. If we find a way how to do that, with
truncation and memory mgmt covered I am all for it.

BTW: truncation is not really a problem in this case but might in others.

bazsi:
I am merging this now, the snprintf() case should be handled as a separate issue, I have even filed an issue #1038

PR #1012, opened 2016-04-06
F/dbparser convert pdbload to state machine

bazsi:
This series of patches converts the patterndb XML loader code to be a state machine to make it
somewhat easier to understand. This also improves error handling by not accepting XML documents that are semantically incorrect.

SAX parsing is not my favourite, but using a single state helps separation (e.g. SRP) and code reusability. 

I still have a few dozen patches in the queue that will take this and allow reusing states by removing the current in_XXXX gboolean variables in PDBLoader.

### DIFF
--- a/doc/xsd/patterndb-4.xsd
+++ b/doc/xsd/patterndb-4.xsd
@@ -180,6 +180,7 @@
                 <xs:unique name="examples">
                     <xs:selector xpath="example"/>
                     <xs:field xpath="test_message"/>
+                    <xs:field xpath="@program"/>
                 </xs:unique>
             </xs:element>
 
--- a/modules/dbparser/pdb-load.c
+++ b/modules/dbparser/pdb-load.c
@@ -45,6 +45,7 @@ enum PDBLoaderState
   PDBL_RULE_TAG,
   PDBL_RULE_EXAMPLES,
   PDBL_RULE_EXAMPLE,
+  PDBL_RULE_EXAMPLE_TEST_MESSAGE,
   PDBL_RULE_ACTIONS,
   PDBL_RULE_ACTION,
   PDBL_RULE_ACTION_MESSAGE,
@@ -307,6 +308,7 @@ pdb_loader_start_element(GMarkupParseContext *context, const gchar *element_name
               if (strcmp(attribute_names[i], "program") == 0)
                 state->current_example->program = g_strdup(attribute_values[i]);
             }
+          state->current_state = PDBL_RULE_EXAMPLE_TEST_MESSAGE;
         }
       else if (strcmp(element_name, "test_values") == 0)
         {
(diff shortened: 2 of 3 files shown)


bazsi:
Please note that this also includes my previous f/dbparser-add-validation-in-tests (PR #1007), so not all patches are relevant, that should go in first.

bazsi:
This seems to parse my internal collection of patterndb xml files just fine. 

@faxm0dem I know you are using patterndb in production, could you test if "pdbtool test" reads your pdb files without reporting an error?

syslog-ng just got a lot more picky about syntax errors in the XML tree.

bazsi:
Rebased to master (still includes pr #1007, which should be merged first)

modules/dbparser/tests/test_parsers_e2e.c
```diff
@@ -38,18 +38,22 @@ do { \
...
+  <ruleset name='test1_program' id='480de478-d4a6-4a7f-bea4-0c0245d361e1'>\
+    <patterns>\
+      <pattern>test</pattern>\
+    </patterns>\
+    <rules>\
+      <rule id='1' class='test1' provider='my'>\
+        <patterns>\
+          <pattern>" /* HERE COMES THE GENERATED PATTERN */ ;
+
+gchar *pdb_parser_skeleton_postfix =  /* HERE IS THE GENERATED PATTERN */ "</pattern>\
```
bkil-syslogng: I like the terminology `suffix` better. Could you perhaps add some modifiers as well, like `static const`?

bazsi: Me too, but this is not new code, I just reindented it for readability.

bazsi:
I have rebased this against master again, this time master contains all dependencies, so this is the next to go in the series.

I know it is a lot of patches, but the point was to split it off to baby-steps in a way that makes it easier to review. You should either 1) review the entire set as a whole or 2) use gitk or some local patch viewer that let's you iterate through patches quickly.

If using the 2nd approach, you might want to "ignore space change", as I was changing indentation levels a lot, causing a lot of churn, but if you ignore space change, you'll immediately focus on the actual changes.

I would really like to get this merged. It's a fragile patchset, should anyone change the module, could cause a lot of conflicts during rebase.

bazsi:
I have accidentaly pushed this local branch to master, which I have reverted but github seems to think that I have closed this PR. Let me see if I can reopen it.

bazsi:
Seems the merges cannot be reopened. :(

I would then simply combine this with pr #1032, which contains these patches anyway. I am copying over the comment about the review above.

faxm0dem:
Hi @baszi sorry I must have missed this one. I'd love to test this for you on my patterndb files, but I have to strip those of all the incubator module calls like getent before - I'll try to do this this week

PR #1016, opened 2016-04-09
F/kvscanner separator

### LINKED ISSUE #1014: Separator option for key-value parser
It would be very useful if the key-value parser had the following options:

* `field-separator`
* `record-separator`

This would allow us for instance to easily parse [mysql logs](https://dev.mysql.com/doc/refman/5.7/en/mysql-logging.html):

```
Mar  7 12:39:25 myhost MysqlClient[20824]: SYSTEM_USER:'oscar', MYSQL_USER:'my_oscar', CONNECTION_ID:23, DB_SERVER:'127.0.0.1', DB:'--', QUERY:'USE test;'
```

bazsi:
This should fix #1014


MrAnno:
:+1:

Issue #910, opened 2016-02-08
Handle option names with hyphen (-) characters in java scls


fekete-robert:
Currently, Java scls do not handle the hyphens in option names, only underscores. This is inconsistent with the rest of the configuration, where hyphens and underscores are equivalent, and can be really confusing.

lbudai:
Hmm... this is not only a Java-specific problem.

lbudai:
PR #1024 should have to solve this.

PR #1024, opened 2016-04-15
scl: handle option names with both hyphen(-) and underscore (-) in SCLs

lbudai:
cfg-args: always normalize option names
    
    We have to normalize option names set by users.
    'Normalize' means that we replace all hyphen('-') to underscore('_') in an
    option name so we will find the key in the underlying hashtable and thus we
    can update or read its value.
    
    With this patch users can use both hyphen(-) and underscore (_) in
    SCLs.
    
    Fixes: #910
    
### DIFF
--- a/lib/cfg-args.c
+++ b/lib/cfg-args.c
@@ -23,6 +23,7 @@
  */
 #include "cfg-args.h"
 #include "messages.h"
+#include "str-utils.h"
 
 struct _CfgArgs
 {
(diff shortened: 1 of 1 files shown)


lib/block-ref-grammar.ym
```diff
@@ -61,7 +62,8 @@ block_arg
           LL_BLOCK
           {
             cfg_lexer_pop_context(lexer);
-            cfg_args_set(*result, $1, $3); free($1); free($3);
+            gchar *normalized_key = __normalize_key($1);
+            cfg_args_set(*result, normalized_key, $3); free($1); free($3); free(normalized_key);
           }
```
bazsi: normalized_key should be freed by g_free() while tokens should be freed by simple free()

lbudai:
@bazsi: fixed

lbudai:
do not merge this, some fixes are still in progress ...

lbudai:
Now, it works.

ihrwein:
@bazsi Could you check the logic, please? Thanks!

bazsi:
this looks good to me, although I originally didn't do this as these were all options that were introduced past the fact that the preference for dash (in contrast to underscore) were introduced.

dnsjts:
LGTM, :+1:

PR #1025, opened 2016-04-17
Added ElasticSearch Destination HTTP Implementation

talien:
This ES Destination uses the Jest library and the HTTP API of the
ElasticSearch Cluster. It uses different options for connections
that the other ES destinations, namely the host and port options
were elminated and cluster_url option was introduced.

The index, message_template, type options remained the same.

### DIFF
--- a/modules/java-modules/elastic-http/README.md
+++ b/modules/java-modules/elastic-http/README.md
@@ -1,0 +1,20 @@
+ElasticSearch HTTP Destination
+==============================
+
+This is a simple ElasticSearch HTTP destination using the Jest library. Although it support bulk insert (flush_limit), unfortunately it doesn't support flush timeout.
+
+Example configuration:
+```
+destination d_test {
+  java(
+    class_path("`module-path`/java-modules/*.jar:/install/jest-jars/*.jar")
+    class_name("org.syslog_ng.elasticsearch_http.ElasticSearchHTTPDestination")
+    option("index", "logs-$YEAR.$MONTH.$DAY")
+    option("message_template", "$(format-json --scope rfc5424 --exclude DATE --key ISODATE --pair @timestamp=$ISODATE)")
+    option("type", "logs")
+    option("cluster_url", "http://localhost:9200")
+    option("flush_limit", "100")
+  );
+};
+
+```
--- a/modules/java-modules/elastic-http/build.gradle
+++ b/modules/java-modules/elastic-http/build.gradle
@@ -1,0 +1,12 @@
+project.buildDir = syslogBuildDir+'/elastic-http/gradle'
+
+dependencies {
+    compile 'io.searchbox:jest:1.0.2'
+    compile name: 'syslog-ng-common'
+    compile name: 'syslog-ng-core'
+    compile 'junit:junit:4.12'
+    compile 'org.hamcrest:hamcrest-core:1.3'
+    compile 'log4j:log4j:1.2.17'
+    compile fileTree(dir: "/usr/lib/syslog-ng-java-module-dependency-jars/jars", includes: ['*.jar'])
+}
+
--- a/modules/java-modules/elastic-http/src/main/java/org/syslog_ng/elasticsearch_http/ElasticSearchHTTPDestination.java
+++ b/modules/java-modules/elastic-http/src/main/java/org/syslog_ng/elasticsearch_http/ElasticSearchHTTPDestination.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright (c) 2015 Balabit
+ * Copyright (c) 2015 Viktor Juhasz <viktor.juhasz@balabit.com>
+ * Copyright (c) 2015 Viktor Tusa <tusavik@gmail.com>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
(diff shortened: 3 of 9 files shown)


czanik:
First of all: thanks! And a few questions:
- which ElasticSearch versions are tested / supported with this module? I checked the Jest docs, but version support is a bit confusing there...
- could you provide a sample configuration to be used in syslog-ng.conf?
- could you make travis happy? The copyright check fails: add your module to tests/copyright/policy

lbudai:
Hi,

first, thank for your contribution.

I think HTTP should be only a mode extension to the existing ES destination implementation (especially as the driver implemented in Java).
In that case we could "fallback" to the HTTP mode when others are failed due to eg. incompatibility issues(or any other problems).

Adding HTTP as a mode extension to the current implementation is not as easy as to add a new one :-)
As the ESClient used by ESMessageProcessor coupled with org.elasticsearch.client.Client .
We need an adapter (interface) which then adapts the EsHttpClient. After that we have two options:
1. [the clean] use dependency injection and remove the abstract createClient
2. keep current createClient but it should return the Adapter client

The 2nd option is just for the sake of completeness :-) I'd prefer the 1st option.

Options are shared (as it almost the same in this implementation as in the original, non-http one - except the transport and node related options are removed, and the new
http options are added), so it should not be a problem.

If it would not use Java at all, then that would be a different sitation ;-)
In that case we would have a native HTTP-ES destination...

talien:
I agree that adding this to the existing one would be a better option, but I had a short time to put it together, and didn't want to complicate the existing module too much. (I guess this is the reason the elastic-v2 module was separated as well).

@czanik: I added a simple README file to the root of the module which contains an example config. I tested only with Elastic 1.7 because I shoot for AWS ES, and it uses the 1.7 version. I tested with AWS ES and it works fine if you don't need IAM auth (for example when you have an IP-restricted resource policy), but for IAM I need to sign the requests going to the AWS ES cluster. I plan to do that as well, that was the other reason I didn't want to merge it with the other module.

I added the module directory to the proper license checking file, but Travis is still failing and the logs do not contain any useful info. I did a make clean && make && make install && make check && make check-copyright on my machine and everything worked fine. I use the syslog-ng-dev Docker image.

lbudai:
@talien : reason of elastic-v2 is the broken internal API provided by elasticsearch :-(

Maybe we could add this destination into incubator or just keep this PR open and if we won't be able to implement this (for whatever reason) until 3.8 is released (exact time is not known currently) then we will integrate your original PR. Our implementation would use the jest part from your PR as it is working nice.

The reason that I don't want to integrate right now your PR to the master without trying to implement the http as an option to the existing elasticsearch destination is backward compatibility: if we would merge this as a new destination then we will have to support it till end of time :-) 

What do you think?

talien:
Basically you support functionality not modules. The user have to choice explicitly between the http or the native transport, this can be either the destination type, or a parameter in the destination. This module shares very little code with the other, only the architecture is the same.

I may start working on merging the two modules, but I have very little time and I would like to focus on the AWS request signing part as well. 

My goal is to put this and the request signing functionality into 3.8 if I can. I kindly ask you to make a sufficient tradeoff in this case or help me with the merging of the two modules.

lbudai:
But you cannot break user interface which is the config file ;-)
Of course, we can help you with merging, that was my original intention. On Friday I'll try to allocate some time for it.

talien:
I wrote to you on gitter to discuss the details.

bkil-syslogng:
@talien The network interface of VM Travis can get overloaded at times, that's why the last build failed. Container Travis in a matrix will help us avoid such cases after it is merged.

Some of your sources contain an unusual copyright declaration. My impression is that you didn't leave room between the list of copyright holders and the license text:
``` 
error: can't parse modules/java-modules/elastic-http/src/main/java/org/syslog_ng/elasticsearch_http/ElasticSearchHTTPOptions.java (nonstandard license declaration?)
```

`distcheck` fails because of an exception when compiling `elastic-http`. As recommended in the error message, you could enable stack tracing of Gradle to get more information (`--stacktrace --info --debug`). You may find this useful: https://stackoverflow.com/questions/29977196/how-can-i-make-gradle-stacktrace-the-default

Let me copy the error message from the raw log:
```
 > :elastic-http:jar
FAILURE:
Build failed with an exception.
* What went wrong:
Failed to capture snapshot of input files for task 'jar' during up-to-date check.  See stacktrace for details.
Failed to create parent directory '/home/travis/build/balabit/syslog-ng/syslog-ng-3.8.0alpha0/modules/java-modules/elastic-http' when creating directory '/home/travis/build/balabit/syslog-ng/syslog-ng-3.8.0alpha0/modules/java-modules/elastic-http/build/tmp/jar'
* Try:
Run with 
--stacktrace
 option to get the stack trace. Run with 
--info
 or 
--debug
 option to get more log output.
BUILD FAILED
```

By the way, you could also temporarily disable parallelism (-j) and `keep-going` in `.travis.yml` if it helps in your investigation.

lbudai:
@talien : I made some refactor around ElasticSearchDestination and add HTTP mode to it.
here is my branch: https://github.com/lbudai/syslog-ng/tree/f/elasticsearch-v2-mode-http

This is very experimental (it's not finished).
(I've tested it with ES2.1, it worked, but my environment is far from production. Could you test it?)

bazsi:
this seems to be superseeded by pr #1053, so I am closing this in favour of that one. @talien could you please review that PR if it indeed resolves your intent here?

thanks

PR #1030, opened 2016-04-21
afprog: Don't kill our own process group

hbakken:
If, for some reason, the program source/destination failed to set up its
own process group, we need to make sure we do not run killpg() on that
process group, as it would kill ourselves.

Signed-off-by: Henrik Grindal Bakken <henribak@cisco.com>


bazsi:
Looks good to me. :+1:

bkil-syslogng:
:+1:

PR #1031, opened 2016-04-21
Pass a GlobalConfig pointer to native_parser_proxy_new()

ihrwein:
Without this change, a native parser cannot use configuration dependent
features, such as templates.

Signed-off-by: Tibor Benke <ihrwein@gmail.com>

### DIFF
--- a/modules/native/parser.c
+++ b/modules/native/parser.c
@@ -40,7 +40,7 @@ __attribute__((visibility("hidden"))) int
 native_parser_proxy_init(struct NativeParserProxy* s);
 
 __attribute__((visibility("hidden"))) struct NativeParserProxy*
-native_parser_proxy_new(void);
+native_parser_proxy_new(GlobalConfig *cfg);
 
 __attribute__((visibility("hidden"))) struct NativeParserProxy*
 native_parser_proxy_clone(struct NativeParserProxy *self);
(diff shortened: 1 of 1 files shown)


ihrwein:
Ready for merge, but I'd like to prepare all modules to support this functionality. However, you can review it.

bazsi:
Please note that over time I want to remove the GlobalConfig dependency in my plugin refactor patches (that are not yet ready). But they are on the f/decouple-cfg branch.

bazsi:
LGTM

bkil-syslogng:
:+1:

lbudai:
@ihrwein : so you don't want this PR being merged until the modules are not finished?

ihrwein:
Thanks for the reviews!

@lbudai Yes, you are right. I'll prepare them tomorrow morning, they need a very little change. If somebody uses them, they build from their master branch which are using the old interface. 

@bazsi How do you imagine the GlobalConfig removal? Plugins won't store pointers to the full config, just copy the relevant data into their own data structure?

bazsi:
well, LogPipe instances would still get a GlobalConfig instance as a
parameter somewhere, but subsystems such as LogTemplate would not need a
GlobalConfig instance just to load plugins.

If you look at my unfinished patch series, you can see that instead of
passing GlobalConfig * here and there I would just pass pointers to
embedded structs. For example the plugin subsystem would use a
PluginContext to store its state, which is then contained by GlobalConfig.
All plugin related functions would only need PluginContext and not
GlobalConfig.


-- 
Bazsi

PR #1032, opened 2016-04-22
F/dbparser action create context

bazsi:
This branch creates a new action in db-parser(), one to create a new context for upcoming messages. It all depends on a number of PRs waiting in the queue.

    
    <rule provider='test' id='12' class='violation'>
      <patterns>
        <pattern>simple-message-with-action-to-create-context</pattern>
      </patterns>
      <actions>
        <action trigger='match'>
          <create-context context-id='1000' context-timeout='60' context-scope='program'>
            <message inherit-properties='context'>
              <values>
                <value name='MESSAGE'>context message</value>
              </values>
            </message>
          </create-context>
        </action>
      </actions>
    </rule>

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -1,3 +1,10 @@
+xsd_DATA	= \
+		doc/xsd/patterndb-1.xsd \
+		doc/xsd/patterndb-2.xsd \
+		doc/xsd/patterndb-3.xsd \
+		doc/xsd/patterndb-4.xsd \
+		doc/xsd/patterndb-5.xsd
+
 EXTRA_DIST	+=  \
 		doc/man/syslog-ng.8.xml \
 		doc/man/syslog-ng.conf.5.xml \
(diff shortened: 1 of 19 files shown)


faxm0dem:
Do you have a use-case example? thanks!

bazsi:
Yup. In ironport logs (an SMTP appliance) sessions can be split multiple
times.

E.g. email reception is under one sessionid (called mid there), and then
each delivery is potentially another.

We wanted to propagate information from one session to the next, so that
the sender is available.

faxm0dem:
okay, neat! very interesting!

bazsi:
It seems that I have unintentionally closed the PR for #1012, which is NOT merged to master yet. Since I can't reopen merged pull requests, I decided to combine #1012 into this one, so we should continue the review here. Sorry for the confusion.

I have added this comment for #1012, which is absolutely relevant when trying to review this branch:

```
I know it is a lot of patches, but the point was to split it off to baby-steps in a way that makes it easier to review. You should either 1) review the entire set as a whole or 2) use gitk or some local patch viewer that let's you iterate through patches quickly.

If using the 2nd approach, you might want to "ignore space change", as I was changing indentation levels a lot, causing a lot of churn, but if you ignore space change, you'll immediately focus on the actual changes.

I would really like to get this merged. It's a fragile patchset, should anyone change the module, could cause a lot of conflicts during rebase.
```

lbudai:
@bazsi : we are trying to process this PR in one of the next few days (and we will also run some extra regression tests). When those tests are passed (internal regression tests), and we finished with the review, we will merge it.

bazsi:
I have now fixed the issues reported by @faxm0dem, improved some error messages (as it was difficult even for me to decipher them) and rebased against master.

It should be ready to go in.

lbudai:
I made some further refactors during the review (small changes around the huge `switch-cases`). I'll setup a PR.

bazsi:
Thanks a lot, this is great news. I agree that the code can be even better,
and I can review the follow-ups. But believe me it was a lot better when
adding the create-context stuff.

lbudai:
Yes, I believe :)

This is why I did not address that in this PR.
After the refactor I made it is still not perfect, but maybe a little bit better .

faxm0dem:
@bazsi if I replace in my patterndb all `==` in string context by `eq` everything works fine, so :+1: LGTM thanks!

PR #1033, opened 2016-04-25
Add ability to use templates in both url and message format

avcbvamorec:
eg:
  java(
    class-path("/usr/lib64/syslog-ng/java-modules/*.jar")
    class-name("org.syslog_ng.http.HTTPDestination")
    option("url", "http://host:8080/prefix-${YEAR}.${MONTH}.${DAY}/mailtrack/${macro}/update")
    option("method", "POST")
    option("template", "$(format-json --scope dot_nv_pairs)")
  );

### DIFF
--- a/modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestination.java
+++ b/modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestination.java
@@ -98,12 +98,20 @@ public class HTTPDestination extends StructuredLogDestination {
                 return false;
             }
             responseCode = connection.getResponseCode();
+            if (isHTTPResponseError(responseCode)) {
+                logger.error("HTTP response code error: " + responseCode);
+                return false;
+            }
         } catch (IOException | SecurityException | IllegalStateException e) {
             logger.debug("error in writing message." +
                     (responseCode != 0 ? "Response code is " + responseCode : ""));
             return false;
         }
         return true;
+    }
+
+    private static boolean isHTTPResponseError(int responseCode) {
+        return responseCode >= 400 && responseCode <= 599;
     }
 
     @Override
--- a/modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestination.java
+++ b/modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestination.java
@@ -39,8 +39,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 
 
-public class HTTPDestination extends TextLogDestination {
-    private URL url;
+public class HTTPDestination extends StructuredLogDestination {
     Logger logger;
     private HTTPDestinationOptions options;
 
@@ -53,7 +52,7 @@ public class HTTPDestination extends TextLogDestination {
 
     @Override
     public String getNameByUniqOptions() {
-        return String.format("HTTPDestination,%s", options.getURL());
+        return String.format("HTTPDestination,%s", options.getURLTemplate());
     }
 
 
@@ -62,12 +61,7 @@ public class HTTPDestination extends TextLogDestination {
         try {
             options.init();
         } catch (InvalidOptionException e) {
-            return false;
-        }
-        try {
-            this.url = new URL(options.getURL());
-        } catch (MalformedURLException e) {
-            logger.error("A properly formatted URL is a required option for this destination");
+            logger.error(e);
             return false;
         }
         return true;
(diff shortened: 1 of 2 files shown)


bkil-syslogng:
If you click on the build details button (red 'x' in this case, you can check the details about why Travis fails. Here is the link in this case:
https://travis-ci.org/balabit/syslog-ng/jobs/125527650

The problem this time is that you have removed the license declaration, however every file must have one, mentioning authors and copyrights. You have also introduced a typo in Balabit's name: instead of all those variants, we have canonized on using simply `Balabit`.

avcbvamorec:
Fixed :-)

bkil-syslogng:
Congratulations :) Although, excuse us for the possible delay until we have time to do a technical review of your changes, but we will be on it. We will leave comments here when the review starts. Thanks again for your efforts.

modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestination.java
```diff
@@ -53,7 +53,7 @@ public HTTPDestination(long handle) {
 
     @Override
     public String getNameByUniqOptions() {
-        return String.format("HTTPDestination,%s", options.getURL());
+        return String.format("HTTPDestination,%s", "http://localhost:80/foo/bar/");
```
lbudai: If you will have two HTTP destinations in the same config then these options are not unique (this is important because this name is used as a unique key eg. in statistics).

modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestination.java
```diff
@@ -62,12 +62,8 @@ public boolean init() {
...
             options.init();
         } catch (InvalidOptionException e) {
-            return false;
-        }
-        try {
-            this.url = new URL(options.getURL());
-        } catch (MalformedURLException e) {
-            logger.error("A properly formatted URL is a required option for this destination");
+            logger.error("error during init: InvalidOptionException.");
+            logger.error(e);
```
lbudai: AFAIK if you are using the `toString` method of the Exception (and implicitly you are using: it will be called when you pass the exception object to a print/log method), then the name of the `Exception` will be also printed which means that it is enough to keep the second `logger.error`.

modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestination.java
```diff
@@ -85,9 +81,12 @@ public boolean isOpened() {
 
 
     @Override
-    public boolean send(String message) {
+    public boolean send(LogMessage msg) {
         int responseCode = 0;
+        String message;
         try {
+            this.url = new URL(options.getURL().getResolvedString(msg, getTemplateOptionsHandle(), LogTemplate.LTZ_SEND));
```
lbudai: As the URL itself is derived from the actual message(and URL is not used only in this method...) it should be a local variable.

modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestination.java
```diff
@@ -96,15 +95,10 @@ public boolean send(String message) {
                 osw.write(message);
                 osw.close();
             } catch (IOException e) {
-                logger.error("error in writing message.");
-                return false;
+		logger.error("error in writing message.");
+		return false;
```
lbudai: If the whitespace changes are intentional changes then it is better to move to a separate commit message.

modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestinationOptions.java
```diff
@@ -58,21 +60,30 @@ public Option getOption(String optionName) {
     return options.get(optionName);
   }
 
-  public String getURL() {
-    return options.get(URL).getValue();
+  public TemplateOption getURL() {
```
lbudai: This should be called `getURLTemplate()` (or `getUrlTemplate`...).

modules/java-modules/http/src/main/java/org/syslog_ng/http/HTTPDestinationOptions.java
```diff
@@ -58,21 +60,30 @@ public Option getOption(String optionName) {
...
 
   private void fillOptions() {
     fillStringOptions();
+    fillTemplateOptions();
   }
 
   private void fillStringOptions() {
-		options.put(new RequiredOptionDecorator(new StringOption(owner, URL)));
-		options.put(new EnumOptionDecorator(new StringOption(owner, METHOD, METHOD_DEFAULT), HTTP_METHODS));
+    //options.put(new RequiredOptionDecorator(new StringOption(owner, URL)));
```
lbudai: Could you remove this comment? (In general, it is confusing/disturbing when a comment message is actually an unused code.)

avcbvamorec:
Hi, 
@lbudai, I think I took all of your corrections and suggestions.
Cheers.

lbudai:
@avcbvamorec:hi, thanks for the patch. We will review soon.

bazsi:
I am afraid i am not qualified to review this, but seems to be lingering here for a long time now. Anyone?

lbudai:
@bazsi : I've already reviewed the original PR, and @avcbvamorec processed them on yesterday.
So what we have to do is to check whether the changes are fix all the issues: I think they are.

@avcbvamorec : only one issue remained: patch formatting.
Could you squash some of the patches?
I think two patches are enough here: one is for implementing the feautre, other is for the https://github.com/balabit/syslog-ng/pull/1033/commits/935901655a2c914540d3af65f09d95d300ffc249
patch.

And the commit message should have the following format:
```
modul: subject

signed-off-by: ....
```

eg.:
```
java-modules/http:  Add ability to use templates in both url and message format

the config example, etc...

signed-off-by: <you>
```

avcbvamorec:
I messed up a bit with git rebase, but I think it's done.

lbudai:
@avcbvamorec : thanks!

Issue #868, opened 2016-01-13
New default value for on-error() template option


mitzkia:
I would like to suggest to change the default option of on-error() template option. My suggestion is to change from "drop-message" to "drop-property".  This option value would be less painful, we will not miss the whole message but only the smallest “invalid” part.

bazsi:
I tend to agree but @algernoone had strong opinions on the matter when he
did that, so lets see what he thinks.

algernon:
My reasoning was that half a message with missing parts is worse than a completely missing one. In the latter case, I get a single error, and the incomplete message does not upset my tools. In the other case, I get a message for every dropped property, and there is a high chance that my tools will drop the message anyway, due to missing properties.

But I no longer use templates, or this functionality, so my opinions are far weaker than they used to be.

Issue #867, opened 2016-01-13
on-error("drop-message") option value renaming


mitzkia:
There is a template option in syslog-ng: on-error(). It has a default value: “drop-message”. The name and the working of this option is a little bit misunderstanding. 
The admin guide refers for this option as:
```
drop-message: Drop the entire message and log an error message to the internal() source. This is the default behavior of syslog-ng OSE.
```
The problem is that syslog-ng does not drop this message (check it with syslog-ng-ctl stats), but emptied it by the template(). I suggest to change the “drop” option name to “empty” also in all similar option names: “drop-message”, “drop-property”, “silently-drop-message”, “silently-drop-property”.

To trigger this option here is an example:
syslog-ng.conf:
```
@version: 3.7
options {
   on_error("drop-message");
   stats_level(3);
};
source s_internal { internal();};
source s_file { file("input2" flags(no-parse)) ;};
parser create_cef_blocks { csv-parser(columns(".cef.block_A", ".cef.blockB", ".cef.blockC") delimiters("@"));} ;
destination d_file {
    file( "output_file.log"
    template("$(format-json --scope everything pid=boolean($.cef.blockB) )\n")
    ); };
log {
    source(s_internal);
    source(s_file);
    parser(create_cef_blocks);
    destination(d_file);
};
```
input2 file content:
```
aaaaaa@bbb b=b|b\b@cccccc INFO: Server startup in 1206 ms

```
check console output:
```
[2016-01-12T14:58:33.601432] syslog-ng starting up; version='3.8.0alpha0'
[2016-01-12T14:58:34.602838] Incoming log entry; line='aaaaaa@bbb b=b|b\b@cccccc INFO: Server startup in 1206 ms'
[2016-01-12T14:58:34.602961] Message parsing complete; result='1', rule='create_cef_blocks', location='syslog-ng.conf:7:28'
[2016-01-12T14:58:34.603081] Initializing destination file writer; template='output_file.log', filename='output_file.log'
[2016-01-12T14:58:34.603733] Casting error; value='$.cef.blockB', type-hint='boolean'
[2016-01-12T14:58:34.603778] Outgoing message; message='\x0a'
^C[2016-01-12T14:58:35.470406] syslog-ng shutting down; version='3.8.0alpha0'
```
checking statistics:
```
root@782f6c7f7ea2:/install/syslog-ng# sbin/syslog-ng-ctl stats | grep drop
dst.file;d_file#0;output_file.log;a;dropped;0
```

[...]

PR #1037, opened 2016-05-02
Add $(sum), $(min) and $(max) template functions

MrAnno:
This PR contains three new template functions, which can be used together with correlation plugins such as [correlation-parser](https://github.com/ihrwein/syslog-ng-rust-modules/tree/master/correlation-parser/correlation), [grouping-by() parser](https://czanik.blogs.balabit.com/2016/04/the-grouping_by-parser-in-syslog-ng-3-8/) and [PatternDB](https://www.balabit.com/sites/default/files/documents/syslog-ng-ose-latest-guides/en/syslog-ng-ose-guide-admin/html-single/index.html#patterndb-correlation).
These correlation plugins create message contexts that contain multiple messages.

The newly introduced `$(sum)`, `$(min)` and `$(max)` functions receive one argument, and operate on message contexts. The argument of the template function will be used to calculate the summation, the minimum and the maximum value of the context.

----

Example usage with the `grouping-by()` parser:
```
parser p_groupingby {
  grouping_by(
    key("${KEY}")
    timeout(5)
    aggregate(
      value("MAX" "$(max ${NUMBER})")
      inherit-mode("none")
    )
    inject-mode("pass-through")
  );
};
```
This parser groups messages based on their `${KEY}` value.
After the group is complete (timeout is reached), the aggregation will be triggered and `${MAX}` will contain the maximum value of the `${NUMBER}` fields in the message group.

### DIFF
--- a/libtest/template_lib.c
+++ b/libtest/template_lib.c
@@ -182,23 +182,32 @@ assert_template_format_with_escaping(const gchar *template, gboolean escaping,
 void
 assert_template_format_with_context(const gchar *template, const gchar *expected)
 {
-  LogTemplate *templ = compile_template(template, FALSE);
-  if (!templ)
-    return;
-
   LogMessage *msg;
-  GString *res = g_string_sized_new(128);
-  const gchar *context_id = "test-context-id";
   LogMessage *context[2];
 
   msg = create_sample_message();
   context[0] = context[1] = msg;
 
-  log_template_format_with_context(templ, context, 2, NULL, LTZ_LOCAL, 999, context_id, res);
+  assert_template_format_with_context_msgs(template, expected, context, 2);
+
+  log_msg_unref(msg);
+}
+
+void
+assert_template_format_with_context_msgs(const gchar *template, const gchar *expected,
+                                         LogMessage **msgs, gint num_messages)
+{
+  LogTemplate *templ = compile_template(template, FALSE);
+  if (!templ)
+    return;
+
+  GString *res = g_string_sized_new(128);
+  const gchar *context_id = "test-context-id";
+
+  log_template_format_with_context(templ, msgs, num_messages, NULL, LTZ_LOCAL, 999, context_id, res);
   expect_nstring(res->str, res->len, expected, strlen(expected), "context template test failed, template=%s", template);
   log_template_unref(templ);
   g_string_free(res, TRUE);
-  log_msg_unref(msg);
 }
 
 void
--- a/modules/basicfuncs/numeric-funcs.c
+++ b/modules/basicfuncs/numeric-funcs.c
@@ -21,6 +21,8 @@
  *
  */
 
+typedef gboolean (*ExtremumFunc)(gint64, gint64);
+
 static gboolean
 tf_num_parse(gint argc, GString *argv[],
 	     const gchar *func_name, gint64 *n, gint64 *m)
@@ -202,4 +204,85 @@ tf_num_sum_call(LogTemplateFunction *self, gpointer s,
 
 TEMPLATE_FUNCTION(TFSimpleFuncState, tf_num_sum,
                   tf_num_prepare, NULL, tf_num_sum_call,
+                  tf_simple_func_free_state, NULL);
+
+static gboolean
+_tf_num_get_first_valid_arg(const TFSimpleFuncState *state,
+                            const LogTemplateInvokeArgs *args,
+                            gint *message_index, gint64 *number)
+{
+  for (*message_index = 0; *message_index < args->num_messages; (*message_index)++)
+    {
+      LogMessage *message = args->messages[*message_index];
+
+      if (_tf_num_parse_arg_with_message(state, message, args, number))
+        return TRUE;
+    }
+
+  return FALSE;
+}
+
+static gboolean
+_tf_num_extremum(const TFSimpleFuncState *state,
+                 const LogTemplateInvokeArgs *args,
+                 ExtremumFunc extremum_compare, gint64 *extremum)
+{
+  gint64 number;
+
+  gint message_index;
+  if (!_tf_num_get_first_valid_arg(state, args, &message_index, extremum))
+    return FALSE;
+
+  for (; message_index < args->num_messages; ++message_index)
+    {
+      LogMessage *message = args->messages[message_index];
+
+      if (_tf_num_parse_arg_with_message(state, message, args, &number)
+          && extremum_compare(number, *extremum))
+        *extremum = number;
+    }
+
+  return TRUE;
+}
+
+static gboolean
+_extremum_minimum(gint64 number, gint64 min)
+{
+  return number < min;
+}
+
+static void
+tf_num_min_call(LogTemplateFunction *self, gpointer s,
+                const LogTemplateInvokeArgs *args, GString *result)
+{
+  TFSimpleFuncState *state = (TFSimpleFuncState *) s;
+  gint64 min;
+
+  if (_tf_num_extremum(state, args, _extremum_minimum, &min))
+    format_int64_padded(result, 0, ' ', 10, min);
+}
+
+TEMPLATE_FUNCTION(TFSimpleFuncState, tf_num_min,
+                  tf_num_prepare, NULL, tf_num_min_call,
+                  tf_simple_func_free_state, NULL);
+
+static gboolean
+_extremum_maximum(gint64 number, gint64 max)
+{
+  return number > max;
+}
+
+static void
+tf_num_max_call(LogTemplateFunction *self, gpointer s,
+                const LogTemplateInvokeArgs *args, GString *result)
+{
+  TFSimpleFuncState *state = (TFSimpleFuncState *) s;
+  gint64 max;
+
+  if (_tf_num_extremum(state, args, _extremum_maximum, &max))
+    format_int64_padded(result, 0, ' ', 10, max);
+}
+
+TEMPLATE_FUNCTION(TFSimpleFuncState, tf_num_max,
+                  tf_num_prepare, NULL, tf_num_max_call,
                   tf_simple_func_free_state, NULL);
--- a/modules/basicfuncs/numeric-funcs.c
+++ b/modules/basicfuncs/numeric-funcs.c
@@ -21,7 +21,7 @@
  *
  */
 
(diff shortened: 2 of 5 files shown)


libtest/template_lib.c
```diff
@@ -182,23 +182,31 @@ assert_template_format_with_escaping(const gchar *template, gboolean escaping,
...
+  msg = create_sample_message();
+  context[0] = context[1] = msg;
+
+  assert_template_format_with_context_msgs(template, expected, context, 2);
+
+  log_msg_unref(msg);
+}
+
+void
+assert_template_format_with_context_msgs(const gchar *template, const gchar *expected, LogMessage **msgs, gint num_messages)
```
bkil-syslogng: A bit long line there

modules/basicfuncs/numeric-funcs.c
```diff
@@ -21,6 +21,8 @@
  *
  */
 
+typedef gboolean (*extremum_func)(gint64, gint64);
+
```
bazsi: Pls use camelcase here as it is a type

modules/basicfuncs/numeric-funcs.c
```diff
@@ -130,3 +132,156 @@ tf_num_mod(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+static GString *
+_make_gstring_buffer(const LogTemplateInvokeArgs *args)
+{
+  if (args->bufs->len == 0)
+    g_ptr_array_add(args->bufs, g_string_sized_new(64));
+
+  return (GString *) g_ptr_array_index(args->bufs, 0);
+}
+
```
bazsi: This function should probably go to lib/template somehow. It is clearly operating on LogTemplateInvokeArgs, eg it seems to smell from feature envy

modules/basicfuncs/numeric-funcs.c
```diff
@@ -130,3 +132,156 @@ tf_num_mod(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+  GString *formatted_template_result = _make_gstring_buffer(args);
+  gint on_error = args->opts->on_error;
+
+  log_template_format(state->argv[0], message, args->opts, args->tz,
+    args->seq_num, args->context_id, formatted_template_result);
+
+  if (!parse_number_with_suffix(formatted_template_result->str, number))
+    {
+      if (!(on_error & ON_ERROR_SILENT))
+        msg_error("Parsing failed, template function's argument is not a number",
```
bazsi: On-error has been used for your hinting so far, so I am not sure we can apply it here.... Or if we do, we should think about other call sites as well. What are the rest of the numeric functions do with NaN values?

modules/basicfuncs/numeric-funcs.c
```diff
@@ -130,3 +132,156 @@ tf_num_mod(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+tf_num_prepare(LogTemplateFunction *self, gpointer s, LogTemplate *parent,
+               gint argc, gchar *argv[], GError **error)
+{
+  g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+  if (argc != 2)
+    {
+      g_set_error(error, LOG_TEMPLATE_ERROR, LOG_TEMPLATE_ERROR_COMPILE,
+        "Template function requires only one argument");
+      return FALSE;
```
bazsi: The error should contain the template function name. Try an actual error in your configuration and check how syslog-ng reports it. I guess is not going to be easy to understand the error.

modules/basicfuncs/numeric-funcs.c
```diff
@@ -130,3 +132,156 @@ tf_num_mod(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+                 const LogTemplateInvokeArgs *args,
+                 extremum_func extremum_compare, gint64 *extremum)
+{
+  gint64 number;
+
+  gint message_index;
+  if (!_tf_num_get_first_valid_arg(state, args, &message_index, extremum))
+    return FALSE;
+
+  for (; message_index < args->num_messages; ++message_index)
```
bazsi: I think you might be able to not code the same for loop twice (once here, once in the first_valid_arg function) if you simply kept track whether you encountered the first valid number already.

bazsi:
This is great and very useful stuff which I'd liked merged. My notes are no biggies, but would be great to see at least some of these to be addressed.

MrAnno:
I'm going to fix these problems tomorrow. Thank you for the review.

modules/basicfuncs/numeric-funcs.c
MrAnno: I tried a few different ways to make an uninterpretable error message, but it usually looks like this:
```
Error parsing dbparser, Error compiling template (Template function requires only one argument) in /home/laszlovarady/install-syslog-ng/etc/syslog-ng.conf at line 21, column 19:

      value("SUM" "$(sum ${MESSAGE} ${NOPE})")
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
But you are absolutely right, I'll make it consistent.

modules/basicfuncs/numeric-funcs.c
MrAnno: Fixed.

libtest/template_lib.c
MrAnno: Fixed.

modules/basicfuncs/numeric-funcs.c
MrAnno: Other numeric functions report the problem with `msg_debug()` and return `"NaN"` as a string.

In this case, I think it wouldn't be a good idea to stop processing the summation, minimum or maximum of the context when an invalid value appears.

Should I just print the error using `msg_error()` or `msg_warning()` without checking this `ON_ERROR_SILENT` field?

bkil-syslogng: You might consider to also check ON_ERROR_DROP_MESSAGE. You could produce an empty output if an invalid entry is found. Otherwise you can ignore it, as has been done here.

See also:
https://github.com/balabit/syslog-ng/issues/868
https://github.com/balabit/syslog-ng/issues/867

modules/basicfuncs/numeric-funcs.c
MrAnno: I moved this function into `template/function.h`. It's not the best place, because LogTemplateInvokeArgs is initialized in `log_template_append_format_with_context` without "taking ownership" of its `GPtrArray *bufs` attribute. Do you have any better idea?

bkil-syslogng: An alternative refactoring would have been to make it only operate on `(GPtrArray *)` instead. Are you sure this can and will be used at other places as well?

bazsi: Yeah, you might be right. If it doesn't lend itself easily I can live with
the private implementation. The whole point of the args member is to
provide a cache instead of having to allocate GString instances on the fast
path. If scratch buffers would be improved (I wouldn't use extend its usage
beyond what it is used for before some serious rethought), then that could
probably be used instead if this private cache.

So if it doesn't fit nicely into template/function.h then let's keep it
private. Perhaps with a better name. It doesn't make GString instances.

MrAnno: I'll move it back to the basicfuncs module with a different name (to make its scope smaller) until we improve the scratch buffer part of the `LogTemplateInvokeArgs`. This improvement part could be a separate pull request from now.

modules/basicfuncs/numeric-funcs.c
MrAnno: We made a small refactor on this part with @bkil-syslogng to reduce the redundancies of iterating through the message context.

MrAnno:
I modified multiple parts based on @bazsi's and @bkil-syslogng's review notes. Can I squash these fixups?

bazsi:
The followups look great, please squash them.

There's one unaddressed review note, which is not a showstopper, so can go in without addressing it, e.g. it's up to you.

The function that gets the first value from the list has the same loop as the aggregation function. Presumably this is because the aggregators are expecting two values. This can be eliminated by using an explicit initialization value for the accumulator.

For maximum aggregation it can by thr int64 minimum value, for sum it is zero and so on.

This is how it is solved in functional languages.

MrAnno:
Originally, I created the `_tf_num_get_first_valid_arg` function to eliminate any "first_found" boolean flag in the loop of minimum and maximum. I like the idea of the explicit initialization to int64's min and max value, but there is one ugly corner case:

What should I do when all values are "NaN" in the context? I think it would be nasty to return int64-min or int64-max in that case.
Also, if I want to detect this corner case, I should use a flag again.

bkil-syslogng:
@bazsi I've offered the same alternative in person, and we've discussed it with @MrAnno that it has drawbacks. For example, you either don't support INT max/min as a value and treat it as completely extremal, or you must keep track of whether we have already found a valid value or not. If we don't encounter a value, we return an empty string (Nothing), otherwise we return the stringified number (Maybe Int). The latter would take extra compute cycles compared to the optimized solution of @MrAnno. The former isn't that elegant.

And what would you do for sum? An empty sum could be defined as `0`, but what about a median or average? We both see what you want to do, but we don't see the benefit of your proposed improvement. 

On the other hand, the iteration might be abstracted again on a functional basis if it hurts that much - I'd love to see more Haskell code in C anyway.

Note that we've implemented it just like `foldl1` (hand compiled a composition with a `filter`/`catMaybes`). `foldl` would need the additional initialization as you have mentioned.

bkil-syslogng:
Okay, as per the above, I'll give my :+1:. I have further improvements in the pipe, but let's not block this feature with those for now. I'll submit them separately. Here's a work in progress about what I had in mind:
https://github.com/bkil-syslogng/syslog-ng/tree/p/ctx-template-funcs2

So if I'm reading @bazsi correctly, he also gave a plus one:
"There's one unaddressed review note, which is not a showstopper, so can go in without addressing it, e.g. it's up to you."

@ihrwein @dnsjts want to merge it?

ihrwein:
:+1:  from me, I'll push the merge button

ihrwein:
@fekete-robert : Could you documeny this new feature, please? Thanks!

Issue #1006, opened 2016-04-03
destination AWS CloudWatch Logs


philtay:
Is there any direct way to send log messages to [AWS CloudWatch Logs](http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/WhatIsCloudWatchLogs.html)? I mean a builtin and supported feature, along the lines of Docker [awslogs driver](https://docs.docker.com/engine/admin/logging/awslogs/), and not something I have to create myself using custom destinations and/or external programs. I think this would be a very useful feature used by a lot of people given the success of Docker's awslogs driver.

bazsi:
It does sound useful indeed. Do you perhaps have a link to the API that
shows how to submit logs there?

philtay:
Sure, the [CloudWatch Logs API Reference](http://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/Welcome.html). It could look a bit overwhelming at first but, if you dig deeper, it's actually quite simple. In the [documentation](http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/WhatIsCloudWatchLogs.html) you will also find info about the [CloudWatch Logs Agent](http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/AgentReference.html) (an official and open source Python daemon that pushes logs from local files to CloudWatch Logs). The already mentioned Docker awslogs driver also offers an example on how to use the CloudWatch Logs API.

bazsi:
Well, it really seems to require a bit more than a stateless http
destination what we currently have. We need to keep the security cookie and
the sequence token while making requests.

Do you know why sequence token is needed? As I see it's use is optional,
but I don't know what happens if it's not included.

philtay:
Yes, pure stateless http is not enough. But:

* You don't have to keep any security cookie, you just sign each http request independently from the others. To do that only some standard cryptographic primitives (e.g. SHA256) are needed. More info and code [here](http://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/CommonParametersV4.html) and [here](http://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html). Basically the security part of the API is 100% stateless.

[...]

PR #1040, opened 2016-05-05
python: introduce flush-lines option

ngergo:
!!! EXPERIMENTAL FEATURE !!!

Syslog-ng python destination respects `flush-lines(<INT>)` option. It
uses `Py_ListObject` to queue incoming messages to the destination and
if the maximum size is reached it sends them to one of the workers as a
python list.

If `flush-lines(1)` is set, the list passed to one of the workers will
contain only one message.

If `flush-lines` is not set, python destination will use `cfg` default.

_Further improvements_:
- make python destination respect `flush-timeout` option
- do not drop messages on reload with different `flush-lines` value

_This feature was needed_ in order to make developers write destinations
that can send batched events.

_Example was modified_ according to the changes.

ref #1006

Signed-off-by: Gergő Nagy grigori.grant@gmail.com

### DIFF
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -49,6 +49,14 @@ typedef struct
 
   struct
   {
+    PyListObject* list;
+    gint current_size;
+    gint max_size;
+    // TODO: mutex is needed
+  } batch;
+
+  struct
+  {
     PyObject *class;
     PyObject *instance;
     PyObject *is_opened;
@@ -91,6 +99,13 @@ python_dd_set_imports(LogDriver *d, GList *imports)
 
   string_list_free(self->imports);
   self->imports = imports;
+}
+
+void
+python_dd_set_flush_lines(LogDriver *d, gint lines)
+{
+    PythonDestDriver *self = (PythonDestDriver *)d;
+    self->batch.max_size = lines;
 }
 
 void
@@ -349,21 +364,51 @@ _py_init_object(PythonDestDriver *self)
   return TRUE;
 }
 
-static worker_insert_result_t
-python_dd_insert(LogThrDestDriver *d, LogMessage *msg)
+static void
+python_dd_free_batch_list(PythonDestDriver *self)
 {
-  PythonDestDriver *self = (PythonDestDriver *)d;
+  self->batch.current_size = 0; // TODO: OK because of list reference handling
+}
+
+static worker_insert_result_t // TODO: shouldnt it return gboolean
+python_dd_batch_flush(PythonDestDriver *self)
+{
   worker_insert_result_t result = WORKER_INSERT_RESULT_ERROR;
   gboolean success;
-  PyObject *msg_object;
-  PyGILState_STATE gstate;
+  PyObject* list_object = (PyObject *)self->batch.list;
 
-  gstate = PyGILState_Ensure();
   if (!_py_invoke_is_opened(self))
     {
       result = WORKER_INSERT_RESULT_NOT_CONNECTED;
       goto exit;
     }
+
+  success = _py_invoke_send(self, list_object);
+  if (success)
+    {
+      result = WORKER_INSERT_RESULT_SUCCESS;
+      python_dd_free_batch_list(self);
+    }
+  else
+    {
+      msg_error("Python send() method returned failure, suspending destination for time_reopen()",
+                evt_tag_str("driver", self->super.super.super.id),
+                evt_tag_str("class", self->class),
+                evt_tag_int("time_reopen", self->super.time_reopen),
+                NULL);
+    }
+
+ exit:
+  return result;
+}
+
+static worker_insert_result_t
+python_dd_insert_one(PythonDestDriver *self, LogMessage *msg)
+{
+  worker_insert_result_t result = WORKER_INSERT_RESULT_ERROR;
+  gboolean success;
+  PyObject *msg_object;
+
   if (self->vp)
     {
       success = py_value_pairs_apply(self->vp, &self->template_options, self->super.seq_num, msg, &msg_object);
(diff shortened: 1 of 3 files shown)


ihrwein:
I'd like to review this tomorrow.

bazsi:
I think flush-timeout was more or less removed from LogWriter and is
present for compatibility only. We always flush if the queue is empty

bkil-syslogng:
Thank you for your contribution. Before proceeding to a review, I'd like to ask that you please make it pass through static analysis (i.e., at least no compiler warnings) and make it compile in Travis.

ngergo:
If u had followed this thread u would have seen it's not a "contribution" but a pilot. I've never had any intentions to ask u to merge this code... just to ask @bazsi or @ihrwein  to check if he is OK with the concept and one can try out the cloudwatch destination end-to-end. I'm not blind not to see the warnings... Thank you for your understanding! Hope cleaned ur cup ;)

ngergo:
@bazsi what do u mean "we flush if the queue is empty"?  I didn't get it. Isn't `flush-timeout` is for flushing the queue when it hasn't been filled in the given period?

juhaszviktor:
Hi @ngergo,

I know that you asked @bazsi, but let me to answer :)
The flushing strategy is, that we flush in two cases (in case of LogWriter):
- reached the flush limit
- the queue becomes empty

But this is irrelevant in your code, because python destination is LogThreadedDestination base, therefore there is no pre-implemented flushing strategy.

ngergo:
👍 Thanks :) got it.

bazsi:
e.g. the flush-timeout() parameter does nothing in current syslog-ng and we
shouldn't reintroduce that in newly written code.

-- 
Bazsi

ngergo:
I see. I think it would be better to tell what I want to reach... Let's say I have a python destination e.g. ES that has `flush-lines` set to 500 to increase ES indexing performance via sending batches. Tough I'm sending logs that should be "realtime" so every 60 seconds I want to get the freshest data that's why I declare `flush-timeout` to 60 secs... that says "flush the batch with no respect to its size if no logs came in the last 60 secs". 

As far as I can understand what @juhaszviktor said there is no such mechanism in the python dest. 
Why `flush-timeout` does nothing in syslog-ng? Should this mechanism be implemented in the python destination as a "private" option?

bazsi:
well, whenever log_queue_pop_head() returns NULL in logthrdestdriver, we
should flush.


-- 
Bazsi

ngergo:
Let's say we flush on `log_queue_pop_head() == NULL`, if I'm right that means in the case of not having any messages in the pipe we flush. As far as I can understand it would be the user's responsibility to try to ensure that the queue gets the batch with the size he wants to deliver batches with. She will lose the possibility of determining time-frame to deliver messages with and that would be a nice-to-have feature and also a pretty popular one.

What if python dest would have this queueing mechanism with the timeout feature (using ivykis let's say) and the docs would mention if the queue is greater than `1` slot u could lose ur batch (same in es).

Java ES destination also implements these features (`flush-lines`, `retires`) separately from the others in `ESBulkMessageProcessor` with a default timeout `1sec`. In a long term it would be nice to move this mechanism to the java destination not implementing privately for each dest. 
*Note*: it would be nice to make this timeout configurable ;)

*The most efficient way of writing to ES is using the *Bulk API*. When sending messages to ES u don't want to send just one message because its inefficient for the cluster to index and store. In production we use an indexer instance that collects logs from every logger in our infrastructure in a reliable way (e.g.: tcp, redis, etc...) and writes the batch to ES with the size of 1000 every 60 secs.*

In the python destination I would like to implement the possibility of writing destinations with a greater efficiency assuring "soft-real-time" log delivery.

So here is what I propose:
* `flush-lines`: let's queue logs in the python destination with to guarantee on not losing logs
* `retries`: how many times should we want to try to send msgs
* `timeout`: after not getting logs for the time specified let's flush
* [maybe] `throttle` to limit the number of logs can be sent in one minute

bazsi:
Hmmm I didn't know about the java feature. flush-timeout() was causing
headaches with the end-to-end acknowledgement mechanism. I don't remember
exactly what it was, and as far as I know ack path is not extended to the
insert callback of a logthrdestination.

ihrwein:
Hi! Thanks for sharing your idea with us!

I'm OK with the changes, if it works for you and makes a user happy, that's enough. Combining batch mode with Python destination could be a very versatile solution for a bunch of problems.

However, I'd like to point out that the argument type changing of the `self()` method is a breaking change. It won't cause problems if you use only the cloudwatch destination, but if other Python destinations are used it will cause problems.

If this feature is needed by other threaded destination, maybe it would be better to implement it there (so all config options and behavior would be consistent) by us.

If compatibility is big concern, maybe this destination can be implemented in pure Python (like `BatchingPythonDestination`), then Cloudwatch destination can subclass this baseclass.

What do you think?

ngergo:
Thanks for the constructive answers! 
I completely agree with your statement @ihrwein, breaking the API is not a wise decision. Tough as @bazsi mentioned there had been a reason of not having `flush-timeout` implemented in the core so it wouldn't be good to reintroduce something that has caused problems before. 

Summing up the things:
* `flush-timeout` cased problems in the logthrdestination
* we shouldn't break API
* batch mode would be fun :) 

Implementing `BatchingPythonDestination` with `send(self, msgs)`, and do queueing there, is not a good idea I think. Queueing and timing should be at the C side for the following reasons:
* it would be nice in the future to start multiple workers for one python destination, then to increase throughput it's advisable to hold the queue at the C side. Otherwise it's possible that none of the workers has enough messages to reach the batch size but the sum of the messages at each workers is higher than the threshold so none of them will flush.
* holding the queue at the C side aggregated for each workers makes the flushing mechanism truly efficient

My proposal:
* introduce a new mode in python destination, that is configurable with the option `batch(true)`, that will make python destination respect the `flush-lines` and `timeout` options.
* hold `send` on the C side as it is now, and implement `send_batch` that will invoke a method on the Python side `send_batch(self, msgs)` 
* so `LogDestination` would have two functions for sending: `send` and `send_batch`
* if you give `batch(true)` in the config, syslog-ng will try to invoke `send_batch` instead of `send`
* by default `LogDestination.send_batch` would throw `NotImplementedException` and if u inherit from this class as it is now u r able to override this method

What do u think? :)

kira-syslogng:

presidento:
It is outdated and was experimental.

PR #1041, opened 2016-05-05
F/ctx template funcs2 - mean operator, further stylistic improvements to #1037

bkil-syslogng:
Further improvements to #1037:
* Rename a few static functions
* Abstract out iteration when finding first element and aggregating
* Implement numeric mean (average)
* Refactor test cases to reduce redundancy

### DIFF
--- a/modules/basicfuncs/basic-funcs.c
+++ b/modules/basicfuncs/basic-funcs.c
@@ -72,6 +72,7 @@ static Plugin basicfuncs_plugins[] =
   TEMPLATE_FUNCTION_PLUGIN(tf_num_sum, "sum"),
   TEMPLATE_FUNCTION_PLUGIN(tf_num_min, "min"),
   TEMPLATE_FUNCTION_PLUGIN(tf_num_max, "max"),
+  TEMPLATE_FUNCTION_PLUGIN(tf_num_average, "average"),
 
   /* ip-funcs */
   TEMPLATE_FUNCTION_PLUGIN(tf_ipv4_to_int, "ipv4-to-int"),
--- a/modules/basicfuncs/numeric-funcs.c
+++ b/modules/basicfuncs/numeric-funcs.c
@@ -21,7 +21,7 @@
  *
  */
 
-typedef gint64 (*AggregateFunc)(gint64, gint64);
+typedef gboolean (*AggregateFunc)(gpointer, gint64);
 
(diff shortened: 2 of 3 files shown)


bazsi:
I would use average instead of mean. I know it's the same but average is
used more often as an aggregator.

bkil-syslogng:
@bazsi I've added aliases for `avg` and `average`. Do you like it this way?

bazsi:
Usually I prefer to have fewer names instead of more, as it makes understanding configs more difficult.

I for one had check if mean is matchematically equivalent to average. It is, but I didn't know for sure. I think the same lack of confidence could apply to others.

That's why I think it is not the best name for it. I can live with mean, but please don't use multiple aliases for the same thing. 

Do you perhaps have a reference in which programming language is this called mean in contrast with average?

faxm0dem:
average has fewer homonyms *i.e.* easier to google

bkil-syslogng:
As a data scientist, I don't use the word `average` often, but it's not me who decides. If you say that the community will understand it better that way, I'd gladly rename it. It's our goal to please our users.

As a side note, we generally use mean and standard deviation at the same time.

Some examples:
* http://sites.fas.harvard.edu/~stat335/overheads/Matlab_IntroII.pdf
* http://www.r-tutor.com/elementary-statistics/numerical-measures/mean
* http://stackoverflow.com/a/20820148
* ...

bkil-syslogng:
By the way, we need to decide whether this should stay as an integer one, change to a floating point version, implement both of them or select between them using a precision argument. That's one of the questions which kept @MrAnno from implementing this in the original, urgent scope.

lbudai:
If we are talking about statistics then `mean` is still not very precise: what you implemented is `arithmetic-mean`. 

So I'd use the `average` ;-)

lbudai:
@bkil-syslogng : at this PR just go with integers, IMHO.

bkil-syslogng:
Renamed `mean` to `average` and removed all aliases.

Backup saved under:
* https://github.com/balabit/syslog-ng/compare/master...bkil-syslogng:t/ctx-template-funcs2-bk0?expand=1

Rebased to `master`.

bazsi:
Thanks for working on this, it looks much better and the $(average) function is really welcome.
Merging this as well.

PR #1049, opened 2016-05-10
lib/logproto/logproto-server.c: refer to utf8 as utf-8

### LINKED ISSUE #1048: Support encoding on glib compiled with libiconv
If the user compiles glib with libiconv on a platform without a native iconv, character encoding can not be specified. (The issue is not easy to reproduce on many systems, but `LD_PRELOAD` should do the trick.)

Given the following config:
```
@version: 3.8

options
{ threaded(yes); }
;
source s_program_37b57ebaaa6a460085d2d35d6f7d56ec
{ program("cat" encoding("UTF-8")); }
;
destination d_file_a3219214c5e541d5a18b0e9ca3eedee2
{ file("/tmp/81525975-13e5-4a41-8e9e-1bbd7cdc3d25.txt"); }
;
log
{ source(s_program_37b57ebaaa6a460085d2d35d6f7d56ec); destination(d_file_a3219214c5e541d5a18b0e9ca3eedee2); flags(flow-control); };
```

Startup fails with the following message:
```
[2016-05-10T11:51:22.050595] Module loaded and initialized successfully; module='afprog'
Error parsing afprog, unknown encoding UTF-8 in /home/tamasnagy/install/opt/syslog-ng/etc/syslog-ng.conf at line 7, column 88:

{ program("/bin/cat /home/tamasnagy/install/opt/syslog-ng/etc/syslog-ng.conf" encoding("UTF-8")); }
                                                                                       ^^^^^^^

syslog-ng documentation: http://www.balabit.com/support/documentation/?product=syslog-ng
mailing list: https://lists.balabit.hu/mailman/listinfo/syslog-ng
```

It worked correctly on syslog-ng 3.7.3.

The root cause is that `lib/logproto/logproto-server.c` checks supported encodings by first trying to convert from "utf8" to the given encoding. However, this always fails because "utf8" is not considered a valid encoding on some of the libraries. Internal canonicalization is done by upper casing each character.

On a separate note, this analysis concludes that the user should be advised that only those codes can be used in a syslog-ng config which are supported by the used `iconv` implementation, i.e., giving `{ program("cat" encoding("UTF8")); }` should also fail on such a system. syslog-ng does not implement an extra aliasing layer and glib's aliasing doesn't seem to be exhaustive in this regard either.

Here are the aliases for `libiconv`:
* http://git.savannah.gnu.org/cgit/libiconv.git/tree/lib/encodings.def?id=48f31c744eb14da699c6ba3fd03f5f232c476bbb#n60

Here are some of the aliases for `glib`:
* https://github.com/GNOME/glib/blob/master/glib/libcharset/config.charset

bkil-syslogng:
Although, most libc implementations provide a native iconv with various
aliases, libiconv does not know about the alias `utf8`.
Neither does glib provide this alias if compiled with libiconv.

Furthermore, be advised that the user can only specify in the syslog-ng
config those iconv codes which are supported by the local iconv or
libiconv.

Fixes #1048

Regression introduced:

3fa60288548b54bcfad7b42deb145ef2eafa029c
"logproto: handle invalid encoding() options at parsing time"

### DIFF
--- a/lib/logproto/logproto-server.c
+++ b/lib/logproto/logproto-server.c
@@ -138,7 +138,7 @@ log_proto_server_options_set_encoding(LogProtoServerOptions *self, const gchar *
   self->encoding = g_strdup(encoding);
 
   /* validate encoding */
-  convert = g_iconv_open("utf8", encoding);
+  convert = g_iconv_open("utf-8", encoding);
   if (convert == (GIConv) -1)
     return FALSE;
   g_iconv_close(convert);


lbudai:
:+1:

MrAnno:
:+1:

ihrwein:
:+1:

ihrwein:
Thanks!

bazsi:
Thanks for tracking this down.

PR #1051, opened 2016-05-13
Add experimental CMake build scripts

### LINKED ISSUE #966: Support an alternative build system: CMake
I opened this issue for getting feedback from syslog-ng's user and developer community about supporting CMake as another build system.

CMake is a build system generator. It generates makefiles on *NIX, Xcode project on OSX, Visual Studio project on Windows.

I'd be nice to build syslog-ng with CMake (*as well*, I don't want to throw autotools away). 

It has the following advantages over autotools:
* no more shell code for configure
* native build system support: make, xcode, visual studio, code blocks
* faster configuration
* faster building
* most IDEs can import CMake projects (QtCreator, Clion, Eclipse, ...)
* optional package generation (debian, rpm)
* GUI for configuration [1]
* out of tree build is the default: it won't touch your source folder (automake does it even if you build your code out of tree)

The two build systems can nicely live next to each other (this is also an aim). Several high profile projects supported both autotools and CMake (e.g. LLVM, Google Test) or transitioned to the latter one (KDE [2]).

The first supported platforms would be Linux and OSX then we can extend it to the other *NIX systems (BSD, AIX, Solaris) or even Windows...

@dnsjts  and me has a PoC version of syslog-ng which can be built with CMake (core & 10+ modules) on Linux and OSX.

[1] http://3.bp.blogspot.com/-iz8Ho-23lNg/UThQoo7lvzI/AAAAAAAAAGM/Oqv0soTPKOM/s1600/cmake-guiqt.png
[2] https://lwn.net/Articles/188693/

ihrwein:
Fixes #966 

This PR adds CMake build system support to syslog-ng. It's experimental, which means that it's not in feature parity with the existing Autotools build system. However, syslog-ng can be built with CMake on Linux and OSX. And it's a turn-key solution.

Missing modules:
* afmongodb
* afamqp

Platforms checks, `ENABLE_*` flags can be missing. We implemented all platform checks what was needed to compile syslog-ng on Linux (without warnings) and on OSX.

The CMake based build system uses only system libraries (so you have to install libivykis and libjsonc). Submodule support can be added, but it wasn't required for the first iteration.

`make install` should work as expected on Linux and OSX.

`make check` (and unit test compilation) isn't ready.

We created more than 230 commits, this PR contains only the essence of our work, the intermediate steps are merged by files.

Here is an image about syslog-ng imported into Qt Creator:

image

# Usage

1. Create a clean clone from syslog-ng and checkout my branch
2. Step into the cloned repo and create a build directory:
 ```
mkdir build

cd build
```
3. Generate the Makefiles with CMake (CMAKE_INSTALL_PREFIX is the same as `--prefix`):
 ```
cmake -DCMAKE_INSTALL_PREFIX=/home/btibi/install/syslog-ng ..
```

4. Run make and install:
 ```
make -j  install
```

(Note, that you didn't have to run autogen.sh!!!)

If you prefer GUI tools to configure and build syslog-ng, you can install cmake-gui and run it from the build directory (`cmake-gui .`):

image

This PR is the result of the joint efforts of @MrAnno @dnsjts and me.

### CODE REVIEW

**Comment by bkil-syslogng** [comment]
To make it work on Travis, you can cherry-pick or copy&paste some of the improvements from the top of this one:
https://github.com/bkil-syslogng/syslog-ng/pull/25

**Comment by bazsi** [comment]
Wow, this one seems pretty comprehensive, especially as the first iteration.

When do you expect feature parity? Can we drop autotools completely at one point?

**Comment by ihrwein** [comment]
Update: I mentioned that `make check` and unit test compilation isn't included in this iteration.

@bazsi I cannot pinpoint a date. This work was started to support an alternative build system, not to drop autotools.

Bigger projects usually support both cmake and autotools, them drop the latter one (like LLVM did).

However, I can define the missing parts:
* the previously mentioned modules
* unit test compilation
* covering all macros defined by `AC_DEFINE*`

After that, we can build syslog-ng on more exotic platforms, like older UNIXes. Then, we can think about the transition process.

**Comment by ihrwein** [comment]
@bkil-syslogng thanks for the additional patches, I cherry-picked all of them.

I also added a new commit (https://github.com/balabit/syslog-ng/pull/1051/commits/aa8e98ce92fc6ebc6e7555bbb5ad951eedab4d35) to fix the building of dbparser (new files were added in the last week).

**Comment by ihrwein** [comment]
The build's output is a little bit mingled (https://travis-ci.org/balabit/syslog-ng/jobs/131657634) because of `make -j`, but we got nice results: "configure" took 5 secs, `make -j install` 72 secs.

**Comment by bazsi** [comment]
What is keeping this outside, any further ACKs needed? I'd love to see this merged.

**Comment by bkil-syslogng** [comment]
It also builds with GCC, but needs `DCMAKE_C_FLAGS="-Werror -std=gnu99"` for that. We should decide whether this should be passed along externally or it should be computed internally (with warning flags appended) as in `autotools`.
https://travis-ci.org/bkil-syslogng/syslog-ng/builds/131734211

I'll help resolving the conflict soon.

**Comment by bkil-syslogng** [comment]
@ihrwein to resolve the conflict:
* remove the commit which touches `.travis.yml` via `git rebasem -i master`
* cherry pick the most recent closing commit (the one touching `.travis.yml`) from https://github.com/bkil-syslogng/syslog-ng/pull/25/commits

**Comment by ihrwein** [comment]
I did an autosquash then replaced the conflicting commit with the proposed one.

**Comment by MrAnno** [comment]
Some naive benchmarks:

```sh
$ time (../configure --enable-debug && make -j)

real    1m36.925s
user    1m41.318s
sys     1m1.967s
```

--------------

```sh
$ time (cmake .. && make -j)

real    0m21.587s
user    0m29.846s
sys     0m8.007s
```

--------------

```sh
$ time (cmake -GNinja .. && ninja)

real    0m13.275s
user    0m18.463s
sys     0m5.256s
```

--------------

.travis.yml
```diff
@@ -1,24 +1,26 @@
 language: c
+git:
+  submodules: false
+
+env: B=autotools
+
 install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -qq curl
   - curl http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_12.04/Release.key | sudo apt-key add -
```
lbudai: Are you sure in that curl is installed by default?

bkil-syslogng: Otherwise command not found error would be found in the logs and the `apt-key add` would also fail. It may or may not be generally installed on desktop systems, but Travis is a development CI which comes pre-installed with various tools.

.travis.yml
```diff
@@ -1,24 +1,26 @@
...
   - echo "deb http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_12.04 ./" | sudo tee --append /etc/apt/sources.list.d/syslog-ng-obs.list
   - sudo apt-get update -qq
-  - sudo apt-get install -qq pkg-config flex bison xsltproc docbook-xsl libevtlog-dev libnet1-dev libglib2.0-dev libdbi0-dev libssl-dev libjson0-dev libwrap0-dev libpcre3-dev libcap-dev libesmtp-dev libgeoip-dev libhiredis-dev sqlite3 libdbd-sqlite3 libriemann-client-dev openjdk-7-jdk gradle-2.2.1 autoconf-archive
+  - sudo apt-get install -qq pkg-config flex bison xsltproc docbook-xsl libivykis-dev libevtlog-dev libnet1-dev libglib2.0-dev libdbi0-dev libjson0-dev libwrap0-dev libcap-dev libesmtp-dev libgeoip-dev libhiredis-dev sqlite3 libdbd-sqlite3 libriemann-client-dev gradle-2.2.1 autoconf-archive
```
lbudai: just a note : libivykis will be installed from my repo (as the version is 0.36.3-1, and in trusty it has the version 0.36.2-1).
I think it's not a problem, but it is good to know.

.travis.yml
lbudai: OK, but previously we had to install it and that time Travis was also a development CI and not a desktop system ;-)
( and travis.yml could document how to setup a  build environment to build syslog-ng from source even on a desktop system, but I'm ok with this change )

bkil-syslogng: I have also given this second though before. The OS packages should provide
good enough documentation about package requirements.

However, we could try to validate requirements via `strace` as a CI
enhancement if you see value in that.

.travis.yml
bkil-syslogng: Do note that many other requirements have also been removed from this long line. I wrapped it in my PRs, but didn't want to touch it here, although we could.

PR #1053, opened 2016-05-16
F/elasticsearch v2 mode http

lbudai:
ElasticSearchDestination-v2: HTTP mode added

    HTTP mode is implemented with the help of Jest
    , a Java HTTP Rest client for ElasticSearch.
    https://github.com/searchbox-io/Jest

    The implementation is partially based on
    https://github.com/balabit/syslog-ng/pull/1025 created by Viktor Tusa.

    Difference is that this implementation is not a separate driver
    but a mode-extension to the existing destination.

Note, that this client mode is able to work together even with ElasticSearch 5 (at least on my test environment).

  *  example config:
   ```
    @version: 3.8
    @include "scl.conf"

    source s_network {
        network(port(5555));
    };

    destination d_elastic {
        elasticsearch2(
            client_mode("http")
            cluster("es-syslog-ng")
            index("x201")
            cluster_url("http://es_admin:qqq123@192.168.33.10:9200")
            type("slng_test_type")
            flush_limit("0")
        );
    };

    log {
       source(s_network);
       destination(d_elastic);
       flags(flow-control);
    };
  ```

### DIFF
--- a/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/ElasticSearchDestination.java
+++ b/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/ElasticSearchDestination.java
@@ -26,12 +26,9 @@ package org.syslog_ng.elasticsearch_v2;
 import org.syslog_ng.LogMessage;
 import org.syslog_ng.LogTemplate;
 import org.syslog_ng.StructuredLogDestination;
-import org.syslog_ng.elasticsearch_v2.ElasticSearchOptions;
-import org.syslog_ng.elasticsearch_v2.client.ESClient;
+import org.syslog_ng.elasticsearch_v2.client.ESNativeClient;
 import org.syslog_ng.elasticsearch_v2.client.ESClientFactory;
 import org.syslog_ng.elasticsearch_v2.client.UnknownESClientModeException;
-import org.syslog_ng.elasticsearch_v2.messageprocessor.ESMessageProcessor;
-import org.syslog_ng.elasticsearch_v2.messageprocessor.ESMessageProcessorFactory;
 import org.syslog_ng.options.InvalidOptionException;
 import org.syslog_ng.logging.SyslogNgInternalLogger;
 import org.apache.log4j.Logger;
@@ -39,7 +36,7 @@ import org.elasticsearch.action.index.IndexRequest;
 
 public class ElasticSearchDestination extends StructuredLogDestination {
 
-	ESClient client;
+	ESNativeClient client;
 	ElasticSearchOptions options;
 	Logger logger;
 
--- a/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/ElasticSearchDestination.java
+++ b/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/ElasticSearchDestination.java
@@ -26,7 +26,7 @@ package org.syslog_ng.elasticsearch_v2;
 import org.syslog_ng.LogMessage;
 import org.syslog_ng.LogTemplate;
 import org.syslog_ng.StructuredLogDestination;
-import org.syslog_ng.elasticsearch_v2.client.ESNativeClient;
+import org.syslog_ng.elasticsearch_v2.client.ESClient;
 import org.syslog_ng.elasticsearch_v2.client.ESClientFactory;
 import org.syslog_ng.elasticsearch_v2.client.UnknownESClientModeException;
 import org.syslog_ng.elasticsearch_v2.messageprocessor.ESIndex;
@@ -36,7 +36,7 @@ import org.apache.log4j.Logger;
 
 public class ElasticSearchDestination extends StructuredLogDestination {
 
-	ESNativeClient client;
+	ESClient client;
 	ElasticSearchOptions options;
 	Logger logger;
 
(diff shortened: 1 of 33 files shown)


lbudai:
When this feature is merged, next step could be to provide a fallback mechanism to mode HTTP when other modes fail to work.

lbudai:
@talien : could you take a look at this PR?

Note, that there is a patch which removes the `cluster_url` option.
If you have a good reason why is it better than building the URL from the existing `server` and `port`, I'll drop that patch (this is why this change is in a separate patch :-) ).

faxm0dem:
how would you specify http or HTTPS, and auth?

lbudai:
@faxm0dem: as options, but we can also use the `cluster_url`...
Or, when user sets the url then that will be used as connection url, otherwise it will be built from server and port.

I'd go with 2nd: support custom `cluster_url` and when it is not specified use the server and port options.

lbudai:
@faxm0dem : code has been modified, re-added the `cluster_url` option (when it is not presented, then the server/port will be used). 

@juhaszviktor , @pzoleex : could you review and merge this PR?

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/ESTransportClient.java
```diff
@@ -63,6 +69,13 @@ protected Client createClient(List<Class<? extends Plugin>> plugins) {
 		transportClient = transportClientBuilder.build();
 		
 		addServersToClient(transportClient);
+
```
ihrwein: We will print (almost) the same message two times.

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/ESClient.java
```diff
@@ -30,14 +30,18 @@
...
 import org.elasticsearch.cluster.health.ClusterHealthStatus;
 import org.elasticsearch.common.settings.SettingsException;
 import org.elasticsearch.common.settings.Settings.Builder;
 import org.syslog_ng.elasticsearch_v2.ElasticSearchOptions;
+import org.syslog_ng.elasticsearch_v2.messageprocessor.ESMessageProcessor;
+import org.syslog_ng.elasticsearch_v2.messageprocessor.ESMessageProcessorFactory;
 
 public abstract class ESClient {
 	private Client client;
+    private ESMessageProcessor messageProcessor;
```
ihrwein: Just a nitpicking, but the whitespace handling is inconsistent with the original content. This is just a FYI to avoid this problem in the future, I don't care much about it if the code is good :)

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/ElasticSearchDestination.java
```diff
@@ -76,13 +76,14 @@ protected boolean open() {
...
-    private IndexRequest createIndexRequest(LogMessage msg) {
+    private ESIndex createIndexRequest(LogMessage msg) {
     	String formattedMessage = options.getMessageTemplate().getResolvedString(msg, getTemplateOptionsHandle(), LogTemplate.LTZ_SEND);
     	String customId = options.getCustomId().getResolvedString(msg, getTemplateOptionsHandle(), LogTemplate.LTZ_SEND);
     	String index = options.getIndex().getResolvedString(msg, getTemplateOptionsHandle(), LogTemplate.LTZ_SEND);
     	String type = options.getType().getResolvedString(msg, getTemplateOptionsHandle(), LogTemplate.LTZ_SEND);
     	logger.debug("Outgoing log entry, json='" + formattedMessage + "'");
-    	return new IndexRequest(index, type, customId).source(formattedMessage);
+
+		return new ESIndex.Builder().formattedMessage(formattedMessage).index(index).id(customId).type(type).build();
```
ihrwein: :+1:

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/ESClient.java
```diff
@@ -0,0 +1,7 @@
+package org.syslog_ng.elasticsearch_v2.client;
+
+/**
+ * Created by stentor on 2016.05.13..
```
ihrwein: I'm not sure you wanted to include these lines.

ihrwein: Nevermind, it's removed in a followup commit.

lbudai: Yeah, it was accidentally.

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/ESClient.java
ihrwein: Nevermind, It is fixed by another commit.

ihrwein:
Don't forget to extend the CMakeLists.txt file in `modules/java-modules`. Otherwise, :+1:

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/ESTransportClient.java
lbudai: Yeah, it strange because I factored out these lines to a method(with an IDE...) .
I'll fix this.

ihrwein:
I created https://github.com/balabit/syslog-ng/issues/1095, this is ready for merge.

lbudai:
Rebased against master, so someone could push that merge button...

PR #1058, opened 2016-05-19
Fix a linker related problem in systemd-journal's tests

ihrwein:
The ivykis library wasn't linked to the test binary.

  CCLD     modules/systemd-journal/tests/test_systemd_journal
/usr/bin/ld: modules/systemd-journal/tests/modules_systemd_journal_tests_test_systemd_journal-test_systemd_journal.o: undefined reference to symbol 'iv_task_registered@@IVYKIS_0.29'
/tmp/pe7-tibi/install/lib/libivykis.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[3]: *** [modules/systemd-journal/tests/test_systemd_journal] Error 1
make[2]: *** [check-am] Error 2
make[1]: *** [check-recursive] Error 1
make: *** [check] Error 2

### DIFF
--- a/modules/systemd-journal/tests/Makefile.am
+++ b/modules/systemd-journal/tests/Makefile.am
@@ -11,5 +11,5 @@ modules_systemd_journal_tests_test_systemd_journal_SOURCES = \
 	modules/systemd-journal/tests/test-source.c
 
 modules_systemd_journal_tests_test_systemd_journal_CFLAGS = $(TEST_CFLAGS) -I$(top_srcdir)/modules/systemd-journal
-modules_systemd_journal_tests_test_systemd_journal_LDADD = $(TEST_LDADD)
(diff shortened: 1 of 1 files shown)


bkil-syslogng:
I don't see where the test references the module to be tested.

(Also note that `iv_task_registered` is called from `modules/systemd-journal/journal-reader.c` and some places from `lib` as well.)

bkil-syslogng:
:-1: I will update this thumb after addressing the issues.

bazsi:
hmm.. I don't see code changes in this pull request, but as it seems it fixes an issue seen in the wild. Is this really worth postponing? I agree that the journal related tests are not the best ones though, at least if this is what you imply in your comment.

So I would merge this if it indeed fixes the issue. I guess the reason that it fails sometimes that ivykis is sometimes embedded into libsyslog-ng.so in other cases it isn't.

lbudai:
@bkil-syslogng:
As this PR solves a problem that we are blocking on, I'll merge this.

PR #1059, opened 2016-05-19
F/dbparser pdbload refact

lbudai:
This PR contains only equivalent changes.
Goal of this PR was to eliminate those huge switch-cases.

### DIFF
--- a/modules/dbparser/pdb-load.c
+++ b/modules/dbparser/pdb-load.c
@@ -1024,83 +1024,139 @@ pdb_loader_end_element(GMarkupParseContext *context, const gchar *element_name, 
     }
 }
 
+static gboolean
+_pdbl_ruleset_pattern_text(PDBLoader *state, const gchar *text, gsize text_len, GError **error)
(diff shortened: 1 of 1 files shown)


bazsi:
I like where this is going, looks a lot better. We might as well eliminate the braces  where single statements are embedded in if/else that would make the code even more concise.

bazsi:
:+1:

modules/dbparser/pdb-load.c
```diff
@@ -93,21 +98,35 @@ typedef struct _PDBProgramPattern
   PDBRule *rule;
 } PDBProgramPattern;
 
+
+static void
+_pdb_state_stack_push(PDBStateStack *self, gint state)
+{
+  g_assert(self->top < PDB_STATE_STACK_MAX_DEPTH - 1);
```
ihrwein: :+1: for the bounds checking with abort

ihrwein:
:+1:

PR #1069, opened 2016-05-27
cfg-tree: fix a potential crash in reload logic

### LINKED ISSUE #1065: Aborting during configuration loading
I found a nasty bug in the configuration loading process.

It can be reproduced by following these steps:

1. start syslog-ng with the following (valid) configuration:

 ```
@version: 3.8
@include "scl.conf"

log {
  source {
    file("/tmp/a.txt");
  };
  destination {
    file("/tmp/a.txt");
  };
};
```

1. Replace the configuration with the following one:
 ```
@version: 3.8
@include "scl.conf"

log {
  source {
    # syntactically valid, but semantically invalid -> init() returns FALSE
    file("/tmp/a.txt" multi-line-prefix("a"));
  };
  destination {
    file("/tmp/a.txt");
    };
};
```

1. Reload syslog-ng
 ```
sbin/syslog-ng-ctl reload
```

Expected result:
```
[2016-05-26T15:29:21.460108] multi-line-prefix() and/or multi-line-garbage() specified but multi-line-mode() is not regexp based (prefix-garbage or prefix-suffix), please set multi-line-mode() properly;
[2016-05-26T15:29:21.460116] Error initializing message pipeline;
[2016-05-26T15:29:21.460123] Error initializing new configuration, reverting to old config;
[2016-05-26T15:29:21.460157] Unable to detect fully qualified hostname for localhost, use_fqdn() will use the short hostname;
[2016-05-26T15:29:21.460168] Compiling #unnamed sequence [log] at [/home/btibi/work/impl-deinit-for-native-parser/install/etc/syslog-ng.conf:15:2]
[2016-05-26T15:29:21.460176]   Compiling #anon-source0 sequence [source] at [/home/btibi/work/impl-deinit-for-native-parser/install/etc/syslog-ng.conf:15:2]
[2016-05-26T15:29:21.460183]     Compiling #unnamed junction [log] at [/home/btibi/work/impl-deinit-for-native-parser/install/etc/syslog-ng.conf:15:10]
[2016-05-26T15:29:21.460191]       Compiling #unnamed single [log] at [/home/btibi/work/impl-deinit-for-native-parser/install/etc/syslog-ng.conf:17:3]
[2016-05-26T15:29:21.460200] Error cloning pipe into its reference point, probably the element in question is not meant to be used in this situation; location='/home/btibi/work/impl-deinit-for-native-par$er/install/etc/syslog-ng.conf:17:3'
Quit (core dumped)
```

This problem can be reproduced with plugins whose `init()` method can return `FALSE`. `affile` and the "native" parsers are certainly affected. They are needed to trigger a config reverting.

We need an other plugin which cannot be cloned/`clone()` could fail. File destinations cannot be cloned, so they are perfect for this purpose.

What happens during the reproduction?

1. original configuration is loaded, everything is fine
 1. `PIF_INLINED` flag is set on the `LogPipe` instances
2. we replace the configuration
3. syslog-ng is reloaded
 1. new configuration is created
 1. old configuration is deinitialized
 1. `PIF_INLINED` flag is still set on the `LogPipe` instance
 1. new configuration is initialized
   * it will fail, because the configuration is invalid
 1. new configuration is deinitialized
 1. old configuration is initialized
   * as `PIF_INLINED` is set, we try to clone the pipe -> it will fail https://github.com/balabit/syslog-ng/blob/master/lib/cfg-tree.c#L461-L474
   * we return with FALSE which will trigger a `kill()` call https://github.com/balabit/syslog-ng/blob/master/lib/mainloop.c#L217-L223

syslog-ng `3.8` is certainly affected and by looking at the code `3.6` and `3.7` as well.

bazsi:
this branch fixes #1065 

cfg_tree_compile() is not idempotent for at least two reasons:
  * it stores the pipes to be initialized in an array, which will grow
    for each cfg_tree_compile() calls
  * it uses PIF_INLINED to indicate if a LogPipe was pasted directly into
    the config rather than cloned

But cfg_tree_start() invokes cfg_tree_compile() unconditionally, which
means that if a config is initialized multiple times, cfg_tree_compile()
will accumulate cruft in its state.

Issue #1065 on github explains one case where syslog-ng crashes on reload:
https://github.com/balabit/syslog-ng/issues/1065

This patch fixes this problem by guarding the compile function with
a variable, which ensures that the body of the function is executed only
once.

### DIFF
--- a/lib/cfg-tree.c
+++ b/lib/cfg-tree.c
@@ -1036,6 +1036,8 @@ cfg_tree_compile(CfgTree *self)
   gint i;
 
   /* resolve references within the configuration */
+  if (self->compiled)
+    return TRUE;
 
   for (i = 0; i < self->rules->len; i++)
     {
@@ -1054,6 +1056,7 @@ cfg_tree_compile(CfgTree *self)
           return FALSE;
         }
     }
+  self->compiled = TRUE;
   return TRUE;
 }
 
--- a/lib/cfg-tree.c
+++ b/lib/cfg-tree.c
@@ -1105,6 +1105,7 @@ cfg_tree_stop(CfgTree *self)
 void
 cfg_tree_init_instance(CfgTree *self, GlobalConfig *cfg)
 {
+  memset(self, 0, sizeof(*self));
   self->initialized_pipes = g_ptr_array_new();
   self->objects = g_hash_table_new_full(cfg_tree_objects_hash, cfg_tree_objects_equal, NULL, (GDestroyNotify) log_expr_node_free);
   self->templates = g_hash_table_new_full(g_str_hash, g_str_equal, NULL, (GDestroyNotify) log_template_unref);
--- a/lib/cfg-tree.h
+++ b/lib/cfg-tree.h
@@ -157,6 +157,7 @@ typedef struct _CfgTree
   /* list of top-level rules */
   GPtrArray *rules;
   GHashTable *templates;
+  gboolean compiled;
 } CfgTree;
 
 gboolean cfg_tree_add_object(CfgTree *self, LogExprNode *rule);


ihrwein:
:+1: thanks for the fix! We'll backport it once it's merged.

bazsi:
Merging based on the review of @ihrwein.

I also noticed the failing unit test which was fixed in the second commit. So it is all green now

PR #1070, opened 2016-05-27
F/fix make func-test typos that keeps us from enabling it in CI

bkil-syslogng:
These are the core fixes that are needed to enable `func_test` in Travis. As enabling all tests in Travis leads to a longer discussion, I have separated the fixes themselves here. They are mostly typos and version incompatibilities.

https://github.com/balabit/syslog-ng/pull/982 can be merged on top of this one.

I will do some quick adjustments now, but after this one gets a green light, it is ready for review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,27 +8,68 @@ install:
   - curl http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_12.04/Release.key | sudo apt-key add -
   - echo "deb http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_12.04 ./" | sudo tee --append /etc/apt/sources.list.d/syslog-ng-obs.list
   - sudo apt-get update -qq
-  - sudo apt-get install -qq pkg-config flex bison xsltproc docbook-xsl libivykis-dev libevtlog-dev libnet1-dev libglib2.0-dev libdbi0-dev libjson0-dev libwrap0-dev libcap-dev libesmtp-dev libgeoip-dev libhiredis-dev sqlite3 libdbd-sqlite3 libriemann-client-dev gradle-2.2.1 autoconf-archive
+  - sudo apt-get install -qq
+      autoconf-archive
+      bison
+      docbook-xsl
+      flex
+      gradle-2.2.1
+      libcap-dev
+      libdbd-sqlite3
+      libdbi0-dev
+      libesmtp-dev
+      libevtlog-dev
+      libgeoip-dev
+      libglib2.0-dev
+      libhiredis-dev
+      libivykis-dev
+      libjson0-dev
+      libnet1-dev
+      libriemann-client-dev
+      libwrap0-dev
+      pkg-config
+      sqlite3
+      xsltproc
   - sudo pip install -r requirements.txt
 before_script:
   - ./autogen.sh
   - unset PYTHON_CFLAGS # HACK
-  - ./configure CFLAGS=-Werror --with-ivykis=internal --prefix=$HOME/install/syslog-ng --enable-pacct --enable-manpages --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl --enable-extra-warnings
-  - . tests/build-log-cflags-propagation.sh;
-    exec_prop_check "make -j V=1 --keep-going";
-    S=$?;
-    if [ "$S" = "42" -o "$S" = "0" ]; then
-      return $S;
-    else
-      make V=1; false;
-    fi # to make error messages more readable on error
+  - ./configure
+      CFLAGS=-Werror
+      --prefix=$HOME/install/syslog-ng
+      --with-ivykis=internal
+      --with-mongoc=internal
+      --with-librabbitmq=internal
+      --with-jsonc=system
+      --disable-env-wrapper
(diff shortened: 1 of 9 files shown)


bazsi:
Lgtm.

I have some doubts about using rpath to locate libjvm.so, because rpath is strongly discouraged by both Debian and fedora, but I couldn't find a better solution within a few minutes of googling.

bazsi:
:+1:

m4/check_java_support.m4
```diff
@@ -83,7 +83,7 @@ AC_DEFUN([AX_CHECK_JAVA_VERSION],
       JNI_LIBDIR=`find $JNI_HOME \( -name "libjvm.so" -or -name "libjvm.dylib" \) \
         | sed "s-/libjvm\.so-/-" \
         | sed "s-/libjvm\.dylib-/-" | head -n 1`
-      JNI_LIBS="-L$JNI_LIBDIR -ljvm"
+      JNI_LIBS="-L$JNI_LIBDIR -ljvm -Wl,-rpath=$JNI_LIBDIR"
```
ihrwein: Why `ldconfig` or `LD_LIBRARY_PATH` are not enough for a CI system?

modules/systemd-journal/systemd-journal-grammar.ym
```diff
@@ -101,7 +101,7 @@ source_systemd_journal_option
     }
   | KW_TIME_ZONE '(' string ')'
     {
-      if (last_journal_reader_options->recv_time_zone);
```
ihrwein: :+1: (did clang complain about it? :))

bkil-syslogng: It was just the propagation of flags and some other configuration fixes that brought this up via a warning.

bkil-syslogng:
Actually I would be good to use RUNPATH as well, but I didn't find a quick
way to convince the linker to use RUNPATH for libjvm, but RPATH for the
other `.so`.

On Sun, May 29, 2016 at 10:59 PM, Balazs Scheidler <notifications@github.com

m4/check_java_support.m4
bkil-syslogng: System-wide `ldconfig` requires root and we'd like to avoid that to support
containerized CI.

`LD_LIBRARY_PATH` needs some extra effort in touching some of our other
scripts, and I was hoping that fixing `RPATH`/`RUNPATH` could also spare
some of our users of having to specify LD_LIBRARY_PATH themselves if they
use OS packages. I would strongly recommend that we use RUNPATH if you know
how we can do that individually, though.

.travis.yml
```diff
@@ -8,27 +8,68 @@ install:
...
+    if [ "$CC" = "gcc" ]; then
       export DISTCHECK_CONFIGURE_FLAGS="CFLAGS=-Werror --enable-extra-warnings";
       exec_prop_check "make distcheck V=1 --keep-going";
     else
-      make func-test V=1;
+      exec_prop_check "make -j V=1 --keep-going install";
+      S=$?;
+      if [ "$S" = "0" ]; then
+        . scripts/get-libjvm-path.sh || return $?;
+        export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$JNI_LIBDIR";
```
bkil-syslogng: I've removed the `RPATH` modification and exchanged it for a lookup script to set `LD_LIBRARY_PATH`:
https://github.com/balabit/syslog-ng/pull/1070/commits/7081ffb1e912573bccc993c7abb9c250632dbe3a

bkil-syslogng:
I've cherry picked two new improvements from the full matrix:
https://github.com/balabit/syslog-ng/pull/1070/commits/028902926177eaee0d487f9ddba91f5010011934
https://github.com/balabit/syslog-ng/pull/1070/commits/927e19af8373e06a8d9d7dc69ebaad6af0fbadff

modules/afsql/afsql.h
```diff
@@ -28,7 +28,7 @@
 #include "mainloop-worker.h"
 #include "string-list.h"
 
-#include <dbi/dbi.h>
+#include <dbi.h>
```
bazsi: This seems to have been the wrong fix. LIBDBI_CFLAGS is set to /usr/include in my case (e.g. it doesn't contain the dbi subdirectory), which comes from the .pc file installed by libdbi.

Our configure script however encodes /usr/include/dbi in case the pc file is not found, which certainly needs this change.

I think we should have changed configure.ac to match the pc file. I'll create a patch within my current branch to do that.

bkil-syslogng: This fix is valid on both Ubuntu Trusty and Xenial:

`pkg-config --cflags dbi`

-I/usr/include/dbi -I/usr/include/dbi/dbi

Please open a new issue if you see a problem on a specific config and share
details on the config on which this does not work as intended.

bazsi: I am using upstream libdbi, so this might be a patch in Ubuntu then.

I'll think it over how we can resolve my scenario then.

bkil-syslogng: This is building upstream libdbi as well, installs it locally and works
with the above patch:
https://github.com/balabit/syslog-ng/pull/982
It uses libdbi-0.9.0 branch, though.

On Tue, Jun 21, 2016 at 10:52 PM, Balazs Scheidler <notifications@github.com

PR #1075, opened 2016-06-02
Create libsyslog-ng.so symlink on installation with CMake

ihrwein:
This is a bugfix commit for #1051 

CMake didn't create the `libsyslog-ng.so` symlink during installation so the linker wasn't able to find syslog-ng with `-lsyslog-ng`.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -279,7 +279,7 @@ set_target_properties(syslog-ng
     PROPERTIES VERSION ${SYSLOG_NG_VERSION}
     SOVERSION ${SYSLOG_NG_VERSION})
 
-install(TARGETS syslog-ng LIBRARY DESTINATION lib NAMELINK_SKIP)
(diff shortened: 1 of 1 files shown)


bkil-syslogng:
You forgot to cherry pick `travis_retry`

ihrwein:
Now it's green.

(I suppose you referred `travis_retry` as a solution to a failing apt-get execution.)

bkil-syslogng:
:+1: This is an improvement.

Though, could you tell us why not `NAMELINK_ONLY`?
https://cmake.org/cmake/help/v3.2/command/install.html#installing-targets

ihrwein:
@bkil-syslogng I like to see the version in the filename:

```
lrwxrwxrwx 1 btibi btibi   26 May 25 14:41 libsyslog-ng.so -> libsyslog-ng.so.3.8.alpha0
-rw-r--r-- 1 btibi btibi 2.2M May 27 11:04 libsyslog-ng.so.3.8.alpha0
```

PR #1079, opened 2016-06-08
F/kv parser robustness improvements, drop empty keys, better test coverage

### LINKED ISSUE #1072: Minor problems in kv-parser()
Hi,

I wrote a couple of edge-case unit tests for `kv-parser()`, some of them have produced questionable results.

- value-separator alone creates an empty key-value pair:
```C
static void
test_kv_scanner_stray_words_are_ignored(void)
{
  kv_scanner_input(kv_scanner, "lorem ipsum foo=bar");
  assert_next_kv_is("foo", "bar");
  assert_no_more_tokens();
// ...
  kv_scanner_input(kv_scanner, "lorem ipsum = foo=bar");
  assert_next_kv_is("", ""); // is it OK?
  assert_next_kv_is("foo", "bar");
  assert_no_more_tokens();
// ...
}
```

---------------

- a comma separated value without a space behaves differently:
```C
  kv_scanner_input(kv_scanner, "key1=value1,key2=value2,key3=value3");
  assert_next_kv_is("key1", "value1,key2=value2,key3=value3");
  assert_no_more_tokens();
```

---------------

- tabulator as a whitespace character is not working properly:
```C
  kv_scanner_input(kv_scanner, "key1=value1    key2=value2,     key3=value3");
  // key=value1 <tab> key2=value2, <space> <tab> key3=value3
  assert_next_kv_is("key1", "value1    key2=value2");
  assert_next_kv_is("key3", "value3");
  assert_no_more_tokens();
```

bkil-syslogng:
Fixes #1072

### DIFF
--- a/modules/kvformat/kv-parser.c
+++ b/modules/kvformat/kv-parser.c
@@ -112,6 +112,17 @@ kv_parser_free(LogPipe *s)
   log_parser_free_method(s);
 }
 
+static gboolean
+kv_parser_process_threaded(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_options, const gchar *input, gsize input_len)
+{
+  LogParser *self = (LogParser *)log_pipe_clone(&s->super);
+
+  gboolean ok = kv_parser_process(self, pmsg, path_options, input, input_len);
+
+  log_pipe_unref(&self->super);
+  return ok;
+}
+
 LogParser *
 kv_parser_new(GlobalConfig *cfg, KVScanner *kv_scanner)
 {
@@ -120,7 +131,7 @@ kv_parser_new(GlobalConfig *cfg, KVScanner *kv_scanner)
   log_parser_init_instance(&self->super, cfg);
   self->super.super.free_fn = kv_parser_free;
   self->super.super.clone = kv_parser_clone;
-  self->super.process = kv_parser_process;
+  self->super.process = kv_parser_process_threaded;
 
   self->kv_scanner = kv_scanner;
   self->formatted_key = g_string_sized_new(32);
--- a/modules/kvformat/tests/test_kv_scanner.c
+++ b/modules/kvformat/tests/test_kv_scanner.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015 Balabit
+ * Copyright (c) 2015-2016 Balabit
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU General Public License version 2 as published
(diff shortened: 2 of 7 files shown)


bkil-syslogng:
Rebased to master. Data race fixes have been cherry picked on top. A leak fix is still expected by tomorrow, which will also be cherry picked on top, so reviewing the previous parts is still valid.

modules/kvformat/tests/test_kv_scanner.c
```diff
@@ -20,196 +20,302 @@
...
+  kv_scanner_input(scanner, input);
+  gboolean expect_more = TRUE;
+  while (fn(scanner, args, &expect_more))
+    {
+    }
+  if (expect_more)
+    _assert_no_more_tokens(scanner);
 }
 
+#define _INVOKE_VA(args, code) \
```
kvch: `_INVOKE_VA(args, code, input)`

kvch:
After the small mistake is fixed, this can be merged. :+1:

bkil-syslogng:
Fixed comments. At first sight, the mentioned memory leak doesn't apply to this version of kv-parser, so this PR seems final. I'll proceed with investigation on a different track tomorrow, though.

bkil-syslogng:
Another `kv-parser` race fixed. The `GString *formatted_key` scratch buffer was also shared. I tried to condition the fixes on whether threading is enabled or not to not cause a speed regression.

modules/kvformat/kv-parser.c
```diff
@@ -112,6 +112,28 @@ kv_parser_free(LogPipe *s)
...
+
+  kv_parser_free(&self->super);
+  return ok;
+}
+
+static gboolean
+kv_parser_process_init(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_options, const gchar *input, gsize input_len)
+{
+  if (s->super.cfg && s->super.cfg->threaded)
+    s->process = kv_parser_process_threaded;
```
bazsi: there are other ways to initiate threading (the threaded flag on an input driver), so this is not really safe.

And I would try to avoid any conditionals in driver code if at all possible.

Can't we avoid malloc/free by using a local variable when cloning? But you know what, only a small fraction of the input will be parsed by kv-parser(), so for now let's stick to the clone() stuff, always. We'll see if there are bottlenecks in the area. If there is, I would probably use a pool of KVScanner instances, one dedicated to all threads.

So I think the conditions on threaded should just be reverted.

bazsi:
With the one comment on the conditionals, this could go in.
:+1:

bkil-syslogng:
Thank you for your time. I've now removed the check and retained the threaded version only.

While the simple `kv_scanner_clone()` solution looked sane of first glance, unfortunately it would introduce a memory leak which would have been caught by https://github.com/balabit/syslog-ng/pull/1080
I've adjusted it to use `log_pipe_clone()` and `log_pipe_unref()` instead.

Can we now merge this with these two +1's?

[The previous fix proposals used local variables (and sprinkled conditioning) instead. I've opted for this cloning solution now because it is not much slower, while being much easier to understand and more difficult to mess up in the future.]

Backup saved:
* https://github.com/balabit/syslog-ng/compare/master...bkil-syslogng:t/kv-parser-robustness-bk3?expand=1

PR #1084, opened 2016-06-11
Fix incorrectly checked integer underflow/overflow

### LINKED ISSUE #1083: basicfuncs unit test seems to be buggy on 32bit systems
https://build.opensuse.org/package/live_build_log/home:laszlo_budai:syslog-ng/syslog-ng-3.8alpha/xUbuntu_15.10/i586

( I don't have the time for investigating this right now, so I'm opening this issue. )
Maybe there is a huge number in the tests? Recently we added some numerical functions...

MrAnno:
From manpage:

> If an underflow occurs, strtol() returns LONG_MIN. If an overflow occurs, strtol() returns LONG_MAX. In both cases, errno is set to ERANGE. Precisely the same holds for strtoll() (with LLONG_MIN and LLONG_MAX instead of LONG_MIN and LONG_MAX).

>The implementation may also set errno to EINVAL in case no conversion was performed (no digits seen, and 0 returned).

Fixes #1083

### DIFF
--- a/lib/parse-number.c
+++ b/lib/parse-number.c
@@ -151,8 +151,7 @@ _parse_number(const gchar *s, gchar **endptr, gint64 *d)
   errno = 0;
   val = strtoll(s, endptr, 10);
 
-  if ((errno == ERANGE && (val == LLONG_MAX || val == LLONG_MIN))
-      || (errno != 0 && val == 0))
+  if (errno == ERANGE || errno == EINVAL)
     return FALSE;
 
   if (*endptr == s)
--- a/lib/parse-number.c
+++ b/lib/parse-number.c
@@ -151,7 +151,7 @@ _parse_number(const gchar *s, gchar **endptr, gint64 *d)
   errno = 0;
   val = strtoll(s, endptr, 10);
 
-  if ((errno == ERANGE && (val == LONG_MAX || val == LONG_MIN))
+  if ((errno == ERANGE && (val == LLONG_MAX || val == LLONG_MIN))
       || (errno != 0 && val == 0))
     return FALSE;
 


juhaszviktor:
:+1:

lbudai:
Thanks!

bazsi:
I had just one question in the issue and not here at the PR. Why do we need to check on LLONG_MIN/LLONG_MAX ?

isn't it enough to check for ERANGE?

lbudai:
@bazsi : that check was removed in the second commit (and a new check also introduced by that commit: EINVAL). However, it would be nice to have an answer ( @MrAnno ? :-)  )

bkil-syslogng:
An even better question would be why not simply use `if (errno) return FALSE;` ? Although, you've covered all errors listed in the man page, this would communicate 'nothing up my sleeves' for future contributors.

MrAnno:
@bazsi , @lbudai: Thanks for pointing this out, I updated the pull request and also removed the `(errno != 0 && val == 0)` condition.

`strtoll()` sets errno to ERANGE when an underflow/overflow occurs and may set errno to EINVAL *in case no conversion was performed*. 

@bkil-syslogng: I thought it's better to explicitly mention these error codes, but I'm absolutely OK with the `if (errno)` version as well. :spades:

lbudai:
@MrAnno : I'm fine with the current patch-set.
so, :+1: 
@bazsi : ?

bkil-syslogng:
My question stemmed from what I said above and this note in the man page:

> the calling program should set errno to 0 before the call, and then determine if an error occurred by checking whether errno has a nonzero value after the call.
http://man7.org/linux/man-pages/man3/strtol.3.html#NOTES

bkil-syslogng:
:+1: I'm merging this because it has +3, it fixes a bug, is clean enough and does not regress code.

bazsi:
Thanks, obviously no objections from my part. :)

:+1:

PR #1085, opened 2016-06-11
Remove unused syslog-ng.conf file

MrAnno:
I couldn't find any reference to this file.


bazsi:
It has definitely no place in lib.

PR #1086, opened 2016-06-13
logmsg: don't release "original" from the cleared message

bazsi:
This patch fixes a potential SIGSEGV for code that uses log_msg_clear()
in production (e.g. syslog-parser()).

The role of log_msg_clear() is to clear up all name-value pairs,
SDATA, tags and sockaddr in order to reset the message into its
default state.

In this function we also relinquished self->original, which points
to the message where we cloned the current one from.

The problem with this was that it actually breaks the acknowledgement
logic, as ack_func was not reset at the same time.

It would be a layering violation to reset it back to its original value,
I simply remove dropping the link to self->original, which is only going to
be used for acks and nothing really else.

### DIFF
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -1050,11 +1050,6 @@ log_msg_clear(LogMessage *self)
     }
   self->saddr = NULL;
 
-  if (self->original)
-    {
-      log_msg_unref(self->original);
-      self->original = NULL;
-    }
   self->flags |= LF_STATE_OWN_MASK;
 }
 


bazsi:
This fixes a crash reproducible always, if there's a syslog-parser() and an error occurs in parsing, so this should be merged ASAP. Any takers on reviewing it?

lbudai:
@bazsi : in progress

lbudai:
@bazsi: so when we cannot parse the message ( `msg_format_inject_parse_error()` ),
we clear the related LogMessage instance  and the problem was that when it is ACKed then `self->original` was NULL (and it crashed when the  `self[NULL]->ack_func` was called in `log_msg_ack`). 

One question here: so in this case we want to ACK the message that we cannot parse (and the answer is yes, as we can forward it even if we cannot parse...).

bazsi:
-- 
Bazsi

On Thu, Jun 16, 2016 at 11:21 AM, Budai Laszlo <notifications@github.com>
wrote:

> @bazsi <https://github.com/bazsi>: so when we cannot parse the message (
> msg_format_inject_parse_error() ),
> we clear the related LogMessage instance and the problem was that when it
> is ACKed then self->original was NULL (and it crashed when the
> self[NULL]->ack_func was called in log_msg_ack).
>
> One question here: so in this case we want to ACK the message that we
> cannot parse (and the answer is yes, as we can forward it even if we cannot
> parse...).
>
> yes, we need to ACK it, as its processing has succeeded, and otherwise we
would block a slot in the input window. So the ack is definitely needed.

lbudai:
@bazsi : yes, agree

Issue #999, opened 2016-03-24
format-json produces invalid json


bobrik:
syslog-ng 3.7.2, template looks like this:

```
$(format-json @timestamp=\"$ISODATE\" logsource=\"$HOST\" program=\"$PROGRAM\" message=\"$MESSAGE\" app=\"something\")
```

This produces the following JSON messages (one per line):

``` json
{"program":"ÿ¿¿¿¿¾3m5","logsource":"a.b.c","app":"nginx","@timestamp":"2016-03-24T16:03:24+00
{"program":"mcmd\\xf4\\xf5\\xefά","logsource":"a.b.c","app":"nginx","@timestamp":"2016-03-24T16:03:24+00:00"}
```

Might be somehow connected to #917, as it is observed in the same installation.

bkil-syslogng:
Could you please test with 3.7.3 or git master as well?

bobrik:
3.7.3 has the same issue.

bkil-syslogng:
I've tested the given template format both from the command line and in a sterile unit testing environment and it works for me in both cases.

We would need more information on the kind of input you feed this template with and a full configuration file as well. It would help a lot if you could minimize them.

bobrik:
Main pipeline sends logs to kafka, but I also steered logs from problematic host to output JSON on stdout too:

```
destination d_debug { file("/dev/stdout" template("$(format-json @timestamp=\"$ISODATE\" logsource=\"$HOST\" program=\"$PROGRAM\" message=\"$MESSAGE\" app=\"nginx\")\n")); };

filter f_hehe { host("hehe"); };

log {
    source(s_net);
    filter(f_hehe);
    destination(d_debug);
};
```

This does not reveal messages with truncated JSON.

Kafka:

```
{"program":"nginx","message":"\\xafcac\\xfb\\xb2\\xcdά","logsource":"hehe","app":"nginx","@timestamp":"2016-04-02T15:35:00+00:00"}
{"program":"nginx","message":"ÿ¿¿¿¿¾\\xfb\\xb2\\xcdά","logsource":"hehe","app":"nginx","@timestamp":"2016-04-02T15:35:00+00
```

Stdout:

```
2016-04-02T18:01:06.000 myhost {"program":"nginx","message":"\\xafcac\\xfb\\xb2\\xcdά","logsource":"hehe","app":"nginx","@timestamp":"2016-04-02T15:35:00+00:00"}
[...]

PR #1088, opened 2016-06-13
Fix invalid handling of UTF-8 strings

MrAnno:
This issue can cause invalid UTF-8 characters in `format-cef-extension`, `format-welf` and `format-json`.

`g_utf8_get_char_validated()` converts a sequence of bytes encoded as UTF-8 to a Unicode character (`gunichar` --> UTF-32), so checking for an UTF-8 leading byte (`< 0xc0`) is not an option here.
`0x80` can be used instead.

This is a performance optimization, where we can append the given unichar with `g_string_append_c()` if it fits in 7 bits.
In all other cases, conversion is required (`g_string_append_unichar()`).

Partial fix for #999

### DIFF
--- a/lib/utf8utils.c
+++ b/lib/utf8utils.c
@@ -39,7 +39,7 @@ _is_character_unsafe(gunichar uchar, const gchar *unsafe_chars)
 static inline void
 _append_unichar(GString *string, gunichar wc)
 {
-  if (wc < 0xc0)
+  if (wc < 0x80)
     g_string_append_c(string, (gchar) wc);
   else
     g_string_append_unichar(string, wc);
--- a/lib/utf8utils.c
+++ b/lib/utf8utils.c
@@ -36,15 +36,6 @@ _is_character_unsafe(gunichar uchar, const gchar *unsafe_chars)
   return _strchr_optimized_for_single_char_haystack(unsafe_chars, (gchar) uchar) != NULL;
 }
 
-static inline void
-_append_unichar(GString *string, gunichar wc)
-{
-  if (wc < 0x80)
-    g_string_append_c(string, (gchar) wc);
-  else
-    g_string_append_unichar(string, wc);
-}
-
 /**
  * This function escapes an unsanitized input (e.g. that can contain binary
  * characters, and produces an escaped format that can be deescaped in need,
@@ -102,7 +93,7 @@ _append_escaped_utf8_character(GString *escaped_output, const gchar **raw,
         else if (_is_character_unsafe(uchar, unsafe_chars))
           g_string_append_printf(escaped_output, "\\%c", (gchar) uchar);
         else
-          _append_unichar(escaped_output, uchar);
+          g_string_append_unichar_optimized(escaped_output, uchar);
         break;
     }
   *raw = g_utf8_next_char(char_ptr);
--- a/lib/str-utils.h
+++ b/lib/str-utils.h
@@ -32,6 +32,16 @@
 GString *g_string_assign_len(GString *s, const gchar *val, gint len);
 void g_string_steal(GString *s);
 
+static inline GString *
+g_string_append_unichar_optimized(GString *string, gunichar wc)
+{
+  if (wc < 0x80)
+    g_string_append_c(string, (gchar) wc);
+  else
+    g_string_append_unichar(string, wc);
+
+  return string;
+}
 
 /*
  * DO NOT USE THIS MACRO UNLESS STRICTLY NECESSARY FOR PERFORMANCE REASONS.
--- a/modules/cef/format-cef-extension.c
+++ b/modules/cef/format-cef-extension.c
@@ -66,7 +66,7 @@ tf_cef_is_valid_key(const gchar *str)
 static inline void
 tf_cef_append_unichar(GString *string, gunichar wc)
 {
-  if (wc < 0xc0)
+  if (wc < 0x80)
     g_string_append_c(string, (gchar) wc);
   else
     g_string_append_unichar(string, wc);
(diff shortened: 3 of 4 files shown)


MrAnno:
@bkil-syslogng Can you have a look at this PR, please?

`_append_unichar()` was introduced in [Format CEF extension](https://github.com/balabit/syslog-ng/pull/842/commits), I'm a little bit confused by this `wc < 0xc0` condition.

bazsi:
Anything below 0xC0 can be encoded in a single byte in utf8. This is a performance optimization, that seriously improved performance, and I don't get why it would break things this badly.

bkil-syslogng:
I think the only fix needed is to patch the function to check `(wc < 0x7f)` instead. This one passes the above tests for me.

bkil-syslogng:
Otherwise good catch.

MrAnno:
@bazsi Thank you. I think it's because `wc` is not an UTF-8 encoded character anymore (`g_utf8_get_char_validated()` converts it to a Unicode character).

Is it acceptable if I change 0xc0 to 0x7f as @bkil-syslogng suggested? This should preserve your optimization for ASCII characters.

bkil-syslogng:
Just for the record, we've now discussed with @bazsi in person that such a workaround would be okay if it is really needed.

Here's the source I used to propose `(wc < 0x7f)`:

> Sixty-five code points (U+0000–U+001F and U+007F–U+009F) are reserved as control codes, and correspond to the C0 and C1 control codes defined in ISO/IEC 6429. Of these U+0009 (Tab), U+000A (Line Feed), and U+000D (Carriage Return) are widely used in Unicode-encoded texts.
https://en.wikipedia.org/wiki/Unicode#Character_General_Category

We do escape control characters < 0x20, not sure what results in appending 0x7f directly.

lib/utf8utils.c
```diff
@@ -36,15 +36,6 @@ _is_character_unsafe(gunichar uchar, const gchar *unsafe_chars)
   return _strchr_optimized_for_single_char_haystack(unsafe_chars, (gchar) uchar) != NULL;
```
bkil-syslogng: Another optimization I see is that we could probably check `(uchar >= 0x7f)` (or `0x80`) in `_is_character_unsafe()`.

MrAnno:
I updated the PR to preserve the mentioned performance optimization. I also fixed the same bug in `format-cef-extension.c`.

bkil-syslogng:
Sorry for the mistake, if I'm reading this correctly, thankfully `0x7F` should be left intact as well, so the check should be `(wc < 0x80)`. The above text was not clear, so we should improve it on Wikipedia.
https://en.wikipedia.org/wiki/C0_and_C1_control_codes#C0_.28ASCII_and_derivatives.29

MrAnno:
@bkil-syslogng U+007F is a control character, why should we append it as is?

bkil-syslogng:
:+1:

bazsi:
This looks good to me and I fully trust @bkil-syslogng about the need to escape 0x7f or not. So if he says it is ok, I agree :)

so  :+1:

bkil-syslogng:
@bazsi **TL;DR**: the reasoning only confirmed that `< 0x80` is the logically sound choice. This PR should fix the said issue of Unicode bytecode breakage in its current form, so it is ready for merging.

Strictly speaking, as `0x7f` is considered a control character, we would probably need to escape it just like `0x00`-`0x1f`. However, this does not affect the said `if`-condition - it's an improvement to be made in the outer loop. Though, because it's an extension to the hot path, we need to be cautious. It would also be wise to research how our users would want to see it. Hence I propose that this possible further enhancement should go into a separate PR.

bkil-syslogng:
@dnsjts Pointed out that the manual inlining could also have been validated from the source (`< 0x80`)
https://git.gnome.org/browse/glib/tree/glib/gstring.c?id=39a22880b6d8544fec51ba3176214f38a8a572b8#n820

lbudai:
Hmm... we have a copy pasted function.
Wouldn't be better to have only one function? Maybe you talked about it and you have your reasons.

@bazsi: how much performance gain can we expect from this? (I'm just curious.)

bazsi:
I can't remember how we ended up with two copies of that function, so that
should probably be eliminated. Maybe @bkil-syslogng remembers.

Doing it with simple append_c instead of append_unichar is a lot faster if
most of the characters are ASCII and they are.

The first is inlined, the 2nd is a function call, in a loop that gets
executed for all characters all affected messages. If we use 20k as message
rate that loop might be executed 2million times per second assuming 100
characters per message.

We do have a bottleneck in format-json as Evan explained in one of his
emails.

bkil-syslogng:
Yes, I remember. I recommended extracting the functionality to a common
place, but I was downvoted.

On Tue, Jun 21, 2016 at 10:50 PM, Balazs Scheidler <notifications@github.com
> wrote:

> I can't remember how we ended up with two copies of that function, so that
> should probably be eliminated. Maybe @bkil-syslogng remembers.
>
> Doing it with simple append_c instead of append_unichar is a lot faster if
> most of the characters are ASCII and they are.
>
> The first is inlined, the 2nd is a function call, in a loop that gets
> executed for all characters all affected messages. If we use 20k as message
> rate that loop might be executed 2million times per second assuming 100
> characters per message.
>
> We do have a bottleneck in format-json as Evan explained in one of his
> emails.

lbudai:
Finally I was able to fix OBS (libdbi and some other deps, libdbi packaging issues, etc...) so Travis should be fine now.

lbudai:
Yeah, the namespace pollution is not the best.

Issue #610, opened 2015-08-03
Assertion error in log_msg_refcache_stop with an empty channel


vincentbernat:
Hey!

With the following configuration, I get reproducibly the following assertion error:

```
ERROR:../lib/logmsg.c:1510:log_msg_refcache_stop: assertion failed: ((logmsg_cached_acks < LOGMSG_REFCACHE_BIAS - 1) && (logmsg_cached_acks >= -LOGMSG_REFCACHE_BIAS))
```

This can be fixed by removing the empty `channel` section. It could also be solved by merging `d_alerts` and `d_kafka` inside `d_remote`.

```conf
@version: 3.6
@include "scl.conf"

@define basepath "/home/bernat/Documents/exoscale/S03-2015/bug1"

options {
  chain_hostnames(off);
  ts-format(iso);
  use_dns(no);
  use_fqdn(no);
  flush-lines(10);
  flush-timeout(1000);
  log-msg-size(65535);
};

destination d_kafka {
  channel {
    rewrite {
      set("filename"   value(".eventv1.filename"));
    };
  };
};

log {
  source {
    file("`basepath`/system-json.log"
         flags(no-parse));
  };

  junction {
    channel {
      rewrite {
        set-tag("warning");
      };
    };
    channel {
    };
  };
  destination(d_remote);
};

destination d_alerts {
  channel {
    rewrite {
      clear-tag("ok");
    };
  };
};

destination d_remote {
  channel {
    destination(d_alerts);
    destination(d_kafka);
  };
};
```

vincentbernat:
In fact, it doesn't seem that the problem is the empty channel, but the fact there are several of them in the junction. The following configuration also triggers the bug:

```
destination d_kafka {
  channel {
    rewrite {
      set("filename"   value(".eventv1.filename"));
    };
  };
};

log {
  source {
    file("`basepath`/system-json.log"
         flags(no-parse));
  };

  junction {
    channel {
      rewrite {
        set-tag("warning");
      };
      flags(final);
    };
    channel {
      rewrite {
        set-tag("warning");
      };
      flags(final);
    };
  };
  destination(d_remote);
};

destination d_alerts {
  channel {
    rewrite {
      set-tag("ok");
    };
  };
};

destination d_remote {
  channel {
    destination(d_alerts);
    destination(d_kafka);
  };
};
```

And removing `set-tag("ok")` is another way to solve it.

vincentbernat:
[...]

PR #1094, opened 2016-06-21
logmpx: remove last_delivery logic

bazsi:
The last_delivery logic kind-of breaks the ack counting logic for junctions
that have branches that modify the LogMessage.

This causes aborts during execution like this:

ERROR:../lib/logmsg.c:1510:log_msg_refcache_stop: assertion failed: ((logmsg_cached_acks < LOGMSG_REFCACHE_BIAS - 1) && (logmsg_cached_acks >= -LOGMSG_REFCACHE_BIAS))

This patch resolves github issue #610.

### DIFF
--- a/lib/logmpx.c
+++ b/lib/logmpx.c
@@ -69,7 +69,6 @@ log_multiplexer_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_op
   LogPathOptions local_options = *path_options;
   gboolean matched;
   gboolean delivered = FALSE;
-  gboolean last_delivery;
   gint fallback;
   
   local_options.matched = &matched;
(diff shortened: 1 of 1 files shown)


lbudai:
So, when a junction has a branch that modifies (calling `make_writable`) a protected logmsg, then
the ack counting breaks.
Is it possible that there are other situations that could lead to the same problem? I mean is it possible to mark a  logmsg as protected somewhere else?

bazsi:
It was not the unprotect call that violated assumptions, but rather this "optimization" in logmpx, which this patch removes.

If an mpx forks off to several branches, the last one got a writable logmsg, as the LogMpx assumed it had a writable access to the message, which it propagated onwards to its last branch. This assumption turned out not to be true, IIRC it can happen that one of the earlier branches keeps a reference to this LogMessage, e.g. it can't be modified.

I can't remember the details honestly, as I wrote this patch almost a year ago. I wanted to research the performance ramifications, but I never did, and I think it should not be that severe. Or of it really turns out to be we have lots of options to improve performance again (e.g. drop copy-on-write cloning and make sure all clone operations happen with a single malloc() instead of the current, clone portions of the data and do graduall mallocs later on as it gets modified).

This patch really makes LogMpx a lot simpler in exchange for some performance and a potential abort. I think it is a fair exchange :)

lbudai:
I see (and yes, it's a fair exchange :-) ) .

PR #1096, opened 2016-06-23
Replaced the old selinux policy installer script with a new one

jszigetvari:
which now has support for Red Hat Enterprise Linux/CentOS/Oracle Linux 5, 6 and 7.

Usage:
syslog_ng.sh    --install-dir <DIRECTORY>    set the installation path of syslog-ng (for non-interactive setups)
        --remove    remove the installed syslog-ng related SELinux policy settings
        --help      prints usage information


lbudai:
@jszigetvari : thanks for the patch!

Travis failed:
make: *** No rule to make target `contrib/selinux/syslog_ng.fc', needed by `distdir'.
 
You should also modify `contrib/Makefile.am` .

lbudai:
sidenote  :
The modified travis yaml is not the best: what is the reason of skipping make distcheck for clang? if we run distcheck only for gcc, then the name should refer to it.
(otherwise it looks like that we have some gcc specific issue...)

jszigetvari:
@ihrwein told me this would happen, and will fix this later on.

lbudai:
@jszigetvari: you mean @ihrwein will fix it?
I wouldn't merge this until it not fixed.

jszigetvari:
@lbudai: Yes @ihrwein told me so. I can ping him later today, with this topic.
Of course, no problem. It can wait.

ihrwein:
Done :)

contrib/selinux/syslog_ng.sh
```diff
@@ -1,63 +1,272 @@
...
+		remove_module
+		exit 0
+		;;
+	showhelp)
+		echo -e "${USAGE}"
+		exit 1
+		;;
+	install)
+		if [ $( id -u ) != 0 ]; then
+			echo 'You must be root to run this script'
```
kvch: This error message should be redirected to stderr, just like other error messages in the script.

contrib/selinux/syslog_ng.sh
```diff
@@ -1,63 +1,272 @@
...
+	echo -n "Please enter your installation path for Syslog-ng PE: [${INSTALL_PATH}] "
+	read INPUT
+}
+
+check_dir() {
+	if [ -d "${1}" ]; then
+		return 0
+	else
+		echo "The directory you specified does not exist!"
+		return 1
```
kvch: Different return values should be used on different errors. If someone runs it from a script, the chance is taken away from him or her to react accordingly based on the error.

contrib/selinux/syslog_ng.sh
```diff
@@ -1,63 +1,272 @@
...
+			;;
+		*)
+			echo -e "ERROR: Invalid option: ${1}\n${USAGE}" >&2
+			exit 1
+			;;
+	esac
+
+done
+
+case "${TASK_SELECTED}" in
```
kvch: I would not juggle with more "switch-cases" based on the value of TASK_SELECTED. Instead I would create functions, and run them directly from the while loop above. Or is there any specific reason of implementing it in such an inexplicable way?

jszigetvari: The primary reason why I designed the code this way is that:
- I wanted to make sure, that all CLI arguments are parsed first
- some arguments take precedence over others: help > remove > install
- I wanted to avoid the script acting differently when the arguments are specified in a different order
- invoking functionality from the first switch would cause the functionality being invoked multiple times, in case an argument were to be written multiple times

This seemed to be the easiest and most efficient way to do it. (Well, at least for me.)

Your other comments are perfectly valid, and I have already corrected them.

jszigetvari:
Corrected most of the bugs discovered by @kvch

contrib/selinux/syslog_ng.sh
```diff
@@ -1,63 +1,282 @@
...
-			echo "Do you want these changes added to policy [y/n]?"
-			read ANS
-			if [ "$ANS" = "y" -o "$ANS" = "Y" ] ; then
-				echo "Updating policy"
-				echo -e "$rules" | audit2allow -R >> syslog_ng.te
-				# Fall though and rebuild policy
-			else
-				exit 0
-			fi
+#!/bin/bash
```
kvch: I ran ``shellcheck`` on this script and it shown a few problems with this script. Please, fix these issues. :)

jszigetvari: Most of the `shellcheck` findings were irrelevant, and there I was explicitly counting on bash breaking up strings along whitespaces.

Despite the above, there were in deed some path-related quotation problems, that I fixed with this commit.

Thanks to @kvch for this interation of the review loop!

contrib/selinux/syslog_ng.sh
kvch: I agree on that if `--help` is supplied as a parameter in the paramlist help should run. But I don't  understand why further parameter precedence is needed. Especially when the user does not even know about it because it is not documented in the script. (Or is it documented somewhere?) Is it really necessary?

To eliminate the ugly lines I propose the following:
```
# check if help can be found in the list of params
for var in "$@"; do
    if [ ${var} = "--help" ]; then
        echo -e "${USAGE}"
    fi
done

# if 
case "${1}" in
	--remove)
		detect_os_version
		setup_vars
		remove_module
		exit 0
		;;
	--install-dir)
        run_install
    *)
        echo -e "ERROR: Invalid option ${1}\n${USAGE}" >&2
        exit 1
        ;;

esac
```

What do you think?

kvch:
We discussed the operator precedence with @jszigetvari and decided that the precedence should be documented. Because this script is expected to be run for atmost 2-3 times on the machines of the subset of syslog-ng users, further code changes seems to be unnecessary. So, after the help text is extended, I give my :+1:

ihrwein:
:+1:

jszigetvari:
I modified the handling of parameters, as per @kvch's suggestion, and now we err out in case of conflicting parameters instead of trying to figure out what the user wanted.
Also, added some additional error checking.

kvch:
Thanks.

PR #1100, opened 2016-06-28
Fix relative path canonicalization in check_java_support.m4

MrAnno:
The `AX_READLINK` macro was created to provide a portable version of the `readlink -f` command (OS X support).

`AX_READLINK` should canonicalize the given file name by following symlinks recursively and it should return an **absolute** path.

In case of relative symlinks, the macro has not returned an absolute path.


bkil-syslogng:
Wouldn't a simple `(cd $DIR && pwd -P)` work?

"Display the physical current working directory (all symbolic links resolved)."

lbudai:
I need this right now :-)
( On Ubuntu 16.04 with the provided gradle 2.10 we cannot build syslog-ng. )

PR #1103, opened 2016-06-29
Implement deinit() for native parsers

ihrwein:
I forgot to forward the `deinit()` calls to the native parsers. This can cause problems with reloads.

I also added a commit for calling `log_parser_init_method()` in `init()`. It didn't produce problems, but other parsers call it, so it's best to call it here as well.


bazsi:
:+1:

juhaszviktor:
:+1:

PR #1109, opened 2016-07-05
include geoip tests only when geoip is enabled

kvch:
Previously, when geoip was found on a machine where automake was run, after moving to another machine geoip tests failed while building.

This patch modifies the makefiles, so geoip tests are only run when geoip is enabled.


bazsi:
I think we don't need the first hunk (e.g. the one in modules/geoip/Makefile.am) as the 2nd one achieves the same, but maybe I am missing something.

Otherwise :+1:, so with the change above this can be merged, (so go ahead and merge it) thanks.

If I miss something, please explain why the change is needed in the root Makefile.am

kvch:
The change was not needed in the root file. I amended the commit. :)

PR #1110, opened 2016-07-06
make xmllint dependency explicit in .travis.yml

### LINKED ISSUE #1087: xmllint is required for building syslog-ng
I was trying to build `syslog-ng(syslog-ng 3.8.0alpha0)` from source. I used the following dependencies:
- https://github.com/balabit/syslog-ng-docker/blob/master/syslog-ng-dev/dev-dependencies.txt
- https://github.com/balabit/syslog-ng/blob/master/.travis.yml

The build succeeds but one of the tests `modules/dbparser/tests/test_patterndb` fails. A quick look in the log file suggests `xmllint` is required for passing this test.

``` shell
FAIL: ASSERTION FAILED: Error validating patterndb, error=Failed to execute child process "xmllint" (No such file or directory)
```

I cannot see the package in the list of dependencies. So, I guess it should be added to them.

kvch:
If someone installs dependencies based on .travis.yml, he or she might miss libxml2 because it is not added there expliticly. Right now patterndb unit tests pass on Travis, because it is installed there by dedfault. But on other machines it might be missing. (Fixes #1087)

Please, note that there is another PR in syslog-ng/syslog-ng-docker, which adds the same lib to the dockerfile.

Signed-off-by: Noémi Ványi <sitbackandwait@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ install:
       libnet1-dev
       libriemann-client-dev
       libwrap0-dev
+      libxml2-utils
       pkg-config
       sqlite3
       xsltproc


bkil-syslogng:
:-1: It was a conscious decision to not document user installation on `.travis.yml`. It slows down CI for no benefit. Travis has many development libraries pre-installed, some of which are uncommon, so this list can never be verified to be complete, which may be again misleading to some.

We use docker and the installation guide for documenting installation.

bkil-syslogng:
By the way, this does not fix #1087.

kvch:
Sorry, my bad. I am closin this PR.

PR #1113, opened 2016-07-07
debian: Fixed wrong man files

jpereira:


bazsi:
why is it better to install the .xml files? does dh_installman support conversion from docbook to groff?

kira-syslogng:

kira-syslogng:

kira-syslogng:

presidento:
@jpereira thank you for your contribution. We have talked about it IRL, and concluded that this would increase the build dependencies, so we would like to stay with the original file format.

PR #1114, opened 2016-07-07
debian: Adding missed compat

jpereira:


lbudai:
This has been already resolved by merging #1125.

PR #1115, opened 2016-07-07
logmsg: add log_msg_new_local() function

ihrwein:
This function creates a new LogMessage with LF_LOCAL flag and its
timestamps are initialized to the same value.

Reason: $YEAR and other time related macros use timestamps[LM_TS_STAMP]
timestamp. If that timestamp isn't initialized (-1), then $YEAR is
evaluated to 1969.

I handle LogMessage instances as opaque pointers, so I don't have access
to their flags. LogMessage doesn't have setter and getter methods for
its flags and I don't want to modify them after an instance is created,
so this is the simplest and less intrusive solution I could think of.

### DIFF
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -567,21 +567,17 @@ log_writer_emit_suppress_summary(LogWriter *self)
 
   msg_debug("Suppress timer elapsed, emitting suppression summary");
 
-  m = log_msg_new_empty();
-  m->timestamps[LM_TS_STAMP] = m->timestamps[LM_TS_RECVD];
-  m->pri = self->last_msg->pri;
-  m->flags = LF_INTERNAL | LF_LOCAL;
+  len = g_snprintf(buf, sizeof(buf), "Last message '%.20s' repeated %d times, suppressed by syslog-ng on %s",
+                   log_msg_get_value(self->last_msg, LM_V_MESSAGE, NULL),
+                   self->last_msg_count,
+                   get_local_hostname_fqdn());
+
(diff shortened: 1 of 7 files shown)


bazsi:
Some kind of test coverage would be nice. There's even a pretty solid, but only partial unit test for log messages.

Also, I guess that flag is used elsewhere can't we use this function anywhere else?

ihrwein:
@bazsi I added a unit test and checked the places where LF_LOCAL was set: logwriter and dbparser. I updated their code as well.

lib/logwriter.c
```diff
@@ -567,10 +567,9 @@ log_writer_emit_suppress_summary(LogWriter *self)
 
   msg_debug("Suppress timer elapsed, emitting suppression summary");
 
-  m = log_msg_new_empty();
-  m->timestamps[LM_TS_STAMP] = m->timestamps[LM_TS_RECVD];
+  m = log_msg_new_local();
   m->pri = self->last_msg->pri;
-  m->flags = LF_INTERNAL | LF_LOCAL;
+  m->flags |= LF_INTERNAL;
```
bazsi: There seems to be a log_msg_new_internal() function as well, which should apply to this case better.

If you update the branch with this, you have my :+1:  do go ahead and merge it.

ihrwein: Thanks! I replaced the last commit which modified logwriter with this one: https://github.com/balabit/syslog-ng/pull/1115/commits/02a8c072529816bfd7594f5365d9894bc629cb7b

The other commits are untouched by the force push.

kvch:
:+1:

PR #1117, opened 2016-07-11
F/fix test kv parser va arg portability

bkil-syslogng:
The new kv parser unit test could not be run on Solaris because of the non-portable usage of `va_arg()`. It worked on GNU/Linux and Mac OS X.

These changes eliminate all usage of `va_arg()`.

The intermediate commit contains parts which are removed by the last commit, and are introduced to ease reviewing.

### DIFF
--- a/modules/kvformat/tests/test_kv_scanner.c
+++ b/modules/kvformat/tests/test_kv_scanner.c
@@ -46,14 +46,6 @@ _assert_no_more_tokens(KVScanner *scanner)
 }
 
 static gboolean
-_scan_next_token(KVScanner *scanner)
-{
-  gboolean ok = kv_scanner_scan_next(scanner);
-  expect_true(ok,  "kv_scanner is expected to return TRUE for scan_next");
-  return ok;
-}
-
-static gboolean
 _assert_current_key_is(KVScanner *scanner, const gchar *expected_key)
 {
   const gchar *key = kv_scanner_get_current_key(scanner);
@@ -78,10 +70,14 @@ _compare_key_value(KVScanner *scanner, va_list args, gboolean *expect_more)
   if (!kv)
     return FALSE;
 
-  if (!_scan_next_token(scanner))
+  gboolean ok = kv_scanner_scan_next(scanner);
+  if (!ok)
     {
       *expect_more = FALSE;
-      return TRUE;
+      expect_true(ok, "kv_scanner is expected to return TRUE for scan_next(), "
+                  "first unconsumed key: [%s]",
+                  kv);
+      return FALSE;
     }
 
   _assert_current_key_is(scanner, kv);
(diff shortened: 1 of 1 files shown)


bazsi:
wouldn't  va_copy()  have helped?

bkil-syslogng:
@bazsi 
* This solution is zero-copy.
* `va_copy()` is a newer standard, in C99, although most of the platforms that we are interested in might provide it (I did not check),
* In my opinion, this one is more clear, better in type safety and boundary checking. It also combines solutions which were favored by others as well according to my survey done during the development of the previous implementation (I did not ask whether they like this synthesis of them).

bkil-syslogng:
And a non-negligible argument is that this one is harder to mess up.

bazsi:
Hi,

This one is from 1996. It probably has been standardized in C99 but has
been around for a lot longer.

https://docs.oracle.com/cd/E19455-01/806-0639/6j9vstu2j/index.html

Anyway, I am not advocating ripping of the new implementation, I just
wanted to add a new piece of information.

I like type safety as well, however, the amount of macro magic in this unit
test starts to go over my thresholds and although I like single-line
testcases, you lost me in the implementation details. And macro complexity
makes debugging/understanding a lot more difficult.


-- 
Bazsi

modules/kvformat/tests/test_kv_scanner.c
```diff
@@ -82,105 +96,98 @@ _compare_key_value(KVScanner *scanner, va_list args, gboolean *expect_more)
...
-_value_was_quoted(KVScanner *scanner, va_list args, gboolean *expect_more)
+_value_was_quoted(KVScanner *scanner, VAIterator *args, gboolean *expect_more)
 {
   if (!_compare_key_value(scanner, args, expect_more))
     return FALSE;
 
-  gboolean was_quoted = va_arg(args, gboolean);
-  g_assert((was_quoted == FALSE) ||(was_quoted == TRUE));
+  gboolean was_quoted = GPOINTER_TO_SIZE(va_get(args));
+  g_assert((was_quoted == FALSE) || (was_quoted == TRUE));
```
ihrwein: This line is very interesting, what's its purpose? :)

ihrwein:
:+1:

modules/kvformat/tests/test_kv_scanner.c
bkil-syslogng: Thanks for the review.

This check was introduced by me in the previous recent refactor and was only present in this intermediate refactor step. Thankfully, it is not present in the final version after the last commit, because the struct array wrapper protects against such mistakes.

This check was preferred because in case of varargs, there was a small chance that invalid usage could cause misalignment/mismatching of the arguments. This could catch some of such cases, because when accidentally casting a string `pointer` to a `gsize`, you rarely get 0 or 1.

lbudai:
( @bkil-syslogng : syslog-ng requires a C99 compiler, so using `va_copy` should not be a problem 
I cannot say +1 because of those macros (test code is production code, it should be clean code), but if others like it, and got the second +1 , I don't mind... When we move to the new unit test framework ( @MrAnno  ), we should have to re-implement this.  )

bazsi:
I wouldn't stop the patches either.

juhaszviktor:
:+1: With restriction written by @lbudai.

PR #1119, opened 2016-07-12
Native parser: fail-fast when configuration setting failed

ihrwein:
This PR changes the signature of `native_parser_set_option()` to return with a boolean value and so makes possible to fail-fast in case of an invalid option value.

What's an invalid option value?
* the value is a path and the file doesn't exist
* the value is a regular expression, but it's invalid
* every other native parser specific 

This PR also contains a little refactor (fixing an include guard, removing some duplicated function prototypes).

There are no unit tests for this change, mainly because I don't know any examples where syslog-ng's grammar tested this way.

This PR is the pair of https://github.com/ihrwein/syslog-ng-rust-modules/pull/35

### DIFF
--- a/modules/native/native-grammar.ym
+++ b/modules/native/native-grammar.ym
@@ -40,7 +40,7 @@
 
 #include <string.h>
 
-void native_parser_set_option(LogParser *s, gchar* key, gchar* value);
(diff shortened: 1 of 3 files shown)


lbudai:
:+1:

PR #1120, opened 2016-07-14
parser: fix invalid pointer usage in log_parser_process_message()

MrAnno:
In #1102, `log_parser_process_message()` was extracted from `log_parser_queue()` to make it possible to invoke template processing from outside code.

Unfortunately, this transformation is not fully equivalent. `self->process()` can create a new LogMessage instance (`log_msg_clone_cow()`).
In this case, only the local `msg` pointer will be updated, `*pmsg` will remain unchanged.

This PR fixes this issue by passing the original `pmsg` variable to the `process()` function.

### DIFF
--- a/lib/parser/parser-expr.c
+++ b/lib/parser/parser-expr.c
@@ -58,7 +58,7 @@ log_parser_process_message(LogParser *self, LogMessage **pmsg, const LogPathOpti
        */
 
       value = log_msg_get_value(msg, LM_V_MESSAGE, &value_len);
-      success = self->process(self, &msg, path_options, value, value_len);
+      success = self->process(self, pmsg, path_options, value, value_len);
       nv_table_unref(payload);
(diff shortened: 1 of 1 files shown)


dnsjts:
Nice catch, this should be merged as soon as possible.

bkil-syslogng:
:+1: Excellent.

I wonder whether some kind of deep value analysis and warn_unused annotations could help us catch such issues. A much simpler model of this issue could have definitely been caught, but we'd need to do some research to say for sure in the real case.

bazsi:
wow, thanks a lot for fixing this. how did you find it?

MrAnno:
We didn't implement write-protection on `LogMessage` objects in our internal repos (`log_msg_clone_cow()` is called directly from parsers), so a lot of parser-related tests have failed.

PR #1121, opened 2016-07-14
Fix segmentation fault when including SCLs

ihrwein:
Let's suppose we have a config file which includes files this way:

`scl/*/*.conf`

If we create a `scl/e/e.conf` file under `etc`, this glob pattern is
expanded as:

```
etc/scl/e/e.conf
share/syslog-ng/include/scl/a/a.conf
share/syslog-ng/include/scl/b/b.conf
```

We have to place these files on the same include stack level, as they were
expanded from the same pattern.

syslog-ng pushed on the include stack every time it evaluated a new
include-path prefix (etc and share/syslog-ng/include/scl). This mishandling caused the SEGFAULT (though I didn't track it down exactly why).

I moved the include stack level increasing one level upper, so it's
increased only once. If syslog-ng isn't able to include a file (so the
top of the stack is unused), we restore the stack to it's original state.

### DIFF
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -489,7 +489,6 @@ cfg_lexer_include_file_glob_at(CfgLexer *self, const gchar *pattern)
 #ifndef SYSLOG_NG_HAVE_GLOB_NOMAGIC
           if (!__glob_pattern_p (pattern))
             {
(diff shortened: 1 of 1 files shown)


kvch:
:+1:

PR #1123, opened 2016-07-15
F/kv tagger

lbudai:
* What is kvtagger?

With `kvtagger` (technically, a syslog-ng `parser`) syslog-ng can use an external database file to append custom name-value pairs on incoming logs (to enrich messages). The 'database' is actually a file that containing `<selector, name, value>` records. Currently only `CSV` format is supported.

It is like `geoip parser` where the selector is `$HOST` , but in this case, the user can define the selector, and also the database contents.

* options
```
database         | mandatory | file path | an absolute or relative path for database (*.csv) file
selector         | mandatory | template  | this value will be used to match on first column in database file
default-selector | optional  | string    | this value will be used to use as a non matching selector in database file
prefix           | optional  | string    | this value will be used as a prefix on name part of a tag

```

* Example:

  * kvtagger-db.csv:

```
localhost,localhost_tag1,localhost_value1
localhost,localhost_tag1,localhost_value2
```

   * config
```
@version: 3.8
@include "scl.conf"

source s_network {
        network(port(5555));
};

destination d_local {
        file("/tmp/test-kv-tagged-msgs.log" template("$MSG injected-tag1:[${mytag.localhost_tag1}]};

parser p_kv_tagger {
       kvtagger(selector("$HOST"), database("kvtagger-db.csv"), prefix("mytag"));
};

log {
        source(s_network);
        parser(p_kv_tagger);
        destination(d_local);
};
```

@fekete-robert : you may interested in this


lbudai:
I'm open to rename `tag` to something else  :-)

bazsi:
where would you rename that? I agree that kvTAGger() can be misleading as
we attach name-value pairs and not tags.

Why don't we call this enrich()

Can we use template strings within the CSV file or just literal strings?


-- 
Bazsi

lbudai:
yeah, I wasn't clear: rename this kvtagger to something else. Enrich: we can use this (we drop this at the beginning because it is too general, but kvtagger is misleading...).

CSV file: only simple strings are supported.

bazsi:
I think we should definitely support template strings in the value part.

With regards to the name, I would use something like enrich-values()

lbudai:
@bazsi : template strings: in another PR?

name: `kvappender`? `kvtagger` is a `strong` name and I think `enrich-values` won't replace it (we always will call it internally as `kvtagger`, appender may replace it)

(note from @kvch ): `enrich` suggests to the user that is makes the message better, but it just adds some name value pair (key-value pair) to the message

fekete-robert:
What about
 * add-external-info
 * add-external-metadata
 * insert-data
 * add-contextual-data
 * fetch-corresponding-data
 * add-related-data

Or some permutation of the above

lbudai:
@fekete-robert :  `insert-contextual-fetch-external-metadata-related` ? :-)

Thanks for the hints!

fekete-robert:
Or rather:
this-thing-reads-stuff-from-a-file-and-adds-it-to-your-messages-as-name-value-pairs-but-we-could-not-agree-on-a-shorter-name

lbudai:
@fekete-robert : `insert-contextual-info` ?

bkil-syslogng:
* annotator
* add-annotation
* annotate-messages
* labeler
* add-labels
* tally

modules/kvtagger/kvtagger-grammar.ym
```diff
@@ -0,0 +1,94 @@
...
+%parse-param {LogParser **instance}
+%parse-param {gpointer arg}
+
+/* INCLUDE_DECLS */
+
+%token KW_KVTAGGER
+%token KW_KVTAGGER_DATABASE
+%token KW_KVTAGGER_SELECTOR
+
+%type	<ptr> parser_expr_csv
```
bkil-syslogng: \t

modules/kvtagger/kvtagger-grammar.ym
```diff
@@ -0,0 +1,94 @@
...
+        ;
+
+
+parser_expr_csv
+        : KW_KVTAGGER '('
+          {
+            last_parser = *instance = kvtagger_parser_new(configuration);
+          }
+          parser_csv_opts
+          ')'					{ $$ = last_parser; }
```
bkil-syslogng: \t\t\t\t\t

modules/kvtagger/kvtagdb.c
```diff
@@ -0,0 +1,264 @@
...
+{
+  g_assert(!self || g_atomic_counter_get(&self->ref_cnt));
+  if (g_atomic_counter_dec_and_test(&self->ref_cnt))
+    {
+      kvtagdb_free(self);
+    }
+}
+
+static element_range*
+_get_range_of_tags(KVTagDB *const self, const gchar *const selector)
```
bkil-syslogng: I still agree that we could and should add `const` where ever possible, however because you never use `*const self` anywhere else, it seems a bit confusing here. The reader could get puzzled whether the `const` was misplaced or not.

modules/kvtagger/kvtagger.c
```diff
@@ -84,10 +85,48 @@ kvtagger_set_database_selector_template(LogParser *p, const gchar *selector)
...
+}
+
+static inline element_range *
+kvtagger_lookup_tag(KVTagger *const self, const gchar *const selector)
+{
+  return g_hash_table_lookup(self->tagdb.index, selector);
+}
+
+gboolean
+_is_default_selector_set(KVTagger *self)
```
bkil-syslogng: This should be static, and could even use `const *self`.

modules/kvtagger/kvtagger.c
```diff
@@ -84,10 +85,48 @@ kvtagger_set_database_selector_template(LogParser *p, const gchar *selector)
...
+{
+  element_range *position_of_tags_to_be_inserted = kvtagger_lookup_tag(self, selector->str);
+  if (position_of_tags_to_be_inserted == NULL && _is_default_selector_set(self))
+    {
+      position_of_tags_to_be_inserted = kvtagger_lookup_tag(self, self->default_selector);
+    }
+  return position_of_tags_to_be_inserted;
+}
+
+void
```
bkil-syslogng: This could be static.

modules/kvtagger/kvtagger.c
```diff
@@ -84,10 +85,48 @@ kvtagger_set_database_selector_template(LogParser *p, const gchar *selector)
...
+_tag_message(KVTagger *self, LogMessage *msg, element_range *position_of_tags_to_be_inserted)
+{
+  for (gsize i = position_of_tags_to_be_inserted->offset;
+       i < position_of_tags_to_be_inserted->offset + position_of_tags_to_be_inserted->length;
+       ++i)
+    {
+      tag_record matching_tag = g_array_index(self->tagdb.data, tag_record, i);
+      log_msg_set_value_by_name(msg, matching_tag.name, matching_tag.value, -1);
```
bkil-syslogng: Loading should store the string length here as well, so we wouldn't need to call `strlen()` all the time due to `-1`.

modules/kvtagger/kvtagger.c
```diff
@@ -378,6 +408,7 @@ kvtagger_parser_new(GlobalConfig *cfg)
   self->super.super.free_fn = kvtagger_parser_free;
   self->super.super.init = kvtagger_parser_init;
   self->scanner = NULL;
+  self->default_selector = NULL;
```
bkil-syslogng: We don't explicitly initialize members which are already initialized to the same value by `g_new0()` (`0`).

kvch:
@bkil-syslogng Are you reviewing outdated diffs on purpose?

bkil-syslogng:
@kvch I am looking at the full diff and individual patches in parallel and I can clearly see that the said line is present in both. In this case, I comment on the root cause.

modules/kvtagger/kvtagger.c
```diff
@@ -0,0 +1,333 @@
...
+
+  self->super.process = kvtagger_parser_process;
+  self->tagdb = kvtagdb_new();
+
+  self->super.super.clone = kvtagger_parser_clone;
+  self->super.super.free_fn = kvtagger_parser_free;
+  self->super.super.init = kvtagger_parser_init;
+  self->default_selector = NULL;
+  self->selector_template_string = NULL;
+  self->prefix = NULL;
```
bkil-syslogng: We don't explicitly initialize members which are already initialized to the same value by `g_new0()` (`0`).

modules/kvtagger/kvtagger.c
```diff
@@ -0,0 +1,333 @@
...
+_is_default_selector_set(KVTagger *self)
+{
+  return (self->default_selector != NULL);
+}
+
+void
+_tag_message(gpointer pmsg, const TagRecord *tag)
+{
+  LogMessage *msg = (LogMessage *) pmsg;
+  log_msg_set_value_by_name(msg, tag->name, tag->value, -1);
```
bkil-syslogng: Loading should store the string length here as well, so we wouldn't need to call `strlen()` all the time due to `-1`.

modules/kvtagger/kvtagger.c
```diff
@@ -0,0 +1,333 @@
...
+  self->default_selector = g_strdup(default_selector);
+}
+
+gboolean
+_is_default_selector_set(KVTagger *self)
+{
+  return (self->default_selector != NULL);
+}
+
+void
```
bkil-syslogng: This could be `static`.

kvch:
You can comment however you like. It was just weid that according to Github it was outdated.

modules/kvtagger/kvtagger.c
```diff
@@ -0,0 +1,333 @@
...
+kvtagger_set_database_default_selector(LogParser *p, const gchar *default_selector)
+{
+  KVTagger *self = (KVTagger *)p;
+
+  g_free(self->default_selector);
+  self->default_selector = g_strdup(default_selector);
+}
+
+gboolean
```
bkil-syslogng: This should be `static`, and could even use `const *self`.

modules/kvtagger/kvtagger-grammar.ym
```diff
@@ -0,0 +1,103 @@
...
+
+/* INCLUDE_DECLS */
+
+%token KW_KVTAGGER
+%token KW_KVTAGGER_DATABASE
+%token KW_KVTAGGER_SELECTOR
+%token KW_KVTAGGER_DEFAULT_SELECTOR
+%token KW_KVTAGGER_PREFIX
+
+%type	<ptr> parser_expr_csv
```
bkil-syslogng: \t

bkil-syslogng:
There are >4 lines longer than 120 columns, much more longer than 100. Please identify via `git grep -E ".{100}"` in the added folders, then decide on the width that you'd like to keep.

lib/scanner/csv-scanner/Makefile.am
```diff
@@ -0,0 +1,7 @@
+csvscannerincludedir			= ${pkgincludedir}/scanner/csv-scanner
+
+csvscannerinclude_HEADERS = 			\
+	lib/scanner/csv-scanner/csv-scanner.h
+
+csvscanner_sources = 				\
+	lib/scanner/csv-scanner/csv-scanner.c
```
bkil-syslogng: \t both here and above

modules/kvtagger/csv-tagrecord-scanner.c
```diff
@@ -0,0 +1,110 @@
...
+csv_tag_record_scanner_free(TagRecordScanner *s)
+{
+  CSVTagRecordScanner *self = (CSVTagRecordScanner *)s;
+  csv_scanner_options_clean(&self->options);
+  csv_scanner_state_clean(&self->scanner);
+  g_free(self);
+}
+
+TagRecordScanner*
+csv_tag_record_scanner_new()
```
bkil-syslogng: `(void)`

modules/kvtagger/csv-tagrecord-scanner.h
```diff
@@ -0,0 +1,39 @@
...
+
+typedef struct _CSVTagRecordScanner
+{
+  TagRecordScanner super;
+  CSVScanner scanner;
+  CSVScannerOptions options;
+
+} CSVTagRecordScanner;
+
+TagRecordScanner* csv_tag_record_scanner_new();
```
bkil-syslogng: `(void)`

modules/kvtagger/csv-tagrecord-scanner.c
bkil-syslogng: There are many more missing, please `grep` and fix.

modules/kvtagger/kvtagger.c
```diff
@@ -0,0 +1,333 @@
...
+
+void
+_tag_message(gpointer pmsg, const TagRecord *tag)
+{
+  LogMessage *msg = (LogMessage *) pmsg;
+  log_msg_set_value_by_name(msg, tag->name, tag->value, -1);
+}
+
+static gboolean
+kvtagger_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_options,
```
lbudai: and here, we should remove the `kvtagger` prefix

modules/kvtagger/csv-tagrecord-scanner.c
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+#include "csv-tagrecord-scanner.h"
+#include "scanner/csv-scanner/csv-scanner.h"
+#include "string-list.h"
+#include <string.h>
+
+
+static gchar*
+_csv_scanner_dup_current_value_with_prefix(CSVScanner *line_scanner, const gchar *prefix)
```
bkil-syslogng: Redundant prefix for static function, although it's good enough for this way if you disagree.

modules/kvtagger/kvtagger.c
```diff
@@ -0,0 +1,333 @@
...
+      f = fopen(filename, "r");
+    }
+
+  return f;
+}
+
+const gchar*
+_get_filename_extension(const gchar *filename)
+{
+   gsize filename_len = strlen(filename);
```
bkil-syslogng: Overindented.

modules/kvtagger/kvtagdb.c
lbudai: This code is a result of co-working (code is written by three different people, started from a poc). This is why we have different styles.. (and other static/non-static issues) And I'm really happy that you found only nitpicks(one relevant note : storing the length of the strings in records), it wasn't easy to reach the final form :-)

modules/kvtagger/tests/test_kvtagdb.c
```diff
@@ -0,0 +1,259 @@
...
+_fill_tagdb(KVTagDB *tagdb, const gchar *selector_base,
+            const gchar *name_base, const gchar *value_base,
+            int number_of_selectors, int number_of_nv_pairs_per_selector)
+{
+  int i, j;
+  for (i = 0; i < number_of_selectors; i++)
+    {
+      for (j = 0; j < number_of_nv_pairs_per_selector; j++)
+        {
+          TagRecord record = {
```
bkil-syslogng: Eclipse thinks that the parenthesis should start in a new line, but I see this convention a lot. Which one is correct?

bkil-syslogng:
You should inspect orphaned whitespace via:
`git grep -n "  *$" | cat -A`

lbudai:
@bkil-syslogng : I think, what we need is not `grep` but a stylechecker ;-)
And when we have, we will use. 

As I see, most of your notes could be produced by such a tool.
Could you use a stylecheck tool and just copy paste the result in the end ? :-)

modules/kvtagger/kvtagger.c
```diff
@@ -0,0 +1,333 @@
...
+  _replace_template(&cloned->selector_template, self->selector_template);
+  kvtagger_set_prefix(&cloned->super, self->prefix);
+  kvtagger_set_filename(&cloned->super, self->filename);
+  kvtagger_set_database_default_selector(&cloned->super, self->default_selector);
+
+  return &cloned->super.super;
+}
+
+static void
+kvtagger_parser_free(LogPipe *s)
```
bkil-syslogng: I'll probably not list all of these redundant prefix/missing underscore `static` cases, but I hope it will be fixed together.

modules/kvtagger/tests/test_kvtagdb.c
```diff
@@ -0,0 +1,259 @@
...
+}
+
+static void
+test_import_with_valid_csv()
+{
+  gchar csv_content[] = "selector1,name1,value1\n"\
+                        "selector1,name1.1,value1.1\n"\
+                        "selector2,name2,value2\n"\
+                        "selector3,name3,value3";
+  FILE *fp = fmemopen(csv_content, strlen(csv_content) + 1, "r");
```
bkil-syslogng: Note that `fmemopen()` is POSIX.1-2008, which may or may not cause platform specific issues.

modules/kvtagger/tests/test_kvtagdb.c
```diff
@@ -0,0 +1,259 @@
...
+
+#include "kvtagdb.h"
+#include "testutils.h"
+#include <stdio.h>
+#include <string.h>
+
+#define KVTAGDB_TESTCASE(testfunc, ...) {testcase_begin("%s(%s)", #testfunc, #__VA_ARGS__); testfunc(__VA_ARGS__); testcase_end();}
+
+static void
+_foreach_ctr(gpointer arg, const TagRecord *record)
```
bkil-syslogng: `ctr` does not sound like a descriptive name, neither is it a common abbreviation.

bkil-syslogng: And because `arg` is always an integer here, why force casting to and from `gpointer`?

modules/kvtagger/tests/test_kvtagdb.c
```diff
@@ -0,0 +1,259 @@
...
+static void
+_test_empty_db(KVTagDB *tagdb)
+{
+  assert_false(kvtagdb_is_loaded(tagdb) == TRUE, "Empty KVTagDB should be in unloaded state.");
+  assert_false(kvtagdb_is_indexed(tagdb) == TRUE, "Empty KVTagDB should be in un-indexed state.");
+  assert_false(kvtagdb_contains(tagdb, "selector") == TRUE, "Method kvtagdb_contains should work with empty KVTagDB.");
+  assert_true(kvtagdb_number_of_tags(tagdb, "selector") == 0, "Method kvtagdb_number should work with empty KVTagDB.");
+  int ctr = 0;
+  kvtagdb_foreach_tag(tagdb, "selector", _foreach_ctr, (gpointer)&ctr);
+  assert_true(ctr == 0, "Method kvtagdb_foreach_tag should work for with empty KVTagDB.");
```
bkil-syslogng: s/ for//

modules/kvtagger/tests/test_kvtagdb.c
```diff
@@ -0,0 +1,259 @@
...
+}
+
+static void
+test_empty_db()
```
bkil-syslogng: The difference between `_test_empty_db(void)` and `test_empty_db(void)` is that one calls the other. Perhaps some more descriptive name? (and missing underscore as noted earlier)

modules/kvtagger/tests/test_kvtagdb.c
```diff
@@ -0,0 +1,259 @@
...
+
+  kvtagdb_unref(tagdb);
+}
+
+static void
+test_purge_empty_db()
+{
+  KVTagDB *tagdb = kvtagdb_new();
+
+  kvtagdb_purge(tagdb);
```
bkil-syslogng: It would be nice to a see a purge test on a non-empty db as well.

modules/kvtagger/tests/test_kvtagdb.c
```diff
@@ -0,0 +1,259 @@
...
+}
+
+static void
+_assert_kvtag_db_contains_name_value_pairs_by_selector(KVTagDB *tagdb, 
+                                                       const gchar *selector,
+                                                       TestNVPair *expected_nvpairs,
+                                                       guint number_of_expected_nvpairs)
+{
+  TestNVPair result[number_of_expected_nvpairs];
+  TestNVPairStore result_store = {.pairs = result, .ctr = 0};
```
bkil-syslogng: All such constants should be declared `const`, I won't list them all here.

modules/kvtagger/tests/test_kvtagdb.c
bkil-syslogng: `sizeof(csv_content)`

modules/kvtagger/kvtagger.c
bkil-syslogng: `strrchr` should be used instead of this hard coding. Some possible edge cases to consider:

    xycsv
    xy.sqlite

modules/kvtagger/tagrecord-scanner.c
```diff
@@ -0,0 +1,66 @@
...
+
+TagRecordScanner*
+create_tag_record_scanner_by_type(const gchar *type)
+{
+  TagRecordScanner *scanner = NULL;
+
+  if (type == NULL)
+    return NULL;
+
+  if (!strncmp(type, "csv", 3))
```
bkil-syslogng: This compares the first 3 characters of the type, i.e., it accepts ".csvx" as well.

modules/kvtagger/kvtagdb.c
```diff
@@ -0,0 +1,264 @@
...
+gboolean
+kvtagdb_is_indexed(KVTagDB *self)
+{
+  return self->is_data_indexed;
+}
+
+gboolean
+kvtagdb_is_loaded(KVTagDB *self)
+{
+  return (self->data != NULL && self->data->len > 0);
```
bkil-syslogng: Does this handle correctly the case where a `csv` file has 0 lines?

modules/kvtagger/kvtagger.c
bkil-syslogng: If we consider what this function does and where it is used, I don't feel the added value of keeping it.

modules/kvtagger/kvtagger.h
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+#ifndef KVTAGGER_H_INCLUDED
+#define KVTAGGER_H_INCLUDED
+
+#include "parser/parser-expr.h"
+#include "syslog-ng.h"
+#include "template/common-template-typedefs.h"
+
+LogParser *kvtagger_parser_new(GlobalConfig *cfg);
```
bkil-syslogng: Could we drop the `_parser` suffix?

lbudai:
As I'll rename `kvtagger` to `add-contextual-value`, I'm closing this PR and will open a new one.
@bkil-syslogng : the relevant review notes will be processed

bkil-syslogng:
@lbudai Okay. I would also squash most of the commits, because I don't feel that so much internal history adds value. Probably one or two preparation, fixing and refactoring commits could be followed by a single 'introduce kvtagger' commit. Also, either this main commit should carry the basic usage information about this feature, or she who merges this PR should copy&paste it to the merge commit.

lbudai:
@bkil-syslogng > commits are needed for reviewers. 
and I don't think that a new feature should  be implented in one commit message...

PR #1126, opened 2016-07-18
NVRegistry: earlier the NVHandle was 16 bit wide, but it has raised t…

juhaszviktor:
…o 32 bit.

The problem is that the hardcoded limit (65535) wasn't followed this modification
nor in nvregistry, nor its unit test.

Testing this 32 bit limit takes a lot of time and causes very high memory usage
That is why we should drop the unit test which tests this limit

### DIFF
--- a/lib/logmsg/nvtable.h
+++ b/lib/logmsg/nvtable.h
@@ -36,6 +36,8 @@ typedef struct _NVHandleDesc NVHandleDesc;
 typedef gboolean (*NVTableForeachFunc)(NVHandle handle, const gchar *name, const gchar *value, gssize value_len, gpointer user_data);
 typedef gboolean (*NVTableForeachEntryFunc)(NVHandle handle, NVEntry *entry, NVIndexEntry *index_entry, gpointer user_data);
 
+#define NVHANDLE_MAX_VALUE ((NVHandle)-1)
+
 struct _NVIndexEntry
 {
   NVHandle handle;
--- a/lib/logmsg/tests/test_nvtable.c
+++ b/lib/logmsg/tests/test_nvtable.c
@@ -30,7 +30,6 @@
 #include <string.h>
 #include <stdlib.h>
 
-
 /* NVRegistry */
 /* testcases:
  *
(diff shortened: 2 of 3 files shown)


lib/logmsg/nvtable.h
```diff
@@ -36,6 +36,8 @@ typedef struct _NVHandleDesc NVHandleDesc;
 typedef gboolean (*NVTableForeachFunc)(NVHandle handle, const gchar *name, const gchar *value, gssize value_len, gpointer user_data);
 typedef gboolean (*NVTableForeachEntryFunc)(NVHandle handle, NVEntry *entry, NVIndexEntry *index_entry, gpointer user_data);
 
+#define NVHandleMaxValue ((NVHandle)-1)
+
```
bazsi: If this is a macro it should be clearly uppercase

juhaszviktor:
Patch is corrected based on review note

lib/logmsg/tests/test_nvtable.c
```diff
@@ -54,77 +52,6 @@
...
-  reg = nv_registry_new(builtins);
-
-  for (i = 0; builtins[i]; i++)
-    {
-      handle = nv_registry_alloc_handle(reg, builtins[i]);
-      TEST_ASSERT(handle == (i+1));
-      name = nv_registry_get_handle_name(reg, handle, &len);
-      TEST_ASSERT(strcmp(name, builtins[i]) == 0);
-      TEST_ASSERT(strlen(name) == len);
-    }
```
lbudai: As I see, `nv_registry_get_handle_name` is tested only in this testcase, so we should keep this part .

kvch:
I have created a patch, so the test of `nv_registry` don't have to be deleted. 

The patch can be found here:  abcdb6224a466bf862ef19734c4499c7d9e3ffdc
Where should I open a new PR? Here or at jviktor's?

kvch:
After discussion with @juhaszviktor I am closing this PR. I am opening a new one and the issue will be resolved there. The new PR is #1130.

PR #1129, opened 2016-07-23
F/add contextual data

lbudai:
*  What is add-context-data?

    With `add-context-data` syslog-ng can use an external database file to append
    custom name-value pairs on incoming logs (to enrich messages). The 'database'
    is actually a file that containing `<selector, name, value>` records.
    Currently only `CSV` format is supported.
    
    It is like `geoip parser` where the selector is `$HOST`, but in this case,
    the user can define the selector, and also the database contents.

    * options

    ```
    database         | mandatory | file path | path for database (*.csv) file (it
                                               can be a relative path)
    
    selector         | mandatory | template  | will be evaulated for all incoming
                                               messages and when match
                                               to one of the selectors defined in
                                               the database, all the <name, value>
                                               pairs belongs to this selector will
                                               be added to the message
    
    default-selector | optional  | string    | user can define a special selector
                                               in the database file and when there
                                               are no matching selectors for a
                                               message, the <name, value> pairs
                                               identified by the default selector
                                               will be added to the message
    
    prefix           | optional  | string    | prefix will be prepended to all the
                                               <name, value> pairs that are added
                                               to the message in case of match
                                               (including default selector)
    
    ```
    
    * Example:
    
      * context-info-db.csv:
    
    ```
    localhost,localhost_name1,localhost_value1
    localhost,localhost_name1,localhost_value2
    ```
    
       * config

    ```
    @version: 3.8
    @include "scl.conf"
    
    source s_network {
            network(port(5555));
    };
    
    destination d_local {
            file("/tmp/test-msgs.log"
            template("$MSG injected-nvpair:[${mymachine.localhost_name1}]};
    
    parser p_add_context_data {
           add_context_data(selector("$HOST"), database("context-info-db.csv"),
                            prefix("mymachine"));
    };
    
    log {
            source(s_network);
            parser(p_add_context_data);
            destination(d_local);
    };
    ```
    
    Notes to the developers:
    * CONFORMING TO
      We are using the POSIX.1-2008 conform
       * `getline()` instead of `fgets()`
       * `fmemopen()`
    * Code style
```
    	astyle --convert-tabs
                   --indent=spaces=2
                   --max-instatement-indent=80
                   --align-pointer=name --style=gnu
```


faxm0dem:
that's awesome and something I've been wanting for a long time. will a DBI interface be available some time?

lbudai:
@faxm0dem : well... we thought that loading data from an external database during initialization time
doesn't fit into syslog-ng's philosophy (this file can be generated from the database, so syslog-ng not needed to connect to the database directly - but I understand your requirement and in the future it may will part of syslog-ng...)

Our main objection against supporting external DBI were
 * it slows down syslog-ng init time
 * several problems can occur during fetching the data from the database during init time

So the main problem is init.
Maybe we could solve these issues by supporting 'asynchronous init' and 'synchronous init': 
 * when user asks syslog-ng to initialize the external data source in an asynchronous way it should mean
   that syslog-ng won't load all the data during init time and until all the data not loaded, the add-contextual-data parser should queue the incoming messages
 * the synchronous way should mean that the user definitely wants to wait until data loading not finished (this means that init will block...)

ihrwein:
+:beer:

faxm0dem:
if it's init time only, I guess it defeats my use-case.
I understand the implications it has though.
Maybe a memcached solution would make sense for event-time lookup?

bazsi:
Hmm would reloading the table automatically when it changes resolve your
use case?

I mean how often is that changed? Reloading would take a few marks and then
lookup performance would be s lot better.

faxm0dem:
you mean like triggering a `syslog-ng-ctl reload` using an external script that checks for changes?

bazsi:
no, just like with db-parser() which reloads its database when it changes.

-- 
Bazsi

faxm0dem:
ah yes, that would be very convenient indeed! I could write a script that updates the table asynchronously when the DB changes

faxm0dem:
so the question is: will the reload be automatic like for patterndb or do we have to trigger the `syslog-ng-ctl reload` externally? Both would be fine for our use-case, as we manage syslog-ng using puppet.

Cheers

PR #1142, opened 2016-07-29
add-contextual-data: fixed double free bug when csv file contains inv…

lbudai:
…alid line

Reproduction:
* create a csv file that contains two lines
* first is valid, second line invalid (even selector cannot be fetch)

syslog-ng won't start as the csv file is invalid, but the way it stops
not what someone could expect double free or corruption...

Reason:
* when scanner reads the first record, it inserts the record into the
ContextInfoDB (the record is added to the g_array)
* when the scanner reads the second, the invalid record, it will free the
content of the last_record (as it could be in a partially loaded state
which means that we allocate memory for example selector, but we cannot
fetch the name: in this case we have to free the selector as during reload
we would have memory leak), but when the selector is invalid (the first item
we fetch), the last_record will contain pointers that are already passed to
the g_array
* the problem is that when we call the purge_db function, it will free
the g_array which then will free the same memory

Solution:
* move handling responsibility of last_record to the parent class
* get_next_record in parent class should have to init/clean the
last_record to avoid shared state...

### DIFF
--- a/modules/add-contextual-data/context-info-db.c
+++ b/modules/add-contextual-data/context-info-db.c
@@ -257,7 +257,7 @@ context_info_db_import(ContextInfoDB *self, FILE *fp,
     {
       if (line_buf[n - 1] == '\n')
         line_buf[n - 1] = '\0';
-      next_record = scanner->get_next(scanner, line_buf);
+      next_record = contextual_data_record_scanner_get_next(scanner, line_buf);
       if (!next_record)
         {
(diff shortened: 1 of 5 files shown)


lbudai:
Original PR conflicted with #1131 , so I merged that first and then resolved the conflicts on my branch.
Now, it should be fine...

kvch:
:+1:

ihrwein:
:+1:

Issue #1144, opened 2016-08-02
3.8 git head fails to build on SLES 11 -- add contextual data


czanik:
When I try to compile 3.8 git head on SLES 11, it fails with the following error:
```
[   98s]   CC       modules/add-contextual-data/modules_add_contextual_data_libadd_contextual_data_la-add-contextual-data.lo
[   99s]   CC       modules/add-contextual-data/modules_add_contextual_data_libadd_contextual_data_la-csv-contextual-data-record-scanner.lo
[   99s]   CC       modules/add-contextual-data/modules_add_contextual_data_libadd_contextual_data_la-contextual-data-record-scanner.lo
[   99s]   CC       modules/add-contextual-data/modules_add_contextual_data_libadd_contextual_data_la-add-contextual-data-grammar.lo
[   99s]   CC       modules/add-contextual-data/modules_add_contextual_data_libadd_contextual_data_la-add-contextual-data-parser.lo
[   99s]   CC       modules/add-contextual-data/modules_add_contextual_data_libadd_contextual_data_la-context-info-db.lo
[   99s] modules/add-contextual-data/context-info-db.c: In function '_new':
[   99s] modules/add-contextual-data/context-info-db.c:109: warning: implicit declaration of function 'g_array_set_clear_func'
[   99s] modules/add-contextual-data/context-info-db.c: In function 'context_info_db_import':
[   99s] modules/add-contextual-data/context-info-db.c:251: error: 'ssize_t' undeclared (first use in this function)
[   99s] modules/add-contextual-data/context-info-db.c:251: error: (Each undeclared identifier is reported only once
[   99s] modules/add-contextual-data/context-info-db.c:251: error: for each function it appears in.)
[   99s] modules/add-contextual-data/context-info-db.c:251: error: expected ';' before 'n'
[   99s] modules/add-contextual-data/context-info-db.c:256: error: 'n' undeclared (first use in this function)
[   99s] make[2]: *** [modules/add-contextual-data/modules_add_contextual_data_libadd_contextual_data_la-context-info-db.lo] Error 1
[   99s] make[1]: *** [all-recursive] Error 1
[   99s] make: *** [all] Error 2
```
It compiled fine on a recent openSUSE.

lbudai:
`ssize_t` : I can fix this by adding an appropriate include.
`g_array_set_clear_func` : it is part of glib since version 2.32, we will check it

lbudai:
[...]

PR #1148, opened 2016-08-02
add-contextual-data: build fixed on SLES-11

lbudai:
* include missing header
 * as `g_array_set_clear_func` is only available from glib version 2.32,
   I removed it

Fixes: #1144

### DIFF
--- a/lib/cfg-tree.c
+++ b/lib/cfg-tree.c
@@ -1074,7 +1074,7 @@ _verify_unique_persist_names_among_pipes(const GPtrArray *initialized_pipes)
 
       if (current_pipe_name != NULL)
         {
-          if (g_hash_table_contains(pipe_persist_names, current_pipe_name))
+          if (g_hash_table_lookup_extended(pipe_persist_names, current_pipe_name, NULL, NULL))
             {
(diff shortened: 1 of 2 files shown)


lib/cfg-tree.c
```diff
@@ -1074,7 +1075,7 @@ _verify_unique_persist_names_among_pipes(const GPtrArray *initialized_pipes)
 
       if (current_pipe_name != NULL)
         {
-          if (g_hash_table_contains(pipe_persist_names, current_pipe_name))
+          if (g_hash_table_lookup_extended(pipe_persist_names, current_pipe_name, &orig_key, &value))
```
MrAnno: The last two parameters can be NULL, if we don't need them in the future (they hid it well in their [documentation](https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#g-hash-table-lookup-extended): orig_key, value - [optional][nullable]).

MrAnno:
:+1:

lib/cfg-tree.c
lbudai: Thanks, I'll update the PR.

kvch:
:+1:

Issue #1118, opened 2016-07-11
3.8 new JAR files in java-modules directory


czanik:
After updating my openSUSE package to latest git head, some new JAR files appeared under /usr/lib/syslog-ng/java-modules What are these? Are all of these necessary? 

```
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/commons-codec-1.9.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/commons-lang3-3.4.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/commons-logging-1.2.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/gson-2.6.2.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/guava-19.0.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/httpasyncclient-4.1.1.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/httpclient-4.5.2.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/httpcore-4.4.4.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/httpcore-nio-4.4.4.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/jest-2.0.2.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/jest-common-2.0.2.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/slf4j-api-1.7.13.jar
```

(My guess is, that these are dependency for the Elasticsearch HTTP client and related to https://github.com/balabit/syslog-ng/pull/1053 )

lbudai:
@czanik: as you know we are using gradle for dependency management, and if gradle says that these are the dependencis, then the answer is yes, all of them are needed.
Is it a bug or can we close this issue?

czanik:
Until now only a few files were installed here by syslog-ng:
```
%dir %{_libdir}/syslog-ng/java-modules
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/syslog-ng-core.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/dummy.jar
%attr(755,root,root) %{_libdir}/syslog-ng/java-modules/elastic.jar
[...]

PR #1155, opened 2016-08-05
elastic-v2: fixed debian packaging

lbudai:
elastic-v2 is part of the mod-elastic package (plugin.conf is common
for elastic and elastic-v2 drivers).
Jest client is also added to this package.

This may be a little bit confusing, but I have my own reasons:
 * syslog-ng supports Elastisearch: not 1.7, 2.x, 5, but Elasticsearch
 * elastic and elastic-v2 exists just because the Elasticsearch Java clients
   are not compatible between 1.7.x and 2.x versions (and the bad thing
that the package names are almost identical, but internal representation
has been changed)
 * elastic-v2 could be packaged separately, but we have a http mode extension
   to that which is able to work with both ElasticSearch 1.x, 2.x and Elastic5

So the elastic is only for 1.7.x series, elastic-v2 could work with 2.x series
with node, transport and shield mode, and elastic-v2 http mode could work
with all versions.

References: #1118

### DIFF
--- a/debian/syslog-ng-mod-elastic.install
+++ b/debian/syslog-ng-mod-elastic.install
@@ -1,2 +1,4 @@
 usr/lib/syslog-ng/*/java-modules/elastic.jar
+usr/lib/syslog-ng/*/java-modules/elastic-v2.jar
+usr/lib/syslog-ng/*/java-modules/elastic-jest-client/*.jar
 usr/share/syslog-ng/include/scl/elasticsearch/*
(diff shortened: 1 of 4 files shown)


lbudai:
Note: I haven't tried the packaging yet.
(I've just compiled it on my machine and started syslog-ng with an elasticsearch mode http config, but without elasticsearch: so I've checked only the classpath changes.)

lbudai:
update: I've created deb packages in my local build env, then successfully installed syslog-ng

juhaszviktor:
:+1:

PR #1157, opened 2016-08-10
afstreams: include missing header (errno.h)

lbudai:
After this small change I was able to compile syslog-ng on Oracle Solaris 11.
( the environment setup was a little bit s*ck, if someone interested in the details , I'll share my experiences )


MrAnno:
:+1:

PR #1163, opened 2016-08-12
Fix deserialization of GSockAddr

MrAnno:
During deserialization, `sockaddr_in` and `sockaddr_in6` should be zeroed out, to avoid uninitialized fields.


lbudai:
Nice catch ;-)

PR #1171, opened 2016-08-20
F/fix parsing empty last argument in templates

bazsi:
This is a bugfix I had sitting in one of my in-progress branches, that can be merged without the rest. I think this is eligible even for 3.8, as it is a low-risk bugfix with test coverage. However it'd be nice if @bkil-syslogng could review this as it changes both $(format-cef-extension) and the --subkeys option in value-pairs, both of which were authored by him.

    template: fix parsing of empty strings as the last argument of template functions
    
    This patch fixes parsing of empty strings if they occur as the last
    argument of a template function:
    
    $(echo '')
    
    This was parsed as a template function with zero arguments, e.g. the last
    empty string was skipped.

### DIFF
--- a/lib/template/compiler.c
+++ b/lib/template/compiler.c
@@ -300,6 +300,7 @@ static gboolean
 log_template_compiler_process_arg_list(LogTemplateCompiler *self, GPtrArray *result)
 {
   GString *arg_buf = g_string_sized_new(32);
+  gboolean arg_buf_has_a_value = FALSE;
   gint parens = 1;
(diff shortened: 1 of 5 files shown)


bkil-syslogng:
If the status is `needs-review`, I give the current state :+1:

bazsi:
thanks. @lbudai can we merge this to the current master, or you want to wait for 3.8 to be branched?

lbudai:
I think, it can goes to current master.

PR #1172, opened 2016-08-21
Create X509_* macros for CN, O, OU of the peer certifacate

supericeboy:
Having macros for these values will allow syslog-ng servers to use attributes from the clients certificate, if presented, in file name paths.

This would allow servers to use values that are unique per client, but unchangeable by the client.

Combined with peer-verify(require-trusted) we can issue clients a certificate with attributes that they cannot change. We can use those values to uniquely identify that client without relying on the client to put an identifier in the actual log message.

### DIFF
--- a/lib/logproto/logproto-framed-server.c
+++ b/lib/logproto/logproto-framed-server.c
@@ -69,7 +69,7 @@ log_proto_framed_server_prepare(LogProtoServer *s, GIOCondition *cond)
 }
 
 static LogProtoStatus
-log_proto_framed_server_fetch_data(LogProtoFramedServer *self, gboolean *may_read)
+log_proto_framed_server_fetch_data(LogProtoFramedServer *self, gboolean *may_read, LogTransportAuxData *aux)
 {
   gint rc;
 
@@ -91,7 +91,7 @@ log_proto_framed_server_fetch_data(LogProtoFramedServer *self, gboolean *may_rea
     return LPS_SUCCESS;
 
   rc = log_transport_read(self->super.transport, &self->buffer[self->buffer_end], self->buffer_size - self->buffer_end,
-                          NULL);
+                          aux);
 
   if (rc < 0)
     {
@@ -176,13 +176,17 @@ read_frame:
       if (!log_proto_framed_server_extract_frame_length(self, &need_more_data))
         {
           /* invalid frame header */
+          log_transport_aux_data_reinit(aux);
           return LPS_ERROR;
         }
       if (need_more_data && try_read)
         {
-          status = log_proto_framed_server_fetch_data(self, may_read);
+          status = log_proto_framed_server_fetch_data(self, may_read, aux);
           if (status != LPS_SUCCESS)
-            return status;
+            {
+              log_transport_aux_data_reinit(aux);
+              return status;
+            }
           try_read = FALSE;
           goto read_frame;
         }
@@ -195,6 +199,7 @@ read_frame:
               msg_error("Incoming frame larger than log_msg_size()",
                         evt_tag_int("log_msg_size", self->super.options->max_msg_size),
                         evt_tag_int("frame_length", self->frame_len));
+              log_transport_aux_data_reinit(aux);
               return LPS_ERROR;
             }
           if (self->buffer_size < self->super.options->max_buffer_size &&
@@ -247,9 +252,12 @@ read_message:
         }
       if (try_read)
         {
-          status = log_proto_framed_server_fetch_data(self, may_read);
+          status = log_proto_framed_server_fetch_data(self, may_read, aux);
           if (status != LPS_SUCCESS)
-            return status;
+            {
+              log_transport_aux_data_reinit(aux);
+              return status;
+            }
           try_read = FALSE;
           goto read_message;
         }
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -238,6 +238,28 @@ tls_session_set_verify(TLSSession *self, TLSSessionVerifyFunc verify_func, gpoin
   self->verify_data_destroy = verify_destroy;
 }
 
+void
+tls_session_info_callback(const SSL *ssl, int where, int ret)
+{
+  TLSSession *self = (TLSSession *)SSL_get_app_data(ssl);
+  if( !self->peer_info.found && where == (SSL_ST_ACCEPT|SSL_CB_LOOP) )
+    {
+      X509 *cert = SSL_get_peer_certificate(ssl);
+
+      if(cert)
+        {
+          self->peer_info.found = 1; /* mark this found so we don't keep checking on every callback */
+          X509_NAME *name = X509_get_subject_name(cert);
+
+          X509_NAME_get_text_by_NID( name, NID_commonName, self->peer_info.cn, X509_MAX_CN_LEN );
+          X509_NAME_get_text_by_NID( name, NID_organizationName, self->peer_info.o, X509_MAX_O_LEN );
+          X509_NAME_get_text_by_NID( name, NID_organizationalUnitName, self->peer_info.ou, X509_MAX_OU_LEN );
+
+          X509_free(cert);
+        }
+    }
+}
+
 static TLSSession *
 tls_session_new(SSL *ssl, TLSContext *ctx)
 {
@@ -248,6 +270,9 @@ tls_session_new(SSL *ssl, TLSContext *ctx)
 
   /* to set verify callback */
   tls_session_set_verify(self, NULL, NULL, NULL);
+
+  SSL_set_info_callback(ssl, tls_session_info_callback);
+
   return self;
 }
 
@@ -257,6 +282,7 @@ tls_session_free(TLSSession *self)
   if (self->verify_data && self->verify_data_destroy)
     self->verify_data_destroy(self->verify_data);
   SSL_free(self->ssl);
+
   g_free(self);
 }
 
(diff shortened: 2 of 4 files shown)


supericeboy:
Issue #1173 opened

lib/tlscontext.h
```diff
@@ -64,8 +65,21 @@ typedef struct _TLSSession
   TLSSessionVerifyFunc verify_func;
   gpointer verify_data;
   GDestroyNotify verify_data_destroy;
+  TLSPeerInformation *peer_info;
 } TLSSession;
```
bazsi: I think we dont need to use a separate pointer here, just embed the struct here.

lib/tlscontext.c
```diff
@@ -236,6 +236,26 @@ tls_session_set_verify(TLSSession *self, TLSSessionVerifyFunc verify_func, gpoin
...
+	    X509_NAME *name = X509_get_subject_name(cert);
+
+	    X509_NAME_get_text_by_NID( name, NID_commonName, self->peer_info->cn, X509_MAX_CN_LEN );
+	    X509_NAME_get_text_by_NID( name, NID_organizationName, self->peer_info->o, X509_MAX_O_LEN );
+	    X509_NAME_get_text_by_NID( name, NID_organizationalUnitName, self->peer_info->ou, X509_MAX_OU_LEN );
+	   
+	    X509_free(cert);
+    }
+  }
+}
```
bazsi: Pls fix indentation here.

lib/tlscontext.c
```diff
@@ -246,6 +266,11 @@ tls_session_new(SSL *ssl, TLSContext *ctx)
 
   /* to set verify callback */
   tls_session_set_verify(self, NULL, NULL, NULL);
+  
+  self->peer_info = g_new0(TLSPeerInformation, 1);;
+
```
bazsi: We don't need this allocation if peer_info is contained

lib/tlscontext.c
```diff
@@ -255,6 +280,9 @@ tls_session_free(TLSSession *self)
   if (self->verify_data && self->verify_data_destroy)
     self->verify_data_destroy(self->verify_data);
   SSL_free(self->ssl);
+  if( self->peer_info ) {
+    g_free(self->peer_info);
+  }
```
bazsi: Neither the free

lib/tlscontext.h
```diff
@@ -64,8 +65,21 @@ typedef struct _TLSSession
...
+#define X509_MAX_OU_LEN 64
+
+typedef struct _TLSPeerInformation
+{
+  int found;
+  gchar o[X509_MAX_O_LEN];
+  gchar ou[X509_MAX_OU_LEN];
+  gchar cn[X509_MAX_CN_LEN];
+} TLSPeerInformation;
+
```
bazsi: I am not sure 64 characters would be enough. Is there a specified/recommended limit somewhere?

lib/transport/transport-tls.c
```diff
@@ -46,6 +46,13 @@ log_transport_tls_read_method(LogTransport *s, gpointer buf, gsize buflen, LogTr
    * SSL_ERROR_WANT_WRITE is specified by libssl */
   self->super.cond = G_IO_IN;
 
+  /* if we have found the peer has a certificate */
+  if( self->tls_session->peer_info->found ) {
+    log_transport_aux_data_add_nv_pair(aux, "X509_CN", self->tls_session->peer_info->cn );
+    log_transport_aux_data_add_nv_pair(aux, "X509_O", self->tls_session->peer_info->o );
+    log_transport_aux_data_add_nv_pair(aux, "X509_OU", self->tls_session->peer_info->ou );
+  }
+
```
bazsi: Pls fix indentation here (GNU style braces)

lib/logproto/logproto-framed-server.c
```diff
@@ -245,7 +245,7 @@ log_proto_framed_server_fetch(LogProtoServer *s, const guchar **msg, gsize *msg_
         }
       if (try_read)
         {
-          status = log_proto_framed_server_fetch_data(self, may_read);
+          status = log_proto_framed_server_fetch_data(self, may_read, aux);
           if (status != LPS_SUCCESS)
             return status;
           try_read = FALSE;
```
bazsi: Nice one, this can be considered a bug fix.

lib/tlscontext.h
supericeboy: RFC 5280 sets the max length for those.

Also I just noticed OU should be 32 chars.

bazsi:
thanks for your contribition. I generally like the implementation, you nicely reused the log transport aux data concept. 

I only have two concerns:

1) naming: I am not sure about X509_ as a prefix. The Unix domain specific name value pairs use .unix.* as namespace. I think we should use .tls.x509.*, but other opinions welcome. This is however not an end-to-end name-value pair (changes hop-by-hop) so this might not be the best idea. @robertfekete, @lbudai

2) tests: we should add some level of testing, preferably unit tests if possible.

See my other nitpicks in my previous review notes.

fekete-robert:
Hi
First of all, I think this is an excellent feature, many thanks for contributing it!

Regarding the prefix: I agree with @bazsi that X509_ is not consistent with our other prefixes, which are usually something like .prefix.fieldname. The .tls.x509 prefix might be a bit overkill,  unless we plan to add other fields under .tls.

bazsi:
I think we should use .tls.x509_* e.g. only use one level of depth.

With this set, could you pls address the review notes and this change, so
we can merge this?

Cheers,
Bazsi

supericeboy:
All the review issues should be fixed and the macros move to the .tls.x509_* namespace

lib/tlscontext.c
```diff
@@ -236,6 +236,27 @@ tls_session_set_verify(TLSSession *self, TLSSessionVerifyFunc verify_func, gpoin
   self->verify_data_destroy = verify_destroy;
 }
 
+void
+tls_session_info_callback(const SSL *ssl, int where, int ret)
+{
+  TLSSession *self = (TLSSession*)SSL_get_app_data(ssl);
+  if( !self->peer_info_found && where == (SSL_ST_ACCEPT|SSL_CB_LOOP) ) {
+    X509* cert = SSL_get_peer_certificate(ssl);
```
bazsi: please try to use our coding conventions with white spaces. There's even PR #1180 where astyle can help you reformat your patches.

lib/tlscontext.h
```diff
@@ -57,13 +57,21 @@ typedef enum
...
   SSL *ssl;
   TLSContext *ctx;
   TLSSessionVerifyFunc verify_func;
   gpointer verify_data;
   GDestroyNotify verify_data_destroy;
+  int peer_info_found;
+  gchar peer_info_o[X509_MAX_O_LEN];
+  gchar peer_info_ou[X509_MAX_OU_LEN];
+  gchar peer_info_cn[X509_MAX_CN_LEN];
 } TLSSession;
```
bazsi: we may have a misunderstanding here, I didn't want you to inline these into the base struct, only to contain that. But this gave me an idea, what do you think of this:

```
 typedef struct _TLSSession
 {
   SSL *ssl;
   TLSContext *ctx;
   TLSSessionVerifyFunc verify_func;
   gpointer verify_data;
   GDestroyNotify verify_data_destroy;
   struct
   {
     gboolean info_found;
     gchar o[X509_MAX_O_LEN];
     gchar ou[X509_MAX_OU_LEN];
     gchar cn[X509_MAX_CN_LEN];
 } peer_info;
};
```

This way you have tls_session->peer_info.ou as the struct member. What do you think?

bazsi:
This is getting into shape. In order to merge this, we still need:

1) please check travis, why it is red
2) fix the nitpicks I still had
3) clear up the branch history, so internal changes are not visible.

an automated test program would be great, but as I can see there's no framework in the area. If you could help with that, that would also be great.

Cheers,
Bazsi

supericeboy:
I think everything should be resolved now. Not sure how to get the last commit squashed into 1 total commit.

bkil-syslogng:
The git history will need some major cleanup before this can be merged. I'm not sure whether it is a matter of a simple rebase now. Will you be able to fix it yourself, or do you need help? Do you perhaps have any backup of the previous state of the branch?

supericeboy:
I guess I will need help. I did a rebase and only see 2 commits on the merge request, and it's only showing the files I modified, so I don't understand what needs changed.

bkil-syslogng:
At the end of development, we usually rebase to master. This goes like so:
* `git checkout` our own branch
* `git remote add` the original upstream balabit/syslog repo
* `git rebase upstream/master`
* force push your updated branch

From the current tree, it looks like as if you merged master into your fork. If you click on the commits tab on GitHub, you will see an empty commit and a huge commit that seems to replicate some of the style changes, but hardly shows your changes. gitk shows the current state of the trees better.

I've checked out your branch, done a git diff from the time of the divergence and cleaned up the commits.

First save a backup of the current state of your branch (simple create a new branch and push it just in case).
Then add my repo as a remote, fetch from it, then reset your original branch to this one. Review whether you are satisfied with the diff. It prompted me about a rebase conflict with NULL, you should invastigate that too.
https://github.com/bkil-syslogng/syslog-ng/tree/f/supericeboy-ssl-cert-macros

supericeboy:
OK, I think I got it fixed. I followed your instructions and I re-pushed it with just the 1 commit.

mitzkia:
I have run some tests, there is no regression.

juhaszviktor:
In this case we should merge it

MrAnno:
:+1:

PR #1181, opened 2016-08-24
Add experimental third-party unit test framework

MrAnno:
This pull request integrates a new third-party unit test framework into syslog-ng: [**Criterion**](https://github.com/Snaipe/Criterion).

Currently, we have a light test utility library (`libtest/testutils`) with helper macros and unit tests that are following different approaches and styles.
The motivation of this feature was to make it easier to write unified tests without unnecessary boilerplate code.

After a lot of research and comparisons, Criterion seemed to be the suitable choice.

From the documentation of Criterion:

------------------------
Criterion is a dead-simple, non-intrusive unit testing framework for C.

Most test frameworks for C require a lot of boilerplate code to set up tests and test suites – you need to create a main, then register new test suites, then register the tests within these suits, and finally call the right functions.

This gives the user great control, at the unfortunate cost of simplicity.

Criterion follows the KISS principle, while keeping the control the user would have with other frameworks.

- C99 and C++11 compatible
- Tests are automatically registered when declared
- Implements a xUnit framework structure
- A default entry point is provided, no need to declare a main unless you want
  to do special handling
- Test are isolated in their own process, crashes and signals can be reported
  and tested
- Supports parameterized tests and theories
- Progress and statistics can be followed in real time with report hooks

------------------------

Criterion has been integrated as an optional "system" dependency. This can be changed in the future; using the framework as a submodule would be a more portable solution.
The library is searched at configure time via `pkg-config`, if it's not found, Criterion related tests will be skipped.

The PR also contains a few tests, these are mostly just test-ports, they are not completely following the "KISS principles" of Criterion.

More info:
- [Documentation](http://criterion.readthedocs.io/en/master/)
- [Debugging Criterion tests](https://gist.github.com/MrAnno/08ef176958d38f36f1e25b12e203df2b)

Further development opportunities:
- Rewrite/port all unit tests
- Connect it with our CMake build scripts (we need to solve the ['-dlpreopen' problem](https://github.com/MrAnno/syslog-ng/commit/debf763af9c524b0b61d889191ef509c1c135cb9) first)
- Start using a mock library, for example, [Mimick](https://github.com/Snaipe/Mimick)

Criterion


ngergo:
👍

bazsi:
I like the general approach of criterian and am grateful for getting rid of libtest. 
:+1:

PR #1183, opened 2016-08-25
LogQueueFifo: take a reference on logqueue, while the finish callback…

juhaszviktor:
… was not executed

Otherwise it could cause a use-after-free,
when the file destination closes the logwriter at time reap (possible race condition)

There were some core files at user,
and in these core files shows the symptoms of a use-after-free incident

This correction is based on only a guess.
Anyway in the logqueue-fifo, there was nothing which guarantee,
that the queue won't be freed before the callback is called.

### DIFF
--- a/lib/logqueue-fifo.c
+++ b/lib/logqueue-fifo.c
@@ -221,7 +221,7 @@ log_queue_fifo_move_input(gpointer user_data)
   log_queue_push_notify(&self->super);
   g_static_mutex_unlock(&self->super.lock);
   self->qoverflow_input[thread_id].finish_cb_registered = FALSE;
-  log_queue_unref((LogQueue *)self->qoverflow_input[thread_id].cb.user_data);
+  log_queue_unref(&self->super);
   return NULL;
 }
 
--- a/lib/logqueue-fifo.c
+++ b/lib/logqueue-fifo.c
@@ -221,6 +221,7 @@ log_queue_fifo_move_input(gpointer user_data)
   log_queue_push_notify(&self->super);
   g_static_mutex_unlock(&self->super.lock);
   self->qoverflow_input[thread_id].finish_cb_registered = FALSE;
+  log_queue_unref((LogQueue *)self->qoverflow_input[thread_id].cb.user_data);
   return NULL;
 }
 
@@ -267,6 +268,7 @@ log_queue_fifo_push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *pat
            * input thread finishes */
 
           main_loop_worker_register_batch_callback(&self->qoverflow_input[thread_id].cb);
+          self->qoverflow_input[thread_id].cb.user_data = log_queue_ref(s);
           self->qoverflow_input[thread_id].finish_cb_registered = TRUE;
         }
 
@@ -504,7 +506,10 @@ log_queue_fifo_free(LogQueue *s)
   gint i;
 
   for (i = 0; i < log_queue_max_threads; i++)
-    log_queue_fifo_free_queue(&self->qoverflow_input[i].items);
+    {
+      g_assert(self->qoverflow_input[i].finish_cb_registered == FALSE);
+      log_queue_fifo_free_queue(&self->qoverflow_input[i].items);
+    }
 
   log_queue_fifo_free_queue(&self->qoverflow_wait);
   log_queue_fifo_free_queue(&self->qoverflow_output);
@@ -539,7 +544,6 @@ log_queue_fifo_new(gint qoverflow_size, const gchar *persist_name)
     {
       INIT_IV_LIST_HEAD(&self->qoverflow_input[i].items);
       worker_batch_callback_init(&self->qoverflow_input[i].cb);
-      self->qoverflow_input[i].cb.user_data = self;
       self->qoverflow_input[i].cb.func = log_queue_fifo_move_input;
     }
   INIT_IV_LIST_HEAD(&self->qoverflow_wait);
--- a/lib/logqueue-fifo.c
+++ b/lib/logqueue-fifo.c
@@ -265,11 +265,14 @@ log_queue_fifo_push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *pat
         {
           /* this is the first item in the input FIFO, register a finish
            * callback to make sure it gets moved to the wait_queue if the
-           * input thread finishes */
+           * input thread finishes
+           * One reference should be held, while the callback is registered
+           * avoiding use-after-free situation
+           */
 
           main_loop_worker_register_batch_callback(&self->qoverflow_input[thread_id].cb);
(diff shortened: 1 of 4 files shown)


juhaszviktor:
Please do not merge it, even if it seems good.
Because we are waiting for the user's feedback

juhaszviktor:
Still we are waiting for feedback, but hopefully the end of this week (2016.09.23) we will get it

lib/logqueue-fifo.c
```diff
@@ -221,6 +221,7 @@ log_queue_fifo_move_input(gpointer user_data)
   log_queue_push_notify(&self->super);
   g_static_mutex_unlock(&self->super.lock);
   self->qoverflow_input[thread_id].finish_cb_registered = FALSE;
+  log_queue_unref((LogQueue *)self->qoverflow_input[thread_id].cb.user_data);
   return NULL;
```
bazsi: I don't really like this cast here. Why is the user data different to self?

lib/logqueue-fifo.c
```diff
@@ -267,6 +268,7 @@ log_queue_fifo_push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *pat
            * input thread finishes */
 
           main_loop_worker_register_batch_callback(&self->qoverflow_input[thread_id].cb);
+          self->qoverflow_input[thread_id].cb.user_data = log_queue_ref(s);
           self->qoverflow_input[thread_id].finish_cb_registered = TRUE;
```
bazsi: Hmm, user data is set to self here, but I don't exactly see the reason why s would be different from self.

Also is this a per message ref here? I can't see the context, but if it is, a perfect would be useful.

bazsi: Perftest

juhaszviktor: This ref will be called only at the first message of the job, not for all of the messages.
Add ref at the job starting and remove if the job is finished.

bazsi: ok, should be good enough then.

bazsi: still, the "self" vs. "s" question applies.

BTW: how can I approve a response to a review note?

MrAnno: You can add a review summary here: https://github.com/balabit/syslog-ng/pull/1183/files#submit-review
(There should be a green "Review changes" button at the top-right corner of every diff)

juhaszviktor: self vs s:
- If I use self, then I have to use &self->super or casting it: (LogQueue *)self 
- But the input parameter is s and s is a LogQueue.
So I thought, that using is the more simple way (My opionon is that as clean as the using self).
But I you want I will use &self->super

lbudai: @bazsi : there is an  ['Add your review '](https://github.com/balabit/syslog-ng/pull/1183/files#submit-review)  link on the page , if this was your question :-)

juhaszviktor:
Review notes corrected

lib/logqueue-fifo.c
```diff
@@ -539,7 +544,6 @@ log_queue_fifo_new(gint qoverflow_size, const gchar *persist_name)
     {
       INIT_IV_LIST_HEAD(&self->qoverflow_input[i].items);
       worker_batch_callback_init(&self->qoverflow_input[i].cb);
-      self->qoverflow_input[i].cb.user_data = self;
       self->qoverflow_input[i].cb.func = log_queue_fifo_move_input;
```
bazsi: I wouldn't move this initialization either, user data is const throughout the execution of this log queue instance. That's I think better reflected by leaving it here.

lib/logqueue-fifo.c
```diff
@@ -267,6 +268,7 @@ log_queue_fifo_push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *pat
            * input thread finishes */
 
           main_loop_worker_register_batch_callback(&self->qoverflow_input[thread_id].cb);
+          self->qoverflow_input[thread_id].cb.user_data = log_queue_ref(&self->super);
           self->qoverflow_input[thread_id].finish_cb_registered = TRUE;
```
bazsi: Together with the other note further down, I think we should not change user data here, we should just take a ref, and before registering the callback.

I would also document why that ref is needed, probably where the rest of the locking assumptions are documented.

bazsi:
My last note is the user data one, I like the rest.

juhaszviktor:
Code is updated based on the review notes.
If it is pass on review, I should merge these commits into one commit.
After it can be merged, because user confirmed, that these modifications solved their problem

lib/logqueue-fifo.c
```diff
@@ -539,8 +547,8 @@ log_queue_fifo_new(gint qoverflow_size, const gchar *persist_name)
     {
       INIT_IV_LIST_HEAD(&self->qoverflow_input[i].items);
       worker_batch_callback_init(&self->qoverflow_input[i].cb);
-      self->qoverflow_input[i].cb.user_data = self;
       self->qoverflow_input[i].cb.func = log_queue_fifo_move_input;
+      self->qoverflow_input[thread_id].cb.user_data = self;
     }
```
bazsi: I think there's a mistake here. If you look at the earlier initialization that was removed, it uses "i" as index, whereas this uses thread_id, which means that not all user_data slots will be initialized. Actually, I can't even find thread_id as a variable declared, but maybe it is a global from somewhere.

juhaszviktor: Ooops, sorry my bad :(
Correction is comming asap

juhaszviktor: It is corrected with a force push.
If there is no other review note, the PR should be merged

BruceHowells:
I think I'm experiencing this bug in 3.10.1; I get a sigsegv after a template-based file is reaped.

[2017-07-27T09:03:16.893709] Syslog connection closed; fd='12', client='AF_INET(127.0.0.1:56550)', local='AF_INET(0.0.0.0:514)'
[2017-07-27T09:03:16.893841] Closing log transport fd; fd='12'
[2017-07-27T09:04:13.975115] Destination timed out, reaping; template='/data/messages/${HOST}/${YEAR}_${MONTH}_${DAY}.log', filename='/data/messages/localhost/2017_07_27.log'
[2017-07-27T09:04:13.975216] Disk-buffer state saved; filename='/usr/local/var/syslog-ng-00011.qf', qout_length='0', qbacklog_length='0', qoverflow_length='0', qdisk_length='0'

Please advise what other information I can provide.

MrAnno:
Hi,

Are you using a file destination with disk buffer?
If so, note that this use-case is not really necessary unless you are using a network based file system (NFS, CIFS, etc.)

Anyways, you are experiencing a crash, so please open a new issue using our issue template (a backtrace and/or reproduction steps are good starting points for us) .

PR #1186, opened 2016-08-29
Fix loggen crash in multi-threaded SSL mode

### LINKED ISSUE #1182: loggen crashes when using tls
reproduction:
start syslog-ng with the config below (keys does not need)
start loggen:
loggen --inet --rate=1000000 --size=200 --interval=30 --active-connections=10 --quiet --use-ssl 169.254.1.21 514

result:
** Error in `/opt/syslog-ng/bin/loggen': double free or corruption (fasttop): 0x00007f86080010c0 ***
*** Error in `/opt/syslog-ng/bin/loggen': double free or corruption (fasttop): 0x00007f85f8001690 ***
*** Error in `/opt/syslog-ng/bin/loggen': double free or corruption (!prev): 0x00007f860c001360 ***

gdb:
(gdb) bt
#0  0x00007f559ee6cc37 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007f559ee70028 in __GI_abort () at abort.c:89
#2  0x00007f559eea92a4 in __libc_message (do_abort=do_abort@entry=1, fmt=fmt@entry=0x7f559efb76b0 "*** Error in `%s': %s: 0x%s ***\n") at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007f559eeb555e in malloc_printerr (ptr=<optimized out>, str=0x7f559efb7878 "double free or corruption (fasttop)", action=1) at malloc.c:4996
#4  _int_free (av=<optimized out>, p=<optimized out>, have_lock=0) at malloc.c:3840
#5  0x00007f559f982f8d in CRYPTO_free () from /lib/x86_64-linux-gnu/libcrypto.so.1.0.0
#6  0x00007f559f985891 in OBJ_NAME_add () from /lib/x86_64-linux-gnu/libcrypto.so.1.0.0
#7  0x00007f559fd40fb1 in SSL_library_init () from /lib/x86_64-linux-gnu/libssl.so.1.0.0
#8  0x0000000000403e76 in ?? ()
#9  0x0000000000402d3a in active_thread ()
#10 0x00007f559f486f05 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#11 0x00007f559f203184 in start_thread (arg=0x7f55a015b700) at pthread_create.c:312
#12 0x00007f559ef3037d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111

root@test20:~# syslog-ng -V
syslog-ng 3.8.1
Installer-Version: 3.8.1
Revision: 3.8.1-1

(3.7 is also affected)
-----
syslog-ng config:
root@test21:/opt/syslog-ng# cat etc/syslog-ng.conf 
@version: 3.8
options {
    keep_hostname(no);
    threaded(yes);
    stats_level(2);
    use_dns(no);
};
source s_network_fa30d29831b34992ab2ef3396f4d2eb2 {
    network(ip(0.0.0.0)
 log_fetch_limit(1000)
 log_iw_size(10000)
 max_connections(10)
 port(514)
 tls(peer_verify(optional-untrusted) cert_file("/var/testdb_working_dir/7532e934-fb97-49dc-8f2b-2169f0952a9a.txt") key_file("/var/testdb_working_dir/d185686d-a039-4ece-bd7b-522733f00d10.txt"))
 transport("tls"));
};
destination d_file_622fb85dc0f4431bb3107d388dabb713 {
    file("/var/testdb_working_dir/cb108da5-0ab7-4221-b48a-6d225f17a8c2" flush_lines(1000)
 log_fifo_size(20000));
};
log {
source(s_network_fa30d29831b34992ab2ef3396f4d2eb2);
destination(d_file_622fb85dc0f4431bb3107d388dabb713);

flags(flow-control);
};

MrAnno:
Using the `--active-connections >1` flag together with `--use-ssl` can cause OpenSSL-related crashes, since OpenSSL's locking_function and threadid_func are not set.

This is the minimum requirement of using OpenSSL in a multi-threaded environment.

Fixes #1182

### DIFF
--- a/tests/loggen/loggen.c
+++ b/tests/loggen/loggen.c
@@ -88,6 +88,8 @@ guint64 sum_count;
 struct timeval sum_time;
 gint raw_message_length;
 
+SSL_CTX *ssl_ctx;
+
 typedef ssize_t (*send_data_t)(void *user_data, void *buf, size_t length);
 
 static ssize_t
@@ -559,20 +561,10 @@ gen_messages_ssl(int sock, int id, FILE *readfrom)
 {
   int ret = 0;
   int err;
-  SSL_CTX *ctx;
   SSL *ssl;
 
-  /* Initialize SSL library */
-  OpenSSL_add_ssl_algorithms();
-
-  if (NULL == (ctx = SSL_CTX_new(SSLv23_client_method())))
+  if (NULL == (ssl = SSL_new(ssl_ctx)))
     return 1;
-
-  if (NULL == (ssl = SSL_new(ctx)))
-    return 1;
-
-  SSL_load_error_strings();
-  ERR_load_crypto_strings();
 
   SSL_set_fd (ssl, sock);
   if (-1 == (err = SSL_connect(ssl)))
@@ -588,7 +580,6 @@ gen_messages_ssl(int sock, int id, FILE *readfrom)
 
   /* Clean up. */
   SSL_free (ssl);
-  SSL_CTX_free (ctx);
 
   return ret;
 }
@@ -608,6 +599,49 @@ gboolean threads_stop;
 gint active_finished;
 gint connect_finished;
 guint64 sum_count;
+
+/* Multithreaded OpenSSL */
+GMutex **ssl_lock_cs;
+
+static void
+ssl_threadid_callback(CRYPTO_THREADID *tid)
+{
+  CRYPTO_THREADID_set_pointer(tid, g_thread_self());
+}
+
+static void
+ssl_locking_callback(int mode, int n, const char *file, int line)
+{
+  if (mode & CRYPTO_LOCK)
+    g_mutex_lock(ssl_lock_cs[n]);
+  else
+    g_mutex_unlock(ssl_lock_cs[n]);
+}
+
+static void
+ssl_thread_setup(void)
+{
+  int i;
+  ssl_lock_cs = g_malloc(CRYPTO_num_locks() * sizeof(GMutex *));
+
+  for (i = 0; i < CRYPTO_num_locks(); ++i)
+    ssl_lock_cs[i] = g_mutex_new();
+
+  CRYPTO_THREADID_set_callback(ssl_threadid_callback);
+  CRYPTO_set_locking_callback(ssl_locking_callback);
+}
+
+static void
+ssl_thread_cleanup(void)
+{
+  int i;
+  CRYPTO_set_locking_callback(NULL);
+
+  for (i = 0; i < CRYPTO_num_locks(); ++i)
+    g_mutex_free(ssl_lock_cs[i]);
+
+  g_free(ssl_lock_cs);
+}
 
 gpointer
 idle_thread(gpointer st)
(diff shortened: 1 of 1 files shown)


tests/loggen/loggen.c
```diff
@@ -88,6 +88,8 @@ guint64 sum_count;
 struct timeval sum_time;
 gint raw_message_length;
 
+SSL_CTX *ssl_ctx;
+
 typedef ssize_t (*send_data_t)(void *user_data, void *buf, size_t length);
```
bazsi: I don't see why this needs to be global.

tests/loggen/loggen.c
```diff
@@ -609,6 +600,49 @@ gint active_finished;
...
+  int i;
+  CRYPTO_set_locking_callback(NULL);
+
+  for (i = 0; i < CRYPTO_num_locks(); ++i)
+    g_mutex_free(ssl_lock_cs[i]);
+
+  g_free(ssl_lock_cs);
+}
+
 gpointer
```
bazsi: we could add all this code perhaps as a separate module? I understand that we don't want to use libsyslog-ng.so (where somewhat similar code is already present), but still we could make loggen.c a bit less convoluted by splitting it up.

MrAnno: Thanks @bazsi, you are right. I just noticed that the same lock functionality exists in `lib/crypto.c`.
Maybe I could link against `lib/crypto` statically and use `crypto_init_threading()` instead. What do you think?

bazsi: Until now, loggen was completely independent from the core, and only
depends on the std library, to be able to run on boxes that do not run
syslog-ng.

This could be reconsidered though.

MrAnno:
After a short discussion with @bazsi, he suggested the following:

We already have separate tools (pdbtool, loggen, dqtool, ...) which are more or less independent from the core of syslog-ng. For example, loggen only needs the `lib/crypto` "sublibrary" to fix #1182.

Because `libsyslog-ng` contains a lot of not-so-reusable code (`apphook`, `mainloop`, etc.), it would be a good idea to

- split libsyslog-ng into two parts (e.g. syslog-ng-core, syslog-ng-rt ?):
  - reusable libraries (general helpers, LogMessage API, plugin/module API, ...)
  - monolithic parts which are strongly coupled and can not be reused

This could be an "incremental process".

@juhaszviktor @lbudai  @bkil-syslogng What do you think of this idea? Can I start this with extracting `lib/crypto` from `libsyslog-ng`?

juhaszviktor:
Sounds good to me :)

lbudai:
@MrAnno: sorry for the late answer, I also like this approach.(This perfectly fits to one of my ideas.)

bazsi:
One thing is that the reusability/assumptions of a given code should be
carefully evaluated before moving to the runtime lib. For instance the
refcache portion of LogMessage is far from being free of assumptions.

Issue #1189, opened 2016-09-02
configure.ac could not find the system-provided libmongoc


mochrul:
Tis is because the name of the pc file is wrong in the configure.ac

presidento:
As I see, it was fixed in #1190.

PR #1190, opened 2016-09-02
Fixed the name of the libmongoc pc file

mochrul:
fixes issue: #1189 

Signed-off-by: SZALAY Attila <sasa@debian.org>


bazsi:
I didn't follow the libmongo client library swap to libmongo-client, but mongo-c-driver really seems to use the libmongoc-1.0.pc as its pkg-config data file.

bazsi:
So, this should be an ok change, but @bkil-syslogng should know better.

:+1:

lbudai:
:+1:

Issue #1191, opened 2016-09-02
Error when the system provided librabbitmq-dev is not found


mochrul:
There is a syntactical error in the line whenchecking the system provided librabbitmq-dev.

If it is not found, then an error happen and the amqp module will not be turned off.

ps.: The checking logic is bit weird, but that is another story.

bazsi:
I guess, since the branch above is merged, this can be closed.

PR #1192, opened 2016-09-02
Fixed the case when system librabbitmq does not found

mochrul:
Fixed issue: #1191 

Signed-off-by: SZALAY Attila sasa@debian.org


bazsi:
:+1:

PR #1193, opened 2016-09-02
Fix time compat library on Mach-based distributions

### LINKED ISSUE #1185: Cannot compile on hurd-i386
I've created a new debian package from version 3.8.1 and uploaded it to the debian experimanetal, where the build daemons try to compile it one every architecture debian is supporting.

Unfortunately, on some architecture this is failing. In most of the time it is because of the missing systemd, which is on me, but on hurd-i386 there is a special issue with a missing file:

```
syslog-ng Open Source Edition 3.8.1 configured
 Compiler options:
  compiler                    : gcc
  compiler options            : -g -O2 -fdebug-prefix-map=/«PKGBUILDDIR»=. -fstack-protector-strong -Wformat -Werror=format-security -Wall -pthread -Wdate-time -D_FORTIFY_SOURCE=2 -pthread -I/usr/include/glib-2.0 -I/usr/lib/i386-gnu/glib-2.0/include -I/usr/include/eventlog   -DLIBNET_LIL_ENDIAN -I/usr/include/dbi -I/usr/include/dbi/dbi  -D_GNU_SOURCE -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64
  linker flags                : -Wl,-z,relro 
  prefix                      : /usr
  linking mode                : dynamic
  embedded crypto             : no
  __thread keyword            : yes
 Submodules:
  ivykis                      : system
  mongoc                      : internal
  librabbitmq                 : internal
  jsonc                       : system
 Features:
  Debug symbols               : no
  GCC profiling               : no
  Memtrace                    : no
  IPV6 support                : yes
  spoof-source support        : yes
  tcp-wrapper support         : yes
  Linux capability support    : no
  Env wrapper support         : no
  systemd support             : no (unit dir: none)
  systemd-journal support     : no
  libmongo-client options     : yes
 Modules:
  Module search path          : ${exec_prefix}/lib/syslog-ng/3.8
  Sun STREAMS support (module): no
  SQL support (module)        : yes
  PACCT module (EXPERIMENTAL) : no
  MongoDB destination (module): yes
  JSON support (module)       : yes
  SMTP support (module)       : yes
  HTTP support (module)       : no
  AMQP destination (module)   : yes
  STOMP destination (module)  : yes
  GEOIP support (module)      : yes
  Redis support (module)      : yes
  Riemann destination (module): yes
  python                      : yes (pkg-config package: python)
  java                        : no
  java modules                : no
  native bindings             : auto
make[1]: Leaving directory '/«PKGBUILDDIR»'
   debian/rules override_dh_auto_build
make[1]: Entering directory '/«PKGBUILDDIR»'
install -d debian/build-tree/modules/afamqp/rabbitmq-c/librabbitmq
dh_auto_build
	make -j1
make[2]: Entering directory '/«PKGBUILDDIR»/debian/build-tree'
make --no-print-directory all-recursive
gcc -DHAVE_CONFIG_H -I. -I../..  -I../../lib -I../../modules -I./lib -I./modules	-fPIC -fvisibility=hidden -I../../modules/native -I./modules/native -Wdate-time -D_FORTIFY_SOURCE=2 -pthread -I/usr/include/glib-2.0 -I/usr/lib/i386-gnu/glib-2.0/include -I/usr/include/eventlog   -DLIBNET_LIL_ENDIAN -I/usr/include/dbi -I/usr/include/dbi/dbi  -D_GNU_SOURCE -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64  -g -O2 -fdebug-prefix-map=/«PKGBUILDDIR»=. -fstack-protector-strong -Wformat -Werror=format-security -Wall -pthread -c -o modules/native/modules_native_libsyslog_ng_native_connector_a-native-grammar.o `test -f 'modules/native/native-grammar.c' || echo '../../'`modules/native/native-grammar.c
In file included from ../../lib/timeutils.h:29:0,
                 from ../../lib/msg-format.h:29,
                 from ../../lib/logmsg/logmsg.h:34,
                 from ../../lib/parser/parser-expr.h:28,
                 from ../../modules/native/native-parser.h:30,
                 from modules/native/native-grammar.y:26:
../../lib/compat/time.h:33:24: fatal error: mach/clock.h: No such file or directory
 #include <mach/clock.h>
                        ^
compilation terminated.
Makefile:8790: recipe for target 'modules/native/modules_native_libsyslog_ng_native_connector_a-native-grammar.o' failed
make[4]: *** [modules/native/modules_native_libsyslog_ng_native_connector_a-native-grammar.o] Error 1
Makefile:13172: recipe for target 'all-recursive' failed
make[3]: *** [all-recursive] Error 1
Makefile:5400: recipe for target 'all' failed
make[2]: *** [all] Error 2
make[2]: Leaving directory '/«PKGBUILDDIR»/debian/build-tree'
```

The full build log is here:
https://buildd.debian.org/status/fetch.php?pkg=syslog-ng&arch=hurd-i386&ver=3.8.1-1&stamp=1472251725

MrAnno:
Fixes #1185

### DIFF
--- a/lib/compat/time.c
+++ b/lib/compat/time.c
@@ -23,7 +23,7 @@
 
 #include "compat/time.h"
 
-#ifdef __MACH__
+#if !defined(SYSLOG_NG_HAVE_CLOCK_GETTIME) && defined(__APPLE__) && defined(__MACH__)
 
 int
 clock_gettime(clock_t clock_id, struct timespec *timestamp)
--- a/lib/compat/time.h
+++ b/lib/compat/time.h
@@ -28,7 +28,7 @@
 #include "compat/compat.h"
 #include <time.h>
 
-#ifdef __MACH__
+#if !defined(SYSLOG_NG_HAVE_CLOCK_GETTIME) && defined(__APPLE__) && defined(__MACH__)
 
 #include <mach/clock.h>
 #include <mach/mach.h>
@@ -44,6 +44,6 @@ int clock_gettime(clock_t clock_id, struct timespec *timestamp);
 #define CLOCK_MONOTONIC CLOCK_REALTIME
 #endif
 
-#endif /* __MACH__ */
+#endif /* !SYSLOG_NG_HAVE_CLOCK_GETTIME && __APPLE__ && __MACH__ */
 
 #endif /* COMPAT_TIME_H_INCLUDED */


lib/compat/time.c
```diff
@@ -23,7 +23,7 @@
 
 #include "compat/time.h"
 
-#ifdef __MACH__
+#ifdef __APPLE__
```
bkil-syslogng: According to the man page, `clock_gettime()` conforms to
* `SUSv2`
* `POSIX.1-2001`

It also mentions the `glibc` feature check macro
* `_POSIX_C_SOURCE >= 199309L`

Do we have our macros set up correctly?

Which version of OSX does this workaround involve?

They say that Mac OS X is UNIX 03 / POSIX certified since 10.5:
* http://www.opengroup.org/openbrand/register/brand3555.htm
* https://blog.opengroup.org/2012/07/25/apple-registers-mac-os-x-10-8-mountain-lion-to-the-unix-03-standard/
* http://www.opengroup.org/openbrand/register/brand3612.htm

If these hints don't help, as a workaround, I wanted to recommend checking for `clock_gettime()` in `configure` to set a variable, but it looks like we already check it and set `SYSLOG_NG_HAVE_CLOCK_GETTIME` accordingly. Could you double check?

lbudai:
@bkil-syslogng :
what I know:
 * OSX is based on Darwin kernel which is based on XNU which contains a Mach microkernel
 * Hurd is also based on Mach , 
so `__MACH__` is defined on both OSX and Hurd.

On systems where `__APPLE__` is defined (OSX) there is no `clock_gettime`, so the patch seems to me correct.

bkil-syslogng:
We've discussed it in detail yesterday and I recall @MrAnno said he knows how to proceed. `clock_gettime` is a libc call and not related to the kernel used in this sense, so checking for the kernel is a bit sketchy. (It should probably be fixed at all occurrences in the codebase later eventually.)

They actually recommend a combination here:
*  https://stackoverflow.com/questions/1529031/what-c-preprocessor-conditional-should-i-use-for-os-x-specific-code

bazsi:
Hmm the libc and the kernel are usually pretty coupled (I know about
uclibc). The best would to do a specific test, eg check if the system
supports clock_gettime(), and if needed also check which header declares
it. Then write the code according to the relevant check results.

I don't like OS centric conditionals in general.

bkil-syslogng:
Yep. Those were my exact same recommendations above.

lib/compat/time.c
MrAnno: Thanks, I'll check the value of the mentioned `_POSIX_C_SOURCE` macro on OS X.

MrAnno:
Thanks for the suggestions.
If `_POSIX_C_SOURCE` doesn't solve the problem, I'll update the PR with a configure-time `clock_gettime() ` test.

MrAnno:
As it turned out, `clock_gettime()` is really not implemented on OS X (10.11 El Capitan), `_POSIX_C_SOURCE` doesn't help at all.

Because our current implementation of `clock_gettime()` in `lib/compat/time.c` is fully "Apple specific code" (it is not available on other Mach based systems), I think the configure-time `HAVE_CLOCK_GETTIME` check is not enough to handle this situation.

Note: I looked into other platform-independent, precise and monotonic time functions, but OS X is always an exception.

bkil-syslogng:
:+1: :penguin:

lbudai:
@bazsi: yeah, OS-centric code is not an elegant / not the best solution, but in this case I think this patch is good enough.

PR #1194, opened 2016-09-02
F/syslog parse date speed up

bkil-syslogng:
Resolve some of the bottlenecks when processing syslog source, 2-3x speedup.

Some more bottlenecks remain to be investigated, but the implemented parts (existing commits) can already be reviewed.

### DIFF
--- a/modules/syslogformat/syslog-format.c
+++ b/modules/syslogformat/syslog-format.c
@@ -421,13 +421,11 @@ __get_normalized_time(LogStamp const timestamp, gint const normalized_hour, gint
 
 /* FIXME: this function should really be exploded to a lot of smaller functions... (Bazsi) */
 static gboolean
-log_msg_parse_date(LogMessage *self, const guchar **data, gint *length, guint parse_flags, glong assume_timezone)
+log_msg_parse_date_unnormalized(LogMessage *self, const guchar **data, gint *length, guint parse_flags, struct tm *tm)
 {
   const guchar *src = *data;
   gint left = *length;
   GTimeVal now;
-  struct tm tm;
-  gint unnormalized_hour;
 
   cached_g_current_time(&now);
 
@@ -437,13 +435,15 @@ log_msg_parse_date(LogMessage *self, const guchar **data, gint *length, guint pa
        * unsynced, '.' if it is known to be synced */
       if (G_UNLIKELY(src[0] == '*'))
         {
-          log_msg_set_value(self, is_synced, "0", 1);
+          if (!(parse_flags & LP_NO_PARSE_DATE))
+            log_msg_set_value(self, is_synced, "0", 1);
           src++;
           left--;
         }
       else if (G_UNLIKELY(src[0] == '.'))
         {
-          log_msg_set_value(self, is_synced, "1", 1);
+          if (!(parse_flags & LP_NO_PARSE_DATE))
+            log_msg_set_value(self, is_synced, "1", 1);
           src++;
           left--;
         }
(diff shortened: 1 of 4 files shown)


bazsi:
Nice refactor. 

FYI: The LP_XXX flags are an ugly layering violation, at one point we should probably fix that part up, but this patch is clearly a step forward. so nice one, :+1:

.travis.yml
```diff
@@ -94,7 +95,7 @@ matrix:
           astyle
       before_script:
       script:
-      - scripts/style-checker.sh check
+      - scripts/style-checker.sh diff
```
lbudai: Do we really want to put the diff in the travis log?

bkil-syslogng: Yes, that was what we have agreed upon some weeks ago, wasn't it? We could
do fine tuning of diff options, though.

It helped me fix the issues detected in this case for example.

juhaszviktor:
The travis/stylechecker patch is not in this context

scripts/style-checker.sh
```diff
@@ -52,12 +52,35 @@ function astyle_c_format
     exit 0
 }
 
+function astyle_c_diff
+{
+    astyle --options="$root_dir/.astylerc" "$root_dir/*.c" |
+    grep "Formatted" |
```
kvch: Why are these lines needed? Why is simply calling `git diff` not enough?

scripts/style-checker.sh
```diff
@@ -52,12 +52,35 @@ function astyle_c_format
     exit 0
 }
 
+function astyle_c_diff
```
kvch: AFAIK we decided that diff view is unecessary.

.travis.yml
kvch: First we thought it was a good idea. But than, it was discarded.

modules/syslogformat/syslog-format.c
```diff
@@ -437,28 +435,29 @@ log_msg_parse_date(LogMessage *self, const guchar **data, gint *length, guint pa
...
           src++;
           left--;
         }
       else if (G_UNLIKELY(src[0] == '.'))
         {
-          log_msg_set_value(self, is_synced, "1", 1);
+          if (!(parse_flags & LP_NO_PARSE_DATE))
+            log_msg_set_value(self, is_synced, "1", 1);
           src++;
           left--;
```
juhaszviktor: We should factor out the setting of the is_synced, shouldn't we?

scripts/style-checker.sh
bkil-syslogng: I don't understand the question. Which lines? Enough for what? Calling it when after what?

.travis.yml
bkil-syslogng: Not sure when it was discarded, it was accepted on the meetings that I have
attended.

I still think it is a very useful feature. Giving me a red exclamation mark
without any detailed reason can put me off. Copyright checker also gives
hints about the fixes. gcc gives a hint about what an error means, it
doesn't just say that one of the specified files is in error.

bkil-syslogng:
You mean we should split off the style checker patch from this PR?

scripts/style-checker.sh
kvch: I mean lines 58-65. AFAIK the sources are formatted, so those files are changed. This, it is enough to simply call `git diff` after formatting files using `astyle`.

bkil-syslogng:
I've removed the [style checker patch](https://github.com/bkil-syslogng/syslog-ng/commit/9dad1cb2d4e02d302fbc9a7bffc772606179235d) from this PR, but it should be merged as well soon as per above.

PR #1195, opened 2016-09-05
logproto: do not enable position tracking by default

lbudai:
There was a performance issue around TCP source.
As it came out LateAckTracker was used instead of EarlyAckTracker.
The reason was that position tracking was enabled by default for the
LogProto type that is used for TCP source.


kvch:
:+1:

PR #1196, opened 2016-09-05
F/pytest framework core micek 6

mitzkia:
This is a big refactor. 
Actual tests can be broke, but I working to pass them.


mitzkia:
Sorry about that. I wanted to send this PR to another fork.

PR #1208, opened 2016-09-12
Fix automatic libcap detection

### LINKED ISSUE #1207: Invalid libcap detection
If the `--enable-linux-caps` flag is not specified **explicitly** when running the configure script, syslog-ng will be compiled without capability support (`SYSLOG_NG_ENABLE_LINUX_CAPS` is always set to 0).

```
  --enable-linux-caps     Enable support for managing Linux capabilities (default: auto)
```

MrAnno:
If the enable-linux-caps flag is not specified explicitly (`enable_linux_caps="auto"`), the corresponding `ENABLE_...` macro in syslog-ng-config.h will be set to a constant FALSE value.

This PR fixes this issue by setting the value of `enable_linux_caps` after calling `AC_CHECK_LIB`.

Fixes #1207


bkil-syslogng:
:+1: :penguin:

PR #1213, opened 2016-09-28
switch to Criterion

kvch:
Following tests were ported
 * `lib/tests/test_cache`
 * `tests/unit/test_zone`
 * `tests/unit/test_logwriter`
 * `tests/unit/test_cache`

Now every test under tests/unit folder depends on Criterion.

### DIFF
--- a/tests/unit/test_logwriter.c
+++ b/tests/unit/test_logwriter.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009-2014 Balabit
+ * Copyright (c) 2009-2016 Balabit
  * Copyright (c) 2009-2014 Viktor Juhasz
  *
  * This program is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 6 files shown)


tests/unit/test_logwriter.c
```diff
@@ -83,72 +156,46 @@ testcase(gchar *msg_string, gboolean input_is_rfc5424, gchar *template, guint wr
...
+    {MSG_BSD_STR, FALSE, NULL, LW_FORMAT_PROTO, EXPECTED_MSG_BSD_TO_PROTO_STR},
+    {MSG_BSD_STR, FALSE, NULL, LW_SYSLOG_PROTOCOL, EXPECTED_MSG_BSD_TO_SYSLOG_STR},
+    {MSG_BSD_EMPTY_STR, FALSE, NULL, LW_FORMAT_PROTO, EXPECTED_MSG_BSD_EMPTY_STR},
+    {MSG_BSD_EMPTY_STR, FALSE, "$PID", LW_FORMAT_PROTO, EXPECTED_MSG_BSD_EMPTY_STR_T},
+    {MSG_ZERO_PRI, FALSE, NULL, LW_FORMAT_PROTO, EXPECTED_MSG_ZERO_PRI_STR},
+    {MSG_ZERO_PRI, FALSE, "$PRI", LW_FORMAT_PROTO, EXPECTED_MSG_ZERO_PRI_STR_T},
+  };
+  gint i, nr_of_cases;
+
   app_startup();
```
lbudai: It is racy.
Could you put it to a `attribute constructor`

tests/unit/test_zone.c
```diff
@@ -71,742 +79,727 @@ timezone_exists(const char *time_zone)
...
-#ifdef __linux__
-  /* Oct 31 01:59:59 2004 (EST) +1000 */
-  testcase("Australia/Victoria", 1099151999, 10*3600);
-  /* Oct 31 03:00:00 2004 (EST) +1100 */
-  testcase("Australia/Victoria", 1099152000, 11*3600);
-  /* Mar 27 02:59:59 2005 (EST) +1100 */
-  testcase("Australia/Victoria", 1111852799, 11*3600);
-  /* Mar 27 02:00:00 2005 (EST) +1000 */
-  testcase("Australia/Victoria", 1111852800, 10*3600);
+TestSuite(hostid, .init = app_startup, .fini = app_shutdown);
```
lbudai: It is racy.
Could you put it to a `attribute constructor`

bazsi:
I don't see the race, care to explain?

lbudai:
@bazsi: criterion runs testcases in a parallel way and when we follow the register `app_startup` and `app_shutdown` as setup/teardown functions for each test cases pattern, test will crash
(even if we have only one testcase, it's a bad practice  as when someone will extend the testcases and will follow the same Testsuite initialization pattern, the test will be broken )

tests/unit/test_pathutils.c
```diff
@@ -22,29 +22,59 @@
...
+  gint i, nr_of_extension_cases, nr_of_extensionless_cases;
+
+  nr_of_extension_cases = sizeof(test_cases_with_extension) / sizeof(test_cases_with_extension[0]);
+  for (i = 0; i < nr_of_extension_cases; i++)
+    _assert_filename_with_extension(test_cases_with_extension[i]);
 
-  return 0;
+  nr_of_extensionless_cases = sizeof(test_cases_without_extension) / sizeof(test_cases_without_extension[0]);
+  for (i = 0; i < nr_of_extensionless_cases; i++)
+    _assert_filename_without_extension(test_cases_without_extension[i]);
```
lbudai: Maybe this could be factored out as a function... (or macro...).

kvch: I would not use macros. I don't see why these should be extracted into a function. This form is readable and easy to understand. But tell me if I am wrong.

lbudai: I can live with this as-is...
( for-cycle is copy-pasted ;)
other idea: what if you would create two functions for the two different testcases (SRP)? Then you can keep your for-cycle in-line :) )

tests/unit/test_logwriter.c
```diff
@@ -83,72 +156,46 @@ testcase(gchar *msg_string, gboolean input_is_rfc5424, gchar *template, guint wr
...
-  if (templ)
-    log_template_unref(templ);
-  log_pipe_unref((LogPipe *) writer);
-  log_msg_unref(msg);
-  log_queue_unref(queue);
-  g_string_free(res, TRUE);
+  log_writer_format_log(writer, msg, result_msg);
+  cr_assert_str_eq(result_msg->str, c.expected_value, "Expected: %s, actual: %s", c.expected_value, result_msg->str);
+
+  _tear_down(templ, writer, msg, queue, result_msg);
```
MrAnno: If you had implemented this assert function as a [parameterized test case](http://criterion.readthedocs.io/en/master/parameterized.html#adding-parameterized-tests), you would have used Criterion's [teardown fixture](http://criterion.readthedocs.io/en/master/starter.html#fixtures) here.

Your solution is not redundant, maybe it is even easier to debug, so I'm OK with this.

PR #1214, opened 2016-09-29
afsocket: tcp: Include missing setsockopt header

kylemanna:
The #ifdef macros disable the TCP_KEEPIDLE, TCP_KEEPCNT, TCP_KEEPINTVL
set socket options as the header file that defines them is not included.

Include the necessary header file to enable the options.

Signed-off-by: Kyle Manna <kyle@kylemanna.com>

### DIFF
--- a/modules/afsocket/socket-options-inet.c
+++ b/modules/afsocket/socket-options-inet.c
@@ -25,6 +25,7 @@
 #include "messages.h"
 
 #include <string.h>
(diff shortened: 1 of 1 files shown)


bazsi:
Good catch, thanks for your contribution. It may have got broken when that code was reorganised.

:+1:

kvch:
Thanks!

PR #1215, opened 2016-09-29
debian: move debian dir to release branch

kvch:
From now on debian directory can be found on the branch named `release`. That branch has to be rebased before creating a new deb package.


lbudai:
@bazsi: we could talk about this IRL

bazsi:
I don't want to keep this open until that happens. So on my part this can be merged

lbudai:
Every future syslog-ng release will be based on the release branch, so the dist tarball will contain the debian directory.

PR #1218, opened 2016-10-01
dnscache: fix cache list corruption causing SIGABRT

bazsi:
As reported on the mailing list, DNS cache corruption occurs when the dns-cache-size limit is reached.
Kudos for the great report go to Claus Albøge.

This patch fixes a SIGABRT that happens when an element is being
removed from the DNS cache, due to the high number of elements already
present. Both the cache_list and the persist_list lists were incorrectly
handled.

### DIFF
--- a/lib/dnscache.c
+++ b/lib/dnscache.c
@@ -159,12 +159,12 @@ dns_cache_store(DNSCache *self, gboolean persistent, gint family, void *addr, co
   if (!persistent)
     {
       entry->resolved = cached_g_current_time_sec();
-      iv_list_add(&self->cache_list, &entry->list);
+      iv_list_add(&entry->list, &self->cache_list);
     }
   else
     {
       entry->resolved = 0;
-      iv_list_add(&self->persist_list, &entry->list);
+      iv_list_add(&entry->list, &self->persist_list);
     }
   hash_size = g_hash_table_size(self->cache);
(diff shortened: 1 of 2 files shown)


clausa:
Thanks for the fix.

As mentioned on the mailing list, it is possible to work-around this issue (if running a release, without this fix) by setting dns-cache-size to a number that is larger than the total active clients. Default for dns-cache-log is 1007

bazsi:
The issue is not a race, but rather an incorrect handling of the LRU list,
so any time you have more hosts than dns-cache-size(), you'll get this
crash, that's an easy DoS, you just need to send a number of messages from
different 10000 IP addresses.

-- 
Bazsi

bazsi:
I have now added a testcase that triggered the issue (and with the patch does not anymore), so I think this can go in. Can I request someone to review this?

@kvch perhaps?
Thanks

lib/dnscache.c
```diff
@@ -159,12 +159,12 @@ dns_cache_store(DNSCache *self, gboolean persistent, gint family, void *addr, co
   if (!persistent)
     {
       entry->resolved = cached_g_current_time_sec();
-      iv_list_add(&self->cache_list, &entry->list);
+      iv_list_add(&entry->list, &self->cache_list);
     }
```
lbudai: ehh... nice catch

PR #1219, opened 2016-10-03
autogen.sh: temporarily disable shm counters in mongo-c-driver

bkil-syslogng:
This is a temporary quick workaround to help testing, do not merge yet.

* TODO: extract mongo-c-driver kludges to a separate function
* TODO: introduce configure option to reenable shm counters with warnings

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -66,7 +66,7 @@ autogen_submodules()
 
 		CONFIGURE_OPTS="--disable-shared --enable-static --with-pic"
 		# kludge needed by make distcheck in mongo-c-driver
-		CONFIGURE_OPTS="$CONFIGURE_OPTS --enable-man-pages"
+		CONFIGURE_OPTS="$CONFIGURE_OPTS --enable-man-pages --disable-shm-counters"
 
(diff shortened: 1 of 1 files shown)


czanik:
Tested (patched latest git head, created dist file, recompiled rpm package and tested the result), and it solved the problem.

bazsi:
I think this could go in as it is. It fixes a reported security issue... I have filed this bug in mongo-c-driver, let's see how that timeline works out and either update our in-tree copy, or just merge the workaround.

bazsi:
https://jira.mongodb.org/plugins/servlet/mobile#issue/CDRIVER-1691/comment/1405406

They said they could fix this in their upcoming version.

bkil-syslogng:
@bazsi
Thanks for not letting this down. I rebased to master and adjusted the commit message (please review whether it is okay), so it should still pass Travis CI. If it does not, please press the retry button...

I agree that it could go in if it does no great harm (other than breaking mongo-stats). We can commit further improvements later on, hopefully before the next release.

bazsi:
@czanikp pls notify the original reporter.

PR #1225, opened 2016-10-11
F/list functions

bazsi:
This branch introduces list related functionality in syslog-ng, that's it provides a standard way to encode multiple items in a single name-value pair.

It basically encodes the multiple values in a comma separated list, elements can be quoted if they contain a quote character.

It works this way:
   $(list-concat $list1 $list2) -- concatenate two lists
   $(list-append $list value1 value2 value3 $MSG) -- append values to a list (properly escaping them)
   $(list-head) -- first element of the list
   $(list-tail) -- all elements, except the first
   $(list-count) -- number of elements

You could format a list merely just using a template, as "foo,bar" is a valid list as well. If received over from a structured transport, lists can be transported this way.

My future plans are two also make it possible to support list() as a type qualifier in value-pairs expressions, e.g. do something like this:

$(format-json a=list($listvalue))

Which would turn a into a JSON array. Also, the same should be supported on the source side, e.g. generate a list from JSON arrays and the like.

The patch as such are not perfect yet, as I've extracted some of the kv-parser() code to a central location, which has already changed in the meantime, thus the kv-parser() related tests are red.

But I'd like to get this out there, as it's been sitting on my hard-drive for a few months now. It'd be great to receive feedback from @lbudai, @bkil-syslogng, @csabamajor, @fekete-robert and perhaps @faxm0dem 

I originally created this to represent lists of email addresses as can be extracted from SMTP logs from example.

Thanks for the review in advance.


fekete-robert:
Hi, 

this sounds really interesting, just a few quick questions off the top of my had: 

 * how would you populate a list from patterndb? For example, if you parse a section of a message with a parser like ESTRING that contains comma-separated values, then it is automatically treated as a list?
 * if the list contains name-value pairs, then can I use kv-parser on the list?
 * would it be possible to somehow iterate the list, for example, to execute an action for each?

faxm0dem:
This is great news thanks @bazsi.
Here are my two bitcoins:

* `$(list-head)` and `$(list-tail)` would benefit from an additional parameter: the number of elements to retrieve (as a list)
* It would be useful to have `$(list-nth)` or `${mylist}[N]` to extract the value by index, with possibility to have negative index to count backwards
* Even better would be to be able to extract array slices *e.g.* `${foo}[3..6]`
* `$(grep)` could be extended to use lists: give list of elements matching the parameter
* `$(format-json)` should definitely support lists, but that can of course be done later
* I can think of many more useful array functions like `$(map tpl-function MACRO)`, `$(sort)`, `$(reverse)`, `$(min/max/avg)` (for number arrays), etc.

bazsi:
The branch has now been updated addressing some of the feedback above. Here's my detailed answer:

> this sounds really interesting, just a few quick questions off the top of my had: 
> 
>  * how would you populate a list from patterndb? For example, if you parse a section of a message with a parser like ESTRING that contains comma-separated values, then it is automatically treated as a list?

Yes, anything that is comma delimited and is propery escaped can be used as a list. You can ensure that a list item is properly escaped by using the following XML element in patterndb:

<value name="foobar">$(list-append ${listvalue} ${parsedvalue})</value>

List append takes a list (already properly parsed, empty is ok), and append $(parsedvalue) which is taken as a literal element and escaped properly.

>  * if the list contains name-value pairs, then can I use kv-parser on the list?

The list is not a name-value pairs list, it is just a list of elements. You could almost use a csv-parser() on it, but the escaping mode is not the same. However, you can use any of the list functions (e.g. $(list-slice)) to parse into the elements within the list.

We might want a list-parser() in the future, that would work very similarly to how CSV parser works, except that it would use the list syntax, instead of the CSV one and would probably a lot more intuitive.

>  * would it be possible to somehow iterate the list, for example, to execute an action for each?
> 

Well, not right now. What would be the usecase there? But certainly it would be possible.


> 
> * `$(list-head)` and `$(list-tail)` would benefit from an additional parameter: the number of elements to retrieve (as a list)

I have now added a $(list-slice) that makes it possible to cut any slice from the list, including the head and the tail by using Python's slice indexing.

$(list-slice :2 ${list}) -- return the first two elements
$(list-tail -2: ${list}) -- return the last two elements

> * It would be useful to have `$(list-nth)` or `${mylist}[N]` to extract the value by index, with possibility to have negative index to count backwards

$(list-slice 5:5 ${list}) -- is exactly this now

> * Even better would be to be able to extract array slices *e.g.* `${foo}[3..6]`

$(list-slice 3:6 ${list})

> * `$(grep)` could be extended to use lists: give list of elements matching the parameter

I have not changed how $(grep) works, but that does not ensure proper escaping, I've added 
$(context-lookup) though, which has the same syntax as $(grep), but ensures the result is a properly escaped list.

> * `$(format-json)` should definitely support lists, but that can of course be done later

Yup, not in this patchset yet.

> * I can think of many more useful array functions like `$(map tpl-function MACRO)`, `$(sort)`, `$(reverse)`, `$(min/max/avg)` (for number arrays), etc.

These would all be possible, though probably we should use names like:

$(list-map)
$(list-min/max/avg)
$(list-reverse)

All except $(list-map) would be easily doable, but I would wait for the use-cases to pop up. $(list-map) would be a lot more difficult, we are not dealing with a fill LISP/Scheme after all :)

Anyway, I think apart from ugly code duplications here and there (just look at grep, context-lookup and context-values), this should be good enough for merging.

bazsi:
I have now updated this branch to be based on top of PR #1271, so that needs to go in first.

bazsi:
I didn't want to close this. Anyway, I did, so opening a new PR.

PR #1232, opened 2016-10-16
diskq: fix qoverflow_size setting in nonreliable mode

bazsi:
In non-reliable mode, qoverflow_size was set to mem_buf_size instead of
mem_buf_length, which is wrong, as the first is measured in bytes, whereas
the latter is measured in number of elements.


bazsi:
@juhaszviktor this branch fixes an issue where the log-fifo-size() setting for a disk buffer is incorrectly parsed, causing the memory buffer to grow until 2^32-1 elements, possibly causing all memory to be consumed, once the allocated disk buffer is full.

I'd appreciate an approval.

juhaszviktor:
LGTM :+1:

PR #1233, opened 2016-10-16
F/fix internal msg timestamps

bazsi:
This patch should fix the timestamp issue for suppression messages, reported by 
Scot Needy scotrn@gmail.com on the mailing list.

This revamps how timestamps are initialized/parsed, so the code becomes less coupled.


juhaszviktor:
Aside from the magic number in the test (was just a very minor note), it looks good to me :+1:

PR #1235, opened 2016-10-18
groupunset: there is set/unset rewrite rules, but the groupset had no…

juhaszviktor:
… pair.

This patch implements the groupunset rewrite rule.
Previously you can unset a batch of values using the groupset with empty value
But, the empty value handling is introduced, this method does not work anymore.

So we need a rewrite rule which can do this job.

That is what this patch does.

usage example:
``` 
groupunset(values(".SDATA.*"));
```
Signed-off-by: Viktor Juhasz <viktor.juhasz@balabit.com>

### DIFF
--- a/lib/rewrite/rewrite-expr-grammar.ym
+++ b/lib/rewrite/rewrite-expr-grammar.ym
@@ -64,6 +64,7 @@ LogRewrite *last_rewrite;
 %token KW_SET_TAG
 %token KW_CLEAR_TAG
 %token KW_GROUP_SET
+%token KW_GROUP_UNSET
 %token KW_SET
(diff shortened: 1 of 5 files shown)


juhaszviktor:
@bazsi, yes I agree with you.

The groupset does not have '-', and I do not want to break the backward compatible.
My opinion is that we should create 'group-set' alias for groupset and an alias for group-unset.
Or simple rename the groupunset to group-unset. But this could break the symmetry.

bazsi:
Yeah, I would do the aliasing to keep them coherent and deprecate the name
without the dash.

-- 
Bazsi

PR #1243, opened 2016-10-26
java-modules: report location information only when 'loglevel <= DEBUG'

MrAnno:
Requesting location information (class name, method name, line number) is a slow operation, this should be displayed only in DEBUG (and TRACE) mode.

### DIFF
--- a/modules/java-modules/common/src/main/java/org/syslog_ng/logging/SyslogNgInternalLogger.java
+++ b/modules/java-modules/common/src/main/java/org/syslog_ng/logging/SyslogNgInternalLogger.java
@@ -38,7 +38,6 @@ public class SyslogNgInternalLogger extends AppenderSkeleton {
         if (logger.getAppender(SyslogNgInternalLogger.NAME) == null) {
             logger.removeAllAppenders();
             logger.addAppender(new SyslogNgInternalLogger());
-            logger.setLevel(Level.DEBUG);
         }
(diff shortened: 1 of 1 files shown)


modules/java-modules/common/src/main/java/org/syslog_ng/logging/SyslogNgInternalLogger.java
```diff
@@ -58,15 +57,24 @@ public boolean requiresLayout() {
...
+
+        String message = null;
+        String internalMessageBody = event.getMessage().toString();
+
+        if (event.getLevel().isGreaterOrEqual(Level.INFO)) {
+            message = internalMessageBody;
+        }
+        else {
+            StringBuilder formatedMessage = new StringBuilder();
+            formatedMessage.append(event.getLocationInformation().getClassName());
```
kvch: nitpicking: is not it cleaner to create a list for part of the formatted message and then loop over that list and append it to `formattedMessage`?

MrAnno: `LocationInfo` doesn't implement the `Iterable` or `Collection` interface, so I would have to add the message parts one by one to the list. I think this is the same ugly logic as the current one.

lbudai:
Thanks!

PR #1244, opened 2016-10-26
Refactor add-contextual-data

kvch:
This PR contains
 * fixes of prefix to be consistent with existing prefix functionality
 * selector as a plugin to make it easier to add new selectors

### DIFF
--- a/modules/add-contextual-data/add-contextual-data-selector.h
+++ b/modules/add-contextual-data/add-contextual-data-selector.h
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2016 Balabit
+ *
+ * This program is free software; you can redistribute it and/or modify it 
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *  
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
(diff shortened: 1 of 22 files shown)


modules/add-contextual-data/add-contextual-data-selector.h
```diff
@@ -0,0 +1,80 @@
...
+      return self->init(self);
+    }
+
+  return FALSE;
+}
+
+static inline AddContextualDataSelector *
+add_contextual_data_selector_clone(AddContextualDataSelector *self, GlobalConfig *cfg)
+{
+  if (self->clone)
```
MrAnno: Why don't you check `self` here as you did in `resolve`, `free` and `init`?

modules/add-contextual-data/add-contextual-data.c
```diff
@@ -104,35 +109,23 @@ _process(LogParser *s, LogMessage **pmsg,
...
+  if (!context_info_db_contains(self->context_info_db, selector) && _is_default_selector_set(self))
     selector = self->default_selector;
 
   if (selector)
     context_info_db_foreach_record(self->context_info_db, selector,
                                    _add_context_data_to_message,
                                    (gpointer) msg);
 
-  g_string_free(selector_str, TRUE);
+  g_free(resolved_selector);
```
MrAnno: Memleak (see previous comment in `add-contextual-data-template-selector.c`)

modules/add-contextual-data/add-contextual-data-template-selector.c
```diff
@@ -0,0 +1,115 @@
...
+_init(AddContextualDataSelector *s)
+{
+  AddContextualDataTemplateSelector *self = (AddContextualDataTemplateSelector *)s;
+  return _compile_selector_template(self);
+}
+
+static gchar *
+_resolve(AddContextualDataSelector *s, LogMessage *msg)
+{
+  GString *selector_str = g_string_new(NULL);
```
MrAnno: GString should be freed with `g_string_free()` (it has additional fields, not just a gchar buffer).

MrAnno:
👍

modules/add-contextual-data/add-contextual-data.c
MrAnno: Fixed.

PR #1255, opened 2016-11-11
Normalize version related warnings

lbudai:


bazsi:
I like the change in general, but a rationale behind removing the version number from the keyword deprecations would be quite useful. the changelog does not really explain that.

Issue #1248, opened 2016-11-02
plugin module loading error should be only warning message


pzoleex:
On my debian, there are no python neither java and I don't use them in the configuration.
When syslog-ng starts with the default configuration, it gives two error messages, however syslog-ng keeps running.
/# /etc/init.d/syslog-ng start
Starting syslog-ng: [2016-11-02T13:19:38.634466] Error opening plugin module; module='mod-python', error='libpython2.7.so.1.0: cannot open shared object file: No such file or directory'
[2016-11-02T13:19:38.641648] Error opening plugin module; module='mod-java', error='libjvm.so: cannot open shared object file: No such file or directory'
[ ok ] OK.

These message should be only warning (if it's not a blocker error and syslog-ngs keeps running) and should log to the internal log instead of to the standard error.

Config:
```
@version: 3.8
#Default configuration file for syslog-ng.
#
# For a description of syslog-ng configuration file directives, please read
# the syslog-ng Administrator's guide at:
#
# https://www.balabit.com/support/documentation
#
@include "scl.conf"

options {
};

######
# sources
source s_local {
# message generated by Syslog-NG
internal();
system();
};


######
# destinations
destination d_messages { file("/var/log/messages"); };


log {
source(s_local);

destination(d_messages);
};
```

bazsi:
Before doing literally what the issue is, I'd like to point out that the
reason these errors are shown are that:

   - we scan the plugin directory for available plugins at startup
   - we log errors to stderr as internal logging is not yet initialized at
   this point

The solution is to scan for candidate modules in a separate process and/or
generate those independently of the main syslog-ng process, I've discussed
this with @bkil-syslogng at one point, but I can't find a github issue.
Maybe it was in a private email.

-- 
Bazsi

PR #1256, opened 2016-11-14
plugin: debug instead of error when unable to open

lbudai:
Fixes: #1248


bazsi:
looks good to me.

-- 
Bazsi

PR #1257, opened 2016-11-15
so_rcvbuf/sndbuf: These options should not be negative, or causing in…

juhaszviktor:
…teger overflow

With incorrect set values, syslog-ng should stop

Signed-off-by: Viktor Juhasz <viktor.juhasz@balabit.com>

### DIFF
--- a/modules/afsocket/afsocket-grammar.ym
+++ b/modules/afsocket/afsocket-grammar.ym
@@ -712,8 +712,16 @@ tls_option
 
 
 socket_option
-	: KW_SO_SNDBUF '(' LL_NUMBER ')'            { last_sock_options->so_sndbuf = $3; }
-	| KW_SO_RCVBUF '(' LL_NUMBER ')'            { last_sock_options->so_rcvbuf = $3; }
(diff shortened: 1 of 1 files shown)


modules/afsocket/socket-options.c
```diff
@@ -76,6 +76,37 @@ socket_options_setup_socket_method(SocketOptions *self, gint fd, GSockAddr *bind
...
+socket_options_set_sndbuf(SocketOptions *s, gint so_sndbuf)
+{
+  if (so_sndbuf < 0)
+    {
+      msg_error("Invalid value for so_sndbuf",
+                evt_tag_int("so_sndbuf", so_sndbuf),
+                evt_tag_int("minimum value", 0),
+                evt_tag_int("maximum value", G_MAXINT));
+      return FALSE;
+    }
```
bazsi: This should be reported by yyerror() and not by a separate log message.

There's a CHECK_ERROR() macro that makes this trivial from grammars.

Also, a number of tags use spaces instead of underscore (minimum value) above.

I would not merge this in this form.

PR #1271, opened 2016-12-04
F/kv parser refinements

bazsi:
This branch merges the two parallel implementations of KVScanner introduced by an earlier pull request. It also adds pair-separator() as an option and does a number of refactors.

Please review and merge.
@csabamajor


bazsi:
@MrAnno thanks for the fix. So as it seems only the review is needed and this is good to go.

I have split the patches to make that easier, but still its a challenge. let me know if you have questions.

bazsi:
I have folded in the fixup by @MrAnno and also added two minor changes to address review notes that we had IRL.

This should be as good as it gets.

lib/str-repr/encode.c
```diff
@@ -0,0 +1,85 @@
...
+str_repr_encode_append(GString *escaped_string, const gchar *str, gssize str_len, const gchar *forbidden_chars)
+{
+  const gchar *apostrophe, *quote;
+
+  if (str_len < 0)
+    str_len = strlen(str);
+
+  if (str_len == 0)
+    {
+      g_string_append_len(escaped_string, "''", 2);
```
lbudai: Shouldn't we return `"\"\""`?

lib/str-repr/decode.c
```diff
@@ -0,0 +1,212 @@
...
+            {
+              cur = new_cur;
+              goto finish;
+            }
+          else
+            {
+              g_string_append_c(value, *cur);
+            }
+          break;
+        }
```
lbudai: Could you extract case branches into separate functions?

lib/str-repr/tests/test_encode.c
```diff
@@ -0,0 +1,107 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+#include "str-repr/encode.h"
+#include "testutils.h"
+
```
lbudai: please use criterion.

lib/str-repr/encode.c
bazsi: You mean to use quotes instead of apostrophes? Both should be fine.

lib/str-repr/decode.c
bazsi: You mean all branches into separate functions?

lib/str-repr/tests/test_decode.c
```diff
@@ -0,0 +1,165 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+#include "str-repr/decode.h"
+#include "testutils.h"
+
```
bazsi: Well, this test predates criterion, and I did convert one complicated testcase in this branch.

Yeah, some of this code is that old. Could this go in with these tests please? Pretty please?

promise to convert the next tests to criterion I am touching.

lbudai: fine :)

lib/str-repr/decode.c
lbudai: Yep.

lib/str-repr/encode.c
lbudai: `\" is` is used by default somewhere else, and using the same everywhere would be more symmetric

lib/str-repr/decode.c
bazsi: I have just addressed this in a subsequent patch. Note that I've refactored the function in about 6-8 steps, from which I had 6 separate commits, but I've folded them up, I hope it's ok.

Tests didn't change.

lib/str-repr/encode.c
bazsi: I don't see where else you saw this as default. I've looked for it, but no match.

It is true that we prefer double quotes if for some reason something is needed. Hey, I've changed it, easier than another iteration.

bazsi:
Apart from criterion, I've submitted changes to all open issues.

Please merge if those are ok.

lib/str-repr/decode.c
```diff
@@ -0,0 +1,212 @@
...
+      cur++;
+    }
+finish:
+  *end = cur;
+  /* check if quotation was not finished or we had extra characters, return FALSE */
+  if (quote_state == KV_QUOTE_INITIAL ||
+      quote_state == KV_EXPECT_DELIMITER ||
+      quote_state == KV_UNQUOTED_CHARACTERS)
```
bazsi: Oops, i did not address this, however this fragment is now in a much smaller function, but i am not at my computer.

bazsi:
The last patch should extract the complicated if condition into a separate function.

So with this, really all notes have been addressed.

lbudai:
@bazsi : thanks!

PR #1285, opened 2016-12-17
F/decouple configuration

bazsi:
These are relatively straightforward patches to decrease coupling between the GlobalConfig class and various subsystems in syslog-ng.

Instead of relying on GlobalConfig to store global settings for these subsystem, a separate *Options struct is introduced which is then embedded within GlobalConfig. and call sites are changed to pass this embedded struct instead of the entire configuration object.

I do have a series of more introsive patches that make the plugin related code a lot more straightforward, but those patches are a long way before they can be integrated, hence I submit these separately.


bazsi:
I have now rebased this to master, most of this should be trivial.

PR #1288, opened 2016-12-21
affile: guard against double-free errors

adrians:
In certain conditions (many SIGHUP signals), multiple threads
enter the log_proto_file_writer_flush function and generate double-free
errors while deallocating self->buffer[i].iov_base elemets.

A mutex was added (called "mutex_write") that allows only one thread
into the critical region that deallocates the variables, while all
the other threads skip over that region and log a warning message.

Signed-off-by: Adrian Stratulat <adrian.stratulat@enea.com>
Signed-off-by: Dilimot Mihai <mihai.dilimot@enea.com>

### DIFF
--- a/modules/affile/logproto-file-writer.c
+++ b/modules/affile/logproto-file-writer.c
@@ -28,7 +28,6 @@
 #include <errno.h>
 #include <sys/uio.h>
 #include <unistd.h>
-#include <pthread.h>
 
 typedef struct _LogProtoFileWriter
 {
@@ -40,11 +39,9 @@ typedef struct _LogProtoFileWriter
   gint fd;
   gint sum_len;
   gboolean fsync;
+  GMutex mutex_write;
   struct iovec buffer[0];
 } LogProtoFileWriter;
-
-pthread_mutex_t mutex_write;
-unsigned short int write_mutex_init = 1;
 
 /*
  * log_proto_file_writer_flush:
@@ -82,12 +79,6 @@ log_proto_file_writer_flush(LogProtoClient *s)
           g_free(self->partial);
           self->partial = NULL;
         }
-    }
-
-  if(write_mutex_init)
-    {
-      write_mutex_init = 0;
-      pthread_mutex_init(&mutex_write, NULL);
     }
 
   /* we might be called from log_writer_deinit() without having a buffer at all */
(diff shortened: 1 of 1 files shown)


bazsi:
Thanks for submitting these patches. Do you have a description how this can be reproduced?

The fix is not perfect, for these reasons:
 * we should not be racing on _flush() that should already only be called when we are back to one thread (in the case on SIGHUP, worker threads are suspended)
 * the mutex itself is pretty bad for performance at that point
 * even if the mutex was needed, we should be using the Glib mutex primitives
 * the mutex should be in the struct and not a global, and its initialization should not be done in the function itself, as the initialization itself could be racy.

If you could show us a backtrace or how you think two threads may actually be running on that code in the same buffer, it would be lovely.

adrians:
Hello,

About your comments - I agree with your performance concerns, that Glib primitives make more sense in this context and the scope of the mutex can be isolated. In this variant of the patch, a daemon crash was replaced with a small performance penalty - not ideal, but still an acceptable trade-off.
Also, this was the first contact with the syslog-ng codebase. I guess I'll have to send another iteration of the patch.

About reproducing the issue - the issue happened on release 3.5.4.1 built from yocto-project, on PowerPC hardware. I couldn't find a reliable way to reproduce it on my workstation, but fortunately the project's build-bot managed to trigger the race-condition from its first try - you can see in the [log](https://travis-ci.org/balabit/syslog-ng/jobs/185813681) at line 2959 that it displayed a warning message added by this commit, which should happen only if two threads are inside the _flush() function at the same time.

What is the expected way to send a second iteration of the patch in this project? I've seen multiple approaches to this and I want to make sure I'm not breaking the flow: should I rewrite my branch, add another commit on the tip of the branch, or make a new branch and PR from scratch?

Best regards,
Adrian Stratulat

bazsi:
During review, you can just push followup patches to the same branch/pr.

Then we might ask you to squash them at the end.

We like to keep the history clean and bisectable.

adrians:
Hi,

In my last commit I changed the fix such that the mutex is inside the structure, and uses GLIB functions instead of PTHREAD functions. Is there anything else I can do at this point?

Best regards,
Adrian Stratulat

bazsi:
I am still at odds why multiple threads happen to run within that critical section. Can you perhaps drop a core once this happens? There really should be only one thread executing that code, as the "writer" path is running on a dedicated thread.

kira-syslogng:

kira-syslogng:

kira-syslogng:

presidento:
@adrians please, test it with the latest syslog-ng (3.11.1), whether it can be reproduced or not. We improved handling the SIGHUP signals.

adrians:
Ok, as the application has changed, I don't think this Pull Request is useful anymore.
I'm closing it.

Best regards,
Adrian Stratulat

PR #1289, opened 2016-12-21
Add PCRE / OpenSSL to INSTALL dependency list

kkirsche:
This will make it clearer / quicker to install for new users

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -42,6 +42,8 @@ compiled, these are:
     (--enable-tcp-wrapper)
   * Glib a generic I/O library used by a wide variety of applications
   * EventLog a generic event logging library developed by BalaBit
+  * pcre to enable Perl Compatible Regular Expressions
+  * OpenSSL for TLS/SSL communication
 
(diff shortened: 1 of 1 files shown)


bazsi:
Thanks for your contribution.

Albeit this change could be integrated, that file is almost completely out of date. Especially the configuration section.

The libraries required/optionally needed is also a much broader list.

Could you perhaps update the list with all the packages that the configure script looks for (should give you a list with configure --help, at the section that discusses environment variables, or you could look at PKG_CHECK_MODULES() invocations in the configure script.

thanks again

kkirsche:
Of course. I'll take a look at doing that as soon as possible

bazsi:
If you can, look at readme.md and harmonize these files.

Thanks for your contribution.
Cheers
Bazsi

bazsi:
I have now decided to remove the INSTALL file in favour of README.md completely. That contained old stuff
most of which is covered in README.md

With that said, I am closing this PR. Thanks again.

PR #1291, opened 2016-12-23
python: add python parser

lbudai:
* example config:

```    @version: 3.9
    log {
      source {
        tcp(port(5555));
      };
      parser {
        python(
            class("sngexample.RegexParser")
            options("regex", "seq: (?P<seq>\\d+), thread: (?P<thread>\\d+), runid: (?P<runid>\\d+), stamp: (?P<stamp>[^ ]+) (?P<padding>.*$)")
        );
      };
      destination {
        file("/tmp/regexparser.log.txt" template("$MSG;$HOST;$X;$runid"));
      };
    };
```

* example regex parser (etc/sngexample.py)

```
    import re
    class RegexParser:
        def init(self, options):
            pattern = options["regex"]
            self.regex = re.compile(pattern)
            return True
        def parse(self, logmsg):
            logmsg.X = 'custom key/value'
            logmsg['HOST'] = logmsg['X']
            match = self.regex.match(logmsg.MSG)
            if match is not None:
                for key, value in match.groupdict().items():
                    setattr(logmsg, key, value)
                return True
            else:
                return False
```


modules/python/python-logmsg.c
```diff
@@ -49,6 +49,15 @@ py_log_message_getattr(PyLogMessage *self, gchar *name)
...
 
 static void
+py_log_message_setattr(PyLogMessage *self, gchar *name, gchar *value)
+{
+  NVHandle handle;
+
+  handle = log_msg_get_value_handle(name);
+  log_msg_set_value(self->msg, handle, value, -1);
+}
+
```
lbudai: @bazsi: for those identifiers `setattr(logmsg, key, value)` also works (but we can support that ... )

lbudai: @bazsi: I've just added the indexed attrib. access support. 
and also addressed the other notes, could you check the PR?

bazsi:
I had some comments, some related to existing code.

bazsi:
Wrong button, sorry. Reopening

lbudai:
@bazsi: I found an interesting issue while debugging (an exception was thrown somehow in the setattr function, but not in the setitem... then I saw theres is settro/getattro which deprecates the setattr/getattr, and when I set these callbacks, it worked as expected... so derefs issues are fixed)

bazsi:
This looks good to me.

bazsi:
I have rebased now to master while doing so I saw a number of things that need changing. I am pushing my changes on a separate branch in a minute.

bazsi:
ugh, it's a bit more involved than I thought.  Here are the issues I have identified:

1) we should not use "python-parser" as the keyword to initiate the python-parser, rather we need to use "python" in the parser context.

2) we don't need a separate grammar module for this. (python-parser-grammar.ym)

3) we shouldn't create a separate set of keywords either (python-parser-parser.c)

The biggest obstacle in merging the grammar file is that we are using dest driver specific calls to add imports, class names etc. We would need to use a common interface for these kind of things.

bazsi:
I have done a number of changes to bring the parser/destination code a lot closer:

https://github.com/balabit/syslog-ng/tree/python-parser-refinements

We could still add further refinements, but I think this is already good enough for merging. Can you have a look?

Thanks

lbudai:
@bazsi: I reset this branch your branch. Thanks for the changes :)

kira-syslogng:

lbudai:
There were some issues, I hope I fixed all of them (Makefile.am, undefined symbol, etc... ; and configuration also changed, so I updated the example configs, too.)

bazsi:
This patch solves the unit test issue for me:
```
diff --git a/modules/python/python-helpers.c b/modules/python/python-helpers.c
index d9044c8..2581a29 100644
--- a/modules/python/python-helpers.c
+++ b/modules/python/python-helpers.c
@@ -252,8 +252,8 @@ _py_invoke_bool_method_by_name_with_args(PyObject *instance, const gchar *method
     {
       PyObject *args_obj = args ? _py_create_arg_dict(args) : NULL;
       result = _py_invoke_bool_function(method, args_obj, class, module);
-      if (args_obj)
-        PyObject_Del(args_obj);
+
+      Py_XDECREF(args_obj);
       Py_DECREF(method);
     }
   return result;
```

bazsi:
btw an end-to-end testcase would be pretty useful to add to tests/functional/test_python.py

The destination test was catching this error after all.

bazsi:
I've sent a pull request to your branch that contains both the fix and the testcase: d909818cf35d04f7df4d5e7fe26c50917cdd5d9c

lbudai:
@bazsi: thanks!!

bazsi:
Finally :)

PR #1297, opened 2017-01-03
syslog-ng: make '--preprocess-into' foreground only

MrAnno:
The behaviour of the  `--preprocess-into` option should be the same as `--syntax-only`:

This patch forces `--preprocess-into` to run in foreground mode and shuts
down syslog-ng after calling main_loop_init() (after parsing the config).

### DIFF
--- a/lib/control/control.c
+++ b/lib/control/control.c
@@ -115,7 +115,8 @@ static GString *
 control_connection_stop_process(GString *command)
 {
   GString *result = g_string_new("OK Shutdown initiated");
-  main_loop_exit();
+  MainLoop *main_loop = main_loop_get_instance();
+  main_loop_exit(main_loop);
   return result;
 }
 
--- a/lib/control/control.c
+++ b/lib/control/control.c
@@ -123,7 +123,8 @@ static GString *
 control_connection_reload(GString *command)
 {
   GString *result = g_string_new("OK Config reload initiated");
-  main_loop_reload_config();
+  MainLoop *main_loop = main_loop_get_instance();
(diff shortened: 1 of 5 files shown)


MrAnno:
Do not merge this, please.

@bazsi I've pushed 2 additional commits. 
The first one is just a preparation step, I made it only to present my idea of refactoring the main loop module into a singleton class/normal class. I know that the singleton pattern is a pretty bad practice, but I think this can be a correct use-case.

What do you think? Should I make it singleton or leave it in this form?

MrAnno:
Thanks @bazsi.

I introduced "static self" parameters for the `main_loop_*` functions.
This can be a weird paradox for someone familiar with real OOP languages, but I didn't find a better way to express this "static class" idea clearly.

The `main_loop_get_instance` function reflects the main problems with the main loop module (thread-safety, multiple instances of MainLoop). These problems can be solved in the future.

bazsi:
Hi,

Albeit the things you did in addition to the original change was a step in good direction, I've felt that we would have to push things even further, thus I've pushed my followup patches to the "mainloop-instance" branch.

I am submitting a PR against your branch in a moment. 

If you feel that my changes would be too much for integration now, I would split the two PRs into separate ones. One for the original --proprocess-into part and the rest, but I wouldn't integrate your "self_instance" version on its own.

bazsi:
I have merged your initial patch to master, and I think the rest should be addressed as a separate PR.

bazsi:
So closing this PR as it was manually merged by me into 23813c6304e9f3552074d27e3b42763dd8e18c34

Issue #1298, opened 2017-01-03
Enhancement: Support IP_FREEBIND (and friends)


subfxnet:
Please add support to syslog-ng for binding to non-local IP addresses.  This is desired for some situations where syslog-ng may be running on several hosts where a shared IP floats between the hosts for particular load-balancing/HA purposes.

PR #1299, opened 2017-01-03
Incorporate support for IP_FREEBIND/SO_BINDANY/IP_BINDANY

subfxnet:
Allows for binding to non-local ip addresses.

Fixes: https://github.com/balabit/syslog-ng/issues/1298

Signed-off-by: Jason Hensley


subfxnet:
I agree.  I can redo my pull request to only set the socket options in the afsocket module if there is a configuration in place similar to how "tcp-keepalive-time()" is setup.

bazsi:
Yup, that should be the perfect spot.

PR #1301, opened 2017-01-04
tests/unit: fix test_msgparse

lbudai:
Build time_t by using mktime() instead of using magic
numbers(that magic number does not
work anymore as last year was a leap year).


kvch:
@lbudai are you replacing the magic numbers with the new function call? or should I merge this now?

lbudai:
@kvch : thanks for the note, I replaced every occurences

bazsi:
Sorry for making such a rookie mistake. That test is a mess btw.

PR #1304, opened 2017-01-04
Add support for IP_FREEBIND

subfxnet:
Fixes: https://github.com/balabit/syslog-ng/issues/1298

Adds support for binding to non-local addresses on network sources via a new configuration entry:
   source x { network(ip(10.0.0.2) port(1999) ip-freebind(yes)); };

Signed-off-by: Jason Hensley <jhensley@subfx.net>

### DIFF
--- a/modules/afsocket/afsocket-grammar.ym
+++ b/modules/afsocket/afsocket-grammar.ym
@@ -139,6 +139,7 @@ systemd_syslog_grammar_set_source_driver(SystemDSyslogSourceDriver *sd)
 %token KW_IP_TTL
 %token KW_SO_BROADCAST
 %token KW_IP_TOS
+%token KW_IP_FREEBIND
 %token KW_SO_SNDBUF
 %token KW_SO_RCVBUF
 %token KW_SO_KEEPALIVE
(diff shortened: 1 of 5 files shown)


subfxnet:
Lack of time/resources on my end.  I selfishly only require Linux support.  I can make time to add the BSD support back in at a later date if you'd like.

subfxnet:
You may want to check the PR again.  I missed one file in my original commit.  I had to move the setsockopt calls to happen prior to the bind() call in transport-mapper.c to get the benefit of IP_FREEBIND.  I didn't see any obvious reason that these operations couldn't be reordered this way.

bazsi:
So, this is still approved, thanks for the patch.

Can you pls squash changes into a single/few patches, so internal iterations get removed from the history? Thanks.

bazsi:
There was a compile error because of this problem, can you please incorporate this into your branch:

```
diff --git a/modules/afsocket/socket-options-inet.c b/modules/afsocket/socket-options-inet.c
index fb4f3b5..cf7b8a4 100644
--- a/modules/afsocket/socket-options-inet.c
+++ b/modules/afsocket/socket-options-inet.c
@@ -111,7 +111,7 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
       if (self->ip_freebind && (dir & AFSOCKET_DIR_RECV))
         {
 #ifdef IP_FREEBIND
-          setsockopt(sock, SOL_IP, IP_FREEBIND, &on, sizeof(on));
+          setsockopt(fd, SOL_IP, IP_FREEBIND, &on, sizeof(on));
 #else
           msg_error("ip-freebind() is set but no IP_FREEBIND setsockopt on this platform");
           return FALSE;
@@ -149,7 +149,7 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
       if (self->ip_freebind && (dir & AFSOCKET_DIR_RECV))
         {
 #ifdef IP_FREEBIND
-          setsockopt(sock, SOL_IPV6, IP_FREEBIND, &on, sizeof(on));
+          setsockopt(fd, SOL_IPV6, IP_FREEBIND, &on, sizeof(on));
 #else
           msg_error("ip-freebind() is set but no IP_FREEBIND setsockopt on this platform");
           return FALSE;

```

bazsi:
Never mind, I have applied my fixes on top of yours and then merged it to master.

NOTE: the ipv6 version had issues, SOL_IPV6, IP_FREEBIND does not exist, one needs to use SOL_IP, IP_FREEBIND (e.g. the same as the ipv4 version), as the IPV6 implementation reuses the ipv4 bit for this purpose and there's no separate setsockopt. I have manually tested the code and now it properly binds non-local addresses for both ipv4 and ipv6.

bazsi:
This is the merge commit: 626310a5a7136498327bc25a142ba87ca49a50a6

PR #1309, opened 2017-01-05
PythonModule: Handle deletion of _syslogng module

mitzkia:
* We should only delete if the "_syslogng" module
exists
  * Without this patch syslog-ng crashes at start
in the following conditions:
    * OS: Ubuntu 16.04
    * config contains: python{};

Signed-off-by: Andras <andras.mitzki@balabit.com>


bazsi:
I have sent you a pull request against this branch to improve both the fix and the error handling around the
area.

I think it makes sense to push through both at the same time.

PR #1311, opened 2017-01-06
allow http module (curl) to be build with cmake

millert:
The non-java http module does not currently get built with cmake.  I've made a stab at this by adapting some existing cmake build files.  This relies on pkg-config to find libcurl but I don't expect that to be an issue.


cmake/Modules/FindCurl.cmake
```diff
@@ -0,0 +1,34 @@
+#############################################################################
+# Copyright (c) 2016 Balabit
+#
```
bazsi: you probably want to put your own copyright here.

modules/http/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+option (ENABLE_CURL "Enable curl destination" ON)
+
+set(HTTP_PARSER_SOURCES
+    http-plugin.h
```
bazsi: This would probably be HTTP_DESTINATION_SOURCES

modules/http/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+
+if (Curl_FOUND)
+    option(ENABLE_CURL "Enable curl destination" ON)
+else()
+    option(ENABLE_CURL "Enable curl destination" OFF)
+endif()
+
+if (ENABLE_CURL)
+    add_library(curl MODULE ${HTTP_PARSER_SOURCES})
+
```
bazsi: I see that "curl" is being used for the plugin name (I just understood that from your patchset), but I think that's a mistake even in the Makefile, it should be "http". The first incarnation of this plugin was called "curl", then it was renamed to http later on.

Please rename the library to "http".

bazsi: Please also rename it in the Makefile by including this patch in your branch:

```
diff --git a/modules/http/Makefile.am b/modules/http/Makefile.am
index a2d246d..4eca10c 100644
--- a/modules/http/Makefile.am
+++ b/modules/http/Makefile.am
@@ -1,7 +1,7 @@
 if ENABLE_HTTP
 
-module_LTLIBRARIES      += modules/http/libcurl.la
-modules_http_libcurl_la_SOURCES = \
+module_LTLIBRARIES      += modules/http/libhttp.la
+modules_http_libhttp_la_SOURCES = \
   modules/http/http-plugin.h        \
   modules/http/http.c               \
   modules/http/http-grammar.y       \
@@ -14,19 +14,19 @@ BUILT_SOURCES       +=      \
   modules/http/http-grammar.c       \
   modules/http/http-grammar.h
 
-modules_http_libcurl_la_CPPFLAGS  =     \
+modules_http_libhttp_la_CPPFLAGS  =     \
   $(AM_CPPFLAGS)            \
   $(LIBCURL_CFLAGS)          \
   -I$(top_srcdir)/modules/http        \
   -I$(top_builddir)/modules/http
 
-modules_http_libcurl_la_LIBADD  = $(MODULE_DEPS_LIBS) $(LIBCURL_LIBS)
+modules_http_libhttp_la_LIBADD  = $(MODULE_DEPS_LIBS) $(LIBCURL_LIBS)
 
-modules_http_libcurl_la_LDFLAGS = $(MODULE_LDFLAGS)
+modules_http_libhttp_la_LDFLAGS = $(MODULE_LDFLAGS)
 
-modules_http_libcurl_la_DEPENDENCIES= $(MODULE_DEPS_LIBS)
+modules_http_libhttp_la_DEPENDENCIES= $(MODULE_DEPS_LIBS)
 
-modules/http modules/http/ mod-http: modules/http/libcurl.la
+modules/http modules/http/ mod-http: modules/http/libhttp.la
 else
 modules/http modules/http/ mod-http:
 endif

```

bazsi:
never mind the Makefile fixes, I just pushed that portion to master, so only focus on the cmake piece.

Cheers,
Bazsi

bazsi:
Closing in favour of #1332

PR #1312, opened 2017-01-07
Grouping by refinements

bazsi:


bazsi:
This only contains changes/clarifications to a number of log messages, so it should be pretty easy to review.

bazsi:
Albeit this did not receive review, all it does is:

1) clarify a few log messages to use grouping-by() instead of correllate(), the way the plugin was called earlier
2) adding a location tag into a few messages to make it possible to identify grouping-by() instances in the config.

I am taking the responsibility and merging this even without approval, it simply can't be controversial, tests run, travis is happy.

bazsi:
I have merged this from the command line: 5f83742652bd7e1400b65ed37366afb72032a24b

Issue #1310, opened 2017-01-06
(3.8.1) tls cert check segfault (no ca in ca-dir)


grinapo:
## Version of syslog-ng

```
syslog-ng 3.8.1
Installer-Version: 3.8.1
Revision: 3.8.1-9
Module-Directory: /usr/lib/syslog-ng/3.8
Module-Path: /usr/lib/syslog-ng/3.8
Available-Modules: afprog,sdjournal,linux-kmsg-format,json-plugin,basicfuncs,dbparser,syslogformat,add-contextual-data,affile,csvparser,kvformat,date,pseudofile,disk-buffer,afsql,confgen,cryptofuncs,afuser,cef,system-source,afmongodb,afsocket
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: off
```

## Platform

Debian/testing

# Issue

Test machine uses self-signed cert, and it's _not_ installed in /etc/ssl/certs/. syslog-ng segfaults on the first occasion to send over something through the tls network channel:

[pid 143944] 17:05:45.069870 stat("/etc/ssl/certs//f4ea497f.0", 0x7f1ef7ffc430) = -1 ENOENT (No such file or directory)
[pid 143944] 17:05:45.069936 stat("/etc/ssl/certs//f4ea497f.0", 0x7f1ef7ffc430) = -1 ENOENT (No such file or directory)
[pid 143944] 17:05:45.069960 stat("/etc/ssl/certs//f4ea497f.0", 0x7f1ef7ffc430) = -1 ENOENT (No such file or directory)
[pid 143944] 17:05:45.069989 --- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0x1d0} ---

## config
While `peer-verify(optional-untrusted)` there is a ca-dir in the config, which may confuse the logic. It should not die on it, but definitely not with a SIGSEGV. 

## gdb
```
#0  CRYPTO_get_ex_data (ad=0x1d0, idx=idx@entry=0) at crypto/ex_data.c:377
#1  0x00007ffff5b4ef2c in SSL_get_ex_data (s=<optimized out>, idx=idx@entry=0) at ssl/ssl_lib.c:3527
[...]

PR #1316, opened 2017-01-09
Hopefully fix segfault in X509_STORE_CTX_get_app_data(ctx)

grinapo:
see https://bz.apache.org/bugzilla/show_bug.cgi?id=32529
fixes issue #1310

### DIFF
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -54,7 +54,7 @@ tls_get_x509_digest(X509 *x, GString *hash_string)
 int
 tls_session_verify_fingerprint(X509_STORE_CTX *ctx)
 {
-  SSL *ssl = X509_STORE_CTX_get_app_data(ctx);
+  SSL *ssl = (SSL *)X509_STORE_CTX_get_ex_data(ctx, SSL_get_ex_data_X509_STORE_CTX_idx());
   TLSSession *self = SSL_get_app_data(ssl);
   GList *current_fingerprint = self->ctx->trusted_fingerpint_list;
   GString *hash;
@@ -106,7 +106,7 @@ tls_x509_format_dn(X509_NAME *name, GString *dn)
 int
 tls_session_verify_dn(X509_STORE_CTX *ctx)
 {
-  SSL *ssl = X509_STORE_CTX_get_app_data(ctx);
+  SSL *ssl = (SSL *)X509_STORE_CTX_get_ex_data(ctx, SSL_get_ex_data_X509_STORE_CTX_idx());
   TLSSession *self = SSL_get_app_data(ssl);
   gboolean match = FALSE;
   GList *current_dn = self->ctx->trusted_dn_list;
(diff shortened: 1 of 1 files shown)


grinapo:
Please review it carefully: I have conjured this up without dwelving into the depths of the function calls. It worked for the apache guys but I made no real efforts to make it sure it's absolutely equivalent.

bazsi:
Thanks for the patch and the diagnosis.

Hmm, the apache guys say it is related to openssl loading and then unloading.

We might do that during startup to discover modules for demand loading later.

We could confirm this theory by turning demand loading off, something like:

@define autoload-compiled-modules 0
@module afsocket
@module ...

You'd need to repeat this for all modules that your configuration uses.

If this works, then the theory is right.

bazsi:
I have implemented a PoC code for plugin discovery in a separate process that should fix this issue without
your patch. It is in this branch:

https://github.com/balabit/syslog-ng/tree/plugin-discovery-in-separate-process

Can you confirm that it also fixes your problem? (e.g. please revert your patch and use mine). If it does, we would probably still need to apply your version, it seems to be the right way forward, but I also wanted to confirm that this is the case.

Thanks.

grinapo:
I have tried, but I have limited time to try to fiddle with it an within 30 minutes I wasn't able cleanly generate debian packages from the source; copying release-3.8's .../debian into the clone results
```
libtoolize: copying file 'm4/lt~obsolete.m4'
configure.ac:1402: error: possibly undefined macro: AC_VAR_TIMEZONE_EXTERNALS
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/bin/autoconf failed with exit status: 1
dh_autoreconf: sh -c autoreconf --force -i && ./autogen.sh sh -c autoreconf --force -i && ./autogen.sh returned exit code 1
debian/rules:65: recipe for target 'override_dh_autoreconf' failed
make[1]: *** [override_dh_autoreconf] Error 2
```
and I fear I don't have time to investigate. I'll gladly try to build it if you have a method which reliably creates the packages under a stretch build env, or actually generate them and throw 'em at me. :-)

As an emergency you may try to docker me but I pretty much dislike it so it'd need pretty specific instructions how to make it to build. ;-)

lbudai:
@grinapo : AFAIK the `AC_VAR_TIMEZONE_EXTERNALS` m4 macro is provided by `autoconf-archive` package .

bazsi:
I have now investigated the issue a bit deeper.

First, it seems that libssl is directly linked into libsyslog-ng.so, so my theory about libssl being unloaded and reloaded is not true. I can't see why that id gets reassigned though, it should really be zero, unless some clunky thing happens.

Second, I've downloaded the 3.8.1-9 debian package and recompiled it locally on my Ubuntu xenial. The crash did not happen.

With all this details, I am inclined to merge your fix as is. There seems to be clearly a case where libssl unloads and reloads or at least which triggers the verify context ID to be remapped during the runtime of syslog-ng.

The demand-loading stuff I've posted still makes sense, but that can go in a separate pull request.

bazsi:
Please supply a "Signed-off-by" line for the patch, so I can merge it. Posting it to this PR should suffice.

grinapo:
`Signed-off-by: Peter Gervai <grin@grin.hu>`

grinapo:
I hope you mean such post and not a git-magical one. :-)

bazsi:
This is perfeftly fine, thank you.

bazsi:
I have merged this into 43d773135406e3869ee038e511fb1872e280ff91, thanks a lot for both the diagnosis and the patch.

PR #1319, opened 2017-01-14
F/elastic http load balancing

faxm0dem:
This allows to specify multiple cluster URLs so jest will supposedly load-balance across nodes:

```
destination d_elasticsearch {
  elasticsearch2(
    client-lib-dir("/usr/share/elasticsearch/lib/")
    index("syslog-${YEAR}.${MONTH}.${DAY}")
    type("syslog")
    time-zone("UTC")
    client_mode("http")
    cluster_url("http://node01:9200 http://node02:9200")
  );
};
```

If `cluster_url` is omitted, we leverage `server` option to connect to those using url list:

```
destination d_elasticsearch {
  elasticsearch2(
    client-lib-dir("/usr/share/elasticsearch/lib/")
    index("syslog-${YEAR}.${MONTH}.${DAY}")
    type("syslog")
    time-zone("UTC")
    client_mode("http")
    server("node01 node02")
    port(9200)
  );
};
```

### DIFF
--- a/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/ElasticSearchOptions.java
+++ b/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/ElasticSearchOptions.java
@@ -28,6 +28,7 @@ package org.syslog_ng.elasticsearch_v2;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Set;
 
 import org.syslog_ng.LogDestination;
 import org.syslog_ng.options.*;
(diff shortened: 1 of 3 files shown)


faxm0dem:
I have a small doubt about changing the `threaded` boolean. Maybe it would be better to switch it to `false` if only one url is specified?

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/ElasticSearchOptions.java
```diff
@@ -147,17 +148,21 @@ public boolean getSkipClusterHealthCheck() {
...
-		String cluster_url = options.get(CLUSTER_URL).getValue();
-		if (cluster_url.isEmpty()) {
-			StringBuilder url = new StringBuilder();
-			url.append("http://");
-			url.append(getServerList()[0]);
-			url.append(":");
-			url.append(getPort());
-			cluster_url = url.toString();
+	public Set<String> getClusterUrl() {
+		String[] cluster_url = options.get(CLUSTER_URL).getValueAsStringList(" ");
```
bazsi: maybe the name of the function should also be changed, getClusterUrls() would indicate it better that there are multiple values.

Also, is space the best separator in that option? Do we use that elsewhere? Just asking, not trying to suggest something.

faxm0dem:
I agree about the function name, I'll change it.
About the separator, `space` is also used in the option `server` whereas `:` is used for `client-lib-dir`. I Think `space` was more appropriate in this context.

Thanks for your comment

bazsi:
I was looking through the jest implementation, and I _think_ (not sure though), multi-threaded only makes sense if one executes asynchronously, otherwise it does not really matter, as we are waiting for the results anyway.

As I read it, we only use execute() and not executeAsync(), thus multi-threaded mode is not helping, it would probably make things slower/more CPU hungry as we are handing over data through threads.

bazsi:
I have now reviewed your entire patchset, and it seems ok to me. Apart from the name change (that you agreed with) and the fact that I think multiThreaded(false) should be enough for our purposes, this seems pretty good thank you.

This was my first encounter with the Java based code though, so would be nice if someone else with more familiarity could also review this, but I am giving you my +1.

faxm0dem:
I made the name change.
However, I'm not convinced only `executeAsyc` is using the threaded mode.
It seems the object returned by `JestClientFactory` is either a `PoolingHttpClientConnectionManager` in case of `threaded=yes` or a `BasicHttpClientConnectionManager` in the other case.

lbudai:
@faxm0dem : thanks :)
( If I'm correct then it will break your other PR, as this patch is also part of that, so you should rebase the https PR. )

PR #1321, opened 2017-01-17
F/elastic-v2 client_mode(https)

faxm0dem:
WIP: please review

implements elastic-v2 client_mode(https)

this was branched on top of #1319 

# implemented auth methods

* basic-auth
* clientcert
* none

# Config examples

## Basic Auth

```
elasticsearch2(
  …
  client_mode("https")
  http_auth_type("basic")
  http_auth_type_basic_username("syslog_ng")
  http_auth_type_basic_password("gn_golsys")
  ssl_truststore_filepath("/etc/elasticsearch/truststore.jks")
  ssl_truststore_password("changeit")
);
```

## Client Certificate

```
elasticsearch2(
  …
  client_mode("https")
  http_auth_type("clientcert")
  ssl_keystore_filepath("/etc/searchguard/example-pki-scripts/syslog_ng-keystore.jks")
  ssl_keystore_password("changeit")
  ssl_truststore_filepath("/etc/elasticsearch/truststore.jks")
  ssl_truststore_password("changeit")
);
```

## None

```
elasticsearch2(
  …
  client_mode("https")
  http_auth_type("none")
  ssl_truststore_filepath("/etc/elasticsearch/truststore.jks")
  ssl_truststore_password("changeit")
);
```

## Insecure

```
elasticsearch2(
  …
  ssl_insecure("true")
  …
);
```

# todo

* SPNEGO/Kerberos
* other?


faxm0dem:
I'd love some feedback @lbudai @juhaszviktor on this, especially:

* Do you have a better idea on how to pass additional options without creating a gazillion top-level options?
* Any remarks on coding style? (new to java)

Thanks!

faxm0dem:
On second thought, I believe it would be better to keep one level of options, as before.
I therefore propose the following:

* ssl_truststore_filepath
* ssl_truststore_password
* ssl_keystore_filepath
* ssl_keystore_password
* kerberos_config_filepath
* kerberos_principal
* https_auth_type: basic,clientcert,kerberos,...
* https_auth_basic_username
* https_auth_basic_password

lbudai:
@faxm0dem : thanks for the PR, I'm not sure when I can answer your question and review the code.
maybe next week

faxm0dem:
thanks @lbudai I updated the PR with the new option paradigm

scl/elasticsearch/plugin.conf
```diff
@@ -68,6 +68,14 @@ block destination elasticsearch2(
...
   skip_cluster_health_check("")
   cluster_url("")
+  ssl_keystore_filepath("")
+  ssl_keystore_password("")
+  ssl_truststore_filepath("")
+  ssl_truststore_password("")
+  ssl_insecure("false")
+  http_auth_type("none")
+  http_auth_type_basic_username("")
+  http_auth_type_basic_password("")
```
bazsi: I think at the very least the SCL wrapper should use option names that are similar to how other destinations get SSL related parameters. Stuff like ca-file() & ca-dir() to specify the trusted CA certificates (ca-file() as a text file containing many CA certs in PEM format, ca-dir() to specify a directory with hashed filenames that point to CA directories).

Also, keystore() is I guess used for private keys, where we use: key-file() for the private key and cert-file() for the certificate. Is it possible to map arguments this way?

faxm0dem: Good idea, I'll have a look.

I have a question about the file format of key and cert files: should we prefer standard file format like pkcs12 or the Java default trust/keystore format as created by the `keytool` command?

bazsi:
Hi,

I would prefer to accept the pem format that other destinations accept.
Then if it makes sense, I would accept the java specific formats with a
different option name. I dont know how different semantically is a keytool
file vs the pem encoded rsa key that openssl uses.

faxm0dem:
The current implementation already accepts PKCS12 private key without modification.
However, as far as I know the PEM format and JKS are semantically different, and java can't load PEM files without additional code and/or dependancies. Maybe @lbudai can comment on that?

bazsi:
In that case this should be an option with a different name, eg one that
indicates the difference in format.

On Jan 23, 2017 3:39 PM, "Fabien Wernli" <notifications@github.com> wrote:

The current implementation already accepts PKCS12 private key without
modification.
However, as far as I know the PEM format and JKS are semantically
different, and java can't load PEM files without additional code and/or
dependancies. Maybe @lbudai <https://github.com/lbudai> can comment on that?

—
You are receiving this because you commented.
<https://github.com/balabit/syslog-ng/pull/1321#issuecomment-274504458>,
or mute
the thread

faxm0dem:
Not sure I get your meaning.
Currently, it's possible to pass either pkcs12 or jks formats indifferently to `ssl_keystore_filepath` and `ssl_truststore_filepath`.

faxm0dem:
Maybe you meant renaming the options to `java_` instead of `ssl_` ?

lbudai:
@faxm0dem , @bazsi:
I think in we should support the jks files, as fax0modem said before: java users will use the java keystores.
But Bazsi is right when he suggested to use a different name.
We should follow the java terminology(keystore, truststore, similar to the javax.net.ssl options)

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
```diff
@@ -106,7 +219,7 @@ public String getClusterName() {
 		try {
 			JestResult result = client.execute(nodesinfo);
 			clusterName = result.getValue("cluster_name").toString();
-		} catch (IOException e) {
+		} catch (IOException | NullPointerException e) {
```
lbudai: Why do you need to catch `NullPointerException`?

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
```diff
@@ -51,12 +69,107 @@ public ESHttpClient(ElasticSearchOptions options) {
...
+			SSLConnectionSocketFactory sslSocketFactory = null;
+			SchemeIOSessionStrategy httpsIOSessionStrategy = null;
+			HostnameVerifier hostnameVerifier = null;
+			TrustStrategy trustStrategy  = null;
+			KeyStore trustStore = null;
+			/* Initialize trust strategy */
+			// trust any server certificate
+			// TODO: configuration option to control this?
+			trustStrategy = new TrustStrategy() {
+				public boolean isTrusted(X509Certificate[] arg0, String arg1) throws CertificateException {
```
lbudai: Names of the arguments should be `chain` and `authType`.

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
```diff
@@ -51,12 +69,107 @@ public ESHttpClient(ElasticSearchOptions options) {
...
 
 	private JestClient createClient() {
-		String connectionUrl = options.getClusterUrl();
+		Set<String> connectionUrls = options.getClusterUrls();
 		JestClientFactory clientFactory = new JestClientFactory();
-		clientFactory.setHttpClientConfig(new HttpClientConfig
-				.Builder(connectionUrl)
-				.multiThreaded(false)
-				.build());
+
```
lbudai: Could you refactor this huge method? (just split into smaller methods)

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
faxm0dem: sorry, I don't understand your comment.

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
faxm0dem: I think this is a leftover from an experiment, sorry

faxm0dem: That was a dirty trick when the cluster name was not available through the API Call.
For instance, when syslog_ng has no permissions to do so:

```
{"error":{"root_cause":[{"type":"security_exception","reason":"no permissions for cluster:monitor/nodes/info"}],"type":"security_exception","r
eason":"no permissions for cluster:monitor/nodes/info"},"status":403}";
```

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
lbudai: name of the arguments are arg0 and arg1:
replace them by chain and authType.

faxm0dem: ack

kira-syslogng:

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
```diff
@@ -106,9 +217,14 @@ public String getClusterName() {
 		String clusterName;
 		try {
 			JestResult result = client.execute(nodesinfo);
-			clusterName = result.getValue("cluster_name").toString();
+			if (result == null ) {
```
lbudai: you mean `result != null` ?

faxm0dem: No: this is when we cannot read cluster name from REST (e.g. 403). We then use user provided one

lbudai: @faxm0dem : but if result is null then calling `result.getValue()` is not the best idea :)

faxm0dem: ah yes, obviously my mistake - will fix ASAP thanks

lbudai:
@faxm0dem : could you try this branch: https://github.com/lbudai/syslog-ng/tree/elastic-v2-client-mode-https (I refactored the code a little bit)

faxm0dem:
@lbudai: I just tested your refactored code using searchguard2 and it works for me :+1:

lbudai:
@faxm0dem : Thanks! Then I'm going to create a new PR that contains your patches and my refactors.

lbudai:
@faxm0dem : I've just opened #1358.

lbudai:
I'm closing this as #1358 has been already merged.

PR #1322, opened 2017-01-17
logproto: fix possible reading loop in case of multi-line file source

### LINKED ISSUE #1318: logproto-buffered-server: problems with encoding when the message is too long
Hi,

I'm trying to fix multiple multi-line file source related issues.

1. our buffer is full of bytes of multiple messages
The buffer should be split after an unsuccessful [log_proto_text_server_extract()](https://github.com/balabit/syslog-ng/blob/23813c6304e9f3552074d27e3b42763dd8e18c34/lib/logproto/logproto-text-server.c#L400) call.
    The issue was addressed by @presidento, it can be fixed easily without any major changes.
    This can cause an infinite file reading loop, if the following corner-case occurs:
```
LogProtoBufferedServer.buffer:
---------------------------------------------------
| P | P | EOM | C | C | C | C | C | C | C | C | C |
---------------------------------------------------
                ^                                 ^
        pending_buffer_pos                pending_buffer_end

P: bytes from the previously fetched message
EOM: end of MESSAGE (this is not equivalent with end-of-line in multi-line case)
C: bytes from the current message

The buffer size equals to log-msg-size (from the config).
```

Because EOM != EOL in this case and we don't have an EOM for the current message in the buffer, the buffer should be split up by calling `log_proto_text_server_split_buffer()`:
```
-------------------------------------------------
| C | C | C | C | C | C | C | C | C |   |   |   |
-------------------------------------------------
  ^                                 ^
pending_buffer_pos          pending_buffer_end
```
In this state, we can read more data. If the last byte will be the end of the message, we are OK.

------------------------------------

2. our buffer is full of bytes of a single message

```
-------------------------------------------------
| C | C | C | C | C | C | C | C | C | C | C | C |
-------------------------------------------------
  ^                                             ^
pending_buffer_pos                       pending_buffer_end
```
In this case, we can agree that `log_proto_text_server_yield_whole_buffer_as_message()` should be called on the chunk (we can't wait for an EOM without reallocating the buffer). Since the buffer is still `log-msg-size` long here, this is a documented behaviour.

This can work perfectly with encoding disabled.
If we enable encoding (e.g. a dummy UTF-8 to UTF-8 conversion for testing purposes), the process won't be the same:

With encoding enabled, we have another buffer called [raw_buffer](https://github.com/balabit/syslog-ng/blob/23813c6304e9f3552074d27e3b42763dd8e18c34/lib/logproto/logproto-buffered-server.c#L636) which is allocated on the stack.
Its size is always `self->super.options->init_buffer_size` (+ leftover bytes).

Due to the foregoing, syslog-ng reads `self->super.options->init_buffer_size` bytes of data in every run.
After reading into the `raw_buffer`, `log_proto_buffered_server_convert_from_raw()` will be called.
The conversion is not always a byte-to-byte procedure, so it is allowed to [increase the output buffer size](https://github.com/balabit/syslog-ng/blob/23813c6304e9f3552074d27e3b42763dd8e18c34/lib/logproto/logproto-buffered-server.c#L124).

Here are the problems:
- We can't guarantee that `log_proto_text_server_yield_whole_buffer_as_message()` is called on a buffer which represents a `log-msg-size` long input, because the conversion function increases the buffer size after we read a fix amount of data into `raw_buffer` (we always read a fix amount of data, because we don't track the raw buffer's fetching position)
- It can happen that the original buffer (`LogProtoBufferedServer.buffer`) "never" gets full:

```
After a log_proto_text_server_split_buffer() call

LogProtoBufferedServer.buffer:
--------------------------------------------------------------
| C1 | C2 | C3 | C4 | C5 | C6 | C7 | C8 | C9 | C10 | C11 |   |
--------------------------------------------------------------
  ^                                                      ^
pending_buffer_pos                                 pending_buffer_end


Cn: bytes from the current message (no EOM yet)
```

In this state, syslog-ng should only read 1 additional byte to the buffer.
With encoding enabled, this means that we should read 1 additional code point which can be represented in 1 or more bytes in UTF-8. This is not a problem, we can extend our buffer and we also have a leftover buffer for this.

The real problem is that we don't track `pending_buffer_pos` in  raw_buffer, we always read a big chunk of data to raw_buffer. The `log_proto_buffered_server_convert_from_raw()` function "has the right" to increase the size of LogProtoBufferedServer.buffer, so this will be done in every round:

```
read(raw_buffer, sizeof(raw_buffer); // trying to read into the whole raw buffer

raw_buffer:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
| C'12 | C'13 | C'14 | C'15 |
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

log_proto_buffered_server_convert_from_raw(LogProtoBufferedServer.buffer, raw_buffer); // this will increase the buffer size 
    if (state->buffer_size < self->super.options->max_buffer_size)
      state->buffer_size *= 2;

LogProtoBufferedServer.buffer:
----------------------------------------------------------------------------------------------------------------------
| C1 | C2 | C3 | C4 | C5 | C6 | C7 | C8 | C9 | C10 | C11 | C12 | C13 | C14 | C15 |   |   |   |   |   |   |   |   |   |
----------------------------------------------------------------------------------------------------------------------
  ^                                                                              ^
pending_buffer_pos                                                        pending_buffer_end
```
 
As you can see, `LogProtoBufferedServer.buffer` grows, but it can happen that it never gets full but we reach the full size which is 6 * log-msg-size.
**If the message is more than 6 * log-msg-size long, syslog-ng will fail with the [following error message](https://github.com/balabit/syslog-ng/blob/23813c6304e9f3552074d27e3b42763dd8e18c34/lib/logproto/logproto-buffered-server.c#L135)**:

```
Incoming byte stream requires a too large conversion buffer, probably invalid character sequence
```
The file will be reread from the beginning after this error ---> reading loop.

-------------------------------
@bazsi I think this part of the codebase (the encoding part) is common with the PE codebase as well.

In my opinion, `raw_buffer` should be placed on the heap and we should track its position too. Thus syslog-ng won't read a fix amount of data, it will just fill the free space in raw_buffer.
After this modification, the `log_proto_buffered_server_convert_from_raw()` function will work correctly, it will extend the buffer only if the conversion itself requires more space.

The big problem for me is how we can implement the position tracking of raw_buffer. It is in a different encoding than LogProtoBufferedServer.buffer, so we must do small decoding steps continuously.
This is a terribly expensive operation (see: [log_proto_text_server_get_raw_size_of_buffer()](https://github.com/balabit/syslog-ng/blob/23813c6304e9f3552074d27e3b42763dd8e18c34/lib/logproto/logproto-text-server.c#L130)), so I don't really know what I could do here.

What do you think?

I'm going to attach reproduction steps and example logs.

MrAnno:
Using multi-line file sources, EOL is not equivalent with EOM therefore two additional cases should be handled:

- the buffer should be split when log_proto_text_server_extract() can not extract the message from the current buffer.
- if the buffer is full with a single message, the buffer should be yielded as a complete message

Previously, the meaning of the log-msg-size() option was not clearly defined when encoding was set.

In multi-line mode, the buffer size increased without limit to the maximum size due to our conversion mechanism.

This PR also defines the meaning of the log-msg-size() option if conversion is enabled:

If the size of incoming message is greater than log-msg-size() (in the internal UTF-8 representation), it will be splitted.

Fixes #1318

### DIFF
--- a/lib/logproto/logproto-text-server.c
+++ b/lib/logproto/logproto-text-server.c
@@ -377,42 +377,36 @@ log_proto_text_server_fetch_from_buffer(LogProtoBufferedServer *s, const guchar 
                                         const guchar **msg, gsize *msg_len)
 {
   LogProtoTextServer *self = (LogProtoTextServer *) s;
-  const guchar *eol;
   LogProtoBufferedServerState *state = log_proto_buffered_server_get_state(&self->super);
   gboolean result = FALSE;
 
-  eol = log_proto_text_server_locate_next_eol(self, state, buffer_start, buffer_bytes);
+  const guchar *eol = log_proto_text_server_locate_next_eol(self, state, buffer_start, buffer_bytes);
 
-  if (!eol &&
-      ((buffer_bytes == state->buffer_size) ||
-       log_proto_buffered_server_is_input_closed(&self->super)))
+  if (!eol)
     {
-      log_proto_text_server_yield_whole_buffer_as_message(self, state, buffer_start, buffer_bytes, msg, msg_len);
-      goto success;
-    }
-  else if (!eol)
-    {
-      log_proto_text_server_split_buffer(self, state, buffer_start, buffer_bytes);
-      goto exit;
-    }
-  else
-    {
-      if (!log_proto_text_server_extract(self, state, buffer_start, buffer_bytes, eol, msg, msg_len))
+      if (buffer_bytes == state->buffer_size || log_proto_buffered_server_is_input_closed(&self->super))
         {
-          if (buffer_bytes == state->buffer_size)
-            {
-              log_proto_text_server_yield_whole_buffer_as_message(self, state, buffer_start, buffer_bytes, msg, msg_len);
-              goto success;
-            }
-          else
-            {
-              log_proto_text_server_split_buffer(self, state, buffer_start, buffer_bytes);
-              goto exit;
-            }
+          log_proto_text_server_yield_whole_buffer_as_message(self, state, buffer_start, buffer_bytes, msg, msg_len);
+        }
+      else
+        {
+          log_proto_text_server_split_buffer(self, state, buffer_start, buffer_bytes);
+          goto exit;
+        }
+    }
+  else if (!log_proto_text_server_extract(self, state, buffer_start, buffer_bytes, eol, msg, msg_len))
+    {
+      if (buffer_bytes == state->buffer_size)
+        {
+          log_proto_text_server_yield_whole_buffer_as_message(self, state, buffer_start, buffer_bytes, msg, msg_len);
+        }
+      else
+        {
+          log_proto_text_server_split_buffer(self, state, buffer_start, buffer_bytes);
+          goto exit;
         }
     }
 
-success:
   log_proto_text_server_remove_trailing_newline(msg, msg_len);
   result = TRUE;
 
@@ -420,7 +414,6 @@ exit:
   log_proto_buffered_server_put_state(&self->super);
   return result;
 }
-
 
 static void
 log_proto_text_server_free(LogProtoServer *s)
--- a/lib/logproto/logproto-text-server.c
+++ b/lib/logproto/logproto-text-server.c
@@ -398,7 +398,18 @@ log_proto_text_server_fetch_from_buffer(LogProtoBufferedServer *s, const guchar 
   else
     {
       if (!log_proto_text_server_extract(self, state, buffer_start, buffer_bytes, eol, msg, msg_len))
-        goto exit;
+        {
+          if (buffer_bytes == state->buffer_size)
+            {
+              log_proto_text_server_yield_whole_buffer_as_message(self, state, buffer_start, buffer_bytes, msg, msg_len);
+              goto success;
+            }
+          else
+            {
+              log_proto_text_server_split_buffer(self, state, buffer_start, buffer_bytes);
+              goto exit;
+            }
+        }
     }
 
 success:
--- a/lib/logproto/logproto-text-server.c
+++ b/lib/logproto/logproto-text-server.c
@@ -117,6 +117,17 @@ log_proto_text_server_prepare(LogProtoServer *s, GIOCondition *cond)
   return avail;
 }
 
+static void
+log_proto_test_server_maybe_realloc_reverse_buffer(LogProtoTextServer *self, gsize buffer_length)
+{
+  if (self->reverse_buffer_len >= buffer_length)
+    return;
+
+  /* we free and malloc, since we never need the data still in reverse buffer */
+  g_free(self->reverse_buffer);
+  self->reverse_buffer_len = buffer_length;
+  self->reverse_buffer = g_malloc(buffer_length);
+}
 
 /*
  * returns the number of bytes that represent the UTF8 encoding buffer
@@ -153,13 +164,9 @@ log_proto_text_server_get_raw_size_of_buffer(LogProtoTextServer *self, const guc
   if (self->convert_scale)
     return g_utf8_strlen((gchar *) buffer, buffer_len) * self->convert_scale;
 
-  if (self->reverse_buffer_len < buffer_len * 6)
-    {
-      /* we free and malloc, since we never need the data still in reverse buffer */
-      g_free(self->reverse_buffer);
-      self->reverse_buffer_len = buffer_len * 6;
-      self->reverse_buffer = g_malloc(buffer_len * 6);
-    }
+
+  /* Multiplied by 6, because 1 character can be maximum 6 bytes in UTF-8 encoding */
+  log_proto_test_server_maybe_realloc_reverse_buffer(self, buffer_len * 6);
 
   avail_out = self->reverse_buffer_len;
   out = self->reverse_buffer;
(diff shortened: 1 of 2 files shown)


lib/logproto/logproto-text-server.c
```diff
@@ -383,36 +398,12 @@ log_proto_text_server_fetch_from_buffer(LogProtoBufferedServer *s, const guchar
 
   eol = log_proto_text_server_locate_next_eol(self, state, buffer_start, buffer_bytes);
 
-  if (!eol &&
-      ((buffer_bytes == state->buffer_size) ||
-       log_proto_buffered_server_is_input_closed(&self->super)))
+  if (!eol || !log_proto_text_server_extract(self, state, buffer_start, buffer_bytes, eol, msg, msg_len))
```
presidento: I suggest using the following condition (single if, and without need of goto):

```C
  eol = log_proto_text_server_locate_next_eol(self, state, buffer_start, buffer_bytes);

  if (eol && log_proto_text_server_extract(self, state, buffer_start, buffer_bytes, eol, msg, msg_len)
      || log_proto_text_server_handle_extract_failures(self, state, buffer_start, buffer_bytes, msg, msg_len))
    {
      log_proto_text_server_remove_trailing_newline(msg, msg_len);
      result = TRUE;
    }

  log_proto_buffered_server_put_state(&self->super);
  return result;
```

lib/logproto/logproto-text-server.c
```diff
@@ -383,36 +398,12 @@ log_proto_text_server_fetch_from_buffer(LogProtoBufferedServer *s, const guchar
...
-              log_proto_text_server_yield_whole_buffer_as_message(self, state, buffer_start, buffer_bytes, msg, msg_len);
-              goto success;
-            }
-          else
-            {
-              log_proto_text_server_split_buffer(self, state, buffer_start, buffer_bytes);
-              goto exit;
-            }
-        }
+      if (!log_proto_text_server_handle_extract_failures(self, state, buffer_start, buffer_bytes, msg, msg_len))
```
MrAnno: @presidento noted that this is not a fully equivalent modification. I'll split this commit.

lib/logproto/logproto-text-server.c
MrAnno: We discussed this IRL, it is not relevant anymore.

lib/logproto/logproto-text-server.c
MrAnno: Fixed.

lib/logproto/logproto-text-server.c
```diff
@@ -153,12 +155,12 @@ log_proto_text_server_get_raw_size_of_buffer(LogProtoTextServer *self, const guc
   if (self->convert_scale)
     return g_utf8_strlen((gchar *) buffer, buffer_len) * self->convert_scale;
 
-  if (self->reverse_buffer_len < buffer_len * 6)
+  if (self->reverse_buffer_len < requried_reverse_buffer_length)
     {
```
bazsi: typo in the name of the variable, but also, I'd probably extract the reallocation routine instead of extracting this to a separate
value.

For instance if you had log_proto_text_server_maybe_realloc_reverse_buffer(buffer_len * 6) with a comment about the "6" or that 6 extracted into a separate value, this whole thing would be even more easy to read.

lib/logproto/logproto-text-server.c
```diff
@@ -377,31 +379,37 @@ log_proto_text_server_fetch_from_buffer(LogProtoBufferedServer *s, const guchar
...
   LogProtoBufferedServerState *state = log_proto_buffered_server_get_state(&self->super);
   gboolean result = FALSE;
 
-  eol = log_proto_text_server_locate_next_eol(self, state, buffer_start, buffer_bytes);
+  const guchar *eol = log_proto_text_server_locate_next_eol(self, state, buffer_start, buffer_bytes);
 
-  if (!eol &&
-      ((buffer_bytes == state->buffer_size) ||
-       log_proto_buffered_server_is_input_closed(&self->super)))
+  gboolean message_reached_max_size = buffer_bytes >= self->super.super.options->max_msg_size;
```
bazsi: I think I would've extracted the condition into a separate function, possibly the prefix chopped off, to make reading the code easier:

```
if (!eol && (_message_size_too_large() || log_proto_buffered_server_is_input_closed(&self->super))
   ;
```
Wouldn't that be better?

MrAnno: Fixed.

lib/logproto/logproto-text-server.c
MrAnno: Fixed.

lib/logproto/logproto-text-server.c
```diff
@@ -117,6 +117,17 @@ log_proto_text_server_prepare(LogProtoServer *s, GIOCondition *cond)
   return avail;
 }
 
+static void
+log_proto_test_server_maybe_realloc_reverse_buffer(LogProtoTextServer *self, gsize buffer_length)
```
MrAnno: Typo again: log_proto_test_server_maybe_realloc_reverse_buffer

bazsi:
I have fixed the typo and merged this to master.

MrAnno:
Thanks, @bazsi.

PR #1326, opened 2017-01-20
syslog-ng as a command line tool

kvch:
This PR contains:
 * configuration file opening and reading refactor
 * SCL wrapper of `/dev/stdin` and `/dev/stdout`
 * stopping `syslog-ng` on EOF from `/dev/stdin`
 * ability to configure drivers via command line

Possible usage:
```
$ echo logs | syslog-ng --cli 'filter facility(4);'
```

### DIFF
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -426,39 +426,45 @@ cfg_load_config(GlobalConfig *self, gchar *config_string, gboolean syntax_only, 
 }
 
 gboolean
-cfg_read_config(GlobalConfig *self, const gchar *fname, gboolean syntax_only, gchar *preprocess_into)
+cfg_open_config(GlobalConfig *self, const gchar *fname)
 {
   FILE *cfg_file;
-  gint res;
 
   self->filename = fname;
-
   if ((cfg_file = fopen(fname, "r")) != NULL)
     {
-      CfgLexer *lexer;
-      GString *preprocess_output = g_string_sized_new(8192);
-
-      lexer = cfg_lexer_new(cfg_file, fname, preprocess_output);
-      res = cfg_run_parser(self, lexer, &main_parser, (gpointer *) &self, NULL);
-      fclose(cfg_file);
-      if (preprocess_into)
-        {
-          cfg_dump_processed_config(preprocess_output, preprocess_into);
-        }
-      g_string_free(preprocess_output, TRUE);
-      if (res)
-        {
-          /* successfully parsed */
-          return TRUE;
-        }
+      self->cfg_file = cfg_file;
     }
   else
     {
       msg_error("Error opening configuration file",
                 evt_tag_str(EVT_TAG_FILENAME, fname),
                 evt_tag_errno(EVT_TAG_OSERROR, errno));
+      return FALSE;
     }
+  return TRUE;
+}
 
+gboolean
+cfg_read_config(GlobalConfig *self, gboolean syntax_only, gchar *preprocess_into)
+{
+  gint res;
+  CfgLexer *lexer;
+  GString *preprocess_output = g_string_sized_new(8192);
+
+  lexer = cfg_lexer_new(self->cfg_file, self->filename, preprocess_output);
+  res = cfg_run_parser(self, lexer, &main_parser, (gpointer *) &self, NULL);
+  fclose(self->cfg_file);
+  if (preprocess_into)
+    {
+      cfg_dump_processed_config(preprocess_output, preprocess_into);
+    }
+  g_string_free(preprocess_output, TRUE);
+  if (res)
+    {
+      /* successfully parsed */
+      return TRUE;
+    }
   return FALSE;
 }
 
--- a/lib/cli.c
+++ b/lib/cli.c
@@ -1,0 +1,222 @@
+/*
+ * Copyright (c) 2016-2017 Noémi Ványi
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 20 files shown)


bazsi:
This seems promising :) ill try to review this over the weekend.

kira-syslogng:

kvch:
I modified the generation of the config file, so now inline config is used. I also added `--destination` option so the output logs can be templated/configured.

Usage
```
cat logs | syslog-ng --cli --destination 'template("$SOURCEIP $MSG\n")'
```

lib/cfg.c
```diff
@@ -426,39 +426,45 @@ cfg_load_config(GlobalConfig *self, gchar *config_string, gboolean syntax_only,
...
+    {
+      cfg_dump_processed_config(preprocess_output, preprocess_into);
+    }
+  g_string_free(preprocess_output, TRUE);
+  if (res)
+    {
+      /* successfully parsed */
+      return TRUE;
+    }
   return FALSE;
```
lbudai: `return res == 1? TRUE : FALSE` ? :)

lib/cli.c
```diff
@@ -0,0 +1,199 @@
...
+  CliParam *self = g_new0(CliParam, 1);
+  self->type = type;
+  self->cfg = cfg;
+  return self;
+}
+
+static gchar *
+_get_config_of_driver(CliParam *driver)
+{
+  return g_strdup_printf(DRIVER_TEMPLATE, driver->type, driver->cfg);
```
lbudai: I wouldn't use these template variable (I think the code is readable when you have the format string next to the variables).

lib/cli.c
```diff
@@ -0,0 +1,199 @@
...
+}
+
+static gchar *
+_concatenate_strings(GList *elements)
+{
+  GList *element;
+  GString *concatenated = g_string_new("");
+
+  for (element = elements; element != NULL; element = element->next)
```
lbudai: You can use the c99 declaration rules.

lib/cli.c
```diff
@@ -0,0 +1,199 @@
...
+}
+
+gboolean
+cli_param_converter_setup(CliParamConverter *self)
+{
+  guint i;
+
+  for (i = 0; self->raw_params[i]; i++)
+    {
+      gchar *raw_param = g_strdup(self->raw_params[i]);
```
lbudai: where do you free `raw_param`?

lib/mainloop.c
```diff
@@ -453,17 +455,55 @@ main_loop_init(MainLoop *self, MainLoopOptions *options)
...
+}
+
+static int
+_prepare_configuration_for_reading(MainLoop *self)
+{
+  if (_is_cli_param_conversion_required(self->options))
+    {
+      return _setup_cli_mode(self->options, self->current_configuration);
+    }
+  else if (!cfg_open_config(self->current_configuration, resolvedConfigurablePaths.cfgfilename))
```
lbudai: Why do you need  `else` ?

lib/cli.c
```diff
@@ -0,0 +1,199 @@
...
+    }
+  return TRUE;
+}
+
+CliParamConverter *
+cli_param_converter_new(gchar **params, gchar *destination_params)
+{
+  CliParamConverter *self = g_new0(CliParamConverter, 1);
+  self->raw_params = params;
+  if (destination_params)
```
lbudai: I'd use block statements here.

modules/affile/affile-common.c
```diff
@@ -112,6 +112,14 @@ _open_fd(const gchar *name, FileOpenOptions *open_opts, FilePermOptions *perm_op
   return fd;
 }
 
+gboolean
+affile_is_linux_dev_stdin(const gchar *filename)
```
lbudai: `/dev/stdin` is not linux specific

bazsi:
What options does the destination have? Do we need the explicit template() term?

Also the way exit is triggered is a bit fishy for me. I would like to look at that in more detail.

kvch:
I thought template option might not be enough for configuring the destination. So it is just simply passes the file destination options to the driver. @bazsi What do you think is template enough?

Why do you think it is fishy? Isn't it what NC_* notification for? :)

bazsi:
The point is that we got to simplify to a point so that stuff becomes
really easy to use, having to remember template () makes that more
difficult.

While thinking about this, maybe I would concentrate on the subset of this
branch only in its ability to read until EOF but use stock syslog-ng syntax
or require a config file.

Then I would play with the grammar a bit, simplifying it a even more

I am thinking like:

* Input is always flags(no-parse)
* Output is a template, and is name=value by default, excluding syslog
fields and matches by default
* Processing steps are parsers mostly, but filters and rewrite ops do make
sense, but we should lean towards parsers in the syntax

Some examples:
```
Normalize key value pairs by reproducing the input after parsing:

$ syslog-ng --pipe kv-parser()

Grab a single field:
$ syslog-ng --pipe 'kv-parser() | template $foo '

Convert csv to key values:
$ syslog-ng --pipe 'csv-parser(columns (foo,bar,baz))'

Filter processed data:
$ syslog-ng --pipe 'apache-accesslog-parser() | filter
${.apache.httpresponse} = 200'

Maybe even add an different output formatters:
$ syslog-ng --pipe 'apache-accesslog-parser() | filter
${.apache.httpresponse} = 200 | table'
```

The same syntax could be used in search queries in ssb to transform search
results.

What do you think?


On Feb 17, 2017 01:28, "Noémi Ványi" <notifications@github.com> wrote:

I thought template option might not be enough for configuring the
destination. So it is just simply passes the file destination options to
the driver. @bazsi <https://github.com/bazsi> What do you think is template
enough?

Why do you think it is fishy? Isn't it what NC_* notification for? :)

—

<https://github.com/balabit/syslog-ng/pull/1326#issuecomment-280600658>,
or mute
the thread

lbudai:
@bazsi: I'd suggest to create a proposal that describes your idea, and we could also discuss it in IRL together with @kvch (OSE handover on tomorrow) . (I like the syntax you suggested above).

bazsi:
Hi,

Still, the infrastructure pieces are valuable in this branch are quite
valuable, and I would concentrate on those first.

kira-syslogng:

presidento:
We are closing it in favor of #1605.

PR #1282, opened 2016-12-15
F/list functions

bazsi:
This branch is based on f/kv-parser-refinements and is the same as the erroneously closed PR #1225 

The primary goal of this branch is to add $(list-XXX) template functions to manipulate lists.

Once #1271 goes in, this can go in second.

PR #1327, opened 2017-01-22
F/list functions

bazsi:
This is a continuation of #1282, which was opened against a non-master branch. It is over a couple of iterations, comments and tests, please see #1282 for more information.

### DIFF
--- a/lib/scanner/list-scanner/list-scanner.c
+++ b/lib/scanner/list-scanner/list-scanner.c
@@ -1,0 +1,202 @@
+/*
+ * Copyright (c) 2016 Balabit
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+#include "scanner/list-scanner/list-scanner.h"
+#include "str-repr/encode.h"
+#include "str-repr/decode.h"
+#include "str-utils.h"
+
+#include <string.h>
+#include <stdarg.h>
+
+static void
+_free_input(ListScanner *self)
+{
+  if (self->argv)
+    {
+      if (self->free_argv_payload)
+        g_ptr_array_foreach(self->argv_buffer, (GFunc) g_free, NULL);
+    }
+  g_ptr_array_set_size(self->argv_buffer, 0);
+}
+
+static void
+_store_input(ListScanner *self, gint argc, gchar **argv, gboolean free_argv)
+{
+  self->argc = argc;
+  self->argv = (gchar **) self->argv_buffer->pdata;
+  self->free_argv_payload = free_argv;
+  self->current_arg_ndx = 0;
+  self->current_arg = self->argv[self->current_arg_ndx];
+}
+
+void
+list_scanner_input_va(ListScanner *self, const gchar *arg1, ...)
+{
+  va_list va;
+  gint i;
+  const char *arg;
+
+  _free_input(self);
+
+  va_start(va, arg1);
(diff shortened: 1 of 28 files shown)


bazsi:
Does it help integration if I split this branch to smaller chunks? Most of these are simply new template functions that should not break existing functionality. The remaining pieces offer an improvement in kv-parser(), but that's less important than the initial couple of patches.

Thanks

lib/scanner/list-scanner/list-scanner.c
```diff
@@ -0,0 +1,202 @@
...
+gboolean
+list_scanner_scan_next(ListScanner *self)
+{
+  g_string_truncate(self->value, 0);
+
+  while (_locate_next_non_empty_arg(self))
+    {
+      if (_parse_unquoted_empty_value(self))
+        {
+          ;
```
lbudai: I think adding this empty statement is superfluous, the empty block statement is enough.
(not a blocking issue, just a note)

lib/scanner/list-scanner/tests/test_list_scanner.c
```diff
@@ -0,0 +1,290 @@
...
+  LIST_SCANNER_TESTCASE(test_list_scanner_unquoted_empty_items_are_skipped_to_make_it_easy_to_concatenate_lists);
+  LIST_SCANNER_TESTCASE(test_list_scanner_quoted_empty_items_are_parsed_as_empty_values);
+  LIST_SCANNER_TESTCASE(test_list_scanner_comma_delimiter_values_are_split);
+  LIST_SCANNER_TESTCASE(test_list_scanner_comma_and_arg_are_equivalent);
+  LIST_SCANNER_TESTCASE(test_list_scanner_works_with_gstring_input);
+  LIST_SCANNER_TESTCASE(test_list_scanner_handles_single_quotes);
+  LIST_SCANNER_TESTCASE(test_list_scanner_handles_double_quotes);
+  LIST_SCANNER_TESTCASE(test_list_scanner_malformed_quotes);
+  return 0;
+}
```
lbudai: For new tests we prefer to use the criterion framework.

modules/basicfuncs/list-funcs.c
```diff
@@ -0,0 +1,254 @@
...
+_adjust_list_slice(gint argc, GString *argv[],
+                   gint *first_ndx, gint *last_ndx)
+{
+  gint count = -1;
+
+  if (*first_ndx < 0 || *last_ndx < 0)
+    count = _list_count(argc, argv);
+
+  if (*first_ndx < 0)
+    *first_ndx += count;
```
lbudai: `first_ndx` can still be a negative number as I see

lbudai: Or `parse_number` accepts only non-negative values and when the user sets a value that cannot be parse as a number then it sets the output parameter value to -1? (I didn't check it)

modules/basicfuncs/context-funcs.c
```diff
@@ -43,3 +43,47 @@ tf_context_length_free_state(gpointer s)
...
+void
+tf_context_lookup_call(LogTemplateFunction *self, gpointer s, const LogTemplateInvokeArgs *args, GString *result)
+{
+  gint i, msg_ndx;
+  gboolean first = TRUE;
+  TFCondState *state = (TFCondState *) s;
+  gint count = 0;
+  GString *buf = g_string_sized_new(64);
+
+  for (msg_ndx = 0; msg_ndx < args->num_messages; msg_ndx++)
```
lbudai: I'd limit the scope of `msg_ndx` to the for cycle. (just a note)

kira-syslogng:

pzoleex:
retest this please

kira-syslogng:
Build SUCCESS

modules/basicfuncs/list-funcs.c
bazsi: parse_number() can parse negative numbers, it does parsing with strtoll(). 

The negative values are actually supported values in slice boundaries and are interpreted similarly to Python does that.

```
$(list-slice -1: $LIST) - is the last element (negative elements index from the end of the list)
$(list-slice :-1 $LIST) - is until the last element
```

_adjust only adds "count", which might still be negative, but as I see those cases are handled in the caller. as the loop to skip initial entries would not be entered, thus we would be using entries from the very first element.

```
$(list-slice -100: a,b,c) - returns "a,b,c"
$(list-slice -1: a,b,c) - returns "c"
```
Similarly for the end index. If it is too low, the 'slicing' loop will not be entered (the 2nd loop in _list_slice), causing no elements of the list being returned.

Anyway, adding a comment to make this clear.

lib/scanner/list-scanner/list-scanner.c
bazsi: I have refactored this code to make this more readable.

bazsi:
I have fixed the notes with a couple of followup patches, please let me know if you are satisfied.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/basicfuncs/list-funcs.c
lbudai: thanks for the info

bazsi:
\o/

PR #1331, opened 2017-01-27
Stop using deprecated api openssl 1.1

lbudai:


lbudai:
@eroen : could you check this?

bazsi:
I like this, so you have my :+1:

bazsi:
I am merging this, @eroen fell free to comment even if it's on master.

PR #1332, opened 2017-01-29
http: allow http module (curl) to be build with cmake

bazsi:
This PR fixes the outstanding issues originally in PR #1311, Please @millert have a look and supply your signed-off-by so that we can merge this.

PS: I am closing #1311 in favour of this one.

Thanks.


bazsi:
@millert can you confirm it is ok that I added your copyright and that you can sign off the patch?

I think this could be merged without those, as it's qestionable whether it's copyrightable at all, but the easiest would be if you could just say so. It is enough if you post a comment here with a signed-off-by line.

Thanks.

PR #1335, opened 2017-02-03
new syslog-ng-ctl command: query

kvch:
Previously, the only possiblity to get statistics of syslog-ng was using `stats` command of `syslog-ng-ctl`. It is limited as it return all of the counters in CSV format. From now on, it is possible to filter, so only desired metrics are shown using the new `query` command. The new format of each counter and its value is `<counter-id>: <value>`.

## Commands
`list` - List all counters with matching name
`get` - Get the value of selected counter
`get --sum` - Get the aggregated value of selected counters

### List all counters
```
$ syslog-ng-ctl query list
center.received.processed
src.internal.s_local#0.processed
src.internal.s_local#0.stamp
center.queued.processed
global.payload_reallocs.processed
global.sdata_updates.processed
destination.d_local.processed
global.msg_clones.processed
source.s_local.processed
```

### List selected counters
```
$ syslog-ng-ctl query list "global.*.*"
global.payload_reallocs.processed
global.sdata_updates.processed
global.msg_clones.processed
```

```
$ syslog-ng-ctl query list "glob?l*"
global.payload_reallocs.processed
global.sdata_updates.processed
global.msg_clones.processed
```

### Retrieve values of selected counters
```
$ syslog-ng-ctl query get "global.*.*"
global.payload_reallocs.processed: 0
global.sdata_updates.processed: 0
global.msg_clones.processed: 0
```

### Retrieve aggregated values of selected counters
```
$ syslog-ng-ctl query get --sum "global.*.*"
0
```

### DIFF
--- a/lib/stats/stats-cluster.c
+++ b/lib/stats/stats-cluster.c
@@ -164,6 +164,12 @@ stats_cluster_untrack_counter(StatsCluster *self, gint type, StatsCounterItem **
   *counter = NULL;
 }
 
+gboolean
+stats_cluster_is_alive(StatsCluster *self, gint type)
+{
+  return ((1<<type) & self->live_mask);
+}
+
 StatsCluster *
 stats_cluster_new(gint component, const gchar *id, const gchar *instance)
 {
--- a/lib/stats/stats-cluster.c
+++ b/lib/stats/stats-cluster.c
@@ -39,18 +39,18 @@ stats_cluster_foreach_counter(StatsCluster *self, StatsForeachCounterFunc func, 
     }
 }
 
+static const gchar *tag_names[SC_TYPE_MAX] =
+{
+  /* [SC_TYPE_DROPPED]   = */ "dropped",
+  /* [SC_TYPE_PROCESSED] = */ "processed",
+  /* [SC_TYPE_STORED]   = */  "stored",
+  /* [SC_TYPE_SUPPRESSED] = */ "suppressed",
+  /* [SC_TYPE_STAMP] = */ "stamp",
+};
+
 const gchar *
 stats_cluster_get_type_name(gint type)
 {
-  static const gchar *tag_names[SC_TYPE_MAX] =
-  {
-    /* [SC_TYPE_DROPPED]   = */ "dropped",
-    /* [SC_TYPE_PROCESSED] = */ "processed",
-    /* [SC_TYPE_STORED]   = */  "stored",
-    /* [SC_TYPE_SUPPRESSED] = */ "suppressed",
-    /* [SC_TYPE_STAMP] = */ "stamp",
-  };
-
   return tag_names[type];
 }
 
@@ -164,6 +164,27 @@ stats_cluster_untrack_counter(StatsCluster *self, gint type, StatsCounterItem **
   *counter = NULL;
 }
 
+static gchar *
+_stats_build_query_key(StatsCluster *self)
+{
+  GString *key = g_string_new("");
+  gchar buffer[64] = {0};
+  const gchar *component_name = stats_cluster_get_component_name(self, buffer, sizeof(buffer));
+
+  g_string_append(key, component_name);
+
+  if (self->id && self->id[0])
+    {
+      g_string_append_printf(key, ".%s", self->id);
+    }
+  if (self->instance && self->instance[0])
+    {
+      g_string_append_printf(key, ".%s", self->instance);
+    }
+
+  return g_string_free(key, FALSE);
+}
+
 gboolean
 stats_cluster_is_alive(StatsCluster *self, gint type)
 {
@@ -179,6 +200,7 @@ stats_cluster_new(gint component, const gchar *id, const gchar *instance)
   self->id = g_strdup(id ? : "");
   self->instance = g_strdup(instance ? : "");
   self->use_count = 0;
+  self->query_key = _stats_build_query_key(self);
   return self;
 }
 
@@ -187,5 +209,6 @@ stats_cluster_free(StatsCluster *self)
 {
   g_free(self->id);
   g_free(self->instance);
+  g_free(self->query_key);
   g_free(self);
 }
--- a/lib/stats/stats-query.c
+++ b/lib/stats/stats-query.c
@@ -1,0 +1,272 @@
+/*
+ * Copyright (c) 2017 Balabit
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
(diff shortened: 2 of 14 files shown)


lib/str-utils.c
```diff
@@ -60,3 +60,29 @@ __normalize_key(const gchar *buffer)
...
+{
+  size_t len = 0;
+  while (arr[len] != NULL)
+    len++;
+  return len;
+}
+
+gchar *
+str_array_join(const gchar *separator, gchar *argv[])
+{
```
bazsi: why not use g_strjoinv() here?

lib/CMakeLists.txt
```diff
@@ -187,6 +188,7 @@ set(LIB_SOURCES
     scratch-buffers.c
     serialize.c
     service-management.c
+    stats-query-commands.c
     str-format.c
```
bazsi: I think this should go in lib/stats, no?

lib/stats-query-commands.c
```diff
@@ -0,0 +1,107 @@
...
+
+#include "stats-query-commands.h"
+#include "stats/stats-query.h"
+#include "messages.h"
+
+enum QueryCommands
+{
+  QUERY_GET = 0,
+  QUERY_GET_AGGREGATE,
+  QUERY_LIST,
```
bazsi: I think we should not use the term "aggregate" as long as it is only a "sum". It's misleading and other aggregations (like average) could be added later on.

lib/stats-query-commands.c
```diff
@@ -0,0 +1,107 @@
...
+  QUERY_CMD_MAX
+};
+
+typedef GString *(*QUERY_CMD)(const gchar *hds_path);
+
```
bazsi: Please don't use ALL_CAPS for type names.
What is "hds_path"? I don't get the meaning of "hds", maybe "stats_path"? Is that the right term?

lib/stats-query-commands.c
```diff
@@ -0,0 +1,107 @@
...
+static GString *
+_query_get_aggregated(const gchar *hds_path)
+{
+  return stats_query_get_aggregated(hds_path);
+}
+
+static gint
+_command_str_to_id(const gchar *cmd)
+{
```
bazsi: enum return?

lib/stats-query-commands.c
```diff
@@ -0,0 +1,107 @@
...
+    return QUERY_GET_AGGREGATE;
+
+  if (g_str_equal(cmd, "GET"))
+    return QUERY_GET;
+
+  if (g_str_equal(cmd, "LIST"))
+    return QUERY_LIST;
+
+  msg_error("syslog-ng-ctl query got unknown command", evt_tag_str("command", cmd));
+
```
bazsi: If I understand correctly, this file is the backend for "control/control-commands", and as such should be a generic interface not to be "infected" by concepts such as "syslog-ng-ctl", especially not in error reporting.

This function should return a FALSE and the error should be printed in higher layers..

lib/stats-query-commands.c
```diff
@@ -0,0 +1,107 @@
...
+  return QUERY_CMD_MAX;
+}
+
+static QUERY_CMD QUERY_CMDS[] =
+{
+  _query_get,
+  _query_get_aggregated,
+  _query_list
+};
+
```
bazsi: if we have an array like this, we might as well use to store the "string" representation of commands and use that instead of an if-else-if function to look up the command id.

lib/stats-query-commands.c
```diff
@@ -0,0 +1,107 @@
...
+static GString *
+_dispatch_query(gint cmd_id, const gchar *hds_path)
+{
+  if (cmd_id < QUERY_GET || cmd_id >= QUERY_CMD_MAX)
+    {
+      msg_error("syslog-ng-ctl got invalid query command",
+                evt_tag_int("cmd_id", cmd_id),
+                evt_tag_str("query", hds_path));
+      return g_string_new("");
```
bazsi: again, syslog-ng-ctl should only be used as a concept in higher layer, or this code could be moved there.

lib/stats/stats-cluster.c
```diff
@@ -39,21 +39,34 @@ stats_cluster_foreach_counter(StatsCluster *self, StatsForeachCounterFunc func,
...
-
   return tag_names[type];
 }
 
+gint
+stats_cluster_get_type_by_name(const gchar *name)
+{
+  if (!name)
+    return SC_TYPE_MAX;
+
```
bazsi: I would prefer extremal values to be negative, instead of the max, it's a lot easier to detect and test for.

lib/stats/stats-cluster.c
```diff
@@ -39,21 +39,34 @@ stats_cluster_foreach_counter(StatsCluster *self, StatsForeachCounterFunc func,
...
+  for (gint i = 0; i < SC_TYPE_MAX; i++)
+    if (!strcmp(name, tag_names[i]))
```
bazsi: please use braces here.

lib/stats/stats-query.c
```diff
@@ -0,0 +1,266 @@
...
+static const gchar *_counter_type[SC_TYPE_MAX+1] =
+{
+  "dropped",
+  "processed",
+  "stored",
+  "suppressed",
+  "stamp",
+  "MAX"
+};
+
```
bazsi: this should be delegated to the "stats-cluster" layer, and not replicated here.

lib/stats/stats-query.c
```diff
@@ -0,0 +1,266 @@
...
+  g_string_append_printf(str, COUNTER_WITH_VALUE_TEMPLATE, sc->query_key, ctr_name ? ctr_name : "", ctr->value);
+  return TRUE;
+}
+
+static gboolean
+_append_counter_without_value(StatsCluster *sc, StatsCounterItem *ctr, const gchar *ctr_name, gpointer user_data)
+{
+  GString *str = (GString *)user_data;
+  g_string_append_printf(str, COUNTER_TEMPLATE, sc->query_key, ctr_name ? ctr_name : "");
+  return TRUE;
```
bazsi: In this cases I find it more problematic to extract the name of the template as it is tightly coupled with the printf arguments anyway. And when I read the code, I need to break out and locate the template in the top of the file. Why is that useful?

tests/unit/Makefile.am
```diff
@@ -13,7 +13,8 @@ tests_unit_TESTS			= \
 	tests/unit/test_hostid		   \
 	tests/unit/test_zone		   \
 	tests/unit/test_pathutils	   \
-	tests/unit/test_logwriter
+	tests/unit/test_logwriter	\
+	tests/unit/test_stats_query
```
bazsi: Please don't add more tests to tests/unit, rather use a more specific directory. test_stats_query should rather go to lib/stats/tests probably.

lib/stats/stats-cluster.c
lbudai: well.. that enum is a little bit strange.
`SC_TYPE_MIN` is 0, as well as `SC_TYPE_DROPPED`,
but `SC_TYPE_MAX` is not equal to `SC_TYPE_STAMP`.
Of course, we can use -1, -2, -n, but it still not symmetric.
So I think  `SC_TYPE_MIN` and `SC_TYPE_MAX` should be the extremal values here :)

lib/stats-query-commands.c
kvch: What do you mean?

kvch:
@bazsi  Thanks for your review notes. I have fixed the code based on your notes. However, I created separate commits, so you can validate more easily if the fixes satisfies your requests.
Also, I am still waiting for responses to finish fixing the code. Then, I will rebase the branch and clear up the unnecessary commits.

PR #1336, opened 2017-02-03
README: rm drone.io as it is no longer maintined by syslog devs

kvch:
Remove build status on drone.io, because the building is no longer maintaned.


lbudai:
indeed...

PR #1344, opened 2017-02-10
astylerc: allow continuation lines to start until column 60

bazsi:
This is one of the things that I fail to properly do with our style-checks. I agree that very long lines should
not "continue", but I would permit lines that start up to the 60. character, indented by spaces. I think because of the long identifiers we have, this is a pretty common case.

The patch contains both the patch to astylerc and the resulting indentations.

I would love to hear your opinions. Merging this may need some thought, as it would probably generate conflicts with the rest of the pull requests, but those should be pretty easy to resolve.

Anyway, would be great to know your thoughts @kvch and @lbudai, whether 

1) you agree
2) when you think this could be merged

This patch adds --max-instatement-indent=60 to astylerc, as we are permitting
120 long lines and it makes sense to increase this as well.

This causes a number of reindents in the code, effectively reverting
part of the changes that were made when astyle was implemented. A lot
of our identifiers are pretty long due to the naming
convention.

### DIFF
--- a/.astylerc
+++ b/.astylerc
@@ -9,6 +9,7 @@
 
 # max line lenght is 120 chars
 --max-code-length=120
+--max-instatement-indent=60
 
 # pointer is aligned to the name of the variable
 --align-pointer=name
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -495,8 +495,8 @@ cfg_lexer_include_file_add(CfgLexer *self, const gchar *fn)
   level->include_type = CFGI_FILE;
 
   level->file.files = g_slist_insert_sorted(level->file.files,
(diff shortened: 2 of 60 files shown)


kira-syslogng:

bazsi:
ok to test

lbudai:
@bazsi: I agree with the changes.
when to merge: I think after travis is green it is ready to merge

kvch:
What astyle version do you use? It is weird that the style check fails now in Travis.

bazsi:
Now rebased and fixed the missing bits (I actually moved some of the required changes to another branch, that's why this was failing).

bazsi:
With the approval, I am merging this.

PR #1359, opened 2017-03-03
python: support clone for Python parsers

lbudai:


lbudai:
@pzoleex : could you test it?

pzoleex:
Ok this works now. If I add the same parser to 2 log path, two objects will be created.

Issue #1379, opened 2017-03-08
python logmsg.get(): not every keys are here


pzoleex:
There was this issue: https://github.com/balabit/syslog-ng/issues/1372

But the fix does not solve the problem in every cases because a lot of keys are missing.
For example, I added this two lines to the parse():
```
print("aaaaaaaaaaaaaaaaaa all keys: %s" % msg.keys())
print("aaaaaaaaaaa date: %s" % (msg['**DATE**']))
```
the result:

> aaaaaaaaaaaaaaaaaa all keys: [**'HOST', 'HOST_FROM', 'MESSAGE', 'PROGRAM', 'PID', 'SOURCE', '.SDATA.meta.sequenceId', '.SDATA.file@18372.4.position', '.SDATA.file@18372.4.size', '.SDATA.file@18372.4.name', '.SDATA.meta.sysUpTime'**]
> aaaaaaaaaaa date: Apr 15 11:01:10

As you can see, there is no DATE key in the keys, but I can access it. the keys() should give all of the available key.

PR #1382, opened 2017-03-10
python-parser: add macros to keys()

lbudai:
Please note, that hard macros are also added to the list of keys.
Hard macros are cannot be overwritten.

Fixes: #1379


pzoleex:
Hi,

This is almost good.
There is a key called '_' without any value.

lbudai:
@pzoleex : is it a real  problem? :)

pzoleex:
I can live with that just "strange"

PR #1389, opened 2017-03-21
cfg-lexer: fix token duplication in the output of '--preprocess-into'

MrAnno:
This PR fixes an old token duplication issue in the output of the `--preprocess-into` command line option.

Since every token in a `CfgTokenBlock` is an 'injected token', the injected flag should be set in all cases.

---------------------------

## Configuration

- `etc/syslog-ng-example.conf`:
```
@version: 3.9

parser p_parser {};
filter f_filter { host(""); };
rewrite r_rewrite {};

source s_source {};
destination d_destination {};

log {};
```

## Steps to reproduce

1. Start syslog-ng: `sbin/syslog-ng -Fedv -f etc/syslog-ng-example.conf --preprocess-into test.txt`

2. cat test.txt

Example output (notice the duplicated '}' tokens):

```
@version: 3.9

parser p_parserke {}};
filter f_filter { host(""); }};
rewrite r_rewrite {}};

source s_source {};
destination d_destination {};

log {};
```

### DIFF
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -818,10 +818,9 @@ relex:
           *yylloc = self->include_stack[self->include_depth].lloc;
           tok = token->type;
           if (token->type == LL_TOKEN)
-            {
-              tok = token->token;
-              injected = TRUE;
-            }
+            tok = token->token;
+
+          injected = TRUE;
 
           goto exit;
         }
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -795,18 +795,12 @@ _invoke__cfg_lexer_lex(CfgLexer *self, YYSTYPE *yylval, YYLTYPE *yylloc)
   return _cfg_lexer_lex(yylval, yylloc, self->state);
 }
 
-int
-cfg_lexer_lex(CfgLexer *self, YYSTYPE *yylval, YYLTYPE *yylloc)
+static gboolean
+cfg_lexer_consume_next_injected_token(CfgLexer *self, gint *tok, YYSTYPE *yylval, YYLTYPE *yylloc)
 {
-  CfgBlockGenerator *gen;
   CfgTokenBlock *block;
   YYSTYPE *token;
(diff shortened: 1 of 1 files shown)


lib/cfg-lexer.c
```diff
@@ -818,10 +818,9 @@ cfg_lexer_lex(CfgLexer *self, YYSTYPE *yylval, YYLTYPE *yylloc)
           *yylloc = self->include_stack[self->include_depth].lloc;
           tok = token->type;
           if (token->type == LL_TOKEN)
-            {
-              tok = token->token;
-              injected = TRUE;
-            }
+            tok = token->token;
```
presidento: I would use an if-else statement for setting the `tok` variable. (Resetting after 1 line does not make sense.)

MrAnno: Sure, thanks.
I'm going to push some refactor steps after @kira-syslogng's tests are completed.

kira-syslogng:
Build ERROR

lib/cfg-lexer.c
MrAnno: Done.

kira-syslogng:
Build ERROR

MrAnno:

kira-syslogng:
Build ERROR

MrAnno:

kira-syslogng:
Build ERROR

bazsi:
wow, just wow. Great work on both identifying the problem of duplication (has been a frustration for me for a long time) and the extraction of the function as well.

This function could still use some more of this elegance, but this is already a great step in a good direction.

I am not sure why @kira-syslogng does not like the patch, seems like an unrelated change to me and I can't access the details.

presidento:
@bazsi: @kira-syslogng fails because the master is red at the moment. (Every PR fails now.) It will be green as soon as we merge #1394.

presidento:

kira-syslogng:
Build SUCCESS

PR #1401, opened 2017-03-28
F/dynamic stats cluster registration

bodnartibi:
Use strings to register stats-cluster names instead of hard-coded constant.


kira-syslogng:

kvch:
@kira-syslogng ok to test

kvch:

lib/stats/stats-cluster.c
```diff
@@ -25,6 +25,103 @@
 
 #include <string.h>
 
+typedef struct _StatsComponentsElement
+{
+  GString *name;
+  int index;
```
kvch: Please use `gint` instead of `int`.

bodnartibi: Done

lib/stats/stats-cluster.c
```diff
@@ -25,6 +25,104 @@
...
+  gint i;
+  g_hash_table_destroy(stats_components);
+  stats_components = NULL;
+
+  for (i = 0; i < SCS_SOURCE_MASK; i++)
+    mapping_from_component_number_to_element[i] = NULL;
+}
+
+int
+stats_components_get_component_index(const gchar *name)
```
kvch: This function does more than its name implies. It would be nice, if it would be renamed or the part of its functionality would be extracted into a separate function.

kira-syslogng:
Build ERROR

kira-syslogng:
Build ERROR

kira-syslogng:
Build ERROR

pzoleex:
Seems this make syslog-ng significantly slower based on my test:
without it: 598488 (msg/sec)
with this: 442807 (msg/sec)

config:
@version: 7.0
options {
    keep_hostname(yes);
    keep_timestamp(no);
    stats_level(2);
    use_dns(no);
};
source s_network_12e38694b3664bd18946f823cd88c8c9 {
    network(ip(0.0.0.0)
 log_fetch_limit(1000)
 log_iw_size(10000)
 max_connections(10)
 port(514));
};
destination d_file_ee3e79478cae43b9a9215534eef3f7f1 {
    file("/var/testdb_working_dir/41466d82-7fa2-444c-bbbc-fab91426ec82.$HOST" flush_lines(1000)
 log_fifo_size(20000));
};
log {
source(s_network_12e38694b3664bd18946f823cd88c8c9);
destination(d_file_ee3e79478cae43b9a9215534eef3f7f1);

flags(flow-control);
};

loggen --inet --rate=1000000 --size=512 --interval=600 --active-connections=10 169.254.1.21 514

bazsi:
I would possibly map the strings to numbers early on, and then use the
numerical value in the fast path.

bodnartibi:
In /lib/logsource.c the stats_register_and_increment_dynamic_counter() function is called frequently, in this pull request this function uses stats_components_get_component_index() every time with the same values (SCS_HOST, SCS_SENDER, SCS_PROGRAM). Another unnecessary call is in /lib/stats-cluster.c when we compare the self->component with the return value of stats_components_get_component_index(SCS_GROUP).
Except these cases we use stats_components_get_component_index() only in init time.
So I have tried out an ugly hack, I got the indexes of these values (SCS_HOST, SCS_SENDER, SCS_PROGRAM, SCS_GROUP) in init time and stored them in global variables and used them instead of stats_components_get_component_index() call.
I measured the same performance (on my machine) as before.
The problem that I cannot find elegant places for these stuff. I do not want to use global variables, but I guess it would not be a nice solution to store these values as attributes of a class (e.g. logsource), because every instance would have the same values here.

kira-syslogng:

kira-syslogng:
Build ERROR

kira-syslogng:

pzoleex:
@kira-syslogng ok to test

kira-syslogng:
Build FAILURE

presidento:
Moved to purgatory: the API changed too much, and there are too many conflicts now, so we cannot merge it now, and it requires significant effort.

Issue #1403, opened 2017-03-29
git head does not compile on many platforms


czanik:
# syslog-ng

## Version of syslog-ng

git head (31849565a24b034618df8230aa72ea3c85ffb619)

## Platform

Fedora 25

# Issue

## Failure

Compilation fails:

```
  GEN      modules/java/org_syslog_ng_InternalMessageSender.h
make --no-print-directory all-recursive
  CC       modules/native/modules_native_libsyslog_ng_native_connector_a-native-grammar.o
In file included from /usr/include/bits/libc-header-start.h:33:0,
                 from /usr/include/limits.h:26,
                 from /usr/lib/gcc/x86_64-redhat-linux/7/include/limits.h:194,
                 from /usr/lib/gcc/x86_64-redhat-linux/7/include/syslimits.h:7,
                 from /usr/lib/gcc/x86_64-redhat-linux/7/include/limits.h:34,
                 from /usr/lib64/glib-2.0/include/glibconfig.h:11,
                 from /usr/include/glib-2.0/glib/gtypes.h:32,
                 from /usr/include/glib-2.0/glib/galloca.h:32,
                 from /usr/include/glib-2.0/glib.h:30,
                 from ./lib/compat/glib.h:30,
                 from ./lib/syslog-ng.h:35,
                 from ./lib/cfg-parser.h:28,
                 from modules/native/native-parser.h:28,
                 from modules/native/native-grammar.y:26:
/usr/include/features.h:181:3: error: #warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE" [-Werror=cpp]
 # warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"
   ^~~~~~~
cc1: all warnings being treated as errors
make[2]: *** [Makefile:9326: modules/native/modules_native_libsyslog_ng_native_connector_a-native-grammar.o] Error 1
make[1]: *** [Makefile:14106: all-recursive] Error 1
make: *** [Makefile:5752: all] Error 2
```

## Steps to reproduce

Compile syslog-ng git head on Fedora

My suspicion is: https://github.com/balabit/syslog-ng/commit/31849565a24b034618df8230aa72ea3c85ffb619 where: "-Werror" was added.

czanik:
[...]

PR #1398, opened 2017-03-27
increase compiler verbosity and clean up warnings

pprindeville:
Warnings about uninitialized variables are often accurate.

Signed-off-by: Philip Prindeville <philipp@redfish-solutions.com>

PR #1404, opened 2017-03-29
Remove hardcoded Werror

presidento:
We have the `Werror` flag in Travis, but we do not want to enforce it in every build environment.

This is a continuation of #1398.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -44,8 +44,8 @@ AM_CFLAGS =
 if ENABLE_EXTRA_WARNINGS
 AM_CFLAGS += \
 	-Wshadow	\
-	-Wno-initializer-overrides
-
+	-Wno-initializer-overrides \
+	-Wno-error=deprecated-declarations
 endif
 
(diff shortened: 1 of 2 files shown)


presidento:
Fixes #1403.

presidento:
I have 2 approvers, I merge it to make master green again.

Makefile.am
```diff
@@ -44,8 +44,8 @@ AM_CFLAGS =
 if ENABLE_EXTRA_WARNINGS
 AM_CFLAGS += \
 	-Wshadow	\
-	-Wno-initializer-overrides
-
+	-Wno-initializer-overrides \
+	-Wno-error=deprecated-declarations
```
pprindeville: So the `-Werror` is being dropped?

pprindeville: Never mind, I just looked at PR #1403.

presidento: A clean compile with Werrors is still required, since the [travis check adds this flag](https://github.com/balabit/syslog-ng/blob/master/.travis.yml#L41).

PR #1405, opened 2017-03-30
Update and extend syslog-debun's capabilities

jszigetvari:
* platform support extensions like AIX, SuSE and Slackware Linux
* enhance the capabilities in case of already supported platformas (Solaris, HPUX)
* extended and improved information gathering (startup scripts)
* refactoring of code
* bug fixes

### DIFF
--- a/contrib/syslog-debun
+++ b/contrib/syslog-debun
@@ -62,8 +62,12 @@ trace="strace -f"
 service_stop="/etc/init.d/syslog-ng stop"
 service_start="/etc/init.d/syslog-ng start"
 service_status="/etc/init.d/syslog-ng status"
-checkpid () { [ -d /proc/"$1" ]; }
-mywait () { jobs -p >${tmpdir}/sdn.jobs ; for i in `grep -v "\<$tailpid\>" ${tmpdir}/sdn.jobs` ; do wait $i ; done ; }
+initfile="/etc/init.d/syslog-ng"
+#checkpid () { [ -d /proc/"$1" ]; }
+checkpid () { [ -n "$1" ] && kill -0 "$1" 2>/dev/null ; }
+#mywait () { jobs -p >${tmpdir}/sdn.jobs ; for i in `grep -v "\<$tailpid\>" ${tmpdir}/sdn.jobs` ; do wait $i ; done ; }
+mywait () { jobs -p >${tmpdir}/sdn.jobs ; for i in `grep -v "^$tailpid\$" ${tmpdir}/sdn.jobs` ; do wait $i ; done ; }
+is_available () { which "$1" >/dev/null 2>&1; }
 
 ###
 ### Show Usage
@@ -475,6 +479,123 @@ debun_extra_debian () {
 	md5sum --quiet -c /var/lib/dpkg/info/syslog-ng*.md5sums && echo Package files are intact
 	echo "list syslog-related packages"
 	dpkg -l |grep syslog > ${tmpdir}/deb.packages
+}
+
+debun_extra_redhat () {
+	printf "\nRedhat specific checks\n"
+	echo "Check package files integrity"
+	rpm -V syslog-ng-premium-edition && echo "syslog-ng-premium-edition: Package files are intact"
+	rpm -V syslog-ng-premium-edition-client && echo "syslog-ng-premium-edition-client: Package files are intact"
+	rpm -V syslog-ng-premium-edition-compact && echo "syslog-ng-premium-edition-compact: Package files are intact"
+	echo "list syslog-related packages"
+	rpm -qa |grep syslog > ${tmpdir}/rpm.packages
+}
+debun_extra_genlinux () {
+	if is_available getenforce; then
+		getenforce >${tmpdir}/sys.selinux
+	else
+		echo "No getenforce in path." >${tmpdir}/sys.selinux
+	fi
+	sysctl -a >${tmpdir}/sys.sysctl.all 2>/dev/null
+}
+
+debun_linux () {
+	if [ "$dist" = "Debian" ]; then
+		add_extra debun_extra_debian
+	elif [ "$dist" = "Ubuntu" ]; then
+		add_extra debun_extra_debian
+	elif [ "$dist" = "CentOS" ]; then
+		add_extra debun_extra_redhat
+	elif [ "$dist" = "RedHatEnterpriseServer" ]; then
+		add_extra debun_extra_redhat
+	elif [ "$dist" = "RHEL" ]; then
+		add_extra debun_extra_redhat
+	elif [ "$dist" = "OracleServer" ]; then
+		add_extra debun_extra_redhat
+	elif [ "$dist" = "EnterpriseEnterpriseServer" ]; then
+		add_extra debun_extra_redhat
+	else
+		echo "Unknown Distro, perhaps unsupported"
+	fi
+
+	add_extra debun_extra_genlinux
+}
+
+debun_extra_gensolaris () {
+	sysdef >${tmpdir}/sys.sysdef
+	kstat >${tmpdir}/sys.kstat
+	cp /etc/release ${tmpdir}/sys.release
+	if is_available showrev ; then
+		showrev >${tmpdir}/sys.showrev
+	fi
+}
+
+### Here comes solaris specific parts
+debun_solaris () {
+	add_extra debun_extra_gensolaris
+
+	pkginfo | grep -i syslog > ${tmpdir}/pkg.packages
+
+	if is_available svcadm ; then
+		service_stop="svcadm disable syslog-ng"
+		service_start="svcadm enable syslog-ng"
+		service_status="svcs syslog-ng"
+	fi
+}
+
+debun_extra_freebsd() {
+	:
+}
+
+debun_freebsd() {
+	add_extra debun_extra_freebsd
+}
+
+debun_extra_hpux () {
+	sysdef >${tmpdir}/sys.sysdef
+	swlist >${tmpdir}/sys.swlist
+}
+
+debun_hpux () {
+	add_extra debun_extra_hpux
+}
+
+detect_env_linux () {
+	if is_available lsb_release ; then
+		lsb_release -a | tee ${tmpdir}/lsb.all
+		dist=`lsb_release -si`
+	elif [ -r /etc/debian_version ]; then
+		dist="Debian"
+	elif [ -r /etc/redhat-release ]; then
+		dist="RHEL"
+	fi
+}
+
+detect_env () {
+	###
+	### Detecting syslog-ng ver: ose or pe
+	###
+
+	echo "Start environment detection"
+	if [ -x /opt/syslog-ng/bin/loggen ] ; then
+		syslogfhs=unix
+		echo "Unix-like FHS detected"
+	elif [ -d /etc/syslog-ng/ ]; then
+		syslogfhs=linux
+		echo "Linux-type FHS detected"
+	else
+		syslogfhs=unknown
+		confdir=/nonexistent
+		echo "No syslog-ng detected"
+	fi
+
+	os=`uname -s`
+	if [ "$os" = "Linux" ]; then
+		detect_env_linux
+	fi
+}
+
+setup_env_debian () {
 	unset distpkgoffile
 	unset distpkgstatus
 	distpkgoffile () {
@@ -492,12 +613,7 @@ debun_extra_debian () {
 	#syslogrealbin=${binprefix}/sbin/syslog-ng
 }
 
-debun_extra_redhat () {
-	printf "\nRedhat specific checks\n"
-	echo "Check package files integrity"
-	rpm -V syslog-ng-pe && echo Package files are intact
-	echo "list syslog-related packages"
-	rpm -qa |grep syslog > ${tmpdir}/rpm.packages
+setup_env_redhat () {
 	unset distpkgoffile
 	unset distpkgstatus
 	distpkgoffile () {
@@ -514,47 +630,40 @@ debun_extra_redhat () {
 	}
 }
 
-debun_extra_genlinux () {
-	if hash getenforce; then
-		getenforce >${tmpdir}/sys.selinux
-	else
-		echo "No getenforce in path." >${tmpdir}/sys.selinux
-	fi
-	sysctl -a >${tmpdir}/sys.sysctl.a
 }
 
-debun_linux () {
-	add_extra debun_extra_genlinux
-	if  hash lsb_release ; then
-		lsb_release -a | tee ${tmpdir}/lsb.all
-		dist=`lsb_release -si`
-		if [ "$dist" = "Debian" ]; then
-			add_extra debun_extra_debian 
-		elif [ "$dist" = "Ubuntu" ]; then
-			add_extra debun_extra_debian
-		elif [ "$dist" = "CentOS" ]; then
-			add_extra debun_extra_redhat
-		elif [ "$dist" = "RHEL" ]; then
-			add_extra debun_extra_redhat
-		else 
-			echo "Unknown Distro, perhaps unsupported"
-		fi
-	elif [ -r /etc/debian_version ]; then
-		add_extra debun_extra_debian
-	elif [ -r /etc/redhat-release ]; then
-		add_extra debun_extra_redhat
 	fi
 	unset myplimit
 	myplimit () { cat /proc/$1/limits ; }
 }
 
+setup_env_linux () {
+	if [ "$dist" = "Debian" ]; then
+		setup_env_debian
+	elif [ "$dist" = "Ubuntu" ]; then
+		setup_env_debian
+	elif [ "$dist" = "CentOS" ]; then
+		setup_env_redhat
+	elif [ "$dist" = "RedHatEnterpriseServer" ]; then
+		setup_env_redhat
+	elif [ "$dist" = "RHEL" ]; then
+		setup_env_redhat
+	elif [ "$dist" = "OracleServer" ]; then
+		setup_env_redhat
+	elif [ "$dist" = "EnterpriseEnterpriseServer" ]; then
+		setup_env_redhat
 debun_extra_gensolaris () {
 	sysdef >${tmpdir}/sys.sysdef
 	kstat >${tmpdir}/sys.kstat
 	cp /etc/release ${tmpdir}/sys.release
 	if hash showrev ; then
 		showrev >${tmpdir}/sys.showrev
+	else
+		echo "Unknown Distro, perhaps unsupported"
 	fi
+
+	unset myplimit
+	myplimit () { [ -f "/proc/$1/limits" ] && cat /proc/$1/limits ; }
 }
 
 ### Here comes solaris specific parts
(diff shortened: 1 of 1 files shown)


kira-syslogng:

MrAnno:
@kira-syslogng ok to test.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

contrib/syslog-debun
```diff
@@ -31,25 +47,34 @@ pseao="ps -eao"
...
-service_stop="/etc/init.d/syslog-ng stop"
-service_start="/etc/init.d/syslog-ng start"
-service_status="/etc/init.d/syslog-ng status"
-checkpid () { [ -d /proc/"$1" ]; }
-mywait () { jobs -p >${tmpdir}/sdn.jobs ; for i in `grep -v "\<$tailpid\>" ${tmpdir}/sdn.jobs` ; do wait $i ; done ; }
+initfile="/etc/init.d/syslog-ng"
+service_stop="${initfile} stop"
+service_start="${initfile} start"
+service_status="${initfile} status"
+#checkpid () { [ -d /proc/"$1" ]; }
```
presidento: Why did you keep the outcommented code? (See below, also)

contrib/syslog-debun
```diff
@@ -456,6 +546,172 @@ debun_extra_debian () {
...
+	elif [ -r /etc/slackware-version ]; then
+		dist="Slackware"
+	elif [ -r /etc/SuSE-release ]; then
+		dist="SUSE LINUX"
+	fi
+}
+
+detect_env () {
+	###
+	### Detecting syslog-ng ver: ose or pe
```
presidento: I can't see how this function "detect version: ose or pe"

contrib/syslog-debun
```diff
@@ -1,3 +1,4 @@
+#!/bin/sh
```
pasztor: Not all the commits and improvements, but some of them is my legacy indeed.

contrib/syslog-debun
```diff
@@ -6,6 +7,17 @@
 ### This software may be used and distributed according to the terms of GNU GPLv2
 ### http://www.gnu.org/licenses/gpl-2.0.html
 
+### Check for "local" variable support
+if type local >/dev/null; then
+	:
+elif type bash >/dev/null; then
```
pasztor: Bash is not available oob on all of the syslog-ng's supported platform.
Eg. couple of freebsd's is amongst the supported platform, where bash is not available by default.
But on those systems the /bin/sh proven to be a good posix compatible shell, with local variable support.
On the other hand a couple of Solaris systems were also on the support matrix, where the /bin/sh was also posix compliant, but didn't provide local variables. But /bin/bash is available there, so exec()-ing "ourselves" to bash was a good alternative.
Based on my tests on the supported systems this two case was what I found:
#1 /bin/sh supports local variables and posix compliant
#2 /bin/sh doesn't support local variables, but /bin/bash is available oob. And /bin/bash supports local variables.
Hint: be careful with bash! Some of the supported platforms may contain a very early, like 2.0.something bash version. They follow posix rules, but don't depend on modern bash things!
If you depend only on what posix demands from a shell, then the script can remain as portable, as it was previously.

contrib/syslog-debun
pasztor: Actually it doesn't detect if it's ose or pe rather the install directory structure if it follows fhs, like
a) /etc, /usr/sbin, /var/whatever
b) /usr/local/ whaterver
c) /opt/syslog-ng/whatever
ose installations are a) if they come from OS package as part of the system, and managed by distro's package management
local installations from source usually follows b)
and PE installations follows c)

contrib/syslog-debun
```diff
@@ -31,25 +47,34 @@ pseao="ps -eao"
...
-service_start="/etc/init.d/syslog-ng start"
-service_status="/etc/init.d/syslog-ng status"
-checkpid () { [ -d /proc/"$1" ]; }
-mywait () { jobs -p >${tmpdir}/sdn.jobs ; for i in `grep -v "\<$tailpid\>" ${tmpdir}/sdn.jobs` ; do wait $i ; done ; }
+initfile="/etc/init.d/syslog-ng"
+service_stop="${initfile} stop"
+service_start="${initfile} start"
+service_status="${initfile} status"
+#checkpid () { [ -d /proc/"$1" ]; }
+checkpid () { [ -n "$1" ] && kill -0 "$1" 2>/dev/null ; }
```
pasztor: Seems interesting idea.
Nice research work!
It may worth a line of comment, where it came from, like this:
http://pubs.opengroup.org/onlinepubs/009695399/functions/kill.html

contrib/syslog-debun
```diff
@@ -152,12 +177,17 @@ done
 [ -n "$timeout" ] && [ -z "$debug_mode" ] && debun_usage 2 "Timeout without debug mode or packet capture"
 [ -n "$privacy_mode" ] && [ "x$debug_params" = "x$default_debug_params" ] && debug_params="$default_ldebug_params"
 #if [ -n "$pcap_params" ]; then
-#	hash tcpdump || debun_usage 2 "Packet capture requested, but no tcpdump in your PATH"
+#	hash tcpdump 2>/dev/null || debun_usage 2 "Packet capture requested, but no tcpdump in your PATH"
 #fi
 syslogbin=${binprefix}/sbin/syslog-ng
-syslogrealbin=${binprefix}/libexec/syslog-ng
```
pasztor: Are you sure, it's a good idea to remove that?
My memories are foggy about that part, but I have some foggy memories, that maybe on solaris versions, the syslog binary was wrapped by a simple shell script in sbin, and the actual binary was inside libexec, so ldd should test that.

contrib/syslog-debun
```diff
@@ -345,18 +376,28 @@ acquire_general_info () {
...
+	mkdir ${tmpdir}/config
+	find . > ${tmpdir}/syslog.etc.files
+	grep '\.key$' ${tmpdir}/syslog.etc.files > ${tmpdir}/syslog.etc.files.removed
+	grep -v '\.key$' ${tmpdir}/syslog.etc.files | \
+		while read FILE; do \
+			if ${grepq} "PRIVATE KEY-----" "${FILE}" 2>/dev/null; then
+				echo "${FILE}" >> ${tmpdir}/syslog.etc.files.removed
+			else
+				echo "${FILE}"
+			fi
```
pasztor: I wouldn't drop out any file, even if it's name end with .key., or if the contain the "PRIVATE KEY --- "string.
I have some foggy memories about I started to investigate this area, but I'm not sure, if I gave you notes about that, but...
Don't throw out keys unconditionally!
I suggest to check if openssl is exits, end openssl rsa -modulus or sg. can print out the modulus of a private key.
So, if you keep the certs, their modulus can be printed to, and you can check if a keyfile and a cert file belongs together without knowing the secret key or broke customer's privacy.
One more thing: I don't know if sng (or the underlying openssl) is able to handle cert and key in the same file, but, it would be wise, if you don't throw out the whole file, which contains private key, just find out the line number where the private key starts, where it ends, cut out that part, and put back a placeholder into what debug bundle creator collects, and that placeholder can contain the modulus.
Be careful! I tried that, and I know it's not easy!
You will bump into another nice thing, that head and tail are not exactly follows posix conventions on linuces, because they are gnu. eg. if you want to print a file from the 11.th line, then tail -n +11 may work on gnu versions, and as far as I remember, tail +11 is needed on non-gnu unices. However, that's another thing what can be detected, and it's better to detect that based on error codes and behaviour! ;-)

contrib/syslog-debun
```diff
@@ -402,44 +449,87 @@ acquire_syslog_var () {
...
+format_ldd_output () {
 	while read x ; do
 		x="${x#*=>}"
 		x="/${x#*/}"
 		x="${x%% (*}"
-		#[ "${x:0:2}" != "/ " ] && echo "$x" >>${tmpdir}/syslog.ldfiles
-		[ "${x}" != "/ " ] && echo "$x" >>${tmpdir}/syslog.ldfiles
-	done <${tmpdir}/syslog.ldd 
+		x="${x%%(*}"
+		[ "${x}" != "/ " ] && [ -f "${x}" ] && echo "$x"
```
pasztor: Nice rework. I assume why is this here, but not sure.
Would you mind add some example comment here showing what it transfrom to what?
Btw, thanks for this! I've just started to doubt in myself, because I have a suspicion, that ${x:0:2} is not in posix, so maybe a dash or ash can not interpret that brace expansion.

contrib/syslog-debun
```diff
@@ -456,6 +546,172 @@ debun_extra_debian () {
...
 	dpkg -l |grep syslog > ${tmpdir}/deb.packages
+}
+
+debun_extra_redhat () {
+	printf "\nRedhat specific checks\n"
+	echo "Check package files integrity"
+	rpm -V syslog-ng-premium-edition && echo "syslog-ng-premium-edition: Package files are intact"
+	rpm -V syslog-ng-premium-edition-client && echo "syslog-ng-premium-edition-client: Package files are intact"
+	rpm -V syslog-ng-premium-edition-compact && echo "syslog-ng-premium-edition-compact: Package files are intact"
+	echo "list syslog-related packages"
```
pasztor: I suggest to not copy code, and solve this problem like this:
pkgfound=0
for pkg in pkgname1 pkgname2 ... 
do
  if rpm -q $pkg ;
    pkgfound=$[pkgfound+1]
    if rpm -V $pkg;
     printf "%s: Package files are intact\n" "$pkg"
   else
     printf "%s: Package files are borked\n" "$pkg"
    fi
  fi
done
if [ $pkgfound -eq 0 ]; then
echo "I haven't found any syslog package on your system"
fi

contrib/syslog-debun
```diff
@@ -456,6 +546,172 @@ debun_extra_debian () {
...
+	find /var/log/packages -name "*sys*log*" | while read -r FILE; do echo "${FILE##*/}"; done > ${tmpdir}/pkg.packages
+}
+
+debun_extra_suse() {
+	printf "\nSuSE specific checks\n"
+	echo "Check package files integrity"
+	rpm -V syslog-ng-premium-edition && echo "syslog-ng-premium-edition: Package files are intact"
+	rpm -V syslog-ng-premium-edition-client && echo "syslog-ng-premium-edition-client: Package files are intact"
+	rpm -V syslog-ng-premium-edition-compact && echo "syslog-ng-premium-edition-compact: Package files are intact"
```
pasztor: Same here, what I wrote above for rpm.
It seems pretty much the same code.
Do not copy them! Consider to put the rpm integrity checking into a separate function, and call them from both place.
Btw.: You may consider to write package validity checking part for .deb systems aswell.

contrib/syslog-debun
```diff
@@ -402,44 +449,87 @@ acquire_syslog_var () {
...
+	echo "systemd detected..."
+	cp "/usr/lib/systemd/system/syslog-ng.service" "${tmpdir}/sys.service.syslog-ng"
+}
+
+acquire_syslog_startup_svc () {
+	echo "Solaris SMF detected..."
+	cp "/lib/svc/method/syslog-ng" "${tmpdir}/sys.svc-method.syslog-ng"
+	chmod 0660 "${tmpdir}/sys.svc-method.syslog-ng"
+	cp "/var/svc/manifest/system/syslog-ng.xml" "${tmpdir}/sys.svc-manifest.syslog-ng.xml"
+}
```
pasztor: Nice work!
Foggy memories again: I remember something, that I implemented some improvement for s*larish systems to support multiple syslog-ng instances.
What I am pretty sure about one of the clients happyly uses that solution.
So, just some hints: You may want to find out from svcs 's output, if there are more than one instance, and collect svcprop for all the instances or whatever you can find out from proper parametering of svcprop.

contrib/syslog-debun
```diff
@@ -31,25 +47,34 @@ pseao="ps -eao"
...
-mywait () { jobs -p >${tmpdir}/sdn.jobs ; for i in `grep -v "\<$tailpid\>" ${tmpdir}/sdn.jobs` ; do wait $i ; done ; }
+initfile="/etc/init.d/syslog-ng"
+service_stop="${initfile} stop"
+service_start="${initfile} start"
+service_status="${initfile} status"
+#checkpid () { [ -d /proc/"$1" ]; }
+checkpid () { [ -n "$1" ] && kill -0 "$1" 2>/dev/null ; }
+#mywait () { jobs -p >${tmpdir}/sdn.jobs ; for i in `grep -v "\<$tailpid\>" ${tmpdir}/sdn.jobs` ; do wait $i ; done ; }
+mywait () { jobs -p >${tmpdir}/sdn.jobs ; for i in `grep -v "^$tailpid\$" ${tmpdir}/sdn.jobs` ; do wait $i ; done ; }
+is_available () { which "$1" >/dev/null 2>&1; }
```
pasztor: Which?
Are you sure about that?
Which may not work on all platform. That's why I used type in most of the cases, because type is defined in posix.

contrib/syslog-debun
```diff
@@ -456,6 +546,172 @@ debun_extra_debian () {
...
+}
+
+debun_aix () {
+	add_extra debun_extra_aix
+
+	echo "Check package files integrity"
+	rpm -V syslog-ng-premium-edition && echo "syslog-ng-premium-edition: Package files are intact"
+	rpm -V syslog-ng-premium-edition-client && echo "syslog-ng-premium-edition-client: Package files are intact"
+	rpm -V syslog-ng-premium-edition-compact && echo "syslog-ng-premium-edition-compact: Package files are intact"
+	echo "list syslog-related packages"
```
pasztor: Same again! Avoid repeating yourself'

contrib/syslog-debun
pasztor: Or maybe the hash was the other solution what I used.
You have to check posix on that. I'm pretty sure, I did the research, just It was so handy at the time when I wrote that, that I haven't thought about documenting that research result into comments, I've just used them more or less consequently.
I suggest you the same: Do again the reseach.
As a starting point, use this link: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/hash.html

Summary: avoid wiches! ;-)

contrib/syslog-debun
pasztor: I can not speak in János's name, but I think I understand his feelings on this:
How he refactored it, seems a good idea. But haven't proved it's stability, portability and robustness yet.
So, if the old code is there in the comments, and you are in a situation, like they are in the support team, and you have to collect the information from the cu asap, then it can be handy, if you still have the old but proven code!
I also understand your feelings, that git remembers everything.
But, please consider these thoughts regarding this script:
This is the support team's swiss-army knife^Wcannon.

contrib/syslog-debun
pasztor: Haven't I wrote a comment here yet, to not repeat yourself? ;-)
Please consider package file consistency checking into a separate function! ;-)

pasztor:
Overall: Nice improvement on the code.
Consider my suggestions.
And one extra: I think it starts to pass level of complexity.
Don't forget to update README file of this project,
Maybe it worth to add a checklist / supportlist there, to entitle those systems where have you tested this.
It also worth some "backgroundcheck" work, to spin up a couple of vm instances, which run sng, and develop some framework for yourself, to automatically log in to the systems, have a couple of parameterlist for this debug script. And collect them all.
Collect an "etalon".
And every "rerun", test if the result is the same as the etalon. eg diff -burN on the etalon dir and on the freshly collected dir, from the same target host, with the same syslog-debun parameters.
You may want to do some "anonymization" on the extracted dir, to mask those things which are subject to change, like pids. But if you collect the "etalon", then collect it again, you will see what changes.
Hint: On your "test instances" I suggest to create more complicated scenarios, which helps you to test all the functions in the script.
One corner case which screams in my head: program destination!
I clearly remember, that on one of my test instances, one of the configs had a program destination, so syslog-ng had a (dumb) child process, which had another child process.
It's not that trivial to use. You may found traces about that from the time, on one of the mailinglists when Lavian@ asked around how to use that.
So that is the source where the whole recursive process list collecting idea came from.
If I haven't mention that previously: GOOD WORK! ;-)

jszigetvari:
Hello Guys,
Unfortunately I won't have much free time on my hands during the better part of the next week.
I will try to answer everyone's concerns, but the earliest time I will be able to do so will be the second part of next week.
Thank you for your patience!

contrib/syslog-debun
jszigetvari: Gyuri basically listed the reasons why it's safer to start off with /bin/sh.
I noticed that Solaris 9 only has some ancient bash, but I tested the scripts there as well, and it worked fine there as well. I only used bashishms more freely for the platforms where bash is readily available (Linux systems).

contrib/syslog-debun
jszigetvari: I just left it there for later reference. If the new code works well, the commented out parts will disappear in a later version. That was the logic behind it.
Usually support or customers simply download this script from somewhere other than git, so it seemed a good idea to leave some old and short code-snippets in the script, but commented out.

contrib/syslog-debun
jszigetvari: Gyuri basically said said it again. We are detecting the directory layout here. The OSE and PE are simply the flavors of the layouts.

contrib/syslog-debun
jszigetvari: The problem was that some Unix version I tested the code with (I can't recall which) did not have procfs, and I had to work around that case.
This seemed a fairly robust solution, and luckily it worked everywhere.
I will include that comment.

contrib/syslog-debun
jszigetvari: For now, I don't plan on implementing this in this current iteration, as it would take too much time.
I will put this on the wishlist for the next version.
I have added some extra code to generate the list of removed files. In later versions this feature could work with that list.

contrib/syslog-debun
jszigetvari: Yes, your method file (the one that can be extended easily to handle multiple instances) is now being shipped with syslog-ng. This current version does in deed not check for multiple instances, but can't see this feature being implemented in this changeset.
I will put this on the wishlist as well for a future version.

contrib/syslog-debun
jszigetvari: Yes, unfortunately type, hash and which are a big mess when compared across the different Unix systems.
Your version employed hash, but there were systems where it returned 0 even on missing commands, then I tried type, but that behaved this way as well on the different platforms.
Therefore I had to fall back to which, and even do a little extra output parsing on certain platforms, in order to get the right results :(

jszigetvari:
@pasztor
Thank you for the comments!
Regarding the README file, I don't feel that there are any changes necessary, as I did not change or add any CLI parameters, or change any behaviors of the script. The old options can be used just the same.
Yes, a platform support list may indeed be a good idea for the future.
I currently employ the testing system you passed on to me, but I added some bits here and there.
I am currently a bit overloaded with work, but I will consider your suggestions.
Thanks for the kind words! ;)

contrib/syslog-debun
jszigetvari: fixed in 928b085

contrib/syslog-debun
jszigetvari: ficed in fixed in 928b085

contrib/syslog-debun
jszigetvari: fixed in 928b085

jszigetvari: fixed in 928b085

contrib/syslog-debun
jszigetvari: added further explanation in 928b085

contrib/syslog-debun
jszigetvari: added mentioned comment in 928b085

kira-syslogng:
Build SUCCESS

contrib/syslog-debun
jszigetvari: I will add the Signed-off-by line.

kira-syslogng:
Build SUCCESS

MrAnno:
Thanks. :)

PR #1412, opened 2017-04-04
cfg.c: default log_msg_size is increased to 64Kbyte from 8Kb

pzoleex:
Because in modern log infrastructure the general log size is higher
and syslog-ng should process it by default.

### DIFF
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -303,7 +303,7 @@ cfg_new(gint version)
   self->time_reap = 60;
 
   self->log_fifo_size = 10000;
-  self->log_msg_size = 8192;
+  self->log_msg_size = 65536;
 
   file_perm_options_global_defaults(&self->file_perm_options);
 


pzoleex:

kira-syslogng:
Build ERROR

pzoleex:
Hm... Really interesting.
The problem is the next:
there is a source file, the size of source file (51 KB) < log_msg_size (64 KB).
If syslog-ng is restarted while it is reading the file, then syslog-ng won't start to read the remaining lines.
Until a new line comes...

In this test:
the source file contains 1000 lines.
syslog-ng starts, reads 400 lines then we stop it.
start syslog-ng -> nothing happens, however there are 600 not processed lines in the source file

append one more line to the source file -> syslog-ng reads all the 601 lines.

This bug is not related by this PR, always happens if log_msg_size > file size.

PR #1419, opened 2017-04-07
affile-source: add exit_on_eof options

kvch:
This PR is the next step to implement the command line functionality. It is the reimplementation of the  infrastructural part of exiting on EOF in #1326 based on feedbacks.

From now on it is possible to configure file and pipe sources to exit when EOF is reached.
```
file("/dev/stdin", exit_on_eof(yes));
```

Exit on EOF option was introduced for `LogTransport`, so it will be easier
to include more modules in the future.

However, another layering violation was introduced by
`log_proto_server_get_exit_on_eof`. I selected this point for the layering
violation, because there was already one and both of the violations are
the consequence of the same root cause. So next time if someone
re-factors the proto and transport parts, it will be easier to eliminate
these violations with the same method.
Another solution would have been to store `exit_on_eof` in
`LogProtoServerOptions` and change behaviour of `LogTransport` based on it.
This way the value could have been accessible both from `LogReader` and
`LogTransport`.  But it would have required more boilerplate code and
more complex "solution".

We agreed that `main_loop_exit` would be called from `GlobalConfig`. But it would have introduced circular dependencies, so `main_loop_exit` is called directly from `logreader.c`.

Signed-off-by: Noémi Ványi <sitbackandwait@gmail.com>

### DIFF
--- a/lib/logproto/logproto-server.h
+++ b/lib/logproto/logproto-server.h
@@ -123,6 +123,13 @@ log_proto_server_get_fd(LogProtoServer *s)
   return s->transport->fd;
 }
 
+static inline gboolean
+log_proto_server_get_exit_on_eof(LogProtoServer *s)
+{
+  /* FIXME: Layering violation */
+  return s->transport->exit_on_eof;
+}
+
 static inline void
 log_proto_server_reset_error(LogProtoServer *s)
 {
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -23,6 +23,7 @@
  */
 
 #include "logreader.h"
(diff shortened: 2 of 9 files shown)


kvch:

kira-syslogng:
Build ERROR

kira-syslogng:
Build ERROR

kvch:

kira-syslogng:
Build ERROR

kira-syslogng:

kira-syslogng:
Build ERROR

bazsi:
I would close it in favour of #1487

Issue #1334, opened 2017-01-31
csv-parser appears not to be thread safe when attached to a network source


cventers:
## Version of syslog-ng

```
syslog-ng 3.8.1
Installer-Version: 3.8.1
Revision:
Module-Directory: //usr/lib64/syslog-ng
Module-Path: //usr/lib64/syslog-ng
Available-Modules: dbparser,afprog,cef,graphite,afsmtp,csvparser,disk-buffer,affile,confgen,linux-kmsg-format,cryptofuncs,syslogformat,json-plugin,afuser,afstomp,pseudofile,add-contextual-data,afsocket,basicfuncs,date,system-source,kvformat,riemann,curl
Enable-Debug: on
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: off
```

## Platform

Linux <redacted hostname> 2.6.32-642.13.1.el6.x86_64 #1 SMP Wed Jan 11 20:56:24 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
CentOS release 6.8 (Final)

# Issue

## Failure

It seems that csv-parser does not work properly when attached to a network source and syslog-ng is running in threaded mode. This may be an unusual configuration, but it's one I've used quite successfully for many years and TBs of data from a 100+ node cluster under the 3.3 series.

The behavior under 3.8.1 is that the parser randomly fills the column macros with garbage and eventually segfaults (often because the current_column is being NULLed while another thread is trying to get its name).

If this is now an unsupported configuration, I recommend either a documentation warning or an assertion if threaded != no and the csv-parser is attached to a network source.

## Configuration

```
source s_feed {
    tcp(ip(IP REDACTED) port(1999) tls(
        key_file("KEY REDACTED")
        cert_file("CRT REDACTED")
        peer_verify(optional-trusted)
        ssl-options(no-sslv2, no-sslv3, no-tlsv1)
    ) max-connections(256) log-iw-size(25600));
};

parser p_hostsep {
    csv-parser(columns("IMSG.EPATH","IMSG.REMMSG")
        delimiters(" ")
        flags(drop-invalid,greedy,escape-none,strip-whitespace)
        quote-pairs("")
    );
};

log {
    source(s_feed);
    parser(p_hostsep);

    # ...
};

```

mitzkia:
I could reproduce the crash, 

used configuration:
```
[...]

PR #1368, opened 2017-03-07
Scratch buffers2

bazsi:
This branch implements an alternative scratch buffers implementation that should be easier to use and that solves the per-thread buffer allocation which is freed automatically.

PR #1421, opened 2017-04-10
csv-parser appears not to be thread safe when attached to a network source

furiel:
For issue:
https://github.com/balabit/syslog-ng/issues/1334

csv scanner object contained a few fields that are used for iterating
through columns. As multiple threads use the same scanner object,
those threads broke the state of the common iterator. This patch
separates the general scanner options and her iterator parts. The
iterator is created on the stack of the scanner parser function, thus
avoiding race condition.

Contextual data scanner also uses this scanner object, so she is also
affected by the internal api change. To maintain original mode of
operation of the contextual data scanner, the CSVScannerIterator
object is added to the CSVContextualDataRecordScanner, and the
contextual data scanner internal api is changed accordingly.


kira-syslogng:

modules/csvparser/csvparser.c
```diff
@@ -117,17 +117,21 @@ csv_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_o
   CSVParser *self = (CSVParser *) s;
   LogMessage *msg = log_msg_make_writable(pmsg, path_options);
 
-  csv_scanner_input(&self->scanner, input);
-  while (csv_scanner_scan_next(&self->scanner))
+  CSVScannerIterator iterator;
+  memset(&iterator, 0, sizeof(iterator));
+
```
bazsi: I think the memset() should be part of the init function.

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -392,43 +392,40 @@ csv_scanner_is_scan_finished(CSVScanner *self)
...
 
 void
-csv_scanner_input(CSVScanner *self, const gchar *input)
+csv_scanner_iterator_init(CSVScannerIterator *iterator, const gchar *input)
 {
-  self->src = input;
-  self->current_column = NULL;
+  iterator->src = input;
+  iterator->current_value = g_string_sized_new(128);
+  iterator->current_column = NULL;
```
bazsi: I think this is a perfect example where scratch_buffers2 would avoid another allocation in the fast path.
See pr #1368

lib/scanner/csv-scanner/csv-scanner.h
```diff
@@ -64,25 +64,28 @@ void csv_scanner_options_set_quotes(CSVScannerOptions *options, const gchar *quo
...
   GList *current_column;
   const gchar *src;
   GString *current_value;
   gchar current_quote;
+} CSVScannerIterator;
+
+typedef struct _CSVScanner
+{
+  CSVScannerOptions *options;
 } CSVScanner;
```
bazsi: The naming bothers me a little, CSVScanner was already implying some form of iteration. what's the role of the two data structures:

CSVScannerIterator is used to scan through the CSV formatted line, but what is CSVScanner?

If at all possible I'd stick to the CSVScanner name for the iterator role. Why can't we get rid off CSVScanner?

lbudai:
@bazsi: we will keep only CSVScanner :) before you entered to the room we discussed it and we concluded the same :)

lib/scanner/csv-scanner/csv-scanner.c
furiel: Good point thanks. Let's do this change in a separate changeset after pr is merged.
I created https://github.com/balabit/syslog-ng/pull/1421 so not to forget.

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -243,146 +243,146 @@ _match_string_delimiters_at_current_position(const char *input, GList *string_de
...
-csv_scanner_get_current_name(CSVScanner *self)
+csv_scanner_get_current_name(CSVScanner *scanner)
 {
-  return (const gchar *) self->current_column->data;
+  return (const gchar *) scanner->current_column->data;
 }
 
 gboolean
-csv_scanner_is_scan_finished(CSVScanner *self)
+csv_scanner_is_scan_finished(CSVScannerOptions *options, CSVScanner *scanner)
```
lbudai: Could you store a pointer `options` in the CSVScanner ?

kira-syslogng:

furiel:
@kira-syslogng retest please

furiel:

furiel:
ok to test

lib/scanner/csv-scanner/csv-scanner.h
```diff
@@ -80,9 +79,7 @@ gboolean csv_scanner_scan_next(CSVScanner *pstate);
 gboolean csv_scanner_is_scan_finished(CSVScanner *pstate);
 gchar *csv_scanner_dup_current_value(CSVScanner *self);
 
-void csv_scanner_input(CSVScanner *pstate, const gchar *input);
-gboolean csv_scanner_parse_input(CSVScanner *pstate);
-void csv_scanner_state_init(CSVScanner *pstate, CSVScannerOptions *options);
+void csv_scanner_init(CSVScanner *pstate, CSVScannerOptions *options, const gchar *input);
 void csv_scanner_state_clean(CSVScanner *pstate);
```
bazsi: isn't state_clean the pair of csv_scanner_init()? the naming does not suggest that.

kira-syslogng:
Build SUCCESS

PR #1432, opened 2017-04-12
csv-parser appears not to be thread safe when attached to a network source

furiel:
csv-parser appears not to be thread safe when attached to a network source

https://github.com/balabit/syslog-ng/issues/1334
csv scanner object contained a few fields that are used for iterating
through columns. As multiple threads uses the same scanner object,
those threads broke the state of the common scanner. This patch creates
the scanner object on the stack of the parse function of the parser,
so each threads will have their own dedicated scanner objects.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #1434, opened 2017-04-17
Fix broken headings in Markdown files

bryant1410:
GitHub changed the way Markdown headings are parsed, so this change fixes it.

See [bryant1410/readmesfix](https://github.com/bryant1410/readmesfix) for more information.

Tackles bryant1410/readmesfix#1


kira-syslogng:

PR #1435, opened 2017-04-17
Db parser deadlock2

bazsi:
This series of patches incorporate a number of refactoring steps within db-parser and also fix a deadlock situation reported by Evan Rempel.

The refactoring steps aim at two things:
 * split out a few helper classes to modules of their own
 * introduce a per-thread state that is used throughout pattern_db_proces() and its chain of function calls to simplify parameter lists

The fix itself builds on the second refactoring step: we use this state to store a list of messages to be emitted, instead of emitting them right away. This way we can release all locks through log_db_parser_emit(), which resolves the deadlock. Earlier a number of workarounds were present in the code, that were only partial fixes, this patch removes the workarounds and resolves the situation for good.

### DIFF
--- a/modules/dbparser/CMakeLists.txt
+++ b/modules/dbparser/CMakeLists.txt
@@ -21,6 +21,8 @@ set(PATTERNDB_SOURCES
     pdb-ruleset.h
     pdb-context.c
     pdb-context.h
+    pdb-ratelimit.c
+    pdb-ratelimit.h
     correllation.c
     correllation.h
     correllation-key.c
--- a/modules/dbparser/CMakeLists.txt
+++ b/modules/dbparser/CMakeLists.txt
@@ -19,6 +19,8 @@ set(PATTERNDB_SOURCES
     pdb-example.h
     pdb-ruleset.c
     pdb-ruleset.h
+    pdb-context.c
+    pdb-context.h
     correllation.c
     correllation.h
     correllation-key.c
--- a/modules/dbparser/CMakeLists.txt
+++ b/modules/dbparser/CMakeLists.txt
@@ -23,6 +23,7 @@ set(PATTERNDB_SOURCES
     pdb-context.h
     pdb-ratelimit.c
     pdb-ratelimit.h
+    pdb-lookup-params.h
     correllation.c
     correllation.h
     correllation-key.c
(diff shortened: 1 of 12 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
Hopefully this one fixes the @kira-syslogng failure.

kira-syslogng:
Build FAILURE

bazsi:
@kira-syslogng please retest

bazsi:
@kira-syslogng please test

presidento:

kira-syslogng:
Build FAILURE

bazsi:
I would appreciate some help with kira, i can't reproduce that locally.

bazsi:
@kira-syslogng test this please test=functions/tools/pdbtool/001-pdbtool-match-test.zts;

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:
Finally I got the bug found by @kira-syslogng and this is good to go.

@faxm0dem Evan has already tested this issue, could you perhaps run some tests yourself as well?

I am pretty confident about the fix itself, but the refactoring steps do have some risks associated. Would be great if this worked when 3.10 goes out without having to do an emergeny release.

Anything that excercises patterndb correllations would be useful.

Thanks

kira-syslogng:
Build FAILURE

bazsi:
Since master seems to be broken according to Kira, this branch now contains a revert for the patch that breaks master. Thus it should not be merged directly.

kira-syslogng:
Build SUCCESS

bazsi:
After the revert of 484e27665f68013a31eab3e762dcf20ef49e97e8 (currently on master), on this branch, kira was green, so the series itself is ok.

I have now removed the revert, which will cause kira to fail, but otherwise this patch series is good to go.

I've discussed it with @lbudai that this could go in, with the feedback from Evan Rempel, so Laci pls approve the PR.

Cheers,

kira-syslogng:
Build FAILURE

bazsi:

milanshah7:
Is this reproducible in v3.8.1? If Yes then Can anyone provide the steps for it?
@bazsi @kira-syslogng

bazsi:
Races are typically difficult to trigger and this issue seems to be quite old to make remembering the reproduction even more difficult.

With that said the original bug is almost certain to be present in 3.8 and I remember some more that come after this.

So I can't offer steps for reproduction and would recommend to test on a newer release. Binary packages for various distros are available and you can build your own (using dbld this is quite easy).

Issue #1436, opened 2017-04-18
Enhancement: Add SSL/TLS options to the http (curl) module


millert:
While not explicitly supported, it is possible to use https URLs with the http module (since curl supports this natively) but there is no way to control the SSL/TLS parameters.

Adding configuration options for the CA certificate directory/file, client certificate, client private key, cipher suite and peer verify flag would make the module much more useful for logging via https.  I will be submitting a PR that adds these options shortly.

PR #1437, opened 2017-04-18
Add SSL/TLS options to the http (curl) module

millert:
Add SSL/TLS related options to the http (curl) module.  It is already possible to use https URLs (since curl supports this natively) but there is no way to control the SSL/TLS parameters.  This diff add the following options: ca_dir, ca_file, cert_file, key_file, cipher_suite and peer_verify.  Note that unlike the main syslog-ng peer_verify option, this one just takes a boolean value.

Closes https://github.com/balabit/syslog-ng/issues/1436

Signed-off-by: Todd C. Miller <Todd.Miller@courtesan.com>

### DIFF
--- a/modules/http/http.c
+++ b/modules/http/http.c
@@ -396,20 +396,11 @@ http_dd_set_ssl_version(LogDriver *d, const gchar *value)
 }
 
 void
-http_dd_set_peer_verify(LogDriver *d, const gchar *value)
+http_dd_set_peer_verify(LogDriver *d, gboolean verify)
 {
-  gboolean verify;
   HTTPDestinationDriver *self = (HTTPDestinationDriver *) d;
 
-  if (type_cast_to_boolean(value, &verify, NULL))
-    {
-      self->peer_verify = verify;
-    }
-  else
-    {
-      msg_warning("curl: non-boolean value for TLS peer verify",
-                  evt_tag_str("peer_verify", value));
-    }
+  self->peer_verify = verify;
 }
 
 gboolean
--- a/modules/http/http.c
+++ b/modules/http/http.c
@@ -353,7 +353,11 @@ http_dd_set_ssl_version(LogDriver *d, const gchar *value)
 
   if (strcmp(value, "default") == 0)
     {
-      /* libcurl default, currently TLS 1.x */
+      /*
+       * Negotiate the version based on what the remote server supports.
+       * SSLv2 is disabled by default as of libcurl 7.18.1.
+       * SSLv3 is disabled by default as of libcurl 7.39.0.
+       */
       self->ssl_version = CURL_SSLVERSION_DEFAULT;
 
     }
@@ -373,21 +377,27 @@ http_dd_set_ssl_version(LogDriver *d, const gchar *value)
       /* SSL 3 only */
       self->ssl_version = CURL_SSLVERSION_SSLv3;
     }
+#ifdef CURL_SSLVERSION_TLSv1_0
   else if (strcmp(value, "tlsv1_0") == 0)
     {
       /* TLS 1.0 only */
       self->ssl_version = CURL_SSLVERSION_TLSv1_0;
     }
+#endif
+#ifdef CURL_SSLVERSION_TLSv1_1
   else if (strcmp(value, "tlsv1_1") == 0)
     {
       /* TLS 1.1 only */
       self->ssl_version = CURL_SSLVERSION_TLSv1_1;
     }
+#endif
+#ifdef CURL_SSLVERSION_TLSv1_2
   else if (strcmp(value, "tlsv1_2") == 0)
     {
       /* TLS 1.2 only */
       self->ssl_version = CURL_SSLVERSION_TLSv1_2;
     }
+#endif
   else
     {
       msg_warning("curl: unsupported SSL version",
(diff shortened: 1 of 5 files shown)


kira-syslogng:

presidento:
@kira-syslogng ok to test.

kira-syslogng:
Build SUCCESS

modules/http/http.c
```diff
@@ -282,6 +299,68 @@ http_dd_get_template_options(LogDriver *d)
...
+  if (type_cast_to_boolean(value, &verify, NULL))
+    {
+      self->peer_verify = verify;
+    }
+  else
+    {
+      msg_warning("curl: non-boolean value for TLS peer verify",
+                  evt_tag_str("peer_verify", value));
+    }
+}
```
bazsi: Apart from this function this patch set looks great. There are two issues from which one might only be a question:

1) is peer verify only a binary switch? Iirc the to destination allow a set of options there and we should keep this consistent

2) even if peer verify should be a boolean the tour safety functions are intended for a very different use case (value pairs type casting). So this function shouldn't be used here. Boolean options can be parsed using the yesno rule in the grammar, which yields an integer that can then be used to set this boolean value directly.

modules/http/http.c
```diff
@@ -153,6 +153,23 @@ _set_curl_opt(HTTPDestinationDriver *self, LogMessage *msg, struct curl_slist *c
...
+
+  if (self->cert_file)
+    curl_easy_setopt(self->curl, CURLOPT_SSLCERT, self->cert_file);
+
+  if (self->key_file)
+    curl_easy_setopt(self->curl, CURLOPT_SSLKEY, self->key_file);
+
+  if (self->ciphers)
+    curl_easy_setopt(self->curl, CURLOPT_SSL_CIPHER_LIST, self->ciphers);
+
```
bazsi: Shouldn't we have defaults in the court suite? Also, we might want to make it possible to control ssl versions or at least disable the vulnerable ones.

modules/http/http.c
ToddMiller1: The SSL syslog destination's peer-verify supports the following four values: optional-untrusted, optional-trusted, required-untrusted, required-trusted

I don't see a way to support the certificate being optional in https, at least not using the curl SSL/TLS options.  That is why I made it a binary switch.  If you think it is better to accept the same values as the syslog SSL destination and just treated the "optional" and "required" versions the same I'm happy to do that.

If you'd rather I used a different option name to avoid confusion, that would also work.

I didn't notice the yesno rules in the grammar, that would certainly be simpler.

modules/http/http.c
millert: By default, libcurl will use the default cipher suite compiled into the crypto library.  For OpenSSL this is normally "ALL:!EXPORT:!LOW:!aNULL:!eNULL:!SSLv2".  The compiled in default value can also be specified as "DEFAULT" for OpenSSL and GnuTLS.  The actual values allowed depends on the underlying crypto libs.  OpenSSL and GnuTLS use the same syntax but the syntax accepted by NSS is completely different.  So hard-coding a default value can be problematic since we don't know what crypto libs libcurl was built with.

Adding the ability to specify the SSL/TLS version is a good idea.  By default, libcurl will use TLS 1.x (not SSL).

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
What do you think about adding this patch to this patch-set as well:

```
diff --git a/modules/afsocket/afsocket-grammar.ym b/modules/afsocket/afsocket-grammar.ym
index 65d7b99..eabd9ac 100644
--- a/modules/afsocket/afsocket-grammar.ym
+++ b/modules/afsocket/afsocket-grammar.ym
@@ -665,6 +665,12 @@ tls_option
         : KW_IFDEF {
 }
 
+       | KW_PEER_VERIFY '(' yesno ')'
+         {
+           last_tls_context->verify_mode = $3
+               ? (TVM_REQUIRED | TVM_TRUSTED)
+               : (TVM_OPTIONAL | TVM_UNTRUSTED);
+          }
        | KW_PEER_VERIFY '(' string ')'
          {
            last_tls_context->verify_mode = tls_lookup_verify_mode($3);

```

This makes the http/tcp tls related options consistent. Cc @fekete-robert

millert:
That makes sense to me, consistency is good.

kira-syslogng:
Build SUCCESS

pzoleex:

kira-syslogng:
Build SUCCESS

bazsi:
As I see travis fails because a too old libcurl.

@czanik: tls version support come with libcurl 7.34. can you check if this
is a problem wrt distro support?

czanik:
@bazsi yes, it will be a problem: Fedora/openSUSE are OK, but even RHEL7 has an older curl version.
I'd make it optional (so distros without up-to-date curl could still build the http destination) and in non-distro repos we could also supply a recent curl package.

millert:
I'm able to build on RHEL 7 with libcurl 7.29 but probably RedHat has backported support for their package.  The simplest thing is to just #ifdef those bits so we don't try to use the defines if they are not present.

kira-syslogng:
Build SUCCESS

modules/http/http.c
bazsi: Thanks for the explanation

Issue #1204, opened 2016-09-09
json-c submodule is missing from the tarball


mochrul:
In the configure.ac I could choose to use internal json-c library (https://github.com/balabit/syslog-ng/blob/master/configure.ac#L201), but the release targz does not contain the lib. This can cause compilation issues.

MrAnno:
This can be fixed after merging json-c/json-c#345.

MrAnno:
There will be an additional requirement: we'll have to install doxygen to generate our tarball.

mochrul:
@bazsi I found this bug. And this issue still persists in 3.20.1.

PR #1442, opened 2017-04-20
Include ivykis and librabbitmq in the dist tarball

MrAnno:
This PR fixes `make dist` when syslog-ng is configured with `--with-ivykis=system` or `--with-librabbitmq-client=system`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -22,6 +22,7 @@
 #############################################################################
 
 SUBDIRS			=
+DIST_SUBDIRS		=
 AM_MAKEFLAGS		= --no-print-directory
 AM_YFLAGS		= -Wno-yacc -Wno-other
 
--- a/configure.ac
+++ b/configure.ac
@@ -944,17 +944,22 @@ dnl ***************************************************************************
 dnl ivykis headers/libraries
 dnl ***************************************************************************
 
+IVYKIS_INTERNAL_SOURCE_EXISTS="test -f $srcdir/lib/ivykis/src/iv_main_posix.c"
+
+if $IVYKIS_INTERNAL_SOURCE_EXISTS; then
+	AC_CONFIG_SUBDIRS([lib/ivykis])
+	IVYKIS_SUBDIRS=lib/ivykis
+fi
+
+
 INTERNAL_IVYKIS_CFLAGS=""
 if test "x$with_ivykis" = "xinternal"; then
-	if test -f "$srcdir/lib/ivykis/src/iv_main_posix.c"; then
-		AC_CONFIG_SUBDIRS([lib/ivykis])
-
+	if $IVYKIS_INTERNAL_SOURCE_EXISTS; then
 		# these can only be used in lib as it assumes
 		# the current directory just one below ivykis
 
 		IVYKIS_LIBS="-Wl,${WHOLE_ARCHIVE_OPT} -L\$(top_builddir)/lib/ivykis/src -livykis -Wl,${NO_WHOLE_ARCHIVE_OPT}"
 		IVYKIS_CFLAGS="-I\$(top_srcdir)/lib/ivykis/src/include -I\$(top_builddir)/lib/ivykis/src/include"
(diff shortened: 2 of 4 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
This behavior was on purpose, as even in this case the source tarball did be the same. Why do we have different source builds depending on where it was configured?

MrAnno:
In aacf1a55, `LIBMONGO_SUBDIRS` is used in the same way, so I thought this patch would be just an unification.

Also, if I run `make dist` with `--with-ivykis=system` or `--with-librabbitmq-client=system`, it fails because it tries to "cd" into a non-existing directory (`builddir/lib/ivykis` and `builddir/modules/afamqp/rabbitmq-c`) and tries to run a specific make target.

These directories (and makefiles) are missing, because the [configure invocation](https://github.com/balabit/syslog-ng/blob/609bbda93b659311dc32cac28d759edd6be42cf3/configure.ac#L950) is also conditional in these submodules.
I couldn't find a better solution, but I may have misunderstood something.

MrAnno:
```
$ ../configure --with-ivykis=system --with-librabbitmq-client=system
$ make dist

  YACC     lib/cfg-grammar.h
  GEN      lib/block-ref-grammar.y
  YACC     lib/block-ref-grammar.c
  YACC     lib/cfg-grammar.c
  LEX      lib/cfg-lex.c
  GEN      lib/filter/filter-expr-grammar.y
  YACC     lib/filter/filter-expr-grammar.c
  GEN      lib/parser/parser-expr-grammar.y
  YACC     lib/parser/parser-expr-grammar.c
...

if test -d "syslog-ng-3.9.1"; then find "syslog-ng-3.9.1" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "syslog-ng-3.9.1" || { sleep 5 && rm -rf "syslog-ng-3.9.1"; }; else :; fi
test -d "syslog-ng-3.9.1" || mkdir "syslog-ng-3.9.1"
 (cd lib/ivykis && make --no-print-directory top_distdir=../../syslog-ng-3.9.1 distdir=../../syslog-ng-3.9.1/lib/ivykis \
/bin/bash: line 14: cd: lib/ivykis: No such file or directory
Makefile:15289: recipe for target 'distdir' failed
make[1]: *** [distdir] Error 1
```

bazsi:
Hmm my goal was to make sure there are no differences between source tarballs based on configure settings. Configure detects compilation options after all.

I do remember this was working in the past, but maybe it worked because all internal libs were configured even if the system version was in use.

MrAnno:
Oh okay, I got it now.

As an alternative, we can run configure on `lib/ivykis` (and on other optional submodules) even if `--with-*=system` is specified. This way, the dist tarball would be complete.  Is this acceptable?

I know this would be weird too, but automake wants to "preinitialize" all directories in `DIST_SUBDIRS`, so makefiles should exist on those dirs when creating the tarball.

MrAnno:
While I was trying to include all submodules in the dist tarball, I encountered further problems:
- json-c was not included in the dist tarball **at all**. I tried to fix it, but they have an incorrect Makefile, so 'make dist' can not be run on this module
- libmongo is using another submodule (bson), which also has the "conditional configure" logic, so this is also a problem

rabbitmq and ivykis are OK, they can be included in the dist tarball easily.

bazsi:
Hi,

I think it's not a problem if not all optional stuff is enabled from a
standalone build, it's just the result should be a tarball that results in
a sensible build. Json could be argued, but rabbitmq is not that important.

With all this said, the important thing is that we produce a consistent
tarball irrespectively from the configure options.

mochrul:
I just want to mention here (as a cross reference) the ticket #1204 which is about the missing json-c module.

MrAnno:
In short:
Now ivykis and librabbitmq will be always included in the dist tarball (even if they are configured with the "system" flag).

json-c and libmongo will be fixed later in another PR.

kira-syslogng:
Build SUCCESS

PR #1444, opened 2017-04-24
cc-mask: Removed unnecessary group name

pzoleex:
syslog-ng automatically stores the first result in the $1 macro thus
we don't need to name it.
Otherwise the original solution did not work because PCRE does not supported group names
started with digits anymore.


pzoleex:
@kira-syslogng test this please test branch=pzolee-rewriter-credit-card; test=functions/other-functions/rewrite/;

kira-syslogng:
Build SUCCESS

Issue #1447, opened 2017-04-25
geoip: crashes at reload when it does not find database


pzoleex:
reproduction:
start syslog-ng with the config below
rm -rf /home/pzolee/GeoIP.dat (this is just a copy of /usr/share/GeoIP/GeoIP.dat)
reload syslog-ng

```
[2017-04-25T16:03:10.822081] Module loaded and initialized successfully; module='syslogformat'
Error Opening file /home/pzolee/GeoIP.dat
[2017-04-25T16:03:10.822144] Error initializing message pipeline;
[2017-04-25T16:03:10.822147] Error initializing new configuration, reverting to old config;
[2017-04-25T16:03:10.822190] Unable to detect fully qualified hostname for localhost, use_fqdn() will use the short hostname;
Error Opening file /home/pzolee/GeoIP.dat
[2017-04-25T16:03:10.822224] Error initializing message pipeline;
Kilépés (core készült)

```
bt:
```
(gdb) bt
#0  0x00007fe10d4be767 in kill () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007fe10ddc9558 in main_loop_reload_config_apply (user_data=0x7fe10e061b20 <main_loop>) at ../lib/mainloop.c:230
#2  0x00007fe10ddcac66 in main_loop_worker_sync_call (func=0x7fe10ddc9347 <main_loop_reload_config_apply>, user_data=0x7fe10e061b20 <main_loop>) at ../lib/mainloop-worker.c:346
#3  0x00007fe10ddc9758 in main_loop_reload_config_initiate (user_data=0x7fe10e061b20 <main_loop>) at ../lib/mainloop.c:287
#4  0x00007fe10ddc98ca in sig_hup_handler (user_data=0x7fe10e061b20 <main_loop>) at ../lib/mainloop.c:345
#5  0x00007fe10de1a60f in iv_signal_event (_this=0x7fe10e061b28 <main_loop+8>) at ../../../../lib/ivykis/src/iv_signal.c:170
#6  0x00007fe10de187b7 in iv_event_raw_got_event (_this=0x7fe10e061b68 <main_loop+72>) at ../../../../lib/ivykis/src/iv_event_raw_posix.c:89
#7  0x00007fe10de18f22 in iv_fd_poll_and_run (st=0x1cdf280, to=0x7fffdcdcd820) at ../../../../lib/ivykis/src/iv_fd.c:163
[...]

PR #1449, opened 2017-04-26
geoip: crash when start->deleting geoip.dat->reload

furiel:
By this patch, syslog-ng will store geoip data in persistent
config. When syslog-ng is reloaded and the data file is not available,
syslog-ng will fall back to the cached data.

Correction for:
https://github.com/balabit/syslog-ng/issues/1447

### DIFF
--- a/modules/geoip/geoip-parser.c
+++ b/modules/geoip/geoip-parser.c
@@ -156,8 +156,6 @@ geoip_parser_free(LogPipe *s)
   g_free(self->database);
   g_free(self->prefix);
 
-  GeoIP_delete(self->gi);
-
   log_parser_free_method(s);
 }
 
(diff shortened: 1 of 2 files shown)


mochrul:
How big this geoip database usually?

kira-syslogng:
Build SUCCESS

furiel:
On a xenial machine the shipped GeoIP.dat is
$ ls -l /usr/share/GeoIP/GeoIP.dat | awk  '{print $5}'
1160739

The maxmind version of the database is somewhat smaller. 
$ wget -q -O - http://geolite.maxmind.com/download/geoip/database/GeoLiteCountry/GeoIP.dat.gz | gunzip | wc -c
1082230

About 1,2M.

mochrul:
I jst asked this because

modules/geoip/geoip-parser.c
```diff
@@ -156,20 +156,29 @@ geoip_parser_free(LogPipe *s)
...
+    {
+      GeoIP *_gi = GeoIP_open(self->database, GEOIP_MMAP_CACHE);
+      if (_gi)
+        {
+          if (self->gi)
+            GeoIP_delete(self->gi);
+          self->gi = _gi;
+        }
+      else
+        self->gi = cfg_persist_config_fetch(cfg, "geoip-obj");
```
furiel: Offline code review has brought an issue to check before merge. Above we open the .dat file with GEOIP_MMAP_CACHE. The question is if it is still safe to to access the object when we delete the backed file in case of mmap allocation.

furiel:
Saving the pointer in persistent config is not enough indeed.

From the man page of mmap: 
[...]
Use of a mapped region can result in these signals:
SIGBUS Attempted access to a portion of the buffer that does not correspond to the file (for example,  beyond the end of the file, including the case **where another process has truncated the file**).

I created a short code to check this behaviour, and I could indeed trigger SIGBUS.
https://gist.github.com/furiel/0b80bb2ab726fa37debb61c28d554196

bazsi:
mmap() remains functional if the file is deleted, but not if it is changed
(the latter is the whole point). However our threat model allows local root
to be stupid, so I don't think we need to handle the modified case.

Package upgrades or cp even will produce a new file. So the most that we
should do is to document this behaviour and advice against in-place
modifications.

PR #1450, opened 2017-04-27
tlscontext: Implement ECDH support

pfeiffersz:


kira-syslogng:
Build SUCCESS

lib/tlscontext.c
```diff
@@ -386,6 +413,10 @@ tls_context_setup_session(TLSContext *self)
         }
       else
         msg_debug("empty ssl options");
+
+      if (!tls_context_setup_ecdh(self->ssl_ctx, SN_X9_62_prime256v1))
+          goto error;
```
bazsi: Is the curve name always available? Why pass it as a parameter, as i see it is only used in one of the code paths, in other cases it is unused.

lib/tlscontext.c
```diff
@@ -305,6 +305,33 @@ file_exists(const gchar *fname)
...
+#elif OPENSSL_VERSION_NUMBER > 0x10001000L
+  EC_KEY *ecdh = EC_KEY_new_by_curve_name(ecdh_nid);
+  if (!ecdh)
+    return FALSE;
+
+  if (SSL_CTX_set_tmp_ecdh(ctx, ecdh) != 1)
+    return FALSE;
+  EC_KEY_free(ecdh);
+#else
+  return FALSE;
```
bazsi: Can't we somehow extract this function to compat/openssl-support? That's where we add upgrade code to old openssl so we can use the latest API in our own code.

Would be great.

Hmm what if we defined SSL_CTX_set_ecdh_auto in there, doing roughly the same that you have here, except for cases where openssl already defines it.

lib/tlscontext.c
pfeiffersz: AFAIK it is available. It will be easier to give a configuration option, if it is a parameter.

presidento:
Close in favor of #1603.

PR #1451, opened 2017-04-27
tlscontext: Prefer server side cipher suite order

pfeiffersz:


mochrul:
I´m not fully aware of this feature, but is this a good thing to do unconditionally? Sometimes the server is not in our hand, so keeping the possibility to control this behavior can be helpful in that case.

kira-syslogng:
Build SUCCESS

bazsi:
By reading about this feature, I think it makes sense to control algorithms on the server side, and this is all what this patch does. So, I'd merge this as is.

PR #1454, opened 2017-04-28
Stats cluster refact

lbudai:
- Motivation
    - support new types of counters that are not fit into the existing StatsCluster implementation
    - example for new types: LogQueue related metrics (queue size in bytes which is definitely not a STORED/PROCESSED/... kind of counter)
- Changes initiated by this PR
    - extracted StatsClusterKey
        - (id, component, instance)  group was used as a key for StatsCluster so this is just an equivalent change and as a side effect the interface is simpler now (number of parameters has been reduced significantly for register_counter method-family)
    - initiated StatsCounterGroup (this name is not the final one, maybe StatsClusterCounters is better... )
        - number of counters can vary
        - the group initializer(that is reponsible for allocating counters) method added to StatsClusterKey

PR #1459, opened 2017-05-03
Add "index" to counter container

kvch:
Motivation
========

The main goal of this PR is to support fully globbing when searching for counters and eliminate workarounds introduced in an earlier PR. So from now on it is possible to search using `src.*` and get all of the counters which name stats with `src.`. There is no need to add further `.*` to get all counters.

Changes
=======

* `StatsCounterItem` has an attribute named `name` which consist of the `query_id` of its `StatsCluster` and the name of the type.
* Add `counter_index` hash table where the key is name of the counter and the value is the pointer to the counter. As the construction of the key is expensive, it is only generated once when a `StatsCounterItem` is queried.
* Introduce `indexed_mask` for `StatsCluster`s to know which counter is in `counter_index`.

Dependencies
------------------

- [x] It is rebased on top of #1454 and can be merged after that one.

### DIFF
--- a/lib/stats/stats-cluster.c
+++ b/lib/stats/stats-cluster.c
@@ -222,6 +222,13 @@ stats_cluster_is_alive(StatsCluster *self, gint type)
   return ((1<<type) & self->live_mask);
 }
 
+
+gboolean
+stats_cluster_is_indexed(StatsCluster *self, gint type)
+{
+  return ((1<<type) & self->indexed_mask);
+}
+
 StatsCluster *
 stats_cluster_new(const StatsClusterKey *key)
 {
--- a/lib/stats/stats-cluster.c
+++ b/lib/stats/stats-cluster.c
@@ -219,6 +219,8 @@ _stats_build_query_key(StatsCluster *self)
 gboolean
 stats_cluster_is_alive(StatsCluster *self, gint type)
 {
+  g_assert(type < self->counter_group.capacity);
+
   return ((1<<type) & self->live_mask);
 }
 
@@ -226,6 +228,8 @@ stats_cluster_is_alive(StatsCluster *self, gint type)
 gboolean
 stats_cluster_is_indexed(StatsCluster *self, gint type)
 {
+  g_assert(type < self->counter_group.capacity);
+
   return ((1<<type) & self->indexed_mask);
 }
 
--- a/lib/stats/stats-cluster.c
+++ b/lib/stats/stats-cluster.c
@@ -242,11 +242,18 @@ stats_counter_group_free(StatsCounterGroup *self)
     self->free_fn(self);
 }
 
+static void
+_free_stats_counters_in_cluster(StatsCluster *sc, gint type, StatsCounterItem *counter, gpointer user_data)
+{
+  stats_counter_free(counter);
+}
+
 void
 stats_cluster_free(StatsCluster *self)
 {
   _stats_cluster_key_cloned_free(&self->key);
   g_free(self->query_key);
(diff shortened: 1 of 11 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/stats/stats-counter.h
```diff
@@ -72,6 +73,15 @@ stats_counter_get(StatsCounterItem *counter)
   return result;
 }
 
+static inline gchar *
+stats_counter_get_name(StatsCounterItem *counter)
+{
+  if (counter && counter->name)
+    return counter->name;
```
lbudai: I think checking whether `counter->name` is not `NULL` is superfluous.
When counter is `NULL`, that's an error, IHMO and it's not the same case when a counter does not have a name.

lbudai: So the question is: do we really need this function?

lib/stats/stats-cluster.c
```diff
@@ -222,6 +222,13 @@ stats_cluster_is_alive(StatsCluster *self, gint type)
   return ((1<<type) & self->live_mask);
 }
 
+
+gboolean
+stats_cluster_is_indexed(StatsCluster *self, gint type)
+{
+  return ((1<<type) & self->indexed_mask);
```
lbudai: I'd add an `assert` for checking if type is in a valid range.

lib/stats/stats-query.c
```diff
@@ -28,125 +28,124 @@
...
+    }
 }
 
-static const gchar *
-_get_name_of_counter_item(NamedStatsCounterItem *item)
+static gchar *
+_construct_counter_item_name(StatsCluster *sc, gint type)
 {
-  return _get_counter_name_by_index(item->index);
+  GString *name = g_string_new("");
```
lbudai: another candidate for using the new scratchbuffer

lib/stats/stats-query.c
```diff
@@ -28,125 +28,124 @@
...
-            }
-        }
-    }
+  counter_full_name = stats_counter_get_name(counter);
+  if (counter_full_name == NULL)
+    counter_full_name = _construct_counter_item_name(sc, type);
+
+  sc->counter_group.counters[type].name = counter_full_name;
+  g_hash_table_insert(counter_index, counter_full_name, &sc->counter_group.counters[type]);
+  sc->indexed_mask |= type;
```
lbudai: Factor out the counter registration to a separate function (and do not pass the `StatsCounterItem`  to that function).

lib/stats/stats-query.c
```diff
@@ -28,125 +28,124 @@
...
 }
 
-void
-_split_expr(const gchar *expr, gchar **key, gchar **ctr)
+static void
+_remove_counter_from_index(StatsCluster *sc, gint type)
```
lbudai: yep... the function I asked for is the symmetry pair of this function.

kira-syslogng:
Build SUCCESS

lib/stats/stats-counter.h
kvch: It also has a `stats_counter_get` which returns the value. Another getter for the name was created to have a consistent interface for `StatsCounterItem`.

kira-syslogng:
Build SUCCESS

kvch:
I addressed the problems in the code. @lbudai Could you please validate the fixes?

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #1461, opened 2017-05-03
cfg-lex: handle API change in flex

### LINKED ISSUE #1460: Infinite loop with flex >= 2.6.1 if config last expression is a comment without newline
If syslog-ng was compiled with flex >=2.6.1, and the last expression in a config is a comment line without newline, the syslog-ng gets in an infinite loop state during config parsing.

## Steps to reproduce

1. Build syslog-ng with flex >= 2.6.1
2. `echo -n "#">etc/syslog-ng.conf`
3. `sbin/syslog-ng -s`
4. Infinite loop with 100% CPU usage.

presidento:
It caused infinite loop if the config file had a comment line without newline as a last expression. (We got 0 instead of EOF, so we never stopped reading.)

Fixes #1460

### DIFF
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -168,7 +168,7 @@ word	[^ \#'"\(\)\{\}\\;\r\n\t,|\.@:]
                              int ch;
 
                              ch = input(yyscanner);
-                             while (ch != '\n' && ch != EOF)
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

PR #1462, opened 2017-05-05
stats-query: lock index before editing it

kvch:
In case of paralell querying through stats_query a race could be possible when editing the index records. But not anymore! :)

### DIFF
--- a/lib/stats/stats-query.c
+++ b/lib/stats/stats-query.c
@@ -30,6 +30,7 @@
 
 
 static GHashTable *counter_index;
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
I think we could follow the same pattern as in case of stats lock...
but I can accept this PR now.

PR #1466, opened 2017-05-08
stats: replace type of StatsCounter value from gint to gssize

lbudai:
On 64 bit platforms we can use 64 bit atomic counters.


kira-syslogng:
Build SUCCESS

furiel:
Ship it!

PR #1467, opened 2017-05-08
kvformat: invalidate value_was_quoted before extracting the next value

MrAnno:
`_skip_initial_spaces()` calls the `_match_delimiter()` function which uses the value of `value_was_quoted`.

Since this variable contains the state of the previously parsed value, it should be invalidated.


kira-syslogng:
Build SUCCESS

PR #1486, opened 2017-05-19
stats-view: fix double free of written counter

kvch:
In a counter name is query, it was freed twice. Once when the views were freed and once when the index was freed. From now on it is fixed,


kira-syslogng:

PR #1487, opened 2017-05-21
Alternative exit on eof

bazsi:
This  supersedes #1419. It makes that implementation simpler by moving the exit trigger to affile-source.

I still see lots of refactorint possibilities, and would probably go toward creating a dedicated stdin() source, no other file source would need this option.

@robert-fekete: I wouldn't document this flag, so we can kill it later.

Btw: an stdin() scl is missing from here. So plan should be:
* Add stdin() scl, document that but not the flag
* Do the refactors, create stdin() driver, and kill both the flag and the scl


kira-syslogng:

kira-syslogng:

bazsi:
I have now added the stdin() & stdin-raw() sources, plus also fixed the travis failure, so this should be ready to go.

kira-syslogng:

kvch:
AFAIK we agreed last time that adding a boolean flag to `log_transport_file/pipe_init` and `log_transport_file/pipe_new` is not the way to introduce `exit_on_eof` configuration. Our common solution was introducing a new transport class for these cases. What changed since? I don't see how this PR is better than the solution we agreed on.

bazsi:
The very first solution pushed the notification interface into the LogTransports layer and decided whether to use the stdin specific LogTransport based on a filename. I had problems with these. The stdin specific constructor was a good idea.

The second iteration dropped the constructor and added  broad interface to LogTransport, used only by this specific case. I think the constructor is a better solution.

While working on the code and rebasing it to current master, I've found a number of other cases where ignoring the eof was a bad idea (program source for instance). So its not just stdin where this behavior is needed, that's why I added the flag.

We might want constructors for separate use-cases: stdin (regardless it's a pipe or a file), tailed file, multi writer pipe (named pipe), single writer pipe (program destination).

Also, we shouldn't try to open /dev/stdin explicitly (which is not portable). we should just be using fd 0. But the code requires an open by name now, and also validates if it's indeed the proper type, displaying a warning if it's not. The current FileOpenOptions based file open should be reworked so this becomes easy and simple.

configure.ac
```diff
@@ -536,6 +536,7 @@ AC_CHECK_HEADERS(strings.h	\
 AC_CHECK_HEADERS(tcpd.h)
 
 AC_CHECK_TYPES([struct ucred, struct cmsgcred], [], [], [#define _GNU_SOURCE 1
+#define _DEFAULT_SOURCE 1
```
presidento: Excellent, I would merge (cherry-pick) this commit now.

lib/logreader.c
```diff
@@ -359,9 +359,11 @@ log_reader_fetch_log(LogReader *self)
       switch (status)
         {
         case LPS_EOF:
+          g_sockaddr_unref(aux.peer_addr);
+          return NC_CLOSE;
         case LPS_ERROR:
           g_sockaddr_unref(aux.peer_addr);
-          return status == LPS_ERROR ? NC_READ_ERROR : NC_CLOSE;
+          return NC_READ_ERROR;
```
presidento: :+1:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

bazsi:
As I have repurposed this branch to cover a pretty large refactor, I am opening a new pull request and closing this.

Issue #1446, opened 2017-04-25
geoip: Invalid database type GeoIP Country Edition warning message


pzoleex:
This is an really annoying warning(?) message:
"Invalid database type GeoIP Country Edition, expected GeoIP City Edition, Rev 1"

If I add the city edition instead then:
"Invalid database type GeoIP City Edition, Rev 1, expected GeoIP Country Edition"

bonus: the function itself works, just gives this message every time when a new message comes.

This message should not come, and especially not to stderr without any formatting (this is not a syslog-ng debug message).

```
pzolee@Thor-E5440:~/projects/syslog-ng-ose/install$ sbin/syslog-ng -Fe
syslog-ng: Error setting capabilities, capability management disabled; error='Operation not permitted'
[2017-04-25T15:51:14.283711] WARNING: Configuration file format is newer than the current version, please specify the current version number (3.9) in the @version directive. syslog-ng will operate at its highest supported version in this mode;
[2017-04-25T15:51:14.283797] Plugin module not found in 'module-path'; module-path='/home/pzolee/projects/syslog-ng-ose/install/lib/syslog-ng', module='geoip'
[2017-04-25T15:51:14.303182] syslog-ng starting up; version='3.9.1'
[2017-04-25T15:51:16.984564] Syslog connection accepted; fd='14', client='AF_INET(127.0.0.1:41188)', local='AF_INET(0.0.0.0:1212)'
Invalid database type GeoIP Country Edition, expected GeoIP City Edition, Rev 1
[2017-04-25T15:51:16.984919] Syslog connection closed; fd='14', client='AF_INET(127.0.0.1:41188)', local='AF_INET(0.0.0.0:1212)'
[2017-04-25T15:52:10.216462] Syslog connection accepted; fd='14', client='AF_INET(127.0.0.1:41192)', local='AF_INET(0.0.0.0:1212)'
Invalid database type GeoIP Country Edition, expected GeoIP City Edition, Rev 1
[2017-04-25T15:52:10.216605] Syslog connection closed; fd='14', client='AF_INET(127.0.0.1:41192)', local='AF_INET(0.0.0.0:1212)'

```
reproduction:
start syslog-ng with the config below
send a message:echo "<38>Jan  1 14:45:22 95.108.142.138 prg00000[1234]: test message" | nc 127.0.0.1 1212

config:
```
@version: 7.0
@module geoip

options {
stats_freq(0);
keep_hostname(yes);
};

[...]

PR #1495, opened 2017-05-24
geoip: legacy library - invalid database warning

furiel:
Fixing: https://github.com/balabit/syslog-ng/issues/663
and https://github.com/balabit/syslog-ng/issues/1446

This patchset removes the invalid database warning that is written by geoip plugin in syslog-ng in certain circumstances.

There are two database types that operators can use: city and the country version. In the original behaviour, syslog-ng tried to interpret the database as if it was city version: GeoIP_record_by_name, and in case it fails, she falls back to the country version: GeoIP_country_code_by_name.
But in case of mismatch, libgeoip writes an unmaskable warning about invalid database. Hence the problem.

The resolution is the following: during init phase, syslog-ng detects the database type, and uses only the appropriate api call that is compatible with the detected database type. This fixes the geoip parser.

For the template function version there is one slight addition. 
When database is not provided, libgeoip will look for the database file in a hard coded path (set by package maintainers using a compiler option): GeoIP_new(GEOIP_MMAP_CACHE);
In the original implementation syslog-ng did not support to specify database location for template parameters. Only the "official" location provided by package maintainers was used. And the fallback mechanic that was included in the parser version was missing, so operators were forced to use the country version.

By this patchset operators can specify database location for geoip template function as well:
destination d_file {
   file("/tmp/output" flags(syslog-protocol)
   template("$(geoip --database GeoLiteCity.dat ${HOST})"));
};

### DIFF
--- a/modules/geoip/geoip-helper.c
+++ b/modules/geoip/geoip-helper.c
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2017 Balabit
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "geoip-helper.h"
+
+#include <messages.h>
+
+gboolean
+is_country_type(GeoIPDBTypes database_type)
+{
+
+  switch (database_type)
+    {
+    case GEOIP_LARGE_COUNTRY_EDITION:
+    case GEOIP_COUNTRY_EDITION:
+    case GEOIP_PROXY_EDITION:
+    case GEOIP_NETSPEED_EDITION:
+      return TRUE;
+
+    case GEOIP_CITY_EDITION_REV0:
+    case GEOIP_CITY_EDITION_REV1:
+      return FALSE;
+
+    default:
+      g_assert_not_reached();
+      return FALSE;
+    }
+}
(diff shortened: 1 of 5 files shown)


kira-syslogng:

modules/geoip/geoip-helper.c
```diff
@@ -0,0 +1,54 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include <glib.h>
+#include <GeoIP.h>
```
bazsi: don't include glib.h directly please, if there's no  local, "root" header for geoip, include "syslog-ng.h", that would pull in config.h and glib.h directly.

modules/geoip/geoip-helper.c
```diff
@@ -0,0 +1,54 @@
...
+
+#include <messages.h>
+
+gboolean
+is_country_type(GeoIPDBTypes database_type)
+{
```
bazsi: If this is an "is" function, it should not log its results as a side effect. e.g. do the logging in the caller.

modules/geoip/geoip-parser.c
```diff
@@ -24,24 +24,25 @@
...
+struct _GeoIPParser
 {
   LogParser super;
   GeoIP *gi;
 
   gchar *database;
   gchar *prefix;
-
+  void (*lookup)(GeoIPParser *, LogMessage *, const gchar *);
   struct
```
bazsi: I like that this has become a virtual function. "lookup" as a name is not the best one though, as it not only looks up (which suggests a read-only function without side effects), it actually changes the message, thus has side effects.

"process" as a term is a bit too generic and is already used by "LogParser", so another term would be better. I am out of really good ideas, but these come to my mind: process_record(), process_geoip(), lookup_and_set_values(), but I think you get the idea.

modules/geoip/geoip-helper.c
```diff
@@ -0,0 +1,54 @@
...
+
+  switch (database_type)
+    {
+    case GEOIP_LARGE_COUNTRY_EDITION:
+    case GEOIP_COUNTRY_EDITION:
+    case GEOIP_PROXY_EDITION:
+    case GEOIP_NETSPEED_EDITION:
+      msg_debug("geoip: country type database detected",
+                evt_tag_int("databaseType", database_type));
+      return TRUE;
```
bazsi: we are using underscory names in event tags, and not camel case ones.

kira-syslogng:

kira-syslogng:

modules/geoip/geoip-helper.c
furiel: Fixed.

PR #1464, opened 2017-05-06
python: retries support for python destination

furiel:
With this patch one can use retries() option for
python destination. For example:
destination d_python {
  python(class("ClassFoo") retries(3));
};
See:
https://github.com/balabit/syslog-ng/issues/1424

PR #1496, opened 2017-05-24
logthrdestdrv: regression: wrong error message when dropping message

furiel:
PR: https://github.com/balabit/syslog-ng/pull/1464
introduced a regression, when the worker instantly drops a message
(returns with WORKER_INSERT_RESULT_DROP),
the same error message is written as if it was dropped due to multiple
retries failure.

For example in case of invalid ttl setting for riemann destination:
like: ttl("string")
"2017-05-24T14:00:12.709410] Multiple failures while sending message to
destination, message dropped; driver='d_riemann#0', number_of_retries='3'"

This is changed for:
"[2017-05-24T14:30:56.081535] Message dropped while sending message
to destinaton; driver='d_riemann#0'"

### DIFF
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -157,6 +157,9 @@ log_threaded_dest_driver_do_insert(LogThrDestDriver *self)
       switch (result)
         {
         case WORKER_INSERT_RESULT_DROP:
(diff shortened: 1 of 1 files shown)


kira-syslogng:

PR #1498, opened 2017-05-29
Riemann destination: various bugfixes including heap corruption

furiel:
This patchset contains various bugfixes/improvements for Riemann destination.

A little comment for one of these issues: in case flush-lines is configured, messages can get stuck in the internal queue of Riemann destination. 

The optimistic path when Riemann server is connected is covered by this patchset. But the failure path when Riemann server is not connected is not covered. In the latter case syslog-ng will try to send batch once, but will not retry again, until the new messages arrived, or the batch filled suddenly completely, which latter case is again covered by this patch.
Also, for now in case flush-lines is used and unaccessible Riemann server, all events except the latest in the batch will be lost, because LogThrDestDriver do not support such message bulk rewind support. Only the last message can be rewind.

Any such issues do not occur when flush-lines is not used.

These issues will be resolved later in another changeset.

### DIFF
--- a/modules/riemann/riemann.c
+++ b/modules/riemann/riemann.c
@@ -560,6 +560,9 @@ riemann_worker_batch_flush(RiemannDestDriver *self)
   riemann_message_t *message;
   int r;
 
+  if (self->event.n == 0)
+    return WORKER_INSERT_RESULT_SUCCESS;
+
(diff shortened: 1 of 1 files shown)


kira-syslogng:

PR #1507, opened 2017-05-30
Remove diskq option from file destinations

gaborznagy:
This pull request is created to disallow the usage of diskq plugin in some destinations (e.g file and pipe destinations). See SYSLOGDEV-3479
For this the diskq module's context in the grammar has to be changed so it will not be generally available for all destination drivers.

Some aspects considered:
- modules which can use the diskq module should be able to use general destination plugins too.
- we want to keep the possibility to have general destination plugins for modules that cannot use diskq.

What has been done:
-   create new context for diskq in lexer and grammar
-  add new plugin_find interface to be able to search with multiple contexts (therefore modules can access to general destination plugins and to diskq module)
- restructured grammar of destination driver options

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -1159,25 +1159,7 @@ dest_driver_option
 
 	: KW_LOG_FIFO_SIZE '(' positive_integer ')'	{ ((LogDestDriver *) last_driver)->log_fifo_size = $3; }
 	| KW_THROTTLE '(' nonnegative_integer ')'         { ((LogDestDriver *) last_driver)->throttle = $3; }
-        | LL_IDENTIFIER
-          {
-            Plugin *p;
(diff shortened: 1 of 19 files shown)


kira-syslogng:

bazsi:
Hmm, pipe() should be able to use disk buffers, cant it?

kira-syslogng:

bazsi:
I think we agreed that this will not go in this form, at least until we checked if it is simpler to actually support disk queues even in the file destination. So I would close it and reopen at the point we have the solution.

bazsi:
Please open a new PR once there's a decision on how to move forward with this.

presidento:
The team decided to keep it open. Previously we had the "purgatory" label for PRs which have good parts, but we do not want to merge, at all. The team decided that the "blocked" label is enough in this case, this is why we kept it open.

PR #1357, opened 2017-03-02
Apply timezone configuration in Java destinations

presidento:
Previously the log messages were always in local timezone independently from the `time-zone()` configuration option.

# Issue

## Configuration

We have a simple file destination and the dummy Java destination, which just prints the received log message to the console. We expect to have the same message in both cases.

```
@version: 3.9

options {
  time_zone("-08:00");
};

source s_oneline { program("echo syslog-ng-test-oneline"); };

destination d_outfile {file(
    '/home/matefarkas/work/syslog-ng/install/out.txt'
    template("FILE / ${DATE} / ${ISODATE}")
);};


destination d_java {
  java(
    class_path("/home/matefarkas/work/syslog-ng/install/lib/syslog-ng/java-modules/*.jar")
    class_name("org.syslog_ng.DummyTextDestination")
    template("JAVA / ${DATE} / ${ISODATE}")
    option("name", "timezonetest")
  );
};

log {
  source(s_oneline);
  destination(d_java);
  destination(d_outfile);
};
```

## Output logs

```
FILE / Mar  2 02:35:55 / 2017-03-02T02:35:55-08:00
JAVA / Mar  2 11:35:55 / 2017-03-02T11:35:55+01:00;
```

It was originally [reported](https://gitter.im/balabit/syslog-ng/archives/2015/05/15) by @faxm0dem.

PR #1508, opened 2017-05-30
Set the default time-zone to UTC for elasticsearch2

czanik:
Elasticsearch and Kibana use UTC internally. Sending logs with UTC,
they end up in the right index and show up at the right time in
Kibana. Helps to avoid surprises and long debugging sessions :)

Suggested by Fabien Wernli on the syslog-ng mailing list.

Signed-off-by: Peter Czanik <peter@czanik.hu>

### DIFF
--- a/scl/elasticsearch/plugin.conf
+++ b/scl/elasticsearch/plugin.conf
@@ -76,6 +76,7 @@ block destination elasticsearch2(
   http_auth_type("none")
   http_auth_type_basic_username("")
   http_auth_type_basic_password("")
+  time_zone("UTC")
 )
 {
   java(
@@ -102,6 +103,7 @@ block destination elasticsearch2(
     option("http_auth_type", `http_auth_type`)
     option("http_auth_type_basic_username", `http_auth_type_basic_username`)
     option("http_auth_type_basic_password", `http_auth_type_basic_password`)
+    option("time_zone", `time_zone`)
     `__VARARGS__`
   );
 };
--- a/scl/elasticsearch/plugin.conf
+++ b/scl/elasticsearch/plugin.conf
@@ -103,7 +103,7 @@ block destination elasticsearch2(
     option("http_auth_type", `http_auth_type`)
     option("http_auth_type_basic_username", `http_auth_type_basic_username`)
     option("http_auth_type_basic_password", `http_auth_type_basic_password`)
-    option("time_zone", `time_zone`)
+    time_zone(`time_zone`)
     `__VARARGS__`
   );
 };


kira-syslogng:

presidento:
:+1: LGTM

bazsi:
I think a review by @lbudai or @pzoleex would be useful

pzoleex:
Hi,

Sorry I don't have syslog-ng code on my computer right now (so I cannot try it),  but I think this patch is partially wrong.

This just passes a "time_zone" option to elasticsearch code, but in that code there is nothing that would process this option.

If the general java destination handles the syslog-ng time_zone() option replace this:
-option("time_zone", `time_zone`)
+time_zone(`time_zone`)

If the java destination does not handle it, then this patch is not enough itself

faxm0dem:
@pzoleex is right, sorry for the oversight

bazsi:
@czanikp can you pls make these changes? Thanks.

presidento:
@pzoleex the Java destination handles it, thank you for noticing the error.

kira-syslogng:

presidento:
@czanik and @faxm0dem  as I know, there are 3 different time zones:

- the one used for example in file name macros in file destination (local timezone)
- the one used within the message templates, including the message timestamp (send timezone)
- and the original timezone

I am not sure which timezones are affected, and whether it is intended or not.

Only the `index` option is interesting (local timezone?), or it is not a problem, if the message contains everything in UTC?

I suppose that in current solution everything will be handled in UTC, and it is good for Elasticsearch, but I am not sure.

presidento:
In #1357 I have fixed in Java destinations that they should use the _send timezone_ (instead of _local timezone_) for formatting template strings, so I think it will be OK, but I am not sure.

faxm0dem:
yes it's probably for the best to use UTC for everything. the most important is for the `index` parameter.

bazsi:
@faxm0dem with the current patch index would be utc, the message would be the original timezone.

Our timestamp in the message does contain the timezone information when we send it to ES (e.g. it is an ISODATE)

If Kibana behaves properly in this combination, I would not change the message template, as we would be changing the timestamp of the message, not send it intact.

faxm0dem:
I agree: change as few things as possible to fix the issue!

FTR in a nutshell ES will happily accept ISODATE regardless of the TZ It is in (the default ES parsing includes the iso format)

bazsi:
Thanks @faxm0dem, this is good enough for merging now, so I'll do so. :)

PR #1354, opened 2017-02-21
Validate non negative numbers

presidento:
Add rules to grammar to be able to require non-negative and positive integers. Some of our configuration options expected those kind of values, but they were not validated before. (Just the program crashed due an invalid setting.)

(@MrAnno collected this list before, it was reviewed and splitted into the positive and the non-negative parts. Copied from https://github.com/balabit/syslog-ng/pull/1330#issuecomment-275225263)

## Positive numbers

* [listen-backlog()](https://www.balabit.com/documents/syslog-ng-ose-3.9-guides/en/syslog-ng-ose-guide-admin/html/reference-source-syslog-chapter.html#idp4343616)
* [log-fetch-limit()](https://www.balabit.com/documents/syslog-ng-ose-3.9-guides/en/syslog-ng-ose-guide-admin/html/reference-source-syslog-chapter.html#idp4358576)
* [log-iw-size()](https://www.balabit.com/documents/syslog-ng-ose-3.9-guides/en/syslog-ng-ose-guide-admin/html/reference-source-syslog-chapter.html#idp4368480)
* [max-connections()](https://www.balabit.com/documents/syslog-ng-ose-3.9-guides/en/syslog-ng-ose-guide-admin/html/reference-source-unixstream.html#idp5303904)
* [retries()](https://www.balabit.com/documents/syslog-ng-ose-3.9-guides/en/syslog-ng-ose-guide-admin/html/reference-destination-stomp.html#idp11689536)
* [log-msg-size()](https://www.balabit.com/documents/syslog-ng-ose-3.9-guides/en/syslog-ng-ose-guide-admin/html/reference-options.html#global-option-log-msg-size)
* [time-reap()](https://www.balabit.com/documents/syslog-ng-ose-3.9-guides/en/syslog-ng-ose-guide-admin/html/reference-options.html#global-option-time-reap)
* [time-reopen()](https://www.balabit.com/documents/syslog-ng-ose-3.9-guides/en/syslog-ng-ose-guide-admin/html/reference-options.html#global-option-time-reopen)
* [flush-lines()](https://www.balabit.com/documents/syslog-ng-ose-3.9-guides/en/syslog-ng-ose-guide-admin/html/reference-options.html#global-option-flush-lines)
* [flush-timeout()](https://www.balabit.com/documents/syslog-ng-ose-3.9-guides/en/syslog-ng-ose-guide-admin/html/reference-options.html#global-option-flush-timeout)
* stats-lifetime()
[...]

PR #1511, opened 2017-05-30
diskq: cover case disk-buf-size(0)

gaborznagy:
Grammar allows disk-buf-size to be 0.
However in that case syslog-ng exits on the first incoming message due to SEGFAULT.

The code detects that the disk-buf-size is not correct, but allows it.
"[2017-05-31T09:46:05.408827] WARNING: The configured disk buffer size is zero. No disk queue file will be created.;"

### DIFF
--- a/modules/diskq/diskq-options.c
+++ b/modules/diskq/diskq-options.c
@@ -43,11 +43,7 @@ disk_queue_options_qout_size_set(DiskQueueOptions *self, gint qout_size)
 void
 disk_queue_options_disk_buf_size_set(DiskQueueOptions *self, gint64 disk_buf_size)
 {
-  if (disk_buf_size == 0)
-    {
-      msg_warning("WARNING: The configured disk buffer size is zero. No disk queue file will be created.");
-    }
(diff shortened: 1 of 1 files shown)


kira-syslogng:

furiel:
The patch looks good to me.

Besides this change, do we want to explicitely deny zero value for disk buffer size?
In diskq-grammar.ym we have
KW_DISK_BUF_SIZE '(' nonnegative_integer ')'
which might be changed to
KW_DISK_BUF_SIZE '(' positive_integer ')'
Or maybe we could even deny all values below MIN_DISK_BUF_SIZE grammar-level.

gaborznagy:
Good point, a couple of months ago we made some general grammar changes regarding options with number value.
https://github.com/balabit/syslog-ng/pull/1354#pullrequestreview-25760667
I was thinking to re-review this step, but first we need to eliminate the above possibility in code as it causes SEGFAULT. It would be dead code anyway if we switch the grammar.

presidento:
:+1: LGTM. The documentation also should be checked.

PR #1355, opened 2017-02-23
Add SSL support for AMQP module

LarryWax:
This enable using SSL when using AMQP module.
Four config options were added:
- "**ssl_cacrt**" : path of the CA certificate in PEM format, setting this will enable SSL. Default value is an empty string. (SSL off)
-  "**ssl_key**" : path of private key in PEM format. This is effective only if "ssl_usercrt" is also set.
-  "**ssl_usercrt**" : path of user certificate in PEM format. This is effective only if "ssl_key" is also set.
-  "**ssl_verify**" : yes/no value, enable or disable peer verification. Default is enabled.
 
Signed-off-by: Giuseppe D'Anna <larrymadewax@gmail.com>

PR #1519, opened 2017-06-06
Add SSL support to AMQP module

LarryWax:
Hi, this reopen (https://github.com/balabit/syslog-ng/pull/1355) that I closed because I've done a mess with my branch :/

Anyway, I've updated params name's according to discussion with @presidento in the old PR  

Thank you for your time.

### DIFF
--- a/modules/afamqp/afamqp.c
+++ b/modules/afamqp/afamqp.c
@@ -68,10 +68,10 @@ typedef struct
   gint32 max_entries;
 
   /* SSL props */
-  gchar *ssl_cacrt;
-  gchar *ssl_key;
-  gchar *ssl_usercrt;
-  gboolean ssl_verify;
+  gchar *ca_file;
+  gchar *key_file;
+  gchar *cert_file;
+  gboolean peer_verify;
 } AMQPDestDriver;
 
 /*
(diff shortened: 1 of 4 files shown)


kira-syslogng:

presidento:
@LarryWax thank you for the rename!
@kira-syslogng ok to test.

kira-syslogng:

modules/afamqp/afamqp.c
```diff
@@ -342,8 +384,41 @@ afamqp_dd_connect(AMQPDestDriver *self, gboolean reconnect)
       msg_error("Error allocating AMQP connection.");
       goto exception_amqp_dd_connect_failed_init;
     }
+  if(strcmp(self->ca_file,"") != 0)
```
bazsi: it would be great if you could extract this large chunk of code to a separate function, one that is meant to create the amqp socket based on various settings.

Also, I don't really like ca_file being the trigger for setting up SSL. what if the user disables peer-verify (I know it's bad and why it is bad, but it happens), and they couldn't enable SSL in any way.

I have two ideas how to do this:

 * in the tcp() destination, we use a tls() block, where all the tls specific options are written. TLS is enabled with the presence of the block, so this enables TLS:

tcp('1.2.3.4' tls())

often, it looks like this:
tcp('1.2.3.4' tls(key-file('aaa.aaa')));

* another idea is to use a tls(yes/no) option

My preference goes to
1) tls block
2) tls(yes/no) option for enabling it specifically

But let me ask @fekete-robert from the Balabit documentation team if he has a better idea

modules/afamqp/afamqp.c
```diff
@@ -608,6 +683,10 @@ afamqp_dd_new(GlobalConfig *cfg)
 
   log_template_options_defaults(&self->template_options);
   afamqp_dd_set_value_pairs(&self->super.super.super, value_pairs_new_default(cfg));
+  afamqp_dd_set_ca_file((LogDriver *) self, "");
+  afamqp_dd_set_key_file((LogDriver *) self, "");
+  afamqp_dd_set_cert_file((LogDriver *) self, "");
+  afamqp_dd_set_peer_verify((LogDriver *) self, TRUE);
```
bazsi: I would rather use NULLs instead of empty values to indicate these options are unset.

bazsi: Also, the values for these pointers are not freed in amqp_dd_free()

modules/afamqp/afamqp.c
LarryWax: Thanks for your review. 
I agree with you regarding ca

LarryWax: Thank you for your review.
I agree with you about implementing another way to switch off/on ssl instead of using ca_file, I used this other approach because i didn't want to excessively alter the code. 
Let's wait documentation team then.

modules/afamqp/afamqp.c
LarryWax: Right, totally missed it

kira-syslogng:

kira-syslogng:

kira-syslogng:

MrAnno:
Oops, something went really wrong during the rebase. I'm going to open a new PR.

MrAnno:
Reopened (#1615).

PR #1523, opened 2017-06-08
Invalid read in system source/nvtable and in file destination

furiel:
This patchset contains fixes for two invalid read issues:
- System source: a value (program name) that needs to be added to nvtable is invalidated due to nvtable realloc.
- File destination: in case target file cannot be created, reopen_timer goes on infinitely, not stopped even after syslog-ng stops. Thus the scheduled timer has invalid cookie, thus an invalid read occurs during syslog-ng stop.

A little more verbose description can be found in commit messages.

### DIFF
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -1357,6 +1357,9 @@ log_writer_deinit(LogPipe *s)
   log_writer_stop_watches(self);
   iv_event_unregister(&self->queue_filled);
 
+  if (iv_timer_registered(&self->reopen_timer))
+    iv_timer_unregister(&self->reopen_timer);
+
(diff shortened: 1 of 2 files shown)


kira-syslogng:

kira-syslogng:

furiel:
I needed to update/force push the file destination part. The previous attempt planned to use the log_writer_stop_watches for deinitializing the time_reopen function as well, but it seems it is not desinged to do that. As I understand log_writer_stop_watches (together with log_writer_start_watches) is only planned to be used to execute reopen/suspend logic. In the end I reverted my change on self->watches_running, and called the iv_timer_unregister for reopen_timer in the deinit function instead.

PR #1524, opened 2017-06-08
gprocess: fix oracle-8-jvm related issue

lbudai:
The java8 oracle implementation puts two additional variables to
the environ(it is reallocated by libc). It caused crash in java
destination during our testing process.

The fix: track how many items we put to the environ array,
and free only those items. This works as libc use memcpy when realloc
the environ array and the new items appeded to the end of the
reallocated array.

### DIFF
--- a/lib/gprocess.c
+++ b/lib/gprocess.c
@@ -104,6 +104,10 @@ static gboolean stderr_present = TRUE;
 static int have_capsyslog = FALSE;
 #endif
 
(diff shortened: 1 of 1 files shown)


kira-syslogng:

bazsi:
Ugh, this seems like a problem that was difficult to diagnose. Well done.

pzoleex:
ok this fix solves the crash (I tried it where the crash happened originally)

PR #1528, opened 2017-06-12
g_process_finish: reverting attempt to fix memleak

furiel:
The reason why environ is copied during g_process_set_argv_space so to be
able to overwrite process title in ps/top commands to
supervisor. In bsd there is setproctitle call which solve this,
but currently it is not available for linux.

The problem is that modules can add their own env variables to the
list, which must not be freed here, otherwise it would result
double free (e.g some version of Oracle Java). One might also would
try to track which environment variables are added by syslog-ng,
and free only those. There is still a problem with this, **environ
itself cannot be freed in some cases. For example libcurl registers
an at_exit function which needs an environment variable, that would
be freed here before at_exit is called, resulting in invalid read.

As this leak does not cause any real problem like accumulating over
time, it is safe to leave it as it is.

### DIFF
--- a/lib/gprocess.c
+++ b/lib/gprocess.c
@@ -104,10 +104,6 @@ static gboolean stderr_present = TRUE;
 static int have_capsyslog = FALSE;
 #endif
 
-#if SYSLOG_NG_HAVE_ENVIRON
-static gint env_ctr;
-#endif
-
 /* global variables */
 static struct
 {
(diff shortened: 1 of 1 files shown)


lib/gprocess.c
```diff
@@ -1446,13 +1446,25 @@ void
...
 #ifdef SYSLOG_NG_HAVE_ENVIRON
-  for (gint i = 0; i < env_ctr && environ[i] != NULL; i++)
-    g_free(environ[i]);
-
-  g_free(environ);
-
-  if (process_opts.argv_orig)
-    free(process_opts.argv_orig);
+  /**
+   * There were various attempts to eliminate this leak. The reason
```
presidento: The _"to eliminate this leak"_ expression is not self-explaining without the context of this patch.

I would rather use the _"to free  the `environ` variable"_ expression.

kira-syslogng:

PR #1485, opened 2017-05-19
Add debun script and doc to release tgz

czanik:
Closes: https://github.com/balabit/syslog-ng/issues/1415

Signed-off-by: Peter Czanik <peter@czanik.hu>

PR #1529, opened 2017-06-12
F/syslog debun extensions

jszigetvari:
contrib/syslog-debun: PKI information gathering and general improvements

    * PKI information gathering
    * extended scope of information gathered
    * bug fixes

### DIFF
--- a/contrib/syslog-debun
+++ b/contrib/syslog-debun
@@ -439,7 +439,7 @@ acquire_syslog_info () {
 }
 
 acquire_syslog_var () {
-	$duks $vardir >${tmpdir}/syslog.duks.var
+	$duks $vardir/ >${tmpdir}/syslog.duks.var
 	read vardu dir <${tmpdir}/syslog.duks.var
 	mkdir ${tmpdir}/var
 	cd "$vardir"
--- a/contrib/syslog-debun
+++ b/contrib/syslog-debun
@@ -48,7 +48,7 @@ dfk="df -k"
 dfh="df -h"
 dfi="df -i"
 duks="du -ks"
-grepq="grep -q"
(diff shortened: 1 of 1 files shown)


kira-syslogng:

MrAnno:
@kira-syslogng ok to test

kira-syslogng:

bazsi:
Great stuff here. Some notes.

I think it would make sense to request debug output to ISSUE_TEMPLATE.md so
that it is requested from those who open issues.

btw: is the debun script part of the installation image right now? I've
just checked and it isn't. So it should be moved out of contrib (maybe in
scripts/?) and then added to the installation packages, so it is really
easy for users to run. Right now, one would have to download it from github
explicitly. Not many people would do that.

also, the name of the script should be changed, it is "syslog-debun",
whereas we are shipping syslog-ng as a package, so it should be
syslog-ng-debun or something like that. Also, "debun" is not really
intuitive for first time users and it would probably make sense to display
a help screen when launching the script instead of going right into
information collection.

Cheers,
Bazsi

-- 
Bazsi

czanik:
It is: see https://github.com/balabit/syslog-ng/pull/1485

A better name is very welcome. For now I put it next to documentation but with a better name it could be installed next to syslog-ng to the sbin directory.

bazsi:
#1485 doesnt add this into the binaries only the source, so it won't be
available to most of the users.

jszigetvari:
If that's okay for you, I will change the name in the next bunch of changes I'll make. (ETA: a few month's time.)
I just did not want to remove the legacy of @pasztor on my own, because he is the one who originally authored and added this script.

Issue #1423, opened 2017-04-10
Custom template for remote syslog temporarily ignored when HUP is received


epontan:
# syslog-ng

Docker image balabit/syslog-ng:latest

## Version of syslog-ng

```
$ docker exec -it syslog-ng syslog-ng --version
syslog-ng 3.9.1
Installer-Version: 3.9.1
Revision: 3.9.1-2
Module-Directory: /usr/lib/syslog-ng/3.9
Module-Path: /usr/lib/syslog-ng/3.9
Available-Modules: confgen,csvparser,add-contextual-data,afprog,mod-java,affile,geoip-plugin,date,pseudofile,syslogformat,json-plugin,basicfuncs,afuser,linux-kmsg-format,afstomp,afamqp,afsocket,kvformat,cef,cryptofuncs,disk-buffer,graphite,afmongodb,dbparser,sdjournal,system-source
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
```

## Platform

```
$ cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
ID_LIKE=archlinux
ANSI_COLOR="0;36"
HOME_URL="https://www.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
```

# Issue

## Failure

In systems where newsyslog is used to rotate log files that are written to disk by syslog-ng, it has been observed that sometimes when the HUP signal (that is sent to syslog-ng process by newsyslog) causes wrong template to be used temporarily for a remote syslog server that has been configured to use a custom template.

I suspect that there is some kind of race-condition when reloading the configuration files.

It can be reproduced quite easily using docker, with a simple syslog server in python acting as remote (catching the wrong template), a python script that generates frequent logging and shell script that sends regular HUP signals.

The custom template is defined without the program name:
```
<$PRI>$DATE $HOST : $MSG
```

Sometimes when the HUP signal is received, this format is sent to the remote server:
```
<190>Apr 10 09:45:29 0c30188d28c5 : myapp: Writing some log data 0
```

Where the expected format is to follow the custom template:
```
<190>Apr 10 09:45:29 0c30188d28c5 : Writing some log data 0
```

## Steps to reproduce

[...]

PR #1530, opened 2017-06-13
affile: Add possibility to only signal re-open of file handles

epontan:
I'm not sure it is feasible to solve issue #1423 in a good way without any penalty in performance given if it is related to race between multiple threads (which I suspect is the problem). So I tackled this from another point of view by avoiding a reload the configuration all together as it is not needed from the problem scenario stated in the issue and adding a new signal handler to only trigger a re-opening of log files instead.

### DIFF
--- a/modules/affile/affile-dest.c
+++ b/modules/affile/affile-dest.c
@@ -35,6 +35,7 @@
 #include "logwriter.h"
 #include "affile-dest-internal-queue-filter.h"
 #include "file-specializations.h"
+#include "apphook.h"
 
 #include <iv.h>
 #include <sys/types.h>
@@ -82,6 +83,8 @@
  * protection of the lock, keeping a the next pipe alive, even if that would
  * go away in a parallel reaper process.
  */
+
+static GList *affile_dest_drivers = NULL;
 
 struct _AFFileDestWriter
 {
@@ -179,7 +182,8 @@ affile_dw_reopen(AFFileDestWriter *self)
       proto = file_opener_construct_dst_proto(self->owner->file_opener, transport,
                                               &self->owner->writer_options.proto_options.super);
 
-      main_loop_call((void *(*)(void *)) affile_dw_arm_reaper, self, TRUE);
+      if (!iv_timer_registered(&self->reap_timer))
+        main_loop_call((void *(*)(void *)) affile_dw_arm_reaper, self, TRUE);
     }
   else
     {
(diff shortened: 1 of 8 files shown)


kira-syslogng:

presidento:
@kira-syslogng ok to test

kira-syslogng:

bazsi:
I like the approach, I would have to review the implementation though.

Also, I think in addition to SIGUSR1 we should support the same via
syslog-ng-ctl.

kira-syslogng:

epontan:
Thanks! I added a reopen command to syslog-ng-ctl as well.

kira-syslogng:

kira-syslogng:

epontan:
Any idea when this could be looked upon? Thanks in advance.

kira-syslogng:

epontan:
Thank you for the review and your comments @furiel ! 
I agree with them and have pushed a new commit that is hopefully to your satisfaction.

I ended up adding and calling the global init function from the affile-plugin.c instead of in apphook.c as it felt wrong to have apphook directly depend on a plugin module (it even caused problem compiling for some other module like pdbtool when linking libsyslog-ng.so).

Also, sorry for the rebase. Will do a squash once the code is approved.

kira-syslogng:

epontan:
Thank you :)

I have squashed the commit and is ready for merge whenever you are.

bazsi:
\o/

This is a feature I've wanted for a long time :) We should try to make an
effort to use this in logrotate scripts in distros. @mochrul @czanik

-- 
Bazsi

czanik:
Let me check.
* In openSUSE / SLES the logrotate file is shared with rsyslog. It even belongs to a separate package, a common dependency of both syslog implementations.
* On fedora / RHEL the file is part of the syslog-ng package, but if it differs from the rsyslog version it might cause pain: sometimes rsyslog is a hard dependency of another package, so it can not be deleted just disabled. At least RHEL7 is already clever enough that if the logrotate files are the same then the two packages don't conflict...

modules/affile/affile-dest.c
```diff
@@ -354,6 +358,29 @@ affile_dw_new(const gchar *filename, GlobalConfig *cfg)
...
+  affile_dw_reopen(writer);
+}
+
+static void
+affile_dd_reopen_all_writers(gpointer data, gpointer user_data)
+{
+  AFFileDestDriver *driver = (AFFileDestDriver *) data;
+  if (driver->single_writer)
+    affile_dw_reopen(driver->single_writer);
+  else
```
epontan: I am sorry, but I think I might have done an incorrect assumption in this logic. That a destination file driver must either have single_writer or writer_hash (in case destination is a template) set.

I was just testing this patch again with a more complex configuration and it seems that sometimes both entries could be NULL as I got this message: `g_hash_table_foreach: assertion 'hash_table != NULL' failed`

In what configuration scenario would this happen and what is the proper logic? An `else if` or just another `if` (i.e. could both be unset or set)? @bazsi @furiel

furiel: As I see the construction of single_writer or writer_hash is deferred until a file needs to be opened. And opening a file is deferred until there is some message to write.
Probably when syslog-ng starts and there was no message to handle, both these objects can be null.

epontan: Ah yes, that makes sense. So an `else if` on the `writer_hash` should account for this.  
I'll create an issue of it and another pull request to solve it. Thanks!

Issue #1506, opened 2017-05-30
SigAbort after reload and stop


mitzkia:
syslog-ng version:
 * builded from actual OSE master 
 * latest git commit: 282ec7c521918d79a6a01189f0c7d8d0bc9ac26c

Reproduction:
1, Create directory for ssl server keys and ca : `mkdir ~/http_server`
2, Create directory for ssl client keys: `mkdir ~/http_client`
3, Copy the following files to ~/http_server dir (remove the last ".jpg" extension from files):

    * ca-crt.pem  
![ca-crt pem](https://cloud.githubusercontent.com/assets/2952076/26579023/79a2f7ea-4532-11e7-81af-3a3d347502fd.jpg)

    * ca-key.pem   
![ca-key pem](https://cloud.githubusercontent.com/assets/2952076/26579102/bd44c3de-4532-11e7-8d69-4ab2af77b687.jpg)

    * server-crt.pem   
![server-crt pem](https://cloud.githubusercontent.com/assets/2952076/26579112/c21d9872-4532-11e7-844a-a3df55f05f3e.jpg)

    * server-key.pem  
![server-key pem](https://cloud.githubusercontent.com/assets/2952076/26579121/c7572646-4532-11e7-9b60-f489e6635865.jpg)

4, Copy the following files to ~/http_client dir

    * client1-crt.pem  
![client1-crt pem](https://cloud.githubusercontent.com/assets/2952076/26579174/fc7a81ce-4532-11e7-9703-54a67e7b8236.jpg)

    * client1-key.pem 
![client1-key pem](https://cloud.githubusercontent.com/assets/2952076/26579178/017560fe-4533-11e7-84fb-f99c2eb50c78.jpg)

5, Start openssl server from ~/http_server dir: `openssl s_server -key server-key.pem -cert server-crt.pem -accept 4433 -www`

6, Start syslog-ng with the following (valid) config (replace [username] with your name):
```
@version: 3.9

options {
    time_reopen(100);
    threaded(yes);
};

source s_network {
    network( ip("127.0.0.1") port(10001));
};

destination d_http {
    http(
        method(post)
        url("https://localhost:4433")
        ca_file("/home/[username]/http_server/ca-crt.pem")
        ca_dir("/home/[username]/http_server/")
        cert_file("/home/[username]/http_client/client1-crt.pem")
        key_file("/home/[username]/http_client/client1-key.pem")
    );
};

log {
    source(s_network);
    destination(d_http);
};
```

[...]

PR #1537, opened 2017-06-16
MainLoop: crash due to async calls might be held back without limit

furiel:
Reload and stop commands are implemented as an async calls. These async calls only executed when there are no workers running.

In case of slow workers or high load, workers might not be exited for a long time, possibly infinitely in case sources are faster than the destinations. This results in such async calls held back.

The original implementation g_asserts that only one async call is running at a given time, though it is possible in the scenarios above that the 100ms is not enough for the previous async call to exit, resulting in crash.
See: https://github.com/balabit/syslog-ng/issues/1506

This patch collects the commands to be executed in a queue. When all workers exited, we iterate through the queue and call all such actions.

This fixes the crash part.

But certainly it is also an issue that workers may not exit for a long period of time. The fact that workers need to exit is registered inside MainLoop. This patch adds a check for this flag in logthrdestdriver, when messages are fetched from the queues.

Test/Reproduce:
I could test/reproduce these issues by the following way:
One needs an nginx with echo module installed. (On ubuntu16.04, nginx-full package contains echo module, while the standard nginx package does not).
Add the following location to nginx:
```
	location = /slow-reply {
	   echo_sleep 5.0;
	   #return 200 'this response would NOT be delayed!';
	   echo 'Slooow motion';
	}
````
This will ensure nginx will send responses only after 5 seconds, so one have time to play around syslog-ng.
Inside syslog-ng config:
`destination d_https { http(url("http://localhost:80/slow-reply"));};`

Push some messages to the destination: (I have a network source with port 5555)
`loggen -n 10 -s 130 -i 127.0.0.1 5555`

Then try to stop syslog-ng: ctrl-c or /syslog-ng-ctl stop
After this patch syslog-ng exits after 5 seconds, otherwise 50 seconds.

### DIFF
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -52,7 +52,8 @@ static void
 log_threaded_dest_driver_message_became_available_in_the_queue(gpointer user_data)
 {
   LogThrDestDriver *self = (LogThrDestDriver *) user_data;
-  iv_event_post(&self->wake_up_event);
+  if (!self->under_termination)
+    iv_event_post(&self->wake_up_event);
 }
 
 static void
@@ -146,7 +147,8 @@ log_threaded_dest_driver_do_insert(LogThrDestDriver *self)
   worker_insert_result_t result;
   LogPathOptions path_options = LOG_PATH_OPTIONS_INIT;
 
-  while (!self->suspended &&
+  while (G_LIKELY(!self->under_termination) &&
+         !self->suspended &&
          (msg = log_queue_pop_head(self->queue, &path_options)) != NULL)
     {
       msg_set_context(msg);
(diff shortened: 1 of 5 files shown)


kira-syslogng:

kira-syslogng:

mitzkia:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

furiel:
@bazsi Could you accept the requested changes and approve the PR? We would like to include this into the next release.

lbudai:
@bazsi: I think it is ready to merge,so I'm merge this (as I want to include it in the OSE 3.11).

gaborznagy:

Issue #1500, opened 2017-05-29
http() destination: time_reopen() option not activated


mitzkia:
syslog-ng version:
 * builded from actual OSE master 
 * latest git commit: 282ec7c521918d79a6a01189f0c7d8d0bc9ac26c

Reproduction:
1, Use the following (valid) syslog-ng config (http server did not listening):
```
@version: 3.9

options {
    time_reopen(1);
};

source s_network {
    network( ip("127.0.0.1") port(10001));
};

destination d_http {
    http(
        method(put)
        url("http://127.0.0.1:8080")
    );
};


log {
    source(s_network);
    destination(d_http);
};
```
2, syslog-ng should send an error message to console log when time_reopen() is activated, but it is missing:
```
$ sbin/syslog-ng -Fe -f ~/micek_work/syslog_ng_configs/syslog_ng_http_3.conf 
syslog-ng: Error setting capabilities, capability management disabled; error='Operation not permitted'
[2017-05-29T20:14:52.651916] WARNING: Configuration file format is newer than the current version, please specify the current version number (3.9) in the @version directive. syslog-ng will operate at its highest supported version in this mode;
[2017-05-29T20:14:52.676323] syslog-ng starting up; version='3.9.1'
^C[2017-05-29T20:14:55.898044] syslog-ng shutting down; version='3.9.1'

```
* similar message from network destination
```
[2017-05-29T20:11:28.882633] Syslog connection failed; fd='13', server='AF_INET(127.0.0.1:9999)', error='Connection refused (111)', time_reopen='10'
```

mitzkia:
Some further notes. 
If we start syslog-ng with an active http destination, and send 1 message to syslog-ng it will transfer it to http destination correctly. But if the http destination goes down, and syslog-ng gets new messages the  time_reopen() will not activated, but the retries() yes. So messages will be dropped automatically.

```
[2017-05-31T09:53:12.247531] Syslog connection accepted; fd='15', client='AF_INET(127.0.0.1:38376)', local='AF_INET(127.0.0.1:10001)'
[...]

PR #1540, opened 2017-06-16
http destination: improvements

furiel:
- Memory leak during reload/stop
- Invalid handling of retry-send when remote is down.
See: https://github.com/balabit/syslog-ng/issues/1500
- Replacing some loops with foreach
- Using scratchbuffers to store body of the http request
- Curl options are set only once during init
- Timeout is configurable for http request
- HTTP responses handled correcty: 2XX ack, 4XX drop, 5XX retry.

Please wait @mitzkia or @pzoleex with the merge.

### DIFF
--- a/modules/http/http-grammar.ym
+++ b/modules/http/http-grammar.ym
@@ -63,6 +63,7 @@
 %token KW_CIPHER_SUITE
 %token KW_SSL_VERSION
 %token KW_PEER_VERIFY
+%token KW_TIMEOUT
 
 %type   <ptr> driver
 %type   <ptr> http_destination
@@ -113,6 +114,7 @@ http_option
     | KW_CIPHER_SUITE '(' string ')'          { http_dd_set_cipher_suite(last_driver, $3); free($3); }
     | KW_SSL_VERSION '(' string ')'           { http_dd_set_ssl_version(last_driver, $3); free($3); }
     | KW_PEER_VERIFY '(' yesno ')'            { http_dd_set_peer_verify(last_driver, $3); }
+    | KW_TIMEOUT '(' nonnegative_integer ')'  { http_dd_set_timeout(last_driver, $3); }
     | dest_driver_option
     | threaded_dest_driver_option
     | { last_template_options = http_dd_get_template_options(last_driver); } template_option
--- a/modules/http/http-grammar.ym
+++ b/modules/http/http-grammar.ym
@@ -103,7 +103,7 @@ http_option
     | KW_USER       '(' string ')'            { http_dd_set_user(last_driver, $3); free($3); }
     | KW_PASSWORD   '(' string ')'            { http_dd_set_password(last_driver, $3); free($3); }
     | KW_USER_AGENT '(' string ')'            { http_dd_set_user_agent(last_driver, $3); free($3); }
-    | KW_HEADERS    '(' string_list ')'       { http_dd_set_headers(last_driver, $3); g_list_free($3); }
+    | KW_HEADERS    '(' string_list ')'       { http_dd_set_headers(last_driver, $3); g_list_free_full($3, free); }
     | KW_METHOD     '(' string ')'            { http_dd_set_method(last_driver, $3); free($3); }
     | KW_BODY       '(' template_content ')'  { http_dd_set_body(last_driver, $3); log_template_unref($3); }
     | KW_CA_DIR     '(' string ')'            { http_dd_set_ca_dir(last_driver, $3); free($3); }
@@ -121,4 +121,3 @@ http_option
 /* INCLUDE_RULES */
 
 %%
-
--- a/modules/http/http-plugin.h
+++ b/modules/http/http-plugin.h
@@ -46,6 +46,7 @@ typedef struct
   int ssl_version;
   gboolean peer_verify;
   short int method_type;
+  glong timeout;
   LogTemplate *body_template;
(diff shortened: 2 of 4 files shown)


kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

presidento:
@kira-syslogng retest this please.

kira-syslogng:

modules/http/http-grammar.ym
```diff
@@ -103,7 +104,7 @@ http_option
     | KW_USER       '(' string ')'            { http_dd_set_user(last_driver, $3); free($3); }
     | KW_PASSWORD   '(' string ')'            { http_dd_set_password(last_driver, $3); free($3); }
     | KW_USER_AGENT '(' string ')'            { http_dd_set_user_agent(last_driver, $3); free($3); }
-    | KW_HEADERS    '(' string_list ')'       { http_dd_set_headers(last_driver, $3); g_list_free($3); }
+    | KW_HEADERS    '(' string_list ')'       { http_dd_set_headers(last_driver, $3); g_list_free_full($3, g_free); }
     | KW_METHOD     '(' string ')'            { http_dd_set_method(last_driver, $3); free($3); }
```
bazsi: unfortunately the lexer does not use g_strdup() but rather strdup() (e.g. without the g_) to allocate strings, and on win32 g_strdup and strdup are not the same :(

so, please either use free() as the destroy notify callback, or fix the lexer. I've always used the first solution in the past, but it's getting really annoying.

modules/http/http.c
```diff
@@ -173,46 +175,65 @@ _set_curl_opt(HTTPDestinationDriver *self, LogMessage *msg, struct curl_slist *c
...
-  if (body_rendered)
-    g_string_free(body_rendered, TRUE);
+  glong http_code = 0;
+  curl_easy_getinfo (self->curl, CURLINFO_RESPONSE_CODE, &http_code);
+  switch (http_code/100)
+    {
+    case 4:
+      msg_debug("curl: 4XX: msg dropped",
+                evt_tag_int("status code", http_code));
+      retval = WORKER_INSERT_RESULT_DROP;
```
bazsi: tags should never contain spaces..., don't forget this will become key-values in a log messages.

modules/http/http.c
```diff
@@ -413,6 +434,13 @@ http_dd_set_peer_verify(LogDriver *d, gboolean verify)
   self->peer_verify = verify;
 }
 
+void http_dd_set_timeout(LogDriver *d, gulong timeout)
+{
```
bazsi: break the return type to its own line please to match conventions.

modules/http/http.c
```diff
@@ -424,11 +452,19 @@ http_dd_init(LogPipe *s)
 
   log_template_options_init(&self->template_options, cfg);
 
+  if (!(self->curl = curl_easy_init()))
+    {
+      msg_error("curl: cannot initialize libcurl", NULL);
+      return FALSE;
```
bazsi: hmm, we need to improve this (and other similar) error messages. You can get the location information about which source we are reporting the issues about with log_pipe_get_location_tag(), reporting the error message returned by libcurl will also help troubleshooting this message.

modules/http/http-plugin.h
```diff
@@ -46,6 +46,7 @@ typedef struct
   int ssl_version;
   gboolean peer_verify;
   short int method_type;
+  gulong timeout;
   LogTemplate *body_template;
```
bazsi: gint should suffice here I think.

modules/http/http.c
```diff
@@ -173,46 +175,65 @@ _set_curl_opt(HTTPDestinationDriver *self, LogMessage *msg, struct curl_slist *c
...
-      return WORKER_INSERT_RESULT_ERROR;
+      return WORKER_INSERT_RESULT_NOT_CONNECTED;
     }
 
-  if (body_rendered)
-    g_string_free(body_rendered, TRUE);
+  glong http_code = 0;
+  curl_easy_getinfo (self->curl, CURLINFO_RESPONSE_CODE, &http_code);
+  switch (http_code/100)
+    {
```
bazsi: I think you should extract the error handling code to a function of its own (e.g. _map_http_status_to_worker_status)

modules/http/http-plugin.h
furiel: In the libcurl documentation I saw long:
https://curl.haxx.se/libcurl/c/CURLOPT_TIMEOUT.html
CURLcode curl_easy_setopt(CURL *handle, CURLOPT_TIMEOUT, long timeout),
that's why I used gulong. I will change this to glong from gulong, but I would leave a long-type here to align to the curl documentation.

modules/http/http.c
furiel: It seems curl_easy_init cannot report back the reason of the error:
https://curl.haxx.se/mail/lib-2009-11/0243.html,
so unfortunalely we cannot give curl-specific hint of the error.

I will insert log_pipe_location_tag() here and into the _insert method when curl_easy_perform is called as you suggest.

kira-syslogng:

kira-syslogng:

mitzkia:
We should enable "retries" functionality if http server returns with 5xx error code.
We should retry as retries([counter]) times.

kira-syslogng:

PR #1539, opened 2017-06-16
diskq: set the type

kvch:
Signed-off-by: Noémi Ványi <sitbackandwait@gmail.com>

PR #1541, opened 2017-06-16
diskq: remove an always TRUE condition in _acquire_queue

MrAnno:
Previously, `queue->type` was not initialized correctly at all (fixed in #1539), so `_acquire_queue` created a new queue in every case.

NOTE: The removed logic is reasonable and valuable, so we should consider reverting this patch after log_queue_disk_load_queue() is fixed.


kira-syslogng:

PR #1544, opened 2017-06-16
diskq: fix crash when diskq is used with file destionation and file is reaped

gaborznagy:
We allow using disk-buffer for file sources, nevertheless syslog-ng crashes.
It is crashing when the file destination is reaping and log_writer_flush is called during deinit.

The reason is that during reaping we deinitialized the diskq and then called log_writer_flush which tries to access the diskq, but its already a NULL pointer.

This patch checks if the diskq is deinitialized before trying to access the diskq.

### DIFF
--- a/modules/affile/affile-dest.c
+++ b/modules/affile/affile-dest.c
@@ -427,8 +427,9 @@ affile_dd_reap_writer(AFFileDestDriver *self, AFFileDestWriter *dw)
       g_static_mutex_unlock(&self->lock);
     }
 
-  log_dest_driver_release_queue(&self->super, log_writer_get_queue(writer));
+  LogQueue *queue = log_writer_get_queue(writer);
   log_pipe_deinit(&dw->super);
+  log_dest_driver_release_queue(&self->super, queue);
   log_pipe_unref(&dw->super);
 }
 


kira-syslogng:

MrAnno:
The root cause of the problem is in  `affile_dd_reap_writer`:
```
log_dest_driver_release_queue(&self->super, log_writer_get_queue(writer));
log_pipe_deinit(&dw->super);
```

log_pipe_deinit() calls `log_writer_flush()` on the previously released queue, which causes a segmentation fault.

I think your fix is a workaround, we need to investigate the problem on the DestDriver/release_queue side.

bazsi:
I agree. The disk based queue should probably be deinitialized by affile_dw_deinit() and not outside of it.

Actually affile_dw_init() acquires the queue, so the deinitialization should happen in deinit also to be symmetric.

MrAnno:
So we've found the following:
- the `release_queue`, `acquire_queue` implementation is correct, everything is working in case of `LogQueueFifo`
- there is a tricky part in the disk queue version of the `release_queue`, `acquire_queue`:
  - `release_queue` calls qdisk's `_save_queue`, and the qdisk instance will be deinitialized (`qdisk_deinit()`)
  - at the same time, the corresponding `LogQueue` instance will be saved in persist config with the deinitialized qdisk
  - `_acquire_queue` does a trick: it always creates a NEW queue and drops the previously saved queue instance (everything is on the disk, so this is "safe")

Because of this "trick", the queue stays in an deinitialized state when `log_writer_flush()` tries to access the queue in `log_writer_deinit()`.

kira-syslogng:

kira-syslogng:

furiel:
We decided to accept the fix in offline review.

PR #1548, opened 2017-06-19
Fix memory usage counter bug #2: disk-buffer, overflow buffer and reload

gaborznagy:
We have found an error around memory usage counter when non-reliable disk-buffer is used and syslog-ng is reloaded.

We have found that the problem is around the overflow buffer which was not re-initialized after reload.
It caused an underflow in memory usage counter when syslog-ng can successfully send the messages from the disk-buffer.

Using the same solution that was done for qout buffer.

### DIFF
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -32,7 +32,7 @@ typedef struct
 {
   guint index_in_queue;
   guint item_number_per_message;
-  gssize *memory_usage_initial_value;
+  gssize *value_accumulator;
 } DiskqMemusageLoaderState;
 
 static gboolean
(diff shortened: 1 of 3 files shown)


kira-syslogng:

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -61,12 +61,18 @@ _start(LogQueueDisk *s, const gchar *filename)
...
-
-  g_queue_foreach(self->qout, _update_memory_usage_during_load, &state);
+  DiskqMemusageLoaderState qout_sum = { .index_in_queue = 0,
+                                        .item_number_per_message = ITEM_NUMBER_PER_MESSAGE,
+                                        .memory_usage_initial_value = &self->super.super.memory_usage_qout_initial_value
+                                      };
+
+  DiskqMemusageLoaderState overflow_sum = { .index_in_queue = 0,
+                                            .item_number_per_message = ITEM_NUMBER_PER_MESSAGE,
+                                            .memory_usage_initial_value = &self->super.super.memory_usage_overflow_initial_value
```
furiel: DiskqMemusageLoaderState struct has this field: memory_usage_initial_value which turned out it needs to be generalized. Could you please rename it accordingly? Something like: .accumulator or similar?

gaborznagy: Can do :+1:

kira-syslogng:

PR #1552, opened 2017-06-20
logsource: add "source suspended/resumed" debug messages

MrAnno:
This information can be useful when a user wants to find out why syslog-ng doesn't receive messages.
@jszigetvari 

Note that these are just debug messages; if you want to measure the "availability of sources", an `iw-size` related stat counter would be a better choice.

### DIFF
--- a/lib/logsource.c
+++ b/lib/logsource.c
@@ -41,7 +41,7 @@ log_source_wakeup(LogSource *self)
   if (self->wakeup)
     self->wakeup(self);
 
-  msg_debug("Source has been resumed", evt_tag_str("group_name", self->options->group_name));
+  msg_debug("Source has been resumed", log_pipe_location_tag(&self->super));
 }
 
 static inline void
@@ -143,7 +143,7 @@ log_source_msg_ack(LogMessage *msg, AckType ack_type)
 void
 log_source_flow_control_suspend(LogSource *self)
 {
-  msg_debug("Source has been suspended", evt_tag_str("group_name", self->options->group_name));
+  msg_debug("Source has been suspended", log_pipe_location_tag(&self->super));
 
   g_atomic_counter_set(&self->suspended_window_size, g_atomic_counter_get(&self->window_size));
   g_atomic_counter_set(&self->window_size, 0);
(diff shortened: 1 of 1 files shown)


kira-syslogng:

kira-syslogng:

lib/logsource.c
```diff
@@ -52,7 +52,10 @@ _flow_control_window_size_adjust(LogSource *self, guint32 window_size_increment)
   g_atomic_counter_set(&self->suspended_window_size, 0);
 
   if (old_window_size == 0)
-    log_source_wakeup(self);
+    {
+      msg_debug("Source has been resumed", evt_tag_str("group_name", self->options->group_name));
+      log_source_wakeup(self);
```
bazsi: Instead of the group name, either the location or the id should be logged. Also, wouldnt it make sense to add this log message to the resume function?

lib/logsource.c
```diff
@@ -242,6 +247,9 @@ log_source_post(LogSource *self, LogMessage *msg)
 
   old_window_size = g_atomic_counter_exchange_and_add(&self->window_size, -1);
 
+  if (G_UNLIKELY(g_atomic_counter_get(&self->window_size) == 0))
+    msg_debug("Source has been suspended", evt_tag_str("group_name", self->options->group_name));
```
bazsi: This atomic operation is in the fast path and is racy. We should check if old_window_size is 1, that avoids both the race and the atomic.

I cant see the context of this change, but wouldnt this be covered by the message in log_source_flow_control_suspend()?

lib/logsource.c
MrAnno: Right, thank you.

lib/logsource.c
MrAnno: Oh I see, I'll use the `old_window_size` in this case.

AFAIK `log_source_flow_control_suspend()` is only called by the the ack tracker when `LogQueue` drops a message with `AT_SUSPENDED` (when the queue is full).

Usually, the `window_size` reaches 0 before the queue is completely filled, this is why I placed this
debug message here.

kira-syslogng:

kira-syslogng:

MrAnno:
Oh oops, I didn't squash my fixup commits.

bazsi:
Yeah, I should have checked them. But nevermind, this happens. I would not info the merge, especially as i merged other things as well.

PR #1554, opened 2017-06-21
LogQueueFifo memory usage underflow and cfg object corruption during reload

furiel:
During reload, memory_usage counter is reset, but the queue might be maintained, for example when it contains messages. This leads counter value inconsistency during reload. After opening the destination, memory_usage will underflow.

To prevent this, memory_usage counter is saved to the persistent storage inside GlobalConfig, so that its value can be reloaded together with the queue.

During the implementation we run into another bug: after reload the reference to the cfg object of writer is corrupted, because when we fetch the writer object from persistent config, the writer object will refer to the already destroyed config object. This needed to be fixed by this patchset as well.

### DIFF
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -319,6 +319,16 @@ log_threaded_dest_driver_start_thread(LogThrDestDriver *self)
                                  self, &self->worker_options);
 }
 
+static void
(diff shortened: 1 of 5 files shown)


kira-syslogng:

kira-syslogng:

kira-syslogng:

PR #1561, opened 2017-06-25
Eventlog cleanup

mochrul:
After the eventlog is merged almost as-is to the syslog-ng tree now it is time to make it more integrated part of the syslog-ng build system.

Removed all files from the libeventlog which is not needed anymore including the configure.ac and also added the tests used in libevtlog to the syslog-ng test system.


kira-syslogng:

MrAnno:
@kira-syslogng ok to test

kira-syslogng:

bazsi:
This looks really good. One more thing we could perhaps do is to get rid off the separate src level in the directory hierarchy and do it like this:

lib/eventlog/ - source files
lib/eventlog/tests - test related files

What do you think?

mochrul:
@bazsi That means we mix the legal things (AUTHORS, COPYING, NEW, README) and the code together. It totally ok to me, but take it into the consideration as well.

bazsi:
ok, let's merge this, I'll do the relicensing stuff and get rid off those files (obviously checking if the code is still my copyright, which it should be).

once we take care of the legal stuff, we can move files to their final resting locations.

cheers,
Bazsi

kira-syslogng:

kira-syslogng:

mochrul:
I'm trying to figure out why the clang compilation was failed (I have found the symptoms, I do not know why it happens.) but I would appreciate any help with it.

I tried to reproduce it in my test environment but the clang just died when I run it with -j and the issue does not appear when I try to run the make -j install after I compiled it with only one thread.

MrAnno:
Hi,

This `make -j install` relinking issue was not introduced by you, you just triggered the problem with the eventlog modification.

With @lbudai, we reproduced this issue without your modifications and without clang.
The root cause of the problem is that automake can not create a **install-time** dependency between `module_LTLIBRARIES` and `lib_LTLIBRARIES`, so the relinking procedure fails when syslogformat is installed before libsyslog-ng.

You can trigger this behaviour on the master branch:
- configure syslog-ng with a --prefix path *that does not exist*
- in the build directory, edit the generated `Makefile`:
```diff
install-libLTLIBRARIES: $(lib_LTLIBRARIES)
        @$(NORMAL_INSTALL)
+       sleep 3; 
        @list='$(lib_LTLIBRARIES)'; test -n "$(libdir)" || list=; \
```
- run make -j3 install

The patch above proves that this is a thread synchronization/timing issue. 


We will come up with a fix somehow. Until then, I wouldn't run `make install`  in parallel (`make` is OK).

lbudai:
some additional info:
you can check the generated `Makefile`: there are two (independent) target, one is for installing the `lib_LTLIBRARIES` and other is for installing the `module_LTLIBRARIES`. In case of `make -j`, the relinking phase (as we are using `RPATH` can fail as the installation of  one of the modules can be done before `libsyslog-ng.so` is installed.
(when we add an explicit dependency rule to the generated makefile, it worked as expected.

MrAnno:
We've disabled the make install parallelization in Travis. Please rebase this branch to master.

mochrul:
I have the feeling that this patch is cursed.

kira-syslogng:

lbudai:
@mochrul : I have the same feeling :) we need to restart the Travis jobs, hope it works at this time.

mochrul:
Still no luck...

MrAnno:
It seems OBS is dying, so Travis can't install our dependencies. I'll keep restarting the job..

MrAnno:
Well, that's unlucky.. There is one more unrelated issue here:

```
$ make distcheck -j 3
...
make -C codegen clean
make: *** codegen: No such file or directory.  Stop.
make[2]: [clean-local] Error 2 (ignored)
```

This is coming from the rabbitmq-c submodule.

mochrul:
What is the status of this patch? It still fails and it still caused by issue outside of the patch as far as I know.

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

mochrul:
Sorry, I'm still trying to find the holy grail.
Meanwhile, I found a bug which talks about a similar case: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=7328

mochrul:
Oh. And in the make distcheck there is still a -j 3, so removing it from the install was/is not enough.

bazsi:
Do we need eventlog to be a separate so file? we could just link or into
libsyslog-ng just like the rest of the core. Wouldn't that fix the
dependency issue?

mochrul:
@bazsi: Yes, that can help in shrinking the race window.
But I do not know if that can be done legally, regarding the different license.

At https://debbugs.gnu.org/cgi/bugreport.cgi?bug=7328 they propose a workaround which is to add the direct dependency between the two target. Unfortunately, the suggested lines generate a syntax error, so I have to figure out how that can be done.

kira-syslogng:

kira-syslogng:

MrAnno:
Thanks.

PR #1565, opened 2017-06-26
str-funcs: ignore locale when detecting whitespaces

MrAnno:


kira-syslogng:

MrAnno:
You are right, thanks.

Actually, we have to change all locale specific `is*()` calls to a non-locale and UTF-8 safe version.
I'll create a new PR with the complete fix.

PR #1567, opened 2017-06-27
RFC: Merge release branch back to "master"

bazsi:
This PR is to facilititate feedback if we could merge the debian/rpm packaging to master. I understand that we have taken those out because:

 * originally, our debian packaging was very limited, not really matching the one in the official Debian tree
 * packagers/maintainers need to have the ability to change the packaging after release, and associate/change a version number
 * these changes often conflicted, causing friction
 * maintainers don't want to go through a Pull Request process, when they are trying to release a new version, already released.

However, this means that pristine source trees checked out of github will not be easy to compile into debs and RPMs and with the advent of the docker based build environment this should be very easy.

So this is what I am proposing:
 * master does contain the debian subdirectory and the spec file
 * material changes to packaging go through the PR process, submitted by maintainers
 * whenever we do a release, we update debian/changelog and the specfile with proper version number bumps and a summary of changes. Debian/RPM revision set to -1, e.g. 3.10.1-1.
 * when the maintainer does a release, he is able to bump the Debian/RPM revision, e.g. produce 3.10.1-2, -3 and so on with changes as he/she sees fit
 * changes incorporated into -2 and further packaging revisions should go through the PR process, so that we are synced up with debian/changelog and packaging in general.

This will allow users to always compile their own debs/RPMs  if they see fit and should avoid conflicts as well, btw, this sed magic helps resolving trivial conflicts in debian/changelog:

```
$  sed -i ':multiline_hack; N;s/<<<<<<< HEAD\n\(.*\)=======\n\(.*\)>>>>>>> ubuntu\n\(.*\)/\2\n\1\3/; T multiline_hack' debian/changelog
```

although this assumes the conflict happens against a branch named "ubuntu".

So my RFC here is whether we can merge back and find a process that is fit for everyone.
@mochrul @czanik @lbudai 

Please don't merge this unless there's an agreement in this ticket.
thanks.


kira-syslogng:

mochrul:
I have s strong opinion related to this. But I'm sure this comes as a no surprise because the release branch was created from my idea.
In my opinion, the upstream work and the packaging work are two (very) different job and mixing the two just cause problems.
On the other side, I understand that there are users who want to build a Debian/Redhat package from any git commit and this is needed for that.

From my point of view, I already implemented safeguards to prevent the conflict, therefore, I need no worry about the merge. But let to be clear. In my opinion, the Debian packaging is not part of the upstream repo and I will act like this.

czanik:
I also have many doubts. There is no such thing as a single spec to package syslog-ng as RPM. There are many different spec files, depending on the following factors:

- distribution (suse, redhat)
- distribution version (rhel 6, rhel7)
- using only distribution packages (no Java destinations or some exotic dependencies this way) or also others (in this case the missing dependencies also need to be sourced from somewhere)

Of course we can choose one of these, but that will cover just one of the many possible use cases. Good for automatic testing and a small part of the users...

bazsi:
The multiple spec file problem can be resolved easily. The question is if we can merge spec file changes back to the main repo.

It shouldn't be more than a pull request away, or am I missing something?

bazsi:
I understand your stance and also understand the need to be able to change the packaging past the release, so that syslog-ng can be patched up and properly integrated to various distros.

My question is whether in addition to that, we could merge all your spec/debian changes to this repo, using the PR process?

For instance, the spec file for the RPM needed to be changed for the 3.10.1 release, and this specfile is not available to syslog-ng users, at least not from this repo. To build an RPM from git head, someone needs to 1) either wait for your packaging, 2) grab an older rpm, patch in the specfile and then get an RPM.

Now I've found the spec file on copr, but I wouldn't have found it, unless I know where to look. That's the pain I would like to solve, and similarly with the Debian directory.

If we keep the two in sync, conflicts should not arise, you would still be the primary maintainers for the packaging work and it would be trivial to compile a git snapsot into rpm/deb packages.

Thanks for your consideration.

lbudai:
@bazsi : could we talk about this IRL on next week?

bazsi:
@mochrul what do you think about my latest proposal? I was also thinking about pulling packaging changes into this repo automatically, but it'd still be better to use a common workflow.

Cheers,
Bazsi

mochrul:
Sorry, for the long silence. As I mentioned earlier I have a solution which will prevent conflicts to occur, at least when I merge the new version. Algernon had a similar solution, but mine is more general, I think.

On the other hand, giving back the changes can happen, but because the two has no common root, it isn't as easy. And also, we have a bit different goals and focus points in the process, which also makes it a bit harder to do it. At least more prompt responses, mainly from me. :)

I mainly agree with @lbudai, an IRL meeting would be good. Even a skype/hangout/messenger meeting could work.

bazsi:
I am hoppy to pay for a coffee/apple juice/beer somewhere the week after
next week to facilitate discussion. Monday or Thursday would work.


-- 
Bazsi

mochrul:
I can go to the BalaBits direction when I'm commuting, so every morning and/or the time after the working hours can work for me.

bazsi:
My last meeting ends at 6pm on Monday. When are you in the area? Or Thursday could work, i finish a bit earlier that day.

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

presidento:
It was decided IRL, we won't merge it.

/cc @mochrul @bazsi @lbudai

PR #1569, opened 2017-06-27
bugfix (fixes #1635): use sd_journald_test_cursor() from systemd-jour…

### LINKED ISSUE #1635: Could not GET 'log4j.pom' when compile java-modules of syslog-ng
issue: unable to get log4j 1.2.6.pom during compile. In Rhel7.2, by default log4j 1.2.7 is installed.

Still the issue persist. Kindly advise.

P.S: server don't have internet access. so unable to connect online mave Repo.

faithlesstomas:
…nal API for additional cursor test in syslog-ng.presit file

short bug description: when syslog-ng checks cursor position from
syslog-ng.persist file it does not check if cursor is "in the future" with
respect to the system.  This causes logging glitch when syslog-ng is started
(syslog-ng waits for the "time to come" to continue logging) when there is no
systemd journal entries with the "future time". This is because
sd_journald_seek_cursor() seeks to the next closest entry (in terms of time)
if no entry matches.

### DIFF
--- a/modules/systemd-journal/journal-reader.c
+++ b/modules/systemd-journal/journal-reader.c
@@ -322,8 +322,9 @@ _seek_to_saved_state(JournalReader *self)
 {
   JournalReaderState *state = persist_state_map_entry(self->persist_state, self->persist_handle);
   gint rc = journald_seek_cursor(self->journal, state->cursor);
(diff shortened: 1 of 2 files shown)


kira-syslogng:

Issue #1513, opened 2017-05-31
Application abort when using amqp destination


freak82:
## Version of syslog-ng

```
[/etc/syslog-ng]# syslog-ng --version
syslog-ng 3.9.1
Installer-Version: 3.9.1
Revision: 
Module-Directory: //usr/lib64/syslog-ng
Module-Path: //usr/lib64/syslog-ng
Available-Modules: add-contextual-data,afamqp,affile,afprog,afsocket,afstomp,afuser,basicfuncs,cef,confgen,cryptofuncs,csvparser,date,dbparser,disk-buffer,graphite,json-plugin,kvformat,linux-kmsg-format,pseudofile,sdjournal,syslogformat,system-source
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on

```

## Platform

```
[/etc/syslog-ng]# cat /etc/centos-release
CentOS Linux release 7.3.1611 (Core) 
[/etc/syslog-ng]# uname -a
Linux c7-syslog-mcn-1 3.10.0-514.21.1.el7.x86_64 #1 SMP Thu May 25 17:04:51 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

## Failure

The syslog-ng deamon aborts when a amqp destination, to Azure Event Hub, is added.

```
May 31 19:30:24 c7-syslog-mcn-1 systemd: Starting System Logger Daemon...
May 31 19:30:24 c7-syslog-mcn-1 systemd: Started System Logger Daemon.
May 31 19:30:41 c7-syslog-mcn-1 syslog-ng: iv_timer_register: called with timer still on the heap
May 31 19:30:41 c7-syslog-mcn-1 systemd: syslog-ng.service: main process exited, code=killed, status=6/ABRT
May 31 19:30:41 c7-syslog-mcn-1 systemd: Unit syslog-ng.service entered failed state.
May 31 19:30:41 c7-syslog-mcn-1 systemd: syslog-ng.service failed.
May 31 19:30:42 c7-syslog-mcn-1 systemd: syslog-ng.service holdoff time over, scheduling restart.
May 31 19:30:42 c7-syslog-mcn-1 systemd: Starting System Logger Daemon...
May 31 19:30:42 c7-syslog-mcn-1 systemd: Started System Logger Daemon.
[...]

PR #1573, opened 2017-06-27
afamqp: layer violation in suspending thrdestdrv results in crash

furiel:
Fixes: https://github.com/balabit/syslog-ng/issues/1513

There were two issues with afamqp that lead to crashing when
timer_reopen registered.

Afamqp did not virtualize connect function, even though it
should have done. As a result, connect/reconnect logic did not work
inside LogThrDestDrv, essentially logthrdestdrv marked the destination
as connected even though connection failed. Therefore, logthrdestdrv
pushed message to afamqp. Afamqp knew well that she was not connected,
so she tried to reconnect. Though the connection failed again, so
WORKER_INSERT_RESULT_NOT_CONNECTED was sent back to
logthrdestdrv. Receiving this error code, logthrdestdrv start the
suspend logic.

Afamqp, as the the connect function was not virtualized, tried to
execute the suspend logic during connect as well.

In the end, suspend logic started two times, both by afamqp and
logthrdestdrv, resulting timer_reopen registered two times, hence the
crash.

This patch virtualizes connect function of afamqp, removes the suspend
logic triggered inside afamqp, and removes
log_threaded_dest_driver_suspend from header because destination
drivers should not call it at all.

### DIFF
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -39,7 +39,7 @@ log_threaded_dest_driver_format_seqnum_for_persist(LogThrDestDriver *self)
   return persist_name;
 }
 
-void
+static void
 log_threaded_dest_driver_suspend(LogThrDestDriver *self)
 {
   iv_validate_now();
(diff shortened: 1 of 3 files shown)


kira-syslogng:

lbudai:
nice. thanks!

PR #1574, opened 2017-06-27
dbparser: add drop-unmatched() option

bazsi:
This patch adds a new drop-unmatched() option to db-parser() that causes
unmatched messages to be dropped, like this:

db-parser(drop-unmatched(yes));

If none of the rules match, the message would be dropped. If there's
a match the message goes "through".


kira-syslogng:

pzoleex:
Nice feature, can we do it for all of the other parsers?

lbudai:
:+1:

bazsi:
I agree this should be available everywhere.

csvparser has the drop-invalid flag, kv-parser always suceeds. I dont know
what date parser does if it fails.

PR #1595, opened 2017-07-10
dbparser: Rework getenv dependency in unit test

gaborznagy:
Some cases the test_patterndb unit test is not working, e.g. running the unit test binary itself natively (not with make check) without setting the "top_srcdir" environment variable.

Fixing the case when ENABLE_DEBUG option is not set and function used a possibly non-existing production-PATH): 
- the getenv "top_srcdir" dependency replaced by a configure set macro
(note: ifdef was not protecting, code-part always compiled, only worked in production because getenv failed)

Unit test can be run standalone.

Interface "pdb_file_validate" is split by use cases: production and tests.
No need to determine it in one function during runtime.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -98,6 +98,7 @@ xsddir='${datadir}/syslog-ng/xsd'
 
 config_includedir='${datadir}/syslog-ng/include'
 scldir="${config_includedir}/scl"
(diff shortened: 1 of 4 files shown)


kira-syslogng:

Issue #1047, opened 2016-05-10
PatternDB unit test failure 


MrAnno:
After introducing [_get_xsddir_in_build](https://github.com/balabit/syslog-ng/pull/1007/files#diff-085601782dbca00fd10d8091ee66d162R84) in #1007, one of the PatternDB's tests (modules/dbparser/tests/test_patterndb) fails with the following error message: 

```
FAIL: ASSERTION FAILED: Error validating patterndb, error=XSD file is not available at /opt/syslog-ng/share/syslog-ng/xsd/patterndb-4.xsd
```

This is only a problem when the `--enable-debug` flag is not set.
I think `top_srcdir` should be used in all cases when we run unit tests.

VIthulan:
This test failed for me as well, when I tried to build using Docker image.

PR #1597, opened 2017-07-11
dbparser: rework getenv dependency, test rework

gaborznagy:
Fixes: #1047 

Fixing an error case when ENABLE_DEBUG option is not set 
and test_patterndb tests use a non-existing "production PATH"(install dir set by prefix).

- the getenv "top_srcdir" dependency replaced by a configure set macro
  Unit test can be run standalone.
- Autoconf [preset variable](https://www.gnu.org/software/autoconf/manual/autoconf.html#Preset-Output-Variables)  "abs_top_srcdir" is not working,
  there is even a [bug report](https://lists.gnu.org/archive/html/bug-autoconf/2013-02/msg00008.html) about it. The workaround is to have an own variable and export that one in syslog-ng-config.h
For CMake this is not an issue.

Interface "pdb_file_validate" is split by use cases: production and tests.
Existing interface is untouched and is using the production path,
a new "pdb_file_validate_in_tests" interface is using the build path.
No need to determine the xsd directory path from one function during runtime.


kira-syslogng:

PR #1602, opened 2017-07-13
cisco-parser: Added prefix option

pzoleex:
With this option a user can change the prefix for cisco specific fields
and for example add it to SDATA. Just change prefix to ".SDATA.cisco."


kira-syslogng:

PR #1607, opened 2017-07-19
Rework of Solaris SMF multi-instance support, to support unified-style default files

jszigetvari:
contrib/solaris-packaging/syslog-ng.method: reworked to support new default files
contrib/solaris-packaging/syslog-ng@default: added an example default file
contrib/solaris-packaging/solaris10_install.txt: added step to install default file

    * added support for systemd-multi-instance type of default files
    * default files should now work the same way as with systemd instances
    * naming of the default files has changed to reflect those supporting systemd
    * reworked the method file to support additional options
    * set up a hierachy how settings take precedence of each other
    * installation directories/prefixes such as /usr/local may be a bit off
      and may require corrections by the end user to match his/her use case
    * renamed config variables to match those used with systemd's instances

Signed-off-by: Janos SZIGETVARI <jszigetvari@gmail.com>

### DIFF
--- a/contrib/solaris-packaging/solaris10_install.txt
+++ b/contrib/solaris-packaging/solaris10_install.txt
@@ -91,7 +91,7 @@ Procedure 4.5. Installing syslog-ng on Sun Solaris 10
                   
    6.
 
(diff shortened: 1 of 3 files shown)


kira-syslogng:

Issue #1598, opened 2017-07-11
riemann destination doesn't set the event's time


faxm0dem:
It seems that events sent to riemann do not contain the time.
This means they will get the current time once inside riemann.
It would be more correct if syslog-ng sent ISODATE by default (converted to milliseconds) and even better if it were configurable like `service`, `host`, *etc.*

#### Event generated by syslog-ng

![riemann-syslog-ng](https://user-images.githubusercontent.com/112638/28075470-e0356fca-665b-11e7-87fe-3c3932a4048d.png)

#### Event generated by collectd

![riemann-collectd](https://user-images.githubusercontent.com/112638/28075471-e035da82-665b-11e7-9319-ccaaf166dfe9.png)

lbudai:
@faxm0dem : we are reopening this issue as we also want to solve the milliseconds issue

PR #1610, opened 2017-07-19
riemann: configurable event time

furiel:
New configuration parameter is added to riemann destination that
controls the calculation of event time: event_time(). In case omitted,
the default value is ${UNIXTIME}.

For example:
destination d_riemann {
    riemann(
    	server("127.0.0.1")
    	port(5555)
        event_time("${UNIXTIME}")
        [...]
    );

Fixes: https://github.com/balabit/syslog-ng/issues/1598

### DIFF
--- a/modules/riemann/riemann-grammar.ym
+++ b/modules/riemann/riemann-grammar.ym
@@ -47,6 +47,7 @@
 %token KW_SERVER
 %token KW_STATE
 %token KW_SERVICE
+%token KW_EVENT_TIME
 %token KW_DESCRIPTION
 %token KW_METRIC
 %token KW_TTL
@@ -88,6 +89,10 @@ riemann_option
         | KW_SERVICE '(' template_content ')'
           {
             riemann_dd_set_field_service(last_driver, $3);
+          }
+        | KW_EVENT_TIME '(' template_content ')'
+          {
+            riemann_dd_set_field_event_time(last_driver, $3);
           }
         | KW_STATE '(' template_content ')'
           {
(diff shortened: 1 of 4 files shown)


kira-syslogng:

furiel:
@faxm0dem could you please check if this patch is fine for you?

I omitted the conversion to milliseconds part, because as far as I know syslog-ng macros only support seconds-precision. However riemann-c-lib supports both through RIEMANN_EVENT_FIELD_TIME, RIEMANN_EVENT_FIELD_TIME_MICROS constants (now I used  RIEMANN_EVENT_FIELD_TIME), so I can add microseconds as an additional configuration parameter also, if there is a use case.

kira-syslogng:

faxm0dem:
thanks so much! I'll test it ASAP

bazsi:
There are macros for fractions of a second as well.

furiel:
Ok I look into the milliseconds parts as well.

furiel:
There is a small problem with the microseconds.

It was added to Riemann a few months ago:
https://github.com/riemann/riemann/releases/tag/0.2.13
and for example the xenial version of riemann-c-client does not support it yet:
from event.h of the  libriemann-client-dev 1.8.1-2build1
>  RIEMANN_EVENT_FIELD_METRIC_F,

>    RIEMANN_EVENT_FIELD_STRING_ATTRIBUTES
>  } riemann_event_field_t;

while in riemann-c-client master:
>    RIEMANN_EVENT_FIELD_METRIC_F,

>    RIEMANN_EVENT_FIELD_STRING_ATTRIBUTES,
>    RIEMANN_EVENT_FIELD_TIME_MICROS
>  } riemann_event_field_t;

So unless users compile the latest riemann-c-client themselves besides syslog-ng, and link it, they cannot use the microseconds version.

I think we can still detect compatibility in configure phase, and conditionally compile support for microseconds.

Still, as the deadline of the current ose release is very close, I'm afraid I can patch this only in the next release, due that the implementation will be somewhat complex. I suggest merge this pr as it is now because this fixes at least a bug when time is missing from events, and leave the referring ticket open for the enhancement part.

faxm0dem:
I agree, thanks for your work!

PR #1613, opened 2017-07-20
logthrdestdrv: add missing GC call

MrAnno:
Like in afsql, an explicit GC call is necessary before (at least) every do_work().


kira-syslogng:

PR #1614, opened 2017-07-20
SCL: Add missing reference for osquery into scl/Makefile.am

mitzkia:
* without this patch "scl/osquery/plugin.conf" will not copied under share/syslog-ng/include/scl/


kira-syslogng:

Issue #1616, opened 2017-07-24
HOST macro is not working for ipv6 addresses in file source


furiel:
```
@version: 3.10
@include "scl.conf"

options {
#    use_dns(no);
    keep_hostname(yes);
    threaded(yes);
};

source s_file {
   file("/tmp/aaa");
};

destination d_file {
   file("/tmp/bbb" template("HOST: ${HOST} \n"));
};

log {
	source(s_file);
	destination(d_file);
};
```

For ipv6, the hostname of the local machine is emitted instead of th ip address:
echo "<38>2017-05-24T13:09:46 2a00:1450:400d:802::200e prg0[1234]: seq: 0000001234, thread: 0000, runid: 1495624186, stamp: 2017-05-24T13:09:46 PADDPADDPADD" >> /tmp/aaa
--> HOST: (HOSTNAME)

For ipv4, this is working as expected:
echo "<38>2017-05-24T13:09:46 1.2.3.4 prg0[1234]: seq: 0000001234, thread: 0000, runid: 1495624186, stamp: 2017-05-24T13:09:46 PADDPADDPADD" >> /tmp/aaa
--> HOST: 1.2.3.4

PR #1617, opened 2017-07-24
syslog-format: cannot parse ipv6 into hostname

furiel:
Fixes:
https://github.com/balabit/syslog-ng/issues/1616

### DIFF
--- a/modules/syslogformat/syslog-format.c
+++ b/modules/syslogformat/syslog-format.c
@@ -635,6 +635,47 @@ _init_parse_hostname_invalid_chars(void)
     }
 }
 
+typedef struct _IPv6Heuristics
+{
+  gint8 current_segment;
+  gint8 digits_in_segment;
+  gboolean heuristic_failed;
+} IPv6Heuristics;
+
+static gboolean
+ipv6_heuristics_feed_gchar(IPv6Heuristics *self, gchar c)
+{
+  if (self->heuristic_failed)
+    return FALSE;
+
+  if (c != ':' && !g_ascii_isxdigit(c))
+    {
+      self->heuristic_failed = TRUE;
+      return FALSE;
+    }
+
+  if (g_ascii_isxdigit(c))
+    {
+      if (++self->digits_in_segment > 4)
+        {
+          self->heuristic_failed = TRUE;
+          return FALSE;
+        }
+    }
+
+  if (c == ':')
+    {
+      self->digits_in_segment = 0;
+      if (++self->current_segment >= 8)
+        {
+          self->heuristic_failed = TRUE;
+          return FALSE;
+        }
+    }
+
+  return TRUE;
+}
+
 static void
 log_msg_parse_hostname(LogMessage *self, const guchar **data, gint *length,
                        const guchar **hostname_start, int *hostname_len,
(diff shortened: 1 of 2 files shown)


kira-syslogng:

kira-syslogng:

furiel:
Not nice but maybe works:

We could use inet_pton to check if we see a valid ipv6 address. If yes, we set the hostname. If not, we go with the original heuristic.

Of course we need to prepare the candidate string for inet pton, so first I need to create nullterminated substring until the first space or '[' character, which then can be passed to inet_pton.

furiel:
Or this is maybe better: g_hostname_is_ip_address (), from glib 2.22

bazsi:
never forget the performance implications, this is a very hot path here.


-- 
Bazsi

kira-syslogng:

furiel:
Instead of explicitely validating for ipv6 address, I added a new version which relaxes the condition for colon being delimiter. The new heuristics just check a few extra characters around the colon, so probably have lower effect on performance than calling a glib function or inet_pton.

presidento:
We should check a bit more strictly that it seems like an ipv6 address. (Hexa numbers, colons...)

kira-syslogng:

kira-syslogng:

furiel:
@bazsi I replaced the previous heuristics with a more strict one. In the end, an ipv6-heuristics class added just checks if the candidate is max 8 blocks of max 4-wide hexa characters. I felt the class useful, because the state of the ipv6 part became more complex than the previous one.

The good news is I could still avoid memory allocation, so I am very hopeful that we do not lose too much performance. Anyway, we will run a performance test for this version next week.

modules/syslogformat/syslog-format.c
```diff
@@ -635,6 +635,48 @@ _init_parse_hostname_invalid_chars(void)
...
+        {
+          self->heuristic_failed = TRUE;
+          return FALSE;
+        }
+    }
+
+  if (c == ':')
+    {
+      self->digits_in_segment = 0;
+      if (self->current_segment > 8)
```
MrAnno: `current_segment` is not incremented.

modules/syslogformat/syslog-format.c
```diff
@@ -635,6 +635,48 @@ _init_parse_hostname_invalid_chars(void)
     }
 }
 
+typedef struct _IPv6Heuristics
+{
+  gchar current_segment;
+  gchar digits_in_segment;
```
MrAnno: Please use `guint8` for 8 bit integers.

modules/syslogformat/syslog-format.c
```diff
@@ -635,6 +635,48 @@ _init_parse_hostname_invalid_chars(void)
...
+  gboolean heuristic_failed;
+  gboolean (*feed_gchar)(struct _IPv6Heuristics *, gchar);
```
MrAnno: Is there any reason why you made this virtual?

kira-syslogng:

kira-syslogng:

kira-syslogng:

pzoleex:
the perftest passed, there was no regression

PR #1619, opened 2017-07-25
autogen: include {debian,tgz2build}/Makefile.am when exists

lbudai:
Unfortunately conditional include of an automake file is not supported.


kira-syslogng:

PR #1624, opened 2017-07-27
cfg: load license after '@module' pragma initialization

MrAnno:
At this point, all plugin info should be available for license.

This is the Lagrangian point of the license module.


kira-syslogng:

PR #1629, opened 2017-07-28
mainloop: regression: reload not finalized if no workers around

furiel:
The reload mechanism is syslog-ng is done in 3 stages.

- Stage1: ctl/hup signal initiates reload
Old configuration saved, new configuration validated, stage2 is
registered to be executed when no workers are
running (main_loop_worker_sync_call).
All workers are notified to stop as soon as possible.

- Stage2: Configuration applied.
Configuration applied, (workers are still stopped), stage3 registered to be executed later.

- Stage3: Reenables workers, reload finished.

There is a special case when no workers are created. In that case the
three stages can be executed in one step. It is handled by a special
if-branch in main_loop_worker_sync_call. But this branch only called
stage2, and did not call stage3. This means the reload was never
finalized, resulting in that users cannot reinitiate reload anymore.

This patch calls stage3 after stage2 in this special case, fixing the
problem.

Regression was introduced in: https://github.com/balabit/syslog-ng/pull/1537

### DIFF
--- a/lib/mainloop-worker.c
+++ b/lib/mainloop-worker.c
@@ -381,6 +381,7 @@ main_loop_worker_sync_call(void (*func)(gpointer user_data), gpointer user_data)
   if (main_loop_workers_running == 0)
     {
       _invoke_sync_call_actions();
+      _reenable_worker_jobs(NULL);
     }
   else
     {


kira-syslogng:

MrAnno:
Reproduction steps for the "no-worker" case:

1. Start syslog-ng with this config:
```
@version: 3.11
@include "scl.conf"

source s_local {
    internal();
};

destination d_local {
    syslog("127.0.0.1" port(4444));
};

log {
    source(s_local);
    destination(d_local);
};
``

Reload syslog-ng with syslog-ctl twice (you can wait between the 2 reloads).

```
[2017-07-28T14:41:42.936909] Error initiating reload, reload is already ongoing
```

I guess this issue is related to the fact that there are now worker jobs (no incoming traffic) before calling a reload.


The main_loop_worker_job_complete() function is not called, so the reenable_jobs task won't be registered at all.

PR #1631, opened 2017-08-01
python: add backtick substitution for inline Python code

lbudai:
After this patch users can use a custom defined syslog-ng variable in inline
Python code.

Example:

@define PORT 5555

python {
  class PythonDst(object):
    def init(self, options):
      return True
    def send(self, msg):
      print("port:%s" % "`PORT`")
};

Requested by a user.

### DIFF
--- a/modules/python/python-grammar.ym
+++ b/modules/python/python-grammar.ym
@@ -70,8 +70,13 @@ start
           {
             gboolean result;
 
-	    result = python_evaluate_global_code(configuration, $4);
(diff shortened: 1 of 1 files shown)


kira-syslogng:

lbudai:
Please do not merge :)
This is just a poc, maybe there is a better way to get the same result.

lbudai:
yeah, this PR can be closed as the backtick subst. is applied to the `options`...

PR #1637, opened 2017-08-03
DockerBuild: Add docker image related fixes

mitzkia:


kira-syslogng:

kira-syslogng:

dbld/images/centos7/Dockerfile
```diff
@@ -3,11 +3,13 @@ MAINTAINER Andras Mitzki <andras.mitzki@balabit.com>
 
 RUN yum install -y wget epel-release
 
-RUN cd /etc/yum.repos.d && wget https://copr.fedorainfracloud.org/coprs/czanik/syslog-ng39/repo/epel-7/czanik-syslog-ng39-epel-7.repo
+RUN cd /etc/yum.repos.d && wget https://copr.fedorainfracloud.org/coprs/czanik/syslog-ng311/repo/epel-7/czanik-syslog-ng311-epel-7.repo
 
 ADD dev-dependencies.txt .
 RUN cat dev-dependencies.txt | grep -v "#" | xargs yum -y  install
-
+ENV GRADLE_HOME=/opt/gradle
```
lbudai: Why this step is needed?

mitzkia: It is necessary basic setup for gradle.
It is described here: https://docs.gradle.org/3.3/userguide/installation.html - 3.4. Environment variables

lbudai: I disagree :)

I think this is a packaging issue (@czanik: it is installed from your repo, is it possible to fix this issue there?).

Of course we can fix it, but well... this is not a required step on other distros and this is a workaround, it fixes the packaging in the Dockerfile.
(and just a note: on centos7 we are installing gradle 2.6 not 3.3  - and the installation notes are for the case when you install it manually not from a package).

Alternative solution: download gradle binaries from here: https://gradle.org/releases/ and set `GRADLE_HOME` and other PATH.

Dockerfile is also a documentation of a development environment. I think we should keep them as clean as possible.
This solution is a mixture of manual and package installation in case of gradle which is well... not the best.
IMHO when someone installs gradle(or other tool) from a package then no additional steps are required from the user (setting PATH and GRADLE_HOME).
What do you think?

In case of Debian/Ubuntu it's working. I also checked Fedora 26: on that system it is working as expected.

So my suggestion is to fix the gradle packaging, or, install gradle from binaries (provide an `install_gradle.sh` which downloads the binaries, install it and setup the environment).

mitzkia: Thanks for the note. After rethinking it I also agreed with you (it seems a packaging issue).
@czanik could you fix the gradle package?
The mentioned alternative solution is ok, but it comes with additional maintenance work: we need to update every time to latest package version e.g: 4.0.1 - 4.0.2 ...

lbudai: I don't know whether syslog-ng is build with latest gradle.

kira-syslogng:

dbld/images/centos7/Dockerfile
mitzkia: Changes are added. Also tested with the latest gradle 4.1 we can build syslog-ng.

kira-syslogng:

dbld/images/centos7/Dockerfile
```diff
@@ -3,11 +3,13 @@ MAINTAINER Andras Mitzki <andras.mitzki@balabit.com>
...
 
-RUN cd /etc/yum.repos.d && wget https://copr.fedorainfracloud.org/coprs/czanik/syslog-ng39/repo/epel-7/czanik-syslog-ng39-epel-7.repo
+RUN cd /etc/yum.repos.d && wget https://copr.fedorainfracloud.org/coprs/czanik/syslog-ng311/repo/epel-7/czanik-syslog-ng311-epel-7.repo
 
 ADD dev-dependencies.txt .
 RUN cat dev-dependencies.txt | grep -v "#" | xargs yum -y  install
-
+ADD gradle_installer.sh .
+RUN ./gradle_installer.sh
+RUN find / -name libjvm.so | xargs dirname | tee --append /etc/ld.so.conf.d/openjdk-libjvm.conf && ldconfig
```
furiel: What is the reason we run find here? On centos, isn't this file on a fix location? Maybe we can hardcode the path here.

mitzkia: Yes, the file generated on different locations depends on java-1.8.0-openjdk-devel version. The java version is in directory name. Example: 
`/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.141-1.b16.el7_3.x86_64/jre/lib/amd64/server`

czanik: You can cheat :) While the directory name is not fix, there is a symlink to it which is fix. So, as you can see in https://www.balabit.com/blog/syslog-ng-and-elasticsearch-5-getting-started-on-rhelcentos/ you can use a hard coded path name:
```
echo /usr/lib/jvm/jre/lib/amd64/server > /etc/ld.so.conf.d/java.conf
ldconfig
```
This will work as long as you work on AMD64

mitzkia: ok, thanks. I will fix this

kira-syslogng:

Issue #1562, opened 2017-06-26
java module crashes on recent kernels addressing CVE-2017-1000364


faxm0dem:
Recent kernels (Opensuse, CentOS, ...) make syslog-ng crash, as reported by @czanik on the [ML](http://lists.balabit.hu/pipermail/syslog-ng/2017-June/023903.html).

I just tested this with CentOS7:

* `kernel-3.10.0-514.21.2.el7` crash
* `kernel-3.10.0-514.10.2.el7` OK

### crash log

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGBUS (0x7) at pc=0x00007fc72d012c26, pid=11359, tid=0x00007fc75fc1eb40
#
# JRE version:  (8.0_131-b12) (build )
# Java VM: OpenJDK 64-Bit Server VM (25.131-b12 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# j  java.lang.Object.<clinit>()V+0
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /tmp/hsperfdata_root/hs_err_pid11359.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
#
Aborted
```

faxm0dem:
Sandor Geller suggests to change the `-Xms` parameter. Apparently JVM settings are initialized [here in the code](/balabit/syslog-ng/blob/master/modules/java/native/java_machine.c#L102-L108)

faxm0dem:
Hi, can you please give us a timeline on the fix? We obviously don't want to freeze the (vulnerable) kernel version for too long on our production nodes.

Thanks so much in advance

lbudai:
@faxm0dem : we will check(and fix) it in the next few(7) weeks.

lbudai:
@faxm0dem : could you test it with kernel-3.10.0-514.26.1.el7 ?

faxm0dem:
I can test easily with any EL7 kernel, just ask

lbudai:
@faxm0dem : I wrote about weeks, not days :)
But I have a PR: you can set the thread stack size (1M in this example) by adding a 
`jvm_options("-Xss1M")` to your java destination.

faxm0dem:
haha sorry thanks I'll test this on monday

faxm0dem:
[...]

PR #1639, opened 2017-08-03
java: add jvm_options to java destination

lbudai:
syslog-ng defines the following options by default:
 * -Djava.class.path
 * -Djava.library.path
 * -Xrs

These java.class.path and java.library.path options are cannot be
overriden from syslog-ng config.

example:
options {
        jvm_options("-Xss1M -XX:+TraceClassLoading");
};

 hdfs(client-lib-dir(/home/stentor/work/hdfs-libs/libs)
      hdfs-uri("hdfs://hdp:8020") hdfs_archive_dir("")
      hdfs_file("data.txt")
      template("$MSG\n")
);

Fixes: #1562

### DIFF
--- a/modules/java/native/java_machine.c
+++ b/modules/java/native/java_machine.c
@@ -30,11 +30,11 @@
 #include "lib/reloc.h"
 #include "plugin.h"
 #include "resolved-configurable-paths.h"
+#include <string.h>
 
 struct _JavaVMSingleton
 {
   GAtomicCounter ref_cnt;
-  JavaVMOption options[3];
   JNIEnv *env;
   JavaVM *jvm;
   JavaVMInitArgs vm_args;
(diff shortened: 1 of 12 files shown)


kira-syslogng:

modules/java/proxies/java-destination-proxy.c
```diff
@@ -192,14 +192,15 @@ java_destination_proxy_free(JavaDestinationProxy *self)
...
+java_destination_proxy_new(const gchar *class_name, const gchar *class_path, gpointer handle, LogTemplate *template,
+                           const gchar *jvm_options)
 {
   JavaDestinationProxy *self = g_new0(JavaDestinationProxy, 1);
   self->java_machine = java_machine_ref();
   self->formatted_message = g_string_sized_new(1024);
   self->template = log_template_ref(template);
 
-  if (!java_machine_start(self->java_machine))
+  if (!java_machine_start(self->java_machine, jvm_options))
```
MrAnno: `self->java_machine` points to a "singleton" object, so only the last (or first?) `jvm-options()` usage in the configuration will be in effect.

Shouldn't we use a global option instead?
We can keep this per-destination option as well, but I would not document this in the admin guide due to its misleading behavior.

lbudai: Yes, it should be a global option.
I'm working on it.

lbudai: I don't really like it as a global option... but it is better than the per-destination based implementation.

kira-syslogng:

modules/java/proxies/java-destination-proxy.c
MrAnno: Yeah :S It looks weird but this way it's not that deceptive.

pzoleex:
@kira-syslogng test this please test branch=pzolee-java-dest-jvm-options;

kira-syslogng:
Build SUCCESS, the tests were executed on test branch: pzolee-java-dest-jvm-options and test suite: functions

modules/java/native/java_machine.c
```diff
@@ -92,24 +92,92 @@ java_machine_unref(JavaVMSingleton *self)
...
+  static const gchar *predefined_options[] =
+  {
+    "Djava.class.path",
+    "Djava.library.path",
+    NULL
+  };
+
+  for (gint i = 0; predefined_options[i] != NULL; i++)
+    {
+      if (!strncmp(option, predefined_options[i], strlen(predefined_options[i])))
```
MrAnno: I wouldn't use `strncmp()` when the length of the string can not be known explicitly (`strlen()` won't stop either).

kira-syslogng:
Build SUCCESS, the tests were executed on test branch: pzolee-java-dest-jvm-options and test suite: functions

Issue #1494, opened 2017-05-23
syslog-ng leaks the internal messages when not sent to somewhere


pzoleex:
So I have this very easy config listed below.
Just start syslog-ng -F (don't need any live destination)
and get the counters:
root@test21:~# /opt/syslog-ng/sbin/syslog-ng-ctl stats  | grep internal
global;internal_queue_length;;a;processed;3
/opt/syslog-ng/sbin/syslog-ng-ctl stats  | grep allocated
global;msg_allocated_bytes;;a;value;3120

As you can see there are 3 messages stored in somewhere (called internal queue whatever it means) and the allocated_bytes are also increased.
If I allow syslog-ng to run for longer time, these values will increase and never become zero.

But If I start syslog-ng with -e, like syslog-ng -Fe, the leaks disappear. The difference that in that case the internal message are displayed in the console.

version:
latest master (compiled from source): commit 2f698b01cb0a46d1057206fc275ae92fda2d3562
```
@version: 3.9
options {
    keep_hostname(yes);
    keep_timestamp(no);
    stats_level(2);
    use_dns(no);
};
source s_network_09f74da273d446de87aeb63e7c3afc13 {
    network(ip(0.0.0.0)
 log_fetch_limit(1000)
 log_iw_size(100000)
 max_connections(100)
 port(514));
};
destination d_network_ffbdff7305ef45709b9ddf8be4079473 {
    network("127.0.0.1" disk_buffer(disk_buf_size(1048576000) mem_buf_length(100000) reliable(no))
 port(2121));
};
log {
source(s_network_09f74da273d446de87aeb63e7c3afc13);
destination(d_network_ffbdff7305ef45709b9ddf8be4079473);

flags(flow-control);
};
```

PR #1643, opened 2017-08-13
afinter: current_internal_source was used for 2 different things

juhaszviktor:
1) Is there any configured internal source in the config
2) Does the configured internal source free to send or not

If there isn't configured internal source,
the messages shouldn't be stored in the internal queue,
because there is no one to read them.

But if there is a configured internal source,
the messages should be stored in the internal queue,
independently from the queue source is ready to work or not.

There is a side effect of this patch, that if there was no
internal source, before reload, but after reload it exists,
then the internal messages generated before reload will be lost

Fixes: #1494

Signed-off-by: Juhász Viktor <juhasz.viktor81@gmail.com>

### DIFF
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -96,7 +96,7 @@ struct _AFInterSource
   struct iv_event schedule_wakeup;
   struct iv_timer mark_timer;
   struct iv_task restart_task;
-  gboolean watches_running:1;
+  gboolean watches_running:1, free_to_send:1;
 };
 
 static void afinter_source_update_watches(AFInterSource *self);
@@ -219,7 +219,7 @@ afinter_source_update_watches(AFInterSource *self)
        * current_internal_source to NULL.  Messages get accumulated into
        * internal_msg_queue.  */
       g_static_mutex_lock(&internal_msg_lock);
-      current_internal_source = NULL;
+      self->free_to_send = FALSE;
       g_static_mutex_unlock(&internal_msg_lock);
 
       /* Possible race:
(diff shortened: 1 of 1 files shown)


kira-syslogng:

lib/afinter.c
```diff
@@ -219,7 +219,7 @@ afinter_source_update_watches(AFInterSource *self)
        * current_internal_source to NULL.  Messages get accumulated into
```
MrAnno: This comment is not relevant anymore :)

lib/afinter.c
```diff
@@ -428,6 +429,12 @@ void
 afinter_message_posted(LogMessage *msg)
 {
   g_static_mutex_lock(&internal_msg_lock);
+  if (!current_internal_source)
+    {
+      log_msg_unref(msg);
```
lbudai: @juhaszviktor : shouldn't we clear (and unref the LogMessage instances) the queue when it contains some items after a reload? Otherwise we could cache some items in situations when 
 * user starts syslog-ng with an internal source connected into a logpath
 * remove the internal source from config
 * reload syslog-ng
In this case, the queue could contain some items. Of course this is a better situation than before when we just "cache" all new internal messages without consuming them. But I think we could remove the cached items here.

bazsi:
@juhaszviktor are you planning to work on this yet or shall we try to massage this until integration?

juhaszviktor:
@bazsi I'm goint to work with on this. On this week.

lib/afinter.c
juhaszviktor: Ok, I'm going to correct this

kira-syslogng:

lib/afinter.c
juhaszviktor: Patch corrected

lbudai:
@juhaszviktor : thanks!

Issue #1317, opened 2017-01-13
Huge memory leak when syslog-ng handles continuous SIGHUP signals


SalutHZ:
## Version of syslog-ng
Name        : syslog-ng
Version     : 3.8.1 
Release     : 3.1   
Architecture: x86_64
Install Date: Fri Jan 13 09:59:05 2017
Group       : System/Daemons
Size        : 3132532 
License     : GPL-2.0 
Signature   : DSA/SHA1, Mon Sep  5 10:33:00 2016, Key ID eabb096191bb4167
Source RPM  : syslog-ng-3.8.1-3.1.src.rpm
Build Date  : Mon Sep  5 10:32:43 2016
Build Host  : build75 
Relocations : (not relocatable)
Vendor      : obs://build.opensuse.org/home:czanik
URL         : http://www.balabit.com/products/syslog_ng/
Summary     : The new-generation syslog-daemon

## Platform
SUSE 12

# Issue
1. We do a simple test for syslog-ng reloading its configuration by sending SIGHUP signal. We find a huge memory leak which may 1G in 10 min when syslog-ng handles continuous SIGHUP signals. 
2. Memory usage as follow:
 PID    USER     PR  NI    VIRT    RES        SHR       S   %CPU  %MEM   TIME+   COMMAND
 4597   root      20   0  849376 491812 239368    R    101.3    6.237   6:11.09    syslog-ng
3. Memory usage is still reachable and not released after we stop sending SIGHUP signals

## Steps to reproduce

1. start syslog-ng

2. send continuous SIGHUP signals
 #! /bin/sh
while :
do
        kill -HUP  $syslog-ng-PID
done

3. top -p PID 

## Configuration
Default without modification

czanik:
Version 3.9.1 is also available, which contains tons of bugfixes. It is packaged for different SUSE Linux releases: https://build.opensuse.org/project/show/home:czanik:syslog-ng39
Please check it also with that version.

SalutHZ:
@czanik thank for your suggestion. I do the same test in 3.9.1, it has the same problem.
Version info as follow:
Name        : syslog-ng
Version     : 3.9.1 
Release     : 1.1   
Architecture: x86_64
Install Date: Fri Jan 13 11:30:21 2017
Group       : System/Daemons
Size        : 3142733 
License     : GPL-2.0 
Signature   : DSA/SHA1, Thu Jan 12 13:49:29 2017, Key ID eabb096191bb4167
[...]

PR #1647, opened 2017-08-18
Memory leak related to cloning pipes

furiel:
There are two kind of memory leaks in syslog-ng related to cloning pipes. One of them is general: all clone invocations are affected. The other one is rewrite rule specific.
Please see the commit messages for detailed explanation of the issues and the patches.

These memory leaks can be reproduced if one defines a named filter/rewrite rule, and refer them twice in the configuration. For example:
```
@version: 3.11
@include "scl.conf"
filter f_warn  { level(warn, err, crit)  };
log {
  filter(f_warn);
  filter(f_warn);
};
```
These leaks are accumulated during reload.

As a workaround: configuration can be changed to use every reference only once.

Reported in: https://github.com/balabit/syslog-ng/issues/1317

### DIFF
--- a/lib/cfg-tree.c
+++ b/lib/cfg-tree.c
@@ -464,6 +464,8 @@ cfg_tree_compile_single(CfgTree *self, LogExprNode *node,
     {
       /* first reference to the pipe uses the same instance, further ones will get cloned */
       pipe->flags |= PIF_INLINED;
+      /* The pipe object is borrowed, so the reference counter must be increased. */
+      log_pipe_ref(pipe);
     }
   else
     {
@@ -477,7 +479,7 @@ cfg_tree_compile_single(CfgTree *self, LogExprNode *node,
         }
       pipe->flags |= PIF_INLINED;
     }
-  g_ptr_array_add(self->initialized_pipes, log_pipe_ref(pipe));
+  g_ptr_array_add(self->initialized_pipes, pipe);
   pipe->expr_node = node;
 
   if ((pipe->flags & PIF_SOURCE) == 0)
--- a/lib/rewrite/rewrite-subst.c
+++ b/lib/rewrite/rewrite-subst.c
@@ -92,7 +92,7 @@ log_rewrite_subst_clone(LogPipe *s)
   LogRewriteSubst *self = (LogRewriteSubst *) s;
   LogRewriteSubst *cloned;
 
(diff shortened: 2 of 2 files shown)


kira-syslogng:

lib/cfg-tree.c
```diff
@@ -476,8 +478,8 @@ cfg_tree_compile_single(CfgTree *self, LogExprNode *node,
           goto error;
         }
       pipe->flags |= PIF_INLINED;
+      g_ptr_array_add(self->initialized_pipes, pipe);
```
presidento: Why don't we need to increase the reference counter?

lib/cfg-tree.h
```diff
@@ -40,20 +40,20 @@ const gchar *log_expr_node_get_content_name(gint content);
 enum
 {
   /* expr node content type */
-  ENC_SOURCE,
```
presidento: Just a side note: I would create two different commit: one for the fix (the `.c` file), and one for the refactor (`.h` file).

lib/cfg-tree.c
presidento: We talked it IRL, I let the `g_ptr_array_add` in the original place, and call only `log_pipe_ref(pipe)` on line 468.

kira-syslogng:

kira-syslogng:

lib/cfg-tree.h
furiel: fixed

kira-syslogng:

lib/cfg-tree.h
```diff
@@ -40,20 +40,20 @@ const gchar *log_expr_node_get_content_name(gint content);
...
   /* expr node layouts type */
-  ENL_SINGLE,
-  ENL_REFERENCE,
-  ENL_SEQUENCE,
-  ENL_JUNCTION,
+  ENL_SINGLE, /* 7 */
+  ENL_REFERENCE, /* 8 */
+  ENL_SEQUENCE, /* 9 */
+  ENL_JUNCTION, /* 10 */
 };
```
bazsi: Why is this "numbering" useful? as I see it only adds maintenance burden, should we need to add new enums.

kira-syslogng:

kira-syslogng:

Issue #1400, opened 2017-03-28
Memory leak in syslog-ng 3.9.1


aghosh2739:
# syslog-ng

## Version of syslog-ng
syslog-ng 3.9.1
Installer-Version: 3.9.1
Revision: 
Module-Directory: /usr/lib64/syslog-ng
Module-Path: /usr/lib64/syslog-ng
Available-Modules: affile,afprog,kvformat,add-contextual-data,cef,date,cryptofuncs,pseudofile,linux-kmsg-format,syslogformat,csvparser,dbparser,disk-buffer,afuser,confgen,afstomp,graphite,afsocket,basicfuncs,system-source
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: off
Enable-TCP-Wrapper: off
Enable-Linux-Caps: off

## Platform

Linux x86-generic-64

# Issue
Memory leak in syslog-ng 3.9.1 
## Failure

==2104== Memcheck, a memory error detector
==2104== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==2104== Using Valgrind-3.10.1 and LibVEX; rerun with -h for copyright info
==2104== Command: syslog-ng -F --enable-core
==2104== Parent PID: 2103
==2104== 
--2104-- 
--2104-- Valgrind options:
--2104--    -v
--2104--    --log-file=valgrind.log
--2104--    --leak-check=full
--2104-- Contents of /proc/version:
--2104--   Linux version 3.10.104.cge-rt117 (ferret@bamboo) (gcc version 4.7.0 (MontaVista Linux G++ 4.7-170301090024) ) #1 SMP PREEMPT Thu Mar 9 08:50:32 IST 2017
--2104-- Arch and hwcaps: AMD64, LittleEndian, amd64-cx16-lzcnt-rdtscp-sse3-avx-avx2-bmi
--2104-- Page sizes: currently 4096, max supported 4096
--2104-- Valgrind library directory: /usr/lib64/valgrind
--2104-- Reading syms from /usr/sbin/syslog-ng
--2104--   Considering /usr/sbin/syslog-ng ..
--2104--   .. CRC mismatch (computed 6ecb1f66 wanted 2175c4e1)
--2104--    object doesn't have a symbol table
--2104-- Reading syms from /usr/lib64/valgrind/memcheck-amd64-linux
[...]

PR #975, opened 2016-03-05
F/improve deinit leaks mutex (add missing global deinitialization, Valgrind fixes)

bkil-syslogng:
We missed global deinitialization of various objects. These changes make the code more self contained and fix some Valgrind warnings as well.

PR #1649, opened 2017-08-21
memleak hunting: cleanup based on valgrind report

furiel:
This patchet intends to fix some entries in valgrind report. All of these issues are minors, i.e non of them accumulate over time. All of these changes intended to be cosmetic, explicitely calling deinits/frees, instead of just adding new exceptions to a valgrind suppress file.

Credits for @bkil-syslogng for https://github.com/balabit/syslog-ng/pull/975, which gave much inspiration for this pull request.
Only there are two commits in 975 that were not ported. I list them here for documentation reason.
-  g_process_set_argv_space. There were a lot of fight on these recently. The proposed change is still incomplete, see https://github.com/balabit/syslog-ng/pull/1528 for details.
- g_static_mutex_free. Syslog-ng mutexes are typically statically initialized. The glib documentation states that "[...] You don't have to call this functions for a GStaticMutex with an unbounded lifetime, i.e. objects declared 'static', [...]". In the glib implementation, when someone calls lock in the first time on such a static mutex, glib will allocate using malloc on the fly, but there is no automatic deallocation mechanism provided. As that malloc is not freed, valgrind reports it as still reachable memory. The suggested patch calls explicitely g_static_mutex_free-s on these objects. This should work, though I feel this change somewhat risky, because the documentation states we do not have to call such function. There is a small chance (very low, because we are talking about a deprecated mutex api to begin with ...) that some automatic deallocation would be introduced glib at some point. Especially, this leak comes from the design of glib, I don't want to fix it from syslog-ng side. That's why I did not pick that patch.

Edit:
<strike>The dns_cache part is reported in: https://github.com/balabit/syslog-ng/issues/1400</strike>
dns_cache will be handled in a separate pull request.

### DIFF
--- a/lib/stats/stats-control.c
+++ b/lib/stats/stats-control.c
@@ -49,3 +49,9 @@ stats_register_control_commands(void)
   control_register_command("STATS", NULL, control_connection_send_stats, NULL);
   control_register_command("RESET_STATS", NULL, control_connection_reset_stats, NULL);
 }
+
+void
+stats_unregister_control_commands(void)
+{
+  reset_control_command_list();
+}
--- a/contrib/valgrind/syslog-ng.supp
+++ b/contrib/valgrind/syslog-ng.supp
@@ -24,6 +24,14 @@
 
 ## Miscellaneous non-leaks and false alarms
 {
+   suppress_journald_open: see https://github.com/systemd/systemd/issues/6539
+   Memcheck:Leak
+   ...
+   fun:journald_open
+   ...
+}
+
+{
    suppress_syslog_ng_set_argv_space
    Memcheck:Leak
    fun:*alloc
@@ -48,17 +56,6 @@
    ...
    fun:app_startup
    fun:main
-}
-
-{
-    suppress_syslog_ng_path_resolver
-    Memcheck:Leak
-    fun:*alloc
-    ...
-    fun:path_resolver_new
-    ...
-    fun:main_loop_global_init
-    fun:main
 }
 
 {
@@ -105,15 +102,6 @@
 }
 
 {
-     suppress_syslog_ng_reloc
-     Memcheck:Leak
-     fun:*alloc
-     ...
-     fun:cache_*
-     fun:get_installation_path_for
-}
-
-{
      suppress_syslog_ng_msg_limit_internal_message
      Memcheck:Leak
      fun:*alloc
@@ -130,14 +118,6 @@
      ...
      fun:msg_event_create
      fun:main_loop_run
-}
-
-{
-     suppress_syslog_ng_aftiner_message_posted
-     Memcheck:Leak
-     fun:*alloc
-     ...
-     fun:afinter_message_posted
 }
 
 {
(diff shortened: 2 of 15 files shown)


kira-syslogng:

lib/dnscache.c
```diff
@@ -469,7 +471,12 @@ dns_caching_thread_deinit(void)
   g_assert(dns_cache != NULL);
 
   G_LOCK(unused_dns_caches);
-  unused_dns_caches = g_list_prepend(unused_dns_caches, dns_cache);
+
+  if (main_loop_is_terminating(main_loop_get_instance()))
+    dns_cache_free(dns_cache);
+  else
```
bazsi: hmm I don't really like this coupling between dnscache and mainloop. Why isn't the dns_caching_global_deinit() enough? That should cover the exit case properly.

lib/gprocess.c
```diff
@@ -1465,6 +1465,7 @@ g_process_finish(void)
 #endif
 
   g_process_remove_pidfile();
+  reloc_deinit(); /* path_cache needed for calculating pidfile location above */
 }
```
bazsi: I hate this one too. I understand that the pidfile location is actually managed by reloc, which then can't be deinitialized in app_shutdown() where it should be deinitialized, or otherwise we'd risk use-after-free, but adding this kind of coupling is worse than that minor, one-time memory leak.

I would either:
1) leave the leak as is
2) put reloc_deinit() into main()
3) use some kind of destructor trick to make sure that hash table gets freed

The easiest is probably 2), which plugs the leak and is acceptable code/coupling wise.

lib/mainloop.c
```diff
@@ -195,7 +195,7 @@ main_loop_initialize_state(GlobalConfig *cfg, const gchar *persist_filename)
   return success;
 }
 
-static inline gboolean
+gboolean
 main_loop_is_terminating(MainLoop *self)
```
bazsi: should not be needed.

lib/stats/stats-control.c
```diff
@@ -49,3 +49,11 @@ stats_register_control_commands(void)
   control_register_command("STATS", NULL, control_connection_send_stats, NULL);
   control_register_command("RESET_STATS", NULL, control_connection_reset_stats, NULL);
 }
+
+#include <stdio.h>
+
```
bazsi: why the stdio.h inclusion?

lib/dnscache.c
furiel: As mainloop does not wait threads to terminate, it can happen that `unused_dns_caches` is freed by the mainloop in `dns_caching_global_deinit` by the time` dns_caching_thread_deinit` is called. This results in an invalid read in the thread:
```
unused_dns_caches = g_list_prepend(unused_dns_caches, dns_cache);
```
as `unused_dns_caches` is already freed.

The leak here is only a side effect, due that `dns_cache` is appended to a non existing list.

bazsi: ugh, I understand. Still I would leave in this as a bug and then fix the termination issue in mainloop perhaps later. 

As far as I understand, the mainloop does have a grab on how many threads are executing, and we could wait for that to drop to 0, perhaps even using a timeout, to ensure that it does not take all eternity to exit.

If the timeout is reached, we would exit drastically (e.g. exit() without calling the shutdown functions), but that should only happen in extreme cases.

kira-syslogng:

lib/gprocess.c
furiel: Thanks, I'll go with 2).

lib/dnscache.c
furiel: Ok, fine for me. I drop this patch and try to create a separate one based on your guidance.

kira-syslogng:

kira-syslogng:

PR #1657, opened 2017-09-04
Register blocks and generators as plugins

bazsi:
This branch changes how config generators (e.g. user-defined blocks, the system() source and the confgen plugin) are implemented. Instead of using a lexer specific interface, this reuses the plugin interface. This is the list of notable changes, benefits that we get from this chunk of code:

**User visible changes:**
  * a new define "module-install-dir" has been added, that can be used to find syslog-ng's module installation directory (normally $prefix/lib/syslog-ng). All java based SCLs were converted to use this, instead of module-path, which might be a colon separated list of directories.
  * module auto-loading now also works for the system(), thus the explicit "@module system-source" statement in SCL has been removed. This will make it relatively easy to create other, auto-loadable plugins (such as the smart-parser I have in mind).

**Code changes**
  * instead of storing plugin framework related fields in GlobalConfig, do that in a separate struct called PluginContext, which is then used everywhere in the plugin code, making it independent of the config where it is embedded into. As a result, cfg.h depends on plugin.h but not vica versa.
  * CfgBlockGenerator has become a proper class, instead of a data struct with a function callback. CfgBlock (e.g. the user-defined block) became a descendant class, and the same happened with the system() source and the confgen module
  * plugin "preference" has been dropped.  It has only been used by the afsocket module when openssl was not an allowed dependency.
  * plugin auto discovery now happens in a separate process, to avoid loading all modules at startup

**Bugfixes:**
  * restores "@define module-path" functionality, which was broken by 484e591e4d 

**Other, unrelated fixes:**
  * fixed a use of uninitialized memory in debugger


kira-syslogng:

kira-syslogng:

bazsi:
I have now rebased to master and travis issues should be fixed up. Could probably go in after the next release.

bazsi:
btw: I have a subsequent series of patches that builds on this one.

kira-syslogng:

kira-syslogng:

kira-syslogng:

bazsi:
Rebased again, as a new plugin was merged into master, which broke this after rebase.

kira-syslogng:

lbudai:
@bazsi: I promise that after the release this PR will be the first :)

lib/cfg.c
```diff
@@ -346,6 +346,8 @@ cfg_new_snippet(gint version)
   stats_options_defaults(&self->stats_options);
 
   cfg_tree_init_instance(&self->tree, self);
+  plugin_context_init_instance(&self->plugin_context);
```
furiel: I think this init is not necessary here as g_new0 will zero it out by default.

lib/debugger/debugger.c
```diff
@@ -106,6 +106,10 @@ _display_source_line(LogExprNode *expr_node)
         buf[0] = 0;
       fclose(f);
     }
+  else
+    {
+      buf[0] = 0;
```
furiel: Two lines below we have buf[strlen(buf) - 1, which will read an invalid position (buf[-1]) in this case. Maybe we should return instead.

bazsi: you are right, although the same can happen if the last line is zero length, without an NL, so that's basically a 2nd bug.

lib/cfg.c
```diff
@@ -362,6 +362,7 @@ cfg_set_global_paths(GlobalConfig *self)
   cfg_args_set(self->lexer->globals, "syslog-ng-include", get_installation_path_for(SYSLOG_NG_PATH_CONFIG_INCLUDEDIR));
   cfg_args_set(self->lexer->globals, "scl-root", get_installation_path_for(SYSLOG_NG_PATH_SCLDIR));
   cfg_args_set(self->lexer->globals, "module-path", resolvedConfigurablePaths.initial_module_path);
+  cfg_args_set(self->lexer->globals, "module-install-dir", resolvedConfigurablePaths.initial_module_path);
```
furiel: Just a note, because I know it out of scope, but this resolvedConfigurablePaths variable leads to a very inconvenient dependency. For instance this resolvedConfigurablePaths.initial_module_path was initialized with get_installation_path_for(SYSLOG_NG_MODULE_PATH) during start. Which in terms builds a cache object, and store the calculated data there. In the end, the same gchar * will be stored inside the cache, inside resolvedConfigurablePaths. If I followed correctly, luckily you strdup in PluginContext, so the dependency stops here, but still it is not safe to store the same pointer in two places. At least for me it was problematic to find out when to free the cache object for reloc due to this.
Either the cache mechanism or resolvedConfigurablePaths should be eliminated imo.

lib/cfg.c
furiel: I see this init is extended with other things than memset in a later commit. My comment was intended only on the memset part.

bazsi:
This reintroduces bug #1423, which still needs to be fixed.

lib/plugin.h
```diff
@@ -29,6 +29,7 @@
 
 typedef struct _Plugin Plugin;
 typedef struct _ModuleInfo ModuleInfo;
+typedef struct _PluginContext PluginContext;
```
MrAnno: First definition

lib/plugin.h
```diff
@@ -85,26 +89,32 @@ struct _ModuleInfo
...
   gint plugins_len;
   /* the higher the better */
-  gint preference;
 };
 
-/* instantiate a new plugin */
-Plugin *plugin_find(GlobalConfig *cfg, gint plugin_type, const gchar *plugin_name);
-gpointer plugin_construct(Plugin *self, GlobalConfig *cfg, gint plugin_type, const gchar *plugin_name);
-gpointer plugin_parse_config(Plugin *plugin, GlobalConfig *cfg, YYLTYPE *yylloc, gpointer arg);
+typedef struct _PluginContext PluginContext;
```
MrAnno: Second definition

lib/syslog-ng.h
```diff
@@ -47,6 +47,8 @@
 typedef struct _LogPipe LogPipe;
 typedef struct _LogMessage LogMessage;
 typedef struct _GlobalConfig GlobalConfig;
+typedef struct _CfgLexer CfgLexer;
+typedef struct _PluginContext PluginContext;
```
MrAnno: Third definition

kira-syslogng:

lib/debugger/debugger.c
bazsi: I've added a fix for this case as well.

lib/cfg.c
bazsi: The original idea behind "resolvedConfigurablePaths" is that they give us a stable pointer for the entire execution of the main binary, that was the whole idea behind the relocation.h API. I know that we considered these leaks and freed them, but the code randomly assumes that these would be good enough even for storing them directly.

Anyway, this case is independent, cfg_args_set() strdups its argument, so it is not stored as a pointer, but rather it is copied.

So I consider this also to be ok.

kira-syslogng:

lib/syslog-ng.h
bazsi: I have removed the extra ones, thanks for noticing.

bazsi:
I have now fixed all review comments, so this should be good to go.

kira-syslogng:

lib/plugin.h
```diff
@@ -67,11 +67,14 @@ struct _Plugin
...
-  void (*setup_context)(Plugin *self, GlobalConfig *cfg, gint plugin_type, const gchar *plugin_name);
-  gpointer (*construct)(Plugin *self, GlobalConfig *cfg, gint plugin_type, const gchar *plugin_name);
+  gpointer (*construct)(Plugin *self);
   void (*free_fn)(Plugin *s);
 };
 
+gpointer plugin_construct(Plugin *self);
+gpointer plugin_construct_from_config(Plugin *self, CfgLexer *lexer, gpointer arg);
+
+typedef struct _ModuleInfo ModuleInfo;
```
MrAnno: Sorry for bothering you again, but this is also a redefinition. The first one is at line 31.

lib/plugin.h
```diff
@@ -59,6 +58,7 @@ typedef struct _PluginCandidate
  * contrast with the "module" which is the shared object itself which
  * registers plugins.  Each module can register a number of plugins,
  * not just one.  */
+typedef struct _Plugin Plugin;
```
MrAnno: Redefinition. The first one is at line 30.

lib/plugin.h
bazsi: I have now removed these. I think they got here because a bogus conflict resolution.

btw: I have now tried to compile gcc-4.7 and it still compiles the multiple typedefs. How far do you need to go in gcc archeology to not support this?

I have even tried with -std=c99 but that didn't compile ivykis, gnu90 does compile the typedefs.

anyway they shouldn't be there, so the gcc version is probably not that important

MrAnno: We tried to compile this with gcc (Ubuntu 4.4.3-4ubuntu5) 4.4.3 (it is the linux-glibc2.11 builder).

This version of gcc fails without using any flags (no pedantic or std=c99 needed).

MrAnno:
```
$ gcc --version
gcc (Ubuntu 4.4.3-4ubuntu5) 4.4.3

$ cat test.c 
typedef struct Test Test;
typedef struct Test Test;

int main() {}

$ gcc -Wall test.c                                                                                                                                 
test.c:2: error: redefinition of typedef 'Test'
test.c:1: note: previous declaration of 'Test' was here
```

kira-syslogng:

lib/plugin.h
MrAnno: Fixed. Thank you.

MrAnno:
@bazsi Can I push a minor fix here?

An old `plugin_load_module()` call remained here:
https://github.com/balabit/syslog-ng/blob/710d682c4f1400ccacf832022556a2ecc01f1109/lib/cfg-lexer.c#L937

kira-syslogng:

kira-syslogng:

MrAnno:
During the rebase, 785bd2aee62d3fe0ad12fa9f241cfd32ae66d447 has been removed from the PR.

@bazsi Can I add this patch back to this branch?

MrAnno:
During the rebase, https://github.com/balabit/syslog-ng/commit/785bd2aee62d3fe0ad12fa9f241cfd32ae66d447 has been removed from the PR.

@bazsi Can I add it patch back to this branch?

lbudai:
@MrAnno: you are right, I'll add this patch and then merge the PR.

kira-syslogng:

lbudai:
Travis has some problems... (style-checker job is blocking)...
I run style checker manually and it does not report any issues so I'll merge this.

PR #1659, opened 2017-09-05
xml parser

furiel:
The xml parser can process input in xml format, and adds the parsed data to the message object.

Configuration example
```
parser xml_parser {
       xml(template("$MSG")
           prefix(".xml")
           drop_invalid(no)
           exclude-tags("tag1", "tag2", "tag3?", "tag*")
           strip-whitespaces(no)
          );
};
```
Output example

From the following xml:
```
<tags attr='attrval'>part1<tag1>Tag1 Leaf</tag1>part2<tag2>Tag2 Leaf</tag2>part3</tags>
```
the following output is generated: (through json-parse template, ignoring exclude-tags now)
```
{"_xml":{"tags":{"tag2":"Tag2 Leaf","tag1":"Tag1 Leaf","_attr":"attrval","tags":"part1part2part3"}}}
```


kira-syslogng:

kira-syslogng:

modules/xml/xml.c
```diff
@@ -0,0 +1,338 @@
...
+
+void
+xml_parser_set_exclude_tags(LogParser *s, GList *exclude_tags)
+{
+  XMLParser *self = (XMLParser *) s;
+
+  g_list_free_full(self->exclude_tags, g_free);
+  self->exclude_tags = g_list_copy(exclude_tags);
+
+  /* g_list_copy_deep would be better, but not supported on all glib versions :( */
```
Kokan: Would it be possible to implement g_list_copy_deep under a compiler switch (do we have g_list_copy_deep or not) ?
In this case the code would document this comment :)

kira-syslogng:

modules/xml/xml.c
furiel: Thanks for the comment! I backported a g_list_copy_deep implementation to use on machines where it is not available. It is less efficient than the one provided by glib master, because this version iterates twice through the list instead of one. But it does not use any internal api from the glib list implementation (like memory allocation), so I felt safer to go with this version.

kira-syslogng:

kira-syslogng:

modules/xml/xml.c
```diff
@@ -0,0 +1,330 @@
...
+joker_or_wildcard(GList *patterns)
+{
+  gboolean retval = FALSE;
+  GList *pattern = patterns;
+  while (pattern != NULL)
+    {
+      gchar *str = ((gchar *)pattern->data);
+      if (strchr(str, '*') || strchr(str, '?'))
+        {
+          retval = TRUE;
```
Kokan: Would it be possible to return right away instead of iterate one more time ?

modules/xml/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+)
+
+generate_y_from_ym(modules/xml/xml-grammar)
+
+bison_target(xmlGrammar
+    ${CMAKE_CURRENT_BINARY_DIR}/xml-grammar.y
+    ${CMAKE_CURRENT_BINARY_DIR}/xml-grammar.c
+COMPILE_FLAGS ${BISON_FLAGS})
+
+option(ENABLE_xml "Enable xml ON")
```
Kokan: Could we use for the name of the option variable upper case ?

Kokan: The description is confusing, because it says "Enable xml ON", but the default value for `option` is *OFF*. It is not stated here otherwise. I would follow the pattern:

```cmake
option(ENABLE_XML, "Enable xml module", ON)
```

modules/xml/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+
+if (ENABLE_xml)
```
Kokan: Personally I would exclude the whole file with the condition, not just the compiling and linking part.

```cmake
option(ENABLE_BLA)

if (ENABLE_BLA)
 everything here
endif()
```

The advantage of this that people are used to search include guards there.

modules/xml/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+set(xml_HEADERS
+    "xml-parser.h"
+    "${CMAKE_CURRENT_BINARY_DIR}/xml-grammar.h"
+)
+
+set(xml_SOURCES
```
Kokan: You are missing the **xml.c** here.

modules/xml/tests/test_xml_parser.c
```diff
@@ -0,0 +1,244 @@
...
+{
+  static ValidXMLTestCase test_cases[] =
+  {
+    {"<tag1>value1</tag1>", ".xml.tag1", "value1"},
+    {"<tag1 attr='attr_value'>value1</tag1>", ".xml.tag1._attr", "attr_value"},
+    {"<tag1><tag2>value2</tag2></tag1>", ".xml.tag1.tag2", "value2"},
+    {"<tag1>part1<tag2>value2</tag2>part2</tag1>", ".xml.tag1", "part1part2"},
+    {"<tag1><tag11></tag11><tag12><tag121>value</tag121></tag12></tag1>", ".xml.tag1.tag12.tag121", "value"},
+    {"<tag1><tag11></tag11><tag12><tag121 attr1='1' attr2='2'>value</tag121></tag12></tag1>", ".xml.tag1.tag12.tag121._attr1", "1"},
+    {"<tag1><tag11></tag11><tag12><tag121 attr1='1' attr2='2'>value</tag121></tag12></tag1>", ".xml.tag1.tag12.tag121._attr2", "2"},
```
Kokan: I was thinking about the following case:

```xml
<tag>
   <tag1>value11</tag1>
   <tag2>value22</tag1>
</tag>
```

What should be the result of `tag.tag1` ?

modules/xml/xml-grammar.ym
```diff
@@ -0,0 +1,92 @@
...
+  ;
+
+xml_options
+        : xml_opt xml_options
+        |
+        ;
+
+xml_opt
+        : KW_PREFIX '(' string ')'      { xml_parser_set_prefix(last_parser, $3); free($3); }
+        | KW_DROP_INVALID '(' yesno ')' { xml_parser_set_forward_invalid(((XMLParser *) last_parser), !$3); };
```
furiel: Signature must use LogParser instead of XMLParser to reduce dependency from grammar towards module.

lib/compat/Makefile.am
```diff
@@ -17,6 +17,7 @@ compat_sources			= 	\
 	lib/compat/inet_aton.c		\
 	lib/compat/memrchr.c		\
 	lib/compat/pio.c		\
+	lib/compat/glib.c		\
```
Kokan: You should add `glib.c` also to the `lib/compat/CMakeLists.txt`.

modules/xml/xml.c
```diff
@@ -0,0 +1,330 @@
...
+        gpointer             user_data,
+        GError             **error)
+{
+  if (text_len == 0)
+    return;
+
+  InserterState *state = (InserterState *)user_data;
+  const gchar *current_value = log_msg_get_value_by_name(state->msg, state->key->str, NULL);
+  GString *value = scratch_buffers_alloc();
+  g_string_printf(value, "%s%s", current_value, text);
```
furiel: Text is _not_ nullterminated here.

furiel:
Added a strip_whitespaces() option. As xmllint also does not strip whitespaces, the default value is false in syslog-ng as well.
When a user creates an xml by hand and adds indentation for cosmetic purposes, the (significant) enters, tabs, spaces are maintained in the nvtable by default. But such whitespaces in the result might seem awkward in some cases, so this option can strip whitespaces if the user sets it.

kira-syslogng:

lib/compat/Makefile.am
furiel: Fixed

kira-syslogng:

kira-syslogng:

modules/xml/xml.h
```diff
@@ -0,0 +1,47 @@
...
+#define XML_H_INCLUDED
+
+#include "parser/parser-expr.h"
+
+typedef struct
+{
+  LogParser super;
+  gchar *prefix;
+  gboolean forward_invalid;
+  GList *exclude_tags;
```
furiel: exclude_tags should be extracted to an XMLParserOptions object

furiel: Both exclude_tags and exclude patterns are not needed. The exclude_tags set function can compile right away, there is no need to store exclude_tags.

bazsi:
Please try not to use sprintf() and friends (g_string_sprintf qualifies) as that ruins performance a lot.

modules/xml/xml.c
```diff
@@ -0,0 +1,376 @@
...
+  while (names[attrs])
+    {
+      attr_key = g_string_overwrite(attr_key, base_index, names[attrs]);
+      log_msg_set_value_by_name(msg, attr_key->str, values[attrs], -1);
+      attrs++;
+    }
+}
+
+static gboolean
+tag_matches_patterns(const gchar *element_name, const GPtrArray *patterns, gboolean should_reverse)
```
lbudai: What if we pass the `reversed_element_name` also? In this case we don't have to propagate this flag to here (and this decision already made by the caller...). On the other hand in this case it is better to pass also the length of the `element_name` (as the caller also has to compute it because of the reverse).

modules/xml/xml.c
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+  g_string_append_printf(state->key, ".%s", element_name);
+  msg_add_attributes(state->msg, state->key, attribute_names, attribute_values);
+}
+
+static gint
+scan_for_last_dot(GString *str)
+{
+  gint i = str->len;
```
lbudai: Use `strrchr` instead. (`pos = strchr(s, '.'); return (pos-s+1);` IIRC

kira-syslogng:

modules/xml/CMakeLists.txt
furiel: We decided as this is non-critical issue and general in other modules, this will not be scope of this patch.

furiel: We decided as this is non-critical issue and general in other modules, this will not be scope of this patch.

modules/xml/CMakeLists.txt
furiel: We decided as this is non-critical issue and general in other modules, this will not be scope of this patch.

modules/xml/tests/test_xml_parser.c
furiel: There is a drop_invalid() with default value: no. User can control this behaviour.

For the first part: if we write tag1 instead of tag2 as well:
```
<tag>
   <tag1>value11</tag1>
   <tag1>value22</tag1>
</tag>
```
the result will be concatenated. The same behaviour is visible with xmllint:
$ xmllint --xpath "/tag/tag1/text()" text.xml
value11value22%
$
It is currently a limitation of the parser users cannot access the individual tags in such vector-like structures. This limitation will be documented in the administrator's guide.

modules/xml/tests/test_xml_parser.c
```diff
@@ -0,0 +1,270 @@
...
+} ValidXMLTestCase;
+
+ParameterizedTestParameters(xmlparser, valid_inputs)
+{
+  static ValidXMLTestCase test_cases[] =
+  {
+    {"<tag1>value1</tag1>", ".xml.tag1", "value1"},
+    {"<tag1 attr='attr_value'>value1</tag1>", ".xml.tag1._attr", "attr_value"},
+    {"<tag1><tag2>value2</tag2></tag1>", ".xml.tag1.tag2", "value2"},
+    {"<tag1>part1<tag2>value2</tag2>part2</tag1>", ".xml.tag1", "part1part2"},
```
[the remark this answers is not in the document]
furiel: These are two different xml-s, but they will result in the same keys values.
This behaviour aligns with xmllint:
```
$ diff -s <(xmllint --xpath "/tag1/text()" <(echo "<tag1>part1<tag2>value2</tag2>part2</tag1>")) <(xmllint --xpath "/tag1/text()" <(echo "<tag1>part1part2<tag2>value2</tag2></tag1>"))
Files /proc/self/fd/11 and /proc/self/fd/13 are identical
```

modules/xml/xml-grammar.ym
```diff
@@ -0,0 +1,94 @@
...
+  ;
+
+xml_options
+        : xml_opt xml_options
+        |
+        ;
+
+xml_opt
+        : KW_PREFIX '(' string ')'      { xml_parser_set_prefix(last_parser, $3); free($3); }
+        | KW_DROP_INVALID '(' yesno ')' { xml_parser_set_forward_invalid(last_parser, !$3); };
```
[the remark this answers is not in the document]
furiel: There is a similar problem/behaviour/option in dbparser, where we added KW_DROP_UNMATCHED. I wanted to align my option with that. That's why I used drop invalid here. Probably subjective feeling, but I think in case of an invalid input, the term 'drop' comes earlier to one's mind than 'forward'.
From coding point of view however, our internal api (parser_process) thinks in "forwarding", so it was easier to work with the positive version it internally. So I use this negation between the user api and the code internals.

modules/xml/tests/test_xml_parser.c
```diff
@@ -0,0 +1,270 @@
...
+{
+  const gchar *input;
+  const gchar *key;
+  const gchar *value;
+} ValidXMLTestCase;
+
+ParameterizedTestParameters(xmlparser, valid_inputs)
+{
+  static ValidXMLTestCase test_cases[] =
+  {
```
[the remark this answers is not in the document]
furiel: I need some suggeston on the complex XMLs, let's talk irl.

On the other hand, I have a single negative test:
Test(xml_parser, test_drop_invalid) 
which tests a single unbalanced tag. I can add a list with negative cases, but I am not sure on its benefit. Those would only test the glib's xml parser implementation, which we implicitely think (work hypothesis) it works. The integration is tested with test_drop_invalid.

modules/xml/xml-grammar.ym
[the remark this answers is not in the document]
lbudai: It's true, but if the default value is yes, then a user want to set this option only if she or he wants to forward.

modules/xml/xml.c
furiel: Very good comment thanks! Especially again, element_name is _NOT_ null terminated, so I should not have called strlen() on it inside this function. I will fix this.

modules/xml/xml.c
```diff
@@ -0,0 +1,376 @@
...
+  if (value)
+    log_msg_set_value_by_name(state->msg, state->key->str, value->str, value->len);
+}
+
+
+static GMarkupParser xml_scanner =
+{
+  .start_element = start_element_cb,
+  .end_element = end_element_cb,
+  .text = text_cb
```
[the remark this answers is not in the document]
furiel: - error
g_markup_parse_context_parse and g_markup_parse_context_end_parse also reports back a gerror, which I turn into an msg_error. For the invalid xml-s I passed, I could see the error logs. So I did not override the method. Did I overlook something and should I override? 

- passthrough
Comments and processing instructions should not be part of the data imo so I did not override this virtual function. By the way xmllint also strips these data from text. For example
```
$xmllint --xpath "/tag/text()" <(echo "<tag>value<?mso-application progid="Excel.Sheet"?></tag>")
value%
```

kira-syslogng:

modules/xml/tests/test_xml_parser.c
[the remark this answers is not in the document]
furiel: In the end we decided to add some failure testcase.

modules/xml/xml-grammar.ym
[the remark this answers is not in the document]
furiel: We decided to have drop_invalid: no as default value.

furiel:
Replaced the g_string_printf functions with a series of g_string_assign, g_string_append and g_string_append_c.

furiel:
This version will not support CDATA. We will document it as a limitation.

kira-syslogng:

kira-syslogng:

PR #1632, opened 2017-08-01
python-dest: string compatibility for python3

furiel:
Two changes in python3 c api compared to python2:
1) In python3, all strings are utf8, so PyBytes_Check will fail. This
patch will use PyUnicode_Check for python3 instead.
2) The function PyBytes_AsString is removed in python3. This patch
will use PyUnicode_AsUTF8 instead.

Fixes: https://github.com/balabit/syslog-ng/issues/1363

PR #1660, opened 2017-09-07
python: regression in python version auto-detection

furiel:
Introduced in: https://github.com/balabit/syslog-ng/pull/1632
A distribution might not label python pkg-config pc file as python2 or
python3. In such case auto-detection will not work.

In case of autodetection, we will look for pc files in this order:
python3, python2, python. This will work on even archlinux where only
python2 or python3 is used.

The decision between v2 and v3 will be done later by a call to
AC_CHECK_FUNCS instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1310,13 +1310,11 @@ dnl python checks
 dnl ***************************************************************************
 if test "x$enable_python" != "xno"; then
   if test "x$with_python" = "xauto"; then
-    PKG_CHECK_MODULES(PYTHON, python3 >= 3, python_found="yes", python_found="no")
-    if test "$python_found" = "yes"; then
-      with_python="python3"
-    else
-      PKG_CHECK_MODULES(PYTHON, python2 >= 2.6, python_found="yes", python_found="no")
-      if test "$python_found" = "yes"; then
-        with_python="python2"
+    PKG_CHECK_MODULES(PYTHON, python3, python_found="yes", python_found="no")
+    if test "$python_found" = "no"; then
+      PKG_CHECK_MODULES(PYTHON, python2, python_found="yes", python_found="no")
+      if test "$python_found" = "no"; then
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -1310,11 +1310,11 @@ dnl python checks
 dnl ***************************************************************************
 if test "x$enable_python" != "xno"; then
   if test "x$with_python" = "xauto"; then
-    PKG_CHECK_MODULES(PYTHON, python3 >= 3, python_found="yes", python_found="no")
+    PKG_CHECK_MODULES(PYTHON, python >= 3, python_found="yes", python_found="no")
```
presidento: I suggest using `python3` for checking Python 3 existence and use `python` for test Python 2.

The [Ubuntu documentation](https://wiki.ubuntu.com/Python/3) says that making Python 3 the default **does not mean**:
- `/usr/bin/python` will point to Python 3. **No, this is not going to happen** (unless [PEP 394](https://www.python.org/dev/peps/pep-0394/) advocates otherwise, which is doubtful for the foreseeable future). `/usr/bin/python` and `/usr/bin/python2` will point to Python 2.7 and `/usr/bin/python3` will point to the latest supported Python 3 version.

kira-syslogng:

kira-syslogng:

configure.ac
MrAnno: `PKG_CHECK_MODULES` checks for the name of the corresponding `.pc`, not for the binary.
This can vary on different platform (for example: Arch Linux uses `python2` and `python3`, Ubuntu has a `python` alias for `python2`. I would check both `python{2,3}` and `python` as well.


Arch:
```
$ pkg-config --modversion python
Package python was not found in the pkg-config search path. 

$ pkg-config --modversion python3
3.6
```

furiel: Thank you both for your comments. I updated accordingly, both ubuntu and arch is supported now.

kira-syslogng:

Issue #1509, opened 2017-05-30
flush-timeout() not implemented in elasticsearch2 destination


sergehauser:
# syslog-ng
## Version of syslog-ng
syslog-ng 3.9.1
Installer-Version: 3.9.1
Revision: 
Module-Directory: //usr/lib64/syslog-ng
Module-Path: //usr/lib64/syslog-ng
Available-Modules: linux-kmsg-format,afstomp,confgen,dbparser,basicfuncs,afsocket,disk-buffer,date,sdjournal,cryptofuncs,afuser,graphite,afprog,pseudofile,csvparser,syslogformat,afamqp,affile,add-contextual-data,kvformat,json-plugin,cef,system-source
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on

## Platform
CentOS 7
Linux saffa 3.10.0-514.16.1.el7.x86_64 #1 SMP Wed Apr 12 15:04:24 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

# Issue
flush-timeout is not implemented for the elasticsearch2 destination.

## Failure
When i have a flush-limit set, syslog-ng will wait forever (or till limit of messages is reached), before sending the logs to the ES server. The only way to send the messages in the buffer to ES is to stop (or probably SIGNAL) syslog-ng

## Steps to reproduce
- set flush-limit
- send less messages
- wait, messages will not arrive

## Configuration
Relevant part of my config:

```
destination d_elastic_local {
        elasticsearch2(
                index("my-logs-${YEAR}.${MONTH}.${DAY}")
                type("my-log")
                cluster("elasticsearch")
                template("$(format-json --scope nv-pairs --exclude 0*,1*,2*,3*,4*,5*,6*,7*,8*,9*,A*,B*,C*,D*,E*,F*,G*,H*,I*,J*,K*,L*,N*,M*,O*,P*,Q*,R*,S*,T*,U*,V*,W*,X*,Y*,Z* @timestamp=${ISODATE})")
                client-lib-dir("/usr/share/elasticsearch/lib")
                flush-limit(2000)
                client-mode("http")
                cluster-url("http://localhost:9200")
        );
};
```

faxm0dem:
I can confirm this behaviour, and indeed it can be a problem on low-traffic instances.
[...]

PR #1661, opened 2017-09-08
elasticsearch2.http: use flush_timeout() # FIXES 1509

pzoleex:


pzoleex:
this fixes https://github.com/balabit/syslog-ng/issues/1509

kira-syslogng:

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/messageprocessor/http/HttpBulkMessageProcessor.java
```diff
@@ -69,7 +78,31 @@ public boolean send(Index index) {
...
 		return true;
 	}
+
+    private long getCurrentUnixEpochSecond() {
+        return System.currentTimeMillis() / 1000L;
+    }
+
+    private void startFlushTimer() {
+        timer = new Timer();
+        timer.scheduleAtFixedRate(new TimerTask() {
```
furiel: This makes the flush-timeout unpredictable, isn't it? This task is executed each second periodically. In case of the last message arrived in the early part of the period, flush would take almost an extra second to be triggered. Can this be a problem? Probably rescheduling a task after each sent message would be more deterministic.

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/ElasticSearchOptions.java
```diff
@@ -85,6 +86,7 @@
 	public static String CLIENT_MODE_DEFAULT = CLIENT_MODE_TRANSPORT;
 	public static String HTTP_AUTH_TYPE_DEFAULT = HTTP_AUTH_TYPE_NONE;
 	public static String CONCURRENT_REQUESTS_DEFAULT = "1";
+	public static String FLUSH_TIMEOUT_DEFAULT = "1";
```
furiel: Shouldn't this be an integer type?

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/ElasticSearchOptions.java
```diff
@@ -35,212 +35,228 @@
...
         }
+    };
 
-        public boolean getSkipClusterHealthCheck() {
-                return options.get(SKIP_CLUSTER_HEALTH_CHECK).getValueAsBoolean();
-        }
+    public static String CLIENT_MODE_DEFAULT = CLIENT_MODE_TRANSPORT;
+    public static String HTTP_AUTH_TYPE_DEFAULT = HTTP_AUTH_TYPE_NONE;
+    public static String CONCURRENT_REQUESTS_DEFAULT = "1";
+    public static String FLUSH_TIMEOUT_DEFAULT = "1";
```
furiel: Hmm somehow github shows my comment outdated earlier. Trying again....
Shouldn't this one be integer?

pzoleex: No, here is everything is string and it is converted later.

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/messageprocessor/http/HttpBulkMessageProcessor.java
pzoleex: it means the flush will happen between flush_timout and flush_timeout+1. I don't think this can be a problem.

The problem is that in java there is no way to reschedule a task. You have to cancel it then recreate. Now if syslog-ng processes 20k msg/sec, you should do the timer cancel+recreate 20k/sec that is waste of resources (everything is very expensive that we do per message).

My solution will spare the resources against a theoretically very low delaying.

bazsi:
With other destinations we basically started to ignore flush-timeout(),
e.g. we flush if there's nothing in the queue. that behaviour is better and
gets rid of an option and complexity. isn't it?



-- 
Bazsi

pzoleex:
@bazsi Yes, we discussed it, flush when the queue is empty will be the solution

pzoleex:
so I close it, instead https://github.com/balabit/syslog-ng/pull/1668 will fix the issue

PR #1666, opened 2017-09-11
dns-cache: during shutdown: possible use after free

furiel:
During shutdown, mainloop did not wait worker threads to stop: app_shutdown and app_thread_stop was racing. This could lead to a use after free issue in the code: a dns_cache object could be returned to an already unallocated pool: unused_dns_caches.

Two possible resolutions to the problem:
- Reference counting for unused_dns_caches.
- Blocking mainloop until all workers exit.

This patchset contains both versions. I would like to open up a discussion if we need either one or both. 

Me seems reference counting is simpler in terms of code complexity, and has less possible side effects comparing to the version of blocking mainloop. Though the latter one could eliminate the currently unknown or possible future race conditions of these functions as well.

I could only manually test the blocking mainloop version by adding some printf-s to the beginning and the end of the racing functions, and see if the debug-log sandwich form turns into a disjoint form. In the end, leaving these debug logs felt awkward in the code, so I removed them in the final version. Feel free to indicate if I should put them back, or if anyone has any more reasonable idea to test such thing.

The reference counting is tested only manually as well: in a simple file source file destination configuration, sending a single message created a thread with a dns_cache, and stopping syslog-ng instantly after that caused an entry in the valgrind report. On my development machine, unused_dns_caches was steadily freed before the dns_cache was appended to the list so reproducing was not a problem.

This ticket intends to resolve the dns_cache part of the two issues reported in: https://github.com/balabit/syslog-ng/issues/1400

### DIFF
--- a/lib/mainloop.c
+++ b/lib/mainloop.c
@@ -89,6 +89,8 @@
  */
 
 ThreadId main_thread_handle;
+GCond thread_halt_cond;
+GMutex workers_running_lock = G_STATIC_MUTEX_INIT;
 
 struct _MainLoop
 {
@@ -295,6 +297,25 @@ main_loop_reload_config_initiate(gpointer user_data)
 
   is_reloading_scheduled = TRUE;
   main_loop_worker_sync_call(main_loop_reload_config_apply, self);
+}
+
+static void
+block_till_workers_exit()
+{
+  gint64 end_time;
+  end_time = g_get_monotonic_time() + 15*G_TIME_SPAN_SECOND;
+
+  g_mutex_lock(&workers_running_lock);
+  while (main_loop_workers_running)
+    if (!g_cond_wait_until(&thread_halt_cond, &workers_running_lock, end_time))
+      {
+        /* timeout has passed. */
+        fprintf(stderr, "Main thread timed out (15s) while waiting workers threads to exit. "
+                "workers_running: %d. Continuing ...\n", main_loop_workers_running);
+        break;
+      }
+
+  g_mutex_unlock (&workers_running_lock);
 }
 
 /************************************************************************************
@@ -511,6 +532,7 @@ main_loop_deinit(MainLoop *self)
   main_loop_call_deinit();
   main_loop_io_worker_deinit();
   main_loop_worker_deinit();
+  block_till_workers_exit();
   scratch_buffers_automatic_gc_deinit();
 }
 
--- a/lib/mainloop-worker.c
+++ b/lib/mainloop-worker.c
@@ -58,7 +58,7 @@ volatile gboolean main_loop_workers_quit;
 volatile gboolean is_reloading_scheduled;
 
 /* number of I/O worker jobs running */
-static gint main_loop_workers_running;
+static gint main_loop_jobs_running;
 
 static struct iv_task main_loop_workers_reenable_jobs_task;
 
@@ -202,7 +202,7 @@ main_loop_worker_job_start(void)
 {
   main_loop_assert_main_thread();
 
-  main_loop_workers_running++;
+  main_loop_jobs_running++;
 }
 
 typedef struct
@@ -250,8 +250,8 @@ main_loop_worker_job_complete(void)
 {
   main_loop_assert_main_thread();
 
-  main_loop_workers_running--;
-  if (main_loop_workers_quit && main_loop_workers_running == 0)
+  main_loop_jobs_running--;
+  if (main_loop_workers_quit && main_loop_jobs_running == 0)
     {
       /* NOTE: we can't reenable I/O jobs by setting
        * main_loop_io_workers_quit to FALSE right here, because a task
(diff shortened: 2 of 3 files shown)


kira-syslogng:

furiel:
One thought to add:
It is not enough for mainloop to wait for worker threads to quit. Due the worker threads have a 10 seconds idle timeout before they quit, it can happen that main thread should wait that 10 seconds as well. So the quit must be early-triggered in the worker threads before the block kicks in. That's why I put the blocking call only after iv_work_pool_put(&main_loop_io_workers);

kira-syslogng:

lib/dnscache.c
```diff
@@ -419,6 +419,31 @@ TLS_BLOCK_END;
 static DNSCacheOptions effective_dns_cache_options;
 G_LOCK_DEFINE_STATIC(unused_dns_caches);
 static GList *unused_dns_caches;
+volatile gint unused_dns_caches_ref_count;
+
```
bazsi: I think it would make sense to encapsulate this global state into a struct and add the refcounter for that struct.

lib/apphook.c
```diff
@@ -215,4 +216,6 @@ app_thread_stop(void)
   main_loop_call_thread_deinit();
   dns_caching_thread_deinit();
   scratch_buffers_allocator_deinit();
+  g_atomic_int_dec_and_test(&main_loop_workers_running);
+  g_cond_signal(&thread_halt_cond);
 }
```
bazsi: I think this should probably go the call-sites where app_thread_start/stop is getting called.

lib/dnscache.c
bazsi: e.g. "DNSCaching" could be a name for this struct, as the functions use this as a prefix.

lib/mainloop.c
```diff
@@ -297,6 +298,27 @@ main_loop_reload_config_initiate(gpointer user_data)
   main_loop_worker_sync_call(main_loop_reload_config_apply, self);
 }
 
+static void
+block_till_workers_exit()
+{
+  gint64 end_time;
```
bazsi: are the worker threads signaled to exit at this point? e.g. IIRC iv_work_pool_put() should wake up the workers and cause them to exit in a timely manner.

our threaded destination code might be blocking for some time still, but I think that's ok.

lib/mainloop.c
```diff
@@ -297,6 +298,27 @@ main_loop_reload_config_initiate(gpointer user_data)
...
+  static GMutex dummy_workers_running_lock = G_STATIC_MUTEX_INIT;
+
+  end_time = g_get_monotonic_time() + 15*G_TIME_SPAN_SECOND;
+  g_mutex_lock(&dummy_workers_running_lock);
+  while (main_loop_workers_running)
+    if (!g_cond_wait_until(&thread_halt_cond, &dummy_workers_running_lock, end_time))
+      {
+        /* timeout has passed. */
+        msg_error("Main thread timed out while waiting workers threads to exit",
+                  evt_tag_int("workers_running", main_loop_workers_running),
```
bazsi: we should include how long we waited and also what would happen next.

btw: this will probably not be printed, as the internal() source is not in operation at this point.

lib/mainloop.c
```diff
@@ -297,6 +298,27 @@ main_loop_reload_config_initiate(gpointer user_data)
...
 
+static void
+block_till_workers_exit()
+{
+  gint64 end_time;
+  static GMutex dummy_workers_running_lock = G_STATIC_MUTEX_INIT;
+
+  end_time = g_get_monotonic_time() + 15*G_TIME_SPAN_SECOND;
+  g_mutex_lock(&dummy_workers_running_lock);
+  while (main_loop_workers_running)
```
bazsi: what is this dummy? it should be the same lock that protects the counter.

furiel:
I will drop down the dns_cache part. Lets go with the synchronization version.

lib/mainloop.c
furiel: Can I resolve the non printing part somehow? Should I use fprintf here writing to stderr instead of msg_error?

furiel: I'll change this to fprintf.

How long we waited: will be 15s as we get here only when timeout happened. As for what happens next, I'm not sure. I think we can just continue with exiting. Or should we abort? The only root cause I can think of is somehow the accounting of main_loop_workers_running is messed up (nonzero). But for sure all workers will be exited: we only get here when there are no jobs, and after 10 seconds every thread is forced to exit, so waiting 15seconds and continue is safe imo.
For text:
 fprintf(stderr, "Main thread timed out (15s) while waiting workers threads to exit. workers_running: %d. Continuing ...\n", main_loop_workers_running);

lib/mainloop.c
furiel: > iv_work_pool_put

I call this block after main_loop_io_worker_deinit(), which calls iv_work_pool_put.
  main_loop_io_worker_deinit();                                                                                                        
  main_loop_worker_deinit();                                                                                                           
  block_till_workers_exit(); 

> our threaded destination code might be blocking for some time still, but I think that's ok.

For the threaded destinations: currently the complete exit mechanism waits for them: the exit mechanism is executed with main_loop_worker_sync_call, which waits for all threaded destinations to call main_loop_worker_job_complete(). So yes, the threaded destination drivers can block the exit, but the block happens earlier than block_till_workers_exit() called, so the timeout with g_cond_wait_until is not affected.

kira-syslogng:

kira-syslogng:

furiel:
@bazsi could you recheck this PR?

PR #1668, opened 2017-09-12
elasticsearch.http: implemented onMessageQueueEmpty function

pzoleex:
Fixes: #1509
Flush the message automatically when there are no more message in the queue.
Earlier it caused problem when there were less messages than flush_limit() and syslog-ng
does not flush them only at reload/restart.

Possible side effects: if the messages come slower syslog-ng will flush more often
instead of waiting flush_limit().

The classes inherited from ESNativeClient does not use this method
because the bulkprocessor here is responsible for the auto flush.

Signed-off-by: Zoltan Pallagi <pzolee@balabit.com>


kira-syslogng:

kira-syslogng:

PR #1675, opened 2017-09-15
HDFS destination: Added append support

pzoleex:
New option: hdfs_append_enabled("true|false"), default: false
When this option is enabled, syslog-ng will append the file if it already exists.
In that case, syslog-ng does not extend the filename with a unique id as earlier.

!!!
Warning: always ensure that your hdfs server supports append (and it is enabled) before enable this option!
!!!

Note: if you have a very small hdfs cluster (number of nodes < replication factor, for example for testing purpose), you can receive
"Failed to replace a bad datanode on the existing pipeline due to no more good datanodes being available to try"
error message at flush. In that case either decrease the value of dfs.replication or
add dfs.client.block.write.replace-datanode-on-failure.policy key with "NEVER" value
in hdfs-site.xml and give it as resource in the configuration. If you receive this error message
in a production ready hdfs cluster, that means your cluster is not healthy.

Signed-off-by: Zoltan Pallagi <pzolee@balabit.com>


kira-syslogng:

PR #531, opened 2015-06-23
F/literal

juhaszviktor:
Fixes #206

PR #1679, opened 2017-09-16
Binary template function

bazsi:
This branch adds a $(binary) template function, evolving the code originally found in PR #531 

The rationale of this function needs some explanation, as the configuration lexer enables us to encode binary characters in any string, by using C like backslash escapes, e.g. this should work:

template("\x00");

The issue with that is that the lexer->grammar interface does not support embedded NUL characters, thus anything past the NUL would be dropped when the configuration is loaded.

With this function the sample above would become:

template("$(binary 0x00)");

Which is properly parsed, and the embedded zero would only be processed by the destination, which might be supporting embedded NULs. All LogWriter based destinations (file, socket based stuff, program(), pipe, etc) do support that, so this is actually a useful feature.

### DIFF
--- a/libtest/template_lib.c
+++ b/libtest/template_lib.c
@@ -146,22 +146,9 @@ compile_template(const gchar *template, gboolean escaping)
 }
 
 void
-assert_template_format(const gchar *template, const gchar *expected)
-{
-  assert_template_format_with_escaping(template, FALSE, expected);
-}
-
-void
-assert_template_format_msg(const gchar *template, const gchar *expected,
-                           LogMessage *msg)
-{
-  assert_template_format_with_escaping_msg(template, FALSE, expected, msg);
-}
-
-
-void
-assert_template_format_with_escaping_msg(const gchar *template, gboolean escaping,
-                                         const gchar *expected, LogMessage *msg)
+assert_template_format_with_escaping_and_context_msgs(const gchar *template, gboolean escaping,
+                                                      const gchar *expected, gssize expected_len,
+                                                      LogMessage **msgs, gint num_messages)
 {
   LogTemplate *templ = compile_template(template, escaping);
   if (!templ)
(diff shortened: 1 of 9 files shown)


kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

libtest/template_lib.c
```diff
@@ -146,49 +146,63 @@ compile_template(const gchar *template, gboolean escaping)
...
 }
 
 void
-assert_template_format_with_escaping(const gchar *template, gboolean escaping,
-                                     const gchar *expected)
+assert_template_format_with_escaping(const gchar *template, gboolean escaping, const gchar *expected)
 {
   LogMessage *msg = create_sample_message();
 
   assert_template_format_with_escaping_msg(template, escaping, expected, msg);
```
furiel: It would worth set log_template_set_escape(templ, escaping) here instead of compile_template instead of passing around. It would reduce the argument list of many functions. It is a bit hard to follow these functions due to the long name and parameter list.

modules/basicfuncs/str-funcs.c
```diff
@@ -433,3 +433,75 @@ tf_string_padding(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+                      token);
+          goto error;
+        }
+      if (number > 0xFF)
+        {
+          g_set_error(error, LOG_TEMPLATE_ERROR, LOG_TEMPLATE_ERROR_COMPILE,
+                      "$(binary) template function only supports 8 bit values as characters, %" G_GUINT64_FORMAT " is above 255", number);
+          goto error;
+        }
+      g_string_append_c(octets, (guint8) number);
```
furiel: This should be casted to gchar instead of guint8.

modules/basicfuncs/tests/test_basicfuncs.c
```diff
@@ -153,6 +153,14 @@ test_str_funcs(void)
...
   assert_template_format("$(padding foo 10 x)", "xxxxxxxfoo");
   assert_template_format("$(padding foo 10 abc)", "abcabcafoo");
+
+  assert_template_failure("$(binary)", "Incorrect parameters");
+  assert_template_failure("$(binary abc)", "unable to parse abc");
+  assert_template_failure("$(binary 256)", "256 is above 255");
+  assert_template_format("$(binary 1)", "\1");
+  assert_template_format("$(binary 1 0x1)", "\1\1");
+  assert_template_format("$(binary 0xFF 255 0377)", "\xFF\xFF\xFF");
+  assert_template_format_with_len("$(binary 0xFF 0x00 0x30)", "\xFF\0000", 3);
```
furiel: The expected value should be \xFF\0030?

modules/basicfuncs/str-funcs.c
bazsi: done

modules/basicfuncs/tests/test_basicfuncs.c
bazsi: No, \000 is an octal zero, 0 is a character, which is the same as 0x30. But I have
now changed 0x30 to 0x40, which is a '@', should be easier to grasp.

libtest/template_lib.c
bazsi: I agree. The entire libtest should be given some more thought. It is coming from a pre-criterion time, all I did this with this patch was to rebase it from an ancient pull request #531.

I would prefer not to do this refactor at this point, rather to do it as separate pull request in the future. (famous words I know, but I wouldn't want to block this PR because of it. It has been in purgatory for all eternity).

kira-syslogng:

kira-syslogng:

libtest/template_lib.c
furiel: Ok, fine for me.

PR #1684, opened 2017-09-20
xml-parser: bugfixes 

furiel:
- strip-whitespace option is not copied during clone
- when exclude pattern is used and the pipe element is cloned, shutdown leads to double free
- memory leak when exclude patterns is used, and the provided patterns force a string reverse

Now unit tests run on a cloned version of the pipe element instead of the original.


kira-syslogng:

PR #1687, opened 2017-09-20
xml-parser: eliminate side effect g_list_copy_deep

furiel:
Due to coding error, the backported g_list_copy_deep modified the
original list instead of the new list, causing a side effect, and
returns with a new list. This causes crash during free in xml parser.


kira-syslogng:

PR #1688, opened 2017-09-20
Grammar: string_list link in reverse order (eliminate reverse overhead)

Kokan:
**Please don't merge it.**
string_list: if we always insert into the first position into the list no reverse is needed in the end.

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -982,11 +982,11 @@ string_or_number
         ;
 
 string_list
-        : string_list_build                     { $$ = g_list_reverse($1); }
+        : string_list_build                     { $$ = $1; }
         ;
 
 string_list_build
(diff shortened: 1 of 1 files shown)


kira-syslogng:

bazsi:
note that this is a config parsing overhead, not a runtime one, so the primary reason for a patch like this is not performance, but readability.

lib/cfg-grammar.y
```diff
@@ -982,11 +982,11 @@ string_or_number
         ;
 
 string_list
-        : string_list_build                     { $$ = g_list_reverse($1); }
+        : string_list_build                     { $$ = $1; }
```
MrAnno: I guess the `string_list_build` nonterminal rule is not necessary anymore.

PR #1689, opened 2017-09-21
Application adapters + experimental transport for syslog-ng <> syslog-ng communication

bazsi:
This branch is actually a combination of three features:

**1. appmodel: add app-parser()**
    
    The app-parser() is a framework that allows automatic parsing of log
    messages.  Application specific logic is encapsulated into a configuration
    snippet, which is then automatically pulled into the main configuration,
    without having to explicitly change the main configuration file.
    
    For instance:
    
    application appfoo[port514] {
            filter { program("appfoo"); };
            parser { appfoo-parser(); };
    };
    
    application appbar[port514] {
            filter { program("appbar"); };
            parser { appbar-parser(); };
    };
    
    With that defined, app-parser() will automatically incorporate "appfoo" &
    "appbar" specific log processing rules in the configuration, if used this
    way:
    
    log {
            source { udp(); };
            parser { app-parser(specialization(port514)); };  // specialization --> topic
            destination { ... };
    };
    
    app-parser() will generate processing code that would parse both "appfoo"
    and "appbar" properly, and result in name-value pairs specific to those
    applications.
    
    This is the basic idea, but there are some other features, not so easy to
    explain in a commit message, but worth at least mentioning here:
      * applications are iterated in registration order (for now), but this may
        change in the future
      * application rules should ideally be non-overlapping.  If they overlap,
        the first one wins.
      * either the filter or the parser should DROP the message if it does not
        match.  Most parsers do support dropping incorrectly formatted messages,
        but if not, please do that explicitly via a filter
      * if an application matches, a tag named ".app.<appname>" would be
        associated with the message.
      * there can be multiple instances of app-parser() with a different set of
        applications being generated (this is the "port514" above, also called a
        specialization of an application).  It is expected that we would be
        using separate specializations if the same application is logging the
        same messages via multiple channels.
    
    This patch series contains a new source driver (called port514()) that
    implicitly contains app-parser() invocation, so the main syslog-ng.conf file
    becomes very simple indeed.

**2. the ability to transport the complete syslog-ng message model accross syslog-ng instances:**

This is the experimental transport for transferring messages in whole
between syslog-ng instances.
EWMM stands for "enterprise wide message model", which is also kind of an
experimental name, but so far it stuck.
Format:
  - program name should be "@syslog-ng" which is used to recognize this message
  - name-values are encoded as JSON in the MSG field
  - name-value pairs that start with "._" should be hop-by-hop fields only
    to transmit syslog-ng <> syslog-ng information and is not considered
    part of the original message.
  - regexp numeric matches ($0 .. $255) are not transmitted

**3.  The default syslog-ng configuration is changed** 

scl/syslog-ng.conf is changed to use app-parser() automatically, both in system() and in network received logs.

To use the new transport mechanism, you just need to use the syslog-ng() destination driver, like this:

        destination { syslog-ng(server(127.0.0.1) port(2601)); };

On the server side, if the new port514() driver is used, this format would be recognized automatically. If not, you can always parse this format explicitly using the ewmm-parser() parser.

This is a culmination of several new abstractions, so I would probably have to write a blog post about.

@czanik @fekete-robert @faxm0dem

### DIFF
--- a/lib/cfg-block-generator.c
+++ b/lib/cfg-block-generator.c
@@ -25,9 +25,9 @@
 #include "cfg-block-generator.h"
 
 gboolean
-cfg_block_generator_generate(CfgBlockGenerator *self, GlobalConfig *cfg, CfgLexer *lexer, CfgArgs *args)
+cfg_block_generator_generate(CfgBlockGenerator *self, GlobalConfig *cfg, CfgArgs *args, GString *result)
 {
-  return self->generate(self, cfg, lexer, args);
+  return self->generate(self, cfg, args, result);
 }
 
 void
--- a/lib/cfg-block-generator.h
+++ b/lib/cfg-block-generator.h
@@ -43,11 +43,12 @@ struct _CfgBlockGenerator
 {
   gint context;
   gchar *name;
-  gboolean (*generate)(CfgBlockGenerator *self, GlobalConfig *cfg, CfgLexer *lexer, CfgArgs *args);
+  gboolean suppress_backticks;
+  gboolean (*generate)(CfgBlockGenerator *self, GlobalConfig *cfg, CfgArgs *args, GString *result);
   void (*free_fn)(CfgBlockGenerator *self);
 };
 
-gboolean cfg_block_generator_generate(CfgBlockGenerator *self, GlobalConfig *cfg, CfgLexer *lexer, CfgArgs *args);
+gboolean cfg_block_generator_generate(CfgBlockGenerator *self, GlobalConfig *cfg, CfgArgs *args, GString *result);
 void cfg_block_generator_init_instance(CfgBlockGenerator *self, gint context, const gchar *name);
 void cfg_block_generator_free_instance(CfgBlockGenerator *self);
 void cfg_block_generator_free(CfgBlockGenerator *self);
--- a/lib/cfg-block.c
+++ b/lib/cfg-block.c
@@ -24,6 +24,8 @@
 #include "cfg-block.h"
 #include "cfg-lexer.h"
 #include "cfg-lexer-subst.h"
+#include "cfg.h"
+#include "str-utils.h"
 
 /*
  * This class encapsulates a configuration block that the user defined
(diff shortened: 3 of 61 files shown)


bazsi:
note, that this depends on pr #1657

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

pzoleex:

(Bazsi: I increased the number of lines for syslog-ng output to display)

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:

bazsi:
@pzoleex The only remaining issue is the change in the memory allocation profile that breaks the testcase which checks for memory allocations.

This patch 20492a902176ca7087cd4fe0f8a87f024a0d2f62 is what changes the defaults. Can you please change the testcase to accept the new value?

Thanks

pzoleex:

kira-syslogng:

czanik:
Tested application adapters and enterprise wide message model using sudo logs and also tested the new store-raw-message flag. All seem to work fine.

It is good to have application adapters parsing logs coming from the system() source by default. But I would recommend to have the possibility of disabling it, as parsing has performance overhead which might be problematic in embedded and/or high traffic environments.

faxm0dem:
Disclaimer: this is awesome, and funcionally covers all + more I wanted from a syslog-ng destination, so thanks a bunch!

*But*, I think the implementation would benefit from a more advanced protocol than JSON over IP.
Some kind of binary format would make things much more efficient, *e.g.* using protocol-buffers or avro.

Cheers!

lbudai:
@bazsi: there was a trivial conflict (modules/confgen/confgen-plugin.c) caused by #1781 .
I resolved the conflict and force pushed your branch (before the force push I saved your original branch: https://github.com/balabit/syslog-ng/tree/original-application-adapters )

kira-syslogng:

kira-syslogng:

Issue #1667, opened 2017-09-12
getent: does not compile on solaris


furiel:
Versions:
```
# uname -a
SunOS solaris12 5.11 11.3 i86pc i386 i86pc
# cat /etc/release 
                             Oracle Solaris 11.3 X86
  Copyright (c) 1983, 2015, Oracle and/or its affiliates.  All rights reserved.
                            Assembled 06 October 2015
```

```
In file included from modules/getent/tfgetent.c:113:0:
modules/getent/getent-protocols.c: In function ‘tf_getent_protocols’:
modules/getent/getent-protocols.c:32:5: error: too many arguments to function ‘getprotobynumber_r’
     getprotobynumber_r((int) d, &proto, buf, sizeof(buf), &res);
     ^
In file included from modules/getent/tfgetent.c:33:0:
/usr/include/netdb.h:249:18: note: declared here
 struct protoent *getprotobynumber_r
                  ^
In file included from modules/getent/tfgetent.c:113:0:
modules/getent/getent-protocols.c:34:5: error: too many arguments to function ‘getprotobyname_r’
     getprotobyname_r(key, &proto, buf, sizeof(buf), &res);
     ^
In file included from modules/getent/tfgetent.c:33:0:
/usr/include/netdb.h:247:18: note: declared here
 struct protoent *getprotobyname_r
                  ^
In file included from modules/getent/tfgetent.c:114:0:
modules/getent/getent-services.c: In function ‘tf_getent_services’:
modules/getent/getent-services.c:32:5: error: too many arguments to function ‘getservbyport_r’
     getservbyport_r((int)ntohs(d), NULL, &serv, buf, sizeof(buf), &res);
     ^
In file included from modules/getent/tfgetent.c:33:0:
/usr/include/netdb.h:238:17: note: declared here
 struct servent *getservbyport_r
                 ^
In file included from modules/getent/tfgetent.c:114:0:
modules/getent/getent-services.c:34:5: error: too many arguments to function ‘getservbyname_r’
     getservbyname_r(key, NULL, &serv, buf, sizeof(buf), &res);
     ^
In file included from modules/getent/tfgetent.c:33:0:
/usr/include/netdb.h:236:17: note: declared here
 struct servent *getservbyname_r
                 ^
In file included from modules/getent/tfgetent.c:115:0:
modules/getent/getent-group.c: In function ‘tf_getent_group’:
modules/getent/getent-group.c:47:9: error: too many arguments to function ‘getgrgid_r’
[...]

PR #1691, opened 2017-09-21
solaris: compile errors in getent and sun_streams

furiel:
There is a little API difference between the linux and solaris version of getent functions.

In Linux, they return errno, and data in an output parameter.
In solaris, they return the data, and set errno.

This patch provides wrapper around the solaris versions, so that we can use them as in linux: errno is returned, and data is written to the output parameter.

There was a minor compile error in sun_streams, it is fixed as well.

Fixes: https://github.com/balabit/syslog-ng/issues/1667 and https://github.com/balabit/syslog-ng/issues/1692

### DIFF
--- a/lib/compat/getent.h
+++ b/lib/compat/getent.h
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2017 Balabit
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#ifndef GETENT_COMPAT_H_INCLUDED
+#define GETENT_COMPAT_H_INCLUDED
+
+#include <sys/types.h>
+#include <grp.h>
+#include <pwd.h>
+#include <netdb.h>
+
+#if defined(sun) || defined(__sun)
+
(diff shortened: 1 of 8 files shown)


kira-syslogng:

modules/getent/getent-group.c
```diff
@@ -20,6 +20,8 @@
  * COPYING for details.
  */
 
+#include "compat.h"
```
Kokan: These files are included into the `tfgetent.c`, which is not the best idea imho.
But if you don't plan to resolve this incleption I think this `compat.h` should be in the `tfgentent.c`. Because it would mix the two concept.
Of course the preferred fix would be to make these files properly independent compilation chunk.

kira-syslogng:

presidento:
Please, move the _compat_ functionality to `lib` folder.
And perhaps we can use a feature macro (like `GNU_SOURCE` or `BSD_SOURCE`.

kira-syslogng:

kira-syslogng:

furiel:
For now I just moved to lib.

About the feature macros:
There are no feature macros for these:
getprotobynumber_r
getprotobyname_r
getservbyport_r
getservbyname_r

But for these there are compatible declarations with feature macros:
getgrnam_r
getgrgid_r
getpwnam_r
getpwuid_r

Checked on Oracle Solaris 11.3

kira-syslogng:

kira-syslogng:

furiel:
For the feature macros, I added this to tfgetent.c
```
+#if defined(sun) || defined(__sun)
+#define _POSIX_PTHREAD_SEMANTICS
+#endif
```

However, as feature macros are involved now, I did not feel comfortable to test the group part only with compile so I added an extra commit: please check: `getent: crash with empty group` that clears: https://github.com/balabit/syslog-ng/issues/1692. This way I could manually test the feature.

lib/compat/getent.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef GETENT_COMPAT_H_INCLUDED
+#define GETENT_COMPAT_H_INCLUDED
+
+#include <sys/types.h>
+#include <grp.h>
+#include <pwd.h>
+#include <netdb.h>
+
+int _getprotobynumber_r(int proto,
```
presidento: For me, the underscore prefix is reserved for protected/private members. I suggest using either a different prefix (`bb_` can be good for me), or shallow the original function call in Solaris with a `#define`, like the one in [`pio.h`](https://github.com/balabit/syslog-ng/blob/master/lib/compat/pio.h#L41) file.

kira-syslogng:

kira-syslogng:

kira-syslogng:

lib/compat/getent.h
furiel: @presidento thanks for comment, fixed.

PR #1694, opened 2017-09-22
Fix python template

lbudai:


kira-syslogng:

kira-syslogng:

lib/logmsg/nvtable.c
```diff
@@ -187,6 +187,8 @@ nv_table_resolve_indirect(NVTable *self, NVEntry *entry, gssize *length)
    * non-zero terminated strings properly handled, thus the caller has
    * to supply a non-NULL value_len */
 
+  g_assert(length != NULL);
```
furiel: I would allow length to be null, in that case we just simply not write to the output parameter, similarly as we do in line 181.
For example in the python code, we indeed do not need the length, we could avoid the dummy parameter.

lbudai: @furiel : based on the comment it is better to assert, IMHO

furiel: Hmmm that's a problem because it would mean PyBytes_FromString() could be possibly called on a non-null terminated string. I'm afraid we should strndup with value_len on the caller site then.

lbudai:
@bazsi: well... I can write  a unit test...

modules/python/tests/test_python_logmsg.c
```diff
@@ -0,0 +1,135 @@
...
+    PyObject *msg_object = py_log_message_new(msg);
+
+    PyDict_SetItemString(_python_main_dict, "test_msg", msg_object);
+
+    gssize len = strlen(test_value);
+    NVHandle indirect_key_handle = log_msg_get_value_handle("indirect");
+    log_msg_set_value_indirect(msg, indirect_key_handle, test_key_handle, 0, 0, len);
+
+    const gchar *get_value_indirect = "indirect = test_msg['indirect']";
+    PyRun_String(get_value_indirect, Py_file_input, _python_main_dict, _python_main_dict);
```
lbudai: Without the fix the test crashes here...

kira-syslogng:

modules/python/tests/test_python_logmsg.c
```diff
@@ -0,0 +1,135 @@
...
+    PyRun_String(set_value, Py_file_input, _python_main_dict, _python_main_dict);
+
+    gchar *res = _dict_clone_value(_python_main_dict, "result");
+    cr_assert_str_eq(res, "test_value");
+    g_free(res);
+
+    NVHandle test_field_handle = log_msg_get_value_handle("test_field");
+    gssize len = 0;
+    const gchar *test_value=log_msg_get_value(msg, test_field_handle, &len);
+    cr_assert_str_eq(test_value, "test_value");
```
bazsi: we have log_msg_get_value_by_name() which would make this simpler.

modules/python/tests/test_python_logmsg.c
```diff
@@ -0,0 +1,135 @@
...
+
+Test(python_log_message, test_python_logmessage_set_value_indirect)
+{
+  const gchar *raw_msg = "test_msg";
+  const gchar *test_value = "test_value";
+  const gchar *test_key = "test_key";
+  LogMessage *msg = log_msg_new(raw_msg, strlen(raw_msg), NULL, &parse_options);
+  NVHandle test_key_handle = log_msg_get_value_handle(test_key);
+  log_msg_set_value(msg, test_key_handle, test_value, -1);
+
```
bazsi: also set_value_by_name()

lib/logmsg/nvtable.c
bazsi: log_msg_get_value() returning a length field is important to support both:

1) indirect values
2) embedded NULs

the 2nd is not yet reality and I am not sure it would ever be, but the first one is.

Please note however that builtin values (e.g. LM_V_MESSAGE and friends) are always nul terminated and never indirect. (there's an assert to ensure that in log_msg_set_value_indirect)

bazsi: With this command:

```
$ git grep log_msg_get_value\( | grep -v '&' | grep -v test | grep -v LM_V
lib/logmsg/logmsg.c:__log_msg_get_value(const LogMessage *self, NVHandle handle, gssize *value_len)
lib/logmsg/logmsg.h:log_msg_get_value(const LogMessage *self, NVHandle handle, gssize *value_len)
lib/logmsg/logmsg.h:  return log_msg_get_value(self, handle, value_len);
modules/python/python-logmsg.c:  const gchar *value = log_msg_get_value(py_msg->msg, handle, NULL);
```

I've doublechecked that the only call-site that was doing this was indeed the Python implementation (well, except for tests).

bazsi: one more comment: there's a macro to ensure a string is NUL terminated, just for this purpose, it is called APPEND_ZERO in str-utils.h

What it does that it checks if there's a NUL at the end of the string, and if there's none, it copies the string to an alloca() provided buffer. It is much faster than doing this with all the values, if we except that most of them will be NUL terminated.

modules/python/tests/test_python_logmsg.c
lbudai: done

modules/python/tests/test_python_logmsg.c
lbudai: In this case I wouldn't modify the code, as I need the handle for the `log_msg_set_value_indirect`.

lib/logmsg/nvtable.c
lbudai: done

kira-syslogng:

kira-syslogng:

lbudai:
ehh...  style-checker should be fine now...

kira-syslogng:

PR #1701, opened 2017-09-27
doc: do not access to the internet during generating manpages

presidento:


kira-syslogng:

presidento:
Hm, this is absolutely wrong way. See in `configure --help`:
```
  --with-docbook-dir=DIR  compiling manpages using docbook in the specified
                          path, if not set online version will be used from
                          http://docbook.sourceforge.net
```
If we want to compile without internet access, we should use this option.

Issue #1210, opened 2016-09-15
Feature: native network load balancing


falzm:
Hi

I'd like to know if it has been considered implementing a native network-related destination load balancing mecanism in syslog-ng? I'm running *loghosts* that receive logs from hundreds of servers: those loghosts are accessed through a DNS-based endpoint using Consul, however once a server is connected to a loghost it streams its logs only to this server until the connection dies or one of the sender/receiver syslog-ng instance is restarted. The problem is that the distribution of the logs is very unbalanced, and some loghosts are much more loaded than others because of the various logging throughputs across the source servers...

I've search for a bit and the closest I've found is a [mailing list thread](https://lists.balabit.hu/pipermail/syslog-ng/2011-September/017323.html) back in 2011 where the ad-hoc solution was to perform sort-of manual distribution by hashing/regexp'ing the `$SEQNUM` variable of log messages and choose one of the destinations based on the output of the hash/regexp – which is very hacky and I guess very inefficient.

It would be very useful to be able to specify several hosts in the network-based destinations – e.g. `syslog(ip("host1", "host2", "hostN") transport(tcp) ...)` – and have syslog-ng able to split the traffic at protocol level (i.e. on a message basis) among the configured hosts, say in a round-robin fashion, thus ensuring a much more fair logging traffic balancing.

bazsi:
Agreed, this would be a great addition. It would probably have to work with
a $HOST based hashing though, as normally syslog messages are expected to
be logged in sequence, at least from the same host.

The "protocol implementation" is encapsulated into a LogWriter instance,
which the AFSocketDestDriver could use multiple of. A similar
implementation is in the affile module, where we do template based queueing
to multiple files.



-- 
Bazsi

falzm:
@bazsi well, if it could be up to the user to choose the key on which the hashing would be performed that would be great, since hashing on `$HOST` would have the same unbalanced effect in my case... In my case it's OK for logs to be spread on multiple destinations.

bazsi:
Hi,

ok, and you don't care about sequenced delivery within the same host? what
about "$HOST$PROGRAM$PID" would that resolve this for you? Although
PROGRAM/PID are not always filled.

-- 
Bazsi

falzm:
In my use case I'm dealing with JSON-serialized messages, where a message is self-contained (i.e. no chance of being split on several consecutive messages). So as long as the user can specify the log element on which perform the hashing – for example `$SEQNUM` –  I'm good :)

[...]

PR #1706, opened 2017-10-04
Simple rr-network load balancer

illesm:
It implements a simple round-robin load balancing between multiple destination.
    Balancing is based on MSEC hashing where destinations share the same configuration
    except the destination address.
    The main purpose to make load balance configuration simple.
    
    It can be used as a simple network destination and all network parameters are
    recognized, the destination servers must be configured as a space separated list
    of host using the "targets" option:
    
    destination mydest {
        load-balancer(
                targets(server1 server2 server2) # mandatory option
    
                # additional network options can be set as well:
                transport(udp)
                port(3456)
        )
    };
    
    NOTE: as failover-server is not supported in this version it is left out. Otherwise
    nodes are used for fail-over servers to maintain availability.
    
    The idea and original configuration comes from Ferenc Sipos.
    Big kudos for the inspiration and for his persistence on pushing me to add
    this function on our flight from New York to Memphis, TN! Thx!

Also extended confgen module:
    confgen: enable passing arguments to confgen external program as env variables
    
    Options can be set when invoking confgen external program. These options are
    passed as environment variables. Note that env variables are overwritten and
    cleared after the external execution.
    
    This comes handy as this way configuration can be generated not only dynamically,
    but also driven from the configuration.
    
    @module confgen context(destination) name(mygen) exec("gen.sh")
    
    And later that can be used as:
    mygen(
        option1(param1)
        option2("my other option")
    )

This functionality probably addresses #1210 mostly.

### DIFF
--- a/modules/confgen/confgen-plugin.c
+++ b/modules/confgen/confgen-plugin.c
@@ -30,6 +30,24 @@
 
 #include <string.h>
 #include <errno.h>
+#include <stdlib.h>
+
+
+static void
+confgen_set_args_as_env(gpointer k, gpointer v, gpointer user_data)
+{
+  gchar buf[1024];
+  g_snprintf(buf, sizeof(buf), "confgen_%s", (gchar *)k);
+  setenv(buf, (gchar *)v, 1);
+}
+
+static void
+confgen_unset_args_from_env(gpointer k, gpointer v, gpointer user_data)
+{
+  gchar buf[1024];
+  g_snprintf(buf, sizeof(buf), "confgen_%s", (gchar *)k);
+  unsetenv(buf);
+}
 
 gboolean
 confgen_generate(CfgLexer *lexer, gint type, const gchar *name, CfgArgs *args, gpointer user_data)
@@ -43,15 +61,11 @@ confgen_generate(CfgLexer *lexer, gint type, const gchar *name, CfgArgs *args, g
   gboolean result;
 
   g_snprintf(buf, sizeof(buf), "%s confgen %s", cfg_lexer_lookup_context_name_by_type(type), name);
-  if (!cfg_args_validate(args, NULL, buf))
-    {
-      msg_error("confgen: confgen invocations do not process arguments, but your argument list is not empty",
-                evt_tag_str("context", cfg_lexer_lookup_context_name_by_type(type)),
-                evt_tag_str("block", name));
-      return FALSE;
-    }
 
+  cfg_args_foreach(args, confgen_set_args_as_env, NULL);
   out = popen((gchar *) user_data, "r");
+  cfg_args_foreach(args, confgen_unset_args_from_env, NULL);
+
   if (!out)
     {
       msg_error("confgen: Error executing generator program",
--- a/scl/loadbalancer/gen-loadbalancer.sh
+++ b/scl/loadbalancer/gen-loadbalancer.sh
@@ -1,0 +1,56 @@
+#!/bin/sh
+#############################################################################
+# Copyright (c) 2017 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+TARGETS=$confgen_targets
+TARGET_PARAMS=$confgen_parameters
+
+target_cnt=$(echo $TARGETS | wc -w)
+target_id=0
+echo 'channel {'
+for target in $TARGETS
+do
+#    other_targets=$(echo $TARGETS | sed -e s/\\b${target}\\b//)
+    echo '  channel {'
+    echo '    filter {'
+    echo -n '    "'
+    echo -n $target_id
+    echo -n '" == "$(% ${R_MSEC} '
+    echo -n $target_cnt
+    echo ')"'
+    echo '    };'
+    echo '    destination {'
+    echo -n '      network("'
+    echo -n $target
+    echo -n '" '
+    echo -n $TARGET_PARAMS
+#echo -n 'failover-servers("'
+#echo -n $other_targets
+#echo -n '")'
+    echo '      );'
+    echo '    };'
+    echo '    flags(final);'
+    echo '  };'
+    target_id=$((target_id+1))
+done
+echo '};'
+
--- a/scl/loadbalancer/plugin.conf
+++ b/scl/loadbalancer/plugin.conf
@@ -1,0 +1,32 @@
+#############################################################################
+# Copyright (c) 2017 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+@module confgen context(destination) name(genlb) exec("`scl-root`/loadbalancer/gen-loadbalancer.sh")
+
+
+block destination network-load-balancer(targets()) {
+  genlb(
+      targets(`targets`)
+      parameters(`__VARARGS__`)
+    )
+};
+
--- a/scl/Makefile.am
+++ b/scl/Makefile.am
@@ -17,7 +17,9 @@ SCL_SUBDIRS	= \
 	logmatic	\
 	snmptrap        \
 	osquery	        \
-        windowseventlog
+        windowseventlog  \
+        loadbalancer
+
 SCL_CONFIGS	= scl.conf syslog-ng.conf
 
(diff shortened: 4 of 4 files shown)


kira-syslogng:

presidento:
@kira-syslogng ok to test.

presidento:
The idea is clever.
It is not a problem that we can loss the order of the incoming logs? As I remember, a previous idea was to do the load balancing based on the source host to keep the sequence of logs coming from a device. But I am not sure.

kira-syslogng:

illesm:
In many LB scenarios you will end up loosing sequence, but of course you use some slng assigned sequence number to reproduce it. Anyhow this is a simple solution that is useful for scenarios where strict sequencing is not important, but you have lot of logs and need to just distribute it.

You can of course use "sticky host" load-balancing, but that might be a limitation where one host produces too many logs. Also in case of larger deployments with multiple nodes producing logs the strict sequence is probably not that important.

kira-syslogng:

bazsi:
Random set of thoughts:

* I like confgen passing the environment variables to the process, however setenv/unsetenv is not very nice, but probably not very problematic either. The usual way is to create a custom environ aray, and use that in execve(), however there's no "e" variant from popen.
* The persist-id should probably be set for each network() destination to be a unique value, as otherwise the queues may not be separated.
* the ordering issue can be fixed by using $HOST instead of MSEC as the routing key. It may not be that balanced as MSEC, but then it is a tradeoff by the customer. What if the template  to produce the routing key can be specified? Then one could use $(sha1 $HOST) and route based on the first character.
* with the latest register-generators-as-plugins refactor this could become a C implementation, thus no need for an external program which causes headaches with SELinux.

scl/loadbalancer/gen-loadbalancer.sh
```diff
@@ -0,0 +1,56 @@
...
+    echo -n '" == "$(% ${R_MSEC} '
+    echo -n $target_cnt
+    echo ')"'
+    echo '    };'
+    echo '    destination {'
+    echo -n '      network("'
+    echo -n $target
+    echo -n '" '
+    echo -n $TARGET_PARAMS
+#echo -n 'failover-servers("'
```
Kokan: Could you remove this commented part of the script ?

scl/loadbalancer/gen-loadbalancer.sh
```diff
@@ -0,0 +1,56 @@
...
+#
+#############################################################################
+
+TARGETS=$confgen_targets
+TARGET_PARAMS=$confgen_parameters
+
+target_cnt=$(echo $TARGETS | wc -w)
+target_id=0
+echo 'channel {'
+for target in $TARGETS
```
Kokan: The `MSEC` max value is a limit of the number of the `target`s. In this case the script do not really need to create the extra `channel`s.

scl/loadbalancer/plugin.conf
```diff
@@ -0,0 +1,32 @@
...
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+@module confgen context(destination) name(genlb) exec("`scl-root`/loadbalancer/gen-loadbalancer.sh")
+
+
+block destination load-balancer(targets()) {
```
Kokan: I would include somehow in the name, that it is currently for `network` destination only, and not a full-fledged general load balancer.

scl/loadbalancer/gen-loadbalancer.sh
```diff
@@ -1,4 +1,25 @@
 #!/bin/sh
+#############################################################################
+# Copyright (c) 2017 Balabit
```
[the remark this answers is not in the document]
illesm: I can, but why?
it does two different things on two different modules.

scl/loadbalancer/gen-loadbalancer.sh
illesm: I do not understand this limit, but if you ommit the channel definition the final flag would stop the msg processing not just for this but for other destination.
what do i miss?

scl/loadbalancer/gen-loadbalancer.sh
```diff
@@ -0,0 +1,56 @@
...
+target_id=0
+echo 'channel {'
+for target in $TARGETS
+do
+#    other_targets=$(echo $TARGETS | sed -e s/\\b${target}\\b//)
+    echo '  channel {'
+    echo '    filter {'
+    echo -n '    "'
+    echo -n $target_id
+    echo -n '" == "$(% ${R_MSEC} '
```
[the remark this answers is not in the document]
illesm: yes, it could work.
@bazsi also suggested using HOST, but that is bit tricky to create a filter as there is not module for string, of course we can use comparison always. Though it would not balance well with many logs coming from one host. Probably that could be made configurable.

Either MSEC or SEQNUM could work, however we wanted something simple.

I would leave it as it is now, and if someone wants to enhance it they can send PR.

scl/loadbalancer/gen-loadbalancer.sh
illesm: i intentionaly left it here as there is failover support in PE6, but not here, so anyone wanting to port it would find it useful.
Of couse it can be removed.

modules/confgen/confgen-plugin.c
```diff
@@ -30,6 +30,24 @@
 
 #include <string.h>
 #include <errno.h>
+#include <stdlib.h>
+
+
+static void
+confgen_set_args_as_env(gpointer k, gpointer v, gpointer user_data)
+{
+  gchar buf[1024];
```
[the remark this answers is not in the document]
illesm: we truncate the variable name, but it is fairly unlikely to have such a long option name

modules/confgen/confgen-plugin.c
```diff
@@ -30,6 +30,24 @@
...
+  g_snprintf(buf, sizeof(buf), "confgen_%s", (gchar *)k);
+  setenv(buf, (gchar *)v, 1);
+}
+
+static void
+confgen_unset_args_from_env(gpointer k, gpointer v, gpointer user_data)
+{
+  gchar buf[1024];
```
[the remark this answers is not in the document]
illesm: we truncate the variable name, but it is fairly unlikely to have such a long option name

modules/confgen/confgen-plugin.c
```diff
@@ -30,6 +30,24 @@
...
 #include <errno.h>
+#include <stdlib.h>
+
+
+static void
+confgen_set_args_as_env(gpointer k, gpointer v, gpointer user_data)
+{
+  gchar buf[1024];
+  g_snprintf(buf, sizeof(buf), "confgen_%s", (gchar *)k);
+  setenv(buf, (gchar *)v, 1);
```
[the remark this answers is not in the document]
illesm: yes we want to overwrite it

we can handle errors, but it is basically an out of memory case only, which is bad anyhow

       EINVAL name is NULL, points to a string of length 0, or contains an '='
              character.

       ENOMEM Insufficient memory to add a new variable to the environment.

modules/confgen/confgen-plugin.c
```diff
@@ -43,15 +61,11 @@ confgen_generate(CfgLexer *lexer, gint type, const gchar *name, CfgArgs *args, g
...
   g_snprintf(buf, sizeof(buf), "%s confgen %s", cfg_lexer_lookup_context_name_by_type(type), name);
-  if (!cfg_args_validate(args, NULL, buf))
-    {
-      msg_error("confgen: confgen invocations do not process arguments, but your argument list is not empty",
-                evt_tag_str("context", cfg_lexer_lookup_context_name_by_type(type)),
-                evt_tag_str("block", name));
-      return FALSE;
-    }
 
+  cfg_args_foreach(args, confgen_set_args_as_env, NULL);
```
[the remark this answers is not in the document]
illesm: good point

scl/loadbalancer/plugin.conf
illesm: network-load-balancer is ok or maybe network-rr?

scl/loadbalancer/gen-loadbalancer.sh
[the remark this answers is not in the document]
Kokan: Sorry I meant not those two commits into one :)

You have currently 4 commits, from which 2 of them just for style check and copyright.
It would be nice to squash this modification into the original two commits (those are well separated).

The **confgen: fixed coding style error** and **confgen: enable passing arguments to confgen external program as env** into one, also the **loadbalancer: added missing copyright headers** and  **loadbalancer: added network load balancer destination** into one.

I hope it make more sense like this :)

scl/loadbalancer/plugin.conf
Kokan: I would prefer `network-load-balancer`.

scl/loadbalancer/gen-loadbalancer.sh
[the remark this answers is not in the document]
illesm: fair, i will do it :)

scl/loadbalancer/gen-loadbalancer.sh
Kokan: If there are `MSEC + n` target, those `n` targets won't be chosen.  Of course it won't cause any issue, but it enlarges the configuration unnecessarily.
It would be a minor *tweak* not even creating them.

scl/loadbalancer/gen-loadbalancer.sh
Kokan: I would rather not include it here.

modules/confgen/confgen-plugin.c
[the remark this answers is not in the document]
Kokan: It is fair enough.

modules/confgen/confgen-plugin.c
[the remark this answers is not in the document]
Kokan: It is fair enough.

scl/loadbalancer/gen-loadbalancer.sh
illesm: check the code, it wont happen, unless they configure more then 2^128 servers destination :)
we do the modulo of the # of targets on MSEC and MSEC is a large number, even in 1 sec there are 1000 different value, so it should be ok up to a dozen target for sure.

kira-syslogng:

scl/loadbalancer/gen-loadbalancer.sh
MrAnno: I think @Kokan just wanted to clarify that `MSEC` is not a large integer, it's a number between 0 and 999.
That is the reason why you may not want to generate more than 1000 channel blocks.

MrAnno: I don't have any problems with the current implementation, we just wanted to let you know about this. :)

illesm: I do not think that this would actually scale to more then a dozen (and also does not seem to be realistic) so we should be fine wit this :)

illesm:
I fixed what I considered important, so feel free to merge it or enhance it further if you want

scl/loadbalancer/gen-loadbalancer.sh
Kokan: Have you tried this out in real life? Or any kind of emulation with high traffic ? It would be nice to see how balanced it gets.

illesm: Feri used a less efficient version (using regexp) in production, ask him for details.

PR #1683, opened 2017-09-19
Astyle for headers, minor travis improvement

Kokan:
* Astyle check apply for headers also
* Astyle excludeded afamqp, while it should have excluded only the rabbitmq-c submodule
* Travis now shows what should be changed in the style-check job (calling format and git diff)

PR #1708, opened 2017-10-05
Style check for amqp

Kokan:
There are three improvement in the initial PR #1683 is blocked by #1657 from which two of them could be merged without issue.

This PR aims to contains only those changes:
1) Style check excludes the *amqp* module, but it should only exclude the *rabbitmq-c* submodule under it.
2) In Travis CI if the style-check fails, in the console output it is going to print, why.


kira-syslogng:

PR #1710, opened 2017-10-06
riemann: microseconds support

furiel:
This patch introduces support for event time as microseconds in Riemann destination. 

The event_time() option in Riemann destination configuration
shall take an extra optional parameter specifying the time format
is in seconds (unixtime) or microseconds.
For example:
```
  event_time("$(* $UNIXTIME 1000000)" microseconds)
  event_time("12345678" microseconds)
  event_time("12345678" seconds)
  event_time("12345678")
```

In case the parameter is omitted, syslog-ng defaults to the seconds version. In case event_time is omitted, syslog-ng will default to the seconds version with $UNIXTIME, the same as the original behaviour.

Dependencies:
 - riemann-c-client: 1.10.0 or newer
 - Riemann: 2.13 or newer

With older riemann-c-client, the microseconds option will not be
available.

Older Riemann cannot understand microseconds, though she will not
signal an error either. In that case the :time of the event will be
set to the timestamp when the message arrived to Riemann.

In case the distribution does not contain recent enough riemann-c-client, one needs to install a newer version from https://github.com/algernon/riemann-c-client to use microseconds (for example xenial ships 1.8.1 which is not recent enough). riemann-c-client provides a pkg-config file (.pc file), so one can optionally install it to a custom location and just append the directory of the .pc file to the environment variable `export PKG_CONFIG_PATH=...` before calling configure.

If all goes well in the end of the configure command one will see:
```
[...]
 Riemann destination (module): yes, microseconds: yes
[...]
```
reporting that microseconds support is successfully detected or not.

Fixes: https://github.com/balabit/syslog-ng/issues/1598

### DIFF
--- a/modules/riemann/riemann-grammar.ym
+++ b/modules/riemann/riemann-grammar.ym
@@ -33,6 +33,7 @@
 #include "cfg-grammar.h"
 #include "cfg-parser.h"
 #include "plugin.h"
+#include "riemann/event.h"
 }
 
 %name-prefix "riemann_"
@@ -48,6 +49,8 @@
 %token KW_STATE
 %token KW_SERVICE
 %token KW_EVENT_TIME
+%token KW_SECONDS
+%token KW_MICROSECONDS
 %token KW_DESCRIPTION
 %token KW_METRIC
 %token KW_TTL
@@ -90,7 +93,7 @@ riemann_option
           {
             riemann_dd_set_field_service(last_driver, $3);
           }
-        | KW_EVENT_TIME '(' template_content ')'
+        | KW_EVENT_TIME '(' template_content event_time_unit ')'
           {
             riemann_dd_set_field_event_time(last_driver, $3);
           }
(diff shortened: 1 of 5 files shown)


kira-syslogng:

modules/riemann/riemann-grammar.ym
```diff
@@ -90,7 +93,7 @@ riemann_option
           {
             riemann_dd_set_field_service(last_driver, $3);
           }
-        | KW_EVENT_TIME '(' template_content ')'
+        | KW_EVENT_TIME '(' template_content event_time_unit ')'
```
Kokan: Do we have a reason not to save the time with ms accuracy if possible ?

Kokan: We spoke IRL about this.
Before this PR, the value provided in the `event_time` is a number which should be in seconds granularity. Because in this field a number is provided without any dimension, and it shall be determine the precision.

furiel:
@faxm0dem could you try the patch to see if it works for you?

faxm0dem:
Unfortunately I will not be able to test this before October 20th.

furiel:
@faxm0dem Ok no problem. We will wait for your review.

furiel:
@faxm0dem did you have time to check this PR?

faxm0dem:
Not yet, I'm resuming work tomorrow

faxm0dem:
testing now, will give you an update by this afternoon

faxm0dem:
The first tests are very positive: compilation works as expected, and riemann sees the event and processes them all right using microsecond resolution.

On a sidenote, @mcorbin kindly volunteered to patch wireshark to support the new riemann protocol buffer layout, so I can test the packets themselves just to be sure (probably unneccesary, as riemann-c-client produces the packets behind the scene).

Anyway, while functional, currently the best way I found to configure the riemann destination correctly is the following:

```
destination d_riemann {
  riemann(
    frac-digits(0)
    event_time("$(+ $(* $UNIXTIME 1000000) $USEC)" microseconds)
  );
};
```

The reason for this is that that the following does not work:

```
destination d_riemann {
  riemann(
    frac-digits(6)
    event_time("$(* $UNIXTIME 1000000)" microseconds)
  );
};
```

as the `*` template function doesn't like fractional numbers:

```
Parsing failed, template function's first argument is not a number; function='*', arg1='1509630911.590283'
```

I'm fine by that, but maybe it would make sense to extend either the `*` template function to allow for floats, or to create a new macro.

faxm0dem:
A patched wireshark (thanks @mcorbin) shows that the riemann payload looks perfectly fine!

Kokan:
@furiel thank you for the implementation
@faxm0dem thank you for the testing it
We started to talk about the float issue, how to address it.

PR #1715, opened 2017-10-13
Provide tls block for tls options in amqp(), http(), riemann() dest. drivers

mitzkia:
Three destination drivers: amqp(), http(), riemann() allows to use tls options inside the driver configuration. 
This patch-set provides to use these tls options also in a general tls() block as it is used in network type drivers.
The patch-set did not brake the original usage for actual drivers.
Example new type of configuration for tls options in riemann destination:
```
destination d_riemann {
	riemann(
		server("127.0.0.1") 
		port(5672) 
		tls(
			ca_file("ca") 
			key_file("key")
		)
	);
};
```

### DIFF
--- a/modules/riemann/riemann-grammar.ym
+++ b/modules/riemann/riemann-grammar.ym
@@ -115,13 +115,7 @@ riemann_option
           {
             riemann_dd_set_field_tags(last_driver, $3);
           }
-        | KW_TYPE '(' LL_IDENTIFIER riemann_tls_options ')'
-          {
-            CHECK_ERROR(riemann_dd_set_connection_type(last_driver, $3), @3,
-                        "Unknown Riemann connection type: %s", $3);
-            free($3);
-          }
-        | KW_TYPE '(' LL_STRING riemann_tls_options ')'
+        | KW_TYPE '(' string riemann_tls_options ')'
           {
             CHECK_ERROR(riemann_dd_set_connection_type(last_driver, $3), @3,
                         "Unknown Riemann connection type: %s", $3);
--- a/modules/riemann/riemann-grammar.ym
+++ b/modules/riemann/riemann-grammar.ym
@@ -56,6 +56,7 @@
 %token KW_CA_FILE
 %token KW_CERT_FILE
 %token KW_KEY_FILE
+%token KW_TLS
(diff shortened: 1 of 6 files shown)


kira-syslogng:

kira-syslogng:

MrAnno:
You can use `KWS_OBSOLETE`, if you want to deprecate a keyword:

https://github.com/balabit/syslog-ng/blob/7faef8ef82e31260aab239e2aaf2c8bd20bb438b/modules/afsocket/afsocket-parser.c#L58

mitzkia:
Thanks, I am ok with the deprecation of original usages. I will create a new patch for that.

modules/riemann/riemann-grammar.ym
```diff
@@ -114,13 +115,13 @@ riemann_option
...
-        | KW_TYPE '(' LL_IDENTIFIER riemann_type_options ')'
+        | KW_TYPE '(' LL_IDENTIFIER riemann_tls_options ')'
           {
             CHECK_ERROR(riemann_dd_set_connection_type(last_driver, $3), @3,
                         "Unknown Riemann connection type: %s", $3);
             free($3);
           }
-        | KW_TYPE '(' LL_STRING riemann_type_options ')'
+        | KW_TYPE '(' LL_STRING riemann_tls_options ')'
           {
```
bazsi: btw, there's a rule called "string" that maps either LL_IDENTIFIER or LL_STRING, so the two rules above can be collapsed into a single one.

kira-syslogng:

mitzkia:
@Kokan as I saw we can only add KWS_OBSOLETE flag to directly options (not to "option groups"). 
Let see in `modules/riemann/riemann-parser.c`:
```
  { "cacert",                   KW_CA_FILE, KWS_OBSOLETE, "The cacert() option is deprecated in favour of ca-file()" },
  { "cert",                     KW_CERT_FILE, KWS_OBSOLETE, "The cert() option is deprecated in favour of cert-file()" },
```
With this method I think we can not "mark" the original usage as obsolete.
If I mark one of the options as obsolete it will be obsolete also in a new way of configuration.

PR #1716, opened 2017-10-13
python-dest: value-pairs reference leak

furiel:
There is a reference leak in python destination when fields are added
to the message through value pairs. The leak happens per message, so
the memory usage of syslog-ng can increase considerably in a short
period of time.

Also, original implementation led to crash when python destination was fed with binary data (for example loggen -U). This patch also checks if the conversion from char*/long to pyobj was successful.

Example configuration:
```
@version: 3.13

destination d_python {
    python(class("Foo") value-pairs(scope(everything)));
};
log {
    source{tcp(port(10514));};
    destination(d_python);
    flags(flow-control);
};

python {
import sys
class Foo(object):
    def init(self, options):
        return True
    def send(self, msg):
        return True
};
```
Fixes: https://github.com/balabit/syslog-ng/issues/1253

### DIFF
--- a/modules/python/python-value-pairs.c
+++ b/modules/python/python-value-pairs.c
@@ -23,10 +23,45 @@
  *
  */
 #include "python-value-pairs.h"
+#include "python-helpers.h"
+#include "messages.h"
 
 /** Value pairs **/
 
-/* TODO escape '\0' when passing down the value */
+static void
+add_long_to_dict(PyObject *dict, const gchar *name, long num)
+{
+  gchar buf[256];
+
+  PyObject *pyobject_to_add = PyLong_FromLong(num);
+  if (!pyobject_to_add)
+    {
+      msg_error("Error while constructing python object",
+                evt_tag_str("exception", _py_fetch_and_format_exception_text(buf, sizeof(buf))));
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


kira-syslogng:

furiel:
As @Kokan pointed out IRL, it is easy to feed python destination with binary data, for example loggen -U (tls), which will lead to PyUnicode_FromString return with NULL, resulting in a crash when adding NULL to the dictionary.
I force pushed a new version when we check the return value of the conversion function as well, and fetch the generated error in case of conversion failure.

kira-syslogng:

kira-syslogng:

bazsi:
Maybe we should feed bytes to python?

furiel:
Right, I will replace PyUnicode with PyBytes to maintain consistency with the index operator in syslog-ng.

furiel:
I force pushed a new version. In the end I used PyBytes_FromStringAndSize, thus I could resolve the following comment in the code: `/* TODO escape '\0' when passing down the value */`
I could test it with the recently added $(binary) template function:
`value-pairs(scope(everything) pair("TESTSTRING" string("TESTS $(binary 0x00) TRING")))`

kira-syslogng:

PR #1717, opened 2017-10-13
Makefile: disable ASLR for Criterion tests

presidento:
The parameterized Criterion tests crash when ASLR is enabled (see https://github.com/Snaipe/Criterion/issues/208), which is the default behavior in Ubuntu since Zesty (17.04). So disable ASLR for tests if Criterion is enabled.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -58,6 +58,10 @@ TEST_LDADD		= $(top_builddir)/libtest/libsyslog-ng-test.a \
 if ENABLE_CRITERION
 TEST_CFLAGS		+= $(CRITERION_CFLAGS)
 TEST_LDADD		+= $(CRITERION_LIBS)
+
+# Disable ASLR, needed for parameterized Criterion tests,
+# see https://github.com/Snaipe/Criterion/issues/208
+AM_TESTS_ENVIRONMENT	+= setarch `uname -m` --addr-no-randomize
 endif
 
 test_ldflags		= -no-install


kira-syslogng:

kira-syslogng:

Makefile.am
```diff
@@ -58,6 +58,10 @@ TEST_LDADD		= $(top_builddir)/libtest/libsyslog-ng-test.a \
 if ENABLE_CRITERION
 TEST_CFLAGS		+= $(CRITERION_CFLAGS)
 TEST_LDADD		+= $(CRITERION_LIBS)
+
+# Disable ASLR, needed for parameterized Criterion tests,
+# see https://github.com/Snaipe/Criterion/issues/208
+AM_TESTS_ENVIRONMENT	+= setarch `uname -m` --addr-no-randomize
```
MrAnno: Running a single test manually will still fail, but I think it is still better than rewriting tests to use [`cr_strdup()`](https://github.com/Snaipe/Criterion/issues/208#issuecomment-301120787).

Kokan: You could use `clang` it works even with ASLR enabled (in the OS) :smiley: 
(The reason https://blog.quarkslab.com/clang-hardening-cheat-sheet.html )

Please also update the dependencies with `util-linux`, and check for `setarch` in case of test (criterion) is enabled.
I would not make it `setarch` hard dependency even if the unit testing is enabled. (e.g. freebsd)

presidento:
TODO: check for setarch on configure time. If it is missing: warning.

presidento:
@furiel said that he will try another solution for ASLR.

presidento:
Closing in favor of #1753.

PR #1721, opened 2017-10-19
cfg: Fix double free/double close crashes

### LINKED ISSUE #1720: Crash when a config file is removed during config reload
# syslog-ng

## Version of syslog-ng
```
syslog-ng 3.10.1                          
```

## Platform
Various

## Debug bundle

```
N/A
```

# Issue

If a config file is removed between an included directory or glob being resolved and the files being opened, a double free can occur. Resulting in a crash

## Failure

```
*[2017-07-18T02:55:55.040698] Error opening include file; filename='', depth='1'
*** Error in `syslog-ng': double free or corruption (!prev): 0x0000000017dca9e0 ***
======= Backtrace: =========
/lib/libc.so.6(+0x97bd0)[0xfff20f9bd0]
/lib/libc.so.6(+0x9f3ac)[0xfff21013ac]
/lib/libc.so.6(+0xa0d40)[0xfff2102d40]
/usr/lib/libsyslog-ng-3.10.so.0(_cfg_lexer_free+0x24)[0xfff251e2bc]
/usr/lib/libsyslog-ng-3.10.so.0(_cfg_lexer__delete_buffer+0x64)[0xfff251e330]
/usr/lib/libsyslog-ng-3.10.so.0(cfg_lexer_free+0x58)[0xfff24f7f58]
/usr/lib/libsyslog-ng-3.10.so.0(cfg_run_parser+0x120)[0xfff24f595c]
/usr/lib/libsyslog-ng-3.10.so.0(cfg_read_config+0x94)[0xfff24f5b88]
/usr/lib/libsyslog-ng-3.10.so.0(main_loop_read_and_init_config+0x58)[0xfff250c100]
syslog-ng(main+0x2a8)[0x10001918]
```

## Steps to reproduce
Where config_file is a file included in a statement similar to "include "syslog-ng.d/":

1. while :; do mv <config_file> <somewhere else>; mv <somewhere_else/config_file> ./; done
2. while :; do syslog-ng-ctl reload; done

## Configuration

```
N/A

```

swstephenson:
Fix the double free of level->yy_buff and double close of
level->file.include_file that can occur during a config reload if a config
file is deleted.

This was observed on a system with frequent dynamic config change. The issue
occurs when a config file is deleted between the glob expansion/directory
scanning in cfg_lexer_include_file_*() functions and files being opened in
cfg_lexer_start_next_include.

Fixes #1720

Signed-off-by: Sam Stephenson <sam.stephenson@alliedtelesis.co.nz>

### DIFF
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -307,8 +307,13 @@ word	[^ \#'"\(\)\{\}\\;\r\n\t,|\.@:]
 <INITIAL><<EOF>>           {
                              if (!cfg_lexer_start_next_include(yyextra))
                                {
-                                 *yylloc = yyextra->include_stack[0].lloc;
-                                 yyterminate();
+                                 if (yyextra->include_depth == 0)
+                                   {
+                                     *yylloc = yyextra->include_stack[0].lloc;
+                                     yyterminate();
+                                   }
+                                 else
+                                    return LL_ERROR;
                                }
                            }
 
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -248,13 +248,17 @@ cfg_lexer_start_next_include(CfgLexer *self)
 
   /* reset the include state, should also handle initial invocations, in which case everything is NULL */
   if (level->yybuf)
-    _cfg_lexer__delete_buffer(level->yybuf, self->state);
+    {
+      _cfg_lexer__delete_buffer(level->yybuf, self->state);
+      level->yybuf = NULL;
+    }
 
   if (level->include_type == CFGI_FILE)
     {
       if (level->file.include_file)
         {
           fclose(level->file.include_file);
+          level->file.include_file = NULL;
         }
     }
 


kira-syslogng:

Kokan:
@kira-syslogng ok to test

kira-syslogng:

gaborznagy:
@swstephenson I have reviewed the PR and found that it indeed solves the double free problem and no crash occurs, but syslog-ng is not handling correctly the error case in cfg_lexer_start_next_include().
When a file read error occurs what happens is that we terminate the lexer and syslog-ng starts without any configuration, doing anything:
```
[2017-11-07T12:59:49.801231] Error opening include file; filename='/home/gabornagy/install/OSE-Issue1720/share/syslog-ng/include/scl/cisco/plugin.conf', depth='2'
[2017-11-07T12:59:49.801341] Unable to detect fully qualified hostname for localhost, use_fqdn() will use the short hostname;
[2017-11-07T12:59:49.801395] Running application hooks; hook='1'
[2017-11-07T12:59:49.801400] Running application hooks; hook='3'
[2017-11-07T12:59:49.801415] syslog-ng starting up; version='3.12.1'
```
I use the same reproduction method which furiel mentioned in #1720:
chmod 000 cisco/plugin.conf

This is not a good way of error handling, syslog-ng should be stopped in such cases, otherwise the user would not spot the issue.

The reason that the return FALSE is not sufficient error reporting is that we also return with FALSE when we finished with including every file:
```
  if (self->include_depth == 0)
    {
      return FALSE;
    }
```
and this is used in lib/cfg-lex.l:
```
if (!cfg_lexer_start_next_include(yyextra))
  {
    *yylloc = yyextra->include_stack[0].lloc;
    yyterminate();
  }
```

Could you resolve this termination part in this PR as well, please?

swstephenson:
Hi @gaborznagy,

If I understand you correctly, I see the following:

If syslog-ng is already running with a config, and one of the files is made unreadable (with the @furiel reproduction method), the reload fails and syslog reverts to it's old config.

If syslog-ng is starting up for the first time and the config load fails (for the same reason) I see it exiting with a non-zero status and an error message that parsing the configuration failed, as below.

```
# syslog-ng -f /etc/syslog-ng/syslog-ng.conf -F
[2017-11-08T16:04:33.331737] Error opening include file; filename='/etc/syslog-ng/conf.d/stream.conf', depth='1'
Error parsing pragma, Error including /etc/syslog-ng/conf.d in /etc/syslog-ng/syslog-ng.conf at line 4, column 10:

@include "/etc/syslog-ng/conf.d"
         ^^^^^^^^^^^^^^^^^^^^^^^

mailing list: https://lists.balabit.hu/mailman/listinfo/syslog-ng
Error parsing config, syntax error, unexpected LL_ERROR, expecting $end in /etc/syslog-ng/syslog-ng.conf at line 4, column 1:

@include "/etc/syslog-ng/conf.d"
^

mailing list: https://lists.balabit.hu/mailman/listinfo/syslog-ng
# echo $?
1
```
Are you getting something different?

(For getting notification of the failure I would have thought the messages and non-zero exit status should be sufficient. Process supervisors should be able to deal with this as well. And #1739 should make it possible to directly detect the reload failure case)

gaborznagy:
Yes I have different scenario than you.
The reason is that I have reproduced the issue with an invalid scl file which is included through globbing in scl.conf
(repr. steps like before: chmod 000 scl/cisco/plugin.conf).

In your case it is an invalid config file, which is included directly from the config file.
If I reproduce the issue the same way as you did, I have the same result.

gaborznagy:
HI @swstephenson !

We have found the source why syslog-ng is not failing when an include fails inside a glob (thanks too Kokan! ).
This error was not discovered before, thank you for your contribution, without your PR it would be still undiscovered!

So what we have found is that when the include procedure returns False (i.e. stop including more files) and the include depth shows that we were not finished we report this as an error state.
I have pasted the diff of the patch.

```diff
diff --git a/lib/cfg-lex.l b/lib/cfg-lex.l
index c16155b..243837e 100644
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -306,8 +306,13 @@ word       [^ \#'"\(\)\{\}\\;\r\n\t,|\.@:]
 <INITIAL><<EOF>>           {
                              if (!cfg_lexer_start_next_include(yyextra))
                                {
-                                 *yylloc = yyextra->include_stack[0].lloc;
-                                 yyterminate();
+                                 if (yyextra->include_depth == 0)
+                                   {
+                                     *yylloc = yyextra->include_stack[0].lloc;
+                                     yyterminate();
+                                   }
+                                 else
+                                    return LL_ERROR;
```

What do you think about this solution?
I think if this patch would be integrated then all error cases would be covered and your PR could be merged too.

swstephenson:
Right, I'm with you now. I was able to reproduce as you describe. Your patch seems to do the trick. Glad this prompted the discovery of another issue!

kira-syslogng:

kira-syslogng:

Issue #1425, opened 2017-04-10
memory keeps increasing when fail to write log caused by no device space left


ylse:
## Version of syslog-ng
syslog-ng version 3.8.1
```
output of syslog-ng --version

syslog-ng 3.8.1
Installer-Version: 3.8.1
Revision: 
Module-Directory: /usr/lib/syslog-ng
Module-Path: /usr/lib/syslog-ng
Available-Modules: syslogformat,kvformat,afuser,add-contextual-data,afprog,afstomp,sdjournal,dbparser,pseudofile,disk-buffer,afsocket,basicfuncs,system-source,cryptofuncs,confgen,graphite,date,cef,affile,linux-kmsg-format,csvparser
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: off
Enable-TCP-Wrapper: off
Enable-Linux-Caps: off

```

## Platform

Linux version 4.9.18

# Issue
We find that the memory used by process of syslog-ng keeps on increasing when it fails to write log because of no space left on device. It can be reproduced by filling device space and then continuing printing log of Level 0/EMERG.   

## Failure
The log will record some error as below
```
 I/O error occurred while writing; fd='17', error='No space left on device (28)'
 I/O error occurred while writing; fd='19', error='No space left on device (28)'
```

The memory of process syslog-ng rises up steadily. 

## Steps to reproduce

1. Start syslog-ng.

2. Keep on print EMERG level logs by <logger -p 0 "This is a emergency log">.

3. Fill the space with data on the device where log files are stored.

4. It is observed that the memory occupied by process syslog-ng keeps rising.

## Configuration

ordinary configuration for syslog-ng

bazsi:
Syslog-ng normally buffers messages, up to the value of log-fifo-size
().Once that's​ reached it should become steady.

ylse:
Thanks for the comment. I find the memory occupied by process syslog-ng increased from 50M to 90M after no space left and it is too large that the process is killed. So I hope to know how I can make the buffer smaller to avoid the process being killed.

ylse:
Is it the set by Option log-fifo-size() described in the Guide? I didn't set it and it should be the default value 10000. I think 40M(increased memory) is a little too large for 10000 messages(buffered).

bazsi:
Well, the internal representation of a message is usually larger than the
raw data. 2-4kb per message is realistic, but it also depends on the config
[...]

PR #1725, opened 2017-10-19
affile: Mark messages as consumed in a write failure.

swstephenson:
Fix a crash that occurs following a write failure when the file writer
buffer becomes full and a config reload is triggered.

If a write failure occurs in log_proto_file_writer_post as the buffer
becomes full, a subsequent call will be made where msg and
self->buffer[self->buff_count].iov_base point to the same memory.

The first call to log_proto_file_writer_flush will write the message
and free the buffer. The following call will raise an abort signal
when it attempts to free the buffer a second time.

If the message has been stored in self->buffer, it should be marked
as consumed irrespective of the result of log_proto_file_writer_flush.

This may relate to #1425

Signed-off-by: Sam Stephenson <sam.stephenson@alliedtelesis.co.nz>

### DIFF
--- a/modules/affile/logproto-file-writer.c
+++ b/modules/affile/logproto-file-writer.c
@@ -179,16 +179,15 @@ log_proto_file_writer_post(LogProtoClient *s, guchar *msg, gsize msg_len, gboole
   ++self->buf_count;
   self->sum_len += msg_len;
 
+  *consumed = TRUE;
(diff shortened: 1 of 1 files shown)


kira-syslogng:

bazsi:
@kira-syslogng : ok to test

kira-syslogng:

PR #1729, opened 2017-10-20
 logreader/logproto: remove preemption

MrAnno:
The "preemption" logic with NC_FILE_SKIP was an implementation detail of the old wildcard file source.
This is dead code now.

This PR removes code from a "hot path", but I guess this is not a significant performance improvement. :)

### DIFF
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -35,7 +35,6 @@ struct _LogReader
   LogSource super;
   LogProtoServer *proto;
   gboolean immediate_check;
-  gboolean waiting_for_preemption;
   LogPipe *control;
   LogReaderOptions *options;
   PollEvents *poll_events;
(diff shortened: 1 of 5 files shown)


MrAnno:
This PR removes code from a "hot path", but I guess this is not a significant performance improvement. :)

kira-syslogng:

MrAnno:
Questions to reviewers:
- Why could I change the value of `NC_REOPEN_REQUIRED`?
- Can I do the same with `LR_THREADED` (consider persist)?

lib/logreader.c
```diff
@@ -329,9 +328,6 @@ log_reader_fetch_log(LogReader *self)
   gboolean may_read = TRUE;
```
Kokan: Would it be possible to also remove `may_read` ?

MrAnno: It is a requirement of "multi-read" which is a used feature in `affile/linux-kmsg.c`.

Kokan: Yeah, you are right.

MrAnno:
Oops, I forgot to make this WIP. I'll fix the value of LR_THREADED in my next PR.

Kokan:
A merged PR is a good PR :)

MrAnno:
😄

Issue #1714, opened 2017-10-12
Segfault when accessing patterndb macro within Java destination


flix-:
## Version of syslog-ng
Tested with 3.10.1, 3.11.1, 3.12.1

## Platform
CentOS Linux release 7.3.1611 (Core)

## Debug Environment
Docker Image (see below)

## Issue
If an indirect value (e.g. a macro that has been set in a patterndb) is accessed within a Java destination through logMessage.getValue(\<MACRONAME\>) a SEGFAULT occurs. Values directly set within patterndb (e.g. through \<value\>) or macros set by syslog-ng such as ${HOST} are not affected.

## Failure
```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007f9afed1b0fd, pid=841, tid=0x00007f9ae5559700
#
# JRE version: OpenJDK Runtime Environment (8.0_144-b01) (build 1.8.0_144-b01)
# Java VM: OpenJDK 64-Bit Server VM (25.144-b01 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libsyslog-ng-3.11.so.0+0x730fd]  nv_table_resolve_indirect+0xdd
#
# Core dump written. Default location: /tmp/core or core.841
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#

---------------  T H R E A D  ---------------

Current thread (0x00007f9ab0004000):  JavaThread "pB" [_thread_in_native, id=852, stack(0x00007f9ae5459000,0x00007f9ae555a000)]

siginfo: si_signo: 11 (SIGSEGV), si_code: 1 (SEGV_MAPERR), si_addr: 0x0000000000000000

Registers:
RAX=0x00007f9ab4014801, RBX=0x00007f9ab4014620, RCX=0x000000000000000b, RDX=0x0000000000000000
RSP=0x00007f9ae5556270, RBP=0x00007f9ab4014698, RSI=0x0000000000000000, RDI=0x000000000000000b
R8 =0x0000000000000003, R9 =0x0000000000000003, R10=0x00007f9ae5556030, R11=0x00007f9afed1b020
R12=0x0000000000000000, R13=0x00007f9ab4014620, R14=0x00007f9ae5556388, R15=0x00007f9ab0004000
[...]

PR #1732, opened 2017-10-22
This fixes access to indirect values within Java destination

flix-:
Related issue: #1714

### DIFF
--- a/modules/java/proxies/java-logmsg-proxy.c
+++ b/modules/java/proxies/java-logmsg-proxy.c
@@ -25,9 +25,8 @@
 #include "java-logmsg-proxy.h"
 #include "java_machine.h"
 #include "messages.h"
+#include "str-utils.h"
 #include "logmsg/logmsg.h"
-
-#include <string.h>
 
 
 #define LOG_MESSAGE "org.syslog_ng.LogMessage"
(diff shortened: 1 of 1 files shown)


kira-syslogng:

Kokan:
@kira-syslogng  ok to test

kira-syslogng:

modules/java/proxies/java-logmsg-proxy.c
```diff
@@ -49,20 +51,24 @@ Java_org_syslog_1ng_LogMessage_getValue(JNIEnv *env, jobject obj, jlong handle,
...
 
-  value = log_msg_get_value_by_name(msg, name_str, NULL);
+  value = log_msg_get_value_by_name(msg, name_str, &value_len);
 
   (*env)->ReleaseStringUTFChars(env, name, name_str);
 
   if (value)
     {
-      return (*env)->NewStringUTF(env, value);
+      gchar stripped_value[value_len + 1];
```
Kokan: I would avoid the allocation of string on the stack, the `value_len` could be big.
You could use the `g_strndup` instead, with that the `\0` termination is guaranteed, although you must `g_free` it after thew `NewStringUTF` call.

Something like that:
```C
gchar *stripped_value = g_strndup(value, value_len);
jstring java_string = (*env)->NewStringUTF(env, stripped_value);
g_free(stripped_value);
return java_string;
```

Note: with this the `#include <string.h>` not needed.

bazsi: strndup() can be slow, and this might be a fastpath, although I am not sure how "fast" it is considering it is called from Java.

Anyway, there's a macro called APPEND_ZERO() which checks if the specific name-value pair is zero terminated already, and avoids the copying in that case. Still, that would allocate memory on the stack, which might be limited to couple of megabytes. but if that's the case, the remaining call sites of APPEND_ZERO would be affected as well.

kira-syslogng:

PR #1735, opened 2017-10-24
Fix distdir target in case of system option

Kokan:
The following dependencies (as git submodules) were optional to include into the dist compressed files:
* ivykis
* librabbitmq-c
* mongoc

This should be only in case of internal option, and not in case of system, as the system shall provide those files.

Signed-off-by: kokan <peter.kokai@balabit.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -233,7 +233,7 @@ AC_ARG_ENABLE(mongodb,
 AC_ARG_WITH(mongoc,
               [  --with-mongoc=[system/internal/auto/no]
                                          Link against the system supplied or the built-in mongo-c-driver library. (default: auto)]
-              ,,with_mongoc="auto")
+              ,,with_mongoc="internal")
 
 AC_ARG_ENABLE(legacy-mongodb-options,
               [  --enable-legacy-mongodb-options=[yes/no]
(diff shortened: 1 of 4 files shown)


kira-syslogng:

MrAnno:
Please not that we've fixed issues such as #1204 previously, to always include all submodules in the dist tarball (no matter how syslog-ng was configured before the `make dist` command).

This PR violates this idea.

MrAnno:
- mongoc: auto (system by default if exists)
- jsonc: auto (system by default if exists)

kira-syslogng:

Kokan:
The *make dist* in the current format creates a compressed file containing a few extra files with dependencies (also attached as git submodules). Therefore a compressed file exists, where all of the *./configure* option shall work. But those extra dependencies are just external project. It is nice to keep tracking the working version in such way, and to ensure the proper function. Also a great place for confusion, the user could want more control such as not included next to syslog-ng addition files such as *jsonc* or *rabbitmq-c*. The *autogen.sh* option for `skip_submodules` also points to this direction, which is already part of the syslog-ng. Of course the default is to include those files.

lbudai:
@Kokan : so the default distribution tarball contains everything, as before, but we have the chance for customizing the dist. tarball (what we currently need in some of our internal build environments).

lbudai:
it could be interest to you (as I know, you are build your own dist tarballs) :
@czanik 
@mochrul

czanik:
@lbudai : Thanks! Tested and I had no problem building syslog-ng from my own dist tarball on openSUSE / SLES / EPEL7

PR #1739, opened 2017-10-25
syslog-ng-ctl: implement synchronous reload (--wait option to reload …

lbudai:
…command)

When `syslog-ng-ctl reload --wait` is called, syslog-ng-ctl will be blocked
until reload finished. So don't need any log parsing to decide whether
the reload processed is finished or not.

Technical details:
 * add SYNCRELOAD command
 * when SYNCRELOAD is called, an AH_POST_CONFIG_LOADED callback is registered
 * a condvar is waiting for a signal emitted by our
   AH_POST_CONFIG_LOADED callback
 * after the signal is received, answer is sent back to the
   syslog-ng-ctl client

Future devlopment could be:
 * send back status of the reload to the user
 * define a condwait timeout at syslog-ng-ctl site

Questions to Reviewer(s):
 * is it possible to miss an `AH_POST_CONFIG_LOADED`? (`control_connection_reload` is called _before_ registering the callback...)
 * is it possible to rename `SYNCRELOAD` to `RELOAD_SYNC`? when can we use `RELOAD_SYNC`?
 * how could we avoid using conditional variables (what should we have to refactor)?
 * is it possible that our registered callback will be remain active after it is called? (what if we send a SIGHUP after we run `syslog-ng-ctl reload --wait` ? )

(please note that these questions is for helping the reviewers, if I won't get answers within two weeks, I will answer these questions :) )

### DIFF
--- a/lib/apphook.h
+++ b/lib/apphook.h
@@ -41,6 +41,7 @@ enum
 typedef void (*ApplicationHookFunc)(gint type, gpointer user_data);
 
 void register_application_hook(gint type, ApplicationHookFunc func, gpointer user_data);
+void register_application_hook_without_checking_current_state(gint type, ApplicationHookFunc func, gpointer user_data);
 void app_startup(void);
 void app_finish_app_startup_after_cfg_init(void);
 void app_post_daemonized(void);
--- a/lib/control/control-commands.c
+++ b/lib/control/control-commands.c
@@ -111,13 +111,30 @@ control_connection_stop_process(GString *command, gpointer user_data)
   return result;
 }
 
+
+static gboolean
+_reload(GString *command, gpointer user_data, GString **result)
+{
+  MainLoop *main_loop = (MainLoop *) user_data;
+  gboolean reload_initiated = main_loop_reload_config(main_loop);
+
+  if (reload_initiated)
+    {
+      *result = g_string_new("OK Config reload initiated");
+    }
+  else
+    {
+      *result = g_string_new("FAIL Failed to request config reload");
+    }
+
+  return reload_initiated;
+}
+
 static GString *
 control_connection_reload(GString *command, gpointer user_data)
 {
-  GString *result = g_string_new("OK Config reload initiated");
-  MainLoop *main_loop = (MainLoop *) user_data;
-
-  main_loop_reload_config(main_loop);
+  GString *result = NULL;
+  _reload(command, user_data, &result);
   return result;
 }
 
@@ -185,8 +202,14 @@ _wait_for_syslog_ng_reload()
 static GString *
 control_connection_reload_sync(GString *command, gpointer user_data)
 {
-  GString *result = control_connection_reload(command, user_data);
+  GString *result = NULL;
+  gboolean success = _reload(command, user_data, &result);
+
+  if (!success)
+    return result;
+
   _wait_for_syslog_ng_reload();
+
   return result;
 }
 
--- a/lib/control/control-commands.c
+++ b/lib/control/control-commands.c
@@ -121,6 +121,75 @@ control_connection_reload(GString *command, gpointer user_data)
   return result;
 }
 
+typedef struct _ReloadCondvar
+{
+  GMutex mutex;
+  GCond condvar;
+  gboolean reloaded;
+} ReloadCondvar;
+
+static ReloadCondvar *
+_reload_condvar_new()
+{
+  ReloadCondvar *self = g_new0(ReloadCondvar, 1);
+  g_mutex_init(&self->mutex);
+  g_cond_init(&self->condvar);
+  self->reloaded = FALSE;
+
+  return self;
+}
+
+static void
+_reload_condvar_free(ReloadCondvar *self)
+{
+  g_mutex_clear(&self->mutex);
+  g_cond_clear(&self->condvar);
+  g_free(self);
+}
+
+static void
+_reload_condvar_emit(ReloadCondvar *self)
+{
+  g_mutex_lock(&self->mutex);
+  self->reloaded = TRUE;
+  g_cond_signal(&self->condvar);
+  g_mutex_unlock(&self->mutex);
+}
+
+static void
+_reload_condvar_wait(ReloadCondvar *self)
+{
+  g_mutex_lock(&self->mutex);
+  while (!self->reloaded)
+    g_cond_wait(&self->condvar, &self->mutex);
+  self->reloaded = FALSE;
+  g_mutex_unlock(&self->mutex);
+}
+
+static void
+_config_reloaded(gint hook_type, gpointer user_data)
+{
+  ReloadCondvar *self = (ReloadCondvar *)user_data;
+  _reload_condvar_emit(self);
+}
+
+static void
+_wait_for_syslog_ng_reload()
+{
+  ReloadCondvar *reload_condvar = _reload_condvar_new();
+  register_application_hook_without_checking_current_state(AH_POST_CONFIG_LOADED, _config_reloaded, reload_condvar);
+  _reload_condvar_wait(reload_condvar);
+  _reload_condvar_free(reload_condvar);
+}
+
+static GString *
+control_connection_reload_sync(GString *command, gpointer user_data)
+{
+  GString *result = control_connection_reload(command, user_data);
+  _wait_for_syslog_ng_reload();
+  return result;
+}
+
 static GString *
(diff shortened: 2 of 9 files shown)


kira-syslogng:

kira-syslogng:

kira-syslogng:

lbudai:
do not merge, I want to squash the last two commits on tomorrow(I’m modifying code by using my mobile and don’t have a git client)

kira-syslogng:

kira-syslogng:

lbudai:
Now, I've rebased the code, so I think it is ready to merge from my part.

kira-syslogng:

lbudai:
With @furiel we identified a conceptual problem. AH_POST_CONFIG_LOADED is not a good candidate for catching the momentum when config is reloaded as the ‘runlevel’ is not set to a lower level at the time when I’m registering the apphook(which results in an immediate call of the registered function).

I found another issue regarding app hooks: there is an enum which defines an order to the AH values but the ordering can be skipped by passing a false parameter to run_application_hook(this is why AH_REOPEN works, I guess) . The problem is that this boolean parameter is not about ordering/runlevel but requesting a single-shot call (after the call the function is removed from the apphook list).

lbudai:
@furiel: finally I run control server in a separate thread... (this is something we wanted to do some months ago when we initiated the stats query)

kira-syslogng:

lib/mainloop.c
```diff
@@ -468,7 +468,7 @@ main_loop_init(MainLoop *self, MainLoopOptions *options)
 
   main_loop_init_events(self);
   if (!self->options->syntax_only)
-    control_init(self, resolvedConfigurablePaths.ctlfilename);
+    control_thread_start(self, resolvedConfigurablePaths.ctlfilename);
   setup_signals(self);
```
bazsi: Do we need to be explicit that the control subsystem is in a thread? I think thats an implementation detail.

lib/control/control-main.c
```diff
@@ -26,18 +26,61 @@
...
 
 void
-control_init(MainLoop *main_loop, const gchar *control_name)
+control_thread_start(MainLoop *main_loop, const gchar *control_name)
 {
-  control_server = control_server_new(control_name,
-                                      control_register_default_commands(main_loop));
-  control_server_start(control_server);
+  control_server_thread.main_loop = main_loop;
+  control_server_thread.control_name = control_name;
```
bazsi: Dont we need to ref the main loop instance?

lib/control/control-main.c
```diff
@@ -26,18 +26,61 @@
...
+_thread(gpointer user_data)
+{
+  ControlServerThread *cs_thread = (ControlServerThread *)user_data;
+  ControlServer *control_server = control_server_new(cs_thread->control_name,
+                                                     control_register_default_commands(cs_thread->main_loop));
+  iv_init();
+  {
+    _register_stop_requested_event(cs_thread);
+    control_server_start(control_server);
+    iv_main();
```
bazsi: I would probably extract these init init/deinit functions, rather than inlineing them here.

lib/control/control-main.c
```diff
@@ -26,18 +26,61 @@
...
+
+typedef struct _ControlServerThread
+{
+  GThread *thread;
+  MainLoop *main_loop;
+  const gchar *control_name;
+  struct iv_event stop_requested;
+} ControlServerThread;
+
+static ControlServerThread control_server_thread;
```
bazsi: Instead of tbe static variable,cant we store this in globalconfig? As I see this is strictly tied to onr mainloop instance.

Hmm now as i think of it, it may not be true though, the control socket is probably arching over multiple globalconfig instances. If this is the case, then it should probably be moved to the main program.

lib/apphook.h
```diff
@@ -41,6 +41,8 @@ enum
 typedef void (*ApplicationHookFunc)(gint type, gpointer user_data);
 
 void register_application_hook(gint type, ApplicationHookFunc func, gpointer user_data);
+void register_future_application_hook(gint type, ApplicationHookFunc func, gpointer user_data);
+void append_application_hook(gint type, ApplicationHookFunc func, gpointer user_data);
 void app_startup(void);
```
bazsi: I dont see an implementation for the append function.

I dont understant the concept of "future", maybe a comment is due? Is there a better term? What does it do exactly?

kira-syslogng:

lib/apphook.h
lbudai: Well... the append_application_hook is removed, just didn't push every change.
Naming: maybe the best would be to rename the original register_application_hook as it hurts SRP: it not just register a callback but also run it sometimes.

lib/control/control-main.c
lbudai: ok, I can move it to main

lib/control/control-main.c
lbudai: This implementation has a pro over the extracted init/deinit: no need to store the control_server, it is a local variable. I think it is as readable as the other, but I'll factor out init/deinit functions.

lib/control/control-main.c
lbudai: Only if the control server can run after mainloop is deinited.
Currently this is not possible (the thread is joinable and I call the join function in mainloop_deinit).
After I refactor the code(and move control server to main), it depends on when I call the mainloop_deinit.  I can extend Mainloop with a refctr logic.

lib/mainloop.c
lbudai: ControlServerLoop?

kira-syslogng:

kira-syslogng:

lib/apphook.h
bazsi: register_application_hook() would run that function only if that specific state has already passed. I can't remember the use-case exactly.

hmmm... This is the possible offender in afstreams.c:

```
      if (self->door_filename)
        {

          /* door creation is deferred, because it creates threads which is
           * not inherited through forks, and syslog-ng forks during
           * startup, but _after_ the configuration was initialized */

          register_application_hook(AH_POST_DAEMONIZED, afstreams_init_door, self);
        }

```

If syslog-ng is already in the background (e.g. during reload), that afstreams_init_door() should be called immediately. If it is starting up, it should be deferred.

Anyway, apphook probably deserves some care, I am not sure if the right time is now. But like I said, "future" hook is not a term  I easily understand.

bazsi: Ah, I see it was renamed to register_application_hook_without_checking_current_state()

It's better. Let me try to understand the logic again ...

kira-syslogng:

kira-syslogng:

kira-syslogng:

mitzkia:
I am tested manually with running after each other `syslog-ng-ctl reload --wait` and `syslog-ng-ctl stop` for a few time. syslog-ng can handle reload and stop correctly.
latest commit id: 5fb2f97db2591abe67c13a803e93cdfa50e4e098

lib/control/control-commands.c
```diff
@@ -111,13 +111,105 @@ control_connection_stop_process(GString *command, gpointer user_data)
...
+  g_cond_signal(&self->condvar);
+  g_mutex_unlock(&self->mutex);
+}
+
+static void
+_reload_condvar_wait(ReloadCondvar *self)
+{
+  g_mutex_lock(&self->mutex);
+  while (!self->reloaded)
+    g_cond_wait(&self->condvar, &self->mutex);
```
MrAnno: This condition is never met in case of an invalid configuration, because `_config_reloaded` is not triggered in that case. `app_post_config_loaded()` is not called in multiple cases, for example, when `cfg_init` fails:

https://github.com/balabit/syslog-ng/blob/3798aa4dc30e6ff60c8db39b07b9944ddb8da6dc/lib/mainloop.c#L227-L259

Or when `cfg_read_config` fails:
https://github.com/balabit/syslog-ng/blob/3798aa4dc30e6ff60c8db39b07b9944ddb8da6dc/lib/mainloop.c#L293-L300

Reproduction:
- start syslog-ng with a correct config
- add a syntax error to the configuration
- execute `sbin/syslog-ng-ctl reload --wait`

As a result, syslog-ng can't be stopped or reloaded peacefully.

faxm0dem:
will the command return a RC different from 0 if the configuration is invalid?

lbudai:
Currently I do not have the time to finish this, so I'm closing it.

PR #1740, opened 2017-10-27
syslog-ng: extend '--version' with config version

lbudai:
Someone might want to check what is the current config version.

@mitzkia : this is what we talked about yesterday ;)


kira-syslogng:

bazsi:
Is this different from the syslog-ng version number?

lbudai:
@bazsi: this PR has testing reasons and related to rolling versions and nightly testing (eg.: currently we have 3.13 config version while syslog-ng version is 3.12). 

If the numbers are the same, then we should have to derive CONFIG_VERSION from VERSION, otherwise we need to test if the versions are the same.

Issue #1736, opened 2017-10-25
stdin driver crashes with reload


furiel:
Example configuration:
```
@version: 3.13

log {
    source { stdin();};
};
```
After calling syslong-ctl reload: syslog-ng crashes with core dump:
```
#0  0x00007fd01fda4767 in kill () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007fd0206b37a6 in main_loop_reload_config_apply (user_data=0x7fd020954d20 <main_loop>)
    at ../syslog-ng/lib/mainloop.c:246
#2  0x00007fd0206b4c5c in _consume_action (action=0x907b90, dummy=0x0) at ../syslog-ng/lib/mainloop-worker.c:229
#3  0x00007fd0203b005c in g_queue_foreach () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#4  0x00007fd0206b4c8d in _invoke_sync_call_actions () at ../syslog-ng/lib/mainloop-worker.c:236
#5  0x00007fd0206b4fc8 in main_loop_worker_sync_call (func=0x7fd0206b358c <main_loop_reload_config_apply>, 
    user_data=0x7fd020954d20 <main_loop>) at ../syslog-ng/lib/mainloop-worker.c:383
#6  0x00007fd0206b39b3 in main_loop_reload_config_initiate (user_data=0x7fd020954d20 <main_loop>)
    at ../syslog-ng/lib/mainloop.c:298
#7  0x00007fd0207045c7 in __iv_event_run_pending_events (_st=0x8b2520)
    at ../../../../syslog-ng/lib/ivykis/src/iv_event.c:56
#8  0x00007fd020704cf0 in iv_run_tasks (st=0x8b2520) at ../../../../syslog-ng/lib/ivykis/src/iv_task.c:48
#9  0x00007fd02070865e in iv_main () at ../../../../syslog-ng/lib/ivykis/src/iv_main_posix.c:99
#10 0x00007fd0206b40bc in main_loop_run (self=0x7fd020954d20 <main_loop>) at ../syslog-ng/lib/mainloop.c:539
#11 0x0000000000401de4 in main (argc=1, argv=0x7fff5e845848) at ../syslog-ng/syslog-ng/main.c:303
(gdb) bt full
#0  0x00007fd01fda4767 in kill () at ../sysdeps/unix/syscall-template.S:84
No locals.
[...]

PR #1741, opened 2017-10-27
stdin: crash with syslog-ng reload

furiel:
In case stdin() driver is used: syslog-ng closes the zero file
descriptor (stdin) before restarting. After restart this same zero
file descriptor is intended to be added back to epoll in
ivykis. Being already closed: epoll_ctl returns with error:
errno=EBADF.  Hence syslog-ng cannot initialize after reload, and
when she tries to fallback, initialization will fail again with
the same reason. Hence the crash.

This patch duplicates the file descriptor of stdin, and the dupped
version will be watched, instead of zero.

Fixes: https://github.com/balabit/syslog-ng/issues/1736

### DIFF
--- a/modules/affile/stdin.c
+++ b/modules/affile/stdin.c
@@ -24,6 +24,8 @@
 #include "transport/transport-file.h"
 #include "affile-source.h"
 
+#include <unistd.h>
+
 static LogTransport *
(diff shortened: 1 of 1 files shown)


kira-syslogng:

PR #1743, opened 2017-10-27
Limit dynamic counters

lbudai:
##### Add config option for limiting dynamic counters
*   Users can limit the number of registered dynamic counters via the `stats_max_dynamics` global option.
    This is not a limit for the number of counters but for the number of _clusters_ which means that users
    can limit the number of different dynamic value occurrences (as each
    cluster has a pre-allocated array of counters).
*   Note that after reload, when users set a lower limit,
    they have to wait for stat-freq-timer being expired to remove the
    previously allocated dynamic clusters.
 *   Dynamic values are
      * source host (from stats level 2)
      * sender host (host from) (from stats level 3)
      * program (from stats level 3)
    
 * By default this change is backward compatible: if someone not uses this feature, number of dynamic counters won't be limited.
 * It is  possible to disable dynamic counters by setting the limit to 0.

##### The following example contains 6 different dynamic values (a sender, a  host and four different programs):

             
```                                                                                                         
    src.sender;;localhost;d;processed;4
    src.sender;;localhost;d;stamp;1509121934
    src.program;;P-18069;d;processed;1
    src.program;;P-18069;d;stamp;1509121933
    src.program;;P-21491;d;processed;1
    src.program;;P-21491;d;stamp;1509121934
    src.program;;P-9774;d;processed;1
    src.program;;P-9774;d;stamp;1509121919
    src.program;;P-14737;d;processed;1
    src.program;;P-14737;d;stamp;1509121931
    src.host;;localhost;d;processed;4
    src.host;;localhost;d;stamp;1509121934
```

Questions to reviewers:
 * What do you think about storing dynamic counters in a separate hash table?
 * What do you think about the reload-issue? (when the user sets a lower value, after a reload the already registered dynamic counters won't be removed until start-freq timer not expires)?

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -192,6 +192,7 @@ extern struct _StatsOptions *last_stats_options;
 %token KW_MARK_MODE                   10081
 %token KW_ENCODING                    10082
 %token KW_TYPE                        10083
+%token KW_STATS_MAX_DYNAMIC           10084
 
 %token KW_CHAIN_HOSTNAMES             10090
 %token KW_NORMALIZE_HOSTNAMES         10091
@@ -931,6 +932,7 @@ stat_option
 	: KW_STATS_FREQ '(' nonnegative_integer ')'          { last_stats_options->log_freq = $3; }
 	| KW_STATS_LEVEL '(' nonnegative_integer ')'         { last_stats_options->level = $3; }
 	| KW_STATS_LIFETIME '(' positive_integer ')'      { last_stats_options->lifetime = $3; }
+  | KW_STATS_MAX_DYNAMIC '(' nonnegative_integer ')'   { last_stats_options->max_dynamic = $3; }
 	;
 
 dns_cache_option
--- a/lib/cfg-parser.c
+++ b/lib/cfg-parser.c
@@ -86,6 +86,7 @@ static CfgLexerKeyword main_keywords[] =
   { "stats_lifetime",     KW_STATS_LIFETIME },
   { "stats_level",        KW_STATS_LEVEL },
   { "stats",              KW_STATS_FREQ, KWS_OBSOLETE, "stats_freq" },
+  { "stats_max_dynamics", KW_STATS_MAX_DYNAMIC },
   { "flush_lines",        KW_FLUSH_LINES },
   { "flush_timeout",      KW_FLUSH_TIMEOUT },
   { "suppress",           KW_SUPPRESS },
--- a/lib/stats/stats-cluster.c
+++ b/lib/stats/stats-cluster.c
@@ -264,6 +264,15 @@ stats_cluster_new(const StatsClusterKey *key)
   return self;
 }
 
+StatsCluster *
+stats_cluster_dynamic_new(const StatsClusterKey *key)
+{
+  StatsCluster *sc = stats_cluster_new(key);
+  sc->dynamic = TRUE;
+
+  return sc;
+}
+
 void
 stats_counter_group_free(StatsCounterGroup *self)
 {
--- a/lib/stats/stats-registry.c
+++ b/lib/stats/stats-registry.c
@@ -34,6 +34,12 @@ typedef struct _StatsClusterContainer
 
 static StatsClusterContainer stats_cluster_container;
 
+static guint
+_number_of_dynamic_clusters(void)
+{
+  return g_hash_table_size(stats_cluster_container.dynamic_clusters);
+}
+
 static GStaticMutex stats_mutex = G_STATIC_MUTEX_INIT;
 gboolean stats_locked;
 
@@ -68,8 +74,15 @@ _grab_dynamic_cluster(const StatsClusterKey *sc_key)
   sc = g_hash_table_lookup(stats_cluster_container.dynamic_clusters, sc_key);
   if (!sc)
     {
+      if (!stats_check_dynamic_clusters_limit(_number_of_dynamic_clusters()))
+        return NULL;
       sc = stats_cluster_dynamic_new(sc_key);
       _insert_cluster(sc);
+      if ( !stats_check_dynamic_clusters_limit(_number_of_dynamic_clusters()))
+        {
+          msg_warning("Number of dynamic cluster limit has been reached.",
+                      evt_tag_int("allowed_clusters", stats_number_of_dynamic_clusters_limit()));
+        }
     }
 
   return sc;
--- a/lib/stats/stats-registry.c
+++ b/lib/stats/stats-registry.c
@@ -26,14 +26,24 @@
 #include "cfg.h"
 #include <string.h>
 
-static GHashTable *stats_cluster_container;
+typedef struct _StatsClusterContainer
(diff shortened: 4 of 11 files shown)


kira-syslogng:

kira-syslogng:

kira-syslogng:

lib/stats/stats-registry.c
```diff
@@ -51,31 +67,65 @@ stats_unlock(void)
...
-  if (!stats_check_level(stats_level))
-    return NULL;
+  sc = g_hash_table_lookup(stats_cluster_container.dynamic_clusters, sc_key);
+  if (!sc)
+    {
+      if (!stats_check_dynamic_clusters_limit(_number_of_dynamic_clusters()))
+        return NULL;
+      sc = stats_cluster_dynamic_new(sc_key);
+      _insert_cluster(sc);
+      if ( !stats_check_dynamic_clusters_limit(_number_of_dynamic_clusters()))
```
bodnartibi: unnecessary space here

lbudai: Hmm... we have a tool that  checks style and blocks merge until we fix format issues.
So I wouldn't fix it here (fixing astyle would be better as currently we don't check this kind of format issue and it is possible that there are other parts of the code which is affected by the same problem).

bodnartibi:
Looks good to me.

lib/stats/stats-registry.c
```diff
@@ -26,14 +26,30 @@
...
+typedef struct _StatsClusterContainer
+{
+  GHashTable *static_clusters;
+  GHashTable *dynamic_clusters;
+} StatsClusterContainer;
+
+static StatsClusterContainer stats_cluster_container;
+
+static guint
+_number_of_dynamic_clusters()
```
bazsi: Missing void in the parameter list

lib/stats/stats-registry.c
```diff
@@ -51,31 +67,65 @@ stats_unlock(void)
...
+    {
+      if (!stats_check_dynamic_clusters_limit(_number_of_dynamic_clusters()))
+        return NULL;
+      sc = stats_cluster_dynamic_new(sc_key);
+      _insert_cluster(sc);
+      if ( !stats_check_dynamic_clusters_limit(_number_of_dynamic_clusters()))
+        {
+          msg_warning("Number of dynamic cluster limit has been reached.",
+                      evt_tag_int("allowed clusters", stats_number_of_dynamic_clusters_limit()));
+        }
```
bazsi: Please use underscore in the tag name.

lib/stats/stats-registry.c
lbudai: done

lib/stats/stats-registry.c
lbudai: done

kira-syslogng:

lib/cfg-grammar.y
```diff
@@ -931,6 +932,7 @@ stat_option
 	: KW_STATS_FREQ '(' nonnegative_integer ')'          { last_stats_options->log_freq = $3; }
 	| KW_STATS_LEVEL '(' nonnegative_integer ')'         { last_stats_options->level = $3; }
 	| KW_STATS_LIFETIME '(' positive_integer ')'      { last_stats_options->lifetime = $3; }
+  | KW_STATS_MAX_DYNAMIC '(' nonnegative_integer ')'   { last_stats_options->max_dynamic = $3; }
```
presidento: What is the default value of this option? What will be on the documentation? As I see:

- `>0` means the limit of the dynamic stat clusters
- `0` means disable storing dynamic stat clusters
- `-1` means there is no limit - but it is not a valid value here...

lib/stats/stats-registry.c
```diff
@@ -51,31 +61,58 @@ stats_unlock(void)
...
-    {
-      /* check that we are not overwriting a dynamic counter with a
-       * non-dynamic one or vica versa.  This could only happen if the same
-       * key is used for both a dynamic counter and a non-dynamic one, which
-       * is a programming error */
+    sc = _grab_static_cluster(sc_key);
 
-      g_assert(sc->dynamic == dynamic);
-    }
+  if (!sc)
```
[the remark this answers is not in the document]
presidento: Okay, I see, in the next commit it can be `NULL`.

lib/cfg-grammar.y
lbudai: -1 is cannot be set by users in the config, it is the _default_ value, so when the option is not set then number of dynamic counters are unlimited as before.

presidento:
Review answers:
> What do you think about storing dynamic counters in a separate hash table?

Looks good to me. We could use a counter, as well, since it can only increase one-by-one, so it would not be so complicated to maintain it. But it is also fine for me.

> What do you think about the reload-issue? (when the user sets a lower value, after a reload the already registered dynamic counters won't be removed until stats-freq timer not expires)?

Also looks good to me. It can be implemented to clean up some elements of the hash table, but in that case it would be good to keep the older stats, or use a random algorithm... As I understand, when stats-freq expires, everything starts from scratch, so now it is easier to explain and maintain. As I see.

lbudai:
@fekete-robert :  some documentation related notes
 * @presidento pointed out that the default value (-1) cannot be set by users, so if someone don't need this feature(->don't want to limit the number of dynamic counters), he or she has to remove the option from the config (and this is intentional from my part)
 * reload won't clear dynamic counters (hin: lifetime, stats-freq)

lib/cfg-parser.c
```diff
@@ -86,6 +86,7 @@ static CfgLexerKeyword main_keywords[] =
   { "stats_lifetime",     KW_STATS_LIFETIME },
   { "stats_level",        KW_STATS_LEVEL },
   { "stats",              KW_STATS_FREQ, KWS_OBSOLETE, "stats_freq" },
+  { "stats_max_dynamics", KW_STATS_MAX_DYNAMIC },
```
mitzkia: Maybe it will be "punctual", sorry about that. 
I think "stats_max_dynamics" should be only "stats_max_dynamic". I know it is only just a "name" here.

lib/stats/stats-cluster.c
```diff
@@ -264,6 +264,15 @@ stats_cluster_new(const StatsClusterKey *key)
   return self;
 }
 
+StatsCluster *
+stats_cluster_dynamic_new(const StatsClusterKey *key)
```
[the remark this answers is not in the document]
lbudai: We talked about this IRL, here is my short answer: I think dynamic cluster represents a different (sub)type (it behaves differently than the non-dynamic clusters) and we already provide a different API for dynamic counters (note: there is no dynamic counter, but dynamic cluster, what makes a counter dynamic is that the counter is registered with one of the `_dynamic` function which track the counter inside a dynamic cluster).
So I think this is not only a simple attribute: this is a kind of type selector and we actually make behavioral decisions based on the value of this attribute.

fekete-robert:
Thanks, I'll add it to our todo list.

PR #1752, opened 2017-11-09
Emit less confusing warnings for diskq

### LINKED ISSUE #1733: Disk buffer options check confusing
I'm running 3.12.1. As soon as I add `mem-buf-size(8192)`  to a non-reliable disk buffer, syslog-ng yields at startup:
```
[2017-10-22T18:00:01.156512] WARNING: Non-reliable queue: the mem-buf-size parameter is omitted;
```

Same thing happens as soon as I add `mem-buf-length(1000)` to a reliable disk buffer:
```
[2017-10-22T18:00:01.156559] WARNING: Reliable queue: the mem-buf-length parameter is omitted;
```

If the warning messages are what the code is really supposed to test, then conditions in `disk_queue_options_check_plugin_settings()` are reversed:
```
void
disk_queue_options_check_plugin_settings(DiskQueueOptions *self)
{
  if (self->reliable)
    {
      if (self->mem_buf_length > 0)
        {
          msg_warning("WARNING: Reliable queue: the mem-buf-length parameter is omitted");
        }
    }
  else
    {
      if (self->mem_buf_size > 0)
        {
          msg_warning("WARNING: Non-reliable queue: the mem-buf-size parameter is omitted");
        }
    }
}
```
They are initalized to `-1` in `disk_queue_options_set_default_options()`, so shouldn't those be testing for less than zero?

And BTW, [OSE documentation](https://www.balabit.com/documents/syslog-ng-ose-latest-guides/en/syslog-ng-ose-guide-admin/html/reference-destination-syslog-chapter.html) adds to the confusion:
* doc[1] says `mem-buf-length()` should be used "if the option `reliable()` is set to no"
* doc[1] says `mem-buf-size()` should be used "if the option `reliable()` is set to yes"

I believe this is contrary to what the plugin code does, because:
* code emits warning if "mem-buf-length parameter is omitted" for reliable=yes
* code emits warning if "mem-buf-size parameter is omitted" for reliable=no

So maybe it's the warning messages that are wrong, and they should say "NOT omitted"?

My syslog-ng version info:
```
syslog-ng 3 (3.12.1)
Installer-Version: 3.12.1
Revision: 
Module-Directory: /usr/lib64/syslog-ng
Module-Path: /usr/lib64/syslog-ng
Available-Modules: affile,graphite,map-value-pairs,afsql,http,kvformat,snmptrapd-parser,cryptofuncs,pseudofile,linux-kmsg-format,dbparser,stardate,tags-parser,basicfuncs,add-contextual-data,afprog,xml,afsocket,system-source,afstomp,confgen,csvparser,tfgetent,afmongodb,cef,disk-buffer,afuser,syslogformat,mod-python,date
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: off
Enable-TCP-Wrapper: off
Enable-Linux-Caps: on
Enable-Systemd: off
```

shastah:
Using the word "omitted" may be confusing, as it sounds like something should have been there, but was not taken into account. Let's use "ignored" instead, as for something that *should not* be there in the first place (and therefore was ignored).
Also be more verbose and suggest the other configuration parameter.

Fixes #1733.

### DIFF
--- a/modules/diskq/diskq-options.c
+++ b/modules/diskq/diskq-options.c
@@ -79,14 +79,14 @@ disk_queue_options_check_plugin_settings(DiskQueueOptions *self)
     {
       if (self->mem_buf_length > 0)
         {
-          msg_warning("WARNING: Reliable queue: the mem-buf-length parameter is omitted");
+          msg_warning("WARNING: mem-buf-length parameter was ignored as it is not compatible with reliable queue. Did you mean mem-buf-size?");
         }
     }
   else
     {
       if (self->mem_buf_size > 0)
         {
-          msg_warning("WARNING: Non-reliable queue: the mem-buf-size parameter is omitted");
+          msg_warning("WARNING: mem-buf-size parameter was ignored as it is not compatible with non-reliable queue. Did you mean mem-buf-length?");
         }
     }
 }


kira-syslogng:

Kokan:
@kira-syslogng ok to test

kira-syslogng:

PR #1753, opened 2017-11-10
criterion: disabling aslr for testcases

furiel:
Parametrized tests in criterion crashes with aslr enabled.
https://github.com/Snaipe/Criterion/issues/208

This can be a problem, because for example in ubuntu 17.04 gcc
produces position  independent executable code by default. With
randomize_va_space > 0, it leads criterion tests to crash.

This patch will compile the tests into a non position independent
binary, avoiding the crash.
Non test related binaries are untouched.

Test:
$ file modules/xml/tests/test_xml_parser
modules/xml/tests/test_xml_parser: ELF 64-bit LSB executable [...]

However for syslog-ng binary:
$ file ../root/sbin/syslog-ng
../root/sbin/syslog-ng: ELF 64-bit LSB shared object

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -60,7 +60,7 @@ TEST_CFLAGS		+= $(CRITERION_CFLAGS)
 TEST_LDADD		+= $(CRITERION_LIBS)
 endif
 
-test_ldflags		= -no-install
+test_ldflags		= -no-install $(NO_PIE_LDFLAG)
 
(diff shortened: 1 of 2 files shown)


furiel:
I needed to remove the quotes in LDFLAGS+=${test_ldflags}. The reason is I got compiler errors: "-no-install -no-pie" is passed as a single argument to the linker, resulting in error.

kira-syslogng:

furiel:
Adding detection for -no-pie, as it seems we cannot expect it to be available on every system.

kira-syslogng:

Issue #716, opened 2015-09-25
sparse syslog-ng.persist file can cause SIGBUS


juhaszviktor:
When starting syslog-ng on disk which has minimal free disk spaces (<16KB),
the persist file will be created with the required size (seek and write will be successful),
but the memcpy on the mmapped structure will be failed with SIGBUS.

PR #717, opened 2015-09-25
F/persist state fixes

juhaszviktor:
fixes #716

PR #1785, opened 2017-12-01
Crash with SIGBUS when persist_file cannot grow

furiel:
Continuing https://github.com/balabit/syslog-ng/pull/717 in a new PR due that I do not have write access to the original PR. Credits for @juhaszviktor for the investigation and putting together the original pr!
Fixes: #716 

As I see the only thing left to do was to inverse the dependency between persist_state and mainloop. This is fixed here.

However, I ran into another problem while testing: 
Test: created a small tmpfs mountpoint for the persist state. Setting wildcard-file source and creating a lot of files with very long names, waiting for crash.

It turned out syslog-ng crashed for another reason:
```
ERROR:../syslog-ng/lib/persist-state.c:295:_alloc_value: assertion failed: (_check_free_space(self, size))
[1]    22773 abort (core dumped)  ./syslog-ng -Fe -R tmpfs/persist.file
```
The idea was that syslog-ng preallocates a fixed sized space that we ensure it is mapped. When syslog-ng sees that the free space is getting low, (less than 4K), then tries to expand the persist file. But still free space is available in the preallocated space, so we can still return with a **valid** handler. But the problem is syslog-ng is not stopped instantly when calling main_loop_exit() (it is possible that halt can be deferred infinitely), so there can be more persist state allocation requests until syslog-ng stops, filling up the preallocated space, until all space is exhausted, crashing in the assert.

So instead, if there is no free space, I try to return with zero handler, as the original implementation did, which would mean the allocation failed for the caller modules. But until the reserved space exhausted we can still return a return **valid** handler.

Though this led to another kind of crash: due that the caller sites were not ready for such allocation failures. In my specific case:
```
ERROR:../syslog-ng/lib/logproto/logproto-buffered-server.c:47:log_proto_buffered_server_get_state: assertion failed: (self->persist_handle != 0)
```

So a little correction is needed on the caller sites. Please check the last patch for reference. For the run-id and host-id modules: I could not trigger the crash, so I just ended up adding a g_assert for handle. However if you feel the necessity, we can add an in-memory storage for these modules too, something similar that happens with self->persist_state vs self->state1 in LogProtoBufferedServer.

An alternative for the situation could have been that if persist_state could not grow, she could allocate an in-memory-only buffer and return a handler with that. My problem was with it is that it would mean we could not ensure persist_state is persistent any more, which would be strange. Also memory allocation could fail as well. That's why I opted the solution explained above.

### DIFF
--- a/lib/host-id.c
+++ b/lib/host-id.c
@@ -24,6 +24,7 @@
 
 #include "host-id.h"
 #include "str-format.h"
+#include "messages.h"
 #include <openssl/rand.h>
 
 guint32 global_host_id = 0;
@@ -42,7 +43,7 @@ _create_host_id(void)
   return host_id.id;
 }
 
-void
+gboolean
 host_id_init(PersistState *state)
 {
   gsize size;
@@ -59,6 +60,12 @@ host_id_init(PersistState *state)
       handle = persist_state_alloc_entry(state, HOST_ID_PERSIST_KEY, sizeof(HostIdState));
     }
 
+  if (!handle)
+    {
+      msg_error("host-id: could not allocate persist state");
+      return FALSE;
+    }
+
   host_id_state = persist_state_map_entry(state, handle);
   {
     if (new_host_id_required)
@@ -72,6 +79,8 @@ host_id_init(PersistState *state)
       }
   }
   persist_state_unmap_entry(state, handle);
+
+  return TRUE;
 }
 
 guint32
--- a/lib/logproto/logproto-buffered-server.c
+++ b/lib/logproto/logproto-buffered-server.c
@@ -527,7 +527,10 @@ error:
       if (new_state_handle)
         log_proto_buffered_server_apply_state(self, new_state_handle, persist_name);
       else
-        self->state1 = new_state;
+        {
+          self->persist_state = NULL;
+          self->state1 = new_state;
+        }
     }
   if (new_state_handle)
     {
--- a/lib/persist-state.c
+++ b/lib/persist-state.c
@@ -26,6 +26,8 @@
 #include "serialize.h"
 #include "messages.h"
 #include "fdhelpers.h"
+#include "misc.h"
+#include "mainloop.h"
 
 #include <sys/types.h>
 #include <unistd.h>
@@ -33,6 +35,7 @@
 #include <sys/mman.h>
 #include <errno.h>
 #include <string.h>
+#include <signal.h>
 
 typedef struct _PersistFileHeader
 {
@@ -59,6 +62,7 @@ typedef struct _PersistFileHeader
 
 #define PERSIST_FILE_INITIAL_SIZE 16384
 #define PERSIST_STATE_KEY_BLOCK_SIZE 4096
+#define PERSIST_FILE_WATERMARK 4096
 
 /*
  * The syslog-ng persistent state is a set of name-value pairs,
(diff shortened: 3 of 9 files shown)


kira-syslogng:

lib/host-id.c
```diff
@@ -59,6 +59,7 @@ host_id_init(PersistState *state)
       handle = persist_state_alloc_entry(state, HOST_ID_PERSIST_KEY, sizeof(HostIdState));
     }
 
+  g_assert(handle);
```
furiel: Initialization should fail here.

lib/persist-state.c
```diff
@@ -266,6 +304,14 @@ _alloc_value(PersistState *self, guint32 orig_size, gboolean in_use, guint8 vers
   persist_state_unmap_entry(self, self->current_ofs);
 
   self->current_ofs += size + sizeof(PersistValueHeader);
+
+  if (!_check_watermark(self) && !_grow_store(self, self->current_size + PERSIST_FILE_INITIAL_SIZE))
+    {
+      msg_error("Can't preallocate space for persist file", NULL);
```
furiel: We should write more information, for example size information

lib/run-id.c
```diff
@@ -52,6 +52,7 @@ run_id_init(PersistState *state)
       handle = persist_state_alloc_entry(state, RUN_ID_PERSIST_KEY, sizeof(RunIDState) );
     }
 
+  g_assert(handle);
```
furiel: Initialization should fail here

lib/logproto/logproto-buffered-server.c
```diff
@@ -42,11 +42,16 @@ typedef struct _BufferedServerBookmarkData
 LogProtoBufferedServerState *
 log_proto_buffered_server_get_state(LogProtoBufferedServer *self)
 {
-  if (self->persist_state)
+  if (self->persist_state && self->persist_handle)
     {
-      g_assert(self->persist_handle != 0);
+      if (G_UNLIKELY(self->persist_handle == 0))
```
furiel: self->state1 is intended for logproto-buffered cases when there is no need to store state in persistent storage, in memory state is enough: for example tcp buffering.
Instead of this, when opening a file in the wildcard file source case, we should return with error.

kira-syslogng:
Build SUCCESS, test branch: master
 406 tests run, 0 skipped, 0 failed.

kira-syslogng:

furiel:

kira-syslogng:

lib/persist-state.c
```diff
@@ -237,6 +263,18 @@ _commit_store(PersistState *self)
...
+static gboolean
+_check_watermark(PersistState *self)
+{
+  return (self->current_ofs + PERSIST_FILE_WATERMARK < self->current_size);
+}
+
+static inline gboolean
+_check_free_space(PersistState *self, guint32 size)
+{
+  return (size + sizeof(PersistValueHeader) +  self->current_ofs) <= self->current_size;
```
Kokan: Shouldn't it only be less and not less equal ?

lib/persist-state.c
```diff
@@ -266,6 +304,16 @@ _alloc_value(PersistState *self, guint32 orig_size, gboolean in_use, guint8 vers
   persist_state_unmap_entry(self, self->current_ofs);
 
   self->current_ofs += size + sizeof(PersistValueHeader);
+
+  if (!_check_watermark(self) && !_grow_store(self, self->current_size + PERSIST_FILE_INITIAL_SIZE))
+    {
+      msg_error("Can't preallocate space for persist file",
+                evt_tag_int("current", self->current_size),
+                evt_tag_int("new_size", self->current_size + PERSIST_FILE_INITIAL_SIZE));
+      if (self->error_handler.handler)
```
Kokan: Maybe it would be nice to have an inline wrapper for this like: `persist_state_error_handler_run(self)`

lib/host-id.c
```diff
@@ -59,6 +60,12 @@ host_id_init(PersistState *state)
       handle = persist_state_alloc_entry(state, HOST_ID_PERSIST_KEY, sizeof(HostIdState));
     }
 
+  if (!handle)
+    {
+      msg_error("host-id: could not allocate persist state", NULL);
```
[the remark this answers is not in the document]
furiel: Thanks I did not realise the if (tag1) in msg_event_create. Removed the null here and in run-id.c as well.

lib/persist-state.c
furiel: Thanks, indeed I pulled out to a function. Though I did not use inline, let's leave to the compiler to decide that.

lib/persist-state.c
furiel: I believe less or equal can be allowed here. Equal means exact fit, meaning we can add that last object without expansion.

Less is used when checking the free space with watermark.

kira-syslogng:

kira-syslogng:

PR #1788, opened 2017-12-01
App parser kill switch

bazsi:
This patch implements auto-parse(yes/no) option into

- the app-paser()
- and through the use of __VARARGS__ system() and default-network-drivers()

this was requested by @czanik

### DIFF
--- a/lib/cfg-args.c
+++ b/lib/cfg-args.c
@@ -24,6 +24,7 @@
 #include "cfg-args.h"
 #include "messages.h"
 #include "str-utils.h"
+#include "str-repr/encode.h"
 
 struct _CfgArgs
 {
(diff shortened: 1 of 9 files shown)


kira-syslogng:

bazsi:
With these patches it is not possible to control which applications are processed by auto-parser:

```
# disable completely
system(auto-parse(no));

# all parsers enabled, except a few
system(auto-parse-exclude("foo,bar,baz"));

# only a few specific are enabled
system(auto-parse-include("foo,bar,baz"));
```

The same options should work with the default-network-drivers() stuff as well.

bazsi:
@czanik feedback and testing would be welcome, as always.

kira-syslogng:

czanik:
I have tried:
```
system(auto-parse-exclude("iptables"));
```
And iptables logs were no more parsed, but sudo logs were still parsed.

czanik:
```
system(auto-parse-include("iptables"));
```
also worked fine.

czanik:
```
system(auto-parse(no));
```
also worked fine.

I also did a very basic performance test with the above setting: no measurable difference with the system source. My suspicion is, that the journal() source is the bottleneck here.

lbudai:
@czanik : thx for feedbacks.

PR #1791, opened 2017-12-04
plugin: decrease the log level of module loading errors

MrAnno:


kira-syslogng:

lbudai:
@pzoleex : could you share the details on why did you request this change?

pzoleex:
@lbudai Because it gives a confusing error message about not existing libjvm.so, when syslog-ng service is started however no java in the configuration.

It's not lucky thing to give a big error message after a brand new installation with a default configuration about something that is not used at all :)

Starting syslog-ng: Error opening plugin module; module='mod-java', error='libjvm.so: cannot open shared object file: No such file or directory'

lbudai:
@pzoleex : thanks!

PR #1342, opened 2017-02-10
fix silent drop of messages in http bulk mode

faxm0dem:
elastic-v2 with `flush-limit(N)` and `N>1` *e.g.* bulk mode would not notice when bulk failed *e.g.* for 40x status codes (for instance 403 Unauthorized).

Issue #1149, opened 2016-08-03
elasticsearch2 HTTP mode woes


faxm0dem:
I just tested the HTTP mode of the new elastic v2 destination.
While it seems to work (documents are indexed), I have the following two issues:
- On startup syslog-ng complains about the destination's logger:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```
- When sending a SIGINT syslog-ng fails to end gracefully:

```
*** Error in `syslog-ng': corrupted double-linked list: 0x0000000000a476e0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x7bda5)[0x7fa66ebfeda5]
/lib64/libc.so.6(+0x7ded5)[0x7fa66ec00ed5]
/lib64/libc.so.6(__libc_malloc+0x4c)[0x7fa66ec0326c]
/usr/lib/jvm/java-1.7.0-openjdk-1.7.0.101-2.6.6.1.el7_2.x86_64/jre/lib/amd64/server/libjvm.so(+0x7ecc2a)[0x7fa66b2c7c2a]
[...]
```

lbudai:
SLF4J is used by the Jest client. I'll check if we can fix it or it is an issue at Jest side (and if it is and if they have a fix in a newer version then we will use that).

The other is much more interesting, I'll check it.

lbudai:
@faxm0dem : I think I know the reason and also have a solution for this problem.
At least in my environment it worked :-) 
download [slfj-1.7.13](http://www.slf4j.org/dist/slf4j-1.7.14.tar.gz) and copy `slf4j-simple-1.7.13.jar` next to `slf4j-api-1.7.13.jar`.

Of course I will create a PR for addressing this issue, but until that it could work.
[ I'll ask 15 minutes from my wife ;-) ]

faxm0dem:
does this solve the SIGINT issue too?

lbudai:
@faxm0dem : yeah, the lack of this library was the reason of the segfault

faxm0dem:
awesome, thanks!

lbudai:
@faxm0dem: does this patch solve your problem?

faxm0dem:
I'm sorry I didn't have time to test yet

faxm0dem:
[...]

PR #1800, opened 2017-12-06
Elastic 2 fix bulk silent drop

lbudai:
@faxm0dem : could you check this?
#1342 is conflicting, so I'm open this new PR
(if you tested my changes, you can rebase your branch against master and then cherry pick my commit ;) )

### DIFF
--- a/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
+++ b/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
@@ -128,11 +128,7 @@ public class ESHttpClient implements ESClient {
 
 	@Override
 	public void close() {
-		try {
-			messageProcessor.flush();
-		} catch (IOException e) {
-			logger.error(e.getMessage());
-		}
+		messageProcessor.flush();
 	}
 
 	@Override
--- a/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
+++ b/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
@@ -128,7 +128,11 @@ public class ESHttpClient implements ESClient {
 
 	@Override
 	public void close() {
-		messageProcessor.flush();
+		try {
+			messageProcessor.flush();
+		} catch (IOException e) {
+			logger.error(e.getMessage());
+		}
 	}
 
 	@Override
--- a/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
+++ b/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
@@ -52,6 +52,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.io.FileInputStream;
+import org.apache.log4j.PropertyConfigurator;
 
 import java.io.IOException;
 
@@ -61,6 +62,11 @@ public class ESHttpClient implements ESClient {
 	private JestClient client;
 	private HttpMessageProcessor messageProcessor;
 	protected Logger logger;
+
+  static  {
+      PropertyConfigurator.configure(System.getProperty("log4j.configuration"));
+  }
+
 
   public class HttpClientBuilderException extends RuntimeException {
     public HttpClientBuilderException() {}
--- a/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/messageprocessor/http/HttpBulkMessageProcessor.java
+++ b/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/messageprocessor/http/HttpBulkMessageProcessor.java
@@ -40,11 +40,10 @@ public class HttpBulkMessageProcessor extends  HttpMessageProcessor {
 
 	public HttpBulkMessageProcessor(ElasticSearchOptions options, ESHttpClient client) {
 		super(options, client);
+		bulk = new Bulk.Builder();
 	}
 	@Override
 	public void init() {
-		bulk = new Bulk.Builder();
-		messageCounter = 0;
 		flushLimit = options.getFlushLimit();
 	}
 
--- a/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/messageprocessor/http/HttpBulkMessageProcessor.java
+++ b/modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/messageprocessor/http/HttpBulkMessageProcessor.java
@@ -49,9 +49,9 @@ public class HttpBulkMessageProcessor extends  HttpMessageProcessor {
 	}
 
 	@Override
-	public void flush() throws IOException {
+	public boolean flush() {
 		JestResult jestResult = null;
-    logger.debug("Flushing messages for ES destination [mode=" + options.getClientMode() + "]");
+		logger.debug("Flushing messages for ES destination [mode=" + options.getClientMode() + "]");
 		Bulk bulkActions = bulk.build();
 		try {
 			jestResult = client.getClient().execute(bulkActions);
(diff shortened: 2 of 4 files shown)


kira-syslogng:

faxm0dem:
it works, but the error message in `internal()` source isn't very helpful. Not sure where the null pointer exception comes from...

```
Dec  7 15:28:49 syslogng syslog-ng[12591]: Connect to ccosvm0968.in2p3.fr:9200 [ccosvm0968.in2p3.fr/172.17.0.208] failed: connect timed out;
Dec  7 15:28:49 syslogng syslog-ng[12591]: Exception occured: java.lang.NullPointerException\x0a     at org.syslog_ng.elasticsearch_v2.messageprocessor.http.HttpBulkMessageProcessor.flush(HttpBulkMessageProcessor.java:63)\x0a     at org.syslog_ng.elasticsearch_v2.messageprocessor.http.HttpBulkMessageProcessor.send(HttpBulkMessageProcessor.java:76)\x0a     at org.syslog_ng.elasticsearch_v2.messageprocessor.http.HttpMessageProcessor.send(HttpMessageProcessor.java:62)\x0a      at org.syslog_ng.elasticsearch_v2.client.http.ESHttpClient.send(ESHttpClient.java:152)\x0a      at org.syslog_ng.elasticsearch_v2.ElasticSearchDestination.send(ElasticSearchDestination.java:95)\x0a    at org.syslog_ng.StructuredLogDestination.sendProxy(StructuredLogDestination.java:35)\x0a;
```

lbudai:
@faxm0dem : well... we can change the error message, but this is come from the jest library.
(and I didn't touch this part of the code :) )

faxm0dem:
I also have quite often the following message:

```
Dec  8 08:30:28 syslogng syslog-ng[12591]: Exception occured: java.lang.NullPointerException\x0a;
Dec  8 08:30:28 syslogng syslog-ng[12591]: Read timed out;
```

lbudai:
When we got an `IOException`, we catch it and `jestResult` variable will be `null`...

lbudai:
PR is updated with this change (->return when we got `IOException`).

kira-syslogng:

faxm0dem:
Unfortunately with this change, messages are silently dropped again.
Back to square one 👎

lbudai:
@faxm0dem : the logic is the same as in your PR... so you should get error messages when 
 * `IOException` occurs, or
 *  client.execute fails.

So your messages are still dropped while you don't get the error messages? This means that your messages are lost in somewhere else... (I mean there is no `IOException` and client.execute does not report any failures).

faxm0dem:
Yes, messages are silently dropped:

* No error message
* Processed stat goes up
* Queued stat stays zero

I will investigate this further next week - but not before Thursday.

faxm0dem:
I tried to dig into this deeper, but I realized the log events from the java program never make it to the internal() or system() or even STDERR.
I can see the messages from jest at startup:

```
[main] INFO io.searchbox.client.JestClientFactory - Idle connection reaping disabled...
```

but nothing thereafter

faxm0dem:
I tried a few things, to no avail: total silence from syslog-ng (on STDOUT/STDERR, internal() and system():

* `jvm-options('-Dlog4j.debug -Dlog4j.configuration=/etc/syslog-ng/log4j.properties')
* `slf4j group: 'org.slf4j', name: 'slf4j-log4j12', version: '1.7.13'` in modules/java-modules/elastic-v2/build.gradle
* setting log4j12 to log to a file using /etc/syslog-ng/log4j.properties

The logfile is never created, and nothing is on STDOUT of syslog-ng, except the usual JEST messages

faxm0dem:
could this somehow be connected to #1149?

lbudai:
@faxm0dem : I don't think so.

lbudai:
@faxm0dem : I'll only have time for working on this on Friday.

faxm0dem:
@lbudai okay, thank you!

lbudai:
@faxmodem: let me summarize the issue

1. in case of bulk insert (and also in case of non-bulk insert...) when an error occurs, syslog-ng does not report any error
2. in case of bulk insert when it fails, syslog-ng just drops the messages

Your patch focuses on 1st point.
2nd point is unfortunately cannot be resolved until syslog-ng does not support late ack mechanism in java site: currently when the messages are passed to the java destination, syslog-ng increase the processed counter and release the message. We don't keep a messages on the queue that are
passed to the java destination (this is the case for Python destinations), we just send an early ACK, so it is not possible to ACK messages in asynchronous way.

faxm0dem:
I understand in case of an error, we lose the batch of messages (as defined in `flush-limit`). But we should at least avoid losing the subsequent batches!

lbudai:
Agree, and this is why we want to return with `false` from `flush()` when exception occurs. The problem is that it seems that we don't get the exception.

faxm0dem:
can you reproduce the behaviour that whatever you log, even in normal flow, it's dropped even on STDIO when running syslog-ng in the FG?

lbudai:
I think I found the root cause.

Each time when the `flush()` (which is called by `send()`) returns with `false`, we increment a counter.
When the value of this counter reach a limit (max retries), syslog-ng drops the messages.
Unfortunately this mechanism is not working for elastic http bulk because the counter never reach the limit.
The reason of this is that we call `init()` in `ESHttpClient.open()` after the java dst is suspended, which is happen when an error is reported by send, and the `init` creates a new Bulk.Builder and reset messagecounter.
The next send will be succeeded as the messagecounter has been reseted (`send()` returning true until the batch size has not been reached, then it calls `flush` which fails again, java dst suspended again, then open is called which calls init, and it will happen again and again, until all the messages are dropped silently). 

After I fix it, there will be another issue: drop counter is incremented by 1, not by batch size.

faxm0dem:
good catch! what about the logging?

lbudai:
JEST logs the connection failures in my environment. (I'm just using iptables rules for producing connection errors.)

faxm0dem:
interesting. can you clarify what logging libraries are used?
IIRC it's SLF4J using [slf4j-simple] which is the [simple](https://www.slf4j.org/apidocs/org/slf4j/impl/SimpleLogger.html) logger and only logs to the console.

If that's correct, it will only be visible when running syslog-ng in the fg?

kira-syslogng:

lbudai:
@faxm0dem : yep, it is visible only on console.
But I can see 'failed to connect' logs and 'Multiple failures while sending' in the `internal()` logs after my latest changes.

lbudai:
When I added this to `jvm_options` :  `-Dorg.slf4j.simpleLogger.logFile=/tmp/jest-slf4j.log` , it put some logs to the given file, but only some initialization related lines... and other still logged to the console.

lbudai:
@faxm0dem : finally I was able to use log4j  (and configure a fileappender).
JEST has a logger and JEST uses the org.apache.commons.http which actually logs those lines that we are really interesed in (->connection errors). The problem is that it's hard to load the `log4j.properties`.
Currently I'm loading it programmatically with the help of PropertyContainer (and trust me, I tried several ways, but did not work and I wanted to see whether it is working or not, so I choose a temporary solution)
.
In syslog-ng config I added these:

```
jvm_options("-Dorg.apache.commons.logging.Log=org.apache.commons.logging.impl.Log4JLogger -Dorg.apache.commons.logging.LogFactory=org.apache.commons.logging.impl.LogFactoryImpl -Dlog4j.debug=true");
```

kira-syslogng:

faxm0dem:
I just tested your recent changes, and it works as expected as far as the queue is concerned, thank you!
However, I'm still unable to see output from Jest even after using the `jvm_options` in my syslog-ng.conf.

lbudai:
@faxmodem: with my last patch (and notes) I was able to get log messages from JEST :
```
options {
 jvm_options("-Dorg.apache.commons.logging.Log=org.apache.commons.logging.impl.Log4JLogger
-Dorg.apache.commons.logging.LogFactory=org.apache.commons.logging.impl.LogFactoryImpl
-Dlog4j.debug=true -Dlog4j.configuration=/tmp/log4j.properties");
};
```

and log4j.properties:
```
log4j.logger.org.elasticsearch=WARN
log4j.logger.org.apache.http=DEBUG, slng, fout
log4j.logger.org.apache.http.wire=DEBUG, slng, fout
log4j.logger.org.apache.http.impl.conn=DEBUG, slng, fout
log4j.logger.org.apache.http.impl.client=DEBUG, slng, fout
log4j.logger.org.apache.http.client=DEBUG, slng, fout

log4j.appender.slng=org.syslog_ng.logging.SyslogNgInternalLogger

log4j.appender.fout=org.apache.log4j.FileAppender
log4j.appender.fout.File=/tmp/log4jjjj.log
log4j.appender.fout.layout=org.apache.log4j.PatternLayout
log4j.appender.fout.layout.ConversionPattern=%d{HH:mm:ss} %-5p %c{1}:%L
- %m%n
```

( DEBUG messages maybe won't be handled by `internal` as they are printed to stderr... )

PR #1811, opened 2017-12-12
cmake: java and afamqp

Kokan:
Contains:
* Fix for the `json-plugin` link against the correct shared object
* Fix for the java module, the shared object was not installed for (`libmod-java.so`)
* `CMakeLists.txt` for `afamqp` and `rabbitmq-c` lib


kira-syslogng:

kira-syslogng:

cmake/Modules/FindRabbitMQ.cmake
```diff
@@ -0,0 +1,39 @@
...
+include(LibFindMacros)
+
+# Find the include directories
+find_path(RabbitMQ_INCLUDE_DIR NAMES amqp.h)
+
+find_library(RabbitMQ_LIBRARY NAMES rabbitmq)
+
+libfind_process(RabbitMQ)
+
+#find_file(_Rabbitmqc_SSL_HEADER NAMES amqp_ssl_socket.h PATHS ${Rabbitmqc_INCLUDE_DIR} NO_DEFAULT_PATH)
```
furiel: Please delete the commented lines.

modules/afamqp/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+generate_y_from_ym(modules/afamqp/afamqp-grammar)
+
+bison_target(AMQPGrammar
+    ${CMAKE_CURRENT_BINARY_DIR}/afamqp-grammar.y
+    ${CMAKE_CURRENT_BINARY_DIR}/afamqp-grammar.c
+    COMPILE_FLAGS ${BISON_FLAGS})
+
+include_directories (${CMAKE_CURRENT_BINARY_DIR})
+include_directories (${CMAKE_CURRENT_SOURCE_DIR})
```
furiel: When using external rabbitmq, I miss the header include directory for it. Why can we avoid setting it?

Kokan: We shall not. Fixed.

The reason why it worked in CI:

**RabbitMQ_SOURCE=system**
The compiler (for now clang) default include directories are:
```
#include "..." search starts here:
#include <...> search starts here:
 /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0
 /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/x86_64-linux-gnu/c++/6.3.0
 /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/backward
 /usr/include/clang/4.0.0/include
 /usr/local/include
 /usr/include/x86_64-linux-gnu
 /usr/include
```
The installed headers of `librabbitmq` are in `/usr/include`.

**RabbitMQ_SOURCE=internal**
In this case our cmake installs it into the `${CMAKE_CURRENT_BINARY_DIR}`, which is included also, thus not causing compile error.

cmake/Modules/FindRabbitMQ.cmake
Kokan: Fixed.

kira-syslogng:

cmake/external_or_find_package.cmake
```diff
@@ -37,13 +37,16 @@ function(external_or_find_package LIB_NAME)
         set(${LIB_NAME}_INTERNAL FALSE)
     endif()
 
+    if (${LIB_NAME}_INTERNAL)
```
furiel: exlude from all is needed only for external linking. I miss a "NOT" here.

furiel: My bad... It should be excluded from all, because we do not want to pull this dependency by default. Only if other target pulls in on demand.

PR #1812, opened 2017-12-13
filter: fix memory leak with multiple call on FilterCall's init

ibmibmibm:
if `filter_call_init` being called twice or more, `filter_expr_ref(self->filter_expr)` will trigger each time, but `filter_call_free` -> `filter_expr_unref(self->filter_expr)` will only trigger once, so `self->filter_expr` will leak.

This commit prevent furthermore initialization when `self->filter_expr` has value.

Signed-off-by: Shen-Ta Hsieh <ibmibmibm.tw@gmail.com>

### DIFF
--- a/lib/filter/filter-call.c
+++ b/lib/filter/filter-call.c
@@ -58,6 +58,10 @@ filter_call_init(FilterExprNode *s, GlobalConfig *cfg)
   FilterCall *self = (FilterCall *) s;
   LogExprNode *rule;
 
+  /* skip initialize if filter_call_init already called. */
+  if (self->filter_expr)
+    return;
+
   rule = cfg_tree_get_object(&cfg->tree, ENC_FILTER, self->rule);
   if (rule)
     {


kira-syslogng:

Kokan:
@kira-syslogng ok to test

kira-syslogng:

furiel:
Thanks for reporting and the patch!

The intended behavior is that each logpipes that share the same filter expression should ref the object, and unref during free. So free should be called for each free of such logpipes, not only once.

Can you share a minimal configuration and reproduction steps together with a valgrind entry that shows the leak?

ibmibmibm:
https://gist.github.com/ibmibmibm/166b6e578ede5eed72da6ef6e4c20fd1

kira-syslogng:

bazsi:
log_pipe_init() might be called more than once, if the initialization of the new config fails and we are reverting back to the old one.

However, this seems to always leak if I understand the valgrind logs posted. But it's true that if log_pipe_init() might be called multiple times, we need to drop the ref before overwriting the pointer and that doesn't happen either.

I don't think the filter_expr lookup would return different values at different init invocation, so the solution to check for a NULL pointer would be just as fine.

So, this looks good to me.

PR #1813, opened 2017-12-13
Mainloop refactor

ibmibmibm:
1. Prevent undefined behavior - [Shift-pass-bitwidth](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)
2. Holding lock on condition variable signaling. See [glibc document](https://developer.gnome.org/glib/stable/glib-Threads.html#g-cond-signal).

### DIFF
--- a/lib/mainloop-worker.c
+++ b/lib/mainloop-worker.c
@@ -194,9 +194,8 @@ main_loop_worker_thread_stop(void)
 
   g_mutex_lock(&workers_running_lock);
   main_loop_workers_running--;
-  g_mutex_unlock(&workers_running_lock);
-
   g_cond_signal(&thread_halt_cond);
+  g_mutex_unlock(&workers_running_lock);
 }
 
 void
--- a/lib/mainloop-worker.c
+++ b/lib/mainloop-worker.c
@@ -103,7 +103,8 @@ _release_thread_id(void)
   g_static_mutex_lock(&main_loop_workers_idmap_lock);
   if (main_loop_worker_id)
     {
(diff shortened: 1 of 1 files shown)


kira-syslogng:

Kokan:
@kira-syslogng ok to test

kira-syslogng:

kira-syslogng:

lib/mainloop-worker.c
```diff
@@ -103,7 +103,8 @@ _release_thread_id(void)
   g_static_mutex_lock(&main_loop_workers_idmap_lock);
   if (main_loop_worker_id)
     {
-      main_loop_workers_idmap[main_loop_worker_type] &= ~(1 << (main_loop_worker_id - 1));
+      const gint id = main_loop_worker_id & (MAIN_LOOP_MAX_WORKER_THREADS - 1);
```
szemere: This change will introduces a hidden dependency that MAIN_LOOP_MAX_WORKER_THREADS is the power of 2. I suggest to use the type of the main_loop_worker_idmap (guint64) explicitly hear. i.e.: sizeof(guint64)*8 -1

Note: This bitmask still allows (in either case) to have bigger number than MAIN_LOOP_MAX_WORKER_THREADS,  but we limit its value in: https://github.com/balabit/syslog-ng/blob/19ee6f82d11c0bd98f6cb9a26c6b60dee7a40d03/lib/mainloop-io-worker.c#L87

kira-syslogng:

PR #1815, opened 2017-12-13
afinter: avoid data race on next_mark_target

ibmibmibm:
writing next_mark_target.tv_sec is not guaranteed as an atomic operation.

### DIFF
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -421,7 +421,9 @@ afinter_postpone_mark(gint mark_freq)
     }
   else
     {
+      g_static_mutex_lock(&internal_mark_target_lock);
       next_mark_target.tv_sec = -1;
+      g_static_mutex_unlock(&internal_mark_target_lock);
     }
 }
 


kira-syslogng:

kira-syslogng:

lib/afinter.c
```diff
@@ -421,7 +421,9 @@ afinter_postpone_mark(gint mark_freq)
     }
   else
     {
+      g_static_mutex_lock(&internal_mark_target_lock);
       next_mark_target.tv_sec = -1;
+      g_static_mutex_unlock(&internal_mark_target_lock);
     }
```
bazsi: This one looks sane theoretically, however it would grab the lock in the fastpath for every message, which is a performance hurdle, and setting this to -1 is not really needed.

The conditional above the patch tests mark_freq, which is constant for the duration of a configuration. So the only thing this might have an affect on is when we are changing mark_freq in the configuration and reload it. Then for a single invocation, this would have an effect, otherwise it doesn't.

On the other hand, this lock is acquired in the other path, so it should be fine.

I am ok with this patch.

PR #1836, opened 2018-01-05
Crash when large variety of keys added to messages

furiel:
Syslog-ng can crash when large amount of new nvhandles are created, and the nvtable is read at the same time. When self->names (GArray) in NVRegistry is expanded, the data is reallocated. This means jobs that have reference to the original area - for example in the middle of reading from nvtable by name - can read from invalid memory area, hence the crash.

There is a lock in NVRegistry that is used when the GArray was written, however the lock was missing in the readers.

This patchset - instead of adding locks to the readers - adds an alternative array implementation to replace GArray, that does not free the data buffer during expansion immediately. The freeing is delayed to a time when no jobs are running. 

For that reason a new application hook is added with a small refactor on the application hooks.

There is a downside: application hook for no jobs are running might be deferred without bound on a busy system. Which means the original buffers might not be freed for a long time. Though I do not see it a serious problem, the memory we are talking about is the same size of the current nvregistry->names->data size (sum of the two powers below a two power is the two-power-1) at worst case.

The crash can be reproduced for example with kvparser or json parser.
Reproduction steps are collected here: https://gist.github.com/furiel/7383126ba738c55a4ed980c4546e693d

The fix-with-locking would have resulted a much simpler patchset. The reason I chose the lockless version over that is I tried both and made little performance test on my development machine. With the json parser configuration and testfile in the gist, running the same loggen command, I got values around 219k msg/s with the lockless version. With the locking-version I got around 190k msg/s. Which means the lockless version was 15% improvement to the locking-version.


kira-syslogng:

furiel:
While working on the lockless version, I was thinkink about an alternative of the the current array implementation that does not need memcpy at all, and there are no garbage chunks to be freed. Although I dropped the idea in the end, it might worth mentioning.

The structure is inspired by the two-level scheme of page tables in a 32bit architecture. The index space of the descriptor array in syslog-ng is 32bit. So the descriptors could be stored in 2^16 lists, each of them 2^16 long. (like a 2^16 x 2^16 matrix).

Here the 2^16 list of pointers needs to be preallocated at init time, so the idle memory consumption is 2^19 = 0.5MB, not a large number. Each time we need to append, the new descriptor can either fit in the current structure, or a new complete chunk will be allocated (chunk size is: 2^16 * sizeof(NVHandleDesc) = 2^16*32 = 2M), and the new descriptor will be inserted into the new chunk. No copy or deallocation is needed.

When fetching a descriptor, we need to derefer two times, in the first and in the second array, using 0xFFFF0000 and 0x0000FFFFF bits as indexes.

The code would be a little simpler: the array implementation would be more self contained, there would be no need for garbage collection. There is no copy when expanding as opposed to the PR version: currently the copy happens under a lock, which might hold back the job that does the expanding.
However such structure is optimized for the case when elements need to be added or deleted from the middle of the list. Such flexibility is paid by the two-time referencing during reading.
Syslog-ng does not need such flexibility, as syslog-ng only needs to append to the end, previous entries are not touched. As reading is more frequent than writing (especially expanding: 32times at maximum) in the syslng descriptor table, I felt it an unnecessary runtime complexity. So in the end I chose the garbage collecting version.

lib/apphook.c
```diff
@@ -79,14 +85,22 @@ register_application_hook(gint type, ApplicationHookFunc func, gpointer user_dat
...
+run_application_hook(gint type)
 {
   GList *l, *l_next;
+  gboolean remove_hook = !(type & AHFLAGS_PERSISTENT);
 
   if (remove_hook)
     {
-      g_assert(current_state <= type);
-      current_state = type;
+      if (!(type & AHFLAGS_INDEPENDENT))
```
Kokan: This could be one condition, something like that:
```c
static gboolean
app_hook_should_remove(gint type)
{
    return (!(type & (AHFLAGS_PERSISTENT | AHFLAGS_INDEPENDENT)));
}
...
if (app_hook_should_remove(type))
{
  app_hook_update_state(type);
}
```

lib/logmsg/tests/test_nvhandle_desc_array.c
```diff
@@ -0,0 +1,48 @@
...
+  NVHandleDesc desc1 = { .name = g_strdup("test1"), .flags = 1, .name_len = strlen("test1")};
+  NVHandleDesc desc2 = { .name = g_strdup("test2"), .flags = 2, .name_len = strlen("test2")};
+  NVHandleDesc desc3 = { .name = g_strdup("test3"), .flags = 3, .name_len = strlen("test3")};
+
+  NVHandleDescArray *array = nvhandle_desc_array_new();
+  nvhandle_desc_array_append(array, &desc0);
+  nvhandle_desc_array_append(array, &desc1);
+  nvhandle_desc_array_append(array, &desc2);
+  nvhandle_desc_array_append(array, &desc3);
+
```
Kokan: An assert would be nice for the `array->len`.

lib/logmsg/tests/test_nvhandle_desc_array.c
```diff
@@ -0,0 +1,48 @@
...
+ * COPYING for details.
+ *
+ */
+
+#include "logmsg/nvhandle-descriptors.h"
+
+#include "string.h"
+#include <criterion/criterion.h>
+
+Test(nvhandle_desc_array, add_elements)
```
Kokan: It would be nice to have a test, when the `NVHandleDesc` internal size expands, and possibly one with a fake app hook call and gc.

lib/apphook.c
furiel: Thanks for comment! More to it: calling update_main_chain_state should not depend on AHFLAGS_PERSISTENT. I will remove the whole if (remove_hook) around it.

furiel:
After discussing irl with the team, we decided that instead of garbage collecting the chunks, these will be deallocated only during shutdown. The benefit is the patch will be much simpler, and the new NVHandleDescArray implementation will be self contained. The downside is the outdated chunks will not be turned back to glibc. However, the total size of these chunks at worst case is the current size of the array, which is more acceptable compared to the garbage collecting alternative.

lib/logmsg/tests/test_nvhandle_desc_array.c
furiel: The gc part is reverted. Not relevant anymore.

lib/logmsg/nvtable.c
```diff
@@ -60,22 +60,25 @@ nv_registry_alloc_handle(NVRegistry *self, const gchar *name)
 
   len = strlen(name);
   if (len == 0)
+    goto exit;
+
+  if (len > 255)
```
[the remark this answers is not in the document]
furiel: Right, I reworded the commit message, and removed the cyclomatic complexity part.

kira-syslogng:

lib/logmsg/nvhandle-descriptors.h
```diff
@@ -34,4 +34,18 @@ struct _NVHandleDesc
...
+  guint   len;
+  guint   alloc;
+} NVHandleDescArray;
+
+#define NVHANDLE_DESC_ARRAY_INITIAL_SIZE 256
+
+NVHandleDescArray *nvhandle_desc_array_new();
+void nvhandle_desc_array_free(NVHandleDescArray *self);
+void nvhandle_desc_array_append(NVHandleDescArray *self, NVHandleDesc *desc);
+#define nvhandle_desc_array_index(self, i)      (((NVHandleDesc*) (self)->data) [(i)])
```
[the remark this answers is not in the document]
furiel: I would expect `nvhandle_desc_array_index()` to return with `NVHandleDesc`, though on the caller site we only need a pointer to it. However with inline function one cannot write `&nvhandle_desc_array_index(...)` because one cannot take the address of an rvalue. With the `#define` version, `&nvhandle_desc_array_index(...)` will work.

Alternatively we could write an inline function that returns a pointer, but then we need to find better name for the function.  Still for dereferencing, me seems more natural to return the whole object than a pointer to it.

So in this case imo the `#define` is better.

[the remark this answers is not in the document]
Kokan: Ok.

kira-syslogng:

lib/logmsg/nvhandle-descriptors.h
```diff
@@ -34,4 +34,19 @@ struct _NVHandleDesc
   guint8 name_len;
 };
 
+typedef struct
+{
+  NVHandleDesc *data;
+  guint   len;
+  guint   alloc;
```
gaborznagy: Would it be better for understanding if 'alloc' would be 'allocated_len'?

kira-syslogng:

PR #1840, opened 2018-01-14
Syslog ng debun improvements

jszigetvari:
Details:

-  cosmetic changes to remove backtick-style code invokations in favor of dollarsign+parentheses
-  gather the persist file even when vardir is too big
-  gather two sets of syslog-ng stats, with Unix timestamps to allow EPS counting based on the diffs

### DIFF
--- a/contrib/syslog-ng-debun
+++ b/contrib/syslog-ng-debun
@@ -7,7 +7,7 @@
 ### This software may be used and distributed according to the terms of GNU GPLv2
 ### http://www.gnu.org/licenses/gpl-2.0.html
 
-version="0.3.10.20180114"
+version="0.3.10.20180119"
 
 ### Check for "local" variable support
 if type local >/dev/null; then
@@ -203,6 +203,8 @@ fi
 [ -n "$privacy_mode" ] && [ "x$debug_params" = "x$default_debug_params" ] && debug_params="$default_ldebug_params"
 
 syslogbin=${binprefix}/sbin/syslog-ng
+syslogngctlbin=${binprefix}/sbin/syslog-ng-ctl
+syslogngquerybin=${binprefix}/sbin/syslog-ng-query
 vardir=${binprefix}/var
 piddir=${vardir}/run
 confdir=${binprefix}/etc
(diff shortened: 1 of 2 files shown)


kira-syslogng:

Kokan:
@kira-syslogng ok to test

kira-syslogng:

contrib/syslog-ng-debun
```diff
@@ -672,20 +678,34 @@ acquire_syslog_pids () {
...
 	echo ${syslogbin}-ctl stats
-	${syslogbin}-ctl query get "*" > ${tmpdir}/syslog.query.all 2>/dev/null
+	tsdata=$( timestamp )
+	${syslogbin}-ctl query get "*" > ${tmpdir}/syslog.query.all.${tsdata} 2>/dev/null
 	echo ${syslogbin}-ctl query get "*"
 	if [ -x ${syslogbin}-query ]; then
 		#if we reach this brach, then syslog.query.all - as generated above - will not contain any meaningful data
-		${syslogbin}-query sum "*" > ${tmpdir}/syslog.query.all 2>/dev/null
+		rm ${tmpdir}/syslog.query.all.${tsdata}
+		ts=$( timestamp )
```
Kokan: This variable is not used anywhere ?

contrib/syslog-ng-debun
```diff
@@ -672,20 +678,34 @@ acquire_syslog_pids () {
...
+	local tsdata
+	tsdata=$( timestamp )
+	#handy-dandy delay magic, triggered when we'd step on our own foot
+	[ -f ${tmpdir}/syslog.stats.${tsdata} ] && sleep 5 && tsdata=$( timestamp )
+
+	${syslogbin}-ctl stats > ${tmpdir}/syslog.stats.${tsdata} 2>&1
 	echo ${syslogbin}-ctl stats
-	${syslogbin}-ctl query get "*" > ${tmpdir}/syslog.query.all 2>/dev/null
+	tsdata=$( timestamp )
+	${syslogbin}-ctl query get "*" > ${tmpdir}/syslog.query.all.${tsdata} 2>/dev/null
```
Kokan: Maybe we could save the `stderr` also for the query.

contrib/syslog-ng-debun
```diff
@@ -1243,6 +1264,7 @@ setup_env_solaris () {
 	netstatlunp () { netstat -P udp -na ; }
 	myplimit () { plimit $1 ; }
 	free () { prtconf | grep Mem ; printf Pagesize:\  ; pagesize -a ; }
+	timestamp () { /usr/bin/perl -e 'print time, "\n";' ; }
```
Kokan: Why the `timestamp` re-defined here ?

jszigetvari: The script has to run on multiple different platforms. This means that we have to handle somehow when a given OS behaves differently than the others. That is why we introduced OS-specific re-definitions of shell functions or variables.

In this specific case, Solaris does not support the date +%s way of getting the unix timestamp, so we had to override the default way of doing it.

contrib/syslog-ng-debun
jszigetvari: Saving the stderr output is not really helpful in this case. I don't want the user to see it, and we won't use it do determine why it did not work. The other parts of the script will give us enough information about syslog's state, so that we won't need stderr here.

contrib/syslog-ng-debun
Kokan: Ok, one minor thing the path of `perl` I think should not be absolute (at least not hard coded absolute path), see also the `sed_equivalent_cmd`, where `perl` is relative.

contrib/syslog-ng-debun
Kokan: Ok.

contrib/syslog-ng-debun
jszigetvari: Corrected.

contrib/syslog-ng-debun
jszigetvari: That's a typo all right.
Corrected.

kira-syslogng:

contrib/syslog-ng-debun
```diff
@@ -649,43 +655,57 @@ acquire_syslog_pids () {
...
-	$syslogbin --version > ${tmpdir}/syslog.version
-	head -1 ${tmpdir}/syslog.version
-	${syslogbin}-ctl stats > ${tmpdir}/syslog.stats 2>&1
+acquire_syslog_stats () {
+	local tsdata
+	tsdata=$( timestamp )
+	#handy-dandy delay magic, triggered when we'd step on our own foot
+	[ -f ${tmpdir}/syslog.stats.${tsdata} ] && sleep 5 && tsdata=$( timestamp )
+
+	${syslogbin}-ctl stats > ${tmpdir}/syslog.stats.${tsdata} 2>&1
```
mitzkia: It would be better (for reading) to create a separate variable for ${syslogbin}-ctl it is used in a few places.

PR #1847, opened 2018-01-16
persist-state: fix potential fd leak on file size too large

ibmibmibm:
Signed-off-by: Shen-Ta Hsieh <ibmibmibm.tw@gmail.com>


kira-syslogng:

PR #1853, opened 2018-01-19
autotools: small fixes related to --disable-all-modules

furiel:
This patchset fixes
- python unit test should not be compiled if python is disabled
- if ipv6 support is disabled, test_filter_netmask6 should not be compiled
- There was a whitespace in libtest makefile after backslash, causing autotools warning
- The compile of http module was unaffected by --enable/disable-all-modules, even though the default detection state was auto


kira-syslogng:

PR #1856, opened 2018-01-21
add conditionals into the config syntax

bazsi:
We are already able to do conditional evaluation with a bit arcane syntax:
    
    junction {
      channel {
        filter { EXPR1 };
        ...
        flags(final);
      };
      channel {
        filter { EXPR2 };
        ...
        flags(final);
      };
      channel {
        flags(final);
      };
    };
    
    However, this is not very intuitive. This patch adds syntactic sugar using
    the if / else keywords to achieve the same, such as:
    
    if (message('foo')) {
       ...
    } elif (message('bar')) {
       ...
    } else {
       ...
    }
    
    This should make the configuration for complex of stuff a lot simpler.
    
### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -172,10 +172,13 @@ extern struct _StatsOptions *last_stats_options;
 %token KW_BLOCK                       10007
 %token KW_JUNCTION                    10008
 %token KW_CHANNEL                     10009
+%token KW_IF			      10010
+%token KW_ELSE			      10011
+%token KW_ELIF			      10012
 
 /* source & destination items */
-%token KW_INTERNAL                    10010
-%token KW_FILE                        10011
+%token KW_INTERNAL                    10020
+%token KW_FILE                        10021
 %token KW_SYSLOG                      10060
 
 /* option items */
@@ -422,6 +425,8 @@ DNSCacheOptions *last_dns_cache_options;
 %type   <ptr> log_last_junction
 %type   <ptr> log_junction
 %type   <ptr> log_content
+%type   <ptr> log_conditional
+%type   <ptr> log_if
 %type   <ptr> log_forks
 %type   <ptr> log_fork
 
@@ -712,6 +717,7 @@ log_item
         | KW_REWRITE '{' rewrite_content '}'    { $$ = log_expr_node_new_rewrite(NULL, $3, &@$); }
         | KW_DESTINATION '(' string ')'		{ $$ = log_expr_node_new_destination_reference($3, &@$); free($3); }
         | KW_DESTINATION '{' dest_content '}'   { $$ = log_expr_node_new_destination(NULL, $3, &@$); }
+        | log_conditional			{ $$ = $1; }
         | log_junction                          { $$ = $1; }
 	;
 
@@ -740,6 +746,42 @@ log_forks
 log_fork
         : KW_LOG '{' log_content '}'            { $$ = $3; }
         | KW_CHANNEL '{' log_content '}'        { $$ = $3; }
+        ;
+
+log_conditional
+        : log_if				{ $$ = $1; }
+        | log_if KW_ELSE '{' log_content '}'
+          {
+            log_expr_node_conditional_set_false_branch_of_the_last_if($1, $4);
+            $$ = $1;
+          }
+        ;
+
+log_if
+        : KW_IF '(' filter_content ')' '{' log_content '}'
+          {
+            $$ = log_expr_node_new_conditional_with_filter($3, $6, &@$);
+          }
+        | KW_IF '{' log_content '}'
+          {
+            $$ = log_expr_node_new_conditional_with_block($3, &@$);
+          }
+        | log_if KW_ELIF '(' filter_content ')' '{' log_content '}'
+          {
+            LogExprNode *false_branch;
+
+            false_branch = log_expr_node_new_conditional_with_filter($4, $7, &@$);
+            log_expr_node_conditional_set_false_branch_of_the_last_if($1, false_branch);
+            $$ = $1;
+          }
+        | log_if KW_ELIF '{' log_content '}'
+          {
+            LogExprNode *false_branch;
+
+            false_branch = log_expr_node_new_conditional_with_block($4, &@$);
+            log_expr_node_conditional_set_false_branch_of_the_last_if($1, false_branch);
+            $$ = $1;
+          }
         ;
 
 log_content
--- a/lib/cfg-parser.c
+++ b/lib/cfg-parser.c
@@ -60,6 +60,9 @@ static CfgLexerKeyword main_keywords[] =
   { "options",            KW_OPTIONS },
   { "include",            KW_INCLUDE, },
   { "block",              KW_BLOCK },
+  { "if",                 KW_IF },
+  { "else",               KW_ELSE },
+  { "elif",               KW_ELIF },
 
   /* source or destination items */
   { "internal",           KW_INTERNAL },
--- a/lib/cfg-tree.c
+++ b/lib/cfg-tree.c
@@ -371,6 +371,8 @@ log_expr_node_lookup_flag(const gchar *flag)
     return LC_FINAL;
   else if (strcmp(flag, "flow_control") == 0 || strcmp(flag, "flow-control") == 0)
     return LC_FLOW_CONTROL;
+  else if (strcmp(flag, "drop_unmatched") == 0 || strcmp(flag, "drop-unmatched") == 0)
+    return LC_DROP_UNMATCHED;
   msg_error("Unknown log statement flag", evt_tag_str("flag", flag));
   return 0;
 }
@@ -616,6 +618,9 @@ cfg_tree_propagate_expr_node_properties_to_pipe(LogExprNode *node, LogPipe *pipe
 
   if (node->flags & LC_FLOW_CONTROL)
     pipe->flags |= PIF_HARD_FLOW_CONTROL;
+
+  if (node->flags & LC_DROP_UNMATCHED)
+    pipe->flags |= PIF_DROP_UNMATCHED;
 
   if (!pipe->expr_node)
     pipe->expr_node = node;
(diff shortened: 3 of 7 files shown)


bazsi:
I would appreciate feedback from @faxm0dem @czanik @fekete-robert @lbudai

kira-syslogng:

faxm0dem:
What about an `always` clause?

bazsi:
What would that mean? Can you pls help me with a sample config?

Thanks

faxm0dem:
Some programming languages provide a branch in if-else statement that is always executed

```
if (message('foo')) {
   ...
} elif (message('bar')) {
   ...
} else {
   ...
} always {
   # this will always be executed
}
```

fekete-robert:
Hi Bazsi, 

I really like the idea, just have a few questions;
 * the if{} statement would be equivalent with the junction{}? So it could appear practically anywhere in the log path.
 * Are there any limitations on the (message('foo')) part, or is it a full-blown filter expression (with regex/template-functions/boolean operators/whatever)?
 * Is nesting such statements supported? Like if x {... elif z { if y {} } }, or you have to bring this to the first level and combine the two levels of filters with boolean operators?

Robi

lbudai:
@faxm0dem : `if-else-always` as programming language construction ??? You mean `try-catch-finally`? : that has a similar semantic, but in another context.

faxm0dem:
@lbudai yeah something like that, and the idea was dumb, as you could just add the statement outside of the block :-)

bazsi:
Couple of answers:

@fekete-robert 

>     the if{} statement would be equivalent with the junction{}? So it could appear practically anywhere in the log path.

yes. Basically it can go anywhere.

>     Are there any limitations on the (message('foo')) part, or is it a full-blown filter expression (with regex/template-functions/boolean operators/whatever)?

It's a full blown filter expression, and uses the same syntax even boolean operators. the message() was just an example.

>     Is nesting such statements supported? Like if x {... elif z { if y {} } }, or you have to bring this to the first level and combine the two levels of filters with boolean operators?

these things can nest as well. Here's the config I've tested this with:

```
@version: 3.13

log {
        source { unix-stream("log"); };
        if (message("almafa")) {
                destination { file("/dev/stdout" persist-name('aaa') template(">>>>ALMA<<<< $ISODATE $HOST $MSGHDR$MSG\n")); };
        } elif (message("belafa")) {
                if (message("magja")) {
                        destination { file("/dev/stdout" persist-name('bbb1') template(">>>>BELA MAG<<<< $ISODATE $HOST $MSGHDR$MSG\n")); };
                } elif (message("termese")) {
                        destination { file("/dev/stdout" persist-name('bbb2') template(">>>>BELA TERMES<<<< $ISODATE $HOST $MSGHDR$MSG\n")); };
                } elif (message("levele")) {
                        destination { file("/dev/stdout" persist-name('bbb3') template(">>>>BELA LEVEL<<<< $ISODATE $HOST $MSGHDR$MSG\n")); };
                };
        } elif (message("celafa")) {
                destination { file("/dev/stdout" persist-name('ccc') template(">>>>CELA<<<< $ISODATE $HOST $MSGHDR$MSG\n")); };
        } else {
                destination { file("/dev/stdout" persist-name('xxx') template(">>>>XXXX<<<< $ISODATE $HOST $MSGHDR$MSG\n")); };
        };
};
```

Sorry for the funny names in the cases :P, you can test it using:

```
$ logger -u log almafa
$ logger -u log belafa magja
$ logger -u log belafa termese
$ logger -u log belafa levele
$ logger -u log celafa
$ logger -u log something else
```

bazsi:
@lbudai 

> Nice feature for users, and well... it is interesting for developers :)
> 
> Is it necessary to set false branch in log_expr_node_new_conditional?
> All the callers set it to NULL and in case of elif it is strange that you create a false_branch with a log_expr_node_new_conditional call and set its false_branch argument to NULL then you set the false_branch as false branch to the conditional LogExprNode created for the if.


Yes, we always need the false branch, even if it is NULL. If the false branch is not set and the if expression does not match, the message would be dropped. We always need an "else" branch, that does nothing, but accepts the message unchanged.

This is basically equivalent to this config:

```
junction {
    channel {
        filter { EXPRESSION };
        flags(final);
    };
    channel {
        flags(final);
    };
};
```

> 
> (Not talking about the complexity of nexts: which is false and true branch: maybe some wrappers could be simplify the sitation: some static inlines, like get/set_false_branch :) )

Yup, possibly we could do a conditional specific wrapper around LogExprNode and provide accessors to manipulate it. That's basically what I've tried to do with adding the "conditional" qualifier to all related functions (that is two add the moment). We could maybe extract all LogExprNode types to a separate file making cfg-tree.c much smaller and the code easier to navigate.

Or to create one further abstraction: a class that is completely grammar specific which then generates the LogExprNode instances at the end of the parsing phase.

> 
> So my questions would be:
> 
>     is it possible that log_expr_node_new_conditional creates a dummy false branch inside the function (->remove the false_branch argument)?

Like explained above, that would not work, at least until we are using LogExprNode instances to represent if expressions.

>     (later it will be replaced by the log_expr_node_conditional_set_false_branch_of_the_last_if function...)
> 
>     is it possible to add some helper functions/alias (-> branches, true branch, false branch)?
> 

Yeah, this could work. Don't forget that LogExprNode is the common denominator for all things parsed from the config, this way its abstraction level is pretty low. However so far, we could get away with this, as we don't have too many constructs (e.g. sequence, junction, pipe, reference, and now the conditional).

I'll try to make the code easier to read with accessors, but as far as we don't increase abstraction somewhat, this would still be somewhat difficult to read. And increasing the abstraction would mean a lot of boilerplate code. (e.g. a class for all different uses of LogExprNode).

One thing that also relates to this question is the question optimizing the configuration. LogExprNode instances almost directly map to the LogPipe structure used at runtime, and due to how the if statements and structured we would probably have a lot of "empty" LogPipe instances that do nothing but forward the message along the pipeline. At runtime, we can easily have 20+ hops for a single configuration and we know that 20+ indirect function calls at runtime, doing it a few hundred thousands per second incurs some penalties :)

Anyhow, I would try to create some accessors, maybe even add some comments to cfg-tree.c, to make it more readable.

Thanks for the comments.

bazsi:
@faxm0dem I think that "always" clause is superficial just as @lbudai has mentioned. You simply add another node to do processing after the if-elif-else structure:

```
log {
    source { whatever(); };
    if (message("foo")) {
    } elif (message("bar")) {
    } elif (message("baz")) {
    } else {
    };
    # something here would execute after any of the branches in the conditionals was taken
};
```

lbudai:
@bazsi: one additional question : do you plan to replace existing junction/channels to if-else? (eg.: in app-parser?)

bazsi:
@lbudai: yes, I planned doing that, although the first one that came into mind was not the app-parser(), but cisco-parser(), my memories are not fond about that.

kira-syslogng:

lbudai:
@bazsi: thanks for the last two patches, and for the detailed description.
It would be nice to find the time to gather all these information into a document...

(Travis just 'joking', I've restarted the job... )
Do you want to add further patches(you mentioned some tests), or can we merge it as-is?

bazsi:
I have now changed cisco-parser() to use the new if construct, seems much nicer to me :)

However, I've found a behaviour which is not intuitive and not necessarily good as it stands:

```
if (message("whatever")) {
    parser { csv-parser(columns(...) flags(drop-invalid)); };
};
```

e.g. a parser that can actually drop messages. How that is currently compiled into LogExprNode's would cause the parser() part of the statement to be part of the condition as well. E.g. if the filter matches and the parser is not, then flags(final) would cause this to propagate to the else branch, even if the condition has matched.

As I think of it, this is actually a useful behaviour, but certainly not expected for an "if" keyword.

Hmm... I have to give it some more thought.

Please don't merge it for now.

kira-syslogng:

kira-syslogng:

bazsi:
The last incarnation should add handling of filter + parser combinations by implementing a syntax originally recommended by Gergely Bodnar. With that you have two syntaxes:

```
# notice there's no condition in the if
if {
   filter { ... };
   parser { .... };
}

# the conditional part is the combination of the filter plus parsers. Should any of them fail, processing continues along the else branch.
```

And the other is:

```
if (condition) {
    ...
};
```

This one works as expected. If the condition matches, the else will not be taken, even if the actions contain a parser that would drop the message.

Look at the changed scl blocks how much this change helps in clarity of syntax.

kira-syslogng:

kira-syslogng:

bazsi:
I have cleaned up the branch and prepared for merging by dropping unrelated changes into separate PRs. See #1898 and #1899

bazsi:
btw: I've manually tested these patches extensively and also attempted to use the pytest based framework by @mitzkia to automate those testcases, but as I see that is not yet ready for prime time. It took me a while just to be able to launch tests. 

I'll still be experimenting in that area, but I would merge this without testing at this point:

1) unit tests are not able to cope with the complexity of these testcases. I could test logmpx in isolation, however that is simply not going to cut it as the various testcases require a more complex log pipeline structure to reveal problems.

2) the old functest framework could be used, but honestly that again is not something I would want to allocate too much time for.

3) I literally have no chance to launch the kira based test suite locally.

At this point I might need to revert to the functest stuff, but I still got to make friends with that idea.

Cheers,
Bazsi

kira-syslogng:

bazsi:
rebased against master.

kira-syslogng:

lbudai:
@bazsi : I'm merging this (and later we will add new kira tests).

MANOJLAL:
Hi Guys,
Will the below flags(final) inside the if condition work? 
It is not working when I tested with messages that matches if condition in statement 1, then are again getting processed by the log statement 2. Could you guys please let me know about this behaviour.

Log statement 1:log{
if{
filter("something");
parser("something");
destination();
**flags(final);**
}else{
parser("something");
}
};

Log Statemnt 2: log{
some other log
}

Issue #1851, opened 2018-01-18
disq location cannot be modified


faxm0dem:
# syslog-ng

When setting up a destination with disk buffer, it is not possible to change the path anymore:

* changing the path in `syslog-ng.conf` and then reloading the config does not use the new path
* changing the path in `syslog-ng.conf` and then restarting syslog-ng does not use the new path

The only workaround I found was to delete the persist-file

## Version of syslog-ng

```
syslog-ng 3.9.1
Installer-Version: 3.9.1
Revision:
Module-Directory: //usr/lib64/syslog-ng
Module-Path: //usr/lib64/syslog-ng
Available-Modules: cryptofuncs,syslogformat,system-source,curl,afstomp,graphite,kvformat,afamqp,date,cef,mod-java,pseudofile,confgen,afprog,riemann,add-contextual-data,affile,afsocket,mod-python,linux-kmsg-format,basicfuncs,geoip-plugin,disk-buffer,afuser,json-plugin,tfgetent,sdjournal,dbparser,afsmtp,csvparser
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on

```

## Platform

CentOS Linux release 7.3.1611 (Core)

# Issue

## Steps to reproduce

1. start syslog-ng
2. change location of disq buffer
3. reload or restart syslog-ng
4. location of disq buffer is unchanged

## Configuration

```
destination {
  elasticsearch2(
    ...
    disk-buffer(reliable(no) dir("/var/lib/syslog-ng-disq-01/") disk-buf-size(53687091200) mem-buf-size(1073741824))
    ...
  );
};
```

Kokan:
I think the reason is because syslog-ng cannot simply start a new disk queue, because in this case the messages in the queue are "lost". Of course the user could send those messages back to the syslog-ng.

But that is manual and I think not trivial that needed. The current version where the user have to explicitly must remove the persist file is a good day.

The syslog-ng could have some way to read the old disk queue, and start a new one where it saves the new messages.

faxm0dem:
But there are times where you simply want to change the directory where the disq files are located, even if the files are empty.
Removing the persist file seems a very clumsy way to achieve it IMHO.

FTR I managed to solve my use-case which was simply to change the storage backend behind `/var/lib/syslog-ng-disq` by doing the following:

[...]

PR #1861, opened 2018-01-25
Disk queue directory change in configuration

Kokan:
When the `disk-buffer` directory is changed, but the destination already has a disk buffer file, syslog-ng continues to use the old path without any warning.
This modification aims to notify the user about the fact that disk queue is not changed.
Solves first part of the #1851.

### DIFF
--- a/modules/diskq/diskq.c
+++ b/modules/diskq/diskq.c
@@ -30,6 +30,16 @@
 #include "logqueue-disk-non-reliable.h"
 #include "persist-state.h"
 
+static gboolean
+log_queue_disk_is_file_in_directory(const gchar *file, const gchar *directory)
+{
+  gchar *basedir = g_path_get_dirname(file);
+  const gboolean result = (0 == strcmp(basedir, directory));
+
+  g_free(basedir);
+  return result;
+}
+
 static LogQueue *
 _acquire_queue(LogDestDriver *dd, const gchar *persist_name, gpointer user_data)
 {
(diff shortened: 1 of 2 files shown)


kira-syslogng:

kira-syslogng:

faxm0dem:
Thanks! but I find the message not very helpful. I'd say:

```
The disk buffer directory was changed in the configuration, but the disk queue file cannot be moved.
To make changes effective, please stop syslog-ng, remove the persist file, and start syslog-ng.
Warning: this will also remove all other stateful information, like the file source location cursors.
```

or something similar

Kokan:
I could change the message, hopefully one that describes the problem at hand.
But I would rather avoid a lengthy message, that describes step-by-step, how to remove it.
I think it should state, what is the not intuitive event that happened, and the maybe a state to check the documentation.
The documentation should cover a detailed description, how to resolve it, especially when it is related to poke the persist file.
Of course I am open to suggestions. But I think your text above a bit too long. I'll try to improve the message.

faxm0dem:
As you wish, but then change it to:

```
The disk buffer directory has changed in the configuration, but the disk queue file cannot be moved.
```

Kokan:
Thanks for the suggestion replaced the warning with that message.

kira-syslogng:

modules/diskq/diskq.c
```diff
@@ -30,6 +30,16 @@
 #include "logqueue-disk-non-reliable.h"
 #include "persist-state.h"
 
+static gboolean
+log_queue_disk_is_file_in_directory(const gchar *file, const gchar *directory)
+{
+  gchar *basedir = g_path_get_dirname(file);
+  const gboolean result = (0 == strcmp(basedir, directory));
```
MrAnno: An `strcmp` call might not be enough here, because `disk_queue_options_set_dir()` uses the specified directory string from the configuration file without any normalization.

For example, the directory specified by the user can be:
`"/var/disk-buffers/"`
or just
`"/var/disk-buffers"`

kira-syslogng:

modules/diskq/diskq.c
Kokan: I've updated, to normalize during the configuration time. The only side effect of this solution, if a `syslog-ng` is started with a persist file, that already has a not normalized path. But in this case, we only do not print this warning, and the functionality is not changed.
Does it acceptable ?

MrAnno: IMHO, it is perfectly acceptable.

PR #1864, opened 2018-01-26
On Mac the modules are in .dylib files

opoplawski:


kira-syslogng:

lib/plugin.c
```diff
@@ -314,6 +317,19 @@ plugin_dlopen_module_on_path(const gchar *module_name, const gchar *module_path)
       if (is_file_regular(plugin_module_name))
         break;
 #endif
+#ifdef __APPLE__
+      /* On Mac the modules are in .dylib files */
+      dot = strrchr(plugin_module_name, '.');
+      if (dot)
+        {
+          *dot = 0;
+          p = g_strdup_printf("%s.dylib", plugin_module_name);
```
Kokan: I think those 3 cases could be replaced with this one:
```C
       dot = strrchr(plugin_module_name, '.');
       if (dot)
         {
           *dot = 0;
           p = g_strdup_printf("%s.%s", plugin_module_name, G_MODULE_SUFFIX);
           g_free(plugin_module_name);
           plugin_module_name = p;
         }
       if (is_file_regular(plugin_module_name))
         break;
```

opoplawski: Yeah, that would do it too.  Thought about that just after submitting this...

Kokan: Could you update the PR with that version ?
The current version executes both `la` and `dylib` detection in case of apple.

Kokan:
@kira-syslogng ok to test

opoplawski:
Alternatively, we could figure out why syslog-ng is producing module shared objects with the .dylib suffix.  I see that many other software loadable modules use the .so suffix, e.g.:

```
/usr/libexec/apache2/libphp5.so
```

kira-syslogng:

Kokan:
I guess those lines has something to do with it: https://github.com/balabit/syslog-ng/blob/489a82520f05f9bbd29bdc7744574df49820f27f/configure.ac#L505-L506
That leads to:

> commit 9656bebb0312fbb70e7b01400e23ac3bb83b07fe
> Date:   Fri Jul 8 12:25:04 2016 +0200
>
>    configure.ac: MODULE_LDFLAGS linking for Mac OS X
>    
>    `-dylib` is needed because plugins can not be used for static linking,
>    unlike in GNU/Linux.


Although I do not have osx to test if it really is needed. I do not think it is an issue to have `dylib` if it is natural for osx.

opoplawski:
And that is because in configure.ac you have -dylib instead of -module:
```
        Darwin)
                MODULE_LDFLAGS="-avoid-version -dylib"
```
From this:
commit 9656bebb0312fbb70e7b01400e23ac3bb83b07fe
Author: bkil-syslogng <tamas.nagy@balabit.com>
Date:   Fri Jul 8 12:25:04 2016 +0200

    configure.ac: MODULE_LDFLAGS linking for Mac OS X

    `-dylib` is needed because plugins can not be used for static linking,
    unlike in GNU/Linux.

    Signed-off-by: bkil-syslogng <tamas.nagy@balabit.com>

But I just don't understand the static linking reference.

Kokan:
Hello,

During the `cmake` with `add_library` also has the option `STATIC`, `SHARED` and `MODULE`, while the `SHARED` and `MODULE` creates the `so`, but the later one is not possible to use for dynamic linking.
The unit tests are written in a way, that they link against the module they are testing. I think it can be the same with `dylib` vs `so` on mac os.

Could you check this assumption ?

I have tried to check it via TravisCI job: https://travis-ci.org/Kokan/syslog-ng/jobs/333957753
But it does not have `Criterion`, which `syslog-ng` is using for the unit tests.

MrAnno:
Hi,

I've started to fix the Travis macOS build. Unfortunately, I won't have time to finish it in the near future, so I share a few things here.

To fix Travis:
- https://github.com/MrAnno/syslog-ng/commit/628b0ef2934d918f632ae17d51baff8aa86502e0
- fix #1866
- fix the ivykis inotify compat issue in affile:
```
./modules/affile/directory-monitor-inotify.h:26:10: fatal error: 'iv_inotify.h' file not found
#include <iv_inotify.h>
         ^
```

Kokan:
@opoplawski As you said the #1865 is blocked by this PR. I think it is mergable, if the module load part is resolved with that minor refactor.

PR #1865, opened 2018-01-30
Support different shared library suffixes, e.g. .dylib on MacOS

opoplawski:


kira-syslogng:

Kokan:
@kira-syslogng ok to test

kira-syslogng:

cmake/add_tests.cmake
```diff
@@ -34,7 +34,9 @@ function (add_unit_test)
   add_executable(${ADD_UNIT_TEST_TARGET} ${ADD_UNIT_TEST_SOURCES})
   target_link_libraries(${ADD_UNIT_TEST_TARGET} ${ADD_UNIT_TEST_DEPENDS} syslog-ng)
   target_include_directories(${ADD_UNIT_TEST_TARGET} PUBLIC ${ADD_UNIT_TEST_INCLUDES})
-  set_property(TARGET ${ADD_UNIT_TEST_TARGET} APPEND PROPERTY LINK_FLAGS "-Wl,--no-as-needed")
+  if (NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
```
Kokan: Are you able to execute the unit test(s) ?

If the unit tests are running and passed without this option, I would rather use a detection for compiler/linker option.

```cmake
INCLUDE(CheckCCompilerFlag)
SET(CMAKE_REQUIRED_FLAGS "-Wl,--no-as-needed")
CHECK_C_COMPILER_FLAG("" RESULT1)
IF(RESULT1)
    MESSAGE("Supported: -Wl,--no-as-needed")
ELSE()
    MESSAGE("Not supported: -Wl,--no-as-needed")
ENDIF()
```

The above detection should be in the root cmake file.

opoplawski:
I'm not yet able to fully compile and test due to issues noted in #1864

PR #1869, opened 2018-02-01
logproto-client: make the logproto_client_get_id to a virtual function

Kokan:
* The `LogProtoClient` provides an interface, that should alone does not have knowledge about the underlying transport's fd.
* The `log_proto_client_get_fd` renamed to `log_proto_client_get_id`, because I think that name describes better, what this function should do. In this case it shows that possible the `fd` is not the only information we could have here.

Does it also okay for the `LogTransport` to have this `log_transport_get_id` function ?

### DIFF
--- a/lib/logproto/logproto-client.c
+++ b/lib/logproto/logproto-client.c
@@ -49,11 +49,18 @@ log_proto_client_free(LogProtoClient *s)
   g_free(s);
 }
 
+gint
+_get_id(LogProtoClient *self)
+{
+  return log_transport_get_id(self->transport);
+}
+
 void
 log_proto_client_init(LogProtoClient *self, LogTransport *transport, const LogProtoClientOptions *options)
 {
   self->validate_options = log_proto_client_validate_options_method;
   self->free_fn = log_proto_client_free_method;
+  self->get_id  = _get_id;
   self->options = options;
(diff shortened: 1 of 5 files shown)


kira-syslogng:

Kokan:
@bazsi By "The result of this function" are you referring to the `log_transport_get_id` ?
I would let that part go.

The main focus would be the `LogProtoClient` and its `log_proto_client_get_fd` function, as that is not really used just to print some log message in the `LogWriter`.
Maybe that function should have been used in other places, where the `transport->fd` would be not make sense not to be an fd. In the log it should be just an identifier to differentiate the `LogProtoClient`s.

Actually I do not see any code - other then the above print - that would force the `LogProtoClient` to have a `transport`, thus it would break the code if there would be a `LogProtoClient` without one.

I would make it virtual, so later a child could redefine to its own liking.

bazsi:
The fd is used at a number of places where we log about the connection, eg when we accept a new connection or connect to a server. These error messages need to be correlated to those.

Fd might not be the best identifier, i would go as far to say that we should be using the peer address in all such messages, then we could drop the fd altogether.

But renaming it to id doesn't help the user, especially the other cases cook l still report the fd.

PR #1871, opened 2018-02-01
Fixing KVParser rare crash in race condition

szemere:
In rare conditions KVParser may crash due to concurrent access to the "formatted_key" field.
Solution: Moved formatted_key from KVParser into a _process local variable.

Also splitted the _get_formatted_key function along the "if prefix" logic, because it can be decided in config time. (Function called for each key in _process.)

Question:
- Is it needed to modify the key_formatter function anywhere else than kv_parser_set_prefix?

### DIFF
--- a/lib/scanner/kv-scanner/kv-scanner.c
+++ b/lib/scanner/kv-scanner/kv-scanner.c
@@ -344,19 +344,3 @@ kv_scanner_init(KVScanner *self, gchar value_separator, const gchar *pair_separa
   self->is_valid_key_character = _is_valid_key_character;
   self->stop_char = 0;
 }
-
-KVScanner *
-kv_scanner_new(gchar value_separator, const gchar *pair_separator, gboolean extract_stray_words)
-{
-  KVScanner *self = g_new0(KVScanner, 1);
-
-  kv_scanner_init(self, value_separator, pair_separator, extract_stray_words);
-  return self;
-}
-
-void
-kv_scanner_free(KVScanner *self)
-{
-  kv_scanner_deinit(self);
-  g_free(self);
-}
--- a/lib/scanner/kv-scanner/kv-scanner.h
+++ b/lib/scanner/kv-scanner/kv-scanner.h
@@ -107,7 +107,5 @@ kv_scanner_set_stop_character(KVScanner *self, gchar stop_char)
 }
 
 gboolean kv_scanner_scan_next(KVScanner *self);
-KVScanner *kv_scanner_new(gchar value_separator, const gchar *pair_separator, gboolean extract_stray_words);
-void kv_scanner_free(KVScanner *self);
 
 #endif
--- a/lib/scanner/kv-scanner/tests/test_kv_scanner.c
+++ b/lib/scanner/kv-scanner/tests/test_kv_scanner.c
@@ -139,12 +139,19 @@ typedef struct Testcase_t
   KVContainer expected;
 } Testcase;
 
-KVScanner *
-create_kv_scanner(const ScannerConfig config)
+static KVScanner *
+create_kv_scanner(const ScannerConfig *config)
 {
-  KVScanner *scanner = kv_scanner_new(config.kv_separator, config.pair_separator, config.extract_stray_words);
-  scanner->transform_value = config.transform_value;
+  KVScanner *scanner = g_new(KVScanner, 1);
+  kv_scanner_init(scanner, config->kv_separator, config->pair_separator, config->extract_stray_words);
+  kv_scanner_set_transform_value(scanner, config->transform_value);
   return scanner;
+}
+
+static inline void
+kv_scanner_free(KVScanner *scanner)
+{
+  g_free(scanner);
 }
 
 #define VARARG_STRUCT(VARARG_STRUCT_cont, VARARG_STRUCT_elem, ...) \
@@ -189,7 +196,7 @@ _expect_kvq_triplets(KVScanner *scanner, KVQContainer args, gchar **error)
 
 #define _IMPL_EXPECT_KVQ(SCANNER_config, TEST_KV_SCAN_input, ...) \
   do { \
-    KVScanner *scanner = create_kv_scanner(SCANNER_config); \
+    KVScanner *scanner = create_kv_scanner(&SCANNER_config); \
     gchar *error = NULL; \
     \
     kv_scanner_input(scanner, TEST_KV_SCAN_input);            \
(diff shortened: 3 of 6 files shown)


kira-syslogng:

kira-syslogng:

modules/kvformat/kv-parser.c
```diff
@@ -99,8 +102,10 @@ _process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_options, co
 {
   KVParser *self = (KVParser *) s;
   KVScanner kv_scanner;
+  self->init_scanner(self, &kv_scanner);
+  if (self->prefix)
+    g_string_assign(kv_scanner.formatted_key, self->prefix);
```
bazsi: I dont like that we are poking this value into kv_scanner

modules/kvformat/kv-parser.c
```diff
@@ -99,8 +102,10 @@ _process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_options, co
...
 
-  kv_parser_init_scanner(self, &kv_scanner);
   log_msg_make_writable(pmsg, path_options);
```
bazsi: It is a convention that we have an inline function around virtual methods, why do we get rid of this?

lib/scanner/kv-scanner/kv-scanner.h
```diff
@@ -46,6 +46,7 @@ struct _KVScanner
   const gchar *pair_separator;
   gsize pair_separator_len;
   gchar stop_char;
+  GString *formatted_key;
 
   KVTransformValueFunc transform_value;
```
bazsi: I can't see why the kvscanner needs the formatted key value. The only reason of moving this here as I understand the patch is that it is allocated on every invocation, but I think that is not enough reason to do feature envy and put a value used by a higher layer to here.

kira-syslogng:

kira-syslogng:

szemere:
@bazsi , thank You for the review comments! Hopefully everything solved properly.

lib/scanner/kv-scanner/kv-scanner.c
```diff
@@ -348,7 +348,7 @@ kv_scanner_init(KVScanner *self, gchar value_separator, const gchar *pair_separa
 KVScanner *
 kv_scanner_new(gchar value_separator, const gchar *pair_separator, gboolean extract_stray_words)
 {
-  KVScanner *self = g_new0(KVScanner, 1);
+  KVScanner *self = g_new(KVScanner, 1);
```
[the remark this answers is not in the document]
szemere: With g_new0 we call memset twice, because the init function (https://github.com/szemere/syslog-ng/blob/a4b2cbd06eff16bfe931a10df7c7a7bf55c6022d/lib/scanner/kv-scanner/kv-scanner.c#L332) also start with it.

I have chosen this one, because we always call init after creating a new instance of KVScanner, but there are cases where we call init on an already used object.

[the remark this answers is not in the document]
furiel: My bad, you can ignore my comment... I mistook with kv_parser_new. It is also interesting that kv_scanner_new() is called in unit tests.

kira-syslogng:

lbudai:
I think there is a kind of 'symmetry-breaking' around setting the `key_formatter` and setting the prefix.

Bazsi's comment is still not resolved: the formatted key is not stored anymore in kvscanner but the design remained the same :)

You are saying that it is possible to decide which format key method will be used in config time but you check prefix all the time in the `_process` method... and later you pass the formatted key to the format key callback. This means that the logic is splitted.

What if the keyformatter method wouldn't be part of `KVParser`?

Unfortunately C does not support lambdas...
In that case we could write something like this:


```
if (self->prefix)
  format_key = _format_key_with_prefix[capture:prefix](char *key) {...};
else
  format_key = _do_not_format_key[](char *key) {return key;};

...

log_msg_set_value_by_name(*pmsg, format_key(self, format_key(key), ...));
```

I'm not sure if this code is more effective than we just check the prefix inside the while-cycle...
(we always call a function via a fptr and passing variables to it...)

If we would check the prefix inside the while-cycle, we could have a method like this:
```
format_key(KVParser *self, const gchar *key)
{
  if (!self->prefix)
    return key;
  return _format_key_with_prefix(self->prefix, key);
}
```

lib/scanner/kv-scanner/tests/test_kv_scanner.c
```diff
@@ -139,14 +139,21 @@ typedef struct Testcase_t
   KVContainer expected;
 } Testcase;
 
-KVScanner *
+static KVScanner *
 create_kv_scanner(const ScannerConfig config)
```
lbudai: `config` variable is copied onto stack (-> `const` is not necessary here, or, pass a pointer to `ScannerConfig` instance).

szemere:
@lbudai : Thank You for the comments! Eliminated the virtual function. (Back to the original, with the small _with_prefix extraction.)

kira-syslogng:

modules/kvformat/kv-parser.c
```diff
@@ -75,16 +76,22 @@ kv_parser_set_stray_words_value_name(LogParser *s, const gchar *value_name)
 }
 
 static const gchar *
-_get_formatted_key(KVParser *self, const gchar *key)
+_get_formatted_key_with_prefix(KVParser *self, const gchar *key, GString *formatted_key)
+{
+  if (formatted_key->len > 0)
+    g_string_truncate(formatted_key, self->prefix_len);
+  else
+    g_string_assign(formatted_key, self->prefix);
```
furiel: I suggest doing this assignment once in the beginning of _process before the while loop. This function can unconditionally truncate then append without the branching.

furiel: Oh I did not realise for some reason we dropped precalculating the formatted key getter function. I liked the idea, we are doing something very similar ing csvparser.c

```
  GString *key_scratch = scratch_buffers_alloc();
  if (self->prefix)
    g_string_assign(key_scratch, self->prefix);

  key_formatter_t _key_formatter = dispatch_key_formatter(self->prefix);
  while (csv_scanner_scan_next(&scanner))
    {

      log_msg_set_value_by_name(msg,
                                _key_formatter(key_scratch, csv_scanner_get_current_name(&scanner), self->prefix_len),
                                csv_scanner_get_current_value(&scanner),
                                csv_scanner_get_current_value_len(&scanner));
    }
```

@szemere you can ignore my previous comment.

lbudai: @furiel, @szemere: my only problem was that we are storing the formatter in KVParser and the logic around setting up the formatter and prefix/formattes ket was splitted. So the formatter itself can be used as in case of csvparser. I also want to highlight that in a language that supports lambdas it can be expressed in a more elegant way(in our case the signature is ‘polluted’ due to the limitations of C programming language, but this is just a matter of taste whether we use fptrs or another solution).

lbudai:
@bazsi: I think all your requeted changes are addressed, so I'm merge this using my admin priv.

PR #1874, opened 2018-02-05
Refactor and increase the maximum record length of disk-buffer()

MrAnno:
There is a hard-coded limitation in disk-buffer: the length of a record can not be more than 10 mebibyte.
A record consists of multiple fields serialized by `log_msg_serialize()`, so it is just a serialized `LogMessage` object.

The limit is used as a heuristic to detect an incorrect record and to prevent extreme memory usage in case of a corrupt message. Since a record includes structured data (the `NVTable`) and not just a simple serialized message, there is no way to derive an appropriate dynamic limit from the value of the `log-msg-size()` option.

The best would be to remove this limitation, but in that case, a few "consistency guarantees" and/or per-record checksum would be required.


kira-syslogng:

PR #1878, opened 2018-02-07
Do not compile criterion unit tests when criterion missing (autotools + cmake)

furiel:
check target in autotools and test target in cmake depend on the detection of criterion. Prior this patch, unit tests were silently excluded from the test set.

- Autotools
As autotools creates check, recursive-check, local-check targets automatically, the best I could come up with is to add an extra dependency to check target, that triggers configure failure in case testing is disabled.

```
$ make check
Makefile:18487: *** "Unit tests disabled".  Stop.
```
- cmake
In cmake, I could completely disable the test target:
```
$ make test
make: *** No rule to make target 'test'.  Stop.
```
Though in cmake: the test code is compiled together with the code, one important difference to autotools. Meaning we need to expicitely disable compiling unit tests.

The first idea is an early return in add_unit_test() function. Though this is not enough, because some tests do not use this function: those need special configuration for their target: extra cflags, make-clean files, compiler-switch values.

The alternative is to add an early return in add_unit_test(), and decorate those exceptions with BUILD_TESTING. The problem is one does not expect that this BUILD_TESTING needs to be used, because typically we go with add_unit_test().

So in the end I added an add_test_subdirectory, which is a wrapper around add_subdirectory(), this is the intended api to use. add_test_subdirectory() will not go in a directory if BUILD_TESTING is disabled.


kira-syslogng:

furiel:

kira-syslogng:

furiel:

kira-syslogng:

pzoleex:

kira-syslogng:

kira-syslogng:

furiel:

kira-syslogng:

furiel:
@Kokan I'll try to implement it.
make test / make check target should depend on criterion.

kira-syslogng:

bazsi:
I agree completely. I never liked the fact that a missing criterion means
that we omit most of our unit test coverage.


-- 
Bazsi

kira-syslogng:

kira-syslogng:

kira-syslogng:

furiel:
For now we depend on criterion-dev in the control file of deb package generation. I will update the pr accordingly so that deb generation should work even if criterion is not installed.

kira-syslogng:

furiel:
Ok, I removed unit test execution by `+override_dh_auto_test: ;`, see latest patch.

@mochrul I checked the debian packaging files in sid, and I see there is no criterion there either, but the tests are enabled. Meaning this PR will break the packaging for you. You will need to apply a similar change in debian packaging.

@czanik not sure about rpm, but probably you are affected as well. Can you check/follow this PR?

PR #1879, opened 2018-02-09
scratchbuffers: crash in test_context_info_db: missing explicit_gc()

furiel:
The root cause is the same as in: #1872 
As garbage collection is not explicitely called, unit test tries to write an internal message (warning) about possible leak. Though the internal messaging is not initialized, hence there is a crash.
For now I created the same fix as in #1872, calling garbage collection explicitely. Though I am starting to feel this is not the correct way to handle these issues.

Another problem is with test_context_info_db testcase is that scratchbuffers is not used at all by the test code, still the test code needs to deal with the lifecycle of scratchbuffers. It is unfortunate that the test code needs to know the internal mechanics of memory management of the tested module.

I am not sure about the alternatives though.

We can cut the dependency between scratchbuffers and messages, if we create a logger class inside scratchbuffers with a public method that can set the logger to use msg_warning. scratch_buffer_alloc_init might initialize logger to fprintf stderr. In case of unit tests, warning on stderr can be ignored, and will be ignored by the test driver.

Though the problem is the unit test still needs to initialize scratchbuffers, even without using it. It would be great if we could lazily initialize scratchbuffers on the first use, but I am not sure if it is possible without performance hit (without a memory barrier for each scratch_buffers_alloc()).

And the scratch_buffers_allocator_deinit() still needs to be called somewhere.

What are your opinions? Do you feel this is a problem we should deal, or it is fine if unit tests take care of the proper intialization/deinitialization for either or both problems?

### DIFF
--- a/cmake/add_tests.cmake
+++ b/cmake/add_tests.cmake
@@ -37,7 +37,8 @@ function (add_unit_test)
   set_property(TARGET ${ADD_UNIT_TEST_TARGET} APPEND PROPERTY LINK_FLAGS "-Wl,--no-as-needed")
 
   if (${ADD_UNIT_TEST_CRITERION})
-    target_link_libraries(${ADD_UNIT_TEST_TARGET} criterion)
+    target_link_libraries(${ADD_UNIT_TEST_TARGET} ${CRITERION_LIBRARIES})
+    target_include_directories(${ADD_UNIT_TEST_TARGET} PUBLIC ${CRITERION_INCLUDE_DIRS})
     set_property(TARGET ${ADD_UNIT_TEST_TARGET} PROPERTY POSITION_INDEPENDENT_CODE FALSE)
   endif()
 
--- a/lib/value-pairs/tests/Makefile.am
+++ b/lib/value-pairs/tests/Makefile.am
@@ -6,7 +6,8 @@ check_PROGRAMS				+= \
 	${lib_value_pairs_tests_TESTS}
 
 lib_value_pairs_tests_test_value_pairs_LDADD	= $(TEST_LDADD)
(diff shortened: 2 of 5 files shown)


kira-syslogng:

furiel:
@kira-syslogng can you check?

furiel:
Try again.

kira-syslogng:

lbudai:
@furiel: I dont really like that we have to initialize the scratchbuffer in test code without using it. Best would be to talk about IRL where we can discuss the pros/cons of each possible solutions.

lbudai:
@furiel: just call the `app_startup` in tests (as `app_startup` calls the `scratch_buffers_global_init`).

kira-syslogng:

furiel:
@lbudai thanks for help! I aligned the code to use `app_startup()` instead of `scratch_buffers_allocator_init()`.

Added two extra commits: it turned out the criterion in autotools and in cmake did not work if criterion is installed to a custom location. I needed this functionality because these tests hard-fail only with bleeding edge criterion this time of writing: 2.3.2 silently fails on these tests.

lbudai:
@furiel:
What is the root cause of the criterion issue? When a unit test crashes then it doesn't report error code to the caller? Shouldn't we upgrade criterion in our build environments? Which version contains this fix?
@MrAnno do you know about it?

furiel:
@lbudai unfortunately only the bleeding edge in criterion contains the fix.
> When a unit test crashes then it doesn't report error code to the caller?

The problem occurs if unit test crashes during teardown. https://github.com/Snaipe/Criterion/commit/99a2c7ec7261ea1c343f968dc9af657b48e90c2b fixes the issue.

See https://github.com/Snaipe/Criterion/issues/242 for reference.

lbudai:
@furiel : thanks!

PR #1880, opened 2018-02-11
Statistics static assert for module_names

Kokan:
The `module_names` must match the list of `SCS_` items defined in the header.

Should the `static_assert` defined in compat layer instead of not doing the check if not exists ?

### DIFF
--- a/lib/stats/stats-cluster.c
+++ b/lib/stats/stats-cluster.c
@@ -23,6 +23,7 @@
  */
 #include "stats/stats-cluster.h"
 
+#include <assert.h>
 #include <string.h>
 
 static StatsClusterKey *
@@ -79,7 +80,7 @@ stats_cluster_get_type_name(StatsCluster *self, gint type)
 static const gchar *
 _get_module_name(gint source)
 {
-  static const gchar *module_names[SCS_MAX] =
+  static const gchar *module_names[] =
   {
     "none",
     "file",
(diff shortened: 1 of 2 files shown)


kira-syslogng:

furiel:
Unfortunately `stdin` name is still missing from statistics output.
```
@version: 3.13
log { source { stdin();}; }
```
After
```
$ ./syslog-ng-ctl query get "source*"
source.#anon-source0.processed=0
```

Kokan:
@furiel Yes, because it is defined with stats level 1. As a few other source. That made me curious, but seems it was not by mistake.

furiel:
Thanks for explanation, the `stdin` string appears nicely indeed on stats-level 1.

furiel:
Hmmm there is a G_STATIC_ASSERT macro in glib, which could be an (maybe more reliable) alternative.
https://developer.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#G-STATIC-ASSERT:CAPS

Kokan:
Yes. My question actually connects to it. I think because `static_assert` is just a defined `_Static_assert` which is part of the C11 standard, therefore most of the case it is better then a library own solution. The later one is just a macro magic, while th first one must be supported by compiler. The underlying logic behind a macro such as `G_STATIC_ASSERT` always could generate a warning, error that is not trivial to fix, but utterly not relevant.

The only downside of the `static_assert` I could think of - where maybe `G_STATIC_ASSERT` could be better - old compiler that does not support `static_assert`.

I digged gcc 4.6 deep, and it supports it. Travis has gcc 4.8.4. While it could be a risk not to catch those issues with older compiler, Travis would always catch it, and also any compiler installed on almost all of the computer. (even ubuntu 12.04 has 4.6 gcc)

I could introduce a macro in order to patch older compilers and use `G_STATIC_ASSERT`, but I think the risk is so minimal compared to that if I create such code it is never going to away despite not used ever.

furiel:
Just want to confirm that I like either version. Feel free to merge from my point of view.

Kokan:

kira-syslogng:

PR #1881, opened 2018-02-12
Fix python issue R1710

mochrul:
This pylint issue alerts if the return statements
are inconsistent. It also alerts when some return
statement return with value, but at the function end
the return statement is missing.

Signed-off-by: SZALAY Attila <sasa@ubainba.hu>


kira-syslogng:

Kokan:
@kira-syslogng ok to test

kira-syslogng:

mochrul:
Before it has been merged. Please check if all added return None makes sense. I added them because that is the python default (so no behavior change) but sometimes I felt that it just cause no problem because it never hit that code path.

PR #1885, opened 2018-02-15
globalconfig: plugin support for global options

norberttak:
This extends the `ModuleConfig` with automatic module load. A module (like `java`) is able to declare a module specific global option(s).

### DIFF
--- a/modules/java/native/java-grammar.ym
+++ b/modules/java/native/java-grammar.ym
@@ -32,6 +32,7 @@
 
 #include "cfg-parser.h"
 #include "cfg-grammar.h"
+#include "java-config.h"
 #include "plugin.h"
 
 }
@@ -49,6 +50,7 @@
 %token KW_CLASS_PATH
 %token KW_CLASS_NAME
 %token KW_OPTION
+%token KW_JVM_OPTIONS
 
 %%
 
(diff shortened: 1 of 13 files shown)


kira-syslogng:

bazsi:
This looks good to me. there are some whitespace issues (probably your tab size is not 8), so some Makefiles don't indent nicely.

Also, we don't align type/variable names in structs/function declarations in the rest of the codebase.

Anyhow, apart from these nitpicks, LGTM.

kira-syslogng:

kira-syslogng:

modules/java/native/java-grammar.ym
```diff
@@ -91,6 +94,9 @@ java_dest_custom_options
 java_dest_custom_option
   : string string_or_number { java_dd_set_option(last_driver, $1, $2); free($1); free($2); }
 
+java_global_option
+        : KW_JVM_OPTIONS '(' string ')' {java_config_set_jvm_options(configuration, g_strdup($3)); free($3);}
```
MrAnno: Specifying the jvm_options more than once causes a small memory leak (this is a very rare case, but our grammar allows it).
You can call `g_strdup()` in `java_config_set_jvm_options()`. The previous instance can be freed in that method as well.

We have a setter pattern for this, for example:
https://github.com/balabit/syslog-ng/blob/9ffca9f8b4d3402ce2a960db492d2704c4d843e0/modules/xml/xml.c#L269-L277

lbudai:
@MrAnno : OK
+1

Issue #1770, opened 2017-11-21
segfault if disk-buffer file is not deserializable


do3meli:
# syslog-ng

## Version of syslog-ng
```
syslog-ng 3 (3.12.1)
Installer-Version: 3.12.1
Revision: 3.12.1-2
Compile-Date: Oct  6 2017 13:41:52
Module-Directory: /usr/lib/syslog-ng/3.12
Module-Path: /usr/lib/syslog-ng/3.12
Available-Modules: csvparser,kvformat,basicfuncs,linux-kmsg-format,pseudofile,syslogformat,cryptofuncs,cef,xml,disk-buffer,afuser,system-source,dbparser,affile,confgen,date,afprog,sdjournal,afsocket
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: on
```

## Platform
```
Distributor ID:	Debian
Description:	Debian GNU/Linux 8.9 (jessie)
Release:	8.9
Codename:	jessie
```

# Issue

## Failure
syslog-ng 3.12.1 on debian 8.9 segfaults. 
the package was installed from the [unofficial package repo](http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/Debian_8.0)
i have 4 other servers with exact same syslog-ng configuration where the server does not segfault.

the last few log line before the segfault:
```
[2017-11-21T12:04:35.800273] Filter rule evaluation begins; msg='0x7fe3f8003840', rule='f_sftp', location='/etc/syslog-ng/syslog-ng.conf:107:16'
[2017-11-21T12:04:35.800313] Filter regexp node evaluation result; msg='0x7fe3f8003840', input='supervise/syslog-ng', result='not-match'
[2017-11-21T12:04:35.800344] Filter node evaluation result; msg='0x7fe3f8003840', result='not-match', type='regexp'
[2017-11-21T12:04:35.800375] Filter rule evaluation result; msg='0x7fe3f8003840', result='not-match', rule='f_sftp', location='/etc/syslog-ng/syslog-ng.conf:107:16'
[2017-11-21T12:04:35.800408] Filter rule evaluation begins; msg='0x7fe3f8003840', rule='f_err', location='/etc/syslog-ng/syslog-ng.conf:94:15'
[...]

PR #1886, opened 2018-02-15
Diskq: fix restart disk-queue

gaborznagy:
If a pop operation fails during reading from the disk-queue file the restart_corrupted method is called.
It has to close the current disk-queue file and rename it to "<filename>.corrupted" then it has to start a new file from scratch.

The function reuses the existing QDisk object after a deinit call, although it forgets to initialize the object properly before use and that causes a crash.

Questions:
* Who creates the DiskQueueOptions object and where is it passed to the QDisk object?
* Is it okay to get a reference of the `options` member?

Probably fixes crash in #1770 (but not the corruption of the disk-queue).

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -786,6 +786,7 @@ qdisk_start(QDisk *self, const gchar *filename, GQueue *qout, GQueue *qbacklog, 
       self->hdr->write_head = QDISK_RESERVED_SPACE;
       self->hdr->backlog_head = self->hdr->read_head;
       self->hdr->length = 0;
+      self->file_size = self->hdr->write_head;
 
       if (!qdisk_save_state(self, qout, qbacklog, qoverflow))
         {
@@ -855,6 +856,7 @@ void
 qdisk_init(QDisk *self, DiskQueueOptions *options)
 {
   self->fd = -1;
+  self->file_size = 0;
   self->options = options;
   if (!self->options->reliable)
     self->file_id = "SLQF";
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -83,9 +83,6 @@ struct _QDisk
   gchar *filename;
   gchar *file_id;
   gint fd;
-  gint64 read_qout_ofs; /* the size in bytes between read_head and qout_ofs */
-  gint64 prev_read_head;
-  gint64 prev_length;
   gint64 file_size;
   QDiskFileHeader *hdr;
   DiskQueueOptions *options;
@@ -460,7 +457,6 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
   qoverflow_count = self->hdr->qoverflow_count;
   qoverflow_len = self->hdr->qoverflow_len;
   qoverflow_ofs = self->hdr->qoverflow_ofs;
-  self->read_qout_ofs = qout_ofs;
 
   if ((self->hdr->read_head < QDISK_RESERVED_SPACE) ||
       (self->hdr->write_head < QDISK_RESERVED_SPACE) ||
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -967,6 +967,12 @@ qdisk_reset_file_if_possible(QDisk *self)
     }
 }
 
+DiskQueueOptions *
+qdisk_get_options(QDisk *self)
+{
+  return self->options;
+}
+
 gint64
 qdisk_get_length(QDisk *self)
 {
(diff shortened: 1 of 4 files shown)


kira-syslogng:

modules/diskq/tests/test_diskq.c
```diff
@@ -315,6 +317,73 @@ testcase_with_threads(void)
...
+      log_queue_disk_load_queue(q, filename);
+      fed_messages = 0;
+      feed_some_messages(q, 100, &parse_options);
+      assert_gint(fed_messages, 100, "Failed to push all messages to the disk-queue!\n");
+
+      LogQueueDisk *disk_queue = (LogQueueDisk *)q;
+      disk_queue->restart_corrupted(disk_queue);
+
+      struct stat file_stat;
+      stat(filename, &file_stat);
```
furiel: Can you assert for the return value? It would support the Enable warnings PR.

modules/diskq/qdisk.c
```diff
@@ -965,6 +963,12 @@ qdisk_reset_file_if_possible(QDisk *self)
     }
 }
 
+DiskQueueOptions *
+qdisk_get_options_reference(QDisk *self)
```
furiel: From the name I would expect DiskQueueOptions is refcounted and this function increases the reference before returning. I suggest just calling qdisk_get_options.

gaborznagy: I had problem naming this function. :smile:
A "getter" would mean that you receive the value of the attribute and not the address of the attribute.
I am ok with `qdisk_get_options` though.

MrAnno: +1 for `qdisk_get_options`.

MrAnno:
> Who creates the DiskQueueOptions object and where is it passed to the QDisk object?

`DiskQueueOptions` is owned by `DiskQDestPlugin`.
`QDisk` holds a reference that is passed via `qdisk_init` in the constructors of `LogQueueDiskNonReliable` and `LogQueueDiskReliable`.

> Is it okay to get a reference of the options member?

I think it's OK. Otherwise, it wouldn't be easy to get the previous options instance in the middle of `logqueue-disk.c`.

modules/diskq/tests/test_diskq.c
```diff
@@ -339,14 +339,6 @@ typedef struct diskq_tester_parameters
   guint32 qout_size;
 } diskq_tester_parameters_t;
 
-
-typedef struct
-{
-  LogQueue *q;
-  guint32 num;
-  guint32 single_msg_size;
-} feed_and_assert_t;
```
MrAnno: Nice. :+1:

modules/diskq/tests/test_diskq.c
```diff
@@ -315,6 +317,73 @@ testcase_with_threads(void)
...
+                  "Invalid read pointer!\n");
+
+      log_queue_unref(q);
+      unlink(filename);
+      unlink(filename_corrupted_dq);
+      disk_queue_options_destroy(&options);
+    }
+}
+#undef TEST_DISK_BUF_SIZE
+#undef TEST_MEM_SIZE
```
MrAnno: I wouldn't unset these macros.

modules/diskq/tests/test_diskq.c
```diff
@@ -315,6 +317,73 @@ testcase_with_threads(void)
...
+    gboolean reliable;
+  } restart_test_parameters;
+
+  restart_test_parameters parameters[] =
+  {
+    {"test-diskq-restart.qf", log_queue_disk_non_reliable_new, FALSE},
+    {"test-diskq-restart.rqf", log_queue_disk_reliable_new, TRUE},
+  };
+
+  gint number_of_variants = sizeof(parameters)/sizeof(restart_test_parameters);
```
MrAnno: Please use `gsize`.

modules/diskq/tests/test_diskq.c
Kokan: Why not simple use a `const` variable ?

modules/diskq/qdisk.c
gaborznagy: Okay, after our discussion I agree.

modules/diskq/tests/test_diskq.c
gaborznagy: `TEST_MEM_SIZE` is not even used for assertion as checking `disk_buf_size` is good enough.
I prefer macros as constants, but for here the const var is good too.
I have changed it to const var. and moved it inside the function.

kira-syslogng:

gaborznagy:

kira-syslogng:

gaborznagy:

kira-syslogng:

furiel:
If style check is fixed, it is ready to merge.

kira-syslogng:

PR #1887, opened 2018-02-16
Passwd prot ssl keys

lbudai:
# Support password protected SSL keys

## Concept
The solution has three key parts
 * need to store the secret keys somewhere, possible in a non-dumpable memory area (credential store)
 * inject secret into syslog-ng with syslog-ng-ctl (syslog-ng-ctl)
 * afsocket:
   * network source
     * make it possible to disable client connections until a valid password is not provided by the user, and open connections when the password is available
   * network destination
     * try to reconnect to destination until a valid password is not available

#### Network source & destination (afsocket)
In case of TCP network source, we call a `TransportMapper` init as part of the main init. This `TransportMapper::init` is responsible for setting up the `TLSContext`.
In the original implementation, when someone uses a password protected SSL key, OpenSSL will ask for the password (from stdin as we did not set the default password callback) and entire syslog-ng blocks until the password is not set.
One of our goal was to avoid blocking the entire syslog-ng instance in such case.
We can do this by splitting the init logic:
 * first part is responsible for allocating resources (socket and bind system calls)
 * second part is to start accepting client connections(listen & accept system calls)

As `TransportMapper` initializes the `TLSContext`, it has the knowledge whether we can finalize the Source driver init or not, so we added an asynchronous TransportMapper init method.
The async init requires a callback which is responsible for finishing the Source initialization (it calls the finalize init method which then calls the listen&accept).

#### Callback-hell
Finalize callback (which is passed to `TransportMapper::async_init`, `_finalize_init` in `afsocket-source.c`) is not the only one callback we have to use.
OpenSSL uses a password callback on its own: this callback is responsible for filling up a buffer with the password.
This callback (`_pem_passwd_callback` in `tlscontext.c`) has a user data argument which is used for passing the key parameter for secret store in our implementation.
Based on the key we can get the stored password.

The third callback is used with secret store. Secret store handles a kind of hashtable which stores the keys and callbacks. Absolute path of the PEM files are used as keys, and
when the user injects a password for a specific key, the registered callback will be called.

#### How it works together?
* When a tcp source is configured, asynchronous transportmapper init method will be called. When tls is not used, it just call the finalize callback and start accepting client connections.
* When tls is used and the pem file is not protected by password, tls context setup will be created successfully, callback is called immediately, source driver is accepting client connections.
* When tls is used and the pem file is protected by password, tls context setup will fail, another callback(secret-store-callback) is registered(`secret_storage_subscribe_for_key`) in secret-storage with
key = pem file name. Source driver init returns successfully but won't be able to receive client connections as it is not listening.
When user inputs the password for the key, the registered function is called: it will try to setup tls context. The tls setup context method will trigger OpenSSL to call its own password callback(with
userdata=pem file), which will get the password from the secret store. If the password is not valid, we re-register the callback in secret-store (and display error message and change status in secret-storage).
When the password is fine, and the tls context setup is succeeded, we call the callback passed to `TransportMapper::async_init` which then calls listen and accept, so the source driver is ready
for accepting client connections.

#### Sequence diagram: source
network-source

#### Destination
We use the same `TransportMapper::async_init`, the only difference is that the finalize init callback calls the `afsocket_dd_try_connect`.
So we handle invalid/unavailable password as it would be a connection error.

#### syslog-ng-ctl
Operators can use `syslog-ng-ctl` to inject secrets to the secret storage.

* The already stored secrets together with their status can be queried by `syslog-ng-ctl credentials status`.

* Secrets can be added to the secret storage by `syslog-ng-ctl credentials add id secret` command.

From ssl passhprase point of view the `id` in `syslog-ng-ctl` must be the exact string written in the configuration for the location of the encrypted key. If not the exact string is used, the unlock will not work. If full path is used inside the configuration, the same full path must be used in `syslog-ng-ctl`. If relative path is used, the same relative path must be used in `syslog-ng-ctl`. The reason is technical: the subscription happens with the string used inside the configuration, and the same key is used by `syslog-ng-ctl` to store the secret.

This behaviour is a counter-intuitive, but there are plans to resolve it in the future.

For sake of convenience, users can either add credentials with `-i/--id, -s/--secret` options, but also with positional parameters. In case an option is missing, `syslog-ng-ctl` tries to fill the data from the positional parameter. If the options could not be filled either way, error will be reported, and a help text is emitted about the options. For sake of the latter: a little refactor was necessary in syslog-ng-ctl: the `GOptionContext` needs to be passed to the callbacks as well.

`syslog-ng-ctl credentials` is a command that has two subcommands: `add` and `status`. So to support this feature, the description of subcommands is extended: now subcommands can also have subcommands. The command descriptor table therefore is now represented as a tree. To find the appropriate command and callback, the tree needs to be traversed from now on, collecting the nodes in the meantime: `find_active_mode`. Once the correct leaf is found, the callback is invoked with the appropriate arguments.

#### secret-storage
The secret-storage module is responsible for storing the credentials - provided by syslog-ng-ctl -, and callbacks - registered by drivers.

Secret-storage is a hashtable with a special memory allocator. The hashtable is indexed by the keys provided by either `syslog-ng-ctl` or `TransportMapper`. An entry is a secret together with some callbacks and a secret status. The values are stored on a specially allocated memory area, that is locked into memory (not written into swap), and not written to a core file.

There is an ulimit for `RLIMIT_MEMLOCK` (total size of locked memory). This restricts the number of secrets handled by syslog-ng in parallel. The memory allocator can allocate only larger chunks than pagesize, which is typically 4k. The default ulimit for `RLIMIT_MEMOLOCK` is 64k, meaning only 16 secrets can be handled only at a time, depending the size of the secrets, including the secrets cloned internally.

Secret storage is a core functionality, but due to a technical limitation, it is shipped as a separate shared library. The reason is that the symbols of the secret storage is set to hidden visibility by default (which we did not want to change in libsyslog-ng.so), meaning a specially crafted syslog-ng module cannot use secret-storage symbols, unless secret-storage explicitely allows it. Specially: the hash table of secrets is also hidden. As a side effect, users of the secret-storage need to explicitely link to libsecret-storage.so. This feature alone does not add much to security, because for now any driver can query any secrets using `secret_storage_get_secret_by_name`, though there are plans to add some level of authentication for drivers in the future.

Any time a driver works with a secret, secret-storage first clones the secret, and provides only a copy for the driver. As a result, the different drivers cannot accidentally modify the secret for each other.

The status of a secret is initialized to `PENDING` when the first subscription comes from `TransportMapper`. After that only the drivers update the status of the secrets.

There is a function added to loop through the secret status: `secret_store_status_foreach`. This is used by `syslog-ng-ctl` to list the stored secrets with status.

#### nondumpable-allocator
A special memory allocator added to store the secrets in memory. The difference between a malloc-ed memory and a nondumpable-memory is the latter is not written to swap, and not written to core file. During deallocation, the area is zeroed out as well.

The allocator works on mapped anonymous pages, calling `mlock` and `madvise` on the pages. Each allocated memory contains a buffer provided for the user, and some metadata about the allocation: the allocated and available size. The available size is used for zeroing during deallocation.

The layout of buffer and metadata tries to imitate the malloc implementation in glibc: the metadata is stored right before the user buffer in memory: i. e. the allocated anonymous map starts with the metadata, however the pointer returned by the allocator is the location of the user buffer inside the anonymous memory. Hence the deallocator can easily find the metadata by a simple substraction from the location provided by the user.

There is an ulimit for locking memory: `RLIMIT_MEMLOCK`, which is usually 64k. As the allocation happens in pages, usually 4k, the number of parallel allocations is very limited by default.

### Future improvements
 * Use crendential store for SQL , mongodb, and amqp destinations (->any other sources, destinations where credential can be set by users).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,6 +169,12 @@ set(Eventlog_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/lib/eventlog/src")
 
 add_custom_target(style-check  COMMAND ${PROJECT_SOURCE_DIR}/scripts/style-checker.sh check  ${PROJECT_SOURCE_DIR} ${PROJECT_BINARY_DIR})
 add_custom_target(style-format COMMAND ${PROJECT_SOURCE_DIR}/scripts/style-checker.sh format ${PROJECT_SOURCE_DIR} ${PROJECT_BINARY_DIR})
+add_custom_target(check-copyright
+  COMMAND ${PROJECT_SOURCE_DIR}/tests/copyright/check.sh . ${PROJECT_BINARY_DIR} policy
+  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+set_target_properties(check-copyright PROPERTIES
+  ADDITIONAL_MAKE_CLEAN_FILES
+  "copyright-run.log;copyright-err.log")
 
 
 include_directories (${PROJECT_BINARY_DIR}/lib)
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -302,6 +302,7 @@ target_link_libraries(
     ${OPENSSL_LIBRARIES}
     ${RESOLV_LIBRARIES}
     ${LIBPCRE_LIBRARIES}
+    secret-storage
     )
 
 set_target_properties(syslog-ng
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -24,6 +24,7 @@ add_subdirectory(value-pairs)
 add_subdirectory(scanner)
 add_subdirectory(str-repr)
 add_subdirectory(eventlog)
+add_subdirectory(secret-storage)
 
 set(LIB_SUBDIR_HEADERS
     ${COMPAT_HEADERS}
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -18,6 +18,7 @@ include lib/scanner/csv-scanner/Makefile.am
 include lib/scanner/list-scanner/Makefile.am
 include lib/scanner/kv-scanner/Makefile.am
 include lib/str-repr/Makefile.am
+include lib/secret-storage/Makefile.am
 
 LSNG_RELEASE		= $(shell echo @PACKAGE_VERSION@ | cut -d. -f1,2)
 
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -41,7 +41,7 @@ LSNG_REVISION		= 0
 LSNG_AGE		= 0
 
 lib_LTLIBRARIES				+= lib/libsyslog-ng.la
-lib_libsyslog_ng_la_LIBADD		= @CORE_DEPS_LIBS@ $(libsystemd_LIBS)
+lib_libsyslog_ng_la_LIBADD		= @CORE_DEPS_LIBS@ $(libsystemd_LIBS) $(top_builddir)/lib/secret-storage/libsecret-storage.la
 lib_libsyslog_ng_la_LDFLAGS		= -no-undefined -release ${LSNG_RELEASE} \
 					  -version-info ${LSNG_CURRENT}:${LSNG_REVISION}:${LSNG_AGE}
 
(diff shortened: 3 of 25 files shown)


lbudai:
wrong branch...

kira-syslogng:

Issue #1855, opened 2018-01-19
[core-dump] Fatal error when monitoring thousands of files


stg-:
## Version of syslog-ng

```
syslog-ng 3 (3.13.2)
Config version: 3.13
Installer-Version: 3.13.2
Revision: 
Compile-Date: Jan  4 2018 10:38:45
Module-Directory: /syslog-ng/exec/lib/syslog-ng
Module-Path: /syslog-ng/exec/lib/syslog-ng
Available-Modules: add-contextual-data,affile,afprog,afsocket,afsql,afuser,appmodel,basicfuncs,cef,confgen,cryptofuncs,csvparser,date,dbparser,disk-buffer,geoip-plugin,graphite,json-plugin,kvformat,linux-kmsg-format,map-value-pairs,mod-python,pseudofile,sdjournal,snmptrapd-parser,stardate,syslogformat,system-source,tags-parser,tfgetent,xml
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: on
```

## Platform

Ubuntu 16.04 LTS

## Debug bundle

-

# Issue

## Failure

When using wildcard-file with a couple of thousands files the following error appears in logs:

```
iv_fd_unregister: called with fd which is not registered
Aborted (core dumped)
```

gdb info:

```
(gdb) thread apply all bt

Thread 5 (Thread 0x7f536b7cb700 (LWP 2467)):
#0  0x00007f537202c153 in connect () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007f537289bca3 in iv_fd_epoll_timerfd_poll (st=0x7f53540008c0, active=0x7f536b7c8a60, abs=0x118bb18) at iv_fd_epoll.c:436
#2  0x00007f5372899115 in iv_fd_poll_and_run (st=st@entry=0x7f53540008c0, abs=<optimized out>) at iv_fd.c:206
#3  0x00007f537289a0da in iv_main () at iv_main_posix.c:112
#4  0x00007f53728982e3 in iv_work_thread (_thr=0x118bad0) at iv_work.c:185
#5  0x00007f537289b04c in iv_thread_handler (_thr=0x6ff78c0) at iv_thread_posix.c:112
#6  0x00007f53722f56fa in start_thread (arg=0x7f536b7cb700) at pthread_create.c:359
[...]

PR #1889, opened 2018-02-21
wildcard file source: fixing various crashes

furiel:
This patchset fixes 3 independent crashes in wildcard file source.

- in persist state, g_cond_wait was not checked in a while loop, though it is possible to it returns early, without g_cond_signal. This resulted in an assert. I could trigger this with creating and deleting non-empty files in busy loop, while reloading syslog-ng, also in a busy loop.

- When a file is deleted wildcard source deinits the reader in the main thread, but in the meantime it was possible the worker threads try to post message to the pipe element, resulting in an assert. The deinit of the reader is moved to another part of the code: it is handled together with the other reader notifications. Those are synchronized with the pipe post mechanism, eliminating the crash.

- Missing check for directory monitoring creation results in a crash. In case too many files were open, due to missing error checking, syslog-ng thought the watches are started successfully. During deinit, syslog-ng tries to stop the watches, resulting in an abort in ivykis about the not tracked fd to be untracked. Only "inotify" version of polling is affected.

Fixes: https://github.com/balabit/syslog-ng/issues/1855

### DIFF
--- a/modules/affile/directory-monitor-inotify.c
+++ b/modules/affile/directory-monitor-inotify.c
@@ -102,10 +102,17 @@ directory_monitor_inotify_new(const gchar *dir, guint recheck_time)
   DirectoryMonitorInotify *self = g_new0(DirectoryMonitorInotify, 1);
   directory_monitor_init_instance(&self->super, dir, recheck_time);
 
+  IV_INOTIFY_INIT(&self->inotify);
+  if (iv_inotify_register(&self->inotify))
+    {
+      msg_error("directory-monitor-inotify: could not create inotify object", evt_tag_errno("errno", errno));
+      directory_monitor_free(&self->super);
+      return NULL;
+    }
+
   self->super.start_watches = _start_watches;
   self->super.stop_watches = _stop_watches;
   self->super.free_fn = _free;
-  IV_INOTIFY_INIT(&self->inotify);
-  iv_inotify_register(&self->inotify);
+
   return &self->super;
 }
--- a/modules/affile/wildcard-source.c
+++ b/modules/affile/wildcard-source.c
@@ -24,6 +24,7 @@
 #include "directory-monitor-factory.h"
 #include "messages.h"
 #include "file-specializations.h"
+#include "mainloop.h"
 
 #include <fcntl.h>
 
@@ -50,6 +51,13 @@ _check_required_options(WildcardSourceDriver *self)
   return TRUE;
 }
 
+static void
+_deleted_cb(FileReader *self, gpointer user_data G_GNUC_UNUSED)
+{
+  log_pipe_deinit(&self->super);
+  file_reader_remove_persist_state(self);
+}
+
 void
 _create_file_reader(WildcardSourceDriver *self, const gchar *full_path)
 {
@@ -67,6 +75,7 @@ _create_file_reader(WildcardSourceDriver *self, const gchar *full_path)
 
   reader = file_reader_new(full_path, &self->file_reader_options, self->file_opener, &self->super, cfg);
   log_pipe_append(&reader->super, &self->super.super.super);
+  reader->missing_cb = _deleted_cb;
   if (!log_pipe_init(&reader->super))
     {
       msg_warning("File reader initialization failed", evt_tag_str("filename", full_path),
@@ -89,6 +98,7 @@ _handle_file_created(WildcardSourceDriver *self, const DirectoryMonitorEvent *ev
       if (!reader)
         {
           _create_file_reader(self, event->full_path);
+          msg_debug("Wildcard: file created", evt_tag_str("filename", event->full_path));
         }
       else
         {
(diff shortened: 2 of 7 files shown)


kira-syslogng:

kira-syslogng:

modules/affile/directory-monitor-inotify.c
```diff
@@ -106,6 +106,11 @@ directory_monitor_inotify_new(const gchar *dir, guint recheck_time)
   self->super.stop_watches = _stop_watches;
   self->super.free_fn = _free;
   IV_INOTIFY_INIT(&self->inotify);
-  iv_inotify_register(&self->inotify);
+  if (iv_inotify_register(&self->inotify))
+    {
+      msg_error("directory-monitor-inotify: could not create inotify object", evt_tag_errno("errno", errno));
+      g_free(self);
```
Kokan: In the `directory_monitor_init_instance` the `dir` is `strdup`-ed, so in this case we leak that memory.
I would advice to call the free method, but the `_free` actually also do not free the `dir`, and also other possible members because parent free is not called.

furiel: This surprised me as well: the reason why the parent free is not called from _free of DirectoryMonitorInotify is: she passes her free method to her parent as a virtual function. For me this is confusing: this latter suggests a plugin style relationship between DirectoryMonitor and DirectoryMonitorInotify, though the implementation of directory_monitor_inotify_new and struct DirectoryMonitorInotify suggests inheritance kind of relationship. This should be cleared first, because otherwise we cannot decide whose responsibility to free the DirectoryMonitor if the constructor of DirectoryMonitorInotify fails.

Probably we need to get rid of free_fn, and create proper init/deinit/new/free functions.

kira-syslogng:

modules/affile/directory-monitor-inotify.c
furiel: In the end I overcomplicated the situation. There is directory_monitor_free() which could be called instead of g_free(). Until that part of the constructor, only the parent class is initialized, so we can simply call the parent's destructor. I updated the patchset.

lbudai:
Problem: 
### When the file is deleted, the remaining messages will be dropped.

#### To solve this issue maybe we should keep a state inside the file reader.
This state should store two events:
 * deleted
 * EOF

#### Why we need to deal with two cases?

* case a) file deleted, EOF is not reached
When the file is deleted, we just store it (or "mark" the reader as 'deleted').
When the reader reaches EOF, it can be deinited if marked as deleted if the file does not exist.

* case b) file deleted after EOF is reached
When the delete event received in EOF state, just call deinit.

I'm afraid that we need to use atomic counters to track these states.

_n.b.: this is not a suggestion how to fix the issue, just wanted to highlight the problem_

kira-syslogng:

kira-syslogng:

kira-syslogng:
success

kira-syslogng:

furiel:

kira-syslogng:
failure

kira-syslogng:

kira-syslogng:

kira-syslogng:

kira-syslogng:
success

furiel:
In the end I could make all tests pass, and all kind of crashes disappeared that I knew of. I propose for review.

About the last patch, I would like to add a little information/explanation, especially because I could not make it right for the first time.

Wildcard file source creates `follow-mode` file readers, meaning all events (like there is data available, file truncated etc) are polled periodically (the `monitor-method("inotify")` option in wildcard filesource has **no** effect on this). One interesting property of such FileReader is that it is attached to a `filename` instead of `fd`. Specially, at the EOF notification the poll_file_changes_check_file() does not stop, but  also checks if the followed-file exists with the same inode. If not (for example when deleted and recreated between two polls), simply reopens herself, and continues to follow the same filename, but with different fd: in the end the same FileReader object is reused.

There was a conflict due to this in a previous version: when a file is deleted, I scheduled the deinit method as an ivykis task, which to be executed after we reach EOF. But in the case above, the same FileReader object could be reused if the deletion-creation happens quickly between two polls. And I also had a deinit method scheduled for the same object, that should have left alone. In the end, the file reader deinitialized, messages were not read from the file anymore, hence the tests failed.

Without better idea in mind, I attached the deinit_task to FileReader, and in case this quick delete-create happens, I cancel the task.

Any suggestions for improvement, especially for this last patch is appreciated.

kira-syslogng:
success

furiel:
After discussing with @juhaszviktor, I could ridiculously simplify the patchset. In the end, there is a new type of file notification for follow-type file readers: nc_file_missing, that will trigger the FileReader deinit in logpipe. No boolean flags for eof or deleted or ivykis task was necessary in the end.

modules/affile/wildcard-source.c
```diff
@@ -208,6 +215,13 @@ _add_directory_monitor(WildcardSourceDriver *self, const gchar *directory)
     .method = self->monitor_method
   };
   DirectoryMonitor *monitor = create_directory_monitor(&options);
+  if (!monitor)
```
Kokan: The `_add_directory_monitor` is also called from the `_init`, thus the wildcard file source could be initialized even if there is no `monitor` object exits, thus does nothing.

kira-syslogng:
success

modules/affile/wildcard-source.c
furiel: Thanks for comment! Unfortunately I do not know how trigger, but I added a new patch that should resolve it.

PR #1890, opened 2018-02-21
Fix stack overflow when an internal message is too long (eventlog cleanup)

MrAnno:
This PR
- fixes the symmetry between debug messages in `LogReader` and `LogWriter`
- fixes a possible stack overflow in `eventlog`
- fixes the reallocation method of `EVTSTR`
- removes unused code, outdated documents from `eventlog`

Questions:
1. You may have noticed the code duplication `evtstr.c`. I didn't make extra effort to refactor my code, because those functions are not used at all. Can I remove the `xmlattr` and `xmltag` support from eventlog? **Removed**
2. There are a lot of other things that are not used in eventlog. Should I remove them as well? **Removed**


kira-syslogng:

lib/eventlog/src/evtstr.c
```diff
@@ -89,105 +90,135 @@ evt_str_append_escape_bs(EVTSTR *es,
                          char escape_char)
 {
   /* a single character is escaped to at most 4 characters: \xXX */
+  size_t escaped_char_max_length = 4;
 
-  char *buf = (char *)alloca(4*unescaped_len + 1);
+  char escaped_buffer[128];
+  size_t escaped_buffer_capacity = sizeof(escaped_buffer);
```
Kokan: Please use `sizeof(escaped_buffer)/sizeof(escaped_buffer[0])`

lib/eventlog/src/evtstr.c
```diff
@@ -89,105 +90,135 @@ evt_str_append_escape_bs(EVTSTR *es,
...
+          escaped_buffer[escaped_buffer_length++] = '\\';
+          escaped_buffer[escaped_buffer_length++] = escape_char;
         }
       else
         {
-          buf[dst++] = unescaped[i];
+          escaped_buffer[escaped_buffer_length++] = unescaped[i];
+        }
+
+      size_t escaped_buffer_remaining_space = escaped_buffer_capacity - escaped_buffer_length;
```
Kokan: I would be cautious with `size_t - size_t` because `size_t` is `unsigned`.
It is possible that in the current iteration there is enough data, but in the next iteration it is possible that there isn't, and it is going to write possible uninitialized memory.
I would check if there is enough memory to write the next data (estimate the next data with max possible length: 4), and if not append to the buffer.

Kokan:
Sorry I've jumped right into the diff page and did not see the questions.
The `eventlog` is not an external library now, so I think unused code could be removed.

lib/eventlog/src/evtstr.c
MrAnno: Right, thanks.

lib/eventlog/src/evtstr.c
MrAnno: If I understand it correctly, the condition below does the same thing you asked for.
The condition checks the availability for the next iteration, not for the current one.

```
if (escaped_buffer_remaining_space < escaped_char_max_length)
```

Kokan: My bad. You are right about it. It is correct.

MrAnno: I'll rephrase the condition as we agreed IRL, it's more "natural" this way:

```
 if (escaped_buffer_capacity < escaped_buffer_length + escaped_char_max_length)
```

lib/eventlog/src/evtstr.c
MrAnno: Fixed.

lib/eventlog/src/evtstr.c
MrAnno: Done.

kira-syslogng:

kira-syslogng:

MrAnno:

kira-syslogng:

kira-syslogng:

PR #1898, opened 2018-02-28
Filter logging improvements

bazsi:
This branch dramatically improves readability of the logs generated by filter/parser evaluation, which makes it much easier to troubleshoot configuration issues.

There are a couple of unrelated fixes as well.

NOTE: I am splitting up a larger branch into smaller, more manageably pieces. This is the first part of that series.

### DIFF
--- a/lib/filter/filter-call.c
+++ b/lib/filter/filter-call.c
@@ -49,6 +49,11 @@ filter_call_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg)
   else
     stats_counter_inc(self->super.not_matched);
 
+  msg_debug("  filter() evaluation result",
+            filter_result_tag(res),
+            evt_tag_str("called-rule", self->rule),
+            evt_tag_printf("msg", "%p", msgs[num_msg - 1]));
+
   return res ^ s->comp;
 }
 
--- a/lib/filter/filter-cmp.c
+++ b/lib/filter/filter-cmp.c
@@ -85,6 +85,13 @@ fop_cmp_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg)
       result = self->cmp_op & FCMP_GT || self->cmp_op == 0;
     }
 
+  msg_debug("  cmp() evaluation result",
+            filter_result_tag(result),
+            evt_tag_str("left", left_buf->str),
+            evt_tag_str("operator", self->super.type),
+            evt_tag_str("right", right_buf->str),
+            evt_tag_printf("msg", "%p", msgs[num_msg - 1]));
+
   scratch_buffers_reclaim_marked(marker);
   return result ^ s->comp;
 }
--- a/lib/filter/filter-expr.c
+++ b/lib/filter/filter-expr.c
@@ -49,10 +49,6 @@ filter_expr_eval_with_context(FilterExprNode *self, LogMessage **msg, gint num_m
   g_assert(num_msg > 0);
 
   res = self->eval(self, msg, num_msg);
-  msg_debug("Filter node evaluation result",
-            evt_tag_printf("msg", "%p", *msg),
-            evt_tag_str("result", res ? "match" : "not-match"),
-            evt_tag_str("type", self->type));
   return res;
 }
 
@@ -96,4 +92,10 @@ filter_expr_unref(FilterExprNode *self)
         self->free_fn(self);
       g_free(self);
     }
+}
+
+EVTTAG *
+filter_result_tag(gboolean res)
+{
+  return evt_tag_str("result", res ? "MATCH" : "UNMATCHED");
 }
--- a/lib/logmatcher.c
+++ b/lib/logmatcher.c
@@ -38,10 +38,24 @@ _shall_set_values_indirectly(NVHandle value_handle)
 }
 
 static void
+log_matcher_store_pattern(LogMatcher *self, const gchar *pattern)
+{
+  g_free(self->pattern);
+  self->pattern = g_strdup(pattern);
+}
+
+static void
+log_matcher_free_method(LogMatcher *self)
+{
+  g_free(self->pattern);
+}
+
+static void
 log_matcher_init(LogMatcher *self, const LogMatcherOptions *options)
 {
   self->ref_cnt = 1;
   self->flags = options->flags;
+  self->free_fn = log_matcher_free_method;
 }
 
 typedef struct _LogMatcherPosixRe
@@ -59,6 +73,7 @@ log_matcher_posix_re_compile(LogMatcher *s, const gchar *re, GError **error)
   gint flags = REG_EXTENDED;
 
   g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+  log_matcher_store_pattern(s, re);
 
   if (re[0] == '(' && re[1] == '?')
     {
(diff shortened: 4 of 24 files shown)


kira-syslogng:

norberttak:
Hi Bazsi,
Regarding your commit https://github.com/balabit/syslog-ng/pull/1898/commits/710bb80ed670fced33c44f4f9a5712bff57083e1:

I updated the jsonc version to 0.13 which resolves the error. This is handled in PR #1900

lib/filter/filter-call.c
```diff
@@ -49,6 +49,11 @@ filter_call_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg)
   else
     stats_counter_inc(self->super.not_matched);
 
+  msg_debug("  filter() evaluation result",
+            filter_result_tag(res),
+            evt_tag_str("called-rule", self->rule),
+            evt_tag_printf("msg", "%p", msgs[num_msg - 1]));
```
furiel: This is a little bit confusing for me. Filter call eval works on a list of msgs. Should we build list of evt_tags for all messages here? Is it enough to consider the typical scenario when filter_call_eval is called only for one message?

lib/filter/filter-cmp.c
```diff
@@ -85,6 +85,13 @@ fop_cmp_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg)
       result = self->cmp_op & FCMP_GT || self->cmp_op == 0;
     }
 
+  msg_debug("  cmp() evaluation result",
+            filter_result_tag(result),
+            evt_tag_str("left", left_buf->str),
+            evt_tag_str("operator", self->super.type),
+            evt_tag_str("right", right_buf->str),
+            evt_tag_printf("msg", "%p", msgs[num_msg - 1]));
```
furiel: Same question as above.

lib/filter/filter-expr.c
```diff
@@ -97,3 +93,9 @@ filter_expr_unref(FilterExprNode *self)
       g_free(self);
     }
 }
+
+EVTTAG *
+filter_result_tag(gboolean res)
```
furiel: I suggest following the naming convention here: evt_tag_filter_result

modules/json/json-parser.c
```diff
@@ -229,7 +229,7 @@ json_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_
   jso = json_tokener_parse_ex(tok, input, input_len);
   if (tok->err != json_tokener_success || !jso)
     {
-      msg_error("Unparsable JSON stream encountered",
+      msg_debug("Unparsable JSON stream encountered",
```
furiel: What is the motivation behind this change?

bazsi:
On Mar 1, 2018 04:21, "furiel" <notifications@github.com> wrote:

*@furiel* commented on this pull request.

In lib/logmatcher.c
<https://github.com/balabit/syslog-ng/pull/1898#discussion_r171493703>:

>
-  self->pattern = g_strdup(pattern);
-  self->pattern_len = strlen(self->pattern);
+  self->pattern_len = strlen(pattern);

This could go into log_matcher_store_pattern.


Makes sense.

In lib/Makefile.am
<https://github.com/balabit/syslog-ng/pull/1898#discussion_r171494479>:

> @@ -80,7 +80,7 @@ endif

 if JSON_INTERNAL
 lib/jsonc/libjson-c.la:
-	${MAKE} -C lib/jsonc
+	${MAKE} CFLAGS="${CFLAGS} -Wno-error" -C lib/jsonc

With #1900 <https://github.com/balabit/syslog-ng/pull/1900> this should be
removed.


I am not convinced we should error out on warnings in dependant libraries.

In lib/filter/filter-call.c
<https://github.com/balabit/syslog-ng/pull/1898#discussion_r171495429>:

> @@ -49,6 +49,11 @@ filter_call_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg)
   else
     stats_counter_inc(self->super.not_matched);

+  msg_debug("  filter() evaluation result",
+            filter_result_tag(res),
+            evt_tag_str("called-rule", self->rule),
+            evt_tag_printf("msg", "%p", msgs[num_msg - 1]));

This is a little bit confusing for me. Filter call eval works on a list of
msgs. Should we build list of evt_tags for all messages here? Is it enough
to consider the typical scenario when filter_call_eval is called only for
one message?


Good question. Filters are generally applied to just the last message of a
context, except when they use templates (e.g. comparison). In hindsight
this was probably a bad idea, because it complicates the whole filter stuff
because of a cornercase.

So the message context is only used when we evaluate templates and template
functions.

All the rest works as if there was only one message in the context.


In lib/filter/filter-cmp.c
<https://github.com/balabit/syslog-ng/pull/1898#discussion_r171495681>:

> @@ -85,6 +85,13 @@ fop_cmp_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg)
       result = self->cmp_op & FCMP_GT || self->cmp_op == 0;
     }

+  msg_debug("  cmp() evaluation result",
+            filter_result_tag(result),
+            evt_tag_str("left", left_buf->str),
+            evt_tag_str("operator", self->super.type),
+            evt_tag_str("right", right_buf->str),
+            evt_tag_printf("msg", "%p", msgs[num_msg - 1]));

Same question as above.

In lib/filter/filter-expr.c
<https://github.com/balabit/syslog-ng/pull/1898#discussion_r171495983>:

> @@ -97,3 +93,9 @@ filter_expr_unref(FilterExprNode *self)
       g_free(self);
     }
 }
+
+EVTTAG *
+filter_result_tag(gboolean res)

I suggest following the naming convention here: evt_tag_filter_result


Well, that poisons the evt namespace, and this tag is not very generic,
pretty tightly coupled with the filter expressions.

 We already have similar tags, such as log_expr_location_tag() or something.

In modules/json/json-parser.c
<https://github.com/balabit/syslog-ng/pull/1898#discussion_r171499939>:

> @@ -229,7 +229,7 @@ json_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_
   jso = json_tokener_parse_ex(tok, input, input_len);
   if (tok->err != json_tokener_success || !jso)
     {
-      msg_error("Unparsable JSON stream encountered",
+      msg_debug("Unparsable JSON stream encountered",

What is the motivation behind this change?


This is a per message error, that can easily be used to amplify one's input
size. Parse errors are generally not reported during operation.

—
You are receiving this because you authored the thread.
<https://github.com/balabit/syslog-ng/pull/1898#pullrequestreview-100352439>,
or mute the thread

lib/logmatcher.c
```diff
@@ -246,9 +261,9 @@ log_matcher_string_compile(LogMatcher *s, const gchar *pattern, GError **error)
   LogMatcherString *self = (LogMatcherString *) s;
 
   g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+  log_matcher_store_pattern(s, pattern);
 
-  self->pattern = g_strdup(pattern);
-  self->pattern_len = strlen(self->pattern);
+  self->pattern_len = strlen(pattern);
```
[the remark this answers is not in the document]
bazsi: This can't go to store_pattern as that is a base class which does not have a pattern_len member.

the pattern_len is only needed for the string matching and not regexp ones.

bazsi:
I know my answers were difficult to follow because I answered them in email, but here's a summary.

1) the store_pattern idea was good but does not apply
2) I would add the -Wno-error for jsonc, I don't think we want to break builds for weird gcc/jsonc combinations
3) I've explained the rest of the comments, I think.

So if you agree, please approve.

Bazsi

bazsi:
Thanks for the review, I would merge this if there are no objections by
tomorrow.

PR #1902, opened 2018-03-01
cmake: fixing different naming of getent between autotools and cmake

furiel:
In autotools, the getent module is called tfgetent. In cmake, it is
called just getent. This fails the getent-related e2e tests in the cmake version. 

The cmake version is aligned to the autotools version.


kira-syslogng:

PR #1905, opened 2018-03-02
afsocket: port basic client side failover functionality

lbudai:

### DIFF
--- a/modules/afsocket/afinet-dest.c
+++ b/modules/afsocket/afinet-dest.c
@@ -118,6 +118,43 @@ afinet_dd_set_tls_context(LogDriver *s, TLSContext *tls_context)
                                         afinet_dd_verify_callback, self);
 }
 
+void
+afinet_dd_add_failovers(LogDriver *s, GList *failovers)
+{
+  AFInetDestDriver *self = (AFInetDestDriver *)s;
+  if (!self->server_candidates)
+    {
+      self->server_candidates = g_list_append(self->server_candidates, self->hostname);
+    }
+  while (failovers)
+    {
+      self->server_candidates = g_list_append(self->server_candidates, g_strdup(failovers->data));
+      g_free(failovers->data);
+      failovers = g_list_delete_link(failovers, failovers);
+    }
+  self->server_candidates = g_list_first(self->server_candidates);
+  self->hostname = self->server_candidates->data;
+}
+
+static gchar *
+_get_next_destination_candidate(AFInetDestDriver *self)
+{
+  gchar *candidate = self->hostname;
+  if (self->server_candidates)
+    {
+      if (self->server_candidates->next)
+        {
+          self->server_candidates = self->server_candidates->next;
+        }
+      else
+        {
+          self->server_candidates = g_list_first(self->server_candidates);
+        }
+      candidate = (gchar *)self->server_candidates->data;
+    }
+  return candidate;
+}
+
 static LogWriter *
 afinet_dd_construct_writer(AFSocketDestDriver *s)
 {
@@ -152,6 +189,23 @@ _determine_port(const AFInetDestDriver *self)
   return port;
 }
 
+static void
+_setup_next_hostname(AFInetDestDriver *self)
+{
+  gchar *next_dst_candidate = _get_next_destination_candidate(self);
+  if (strcmp(next_dst_candidate, self->hostname) == 0)
+    {
+      msg_warning("The server is inaccessible, trying the old host again", evt_tag_str("host", self->hostname));
+    }
+  else
+    {
+      msg_warning("Current server is inaccessible, sending the messages to the next failover server",
+                  evt_tag_str("current", self->hostname),
+                  evt_tag_str("failover", next_dst_candidate));
+      self->hostname = next_dst_candidate;
+    }
+}
+
 static gboolean
 afinet_dd_setup_addresses(AFSocketDestDriver *s)
 {
@@ -169,6 +223,8 @@ afinet_dd_setup_addresses(AFSocketDestDriver *s)
   if (self->bind_port)
     g_sockaddr_set_port(self->super.bind_addr, afinet_lookup_service(self->super.transport_mapper, self->bind_port));
 
+
+  _setup_next_hostname(self);
   if (!resolve_hostname_to_sockaddr(&self->super.dest_addr, self->super.transport_mapper->address_family, self->hostname))
     return FALSE;
 
@@ -414,12 +470,25 @@ afinet_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options,
   log_dest_driver_queue_method(s, msg, path_options, user_data);
 }
 
+static void
+_free_server_candidates(AFInetDestDriver *self)
+{
+  while (self->server_candidates)
+    {
+      g_free(self->server_candidates->data);
+      self->server_candidates = g_list_delete_link(self->server_candidates, self->server_candidates);
+    }
+}
+
 void
 afinet_dd_free(LogPipe *s)
 {
   AFInetDestDriver *self = (AFInetDestDriver *) s;
 
-  g_free(self->hostname);
+  if (self->server_candidates)
+    _free_server_candidates(self);
+  else
+    g_free(self->hostname);
   g_free(self->bind_ip);
   g_free(self->bind_port);
   g_free(self->dest_port);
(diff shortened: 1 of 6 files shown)


kira-syslogng:

modules/afsocket/afinet-dest.c
```diff
@@ -118,6 +118,43 @@ afinet_dd_set_tls_context(LogDriver *s, TLSContext *tls_context)
...
 
+void
+afinet_dd_add_failovers(LogDriver *s, GList *failovers)
+{
+  AFInetDestDriver *self = (AFInetDestDriver *)s;
+  if (!self->server_candidates)
+    {
+      self->server_candidates = g_list_append(self->server_candidates, self->hostname);
+    }
+  while (failovers)
```
furiel: Can we use g_list_concat here?

modules/afsocket/afinet-dest.c
```diff
@@ -118,6 +118,43 @@ afinet_dd_set_tls_context(LogDriver *s, TLSContext *tls_context)
...
+    {
+      self->server_candidates = g_list_append(self->server_candidates, g_strdup(failovers->data));
+      g_free(failovers->data);
```
furiel: If g_list_concat is not appropriate, I would still remove this g_strdup g_free combination. We could simply pass the reference to the other list.

modules/afsocket/afinet-dest.c
```diff
@@ -118,6 +118,43 @@ afinet_dd_set_tls_context(LogDriver *s, TLSContext *tls_context)
...
+
+static gchar *
+_get_next_destination_candidate(AFInetDestDriver *self)
+{
+  gchar *candidate = self->hostname;
+  if (self->server_candidates)
+    {
+      if (self->server_candidates->next)
+        {
+          self->server_candidates = self->server_candidates->next;
```
furiel: g_list_next can be used here

modules/afsocket/afinet-dest.c
```diff
@@ -118,6 +118,43 @@ afinet_dd_set_tls_context(LogDriver *s, TLSContext *tls_context)
...
+      failovers = g_list_delete_link(failovers, failovers);
+    }
+  self->server_candidates = g_list_first(self->server_candidates);
+  self->hostname = self->server_candidates->data;
+}
+
+static gchar *
+_get_next_destination_candidate(AFInetDestDriver *self)
+{
+  gchar *candidate = self->hostname;
```
furiel: Just a suggestion but for me it would be a little bit clearer to early return with self->hostname in case there are no failover servers.
```
if (!self->server_candidates)
   return self->hostname

...
```

modules/afsocket/afinet-dest.c
```diff
@@ -152,6 +189,23 @@ _determine_port(const AFInetDestDriver *self)
...
+  gchar *next_dst_candidate = _get_next_destination_candidate(self);
+  if (strcmp(next_dst_candidate, self->hostname) == 0)
+    {
+      msg_warning("The server is inaccessible, trying the old host again", evt_tag_str("host", self->hostname));
+    }
+  else
+    {
+      msg_warning("Current server is inaccessible, sending the messages to the next failover server",
+                  evt_tag_str("current", self->hostname),
+                  evt_tag_str("failover", next_dst_candidate));
```
furiel: Would it make sense to print the log_expr_node_location_tag as well? Maybe multiple writers can share the same failover list. (well in that case probably both will fail on this specific candidate, so again this is just a suggestion)

modules/afsocket/afinet-dest.c
```diff
@@ -414,12 +470,25 @@ afinet_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options,
   log_dest_driver_queue_method(s, msg, path_options, user_data);
 }
 
+static void
+_free_server_candidates(AFInetDestDriver *self)
+{
+  while (self->server_candidates)
+    {
+      g_free(self->server_candidates->data);
```
furiel: We could use g_list_free_full here.

kira-syslogng:

modules/afsocket/afinet-dest.c
```diff
@@ -152,6 +189,23 @@ _determine_port(const AFInetDestDriver *self)
   return port;
 }
 
+static void
+_setup_next_hostname(AFInetDestDriver *self)
+{
+  gchar *next_dst_candidate = _get_next_destination_candidate(self);
+  if (strcmp(next_dst_candidate, self->hostname) == 0)
+    {
+      msg_warning("The server is inaccessible, trying the old host again", evt_tag_str("host", self->hostname));
```
furiel: Can we unconditionally write warning here? Maybe I miss something, but because this function is called in afinet_dd_setup_addresses() unconditionally: in the very first attempt, of when there are no candidates at all, even when there was no error, we hit this line. In those cases this warning is false positive.

modules/afsocket/afinet-dest.c
```diff
@@ -414,12 +470,25 @@ afinet_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options,
...
+    }
+}
+
 void
 afinet_dd_free(LogPipe *s)
 {
   AFInetDestDriver *self = (AFInetDestDriver *) s;
 
-  g_free(self->hostname);
+  if (self->server_candidates)
```
furiel: with g_list_free full, this if is probably not needed. I ran through the latest greatest glib code, and as I see it handles null well.

lbudai: @furiel: thanks for comments
( keep in mind that I just ported what I found in another version and will fix only on next week the comments :) )

modules/afsocket/afsocket-parser.c
```diff
@@ -82,6 +82,7 @@ static CfgLexerKeyword afsocket_keywords[] =
   { "listen_backlog",     KW_LISTEN_BACKLOG },
   { "keep_alive",         KW_KEEP_ALIVE },
   { "systemd_syslog",     KW_SYSTEMD_SYSLOG  },
+  { "failover_servers",   KW_FAILOVERS },
   { NULL }
```
bazsi: please match the KW_XXX define with the actual keyword, so it should be KW_FAILOVER_SERVERS

modules/afsocket/afinet-dest.c
bazsi: the string_list rule in the grammar generates a GList with the data member pointing to a g_strdup()-ed string. So concatentating the two lists should just work.

modules/afsocket/afinet-dest.c
```diff
@@ -118,6 +118,43 @@ afinet_dd_set_tls_context(LogDriver *s, TLSContext *tls_context)
...
+      if (self->server_candidates->next)
+        {
+          self->server_candidates = self->server_candidates->next;
+        }
+      else
+        {
+          self->server_candidates = g_list_first(self->server_candidates);
+        }
+      candidate = (gchar *)self->server_candidates->data;
+    }
```
bazsi: I would probably not mutate the server_candidate list to track the current server, why don't we have a separate current_server_candidate pointer pointing the current element. I think that would make things simpler to read.

modules/afsocket/afinet-dest.c
bazsi: I don't understand this message. Does it get logged when we tried all servers in our list? Because if that's the case the log message should be clarified.

modules/afsocket/afinet-dest.c
bazsi: both the id and the location would be useful, I think, not just here but in the rest of the messages.

modules/afsocket/afinet-dest.c
lbudai: @bazsi: you are the original author of the log message :)
But yes, agree, list handling and other log messages, could be clarified. This is just the very-first stage of the port (adopt the functionality from pe and let's make it work). After this step it is possible to extend zts testcases by failover tests. My plan is that after all the tests passed (and have a green set of testcases) I'll do the refactors.

bazsi: I am absolutely positive that I didn't write that code myself. I may have ported it though :)

failover-servers was done either by @juhaszviktor or Attila Nagy, can't remember exactly.

lbudai: :)

kira-syslogng:
success

kira-syslogng:
failure

lbudai:
@kira-syslogng : retest this, please

pzoleex:
@kira-syslogng : retest this, please

kira-syslogng:
failure

pzoleex:
@kira-syslogng : retest this, please

kira-syslogng:
failure

pzoleex:
@kira-syslogng : retest this, please

kira-syslogng:
failure

kira-syslogng:
success

PR #1913, opened 2018-03-03
python3 compatibility fixes, virtualenv support & fix for the interactive debugger

bazsi:
This is an attempt to fix Python3 compatibility and includes a number of related fixes. Please check individual patches for a summary.


kira-syslogng:

kira-syslogng:

kira-syslogng:

bazsi:
Rebased against latest master and against my dbld-login-and-image-targets branch.

This branch contains all my efforts to make syslog-ng Python3 compatible again and make sure it stays that way :)

I know it does have some unrelated commits (like the dbld related ones) and a geoip fix. Let me know if I should separate them out.

@szemere please look at the dbld changes, I've moved requirements.txt back to the root, as I would like it to be independent of the docker stuff and make it easy to use even outside docker. Instead of Dockerfiles doing the heavy lifting I have added a target to the Makefile to bootstrap the venv and then use that from travis, docker and local builds. I've also added a bootstrap target, which I've felt would be useful for initial setup.

kira-syslogng:

kira-syslogng:
success

modules/geoip/geoip-parser.c
```diff
@@ -127,9 +127,7 @@ geoip_parser_process(LogParser *s, LogMessage **pmsg,
   GeoIPParser *self = (GeoIPParser *) s;
   LogMessage *msg = log_msg_make_writable(pmsg, path_options);
 
-  if (!self->dest.country_code &&
-      !self->dest.latitude &&
-      !self->dest.longitude)
+  if (!self->add_geoip_result)
```
furiel: Can you extract this to a function or a local variable? Like `is_initialized(GeoIPParser *self)` or `gboolean is_initialized = self->add_geoip_result != 0`? It has a taste like a magic number for me.

In geoip_parser_init(), there is:
```
  if (!self->gi)
    return FALSE;
```
which should fail the initialization if the database is not available. Theoretically we should not get to the _process method at all. What are the reproduction steps of the crash?

dbld/rules
```diff
@@ -16,6 +16,7 @@ RELEASE_DIR=$(DBLD_DIR)/release
 VERSION=$(shell cat VERSION)
 TARBALL=$(BUILD_DIR)/syslog-ng-$(VERSION).tar.gz
 MODE=snapshot
+CONFIGURE_OPTS=--enable-manpages --with-python=3.5 --prefix=/install --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl
```
furiel: Can we just use --with-python=3? Theoretically autotools and cmake should pick up the installed version automatically with it. Otherwise I'm afraid some specific versions are not available in some distributions.

bazsi: did just that. I have also added a CONFIGURE_ADD environment variable to customize environment variables during the bootstrap process.

modules/geoip/geoip-parser.c
bazsi: it was the unit test that crashed, not production. and that crashed because it doesn't check
initialization problems.

kira-syslogng:

kira-syslogng:
failure

gaborznagy:
We had put back this PR to `in progress` state as among failing kira & jenkins some changes requested as well.

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:

kira-syslogng:
failure

kira-syslogng:

kira-syslogng:
failure

kira-syslogng:

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
failure

bazsi:
With #1937 going in, this has collapsed to become a much smaller branch.

Now it's basically the followings:

1) add support for Python virtualenvs in both cmake and Makefiles
2) fixing up the PyBytes/PyUnicode mess that was here earlier (the bulk of the branch)
3) pylint and C warning fixes
4) fixing up the name for mod-python in the core, that is auto-loaded in case  --interactive command line option is present. This makes the interactive debugger to work again.

From this, only #2 is review intensive, the others are pretty much trivia.

kira-syslogng:
failure

bazsi:
I made a stupid rebase bug, causing a compilation error. This now compiles, make check runs fine.

furiel:
@syslog-ng retest this please

kira-syslogng:
failure

furiel:
@bazsi the e2e tests fail because python module used to accept integers when setting the logmessage. Now this does not work. Example from the python parser test:
```
    msg["MY_COUNTER"] = self.counter
    self.counter += 1
```

bazsi:
Thanks for diagnosing these. I didnt yet get around to check the logs.

I think that supportint this is not a good idea, as syslog-ng will store
stuff as strings inside. But i have to understand how that +1 actually
worked.

Bazsi

furiel:
Sorry I was confusing before: the messages of course did not store integer objects. We were only allowed to pass integer objects to setattr. With the +1 above I only intended to show the type of self.counter is int. The complete process function is
```
    def parse(self, msg):
        match = self.regex.match(msg["MESSAGE"])
        if match:
            for key, value in match.groupdict().items():
                print("key: %s, value: %s" % (key, value))
                msg[key] = value
            msg["MY_COUNTER"] = self.counter
            self.counter += 1
            return True
        return False
```
when `self.counter` is initialized to 0 in the `init` method. In the end we just check the counter value as string in a templated filesource:
```
template='"[...] my_counter: $MY_COUNTER\\n"'
```

bazsi:
Now I get it. But I am on the same opinion as before, we shouldn't do an str() on the object as that can cause all strange stuff, like unicode/bytes mismatch. This was one of the things I've removed.

I think it is better to be explicit in the Python side, so if one wants to store an int there, we can explicitly call str() on the object, or create a wrapper around the message object and do the type conversion there.

Agree?

bazsi:
@kira-syslogng retest this please branch=dont-store-ints-in-python-logmsg

furiel:
Yes, I agree. Otherwise users might assume logmessage can store other kind of objects than strings, which is not the case.

bazsi:
Let's see if @kira-syslogng agrees. :) then it should be all green.

kira-syslogng:
failure

bazsi:
@kira-syslogng retest this please; branch=dont-store-ints-in-python-logmsg

kira-syslogng:
failure

bazsi:
@kira-syslogng test this please test branch=dont-store-ints-in-python-logmsg; test=functions/other-functions/python-parser;

kira-syslogng:
success

bazsi:
@kira-syslogng test this please test branch=dont-store-ints-in-python-logmsg;

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
failure

bazsi:
I've forgotten to push to the testdb branch. Lets see if this works now.


@kira-syslogng test this please test branch=dont-store-ints-in-python-logmsg;

kira-syslogng:
failure

furiel:
global_msg_allocated_bytes has the same problem: we add fields to the message with python parser and check if allocated bytes (statistics) if it is between an expected range. As the parser fails, no field is added, so we fail on the lower bound.

PR #1914, opened 2018-03-04
Make the data type explicit in transport mock

mochrul:
Previously there was a heuristic in that object
which tried to detect if the data is a string or an
error code. But this heuristic was failing in GNU Hurd.
With this patch, the heuristic is replaced with a
type field in a structure.

This patch fixes issue #1912

I welcome every suggestion and, as you can see, the giving name is still not my cup of tea, so if somebody knows a better name for the variables or the new structure, I would be more than happy to rename them.

### DIFF
--- a/libtest/mock-transport.c
+++ b/libtest/mock-transport.c
@@ -28,14 +28,27 @@
 #include <string.h>
 #include <errno.h>
 
+typedef enum data_type
+{
+  DATA_STRING,
+  DATA_ERROR,
+} data_type_t;
+
+typedef struct data
+{
+  data_type_t type;
+  struct iovec iov;
+  guint error_code;
+} data_t;
+
 typedef struct
 {
   LogTransport super;
   /* data is stored in a series of data chunks, that are going to be returned as individual reads */
-  struct iovec iov[32];
-  gint iov_cnt;
+  data_t value[32];
+  gint value_cnt;
   /* index currently read i/o chunk */
-  gint current_iov_ndx;
+  gint current_value_ndx;
   /* position within the current I/O chunk */
   gint current_iov_pos;
   gboolean input_is_a_stream;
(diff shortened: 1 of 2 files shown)


kira-syslogng:

MrAnno:
@kira-syslogng ok to test

kira-syslogng:

gaborznagy:
Hi @mochrul!

Just a question out of curiosity:
Which version of syslog-ng will you use/need on Hurd? Just because you state the problem is on 3.6 and you are patching it in upstream.

mochrul:
If you check the bug ticket you can see that that is just a replica.
The original ticket is dated back to 2014 when 3.6 is not yet released officially (or just barely released). I just copied all the information from that ticket into this new, because that ticket is closed and I was not able to reopen it. And because the bug is more than 3 years old, I think fixing this in HEAD is enough.

Also, I do not use Hurd but I'm the Debian maintainer of the package and Debian wants to compile it on Hurd. Most of my bug reports and patches are related to the Debian packaging.

gaborznagy:
Thanks for the clarification, I missed to see the 2014 date and that it is
the same as #355.

libtest/mock-transport.c
```diff
@@ -115,19 +134,28 @@ log_transport_mock_init(LogTransportMock *self, gchar *read_buffer1, gssize read
...
-
-      if (length < 0)
-        length = strlen(buffer);
-
-      self->iov[self->iov_cnt].iov_base = buffer;
-      self->iov[self->iov_cnt].iov_len = length;
-      self->iov_cnt++;
+      g_assert(self->value_cnt < sizeof(self->value) / sizeof(self->value[0]));
+
+      if (length == -2)
```
Kokan: It is kinda hard to connect this `-2` value and the one in the `LTM_INJECT_ERROR` macro.
Would you create a constant for this ?
```c
#define LTM_INJECT_ERROR_LENGTH -2
#define LTM_INJECT_ERROR(err)   (GUINT_TO_POINTER(err)), LTM_INJECT_ERROR_LENGTH
```

kira-syslogng:

kira-syslogng:
success

libtest/mock-transport.c
mochrul: Done.

kira-syslogng:

kira-syslogng:
success

libtest/mock-transport.c
Kokan: Thank you, one small thing. Your update does not compile because you forget the `-2` from the `LTM_InJECT_ERROR_LENGTH` definition.

kira-syslogng:
success

kira-syslogng:

kira-syslogng:
success

kira-syslogng:

kira-syslogng:
success

PR #1918, opened 2018-03-08
Dbld: Update package list for Debian Jessie

mitzkia:


dbld/images/required-apt/all.txt
```diff
@@ -44,10 +44,12 @@ strace
 vim
 
 # debian packaging tools
+fakeroot
```
szemere: If you put a package into the "all.txt" list, than you should:
- check if the package is really available on all platforms (fakeroot is)
- remove it from the distribution specific lists (in this case: trusty)

mitzkia: Thanks for the feedback.
I have removed dh-exec and fakeroot from platform specific files.

kira-syslogng:

kira-syslogng:

PR #1924, opened 2018-03-14
Diskq: Fix crash when corrupt disk-queue is read (during move)

gaborznagy:
After fixing the restart functionality of disk-queue (see PR #1886) our test cases found another bug in non-reliable disk-queue.

After a successful message pop from disk-queue (either from qout or from overflow buffer or from the disk itself) we call a `move_disk` function to move the messages between the different queues.
Order of moving messages:
qout <-- qdisk <-- qoverflow.

In this `move_disk` function there is a lack of error handling when reading from the disk which results in a crash when `log_msg_get_size` is called with passing a NULL pointer to it.

Note that reliable disk-buffer is not affected.

Question:
* When will the invalid read reported to the user of disk-queue? `_move_disk()` is a void function.

### DIFF
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -1852,6 +1852,8 @@ log_msg_lookup_time_stamp_name(const gchar *name)
 
 gssize log_msg_get_size(LogMessage *self)
 {
+  if(!self) return 0;
+
   return
     sizeof(LogMessage) + // msg.static fields
     + self->alloc_sdata * sizeof(self->sdata[0]) +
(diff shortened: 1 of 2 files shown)


kira-syslogng:
success

kira-syslogng:

lib/logmsg/logmsg.c
```diff
@@ -1852,6 +1852,8 @@ log_msg_lookup_time_stamp_name(const gchar *name)
 
 gssize log_msg_get_size(LogMessage *self)
 {
+  if(!self) return 0;
```
furiel: I would prefer an assert here.

gaborznagy: I agree with your reason that a NULL is not a valid LogMessage, but I would not change it.
Using assert would lead an immediate crash which would be unnecessary if we can handle it gracefully.
* There is no valid LogMessage that would be evaluated to 0, therefore a 0 is always an error.
* If we would like to avoid syslog-ng crashes and assert would be preferred then each `log_msg_get_size` call should be guarded. Of course with returning 0 then the user of `log_msg_get_size` should make the error handling.

furiel: Thanks for explaining, fine for me

kira-syslogng:

kira-syslogng:
success

PR #1926, opened 2018-03-16
basicfuncs: add support for negative indexes in $(list-nth)

bazsi:
Negative indexes are interpreted in a similar way to $(list-slice), e.g.
they are calculated from the end of the list.

### DIFF
--- a/modules/basicfuncs/list-funcs.c
+++ b/modules/basicfuncs/list-funcs.c
@@ -86,8 +86,8 @@ _list_count(gint argc, GString *argv[])
 }
 
 static void
-_adjust_list_slice(gint argc, GString *argv[],
-                   gint *first_ndx, gint *last_ndx)
+_translate_negative_list_slice_indices(gint argc, GString *argv[],
+                                       gint *first_ndx, gint *last_ndx)
 {
   gint count = -1;
 
@@ -111,7 +111,7 @@ _list_slice(gint argc, GString *argv[], GString *result,
   if (argc == 0)
     return;
 
-  _adjust_list_slice(argc, argv, &first_ndx, &last_ndx);
+  _translate_negative_list_slice_indices(argc, argv, &first_ndx, &last_ndx);
 
   /* NOTE: first_ndx and last_ndx may be negative, so these loops must cover
    * that case, by interpreting negative first_ndx as "0", and negative
@@ -140,6 +140,14 @@ _list_slice(gint argc, GString *argv[], GString *result,
 }
 
 static void
+_translate_negative_list_index(gint argc, GString *argv[],
+                               gint *ndx)
+{
+  if (*ndx < 0)
+    *ndx += _list_count(argc, argv);
+}
+
+static void
 _list_nth(gint argc, GString *argv[], GString *result, gint ndx)
 {
   ListScanner scanner;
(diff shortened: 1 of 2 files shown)


kira-syslogng:

kira-syslogng:
success

modules/basicfuncs/list-funcs.c
```diff
@@ -139,6 +139,19 @@ _list_slice(gint argc, GString *argv[], GString *result,
...
+static void
+_adjust_list_index(gint argc, GString *argv[],
+                   gint *ndx)
+{
+  gint count = -1;
+
+  if (*ndx < 0)
+    count = _list_count(argc, argv);
+
+  if (*ndx < 0)
```
szemere: Didn't you mean? `count > 0`

modules/basicfuncs/list-funcs.c
```diff
@@ -151,6 +164,8 @@ _list_nth(gint argc, GString *argv[], GString *result, gint ndx)
   list_scanner_init(&scanner);
   list_scanner_input_gstring_array(&scanner, argc, argv);
 
+  _adjust_list_index(argc, argv, &ndx);
```
szemere: This is clearly just a personal opinion! Leave it as is if you want.
I would prefer: _translate_negative_index_into_nth

modules/basicfuncs/list-funcs.c
```diff
@@ -151,6 +164,8 @@ _list_nth(gint argc, GString *argv[], GString *result, gint ndx)
...
+
   i = 0;
   while (i < ndx && list_scanner_scan_next(&scanner))
```
szemere: I don't know what is the desired behaviour in case of index errors, but with the original code in case of "index overflow" we will do nothing.
After this loop `i == ndx`, and the `list_scanner_scan_next` will return FALSE in the next if statement.

With this modification in case of negative "overflow". i.e: `length == 3; ndx == -4`
`_adjust_list_index` will produce: `ndx == -1`
In the loop we will exit with `i==0 > ndx==-1`.
And in the next if statement: `i==0 > ndx == -1` && `list_scanner_scan_next` will return TRUE and set the position to the first element.

modules/basicfuncs/list-funcs.c
bazsi: This was originally modeled after _adjust_list_slice(), which only triggered the count calculation
if either first_ndx or last_ndx was negative. With this one, only the ndx can be negative, so no point in having both conditionals. I have now merged them.

Thanks for catching this.

modules/basicfuncs/list-funcs.c
bazsi: This was a good catch, and indeed my testcases were not enough to catch this. I have now added a couple more tests and also fixed the code.

Thanks for spotting this.

bazsi:
With that I think I have addressed all review comments.

modules/basicfuncs/list-funcs.c
```diff
@@ -139,6 +139,19 @@ _list_slice(gint argc, GString *argv[], GString *result,
   list_scanner_deinit(&scanner);
 }
 
+static void
+_translate_negative_list_index(gint argc, GString *argv[],
+                   gint *ndx)
+{
+  gint count = -1;
```
furiel: This line is not used anymore. The whole variable can be eliminated.

kira-syslogng:
success

kira-syslogng:
success

PR #1928, opened 2018-03-16
template: remove unneeded name argument from call sites

bazsi:
The LogTemplate->name member is used to store the name of the template
statement, so it is not needed in case we are creating inline template
instances. Get rid off that parameter from everywhere.


kira-syslogng:

kira-syslogng:
success

PR #1931, opened 2018-03-21
Python logmsg exceptions

bazsi:
This one improves the Python API by adding exceptions to the error paths. This branch depends on #1913 to get in first.

It is an alternative implementation to #1920

### DIFF
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -226,7 +226,8 @@ _py_init_bindings(PythonDestDriver *self)
       msg_error("Error looking Python driver class",
                 evt_tag_str("driver", self->super.super.super.id),
                 evt_tag_str("class", self->class),
-                evt_tag_str("exception", _py_fetch_and_format_exception_text(buf, sizeof(buf))));
+                evt_tag_str("exception", _py_format_exception_text(buf, sizeof(buf))));
+      _py_finish_exception_handling();
       return FALSE;
     }
 
@@ -238,7 +239,8 @@ _py_init_bindings(PythonDestDriver *self)
       msg_error("Error instantiating Python driver class",
                 evt_tag_str("driver", self->super.super.super.id),
                 evt_tag_str("class", self->class),
-                evt_tag_str("exception", _py_fetch_and_format_exception_text(buf, sizeof(buf))));
+                evt_tag_str("exception", _py_format_exception_text(buf, sizeof(buf))));
+      _py_finish_exception_handling();
       return FALSE;
     }
 
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -398,14 +398,6 @@ python_dd_worker_init(LogThrDestDriver *d)
 }
 
 static void
-python_dd_worker_deinit(LogThrDestDriver *d)
-{
-  PythonDestDriver *self = (PythonDestDriver *)d;
-
-  python_dd_close(self);
-}
-
-static void
 python_dd_disconnect(LogThrDestDriver *d)
 {
   PythonDestDriver *self = (PythonDestDriver *) d;
@@ -511,7 +503,6 @@ python_dd_new(GlobalConfig *cfg)
   self->super.messages.retry_over = python_dd_over_message;
 
   self->super.worker.thread_init = python_dd_worker_init;
-  self->super.worker.thread_deinit = python_dd_worker_deinit;
   self->super.worker.disconnect = python_dd_disconnect;
   self->super.worker.insert = python_dd_insert;
 
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -22,6 +22,7 @@
  *
  */
 #include "python-logmsg.h"
+#include "python-helpers.h"
 #include "logmsg/logmsg.h"
 #include "messages.h"
 #include "str-utils.h"
@@ -69,13 +70,13 @@ _is_key_blacklisted(const gchar *key)
 static PyObject *
 _py_log_message_getattr(PyObject *o, PyObject *key)
 {
-  if (!py_is_string(key))
+  if (!_py_is_string(key))
     {
       PyErr_SetString(PyExc_TypeError, "key is not a string object");
       return NULL;
     }
 
-  const gchar *name = py_object_as_string(key);
+  const gchar *name = _py_get_string_as_string(key);
 
   if (_is_key_blacklisted(name))
     {
@@ -101,31 +102,32 @@ _py_log_message_getattr(PyObject *o, PyObject *key)
 static int
 _py_log_message_setattr(PyObject *o, PyObject *key, PyObject *value)
 {
-  if (!py_is_string(key))
+  if (!_py_is_string(key))
     {
       PyErr_SetString(PyExc_TypeError, "key is not a string object");
       return -1;
     }
 
-  PyLogMessage *py_msg = (PyLogMessage *)o;
-  const gchar *name = py_object_as_string(key);
+  PyLogMessage *py_msg = (PyLogMessage *) o;
+  const gchar *name = _py_get_string_as_string(key);
   NVHandle handle = log_msg_get_value_handle(name);
   PyObject *value_as_strobj = PyObject_Str(value);
+
   if (value_as_strobj)
     {
       if(log_msg_is_write_protected(py_msg->msg))
         {
           msg_debug("python: error while modifying msg",
-                    evt_tag_printf("msg", "%p", py_msg),
+                    evt_tag_printf("msg", "%p", py_msg->msg),
                     evt_tag_str("name", name),
-                    evt_tag_str("value", py_object_as_string(value_as_strobj)));
+                    evt_tag_str("value", _py_get_string_as_string(value_as_strobj)));
 
           PyErr_SetString(PyExc_RuntimeError, "msg is write protected");
           Py_DECREF(value_as_strobj);
           return -1;
         }
 
-      log_msg_set_value(py_msg->msg, handle, py_object_as_string(value_as_strobj), -1);
+      log_msg_set_value(py_msg->msg, handle, _py_get_string_as_string(value_as_strobj), -1);
       Py_DECREF(value_as_strobj);
     }
   else
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -122,16 +122,18 @@ _py_log_message_ass_subscript(PyObject *o, PyObject *key, PyObject *value)
     }
 
   NVHandle handle = log_msg_get_value_handle(name);
-  PyObject *value_as_strobj = PyObject_Str(value);
 
-  if (value_as_strobj)
+  if (value && _py_is_string(value))
     {
-      log_msg_set_value(py_msg->msg, handle, _py_get_string_as_string(value_as_strobj), -1);
-      Py_DECREF(value_as_strobj);
+      log_msg_set_value(py_msg->msg, handle, _py_get_string_as_string(value), -1);
     }
   else
     {
-      /* propagate exception coming from the str() call */
+      PyErr_Format(PyExc_ValueError,
+                   "str or unicode object expected as log message values, got type %s (key %s). "
+                   "Earlier syslog-ng accepted any type, implicitly converting it to a string. "
+                   "With this version please convert it explicitly to a string using str()",
+                   value->ob_type->tp_name, name);
       return -1;
     }
 
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -235,7 +235,7 @@ static PyMethodDef py_log_message_methods[] =
 
 static PyTypeObject py_log_message_type =
 {
-  PyObject_HEAD_INIT(&PyType_Type)
+  PyVarObject_HEAD_INIT(&PyType_Type, 0)
   .tp_name = "LogMessage",
   .tp_basicsize = sizeof(PyLogMessage),
   .tp_dealloc = (destructor) py_log_message_free,
@@ -244,6 +244,7 @@ static PyTypeObject py_log_message_type =
   .tp_new = PyType_GenericNew,
   .tp_as_mapping = &py_log_message_mapping,
   .tp_methods = py_log_message_methods,
+  0,
 };
 
 void
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -68,7 +68,7 @@ _is_key_blacklisted(const gchar *key)
 }
 
 static PyObject *
-_py_log_message_getattr(PyObject *o, PyObject *key)
+_py_log_message_subscript(PyObject *o, PyObject *key)
 {
   if (!_py_is_string(key))
     {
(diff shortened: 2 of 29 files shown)


kira-syslogng:
failure

kira-syslogng:

kira-syslogng:
failure

kira-syslogng:
failure

bazsi:
@kira-syslogng test this please test branch=dont-store-ints-in-python-logmsg;

bazsi:
This is now a combination of both the original #1913 and the stuff that was submitted in #1931. This is easier to review as the first branch only started the changes which this branch finished at the end.

There are a few breaking changes in here:

1) now the Python destination driver would raise exceptions of various error conditions instead only logging them in the syslog-ng internal log. This means that for invalid name-value pairs the Python code may get exceptions they didn't get before.

2) we don't explicitly call an str() for all values stored in a LogMessage, which means that anything that tried to set a non-str non-unicode object in the LogMessage would result in an exception, whereas earlier we silently converted it to a string.

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
success

kira-syslogng:
success

modules/python/python-dest.c
```diff
@@ -226,7 +226,8 @@ _py_init_bindings(PythonDestDriver *self)
       msg_error("Error looking Python driver class",
                 evt_tag_str("driver", self->super.super.super.id),
                 evt_tag_str("class", self->class),
-                evt_tag_str("exception", _py_fetch_and_format_exception_text(buf, sizeof(buf))));
+                evt_tag_str("exception", _py_format_exception_text(buf, sizeof(buf))));
+      _py_finish_exception_handling();
```
furiel: Can you wrap this (msg_error+finish) up into a macro or a va_arg function? I think this `_py_finish_exception_handling` is easy to miss, and this pattern is very frequent.

modules/python/python-logmsg.c
```diff
@@ -109,30 +109,30 @@ _py_log_message_setattr(PyObject *o, PyObject *key, PyObject *value)
     }
 
   PyLogMessage *py_msg = (PyLogMessage *) o;
+  LogMessage *msg = py_msg->msg;
   const gchar *name = _py_get_string_value_as_utf8(key);
+
+  if (log_msg_is_write_protected(msg))
+    {
+      PyErr_Format(PyExc_TypeError,
```
furiel: This should be a RuntimeError.

modules/python/python-logmsg.c
```diff
@@ -100,7 +100,7 @@ _py_log_message_getattr(PyObject *o, PyObject *key)
 }
 
 static int
-_py_log_message_setattr(PyObject *o, PyObject *key, PyObject *value)
+_py_log_message_ass_subscript(PyObject *o, PyObject *key, PyObject *value)
```
furiel: What does `ass_subscript` abbreviate?

cmake/python_functions.cmake
```diff
@@ -0,0 +1,30 @@
...
+#
+#############################################################################
+
+find_program(PEP8_EXECUTABLE NAMES pep8 pycodestyle HINTS ${CMAKE_BINARY_DIR}/.venv/bin)
+find_program(NOSETESTS_EXECUTABLE nosetests HINTS ${CMAKE_BINARY_DIR}/.venv/bin)
+find_program(PYLINT_EXECUTABLE pylint HINTS ${CMAKE_BINARY_DIR}/.venv/bin)
+
+
+add_custom_target(python-venv
+	COMMAND rm -rf .venv && virtualenv -p ${PYTHON_EXECUTABLE} .venv && .venv/bin/pip install -r ${CMAKE_HOME_DIRECTORY}/requirements.txt)
```
furiel: I just wanted to try this out from a clean environment. In my case there is no .venv directory, so `make python-venv` failed.

Does `virtualenv` work with python3? As far as I know, for python3, we should use `pyvenv`. This should be a problem in the autotools version as well.

cmake/python_functions.cmake
```diff
@@ -0,0 +1,30 @@
...
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+find_program(PEP8_EXECUTABLE NAMES pep8 pycodestyle HINTS ${CMAKE_BINARY_DIR}/.venv/bin)
+find_program(NOSETESTS_EXECUTABLE nosetests HINTS ${CMAKE_BINARY_DIR}/.venv/bin)
+find_program(PYLINT_EXECUTABLE pylint HINTS ${CMAKE_BINARY_DIR}/.venv/bin)
```
furiel: Can you explain how to use the virtualenv? (What commands should I execute on a clean build directory)

What I don't understand is: if I want to use the python-venv target, I need to specify what python version I want to use. So before `make python-venv`, I need to run `cmake [...] -DPYTHON_VERSION=3`. But if I execute cmake without venv existing beforehand, the pep8, nosetests, pylint detection will fail during that, because the directories in HINTS do not exist. So in the end, it will pick up the python2 version from the system. The problem should exist in the autotools version.

As I think, we cannot reliably create python-venv target, because it depends on the configure/cmake due to python version, but it also needs to modify behaviour of configure/cmake due to pep8 nosetests etc (some kind of chicken-egg problem).

modules/python/python-dest.c
bazsi: It was originally part of _py_fetch_and_format_exception_text() and I explicitly extracted that, simply because then it was seemingly a log message that also happened to clear the error state.

Or you mean that I should wrap the msg_error() plus the cleanup? I'll try to see how that looks.

modules/python/python-logmsg.c
bazsi: no clue, https://docs.python.org/3/c-api/typeobj.html#c.PyMappingMethods

cmake/python_functions.cmake
bazsi: rm -rf should remove .venv without an error if it doesn't exist. virtualenv does work for python3 as well, yes.

cmake/python_functions.cmake
bazsi: my understanding was that find_program() would try the hints and then the normal executables without a path. Isn't this the behavior?

kira-syslogng:
failure

modules/python/python-dest.c
furiel: Yes, I would be happy to see msg_error and  _py_finish_exception_handling() bundled together. Maybe into _py_msg_exception() or similar function/macro. Thanks.

modules/python/python-logmsg.c
furiel: Oh ok, so it comes from the official cython api.

cmake/python_functions.cmake
furiel: You are right! The problem was not with rm -rf, but with `${PYTHON_EXECUTABLE}`:
```
The path .venv (from --python=.venv) is not an executable file
```
Probably the executable detection happens in a later phase, so `${PYTHON_EXECUTABLE}` is an empty string, then virtualenv fails because she tries to execute `virtualenv -p .venv`.

Right, thanks for explaining virtualenv and python3. I was confused because python3-virtualenv on ubuntu does not ship the virtualenv executable, that's why I was thinking they are incompatible. If I pass python3 binary to -p, it works indeed.

cmake/python_functions.cmake
furiel: Yes, but the problem is the time of the detection. As an example: in this case PEP8_EXECUTABLE will be filled when you call `cmake`. During that, indeed cmake will try the hint. But that time, we do not have `.venv`, so it will fall back to the global version. Then cmake will cache it, and it will not change even if you reexecute cmake.
From the [documentation](https://cmake.org/cmake/help/v3.4/command/find_path.html):
```
 If the file in a directory is found the result is stored in the variable and the search will not be repeated unless the variable is cleared
```

This is probably something different than in autotools. In case of autotools
```
PYTHON = $(or $(wildcard $(PYTHON_VENV_DIR)/bin/python),$(PYTHON))
```
autotools will generate a makefile with:
````
Makefile
4146:VPYTHON = $(or $(wildcard $(PYTHON_VENV_DIR)/bin/python),$(PYTHON))
```
so in the end the expression will probably recalculated each time you call make, thus causing no problem.

Mk/Makefile.am
```diff
@@ -9,3 +23,8 @@ tools_DATA	= \
 
 tools_SCRIPTS	= \
 		lib/merge-grammar.py
+
+python-venv:
+	rm -rf $(PYTHON_VENV_DIR)
```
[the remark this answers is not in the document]
bazsi: I disagree here, if you want to use a custom venv just create and activate it before launching the syslog-ng build system. In that case the PATH would be set up properly and we would find pep8 and the rest in that virtualenv via PATH.

Running python-venv is completely optional and it is typically managed by the build system itself. What we do basically is that for this special virtualenv, we "auto-activate" it during build without you having to care about it.

bazsi:
I think I would separate python-venv support from the bulk of this branch. I think except of virtualenv the rest could go in. What do you think?

furiel:
@bazsi unfortunately every testcase fails with python3, but I think the root cause is the same: when options is used, there is a type mismatch in the lookup, hence a keyerror.

Example:
config:
```
destination d_python {
    python(class(mymodule.MyDestination) options("key" "value"));
};
```

module:
```
class MyDestination(object):
    def init(self, options):
        print(options["key"])
        return True
    def send(self, message):
        return True
```

syslog-ng log:
```
[2018-04-10T19:30:28.674139] Exception while calling a Python function; caller='d_python#0', class='mymodule.MyDestination', function='init', exception='KeyError: \'key\''
[2018-04-10T19:30:28.674177] Error initializing Python driver object, init() returned FALSE; driver='d_python#0', class='mymodule.MyDestination'
```

When I print the whole options object, I get: `{b'key': b'value'}`, so we store bytes inside the `options`.
When I print `print(options[b"key"])`, syslog-ng can start successfully.

bazsi:
Another iteration.
1) rebased against current master
2) dropped the venv related stuff in favour of doing that in a separate branch
3) I've fixed the options related problem for Python3

Hopefully this goes green :)

kira-syslogng:
success

bazsi:
I hope this is the final iteration at this point. These are the changes I've applied to this branch since the last review round:

* dropped venv support into a separate pull request #1985
* python3 compatibility fixes for the detailed traceback reporting (we needed to use unicode for the source instead of bytes)
* the filename of the Python main module in syslog-ng.conf now includes the line number where the python block begins
* error handling in Python exception handling was improved, earlier I just assumed the needed symbols are there, but in reality they were not in Python3. Now we properly handle if "traceback" module is not present or a print_exception() function does not exist.

furiel:
The options problem is resolved! Thank you. Still all tests fail with python3 :( but as I see these are testcase errors. For some reason when opening a log file from user code, we pass an `encoding=0` parameter to open, which fails with python3: 
```
>>> f = open("tmp.txt", "a", 0)
  File "<stdin>", line 1, in <module>
ValueError: can't have unbuffered text I/O
```
I'll try to patch the tests asap.

kira-syslogng:
success

kira-syslogng:
success

furiel:
@bazsi, the python3 version is still not working. The values returned from the the message object should be of string type (unicode), and not byte.

A simple
 ```
class MyDestination(object):
    def send(self, msg):
        print(msg["MSG"])
        print(type(msg["MSG"]))
        return True
```
will produce:
```
b'asdfdasf'
<class 'bytes'>
```

**Comment by furiel** [comment]
@bazsi there is a problem with exception handling as well. With simply starting syslog-ng syslog-ng, syslog-ng will emit
```
2018-04-12T11:21:27.177934] Error printing proper Python traceback for the exception, printing the error caused by print_exception() itself;
SystemError: NULL object passed to Py_BuildValue
```

**Comment by bazsi** [comment]
We've discussed the LogMessage returning bytes instead of unicode and concluded that it remains this way, as syslog-ng can't guarantee that name-value pairs are utf8 encoded, thus it might not be possible to convert them to unicode.

the Python code can undertake the conversion and handle it appropriately.

**Comment by bazsi** [comment]
I have now fixed the ugly exception message. We were handling an exception that was already handled.

Also, I've noticed that python destinations called close() twice, once in worker_deinit() and once in disconnect(). As disconnect is automatically called by LogThreadedDriver one of the calls could be removed, thus I eliminated the worker_deinit() method.

With this the only remaining thing is to fix Kira's test database to work on Python3 as well.

**Comment by kira-syslogng** [comment]
success

**Comment by furiel** [comment]
@kira-syslogng retest this please test branch=buffering-and-open;

**Comment by kira-syslogng** [comment]
failure

**Comment by furiel** [comment]
@kira-syslogng retest this please test branch=buffering-and-open;

**Comment by kira-syslogng** [comment]
success

**Comment by furiel** [comment]
@bazsi unfortunately I ran into another issue: with python template function.
With:
```
@version: 3.14

log {
    source { stdin(flags(no-parse)); };
    destination { file("/dev/stdout" template("$(python foo)")); };
};

python {
def foo(msg):
    return "bar"
};
```
I get
```
[2018-04-12T19:11:59.383326] $(python): The return value is not a string; function='foo', type='str'
```

If I use `return b"bar"`, instead of `return "bar"`, it works as expected.

How to proceed with this? Me seems strange that users need to convert to bytes for the return value. I agree the message object should carry PyBytes, but for the string that is created with a template, I'm not sure.

In any case, the error message is misleading here.

kira-syslogng:
failure

bazsi:
@kira-syslogng retest this please test branch=buffering-and-open;

bazsi:
I have fixed the template function invocation as well. It now accepts either an unicode or a bytes object.

bazsi:
Kira seems to have had an issue fetching the repos from github, let's start it again.


@kira-syslogng retest this please test branch=buffering-and-open;

kira-syslogng:
failure

pzoleex:
@kira-syslogng retest this please test branch=buffering-and-open;

pzoleex:
@kira-syslogng retest this please test branch=buffering-and-open

kira-syslogng:
failure

pzoleex:
@kira-syslogng retest this please test branch=buffering-and-open

kira-syslogng:
success

bazsi:
@furiel finally everything is green, if your separate python3 related tests run fine, this can be merged, I think.

I am waiting for your approval, once there I'd merge it, unless there's a reason not to.

furiel:
Thanks @bazsi. Only one testcase still fails, the loaders test (hy module) in the parsers. But the problem is not with the python compatibility, something is with the environment, or the testcase itself. Strangely, when I execute it locally, it passes even with python3 (and hy3). Moreover, the destination version of the same test passes even in kira. I will fight with it later, let's not stop the merge due to this.

I checked these last patches, and they look good as well. Great patchet, it was a very complex problem. I approve it for merge.

MrAnno:
I've checked, this PR compiles on macOS correctly, so the rebase is not necessary here. :+1:

bazsi:
Thanks for the feedback and the help.

PR #1932, opened 2018-03-21
Improve error reporting

bazsi:
This patch vastly improves error reporting in case a syntax error happens in the configuration. It is in a way doing something similar to what #1931 does with Python exceptions.

Should make troubleshooting config issues a lot simpler.

### DIFF
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -920,14 +920,20 @@ relex:
       (gen = cfg_lexer_find_generator(self, self->cfg, cfg_lexer_get_context_type(self), yylval->cptr)))
     {
       CfgArgs *args;
+      CfgIncludeLevel *level = &self->include_stack[self->include_depth];
 
       self->preprocess_suppress_tokens++;
+
+      gint saved_line = level->lloc.first_line;
+      gint saved_column = level->lloc.first_column;
       if (cfg_parser_parse(&block_ref_parser, self, (gpointer *) &args, NULL))
         {
           gboolean success;
           gchar buf[256];
           GString *result = g_string_sized_new(256);
 
+          level->lloc.first_line = saved_line;
+          level->lloc.first_column = saved_column;
           self->preprocess_suppress_tokens--;
           success = cfg_block_generator_generate(gen, self->cfg, args, result);
 
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -243,6 +243,7 @@ cfg_lexer_clear_include_level(CfgLexer *self, CfgIncludeLevel *level)
   else if (level->include_type == CFGI_BUFFER)
     {
       g_free(level->buffer.content);
+      g_free(level->buffer.original_content);
     }
   memset(level, 0, sizeof(*level));
 }
@@ -300,7 +301,10 @@ cfg_lexer_start_next_include(CfgLexer *self)
       g_free(level->name);
 
       if (level->include_type == CFGI_BUFFER)
-        g_free(level->buffer.content);
+        {
+          g_free(level->buffer.content);
+          g_free(level->buffer.original_content);
+        }
 
       memset(level, 0, sizeof(*level));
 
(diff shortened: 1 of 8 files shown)


kira-syslogng:
success

kira-syslogng:

lib/cfg-lexer.c
```diff
@@ -226,6 +226,27 @@ cfg_lexer_lookup_keyword(CfgLexer *self, YYSTYPE *yylval, YYLTYPE *yylloc, const
...
+      if (level->file.include_file)
+        fclose(level->file.include_file);
+      g_slist_foreach(level->file.files, (GFunc) g_free, NULL);
+      g_slist_free(level->file.files);
+    }
+  else if (level->include_type == CFGI_BUFFER)
+    {
+      g_free(level->buffer.content);
+    }
+  memset(level, 0, sizeof(*level));
```
furiel: As I see this patch is just a refactor, except this line, which is a new addition. Can you explain the purpose of this line? If needed, probably better do outside the function. For now this function modifies her input parameter.

lib/cfg-parser.c
```diff
@@ -186,68 +186,100 @@ CfgParser main_parser =
...
   if (f)
     {
-      while (fgets(buf, sizeof(buf), f) && lineno < yylloc->first_line)
-        lineno++;
-      if (lineno != yylloc->first_line)
-        buf[0] = 0;
+      while (fgets(buf, sizeof(buf), f) && lineno < (gint) yylloc->first_line + CONTEXT)
+        {
+          lineno++;
+          if (lineno > (gint) yylloc->first_line + CONTEXT)
```
furiel: The while condition above ensures, `lineno < (gint) yylloc->first_line + CONTEXT`, together with lineno++ we can only have `lineno <= (gint) yylloc->first_line + CONTEXT`. I think this condition cannot be triggered, it can be safely deleted.

furiel:
I tried out the patch, and the error reporting is fabulous indeed. :) Thanks for putting this together. It works nicely with scl-s too.

bazsi:
I have now fixed your review comments, and went on to improve the error reporting even more. Instead of just the erroring context, this now prints the entire stack of inclusions, like this:

```
Error parsing rewrite expression, rewrite plugin xxxx not found in block parser cisco-timestamp-parser (at /install/share/syslog-ng/include/scl/cisco/plugin.conf:41:1):
1       
2           channel {
3               rewrite {
4-----> xxxx            set("$1" value('1'));
4-----> ^^^^
5               };
6       
7               if {
8                   filter { match('^[.*]?([A-Za-z]{3} [0-9 ]\d \d{2}:\d{2}:\d{2})' value('1') flags(store-matches)); };
9                   parser { date-parser(format('%b %d %H:%M:%S') template("$1")); };

Included from block parser cisco-parser (at /install/share/syslog-ng/include/scl/cisco/plugin.conf:61:1):
17                  subst("^(<[0-9]+>)?([0-9]+)?(: )?", "", value('1'));
18      
19              };
20      
21              if {
22---->             parser { cisco-timestamp-parser(template("$1")
22---->                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
23      
24      
25      
26      
27      ); };

Included from parser generator app-parser:
1       
2       channel {
3           junction {
4       channel {
5----->     parser {  cisco-parser();  };
5----->               ^^^^^^^^^^^^^^
6           rewrite { set-tag('.app.cisco'); };
7           flags(final);
8       };
9           };
10      }

Included from block source default-network-drivers (at /install/share/syslog-ng/include/scl/default-network-drivers/plugin.conf:24:1):
10      				port(514)
11      				log-msg-size(65536)
12      				flags(no-parse, ));
13      		};
14      		channel {
15----> 			parser { app-parser(topic(syslog-raw) ); };
15----> 			         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
16      			flags(final);
17      		};
18      		channel {
19      			parser {
20      				syslog-parser(flags(syslog-protocol));

Included from /install/etc/syslog-ng.conf:
10      	internal();
11      	unix-stream("log");
12      };
13      
14      source s_network {
15----> 	default-network-drivers();
15----> 	^^^^^^^^^^^^^^^^^^^^^^^^^
16      };
17      
18      destination d_local {
19      	file("/var/log/messages");
20      	file("/var/log/messages-kv.log" template("$ISODATE $HOST $(format-welf --scope all-nv-pairs)\n") frac-digits(3));


contact: https://lists.balabit.hu/mailman/listinfo/syslog-ng
```

kira-syslogng:
success

PR #1933, opened 2018-03-21
[TEST][WIP] LogPipe destruction

MrAnno:
This is just a test PR trying to force the LogPipe destruction to run in the main thread.

Calling `new()`, `init()`, `deinit()`, `free()` on a LogPipe instance is, by convention, only valid in the main thread. There are places in our codebase where `log_pipe_unref()` is called from worker threads (for example: `LateAckTracker`). This may result in use-after-free or other problems that cause crashes.

Work in progress, this is NOT the final fix.

### DIFF
--- a/lib/logpipe.c
+++ b/lib/logpipe.c
@@ -24,6 +24,7 @@
 
 #include "logpipe.h"
 #include "cfg-tree.h"
+#include "mainloop-call.h"
 
 gboolean (*pipe_single_step_hook)(LogPipe *pipe, LogMessage *msg, const LogPathOptions *path_options);
 
@@ -95,7 +96,7 @@ log_pipe_unref(LogPipe *self)
 
   if (self && (g_atomic_counter_dec_and_test(&self->ref_cnt)))
     {
-      _free(self);
+      main_loop_call((MainLoopTaskFunc) _free, self, TRUE);
     }
 }
 


kira-syslogng:
success

bazsi:
I don't necessarily like this approach, sometimes it is safe to call log_pipe_unref(), but I'd like to understand what triggered this.

MrAnno:
Hi @bazsi,

This is just a workaround for a bigger problem in `LateAckTracker` (and for other possible places where log_pipe_unref() is called outside of the main thread). I just wanted to try this idea with Kira and discuss the exact problem in `LateAckTracker`:

Every `LogMessage` is bound to its `LogSource` through `AckRecord` (`AckTracker`).
When a message is acked by the destination, its source should exist to receive the acknowledgement (save bookmarks, ...). 
First question: Is this a necessary requirement? I think the answer is yes.

Since it would be possible that the source is already destroyed when one of its message is acked, `AckTracker` calls an extra `log_pipe_ref()` + `log_pipe_unref()` pair on the source. This is perfectly OK except one thing:
`AckTracker` runs in our worker threads. If the last `log_pipe_unref()` call is coming from the `AckTracker`, it violates our convention about the `free()` methods of `LogPipe` (No one expects `free` to be thread-safe).

lbudai:
@bazsi: if we would be able to decide whether the source is still alive or not, then it would not be an issue (unfortunately acktracker needs a source to adjust window size when an ACK is received...)

bazsi:
The log_pip_unref() should be safe to be called from any threads. deinit()
is the problem. And free() does not need to be thread safe, as there would
be only one call to it. And I don't think free() should be anything that
would need to cause problems, as it only calls free and such.

-- 
Bazsi

MrAnno:
@bazsi Thanks, I understand now.

We are working on a custom `LogProtoServer` implementation, where we had to use ivykis timers and other resources.
Unfortunately, LogProto* does not have a `deinit()` method, so we used `free()` incorrectly (we released timers, etc. in the free method causing a crash when it was called outside of the main thread).

Adding `deinit()` to the LogProto interface and calling it in the deinit method of `LogReader` would comply with the rules you mentioned, but then we would face a very similar problem I guess: We either have to make all `deinit()`s threaf-safe or force them to run in the main thread.

MrAnno:
Yes, I can also confirm the afsocket source connection part.
The tricky part is in the `AckTracker`, but I think now I have everything I need to fix the problem correctly.

Sorry for the weird PR again, and thank you for the help! :)

MrAnno:
Will be fixed in a seperate PR.

PR #1935, opened 2018-03-22
file source filename corruption in case of missing path

### LINKED ISSUE #1934: file source wrong filename in case of missing root directory in path
# syslog-ng

## Version of syslog-ng
$ sbin/syslog-ng -V
syslog-ng 3.14.1
Config version: 3.14
Installer-Version: 3.14.1
Revision: 3.14.1
Module-Directory: ~/github.com/syslog-ng/build/install/lib/syslog-ng
Module-Path: ~/github.com/syslog-ng/build/install/lib/syslog-ng
Available-Modules: system-source,date,tags-parser,json-plugin,geoip2-plugin,graphite,mod-python,redis,afuser,afprog,sdjournal,afsocket,appmodel,cef,disk-buffer,afamqp,basicfuncs,stardate,http,pseudofile,snmptrapd_parser,add_contextual_data,getent,geoip-plugin,dbparser,xml,afsql,kvformat,confgen,syslogformat,affile,afsmtp,cryptofuncs,linux-kmsg-format,pacctformat,map_value_pairs,csvparser,afstomp
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: off
Enable-Systemd: off

# Issue

## Failure

```
...
[2018-03-22T06:55:12.807771] Follow-mode file source not found, deferring open; filename='/foo'
[2018-03-22T06:55:12.807809] Follow-mode file source not found, deferring open; filename='/foo'
[2018-03-22T06:55:12.807829] Error checking the uniqueness of the persist names, please override it with persist-name option. Shutting down.; persist_name='affile_sd_curpos(/foo)', location='/home/peterkokai/github.com/syslog-ng/build/install/etc/syslog-ng.conf:10:3'
```

There are two issue (the root cause is the same).
* It starts to follow `/foo` for both source.
* The persist name is `/foo` for both of them.

## Steps to reproduce

1. The `/foo` should not exists.
2. start syslog-ng

## Configuration

```
@version: 3.14

source s_a {
  file("/foo/a/path/that/not/exists.txt");
};

source s_b {
  file("/foo/a/path/that/donot/exists.txt");
};

destination dd {
  file("/dev/stdout");
};

log { source(s_a); destination(dd); };
log { source(s_b); destination(dd); };
```

Kokan:
The file source checks if the provided path exists, and in case of the path does not exists it has a side effect to put a terminating zero, thus corrupting the passed filename argument. That parameter should be read only.
Because of this I also modified the rest of the `file_perm_*` functions, where it did not cause any issue.

* Does it okay to `memcpy` the `name` paramater from performance point of view? (Does it have per message performance penalty?!)

Fixes #1934

### DIFF
--- a/lib/file-perms.c
+++ b/lib/file-perms.c
@@ -181,32 +181,32 @@ file_perm_options_inherit_dont_change(FilePermOptions *self)
 }
 
 gboolean
-file_perm_options_apply_file(const FilePermOptions *self, gchar *name)
+file_perm_options_apply_file(const FilePermOptions *self, const gchar *path)
 {
 #ifndef _MSC_VER
   gboolean result = TRUE;
 
-  if (self->file_uid >= 0 && chown(name, (uid_t) self->file_uid, -1) < 0)
+  if (self->file_uid >= 0 && chown(path, (uid_t) self->file_uid, -1) < 0)
     result = FALSE;
-  if (self->file_gid >= 0 && chown(name, -1, (gid_t) self->file_gid) < 0)
+  if (self->file_gid >= 0 && chown(path, -1, (gid_t) self->file_gid) < 0)
     result = FALSE;
-  if (self->file_perm >= 0 && chmod(name, (mode_t) self->file_perm) < 0)
+  if (self->file_perm >= 0 && chmod(path, (mode_t) self->file_perm) < 0)
     result = FALSE;
   return result;
 #endif
 }
 
 gboolean
-file_perm_options_apply_dir(const FilePermOptions *self, gchar *name)
+file_perm_options_apply_dir(const FilePermOptions *self, const gchar *path)
 {
 #ifndef _MSC_VER
   gboolean result = TRUE;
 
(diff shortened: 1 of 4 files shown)


furiel:
As I see this is only called when a file is opened, so I don't think the strdup would cause too much performance problem.

If you want to be sure, you can still use scratchbuffer. Scratchbuffer would have another advantage though: in the current version you need to introduce a retval variable to free the strdupped variable, which would not be a problem with scratchbuffers. The patch could be a little bit simpler.

kira-syslogng:
success

kira-syslogng:

kira-syslogng:
success

kira-syslogng:

MrAnno:
> Does it okay to memcpy the name paramater from performance point of view? (Does it have per message performance penalty?!)

It's OK, `file_perm_options_create_containing_directory` is called only when the `FileReader` is initialized and when it is reopened (`NC_FILE_MOVED`, `NC_READ_ERROR`).

PR #1937, opened 2018-03-23
Dbld refinements

bazsi:
These are the dbld changes that I've backed out from #1913. This one should go in first, the other depends on this.

The changes are pretty minor and self-explanatory, however the patches are pretty independent, so it might be easier to review them by patches and not as a whole.

### DIFF
--- a/dbld/bootstrap
+++ b/dbld/bootstrap
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+cd /source
+./autogen.sh
+
+cd /build
+echo "Running configure with: ${CONFIGURE_OPTS}"
+/source/configure ${CONFIGURE_OPTS}
+make python-venv
--- a/dbld/deb
+++ b/dbld/deb
@@ -16,6 +16,6 @@ if [ ! -d debian ]; then
 	echo "You have no Debian packaging in debian/, this should already by in the tarball at this point"
 	exit 1
 fi
-dpkg-buildpackage -rfakeroot
+dpkg-buildpackage -rfakeroot -d
 
 echo "Your debs are in /build, also available on the host in \$(top_srcdir)/dbld/build"
--- a/dbld/rules
+++ b/dbld/rules
@@ -99,9 +99,9 @@ login-%: setup
 	$(DOCKER) exec -ti  $$container /bin/bash
 
 setup:
-	mkdir -p dbld/build || true
-	mkdir -p dbld/install || true
-	mkdir -p dbld/release || true
+	@mkdir -p dbld/build || true
+	@mkdir -p dbld/install || true
+	@mkdir -p dbld/release || true
 
 
 # list apt packages
--- a/dbld/rules
+++ b/dbld/rules
@@ -2,6 +2,8 @@
 
 IMAGES=centos6 centos7 jessie artful xenial trusty
 DEFAULT_IMAGE=xenial
+DEFAULT_DEB_IMAGE=$(DEFAULT_IMAGE)
+DEFAULT_RPM_IMAGE=centos7
 DOCKER=docker
 DOCKER_RUN_ARGS=-e USER_NAME_ON_HOST=$(shell whoami)	\
         --network=host \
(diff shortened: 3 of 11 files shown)


kira-syslogng:
success

dbld/tarball
```diff
@@ -7,6 +7,7 @@ VERSION=`cat VERSION`
 ./autogen.sh
 
 cd /build
+rm -rf *
```
furiel: As /build is retained, people might try save things there. Like [bash history](https://github.com/balabit/syslog-ng/pull/1932/commits/e836a7bd26e0f374ffb3092ae83a3fcf6624736d) :) (though in that specific example we are in luck that hidden directories are not deleted by rm -rf *). Better be safe, I suggest create a dedicated tarball directory or mktemp something

dbld/bootstrap
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+cd /source
+./autogen.sh
+
+cd /build
+rm -rf *
```
furiel: Just mark this rm -rf for the same rethinking as in the previous patch.

dbld/rules
```diff
@@ -19,6 +19,7 @@ RELEASE_DIR=$(DBLD_DIR)/release
 VERSION=$(shell cat VERSION)
 TARBALL=$(BUILD_DIR)/syslog-ng-$(VERSION).tar.gz
 MODE=snapshot
+CONFIGURE_OPTS=--enable-debug --enable-manpages --with-python=3 --prefix=/install --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl $(CONFIGURE_ADD)
```
furiel: In the admin guide we state: Currently only Python 2.7 is supported. Let's align --with-python=2 with the documentation.

dbld/deb
```diff
@@ -16,6 +16,6 @@ if [ ! -d debian ]; then
 	echo "You have no Debian packaging in debian/, this should already by in the tarball at this point"
 	exit 1
 fi
-dpkg-buildpackage -rfakeroot
+dpkg-buildpackage -rfakeroot -d
```
furiel: I understand this helps with the python dependencies. But this will ignore the non-python dependencies too, won't it? Is it safe to do so?

dbld/tarball
bazsi: Indeed. The separate directory is what zwa used to do, but I wanted to simplify. But you're right we shouldn't manage /build.

dbld/bootstrap
bazsi: ok

dbld/rules
bazsi: We can do that in the initial merging, however I would like us to start testing Python3 in Jenkins and possibly change the documentation later. We used to work with Python3 and then broke it. This was the state that got documented.

But I'll change it for now.

dbld/deb
bazsi: Yes it will, but I think it is safe to do so.

The build dependencies are used by the distribution's build service and there's value in validating them, but their primary purpose is to control builds in the distro.

Now, we could use build-depends, but that would mean that _all_ build dependencies would need to be converted into .deb packages and hosted on OBS or back in the distribution, which would slow us down. I wouldn't want to work on packaging the latest version of pep8/pycodestyle or pylint. And if we don't use them from pip, we would get errors down the path, when users start to use new versions of distros.

So my rationale is:
  * build-dependencies is validated when building in the distribution's context. There we can use an older version of pylint/pycodestyle.
  * we use pip instead of python packages during our private builds, thereby ensuring that we are warning free with the latest/greatest.

furiel: Ok I see this will not cause problem. Binary api problems will not occur because we do not use <= relation in the versions for now, and we need to build on the same distribution version anyway, that we target the deb packages. And if build dependency is missing, we will see a build error anyway.

kira-syslogng:
success

dbld/rules
furiel: If we want jenkins to use this line, then this should be parametrized somehow. As we do not want to drop support for v2, jenkins should do both.

bazsi: You can parametrize it using the CONFIGURE_ADD variable that you can pass via the make command line or even as environment variable.

Or you can override CONFIGURE_OPTS as a whole.

make -f dbld/rules CONFIGURE_OPTS=whatever

furiel: Ok fine for me. Thanks for explaining. I have no other comments.

bazsi:
I have now changed the --with-python argument to default to 2. with that I think this can be merged.

kira-syslogng:
success

bazsi:
I have rebased now to master, and with that I am merging this.

kira-syslogng:
success

bazsi:
everything is green, I am merging this with an approval from @furiel

mitzkia:
Sorry for the late response. 
The `bootstrap` uses a `make python-venv` command, but the `python-venv` target did found nowhere. As I remember previously it was defined in `Mk/Makefile.am`, but this patch set did not contains it.

mitzkia:
Issue1: `$ dbld/rules shell`  (for newly built xenial image) is stuck.
Error:
```
...
dbld/rules:95: recipe for target 'shell-xenial' failed
```

bazsi:
Right, I separated that out into another branch...

-- 
Bazsi

bazsi:
> Issue1: $ dbld/rules shell (for newly built xenial image) is stuck.
> Error:
>
> ...
> dbld/rules:95: recipe for target 'shell-xenial' failed
>
> I am using this every day. But I am rebuilding the image to make sure it
works.

mitzkia:
Thanks. The found issue was because of not clean env, after started from beginning it worked

PR #1938, opened 2018-03-23
pkg-config files correction

Kokan:
There was two issue I found with the `pc` files.
* syslog-ng-add-contextual-data.pc is not required anymore
* syslog-ng.pc should list the `libevtlog` library, as it is needed for linking.


bazsi:
The libevtlog lib should be folded into libsyslog-ng.so

However i understand if that's deemed more complicated that the pc file change. I just wanted to state this so everyone is on the same page.

kira-syslogng:
success

PR #1939, opened 2018-03-23
[WIP] logmsg: processed timestamps and macros(prefix P_)

Kokan:
The `C_` prefixed timestamps always evaluated as the actual time. This gives the possibility to get the time when the destination handles the message (or any `LogPipe` expands macro), as there could be a difference between the received and processed time. But this timestamp is not stored into the internal `LogMessage` structure, in case of the message is serialized and deserialized this information is lost.

The serialized message has a space for one timestamp. This reuses that space to store a new timestamp `PROCESSED`. That timestamp should contains the time when the destination handles the `LogMessage` (start processing), but it should be updated before macro expansion.

IMPORTANT
The `PROCESSED` timestamp is not updated for the destination, that is something the destination should take care of if it has to store that time as well.

This timestamp also exposed as macro the same way as received, sent timestamp related macros (see `R_`, `S_`).
The corresponding prefix is `P_`. (For example: `P_UNIXTIME`.)


kira-syslogng:
success

kira-syslogng:
success

kira-syslogng:
success

lib/logmsg/logmsg.c
```diff
@@ -1089,6 +1089,7 @@ log_msg_init(LogMessage *self, GSockAddr *saddr)
   self->timestamps[LM_TS_RECVD].tv_usec = tv.tv_usec;
   self->timestamps[LM_TS_RECVD].zone_offset = get_local_timezone_ofs(self->timestamps[LM_TS_RECVD].tv_sec);
   self->timestamps[LM_TS_STAMP] = self->timestamps[LM_TS_RECVD];
+  self->timestamps[LM_TS_PROCESSED] = self->timestamps[LM_TS_RECVD];
```
furiel: Let's initialize this to something else, like -1 so not to suggest it is an alias. Other than that, this patch looks good to me.

MrAnno:
Hi,

We've just merged a CI-related pull request into our master branch.
Since the current Travis CI checks in your PR are outdated, I'd like to ask you to rebase against master.

kira-syslogng:
success

bazsi:
What we've discussed is that we should handle P_ macros specially, as we can't change that value in the destination drivers as the LogMessage may be shared accross multiple drivers. So the "P" slot in the timestamp fields should only be used once the message has been serialized and the deserialized.

This means that we should:
* initialize the processed timestamp to an extremal value (e.g. -1)
* the P_* macros should resolve to the R_ timestamp as long as the P timestamp is this extremal value
* log containers that encapsulate the serialized format (e.g. disk queue and logstore) are free to write the current time instead of the extremal value.
* consumers of the serialized format are free to reuse the processed timestamp, once it is guaranteed that the LogMessage instance is not shared.

This way:
* if the message is sitting on a disk queue and then deserialized, we can see when it was serialized into the queue
* if the message is sitting in a logstore, we can see when it was written there
* if there's an extremal consumer (e.g. the indexer process) over a named pipe, that can do the same.

kira-syslogng:
Build FAILURE

Issue #1909, opened 2018-03-02
frac-digits ignored in amqp destination


abraxxa:
# syslog-ng

## Version of syslog-ng 

```
syslog-ng 3.8.1
Installer-Version: 3.8.1
Revision: 3.8.1-10
Module-Directory: /usr/lib/syslog-ng/3.8
Module-Path: /usr/lib/syslog-ng/3.8
Available-Modules: sdjournal,affile,afprog,afsocket,afuser,basicfuncs,cef,confgen,cryptofuncs,csvparser,date,dbparser,disk-buffer,kvformat,linux-kmsg-format,pseudofile,syslogformat,system-source,afsql,afmongodb,json-plugin,add-contextual-data,afamqp,geoip-plugin,graphite,mod-python,redis,riemann,afsmtp,afstomp
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: off
```

## Platform

Debian 9.3

## Debug bundle

syslog-ng-debun -r isn't bundled in the Debian syslog-ng-core package

# Issue
When frac-digits(6) is configured in an amqp output the R_ISODATE macro still doesn't contain milli- and microseconds.
It works when configured globally.

## Steps to reproduce

1. configure syslog-ng as shown below

2. start syslog-ng

3. check (unencrypted) AMQP packets using e.g. tcpdump, tshark or wireshark

## Configuration

```
destination d_amqp {
    amqp(
         body('{"timestamp":"${R_ISODATE}","received_from":"${SOURCEIP}","message":"${MESSAGE}"}')
        frac-digits(6)
    );
}

```

## Input and output logs (if possible)
none

gaborznagy:
Hello @abraxxa!

Thank you for reporting the above issue.
I have checked the source code and your report is valid:
`frac-digit()` in global option works while in driver options is not.

Option `frac-digits` is not the only thing that is not working, but:
`ts-format()`, `time-zone()`, `send-time-zone()`, `local-time-zone()`, `on-error()` is not working either.

The reason is that the set `LogTemplateOptions` are not used in ampq.
It has to be fixed.

gaborznagy:
Fixed by #1944.

gaborznagy:
@abraxxa I have tested it too and hence closing the ticket.

PR #1944, opened 2018-03-28
afamqp: LogTemplateOptions fix

definit3:
Fixes issue #1909 . LogTemplateOptions were ignored in the driver because it was not called by log_template_format() function.

### DIFF
--- a/modules/afamqp/afamqp.c
+++ b/modules/afamqp/afamqp.c
@@ -540,13 +540,13 @@ afamqp_worker_publish(AMQPDestDriver *self, LogMessage *msg)
   props.delivery_mode = self->persistent;
   props.headers = table;
 
-  log_template_format(self->routing_key_template, msg, NULL, LTZ_LOCAL,
+  log_template_format(self->routing_key_template, msg, &self->template_options, LTZ_LOCAL,
                       self->super.seq_num,
                       NULL, routing_key);
 
   if (self->body_template)
     {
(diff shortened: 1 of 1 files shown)


kira-syslogng:

gaborznagy:

gaborznagy:
@kira-syslogng ok to test

pzoleex:
@kira-syslogng ok to test

kira-syslogng:
success

bazsi:
good catch. and the patch looks good to me.

While reviewing I just had one question that came up. We seem to be using the send-time-zone() when formatting name-value pairs and the local-time-zone() when formatting the body and the routing key.

I think the routing key is ok this way, we generally use the local timezone in formatting filenames, but I am not sure about the body. I am not an amqp expert so I don't really know whether body is in the same category as the value pairs we are generating, but my gut feeling is that it should be.

Do you know the semantic difference between a body and "props"? What is you opinion?

bazsi:
Also, can you pls add a more descriptive title for the patch? Like "fix the application of template options to body & routing_key".

Thanks, and thanks for submitting the patch to us.

bazsi:
Anyhow the patch is good as is, but I'd love to see the timezone for body() clarified as well.

definit3:
@bazsi  Even I am not that much familiar, let me look it into it in detail.

PR #1946, opened 2018-03-28
Fix scratch-buffer memory usage in threaded destinations

gaborznagy:
There is an explicit scratch_buffer_gc call in `log_threaded_dest_driver_do_work()`
(see `main_loop_worker_run_gc`), 
but the insert() method can be called several times in one do_work call.
If `scratch_buffers_alloc()` is used in the insert method and have been called several times then the memory usage will be increased.

With this patch the used scratch buffers will be reclaimed after each insert() call.

Question:
- Why should the current `main_loop_worker_run_gc` remain in do_work() function?

### DIFF
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -25,6 +25,7 @@
 #include "stats/stats-cluster-logpipe.h"
 #include "logthrdestdrv.h"
 #include "seqnum.h"
+#include "scratch-buffers.h"
 
 #define MAX_RETRIES_OF_FAILED_INSERT_DEFAULT 3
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
success

bazsi:
The explicit_gc() should remain there, as we only reset to the last know location and there might still be other calls to scratch_buffers_alloc() outside of insert()

This looks good to me. I've checked the very same possible bug at least twice, but didn't find that insert() was called multiple times within a same do_work invocation. Strange :) anyhow good catch.

gaborznagy:
@MrAnno @bazsi : sorry for pushing to the branch after review. Can you please recheck?

* I have reworded the threaded dest. commit (missed the `logthrdestdrv` word :) )
* really small refactor: replaced `scratch_buffers_explicit_gc()` to `main_loop_worker_run_gc()` in main_loop_io_worker.

MrAnno:
Still :+1: from my side.
I like the new refactor patch.

bazsi:
Agreed.

PR #1947, opened 2018-03-28
Fix default network drivers

### LINKED ISSUE #1945: syslog-ng can not started with default-network-drivers()
Reproduction:
 * syslog-ng was built from source: commit 778cfa3724364
 * with enabling the following line in generated default config
https://github.com/balabit/syslog-ng/blob/778cfa37243642c079f5d66eeb78720f22bdd6bb/scl/syslog-ng.conf#L27

syslog-ng will use a new `default-network-drivers()` as a source. 
syslog-ng can not started with this new configuration because of the following error:
```
Error setting up TLS session context; tls_error='(null):(null):(null)'
```
This is because the scl did not define a valid key-file() and cert-file() in TLS options. 
Related SCL snippet from: `scl/default-network-drivers/plugin.conf`
https://github.com/balabit/syslog-ng/blob/778cfa37243642c079f5d66eeb78720f22bdd6bb/scl/default-network-drivers/plugin.conf#L69-L72

bazsi:
This should fix #1945 

@lbudai @MrAnno please review the SSL piece, as it might somewhat be questionable, but I am open for suggestions.

### DIFF
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -54,6 +54,7 @@ struct _TLSContext
   GList *trusted_fingerprint_list;
   GList *trusted_dn_list;
   gint ssl_options;
+  gchar *location;
 };
 
 typedef enum
@@ -173,14 +174,16 @@ tls_session_verify(TLSSession *self, int ok, X509_STORE_CTX *ctx)
   /* accept certificate if its fingerprint matches, again regardless whether x509 certificate validation was successful */
   if (ok && ctx_error_depth == 0 && !tls_session_verify_fingerprint(ctx))
     {
-      msg_notice("Certificate valid, but fingerprint constraints were not met, rejecting");
+      msg_notice("Certificate valid, but fingerprint constraints were not met, rejecting",
+                 tls_context_format_location_tag(self->ctx));
       return 0;
     }
 
   X509 *current_cert = X509_STORE_CTX_get_current_cert(ctx);
   if (ok && ctx_error_depth != 0 && (X509_get_extension_flags(current_cert) & EXFLAG_CA) == 0)
     {
-      msg_notice("Invalid certificate found in chain, basicConstraints.ca is unset in non-leaf certificate");
+      msg_notice("Invalid certificate found in chain, basicConstraints.ca is unset in non-leaf certificate",
+                 tls_context_format_location_tag(self->ctx));
       X509_STORE_CTX_set_error(ctx, X509_V_ERR_INVALID_CA);
       return 0;
     }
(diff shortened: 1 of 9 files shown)


kira-syslogng:
success

lib/tlscontext.c
```diff
@@ -602,15 +602,15 @@ _are_key_and_cert_files_accessible(TLSContext *self)
 }
 
 static gboolean
-_client_key_and_cert_files_are_not_specified(TLSContext *self)
+_key_and_cert_files_are_not_specified(TLSContext *self)
 {
-  return self->mode == TM_CLIENT && (!self->key_file && !self->cert_file);
```
MrAnno: Omitting the key in server mode should not be allowed in my opinion.
There is no secure cipher in OpenSSL that works without a key, so every connection attempt will fail with an SSL error: "no shared cipher".

I understand that this is required to fix `default-network-drivers` (since we can't conditionally remove a source from the config based on the value of a SCL block parameter). Adding such a feature to the grammar would be a much bigger job, so I can reconcile with this patch :smile: .

MrAnno:
Omitting the key in server mode should not be allowed in my opinion.
There is no secure cipher in OpenSSL that works without a key, so every connection attempt will fail with an SSL error: "no shared cipher".

I understand that this is required to fix `default-network-drivers` (since we can't conditionally remove a source from the config based on the value of a SCL block parameter). Adding such a feature to the grammar would be a much bigger job, so I can reconcile with this patch :smile: .

MrAnno:
Just a question:
Why don't we remove the TLS source from `default-network-drivers`? For me, default means "it can work without additional configuration". I know it's a weird question, but I'm interested in your opinion about this.

gaborznagy:
@bazsi, I've put the WIP flag on the PR according to the today's discussion (adding some warning messages, but correct me if I don't remember correctly).
You have also explained why should the tls transport included in `default-network-drivers`, i.e. you  would not prefer to have separate `default-network-drivers` for TLS. And also why should the empty `tls()` block accepted in the config.

mitzkia:

kira-syslogng:
success

bazsi:
I have now rebased this against master and also addressed all the comments.

It also improves the TLS related error handling so that error messages include the tls() location in the configuration file.

I've removed the WIP mark, so once it is green  it should be ready to go in and fix #1945

kira-syslogng:
success

kira-syslogng:
success

MrAnno:
Looks really good. 👍

lib/tlscontext.c
```diff
@@ -321,30 +328,39 @@ tls_session_free(TLSSession *self)
...
 {
   if (!fname)
     return FALSE;
   if (access(fname, R_OK) < 0)
     {
-      msg_error("Error opening TLS file",
+      msg_error("Error opening TLS related key or certificate file",
                 evt_tag_str("filename", fname),
-                evt_tag_errno("error", errno));
+                evt_tag_errno("error", errno),
```
furiel: This is a general problem, but it is not safe to use errno this way. The malloc, strdup called from evt_tag_str can overwrite it. As the parameters are evaluated in any order, it is not even enough to move it to the first parameter.

As the original code had this defect too, and probably we will not run into this from practical point of view, I leave just a note here, and I create an issue to fix these in one swipe within the code. Maybe we can even assert if errno is passed directly to evt_tag_errno.

furiel:
@bazsi @fekete-robert as I see default-network-sources is part of the code since 3.12, but I do not see it in the documentation. Should we write something about it?

PR #1950, opened 2018-03-29
Dbld refinements2

bazsi:
Another set of dbld improvements that I've encountered while doing all my development work using the docker containers. Mostly small stuff. Tarball generation should not be risky, The only change that could affect that is that I've added joe and netcat to the apt based images.

### DIFF
--- a/dbld/images/required-apt/all.txt
+++ b/dbld/images/required-apt/all.txt
@@ -31,19 +31,15 @@ tzdata
 # required for make/cmake
 make
 cmake
-
-# python stuff
-virtualenv
+astyle
 
 # required for make check
 libxml2-utils
 
-# other tools
-gdb
+# required for make dist
 git
-lsof
-strace
-vim
+
+# interactive tools
 less
 
 # debian packaging tools
--- a/dbld/rules
+++ b/dbld/rules
@@ -1,6 +1,6 @@
 #!/usr/bin/make -f
 
-IMAGES=centos6 centos7 jessie artful xenial trusty
+IMAGES=centos6 centos7 jessie artful xenial trusty devshell
 DEFAULT_IMAGE ?= xenial
 DEFAULT_DEB_IMAGE=$(DEFAULT_IMAGE)
 DEFAULT_RPM_IMAGE=centos7
--- a/dbld/rules
+++ b/dbld/rules
@@ -46,7 +46,8 @@ help:
 	@echo "Development:"
 	@echo "  bootstrap: bootstrap source from a git clone (e.g. autogen.sh and friends)"
 	@echo "  shell: start a shell in the specified OS"
-	@echo "  login: start a 2nd shell (as root) alongside an already running one"
+	@echo "  login: start a 2nd shell alongside an already running one"
+	@echo "  root-login: start a 2nd shell (as root) alongside an already running one"
 	@echo "  tarball: generate a tarball in a controlled environment (e.g. docker)"
 	@echo "  release: generate a release deb package in a controlled environment (e.g. docker)"
 	@echo ""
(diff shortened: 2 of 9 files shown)


kira-syslogng:
success

kira-syslogng:
success

mitzkia:
I am also thinking about a dbld related PR. It is about to make docker images thinner. And make clear which package is installed for what purpose. Like now: packages for automake/packages for configure, but deeper detailed in case of configure and pip install. Which package is installed for which module support. I think for now our images contains to many packages, and already there are a few of them where I do not know the exact purpose. Our built image is around 1GB, which  I think is quite much. If we follow the current "principle" (just adding a new packages for installation) where will it go? I think where we can make a decision is in "other tools" and "debian packaging tools" sections. What do you think about that to build a new (separate) image for debian packaging on top of a "basic dev image" (which we generate now).  This whole comment is came from there I saw new package installs like: astyle, netcat, ... Finally I think our images should contain only a strictly required packages for building and installing surely and maybe for debugging syslog-ng.

dbld/images/required-apt/all.txt
```diff
@@ -31,6 +31,7 @@ tzdata
 # required for make/cmake
 make
 cmake
+astyle
```
mitzkia: When do we depend on this package?

dbld/rules
```diff
@@ -59,6 +59,9 @@ bootstrap: bootstrap-$(DEFAULT_IMAGE)
 bootstrap-%: setup
 	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -e CONFIGURE_OPTS="$${CONFIGURE_OPTS:-$(CONFIGURE_OPTS)}" -ti  balabit/syslog-ng-$* /source/dbld/bootstrap
 
+make-%: setup
+	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -ti  balabit/syslog-ng-$(DEFAULT_IMAGE) /source/dbld/make $(MAKE_ARGS) $*
```
mitzkia: I have two comments for this.
1, Is it enough to execute only a "make" on source? make install is not needed?
2, I can not find the definition of MAKE_ARGS (or is it a general env. variable)

dbld/rules
```diff
@@ -103,7 +106,7 @@ image-%:
 login: login-$(DEFAULT_IMAGE)
 login-%: setup
 	container=`$(DOCKER) ps | grep syslog-ng-$* | head -1 | cut -d ' ' -f1`; \
-	$(DOCKER) exec -ti  $$container /bin/bash
+	$(DOCKER) exec -ti  $$container gosu $(shell whoami) /source/dbld/shell
```
mitzkia: As I understand the login target was introduced for getting a root shell inside a container, but with this change it will modified/broke.

dbld/images/required-apt/all.txt
bazsi: when we do make style-check

dbld/rules
bazsi: you can call the makefile using:

```
$ dbld/rules make-install MAKE_ARGS=-j4
```

make works so that if a variable is not defined it is populated from the environment or the command line.

The above rule does not support make without any arguments, but the default target for make is usually "all", so you can achive the same with:

```
$ dbld/rules make-all
```

The whole thing is about being able to compile right from the source directory, without having to enter the docker container using "shell".

dbld/rules
bazsi: No, login was created so you can enter the same container multiple times. If was only a side effect that it was doing it as root. But I can agree that a root shell would be useful. Maybe a separate root-login target?

I can't remember why we can't do sudo inside the container though, that would be more handy. I think it was @szemere or @furiel who said that it was not secure or something.

bazsi:
> I am also thinking about a dbld related PR. It is about to make docker images thinner. And make clear which package is installed for what purpose. Like now: packages for automake/packages for configure, but deeper detailed in case of configure and pip install. For example which package is installed for which module support. I think for now our images installs to many packages, and already there are a few of them where I do not know the exact purpose. Our built image is around 1GB, which I think is quite much. If we follow the current "principle" (just adding a new packages for installation) where will it go? I think where we can make a decision is in "other tools" and "debian packaging tools" sections. What do you think about that to build a new (separate) image for debian packaging on top of a "basic dev image" (which we generate now). This whole comment is came from there I saw new package installs like: astyle, netcat, ... Finally I think our images should contain only a strictly required packages for building and installing surely and maybe for debugging syslog-ng.

All I did was exactly to make it comfortable to work on syslog-ng right in the image. I am not sure how much thinner we can make it, or if it is that important. The runtime package should be small if possible (maybe even built one on alpine linux or something), but the dev images can be larger than that.

I installed astyle, because make style-format/check uses it. I've used netcat, because I wanted to connect to syslog-ng after I started it up within the container.

bazsi:
I have now added a separate "devshell" image for interactive development and dropped various debugging tools from the build image.

kira-syslogng:
success

kira-syslogng:
success

bazsi:
I have now manually tested a number of use-cases (image creation, deb compilation, devshell, login, root-login, etc), added new commands to help. 

Addressed the concern of @mitzkia with regards to image size. With that and the approval from @furiel  I am merging this.

kira-syslogng:
success

PR #1951, opened 2018-03-29
add hook-commands module

bazsi:
This module adds a new plugin that is capable of injecting itself into both
source and destination drivers.  It makes it possible to execute external
programs when they are initialized or torn down.
    
Here's an example that opens an iptables port automatically as syslog-ng is
started/stopped.
    
Assumption: the LOGCHAIN chain is part of a larger ruleset that routes
traffic to it.  Whenever the syslog-ng created rule is there, packets can
flow, otherwise the port is closed.  This is a silly example, but I think
you get the idea.
```
source {
  network(transport(udp)
              hook-commands(setup("iptables -I LOGCHAIN 1 -p udp --dport 514 -j ACCEPT")
                          teardown("iptables -D LOGCHAIN 1")));
};
```    
It might even be combined with something like salt for using more
declarative actions instead of simply coding add/remove transactions which
might break if syslog-ng dies.

### DIFF
--- a/lib/logpipe.h
+++ b/lib/logpipe.h
@@ -214,16 +214,31 @@ struct _LogPipe
 {
   GAtomicCounter ref_cnt;
   gint32 flags;
+
+  /* moving queue to the front should improve performance somewhat, although
+   * I didn't measure it at all.  Rationale: the pointer pointing to this
+   * LogPipe instance is resolved right before calling queue and the
+   * colocation of flags and the queue pointer ensures that they are on the
+   * same cache line, and since the CPU returns memory reads sequentially,
+   * it is among the first values that are read ahead.  It might need to be
+   * measured sometime, but this method is probably the single most often
+   * called virual method, so I felt, that even though the optimization is
+   * probably premature without explicit tests, I can do this simply by
+   * believing it helps :) */
+
+  void (*queue)(LogPipe *self, LogMessage *msg, const LogPathOptions *path_options);
+
   GlobalConfig *cfg;
   LogExprNode *expr_node;
   LogPipe *pipe_next;
   StatsCounterItem *discarded_messages;
   const gchar *persist_name;
-
-  void (*queue)(LogPipe *self, LogMessage *msg, const LogPathOptions *path_options);
   gchar *plugin_name;
+
+  gboolean (*pre_init)(LogPipe *self);
   gboolean (*init)(LogPipe *self);
   gboolean (*deinit)(LogPipe *self);
+  void (*post_deinit)(LogPipe *self);
 
   const gchar *(*generate_persist_name)(const LogPipe *self);
 
@@ -269,6 +284,8 @@ log_pipe_init(LogPipe *s)
 {
   if (!(s->flags & PIF_INITIALIZED))
     {
+      if (s->pre_init && !s->pre_init(s))
+        return FALSE;
       if (!s->init || s->init(s))
         {
           s->flags |= PIF_INITIALIZED;
@@ -287,6 +304,9 @@ log_pipe_deinit(LogPipe *s)
       if (!s->deinit || s->deinit(s))
         {
           s->flags &= ~PIF_INITIALIZED;
+
+          if (s->post_deinit)
+            s->post_deinit(s);
           return TRUE;
         }
       return FALSE;
--- a/lib/logpipe.h
+++ b/lib/logpipe.h
@@ -169,11 +169,13 @@
  *   The way to override a method by an external object is as follows:
  *
  *     - it should save the current value of the method address (for
- *       example "queue" for the queue method), and the associated
- *       user_data pointer (queue_data in this case)
+ *       example "queue" for the queue method)
  *
  *     - it should change the pointer pointing to the relevant method to
  *       its own code (e.g. change "queue" in LogPipe)
+ *
+ *     - once the hook is invoked, it should take care about calling the
+ *       original function
  **/
 
 struct _LogPathOptions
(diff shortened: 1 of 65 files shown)


kira-syslogng:
failure

furiel:
I would like to ask you a few questions about the design. Sorry for asking earlier than checking the patchset itself, I just want to make sure and I understand the concept here before delving into the review. Maybe it would worth updating the description with the answers to help the documentation team.

- What happens if `setup` fails? Syslog-ng will not start? That would be unfortunate, we were giving special focus during password protected ssl feature so that those drivers that are not affected with the secret store, in our case with setup, that those should start. Logging is a system critical service, at least file file sources/local logging should start even if an iptables fails.

- What happens if teardown fails? In some cases we should be very loud about it, for example when we cannot close the firewall behind us. How can we notify the user?

- What happens during reload? Shall we setup and teardown during each reload? In some cases maybe we should, but in case of iptables manipulation, I think modifying iptables just because of a log rotation does not seem optimal.

- What if syslog-ng crashes? Is there any way we can call the pending teardowns?

- Should setup/teardown be synchron/asynchron? Can we hold the mainloop for a long time, for example when a complex salt script is executed in them?

- Should we support shared/global setup or teardown? If two drivers need the same setup mechanism, do they need to execute it two times? This implicitely enforces users to use idempotent setup and teardown functions: during reload, two setup and two teardown will be called after each other.

- Such setup/teardown feature seems the responsibility of a service management system. For example in systemd, users can specify complex dependency on the setup teardown mechanism. Who are the target users? What is the benefit configuring syslog-ng than configuring it in systemd?

kira-syslogng:
failure

lbudai:
@bazsi : Based on what I know about this feature, I'm still not convinced that this kind of resource management is the responsibility of syslog-ng.

I think we are not prepared for supporting external software components from packaging point of view.
Which means that we cannot deliver SCLs that are depending on external softwares.
I also think that even if we could support these kind of SCLs the config snippets would not be platform/OS independent. So everyone should write their own variant of config snippets for the same problem.

Could we talk about it this IRL?

bazsi:
I have now rebased it against the current master, resolving the conflicts.

Couple of answers to a comment above:

> I would like to ask you a few questions about the design. Sorry for asking earlier than checking the patchset itself, I just want to make sure and I understand the concept here before delving into the review. Maybe it would worth updating the description with the answers to help the documentation team.
> 
>     * What happens if setup fails? Syslog-ng will not start? That would be unfortunate, we were giving special focus during password protected ssl feature so that those drivers that are not affected with the secret store, in our case with setup, that those should start. Logging is a system critical service, at least file file sources/local logging should start even if an iptables fails.

I think that SSL is a separate case. There if the key is encrypted, we would _always_ run into a problem. Whereas as in this case, the script being executed is expected to succeed in most cases.

If the script enables a critical portion of a log source (e.g. enables audit events or configures osquery to perform regular queries), then I don't see too much point in starting up and behave as if nothing happened.

I understand that logging is a critical service, but failing silently is not good practice. Anyhow, we should probably discuss this in real life, once I am in the office (probably Wed?)

> 
>     * What happens if teardown fails? In some cases we should be very loud about it, for example when we cannot close the firewall behind us. How can we notify the user?

I am thinking about redirecting our own stderr to the internal log somehow instead of sending it to /dev/null. That should solve the issue.

btw: the iptables example was just an example, I wouldn't consider it that important use-case. If someone wants to manage something as complex as iptables, I would probably use a salt-like declarative solution.

But now as I think about it, maybe it would be worthwhile to also add a time-based hook, e.g. one that runs every once in a while. This could take care about startup failing due to a missing dependency, OR for that matter could solve the logrotate problem as well. Think about this:

file("/var/log/remote/$HOST/messages" hook-commands(cron("*/5 * * * * * logrotate /whatever"))

> 
>     * What happens during reload? Shall we setup and teardown during each reload? In some cases maybe we should, but in case of iptables manipulation, I think modifying iptables just because of a log rotation does not seem optimal.

I have added a startup/shutdown hook as well.

> 
>     * What if syslog-ng crashes? Is there any way we can call the pending teardowns?

not really, I wouldn't want to track their state. That's exactly what something like salt could do.

> 
>     * Should setup/teardown be synchron/asynchron? Can we hold the mainloop for a long time, for example when a complex salt script is executed in them?

I would do it synchronously. People shouldn't do expensive things there. If they do, they'd possibly shoot themselves on the foot, but that's ok.

> 
>     Should we support shared/global setup or teardown? If two drivers need the same setup mechanism, do they need to execute it two times? This implicitely enforces users to use idempotent setup and teardown functions: during reload, two setup and two teardown will be called after each other.

Again, this is exactly what salt could do instead of us. I am not sure if salt can be applied on a local host, in contrast to building on its as a distributed management system, but I guess it can be.

> 
>     Such setup/teardown feature seems the responsibility of a service management system. For example in systemd, users can specify complex dependency on the setup teardown mechanism. Who are the target users? What is the benefit configuring syslog-ng than configuring it in systemd?
> 

Like I said, syslog-ng alone can't distribute systemd units.

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
success

lbudai:
@fekete-robert : FYI

lib/logpipe.h
```diff
@@ -214,16 +214,31 @@ struct _LogPipe
...
+
+  /* moving queue to the front should improve performance somewhat, although
+   * I didn't measure it at all.  Rationale: the pointer pointing to this
+   * LogPipe instance is resolved right before calling queue and the
+   * colocation of flags and the queue pointer ensures that they are on the
+   * same cache line, and since the CPU returns memory reads sequentially,
+   * it is among the first values that are read ahead.  It might need to be
+   * measured sometime, but this method is probably the single most often
+   * called virual method, so I felt, that even though the optimization is
+   * probably premature without explicit tests, I can do this simply by
```
furiel: I tend to agree this might reduce cache use, because the skipped fields are five pointers = 40bytes is relately high compared to an i7 cache line of 64byte, so the chance they end up on different cache lines are not negliglible.

With similar reasoning, do you think it would worth move `pipe_next` upwards as well? log_pipe_forward is called at the same time as ->queue just on the other branch on the if.

At the same time, I doubt we could measure the difference: we are talking about cache sizes of 256kb of L1 and 4mb of L2, so using one extra cache line should not mean too much difference. But to be safe, I do not problem with the rearrangement, it can only improve the performance.

However, this comment is in a **very** bad position in the code. It looks as if it would be a `docstring` explaining how LogPipe works, but it is about a very low level optimization, that should not grab the focus for typical investigations. I was just wandering here with gqueue push pop context in my head, and I could only realize this is about rearranging fields only after I found the patch with git blame. We should delete this comment as soon as possible. It would have been a good comment in the commit message though...

PR #1952, opened 2018-03-29
Error reporting on misspelled block args

bazsi:
This branch makes the __VARARGS__ functionality in SCL blocks explicit, as currently it is very easy to mis-spell an argument, which is not logged in any way. It changes the syntax slightly:

```
block source something(foo() bar() ...) {
};
```

The ... at the end of the argument list tells syslog-ng that this macro indeed wants to accept __VARARGS__, thus any name-value pair can be passed without validation.

### DIFF
--- a/lib/cfg-block.c
+++ b/lib/cfg-block.c
@@ -61,6 +61,50 @@ cfg_block_format_name(CfgBlockGenerator *s, gchar *buf, gsize buf_len)
   return buf;
 }
 
+/* token block args */
+
+static void
+_validate_args_callback(gpointer k, gpointer v, gpointer user_data)
+{
+  CfgArgs *defs = ((gpointer *) user_data)[0];
+  const gchar *reference = ((gpointer *) user_data)[1];
+  gboolean *problem_found = ((gpointer *) user_data)[2];
+
+  if ((!defs || cfg_args_get(defs, k) == NULL))
+    {
+      if (cfg_args_is_accepting_varargs(defs))
+        {
+          msg_verbose("Unknown argument, adding it to __VARARGS__",
+                      evt_tag_str("argument", k),
+                      evt_tag_str("value", v),
+                      evt_tag_str("reference", reference));
+        }
+      else
+        {
+          msg_error("Unknown argument specified to block reference",
+                    evt_tag_str("argument", k),
+                    evt_tag_str("value", v),
+                    evt_tag_str("reference", reference));
+          *problem_found = TRUE;
+        }
+    }
+}
+
+static gboolean
+_validate_args(CfgArgs *self, CfgArgs *defs, const gchar *reference)
+{
+  gboolean problem_found = FALSE;
+  gpointer validate_params[] = { defs, (gchar *) reference, &problem_found };
+
+  cfg_args_foreach(self, _validate_args_callback, validate_params);
+
+  if (problem_found)
+    return FALSE;
+
+  return TRUE;
+}
+
+
 /*
  * cfg_block_generate:
  *
@@ -75,15 +119,19 @@ cfg_block_generate(CfgBlockGenerator *s, GlobalConfig *cfg, CfgArgs *args, GStri
   gchar *value;
   gsize length;
   GError *error = NULL;
+  gchar buf[256];
 
-  cfg_args_set(args, "__VARARGS__", cfg_args_format_varargs(args, self->arg_defs));
+  if (!_validate_args(args, self->arg_defs, reference))
+    return FALSE;
+
+  if (cfg_args_is_accepting_varargs(self->arg_defs))
+    cfg_args_set(args, "__VARARGS__", cfg_args_format_varargs(args, self->arg_defs));
 
   value = cfg_lexer_subst_args_in_input(cfg->globals, self->arg_defs, args, self->content, -1, &length, &error);
   if (!value)
     {
       msg_warning("Syntax error while resolving backtick references in block",
-                  evt_tag_str("context", cfg_lexer_lookup_context_name_by_type(self->super.context)),
-                  evt_tag_str("block", self->super.name),
+                  evt_tag_str("block_definition", cfg_block_generator_format_name(s, buf, sizeof(buf))),
                   evt_tag_str("error", error->message),
                   NULL);
       g_clear_error(&error);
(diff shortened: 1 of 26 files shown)


bazsi:
btw, this depends on the improve-error-handling branch, so that needs to go in first, I just wanted to throw this into the mix.

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
success

bazsi:
rebased against master, remove WIP tag, this should be ok to review.

kira-syslogng:
success

lib/cfg-lexer.c
```diff
@@ -946,18 +946,22 @@ cfg_lexer_lex(CfgLexer *self, YYSTYPE *yylval, YYLTYPE *yylloc)
...
             {
+              cfg_block_generator_format_name(gen, buf, sizeof(buf));
+
+              if (gen->suppress_backticks)
+                success = cfg_lexer_include_buffer_without_backtick_substitution(self, buf, result->str, result->len);
+              else
+                success = cfg_lexer_include_buffer(self, buf, result->str, result->len);
+              g_string_free(result, TRUE);
+
               goto relex;
```
furiel: The original code conditionally goto-ed to relex, when `g_lexer_include_buffer_without_backtick_substitution` and `cfg_lexer_include_buffer` returned successful. Can we skip to check the return values now?

lib/cfg-block.c
```diff
@@ -61,6 +61,50 @@ cfg_block_format_name(CfgBlockGenerator *s, gchar *buf, gsize buf_len)
...
 
+/* token block args */
+
+static void
+_validate_args_callback(gpointer k, gpointer v, gpointer user_data)
+{
+  CfgArgs *defs = ((gpointer *) user_data)[0];
+  const gchar *reference = ((gpointer *) user_data)[1];
+  gpointer *problem_found = &((gpointer *) user_data)[2];
+
```
furiel: As I see problem_found is write-only now? What is the purpose? Is an early return missing in this function?

lib/cfg-block.c
```diff
@@ -61,6 +61,50 @@ cfg_block_format_name(CfgBlockGenerator *s, gchar *buf, gsize buf_len)
...
+                    evt_tag_str("reference", reference));
+          *problem_found = (gpointer) TRUE;
+        }
+    }
+}
+
+static gboolean
+_validate_args(CfgArgs *self, CfgArgs *defs, const gchar *reference)
+{
+  gpointer validate_params[] = { defs, (gchar *) reference, NULL };
```
furiel: Please use FALSE instead of NULL.

lib/cfg-block.c
furiel: Ok I see it is used on line 100 with `if (validate_params[2])`. The early early return might still be useful.

lib/cfg-block.c
```diff
@@ -53,11 +53,11 @@ cfg_block_format_name(CfgBlockGenerator *s, gchar *buf, gsize buf_len)
 {
   CfgBlock *self = (CfgBlock *) s;
 
-  g_snprintf(buf, buf_len, "block %s %s (at %s:%d:%d)",
+  g_snprintf(buf, buf_len, "block %s %s() at %s:%d",
```
furiel: () is left for separator purposes?

kira-syslogng:
success

lib/cfg-block.c
bazsi: what do you mean on early return? the entire function is just one if statement, the iteration happens in the foreach which uses this as a callback. You think that the return would make this more readable?

lib/cfg-block.c
bazsi: I've changed this code to pass a gboolean* as third parameter so that I can name the returned value, which makes the code easier to read. I'll push this in a moment.

lib/cfg-block.c
bazsi: In the documentation and log messages we use a convention that a paren is there at the end of the driver, e.g. unix-stream() instead of unix-stream.

This just follows that convention.

lib/cfg-block.c
furiel: I was thinking if the previous iteration ran into an error, then we could skip the execution of this function if we return when problem_found is true.

Maybe it would be better use g_hash_table_find instead of g_hash_table_foreach to avoid unnecessary calls. With g_hash_table_find, probably we could eliminate the problem_found variable.

Though I agree this is not a performance heavy path, so we can stick with anything that is more readable than optimal.

bazsi:
I think I have now addressed the concerns, except for the "early return" that @furiel requested and I didn't completely understand. I've also rebased to master.

One thing that occurred me is if we want to do this selectively on version number, @fekete-robert @czanik what do you think?

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/cfg-block.c
```diff
@@ -61,6 +61,50 @@ cfg_block_format_name(CfgBlockGenerator *s, gchar *buf, gsize buf_len)
...
+{
+  CfgArgs *defs = ((gpointer *) user_data)[0];
+  const gchar *reference = ((gpointer *) user_data)[1];
+  gboolean *problem_found = ((gpointer *) user_data)[2];
+
+  if ((!defs || cfg_args_get(defs, k) == NULL))
+    {
+      if (cfg_args_is_accepting_varargs(defs))
+        {
+          msg_verbose("Unknown argument, adding it to __VARARGS__",
```
gaborznagy: My understanding is that here we cannot differentiate between valid driver options and invalid ones.
We can only know if the option is defined in the scl block or not.

I wonder how this part helps in debugging as if the user configures a lot of options and an invalid option too, he/she will see the parsing error anyway.

bazsi: This makes it possible to see that an argument was indeed not defined explicitly, so even if you have '...' as argument, you'll see (in debug mode) that an argument is using the wildcard mode.

I agree that it could trigger an error later on (when pasting __VARARGS__ into somewhere), but this message can be a useful when __VARARGS__ is pasted to somewhere where it is not actually validated.

Maybe this should be a debug message, instead of a verbose, I'll do that before merging.

kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build SUCCESS

PR #1956, opened 2018-04-02
pragma module-dependency for optional include

### LINKED ISSUE #827: scl improvent: pragma @module-dependency
based on @bazsi's comment ( #678  )
this is a pragma that aborts the current include file if a module is not
available.

@module-dependency json

If this is not met, the entire include file is skipped. or just using the
current @module directive and extend its behaviour (it currently ignores
errors).

Kokan:
A lot of `scl` has a hidden dependency for modules to be present. In case of that `scl` uses a plugin, that is not installed; the syslog-ng is not going to start up even if that specific block is not used in the configuration (but only included).

One of the most common example is the `json`, if the `json` module is not installed, with the default set of `scl`s syslog-ng is not going to start. There are two main workaround exists for this:
* install the module
* remove the scl file

This PR aim to give a solution by creating a new pragma keyword, which can specify module requirement. If the requirement is not specified the syslog-ng is going to skip the current file, thus not causing an issue in case of missing module.

Fixes #827.

### DIFF
--- a/lib/pragma-grammar.ym
+++ b/lib/pragma-grammar.ym
@@ -46,6 +46,7 @@
 %token KW_VERSION
 %token KW_DEFINE
 %token KW_MODULE
+%token KW_REQUIRES
 
 %code {
 
@@ -92,6 +93,7 @@ pragma_stmt
         : KW_VERSION ':' string_or_number               { configuration->parsed_version = __process_version_string($3); free($3); }
 	| define_stmt
         | module_stmt
+        | requires_stmt
 	;
 
 include_stmt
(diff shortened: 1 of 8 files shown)


kira-syslogng:
success

kira-syslogng:
failure

kira-syslogng:
success

kira-syslogng:
success

lbudai:
... and thanks for the PR !

kira-syslogng:
success

lbudai:
@furiel : thanks!

bazsi:
cool \o/


-- 
Bazsi

PR #1957, opened 2018-04-03
Fix stomp template option

gaborznagy:
As discovered in issue #1909 amqp `LogTemplateOptions` were not used in `log_format_template` call.

I have checked syslog-ng code for every template usage and discovered that `stomp()` dest. driver has the same problem, but only for the body part of the message.
Header part works: value_pairs call is using template options.


kira-syslogng:
success

PR #1302, opened 2017-01-04
[WIP] Enable warnings

bazsi:
(edited)
Trivia:
--enable-extra-warnings is **not** a gcc/clang compiler/linker option. It is a syslog-ng configure option (like any --disable-[some-module] options). It is already part of syslog-ng, marked as recommended to be used, and its purpose was to enable three specific kind of warnings during build.

The ultimate goal is of course that there would not be any kind of warnings: to add -Wall compile option, which is part of this patch. The problem is a lot of warnings appear this way (tens of megabytes of log), which is not practical to handle. So we need to reduce the appearing warnings to a manageable size by whitelisting warnings, and the whitelist would be later reduced in other patches.

In short, this patchset replaces the backlisting method on warnings to whitelisting method, and try to enforce the current whitelisting with travis.
(/edited)

This branch rebases #1128 against current master and refines options somewhat and CLAGS as well.

Major changes against #1128:
 * use g_path_basename() instead of a const basename() implementation
 * it sets the default CFLAGS even if --enable-extra-warnings is not defined, e.g -Wshadow is always defined
 * all warning related CFLAGS are set in the Makefile and not in configure.ac

I am not yet sure if travis should run with --enable-extra-warnings, it bloats the build log. My feeling is that we should merge this without the travis part (e.g. build without the extra warnings even in travis). But I am open to ideas.

It would be great if we could merge this at this time, so we don't need to rebase it again in 6 months time :)

PR #1966, opened 2018-04-04
declaration related warnings fixes

### LINKED ISSUE #1962: warning elimination: declaration related warnings

This issue should cover the declaration related warnings.
```
      2 [-Wimplicit-function-declaration]
      1 [-Wnested-externs]
      2 [-Wold-style-declaration]
     16 [-Wstrict-prototypes]
```

mehul-m-prajapati:
Resolves #1962.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,12 @@ before_script:
   - sudo dpkg-reconfigure --frontend noninteractive tzdata
   - ./autogen.sh
   - unset PYTHON_CFLAGS # HACK
+  - if [ "$CC" = "gcc" ]; then
+      EXTRA_WARN="--enable-extra-warnings";
+    fi
+  - if [ "$CC" = "clang" ]; then
+      DISABLE_MONGODB="--disable-mongodb";
+    fi
   - CONFIGURE_FLAGS="
       CFLAGS=-Werror
       --prefix=$HOME/install/syslog-ng
@@ -61,11 +67,9 @@ before_script:
       --enable-manpages
       --with-python=2
       --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl
-      --enable-extra-warnings
+      $EXTRA_WARN
+      $DISABLE_MONGODB
       "
-  - if [ "$CC" = "clang" ]; then
-      CONFIGURE_FLAGS="$CONFIGURE_FLAGS --disable-mongodb" ;
-    fi
   - ./configure $CONFIGURE_FLAGS
 script:
   - . tests/build-log-cflags-propagation.sh;
--- a/Makefile.am
+++ b/Makefile.am
@@ -45,7 +45,11 @@ if ENABLE_EXTRA_WARNINGS
 AM_CFLAGS += \
 	-Wshadow	\
 	-Wno-initializer-overrides \
-	-Wno-error=deprecated-declarations
+	-Wno-error=deprecated-declarations \
+	-Wimplicit-function-declaration \
+      	-Wnested-externs \
+      	-Wold-style-declaration \
+     	-Wstrict-prototypes
 endif
 
 TEST_CFLAGS		= -I$(top_srcdir)/libtest $(AM_CFLAGS)
--- a/configure.ac
+++ b/configure.ac
@@ -1078,8 +1078,8 @@ if test "x$with_ivykis" = "xinternal"; then
 		# the current directory just one below ivykis
 
 		IVYKIS_LIBS="-Wl,${WHOLE_ARCHIVE_OPT} -L\$(top_builddir)/lib/ivykis/src -livykis -Wl,${NO_WHOLE_ARCHIVE_OPT}"
-		IVYKIS_CFLAGS="-I\$(top_srcdir)/lib/ivykis/src/include -I\$(top_builddir)/lib/ivykis/src/include"
-                INTERNAL_IVYKIS_CFLAGS="-I\${includedir}/syslog-ng/ivykis"
+		IVYKIS_CFLAGS="-isystem \$(top_srcdir)/lib/ivykis/src/include -isystem \$(top_builddir)/lib/ivykis/src/include"
+                INTERNAL_IVYKIS_CFLAGS="-isystem \${includedir}/syslog-ng/ivykis"
 
 		# LIBS to use when libtool is not applicable (when linking the main syslog-ng executable in mixed linking mode)
 		IVYKIS_NO_LIBTOOL_LIBS="-Wl,${WHOLE_ARCHIVE_OPT} -L\$(top_builddir)/lib/ivykis/src/.libs -livykis -Wl,${NO_WHOLE_ARCHIVE_OPT}"
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -201,7 +201,7 @@ cfg_load_candidate_modules(GlobalConfig *self)
 void
 cfg_load_forced_modules(GlobalConfig *self)
 {
-  const static gchar *module_list[] =
+  static const gchar *module_list[] =
   {
 #if (!SYSLOG_NG_ENABLE_FORCED_SERVER_MODE)
     "license"
--- a/lib/gprocess.c
+++ b/lib/gprocess.c
@@ -103,6 +103,9 @@ static gboolean stderr_present = TRUE;
 #if SYSLOG_NG_ENABLE_LINUX_CAPS
 static int have_capsyslog = FALSE;
 #endif
+#ifdef SYSLOG_NG_HAVE_ENVIRON
+extern char **environ;
+#endif
 
 /* global variables */
 static struct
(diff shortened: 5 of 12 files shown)


kira-syslogng:

Kokan:
@kira-syslogng ok to test

furiel:
@mehul-m-prajapati about ivykis: can you try to add an exception for that specific warning in lib/Makefile.am for ivykis? I am fine with not applying all these warnings for such modules.
```
lib/ivykis/src/libivykis.la:
	${MAKE} CFLAGS="${CFLAGS} -Wno-error" -C lib/ivykis
```

furiel:
@mehul-m-prajapati about the clang error for unknown warning:
you can try to cherry-pick cac74376a74aea5cdf0feffdb2fcea34a4fd3344 (.travis.yml: only enable extra warnings on gcc) from #1302, which should switch off enable-extra-warnings for clang.

kira-syslogng:
success

kira-syslogng:
success

mehul-m-prajapati:
Hi @furiel ,

I am still getting error for iv.h evn after adding -Wno-error=strict-prototypes
```
In file included from ../lib/logthrdestdrv.h:33:0,
                 from lib/cfg-grammar.y:41,
                 from modules/native/native-grammar.y:39:
./lib/ivykis/src/include/iv.h:57:14: error: function declaration isn’t a prototype [-Werror=strict-prototypes]
 const struct timespec *__iv_now_location_valid();
```

furiel:
Oh I see I overlooked it, we are not failing while building ivykis, but when we include the ivykis headers.

After googling a bit, I found gcc will ignore warnings if header directories are added by `-isystem` instead of `-I`. Actually it seems vendor provided headers should use -isystem anyway, so maybe you found a bug?

https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html#Directory-Options
> You can use -I to override a system header file, substituting your own version, since these directories are searched before the standard system header file directories. However, you should not use this option to add directories that contain vendor-supplied system header files; use -isystem for that. 

After replacing `-I` with `-isystem` in three places at `IVYKIS_CFLAGS` and `INTERNAL_IVYKIS_CFLAGS`, gcc did not complain anymore for ivykis. So this can be one option. Though I am interested in your and other's opinion if we are fine to change this. I am low on xp on this topic, what other effects this can have.

Another alternative I found is to temporarily disable warnings with `#pragma`, such includes could be decorated with it. Though this needs a little experiment, because such options are only available from a specific version of gcc, and it would be unfortunate to fail to compile do to this on older platforms. As I counted, iv.h is included in 31 places, so in this option probably it would be better have a special include file in lib/compat, something like 'syslog-iv.h', which encapsulates this ugly pragma detail, and that header should be used in the code at those 31 places.

All that said, I still ran into trouble with sql source... The libdbi driver has the same problem. Now the problem is we cannot touch that with -isystem, the options come from a .pc file, we do not have control on that. Maybe the pragma version would work there, but: as travis compiles without sql support, I think we can delay solving it now. When you test locally, just pass `--disable-sql` to configure, the same way as travis does.

Unfortunately, that's not all, jsonc also emits warning but for another very strange reason, it signals error saying my system does not have strdup. Let alone, travis does not compile jsonc, it instead installs it from deb, and uses the system version of jsonc: adding `--with-jsonc=system` to configure, so it is not an issue to solve now (though I am utmost interested why it cannot detect my strdup). With jsonc from system, I could finally build locally your patchset without error.

mehul-m-prajapati:
Hi @furiel ,

Is there any other way to commit following change ?
```
-const struct timespec *__iv_now_location_valid();
+const struct timespec *__iv_now_location_valid(void);
```

furiel:
As ivykis is a submodule, I am afraid the only way would be a local patch over the vendor code. I would like to avoid this, because it further increases the complexity to maintain these modules.

You can try to contribute to ivykis upstream, and with a stable ivykis release, we can update the submodule version. Though it would not be future proof, we cannot enforce appliance for submodules from a syslog-ng build script, ivykis can break it later on. Also, there might be other warnings that need to be managed. So we need a general solution without enforcing anything to the ivykis code.

(But this should not hold you back to contribute to ivykis, it would still be a good patch :) )

gaborznagy:
@mehul-m-prajapati! Can you rebase your branch against latest master, please? 
We have just merged a warning related PR.

mehul-m-prajapati:
Hi @gaborznagy ,

I have rebased my branch against latest master. Thanks for the update.

@furiel : 
```
After replacing -I with -isystem in three places at IVYKIS_CFLAGS and INTERNAL_IVYKIS_CFLAGS, gcc did not complain anymore for ivykis.
```
I see these flags in **configure.ac** .

Is it the correct place where I need to replace  ```-I with -isystem.```

furiel:
@mehul-m-prajapati that's right, two in the IVYKIS_CFLAGS line, one in the INTERNAL_IVYKIS_CFLAGS line:
```
--- a/configure.ac
+++ b/configure.ac
[...]
-               IVYKIS_CFLAGS="-I\$(top_srcdir)/lib/ivykis/src/include -I\$(top_builddir)/lib/ivykis/src/include"
-                INTERNAL_IVYKIS_CFLAGS="-I\${includedir}/syslog-ng/ivykis"
+               IVYKIS_CFLAGS="-isystem \$(top_srcdir)/lib/ivykis/src/include -isystem \$(top_builddir)/lib/ivykis/src/include"
+                INTERNAL_IVYKIS_CFLAGS="-isystem \${includedir}/syslog-ng/ivykis"
```

kira-syslogng:
success

kira-syslogng:
success

mehul-m-prajapati:
Hi @furiel ,

All warning related errors are removed.

There are still 2 things which fails in Travis CI,
1. In gcc compiler, 
```
mkdir: cannot create directory ‘syslog-ng-3.14.1/_build’: File exists
make: *** [distcheck] Error 1
```
2. In clang compiler,
```
lib/scanner/kv-scanner/kv-scanner.c:264:24: error: initializer overrides prior
      initialization of this subobject [-Werror,-Winitializer-overrides]
    .match_delimiter = _match_delimiter,
                       ^~~~~~~~~~~~~~~~
lib/scanner/kv-scanner/kv-scanner.c:263:5: note: previous initialization is here
    0,
    ^
1 error generated.
```

Do you have any idea on this?

Thanks,
Mehul

kira-syslogng:
success

furiel:
About the gcc: 
Yes it is unfortunate that our distcheck part in travis is bug-ish.
The long story: we want distcheck run fast, so start with -j3. But if parallel build is used, it is hard to spot the problem because of the interleaving build logs. So when a make -j3 distcheck fails, we restart make distcheck with -j1, hoping that the already compiled parts are skipped, so the first error will be the error we are looking for. This tecnique works fine with simple make command, but not with distcheck, because distcheck locks the _build directory, so the second round cannot even start. Essentially, any problem during distcheck results in this strange error _build exists.

Short story, you need to scroll up a little bit to see the problem:
make[4]: Nothing to be done for `../modules/python/pylib/test_pylib.sh'.
```
../lib/secret-storage/tests/test_secret_storage.c:39:1: error: function declaration isn’t a prototype [-Werror=strict-prototypes]
 secret_storage_testsuite_init()
 ^
../lib/secret-storage/tests/test_secret_storage.c:46:1: error: function declaration isn’t a prototype [-Werror=strict-prototypes]
 secret_storage_testsuite_deinit()
```

About clang, me seems the problem is as if StrReprDecodeOptions options would be initialized two times: first with zero, and then all fields are enumerated and initialized separately. So in the end the `,0` does not make sense and it probably causes the problem. But it is just a guess, I did not test it.

mehul-m-prajapati:
Hi @furiel,

You are right on both the cases.

1. I have made changes in test_secret_storage.c to remove warnings.

2. I have remove double initialization to resolve clang error.

Thanks,
Mehul

kira-syslogng:
success

kira-syslogng:
success

furiel:
Well done! I just have some cosmetic requests.

Please use `Signed-off-by: ...` in your commit message. Commits are cherrypicked, ported, rebased, moved around time to time, so the author of the commit can change. However Signed-off-by is eternal, maintaining your fame and glory forever as contributor to syslog-ng. You can find example for the format in almost every commit messages, or you can generate it with git commit -s/--signoff command line option.

And this leads to the cherry-picked part of your patch, please revert it to a separate patch, maintaining the original signed-off-by line. (As in this case that patch is an ugly hack because we cannot do better, you are better of not linking your name to commit :)) 

Finally, please add a space after -isystem. Strangely enough I tried and it works as it is now, though I'm not sure why. In my previous experiences, I always seen space when others used -system includedir, moreover the manual of gcc does not explicitely enables this: while`-Idir...` is allowed without space, all occurrences I found in the gcc manual uses space. It is safer to follow it.
```

kira-syslogng:
success

kira-syslogng:
success

mehul-m-prajapati:
Hi @furiel ,

Done the changes. Please look at it now ...

Thanks,
Mehul

kira-syslogng:

kira-syslogng:
success

PR #1968, opened 2018-04-05
testing: fixing symbol redefinition error

furiel:
Only testing is affected!

Now both cr_template and msg_parse_lib contain the same symbol: `MsgFormatOptions parse_options`. Also the syslogformat module initializer/deinitializer functions are copy-pasted between the two modules.

This patch resolves the problem: cr_template should use these symbols from msg_parse_lib.


kira-syslogng:
success

PR #1970, opened 2018-04-06
Fix possible LogMessage-related crash (NVTable)

MrAnno:
Subtracting an "out-of-boundary" number from a pointer is undefined.
The result may underflow making the condition false, causing a crash.

(`alloc_size` is the size of a new entry that we want to store in NVTable.)

Apart from this, my change is an equivalent transformation.

This bug has been identified on a Solaris 11 machine, where `NVTable` is often stored on low addresses.


kira-syslogng:
success

bazsi:
Indeed good catch, and I guess it was very difficult to diagnose. Nice job!

PR #1971, opened 2018-04-06
Enable warning: warning manipulation

furiel:
The final helper PR for #1302.

This is the last part I would like to extract from #1302. The idea is the following:
- This patchset adds those warnings to enable extra warnings that do not generate warnings for now. The other part of the list is left in #1302.
- Besides that part, I ported the CFLAGS specification cleanup patch of @bazsi.

The plan is #1302 will only contain those warning options, that we need to deal with code change. When we have a list, we will write issues about those and expect those be fixed/analyzed through the newly created issues and prs. As such, #1302 will be resolved.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -43,9 +43,16 @@ AM_CPPFLAGS		= -I$(top_srcdir)/lib -I$(top_srcdir)/modules -I$(top_builddir)/lib
 # configure script gives us -g and/or -O2, but all warning related settings
 # should be here
 
+# Important warnings
 AM_CFLAGS = \
 	-Wshadow \
 	-fcommon
+
+# Acceptable warnings
+AM_CFLAGS += \
+	-Wno-stack-protector \
+	-Wno-unused-parameter \
+	-Wno-variadic-macros
 
 if ENABLE_EXTRA_WARNINGS
 AM_CFLAGS += \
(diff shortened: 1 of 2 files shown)


kira-syslogng:
success

gaborznagy:
These are just notes in general for warnings. Sorry if it's spam from this PR's point of view.
- list of warnings we would like to have
In the description of #1302 there is a list of warnings that we would like to have.
Is that list complete? I have just spotted a minor warning that is not covered although its cousin has been covered already: `-Wold-style-definition` is not on the list, while `-Wold-style-declaration` is fixed already.

- clang support of enable-extra-warnings
We have turned off `--enable-extra-warnings` in travis for clang since https://github.com/balabit/syslog-ng/commit/248cf78be9e67feaf90304fe78797c077c1b8456.
It would be good to have it again, although it would not give us a lot.
Clang does not support a lot of warnings that gcc does.
Also there are some warnings from gcc that is supported in clang, but just for compatibility purpose and they do nothing...
E.g.
https://clang.llvm.org/docs/DiagnosticsReference.html#wold-style-definition

- conditional usage of some of the warning switches
I see that this patch adds more conditional usage of warnings that I mentioned in clang support above.
https://github.com/balabit/syslog-ng/blob/518dc5da1f11e2d33a0afcb99176995a749563e2/configure.ac#L495-L497
Can we use this to solve the clang problem, or it would increase complexity hugely?

PR #1973, opened 2018-04-09
Switch case default warning fixes

### LINKED ISSUE #1963: warning elimination: switch case default branch

This issue should cover the switch case default branch related warnings
```
     54 [-Wswitch-default]
```

mehul-m-prajapati:
Resolves #1963

### DIFF
--- a/lib/block-ref-grammar.ym
+++ b/lib/block-ref-grammar.ym
@@ -24,6 +24,7 @@
 
 %code requires {
 
+#pragma GCC diagnostic ignored "-Wswitch-default"
 #include "block-ref-parser.h"
 
 #include <string.h>
--- a/lib/cfg-lexer-subst.c
+++ b/lib/cfg-lexer-subst.c
@@ -92,6 +92,9 @@ _track_string_state(CfgLexerSubst *self, CfgLexerStringTrackState last_state, co
       if (*p == '\'')
         return CLS_NOT_STRING;
       return CLS_WITHIN_QSTRING;
+    default:
+      g_assert_not_reached();
+      break;
     }
   g_assert_not_reached();
 }
--- a/lib/cfg-parser.c
+++ b/lib/cfg-parser.c
@@ -421,6 +421,9 @@ cfg_process_flag(CfgFlagHandler *handlers, gpointer base, const gchar *flag_)
               else
                 *field = (*field) & ~handler->param;
               return TRUE;
+            default:
+              g_assert_not_reached();
+              break;
             }
         }
     }
--- a/lib/logproto/logproto-multiline-server.c
+++ b/lib/logproto/logproto-multiline-server.c
@@ -48,6 +48,10 @@ log_proto_multiline_server_new(LogTransport *transport,
                                                           options->garbage);
     case MLM_NONE:
       return log_proto_text_server_new(transport, &options->super);
+
+    default:
+      g_assert_not_reached();
+      break;
     }
   g_assert_not_reached();
 }
--- a/lib/template/macros.c
+++ b/lib/template/macros.c
@@ -570,6 +570,9 @@ log_macro_expand(GString *result, gint id, gboolean escape, const LogTemplateOpt
           length = format_zone_info(buf, sizeof(buf), zone_ofs);
           g_string_append_len(result, buf, length);
           break;
+        default:
+          g_assert_not_reached();
+          break;
         }
       break;
     }
--- a/lib/template/templates.c
+++ b/lib/template/templates.c
@@ -162,6 +162,9 @@ log_template_append_format_with_context(LogTemplate *self, LogMessage **messages
           g_static_mutex_unlock(&self->arg_lock);
           break;
         }
+        default:
+          g_assert_not_reached();
+          break;
         }
     }
 }
--- a/modules/afamqp/afamqp.c
+++ b/modules/afamqp/afamqp.c
@@ -348,6 +348,12 @@ afamqp_is_ok(AMQPDestDriver *self, const gchar *context, amqp_rpc_reply_t ret)
                     evt_tag_int("time_reopen", self->super.time_reopen));
           return FALSE;
         }
+    default:
+      msg_error(context,
+                evt_tag_str("driver", self->super.super.super.id),
+                evt_tag_int("reply_type", ret.reply_type),
+                evt_tag_str("error", "unknown response type"),
+                evt_tag_int("time_reopen", self->super.time_reopen));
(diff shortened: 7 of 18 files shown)


kira-syslogng:

gaborznagy:
@kira-syslogng ok to test

kira-syslogng:
success

kira-syslogng:
success

mehul-m-prajapati:
Hi @furiel .

I have done following changes,
1.  Added missing default statements in source code.
2. Added pragma to ignore warnings from yacc/lex/bison generate source files.

Please review it.

Thanks,
Mehul

lib/cfg-lexer-subst.c
```diff
@@ -92,6 +92,8 @@ _track_string_state(CfgLexerSubst *self, CfgLexerStringTrackState last_state, co
       if (*p == '\'')
         return CLS_NOT_STRING;
       return CLS_WITHIN_QSTRING;
+    default:
```
furiel: In this case we are in a state machine. It is a fatal error to enter to the default state. Please add g_assert_not_reached() into the default branch.

lib/cfg-parser.c
```diff
@@ -421,6 +421,8 @@ cfg_process_flag(CfgFlagHandler *handlers, gpointer base, const gchar *flag_)
               else
                 *field = (*field) & ~handler->param;
               return TRUE;
+            default:
```
furiel: Operation should be either SET or CLEAR. This should be g_assert_not_reached() again.

lib/logproto/logproto-multiline-server.c
```diff
@@ -48,6 +48,9 @@ log_proto_multiline_server_new(LogTransport *transport,
                                                           options->garbage);
     case MLM_NONE:
       return log_proto_text_server_new(transport, &options->super);
+
+    default:
```
furiel: This case should cover all constants for multiline options. This should be a g_assert_not_reached()

lib/template/macros.c
```diff
@@ -570,6 +570,8 @@ log_macro_expand(GString *result, gint id, gboolean escape, const LogTemplateOpt
           length = format_zone_info(buf, sizeof(buf), zone_ofs);
           g_string_append_len(result, buf, length);
           break;
+        default:
```
furiel: This should cover all macro types. The default branch should indicate a bug. Please add g_assert_not_reached() here.

lib/template/templates.c
```diff
@@ -162,6 +162,8 @@ log_template_append_format_with_context(LogTemplate *self, LogMessage **messages
           g_static_mutex_unlock(&self->arg_lock);
           break;
         }
+        default:
```
furiel: A template should be either a function, a macro or a value. This switch case should cover all these. Here we should use g_assert_not_reached().

modules/afsocket/socket-options-inet.c
```diff
@@ -159,6 +159,8 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
       break;
     }
 #endif
+    default:
```
furiel: IP address can be either ipv4 or ipv6. We should use g_assert_not_reached() here.

kira-syslogng:
success

mehul-m-prajapati:
Hi @furiel ,

Thanks for your valuable review comments. 
I got to know about insights of the code which I modified.

Regards,
Mehul

furiel:
Cool thanks. I see I added my comments to wrong line thus github did not detect the change. I just remove those because they are not relevant anymore.

lib/block-ref-grammar.ym
```diff
@@ -24,6 +24,7 @@
 
 %code requires {
 
+#pragma GCC diagnostic ignored "-Wswitch-default"
```
gaborznagy: For now `--enable-extra-warnings` is turned off for clang, but I would like to avoid diversion between clang and gcc.
As far as I found out clang support gcc pragmas, but can you verify it too please?
Here is the documentation I found about clang and pragmas
https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-command-line-flags

mehul-m-prajapati: Yes @gaborznagy. Clang supports gcc pragmas.

Thus, we should add one more line as ```#pragma clang diagnostic ignored "-Wswitch-default"```.

May I do the this change and commit ?

gaborznagy: As clang support gcc pragmas we should not add an extra line with a clang pragma.
The second option you suggested should be used only if you would like to ignore warnings for specific lines instead of the whole file.

So what I suggest is to have gcc pragmas only.

mehul-m-prajapati: Ok. I got the point.

mehul-m-prajapati:
Hi @furiel ,

I have a doubt whether we should use,
```
#pragma GCC diagnostic ignored "-Wswitch-default"
```

**OR**

```
#pragma GCC diagnostic ignored \"-Wswitch-default\"
```

Which is correct according to you ?

Thanks,
Mehul

furiel:
I believe we do not need to escape. The lexer should handle those blocks as verbatim. Did you run into some kind of trouble without escaping?

mehul-m-prajapati:
Nope. I have not faced any trouble without escaping.
It works fine without escaping on machine ...

lib/template/macros.c
```diff
@@ -570,6 +570,9 @@ log_macro_expand(GString *result, gint id, gboolean escape, const LogTemplateOpt
...
```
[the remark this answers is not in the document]
mehul-m-prajapati: So, this change is correct ...

modules/afamqp/afamqp.c
```diff
@@ -348,6 +348,11 @@ afamqp_is_ok(AMQPDestDriver *self, const gchar *context, amqp_rpc_reply_t ret)
                     evt_tag_int("time_reopen", self->super.time_reopen));
           return FALSE;
         }
+    default:
```
gaborznagy: Can you add the numeric value of the unknown reply type, please?
And replace the "unknown server error" message to something unique, please?
E.g. "unknown response type" or "unknown response code"

mehul-m-prajapati: Done the mentioned changes..,

gaborznagy: Thanks! Looks good to me.

kira-syslogng:
success

PR #1975, opened 2018-04-09
log_writer_write_message: same modification as PR 1890

szemere:
The else branch has a same debug message as the modified one in #1890

### DIFF
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -1100,7 +1100,7 @@ log_writer_write_message(LogWriter *self, LogMessage *msg, LogPathOptions *path_
   if (!(msg->flags & LF_INTERNAL))
     {
       msg_debug("Outgoing message",
-                evt_tag_printf("message", "%.*s", (gint) self->line_buffer->len, self->line_buffer->str));
+                evt_tag_printf("message", "%s", self->line_buffer->str));
     }
 
   if (self->line_buffer->len)
@@ -1150,7 +1150,7 @@ log_writer_write_message(LogWriter *self, LogMessage *msg, LogPathOptions *path_
   else
     {
       msg_debug("Can't send the message rewind backlog",
-                evt_tag_str("message",self->line_buffer->str));
+                evt_tag_printf("message", "%s", self->line_buffer->str));
 
       log_queue_rewind_backlog(self->queue, 1);
 


kira-syslogng:
success

lib/logwriter.c
```diff
@@ -1150,7 +1150,7 @@ log_writer_write_message(LogWriter *self, LogMessage *msg, LogPathOptions *path_
   else
     {
       msg_debug("Can't send the message rewind backlog",
-                evt_tag_str("message",self->line_buffer->str));
+                evt_tag_printf("message", "%.*s", (gint) self->line_buffer->len, self->line_buffer->str));
```
furiel: This change is not needed IMO. Internally we will write into a 1024 buffer with vsnprinf anyway. So I do not see danger for overflow here.

szemere: Good catch. I also modified the original printf in the "true branch".

kira-syslogng:
success

lib/logwriter.c
```diff
@@ -1100,7 +1100,7 @@ log_writer_write_message(LogWriter *self, LogMessage *msg, LogPathOptions *path_
   if (!(msg->flags & LF_INTERNAL))
     {
       msg_debug("Outgoing message",
-                evt_tag_printf("message", "%.*s", (gint) self->line_buffer->len, self->line_buffer->str));
+                evt_tag_printf("message", "%s", self->line_buffer->str));
     }
```
bazsi: if the message is NUL terminated (is it?), then why can't we use evt_tag_str() here?

BTW: I think it is not guaranteed to be NUL terminated, although possibly this is ok, as we are not perfectly NUL safe.

But with $(binary 0x00) you can embed a NUL character in the output stream.

bazsi: I think it is perfectly fine to truncate in an internal message, especially if it is about our outgoing stream.

So, it is fine as it is, I just wanted to mention that evt_tag_str() should be ok as well. That doesn't have the truncation as a feature though.

szemere: We use evt_tag_printf with a `%.*s` format string in several places, just to truncate the output.
I think the format string is quite a big overhead to just use it as a "truncate" feature.

I propose a new function:
evt_tag_strn(const char* s, int limit, int length = -1)

If ok, I will create a new PR, and replace the evt_tag_printf in the code. (Including this one.)

PR #1976, opened 2018-04-09
Fix possible crash in afsocket

furiel:
This patchset contains a fix initialization related crash in afsocket.

### DIFF
--- a/modules/afsocket/afsocket-source.c
+++ b/modules/afsocket/afsocket-source.c
@@ -95,6 +95,12 @@ afsocket_sc_init(LogPipe *s)
 
       proto = log_proto_server_factory_construct(self->owner->proto_factory, transport,
                                                  &self->owner->reader_options.proto_options.super);
+      if (!proto)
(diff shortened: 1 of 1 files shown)


kira-syslogng:
success

bazsi:
Why do we need the controller passed to the LogProto layer? I don't see it actually being used.

MrAnno:
One of our internal protos has multiple idle/command/etc. timers that can, for example, close the connection. We didn't want to contaminate `LogReader` with such timeout logic (which would produce more "dead code" on the OSE side), so we passed the controller and notified the controller pipe directly to close the connection (NC_CLOSE).

furiel:
@bazsi I dropped down the two controversal patches : virtual deinit and passing control pipe to constructor. It now only contains only a single patch: @MrAnno 's bugfix. I believe when the tests pass, this can be merged.

bazsi:
agreed. looks good to me.

Issue #1857, opened 2018-01-22
"dbld/rules rpm-centos7" fails on current git head


czanik:
# syslog-ng

## Version of syslog-ng

git head as of today

## Platform

CentOS 7

# Issue

## Failure

Compilation fails when using:
```
dbld/rules rpm-centos7
```
Error message is:
```
 /usr/bin/mkdir -p '/root/rpmbuild/BUILDROOT/syslog-ng-3.13.2-7+20180122T123321.el7.centos.x86_64/usr/share/syslog-ng/xsd'
 /usr/bin/install -c -m 644 doc/xsd/patterndb-1.xsd doc/xsd/patterndb-2.xsd doc/xsd/patterndb-3.xsd doc/xsd/patterndb-4.xsd doc/xsd/patterndb-5.xsd '/root/rpmbuild/BUILDROOT/syslog-ng-3.13.2-7+20180122T123321.el7.centos.x86_64/usr/share/syslog-ng/xsd'
+ /usr/bin/install -d -m 755 /root/rpmbuild/BUILDROOT/syslog-ng-3.13.2-7+20180122T123321.el7.centos.x86_64/etc/syslog-ng/conf.d
+ /usr/bin/install -p -m 644 packaging/rhel/syslog-ng.conf /root/rpmbuild/BUILDROOT/syslog-ng-3.13.2-7+20180122T123321.el7.centos.x86_64/etc/syslog-ng/syslog-ng.conf
/usr/bin/install: cannot stat 'packaging/rhel/syslog-ng.conf': No such file or directory
error: Bad exit status from /var/tmp/rpm-tmp.h3PawN (%install)


RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.h3PawN (%install)
make: *** [rpm-centos7] Error 1
```

## Steps to reproduce

Checkout syslog-ng from git and run "dbld/rules rpm-centos7"

asaf400:
Any updates on this issue \ how to workaround it?

gaborznagy:
Hi @asaf400!

We are currently working on this, however we don't know if we can release this in the upcoming version (3.15).

PR #1876, opened 2018-02-06
Reorganize dbld

szemere:
New versions of Docker (command line docker build, and DockerHub automated builds) accepts Dockerfile names as a parameter. Thus we no longer bounded to the concept of individual directory/dockerfile. With the new structure we can eliminate duplicated files. (Common distributions has very similar helper resources.)

Also changed the structure of the Dockerfiles itself.
- Separated the installation of distribution provided packages from the private repositories (obs/epel).
- Experimental introduction of pip packages (usually newer versions available, compared to the distributions stable branch)
- Common steps moved into [functions.sh](dbld/images/helpers/functions.sh)

Written a general README.md rather than the the distribution related ones. (Using Ubuntu Xenial as a sample where it is required.) - Suggestions/improvements are warmly welcome in the comments.

With the new pylint there were some - currently accepted - warning messages. (`make check` runs it)

Ubuntu Zesty (17.04.) reached EOL (2018.01.13.)
- Removed it.
- Introduced Artful Aardvark (17.10)
- Changed the default **deb, shell, tarball** targets to Xenial

Questions:
- Currently pip packages can not be verified by the `dpkg-checkbuilddeps`. (Makes sens, since they are not dpkg packages.) Should we create some "custom workaround" for addressing this case? i.e. empty dpkg packages as placeholders, or remove the already available binaries from the debian control file?
- How do you decide where to put a package? (pip, dist, obs/epel)

@czanik : I will appreciate some notes on the CentOS related part.
@mitzkia : Can you help me with the DockerHub, Travis integration?

resolves: #1822

PR #1980, opened 2018-04-11
rpm generation process update

szemere:
RPM generation process update from @Czanik.
This is the first step in a long process, the final goal is to harmonize Peter's work and `/dbld`
  - remove syslog-ng-add-contextual-data.pc from file list
  - update to 3.14.1
  - re-enable amqp & mongodb on RHEL only
  - keep tcp wrappers support only on RHEL

Fixes: #1857

### DIFF
--- a/dbld/images/required-epel/all.txt
+++ b/dbld/images/required-epel/all.txt
@@ -2,3 +2,5 @@
 libdbi
 libdbi-drivers
 riemann-c-client-devel
+syslog-ng-java-deps
+gradle
--- a/dbld/rpm
+++ b/dbld/rpm
@@ -7,6 +7,10 @@ VERSION=`cat VERSION`
 RPMBUILD=${HOME}/rpmbuild
 RPMBUILD_SOURCES=$RPMBUILD/SOURCES
 
+# CzP's spec file expect it in the pwd
+# where the build was initiated
+cp packaging/rhel/syslog-ng.* /build
+
 cd /build
 mkdir -p $RPMBUILD_SOURCES
 cp  syslog-ng-${VERSION}.tar.gz $RPMBUILD_SOURCES
--- a/dbld/images/helpers/functions.sh
+++ b/dbld/images/helpers/functions.sh
@@ -14,7 +14,7 @@ function add_epel_repo {
     if [ "$1" == "centos6" ]; then
         wget https://copr.fedorainfracloud.org/coprs/czanik/syslog-ng39epel6/repo/epel-6/czanik-syslog-ng39epel6-epel-6.repo
     elif [ "$1" == "centos7" ]; then
-        wget https://copr.fedorainfracloud.org/coprs/czanik/syslog-ng311/repo/epel-7/czanik-syslog-ng311-epel-7.repo
+        wget https://copr.fedorainfracloud.org/coprs/czanik/syslog-ng-githead/repo/epel-7/czanik-syslog-ng-githead-epel-7.repo
     else
         return 1
     fi
--- a/packaging/rhel/syslog-ng.spec
+++ b/packaging/rhel/syslog-ng.spec
@@ -1,26 +1,23 @@
-# %global evtlog_ver 0.2.13
 %global ivykis_ver 0.36.1
 
 Name: syslog-ng
-Version: 3.12.1
-Release: 7%{?dist}
+Version: 3.15.1
+Release: 8%{?dist}
 Summary: Next-generation syslog server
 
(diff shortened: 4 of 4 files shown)


kira-syslogng:
success

dbld/images/required-epel/all.txt
```diff
@@ -2,3 +2,5 @@
 libdbi
 libdbi-drivers
 riemann-c-client-devel
+syslog-ng-java-deps
```
bazsi: Where does this package come from?

dbld/rpm
```diff
@@ -8,6 +8,10 @@ USER=`whoami`
 RPMBUILD=/home/${USER}/rpmbuild
 RPMBUILD_SOURCES=$RPMBUILD/SOURCES
 
+# CzP's spec file expect it in the pwd
+# where the build was initiated
+cp packaging/rhel/syslog-ng.* /build
+
```
bazsi: My understanding is that they need to be in the rpmbuild source directory. Isn't that the case?

dbld/images/required-epel/all.txt
czanik: From my Copr repo: https://copr.fedorainfracloud.org/coprs/czanik/syslog-ng-githead/package/syslog-ng-java-deps/

dbld/rpm
czanik: It is looking for the files where rpmbuild is initiated. In our case it is '''/build'''

gaborznagy: I was experimenting with it as well.
rpmbuild is using the dist tarball, isn't it?
Why don't we include it our Makefile instead of this copy? I have tried it out and it works, it would be similar to what our debian packaging does.

dbld/images/required-epel/all.txt
gaborznagy: I think it's not the githead branch in your repo, but version "3.11":
https://github.com/balabit/syslog-ng/blob/4ab34df81a7f97740e60aa5b0748c54b6518338b/dbld/images/helpers/functions.sh#L14-L17
This is used during docker image creation.

gaborznagy:
This is a huge step to resolve #1857 :+1:

MrAnno:
Hi,

We've just merged a CI-related pull request into our master branch.
Since the current Travis CI checks in your PR are outdated, I'd like to ask you to rebase against master.

kira-syslogng:
success

kira-syslogng:
success

czanik:
Did a quick test in a completely clean environment. Running **dbld/rules rpm-centos7** fails with dependency problems only if **dbld/rules image-centos7** is not run first. Thanks to @szemere it now also works if it was invoked as root.

As discussed IRL with @gaborznagy , my githead repo should be used instead of 311 to make sure that the right version of dependencies are used.

gaborznagy:
@czanik : by completely clean environment you mean you did not have any of the docker images prior testing, right?
We've rebuilt the centos7 docker images on dockerhub a couple of days ago, they should

About the changes:
- I have tested rpm generating in dbld and it worked when I built the image prior to testing.
- I have not tested it dbld with your githead repo so far.
- I still need to find resolution for the problems I found when I included the packaging/rhel/ dir's content in the dist tarball:
1. There is another spec in the repo, I have removed it from the dist tarball for testing, so there will be only the packaging/rhel spec file in the tarball.
2. I tried to remove many unnecessary copying of the packaging dir (this is common in dbld).
2. Your spec file expects files under the build/ dir which now has to be copied (this is what's in this PR), I try to change this to use files from the tarball instead.

szemere:
> Running dbld/rules rpm-centos7 fails with dependency problems only if dbld/rules image-centos7 is not run first.

You are absolutely right. A little background information:
Because of a configuration mistake in Docker Hub, the **latest** TAG pointed to an outdated version of the image, from the time before #1876 . (Obviously with outdated dependencies.)
By running **dbld/rules image-centos7** the image was (re)built locally and overrode whatever was downloaded from Docker Hub.
I corrected the Docker Hub config on April 17th, so the **latest** tag will represent the last commit from the master branch. By running **docker pull balabit/syslog-ng-centos7** manually you can override your local version without the long build process.

gaborznagy:
This PR is awaited. 
Let's test "githead" repo instead of "3.11" and if it works, merge the PR in.

The resolution of the many copying of the packaging dir in dbld will be resolved later.

szemere:
Updated epel repo url to git head.
Version bumped the spec file.

PR #1981, opened 2018-04-11
dbld: corrected path's in case of root user

szemere:
root users home directory is not under /home, replaced `home/$(whoami)` in rpmbuild.


bazsi:
This looks good to me. Consider it approved.

kira-syslogng:
success

PR #1983, opened 2018-04-12
compat: strnlen impl. if not present

### LINKED ISSUE #1982: tests/loggen/loggen.c uses strnlen which fails on solaris10
# syslog-ng

Configure should check for the existence of strnlen() (the mongo-c-driver/libbson library does just that)

## Version of syslog-ng

syslog-ng-3.14.1

## Platform

Solaris 10, i386

Kokan:
Implements `strnlen` if it is missing from the `string.h` header.

@pryankster please test this

Fixes #1982


kira-syslogng:
success

lib/compat/strnlen.c
```diff
@@ -0,0 +1,36 @@
...
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+#include "compat/string.h"
+
+#if !SYSLOG_NG_HAVE_STRNLEN
+
+#include <stdint.h>
```
furiel: according to man 3 memchr, memchr is included from string.h, which is included from compat/string.h. Do we need this include?

Kokan: This is not included because of the `memchr`, that is guaranteed to be included for the reason you stated above.
And I totally meant `stddef.h` not `stdint.h` :), that is needed for `size_t`. I guess the `stddef.h` is included at some place either way.

kira-syslogng:
success

PR #1984, opened 2018-04-12
dbld/images/devshell: generate utf8 locale

bazsi:
This patch adds locale and ensures en_US.utf8 is generated during
image build. This is required to use an utf8 locale at runtime, which
is the most common setup these days.

### DIFF
--- a/dbld/images/devshell.dockerfile
+++ b/dbld/images/devshell.dockerfile
@@ -13,3 +13,5 @@ RUN cat /devshell-pip/* | grep -v '^$\|^#' | xargs pip install
 
 COPY devshell-apt/all.txt devshell-apt/${DISTRO}*.txt /devshell-apt/
 RUN cat /devshell-apt/* | grep -v '^$\|^#' | xargs apt-get install --no-install-recommends -y
+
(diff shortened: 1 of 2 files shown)


bazsi:
NOTE: I've found this while trying to use Python unicode object in our python module, which is default for Python3.

dbld/images/devshell.dockerfile
```diff
@@ -13,3 +13,5 @@ RUN cat /devshell-pip/* | grep -v '^$\|^#' | xargs pip install
 
 COPY devshell-apt/all.txt devshell-apt/${DISTRO}*.txt /devshell-apt/
 RUN cat /devshell-apt/* | grep -v '^$\|^#' | xargs apt-get install --no-install-recommends -y
+
+RUN echo en_US.utf8 > /etc/locale.gen && locale-gen
```
Kokan: I think this should be `en_US.UTF-8 UTF-8` instead.

kira-syslogng:
success

bazsi:
Thanks. I just tried to capture what I did manually in the container. This time I also built it :)

Thanks for the feedback and sorry for the noise.

kira-syslogng:
success

PR #1985, opened 2018-04-12
[WIP] Python venv support

bazsi:
This series of patches add support for a Makefile managed Python virtualenv so that one does not need to use the system deployed packages. This can be used when working with the syslog-ng source tree outside of the developer's docker images.


kira-syslogng:
success

kira-syslogng:
success

bazsi:
abandoning

PR #1989, opened 2018-04-13
test_run_id: rewrite test in criterion

nobles:
In the previous implementation, all tests used the same persist file. This caused problems with criterion, because criterion - by default - executes tests in parallel. As a result the tests overwrote each others's persist file. To fix this, I needed to dedicate persist files for each tests. That's why the signature of create_persist_state() needed to be modified to accept the name of the persist file.


kira-syslogng:

MrAnno:
@kira-syslogng ok to test

kira-syslogng:

kira-syslogng:
success

lib/tests/test_runid.c
```diff
@@ -28,24 +28,28 @@
...
+#include <criterion/criterion.h>
 
+PersistState *create_persist_state(gchar *persist_file_name);
+
+TestSuite(test_run_id, .init = app_startup, .fini = app_shutdown);
 #define RUN_ID_FIRST 1
 
 PersistState *
-create_persist_state(void)
+create_persist_state(gchar *persist_file_name)
```
Kokan: This function is  not really needed, just calls directly the other one.

MrAnno:
@nobles Nice. Thank you very much for the PR!

PR #1991, opened 2018-04-16
Fix connection close in network sources

MrAnno:
Previously, the connection was closed after calling the destructor of LogReader.

The destruction of LogReader might be delayed, for example, by AckTracker, so an explicit close method is required.

### DIFF
--- a/lib/logproto/logproto-server.c
+++ b/lib/logproto/logproto-server.c
@@ -107,9 +107,19 @@ log_proto_server_validate_options_method(LogProtoServer *s)
 }
 
 void
+log_proto_server_close_transport(LogProtoServer *s)
+{
+  if (s->transport)
+    {
+      log_transport_free(s->transport);
+      s->transport = NULL;
+    }
+}
+
+void
 log_proto_server_free_method(LogProtoServer *s)
 {
-  log_transport_free(s->transport);
+  log_proto_server_close_transport(s);
 }
 
 void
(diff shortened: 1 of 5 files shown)


kira-syslogng:
failure

MrAnno:

kira-syslogng:
success

MrAnno:

kira-syslogng:
success

lib/logproto/logproto-server.c
```diff
@@ -106,10 +106,21 @@ log_proto_server_validate_options_method(LogProtoServer *s)
...
+      s->transport = NULL;
+    }
+}
+
 void
 log_proto_server_free_method(LogProtoServer *s)
 {
-  log_transport_free(s->transport);
+  if (s->transport)
+    log_transport_free(s->transport);
```
Kokan: I think free should set `transport` to `NULL`, just in case.

modules/afsocket/afsocket-source.c
```diff
@@ -212,6 +212,8 @@ afsocket_sd_kill_connection(AFSocketSourceConnection *connection)
 {
   log_pipe_deinit(&connection->super);
 
+  log_reader_close_connection(connection->reader);
```
Kokan: I think this could be applied to other source plugins.

MrAnno: Right, I'll check them. The name "close_connection" is also wrong here. LogReader has nothing to do with connections. "close_transport" might be a bit better name.

lib/logproto/logproto-server.c
MrAnno: The entire `LogProtoServer` is destroyed after calling this function, but I can add the NULL assignment to keep the symmetry with `close_connection` and just to be sure.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

PR #1992, opened 2018-04-16
LogReader async wakeup

MrAnno:
This PR adds a thread-safe suspend and wakeup functionality to LogReader.

LogProtoServer is now able to instruct LogReader to suspend until an asynchronous event occurs.

----------------

Questions:
- I haven't changed the signature of LogProtoClient's prepare(). Since it does not have "suspend" functionality, I think it might be OK to leave it as is. What do you think?

### DIFF
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -325,7 +325,7 @@ log_reader_update_watches(LogReader *self)
       return;
     }
 
-  gboolean line_is_ready_in_buffer = log_proto_server_prepare(self->proto, &cond, &idle_timeout);
+  LogProtoPrepareAction prepare_action = log_proto_server_prepare(self->proto, &cond, &idle_timeout);
 
   if (idle_timeout > 0)
     {
@@ -337,19 +337,27 @@ log_reader_update_watches(LogReader *self)
       iv_timer_register(&self->idle_timer);
     }
 
-  if (self->immediate_check || line_is_ready_in_buffer)
+  if (self->immediate_check)
     {
       log_reader_force_check_in_next_poll(self);
       return;
     }
 
-  if (!cond)
+  switch (prepare_action)
     {
+    case LPPA_POLL_IO:
+      poll_events_update_watches(self->poll_events, cond);
+      break;
+    case LPPA_FORCE_SCHEDULE_FETCH:
+      log_reader_force_check_in_next_poll(self);
+      break;
+    case LPPA_SUSPEND:
       log_reader_suspend_until_awoken(self);
-      return;
+      break;
+    default:
+      g_assert_not_reached();
+      break;
     }
-
-  poll_events_update_watches(self->poll_events, cond);
 }
 
 static void
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -72,6 +72,8 @@ static void log_reader_idle_timeout(void *cookie);
 
 static void log_reader_update_watches(LogReader *self);
 
+static void log_reader_wakeup(LogSource *s);
+
 static void
 log_reader_apply_proto_and_poll_events(LogReader *self, LogProtoServer *proto, PollEvents *poll_events)
 {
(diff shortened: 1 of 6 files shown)


kira-syslogng:
success

kira-syslogng:
success

lib/logreader.c
```diff
@@ -275,19 +291,26 @@ log_reader_update_watches(LogReader *self)
...
 
-  line_is_ready_in_buffer = log_proto_server_prepare(self->proto, &cond);
+  gboolean line_is_ready_in_buffer = log_proto_server_prepare(self->proto, &cond);
   if (self->immediate_check || line_is_ready_in_buffer)
     {
       log_reader_force_check_in_next_poll(self);
       return;
     }
+
+  if (!cond)
```
lbudai: Would it be possible to return a value from an enum?
It would be much more readable than using the default 0 value .

kira-syslogng:
success

bazsi:
It was an enum initially, but then it was redundant, but @mranno has the
details.

MrAnno:
The biggest disadvantage of this "cond = 0" approach can be seen on the LogProtoServer side: 
You can't explicitly tell in code that you want to suspend the LogReader (you can't set `cond` to 0, because cond can only be changed if the transport leaves it empty), so I can only do something like this:

```C
log_proto_my_custom_server_prepare(LogProtoMyCustomServer* self, GIOCondition* cond)
{
    ...
    if (self->state == STATE_WAIT_FOR_EXTERNAL_EVENT)
    {
        /* leave cond as is --> suspend */
        return FALSE;
    }
    ...
}
```

Or another approach :
```C
log_proto_my_custom_server_prepare(LogProtoMyCustomServer* self, GIOCondition* cond)
{
    ...
    if (self->state == STATE_WAIT_FOR_EXTERNAL_EVENT)
    {
        /* this is actually a NOOP, so it also requires an explanation comment */
        if (*cond == 0)
            *cond = 0;
        return FALSE;
    }
    ...
}
```

So I prefer the enum version, but there was definitely redundancy in my implementation. I'll give it another try making this clean.

kira-syslogng:
failure

kira-syslogng:
success

kira-syslogng:
success

lib/filter/tests/test_filters_in_list.c
```diff
@@ -23,8 +23,8 @@
  */
 
 #include <stdlib.h>
-#include <glib.h>
 
+#include "compat/glib.h"
 #include "cfg.h"
```
bazsi: syslog-ng.h already includes glib.h, so we shouldn't include glib directly, but we should include syslog-ng.h as the first thing except if there's already a header that does that.

MrAnno: Oh okay, I'll replace them.
Thanks, @bazsi.

MrAnno: I just leave this here as a note:

In my opinion, the way we use `GLIB_DISABLE_DEPRECATION_WARNINGS` is not sustainable, I find it risky.
We should make it at least independent from the include order of `compat/glib.h`.

Using `-DGLIB_DISABLE_DEPRECATION_WARNINGS` as a CPPFLAG would be an alternative.

pzoleex:

kira-syslogng:
success

kira-syslogng:
success

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

MrAnno:

Issue #1953, opened 2018-03-31
With valgrind the first item (source or destination) will be duplicated


juhaszviktor:
## Version of syslog-ng
3.14
master with configuration opts: 
$ /home/jviktor/work/syslog-ng/configure --prefix=/home/jviktor/work/syslog-ng/install CFLAGS=-g CXXFLAGS=-g --disable-java --disable-python --disable-mongodb --disable-rieman --no-create --no-recursion
```
jviktor@jviktor-X550LB:~/work/syslog-ng/install$ sbin/syslog-ng -V
syslog-ng 3 (3.14.1)
Config version: 3.14
Installer-Version: 3.14.1
Revision: 
Compile-Date: Mar 31 2018 13:10:42
Module-Directory: /home/jviktor/work/syslog-ng/install/lib/syslog-ng
Module-Path: /home/jviktor/work/syslog-ng/install/lib/syslog-ng
Available-Modules: afprog,dbparser,afsocket,affile,system-source,kvformat,cryptofuncs,confgen,tfgetent,basicfuncs,afamqp,graphite,add-contextual-data,pseudofile,http,map-value-pairs,stardate,afstomp,tags-parser,xml,disk-buffer,syslogformat,snmptrapd-parser,sdjournal,afuser,linux-kmsg-format,cef,appmodel,csvparser,date,json-plugin
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: off
Enable-TCP-Wrapper: off
Enable-Linux-Caps: off
Enable-Systemd: off
```

## Platform
Ubuntu 16.04
uname -a
Linux jviktor-X550LB 4.4.0-116-generic #140-Ubuntu SMP Mon Feb 12 21:23:04 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

# Issue

## Failure
With valgrind the first expression node (source or destination) will be duplicated, and syslog-ng doesn't start.
You can see, that the preprocessed configuration file has an additional destination.
If I change the order of destination and source, then the source will be duplicated
```
## Steps to reproduce

1. run syslog-ng with valgrind: valgrind --log-file=valgrind.log sbin/syslog-ng -Fe --preprocess-into=bad.conf
2. output will be:
```
jviktor@jviktor-X550LB:~/work/syslog-ng/install$ valgrind --log-file=valgrind.log sbin/syslog-ng -Fe --preprocess-into=bad.conf
[...]

Issue #1866, opened 2018-01-30
fmemopen() is not available on BSD/MacOS


opoplawski:
# syslog-ng

## Version of syslog-ng
git master

## Platform

MacOS
Darwin 15.6.0

# Issue

## Failure

```
[ 48%] Building C object modules/add-contextual-data/tests/CMakeFiles/test_context_info_db.dir/test_context_info_db.c.o                                                                               
/Users/orion-admin/syslog-ng/modules/add-contextual-data/tests/test_context_info_db.c:204:14: warning: 
      implicit declaration of function 'fmemopen' is invalid in C99
      [-Wimplicit-function-declaration]
  FILE *fp = fmemopen(csv_content, strlen(csv_content) + 1, "r");
             ^
/Users/orion-admin/syslog-ng/modules/add-contextual-data/tests/test_context_info_db.c:204:9: warning: 
      incompatible integer to pointer conversion initializing 'FILE *' (aka 'struct __sFILE *')
      with an expression of type 'int' [-Wint-conversion]
  FILE *fp = fmemopen(csv_content, strlen(csv_content) + 1, "r");
        ^    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/orion-admin/syslog-ng/modules/add-contextual-data/tests/test_context_info_db.c:244:9: warning:
      incompatible integer to pointer conversion initializing 'FILE *' (aka 'struct __sFILE *')
      with an expression of type 'int' [-Wint-conversion]
  FILE *fp = fmemopen(csv_content, sizeof(csv_content), "r");
        ^    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/orion-admin/syslog-ng/modules/add-contextual-data/tests/test_context_info_db.c:322:9: warning:
      incompatible integer to pointer conversion initializing 'FILE *' (aka 'struct __sFILE *')
      with an expression of type 'int' [-Wint-conversion]
  FILE *fp = fmemopen(csv_content, strlen(csv_content) + 1, "r");
        ^    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/orion-admin/syslog-ng/modules/add-contextual-data/tests/test_context_info_db.c:344:9: warning:
[...]

PR #1940, opened 2018-03-23
 PytestFramework: Add initial test framework

mitzkia:
**Current functionality**

* Because it was too hard to review the previous version of the framework, we decied to minimalize it as much as possible.
* The idea was to be able to write a "smallest" testcase (in used features).

## How to use the framework

1, Dependencies:
 * The test-framework was built on top of Python2 and Python3 support
 * The root directory of test-framework contains a requirements.txt file which contains framework dependencies.

2, Usage with make:
 * Here you cannot use custom pytest commands (yet).
 * This mode has only 2 hardcoded commands:
   * `$ make pytest-check` This command runs all functional tests under "functional_test_examples/" dir without verbose console log
   * `$ make pytest-self-check` This command runs all self tests for test-framework

3, Usage without make:
 * Here you can use your favorite pytest commands
 * Here you must define your syslog-ng install path after: "--installdir" option
 * Example1: A standard start for all functional tests with verbose output on console
   * `$ cd tests/pytest_framework`
   * `$ python3 -m pytest --verbose --showlocals -s functional_test_examples/ --installdir=[syslog-ng install path]`
 * Example2: Start only special testcase (or testcases) which are matching on "pattern from testcase"
   * `$ cd tests/pytest_framework`
   * `$ python3 -m pytest --verbose --showlocals -s functional_test_examples/ -k [pattern from testcase] --installdir=[syslog-ng install path]`

4, Reporting:
 * The framework uses capabilities of the builtin `logging` module.
 * Available log levels: INFO (default), DEBUG, ERROR
 * Console log:
   * Pytest display verbose log from testcases when it was started with: "--verbose --showlocals -s"
   * Without this start arguments pytest will not log anything on console, only if there is an error in testcase
 * File log:
   * Self unit-tests are not, but functional-tests are using a separate report directory which contains followings:
     * syslog-ng related files: config, pid, persist, ctl, console stdout/stderr 
     * source-destination related files: input/output files
     * testcase related files: full testcase log, also a started testcase

## The minimal feature set is the following: 
- user can generate config with a file source and file destination
    - mandatory options are not required, framework can be generate it automatically
    - multiple file sources and file destinations can be used (1 statement uses 1 file driver)
- user can start/reload/stop syslog-ng
    - wrong configuration can be detected
    - separate syslog-ng instances can be started with different configurations
- user can generate BSD message(s)
    - message fields can be filled by user or use defaults
- user can write message(s) into the resource of the file source
- user can read message(s) from the resource of the file destination
[...]

PR #1995, opened 2018-04-16
plugin discovery: partial revert to bring back valgrind

furiel:
This patch partially reverts 03771dbba823483e0b285c1c4c8a60c5d16a6f57,
while maintaining as much from the original patch as it
can (structure and codechange). The problem is syslog-ng cannot be
started with valgrind, syslog-ng instantly exits with a strange lexer error.
Running valgrind with --run-libc-freeres=no would start syslog-ng, but
causes a lot of false positive results (even a simple printf appears
in a log). The problem seems to be connected with the fork() and
exit(), but the root cause is yet to be known.

Fixes: https://github.com/balabit/syslog-ng/issues/1953
CC: @juhaszviktor 

Unwanted side effect:
The revert introduces an fmemopen, that is not available in MacOS. This ticket needs to be reopened.
https://github.com/balabit/syslog-ng/issues/1866. After this is merged, the fmemopen needs to be properly ported into compat.


kira-syslogng:
success

lib/plugin.c
```diff
@@ -479,19 +474,19 @@ _enumerate_and_dump_plugin_info_in_modules(const gchar *module_path, int output_
...
+_parse_and_load_plugin_info_in_modules(GString *module_descriptors)
 {
   FILE *discovery;
   GList *candidate_plugins = NULL;
   gchar module_name[4096];
   gint plugin_type;
   gchar plugin_name[256];
 
-  discovery = fdopen(input_fd, "r");
+  discovery = fmemopen(module_descriptors->str, module_descriptors->len, "r");
```
MrAnno: `fmemopen` is not really portable. We should add a compat functions for this.

furiel:
Added WIP flag because I need to compat fmemopen as @MrAnno suggested, and add a testcase that helps avoiding this kind of regression again. For example #1940 of @mitzkia could help adding a test with valgrind.

furiel:
As discussed in an offline review, we will avoid fmemopen and increase the complexity of the revert instead.

kira-syslogng:
success

kira-syslogng:
success

furiel:
I pushed the complete revert of 03771db as discussed. Removed the wip flag. Ready to review.

furiel:
Force push to resolve conflict.

kira-syslogng:
failure

kira-syslogng:
success

PR #1987, opened 2018-04-12
Loggen refactor

norberttak:
Loggen has been splitted into smaller chunks and prepared to
support plugins which can implement any further protocol
See technical details in loggen.md file

Signed-off-by: norberttakacs <norbert.takacs@balabit.com>

Issue #1965, opened 2018-04-03
warning elimination: the rest


furiel:
As the result of #1302 we want to eliminate some special warnings. For this reason I open issues with the different warning types to split the problem into subtasks.

The task is: add these warnings to the ENABLE_EXTRA_WARNINGS blocks, both for autotools and cmake. And fix these warnings.

It is not necessary to fix everything in one round. We can split further this ticket.

```
0    -Wbad-function-cast
1077 -Wconversion
655  -Wmissing-prototypes
220  -Wsign-compare
1455 -Wsign-conversion
0    -Wmissing-declarations
0    -Woverride-init
0    -Winline
144  -Wcast-qual
6    -Wformat-nonliteral -Wformat

```
(The numbers in the statistics are just an approximate)

Hint:
- autotools
Add this to configure: `../syslog-ng/configure CFLAGS="-Werror" --enable-extra-warnings ...`
- cmake
Add this to cmake call: `cmake ../syslog-ng -DENABLE_EXTRA_WARNINGS=yes -DCMAKE_C_FLAGS="-Werror" ...`

furiel:
I updated the description with the remaining warnings from #1302.

alltilla:
I can see that some of this are already fixed, but there's some which are yet to be fixed.

@furiel: Is it possible to make another round of `warning elimination` issues with the remaining flags? I mean, are the not-yet handled flags in game, still?
@vivinperis: If it is possible, would you care to create some issues about those flags? :)

furiel:
I update the description, because me seems some of the warnings are fixed. Also, I am almost sure my statistics are wrong. According to `cloc $(git ls-files)`, we have `132350` lines of C code. I cannot believe that there are 5K sign conversion related warnings.

vivinperis:
I'll look into it @alltilla

furiel:
I updated the description. I gave another thought about the individual warnings, and I removed some:

- `-Waggregate-return`. 110 occurrences. Warns if a function returns a structure. We are using it in a lot of places. I think we cannot eliminate it reasonably It does not give too much trouble either, unless large size of data needs to be copied due to this.
- `-Wcast-qual`. 144 occurrences. Warns when you remove const keyword with a pointer casting.
- `-Wredundant-decls`. 92 occurrences. Warns if if anything is declared multiple times. Our grammar constructs (parsers, debug flags) fall this case. I think we cannot remove them easily, nor they cause too much trouble.
[...]

PR #1996, opened 2018-04-17
floating point warning fixes

### LINKED ISSUE #1959: warning elimination: floating point related warnings

This issue should cover the floating point related warnings.
```
      5 [-Wdouble-promotion]
      4 [-Wfloat-equal]
```

mehul-m-prajapati:
Fixes #1959


kira-syslogng:

kira-syslogng:

furiel:
@kira-syslogng ok to test

mehul-m-prajapati:
Hi @furiel ,

I see **gcc: error: unrecognized command line option ‘-Wfloat-conversion’**

How can I resolve it ?

Thanks,
Mehul

gaborznagy:
@mehul-m-prajapati 
Our travis CI uses Ubuntu Trusty which has gcc 4.8. In gcc 4.8 there is no -Wfloat-conversion option, although -Wconversion exists.
In 5.4 gcc it is documented that Wfloat-conversion is turned on by Wconversion:
https://gcc.gnu.org/onlinedocs/gcc-5.4.0/gcc/Warning-Options.html#Warning-Options
>-Wfloat-conversion
Warn for implicit conversions that reduce the precision of a real value. This includes conversions from real to integer, and from higher precision real to lower precision real values. This option is also enabled by -Wconversion. 
I suggest to use Wconversion instead.

mehul-m-prajapati:
Hi @furiel ,

There are 1685 [-Wconversion] warnings.

I believe that we should have a separate PR for this only.

Thanks,
Mehul

kira-syslogng:
success

kira-syslogng:
success

gaborznagy:
@mehul: You are right Wconversion is in #1965.

Nevertheless if you would like you fix Wconversion too since you started it, just put it in a second commit in this PR for easier review.
You can also split #1965 and move Wconversion into a new issue!
What do you think?

mehul-m-prajapati:
Hi @gaborznagy ,

I will open up a new issue for -Wconversion. Then, I will do a PR to fix it ...

Thanks,
Mehul

kira-syslogng:
failure

lib/type-hinting.c
```diff
@@ -145,9 +145,9 @@ type_cast_to_double(const gchar *value, gdouble *out, GError **error)
 
   errno = 0;
   *out = strtod(value, &endptr);
-  if (errno == ERANGE && (*out == HUGE_VAL || *out == -HUGE_VAL))
+  if (errno <= ERANGE && (*out <= HUGE_VAL || *out >= -HUGE_VAL))
```
furiel: errno should stay ==

On the other two instances, you used wrong comparison. For example `*out >= HUGE_VAL` should have been used.

Mathematically, `double == HUGE_VAL` <=> `(double <= HUGE_VAL and double >= HUGE_VAL)`. Now `double <= HUGE_VAL` is true by definition. So in the end: `double == HUGE_VAL` <=> `double >= HUGE_VAL`.

kira-syslogng:
failure

lib/type-hinting.c
mehul-m-prajapati: ok

mehul-m-prajapati:
Hi @furiel ,

I am not able to resolve following error,
```
../modules/pacctformat/pacct-format.c: In function ‘pacct_format_handler’:
../modules/pacctformat/pacct-format.c:60:30: warning: implicit conversion from ‘float’ to ‘double’ when passing argument to function [-Wdouble-promotion]
 #define PACCT_CONVERT_NOP(x) (x)
                              ^
../modules/pacctformat/pacct-format.c:69:54: note: in expansion of macro ‘PACCT_CONVERT_NOP’
     __len = g_snprintf(__buf, sizeof(__buf), format, convert(rec->field)); \
                                                      ^
../modules/pacctformat/pacct-format.c:74:3: note: in expansion of macro ‘PACCT_FORMAT_CONVERT’
   PACCT_FORMAT_CONVERT(msg, rec, field, format, PACCT_CONVERT_NOP)
   ^
../modules/pacctformat/pacct-format.c:139:3: note: in expansion of macro ‘PACCT_FORMAT’
   PACCT_FORMAT(msg, rec, ac_etime, "%9.2f");
```

lib/type-hinting.c
furiel: That's right, this is the one!

kira-syslogng:
success

kira-syslogng:
success

kira-syslogng:
failure

kira-syslogng:
success

lib/tests/test_lexer.c
```diff
@@ -156,7 +156,7 @@ _current_lloc(void)
 #define assert_parser_float(expected)                           \
   _next_token();                                                        \
   assert_token_type(LL_FLOAT);                                         \
-  assert_true(_current_token()->fnum == expected, "Bad parsed value at %s:%d", __FUNCTION__, __LINE__);
+  assert_true(_current_token()->fnum >= expected, "Bad parsed value at %s:%d", __FUNCTION__, __LINE__);
```
Kokan: It is not enough to check if the `expected` is smaller than the `fnum`, it should check if the distance between the two number are so small that it is insignificant.

lib/type-hinting.c
```diff
@@ -145,9 +145,9 @@ type_cast_to_double(const gchar *value, gdouble *out, GError **error)
 
   errno = 0;
   *out = strtod(value, &endptr);
-  if (errno == ERANGE && (*out == HUGE_VAL || *out == -HUGE_VAL))
+  if (errno == ERANGE && (*out >= HUGE_VAL || *out <= -HUGE_VAL))
     success = FALSE;
-  if (*out == 0 && endptr == value)
+  if (*out >= 0 && endptr == value)
```
Kokan: The `strtod` could return with `0` for two reason, of which only one of them is when the `value` is `0`. The other reason is handled right here. (See https://linux.die.net/man/3/strtod) Checking if the `0` is larger or equal to zero does not make sense here.

libtest/testutils.c
```diff
@@ -212,7 +212,7 @@ assert_gdouble_non_fatal(gdouble actual, gdouble expected, const gchar *error_me
 {
   va_list args;
 
-  if (actual == expected)
+  if (actual <= expected)
```
Kokan: You cannot simple replace float equal with less equal or larger equal.

lib/type-hinting.c
mehul-m-prajapati: ok. I read the manual and got your point. We can do the following,
```
if ((int)(*out) == 0 && endptr == value)
```

libtest/testutils.c
mehul-m-prajapati: We can do following comparison,
```
if (fabs(actual - expected) < 1e-12)
```

lib/tests/test_lexer.c
mehul-m-prajapati: ok. We can compare the difference like,
```
assert_true((_current_token()->fnum - expected) < 1e-12, "Bad parsed value at %s:%d", __FUNCTION__, __LINE__);

```

Kokan: Just like the example in the 3rd comment, you should use `fabs`.

mehul-m-prajapati: ok. Thanks @Kokan

lib/type-hinting.c
Kokan: Yes.

mehul-m-prajapati:
Hi @Kokan ,

I will explore Criterion and try to plug ```cr_assert_float_eq``` to resolve those asserts.

Thanks,
Mehul

kira-syslogng:
success

Kokan:
The conversion is not just about replacing asserts, this is an example PR which converts test case to criterion:  #1989

kira-syslogng:
success

kira-syslogng:
success

lib/tests/test_lexer.c
```diff
@@ -155,8 +159,8 @@ _current_lloc(void)
 
 #define assert_parser_float(expected)                           \
   _next_token();                                                        \
-  assert_token_type(LL_FLOAT);                                         \
-  assert_true(_current_token()->fnum == expected, "Bad parsed value at %s:%d", __FUNCTION__, __LINE__);
+  cr_assert_token_type(LL_FLOAT);                                         \
+  cr_assert_float_eq(_current_token()->fnum, expected, 1e-12, "Bad parsed value at %s:%d", __FUNCTION__, __LINE__);
```
furiel: Out of curiosity: how is this 1e-12 was chosen?

mehul-m-prajapati: I saw this value in one open source package. 

What do you suggest on epsilon value ?

lib/type-hinting.c
```diff
@@ -145,9 +145,9 @@ type_cast_to_double(const gchar *value, gdouble *out, GError **error)
 
   errno = 0;
   *out = strtod(value, &endptr);
-  if (errno == ERANGE && (*out == HUGE_VAL || *out == -HUGE_VAL))
+  if (errno == ERANGE && (*out >= HUGE_VAL || *out <= -HUGE_VAL))
     success = FALSE;
-  if (*out == 0 && endptr == value)
+  if ((int)(*out) == 0 && endptr == value)
```
furiel: This confused me because I thought we would fail here with a float like 0.01, but the trick is endptr == value would tell us that any kind of parsing is happened. That led me thinking, why do we need to check *out == 0 at all? Would it be enough to check endptr == value? The code would be much cleaner.

tests/loggen/loggen.c
```diff
@@ -456,7 +456,7 @@ gen_messages(send_data_t send_func, void *send_func_ud, int thread_id, FILE *rea
           /* check rate every 0.1sec */
           long new_buckets;
 
-          new_buckets = (rate * diff_usec) / USEC_PER_SEC;
+          new_buckets = (long) ((rate * diff_usec) / USEC_PER_SEC);
```
furiel: CC: @norberttak for #1987 :)

lib/type-hinting.c
mehul-m-prajapati: @furiel : You are correct. We can use endptr == value only to check if no conversion is performed.

lib/tests/test_lexer.c
furiel: Yes you are right. Technically the floating point precision is more than 7, something like (24*log_2(10) = 7.225), so it's not worth go below 1e-7. From mathematical point of view, as this precision is just a precision of the representation, so any result of the arithmetic computation would have worse precision. Personally, I wouldn't have a problem even with a 1e-5, though in this specific assert does not involve arithmetics, just simple parsing, so 1e-7 is fine.

kira-syslogng:
success

tests/loggen/loggen.c
mehul-m-prajapati: @furiel : What do I change here ?

furiel: I left the note for @norberttak because I believe these two requests will conflict on loggen.c. Looking at the changeset in #1987, it is possible that the rebase is not trivial, and this change needs to be integrated manually to the other request. But probably I am overcomplicating things now. The point is: you do not need to do any change here, unless #1987 will go in earlier than your PR. Nevertheless, the race is ON :)

furiel:
I only have one comment with the patchset: the rewrite in criterion is done only partially. One part is rewritten in criterion in the file, but still there is a main function with testcases that should be eliminated. The problem is if go with a hybrid solution, it is very hard to detect later that this rewrite is incomplete. We should either rewrite it completely, or leave it untouched.

furiel:
Good to hear, thank you! As this rewrite might be somewhat complex, let me hint you to check examples with TestSuite and ParametrizedTests in syslog-ng. Thode macros can help a lot.

kira-syslogng:
success

kira-syslogng:
success

mehul-m-prajapati:
Hi @furiel ,

Please review code in test_lexer.c file

Thanks,
Mehul

lib/tests/test_lexer.c
```diff
@@ -127,54 +109,84 @@ _current_lloc(void)
   return parser->yylloc;
 }
 
-#define assert_token_type(expected)                                     \
-  assert_gint(_current_token()->type, expected, "Bad token type at %s:%d", __FUNCTION__, __LINE__);
+static void
```
furiel: The correct expansion of __FUNCTION__ and __LINE__ requires this assert to be a macro. With this change __FUNCTION__ would be replaced by `"assert_token_type"` unconditionally, losing the name of the calling function

mehul-m-prajapati: ok. Do I need to replace all assert functions as macros ?

kira-syslogng:
success

lib/tests/test_lexer.c
furiel: The guiding principle is to use functions whenever possible for sake of simple code. Though there are cases when macro is necessary: either because it is impossible to solve the problem with functions (see #1999 for example when the evaluation of the parameters must be delayed), or using macros would result in a simple/more understandable code. Your specific case belongs to the latter: one can resolve the expansion problem by passing `__FUNCTION__` and `__LINE__` as parameter to a regular function, but it would be inconvenient, and it increases the code complexity of the internal api.

lib/tests/test_lexer.c
```diff
@@ -128,53 +110,80 @@ _current_lloc(void)
...
-#define assert_parser_string(expected)                          \
-  _next_token();                                                        \
-  assert_token_type(LL_STRING);                                        \
-  assert_string(_current_token()->cptr, expected, "Bad parsed string at %s:%d", __FUNCTION__, __LINE__); \
+static void
+assert_parser_string(gchar *expected)
+{
+  _next_token();
+  assert_token_type(LL_STRING);
+  cr_assert_str_eq(_current_token()->cptr, expected, "Bad parsed string at %s:%d", __FUNCTION__, __LINE__);
```
furiel: This change suffers the same expasion problem we discussed in assert_token_type. This should be a macro.

mehul-m-prajapati: ok. done

kira-syslogng:
success

lib/tests/test_lexer.c
```diff
@@ -128,53 +110,77 @@ _current_lloc(void)
...
-#define assert_parser_block(expected) \
-  _next_token();                                                        \
-  assert_token_type(LL_BLOCK);                                         \
-  assert_string(_current_token()->cptr, expected, "Bad parsed string at %s:%d", __FUNCTION__, __LINE__);
+static void
+assert_parser_block(gchar *expected)
+{
+  _next_token();
+  assert_token_type(LL_BLOCK);
+  cr_assert_str_eq(_current_token()->cptr, expected, "Bad parsed string at %s:%d", __FUNCTION__, __LINE__);
```
furiel: This is another occurrence for broken expansion.  Please reevaluate the patch and cleanup all the other places.

lib/tests/test_lexer.c
```diff
@@ -128,53 +110,77 @@ _current_lloc(void)
...
+}
 
-#define assert_parser_block_bad(parser) \
-  _next_token();                                                        \
+static void
+assert_parser_block_bad(void)
+{
+  _next_token();
   assert_token_type(LL_ERROR);
```
furiel: This is a tricky one, because the `__FUNCTION__` and `__LINE__` does not occur directly, but in the end assert_token_type is a macro, so implicitely, these compiler variables occur here too.

lib/tests/test_lexer.c
mehul-m-prajapati: ok. Rechecking all and committing again.

kira-syslogng:
success

kira-syslogng:
success

lib/tests/test_lexer.c
```diff
@@ -128,17 +110,17 @@ _current_lloc(void)
...
-  assert_string(_current_token()->cptr, expected, "Bad parsed string at %s:%d", __FUNCTION__, __LINE__); \
+  cr_assert_str_eq(_current_token()->cptr, expected, "Bad parsed string at %s:%d", __FUNCTION__, __LINE__); \
 
 #define assert_parser_block(expected) \
   _next_token();                                                        \
   assert_token_type(LL_BLOCK);                                         \
-  assert_string(_current_token()->cptr, expected, "Bad parsed string at %s:%d", __FUNCTION__, __LINE__);
+  cr_assert_str_eq(_current_token()->cptr, expected, "Bad parsed string at %s:%d", __FUNCTION__, __LINE__);
 
 #define assert_parser_block_bad(parser) \
```
Kokan: You should remove the `parser` parameter, as it is not passed when called, nor used in the macro expansion.

lib/tests/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ add_unit_test(LIBTEST TARGET test_parse_number)
 add_unit_test(LIBTEST TARGET test_reloc)
 add_unit_test(LIBTEST TARGET test_hostname)
 add_unit_test(LIBTEST TARGET test_rcptid)
-add_unit_test(LIBTEST TARGET test_lexer)
+add_unit_test(CRITERION TARGET test_lexer)
```
Kokan: This modification should be in the other patch.

libtest/testutils.c
```diff
@@ -212,7 +213,11 @@ assert_gdouble_non_fatal(gdouble actual, gdouble expected, const gchar *error_me
 {
   va_list args;
 
-  if (actual == expected)
+  /* NOTE: inf means infinity i.e. 1/0.00 */
```
Kokan: This comment does not have any useful information/explanation, therefore should not exists. Please remove.

kira-syslogng:
success

tests/loggen/loggen.c
norberttak: I addedd this modification to #1987 manually.

Issue #1990, opened 2018-04-15
unsafe usage of evt_tag_errno


furiel:
There are `evt_tag_errno` calls all around the code that use the `errno` variable directly. In some cases it is not safe, because other evt_tags in msg_(error/warning/etc) can call functions that can overwrite errno. For example in the snippet below, either malloc or strdrup in evt_tag_str can overwrite errno.

As compiler is free to evaluate parameters in any order, it is not enough to move evt_tag_errno to the first parameter. 

```
      cap_text = cap_to_text(caps, NULL);
      msg_error("Error managing capability set, cap_set_proc returned an error",
                evt_tag_str("caps", cap_text),
                evt_tag_errno("error", errno));
```

The resolution would be to store errno on the stack, and pass the local variable to errno.

As a bonus: evt_tag_errno might disallow users to pass errno directly: for example compare if `&errno` is passed to it, and assert in that case. G_LIKELY/UNLIKELY can be used inside the if statement for optimization. 

As an estimate, 64/76 of evt_tag_errno uses errno directly.
```
$ ag evt_tag_errno | grep -F "errno)" | wc -l
64
$ ag evt_tag_errno | wc -l
76
```

PR #1999, opened 2018-04-20
errno safe evtlog

furiel:
Fixes: https://github.com/balabit/syslog-ng/issues/1990

Problem summary: the evaluation of `evt_tags` can happen in any order, any of which can modify `errno`, which makes hard to reason about code correctness when `evt_tag_errno` is used. More detailed explanation and example snippet is written in the description of #1990.

The resolution is:
- evtlog api is untouched, which is good because we will not break anything. Users can still use `evt_tag_errno(<string-literal>, int errno)` onwards.
- a new call is added: `evt_tag_error(<string_literal>)` to the `messages` module. The `messages` module is modified that `evt_tag_error` grabs `errno` safely.

Some technical info for developers
- By safely, I mean `msg_error` etc macros expand into a block which will capture `errno` in a block-local variable. Besides that `evt_tag_error` expands into an `evt_tag_errno`, but this time the block-local variable is passed, instead of the thread-local `errno`.
- Unit test added, which when was used with evt_tag_errno, showed the failing symthom.
- Originally I wanted the `int __local_copy_of_errno` to be a gensym, but I did not be able to make it. The problem is the gensym should have been generated by `CAPTURE_ERRNO` macro, but this same gensym should have been used by the expansion of `evt_tag_errno`. This means `CAPTURE_ERRNO` should have created the `evt_tag_error` macro, but macro creating macros are not supported by the c standard (and gcc was unwilling to compile it as well).
- Luckily there was no shadow variable warning, because the blocks built by `CAPTURE_ERRNO `are disjoint, considering you cannot call `msg_error` in any `evt_tag_*` kind of functions. Nested blocks are not possible, at least this time of writing.
- It is unfortunate, but not because not all expansions will use `__local_copy_of_errno`, in those expansions I got -Wunused-variable. So in the end I needed to add `G_GNUC_UNUSED` to the variable, even though it is a used variable in some expansions. The question naturally arises, am I allowed to use an `__attribute__(unused)` variable in an expansion? It seems the unit test does not have problem with that, so I believe it is fine.

### DIFF
--- a/lib/tests/test_messages.c
+++ b/lib/tests/test_messages.c
@@ -26,6 +26,7 @@
 #include "logmsg/logmsg.h"
 
 #include <criterion/criterion.h>
+#include <errno.h>
 
 TestSuite(test_messages, .init = app_startup, .fini = app_shutdown);
 
@@ -39,4 +40,36 @@ Test(test_messages, simple_test)
 {
   msg_set_post_func(simple_test_asserter);
   msg_error("simple test");
+}
+
+void test_errno_asserter(LogMessage *msg)
+{
+  gchar expected_message[100];
+  g_snprintf(expected_message, sizeof(expected_message), "test errno; error='%s (%d)'", strerror(5), 5);
+  cr_assert_str_eq(log_msg_get_value_by_name(msg, "MESSAGE", NULL), expected_message);
+  log_msg_unref(msg);
+}
+
+Test(test_messages, test_errno)
+{
+  errno = 5;
+  msg_set_post_func(test_errno_asserter);
+  msg_error("test errno",
+            evt_tag_errno("error", errno));
+}
+
+void test_errno_capture_asserter(LogMessage *msg)
+{
+  gchar pattern[100];
+  g_snprintf(pattern, sizeof(pattern), "error='%s (%d)'", strerror(9), 9);
+  const gchar *msg_content = log_msg_get_value_by_name(msg, "MESSAGE", NULL);
+  cr_assert(g_strstr_len(msg_content, -1, pattern), "searching: >>%s<< in >>%s<<", pattern, msg_content);
+  log_msg_unref(msg);
+}
+
+Test(test_messages, test_errno_capture)
+{
+  errno = 9;
+  msg_set_post_func(test_errno_capture_asserter);
+  msg_error("test errno capture", (errno=10,evt_tag_capture_errno("error")));
 }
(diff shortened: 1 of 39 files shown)


kira-syslogng:
success

lib/tests/test_messages.c
```diff
@@ -0,0 +1,77 @@
...
+  cr_assert(g_strstr_len(msg_content, -1, pattern), "searching: >>%s<< in >>%s<<", pattern, msg_content);
+}
+
+Test(test_messages, test_errno_capture)
+{
+  errno = 9;
+  msg_set_post_func(test_errno_capture_asserter);
+#pragma GCC diagnostic ignored "-Wsequence-point"
+  msg_error("test errno capture",
+            evt_tag_int("int", (errno = 10)),
```
Kokan: Instead of depending on the argument processing order, and side effects, you could use something like this:
```diff
diff --git a/lib/tests/test_messages.c b/lib/tests/test_messages.c
index 62e02fe24..25df032b0 100644
--- a/lib/tests/test_messages.c
+++ b/lib/tests/test_messages.c
@@ -68,10 +68,9 @@ Test(test_messages, test_errno_capture)
 {
   errno = 9;
   msg_set_post_func(test_errno_capture_asserter);
-#pragma GCC diagnostic ignored "-Wsequence-point"
+
   msg_error("test errno capture",
-            evt_tag_int("int", (errno = 10)),
-            evt_tag_error("error"),
-            evt_tag_int("int", (errno = 100)));
+            (errno=10,evt_tag_error("error")));
 
+  cr_assert_eq(errno, 10);
 }

```

lib/tests/test_messages.c
```diff
@@ -0,0 +1,42 @@
...
+#include "messages.h"
+#include "apphook.h"
+#include "logmsg/logmsg.h"
+
+#include <criterion/criterion.h>
+
+TestSuite(test_messages, .init = app_startup, .fini = app_shutdown);
+
+void simple_test_asserter(LogMessage *msg)
+{
```
furiel: This should unref msg, otherwise the unit test would memleak.

lib/tests/test_messages.c
furiel: Thanks @Kokan for the hint, I did not know that comma operator. Both the patch and the description is simplified due to this.

furiel:
Resolved the comments, updated the description and removed the WIP flag. Ready to review again.

furiel:
No code changed, just squashed the two commits on the top.

PR #2007, opened 2018-04-28
CMake modernization pt. 1

MrAnno:
## CMake modernization pt. 1

The goal of this PR is to improve our CMake build using "modern features" in CMake (>= 3.5).

Most of these changes aim to clarify and simplify dependency management between libraries/executables. The clarification of dependencies is also good for catching implicit and possibly incorrect dependencies.

*For example, after removing all global "include_directory" commands, it turned out that `lib/compat` should be a separate static library, because it is used by targets that do not link against libsyslog-ng (loggen, secret-storage).*

It will also help making our "core" (everything under the `lib` directory) more modular in the future.

-------------------------------------------

I followed 4 principles while I was working on this PR:

### 1. "Build specification" vs "usage requirements"

CMake can **automatically inherit all usage requirements** of a library when it is linked with `target_link_libraries`.

With `target_link_libraries`, you can inherit, for example, transitive dependencies, *include paths of a dependency*, and compiler features. This means that you don't have to append CFLAGS and LDFLAGS manually with `${SOMELIBRARY_INCLUDE_DIR}` and `${SOMELIBRARY_LIBRARIES}`.
Of course this only works for targets that supports the new "build specification" approach.

Glib and ivykis currently **do not support** the new approach, so we have to write this:
```
target_include_directories(syslog-ng
    ${GLIB_INCLUDE_DIRS}
    ${IVYKIS_INCLUDE_DIR}
)

target_link_libraries(syslog-ng
    ${GLIB_GMODULE_LIBRARIES}
    ${GLIB_GTHREAD_LIBRARIES}
    ${GLIB_LIBRARIES}
    ${IVYKIS_LIBRARY}
)
```

OpenSSL **supports** the new approach, so we can write this:
```
target_link_libraries(syslog-ng
    OpenSSL::SSL
    OpenSSL::Crypto
)
```
That's all, you don't have to specify include directories or bother with transitive dependencies.

To achieve such elegance above, our modules should "advertise" their build specifications.

For example, eventlog can tell to its users to inherit a include directory, so you can `#include <evtlog.h>` without setting any include paths on the user side:
```
target_include_directories(eventlog SYSTEM INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/src)
```

Or another example: libsyslog-ng links against eventlog *publicly*, so you can use `evt_tag_str` and other functions in syslog-ng modules without mentioning eventlog in the module's CMake file (our without setting a global include/linker flag):
```
target_link_libraries(syslog-ng PUBLIC eventlog)
```

### 2. Using variables are prone to error (an unset variable expands to empty string)
  - alternative: use properties and targets, use variables only when necessary

### 3. Targets can be considered as OOP objects with visibility modifiers
  - add_executable and add_library "construct" an object
  - target_include_directories, set_target_properties, ... are "member functions"
  - target properties are "member fields"
  - target_link_libraries can be used to link and "inherit" things

### 4. Do
  - NOT set CFLAGS, LDFLAGS explicitly when possible. alternative: use compile features, and everything from 3.
  - NOT use `add_dependencies` after `target_link_libraries`, it's not necessary.
  - use `PROJECT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR` (to support nested projects)


kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
failure

furiel:

lib/eventlog/CMakeLists.txt
```diff
@@ -1,15 +1,15 @@
...
+  src/evtctx.c
+  src/evtfmt.c
+  src/evtout.c
+  src/evtrec.c
+  src/evtstr.c
+  src/evtsyslog.c
+  src/evttags.c
 )
 
+target_include_directories(eventlog SYSTEM INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/src)
```
furiel: evtlog should not -isystem her own include directory.

kira-syslogng:
success

lib/eventlog/CMakeLists.txt
MrAnno: It's not its own include directory, its here for the users of evtlog (interface).

furiel: Thanks for explanation!

kira-syslogng:
success

modules/dbparser/tests/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
...
-  DEPENDS patterndb basicfuncs syslogformat)
-add_unit_test(TARGET test_radix INCLUDES ${PATTERNDB_INCLUDE_DIR} DEPENDS patterndb)
+add_unit_test(TARGET test_timer_wheel DEPENDS patterndb)
+add_unit_test(TARGET test_patternize DEPENDS patterndb syslogformat)
+add_unit_test(LIBTEST TARGET test_patterndb DEPENDS patterndb basicfuncs syslogformat)
+add_unit_test(TARGET test_radix DEPENDS patterndb)
 target_compile_options(test_radix PRIVATE "-Wno-error=pointer-sign")
-add_unit_test(LIBTEST TARGET test_parsers INCLUDES ${PATTERNDB_INCLUDE_DIR})
+# test_parsers includes a .c file
+add_unit_test(LIBTEST TARGET test_parsers INCLUDES "${CMAKE_CURRENT_SOURCE_DIR}/../")
```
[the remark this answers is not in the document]
MrAnno: Nice catch. I'll check it.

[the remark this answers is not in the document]
MrAnno: Oh okay.. so as I noted in line 6, this test includes a .c file instead of linking to dbparser and testing the public interface.

`../` looks weird, I'll do something about it.

kira-syslogng:
success

PR #2008, opened 2018-04-29
mainloop-worker: fix thread_id allocation for more than 32 CPUs

bazsi:
We are storing the allocated thread ids as a bit array stored in a
guint64. However when trying to check single bits, we used an int
to shift into the various bit locations. ints are 32 bits, thus
when we shifted 1 << 32, we got 0, which is not correct.

This patch properly adds the LL suffix to those literal numbers, making
sure that we actually shift 64 bit values.

This fixes a crash on machines with more than 32 CPUs/cores.

Originally reported by @czanik in an email.

### DIFF
--- a/lib/mainloop-worker.c
+++ b/lib/mainloop-worker.c
@@ -84,12 +84,12 @@ _allocate_thread_id(void)
     {
       for (id = 0; id < MAIN_LOOP_MAX_WORKER_THREADS; id++)
         {
-          if ((main_loop_workers_idmap[main_loop_worker_type] & (1 << id)) == 0)
+          if ((main_loop_workers_idmap[main_loop_worker_type] & (1LL << id)) == 0)
             {
               /* id not yet used */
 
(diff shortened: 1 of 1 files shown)


kira-syslogng:
failure

lib/mainloop-worker.c
```diff
@@ -84,12 +84,12 @@ _allocate_thread_id(void)
     {
       for (id = 0; id < MAIN_LOOP_MAX_WORKER_THREADS; id++)
         {
-          if ((main_loop_workers_idmap[main_loop_worker_type] & (1 << id)) == 0)
+          if ((main_loop_workers_idmap[main_loop_worker_type] & (1LL << id)) == 0)
```
MrAnno: I think `ULL` would be even better.

Kokan:

kira-syslogng:
success

furiel:
What was the behaviour of the original version? Could it result in a crash? Or just the number of threads were implicitely limited to 32?

PR #2009, opened 2018-05-02
LogPipe: Improved debug logs

mitzkia:
Adding improved debug log for parsers, rewrite rules and logsources.

**Motivation came from**: 
 * after adding new config elements like: default-network-drivers, ewmm, app-parser..., syslog-ng's debug log became hard readable
 * improved debug logs of filters 9e3415b4e49b1addc8fac6eccd9f9c814e395e3d

**Conception**:
 1. Log debug message at the beginning and at the end of queuing methods, like: 
    * log_source_queue(), 
    * log_parser_queue(), 
    * log_rewrite_queue()
 * Debug message should contains:
    * name of the relevant logpipe element
    * rule name
    * location
    * msg pointer
    * result of filter_expr_eval_root() or log_parser_process_message() -> only at the end of queuing methods
 * Example:
```
>>>>>> parser rule evaluation begin; rule='#anon-parser0', location='parser generator app-parser:5:15', msg='0x7fc2f4013e00'
...
<<<<<< parser rule evaluation result; result='UNMATCHED', rule='#anon-parser0', location='parser generator app-parser:5:15', msg='0x7fc2f4013e00'
```

 2. Log a debug message from parsers at the beginning and on failures
 * Example
```
json-parser started message processing; input='AAAAAA', prefix='.cim.', marker='@cim:', msg='0x7fc2f4013e00'
json-parser failed; error='marker not found', input='AAAAAA', marker='@cim:'
```

 3. Log a debug message when a message macro will get a value during:
     * log_msg_parse_legacy()
     * log_msg_parse_syslog_proto()
     * syslog_format_handler()

 * Example:
```
Initial message parsing;
Setting macros on BSD syslog message;
Setting value; macro='.unix.pid', value='3763', msg='0x7ff6c0013e00'
Setting value; macro='.unix.uid', value='1000', msg='0x7ff6c0013e00'
Setting value; macro='.unix.gid', value='1000', msg='0x7ff6c0013e00'
```

**Questions**:
  * What would be a good debug message level from parsers/filters/ when they are failed?
  * Could we display a "parent" logpipe element of an app-parser in a log_pipe_location_tag()?
     * current: `location='parser generator app-parser:5:15'`
     * expected: `location='source generator system app-parser:6:15'`
  * Is it intentional that a location of a FileSource shows as "#unknown" from a log_source_queue()?
  * Would it be better to create a general method for filter_result_tag() -> it will be used also from rewrite-expr.c and parser-expr.c. What would be a good place of this new function: lib/eventlog/src/evtlog.h? What would be a good name of this new function: expression_result_tag() / evt_result_tag()?

### DIFF
--- a/modules/affile/file-reader.c
+++ b/modules/affile/file-reader.c
@@ -340,6 +340,7 @@ file_reader_init_instance (FileReader *self, const gchar *filename,
   self->options = options;
   self->opener = opener;
   self->owner = owner;
+  self->super.expr_node = owner->super.super.expr_node;
 }
 
 FileReader *


lib/rewrite/rewrite-expr.c
```diff
@@ -23,6 +23,7 @@
  */
 
 #include "rewrite/rewrite-expr.h"
+#include "filter/filter-expr.h"
```
furiel: Does this include still necessary? filter_result_tag is commented out in the end.

lib/rewrite/rewrite-expr.c
```diff
@@ -48,15 +53,12 @@ log_rewrite_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_option
...
-      value = log_msg_get_value(msg, self->value_handle, &length);
-      msg_debug("Rewrite expression evaluation result",
-                evt_tag_str("value", log_msg_get_value_name(self->value_handle, NULL)),
-                evt_tag_printf("new_value", "%.*s", (gint) length, value),
-                evt_tag_str("rule", self->name),
-                log_pipe_location_tag(s));
-    }
+  value = log_msg_get_value(msg, self->value_handle, &length);
+  msg_debug("<<<<<< rewrite rule evaluation result",
+            // filter_result_tag(res),
```
furiel: Please delete this comment.

lib/rewrite/rewrite-expr.c
```diff
@@ -48,15 +53,12 @@ log_rewrite_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_option
...
-  if (G_UNLIKELY(debug_flag))
-    {
-      value = log_msg_get_value(msg, self->value_handle, &length);
-      msg_debug("Rewrite expression evaluation result",
-                evt_tag_str("value", log_msg_get_value_name(self->value_handle, NULL)),
-                evt_tag_printf("new_value", "%.*s", (gint) length, value),
-                evt_tag_str("rule", self->name),
-                log_pipe_location_tag(s));
-    }
+  value = log_msg_get_value(msg, self->value_handle, &length);
```
furiel: This is an unused variable. Should include in the msg_debug below?

furiel: Please add the if (debug_flag) around this, because this lookup is on a heavy path.

lib/logmsg/logmsg.c
```diff
@@ -524,9 +524,9 @@ log_msg_set_value(LogMessage *self, NVHandle handle, const gchar *value, gssize
   if (_log_name_value_updates(self))
     {
       msg_debug("Setting value",
-                evt_tag_printf("msg", "%p", self),
-                evt_tag_str("name", name),
-                evt_tag_printf("value", "%.*s", (gint) value_len, value));
+                evt_tag_str("macro", name),
```
furiel: Why is macro better than name here? This is a name from a name-value pair.

modules/csvparser/csvparser.c
```diff
@@ -145,13 +149,6 @@ csv_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_o
   gboolean result = csv_scanner_is_scan_finished(&scanner);
   csv_scanner_deinit(&scanner);
 
-  if (!result)
-    {
-      msg_debug("csv-parser() failed to parse all columns from the message, and drop-invalid flag is specified",
```
furiel: The fact that the parsing was unsuccessful is still interesting. Shouldn't we emit a log here?

modules/date/date-parser.c
```diff
@@ -194,13 +198,6 @@ date_parser_process(LogParser *s,
                                                 &msg->timestamps[self->time_stamp],
                                                 input);
 
-  if (!res)
-    {
-      msg_debug("date-parser() failed to parse its input",
```
furiel: Same comment as above.

modules/json/json-parser.c
```diff
@@ -215,10 +215,21 @@ json_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_
...
+            evt_tag_str ("prefix", self->prefix),
+            evt_tag_str ("marker", self->marker),
+            evt_tag_printf("msg", "%p", *pmsg));
   if (self->marker)
     {
       if (strncmp(input, self->marker, self->marker_len) != 0)
-        return FALSE;
+        {
+          msg_debug("json-parser failed",
+                    evt_tag_str ("error", "marker did not found in input"),
```
furiel: Cosmetic: marker did not found -> marker not found

kira-syslogng:
success

modules/xml/xml.c
```diff
@@ -230,8 +234,9 @@ xml_parser_process(LogParser *s, LogMessage **pmsg,
   return TRUE;
 
 err:
-  msg_error("xml: error",
-            evt_tag_str("str", error->message));
+  msg_error("xml-parser failed",
+            evt_tag_str("error", error->message),
+	          evt_tag_int("forwarding message to next pipe element", self->forward_invalid));
```
furiel: evt_tags should not contain spaces. evt_tag_int("forward_invalid", self->forward_invalid); would be enough.

lib/rewrite/rewrite-expr.c
mitzkia: Thanks, I am fixing this.

lib/logmsg/logmsg.c
mitzkia: Yes, I know where is it came from, but I think the "name" is misleading in this case. I would change the reference from name-value (this is an internal association) to macro-value (this association is more message related)
Just compare old - new style.
 * old:
```
Setting value; msg='0x7fc5e0013e00', name='SOURCE', value='s_local'
```
 * new
```
Setting value; macro='SOURCE', value='s_local', msg='0x7f44a8013e00'
```

modules/csvparser/csvparser.c
mitzkia: The unsuccessful parsing will handle from general other place - here:
https://github.com/balabit/syslog-ng/blob/06a54a8020328ae7308bb4fa4359df25d24e90fa/lib/parser/parser-expr.c#L95

modules/date/date-parser.c
mitzkia: Also same comment as above

modules/json/json-parser.c
mitzkia: Thanks, will fix it

modules/xml/xml.c
mitzkia: Thanks, will fix it

lib/rewrite/rewrite-expr.c
mitzkia: The "value" not needed anymore, deleted it

kira-syslogng:
failure

kira-syslogng:
success

lib/rewrite/rewrite-expr.c
mitzkia: Yes we need this part, we will use: filter_result_tag() later -> I remove the comment sign from there.

lib/rewrite/rewrite-expr.c
mitzkia: This part will be used. Comment sign was removed.

kira-syslogng:
success

kira-syslogng:
success

lib/logmsg/logmsg.c
algernon: My first instinct was to dislike the `macro` name, because these are key-value pairs, thus `key` might be even better. But there's a very good case for the `macro` name: our documentation refers to these things as such. It makes sense to follow that naming convention in log messages too.

I still don't like that they're called macros, but changing `name` to `macro` here makes our naming more consistent with the documentation, and that's a good thing.

kira-syslogng:
success

lib/logmsg/logmsg.c
```diff
@@ -524,9 +524,9 @@ log_msg_set_value(LogMessage *self, NVHandle handle, const gchar *value, gssize
...
+                evt_tag_printf("value", "%.*s", (gint) value_len, value),
+                evt_tag_printf("msg", "%p", self));
```
algernon: The `evt_tag_printf(<name>, "%p", <pointer>)` construct is used in a fair amount of places. Wouldn't it make sense to introduce an `evt_tag_pointer`, that abstracted away the common parts?

An aesthetic change, but it would convey the intent better, it would be immediately clear what it does, without having to read the whole line.

lib/logmsg/logmsg.c
```diff
@@ -524,9 +524,9 @@ log_msg_set_value(LogMessage *self, NVHandle handle, const gchar *value, gssize
   if (_log_name_value_updates(self))
     {
       msg_debug("Setting value",
-                evt_tag_printf("msg", "%p", self),
-                evt_tag_str("name", name),
-                evt_tag_printf("value", "%.*s", (gint) value_len, value));
+                evt_tag_str("macro", name),
+                evt_tag_printf("value", "%.*s", (gint) value_len, value),
```
algernon: This feels like it could be a new evtlog function, say `evt_tag_strn(const char *tag, gint value_len, const char *value)`... It may not be used a lot, but having a dedicated function for this feels much cleaner.

Mind you, this may not need to be part of *this* pull request, but... let this be a food for thought!

lib/logmsg/logmsg.c
mitzkia: Agreed it would more readable.

lib/logmsg/logmsg.c
mitzkia: Agreed.

lib/rewrite/rewrite-expr.c
bazsi: You shouldn't reuse filter_result_tag() for rewrite and parser expressions, since a parser or rewrite result is conceptually different from a filter result.

I would simply copy-paste the function with a different name in this case. Obviously if it wasn't just one line, I would reuse its implementation, but it is just one line, so copying it is ok.

lib/logmsg/logmsg.c
bazsi: macro is NOT a good name in this context. Once upon a time, we've concluded what would be good names for everything, and we basically said that everything could become a "property", macros would be read-only properties whereas name-value pairs would become just properties.

But until that happens "macros" should not be used, as everywhere else we are referencing the concept of name-value pairs.

How they are referenced unfortunately is not uniform either, sometimes we use "name" in other cases we just use "value" to identify the name value-pair. In the context above, the tag "name" should be ok to understand the message, without introducing a very different way of capturing the same concept.

Introducing the "property" naming would be cool, but I am afraid that's a much larger project that has ties in documentation, configuration syntax and log messages. So large, that it might never happen. But please let's not introduce a new one.

modules/json/json-parser.c
```diff
@@ -215,10 +215,21 @@ json_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_
...
+            evt_tag_str ("marker", self->marker),
+            evt_tag_printf("msg", "%p", *pmsg));
   if (self->marker)
     {
       if (strncmp(input, self->marker, self->marker_len) != 0)
-        return FALSE;
+        {
+          msg_debug("  json-parser failed",
+                    evt_tag_str ("error", "marker not found"),
+                    evt_tag_str ("input", input),
```
bazsi: on the error field:"json marker not found at the beginning of the message"

modules/snmptrapd-parser/snmptrapd-parser.c
```diff
@@ -195,10 +199,20 @@ snmptrapd_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *
...
+                evt_tag_str ("input", input));
+      return FALSE;
+    };
 
   if (!_parse_varbindlist(&nv_context, &input, &input_len))
-    return FALSE;
+    {
+      msg_debug("  snmptrapd-parser failed",
+                evt_tag_str ("error", "can not varbindlist"),
+                evt_tag_str ("input", input));
```
bazsi: can not parse name-value pairs in the input

modules/snmptrapd-parser/snmptrapd-parser.c
```diff
@@ -195,10 +199,20 @@ snmptrapd_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *
   log_msg_set_value(nv_context.msg, LM_V_PROGRAM, "snmptrapd", -1);
 
   if (!snmptrapd_header_parser_parse(&nv_context, &input, &input_len))
-    return FALSE;
+    {
+      msg_debug("  snmptrapd-parser failed",
+                evt_tag_str ("error", "can not parse header"),
+                evt_tag_str ("input", input));
```
bazsi: cannot parse snmptrapd header

modules/syslogformat/syslog-format.c
```diff
@@ -1030,6 +1030,7 @@ log_msg_parse_legacy(const MsgFormatOptions *parse_options,
       goto error;
     }
 
+  msg_debug("Setting macros on BSD syslog message");
   log_msg_parse_seq(self, &src, &left);
```
bazsi: What's the added value of this message? I guess there's an "initial parse starts" message, isn't there?

bazsi: Also, please don't use the "macros" term.

modules/syslogformat/syslog-format.c
```diff
@@ -1170,7 +1171,7 @@ log_msg_parse_syslog_proto(const MsgFormatOptions *parse_options, const guchar *
     {
       return log_msg_parse_legacy(parse_options, data, length, self, position);
     }
-
+  msg_debug("Setting macros on IETF syslog message");
   if (!log_msg_parse_skip_space(self, &src, &left))
```
bazsi: similar to what I had on BSD

lib/logmsg/logmsg.c
mitzkia: Ok. 
Thanks for the explanation. I will change it back to "name".

lib/rewrite/rewrite-expr.c
mitzkia: Thanks. I have created parser_result_tag() and rewrite_result_tag() functions.

lib/logmsg/logmsg.c
```diff
@@ -1272,6 +1272,7 @@ log_msg_new(const gchar *msg, gint length,
 
   if (G_LIKELY(parse_options->format_handler))
     {
+      msg_debug("Initial message parsing");
       parse_options->format_handler->parse(parse_options, (guchar *) msg, length, self);
```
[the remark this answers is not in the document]
mitzkia: Thanks, fixed.

lib/parser/parser-expr.c
```diff
@@ -82,19 +83,24 @@ log_parser_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options
...
 
+  msg_debug(">>>>>> parser rule evaluation begin",
+            evt_tag_str("rule", self->name),
+            log_pipe_location_tag(s),
+            evt_tag_printf("msg", "%p", msg));
+
   success = log_parser_process_message(self, &msg, path_options);
+  msg_debug("<<<<<< parser rule evaluation result",
+            filter_result_tag(success),
+            evt_tag_str("rule", self->name),
```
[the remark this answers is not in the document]
mitzkia: Thanks, fixed.

modules/add-contextual-data/add-contextual-data.c
```diff
@@ -126,6 +126,9 @@ _process(LogParser *s, LogMessage **pmsg,
 {
   AddContextualData *self = (AddContextualData *) s;
   LogMessage *msg = log_msg_make_writable(pmsg, path_options);
+  msg_debug("  add-contextual-data started message processing",
+            evt_tag_str ("input", input),
```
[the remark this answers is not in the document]
mitzkia: Thanks, fixed.

modules/csvparser/csvparser.c
```diff
@@ -125,6 +125,10 @@ csv_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_o
   CSVParser *self = (CSVParser *) s;
   LogMessage *msg = log_msg_make_writable(pmsg, path_options);
 
+  msg_debug("  csv-parser started message processing",
+            evt_tag_str ("input", input),
```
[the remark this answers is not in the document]
mitzkia: Thanks, fixed.

modules/date/date-parser.c
```diff
@@ -183,6 +183,10 @@ date_parser_process(LogParser *s,
 {
   DateParser *self = (DateParser *) s;
   LogMessage *msg = log_msg_make_writable(pmsg, path_options);
+  msg_debug("  date-parser started message processing",
+            evt_tag_str ("input", input),
```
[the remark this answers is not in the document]
mitzkia: Thanks, fixed.

modules/date/date-parser.c
```diff
@@ -183,6 +183,10 @@ date_parser_process(LogParser *s,
...
+            evt_tag_int ("time_stamp", self->time_stamp),
+            evt_tag_printf("msg", "%p", *pmsg));
```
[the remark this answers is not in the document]
mitzkia: Thanks, fixed.

modules/json/json-parser.c
mitzkia: Thanks, fixed.

modules/snmptrapd-parser/snmptrapd-parser.c
mitzkia: Thanks, fixed.

modules/snmptrapd-parser/snmptrapd-parser.c
mitzkia: Thanks, fixed.

modules/syslogformat/syslog-format.c
mitzkia: Thinking it again there is no additional value in this message, I will delete it.
It was just good to know if there was a legacy message parsing or syslog message parsing.

modules/syslogformat/syslog-format.c
mitzkia: Thinking it again there is no additional value in this message, I will delete it.

mitzkia:
I have updated the PR with the review fixes. 
Still remaining open questions:

  * What would be a good debug message level from parsers/filters/ when they are failed?
  * Could we display a "parent" logpipe element of an app-parser in a log_pipe_location_tag()?
     * current: `location='parser generator app-parser:5:15'`
     * expected: `location='source generator system app-parser:6:15'`
  * Is it intentional that a location of a FileSource shows as "#unknown" from a log_source_queue()?

kira-syslogng:
success

bazsi:
-- 
Bazsi

On Mon, May 7, 2018 at 10:38 AM, mitzkia <notifications@github.com> wrote:

> I have updated the PR with the review fixes.
> Still remaining open questions:
>
>    - What would be a good debug message level from parsers/filters/ when
>    they are failed?
>
>
well, the problem is that the answer is it depends.

   - if they are part of a condition (e.g. we are trying to identify which
   application a log is coming from), then the message level should be debug
   (so it does not show in production)
   - if they are processing messages once the condition has been met, the
   debug level should be higher, but probably rate controlled. As in this
   case, classification has failed.
   - the issue is that we have no way to know which of the above happened.
   - I am thinking about adding a notice/info log messages when none of the
   if() branches match the message, because that might be interesting. but
   again, this would produce a lot of messages, so probably some kind of
   suppression would be needed.

In the context of this branch, I would keep it as "debug", and maybe
reiterate this problem as an issue of its own. Should it ever become so
important that we want to address it.

mitzkia:
Thank you for the answer.

furiel:
I have only one last comment:

```
msg_error("  xml-parser failed", ...
msg_debug("<<<<<< rewrite rule evaluation result", ...
```

Probably the `  ` (spaces) and `<<<<<<` are added to help the eye to structure the logs. As such, I feel the patchset is highly optimized for the case when syslog is running in the foreground. When I think of `internal()` source, they would seem awkward.

MrAnno:
> - Is it intentional that a location of a FileSource shows as "#unknown" from a log_source_queue()?

I think it is not, it's a bug. `FileReader` should steal its `expr_node` from its owner.

kira-syslogng:
success

kira-syslogng:
success

kira-syslogng:
success

kira-syslogng:
success

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
Build FAILURE

pzoleex:

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

PR #2003, opened 2018-04-25
Rewrite some unit tests to criterion based

Kokan:
This PR has a few changes that is not a criterion conversion. 
* Tweaks the Travis.
* "Performance" tests take too long, while does not catch perf issues at all, so the iteration is decreased 
* The control command tests should use only the interface and shall not include `. c` files either.

PR #2011, opened 2018-05-03
CMake compile issue because of PR #2003 and #2007

Kokan:
The CMake modernization PR changed some include paths, and the UT refactor used the old one the lack of rebasing causes a compile error on the master. This PR fixes the include path in the test source file.


MrAnno:
Our master is "red", I'm merging this with only one review.

kira-syslogng:
success

PR #2012, opened 2018-05-04
ivykis: update to 0.42.3

MrAnno:
- Explicitly build with -D_REENTRANT on Solaris.
- Don't install IV_FD_PUMP_INIT.3 when on a case sensitive filesystem.

From Lennert Buytenhek:
> There are two bugfixes in this version relative to 0.42.2, one for
case-insensitive filesystems, where the presence of both
IV_FD_PUMP_INIT.3 and iv_fd_pump_init.3 would confuse the build system,
and one for Solaris, where we would fail to use the threadsafe version
of 'errno' ever since we removed the hard dependency on libpthread.


kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
failure

kira-syslogng:
success

kira-syslogng:
success

kira-syslogng:
success

MrAnno:
@kira-syslogng do perftest

furiel:
> Don't install IV_FD_PUMP_INIT.3 when on a case sensitive filesystem.
The effect of this is that we can build ivykis from submodule on osx from now?

MrAnno:
@furiel Yes. I wanted to enable the internal ivykis for macOS in this PR, but 2 wildcard tests failed.
I'll check them separately.

MrAnno:
However, `brew` has updated their ivykis to 0.42.3 a few minutes ago, so we are already using the new version in Travis (macOS).

kira-syslogng:
Performance test: Build SUCCESS, the tests were executed on test branch: master and test suite: functions

PR #2016, opened 2018-05-07
Re-write unit tests based on criterion lib

norberttak:
test_parse_number
test_string_list
test_type_hints

### DIFF
--- a/lib/tests/test_type_hints.c
+++ b/lib/tests/test_type_hints.c
@@ -20,180 +20,255 @@
  * COPYING for details.
  *
  */
-#include "testutils.h"
+
 #include "type-hinting.h"
 #include "apphook.h"
+#include <glib.h>
 
+#include <criterion/criterion.h>
+#include <stdlib.h>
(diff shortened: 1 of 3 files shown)


lib/tests/test_type_hints.c
```diff
@@ -20,180 +20,255 @@
...
+    {"int32",     TYPE_HINT_INT32},
+    {"int64",     TYPE_HINT_INT64},
+    {"double",    TYPE_HINT_DOUBLE},
+    {"datetime",  TYPE_HINT_DATETIME},
+    {"default",   TYPE_HINT_DEFAULT},
+    {NULL,0} // only for terminating the array
+  };
+
+  int i=0;
+  while (string_value_pairs[i].string)
```
Kokan: There is an option to create Parameterized test case in criterion, it is very similar to what you created but utilize the criterion functions (handles the loop itself).

kira-syslogng:
failure

PR #2017, opened 2018-05-08
LogTransportMock enhancement

furiel:
This patchset adds a few enhancements to LogTransportMock:
- Exposing a new `log_transport_mock_inject_data` function, which can be used to add data to the transport. Originally, all data must have been added in the constructor. Now data can be injected after LogTransportMock is created
- General purpose user data is added to LogTransportMock, which can be used to carry test-specific data in the mock. For example one can give a name to the mock.
- Support for cloning LogTransportMock. Cloning can be used for testing drivers that can replace their transport on the fly. In that case, clone can carry the already injected and written data, that will be available after changing transport.
- Removing restriction of the number of injectable data chunks. Originally it was limited to 32.
- Support for two-way transports: a new write_buffer and write function is added. For drivers which not only read but also write the transport, tests can access the written data. Users can either read data from write buffer as stream, or they can fetch data by chunks.

### DIFF
--- a/libtest/mock-transport.c
+++ b/libtest/mock-transport.c
@@ -67,13 +67,38 @@ struct _LogTransportMock
   gpointer user_data;
 };
 
+static void
+destroy_write_buffer_element(gpointer d)
+{
+  data_t *data = (data_t *)d;
+  if (data->type == DATA_STRING)
+    g_free((gchar *)data->iov.iov_base);
+};
+
+static GArray *
+clone_write_buffer(GArray *orig_write_buffer)
+{
+  GArray *write_buffer = g_array_new(TRUE, TRUE, sizeof(data_t));
+  g_array_set_clear_func(write_buffer, destroy_write_buffer_element);
+  g_array_append_vals(write_buffer, orig_write_buffer->data, orig_write_buffer->len);
+  for (int i = 0; i < write_buffer->len; i++)
+    {
+      data_t *data = &g_array_index(write_buffer, data_t, i);
+      if (data->type == DATA_STRING)
+        data->iov.iov_base = g_strndup(data->iov.iov_base, data->iov.iov_len);
+    }
+
+  return write_buffer;
+}
+
 LogTransportMock *
 log_transport_mock_clone(LogTransportMock *self)
 {
   LogTransportMock *clone = g_new(LogTransportMock, 1);
   *clone = *self;
-  clone->write_buffer = g_array_new(TRUE, TRUE, 1);
-  g_array_append_vals(clone->write_buffer, self->write_buffer->data, self->write_buffer->len);
+
+  clone->write_buffer = clone_write_buffer(self->write_buffer);
+
   clone->value = g_array_new(TRUE, TRUE, sizeof(data_t));
   g_array_append_vals(clone->value, self->value->data, self->value->len);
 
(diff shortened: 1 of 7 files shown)


kira-syslogng:
success

kira-syslogng:
success

libtest/mock-transport.c
```diff
@@ -188,26 +188,35 @@ log_transport_mock_free_method(LogTransport *s)
...
+  }
+                };
+  g_array_append_val(self->value, data);
+
+}
+
 void
 log_transport_mock_inject_data(LogTransportMock *self, const gchar *buffer, gssize length)
 {
   if (length == LTM_INJECT_ERROR_LENGTH)
```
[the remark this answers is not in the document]
furiel: The unfortunate thing is we cannot get rid of LTM_INJECT_ERROR_LENGTH cleanly because the log_transport_mock constructors support injecting data inside the constructor parameters.

```
LogTransport *
log_transport_mock_records_new(const gchar *read_buffer1, gssize read_buffer_length1, ...)
```

In the constructors, we still need to allow injecting error.

Not that this feature is used: otherwise some unit tests should have failed on your earlier comment when I compared -2 to an unsigned variable. So this is probably a dead code now. The reason might be is that LogTransportMock already contains a very lightweight error simulation: each even (2k+1) reads returns an EAGAIN, each odd (2k) reads return a chunk.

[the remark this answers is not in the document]
MrAnno: Thanks for the explanation.

We could get rid of the -2 values here as well if record objects would be used (for example, MockRecord as an interface with 2 implementations: DataMockRecord and ErrorMockRecord), but this is too much work for this PR. Let's just leave it as is. :+1:

libtest/mock-transport.c
```diff
@@ -124,6 +146,22 @@ log_transport_mock_read_method(LogTransport *s, gpointer buf, gsize count, LogTr
...
+  g_array_append_vals(self->write_buffer, buf, count);
+  return count;
+}
+
+static void
+log_transport_mock_free_method(LogTransport *s)
+{
+  LogTransportMock *self = (LogTransportMock *)s;
+  g_array_free(self->write_buffer, TRUE);
+  log_transport_free_method(s);
```
MrAnno: In `mock_init()`, `log_transport_init_instance` can be called (holds symmetry with `log_transport_free_method`).

libtest/mock-transport.c
```diff
@@ -65,6 +65,19 @@ struct _LogTransportMock
...
 };
 
+LogTransportMock *
+log_transport_mock_clone(LogTransportMock *self)
+{
+  LogTransportMock *clone = g_new(LogTransportMock, 1);
+  *clone = *self;
+  clone->write_buffer = g_array_new(TRUE, TRUE, 1);
+  g_array_append_vals(clone->write_buffer, self->write_buffer->data, self->write_buffer->len);
+  clone->value = g_array_new(TRUE, TRUE, sizeof(data_t));
```
MrAnno: In this context, `value` is not an GArray yet.
This belongs to another patch.

furiel: Thanks, fixed.

kira-syslogng:
success

furiel:
Adding wip flag because currently LogTransportMock is biased, only capable of mocking read oriented tests (i.e. LogProtoServer). I plan to extend the functionality to work for LogProtoClient as well.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng rétest this please

libtest/mock-transport.c
furiel: Done thanks.

furiel:
Now the written data will be available in two ways. Tests can fetch written data either as stream, or they can fetch them by chunk.

Removing the WIP flag. The PR is ready to be reviewed.

kira-syslogng:
Build SUCCESS

furiel:
Force push: I added a small comment near the assert for EAGAIN in the unit test.

PR #2019, opened 2018-05-08
Assert clone exist 

Kokan:
The clone should be defined before the usage, this assert should force it. 

Signed-off-by: kokan <peter.kokai@balabit.com>


kira-syslogng:
success

PR #2027, opened 2018-05-11
DefaultNetworkDrivers: added missing 'else {};' statement to legacy p…

mitzkia:
…art in SCL

 * without this change legacy messages which are not matching on ewmm-parser()
 or on app-parser(topic(syslog)) will _not_ forwarded to destinations,
 they will dropped.


kira-syslogng:
Build SUCCESS

PR #2031, opened 2018-05-13
 Dbld coverage

bazsi:
This patch set makes it possible to generate coverage reports in the devshell image. It contains a set of fixes that:

1) removes python-venv from dbld/bootstrap
2) adds lcov package to the devshell image
3) fixes an lcov compatibility issue that comes up with the lcov version in artful
4) makes it possible to customize dbld options locally from a dlbd/rules.conf file that is included into the main makefile.

With that, if --enable-gcov is added as a configure option, "make coverage" would work from within the devshell image, generating an HTML report into dbld/build/coverage.html

We roughly have 67% coverage with the unit tests alone. Could be better, but some parts look really nice, others less so.

### DIFF
--- a/dbld/images/devshell-apt/all.txt
+++ b/dbld/images/devshell-apt/all.txt
@@ -11,3 +11,4 @@ valgrind
 linux-tools-generic
 libjemalloc-dev
 locales
+lcov
--- a/dbld/rules
+++ b/dbld/rules
@@ -21,6 +21,8 @@ TARBALL=$(BUILD_DIR)/syslog-ng-$(VERSION).tar.gz
 MODE=snapshot
 CONFIGURE_OPTS=--enable-debug --enable-manpages --with-python=2 --prefix=/install --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl $(CONFIGURE_ADD)
 
+-include dbld/rules.conf
+
 help:
 	@echo "This script allows you to build release/snapshot artifacts, such "
 	@echo "as rpm, deb packages and tarballs in controlled environments (e.g. "
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ m4/ax*.m4
 dbld/build/
 dbld/release/
 dbld/install/
+dbld/rules.conf
 debian/build-tree/
 debian/*.log
 debian/autoreconf.*
--- a/dbld/bootstrap
+++ b/dbld/bootstrap
@@ -8,4 +8,3 @@ cd /source
 cd /build
 echo "Running configure with: ${CONFIGURE_OPTS}"
 /source/configure ${CONFIGURE_OPTS}
-make python-venv
--- a/tests/collect-cov.sh
+++ b/tests/collect-cov.sh
@@ -23,7 +23,7 @@
 
 set -e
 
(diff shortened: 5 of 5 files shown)


kira-syslogng:
Build SUCCESS

dbld/images/required-apt/artful.txt
```diff
@@ -6,6 +6,7 @@ libmongoc-dev
 librabbitmq-dev
 libsystemd-dev
 openjdk-8-jdk
+libriemann-client-dev
```
szemere: We install this package from OBS on the following platforms:

$grep -rn libriemann-client-dev
required-obs/trusty.txt:6:libriemann-client-dev
required-obs/xenial.txt:1:libriemann-client-dev
required-obs/jessie.txt:4:libriemann-client-dev

Can we install it from the official repositories on the other platforms? Or should we change this one to?

(Note: Based on the outcome, there is a chance we can move it to the "all.txt")

bazsi:
I would rather decrease dependency on obs packages over time, do if the one
in artful is good enough I would use that.

lbudai:
@bazsi: could you check it and modify the PR accordingly? (I think we could remove obs:libriemann, but we have to test it)

bazsi:
Right now, I am using libriemann from artful instead of obs. Or you mean checking the rest of the platforms?

lbudai:
@bazsi: I know it is not strongly related to this PR but also think that we should check this as we want to less depend on obs.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

dbld/rules
```diff
@@ -21,6 +21,8 @@ TARBALL=$(BUILD_DIR)/syslog-ng-$(VERSION).tar.gz
 MODE=snapshot
 CONFIGURE_OPTS=--enable-debug --enable-manpages --with-python=2 --prefix=/install --with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl $(CONFIGURE_ADD)
 
+-include dbld/rules.conf
```
gaborznagy: Sorry if I missed it, but I don't see this rules.conf file

bazsi: That file can be added locally so you can customize your dbld related options (for me, I override the default docker image to be used and the default bootstrap arguments).

very handy.

bazsi:
I have now rebased and dropped the riemann bits. This should be clear cut now. I don't see how kira couldn't like it :)

Anyhow, I think this is good enough to go in, after a review.

kira-syslogng:
Build SUCCESS

dbld/images/devshell-apt/all.txt
```diff
@@ -11,3 +11,4 @@ valgrind
 linux-tools-generic
 libjemalloc-dev
 locales
+lcov
```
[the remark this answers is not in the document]
bazsi: The agreement last time was that the various images should only install packages strictly required for building and dev tools should not be on them (e.g. not even gdb or an editor).

That's why this went into devshell. I agree that it possibly made sense to get these in travis/kira, but let's migrate at least one of them to use these images and get back to this question at that point.

Until now it is manual only.

PS: I could even live with trying to merge the build/dev usecases and only build one image, but in that case, that image would probably be pretty large.

PR #2032, opened 2018-05-14
LogPipe: Improved debug logs 

mitzkia:
Adding improved debug log for parsers, rewrite rules and logsources.

**Motivation came from**: 
 * after adding new config elements like: default-network-drivers, ewmm, app-parser..., syslog-ng's debug log became hard readable
 * improved debug logs of filters 9e3415b4e49b1addc8fac6eccd9f9c814e395e3d

**Conception**:
 1. Log debug message at the beginning and at the end of queuing methods, like: 
    * log_source_queue(), 
    * log_parser_queue(), 
    * log_rewrite_queue()
 * Debug message should contains:
    * name of the relevant logpipe element
    * rule name
    * location
    * msg pointer
    * result of filter_expr_eval_root() or log_parser_process_message() -> only at the end of queuing methods
 * Example:
```
>>>>>> parser rule evaluation begin; rule='#anon-parser0', location='parser generator app-parser:5:15', msg='0x7fc2f4013e00'
...
<<<<<< parser rule evaluation result; result='UNMATCHED', rule='#anon-parser0', location='parser generator app-parser:5:15', msg='0x7fc2f4013e00'
```

 2. Log a debug message from parsers at the beginning and on failures
 * Example
```
json-parser started message processing; input='AAAAAA', prefix='.cim.', marker='@cim:', msg='0x7fc2f4013e00'
json-parser failed; error='marker not found', input='AAAAAA', marker='@cim:'
```

 3. Log a debug message when a message macro will get a value during:
     * log_msg_parse_legacy()
     * log_msg_parse_syslog_proto()
     * syslog_format_handler()

 * Example:
```
Initial message parsing;
Setting macros on BSD syslog message;
Setting value; macro='.unix.pid', value='3763', msg='0x7ff6c0013e00'
Setting value; macro='.unix.uid', value='1000', msg='0x7ff6c0013e00'
Setting value; macro='.unix.gid', value='1000', msg='0x7ff6c0013e00'
```

### DIFF
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -1272,6 +1272,7 @@ log_msg_new(const gchar *msg, gint length,
 
   if (G_LIKELY(parse_options->format_handler))
     {
+      msg_debug("Initial message parsing follows");
       parse_options->format_handler->parse(parse_options, (guchar *) msg, length, self);
     }
   else
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -503,7 +503,7 @@ _log_name_value_updates(LogMessage *self)
    * log_msg_new_internal() calling log_msg_set_value(), which in turn
    * generates an internal message, again calling log_msg_set_value()
    */
-  return (!self->initial_parse && (self->flags & LF_INTERNAL) == 0);
+  return (self->flags & LF_INTERNAL) == 0;
 }
 
 void
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -524,9 +524,9 @@ log_msg_set_value(LogMessage *self, NVHandle handle, const gchar *value, gssize 
   if (_log_name_value_updates(self))
     {
       msg_debug("Setting value",
-                evt_tag_printf("msg", "%p", self),
                 evt_tag_str("name", name),
-                evt_tag_printf("value", "%.*s", (gint) value_len, value));
+                evt_tag_printf("value", "%.*s", (gint) value_len, value),
+                evt_tag_printf("msg", "%p", self));
     }
 
   if (value_len < 0)
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -1212,6 +1212,10 @@ log_msg_clone_cow(LogMessage *msg, const LogPathOptions *path_options)
   memcpy(self, msg, sizeof(*msg));
   msg->allocated_bytes = allocated_bytes;
 
+  msg_debug("Message was cloned",
+            evt_tag_printf("original_msg", "%p", msg),
+            evt_tag_printf("new_msg", "%p", self));
+
   /* every field _must_ be initialized explicitly if its direct
    * copying would cause problems (like copying a pointer by value) */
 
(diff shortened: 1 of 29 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/parser/parser-expr.h
```diff
@@ -52,6 +52,8 @@ log_parser_deinit_method(LogPipe *s)
   return TRUE;
 }
 
+EVTTAG *parser_result_tag(gboolean res);
```
furiel: Does this need to be part of the public interface? I think this function can be static. Similar comment to the other two variants.

modules/dbparser/dbparser.c
```diff
@@ -218,10 +220,9 @@ log_db_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *pat
 
   if (self->drop_unmatched && !matched)
     {
-      msg_debug("db-parser() failed to parse its input and drop-unmatched flag was specified",
-                evt_tag_str("input", input),
-                log_pipe_location_tag(&s->super),
```
furiel: Is this tag accidentally removed?

modules/python/python-logparser.c
```diff
@@ -182,10 +182,10 @@ python_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *pat
   {
     LogMessage *msg = log_msg_make_writable(pmsg, path_options);
 
-    msg_debug("Invoking the Python parse() method",
+    msg_debug("python-parser message processing started",
+              evt_tag_str ("input", input),
               evt_tag_str("parser", self->super.name),
               evt_tag_str("class", self->class),
-              log_pipe_location_tag(&self->super.super),
```
furiel: Why is the location tag deleted here?

modules/dbparser/dbparser.c
mitzkia: No it was intentional, there would be a redundancy. From now log_pipe_location_tag() would be printed in every case from the very first place: log_source_queue(), log_parser_queue(), log_rewrite_queue(), log_filter_queue().

modules/python/python-logparser.c
mitzkia: In my previous comment I have described why it is deleted from here.

lib/parser/parser-expr.h
mitzkia: Agreed, I will change this.

mitzkia: Fixed.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/logmsg/logmsg.c
```diff
@@ -1212,6 +1212,10 @@ log_msg_clone_cow(LogMessage *msg, const LogPathOptions *path_options)
   memcpy(self, msg, sizeof(*msg));
   msg->allocated_bytes = allocated_bytes;
 
+  msg_debug("Message was cloned",
```
furiel: This should be trace imo.

kira-syslogng:
Build SUCCESS

bazsi:
Sometimes the fact that the message was cloned can be important to
understand debug logs. We include the address of the message for the very
same reason, but that's not that appearent.

kira-syslogng:
Build SUCCESS

PR #2034, opened 2018-05-14
Dbld: Add fixes to functions.sh

mitzkia:
- enable_dbgsyms(): fix keyserver.ubuntu.com URI, without this change fetching can be stuck
 - install_perf(): install OS native version of linux-tools instead of hosts OS version

### DIFF
--- a/dbld/images/helpers/functions.sh
+++ b/dbld/images/helpers/functions.sh
@@ -47,12 +47,12 @@ function enable_dbgsyms {
 deb http://ddebs.ubuntu.com $(lsb_release -cs)-updates main restricted universe multiverse
 deb http://ddebs.ubuntu.com $(lsb_release -cs)-proposed main restricted universe multiverse" | \
     tee -a /etc/apt/sources.list.d/ddebs.list
-    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 428D7C01 C8CAB6595FDFF622
+    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 428D7C01 C8CAB6595FDFF622
     apt-get update
 }
 
 function install_perf {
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

dbld/images/helpers/functions.sh
```diff
@@ -47,12 +47,12 @@ function enable_dbgsyms {
...
     tee -a /etc/apt/sources.list.d/ddebs.list
-    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 428D7C01 C8CAB6595FDFF622
+    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 428D7C01 C8CAB6595FDFF622
     apt-get update
 }
 
 function install_perf {
-    apt-get install --no-install-recommends -y linux-tools-$(uname -r)
+    apt-cache search linux-tools | grep 'linux-tools-.*-generic' | cut -d" " -f1 | tail -n1 | cut -d"-" -f1-4 | xargs apt-get install --no-install-recommends -y
 }
```
bazsi: unfortunately perf needs to match the kernel version that is running the container and not the one which is installed.

mitzkia: The issue was that my host is Ubuntu:18.04, and the OS of devshell image is Ubuntu:Artful.  Artful's package repository does not contain linux-tools version of 4.15.0-20 (which is Ubuntu Bionic's kernel version). Other option we upgrade DBLD devshell's docker OS version to Ubuntu:18.04. But this needs more work.

bazsi:
hmm.. that's bad :( I don't think we can solve this issue in a clean way.
that perf package might install but may not actually work on your kernel.
hmm.. in that case, your fix should be ok. the install finishes and as long
as someone is running under the proper kernel it will work. If not, that's
ok too. this is a developer tool.


-- 
Bazsi

PR #2035, opened 2018-05-14
Fix lloc tracking for multi line blockrefs

bazsi:
The issue I noticed while finishing up a previous PR was that line tracking got broken by 2e196c393e. While fixing the problem I've also improved error logging a bit more, which I've piggybacked to the same branch.

    cfg-lex.l: fix location tracking for multi-line block-refs
    
    For tokens that span multiple lines (in our case this only happens when
    a block-ref is parsed that spans multiple lines), we blew up the line
    number counter, so whenever reporting locations or errors we got an
    incorrect value.
    
    This patch fixes that by shifting to the new line number relative to
    last_line, instead of simply increasing the first_line.
    
### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -543,7 +543,7 @@ rewrite_stmt
 
 log_stmt
         : KW_LOG
-          { cfg_lexer_push_context(lexer, LL_CONTEXT_LOG, NULL, "log"); }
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
I'll fix the style issue and merge, thanks for the review.

MrAnno:
I've fixed it for you :)

kira-syslogng:
Build SUCCESS

PR #2037, opened 2018-05-15
autotools: fixing emitted warnings due to -no-pie detection for gcc 4.4

furiel:
This patchset contains the following changes:
- fixing no-pie detection for gcc 4.4. With version 4.4, gcc cannot be configured to fail on unknown cli parameters. As a result, `-no-pie gcc: error: unrecognized command line option '-no_pie'` is emitted during the build. This did not cause any problem, just an unnecessary warning is emitted.
- Adding an assert for the exciting cpu-cache optimization introduced in  9fc86dfc6de3ad66862218522b2588d4553c70de.
- using glib macro in format string to write uint64 number.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -446,10 +446,17 @@ fi
 dnl ***************************************************************************
 dnl Set up NO_PIE_LDFLAG: -no-pie is compatible with $CC.
 
+AX_COMPILER_VENDOR
+AX_COMPILER_VERSION
+
 if test $ostype = "Darwin"; then
   no_pie_ldflags="-Wl,-no_pie"
 else
-  no_pie_ldflags="-no-pie"
+  if test x$ax_cv_c_compiler_vendor = "xgnu"; then
+    AX_COMPARE_VERSION($ax_cv_c_compiler_version, gt, [4.5], [no_pie_ldflags="-no-pie"], [no_pie_ldflags=""])
+  else
+    no_pie_ldflags="-no-pie"
+  fi;
 fi
 
 old_LDFLAGS="$LDFLAGS"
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

m4/ax_compare_version.m4
```diff
@@ -0,0 +1,177 @@
+# ===========================================================================
+#    https://www.gnu.org/software/autoconf-archive/ax_compare_version.html
+# ===========================================================================
```
MrAnno: Why did you copy `ax_compare_version.m4` and `ax_compiler_vendor.m4` into our repository?
`autoconf-archive` is our hard dependency. Are these macros outdated in distro's official packages?

furiel: I wanted to be sure these macros will be available independently from autoconf-archive version. Though most probably you are right, these are very basic macros, this has low risk. Let me check if we can compile without them.

MrAnno: (I've only checked with latest package in Xenial.)

furiel:
@MrAnno I could build syslog-ng manually without hardcoded ax macros.

Though there is a little inconvenience with the release tarball generation. The process is: tarball generation executes autogen, which will steal the scripts from the host machine, and those versions of the scripts will be put in the release tarball. In the end, the content of the release tarball depends on the environment where the release tarball is created.

kira-syslogng:
Build SUCCESS

MrAnno:
@furiel Yes, that is correct. Our configure script generation completely depends on the environment where the tarball is created because of `autoconf` and `autoconf-archive`.

We could get rid of at least `autoconf-archive`, but then all other macros should be copied into our repo (we also use other macros from this package).

PR #2039, opened 2018-05-15
Rewrite some unittest based on criterion

norberttak:


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lib/tests/test_userdb.c
```diff
@@ -23,34 +23,43 @@
...
-    return 1;
+  cr_assert(resolve_user("0", &uid));
+  cr_assert_eq(uid, 0);
+}
+
+Test(user_db, resolve_group)
+{
+  gint gid;
 
-  if (resolve_group("nincsily", &gid))
```
Kokan: I think this was meant as `resolve_user` instead of group, because later it has a group check.
This is not part of the unit test conversation, but it would be nice if you could add a test for the `resolve_user` also.

lib/tests/test_utf8utils.c
```diff
@@ -22,89 +22,83 @@
...
+    {"árvíztűrőtükörfúrógép", "árvíztűrőtükörfúrógép", NULL, -1},
+    {"árvíztűrőtükörfúrógép\n", "árvíztűrőtükörfúrógép\\n", NULL, -1},
+    {"\x41", "A", NULL, -1},
+    {"\x7", "\\x07", NULL, -1},
+    {"\xad", "\\xad", NULL, -1},
+    {"Á\xadÉ", "Á\\xadÉ", NULL, -1},
+    {"\xc3\x00\xc1""", "\\xc3", NULL, -1},
+    {"\"text\"", "\\\"text\\\"", "\"", -1},
+    {"\"text\"", "\\\"te\\xt\\\"", "\"x", -1},
+    {"\xc3""\xa1 non zero terminated", "\\xc3", NULL, 1},
```
Kokan: Nice refactor :+1:

furiel:
Please edit CMakeLists.txt. These tests need CRITERION flag. If criterion was installed to a non-standard location, we probably wouldn't compile.

kira-syslogng:
Build SUCCESS

lib/userdb.c
```diff
@@ -35,8 +35,11 @@ resolve_user(const char *user, gint *uid)
   struct passwd *pw;
   gchar *endptr;
 
+  if (!uid)
```
furiel: Please use assert here.

lib/userdb.c
```diff
@@ -35,8 +35,11 @@ resolve_user(const char *user, gint *uid)
...
+    return FALSE;
   *uid = 0;
-  if (!(*user))
+
+  if (!user || !(*user))
```
furiel: The !user part should be assert instead. Return for !(*user) with false is fine.

lib/userdb.c
```diff
@@ -57,8 +60,11 @@ resolve_group(const char *group, gint *gid)
   struct group *gr;
   gchar *endptr;
 
+  if (!gid)
```
furiel: This should be an assert.

lib/userdb.c
```diff
@@ -57,8 +60,11 @@ resolve_group(const char *group, gint *gid)
...
+    return FALSE;
   *gid = 0;
-  if (!(*group))
+
+  if (!group || !(*group))
```
furiel: The !group part should be assert. The !(*group) return FALSE part is ok.

kira-syslogng:
Build SUCCESS

norberttak:
I did the requested modifications

lib/userdb.c
norberttak: I restored the original form of userdb.c with no checks for NULL pointers.

PR #2043, opened 2018-05-16
syslog-ng-ctl: fix shifting of the raw_query_params

Kokan:
On ubuntu 18.04 the `syslog-ng-ctl query` command does not work because the command send the wrong data over the control file. The global `raw_query_params` "duplicates" the first parameter.

[Edit] As it turned out it was not related to the distribution, see my comment below.

### DIFF
--- a/syslog-ng-ctl/syslog-ng-ctl.c
+++ b/syslog-ng-ctl/syslog-ng-ctl.c
@@ -298,12 +298,11 @@ _is_query_params_empty(void)
   return raw_query_params == NULL;
 }
 
-static gchar *
+static void
 _shift_query_command_out_of_params(void)
 {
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

lbudai:
why this is a problem only on ubuntu 18.04?

Kokan:
@lbudai it is not, it happened to me after upgrade, and I had other priorities so I only pushed this so far so it works.

It is actually compiler dependent. The clang and the gcc produces different code, the gcc creates a code we are expecting, and the clang does not. BUT both of them are correct, as the order of evaluating the shift function and the left side of the assignment is unspecified. The clang did dereference before the call, and the gcc did after the call.

PR #2044, opened 2018-05-17
mock-transport: fix compiler error for gcc 4.4

furiel:
In gcc 4.4, one cannot initialize unions using compound literals. It results in a compiler error.


kira-syslogng:
Build SUCCESS

PR #2045, opened 2018-05-18
[WIP] Fix window leak in log_source_flow_control_suspend()

MrAnno:
This PR contains 3 fixes that prevents a source getting stuck/leaking the window.

- `g_atomic_counter_set()` was not atomic. I would have called it `g_atomic_counter_init_racy`.
- `AckTracker` did not increase the window size by 1 in case of an `AT_SUSPENDED` ack type.
- `log_source_flow_control_suspend()` was not "atomic enough":

```C
void
log_source_flow_control_suspend(LogSource *self)
{
  msg_debug("Source has been suspended", log_pipe_location_tag(&self->super));

  g_atomic_counter_set(&self->suspended_window_size, /* ! */ g_atomic_counter_get(&self->window_size));
  /* !!! A context switch here and an incrementation of window_size from a different thread
   * make it possible to lose acknowledged "values" from the window
   */
  g_atomic_counter_set(&self->window_size, 0);
  _flow_control_rate_adjust(self);
}
```

Between setting `suspended_window_size` and setting the original window size to 0, it is possible that `window_size` is incremented by another message consumer in a different thread.

This window leak is permanent, it can not be reversed without reloading syslog-ng. Since the problem can occur multiple times in a multithreaded environment, the source may also get stuck.

-------------

Questions:
- **In late-ack-tracker, should I adjust the window size by 1 or `ack_range_length`?**

### DIFF
--- a/lib/atomic.h
+++ b/lib/atomic.h
@@ -32,6 +32,8 @@ typedef struct
   gint counter;
 } GAtomicCounter;
 
+typedef GAtomicCounter GAtomicBool;
+
 static inline void
 g_atomic_counter_inc(GAtomicCounter *c)
 {
@@ -70,6 +72,19 @@ static inline void
 g_atomic_counter_set(GAtomicCounter *c, gint value)
 {
   g_atomic_int_set(&c->counter, value);
+}
+
+
+static inline gboolean
+g_atomic_bool_get(GAtomicBool *b)
+{
+  return !!g_atomic_counter_get(b);
+}
+
+static inline void
+g_atomic_bool_set(GAtomicBool *b, gboolean value)
+{
+  g_atomic_counter_set(b, !!value);
 }
 
 #endif
--- a/lib/atomic.h
+++ b/lib/atomic.h
@@ -25,6 +25,7 @@
 #ifndef ATOMIC_H_INCLUDED
 #define ATOMIC_H_INCLUDED
 
+#include "compat/glib.h"
 
 typedef struct
 {
@@ -68,13 +69,7 @@ g_atomic_counter_racy_get(GAtomicCounter *c)
 static inline void
 g_atomic_counter_set(GAtomicCounter *c, gint value)
 {
-  /* FIXME: we should use g_atomic_int_set, but that's available starting
-   * with GLib 2.10 only, and we only use this function for initialization,
-   * thus atomic write is not strictly needed as there's no concurrency
-   * while initializing a refcounter.
-   */
-
-  c->counter = value;
+  g_atomic_int_set(&c->counter, value);
 }
 
 #endif
--- a/lib/late_ack_tracker.c
+++ b/lib/late_ack_tracker.c
@@ -141,7 +141,10 @@ late_ack_tracker_manage_msg_ack(AckTracker *s, LogMessage *msg, AckType ack_type
         _drop_range(self, ack_range_length);
 
         if (ack_type == AT_SUSPENDED)
-          log_source_flow_control_suspend(self->super.source);
+          {
+            log_source_flow_control_adjust(self->super.source, 1);
+            log_source_flow_control_suspend(self->super.source);
+          }
         else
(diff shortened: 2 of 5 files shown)


MrAnno:
@kira-syslogng do perftest

kira-syslogng:
Build SUCCESS

kira-syslogng:
Performance test: Build SUCCESS, the tests were executed on test branch: master and test suite: functions

MrAnno:
@kira-syslogng do perftest again

kira-syslogng:
Performance test: Build FAILURE, $TEST_DESC_COMMENT

MrAnno:
@kira-syslogng do perftest

kira-syslogng:
Performance test: Build SUCCESS, the tests were executed on test branch: master and test suite: functions

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lib/atomic.h
```diff
@@ -68,13 +71,20 @@ g_atomic_counter_racy_get(GAtomicCounter *c)
...
-   */
+  g_atomic_int_set(&c->counter, value);
+}
+
 
-  c->counter = value;
+static inline gboolean
+g_atomic_bool_get(GAtomicBool *b)
+{
+  return g_atomic_counter_get(b);
```
furiel: Now this return should be bool: `!!g_atomic_counter_get(b);` would be better.

lib/atomic.h
```diff
@@ -68,13 +71,20 @@ g_atomic_counter_racy_get(GAtomicCounter *c)
...
+}
+
+static inline void
+g_atomic_bool_set(GAtomicBool *b, gboolean value)
+{
+  g_atomic_counter_set(b, value);
```
furiel: !!value would be better

lib/logsource.h
```diff
@@ -83,7 +83,8 @@ struct _LogSource
 static inline gboolean
 log_source_free_to_send(LogSource *self)
 {
-  return g_atomic_counter_get(&self->window_size) > 0;
+  return g_atomic_bool_get(&self->forced_suspend) == FALSE
```
furiel: matter of taste, but as g_atomic_bool_get is boolean, I would use `!g_atomic_bool_get(...)`, provided either setter or getter forcefully translates the counter value to gboolean with !!.

lib/atomic.h
furiel: @algernon you are right, overlook from my side.

kira-syslogng:
Build FAILURE

lib/atomic.h
MrAnno: Done, thanks.

MrAnno:
Question:
- In late-ack-tracker, should I adjust the window size by 1 or `ack_range_length`?

kira-syslogng:
Build SUCCESS

lib/late_ack_tracker.c
```diff
@@ -141,7 +141,10 @@ late_ack_tracker_manage_msg_ack(AckTracker *s, LogMessage *msg, AckType ack_type
         _drop_range(self, ack_range_length);
 
         if (ack_type == AT_SUSPENDED)
-          log_source_flow_control_suspend(self->super.source);
+          {
+            log_source_flow_control_adjust(self->super.source, 1);
+            log_source_flow_control_suspend(self->super.source);
+          }
```
lbudai: I think the original code is not correct.
We should suspend immediately when an `AT_SUSPENDED` ack received even if we don't have continuous (acked=true) block.
So we should mark the ringbuffer record as acked, then call suspend.
The adjustment is not required here as the ringbuffer size is not decreasing.

lbudai:
@MrAnno : thanks for the question!!

MrAnno:
Closing in favor of #2167.
I'm going to open a separate PR with bc7f215df45b0c77aeb82a6b1dc673c856e54336.

Issue #1943, opened 2018-03-27
Problem with Java destinations JDK9 and above


sergio-limone:
# syslog-ng

## Version of syslog-ng

syslog-ng 3 (3.14.1)
Config version: 3.14
Installer-Version: 3.14.1
Revision: 
Compile-Date: Feb 28 2018 07:34:20
Module-Directory: //usr/lib64/syslog-ng
Module-Path: //usr/lib64/syslog-ng
Available-Modules: afsocket,json-plugin,cef,stardate,graphite,disk-buffer,sdjournal,afuser,afamqp,afprog,date,linux-kmsg-format,system-source,map-value-pairs,tags-parser,affile,cryptofuncs,xml,dbparser,appmodel,kvformat,snmptrapd-parser,basicfuncs,add-contextual-data,confgen,mod-java,afstomp,syslogformat,pseudofile,csvparser,tfgetent
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: on

## Platform

Linux statia 3.10.0-693.5.2.el7.x86_64 #1 SMP Fri Oct 20 20:32:50 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
Centos 7.4

# Issue

I tried to start elasticsearch2 with jre 9.0.4 , seems the elastic2 destination is not jdk9 ready yet.
Same configuration works perfect with jre1.8.0_161-b12
RPMs are from https://copr-be.cloud.fedoraproject.org/results/czanik/syslog-ng314/

## Failure

Fails to start with java exception on stdout:

WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.syslog_ng.SyslogNgClassLoader (file:/usr/lib64/syslog-ng/java-modules/syslog-ng-core.jar) to method java.net.URLClassLoader.addURL(java.net.URL)
WARNING: Please consider reporting this to the maintainers of org.syslog_ng.SyslogNgClassLoader
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
[2018-03-27T17:01:53.585521] Error while expanding path list: object is not an instance of declaring class;
java.lang.IllegalArgumentException: object is not an instance of declaring class
[...]

Issue #1726, opened 2017-10-19
version 3.12 does not compile on FreeBSD 11-1 p1


redtorinogt:
# syslog-ng

## Version of syslog-ng

3.12 from freshports  sysutils/syslog-ng312

## Platform

FreeBSD 11.1 p1

# Issue

Trying to build syslog-ng with support for ElasticSearch.  Failure to compile with following configuration options: 
     DOCS=on 
     IPV6=on
     JAVA=on
     JAVA_MOD=on
     JSON=on
     CURL=off
     GEOIP=off 
     GEOIP2=off
     MONGO=off
     PYTHON=off
     REDIS=off
     SMTP=off
     SPOOF=off
     SQL=off
     TCP_WRAPPERS=off

## Failure

Could not resolve all files for configuration ':common:compileClasspath'.
> Could not resolve log4j:log4j:1.2.16.
  Required by:
      project :common
   > Could not resolve log4j:log4j:1.2.16.
      > Could not get resource 'https://repo1.maven.org/maven2/log4j/log4j/1.2.16/log4j-1.2.16.pom'.
         > Could not GET 'https://repo1.maven.org/maven2/log4j/log4j/1.2.16/log4j-1.2.16.pom'.
            > repo1.maven.org
   > Could not resolve log4j:log4j:1.2.16.
      > Could not get resource 'http://maven.elasticsearch.org/releases/log4j/log4j/1.2.16/log4j-1.2.16.pom'.
         > Could not GET 'http://maven.elasticsearch.org/releases/log4j/log4j/1.2.16/log4j-1.2.16.pom'.
            > maven.elasticsearch.org

Am able to resolve both maven.elasticsearch.org and repo1.maven.org.  Can manually download log4j-1.2.16.pom from https://repo1.maven.org/maven2/log4j/log4j/1.2.16/log4j-1.2.16.pom

second URL prompts for user credentials.

Kokan:
In the #2047 the issue was identified, which is not related to syslog-ng: https://bugs.openjdk.java.net/browse/JDK-8196141

PR #2047, opened 2018-05-18
[WIP] Javah removed from JDK 10

MrAnno:
This PR fixes the following incompatibilities:
- `javah` has been removed from JDK 10, `javac -h` should be used instead


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/java/CMakeLists.txt
```diff
@@ -84,42 +66,34 @@ endif()
...
 
-  gen_java_header(
-      ${CMAKE_CURRENT_BINARY_DIR}/org_syslog_ng_LogMessage.h
-      ${CMAKE_CURRENT_BINARY_DIR}/syslog-ng-core.jar
-      ${CMAKE_CURRENT_BINARY_DIR}
-      "org.syslog_ng.LogMessage"
-      "syslog-ng-core"
+  create_javah(GENERATED_FILES BUILT_HEADERS
+        CLASSES    org.syslog_ng.LogMessage org.syslog_ng.InternalMessageSender org.syslog_ng.LogDestination org.syslog_ng.LogTemplate
+        CLASSPATH ${CMAKE_CURRENT_SOURCE_DIR}/src/main/java/
```
MrAnno: This seems to be tricky. Why did you set CLASSPATH to the source dir?

modules/java/CMakeLists.txt
```diff
@@ -84,42 +66,34 @@ endif()
...
   )
 
-  gen_java_header(
-      ${CMAKE_CURRENT_BINARY_DIR}/org_syslog_ng_LogMessage.h
-      ${CMAKE_CURRENT_BINARY_DIR}/syslog-ng-core.jar
-      ${CMAKE_CURRENT_BINARY_DIR}
-      "org.syslog_ng.LogMessage"
-      "syslog-ng-core"
+  create_javah(GENERATED_FILES BUILT_HEADERS
+        CLASSES    org.syslog_ng.LogMessage org.syslog_ng.InternalMessageSender org.syslog_ng.LogDestination org.syslog_ng.LogTemplate
```
MrAnno: `org.syslog_ng.LogPipe` has a native method, so it should be listed here.

modules/java/Makefile.am
```diff
@@ -43,9 +43,18 @@ $(SYSLOG_NG_CORE_JAR): $(JAVA_CLASS_FILES) $(SYSLOG_NG_CORE_LIBS_DIR)
...
 
+if HAVE_JAVAH
+
 $(JAVA_HEADER_FILES): $(JAVA_BUILDDIR)/org_syslog_ng_%.h: $(SYSLOG_NG_CORE_PACKAGE_DIR)/%.class
 	$(AM_V_GEN)$(JAVAH) -classpath  $(SYSLOG_NG_CORE_CLASSES_DIR) -d $(JAVA_BUILDDIR) org.syslog_ng.$*
 
+else
+
+JAVAC_HEADER_OPTIONS = -h $(JAVA_BUILDDIR)
+$(JAVA_HEADER_FILES): $(JAVA_CLASS_FILES)
```
MrAnno: The `JAVA_CLASS_FILES` target will generate our headers as a side effect. I can't express this explicitly in the makefile because a [static pattern](https://www.gnu.org/software/make/manual/html_node/Static-Usage.html#Static-Usage) is used in `JAVA_CLASS_FILES`.

MrAnno:
I can't compile `java-modules` on Ubuntu 18.04 (JDK 10).

```
> Could not resolve log4j:log4j:1.2.17.
  Required by:
      project :common
   > Could not resolve log4j:log4j:1.2.17.
      > Could not get resource 'https://repo1.maven.org/maven2/log4j/log4j/1.2.17/log4j-1.2.17.pom'.
         > Could not GET 'https://repo1.maven.org/maven2/log4j/log4j/1.2.17/log4j-1.2.17.pom'.
            > java.lang.RuntimeException: Unexpected error: java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty
```
Related to #1726.

It seems something is not OK with `/etc/ssl/certs/java/cacerts`. I had to copy my cacerts from Arch Linux to resolve the problem

Hint(?): https://bugs.openjdk.java.net/browse/JDK-8196141

kira-syslogng:
Build FAILURE

modules/java/CMakeLists.txt
Kokan: Long story short, it works if you specify the source instead of classpath, but that's as you pointed out not a good thing to do. Now corrected with a proper `get_property` call.

modules/java/CMakeLists.txt
Kokan: It was not listed in the original list, that could be extended... the new way of header generation (after 1.8) generates the `LogPipe` anyway.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

MrAnno:
I was able to compile `java` and `java-modules` using Java 10 on my machine, so I think I can remove the WIP flag.

There is a remaining CAcert-related issue with Gradle on newer Ubuntu systems, but it can be fixed in another pull request (if it is our fault).

m4/check_java_support.m4
```diff
@@ -4,12 +4,17 @@ AU_ALIAS([AC_CHECK_GRADLE_VERSION], [AX_CHECK_GRADLE_VERSION])
...
     cd $(dirname "$READLINK_TARGET")
-  done
-  echo $(pwd -P)/$(basename "$READLINK_TARGET")
+    while test -L "$READLINK_TARGET"; do
+      READLINK_TARGET=$(readlink "$READLINK_TARGET")
+      cd $(dirname "$READLINK_TARGET")
+    done
+    echo $(pwd -P)/$(basename "$READLINK_TARGET")
+  else
+    echo "$READLINK_TARGET"
```
furiel: test -n means string is nonzero length, and we are now in the else branch, so now READLINK_TARGET is empty. Why are we echo-ing it? What does this patch fix?

modules/java/CMakeLists.txt
```diff
@@ -84,42 +66,40 @@ endif()
...
   )
 
-  gen_java_header(
-      ${CMAKE_CURRENT_BINARY_DIR}/org_syslog_ng_LogTemplate.h
-      ${CMAKE_CURRENT_BINARY_DIR}/syslog-ng-core.jar
-      ${CMAKE_CURRENT_BINARY_DIR}
-      "org.syslog_ng.LogTemplate"
-      "syslog-ng-core"
-  )
+  set(BUILT_HEADERS "${CMAKE_CURRENT_BINARY_DIR}/syslog-ng-core.jar")
```
furiel: It looks strange to call the jar as BUILT_HEADERS, which originally contained the headers. Can you just explicitely call set_source_files_properties on the jar?

Kokan: Hmm, now I kinda do not remember why was it written in this way. But I do not think it is right.

m4/check_java_support.m4
MrAnno: It prevents returning an invalid path when the target is empty.
(`dirname` returns the current working dir when its first parameter is empty)

modules/java/CMakeLists.txt
MrAnno: @Kokan Could you have a look at this again, so we can finish the PR?

m4/check_java_support.m4
furiel: As we discussed irl, the else brach is not necessary. It does not cause problem either, but not necessary.

MrAnno: Removed, thanks.

kira-syslogng:
Build SUCCESS

Kokan:
It would be great if somebody could test it with cmake 3.11 or above. Maybe @MrAnno ?

modules/java/CMakeLists.txt
Kokan: @MrAnno @furiel sorry for the delay

kira-syslogng:
Build SUCCESS

MrAnno:
I'll check it.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
I think the name of the PR is confusing, because this actually fixes the case when `javah` is missing, and that is obsolete since java `1.8`.
While normally I would say we should also fix the above issue in this PR, actually it was present before this patch, and by fixing it `syslog-ng` won't be java 10 compatible either see #1943.

This patch improves java `1.8`, and did not make it worth for other versions.

I think the actual fix of the issue #1943 is not scope of this PR.

@MrAnno if you agree please rename this PR.

lbudai:
Hmmmmmmm.....

There are two bugs around our java build system
1. jdk bug
2. gradle bug...

Finally I was able to build syslog-ng on Ubuntu 18.04.
There is a workaround for the jdk bug with the certificates: 
Pass `-Djavax.net.ssl.trustStorePassword=changeit` option to the jvm that uses by gradle.
Don't replace `changeit`, otherwise it won't work!
I simply set the `GRADLE_OPTS` environment variable. I think we could use this workaround until Ubuntu won't upgrade the jdk-11.

The gradle story is a little bit different... I had to upgrade it to version 4.9 (Ubuntu 18.04 still uses  version 3.4.1, and it is possible that it works well with other versions).
I think we could use the [sdkman](https://sdkman.io/) tool for installing gradle.

One additional note: I changed the elastic maven repo from `http://maven.elasticsearch.org/releases` to `"https://artifacts.elastic.co/maven"`.

Next week I try to setup a PR that makes it possible to build syslog-ng on Ubuntu 18.04.

MrAnno:
Will reopen once it's needed.

PR #482, opened 2015-05-12
completerlang: quick fix for PLY-3.6

lbudai:
Local variable 'state' is not in frame 1 anymore, it is in frame 2 in
PLY-3.6 (there is a callback which calls our function this is the reason why
the state variable is in frame 2).

Note that this is an ugly-ugly hack :-)
We have to eliminate this sitation somehow.

Signed-off-by: Laszlo Budai <Laszlo.Budai@balabit.com>

PR #2050, opened 2018-05-22
requirements.txt: drop explicit version numbers

bazsi:
We are only using requirements.txt to bootstrap the development environment,
and that file is not really used in production. It is better to rely on
the latest and greatest so we actually follow pep8/pylint etc changes as
they happen.

If we get too many issues with that, we might pin some of the version
numbers in te future, but so far the experience has been that we simply
forget to bump versions and stick to very old versions instead.

### DIFF
--- a/dbld/images/required-pip/all.txt
+++ b/dbld/images/required-pip/all.txt
@@ -1,6 +1,6 @@
-nose==1.3.7
-ply==3.10
-pep8==1.7.1
-pylint==1.8.2
-astroid==1.6.1
-logilab-common<=0.63.0
+nose
+ply
+pep8
+pylint
+astroid
+logilab-common


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

dbld/images/required-pip/all.txt
```diff
@@ -1,6 +1,6 @@
-nose==1.3.7
-ply==3.10
-pep8==1.7.1
-pylint==1.8.2
-astroid==1.6.1
-logilab-common<=0.63.0
+nose
+ply
+pycodestyle
```
MrAnno: We should use `pycodestyle` in `pylib/Makefile.am` if we depend on the new name (this is why the macOS build fails).

https://github.com/balabit/syslog-ng/blob/28a34968d572ab6dfecf71d9d3fff1b6ccb9ecae/modules/python/pylib/Makefile.am#L71-L72

lbudai:
@bazsi: just a reminder to #482
If we don't set a fix version for `ply` we can easily get crash... and it's not an easy job to investigate.

I can live with this, but won't forget this fragile code chunk ;)

bazsi:
Unfortunately setting the version here doesn't solve the various
environments our users use, because they will probably not run this in a
venv, but rather use the supplied binaries of their distribution.

-- 
Bazsi

dbld/images/required-pip/all.txt
bazsi: ahh.. I've removed the rest of the branch, one of the patches did autodetected whether to use
pep8 or pycodestyle.

I think we can work with the warning that pep8 generates, so I change this patch to use pep8 instead.

bazsi:
I've reverted the change that migrates to pycodestyle and remains on pep8. This will display a warning at runtime, but we can live with that for some time yet.

Also, it seems that ply indeed added the requested "state" attribute to parser objects, so I could get rid of the getframe() stuff that @lbudai was mentioning. I've sent another PR to remove them.

With that, if this is green, I think it can be merged.

bazsi:
Travis failed with a java build related error, but otherwise this patch seems to be ok and not really related to the build issue. I am merging it.

lbudai:
@bazsi : I understand that there was a java(mvn repo/whatever) related error but it would be better if not merging your own PR when CI reports failure: it's a really bad practice. Even if you are right in that there was no other error just the java which you didn't touch... it's could become an example that is followed by others. And this is why I'm against this kind of way of working (merging your own pr _when_ ci is red).

And nowadays we are trying to be more active in merging/reviewing PRs (even in our private time), so if I may suggest, next time just ping the team, I'm sure that when someone see your request, will react to it, even at midnight :)

bazsi:
Ok, got it.

PR #2055, opened 2018-05-24
Multitransport

lbudai:
## MultiTransport

Responsibility of this class is doing runtime switches between the supported `LogTransports` inside a `LogProto`.
`Multitransport` *is-a* `LogTransport` so it can be passed to any `LogProto`'s constructor.
This also means that this is a polymorphic transport. Polymorphism is implemented via runtime delegation: each `MultiTransport` *has-a* `LogTransport`, called `active_transport` and all the read/write methods are delegate I/O to `active_transport`.

### Motivation

Why would someone need `MultiTransport`? There can be `LogProtos` that support more than one `LogTransport`and switching between the `LogTransports` are done in runtime (negotiation).
Example: a `LogProtoServer` starts to communicate with its peer via a plain TCP LogTransport. Later, a client request SSL, or compression. In this case `LogProtoServer` has to be able to switch between `LogTransports`.

### Proposed solution
* Need a class that responsible for transport switching: this is `Multitransport`.
* Need `TransportFactory` which is responsible for creating `LogTransport` instances. A factory has all the resources that are needed for being able to create a `LogTransport` instance.
So in case of TLS, the `TransportFactoryTLS` contains `TLSContext` and `TLSSessionVerifyFunc` with `verify_data`. Using these factories eliminates the problem of *how to pass a driver-level
option defined resource to a proto.

   * What is this problem exactly?
Suppose we have a network source driver that has a proto which supports negotiate transport in runtime.
Each network source driver support tls options. A `TLSContext` is constructed with these options. But the constructed `TLSContext` is actually used by the proto and the proto is constructed as a plugin.
How can we pass this resource to the proto? We have only one option: via the global constructor.  Without changing the proto ctor interface we have several options, but I think only `Multitransport` is viable. See Alternatives section .
* Need a `TransportFactoryRegistry` which is responsible for storing factory instances and support factory lookup.
* Need a `TransportFactoryId` which is responsible for acting as a uniq-id for `TransportFactory` instances.

* Multitransport is started with a default transport, and provides interface for adding other factories. These factories can be added anywhere, so it contains only the allowed transport switches.

multitransport

### Alternatives
 * polluting `LogProto's` constructor with transport resources, like `TLSContext`, or gpointer
 * using gpointers in `LogTransport` and use it for referencing factory resources (this is also a kind of pollution)
 * implement a facade for all the possible transport switches (`switchToTls`)
   * The problem with this approach is that it's suppose we have only *global* transports and doesn't support *locals*.
What is a global transport? `LogProtos` has-a `LogTransport` and `LogProtos` can be implemented as modules (plugins).
All the transports that are part of the core (located in lib) are known for this facade and can be make part of it.
But what about local transports? We cannot extend easily the facade with locals.
   * Another problem with the facade approach: factory resources. As all the global transports switching are part of the interface, we need to manage
the allowed transport-switches.

### What are missing?
Currently only Plain TCP and TLS transports are supported.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
Hmm... where is Traivs? It's weird... 
I restarted the build in travis-ci.org : https://travis-ci.org/balabit/syslog-ng/builds/383136853

Do not merge until Travis is not green!

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
I'm reopen the PR and hope that Travis will work at next time...

PR #2057, opened 2018-05-24
Multitransport

lbudai:
## MultiTransport

Responsibility of this class is doing runtime switches between the supported `LogTransports` inside a `LogProto`.
`Multitransport` *is-a* `LogTransport` so it can be passed to any `LogProto`'s constructor.
This also means that this is a polymorphic transport. Polymorphism is implemented via runtime delegation: each `MultiTransport` *has-a* `LogTransport`, called `active_transport` and all the read/write methods are delegate I/O to `active_transport`.

### Motivation

Why would someone need `MultiTransport`? There can be `LogProtos` that support more than one `LogTransport`and switching between the `LogTransports` are done in runtime (negotiation).
Example: a `LogProtoServer` starts to communicate with its peer via a plain TCP LogTransport. Later, a client request SSL, or compression. In this case `LogProtoServer` has to be able to switch between `LogTransports`.

### Proposed solution
* Need a class that responsible for transport switching: this is `Multitransport`.
* Need `TransportFactory` which is responsible for creating `LogTransport` instances. A factory has all the resources that are needed for being able to create a `LogTransport` instance.
So in case of TLS, the `TransportFactoryTLS` contains `TLSContext` and `TLSSessionVerifyFunc` with `verify_data`. Using these factories eliminates the problem of *how to pass a driver-level
option defined resource to a proto.

   * What is this problem exactly?
Suppose we have a network source driver that has a proto which supports negotiate transport in runtime.
Each network source driver support tls options. A `TLSContext` is constructed with these options. But the constructed `TLSContext` is actually used by the proto and the proto is constructed as a plugin.
How can we pass this resource to the proto? We have only one option: via the global constructor.  Without changing the proto ctor interface we have several options, but I think only `Multitransport` is viable. See Alternatives section .
* Need a `TransportFactoryRegistry` which is responsible for storing factory instances and support factory lookup.
* Need a `TransportFactoryId` which is responsible for acting as a uniq-id for `TransportFactory` instances.

* Multitransport is started with a default transport, and provides interface for adding other factories. These factories can be added anywhere, so it contains only the allowed transport switches.

multitransport

### Alternatives
 * polluting `LogProto's` constructor with transport resources, like `TLSContext`, or gpointer
 * using gpointers in `LogTransport` and use it for referencing factory resources (this is also a kind of pollution)
 * implement a facade for all the possible transport switches (`switchToTls`)
   * The problem with this approach is that it's suppose we have only *global* transports and doesn't support *locals*.
What is a global transport? `LogProtos` has-a `LogTransport` and `LogProtos` can be implemented as modules (plugins).
All the transports that are part of the core (located in lib) are known for this facade and can be make part of it.
But what about local transports? We cannot extend easily the facade with locals.
   * Another problem with the facade approach: factory resources. As all the global transports switching are part of the interface, we need to manage
the allowed transport-switches.

### What are missing?
Currently only Plain TCP and TLS transports are supported.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lib/transport/tests/test_transport_factory.c
```diff
@@ -91,8 +91,8 @@ _fake_transport_factory_new(void)
 {
   FakeTransportFactory *instance = g_new0(FakeTransportFactory, 1);
   instance->super.id = _fake_transport_factory_id();
-  instance->super.construct = _transport_factory_construct;
-  instance->super.destroy = _transport_factory_destroy; //TODO: confusing the naming: construct_transport... destroy_self
+  instance->super.construct_transport = _transport_factory_construct;
+  instance->super.free_fn = _transport_factory_free;
```
MrAnno: I would squash this commit into `transport: add TransportFactory interface`.

lib/transport/transport-factory-id.h
```diff
@@ -34,14 +34,12 @@ void transport_factory_id_global_deinit(void);
...
 const gchar *transport_factory_id_to_string(const TransportFactoryId *);
+guint transport_factory_id_hash(gconstpointer);
+gboolean transport_factory_id_equal(const TransportFactoryId *, const TransportFactoryId *);
+TransportFactoryId *transport_factory_id_clone(const TransportFactoryId *);
 
 #define TRANSPORT_FACTORY_ID_NEW(name) ({GString *str = g_string_new(""); g_string_printf(str, "%s-%s:%d", name, __FILE__, __LINE__); str;})
-#define TRANSPORT_FACTORY_ID_FREE_FUNC transport_factory_id_free
-#define TRANSPORT_FACTORY_ID_HASH_FUNC g_string_hash
-#define TRANSPORT_FACTORY_ID_CMP_FUNC g_string_equal
-#define TRANSPORT_FACTORY_ID_CLONE transport_factory_id_clone
```
MrAnno: I'd also squash this commit into `transport: add transport-factory-id` and use the new non-macro interface in `transport: add TransportFactoryRegistry`.

lib/transport/transport-factory-socket.c
```diff
@@ -27,7 +27,7 @@
 
 typedef struct _TransportFactorySocket TransportFactorySocket;
 
-DEFINE_TRANSPORT_FACTORY_ID_FUN(transport_factory_socket_id);
+DEFINE_TRANSPORT_FACTORY_ID_FUN(socket, transport_factory_socket_id);
```
MrAnno: I think it would be more understandable if we passed the name of the transport as a string.

For example:
```C
DEFINE_TRANSPORT_FACTORY_ID_FUN("socket", transport_factory_socket_id);
```

lib/transport/multitransport.c
```diff
@@ -74,10 +74,9 @@ _construct_transport(const TransportFactory *factory, gint fd)
 
 gboolean multitransport_switch(MultiTransport *self, const TransportFactoryId *factory_id)
 {
-  const TransportFactoryId *old = transport_factory_get_id(self->active_transport_factory);
   msg_debug("Transport switch requested",
-            evt_tag_str("Old TransportFactoryId", transport_factory_id_to_string(old)),
-            evt_tag_str("Requested TransportFactoryId", transport_factory_id_to_string(factory_id)));
+            evt_tag_str("active-transport", self->active_transport->name),
+            evt_tag_str("requested-transport", transport_factory_id_get_transport_name(factory_id)));
```
MrAnno: Please use underscores in tag names.

lib/transport/multitransport.c
```diff
@@ -89,8 +88,8 @@ gboolean multitransport_switch(MultiTransport *self, const TransportFactoryId *f
 
   _do_transport_switch(self, transport, transport_factory);
 
-  msg_debug("multitransport_switch succeded",
-            evt_tag_str("New TransportFactoryId", transport_factory_id_to_string(factory_id)));
+  msg_debug("Transport switch succeded",
+            evt_tag_str("new-active-transport", self->active_transport->name));
```
MrAnno: Please use underscores in tag names.

lib/transport/transport-factory-socket.c
MrAnno: Sorry, already fixed (`transport: use a sequence as uniq id instead of location macros `)

kira-syslogng:
Build SUCCESS

PR #2060, opened 2018-05-28
multitransport: older glib support

MrAnno:
https://developer.gnome.org/glib/2.32/glib-Hash-Tables.html#g-hash-table-insert

### DIFF
--- a/lib/transport/transport-factory-registry.c
+++ b/lib/transport/transport-factory-registry.c
@@ -56,11 +56,11 @@ void transport_factory_registry_free(TransportFactoryRegistry *self)
   g_free(self);
 }
 
-gboolean
+void
 transport_factory_registry_add(TransportFactoryRegistry *self, TransportFactory *factory)
 {
   const TransportFactoryId *id = transport_factory_get_id(factory);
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

lib/transport/transport-factory-registry.c
```diff
@@ -50,17 +50,18 @@ TransportFactoryRegistry *transport_factory_registry_new(void)
...
+transport_factory_registry_free(TransportFactoryRegistry *self)
 {
   g_hash_table_unref(self->registry);
   g_free(self);
 }
 
-gboolean
+void
 transport_factory_registry_add(TransportFactoryRegistry *self, TransportFactory *factory)
 {
```
lbudai: and what if we’d create an insert method in compat that reports back result of insertion (lookup&insert?)
I didn’t cover the false case with unit tests, maybe we have to.

MrAnno: Yeah, I would also check the return value of `transport_factory_registry_add()` in `multitransport_new()` and `multitransport_add_factory()`.

I'll add the compat function and fix all invocation in a separate PR.

PR #2061, opened 2018-05-28
transport-factory-registry: overwriting of a factory instance is not …

lbudai:
…allowed

Assert when someone wants to register a difference TransportFactory
instance with the same TransportFactoryId.


kira-syslogng:
Build SUCCESS

PR #2063, opened 2018-05-28
LogThreadedDestDriver batching

bazsi:
This is still a work in progress, however the basic function did work and I wanted to solicit feedback on it.

So this basically adds a batched API to LogThreadedDestDriver and enables it in the HTTP destination,
which would send batches of 10 messages in a single HTTP transaction.

The bulk of the patch is a largish cleanup of LogThreadedDestDriver with lots of non-functional changes, renaming functions, moving them around, making stuff easier to navigate there.

The infrastructure to support batching is pretty basic: you can now return WORKER_INSERT_RESULT_QUEUED to add the current message into a batch, whose size is tracked by LogThreadedDestDriver. Any time the driver returns SUCCESS, the entire batch is considered to be sent. If the driver returns an error, the entire batch is retried and dropped after 3 retries. I have also started to add an explicit acknowledgement model, but that's the most foggy at this point.

NOTE: this is not final yet, some of the stuff is not configurable and a number of corner cases not yet thought through. Not to mention that I did the final changes, rebased and writing this note after roughly being awake for ~24 hours.


kira-syslogng:
Build FAILURE

furiel:
The problem with the proposed http bulk sending (concatenating messages into a single http request) is that the remote needs a method to extract the messages from the single request. Which means each remote/product - if supports bulk sending at all - will specify some kind of structure for the http request (list of structured json, split by new lines). That would mean a single http destination driver needs to support a variety of product specific protocols, which is difficult to achieve.

The better way is that http destination would create multiple parallel connections to the same remote. In this case each http request would contain a single message. But these messages would be sent parallelly on the different connections. This would result in a speedup. This method would work even if the remote does not support the abovementioned bulk send. This version is very similar to the way http2 works, we could probably introduce support for http2 relatively easily with this version.

I know I am complaining about the least important part of this patchset, I am just afraid with this part merged, it would be difficult to introduce this other http bulk method without breaking the configuration.

lbudai:
I agree with furiel and I also think that HttpDestDriver could be a base class for other destination drivers(like REST basef Elasticsearch destination driver). 
What if `append_body` would be a virtual method? 

I think we can add it later. The question is that do we agree in this(http dst driver is a base class for other drivers and append body is virtual)?

bazsi:
I actually want to implement both of:
* Being able to add multiple messages into the payload and
* Running multiple workers at the same time, each serializing a subset of
the incoming messages.

We can already provide enough framing with our templating mechanism, and my
guess is that one json per line is a pretty common solution for receivers.
IIRC splunk hec does that.

bazsi:
To @furiel's comment above:

I have just checked:
* Splunk HEC: accepts batching, each message is a separate JSON object with no separator between them (e.g. just JSON objects following one another), source: https://github.com/splunk/splunk-javascript-logging/blob/master/splunklogger.js

* elastic search uses newline delimited JSON objects: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html

Both formats can be produced by a template string and simply appending the results. Regardless, the http destination driver can be extended with more sophisticated logic, should the ones above not be sufficient, but I guess those would be new drivers and not ones based off http() directly.

With regards to multiple destinations, I envision something like the following:

```

destination foo { http(url("http://localhost:9200/") batch-size(1000) workers(10) worker-routing-by("$HOST$PROGRAM")); };

```

Where
 * batch-size specifies the amount of messages in a single request
 * workers specifies the number of worker threads that send messages
 * worker-routing-by specifies the key upon which messages are distributed to workers, e.g. this value is produced for all incoming messages, a hash is calculated and that would drive the decision which queue a specific message gets into.

The exact option names can still be refined, I am not 100% satisfied with them either, but I think you get the idea.

I think the format above makes sense and it is possible to combine these ideas into one concept. What do you think?

pzoleex:

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
This had a compile error, but anyways it is not yet ready. Just wanted to share in time.

I've just pushed a commit that should make it compile, but still proper error handling on the flush() side is missing.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

furiel:
@bazsi how should bulk sending work together with throttling? Now we check throttle before _perform_inserts, which is the original behaviour, throttle applied globally per message.

In my opinion, in case of bulk sending (single worker but many messages in same http request), one bulk should be counted as a single send. When we have multiple workers, then throttle should be applied per worker.

bazsi:
Yeah, i would throttle per worker. Weather to throttle per batch or per message I don't know yet, I think both makes sense but I would do whatever is easier, And i would probably lean towards the message based throttling as other destinations do that. But i don't yet have a firm answer.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/logthrdestdrv.c
```diff
@@ -201,6 +201,73 @@ _rewind_batch(LogThreadedDestDriver *self)
...
+      _rewind_batch(self);
+      _disconnect_and_suspend(self);
+      break;
+
+    case WORKER_INSERT_RESULT_EXPLICIT_ACK_MGMT:
+      /* we require the instance to use explicit calls to ack_messages/rewind_messages */
+      break;
+
+    case WORKER_INSERT_RESULT_SUCCESS:
+      stats_counter_inc(self->written_messages);
```
furiel: When talking about statistics, should we increase any per-message counters with batch size? It might worth to cover this with some tests. e2e is great, but probably this could be even covered with a unit test.

lib/logthrdestdrv.c
```diff
@@ -220,9 +220,6 @@ _process_result(LogThreadedDestDriver *self, gint result, LogMessage *triggering
 
       if (self->retries.counter >= self->retries.max)
         {
-          if (self->messages.retry_over && triggering_message)
-            self->messages.retry_over(self, triggering_message);
```
furiel: Is there any other way to know which message is dropped?

modules/python/python-dest.c
```diff
@@ -177,14 +169,6 @@ _py_invoke_is_opened(PythonDestDriver *self)
   return _dd_py_invoke_bool_function(self, self->py.is_opened, NULL);
 }
 
-static void
-_py_invoke_retry_error(PythonDestDriver *self, PyObject *dict)
-{
-  if (!self->py.retry_error)
-    return;
-  _dd_py_invoke_void_function(self, self->py.retry_error, dict);
```
furiel: As I see this is the only way for python binding to know syslog-ng gave up on a message. Maybe this can be used if user wants to implement a network like destination, and when they see the message is dropped due to server error (after 3 tries), they want to fallback to another server. Or just want to do their own logging on the failed message, like report an incident.

lib/logthrdestdrv.c
```diff
@@ -34,7 +34,7 @@ log_threaded_dest_driver_set_max_retries(LogDriver *s, gint max_retries)
 {
   LogThreadedDestDriver *self = (LogThreadedDestDriver *)s;
 
-  self->retries.max = max_retries;
+  self->retries_max = max_retries;
```
furiel: Personally I like the coupling of max and counter fields. Especially if the `retries_over` were part of the same structure.

modules/http/http.h
```diff
@@ -30,10 +30,13 @@
 
 #include "logthrdestdrv.h"
 
+#define CURL_NO_OLDIES 1
```
furiel: Just a note to other reviewers how I tried to validate this: the problem was introduced in libcurl with 9adf3c473a01b289c781aab111f9ad2fc541ed4e (libcurl 7.50), which was fixed in d6604524ad24daf4581efbe0020da058d2b3af84 (libcurl 7.51) by introducing the underlying define.

So as I understand, this has effect only in very recent libcurl  >= 7.51 (for comparison, xenial ships with 7.47), and has no effect below <7.50. In summary, if this compiles with xenial or trusty, then probably will on all supported platforms.

lib/logthrdestdrv.c
```diff
@@ -160,148 +287,150 @@ log_threaded_dest_driver_do_insert(LogThrDestDriver *self)
...
-        }
+      self->batch_size++;
+      _process_result(self, result);
 
+      log_msg_unref(msg);
       msg_set_context(NULL);
       log_msg_refcache_stop();
     }
-  if (!self->suspended)
+  if (!self->suspended && self->batch_size > 0 && self->worker.flush)
```
furiel: This is the right direction, but we will need to be careful in the documentation that flush_lines and flush_bytes are only maximal values, and syslog-ng will flush the first time she can, even if these values are not hit.

This phenomenon might worth reflecting in the configuration: instead of flush_lines and flush_bytes -> batch_mode(on/off), max_lines, max_bytes.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/logthrdestdrv.c
bazsi: good catch. just pushed a couple of patches to at least fix the issues (both the statistics and also a related issue with $SEQNUM).

A dummy threaded destination would be very useful for testing, which is probably the thing I would do next, before actually making it possible to use multiple workers.

These are the fixes (I've kept them separate patches on top of my branch):
67785c68f776190c385aa8a992d56b8ce89769e8 logthrdestdrv: clean up sequence number increments
20242792dfdf67eba14c178aa4fc71528084ab26 logthrdestdrv: written_messages should get updated with the entire batch

lib/logthrdestdrv.c
bazsi: I would prefer to add this mechanism when the need arises. I am not sure this interface is that good at solving this issue. I mean the insert() function already knows that  a message would be dropped, because it is the one that decides that there was an error. The retry mechanism would probably not be needed if someone has an alternative location.

I was actually looking at what used this interface and there were two locations:
 1) the python binding: forwarded to the python code
 2) mongodb: to log the details of the message.

If we wanted to retain this function, then probably this would have to become batch based, as we are retrying batches and not just messages. 

Anyway, back to my initial point: I would wait for an actual use-case to surface.

Is this ok?

modules/python/python-dest.c
bazsi: Don't forget that we now have flush() as an interface (not yet exposed to Python, but I guess that's one of the next steps). It is in flush() where the plugin decides if a batch can be sent or not. It can even do its retry tracking logic (e.g. it can count the number of WORKER_INSERT_RESULT_ERROR return values in a row).

Logging can also happen in flush() or even during insert().

btw: I am not against the interface per-se, it was just difficult to retain in its current form because of batching. I've checked there were no important in-tree users, so I've dropped it. Reimplementing it would have increased the scope yet again and at 43 patches, this series is already huge.

lib/logthrdestdrv.c
bazsi: I also like it, but only when we are not using it for prefixing and are actually grouping related fields. The worst offender was format.stats_instance, with a single member, but we also had messages.retry_over, again as a single member.

I might have gone the other side too much by also getting rid off retries, but I did that because the original structure was completely mixed and has been confusing. I did keep "worker" though, I even added a few members.

As I look at the structure, I would still probably  not use "retries" as a struct name, rather "batch" or "batching", which could contain:
* size
* retry_max
* retry_attempts

Again to cluster not just instead of prefixing, but along functionality. Do you want me to do that?

bazsi:
@furiel 

> @bazsi just an idea: as this patch looks like a critical fix: 63ad460, it might worth separate from the rest to be sure it is merged into the next ose release (and maybe move some easily reviewable low-risk refactor commits from this PR with it, so that the size of this pr is reduced).

There was no actual bug that I've found, so I didn't expedite that one fix.

bazsi:
@kira-syslogng test this please test branch=logthreadeddst-multi-threading; test=functions/destination-drivers/http-destination-driver/;

kira-syslogng:
Build FAILURE

bazsi:
@kira-syslogng test this please test branch=logthreadeddst-multi-threading; test=functions/destination-drivers/http-destination-driver/;

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
@kira-syslogng test this please test branch=logthreadeddst-multi-threading; test=functions/destination-drivers/;

kira-syslogng:
Build FAILURE

modules/python/python-dest.c
furiel: Ok sounds good to me. I do not have any other use case in mind.

modules/http/http.c
```diff
@@ -344,6 +344,11 @@ _setup_static_options_in_curl(HTTPDestinationDriver *self)
   curl_easy_setopt(self->curl, CURLOPT_DEBUGFUNCTION, _curl_debug_function);
   curl_easy_setopt(self->curl, CURLOPT_VERBOSE, (long) (debug_flag || trace_flag));
 
+  curl_easy_setopt(self->curl, CURLOPT_FOLLOWLOCATION, 1);
+  curl_easy_setopt(self->curl, CURLOPT_POSTREDIR, CURL_REDIR_POST_ALL);
+  curl_easy_setopt(self->curl, CURLOPT_REDIR_PROTOCOLS, CURLPROTO_HTTP | CURLPROTO_HTTPS);
+  curl_easy_setopt(self->curl, CURLOPT_MAXREDIRS, 3);
```
furiel: Exciting new feature! It might worth and e2e test, and a note in the documentation.

modules/http/http.c
```diff
@@ -390,6 +390,7 @@ _format_request_headers(HTTPDestinationDriver *self, LogMessage *msg)
   struct curl_slist *headers = NULL;
   GList *l;
 
+  headers = _add_header(headers, "Expect", "");
```
furiel: How does this work? Does Libcurl detect internally that expect header was already set? Or two expect headers will be sent in the header, and this one overrides the other one?

bazsi:
@kira-syslogng test this please test branch=logthreadeddst-multi-threading; test=functions/destination-drivers/java-destination-driver;

kira-syslogng:
Build FAILURE

lib/logthrdestdrv.c
```diff
@@ -32,13 +32,13 @@
 void
 log_threaded_dest_driver_set_max_retries(LogDriver *s, gint max_retries)
 {
-  LogThrDestDriver *self = (LogThrDestDriver *)s;
+  LogThreadedDestDriver *self = (LogThreadedDestDriver *)s;
```
furiel: The "Threaded" name in `LogThreadedDestDriver` is a little unfortunate. I had a preconception due to this for quite some time, thinking LogThreadedDestDriver is a destination driver that works with multiple threads, hence the name. Then it turned out it is still single thread, just the thread is not the main loop. Here the name carries some implementation detail, instead of the purpose of the concept. As it is natural expectation that a worker is a thread, we could just drop Threaded, instead of making it more explicit. We could call DestinationDriver (LogDestDriver is already taken)?

bazsi:
The default more of operation of syslog-ng has been asynchronous, thus the
threaded qualifier makes sense, I think.

And LogDestDriver vs DestinationDriver is not sufficiently different in my
opinion.

But anyhow my next plan is to make the threaded infrastructure support
multiple threads, in which case this should be moot.

bazsi:
I am rebasing this one on top of #2123 , so this should not be merged until that is in.

kira-syslogng:
Build FAILURE

bazsi:
I am repurposing this branch to be the batching support in the original logthreaddstdrv module. I've pushing the http related changes to a separate PR and I am working on introducing unit tests into this one.

kira-syslogng:
Build FAILURE

lib/logthrdestdrv.h
```diff
@@ -59,6 +59,11 @@ typedef struct _LogThreadedDestWorker
 struct _LogThreadedDestDriver
 {
   LogDestDriver super;
+  struct iv_task  do_work;
+  struct iv_event wake_up_event;
```
furiel: Just for curiousity: can you explain the guiding principle here?

I was thinking we are trying to move the frequent fields to the top for cpu-cache optimization?
I see do_work used a lot, but wake_up_event less, shutdown_event very rarely, while batch, processed_message, memory_usage, seq_num much more, but they are lower in the structure.

lib/logthrdestdrv.c
```diff
@@ -255,10 +245,11 @@ _process_result(LogThreadedDestDriver *self, gint result)
 
     case WORKER_INSERT_RESULT_SUCCESS:
       _accept_batch(self);
+      step_sequence_number(&self->seq_num);
```
furiel: Theoretical question: do we expect the sequence number to be continuous without skips? Should we step down the sequence number in case of drop?

furiel:
@kira-syslogng retest this please test branch=logthrdestdrv-batching;

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/logthrdestdrv.c
```diff
@@ -187,74 +270,27 @@ _perform_inserts(LogThreadedDestDriver *self)
...
-          break;
-        }
+      _process_result(self, result);
 
+      log_msg_unref(msg);
       msg_set_context(NULL);
       log_msg_refcache_stop();
     }
-  if (!self->suspended)
+  if (!self->suspended && self->batch_size > 0 && self->worker.flush)
```
furiel: Until the send in language bindings return with worker_insert_result_t, we cannot add batch_size > 0, as there is no way to do batching after the change, and we break the only way for batching prior the patchset: send returns true, then on-empty is called unconditionally.

lib/logthrdestdrv.h
bazsi: I wanted do_work up there and the rest is not that important, so I've kept the grouping of ivykis events/tasks. 

This is probably just noise, so I will skip this patch.

lib/logthrdestdrv.c
bazsi: No we shouldn't step down the sequence number. The behaviour of SEQNUM is pretty much syslog specific, so might not even matter that much.

The point of SEQNUM is to be able to detect drops and order log messages. I would try to avoid duplicate SEQNUMs at all costs and it is still better to produce gaps between them.

Quoting https://www.cisco.com/c/en/us/td/docs/routers/access/wireless/software/guide/SysMsgLogging.html#wp1054751

> Because there is a chance that more than one log message can have the same timestamp, 
> you can display messages with sequence numbers so that you can unambiguously refer to a > single message. By default, sequence numbers in log messages are not displayed. 

The RFC5424 [meta sequenceId] is structured data is coming from the same roots.

With that said, I simply left SEQNUM always increasing as we do in all other drivers. (e.g. logwriter based ones, and sql).

While researching the subject, I've noticed that logwriter based destinations would use a separate sequence number for local messages and for relayed messages (reproducing the original one the relay has received). Honestly I am a bit confused, which is sad, given that I have written that code, roughly 10 years ago.

What I am sure about is that I don't want to fix SEQNUM logic in this PR. :)

In SQL (which I would like to migrate to this framework in a followup PR), the sequence number if just monotonically increasing, even if the transaction fails is being retried, etc.

So I would just reproduce that logic here and then file an issue to fix this mess up.

bazsi: #2152 is the issue number for the SEQNUM mess.

lib/logthrdestdrv.c
bazsi: good catch. I would remove the batch_size condition from here, that does not affect my flush logic that much, it just has to honour the batch_size properly (e.g. do nothing if batch_size is zero). It is not a high price to pay for compatibility.

Once all bindings are migrated, we can put this condition back here.

bazsi:
I am pushing an update to this branch, which now also contains a unit test program and a couple of fixes.

The followup fixes are in separate patches, but you can always combine against the last version by comparing against 511c49dd24e8c7268834b60ededa880919b2be81

I've also rebased against current master. I am pushing stuff in a minute.

bazsi:
+1 note: the unit tests do not cover yet batched behavior completely. But I wanted to push it so, noone reviews stale patches.

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=logthrdestdrv-batching;

kira-syslogng:
Build FAILURE

furiel:
@bazsi something is broken with the stats counter names, that's why most of the tests fail. In the counter names we can see #anon-[...] instead of the actual name.

The strange thing is not only the threaded destination drivers are affected, but seemingly any driver:
```
@version: 3.16
options { stats-level(3); };
source s_stdin { stdin(); };
log { source(s_stdin); };
```
shows
```
src.stdin;#anon-source1#0;-;a;processed;0
```
instead of
```
source;s_stdin;;a;processed;0
```

bazsi:
Thanks. I have an idea what I have broken. I am still working on the unit
test, so putting back the WIP marker.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=logthrdestdrv-batching;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=logthrdestdrv-batching;

kira-syslogng:
Build FAILURE

furiel:
@bazsi this last test: 013-java-destination-group-remove-add.zts is known to be failing, @mitzkia is investigating the problem. Other than that, all tests pass.

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=logthrdestdrv-batching;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=logthrdestdrv-batching;

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:
thanks a lot for helping out with @kira-syslogng

furiel:
@bazsi all looks good to me. One final request: can you add the new unit test to the cmake? It is only part of autotools now.

bazsi:
Sure, I forgot it. Good catch.

bazsi:
I have now added the cmake piece, fingers crossed that it passes in travis.

bazsi:
@kira-syslogng retest this please test branch=logthrdestdrv-batching;

kira-syslogng:
Build SUCCESS

mitzkia:
Internal installer was generated, and tests were Passed.

lbudai:
@mitzkia : thanks!

bazsi:
@lbudai first of all thanks for reviewing and merging the patchset. 

With regards to extra patches, I included them because of the unit test program. Can you be more specific what you consider extra in the series? Thanks

PR #2064, opened 2018-05-29
afsocket: add default_inet_port to protocol plugins

MrAnno:


kira-syslogng:
Build FAILURE

MrAnno:

lib/logproto/logproto-client.h
```diff
@@ -196,6 +196,7 @@ typedef struct _LogProtoClientFactory LogProtoClientFactory;
 struct _LogProtoClientFactory
 {
   LogProtoClient *(*construct)(LogTransport *transport, const LogProtoClientOptions *options);
+  gint default_inet_port;
```
gaborznagy: Is `DEFINE_LOG_PROTO_CLIENT` and `DEFINE_LOG_PROTO_SERVER` affected?

kira-syslogng:
Build SUCCESS

lib/logproto/logproto-client.h
gaborznagy: We've discussed this IRL. :+1:

MrAnno: I leave a note here, because I'm not really proud of this patch.

Currently, our protocols (implementations of LogProtoServer and LogProtoClient) are general. This means, for example that `LogProtoTextServer` is used both for file-based and inet-based drivers.

(If you want to use `LogProtoTextServer` on a file, you should call its `restart_with_state()` method and it will track file positions and do other file-related operations. I think it can be seen clearly that the responsibility of `LogProtoTextServer`(`LogProtoBufferedServer`) is not singular).

Since we don't have a `LogProtoInetServer` abstract class or anything similar, I could not find a better place for the `default_inet_port` field. In case of a new protocol plugin, its default port needs to be defined by the protocol itself, but it also has to be accessible by afsocket's transport mapper (where `bind()` is called).

TLDR: I haven't changed the signature of `DEFINE_LOG_PROTO_CLIENT` and `DEFINE_LOG_PROTO_SERVER`, because I know that this field does not belong here and I don't want to build anything on top of this field.

PR #1967, opened 2018-04-05
nvtable: deserialized handle access crash

Kokan:
This crash happens, when a message is deserialized. (An example for this is the disk queue file.)

The deserialization process has two steps. One that actually make the deserialization, and an additional *fixup* step.

The additional *fixup* step tries to validate and updated the index table with the new handles (if they are new).

The issue comes from this fixup, because first it creates a temporary index table, if a handle has a new handle it saves the updated entry to the temporary table.
After going through all of the entries, if there was even one updated entry it just swaps the two table (temporary, and active).

The issue here, that when there were some handle which has been updated, and there were some that have not, the temporary table has gaps with uninitialized entities.

This probably Fixes #1919.

Issue #1919, opened 2018-03-08
Segfault in nv_registry_get_handle_flags


akunszt:
# syslog-ng

## Version of syslog-ng

```
syslog-ng 3.8.1
Installer-Version: 3.8.1
Revision: 3.8.1-10
Module-Directory: /usr/lib/syslog-ng/3.8
Module-Path: /usr/lib/syslog-ng/3.8
Available-Modules: basicfuncs,disk-buffer,cryptofuncs,afuser,linux-kmsg-format,cef,afprog,kvformat,date,sdjournal,dbparser,pseudofile,confgen,syslogformat,affile,afsocket,csvparser,system-source
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: off
```

## Platform

A docker image created from debian:stretch with the following Dockerfile.

```
FROM debian:stretch

RUN apt-get update && \
        apt-get -y --no-install-recommends upgrade && \
        apt-get -y --no-install-recommends install syslog-ng-core syslog-ng-mod-journal && \
        apt-get clean && \
        rm -rf /var/lib/apt/lists/*

ENTRYPOINT [ "/usr/sbin/syslog-ng", "-F" ]
```

## Debug bundle

There isn't syslog-ng-debun in this version of syslog-ng, yet.

# Issue

## Failure

```
(gdb) bt
#0  nv_registry_get_handle_flags (self=0x564a4cee51b0, handle=1290888120) at ../../lib/logmsg/nvtable.h:91
#1  log_msg_append_format_sdata (self=self@entry=0x564a4cf16010, result=result@entry=0x564a4cf106c0, seq_num=seq_num@entry=0) at ../../lib/logmsg/logmsg.c:858
#2  0x00007fdb70cb06e9 in log_writer_format_log (self=self@entry=0x564a4cf15830, lm=lm@entry=0x564a4cf16010, result=0x564a4cf106c0) at ../../lib/logwriter.c:923
#3  0x00007fdb70cb1346 in log_writer_write_message (self=self@entry=0x564a4cf15830, msg=0x564a4cf16010, path_options=path_options@entry=0x7fdb6c1f2970, write_error=write_error@entry=0x7fdb6c1f296c) at ../../lib/logwriter.c:1101
[...]

PR #2067, opened 2018-05-29
nvtable: deserialize sdata invalid if fixed up

Kokan:
The #1919 was not fixed by the #1967 because the sdata should has also a fixup, that has the same fault.
This PR tries to cover also sdata.

### DIFF
--- a/lib/logmsg/logmsg-serialize-fixup.c
+++ b/lib/logmsg/logmsg-serialize-fixup.c
@@ -237,12 +237,10 @@ _fixup_entry(NVHandle old_handle, NVEntry *entry, NVIndexEntry *index_entry, gpo
   if (index_entry)
     _fixup_handle_in_index_entry(state, index_entry, new_handle);
 
-  if (new_handle != old_handle)
-    {
-      if (log_msg_is_handle_sdata(new_handle))
-        _fixup_sdata_handle(state, old_handle, new_handle);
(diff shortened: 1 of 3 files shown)


Kokan:
@bazsi When I looked up the original PR I was thinking about why did not I added a test for that case, because I remember to have some half baked UT. But that required I think some kind of nasty operation.
I'll check it and come back hopefully with an UT.

furiel:
The patch looks good to me. I would be happy with a supporting unit test as well. Though I would like to see this patch in the next ose release, so it is fine for me if the unit test is added later.

gaborznagy:
I have tested it with the reproduction steps  described in https://github.com/balabit/syslog-ng/issues/1919#issuecomment-374273034 and it works!

Kokan:
I've checked about unit test. The issue came up because the `log_msg_fixup_handles_after_deserialization` function has array on the stack that could have memory garbage. Due to this the reproduction depends on that there is memory garbage.

I would not go in that direction and leave it as it is.

As a side effect I've converted tests to criterion.

kira-syslogng:
Build SUCCESS

bazsi:
Agreed. Thanks for checking it.

lbudai:
Based on today's OSE meeting and on Bazsi's feedback I'm merging the PR.

PR #2069, opened 2018-05-30
cmake: add enable extra warnings option to cmake

Kokan:
The autotools build system has this warnings, so should the cmake.
The 3rd append with the big list of switches should be turned into something more readable.


kira-syslogng:
Build SUCCESS

PR #2072, opened 2018-05-30
cmake: support for ENABLE_EXTRA_WARNINGS

furiel:
Opposed to autotools version, this option is enabled by default. One can disable extra warnings with -DENABLE_EXTRA_WARNINGS=no


CMakeLists.txt
```diff
@@ -20,7 +20,36 @@ set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/Modules ${PROJECT_SOURCE_DIR}/
 set(CMAKE_POSITION_INDEPENDENT_CODE 1)
 set(BISON_FLAGS "-Wno-other")
 
-add_compile_options("-Wno-initializer-overrides")
+set(IMPORTANT_WARNINGS
```
Kokan: I would move this block after detection, because for detection those are not relevant and could cause issue.

modules/afsocket/transport-unix-socket.c
```diff
@@ -34,6 +34,8 @@
 #include <errno.h>
 #include <unistd.h>
 
+#pragma GCC diagnostic ignored "-Wunused-function"
```
Kokan: I would really not like to follow the path of pragmas.

kira-syslogng:
Build SUCCESS

modules/afsocket/transport-unix-socket.c
furiel: I could resolve this with adding G_GNUC_UNUSED to the conflicting function definitions.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2076, opened 2018-05-31
KW_FILE token in a global namespace

Kokan:
The following grammar does not parsed successfully, the syslog-ng reports that the `file` is an unexpected symbol (which is clearly not).

```
parser test_parser {
channel {
if (message('foo')) {
filter {message('foo2')};
parser { date-parser();};
destination { file("/tmp/out2" ); };
};
};
```
Thanks @mitzkia to come up with this configuration snippet.

The strange thing is that it works with every other destination (network, redis, ...) as far as I could see only the file was effected.

As it turned out the KW_FILE token was in a global namespace, every parser has a KW_FILE in it. In this situation the tokenazier must have found the KW_FILE (which is should not found, but because the lack of this token it should try to load the parser for the affile plugin.), but because it found this token, it went on parsing the grammar, and of course it cannot, so it roll back and it cannot do anything other than report unexpected `file`.

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -180,7 +180,6 @@ extern struct _StatsOptions *last_stats_options;
 
 /* source & destination items */
 %token KW_INTERNAL                    10020
-%token KW_FILE                        10021
 %token KW_SYSLOG                      10060
 
 /* option items */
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build FAILURE

Kokan:

kira-syslogng:
Build SUCCESS

PR #2087, opened 2018-06-04
Multitransport contains

lbudai:


kira-syslogng:
Build SUCCESS

lib/transport/transport-factory-id.h
```diff
@@ -40,15 +41,20 @@ TransportFactoryId *transport_factory_id_clone(const TransportFactoryId *);
...
 const gchar *transport_factory_id_get_transport_name(const TransportFactoryId *);
 
+void _transport_factory_id_lock(void);
+void _transport_factory_id_unlock(void);
+
 #define DEFINE_TRANSPORT_FACTORY_ID_FUN(transport_name, func_name) \
   const TransportFactoryId* func_name(void) \
   {\
     static TransportFactoryId *id;\
+    _transport_factory_id_lock(); \
```
furiel: Feel free to ignore this comment if you feel too much of boilerplate.

From clean code perspective having the constructor (not the register) under lock feels strange. The reason we do this is the sequence number needs to be safely incremented. So we reuse the lock created for GList *transport_factory_ids. This should work, but normally these two should be independent. Instead, I suggest having sequence as an atomic counter, sequence_inc could return int instead of void, and internally it could call g_atomic_int_add (because this returns the original value safely). Then the constructor could be moved outside the lock.

furiel:
I have no other comments. Just ping me if you want to drop my optional comment, and I will merge the pr.

lib/transport/transport-factory-id.h
lbudai: @furiel: thanks for the comment.
The lock is not for the sequence (using atomic is reasonable) but for checking if `id` is exists or not.

furiel: Oh ok. Then the lock is needed indeed.

PR #2088, opened 2018-06-05
cfg-block: mandatory parameters

furiel:
The goal of this patchset is that block definitions could express that a parameter must be overridden by user. When a parameter is defined with empty brackets, then the parameter becomes mandatory.

Example:
```
block parser foo(bar()) {
  kv-parser();
};

log {
    parser { foo(); };
};
```

In the example above, the `bar` parameter is mandatory, but user did not pass it in the reference block. The result is an error message, and initialization failure.

```
[2018-06-05T08:26:48.015264] Mandatory parameter not overridden; parameter='bar', reference='../etc/block.conf:8:14'
Error parsing log statement, syntax error, unexpected LL_IDENTIFIER, expecting '}' in ../etc/block.conf:
3       block parser foo(bar()) {
4         kv-parser();
5       };
6       
7       log {
8----->     parser { foo(); };
8----->              ^^^^^
9       };
```

If an scl still needs a parameter to expand into nothing by default, because there is no natural default parameter (like hook-commands or tls), an empty double doublequote will that. Check default-network-drivers in this patchset for example usage.

From implementation point of view:
In the original implementation, when a parameter had empty bracket, the grammar generated an empty string. As such, we are not able do decide if the parameter is "mandatory", or the definition has the empty string as default value: `bar()` vs `bar("")`.

So I changed the block grammar, so that when block finds an empty bracket, `NULL` pointer is generated instead of empty string. With this change, I could iterate through the parameters looking for `NULL` value, and check if if user provided a value in the reference block.


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

furiel:
It seems I cannot get default-network-drivers fixed cleanly. The example of default-network-drivers shows that it is useful if I can generate truly empty output. Note if someone uses `foo("")` in the current version then it will be expanded to `""` inside the user block, which is not an empty string.

I am thinking of creating a special keyword in the block grammar: like `foo(_EMPTY_)`, that would be handled as a special case: empty string is generated and the mandatory error is not triggered.

The alternative would be to leave anything alone, and create a special keyword for the NULL generation: like `foo(NULL)` or `foo(REQUIRED)`. My problem with the former is it is counter-intuitive that this means required. My problem with the latter is that from block point of view, REQUIRED does not mean much in that context.

That's why I would go that `foo()` means mandatory, and `foo(_EMPTY_)` means empty expansion.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

scl/default-network-drivers/plugin.conf
```diff
@@ -27,9 +27,9 @@ block source default-network-drivers(
 	rfc5424-tls-port(6514)
 	rfc5424-tcp-port(601)
 	log-msg-size(65536)
-	flags()
-	tls()
-	hook-commands()
+	flags(_EMPTY_)
```
bazsi: The idea is great, but i don't like the token _EMPTY_. Can't we simply use the quote characters?

eg. flags("") means that it's optional and the default is an empty string.

lib/cfg-block.c
```diff
@@ -121,6 +151,9 @@ cfg_block_generate(CfgBlockGenerator *s, GlobalConfig *cfg, CfgArgs *args, GStri
   GError *error = NULL;
   gchar buf[256];
 
+  if (!_validate_mandatory_options(self->arg_defs, args, reference))
+    return FALSE;
+
   if (!_validate_args(args, self->arg_defs, reference))
     return FALSE;
```
bazsi: It's a nitpick but i think mandatory option checking should be part of validate_args. It's getting the same arguments but in a different order.

scl/default-network-drivers/plugin.conf
furiel: The only downside I see is when an option does not expect string, it looks strange to pass "" to it, for example in case of hook-commands or tls. Other than that, I do not see any technical reason why shouldn't this work.

I will try this out and report back if I run into any trouble with it.

bazsi: hmm.. you have a point here. The "" may not be the best token after all.

Just le me try to think about it:
  * port() -- the argument is required
  * port(anything) -- the argument has a default and is not required
  * port("") -- the argument defaults to an empty string and is not required

And the last one is what we are still debating, right? Clearly if port is assumed to be numeric, this might look strange. My argument on the other side is that if something indeed can be an empty string, then it's actually a string and not anything else.

I am ok with this format, are you?

furiel: Yes, we need some token to express empty string. If we can reuse `""`, it is less awkward than having an `_EMPTY_` token for the same purpose. The thing is the suggested version does not work seamlessly, I ran into trouble with an scl, which contained prefix(`prefix`), which was expanded to prefix(), hence configuration error. I could workaround this instance by writing `prefix("`prefix"`)` instead. I'll try to patch all similar cases and see if any other problem occurs.

bazsi:
I dont understand the example, probably github garbled the characters a
bit. Can you name the scl file where the problem occurred?

kira-syslogng:
Build SUCCESS

furiel:
Using the double doublequotes instead of the `_EMPTY_` token passed all tests. The new version feels more natural indeed, so I updated the PR accordingly. Removing WIP flag.

lib/cfg-args.c
```diff
@@ -57,7 +57,7 @@ _resolve_unknown_blockargs_as_varargs(gpointer key, gpointer value, gpointer use
   CfgArgs *defaults = ((gpointer *) user_data)[0];
   GString *varargs = ((gpointer *) user_data)[1];
 
-  if (!defaults || cfg_args_get(defaults, key) == NULL)
+  if (!defaults || !cfg_args_contains(defaults, key))
```
gaborznagy: what about cfg_args_get() normalize_key() call? Isn't it missing here?

lib/cfg-block.c
```diff
@@ -90,9 +90,43 @@ _validate_args_callback(gpointer k, gpointer v, gpointer user_data)
...
+};
+
 static gboolean
 _validate_args(CfgArgs *self, CfgArgs *defs, const gchar *reference)
 {
+  if (!_validate_mandatory_options(defs, self, reference))
+    return FALSE;
+
   gboolean problem_found = FALSE;
   gpointer validate_params[] = { defs, (gchar *) reference, &problem_found };
```
Kokan: Would it be possible to hide this check also with a function just like the `_validate_args_callback` ?

lib/cfg-lex.l
```diff
@@ -255,6 +255,8 @@ word	[^ \#'"\(\)\{\}\[\]\\;\r\n\t,|\.@:]
      /* rules to parse a block as a LL_BLOCK token */
 <block>{white}+            ;
 <block>[^{( \t]+           { return LL_ERROR; }
+<block>\({white}*\)        { yy_pop_state(yyscanner); yylval->cptr = NULL; return LL_BLOCK; }
+<block>\({white}*\"\"{white}*\) { yy_pop_state(yyscanner); yylval->cptr = g_strdup(""); return LL_BLOCK; }
```
Kokan: I think this rule is redundant, does the same thing without it.

lib/cfg-lex.l
```diff
@@ -255,6 +255,8 @@ word	[^ \#'"\(\)\{\}\[\]\\;\r\n\t,|\.@:]
      /* rules to parse a block as a LL_BLOCK token */
 <block>{white}+            ;
 <block>[^{( \t]+           { return LL_ERROR; }
+<block>\({white}*\)        { yy_pop_state(yyscanner); yylval->cptr = NULL; return LL_BLOCK; }
```
gaborznagy: I'm not an expert of lexer. Do we need pop_state here?
Can it cause an issue if in block context/state we pop the state, but in the next line we find an option with default value?
I've tried to test it but so far no problem found.

lib/cfg-args.c
MrAnno: Yeah, I also think `cfg_args_contains()` should normalize the key before the lookup.

lib/cfg-lex.l
Kokan: Yes, that pop state is needed. Although somewhat not intuitive because the push state is triggered from the grammar, and one could argue that the pop could be moved there for the same reason. (Not directly in the grammar, but triggered by an item pushed by the grammar) But we need to pop the block state at this point, because the next item going to be an `LL_INDENTIFIER`  (if config is valid), that is not part of the block.

No it cannot.  There is a general rule `<*>\r?\n` that takes care of the new line, so this is a reason it is not a problem.

gaborznagy: Thanks for clarifying!

lib/cfg-args.c
furiel: Nice catch! For some reason cfg_args_get() first tries to get without normalizing, and in case of failure, she falls back to normalizing. I don't know why, because cfg_args_set always normalizes. I skip the non-normalized check in my patch.

lib/cfg-block.c
furiel: Fixed it: I named it _validate_spurious_args()

furiel:
Back to wip. I try to create a unit test, both for the mandatory options and the varargs, if possible.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2098, opened 2018-06-06
WIP: Riemann batching

bazsi:
Just like the SQL version, this one migrates riemann to use the new batching capabilities of LogThreadedDestDriver instead of its slightly broken version.

This is still work in progress, don't merge.


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:
Is there anyone who can test riemann end-to-end? These patches are pretty trivial, however I don't have a riemann environment to test it against.

@algernon perhaps?

bazsi:
After some kind of testing this should be ready to go in.

kira-syslogng:
Build SUCCESS

bazsi:
I did an end-to-end testing with the help of https://hub.docker.com/r/rlister/riemann/ and batching
seems to work fine.

In order to make it easier to confirm this, I've added a number of new log messages and also added extra data to the existing ones.

This series also fixes a major memory leak in the riemann driver in case sending the event fails.

I think with this testing this should be good enough to go in.

I hope that those who use riemann will eventually confirm that I didn't break anything.
@faxm0dem

kira-syslogng:
Build SUCCESS

bazsi:
Rebased to master, so the test failure should be addressed.

kira-syslogng:
Build FAILURE

bazsi:
 
for some reason kira failed in the cmake version of the build on source file related tests. This seems to be a race condition of some kind. I am restarting the tests. @pzoleex

kira-syslogng:
Build SUCCESS

modules/riemann/riemann.c
```diff
@@ -726,7 +718,6 @@ riemann_dd_new(GlobalConfig *cfg)
 
   self->super.worker.disconnect = riemann_dd_disconnect;
   self->super.worker.insert = riemann_worker_insert;
-  self->super.worker.thread_deinit = riemann_worker_thread_deinit;
```
gaborznagy: Maybe I don't understand, but does LogThreadedDestDriver really flush on deinit?
I don't see any registration of the `do_work` task which would eventually call `_perform_flush`.

Handling shutdown:
- I see `_shutdown_event_callback` event handler, which is called on deinit, but I only see task un-registering steps and no `do_work` scheduling or direct `flush` calling.
- I don't see any flush operation in `_worker_thread` either.

However, the original code with `riemann_worker_thread_deinit` calling `riemann_worker_batch_flush`  would not work either since `log_threaded_dest_worker_thread_deinit` is called after `_disconnect`.

As far as I see this concerns the http module as well.

bazsi:
The perform_work() did an explicit flush after the loop, however you are be
right that with flush-timeout() this might have changed.

This means that the thread might exit without flushing messages, but this
shouldn't be in thread_deinit even in this case.

We should do an explicit flush, and possibly a rewind when the thread exits.

PR #2099, opened 2018-06-07
force syslog-ng SIGCHLD handlers on top

Kokan:
The initial issue:

In case of both java destination and program source/destination configured, syslog-ng cannot reap the program source/destination, which is started in a fork. Those fork remains in Zombie state until syslog-ng is restarted.

This is because the java virtual machine  sets  its own  signal handler. The syslog-ng reaping
happens when the SIGCHLD handler is  called, but the java replaced our SIGCHLD
handler. Therefore every forked process became a zombie, there is nobody calling wait on them.

A proposed solution:

The sigaction is registers the signal handlers, to overcome the java own
registration process in syslog-ng we implement our own sigaction implementation,
which is called instead of the one in libc. Every single case our sigaction
calls the original one, except in case of SIGCHLD. If it is called with SIGCHLD
out of setup phase, the new handler is saved and called, when we are not waiting
for SIGCHLD.

Questions:
* Right now it is called signal chaining, but only one handler is saved and triggered. Is it okay to save just the latest and do not trigger the other handlers ?
* Should it trigger the saved signal handler even if syslog-ng is sure that it was sent by the system because of forked process ?

### DIFF
--- a/lib/children.c
+++ b/lib/children.c
@@ -69,7 +69,7 @@ child_manager_unregister(pid_t pid)
 void
 child_manager_register_external_sigchld_handler(void (*callback)(int))
 {
-   sigchld_callback = callback;
+  sigchld_callback = callback;
 }
 
 void
--- a/lib/children.c
+++ b/lib/children.c
@@ -33,6 +33,7 @@ typedef struct _ChildEntry
 } ChildEntry;
 
 GHashTable *child_hash;
+static void (*sigchld_callback)(int);
 
 static void
 child_manager_child_entry_free(ChildEntry *ce)
@@ -66,6 +67,12 @@ child_manager_unregister(pid_t pid)
 }
 
 void
+child_manager_register_external_sigchld_handler(void (*callback)(int))
+{
+   sigchld_callback = callback;
+}
+
+void
 child_manager_sigchild(pid_t pid, int status)
 {
   ChildEntry *ce;
@@ -76,6 +83,9 @@ child_manager_sigchild(pid_t pid, int status)
       ce->exit_callback(pid, status, ce->callback_data);
       g_hash_table_remove(child_hash, &pid);
     }
+
+  if (sigchld_callback)
+    sigchld_callback(SIGCHLD);
 }
 
 void
--- a/lib/signal-handler.c
+++ b/lib/signal-handler.c
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2018 Balabit
+ * Copyright (c) 2018 Kokan
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+#include "signal-handler.h"
+
+#include "syslog-ng.h"
+#include "children.h"
+
+#ifndef _WIN32
+#include <dlfcn.h>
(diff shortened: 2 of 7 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

Kokan:

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

Kokan:
@kira-syslogng retest this please;  branch=micek_SYSLOGDEV-4290; test=functions/destination-drivers/hdfs-destination/009-reload-when-child-process-exists.zts;

lib/children.c
```diff
@@ -33,6 +33,7 @@ typedef struct _ChildEntry
 } ChildEntry;
 
 GHashTable *child_hash;
+static void (*sigchld_callback)(int) = NULL;
```
furiel: There is no need to initialize here.

lib/signal-handler.c
```diff
@@ -0,0 +1,96 @@
...
+#include "signal-handler.h"
+
+#include "syslog-ng.h"
+#include "children.h"
+
+#ifndef _WIN32
+#include <dlfcn.h>
+#endif
+
+static const struct sigaction *sgchld_handler = NULL;
```
furiel: Initialization is not needed here.

lib/signal-handler.c
```diff
@@ -0,0 +1,96 @@
...
+
+#ifndef _WIN32
+static int
+call_original_sigaction(int signum, const struct sigaction *act, struct sigaction *oldact)
+{
+  static int (*real_sa)(int, const struct sigaction *, struct sigaction *) = NULL;
+
+  if (real_sa == NULL)
+    {
+      real_sa = dlsym(RTLD_NEXT, "sigaction");
```
furiel: > Right now it is called signal chaining, but only one handler is saved and triggered. Is it okay to save just the latest and do not trigger the other handlers ?

Would the problem resolved by collection the function pointer into GPtrArray. It would not introduce much extra code complexity, considering how tricky the implementation is.

lib/signal-handler.c
```diff
@@ -0,0 +1,96 @@
...
+    {
+      sgchld_handler->sa_handler(signum);
+    }
+}
+
+#ifndef _WIN32
+static int
+call_original_sigaction(int signum, const struct sigaction *act, struct sigaction *oldact)
+{
+  static int (*real_sa)(int, const struct sigaction *, struct sigaction *) = NULL;
```
furiel: Initialization is not needed here.

lib/signal-handler.c
furiel: We concluded with @Kokan that my suggestion wouldn't work. Other modules instances could be written with actual care, and save the original function pointer before overwriting. So in the end, I could not decide if a pointer in GPtrArray should be called or not.

I agree with the current version.

lib/signal-handler.c
```diff
@@ -0,0 +1,96 @@
...
+    {
+      real_sa = dlsym(RTLD_NEXT, "sigaction");
+    }
+  return real_sa(signum, act, oldact);
+}
+
+static gboolean
+_save_handler(const struct sigaction *act)
+{
+  static gboolean is_first_handler = FALSE;
```
lbudai: I think it would be better to change this to this one:
```
  static gboolean is_first_handler = TRUE;                                                                                                                                                                          
  if (is_first_handler)                                                                                                                                                                                             
    {                                                                                                                                                                                                               
      is_first_handler = FALSE;                                                                                                                                                                                     
      return FALSE;                                                                                                                                                                                                 
    } 
```

lib/children.c
```diff
@@ -65,6 +66,12 @@ child_manager_unregister(pid_t pid)
     }
 }
 
+void
+child_manager_register_external_sigchld_handler(void (*callback)(int))
+{
+  sigchld_callback = callback;
```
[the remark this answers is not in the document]
Kokan: There are three component, where the `sigaction` is called (there are other guarantee as well):
* mainloop `setup_signals`
  It is called before the configuration parse, or any pipe is created.
* python module, when the module loaded
  There cannot be race because it is called - during the configuration parse - from the same main thread as the `setup_signals`.
* java module; this is actually for certainly called from a different thread than the other two, the one that is created by the `JVM`. But the `JVM` is started only, when the java destination `init` is called. The `init` function for the pipes are called after parsing the configuration file (only after the modules are already loaded).

Kokan:
@lbudai Yes, it is from you ("The hypothesis behind this question maybe originated from me..")

kira-syslogng:
Build FAILURE

Kokan:

kira-syslogng:
Build SUCCESS

PR #2101, opened 2018-06-08
Always add test datafiles to the dist tarball

MrAnno:
`lib/tests` contains `EXTRA_DIST` inputs for tests, for example: `lib/tests/testdata-lexer/`. They should not be excluded from the dist tarball, even if testing is disabled.

Note that only `EXTRA_DIST` is affected by automake's `if` conditional, source files can't be excluded from the tarball by `if`. automake handles them automatically (see `SOURCES_DIST`).


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2103, opened 2018-06-10
Fix a bug in the old LogMessage deserialization

MrAnno:
`sd_element_name` is a pointer, `sizeof()` can not be called on it to determine the array size.

Detected by gcc 8.1.1, configured with `--enable-extra-warnings`.


kira-syslogng:
Build SUCCESS

Kokan:
Nice catch! I am kinda sad that it works anyway.

PR #2108, opened 2018-06-13
confgen: fix reading the output of the program

bazsi:

### DIFF
--- a/modules/confgen/confgen-plugin.c
+++ b/modules/confgen/confgen-plugin.c
@@ -54,13 +54,25 @@ typedef struct _ConfgenExec
   gchar *exec;
 } ConfgenExec;
 
+static void
+_read_program_output(FILE *out, GString *result)
+{
+  gchar buf[1024];
+  gsize res;
+
+  while ((res = fread(buf, 1, sizeof(buf), out)) > 0)
+    {
+      g_string_append_len(result, buf, res);
+    }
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

modules/confgen/confgen-plugin.c
```diff
@@ -79,12 +91,8 @@ confgen_exec_generate(CfgBlockGenerator *s, GlobalConfig *cfg, CfgArgs *args, GS
                 evt_tag_error("error"));
       return FALSE;
     }
-  g_string_set_size(result, 1024);
```
szemere: note:
inside [cfg_lexer_lex](https://github.com/balabit/syslog-ng/blob/89e435abc964ca566ae75e64eae2d2dbfbc2ffa3/lib/cfg-lexer.c#L868), before calling _confgen_exec_generate_ via [cfg_block_generator_generate](https://github.com/balabit/syslog-ng/blob/89e435abc964ca566ae75e64eae2d2dbfbc2ffa3/lib/cfg-lexer.c#L944) we created **result** as a 256byte Gstring. With the old version, we resized **result** to 1024bytes before using it.
Any final thoughts about the optimal "buffer" size? (The new g_string_append_len LGTM, should we remove the pre- allocation?)

PR #2117, opened 2018-06-16
[WIP] $(context-length) should not count synthetic messages (PatternDB, grouping-by())

MrAnno:
Counting synthetic messages with $(context-length) seems intuitive, but its side effects are not intuitive at all.

For example:
Triggering a context close event with `"$(context-length)" == "N"`, where the user wants to include `$(context-length)` in the message looks really weird (the output is N+1).

```
@version: 3.16

parser contextlength_test {
    grouping-by(
        key("${PROGRAM}")
        trigger("$(context-length)" == "5")
        aggregate(
            value("MESSAGE" "We have $(context-length) messages out of 5 considering a 30 secs timeout.")
        )
        timeout(30)
    );
};

log {
    source { stdin(); };
    parser(contextlength_test);
};
```

Input:
```
program1 m1
program1 m2
program1 m3
program1 m4
program1 m5
```

Output prior to this PR:
```
We have 6 messages out of 5 considering a 30 secs timeout.
```

--------

The same applies to PatternDB:
```
<action condition='"$(context-length)" == "5"'>
        <message inherit-properties='true'>
        <tags>
        <tag>pdb_action</tag>
        </tags>

        <values> <value name="ACTION_MSG">$(context-length) is equal to 5 for '${.CEF.dhost}'</value> </values>
        </message>
</action>
```

--------

Questions:
- @bazsi What do you think about this? AFAIK this can not break old configs, but it may change the output of the last synthetic message. I know it can be argued whether this is a bug or not..

This problem was reported internally by a customer.

### DIFF
--- a/modules/dbparser/synthetic-context.h
+++ b/modules/dbparser/synthetic-context.h
@@ -41,6 +41,4 @@ void synthetic_context_set_context_scope(SyntheticContext *self, const gchar *sc
 void synthetic_context_init(SyntheticContext *self);
 void synthetic_context_deinit(SyntheticContext *self);
 
-gint synthetic_context_lookup_context_scope(const gchar *context_scope);
-
 #endif
--- a/modules/dbparser/synthetic-message.c
+++ b/modules/dbparser/synthetic-message.c
@@ -215,9 +215,7 @@ synthetic_message_generate_with_context(SyntheticMessage *self, CorrellationCont
       g_assert_not_reached();
       break;
     }
-  g_ptr_array_add(context->messages, genmsg);
   synthetic_message_apply(self, context, genmsg, buffer);
-  g_ptr_array_remove_index_fast(context->messages, context->messages->len - 1);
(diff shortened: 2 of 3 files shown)


kira-syslogng:
Build FAILURE

MrAnno:
Well, I did change message indexing with this path, which definitely breaks old configs.. :disappointed:

MrAnno:
Well, I completely forgot about an essential functionality: referencing earlier messages of the context. 
This PR breaks that, back to WIP.

MrAnno:
I think I should rephrase the problem:
Synthetic messages must be placed inside their context.

But do we want to count them when $(context-length) is used?

bazsi:
It has always been part of the context, thus $(context-length) should count
them.

MrAnno:
Thanks.
Then I'll just add an example to the documentation about the mentioned confusing case.

PR #2118, opened 2018-06-17
Rewrite patternDB test cases based on criterion

norberttak:
Rewrite patternDB test cases based on criterion

### DIFF
--- a/modules/dbparser/tests/CMakeLists.txt
+++ b/modules/dbparser/tests/CMakeLists.txt
@@ -1,6 +1,7 @@
 add_unit_test(TARGET test_timer_wheel DEPENDS patterndb)
 add_unit_test(TARGET test_patternize DEPENDS patterndb syslogformat)
 add_unit_test(CRITERION LIBTEST TARGET test_patterndb DEPENDS patterndb basicfuncs syslogformat)
+add_unit_test(CRITERION TARGET test_parsers_e2e DEPENDS patterndb basicfuncs syslogformat)
 add_unit_test(TARGET test_radix DEPENDS patterndb)
 target_compile_options(test_radix PRIVATE "-Wno-error=pointer-sign")
 
--- a/modules/dbparser/tests/CMakeLists.txt
+++ b/modules/dbparser/tests/CMakeLists.txt
@@ -5,5 +5,5 @@ add_unit_test(TARGET test_radix DEPENDS patterndb)
 target_compile_options(test_radix PRIVATE "-Wno-error=pointer-sign")
 
 # test_parsers includes a .c file
-add_unit_test(LIBTEST TARGET test_parsers INCLUDES ${PATTERNDB_INCLUDE_DIR})
+add_unit_test(CRITERION TARGET test_parsers INCLUDES ${PATTERNDB_INCLUDE_DIR})
 target_compile_options(test_parsers PRIVATE "-Wno-error=pointer-sign")
--- a/modules/dbparser/tests/CMakeLists.txt
+++ b/modules/dbparser/tests/CMakeLists.txt
@@ -1,6 +1,6 @@
 add_unit_test(TARGET test_timer_wheel DEPENDS patterndb)
 add_unit_test(TARGET test_patternize DEPENDS patterndb syslogformat)
-add_unit_test(LIBTEST TARGET test_patterndb DEPENDS patterndb basicfuncs syslogformat)
+add_unit_test(CRITERION LIBTEST TARGET test_patterndb DEPENDS patterndb basicfuncs syslogformat)
 add_unit_test(TARGET test_radix DEPENDS patterndb)
 target_compile_options(test_radix PRIVATE "-Wno-error=pointer-sign")
 
--- a/modules/dbparser/tests/test_parsers.c
+++ b/modules/dbparser/tests/test_parsers.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013 Balabit
+ * Copyright (c) 2018 Balabit
  * Copyright (c) 2013 Balázs Scheidler <balazs.scheidler@balabit.com>
  *
  * This program is free software; you can redistribute it and/or modify it
@@ -21,30 +21,12 @@
  *
  */
 
-#include "testutils.h"
+#include <criterion/criterion.h>
+#include <criterion/parameterized.h>
 #include <string.h>
 
 /* NOTE: including the implementation file to access static functions */
 #include "radix.c"
-
-
-#define PARSER_TESTCASE(x, ...) do { parser_testcase_begin(#x, #__VA_ARGS__); x(__VA_ARGS__); parser_testcase_end(); } while(0)
-
-#define parser_testcase_begin(func, args)                     \
-  do                                                            \
-    {                                                           \
-      testcase_begin("%s(%s)", func, args);                     \
-    }                                                           \
-  while (0)
-
-#define parser_testcase_end()                                 \
-  do                                                            \
-    {                                                           \
-      testcase_end();                                           \
-    }                                                           \
-  while (0)
-
-
 
 static gboolean
 _invoke_parser(gboolean (*parser)(guint8 *str, gint *len, const gchar *param, gpointer state, RParserMatch *match),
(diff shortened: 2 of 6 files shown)


kira-syslogng:
Build SUCCESS

modules/dbparser/tests/test_patterndb.c
```diff
@@ -706,434 +441,223 @@ test_simple_rule_with_rate_limited_action(void)
...
-  g_file_open_tmp("patterndbXXXXXX.xml", &filename, NULL);
-  g_file_set_contents(filename, tag_outside_of_rule_skeleton,
-                      strlen(tag_outside_of_rule_skeleton), NULL);
+ParameterizedTestParameters(pattern_db, test_rules)
+{
+  static PatternDBTestParam parser_params[] =
+  {
+    // test_conflicting_rules_with_different_parsers
+    {
+      .pattern_db_file = PATH_PREFIX"pdb_conflicting_rules_with_different_parsers.xml",
```
Kokan: I think this points toward a wrong direction. The unit test should not have external files that they are depending on. 
I know that the existing version created and loaded that file (but I think that was less evil.)

This is probably out of current PR's scope, just to give you a hint, where I think it should move and why is it important not to imply the current is the good direction.

The `PatternDB` itself should not know where its contents come: file, user input, stream or whatever. Try to think how you would modify the `PatternDB` if you would want to add a feature like embedded `PatternDB` in syslog-ng configuration (like `python`).

```
patterndb inlinepattern {
 <patterndb version='5' pub_date='2010-02-22'>
 ... 
</patterndb>
};
```

Actually this is exactly what we would like to achieve in the unit test IMHO.

(Anyway if this is kept as it is - :disappointed: , those files should be added to the `EXTRA_DIST` target.)

modules/dbparser/tests/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 add_unit_test(TARGET test_timer_wheel DEPENDS patterndb)
 add_unit_test(TARGET test_patternize DEPENDS patterndb syslogformat)
-add_unit_test(CRITERION LIBTEST TARGET test_patterndb DEPENDS patterndb basicfuncs syslogformat)
+add_unit_test(LIBTEST TARGET test_patterndb DEPENDS patterndb basicfuncs syslogformat)
```
Kokan: This patch reverts this line, this line should not be changed in this patch.

kira-syslogng:
Build SUCCESS

modules/dbparser/tests/test_parsers.c
```diff
@@ -74,56 +56,77 @@ _invoke_parser(gboolean (*parser)(guint8 *str, gint *len, const gchar *param, gp
...
-  gboolean result;
+typedef gboolean (*ParserFunc)(guint8 *str, gint *len, const gchar *param, gpointer state, RParserMatch *match);
 
-  result = _invoke_parser(parser, str, param, state, &result_string);
-  assert_true(result, "Mismatching parser result");
-  assert_string(result_string, expected_string, "Mismatching parser result");
-  g_free(result_string);
+typedef struct _parser_test_param
+{
+  ParserFunc parser;
```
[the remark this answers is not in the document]
norberttak: OK, removed the parser member from structure

modules/dbparser/tests/test_parsers.c
```diff
@@ -136,35 +139,16 @@ compile_qstring_state(const gchar *quotes)
...
-
-void
-test_qstring(void)
+ParameterizedTest(ParserQStringTestParam *param, parser, test_qstring_parser)
 {
-  PARSER_TESTCASE(test_qstring_parser_extracts_word_from_quotes);
-}
-
-/* NOTE: incomplete, the rest of the testcases are missing */
+//  assert_parser_success(r_parser_qstring, str, quotes, _compile_qstring_state(quotes), expected_string);
```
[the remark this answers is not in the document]
norberttak: removed

modules/dbparser/tests/test_parsers.c
```diff
@@ -74,56 +56,77 @@ _invoke_parser(gboolean (*parser)(guint8 *str, gint *len, const gchar *param, gp
...
-  result = _invoke_parser(parser, str, param, state, &result_string);
-  assert_true(result, "Mismatching parser result");
-  assert_string(result_string, expected_string, "Mismatching parser result");
-  g_free(result_string);
+typedef struct _parser_test_param
+{
+  ParserFunc parser;
+  gchar *str;
+  gpointer param;
+  gpointer state;
```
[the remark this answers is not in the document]
norberttak: removed

modules/dbparser/tests/test_patterndb.c
norberttak: I don't really catch the point. I need some more details about your concern. let's discuss it personally...

norberttak: I removed the XML files and moved their content into test_patterndb.h header file.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

norberttak:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/dbparser/tests/test_parsers.c
```diff
@@ -74,56 +56,72 @@ _invoke_parser(gboolean (*parser)(guint8 *str, gint *len, const gchar *param, gp
...
+    {.str = "{foo}", .quotes = "{}", .expected_string = "foo"},
+  };
 
-void
-test_string(void)
-{
-  PARSER_TESTCASE(test_string_parser_without_parameter_parses_a_word);
-  PARSER_TESTCASE(test_string_parser_with_additional_end_characters);
+  return cr_make_param_array(ParserQStringTestParam, parser_params,
+                             sizeof(parser_params) / sizeof(parser_params[0]));
```
furiel: Can I suggest https://developer.gnome.org/glib/stable/glib-Standard-Macros.html#G-N-ELEMENTS:CAPS?

modules/dbparser/tests/test_parsers.c
```diff
@@ -74,56 +56,72 @@ _invoke_parser(gboolean (*parser)(guint8 *str, gint *len, const gchar *param, gp
...
+  gchar *result_string = NULL;
   gboolean result;
 
-  result = _invoke_parser(parser, str, param, state, NULL);
-  assert_false(result, "Mismatching parser result");
+  result = _invoke_parser(r_parser_string, param->str, param->param, NULL, &result_string);
+  if (param->expected_result == TRUE)
+    {
+      cr_assert(result, "Mismatching parser result");
+      cr_assert_str_eq(result_string, param->expected_string, "Mismatching parser result");
```
furiel: I'm afraid this hides the strings in case of error. We should either not provide an error string, or print the underlying strings similarly to the bottom of the file as %s %s.

modules/dbparser/tests/test_patterndb.h
```diff
@@ -0,0 +1,538 @@
...
+              </tags>\
+            </message>\
+          </action>\
+        </actions>\
+     </rule>\
+    </rules>\
+  </ruleset>\
+</patterndb>"
+
+#define pdb_msg_count_skeleton "<patterndb version='4' pub_date='2010-02-22'>\
```
furiel: If I understand correctly, these were originally gchar * types. Why is these changed to defines?

modules/dbparser/tests/CMakeLists.txt
Kokan: The 2nd patch modified the `add_unit_test(... test_patterndb ...)`, so it is `add_unit_test(CRITERION LIBTEST TARGET test_patterndb DEPENDS patterndb basicfuncs syslogformat)` with Criterion.
And the last (3rd patch), reverts the above change, so the result is the `test_patterndb` is marked as non-criterion test.
(In the unified diff it cannot be seen.)

modules/dbparser/tests/test_parsers.c
norberttak: I added the result and expected strings to printout.

modules/dbparser/tests/test_patterndb.h
norberttak: I made the XML strings as a macro because I don't like to create variables in header files. If we want to use the const gchar*  then we should create C file instead of H.
I'd like to keep it as is.

modules/dbparser/tests/test_parsers.c
norberttak: I added the G_N_ELEMENTS macro.

modules/dbparser/tests/CMakeLists.txt
norberttak: Thanks for the finding. I clean up the commits.

kira-syslogng:
Build SUCCESS

PR #2121, opened 2018-06-18
Fix global "center;;received" counter when systemd-journal() is used

MrAnno:
This PR enables the global "center;;received" counter on the `systemd-journal` source.


kira-syslogng:
Build SUCCESS

PR #2122, opened 2018-06-19
Confgen and pragma improvements

bazsi:
Add unit/integration test for the confgen module.

### DIFF
--- a/lib/pragma-grammar.ym
+++ b/lib/pragma-grammar.ym
@@ -52,6 +52,10 @@
 
 #include <stdlib.h>
 
+#define PRAGMA_BEGIN() ({ lexer->tokenize_eol++; })
+#define PRAGMA_END() ({ lexer->tokenize_eol--; })
+#define PRAGMA_ERROR() ({ PRAGMA_END(); YYERROR; })
+
 CfgArgs *last_module_args = NULL;
 
 gint
@@ -81,8 +85,8 @@ __process_version_string(gchar *value)
 %%
 
 start
-        : { lexer->tokenize_eol++; }
-          stmt_and_eol                                  { lexer->tokenize_eol--; if (yychar != YYEMPTY) { cfg_lexer_unput_token(lexer, &yylval); } YYACCEPT; }
+        : { PRAGMA_BEGIN(); }
+          stmt_and_eol                                  { if (yychar != YYEMPTY) { cfg_lexer_unput_token(lexer, &yylval); } PRAGMA_END(); YYACCEPT; }
 
 stmt_and_eol
         : pragma_stmt LL_EOL
@@ -117,8 +121,8 @@ module_stmt
             free($2);
             cfg_args_unref(last_module_args);
             last_module_args = NULL;
-	    if (!success)
-              YYERROR;
+            if (!success)
+              PRAGMA_ERROR();
           }
 	;
 
@@ -133,7 +137,7 @@ requires_stmt
                                 evt_tag_str("module", $2),
                                 cfg_lexer_format_location_tag(lexer,&@1));
                     free($2);
-                    YYERROR;
+                    PRAGMA_ERROR();
                   }
                 else
                   {
--- a/lib/pragma-grammar.ym
+++ b/lib/pragma-grammar.ym
@@ -113,10 +113,12 @@ define_stmt
 module_stmt
 	: KW_MODULE string { last_module_args = cfg_args_new(); } module_params
           {
-            plugin_load_module(&configuration->plugin_context, $2, last_module_args);
+            gboolean success = plugin_load_module(&configuration->plugin_context, $2, last_module_args);
             free($2);
             cfg_args_unref(last_module_args);
             last_module_args = NULL;
+	    if (!success)
+              YYERROR;
           }
 	;
 
--- a/lib/pragma-grammar.ym
+++ b/lib/pragma-grammar.ym
@@ -115,14 +115,21 @@ define_stmt
 								  cfg_args_set(configuration->globals, $2, $3); free($2); free($3); }
 
 module_stmt
-	: KW_MODULE string { last_module_args = cfg_args_new(); } module_params
+	: KW_MODULE string { last_module_args = cfg_args_new(); } module_param module_params
           {
             gboolean success = plugin_load_module(&configuration->plugin_context, $2, last_module_args);
             free($2);
+
             cfg_args_unref(last_module_args);
             last_module_args = NULL;
+
             if (!success)
               PRAGMA_ERROR();
+          }
+	| KW_MODULE string
+          {
+            plugin_load_module(&configuration->plugin_context, $2, NULL);
+            free($2);
           }
 	;
 
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -1013,10 +1013,19 @@ cfg_lexer_preprocess(CfgLexer *self, gint tok, YYSTYPE *yylval, YYLTYPE *yylloc)
   else if (tok == LL_PRAGMA)
     {
       gpointer dummy;
+      CfgIncludeLevel *level = &self->include_stack[self->include_depth];
 
       cfg_lexer_append_preprocessed_output(self, "@");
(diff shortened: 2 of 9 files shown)


bazsi:
There are two extra patches on top of the new test program:

* fix location for syntax errors in pragma statements (similar to what we had for generator references a few weeks back)
* add validation for the context() value in confgen, which was completely unvalidated earlier, causing confgen declarations not to be found.

Otherwise it is a few libtest refactors and a new unit test program. Let me know if you want me to extract the two extra patches into a separate PR.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/tests/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ add_unit_test(CRITERION TARGET test_parse_number)
 add_unit_test(CRITERION TARGET test_reloc)
 add_unit_test(CRITERION TARGET test_hostname)
 add_unit_test(CRITERION LIBTEST TARGET test_rcptid)
-add_unit_test(CRITERION TARGET test_lexer)
+add_unit_test(LIBTEST TARGET test_lexer)
```
Kokan: You should not remove `CRITERION` from this test, as it is still criterion based.
The `CRITERION` and `LIBTEST` are kinda flags for `add_unit_test`, so you could apply both for the same test, and it means which should be linked/inlucded to the current test. In this case probably both of them should be applied.

libtest/Makefile.am
```diff
@@ -21,6 +21,8 @@ libtest_libsyslog_ng_test_a_SOURCES =   \
 	libtest/persist_lib.c		\
 	libtest/mock-transport.c	\
 	libtest/mock-transport.h	\
+	libtest/mock-cfg-parser.c	\
```
Kokan: This should probably added to the `CMakeLists.txt` as well.

libtest/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ set(LIBTEST_HEADERS
     queue_utils_lib.h
     stopwatch.h
     cr_template.h
+    grab-logging.h
```
Kokan: [optional] Only if you modify the patches anyway, and you feel like it:  this would be better in the previous patch (where you extracted the internal message logging).

bazsi:
On Jun 19, 2018 19:29, "Kókai Péter" <notifications@github.com> wrote:

*@Kokan* commented on this pull request.

In lib/pragma-grammar.ym
<https://github.com/balabit/syslog-ng/pull/2122#discussion_r196497635>:

>              free($2);
             cfg_args_unref(last_module_args);
             last_module_args = NULL;
+	    if (!success)

The similar pragma @requires has different effect based on, where it is
called and the module cannot be loaded.
When it is called from a file included from the main configuration file, it
stops processing that particular file, and continues with the next one.
On the other hand, when it is called from the top level configuration file,
the whole configuration is reported as invalid.

I always considered @module as a silent sibling of @requires that tries,
but never fails. Otherwise in case of top level configuration they would
not differ, and in that case it is redundant.
One could argue that since module detection and automatic load, it does not
make sense either not to fail if it cannot load.

So we are at a point where I think @module should be deprecated in favor of
the @requires.


There's one usecase in @module which might not immediately be appearent: it
can provide arguments to module initialization, which confgen relies on.

And with the recent change that can fail.

I am not sure about the right solution though.

In lib/tests/CMakeLists.txt
<https://github.com/balabit/syslog-ng/pull/2122#discussion_r196506063>:

> @@ -5,7 +5,7 @@ add_unit_test(CRITERION TARGET test_parse_number)
 add_unit_test(CRITERION TARGET test_reloc)
 add_unit_test(CRITERION TARGET test_hostname)
 add_unit_test(CRITERION LIBTEST TARGET test_rcptid)
-add_unit_test(CRITERION TARGET test_lexer)
+add_unit_test(LIBTEST TARGET test_lexer)

You should not remove CRITERION from this test, as it is still criterion
based.
The CRITERION and LIBTEST are kinda flags for add_unit_test, so you could
apply both for the same test, and it means which should be linked/inlucded
to the current test. In this case probably both of them should be applied.


Thanks. I thought libtest includes criterion, but I'll add it.

In libtest/Makefile.am
<https://github.com/balabit/syslog-ng/pull/2122#discussion_r196507772>:

> @@ -21,6 +21,8 @@ libtest_libsyslog_ng_test_a_SOURCES =   \
 	libtest/persist_lib.c		\
 	libtest/mock-transport.c	\
 	libtest/mock-transport.h	\
+	libtest/mock-cfg-parser.c	\

This should probably added to the CMakeLists.txt as well.


I thought i did. But I'll check.

In libtest/CMakeLists.txt
<https://github.com/balabit/syslog-ng/pull/2122#discussion_r196510172>:

> @@ -8,6 +8,7 @@ set(LIBTEST_HEADERS
     queue_utils_lib.h
     stopwatch.h
     cr_template.h
+    grab-logging.h

[optional] Only if you modify the patches anyway, and you feel like it:
this would be better in the previous patch (where you extracted the
internal message logging).


Will do.

—
You are receiving this because you authored the thread.
<https://github.com/balabit/syslog-ng/pull/2122#pullrequestreview-130061610>,
or mute the thread

kira-syslogng:
Build SUCCESS

MrAnno:
@bazsi This PR has a few conflicts and 2 minor unresolved review notes.

Do you mind if I resolve them?

lib/tests/CMakeLists.txt
MrAnno: Done.

libtest/Makefile.am
MrAnno: Done.

MrAnno:
@Kokan Could you confirm that we've addressed your review notes, please? :)

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

MrAnno:
@bazsi Please have a look at the last 3 commits. I had to modify your test and fix`tokenize_eol` in `pragma-grammar`.

lib/pragma-grammar.ym
```diff
@@ -117,8 +117,11 @@ module_stmt
...
             cfg_args_unref(last_module_args);
             last_module_args = NULL;
-	    if (!success)
-              YYERROR;
+            if (!success)
+              {
+                lexer->tokenize_eol--;
+                YYERROR;
+              }
           }
```
bazsi: hmm maybe we should extract this as a macro? 

e.g. PRAGMA_ENTER() to increment tokenize_eol
PRAGMA_LEAVE() to decrement it.
PRAGMA_ERROR to indicate failure.

kira-syslogng:
Build SUCCESS

lib/pragma-grammar.ym
```diff
@@ -113,10 +113,12 @@ define_stmt
 module_stmt
 	: KW_MODULE string { last_module_args = cfg_args_new(); } module_params
           {
-            plugin_load_module(&configuration->plugin_context, $2, last_module_args);
+            gboolean success = plugin_load_module(&configuration->plugin_context, $2, last_module_args);
             free($2);
             cfg_args_unref(last_module_args);
             last_module_args = NULL;
+	    if (!success)
```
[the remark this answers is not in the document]
Kokan: Yes, but also that - not part of this PR - the `confgen` could be independent of `@module`, and the later one could be deprecated at that point.

Kokan:
@MrAnno see my reply https://github.com/balabit/syslog-ng/pull/2122#discussion_r206033210

lib/pragma-grammar.ym
MrAnno: Good idea, thanks.

lib/pragma-grammar.ym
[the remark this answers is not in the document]
MrAnno: Done.

kira-syslogng:
Build SUCCESS

PR #2127, opened 2018-06-20
Minor updates to SELinux policy installer script

jszigetvari:
src/root_safe/syslog_ng.el6.fc.in: added conf.d/*.conf to be marked as config files
src/root_safe/syslog_ng.el7.fc.in: added conf.d/*.conf to be marked as config files, removed labels for the init script
src/syslog_ng.el67.te.in: bumped module version to 1.0.6
syslog_ng.sh: updated OS detection code to properly detect CentOS 7, updated code for init-script labeling

    * added conf.d/*.conf to be marked with the label syslog_conf_t
    * removed labels for the init script for RHEL/CentOS 7
    * module version bump to 1.0.6
    * OS detection had problems with CentOS 7, when lsb_release was present, fixed it
    * updated init script relabeling code, to handle when the init scripts are not present

Signed-off-by: Szigetvari Janos <jszigetvari@gmail.com>


kira-syslogng:

Kokan:
@kira-syslogng ok to test

Issue #2078, opened 2018-06-01
Crash when syslog-parser() message parsing fails with a "no-parse" source


gaborznagy:
If a log path with a source with `flags(no-parse)` and a `syslog-parser()` receives a message with an invalid message format, then syslog-ng crashes.

Based on backtrace crash caused by an assert() in `nv_table_clear()` when syslog parsing in `syslog-parser()` fails and syslog wants to inject an error message into the log message with `msg_format_inject_parse_error`.
In `nv_table_clear` there is an assertion on the NVTable's `ref_cnt`, which had been set to 2 by `log_parser_process_message()`.
`nv_table_clear` is ONLY called if `LF_STATE_OWN_PAYLOAD` flag of the LogMessage is set, which could be done by the source when it sets the PROGRAM, SOURCE, HOST, or MSG part of the message.

## Version of syslog-ng
3.15.1

## Backtrace
```
  #0  0x00007f37bd9db1f7 in raise () from /lib64/libc.so.6
  #1  0x00007f37bd9dc8e8 in abort () from /lib64/libc.so.6
  #2  0x00007f37bea6800a in g_assertion_message () from /opt/syslog-ng/lib/libglib-2.0.so.0
  #3  0x00007f37bea6806f in g_assertion_message_expr () from /opt/syslog-ng/lib/libglib-2.0.so.0
  #4  0x00007f37bf5d79e0 in nv_table_clear (self=<optimized out>) at lib/logmsg/nvtable.c:641
  #5  0x00007f37bf5d23a3 in log_msg_clear (self=0x7f37ac0024d0) at lib/logmsg/logmsg.c:1101
  #6  0x00007f37bf5a5f6f in msg_format_inject_parse_error (msg=0x7857, data=0x78b8 <Address 0x78b8 out of bounds>, length=6, problem_position=-1) at lib/msg-format.c:35
  #7  0x00007f37b8c3fdb3 in syslog_format_handler (parse_options=0x20bd030, data=0x7f37ac0026f9 "1       LPU", ' ' <repeats 11 times>, "8% 13%  524MB/3775MB", length=42, self=0x7f37ac0024d0) at modules/syslogformat/syslog-format.c:1290
[...]

PR #2128, opened 2018-06-20
Fix possible crash when using log_msg_clear()

gaborznagy:
log_msg_clear is used when we want to reuse a log message, but
we don't need its contents (SDATA, tags, NVTable), just
its properties (log path related properties, ack, etc.).
It's used only in msg-format handlers (syslog, kmsg).

This function had a condition which decided if we are
allowed to modify the NVTable or we should create a new one
(protecting other users of the log message).
The condition was not perfect and it can result
in an assertion failure when nv_table_clear called (NVTable ref_cnt), as
in case of parsers it happened.

Instead of improving the condition, which could only be achieved
with a layering violation (checking the NVTable ref_cnt),
it was decided that we should always use a new NVTable 
and free the previous one if it is not used by anyone.

Removing this conditional nv_table_clear results
in no performance degradation in the fast path,
this is an error handling.

As nv_table_clear is not used anywhere else I have removed it.

Fixes: #2078


kira-syslogng:
Build SUCCESS

bazsi:
Well done, ok on my part.

gaborznagy:
I would like to investigate something that came up today. Please dont merhe it yet

gaborznagy:
I've changed the nv_table_unref() to do it conditionally to prevent freeing a cloned messages NVTable.
With prev. patch only the `borrowed` flag saved it from being freed and it had a ref_cnt = 0.

kira-syslogng:
Build SUCCESS

PR #2129, opened 2018-06-20
logwriter: disable ack for mark mode

furiel:
Mark mode messages need not be acked. Without this patch, protos can
crash that do not immediately acknowledge back messages.

### DIFF
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -774,6 +774,7 @@ log_writer_mark_timeout(void *cookie)
   main_loop_assert_main_thread();
 
   msg = log_msg_new_mark();
+  path_options.ack_needed = FALSE;
   /* timeout: there was no new message on the writer or it is in periodical mode */
   hostname = resolve_sockaddr_to_hostname(&hostname_len, msg->saddr, &self->options->host_resolve_options);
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
I just had a look on the `mark-mode(internal)` (`afinter` side):

https://github.com/balabit/syslog-ng/blob/35d336db674c8bcc189a8a99b5b4972ebc53c906/lib/afinter.c#L142-L145

`ack_needed` is already disabled here. At first glance, it seemed to be incorrect, but here is the trick:
`log_pipe_queue` is used to forward the message instead of `log_source_post`, so the window won't be decremented -> ack is not needed here too.

bazsi:
log_source_post() was probably a newer addition than posting of the mark
message. And probably it could be refactored to use log_source_post()


-- 
Bazsi

lbudai:
@bazsi: agree with you, but I’d merge this as-is (because it solves a critical (crash) for us). we are going to create a task for the refactor.

bazsi:
Agreed. I just mentioned why that code wasn't using post ()

PR #2130, opened 2018-06-21
test_thread_wakeup: activate test

furiel:
This test was not compiled and executed. This patch fixes this problem.


kira-syslogng:
Build SUCCESS

PR #2132, opened 2018-06-21
RewriteExpr: Remove misleading print of result from log_rewrite_queue()

mitzkia:
- this can be misleading when rewrite rule is used without condition,
for example with: set()

### DIFF
--- a/lib/rewrite/rewrite-expr.c
+++ b/lib/rewrite/rewrite-expr.c
@@ -26,12 +26,6 @@
 
 /* rewrite */
 
-static EVTTAG *
-rewrite_result_tag(gboolean res)
-{
-  return evt_tag_str("result", res ? "MATCH - Forwarding message" : "UNMATCHED - Forwarding message");
-}
-
 void
 log_rewrite_set_condition(LogRewrite *self, FilterExprNode *condition)
 {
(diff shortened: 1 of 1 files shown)


mitzkia:
The evaluated "res" could not be changed in else{ self->process() } branch.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/rewrite/rewrite-expr.c
```diff
@@ -58,7 +52,6 @@ log_rewrite_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_option
       self->process(self, &msg, path_options);
     }
   msg_debug("<<<<<< rewrite rule evaluation result",
```
furiel: As the result is not available anymore, you can delete the result word from here.

furiel:
@mitzkia in case condition is used, the value of res would be still useful, isn't it? How about conditionally emit it?

For example a dirty way that exploits that the final parameter in varargs can/must be null
```
  msg_debug("<<<<<< rewrite rule evaluation result",
            evt_tag_str("rule", self->name),
            log_pipe_location_tag(s),
            evt_tag_printf("msg", "%p", msg),
            self->condition ? rewrite_result_tag(res) : NULL);
```

mitzkia:
@furiel Thanks for the review. I like the conditionally emit idea. 
@gaborznagy What do you think? Would it be useful in this way?

gaborznagy:
@mitzkia: I agree with furiel's proposal.
Although I don't know if the previous msg_debug call would be redundant or not:
`msg_debug("Rewrite condition unmatched, skipping rewrite",`

furiel:
@gaborznagy is right, res + the msg_debug above are redundant, one can do the math without either one. With that, the original version is better when deleting `rewrite_result_tag(res)`. Please rephase `<<<<<< rewrite rule evaluation result`, then approve from my side.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2134, opened 2018-06-22
cisco: add parsing of Cisco unified call manager

bazsi:
Jun 14 11:57:27 PM.685 UTC :  %UC_LOGIN-4-AuthenticationFailed: %[Login Date/Time=06/15/2018 at 09:57:27][Login IP Address/Hostname=10.25.1.16][Login Interface=cucm-uds][Login UserID=POBAR][App ID=Cisco Tomcat][Cluster ID=][Node ID=cucmsub-prd-t2]: Login Authentication failed.

### DIFF
--- a/scl/cisco/plugin.conf
+++ b/scl/cisco/plugin.conf
@@ -60,10 +60,6 @@ block parser cisco-timestamp-parser(template()) {
             filter { match('^[.*]?([A-Za-z]{3} [0-9 ]\d \d{4} \d{2}:\d{2}:\d{2})' value('1') flags(store-matches)); };
             parser { date-parser(format('%b %d %Y %H:%M:%S') template("$1")); };
         };
-
-        rewrite {
-            unset(value('1'));
-        };
     };
 };
 
@@ -96,11 +92,6 @@ block parser cisco-parser(prefix(".cisco.")) {
             filter { match("^(?'HOST'[^:]+)$" value('1') flags(store-matches) type(pcre)); };
         } else {
             filter { match("^$" value('1') flags(store-matches) type(pcre)); };
-        };
-        rewrite {
-            unset(value("1"));
-            unset(value("2"));
-            unset(value("3"));
         };
     };
 };
--- a/scl/cisco/plugin.conf
+++ b/scl/cisco/plugin.conf
@@ -45,9 +45,18 @@ block parser cisco-timestamp-parser(template()) {
         };
 
         if {
+            # timestamp from Cisco Unified Call Manager, example "Jun 14 11:57:27 PM.685 UTC"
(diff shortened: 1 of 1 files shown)


scl/cisco/plugin.conf
```diff
@@ -45,6 +45,10 @@ block parser cisco-timestamp-parser(template()) {
         };
 
         if {
+	    # drops msecs and timezone
+            filter { match('^[.*]?([A-Za-z]{3} [0-9 ]\d \d{2}:\d{2}:\d{2} ((AM)|(PM)))' value('1') flags(store-matches)); };
```
MrAnno: I think it's worth adding an example at the top of the config.
`Jun 14 11:57:27 PM.685 UTC`

kira-syslogng:
Build SUCCESS

scl/cisco/plugin.conf
bazsi: good idea, done.

scl/cisco/plugin.conf
```diff
@@ -45,9 +45,18 @@ block parser cisco-timestamp-parser(template()) {
         };
 
         if {
+            # timestamp from Cisco Unified Call Manager, example "Jun 14 11:57:27 PM.685 UTC"
+            # NOTE: drops msecs and timezone
+            filter { match('^[.*]?([A-Za-z]{3} [0-9 ]\d \d{2}:\d{2}:\d{2} ((AM)|(PM)))' value('1') flags(store-matches)); };
```
gaborznagy: Do we need to put AM and PM into capture groups? Isn't (AM|PM) enough?
Also should we unset additional capture groups, e.g. $4 as well?
Testing with the input message we will have these nv-pairs as well:
"4":"PM", 
"0":"Jun 14 11:57:27 PM"

kira-syslogng:
Build SUCCESS

scl/cisco/plugin.conf
bazsi: With the already merged ewmm() driver that excludes all numeric matches I think we can get rid of all unsets() everywhere. That's fragile anyway (e.g I broke it again).

bazsi: done in the recent push.

gaborznagy: That functionality is in format-ewmm template function, right?
Which is used in syslog-ng() destination driver currently.

What if somebody wants to use only cisco-parser()? Should cisco-parser give you these temporary variables?
Or I change the question, is it a problem for the user if cisco-parser unsets these matches?

I know that name-value pairs are not transferred automatically, it's a sort of reference question for developers how to write processing elements.

kira-syslogng:
Build SUCCESS

PR #2135, opened 2018-06-22
Http batching

bazsi:


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
I have now removed the batch-prefix/suffix options in favour of putting those in a separate branch. This now focuses clearly on the batched use-cases.

kira-syslogng:
Build FAILURE

bazsi:
With the merge of #2063 this should be unblocked. I've now rebased to master and hope everything will magically turn green :)

kira-syslogng:
Build FAILURE

bazsi:
kira fails because I have removed a couple of redundant log messages which is now breaking the tests, as they rely on the presence of those log messages. I'll see what I can do with that.

bazsi:
I've updated the kira tests here:
https://git.balabit/syslog-ng/syslog-ng-pe-testdb--mainline--5.0/merge_requests/37

@kira-syslogng retest this please test branch=http-batching; test=functions/destination-drivers/http-destination-driver007-http-retries-option.zts/;

kira-syslogng:
Build SUCCESS

bazsi:
The patch to kira testdb seems to have resolved the issue. @mitzkia can you please review the testdb patch? Also, once this is green, we should synchronize merging the testdb and the code bits.

mitzkia:
@bazsi testdb patch is approved.

modules/http/http.c
```diff
@@ -491,6 +223,324 @@ http_dd_set_timeout(LogDriver *d, glong timeout)
...
+      msg_error("curl: error sending HTTP request",
+                evt_tag_str("error", curl_easy_strerror(ret)),
+                log_pipe_location_tag(&self->super.super.super.super));
+      retval = WORKER_INSERT_RESULT_NOT_CONNECTED;
+      goto exit;
+    }
+
+  glong http_code = 0;
+
+  curl_easy_getinfo(self->curl, CURLINFO_RESPONSE_CODE, &http_code);
```
Kokan: According to the documentation the `http_code` is filled only if the function returns with `CURLE_OK`, and the current code initialize `http_code` to `0`, and later divides it with `100` and if it is not `4` or `5` it is success.
It could report unsuccessful calls as successful one. (I know it has a low chance, as this is just getting the information, but still.)

modules/http/http.c
```diff
@@ -491,6 +223,324 @@ http_dd_set_timeout(LogDriver *d, glong timeout)
...
+    }
+}
+
+static worker_insert_result_t
+_map_http_status_to_worker_status(glong http_code)
+{
+  worker_insert_result_t retval;
+
+  switch (http_code/100)
+    {
```
Kokan: The `0` is invalid http response code, I would not handle it as successful case (and now it possible to get `0` see my comment below.)

modules/http/http.c
```diff
@@ -428,38 +457,77 @@ _map_http_status_to_worker_status(glong http_code)
...
+  _add_message_to_batch(self, msg);
+  return _flush(&self->super);
+}
+
+static worker_insert_result_t
+_insert(LogThreadedDestDriver *s, LogMessage *msg)
+{
+  HTTPDestinationDriver *self = (HTTPDestinationDriver *) s;
+
+  if (self->flush_lines > 0)
```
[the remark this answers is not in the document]
bazsi: Thanks, fixed.

bazsi:
@kira-syslogng retest this please test branch=http-batching;

kira-syslogng:
Build FAILURE

bazsi:
@kira-syslogng retest this please test branch=http-batching;

kira-syslogng:
Build SUCCESS

modules/http/http.c
```diff
@@ -223,6 +223,22 @@ http_dd_set_timeout(LogDriver *d, glong timeout)
   self->timeout = timeout;
 }
 
+void
+http_dd_set_flush_lines(LogDriver *d, glong flush_lines)
```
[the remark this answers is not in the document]
bazsi: I've addressed this in 882d032510b4d8368e0b16c42bff7a2d2d566597

With that, could you please approve the patchset? The rest of the changes requested were added to a subsequent PR, as those would cause massive conflicts. Thanks.

bazsi:
This should be ready for merging, once 3.17.1 has been cut.

modules/http/http.c
```diff
@@ -491,6 +223,324 @@ http_dd_set_timeout(LogDriver *d, glong timeout)
...
+static struct curl_slist *
+_format_request_headers(HTTPDestinationDriver *self, LogMessage *msg)
+{
+  struct curl_slist *headers = NULL;
+  GList *l;
+
+  headers = _add_header(headers, "Expect", "");
+  if (msg)
+    {
+      /* NOTE: I have my doubts that these headers make sense at all.  None of
```
furiel: Optional request:
Good topic, thanks for opening it! Maybe these can be removed, especially if we can provide alternative? For example if headers could handle templates? For now they don't.
Still I would **not** commit this comment into the code, but move the discussion into an issue or PR. Because discussion inside the code makes participating hard. As this would require participants to open a pr to add a comment.

modules/http/http.c
```diff
@@ -491,6 +223,324 @@ http_dd_set_timeout(LogDriver *d, glong timeout)
...
+
+  if (self->ciphers)
+    curl_easy_setopt(self->curl, CURLOPT_SSL_CIPHER_LIST, self->ciphers);
+
+  curl_easy_setopt(self->curl, CURLOPT_SSLVERSION, self->ssl_version);
+  curl_easy_setopt(self->curl, CURLOPT_SSL_VERIFYHOST, self->peer_verify ? 2L : 0L);
+  curl_easy_setopt(self->curl, CURLOPT_SSL_VERIFYPEER, self->peer_verify ? 1L : 0L);
+
+  curl_easy_setopt(self->curl, CURLOPT_DEBUGFUNCTION, _curl_debug_function);
+  curl_easy_setopt(self->curl, CURLOPT_VERBOSE, (long) (debug_flag || trace_flag));
```
furiel: Regression.
debug_flag and trace_flag can be changed runtime with syslog-ng-ctl, without reload. With this patch, the http loglevel cannot follow such change. That's why we explicitely set verbose level to one in the original version, and have an early return based on debug_flag in _curl_debug_function.

furiel:
@bazsi 7fd860d does not work as intended: curl logs are written by msg_debug, so trace level would not affect behaviour. That patch also causes a minor regression I described earlier,  so I added a small revert patch to the top, because I did not want to hold this PR on such small issue. Lets discuss the goal of that patch, and handle in a separate PR.

furiel:
@kira-syslogng retest this please test branch=http-batching;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=http-batching;

kira-syslogng:
Build SUCCESS

lbudai:
@bazsi: could you resolve the conflict?
After that we would merge this.

bazsi:
Conflict resolved. Light manual end-to-end testing still shows that its operational.

So let's hope Travis/kira agrees. :)

modules/http/http.c
bazsi: While I agree that work items would need to be scheduled/filed, I think notes like this help navigation and subsequent adjustment of functionality for someone coming later to read the code.

For example, at the end I could easily retain this functionality, but only in case batching is disabled. With that I dont see removing that code is priority. Should we need to change the http module in the future, this comment helps to add a +1 to cases where this functionality becomes a bottleneck for further improvements, bumping the motivation to actually remove it. Adding a low prio ticket that will never get scheduled will not achieve the same, as it is disconnected from the code. We could add the ticket and a reference from the code, which is a bit of an improvement, I tend to agree. I was lazy not filing, but I dont think the ticket would help that much.

bazsi:
@furiel thanks foe the thorough review and comments, and the patch to fix the curl debug functionality. I accept the reason why it is always set, but I think it would be a bit better if we only set the callback for new requests and not always, as i am sure those debug callbacks have an overhead. Anyway, http is probably not the best performing transport, so that overhead is negligible, so it's fine as it is. Thanks again

kira-syslogng:
Build FAILURE

bazsi:
@kira-syslogng retest this please test branch=http-batching;

kira-syslogng:
Build SUCCESS

bazsi:
Finally, this is green. \o/

PR #2136, opened 2018-06-22
http: enable supporting HTTP redirects

bazsi:
Redirects can be useful in load-balanced environments, so enable supporting
them. We would only follow maximum 3 redirects and only if they are
http or https based.


furiel:
It would be useful if users could configure these parameters. Some administrators might not want to follow redirections, others might want to increase the limit beyond 3. We do not need separate parameter for follow_location and max_redirs: max_redirs could express both. -1 would mean follow_location+infinite max-redirs (this is what curl does), 0 would disable follow_location, and any positive integer would enable follow_location with setting max_redirs to the same value.

As for default values, follow location on+ max redirs 3 seems fine.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
@furiel I would wait until someone presents a use-case where configuring this makes sense. Supporting redirects does make sense, in the case of a server being brought down, but I am not expecting more than 1 single redirects. And keeping servers in production which does redirect reduces performance a lot.

It might make sense to turn redirects on or off, but again I don't see this would hurt and I don't see too much value in making this configurable. If such a use-case presents itself, it's just as easy to make this configurable at that point.

I have rebased this now to the latest version of http-batching that should resolve the conflicts.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

furiel:
It seems this PR has conflicts with master. @bazsi can you fix it?

bazsi:
Not until a few of my other branches get integrated. This would cause a
huge conflict.

gaborznagy:
@bazsi: As far as I see the only depending http related branch is the "flush-timeout" branch with PR #2251, right

bazsi:
And one I have privately on my notebook, which I promise would push, once I
am done with the monster conflict I inflicted on myself.

bazsi:
Just rebased it on top of current master.

kira-syslogng:
Build FAILURE

PR #2137, opened 2018-06-22
Fix CMake unit test compilation (no-pie)

MrAnno:
This PR contains a workaround for a bug in the implementation of the `POSITION_INDEPENDENT_CODE` property (CMake).

https://gitlab.kitware.com/cmake/cmake/issues/16561

### DIFF
--- a/cmake/add_tests.cmake
+++ b/cmake/add_tests.cmake
@@ -40,7 +40,7 @@ function (add_unit_test)
   target_compile_definitions(${ADD_UNIT_TEST_TARGET} PRIVATE TOP_SRCDIR="${PROJECT_SOURCE_DIR}")
   target_link_libraries(${ADD_UNIT_TEST_TARGET} ${ADD_UNIT_TEST_DEPENDS} syslog-ng)
   target_include_directories(${ADD_UNIT_TEST_TARGET} PUBLIC ${ADD_UNIT_TEST_INCLUDES})
-  if (NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
+  if (NOT APPLE)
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

cmake/add_tests.cmake
```diff
@@ -40,19 +40,24 @@ function (add_unit_test)
...
     set_property(TARGET ${ADD_UNIT_TEST_TARGET} PROPERTY POSITION_INDEPENDENT_CODE FALSE)
 
-    if (${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
-      # https://gitlab.kitware.com/cmake/cmake/issues/16561
-      set_property(TARGET ${ADD_UNIT_TEST_TARGET} APPEND PROPERTY LINK_FLAGS "-Wl,-no_pie")
+    # https://gitlab.kitware.com/cmake/cmake/issues/16561
+    include(CheckCCompilerFlag)
+    check_c_compiler_flag(-no-pie NO_PIE_AVAILABLE)
+    if (NO_PIE_AVAILABLE)
+      set_property(TARGET ${ADD_UNIT_TEST_TARGET} APPEND_STRING PROPERTY LINK_FLAGS " -no-pie")
```
furiel: So in summary `set_property(TARGET ${ADD_UNIT_TEST_TARGET} PROPERTY POSITION_INDEPENDENT_CODE FALSE)` does not work with mac and gcc v>=6, hence the workaround. Does the POSITION_INDEPENDENT_CODE give anything after this workaround, or we could just delete that line?

MrAnno: We could remove that line, it does not give anything to the tests.
However, the code below is just a workaround that we'll remove after a "few" years (:smiley:), so I would leave `POSITION_INDEPENDENT_CODE` here.

furiel: Ok fine for me.

PR #2140, opened 2018-06-25
logqueue-fifo: memory_usage - possible underflow (afsql)

furiel:
In the original code, the content of backlog is dropped earlier, then
the empty list is iterated through to update memory_usage
counter. This means when messages put back to backlog later again,
the size was subtracted twice, hence the underflow.

This patch first updates statistics, then drops the content of backlog, fixing the problem.

The only place when this code path is walked in ose is in afsql, where I could reproduce the problem:
created an sql destination with type(sqlite3), setting flush-lines(1000). After starting to send logs, I deleted the sql database file. This triggered an underflow in memory_usage. After this patch, the memory_usage counter goes back to zero, when the load stops.

### DIFF
--- a/tests/unit/test_logqueue.c
+++ b/tests/unit/test_logqueue.c
@@ -284,3 +284,28 @@ Test(logqueue, test_with_threads)
     }
   fprintf(stderr, "Feed speed: %.2lf\n", (double) TEST_RUNS * MESSAGES_SUM * 1000000 / sum_time);
 }
+
+Test(logqueue, log_queue_fifo_rewind_all_and_memory_usage)
+{
+  LogQueue *q = log_queue_fifo_new(OVERFLOW_SIZE, NULL);
+  log_queue_set_use_backlog(q, TRUE);
+
+  StatsClusterKey sc_key;
+  stats_lock();
+  stats_cluster_logpipe_key_set(&sc_key, SCS_DESTINATION, q->persist_name, NULL );
+  stats_register_counter(1, &sc_key, SC_TYPE_MEMORY_USAGE, &q->memory_usage);
+  stats_unlock();
+
+  feed_some_messages(q, 1, &parse_options);
+  gint size_when_single_msg = stats_counter_get(q->memory_usage);
+
+  feed_some_messages(q, 9, &parse_options);
+  cr_assert_eq(stats_counter_get(q->memory_usage), 10*size_when_single_msg);
+
+  send_some_messages(q, 10);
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng do perftest

kira-syslogng:
Build SUCCESS

tests/unit/test_logqueue.c
```diff
@@ -284,3 +284,28 @@ Test(logqueue, test_with_threads)
...
+  feed_some_messages(q, 1, &parse_options);
+  gint size_when_single_msg = stats_counter_get(q->memory_usage);
+
+  feed_some_messages(q, 9, &parse_options);
+  cr_assert_eq(stats_counter_get(q->memory_usage), 10*size_when_single_msg);
+
+  send_some_messages(q, 10);
+  cr_assert_eq(stats_counter_get(q->memory_usage), 0);
+  log_queue_rewind_backlog_all(q);
+  cr_assert_eq(stats_counter_get(q->memory_usage), 10*size_when_single_msg);
```
bazsi: hmm... the backlog is not accounted in the memory usage counter? That is also stored in memory, isn't it?

furiel: No it is not accounted. Maybe not optimal, but I try to reason why I implemented memory_usage like this.

When we tried to find a definition for memory_usage, we did not want to include all memory usage of the drivers, either because it is hard to estimate (how much memory does any language bindings use?), either not interesting from scaling point of view. The intended use for memory usage is capacity calculation and scaling, and in case of problem, the memory usage of a queue would dominate the other memory allocations. So we restricted ourselves to memory usage of the queues.

Within the memory usage of queues, the auxiliary allocations are not interesting again, so we wanted to focus on the allocations per message. But the concept is still vague, so I chose the definition to be as simplest as possible: when a message is pushed to the queue, the memory usage should be increased with the size of the message, and decreased with the size of the message when a message is popped.

When a message is popped, it is like the queue has passed the message to the destination driver. In case a destination needs a backlog, the message goes indeed to the backlog. But it is an interesting coupling why the backlog is part of the logqueue. From conceptual point of view, the queue does not own the message after pop. Hence it is not strictly part of memory_usage.

The behaviour is the same in self->super.queued_messages counter: it is decreased after each pop, even if the message is added to the backlog.

bazsi:
Thanks for the explanation, makes sense.

lbudai:
Could someone review&approve(...and merge) this PR?

PR #2143, opened 2018-06-26
All drivers should support inner dst or src plugins

bazsi:
This branch fixes a couple of cases where the source_driver_option or dest_driver_option rules were not
referenced.
 
It also fixes scl source/destination drivers to include __VARARGS__.

Both of these changes support the ability to use inner source/dest plugins such as hook-commands() in all drivers.


kira-syslogng:
Build SUCCESS

PR #2145, opened 2018-06-26
dns: Inform about the right dns-cache() configuration

bjacquin:
Since bad916a9e6c5 ("Enforce use-dns-cache(no) when use-dns(no) is
set"), the emitted warning does not refer to an existing setting


kira-syslogng:

kira-syslogng:
Build SUCCESS

PR #2146, opened 2018-06-27
afinter: adjusting window size for internal mark mode

furiel:
Prior this patch, internal mark mode messages were generated independently
from flow control. This means mark mode messages were sent even if the
destination cannot handle messages.

With the patch, mark mode messages can fill now the available window,
as such, internal source can be suspended properly because of them,
and no mark mode messages are generated in that case.

How to test?

Leave syslog-ng running with -Fevd without a destination. You will see that around 100 mark mode messages, the mark mode generation stops. Without this patch, mark mode messages are generated infinitely.
```
@version: 3.16

options {
  time-reopen(1);
  mark-freq(1);
};

log {
  source { internal(); };
  source { network(port(4444) transport(tcp)); };
  destination { network(localhost port(5555) transport(tcp) mark-mode(internal)); };
  flags(flow-control);
};
```

Motivated by the discussion of @bazsi @MrAnno @lbudai in
https://github.com/balabit/syslog-ng/pull/2129

### DIFF
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -100,6 +100,7 @@ struct _AFInterSource
 };
 
 static void afinter_source_update_watches(AFInterSource *self);
+void afinter_message_posted(LogMessage *msg);
 
 static void
 afinter_source_post(gpointer s)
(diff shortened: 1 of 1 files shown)


bazsi:
looks good to me.

kira-syslogng:
Build SUCCESS

PR #2147, opened 2018-06-28
pragma: lower the require message level to debug

### LINKED ISSUE #2138: Telegram requires http module but doesn't fail build when it's not built
# syslog-ng

## Version of syslog-ng
3.16.1
```
## Platform
AWS Linux

When starting syslog-ng 3.16.1 I see this message
Starting /opt/syslog-ng/sbin/syslog-ng: [2018-06-23T16:08:15.956848] Plugin module not found in 'module-path'; module-path='/opt/syslog-ng/lib/syslog-ng', module='http'
[2018-06-23T16:08:15.956883] Included file was skipped because of a missing module; module='http', location='/opt/syslog-ng/share/syslog-ng/include/scl/telegram/telegram.conf:23:11'

I'm using the following configure:
./configure --prefix=/opt/syslog-ng --enable-json --enable-java --enable-java-modules --enable-python --with-python=2.7

I tracked it down to the telegram module is being built by default however the http module is not being built becuase i haven't installed the libcurl-devel packages. If i install the package and then build i'm able to start syslog ng with out the above message.

When i specifically say build http without the libcurl-devel then the configure fails.

The problem looks like telegram may not be forcing http to be built and thus my compile and install succeeds. 

Since telegram is requiring the http module it should be forcing it to be enabled so if it can't be built then the build fails rather than continuing.
The other possibility is that it's not built at all if http can't be built.

Kokan:
The `@requires' keyword wrote a warning, when the component was missing. 
It is probably too verbose, this PR moves that warning message into the debug level.

Fixes #2138

### DIFF
--- a/lib/pragma-grammar.ym
+++ b/lib/pragma-grammar.ym
@@ -135,9 +135,9 @@ requires_stmt
                   }
                 else
                   {
-                    msg_warning("Included file was skipped because of a missing module",
-                                evt_tag_str("module", $2),
-                                cfg_lexer_format_location_tag(lexer,&@1));
+                    msg_debug("Included file was skipped because of a missing module",
+                              evt_tag_str("module", $2),
+                              cfg_lexer_format_location_tag(lexer,&@1));
                     cfg_lexer_start_next_include(lexer);
                   }
               }


kira-syslogng:
Build SUCCESS

PR #2153, opened 2018-06-30
DiskQDestPlugin was not freed

Kokan:
This fixes two leaks connected to the `DiskQDestPlugin`.

The first could be triggered during configuration parsing, if the parsing was not successful the `cleanup` function were `log_pipe_unref`, but the `DiskQDestPlugin` has no relationship with the `LogPipe`. It was at least a leak, or arbitrary code execution from random memory.

The second one was easier, the `_free` function of `DiskQDestPlugin` did not called `g_free` on itself.

[edit]
Also the same have been done for hook commands.

### DIFF
--- a/modules/diskq/diskq-parser.c
+++ b/modules/diskq/diskq-parser.c
@@ -48,7 +48,7 @@ CfgParser diskq_parser =
   .name = "disk_buffer",
   .keywords = diskq_keywords,
   .parse = (int (*)(CfgLexer *lexer, gpointer *instance, gpointer arg)) diskq_parse,
-  .cleanup = (void (*)(gpointer)) log_pipe_unref,
+  .cleanup = (void (*)(gpointer)) log_driver_plugin_free,
 
 };
 
--- a/modules/diskq/diskq.c
+++ b/modules/diskq/diskq.c
@@ -163,6 +163,10 @@ _free(LogDriverPlugin *s)
   disk_queue_options_destroy(&self->options);
 }
 
+DiskQueueOptions *diskq_get_options(DiskQDestPlugin *self)
+{
+  return &self->options;
+}
 
 DiskQDestPlugin *
 diskq_dest_plugin_new(void)
--- a/modules/diskq/diskq.c
+++ b/modules/diskq/diskq.c
@@ -32,6 +32,12 @@
 
 #define DISKQ_PLUGIN_NAME "diskq"
 
+struct _DiskQDestPlugin
+{
+  LogDriverPlugin super;
+  DiskQueueOptions options;
+};
+
 static gboolean
 log_queue_disk_is_file_in_directory(const gchar *file, const gchar *directory)
 {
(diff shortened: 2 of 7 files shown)


modules/diskq/diskq-parser.c
```diff
@@ -48,7 +48,7 @@ CfgParser diskq_parser =
   .name = "disk_buffer",
   .keywords = diskq_keywords,
   .parse = (int (*)(CfgLexer *lexer, gpointer *instance, gpointer arg)) diskq_parse,
-  .cleanup = (void (*)(gpointer)) log_pipe_unref,
+  .cleanup = (void (*)(gpointer)) log_driver_plugin_free,
```
MrAnno: The same is true for `hook-commands`, we should fix that as well.

modules/diskq/diskq.c
```diff
@@ -161,10 +167,16 @@ _free(LogDriverPlugin *s)
 {
   DiskQDestPlugin *self = (DiskQDestPlugin *)s;
   disk_queue_options_destroy(&self->options);
+  g_free(self);
```
MrAnno: `log_driver_plugin_free_method()` would be better instead of `g_free()`.

kira-syslogng:
Build SUCCESS

modules/diskq/diskq-parser.c
Kokan: Done.

MrAnno: Thanks.

modules/hook-commands/hook-commands.c
```diff
@@ -27,30 +27,45 @@
...
+  HookCommandsPlugin *self = (HookCommandsPlugin *)s;
   g_free(self->teardown);
   self->teardown = g_strdup(teardown);
 }
 
 void
-hook_commands_plugin_set_shutdown(HookCommandsPlugin *self, const gchar *shutdown)
+hook_commands_plugin_set_shutdown(LogDriverPlugin *s, const gchar *shutdown)
 {
+  HookCommandsPlugin *self = (HookCommandsPlugin *)s;
```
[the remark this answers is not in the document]
Kokan: In the grammar, where it is called, there is no reference to `HookCommandsPlugin`, but only to `LogDriverPlugin`, as the new function should return with the base class.

So either in the grammar the `LogDriverPlugin` should be casted to `HookCommandsPlugin` or in the setters functions.
I would like to keep the grammar clean, so stuff like casting should not be part of grammar if possible.

Why casting multiple time is a disadvantage ?
* The source code gets bigger, but I think it is not a concern in this case.
* Run time it is free(, and even if it would not be it is configuration time.)
* Compile time is probably bigger, but still not a concern I guess.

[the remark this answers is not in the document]
MrAnno: I mean `last_hook_commands` with type `HookCommandsPlugin` was in the grammar. With that, we don't have to cast anything at all.

There was only one cast in the grammar:
```
*instance = (LogDriverPlugin *) last_hook_commands;
```
But this is an "upcast", which is much acceptable in my opinion (and we have only one instead of N).

kira-syslogng:
Build SUCCESS

PR #2154, opened 2018-07-01
undefined behavior fix

Kokan:
* In the `cfg_tree_compile_junction` function the `fork_mpx` unconditionally dereferenced, and it can be a null pointer.
* When the `LogMessage` is deserialized and the `sdata` does not have data (`sdata==NULL`), it tries to read zero byte into a zero length buffer (null pointer).  Reading zero was a valid operation before this PR, it just make sure, that the read function does not touch anything as they should not.

Question:
* Is it okay to move the `outer_pipe_head` initialization ?

### DIFF
--- a/lib/cfg-tree.c
+++ b/lib/cfg-tree.c
@@ -1056,6 +1056,7 @@ cfg_tree_compile_junction(CfgTree *self,
           if (!fork_mpx)
             {
               fork_mpx = cfg_tree_new_mpx(self, node);
+              *outer_pipe_head = &fork_mpx->super;
             }
           log_multiplexer_add_next_hop(fork_mpx, sub_pipe_head);
         }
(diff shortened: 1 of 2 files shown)


lib/cfg-tree.c
```diff
@@ -1082,7 +1083,6 @@ cfg_tree_compile_junction(CfgTree *self,
         }
     }
 
-  *outer_pipe_head = &fork_mpx->super;
```
szemere: Discussed IRL. We fortunately get away with the NULL pointer dereferation, and set `outer_pipe_head` to NULL in those cases, due to a compiler optimization. (Also checked in assembly.)
Also checked every calling point of `cfg_tree_compile_junction` since now we "lost the initialization" of `outer_pipe_head`.

PR #2158, opened 2018-07-02
[WIP] Increase the default value of log-fetch-limit() to 100

MrAnno:


MrAnno:
@kira-syslogng do perftest

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

MrAnno:
@pzoleex Help me please, what happened here? :smiley:

pzoleex:
Memory usage was doubled. I don't know if increasing log fetch limit can cause such a high memory usage, we should investigate if this is normal or a possible leak.

lbudai:
@MrAnno: why would we need to increase this limit?

MrAnno:
We've measured previously that this increases the performance of the wildcard file source, so an internal issue has been created to change the default.

MrAnno:
Just a note:
The documentation is currently incorrect: it says the default is already 100.

MrAnno:
Outdated.

PR #2160, opened 2018-07-03
regression - scl: aligning java related scl-s with mandatory parameters

furiel:
The regression introduced in https://github.com/balabit/syslog-ng/pull/2088. This PR fixes the remaining affected scls.


kira-syslogng:
Build SUCCESS

PR #2161, opened 2018-07-03
[WIP]Log source max memory poc

lbudai:
First PR of the memory-limited flow control concept.
I'd discuss this topic IRL on tomorrow.

The concept:
 * `max-memory` could be defined for sources (also as a global option, which means that each source has the same memory limit, when defined per source , it can express a sort of priority between the sources)
 * when `max-memory` is used the fifo queues would become dynamic, also the window size calculation
Note: `max-memory` is calculated _after_ the message went through the processing pipeline, so the final size is known, this is not an exact number,  cloned messages are not counted... 

This PR just adds the memory-limitation for sources, without touching the destination side and iw-size calculations.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -53,6 +53,7 @@ set (LIB_HEADERS
     alarms.h
     apphook.h
     atomic.h
+    atomic-gssize.h
     block-ref-parser.h
     cache.h
     cfg.h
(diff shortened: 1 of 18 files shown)


kira-syslogng:
Build SUCCESS

lbudai:
I'll open a final PR when finished.

PR #2162, opened 2018-07-04
Fix missing normalize flags

furiel:
Release notes:
- syslog-ng-ctl: fixing a memleak
- afsql: normalizing key when checking flags
This can be useful on a rare occasion when user tries to use `dont_create-tables` or dont-create_tables`.
- Normalizing flags in log writer options. This can be useful on a rare occasion when user tries to use `no-multi_line` or `no_multi-line`.
- csvparser: normalizing flags was not supported earlier at all.

Developer notes:
- new token added: normalized_flag that can be used in grammar to replace underscores with hyphens.
- log_flags: just a refactor to use the new normalized_flag token

### DIFF
--- a/lib/str-utils.c
+++ b/lib/str-utils.c
@@ -55,6 +55,13 @@ str_replace_char(const gchar *str, const gchar from, const gchar to)
   return ret;
 }
 
+/*
+  This function normalizes differently than the flags version. TODO:
+  investigate if the two can be aligned. Hint: as flag normalization has
+  strong position in the code, aligning block normalization to flag
+  normalization might prove easier. Some more info:
+  https://github.com/balabit/syslog-ng/pull/2162#discussion_r202247468
+*/
 gchar *
 __normalize_key(const gchar *buffer)
 {
--- a/lib/str-utils.c
+++ b/lib/str-utils.c
@@ -60,3 +60,9 @@ __normalize_key(const gchar *buffer)
 {
   return str_replace_char(buffer, '-', '_');
 }
+
+gchar *
+normalize_flag(const gchar *buffer)
+{
+  return str_replace_char(buffer, '_', '-');
+}
--- a/modules/afsql/afsql.c
+++ b/modules/afsql/afsql.c
@@ -33,6 +33,7 @@
 #include "apphook.h"
 #include "timeutils.h"
 #include "mainloop-worker.h"
+#include "str-utils.h"
 
 #include <string.h>
 #include <errno.h>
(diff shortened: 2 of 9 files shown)


modules/afsql/afsql.c
```diff
@@ -1478,12 +1479,17 @@ afsql_dd_new(GlobalConfig *cfg)
...
-    return AFSQL_DDF_EXPLICIT_COMMITS;
-  else if (strcmp(flag, "dont-create-tables") == 0 || strcmp(flag, "dont_create_tables") == 0)
-    return AFSQL_DDF_DONT_CREATE_TABLES;
+  gint retval = 0;
+  const gchar *flag_copy = __normalize_key(flag);
+
+  if (strcmp(flag_copy, "explicit_commits") == 0)
+    retval = AFSQL_DDF_EXPLICIT_COMMITS;
+  else if (strcmp(flag_copy, "dont_create_tables") == 0)
+    retval =  AFSQL_DDF_DONT_CREATE_TABLES;
```
MrAnno: There is another one in `log_writer_options_lookup_flag`: `no_multi-line`

furiel: Oh I see the same problem exists in that log_writer_options_lookup_flag too. Fine for me, I will fix that too.

MrAnno: Thanks.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/str-utils.c
```diff
@@ -60,3 +60,9 @@ __normalize_key(const gchar *buffer)
 {
   return str_replace_char(buffer, '-', '_');
 }
+
+gchar *
+normalize_flag(const gchar *buffer)
+{
+  return str_replace_char(buffer, '_', '-');
```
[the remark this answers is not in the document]
furiel: I had an earlier version when I tried to use `__normalize_key`, but I ran into all kind of problems, the code broke in unexpected ways, that's why ended up with the opposite normalization.

The thing is syslog-ng already has `cfg_process_flag` which is used to normalize flags. Also `__normalize_key` is not used for flags, but for block arguments, so it would be more important to align to `cfg_process_flag` instead of `__normalize_key`.

`cfg_process_flag` works differently than `__normalize_key`, it transforms underscores to hyphen. In my first attempt, I tried to use `__normalize_key` for every cases, even in `cfg_process_flag`. But the patch complexity increased quickly, and unit tests broke unexpectedly, as it turned out `cfg_process_flag` is used not only for flags, but in some other cases, for example template function command line arguments that depend on value pairs: like format-json. You can follow this by starting from `cfg_process_flag` -> `value_pairs_add_scope` -> `vp_cmdline_parse_scope` -> `value_pairs_new_from_cmdline` -> references.

In summary, the normalization from underscore and hyphen is more dominant in the current core philosophy, and I turned to investigate if I can align the exceptions to `cfg_process_flag`. This resulted in a much simpler patch.

Then I investigated, if I could use `cfg_process_flag` everywhere thus avoiding introducing a new function. It turned out it is not easy either: `cfg_process_flag` works by side effects (setting flag into a memory area), and in some cases we need explicit lookup function without side effects (check log_expr_node_lookup_flag("final") in cfg-tree.c).

Still I could have rewriten some cases to use `cfg_process_flag`, but its API is somewhat complex, and I could have not done it without severe complexity on the patch. I did not feel the the benefit to do so.

The only question that remains, if we can align the block argument normalization and flag normalization: both use normalizing from underscore to hyphen. Maybe it is possible, it can be investigated in another day.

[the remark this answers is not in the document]
Kokan: Thanks for the explanation, I can agree with you to move this PR forward as is.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2167, opened 2018-07-05
Fix suspend race cond

lbudai:
@MrAnno : please check whether it supersedes #2045 or not. 

 * suspend is set by `log_source_flow_control_suspend`
 * suspend is unset by `log_source_wakeup`
 * `free_to_send` blocks reading until suspend bit is set or the window size is 0
 * when window size was 0 before `log_source_flow_control_adjust` is called, then we call `log_source_wakeup`
 * when ack_tracker receives an ACK, then it checks if source is suspended or not (and based on the acknowledge type it makes a decision whether we need to suspend/wakeup the source)

### DIFF
--- a/lib/early_ack_tracker.c
+++ b/lib/early_ack_tracker.c
@@ -63,8 +63,8 @@ early_ack_tracker_manage_msg_ack(AckTracker *s, LogMessage *msg, AckType ack_typ
 
   if (ack_type == AT_SUSPENDED)
     log_source_flow_control_suspend(self->super.source);
-  else
-    log_source_flow_control_adjust(self->super.source, 1);
+
+  log_source_flow_control_adjust(self->super.source, 1);
 
   log_msg_unref(msg);
   log_pipe_unref((LogPipe *)self->super.source);
--- a/lib/logsource.c
+++ b/lib/logsource.c
@@ -53,17 +53,25 @@ log_source_window_empty(LogSource *self)
 }
 
 static inline void
-_flow_control_window_size_adjust(LogSource *self, guint32 window_size_increment)
+_flow_control_window_size_adjust(LogSource *self, guint32 window_size_increment, gboolean last_ack_type_is_suspended)
 {
-  guint32 old_window_size;
+  gboolean suspended;
+  gsize old_window_size = window_size_counter_add(&self->window_size, window_size_increment, &suspended);
 
-  window_size_increment += g_atomic_counter_get(&self->suspended_window_size);
-  old_window_size = g_atomic_counter_exchange_and_add(&self->window_size, window_size_increment);
-  g_atomic_counter_set(&self->suspended_window_size, 0);
+  msg_debug("Window size adjustment",
+            evt_tag_int("old_window_size", old_window_size),
+            evt_tag_int("window_size_increment", window_size_increment),
+            evt_tag_str("suspended_before_increment", suspended ? "TRUE" : "FALSE"),
+            evt_tag_str("last_ack_type_is_suspended", last_ack_type_is_suspended ? "TRUE" : "FALSE"));
 
-  if (old_window_size == 0)
+
+  gboolean need_to_resume_counter = !last_ack_type_is_suspended && suspended;
+  if (need_to_resume_counter)
+    window_size_counter_resume(&self->window_size);
+  if (old_window_size == 0 || need_to_resume_counter)
     log_source_wakeup(self);
-  if (g_atomic_counter_get(&self->window_size) == self->options->init_window_size)
+
+  if (old_window_size+window_size_increment == self->options->init_window_size)
     log_source_window_empty(self);
 }
 
@@ -133,7 +141,14 @@ _flow_control_rate_adjust(LogSource *self)
 void
 log_source_flow_control_adjust(LogSource *self, guint32 window_size_increment)
 {
-  _flow_control_window_size_adjust(self, window_size_increment);
+  _flow_control_window_size_adjust(self, window_size_increment, FALSE);
+  _flow_control_rate_adjust(self);
+}
+
+void
+log_source_flow_control_adjust_when_suspended(LogSource *self, guint32 window_size_increment)
+{
+  _flow_control_window_size_adjust(self, window_size_increment, TRUE);
   _flow_control_rate_adjust(self);
 }
 
@@ -153,11 +168,11 @@ log_source_msg_ack(LogMessage *msg, AckType ack_type)
 void
 log_source_flow_control_suspend(LogSource *self)
 {
-  msg_debug("Source has been suspended", log_pipe_location_tag(&self->super));
+  msg_debug("Source has been suspended",
+            log_pipe_location_tag(&self->super),
+            evt_tag_str("function", __FUNCTION__));
 
-  g_atomic_counter_set(&self->suspended_window_size, g_atomic_counter_get(&self->window_size));
-  g_atomic_counter_set(&self->window_size, 0);
-  _flow_control_rate_adjust(self);
+  window_size_counter_suspend(&self->window_size);
 }
 
 void
@@ -259,10 +274,14 @@ log_source_post(LogSource *self, LogMessage *msg)
   log_msg_add_ack(msg, &path_options);
   msg->ack_func = log_source_msg_ack;
 
-  old_window_size = g_atomic_counter_exchange_and_add(&self->window_size, -1);
+  old_window_size = window_size_counter_sub(&self->window_size, 1, NULL);
 
   if (G_UNLIKELY(old_window_size == 1))
-    msg_debug("Source has been suspended", log_pipe_location_tag(&self->super));
+    {
+      msg_debug("Source has been suspended",
+                log_pipe_location_tag(&self->super),
+                evt_tag_str("function", __FUNCTION__));
+    }
 
   /*
    * NOTE: this assertion validates that the source is not overflowing its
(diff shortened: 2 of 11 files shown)


lib/logsource.c
```diff
@@ -55,15 +59,17 @@ log_source_window_empty(LogSource *self)
...
+  msg_debug("window_size_adjust", evt_tag_int("window_size",
+                                             window_size_counter_get(&self->window_size, NULL)),
+           evt_tag_str("suspended:", suspended ? "TRUE" : "FALSE"));
 
-  if (old_window_size == 0)
+  if (suspended)
     log_source_wakeup(self);
-  if (g_atomic_counter_get(&self->window_size) == self->options->init_window_size)
+
+  if (old_window_size == self->options->init_window_size)
```
MrAnno: The new (incremented) window size should be used here instead of the old one.

Instead of calling `window_size_counter_get` here, we could write `old_window_size + window_size_increment == ...` which is guaranteed to preserve the value of the previous `window_size_counter_add` call.

kira-syslogng:
Build FAILURE

lib/logsource.c
```diff
@@ -38,6 +38,10 @@ gboolean accurate_nanosleep = FALSE;
 void
 log_source_wakeup(LogSource *self)
 {
+  if (!window_size_counter_suspended(&self->window_size))
```
MrAnno: I would not add this condition to `log_source_wakeup`.
The name of this function tells me that it will wake the source up without condition.

kira-syslogng:
Build FAILURE

lib/early_ack_tracker.c
```diff
@@ -61,10 +61,12 @@ early_ack_tracker_manage_msg_ack(AckTracker *s, LogMessage *msg, AckType ack_typ
 {
   EarlyAckTracker *self = (EarlyAckTracker *)s;
 
+  log_source_flow_control_adjust(self->super.source, 1);
+
   if (ack_type == AT_SUSPENDED)
     log_source_flow_control_suspend(self->super.source);
   else
-    log_source_flow_control_adjust(self->super.source, 1);
+    log_source_wakeup(self->super.source);
```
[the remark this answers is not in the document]
lbudai: I think after my change it's not... 
Maybe it's a wrong direction but trying to make suspend/wakeup more symmetrical:
suspend bit is set by suspend and unset by wakeup...

lib/logsource.c
lbudai: yes, you are right

lib/logsource.c
lbudai: well... someone should check this suspend bit...
This is not the best and not the final place (this is why I added the wip flag :) )

MrAnno: Oops, okay, sorry. I'll check it.

lbudai: I'm also playing with this a little bit.

kira-syslogng:
Build FAILURE

lib/early_ack_tracker.c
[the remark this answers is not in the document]
lbudai: I made a small change: 
 * suspend is set by `log_source_flow_control_suspend`
 * suspend is unset by `log_source_wakeup`
 * `free_to_send` blocks reading until suspend bit is set or the window size is 0
 * when window size was 0 before `log_source_flow_control_adjust` is called, then we call `log_source_wakeup`
 * when ack_tracker receives an ACK, then it checks if source is suspended or not (and based on the acknowledge type it makes a decision whether we need to suspend/wakeup the source)

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/atomic-gssize.h
```diff
@@ -0,0 +1,105 @@
...
+}
+
+static inline gsize
+atomic_gssize_get_unsigned(atomic_gssize *a)
+{
+  return (gsize)g_atomic_pointer_get(&a->value);
+}
+
+static inline gssize
+atomic_gssize_racy_get(atomic_gssize *a)
```
MrAnno: I'd remove these 3 racy operations. We had only problems with them in the past.

lbudai: ...and I would not, not part of this PR. I agree with you that we should consider to remove all the racy atomic operations, but... but I also know that I don't want to discuss it it as part of this PR.
This is why I wanted to atomic-gssize PR handled separately. Please, add these notes to there, or, after that PR merged, we could open a new PR where we remove all these operations and explain why we don't need them anymore.

If no other issues, don't forget to approve the requested changes :)

kira-syslogng:
Build SUCCESS

lib/ack_tracker.h
```diff
@@ -84,4 +84,19 @@ ack_tracker_manage_msg_ack(AckTracker *self, LogMessage *msg, AckType ack_type)
   self->manage_msg_ack(self, msg, ack_type);
 }
 
+static inline void
+_ack_tracker_update_source_when_ack_received(LogSource *source, AckType ack_type)
+{
+  if (ack_type == AT_SUSPENDED)
+    {
+      if (log_source_free_to_send(source))
```
MrAnno: I think we should not check free_to_send() here. We might skip suspend() while another thread wakes up the source.

lib/ack_tracker.h
```diff
@@ -84,4 +84,19 @@ ack_tracker_manage_msg_ack(AckTracker *self, LogMessage *msg, AckType ack_type)
...
+        log_source_flow_control_suspend(source);
+    }
+  else
+    {
+      if (!log_source_free_to_send(source))
+        log_source_wakeup(source);
```
MrAnno: The free_to_send() call is a problem here too.
Instead of calling `wakeup` here, I would reset the suspend bit in `_flow_control_window_size_adjust()`, where we can do it safely.

lib/atomic-gssize.h
MrAnno: Okay. I've added my comment to #2159.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lib/window-size-counter.c
```diff
@@ -0,0 +1,101 @@
...
+ * COPYING for details.
+ *
+ */
+
+#include "window-size-counter.h"
+
+#define COUNTER_MASK (((gsize)1<<(8*sizeof(gsize)-1)) - 1)
+static const gsize COUNTER_MAX = COUNTER_MASK;
+static const gsize SUSPEND_MASK = G_MAXSIZE ^ COUNTER_MASK;
+static const gsize RESUME_MASK = G_MAXSIZE >> 1;
```
MrAnno: `atomic_gssize` is implemented with atomic pointers, so I would add a static_assert to compare its size with gsize if we want to use G_MAXSIZE on it.

lib/window-size-counter.c
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "window-size-counter.h"
+
+#define COUNTER_MASK (((gsize)1<<(8*sizeof(gsize)-1)) - 1)
```
MrAnno: Should not it be `1ULL << ...`? I think the cast would be unnecessary this way.

lib/logsource.c
```diff
@@ -53,17 +53,23 @@ log_source_window_empty(LogSource *self)
...
-  g_atomic_counter_set(&self->suspended_window_size, 0);
-
-  if (old_window_size == 0)
+  gboolean suspended;
+  gsize old_window_size = window_size_counter_add(&self->window_size, window_size_increment, &suspended);
+
+  msg_debug("window_size_adjust", evt_tag_int("window_size",
+                                              old_window_size+window_size_increment),
+            evt_tag_str("suspended:", suspended ? "TRUE" : "FALSE"),
+            evt_tag_str("AT_SUSPENDED:", last_ack_type_is_suspended ? "TRUE" : "FALSE"));
```
MrAnno: Would you like to leave this debug message in the final patch?

lib/logsource.c
```diff
@@ -259,10 +270,10 @@ log_source_post(LogSource *self, LogMessage *msg)
   log_msg_add_ack(msg, &path_options);
   msg->ack_func = log_source_msg_ack;
 
-  old_window_size = g_atomic_counter_exchange_and_add(&self->window_size, -1);
+  old_window_size = window_size_counter_sub(&self->window_size, 1, NULL);
 
   if (G_UNLIKELY(old_window_size == 1))
-    msg_debug("Source has been suspended", log_pipe_location_tag(&self->super));
+    msg_debug("Source has been suspended", log_pipe_location_tag(&self->super), evt_tag_str("Function", __FUNCTION__));
```
MrAnno: Please consider adding this new evt tag to `log_source_flow_control_suspend` as well. We have the same debug message there.

I don't know about our evt tag conventions, but it seems we prefer lowercase tag names (`function` vs. `Function`).

MrAnno:
@kira-syslogng do perftest

kira-syslogng:
Build SUCCESS

lib/window-size-counter.c
lbudai: Wouldn't be the type of the expression unsigned long long? I'd stick to the `gsize` type.

lib/logsource.c
lbudai: Well.. I found it very useful during development... but it is possible to print out a better message, I'll try to do that. But I'd keep a debug message here.

lib/logsource.c
lbudai: ok, will check it.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2168, opened 2018-07-06
afsocket: Add ref-counted TLSVerifier struct

bernieh-atlnz:
When a TLS session is initiated, the afsocket module passes a
reference to the AFInetDestDriver instance to the TLSSession
structure to use as verification data. Due to the timing of
configuration changes, the driver structure may be freed before
the TLSSession structure, meaning that TLSSession will contain a
dangling pointer to the AFInetDestDriver. The afsocket module may
try to verify the session using the invalid data, which result in
a crash.

This patch adds two extra function pointers to the
TransportMapperInet structure, one that is used to increment the
driver's reference counter when it is passed to the TLSSession,
and one that is called to decrement the driver's reference
counter when the session structure is destroyed. This ensures
that the driver is only freed once there are no sessions
referencing it.

Signed-off-by: Bernie Harris <bernie.harris@alliedtelesis.co.nz>

### DIFF
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -250,8 +250,8 @@ tls_session_verify_callback(int ok, X509_STORE_CTX *ctx)
 
       tls_log_certificate_validation_progress(ok, ctx);
 
-      if (self->verify_func)
-        return self->verify_func(ok, ctx, self->verify_data);
+      if (self->verifier && self->verifier->verify_func)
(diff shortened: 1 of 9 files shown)


Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

lbudai:
@bernieh-atlnz : thanks for the patch! will check it soon.

kira-syslogng:
Build SUCCESS

PR #2173, opened 2018-07-08
Leak & invalid memory access

Kokan:
Instead of explaining/listing each patch (as they should be more or less self explanatory), here is the main drive of those patches: 
I would like to use [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) for unit tests and as a long term also for functional tests.

It is fairly easy to set it up, it needs only the following:
* CFLAGS  `-O1  -fsanitize=address -fno-omit-frame-pointer`
* The executable should be started with `ASAN_OPTIONS=detect_odr_violation=0` environment variable.

The last environment variable is needed because of the `module_info` is in global scope with the same name in the modules. That is by design, so I think we can ignore those for now.

There are issues, that still exists, but this PR should be fine without finishing the whole thing.

* The following test cases are still failing: `test_stats_cluster, test_secret_storage, test_diskq_full, test_geoip2_parser,test_radix`
* Criterion based tests won't fail on memory leak, but only memory access violation. (This should be checked, what can be done


kira-syslogng:
Build SUCCESS

lib/eventlog/src/evtrec.c
```diff
@@ -74,6 +74,7 @@ evt_rec_add_tags(EVTREC *e, EVTTAG *first, ...)
   va_list ap;
 
   va_start(ap, first);
+  evt_rec_add_tag(e, first);
```
furiel: Can you check the users of this function? For example this should be deleted after the patch:
https://github.com/balabit/syslog-ng/blob/c71f47dfca3316398c833d57253d0514e5995a91/lib/messages.c#L210

modules/dbparser/tests/test_parsers_e2e.c
```diff
@@ -57,9 +57,9 @@ _assert_pattern(PatternDB *patterndb, const gchar *pattern, const gchar *rule, g
   gboolean result;
   LogMessage *msg = log_msg_new_empty();
 
-  log_msg_set_value(msg, LM_V_HOST, MYHOST, strlen(MYHOST));
-  log_msg_set_value(msg, LM_V_PROGRAM, "test", strlen(MYHOST));
-  log_msg_set_value(msg, LM_V_MESSAGE, pattern, strlen(pattern));
+  log_msg_set_value(msg, LM_V_HOST, MYHOST, -1);
```
furiel: Is there an actual problem with the original version, or this patch is a refactor? log_msg_set_value seems to do the same (calculating strlen) when negative value is passed.

Kokan: Yes, the second line sets `"test"` and uses `strlen(MYHOST)`.

furiel: hmmm ok :) I did not catch it.

lib/eventlog/src/evtrec.c
Kokan: Sure there is no user, but only unit test. (Which does not really test much sadly.)
I think the whole test should be redesigned for eventlog, and because of that it should be done when it is converted to criterion.

furiel: Overlook from my side, the caller is evt_rec_add_tagsv instead of evt_rec_add_tags

modules/systemd-journal/journal-reader.c
```diff
@@ -233,6 +233,7 @@ _set_program(JournalReaderOptions *options, LogMessage *msg)
     }
   else
     {
+      g_free(value);
```
lbudai: I refactored a little bit this part of the code, the refactored one is simpler than this. Opened a pr to your branch, please merge.
And thanks for the leak-fixes!

Kokan: Thanks, merged.

kira-syslogng:
Build SUCCESS

PR #2175, opened 2018-07-09
csv_parser test: revwrite based on criterion

norberttak:
csv_parser test: revwrite based on criterion


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

norberttak:
@lbudai : some places the expect can be used instead of assert, you are right. At some places not. I'm sure the line 803 shall be assert for example (as the max_columns shall be correct in later code part)

Other asserts seems to be valid with expect as well. I use the assert because I'm lazy :-) With asserts I don't need to care about the results of previous evaluation. If you wish I'm open to modify the asserts to expect.

lbudai:
@norberttak: I’m ok with asserts, just wanted to know whether we take into account this question(assert vs expect) in case uni test porting.

PR #1897, opened 2018-02-27
Added g_list_free_full in lib/compat/glib & configure.ac

definit3:
Fixes issue #1270  
SLES 11 has glib version <=2-26, in which g_list_free_full function is not present.
Added compat : glib patch to fix the issue.

PR #2182, opened 2018-07-13
Make g_atomic_counter_set() atomic and update minimum glib version to 2.26

MrAnno:
Extracted from #2045.

This function was not used only for initialization, because its name deceived others.
The name does not imply being racy, so I've fixed the implementation.

### DIFF
--- a/lib/atomic.h
+++ b/lib/atomic.h
@@ -25,6 +25,7 @@
 #ifndef ATOMIC_H_INCLUDED
 #define ATOMIC_H_INCLUDED
 
+#include "compat/glib.h"
 
 typedef struct
 {
@@ -68,13 +69,7 @@ g_atomic_counter_racy_get(GAtomicCounter *c)
 static inline void
 g_atomic_counter_set(GAtomicCounter *c, gint value)
 {
-  /* FIXME: we should use g_atomic_int_set, but that's available starting
-   * with GLib 2.10 only, and we only use this function for initialization,
-   * thus atomic write is not strictly needed as there's no concurrency
-   * while initializing a refcounter.
-   */
-
-  c->counter = value;
+  g_atomic_int_set(&c->counter, value);
 }
 
 #endif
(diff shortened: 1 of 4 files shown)


MrAnno:
@kira-syslogng do perftest

lib/atomic.h
```diff
@@ -68,13 +69,7 @@ g_atomic_counter_racy_get(GAtomicCounter *c)
...
 g_atomic_counter_set(GAtomicCounter *c, gint value)
 {
-  /* FIXME: we should use g_atomic_int_set, but that's available starting
-   * with GLib 2.10 only, and we only use this function for initialization,
-   * thus atomic write is not strictly needed as there's no concurrency
-   * while initializing a refcounter.
-   */
-
-  c->counter = value;
+  g_atomic_int_set(&c->counter, value);
```
Kokan: The Centos 6 has `2.28` glib2 version, so I guess we are way passed the `2.10` barrier.
Please update the minimum glib version in `configure.ac` and `CMakeLists.txt`

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/atomic.h
MrAnno: We have compat function for `g_mapped_file_unref` introduced in v2.22, so I'm not sure about the correct minimum version.

MrAnno: Also:
https://github.com/balabit/syslog-ng/pull/1897

MrAnno: @kira-syslogng has a glib-2.26 builder, that might be a good minimum version. The `g_mapped_file_unref` compat function should be removed after the change.

Kokan: I would go with the minimum version of the following set:
* `2.28` distributions min glib version
   (the centos 6 means that to me, but maybe my heuristics is not accurate)
* `2.26` there is a Kira job to catch issues with that version
   (I think it was in connection with centos 6, but @furiel created that job)
*  `2.32`

I think that compat function just rotten there, for the same reason as the atomic set.

MrAnno:
A note:
Macros used by `compat/glib` are missing from CMake.

kira-syslogng:
Build SUCCESS

PR #2183, opened 2018-07-13
[WIP] Client side failback mode

szemere:
Add failback mode to the client side failover function.

The normal **failover-servers** option threats the address in the network destination driver and in the failover-servers list equally.

With this modification by adding the failback option to the current failover configuration, we nominate the address in the network driver configuration to a special role. It will be the **primary** server.

In case the client losts the connection towards the primary server and starts to use one of the failover connections, it will start a probing mechanism to check the health of the primary connection. The probing is realised by TCP connection attempts. If the connection can be established at least **successfull_probes_required** times, client will close the current connection and starts to use the last established connection. (In case we can not connect to the primary server, the counter for successful probes will be cleard.)
The probes will happen in **tcp-probes-interval** intervals, but we will always wait for the result of the last probe before sending out the next one. This might cause bigger time periods between probes if the settings for connection timeout is bigger than the chosen interval. (note: The settings on a Ubuntu based system is usually 2min 10 seconds.)

note1: Until the primary server is recognised as stable, client will use the fail over servers in the "normal failover way", using the already in place **time-reopen** timer to switch between servers if necessary.

### Example configuration
```
destination dst{
      syslog('a"
            failover-servers("b", "c", "d"
                  failback(
                        tcp-probe-interval(60)
                        successful-probes-required(3)
                  )
            )
      transport("tcp")
      port(6514)
      time-reopen(60)
      );
};
```

### Default values
tcp-probe-interval: 60 seconds
successful-probes-required: 3

### Use cases for connection change

- C - client who tries to connect, and send the log messages
- P - primary server (configured in network driver)
- Fn - Nth failover server

### A)
Client is connected to the Primary server. Connection lost.
```
C  P  F1  F2
   ^
```

Client will start to join to the first Failover server, and start to probe the Primary server.
```
C  P  F1  F2
      ^
```

### B)
Client is connected to the first Failover server. Connection lost. Primary server is still not reachable. (Probing is in progress in the background.)
```
C  P  F1  F2
      ^
```

Client will start to join to the second Failover server.
```
C  P  F1  F2
          ^
```

### C)
Client is connected to the last Failover server. Connection lost. Primary server is still not reachable.
```
C  P  F1  F2
          ^
```

Client will start to join to the first Failover server.
```
C  P  F1  F2
      ^
```

### D)
Client is connected to one of the Failover servers. Primary server responded successfully to the "last required" probe.

Client closes the connection with the current Failover server, and will start to build up the higher levels of the connection with the Primary server.
```
C  P  F1  F2  F3
          ^
```

IF the probe was succesfull, the client will use the new connection, and start to send log messages towards the Primary server.
```
C  P  F1  F2  F3
   ^
```

IF building up the connection with the Primary server fails. Client will start to join to the first Failover server.
```
C  P  F1  F2  F3
      ^
```

TODO:
- [ ] properly break down the current connection in case `afsocket_connected_with_fd`
- [ ] split afinet_dd_setup_addresses into:
  - [ ] resolve the current address into GSockAddr
  - [ ] actually change the address. (We want to skip the first action in case we are changing the connection from afsocket_connected_with_fd.)


kira-syslogng:
Build FAILURE

lbudai:
I'd still ask you to consider the following configuration:
```
destination dst{
      syslog("a"
            failover(
                  servers("b", "c", "d")
                  failback(
                        tcp-probe-interval(60)
                        successful-probes-required(3)
                  )
            )
      transport("tcp")
      port(6514)
      time-reopen(60)
      );
};
```
and make `failover-servers` deprecated.

kira-syslogng:
Build FAILURE

szemere:
László, thank you for your comment!
@lbudai, @norberttak : please check the new configuration syntax.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

modules/afsocket/afinet-dest.c
```diff
@@ -367,18 +364,33 @@ _current_server_candidate_hostname(AFInetDestDriver *self)
...
-  if (!self->server_candidates)
-    return self->hostname;
-
-  if (!self->current_server_candidate)
+  if (self->current_server_candidate == g_list_first(self->server_candidates))
     {
-      self->current_server_candidate = g_list_first(self->server_candidates);
-      return _current_server_candidate_hostname(self);
+      _afinet_dd_start_failback_timer(self);
+      self->current_server_candidate = g_list_next(self->current_server_candidate);
```
lbudai: Is it possible that `g_list_next` returns `NULL`?

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

mitzkia:
@kira-syslogng retest this please test branch=micek_failover_failback2;

MrAnno:
Please note that `time-reopen()` is a global option, it can not be used in destinations (the example in the description is incorrect).

modules/afsocket/afinet-dest-failover.c
```diff
@@ -0,0 +1,429 @@
...
+
+void
+afinet_dd_failover_free(AFInetDestDriverFailover *self)
+{
+  if (!self)
+    return;
+
+  _afinet_dd_stop_failback_handlers(self);
+  g_list_free_full(self->servers, g_free);
+  g_sockaddr_unref(self->primary_addr);
```
MrAnno: `g_free(self)` is missing, memory leak.

modules/afsocket/afinet-dest-failover.c
```diff
@@ -0,0 +1,429 @@
...
+
+  if (self->probes_received >= self->probes_required)
+    {
+      msg_notice("Primary server seems to be stable, reconnecting to primary server");
+      _afinet_dd_hand_over_connection_to_afsocket(self);
+    }
+  else
+    {
+      close(self->fd.fd);
+      g_sockaddr_unref(self->primary_addr);
```
MrAnno: `self->primary_addr` should be set to NULL to avoid a double-free in `afinet_dd_failover_free()`.

modules/afsocket/afinet-dest-failover.c
```diff
@@ -0,0 +1,429 @@
...
+      _afinet_dd_start_failback_timer(self);
+    }
+}
+
+static void
+_afinet_dd_tcp_probe_failed(AFInetDestDriverFailover *self)
+{
+  self->probes_received = 0;
+  g_sockaddr_unref(self->primary_addr);
```
MrAnno: `self->primary_addr` should be set to NULL to avoid a double-free in `afinet_dd_failover_free()`.

modules/afsocket/afinet-dest-failover.c
MrAnno: Resolved.

kira-syslogng:
Build SUCCESS

modules/afsocket/afinet-dest-failover.c
```diff
@@ -0,0 +1,443 @@
...
+  return self;
+}
+
+void
+afinet_dd_failover_free(AFInetDestDriverFailover *self)
+{
+  if (!self)
+    return;
+
+  _afinet_dd_stop_failback_handlers(self);
```
MrAnno: This must be called in `deinit()` instead of `free()`, because you initialize handlers in `init()`.

If you create a destination but you forget to add it to a logpath, this will crash.

We don't call `init()` in that case, but `free()` is always called and ivykis crashes if you try to unregister a zero-initialized struct (ivykis prefers -1 values instead of 0, that's why we call `IV_FD_INIT` and `IV_TIMER_INIT`).

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/afsocket/afinet-dest-failover.c
```diff
@@ -0,0 +1,443 @@
...
+}
+
+static void
+_init_current_server(AFInetDestDriverFailover *self )
+{
+  self->current_server = _is_failback_enabled(self) ? g_list_next(_primary(self)) : _primary(self);
+
+  if (_primary(self) == self->current_server)
+    {
+      msg_warning("Last failover server reached, trying the original host again",
```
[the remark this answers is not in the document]
szemere: Discussed with the testers and the documentation team. (It is a 50-50 scenario.) We can change it, or stay with the "original" keyword. (I prefer to leave the message as is, introduced with this patch: https://github.com/balabit/syslog-ng/commit/4be247f507fb086603d68332bd0d9d286ae921ff)

modules/afsocket/afinet-dest.c
```diff
@@ -281,19 +270,54 @@ afinet_dd_setup_addresses(AFSocketDestDriver *s)
...
 afinet_dd_get_dest_name(const AFSocketDestDriver *s)
 {
   const AFInetDestDriver *self = (const AFInetDestDriver *)s;
   static gchar buf[256];
+  const gchar *hostname = _afinet_dd_get_hostname(self);
 
-  if (strchr(self->hostname, ':') != NULL)
-    g_snprintf(buf, sizeof(buf), "[%s]:%d", self->hostname, _determine_port(self));
+  if (strchr(hostname, ':') != NULL)
+    g_snprintf(buf, sizeof(buf), "[%s]:%d", hostname, _determine_port(self));
```
MrAnno: I've noticed 3 things here (These are not a regressions, so we **should not block the PR** with this, I just want to leave a note, because we might want to change them in the future):

1. `afinet_dd_get_dest_name` is used for multiple purposes:
- to generate a persist name for diskq
- to generate a name for the connection (for example, for statistics)
- to generate persist name for the driver

When we generate a persist name for diskq or for the driver, we should use a static name that does not change based on the current failover server (we don't want a separate queue for each failover server).

The name generated for stats can be (IMHO should be) changed based on the current failover server name.

2. When syslog-ng is reloaded, we reconnect to the primary server, which is not a big problem in my opinion, but it worked differently in syslog-ng PE 6.

3. Even if we don't reconnect to the primary server after reloading syslog-ng, the iterator in `AFInetDestDriverFailover` is lost (we store only the `LogWriter` instance in the persist config), so we will start from the beginning. This worked the same weird way in syslog-ng PE 6.

PR #2184, opened 2018-07-13
Fix a groupset/groupunset and map-value-pairs() crash

### LINKED ISSUE #1673: segfault when using question mark in glob
I was playing around with `groupunset` and stumbled on this bug: when using `?` in the expression, syslog-ng crashes:

```
@version: 3.9

log {
  source {
    file("/dev/stdin" flags(no-parse));
  };
  rewrite {
    groupunset(
      values("????")
    );
  };
  destination {
    file("/dev/stdout" template("$(format-json -s nv-pairs)\n"));
  };
};
```

```
[2017-09-14T14:29:54.863177] syslog-ng starting up; version='3.9.1'
test
[2017-09-14T14:29:56.717781] Incoming log entry; line='test'
Segmentation fault
```

MrAnno:
The `template_options` parameter of `value_pairs_foreach()` is mandatory, because it is used by `log_macro_expand()` (through `vp_merge_builtins()`).

There are additional (serious) problems with groupset/groupunset and value-pairs (see: https://github.com/balabit/syslog-ng/issues/1673#issuecomment-331413086 and https://github.com/balabit/syslog-ng/issues/1673#issuecomment-331438954), but fixing the crash itself might not be a bad idea.

Fixes #1673

### DIFF
--- a/lib/rewrite/rewrite-groupset.c
+++ b/lib/rewrite/rewrite-groupset.c
@@ -70,12 +70,13 @@ static void
 log_rewrite_groupset_process(LogRewrite *s, LogMessage **msg, const LogPathOptions *path_options)
 {
   LogRewriteGroupSet *self = (LogRewriteGroupSet *) s;
+  GlobalConfig *cfg = log_pipe_get_config(&s->super);
   LogRewriteGroupSetCallbackData userdata;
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

PR #2186, opened 2018-07-17
New linux-audit() source as SCL

czanik:
Add a new SCL: linux-audit() source. It reads and automatically
parses the Linux audit logs. You can override the file name using
the filename() parameter and the prefix for the created
name-value pairs using the prefix() parameter. Any additional
parameters are passed to the file source.

Example:

```
source s_auditd {
  linux-audit(
    prefix("test.")
    hook-commands(
      startup("auditctl -w /etc/ -p wa")
      shutdown("auditctl -W /etc/ -p wa")
    )
  );
};
```

Signed-off-by: Peter Czanik <peter@czanik.hu>

### DIFF
--- a/scl/linux-audit/linux-audit.conf
+++ b/scl/linux-audit/linux-audit.conf
@@ -1,0 +1,31 @@
+#############################################################################
+# Copyright (c) 2018 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+block source linux-audit(filename("/var/log/audit/audit.log") prefix(".auditd.") ...) {
+  channel {
+    source { file("`filename`" flags(no-parse) `__VARARGS__`); };
+    parser { linux-audit-parser (prefix("`prefix`")); };
+    parser { kv-parser (template("${`prefix`msg}") prefix("`prefix`msg.")); };
+    rewrite { unset(value("`prefix`msg")); };
+  };
+};
+


kira-syslogng:
Build SUCCESS

scl/linux-audit/linux-audit.conf
```diff
@@ -0,0 +1,31 @@
...
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+block source linux-audit(filename("/var/log/audit/audit.log") prefix(".auditd.") ...) {
+  channel {
+    source { file("`filename`" flags(no-parse)`__VARARGS__`); };
+    parser { linux-audit-parser (prefix("`prefix`")); };
+    parser { kv-parser (template("${`prefix`msg}") prefix("`prefix`msg.")); };
```
furiel: What is the purpose of this key in the message, and who sets it?
```
`prefix`msg
```
I quickly ran through the linux-audit-parser documentation, but I do not see linux-audit-parser sets such key (maybe I overlook something?). This should always be empty in my opinion.

What is the purpose of the call to kv-parser?

czanik: This is where most of the content of the audit.log message is stored after parsing. And everything inside are key-value pairs. That's why after the linux-audit parser the kv-parser is also used.
Example:
```
type=CRED_REFR msg=audit(1531902380.657:125): pid=1982 uid=0 auid=0 ses=1 msg='op=PAM:setcred grantors=pam_unix acct="root" exe="/usr/sbin/sshd" hostname=172.16.167.1 addr=172.16.167.1 terminal=ssh res=success'
```
Is turned into:
```
{"test":{"uid":"0","type":"CRED_REFR","ses":"1","pid":"1982","msg":{"terminal":"ssh","res":"success","op":"PAM:setcred","hostname":"172.16.167.1","grantors":"pam_unix","exe":"/usr/sbin/sshd","addr":"172.16.167.1","acct":"root"},"auid":"0"},"SOURCE":"s_auditd","PRIORITY":"notice","MESSAGE":"type=CRED_REFR msg=audit(1531902380.657:125): pid=1982 uid=0 auid=0 ses=1 msg='op=PAM:setcred grantors=pam_unix acct=\"root\" exe=\"/usr/sbin/sshd\" hostname=172.16.167.1 addr=172.16.167.1 terminal=ssh res=success'","ISODATE":"2018-07-18T10:26:21+02:00","HOST_FROM":"centos7","HOST":"centos7","FILE_NAME":"/var/log/audit/audit.log","FACILITY":"user"}
```

scl/linux-audit/linux-audit.conf
```diff
@@ -0,0 +1,31 @@
...
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+block source linux-audit(filename("/var/log/audit/audit.log") prefix(".auditd.") ...) {
+  channel {
+    source { file("`filename`" flags(no-parse)`__VARARGS__`); };
```
furiel: Can you add a space before `__VARARARGS__`?

scl/linux-audit/linux-audit.conf
furiel: Thanks for explanation, I understand now. `prefix`msg is set by linux-audit-parser

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Issue #2187, opened 2018-07-18
dbld/rules image-centos7 fails in python setup part


czanik:
# syslog-ng

## Version of syslog-ng

Latest git snapshot

## Platform

Docker host: openSUSE 42.3, but using the CentOS 7 image

# Issue

## Failure

When trying to run ``dbld/rules image-centos7`` it fails in python setup part:
```
linux-6chp:~/syslog-ng # dbld/rules image-centos7
cd /root/syslog-ng/dbld/images/ && docker build --network=host -t balabit/syslog-ng-centos7 -f centos7.dockerfile .
Sending build context to Docker daemon    129kB
Step 1/21 : FROM centos:7
 ---> 49f7960eb7e4
Step 2/21 : ARG DISTRO=centos7
 ---> Using cache
 ---> 657157439c5f
Step 3/21 : LABEL maintainer "Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>"
 ---> Using cache
 ---> 87fcdc21f1a1
Step 4/21 : COPY helpers/* /helpers/
 ---> Using cache
 ---> 0cc24ff1e93f
Step 5/21 : RUN yum install -y   wget   epel-release
 ---> Using cache
 ---> e664067bd6ef
Step 6/21 : RUN yum install -y   python-pip   python-setuptools
 ---> Using cache
 ---> 0322c2d0a424
Step 7/21 : COPY required-pip/all.txt required-pip/${DISTRO}*.txt /required-pip/
 ---> Using cache
 ---> c82985d049f5
Step 8/21 : RUN cat /required-pip/* | grep -v '^$\|^#' | xargs pip install
 ---> Running in 67c5a4b07245
Collecting nose
  Downloading https://files.pythonhosted.org/packages/99/4f/13fb671119e65c4dce97c60e67d3fd9e6f7f809f2b307e2611f4701205cb/nose-1.3.7-py2-none-any.whl (154kB)
Collecting ply
  Downloading https://files.pythonhosted.org/packages/a3/58/35da89ee790598a0700ea49b2a66594140f44dec458c07e8e3d4979137fc/ply-3.11-py2.py3-none-any.whl (49kB)
Collecting pep8
  Downloading https://files.pythonhosted.org/packages/42/3f/669429ce58de2c22d8d2c542752e137ec4b9885fff398d3eceb1a7f5acb4/pep8-1.7.1-py2.py3-none-any.whl (41kB)
[...]

PR #2188, opened 2018-07-18
dbld: fix pip package versions on older Distro releases

bazsi:
Some packages (e.g.  pylint 2.0.0) declare package tags that are not
supported by ancient setuptools packages, that are available in centos 6 and
7, therefore use fixed versions on ancient distributions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ install:
       criterion-dev
       libmaxminddb-dev
       libxml2-utils
-  - sudo pip install -r requirements.txt
+  - sudo pip install -r dbld/images/required-pip/devshell.txt
 before_script:
   - echo 'Europe/Budapest' | sudo tee /etc/timezone
   - sudo dpkg-reconfigure --frontend noninteractive tzdata
@@ -169,7 +169,7 @@ matrix:
         - export PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig:$PKG_CONFIG_PATH
         - brew update > /dev/null
         - brew install python@2 &> /dev/null || brew link --overwrite python@2
-        - pip2 install --user -r dbld/images/required-pip/all.txt
+        - pip2 install --user -r dbld/images/required-pip/devshell.txt
         - brew bundle --file=contrib/Brewfile
       before_script:
         - ./autogen.sh
--- a/dbld/images/required-apt/all.txt
+++ b/dbld/images/required-apt/all.txt
@@ -52,3 +52,4 @@ dh-systemd
 docbook
 docbook-xsl
 dh-exec
+gnupg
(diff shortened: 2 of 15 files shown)


bazsi:
this should fix #2187

kira-syslogng:
Build SUCCESS

lbudai:
Hmm... Ubuntu 17.10 itself an older distro (https://wiki.ubuntu.com/Releases).

I think we should have a dev-image (which could be latest Ubuntu LTS, or just the latest Ubuntu, or even Debian testing). And only in that image should we use the non-versioned requirements.
Every other image should be a kind of 'support' image and on those images we should go with the versioned requirements.

bazsi:
@lbudai I was trying to use unversioned for the latest of all distros (well except CentOS 7).

This is what I've concluded with:
* artful seems to work fine so far, but I can easily foresee that this should use the versioned stuff eventually (maybe a bit newer than what is currently in the requirements.txt file)
* jessie didn't work, so I added version numbers. 
* I've added stretch without versions as a separate PR.

I am using bionic locally, but java does not compile in that, so I am not ready to submit that as a PR.

So as a policy, I would agree, older releases should be using explicit versions, latest should be unversioned so we uncover incompability issues.

szemere:
Since last time we made the `/requirements.txt` to be a symlink on `/dbld/images/required-pip/all.txt` (which is now empty), travis now fails on pylint.
I built the image locally, but somehow I got error messages on the mongodb module during the build of syslog-ng.

(note: The patch perfectly follows the logic behind the dbld structure! Thank You for that!)

MrAnno:
Sorry, my fault. I'm pushing a commit to fix the macOS build.

kira-syslogng:
Build SUCCESS

czanik:
/me tested centos7 and xenial and those built without a problem

czanik:
@szemere: I can confirm, that stretch builds. I don't have any .deb based systems to test the results.

szemere:
Discussed IRL with @czanik and @algernon , thanks for the help!

The problem only occurs when we try to build syslog-ng from bare source on Debian stretch.
Using the tarball method works fine. (Because however we include the embedded mongo-c driver into the tarball, we actually do not use it.)
Fortunately the mongoc driver is very good ad backward compatibility, so according to my experience syslog-ng works fine with the system provided version. @algernon already opened an issue #2200 about dropping the built in version. If we decide to accept the proposal naturally it will also affects dbld.

I will remove the requested changes, as a workaround we can add `--with-mongoc=system` to configure during building from the source. (But this issue has to be addressed in an another PR.)

kira-syslogng:
Build SUCCESS

gaborznagy:
Can we clear the git tree of the two branches of the 2 PRs #2189 and #2188, please?

It's hard to review, we merge something that we don't want, the other branch got changed, etc.
@bazsi: I think in the past you opened the PR on the other PR's branch, this way it was clear what's actually in the PR.

PR #2190, opened 2018-07-18
Http framing support for batches

bazsi:
This series of patches add body-prefix(), body-suffix() and delimiter() option to the http driver, so that
one case create proper framing around items (formatted by the template specified with body()).

This allows us to post proper JSON encoded arrays as POST payloads, which is required by a number of REST APIs.

I am working on splitting the HTTPDestinationDriver class into a separate Worker and Driver class, which would make it easier to do actual testing in the http destination. Right now, I only did manual tests and it does work and the patches are simple enough. My refactorings are subsequent to this, and it is far from trivial to do bring testing in front of this patch, so please merge this without tests at this point.

Thanks.

### DIFF
--- a/modules/http/http.c
+++ b/modules/http/http.c
@@ -239,6 +239,21 @@ http_dd_set_flush_bytes(LogDriver *d, glong flush_bytes)
   self->flush_bytes = flush_bytes;
 }
 
+void
+http_dd_set_body_prefix(LogDriver *d, const gchar *body_prefix)
+{
+  HTTPDestinationDriver *self = (HTTPDestinationDriver *) d;
+
+  g_string_assign(self->body_prefix, body_prefix);
+}
+
+void
+http_dd_set_body_suffix(LogDriver *d, const gchar *body_suffix)
+{
+  HTTPDestinationDriver *self = (HTTPDestinationDriver *) d;
+
+  g_string_assign(self->body_suffix, body_suffix);
+}
 
 static gchar *
 _sanitize_curl_debug_message(const gchar *data, gsize size)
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build FAILURE

bazsi:
rebased against http-batching.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
One comment I've received that with the changes I've made, prefix is part of the flush_bytes calculation whereas suffix is not. This can be addressed, even though I don't consider this fatal.

bazsi:
This series of patches add body-prefix(), body-suffix() and delimiter() option to the http driver, so that
one case create proper framing around items (formatted by the template specified with body()).

This allows us to post proper JSON encoded arrays as POST payloads, which is required by a number of REST APIs.

I am working on splitting the HTTPDestinationDriver class into a separate Worker and Driver class, which would make it easier to do actual testing in the http destination. Right now, I only did manual tests and it does work and the patches are simple enough. My refactorings are subsequent to this, and it is far from trivial to do bring testing in front of this patch, so please merge this without tests at this point.

Thanks.

kira-syslogng:
Build SUCCESS

modules/http/http.c
```diff
@@ -466,6 +496,8 @@ _flush(LogThreadedDestDriver *s)
   if (self->super.batch_size == 0)
     return WORKER_INSERT_RESULT_SUCCESS;
 
+  _finish_request_body(self);
```
gaborznagy: Sorry if my comment is a matter-of-taste comment (i.e. how I would organize the code),
but I think prefix and suffix should be part of message construction and not part of flush().
Flush should be about the actual message sending and the logic around it, in my opinion.

e.g. _insert_batched()
``` diff
if (_should_initiate_flush(self))
    {
+     _finish_request_body(self);
      return _flush(&self->super);
    }
```
insert_single()
``` diff
self->request_headers = _format_request_headers(self, msg);
+init_request_body(self);
_add_message_to_batch(self, msg);
+_finish_request_body(self);
return _flush(&self->super);

```
What do you think?

bazsi:
Hi,

Flush is also called by the framework, and not just from http itself. In
that case, the insert does not yet know that a flush is coming.

gaborznagy:
I see your point.
Flush method is called in LogThreadedDestDriver if the queue is empty, but we have leftover messages on the worker side.

PR #2192, opened 2018-07-19
Reliable disk buffer non-empty backlog

Kokan:
The reliable disk buffer can contain messages in the backlog, and during syslog-ng lifetime it is perfectly fine (by design).
But if there is a message in the backlog when syslog-ng is starting, it is not expected to have message in the backlog, because it should have been acknowledged or rewinded.
As of now it keeps that backlog slot as a sliding window, and message could be stuck there until other message does not replace it. (No message stuck there forever.)

This patch contains 2 things:
* When reliable disk buffer is started, it check if there is a message in the backlog, and moves them back the reliable queue, therefore they are going to be reprocessed again.
* dqtool is extended, so when the `dqtool info` is used with `-d` (debug) option, it also prints extra information: `backlog_head`, `read_head`, `write_head`, `backlog_len`.

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -214,6 +214,21 @@ _truncate_file(QDisk *self, gint64 new_size)
   return success;
 }
 
+gint64
+qdisk_get_empty_space(QDisk *self)
+{
+  gint64 wpos = qdisk_get_writer_head(self);
+  gint64 bpos = qdisk_get_backlog_head(self);
+
+  if (wpos > bpos)
+    {
+      return (qdisk_get_size(self) - wpos) +
+             (bpos - QDISK_RESERVED_SPACE);
+    }
+
+  return bpos - wpos;
+}
+
 gboolean
 qdisk_push_tail(QDisk *self, GString *record)
 {
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -538,6 +538,20 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
         _truncate_file(self, end_ofs);
     }
 
+  if (!self->options->read_only && self->options->reliable)
+    {
+      if (self->hdr->backlog_len > 0 || self->hdr->backlog_head != self->hdr->read_head)
+        {
+          msg_warning("Reliable disk-buffer backlog is not empty, resetting",
+                      evt_tag_long("backlog_head", self->hdr->backlog_head),
+                      evt_tag_long("backlog_len", self->hdr->backlog_len));
+
+          self->hdr->length += self->hdr->backlog_len;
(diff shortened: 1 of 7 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -558,6 +558,13 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
                evt_tag_str("filename", self->filename),
                evt_tag_int("queue_length", self->hdr->length),
                evt_tag_int("size", self->hdr->write_head - self->hdr->read_head));
+
+      msg_debug("Reliable disk-buffer internal state",
+                evt_tag_str("filename", self->filename),
+                evt_tag_int("backlog_head", self->hdr->backlog_head),
+                evt_tag_int("read_head", self->hdr->read_head),
```
furiel: Please use evt_tag_long.
evt_tag_int uses `"%d"`, so it might not work with gint64.

modules/diskq/qdisk.c
```diff
@@ -538,6 +538,20 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
         _truncate_file(self, end_ofs);
     }
 
+  if (!self->options->read_only && self->options->reliable)
+    {
+      if (self->hdr->backlog_len > 0 || self->hdr->backlog_head != self->hdr->read_head)
+        {
+          msg_warning("Reliable disk-buffer backlog is not empty, resetting",
+                      evt_tag_int("backlog_head", self->hdr->backlog_head),
+                      evt_tag_int("backlog_len", self->hdr->backlog_len));
```
furiel: evt_tag_long here too

modules/diskq/qdisk.c
Kokan: Nice catch, I updated in other messages also (in disk buffer).

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -523,6 +538,20 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
...
+          msg_warning("Reliable disk-buffer backlog is not empty, resetting",
+                      evt_tag_long("backlog_head", self->hdr->backlog_head),
+                      evt_tag_long("backlog_len", self->hdr->backlog_len));
+
+          self->hdr->length += self->hdr->backlog_len;
+          self->hdr->read_head = self->hdr->backlog_head;
+          self->hdr->backlog_len = 0;
+        }
+    }
+
```
lbudai: I'd factor out this to a `static _reset_backlog(...)` function.

Issue #2084, opened 2018-06-01
syslog-ng random segfault in main thread, _is_timestamp at lib/stats/stats-query.c:344


nikat:
# syslog-ng

## Version of syslog-ng

```
syslog-ng 3 (3.15.1)
Config version: 3.15
Installer-Version: 3.15.1
Revision: 
Compile-Date: Jun  1 2018 18:30:48
Module-Directory: //usr/lib64/syslog-ng
Module-Path: //usr/lib64/syslog-ng
Available-Modules: cryptofuncs,hook-commands,snmptrapd-parser,map-value-pairs,confgen,sdjournal,graphite,json-plugin,appmodel,afstomp,syslogformat,date,system-source,pseudofile,basicfuncs,afprog,affile,disk-buffer,linux-kmsg-format,kvformat,add-contextual-data,afuser,dbparser,tags-parser,xml,stardate,tfgetent,cef,csvparser,afsocket
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: on
```

## Platform

CentOS Linux 7.3

# Issue

Random segfault in main thread, _is_timestamp at lib/stats/stats-query.c:344

## Failure

```
[Jun 1 22:47] syslog-ng[4359]: segfault at 480 ip 00007fa7ef24a365 sp 00007ffd8edfa668 error 4 in libc-2.17.so[7fa7ef115000+1b7000]
[Jun 1 23:00] traps: syslog-ng[28158] general protection ip:7f14b0162694 sp:7ffd8e8fb6b0 error:0 in libc-2.17.so[7f14b011a000+1b7000]
```

```
(gdb) backtrace full
#0  0x00007f558f31c3b0 in __strrchr_sse42 () from /usr/lib64/libc.so.6
No symbol table info available.
#1  0x00007f5590e0f0e6 in _is_timestamp (counter_name=<optimized out>) at lib/stats/stats-query.c:344
        last_dot = <optimized out>
#2  _sum_selected_counters (counters=counters@entry=0x181ed20, user_data=user_data@entry=0x7ffca57e1a10) at lib/stats/stats-query.c:357
        counter = 0x7f5584004418
        c = 0x1818c00
        args = 0x7ffca57e1a10
        sum = 0x7ffca57e1a08
[...]

PR #2193, opened 2018-07-19
Minor stats-query fixes and refactor

MrAnno:
During the investigation of #2084, I fixed/refactored a few stats-related things.

This PR contains:
- minor "simplifications" in the implementation of `stats-query`
- a fix for the usage of the format callback of aggregated queries
- a fix for `remove_counter_from_index()`
- a modification in the deallocation of `StatsCounterItem::name`

Questions:
- Why do we use a HashTable (`counter_index`) in `stats-query.c` when we do not call one single lookup function on it? (Only a full iteration is used [linear search])
- This "index" is updated before every query (which requires a full iteration of `stats_cluster_foreach_counter`) call. Does this increase the query performance?

### DIFF
--- a/lib/stats/stats-cluster.c
+++ b/lib/stats/stats-cluster.c
@@ -286,15 +286,15 @@ stats_counter_group_free(StatsCounterGroup *self)
 
 
 static void
-stats_cluster_free_counter_name(StatsCluster *self, gint type, StatsCounterItem *item, gpointer user_data)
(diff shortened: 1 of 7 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng do perftest

kira-syslogng:
Build SUCCESS

PR #2194, opened 2018-07-19
LogProto partial write

Kokan:
The `LogQueue` and the `LogProto` has shared knowledge about the messages states.
The `LogQueue` holds the messages(both need processing, and ones in processing), in case of partial message(the whole message was not sent), the `LogProto` knows, which part of the message is sent and not, from the `LogQueue` point of view that partial message is still in progress. Which is fine, as long as we have both the `LogQueue` and the `LogProto`.

Actually we keep the `LogProto` around as long as the connection is alive, when the connection is closed the `LogProto` and the partial state of the message is thrown away.

Base on the above, if there is a partial message, when the connection is closed (by destination), the partial (and correct) state of the actually processed message is gone. That message is stuck in the `backlog`, from which it cannot be poped out, at least one message have to be in that area.

A proposed solution would be to rewind all item from backlog, when the connection is closed, and there have been a partial write reported by `LogProtoClient`.

From now on the `LogProtoClient` is responsible to send `LPS_PARTIAL` if the last write was partial, and it cannot continue after connection reset.

### DIFF
--- a/lib/logproto/logproto-text-client.c
+++ b/lib/logproto/logproto-text-client.c
@@ -70,7 +70,7 @@ log_proto_text_client_flush(LogProtoClient *s)
   if (rc != len)
     {
       self->partial_pos += rc;
-      return LPS_SUCCESS;
+      return LPS_PARTIAL;
     }
 
   if (self->partial_free)
@@ -129,7 +129,7 @@ log_proto_text_client_post(LogProtoClient *s, LogMessage *logmsg, guchar *msg, g
       return status;
     }
 
-  if (self->partial)
+  if (self->partial || LPS_PARTIAL == status)
     {
       /* NOTE: the partial buffer has not been emptied yet even with the
        * flush above, we shouldn't attempt to write again.
@@ -142,7 +142,7 @@ log_proto_text_client_post(LogProtoClient *s, LogMessage *logmsg, guchar *msg, g
        * This obviously would cause the framing to break. Also libssl
        * returns an error in this case, which is how this was discovered.
        */
-      return status;
+      return LPS_PARTIAL;
     }
 
   *consumed = TRUE;
--- a/lib/logproto/logproto-text-client.c
+++ b/lib/logproto/logproto-text-client.c
@@ -46,45 +46,45 @@ log_proto_text_client_flush(LogProtoClient *s)
   LogProtoTextClient *self = (LogProtoTextClient *) s;
   gint rc;
 
-  /* attempt to flush previously buffered data */
-  if (self->partial)
+  if (!self->partial)
     {
-      gint len = self->partial_len - self->partial_pos;
-
-      rc = log_transport_write(self->super.transport, &self->partial[self->partial_pos], len);
-      if (rc < 0)
-        {
-          if (errno != EAGAIN && errno != EINTR)
-            {
-              msg_error("I/O error occurred while writing",
-                        evt_tag_int("fd", self->super.transport->fd),
-                        evt_tag_error(EVT_TAG_OSERROR));
-              return LPS_ERROR;
-            }
-          return LPS_SUCCESS;
-        }
-      else if (rc != len)
-        {
-          self->partial_pos += rc;
-          return LPS_SUCCESS;
-        }
-      else
-        {
-          if (self->partial_free)
-            self->partial_free(self->partial);
-          self->partial = NULL;
-          if (self->next_state >= 0)
-            {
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/logwriter.c
```diff
@@ -1144,6 +1144,10 @@ log_writer_write_message(LogWriter *self, LogMessage *msg, LogPathOptions *path_
                                                     &consumed);
 
       self->partial_write = status == LPS_PARTIAL;
+      if (self->partial_write)
+       {
+         msg_debug("Partial message writer", evt_tag_printf("message", "%s", self->line_buffer->str));
```
lbudai: If it will be part of the final PR then I'd recommend to use `msg_trace` instead.

lib/logproto/logproto-text-client.c
```diff
@@ -119,15 +119,14 @@ static LogProtoStatus
 log_proto_text_client_post(LogProtoClient *s, LogMessage *logmsg, guchar *msg, gsize msg_len, gboolean *consumed)
 {
   LogProtoTextClient *self = (LogProtoTextClient *) s;
-  gint rc;
 
   /* try to flush already buffered data */
   *consumed = FALSE;
-  rc = log_proto_text_client_flush(s);
-  if (rc == LPS_ERROR)
+  const LogProtoStatus status = log_proto_text_client_flush(s);
```
lbudai: Why the `const` ?

lib/logproto/logproto-text-client.c
```diff
@@ -46,45 +46,48 @@ log_proto_text_client_flush(LogProtoClient *s)
...
-          return LPS_SUCCESS;
+          msg_error("I/O error occurred while writing",
+                    evt_tag_int("fd", self->super.transport->fd),
+                    evt_tag_error(EVT_TAG_OSERROR));
+          return LPS_ERROR;
         }
-      else
+      return LPS_SUCCESS;
+    }
+  else if (rc != len)
```
lbudai: As the `if` branch returns, the `else` is not necessary.

lib/logproto/logproto-text-client.c
```diff
@@ -46,45 +46,48 @@ log_proto_text_client_flush(LogProtoClient *s)
...
+    {
+      self->partial_pos += rc;
+      return LPS_SUCCESS;
+    }
+  else
```
lbudai: I think this `else` also can be removed.

lib/logproto/logproto-text-client.c
```diff
@@ -71,23 +71,19 @@ log_proto_text_client_flush(LogProtoClient *s)
...
+    {
+      self->state = self->next_state;
+      self->next_state = -1;
     }
 
+  log_proto_client_msg_ack(&self->super, 1);
+
+  /* NOTE: we return here to give a chance to the framed protocol to send the frame header. */
   return LPS_SUCCESS;
 }
```
lbudai: Squash this patch into the previous patch.

lib/logwriter.c
```diff
@@ -1142,6 +1143,8 @@ log_writer_write_message(LogWriter *self, LogMessage *msg, LogPathOptions *path_
                                                     self->line_buffer->len,
                                                     &consumed);
 
+      self->partial_write = status == LPS_PARTIAL;
```
lbudai: Use braces around the logical expression.

lib/logwriter.c
Kokan: I did not plan to keep it.

lib/logproto/logproto-text-client.c
Kokan: I split those into two commit on purpose, because I wanted to create as small changes in commit as I can get. But well I do not mind to squash them.

lib/logproto/logproto-text-client.c
Kokan: I like to give the compiler and the reader a hint, it won't change.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
Nice catch @gaborznagy I've updated the last commit based on your comment, please check it.

Issue #2172, opened 2018-07-08
template support for python binding


furiel:
When someone tries to use template in python destination:
```
destination { python(class(MyDestination) template("hello\n")); };
```
receives
```
Error parsing python, syntax error, unexpected KW_TEMPLATE in
[...]
27---->       destination { python(class(Test) template("hello\n")); };
27---->                                        ^^^^^^^^
```

As a workaround, I can still evaluate and send template to python destination with value_pairs()
```
destination { python(class(MyDestination) value_pairs(pair("foo" : "hello\n"))); };
```
though python destination could support template directly for convenience.

EDIT: java binding supports both template option in grammar, and evaluation of the template inside the binding. Both features are missing from python binding, the latter one seems even more interesting.

PR #2196, opened 2018-07-20
python-destination: template support

furiel:
Fixes: https://github.com/balabit/syslog-ng/issues/2172

This patchset adds support for template formatting in python destination. All necessary parameters are exposed to the user code: LTZ_SEND, LTZ_LOCAL, sequence number, and LogTemplateOptions instance of the destination. This patchset does not support construction of LogTemplateOptions, nor their fields are exposed to the user, only that is generated by the grammar is available to the user. Though such functionality might be useful, and can be added later.

LTZ_SEND, LTZ_LOCAL is exposed in a newly created syslog-ng module. LogTemplateOptions and sequence number is exposed through self (the class, provided by the user).

Examples:
(Don't forget to add `import syslogng`)

* Creating a new LogTemplate object 
```
self.log_template = syslogng.LogTemplate("${S_STAMP}")
self.log_template = syslogng.LogTemplate("${S_STAMP}", self.template_options)
```

* LogTemplateException

In case a malformed template is provided, an exception will be raised. For that sake, a new exception class is added: syslogng.LogTemplateException:
```
try:
    syslogng.LogTemplate("$(incomplete");
except syslogng.LogTemplateException as e:
     print(e)
```

* Formatting template:
Templates can be formatted using the format method. 
Return value: with python3, the return value is unicode (the standard string representation) if the result is unicode or covertible to unicode, else PyBytes. With python2, the return value is PyBytes (the standart string representation).
```
self.log_template.format(msg)
self.log_template.format(msg, self.template_options)
self.log_template.format(msg, self.template_options, syslogng.LTZ_SEND)
self.log_template.format(msg, self.template_options, syslogng.LTZ_SEND, self.seqnum)
```
Only msg and is mandatory for the `format` method.

* Formatting template works with keyword arguments as well:
```
self.log_template.format(msg=msg, options=self.template_options, tz=syslogng.LTZ_SEND, seqnum=self.seqnum))
```

Template options must be either provided in the constructor or in the `format` method. Otherwise syslog-ng will raise RuntimeError during in the `format` method.

Notes to developers:
- In practice some macros could work without any LogTemplate options provided (NULL passed to log_template_format). Though in case of some macros: like STAMP, syslog-ng would crash, which I wanted to avoid in a language binding. That's why I am enforcing that LogTemplateOptions must be passed either in constructor or in the format method.

- I was trying hard to keep reference counts at bay. Though there is one place (somewhat hidden hidden :( )
that I know there is a reference leak: PY_LTZ_LOCAL, PY_LTZ_SEND. So that I could unref these, there should be a pair for python_module_init, like deinit, but syslog-ng currently does not support deinit for modules (so files). Or some kind of garbage collection mechanism. I created an internal jira ticket about it. As for this reference leak, this does **not** result in memory leak. Interestingly enough, it seems many (all?) PyObject numbers share the same object, so by the time I created these objects,  the reference count was already 433 for PY_LTZ_SEND (0), which means the zero-object was already reffed many times, and most probably will not disappear when syslog-ng exits. We cannot unload the zero object. So this is a cosmetic reference leak.

### DIFF
--- a/modules/python/CMakeLists.txt
+++ b/modules/python/CMakeLists.txt
@@ -35,6 +35,8 @@ set(PYTHON_SOURCES
     python-logmsg.c
     python-logtemplate.h
     python-logtemplate.c
+    python-logtemplate-options.h
+    python-logtemplate-options.c
     python-global-code-loader.h
     python-global-code-loader.c
     python-debugger.c
--- a/modules/python/CMakeLists.txt
+++ b/modules/python/CMakeLists.txt
@@ -43,6 +43,8 @@ set(PYTHON_SOURCES
     python-debugger.h
     python-logparser.h
     python-logparser.c
+    python-integerpointer.h
+    python-integerpointer.c
     compat/compat-python.c
     ${CMAKE_CURRENT_BINARY_DIR}/python-grammar.c
     ${CMAKE_CURRENT_BINARY_DIR}/python-grammar.h
--- a/modules/python/CMakeLists.txt
+++ b/modules/python/CMakeLists.txt
@@ -33,6 +33,8 @@ set(PYTHON_SOURCES
     python-parser.h
     python-logmsg.h
     python-logmsg.c
+    python-logtemplate.h
+    python-logtemplate.c
     python-global-code-loader.h
     python-global-code-loader.c
     python-debugger.c
--- a/modules/python/compat/compat-python-v3.c
+++ b/modules/python/compat/compat-python-v3.c
@@ -21,9 +21,17 @@
  *
  */
 
+#include "compat-python.h"
+
 void
 py_init_argv(void)
 {
   static wchar_t *argv[] = {L"syslog-ng"};
   PySys_SetArgvEx(1, argv, 0);
 }
+
+PyObject *
+int_as_pyobject(gint num)
+{
+  return PyLong_FromLong(num);
+};
--- a/modules/python/compat/compat-python.h
+++ b/modules/python/compat/compat-python.h
@@ -36,5 +36,6 @@
 #endif
 
 void py_init_argv(void);
+PyObject *int_as_pyobject(gint num);
 
 #endif
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -27,6 +27,7 @@
 #include "python-value-pairs.h"
 #include "python-logmsg.h"
 #include "python-logtemplate-options.h"
+#include "python-integerpointer.h"
 #include "python-helpers.h"
 #include "logthrdestdrv.h"
 #include "stats/stats.h"
@@ -52,6 +53,7 @@ typedef struct
     PyObject *is_opened;
     PyObject *send;
     PyObject *log_template_options;
+    PyObject *seqnum;
   } py;
 } PythonDestDriver;
 
@@ -219,6 +221,8 @@ _py_init_bindings(PythonDestDriver *self)
 
   self->py.log_template_options = py_log_template_options_new(&self->template_options);
   PyObject_SetAttrString(self->py.class, "template_options", self->py.log_template_options);
+  self->py.seqnum = py_integer_pointer_new(&self->super.seq_num);
+  PyObject_SetAttrString(self->py.class, "seqnum", self->py.seqnum);
   Py_DECREF(self->py.log_template_options);
 
   self->py.instance = _py_invoke_function(self->py.class, NULL, self->class, self->super.super.super.id);
@@ -254,6 +258,7 @@ _py_free_bindings(PythonDestDriver *self)
   Py_CLEAR(self->py.is_opened);
   Py_CLEAR(self->py.send);
   Py_CLEAR(self->py.log_template_options);
+  Py_CLEAR(self->py.seqnum);
 }
 
 static gboolean
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -26,6 +26,7 @@
 #include "python-module.h"
 #include "python-value-pairs.h"
 #include "python-logmsg.h"
+#include "python-logtemplate-options.h"
 #include "python-helpers.h"
 #include "logthrdestdrv.h"
 #include "stats/stats.h"
@@ -50,6 +51,7 @@ typedef struct
     PyObject *instance;
     PyObject *is_opened;
     PyObject *send;
+    PyObject *log_template_options;
   } py;
 } PythonDestDriver;
 
@@ -215,6 +217,10 @@ _py_init_bindings(PythonDestDriver *self)
       return FALSE;
     }
 
+  self->py.log_template_options = py_log_template_options_new(&self->template_options);
+  PyObject_SetAttrString(self->py.class, "template_options", self->py.log_template_options);
+  Py_DECREF(self->py.log_template_options);
+
   self->py.instance = _py_invoke_function(self->py.class, NULL, self->class, self->super.super.super.id);
   if (!self->py.instance)
     {
@@ -247,6 +253,7 @@ _py_free_bindings(PythonDestDriver *self)
   Py_CLEAR(self->py.instance);
   Py_CLEAR(self->py.is_opened);
   Py_CLEAR(self->py.send);
+  Py_CLEAR(self->py.log_template_options);
 }
 
 static gboolean
--- a/modules/python/python-integerpointer.c
+++ b/modules/python/python-integerpointer.c
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2018 Balabit
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "python-integerpointer.h"
+
+PyObject *
+py_integer_pointer_new(gpointer ptr)
+{
+  PyIntegerPointer *self = PyObject_New(PyIntegerPointer, &py_integer_pointer_type);
+  if (!self)
+    return NULL;
+
+  self->ptr = ptr;
+
+  return (PyObject *) self;
+}
+
+static PyObject *
+_as_int(PyObject *s)
+{
+  PyIntegerPointer *self = (PyIntegerPointer *)s;
+  return PyLong_FromLong(*self->ptr);
+}
+
+static PyObject *
+_as_str(PyObject *s)
+{
+  PyIntegerPointer *self = (PyIntegerPointer *)s;
+  return PyUnicode_FromFormat("%d", *self->ptr);
+}
+
+PyTypeObject py_integer_pointer_type =
+{
+  PyVarObject_HEAD_INIT(&PyType_Type, 0)
+  .tp_name = "IntegerPointer",
+  .tp_basicsize = sizeof(PyIntegerPointer),
+  .tp_dealloc = (destructor) PyObject_Del,
+  .tp_flags = Py_TPFLAGS_DEFAULT,
+  .tp_doc = "IntegerPointer class exposes an integer by a pointer",
+  .tp_as_number = &(PyNumberMethods){ .nb_int = _as_int },
+  .tp_str = _as_str,
+  0,
+};
+
+void
+py_integer_pointer_init(void)
+{
+  PyType_Ready(&py_integer_pointer_type);
+}
--- a/modules/python/python-logtemplate-options.c
+++ b/modules/python/python-logtemplate-options.c
@@ -23,8 +23,6 @@
 
 #include "python-logtemplate-options.h"
 
-static PyTypeObject py_log_template_options_type;
-
 PyObject *
 py_log_template_options_new(LogTemplateOptions *template_options)
 {
@@ -38,7 +36,7 @@ py_log_template_options_new(LogTemplateOptions *template_options)
 }
 
 
-static PyTypeObject py_log_template_options_type =
+PyTypeObject py_log_template_options_type =
 {
   PyVarObject_HEAD_INIT(&PyType_Type, 0)
   .tp_name = "LogTemplateOptions",
--- a/modules/python/python-logtemplate-options.c
+++ b/modules/python/python-logtemplate-options.c
@@ -23,6 +23,12 @@
 
 #include "python-logtemplate-options.h"
 
+int
+py_is_log_template_options(PyObject *obj)
+{
+  return PyType_IsSubtype(Py_TYPE(obj), &py_log_template_options_type);
+}
+
 PyObject *
 py_log_template_options_new(LogTemplateOptions *template_options)
 {
--- a/modules/python/python-logtemplate-options.c
+++ b/modules/python/python-logtemplate-options.c
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2018 Balabit
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "python-logtemplate-options.h"
+
+static PyTypeObject py_log_template_options_type;
+
+PyObject *
+py_log_template_options_new(LogTemplateOptions *template_options)
+{
+  PyLogTemplateOptions *self = PyObject_New(PyLogTemplateOptions, &py_log_template_options_type);
+  if (!self)
+    return NULL;
+
+  self->template_options = template_options;
+
+  return (PyObject *) self;
+}
+
+
+static PyTypeObject py_log_template_options_type =
+{
+  PyVarObject_HEAD_INIT(&PyType_Type, 0)
+  .tp_name = "LogTemplateOptions",
+  .tp_basicsize = sizeof(PyLogTemplateOptions),
+  .tp_dealloc = (destructor) PyObject_Del,
+  .tp_flags = Py_TPFLAGS_DEFAULT,
+  .tp_doc = "LogTemplateOptions class encapsulating a syslog-ng LogTemplateOptions",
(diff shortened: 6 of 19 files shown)


furiel:
@kira-syslogng retest this please test branch=furiel-python-template-support;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=furiel-python-template-support;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=furiel-python-template-support;

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/python/CMakeLists.txt
```diff
@@ -33,6 +33,10 @@ set(PYTHON_SOURCES
     python-parser.h
     python-logmsg.h
     python-logmsg.c
+    python-logtemplate.h
+    python-logtemplate.c
+    python-logtemplateoptions.h
```
Kokan: Optional: It would be easier to read IMHO if it would be python-logtemplate-options.

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,145 @@
...
+static void
+py_log_template_free(PyLogTemplate *self)
+{
+  log_template_unref(self->template);
+  g_free(self->template_options);
+  PyObject_Del(self);
+}
+
+PyObject *
+py_log_template_new(gchar *template)
```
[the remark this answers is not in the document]
furiel: This is a dead code indeed. I started off by copying PyLogMessage, and accidentally created this as well. I will remove it.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng retest this please test branch=furiel-python-template-support;

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=furiel-python-template-support;

kira-syslogng:
Build SUCCESS

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,136 @@
...
+}
+
+static PyObject *
+py_log_template_constructor(PyTypeObject *type, PyObject *args, PyObject *kwds)
+{
+  gchar *template_string;
+  if (!PyArg_ParseTuple(args, "s", &template_string))
+    return NULL;
+
+  LogTemplate *template = log_template_new(configuration, NULL);
```
furiel: Note to myself: double check: NULL instead of configuration?

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng retest this please test branch=furiel-python-template-support;

furiel:
@kira-syslogng retest this please test branch=furiel-python-template-support;

kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng retest this please test branch=furiel-python-template-support;

kira-syslogng:
Build SUCCESS

furiel:
I finished with the update, removing the WIP flag.

@Kokan I resolved your comments. There were a few changes since you last reviewed. These are the changes:
- In log_template_new, configuration is replaced with NULL. API forces users to pass LogTemplateOptions either to the constructor, or format method.
- Constructor now can receive PyTemplateOptions, and can save it for later use.
- Parameters of the format method got fancy names. The format method can be called using keywords, not only with positional parameters.
- Checking type of the first parameter (msg) in format method, and raise TypeError if necessary.
- Added unit tests

I updated the description of the PR accordingly.

modules/python/python-logtemplate.c
```diff
@@ -52,6 +56,56 @@ py_log_template_new(gchar *template)
...
+    {
+      PyErr_Format(PyExc_RuntimeError,
+                   "Error compiling template: %s", error->message);
+      g_clear_error(&error);
+      log_template_unref(template);
+      return NULL;
+    }
+
+  PyLogTemplate *self = (PyLogTemplate *)type->tp_alloc(type, 0);
+  if (!self)
```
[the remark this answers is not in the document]
Kokan: Resolved.

modules/python/compat/compat-python.h
```diff
@@ -36,5 +36,7 @@
 #endif
 
 void py_init_argv(void);
+PyObject *py_as_int(gint num);
```
furiel: Wrong name: should be int_as_pyobject

modules/python/compat/compat-python-v3.c
```diff
@@ -21,9 +21,23 @@
...
+
+PyObject *
+py_as_int(gint num)
+{
+  return PyLong_FromLong(num);
+};
+
+PyObject *
+slng_PyMethod_New(PyObject *func, PyObject *class)
+{
```
bazsi: I would rather use a prefix the is not related to syslog-ng, e.g. compat_PyMethod_New or something along those lines, that clearly indicates that it is only a wrapper around PyMethod_New, and exists for compatibility reasons only.

modules/python/python-dest.c
```diff
@@ -199,6 +202,26 @@ _py_invoke_deinit(PythonDestDriver *self)
...
+{
+  static PyMethodDef getSeqNumMethod = {"getSeqNum", (PyCFunction)_py_dest_get_seqnum, METH_VARARGS, "Read sequence number"};
+  PyObject *dest_driver = PyLong_FromVoidPtr(self);
+  PyObject *func = PyCFunction_New(&getSeqNumMethod, dest_driver);
+  Py_DECREF(dest_driver);
+  self->py.seqnum_method = slng_PyMethod_New(func, self->py.class);
+  Py_DECREF(func);
+  PyObject_SetAttrString(self->py.class, "getSeqNum", self->py.seqnum_method);
+  Py_DECREF(self->py.seqnum_method);
+}
```
bazsi: it's much easier to register methods through the use of the tp_methods member of the PyType that you register. (see python-logmsg for an example). Is there a specific reason why you are doing it this way?

bazsi: I think I now understand, the object is instantiated by a Python class that does not derive from any superclass, this way we can't publish methods from our C code.

I think that's a design issue in the interface that we should - eventually - rectify. Backward compatibility could be an issue, but I think it would be a good tradeof that if you want the new fields (e.g. template_options and getSeqNum), you need to derive your class from syslogng.DestinationDriver or something like that. The java interface also uses base classes and I think that's a good solution.

modules/python/python-dest.c
```diff
@@ -215,6 +238,12 @@ _py_init_bindings(PythonDestDriver *self)
       return FALSE;
     }
 
+  self->py.log_template_options = py_log_template_options_new(&self->template_options);
+  PyObject_SetAttrString(self->py.class, "template_options", self->py.log_template_options);
+  Py_DECREF(self->py.log_template_options);
+
```
bazsi: again, this should be a derived attribute, not defined dynamically this way.

modules/python/python-logtemplate-options.c
```diff
@@ -0,0 +1,49 @@
...
+    return NULL;
+
+  self->template_options = template_options;
+
+  return (PyObject *) self;
+}
+
+
+PyTypeObject py_log_template_options_type =
+{
```
bazsi: You need to call PyType_Ready() on this before using it.

modules/python/python-logtemplate-options.h
```diff
@@ -0,0 +1,41 @@
...
+{
+  PyObject_HEAD
+  LogTemplateOptions *template_options;
+} PyLogTemplateOptions;
+
+extern PyTypeObject py_log_template_options_type;
+
+PyObject *py_log_template_options_new(LogTemplateOptions *template_options);
+void py_log_template_options_init(void);
+
```
bazsi: this function is not defined anywhere (and probably should contain the missing PyType_Ready call)

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,157 @@
...
+  gint seqnum = 0;
+
+  static const gchar *kwlist[] = {"msg", "template_options", "tz", "seqnum", NULL};
+
+  if (!PyArg_ParseTupleAndKeywords(args, kwrds, "O|Oii", (gchar **)kwlist,
+                                   &msg, &log_template_options, &tz, &seqnum))
+    return NULL;
+
+  if (Py_TYPE(msg) != &py_log_message_type)
+    {
```
bazsi: Please extract the condition into a macro/inline function. The conventions in Python are PySomething_Check(), e.g. PyString_Check().

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,157 @@
...
+      PyErr_Format(PyExc_TypeError,
+                   "LogMessage expected in the first parameter");
+      return NULL;
+    }
+
+  if (log_template_options && (Py_TYPE(log_template_options) != &py_log_template_options_type))
+    {
```
bazsi: again, please extract the type check to a macro/inline function.

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,157 @@
...
+      PyErr_Format(PyExc_TypeError,
+                   "LogTemplateOptions expected in the second parameter");
+      return NULL;
+    }
+
+  LogTemplateOptions *template_options = log_template_options ? log_template_options->template_options :
+                                         self->template_options;
+  if (!template_options)
```
bazsi: naming here is a bit unfortunate. template_options is easily confused by log_template_options.

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,157 @@
...
+{
+  PyLogTemplate *self = (PyLogTemplate *)s;
+
+  PyLogMessage *msg;
+  PyLogTemplateOptions *log_template_options = NULL;
+  gint tz = LTZ_SEND;
+  gint seqnum = 0;
+
+  static const gchar *kwlist[] = {"msg", "template_options", "tz", "seqnum", NULL};
+
```
bazsi: I would only use "options" as the name of the kwarg. template is easily implied.

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,157 @@
...
+PyObject *
+py_log_template_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
+{
+  gchar *template_string;
+  PyLogTemplateOptions *log_template_options = NULL;
+  if (!PyArg_ParseTuple(args, "s|O", &template_string, &log_template_options))
+    return NULL;
+
+  if (log_template_options && (Py_TYPE(log_template_options) != &py_log_template_options_type))
+    {
```
bazsi: PY_TYPE() extraction.

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,157 @@
...
+                   "LogTemplateOptions expected in the second parameter");
+      return NULL;
+    }
+
+  LogTemplate *template = log_template_new(NULL, NULL);
+  GError *error = NULL;
+  if (!log_template_compile(template, template_string, &error))
+    {
+      PyErr_Format(PyExc_RuntimeError,
+                   "Error compiling template: %s", error->message);
```
bazsi: could we use a specific exception class? RuntimeError is very generic.

kira-syslogng:
Build SUCCESS

modules/python/python-logtemplate.c
furiel: Good idea! I created a new LogTemplateException class. Updated the description.

modules/python/python-logtemplate-options.h
furiel: Thanks I added the missing PyType_Ready. When to call this init? I felt a good idea to to link logtemplate and logtemplateoptions: so in the end I called py_log_template_options_init() from py_log_template_init(). That would only be unfortunate if LogTemplateOptions could live in a module while LogTemplate not, but I cannot imagine such scenario, except for unit tests. But for now I put every such tests into test_python_template.

modules/python/python-logtemplate-options.c
furiel: Fixed thanks

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/python/python-logtemplate.c
bazsi: I wasn't trying to suggest a name for the function, only give you a pointer how these functions usually look in CPython. 

I _think_ it would read better if we used our own internal style (e.g. _py_is_log_message or something like that) conventions only use CPython naming when calling CPython or doing some compat functions.

Sorry for the confusion.

furiel:
I want to fight with `getSeqNum` another round. It feels a little bit awkward for me that we have a getter function for this. It would me more natural to use
```
template.format(self, self.temlate_options, 0, self.seqnum) 
```
instead of
```
template.format(self, self.temlate_options, 0, self.getSeqNum()) 
```
This  would result in a deviation from the java binding, where we have only getters, but I do not feel that a problem.

Adding WIP flag.

kira-syslogng:
Build SUCCESS

furiel:
@bazsi I was thinking about introducing the inheritance. Here is how I can summarize.

As I see, the benefit of inheritance would be code simplicity at the cost of user API complexity:
we can define class methods with `tp_methods` array at the cost that users need to write `class ...(DestinationDriver)` instead of `class ...(object)`. Is there any other benefit to introduce inheritance?

There is a little difference between `self.getSeqNum` and `self.template_options`: while `getSeqNum` is a `class object` (method), `template_options` is an `instance object`. With that, `template_options` cannot profit (in terms of code simplicity) from inheritance: I still need to `setattr` it, I just need to do it in the superclass constructor, because `PyTypeObject` define class methods and not instance methods/objects. Generally speaking, instance variables cannot profit from inheritance, only class methods.

What kind of methods should the superclass have. For now the only candidate is the getSeqNum getter function. But I created a mini poc, and I believe I can replace the getter function with a seqnum instance variable. 

In java binding, we have 4 methods inherited, and 3 are getter functions: config, template options and sequence number, and we have `getOption` (not LogTemplateOptions, but java destination options), but that is already handled differently in python. That is a good approximation about the candidates. Except for config, because I do not need to expose to use templates because I am forcing to use `LogTemplateOptions`. Leaving sequence number and template options in the loop.

The question is, what is our preference: getters or instance variables? If instance variables, I think there is no code simplicity gain. If getters, then we can evaluate evaluate if the code simpicity worth the api complexity.  Personally I like variables over getter functions, but thats only matter of taste. If we opt for getters, then I would of course rewrite `self.template_options` to be `getTemplateOptions()` too.

I am not much worried of backward compatibility in the latter case, the current user codes would still work (when inherited from Object). Only new codes need to inherit, and only when users want to use templates.

I am wondering we have inheritance in java because it is neat, or we are forced to? Could we add variables/methods to the user class to the user class in such dynamic way as in python?

bazsi:
I think this is a perfect topic for in-person discussion, however it would
be pretty difficult to conclude in writing. Let's discuss this when we are
both at the office next.

Bazsi

furiel:
@kira-syslogng retest this please test branch=furiel-python-template-support;

kira-syslogng:
Build SUCCESS

furiel:
Removed the wip flag. getSeqNum() getter function turned into an attribute: self.seqnum. The code is somewhat simplified, because I do not need to create a function dynamically. Though I needed to add a new type: I named it class IntegerPointer. An instance of this class can encapsulate an integer behind a pointer, and automatically converts to python integer by dereferencing the pointer. Description is updated accordingly.

The pr is ready to review.

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,163 @@
...
+
+static PyMethodDef py_log_template_methods[] =
+{
+  { "format", (PyCFunction)py_log_template_format, METH_VARARGS | METH_KEYWORDS, "format template" },
+  { NULL }
+};
+
+PyTypeObject py_log_template_type =
+{
+  PyVarObject_HEAD_INIT(&PyType_Type, 0)
```
MrAnno: Shouldn't we use `PyObject_HEAD_INIT` here? PyLogTemplate is a `PyObject`, not a `PyVarObject`.

modules/python/python-logtemplate-options.c
```diff
@@ -0,0 +1,61 @@
...
+  PyVarObject_HEAD_INIT(&PyType_Type, 0)
```
MrAnno: Shouldn't we use PyObject_HEAD_INIT here too?

modules/python/python-integerpointer.c
```diff
@@ -0,0 +1,70 @@
...
+static PyObject *
+_as_str(PyObject *s)
+{
+  PyIntegerPointer *self = (PyIntegerPointer *)s;
+  return PyUnicode_FromFormat("%d", *self->ptr);
+}
+
+PyTypeObject py_integer_pointer_type =
+{
+  PyVarObject_HEAD_INIT(&PyType_Type, 0)
```
MrAnno: Shouldn't we use PyObject_HEAD_INIT here too?

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,163 @@
...
+  GString *result = scratch_buffers_alloc();
+  log_template_format(self->template, msg->msg, log_template_options, tz, seqnum, NULL, result);
+
+  return _py_string_from_string(result->str, result->len);
+}
+
+PyObject *
+py_log_template_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
+{
+  gchar *template_string;
```
MrAnno: This should be const (`PyArg_ParseTuple` with "s" specifies const)

modules/python/python-integerpointer.c
```diff
@@ -0,0 +1,70 @@
...
+  .tp_name = "IntegerPointer",
+  .tp_basicsize = sizeof(PyIntegerPointer),
+  .tp_dealloc = (destructor) PyObject_Del,
+  .tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
+  .tp_doc = "IntegerPointer class exposes an integer by a pointer",
+  .tp_new = PyType_GenericNew,
```
MrAnno: I think we don't want to create `IntegerPointer` instances from Python code.
You could set `tp_new` to NULL to prevent Python-side instantiation.
`Py_TPFLAGS_BASETYPE` is also is unnecessary.

modules/python/python-logtemplate-options.c
```diff
@@ -0,0 +1,57 @@
...
+
+static PyTypeObject py_log_template_options_type =
+{
+  PyVarObject_HEAD_INIT(&PyType_Type, 0)
+  .tp_name = "LogTemplateOptions",
+  .tp_basicsize = sizeof(PyLogTemplateOptions),
+  .tp_dealloc = (destructor) PyObject_Del,
+  .tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
+  .tp_doc = "LogTemplateOptions class encapsulating a syslog-ng LogTemplateOptions",
+  .tp_new = PyType_GenericNew,
```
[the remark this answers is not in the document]
furiel: Thanks I agree and fix it.

In the future, we might want to allow to create LogTemplateOptions, and then I will add the correct tp_new vfunction. Btw users still cannot instantiate it even with PyType_GenericNew, because the class object is not added to any namespace, so the name itself is not available. Though still better use NULL.

furiel:
> It might be trivial, but I think we should mention that self.log_template.format(msg) returns the result as a string.

Thanks, I updated the description.

modules/python/python-logtemplate.c
```diff
@@ -0,0 +1,163 @@
...
+
+PyTypeObject py_log_template_type;
+PyObject *PyExc_LogTemplate;
+
+void
+py_log_template_free(PyLogTemplate *self)
+{
+  log_template_unref(self->template);
+  g_free(self->template_options);
+  PyObject_Del(self);
```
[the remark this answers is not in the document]
MrAnno: If I'm not mistaken:

```diff
-PyObject_del(self);
+Py_TYPE(self)->tp_free((PyObject*) self);
```

furiel:
Thanks for the thorough review. Yes I think that would be useful for python parser too, or maybe even for python source. Though I do not not plan to work on it in the near future.

kira-syslogng:
Build SUCCESS

Issue #1761, opened 2017-11-16
python destination: API needs to be discussed


furiel:
I start with a quick summary of the threaded destination driver message handling in terms of flow control (you can skip if you are familiar with the insert method error codes).
Currently threaded destination driver provides the following possible behaviour:
- _success_: all went well, message handled correctly, message acknowledged back to core <-> python send method returns with True
- _temporary error_: some temporary error happened, message put back to queue, syslog-ng tries to resend it retries() number of times, waiting time-reopen() in-between <-> send method returns with false
- _not connected_: there is nothing wrong with the message, but the message still could not be processed due to some other persistent problem (for example connection lost towards a database), in this case message is put back to the queue, and threaded destination driver tries to reopen connection calling thrdestdrv->connect method infinitely, until reopen successful <-> this is (almost, but I will get to that later) achievable if python send sets some side effect (the side effect will be queried with is_opened method), then send should return with false.
- _drop_: there is some permanent unrecoverable problem with the message, the message should be instantly dropped (for example missing fields, malformed data, invalid data), the message should be instantly dropped.

I identified the following problems with the current python API for user code.

- WORKER_INSERT_RESULT_DROP is not supported.
Currently there is no way to express the drop case. One might suggest the user could write the send method that just does not handle a message, and return with True (returning false would result in retry with the same message). But returning True from send will mess up the dropped statistics counter.

- WORKER_INSERT_RESULT_NOT_CONNECTED
This bullet can possibly fixed in the code, but I still want to mention here, because it arrives from the complexity of the python API.
The only way to trigger this is to set some side effect (some class-local variable that expresses we are not connected anymore) by send method, and return with False. (The side effect will be queried by is_opened() later) Now the problem is return False overlaps with WORKER_INSERT_RESULT_ERROR. In case retries(1) is used: no retry will happen. (It is strange for me 1 means zero, nonetheless it is not python specific, so I just note here).
In that case the message is instantly dropped, and the open/close method is not activated. The expected behaviour is that the message should be put back to the queue, and try resend after successful connect.
If retries > 2, then this seems to work.

- is_opened method
Threaded destination driver is capable of tracking the connected (or opened in python dest terminology) state of the destination: there is a gboolean self->worker.connected for this purpose.
But python destination does not utilize this feature: the connect method is not overridden, only the disconnect. As a result, users themselves need to reimplement this logic: using is_opened method. Users need to store the opened state inside the python class, which must be updated by open, close and even send methods as side effect.
[...]

PR #2197, opened 2018-07-20
[wip] Rename onmessagequeueempty to flush in java destinations

lbudai:
This should merge only after #2063 is merged.

### DIFF
--- a/modules/java/src/main/java/org/syslog_ng/LogDestination.java
+++ b/modules/java/src/main/java/org/syslog_ng/LogDestination.java
@@ -56,8 +56,8 @@ public abstract class LogDestination extends LogPipe {
 
 	private native int getSeqNum(long ptr);
 
-	protected void onMessageQueueEmpty() {
-		return;
+	protected boolean flush() {
+		return true;
 	}
 
 	public boolean openProxy() {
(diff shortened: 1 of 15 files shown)


furiel:
The patch itself looks good to me, but it feels incomplete. In a previous version of #2063 there was an additional condition before flush in logthrdestdrv.c: batch size must not be empty. I think we should add this to make flush api complete: no reason to flush on an empty batch. 

Though that introduces another problem. As currently the language binding inserts return with boolean (we are facing #1761 again) we effectively disable batch insert: either a message is dropped or accepted, no way to mark as queued. With onEmpty behaviour, users could simulate batch sending by accepting each message, save them inside the binding, then flush with the onEmpty notification. Now there is no convenient way to flush, except counting inside the binding + timeout.

In short, language bindings expect onEmpty api until #1761.

kira-syslogng:
Build FAILURE

lbudai:
@furiel: Thanks for the note. My plan is to discuss this topic on next OSE handover.

kira-syslogng:
Build FAILURE

modules/java/src/main/java/org/syslog_ng/LogDestination.java
```diff
@@ -56,8 +56,8 @@ public int getSeqNum() {
 
 	private native int getSeqNum(long ptr);
 
-	protected void onMessageQueueEmpty() {
-		return;
+	protected boolean flush() {
```
Kokan: I understand that we want to change the interface here, and without deprecating the old one. (I think we could even get away with it.)
But if we accept this change, does this change enough ?
Are we going to feel that the java interface is good enough for at least mid term ?
I think we should hurt the user once, and not continuously.

lbudai:
@kokan: see my last comment ;)
my aim with this patch set was to point out some problems, including this one.
after batching has been merged, we need to figure out the next steps.

PR #2199, opened 2018-07-23
Introduce syslog-ng() source

jszigetvari:
scl/ewmm/ewmm.conf: added definition of syslog-ng() source

    * introduced the syslog-ng() source for ewmm formatted logs
    * originally only the syslog-ng() destination was defined here, but in
      most cases sources and destinations arrive in pairs, so I just
      implemented that

Signed-off-by: Janos SZIGETVARI <jszigetvari@gmail.com>

### DIFF
--- a/scl/ewmm/ewmm.conf
+++ b/scl/ewmm/ewmm.conf
@@ -65,3 +65,23 @@ block destination syslog-ng(server('127.0.0.1') transport(tcp) port(514) ...) {
 		`__VARARGS__`
         );
 };
+
+block source syslog-ng(
+        ip('0.0.0.0')
+        port(514)
+        transport(tcp)
+        flags("")
+        ...) {
+
+        channel {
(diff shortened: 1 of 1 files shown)


kira-syslogng:

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build FAILURE

Kokan:
@syslog-ng retest this please

kira-syslogng:
Build SUCCESS

gaborznagy:
I prefer the current implementation over the above mentioned "no-parse" variant.
I think syslog-ng() driver should expect the ewmm format explicitly and shouldn't allow other formats.
For that functionality the `default-network-drivers()` driver is preferred.

Besides, syslog-parser injects an error message upon parsing error rather than dropping the message.

PR #2204, opened 2018-07-24
client-lib-dir, loggen, eventlog: minor fixes

MrAnno:
- elastic, hdfs, kafka: make client-lib-dir() optional
- loggen: fix strncpy compiler warning
- eventlog: use long long parameter for evt_tag_long()

### DIFF
--- a/tests/loggen/loggen_helper.c
+++ b/tests/loggen/loggen_helper.c
@@ -148,7 +148,15 @@ int connect_unix_domain_socket(int sock_type, const char *path)
 
   DEBUG("unix domain socket: %s\n",path);
   saun.sun_family = AF_UNIX;
-  strncpy(saun.sun_path, path, sizeof(saun.sun_path));
+
+  gsize max_target_path_size = sizeof(saun.sun_path);
+  if (strlen(path) >= max_target_path_size)
+    {
+      ERROR("Target path is too long; max_target_length=%" G_GSIZE_FORMAT "\n", max_target_path_size - 1);
+      return -1;
+    }
+
+  strcpy(saun.sun_path, path);
 
   dest_addr = (struct sockaddr *) &saun;
   dest_addr_len = sizeof(saun);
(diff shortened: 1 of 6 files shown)


kira-syslogng:
Build SUCCESS

tests/loggen/loggen_helper.c
```diff
@@ -148,7 +148,15 @@ int connect_unix_domain_socket(int sock_type, const char *path)
 
   DEBUG("unix domain socket: %s\n",path);
   saun.sun_family = AF_UNIX;
-  strncpy(saun.sun_path, path, sizeof(saun.sun_path));
+
+  gsize max_target_path_size = sizeof(saun.sun_path);
+  if (strlen(path) >= max_target_path_size)
```
Kokan: Just a few thoughts out-load, consider it as a note (this is not a change request):
This is not equivalent in a sense that previous code handled not null terminated `path`, and this is not.
Anyhow neither version handles the fact that the `sun_path` shall be null terminated :)

MrAnno: That's true, thanks for pointing out :) Relying on the fact that `path` is coming from `argv[1]` is not the best idea, so a forced NULL-termination would be a good idea.

MrAnno: Sorry, I was wrong. The NULL-termination was never handled, `strncpy` won't terminate a string if `src` is not terminated.

MrAnno: We agreed that the current state of this PR is good enough, it is assumed that `path` is a string (null-terminated).

bazsi:
Just a note, iirc sun_path is not required to be NUL terminated.

MrAnno:
Linux man pages specify sun_path to be null-terminated.

> The pathname in sun_path should be null-terminated.

Exceptions are the abstract (it starts with a \0) and unnamed sockets.

PR #2207, opened 2018-07-26
RFC: m4/check_java_support: add -rpath to libmod-java so it finds libjvm.so

bazsi:
This is primarily an RFC pull request, with this we would be encoding the path to libjvm.so into our libmod-java.so plugin, avoiding having to set LD_LIBRARY_PATH at runtime.

I think it is pretty common to run into java() being unavailable because of the missing LD_LIBRARY_PATH setting.

The risk is that not all platforms might support rpath the same. Also, Debian has a policy against rpath, but I think it does not apply (but I am not sure). @mochrul 

I haven't implemented the cmake variant.

What do you, especially package maintainers, think?
@czanik

### DIFF
--- a/m4/check_java_support.m4
+++ b/m4/check_java_support.m4
@@ -91,7 +91,7 @@ AC_DEFUN([AX_CHECK_JAVA_VERSION],
       JNI_LIBDIR=`find $JNI_HOME \( -name "libjvm.so" -or -name "libjvm.dylib" \) \
         | sed "s-/libjvm\.so-/-" \
         | sed "s-/libjvm\.dylib-/-" | head -n 1`
-      JNI_LIBS="-L$JNI_LIBDIR -ljvm"
+      JNI_LIBS="-L$JNI_LIBDIR -ljvm -Wl,-rpath -Wl,$JNI_LIBDIR"
       JNI_INCLUDE_DIR=`find $JNI_HOME -name "jni.h" |  sed "s/\(.*\)jni.h/\1/" | head -n 1`
       JNI_CFLAGS="-I$JNI_INCLUDE_DIR"
 


kira-syslogng:
Build SUCCESS

lbudai:
@bazsi : could work, when building from source 

From packages point of view: what if the distro provides more than one java version? libjvm.so is not handled by the alternatives in Debian/Ubuntu, so it is easy to build a binary that contains an rpath pointing to a different(, or even nonexistent jvm.so...) which can be sometimes an issue. Or, in other words, it will only work when the builder environment and the runtime environment has the same java version (and distribution) installed... so if the user has an oracle jre, but we are building syslog-ng with openjdk, then rpath will points to nowhere on the client machine.

czanik:
@bazsi Yeah, would be fantastic. But as soon as the user chooses a different Java vendor or simply just updates the package containing libvjm.so, it will stop working.

bazsi:
Are different libjvm versions indeed compatible? e.g. can I actually use Oracle 8 for a binary that was compiled with Oracle 7? Do the same apply with OpenJDK?

czanik:
Yes. At least packages built with OpenJDK 7 work with OpenJDK 8, and AFAIR also with Oracle JDK (but RPM package dependencies force OpenJDK to be installed, even if not used). JDK 9+ needs some compatibility fixes, which is AFAIR still WIP.

MrAnno:
But the `JNI_LIBDIR` of OpenJDK 7 and OpenJDK 8 are different even on the same platform.

bazsi:
As I see theres a java-wrapper package for writing shell scripts that
invoke java apps with properly set LD_LIBRARY_PATH settings.

Anyhow this patch will not fly, so I am closing the pr.

PR #2209, opened 2018-07-26
scl/ewmm: rename syslog-ng() source to ewmm()

bazsi:
syslog-ng is just an alias for the ewmm destination there's no syslog-ng()
source and shouldn't be one, even if it sounds asymmetrical.

Rationale:
  - on the source side we want ewmm to be processed on the same channel as
    everything else is coming in.
  - we already created default-network-drivers() for exactly this use-case
  - the ewmm() source below is a pretty limited use-case that is probably
    only good enough for testing.
  - with this, we have symmetry: ewmm() source and destination. And we
    also reserved the name "syslog-ng()" to send messages to an actual
    syslog-ng instance.

if someone needs a different use-case, the recommended way is to construct
the source plus parsing bits at the configuration file level, as that
use-case is not generic enough.

### DIFF
--- a/scl/ewmm/ewmm.conf
+++ b/scl/ewmm/ewmm.conf
@@ -58,15 +58,38 @@ block parser ewmm-parser() {
 
 template-function "format-ewmm" "<$PRI>1 $ISODATE $LOGHOST @syslog-ng - - ${SDATA:--} $(format-json --leave-initial-dot --scope all-nv-pairs --exclude 0* --exclude 1* --exclude 2* --exclude 3* --exclude 4* --exclude 5* --exclude 6* --exclude 7* --exclude 8* --exclude 9* --exclude SOURCE --exclude .SDATA.* ._TAGS=${TAGS})\n";
 
-block destination syslog-ng(server('127.0.0.1') transport(tcp) port(514) ...) {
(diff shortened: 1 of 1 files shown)


bazsi:
@jszigetvari I am basically undoing your syslog-ng() source with this patch, please have a look at the rationale I have added to the file as a comment and let me know if you have issues with this patch.

Thanks in advance,

kira-syslogng:
Build SUCCESS

lbudai:
Talked to @jszigetvari and he will react to this PR soon.
Note to other reviewers: until that we should keep it open.

jszigetvari:
@bazsi Originally there was no separate ewmm() destination, but with this change the symmetry has been restored.

IMHO the default-network-drivers() source has a much wider feature set, than what one needs in case of simply wanting to forward EWMM logs.
The solution of having a dedicated evmm() source enables users to have much more control over their syslog-ng instance.
For example, if we look at a relay use-case which forwards the logs in EVMM format, then on the remote destination we simply create an evmm() source, which only does what it's supposed to do, as opposed to the default-network-drivers() source.

So the bottom like it, that I am okay with the way you changed my original contribution.

PR #2213, opened 2018-08-01
diskq: dqtool cat print backlog items

Kokan:
The `dqtool` currently does not print out the backlog items from the disk queue file.
This patch allow `dqtool` to use the backlog, and rewinds messages (only in memory), so `dqtool` could access them.


kira-syslogng:
Build SUCCESS

PR #2214, opened 2018-08-01
WIP: Rewind backlog in case of stateless LogProtoClient

Kokan:
There was an original fix (#2192) for reliable disk buffer. That solution focused only on reliable disk buffer.
With this change also the non-reliable disk buffer and logqueue fifo is going to be fixed, with an additional change. If there is a stateful LogProtoClient the backlog rewind actually is not needed, because the the Proto has the knowledge about the queue state, and it can rewind itself.


kira-syslogng:
Build SUCCESS

Kokan:
@lbudai Thanks for the review. I totally agree on *default* instead of *preferred*, so I updated only the comment of that commit.

kira-syslogng:
Build SUCCESS

PR #2215, opened 2018-08-02
[WIP]Resolve circular dep ack tracker logsource

lbudai:
This was a side effect of another work, but it can be handled separately from that.

@MrAnno : I know that we need to follow up these changes on some internal branches where we have a session cache which is based on the fact that we store a pointer to a `LogSource` instance in the `AckTracker`...

### DIFF
--- a/lib/ack_tracker.h
+++ b/lib/ack_tracker.h
@@ -26,7 +26,12 @@
 #define ACK_TRACKER_H_INCLUDED
 
 #include "syslog-ng.h"
-#include "logsource.h"
+#include "ack-type.h"
+#include "bookmark.h"
+
+typedef struct _AckTracker AckTracker;
+typedef struct _AckRecord AckRecord;
+typedef struct _LogSource LogSource;
 
 struct _AckTracker
 {
--- a/lib/ack_tracker.h
+++ b/lib/ack_tracker.h
@@ -35,6 +35,7 @@ struct _AckTracker
   Bookmark *(*request_bookmark)(AckTracker *self);
   void (*track_msg)(AckTracker *self, LogMessage *msg);
   void (*manage_msg_ack)(AckTracker *self, LogMessage *msg, AckType ack_type);
+  void (*free_fn)(AckTracker *self);
 };
 
 struct _AckRecord
@@ -45,18 +46,12 @@ struct _AckRecord
 AckTracker *late_ack_tracker_new(LogSource *source);
 AckTracker *early_ack_tracker_new(LogSource *source);
 
-void late_ack_tracker_free(AckTracker *self);
-void early_ack_tracker_free(AckTracker *self);
-
 static inline void
 ack_tracker_free(AckTracker *self)
 {
-  if (self)
+  if (self && self->free_fn)
     {
-      if (self->late)
-        late_ack_tracker_free(self);
-      else
-        early_ack_tracker_free(self);
+      self->free_fn(self);
     }
 }
 
--- a/lib/ack_tracker.h
+++ b/lib/ack_tracker.h
@@ -34,7 +34,6 @@ typedef struct _AckRecord AckRecord;
 
 struct _AckTracker
 {
-  gboolean late;
   Bookmark *(*request_bookmark)(AckTracker *self);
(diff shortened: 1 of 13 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/early_ack_tracker.c
```diff
@@ -52,50 +51,52 @@ static void
...
   self->super.request_bookmark = early_ack_tracker_request_bookmark;
   self->super.track_msg = early_ack_tracker_track_msg;
   self->super.manage_msg_ack = early_ack_tracker_manage_msg_ack;
+  self->super.free_fn = early_ack_tracker_free;
+}
+
+static void
+early_ack_tracker_init_instance(EarlyAckTracker *self)
+{
+  self->super.late = FALSE;
```
furiel: This hack is not needed anymore due to the previous patch. The `late` field can be eliminated.

lib/early_ack_tracker.c
```diff
@@ -70,6 +70,14 @@ early_ack_tracker_manage_msg_ack(AckTracker *s, LogMessage *msg, AckType ack_typ
   log_pipe_unref((LogPipe *)self->super.source);
 }
 
+static void
+early_ack_tracker_free(AckTracker *s)
+{
+  EarlyAckTracker *self = (EarlyAckTracker *)s;
+
+  g_free(self);
+}
```
bazsi: g_free() is usually part of the inline function, so that allocation policy is not encoded in the free function.

lib/ack_tracker.h
```diff
@@ -26,7 +26,12 @@
 #define ACK_TRACKER_H_INCLUDED
 
 #include "syslog-ng.h"
-#include "logsource.h"
+#include "ack-type.h"
+#include "bookmark.h"
+
+typedef struct _AckTracker AckTracker;
+typedef struct _AckRecord AckRecord;
+typedef struct _LogSource LogSource;
```
bazsi: why isn't this in logsource.h? this file includes that anyway and seems like a better fit.

bazsi: seems you are already doing in a subsequent patch, sorry for the noise.

lib/early_ack_tracker.c
lbudai: indeed...

lib/early_ack_tracker.c
lbudai: thanks, will fix

kira-syslogng:
Build FAILURE

lib/early_ack_tracker.c
bazsi: that's why we usually have _new() functions (which contains the g_new call) and in case we want to change the allocation policy, we also have an init_instance() function, which does initialization in the face of an existing memory block.

the existing convention is to put the g_free() in the inline function, this way it's less likely to be forgotten as further derived classes are added (as the free is always called regardless of the virtual function).

we could change the convention, but I don't think it is better the other way.

lbudai: @bazsi : I agree when we are talking about base classes/abstract classes, but not in case of interfaces...  there is no data member in the interface which means that every allocations for data members need to be done in the `init` functions. But it's just a matter of taste, I think (as we are talking about mimicking OOP in C).

lbudai:
@bazsi: thanks for the patch and for the note, I'll check it in details soon (I also hated to add the `LogSource*` to the `LogMessage` :) ).

The ack func with data concept could work (as a function object) of course, but this is just a technical trick to hide the fact: the source need to be kept alive until the last message is not acked, which means that there is a dependency between the two.

If the only issue is the techn. dependency (adding a  `LogSource *` to the `LogMessage`), then maybe we could factor out a class that represents the flow control... and then we could include only that part of the LogSource. Are you ok with this idea?

lbudai:
I'm closing this and will solve it later, as part of the memory limited flow control PR.

PR #2216, opened 2018-08-02
test_python: flush before stopping syslog-ng

lbudai:
Otherwise it is possible that not every message received (eg.: when
start syslog-ng with 'Fevdt' in control.py).


kira-syslogng:
Build SUCCESS

PR #2218, opened 2018-08-04
Decorating generated configuration

Kokan:
There are configuration features, that in the background generates additional configuration. They are pretty cool, but sometimes hard to debug as a user.

A good example is the `system` source, there is a script to print out what is generated emplace of the `system` source: `modules/system-source/system-expand.sh`

The result is going to be the following: (good luck with that)
```

channel {
    source {
channel {
    source { unix-dgram("/dev/log" so_rcvbuf(8192));
 };
    rewrite { set("${.unix.pid}" value("PID") condition("${.unix.pid}" ne "")); };
};
file("/dev/kmsg" program-override("kernel") flags(kernel) format(linux-kmsg) keep-timestamp(no));
    }; # source
channel {
  channel {
    parser {
       
channel {
    junction {
    };
}
;
       
channel {
    junction {
channel {
    parser {  json-parser(prefix('.cim.') marker('@cim:'));  };
    rewrite { set-tag('.app.cim'); };
    flags(final);
};
channel {
    filter {  facility(kern) and
                 program("kernel" type(string)) and
                 message("PROTO=" type(string) flags(substring));  };
    parser {  
	kv-parser(prefix('.iptables.'));
;  };
    rewrite { set-tag('.app.iptables'); };
    flags(final);
};
channel {
    filter {  program("sudo" type(string));  };
    parser {  
    channel {
        parser {
            kv-parser(prefix('.sudo.') pair-separator(';') extract-stray-words-into('0'));
            csv-parser(columns(".sudo.SUBJECT") template("$(list-head $0)") delimiters(' '));
        };
	# drop everything which does not have a command (to exclude pam
	# related logs)
        filter { not match("" value(".sudo.COMMAND") type(string)); };
    };
;  };
    rewrite { set-tag('.app.sudo'); };
    flags(final);
};
    };
}
;
    };
    flags(final);
  };
  channel { flags(final); };
};
}; # channel
;
```
This modification actually decorates with **Start** and **End** in case of generated configuration, so it helps to identify the origin of the lines:

The above example after the patch:
```## system() expands to:

#Start Block source generator system
channel {
    source {
channel {
    source { unix-dgram("/dev/log" so_rcvbuf(8192));
 };
    rewrite { set("${.unix.pid}" value("PID") condition("${.unix.pid}" ne "")); };
};
file("/dev/kmsg" program-override("kernel") flags(kernel) format(linux-kmsg) keep-timestamp(no));
    }; # source
channel {
  channel {
    parser {
       
#Start Block parser generator app-parser
channel {
    junction {
    };
}
#End Block parser generator app-parser
;
       
#Start Block parser generator app-parser
channel {
    junction {
#Start Application cim
channel {
    parser {  json-parser(prefix('.cim.') marker('@cim:'));  };
    rewrite { set-tag('.app.cim'); };
    flags(final);
};
#End Application cim
#Start Application iptables
channel {
    filter {  facility(kern) and
                 program("kernel" type(string)) and
                 message("PROTO=" type(string) flags(substring));  };
    parser {  
#Start Block block parser iptables-parser() at /tmp/install/share/syslog-ng/include/scl/iptables/iptables.conf:23
	kv-parser(prefix('.iptables.'));
#End Block block parser iptables-parser() at /tmp/install/share/syslog-ng/include/scl/iptables/iptables.conf:23
;  };
    rewrite { set-tag('.app.iptables'); };
    flags(final);
};
#End Application iptables
#Start Application sudo
channel {
    filter {  program("sudo" type(string));  };
    parser {  
#Start Block block parser sudo-parser() at /tmp/install/share/syslog-ng/include/scl/sudo/sudo.conf:23
    channel {
        parser {
            kv-parser(prefix('.sudo.') pair-separator(';') extract-stray-words-into('0'));
            csv-parser(columns(".sudo.SUBJECT") template("$(list-head $0)") delimiters(' '));
        };
	# drop everything which does not have a command (to exclude pam
	# related logs)
        filter { not match("" value(".sudo.COMMAND") type(string)); };
    };
#End Block block parser sudo-parser() at /tmp/install/share/syslog-ng/include/scl/sudo/sudo.conf:23
;  };
    rewrite { set-tag('.app.sudo'); };
    flags(final);
};
#End Application sudo
    };
}
#End Block parser generator app-parser
;
    };
    flags(final);
  };
  channel { flags(final); };
};
}; # channel
#End Block source generator system
;
```

### DIFF
--- a/lib/cfg-block-generator.c
+++ b/lib/cfg-block-generator.c
@@ -38,7 +38,14 @@ gboolean
 cfg_block_generator_generate(CfgBlockGenerator *self, GlobalConfig *cfg, CfgArgs *args, GString *result,
                              const gchar *reference)
 {
-  return self->generate(self, cfg, args, result, reference);
+  gchar block_name[1024];
+  cfg_block_generator_format_name(self, block_name, sizeof(block_name)/sizeof(block_name[0]));
+
+  g_string_append_printf(result, "\n#Start Block %s\n", block_name);
+  const gboolean res = self->generate(self, cfg, args, result, reference);
+  g_string_append_printf(result, "\n#End Block %s\n", block_name);
+
+  return res;
 }
 
 void
--- a/modules/appmodel/tests/test_app_parser_generator.c
+++ b/modules/appmodel/tests/test_app_parser_generator.c
@@ -77,11 +77,48 @@ _build_cfg_args(const gchar *key, const gchar *value, ...)
   return args;
 }
 
+
+static GString *
+_remove_comments(const GString *str)
+{
+  GString *prune = g_string_new("");
+  gboolean is_inside_comment = FALSE;
+
+  for (gint i=0; i < str->len; ++i)
+    {
+      const gchar curr = str->str[i];
+      const gboolean is_comment_start = '#' == curr;
+      const gboolean is_comment_end = curr == '\n';
+      const gboolean does_comment_continue_next_line = is_comment_end && (i+1) < str->len && '#' == str->str[i+1];
+
+      if (is_comment_start || does_comment_continue_next_line)
+        {
+          is_inside_comment = TRUE;
+        }
+
+      if (!is_inside_comment)
+        {
+          g_string_append_c(prune, str->str[i]);
+        }
+
+      if (is_comment_end)
+        {
+          is_inside_comment = FALSE;
+        }
+
+    }
+
+  return prune;
+}
+
 static void
 _app_parser_generate_with_args(const gchar *topic, CfgArgs *args)
 {
+  GString *tmp = g_string_new("");
   cfg_args_set(args, "topic", topic);
-  cfg_block_generator_generate(app_parser, configuration, args, result, "dummy-reference");
+  cfg_block_generator_generate(app_parser, configuration, args, tmp, "dummy-reference");
+  result = _remove_comments(tmp);
+  g_string_free(tmp, TRUE);
   cfg_args_unref(args);
 }
 
--- a/lib/cfg-block.c
+++ b/lib/cfg-block.c
@@ -178,7 +178,7 @@ cfg_block_generate(CfgBlockGenerator *s, GlobalConfig *cfg, CfgArgs *args, GStri
       return FALSE;
     }
 
-  g_string_assign_len(result, value, length);
+  g_string_append_len(result, value, length);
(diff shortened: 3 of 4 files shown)


kira-syslogng:
Build SUCCESS

modules/appmodel/tests/test_app_parser_generator.c
```diff
@@ -77,11 +77,48 @@ _build_cfg_args(const gchar *key, const gchar *value, ...)
...
+{
+  GString *prune = g_string_new("");
+  gboolean is_inside_comment = FALSE;
+
+  for (gint i=0; i < str->len; ++i)
+    {
+      const gchar curr = str->str[i];
+      const gboolean is_comment_start = '#' == curr;
+      const gboolean is_comment_end = curr == '\n';
+      const gboolean does_comment_continue_next_line = is_comment_end && (i+1) < str->len && '#' == str->str[i+1];
```
furiel: Why do we need this variable? To join the before-after lines of a comment block?

lib/cfg-block-generator.c
```diff
@@ -38,7 +38,14 @@ gboolean
 cfg_block_generator_generate(CfgBlockGenerator *self, GlobalConfig *cfg, CfgArgs *args, GString *result,
                              const gchar *reference)
 {
-  return self->generate(self, cfg, args, result, reference);
+  gchar block_name[1024];
+  cfg_block_generator_format_name(self, block_name, sizeof(block_name)/sizeof(block_name[0]));
```
furiel: Just a minor: [G_N_ELEMENTS](https://developer.gnome.org/glib/stable/glib-Standard-Macros.html#G-N-ELEMENTS:CAPS) could be used.

modules/appmodel/tests/test_app_parser_generator.c
Kokan: The aim is to remove the `\n` from *multi line* comment blocks like this:
```
#Comment1\n
#Comment2\n
log {};
```

Without this it would became:
```
\n
\n
log {};
```

I did not want to change the assert, but rather remove what not needed.
This is actually not perfect, because it would detect as a comment in strings: `file('/tmp/#bla');'

But I would rather avoid parsing, and would require from UT writer to keep that in mind.

PR #2220, opened 2018-08-07
debugger: fix template function evaluation

bazsi:
The debugger runs in its dedicated thread, so we need to initialize it
as a thread so stuff like template function evaluation works, since
it depends on scratch buffers.


kira-syslogng:
Build SUCCESS

PR #2223, opened 2018-08-10
Base64 encoding template function

bazsi:
This patch adds $(base-encode ...) template function, which encodes
the concatenated arguments using base64. No delimiters are added
between strings, so:
    
$(base64-encode foo bar) would encode "foobar".

Originally contributed by Ferenc Sipos <ferenc.sipos@oneidentity.com>

### DIFF
--- a/modules/basicfuncs/basic-funcs.c
+++ b/modules/basicfuncs/basic-funcs.c
@@ -106,6 +106,7 @@ static Plugin basicfuncs_plugins[] =
   TEMPLATE_FUNCTION_PLUGIN(tf_template, "template"),
   TEMPLATE_FUNCTION_PLUGIN(tf_urlencode, "urlencode"),
   TEMPLATE_FUNCTION_PLUGIN(tf_urlencode, "url-encode"),
+  TEMPLATE_FUNCTION_PLUGIN(tf_base64encode, "base64-encode")
 };
 
 gboolean
--- a/modules/basicfuncs/basic-funcs.c
+++ b/modules/basicfuncs/basic-funcs.c
@@ -104,7 +104,8 @@ static Plugin basicfuncs_plugins[] =
   /* misc funcs */
   TEMPLATE_FUNCTION_PLUGIN(tf_env, "env"),
   TEMPLATE_FUNCTION_PLUGIN(tf_template, "template"),
-  TEMPLATE_FUNCTION_PLUGIN(tf_urlencode, "urlencode")
+  TEMPLATE_FUNCTION_PLUGIN(tf_urlencode, "urlencode"),
+  TEMPLATE_FUNCTION_PLUGIN(tf_urlencode, "url-encode"),
 };
 
 gboolean
--- a/modules/basicfuncs/str-funcs.c
+++ b/modules/basicfuncs/str-funcs.c
@@ -597,6 +597,27 @@ tf_base64encode(LogMessage *msg, gint argc, GString *argv[], GString *result)
                              &state, &save);
     }
   g_string_set_size(result, init_len + out_len + _get_base64_encoded_size(0));
+
+#if !GLIB_CHECK_VERSION(2, 54, 0)
+  /* NOTE: this is an ugly workaround for glib versions < 2.54 (which is
+   * pretty recent and not widely available yet) to fix an encoding issue.
+   *
+   * This is the bug:
+   *    https://bugzilla.gnome.org/show_bug.cgi?id=780066
+   *
+   * This is the fix:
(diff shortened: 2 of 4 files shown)


modules/basicfuncs/basic-funcs.c
```diff
@@ -104,7 +104,8 @@ static Plugin basicfuncs_plugins[] =
   /* misc funcs */
   TEMPLATE_FUNCTION_PLUGIN(tf_env, "env"),
   TEMPLATE_FUNCTION_PLUGIN(tf_template, "template"),
-  TEMPLATE_FUNCTION_PLUGIN(tf_urlencode, "urlencode")
+  TEMPLATE_FUNCTION_PLUGIN(tf_urlencode, "urlencode"),
+  TEMPLATE_FUNCTION_PLUGIN(tf_urlencode, "url-encode"),
```
furiel: Users might got used to that hyphens and underscores are interchangeable. Should we support "url_encode" as well?

modules/basicfuncs/str-funcs.c
```diff
@@ -570,3 +570,36 @@ tf_binary_free_state(gpointer s)
...
+  gint save = 0;
+  gsize out_len = 0;
+  gsize init_len = result->len;
+
+  for (i = 0; i < argc; i++)
+    {
+      /* expand the buffer and add space for the base64 encoded string */
+      g_string_set_size(result, init_len + out_len + _get_base64_encoded_size(argv[i]->len));
+      out_len +=
+        g_base64_encode_step((guchar *) argv[i]->str, argv[i]->len,
```
furiel: Why do we use g_base64_encode_step instead of g_base64_encode? To avoid memory allocation?

modules/basicfuncs/basic-funcs.c
bazsi: no, the reason _ and - is interchangeable as it wasn't consistent in the early years. the documented/preferred form is always the dash based one. (btw we could issue a warning if that's not the case).

When the first template functions were introduced, the dash based convention was already present, so I wouldn't want to publish these with underscores as well.

modules/basicfuncs/str-funcs.c
bazsi: yes.

modules/basicfuncs/basic-funcs.c
furiel: Ok I see.

modules/basicfuncs/tests/test_basicfuncs.c
```diff
@@ -184,6 +184,15 @@ Test(basicfuncs, test_str_funcs)
   assert_template_format("$(binary 1 0x1)", "\1\1");
   assert_template_format("$(binary 0xFF 255 0377)", "\xFF\xFF\xFF");
   assert_template_format_with_len("$(binary 0xFF 0x00 0x40)", "\xFF\000@", 3);
+
+  assert_template_format("[$(base64-encode abc)]", "[YWJj]");
```
szemere: I am missing a test case where argc > 1 (this is why we have a loop inside the template function), because some might accidentally expect to get their parameters encoded one by one.
 - encode(apple) + encode(apple) `!=` encode(appleapple)

while
 - decode( encode(apple) + encode(apple) ) `==` decode (encode(appleapple) )

note: This only needs to be considered in cases where someone might want to handle the raw output of the encoding.


I also wrote two simple tests, to raise the attention to use proper escaping:
`assert_template_format("[$(base64-encode 'X  X')]", "[WCAgWA==]");`
`assert_template_format("[$(base64-encode X  X)]", "[WFg=]");`
`// $(base64-encode X  X) == $(base64-encode 'X'   'X') == $(base64-encode 'XX')`

modules/basicfuncs/str-funcs.c
```diff
@@ -570,3 +570,36 @@ tf_binary_free_state(gpointer s)
...
+                             FALSE /* break_lines */,
+                             result->str + init_len + out_len,
+                             &state, &save);
+    }
+  g_string_set_size(result, init_len + out_len + _get_base64_encoded_size(0));
```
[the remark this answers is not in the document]
bazsi: you might be right that the last allocation might not be needed, but only if we entered the loop. If we didn't, the +4 allocation might not have happened.

modules/basicfuncs/tests/test_basicfuncs.c
bazsi: Good point. I have added a testcase for the loop, thanks for the heads up.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
I have added an ugly workaround for glib < 2.54, but this actually should fix an invalid base64 issue.

PR #2224, opened 2018-08-10
early-ack-tracker: do not wakeup source when the ack type is SUSPENDED

lbudai:

### DIFF
--- a/lib/early_ack_tracker.c
+++ b/lib/early_ack_tracker.c
@@ -61,10 +61,10 @@ early_ack_tracker_manage_msg_ack(AckTracker *s, LogMessage *msg, AckType ack_typ
 {
   EarlyAckTracker *self = (EarlyAckTracker *)s;
 
+  log_source_flow_control_adjust(self->super.source, 1);
+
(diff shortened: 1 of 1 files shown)


lbudai:
Reproduction of the bug:

* listen with netcat (and use pv tool to make partial write happen)
```
  nc -l 61002 | pv -L 40
```

* run syslog-ng with the following config:
```
@version: 3.17
options {
   time_reopen(3);
   stats_level(1);
};
source s_network {
   network(ip("127.0.0.1")
   port(5555));
};

destination d_network {
   network("127.0.0.1" disk_buffer(mem-buf-size(1000) disk-buf-size(1048576) reliable(yes) dir('/tmp/dq'))
   ip_protocol(4)
   port(6666)
   so-sndbuf(4096));
};

log {
  source(s_network);
  destination(d_network);

  flags(flow-control);
};
```

* send logs
```
  loggen -r 100000 -I 200 127.0.0.1 10001 --inet
```

* wait a few seconds then stop netcat

* result: syslog-ng will drops all the incoming messages

PR #2226, opened 2018-08-12
Use msg_trace() when emitting trace information

MrAnno:
I think our debug log is flooded, so I've opened this PR to
- make it possible to enable trace messages with the `--trace` or `-t` command line flag, even if syslog-ng was not compiled in debug mode
- move frequent, per-message internal messages to the trace log level.

This won't have any performance implications, because we already have 166 debug messages compared to 10 existing trace messages.

My intention is to make our debug logs manageable in case there are more than 10-20 messages passing through syslog-ng. This way we can ask for debug logs from production environments without generating ~15-50 lines of debug information for each incoming message (15 using a trivial configuration, 50 with a non-trivial one).

But of course, when trace is required, it can be enabled without recompiling syslog-ng:
```sh
sbin/syslog-ng -Fdevt
# or
sbin/syslog-ng-ctl --set=1 trace
```

My guidelines I followed:
1. Leave all nonfrequent events on debug level, they are extremely useful.
2. Leave a few frequent (per-message) debug messages that are absolutely necessary for debugging.
   For example: source suspend/wakeup events, incoming/outgoing log entries, destination queue full events, the result of filters/parsers, throttling information, etc.
3. Move all other per-message "trace information" to the trace level.
   For example: starting/ending filters/parsers/rewrites/etc, setting an NV-pair in a message, ... 

Questions:
- **What do you think of the guidelines above?**
- **Please have a look at the output of `git grep msg_debug` and `git grep msg_trace` executed on my branch. Do you agree with those changes?**

### DIFF
--- a/lib/filter/filter-call.c
+++ b/lib/filter/filter-call.c
@@ -49,7 +49,7 @@ filter_call_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg)
   else
     stats_counter_inc(self->super.not_matched);
 
-  msg_debug("filter() evaluation started",
+  msg_trace("filter() evaluation started",
             evt_tag_str("called-rule", self->rule),
             evt_tag_printf("msg", "%p", msgs[num_msg - 1]));
 
--- a/lib/filter/filter-cmp.c
+++ b/lib/filter/filter-cmp.c
@@ -85,7 +85,7 @@ fop_cmp_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg)
       result = self->cmp_op & FCMP_GT || self->cmp_op == 0;
     }
 
-  msg_debug("cmp() evaluation started",
+  msg_trace("cmp() evaluation started",
             evt_tag_str("left", left_buf->str),
             evt_tag_str("operator", self->super.type),
             evt_tag_str("right", right_buf->str),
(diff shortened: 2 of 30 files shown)


MrAnno:
@kira-syslogng do perftest

MrAnno:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

szemere:
Forwarding the words of @mitzkia : I would move the following messages from debug to trace too:

`lib/parser/parser-expr.c:  msg_debug(">>>>>> parser rule evaluation begin",`
`lib/filter/filter-pipe.c:  msg_debug(">>>>>> filter rule evaluation begin",`
(Of course with the closing parts.)

MrAnno:
The result of rewrites/filters has been on debug lever for a very long time. Are you sure?

mitzkia:
I would say to put that messages also on trace because:
 - symmetry (same type of messages: ">>>>", "<<<<" can be appear on same level)
 - the debug log will be cleaner :)

MrAnno:
Okay, will do :)

MrAnno:

kira-syslogng:
Build SUCCESS

PR #2235, opened 2018-08-17
Hdfs: disable archive when append-enabled is configured

gaborznagy:
Archiving and append are 2 conflicting use cases.
Archiving is for moving closed file to a dir declaring that processing on that file have finished.
If we would allow appending an archived file it would not really be archived.

Originally I suspected a bug when the two options used simultaneously,
however the rename() call from hdfs lib prevents overwriting an existing file.
This patch fixes the current issue of archiving a file once, then using
the original file indefinitely.

### DIFF
--- a/modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
+++ b/modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
@@ -56,6 +56,7 @@ public class HdfsDestination extends StructuredLogDestination {
     private boolean isOpened;
     private int maxFileNameLength = 255;
     private boolean appendEnabled;
+    private String archiveDir;
 
     HashMap<String, HdfsFile> openedFiles;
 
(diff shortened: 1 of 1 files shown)


gaborznagy:
This affects documentation as well: I have discussed this change with @balabit-docs team.

gaborznagy:

kira-syslogng:
Build SUCCESS

gaborznagy:
The reasons to query archivedir() options only one are:
- it's unnecessary to query it every time it's needed, there is a point
where we call it 3 times in a row,
- the option is not going to change after we queried one time: you can only
set archivedir() option in config, and on reload we will load the new one
too, the change is equivalent the prev. version,
- this way it's also easier to disable archivedir() option and looks better
(I have thought other options, but there were not better),
- minimal performance improvement: options are stored in java_dd, so the
java side has to ask it everytime

lbudai:
@gaborznagy : maybe we could store the values "java init" time.

PR #2236, opened 2018-08-17
Fix threaded destination test cases

MrAnno:
1. The queued counter is decremented before calling `insert()` or `flush()`. Therefore, `queued_messages` is not suitable for conditional waiting in `test_explicit_ack_accept`.

2. `flush()` is called after every `_perform_inserts()` even when the batch size is 0.
    Because 3 batch-related test cases check whether the batch size remains the same across retries, `WORKER_INSERT_RESULT_SUCCESS` should be returned from `flush()` in case of an empty batch.
    
Reproduction: 
```sh
while lib/tests/test_logthrdestdrv --filter "logthrdestdrv/when_batched_*"; do :; done
```

From `_perform_flush`:

>  NOTE: earlier we had a condition on only calling flush() if batch_size is non-zero.  This was removed, as the language bindings that were done  _before_ the batching support in LogThreadedDestDriver relies on  flush() being called always, even if WORKER_INSERT_RESULT_SUCCESS is returned, in which case batch_size is already zero at this point.

### DIFF
--- a/lib/tests/test_logthrdestdrv.c
+++ b/lib/tests/test_logthrdestdrv.c
@@ -377,7 +377,7 @@ Test(logthrdestdrv, batched_set_of_messages_are_dropped_as_a_whole)
 }
 
 static inline void
-_assert_batch_size_remains_the_same_accross_retries(TestThreadedDestDriver *self)
+_assert_batch_size_remains_the_same_across_retries(TestThreadedDestDriver *self)
 {
   if (self->super.retries_counter > 0)
     cr_assert(self->super.batch_size == self->prev_flush_size);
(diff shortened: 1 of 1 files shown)


lbudai:
@MrAnno : WIP flag is because you are waiting for some internal tests?

MrAnno:
@lbudai I've found 2 additional bugs in this test case, so I have to fix them as well.
(One of them is our "release blocker" issue.)

kira-syslogng:
Build FAILURE

MrAnno:
@bazsi Why do we have to call `_perform_flush()` when `batch_size` is zero?

I've read your compatibility note about language bindings, but it still looks weird because one can return `WORKER_INSERT_RESULT_ERROR` from a zero-batched `flush()` call. 

Is it possible somehow to get rid of this?

kira-syslogng:
Build SUCCESS

MrAnno:

MrAnno:

kira-syslogng:
Build SUCCESS

bazsi:
I would love to, but it was requested by@furiel that we call that function
because we have legacy support for batching, when the language binding
relies on the fact that we always call onmessagequeueempty.

If we fix the language bindings we could get rid of the flush call in the
case of batch size being 0.

PR #2242, opened 2018-08-21
msg parameters: remove last NULL parameter from msg_ macros

bodnartibi:
msg_* macros put a NULL at the end of parameter list when call msg_event_suppress_recursions_and_send()


kira-syslogng:

Kokan:
@syslog-ng ok to test

bazsi:
We used to not have that NULL termination, and it seems we didn't remove all explicit NULLs or these messages were in-flight at the time.

Looks good to me.

Kokan:
@kira-syslogng ok to test

PR #2243, opened 2018-08-22
Processed timestamp

bodnartibi:
Now we can set our own timestamp.
It is useful when we want to book the time of the msg processing (for example the time of the serialization).


kira-syslogng:

lib/template/templates.h
```diff
@@ -93,14 +94,16 @@ struct _LogTemplateOptions
 void log_template_set_escape(LogTemplate *self, gboolean enable);
 gboolean log_template_set_type_hint(LogTemplate *self, const gchar *hint, GError **error);
 gboolean log_template_compile(LogTemplate *self, const gchar *template, GError **error);
-void log_template_format(LogTemplate *self, LogMessage *lm, const LogTemplateOptions *opts, gint tz, gint32 seq_num,
+void log_template_format(LogTemplate *self, LogMessage *lm, const LogTemplateOptions *opts, const LogStamp *processed,
+                         gint tz, gint32 seq_num,
                          const gchar *context_id, GString *result);
-void log_template_append_format(LogTemplate *self, LogMessage *lm, const LogTemplateOptions *opts, gint tz,
```
bazsi: I think the number of arguments passed here has surpassed the limit where we should encapsulate these into a struct and pass that.

We even had the beginnings of that in the form of LogTemplateInvokeArgs, that is only used during template function expansion, but it would be much better, if we extended its use into all call sites.

This would mean that even if we had further similar arguments, it would become a much smaller patch.

bazsi: I have this ancient branch, where I started this transition:

https://github.com/bazsi/syslog-ng/tree/wip/template-eval-arguments-as-a-struct-instead-of-parameters

I can't remember why I got distracted, maybe it was just time or I ran into some problems that I deemed was too difficult to solve.

bodnartibi: I have pushed some WIP commits.
I am not sure what would be the best.
- If LogTemplateEvalArgs has the same properties as in your branch, we could reduce the parameters of the log_template_append_format_with_context() and log_template_format_with_context() functions, and they looks good I guess.
- But in this case other functions in templates.h will have the same number of arguments as before, and it is ugly that we have similar functions with totally different argument list
- In the refactor commit I move messages and num_messages variables out of LogTemplateEvalArgs. In this case we can reduce all of these functions' parameters, and I guess it would look nice. I have not do that yet, because in this case we should modify so many files, and before every function call we have to create a structure filled with the parameters.
I am not sure what should I do, so I did finish non of these options.

kira-syslogng:
Build SUCCESS

lib/template/templates.h
bodnartibi: We agreed that the solution of this note needs a dedicated pull request.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bodnartibi:
I have fixed some issus:
- remove deadcode
- keep log_msg_serialize() function to have 2 parameters, and create log_msg_serialize_with_ts_processed() with extra parameter which is the processed timestamp (the first function is a wrapper for the other one)
- change timestamp_serialize() behavior (when it fills the timestamp gap). It is better to read the code than write it doen here
Please rereview it.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/logmsg/timestamp-serialize.c
```diff
@@ -54,11 +54,19 @@ _read_log_stamp(SerializeArchive *sa, LogStamp *stamp)
 
 
 gboolean
-timestamp_serialize(SerializeArchive *sa, LogStamp *timestamps)
+timestamp_serialize(SerializeArchive *sa, LogStamp *timestamps, const LogStamp *processed)
 {
+  const LogStamp *ts;
+  if (processed != NULL)
+    ts = processed;
+  else if (timestamps[LM_TS_PROCESSED].zone_offset == -1)
```
bodnartibi: a function is needed here to be readable

bodnartibi: is -1 valid here?

lib/template/macros.h
```diff
@@ -112,7 +113,8 @@ extern LogMacroDef macros[];
 
 /* low level macro functions */
 guint log_macro_lookup(const gchar *macro, gint len);
-gboolean log_macro_expand(GString *result, gint id, gboolean escape, const LogTemplateOptions *opts, gint tz,
+gboolean log_macro_expand(GString *result, gint id, gboolean escape, const LogTemplateOptions *opts,
+                          const LogStamp *processed, gint tz,
```
bodnartibi: revert the macro related stuffs

lbudai:
Some additional notes:
* The macro expansion can be removed (we talked about IRL).
* The way we are checking whether the processed timestamp is already set is not so good (we talked about IRL).
* The reason of passing an additional `LogStamp *processed` to the serialize function is that the messages are not cloned to each logpath... (and setting the processed timestamp would be racy).
This implementation is not something I'd merge... : processed timestamp is part of the message, but in some places we handle it as it wouldn't.

Two suggestions:
* maybe we could do a shallow-copy of the messages (in that case every destination could modifiy the timestamps -  I think it is not a bad thing if a destination is able to modify some parts of the message without cloning the whole nvtable), or, 
* store processed timestamp in the NVTable (note, this will result in a deep-copy of the NVtable)
* third option would be to clone the whole message... (could have huge performance impact)

kira-syslogng:
Build SUCCESS

lib/logmsg/logmsg-serialize.c
```diff
@@ -36,19 +36,40 @@
 
 #define OLD_LMM_REF_MATCH 0x0001
 
+static void
+_set_ts_processed(LogStamp *timestamps, const LogStamp *processed)
```
furiel: Optional: the naming is a little unfortunate, as this is not a set function. `_init_ts_processed` or `_setup_ts_processed` perhaps?

lib/logmsg/logmsg-serialize.c
```diff
@@ -36,19 +36,40 @@
...
+{
+  if (processed != NULL)
+    {
+      timestamps[LM_TS_PROCESSED].tv_sec = processed->tv_sec;
```
furiel: Optional: just simply `timestamps[LM_TS_PROCESSED] = *processed`?

lib/logmsg/logmsg-serialize.c
```diff
@@ -36,19 +36,40 @@
...
+      timestamps[LM_TS_PROCESSED].tv_usec = processed->tv_usec;
+      timestamps[LM_TS_PROCESSED].zone_offset = processed->zone_offset;
+    }
+  else if (timestamps[LM_TS_PROCESSED].zone_offset == LOGSTAMP_ZONE_OFFSET_UNSET)
+    {
+      timestamps[LM_TS_PROCESSED].tv_sec = timestamps[LM_TS_RECVD].tv_sec;
```
furiel: Optional: just simply `timestamps[LM_TS_PROCESSED]=timestamps[LM_TS_RECVD]`?

lbudai:
@bodnartibi , @Kokan : thanks!

PR #2244, opened 2018-08-22
compat/getent: add support for platforms that lack the _r versions of…

bazsi:
… proto/service lookup

This patch implements a limited version of the reentrant service/protocol
lookup functions from nss. It is limited so that not all fields are
available (e.g. the aliases array and the proto field).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -778,6 +778,7 @@ AC_CHECK_FUNCS(strdup		\
 	strcasestr		\
 	memrchr			\
 	localtime_r		\
+	getprotobynumber_r	\
 	gmtime_r		\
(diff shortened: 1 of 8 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
The reporter sent the following feedback:

> Syslog-ng 3.17.2 compiles fine on OpenWrt master with musl when the patch from pull request #2244 is applied. Running "syslog-ng --module-registry" no longer prints an error about loading the tfgetent module as it did with syslog 3.17.2 vanilla.

I have tried the associated functionality manually as getent seems to lack a unit test.

PR #2249, opened 2018-08-24
[WIP] kv-parser: stray word contained previouse kv pair

Kokan:
The stray words in wrongly detected, when the closing character of a quoted string did not followed by a key-value separator. The key-value pair was correct, but the stray word contained the previouse key-value pair as well.

Example:
```
kv-separator: [space]
Input: foo='qu'x asd
```

This PR also causes some regression in case of string parsing (as the function is shared), this break could be avoided with untangling the `str decode` and `kv parser`.

### DIFF
--- a/lib/str-repr/decode.c
+++ b/lib/str-repr/decode.c
@@ -75,6 +75,7 @@ typedef struct _StrReprDecodeState
 {
   GString *value;
   const gchar *cur;
+  const gchar *closing_delimiter;
   gchar quote_char;
   const StrReprDecodeOptions *options;
 } StrReprDecodeState;
@@ -161,6 +162,7 @@ _process_backslash_escaped_character_in_strings(StrReprDecodeState *state)
 static gint
 _process_delimiter_characters_after_a_quoted_string(StrReprDecodeState *state)
 {
+  state->closing_delimiter = state->cur;
   if (_match_and_skip_delimiter(state))
     return KV_FINISH_SUCCESS;
   return KV_QUOTE_ERROR;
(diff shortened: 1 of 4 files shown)


lib/str-repr/decode.c
```diff
@@ -238,6 +240,7 @@ str_repr_decode_append_with_options(GString *value, const gchar *input, const gc
   {
     .value = value,
     .cur = input,
+    .closing_delimiter = 0,
```
bazsi: this should probably be NULL instead of 0.

lib/str-repr/decode.c
```diff
@@ -161,6 +162,7 @@ _process_backslash_escaped_character_in_strings(StrReprDecodeState *state)
 static gint
 _process_delimiter_characters_after_a_quoted_string(StrReprDecodeState *state)
 {
+  state->closing_delimiter = state->cur;
```
bazsi: does this point to the closing apostrophe or rather it is the delimiter e.g. the comma?

bazsi:
This looks good to me, I just asked a question which I didn't understand without running the code.

Kokan:
A note about Kira failure, it is failing on *snmtp* source, I started to investigate the differences but there are quite many (but they are similar).
I went through roughly half of the differences and as far as I could see 2 major change, one of them in my opinion was a fix and the other one seems like an issue, so I have to investigate that one.

Kokan:
I am closing this PR as this solves an actually `kv-parser` issue, but due the `kv-parser` is internally used by `snmtp parser`, which also changes its behavior. That in some case are better, and in some case make it worse (that was also not the best option).

The things that probably should be addressed before this PR:
* `snmtp parser` coupled with `kv-parser`
* the `kv-parser` and other components based on a str decode function, which has a boat load of heuristic (there are hidden gems like the `space` always considered as a separator...)

Issue #2131, opened 2018-06-21
increased CPU usage with elasticsearch2 destination


meierfra-ergon:
# syslog-ng

## Version of syslog-ng
3.15.1-2

output of syslog-ng --version
syslog-ng 3 (3.15.1)
Config version: 3.15
Installer-Version: 3.15.1
Revision: 
Compile-Date: May 16 2018 07:23:22
Module-Directory: //usr/lib64/syslog-ng
Module-Path: //usr/lib64/syslog-ng
Error opening plugin module; module='mod-java', error='libjvm.so: cannot open shared object file: No such file or directory'
Available-Modules: hook-commands,stardate,kvformat,add-contextual-data,tfgetent,affile,afamqp,snmptrapd-parser,appmodel,afprog,disk-buffer,cef,confgen,afsocket,basicfuncs,pseudofile,afstomp,graphite,csvparser,syslogformat,cryptofuncs,linux-kmsg-format,system-source,xml,sdjournal,tags-parser,afuser,map-value-pairs,dbparser,date,json-plugin
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: on

## Platform
CentOS 7.5

## Debug bundle

# Issue
We are using Syslog-NG to write our Logs into Files and also into an Elasticsearch DB. Before we were using Syslog-NG version 3.11.1-2 and since the Update to 3.15.1-2 we experience a substantial increase in CPU usage. The increase does only exists, if the elasticsearch2 destination is active, so I think the the Problem exists in this component.

## Failure
20% higher CPU usage

## Steps to reproduce
1. use elasticsearch2 destination
2. start a load-test with 3.11.1-2 
3. start a load-test with 3.15.1-2
4. compare CPU usages

## Configuration
(only elastcsearch2)
```
destination d_elastic_local {
        elasticsearch2(
                index("logs-${YEAR}.${MONTH}.${DAY}")
                type("log")
                cluster("elasticsearch")
                persist_name("elastic-local")
[...]

PR #2251, opened 2018-08-27
implement flush-timeout() for http

bazsi:
This branch implements the flush-timeout() option for LogThreadedDestDriver, which works very similarly to what we had earlier (e.g. pre syslog-ng 3.3).

Then it makes this option available to the HTTP destination driver. This should fix CPU consumption issues for elasticsearch, such as #2131.

After giving some though whether we should simply use the old option name (e.g. flush-timeout()) or introduce a new one, I've finally concluded on simply using the old one, meaning that it should be brought out of deprecation. But looking at it closely, that deprecation wasn't very successful, in the sense it has not been documented.

This also means that the flush-timeout() and flush-lines() options would be inherited automatically from the global options. This branch implements that as well for the generic case, and then disables this inheritance in the http driver, as that would be unexpected.

NOTE: this is not a complete solution for #2131. To solve #2131, we would need to add this option to the Java binding as well, or feed Elastic with the http destination. The latter should be an interesting experiment.

### DIFF
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -45,6 +45,15 @@ log_threaded_dest_driver_set_flush_lines(LogDriver *s, gint flush_lines)
   self->flush_lines = flush_lines;
 }
 
+void
+log_threaded_dest_driver_set_flush_timeout(LogDriver *s, gint flush_timeout)
+{
+  LogThreadedDestDriver *self = (LogThreadedDestDriver *) s;
+
+  self->flush_timeout = flush_timeout;
+}
+
+
 /* this should be used in combination with WORKER_INSERT_RESULT_EXPLICIT_ACK_MGMT to actually confirm message delivery. */
 void
 log_threaded_dest_driver_ack_messages(LogThreadedDestDriver *self, gint batch_size)
@@ -72,6 +81,24 @@ log_threaded_dest_driver_rewind_messages(LogThreadedDestDriver *self, gint batch
   self->batch_size -= batch_size;
 }
 
+static gboolean
+_should_flush_now(LogThreadedDestDriver *self)
+{
+  struct timespec now;
+  glong diff;
+
+  if (self->flush_timeout <= 0 ||
+      self->flush_lines <= 1 ||
+      !self->enable_flush_timeout)
+    return TRUE;
+
+  iv_validate_now();
+  now = iv_now;
+  diff = timespec_diff_msec(&now, &self->last_flush_time);
+
+  return (diff > self->flush_timeout);
+}
+
 static gchar *
 _format_seqnum_persist_name(LogThreadedDestDriver *self)
 {
@@ -97,6 +124,10 @@ _stop_watches(LogThreadedDestDriver *self)
   if (iv_timer_registered(&self->timer_throttle))
     {
       iv_timer_unregister(&self->timer_throttle);
+    }
+  if (iv_timer_registered(&self->timer_flush))
+    {
+      iv_timer_unregister(&self->timer_flush);
     }
 }
 
@@ -243,6 +274,7 @@ _process_result(LogThreadedDestDriver *self, gint result)
       break;
 
     case WORKER_INSERT_RESULT_QUEUED:
+      self->enable_flush_timeout = TRUE;
       break;
 
     default:
@@ -260,6 +292,15 @@ _perform_inserts(LogThreadedDestDriver *self)
   LogMessage *msg;
   worker_insert_result_t result;
   LogPathOptions path_options = LOG_PATH_OPTIONS_INIT;
+
+  if (self->batch_size == 0)
+    {
+      /* first message in the batch sets the last_flush_time, so we
+       * won't expedite the flush even if the previous one was a long
+       * time ago */
+
+      self->last_flush_time = iv_now;
+    }
 
   while (G_LIKELY(!self->under_termination) &&
          !self->suspended &&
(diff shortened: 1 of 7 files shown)


lib/logthrdestdrv.c
```diff
@@ -303,9 +294,9 @@ _perform_flush(LogThreadedDestDriver *self)
    * flush() being called always, even if WORKER_INSERT_RESULT_SUCCESS is
    * returned, in which case batch_size is already zero at this point.
    */
-  if (!self->suspended && self->worker.flush)
+  if (!self->suspended)
     {
-      worker_insert_result_t result = self->worker.flush(self);
+      worker_insert_result_t result = log_threaded_dest_worker_flush(self);
```
furiel: This change is not an equivalent transformation. In the original version, accepting the batch happens only if flush vfunction is provided. In the new code it happens everytime. This can only be a problem when a driver wants to use the batching framework without specifying a flush. I cannot tell any reasonable usecase for this,  though the API allows to do so.

The cleanest would be to start the function with:
```
if (!self->worker.flush)
  return;
```

lib/logthrdestdrv.c
```diff
@@ -325,16 +363,40 @@ _message_became_available_callback(gpointer user_data)
...
+    iv_task_register(&self->do_work);
+}
+
+static void
+_schedule_restart_on_next_flush(LogThreadedDestDriver *self)
+{
+  if (self->suspended)
+    _schedule_restart_on_suspend_timeout(self);
+  else if (!_flush_time_has_elapsed(self))
+    _schedule_restart_on_flush_timeout(self);
```
furiel: Here in the `!_flush_time_has_elapsed(self)` branch we should still call `iv_task_register(&self->do_work);`, if there are messages in the queue, we should handle them.

lib/logthrdestdrv.c
lbudai: @furiel: I think it is equivalent. Just check the implementation of `log_threaded_dest_worker_flush`.
(and IMHO it is better as checking of the existence of a virtual method is moved to the right place)

lib/logthrdestdrv.c
lbudai: @furiel: flush_handler and work_handler is the same : `_perform_work`, so I think it is correct, but maybe I'm wrong.

lib/logthrdestdrv.c
furiel: I still think it is not equivalent, at least not mathematically. In the original implementation with self->worker.flush == 0, the whole if block is skipped. In the new implementation, we simulate a dummy flush method by returning `WORKER_INSERT_RESULT_SUCCESS`. But that has a side effect: acknowledging a half-full batch, which the original implementation did not do. Maybe we can reason why it is not a problem (I am not 100% convinced yet), but this is a functional change in the middle of an otherwise innocent-looking refactor patch, which is unfortunate.

lib/logthrdestdrv.c
furiel: They almost do the same, but `flush_handler` executes `_perform_work` only `flush_timeout` later. But when we are in the `!suspended` branch, we should have a message to be handled. It is possible that is the last message missing from a batch to flush, which would delay the batch flush unnecessarily.

lbudai:
@bazsi : `test_logthrdestdrv`  is broken currently

lib/logthrdestdrv.c
bazsi: I agree this is not functionally equivalent, however after giving some thought this should have been the way it was coded originally. I should have added more comments to the commit message, sorry for that.

As for the rationale:

Remember that in earlier versions of this code, on_message_queue_empty() was called if provided and flush superceedes that logic. 

The original on_message_queue_empty() method was returning void, basically assuming that any flush that needed happening indeed happened successfully, there was no way to indicate failure.

With the new flush logic, in this codepath, one of two things could happen:

*1) the flush() method is provided*

In this case  we work as expected.

*2) flush() is not set*

then I'd expect the driver never return a QUEUED verdict as there's no way to handle queued messages. So
instead of QUEUED it would return a final verdict like SUCCESS or ERROR. In either case the insert() method would already call _process_result(), causing batch_size to be reset to zero. This in turn means that the process_result() here would become a noop, as anything it does is basically dependant on batch_size. (well, it also resets the retries_counter, but again that shouldn't matter that we set that to zero twice).

lib/logthrdestdrv.c
bazsi: do_work would wake up immediately and not when the message arrives. This means that we would spin in the thread, as we would wake up immediately, finding that the queue is empty.

The way this is handled is that log_queue_check_items() registers a callback function whenever there is NO message in the queue and that gets called if one is added, this function is _message_became_available_callback() and that wakes up our thread immediately.

So in summary:
1) this function is only called when we have some items in the batch but no items in the queue
2) in this case message_became_available_callback() is registered and would be called by the LogQueue
3) any new message thus wakes up the thread, which would pull all of the messages it finds there.
4) if the batch is still not complete we would repeat the same process until either the flush timeout elapses, or we get enough items  to fill a batch.

I think this should be what we expect. I agree though that the wakeup logic is a bit messy here. I tried a few alternatives, but none seemed to read better :(

bazsi:
The last set of changes broke the test (e.g. that one that migrates this to be flush-timeout and implements inheritance from the global options), but works for me now.

bazsi:
I've also added two testcases to test_logthrdestdrv to test flush-timeout(), there are some timing magic involved so please review it with a critical eye.

lib/logthrdestdrv.c
furiel: Thanks for the detailed explanation. This was the missing puzzle for me, without flush but returning QUEUED users cannot accept batch. I see `_accept_batch` is not exposed to the `logthrdestdrv.h` header. (Not static function either, maybe worth change it to static later.).

furiel:
@bazsi I do not have any other comments. The unit test for flush timeout looks good to me. When travis and kira is green, we can merge this.

bazsi:

bazsi:
This is now ready for merging, unfortunately I've rebased it a number of times invalidating the reviews so far.

The important changes were:
1) fix a behavioural change, flush-timeout() defaults to 10 seconds for destinations that do not override it specifically (e.g. java). But that shouldn't be applied unless at least one message is QUEUED into the batch.

2) in the course of fixing 1) above, I've introduced a race condition that I've also fixed, but took me a few iterations until I've found the root cause. I've squashed the issue and its fix into the base patch, so it is not visible with the current set of patches.

3) a few improvements in .travis.yml on macOS, as that platform was the most reliable in finding the race mentioned in 2) above.

lib/logthrdestdrv.c
```diff
@@ -243,6 +274,7 @@ _process_result(LogThreadedDestDriver *self, gint result)
       break;
 
     case WORKER_INSERT_RESULT_QUEUED:
+      self->enable_flush_timeout = TRUE;
```
furiel: Should we set this variable to false in _accept_batch and _drop_batch?

bazsi:
No, it is only a "turn-on" mechanism, to be compatible with legacy drivers,
which shouldn't be subject to flush-timeout() at all.

lib/logthrdestdrv.c
furiel: Oh ok I thought we could use self->flush_timeout, but that would not work because it can be set with global options, so we need an additional filtering mechanism not to start the timer for drivers that do not support it. Marking as resolved.

bazsi:
Thanks \o/

PR #2256, opened 2018-08-30
dbld: use upstream gradle binaries

bazsi:
The distribution provided gradle binaries are often not good enough,
either because they are too old or they are buggy.

Use the binary provided by the gradle project.


lbudai:
@bazsi: what do you think about using [sdkman](https://sdkman.io) ?

bazsi:
I am not against sdkman per se, however the gradle installation was already there, and I wanted to fix the build. Since I would concentrate on getting this into kira, I wouldn't spend much cycles on sdkman.

Also, it seems to be a tool for interactive use on dev workstations, whereas our build environments are much less used in that way, plus distros usually have their own java packaging and we should integrate with that.

Maybe in the devshell image we should be using sdkman and stick to normal packages in the build images.

lbudai:
@bazsi: well... apt is also an interactive tool ;)
Not only gradle has problems, but also java (with cacerts) in the latest Ubuntu LTS... this is why I'm prefer using stable java/gradle versions (and sdkman is a tool which supports installing both).

bazsi:
Well, "apt" is, but "apt-get" is an API. :)

PR #2257, opened 2018-08-31
Hdfs time-reap

### LINKED ISSUE #917: Java heap size explodes for Kafka destination, possible memleak
We run 3.7.2 to read from syslog input and write to kafka output. It works great, but hits CGroup OOM. I suspect that java heap is set to something that we do not expect.

12h time frame from monitoring for 10 syslog-ng instances looks like this:

image

Bold red line is memory limit.

This is how `ps aux` looks inside of a container:

```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1 42.0  1.9 40336700 2569312 ?    Ssl  10:53 152:43 /usr/sbin/syslog-ng -F -f /etc/syslog-ng/syslog-ng.conf
```

Config file:

```
@version: 3.7
@module mod-java

options {
    use_dns(no);
    keep_hostname(yes);
    create_dirs(yes);
    ts_format(iso);
};

source s_net {
    network(
      ip(myhost.mydomain)
      port(31917)
      so-keepalive(yes)
      log-iw-size(100000)
      max-connections(1000)
      transport("tls")
      tls(
        peer-verify(optional-untrusted)
        key-file('/etc/syslog-ng/key.d/example.com')
        cert-file('/etc/syslog-ng/cert.d/example.com')
      )
    );
};

destination kafka {
  java(
    class_path("/usr/lib/syslog-ng/java-modules/*.jar:/usr/local/kafka/libs/*.jar")
    class-name("org.syslog_ng.kafka.KafkaDestination")

    option("kafka-bootstrap-servers", "some.zk.nodes:2181")
    option("topic", "mytopic")

    option("key", "$HOST")
    option("template", "$(format-json @timestamp=\"$ISODATE\" logsource=\"$HOST\" program=\"$PROGRAM\" message=\"$MESSAGE\" app=\"something\")")
  );
};

log {
    source(s_net);
    destination(kafka);
};
```

Kokan:
This patchset introduces the `time_reap` option for the `hdfs` destination. The user could simply set this option in seconds after that every file that has not been written since `time_reap` seconds is going to be closed.
(Note: In case of archive is configured it also triggers the archiving mechanism for that file only.)
```
  hdfs( ... time_reap(60));
```

The option default value is `0`, which means disable the timeout.

The `java: option set/get key leaked` commit fixes a general java based option leak, every get/set option leaked per call.
Fixes #917

### DIFF
--- a/modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
+++ b/modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
@@ -232,14 +232,11 @@ public class HdfsDestination extends StructuredLogDestination {
 
         logger.debug("Flushing hdfs");
         for (Map.Entry<String, HdfsFile> entry : openedFiles.entrySet()) {
-            FSDataOutputStream outputStream = entry.getValue().getFsDataOutputStream();
-            if (outputStream != null) {
-                try {
-                    outputStream.hflush();
-                    outputStream.hsync();
-                } catch (IOException e) {
-                    logger.debug(String.format("Flush failed on file %s, reason: %s", entry.getKey(), e.getMessage()));
-                }
+            HdfsFile hdfsfile = entry.getValue();
+            try {
+                hdfsfile.flush();
+            } catch (IOException e) {
+                logger.debug(String.format("Flush failed on file %s, reason: %s", entry.getKey(), e.getMessage()));
             }
         }
     }
--- a/modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
+++ b/modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
@@ -43,6 +43,12 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.UUID;
 
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import static java.lang.Math.abs;
+
 public class HdfsDestination extends StructuredLogDestination {
     private static final String LOG_TAG = "HDFS:";
     private static final String HADOOP_SECURITY_AUTH_KEY = "hadoop.security.authentication";
@@ -60,11 +66,92 @@ public class HdfsDestination extends StructuredLogDestination {
 
     HashMap<String, HdfsFile> openedFiles;
 
+    private Timer timeReap;
+
+    private Lock lock;
+
+    class TimeReapTask extends TimerTask {
+        private boolean closeFile(HdfsFile hdfsfile) {
+           logger.info("Closing file " + hdfsfile.getPath() + "; reason=reaping");
+
+           try {
+               hdfsfile.flush();
+               hdfsfile.close();
+           } catch (IOException e) {
+               printStackTrace(e);
+               return false;
+           }
+
+           if (archiveDir != null) {
+               archiveFile(hdfsfile);
+           }
+
+           return true;
+        }
+
+        private void closeExpiredFiles() {
+           for (Iterator<Map.Entry<String, HdfsFile>> it = openedFiles.entrySet().iterator(); it.hasNext();) {
+               HdfsFile file = it.next().getValue();
+               long current = file.timeSinceLastWrite();
+
+               if (current >= options.getTimeReap()) {
+                  closeFile(file);
+                  it.remove();
+               }
+           }
+        }
+
+        private long calculateNextWakeUp() {
+           if (openedFiles.isEmpty())
+              return 0;
+
+           long nextTimerExpires = -1;
+           for (Map.Entry<String, HdfsFile> entry : openedFiles.entrySet()) {
+               logger.debug("key: " + entry.getKey());
+               long current = entry.getValue().timeSinceLastWrite();
+
+               if (current > nextTimerExpires) {
+                  nextTimerExpires = current;
+               }
+           }
+
+           if (nextTimerExpires <= 0)
+              return 0;
+
+           nextTimerExpires = Math.abs(options.getTimeReap() - nextTimerExpires);
+           logger.trace("Next schedule in " + nextTimerExpires);
+
+           return nextTimerExpires;
+        }
+
+        @Override
+        public void run() {
+           logger.debug("time-reap run");
+           lock.lock();
+           try {
+               closeExpiredFiles();
+
+               long nextTimerExpires = calculateNextWakeUp();
+
+               if (nextTimerExpires <= 0) {
+                  timeReap.cancel();
+                  timeReap = null;
+                  return;
+               }
+
+               timeReap.schedule(new TimeReapTask(), nextTimerExpires);
+           } finally {
+             lock.unlock();
+           }
+        }
+    }
+
     public HdfsDestination(long handle) {
         super(handle);
         logger = Logger.getRootLogger();
         SyslogNgInternalLogger.register(logger);
         options = new HdfsOptions(this);
+        lock = new ReentrantLock();
     }
 
     @Override
@@ -93,6 +180,7 @@ public class HdfsDestination extends StructuredLogDestination {
     @Override
     public boolean open() {
         logger.debug("Opening hdfs");
+        lock.lock();
         openedFiles = new HashMap<String, HdfsFile>();
         isOpened = false;
         try {
@@ -111,6 +199,8 @@ public class HdfsDestination extends StructuredLogDestination {
         } catch (Exception e) {
             printStackTrace(e);
             closeAll(false);
+        } finally {
+          lock.unlock();
         }
         return isOpened;
     }
@@ -137,10 +227,23 @@ public class HdfsDestination extends StructuredLogDestination {
         UserGroupInformation.loginUserFromKeytab(options.getKerberosPrincipal(), options.getKerberosKeytabFile());
     }
 
+
+    protected void updateFilesTimer(HdfsFile hdfsfile) {
+        if (options.getTimeReap() == 0)
+           return;
+
+        if (timeReap != null)
+           return;
+
+        timeReap = new Timer();
+        timeReap.schedule( new TimeReapTask(), options.getTimeReap());
+    }
+
     @Override
     public boolean send(LogMessage logMessage) {
         isOpened = false;
         String resolvedFileName = options.getFileNameTemplate().getResolvedString(logMessage);
+        lock.lock();
         HdfsFile hdfsfile = getHdfsFile(resolvedFileName);
         if (hdfsfile == null) {
             // Unable to open file
@@ -156,11 +259,16 @@ public class HdfsDestination extends StructuredLogDestination {
             byte[] rawMessage = formattedMessage.getBytes(Charset.forName("UTF-8"));
 
             hdfsfile.write(rawMessage);
+
+            updateFilesTimer(hdfsfile);
         } catch (IOException e) {
             printStackTrace(e);
             closeAll(false);
             return false;
+        } finally {
+          lock.unlock();
         }
+
 
         isOpened = true;
         return true;
@@ -236,6 +344,8 @@ public class HdfsDestination extends StructuredLogDestination {
          */
 
         logger.debug("Flushing hdfs");
+        lock.lock();
+        try {
         for (Map.Entry<String, HdfsFile> entry : openedFiles.entrySet()) {
             HdfsFile hdfsfile = entry.getValue();
             try {
@@ -244,16 +354,30 @@ public class HdfsDestination extends StructuredLogDestination {
                 logger.debug(String.format("Flush failed on file %s, reason: %s", entry.getKey(), e.getMessage()));
             }
         }
+        } finally {
+          lock.unlock();
+        }
     }
 
     @Override
     public void close() {
         logger.debug("Closing hdfs");
+
+        lock.lock();
+        try {
         closeAll(true);
+        } finally {
+          lock.unlock();
+        }
         isOpened = false;
     }
 
     private void closeAll(boolean isArchiving) {
+        if (timeReap != null) {
+            timeReap.cancel();
+            timeReap = null;
+        }
+
         closeAllHdfsFiles();
 
         if (isArchiving) {
--- a/modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
+++ b/modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
@@ -144,7 +144,7 @@ public class HdfsDestination extends StructuredLogDestination {
         HdfsFile hdfsfile = getHdfsFile(resolvedFileName);
         if (hdfsfile == null) {
             // Unable to open file
-            closeAll(archiveDir != null);
+            closeAll(true);
             return false;
         }
 
@@ -245,7 +245,7 @@ public class HdfsDestination extends StructuredLogDestination {
     @Override
     public void close() {
         logger.debug("Closing hdfs");
-        closeAll(archiveDir != null);
+        closeAll(true);
         isOpened = false;
     }
 
--- a/modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
+++ b/modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
@@ -289,20 +289,25 @@ public class HdfsDestination extends StructuredLogDestination {
         }
     }
 
+    private void archiveFile(HdfsFile hdfsfile) {
+        Path filePath = hdfsfile.getPath();
+        logger.debug(String.format("Trying to archive %s to %s", filePath.getName(), archiveDir));
+        Path archiveDirPath = new Path(String.format("%s/%s", options.getUri(), archiveDir));
+
+        try {
+           hdfs.mkdirs(archiveDirPath);
+           hdfs.rename(filePath, archiveDirPath);
+        } catch (IOException e) {
+            logger.debug(String.format("Unable to archive, reason: %s", e.getMessage()));
+        }
+    }
+
     private void archiveFiles() {
         if (archiveDir == null)
            return;
 
         for (Map.Entry<String, HdfsFile> entry : openedFiles.entrySet()) {
-            Path filePath = entry.getValue().getPath();
-            logger.debug(String.format("Trying to archive %s to %s", filePath.getName(), archiveDir));
-            Path archiveDirPath = new Path(String.format("%s/%s", options.getUri(), archiveDir));
-            try {
-                hdfs.mkdirs(archiveDirPath);
-                hdfs.rename(filePath, archiveDirPath);
-            } catch (IOException e) {
-                logger.debug(String.format("Unable to archive, reason: %s", e.getMessage()));
-            }
+            archiveFile( entry.getValue() );
         }
     }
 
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -60,11 +66,97 @@
...
+        @Override
+        public void run() {
+           logger.debug("time-reap run");
+           lock.lock();
+           try {
+           closeExpiredFiles();
+
+           long nextTimerExpires = caculateNextWakeUp();
+
+           timeReap.cancel();
```
gaborznagy: I don't think we need to cancel other tasks, since as I see it we create a new timer every time we need to reschedule.
The other schedule logic is in `send()` method, which can only schedule a timer if there is none.
We can't have multiple tasks with the same timer.

I'm not a java expert, but I saw that you can reuse a timer, just add a new TimerTask instance.
Otherwise I would use `Timer.purge()` as well to be able to gc the cancelled tasks.
See [Timer](https://docs.oracle.com/javase/8/docs/api/java/util/Timer.html#cancel())

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -60,11 +66,97 @@
...
+        public TimeReap(Map<String, HdfsFile> files) {
+            logger.trace("Create TimeReap");
+            this.files = files;
+        }
+
+        private boolean closeFile(HdfsFile hdfsfile) {
+           logger.info("Close file" + hdfsfile.getPath());
+
+           try {
+               hdfsfile.flush();
```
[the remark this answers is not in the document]
Kokan: There is no need for the `flush()`, as it does not do much. I wanted to call it, because later on this could mean actual flush. But sure I can remove it now, and we can add if needed.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -60,11 +66,97 @@
...
+    private Lock lock;
+
+    class TimeReap extends TimerTask {
+        private Map<String, HdfsFile> files;
+        public TimeReap(Map<String, HdfsFile> files) {
+            logger.trace("Create TimeReap");
+            this.files = files;
+        }
+
+        private boolean closeFile(HdfsFile hdfsfile) {
```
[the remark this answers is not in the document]
Kokan: We don't really. Actually when I started to do some refactor, it would have been possible to have one close. The issue was with archive, because the archiving is started if configured and the magical `bool` flag is flipped. If I wanted to have one `close` for both the timer and for the destination, either I had to deal with archive or just pass an extra parameter as in `closeAllFiles`.

I draw the refactor line at archive, thus the above divided closing.

kira-syslogng:
Build SUCCESS

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsFile.java
```diff
@@ -55,4 +55,12 @@ public void flush() throws IOException {
          fsDataOutputStream.hflush();
          fsDataOutputStream.hsync();
     }
+
+    public void close() throws IOException {
+         if (fsDataOutputStream == null)
+            return;
+
+          fsDataOutputStream.close();
+          fsDataOutputStream = null;
```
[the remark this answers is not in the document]
Kokan: sure.

kira-syslogng:
Build SUCCESS

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -60,11 +66,97 @@
...
+           logger.info("Close file" + hdfsfile.getPath());
```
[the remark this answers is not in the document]
gaborznagy: Yes, that would be good. The point is to emphasize "reap" is the cause of the closing.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsFile.java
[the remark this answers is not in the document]
Kokan: I cannot set it `null`, because `archive` is relying on having a path. That dependency is not nice, it can be refactored on some other time.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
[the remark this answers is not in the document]
Kokan: As we speak in IRL, I extend the calls where `close()` is called with `flush()` covering the feature where it makes sense to call `flush()`.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -289,20 +289,25 @@ private void closeHdfs() {
         }
     }
 
+    private void archiveFile(HdfsFile hdfsfile) {
+        Path filePath = hdfsfile.getPath();
+        logger.debug(String.format("Trying to archive %s to %s", filePath.getName(), options.getArchiveDir()));
```
gaborznagy: I forgot previously, although it was my [patch](https://github.com/balabit/syslog-ng/commit/a35d5c1b09c63b0900bf1744ebf82be4f46df806):
if you use `options.getArchiveDir()` here instead of the `archiveDir` attribute, then you will override a previously disabled archiving.
I had an [alternative](https://github.com/gaborznagy/syslog-ng/commit/bb98c77b9d9944ae263c0a60c34087254570bd5b) to my patch, but I did not like it.

kira-syslogng:
Build FAILURE

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
gaborznagy: Thanks for fixing.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -60,11 +66,97 @@
...
+
+        long caculateNextWakeUp() {
+           if (files.isEmpty())
+              return 0;
+
+           long nextTimerExpires = -1;
+           for (Map.Entry<String, HdfsFile> entry : files.entrySet()) {
+               long current = entry.getValue().timeSinceLastWrite();
+
+               if (current > nextTimerExpires || nextTimerExpires == -1) {
```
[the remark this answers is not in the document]
gaborznagy: thanks.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -259,18 +259,15 @@ private void closeAll(boolean isArchiving) {
...
-                try {
-                    logger.debug(String.format("Closing file: %s", entry.getKey()));
-                    outputStream.close();
-                } catch (IOException e) {
-                    printStackTrace(e);
-                }
+            HdfsFile hdfsfile = entry.getValue();
+            try {
+                logger.debug(String.format("Closing file: %s", entry.getKey()));
+                hdfsfile.close();
```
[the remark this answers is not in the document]
gaborznagy: thanks.
At destination side's send() method we don't remove the file neither on close nor after archive.
However since after close() we need to call open(), `openedFiles` will be re-created, so this is fine.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
gaborznagy: thanks.

mitzkia:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -137,45 +222,72 @@ private void setKerberosAuth(Configuration configuration) throws IOException {
...
 
+
         isOpened = true;
         return true;
     }
 
-    private FSDataOutputStream getFSDataOutputStream(String resolvedFileName) {
+    private HdfsFile getHdfsFile(String resolvedFileName) {
         HdfsFile hdfsFile = openedFiles.get(resolvedFileName);
         if (hdfsFile == null) {
```
pzoleex: this could be done easier: hdfsFile == null || !hdfsFile.isOpen()

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -137,45 +222,72 @@ private void setKerberosAuth(Configuration configuration) throws IOException {
...
         isOpened = false;
         String resolvedFileName = options.getFileNameTemplate().getResolvedString(logMessage);
-        FSDataOutputStream fsDataOutputStream = getFSDataOutputStream(resolvedFileName);
-        if (fsDataOutputStream == null) {
+        lock.lock();
+        HdfsFile hdfsfile = getHdfsFile(resolvedFileName);
+        if (hdfsfile == null) {
             // Unable to open file
-            closeAll(archiveDir != null);
+            closeAll(true);
```
pzoleex: hm... seems you want to always archive regardless if it is asked or not

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -231,28 +343,41 @@ public void onMessageQueueEmpty() {
...
     }
 
     @Override
     public void close() {
         logger.debug("Closing hdfs");
-        closeAll(archiveDir != null);
+
+        lock.lock();
+        try {
+        closeAll(true);
```
pzoleex: hm... seems you want to always archive regardless if it is asked or not

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
Kokan: I think the logic here was too complex, the `archiveDir` has been checked twice once as a parameter and once later in the `archiveFiles` function.
Actually it does not make a difference if you pass `true` or `false` if you want to `archive` if archiving is configured. (see even the sentence is too complicated.)

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
Kokan: Well for me the two separate condition was more readable, but @gaborznagy also commented this, so I guess it is just me and time to change this part.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -43,28 +43,110 @@
...
+
+           try {
+               hdfsfile.flush();
+               hdfsfile.close();
+           } catch (IOException e) {
+               printStackTrace(e);
+               return false;
+           }
+
+           if (archiveDir != null)
```
pzoleex: I would not recommend to use this form (no braces).
Later it will be really hard to maintain this code (easy to make errors, just extend with one line and forget about the brace) and the general code style is against this: http://checkstyle.sourceforge.net/reports/google-java-style-20170228.html#s4.1-braces

Please fix this other places too (I can see it several places in the code)

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
pzoleex: in that case just call closeAll(), remove the parameter and decide about archiving within this function

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
pzoleex: the main problem is the code duplication, within the two if, you are doing the same thing twice

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsFile.java
```diff
@@ -26,23 +26,57 @@
...
+         if (!isOpen())
+            throw new IOException("File is not open: "+path);
+
+         fsDataOutputStream.write(message);
+
+         lastWrite = System.currentTimeMillis();
+    }
+
+    public long timeSinceLastWrite() {
+       if (!isOpen())
```
pzoleex: I think this shouldn't be here, it should never happen that we call this function and the file is not open.
currently we just hide if there is something wrong with the code outside

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -43,28 +43,110 @@
...
+
+           return nextTimerExpires;
+        }
+
+        @Override
+        public void run() {
+           logger.debug("time-reap run");
+           lock.lock();
+           try {
+           closeExpiredFiles();
```
pzoleex: seems here is missing some indentation

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
Kokan: Sure I figured what is bugging you.
Of course it is a code duplication but very local - really hard to miss - thus I won't think it would have caused any issue later on. But as I said I am going to change it as it seems its just me who preferred it that way.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
Kokan: The `archiveDir != null` to `true` change was functionally equivalent change. (refactor)
What you are suggestion is not, and that's why I think that should be a in a different PR.
At least I do not want to overload this PR anymore.
(It already has a leak and hdfs fix.)

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
Kokan: I mark this as resolved as it is duplicate of the above and either one would be changed or not.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsFile.java
Kokan: Well yes, it is not called now. But that does not mean it should not be here at all. As the `HdfsFile` is a separate class that should work regardless how it is called.
This class has an internal state of not open, that should be respected by all of its method.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
Kokan: From `CONTRIBUTING.md`: 
/`We are using a coding style very similar to [GNU Coding Standards]`/
As far as I know there is no other style accepted. Therefore bringing up the `google-java-style` guide is bogus.
In fact we have a few examples not to use braces, even in java or C.

But as here I do not really have a preference I could change.

pzoleex: I don't agree. almost all java coding standards say DO NOT use if without brace (google, oracle, others).
would be nice not to refactor this code to a worse format.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsFile.java
pzoleex: can you tell me the decision logic when to raise an exception and when to do other things if the file is not open?
(e.g.: write will thrown an exception, flush will does not nothing, this will return with zero)

Kokan: * `flush`: there was no decision on my part at all it should be identical to the code before.
* `write`: here the logic was to keep the stream write behavior as throwing an exception
* `timeSinceLastWrite`: the `return 0;` just meant as `return`, but I cannot do the later one in case of return value

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -172,10 +172,11 @@ private HdfsFile getHdfsFile(String resolvedFileName) {
     }
 
     private HdfsFile createHdfsFile(String resolvedFileName) {
-        HdfsFile hdfsFile = new HdfsFile();
         Path filePath = getFilePath(resolvedFileName);
-        hdfsFile.setPath(filePath);
-        hdfsFile.setFsDataOutputStream(createFsDataOutputStream(hdfs, filePath));
+        FSDataOutputStream fsDataOutputStream = createFsDataOutputStream(hdfs, filePath);
+
+        HdfsFile hdfsFile = new HdfsFile(filePath, fsDataOutputStream);
```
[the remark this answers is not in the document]
Kokan: Actually for me the issue is not the `filePath` but the `FSDataOutputStream`. Because for a file to me it is natural to provide the path, not some magical stream (That should be the `HdfsFile` internal business.)
But for that I have to deal with append and archive (maybe just append). The actual purpose of this PR was not to refactor, and I made a cut here as for me it was acceptable (compared to the original.)

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsFile.java
```diff
@@ -32,22 +32,19 @@
...
     }
 
     public Path getPath() {
         return path;
     }
 
-    public void setPath(Path path) {
-        this.path = path;
-    }
-
```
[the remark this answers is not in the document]
Kokan: As far as I know you cannot ask this stream about it's path.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -144,7 +144,7 @@ public boolean send(LogMessage logMessage) {
         HdfsFile hdfsfile = getHdfsFile(resolvedFileName);
         if (hdfsfile == null) {
             // Unable to open file
-            closeAll(archiveDir != null);
+            closeAll(true);
```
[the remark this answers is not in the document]
Kokan: I could split this function into `closeAll` and `closeAndArchiveAll`, but the issue with that that the later function either does archive or not which is very confusing for me (and not better).
There are a lot of things going on in this function and those are tightly coupled (which is not good). But I draw the line at `archive`.
As I wrote before there is a place for improvement.

modules/java-modules/hdfs/src/main/java/org/syslog_ng/hdfs/HdfsDestination.java
```diff
@@ -168,7 +168,7 @@ public boolean send(LogMessage logMessage) {
 
     private HdfsFile getHdfsFile(String resolvedFileName) {
         HdfsFile hdfsFile = openedFiles.get(resolvedFileName);
-        if (hdfsFile == null) {
+        if (hdfsFile == null || !hdfsFile.isOpen()) {
```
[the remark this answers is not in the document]
Kokan: Yes, but there I cannot create a nice `open` but only like the constructor with filepath and stream for the same reason.

PR #2260, opened 2018-09-04
DbldPackages: Add missing packages to DBLD

mitzkia:
- add packages to pass tests in Kira

### DIFF
--- a/dbld/images/required-apt/artful.txt
+++ b/dbld/images/required-apt/artful.txt
@@ -15,3 +15,7 @@ python-pep8
 # other tools
 
 # debian packaging tools
+
+# kira
+netbase
+systemd-sysv
--- a/dbld/images/required-apt/xenial.txt
+++ b/dbld/images/required-apt/xenial.txt
@@ -16,3 +16,6 @@ python-pep8
 libbson-dev
 librabbitmq-dev
 libmongoc-dev
+
+# kira
+netbase


Kokan:
This is a nice example. This PR cannot be merged even though this fixes the `Build FAILURE` issue with Kira. As Kira still uses the docker image created from the master, which is not correct.
Also it is possible to merge something that is going to break Kira, but won't be discovered only after merged :)

mitzkia:
Sorry, I forgot to mention, @Kokan is absolutely right. 
The goal of this PR is to pass Kira testcases on PRs.

The detail description why getent specific testcase fails on our balabit/syslog-ng-xenial:latest image:
 * We have updated our xenial:latest image to the recent DBLD
 * The failing testcase in Kira uses getent command, but it returns with nothing. 
 * This is because the getent command wanted to use /etc/protocols file, but it does not exist in our image. 
 * The netbase package will install this file.

furiel:
$ dpkg --listfiles netbase | grep protocols 
/etc/protocols

That part sounds good to me. But why do we need `systemd-sysv`?

mitzkia:
Yes, the systemd-sysv package we needed because we want to run Kira tests also on Artful, and this package will contains the necessary /sbin/init file for us. On other platforms (until Xenial) this is already exist.

furiel:
Why do we need /sbin/init? Entrypoint as `$ sleep infinity` can be an alternative?

PR #2265, opened 2018-09-06
afsocket: implemented min_iw_size_per_connection global option

gyorkop:
Decreasing log_iw_size to a low value could cause performance regression in
syslog-ng, but in some cases in order to reduce the memory consumption, we need
to reduce it to a certain value.

To preserve backward compatibility, a new global option has been introduced to
mitigate this issue. Using min_iw_size_per_connection we can lower that limit
from the previously hard-coded value.

Warning: lowering this value could cause performance degradation.

Signed-off-by: Peter Gyorko <peter.gyorko@balabit.com>

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -197,6 +197,7 @@ extern struct _StatsOptions *last_stats_options;
 %token KW_ENCODING                    10082
 %token KW_TYPE                        10083
 %token KW_STATS_MAX_DYNAMIC           10084
+%token KW_MIN_IW_SIZE_PER_READER      10085
 
 %token KW_CHAIN_HOSTNAMES             10090
 %token KW_NORMALIZE_HOSTNAMES         10091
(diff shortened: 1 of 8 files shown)


kira-syslogng:

Kokan:
@kira-syslogng ok to test

lbudai:
Shouldn't we extend this also to the wildcard file source driver?
In that case we should rename the option (as in case of wildcard file source driver we have `max-files` instead of `max-connections`).

gyorkop:
@lbudai : You are right, I've just implemented a similar behaviour in wildcard file source driver. The name of the global option had to be changed as well.

lbudai:
@gyorkop : thanks!

Would it be possible to use the same global option (with a different name, eg.: `default_min_iw_size`) for both cases? What do you think?

gyorkop:
@lbudai : If you check the latest diff, you can see that we use the same global option, but after some discussion with @Kokan , we renamed it to `min_iw_size_per_reader`.

lbudai:
@gyorkop : ok.

@Kokan :
The name is a little bit technical... what if we just call it as `min_iw_size`?
(on the other hand: `log-iw-size` _is_ technical... so I'm ok with this )

Kokan:
@lbudai I can agree on that this name is technical.

But the `_per_connection` was a nice suffix in case of network connection, so we wanted to somehow capture that this value is not the minimal for the whole source (or at least not always).

We could not come up with a better alternative for `per_connection` that also works with file.

The reason that I think it is not so bad that this is too technical that this value is not something that the user should normally not modify, and those who want to really change it, shall understand a little bit more about what is under the hood. Maybe it is good that not trivial from the name, what does this option change.

Issue #2262, opened 2018-09-05
gradle problem when running dbld/rules rpm-centos7


czanik:
# syslog-ng

## Version of syslog-ng

git snapshot from this morning

## Platform

not relevant here, but it's openSUSE 42.3

# Issue

## Failure
Building the CentOS 7 rpm-s using dbld fails:
```
# dbld/rules rpm-centos7
if [ -f /root/syslog-ng/dbld/build/syslog-ng-3.17.2.tar.gz ]; then \
	count=`find /root/syslog-ng -newer /root/syslog-ng/dbld/build/syslog-ng-3.17.2.tar.gz | grep -v "^/root/syslog-ng/dbld/" | wc -l`; \
else \
	count=1; \
fi; \
[ "${count}" -gt 0 ] && docker run -e USER_NAME_ON_HOST=root --network=host --privileged --ulimit nofile=1024:1024 -v /root/syslog-ng:/source -v /root/syslog-ng/dbld/build:/build -v /root/syslog-ng/dbld/install:/install -v ~/.gitconfig:/root/.gitconfig --rm -ti balabit/syslog-ng-xenial /source/dbld/tarball || true
docker run -e USER_NAME_ON_HOST=root --network=host --privileged --ulimit nofile=1024:1024 -v /root/syslog-ng:/source -v /root/syslog-ng/dbld/build:/build -v /root/syslog-ng/dbld/install:/install -v ~/.gitconfig:/root/.gitconfig --rm -ti balabit/syslog-ng-xenial /source/dbld/pkg-tarball snapshot
Generating snapshot version in debian/
Generating snapshot version in syslog-ng.spec
docker run -e USER_NAME_ON_HOST=root --network=host --privileged --ulimit nofile=1024:1024 -v /root/syslog-ng:/source -v /root/syslog-ng/dbld/build:/build -v /root/syslog-ng/dbld/install:/install -v ~/.gitconfig:/root/.gitconfig --rm -ti  balabit/syslog-ng-centos7 /source/dbld/rpm
warning: bogus date in %changelog: Wed Aug 10 2018 Laszlo Budai <laszlo.budai@balabit.com> - 3.17.2-1
warning: bogus date in %changelog: Wed Aug 9 2018 Laszlo Budai <laszlo.budai@balabit.com> - 3.17.1-1
error: Failed build dependencies:
	gradle is needed by syslog-ng-3.17.2-1+20180905T085833.el7.x86_64
[...]

PR #2266, opened 2018-09-06
syslog-ng.spec: remove gradle from the list of "BuildRequires"

szemere:
Fixes: #2262, by installing gradle manually we do not fulfill the "package requirement" of the specfile.

### DIFF
--- a/packaging/rhel/syslog-ng.spec
+++ b/packaging/rhel/syslog-ng.spec
@@ -34,7 +34,7 @@ BuildRequires: hiredis-devel
 BuildRequires: riemann-c-client-devel
 BuildRequires: python-devel
 BuildRequires: java-devel
-BuildRequires: gradle
+#BuildRequires: gradle    # Upstream version installed manually, discussed in: https://github.com/balabit/syslog-ng/issues/2262
 BuildRequires: syslog-ng-java-deps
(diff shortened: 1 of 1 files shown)


packaging/rhel/syslog-ng.spec
```diff
@@ -34,7 +34,7 @@ BuildRequires: hiredis-devel
 BuildRequires: riemann-c-client-devel
 BuildRequires: python-devel
 BuildRequires: java-devel
-BuildRequires: gradle
```
gaborznagy: We already decided to use upstream gradle.
Shouldn't we check if that one exists, e.g. with a `gradle -v` call?

bazsi: It would be an improvement to check the gradle version in our configure scripts, to check
if it meets minimum requirements. But I think it would be a mistake to do the same in the spec file.

Also, unfortunately, the version number doesn't seem to be a good enough indicator, the gradle versions did work, but someone (either gradle upstream or debian or ubuntu) br0ke it.

gaborznagy: We talked about it in person and I accept that we should not check whether gradle is installed locally (with `gradle -v`).

PR #2267, opened 2018-09-06
wildcard-file: poll method directory delete related leak

furiel:
With monitor_method(poll), when a tracked directory is deleted: g_dir_open allocated a
GError structure with the error: No such file or directory. This
must be freed by syslog-ng.

### DIFF
--- a/modules/affile/directory-monitor-poll.c
+++ b/modules/affile/directory-monitor-poll.c
@@ -23,6 +23,7 @@
 #include <collection-comparator.h>
 #include "directory-monitor-poll.h"
 #include "timeutils.h"
+#include "messages.h"
 
 #include <iv.h>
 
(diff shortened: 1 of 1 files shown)


modules/affile/directory-monitor-poll.c
```diff
@@ -99,6 +100,10 @@ _rescan_directory(DirectoryMonitorPoll *self)
     {
       collection_comparator_stop(self->comparator);
       _handle_deleted_self(self);
+      msg_debug("Error while opening directory",
+                evt_tag_str("dirname", self->super.real_path),
+                evt_tag_str("error", error->message));
+      g_clear_error(&error);
```
Kokan: I know it is not fare, but could you fix the `_start_watches` which also leaks the `error` :)

furiel: Thanks for note. I pushed a patch. I tried to trigger that code path by messing with the permissions, but could not. This was because the caller also tried to open, and report an error there. So I just passed NULL to `g_dir_open`, and not doing error reporting and GError cleanup.

Issue #2228, opened 2018-08-14
SIGSEGV in log_expr_node_format_location


mitzkia:
This is not the first time that we found SIGSEGV around: log_expr_node_format_location
- [issue 1986](https://github.com/balabit/syslog-ng/issues/1986)
- [PR 2028](https://github.com/balabit/syslog-ng/pull/2028)

Reproduction
1, start syslog-ng with the following config and with the following prerequsities:
* do not open destination,
* create file with `log_iw_size` line of messages, so flow-control kicks in,
* enable debug mode (either with -d option or through `syslog-ng-ctl`)
```
@version: 3.16
source s_file {
    file("/tmp/input" log_iw_size(10));
};
destination d_network {
    network("127.0.0.1" port(9999));
};
log { source(s_file); destination(d_network); flags(flow-control); };
```
2, send **2** reloads to syslog-ng with `sbin/syslog-ng-ctl reload`
3, stop syslog-ng -> crash happens

bt: 
```
#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:93
#1  0x00007fda7e7534d3 in _IO_vfprintf_internal (s=s@entry=0x7ffde47b96b0, format=format@entry=0x7fda7eead64b "%s:%d:%d", ap=ap@entry=0x7ffde47b9830) at vfprintf.c:1643
#2  0x00007fda7e828169 in ___vsnprintf_chk (s=0x7ffde47b9960 "", maxlen=<optimized out>, flags=1, slen=<optimized out>, format=0x7fda7eead64b "%s:%d:%d", args=0x7ffde47b9830) at vsnprintf_chk.c:63
#3  0x00007fda7eb77ecf in g_snprintf () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#4  0x00007fda7ee36d46 in log_expr_node_format_location (self=0x55b8210dac60, buf=0x7ffde47b9960 "", buf_len=128) at ../lib/cfg-tree.c:146
#5  0x00007fda7ee36dbe in log_expr_node_location_tag (self=0x55b8210dac60) at ../lib/cfg-tree.c:161
#6  0x00007fda7ee41b3f in log_pipe_location_tag (pipe=0x55b8210e1370) at ../lib/logpipe.c:33
#7  0x00007fda7ee45b9b in log_source_wakeup (self=0x55b8210e1370) at ../lib/logsource.c:44
[...]

PR #2269, opened 2018-09-10
Fix log expr node use after free

lbudai:
fixes: #2228

### DIFF
--- a/lib/cfg-tree.c
+++ b/lib/cfg-tree.c
@@ -28,6 +28,33 @@
 
 #include <string.h>
 
+static void _log_expr_node_free(LogExprNode *self);
+
+LogExprNode *
+log_expr_node_ref(LogExprNode *self)
+{
+  g_assert(!self || g_atomic_counter_get(&self->ref_cnt) > 0);
+
+  if (self)
+    {
+      g_atomic_counter_inc(&self->ref_cnt);
+    }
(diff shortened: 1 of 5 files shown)


lib/cfg-tree.c
```diff
@@ -212,6 +212,22 @@ log_expr_node_set_children(LogExprNode *self, LogExprNode *children)
...
 }
 
+void
+log_expr_node_subscribe_pipe(LogExprNode *self, LogPipe *pipe)
+{
+  pipe->expr_node = log_expr_node_ref(self);
+}
+
+void
+log_expr_node_unsubscribe_pipe(LogExprNode *self, LogPipe *pipe)
```
MrAnno: `subscribe()` and `unsubscribe()` do not just increment/decrement a ref counter, they also set a field in the `LogPipe` instance. Because a `LogPipe` field is modified, shouldn't we move these functions to `logmsg.h`?

`log_pipe_subscribe_expr_node` and `log_pipe_unsubscribe_expr_node` would be their name. What do you think?

lbudai: thanks for the note!
(original implementation was 'closer' to this name, but well... it is better to move to `LogPipe` and call it `attach/detach` instead of `subscribe/unsubscribe`)

PR #2273, opened 2018-09-12
Undefined filter reference

Kokan:
The `filter` filter could refer to other filter, which is checked during the initialization time instead of configuration parse time. But at this point for some reason syslog-ng only reported as an error message:
`Referenced filter rule not found in filter() expression ...`

The syslog-ng should not start in this case, and this patch solves that.


Kokan:

PR #2275, opened 2018-09-12
logthrdestdrv infinite loop with ivykis prior 0.39

furiel:
I moved some `iv_validate_now` around, added a few `iv_invalidate_now`, but the main root cause was that the function that check if we need to flush expected strictly more time to pass than flush_timeout (> instead of  >=). Although when flush triggered by the timer, in some cases exactly flush_timeout passed, which resulted in skipping the flush and reregistering the same timer in the past, which resulted in an infinite loop.

For more detailed analysis, please check commit messages.

### DIFF
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -96,7 +96,7 @@ _should_flush_now(LogThreadedDestDriver *self)
   now = iv_now;
   diff = timespec_diff_msec(&now, &self->last_flush_time);
 
-  return (diff > self->flush_timeout);
+  return (diff >= self->flush_timeout);
 }
 
 static gchar *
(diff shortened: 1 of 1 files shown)


lib/tests/test_logthrdestdrv.c
```diff
@@ -776,6 +777,7 @@ static void
 setup(void)
 {
   app_startup();
+  init_and_load_syslogformat_module();
```
Kokan: Instead of using an unneeded dependency (`syslogformat`) wouldn't it be possible to create a new logmsg, I mean this test has nothing to do with `syslogformat`

furiel: Good point. But for now I just simply drop that patch because it is a cosmetic change.

lbudai:
Thanks!

PR #2279, opened 2018-09-14
Fix memory leak caused by saving stats counter to persist config

lbudai:
Leak:

* start syslog-ng with a simple config

```
@version: 3.17

options {
  time_reopen(5);
  stats-level(3);
};

source s_file {
  file("/tmp/input-logs.txt" log-iw-size(10));
};

destination d_file {
  file("/tmp/output-logs.txt");
};

destination d_net {
 network("localhost" port(5555) 
        );
};

log {
  source (s_file);
  destination (d_net);
};

```

* reload syslog-ng
Each time a memory-usage counter is written into a variable that is dynamically allocated on the heap and not freeing up (on my machine 8 bytes are leaked for each reload).

### DIFF
--- a/lib/logqueue.c
+++ b/lib/logqueue.c
@@ -168,23 +168,6 @@ log_queue_check_items(LogQueue *self, gint *timeout, LogQueuePushNotifyFunc para
   return TRUE;
 }
 
-void
-log_queue_set_counters(LogQueue *self, StatsCounterItem *queued_messages, StatsCounterItem *dropped_messages,
-                       StatsCounterItem *memory_usage)
-{
-  self->queued_messages = queued_messages;
-  self->dropped_messages = dropped_messages;
-  self->memory_usage = memory_usage;
-}
-
-void
-log_queue_init_counters(LogQueue *self)
-{
-  stats_counter_set(self->memory_usage,
-                    self->memory_usage_qout_initial_value + self->memory_usage_overflow_initial_value);
-  stats_counter_set(self->queued_messages, log_queue_get_length(self));
-}
-
 static void
 _reset_counters(LogQueue *self)
 {
--- a/lib/logqueue.c
+++ b/lib/logqueue.c
@@ -185,6 +185,60 @@ log_queue_init_counters(LogQueue *self)
   stats_counter_set(self->queued_messages, log_queue_get_length(self));
 }
 
+static void
+_reset_counters(LogQueue *self)
+{
+  stats_counter_set(self->memory_usage,
+                    self->memory_usage_qout_initial_value + self->memory_usage_overflow_initial_value);
+  stats_counter_set(self->queued_messages, log_queue_get_length(self));
+}
+
+static void
+_register_common_counters(LogQueue *self, gint stats_level, const StatsClusterKey *sc_key)
+{
+  gboolean need_to_reset_counters = TRUE;
+
+  stats_register_counter(stats_level, sc_key, SC_TYPE_QUEUED, &self->queued_messages);
+
+  if (stats_check_level(STATS_LEVEL1))
+    need_to_reset_counters = !stats_contains_counter(sc_key, SC_TYPE_MEMORY_USAGE);
+
+  stats_register_counter_and_index(STATS_LEVEL1, sc_key, SC_TYPE_MEMORY_USAGE, &self->memory_usage);
+
+  if (need_to_reset_counters)
+    _reset_counters(self);
+}
+
+void
+log_queue_register_stats_counters(LogQueue *self, gint stats_level, const StatsClusterKey *sc_key)
+{
+  _register_common_counters(self, stats_level, sc_key);
+
+  if (self->register_stats_counters)
+    self->register_stats_counters(self, stats_level, sc_key);
+}
+
+static void
+_unregister_common_counters(LogQueue *self, const StatsClusterKey *sc_key)
+{
+  stats_unregister_counter(sc_key, SC_TYPE_QUEUED, &self->queued_messages);
+  stats_unregister_counter(sc_key, SC_TYPE_MEMORY_USAGE, &self->memory_usage);
+}
+
+void
+log_queue_unregister_stats_counters(LogQueue *self, const StatsClusterKey *sc_key)
+{
+  _unregister_common_counters(self, sc_key);
+
+  if (self->unregister_stats_counters)
+    self->unregister_stats_counters(self, sc_key);
+}
+void
+log_queue_set_dropped_counter(LogQueue *self, StatsCounterItem *dropped_messages)
+{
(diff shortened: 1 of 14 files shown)


lbudai:
kira link : https://ci.syslog-ng.com/jenkins/view/all/job/kira-test-compile-automake-run-testdb/1264/

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lbudai:
hmmm... the reason why Travis/Kira is timeouted is that I have a mutex in a locked state.
The reason of this is that one of the the asserts has failed _before_ the unlock is called... (in c++ this is not an issue as one could use RAII and scope lock). Using `cr_expect` instead of `cr_assert` solves this blocking issue.

lbudai:
...and the failing assert was a mistake by myself... now I think it is fixed

bazsi:
I would depend on this patch in my threaded logthrdestdriver work, as it wasn't easy to resolve the same
in a multi-worker environment.

So this is not just a minor leak, but helps me a big deal. I'll try to review this tonight.

Thanks for fixing this!

lbudai:
@bazsi: there are some 'ugly' parts in the PR: I didn't want to lock/unlock the stats twice, so I extended the `register_stats` methods with assigning the counters to the `LogQueue`.

lbudai:
I’m going to check it on tomorrow and move it to LogQueue.
(when a counter exists, the queue resets the counter... this is what I called `init`, but maybe `reset` would be a better name ).

lbudai:
@bazsi: in case of dropped counter the ownership is at logwriter/logthrdstdrv/afsql (and I would not touch that part), but the other two counters are owned by the LogQueue, so I'll modify the code.

lbudai:
@bazsi: I think (I'm sure...), later the dropped counter can be removed from the `LogQueue`, so until that I'd keep this `log_queue_set_dropped_counter` method.
Thanks for suggesting this change, it's better now (IMHO).

note: in afsql module the stats counter registration point changed, but I think it still works.

kira-syslogng:
Build SUCCESS

lbudai:

lbudai:
@kira-syslogng Do perftest

kira-syslogng:
Build SUCCESS

furiel:
`logthrdestdrv: init queue counters only when registered for the first time`
and
`logqueue: do not initialize counters in set method`
patches are reverted in the end, (maybe more?). From these, the first one is especially hard to understand and validate by review. Can you rebuild the patchset without such reverts? That would help the review a lot.

lib/stats/stats-cluster.c
```diff
@@ -244,15 +257,15 @@ stats_cluster_is_alive(StatsCluster *self, gint type)
 {
   g_assert(type < self->counter_group.capacity);
 
-  return ((1<<type) & self->live_mask);
+  return ((1<<type) & self->live_mask) == (1 << type);
```
furiel: For me: `return !!(1<<type) & self->live_mask)` would be more readable

lib/logqueue.c
```diff
@@ -168,18 +168,60 @@ log_queue_check_items(LogQueue *self, gint *timeout, LogQueuePushNotifyFunc para
...
 }
 
+static void
+_register_common_counters(LogQueue *self, gint stats_level, const StatsClusterKey *sc_key)
+{
+  gboolean need_to_reset_counters = TRUE;
+
+  stats_register_counter(stats_level, sc_key, SC_TYPE_QUEUED, &self->queued_messages);
+
+  if (stats_check_level(STATS_LEVEL1))
```
furiel: Why do we need this check? Can we just omit?

lib/logqueue.c
```diff
@@ -168,18 +168,60 @@ log_queue_check_items(LogQueue *self, gint *timeout, LogQueuePushNotifyFunc para
...
+    need_to_reset_counters = !stats_contains_counter(sc_key, SC_TYPE_MEMORY_USAGE);
```
furiel: We could go this check into _reset_counters.

lib/stats/stats-cluster.c
```diff
@@ -208,6 +208,19 @@ stats_cluster_track_counter(StatsCluster *self, gint type)
   return &self->counter_group.counters[type];
 }
 
+StatsCounterItem *
+stats_cluster_get_counter(StatsCluster *self, gint type)
```
furiel: In the end, this function is not used. Can you drop the patch?

lbudai: I want to keep it as I’ll need it later.

lbudai:
@furiel: thanks for the review notes.

I’ll check them but... but I’m not sure I want to rebuild the whole patchset. It is true that I’m blocking on this PR, but for me it is just a marginal issue what I had to solve and I want to focus on other things now.

lib/logqueue.c
lbudai: No, we cannot: if we wouldn't check this than in case of stats level 0 the `need_to_reset_counters` would be set to `FALSE` which would ended in an unitialized queued counter.
(note that `stats_counter_set` handles the case when NULL ptr passed to it as counter).

lib/logqueue.c
lbudai: No, it couldn't.
Why?
Because at that point the counter is registered... so it would return `TRUE`  (of course when stats-level is >= 1).

lib/logwriter.c
```diff
@@ -1349,17 +1345,11 @@ _register_counters(LogWriter *self)
       stats_register_counter(self->options->stats_level, &sc_key, SC_TYPE_SUPPRESSED, &self->suppressed_messages);
     stats_register_counter(self->options->stats_level, &sc_key, SC_TYPE_DROPPED, &self->dropped_messages);
     stats_register_counter(self->options->stats_level, &sc_key, SC_TYPE_PROCESSED, &self->processed_messages);
```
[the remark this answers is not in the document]
lbudai: What do you mean under `_common_counters`? 
`_common_counter` is at `LogQueue` implementation, a private (static) method for that, and has nothing to do with `LogWriter`.

lbudai:
I can do some modifications...
1) modify the predicate function (return `!!x`)
2) modifiy the code and use the `stats_cluster_get_counter` method

but only in if the PR get merged on today.

I still don't want to rebuild the whole patchset. 
I think, the first version could be also merged, but totally agreed with Bazsi that this naming is not the best. Changing the `LogQueue` interface is not part of the original problem, but a make-things-better changset, or, in other words: it could be part of a separate PR... but I'd not open another PR for it.

lbudai:
@furiel: you were too fast :) 
then I'm going to open a PR  for addressing those two.

PR #2280, opened 2018-09-14
syslog-ng-ctl config to print pre-processed configuration

Kokan:
Currently there is no way to tell what configuration syslog-ng has in effect.
Sure the file in the disk can be checked, and most of the time it would be save to assume the active config and the disk content are equal.

This gives an opportunity to both get the content of the top level configuration file and also the preprocessed configuration.

The way it could be queried via `syslog-ng-ctl`:
```
> syslog-ng-ctl 
Syntax: /tmp/install/sbin/syslog-ng-ctl <command> [options]
Possible commands are:
    stats                Get syslog-ng statistics in CSV format
...
    config               Print current config
```

This prints the top level configuration file:
```
> syslog-ng-ctl config
```

With an option `--preprocess` it prints the preprocessed version of the config:
```
> syslog-ng-ctl config --preprocessed
```

### DIFF
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -535,16 +535,16 @@ cfg_read_config(GlobalConfig *self, const gchar *fname, gboolean syntax_only, gc
   if ((cfg_file = fopen(fname, "r")) != NULL)
     {
       CfgLexer *lexer;
-      GString *preprocess_output = g_string_sized_new(8192);
+      self->preprocess_config = g_string_sized_new(8192);
 
-      lexer = cfg_lexer_new(self, cfg_file, fname, preprocess_output);
+      lexer = cfg_lexer_new(self, cfg_file, fname, self->preprocess_config);
       res = cfg_run_parser(self, lexer, &main_parser, (gpointer *) &self, NULL);
       fclose(cfg_file);
       if (preprocess_into)
         {
-          cfg_dump_processed_config(preprocess_output, preprocess_into);
+          cfg_dump_processed_config(self->preprocess_config, preprocess_into);
         }
-      g_string_free(preprocess_output, TRUE);
+
       if (res)
         {
           /* successfully parsed */
@@ -587,6 +587,9 @@ cfg_free(GlobalConfig *self)
 
   if (self->state)
     persist_state_free(self->state);
+
+  if (self->preprocess_config)
+    g_string_free(self->preprocess_config, TRUE);
 
   g_free(self);
 }
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -524,6 +524,21 @@ cfg_dump_processed_config(GString *preprocess_output, gchar *output_filename)
     }
 }
 
+static GString *
+_load_file_into_string(const gchar *fname)
+{
+  gchar *buff;
+  GString *content = g_string_new("");
+
+  if (g_file_get_contents(fname, &buff, NULL, NULL))
+    {
+      g_string_append(content, buff);
+      g_free(buff);
+    }
+
+  return content;
+}
+
 gboolean
 cfg_read_config(GlobalConfig *self, const gchar *fname, gboolean syntax_only, gchar *preprocess_into)
 {
@@ -536,6 +551,7 @@ cfg_read_config(GlobalConfig *self, const gchar *fname, gboolean syntax_only, gc
     {
       CfgLexer *lexer;
       self->preprocess_config = g_string_sized_new(8192);
+      self->original_config = _load_file_into_string(fname);
 
       lexer = cfg_lexer_new(self, cfg_file, fname, self->preprocess_config);
       res = cfg_run_parser(self, lexer, &main_parser, (gpointer *) &self, NULL);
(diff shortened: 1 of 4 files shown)


bazsi:
This is a good idea, but is it also that it gives you the preprocessed configuration? I would have the assumption that it would return the one that matches the disk file.

OTOH, it is true, that include files wouldn't be part of that, whereas this one contains them.

In that case, I think I would rename the syslog-ng-ctl command to reflect this fact, e.g. syslog-ng-ctl preprocessed-config (which I don't like too much) or running-config (if that's enough to indicate that it might not be equal to disk file) or something else. the key that it should be clear that it is different.

Kokan:
I was thinking about to make it possible to provide the original configuration and in such case add an additional option to `syslog-ng-ctl config` for example `--preprocessed`.

It was a quick patch, I did not want to complicate things. But you have a valid point, this should be cleared now.

Does this alternatives look good to you ?
`syslog-ng-ctl config` would return with the config without preprocessing.
`syslog-ng-ctl config --preprocessed` would return with the preprocessed configuration

This syntax would support an option like `--verify` later on, which could check against the config stored on disk.

kira-syslogng:
Build SUCCESS

lib/cfg.c
```diff
@@ -535,16 +535,16 @@ cfg_read_config(GlobalConfig *self, const gchar *fname, gboolean syntax_only, gc
...
-      lexer = cfg_lexer_new(self, cfg_file, fname, preprocess_output);
+      lexer = cfg_lexer_new(self, cfg_file, fname, self->preprocess_config);
       res = cfg_run_parser(self, lexer, &main_parser, (gpointer *) &self, NULL);
       fclose(cfg_file);
       if (preprocess_into)
         {
-          cfg_dump_processed_config(preprocess_output, preprocess_into);
+          cfg_dump_processed_config(self->preprocess_config, preprocess_into);
         }
-      g_string_free(preprocess_output, TRUE);
```
furiel: My question is, is it safe to move the g_string_free to cfg_free?

In the `main_loop_read_and_init_config` I see the only place when `cfg_free` is not called in the failure branch, that might leak this value. Can we add a `cfg_free` there? Otherwise I think g_string_free method could be moved into `cfg_free`.

lib/cfg.c
```diff
@@ -524,6 +524,27 @@ cfg_dump_processed_config(GString *preprocess_output, gchar *output_filename)
...
+  FILE  *fd = fopen(fname, "r");
+
+  if (!fd)
+    return NULL;
+
+  gchar buf[1024];
+  gsize res;
+  GString *result = g_string_new("");
+
+  while ((res = fread(buf, 1, sizeof(buf), fd)) > 0)
```
furiel: Optional:
If the goal is just to read the content, you could use `g_file_get_contents`.
https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-file-get-contents

Though in the end you need `GString`, though I would still use a library function to read into cstring, then convert to glib, as it would result in cleaner code, and performance is not important here.

lib/cfg.c
```diff
@@ -524,27 +524,25 @@ cfg_dump_processed_config(GString *preprocess_output, gchar *output_filename)
...
+_load_file_into_string(const gchar *fname)
 {
-  gint res;
-  CfgLexer *lexer;
-  GString *preprocess_output = g_string_sized_new(8192);
+  FILE  *fd = fopen(fname, "r");
 
-  lexer = cfg_lexer_new_buffer(self, config_string, strlen(config_string));
-  lexer->preprocess_output = preprocess_output;
+  if (!fd)
```
furiel: It would be nice to log the reason here.

lib/cfg.c
Kokan: Thanks, actually I wanted to get such function but my query was not successful, and I did not try that hard.

lib/cfg.c
Kokan: The file is opened again where the reason not able to open a file is logged. I wanted to avoid logging the same thing twice (the two open's result could differ, but I think that has a low chanse, while this failing this open is not critical, while the other is.)

lib/cfg.c
Kokan: I do not think there is much choice about using `cfg_free`, the `GlobalConfig` own's that string, that should be freed by `cfg_free`.

On the other hand your findings - after just a glimpse - looks valid, but that is a general issue. I do not want to mix that with this PR.

furiel: Okok.

kira-syslogng:
Build SUCCESS

lib/cfg.c
gaborznagy: Correct me if I'm wrong but I think that possible leak you mentioned does not have a high severity as it would only happen in an error case, after which syslog-ng is terminated.
This all happens on early startup, it's not affecting production environment.

Of course an error-handling is missing and we could build something else on `main_loop_read_and_init_config` too with the possibility to leak memory.

Although it is a nice catch! :+1:

PR #2281, opened 2018-09-14
Update data type to avoid conversion.

y-l-i:
The output of atomic_gssize_get_unsigned is gsize, and output of stats_counter_get is also gsize. Instead of gssize, declare result as gsize matches the unsigned type of the return value.

This fixes an alignment trap on ARMv5 when using with glib-2.56.x.

### DIFF
--- a/lib/stats/stats-counter.h
+++ b/lib/stats/stats-counter.h
@@ -75,7 +75,7 @@ stats_counter_set(StatsCounterItem *counter, gsize value)
 static inline gsize
 stats_counter_get(StatsCounterItem *counter)
 {
-  gssize result = 0;
+  gsize result = 0;
 
   if (counter)
     result = atomic_gssize_get_unsigned(&counter->value);


kira-syslogng:

Kokan:
@kira-syslogng ok to test

furiel:
@y-l-i thanks for the patch!

> The output of atomic_gssize_get_unsigned is gsize, and output of stats_counter_get is also gsize. Instead of gssize, declare result as gsize matches the unsigned type of the return value.

I understand and agree with this one. 

> This fixes an alignment trap on ARMv5 when using with glib-2.56.x.

Out of curiosity, can you share some details on this? I am yet to see where the misalignment can happen. Which variable is misaligned? How does it depend on glib version?

lbudai:
@furiel: maybe this is related to pointer alignments... (we are delegate atomic operations to `g_atomic_pointer`s in `atomic_gssize` impl.)

y-l-i:
@furiel I used latest syslog-ng release with latest glib-2.56.1 then upgrade to glib-2.56.2 on my ARMv5 based platform. Haven't tested with other versions.

The kernel reported alignment trap on every 10 minutes. It leads to the timer handler for STATS messages, then nailed down to here:
```
static void
stats_log_format_counter(StatsCluster *sc, gint type, StatsCounterItem *item, gpointer user_data)
{
  EVTREC *e = (EVTREC *) user_data;
  EVTTAG *tag;
  gchar buf[32];

  tag = evt_tag_printf(stats_cluster_get_type_name(sc, type), "%s(%s%s%s)=%"G_GSIZE_FORMAT,
                       stats_cluster_get_component_name(sc, buf, sizeof(buf)),
                       sc->key.id,
                       (sc->key.id[0] && sc->key.instance[0]) ? "," : "",
                       sc->key.instance,
                       stats_counter_get(&sc->counter_group.counters[type]));
  evt_rec_add_tag(e, tag);
}
```

When misalignment happens, it dereferences:
`stats_counter_get(&sc->counter_group.counters[type])`

So I guess the misalignment happens in either
```
typedef struct _StatsCluster
{
  StatsClusterKey key;
  StatsCounterGroup counter_group;
  guint16 use_count;
  guint16 live_mask;
  guint16 indexed_mask;
  guint16 dynamic:1;
  gchar *query_key;
} StatsCluster;
```
or 
```
struct _StatsCounterGroup
{
  StatsCounterItem *counters;
  const gchar **counter_names;
  guint16 capacity;
  void (*free_fn)(StatsCounterGroup *self);
```

When handling counters[id].value, the error happened. But it more looks like the below structure, which has guint16 declared and placed before the member counter_group.

```
struct _StatsClusterKey
{
  /* syslog-ng component/driver/subsystem that registered this cluster */
  guint16 component;
  const gchar *id;
  const gchar *instance;
  StatsCounterGroupInit counter_group_init;
};
```

furiel:
Thanks for the detailed analysis and the PR. With the 10 minutes interval, I suspect it is triggered by the periodic statistic logging in syslog-ng. As a workaround, you can disable it with stats-freq(0).
https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.17/administration-guide/54#TOPIC-989754

Approving now.

lbudai:
@y-l-i : but why the gssize ->gsize conversion solves the issue?
(Of course we can merge it, and will merge it, just want to be sure that it really solves the issue.)

PR #2282, opened 2018-09-15
dbld: fix dbld/rules deb failure

furiel:
Due to the missing unzip dependency, gradle could not be installed. Then dh_install failed because it missed some jar files. Then dpkg-buildpackage validation failed because not all packages were installed in the dh_install step.

I added the missing unzip dependency.

Also added set -e to functions.sh. Until now those functions could fail silently.


kira-syslogng:
Build SUCCESS

PR #2283, opened 2018-09-17
filter: detect loop

Kokan:
The following config would cause stack overflow because of recursive function call:
```
@version: 3.17

source s_inter { internal(); };

filter f { filter(f); };

log { source(s_inter); filter(f); };
```
Edit:
You should start `syslog-ng -F` (without enabling either of those switches: `-d -e -t -v`)

### DIFF
--- a/lib/filter/filter-call.c
+++ b/lib/filter/filter-call.c
@@ -30,6 +30,7 @@ typedef struct _FilterCall
   FilterExprNode super;
   FilterExprNode *filter_expr;
   gchar *rule;
+  gboolean visited; /* Used for filter call loop detection */
 } FilterCall;
 
 static gboolean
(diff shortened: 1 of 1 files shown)


lib/filter/filter-call.c
```diff
@@ -62,6 +63,13 @@ filter_call_init(FilterExprNode *s, GlobalConfig *cfg)
...
+  if (self->visited)
+    {
+      msg_error("Loop detected in filter rule", evt_tag_str("rule", self->rule));
+      return FALSE;
+    }
+  self->visited = TRUE;
+
   /* skip initialize if filter_call_init already called. */
   if (self->filter_expr)
     return TRUE;
```
lbudai: Is it possible that `self->filter_expr` is not `NULL`  when we check the `visited` flag?

PR #2285, opened 2018-09-17
lib/gsockaddr.c: modify the unix salen calculation

szemere:
It is not required on all platforms to add the size of the
terminating null character (+1) to the value of the `addrlen`
parameter during the `bind` system call. Most platforms has a
built in MACRO to calculate the proper value, called `SUN_LEN`.
SUN_LEN is **not** a POSIX standard, so we use the formula from
the Linux man page if it is not available.

Signed-off-by: Laszlo Szemere <laszlo.szemere@balabit.com>

### DIFF
--- a/lib/gsockaddr.c
+++ b/lib/gsockaddr.c
@@ -529,21 +529,6 @@ static GSockAddrFuncs unix_sockaddr_funcs =
   .format = g_sockaddr_unix_format
 };
 
-/*
-  SUN_LEN is not a POSIX standard, thus not available on all platforms.
-  If it is available we should rely on it. Otherwise we use the formula
-  from the Linux man page.
-*/
-static int
-_calculate_salen(GSockAddrUnix *addr)
-{
-#ifdef SUN_LEN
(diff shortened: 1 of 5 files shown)


lib/gsockaddr.c
```diff
@@ -529,6 +529,21 @@ static GSockAddrFuncs unix_sockaddr_funcs =
...
+static int
+_calculate_salen(GSockAddrUnix *addr)
+{
+#ifdef SUN_LEN
+  return SUN_LEN(&(addr->saun));
+#else
+  return sizeof(addr->saun) - sizeof(addr->saun.sun_path) + strlen(addr->saun.sun_path) + 1;
+#endif
+}
+
```
lbudai: So we have a platform where `SUN_LEN` is not defined? :)
I'd define `SUN_LEN` in a `compat` header.
What do you think?

an alternative impl.:
http://lists.llvm.org/pipermail/lldb-commits/Week-of-Mon-20151026/024736.html
Does this have the same result?

lbudai:
@szemere : Thanks!

PR #2286, opened 2018-09-18
json: fix the literal() type hint

bazsi:
This makes it possible to add snippets of literal values into the
json, without being represented as strings.

$(format-json bar=literal($(format-json --subkeys .foo.)))

Signed-off-by: Balazs Scheidler <balazs.scheidler@oneidentity.com>

PR #2287, opened 2018-09-18
Value pairs improvements

bazsi:
This branch adds a couple of improvements to our value-pairs functionality:

1) adds support for --shift-levels key rewrite method, a bit similar to --shift, but instead of working with characters, it works with dot delimited "levels". --shift-levels 2 drops the prefix up to the second dot in the name of the key. e.g. .iptables.SRC becomes SRC (the initial dot is calculated too).

2) fixes a bug in type hint parsing, in case the hinted template expression contains parenthesis on its own. This is also related to #2286 where we actually make use of literal type hints in connection with $(format-json)

3) it adds a scope named "none" that resets the previously added scopes, making it possible to get rid of automatically added name-value pairs. This is not really changing functionality, however improves performance as the internal table of macros need not be consulted.

The 3rd option works like this:

$(format-welf --scope none .iptables.*)

This way the internal table of macros are not attempted to be matched against the wild-card. This is basically a performance hack, as I think the default behavior is not good.

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -332,13 +332,14 @@ extern struct _StatsOptions *last_stats_options;
 %token KW_KEY                         10504
 %token KW_SCOPE                       10505
 %token KW_SHIFT                       10506
-%token KW_REKEY                       10507
-%token KW_ADD_PREFIX                  10508
-%token KW_REPLACE_PREFIX              10509
+%token KW_SHIFT_LEVELS                10507
+%token KW_REKEY                       10508
+%token KW_ADD_PREFIX                  10509
+%token KW_REPLACE_PREFIX              10510
 
-%token KW_ON_ERROR                    10510
+%token KW_ON_ERROR                    10511
 
-%token KW_RETRIES                     10511
+%token KW_RETRIES                     10512
 
 /* END_DECLS */
 
@@ -1393,7 +1394,8 @@ vp_rekey_options
 	;
 
 vp_rekey_option
-	: KW_SHIFT '(' LL_NUMBER ')' { value_pairs_transform_set_add_func(last_vp_transset, value_pairs_new_transform_shift($3)); }
+	: KW_SHIFT '(' positive_integer ')' { value_pairs_transform_set_add_func(last_vp_transset, value_pairs_new_transform_shift($3)); }
+	| KW_SHIFT_LEVELS '(' positive_integer ')' { value_pairs_transform_set_add_func(last_vp_transset, value_pairs_new_transform_shift_levels($3)); }
 	| KW_ADD_PREFIX '(' string ')' { value_pairs_transform_set_add_func(last_vp_transset, value_pairs_new_transform_add_prefix($3)); free($3); }
 	| KW_REPLACE_PREFIX '(' string string ')' { value_pairs_transform_set_add_func(last_vp_transset, value_pairs_new_transform_replace_prefix($3, $4)); free($3); free($4); }
 	;
--- a/lib/value-pairs/cmdline.c
+++ b/lib/value-pairs/cmdline.c
@@ -169,7 +169,7 @@ value_pairs_parse_type(gchar *spec, gchar **value, gchar **type)
       return;
     }
 
-  ep = strchr(sp, ')');
+  ep = strrchr(sp, ')');
   if (ep == NULL || ep[1] != '\0')
     {
       *value = spec;
--- a/lib/value-pairs/cmdline.c
+++ b/lib/value-pairs/cmdline.c
@@ -331,6 +331,15 @@ vp_cmdline_parse_rekey_shift (const gchar *option_name, const gchar *value,
   gpointer *args = (gpointer *) data;
   ValuePairsTransformSet *vpts = (ValuePairsTransformSet *) args[2];
   gchar *key = (gchar *) args[3];
+  gchar *end = NULL;
+  gint number_to_shift = strtol(value, &end, 0);
+
+  if (number_to_shift <= 0 || *end != 0)
+    {
+      g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
+                   "Error parsing value-pairs, argument to --shift is not numeric or not a positive number");
+      return FALSE;
+    }
 
   vpts = vp_cmdline_rekey_verify (key, vpts, data);
   if (!vpts)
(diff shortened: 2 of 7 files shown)


kira-syslogng:
Build SUCCESS

lib/cfg-grammar.y
```diff
@@ -1394,6 +1395,7 @@ vp_rekey_options
 
 vp_rekey_option
 	: KW_SHIFT '(' LL_NUMBER ')' { value_pairs_transform_set_add_func(last_vp_transset, value_pairs_new_transform_shift($3)); }
+	| KW_SHIFT_LEVELS '(' LL_NUMBER ')' { value_pairs_transform_set_add_func(last_vp_transset, value_pairs_new_transform_shift_levels($3)); }
```
Kokan: If I understand correctly the main use case for `shift-level` would be to cut the `--key` part, though it is capable more also it possible to misuse by accident. (The level of the key and the supplied value does not match.)

In that case I think it would be better to have an option like `--skip-key-prefix` which would automatically skip the key part.

Also the `shift-levels` has a side effect when the key name could be not unique anymore, which one should be preferred ? 
```
.a.b.c
.a.c.c
```
With shift level 3.

bazsi: Value-pairs would deduplicate the result, ie would use one of them, I think the last one.

But this is not necessarily a real problem. The convention in syslog-ng is to use an application specific prefix, so most value pairs would be unique under that prefix.

If there are exceptions (e.g. .unix or .SDATA) they can be handled separately.

Btw, the same can be true for --shift or any other key manipulation that removes parts from the key.

I don't understand --skip-key-prefix, can you elaborate on that?

Kokan: Based on this example: `$(format-json --key .SDATA.* --shift-levels 2)` and also on how I would use the `--shift-levels` option, for me it would be to drop the `.SDATA.` part of the key. If that is true, I think a simplified option could be used instead of `--shift-levels`.

Instead asking the user to count the dots in the key, syslog-ng could do the same with an option like `--skip-key-prefix` (maybe just `--skip-key`).

The `$(format-json --key <key> --skip-key)` would be equal to the `$(format-json --key <key> --shift-levels <number of dots in key>)`.

lib/value-pairs/transforms.c
```diff
@@ -152,6 +151,39 @@ value_pairs_new_transform_shift (gint amount)
...
+  gint levels_left = self->amount - 1;
+
+  dot = strchr(key->str, '.');
+  while (dot && levels_left > 0)
+    {
+      dot = strchr(dot + 1, '.');
+      levels_left--;
+    }
+
+  g_string_erase(key, 0, dot + 1 - key->str);
```
furiel: Is there always dot in a name? If not, this will assert in glib.

lib/value-pairs/transforms.c
```diff
@@ -152,6 +151,39 @@ value_pairs_new_transform_shift (gint amount)
...
+/* shift-levels() */
+
+static void
+vp_trans_shift_levels(ValuePairsTransform *t, GString *key)
+{
+  VPTransShift *self = (VPTransShift *)t;
+  const gchar *dot;
+  gint levels_left = self->amount - 1;
+
+  dot = strchr(key->str, '.');
```
furiel: Now we use `LL_NUMBER` in grammar, which allows 0 or less. For those cases, we will also shift one level. Maybe it would better use `positive_integer` in the grammar.

lib/cfg-grammar.y
```diff
@@ -332,13 +332,14 @@ extern struct _StatsOptions *last_stats_options;
 %token KW_KEY                         10504
 %token KW_SCOPE                       10505
 %token KW_SHIFT                       10506
-%token KW_REKEY                       10507
-%token KW_ADD_PREFIX                  10508
-%token KW_REPLACE_PREFIX              10509
+%token KW_SHIFT_LEVELS                10507
```
furiel: Just a minor thing but I think changing these id's will break the binary ABI. If a maintainer wants to port it back, all dependant modules need to be recompiled.

lib/value-pairs/cmdline.c
```diff
@@ -340,8 +340,29 @@ vp_cmdline_parse_rekey_shift (const gchar *option_name, const gchar *value,
...
+  vpts = vp_cmdline_rekey_verify (key, vpts, data);
+  if (!vpts)
+    {
+      g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
+                   "Error parsing value-pairs: --shift-levels used without --key or --rekey");
+      return FALSE;
+    }
+
+  value_pairs_transform_set_add_func(vpts,
+                                     value_pairs_new_transform_shift_levels (atoi (value)));
```
furiel: We should exclude zero or less for atoi(value).

lib/value-pairs/value-pairs.c
```diff
@@ -822,8 +822,17 @@ value_pairs_add_scope(ValuePairs *vp, const gchar *scope)
 {
   gboolean result;
 
-  result = cfg_process_flag(value_pair_scope, vp, scope);
-  vp_update_builtin_list_of_values(vp);
+  if (strcmp(scope, "none") != 0)
```
furiel: Can you add an example (preferably to the PR description), how a user can exploit this?

lib/cfg-grammar.y
bazsi: Hmm.. nothing restricts syslog-ng name-value pairs to only have 1, 2 or 3 or more dots. Howlong prefix would --skip-key-prefix drop?

BTW: I tend to agree that value-pairs needs a serious reevaluation of its functionality, this is basically a quick hack to make name-value pairs map to Splunk nicely.

Kokan: It would use the string provided in the `--key` to determine the level, so in your example `--key .SDATA.*` it has 2 dots, therefore 2 level.

Example:
```
nv-pairs
a.b.c.d
a.b.d.e
a.c.g.f
```
The `--key a.b.* --skip-key` would result in: `c.d` and `d.e`.
The `--key a.b.c.* --skip-key` would result in `d`.

bazsi: ah.. that's actually a good idea. let me think about that.

lib/value-pairs/transforms.c
bazsi: fixed. dot is not guaranteed.

lib/value-pairs/transforms.c
bazsi: fixed both for --shift and --shift-levels

lib/cfg-grammar.y
bazsi: true, but I don't think we have ABI guarantees.

lib/value-pairs/cmdline.c
bazsi: resolved with using positive_integer in the grammar.

bazsi: wait, this is the cmdline parser...

bazsi: I've added validation and error handling.

lib/value-pairs/value-pairs.c
furiel: Thanks

kira-syslogng:
Build SUCCESS

lib/cfg-grammar.y
bazsi: I gave it some thought, and could be implemented at this point as the wildcard would be available at the transformation rule.

My rationale:
1) I am generally unhappy with the value-pairs implementation and I am not sure the wildcard would be readily available to the transformation in the new one. This change does not introduce any new concepts/interdependencies in the command line syntax.

2) the wildcard might not be based on the prefix, in that case, it'd be difficult to decide what to do with --skip-key.

With that rationale, I'd keep the implementation as is. The value-pairs syntaxes are in a sorry state and should be overhauled at one point in the future. But since I am working something very different now, I wouldn't start that work. With this, I'd resolve this conversation.

kira-syslogng:
Build SUCCESS

lib/value-pairs/cmdline.c
```diff
@@ -340,8 +340,38 @@ vp_cmdline_parse_rekey_shift (const gchar *option_name, const gchar *value,
       return FALSE;
     }
 
-  value_pairs_transform_set_add_func
-  (vpts, value_pairs_new_transform_shift (atoi (value)));
+  value_pairs_transform_set_add_func(vpts,
+                                     value_pairs_new_transform_shift (atoi (value)));
```
furiel: From PR perspective this was just a whitespace change, so this is an optional request, but the same value validation could be applied here too.

lib/value-pairs/cmdline.c
```diff
@@ -340,8 +340,38 @@ vp_cmdline_parse_rekey_shift (const gchar *option_name, const gchar *value,
...
+  vpts = vp_cmdline_rekey_verify (key, vpts, data);
+  if (!vpts)
+    {
+      g_set_error (error, G_OPTION_ERROR, G_OPTION_ERROR_FAILED,
+                   "Error parsing value-pairs: --shift-levels used without --key or --rekey");
+      return FALSE;
+    }
+
+  value_pairs_transform_set_add_func(vpts,
+                                     value_pairs_new_transform_shift_levels (atoi (value)));
```
furiel: Thanks for adding the validation. A few lines above we convert the `value` string to `integer` with `strol`, but in this line we redo the calculation with `atoi`. My question is, do we need to use different conversion function, or even would be better, can we reuse the already calculated integer here too?

bazsi: ugh, sorry. I certainly meant that. Thank you for catching this.

bazsi:
I've fixed the shift-levels omission and did the same for --shift.

PR #2288, opened 2018-09-18
filter: multiple filter ref incorrect loop detection

Kokan:
In case of skip initialized the `visited` variable was not set back to `FALSE`, which would cause false loop detection if the same `filter` filter would be used twice.


lbudai:
I know that it would be to unfair to request a test... but maybe in a separate PR... ?

PR #2289, opened 2018-09-18
templates: get rid off the args_lock

bazsi:
This patch gets rid off the argument locking in template functions, thereby making the use of the same
template (with template functions) from multiple threads scale.

[Edit by Kokan]
This patch also limits the template function arguments to 64.

### DIFF
--- a/lib/template/function.h
+++ b/lib/template/function.h
@@ -29,18 +29,14 @@
 #include "plugin-types.h"
 #include "common-template-typedefs.h"
 
+#define TEMPLATE_INVOKE_MAX_ARGS 64
+
 /* This structure contains the arguments for template-function
  * expansion. It is defined in a struct because otherwise a large
  * number of function arguments, that are passed around, possibly
  * several times. */
 typedef struct _LogTemplateInvokeArgs
 {
-  /* scratch buffers, stores GString *, elements are managed by the
-   * function, storage/free is performed by the core. Can be used to
-   * avoid allocating GString buffers in the fast-path. */
-
-  GPtrArray *bufs;
-
   /* context in case of correllation */
   LogMessage **messages;
   gint num_messages;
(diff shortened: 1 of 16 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/template/function.h
```diff
@@ -29,18 +29,14 @@
 #include "plugin-types.h"
 #include "common-template-typedefs.h"
 
+#define TEMPLATE_INVOKE_MAX_ARGS 64
```
Kokan: Is it okay to limit the maximum arguments ? How did you come up with this number ?

lib/template/tests/test_template.c
```diff
@@ -126,6 +132,7 @@ setup(void)
 void
 teardown(void)
 {
+  scratch_buffers_explicit_gc();
```
Kokan: The scratch buffer is only used in the thread `format_template_thread`, do we really have to call gc here?

lib/template/function.h
bazsi: I consider it ok :) I am yet to see a config where we've used more than 15-20, that being the ewmm template function. There I'd even want to decrease the number by adding a couple of features to value-pairs.

If this is deemed too small, it's pretty easy to increment it, but by having this limit the code is much simpler. It is also an option to make it dynamic, but that's substantially more complex code.

I think it should be good for now.

lib/template/tests/test_template.c
bazsi: The scratch buffers system is also used in other tests, but in those threads app_startup() covers the initialization.

With this change scratch-buffers are used in all template function invocations, be them either multi or single threaded.

lib/template/function.h
Kokan: Okay, but because of this I mark it as user visible feature, we should at least document this limitation. (I'll notify the doc team.)

PR #2290, opened 2018-09-18
app-parser: set ${.app.name} to the application identified

bazsi:
Rationale:
* whenever we are sending structured data upstream, this name-value tag grabs the "kind" of application we parsed. This could be used as "sourcetype" in splunk. The same information is also available as a tag, however it is not easy to use that in an output template.


furiel:
@bazsi do we need both? Maybe set-tag could be dropped. Otherwise the patch looks good to me.

Issue #2276, opened 2018-09-14
Missing flush on deinit in LogThreadedDestDriver


gaborznagy:
## Version: 
3.17.2
## Description:

During review of #2098 we realized that there is no flush when worker thread is de-inited.
Quoting from 2098:

> The perform_work() did an explicit flush after the loop, however you are be
right that with flush-timeout() this might have changed.
> 
> This means that the thread might exit without flushing messages, but this
shouldn't be in thread_deinit even in this case.
> 
> We should do an explicit flush, and possibly a rewind when the thread exits.

MrAnno:
Fixed in #2291.

PR #2291, opened 2018-09-18
Http multi threading

bazsi:
This branch implements multithreaded http destination, that uses the specified number of worker threads to send messages to an upstream HTTP server.

This should drastically improve performance of HTTP based transports, especially if combined with #2289, #2287, #2286

### DIFF
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -653,7 +653,7 @@ _worker_thread(gpointer arg)
   iv_deinit();
   return;
 
- error:
+error:
   _signal_startup_failure(self);
   iv_deinit();
 }
@@ -661,9 +661,8 @@ _worker_thread(gpointer arg)
 gboolean
 log_threaded_dest_worker_init_method(LogThreadedDestWorker *self)
 {
-  self->queue = log_dest_driver_acquire_queue(
-                         &self->owner->super,
-                         _format_queue_persist_name(self));
+  self->queue = log_dest_driver_acquire_queue(&self->owner->super,
+                                              _format_queue_persist_name(self));
 
   if (!self->queue)
     return FALSE;
@@ -937,7 +936,7 @@ log_threaded_dest_driver_init_method(LogPipe *s)
   _register_stats(self);
 
   self->shared_seq_num = GPOINTER_TO_INT(cfg_persist_config_fetch(cfg,
-                                                                  _format_seqnum_persist_name(self)));
+                                         _format_seqnum_persist_name(self)));
   if (!self->shared_seq_num)
     init_sequence_number(&self->shared_seq_num);
 
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -56,8 +56,8 @@ log_threaded_dest_driver_ack_messages(LogThreadedDestDriver *self, gint batch_si
 {
   log_queue_ack_backlog(self->worker.instance.queue, batch_size);
   stats_counter_add(self->written_messages, batch_size);
-  self->retries_counter = 0;
-  self->batch_size -= batch_size;
+  self->worker.instance.retries_counter = 0;
+  self->worker.instance.batch_size -= batch_size;
 }
 
 void
@@ -65,16 +65,16 @@ log_threaded_dest_driver_drop_messages(LogThreadedDestDriver *self, gint batch_s
 {
   log_queue_ack_backlog(self->worker.instance.queue, batch_size);
   stats_counter_add(self->dropped_messages, batch_size);
-  self->retries_counter = 0;
-  self->batch_size -= batch_size;
+  self->worker.instance.retries_counter = 0;
+  self->worker.instance.batch_size -= batch_size;
 }
 
 void
 log_threaded_dest_driver_rewind_messages(LogThreadedDestDriver *self, gint batch_size)
 {
   log_queue_rewind_backlog(self->worker.instance.queue, batch_size);
-  self->rewound_batch_size = self->batch_size;
-  self->batch_size -= batch_size;
+  self->worker.instance.rewound_batch_size = self->worker.instance.batch_size;
+  self->worker.instance.batch_size -= batch_size;
 }
 
 static gboolean
@@ -90,7 +90,7 @@ _should_flush_now(LogThreadedDestDriver *self)
 
   iv_validate_now();
   now = iv_now;
-  diff = timespec_diff_msec(&now, &self->last_flush_time);
+  diff = timespec_diff_msec(&now, &self->worker.instance.last_flush_time);
 
   return (diff >= self->flush_timeout);
 }
@@ -158,7 +158,7 @@ _shutdown_event_callback(gpointer data)
 static void
 _suspend(LogThreadedDestDriver *self)
 {
-  self->suspended = TRUE;
+  self->worker.instance.suspended = TRUE;
 }
 
 /* NOTE: runs in the worker thread */
@@ -196,21 +196,21 @@ _disconnect_and_suspend(LogThreadedDestDriver *self)
 static void
 _accept_batch(LogThreadedDestDriver *self)
 {
-  log_threaded_dest_driver_ack_messages(self, self->batch_size);
+  log_threaded_dest_driver_ack_messages(self, self->worker.instance.batch_size);
 }
 
 /* NOTE: runs in the worker thread */
 static void
 _drop_batch(LogThreadedDestDriver *self)
 {
-  log_threaded_dest_driver_drop_messages(self, self->batch_size);
+  log_threaded_dest_driver_drop_messages(self, self->worker.instance.batch_size);
 }
 
 /* NOTE: runs in the worker thread */
 static void
 _rewind_batch(LogThreadedDestDriver *self)
 {
-  log_threaded_dest_driver_rewind_messages(self, self->batch_size);
+  log_threaded_dest_driver_rewind_messages(self, self->worker.instance.batch_size);
 }
 
 static void
@@ -221,22 +221,22 @@ _process_result(LogThreadedDestDriver *self, gint result)
     case WORKER_INSERT_RESULT_DROP:
       msg_error("Message(s) dropped while sending message to destination",
                 evt_tag_str("driver", self->super.super.id),
-                evt_tag_int("batch_size", self->batch_size));
+                evt_tag_int("batch_size", self->worker.instance.batch_size));
 
       _drop_batch(self);
       _disconnect_and_suspend(self);
       break;
 
     case WORKER_INSERT_RESULT_ERROR:
-      self->retries_counter++;
+      self->worker.instance.retries_counter++;
 
-      if (self->retries_counter >= self->retries_max)
+      if (self->worker.instance.retries_counter >= self->retries_max)
         {
           msg_error("Multiple failures while sending message(s) to destination, message(s) dropped",
                     evt_tag_str("driver", self->super.super.id),
                     log_expr_node_location_tag(self->super.super.super.expr_node),
-                    evt_tag_int("retries", self->retries_counter),
-                    evt_tag_int("batch_size", self->batch_size));
+                    evt_tag_int("retries", self->worker.instance.retries_counter),
+                    evt_tag_int("batch_size", self->worker.instance.batch_size));
 
           _drop_batch(self);
         }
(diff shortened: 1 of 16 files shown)


kira-syslogng:
Build FAILURE

bazsi:
This one depends on #2279, as I've removed the patch from the series that works around a problem, and without that, this one crashes.

My expectation is that if #2279 goes in, this should turn green, sans any unrelated problems (which now is riemann).

I just wanted to publish this, for anyone to see what I've been working on in the recent weeks.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

Kokan:
@bazsi could you please resolve the conflict ?

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
This now depends on #2302 that it also implicitly contains. If that goes in, this could be next.

I am doing another round of manual testing if it survived all the rebase rounds :)

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
@kira-syslogng do perftest

kira-syslogng:
Build FAILURE

lbudai:
@pzoleex : is this the same issue that you've found internally?

lbudai:
@bazsi : additional information: the issue is not introduced by your PR, we think it is triggered by the stats related changes (it is possible that it is not _caused_ by that stats PR but somehow it triggered something...), so your PR still mergeable and doesn't require any action from you

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/http/http.c
```diff
@@ -202,20 +202,40 @@ _add_message_to_batch(HTTPDestinationWorker *self, LogMessage *msg)
...
+      break;
+    case 4:
+      msg_notice("Server returned with a 4XX (client errors) status code, which means we are not "
+                 "authorized or the URL is not found. Trying again",
+                 evt_tag_str("url", owner->url),
+                 evt_tag_int("status_code", http_code),
+                 log_pipe_location_tag(&owner->super.super.super.super));
       retval = WORKER_INSERT_RESULT_ERROR;
       break;
     default:
```
furiel: Cosmetic, but I would prefer default statement to be the last case.

modules/http/http.c
```diff
@@ -24,569 +24,656 @@
...
+                 evt_tag_str("url", owner->url),
+                 evt_tag_int("status_code", http_code),
+                 log_pipe_location_tag(&owner->super.super.super.super));
+      retval = WORKER_INSERT_RESULT_ERROR;
+      break;
+    default:
+      msg_error("Unknown HTTP response code",
+                evt_tag_str("url", owner->url),
+                evt_tag_int("status_code", http_code),
+                log_pipe_location_tag(&owner->super.super.super.super));
```
furiel: We should set an error code in the default case too. Otherwise memory garbage is returned.

lib/logthrdestdrv.h
```diff
@@ -63,6 +65,7 @@ struct _LogThreadedDestWorker
   gint batch_size;
   gint rewound_batch_size;
   gint retries_counter;
+  gint seq_num;
```
furiel: I think this should remain gint32: step_sequence_number_atomic returns with gint32.

lib/logthrdestdrv.c
```diff
@@ -888,9 +931,9 @@ log_threaded_dest_driver_init_instance(LogThreadedDestDriver *self, GlobalConfig
   self->time_reopen = -1;
   self->flush_lines = -1;
   self->flush_timeout = -1;
-  self->worker.instance.batch_size = 0;
 
   self->retries_max = MAX_RETRIES_OF_FAILED_INSERT_DEFAULT;
   self->lock = g_mutex_new();
-  self->worker.instance.started_up = g_cond_new();
+  log_threaded_dest_worker_init_instance(&self->worker.instance, self);
+  _init_worker_compat_layer(&self->worker.instance);
```
furiel: This line should be moved before init_instance (->init_watches)

lib/logthrdestdrv.c
```diff
@@ -804,6 +850,17 @@ log_threaded_dest_driver_set_max_retries(LogDriver *s, gint max_retries)
   self->retries_max = max_retries;
 }
 
+LogThreadedDestWorker *
+_lookup_worker(LogThreadedDestDriver *self, LogMessage *msg)
```
furiel: I would call select_worker or pick_worker or similar. As we have an array of workers, maybe we need a real lookup method later.

lib/logthrdestdrv.c
```diff
@@ -886,7 +944,16 @@ log_threaded_dest_driver_init_method(LogPipe *s)
...
 
-  return _start_worker_thread(self);
+  /* free previous workers array if set to cope with num_workers change */
+  g_free(self->workers);
+  self->workers = g_new0(LogThreadedDestWorker *, self->num_workers);
+
+  gboolean startup_success = TRUE;
+  for (gint i = 0; startup_success && i < self->num_workers; i++)
+    {
+      startup_success &= _start_worker_thread(self);
```
furiel: The deinit method needs similar code for the those threads that are successfully initialized. self->workers_started might help with that.

furiel:
@bazsi about the last patch: http: split HTTPDestinationWorker, can you explain the changes or a general idea behind it?

I expect that should be a straigthforward patch that just follows up the changes in the logthrdestdrv driver. But  as methods are moved around in the same patch, it makes difficult to follow the diff.

Should we look out for something special about that patch?

lib/logthrdestdrv.c
furiel: My bad, confused with start_watches. Deleting my comment.

lib/logthrdestdrv.c
bazsi: Agree

modules/http/http.c
```diff
@@ -202,20 +202,40 @@ _add_message_to_batch(HTTPDestinationWorker *self, LogMessage *msg)
...
+_map_http_status_to_worker_status(HTTPDestinationWorker *self, glong http_code)
 {
+  HTTPDestinationDriver *owner = (HTTPDestinationDriver *) self->super.owner;
   worker_insert_result_t retval;
 
   switch (http_code/100)
     {
-    case 4:
-      retval = WORKER_INSERT_RESULT_DROP;
+    case 1:
```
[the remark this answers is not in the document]
bazsi: good catch, I have set this to return an ERROR and not NOT_CONNECTED. I think this should be rather rare, and if our retry mechanism does not solve it, we shouldn't repeat the same batch.

I have fixed it in a followup patch.

kira-syslogng:
Build FAILURE

modules/http/http.c
bazsi: the default falls through to the 5XX case that's why it was in that order. But I've added a followup patch, that among other things makes "default" the last one.

modules/http/http.c
bazsi: the default was falling through to the 5xx case, so a default was set. but my followup patch makes that absolutely clear.

lib/logthrdestdrv.h
bazsi: good catch.

kira-syslogng:
Build FAILURE

lib/logthrdestdrv.c
bazsi: workers are automatically terminated at exit, so I think they are cleared up just fine.

main_loop_create_worker_thread() gets a function callback that is invoked when a thread needs to exit due to a reload/stop being processed. In that case _request_worker_exit() gets called, which sets a variable and posts an iv_event.

Then the mainloop mechanism will wait for all worker threads to shut down and that's when deinit() gets called. By that time workers exited.

I think that addresses the issue, do you agree?

bazsi:
Responding to https://github.com/balabit/syslog-ng/pull/2291#issuecomment-425629002, @furiel 

Unfortunately at one point I squashed the http refactors into a single patch, but it is should be nothing but extracting HTTPDestinationWorker from the Driver, changing the signatures of all functions and moving code so that code for each of the classes are in one block. I'll see if I can resplit the patch. The riemann destination was similarly refactored, but I did that in 3 steps, but it's not going to be simpler than that.

bazsi:
I have now extracted the biggest chunk of code moves around in that patch, making it much easier to review. The patch that moves the code around was actually redone now and squashed out of the refactor, so I am 100% certain that no other changes than code moves are in that patch.

I hope this helps with the review.

lib/logthrdestdrv.c
furiel: I was thinking that the individual workers also have init deinit mechanism, but that's not the case as only the "master" has init-deinit, hence was my confusion. I agree. Marking as resolved.

furiel:
Thank you @bazsi for the resplit! It helped a lot with the review.

gaborznagy:
Fixes #2276 !

bazsi:
On Fri, Oct 5, 2018, 14:27 Gábor Nagy <notifications@github.com> wrote:

> *@gaborznagy* commented on this pull request.
>
> In lib/logthrdestdrv.c
> <https://github.com/balabit/syslog-ng/pull/2291#discussion_r223025110>:
>
> >
>    _schedule_restart(self);
>    iv_main();
>
> +  log_threaded_dest_worker_flush(self);
> +  log_queue_rewind_backlog_all(self->queue);
>
> It's a minor thing, not affecting behaviour:
> Shouldn't we highlight that rewinding after flush should happen only if
> the flush failed and the queue has remaining messages?
> We could use the log_threaded_dest_worker_flush return value.
>

rewinding an empty backlog will do nothing. I don't see much value in
checking that.

Actually I am rather thinking about a flush-on-exit() parameter that would
control whether:

1) we flush batched data but only that
2) we flush the entire queue
3) we don't flush at all.

The default for persistent queues would be 1), non persistent queues would
be 2). This would allow the user to solve cases where the destination is
slow, and we dont want to wait the flush of the entire queue on each reload.


>
> In lib/logthrdestdrv.c
> <https://github.com/balabit/syslog-ng/pull/2291#discussion_r223039070>:
>
> >    self->startup_finished = TRUE;
> +  self->startup_failure |= thread_failure;
>
> Just out of curiosity: why do we need to use bitwise OR when setting
> failure?
> It looks like if workers can be started multiple times and we don't want a
> failure to be masked out.
>
We want to fail if any of the workers failed.

lib/logthrdestdrv.c
```diff
@@ -555,70 +559,292 @@ _init_watches(LogThreadedDestDriver *self)
 }
 
 static void
-_signal_startup_finished(LogThreadedDestDriver *self)
+_signal_startup_finished(LogThreadedDestWorker *self, gboolean thread_failure)
 {
-  g_mutex_lock(self->lock);
+  g_mutex_lock(self->owner->lock);
   self->startup_finished = TRUE;
+  self->startup_failure |= thread_failure;
```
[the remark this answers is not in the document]
gaborznagy: > We want to fail if any of the workers failed.
My point was that `startup_failure` is a Worker attribute and not the Driver's, therefore I don't think it can be evaluated multiple times, hence .
But it was just a question, it caught my attention, it looked interesting. :)

lib/logthrdestdrv.h
```diff
@@ -116,7 +119,13 @@ struct _LogThreadedDestDriver
...
 
   gint stats_source;
-  gint32 seq_num;
+
+  /* this counter is not thread safe if there are multiple worker threads,
+   * in that case, one needs to use LogThreadedDestWorker->seq_num, which is
+   * static for a single insert() invocation, whereas this might be
+   * increased in parallel by the multiple threads. */
+
+  gint32 shared_seq_num;
```
lbudai: Shouldn't we use a 64 bit atomic counter (`atomic-gsize`) for `shared_seq_num`?
It will overflow after ~3hours when the message rate is 400k/sec.
Or it's not a problem as it is used with timestamp?  and (currently :) ) we cannot do 2^32-1 msg/sec...

lib/logthrdestdrv.c
bazsi: I would change this name in a followup branch, not this one, as it already received some feedback and I have stacks of followup work on top of this.

But please call me on this should I not address that in the load balancing patches.

Thanks.

lib/logthrdestdrv.h
bazsi: It's only meant to be used for ordering and is explicitly sized as 32 bits:

quoting rfc5424:

7.3.1.  sequenceId

   The "sequenceId" parameter tracks the sequence in which the
   originator submits messages to the syslog transport for sending.  It
   is an integer that MUST be set to 1 when the syslog function is
   started and MUST be increased with every message up to a maximum
   value of 2147483647.  If that value is reached, the next message MUST
   be sent with a sequenceId of 1.

PR #2295, opened 2018-09-20
Revert sticky hook option

Kokan:
The PR which introduced the `reopen` also changes the hook behavior optionally, so the user could register a hook that is not removed after it was triggered.
My issue with that solution that the code that triggered a hooks was responsible to decide if all of the triggered hooks should be removed, or kept.
While I think it would be more reasonable for the code that registers to decide if it want the registered hook to be triggered always or just once.

This PR makes all the hook one shoot again, and registers the `reopen` hook, when it was triggered.

### DIFF
--- a/lib/apphook.c
+++ b/lib/apphook.c
@@ -98,21 +98,18 @@ static void
 _update_current_state(gint type)
 {
   if (AH_REOPEN == type)
-     return;
+    return;
 
   g_assert(current_state <= type);
   current_state = type;
 }
 
 static void
-run_application_hook(gint type, gboolean remove_hook)
+run_application_hook(gint type)
 {
   GList *l, *l_next;
 
-  if (remove_hook)
-    {
-      _update_current_state(type);
-    }
+  _update_current_state(type);
 
   msg_debug("Running application hooks", evt_tag_int("hook", type));
   for (l = application_hooks; l; l = l_next)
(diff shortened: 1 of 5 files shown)


Kokan:

lib/apphook.c
```diff
@@ -95,15 +95,21 @@ register_application_hook(gint type, ApplicationHookFunc func, gpointer user_dat
 }
 
 static void
-run_application_hook(gint type, gboolean remove_hook)
+_update_current_state(gint type)
+{
+  if (AH_REOPEN == type)
```
furiel: Why do we need this early return?

Kokan: Because the `AH_REOPEN` is kinda hack here. The app hook has its internal state machine, stepping from daemonize, pre config, post config, and so on. This function updates that.
A real deal would be to untangle the reopen from the rest of the hooks.

furiel: Oh ok this is indeed an equivalent transformation, just a tricky one: only AH_REOPEN had remove hook false.

lib/apphook.c
```diff
@@ -98,21 +98,18 @@ static void
 _update_current_state(gint type)
 {
   if (AH_REOPEN == type)
-     return;
```
lbudai: ahh.. so you did not run the style-format for the first patch :)

Kokan: :(

PR #2296, opened 2018-09-20
riemann crashes in flush

Kokan:
I am just stealing a patch from #2291 which fixes a riemann issue caused by #2098 so it could go in faster.

### DIFF
--- a/modules/riemann/riemann-grammar.ym
+++ b/modules/riemann/riemann-grammar.ym
@@ -138,7 +138,7 @@ riemann_option
           }
         | KW_FLUSH_LINES '(' nonnegative_integer ')'
           {
-            riemann_dd_set_flush_lines(last_driver,  $3);
+            log_threaded_dest_driver_set_flush_lines(last_driver,  $3);
           }
         | threaded_dest_driver_option
         | KW_TLS '(' riemann_tls_options ')'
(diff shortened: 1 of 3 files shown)


bazsi:
It's good for me, obviously I am the original author, but I think we can warrant one approver in your name, as you seem to agree with the patch :)

Kokan:
@bazsi I just need to check one thing pointed out by @furiel

bazsi:
I think right now <=1 means no batching.

furiel:
@bazsi @Kokan unfortunately there is still an invalid memory access. The problem comes when flush-lines is not specified:
```
destination d_riemann {
    riemann(description($MESSAGE)
    port(5555)
#    flush-lines(0)
    server("localhost"));
};
```

In that case, after the MAX calculation, there will be only room for one event after the malloc. Though batching is disabled now, and in that case logthrdestrv will send every message she can at the read loop to riemann, that means more than one events will be appended to a fix size array.

You can reliably reproduce with an stdin source:
```(echo "line1\nline2\nline3\n"; sleep 5) | valgrind ./syslog-ng -Fe -f ../etc/riemann.conf```

Now `riemann_worker_insert` can only return with QUEUED after inserting a single message, meaning riemann does not support non-batching mode now. Maybe this is where the problem comes from.

bazsi:
Apart from a small change to use the log_threaded_dest_driver_flush() flush wrapper, this still looks good to me.

kira-syslogng:
Build SUCCESS

PR #2294, opened 2018-09-20
Small stats fixes

lbudai:

PR #2297, opened 2018-09-20
[wip]Stats autoreset ctr

lbudai:
@bazsi: could you check it if works?
If yes, then we could merge the first part of this PR (without the single-stats), then later, in another PR I'll change the single-stats-cluster.

Note that it is rebased to #2294

(Stats is a little bit complicated.)

### DIFF
--- a/lib/logqueue.c
+++ b/lib/logqueue.c
@@ -178,8 +178,8 @@ _register_queued_messages_counter(LogQueue *self, gint stats_level, const StatsC
   else
     {
       stats_register_counter(stats_level, sc_key, SC_TYPE_QUEUED, &self->queued_messages);
-      stats_counter_set(self->queued_messages, log_queue_get_length(self));
     }
+  stats_counter_add(self->queued_messages, log_queue_get_length(self));
 }
 
 static void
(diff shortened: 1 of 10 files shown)


lbudai:
@kira-syslogng : Test this please

lbudai:

bazsi:
This does not add counters but set them, or.am I missing something?

lbudai:
@bazsi: pushed

bazsi:
Thanks, will check it out

On Sep 20, 2018 20:10, "Budai Laszlo" <notifications@github.com> wrote:

@bazsi <https://github.com/bazsi>: pushed

—
<https://github.com/balabit/syslog-ng/pull/2297#issuecomment-423280620>,
or mute
the thread

lbudai:

bazsi:
It still has some issues, but I think is in the right direction.

1) the logqueue initial allocation is strange, the LogQueue interface has disk queue specific fields, that are always zero for memory based queues, thus their memory allocation is lost. This is probably the reason for the kira failure as well. With a simple reload, the memory_usage counter remains zero, even though the addition is there. The "queued" value seems to work.

2) the queues may need to save their original memory usage in an instance variable to be able to be able restore it, should they be still used after reload.

3) we might want to get rid off "use" function, why don't we simply register the counter again with the same key? As far as I see it's basically the same. We could get rid of the "is_registered" style conditionals in the registration functions and most of the bodies of _use and _track is the same anyway.

4) _register_memory_usage_counter seems to ignore its stats_level argument and uses a fixed value instead.

That's it for now.

lbudai:
@bazsi: I’m not sure this is the right way.
If a memory usage counter contains a value, and the counter is not reset during reload, why do we want to save values and then accumulate them again? The stats counters are located in a hashtable(to be precise, the statscluster are stored in this table) which is like persist-config: values are accessible after a reload.

bazsi:
Because we might have multiple queues and the number of queues might change
accross reloads, registered to the same counters.

lbudai:
@bazsi: indeed... then I'll change the code accordingly

pzoleex:

pzoleex:

pzoleex:

pzoleex:

pzoleex:

kira-syslogng:
Build FAILURE

PR #2301, opened 2018-09-21
Fixing some compiler warnings from armv7l

furiel:
I was compiling syslog-ng on armv7l and I found some warnings, that I intend to fix now. (Probably glong->gint64 change is not arm specific.)

I wanted to add `-Wincompatible-pointer-types` to extra warnings, but 4.8 gcc and clang does not know that warning yet. Though with xenial gcc, I could compile without these warnings.

Besides these, there were  43 instances `warning: cast increases required alignment of target type [-Wcast-align]` that I do not fix for now.


kira-syslogng:
Build SUCCESS

PR #2302, opened 2018-09-21
[wip] Stats prepare for multiple queues per destdrv

lbudai:


lbudai:
@bazsi: I think this is what we've talked about

* track the counters internally, when one of the counters changed, change the 'cached' value
* counter reg: increase with the cached value
* counter unreg: decrease with the cached value

( unit tests are missing, will added after you tried it)

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lbudai:
@bazsi: I checked again why tests has failed, and I think in case of diskq we cannot rely on a cached queued_messages value _after_ the reload.
We store the `LogQueue` instance in the persist-config, this is why it work for in-memory `LogQueue`.
In case of diskq when we found a `LogQueue` instance in the persist config, the very first thing what we do is to unref it... so the cached values are dropped. This is why the tests in Kira failed.
In case of `queued_messages` we have to use the `log_queue_get_length` but instead of *set* we should *add* it to the current value... 
The cached queued_messages is still need, unless we want to implement the autoreset...

I hope that it solve all the issues.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:
I have now rebased my patches against this series, and code-wise it seems to have worked ok. I am yet to do functional testing.

At the end I've pushed the DROPPED counter down into log_queue_register_stats_counters(), the
dropped counter felt completely out of place, and we'd be relying on being able to register the same
counter multiple times anyway.

This way both the driver and the associated LogQueue would grab the same counter, the only
risk is that the StatsClusterKey/stats_level would somehow be different for the two registrations. However both of these values are actually passed in as arguments.

This is the patch I am referring to: https://github.com/balabit/syslog-ng/pull/2291/commits/be8a4bb5e71ef5e32372ddeeddc0b976b829bab7

lbudai:
@bazsi:
I'm ok with this change.

Should I add this to my PR? (if it causes conflict, I'd not touch it...)

bazsi:
I have pushed that change to the bottom of my branch, so please do cherry-pick it over. I still have some rebasing work to do.

This is the one: 551dc838293146345f2c6cdf92fdd20e05f2221f

kira-syslogng:
Build SUCCESS

PR #2314, opened 2018-09-27
syslog-parser: fix unexpected flag check-hostname

Kokan:
The following configuration was not accepted by:
```
parser  { syslog-parser(flags(check-hostname)); };
```

It was not accepted because the lexer has `KW_CHECK_HOSTNAME`, and the logic if something is a keyword or a string is looking up the keyword list. But the flags option expect it to be a string instead of a keyword, therefore the grammer fails.

### DIFF
--- a/modules/syslogformat/syslog-parser-grammar.ym
+++ b/modules/syslogformat/syslog-parser-grammar.ym
@@ -81,8 +81,9 @@ parser_syslog_opt
         ;
 
 parser_syslog_opt_flags
-        : string parser_syslog_opt_flags     { CHECK_ERROR(msg_format_options_process_flag(&((SyslogParser *) last_parser)->parse_options, $1), @1, "Unknown flag %s", $1); free($1); }
-	|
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
So in the end if some string exist as keyword too, grammar takes precedence to it. If I understand correctly, if we delete KW_CHECK_HOSTNAME (I know we can't) then this should start working. Also with flags("check-hostname"), we can force the string path.

gaborznagy:
I see the exact same solution in cfg-grammar.y:
```
source_reader_option_flags
      : string source_reader_option_flags     { CHECK_ERROR(log_reader_options_process_flag(last_reader_options, $1), @1, "Unknown flag %s", $1); free($1); }
      | KW_CHECK_HOSTNAME source_reader_option_flags     { log_reader_options_process_flag(last_reader_options, "check-hostname"); }
```
I wish there could be an alternative for this, as it is hard to maintain if we make this conflict with a new option, or using it somewhere else.

Kokan:
@furiel yes, as you stated. in that case it should work.
@gaborznagy I cheated a little about the fix :)

A possible a better resolution would be to move all of the flag parsing to the grammar/lexer level instead of the `cfg_process_flag` function.

PR #2318, opened 2018-09-28
DBLD: Support Ubuntu Bionic and update existing images

mitzkia:
Updated:
see my last comment: https://github.com/balabit/syslog-ng/pull/2318#issuecomment-427330823
Mainly the PR is about to add Ubuntu Bionic as new supported OS.
The PR is also added some fixes to other parts of the DBLD

Other changes that this PR is introduce:
- added new type of packages file: This file will contains all of the packges for all platforms. There are 
too many package files (apt, pip - for each platform and for all), and it is not so easy to maintain them.
- make target: add support to run make for custom images
- make-check target: add support to run make-check

How to test:
$ dbld/rules image-bionic 
$ dbld/rules bootstrap-bionic 
$ dbld/rules make-bionic 
$ dbld/rules make-check-bionic

Todo:
I also wanted to update old images to the newer form.
Check there is no failures with the new images.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

dbld/images/xenial.dockerfile
```diff
@@ -1,41 +1,40 @@
...
-ARG OBS_REPO=xUbuntu_16.04
+LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>, "
 
-LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>"
+ENV DEBIAN_FRONTEND=noninteractive
+ENV DEBCONF_NONINTERACTIVE_SEEN=true
+ENV LANG C.UTF-8
+ENV OS_PLATFORM xenial
 
-COPY helpers/* /helpers/
```
szemere: Could you explain the benefits of copying the "helper" files one-by-one?

szemere:
Could you please explain why did you splitted up the functions.sh?

mitzkia:
Thanks for your notice. It was just a first try way. Since than I have finished package list for all platforms, and it seems it wont be a good way. I already have a second way POC for package installs. Soon I will update the PR.

dbld/images/bionic.dockerfile
```diff
@@ -0,0 +1,40 @@
...
+ENV DEBIAN_FRONTEND=noninteractive
+ENV DEBCONF_NONINTERACTIVE_SEEN=true
+ENV LANG C.UTF-8
+ENV OS_PLATFORM bionic
+
+RUN apt-get update -qq -o Acquire::CompressionTypes::Order::=gz
+
+COPY required-apt/packages.txt /packages.txt
+RUN grep ${OS_PLATFORM} /packages.txt | cut -d"[" -f1 | xargs apt-get install --no-install-recommends --yes
+
```
bazsi: this grep and cut magic should be a helper script that invokes apt based on the packages file.

mitzkia: Ok, agreed. In the next update I will fix this.

dbld/images/bionic.dockerfile
```diff
@@ -0,0 +1,40 @@
...
+COPY required-pip/requirements.txt /requirements.txt
+RUN python -m pip install --upgrade pip && \
```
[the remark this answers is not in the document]
mitzkia: Ok, agreed. In the next update I will fix this.

dbld/images/bionic.dockerfile
szemere: During the development of dbld, I have intentionally tried to avoid extracting this command from the Dockerfile. IMHO "hiding" the install steps somehow mystifies the actual work. (Using text files was already a big step back.)

i.e.: Since the Dockerfile's syntax is limited, the easiest Dockerfile will be:
`FROM imagename`
`COPY bar /foo`
`RUN foo/create.sh`
`ENTRYPOINT["foo/foo.sh"]`

Obviously, we don't want to do that. However, IF we take this step, I think we can use some other formats which will furtherly reduce the maintenance complexity of the package lists, and provide extra functionality at the same time. i.e.:
- warning about duplications (some kind of stylecheck)
- creating "groups" of platforms, to sorten the lists.
- able to provide version numbers (the same method can be used for pip)

mitzkia:
I have updated this PR. I have reduced the scope only for the Ubuntu Bionic support. 
From my point of view this PR is depends on #2329 
If this PR is accepted I would like to update other dbld images for using methods introduced in this PR e.g:
- using functions from dependencies.sh
- using packages from packages.txt

kira-syslogng:
Build SUCCESS

mitzkia:
Or maybe if I updates also the xenial image in this PR, it will automatically tested by Kira.

kira-syslogng:
Build FAILURE

mitzkia:
This PR is not depends on #2329 anymore.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

dbld/Makefile.am
```diff
@@ -9,23 +9,27 @@ EXTRA_DIST += \
...
+	dbld/images/centos_6.dockerfile \
+	dbld/images/centos_7.dockerfile \
+	dbld/images/debian_8_jessie.dockerfile \
+	dbld/images/debian_9_stretch.dockerfile \
+	dbld/images/debian_10_buster.dockerfile \
+	dbld/images/ubuntu_1404_trusty.dockerfile \
+	dbld/images/ubuntu_1604_xenial.dockerfile \
+	dbld/images/ubuntu_1804_bionic.dockerfile \
+	dbld/images/ubuntu_1810_cosmic.dockerfile \
 	dbld/make \
```
bazsi: I would either use the code name (e.g. xenial) or the version number in the names of files but not both. If you ask me, I'd prefer names, but YMMV.

dbld/images/helpers/dependencies.sh
```diff
@@ -0,0 +1,125 @@
...
+    download_target "https://github.com/maxmind/libmaxminddb/releases/download/${LIBMAXMINDDB_VERSION}/libmaxminddb-${LIBMAXMINDDB_VERSION}.tar.gz" /tmp/libmaxminddb.tar.gz
+    cd /tmp/
+    tar xvf /tmp/libmaxminddb.tar.gz
+    cd /tmp/libmaxminddb-${LIBMAXMINDDB_VERSION}
+    ./configure
+    make
+    make install
+    ldconfig
+    rm -rf /tmp/libmaxminddb.tar.gz /tmp/libmaxminddb-${LIBMAXMINDDB_VERSION}
+}
```
bazsi: runtime dependencies should not be installed from source, unless strictly necessary. libmaxminddb is available in ubuntu repositories, so we should just use that. Ah.. I see that you were only doing this on platforms where this is not available.

For non-critical stuff, I'd simply not compile them on the platform, for critical stuff (I am not sure there's any), this could be an option, but then we would probably have to supply the binaries bundled.

For dev tools, it is ok to pull in 3rd party tools, but again I'd try to avoid that if possible.

dbld/images/helpers/dependencies.sh
```diff
@@ -0,0 +1,125 @@
...
+    cd /tmp/
+    tar xvf /tmp/mongo-c-driver.tar.gz
+    cd /tmp/mongo-c-driver-${MONGOC_VERSION}
+    mkdir cmake-build/
+    cd cmake-build/
+    cmake -DENABLE_AUTOMATIC_INIT_AND_CLEANUP=OFF ..
+    make
+    make install
+    rm -rf /tmp/mongo-c-driver.tar.gz /tmp/mongo-c-driver-${MONGOC_VERSION}
+}
```
bazsi: again runtime lib, should probably be skipped on platforms that miss this.

dbld/images/helpers/dependencies.sh
```diff
@@ -0,0 +1,125 @@
...
+    cd /tmp/
+    tar xfz /tmp/rabbitmq-c.tar.gz
+    cd /tmp/rabbitmq-c-${RABBITMQC_VERSION}/
+    mkdir cmake-build/
+    cd cmake-build/
+    cmake ..
+    make
+    make install
+    rm -rf /tmp/rabbitmq-c.tar.gz /tmp/rabbitmq-c-${RABBITMQC_VERSION}/
+}
```
bazsi: same.

dbld/images/helpers/dependencies.sh
```diff
@@ -0,0 +1,125 @@
...
+    download_target "https://github.com/algernon/riemann-c-client/archive/riemann-c-client-${RIEMANN_VERSION}.tar.gz" /tmp/riemann.tar.gz
+    cd /tmp/
+    tar xfz /tmp/riemann.tar.gz
+    cd /tmp/riemann-c-client-riemann-c-client-${RIEMANN_VERSION}
+    autoreconf -i
+    ./configure
+    make
+    make install
+    rm -rf /tmp/riemann.tar.gz /tmp/riemann-c-client-riemann-c-client-${RIEMANN_VERSION}
+}
```
bazsi: same.

dbld/images/packages/packages.txt
```diff
@@ -0,0 +1,171 @@
+#autogen
```
bazsi: since this has become a simple file now, I'd omit the directory and move it to the images/ subfolder.

dbld/images/packages/packages.txt
```diff
@@ -0,0 +1,171 @@
...
+[
+    autoconf                    [centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+    autoconf268                 [centos6]
```
bazsi: the names of distros/version numbers should be unified, I'd go for: <distro>-<version>, preferring code-names if those are widely used (e.g. ubuntu).

e.g. centos-6, centos-7, debian-stretch, debian-buster, ubuntu-xenial, ubuntu-trusty.

dbld/images/packages/packages.txt
```diff
@@ -0,0 +1,171 @@
...
+]
+autoconf-archive                [centos6, centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+automake                        [centos6, centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+ca-certificates                 [centos6, centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+git                             [centos6, centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+libtool                         [centos6, centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+[
```
bazsi: could we add support for "*" element here? it could become cumbersome to maintain this list, as any new release would need a new entry to all of the lines.

dbld/images/packages/packages.txt
```diff
@@ -0,0 +1,171 @@
...
+    pkg-config                  [d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+    pkgconfig                   [centos6, centos7]
+]
+
```
bazsi: I like the ability to group things this way.

dbld/images/packages/requirements.txt
```diff
@@ -0,0 +1,6 @@
+#make_check
+hy                      [centos6, centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+nose                    [centos6, centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+pep8                    [centos6, centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+ply                     [centos6, centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
+pylint                  [centos7, d_jessie, d_stretch, d_buster, u_trusty, u_xenial, u_bionic, u_cosmic]
```
bazsi: in this way, this is not in pip format anymore, so I would at the very least use a different name for the file then.

dbld/images/ubuntu_1404_trusty.dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM ubuntu:14.04
+LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>, Balazs Scheidler <balazs.scheidler@oneidentity.com>"
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV DEBCONF_NONINTERACTIVE_SEEN=true
+ENV LANG C.UTF-8
+ENV OS_PLATFORM trusty
+
```
bazsi: why does it not have to match the distro names in the packages.txt file?

Ah, I think I get it now, it does not need to match as long as "trusty" as a string is present. I think this is not a good idea, and we should use an exact match.

also note that OS_PLATFORM as an environment variable is exported to dependencies.sh, so no need to pass it as an argument, it should be available as an environment variable, making the dockerfile a bit more concise.

dbld/rules
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/make -f
 
-IMAGES=centos6 centos7 jessie artful xenial trusty devshell
+IMAGES=centos_6 centos_7 jessie artful xenial trusty devshell bionic cosmic buster
 DEFAULT_IMAGE ?= xenial
```
bazsi: again, let's use debian-jessie here and match them up with tokens in packages.txt

this should also be inherited in docker image names, I don't think that would be an issue, but let me know if you think it is.

dbld/rules
```diff
@@ -104,7 +104,7 @@ shell-%: setup
 images: $(foreach image,$(IMAGES), image-$(image))
 image: image-$(DEFAULT_IMAGE)
 image-%:
-	cd $(DBLD_DIR)/images/ && $(DOCKER) build --network=host -t balabit/syslog-ng-$* -f $*.dockerfile .
+	cd $(DBLD_DIR)/images/ && $(DOCKER) build --network=host -t balabit/syslog-ng-$* -f $$(ls *dockerfile | grep $*) .
```
bazsi: once we match image names with files, this should not be needed.

dbld/images/helpers/dependencies.sh
```diff
@@ -0,0 +1,51 @@
...
+function download_target() {
+    target=$1
+    output=$2
+    wget --no-check-certificate $target --output-document=$output
+}
+
+function install_apt_packages {
+    OS_PLATFORM=$1
+    apt-get update -qq -o Acquire::CompressionTypes::Order::=gz
+    grep ${OS_PLATFORM} /packages/packages.txt | cut -d"[" -f1 | xargs apt-get install --no-install-recommends --yes
```
szemere: optional: to avoid unintentional matches I would filter out comments before matching to OS_PLATFORM.
optional2: add the surrounding '[' brackets to the expression to make sure no package name will conflict with any platform.

mitzkia:
After an IRL discussion of review comments I have updated this PR.
Hope commit messages will help in understanding.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

dbld/images/xenial.dockerfile
bazsi: This has been resolved since.

dbld/images/helpers/dependencies.sh
bazsi: We agreed on this and the latest version dropped the private compilations during build, so this is resolved.

dbld/images/packages/packages.txt
bazsi: we've discussed this and came to the conclusion that we can use the name of the distro (e.g. ubuntu or centos) without version number to specify the name for a specific distro as a whole.

this is a simple form of wildcarding, and more complex would complicate the file parsing a lot (which is very simple/dumb now).

This should work for now.

bazsi:
The reason this fails in kira is that it is trying to use the old image name (image-xenial instead of image-ubuntu-xenial). We could perhaps add a compatibility target to dbld/rules like this:

image-xenial: image-ubuntu-xenial

This still does not resolve the problem as kira would be still be using syslog-ng-xenial as the docker image name. To resolve that we would either need a flag day and manually push the transition over (e.g. rebuild dockerhub images with new names, merge the PR, and change kira as well), or make sure that we have compatibility names for Makefile targets as well as docker image names.

dbld/images/devshell.dockerfile
```diff
@@ -1,17 +1,14 @@
...
 
-LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>"
+ENV DEBIAN_FRONTEND=noninteractive
+ENV DEBCONF_NONINTERACTIVE_SEEN=true
+ENV LANG C.UTF-8
+ENV OS_PLATFORM devshell
 
 COPY helpers/* /helpers/
 
 RUN /helpers/functions.sh enable_dbgsyms
```
szemere: functions.sh -> dependencies.sh

dbld/images/helpers/dependencies.sh
```diff
@@ -0,0 +1,145 @@
...
+    fi
+}
+
+function enable_dbgsyms {
+    echo "deb http://ddebs.ubuntu.com $(lsb_release -cs) main restricted universe multiverse
+deb http://ddebs.ubuntu.com $(lsb_release -cs)-updates main restricted universe multiverse
+deb http://ddebs.ubuntu.com $(lsb_release -cs)-proposed main restricted universe multiverse" | \
+    tee -a /etc/apt/sources.list.d/ddebs.list
+    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 428D7C01 C8CAB6595FDFF622
+    apt-get update
```
szemere: syncronise it with `/helpers/dependencies.sh install_apt_packages` (apt-get update)

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

pzoleex:
internal mi do test

kira-syslogng:
Build FAILURE

pzoleex:
internal mi do test

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:

kira-syslogng:
Build FAILURE

bazsi:

kira-syslogng:
Build SUCCESS

bazsi:
I have installed make on the slaves and now it ran properly

PR #2319, opened 2018-09-28
python block skip string check in comments

### LINKED ISSUE #2317: inline Python code: ' in comment generates syntax error
# syslog-ng

## Version of syslog-ng

git snapshot from yesterday

## Platform

CentOS 7

# Issue

## Failure

While testing the Python parser.

If there is a ' in a comment, like in:
```
# Don't do something
```
Then syslog-ng does not start, but gives an error message and the ^^^ is at the wrong place:
```
[2018-09-28T15:13:52.655555] Module loaded and initialized successfully; module='mod-java'
[2018-09-28T15:13:52.655701] Finishing include; content='block destination elasticsearch2() at /usr/share/syslog-ng/include/scl/elasticsearch/plugin.conf:59', depth='2'
[2018-09-28T15:13:52.671001] Module loaded and initialized successfully; module='mod-python'
Error parsing Python code, syntax error, unexpected $end, expecting LL_BLOCK in /etc/syslog-ng/conf.d/suricata.conf:
56      """
57      
58      import socket
59      
60      class SngResolver(object):
61---->     def parse(self, log_message):
61----> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
62              """
63              Resolves IP to hostname
64              """
65      
66              # Don't resolve DNS transactions

Included from /etc/syslog-ng/syslog-ng.conf:
63      log { source(s_sys); filter(f_cron); destination(d_cron); };
64      
65      
66      # Source additional configuration files (.conf extension only)
67      @include "/etc/syslog-ng/conf.d/*.conf"
68----> 
68----> ^
69      
70      # vim:ft=syslog-ng:ai:si:ts=4:sw=4:et:
```

## Configuration

```
[...]
parser p_resolver {
    python(
        class("SngResolver")
    );
};

python {

"""
simple syslog-ng Python parser example
resolves IP to hostname
"""

import socket

class SngResolver(object):
    def parse(self, log_message):
        """
        Resolves IP to hostname
        """

        # Don't resolve DNS transactions
        dest_port = log_message['suricata.dest_port']
        if dest_port == '53':
            return True
[...]
```
If I use """ for comment instead of #, the ' does not cause a problem any more. The same Python code runs just fine when used in a .py file instead of inline, and pylint does not complain about it either.

Kokan:
Fixes #2317 (details in the issue).

Also this fixes an other case
```
python {
#}
};
```

### DIFF
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -127,6 +127,23 @@ yy_input_run_backtick_substitution(CfgLexer *self, gchar *buf, gsize buf_size)
     longjmp(__yyg->yyextra_r->fatal_error, 1);				\
   } while(0)
 
+#define _process_commented_line(scanner,content,cond) \
+  do {                                                \
+    int ch;                                           \
+                                                      \
+    ch = input(scanner);                              \
+    while (ch != '\n' && ch != EOF && ch != 0)        \
+      {                                               \
+        if (cond)                                     \
+          g_string_append_c(content, ch);             \
+        ch = input(scanner);                          \
+      }                                               \
+    if (ch == '\n')                                   \
+      {                                               \
+        unput(ch);                                    \
+      }                                               \
+  } while (0)
+
 
 static void
 _cfg_lex_move_token_location_to_new_line(CfgLexer *lexer)
@@ -197,19 +214,7 @@ word	[^ \#'"\(\)\{\}\[\]\\;\r\n\t,|\.@:]
 %%
 
 \#                         {
-                             int ch;
-
-                             ch = input(yyscanner);
-                             while (ch != '\n' && ch != EOF && ch != 0)
-                               {
-                                 if (yyextra->token_text)
-                                   g_string_append_c(yyextra->token_text, ch);
-                                 ch = input(yyscanner);
-                               }
-                             if (ch == '\n')
-                               {
-                                 unput(ch);
-                               }
+                             _process_commented_line(yyscanner,yyextra->token_text,yyextra->token_text);
                            }
 ^@                         {
                              return LL_PRAGMA;
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -308,6 +308,11 @@ word	[^ \#'"\(\)\{\}\[\]\\;\r\n\t,|\.@:]
                              yy_push_state(block_qstring, yyscanner);
                            }
 
+<block_content>\#          {
+                             g_string_append_c(yyextra->string_buffer, yytext[0]);
+                             _process_commented_line(yyscanner, yyextra->string_buffer, TRUE);
+                           }
+
 <block_string>[^"]+        { g_string_append(yyextra->string_buffer, yytext); }
 <block_string>\\\"         { g_string_append(yyextra->string_buffer, yytext); }
 <block_string>\"           {


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/cfg-lex.l
```diff
@@ -127,6 +127,23 @@ yy_input_run_backtick_substitution(CfgLexer *self, gchar *buf, gsize buf_size)
     longjmp(__yyg->yyextra_r->fatal_error, 1);				\
   } while(0)
 
+#define _read_until_eof_or_eol(scanner,content,cond)  \
+  do {                                                \
```
bazsi: I think this could become a function instead of a macro. The others are overriding lexer functionality, but this should just be a function. It does not even change the control flow.

bazsi: I would also change the verb "read" to something like, I don't have a firm opinion, the best I could come up with is  _process_commented_line(), or instead of process, it could be scan/parse/skip, however reading as a term does not apply.

This is probably nitpicking, but if you change this to a function anyway, it'd be easy to rename it as well.

bazsi:
Apart from the minor comments I have made, this looks good to me, even as it is.

lib/cfg-lex.l
Kokan: I had trouble with `input` / `unput`, when it was a function; otherwise I would it to be a function.
Sure renaming make sense.

kira-syslogng:
Build SUCCESS

PR #2320, opened 2018-09-30
Logthreadeddst batching

shivamk01:
We have to add hashes values using MD5 here to basically generate logs for each system value so that we can check for malicious files online like on tolls like virus total.


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

PR #2329, opened 2018-10-03
DBLD: update make target, support make-check target

mitzkia:
- Added some minor changes to `dbld/rules make`
- Support to run `dbld/rules make-check`
- Goal of this changes is to run following commands from CI easily
$ dbld/rules make-[image] 
$ dbld/rules make-check-[image] 

### DIFF
--- a/dbld/make
+++ b/dbld/make
@@ -4,3 +4,4 @@ set -e
 
 cd /build
 make "$@"
+make install
--- a/dbld/rules
+++ b/dbld/rules
@@ -66,6 +66,10 @@ make: setup-$(DEFAULT_IMAGE)
 make-%: setup
 	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -ti  balabit/syslog-ng-$* /source/dbld/make $(MAKE_ARGS)
 
+make-check: setup-$(DEFAULT_IMAGE)
+make-check-%: setup
+	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -ti  balabit/syslog-ng-$* /source/dbld/make-check
+
 tarball: tarball-$(DEFAULT_IMAGE)
 tarball-%: setup
 	if [ -f $(TARBALL) ]; then \
--- a/dbld/rules
+++ b/dbld/rules
@@ -62,8 +62,9 @@ bootstrap: bootstrap-$(DEFAULT_IMAGE)
 bootstrap-%: setup
 	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -e CONFIGURE_OPTS="$${CONFIGURE_OPTS:-$(CONFIGURE_OPTS)}" -ti  balabit/syslog-ng-$* /source/dbld/bootstrap
 
+make: setup-$(DEFAULT_IMAGE)
 make-%: setup
-	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -ti  balabit/syslog-ng-$(DEFAULT_IMAGE) /source/dbld/make $(MAKE_ARGS) $*
+	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -ti  balabit/syslog-ng-$* /source/dbld/make $(MAKE_ARGS)
 
 tarball: tarball-$(DEFAULT_IMAGE)
 tarball-%: setup
--- a/dbld/make-check
+++ b/dbld/make-check
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+cd /build
+make check ; ret=$? ; if [ $ret -ne 0 ]; then cat test-suite.log ; exit $ret ; fi
+make func-test ; ret=$? ; if [ $ret -ne 0 ]; then cat test-suite.log ; exit $ret ; fi


kira-syslogng:
Build FAILURE

mitzkia:

kira-syslogng:
Build FAILURE

mitzkia:

kira-syslogng:
Build SUCCESS

dbld/make
```diff
@@ -4,3 +4,4 @@ set -e
 
 cd /build
 make "$@"
+make install
```
bazsi: hmm, why are we adding an explicit "install" after we already run what the user requested?

e.g. ./dbld/rules make-check  would execute this with "check", thus the make shell script would be executed with "check", and with the above change we would be installing as well. Is there a reason?

Thanks

dbld/rules
```diff
@@ -62,8 +62,13 @@ bootstrap: bootstrap-$(DEFAULT_IMAGE)
 bootstrap-%: setup
 	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -e CONFIGURE_OPTS="$${CONFIGURE_OPTS:-$(CONFIGURE_OPTS)}" -ti  balabit/syslog-ng-$* /source/dbld/bootstrap
 
+make: setup-$(DEFAULT_IMAGE)
 make-%: setup
-	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -ti  balabit/syslog-ng-$(DEFAULT_IMAGE) /source/dbld/make $(MAKE_ARGS) $*
+	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -ti  balabit/syslog-ng-$* /source/dbld/make $(MAKE_ARGS)
+
```
bazsi: hm. with this change I would be unable to run make-check and make-install. I am not against changing it, but I don't see how this functionality would be available otherwise?

Previously make only supported specifying the image by supplying DEFAULT_IMAGE on the make command line, e.g.

./dbld/rules make-check DEFAULT_IMAGE=xenial

The goals behind the targets were:
* image related targets expected the image name as a part of the target (e.g. image-xenial or image-centos6)
* development related targets (bootstrap, release, etc) used the DEFAULT_IMAGE variable, as it is expected that an interactive build would always happen in the same environment, that can even be specified in ./dbld/rules.conf
* to still override the image, one could use the DEFAULT_IMAGE override, e.g.

```
$ ./dbld/rules bootstrap DEFAULT_IMAGE=xenial
$ ./dbld/rules make-check DEFAULT_IMAGE=xenial
```

In hindsight, I am not entirely sure the Makefile format was such a good idea, especially as it was using shell scripts for more complicated problems anyway, so that it could run them easily under docker.

dbld/rules
```diff
@@ -62,8 +62,13 @@ bootstrap: bootstrap-$(DEFAULT_IMAGE)
...
+make-check: setup-$(DEFAULT_IMAGE)
+make-check-%: setup
+	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -ti  balabit/syslog-ng-$* /source/dbld/make-check
```
bazsi: as I wrote previously, I don't think we need a make-check target, we already had that even without this patch.

dbld/make
mitzkia: Thanks for the note and the example. 
Before the patch the "$*" arg (from rules->make) was used  to pass the "install" into the make script. But in this patch I wanted to use the "$*" for the defination of the image/platform (see my last comment).

dbld/rules
mitzkia: Thanks for the notes about the image related targets and developement related targets.
I am answering in a last comment.

mitzkia:
@bazsi , thanks for the notes and clarifications.

Before the patch a complete build pipeline would look like the following. 
I have intentionally did not specify an image (platform), because I belive later they will be "identical".
1, dbld/rules image-[MY-PLATFORM]
2, dbld/rules bootstrap-[MY-PLATFORM] (autogen.sh, configure)
3, export DEFAULT_IMAGE=[MY-PLATFORM]
4, dbld/rules make-install (make, make install)
5, dbld/rules make-check (make-check)
6, dbld/rules make-func-test (make-func-test)
7, dbld/rules deb-[MY-PLATFORM]

After the patch it will look like:
1, dbld/rules image-[MY-PLATFORM]
2, dbld/rules bootstrap-[MY-PLATFORM] (autogen.sh, configure)
3, dbld/rules make-[MY-PLATFORM] (make, make install)
4, dbld/rules make-check-[MY-PLATFORM] (make-check, make-func-test)
5, dbld/rules deb-[MY-PLATFORM]

The main difference is for me, that in the second way I should not specify my platform as a DEFAULT_IMAGE.
To specify my platform as a DEFAULT_IMAGE is not self-evident after the 1 and 2 steps.
The goal is that we wanted to use these steps/targets in a CI job in a most handy way for different platforms.

furiel:
> 3, dbld/rules make-[MY-PLATFORM] (make, make install)
> 4, dbld/rules make-check-[MY-PLATFORM] (make-check, make-func-test)

Can you run make check before make install? Make check should not depend install. Otherwise we can shadow some exciting linking bugs on the unit tests. 

About func-tests, in my opinion it can stay after install if you like so, but it is also prepared to be able to executed from the build directory.

mitzkia:
@furiel thanks for the note.
Yes, the order of the commands can be different (it was just an example order): make-check not depend on make install.
The goal is that: dbld/rules [some target]-[MY-PLATFORM] would be good to exist (symmetricity).

kira-syslogng:
Build SUCCESS

bazsi:
Please dont forget that I am using this interactively, and in that case I
dont want to specify the target platform. And I am invoking these commands
a lot, which I do manually, in contrast with Jenkins which only need
setting up once.

Why dont you introduce separate targets and leave these low level ones
alone?

Like:

* ci-build-%
* ci-test-%
* ci-package-%

It is even possible to do a special "ci-build-all" target that would
iterate all platforms supported.

mitzkia:
@bazsi Thanks for the comment and reminding you run these commands regularly.
In that case new targets can be work. I will update this PR in this way.
About the "ci-build-all" I think, this functionality belongs to the CI. I would make the build-test-package jobs run parallely for different platforms.

mitzkia:
I am chaning my mind. Finally I am agreed with @bazsi . The functionality is already here.
I did not want to add more new targets which will be only redundant. 
It is fine to setup CI job with already existing targets.
So I am closing this PR.

PR #2333, opened 2018-10-03
Message about not supported cap_syslog only at debug level

Kokan:
If the syslog-ng was running on a system where `CAP_SYSLOG` is not present, it would print to the `stderr` unconditionally a message about fallbacking to `CAP_SYS_ADMIN`.
This prevent that log by default, and moves it into the debug level.

### DIFF
--- a/lib/gprocess.c
+++ b/lib/gprocess.c
@@ -122,6 +122,7 @@ static struct
   const gchar *pidfile_dir;
   const gchar *cwd;
   const gchar *caps;
+  gboolean enable_caps;
   gint  argc;
   gchar **argv;
   gchar *argv_start;
@@ -142,7 +143,9 @@ static struct
   .check_period = -1,
   .check_fn = NULL,
   .uid = -1,
-  .gid = -1
+  .gid = -1,
+  .enable_caps = TRUE,
+  .caps = NULL
 };
 
 #if SYSLOG_NG_ENABLE_SYSTEMD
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build FAILURE

Kokan:

lib/gprocess.c
```diff
@@ -384,6 +393,12 @@ g_process_check_cap_syslog(void)
     }
 }
 
+gboolean
+g_process_is_cap_enabled(void)
+{
+  return process_opts.enable_caps;
+}
```
bazsi: this seems like a bit redundant to me, I mean we have a boolean parameter in addition caps being NULL or not NULL. I could accept the redundancy, however as it seems the --no-caps argument manipulates caps and not this variable, thus we'd still try to use caps.

I am a bit tired now and may be missing something though.

Kokan: The `--no-caps` manipulates `enable_caps` to `FALSE`.  The `caps` & `!caps` won't fly anymore, because actually we need 3 state:
* default caps
* caps from user
* no caps

Before this patch, this 3 state was solved by initializing with `default caps` and when user overwrite it everything was as expected.
But because I need to have the `debug_flag` before setting its default value, that trick won't help anymore.

lbudai: @Kokan : add these information to the commit message (where `enable_caps` is introduced)

Kokan: Sure.

kira-syslogng:
Build SUCCESS

lib/gprocess.c
lbudai: @Kokan : thanks!!

PR #2337, opened 2018-10-04
python: include python2/3 in plugin description

### LINKED ISSUE #2336: print Python version with syslog-ng -V
# Issue

syslog-ng already prints lots of useful information when the -V option is used. I'd add one more: print the version of Python used to compile mod-python. Most Linux distributions support multiple Python versions and often have multiple versions installed. Many packages still come with Python 2.7 support, but gradually changing to 3.X. 

For example RHEL has 2.7 and 3.4 in the distro, and 3.6 in EPEL. It is possible to find this information using package management tools, like on openSUSE:
```
linux-mewy:~ # rpm -q --requires syslog-ng-python
libc.so.6()(64bit)
libc.so.6(GLIBC_2.14)(64bit)
libc.so.6(GLIBC_2.2.5)(64bit)
libc.so.6(GLIBC_2.4)(64bit)
libpthread.so.0()(64bit)
libpython3.6m.so.1.0()(64bit)
libsyslog-ng-3.17.so.0()(64bit)
python(abi) = 3.6
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsXz) <= 5.2-1
syslog-ng = 3.17.2
```
But having this information in ``syslog-ng -V`` would make much easier to find this information and also make it distribution neutral.

Kokan:
This changes the plugin description, so it includes the python main version (3/2 as of now).
The user could query which version was the python module compiled:
```
syslog-ng --module-registry
...
Module: mod-python
Status: ok
Version: 3.17.2
Core-Revision: 3.17
Description:
  The python (python2) module provides Python scripted destination support for syslog-ng.
Plugins:
  destination     python
  root            python
  parser          python
  template-func   python
...
```

* Should it include also the minor/patch version of the python ?

Fixes #2336

### DIFF
--- a/modules/python/compat/compat-python.h
+++ b/modules/python/compat/compat-python.h
@@ -29,10 +29,12 @@
 
 #if (SYSLOG_NG_ENABLE_PYTHONv2)
 #define PYTHON_BUILTIN_MODULE_NAME "__builtin__"
+#define PYTHON_MODULE_VERSION "python2"
 #endif
 
 #if (SYSLOG_NG_ENABLE_PYTHONv3)
 #define PYTHON_BUILTIN_MODULE_NAME "builtins"
+#define PYTHON_MODULE_VERSION "python3"
 #endif
 
 void py_init_argv(void);
--- a/modules/python/python-plugin.c
+++ b/modules/python/python-plugin.c
@@ -89,7 +89,7 @@ const ModuleInfo module_info =
 {
   .canonical_name = "python",
   .version = SYSLOG_NG_VERSION,
-  .description = "The python module provides Python scripted destination support for syslog-ng.",
+  .description = "The python ("PYTHON_MODULE_VERSION") module provides Python scripted destination support for syslog-ng.",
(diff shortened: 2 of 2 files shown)


czanik:
Yes, please also include minor version. RHEL using EPEL has multiple 3.X versions available, so knowing which minor version is used is important information.

kira-syslogng:
Build SUCCESS

Kokan:
@czanik I can provide the version it was compiled with, not the version it is loaded by the OS. If you want to know about that you could simple `ldd ${prefix}/lib/syslog-ng/libmod-python.so|grep python`

czanik:
The version compiled with is perfect.

furiel:
What is the conclusion here? Do you add the minor version for the compiled version, or it is ready to be merged?

Kokan:
I think the minor version does not have much significant in this case, and would require additional digging. I would merge this as is, and later somebody could improve if they need it.

PR #2342, opened 2018-10-04
LogSource: reorder some operations on LogMessage

gaborznagy:
Some mangle callbacks depends on the original LogMessage that is received, therefore their invocation has to be made sooner.

As some processing is elemantary (e.g. hostname resolving), I hesitatedto call it first before anything else to avoid breaking things which assumes differently.

### DIFF
--- a/lib/logsource.c
+++ b/lib/logsource.c
@@ -348,6 +348,22 @@ _increment_dynamic_stats_counters(const gchar *source_id, const LogMessage *msg)
 }
 
 static void
+log_source_override_host(LogSource *self, LogMessage *msg)
+{
+  if (self->options->host_override_len < 0)
+    self->options->host_override_len = strlen(self->options->host_override);
+  log_msg_set_value(msg, LM_V_HOST, self->options->host_override, self->options->host_override_len);
+}
+
+static void
+log_source_override_program(LogSource *self, LogMessage *msg)
+{
+  if (self->options->program_override_len < 0)
+    self->options->program_override_len = strlen(self->options->program_override);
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

lib/logsource.c
```diff
@@ -395,14 +401,18 @@ log_source_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options
 
   log_msg_set_tag_by_id(msg, self->options->source_group_tag);
 
+
+  if (!_invoke_mangle_callbacks(s, msg, path_options))
+    return;
+
+  log_source_override_host(self, msg);
+  log_source_override_program(self, msg);
```
MrAnno: I would not extract the `self->options->*_override` conditions, if the names of the extracted functions do not indicate conditionality.

For example,
```C
if (self->options->program_override)
  log_source_override_program(self, msg);
```
is less misleading in my opinion.

MrAnno:
The functionality of `host_override` conflicts with one of our internal mangle callback implementations that wants to use the _original_ hostname.

Moving `log_source_override_host()` after the mangle callback invocation fixes the issue.

bazsi:
I like the extraction into functions, agreeing with the caveat of @MrAnno about naming.

Can't comment on the ordering but it's ok for me either way.

lib/logsource.c
MrAnno: Thanks.

kira-syslogng:
Build SUCCESS

PR #2347, opened 2018-10-09
Http load balancer

bazsi:
This branch implements client-side load balancing logic into the HTTP destination.


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

pzoleex:

modules/http/http-loadbalancer.c
```diff
@@ -0,0 +1,285 @@
...
+
+static HTTPLoadBalancerTarget *
+_locate_target(HTTPLoadBalancer *self, HTTPLoadBalancerClient *lbc)
+{
+  gint start_index = lbc->target
+                     ? (lbc->target->index + 1) % self->num_targets
+                     : 0;
+  for (gint i = 0; i < self->num_targets; i++)
+    {
+      HTTPLoadBalancerTarget *target = &self->targets[i + start_index];
```
lbudai: I think a modulo operation is missing from the index... 
As I understand, the key point here is to form a ring from the `targets` array, and the `start_index` would point to the head... but when the real index is computed, index can be out of bounds).

modules/http/http.c
```diff
@@ -306,14 +306,17 @@ http_dd_init(LogPipe *s)
   HTTPDestinationDriver *self = (HTTPDestinationDriver *)s;
   GlobalConfig *cfg = log_pipe_get_config(s);
 
+  if (self->load_balancer->num_targets == 0)
+    http_load_balancer_add_target(self->load_balancer, HTTP_DEFAULT_URL);
+
+  /* we need to set up url before we call the inherited init method, so our stats key is correct */
+  self->url = self->load_balancer->targets[0].url;
```
lbudai: Does this (using `targets[0]` as `url`) mean that when the user prepend a new target to the target list and reloads syslog-ng, we lost the diskq files?

modules/http/http-loadbalancer.h
```diff
@@ -0,0 +1,90 @@
...
+  gchar *url;
+  gint index;
+  /* read-write data, locking must be in effect */
+  HTTPLoadBalancerTargetState state;
+  gint number_of_clients;
+  gint max_clients;
+  time_t last_failure_time;
+};
+
+struct _HTTPLoadBalancerClient
```
lbudai: Note1:
We've talked about this IRL, but for completeness I'd still mention that I have the feeling that this hurts the YAGNI principle... on the other side, I know that when we extend the functionality, we will definitely need this structure.
Note 2: 
Maybe I'd use an index (that points to the targets array) instead of storing a pointer, but it's just a matter of taste (index or pointer... the result is almost the same, for me it is easier to read the code when index is used).

kira-syslogng:
Build FAILURE

modules/http/http-loadbalancer.h
```diff
@@ -0,0 +1,90 @@
...
+
+void http_lb_client_init(HTTPLoadBalancerClient *, HTTPLoadBalancer *lb);
+void http_lb_client_deinit(HTTPLoadBalancerClient *);
+
+struct _HTTPLoadBalancer
+{
+  GStaticMutex lock;
+  HTTPLoadBalancerTarget *targets;
+  gint num_targets;
+  gint num_clients;
```
lbudai: just a note (to other reviewers): 
`num_clients` is the number of workers (when I read the code, I always had to remember myself to this fact, otherwise it is easier to lost in num_* variables ). I understand that you don't want to connect the LoadBalancer class to the workers, but the num_clients maybe too 'sterile'

modules/http/http-loadbalancer.c
```diff
@@ -0,0 +1,285 @@
...
+    {
+      *new_target = _locate_target(self, lbc);
+      return TRUE;
+    }
+  return FALSE;
+}
+
+HTTPLoadBalancerTarget *
+http_load_balancer_choose_target(HTTPLoadBalancer *self, HTTPLoadBalancerClient *lbc)
+{
```
lbudai: 1.What if we wouldn't choose a new target when the number of workers less then the number of targets(`num_clients` < num_targets)?

modules/http/http-worker.c
```diff
@@ -355,23 +355,47 @@ _flush(LogThreadedDestWorker *s)
...
 
   if (self->super.batch_size == 0)
     return WORKER_INSERT_RESULT_SUCCESS;
 
   _finish_request_body(self);
 
   target = http_load_balancer_choose_target(owner->load_balancer, &self->lbc);
-  retval = _flush_on_target(self, target);
 
+  while (--retry_attempts >= 0)
```
lbudai: Could you factor out the this logic to a separate function?

modules/http/http-loadbalancer.c
bazsi: done. thanks for catching this.

modules/http/http.c
bazsi: I have added a warning for this case, advising the user to set persist-name()

modules/http/http-loadbalancer.h
bazsi: it's the number of HTTPLoadBalancerClient instances, in this way it is really the number of clients. Whether we have threads in this case is not a requirement.

I'd leave this as is.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lbudai:
@bazsi : I've added one additional patch (after the rebase/conflict res.) : `url` supports setting multiple urls, so the `urls` keyword is removed

kira-syslogng:
Build FAILURE

pzoleex:
@kira-syslogng test this please test branch=pzolee-http-load-balancer;

kira-syslogng:
Build SUCCESS

PR #2351, opened 2018-10-11
Fix license cmd

lbudai:
@czanik : FYI

### DIFF
--- a/lib/control/control-commands.c
+++ b/lib/control/control-commands.c
@@ -161,6 +161,12 @@ exit:
 }
 
 static GString *
+show_ose_license_info(GString *command, gpointer user_data)
+{
+  return g_string_new("You are using the Open Source Edition of syslog-ng.");
+}
+
+static GString *
 control_connection_reload(GString *command, gpointer user_data)
 {
   GString *result = g_string_new("OK Config reload initiated");
@@ -271,6 +277,7 @@ ControlCommand default_commands[] =
   { "QUERY", NULL, process_query_command },
   { "PWD", NULL, process_credentials },
   { "CONFIG", NULL, control_connection_config },
+  { "LICENSE", NULL, show_ose_license_info },
   { NULL, NULL, NULL },
 };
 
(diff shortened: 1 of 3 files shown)


lib/control/control-commands.c
```diff
@@ -58,6 +58,25 @@ control_register_command(const gchar *command_name, const gchar *description, Co
...
 
+void
+control_replace_command(const gchar *command_name, const gchar *description, CommandFunction function,
+                        gpointer user_data)
+{
+  GList *command_it =  g_list_find_custom(command_list, command_name,
+                                          (GCompareFunc)control_command_start_with_command);
+  if (!command_it)
+    {
+      msg_error("Failed to replace control command",
```
lbudai: hmm... I think we could register the command when we want to replace a non-existent (and print a warning msg): I'll change it.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/control/control-commands.c
```diff
@@ -58,6 +58,26 @@ control_register_command(const gchar *command_name, const gchar *description, Co
...
+control_replace_command(const gchar *command_name, const gchar *description, CommandFunction function,
+                        gpointer user_data)
+{
+  GList *command_it =  g_list_find_custom(command_list, command_name,
+                                          (GCompareFunc)control_command_start_with_command);
+  if (!command_it)
+    {
+      msg_warning("Trying to replace a non-existent command. Command will be registered as a new command.",
+                  evt_tag_str("command", command_name));
+      control_register_command(command_name, description, function, user_data);
```
bazsi: Do we need the warning here? 

If we do, I think we need a bit more context, like mentioning that this was a control socket command, but I like the idea that modules can register such stuff.

But all in all I would make this msg_debug().

gaborznagy:
I think nothing protects from re-registering a command with `control_register_command`, although you cannot override an already registered due to GList find.

We should rethink the `control_register_command` behaviour.

lib/control/control-commands.c
lbudai: changed to debug

lbudai:
@gaborznagy : fixed

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2356, opened 2018-10-13
 non-reliable diskq: fixing a false positive corruption detection

furiel:
This patchset addreses the following problems:
For users:
- When diskqueue was read by dqtool cat, syslog-ng tried to truncate the file even though it was opened as readonly, hence an error message was emitted. The code is fixed that syslog-ng will not try to truncate with readonly mode
- Fixed the dumped variables in the error log that is emitted when a truncate was unsuccessful
- Removing a false positive corruption detection error log.

For developers
- When corruption was detected, the error message was missing the write_head variable that is important information from debugging point of view.
- qdisk_init contained a branching based on reliable-non-reliable queue, that I could eliminate (refactor)
- _load_state contained a large chunk of code that was responsible for loading the queues of a non-reliable diskqueue, with a little code repetition. I could factor out a function, and eliminate the code duplications

A little more information about the corruption detection problem:

Previously, when a non-reliable diskqueue file is loaded, even though the qout, qbacklog qoverflow queues were truncated from the file, their pointers were not reset. As a consequence, the diskqueue file is kept in an invalid state while syslog-ng was running. These pointers were only set to valid values only when syslog-ng stopped gracefully, while saving the diskqueue. Suppose more messages are added to the queue, so that the write_head passes the invalid qout_ofs, then crash syslog-ng. In that case, the original invalid qout_ofs was kept in the diskqueue file, which is detected as a corruption at the next start, though the content of the disk queue was valid: at least the memory queue point of view, as everything in the memory was lost. 

Reproduction steps:
```
log {
    source { tcp(port(4444) flags(no-parse)); };
    destination { tcp(127.0.0.1 port(5555)
    disk-buffer(disk-buf-size(31457280) reliable(no)
    qout_size(64)));
  };
  flags(flow-control);
};
```

The destination must **NOT** be reachable.

Send in 128 messages:
```
for i in `seq 1 128`; do echo "hello AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" | nc localhost 4444; done
```

As qout is 64, 64 messages will be put into qout, 64 messages will be put into the disk queue.

Stop syslog-ng gracefully, with syslog-ng-ctl stop or ctrl-c. Then the qout will be written into the end of the diskqueue file, meaning qout will be larger than write_head.

Start syslog-ng again, and send in more messages that are stored in qout, for example again:
```
for i in `seq 1 128`; do echo "hello AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" | nc localhost 4444; done
```

In this case the write_head will go past qout. Note qout is not changed only when the queue is saved.

Then kill syslog-ng: for example with pkill -9 syslog-ng.

The result will be:
```

Inconsistent header data in disk-queue file, ignoring qout; filename='[...]', qout_ofs='35328', write_head='97792', qdisk_length='192'
--
Disk-buffer state loaded; filename='[...]', qout_length='64', qbacklog_length='0', qoverflow_length='0', qdisk_length='192'
Oct 12 13:28:09.084 localhost hello AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
[...]
```

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -60,15 +60,9 @@ typedef union _QDiskFileHeader
     gint64 write_head;
     gint64 length;
 
-    gint64 qout_ofs;
-    gint32 qout_len;
-    gint32 qout_count;
-    gint64 qbacklog_ofs;
-    gint32 qbacklog_len;
-    gint32 qbacklog_count;
-    gint64 qoverflow_ofs;
-    gint32 qoverflow_len;
-    gint32 qoverflow_count;
+    QDiskQueuePosition qout_pos;
+    QDiskQueuePosition qbacklog_pos;
+    QDiskQueuePosition qoverflow_pos;
     gint64 backlog_head;
     gint64 backlog_len;
   };
@@ -474,15 +468,15 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
       return FALSE;
     }
 
-  qout_count = self->hdr->qout_count;
-  qout_len = self->hdr->qout_len;
-  qout_ofs = self->hdr->qout_ofs;
-  qbacklog_count = self->hdr->qbacklog_count;
-  qbacklog_len = self->hdr->qbacklog_len;
-  qbacklog_ofs = self->hdr->qbacklog_ofs;
-  qoverflow_count = self->hdr->qoverflow_count;
-  qoverflow_len = self->hdr->qoverflow_len;
-  qoverflow_ofs = self->hdr->qoverflow_ofs;
+  qout_count = self->hdr->qout_pos.count;
+  qout_len = self->hdr->qout_pos.len;
+  qout_ofs = self->hdr->qout_pos.ofs;
+  qbacklog_count = self->hdr->qbacklog_pos.count;
+  qbacklog_len = self->hdr->qbacklog_pos.len;
+  qbacklog_ofs = self->hdr->qbacklog_pos.ofs;
+  qoverflow_count = self->hdr->qoverflow_pos.count;
+  qoverflow_len = self->hdr->qoverflow_pos.len;
+  qoverflow_ofs = self->hdr->qoverflow_pos.ofs;
 
   if ((self->hdr->read_head < QDISK_RESERVED_SPACE) ||
       (self->hdr->write_head < QDISK_RESERVED_SPACE) ||
@@ -690,17 +684,17 @@ qdisk_save_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
 
   memcpy(self->hdr->magic, self->file_id, 4);
 
-  self->hdr->qout_ofs = qout_ofs;
-  self->hdr->qout_len = qout_len;
-  self->hdr->qout_count = qout_count;
+  self->hdr->qout_pos.ofs = qout_ofs;
+  self->hdr->qout_pos.len = qout_len;
+  self->hdr->qout_pos.count = qout_count;
 
-  self->hdr->qbacklog_ofs = qbacklog_ofs;
-  self->hdr->qbacklog_len = qbacklog_len;
-  self->hdr->qbacklog_count = qbacklog_count;
+  self->hdr->qbacklog_pos.ofs = qbacklog_ofs;
+  self->hdr->qbacklog_pos.len = qbacklog_len;
+  self->hdr->qbacklog_pos.count = qbacklog_count;
 
-  self->hdr->qoverflow_ofs = qoverflow_ofs;
-  self->hdr->qoverflow_len = qoverflow_len;
-  self->hdr->qoverflow_count = qoverflow_count;
+  self->hdr->qoverflow_pos.ofs = qoverflow_ofs;
+  self->hdr->qoverflow_pos.len = qoverflow_len;
+  self->hdr->qoverflow_pos.count = qoverflow_count;
 
   if (!self->options->reliable)
     msg_info("Disk-buffer state saved",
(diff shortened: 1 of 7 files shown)


kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -445,15 +449,75 @@ _load_queue(QDisk *self, GQueue *q, gint64 q_ofs, gint32 q_len, gint32 q_count)
...
+    return FALSE;
+
+  if (!_try_to_load_queue(self, qoverflow))
+    return FALSE;
+
+  return TRUE;
+}
+
+static void
+_reset_queue_pointers(QDisk *self)
```
gaborznagy: We should reset the `..._count` members in the file too, they are already used in load_queue methods,
any later reference in load_state can use the saved values before load_queue.

modules/diskq/qdisk.h
```diff
@@ -41,6 +41,7 @@ typedef struct _QDisk QDisk;
 
 QDisk *qdisk_new(void);
```
gaborznagy: What about adding an extra argument to qdisk_init() ?
That is already a constructor to QDisk

modules/diskq/qdisk.c
```diff
@@ -511,6 +527,22 @@ _load_non_reliable_queues(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *q
   return TRUE;
 }
 
+static gboolean
+_load_non_reliable_queues(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
```
gaborznagy: I like this refactor, but I think we can resolve the code repeating problem without using macros and string mangling.
How about creating structs which contain the queue, and the additional members (ofs, len, count), then we create
an array out of these structures, on which we can freely iterate.

The error handling should be done outside of the iterator where we can access the names of the current index "qout", "qbacklog", etc.

kira-syslogng:
Build FAILURE

modules/diskq/qdisk.h
furiel: I tried, but in logqueue-disk.c: disk_init() is also used in _restart_diskq. In that occurrence it would be necessary to steal the magic id from the already created disk queue beforehand, that would be rewritten back to the disk. That would look strange from initialization perspective. Also, in that module QDisk is forward declared, so it would be necessary to add a getter to the api for the magic id. So in the end, my setter function would be replaced with a getter function, so the api complexity would not change much. I would leave as it is.

modules/diskq/qdisk.c
furiel: The main reason this macro complexity is involved is because I wanted to retain the original error message in case of corruption detection. By the way testdb also tests the error message. Though we believe with Gabor that it might worth breaking the error message for sake of reduced complexity.
With that, I believe I can simplify the macros considerably, and turn `_try_to_load_queue` into a function. I will try it out and get back with a new version.

modules/diskq/qdisk.h
gaborznagy: Do you mean "as it is", by leaving the `set_file_id()` function?

furiel: I've got an idea: the problem is LogQueueDisk wants to restart, but does not know how, because the restart depends on the type of the QDisk. But this can be resolved, if I add a `restart` virtual function to QDisk. With that, we would not need neither the setter and the getter, and qdisk_init could be leaned towards a true init function of qdisk by extending the parameters as you suggested. I'll try to implement this, and come back with a patch.

furiel:
Before merge, I would like to ask for help and approve from @pzoleex or @mitzkia because 002-dqtool-wrong-disk-buffers.zts needs to be updated.

modules/diskq/qdisk.c
```diff
@@ -453,50 +453,60 @@ _load_queue(QDisk *self, GQueue *q, gint64 q_ofs, gint32 q_len, gint32 q_count)
...
+{
+  gint64 ofs;
+  gint32 count;
+  gint32 len;
+  const gchar *type;
+} LoadParameters;
 
 static gboolean
-_load_non_reliable_queues(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
+_try_to_load_queue(QDisk *self, GQueue *queue, LoadParameters *params)
```
MrAnno: It's beautiful! 👍

modules/diskq/qdisk.c
```diff
@@ -494,20 +494,21 @@ _try_to_load_queue(QDisk *self, GQueue *queue, LoadParameters *params)
...
+    retval = FALSE; \
+  retval; \
+} )
+
 static gboolean
 _load_non_reliable_queues(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
 {
-  LoadParameters qout_parameters = create_load_parameter_for(qout);
-  if (!_try_to_load_queue(self, qout, &qout_parameters))
-    return FALSE;
```
MrAnno: I think this repetition is not harmful at all, especially if you have to use macros and token pasting operators to replace them.

I understand the necessity of `create_load_parameter_for()`, but in my opinion the macro-based `try_load_queue()` adds unnecessary complexity to a straightforward/harmless repetition.

modules/diskq/qdisk.c
gaborznagy: Thanks, I like the changes.

kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng retest this please test branch=diskq-corruption; test=functions/tools/dqtool/;

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -497,20 +497,21 @@ _try_to_load_queue(QDisk *self, GQueue *queue, LoadParameters *params)
...
 
+#define try_load_queue(queue) \
+( { \
+  gboolean retval = TRUE; \
+  LoadParameters queue ## _parameters = create_load_parameter_for(queue); \
+  if (!_try_to_load_queue(self, queue, &queue ## _parameters)) \
+    retval = FALSE; \
+  retval; \
+} )
+
```
lbudai: hmmmm....  I'd suggest to follow a different approach here.
What if we would have three different functions for loading `qout`, `qoverflow` and `qbacklog`?
Or, we could group the variables per queue into a type... and the loader could receive pointer to an instance with that type?

furiel: Fixed.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -63,15 +60,9 @@ typedef union _QDiskFileHeader
...
-    gint32 qout_count;
-    gint64 qbacklog_ofs;
-    gint32 qbacklog_len;
-    gint32 qbacklog_count;
-    gint64 qoverflow_ofs;
-    gint32 qoverflow_len;
-    gint32 qoverflow_count;
+    QDiskQueuePosition qout_pos;
+    QDiskQueuePosition qbacklog_pos;
+    QDiskQueuePosition qoverflow_pos;
```
MrAnno: I've validated that the offset of every field remained the same on my machine, but I'm not sure about other platforms. I find this refactor step a bit risky.

As we pointed out IRL, the manual padding in`QDiskFileHeader` is incorrect (it was incorrect in the original version as well), an addition 1-byte padding is missing after `guint8 _pad1;`

lbudai: @MrAnno: we can do further tests on various cpu architectures... ( as each 'group' is 128=64+32+32 bytes, I don't think it would be an issue, more details:
 * in C, there are no lead padding for structs
 * alignment will be 64 (as the widest type is uint64_t in the struct)
 * as the other two members are 32 bits, I think there is no chance for having a padding inside the struct (nor trailing padding)
 * we don't have to take structure reordering into account (but it wouldn't cause any issue)
 )

PR #2358, opened 2018-10-15
Move some messages to trace

bazsi:
I find the latest messages clutter up debug messages a lot, so move some of the less interesting ones
to msg_trace(), now that we have that.

### DIFF
--- a/lib/logproto/logproto-buffered-server.c
+++ b/lib/logproto/logproto-buffered-server.c
@@ -667,8 +667,8 @@ log_proto_buffered_server_fetch_into_buffer(LogProtoBufferedServer *self)
   else if (rc == 0)
     {
       /* EOF read */
-      msg_verbose("EOF occurred while reading",
-                  evt_tag_int(EVT_TAG_FD, self->super.transport->fd));
+      msg_trace("EOF occurred while reading",
+                evt_tag_int(EVT_TAG_FD, self->super.transport->fd));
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

szemere:
FYI: I am currently checking control flow, if there are any message "before/after" the changed ones which should be also changed as well.

lib/logproto/logproto-buffered-server.c
```diff
@@ -667,8 +667,8 @@ log_proto_buffered_server_fetch_into_buffer(LogProtoBufferedServer *self)
   else if (rc == 0)
     {
       /* EOF read */
-      msg_verbose("EOF occurred while reading",
-                  evt_tag_int(EVT_TAG_FD, self->super.transport->fd));
+      msg_trace("EOF occurred while reading",
```
gaborznagy: Would you change it in `logproto-framed-server.c` too, please?

gaborznagy:
@bazsi what do you think about `Message was cloned` debug message?
I don't want to discuss every part I just came across it due to `default-network-drivers()`. :)

bazsi:
I have now added the call sites requested in the comments, however I would not want to add more, so if you have any additions, please open another PR.

MrAnno:

kira-syslogng:
Build SUCCESS

PR #2371, opened 2018-10-24
basicfuncs: fix $(basename) and $(dirname) in the presence of a prefix

### LINKED ISSUE #2361: Problem with basename function of syslog-ng
## Version of syslog-ng
syslog-ng 3.14.1

## Platform
Ubuntu 16.04.3(server-amd64)

## Steps to reproduce
1. Configuration
vim /etc/syslog-ng/syslog-ng.conf

```
source s_dir {
    wildcard-file(
        base-dir("/var/log/test/")
        filename-pattern("*.log")
        recursive(no)
    );
};

destination d_file {
    file("/data/logs/$(basename ${FILE_NAME})"
        create-dirs(yes)
        dir-perm(0755)
    );
};

log { source(s_dir); destination(d_file); };
```

2. start syslog-ng
systemctl restart syslog-ng

3. output logs 
"/var/log/test/*.log" expected to be collected under "/data/logs/" directory, but actually under "/" directory.

bazsi:
This should fix #2361

### DIFF
--- a/modules/basicfuncs/fname-funcs.c
+++ b/modules/basicfuncs/fname-funcs.c
@@ -30,7 +30,7 @@ tf_basename(LogMessage *msg, gint argc, GString *argv[], GString *result)
   gchar *base;
 
   base = g_path_get_basename(argv[0]->str);
-  g_string_assign(result, base);
+  g_string_append(result, base);
   g_free(base);
 }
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

modules/basicfuncs/fname-funcs.c
```diff
@@ -30,7 +30,7 @@ tf_basename(LogMessage *msg, gint argc, GString *argv[], GString *result)
   gchar *base;
 
   base = g_path_get_basename(argv[0]->str);
-  g_string_assign(result, base);
+  g_string_append(result, base);
```
szemere: Just for the reference:
https://github.com/balabit/syslog-ng/blob/da326364d52a21ebaf2d8fd17f4077fc5cc954e6/lib/template/templates.c#L170

For me this was not trivial why we can eliminate the initialization of the result variable. Seems like it was unnecessary in the first place.

PR #2372, opened 2018-10-24
tls: disable ssl compression by default

furiel:
Currently syslog-ng does not change the tls compression setting of tls transport. This means the default setting depends on the patch level and compile options of openssl. H̶o̶w̶e̶v̶e̶r̶ ̶t̶h̶e̶ ̶d̶e̶f̶a̶u̶l̶t̶ ̶c̶o̶m̶p̶r̶e̶s̶s̶i̶o̶n̶ ̶s̶e̶t̶t̶i̶n̶g̶ ̶s̶h̶o̶u̶l̶d̶ ̶b̶e̶ ̶d̶i̶s̶a̶b̶l̶e̶d̶ ̶b̶y̶ ̶d̶e̶f̶a̶u̶l̶t̶ ̶t̶o̶ ̶a̶v̶o̶i̶d̶ ̶C̶R̶I̶M̶E̶ ̶v̶u̶l̶n̶e̶r̶a̶b̶i̶l̶i̶t̶y̶.̶ Edit: syslog-ng is not vulnerable to CRIME.

T̶h̶i̶s̶ ̶p̶a̶t̶c̶h̶s̶e̶t̶ ̶e̶x̶p̶l̶i̶c̶i̶t̶e̶l̶y̶ ̶d̶i̶s̶a̶b̶l̶e̶s̶ ̶t̶l̶s̶ ̶c̶o̶m̶p̶r̶e̶s̶s̶i̶o̶n̶ ̶t̶o̶ ̶b̶e̶ ̶s̶u̶r̶e̶.̶
This patchset makes tls compression configurable, with disabled by default.

Note: allow-compress(yes) does not necessarily mean the session will be compressed. It only means endpoints are allowed to choose tls compression during negotiation. For example if the openssl is compiled without zlib either on the local or the remote side, compression will not happen.

### DIFF
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -66,6 +66,19 @@ typedef enum
   TLS_CONTEXT_PASSWORD_ERROR
 } TLSContextLoadResult;
 
+void
+tls_session_configure_allow_compress(TLSSession *tls_session, gboolean allow_compress)
+{
+  if (allow_compress)
+    {
+      SSL_clear_options(tls_session->ssl, SSL_OP_NO_COMPRESSION);
+    }
+  else
+    {
+      SSL_set_options(tls_session->ssl, SSL_OP_NO_COMPRESSION);
+    }
+}
+
 gboolean
 tls_get_x509_digest(X509 *x, GString *hash_string)
 {
--- a/lib/tlscontext.h
+++ b/lib/tlscontext.h
@@ -108,6 +108,7 @@ TLSVerifier *tls_verifier_ref(TLSVerifier *self);
 void tls_verifier_unref(TLSVerifier *self);
 
 
+void tls_session_configure_allow_compress(TLSSession *tls_session, gboolean allow_compress);
 gboolean tls_context_set_verify_mode_by_name(TLSContext *self, const gchar *mode_str);
 gboolean tls_context_set_ssl_options_by_name(TLSContext *self, GList *options);
 gint tls_context_get_verify_mode(const TLSContext *self);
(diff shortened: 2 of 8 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
I am yet to be convinced that CRIME applies to syslog, and at one point it
was explicitly requested to allow compression in one way or the other, and
ssl was the solution.

So in what way are we vulnerable to crime? And can you point me how this
was fixed in openssl in distros? I would even argue to make it possible to
enable compression even if its disabled by default.

Ssl is not just for http.

Bazsi

lbudai:
@bazsi :
I think, syslog-ng is not vulnerable to CRIME. On the other hand, I don't see any problem if someone need to explicitly set the allow-compression flag... (especially if newest OpenSSLs are built without compression support by distros; just check CONFARGS in debian/rules : https://launchpad.net/ubuntu/+source/openssl/1.1.0g-2ubuntu4.1, they are using `no-zlib` option ). As I understand, users can still set the allow-compress flag, but they need to set it manually.

kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng test this please test branch=micek_allow_compress;

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2373, opened 2018-10-24
Add contextual data error reporting improvements & csv-scanner refactor

bazsi:
This branch addresses a couple of things:

1) it improves on add-contextual-data() error handling a lot, reports the unparseable lines, the kind of structure we expect in the CSV file.
2) it refactors csv-scanner to use an explicit state variable, instead of deducing state based on the value of various internal variables. It also adds a unit test program.
3) it fixes csv-parser(flags(drop-invalid)) so it actually drops empty lines (which was not an intended behaviour)
4) to enables add-contextual-data() to ignore empty lines in its CSV file input.

It also improves add-contextual-data() runtime messages during processing to make its behavior easier to understand.


kira-syslogng:
Build FAILURE

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -376,7 +376,9 @@ csv_scanner_scan_next(CSVScanner *self)
 const gchar *
 csv_scanner_get_current_name(CSVScanner *self)
```
szemere: With the original implementation there are only two cases where `csv_scanner_get_current_name` could not work:
 - There are more values than columns in the csv file, and we are not `CSV_SCANNER_GREEDY` . (In this case the `current_column` was setted to `NULL` during switch to next column, and we neglected the rest of the data.)
 - The other case was when there were absolutely no data at all in the line. But this was handled here https://github.com/balabit/syslog-ng/blob/da326364d52a21ebaf2d8fd17f4077fc5cc954e6/modules/csvparser/csvparser.c#L140
by the fact, that `csv_scanner_scan_next` returns false, and we stopped the parsing, and did not try to print any message with the `current_column`.

After the patch, `csv_scanner_get_current_name` will return wrong value in the first case. Fortunately we still do not print any message in this case. So practically it will not cause any problem. (Just CleanCode point of view.)

The second scenario - when the input is empty - is still valid, and the "selector" name is a pretty good error message for this. However I would prefer to handle the "no data" case by skipping the line. (This will also bring an another benefit: users may seperate their entry blocks with empty lines.)

Thank You @furiel for the investigation.

modules/add-contextual-data/add-contextual-data.c
```diff
@@ -249,8 +249,9 @@ _load_context_info_db(AddContextualData *self)
   FILE *f = _open_data_file(self->filename);
   if (!f)
     {
-      msg_error("Error loading add_contextual_data database",
-                evt_tag_str("filename", self->filename));
+      msg_error("add-contextual-data(): Error opening database",
+                evt_tag_str("filename", self->filename),
+                evt_tag_str("error", g_strerror(errno)));
```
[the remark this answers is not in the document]
bazsi: thanks for spotting this, fixed.

lib/scanner/csv-scanner/csv-scanner.c
bazsi: I am not sure I understand everything here. I was handling the empty line case, but not in csv-parser() rather in add-contextual-data(), which has not handled empty lines, and I have not added that.

I agree It'd be a good idea to handle them there, as the csv file is indeed a configuration file, however as csv-parser() is intended to be used on incoming logs, we shouldn't specially handle empty lines.

And indeed, csv-parser() accepts empty lines, but I think it is a bug. 

I think we should discuss this IRL.

kira-syslogng:
Build FAILURE

lib/scanner/csv-scanner/csv-scanner.c
bazsi: Ssoo, I have added a refactor of csv-scanner so that it behaves more consistently and also addresses the things we discussed in this comment:

 * csv-parser() will drop empty lines instead of accepting them (this was a bug)
 * add-contextual-data() will ignore empty lines, instead of frowning on them (this is a feature)

At the same time, I have added a unit test at the csv-scanner level (earlier it was only tested through csv-parser()). This unit test is not yet complete, but at least shows the consistent usage of the API. Stuff like separators, and escaping modes are not tested here, rather they still are tested through csv-parser().

with that I consider this comment addressed. I am rewriting the summary of the PR so it better reflects what we have here.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng test this please test branch=add-contextual-data-refactor;

pzoleex:
@kira-syslogng test this please test branch=add-contextual-data-refactor;

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng test this please test branch=add-contextual-data-refactor;

furiel:
@kira-syslogng test this please test branch=add-contextual-data-refactor;

kira-syslogng:
Build SUCCESS

lib/scanner/csv-scanner/csv-scanner.c
```diff
@@ -344,23 +322,59 @@ _translate_value(CSVScanner *self)
...
+_switch_to_next_column(CSVScanner *self)
+{
+  g_string_truncate(self->current_value, 0);
+
+  switch (self->state)
+    {
+    case CSV_STATE_INITIAL:
+      self->state = CSV_STATE_COLUMNS;
+      self->current_column = self->options->columns;
+      if (self->current_column)
```
szemere: Do we allow scanning without defining columns?

PR #2370, opened 2018-10-23
Issue 2306 | auto reload syslog when config changed on disk

madsalama:
Closes #2306

PR #2375, opened 2018-10-25
Auto reload syslog when config changed on disk

### LINKED ISSUE #2306: auto reload syslog when config changed on disk
_Originally posted by @faxm0dem in https://github.com/balabit/syslog-ng/issues/2304#issuecomment-423956534_

This option would watch the configuration file for changes, and when there have been a change it automatically reloads syslog-ng.

madsalama:
- Closes #2306 | Invalidates PR #2370


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
internal mi: This user does not have permission to start the build. Can one of the admins verify this patch and start the build? (admin please type: ok to test)

bazsi:
Guys, do we really need a complex polling implementation instead of simply checking if the config changed every second?

The only flipside is that we would need roughly 1sec to notice that the file has changed.

All it needs is a single timer that runs every second.

madsalama:
@bazsi - I agree this might be an overkill, but ``inotify`` is a better solution for systems that support it; since its events are triggered by the kernel (which does the I/O read/write anyway) instead of the process blocking/polling and re-accessing the file and comparing every once in a while. In my opinion its easier to use (but it has its quirks which I handled already).

Kokan:
@bazsi I would also prefer to have an `inotify` based solution, as the timer would also run the handler in the main thread, which most likely won't cause any performance degradation; but on the other hand the timer would be triggered every 1s that is a big waste compared to the expected frequency of the config file change.

madsalama:
@Kokan - For handling the case where ``inotify`` is not supported, how long do you recommend the timeout to be?

madsalama:
@Kokan - any updates on this? I'd like to continue working on it if needed modifications.

lib/monitor.c
```diff
@@ -0,0 +1,173 @@
...
+
+void
+file_monitor_start(FileMonitor *fm)
+{
+  IV_INOTIFY_INIT(&(fm->inotify));
+  iv_inotify_register(&(fm->inotify));
+
+  IV_INOTIFY_WATCH_INIT(fm->w);
+  fm->w->inotify = &(fm->inotify);
+  fm->w->mask = IN_ALL_EVENTS;
```
Kokan: It should only subscribe to events, that are relevant (let others filter instead of us).

lib/monitor.c
```diff
@@ -0,0 +1,173 @@
...
+handler(void *_fm, struct inotify_event *event)
+{
+  FileMonitor *fm = (FileMonitor *) _fm;
+
+  if (event->mask & IN_MODIFY)
+    fm->main_loop_reload_config(fm->self);
+
+  // 1- handle case where editing the file using vim causes an event 'IN_IGNORED'
+  //     and subsequently the watch gets deleted by ivykis: ivykis/src/iv_inotify.c:88
+  //   2- handle case when file is edited and contains a parse error, syslog-ng changes
```
Kokan: I do not think parsing error need special handling, just simply trigger the reload (and that if the config fails fall back if can to the previous config).
I would ignore the case when because of attribute change the config is not possible to read anymore. Most of the time syslog-ng is running as root anyway (or with DAC_OVERRIDE cap by default). If it turns out that is a needed feature, probably somebody is going to ask for it, and it won't be hard to do that time.

lib/monitor.c
```diff
@@ -0,0 +1,173 @@
...
+  //     file attributes and file becomes unreadable
+
+  //    - note : for case when file was deleted, syslog-ng would still attempt to reload,
```
Kokan: Maybe this case should be handled ? (deletion)
But I think the other use case is when a file is moved, as I think that is not handled correctly now.

lib/monitor.c
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+  fm->file_path = file_path;
+  read_file(fm->file_path, &(fm->current_config));
+
+  return fm;
+}
+
+void
+file_monitor_config_reload_callback_function(FileMonitor *fm, void (*mainloop_callback)(void *), gpointer *self)
```
Kokan: This function is repeated because of the `#if #else #endif`, but it can be avoided, as this function is the same in either case.

lib/monitor.c
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+static void
+handler(void *_fm)
+{
+  FileMonitor *fm = (FileMonitor *) _fm;
+  gchar *file_contents = NULL;
+
+  read_file(fm->file_path, &file_contents);
```
Kokan: Instead of this, you could just use `stat` and get the modified time and compare to the one when started.

madsalama: That's indeed a better idea!

lib/monitor.c
madsalama: If the file is moved, the watch on the original file is eventually removed by the kernel so it's the same as if the file was deleted (nothing to watch for modifications - since the watch is not dynamic and will not track original file path changes). 

We can watch on the event ``IN_MOVED`` but it's not very useful since if user moved the original file back to its path, the watch would've been gone anyway and no way to re-watch other than restarting syslog-ng.

lib/monitor.c
madsalama: The special handling is not related to inotify itself, but the way syslog-ng - for some reason - modifies the attributes of the configuration file if a reload has been issued with a problematic file, and then inotify is not able to watch it anymore because it's marked as 'deleted'. I don't know the root cause for that but I will reproduce it and confirm to double check.

Kokan:
@madsalama I spoke with @bazsi and talked about the issue he wrote you about before.
I know this PR has its long life, but the conclusion was that it would be nice if the `DirectoryMonitor` could be changed so it supports both use case (affile and this config file change trigger).
Is is something that you would explore ? One option would be to make: `directory_monitor_set_new_file_callback` and `directory_monitor_set_rm_file_callback`, something like that.

madsalama:
> 
> 
> @madsalama I spoke with @bazsi and talked about the issue he wrote you about before.
> I know this PR has its long life, but the conclusion was that it would be nice if the `DirectoryMonitor` could be changed so it supports both use case (affile and this config file change trigger).
> Is is something that you would explore ? One option would be to make: `directory_monitor_set_new_file_callback` and `directory_monitor_set_rm_file_callback`, something like that.

Actually that was my original plan initially. You can find relevant questions on the original PR. I disregarded it as it was easier to build something from scratch. Nonetheless I think I can give it a try, not sure how feasible it could be though.

lib/mainloop.c
```diff
@@ -607,6 +608,14 @@ main_loop_run(MainLoop *self)
       cfg_load_module(self->current_configuration, "mod-python");
       debugger_start(self, self->current_configuration);
     }
+
+  if (self->options->auto_reload)
+    {
+      FileMonitor *fm = file_monitor_create(resolvedConfigurablePaths.cfgfilename);
```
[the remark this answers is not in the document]
madsalama: @Kokan - I took that into consideration for the init/deinit of the monitor in ``mainloop.c``. Please check.

lib/Makefile.am
```diff
@@ -258,6 +259,7 @@ lib_libsyslog_ng_la_SOURCES		= \
 	lib/host-id.c			\
 	lib/resolved-configurable-paths.c \
 	lib/window-size-counter.c \
+        lib/monitor.c \
```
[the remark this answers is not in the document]
madsalama: It is still failing for some reason, I copied the directory monitor files into ``lib`` and updated ``Makefile.am``.

lib/directory-monitor-poll.c
```diff
@@ -85,16 +109,31 @@ _rescan_directory(DirectoryMonitorPoll *self)
...
       const gchar *filename;
       while((filename = g_dir_read_name(directory)))
         {
-          collection_comparator_add_value(self->comparator, filename);
+          full_path = build_filename(self->super.real_path, filename);
+          collection_comparator_add_value(self->comparator, full_path);
+
+          /* compare entry mtime vs stat result */
+          CollectionComparatorEntry *entry = g_hash_table_lookup(self->comparator->original_map, filename);
+          if (entry->flag == IN_BOTH)
```
madsalama: @Kokan - FYI, I've tried this with and without the ``IN_BOTH`` condition and still only worked while editing the file and doing ``:w!`` only in vim.

lib/collection-comparator.h
```diff
@@ -24,6 +24,29 @@
 
 #include "syslog-ng.h"
 
+#define IN_OLD   0x01
```
Kokan: I do not think this shall be exposed in the header.

lib/directory-monitor-poll.c
```diff
@@ -85,16 +106,37 @@ _rescan_directory(DirectoryMonitorPoll *self)
...
+
   collection_comparator_start(self->comparator);
   if (directory)
     {
       const gchar *filename;
       while((filename = g_dir_read_name(directory)))
         {
           collection_comparator_add_value(self->comparator, filename);
+
+          full_path = build_filename(self->super.real_path, filename);
```
Kokan: This block imho in the wrong level of abstraction. The poll version of directory monitor uses the `collection_comparator` inside to decide what state a file is in. (created, deleted, ...)
The part where it request `stat` and compare should be in the `collection_comparater`, and here the `collection_comparator_add_value` should hide the change inside it.

lib/directory-monitor-poll.c
```diff
@@ -85,16 +106,37 @@ _rescan_directory(DirectoryMonitorPoll *self)
...
+          entry = g_hash_table_lookup(self->comparator->original_map, filename);
+
+          if (g_stat(full_path, &st))
+            perror(full_path);
```
Kokan: We do not use `perror` to report issues, please use the `msg_` macros to do the same. (`msg_errror`, `msg_warning`, ..., `msg_trace`).

lib/directory-monitor-poll.c
```diff
@@ -85,16 +106,37 @@ _rescan_directory(DirectoryMonitorPoll *self)
...
+          else if (self->initial_rescan)
+            entry->mtime = st.st_mtime;
```
Kokan: If the current file is not in the map, this line is a null pointer dereference.

lib/directory-monitor-poll.c
```diff
@@ -85,16 +106,37 @@ _rescan_directory(DirectoryMonitorPoll *self)
...
+          else if (entry->mtime != st.st_mtime)
```
Kokan: If `g_stat` was not successful can we use the `st` ? I think we should assume in that case that the `st` is uninitialized.

lib/mainloop.c
```diff
@@ -365,6 +366,51 @@ main_loop_verify_config(GString *result, MainLoop *self)
...
+
+/* callback */
+static void
+_on_file_modified(const DirectoryMonitorEvent *event, gpointer user_data)
+{
+  MainLoop *self = (MainLoop *) user_data;
+
+  if (!strcmp(event->full_path, resolvedConfigurablePaths.cfgfilename)
+      && ((event->event_type == FILE_MODIFIED) || (event->event_type == FILE_CREATED))
+      && (monitor->initial_collect == FALSE))
```
Kokan: I do not like the idea of `initial_collect` flag, but I guess this is triggered for the `FILE_CREATED` when it starts.
* I would either say to make the directory monitor configurable whether it should ignore the initial triggers at first scan or not. (this way it is hidden in the directory monitor)
* Handle here that it ignores the first `FILE_CREATED` for the `cfgfilename` (we can assume if we are here that there was a file when syslog-ng started).

I guess it is more like the call side responsibility, so I would do the second, even thou that not pretty.
Something like this:
```c
static void
 _on_file_modified(const DirectoryMonitorEvent *event, gpointer user_data)
 {
   static gboolean ignore_first_time_created_event = TRUE;
   MainLoop *self = (MainLoop *) user_data;

   if (strcmp(event->full_path, resolvedConfigurablePaths.cfgfilename))
      return;

   if (ignore_first_time_created_event && event_type == FILE_CREATED)
     {
         ignore_first_time_created_event = FALSE;
         return;
     } 

   if ((event->event_type == FILE_MODIFIED) || (event->event_type == FILE_CREATED))
     {
       main_loop_reload_config(self);
     }
 }
```
(disclaimer: this is just a sketch and not tested/compiled)

lib/directory-monitor-poll.c
madsalama: Could there be a case where it not going to be in the map after ``collection_comparator_add_value(self->comparator, filename);`` has just been called?

Kokan: Well at this level the `collection_comparator_add_value` is black box, but I would say move these things inside, and probably this won't be an issue.

lib/mainloop.c
madsalama: Shouldn't ``static gboolean ignore_first_time_created_event = TRUE;`` be outside the callback though?

lib/directory-monitor-poll.c
madsalama: I judged based on the fact that ``collection_comparator_add_value`` will either add an entry if not exists or update the entry flag if it already exists, thus, there will always be an entry. But I see your perspective.

lib/mainloop.c
Kokan: Why should it ? It is only needed in the that function.

madsalama: I'm thinking, it will be set to ``TRUE`` every time the callback is invoked, isn't it?

madsalama: I'm thinking, it will be set to ``TRUE`` every time the callback is invoked, won't it?

Kokan: No, it is assigned only the first time. You can try:
```c#include <stdio.h>

void sayhello()
{
  static int once = 1;

  if (!once) return;

  once = 0;
  printf("hello\n");
}


int main()
{
  sayhello();
  sayhello();
  sayhello();
  sayhello();
 
  return 0;
}

```

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

lib/directory-monitor-poll.c
```diff
@@ -123,15 +157,32 @@ _start_watches(DirectoryMonitor *s)
...
       const gchar *filename = g_dir_read_name(directory);
       while (filename)
         {
-          collection_comparator_add_initial_value(self->comparator, filename);
+          fullpath = build_filename(self->super.real_path, filename);
+          if (g_stat(fullpath, &stat_result))
+            {
+              msg_debug("Cannot stat file/directory",
+                        evt_tag_str("filename", fullpath),
+                        evt_tag_str("error", error->message));
```
Kokan: The `error` is not used before - but initialized to `NULL` - thus `error->message` is a null pointer dereference.

madsalama: Now fixed for the ``stat`` block. FYI - ``error`` is actually used:
```
  GError *error = NULL;
  gchar *fullpath = NULL;
  GStatBuf stat_result;
  GDir *directory = g_dir_open(self->super.real_path, 0, &error);
```

lib/tests/test_collection_comparator.c
```diff
@@ -59,6 +59,12 @@ _handle_new(const gchar *filename, gpointer data)
   g_hash_table_insert(self->new_entries, g_strdup(filename), (gchar *)TEST);
 }
 
+static void
+_handle_modify(const gchar *filename, gpointer data)
+{
+  /* TODO: write test code */
```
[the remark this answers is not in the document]
madsalama: I'm not quite sure I understand what's needed here. Could you please elaborate? I'm new to Criterion.

lib/directory-monitor.c
```diff
@@ -172,7 +172,12 @@ _collect_all_files(DirectoryMonitor *self, GDir *directory)
       gchar *filename_real_path = resolve_to_absolute_path(filename, self->real_path);
       event.full_path = build_filename(self->real_path, filename);
       event.event_type = g_file_test(filename_real_path, G_FILE_TEST_IS_DIR) ? DIRECTORY_CREATED : FILE_CREATED;
-      self->callback(&event, self->callback_data);
+
+      if(self->callback)
```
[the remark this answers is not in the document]
madsalama: I may revert that change since it used to be ``self->callback(&event, self->callback_data);`` and that check wasn't done anyway. (This seems outside of the scope for this PR).

kira-syslogng:
Build FAILURE

lib/directory-monitor.c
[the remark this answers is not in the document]
Kokan: I think revert would be a best option; maybe an assert. But I would not do the `msg_error`, as that could be ignored.

lib/tests/test_collection_comparator.c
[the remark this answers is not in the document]
Kokan: At first I would avoid having this `TODO` comment.
You have to create this dummy function as the collection comperator interface changed, but none of the existing test should call this new function, and you could assert for that. Like if this function is called that mean a test failed.
For tests that are testing modify trigger, sure there is a need for a different function.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lib/directory-monitor-poll.c
madsalama: @Kokan - now ``error`` usage removed from all occurrences relating to this PR functionality. I left the one relating to ``open_dir`` since it's not related to this PR.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/directory-monitor-inotify.c
```diff
@@ -56,6 +56,10 @@ _get_event_type(struct inotify_event *event, gchar *filename)
     {
       return DIRECTORY_DELETED;
     }
+  else if (event->mask & IN_MODIFY)
```
furiel: Similar question again: do we need to change inotify based monitoring?

lib/mainloop.c
```diff
@@ -398,6 +399,59 @@ main_loop_verify_config(GString *result, MainLoop *self)
...
+  if (strcmp(event->full_path, resolvedConfigurablePaths.cfgfilename))
+    return;
+
+  if (ignore_first_time_created_event && (event->event_type == FILE_CREATED))
+    {
+      ignore_first_time_created_event = FALSE;
+      return;
+    }
+
+  if ((event->event_type == FILE_MODIFIED) || (event->event_type == FILE_CREATED))
```
furiel: If user_data would store the last modification time of the cfgfile besides mainloop, then you could compare the modification time from `g_stat` here. That would eliminate the need of `mtime` parameter in `collection_comparator`.

lib/collection-comparator.c
```diff
@@ -79,12 +83,18 @@ collection_comparator_start(CollectionComparator *self)
...
+collection_comparator_add_value(CollectionComparator *self, const gchar *value, time_t mtime)
 {
   CollectionComparatorEntry *entry = g_hash_table_lookup(self->original_map, value);
+
   if (entry)
     {
       entry->flag = IN_BOTH;
+      if (mtime && (entry->mtime != mtime))
+        {
+          self->handle_modified_entry(entry->value, self->callback_data);
```
furiel: As mtime is never zero (except when `g_stat` fails), this callback is called for each files that are already tracked. Which currently includes a memory allocation, full path calculation, and a deallocation, which is unnecessary from wildcard file source perspective. Wildcard filesource can work on thousands of files, and executing these per rescan is unnecessary work.

This problem persists even with my previous comment how we could eliminate mtime here.

To resolve this, either handle_modified_entry should be set optionally, and called here like:
```
if (self->handle_modified_entry)
  self->handle_modified_entry(entry->value, self->callback_data);
```
or collection comparator could remember which events it should handle. For this case a mode field could be added  to `CollectionComparator` (for example `guint32 events`, with flags `COLLECTION_COMPARATOR_EVENT_NEW`,` ..._DELETED`,` ..._MODIFIED`).

Probably this latter is easier to impement.

lib/directory-monitor-inotify.c
```diff
@@ -83,7 +87,7 @@ _start_watches(DirectoryMonitor *s)
   IV_INOTIFY_WATCH_INIT(&self->watcher);
   self->watcher.inotify = &self->inotify;
   self->watcher.pathname = self->super.dir;
-  self->watcher.mask = IN_CREATE | IN_DELETE | IN_MOVE | IN_DELETE_SELF | IN_MOVE_SELF;
+  self->watcher.mask = IN_CREATE | IN_DELETE | IN_MOVE | IN_DELETE_SELF | IN_MOVE_SELF | IN_MODIFY;
```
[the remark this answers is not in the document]
madsalama: The configuration checking uses the ``DirectoryMonitor`` module - which decides which monitoring type to use - either only polling or inotify (used when supported). Thus I decided that the inotify part should be modified for syslog-ng to register for the ``IN_MODIFY`` event and define a corresponding callback.

lib/directory-monitor-poll.c
```diff
@@ -86,14 +104,28 @@ static void
...
   GDir *directory = g_dir_open(self->super.real_path, 0, &error);
   collection_comparator_start(self->comparator);
+
   if (directory)
     {
       const gchar *filename;
       while((filename = g_dir_read_name(directory)))
         {
-          collection_comparator_add_value(self->comparator, filename);
+          fullpath = build_filename(self->super.real_path, filename);
```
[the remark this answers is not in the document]
madsalama: I have not checked the wildcard file source implementation in detail, but it seems that fullpath is needed for ``g_stat``. As far as I remember - it didn't work otherwise. 
[https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-stat](url)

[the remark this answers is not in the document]
furiel: Can you highlight the sentence that writes we need full path? I must be overlooking something, but I cannot find it.

I tried with a small c code, and g_stat works for me for relative paths:
```
$ touch relative_path.txt; cat | gcc -xc - `pkg-config glib-2.0 --cflags` `pkg-config glib-2.0 --libs` && ./a.out
#include <glib/gstdio.h>
#include <assert.h>

int main() {
  GStatBuf result;
  assert(0 == g_stat("./relative_path.txt", &result));
  printf("size: %lu\n", result.st_mtime);
};
size: 1546934223
$ 
```

lib/directory-monitor-inotify.c
furiel: I see now inotify needs to be changed too from your explanation above. Resolving the comment.

lib/directory-monitor-poll.c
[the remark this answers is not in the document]
madsalama: you're right I need to take a deeper look on this.

lib/mainloop.c
furiel: Yes sorry, I should have explained my problem first, before offering an alternative solution.

From clean code perspective I have difficulties in understanding the `CollectionComparator` mechanics the way it is implemented now. The original design and implementation was clean, it served the purpose of checking tracked (general purpose) values, and executed callbacks based on a value has newly added or removed from the tracked list. And I can also imagine a improved version with modification time: we have 3 callbacks, new, removed, modified, that can be called respectively, which is your current intention to implement. Storing `mtime` for this sake in `CollectionComparatorEntry` is also reasonable.

My problem is the way `mtime` is injected into `CollectionComparator` when calling `collection_comparator_add_value`. The current implementation passes `time_t modified`, which looks out of place.

Passing modified time during `..._add` adds too much freedom, callers allow to update modification time at their will, they can set to zero, they can set into the the past or the future, without control. To avoid this, users are highly suggested (forced) to create a function that calculates the modification time before calling `..._add`, resulting in a repetitive pattern.

In my opinion it would be much cleaner if `CollectionComparator` could be taught how it could detect the modification change (which could use even a method that is not based on mtime, but I would not go that far in this scope, as it would probably YAGNI).

And there are alternatives to achieve this.
- One would be to specify another callback for collection comparator (`gboolean entry_modified_p(...)`), that is used to check if `handle_modified_entry` should be called or not.
- Or we could say that `handle_modified_entry` is called unconditionally (when specified), and users responsibility to provide a callback that check if the modification and necessary, and early return otherwise. In that case however, updating mtime in the entry would be the user's responsibility. That means `CollectionComparatorEntry` should be passed instead of `const gchar *`. With that it would be elegant to store `gpointer modified_data` instead of `mtime modified`, but I can live without that. Still it would be nice to unify the api, that `handle_new_entry` etc signatures should be updated. With that said, the other option would be better.

For our purpose however, as only a single file modification is enough, that's why I proposed my alternative above, that the directory monitor callback could deal with the update, which would eliminate many complexities from above. And I am not sure mtime tracking would be used by other features in the future apart from auto reload.

About goal shifting, I would like to leave that decision to you. My request is to address specifying mtime in `..._add` problem somehow.

lib/directory-monitor-poll.c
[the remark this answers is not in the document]
madsalama: I remember now. I think the problem was that the existing code only had 'filename' in each iteration, but ``g_stat`` needs a path (or a relative path) - which would be different per platform (windows/*nix) and thus, current working directory path must actually be built.

Thus, the code modification below gave errors like: `` [2019-01-08T05:54:31.725568] Cannot stat file/directory; filename='syslog-ng.conf' `` 
```
  if (directory)
    {
      const gchar *filename;
      while((filename = g_dir_read_name(directory)))
        {
          if (g_stat(filename, &stat_result))
            {
              msg_debug("Cannot stat file/directory",
                        evt_tag_str("filename", filename));
            }
          else
            {
              collection_comparator_add_value(self->comparator, filename, stat_result.st_mtime);
            }
        }
      g_dir_close(directory);
      collection_comparator_stop(self->comparator);

```

Kokan:
Hello @madsalama,

There was a long discussion about this feature request. The feature itself is questionable (creating an issue is on me).

Thanks to you we now see exactly what work has to be done in order to achieve this functionality.

The consensus is that configuration auto reload does not bring enough value on the table as much potential issue could come up, and makes - at least the file source - more volatile, and the more complex.

As of the result, I am regret to inform you that we cannot merge this pull request.

On the other hand we value your contribution and effort you made, and we would like to reward you.

Again big thanks for your work.

PR #2378, opened 2018-10-28
Block location tracking fixes

bazsi:
This branch fixes error reporting in the case of embedded newlines within strings within blocks. It also adds a number of unit tests to make sure it won't break again.

One other change is that is included in this branch and is somewhat independent is one where we highlight the entire, multi-line error block and not just the first line when displaying an error.


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

PR #2379, opened 2018-10-29
Udp reuseport

bazsi:


kira-syslogng:
Build SUCCESS

modules/afsocket/socket-options.c
```diff
@@ -49,6 +49,20 @@ socket_options_setup_socket_method(SocketOptions *self, gint fd, GSockAddr *bind
...
+            {
+              msg_error("The kernel refused our SO_REUSEPORT setting, which should be supported by Linux 3.9+",
+                        evt_tag_error("error"));
+              return FALSE;
+            }
+#else
+          msg_error("You set so-reuseport() to 1, but your platform does not support SO_REUSEPORT, which is supported on Linux 3.9+");
+          return FALSE;
+#endif
+        }
```
MrAnno: This could be extracted to a separate method, `socket_options_set_reuseport()` for example.

modules/afsocket/socket-options.c
```diff
@@ -49,6 +49,20 @@ socket_options_setup_socket_method(SocketOptions *self, gint fd, GSockAddr *bind
...
+        {
+#ifdef SO_REUSEPORT
+          if (setsockopt(fd, SOL_SOCKET, SO_REUSEPORT, &self->so_reuseport, sizeof(self->so_reuseport)) < 0)
+            {
+              msg_error("The kernel refused our SO_REUSEPORT setting, which should be supported by Linux 3.9+",
+                        evt_tag_error("error"));
+              return FALSE;
+            }
+#else
+          msg_error("You set so-reuseport() to 1, but your platform does not support SO_REUSEPORT, which is supported on Linux 3.9+");
```
MrAnno: `so-reuseport()` is a yes/no option. Should not we rephrase this error message? For example: "You set  so-reuseport() to yes ..." or "You enabled so-reuseport(), but ..."

furiel:
@bazsi can you update the description with an explanation on the feature, and maybe an example configuration? Anything that would worth mentioning in the documentation.

Is there any way we can test it?

modules/afsocket/socket-options.c
bazsi: I extracted all funcions.

modules/afsocket/socket-options.c
bazsi: I clarified the error message, thanks for pointing that out.

bazsi:
@furiel I have added a configuration sample, a more detailed description and a pointer to further
information into the commit message.

kira-syslogng:
Build SUCCESS

modules/afsocket/socket-options.c
```diff
@@ -49,6 +49,20 @@ socket_options_setup_socket_method(SocketOptions *self, gint fd, GSockAddr *bind
                           evt_tag_int("so_rcvbuf_set", so_rcvbuf_set));
             }
         }
+      if (self->so_reuseport)
+        {
+#ifdef SO_REUSEPORT
+          if (setsockopt(fd, SOL_SOCKET, SO_REUSEPORT, &self->so_reuseport, sizeof(self->so_reuseport)) < 0)
```
szemere: Note for the future (since this will be tricky to find otherwise): SO_REUSEPORT does work on WIN32 platforms, but has a very different meaning. Currently it is out of scope.

modules/afsocket/socket-options.c
MrAnno: @bazsi Thank you!

PR #2381, opened 2018-10-29
DBLD: Add new target "list-builder-images"

mitzkia:
- this target is useful when we want to query actual
 builder images from CI system
- introduced new BUILDER_IMAGES variable to store
OS images
- extend IMAGES variable with new BUILDER_IMAGES to
not broke old behavior

Signed-off-by: Micek <mitzkia@gmail.com>


kira-syslogng:
Build SUCCESS

PR #2383, opened 2018-10-30
Leak redis connection

Kokan:
* The redis destination leaked the connection at every reconnect and/or reload.
* Removed the `reconnect` arg, that did not prevent reconnection (it is better to check if the connection is alive)


Kokan:
@HanouaJ could you also test with this version ?

PR #2384, opened 2018-10-30
[wip]Fix http flush

lbudai:


kira-syslogng:
Build SUCCESS

lbudai:
other possible solution: when the queue is empty but we have pending elements in the batch, we call flush().

lib/logmsg/logmsg.c
```diff
@@ -1863,6 +1864,14 @@ gssize log_msg_get_size(LogMessage *self)
     nv_table_get_memory_consumption(self->payload); // msg.payload (nvtable)
 }
 
+gboolean log_msg_is_source_suspended(LogMessage *self)
+{
+  if (!self->ack_record)
+    return FALSE;
+  LogSource *src = self->ack_record->tracker->source;
+  return !log_source_free_to_send(src);
+}
```
bazsi: I dont like this. The threaded destination driver should cover this case by doing the flush when the queue is empty. Why doesnt that resolve this situation?

bazsi: Ah the reason i don't like this is that LogMessage should not care less about is source being suspended. It should just be a vehicle for transferring messages.

lbudai: @bazsi: I don’t like it either. this patch just wanted to highlight the fact: iw-size and flush-lines/bytes are not independent.

My other suggestion could resolve the issue(calling flush when queue becomes empty).

bazsi:
As we've discussed IRL, neither of the proposed solution is a good one.

1) flushing when the queue is empty is exactly what I wanted to avoid with flush-timeout(). With some destinations a flush operation has a lot of overhead, and syslog-ng is too "fast" depleting its queue, so in reality we always post 1-5 messages instead of a full batch, that causes the end system to be slower.

2) this solution is not good, as it'd cause pretty insidious coupling between the source and the destination side through a structure that is meant to represent messages and not flow control state.

The solution we've discussed should be better is to enumerate the sources that feed a specific destination and based on its log-iw-size() settings:

1) warn about the potential configuration issue
2) maybe even size the various parameters automatically.

2) would be more preferred, as this is a pretty common problem. I'll check if I can draft something up to resolve the situation.

lbudai:
@bazsi: currently the flush-timeout is inherited from the global opts, so I think as a first step we should break this inheritance, and disable flush-timeout by default (in this case, the flush would be called when queue becomes empty).

PR #2389, opened 2018-11-02
AFSocket: Make it possible to set the interface for socket.

juhaszviktor:
There are some use cases, when the ip address for the interface isn't known,
or there are more ip addresses for one interface

Also, there are some HA architecture, where the master and standby is on the physical network,
The active's ip address is predefined, but when syslog-ng starts it doesn't know that it runs on the master or on the standby.

This patch helps to define the interface instead of ip addresses in these cases

Usage example: s_network{ tcp(port(6666) interface("wlp3s0")) };

Signed-off-by: Juhasz Viktor <juhasz.viktor81@gmail.com>

### DIFF
--- a/modules/afsocket/socket-options-inet.c
+++ b/modules/afsocket/socket-options-inet.c
@@ -23,6 +23,7 @@
 #include "socket-options-inet.h"
 #include "gsockaddr.h"
 #include "messages.h"
+#include "gprocess.h"
 
 #include <string.h>
 #include <netinet/tcp.h>
@@ -37,6 +38,22 @@
 
 #ifndef SOL_TCP
 #define SOL_TCP IPPROTO_TCP
+#endif
+
+#ifdef SO_BINDTODEVICE
+static int
+socket_options_inet_set_interface(SocketOptionsInet *self, gint fd)
+{
+  cap_t saved_caps;
+  int result;
+
+  saved_caps = g_process_cap_save();
+  g_process_enable_cap("cap_net_raw");
+  result = setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, self->interface_name,
+                      strlen(self->interface_name)) < 0 ? errno : 0;
+  g_process_cap_restore(saved_caps);
+  return result;
+}
 #endif
 
 static gboolean
@@ -73,6 +90,21 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
       setsockopt(fd, SOL_TCP, TCP_KEEPINTVL, &self->tcp_keepalive_intvl, sizeof(self->tcp_keepalive_intvl));
 #else
       msg_error("tcp-keepalive-intvl() is set but no TCP_KEEPINTVL setsockopt on this platform");
+      return FALSE;
+#endif
+    }
+
+  if (self->interface_name)
+    {
+#ifdef SO_BINDTODEVICE
+      int result = socket_options_inet_set_interface(self, fd);
+      if (result != 0)
+        {
+          msg_error("Can't bind to interface", evt_tag_str("interface", self->interface_name), evt_tag_errno("error", result));
+          return FALSE;
+        }
+#else
+      msg_error("interfacel() is set but no SO_BINDTODEVICE setsockopt on this platform");
       return FALSE;
 #endif
     }
@@ -166,6 +198,26 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
   return TRUE;
 }
 
+void
+socket_options_inet_set_interface_name(SocketOptionsInet *self, const gchar *interface_name)
+{
+  if (self->interface_name)
+    {
+      g_free(self->interface_name);
+    }
+  self->interface_name = g_strdup(interface_name);
+}
+
+void
+socket_options_inet_free(gpointer s)
+{
+  SocketOptionsInet *self = (SocketOptionsInet *) s;
+  if (self->interface_name)
+    {
+      g_free(self->interface_name);
+    }
+}
+
 SocketOptionsInet *
 socket_options_inet_new_instance(void)
 {
@@ -179,6 +231,7 @@ socket_options_inet_new_instance(void)
   self->tcp_keepalive_intvl = 10;
   self->tcp_keepalive_probes = 6;
 #endif
+  self->super.free = socket_options_inet_free;
   return self;
 }
 
--- a/modules/afsocket/socket-options-inet.c
+++ b/modules/afsocket/socket-options-inet.c
@@ -45,14 +45,33 @@ static int
 socket_options_inet_set_interface(SocketOptionsInet *self, gint fd)
 {
   cap_t saved_caps;
-  int result;
+  int result, rc;
 
   saved_caps = g_process_cap_save();
   g_process_enable_cap("cap_net_raw");
-  result = setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, self->interface_name,
-                      strlen(self->interface_name)) < 0 ? errno : 0;
+  rc = setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, self->interface_name, strlen(self->interface_name));
+  result = (rc < 0) ? errno : 0;
   g_process_cap_restore(saved_caps);
   return result;
+}
+
+static gboolean
+socket_options_inet_try_to_set_interface(SocketOptionsInet *self, gint fd)
+{
+  int result = socket_options_inet_set_interface(self, fd);
+  if (result != 0)
+    {
+      msg_error("Can't bind to interface", evt_tag_str("interface", self->interface_name), evt_tag_errno("error", result));
+      return FALSE;
+    }
+  return TRUE;
+}
+#else
+static gboolean
+socket_options_inet_try_to_set_interface(SocketOptionsInet *self, gint fd)
+{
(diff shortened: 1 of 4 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

modules/afsocket/socket-options-inet.c
```diff
@@ -166,6 +198,26 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
   return TRUE;
 }
 
+void
+socket_options_inet_set_interface_name(SocketOptionsInet *self, const gchar *interface_name)
+{
+  if (self->interface_name)
```
Kokan: Optional: no need to check if it is a null pointer.

modules/afsocket/socket-options-inet.h
```diff
@@ -35,8 +35,11 @@ typedef struct _SocketOptionsInet
   gint tcp_keepalive_time;
   gint tcp_keepalive_intvl;
   gint tcp_keepalive_probes;
+  char *interface_name;
```
Kokan: Please use `gchar` as that type is used in the setter function.

kira-syslogng:
Build SUCCESS

modules/afsocket/socket-options-inet.c
```diff
@@ -166,6 +198,26 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
...
+    {
+      g_free(self->interface_name);
+    }
+  self->interface_name = g_strdup(interface_name);
+}
+
+void
+socket_options_inet_free(gpointer s)
+{
+  SocketOptionsInet *self = (SocketOptionsInet *) s;
```
Kokan: You should free `self` otherwise it is leaked.

modules/afsocket/socket-options-inet.c
```diff
@@ -77,6 +94,21 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
...
+    {
+#ifdef SO_BINDTODEVICE
+      int result = socket_options_inet_set_interface(self, fd);
+      if (result != 0)
+        {
+          msg_error("Can't bind to interface", evt_tag_str("interface", self->interface_name), evt_tag_errno("error", result));
+          return FALSE;
+        }
+#else
+      msg_error("interfacel() is set but no SO_BINDTODEVICE setsockopt on this platform");
```
Kokan:
```suggestion
      msg_error("interface() is set but the SO_BINDTODEVICE option not supported on this platform");
```

modules/afsocket/socket-options-inet.c
```diff
@@ -39,6 +40,22 @@
...
+#ifdef SO_BINDTODEVICE
+static int
+socket_options_inet_set_interface(SocketOptionsInet *self, gint fd)
+{
+  cap_t saved_caps;
+  int result;
+
+  saved_caps = g_process_cap_save();
+  g_process_enable_cap("cap_net_raw");
+  result = setsockopt(fd, SOL_SOCKET, SO_BINDTODEVICE, self->interface_name,
```
Kokan: Optional:
This is kinda hard to read for me with the `?:` operator in the end.
Would you consider chaning like:
```c
result = setsockopt(...);
result = result < 0 ? errno : 0;
...
return result;
```
well maybe this suggestion is not the best, but I would still not *hide* the `?:` operator.

modules/afsocket/socket-options-inet.c
```diff
@@ -166,6 +198,26 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
...
+
+void
+socket_options_inet_free(gpointer s)
+{
+  SocketOptionsInet *self = (SocketOptionsInet *) s;
+  if (self->interface_name)
+    {
+      g_free(self->interface_name);
+    }
+}
```
MrAnno: `g_free(s)` is missing from here. There is no `socket_options_free_method()`, so I guess we have to add it or just use `g_free(s)` directly.

modules/afsocket/socket-options-inet.c
MrAnno: Typo: interfacel

modules/afsocket/socket-options-inet.h
MrAnno: int, gint; char, gchar. They are the same 😄, but glib types are preferred.

modules/afsocket/socket-options-inet.c
```diff
@@ -77,6 +94,21 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
...
+      int result = socket_options_inet_set_interface(self, fd);
+      if (result != 0)
+        {
+          msg_error("Can't bind to interface", evt_tag_str("interface", self->interface_name), evt_tag_errno("error", result));
+          return FALSE;
+        }
+#else
+      msg_error("interfacel() is set but no SO_BINDTODEVICE setsockopt on this platform");
+      return FALSE;
+#endif
```
MrAnno: Optional review note:

This function is already huge. Could you extract this code with those ifdefs, for example:

```
if (self->interface_name && !socket_options_inet_try_bind_to_interface(self, fd))
  return FALSE;
```

or

```
if (self->interface_name)
{
  if (!socket_options_inet_try_bind_to_interface(self, fd))
    return FALSE;
}
```

furiel:
I tried:
```
    source { network(interface("notexist") port(5555)); }
```

I received a bind error
```
[2018-11-03T13:33:20.886431] Can't bind to interface; interface='notexist', error='Operation not permitted (1)'
```
which is great. Other than that, syslog-ng started successfully, binding to 0 0.0.0.0:5555. I would expect such error results in an init failure.

modules/afsocket/socket-options-inet.c
```diff
@@ -39,6 +40,41 @@
...
+  return result;
+}
+
+static gboolean
+socket_options_inet_try_to_set_interface(SocketOptionsInet *self, gint fd)
+{
+  int result = socket_options_inet_set_interface(self, fd);
+  if (result != 0)
+    {
+      msg_error("Can't bind to interface", evt_tag_str("interface", self->interface_name), evt_tag_errno("error", result));
```
furiel: Seemingly this should signal an error, but for me syslog-ng started successfully, as if I had just omitted the interface option.
https://github.com/balabit/syslog-ng/pull/2389#issuecomment-435584668

kira-syslogng:
Build SUCCESS

furiel:
For some reason interface does not work for me :(

```
@version: 3.18

log {
  source { network(interface("wlp2s0")
               port(5555)); };
};
```

Even though I have such interface
```
$ ip addr | grep wlp2s0 
3: wlp2s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    inet 192.168.2.110/24 brd 192.168.2.255 scope global wlp2s0
```

There are no apparent debug or error logs that would help in the investigation.

```
[plugin load related logs]
[2018-11-03T14:06:17.737337] Module loaded and initialized successfully; module='syslogformat'
[2018-11-03T14:06:17.737461] Error initializing message pipeline; plugin_name='network', location='../etc/bind-interface.conf:4:12'
```

Tried with both tcp and network driver.

@juhaszviktor do you happen to have any hint what I am I doing wrong?

modules/afsocket/socket-options-inet.c
juhaszviktor: @furiel Have you tried with root or non-root user?
Because it requires to cap_net_raw, and without superuser, I don't sure, that you can get it.

juhaszviktor:
@furiel Have you tried with root or non-root user?
Because it requires to cap_net_raw, and without superuser, I don't sure, that you can get it.

modules/afsocket/socket-options-inet.c
furiel: I tried this one with non-root user, so I expected it to fail. The issue here is that syslog-ng starts after the fail detection. When this option is not applied, it can cause connectivity issue, which can go semi-silent (we have an apparent error log, but in similar cases like when binding for an already bound port, syslog-ng does not start at all).

modules/afsocket/socket-options-inet.c
```diff
@@ -77,6 +113,12 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
 #endif
     }
 
+  if (self->interface_name)
+    {
+      if (socket_options_inet_try_to_set_interface(self, fd))
```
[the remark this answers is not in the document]
MrAnno: After fixing it, it does work perfectly:

```
LISTEN       0             128                    0.0.0.0%wlp2s0:4444                       0.0.0.0:*
```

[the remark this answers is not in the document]
juhaszviktor: Ahh, yes! It was my bad! Thanks @MrAnno :)
@furiel I've updated the correction patch.

kira-syslogng:
Build SUCCESS

Kokan:
Note: 
1) This works for source and destination as well.
2) Also works when ip specified, but it is possible to configure an ip and interface that won't match, causing no traffic in/out; and there is no message about the issue. (this could be an improvement later on).

PR #2391, opened 2018-11-05
hdfs: do not try to write unopened file

furiel:
In some cases file open in hdfs can fail. 

In hdfs, only one writer can open a file for writing exclusively. In case of append, this is ensured by a LEASE_SOFTLIMIT_PERIOD (1min) timer in the server. During that time, no other writers can open a file in hdfs.

When syslog-ng crashed and started immediately (before the 1min timer), the file opening will fail because of the exclusive lease period.
Similarly, if two syslog-ng tries to open the same file simultaneously, the file opening will fail again.

In the current code the file entry in the hashtable is created but as unopened, but syslog-ng still tries to write into the unopened file. Hence we have an exception:

```
[2018-11-05T12:29:25.836982] java.io.IOException: File is not open: hdfs://[... file path]
	at org.syslog_ng.hdfs.HdfsFile.write(HdfsFile.java:60)
	at org.syslog_ng.hdfs.HdfsDestination.send(HdfsDestination.java:256)
	at org.syslog_ng.StructuredLogDestination.sendProxy(StructuredLogDestination.java:35)
```

This patch checks if the file is successfully opened before writing.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

pzoleex:
Regression test passed

PR #2386, opened 2018-10-31
Flush lines cleanup

lbudai:
* make global `flush_timeout` deprecated
* in http rename
  * `flush_bytes` to `batch_bytes`
  * `flush_lines` to `batch_lines`
  * `flush_timeout` to `batch_timeout`

PR #2392, opened 2018-11-05
Afsql flush lines cleanup

lbudai:
depends on #2386


kira-syslogng:
Build FAILURE

lib/cfg-parser.c
```diff
@@ -92,7 +92,7 @@ static CfgLexerKeyword main_keywords[] =
   { "stats_max_dynamics", KW_STATS_MAX_DYNAMIC },
   { "min_iw_size_per_reader", KW_MIN_IW_SIZE_PER_READER },
   { "flush_lines",        KW_FLUSH_LINES },
-  { "flush_timeout",      KW_FLUSH_TIMEOUT },
+  { "flush_timeout",      KW_FLUSH_TIMEOUT, KWS_OBSOLETE, "flush_timeout() has been deprecated, remove from your configuration." },
   { "suppress",           KW_SUPPRESS },
```
bazsi: I think we should add this to the description above (as it is displayed to the user).

"Some drivers support batch-timeout() instead that you can specify at the destination level."

modules/afsql/afsql.c
```diff
@@ -859,10 +845,19 @@ afsql_dd_should_begin_new_transaction(const AFSqlDestDriver *self)
...
 }
 
+static gint
+_batch_size(const AFSqlDestDriver *self)
+{
+  if (self->super.batch_lines <= 0)
+    return DEFAULT_SQL_TX_SIZE;
+
+  return self->super.batch_lines;
+}
```
bazsi: why don't we initialize batch_lines to DEFAULT_SQL_TX_SIZE? Also, I think by default the sql driver wasn't batching as it was also depending on the "explicit-transactions" flag (the name might be somewhat different).

modules/http/http-parser.c
```diff
@@ -48,7 +48,10 @@ static CfgLexerKeyword http_keywords[] =
   { "accept_redirects", KW_ACCEPT_REDIRECTS },
   { "timeout",          KW_TIMEOUT },
   { "tls",              KW_TLS },
-  { "flush_bytes",      KW_FLUSH_BYTES },
+  { "flush_bytes",      KW_BATCH_BYTES, KWS_OBSOLETE, "The flush-bytes option is deprecated. Use batch-bytes instead." },
+  { "batch_bytes",      KW_BATCH_BYTES },
+  { "flush_lines",      KW_BATCH_LINES, KWS_OBSOLETE, "The flush-lines option is deprecated. Use batch-lines instead."},
+  { "flush_timeout",    KW_BATCH_TIMEOUT, KWS_OBSOLETE, "The flush-timeout option is deprecated. Use batch-timeout instead."},
   { "body_prefix",      KW_BODY_PREFIX },
```
bazsi: I think we can the batching related keywords as they are already defined in cfg-parser.c, which should work here too.

modules/riemann/riemann-parser.c
```diff
@@ -48,6 +48,7 @@ static CfgLexerKeyword riemann_keywords[] =
   { "ca_file",                  KW_CA_FILE },
   { "cert_file",                KW_CERT_FILE },
   { "key_file",                 KW_KEY_FILE },
+  { "flush_lines",              KW_BATCH_LINES, KWS_OBSOLETE, "The flush-lines option is deprecated in favour of batch-lines()" },
```
bazsi: again, we probably don't need this as the cfg-parser.c definition prevails.

kira-syslogng:
Build SUCCESS

modules/afsql/afsql-parser.c
```diff
@@ -50,8 +50,8 @@ static CfgLexerKeyword afsql_keywords[] =
   { "local_time_zone",    KW_LOCAL_TIME_ZONE },
   { "null",               KW_NULL },
   { "retry_sql_inserts",  KW_RETRIES },
-  { "flush_lines",        KW_FLUSH_LINES },
-  { "flush_timeout",      KW_FLUSH_TIMEOUT, KWS_OBSOLETE, "The flush-timeout() option for sql() destinations has been broken without deprecation since " VERSION_3_5 },
+  { "flush_lines",        KW_BATCH_LINES, KWS_OBSOLETE, "The flush-lines() option is deprecated, use batch-lines instead." },
+  { "flush_timeout",      KW_BATCH_TIMEOUT, KWS_OBSOLETE, "The flush-timeout() option for sql() destinations has been broken without deprecation since " VERSION_3_5 },
   { "flags",              KW_FLAGS },
```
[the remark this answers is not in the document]
lbudai: indeed, will fix, thanks

modules/afsql/afsql.c
lbudai: Well... it could be set, I just used a private function.
I did not change the default: whenever this function is called, the explicit-transaction flag is checked before.

modules/riemann/riemann-parser.c
lbudai: The reason why I've added this to x-parser.c is to mark it as an obsolete option and recommend a new option name.
In cfg-parser it is not obsolete (and don't want to make it obsolete currently, and even if it would be obsolete, the recommended option name would be different one... so I think we need this change).

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

gaborznagy:
@lbudai can you do a rebase, please as #2386 merged?
I think the PR will be just 1 commit.

lbudai:
@gaborznagy : rebased

kira-syslogng:
Build SUCCESS

modules/afsql/afsql.c
gaborznagy: @bazsi I have found the flag: "explicit-commits". And you're right, it's another requirement for batching.
The other comment about default value placement seems to be optional for me, so I mark this as resolved.

modules/afsql/afsql.h
```diff
@@ -87,7 +87,6 @@ typedef struct _AFSqlDestDriver
   AFSqlField *fields;
   gchar *null_value;
   gint num_retries;
```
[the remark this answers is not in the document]
lbudai: thanks!

lbudai:
updated

kira-syslogng:
Build SUCCESS

PR #2398, opened 2018-11-07
DBLD: Add missing docbook-xsl packages

mitzkia:
- it installs required xsl files
 - handy when 'http://docbook.sourceforge.net' is unavailable

### DIFF
--- a/dbld/images/helpers/packages.manifest
+++ b/dbld/images/helpers/packages.manifest
@@ -125,6 +125,10 @@ make                            [centos, debian, ubuntu]
     xsltproc                    [debian, ubuntu]
     libxslt                     [centos]
 
+    # docbook
+    docbook-xsl                 [debian, ubuntu]
+    docbook-style-xsl           [centos]
+
 
 
 


MrAnno:
`--with-docbook=/usr/share/xml/docbook/stylesheet/docbook-xsl/manpages/docbook.xsl` is hardcoded in dbld/rules. Is it a valid path for all images?

lbudai:
@MrAnno : it might not work on every distrib, but for ubuntu/debian, it is a working path currently.

kira-syslogng:
Build SUCCESS

mitzkia:
@MrAnno Thanks for the finding after installing the 'docbook-style-xsl' package on centos7, docbook.xsl file will be available here: /usr/share/sgml/docbook/xsl-stylesheets-1.78.1/manpages/docbook.xsl. 
So the path will be different.

bazsi:
I don't think we need to support tarball generation in all distros. We just
need *one* official where we generate release tarballs. That should
probably be the one devshell is based on.

gaborznagy:
If that is the decision then we need to remove multi-platform tarball
generation support from dbld, right?

lbudai:
@gaborznagy : as @bazsi wrote: we need only one official release tarball...
but I would not remove this feature as sometimes (due to some magic issues around autohell tools...) someone need to generate a tarball on a different image than devshell. But this should be a best-effort thing (and I also think that if someone need this feature, that person can easily modify the things that can be modified in order to make it work...)

PR #2399, opened 2018-11-07
http: when ca-file() is not set, unset CURLOPT_CAINFO

lbudai:
Reason:
 * when ca-dir() is set but ca-file() is not, AND the default cert file
   does not exist, syslog-ng cannot establish the SSL connection
   (could be a typical problem when syslog-ng is built on a different distrib)

### DIFF
--- a/modules/http/http.c
+++ b/modules/http/http.c
@@ -106,8 +106,7 @@ _setup_static_options_in_curl(HTTPDestinationWorker *self)
   if (owner->ca_dir)
     curl_easy_setopt(self->curl, CURLOPT_CAPATH, owner->ca_dir);
 
-  if (owner->ca_file)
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

modules/http/http.c
```diff
@@ -108,6 +108,8 @@ _setup_static_options_in_curl(HTTPDestinationWorker *self)
 
   if (owner->ca_file)
```
Kokan: I guess there is no need for the if, simple: `curl_easy_setopt(self->curl, CURLOPT_CAINFO, owner->ca_file);` would do the same.

lbudai: indeed! will change it.
(background: originally, I set an empty string instead of `NULL`, but it did not work...  as we need to set an explicit `NULL`, the `else` is superfluous)

kira-syslogng:
Build SUCCESS

PR #2401, opened 2018-11-09
Remove template hex, oct number parse in configuration

### LINKED ISSUE #2400: 3.16 sometimes says "NaN" for $(/ ${C_MIN} 6)
# syslog-ng
## Version of syslog-ng
```
syslog-ng 3 (3.16.1)
Config version: 3.16
Installer-Version: 3.16.1
Revision:
Compile-Date: Jun 21 2018 16:34:35
Module-Directory: //usr/lib64/syslog-ng
Module-Path: //usr/lib64/syslog-ng
Available-Modules: cryptofuncs,kvformat,afstomp,confgen,tfgetent,afprog,linux-kmsg-format,hook-commands,snmptrapd-parser,afsocket,cef,graphite,affile,http,map-value-pairs,system-source,dbparser,pseudofile,syslogformat,afuser,date,stardate,disk-buffer,basicfuncs,csvparser,add-contextual-data,geoip2-plugin,json-plugin,appmodel,tags-parser,xml
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: off
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: off
```

## Platform
* CentOS-6.7
* OEL-7.x

# Issue
*  Failure: 3.16 sometimes says "NaN" for $(/ ${C_MIN} 6)
* 3.8.1-2 and older did not do that

## Steps to reproduce
* Write `template("'${C_MIN}','$(/ ${C_MIN} 6)','...` to a test file someplace
* Observe that if `$C_MIN` == '08' or '09' *and* we're doing `$(/ ${C_MIN} 6)` we get "NaN"
* Note using just `${C_MIN}` works as expected
* Details below

## Configuration
```
$ cat /etc/syslog-ng/syslog-ng.conf
@version: 3.16
# syslog-ng.conf--Configuration file for syslog-ng on a v3 Sentry
...
options {
    # [ ] = possible options, >< = default option
    create_dirs (no);     # [yes or >no<] Enable or disable directory creation for destination files
    keep_hostname (yes);  # [yes or >no<] Enable or disable hostname rewriting macros
                          # "yes" REQUIRED for $HOST
    chain_hostnames(no);  # [yes or >no<] Enable or disable the chained hostname format (was long_hostnames())
    log_fifo_size (1000); # [number >= 1000] The number of lines fitting to the output queue
    flush_lines(0);       # [number >100<] Specifies how many lines are flushed to a destination at a time (was called sync())
    mark-mode(none);      # (various) Keep-alive and state "mark" messages
    time_reopen (10);     # [number >60<] The time to wait in seconds before a dead connection is reestablished
    threaded(yes);        # [>yes<, no] Enable syslog-ng OSE to run in multithreaded mode and use multiple CPUs
    use_dns (no);         # [>yes<, no, persist_only] Enable or disable DNS usage
    use_fqdn (no);        # [>yes<, no] Add Fully Qualified Domain Name instead of short hostname
};

####### sources
# Broken out so Syslog-NG "stats" are more granular and useful
source s_syslogng {
    internal(); # syslog-ng itself
};
source s_local {
    unix-stream("/dev/log");
    file("/proc/kmsg" program_override("kernel: ") );
};
source s_remote_tcp {
    network( ip(0.0.0.0) port(514) transport("tcp") max-connections(256) log-iw-size(25600) );
};
source s_remote_udp {
    network( ip(0.0.0.0) port(514) transport("udp") );
};
...

####### filters
...

####### rewrites
...

####### destinations
destination d_file4arc { file(
    ".../${LOGHOST}_${C_YEAR}-${C_MONTH}-${C_DAY}-${C_HOUR}_$(/ ${C_MIN} 6).events"
    template("'${LOGHOST}','${HOST}','<${TAG}>${S_DATE} ${HOST} ${MSGHDR}${MSG}'\n") template-escape(no)
    owner("user")
    group("user")
    perm(0644)
    create-dirs(yes)
    flags("no-multi-line")
); };
...

####### logging (ORDER, both lines and in-line, MATTERS)
# Extract and deal with "internal" SNMP first (and do not write SNMP to alllog)
...
log { source(s_remote_tcp);     destination(d_file4arc); };
log { source(s_remote_udp);     destination(d_file4arc); };
...

# vim:ft=syslog-ng:ai:si:ts=4:sw=4:et:
```

## Details:

I have a `"${LOGHOST}_${C_YEAR}-${C_MONTH}-${C_DAY}-${C_HOUR}_$(/ ${C_MIN} 6).events"` template that gives me 10 files per hour "indexed" from 0-9.  This has been working for a few years, but now with an upgrade from (I know old) Syslog-NG 3.8 to 3.16 I get files like:
```
host1 | SUCCESS | rc=0 >>
-rw-r--r-- 1 user user 23853 Oct 31 00:10 /archive/2018/10/31/host1_2018-10-31-00_NaN.gz
-rw-r--r-- 1 user user 26835 Oct 31 01:10 /archive/2018/10/31/host1_2018-10-31-01_NaN.gz
-rw-r--r-- 1 user user 82145 Oct 31 02:10 /archive/2018/10/31/host1_2018-10-31-02_NaN.gz
...
-rw-r--r-- 1 user user 24779 Oct 31 21:10 /archive/2018/10/31/host1_2018-10-31-21_NaN.gz
-rw-r--r-- 1 user user 25233 Oct 31 22:10 /archive/2018/10/31/host1_2018-10-31-22_NaN.gz
-rw-r--r-- 1 user user 85825 Oct 31 23:10 /archive/2018/10/31/host1_2018-10-31-23_NaN.gz
...
host2 | SUCCESS | rc=0 >>
-rw-r--r--. 1 user user 3266 Oct 31 00:10 /archive/2018/10/31/host2_2018-10-31-00_NaN.gz
-rw-r--r--. 1 user user 1924 Oct 31 01:10 /archive/2018/10/31/host2_2018-10-31-01_NaN.gz
-rw-r--r--. 1 user user 4096 Oct 31 02:10 /archive/2018/10/31/host2_2018-10-31-02_NaN.gz
...
-rw-r--r--. 1 user user 2263 Oct 31 17:10 /archive/2018/10/31/host2_2018-10-31-17_NaN.gz
-rw-r--r--. 1 user user 2860 Oct 31 18:10 /archive/2018/10/31/host2_2018-10-31-18_NaN.gz
-rw-r--r--. 1 user user 1587 Oct 31 19:10 /archive/2018/10/31/host2_2018-10-31-19_NaN.gz
...
```

I want the "index" because reasons, and what "index" `$(/ ${C_MIN} 6)` creates is as follows:
```
Index:   0    1      2      3      4      5      6      7      8      9
Minute: 0-5  6-11  12-17  18-23  24-29  30-35  36-41  42-47  48-53  54-59
```

So to reproduce this I created:
```
destination d_file4TEST { file(
    "/tmp/${LOGHOST}_${C_YEAR}-${C_MONTH}-${C_DAY}-${C_HOUR}_$(/ ${C_MIN} 6).events"
    template("'${C_MIN}','$(/ ${C_MIN} 6)','${LOGHOST}','${HOST}','<${TAG}>${S_DATE} ${HOST} ${MSGHDR}${MSG}'\n") template-escape(no)
    owner("user")
    group("user")
    perm(0644)
    create-dirs(yes)
    flags("no-multi-line")
); };
```

And I get:
```
$ ll /tmp/*events*
-rw-r--r-- 1 user user 190K Nov  8 00:41 /tmp/host-904_2018-11-08-00_6.events
-rw-r--r-- 1 user user 3.5M Nov  8 00:47 /tmp/host-904_2018-11-08-00_7.events
-rw-r--r-- 1 user user 3.5M Nov  8 00:54 /tmp/host-904_2018-11-08-00_8.events
-rw-r--r-- 1 user user 3.5M Nov  8 00:59 /tmp/host-904_2018-11-08-00_9.events
-rw-r--r-- 1 user user 3.5M Nov  8 01:06 /tmp/host-904_2018-11-08-01_0.events
-rw-r--r-- 1 user user 1.2M Nov  8 01:09 /tmp/host-904_2018-11-08-01_NaN.events
-rw-r--r-- 1 user user 2.3M Nov  8 01:12 /tmp/host-904_2018-11-08-01_1.events
-rw-r--r-- 1 user user 3.5M Nov  8 01:17 /tmp/host-904_2018-11-08-01_2.events
-rw-r--r-- 1 user user 3.3M Nov  8 01:23 /tmp/host-904_2018-11-08-01_3.events

$ cut -d"," -f1-2 /tmp/*events* | sort -u
'00','0'
'01','0'
'02','0'
'03','0'
'04','0'
'05','0'
'06','1'
'07','1'
'08','NaN'
'09','NaN'
'10','1'
'11','1'
'12','2'
'13','2'
'14','2'
'15','2'
'16','2'
'17','2'
'18','3'
'19','3'
'20','3'
'21','3'
'22','3'
'23','3'
'24','4'
'41','6'
'42','7'
'43','7'
'44','7'
'45','7'
'46','7'
'47','7'
'48','8'
'49','8'
'50','8'
'51','8'
'52','8'
'53','8'
'54','9'
'55','9'
'56','9'
'57','9'
'58','9'
'59','9'

```

https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.16/release-notes/template-functions-of-syslog-ng-ose says:
* _Description_: These template functions allow you to manipulate numbers, that is, to perform addition ( + ), substraction ( - ), multiplication ( * ), division ( / ), and modulus ( % ). All of them require two numeric arguments. *The result is NaN (Not-a-Number) if the parameters are not numbers, cannot be parsed, or if a division by zero would occur.*
* (Also note the typo in the docs: "substraction" should be "subtraction".)

But since `${C_MIN}` (current minute) and `6` are always there and always parseable, how is `$(/ ${C_MIN} 6)` giving me "NaN"?

Kokan:
The #1679 enhanced number parsing, so in configuration a numbers formatted in hexadecimal (`0xFA`) and octal (`07`) could be also interpreted.

But sadly that also broke configuration that try to use `08` numbers like this (not valid octal, but seems like valid), and also in many cases the `010` is not expected to be actually `8`.

The following configuration was accepted by syslog-ng before that patch (3.13)
```
network(port(08));
```
But not after, it says configuration error `08` is not a number.

On the other hand the `$(binary)` function demands those above mentioned formats.

The final solution is:
* The grammar can detect hex, oct, dec; and if a number starts with `0` or `0x`, it must be in the expected format (oct, hex). The `08` is considered invalid in config.
* The template argument: the numbers in template function can only be in dec format, and therefore the `08` is a valid `8` value.

There is an exception for the template argument:
* `$(binary)` accepts hex, dec, oct numbers, and works like in the grammar. (`08` is invalid).

Fixes #2400 
Thanks @vossenjp for reporting the issue, and providing ideas about possible solution.

### DIFF
--- a/lib/parse-number.c
+++ b/lib/parse-number.c
@@ -28,6 +28,8 @@
 #include <errno.h>
 #include <stdlib.h>
 
+static const gint DETECT_BASE  = 0;
+
 static gboolean
 _valid_unit(const gchar unit_char)
 {
@@ -144,12 +146,12 @@ _process_suffix(const gchar *suffix, gint64 *d)
 }
 
 static gboolean
-_parse_number(const gchar *s, gchar **endptr, gint64 *d)
+_parse_number(const gchar *s, gchar **endptr, gint base, gint64 *d)
 {
   gint64 val;
 
   errno = 0;
-  val = strtoll(s, endptr, 0);
+  val = strtoll(s, endptr, base);
 
   if (errno == ERANGE || errno == EINVAL)
     return FALSE;
@@ -166,7 +168,7 @@ parse_number(const gchar *s, gint64 *d)
 {
   gchar *endptr;
 
-  if (!_parse_number(s, &endptr, d))
+  if (!_parse_number(s, &endptr, DETECT_BASE, d))
     return FALSE;
   if (*endptr)
     return FALSE;
@@ -178,7 +180,7 @@ parse_number_with_suffix(const gchar *s, gint64 *d)
 {
   gchar *endptr;
 
-  if (!_parse_number(s, &endptr, d))
+  if (!_parse_number(s, &endptr, DETECT_BASE, d))
     return FALSE;
   return _process_suffix(endptr, d);
 }
--- a/lib/parse-number.c
+++ b/lib/parse-number.c
@@ -28,6 +28,7 @@
 #include <errno.h>
 #include <stdlib.h>
 
+static const gint DECIMAL_BASE = 10;
 static const gint DETECT_BASE  = 0;
 
 static gboolean
@@ -171,6 +172,16 @@ parse_number(const gchar *s, gint64 *d)
   if (!_parse_number(s, &endptr, DETECT_BASE, d))
     return FALSE;
   if (*endptr)
+    return FALSE;
+  return TRUE;
+}
+
+gboolean
+parse_dec_number(const gchar *s, gint64 *d)
+{
+  gchar *endptr;
+
+  if (!_parse_number(s, &endptr, DECIMAL_BASE, d))
     return FALSE;
   return TRUE;
 }
--- a/lib/parse-number.c
+++ b/lib/parse-number.c
@@ -147,12 +147,6 @@ _process_suffix(const gchar *suffix, gint64 *d)
 }
 
 static gboolean
-_no_suffix(const gchar *suffix)
-{
-  return (*suffix == '\0' || g_ascii_isspace(*suffix));
-}
-
-static gboolean
 _parse_number(const gchar *s, gchar **endptr, gint base, gint64 *d)
 {
   gint64 val;
@@ -179,7 +173,7 @@ parse_number(const gchar *s, gint64 *d)
     return FALSE;
   if (*endptr)
     return FALSE;
-  return _no_suffix(endptr);
+  return TRUE;
 }
 
 gboolean
@@ -189,7 +183,9 @@ parse_dec_number(const gchar *s, gint64 *d)
 
   if (!_parse_number(s, &endptr, DECIMAL_BASE, d))
     return FALSE;
-  return _no_suffix(endptr);
+  if (*endptr)
+    return FALSE;
+  return TRUE;
 }
 
 gboolean
--- a/lib/parse-number.c
+++ b/lib/parse-number.c
@@ -147,6 +147,12 @@ _process_suffix(const gchar *suffix, gint64 *d)
 }
 
 static gboolean
+_no_suffix(const gchar *suffix)
+{
+  return (*suffix == '\0' || g_ascii_isspace(*suffix));
+}
+
+static gboolean
 _parse_number(const gchar *s, gchar **endptr, gint base, gint64 *d)
 {
   gint64 val;
@@ -173,7 +179,7 @@ parse_number(const gchar *s, gint64 *d)
     return FALSE;
   if (*endptr)
     return FALSE;
-  return TRUE;
+  return _no_suffix(endptr);
 }
 
 gboolean
@@ -183,7 +189,7 @@ parse_dec_number(const gchar *s, gint64 *d)
 
   if (!_parse_number(s, &endptr, DECIMAL_BASE, d))
     return FALSE;
-  return TRUE;
+  return _no_suffix(endptr);
 }
 
 gboolean
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -233,8 +233,20 @@ word	[^ \#'"\(\)\{\}\[\]\\;\r\n\t,|\.@:]
 \.\.                       { return LL_DOTDOT; }
 \.\.\.                     { return LL_DOTDOTDOT; }
 (-|\+)?{digit}+\.{digit}+  { yylval->fnum = strtod(yytext, NULL); return LL_FLOAT; }
-0x{xdigit}+ 		   { yylval->num = strtoll(yytext + 2, NULL, 16); return LL_NUMBER; }
-0{odigit}+		   { yylval->num = strtoll(yytext + 1, NULL, 8); return LL_NUMBER; }
+0x{xdigit}+ 		   {
+                             if (!parse_number_with_suffix(yytext, &yylval->num))
+                               {
+                                 return LL_ERROR;
+                               }
+                             return LL_NUMBER;
+                           }
+0{odigit}+		   {
+                             if (!parse_number_with_suffix(yytext, &yylval->num))
+                               {
+                                 return LL_ERROR;
+                               }
+                             return LL_NUMBER;
+                           }
 (-|\+)?{digit}+(M|m|G|g|k|K)?(i|I)?(b|B)? {
                              if (!parse_number_with_suffix(yytext, &yylval->num))
                                {
--- a/lib/parse-number.h
+++ b/lib/parse-number.h
@@ -29,5 +29,6 @@
 
 gboolean parse_number_with_suffix(const gchar *str, gint64 *result);
 gboolean parse_number(const gchar *str, gint64 *result);
+gboolean parse_dec_number(const gchar *str, gint64 *result);
 
 #endif
--- a/lib/tests/test_parse_number.c
+++ b/lib/tests/test_parse_number.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2013-2018 Balabit
  * Copyright (c) 2013 Gergely Nagy <algernon@balabit.hu>
+ * Copyright (c) 2018 Kokan <kokaipeter@gmail.com>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -81,8 +82,16 @@ Test(parse_number, test_simple_numbers_are_parsed_properly)
 Test(parse_number,test_c_like_prefixes_select_base)
 {
   assert_parse("0x20", 32);
+  assert_parse("0xFF", 255);
+  assert_parse("-0x09", -9);
+
   assert_parse("020", 16);
+  assert_parse("-010", -8);
+  assert_parse_fails("08");
+  assert_parse_fails("0A");
+
   assert_parse("20", 20);
+  assert_parse_fails("FF");
 }
 
 Test(parse_number_with_suffix, test_simple_numbers_are_parsed_properly)
@@ -95,8 +104,16 @@ Test(parse_number_with_suffix, test_simple_numbers_are_parsed_properly)
 Test(parse_number_with_suffix, test_c_like_prefixes_select_base)
 {
   assert_parse_with_suffix("0x20", 32);
+  assert_parse_with_suffix("0xFF", 255);
+  assert_parse_with_suffix("-0x09", -9);
+
   assert_parse_with_suffix("020", 16);
+  assert_parse_with_suffix("-010", -8);
+  assert_parse_with_suffix_fails("08");
+  assert_parse_with_suffix_fails("0A");
+
   assert_parse_with_suffix("20", 20);
+  assert_parse_with_suffix_fails("FF");
 }
 
 Test(parse_number_with_suffix, test_exponent_suffix_is_parsed_properly)
--- a/lib/tests/test_parse_number.c
+++ b/lib/tests/test_parse_number.c
@@ -72,6 +72,29 @@ assert_parse_fails(const gchar *str)
   cr_assert_eq(res, FALSE, "Parsing (w/o suffix) %s succeeded, while expecting failure", str);
 }
 
+static void
+assert_parse_dec(const gchar *str, gint64 expected)
+{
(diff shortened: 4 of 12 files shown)


furiel:
Just two thoughts:

I am afraid the same problem still persists inside binary template function. $(binary 09) would still be a NaN. Type hinting sounds good, or we could just leave octal numbers out of the game: first we try to detect as hexadecimal (i.e. look for 0x), then fallback to decimal.

Can you add some test that avoid this kind of regression later on?
Test(parse_number, test_simple_numbers_are_parsed_properly) looks like a good candidate in test_parse_number.c
```
+  assert_parse("09", 9);
```

bazsi:
@furiel I think the case with $(binary) is less problematic, as we would get an error message during configuration parsing, $(binary) processes its arguments in the prepare() phase for template functions, e.g. it does not support resolving macros.

Also, we could make the error message in $(binary) a bit more helpful, in the case of errors and explicitly mention that 0 indicates octal.

What about this patch:

```
diff --git a/modules/basicfuncs/str-funcs.c b/modules/basicfuncs/str-funcs.c
index 9a04ee7e4..319dfa550 100644
--- a/modules/basicfuncs/str-funcs.c
+++ b/modules/basicfuncs/str-funcs.c
@@ -529,9 +529,14 @@ tf_binary_prepare(LogTemplateFunction *self, gpointer s, LogTemplate *parent, gi
       gchar *token = argv[i];
       if (!parse_number(token, &number))
         {
-          g_set_error(error, LOG_TEMPLATE_ERROR, LOG_TEMPLATE_ERROR_COMPILE,
-                      "$(binary) template function requires list of dec/hex/oct numbers as arguments, unable to parse %s as a number",
-                      token);
+          if (token[0] == '0')
+            g_set_error(error, LOG_TEMPLATE_ERROR, LOG_TEMPLATE_ERROR_COMPILE,
+                        "$(binary) argument starts with '0' that indicates an octal number, unable to parse %s as an octal number",
+                        token);
+          else
+            g_set_error(error, LOG_TEMPLATE_ERROR, LOG_TEMPLATE_ERROR_COMPILE,
+                        "$(binary) template function requires list of dec/hex/oct numbers as arguments, unable to parse %s as a number",
+                        token);
           goto error;
         }
       if (number > 0xFF)
```

furiel:
@bazsi that sounds good. I was thinking it is valid to use template/macro as a parameter for `binary`, for example `$(binary $MSG)`, that's why I was proposing the runtime check. But we do the validation in `prepare` and not in `call`, so binary template function can work only with constants.

Kokan:
@bazsi 
"Also, I think the lexer should still use the octal detection stuff, the problematic piece comes when we are expanding macros and parsing is applied to those. Octal comes handy when specifying permission bits."

Actually the lexer itself handles the hex, octal format. So this makes me more sense, but its strange while the network port option did not used that (it is `string_or_number`), so it is passed as string from grammar point of view, and later on converted :)

I'll check a few things like above, so I've mark it WIP till that time.

bazsi:
@Kokan the reason it is "string_or_number" is that you can use a service name in place of a port number there and that's resolved in init time.

bazsi:
I have now reviewed the patch and it looks good to me. The only thing we've agreed on is to merge the rules in cfg-lex.l so we use the same auto-detection logic both in the grammar and later in template functions. For that something like the following untested patch is needed:

```
diff --git a/lib/cfg-lex.l b/lib/cfg-lex.l
index 67cd7f196..562769b16 100644
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -233,9 +233,7 @@ word        [^ \#'"\(\)\{\}\[\]\\;\r\n\t,|\.@:]
 \.\.                       { return LL_DOTDOT; }
 \.\.\.                     { return LL_DOTDOTDOT; }
 (-|\+)?{digit}+\.{digit}+  { yylval->fnum = strtod(yytext, NULL); return LL_FLOAT; }
-0x{xdigit}+               { yylval->num = strtoll(yytext + 2, NULL, 16); return LL_NUMBER; }
-0{odigit}+                { yylval->num = strtoll(yytext + 1, NULL, 8); return LL_NUMBER; }
-(-|\+)?{digit}+(M|m|G|g|k|K)?(i|I)?(b|B)? {
+(-|\+)?(0x)?{digit}+(M|m|G|g|k|K)?(i|I)?(b|B)? {
                              if (!parse_number_with_suffix(yytext, &yylval->num))
                                {
                                  return LL_ERROR;

```

Kokan:
@bazsi I think in a clean way we cannot use one lexer rule to parse the number.
Your patch(`(-|\+)?(0x)?{digit}+(M|m|G|g|k|K)?(i|I)?(b|B)?`) has to be extended with `A-F`, and in that case something like this `ace` is considered as number, while that is clearly not a number. The hex number requires the context in, so the way I guess would be to make rules like `(-|\+)?(dec|hex|oct)(M|m|...|K)?...`, in witch case this is almost the same if there were three rule, but all of them calls the same function (`parse_number_with_suffix`).

kira-syslogng:
Build SUCCESS

lib/parse-number.c
```diff
@@ -146,6 +147,12 @@ _process_suffix(const gchar *suffix, gint64 *d)
   return TRUE;
 }
 
+static gboolean
+_no_suffix(const gchar *suffix)
+{
+  return (*suffix == '\0' || isspace(*suffix));
```
furiel: Did you happen to consider `g_ascii_isspace()` for this purpose?

furiel: Can you explain when the `isspace` branch is possible?

Kokan: No, but I can use it.

Kokan: The previous implementation was faulty. The parsing only checked if the `endptr` is equal to the first character in the string representation of the number, but it is possible that the `strtoll` fails on other point and not the start. For example, if `strtoll` is called with base detection and the following number is supplied: `018`.

From `strtoll` point of view, the number is `1`, and the `endptr` is `8`, but that number is clearly not valid, the previous implementation marked it as a valid number, ignoring the `8`.

I thought I had to define a number in a way, that it is followed by something. I came up with `\0` and whitespace.

kira-syslogng:
Build SUCCESS

gaborznagy:
Review deadline: Dec 6 end of day (Santa will do it).

bazsi:
I have tried to push one more patch, but at the end I couldn't:

```
Author: Balazs Scheidler <balazs.scheidler@oneidentity.com>
Date:   Tue Dec 4 16:45:11 2018 +0100

    parse-number: get rid off _no_suffix
    
diff --git a/lib/parse-number.c b/lib/parse-number.c
index 6161aa406..ee8eba4f3 100644
--- a/lib/parse-number.c
+++ b/lib/parse-number.c
@@ -146,12 +146,6 @@ _process_suffix(const gchar *suffix, gint64 *d)
   return TRUE;
 }
 
-static gboolean
-_no_suffix(const gchar *suffix)
-{
-  return (*suffix == '\0' || g_ascii_isspace(*suffix));
-}
-
 static gboolean
 _parse_number(const gchar *s, gchar **endptr, gint base, gint64 *d)
 {
@@ -179,7 +173,7 @@ parse_number(const gchar *s, gint64 *d)
     return FALSE;
   if (*endptr)
     return FALSE;
-  return _no_suffix(endptr);
+  return TRUE;
 }
 
 gboolean
@@ -189,7 +183,9 @@ parse_dec_number(const gchar *s, gint64 *d)
 
   if (!_parse_number(s, &endptr, DECIMAL_BASE, d))
     return FALSE;
-  return _no_suffix(endptr);
+  if (*endptr)
+    return FALSE;
+  return TRUE;
 }
 
 gboolean

```

With this consider this branch approved from my side. I've discussed the above with @Kokan IRL.

Kokan:
@bazsi I changed the rights to allow maintainers to update this PR, now it should work.

kira-syslogng:
Build SUCCESS

PR #2402, opened 2018-11-09
logthrdestdrv: process flush result after worker thread exits

lbudai:
The messages are ACKed by the `_process_result()` method.
UnACKed messages are move back to the queue to be sent by syslog-ng
later.

Problems the patch fixes in batching mode:
1) when a queue contains unsent items during a stop request, worker
threads are blocked until the batch timer not elapsed

2) when the queue contains items during reload (less then a `batch_lines`), then we perform a flush after the worker thread exits, but as the result was not processed, these messages are never ACKed, so they were moved back to the queue, and could be sent multiple times

### DIFF
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -639,7 +639,8 @@ _worker_thread(gpointer arg)
   _schedule_restart(self);
   iv_main();
 
-  log_threaded_dest_worker_flush(self);
+  worker_insert_result_t result = log_threaded_dest_worker_flush(self);
+  _process_result(self, result);
   log_queue_rewind_backlog_all(self->queue);
 
   _disconnect(self);


kira-syslogng:
Build SUCCESS

lib/logthrdestdrv.c
```diff
@@ -639,7 +639,8 @@ _worker_thread(gpointer arg)
   _schedule_restart(self);
   iv_main();
 
-  log_threaded_dest_worker_flush(self);
+  worker_insert_result_t result = log_threaded_dest_worker_flush(self);
+  _process_result(self, result);
```
Kokan: Using the `_process_result` has a few drawback, as it could trigger `_disconnect` or `_rewind_batch`. Although they are fine, as both of them should handle consecutive calls.

But it has also a tricky side effect about logs. The logs in the `_process_result` was not composed to use in this case.
For example, if the destination is not connected while the thread is in shutdown, this is going to print on info level the following: `Server disconnected while preparing messages for sending, trying again`. That could be confusing to see.

I did not requested change, as I think the issue it solves outweights the above small inconveniences.

pzoleex:
@kira-syslogng test this please test branch=pzolee-splunk;

kira-syslogng:
Build SUCCESS

PR #2405, opened 2018-11-10
Template function prefix coverage

bazsi:
This branch improves test coverage by always checking that the template functions properly handle prefixes, e.g. they don't change behavior if the result string already has some value.

The template functions always "append", thus the incoming result value is significant and forms part of
the output. If any template function changes output based on that, it's a bug.

With that I've identified some bugs, fixes are in the same branch. I could also remove some explicit testcases, which are now covered implicitly.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/basicfuncs/list-funcs.c
```diff
@@ -25,9 +25,9 @@
 #include "str-repr/encode.h"
 
 static void
-_append_comma(GString *result)
+_append_comma(GString *result, gsize initial_len)
 {
-  if (result->len == 0)
+  if (result->len == initial_len)
```
furiel: From clean code perspective it is hard to understand what does `initial_len` do for an `_append_comma` function. Previously, when initial len was zero, that version was also strange but I could reason with something similar: I do not need to use comma when appending to an empty string. With this change, although the logic is very similar, still it is harder to reason about it.

Maybe this condition should be moved outside the `_append_comma`?

libtest/cr_template.c
```diff
@@ -154,7 +154,7 @@ assert_template_format_with_escaping_and_context_msgs(const gchar *template, gbo
 {
   LogTemplate *templ = compile_template(template, escaping);
   const gchar *prefix = "somevoodooprefix/";
-  gsize prefix_len = strlen(prefix);
+  gint prefix_len = strlen(prefix);
```
furiel: Can you edit the commit message?

kira-syslogng:
Build SUCCESS

modules/basicfuncs/list-funcs.c
bazsi: I have renamed the function to better indicate that this is basically a private function, part of the various list functions and doing a bit more than just appending a comma. While I agree it is a bit cumbersome to follow the logic and why initial_len is important to the decision of appending a comma, but I would rather not do more than the minimal fix to the issue. The goal of this branch was to improve test coverage after all. I'd defer refactoring of the list functions to a later date.

I've also added a comment to explain the reasons for that check.

PR #2406, opened 2018-11-12
Fix non-reliable disk-queue truncating problem on load

gaborznagy:
On loading exiting disk-queues it was possible that a disk-queue only
contained saved memory queues (e.g. qout, qbacklog) and the disk-buf-size part was empty.

Due to the faulty logic of truncate on load, we could have empty disk-queues with possibly big sizes.

Also, the truncating logic assumed that qout will be the lowest offset among the saved
memory queues, however it is possible that only qbacklog is saved to the disk, which
resulted into a huge, empty disk-queue file.

### DIFF
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -388,14 +388,15 @@ _load_queue (LogQueueDisk *s, const gchar *filename)
 static gboolean
 _save_queue (LogQueueDisk *s, gboolean *persistent)
 {
+  gboolean success = FALSE;
   LogQueueDiskNonReliable *self = (LogQueueDiskNonReliable *) s;
   if (qdisk_save_state (s->qdisk, self->qout, self->qbacklog, self->qoverflow))
     {
       *persistent = TRUE;
-      qdisk_stop (s->qdisk);
-      return TRUE;
+      success = TRUE;
     }
-  return FALSE;
+  qdisk_stop (s->qdisk);
+  return success;
 }
 
 static void
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -380,7 +380,7 @@ static gboolean
 _load_queue (LogQueueDisk *s, const gchar *filename)
 {
   /* qdisk portion is not yet started when this happens */
-  g_assert(!qdisk_initialized (s->qdisk));
+  g_assert(!qdisk_started (s->qdisk));
 
   return _start(s, filename);
 }
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -402,7 +402,7 @@ static void
 _restart(LogQueueDisk *s, DiskQueueOptions *options)
 {
   LogQueueDiskNonReliable *self = (LogQueueDiskNonReliable *) s;
-  qdisk_init(self->super.qdisk, options, "SLQF");
+  qdisk_init_instance(self->super.qdisk, options, "SLQF");
 }
 
 static void
@@ -427,7 +427,7 @@ log_queue_disk_non_reliable_new(DiskQueueOptions *options, const gchar *persist_
   g_assert(options->reliable == FALSE);
   LogQueueDiskNonReliable *self = g_new0(LogQueueDiskNonReliable, 1);
   log_queue_disk_init_instance(&self->super, persist_name);
-  qdisk_init(self->super.qdisk, options, "SLQF");
+  qdisk_init_instance(self->super.qdisk, options, "SLQF");
   self->qbacklog = g_queue_new ();
   self->qout = g_queue_new ();
   self->qoverflow = g_queue_new ();
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -339,7 +339,7 @@ _push_tail (LogQueueDisk *s, LogMessage *msg, LogPathOptions *local_options, con
                          evt_tag_str  ("filename", qdisk_get_filename (self->super.qdisk)),
                          evt_tag_long ("queue_len", _get_length(s)),
                          evt_tag_int  ("mem_buf_length", self->qoverflow_size),
-                         evt_tag_long ("size", qdisk_get_size (self->super.qdisk)),
+                         evt_tag_long ("disk_buf_size", qdisk_get_size (self->super.qdisk)),
                          evt_tag_str  ("persist_name", self->super.super.persist_name));
               return FALSE;
             }
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -392,7 +392,7 @@ _save_queue (LogQueueDisk *s, gboolean *persistent)
   if (qdisk_save_state (s->qdisk, self->qout, self->qbacklog, self->qoverflow))
     {
       *persistent = TRUE;
-      qdisk_deinit (s->qdisk);
+      qdisk_stop (s->qdisk);
       return TRUE;
     }
   return FALSE;
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -557,27 +557,12 @@ qdisk_header_is_inconsistent(QDisk *self)
 static gboolean
 _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
 {
-  gint64 qout_ofs;
-  gint qout_count;
-  gint64 qbacklog_ofs;
-  gint qbacklog_count;
-  gint64 qoverflow_ofs;
-  gint qoverflow_count;
-  gint64 end_ofs = QDISK_RESERVED_SPACE;
-
   if (memcmp(self->hdr->magic, self->file_id, 4) != 0)
     {
       msg_error("Error reading disk-queue file header",
                 evt_tag_str("filename", self->filename));
       return FALSE;
     }
-
-  qout_count = self->hdr->qout_pos.count;
-  qout_ofs = self->hdr->qout_pos.ofs;
-  qbacklog_count = self->hdr->qbacklog_pos.count;
-  qbacklog_ofs = self->hdr->qbacklog_pos.ofs;
-  qoverflow_count = self->hdr->qoverflow_pos.count;
-  qoverflow_ofs = self->hdr->qoverflow_pos.ofs;
 
   if (qdisk_header_is_inconsistent(self))
     {
@@ -594,6 +579,7 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
       if (!_load_non_reliable_queues(self, qout, qbacklog, qoverflow))
         return FALSE;
 
+      gint64 end_ofs = QDISK_RESERVED_SPACE;
       if (!self->options->read_only)
         {
           qdisk_try_to_truncate_file_to_minimal(self, &end_ofs);
@@ -606,9 +592,9 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
 
       msg_info("Disk-buffer state loaded",
                evt_tag_str("filename", self->filename),
-               evt_tag_int("qout_length", qout_count),
-               evt_tag_int("qbacklog_length", qbacklog_count),
-               evt_tag_int("qoverflow_length", qoverflow_count),
+               evt_tag_int("qout_length", self->hdr->qout_pos.count),
+               evt_tag_int("qbacklog_length", self->hdr->qbacklog_pos.count),
+               evt_tag_int("qoverflow_length", self->hdr->qoverflow_pos.count),
                evt_tag_long("qdisk_length", self->hdr->length));
     }
   else
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -906,7 +906,7 @@ qdisk_init_instance(QDisk *self, DiskQueueOptions *options, const gchar *file_id
 }
 
 void
-qdisk_deinit(QDisk *self)
+qdisk_stop(QDisk *self)
 {
   if (self->filename)
     {
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -933,29 +933,6 @@ qdisk_stop(QDisk *self)
 }
 
 gssize
-qdisk_read_from_backlog(QDisk *self, gpointer buffer, gsize bytes_to_read)
-{
-  gssize res;
-  res = pread(self->fd, buffer, bytes_to_read, self->hdr->backlog_head);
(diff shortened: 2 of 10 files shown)


kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -597,15 +597,12 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
     {
       if (!_load_non_reliable_queues(self, qout, qbacklog, qoverflow))
         return FALSE;
-    }
 
-  if (!self->options->read_only)
-    {
-      end_ofs = qdisk_try_to_truncate_file_to_minimal(self);
-    }
+      if (!self->options->read_only)
```
furiel: Originally this block was called in both reliable and nonreliable case. Now only nonreliable case. Is this behavior change intended?

furiel:
It would worth mentioning that a possible fd and memory leak is fixed too -> diskq: Fix missing file close in case memory queue saving fails

modules/diskq/qdisk.c
```diff
@@ -560,18 +601,12 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
 
   if (!self->options->read_only)
     {
-      end_ofs = qout_ofs;
```
furiel: Just a note: so the bug was in this line: if qout_ofs was zero, the minimum calculation logic did nothing, and in the end syslog wouldn't truncate.

Kokan:
@Kokan review deadline tomorrow

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -0,0 +1,172 @@
...
+  configuration = cfg_new_snippet();
+  assert_true(cfg_init(configuration), "cfg_init failed!");
+  cfg_load_module(configuration, "syslogformat");
+  msg_format_options_defaults(&parse_options);
+  msg_format_options_init(&parse_options, configuration);
+
+  // File-queue is the part of disk-queue that is used when qout is full
+  test_diskq_truncate_with_diskbuffer_used();
+  test_diskq_truncate_without_diskbuffer_used();
+
```
furiel: I miss deinitialization: like cfg_free, and app_shutdown.

modules/diskq/qdisk.c
```diff
@@ -560,18 +601,12 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
...
-        end_ofs = qoverflow_ofs;
-      if(end_ofs > QDISK_RESERVED_SPACE)
-        _truncate_file(self, end_ofs);
+      end_ofs = qdisk_try_to_truncate_file_to_minimal(self);
     }
 
   if (!self->options->reliable)
     {
-      self->file_size = qout_ofs;
+      self->file_size = end_ofs;
```
furiel: Shouldn't this be the maximum of (end_ofs and QDISK_RESERVED_SPACE)?

modules/diskq/tests/Makefile.am
```diff
@@ -31,3 +32,10 @@ modules_diskq_tests_test_reliable_backlog_LDADD = $(DISKQ_TEST_LD_ADD)
 modules_diskq_tests_test_reliable_backlog_SOURCES = \
 	modules/diskq/tests/test_reliable_backlog.c \
 	modules/diskq/tests/test_diskq_tools.h
+
```
furiel: [...]_DEPENDENCIES = $(top_builddir)/modules/diskq/libdisk-buffer.la is missing. 
(Unfortunately this is missing from the old testfiles too.)

modules/diskq/tests/Makefile.am
```diff
@@ -5,6 +5,7 @@ DISKQ_TEST_LD_ADD = $(TEST_LDADD)
 modules_diskq_tests_TESTS = \
   modules/diskq/tests/test_diskq \
   modules/diskq/tests/test_diskq_full \
+  modules/diskq/tests/test_diskq_truncate \
```
furiel: Please add the tests to CMakeLists.txt too.

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -0,0 +1,172 @@
...
+#include "mainloop-call.h"
+#include "mainloop-io-worker.h"
+
+#include "queue_utils_lib.h"
+#include "test_diskq_tools.h"
+#include "testutils.h"
+
+#include <stdlib.h>
+#include <string.h>
+#include <iv.h>
```
furiel: iv.h is not necessary.

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -0,0 +1,172 @@
...
+#include "logqueue-fifo.h"
+#include "logqueue-disk.h"
+#include "logqueue-disk-reliable.h"
+#include "logqueue-disk-non-reliable.h"
+#include "diskq.h"
+#include "qdisk.c"
+#include "logpipe.h"
+#include "apphook.h"
+#include "plugin.h"
+#include "mainloop.h"
```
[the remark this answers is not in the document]
gaborznagy: okay will fix among with others.

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -0,0 +1,172 @@
...
+  GString *filename = g_string_new("test_dq_truncate.qf");
+
+  q = _create_diskqueue(filename->str, &options);
+  assert_gint(log_queue_get_length(q), 0, "Invalid queue length!");
+
+  feed_some_messages(q, params.number_of_msgs_to_push, &parse_options);
+  assert_gint(log_queue_get_length(q), params.number_of_msgs_to_push, "Invalid queue length!");
+  gint64 size_after_push = ((LogQueueDisk *)q)->qdisk->file_size;
+  struct stat after_push;
+  stat(filename->str, &after_push);
```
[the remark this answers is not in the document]
gaborznagy: I think we should not assert every step in a testcase:
- save_queue is not asserted
- unlink is not asserted either currently.

It is a good question what happens if stat failed (however I think it will not, as previous function would fail too).
I guess we would access uninitialized members inside the "struct stat".
Okay, it's not a major thing, I assert the stat() calls too.

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -0,0 +1,172 @@
...
+  {
+    .number_of_msgs_to_push = 100,
+     .number_of_msgs_to_pop = 80,
+      .should_file_be_empty_after_truncate = TRUE
+  });
+}
+
+
+int
+main(void)
```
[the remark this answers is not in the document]
gaborznagy: I plan to, also with the other dq tests, just not have that in the scope of this PR.
The queue libraries are also needs reworking, and I'm afraid I will have to do spend a much bigger effort into it than it looks now.

modules/diskq/qdisk.c
gaborznagy: Yes.
If you look the previous code it was called on reliable queues as well, but
since it started by checkout qout_ofs, which was 0 in case of reliable, we never truncated in that case.
I think it's much cleaner now.

modules/diskq/tests/test_diskq_truncate.c
gaborznagy: will fix.

modules/diskq/qdisk.c
gaborznagy: Wouldn't it will work the same as currently, just implemented differently?

modules/diskq/tests/Makefile.am
gaborznagy: Thanks, I will fix it.

modules/diskq/tests/test_diskq_truncate.c
gaborznagy: Includes re-checked.

modules/diskq/qdisk.c
gaborznagy: Yes.

modules/diskq/tests/Makefile.am
gaborznagy: Good catch, thanks.
I guess it worked before because of this documented behaviour of Automake:
> If _DEPENDENCIES is not supplied, it is computed by Automake. The automatically-assigned value is the contents of _LDADD or _LIBADD, with most configure substitutions, -l, -L, -dlopen and -dlpreopen options removed.

I will add these.

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
gaborznagy: Thanks for eplaining, I should fix this.

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -0,0 +1,168 @@
...
+{
+  gint number_of_msgs_to_push;
+  gint number_of_msgs_to_pop;
+  gboolean should_file_be_empty_after_truncate;
+} truncate_test_params;
+
+static void
+_test_diskq_truncate(truncate_test_params params)
+{
+  LogQueue *q;
```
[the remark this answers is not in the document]
gaborznagy: As we discussed, using `file_size` is not optimal:
 if disk_queue truncate logic is broken, but `file_size` updated accordingly and we just ask the file size with `stat` then we wouldn't see the incorrent truncate.
A better expected value needed when testing truncate.

Also, it's not enough to check only if the file is not empty.
We agreed that one way to do this is to calculate the expected size after truncate:
- serialize 1 input message as it happens in the disk-queue
- multiply with the number of expected messages
- also take into consideration the header size (QDISK_RESERVED_SPACE)
This vale can be used against a `stat()` result as expected value.

The only drawback is, that we don't use an API of the disk-queue, but reimplement something that syslog-ng does, hence it can be easily broke as well.
Hopefully not silently.

modules/diskq/qdisk.c
```diff
@@ -224,6 +224,47 @@ _truncate_file(QDisk *self, gint64 new_size)
   return success;
 }
 
+static gint64
+qdisk_get_lowest_used_queue_offset(QDisk *self)
+{
```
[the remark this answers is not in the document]
gaborznagy: Maybe the iteration does not look bad (you mentioned the `offsets[i]`), 
it's just it can be done nice.

Your suggestion isn't perfect, as in case of empty queues (offset = 0), it would result in `GINTMAX`.

modules/diskq/qdisk.c
```diff
@@ -544,30 +582,22 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
...
-        end_ofs = qoverflow_ofs;
-      if(end_ofs > QDISK_RESERVED_SPACE)
-        _truncate_file(self, end_ofs);
-    }
 
-  if (!self->options->reliable)
-    {
-      self->file_size = qout_ofs;
-      _reset_queue_pointers(self);
+      gint64 end_ofs = QDISK_RESERVED_SPACE;
```
szemere: note1: The old code did not truncate in case the file was "empty". check: Is it implemented elsewhere (in the event of "emptying" the queue)? I think this is a nice feature from the new one.

note2: _truncate_file returns with a bool, neither the old code, or the new one handles failure.

modules/diskq/qdisk.c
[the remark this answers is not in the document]
gaborznagy: I will resolve this based on @Kokan's comment, although the return value will be 0 in case no memory queue have been used (positions are 0).

gaborznagy:
Marked as WIP since it's been waiting for me for a while.

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -392,7 +392,7 @@ _save_queue (LogQueueDisk *s, gboolean *persistent)
   if (qdisk_save_state (s->qdisk, self->qout, self->qbacklog, self->qoverflow))
     {
       *persistent = TRUE;
-      qdisk_deinit (s->qdisk);
+      qdisk_stop (s->qdisk);
```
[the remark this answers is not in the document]
gaborznagy: marking as `qdisk_started` comment had been [resolved](https://github.com/balabit/syslog-ng/pull/2406/files/9b65c7ab008835536eafbf877c2e90e60570f72d#diff-85691d1ef3d403403e3710842516d838).

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
gaborznagy: About note1: Yes, truncating can happen in several cases in current implementation, e.g. during:
- popping messages from the queue (when the file becomes empty, only in non-rel. disk-queue)
- pushing messages to the queue (when there is unused space after write_head, i.e. read follows write)
- acking back messages from backlog (in case of reliable disk-queue where backlog is saved on the disk as well)
- opening, loading existing disk-queues (had a bug, which is fixed by this PR)

note2:
That's a good observation, indeed the result of the truncating is ignored in the entire disk-queue code.
It needs some thoughts to given, as it is not trivial how to resolve issues like that.
We have a restart mechanism (close current file as corrupted and open a new disk-queue file), which currently only used in case of a disk read operation fails during message popping.
The restart() mechanism is defined on LogQueueDisk level, we are on QDisk level.
We could propagate the error and see if we could either start a new disk-queue file or do something else.

Anyway, as you mentioned handling file truncating failures never handled, and I did not changed that either, just fixed a bug in one of the truncating scenarios.
I have made a lot of refactors already, and don't want to increase this PR with that too.
Please remove the change request, probably we should open a new issue about it.

modules/diskq/qdisk.c
gaborznagy: Fixed.

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -0,0 +1,168 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "logqueue.h"
+#include "logqueue-fifo.h"
```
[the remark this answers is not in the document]
gaborznagy: Removed.

modules/diskq/qdisk.c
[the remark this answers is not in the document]
gaborznagy: Reworked based on @Kokan's comment.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

gaborznagy:
I have added the requested change to the unit test:
- it calculates the expected file size after truncate and compares it to the actual file size (not to the internally stored size)
- additionally the test still checks the internally stored file size and the actual.

Please re-review the patches as I think some review comments are obsolete and for some I have not got an answer for  a while.

kira-syslogng:
Build FAILURE

gaborznagy:

kira-syslogng:
Build SUCCESS

PR #2408, opened 2018-11-12
system-source: check if /proc/kmsg can be opened

bazsi:
Instead of using an explicit exclude-kmsg() option, let's use an alternative detection mechanism to see
if we can pull /proc/kmsg. This alternative mechanism is implemented in this PR.

I've removed support for the exclude-kmsg() option but silently ignoring it if it's present in the configuration.

At the same time I've submitted a PR in syslog-ng-docker to remove system().
https://github.com/balabit/syslog-ng-docker/pull/37

### DIFF
--- a/modules/system-source/system-source.c
+++ b/modules/system-source/system-source.c
@@ -44,29 +44,6 @@
 #include <inttypes.h>
 #endif
 
-static gboolean
-_system_normalized_compare(const gchar *s1, const gchar *s2)
-{
-  gchar *normalized_s1 = __normalize_key(s1);
-  gchar *normalized_s2 = __normalize_key(s2);
-  gboolean result = g_ascii_strcasecmp(s1, s2) == 0;
-  g_free(normalized_s1);
-  g_free(normalized_s2);
-  return result;
-}
-
-static gboolean
-_system_option_normalized_test_and_remove(CfgArgs *args, const gchar *option, const gchar *test)
-{
-  const gchar *option_string = cfg_args_get(args, option);
-  if (option_string == NULL)
-    return FALSE;
-
-  gboolean result = _system_normalized_compare(option_string, test);
-  cfg_args_remove(args, option);
-  return result;
-}
-
 static void
 system_sysblock_add_unix_dgram_driver(GString *sysblock, const gchar *path,
                                       const gchar *perms, const gchar *recvbuf_size)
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build FAILURE

Kokan:
I think if there is not sufficient permission to do what is in the configuration, that is a mistake in the configuration, and it is nice that syslog-ng fails.

It could be argued that the `system()` source something that should just work and do the sane thing, and this patch does that in case of container (at least in case of docker); but does not if the syslog-ng is the main logger in the system, as the permission could be an user mistake, where the info level of message I think too soft.

My suggestion would be to change the default configuration in the docker image as that is less intrusive.

szemere:
Just for the reference: https://github.com/balabit/syslog-ng-docker/pull/36

bazsi:
@Kokan thanks for the insight, I can agree with most of what you said. I agree that the default config should be changed in docker, but I argue that we didn't go far enough. If we are running syslog-ng in a container, it should NOT even use the system() source. There's no point in pulling /dev/log in a container, especially if the only process in that container is syslog-ng itself. We might want to add a default configuration for containers.

That should resolve this issue, albeit those users who supply their own configuration to the image will need to change it, which is still not ideal. I've checked the documentation for how we document the system source, it says:

```Starting with version 3.2, syslog-ng OSE can automatically collect the system-specific log messages of the host on a number of platforms using the system() driver. If the system() driver is included in the syslog-ng OSE configuration file, syslog-ng OSE automatically adds the following sources to the syslog-ng OSE configuration.
```

Also, on Linux it says:

```

unix-dgram("/dev/log");
file("/proc/kmsg" program-override("kernel") flags(kernel)); 

Note that on Linux, the so-rcvbuf() option of the system() source is automatically set to 8192. If the host is running under systemd, syslog-ng OSE reads directly from the systemd journal file using the systemd-journal() source. 

If the kernel of the host is version 3.5 or newer, and /dev/kmsg is seekable, syslog-ng OSE will use that instead of /proc/kmsg, using the multi-line-mode(indented), keep-timestamp(no), and the format(linux-kmsg) options. 

If syslog-ng OSE is running in a jail or a Linux Container (LXC), it will not read from the /dev/kmsg or /proc/kmsg files.

```

At least the documentation should be updated to remove the container autodetection pieces. Let's discuss this IRL.

bazsi:
We've discussed this and decided to remove exclude-kmsg() in favour of trying to open /proc/kmsg if it works. I've also submitted a PR in the docker image repository.

https://github.com/balabit/syslog-ng-docker/pull/37

kira-syslogng:
Build SUCCESS

szemere:
@bazsi , @lbudai , @Kokan , @fekete-robert : don't forget to update the documentation. Thank You!

PR #2410, opened 2018-11-13
http: do not load certs from default CURL ca-dir

lbudai:
when ca-dir() is not set, syslog-ng uses the default ca-dir, provided by
CURL (on my machine: /etc/ssl/certs)


kira-syslogng:
Build SUCCESS

PR #2415, opened 2018-11-15
pytest-fw to travis

mitzkia:
- run pytest-fw related make targets from Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ install:
       criterion-dev
       libmaxminddb-dev
       libxml2-utils
-  - sudo pip install -r requirements.txt
+  - pip install --user -r requirements.txt
 before_script:
   - echo 'Europe/Budapest' | sudo tee /etc/timezone
   - sudo dpkg-reconfigure --frontend noninteractive tzdata
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,6 +88,8 @@ script:
         . scripts/get-libjvm-path.sh || return $?;
         export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$JNI_LIBDIR";
         make func-test V=1;
+        make pytest-self-check;
+        make pytest-check;
       elif [ "$S" = "42" ]; then
         return $S;
       else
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build FAILURE

.travis.yml
```diff
@@ -88,6 +88,8 @@ script:
         . scripts/get-libjvm-path.sh || return $?;
         export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$JNI_LIBDIR";
         make func-test V=1;
+        make VERBOSE=1 pytest-self-check;
```
MrAnno: We could add a separate parallel job for `self-check` in the future. It does not depend on the compilation of syslog-ng.

furiel:

.travis.yml
```diff
@@ -88,6 +88,8 @@ script:
...
+        make VERBOSE=1 pytest-check;
```
Kokan: Tip: If you want a verbose build with `autotools` you should use `V=1`, and the `VERBOSE=1` is for `cmake`.

.travis.yml
mitzkia: Yes, you are right. For now is it ok to leave it here?

.travis.yml
mitzkia: Oh, thanks for the tip. Otherwise as I saw the output, it does not add any other information for us. So I would remove this VERBOSE=1 from here, is it ok?
Otherwise when we want to get some debug info from failing tests we can run pytest locally with "python -m pytest ... -lvs --loglevel=debug [some test]". This information also can be public, I mean I can extend our wiki page in GitHub with some useful pytest commands, or can add a Readme.md to tests/pytest-framework/. What do you think?

Kokan: I agree to remove the verbosity.

I did not try the `autotool` version, but in case of `cmake` (+`Make`) the output of the pytest is still printed, what is not printed that what command is used to start pytest.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2416, opened 2018-11-15
Fix telegram dst default ca dir

lbudai:
PR #2410 breaks the Telegram destination.

This patch adds new option to HTTP destination (`enable-default-ca-dir`).
When defaults are enabled and the user does not set the `ca-dir` option,a default ca-dir is used, provided by the CURL library.

When both options are set (`ca-dir` and `enable-default-ca-dir`), of course, `ca-dir` will be used.

### DIFF
--- a/modules/http/http-grammar.ym
+++ b/modules/http/http-grammar.ym
@@ -61,6 +61,7 @@
 %token KW_CERT_FILE
 %token KW_KEY_FILE
 %token KW_CIPHER_SUITE
+%token KW_USE_SYSTEM_CERT_STORE
 %token KW_SSL_VERSION
 %token KW_PEER_VERIFY
 %token KW_TIMEOUT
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build FAILURE

modules/http/http.c
```diff
@@ -103,7 +103,8 @@ _setup_static_options_in_curl(HTTPDestinationWorker *self)
   if (owner->user_agent)
     curl_easy_setopt(self->curl, CURLOPT_USERAGENT, owner->user_agent);
 
-  curl_easy_setopt(self->curl, CURLOPT_CAPATH, owner->ca_dir);
+  if (owner->ca_dir || owner->enable_default_ca_dir)
```
MrAnno: Shouldn't we negate the right side?

kira-syslogng:
Build SUCCESS

modules/http/http-grammar.ym
```diff
@@ -134,6 +135,7 @@ http_tls_options
 
 http_tls_option
     : KW_CA_DIR     '(' string ')'            { http_dd_set_ca_dir(last_driver, $3); free($3); }
+    | KW_ENABLE_DEFAULT_CA_DIR '(' yesno ')'  { http_dd_set_enable_default_ca_dir(last_driver, $3); free($3); }
```
MrAnno: Oops, I haven't noticed, but this `free()` should be removed.

lbudai: copy-paste :/ (I've just wanted to close this topic as fast as possible, but maybe too fast...)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
@bazsi: I heard that you don't really like the name of the option :) I'm open to replace it to (almost...) anything (`use-default-ca-dir`, whatever).

kira-syslogng:
Build SUCCESS

bazsi:
I was thinking to make it less closely related to ca-dir option, and make
it something like:

use-system-cert-store(yes/no) and that would affect both ca-dir() and
ca-file().

I am not completely satisfied with this either, my point is that it should
relate to the system ca store as a concept and not specific to one part of
it (e.g. ca-dir).

kira-syslogng:
Build SUCCESS

PR #2423, opened 2018-11-23
Adding build flag -Wmissing-format-attribute and -Wsuggest-attribute=noreturn

### LINKED ISSUE #1961: warning elimination: function attribute suggestions

This issue should cover the function attribute suggestion related warnings.
```
     10 [-Wsuggest-attribute=format]
      2 [-Wsuggest-attribute=noreturn]
      ? [-Wformat=2]
      ? [-Wmissing-format-attribute] (probably alias for -Wsuggest-attribute=format)
```

alltilla:
This PR fixes #1961.

-Wmissing-format-attribute is the alias for -Wsuggest-attribute=format.
This attribute can be added to any function, which has a format string in its arguments, and possibly additional variable arguments, intended to match with the format string.
When 'format' attribute is added, the compiler checks, whether the format string is valid, and the variable arguments match the format.
After enabling this flag, several functions were shown to be candidates for the format attribute.
As clang and gcc has different syntactic, I made macros for every module, which was affected, with different syntactic for the mentioned compilers.
I wanted to make only one macro, but I did not find the mutual source file for it.
After marking each candidate functions with 'format' attribute, the compiler found some calls, where the format string and the arguments were not compatible. I fixed these instances.

-Wsuggest-attribute=noreturn suggests the compiler, that the code will not return from the function call.
I found, that clang and gcc had the same syntactic, so macro was not needed here.
After enabling this warning flag, I found some instances, where the attribute could be added to a function.
I checked the functions, added the 'noreturn' attribute and the warnings disappeared.

I do not suggest turning on -Wformat=2 make flag, as it throws warnings for such cases, where the format string is built run-time, and the compiler is not able to check those cases with -Wmissing-format-attribute. So the warnings caused by -Wformat=2 cannot be eliminated.

I tested the PR with Travis and it passed, but I found, that with clang, the extra warnings flags are not enabled, so I tested this scenario manually, by configuring with --enable-extra-warnings and building with clang. I found the build output to be as expected, the new warnings (caused by turning on the make flags) were cleared with the code modification.

### DIFF
--- a/lib/eventlog/src/evtlog.h
+++ b/lib/eventlog/src/evtlog.h
@@ -48,6 +48,7 @@
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
+#include <glib.h>
 
 #include "evtmaps.h"
 
@@ -82,12 +83,6 @@
 #define EVT_FAC_LOCAL5      (21<<3) /* reserved for local use */
 #define EVT_FAC_LOCAL6      (22<<3) /* reserved for local use */
 #define EVT_FAC_LOCAL7      (23<<3) /* reserved for local use */
-
-#ifdef __GNUC__
-#define EVT_GNUC_PRINTF_FUNC(format_idx, first_arg_idx)  __attribute__((format(printf, format_idx, first_arg_idx)))
-#else
-#define EVT_GNUC_PRINTF_FUNC(format_idx, first_arg_idx)
-#endif
 
 /* EVTCONTEXT encapsulates logging specific parameters like the
  * program name and facility to use */
@@ -148,7 +143,7 @@ EVTTAG *evt_tag_str(const char *tag, const char *value);
 EVTTAG *evt_tag_int(const char *tag, int value);
 EVTTAG *evt_tag_long(const char *tag, long long value);
 EVTTAG *evt_tag_errno(const char *tag, int err);
-EVTTAG *evt_tag_printf(const char *tag, const char *format, ...) EVT_GNUC_PRINTF_FUNC(2, 3);
+EVTTAG *evt_tag_printf(const char *tag, const char *format, ...) G_GNUC_PRINTF(2, 3);
 EVTTAG *evt_tag_inaddr(const char *tag, const struct in_addr *addr);
 EVTTAG *evt_tag_inaddr6(const char *tag, const struct in6_addr *addr);
 
@@ -180,8 +175,8 @@ int evt_log(EVTREC *e);
 /* syslog wrapper */
 void evt_openlog(const char *ident, int option, int facility);
 void evt_closelog(void);
-void evt_vsyslog(int pri, const char *format, va_list ap);
-void evt_syslog(int pri, const char *format, ...) EVT_GNUC_PRINTF_FUNC(2, 3);
+void evt_vsyslog(int pri, const char *format, va_list ap) G_GNUC_PRINTF(2, 0);
+void evt_syslog(int pri, const char *format, ...) G_GNUC_PRINTF(2, 3);
 
 #ifdef EVENTLOG_SYSLOG_MACROS
 
--- a/lib/logproto/tests/test-text-server.c
+++ b/lib/logproto/tests/test-text-server.c
@@ -360,7 +360,7 @@ accumulator_assert_that_lines_are_starting_with_sequence_number(LogProtoTextServ
 {
   assert_true((msg[0] - '0') == accumulate_seq,
               "accumulate_line: Message doesn't start with sequence number, msg=%.*s, seq=%d",
-              msg_len, msg, accumulate_seq);
+              (int)msg_len, msg, accumulate_seq);
   assert_gint(consumed_len, -1, "Initial invocation of the accumulator expects -1 as consumed_len");
   accumulate_seq++;
   return LPT_CONSUME_LINE | LPT_EXTRACTED;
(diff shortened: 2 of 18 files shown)


kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

modules/diskq/dqtool.c
```diff
@@ -225,7 +225,7 @@ dqtool_mode(int *argc, char **argv[])
   return NULL;
 }
 
-void
+void __attribute__((noreturn))
```
Kokan: I would avoid using `__attribute__` as much as it can be.
Possible it is either way better not to `exit(1)` in the `usage` method, it should just print the usage to `stderr` (do one thing).
The caller could do the `exit(1)`, but that is actually being the main could simple `return 1`.

(Possible this can be applied to the `pdbtool` as well.)

kira-syslogng:
Build SUCCESS

lib/eventlog/src/evtlog.h
```diff
@@ -84,9 +84,11 @@
 #define EVT_FAC_LOCAL7      (23<<3) /* reserved for local use */
 
 #ifdef __GNUC__
-#define EVT_GNUC_PRINTF_FUNC(format_idx, first_arg_idx)  __attribute__((format(printf, format_idx, first_arg_idx)))
+#define EVT_FORMAT_PRINTF_FUNC(format_idx, first_arg_idx)  __attribute__((format(printf, format_idx, first_arg_idx)))
```
Kokan: We are using `glib`, that actually defined this macro: `G_GNUC_PRINTF`. I think it would easier to use that (and possibly more compatible).

lib/logproto/tests/test-text-server.c
```diff
@@ -360,7 +360,7 @@ accumulator_assert_that_lines_are_starting_with_sequence_number(LogProtoTextServ
 {
   assert_true((msg[0] - '0') == accumulate_seq,
               "accumulate_line: Message doesn't start with sequence number, msg=%.*s, seq=%d",
-              msg_len, msg, accumulate_seq);
+              (int)msg_len, msg, accumulate_seq);
```
Kokan: The `gsize` is probably a `long`, you should not cast to `int` because of printing. I would use the `G_GSSIZE_FORMAT` format string, which has the proper format string to this type.

alltilla: In this case, `msg_len` is not printed, it just sets the precision of the `msg` string.
The precision can only be given as an `int`, this is why I only casted it. Alternatively we could get rid of the precision, or convert `msg_size` to int with a safer routine. Which one would you prefer? Thanks!

lib/eventlog/src/evtlog.h
Kokan: The above macro `G_GNUC_PRINTF` is going to work with clang also, as clang pretend to be gcc `4.x` so defines `__GNUC__`.

lib/logproto/tests/test-text-server.c
Kokan: Sorry, my bad. In that case the `(int)` conversion is fine.

lib/eventlog/src/evtlog.h
Kokan: Oh and if it is not clear, so get ride of those macros and use this one directly.

alltilla: Thank you Kokan, this clears up a lot of copy pasted code. I will change it as you asked.

kira-syslogng:
Build FAILURE

libtest/testutils.c
```diff
@@ -142,7 +142,7 @@ assert_guint64_non_fatal(guint64 actual, guint64 expected, const gchar *error_me
     return TRUE;
 
   va_start(args, error_message);
-  print_failure(error_message, args, "actual=%llu, expected=%llu", actual, expected);
+  print_failure(error_message, args, "actual=%lu, expected=%lu", actual, expected);
```
MrAnno: I'd use `G_GUINT64_FORMAT`.

libtest/testutils.c
```diff
@@ -333,13 +333,13 @@ assert_gboolean_non_fatal(gboolean actual, gboolean expected, const gchar *error
...
 
-gboolean
+gboolean G_GNUC_PRINTF(2, 0)
 assert_null_non_fatal_va(const void *pointer, const gchar *error_message, va_list args)
 {
   if (pointer == NULL)
     return TRUE;
 
-  print_failure(error_message, args, "Pointer expected to be NULL; pointer=%llx", (guint64)pointer);
+  print_failure(error_message, args, "Pointer expected to be NULL; pointer=%lx", (guint64)pointer);
```
MrAnno: Why don't we use `%p` here as you did below?

MrAnno:
`eventlog` is a separate shared library in the "core" of syslog-ng. Since you've started to use
glib in that library, you should add its include directories to the CMake build (I guess linking to glib is not necessary at the moment, but it looks cleaner doing it).

Something like this should work:
```diff
--- a/lib/eventlog/CMakeLists.txt
+++ b/lib/eventlog/CMakeLists.txt
@@ -8,7 +8,15 @@ add_library(eventlog SHARED
   src/evttags.c
 )

-target_include_directories(eventlog SYSTEM INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/src)
+target_include_directories(eventlog
+  SYSTEM INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/src
+  PRIVATE ${GLIB_INCLUDE_DIRS}
+)
+
+target_link_libraries(eventlog
+  PRIVATE
+    ${GLIB_LIBRARIES}
+)
```

`lib/eventlog/src/Makefile.am` has to be modified as well, but it works without it for some reasons (transitivity might be the reason).

kira-syslogng:
Build SUCCESS

PR #2425, opened 2018-11-23
internal source + trace level causes message storm

furiel:
Problem: when internal source is enabled and trace level set, infinite message is generated in an exponential manner.

There are two independent root cause for this, both of these need to be addressed.

1)
There is a loop detection that was not working, because the order msg_set_context and the triggering messages were swapped. With that change, the `Source side message processing begin` messages are suppressed.

With that `internal() messages are looping back, preventing loop by ...` messages occur, so the detection kicks in, avoiding the infinite loop, but still causing minor inconvenience.

2)
There is a message about window size adjustment, that also triggers a message each time a message arrives into the internal queue. However, the loop detection does not work on this, and should not, the window size change is completely valid event in this case.

For now I just deleted the msg_trace, but this is probably suboptimal solution.

-----

I was thinking that maybe trace level messages could be written to stderr unconditionally. To my undrestanding, that would also stop the infinite generation of messages, because the trace messages are not put back into the internal queue.

### DIFF
--- a/lib/logsource.c
+++ b/lib/logsource.c
@@ -369,12 +369,12 @@ log_source_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options
   LogSource *self = (LogSource *) s;
   gint i;
 
+  msg_set_context(msg);
+
   msg_trace(">>>>>> Source side message processing begin",
             evt_tag_str("instance", self->stats_instance ? self->stats_instance : "internal"),
             log_pipe_location_tag(s),
             evt_tag_printf("msg", "%p", msg));
-
-  msg_set_context(msg);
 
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

furiel:
We discussed the topic IRL, and the conclusion is: there will be an `msg_diagnostics`, that will write to stderr unconditionally.  That will be used for these three messages.

kira-syslogng:
Build SUCCESS

lbudai:
I still don’t really like this, but I do understand why this is the solution.

lbudai:
@furiel: and thanks for the investigation, also for the detailed commit messages!

PR #2426, opened 2018-11-24
Make include-path more discoverable

bazsi:
I received feedback that the support for include-path was not intuitive at all, and while going into this I've discovered, that it's:

* not documented
* none of the log messages contain anything on it

So I've decided to rectify the situation on the code side, this patch adds a debug message and an additional line to --version output, so both add the include-path.

Also @fekete-robert it would be great if we could document this in https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.18/administration-guide/15


kira-syslogng:
Build SUCCESS

fekete-robert:
I've added it to our todo list

PR #2430, opened 2018-11-26
system-source: allow syslog-protocol on unix-dgram() sources

bazsi:
FreeBSD has started to use RFC5424 format between applications and syslogd
when using the syslog() API. Allow this use by enabling the syslog-protocol
flag.


kira-syslogng:
Build SUCCESS

PR #2434, opened 2018-11-27
mainloop-worker-io: possible refcount leak during reload/shutdown

furiel:
When reload/shutdown is triggered, syslog-ng waits for the ongoing jobs to finish. Specially, a MainLoopIOWorkerJob early returns without scheduling the ivykis task. The problem with the current implementation is that api users call `log_pipe_unref` in the `completion` callback, that is not called in this case, because the ivykis task is not scheduled at all.

To mitigate that, users can specify a destroy function that will be used to `unref` the user_data before returning. Practically the ownership of user_data is given to (shared between the pipe and) MainLoopIOWorkerJob.

Reproduction steps:

Starting the config below with -Fv (verbose is mandatory, must not use -e)

```
@version: 3.18
@include "scl.conf"

options{
    time-reap(5);
};

destination d_java {
  java( ... );
};

log {
  source { internal(); };
  destination { file("/tmp/tmp.txt"); };
};

log {
  destination(d_java);
};
```
after reload one can see syslog-ng leaked an fd related to /tmp/tmp.txt (check output of lsof -p ...). Java destination is not necessary for the condition to trigger, but for some reason such unused destination increased the condition frequency dramatically.

### DIFF
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -214,10 +214,8 @@ log_reader_io_process_input(gpointer s)
        */
       if (!main_loop_worker_job_quit())
         {
-          log_pipe_ref(s);
           log_reader_work_perform(s);
           log_reader_work_finished(s);
-          log_pipe_unref(s);
         }
     }
 }
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -197,9 +197,9 @@ log_reader_io_process_input(gpointer s)
   LogReader *self = (LogReader *) s;
 
   log_reader_stop_watches(self);
-  log_pipe_ref(&self->super.super);
   if ((self->options->flags & LR_THREADED))
     {
+      log_pipe_ref(&self->super.super);
       main_loop_io_worker_job_submit(&self->io_job);
     }
   else
@@ -214,6 +214,7 @@ log_reader_io_process_input(gpointer s)
        */
       if (!main_loop_worker_job_quit())
         {
+          log_pipe_ref(s);
           log_reader_work_perform(s);
           log_reader_work_finished(s);
           log_pipe_unref(s);
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -199,7 +199,6 @@ log_reader_io_process_input(gpointer s)
   log_reader_stop_watches(self);
   if ((self->options->flags & LR_THREADED))
     {
-      log_pipe_ref(&self->super.super);
       main_loop_io_worker_job_submit(&self->io_job);
     }
   else
@@ -243,6 +242,7 @@ log_reader_init_watches(LogReader *self)
   self->io_job.user_data = self;
   self->io_job.work = (void (*)(void *)) log_reader_work_perform;
   self->io_job.completion = (void (*)(void *)) log_reader_work_finished;
+  self->io_job.engage = (void (*)(void *)) log_pipe_ref;
   self->io_job.release = (void (*)(void *)) log_pipe_unref;
 }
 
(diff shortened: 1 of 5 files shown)


bazsi:
This is a nice catch and improves the readability of the various _work() methods so they don't need to unref(self), which is much better. I have some nitpicks, which I am filing as comments.

lib/logreader.c
```diff
@@ -215,8 +214,10 @@ log_reader_io_process_input(gpointer s)
        */
       if (!main_loop_worker_job_quit())
         {
+          log_pipe_ref(&self->super.super);
           log_reader_work_perform(s);
           log_reader_work_finished(s);
+          log_pipe_unref(s);
         }
```
bazsi: I don't think it is necessary to take + drop a reference here, unless I am missing something. 

We have a borrowed reference due to our argument. Also, should it still be needed there's an asymmetry 
between the ref/unref call, one taking self->super.super, the other taking an untyped "s".

lib/logwriter.c
```diff
@@ -265,8 +262,10 @@ log_writer_io_flush_output(gpointer s)
 
       if (!main_loop_worker_job_quit())
         {
+          log_pipe_ref(&self->super);
           log_writer_work_perform(s);
           log_writer_work_finished(s);
+          log_pipe_unref(s);
         }
```
bazsi: again, no need for the ref/unref here.

lib/logwriter.c
```diff
@@ -247,9 +244,9 @@ log_writer_io_flush_output(gpointer s)
   main_loop_assert_main_thread();
 
   log_writer_stop_watches(self);
-  log_pipe_ref(&self->super);
   if ((self->options->options & LWO_THREADED))
     {
+      log_pipe_ref(&self->super);
       main_loop_io_worker_job_submit(&self->io_job);
     }
```
bazsi: I think this could be made a bit more explicit that self->io_job->user_data equals to self. What about this
```
self->io_job->user_data = log_pipe_ref(&self->super);
```
and the same on the logreader side.

lib/mainloop-io-worker.h
```diff
@@ -32,6 +32,7 @@ typedef struct _MainLoopIOWorkerJob
 {
   void (*work)(gpointer user_data);
   void (*completion)(gpointer user_data);
+  void (*destroy)(gpointer user_data);
   gpointer user_data;
```
bazsi: I might suggest renaming the destroy() function, it does not destroy the Job object at all. I thinkit merely cleans it up, so cleanup() would be more appropriate.

Also, I am a bit unconvinced about the coupling between the destroy callback and the fact that we are taking the ref outside of the responsibilities of the Job class. What if we had two methods? One that would take the ref (in the main thread) and another that would release it (in complete). Both would be methods of the MainLoopIOWorkerJob
class, thus the ref/unref dance would nicely be encapsulated and would be easy to follow.

The tricky question is how we would name these functions. What about:
  * engage(): right before submitting the job
  * release(): at completion

This way we wouldn't even need to call destroy in the main_loop_io_worker_submit() function, as we would bail before calling engage().

What do you think?

lib/logreader.c
furiel: Good point, I do not see any reason either. I just blindly copied the ref/unref to maintain original behavior, but I agree these are not necessary in the nonthreaded variant. I will remove these.

lib/mainloop-io-worker.h
furiel: I like the preparation+cleanup function idea, and the engage/release names also sound good. I will rewrite the patch according to your suggestion.

kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build SUCCESS

PR #2438, opened 2018-12-01
Fix Python package/module name collision

### LINKED ISSUE #2364: the "syslogng" Python module name is used for two purposes, one shadowing the other
# syslog-ng

## Version of syslog-ng

3.18.1 and latest master.
## Platform

Linux, but is any.

# Issue

We are using the "syslogng" python module for two purposes and one use shadows the other, breaking the interactive debugger.

1) it is used as the package name for all Python code supplied to syslog-ng (e.g. modules/python/pylib)
2) it is used as a builtin module by the Python fetcher/source/template stuff

We already had a _syslogng module, and I would propose to use that to publish internal stuff and wrap it from the python sources to make it available to user code. This should improve flexibility in situations where compatibility needs to be retained.

## Failure

The failure is that the interactive debugger (-i command line switch) cannot import its tab completer implementation.

## Steps to reproduce

1. start syslog-ng with -i
2. send a message to trigger a breakpoint
3. see the error message about not being able to import syslogng.debuggercli

## Configuration

any

any log message should trigger it.

MrAnno:
Previously, `syslogng` was a package and a module name at the same time, breaking the interactive debugger.

C-side Python objects have been moved into the `_syslogng` module, and these objects have been published in the `syslogng` package.

From now on, the main module runs in the `_syslogng_main` module instead of `_syslogng`.

The installpath of pylib has been changed to `lib/syslog-ng/python`. This path is automatically appended to `PYTHONPATH`. Relocation is also supported using the `SYSLOGNG_PREFIX` env variable.

Fixes #2364


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

MrAnno:
Automatically appending `PYTHONPATH` with the corresponding path is not as clean as I thought.

The path is "generated" by `setup.py` during installation, however, we have the path of every installed file in `builddir/modules/python/pylib/install-manifest.txt`.

The Java binding's module-dir was a bit easier, because it is a hard-coded path, it does not change based on binding versions (for example, `syslog-ng/ose/install/lib/python3.7/site-packages/`).

furiel:
I feel this PR is a little bit stuck. Just an idea: we could move the debugger to `_syslogng` package instead, to avoid collision. It is far from optimal solution, as there are other benefits if we could write code in python for the c binding. But as quick workaround to get the debugger running, it might be enough.

furiel:
`setup.py` has a `--install-lib` option that can override the installation directory entirely. That means the path would not depend on python version. With that we could force a known installation path, so we can update PYTHONPATH correctly.

For example after adding `--install-lib=/tmp/test` to `SetupPyLib` target, after the installation I got

<details>

```
/tmp/test/
├── syslogng
│   ├── debuggercli
│   │   ├── choicecompleter.py
│   │   ├── commandlinelexer.py
│   │   ├── completerlang.py
│   │   ├── completer.py
│   │   ├── debuggercli.py
│   │   ├── debuglang.py
│   │   ├── getoptlexer.py
│   │   ├── __init__.py
│   │   ├── langcompleter.py
│   │   ├── lexer.py
│   │   ├── lexertoken.py
│   │   ├── macrocompleter.py
│   │   ├── __pycache__
│   │   │   ├── choicecompleter.cpython-35.pyc
│   │   │   ├── commandlinelexer.cpython-35.pyc
│   │   │   ├── completer.cpython-35.pyc
│   │   │   ├── completerlang.cpython-35.pyc
│   │   │   ├── debuggercli.cpython-35.pyc
│   │   │   ├── debuglang.cpython-35.pyc
│   │   │   ├── getoptlexer.cpython-35.pyc
│   │   │   ├── __init__.cpython-35.pyc
│   │   │   ├── langcompleter.cpython-35.pyc
│   │   │   ├── lexer.cpython-35.pyc
│   │   │   ├── lexertoken.cpython-35.pyc
│   │   │   ├── macrocompleter.cpython-35.pyc
│   │   │   ├── readline.cpython-35.pyc
│   │   │   ├── syslognginternals.cpython-35.pyc
│   │   │   ├── tablexer.cpython-35.pyc
│   │   │   ├── templatelang.cpython-35.pyc
│   │   │   ├── templatelexer.cpython-35.pyc
│   │   │   └── tflang.cpython-35.pyc
│   │   ├── readline.py
│   │   ├── syslognginternals.py
│   │   ├── tablexer.py
│   │   ├── templatelang.py
│   │   ├── templatelexer.py
│   │   └── tflang.py
│   ├── __init__.py
│   └── __pycache__
│       └── __init__.cpython-35.pyc
└── syslogng-1.0-py3.5.egg-info
```

</details>

MrAnno:
@furiel Thanks.

I've used the relocation API, because `SYSLOGNG_PREFIX` is actively used in syslog-ng PE.

There is something wrong with configs containing inline Python code (`python_evaluate_global_code`), I'll fix that before removing the WIP flag.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

MrAnno:
It seems the `_syslogng` module is also overloaded.
It is used to store C-side Python objects and to load and run inline Python code too.

When executing inline Python code, `_syslogng` is removed from the module registry and it is recreated after every reload (`_py_construct_main_module`). This removes all C-side objects that are registered in `_py_init_interpreter`.

I'll just rename the main module to `_syslogng_main` to resolve this issue.

kira-syslogng:
Build SUCCESS

PR #2439, opened 2018-12-01
python-debugger: fix macro completion

MrAnno:
This PR fixes the macro completion feature of the interactive debugger.

----------

`syslognginternals.py` contains a line to override the Python-side
implementation with the C implementation:

`from _syslogngdbg import *`

This means that the C side has to be published under the `_syslogngdbg`
module.

The `macro.decode()` call is required to be compatible with Python 3.

### DIFF
--- a/modules/python/python-debugger.c
+++ b/modules/python/python-debugger.c
@@ -34,7 +34,8 @@ _add_nv_keys_to_list(gpointer key, gpointer value, gpointer user_data)
   PyObject *list = (PyObject *) user_data;
   const gchar *name = (const gchar *) key;
 
-  PyObject *py_name = PyBytes_FromString(name);
+  PyObject *py_name = _py_string_from_string(name, -1);
+
   PyList_Append(list, py_name);
   Py_XDECREF(py_name);
 }
@@ -63,7 +64,7 @@ static PyMethodDef _syslogngdbg_functions[] =
 static struct PyModuleDef syslogngdbgmodule =
 {
   .m_base    = PyModuleDef_HEAD_INIT,
-  .m_name    = "syslogngdbg",
+  .m_name    = "_syslogngdbg",
   .m_size    = -1,
   .m_methods = _syslogngdbg_functions
 };
@@ -84,7 +85,7 @@ static void
 PyInit_syslogngdbg(void)
 {
   PyGILState_STATE gstate = PyGILState_Ensure();
-  Py_InitModule("syslogngdbg", _syslogngdbg_functions);
+  Py_InitModule("_syslogngdbg", _syslogngdbg_functions);
   PyGILState_Release(gstate);
 }
 
@@ -93,7 +94,7 @@ PyInit_syslogngdbg(void)
 void
 python_debugger_append_inittab(void)
 {
-  PyImport_AppendInittab("syslogngdbg", &PyInit_syslogngdbg);
+  PyImport_AppendInittab("_syslogngdbg", &PyInit_syslogngdbg);
 }
 
 #define DEBUGGER_FETCH_COMMAND "syslogng.debuggercli.fetch_command"


kira-syslogng:
Build SUCCESS

bazsi:
The collision that I've found was not on the debugger symbols, rather it was on the module named "syslogng". Because our pylib installs a syslogng package and we also initialize it internally.

These PyImport_AddModule() would register syslogng without reading syslogng/__init__.py.

```
bazsi@bzorp:~/zwa/projects/syslog-ng-ose-3.6/source/syslog-ng/modules/python$ git grep -n '"syslogng"'
python-fetcher.c:571:  PyModule_AddObject(PyImport_AddModule("syslogng"), "LogFetcher", (PyObject *) &py_log_fetcher_type);
python-logmsg.c:352:  PyModule_AddObject(PyImport_AddModule("syslogng"), "LogMessage", (PyObject *) &py_log_message_type);
python-logtemplate.c:151:  PyModule_AddObject(PyImport_AddModule("syslogng"), "LogTemplate", (PyObject *) &py_log_template_type);
python-logtemplate.c:155:  PyObject_SetAttrString(PyImport_AddModule("syslogng"), "LTZ_LOCAL", PY_LTZ_LOCAL);
python-logtemplate.c:156:  PyObject_SetAttrString(PyImport_AddModule("syslogng"), "LTZ_SEND", PY_LTZ_SEND);
python-logtemplate.c:162:  PyModule_AddObject(PyImport_AddModule("syslogng"), "LogTemplateException", (PyObject *)PyExc_LogTemplate);
python-source.c:582:  PyModule_AddObject(PyImport_AddModule("syslogng"), "LogSource", (PyObject *) &py_log_source_type);
```

This way, syslogng.debuggercli as a module will not be available. You might be right that that import is bogus, but I think that's not the only problem.

MrAnno:
This PR fixes only a macro completion bug. I've opened #2438 separately, that is the one that should fix the problem you've found.

bazsi:
You are right, i just noticed the other branch, and i approved that. I'm this case this looks good as well.

bazsi:
Hmm just one note, shouldn't we return strings instead of bytes, so that we don't need an exlicit decode? I think that would make more sense. Name value pairs should be utf8. And if they aren't, we can simply sanitize or skip them when the debugger queries the registry.

furiel:
Now I feel awkward, because I wanted to say that @bazsi has the details :).

Well it was quite some time ago, but here is how I remember. The decision was made during
https://github.com/balabit/syslog-ng/pull/1931. Specially in commit  https://github.com/balabit/syslog-ng/pull/1931/commits/bb3f3886ba94f3c924a4a77d09d483ce87a5824c when we made expicit that `LogMessage` should carry `PyBytes` with a relaxation: if not `PyBytes` are passed, we presumed it is a unicode string. This works nicely with utf8, and also gives a possibility to carry binary data in a logmessage. This way the task for conversion was delegated to the user.
The alternative would have been that syslog-ng always does the conversion during set and get, but that could have been problematic if the data weren't convertible, and also had a performance cost.
From v2/v3 compatibility point of view, we already used PyBytes prior that PR, just the users might weren't conscious about it, because in python2 PyBytes was equivalent to string. With python3, PyBytes is not string anymore, and by (CPython) design, the conversion problem was delegated to the users. So the problem is not syslog-ng specific. Therefore, python3 users would already expect such kind of troubles, and know how to deal wih them.

As such, our e2e tests were broken in python3, because in python3 the strings in test results started to receive a `b'` prefix, that I could resolve by `.decode()`-ing every occurrence of `msg["MSG"].decode()`. That worked with python2 and python3 too. That's why I suggested using `.decode()` for @MrAnno here too.

bazsi:
Hi,

But this is not the _value_ of the nvpair, but the key and this is an
internal API. Let me go there IRL and discuss this.

Cheers,
Bazsi

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2440, opened 2018-12-03
PytestPART1: Add some bug fixes

mitzkia:
This patch set fixes some previous bugs in pytest fw.

### DIFF
--- a/tests/pytest_framework/conftest.py
+++ b/tests/pytest_framework/conftest.py
@@ -29,7 +29,7 @@ from src.setup.unit_testcase import SetupUnitTestcase
 
 
 def pytest_addoption(parser):
-    parser.addoption("--runslow", action="store_true", help="Also run @slow tests.")
+    parser.addoption("--runslow", action="store_true", default=False, help="run slow tests")
     parser.addoption(
         "--loglevel",
         action="store",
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build FAILURE

furiel:

tests/pytest_framework/src/syslog_ng_config/statements/destinations/destination_driver.py
```diff
@@ -35,11 +35,12 @@ def __init__(self, logger_factory, native_driver_io_ref):
     def dd_read_logs(self, path, counter):
         if path not in self.__native_driver_io_collection.keys():
             native_driver_io = self.__native_driver_io_ref(self.__logger_factory, path)
-            self.__native_driver_io_collection.update({path: native_driver_io})
             native_driver_io.wait_for_creation()
             message_reader = MessageReader(
                 self.__logger_factory, native_driver_io.read, SingleLineParser(self.__logger_factory)
             )
+            self.__native_driver_io_collection.update({path: message_reader})
```
furiel: Just a side note: I do not see why we need a hashmap of path->message_reader objects for a destination driver. Probably simply storing a single reader would be enough. At a later PR, this might be factored out.

For now, this fix is needed.

kira-syslogng:
Build SUCCESS

tests/pytest_framework/conftest.py
```diff
@@ -65,9 +65,13 @@ def loglevel(request):
     return request.config.getoption("--loglevel")
 
 
-@pytest.fixture
-def runslow(request):
-    return request.config.getoption("--runslow")
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("--runslow"):
```
bazsi: how are slow tests marked? because as I see the runslow fixture was removed.

how do you set item.keywords?

mitzkia: Sorry, I forgot add example here. 
from testcase you can use pytest 'mark' feature:
import pytest
@pytest.mark.slow
def test_my_slow_test():
   ....


docs:
https://docs.pytest.org/en/latest/mark.html
https://docs.pytest.org/en/latest/example/simple.html

bazsi:
I understand, its just good practice to not have to read the code to know
this :)

PR #2442, opened 2018-12-03
PytestPART3: SyslogNgCli updates

mitzkia:
This patch set extends SyslogNgCli with core file detection and with --run-with-valgrind fixture.
The --run-with-valgrind fixture enables to run syslog-ng in any functional test behind valgrind. 
In this case on syslog-ng stop in handle_valgrind_log() we can assert on any unexpected behaviour (invalid read).

### DIFF
--- a/tests/pytest_framework/src/syslog_ng/syslog_ng.py
+++ b/tests/pytest_framework/src/syslog_ng/syslog_ng.py
@@ -21,15 +21,10 @@
 #
 #############################################################################
 
-from src.syslog_ng.console_log_reader import ConsoleLogReader
-from src.syslog_ng.syslog_ng_cli import SyslogNgCli
-
 
 class SyslogNg(object):
-    def __init__(self, logger_factory, instance_paths, syslog_ng_ctl):
-        self.__syslog_ng_cli = SyslogNgCli(
-            logger_factory, instance_paths, ConsoleLogReader(logger_factory, instance_paths), syslog_ng_ctl
-        )
+    def __init__(self, syslog_ng_cli):
+        self.__syslog_ng_cli = syslog_ng_cli
 
     def start(self, config):
         self.__syslog_ng_cli.start(config)
--- a/tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
+++ b/tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
@@ -23,15 +23,16 @@
 
 from pathlib2 import Path
 from src.syslog_ng.syslog_ng_executor import SyslogNgExecutor
-
+from src.syslog_ng.console_log_reader import ConsoleLogReader
+from src.syslog_ng_ctl.syslog_ng_ctl import SyslogNgCtl
 
 class SyslogNgCli(object):
-    def __init__(self, logger_factory, instance_paths, console_log_reader, syslog_ng_ctl):
+    def __init__(self, logger_factory, instance_paths):
         self.__instance_paths = instance_paths
-        self.__console_log_reader = console_log_reader
+        self.__console_log_reader = ConsoleLogReader(logger_factory, instance_paths)
         self.__logger = logger_factory.create_logger("SyslogNgCli")
         self.__syslog_ng_executor = SyslogNgExecutor(logger_factory, instance_paths)
-        self.__syslog_ng_ctl = syslog_ng_ctl
+        self.__syslog_ng_ctl = SyslogNgCtl(logger_factory, instance_paths)
         self.__process = None
 
     # Application commands
--- a/tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
+++ b/tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
@@ -26,13 +26,15 @@ from src.syslog_ng.syslog_ng_executor import SyslogNgExecutor
 from src.syslog_ng.console_log_reader import ConsoleLogReader
 from src.syslog_ng_ctl.syslog_ng_ctl import SyslogNgCtl
 
+
 class SyslogNgCli(object):
-    def __init__(self, logger_factory, instance_paths):
+    def __init__(self, logger_factory, instance_paths, testcase_parameters):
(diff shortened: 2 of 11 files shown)


kira-syslogng:
Build FAILURE

mitzkia:

kira-syslogng:
Build SUCCESS

furiel:
We discussed with @mitzkia IRL, that we see two scenarios for valgrind execution.
- Every test should be run with valgrind enabled. With a well written suppression file, that is a very strong test suite, however it severely affects test suite execution. So we are thinking this version to be executed once in a day in a nightly build, and not per pull request.
- Specific tests executed with valgrind, for smoke check, or for explicit bug detection. For smoke check, there could be a giant test with all known drivers included, with a simple message sending and reload. That would be executed under valgrind per pull request.

This patchset intends to address the first category when every test would run under valgrind in a nightly build.

tests/pytest_framework/src/syslog_ng/syslog_ng.py
```diff
@@ -26,9 +26,10 @@
 
 
 class SyslogNg(object):
-    def __init__(self, logger_factory, instance_paths, syslog_ng_ctl):
+    def __init__(self, logger_factory, instance_paths, syslog_ng_ctl, valgrind_usage):
```
furiel: Adding valgrind_usage to the SyslogNg constructor looks strange to me, which I would like to avoid.

As I see:
We have 3 objects now: SyslogNg, SyslogNgCtl, and SyslogNgCli.

SyslogNg needs SyslogNgCli, because the start, stop etc messages needs to be forwarded to syslog-ng cli. But SyslogNg does not need SyslogNgCtl.

SyslogNgCli needs SyslogNgCtl, because some commands need to use the control socket.

So my proposal is, we should inject `syslog_ng_cli` instead of `syslog_ng_ctl` to `SyslogNg`. So syslog_ng_cli would be created `__register_instance`. Creating syslog_ng_ctl, should be moved into the constructor of `SyslogNgCli` from `__register_instance`. The `new_syslog_ng_ctl` should be removed.

valgrind_usage would be parameter of the syslog_ng_cli constructor instead. But I still feel this boolean parameter too specific, instead I would inject the whole `testcase_parameters` variable, and read the valgrind state through that.

tests/pytest_framework/src/syslog_ng/syslog_ng.py
```diff
@@ -46,3 +47,6 @@ def restart(self, config):
 
     def get_version(self):
         return self.__syslog_ng_cli.get_version()
+
+    def get_instance_paths(self):
```
furiel: This is unused code. Probably leftover from the splitting?

kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -91,7 +93,7 @@ def reload(self, config):
 
     def stop(self, unexpected_messages=None):
         self.__logger.info("Beginning of syslog-ng stop")
-        if self.__process:
+        if self.__process and not self.__core_file_exist:
```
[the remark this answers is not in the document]
mitzkia: ok, +1

tests/pytest_framework/src/syslog_ng/syslog_ng.py
mitzkia: Ok, agreed with the changing the parameter list of SyslogNg, SyslogNgCtl, and SyslogNgCli classes. 
Also agreed with removing syslog_ng_ctl object from the hashmap in register_instance() .
Also ok for passing testcase_parameters instead of a specific value.

mitzkia:
@furiel Thanks again for the review comments, I am agreed with them. Fix is uploaded. Removing WIP.

mitzkia:
I have added 1 new (middle) commit -> there were some refactors. The commit which contains the pure valgrind support is goes up to this.

kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -112,3 +112,12 @@ def stop(self, unexpected_messages=None):
...
         self.__console_log_reader.dump_stderr()
+        self.__handle_core_file(self.__process)
+
+    def __handle_core_file(self, process):
+        if process.wait(1) != 0:
+            for core_file in Path(".").glob("*core*"):
+                self.__process = None
+                self.__syslog_ng_executor.get_backtrace_from_core(core_file=str(core_file))
+                core_file.replace(Path(self.__instance_paths.get_working_dir(), core_file))
+            raise Exception("syslog-ng core file found and processed")
```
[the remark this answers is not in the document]
mitzkia: Of course, agreed. Soon will be fixed

kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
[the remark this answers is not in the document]
mitzkia: Thanks for the note, I have fixed it.

kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -112,3 +112,15 @@ def stop(self, unexpected_messages=None):
     # Helper functions
     def __error_handling(self):
         self.__console_log_reader.dump_stderr()
+        self.__handle_core_file(self.__process)
+
+    def __handle_core_file(self, process):
```
alltilla: Maybe `process` is not needed, as it can be acquired from `self`.

mitzkia: @alltilla thx for the finding, fixed it.

kira-syslogng:
Build SUCCESS

PR #2455, opened 2018-12-09
cfg-parser: printing error positions in case of parse failure

furiel:
I would like to write a [flycheck](https://www.flycheck.org/en/latest/) module for syntax checking syslog-ng configuration. For that, I need to find the offending line along with the error message from the output of syslog-ng -s.

The current syntax check printout looks awesome and helps users visually detect the problem quickly. But to find the problem algorithmically it is a little challenging because the error position does not occur explicitely, only implicitely (for example: to find first line number, one needs to look for an arrow). 
```
Error parsing log statement, syntax error, unexpected KW_DESTINATION, expecting ';' in ../etc/tmp.conf:
1       @version: 3.18
2       
3       log {
4         source { stdin(flags(no-parse)); };
5         parser {}
6----->   destination { file(/dev/stdout template('$(list-nth $LEVEL_NUM "a,b,c,d")')); };
6----->   ^^^^^^^^^^^
7       };
```

This patch adds the positional parameters to the end of the error reporting line, so that a script could parse it more easily.

Example:
```
Error parsing log statement, syntax error, unexpected KW_DESTINATION, expecting ';' in ../etc/tmp.conf:6:3-6:14:
```

### DIFF
--- a/lib/cfg-parser.c
+++ b/lib/cfg-parser.c
@@ -334,7 +334,11 @@ report_syntax_error(CfgLexer *lexer, YYLTYPE *yylloc, const char *what, const ch
         }
       else
         {
-          fprintf(stderr, "Included from %s:\n", from->name);
+          fprintf(stderr, "Included from %s:%d:%d-%d:%d:\n", from->name,
+                  from->lloc.first_line,
+                  from->lloc.first_column,
+                  from->lloc.last_line,
+                  from->lloc.last_column);
         }
       if (from->include_type == CFGI_FILE)
         {
--- a/lib/cfg-parser.c
+++ b/lib/cfg-parser.c
@@ -323,10 +323,14 @@ report_syntax_error(CfgLexer *lexer, YYLTYPE *yylloc, const char *what, const ch
     {
       if (from == level)
         {
-          fprintf(stderr, "Error parsing %s, %s in %s:\n",
+          fprintf(stderr, "Error parsing %s, %s in %s:%d:%d-%d:%d:\n",
                   what,
                   msg,
-                  yylloc->level->name);
+                  from->lloc.level->name,
+                  from->lloc.first_line,
+                  from->lloc.first_column,
+                  from->lloc.last_line,
+                  from->lloc.last_column);
         }
       else
         {


kira-syslogng:
Build SUCCESS

furiel:
> Do we want a user-friendly one line version when `first_line == last_line`?

I deliberately tried to trade user-friendliness for parsability. But if this version is too intrusive, I have alternatives in mind:
- For flycheck to work, I only need the first line number so what I did above is overkill. For me something similar would be enough
`[...] in ../etc/tmp.conf at line 6:`
I just added the `firstline:firstchar-lastline:lastchar` format in case someone else uses another "online" checker that can utilize these data (for example underlining offending parts similarly to syslog-ng).
- Do not touch the current output. Instead I can add a new option: like `--porcelain` similarly for `git log/status` etc, that uses an even more scriptable format. For example message should be the last part of the line. Something like this:
`error:filename:firstline:firstchar:lastline:lastchar:message string`.

Kokan:
If the error is not in the top level configuration file, the syslog-ng prints the *stack* configuration:

```
Error parsing source statement, source plugin something not found in block source default-network-drivers() at /tmp/install/share/syslog-ng/include/scl/default-network-drivers/plugin.conf:24:5:1-5:10:
1       
2       #Start Block block source default-network-drivers() at /tmp/install/share/syslog-ng/include/scl/default-network-drivers/plugin.conf:24
3       
4       
5-----> something();
5-----> ^^^^^^^^^
6       	channel {
7       		source {
8       			network(transport(tcp)
9       				port(514)
10      				log-msg-size(65536)

Included from /tmp/install/etc/syslog-ng.conf:
10      	system();
11      	internal();
12      };
13      
14      source s_network {
15----> 	default-network-drivers(
15----> 	^^^^^^^^^^^^^^^^^^^^^^^^^
16----> 		# NOTE: TLS support
17----> 		#
18----> 		# the default-network-drivers() source driver opens the TLS
19----> 		# enabled ports as well, however without an actual key/cert
20----> 		# pair they will not operate and syslog-ng would display a


contact: https://lists.balabit.hu/mailman/listinfo/syslog-ng
```
The line and column information is printed for the lowest level: **scl/default-network-drivers/plugin.conf:24:5:1-5:10:**
While it does not print the same information for the other levels: **Included from /tmp/install/etc/syslog-ng.conf:** which maybe a good idea to do so.

furiel:
Thanks @Kokan, my flycheck module did not work indeed without the location reported in the include chain.

I pushed a commit to fix it, (and also edited the first commit to use `from->lloc` instead of `yylloc`, because I think it is more readable).

Here is the current printout:
```
Error parsing source statement, source plugin abc not found in block source default-network-drivers() at /home/furiel/workspace/test-syslogng/root/share/syslog-ng/include/scl/default-network-drivers/plugin.conf:24:5:9-5:12:
1       
2       #Start Block block source default-network-drivers() at /home/furiel/workspace/test-syslogng/root/share/syslog-ng/include/scl/default-network-drivers/plugin.conf:24
3       
4       
5----->         abc
5----->         ^^^
6       	channel {
7       		source {
8       			network(transport(tcp)
9       				port(514)
10      				log-msg-size(65536)

Included from ../etc/tmp.conf:7:12-7:47:
2       
3       @include "scl.conf"
4       
5       log {
6         source { stdin(flags(no-parse)); };
7----->   source { default-network-drivers(dummy(abc)); };
7----->            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
8         parser {};
9         destination { file(/dev/stdout template('$(list-nth $LEVEL_NUM "a,b,c,d")')); };
10      };
```

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build SUCCESS

PR #2458, opened 2018-12-11
[WIP] Throttle: Change throttling to limit batches sent, not log lines sent

### LINKED ISSUE #2453: Enabling throttle with batching (HTTP) has unexpected behavior
# syslog-ng

## Version of syslog-ng
```
output of syslog-ng --version
attilaszakacs[ ~/install ]$ ./sbin/syslog-ng --version
syslog-ng 3 (3.18.1)
Config version: 3.18
Installer-Version: 3.18.1
Revision:
Compile-Date: Dec  6 2018 11:02:19
Module-Directory: /home/attilaszakacs/install/lib/syslog-ng
Module-Path: /home/attilaszakacs/install/lib/syslog-ng
Include-Path: /home/attilaszakacs/install/share/syslog-ng/include
Error opening plugin module; module='mod-java', error='libjvm.so: cannot open shared object file: No such file or directory'
Available-Modules: syslogformat,kvformat,map-value-pairs,geoip2-plugin,tfgetent,examples,afsocket,afamqp,http,basicfuncs,cef,stardate,afsmtp,afmongodb,appmodel,json-plugin,cryptofuncs,tags-parser,add-contextual-data,afuser,redis,system-source,disk-buffer,pseudofile,linux-kmsg-format,hook-commands,afsql,affile,afprog,mod-python,graphite,sdjournal,date,csvparser,geoip-plugin,afstomp,confgen,dbparser,snmptrapd-parser,riemann,xml
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: off
```

## Platform

```
attilaszakacs[ ~/install ]$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.1 LTS
Release:        18.04
Codename:       bionic
```

# Issue

## Failure

When I use the following config, and generate 13 messages with loggen, everything works as expected.
```
@version: 3.18
@include "scl.conf"

source s_network1 {
  network(ip("127.0.0.1") port(8888));
};

destination d_network1 {
  http(
    url("localhost:9999")
    batch-lines(5)
    batch-timeout(1000)
  );
};

log {
  source(s_network1);
  destination(d_network1);
};
```
First, two HTTP requests are sent instantly, with 5 log lines batched each. Then there is only 3 logs in the queue, so syslog-ng waits 1000 ms, to possibly gather more, then when the 1000 ms runs out, it sends the 3 batched logs in one request.
So I have:
* two requests at t=0s, with 5 logs batched, each
* one request at t=1s, with 3 logs batched

BUT, I would like to limit the HTTP requests 1 request/second, as let's say, my destination does not accept more, and the first two requests are sent immediately with this config, so that violates the requests/sec limit.
Let's turn on throttle with `throttle(1)`.
```
@version: 3.18
@include "scl.conf"

source s_network1 {
  network(ip("127.0.0.1") port(8888));
};

destination d_network1 {
  http(
    url("localhost:9999")
    throttle(1)
    batch-lines(5)
    batch-timeout(1000)
  );
};

log {
  source(s_network1);
  destination(d_network1);
};
```
I expect to have:
* one request at t=0s, with 5 logs batched
* one request at t=1s, with 5 logs batched
* one request at t=2s, with 3 logs batched

But instead, I have: 
* one request at t=0s, with 2 logs batched
* one request at t=2s, with 2 logs batched
* one request at t=4s, with 2 logs batched
* ...

I might have misunderstood the functionality of the `batch-lines` and `throttle` options. If I did, sorry for that :(

## Steps to reproduce
To make it easy to reproduce, I used a python based HTTP server, which always responds 200 to POST requests. It logs the responds sent, so it is easy to see, when did a request arrive.

1. start syslog-ng with the given config: `./sbin/syslog-ng -f throttle.conf -F -edtv`
2. start the HTTP server: `./http.py`
3. generate messages with loggen: `./bin/loggen -n 13 -i 127.0.0.1 8888`
4. check the output of the HTTP server to see the timestamps of the requests arrived
5. check the output of syslog-ng to see the timestamps and the payload of the requests sent

I attach these in a tar.gz:
* `http.py`: the HTTP server's code
* `httpserver.log`: the HTTP server's output
* `nothrottle.log`: the syslog-ng's output with no throttle enabled
* `throttle1.log`: the syslog-ng's output with throttle(1)
* `throttle.conf`: the config

[throttle.tar.gz](https://github.com/balabit/syslog-ng/files/2656725/throttle.tar.gz)

alltilla:
This PR intends to fix #2453.

The LogQueue's throttle_buckets variable counts how many logs can be popped from the queue, and its throttle variable is responsible for calculating the right amount of sleep, so we do not exceed the 
message / sec limit.
Currently these two variables are equal, so throttle limits for log lines sent, and not batches sent, which is okay, when we send 1 log line / batch, but if we batch several log lines to 1 batch it causes unexpected behavior.
In case of batching, the initial value of the LogQueue's throttle_buckets should be batch_lines * throttle,  so the LogQueue can grab enough log lines for the batch, but throttle should be intact, because the sleep is calculated from it.
However LogQueue should not know about batch_lines, but we can add an option to have a multiplicant relation between throttle and throttle buckets, so the unchanged throttle will cause the right amount of sleep, and the multiplied throttle_buckets will provide the right amount of log lines, which will be batched later in the Threaded Destination's _perform_insert() method.

For the reviewers:
1. Is it more logical for you as well, to limit the batches sent (e.g.: number of HTTP requests) / sec, instead of log lines sent / sec?
2. Is this the right way to provide enough log lines to the driver to batch?
3. Is `logs_per_throttle_bucket` name understandable?
4. Is it okay to set the LogDestDriver's `logs_per_throttle_bucket` attribute in `log_threaded_dest_driver_set_batch_lines()`?
5. How should we test this?

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -1192,7 +1192,7 @@ threaded_dest_driver_option
           log_threaded_dest_driver_set_max_retries(last_driver, $3);
         }
         | KW_BATCH_LINES '(' nonnegative_integer ')' { log_threaded_dest_driver_set_batch_lines(last_driver, $3); }
-        | KW_BATCH_TIMEOUT '(' positive_integer ')' { log_threaded_dest_driver_set_batch_timeout(last_driver, $3); }
+        | KW_BATCH_TIMEOUT '(' nonnegative_integer ')' { log_threaded_dest_driver_set_batch_timeout(last_driver, $3); }
         | dest_driver_option
         ;
 
--- a/lib/driver.c
+++ b/lib/driver.c
@@ -254,7 +254,7 @@ log_dest_driver_acquire_queue_method(LogDestDriver *self, const gchar *persist_n
   if (!queue)
     {
       queue = log_queue_fifo_new(self->log_fifo_size < 0 ? cfg->log_fifo_size : self->log_fifo_size, persist_name);
-      log_queue_set_throttle(queue, self->throttle);
+      log_queue_set_throttle(queue, self->throttle, self->logs_per_throttle_bucket);
     }
   return queue;
 }
@@ -365,6 +365,7 @@ log_dest_driver_init_instance(LogDestDriver *self, GlobalConfig *cfg)
   self->release_queue = log_dest_driver_release_queue_method;
   self->log_fifo_size = -1;
   self->throttle = 0;
+  self->logs_per_throttle_bucket = 1;
 }
 
 void
--- a/lib/driver.h
+++ b/lib/driver.h
@@ -167,6 +167,7 @@ struct _LogDestDriver
 
   gint log_fifo_size;
   gint throttle;
+  gint logs_per_throttle_bucket;
   StatsCounterItem *queued_global_messages;
 };
 
--- a/lib/logqueue.c
+++ b/lib/logqueue.c
@@ -190,7 +190,8 @@ log_queue_check_items(LogQueue *self, gint *timeout, LogQueuePushNotifyFunc para
            * rounded to zero when only a minimal interval passes between
            * poll iterations.
            */
-          self->throttle_buckets = MIN(self->throttle, self->throttle_buckets + new_buckets);
+          self->throttle_buckets = MIN(self->throttle, self->throttle_buckets + new_buckets)
+                                   * self->logs_per_throttle_bucket;
           self->last_throttle_check = now;
         }
       if (num_elements && self->throttle_buckets == 0)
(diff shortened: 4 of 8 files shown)


kira-syslogng:
Build SUCCESS

alltilla:
This PR would not fix the issue in every scenario, we will get back to this with another approach later, if we decide to cleanup some code, related to threaded destination, http destination, ... In the meantime the current behavior works, as it does not violate the msg/sec limit, although not so efficient with batching.

PR #2464, opened 2018-12-13
dbld: release target should use the default image

furiel:
For now it uses xenial, instead of ubuntu-xenial, so releasing goes off track due to using old image. With this change, it will use the default image, which is ubuntu-bionic this time of writing.


kira-syslogng:
Build SUCCESS

PR #2465, opened 2018-12-13
configure: fix "--with-docbook*" option usage

gaborznagy:
"--with-docbook-dir" option changed to "--with-docbook"

In dbld and in travis the "--with-docbook=" option is used, which does not exist,
therefore we always use the online version.

At first it seems trivial to change travis, dbld, etc to use the actual option,
although the config option name is misleading.
According to `xsltproc` man page, it is a file uses as an XML STYLESHEET.

Therefore I've changed the configure option and not the users of the option.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -320,10 +320,10 @@ AC_ARG_WITH(python,
               [  --with-python=VERSION    Build with a specific version of python]
               ,,with_python="auto")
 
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

gaborznagy:
OSE 3.19 released, I have removed the WIP flag.

PR #2467, opened 2018-12-17
travis: refactor - use brew and apt addon instead of shell execution

stephengroat:


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

gaborznagy:
Hi @stephengroat!

Thank you for your contribution!
Can I ask you to give some background information about the patch?
Please add it to the commit messages as well for future reference!

kira-syslogng:
Build SUCCESS

stephengroat:
@gaborznagy no worries, added a bit in the comment

travis `addons` seem to be better for speed and cleanliness

kira-syslogng:
Build SUCCESS

furiel:
I agree the travis file is more readable. It also modifiles the linux versions, which also look great (those changes might have gone to a separate patch?).

However, I do not see the speed benefit.
The current-ish master: https://travis-ci.org/balabit/syslog-ng/builds/469395508
osx execution time: 21min 19sec
With this PR:
https://travis-ci.org/balabit/syslog-ng/builds/469145573?utm_source=github_status&utm_medium=notification
osx execution time: 22min 45sec

So actually the execution time increased with more than a minute (above the 5% margin error).

@stephengroat do you happen to know the reason? Can you recheck the patch?

Also, the runtime of style-check build is doubled, because now we install packages that are unnecessary from style-check point of view. But probably this can be traded for readibility (1min 45s is still not much)

PS. I like the changes, it just does not help with the speed as intended.

stephengroat:
@furiel sorry about that.

the travis `homebrew` `addon` uses a specific ruby version (see https://travis-ci.org/balabit/syslog-ng/jobs/469145580#L101 and https://travis-ci.org/balabit/syslog-ng/jobs/469145580#L1211) that's supposed to be faster, but it's definitely more reliable. if you look at (https://github.com/travis-ci/travis-ci/issues/8552), you can see a previous issue where an osx image update caused `brew` to break because of ruby versioning. this extension was meant to lock in a tested version of ruby for `brew` to increase reliability (and speed). too bad the speed isn't increasing (has on some of my builds), but the reliability _should_ be better (no promises though!)

also, if desired, in the https://github.com/balabit/syslog-ng/pull/2467/files#diff-354f30a63fb0907d4ad57269548329e3R111 block, it is possible to override the `addons` block with a seperate one to avoid installing unnecessary packages. i don't think the override is worth it, but that's just one opinion. i'll clearly default to you on this, if it is, just left me know and i can override with something like this:
```
 - env: B=style-check
      dist: trusty
      sudo: required
      addons:
        apt:
          sources:
            - sourceline: 'deb http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_14.04 ./'
              key_url: 'http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_14.04/Release.key'
          packages:
            - astyle
      before_script: astyle --version
```

furiel:
Fine for me. Let's just rephrase the PR title then. Something like:
`travis: refactor - use brew and apt addon instead of shell execution`

contrib/Brewfile
```diff
@@ -16,7 +16,8 @@ brew "openssl"
 # brew "pcre"
 
 brew "gradle"
-brew "python@2"
+brew "python", link: false
```
furiel: This line was missing from the original Brewfile. Why this line is needed?

PR #2469, opened 2018-12-18
Fix app-parser() per reload memory leak

bazsi:
The lexer didn't handle generators that were allocated dynamically for every invocation properly.

This patch adds reference counting for CfgBlockGenerators which makes the fix trivial. 

Note, this is not app-parser() related, this is a bug in the core, which app-parser() and potentially confgen triggers.


kira-syslogng:
Build SUCCESS

furiel:
system source suffers from the same problem which is also fixed by this patchset!

PR #2471, opened 2018-12-20
Websense parser

bazsi:
This branch adds a parser/application adapter for fixing up websense logs.

This branch also adds a trace message into subst() to make that easier to debug.

### DIFF
--- a/lib/rewrite/rewrite-subst.c
+++ b/lib/rewrite/rewrite-subst.c
@@ -62,7 +62,26 @@ log_rewrite_subst_process(LogRewrite *s, LogMessage **pmsg, const LogPathOptions
                                   &new_length);
   if (new_value)
     {
+      msg_trace("Performing subst() rewrite",
+                evt_tag_str("rule", s->name),
+                evt_tag_str("value", log_msg_get_value_name(s->value_handle, NULL)),
+                evt_tag_printf("input", "%.*s", (gint) length, value),
+                evt_tag_str("type", self->matcher_options.type),
+                evt_tag_str("pattern", self->matcher->pattern),
+                evt_tag_str("replacement", self->replacement->template),
+                log_pipe_location_tag(&s->super));
       log_msg_set_value(*pmsg, self->super.value_handle, new_value, new_length);
+    }
+  else
+    {
+      msg_trace("Performing subst() rewrite failed, pattern did not match",
+                evt_tag_str("rule", s->name),
+                evt_tag_str("value", log_msg_get_value_name(s->value_handle, NULL)),
+                evt_tag_printf("input", "%.*s", (gint) length, value),
+                evt_tag_str("type", self->matcher_options.type),
+                evt_tag_str("pattern", self->matcher->pattern),
+                evt_tag_str("replacement", self->replacement->template),
+                log_pipe_location_tag(&s->super));
     }
   g_free(new_value);
 }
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

scl/websense/plugin.conf
```diff
@@ -0,0 +1,51 @@
...
+
+# Example log message (expected to be received via flags(no-parse))
+# <159>Dec 19 10:48:57 EST 10.203.28.21 vendor=Websense product=Security product_version=7.7.0 action=permitted severity=1 category=153 user=- src_host=10.64.134.74 src_port=62189 dst_host=mail.google.com dst_ip=74.125.224.53 dst_port=443 bytes_out=197 bytes_in=76 http_response=200 http_method=CONNECT http_content_type=- http_user_agent=Mozilla/5.0_(Windows;_U;_Windows_NT_6.1;_enUS;_rv:1.9.2.23)_Gecko/20110920_Firefox/3.6.23 http_proxy_status_code=200 reason=- disposition=1034 policy=- role=8 duration=0 url=https://mail.google.com
+
+block parser websense-parser(prefix(".websense.")) {
+    channel {
+        rewrite {
+            # normal BSD timestamp, plus a timezone code. Remove the
+            # timezone information for now.
+            subst('([A-Za-z]{3} [0-9 ]\d \d{2}:\d{2}:\d{2}) [A-Z]{3,4}' "$1 " flags(global) value("MSG"));
```
furiel: optional: `flag(global)` is not needed. We need to replace only the first occurrence.

scl/Makefile.am
```diff
@@ -25,6 +25,7 @@ SCL_SUBDIRS	= \
 	sudo		\
 	syslogconf	\
 	system		\
+	websense	\
```
furiel: Please add this to scl/CMakeLists.txt too.

furiel: Please update packaging too:
packaging/debian/syslog-ng-core.install

lib/rewrite/rewrite-subst.c
```diff
@@ -62,8 +62,27 @@ log_rewrite_subst_process(LogRewrite *s, LogMessage **pmsg, const LogPathOptions
...
+                log_pipe_location_tag(&s->super));
       log_msg_set_value(*pmsg, self->super.value_handle, new_value, new_length);
     }
+  else
+    {
+      msg_trace("Performing subst() rewrite failed, pattern did not match",
+                evt_tag_str("rule", s->name),
+                evt_tag_str("value", log_msg_get_value_name(s->value_handle, NULL)),
+                evt_tag_printf("input", "%.*s", (gint) length, value),
+                evt_tag_str("type", self->matcher_options.type),
```
furiel: optional: I would try to avoid duplication. A new function might be suitable: `rewrite_subst_trace(const gchar *str, LogRewriteSubst *self, gint length, const gchar *value);`

scl/websense/plugin.conf
bazsi: thanks for spotting this. fixed.

scl/Makefile.am
bazsi: done

bazsi: done.

lib/rewrite/rewrite-subst.c
bazsi: I prefer that log messages are trivial to find using a simple grep, and also I find that arguments to a specific log message is a concept that might easily change and although it seems like code duplication, the concepts should be handled separately.

Also, this is not a pattern we do in general, and I am a bit opposed to it, if we did.

With that said I wouldn't do this now.

kira-syslogng:
Build SUCCESS

PR #2477, opened 2019-01-02
loggen: paramter handling

### LINKED ISSUE #2239: loggen does not start with -D /dev/log (2 arguments)
# syslog-ng

## Version of syslog-ng

```
since 3.16.1
```

# Issue

## Failure

It won't start because it excepts 3 argument, but in this case 2 should be enough.

## Steps to reproduce

1. start loggen with `-D /dev/log` arguments

## Input and output logs

```
error [loggen.c:main:454] please specify target and port at least
```

Just an idea but each plugin could register an option, and if that is supplied that plugin could be started, and such checks like it has every options could be made plugin/option specific.

norberttak:
Move the parameter count check into plugins.
This allow us to run plugin options which
doesn't need target and port port parameters.
Fixes #2239

### DIFF
--- a/tests/loggen/socket_plugin/socket_plugin.c
+++ b/tests/loggen/socket_plugin/socket_plugin.c
@@ -133,6 +133,20 @@ start(PluginOption *option)
       return;
     }
 
+  if (unix_socket_x)
+    {
+      if (!option->target)
+        {
+          ERROR("in case of unix domain socket please specify target parameter\n");
+          return;
+        }
+    }
+  else if (!option->target || !option->port)
+    {
+      ERROR("in case of TCP or UDP socket please specify target and port parameters\n");
+      return;
+    }
+
   DEBUG("plugin (%d,%d,%d,%d)start\n",
         option->message_length,
         option->interval,
--- a/tests/loggen/ssl_plugin/ssl_plugin.c
+++ b/tests/loggen/ssl_plugin/ssl_plugin.c
@@ -127,6 +127,12 @@ start(PluginOption *option)
       return;
     }
 
+  if (!option->target || !option->port)
+    {
+      ERROR("please specify target and port parameters\n");
+      return;
+    }
+
   DEBUG("plugin (%d,%d,%d,%d)start\n",
         option->message_length,
         option->interval,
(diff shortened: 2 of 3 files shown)


kira-syslogng:
Build SUCCESS

tests/loggen/socket_plugin/socket_plugin.c
```diff
@@ -308,9 +318,19 @@ active_thread_func(gpointer user_data)
...
+      if (!option->target)
+        ERROR("please specify target parameter\n");
+
+      fd = connect_unix_domain_socket(sock_type, option->target);
+    }
   else
-    fd = connect_ip_socket(sock_type, option->target, option->port, option->use_ipv6);
+    {
+      if (!option->target || !option->port)
+        ERROR("please specify target and port parameters\n");
```
MrAnno: This validation should not be duplicated, it has nothing to do with active or idle threads IMHO.

The validation is based only on command line parameters and flags (`unix_socket_x` for example), so they can be checked before initializing threads/other functionality.

`start()` might be a better place, but adding a new method to the plugin interface is also an option (you already have `get_options_list`, something like `bool validate_options(PluginOption, GError**)` looks good to me).

tests/loggen/ssl_plugin/ssl_plugin.c
```diff
@@ -296,6 +299,9 @@ active_thread_func(gpointer user_data)
   PluginOption *option = thread_context->option;
 
   char *message = g_malloc0(MAX_MESSAGE_LENGTH+1);
+
+  if (!option->target || !option->port)
+    ERROR("please specify target and port parameters\n");
```
MrAnno: See my last comment.

tests/loggen/socket_plugin/socket_plugin.c
norberttak: I moved the check into start function

tests/loggen/ssl_plugin/ssl_plugin.c
norberttak: I moved the check into start function

kira-syslogng:
Build SUCCESS

tests/loggen/ssl_plugin/ssl_plugin.c
MrAnno: Thanks.

PR #2478, opened 2019-01-03
elasticsearch2: fix bulk send for client-mode("http")

furiel:
The motivation of this patchset is to fix https://github.com/balabit/syslog-ng/issues/1993.

Changes:
- flush-timeout() option is added for elasticsearch2 client-mode("http")
- retry mechanism is applied for unsuccessful batch sends
- Fix class loading problem of elasticsearch when compiled with cmake
- `void onMessageQueueEmpty` method in java binding is removed. It is replaced by `int flush()`. API breaking change.
- `boolean send()` method is replaced in java binding is replaced with `int send()`. API breaking change.
- batch-size based flush logic is pulled up into threaded destination driver framework

Notes to reviewers.
- Prior this patchset, the individual destination drivers were responsible for flushing when a batch-lines is reached. I pulled this logic up into logthrdestdrv level, because the batch size accounting is already done there, and this way it is not necessary to inject the current batch size into the individual drivers. With elasticsearch, this led to a major simplification, because a separate java module was responsible for batching. But it also lead to minor simplification in http, afsql, and riemann.
- When pushing down the int send() signature change and int flush() in from `class LogDestination` to `interface EsMessageProcessor`, I separated each layer changes into separate commits. As a result, changes done in a previous patch are reverted in the next patch, but I hope it will not cause much confusion. It gave me a lot of confidence and control over pushing through the changes step by step.
- First I created a separate class for the `boolean send() -> int send()` change to avoid api break as much as possible, which resulted in a little code duplication and strange sounding classes like `class ThreadedStructuredLogDestination`. The api breaking changes are put into the last two patches. If we decide that we want to maintain the api, then I just need to drop those two patches.
(PS: I experimented with java reflection, so that maybe the `Threaded...LogDestination` and `...LogDestination` classes can be merged without breaking the api, but unfortunately getMethod() could not catch the methods of subclasses, so I could not detect with it, if a user provided a send method with int or boolean return value.)

As many modules affected (hdfs, elasticsearch, elasticsearch2, kafka, afsql, http, riemann), I would like to ask for explicit approve form the testers.


furiel:
@kira-syslogng test this please test branch=elastic-bulk;

Kokan:
Also fixes #1095

bazsi:
This pull request **fixes 1 alert** when merging 8eddf1c29f2d0ef7e6be571a2d02b9fa8b37e745 into acc44a062c998c93ced9c9339e923b93084ed5ec - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-962092bf64e6ed57fa583ecde5be137f057537e2)

**fixed alerts:**

* 1 for FIXME comment

---

kira-syslogng:
Build SUCCESS

modules/java/proxies/java-destination-proxy.c
```diff
@@ -145,6 +145,12 @@ __load_destination_object(JavaDestinationProxy *self, const gchar *class_name, c
                                                    self->dest_impl.mi_constructor, handle);
   if (!self->dest_impl.dest_object)
     {
+      jthrowable exc = CALL_JAVA_FUNCTION_VOID(java_env, ExceptionOccurred);
+      if (exc)
+        {
+          CALL_JAVA_FUNCTION_VOID(java_env, ExceptionDescribe);
```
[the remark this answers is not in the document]
furiel: I was experimenting with that, but in the end I could not find a solution after putting a small effort into it.

I found a function call (I don't remember now which was that, but I can regoogle that if interesting) that could extract the error message from the exception, that needed to be converted to utf8 string through java (it was a little bit complex code), and also the result was not satisfactory, because the stacktrace part was missing.

In my case the exception came from the framework itself (missing jar files from make install), and not triggered from user code. Maybe this could be triggered from user code, but users can write their code with appropriate logging that they do not need to rely on us for do so. That's why I did not add to much importance to the matter.

So I left this change in the patchset as good enough. But of course it can probably improved into an msg_error() message later.

modules/java/src/main/java/org/syslog_ng/LogThrDestDrvConsts.java
```diff
@@ -0,0 +1,32 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+package org.syslog_ng;
+public abstract interface LogThrDestDrvConsts {
```
lbudai: * use [java enums](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html) and
* use [static imports](https://docs.oracle.com/javase/1.5.0/docs/guide/language/static-import.html)
(I'd call it `WorkerInsertResults`.)
You can import enum members statically, eg.: 
```java
import static org.syslog_ng.WorkerInserResultCodes.*;
```
In this way you can use these consts the same way then currently but without this weird interface-usage.

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
```diff
@@ -53,10 +53,11 @@
 import java.security.KeyStoreException;
 import java.io.FileInputStream;
 import org.apache.log4j.PropertyConfigurator;
+import org.syslog_ng.LogThrDestDrvConsts;
 
 import java.io.IOException;
 
-public class ESHttpClient implements ESClient {
+public class ESHttpClient implements ESClient, LogThrDestDrvConsts {
```
lbudai: By replacing `LogThrDestDrvConsts` with a statically imported enum class, you can remove this extra interface.

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/messageprocessor/http/HttpBulkMessageProcessor.java
```diff
@@ -57,27 +52,20 @@ private boolean _flush() {
...
 			logger.error(e.getMessage());
-			return false;
+			retval = false;
 		}
 		if (! jestResult.isSucceeded()) {
 			logger.error(jestResult.getErrorMessage());
-			return false;
+			retval = false;
 		}
 		bulk = new Bulk.Builder();
```
lbudai: I think, we could extend `try-catch` with a `finally`, where we could assign a new `Bulk.Builder` to `bulk`·
(and in this case we can keep early returns)

lib/logthrdestdrv.c
```diff
@@ -329,6 +352,9 @@ _perform_inserts(LogThreadedDestWorker *self)
 
       _process_result(self, result);
 
+      if (self->enable_batching && self->batch_size == self->owner->batch_lines)
```
lbudai: shouldn't we check `self->batch_size >= self->owner->batch_lines`?

modules/afsql/afsql.c
```diff
@@ -857,7 +857,7 @@ _batch_lines(const AFSqlDestDriver *self)
 static inline gboolean
 afsql_dd_should_commit_transaction(const AFSqlDestDriver *self)
```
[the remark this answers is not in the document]
furiel: _batch_lines still needed in the init method of afsql, to initialize the default batch_lines variable to custom value (which is different than the default value of logthrdestdrv).

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/messageprocessor/ESMessageProcessor.java
```diff
@@ -24,5 +24,5 @@
 package org.syslog_ng.elasticsearch_v2.messageprocessor;
 
 public interface ESMessageProcessor {
```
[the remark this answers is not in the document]
furiel: Thanks for the comment. I implemented this, and the patchet is simplified considerably.

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/client/http/ESHttpClient.java
furiel: I resolve this, because in the end I could eliminate the whole class.

modules/java-modules/elastic-v2/src/main/java/org/syslog_ng/elasticsearch_v2/messageprocessor/http/HttpBulkMessageProcessor.java
furiel: Indeed, it resulted in a much cleaner code. Thanks for the comment.

lib/logthrdestdrv.c
furiel: fixed thanks

modules/java/src/main/java/org/syslog_ng/LogThrDestDrvConsts.java
furiel: I resolve this because in the end I could eliminate the whole class.

furiel:
@kira-syslogng test this please test branch=elastic-bulk;

bazsi:
This pull request **fixes 1 alert** when merging c337706caf28d2d47b8b3aaec975aaafcb553d4e into 4e125dd2fdc7a11df8133dab0753dade9f8a6647 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-11a8fcf97599a68b880d0a37b5418dff4048aaaa)

**fixed alerts:**

* 1 for FIXME comment

---

modules/java/proxies/java-destination-proxy.c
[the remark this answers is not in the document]
Kokan: I do not think this is mandatory as; there is a general error about a missing constructor via internal log, so the user could be notified that something is wrong; this just gives an extra help for finding it out.
Maybe this could be moved to debug level; but I am fine with it as it is.

furiel:
@kira-syslogng test this please test branch=elastic-bulk;

kira-syslogng:
Build SUCCESS

furiel:
The regression passed for this codechange.

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng test this please test branch=elastic-bulk;

kira-syslogng:
Build SUCCESS

modules/java/native/java-destination.c
```diff
@@ -173,7 +173,15 @@ java_dd_deinit(LogPipe *s)
 gint
 java_dd_send_to_object(JavaDestDriver *self, LogMessage *msg)
 {
-  return java_destination_proxy_send(self->proxy, msg);
+  gint result = java_destination_proxy_send(self->proxy, msg);
+  if (result < 0 || result >= WORKER_INSERT_RESULT_MAX)
+    {
+      msg_error("java_destination: worker insert result out of range. Retrying message later",
```
Kokan: optional: the message could include destination location tag

furiel: Good idea, fixed thanks.

furiel:
@kira-syslogng test this please test branch=elastic-bulk;

bazsi:
This pull request **fixes 1 alert** when merging 1731ad7ba0e5b7e7c70dc60929d32a116963010d into 9554237671dca75107f8c071d7ba3b345195d5b9 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-9d84f9c426d7838941b46a4d086ab55dcb2ed9f1)

**fixed alerts:**

* 1 for FIXME comment

---

kira-syslogng:
Build SUCCESS

PR #2482, opened 2019-01-07
lib/compat: replace strcasestr() implementation

bazsi:
This patch removes the old strcasestr() implementation and replaces it
with the one in NetBSD, which uses the 3 clause BSD license.

### DIFF
--- a/.astylerc
+++ b/.astylerc
@@ -27,3 +27,4 @@
 --exclude="lib/jsonc"
 --exclude="modules/afamqp/rabbitmq-c"
 --exclude="modules/afmongodb/mongo-c-driver"
+--exclude="lib/compat/strcasestr.c"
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

gaborznagy:
I see it has style-check problems and copyright error as well.
I don't know the reason for the latter.

bazsi:
I have tried to resolve the copyright issue, with no success so far. I was
battling with the sed magic there.

With the style issues I would rather ignore them, this is ported code after
all. Is there a way for astyle issues to be ignored or a file be excluded
from scanning?

MrAnno:
The `exclude` option in `.astylerc` can be used to exclude directories (and files as well).

https://github.com/balabit/syslog-ng/blob/4e125dd2fdc7a11df8133dab0753dade9f8a6647/.astylerc#L26

kira-syslogng:
Build SUCCESS

PR #2483, opened 2019-01-07
timeutils refactor

bazsi:
This branch is a second set of patches to refactor time handling code in syslog-ng to be much more readable and easier to change.

The branch achieves a number of things:
* it splits out timeutils.c to a set of independent modules in the lib/timeutils directory
* it adds a thin abstraction layer around "struct tm" called WallClockTime
* it also introduces UnixTime that is a more generic version of the previous LogStamp structure
* it centralizes all functions around the new abstractions and refactors all users to use the new interfaces (templates, date-parser, snmptrapd-header-parser, syslog-format)
* contains a number of bugfixes

Bugfixes:
* the P_ macros were using an invalid value in case the PROCESSED timestamp was not initialized (which is not before serialization)
* Pythonv2 we were using the GMT as timezone, instead of the local timezone

~~I still want to add low-level unit tests, so far I was only moving code around and we have some coverage through the existing test suite. It would be much better to test the functionality through the low-level interfaces.~~ 

~~I am opening this PR to show how it is progressing and to solicit feedback.~~

**I have now added unit tests for UnixTime and WallClockTime** With those this PR should be complete.

PS: the patchset seems huge, but it is broken down to tiny little patches and each should only make one thing that is trivial to review. I was making an effort to even compile patches one-by-one, which I may have broken in the last round of rebases, but should be easy to fix.


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/timeutils/wallclocktime.h
```diff
@@ -0,0 +1,128 @@
...
+        .tm_sec = -1,			\
+        .tm_isdst = -1,			\
+      }, 				\
+      .wct_gmtoff = -1,			\
+      .wct_zone = NULL,			\
+      .wct_usec = 0,			\
+  }
+
+#endif
+
```
lbudai: Nice.

lib/logmsg/tests/test_logmsg_serialize.c
```diff
@@ -239,12 +239,11 @@ Test(logmsg_serialize, simple_serialization)
 
   log_msg_deserialize(msg, sa);
 
-  LogStamp ls =
-  {
-    .tv_sec = msg->timestamps[LM_TS_RECVD].tv_sec,
-    .tv_usec = msg->timestamps[LM_TS_RECVD].tv_usec,
-    .zone_offset = msg->timestamps[LM_TS_RECVD].zone_offset
-  };
+  LogStamp ls;
```
lbudai: Why did you drop designated initializers? (I think that initialization way is superior).

lib/timeutils/cache.h
```diff
@@ -25,13 +25,33 @@
...
+static inline time_t
+cached_mktime_wct(WallClockTime *wct)
+{
+  return cached_mktime(&wct->tm);
+}
+
+static inline void
+cached_gmtime_wct(time_t *when, WallClockTime *wct)
+{
+  return cached_gmtime(when, &wct->tm);
```
lbudai: Remove `return`.

lib/logmsg/tests/test_logmsg_serialize.c
bazsi: I agree, the reason that the UnixStamp was a member of LogStamp, but at the end I've decided against it, and alias LogStamp to UnixTime, with that the initializers could come back.

So agreed, thanks for catching this.

bazsi: I've sent a followup patch to revert this piece and go back to C99 initializers.

kira-syslogng:
Build FAILURE

mitzkia:

kira-syslogng:
Build SUCCESS

lib/timeutils/cache.h
bazsi: fixed.

bazsi:
I have squashed all changes back to the original commits.

There's still one open item, the one to eliminate LogStamp in favour of UnixTime. That I would do in the last round of patches, the one that will also do functional changes (e.g. the original aim of guessing the timezone value for incoming timestamps).

So right now we are:
1) waiting for one more review
2) and to check if this broke any of the broader functional tests.

kira-syslogng:
Build SUCCESS

bazsi:
I have squashed all changes back to the original commits.

There's still one open item, the one to eliminate LogStamp in favour of UnixTime. That I would do in the last round of patches, the one that will also do functional changes (e.g. the original aim of guessing the timezone value for incoming timestamps).

So right now we are:
1) waiting for one more review
2) and to check if this broke any of the broader functional tests.

lib/timeutils/unixtime.c
```diff
@@ -34,3 +37,39 @@ unix_time_unset(UnixTime *self)
...
+{
+  self->ut_gmtoff = wct->wct_gmtoff;
+
+  if (self->ut_gmtoff == -1)
+    self->ut_gmtoff = gmtoff_hint;
+  self->ut_usec = wct->wct_usec;
+
+  /* FIRST: We convert the timestamp as it was in our local time zone. */
+  gint unnormalized_hour = wct->wct_hour;
+  wct->wct_isdst = -1;
```
furiel: What is the meaning/purpose of this `wtc_isdst`?

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/timeutils/wallclocktime.h
```diff
@@ -0,0 +1,128 @@
...
+#define wct_sec   tm.tm_sec
+#define wct_isdst tm.tm_isdst
+  struct tm tm;
+
+
+/* We might need to separate tm_zone to a different conditional/autoconf
+ * check.  At least on Linux/FreeBSD they were introduced the same time.
+ * Some platforms may lack tm_zone, even though they have tm_gmtoff */
+
+#ifndef HAVE_STRUCT_TM_TM_GMTOFF
```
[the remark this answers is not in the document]
bazsi: I've fixed this, which uncovered a couple of suprises, but I am adding a comment at the end to give you more details.

lib/timeutils/wallclocktime.h
```diff
@@ -0,0 +1,128 @@
...
+  gint32 wct_gmtoff;
+  const char *wct_zone;
+#else
+#define wct_gmtoff tm.tm_gmtoff
+#define wct_zone tm.tm_zone
+#endif
+  gint32 wct_usec;
+} WallClockTime;
+
+#ifdef HAVE_STRUCT_TM_TM_GMTOFF
```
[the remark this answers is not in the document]
bazsi: fixed.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:
I have now concluded on all my pending changes, so this should be good to go in, once the tests pass.

I promised to include some details about a surprise I've uncovered when fixing the conditional compilation that @furiel found in one of the notes, so here it comes a summary of changes since that point.

* I have reverted to C99 initializers in test_logmsg_serialize as requested by @lbudai 
* removed unneeded returns from functions with a void return value in cache.h
* the wct_gmtoff field was converted to "long" as that's the type used in struct tm->tm_gmtoff, this cascaded down to stuff like our strptime() implementation
* I have introduced separate unix_time_set_from_wall_clock_time() functions that 1) change the incoming WallClockTime, 2) one that does not. This makes the mutation at least explicit. We are not using the non-normalized ones, but I've kept those in the code to make the distinction trivial. The tests do excercise the non-normalized ones.

I also had a change in time conversion during daylight saving transition hours, that I've abandoned at the end, so there's no behavior change at this point. This resulted in 2 more testcases though.

With all that said, I'm hopeful that travis agrees with me and yields green test result. I am also initiating an internal test run if I can :) I would be grateful for a timely review, this has become much longer than I've had hoped.

kira-syslogng:
Build SUCCESS

lbudai:
@bazsi : trying to review on today/tomorrow

kira-syslogng:
Build SUCCESS

bazsi:
Apart from the macos failure which is not specific to this PR, this has passed:
1) travis
2) kira
3) internal tests (http://jenkins.syslog-ng.balabit/job/pe7-MI-test-engine/894/)

It might make sense to add the first patch of #2517 here, as that still introduces incompatibilities by eliminating LogStamp completely. Should I pick that patch here, or let's do that in the next round?

lib/timeutils/unixtime.c
```diff
@@ -90,6 +90,8 @@ unix_time_set_from_normalized_wall_clock_time_with_tz_hint(UnixTime *self, WallC
                  - target_gmtoff;
 
   self->ut_gmtoff = target_gmtoff;
+  wct->wct_gmtoff = self->ut_gmtoff;
```
furiel: This function is like an assignment for me. But shouldn't we consider wct as const in this context? It looks very strange that the assignee is assigned back to the assigner.

bazsi: btw: I wrote a comment in the unixtime.h header as well, so it's not just a commit message.

kira-syslogng:
Build FAILURE

bazsi:

kira-syslogng:
Build SUCCESS

lib/timeutils/decode.c
```diff
@@ -228,25 +229,25 @@ __determine_recv_timezone_offset(LogStamp *const timestamp, glong const recv_tim
 }
 
 static void
-__fixup_hour_in_struct_tm_within_transition_periods(LogStamp *stamp, struct tm *tm, glong recv_timezone_ofs)
+__fixup_hour_in_struct_tm_within_transition_periods(LogStamp *stamp, WallClockTime *wct, glong recv_timezone_ofs)
 {
   /* save the tm_hour value as received from the client */
```
szemere: please update the comment with the new class member name

furiel:
@bazsi we started a review with @szemere and we reviewed up to commit
```
timeutils: add wall_clock_time_strptime() function
```
(by the order of gitk, github messed up the commit orders)
We do not have comment to that slice, except a minor one we think `stdint.h` needs not to be included. That slice is useful even on its own, it nicely splits up timeutils into smaller parts.

Can you split the PR there?

lib/timeutils/timeutils.h
```diff
@@ -28,6 +28,9 @@
 #include "syslog-ng.h"
 #include "compat/time.h"
 
+typedef struct _UnixTime UnixTime;
+typedef struct _WallClockTime WallClockTime;
```
furiel: We discussed about this IRL, but we still believe this is not good direction.

The original timeutils was split into three modules, UnixTime, WallClockTime, TimeUtils.
UnixTime contains conversion functions from WallClockTime to UnixTime, (besides the UnixTime structure).
WallClockTime contains the conversion functions from WallClockTime to UnixTime (besides the WallClockTime structure).
TimeUtils remains with some random leftover functions.

The problem is with this design is the circular dependency between UnixTime and WallClockTime due to the conversion functions. While timeutils does not mean much with these forward declaration on their own, as they are not used in timeutils at all.

These forward declarations workaround the problem by exploiting c language specific trick. However the design problem still remains.

The proper way to address is the conversion functions, or any other function that needs both class should reside in a third module. A good candidate for this would be timeutils. Instead UnixTime and WallClockTime includes timeutils, timeutils should include them. Thinkink timeutils become an adapter or a thin facade module in front of them.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

lib/timeutils/timeutils.h
bazsi: I had a number of discussions with @szemere and I have now added couple of patches on top, with a solution that I think addresses your concerns but the coupling between unixtime and wallclocktime remains.

The solution was to split timeutils.h into two:
1) misc.h to contain miscellaneous functions, that are concrete
2) timeutils-internal.h that contains typedef forward declarations.

unixtime and wallclocktime will then both include timeutils-internals.h. Users will include either unixtime or wallclocktime or both.

If something is primarily using unixtime, that module can work just by including unixtime.h, if something is primarily working with wallclocktime, you can do that as well.

There is an inherent coupling between the unix/wallclock time functions and albeit we could put them in the same module, I think it is better to separate them.

Also, I wouldn't consider forward type declarations a "trick". I agree that circular dependency is not a good thing by itself, _except_ when the concepts we are representing are deeply coupled in reality as well.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/timeutils/timeutils.h
furiel: I checked the changes and I still believe it is not a good idea.

The forward declarations are just cosmetic. UnixTime still includes WallClockTime, and WallClockTime still includes UnixTime. If the forward declarations were moved into the respective headers, we would have the same result with less headers. In the end, it is the header guards that avoid recursive include. I do not see what is the benefit of the forward declarations now.

> If something is primarily using unixtime, that module can work just by including unixtime.h

As unixtime includes wallclocktime, in the end those modules are not separated in practice.

For now all conversion functions are asymetric in the sense that we can decide where to put each one. However, as soon a function needs to know the internals of both types, then we could not find a good place for them.
As an example: lets say at some point we want to add a function that checks if a wallclocktime is equal to a unixtime. Where should one put that function? Into wallclocktime or into unixtime?
Which header should contain the declaration? unixtime.h or wallclocktime.h, or should be put into timeutils-internals.h instead?
I am afraid there is no good answer for those questions in this architecture.

lib/logmsg/timestamp-serialize.c
```diff
@@ -25,15 +25,15 @@
 #include "timestamp-serialize.h"
 
 static gboolean
-_write_log_stamp(SerializeArchive *sa, const LogStamp *stamp)
+_write_log_stamp(SerializeArchive *sa, const UnixTime *stamp)
```
furiel: I accidentally commented stg into the other PR, commenting here too to have everything in one place.

Actually, I believe LogStamp is a better name than unixtime. In case of unixtime, one would associate the "unix epoch", the number of seconds since 1970 january 01, however now that includes timezone and microseconds information too, which can be confusing.
Syslog-ng also has an unixtime macro, that is expanded into "unix epoch", and not printing the timezone. 

Using LogStamp as name would be more clear in my opinion. It would reduce the size of the diff too.

Related to this, the original logstamp module module was eliminated, and unixtime was put in place. That seems also unnecessary for me.

bazsi:
Hi,

On Sun, Feb 10, 2019, 14:35 furiel <notifications@github.com wrote:

> *@furiel* commented on this pull request.
>
> In lib/timeutils/timeutils.h
> <https://github.com/balabit/syslog-ng/pull/2483#discussion_r255337889>:
>
> > @@ -28,6 +28,9 @@
>  #include "syslog-ng.h"
>  #include "compat/time.h"
>
> +typedef struct _UnixTime UnixTime;
> +typedef struct _WallClockTime WallClockTime;
>
> I checked the changes and I still believe it is not a good idea.
>
> The forward declarations are just cosmetic. UnixTime still includes
> WallClockTime, and WallClockTime still includes UnixTime. If the forward
> declarations were moved into the respective headers, we would have the same
> result with less headers.
>
This is not true. The unixtime.h does not include wallclocktime.h and the
unixtime.h interface does not depend on the specific details of
WallClockTime.

If the headers would define their own typedefs (e.g. unixtime.h would
define the UnixTime typedef) then we couldn't define the wallclocktime
interface without including unixtime.h. I forward declare the types exactly
as I don't want to include one from the other.

> the end, it is the header guards that avoid recursive include.
>

No. It's not that. The headers dont include one another.

I do not see what is the benefit of the forward declarations now.
>
> If something is primarily using unixtime, that module can work just by
> including unixtime.h
>
> As unixtime includes wallclocktime, in the end those modules are not
> separated in practice.
>
You can work with unixtime without including wallclocktime as long as you
only use a pointer to a WallClockTime.


For now all conversion functions are asymetric in the sense that we can
> decide where to put each one. However, as soon a function needs to know the
> internals of both types, then we could not find a good place for them.
>

As long as the functions are asymmetric they can go to one or the other.

As an example: lets say at some point we want to add a function that checks
> if a wallclocktime is equal to a unixtime. Where should one put that
> function? Into wallclocktime or into unixtime?
>

If we check if a wallclocktime equals to a unixtime that goes to
wallclocktime. As long as we dont want the implementation as an inline
function in the header.

Which header should contain the declaration? unixtime.h or wallclocktime.h,
> or should be put into timeutils-internals.h instead?
> I am afraid there is no good answer for those questions in this
> architecture.
>
Again, I think you misunderstood that the headers include one another,
which they dont.
Only their implementation depends on both structures, the declarations of
either the types or functions dont.

If we need a declaration that needs both headers that will need to go to a
new header that depends on the current ones.

bazsi:
On Sun, Feb 10, 2019, 14:47 furiel <notifications@github.com wrote:

> *@furiel* commented on this pull request.
>
> In lib/logmsg/timestamp-serialize.c
> <https://github.com/balabit/syslog-ng/pull/2483#discussion_r255338325>:
>
> > @@ -25,15 +25,15 @@
>  #include "timestamp-serialize.h"
>
>  static gboolean
> -_write_log_stamp(SerializeArchive *sa, const LogStamp *stamp)
> +_write_log_stamp(SerializeArchive *sa, const UnixTime *stamp)
>
> I accidentally commented stg into the other PR, commenting here too to
> have everything in one place.
>
> Actually, I believe LogStamp is a better name than unixtime. In case of
> unixtime, one would associate the "unix epoch", the number of seconds since
> 1970 january 01, however now that includes timezone and microseconds
> information too, which can be confusing.
> Syslog-ng also has an unixtime macro, that is expanded into "unix epoch",
> and not printing the timezone.
>
The UNIXTIME macro also contains the microseconds part if you use
frac-digits(). If you get the current time with gettimeofday() you'll get
fractions of a second. Python returns fractions of a second if you query
the time using time.time().

So I think that qualifies fractions of the second to be part of the unix
idea of time.

The timezone is a bit more subtle, still you have the timezone offset as a
global variable in all programs, not to mention the TZ environment variable.

Also for any globally used application it makes sense to store the original
timezone value even if the application chooses to use epoch based time
representation. If you don't, you are losing information.

For example unix filesystems do not store the time zone for file
modification. Therefore if you plug the same file system into different
computers ls would report the time it was changed in the local timezone. Is
that the right thing to do? In a lot of cases it probably is, in other
cases remembering the time zone offset next to the epoch makes sense.

If you want to convert wall clock time to an epoch based timestamp, you'd
lose information. You can't convert it back unless you also specify the
timezone. This is not normally a problem in systems that primarily operate
in one timezone, as that is implicit.

Using LogStamp as name would be more clear in my opinion. It would reduce
> the size of the diff too.
>
I explicitly wanted to get away from the Log prefix. There's nothing that
is log specific there.

I am open to other names, but my goal was to make this decoupled from the
Log concept.


Related to this, the original logstamp module module was eliminated, and
> unixtime was put in place. That seems also unnecessary for me.
>
Since it's not just log related, it is better located in the timeutils
directory.

>

lib/logmsg/timestamp-serialize.c
furiel: Agreeing to https://github.com/balabit/syslog-ng/pull/2483#issuecomment-462163233, marking resolved.

lbudai:
@bazsi: out of scope: circular dependency is a programming language indepent issue, or a design problem. To resolve such an issue we usually have more than one way. Yes, we can use forward decl. but in this case initiating a facade-like “unit” that includes A and B is superior. In this case you can easily extend the functionality while you also break the circ. dependency. It’s also future proof if someone need to add new conversion to A and B, or if someone need to initiate a new type, C. 
I think this is why @furiel doesn’t like the forward decl.
(and another thing came into my mind: c++ provides virtual base classes to resolve some dependency issues but no one want use it in modern c++.
of course forward decls is better than virt. base classes, but the message is clear: try to change the mode to avoid circ. dependency issues :)

kira-syslogng:
Build FAILURE

bazsi:
Albeit I forgot to change the cmakelists files, I did make the changes that we agreed on, even if the automated tests are red. Ill fix them possibly in the evening.

lib/timeutils/conv.c
```diff
@@ -0,0 +1,112 @@
...
+
+  /* SECOND: adjust ut_sec as if we converted it according to our timezone. */
+  gint local_gmtoff = get_local_timezone_ofs(self->ut_sec);
+  if (target_gmtoff == -1)
+    {
+      target_gmtoff = local_gmtoff;
+    }
+  self->ut_sec = self->ut_sec
+                 + local_gmtoff
+                 - (normalized_hour - unnormalized_hour) * 3600
```
furiel: Can you explain why we need to repair with `- (normalized_hour - unnormalized_hour) * 3600`? `mktime` should deal with aligning calendar time due to normalization before returning. I think we do not need to repair here.

lib/template/macros.c
```diff
@@ -515,97 +515,96 @@ log_macro_expand(GString *result, gint id, gboolean escape, const LogTemplateOpt
...
-        zone_ofs = stamp->zone_offset;
-
-      t = stamp->tv_sec + zone_ofs;
-
-      cached_gmtime(&t, &tm_storage);
-      tm  = &tm_storage;
+      WallClockTime wct;
 
+      convert_unix_time_to_wall_clock_time_with_tz_override(&wct, stamp,
+                                                            time_zone_info_get_offset(opts->time_zone_info[tz], stamp->ut_sec));
```
furiel: The last parameter looks strange for me. According to the `convert_unix_time_to_wall_clock_time_with_tz_override` signature, that should be a `gmtoff_override` which should be related to timezone. But stamp->ut_sec is the calendar time itself. Shouldn't that be zero?

lib/timeutils/conv.c
bazsi: We do, as we are using local timezone specific functions (e.g. mktime) to convert timestamps in a remote timezone.

What we do is that we convert these timestamps and then cancel out any local timezone offsets. And the normalized/unnormalized dance is needed for daylight saving transition hours. The US changes dst at a different time than in Europe, so if mktime() uses that to change the local hour value, then that is incorrect.

Also, this branch explicitly does not want to change behavior, and this is the same as the one before.

furiel: Oh I see. I was thinking this is a new implementation, then I found `_adjust_tvsec_to_move_it_into_given_timezone` which worked exactly this way. I think I understand this one.

lib/template/macros.c
bazsi: The only use for stamp->ut_sec is the parameter for time_zone_info_get_offset () which will return a gmtoff value.

This actually forces a timestamp to be in a specific timezone, with its associated daylight saving rules.

kira-syslogng:
Build SUCCESS

lib/timeutils/unixtime.c
bazsi: with the new convert_XXX() functions this should be even more appearent, so resolving this conversation, reopen if you disagree.

lib/timeutils/decode.c
bazsi: I think this has been eliminated in the meantime. I even grepped the source for tm_hour and I found no occurences.

lib/timeutils/timeutils.h
bazsi: Albeit we've disagreed, I've changed this code so that the circular dependency is moved to a 3rd module.

bazsi:
I have now reordered the parameters for the conversion functions to make them more logical. I hope with that this branch is complete.

kira-syslogng:
Build SUCCESS

lib/template/macros.c
furiel: Looking at `convert_unix_time_to_wall_clock_time_with_tz_override` I found something strange:

In this call it has 4 parameters:
https://github.com/bazsi/syslog-ng/blob/2c17c0620f87a696aa4c2282df3ff286db8edd4b/lib/template/macros.c#L520

In the definition it has only 3:
https://github.com/bazsi/syslog-ng/blob/2c17c0620f87a696aa4c2282df3ff286db8edd4b/lib/timeutils/conv.c#L98

Which is strange because the tests should have detected that. Do I miss something?

bazsi:
I can see only 3 in that call too.

lib/template/macros.c
furiel: Oh right sorry!

kira-syslogng:
Build SUCCESS

lib/timeutils/tests/test_scan-timestamp.c
```diff
@@ -287,10 +289,6 @@ Test(parse_timestamp, cisco_timestamps)
 
 Test(parse_timestamp, rfc5424_timestamps)
 {
-  /* the timestamp '-' means the current time */
-  _expect_rfc5424_timestamp_eq("-", "2017-12-13T09:10:23.123+01:00");
```
furiel: As I see this functionality is still available, just pulled up into `log_msg_parse_timestamp`. Would it make sense to add a testcase for that to maintain coverage?

bazsi: I've pushed another patch on top, which adds a testcase to test_msgparse that covers the "-"
timestamp.

bazsi:
@furiel was kind enough to do a thorough review, we now need another reviewer. All known issues were addressed in the branch and it would be appreciated if I could move forward with the patches I have piled on top of this.

kira-syslogng:
Build SUCCESS

lib/timeutils/timeutils-internals.h
```diff
@@ -0,0 +1,31 @@
+/*
```
szemere: As a result of the - VERY appreciated - conv.c/h modification this file became futhile. (If I see it correctly, it is already removed from compiling, only the file left in the repo.)

PR #2487, opened 2019-01-10
python-destination: send can return worker_insert_result_t, flush support

furiel:
This patchset changes python destination, so that send can now return `worker_insert_result_t` instead of bool. This fixes almost everything (except from `isOpened` problem) from https://github.com/balabit/syslog-ng/issues/1761, and aligns well with https://github.com/balabit/syslog-ng/pull/2478 (which does the same with java binding).

- The worker insert result constants are exposed through `self`.
- Returning insert result constants is optional, send can still return with bool, if it is more convenient.
- new flush method is available, that will use the batching mechanism of threaded destination driver framework in terms of `batch_lines` and `batch_timeout`.

Available constants:
`WORKER_INSERT_RESULT_DROP`: message is dropped instantly
`WORKER_INSERT_RESULT_ERROR`: message is put back to the queue and retried later (`retries` retry number). In the meantime the destination is suspended for time-reopen() seconds.
`WORKER_INSERT_RESULT_SUCCESS`: message is accepted
`WORKER_INSERT_RESULT_QUEUED`: this constant should be returned for batching. Message acknowledgement of batches are controlled by the `flush` function.
`WORKER_INSERT_RESULT_NOT_CONNECTED`: Message is put back to the queue, and the destination is suspended. `open` will be retried later for the same message/batch.
`WORKER_INSERT_RESULT_MAX`: Maximum number of available constants.

Example:
```
class MyDestination(object):
    def init(self, options):
        self.bulk = list()
        return True

    def send(self, msg):
         self.bulk.append(msg["MSG"].decode())
         return self.WORKER_INSERT_RESULT_QUEUED

    def flush(self):
        print("flushing: " + ",".join(self.bulk))
        self.bulk = list()
        return self.WORKER_INSERT_RESULT_SUCCESS
```

The batching part depends on https://github.com/balabit/syslog-ng/pull/2478. Without it, batch_lines based batching is not working.


kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng test this please test branch=python-flush;

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

modules/python/python-dest.c
```diff
@@ -483,3 +550,19 @@ python_dd_new(GlobalConfig *cfg)
...
+    PyModule_AddObject(PyImport_AddModule("syslogng"), \
+  #value, int_as_pyobject(value))
+
+  register_worker_result(WORKER_INSERT_RESULT_DROP);
+  register_worker_result(WORKER_INSERT_RESULT_ERROR);
+  register_worker_result(WORKER_INSERT_RESULT_EXPLICIT_ACK_MGMT);
+  register_worker_result(WORKER_INSERT_RESULT_SUCCESS);
+  register_worker_result(WORKER_INSERT_RESULT_QUEUED);
+  register_worker_result(WORKER_INSERT_RESULT_NOT_CONNECTED);
+  register_worker_result(WORKER_INSERT_RESULT_MAX);
```
MrAnno: I would not inject destination-related constants to the main syslogng namespace. Could you add them to the destionation object (since we don't have a superclass or a separate module for destinations)?

modules/python/python-dest.c
```diff
@@ -483,3 +550,19 @@ python_dd_new(GlobalConfig *cfg)
 
   return (LogDriver *)self;
 }
+
+void
+py_destination_init(void)
+{
+#define register_worker_result(value) \
```
MrAnno: Why did you use a macro?

furiel: The injection needs both the token and the string literal generated from the token. With functions, I would need to specify both.

modules/python/python-dest.c
```diff
@@ -185,10 +185,56 @@ _py_invoke_close(PythonDestDriver *self)
...
+  if (result == -1 && PyErr_Occurred())
+    {
+      gchar buf[256];
+      msg_error("Error converting PyObject to int. Dropping message",
+                evt_tag_str("exception", _py_format_exception_text(buf, sizeof(buf))));
+      _py_finish_exception_handling();
+      return WORKER_INSERT_RESULT_DROP;
+    }
+
+  if (result < 0 || result >= WORKER_INSERT_RESULT_MAX)
```
Kokan: In case of the other language binding (java) I did not saw such safe guarding. Would it be possible to keep them in sync ?

Is it right to drop the message ?

modules/python/python-dest.c
furiel: Yes sure! I will do so.

modules/python/python-dest.c
furiel: Good idea, I will push a patch on the top of https://github.com/balabit/syslog-ng/pull/2478., just let's agree on the behaviour first.

I think the dropping is preferable here.
From development point of view, the behaviour does not really matter, the interesting question is when the condition hits a production environment. When this condition occurs, it is already a very bad situation, meaning that the driver implementation is faulty. It should not be that faulty that the condition is triggered for every messages (talking about production code), the problem should be message specific. When talking about message specific, it is probably not worth retrying the same message, because the same input will probably trigger the same condition.

With that said, I think the best is to drop the message so that other messages could be handled without delay.
The second best option would be disconnect, which will suspend the driver (probably infinitely as the same message will be retried over and over), when the messages will accumulate into the queue. This would minimize message loss to some extent, but depending on how quickly the problem is detected and addressed by the operators, this will still result in either message loss in syslog-ng (without flow control), or congestion at some other part of the operator network. And causes the largest possible delay in the messages depending how quickly can the message queue be salvaged.
The least preferable option is retry mechanism, which would be the most optimistic version, but I think if a code path fails on a message, it will fail again later.

What do you think?

modules/python/python-dest.c
furiel: I fixed the issue. I added an extra refactor patch to store the references to clear in an array instead explicitely calling them in `_py_free_bindings`. I felt storing the individual objects would have been too much boilerplate.

kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng test this please test branch=python-flush;

kira-syslogng:
Build SUCCESS

modules/python/python-dest.c
furiel: Okok, I can align align to that.

Also, just occurred to me that any uncatched exception in the send method is very similar scenario to this one: and we return `WORKER_INSERT_RESULT_ERROR` in such case. So using `WORKER_INSERT_RESULT_ERROR` here is the consistent choice.

bazsi:
I had a comment on one of the commits but otherwise this looks good to me.

And its very much appreciated making is possible to do proper batching even from Python.

furiel:
@bazsi thanks for the idea. That would also solve my intention: to be able to free all references in one step. I am just thinking if there are more alternatives. I think the cleanest would be if python add a reference during setattr, and unref when the class itself is released. In that case we could release the reference right after setattr, and would not have any work in the `_py_free_bindings()` function. I merge this for now as it is, but I plan to experiment with the topic.

PR #2489, opened 2019-01-11
syslog-ng@default: use pid file location on control socket

andy9a9:
When the pid file location was changed via `--with-pidfile-dir=` control socket was still pointing to `/var/lib/syslog-ng` even if they are using same path definition.

### DIFF
--- a/contrib/systemd/syslog-ng@default
+++ b/contrib/systemd/syslog-ng@default
@@ -1,5 +1,5 @@
 CONFIG_FILE=/etc/syslog-ng.conf
 PERSIST_FILE=/var/lib/syslog-ng/syslog-ng.persist
-CONTROL_FILE=/var/lib/syslog-ng/syslog-ng.ctl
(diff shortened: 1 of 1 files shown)


Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

gaborznagy:
Hi @andy9a9!

Thanks for your contribution!

Can I ask which distribution do you use for syslog-ng?
Do you use unofficial builds or you build it from source?

Because the packaging for various distributions can be different and the service file you edited maybe not used in those distros.
Nevertheless what you have changed follows a convention (what you have found in lib/syslog-ng.h)!

andy9a9:
Hi @gaborznagy !
I am building syslog-ng with Yocto (Open-embedded) distribution. Somebody change the pid file location. I was wondering, that .ctl socket location was also changed. So I was searching a little bit about and found this issue.

PR #2490, opened 2019-01-11
pytest_framework: support for implicit groups in config.create_logpath

furiel:
There is a little repetitive process in test writing in pytest framework: each driver (for example `FileSource`) must be encapsulated into a `StatementGroup` (for example `source`), prior constructing `LogPath`. If the group is not needed otherwise (one does not want to have multiple drivers in the same group, or update group on the fly), the explicit construction of the intermediate `StatementGroup` is unnecessary.

This patchset features logpath creation without explicitely creating these statement groups.

Note to developers:
 I like the idea of having the same api for the group/non-group based construction in `config.create_logpath`, which means the function should branch somehow on the parameter type, that's why I think the `isinstance` part in the last patch is reasonable. At least I don't have any clue how can I eliminate it from clean code point of view for now :(. If you feel using `isinstance` too offensive, I can offer to create a `def create_logpath(statements=[], groups=[])` separation and delegate the classification problem to the users. But I think the current api looks neat form user perspective.

### DIFF
--- a/tests/pytest_framework/functional_tests/logpath/test_flags_catch_all.py
+++ b/tests/pytest_framework/functional_tests/logpath/test_flags_catch_all.py
@@ -42,17 +42,12 @@ def test_flags_catch_all(tc):
     config = tc.new_config()
 
     file_source = config.create_file_source(file_name="input.log")
-    source_group = config.create_statement_group(file_source)
-
     file_destination1 = config.create_file_destination(file_name="output1.log")
-    destination_group = config.create_statement_group(file_destination1)
+    catch_all_destination = config.create_file_destination(file_name="output2.log")
 
-    file_destination2 = config.create_file_destination(file_name="output2.log")
-    catch_all_destination = config.create_statement_group(file_destination2)
+    inner_logpath = config.create_inner_logpath(statements=[file_destination1])
 
-    inner_logpath = config.create_inner_logpath(statements=[destination_group])
-
-    config.create_logpath(statements=[source_group, inner_logpath])
+    config.create_logpath(statements=[file_source, inner_logpath])
     config.create_logpath(statements=[catch_all_destination], flags="catch-all")
 
     bsd_message = write_dummy_message(tc, file_source)
@@ -66,7 +61,7 @@ def test_flags_catch_all(tc):
     # message should arrived into destination1
     assert tc.format_as_bsd(expected_bsd_message) in dest1_logs
 
-    dest2_logs = file_destination2.read_log()
+    dest2_logs = catch_all_destination.read_log()
     # message should arrived into destination2
     # there is a flags(catch-all)
     assert tc.format_as_bsd(expected_bsd_message) in dest2_logs
--- a/tests/pytest_framework/functional_tests/logpath/test_flags_catch_all.py
+++ b/tests/pytest_framework/functional_tests/logpath/test_flags_catch_all.py
@@ -42,13 +42,13 @@ def test_flags_catch_all(tc):
     config = tc.new_config()
 
     file_source = config.create_file_source(file_name="input.log")
-    source_group = config.create_source_group(file_source)
+    source_group = config.create_statement_group(file_source)
 
     file_destination1 = config.create_file_destination(file_name="output1.log")
-    destination_group = config.create_destination_group(file_destination1)
+    destination_group = config.create_statement_group(file_destination1)
 
     file_destination2 = config.create_file_destination(file_name="output2.log")
-    catch_all_destination = config.create_destination_group(file_destination2)
+    catch_all_destination = config.create_statement_group(file_destination2)
 
     inner_logpath = config.create_inner_logpath(statements=[destination_group])
 
--- a/tests/pytest_framework/functional_tests/logpath/test_flags_catch_all.py
+++ b/tests/pytest_framework/functional_tests/logpath/test_flags_catch_all.py
@@ -42,10 +42,10 @@ def test_flags_catch_all(tc):
     config = tc.new_config()
 
     file_source = config.create_file_source(file_name="input.log")
-    file_destination1 = config.create_file_destination(file_name="output1.log")
-    catch_all_destination = config.create_file_destination(file_name="output2.log")
+    file_destination = config.create_file_destination(file_name="output.log")
+    catch_all_destination = config.create_file_destination(file_name="catchall_output.log")
 
-    inner_logpath = config.create_inner_logpath(statements=[file_destination1])
+    inner_logpath = config.create_inner_logpath(statements=[file_destination])
 
     config.create_logpath(statements=[file_source, inner_logpath])
     config.create_logpath(statements=[catch_all_destination], flags="catch-all")
@@ -57,11 +57,11 @@ def test_flags_catch_all(tc):
     syslog_ng = tc.new_syslog_ng()
     syslog_ng.start(config)
 
-    dest1_logs = file_destination1.read_log()
+    destination_logs = file_destination.read_log()
     # message should arrived into destination1
-    assert tc.format_as_bsd(expected_bsd_message) in dest1_logs
+    assert tc.format_as_bsd(expected_bsd_message) in destination_logs
 
-    dest2_logs = catch_all_destination.read_log()
-    # message should arrived into destination2
+    catch_all_destination_logs = catch_all_destination.read_log()
+    # message should arrived into catch_all_destination
     # there is a flags(catch-all)
-    assert tc.format_as_bsd(expected_bsd_message) in dest2_logs
+    assert tc.format_as_bsd(expected_bsd_message) in catch_all_destination_logs
--- a/tests/pytest_framework/functional_tests/source_drivers/file_source/test_acceptance.py
+++ b/tests/pytest_framework/functional_tests/source_drivers/file_source/test_acceptance.py
@@ -26,12 +26,9 @@ def test_acceptance(tc):
     config = tc.new_config()
 
     file_source = config.create_file_source(file_name="input.log")
-    source_group = config.create_statement_group(file_source)
-
     file_destination = config.create_file_destination(file_name="output.log")
-    destination_group = config.create_statement_group(file_destination)
 
-    config.create_logpath(statements=[source_group, destination_group])
+    config.create_logpath(statements=[file_source, file_destination])
 
     log_message = tc.new_log_message()
     bsd_log = tc.format_as_bsd(log_message)
--- a/tests/pytest_framework/functional_tests/source_drivers/file_source/test_acceptance.py
+++ b/tests/pytest_framework/functional_tests/source_drivers/file_source/test_acceptance.py
@@ -26,10 +26,10 @@ def test_acceptance(tc):
     config = tc.new_config()
 
     file_source = config.create_file_source(file_name="input.log")
-    source_group = config.create_source_group(file_source)
+    source_group = config.create_statement_group(file_source)
 
     file_destination = config.create_file_destination(file_name="output.log")
-    destination_group = config.create_destination_group(file_destination)
+    destination_group = config.create_statement_group(file_destination)
 
     config.create_logpath(statements=[source_group, destination_group])
 
--- a/tests/pytest_framework/src/syslog_ng_config/statement_group.py
+++ b/tests/pytest_framework/src/syslog_ng_config/statement_group.py
@@ -25,13 +25,11 @@ import functools
 from src.common.random_id import get_unique_id
 from src.common.operations import cast_to_list
 
-class StatementGroup(object):
+class StatementGroup(list):
     def __init__(self, statements):
+        super(StatementGroup, self).__init__(cast_to_list(statements))
         self.__group_type = self.__calculate_group_type(cast_to_list(statements))
         self.__group_id = "%s_%s" % (self.__group_type, get_unique_id())
-        self.__statements = []
-        if statements:
-            self.update_group_with_statements(cast_to_list(statements))
 
     @staticmethod
     def __calculate_group_type(statements):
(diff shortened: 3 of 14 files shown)


kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/syslog_ng_config/statement_group.py
```diff
@@ -22,13 +22,24 @@
 #############################################################################
 
 from src.common.random_id import get_unique_id
-
+import functools
```
mitzkia: This should be the first import statement (in the order)

tests/pytest_framework/src/syslog_ng_config/statement_group.py
```diff
@@ -22,13 +22,24 @@
...
 class StatementGroup(object):
-    def __init__(self, group_type):
-        self.__group_type = group_type
-        self.__group_id = "%s_%s" % (group_type, get_unique_id())
+    def __init__(self, statements):
+        self.__group_type = self.__calculate_group_type(cast_to_list(statements))
+        self.__group_id = "%s_%s" % (self.__group_type, get_unique_id())
         self.__statements = []
+        if statements:
+            self.update_group_with_statements(cast_to_list(statements))
```
mitzkia: I would not put self.update_group_with_statements() in to constructor. From the caller side I would not expect this.
I would call it separately after the StatementGroup() is created. I understand it is one more step which we can avoid.

tests/pytest_framework/src/syslog_ng_config/statement_group.py
```diff
@@ -22,13 +22,24 @@
...
+
+    def __calculate_group_type(self, statements):
+        def check_consistency(stmt1, stmt2):
+            if stmt1.group_type != stmt2.group_type:
+                raise TypeError("Conflict in statement types: {} and {}".format(x, y))
```
mitzkia: typo: 
x -> stmt1.group_type, y -> stmt2.group_type.

tests/pytest_framework/src/syslog_ng_config/syslog_ng_config.py
```diff
@@ -54,17 +54,15 @@ def write_config_content(self):
...
+    def __create_logpath_group(self, statements=None, flags=None):
         logpath = LogPath()
         if statements:
-            logpath.update_logpath_with_groups(cast_to_list(statements))
+            for statement in cast_to_list(statements):
+                if isinstance(statement, StatementGroup) or isinstance(statement, LogPath):
+                    logpath.add_group(statement)
+                else:
+                    group = self.create_group(statement)
+                    logpath.add_group(group)
```
mitzkia: From now the __create_logpath_group() does two things: 
1, it can add statements to the logpath group
2, can create a new statement group for drivers
I would separate the second logic before calling the __create_logpath_group() -> with this we can eliminate the if statement

Other thing: the incoming "statements" parameter could be a driver or a statement, it can be confusing. 
Can we choose other name for this two entities?

tests/pytest_framework/src/syslog_ng_config/syslog_ng_config.py
```diff
@@ -81,20 +79,10 @@ def create_file_destination(self, **kwargs):
...
-    def create_destination_group(self, drivers):
-        destination_group = self.__create_statement_group("destination", drivers)
-        self.__syslog_ng_config["statement_groups"].append(destination_group)
-        return destination_group
-
-    def create_filter_group(self, filters):
-        filter_group = self.__create_statement_group("filter", filters)
-        self.__syslog_ng_config["statement_groups"].append(filter_group)
-        return filter_group
+    def create_group(self, drivers):
```
mitzkia: Optional: Consider to call this function as: create_statement_group().
I would call the parameter not drivers, but statements (filter also can be added),
I would call the inner variable not group, but statement_group

tests/pytest_framework/src/syslog_ng_config/statement_group.py
```diff
@@ -22,13 +22,24 @@
...
-        self.__group_type = group_type
-        self.__group_id = "%s_%s" % (group_type, get_unique_id())
+    def __init__(self, statements):
+        self.__group_type = self.__calculate_group_type(cast_to_list(statements))
+        self.__group_id = "%s_%s" % (self.__group_type, get_unique_id())
         self.__statements = []
+        if statements:
+            self.update_group_with_statements(cast_to_list(statements))
+
+    def __calculate_group_type(self, statements):
```
mitzkia: This function can be static

tests/pytest_framework/functional_tests/source_drivers/file_source/test_acceptance.py
```diff
@@ -26,12 +26,9 @@ def test_acceptance(tc):
     config = tc.new_config()
 
     file_source = config.create_file_source(file_name="input.log")
-    source_group = config.create_source_group(file_source)
-
     file_destination = config.create_file_destination(file_name="output.log")
-    destination_group = config.create_destination_group(file_destination)
 
-    config.create_logpath(statements=[source_group, destination_group])
+    config.create_logpath(statements=[file_source, file_destination])
```
mitzkia: A little bit confusing parameter name: drivers are not statements.

tests/pytest_framework/src/syslog_ng_config/statement_group.py
furiel: Sorry I cannot understand this one. Why does it look unexpected from the caller site? What would be the definition/short explanation of `StatementGroup`?

This is how I think about it: `StatementGroup` is a container type that is intended to store statements. With that, it is very close in feeling to a regular list, just with a few extras. As I do not feel strange to initialize a regular list through constructor, it feels analogous here too.

But I know it is not how it is implemented, and this might be the source of disagreement. Now to mention it, I feel strange about the implementation.
StatementGroup should be a list(-ish) from ducktype point of view, with the add/remove (update_group_with_statement/remove_statement). But you cannot access the content as you would do with a list, because the content is exposed through a property.

What do you think about making the list analogue more explicit? I can reimplement it by inheriting from `list` directly. That would remove some code: the add/remove methods are not needed anymore. And also would have a lot of exciting benefit: for example you could write statement_group[0], or can use pop, insert, clear, copy, etc. Also for the iteration, it would look very neat to write `for statement in statement_group:` compared to `for statement in statement_group.statements`.

What do you think?

tests/pytest_framework/src/syslog_ng_config/syslog_ng_config.py
furiel: Good point thanks. Agreed on the first part, I will do so.

About the naming part, unfortunately I cannot come up with a good name. How does it sound **not** choosing a name? My suggestion for signature: `def create_logpath(items, flags=None)`? I think it would be as convenient as the original one from user point of view.

tests/pytest_framework/src/syslog_ng_config/syslog_ng_config.py
furiel: Agreed. I will do so. The only downside would be is that it means more typing (/longer lines) for the user. But it is a good clarification, as it is a member of the config, and statement might not be the first thing that comes to my mind when someone mentions group in config context. Also, with this pr merged, users not necessarily need to make these groups explicitely, so the longer typing is reasonable.

kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/syslog_ng_config/syslog_ng_config.py
mitzkia: After checking the update, I have new comment on this part.
About splitting the __create_logpath_group():
- with the new version there is an almost duplicated function: __create_logpath() vs create_logpath().
- New functions I feel a little bit complex
- however I would do this part in this way, what do you think?
```
    @staticmethod
    def __create_logpath_group(statements=None, flags=None):
        logpath = LogPath()
        if statements:
            logpath.add_groups(statements)
        if flags:
            logpath.add_flags(cast_to_list(flags))
        return logpath

    def create_statement_group(self, statements):
        statement_group = StatementGroup(statements)
        self.__syslog_ng_config["statement_groups"].append(statement_group)
        return statement_group

    def __create_statement_group_if_needed(self, statements):
        for statement in statements:
            if not isinstance(statement, (LogPath, StatementGroup)):
                self.create_statement_group(statement)

    def create_logpath(self, statements=None, flags=None):
        self.__create_statement_group_if_needed(statements)
        logpath = self.__create_logpath(statements, flags)
        self.__syslog_ng_config["logpath_groups"].append(logpath)
        return logpath

    def create_inner_logpath(self, statements=None, flags=None):
        self.__create_statement_group_if_needed(statements)
        inner_logpath = self.__create_logpath(statements, flags)
        return inner_logpath
```
- about the "statement" parameter name: we can go with this for now (later we can change it if it will confusing)

tests/pytest_framework/src/syslog_ng_config/statement_group.py
mitzkia: I understand the benefits of inheritance from list object. 
But we should be able to update (for now we have this with: update_group_with_statements()) the statement_group list with more than one statements in one turn. 
Originally lists has append(), but it is not accepts more than one items. We would need such function (will be this understandable?):
```
def append_items(self, items):
   for item in items:
      self.append(item)
```

mitzkia: If we will not lose functionality from StatementGroup() than +1 from my side

furiel: Ok, I will do so. The builtin list contains an `extend` method that adds many elements based on an iterator that should be available for us. So we would not even need `append_items`.

tests/pytest_framework/src/syslog_ng_config/syslog_ng_config.py
furiel: Unfortunately I do not see the goal here. `__create_logpath()` vs `create_logpath` are not a duplicated functions (except the name, but I can rename to `__create_logpath_wrapper` or `__create_logpath_with_conversion`). However using `self.__create_statement_group_if_needed` as above is real a code duplication of the same logic in `create_logpath` and `create_inner_logpath`. If I merge the duplicated logic into a single function, then the only difference I see between the two code structures is that somehow the map function is eliminated. I believe that is intended to be eliminated  by `__create_statement_group_if_needed`, but the pseudocode is incomplete, at least I miss the aggregation of the converted `StatementGroup` into a list that would be returned. Which would be solved by the map function.

The wrapper function is a restriction/necessity that originates the logic duplication in `create_logpath` vs `create_inner_logpath`. We shortly discussed this in https://github.com/balabit/syslog-ng/pull/2443#discussion_r245476988. In the original proposal, they was only one function, which we split into two functions to simplify the api from user point of view. The original problem came from that `create_logpath` constructs a logpath instance, but also registers it as a side effect into the config, which must not be done in case of inner paths. This latter was addressed by splitting into two functions based in inner logpath-toplevel logpath, but did not eliminate the original root cause. In consequence, whenever logpath logic changed, that must be followed in both functions (inner and toplevel), as a result a function is needed for the common parts, and this is the lead I implemented above.

As an alternative, we might rethink the whole path creation and registration process, and go with the new_logpath()+add_logpath() duet, when new_logpath() would be used to construct both inner and toplevel logpaths. With that, I can avoid the wrapper function, and execute the conversion right in `new_logpath`.

kira-syslogng:
Build SUCCESS

tests/pytest_framework/functional_tests/logpath/test_flags_catch_all.py
```diff
@@ -42,17 +42,12 @@ def test_flags_catch_all(tc):
     config = tc.new_config()
 
     file_source = config.create_file_source(file_name="input.log")
-    source_group = config.create_statement_group(file_source)
-
     file_destination1 = config.create_file_destination(file_name="output1.log")
```
alltilla: In `file_destination1`, `1` could be removed, as there is no other `file_destination` in this function.

tests/pytest_framework/functional_tests/logpath/test_flags_catch_all.py
```diff
@@ -66,7 +61,7 @@ def test_flags_catch_all(tc):
     # message should arrived into destination1
     assert tc.format_as_bsd(expected_bsd_message) in dest1_logs
 
-    dest2_logs = file_destination2.read_log()
+    dest2_logs = catch_all_destination.read_log()
```
alltilla: Might not be the scope of this change, but `dest2_logs` is not so talkative. IMO `catch_all_dest_logs` or something might fit better. Same applies for `dest1_logs`

tests/pytest_framework/functional_tests/logpath/test_flags_catch_all.py
furiel: Fixed thanks.

tests/pytest_framework/functional_tests/logpath/test_flags_catch_all.py
furiel: Fixed thanks.

kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/syslog_ng_config/syslog_ng_config.py
mitzkia: Thanks, for the explanation. After rethought my pseudo code it missed the return list (as you pointed out), and with that it will have a same complexity. This part is clear. I would add some function renaming:
```
__maybe_convert_to_statement_group() -> create_statement_group_if_needed(), 
```
```
__create_logpath() -> __create_logpath_with_conversion() 
```
and 

```
isinstance(item, StatementGroup) or isinstance(item, LogPath) -> isinstance(item, (StatementGroup, LogPath))
```

mitzkia:
I have only some optional minor function rename comments. This is PR is approved.

tests/pytest_framework/src/syslog_ng_config/syslog_ng_config.py
furiel: I like the suggestions. I fixed these.

PR #2492, opened 2019-01-14
pytest_framework: faster failure detection for semantically invalid configs

furiel:
Prior this patchet, the syslog-ng start validation was the following: we relied on `syslog-ng -s` to detect if syslog-ng can be started. After that it waited for the control socket with a 10s timeout. However, when the configuration passes `syslog-ng -s`, the control socket may or may not be created, if there are other kind of errors in configuration that `syslog-ng -s` does not detect. That introduces a 10s delay for detecting failure in those cases.

For that sake, this patch checks if syslog-ng exited unexpectedly besides the control socket check.

This patch also eliminates the `class SyslogNgCtl` (or `/SyslogNgCtlCli`, depending on point of view).
Edit: We decided not to eliminate `SyslogNgCtl` here.

### DIFF
--- a/tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
+++ b/tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
@@ -25,7 +25,7 @@ from pathlib2 import Path
 from src.syslog_ng.syslog_ng_executor import SyslogNgExecutor
 from src.syslog_ng.console_log_reader import ConsoleLogReader
 from src.syslog_ng_ctl.syslog_ng_ctl import SyslogNgCtl
-from src.common.blocking import wait_until_true
+from src.common.blocking import wait_until_true, wait_until_false
 
 class SyslogNgCli(object):
     def __init__(self, logger_factory, instance_paths, testcase_parameters):
@@ -121,9 +121,9 @@ class SyslogNgCli(object):
             # wait for stop and check stop result
             if result["exit_code"] != 0:
                 self.__error_handling()
-            if not self.__syslog_ng_ctl.wait_for_control_socket_stopped():
+            if not wait_until_false(self.__is_process_running):
                 self.__error_handling()
-                raise Exception("Control socket still alive")
+                raise Exception("syslog-ng did not stop")
             if not self.__console_log_reader.wait_for_stop_message():
                 self.__error_handling()
                 raise Exception("Stop message not arrived")
--- a/tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
+++ b/tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
@@ -54,23 +54,13 @@ class SyslogNgCli(object):
             command_short_name="syntax_only", command=["--syntax-only", "--cfgfile={}".format(config_path)]
         )
 
-    # Process commands
-    def start(self, config):
-        self.__logger.info("Beginning of syslog-ng start")
-        config.write_config_content()
-
-        # syntax check
+    def __syntax_check(self):
         result = self.__syntax_only()
         if result["exit_code"] != 0:
             self.__logger.error(result["stderr"])
             raise Exception("syslog-ng can not started")
 
-        # effective start
-        if self.__valgrind_usage:
-            self.__process = self.__syslog_ng_executor.run_process_with_valgrind()
-        else:
-            self.__process = self.__syslog_ng_executor.run_process()
-
+    def __wait_for_start(self):
         # wait for start and check start result
         if not self.__syslog_ng_ctl.wait_for_control_socket_alive():
             self.__error_handling()
@@ -78,6 +68,22 @@ class SyslogNgCli(object):
         if not self.__console_log_reader.wait_for_start_message():
             self.__error_handling()
             raise Exception("Start message not arrived")
+
+    def __start_syslog_ng(self):
+        if self.__valgrind_usage:
+            self.__process = self.__syslog_ng_executor.run_process_with_valgrind()
+        else:
+            self.__process = self.__syslog_ng_executor.run_process()
+        self.__wait_for_start()
+
+    # Process commands
+    def start(self, config):
+        self.__logger.info("Beginning of syslog-ng start")
+        config.write_config_content()
+
+        self.__syntax_check()
+        self.__start_syslog_ng()
+
         self.__logger.info("End of syslog-ng start")
 
     def reload(self, config):
--- a/tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
+++ b/tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
@@ -25,7 +25,7 @@ from pathlib2 import Path
 from src.syslog_ng.syslog_ng_executor import SyslogNgExecutor
 from src.syslog_ng.console_log_reader import ConsoleLogReader
 from src.syslog_ng_ctl.syslog_ng_ctl import SyslogNgCtl
-
+from src.common.blocking import wait_until_true
 
 class SyslogNgCli(object):
     def __init__(self, logger_factory, instance_paths, testcase_parameters):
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -25,7 +25,7 @@
 from src.syslog_ng.syslog_ng_executor import SyslogNgExecutor
 from src.syslog_ng.console_log_reader import ConsoleLogReader
 from src.syslog_ng_ctl.syslog_ng_ctl import SyslogNgCtl
-from src.common.blocking import wait_until_true
+from src.common.blocking import wait_until_true, wait_until_false
```
alltilla: I think this belongs to the previous commit (`pytest_framework: look for process instead of stats when exiting`).

kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -54,30 +54,51 @@ def __syntax_only(self, config_path=None):
...
-        if self.__valgrind_usage:
-            self.__process = self.__syslog_ng_executor.run_process_with_valgrind()
-        else:
-            self.__process = self.__syslog_ng_executor.run_process()
-
+    def __is_process_running(self):
+        return self.__process.poll() == None
+
+    def wait_for_control_socket_alive(self):
+        class StoppedUnexpectedly(Exception):
```
mitzkia: I would not create separate Exception class for now. We already have several raise calls with standard Exception class, they would be strange to read.
If we wanted to add description about the raise we can use Exception("why we raises") form. I understand this is a convention. But if we started to create classes for raising let collect them in separate files, and we can iteratively extend them. I wanted to structure pytest_framework as clear as possible.
My other comment to here: 
I would not create wait_for_control_socket_alive() in this way. What we wanted here is to wait if syslog-ng process can started or not with: self.process.poll(), this is welcome.
But, I would create only a new class function:
```
def __wait_for_process_alive(self): 
    return wait_until_true(__is_process_running)
```
Maybe __is_process_running() also should be contains a check with an inside raise (as you wrote).
But, after this we can use it like (I think it is much readable):
```
    def wait_for_start(self):
        # wait for start and check start result
        if not self.wait_for_process_alive():
            self.__error_handling()
        if not self.wait_for_control_socket_alive():
            self.__error_handling()
        ...
```
As I think it should do the same as you wanted

Ok, after this I realize there are some repetative code what we can refactor into one function:
- we wanted to wait for process, for socket, for console log message.

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -54,30 +54,51 @@ def __syntax_only(self, config_path=None):
...
+            return s.__syslog_ng_ctl.is_control_socket_alive()
+        try:
+            return wait_until_true(is_alive, self)
+        except StoppedUnexpectedly:
+            return False
+
+    def wait_for_start(self):
         # wait for start and check start result
-        if not self.__syslog_ng_ctl.wait_for_control_socket_alive():
+        if not self.wait_for_control_socket_alive():
```
mitzkia: This will not needed

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -88,7 +109,7 @@ def reload(self, config):
         self.__syslog_ng_ctl.reload()
 
         # wait for reload and check reload result
-        if not self.__syslog_ng_ctl.wait_for_control_socket_alive():
+        if not self.wait_for_control_socket_alive():
```
mitzkia: This will not needed, but the new __wait_for_process_alive()

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -105,9 +126,9 @@ def stop(self, unexpected_messages=None):
             # wait for stop and check stop result
             if result["exit_code"] != 0:
                 self.__error_handling()
-            if not self.__syslog_ng_ctl.wait_for_control_socket_stopped():
+            if not wait_until_false(lambda s : s.__is_process_running(), self):
```
mitzkia: Ok, agreed to change from checking stats to checking process, but lets use the new __wait_for_process_alive()

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
furiel: Unfortunately `wait_for_process_alive` would not help here, because it will instantly return, because the process is not started. We to look for "not started" instead of "started" somehow.

With @mitzkia help, the inner class is eliminated.

Resolving the issue.

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
furiel: as mentioned above, wait_for_process_alive cannot help here.

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
furiel: as mentioned above, wait_for_process_alive cannot help us now.

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
furiel: As mentioned above, wait_for_process_alive cannot help us now.

kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -54,30 +54,46 @@ def __syntax_only(self, config_path=None):
...
 
-        # effective start
-        if self.__valgrind_usage:
-            self.__process = self.__syslog_ng_executor.run_process_with_valgrind()
-        else:
-            self.__process = self.__syslog_ng_executor.run_process()
+    def __is_process_running(self):
+        return self.__process.poll() == None
+
+    def wait_for_control_socket_alive(self):
```
mitzkia: This function can be private

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -54,30 +54,46 @@ def __syntax_only(self, config_path=None):
...
+        def is_alive(s):
+            if not s.__is_process_running():
+                raise Exception("syslog-ng could not start")
+            return s.__syslog_ng_ctl.is_control_socket_alive()
+        return wait_until_true(is_alive, self)
 
+    def wait_for_start(self):
```
mitzkia: This function also can be private

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -105,9 +121,9 @@ def stop(self, unexpected_messages=None):
             # wait for stop and check stop result
             if result["exit_code"] != 0:
                 self.__error_handling()
-            if not self.__syslog_ng_ctl.wait_for_control_socket_stopped():
+            if not wait_until_false(lambda s : s.__is_process_running(), self):
```
mitzkia: is not enough only:
```
if not wait_until_false(self.__is_process_running)
```

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
furiel: Fixed.

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
furiel: Fixed

tests/pytest_framework/src/syslog_ng/syslog_ng_cli.py
furiel: Indeed. I overthought it a little :)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
@mitzkia right! I dropped the top patch.

PR #2493, opened 2019-01-14
pytest_framework: timeout parameter for wait_until_true

furiel:
Prior this patch, pytest_framework did not support specifying timeout
as parameter for `wait_until_true`. Instead, a global variable was
responsible for the timeout. The default timeout (10s) was good for
the functional tests, but not for the unit tests, which executed
properly but unnecessarily long. For unit tests this global variable
was changed to 0.1s, which is too small for the functional
tests. Consequently, unit tests and functional tests cannot not be
executed together.

This patch adds timeout parameter for `wait_until_true`, and set the
proper timeout locally for unit tests. Hence unit tests and e2e test
can be executed together.

For example:
```
python3 -m pytest --installdir ../root/ tests/pytest_framework/src tests/pytest_framework/functional_tests
```

The way of parameterization (arguments as tuple) was motivated from
the way `threading.Thread` works.
One inconvenient thing worth mentioning: in python - single element
tuples need explicit comma ... `(1) == 1`. One needs to use `(1,)`.

### DIFF
--- a/tests/pytest_framework/src/common/blocking.py
+++ b/tests/pytest_framework/src/common/blocking.py
@@ -23,13 +23,15 @@
 
 import time
 
-MONITORING_TIME = 10  # 10 sec
+DEFAULT_TIMEOUT = 10  # 10 sec
 POLL_FREQ = 0.001  # 1 millisecond
 
-
 def wait_until_true(func, *args):
+    return wait_until_true_custom(func, args)
+
+def wait_until_true_custom(func, args=(), timeout=DEFAULT_TIMEOUT):
     # Python 2 compatibility note: time.monotonic() is missing
-    t_end = time.time() + MONITORING_TIME
+    t_end = time.time() + timeout
     while time.time() <= t_end:
         result = func(*args)
         if result:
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

tests/pytest_framework/src/common/blocking.py
```diff
@@ -23,21 +23,19 @@
 
 import time
 
-MONITORING_TIME = 10  # 10 sec
-POLL_FREQ = 0.001  # 1 millisecond
+DEFAULT_TIMEOUT = 10  # 10 sec
+DEFAULT_POLL_FREQ = 0.001  # 1 millisecond
```
alltilla: Do we plan to change `DEFAULT_POLL_FREQ` anywhere? If not I believe the `DEFAULT_` part is unnecessary. Although, I can live with this. :)

furiel: Yes, I believe we will. We already do this in the unit tests, but in the future I think the timeout parameter will be exposed up until the user API. For example, for now read_logs waits for 10s unconditionally, but later this will be probably changed. Depending on a scenario, 10s maybe too small, or unnecessarily high, and we need to let the user specify the timeout.

So I would leave the name as it is.

furiel: Whops I misread the request. POLL_FREQ is not changed indeed, nor think it will. I rename it back.
It would be the best eliminate altogether, and use some kind of exponential backoff algorithm instead. Polling continuously for 10 seconds with milliseconds frequency seems impractical.

kira-syslogng:
Build SUCCESS

mitzkia:
Sorry, but I as wrote earlier I think we did not need this change.
We can workaround it another way.

furiel:
After discussing with @mitzkia IRL, we concluded to mitigate impact of the `wait_until_true` API change by maintaining the original API, but creating a customizable version called `wait_until_true_custom`, that can be used by the unit tests. Marking as WIP until that.

kira-syslogng:
Build SUCCESS

PR #2496, opened 2019-01-15
Rdkafka destination

bazsi:
This branch adds a librdkafka based destination to syslog-ng, building from the work of @pyr and @Kokan.

*Constraints:*
* we feed messages to rdkafka, acking back all messages to the syslog-ng core
* these are kept in memory in the librdkafka buffers for as long as they are not sent
* at exit, syslog-ng waits 1 minute for this buffer to empty, the timeout can also be controlled using the flush_timeout_on_shutdown() parameter
* if messages cannot be sent in this 1 minute, they will be backfed into the queue, so in case it is a disk buffer, they will be retransmitted. in case of a reload, they will be retransmitted even without a disk buffer.
* throttling is handled using the blocking semantics of librdkafka, syslog-ng will block as long as the librdkafka internal buffers are full. (controlled by `queue.buffering.max.messages` and `queue.buffering.max.kbytes` configuration settings of rdkafka)

It is possible to use the syslog-ng disk buffer, but that only saves messages that are still on syslog-ng side, e.g. not in rdkafka buffers.

*Example configuration*

I have started a dockerized kafka on my computer, this way:
```
$ docker run --rm  --net=host   spotify/kafka
```

Then, I was using this configuration for syslog-ng:

```
          kafka(global-config(metadata.broker.list("localhost:9092")
                                         queue.buffering.max.ms("1000")
                      )
                      topic("syslog-ng")
                      flush_timeout_on_shutdown(60000)
                      flush_timeout_on_reload(1000)
                      key("${HOST}_${PROGRAM}_${PID}")
                      message("$(format-json --scope nv-pairs)"));
```

*Features*

* you can use the syslog-ng template language in message() and key() to specify the payload and key for messages.
* we are relying on librdkafka to separate the workload into partitions, we are using the murmur2 hash by default. You can customize that by changing the `partitioner` configuration setting of librdkafka
* ordering is only ensured within partitions by kafka, so make sure you are using an appropriate key() value, if you want stable ordering across a set of name-value pairs.
* for scalability on the destatination side, you'll need to use a key() value that has a distribution. If everything falls to the same partition, scalability will be limited
* you can specify both global and topic configuration right within the syslog-ng configuration file (global-config() and topic-config() options respectively).
* syslog-ng supports the workers() option, e.g. we can run the librdkafka feeder on multiple threads. This may or may not be the bottleneck. These threads basically format the key() and the message() templates and call rd_kafka_produce().

*Future*
* it might make sense to add support for Java style properties files as the configuration mechanism of librdkafka. As I see it is pretty general to use those, at least in the Java implementation.
* kafka timestamp support is missing
* I think it would be great if we could connect syslog-ng's acknowledgement system with that of Kafka. That would mean that the normal syslog-ng flow control/throttling would extend to the librdkafka buffer and that inflight messages would be saved into the disk buffer.
* if the previous point would not work out, we could save the outstanding messages in a separate queue for resending. We could even persist that across restarts.

*Questions*
* I am not sure about the configuration syntax. In a way I like because it's pretty natural. On the other hand we are not using similar formats in other plugins.


kira-syslogng:
Build SUCCESS

bazsi:
@pyr I am not yet finished with all things I wanted to do, but you can check how I've changed the code/configuration syntax of your original kafka-c destination.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lbudai:
@Kokan: are you sure in that this goes into 3.21?

Kokan:
@lbudai no, but it would be great(ans still there is like a month till 3.21), and as far as I understand that list is not a commitment for the content.

kira-syslogng:
Build FAILURE

bazsi:
This should be good enough for testing, review.

furiel:
@bazsi some of the patches, especially in the beginning are not kafka related. Could you move them into a separate PR? Arguing them separately would make this review cleaner.
Specially, it would be great if you could separate `diskq: fix deadlock in log_queue_push_head()`, with some explanation about that patch.
The introduction of `expedite` is also an interesting question, that could be introduced separately from rdkafka: http would also be user of that feature (and hence the `http` related patch could be picked from the middle of rdkafka patches). Personally I have a little concern about the introduced api: a `flush` method with boolean parameter seems a little awkard for me.

bazsi:
Hi,

On Mon, Apr 1, 2019, 09:41 furiel <notifications@github.com wrote:

> @bazsi <https://github.com/bazsi> some of the patches, especially in the
> beginning are not kafka related. Could you move them into a separate PR?
> Arguing them separately would make this review cleaner.
>

I would, I just would like to clarify this request a bit. So far I have
considered such fixes/refactors to be better placed in branches that
triggered them, as they provide an obvious trigger for their existance. And
apart from review bandwidth questions I think it is better to review them
in the same PR.

Has this approach been changed? My opinion is that I don't think we should
split branches/PRs unless simply because of size.




Specially, it would be great if you could separate diskq: fix deadlock in
> log_queue_push_head(), with some explanation about that patch.
>

There's a double acquisition of the same lock, causing a lock in all
invocations of push_head() for the disk buffer.

The introduction of expedite is also an interesting question, that could be
> introduced separately from rdkafka: http would also be user of that feature
> (and hence the http related patch could be picked from the middle of
> rdkafka patches).
>

Like I mentioned I wouldn't remove the 3 lines of http change from an
rdkafka change if it explains the idea better.


Personally I have a little concern about the introduced api: a flush method
> with boolean parameter seems a little awkard for me.
>
I agree that "expedite" as a boolean parameter could be improved, but
please do not that logwriter has similarly a flush_mode argument.

We could probably add a dedicated flush method that does the flush at the
time of exit/reload.

furiel:
> Has this approach been changed? My opinion is that I don't think we should split branches/PRs unless simply because of size.

No, I think this has not changed. Especially, if the only user of the change is in the PR, the separation makes especially difficult (i.e impossible) to review.

Separation makes sense if the change is unrelated to the topic, or it has such importance that we need to increase the visibility about it. Specially, in case a deadlock patch, if it is hidden in this pr, it will not be visible in the changelog. For example, users might do upgrade if they know there was a deadlock fixed in a release, but not necessarily for sake of kafka, if they do not need it otherwise.

bazsi:
I am not sure there's any user of push_head() in the codebase, apart from
kafka, but I dont have my notebook handy.

modules/kafka/kafka-dest-worker.c
```diff
@@ -0,0 +1,156 @@
...
+  msg_debug("kafka: message published",
+            evt_tag_str("topic", owner->topic_name),
+            evt_tag_str("key", key->str ? : "NULL"),
+            evt_tag_str("message", message->str),
+            evt_tag_str("driver", owner->super.super.super.id),
+            log_pipe_location_tag(&owner->super.super.super.super));
+
+  /* we passed the allocated buffers to rdkafka, which will eventually free them */
+  g_string_steal(message);
+  if (key)
```
lbudai: @bazsi: key won’t be `NULL` here

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
This is now considered ready. It was renamed to kafka-c() and the SCL is modified to make it trivial
to switch between the legacy java based driver and this one.

The configuration options have been revamped to make them compatible with the  java based driver
to make the transition seamless. In this process the configuration was made a bit neater, so if
you know you are using the C based version, your configuration will be nicer at the cost of incompatibility with the Java stuff.

Here's my test config:
```
                kafka(config(
                             queue.buffering.max.ms("1000")
                             message.timeout.ms("5000")
                             debug(all)
                             "queue.buffering.max.ms" => "1000",
                             "message.timeout.ms" => "5000",
                             "debug" => "all"
                      )
                      properties-file("/install/etc/kafka.properties")
                      option("debug", "none")
                      topic("syslog-ng")
                      bootstrap-servers("localhost:9092")
                      flush-timeout-on-shutdown(1000)
                      client-lib-dir("/foo/bar")
                      key("${HOST}_${PROGRAM}_${PID}")
                      message("$(format-json --scope nv-pairs)"));

```

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

dbld/images/helpers/dependencies.sh
```diff
@@ -109,9 +109,21 @@ function filter_packages_by_platform {
...
+    echo 'deb http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_18.04 ./' | tee /etc/apt/sources.list.d/lbudai.list
+    cat | tee /etc/apt/preferences.d/lbudai <<EOF
+Package: *
+Pin: origin "download.opensuse.org"
+Pin-Priority: 1
+EOF
+    wget -qO - http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_18.04/Release.key | apt-key add -
+    apt-get update
+}
+
```
bazsi: This is how I ensured that OBS packages are not added unless explicitly requested by version constraints.

Please review @szemere @mitzkia

kira-syslogng:
Build SUCCESS

bazsi:
Wow, travis agrees that this is in good shape \o/

Kokan:
Travis is false positive for cmake:

```patch
diff --git a/modules/kafka/CMakeLists.txt b/modules/kafka/CMakeLists.txt
index 52346ac8b..5717ef850 100644
--- a/modules/kafka/CMakeLists.txt
+++ b/modules/kafka/CMakeLists.txt
@@ -21,6 +21,10 @@ add_library(kafka SHARED
   kafka-parser.c
   kafka-parser.h
   kafka-plugin.c
+  kafka-dest-driver.c
+  kafka-dest-worker.c
+  kafka-plugin.c
+  kafka-props.c
   ${CMAKE_CURRENT_BINARY_DIR}/kafka-grammar.c
   ${CMAKE_CURRENT_BINARY_DIR}/kafka-grammar.h
 )
@@ -37,4 +41,6 @@ bison_target(KafkaGrammar
     ${CMAKE_CURRENT_BINARY_DIR}/kafka-grammar.c
     COMPILE_FLAGS ${BISON_FLAGS})

+add_test_subdirectory(tests)
+
 install(TARGETS kafka LIBRARY DESTINATION lib/syslog-ng/ COMPONENT kafka)
diff --git a/modules/kafka/tests/CMakeLists.txt b/modules/kafka/tests/CMakeLists.txt
index d47ba2b74..1a7c8f831 100644
--- a/modules/kafka/tests/CMakeLists.txt
+++ b/modules/kafka/tests/CMakeLists.txt
@@ -1 +1 @@
-add_unit_test(CRITERION TARGET test_kafka-props DEPENDS kafka)
+add_unit_test(CRITERION LIBTEST TARGET test_kafka-props DEPENDS kafka)

```

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:
On Fri, May 3, 2019 at 4:05 PM Budai Laszlo <notifications@github.com>
wrote:

> *@lbudai* approved this pull request.
>
> A question to the flow control:
> The librdkafka supports delivery reports for each message, so would be
> nice if we could provide a more reliable message delivery. But we cannot
> ACK messages randomly... So if would like to support E2E flow control, then
> we would need to support random access ACK mechanism. Is this true?
>

Yes, spot on. Implementing this is absolutely feasible for memory based
queues and probably possible with non-reliable disk buffers (as the
overflow queue part is memory based). However, it is completely out of
question for reliable disk buffers.


>
> In modules/kafka/kafka-props.c
> <https://github.com/balabit/syslog-ng/pull/2496#discussion_r280769326>:
>
> > +{
> +  g_free(self->name);
> +  g_free(self->value);
> +  g_free(self);
> +}
> +
> +void
> +kafka_property_list_free(GList *l)
> +{
> +  g_list_foreach(l, (GFunc) kafka_property_free, NULL);
> +
> +  g_list_free(l);
> +}
> +
> +
> +typedef struct _KafkaPropertyFileReader
>
> Wouldn't be better to add a PropertyFileReader to lib?
>

I would do that if there's another user.


>
> In modules/kafka/kafka-dest-driver.c
> <https://github.com/balabit/syslog-ng/pull/2496#discussion_r280777854>:
>
> > +  msg_event_send(msg_event_create(level, buf, NULL));
> +  g_free(buf);
> +}
> +
> +static void
> +_kafka_delivery_report_cb(rd_kafka_t *rk,
> +                          void *payload, size_t len,
> +                          rd_kafka_resp_err_t err,
> +                          void *opaque, void *msg_opaque)
> +{
> +  KafkaDestDriver *self = (KafkaDestDriver *) opaque;
> +  LogMessage *msg = (LogMessage *) msg_opaque;
> +
> +  /* we already ACKed back this message to syslog-ng, it was kept in
> +   * librdkafka queues so far but successfully delivered, let's unref it */
> +
>
> The comment above is little bit misleading.
>

why?

lbudai:
there are two branches:
1) when the message is successfully delivered
2) when not... 

and when it is failed, we not unref the msg, but push back to the queue.

PR #2501, opened 2019-01-18
fix static uClibc-ng support

### LINKED ISSUE #2491: syslog-ng can't be build with static libc
# syslog-ng

## Version of syslog-ng

```
syslog-ng 3.19.1
```

## Platform

buildroot (linux)

# Issue

## Failure

building on a system with only static libraries is no longer possible as of 3.19.1. It used to work on 3.10.1.

http://autobuild.buildroot.net/results/c0bca852e053cd0e021205b3a949e750ae602333/build-end.log

The build failure was introduced by commit 01258e56abc5804f8ea977ff9ab98cc8e2fd4702 .

This may or may not be something that syslog-ng actually wants to support. The plugin architecture probably means that dynamic linking is required but there is enough it configure.ac that seems to handle static, shared and a mix of both.

For now I'll submit a patch to buildroot to disable syslog-ng for static-link-only configurations. That can remain in place if this get's closed as 'not-a-bug' or reverted if a fix is made.

Kokan:
In case of static `uClibc-ng` lib, the `<dlfcn.h>` does not exists.

The `signal-handler.c` solves an issue in case of `program` src/dst
and `Java` destination. That combination is really rare if not impossible.

This patch simply disables own `sigaction` if `<dlfcn.h>` is not present.

Fixes #2491

@cpackham could you test with this patch, if any other issue is present ?


kira-syslogng:
Build SUCCESS

cpackham:
@Kokan this fixes the build failure. Thanks.

Kokan:
@furiel I was thinking about it, but compiling those modules are fine, even using them in a different syslog-ng instance are ok; using only a warning could be a way, but imho won't be noticed.
As I wrote above this is probably not a use case anybody has, I tend to not to handle at all.

If you or anybody still think a warning would be nice, I can update the patch accordingly.

furiel:
@Kokan I agree warning is not necessary here.

PR #2505, opened 2019-01-21
python: internal() source exposed via syslogng.Logger

### LINKED ISSUE #1375: python-based-func: Add log_debug like function
I would like to add some debug message to my python code (in my case to python parser), but I want to display them only if syslog-ng is running with debug mode.
Currently this is not possible because the python code does not know the current debug level.

Would be nice either to know the debug level or to have some log_* callback function, like log_debug, log_info, log_warning or similar that I can use to display any debug message.

Kokan:
Introduces a logger that can send messages into the `internal()` source, which also respects the syslog-ng log level.

```python
import syslogng

logger = syslogng.Logger()

logger.error("plain text message")
logger.warning("plain text message")
logger.info("plain text message")
logger.debug("plain text message")
logger.trace("plain text message")
```

Question:
* Could syslogng utilize the built-in `logger` class by extending it (creating only a handler for internal source) ?
* Does the apsense of `evt_tag_*` formatting cause significant performance issue? Should it be solved ?

Fixes #1375

### DIFF
--- a/modules/python/CMakeLists.txt
+++ b/modules/python/CMakeLists.txt
@@ -45,6 +45,8 @@ set(PYTHON_SOURCES
     python-logparser.c
     python-integerpointer.h
     python-integerpointer.c
+    python-logger.h
+    python-logger.c
     python-source.h
     python-source.c
     python-fetcher.h
--- a/modules/python/Makefile.am
+++ b/modules/python/Makefile.am
@@ -48,6 +48,8 @@ modules_python_libmod_python_la_SOURCES		= \
 	modules/python/python-logtemplate-options.c\
 	modules/python/python-integerpointer.c    \
 	modules/python/python-integerpointer.h    \
+	modules/python/python-logger.c    \
+	modules/python/python-logger.h    \
 	modules/python/python-global-code-loader.h\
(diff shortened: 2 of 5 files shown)


kira-syslogng:
Build SUCCESS

furiel:
> Could syslogng utilize the built-in logger class by extending it (creating only a handler for internal source) ?

I don't see much benefit to do so. On the other hand it might be a source of confusion, for example when someone wants to set debug level through that.

> Does the apsense of evt_tag_* formatting cause significant performance issue? Should it be solved ?

I don't think the absense would cause performace issue. However the presence might be. I investigated this one too, and me seemed python developers deliberately avoid lazy evaluation as we would need here. I do not see any way to avoid the function parameter evaluation with pure python, for example with debug messages when debug is disabled. If it were `hylang`,  maybe we could do better ...

lbudai:
Could syslogng utilize the built-in logger class by extending it (creating only a handler for internal source) ?
This is only a nice to have thing... (maybe when someone working with a 3rd party library that uses a python logger internally (`logging.getLogger(...)`) could benefit from such class... but I'm not sure).

Performance: I wouldn't worry about it. The feature is good enough for now, I think.

PR #2506, opened 2019-01-21
simpler names for WORKER_INSERT_RESULT_T in language bindings

furiel:
I got a valuable comment from @pzoleex that the recently introduced constant names in java and python destinations could be more user friendly.

In an attempt to do so, I did a little renaming on them. Also, the python example is updated.

I also removed the `WORKER_INSERT_RESULT_EXPLICIT_ACK_MGMT`, that could not be used by the bindings this time of writing.

Please review this with in mind that we conclude here, that would be published in the documentation.

### DIFF
--- a/modules/java/src/main/java/org/syslog_ng/LogDestination.java
+++ b/modules/java/src/main/java/org/syslog_ng/LogDestination.java
@@ -26,12 +26,11 @@ package org.syslog_ng;
 
 public abstract class LogDestination extends LogPipe {
 
-	protected static final int WORKER_INSERT_RESULT_DROP = 0;
-	protected static final int WORKER_INSERT_RESULT_ERROR = 1;
-	protected static final int WORKER_INSERT_RESULT_EXPLICIT_ACK_MGMT = 2;
-	protected static final int WORKER_INSERT_RESULT_SUCCESS = 3;
-	protected static final int WORKER_INSERT_RESULT_QUEUED = 4;
-	protected static final int WORKER_INSERT_RESULT_NOT_CONNECTED = 5;
+	protected static final int DROP = 0;
+	protected static final int ERROR = 1;
+	protected static final int SUCCESS = 3;
+	protected static final int QUEUED = 4;
+	protected static final int NOT_CONNECTED = 5;
 
 	public LogDestination(long pipeHandle) {
 		super(pipeHandle);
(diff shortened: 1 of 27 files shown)


furiel:
@kira-syslogng retest this please test branch=binding-user-friendly-names;

kira-syslogng:
Build FAILURE

modules/python/sngexample.py
```diff
@@ -45,12 +45,53 @@ def deinit(self):
...
+        After that the same message is retried until retries() times.
+
+        Integer:
+        self.SUCCESS: message sending was successful (same as boolean True)
+        self.ERROR: message sending was unsuccessful. Same message is retried.
+            (same as boolean False)
+        self.DROP: message cannot be sent, it should be dropped immediately.
+        self.QUEUED: message is not sent immediately, it will be sent with the flush method.
+        self.NOT_CONNECTED: message is put back to the queue, open method will be called until success."""
+
```
bazsi: If the Java destination defines these values in the java side, i am wondering if we could do the same in Python.

That would simplify the python part and make this pretty explicit.

modules/java/src/main/java/org/syslog_ng/LogDestination.java
```diff
@@ -26,12 +26,11 @@
...
-	protected static final int WORKER_INSERT_RESULT_SUCCESS = 3;
-	protected static final int WORKER_INSERT_RESULT_QUEUED = 4;
-	protected static final int WORKER_INSERT_RESULT_NOT_CONNECTED = 5;
+	protected static final int DROP = 0;
+	protected static final int ERROR = 1;
+	protected static final int SUCCESS = 3;
+	protected static final int QUEUED = 4;
+	protected static final int NOT_CONNECTED = 5;
 
 	public LogDestination(long pipeHandle) {
```
bazsi: I like these names a lot better. I would love to see the c names to be simplified as well. But that's optional.

furiel:
We discussed this with @bazsi , and I will sed the logthrdestdv constants too. `worker_insert_result_t` will be `LogThrResult`,`WORKER_INSERT_RESULT_SUCCESS` will be `LTR_SUCCESS` etc.

modules/python/sngexample.py
furiel: Agreed. Injecting `template_options` was also a bit strange, but I solely for that i did not want to introduce inheritance through parent class. However, the problem is much more visible now, because of these newly injected return values, that are intended to be used by all user scripts.
After resolving https://github.com/balabit/syslog-ng/pull/2438, we should introduce a proper inheritance from a python destination parent class.

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=binding-user-friendly-names;

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

PR #2514, opened 2019-01-28
riemann: fix ref/unref-ing templates

alltilla:
Setting the values of riemann driver calls `log_template_ref`, which must be unref'd after (in the ym file).

Previously it was not unref'd, which caused memory leak every reload, because at `riemann_dd_free()`, the ref counters were 2.

I have seen this behavior in other drivers, too:
afsmtp, http, msg-generator

### DIFF
--- a/modules/riemann/riemann-grammar.ym
+++ b/modules/riemann/riemann-grammar.ym
@@ -89,30 +89,37 @@ riemann_option
         | KW_HOST '(' template_content ')'
           {
             riemann_dd_set_field_host(last_driver, $3);
+            log_template_unref($3);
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

alltilla:
> I think `redis` is affected too.

I will check tomorrow, and open a PR, if that is the case. :)

bazsi:
hmm... we should probably add a riemann_set_XXX_ref() function that consumes the ref that gets
passed to it.

That's the pattern when the call-site has a ref that would need to be unrefed.

PR #2515, opened 2019-01-29
collect-cov.sh: make coverage should work with lcov in bionic

bazsi:
Our devshell image is bionic based so make sure our coverage targets
work with the version of lcov that is on bionic.

lcov seems to be changing configuration options quite frequently, but
I think it is enough to work in devshell.

### DIFF
--- a/tests/collect-cov.sh
+++ b/tests/collect-cov.sh
@@ -23,8 +23,10 @@
 
 set -e
 
-lcov --capture --directory `pwd` --compat-libtool --output-file coverage.info --rc geninfo_auto_base=1
+lcov --capture --directory `pwd` --compat-libtool --output-file coverage.info
 lcov --extract coverage.info "${top_srcdir}/*" --output-file coverage.info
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

tests/collect-cov.sh
```diff
@@ -23,8 +23,10 @@
 
 set -e
 
-lcov --capture --directory `pwd` --compat-libtool --output-file coverage.info --rc geninfo_auto_base=1
+lcov --capture --directory `pwd` --compat-libtool --output-file coverage.info
 lcov --extract coverage.info "${top_srcdir}/*" --output-file coverage.info
 lcov --remove coverage.info "${top_srcdir}/modules/afmongodb/mongo-c-driver/*" --output-file coverage.info
```
Kokan: Not related to this PR, but the `mongo-c-driver` and `rabbitmq-c` submodules are no longer part of the projects so there is no need to exclude them.

lbudai: @Kokan: indeed, we should remove them.

lbudai: @Kokan : #2525

PR #2512, opened 2019-01-24
threaded-dest: Fix integer overflow

alltilla:
A negative number's modulo will be negative, which in our case pointed to a negative entry of the workers array, causing segfault.

Fixes #2466.

Signed-off-by: Attila Szakacs <attila.szakacs@balabit.com>

PR #2516, opened 2019-01-29
threaded-dest: make persist_name local

alltilla:
Before this commit, the worker's queue's `persist_name` was stored in a static string, and it was accessed and changed by multiple threads in the same time.

The static string was there to avoid having to care about malloc/free, it is not performance critical, so it can be changed to local.

Follow up for the comments at #2512.

### DIFF
--- a/lib/logthrdestdrv.c
+++ b/lib/logthrdestdrv.c
@@ -76,24 +76,23 @@ log_threaded_dest_worker_rewind_messages(LogThreadedDestWorker *self, gint batch
   self->batch_size -= batch_size;
 }
 
-static const gchar *
+static gchar *
 _format_queue_persist_name(LogThreadedDestWorker *self)
 {
-  static gchar persist_name[1024];
   LogPipe *owner = &self->owner->super.super.super;
 
   if (self->worker_index == 0)
     {
       /* the first worker uses the legacy persist name, e.g.  to be able to
        * recover the queue previously used.  */
-
-      return owner->generate_persist_name(owner);
+      return g_strdup(owner->generate_persist_name(owner));
     }
   else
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

lib/logthrdestdrv.c
```diff
@@ -76,24 +76,21 @@ log_threaded_dest_worker_rewind_messages(LogThreadedDestWorker *self, gint batch
   self->batch_size -= batch_size;
 }
 
-static const gchar *
-_format_queue_persist_name(LogThreadedDestWorker *self)
+static void
+_format_queue_persist_name(LogThreadedDestWorker *self, gchar *persist_name, size_t len)
 {
-  static gchar persist_name[1024];
   LogPipe *owner = &self->owner->super.super.super;
```
bazsi: I would probably convert this (and other similar call sites) to use g_strdup_printf() and ensure the result is freed by the caller.

That way we don't have to care about 1024 character limits.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/logthrdestdrv.c
```diff
@@ -76,24 +76,21 @@ log_threaded_dest_worker_rewind_messages(LogThreadedDestWorker *self, gint batch
   self->batch_size -= batch_size;
 }
 
-static const gchar *
+static gchar *
```
furiel: Why did you need to remove const? `g_strdup_printf` indeed returns `gchar *`, but I think that is converted automatically without warnings to `const gchar *`.

kira-syslogng:
Build SUCCESS

PR #2517, opened 2019-01-29
Guess timezone based on difference3

bazsi:
NOTE: This is against guess-timezone-based-on-differences2 (PR #2483)

This patch implements a new flag "guess-timezone" for source drivers, which if specified causes syslog-ng to guess the sender's timestamp, based on the time difference between the local clock
and the incoming timestamp in the message.

Constraints:
* the sender clock needs to be roughly accurage (+/- 30 seconds is accepted)
* this should be used at the entry point of the managed log management infrastructure, so that no delays in reception can be expected (e.g. incoming disk buffer)
* it is pretty much incompatible with the time-zone() source option
* we have a list of valid timezone offsets and apart from even hours, we also support timezones with 30 or 45 minutes offsets. (e.g. India with IST +05:30)

Other piggybacked changes:
* This also includes a refactoring step of eliminating LogStamp to use UnixTime directly instead. This should have been in the previous PR, but it here, as the first patch.
* It also fixes timezone value for messages where an explicit time-zone() option is set AND 1) the timestamp is unparseable, 2) if keep-timestamp(no) is specified. In those cases, syslog-ng used the local timezone, instead of the one specified.


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/logmsg/logmsg-serialize.c
```diff
@@ -37,7 +37,7 @@
 #define OLD_LMM_REF_MATCH 0x0001
 
 static void
-_setup_ts_processed(LogStamp *timestamps, const LogStamp *processed)
+_setup_ts_processed(UnixTime *timestamps, const UnixTime *processed)
```
furiel: We think we should use the original name (LogStamp) instead, and should not do this renaming.

unixtime is a widely used term that includes only seconds. It sounds strange to include timezone info and microseconds. However, these concepts fit well with the LogStamp concept.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lib/timeutils/tests/test_unixtime.c
```diff
@@ -60,12 +61,12 @@ Test(unixtime, unix_time_fix_timezone_adjusts_timestamp_as_if_was_parsed_assumin
 
   _wct_initialize(&wct, "Jan 19 2019 18:58:48");
   wct.wct_gmtoff = 3600;
-  convert_wall_clock_time_to_unix_time(&ut, &wct);
+  convert_wall_clock_time_to_unix_time(&wct, &ut);
```
furiel: As I see the variable orders are fixups for the previous commit: `unixtime: add unix_time_{set,fix}_timezone() functions and associated testing`.

lib/timeutils/unixtime.c
```diff
@@ -45,6 +47,133 @@ unix_time_set_now(UnixTime *self)
   self->ut_gmtoff = get_local_timezone_ofs(self->ut_sec);
 }
 
+static glong
+_div_round(glong n, glong d)
```
furiel: Just a minor thing, but I think it would be more clear to abstract out a `_round(number, base)` function instead: the multiplication should be part of it.

This function is a special version of division in a way that after multiplying we would get the correct result in terms of rounding. But this function cannot stand on its own, or at least I cannot see its usefulness in another context.

furiel:
I will be out of office until next week. Feel free to merge this after the rebase: I had only cosmetic comments. Approve from my side.

kira-syslogng:
Build SUCCESS

lib/timeutils/tests/test_unixtime.c
bazsi: that's right, lost during the rebase. This is fixed now.

lib/timeutils/unixtime.c
bazsi: This is a division that rounds the result according to the mathematical definition, e.g. one that rounds upwards above .5 and downwards below that.

5/2 = 3
5/3 = 2
5/4 = 1

With non-floating point division, 5/2 would be 2. The name of the function could be improved somewhat, and you might be right that we should include the multiplication, let me think if there's a good function name for that...

kira-syslogng:
Build SUCCESS

lib/timeutils/unixtime.c
bazsi: I gave it a thought and I think the current name is not ultimately bad. It is a function that does division rounding the result. With that i resolve this conversation

furiel: I agree the function name is fine.

I wanted to imply that wrong function was extracted. Rounding has a strict mathematical formula, but the patch does not contain function for that. Instead a subexpression from the formula was extracted into the function. I think it would be better extracting the rounding itself into the function, instead of extracting _div_round. (Or both could be extracted.)

Still, this is just a minor thing. It is fine for me if this stays as it is.

bazsi:
Removing WIP flag as all dependencies have gone in. @furiel  has kindly reviewed this already, so just one more review to go!

lib/timeutils/tests/test_unixtime.c
```diff
@@ -44,6 +54,42 @@ Test(unixtime, unix_time_initialization)
...
+{
+  UnixTime ut = UNIX_TIME_INIT;
+  WallClockTime wct = WALL_CLOCK_TIME_INIT;
+
+  _wct_initialize(&wct, "Jan 19 2019 18:58:48");
+  wct.wct_gmtoff = 3600;
+  convert_wall_clock_time_to_unix_time(&wct, &ut);
+  cr_expect(ut.ut_gmtoff == 3600);
+
+  unix_time_fix_timezone(&ut, -5*3600);
```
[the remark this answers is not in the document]
bazsi: I don't understand the comment, sorry. This is using the conversion, as that's the key element of the timestamp guessing code. The hour of the day should not be changed, only the timezone.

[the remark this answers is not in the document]
Kokan: Simply put these test cases as is I think a conversion test, rather then a Unix time. There is a fine test set for them in  `test_conv.c`. So tests remain clean.
Still testing here `unix_time_fix_timezone` is ok, but only that, withouth any convertion:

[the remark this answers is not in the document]
bazsi: So you say, that instead of using a timestamp to initialize the Unix timestamp I should use an utc timestamp and that would make the test more readable?

I understand that the conversion is in a separate module, but surely a test depending on more modules doesn't introduce circular dependency. While in my opinion testing a bit more broadly we cover more code and the testcase effectively contains our main requirement.

The single requirement I want to test here is that the hour value remains and the timezone changes. How to do that without converting to/from hours.

If I avoid using wct, then all the test would contain is a change in the utc/timezone fields.

Of course that would be equivalent, but the main requirement would be completely missing from the test.

I feel that we are arguing a bit about how pure a unit test needs to be. I believe that they should encapsulate as much as needed, with the real code, especially if it makes the test requirements easier to code, and test run fast enough.

I remember the two camps of unit test practitioners, one camp always uses mocks so that we only use the subject under test the other incorporates, other, real classes as long as they are fast enough. I am a firm believer of the 2nd group. Here is a post that basically describes why, I am sure there are other explanations.

https://stackoverflow.com/questions/180413/mocks-or-real-classes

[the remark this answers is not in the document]
Kokan: Yes, if you initialise `UnixTime` directly, and call fix function, and do assert after; I think it can be as readable as asserting for hour, min, sec. As after calling fix function your requirement that the hour not changed, should already been met. The two case almost the same, but without `WallClockTime` it became simpler, thus more readable.

I am not worried here about circular dependencies, but rather if I change either `WallClockTime`, why should I change tests for `UnixTime` ? (in this case I might).

"I feel that we are arguing a bit about how pure a unit test needs to be." Mostly, and that's why I did not ask for change via change request.

While I do not see a relevance of mocking or not here; I agree on not mocking everything.

PR #2519, opened 2019-01-30
value-pairs: add an --omit-empty-values option

bazsi:
This new option will allow a template function to skip name-value pairs
that are empty. This used to be the default behavior until unset() was
introduced, since which time we only omit unset name-value pairs.

For example, ElasticSearch needs a transaction description as a JSON
before each record, that can be generated like this:

$(format-json --omit-empty-values index._type=`type` index._index=`index` index._id=`custom_id`)

If $(format-json) would include index._id as an empty string, that would
invalidate the JSON in the eyes of ElasticSearch.

### DIFF
--- a/lib/value-pairs/cmdline.c
+++ b/lib/value-pairs/cmdline.c
@@ -393,6 +393,8 @@ value_pairs_new_from_cmdline (GlobalConfig *cfg,
   ValuePairs *vp;
   GOptionContext *ctx;
 
+  vp = value_pairs_new();
+
   GOptionEntry vp_options[] =
   {
     {
@@ -440,6 +442,10 @@ value_pairs_new_from_cmdline (GlobalConfig *cfg,
       NULL, NULL
     },
     {
+      "omit-empty-values", 0, 0, G_OPTION_ARG_NONE, &vp->omit_empty_values,
+      NULL, NULL
+    },
+    {
       G_OPTION_REMAINING, 0, 0, G_OPTION_ARG_CALLBACK, vp_cmdline_parse_pair_or_key,
       NULL, NULL
     },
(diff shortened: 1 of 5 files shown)


lib/value-pairs/cmdline.c
```diff
@@ -439,6 +441,10 @@ value_pairs_new_from_cmdline (GlobalConfig *cfg,
       "subkeys", 0, 0, G_OPTION_ARG_CALLBACK, vp_cmdline_parse_subkeys,
       NULL, NULL
     },
+    {
+      "omit-empty-values", 0, 0, G_OPTION_ARG_NONE, &vp->omit_empty_values,
```
Kokan: I would rather create an `value_pairs_omit_empty()` and call it from here (so the struct does not need to be moved to the header).

kira-syslogng:
Build SUCCESS

furiel:
Adding wip due to breaking tests and the open question.

lib/value-pairs/cmdline.c
MrAnno: > I would rather create an `value_pairs_omit_empty()` and call it from here (so the struct does not need to be moved to the header).

It sounds good.
@bazsi Can I update this branch with the change requested by @Kokan?

kira-syslogng:
Build SUCCESS

lib/value-pairs/cmdline.c
MrAnno: I went ahead and updated the branch. Let me know if you don't like it.

kira-syslogng:
Build SUCCESS

bazsi:
The change is fine to me this way, thank you.

kira-syslogng:
Build FAILURE

MrAnno:

kira-syslogng:
Build SUCCESS

PR #2520, opened 2019-01-31
travis: temporarily downgrade to bison 3.2 (macOS)

MrAnno:
Bison 3.3 introduced multiple breaking changes that has to be followed in our codebase (maintaining backwards compatibility with older versions).

Until then, this patch downgrades bison to version 3.2 in Travis (macOS).


kira-syslogng:
Build SUCCESS

PR #2526, opened 2019-02-01
grammar: fix bison 3.3 deprecation warnings

bazsi:
```
    This patch removes the use of deprecated bison features, but at this point
    depends on an unreleased bison function, namely that api.location.type works
    for the C language, which was implemented by my request in:
    
    https://github.com/akimd/bison/commit/9260e5ca4fa62ff971fc52a2dd64474dd60e5990
    
    By merging this branch to master, we would be depending on a future bison
    release, probably 3.4.1.  This is added as a dependency in the configure
    script, but we should delay inclusion of this patch at the very least until
    that bison is released, and maybe even somewhat more until it appears in
    distros.
```


kira-syslogng:
Build FAILURE

MrAnno:
The minimum required Bison version in our configure/CMake files is currently 2.4.
I think we should increase the version at least up to 2.6, where `api.prefix` was introduced.

bazsi:
Yeah, we need to do that, but I am not sure that's available everywhere.
Although we can (and should include the generated .c/.h files into our
tarball, so we only need the recent bison versions on the machine we do
releases in, e.g. the devshell docker imagE).

czanik:
openSUSE Leap 15.1 also changed to bison 3.3.1 and build now fails:
```
[   70s]   YACC     lib/filter/filter-expr-grammar.h
[   70s] /home/abuild/rpmbuild/BUILD/syslog-ng-3.19.1/lib/filter/filter-expr-grammar.y:53.1-27: warning: deprecated directive, use '%define api.prefix {filter_expr_}' [-Wdeprecated]
[   70s]  %name-prefix "filter_expr_"
[   70s]  ^~~~~~~~~~~~~~~~~~~~~~~~~~~
[   70s] /home/abuild/rpmbuild/BUILD/syslog-ng-3.19.1/lib/filter/filter-expr-grammar.y:64.1-14: warning: deprecated directive, use '%define parse.error verbose' [-Wdeprecated]
[   70s]  %error-verbose
[   70s]  ^~~~~~~~~~~~~~
[   70s] /usr/bin/m4:/usr/share/bison/skeletons/yacc.c:1652: undefined macro `b4_symbol(229, has_type)'
[   70s] /usr/bin/m4:/usr/share/bison/skeletons/yacc.c:1652: undefined macro `b4_symbol(229, has_type)'
[   70s] /usr/share/bison/skeletons/yacc.c:1652: error: b4_symbol_if: field has_type of 229 is not a Boolean: 
[   70s] /usr/share/bison/skeletons/yacc.c:1652: the top level
[   70s]   YACC     lib/block-ref-grammar.c
[   70s] /home/abuild/rpmbuild/BUILD/syslog-ng-3.19.1/lib/filter/filter-expr-grammar.y:53.1-27: warning: deprecated directive, use '%define api.prefix {filter_expr_}' [-Wdeprecated]
[   70s]  %name-prefix "filter_expr_"
[   70s]  ^~~~~~~~~~~~~~~~~~~~~~~~~~~
[   70s] /home/abuild/rpmbuild/BUILD/syslog-ng-3.19.1/lib/filter/filter-expr-grammar.y:64.1-14: warning: deprecated directive, use '%define parse.error verbose' [-Wdeprecated]
[   70s]  %error-verbose
[   70s]  ^~~~~~~~~~~~~~
[   70s] /usr/bin/m4:/usr/share/bison/skeletons/yacc.c:1652: undefined macro `b4_symbol(229, has_type)'
[   70s] /usr/bin/m4:/usr/share/bison/skeletons/yacc.c:1652: undefined macro `b4_symbol(229, has_type)'
[   70s] /usr/share/bison/skeletons/yacc.c:1652: error: b4_symbol_if: field has_type of 229 is not a Boolean: 
[   70s] /usr/share/bison/skeletons/yacc.c:1652: the top level
[   70s]   YACC     lib/block-ref-grammar.h
```

MrAnno:
There was a bug in Bison 3.3.1 about unused nonterminal symbols that they've fixed in 3.3.2.
I hope it will be upgraded in Leap soon.

https://lists.gnu.org/archive/html/bison-announce/2019-02/msg00000.html

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
There's a prerelease of bison 3.4.1 called 3.3.90, so I've now updated the branch to depend on 3.3.90 or later.

Before merging, we'd at least port bison 3.3.90 to the dbld devshell to be able to generate tarballs OR we need to wait until bison 3.4.x hits the major distros.

kira-syslogng:
Build FAILURE

bazsi:
fyi, bison 3.4 was released and the feature we depend on was included there. so the official dependency of this branch will be bison 3.4

homebrew already has 3.4
suse tumbleweed has 3.4
archlinux is still on 3.3.2
debian sid is 3.3.2
fedora is on 3.0

I guess I would wait until debian sid has 3.4 and then attempt the migration.

bazsi:
in the meantime, I've rebased this to latest master and updated the build dependency on bison 3.4

kira-syslogng:
Build FAILURE

xnox:
Ubuntu & Debian sid have 2:3.4.1+dfsg-2

alltilla:
Thanks for the heads-up! We are planning to wait, until it is updated in Arch, then move forward with the PR.

MrAnno:
Arch has been updated, but we've found multiple issues with the new version.

The first one is a dangling pointer issue which could result in gigabytes of whitespaces on stderr:
https://lists.gnu.org/archive/html/bug-bison/2019-08/msg00002.html
https://lists.gnu.org/archive/html/bison-patches/2019-08/msg00007.html

We've also found leaks that seem to be new, it might be caused by the Bison upgrade. @Kokan is investigating this one.

Kokan:
The reproduction of the issue I am dealing with:
* compile with bison 3.4
* start syslog-ng with default configuration
* stop syslog-ng
* syslog-ng crashes in `free` when in date parser tries to free formats: `string_list_free(self->date_formats);`

The corresponding grammar is:
```bison
date_parser_option
        : KW_FORMAT '(' string_list ')'             { printf("%s\n", $3); date_parser_set_formats(last_parser, $3); }
        | KW_TIME_ZONE '(' string ')'               { date_parser_set_timezone(last_parser, $3); free($3); }

```

The print writes a string instead of some memory junk (so I think somehow instead of `string_list` rule the `string` rule is applied here).
The strangest when I change the next line from `string` to `string_list`, it works just fine.
This gives me the impression that something went wrong creating the state machine (diffed the states, and there is some change I do not yet understand).
I'll try to reproduce without syslog-ng and report upstream.

Kokan:
There are several issues with newer bison (see comments above) so it would not be the best action to upgrade the minimal version as of now. We should wait till those issues are fixed and released. When released and available in stable distributions that point imho we could take a big step and use this patch. Closing this PR until that time.

kira-syslogng:
Build FAILURE

bazsi:
I think it is worth keeping this open to track the branch. In any case, the distro status is as follows:

Debian - "testing" has 3.4.2 (fixing the whitespace issue), buster only 3.3.2
Ubuntu - "disco" has 3.3.2, current dev has 3.4.1, but I guess that would be updated by the time of the release from Debian
Arch - 3.4.2
Fedora - 3.0.5

The memory-leak/corruption issue is definitely a blocker for this item, so putting back to WIP.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
I have rebased and retested the entire branch and I couldn't find any bison related problems: not the string_list case you mention above and not even any new memory leaks. I was using the default configuration file, and I did find a leak (which I have fixed in the branch now), but that was not related  to bison. Are you sure you regenerated/rebuilt all the bison files properly and compiled syslog-ng with those?

I have also added bison 3.4.2 to our devshell image (with a hack, probably should be removed before merging).

With all this I think we should do the following:
- make sure we have a "single way" of doing a release (e.g. continue the work on ci.syslog-ng.com and the Jenkinsfile to automate official releases and not just snapshots)
- this mechanism would use dbld to cut releases
- make sure the "official" dbld build image to build tarballs contains bison 3.4.2+ 
- currently TARBALL_IMAGE is set to ubuntu-bionic. eoan will probably have the right bison, but that's not going to happen before 2020 april. If I get around to do the stuff above, I might change TARBALL_IMAGE to Debian testing or flesh out a dedicated "release" image just like devshell.

With those, we could do a release and build binaries from all dbld-supported distros, we would however lose the ability to build our yacc files on anything older. (e.g. the yacc would need to be built and .c files copied over to compile anything interactively).

This will _NOT_ get better until we are supporting the current distro versions, e.g. we are either merging something like this OR we'll be waiting a "decade" to switch to the newer bison version.

With that said, I would do the transition as soon as the release bits are ready.
What do you tihnk?

kira-syslogng:
Build FAILURE

bazsi:
The current timeline is that we would merge this once Ubuntu disco contains bison 3.4.2

bazsi:
ubuntu 20.04 seems to contain bison 3.4.2, finally. But that's only due in April.

kira-syslogng:
Build FAILURE

bazsi:
Rebased against latest master, resolved all conflicts and updated modules that were merged between
now and the time this was written.

This is still contains the "borrowal" of the bison package from sid in our devshell, but the true solution would be to update our TARBALL_IMAGE and devshell to Ubuntu focal, where the bison version is already available.

kira-syslogng:
Build FAILURE

bazsi:
Rebased against latest master. This makes an attempt to install bison from Debian sid into our travis
environment. The fate of that change needs to be decided once all dependencies go in and we are ready to merge this.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

MrAnno:
I'm currently testing this branch internally with Bison v2.6.4.

kira-syslogng:
Build FAILURE

dbld/images/devshell.dockerfile
```diff
@@ -9,3 +9,4 @@ RUN /dbld/builddeps install_apt_packages
 RUN /dbld/builddeps install_pip2
 RUN /dbld/builddeps install_pip2_packages
 RUN /dbld/builddeps install_pip3_packages
+RUN /dbld/builddeps install_bison_from_sid
```
MrAnno, line 12: This is not necessary in the new devshell image.

lib/cfg-grammar.y
```diff
@@ -69,21 +69,24 @@ extern struct _HostResolveOptions *last_host_resolve_options;
 extern struct _StatsOptions *last_stats_options;
 extern struct _LogRewrite *last_rewrite;
 
+#define MAIN_TOKENTYPE 1
```
MrAnno, line 72: I might be mistaken, but this seems to be a define trick to avoid redefinitions in plugins.

Unfortunately, this does not work starting from Bison 3.6, because the MAIN_EMPTY token's existence depends on this macro:

```
lib/cfg-grammar.c:3469:12: error: ‘MAIN_EMPTY’ undeclared (first use in this function); did you mean ‘MAIN_EOF’?
 3469 |   yychar = MAIN_EMPTY; /* Cause a token to be read.  */
      |            ^~~~~~~~~~
      |            MAIN_EOF
```

I might be a bug in Bison 3.6, of course, I'll check it.

modules/native/native-grammar.ym
```diff
@@ -36,13 +36,13 @@
 #include "syslog-names.h"
 #include "messages.h"
 #include "plugin.h"
-#include "cfg-grammar.h"
+// #include "cfg-grammar.h"
```
MrAnno, line 39: I'm investigating this one as well because I can't build this module internally with Bison 3.5-3.6.

kira-syslogng:
Build FAILURE

bazsi:
The dependencies of this PR have gone in, however we fail to compile again on bison 3.6, which has arrived to Debian sid. Putting this back to WIP.

bazsi:
I have now rebased this against latest master and also fixed the compile issue with bison 3.6+

So we should be able to work with 3.4.2 onwards and we now don't depend on internal bison interfaces.

bazsi:
I'd appreciate help with cmake, namely the configure script now exports the bison version as SYSLOG_NG_BISON_MAJOR/MINOR preprocessor defines as bison's skeleton code doesn't have anything that we could use to find out which bison was used to generate the parser.

The same logic would need to be added to cmake as well, otherwise cmake builds will only work with bison 3.6+

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
So, all the cmake related builds are red, kira uses bison 3.0.4, lgtm uses 3.4.1

bazsi:
A humble attempt at implementing cmake support. All I have to say that my hatred for cmake has only intensified in the process.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

mitzkia:
@bazsi  in my opinion 'Kira-starter' failing should fix this code:

```bash
diff --git a/dbld/builddeps b/dbld/builddeps
index 67fa85b99..6d9173a0f 100755
--- a/dbld/builddeps
+++ b/dbld/builddeps
@@ -179,7 +179,7 @@ function install_pip_packages {
         ubuntu-focal)
             python_executables="python3"
             ;;
-        devshell)
+        devshell|kira)
             python_executables="python2 python3"
             ;;
         *)
diff --git a/dbld/images/kira.dockerfile b/dbld/images/kira.dockerfile
index f57f6a00a..2aebf89fb 100644
--- a/dbld/images/kira.dockerfile
+++ b/dbld/images/kira.dockerfile
@@ -4,4 +4,5 @@ LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <las
 ENV OS_PLATFORM kira
 
 RUN /dbld/builddeps install_apt_packages
+RUN /dbld/builddeps install_pip2
 RUN /dbld/builddeps install_pip_packages
```

bazsi:
Rebased against latest master. I've commited the kira related changes @mitzkia offered helpfully and now kira seems to have started. There's still an issue with an older clang in one of the builds, which I'll have to check.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

mitzkia:
@bazsi  I will update Kira testdb to make tests pass on Ubuntu Focal.

mitzkia:
@kira-syslogng retest this please test branch=pzolee-ssl-4k-keys;

kira-syslogng:
Build FAILURE

mitzkia:
@bazsi about the Kira failed tests: we have created 2 separate testdb branches for the failures, which seems are correct. We need some additional internal test runs, than we can merge them.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

mitzkia:

kira-syslogng:
Build FAILURE

mitzkia:
@bazsi Previously there were 2 different Testdb failures in Kira.
1, TLS key size was too small, we have fixed this, and the corresponding testdb MR was merged, in the last Kira job this testcases are passed.
2, (This is still an issue, but as I think not in our testdb) In python related testcases syslog-ng failed to start with:
```
Error parsing parser expression, parser plugin python not found
```
This can be because (configure summary):
```
python                      : no (pkg-config package: python2)
```
Maybe pkg-config searches for a `python2` binary which is not exist in Focal.

alltilla:
I added `python2-dev` and `python2-dbg` to the kira image. `--with-python=2` works now.

See: https://github.com/bazsi/syslog-ng/pull/6

kira-syslogng:
Build SUCCESS

alltilla:
LGTM does not have fresh enough bison :(
https://lgtm.com/projects/g/syslog-ng/syslog-ng/logs/rev/pr-6e37c814e66c231dea579f91247b62c1feed912c/lang:cpp/stage:Build%20merge_6b6dd8ea44b6c8a40c04faf67eedaba8d051f08c
```
[2020-10-09 08:40:42] [build] configure: WARNING: bison is found, but your bison version 3.4.1 is not recent enough, at least 3.4.2 is required
[2020-10-09 08:40:42] [build] configure: error: syslog-ng requires bison 3.4.2 or later. Your source tree seems to be from git, which doesn't have the bison generated files (like cfg-grammar.c). Please install/upgrade bison or use a distribution tarball.
```

alltilla:
@bazsi 
I created a patch, where we build bison 3.6.4 on LGTM's build environment. https://github.com/bazsi/syslog-ng/pull/7
Can you try this out?

Thanks!

kira-syslogng:
Build SUCCESS

bazsi:
@lbudai  it seems to be green and there was team consensus on this being in good shape. can you please do another review and remove your "change requested" status?

thanks.

I think this is finally ready to go in.

alltilla (approved):

lbudai (approved):

bazsi:
\o/ finally, we are working with the latest bison versions without warnings.

@gcsideal

PR #2528, opened 2019-02-01
logthrdestdrv: factor out _process_result_xxx functions

lbudai:
small refactor (I think it is easier to modify the `process_result` function after this change)


kira-syslogng:
Build SUCCESS

PR #2530, opened 2019-02-04
redis: consume ref at set_command

alltilla:
Similar issue as what we have seen in #2514.

### DIFF
--- a/modules/redis/redis.h
+++ b/modules/redis/redis.h
@@ -31,9 +31,9 @@ LogDriver *redis_dd_new(GlobalConfig *cfg);
 void redis_dd_set_host(LogDriver *d, const gchar *host);
 void redis_dd_set_port(LogDriver *d, gint port);
 void redis_dd_set_auth(LogDriver *d, const gchar *auth);
-void redis_dd_set_command(LogDriver *d, const gchar *command,
-                          LogTemplate *key,
-                          LogTemplate *param1, LogTemplate *param2);
+void redis_dd_set_command_ref(LogDriver *d, const gchar *command,
+                              LogTemplate *key,
+                              LogTemplate *param1, LogTemplate *param2);
 LogTemplateOptions *redis_dd_get_template_options(LogDriver *d);
 
 #endif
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

modules/redis/redis.h
```diff
@@ -34,6 +34,9 @@ void redis_dd_set_auth(LogDriver *d, const gchar *auth);
 void redis_dd_set_command(LogDriver *d, const gchar *command,
                           LogTemplate *key,
                           LogTemplate *param1, LogTemplate *param2);
+void redis_dd_set_command_ref(LogDriver *d, const gchar *command,
```
Kokan: Does it really make sense to create a new function which calls the original, and now the original is not used only by the new function ? It would if it were some kind of API, but it is not.

The whole think just gets too complicated, what is a downside of assuming that the set function grabs ownership as well ?

bazsi:
we could just rename _set_x() to _set_x_ref() so it documents that it grabs
the reference and get rid off the simple set.

modules/redis/redis.h
alltilla: Agreed, I will fix it as Bazsi suggested :)

alltilla: I updated the PR with renaming the `set` function and unrefing the templates.
Could I just remove the `log_template_ref` calls and not even call `unref` later? It could be easier to read.

alltilla: I see now, if i remove the `ref` call, `g_assert(s->ref_cnt > 0);` will not be checked, so I guess it is needed after all.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/redis/redis.h
alltilla: Thanks! I think I finally have the desired fix :D

Kokan: Thanks.

kira-syslogng:
Build SUCCESS

PR #2531, opened 2019-02-04
LogReader refactor + performance improvements

bazsi:
This branch eliminates a lot of system calls on the LogReader input handling path, thereby increasing performance, especially if the per-wakeup overheads cannot be amortized over a lot of messages (e.g. it does not help much in case we are running with log-fetch-limit(1000) and sufficient window, but it does if we go to sleep every 10 messages).

The performance change is basically that we don't call  `iv_unregister_fd()` at every input callback, rather we'll only set the interesting poll mask to zero. This removes a number of epoll_ctl/setsockopt calls, whose overhead can be quite visible.

On the other hand this branch also contains a quite significant refactor of the LogReader, by clearing up various internal layers. These changes are:

* "watches": are the set of events that we need wrt. the input protocols (poll_events and idle_timeout)
  - we can start/stop/disable watches
  - init starts, deinit stops watches
  - at every input processing we disable watches (in contrast to stopping them).

* I've removed the log_reader_reopen() concept, as we are not using that anymore, this has become log_reader_open(), and this allowed me to remove complexity

* I've reordered a functions to make it more amangeable, related concepts are moved to the same place, with a "comment flag" on top to indicate what each is. Otherwise I was jumping back and forth in the source code a lot.

Most of these changes were doing in a pair programming session with @gaborznagy

### DIFF
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -192,12 +192,20 @@ log_reader_window_empty(LogSource *s)
 }
 
 static void
+log_reader_disable_watches(LogReader *self)
+{
+  g_assert(self->watches_running);
+  if (self->poll_events)
+    poll_events_suspend_watches(self->poll_events);
+  log_reader_stop_idle_timer(self);
+}
+
+static void
 log_reader_io_handle_in(gpointer s)
 {
   LogReader *self = (LogReader *) s;
 
-  log_reader_stop_watches(self);
-  log_reader_stop_idle_timer(self);
+  log_reader_disable_watches(self);
   if ((self->options->flags & LR_THREADED))
     {
       main_loop_io_worker_job_submit(&self->io_job, G_IO_IN);
@@ -250,12 +258,12 @@ log_reader_init_watches(LogReader *self)
 static void
 log_reader_stop_watches(LogReader *self)
 {
-  if (self->watches_running)
-    {
-      poll_events_stop_watches(self->poll_events);
+  g_assert(self->watches_running);
+  if (self->poll_events)
+    poll_events_stop_watches(self->poll_events);
+  log_reader_stop_idle_timer(self);
 
-      self->watches_running = FALSE;
-    }
+  self->watches_running = FALSE;
 }
 
 static void
@@ -266,13 +274,12 @@ log_reader_stop_idle_timer(LogReader *self)
 }
 
 static void
-log_reader_start_watches_if_stopped(LogReader *self)
+log_reader_start_watches(LogReader *self)
 {
-  if (!self->watches_running)
-    {
-      poll_events_start_watches(self->poll_events);
-      self->watches_running = TRUE;
-    }
+  g_assert(!self->watches_running);
+  if (self->poll_events)
+    poll_events_start_watches(self->poll_events);
+  self->watches_running = TRUE;
 }
 
 static gboolean
@@ -285,7 +292,7 @@ static void
 log_reader_suspend_until_awoken(LogReader *self)
 {
   self->immediate_check = FALSE;
-  poll_events_suspend_watches(self->poll_events);
+  log_reader_disable_watches(self);
   self->suspended = TRUE;
 }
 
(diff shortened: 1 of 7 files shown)


kira-syslogng:
Build SUCCESS

lib/logreader.c
```diff
@@ -196,7 +196,7 @@ log_reader_io_handle_in(gpointer s)
 {
   LogReader *self = (LogReader *) s;
 
-  log_reader_stop_watches(self);
+  poll_events_suspend_watches(self->poll_events);
```
gaborznagy: Fd polling will be restored in `update_watches`, 
- in `log_reader_start_watches_if_stopped` which no longer will be called unless `log_reader_stop_watches` is not called prior.
I think a registered immediate_check (`restart_task`) must be stopped if we got a poll event.
- in `poll_events_update_watches` fd handlers changed according to `log_proto_server_prepare` result

Shouldn't be the `log_reader_stop_watches` changed to use `poll_events_suspend` instead of `poll_events_stop_watches`?
Of course in deinit, we need to call unregister, so the same stop function wouldn't be feasible.

gaborznagy:
We should do this on LogWriter side as well, right?

kira-syslogng:
Build SUCCESS

lib/logreader.c
bazsi: We've addressed this while we've worked on the refactor.

gaborznagy:
@kira-syslogng do perftest

kira-syslogng:

gaborznagy:
@bazsi I have only a minor comment, otherwise it is approved from my side and it's ready for merge.

lib/logreader.c
```diff
@@ -285,7 +292,7 @@ static void
 log_reader_suspend_until_awoken(LogReader *self)
 {
   self->immediate_check = FALSE;
-  poll_events_suspend_watches(self->poll_events);
+  log_reader_disable_watches(self);
```
Kokan: Disabling the `idle_timer` during *suspend* does make things awkward, as the `LogReader` and the `LogProtoServer` has an interface that allow the proto to set the *timeout* value for `idle_timer` (`log_proto_server_prepare`). Where it has to specify a `LogProtoPrepareAction` and can change the *timeout*.

From the proto point of view it makes sense to ask the `LogReader` not to do anything (`LPPA_SUSPEND`), and set an amount of time to keep the "connection".
This breaks any proto that may want to set `idle_timeout` now.

MrAnno: It is an interesting question whether we want the idle timer running when a proto suspends the source with `LPPA_SUSPEND`.

Other callers of `log_reader_suspend_until_awoken()` are interesting too. When a destination (with flow control) suspends the source, it seems intuitive to stop the timer, since we want to close _idle_ connections, and a suspended connection is obviously not idle (it was actually active and faster than the destination, this triggered the suspension).

However, we can't detect a closed or broken connection when the source is suspended, so a timeout could actually help this scenario. There are other ways to do something with closed/broken connections, I just wanted to mention this.

gaborznagy:
@bazsi:
After our discussion, I have put a fixup commit on top of your branch:
> In log_reader_force_check_in_next_poll: Can we replace poll_events_suspend_watches to log_reader_disable_watches ?
I think we can stop the idle_timer too (well it will be anyway in io_handle_in, but it would be similar to suspend).
It would make sense too, as we know we have to do an I/O operation (handshake or actual fetch), we should avoid that the idle_timer timeouts before the iv_task is scheduled.

Also, please share your opinion about suspended sources and using/stopping the idle timer.
See [conversation](https://github.com/balabit/syslog-ng/pull/2531#pullrequestreview-230665347).

I really would like this PR in OSE 3.21, as we made the refactor almost a month ago!

lib/logreader.c
gaborznagy: We just had a discussion with @szemere about this topic from `LogProtoServer` point of view.
My argument was, that why would we need to care if a connections is broken on a suspended connection?
Clients would try to re-connect to our server which was just suspended, probably making a new connection, and eating up another window size from the queue.

Also, the broken link can be re-established while it is suspended, which would make the error transparent, right?

With all that, I think we can stop the idle timer when a source is suspended.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/logreader.c
gaborznagy: >We discussed the open question in the team meeting, double checked the code, and decided to stick with the current version.

Issue #2368, opened 2018-10-22
python source crashes if post called from a thread


furiel:
I found a crash while experimenting with python source. The crash happens if the post is called from another thread, spawned from the user code. Without any destination, there is no crash. However, there was a crash with either file destination and network destination.

```
@version: 3.18

log {
  source { python(class(PySource) flags(no-parse)); };
  destination { file(/dev/stdout); };
#  destination { network(localhost port(5555)); };
};

python {
from syslogng import LogSource, LogMessage
import threading

class PySource(LogSource):
    def run(self):
        #self.post_message(LogMessage("example"))

        t = threading.Thread(target=self.post_message, args=(LogMessage("example"),))
        t.start()
        t.join()

    def request_exit(self):
        pass
};
```

Backtrace
```
#0  0x00007fdb6b291c58 in g_cond_wait () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007fdb6b5676ac in main_loop_call (func=0x7fdb67652051 <affile_dd_open_writer>,
    user_data=0x7fdb67436370, wait=1)
    at [...]/syslog-ng/lib/mainloop-call.c:94
#2  0x00007fdb67652382 in affile_dd_queue (s=0xbf13b0, msg=0x7fdb60003730,
    path_options=0x7fdb67436430)
    at [...]/syslog-ng/modules/affile/affile-dest.c:693
#3  0x00007fdb6b558c5d in log_pipe_queue (s=0xbf13b0, msg=0x7fdb60003730,
    path_options=0x7fdb67436430)
    at [...]/syslog-ng/lib/logpipe.h:362
#4  0x00007fdb6b558ebb in log_multiplexer_queue (s=0xb6b820, msg=0x7fdb60003730,
    path_options=0x7fdb67436800) at [...]/syslog-ng/lib/logmpx.c:96
#5  0x00007fdb6b551729 in log_pipe_queue (s=0xb6b820, msg=0x7fdb60003730,
    path_options=0x7fdb67436800)
    at [...]/syslog-ng/lib/logpipe.h:362
#6  0x00007fdb6b5515d7 in log_pipe_forward_msg (self=0xb6b790, msg=0x7fdb60003730,
    path_options=0x7fdb67436800)
    at [...]/syslog-ng/lib/logpipe.h:328
#7  0x00007fdb6b551742 in log_pipe_queue (s=0xb6b790, msg=0x7fdb60003730,
    path_options=0x7fdb67436800)
[...]

PR #2533, opened 2019-02-05
python-source: fix crash when posting from python thread

furiel:
Fixes: https://github.com/balabit/syslog-ng/issues/2368

In syslog-ng, logwriters are lazily created: the first message
triggers the creation, and the creation is done by the posting
thread. To avoid race condition, the process is done through a
`main_loop_call()` function. So that to work, on needs to initialize
the posting thread with `main_loop_call_thread_init()`. However, this
api is not exposed to python, hence posting from thread cannot work
from a python thread other than the one started by the threaded
source.

This patch saves the assigned thread id, and compares the current
thread id of the `post_message` caller. In case of mismatch, exception
is thrown.

### DIFF
--- a/modules/python/python-source.c
+++ b/modules/python/python-source.c
@@ -25,6 +25,7 @@
 #include "python-logmsg.h"
 #include "python-helpers.h"
 #include "logthrsource/logthrsourcedrv.h"
+#include "thread-utils.h"
 #include "str-utils.h"
 #include "string-list.h"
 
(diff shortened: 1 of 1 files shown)


modules/python/python-source.c
```diff
@@ -434,6 +436,20 @@ static PyObject *
...
+    {
+      /*
+         Message posting must happen in a syslog-ng thread that was
+         initialized by main_loop_call_thread_init(), which is not
+         exposed to python. Hence posting from a python thread can
+         crash syslog-ng.
+      */
+
+      PyErr_Format(PyExc_RuntimeError, "post_message must be called from main thread");
+      return 0;
```
MrAnno: Please return either `NULL` or the return value of `PyErr_Format` itself.

kira-syslogng:
Build SUCCESS

Issue #1870, opened 2018-02-01
Template substitution not working in Elasticsearch2 destination


micos76:
# syslog-ng

(Please, remove the unnecessary notes, including this one.)

## Version of syslog-ng

```
syslog-ng 3 (3.13.2)
Config version: 3.13
Installer-Version: 3.13.2
Revision: 3.13.2-1
Compile-Date: Dec  5 2017 13:31:27
Module-Directory: /usr/lib/syslog-ng/3.13
Module-Path: /usr/lib/syslog-ng/3.13
Available-Modules: date,map-value-pairs,afamqp,csvparser,xml,riemann,syslogformat,system-source,json-plugin,tfgetent,afstomp,afsql,snmptrapd-parser,afsocket,http,cef,pseudofile,disk-buffer,geoip-plugin,mod-python,appmodel,confgen,afprog,afuser,tags-parser,kvformat,affile,cryptofuncs,graphite,afsmtp,basicfuncs,stardate,mod-java,geoip2-plugin,linux-kmsg-format,afmongodb,add-contextual-data,redis,sdjournal,dbparser,pacctformat
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: on
```

## Platform

Linux 4.4.0-112-generic #135-Ubuntu SMP Fri Jan 19 11:48:36 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

# Issue

When using the Elasticsearch2 destination with a custom defined template it does not get substituted correctly with the contents of the template.
Using the same template in a file destination instead works correctly.

## Failure
When trying to send some parsed data in JSON format to ElasticSearch syslog-ng was showing the following error:

`syslog-ng    | 17:59:00.897 [?3?] ERROR  - {"root_cause":[{"type":"mapper_parsing_exception","reason":"failed to parse"}],"type":"mapper_parsing_exception","reason":"failed to parse","caused_by":{"type":"not_x_content_exception","reason":"Compressor detection can only be called on some xcontent bytes or compressed xcontent bytes"}}
`
The only way for me to debug this issue was to capture network traffic between syslog-ng and Elasticsearch. I then realised that whatever template name I was specifying in the ES destination the name of the template itself was being sent as string data to ES instead of JSON objects.

`**template(t_sonicwall)**`

[...]

PR #2534, opened 2019-02-05
java, elasticsearch2, explicit unsupport for named templates

furiel:
Fixes: https://github.com/balabit/syslog-ng/issues/1870

As reported in #1870, elasticsearch destination does not support named templates. We concluded in that thread that there is no plan to support for it.

This patchset adds error detection in case someone tries to use named templates, so that the problem does not go undetected.

I added similar detection for java destination too. Interestingly, elasticsearch2 destination did not use the template grammar from java destination, so different implementation was needed there.

The motivation is that we would like to promote `@define` keyword instead of named templates.


kira-syslogng:
Build SUCCESS

Issue #1832, opened 2017-12-30
compile syslog-ng python module for python3


mochrul:
# syslog-ng

## Version of syslog-ng

3.13.2

## Platform

Debian (unstable)

# Issue

As part of the compile, the build system tries to run pylint and nosetests.
But on a Debian system, the python3 counterparts of this programs are pylint3 and nosetests3.

mochrul:
I would like to extend this with the issue at modules/python/pylib/setup.py.

That script is called from the Makefile.am and use the $(PYTHON) as the interpreter.
That $(PYTHON) variable comes from the configure.ac, but can only contain the python executable, which can be a link to a python2 executable only. And this cause that the setup.py will install the python scripts into a python2 related path. The configure.ac does not even honor the --with-python option when deals with this.

furiel:
Since https://github.com/balabit/syslog-ng/pull/1868, python2 is not a hard dependency for syslog-ng anymore. You can install syslog-ng on a pythonv3-only system. @mochrul Is this problem still relevant?

mochrul:
From Debian point of view, yes.

The trick here is that Debian follows that python recommendation, that the python2 and python3 implementation should be totally separated. Which means, that if you try to run pylint for python3, you should run pylint3. And the same is required for nosetests. As far as I followed the development these issues are still not remediated. And the second comment is still valid too I think.

The other ticket was about "running" syslog-ng on a python3 environment. This is about compiling it.

furiel:
Ok thanks for explaining. Just reciting all above: the problem is there is no detection for pylint and nosetests, they are called directly from the code. We need to detect the correct binary names (honoring with-python), and call those binaries using the detected name.

PR #2537, opened 2019-02-06
autotools, cmake: add detection for pylint, nosetests

furiel:
Fixes: https://github.com/balabit/syslog-ng/issues/1832

The patchset uses detection for pylint nosetests:
for python2 we detect pylint, nosetests
for python3 we detect pylint3, nosetests3

If deployment does not use these recommendations, users can override binary with `--with_pylint=` and `--with_nosetests` for autotools, or `-DPYLINT=` and `-DNOSETESTS` for cmake.


kira-syslogng:
Build SUCCESS

configure.ac
```diff
@@ -1375,13 +1375,21 @@ dnl ***************************************************************************
 if test "x$with_nosetests" != "x"; then
     NOSETESTS=$with_nosetests
 else
-    AC_CHECK_PROG(NOSETESTS, nosetests, nosetests)
+    if (echo "$with_python" | grep -Eq "python-?3.*"); then
```
Kokan: [optional] The same *grep* is used to create the `ENABLE_PYTHONv2/3`, I think it would be beneficial if the *magic* would be in one place, and we have `enable_pythonv2/3` variable in configure  (or `python_version == 2/3`).

furiel: Good idea. I will do so.

furiel: I just force pushed the correction. During teh fix I saw another problem: nosetests/pylint detection was still executed even if python was disabled. I fixed this by checking the `$python_found` variable.

Kokan: Thanks.

kira-syslogng:
Build SUCCESS

PR #2538, opened 2019-02-06
xml/geoip2: make prefix optional

gaborznagy:
Until now, prefix for these parsers were mandatory: it was either the default prefix, or a user-defined.
After this PR, it is possible to set an empty prefix for geoip2 and xml parsers.
```
xml(prefix(""));
geoip2(prefix(""));
```
The resulting nv-pairs would be created without a prefix:
```
name1 = value1
name2 = value2
```

### DIFF
--- a/modules/xml/tests/test_xml_parser.c
+++ b/modules/xml/tests/test_xml_parser.c
@@ -156,10 +156,8 @@ ParameterizedTest(ValidXMLTestCase *test_cases, xmlparser, valid_inputs)
   log_msg_unref(msg);
 }
 
-Test(xml_parser, test_drop_invalid)
+Test(xmlparser, test_drop_invalid)
 {
-  setup();
-
   LogParser *xml_parser = _construct_xml_parser((XMLParserTestOptions) {});
 
   LogMessage *msg = log_msg_new_empty();
@@ -175,8 +173,6 @@ Test(xml_parser, test_drop_invalid)
   log_pipe_deinit((LogPipe *)xml_parser);
   log_pipe_unref((LogPipe *)xml_parser);
   log_msg_unref(msg);
-
-  teardown();
 }
 
 typedef struct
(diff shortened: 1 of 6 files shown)


bazsi:
The dot should be part of the prefix. There are usecases for both dotted
and dotless versions.

kira-syslogng:
Build SUCCESS

gaborznagy:
> The dot should be part of the prefix. There are usecases for both dotted and dotless versions.

@bazsi: Do you mean that if the prefix set to empty, the parser generated name-value pairs must always be started with a dot?
I.e. `name = value` is not valid, but `.name = value` always?

bazsi:
No, if the prefix is empty, no prefix should be generated. Not even a dot.

gaborznagy:
Setting WIP flag:
geoip2/xml needs more refactor to be able to use an empty prefix:
`prefix("") ` -> `name = value` and not `.name = value`
- xml: key removing in `before_last_dot`

@bazsi: I find it interesting, that with other parsers (e.g. kv-parser()) if a prefix is missing a trailing dot,
then each name-value pairs will miss the dot separating the prefix and the name:
`prefix(".geoip")` -> `.geoipcity = Budapest` and `.geoipcountry = Hungary`.
I don't think that would be a use case.

kira-syslogng:
Build FAILURE

gaborznagy:
Updated the PR: if no prefix is configured xml/geoip2 parser will create name-value pairs without a leading dot:
`prefix("") ` -> `name = value` and not `.name = value`

kira-syslogng:
Build SUCCESS

modules/xml/tests/test_xml_parser.c
```diff
@@ -299,14 +293,10 @@ Test(xml_parser, test_multiple_exclude_tags)
   log_msg_unref(msg);
 
   g_list_free(exclude_tags);
-
-  teardown();
 }
 
-Test(xml_parser, test_strip_whitespaces)
+Test(xmlparser, test_strip_whitespaces)
```
MrAnno: Nice catch!

PR #2541, opened 2019-02-07
scl: add parser for netskope logs

bazsi:

### DIFF
--- a/scl/netskope/plugin.conf
+++ b/scl/netskope/plugin.conf
@@ -1,0 +1,43 @@
+#############################################################################
+# Copyright (c) 2019 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+
+# sample log
+#   <134>{"count": 1, "supporting_data": {"data_values": ["x.x.x.x", "user@domain.com"], "data_type": "user"}, "organization_unit": "domain/domain/Domain Users/Enterprise Users", "severity_level": 2, "category": null, "timestamp": 1547421943, "_insertion_epoch_timestamp": 1547421943, "ccl": "unknown", "user": "user@domain.com", "audit_log_event": "Login Successful", "ur_normalized": "user@domain.com", "_id": "936289", "type": "admin_audit_logs", "appcategory": null}
+
+# assumed to be running with flags(no-parse), e.g. the entire incoming log
+# message is in $MSG
+block parser netskope-parser(prefix(".netskope.")) {
+    channel {
+        rewrite {
+            subst("^<[0-9]+>", "");
+        };
+        parser {
+            json-parser(prefix(`prefix`));
+	};
+    };
+};
+
+application netskope[syslog-raw] {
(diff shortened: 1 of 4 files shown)


scl/netskope/plugin.conf
```diff
@@ -0,0 +1,22 @@
+
+
+# sample log
+#   <134>{"count": 1, "supporting_data": {"data_values": ["x.x.x.x", "user@domain.com"], "data_type": "user"}, "organization_unit": "domain/domain/Domain Users/Enterprise Users", "severity_level": 2, "category": null, "timestamp": 1547421943, "_insertion_epoch_timestamp": 1547421943, "ccl": "unknown", "user": "user@domain.com", "audit_log_event": "Login Successful", "ur_normalized": "user@domain.com", "_id": "936289", "type": "admin_audit_logs", "appcategory": null}
+
+# assumed to be running with flags(no-parse), e.g. the entire incoming log
```
szemere: Can we also assume that netskope logs are coming over network? In this case we can "include the `flags(no-parse)` assumption" by creating a whole source block instead of a parser.

kira-syslogng:
Build SUCCESS

bazsi:
That is coming from default-network-drivers(). Theres no point in creating
a separate source for all device types.

kira-syslogng:
Build SUCCESS

scl/netskope/plugin.conf
```diff
@@ -0,0 +1,43 @@
...
+#
+#############################################################################
+
+
+# sample log
+#   <134>{"count": 1, "supporting_data": {"data_values": ["x.x.x.x", "user@domain.com"], "data_type": "user"}, "organization_unit": "domain/domain/Domain Users/Enterprise Users", "severity_level": 2, "category": null, "timestamp": 1547421943, "_insertion_epoch_timestamp": 1547421943, "ccl": "unknown", "user": "user@domain.com", "audit_log_event": "Login Successful", "ur_normalized": "user@domain.com", "_id": "936289", "type": "admin_audit_logs", "appcategory": null}
+
+# assumed to be running with flags(no-parse), e.g. the entire incoming log
+# message is in $MSG
+block parser netskope-parser(prefix(".netscope.")) {
```
gaborznagy: I guess it's a typo : `".netscope."`

scl/netskope/plugin.conf
```diff
@@ -0,0 +1,43 @@
...
+            subst("^<[0-9]+>", "");
+        };
+        parser {
+            json-parser(prefix(`prefix`));
+	};
+    };
+};
+
+application netskope[syslog-raw] {
+	filter { message("<134>{" type(string) flags(prefix)) and message("_insertion_epoch_timestamp"); };
```
gaborznagy: Can you please link a reference document for netskope log format? I have not found any with a quick search.
It seems that "<134>" is hard coded to every message, and `_insertion_epoch_timestamp` is a fingerprint as well.

bazsi:
Both of these are true observations. This is a heuristics to fix the logs
coming from Netskope and the format is potentially not documented and can
change any time.

kira-syslogng:
Build SUCCESS

scl/netskope/plugin.conf
gaborznagy: fixed

gaborznagy:
>Both of these are true observations. This is a heuristics to fix the logs coming from Netskope and the format is potentially not documented and can change any time.

Then is it worth to include something fragile into `default-network-drivers` via app-parser?
How are we gonna test that it's broken?

It sound to me, that the `netskope-parser` is enough.
I know, Cisco is the same: it uses heuristics and the format can broke as well, but integrating that many things by default that can broke easily seems to be hard to maintain.

scl/netskope/plugin.conf
bazsi: quoting a comment I sent via email: That is coming from default-network-drivers(). Theres no point in creating a separate source for all device types.

bazsi:
> Then is it worth to include something fragile into default-network-drivers via app-parser?
> How are we gonna test that it's broken?

Well, we won't. These are "as is" contributions that anyone is welcome to adapt to their own liking and PRs are appreciated. Bugreports are also appreciated, which we should handle on a best effort basis.

Usually though, if there's a problem, we already have a log sample to fix the parser with. So in fact, these kind of bugreports are a good source of getting log samples :)

And these kind of support levels are pretty standard, log formats are breaking everywhere, still the value that we deliver is that we compiled them to make it easy to use and then make it easy to fix/upgrade.

bazsi:
@MrAnno was nice enough to fix the outstanding issues which I have added to this branch.

Removing wip

kira-syslogng:
Build SUCCESS

PR #2536, opened 2019-02-05
Json list support

bazsi:
json: added support for arrays using the list() type-hints
    
This patch adds support for generating JSON arrays assuming the input is a syslog-ng list.
**example**: `template("$(format-json message=list($MESSAGE))")`

PR #2544, opened 2019-02-08
WIP: xml-parser: add list support

gaborznagy:
XML-parser's current limitation is that it cannot handle array-like structures in the XML.
Example
```
<List>
  <Data>Info 1</Data>
  <Data>Info 2</Data>
</List>
```
Formerly we just concatenated the different tag's content:
`List.Data = "Info 1Info2"`

This concatenated format is unsuitable for the users.
I've added existing list-handling method (escaping values on space and comma, which is the default list separator), this way the user can separately index different values from one name-value pair (NVTable does not support arrays, hence the list).
Result
`List.Data = "Info 1","Info 2"`

What's a drawback is, not just real lists will be escaped, but name-value pairs with single value.
This could surprise users as they don't understand why their text/content got quoted.

Applying list-handling only on array-like data cannot be done unless a
a) native nv-table type support or
b) an additional data structure in xml to handle list-handling for array-like data only
is implemented. None of them exists or suitable.

### Using the newly created lists in config
**As every value in nv-pairs can be quoted, it is highly recommended to access these name-values as lists!!!**
Here are some ways to do that:
* **List related template-functions** can be used on the list created by xml-parser (e.g. `list-count, list-nth`).
* **list type-hint** in `format-json` (since #2536 ): example config
```
template("$(format-json --scope dot-nv-pairs LIST=list(${.xml.Event.EventData.Data}))\n")
```

### SDATA and quoting:

According to RFC5424, SDATA param values must be quoted with double-quote  ('"') characters.
If the value contains double-quotes, they must be escaped with a backslash ('\').

Due to xml-parser() list-handling functionality, parsed xml text contents will be quoted with a '"' too.
As these are part of the message, they will be quoted when put into SDATA param value.

E.g.
```
<Event>
<Data>Not available</Data>
</Event>
```
before xml-parser list-handling:
`[Event Data="Not available"]`
after xml-parser list-handling:
`[Event Data="\"Not available\""]`

### Compatibility
This functionality can be turned off with the new `create_lists(yes/no)` option.                                                                                                                      
The default value set to be "yes".
Example config:
```
parser p_xml {
    xml(create_lists(no));
};
```

**This PR depends on ~#2459~, #2569**.


gaborznagy:
Added WIP flag as it is blocked by #2459 .

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please branch=xml-parser-list-handling

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please branch=xml-parser-list-handling

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please test branch=xml-parser-list-handling

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please branch=xml-parser-list-handling;

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please test branch=xml-parser-list-handling;

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please test branch=xml-parser-list-handling;

kira-syslogng:
Build SUCCESS

gaborznagy:
Marked as WIP as some investigations are still needed before the patch is ready for merge:
- should we introduce this compatibility breaking feature: "not just real lists will be escaped, but name-value pairs with single value" or introduce an option which allows list-generating

gaborznagy:
I have discussed the compatibility question about quoting and every feedback was that this improves the xml-parser logic.

**I have also checked a question about SDATA and quoting:**

According to RFC5424, SDATA param values must be quoted with double-quote  ('"') characters.
If the value contains double-quotes, they must be escaped with a backslash ('\').

Due to xml-parser() list-handling functionality, parsed xml text contents will be quoted with a '"' too.
As these are part of the message, they will be quoted when put into SDATA param value.

E.g.
```
<Event>
<Data>Not available</Data>
</Event>
```
before xml-parser list-handling:
`[Event Data="Not available"]`
after xml-parser list-handling:
`[Event Data="\"Not available\""]`

It seems wrong at first, but imagine if there are multiple nodes
of the same tag.
```
<Event>
<Data>42</Data>
<Data>Dont panic</Data>
</Event>
```
the expected name-value pair is
`Event.Data = 42,"Dont panic"`
putting it into SDATA:
`[Event Data="42,\"Dont panic\""]`
Here it looks natural to escape the double quotes, while in the first example it looks unnecessary.
I don't think that the first case should be handled differently, there should be no heuristics when quoted strings are put into SDATA.

gaborznagy:
@kira-syslogng retest this please test branch=xml-parser-list-handling;

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please test branch=xml-parser-list-handling;

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

bazsi:
I wouldn't use a different representation of lists even if we have native
support for it in nvtable. We could increase performance, still we have to
be able to serialize lists in things like welf.

Also, the $TAGS macro uses this format already, it just doesn't have
escaping, as tags dont have special characters in them.

The list functions should be available in all syslog-ng deployments, it
doesnt have an extra dependency.

Btw, one idea to solve the compatibility issue is to use a separate name
value pair for the very first value (unescaped) and then a separate name
value pair for the list.

Bazsi

gaborznagy:
@MrAnno thanks for your opinion.
@bazsi thanks for the update too!

It was hard to decide how should the list-handling be introduced (optional, or default list-handling for all nv-pairs). Initially I was reluctant to introduce a breaking change, i.e. adding quotes to simple name-value pairs.
I have discussed this internally with several people, including product team.

Currently I think we should be able to have the original behaviour (however bad it was) for users who are sensitive about this extra unwanted quoting (it can be sensitive in SIEMs or just in a filter expression).
For this a boolean config option could help.

@bazsi, how do you see the list/type representation in the config if nvtable supports lists or other types?
With the current patch, the user always has to use list-handling template-functions, since he/she does not know the type/content of his/her data.
Do you think an implicit type resolution would be possible, so templates would be resolved differently in case of a list or simple name-value and we will not need type-hints or template-functions?
e.g. a `{$.xml.System.Provider}` template would produce different things.

bazsi:
It would be great to discuss how list support should be fleshed out in the
future, so we don't surprise each other. Also, I don't have an answer for
these questions yet, hence a discussion would be useful.

gaborznagy:
We have discussed this IRL, in summary:
- we will introduce a new boolean option, .e.g `create-lists()`:
  * on "yes", the newly added list-handling will be used
  * on "no", the former appending without separating gonna be used, so user can fallback if they don't like the quoting introduced with list-handling
- default will be "yes" and a config version update will be added too,
- documentation: using list-handling template functions with xml-parser() will be emphasized.

I've added it to OSE 3.20 milestone.

gaborznagy:
@kira-syslogng retest this please test branch=xml-parser-list-handling;

gaborznagy:
Updated PR and rebased onto the new OSE 3.20 release preparation PR #2569.

Marking it as WIP until #2569 is not merged.

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please test branch=xml-parser-list-handling;

kira-syslogng:
Build SUCCESS

modules/xml/xml.c
```diff
@@ -40,29 +46,42 @@ remove_trailing_dot(gchar *str)
     str[strlen(str)-1] = 0;
 }
 
+static void
+encode_and_append_value(GString *result, const gchar *current_value, gssize current_value_len)
+{
+  if (result->len > 0)
+    g_string_append_c(result, ',');
+  str_repr_encode_append(result, current_value, current_value_len, ",");
```
alltilla: Not the scope for this PR, but I wonder if we really need to escape the special characters (`"\b\f\n\r\t\\ "`), defined in `str_repr_encode_append()` for lists.

Another finding:
```
if (result->len > 0)
  g_string_append_c(result, ',');
str_repr_encode_append(result, additinal_value, len, ",")
```
is used for list element handling in other places as well, maybe we could give a `list_append` function, which would only escape `,` and would do the "check whether this is the first element or not" as well. Actually, that function would look something like `encode_and_append_value()`, so it could be enough to move this function out and be used by the other source codes.
Source codes, where I found similar behavior:
```
lib/logmsg/logmsg.c
lib/scanner/kv-scanner/kv-scanner.c
modules/basicfuncs/context-funcs.c
modules/basicfuncs/list-funcs.c
modules/xml/xml.c
```

gaborznagy: We discussed this in person, thanks for summarizing!
- should `str_repr_encode` quote our string in case of a whitespace, while the only special character in our lists is the comma characeter is a good point. Right now I don't think in this context it is needed.
- adding a higher level abstraction for list-handling is a good idea, I guess there were not many users before. Now kv-parser(), tags-parser(), xml() parser. 
- as you started with, it has to be defined what should be escaped in lists in general, before creating such API.

kira-syslogng:
Build FAILURE

PR #2547, opened 2019-02-11
persist-tool: remove criterion as dependency

furiel:
`persist-tool` was linked to `TEST_LDADD`, which contains `criterion`. This caused problem during deb package generation on debian jessie: the dpkg-buildpackage framework picked criterion as deb package dependency automatically. As a consequence, the resulting deb package could not be installed due to missing criterion package.

Replacing `TEST_LDADD` with the usual dependencies solves the problem.

### DIFF
--- a/persist-tool/Makefile.am
+++ b/persist-tool/Makefile.am
@@ -11,11 +11,12 @@ persist_tool_persist_tool_SOURCES	=	\
 	persist-tool/generate.c \
 	persist-tool/generate.h \
 	persist-tool/persist-tool.c \
-	persist-tool/persist-tool.h 
+	persist-tool/persist-tool.h
 
 persist_tool_persist_tool_LDADD	= \
-  $(TEST_LDADD) \
-	@GLIB_LIBS@ \
-	@BASE_LIBS@
+	$(MODULE_DEPS_LIBS) \
+	$(TOOL_DEPS_LIBS)
 
 EXTRA_DIST += persist-tool/CMakeLists.txt
+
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

furiel:
I was falsely thinking that the patch does not work, but in the end it turned out there was a problem with the test system. I reopened and removed the WIP flag.

kira-syslogng:
Build SUCCESS

persist-tool/Makefile.am
```diff
@@ -11,11 +11,12 @@ persist_tool_persist_tool_SOURCES	=	\
...
 persist_tool_persist_tool_LDADD	= \
-  $(TEST_LDADD) \
-	@GLIB_LIBS@ \
-	@BASE_LIBS@
+	$(MODULE_DEPS_LIBS) \
+	$(TOOL_DEPS_LIBS)
 
 EXTRA_DIST += persist-tool/CMakeLists.txt
+
+persist_tool_persist_tool_DEPENDENCIES	= lib/libsyslog-ng.la
```
gaborznagy: wouldn't it better to have `MODULE_DEPS_LIBS` here too?
from configure.ac:
`MODULE_DEPS_LIBS="\$(top_builddir)/lib/libsyslog-ng.la"`

lbudai:
@gaborznagy: I guesss @furiel just followed the pattern found in `syslog-ng-ctl` ... and I think, it’s fine.

Issue #2457, opened 2018-12-11
syslog-ng (3.18.1) suddenly hangs


mikeBashStuff:
## Version of syslog-ng

```
syslog-ng 3 (3.18.1)
Config version: 3.18
Installer-Version: 3.18.1
Revision: 
Compile-Date: Dec 10 2018 14:21:58
Module-Directory: /usr/lib/syslog-ng
Module-Path: /usr/lib/syslog-ng
Available-Modules: system-source,stardate,cef,kvformat,cryptofuncs,appmodel,afuser,xml,graphite,map-value-pairs,afprog,afstomp,basicfuncs,syslogformat,tags-parser,add-contextual-data,date,tfgetent,affile,snmptrapd-parser,examples,afsocket,confgen,hook-commands,sdjournal,disk-buffer,pseudofile,dbparser,csvparser,linux-kmsg-format
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: off
Enable-TCP-Wrapper: off
Enable-Linux-Caps: off
Enable-Systemd: off
```
Compilation config:
```
./configure \
        --enable-ipv6 \
        --prefix=/usr \
        --sysconfdir=/etc/syslog-ng \
        --localstatedir=/var/run/syslog-ng \
        --disable-linux-caps \
        --disable-java \
        --disable-python \
        --with-mongoc=no \
        --enable-mongodb=no \
        --enable-json=no \
        --with-jsonc=no \
        --disable-legacy-mongodb-options \
        --with-default-modules=affile,afprog,afsocket,afuser,basicfuncs,csvparser,syslogformat
```
## Platform

OS: Linux (embedded distro) 
Kernel version: 4.4.88

# Issue
it's been noticed that syslog-ng, under a bit heavier load, suddenly hangs - it locks on a single cpu thread, not writing to any log files, not responding to most of the signals (including term ones. It dies when you bash it with those that can't be handled|ignored + SEGV). When that happens, every attempt to connect() to /dev/log blocks as well (e.g. via logger). I am able to trigger that behavior with a  ```./stress_syslog 6 ``` where "stress_syslog" is just:
```
#!/bin/bash
trap 'wait' 0 1 2 15

levels=() facilities=()

levels=(
                emerg alert crit
                err warning notice
                info debug panic
                error warn
       )

facilities=(
                auth authpriv cron
                daemon syslog user
                local{0..7} security
           )

[...]

PR #2549, opened 2019-02-12
WIP: logreader: remove an ordering issue between event registration and de…

bazsi:
This is an attempt at fixing #2457 

I just wanted to open the PR to see if @kira-syslogng agrees that it helps the issue.

This was actually in the codebase for a long time, however 33c4716579 makes things worse.

Those opening #2457, I would appreciate if you could try whether this fixes your problem.

### DIFF
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -47,7 +47,7 @@ struct _LogReader
   struct iv_event schedule_wakeup;
   struct iv_event last_msg_sent_event;
   MainLoopIOWorkerJob io_job;
-  gboolean watches_running:1, suspended:1;
+  gboolean watches_running:1, suspended:1, wakeup_events_registered:1;
   gint notify_code;
 
 
@@ -94,6 +94,9 @@ static void
 _last_msg_sent(gpointer s)
 {
   LogReader *self = (LogReader *) s;
+
+  if (!(self->super.super.flags & PIF_INITIALIZED))
+    return;
   log_pipe_notify(self->control, NC_LAST_MSG_SENT, self);
 }
 
@@ -151,6 +154,9 @@ log_reader_wakeup_triggered(gpointer s)
 {
   LogReader *self = (LogReader *) s;
 
+  if (!(self->super.super.flags & PIF_INITIALIZED))
+    return;
+
   if (!self->io_job.working && self->suspended)
     {
       /* NOTE: by the time working is set to FALSE we're over an
@@ -180,16 +186,14 @@ log_reader_wakeup(LogSource *s)
    *
    */
 
-  if (self->super.super.flags & PIF_INITIALIZED)
-    iv_event_post(&self->schedule_wakeup);
+  iv_event_post(&self->schedule_wakeup);
 }
 
 static void
 log_reader_window_empty(LogSource *s)
 {
   LogReader *self = (LogReader *) s;
-  if (self->super.super.flags & PIF_INITIALIZED)
-    iv_event_post(&self->last_msg_sent_event);
+  iv_event_post(&self->last_msg_sent_event);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


bazsi:
This pull request **introduces 1 alert** when merging 13bc6f6f1ae849929fe909a4098cf4dfe4b29130 into ac66d611cee0be5b5fbbbaec6aa776d4f499820d - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-b24277dacef8a4fb8c33faf55c21f86c24e831b7)

**new alerts:**

* 1 for Ambiguously signed bit\-field member

---

kira-syslogng:
Build SUCCESS

mikeBashStuff:
> Those opening #2457, I would appreciate if you could try whether this fixes your problem.

Will try to take it for a spin this week.

mikeBashStuff:
I have tested a 3.18.1 syslog-ng build with a patch from this PR, however, issue seems to be still present. I tried to reproduce it same way as in https://github.com/balabit/syslog-ng/issues/2457. First time, it was promising as syslog-ng was spinning without any hiccups for over an hour, however, soon after that it run into same condition as described in the ticket. 

Second time was a bit more interesting - I increased number of jobs that were executing logger from 6 to 300 (it resulted in ~1.5MB/s of data being written to a dst file) and here's what the syslog-ng was doing:

[syslog_dump.txt](https://github.com/balabit/syslog-ng/files/2869525/syslog_dump.txt)

The above was generated with:

 `top -b -H -d01 -c -p$(</var/service/syslog-ng/supervise/pid) >syslog_dump.txt`

 Around line 549 most of the threads are gone again, however, the main one doesn't lock on the cpu, messages are still being written, but note what happens with memory - syslog-ng starts to overcommit quite fast where VmRSS goes over 6GB (the system had total of ~100GB so I can just assume that syslog-ng would at some point just force kernel into entering OOM frenzy). At some point syslog-ng locks on the cpu again where we go back to the initially reported condition (+ the mem leak). 

Now, the funny thing is that I am currently running the same test as described in the paragraph above for almost 10 minutes (test seen in .txt took over a 1m before I aborted it) now and everything has been working fine so far, so it's quite hard for me to determine what is that may be triggering this behavior. Maybe it's worth to mention that every minute we call newsyslog via crond to rotate the log files, with limit of 5MB for the file that messages generated by logger ends up in, and when the rotation is done, newsyslog sends USR1 signal to syslog-ng to reopen the files. That's the only process I can think of that may be poking syslog-ng at the same time the test is running. I have also sent HUP and USR1, randomly, couple of times, to see if that may disrupt syslog-ng somehow, but it didn't.

Let me know if you would like me to run some specific tests, grab some additional data, etc. to troubleshoot this further.

bazsi:
Thanks for testing this. I think I have found an issue with my fix, so this doesn't solve the problem. I am going to give it another spin.

bazsi:
The problem is that with my patch iv_event_unregister() is pushed to log_msg_unref, which might be in a non-main-thread, thus the iv_event will be unregistered from the wrong thread.

If you don't use wildcard file sources, this can be a workaround:
```
diff --git a/lib/logreader.c b/lib/logreader.c
index 7777b0890..fabf50f7c 100644
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -193,7 +193,7 @@ static void
 log_reader_window_empty(LogSource *s)
 {
   LogReader *self = (LogReader *) s;
-  iv_event_post(&self->last_msg_sent_event);
+//  iv_event_post(&self->last_msg_sent_event);
 }
 
 static void
```

MrAnno:
@mikeBashStuff We are trying to fix #2457 by removing `last_msg_sent` from `LogReader` and using a different approach that does not have such an ordering issue. This may take some time and effort, but we'll keep you updated in this pull request.

MrAnno:
FYI, #2589 is another attempt at fixing #2457.
This is still not the final fix, but hopefully we are one step closer.

mikeBashStuff:
@MrAnno @bazsi Appreciate your efforts. I will take a look into testing #2589 on my end as well, though it may take some time due to other priorities. But if I you guys come up with some final patch you would like me to test specifically, just let me know and I will try to get it done.

bazsi:
I am closing this pr in favour of the other. Thanks.

PR #2551, opened 2019-02-13
Http auth header

lbudai:


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/Makefile.am
```diff
@@ -22,6 +22,7 @@ include lib/str-repr/Makefile.am
 include lib/timeutils/Makefile.am
 include lib/secret-storage/Makefile.am
 include lib/logthrsource/Makefile.am
+include lib/http-auth/Makefile.am
```
bazsi: Why do we need to add this to lib?

lib/cfg-lexer.c
```diff
@@ -1188,6 +1188,7 @@ static const gchar *lexer_contexts[] =
   [LL_CONTEXT_CLIENT_PROTO] = "client-proto",
   [LL_CONTEXT_SERVER_PROTO] = "server-proto",
   [LL_CONTEXT_OPTIONS] = "options",
+  [LL_CONTEXT_HTTP_AUTH_HEADER] = "http-auth-header",
 };
```
bazsi: This is pluggable? We could perhaps reuse the context_inner_destination context and introduce an api between the plugins.

lib/Makefile.am
lbudai: Without an interface how could one implement a pe-only plugin??

lib/cfg-lexer.c
lbudai: Why this is a big issue?

modules/http/http-worker.c
```diff
@@ -269,7 +269,7 @@ map_http_status_to_worker_status(HTTPDestinationWorker *self, const gchar *url,
                  evt_tag_int("status_code", http_code),
                  evt_tag_str("driver", owner->super.super.super.id),
                  log_pipe_location_tag(&owner->super.super.super.super));
-      retval = LTR_DROP;
+      retval = LTR_NOT_CONNECTED;
```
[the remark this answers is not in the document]
lbudai: indeed, I’d not remove the patch from this branch. this is why I opened pr #2557, just merge that one(after the fixes) and close this pr :)

modules/http/http-worker.c
```diff
@@ -160,6 +160,12 @@ _append_auth_header(struct curl_slist *list, HTTPDestinationDriver *owner)
 {
   const gchar *auth_header_str = http_auth_header_get_as_string(owner->auth_header);
 
+  if (!auth_header_str)
+    {
+      http_dd_auth_header_renew(&owner->super.super.super);
```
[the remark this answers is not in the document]
lbudai: will check it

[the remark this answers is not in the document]
lbudai: yes, it is better to handle that case, thx.
fixed.

lib/cfg-lexer.c
lbudai: @bazsi: yes, it can be done by using the `LL_CONTEXT_INNER_DEST`, but I guess we need the very same to the http source (and yes, I know that there is `LL_CONTEXT_INNER_SRC`).
So what you really want is to remove the `LL_CONTEXT_HTTP_AUTH_HEADER` and 
use the `LL_CONTEXT_INNER_DEST` instead? Is that all? Any other issue?

kira-syslogng:
Build FAILURE

lib/cfg-lexer.c
lbudai: @bazsi: hmmm... after we discussed it with @MrAnno, I think -as it is a requirement to make it possible to implement auth-headers as plugins(including PE-only)- using inner_context is  not a good idea.
On the other hand, I understand your hesitation on 'polluting' the 'context-space', but I think it can be changed/fixed in the future without breaking the config syntax, so I mark your change request as resolved.

modules/http/http-worker.c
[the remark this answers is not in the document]
lbudai: @furiel: based on our discussion, I'm closing #2557 , and remove the patch that maps 40x to `LTR_NOT_CONNECTED`

kira-syslogng:
Build SUCCESS

lib/cfg-lexer.c
MrAnno: Adding the `http-auth` interface to the core lib looks a bit weird, but the current licensing forced us to do such thing AFAIK.

However, I think reusing the inner destination context is not a good idea. For `hook-commands` and `disk-buffer`, it works perfectly, they don't have any "communication interface" with the specific driver. In case of `HttpAuthHeader`, `cfg_parse_plugin()` returns a instance of this interface, so we assume a `HttpAuthHeader` instance here.

If we changed the context to inner destination, it would allow the user to specify a `disk-buffer()` or `hook-commands()` plugin in `auth-header()`. With the current implementation, it results in a crash.

modules/http/http.c
```diff
@@ -299,6 +307,135 @@ _format_stats_instance(LogThreadedDestDriver *s)
...
+
+  persist_state_alloc_string(cfg->state, persist_name, auth_header_str, -1);
+}
+
+static gboolean
+_auth_header_renew(HTTPDestinationDriver *self)
+{
+  gboolean ret = TRUE;
+
+  if (self->auth_header && http_auth_header_has_expired(self->auth_header))
```
gaborznagy: `http_auth_header_has_expired` is optional to implement, without a has_expired() method it will be always FALSE.
This means that we cannot force a renew on error.

We should either:
- make `has_expired` mandatory
- or create a separate function, e.g. `_auth_header_force_renew`.

lbudai: @gaborznagy : indeed, but renewal should be always explicitly tested
Making it as mandatory method is a valid option.

gaborznagy: Thanks, we can open a new PR later with asserting `has_expired` exists. :)

PR #2555, opened 2019-02-15
Fix threaded source/destination crash when reverting configuration

MrAnno:
If a new configuration turns out to be invalid (I mean init()-time validation) during reload, the old one has to be restored.

In such case, the new configuration is in a half-initialized state, so it needs to be deinitialized. This must include a step called `main_loop_worker_sync_call()`, where worker threads are requested and waited to be stopped.

### DIFF
--- a/lib/mainloop-worker.c
+++ b/lib/mainloop-worker.c
@@ -235,7 +235,7 @@ _register_sync_call_action(GQueue *q, void (*func)(gpointer user_data), gpointer
 }
 
 void
-_consume_action(SyncCallAction *action, gpointer dummy)
+_consume_action(SyncCallAction *action)
 {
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
This looks good to me. The only suspicious thing I saw was weather a task is good enough to avoid recursion, but it seems so.

So approving. Nicely structured branch.

kira-syslogng:
Build SUCCESS

MrAnno:
Thanks for the suggestions! :)

PR #2559, opened 2019-02-20
python, java destination add already introduced retry

Kokan:
A new possible value added to python, java (as the C lib already has it): `RETRY`.

### DIFF
--- a/modules/python/sngexample.py
+++ b/modules/python/sngexample.py
@@ -56,7 +56,8 @@ class LogDestination(object):
             (same as boolean False)
         self.DROP: message cannot be sent, it should be dropped immediately.
         self.QUEUED: message is not sent immediately, it will be sent with the flush method.
-        self.NOT_CONNECTED: message is put back to the queue, open method will be called until success."""
+        self.NOT_CONNECTED: message is put back to the queue, open method will be called until success.
+        self.RETRY: message is put back to the queue, try to send again until 3 times, then fallback to self.NOT_CONNECTED."""
 
         pass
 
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build FAILURE

MrAnno:
I think we should not use these prefixes in languages where precise visibility and namespace features are available. However, grepping code is easier this way, that's true.

kira-syslogng:
Build SUCCESS

modules/python/sngexample.py
```diff
@@ -56,7 +56,8 @@ def send(self, msg):
             (same as boolean False)
         self.DROP: message cannot be sent, it should be dropped immediately.
         self.QUEUED: message is not sent immediately, it will be sent with the flush method.
-        self.NOT_CONNECTED: message is put back to the queue, open method will be called until success."""
+        self.NOT_CONNECTED: message is put back to the queue, open method will be called until success.
+        self.RETRY: message is put back to the queue, try to send again until retries_max time, then fallback to self.NOT_CONNECTED."""
```
furiel: I'm afraid retries_max is not configurable at this point. For now we need to explicitely write 3.

Kokan: Would `retiries_max(3)` acceptable ? (as I do not think anybody ever going to update this comment, when introduces to change this limit)

furiel: I don't know if the configurability will be implemented in the future at all. And if so, it will be probably implemented under a different name. Also, writing retries_max() can make users believe this is configurable at this point, which is misleading. So I tend to think it would be better to just write 3 for now.

Kokan: My understanding that this is just an arbitrary number, so it hard to defend, so the next logical conclusion could be to make it configurable. But well I'll go with the flow.

kira-syslogng:
Build SUCCESS

modules/python/sngexample.py
furiel: Thank you!

Issue #2178, opened 2018-07-11
Execute Travis unit tests with Address Sanitizer


Kokan:
There are couple of Sanatizer supported by clang (not sure about gcc), and those Sanatizer could help catch feature bugs, so I think it would be great idea if we configure the sanatizer for at least one job.

This particular issue aim is to use the AddressSanatizer which main purpose to check invalid memory access, but also has leak detection, and it is much faster then `valgrind`.

The following travis branch does run with this sanatizer: https://travis-ci.org/Kokan/syslog-ng/jobs/401312118

The #2173 already fixed a few of those issues.

The remaining tasks:
* [ ] Criterion does not work well with memory leak detection
* [ ] Fix issues from test_stats_cluster
* [ ] Fix issues from test_secret_storage
* [ ] Fix issues from test_diskq_full
* [ ] Fix issues from test_geoip2_parser
* [ ] Fix issues from test_radix (Probably this is going to be a biggest work, but it can be divided into subtasks.)

One thing that I turned off is the `odr` detection: `ASAN_OPTIONS=detect_odr_violation=0`, because the `module_info` violates it, and that is by design does so.

lbudai:
We could run this inside dbld. What do you think?
( @bazsi has a [branch](https://github.com/balabit/syslog-ng/commits/travis-use-docker) where he is working on travis-dbld integration. )

Kokan:
Criterion already has an issue for breaking snatizers: https://github.com/Snaipe/Criterion/issues/220

Kokan:
Small status update:

```
94% tests passed, 10 tests failed out of 156
Total Test time (real) = 106.00 sec
The following tests FAILED:
	 25 - test_stats_cluster (Failed)
	 26 - test_stats_query (Failed)
	 54 - test_logthrsourcedrv (Failed)
	 82 - test_context_info_db (Failed)
	 84 - test_selector_filter (Failed)
	 85 - test_wildcard_source (Failed)
	115 - test_diskq_full (Failed)
	119 - test_geoip2_parser (Failed)
	134 - test_python_logmsg (Failed)
	135 - test_python_template (Failed)
```
/ https://travis-ci.org/Kokan/syslog-ng/jobs/561134891 /

```
	134 - test_python_logmsg (Failed)
	135 - test_python_template (Failed)
```
[...]

PR #2562, opened 2019-02-21
Cmake clang sanitizer

Kokan:
This make it possible to configure with a simple option (`sanitizer`) in case of cmake to use the clang address (or any other sanitizer).

https://clang.llvm.org/docs/AddressSanitizer.html

It is connected to the #2178 issue, which has a connected branch with memory issue fixes. I do rebase the current master from time to time, and since the last time (end of December) I got a lot more test failing, so I though I share a simple way to deploy it for others (and make my life easier not to switch, rebase branch).

This is cmake only, as I won't touch autotools before every unit tests passed.

What you actually need to set this up is: clang, cmake an example can be found on the original branch: https://github.com/Kokan/syslog-ng/blob/address-sanatize/.travis.yml#L157

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -334,6 +334,17 @@ endif()
 
 add_compile_options(${IMPORTANT_WARNINGS} ${ACCEPTABLE_WARNINGS} ${EXTRA_WARNINGS})
 
+# Sanatizer configuration
+set(SANITIZER "OFF" CACHE STRING "Enable clang sanitizer")
+
+set_property(CACHE SANITIZER PROPERTY STRINGS OFF address undefined)
+
+if (SANITIZER)
+  SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O1 -fsanitize=${SANITIZER} -fno-omit-frame-pointer")
+endif()
+
+
+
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

CMakeLists.txt
```diff
@@ -334,6 +334,17 @@ endif()
 
 add_compile_options(${IMPORTANT_WARNINGS} ${ACCEPTABLE_WARNINGS} ${EXTRA_WARNINGS})
 
+# Sanatizer configuration
+set(SANITIZER "OFF" CACHE STRING "Enable clang sanitizer")
+
+set_property(CACHE SANITIZER PROPERTY STRINGS OFF address undefined)
+
+if (NOT ${SANITIZER} STREQUAL "OFF")
```
furiel: OFF, NO, etc evaluate to false. Anything else including `address` and `undefined` evaluates to true. So it is enough to simply write `if (SANITIZER)`.

CMakeLists.txt
```diff
@@ -334,6 +334,17 @@ endif()
...
+  SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O1 -fsanitize=${SANITIZER} -fno-omit-frame-pointer")
```
furiel: Why did you use -O1? Couldn't we more accurate results with `-O0`?

CMakeLists.txt
Kokan: Thanks for the suggestion, fixed.

kira-syslogng:
Build SUCCESS

PR #2564, opened 2019-02-21
Improve nosetests, pylint detection in case of python3

gaborznagy:
With python3 pylint and nose detection doesn't work in case they were installed via pip.
Currently it could work if the packages were installed via a package manager
(tested with apt on Ubuntu Xenial).

- pip installs a `nosetests` and a `nosetests-x.y` file
(both in case of python2 and python3)
- apt package manager installs
  - a `nosetests3` file only in case of python3
  - a `nosetests` and a `nosetests-2.7` in case of python2

**EDIT:**
This patch adds an additional step and searches for the non-versioned form of the binary, e.g. `nosetests`.
Explicit version could be used to avoid using a python2 version of nosetests.
However the version number should be a wildcard instead of the python interpreter's version,
as it could mismatch. 
```
$python3 --version
Python 3.6.7
$pip3 show -f  nose
...
  ../../../bin/nosetests
  ../../../bin/nosetests-3.4
```


kira-syslogng:
Build SUCCESS

gaborznagy:
Sorry, but I just realized that this can go wrong resulting a false-positive dependency detection in case of:
- both python2 and python3 installed
- only python2 nose package is installed via pip

In case of python3 is selected as preferred version, then `nosetests3` program will not be found, but `nosetests` will be, which is a python2 program.

I think the patch needs more work.

furiel:
@gaborznagy I believe the false detection is not a big issue here. As we need to support both python2 and python3, it does not really matter which nosetests etc will be used, we need to comply for both. I see it might be suboptimal to check python3 with python2-nosetests, but the benefit that we can work with pip packages worth it. If this causes problem for someone (I cannot imagine how), then there is still option to set whichever binary they want during compile time. In short, I'm fine with the change. as it is.

gaborznagy:
Thanks @furiel and @Kokan for clarifying this, I agree with you. :)
Even if a more strict pattern would be used, e.g. `nosetests-3.x` there is not guarantee it would work.

I leave the patch as it is.

PR #2565, opened 2019-02-22
improvements in example-msg-generator

nobles:
In this patch I added two changes to `example-msg-generator`. With these changes, `example-msg-generator` can be used for testing purposes easier.

- First message is sent immediately: we do not need to wait `freq`. With this, the tests can speed up.
- With the new `num` option, users can limit how many messages should be generated. Zero means infinite messages (this is the default). Positive integer means the maximum number of messages.

Example:
```
source { example-msg-generator(freq(1) num(5)); };
```

### DIFF
--- a/modules/examples/sources/msg-generator/msg-generator-source.c
+++ b/modules/examples/sources/msg-generator/msg-generator-source.c
@@ -36,6 +36,7 @@ struct MsgGeneratorSource
   LogSource super;
   MsgGeneratorSourceOptions *options;
   struct iv_timer timer;
+  gint num;
 };
 
 static void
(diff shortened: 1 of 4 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

furiel:
@kira-syslogng ok to test

modules/examples/sources/msg-generator/msg-generator-source.c
```diff
@@ -55,14 +56,23 @@ _start_timer(MsgGeneratorSource *self)
...
+
 static gboolean
 _init(LogPipe *s)
 {
   MsgGeneratorSource *self = (MsgGeneratorSource *) s;
   if (!log_source_init(s))
     return FALSE;
-
-  _start_timer(self);
+  if (self->options->max_num)
```
Kokan: Please check the same way as below (` ..max_num < 0`).

kira-syslogng:
Build SUCCESS

modules/examples/sources/msg-generator/msg-generator-source.c
nobles: Thanks. Fixed.

kira-syslogng:
Build SUCCESS

PR #2571, opened 2019-02-27
light: example-msg-generator support

furiel:
This patchset adds initial support for `example-msg-generator` in the light test framework.

For now, only template can be specified. In a followup patch, I plan to support specific message list that is emitted by the generator.

I needed to do a small refactor on the code: prior this patch positional parameters were mandatory in drivers, that I needed to eliminate. 

The patchset contains a minor fix in the `SingleLineParser` code in the test framework.

### DIFF
--- a/tests/python_functional/functional_tests/source_drivers/generator_source/test_generator_source.py
+++ b/tests/python_functional/functional_tests/source_drivers/generator_source/test_generator_source.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+#############################################################################
+# Copyright (c) 2019 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+
+def test_generator_source(tc):
+    config = tc.new_config()
+
+    generator_source = config.create_example_msg_generator(num=1)
+    file_destination = config.create_file_destination(file_name="output.log", template="'$MSG\n'")
+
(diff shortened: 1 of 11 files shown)


kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
```diff
@@ -108,3 +109,10 @@ def create_logpath(self, statements=None, flags=None):
     def create_inner_logpath(self, statements=None, flags=None):
         inner_logpath = self.__create_logpath_with_conversion(statements, flags)
         return inner_logpath
+
+    def create_generator_source(self, freq=1, num=0, template=None):
```
mitzkia: 1, About the config-create function names (in general), my preference is: "create+current_driver_name()" -> `create_example_msg_generator()` would be better here. Be consequent with our driver names. 
But...
2, I like the approach to create drivers with default options (I am also thinking about this in one of my next PRs),
but I would not dismiss the possibility to create drivers with any options. So I would do the following changes here:
a, rename the current function to: `create_basic_example_msg_generator()` -> I would define the basic default options from inside the function (as you did it from parameters).
b, create a new function like: `create_example_msg_generator(**kwargs)` -> Further options can be passed via **kwargs

mitzkia:
@furiel Thank you for the new driver support, for the simplifications and for the bug fix.

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
furiel: Sure thing. I will do so.

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/source_drivers/generator_source/test_generator_source.py
```diff
@@ -0,0 +1,35 @@
...
+    config = tc.new_config()
+
+    generator_source = config.create_example_msg_generator(num=1)
+    file_destination = config.create_file_destination(file_name="output.log", template="'$MSG\n'")
+
+    config.create_logpath(statements=[generator_source, file_destination])
+    syslog_ng = tc.new_syslog_ng()
+    syslog_ng.start(config)
+    log = file_destination.read_log()
+    assert log == generator_source.DEFAULT_MESSAGE + "\n"
```
Kokan: Question: as I did a little dig in, but cannot answer myself; does this also check if the file destination has only 1 message ?

furiel: No, it only checks only the top message.

For an acceptance test, it would have been a good idea to check all messages. I have followup work on the example msg generator so I'll fix this one later. Thanks for the comment.

PR #2546, opened 2019-02-10
light: Switch to native logger

mitzkia:
For now internal classes: Logger and LoggerFactory are strong dependencies inside the lib. They are passed for those classes where we want to print some log messages, but they are not used everywhere.
Pytest also has a good support for using native logging mechanism. We can easily add support for this.
Until now there was one missing feature from pytest side namely: it could not create report file (log-file) dynamically (from hook). This has been changed, there is an upcoming PR: https://github.com/pytest-dev/pytest/pull/4752

Update (2019-02-22):
* New pytest releated PR: https://github.com/pytest-dev/pytest/pull/4812

Missing steps:
- [x] save AssertionErrors also into report files (as earlier)
- [x] check created logger objects in classes/modules do they really needed
- [x] remove --loglevel starter option

Decision needed:
- about logging format (example): `%(asctime)s:%(msecs)d - %(filename)s:%(lineno)d->%(funcName)s - %(levelname)s - %(message)s`

WIP state:
- This PR should be in WIP state until the pytest's PR is not merged and released
- TravisCI is failing because of this

How logging will work:
1, setup initial logging parameters via command line or pytest.ini
2, Add `import logging` into module where you want to print some log
3, create new logger for that class or module
4, use created logger with a proper debug level

PR #2587, opened 2019-03-01
light: Remove SetupTestcase() dependency 

mitzkia:
For now SetupTestcase() is the main dependency for every test method. It works as a Facade pattern.
We can eliminate this dependency. Instead of it we can use native pytest fixtures.
We can move our setup logic into `pytest_runtest_setup()`

Notes: 
- this patch-set is based on my previous PR: #2546 
- this PR is not yet review ready, I have still some todos


kira-syslogng:
Build SUCCESS

MrAnno:
Let's just use the `light-conftest: Save failure descriptions` or `light: conftest: Save failure descriptions` format or something like @furiel chose in #2571 for commit messages.

@mitzkia What do you think? :)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

mitzkia:
For now I have rebased to master (also still depends on #2546). 
The code seems finished, but needs to split at least 3 separate commits.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

tests/python_functional/src/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@pytest.fixture
+def fake_testcase_parameters(request, tmpdir):
+    orig_installdir = request.config.option.installdir
+    request.config.option.installdir = tmpdir.join("installdir")
+    request.config.option.reports = tmpdir.join("reports")
+    request.config.option.runwithvalgrind = False
+    yield TestcaseParameters(request)
+    request.config.option.installdir = orig_installdir
```
furiel: Teardown fixture! Nice :)
(Actually this is a bugfix, it would worth its own commit).

tests/python_functional/src/executors/tests/test_process_executor.py
```diff
@@ -24,11 +24,11 @@
 from src.executors.process_executor import ProcessExecutor
 
 
-def test_start_stop_process(tc_unittest):
+def test_start_stop_process(tmpdir):
     process_executor = ProcessExecutor()
-    process_command = ["python", "-c", "import time; time.sleep(3)"]
-    stdout_file = tc_unittest.get_temp_file()
-    stderr_file = tc_unittest.get_temp_file()
+    process_command = ["python", "-c", "import time; time.sleep(1)"]
```
furiel: A simple `"ls"` command is enough.

tests/python_functional/src/executors/tests/test_process_executor.py
```diff
@@ -24,11 +24,11 @@
...
+    stdout_file = tmpdir.join("stdout.log")
+    stderr_file = tmpdir.join("stderr.log")
```
furiel: Is it necessary to set these? Stdout and stderr is not asserted.

tests/python_functional/functional_tests/logpath/test_flags_fallback.py
```diff
@@ -54,20 +48,23 @@ def test_flags_fallback(tc):
 
     config.create_logpath(statements=[file_source, inner_logpath1, inner_logpath2])
 
-    bsd_message1 = write_message_with_fields(tc, file_source, hostname="host-A", message="message from host-A")
-    bsd_message2 = write_message_with_fields(tc, file_source, hostname="host-B", message="message from host-B")
+    log_message1 = LogMessage().hostname("host-A")
```
furiel: Abstracting out this chunk into `write_message_wit_fields` seemed nice to me. Is there a reason that this cannot work anymore?

Same question for the other testcases.

tests/python_functional/src/executors/tests/test_process_executor.py
mitzkia: +1

tests/python_functional/src/executors/tests/test_process_executor.py
mitzkia: We have not assert it, but only passed to `process_executor.start()`  in the next line.

tests/python_functional/functional_tests/logpath/test_flags_fallback.py
mitzkia: Thanks for the note. I will come up with sg here.

tests/python_functional/src/executors/tests/test_process_executor.py
furiel: It would be great if we could pass `None` instead of them.

mitzkia: +1, I will fix this to None

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

tests/python_functional/src/executors/tests/test_process_executor.py
mitzkia: For now I would remain it as it is (using files in stdout/stderr). I have tried both dev/null and None, and with this file has been created in python_functional/ dir

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

mitzkia:
Sorry about that, but I needed to split this PR into smaller commits. After re-reading my previous commits it was not easy to deal with it, and also it was not bisectable.
For now it should be bisectable.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

tests/python_functional/conftest.py
```diff
@@ -63,18 +80,27 @@ def pytest_collection_modifyitems(config, items):
...
+        file_name = "testcase_{}.log".format(testcase_name)
+        return Path(relative_report_dir, testcase_name, file_name).absolute()
 
-def get_current_date():
-    return datetime.now().strftime("%Y-%m-%d-%H-%M-%S-%f")
+    config = item.config
+    logging_plugin = config.pluginmanager.get_plugin("logging-plugin")
+    report_file_path = construct_report_file_path(item)
+    logging_plugin.set_log_path(report_file_path)
+    yield
```
furiel: Is this `yield` needed?

tests/python_functional/functional_tests/conftest.py
```diff
@@ -0,0 +1,58 @@
...
+        return str(Path(relative_report_dir, testcase_name, file_name).absolute())
+
+    config = item.config
+    logging_plugin = config.pluginmanager.get_plugin("logging-plugin")
+    report_file_path = construct_report_file_path(item)
+    logging_plugin.set_log_path(report_file_path)
+    item.user_properties.append(('report_file_path', report_file_path))
+    prepare_testcase_working_dir(item._request)
+    item._request.addfinalizer(lambda: logger.info("Report file path\n{}\n".format(report_file_path)))
+    yield
```
furiel: Is this `yield` necessary?

tests/python_functional/conftest.py
mitzkia: Yes, it is needed to returning from the pytest hook (this is a pytest internal thing)

tests/python_functional/functional_tests/conftest.py
mitzkia: Yes, it is needed to returning from the pytest hook (this is a pytest internal thing)

mitzkia: Ok, I have checked, and for now we did not need it. (related docs: https://pluggy.readthedocs.io/en/latest/)

mitzkia:
I have fixed the found review note and force pushed the PR.

kira-syslogng:
Build SUCCESS

tests/python_functional/src/common/operations.py
```diff
@@ -50,3 +50,9 @@ def delete_session_file(shared_file_name, syslog_ng_testcase):
     working_dir = syslog_ng_testcase.testcase_parameters.get_working_dir()
     shared_file_name = Path(working_dir, shared_file_name)
     shared_file_name.unlink()
+
+
+def calculate_testcase_name(pytest_request):
+    # In case of parametrized tests we need to replace "[" and "]" because
+    # testcase name will appear in directory name
+    return pytest_request.node.name.replace("[", "_").replace("]", "_")
```
Kokan: The commit is a little misleading as you remove the `node.originalname` as well. Can we left that out ?

mitzkia: Sorry, I forgot to mention it in the commit. So yes, we can leave originalname. It is the same as `node.name` but without any decorations for example: suffixes which came from parametrization, it is only the pure testcase function name. 
And we did not need this, because we always need the parametrization suffix to create separate directories also for parametrized tests. With using `node.name` we will get uniq testcase names also in case of parametrized tests.

PR #2590, opened 2019-03-03
Include json-c in the dist tarball

### LINKED ISSUE #1204: json-c submodule is missing from the tarball

MrAnno:
Note:
`json-c` has a dist hook, where `doxygen` is used unconditionally:

https://github.com/json-c/json-c/blob/3e81b4abe359c8128bb2b4127f4e8c8c057fb004/Makefile.am#L8-L11

This means `doxygen` has to be installed in order to generate the tarball.

Questions:
- I don't think it is acceptable to depend on `doxygen`. Shouldn't we try to contribute and make it conditional in `json-c`?

Fixes #1204

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,7 @@ addons:
       - libmaxminddb-dev
       - libxml2-utils
       - astyle
+      - doxygen
 install:
       - pip install --user -r requirements.txt
       - pip list
--- a/configure.ac
+++ b/configure.ac
@@ -1147,6 +1147,13 @@ dnl ***************************************************************************
 
 enable_json="no"
 
+JSON_INTERNAL_SOURCE_EXISTS="test -f $srcdir/lib/jsonc/json.h"
+
+if $JSON_INTERNAL_SOURCE_EXISTS; then
+  AC_CONFIG_SUBDIRS([lib/jsonc])
+  JSON_SUBDIRS="lib/jsonc"
+fi
+
 if test "x$with_jsonc" = "xsystem" -o "x$with_jsonc" = "xauto"; then
    enable_json="yes"
    PKG_CHECK_EXISTS(json-c, json_module_name="json-c",
@@ -1158,17 +1165,16 @@ if test "x$with_jsonc" = "xsystem" -o "x$with_jsonc" = "xauto"; then
 fi
 if test "x$with_jsonc" = "xinternal" -o "x$with_jsonc" = "xauto"; then
   AC_MSG_CHECKING(for JSON)
-  if test -f "$srcdir/lib/jsonc/json.h"; then
-      AC_CONFIG_SUBDIRS([lib/jsonc])
+  if $JSON_INTERNAL_SOURCE_EXISTS; then
       JSON_LIBS="\$(top_builddir)/lib/jsonc/libjson-c.la"
       JSON_DEPENDENCY="$JSON_LIBS"
       JSON_CFLAGS="-I\$(top_srcdir)/lib/jsonc -I\$(top_builddir)/lib/jsonc"
-      JSON_SUBDIRS="lib/jsonc"
       enable_json="yes"
       with_jsonc="internal"
   else
       AC_MSG_WARN([Internal jsonc sources not found in lib/jsonc])
       enable_json="no"
+      with_jsonc="no"
   fi
   AC_MSG_RESULT($enable_json)
 fi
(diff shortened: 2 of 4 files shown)


kira-syslogng:
Build SUCCESS

Kokan:
`I don't think it is acceptable to depend on doxygen. Shouldn't we try to contribute and make it conditional in json-c?`
I think it worth a trial.

But I would rather ask the question: why do we ship `json-c` ?

bazsi:
I got this request recently too, from an organization that maintains its
own repository of software, compiling internally and where adding
dependencies recursively poses a problem.

I was under the impression that we already include jsonc in the tarball.

Depending on doxygen is not very nice, however we only need this in the
dbld docker image where we generate tarballs, compiling syslog-ng will not
need this, right?



On Mar 3, 2019 22:39, "Kókai Péter" <notifications@github.com> wrote:

I don't think it is acceptable to depend on doxygen. Shouldn't we try to
contribute and make it conditional in json-c?
I think it worth a trial.

But I would rather ask the question: why do we ship json-c ?

—
You are receiving this because you are subscribed to this thread.
<https://github.com/balabit/syslog-ng/pull/2590#issuecomment-469066806>,
or mute
the thread

MrAnno:
Yes, compilation should not be affected by this hook.

lbudai:
Other solution: remove json-c as submodule...

MrAnno:
> I got this request recently too, from an organization that maintains its own repository of software, compiling internally and where adding dependencies recursively poses a problem.

Just for the record: The `skip_submodules` environment variable disables submodule initialization in `autogen.sh`. We also use this internally, when recursive dependencies are not allowed (and/or when there is no internet on the builder). The tarball generation won't fail in that case, submodules will just be left out.

bazsi:
He was talking about _binary_ dependencies, not source ones. He is not
permitted to pull in newer versions or source packages that he built
himself.

lbudai:
If we are talking about (Debian) packaging, then I think the problem is come from the release tarball generation (as I know, they generate their own tarball which fails due to json-c).
As they won't use the internally shipped json-c (and any other internally shipped 3rd party dependencies), then when they are generating the tarball  they should have to use the `skip_submodules` option as @MrAnno  [mentioned](https://github.com/balabit/syslog-ng/pull/2590#issuecomment-469251988).

Or, other solutions:
1) remove json-c submodule... 
2) fix json-c

But I think the 'skip-submodules' is the real solution for package maintainers.

czanik:
And it broke ```dbld/rule tarball```.

@mitzkia: do I guess well, that I have to rebuild the base images locally until those on Docker hub are rebuilt?

```
test -d "../../syslog-ng-3.20.1/lib/jsonc" || mkdir "../../syslog-ng-3.20.1/lib/jsonc"
 (cd tests && make  top_distdir=../../../syslog-ng-3.20.1 distdir=../../../syslog-ng-3.20.1/lib/jsonc/tests \
make  \
  top_distdir="../../syslog-ng-3.20.1" distdir="../../syslog-ng-3.20.1/lib/jsonc" \
  dist-hook
test -d "../../syslog-ng-3.20.1/lib/jsonc/doc" || mkdir "../../syslog-ng-3.20.1/lib/jsonc/doc"
chmod -R u+w "../../syslog-ng-3.20.1/lib/jsonc/doc"
cd "../../syslog-ng-3.20.1/lib/jsonc" && doxygen
/bin/bash: doxygen: command not found
Makefile:1100: recipe for target 'dist-hook' failed
make[3]: *** [dist-hook] Error 127
Makefile:766: recipe for target 'distdir' failed
make[2]: *** [distdir] Error 2
Makefile:21461: recipe for target 'distdir' failed
make[1]: *** [distdir] Error 1
Makefile:21560: recipe for target 'dist' failed
make: *** [dist] Error 2
```

bazsi:
could you please update your docker image?

dbld/rules image-<yourplatform> or wait until dockerhub does and explicitly
pull a new image.

czanik:
The ```tarball``` target needs Ubuntu Bionic, so I did a ```docker rmi  balabit/syslog-ng-ubuntu-bionic``` and ```dbld/rules image-ubuntu-bionic``` and my problem was resolved.

mochrul:
To be clear, the problem is not with Debian building. If this had cause problem in Debian, that would have been reported many years ago. But because in Debian we actually do/could not use the embedded libs, this isn't a problem there.

But I can imagine other firms, which would want to use syslog-ng, would want to compile itself (for example because they want to use in an OS no binary package is available, or using a pre-built binary package is not allowed for security reasons), but it's hard for them to maintain all the dependencies/not allowed to freely download dependencies. And in this case, the fact that all these dependencies are bundled into one tgz which also includes the syslog-ng itself can make the life of that firm unbelievably easier.

PR #2601, opened 2019-03-07
value-pairs: move ValuePairs struct declaration to a private header

bazsi:
To make the ValuePairs struct opaque to users, move its definition to
a private header, so that cmdline.c can have access, but still the
rest of the codebase can't.


kira-syslogng:
Build SUCCESS

PR #2603, opened 2019-03-07
test-stats-query: fix unit test

MrAnno:
This test bug was caught by our macOS Travis job.

The ordering of `stats_query_get()` is undefined, its implementation is based on a hash table.

In case of multiple matching counters, the test case had to be fixed by checking the existence of each counter in an unordered manner.


kira-syslogng:
Build SUCCESS

PR #2619, opened 2019-03-10
test_systemd_journal: rewrite test cases in criterion

### LINKED ISSUE #2611: convert test_systemd_journal to criterion
Use Criterion to rewrite this unit test.

mehul-m-prajapati:
fixes #2611

### DIFF
--- a/modules/systemd-journal/tests/test_systemd_journal.c
+++ b/modules/systemd-journal/tests/test_systemd_journal.c
@@ -25,15 +25,32 @@
 #include "test-source.h"
 #include "journald-helper.c"
 #include "journal-reader.c"
-#include "testutils.h"
 #include "apphook.h"
-
-#define JOURNALD_TESTCASE(testfunc, ...) { testcase_begin("%s(%s)", #testfunc, #__VA_ARGS__); testfunc(__VA_ARGS__); testcase_end(); }
-
-#define TEST_PERSIST_FILE_NAME "test_systemd_journal.persist"
+#include <criterion/criterion.h>
 
 static gboolean task_called;
 static gboolean poll_triggered;
+
+static GlobalConfig *cfg;
+
+static void
+_init_cfg_with_persist_file(const gchar *persist_file)
+{
+  main_thread_handle = get_thread_id();
+  cfg = cfg_new_snippet();
+  cfg->threaded = FALSE;
+  cfg->state = persist_state_new(persist_file);
+  cfg->keep_hostname = TRUE;
+  persist_state_start(cfg->state);
+}
+
+static void
+_deinit_cfg(const gchar *persist_file)
+{
+  persist_state_cancel(cfg->state);
+  unlink(persist_file);
+  cfg_free(cfg);
+}
 
 void
 add_mock_entries(gpointer user_data)
@@ -57,7 +74,7 @@ handle_new_entry(gpointer user_data)
 {
   Journald *journald = user_data;
   journald_process(journald);
-  assert_false(poll_triggered, ASSERTION_ERROR("Should called only once"));
+  cr_assert_not(poll_triggered, "%s", "Should called only once");
   poll_triggered = TRUE;
 }
 
@@ -71,9 +88,9 @@ void
 __test_seeks(Journald *journald)
 {
   gint result = journald_seek_head(journald);
-  assert_gint(result, 0, ASSERTION_ERROR("Can't seek in empty journald mock"));
+  cr_assert_eq(result, 0, "%s", "Can't seek in empty journald mock");
   result = journald_next(journald);
-  assert_gint(result, 0, ASSERTION_ERROR("Bad next step result"));
+  cr_assert_eq(result, 0, "%s", "Bad next step result");
 
   MockEntry *entry = mock_entry_new("test_data1");
   mock_entry_add_data(entry, "MESSAGE=test message");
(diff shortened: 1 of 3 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

mehul-m-prajapati:
Hi @MrAnno ,

Thanks for noting down the error log.

lbudai:
@mehul-m-prajapati : when you port a test to criterion, keep in mind, that criterion runs the test cases in a parallel way. So be very careful with resource management.

( Criterion has an option where you can set the number of concurrent jobs(`-j`) - I usually run tests also with `-j1` option and also without limiting the concurrent jobs. )

lbudai:
@mehul-m-prajapati : just take a look at [test_hostid.c](https://github.com/balabit/syslog-ng/blob/master/tests/unit/test_hostid.c)

mehul-m-prajapati:
@lbudai, @MrAnno : I have changed the code as per your suggestion.

kira-syslogng:
Build SUCCESS

modules/systemd-journal/tests/test_systemd_journal.c
```diff
@@ -25,16 +25,30 @@
...
+#include <criterion/criterion.h>
 
 static gboolean task_called;
 static gboolean poll_triggered;
 
+static void
+_init_cfg_with_persist_file(const gchar *persist_file)
+{
+  main_thread_handle = get_thread_id();
+  configuration = cfg_new_snippet();
```
lbudai: I think it would be better to create a `GlobalConfig` object instead of using the global `configuration`.

kira-syslogng:
Build SUCCESS

modules/systemd-journal/tests/test_systemd_journal.c
mehul-m-prajapati: ok.

MrAnno: Oh sorry, I did not see that this one has not been resolved yet. I'm marking this unresolved.

mehul-m-prajapati: If I use GlobalConfig object, one of the test case on my machine fails.
```
[FAIL] systemd_journal::test_journal_reader: CRASH!
```

MrAnno: Yes, you have to pass your config object in that test:
```
test_source_new(configuration) // <-- here,
```

One way of doing that would be, if `_init_cfg_with_persist_file()` returned the created `GlobalConfig` object, so you can pass it to the constructor of the test source.

mehul-m-prajapati: Yes, Thanks @MrAnno

Kokan: @mehul-m-prajapati why does it crashes ? maybe there is some issue the test found this way.

mehul-m-prajapati: @Kokan : That might be the case. I am still debugging the issue.

Kokan: Maybe the global `configuration` is used somewhere in the test, that is not initialised causing null pointer dereference.

mehul-m-prajapati: @Kokan : You are correct. I can see below code in test-source.c
```
static gboolean
__init(LogPipe *s)
{
  TestSource *self = (TestSource *)s;
  self->reader = journal_reader_new(configuration, self->journald_mock);
  journal_reader_options_defaults(&self->options);
  if (self->current_test_case && self->current_test_case->init)
    {
      self->current_test_case->init(self->current_test_case, self, self->journald_mock, self->reader, &self->options);
    }
  journal_reader_options_init(&self->options, configuration, "test");
```

Now, how can i pass local GlobalConfig object in a __init callback of test-source.c ?

Kokan: The `LogPipe`'s init function signature only accepts `LogPipe`, so passing anything can be done via `self`. Luckily your use case have been foreseen by others and `log_pipe_get_config` might help you out :)

mehul-m-prajapati: @Kokan : Thanks for the suggestion. I have modified the patch. Please review the same.

modules/systemd-journal/tests/test_systemd_journal.c
```diff
@@ -487,21 +511,24 @@ _test_program_field_test(TestCase *self, TestSource *src, LogMessage *msg)
...
 }
 
-void
-test_journal_reader(void)
+Test(systemd_journal, test_journal_reader)
 {
-  TestSource *src = test_source_new(configuration);
+  const gchar *persist_file = "test_systemd_journal1.persist";
+
+  _init_cfg_with_persist_file(persist_file);
```
Kokan: As far as I can see all of your tests starts with this functions, it is just preparing common things for the tests. I would use this and the deinit for `init` and `fini`.

mehul-m-prajapati: @Kokan : I want a different file name for each test case. So, I guess putting this function in init and fini will give me same file name everytime.

Kokan: Yes, still some part could be moved there; or perhaps we could have some different api creating persist files with generated names for UT.

mehul-m-prajapati: @Kokan : So, do I need to change the code with your suggestion?

Kokan: I am fine with it like this (see approve from my side), but sure would prefer some more cleanup.

lbudai:
I have only one issue with this: this works just because the criterion _forks_ the tests... so the global variable will be global for all test cases. We can depend on this fact, but I think it's not elegant.
But don't want to block this PR, so I'm merging this.

PR #2621, opened 2019-03-11
test_diskq: rewrite test cases in criterion

### LINKED ISSUE #2612: convert test_diskq to criterion
Use Criterion to rewrite this unit test.

- [ ] test_reliable_backlog
- [ ] test_diskq_truncate
- [ ] test_diskq
- [ ] test_diskq_full

mehul-m-prajapati:
fixes #2612

### DIFF
--- a/modules/diskq/tests/test_diskq_truncate.c
+++ b/modules/diskq/tests/test_diskq_truncate.c
@@ -30,7 +30,7 @@
 
 #include "queue_utils_lib.h"
 #include "test_diskq_tools.h"
-#include "testutils.h"
+#include <criterion/criterion.h>
 
 #include <stdlib.h>
 #include <sys/stat.h>
@@ -80,9 +80,10 @@ _assert_diskq_actual_file_size_with_stored(LogQueue *q)
   gchar *filename = qdisk->filename;
 
   struct stat diskq_file_stat;
-  assert_gint(stat(filename, &diskq_file_stat), 0, "Stat call failed, errno:%d", errno);
-  assert_gint(qdisk->file_size, diskq_file_stat.st_size, "File size does not match with stored size");
-  fprintf(stderr, "Actual file size: %ld\n", diskq_file_stat.st_size);
+  cr_assert_eq(stat(filename, &diskq_file_stat), 0, "Stat call failed, errno:%d", errno);
+  cr_assert_eq(qdisk->file_size, diskq_file_stat.st_size,
+               "File size does not match with stored size; Actual file size: %ld, Expected file size: %ld\n", diskq_file_stat.st_size,
+               qdisk->file_size);
 }
 
 static void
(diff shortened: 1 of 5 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -177,22 +177,21 @@ test_diskq_truncate_without_diskbuffer_used(void)
...
   msg_init(TRUE); // internal messages will go to stderr(), msg_init() is idempotent
   app_startup();
 
   configuration = cfg_new_snippet();
-  assert_true(cfg_init(configuration), "cfg_init failed!");
-
-  // Diskbuffer is the part of disk-queue that is used only when qout is full
-  test_diskq_truncate_with_diskbuffer_used();
-  test_diskq_truncate_without_diskbuffer_used();
+  cr_assert(cfg_init(configuration), "cfg_init failed!");
```
Kokan: The `setup` and `teardown` phases are not designed to assert in them. `assert` and `expect` should be kept in the `Test` phase.

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -177,22 +177,21 @@ test_diskq_truncate_without_diskbuffer_used(void)
...
+static void
+teardown(void)
+{
   app_shutdown();
   cfg_free(configuration);
-
-  return 0;
 }
+
+TestSuite(diskq_truncate, .init = setup, .fini = teardown);
```
Kokan: This test case fails in travis please fix it.

mehul-m-prajapati: ok..

modules/diskq/tests/test_diskq_truncate.c
mehul-m-prajapati: okay. I will remove assert from setup and teardown.

Kokan:
One of the statistics ut crashes for me, could you please check ? Also please solve the merge conflict.

modules/diskq/tests/test_diskq_truncate.c
gaborznagy: What about using other error handling, but keeping cfg_init in setup function?
Maybe C standard `assert` works in setup?
I don't think it's a good way to have init functions per test cases.

mehul-m-prajapati: @gaborznagy : ok, I will use standard assert.

Kokan:
Travis fails because recently we moved to 3.1 version of astyle and it has a few format fixes. (You can run the *correct* version with `./dbld/rules run RUN_COMMAND="cd /source && astyle --version && scripts/style-checker.sh format"`

mehul-m-prajapati:
@Kokan : Thanks

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

Kokan:
@mehul-m-prajapati on my machine the `test_diskq` test crashes (specificly the tests_diskq_statistics with non reliable disk buffer).

This is just a feedback that despite the test is passed in Travis I would not like to merge until the issue is well understood and/or solved.

Some details I could dig up till now:
The crash happens when the `testcase_diskq_prepare` tries to call the `q = parameters->constructor(options, NULL);` function, if the function pointer should be `log_queue_disk_non_reliable_ne
w`.

This crash does not happen if the kernel ASLR is disabled. (Note: Criterion is known not to work well with ASLR, which has been addressed with: https://github.com/balabit/syslog-ng/pull/1753
or we thought so)

At the moment I think the key is on the *pointer should be `log_queue_disk_non_reliable_new`*.
If you use debugger to print the addresses of the parameters:

```
$3 = {{
...
    constructor = 0x7f28ac1ed0fe <log_queue_disk_non_reliable_new>,
...
  }, {
...
    constructor = 0x402290 <log_queue_disk_reliable_new@plt>,
...
  }, {
```

It seems that the `log_queue_disk_reliable_new@plt` uses PLT to reach the original function, and that works. While the `log_queue_disk_non_reliable_new` tries to call directly, which fails. C
rash because of ASLR make sense here, see the issue linked above.

The question remains why the different method to reach the function is used, btw. I could somehow managed to get PLT usage if I created a proxy function in the test, and the test passed.

```
gcc (GCC) 8.2.1 20181127
GNU ld (GNU Binutils) 2.31.1
Linux foo 5.0.5-arch1-1-ARCH #1 SMP PREEMPT Wed Mar 27 17:53:10 UTC 2019 x86_64 GNU/Linux
```

Kokan:
I've tried this PR with older version of the linker, and compiler:
```
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609
GNU ld (GNU Binutils for Ubuntu) 2.26.1
```

The issue is not present with those versions, thou I would like to find out more details, we discussed it would not be fare to block this PR, as it seems the code is fine.

PR #2623, opened 2019-03-11
Force C99 with GNU

Kokan:
Enables a switch to force GNU99 as the current used C standard is C99.
In case of CMake the GNU99 was already enabled, but warnings were excluded with `-isystem` style include.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -180,6 +180,9 @@ AC_ARG_ENABLE(forced_server_mode,
 AC_ARG_ENABLE(debug,
               [  --enable-debug      Enable debugging code.],, enable_debug="no")
 
+AC_ARG_ENABLE(force_gnu99,
+              [  --enable-force-gnu99      Enforce C99 with gnu extensions.],, force_gnu99="no")
+
 AC_ARG_ENABLE(extra-warnings,
               [  --enable-extra-warnings      Enable extra compiler warnings (recommended).],, enable_extra_warnings="no")
 
@@ -480,7 +483,6 @@ dnl ***************************************************************************
 dnl Set up CFLAGS
 
 
-
 dnl The rest of CFLAGS. We produce
 dnl -g/-O2/-pg/-fprofile-arcs/-ftest-coverage combinations, the rest is
 dnl supplied by Makefile.am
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

configure.ac
```diff
@@ -507,6 +509,12 @@ else
 fi
 CFLAGS_ADD="${CFLAGS_ADD} -pthread"
 
+dnl The current supported standard is C99 with gnu extensions
+
+if test "x$force_gnu99" = "xyes"; then
+	CFLAGS_ADD="${CFLAGS_ADD} --std=gnu99"
```
MrAnno: This should be `-std` instead of `--std`

Kokan: thanks, done.

MrAnno: Well, it does not really matter in case of gcc:
https://github.com/gcc-mirror/gcc/blob/74157df9c26d1f053678d76e60d065a67815660f/gcc/opts-common.c#L456-L457

But only `-std` is documented. I don't know about clang, it should work as well.

MrAnno: Thanks!

kira-syslogng:
Build SUCCESS

configure.ac
```diff
@@ -180,6 +180,9 @@ AC_ARG_ENABLE(forced_server_mode,
 AC_ARG_ENABLE(debug,
               [  --enable-debug      Enable debugging code.],, enable_debug="no")
 
+AC_ARG_ENABLE(force_gnu99,
+              [  --enable-force-gnu99      Enforce C99 with gnu extensions.],, force_gnu99="no")
```
furiel: Can we turn this on by default?

Kokan: I wanted to do it as opt-in as c99 is something anybody should really force. As you may miss some compiler optimisation in this way - even if your codebase follows c99 (I do not say it is a fact, just a feeling). This is not really a *good* enough reason.

Why would you run this on by default ?

furiel: I just wanted to avoid the possibility to write some code locally, that would not compile in CI: just for catching the errors easier. But I can make an "alias" for this option in my development environment, so not really important request. Marking as resolved.

PR #2626, opened 2019-03-13
afamqp: add support for external authentication

spencercw:
AMQP has support for an 'external' authentication scheme which uses some out-of-band method for authenticating the user. For instance, the user may be identified from the Distinguished Name or Common Name in their client certificate. The method by which the user is identified when using external authentication depends on the server configuration.

Add a new 'auth-method' parameter to the amqp destination configuration which may be set to 'plain' or 'external'. When this parameter is omitted or set to 'plain', the existing behaviour is unchanged. When set to 'external' then external authentication is enabled and the username, if supplied, will be used as the identity, although this is generally not needed and may be omitted.

### DIFF
--- a/modules/afamqp/afamqp.c
+++ b/modules/afamqp/afamqp.c
@@ -52,6 +52,7 @@ typedef struct
   gchar *host;
   gint port;
 
+  gint auth_method;
   gchar *user;
   gchar *password;
 
@@ -77,6 +78,21 @@ typedef struct
 /*
  * Configuration
  */
+
+gboolean
+afamqp_dd_set_auth_method(LogDriver *d, const gchar *auth_method)
+{
+  AMQPDestDriver *self = (AMQPDestDriver *) d;
+
+  if (strcasecmp(auth_method, "plain") == 0)
+    self->auth_method = AMQP_SASL_METHOD_PLAIN;
+  else if (strcasecmp(auth_method, "external") == 0)
+    self->auth_method = AMQP_SASL_METHOD_EXTERNAL;
+  else
+    return FALSE;
+
+  return TRUE;
+}
 
 void
 afamqp_dd_set_user(LogDriver *d, const gchar *user)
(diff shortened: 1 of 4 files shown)


kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

modules/afamqp/afamqp.c
```diff
@@ -463,8 +488,28 @@ afamqp_dd_connect(AMQPDestDriver *self, gboolean reconnect)
...
+    case AMQP_SASL_METHOD_EXTERNAL:
+      // The identity is generally not needed for external authentication, but must be set to
+      // something non-empty otherwise the API call will fail, so just default it to something
+      // meaningless
+      identity = (self->user && self->user[0] != '\0') ? self->user : ".";
+      ret = amqp_login(self->conn, self->vhost, self->max_channel, self->frame_size, 0,
+                       self->auth_method, identity);
+      break;
+
+    default:
```
bazsi: Since we validate when the grammar is parsed, this can easily become an assertion. No need to handle errors as they will essentially be dead code. Just use g_assert_not_reached ()

modules/afamqp/afamqp.c
```diff
@@ -463,8 +488,28 @@ afamqp_dd_connect(AMQPDestDriver *self, gboolean reconnect)
       goto exception_amqp_dd_connect_failed_init;
     }
 
-  ret = amqp_login(self->conn, self->vhost, self->max_channel, self->frame_size, 0,
-                   AMQP_SASL_METHOD_PLAIN, self->user, self->password);
+  switch (self->auth_method)
```
bazsi: Could you extract this login code to a function of it's own right? This connect function started to become to large.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

spencercw:
External authentication is mentioned in the [RabbitMQ documentation](https://www.rabbitmq.com/authentication.html). I'm having trouble finding anything AMQP-specific so it's possible it's just a RabbitMQ thing.

I have addressed your comments. Thanks.

modules/afamqp/afamqp.c
Kokan: Travis compiles with `-Wswitch-default`, so it is kinda forced by us.

PR #2627, opened 2019-03-13
version from git describe

Kokan:
The `git describe` is capable to get a much nicer cheap version.

A few things left to do:
- [x] test dbld/rules tarball
- [x] fix dbld/rules deb (some strange error occurs)
- [x] test dbld/rules rpm (at least not worse then the current)
- [x] cmake
- [x] enforce different versioning

### DIFF
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -24,5 +24,11 @@
 
 BASEDIR=$(dirname "$0")
 
-cat $BASEDIR/../VERSION
+cd $BASEDIR/../
+
+if [ -d .git ]; then
+  git describe --tags --dirty --abbrev=7 | sed 's/^syslog-ng-//'
+else
+  cat VERSION
+fi
 
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+#############################################################################
+# Copyright (c) 2019 Balabit
+# Copyright (c) 2018 Kokan <kokaipeter@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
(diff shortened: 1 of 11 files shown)


kira-syslogng:
Build FAILURE

Kokan:
@pzoleex is this a kinda issue you are looking for ?

pzoleex:
@Kokan No, it's another error (the original was permission denied, this is an undefined flag). But this error is caused by the new image (just we didn't test it with dbld change), so we will fix it.

kira-syslogng:
Build FAILURE

folti:

kira-syslogng:
Build FAILURE

folti:

kira-syslogng:
Build SUCCESS

Kokan:
@folti thanks for the fix.

scripts/version.sh
```diff
@@ -0,0 +1,34 @@
...
+# COPYING for details.
+#
+#############################################################################
+
+BASEDIR=$(dirname "$0")
+
+cd $BASEDIR/../
+
+if [ -d .git ]; then
+  git describe --always --tags --dirty --abbrev=7 | cut -d'-' -f3- | tr '-' '+' | tr -d '\n'
```
MrAnno: What do you think of `sed 's/^syslog-ng-//; s/-/+/g' | tr -d '\n'`?
I think removing `syslog-ng` explicitly is better than `cut -d'-' -f3-`.

Do we want `--tags`? I know, we sometimes forget to use annotated tags, but I think we should not.

We might also want to add `--match "syslog-ng-*"` in order to avoid non-version tags.

Kokan: As long as github does not do annotated tags I think `--tags` should be kept.
The current version supports if we may want to use different versioning system like (3.20.1).

kira-syslogng:
Build SUCCESS

PR #2632, opened 2019-03-16
contrib: fix default config file location (systemd)

MrAnno:
syslog-ng has multiple configuration files (syslog-ng.conf, scl.conf,
patterndb.d), so they are usually installed under the /etc/syslog-ng
directory.

Signed-off-by: Janos SZIGETVARI
Signed-off-by: László Várady

### DIFF
--- a/contrib/systemd/syslog-ng@default
+++ b/contrib/systemd/syslog-ng@default
@@ -1,4 +1,4 @@
-CONFIG_FILE=/etc/syslog-ng.conf
+CONFIG_FILE=/etc/syslog-ng/syslog-ng.conf
 PERSIST_FILE=/var/lib/syslog-ng/syslog-ng.persist
 CONTROL_FILE=/var/run/syslog-ng.ctl
 PID_FILE=/var/run/syslog-ng.pid


kira-syslogng:
Build SUCCESS

bazsi:
LGTM pull request analysis was skipped for 442d7458ce23526b83a5262d6e14840425c7ebb7 by [MrAnno](https://github.com/MrAnno). Analysis of future commits will happen as normal.

---

furiel:
Currently syslog-ng installs `syslog-ng.conf` and `scl.conf` into `/etc`, and not `/etc/syslog-ng`. Do you plan to update the install scripts too?
https://github.com/balabit/syslog-ng/blob/b21d85690697f9ae0c5a67b2e22edc524263ba6f/scl/Makefile.am#L44
Similarly for cmake.

There are other places in the code when `/etc/syslog-ng.conf` is hard coded. Some of them might be relevant too.
```
$ ag -F "/etc/syslog-ng.conf" | wc -l
20
```

bazsi:
Never forget the case when syslog-ng is installed for a specific prefix,
e.g. /opt/syslog-ng, in which case etcdir is /opt/syslog-ng/etc and not
/opt/syslog-ng/etc/syslog-ng

The question when that systemd unitfile is used, by the look of it, it only
applies to distribution binaries, in which case /etc/syslog-ng would be
just fine.

Alternatively, we could generate the unit file using the configure script
so that it contains the configured path.

MrAnno:
Currently, it is used by package maintainers, we don't install anything from `contrib`. For example, in Arch Linux, this file is the part of their binary package, but they have to adjust the path first according to the value of the `--sysconfdir`  configure flag.

The auto-generation is a good idea. In that case, we should move the content of the `contrib/systemd` folder out of `contrib`.

MrAnno:
@furiel At this point, my PR could be closed until the auto-generation is implemented (which would open other questions like: where and when do we want to install the generated service file), but I don't think that the current patch is inconsistent or harmful, especially since it is under the `contrib` folder.

For example, we used `/etc/syslog-ng/`in `contrib/upstart/syslog-ng.conf.upstart` as well, thinking only of distribution binaries.
And a few others:

```
contrib/fedora-packaging/syslog-ng.init:10:# config:      /etc/syslog-ng/syslog-ng.conf
contrib/init.d.SuSE:90: test /etc/syslog-ng/syslog-ng.conf -nt /var/run/syslog-ng.pid && echo reload
contrib/init.d.SunOS:36:INIT_CONF="/etc/syslog-ng/syslog-ng.conf" # this file has to exist
contrib/init.d.solaris:12:OPTIONS="-f /etc/syslog-ng/syslog-ng.conf"
contrib/lfs-packaging/syslog-ng.init.lfs:10:# config:      /etc/syslog-ng/syslog-ng.conf
contrib/syslog-ng-debun:1085:   elif [ -d /etc/syslog-ng/ ]; then
contrib/upstart/syslog-ng.conf.upstart:14:  CONF="/etc/syslog-ng/syslog-ng.conf"

packaging/debian/syslog-ng-core.syslog-ng.upstart:14:  CONF="/etc/syslog-ng/syslog-ng.conf"
packaging/debian/syslog-ng.conf:161:@include "/etc/syslog-ng/conf.d/*.conf"

packaging/rhel/syslog-ng.conf:66:@include "/etc/syslog-ng/conf.d/*.conf"
```

furiel:
Ok thanks. File for me. The correct solution can be implemented later.

PR #2636, opened 2019-03-19
graylog2: Add default value for PID

Gorian:
When you forward logs from a remote destination, there is no PID value. This results in a key without a value,

I.E.
```json
{
	"version": "1.1",
	"timestamp": 1553016201,
	"short_message": "message",
	"level": 0,
	"host": "host",
	"_program": "program",
	"_pid": ,
	"_facility": "facility",
	"_class": ""
}
```
Because there is no value for the key `_pid`, this json object is invalid, and thus rejected by the graylog server.

### DIFF
--- a/scl/graylog2/plugin.conf
+++ b/scl/graylog2/plugin.conf
@@ -23,7 +23,7 @@
 
 @requires json-plugin
 
-template-function "format-gelf" "$(format-json version='1.1' host='${HOST}' short_message='${MSG}' level=int(${LEVEL_NUM}) timestamp=int64(${R_UNIXTIME}) _program='${PROGRAM}' _pid=int(${PID}) _facility='${FACILITY}' _class='${.classifier.class}' --key .* --key _*)$(binary 0x00)";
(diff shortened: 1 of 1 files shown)


bazsi:
This is a bug in $(format-json) unfortunately, this is just a workaround for that bug.

The bug in $(format-json) and the underlying type hinting infrastructure is that it accepts the empty string as a valid value (we drop these by default), but then it assumes that an empty string also makes sense when embedded in a JSON.

#2639 is a more comprehensive version of the fix, which should make this unnecessary.

bazsi:
One question: what would you prefer in case PID is not available in the message? My branch above (#2639) means that the PID key will be dropped completely. I was thinking that's preferrable, but I don't know graylog enough. Your patch will include PID with a value of 0.

If we prefer the latter, then both branches are needed, if it's good that the entire key is dropped then this one isn't, only #2639.

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

bazsi:
As #2639  is merged, I would close this without merging, unless @Gorian can elaborate if this is needed in addition to that.

Gorian:
> As #2639 is merged, I would close this without merging, unless @Gorian can elaborate if this is needed in addition to that.

Nope! If you guys fixed it with another PR, that's great! Thanks :)

bazsi:
Thanks for the feedback, closing...

PR #2637, opened 2019-03-19
Add support for Graylog GELF over TLS and UDP

Gorian:
Graylog GELF supports both UDP and TCP+TLS. There is no reason you can't default to TCP but add support for using the plugin while changing the transport to UDP. Also added a block for TLS.

### DIFF
--- a/scl/graylog2/plugin.conf
+++ b/scl/graylog2/plugin.conf
@@ -25,10 +25,23 @@
 
 template-function "format-gelf" "$(format-json version='1.1' host='${HOST}' short_message='${MSG}' level=int(${LEVEL_NUM}) timestamp=int64(${R_UNIXTIME}) _program='${PROGRAM}' _pid=int(${PID}) _facility='${FACILITY}' _class='${.classifier.class}' --key .* --key _*)$(binary 0x00)";
 
-block destination graylog2(host("127.0.0.1") port(12201) template("$(format-gelf)") ...) {
+block destination graylog2(host("127.0.0.1") port(12201) transport(tcp) template("$(format-gelf)") ...) {
 	network("`host`"
                 port(`port`)
-		transport(tcp)
+		transport(`transport`)
 		template("`template`")
 		`__VARARGS__`);
+};
+
+block destination graylog2-tls(host("127.0.0.1") port(12201) ca-dir("/etc/ssl/certs") peer-verify("required-trusted") template("$(format-gelf)") ...) {
+    network(
+        "`host`"
+        port(`port`)
+        transport(tls)
+        tls (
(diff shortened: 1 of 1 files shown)


Kokan:

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build FAILURE

Kokan:

kira-syslogng:
Build FAILURE

Kokan:

kira-syslogng:
Build SUCCESS

scl/graylog2/plugin.conf
```diff
@@ -25,10 +25,23 @@
...
+block destination graylog2(host("127.0.0.1") port(12201) transport(tcp) template("$(format-gelf)") ...) {
 	network("`host`"
                 port(`port`)
-		transport(tcp)
+		transport(`transport`)
 		template("`template`")
 		`__VARARGS__`);
 };
+
+block destination graylog2-tls(host("127.0.0.1") port(12201) ca-dir("/etc/ssl/certs") peer-verify("required-trusted") template("$(format-gelf)") ...) {
```
furiel: Instead of setting `ca-dir` to` "/etc/ssl/certs"`, could you try if `use_system_cert_store(yes)` would work for you? Different distributions handle cert store different way, and I think it would be safer to use that.

Another question: do we need separate block for tls? Me seems the main difference is that we want to prepare a tls block in the latter case. But one could pass it as `__VARARGS_` using the original `graylog2` block, I think.

Something like:
```
destination { graylog2(transport(tls) tls(peer-verify(yes) [...])) };
```

Would this work for you?

Gorian: Sure. I don't know syslog-ng that well, and I didn't know that was an option. Works fine for me.

furiel: I just found out that the `use-system-cert-store` part of my request is invalid: it is a `http` destination option, and it is not available for the `network` driver.

Passing `tls` block in `__VARARGS__` is accepted by the grammar on the other hand, so I think that part could still work.

Kokan: @Gorian are you planning to do a follow up based on @furiel's comment ?

Kokan:
@Gorian I did the updates @furiel suggested, and created a PR based on your work (https://github.com/balabit/syslog-ng/pull/2657). I am closing this PR in favour of that one.

Thank you for your contribution.

PR #2640, opened 2019-03-20
app-parser, pseudofile: fix crash with grammar error

furiel:
See the following example.

```
@version: 3.20
application a[my-apparser] {
    rewrite { set("b" value("b") };
};

log {
  parser { app-parser(topic(my-apparser)); };
};
```

In this case, the error happens in the `application_options` in the
grammar file (`KW_REWRITE` is not allowed), after we freed `$2` and
`$4`. However, in case of grammar error, lexer frees these tokens
too, hence double free.

As a resolution, the `free` calls are moved to the end of the grammar file.
This way only grammar will free these tokens.

-----

Similarly, with pseudofile:
```
@version: 3.20

log {
  destination { pseudofile("/tmp/foo" disk-buffer(a) ); };
};
```
causes crash.

### DIFF
--- a/modules/appmodel/appmodel-grammar.ym
+++ b/modules/appmodel/appmodel-grammar.ym
@@ -70,12 +70,12 @@ application_definition
         : KW_APPLICATION string '[' string ']'
           {
 	    last_application = application_new($2, $4);
-            free($2);
-            free($4);
           }
 	  '{' application_options '}'
 	  {
 	    $$ = last_application;
+            free($2);
+            free($4);
           }
         ;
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build SUCCESS

Kokan:
@furiel is this the only place with such misuse of `$n` variables ?

furiel:
> is this the only place with such misuse of $n variables ?

I'm afraid I don't know that, nor know a way to find such problems (other than checking every free in every grammar). There are no such instances that I know of currently besides these.

furiel:
@Kokan found the same error with `pseudofile`. Added a patch to fix that occurrence too.

kira-syslogng:
Build SUCCESS

PR #2643, opened 2019-03-22
python: set_timestamp normalization

Kokan:
The current `LogMessage.set_timestamp` behaved differently for py2 or py3; and had issue with *aware* datetime.

Fixes:
* py3 use local timezone in case of *naive* datetime
* py2 could not accept *aware* datetime

Also creates a code that is compatible with both py2 and py3 so does not have to be in compat layer.

Example for creating *aware* datetime
```
import datetime
import pytz

datetime.datetime(2019,9,14,12,34,56,tzinfo=pytz.timezone('Asia/Shanghai'))
```

### DIFF
--- a/modules/python/compat/compat-python.h
+++ b/modules/python/compat/compat-python.h
@@ -26,7 +26,6 @@
 
 #include <Python.h>
 #include "syslog-ng.h"
-#include "timeutils/unixtime.h"
 
 #if (SYSLOG_NG_ENABLE_PYTHONv2)
 #define PYTHON_BUILTIN_MODULE_NAME "__builtin__"
@@ -45,9 +44,6 @@
 
 void py_init_argv(void);
 PyObject *int_as_pyobject(gint num);
-
-void py_datetime_init(void);
-gboolean py_datetime_to_logstamp(PyObject *py_timestamp, UnixTime *logstamp);
 
 gint pyobject_as_int(PyObject *object);
 #endif
--- a/modules/python/compat/compat-python.h
+++ b/modules/python/compat/compat-python.h
@@ -31,6 +31,11 @@
 #if (SYSLOG_NG_ENABLE_PYTHONv2)
 #define PYTHON_BUILTIN_MODULE_NAME "__builtin__"
 #define PYTHON_MODULE_VERSION "python2"
+
+/*
+ * Macro from python3 Include/datetime.h file
+ */
+#define PyDateTime_DELTA_GET_SECONDS(o)      (((PyDateTime_Delta*)o)->seconds)
 #endif
 
 #if (SYSLOG_NG_ENABLE_PYTHONv3)
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -22,12 +22,16 @@
  *
  */
 #include "python-logmsg.h"
+#include "compat/compat-python.h"
 #include "python-helpers.h"
 #include "logmsg/logmsg.h"
 #include "messages.h"
 #include "timeutils/cache.h"
 #include "str-utils.h"
 #include "timeutils/unixtime.h"
+#include "timeutils/misc.h"
+
+#include <datetime.h>
 
 int
 py_is_log_message(PyObject *obj)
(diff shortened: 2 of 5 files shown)


kira-syslogng:
Build SUCCESS

modules/python/compat/compat-python.h
```diff
@@ -31,6 +31,11 @@
 #if (SYSLOG_NG_ENABLE_PYTHONv2)
 #define PYTHON_BUILTIN_MODULE_NAME "__builtin__"
 #define PYTHON_MODULE_VERSION "python2"
+
+/*
+ * Macro from python3 Include/datetime.h file
+ */
+#define PyDateTime_DELTA_GET_SECONDS(o)      (((PyDateTime_Delta*)o)->seconds)
```
furiel: Can you add an `#ifndef` guard around it?

modules/python/python-logmsg.c
```diff
@@ -270,6 +274,83 @@ py_log_message_set_pri(PyLogMessage *self, PyObject *args, PyObject *kwrds)
...
+static gboolean
+py_datetime_to_logstamp(PyObject *py_timestamp, UnixTime *logstamp)
+{
+  if (!PyDateTime_Check(py_timestamp))
+    {
+      PyErr_Format(PyExc_TypeError, "datetime expected in the first parameter");
+      return FALSE;
+    }
+
+  PyObject *py_posix_timestamp = _datetime_timestamp(py_timestamp);
```
furiel: `_datetime_timestamp` can return `NULL`. Please handle error here.

modules/python/python-logmsg.c
```diff
@@ -270,6 +274,83 @@ py_log_message_set_pri(PyLogMessage *self, PyObject *args, PyObject *kwrds)
...
+{
+  PyObject *py_utcoffset = _py_invoke_method_by_name(py_datetime, "utcoffset", NULL, "PyDateTime",
+                                                     "py_datetime_to_logstamp");
+  if (!py_utcoffset)
+    {
+      PyErr_Format(PyExc_ValueError, "Error obtaining timezone info");
+      return -1;
+    }
+
+  if (py_utcoffset == Py_None)
```
furiel: Even in case of `Py_None`, py_utcoffset must be decrefed: https://docs.python.org/3/c-api/none.html#c.Py_None.

modules/python/python-logmsg.c
```diff
@@ -270,6 +274,83 @@ py_log_message_set_pri(PyLogMessage *self, PyObject *args, PyObject *kwrds)
...
+
+static gint
+_datetime_get_gmtoff(PyObject *py_datetime)
+{
+  PyObject *py_utcoffset = _py_invoke_method_by_name(py_datetime, "utcoffset", NULL, "PyDateTime",
+                                                     "py_datetime_to_logstamp");
+  if (!py_utcoffset)
+    {
+      PyErr_Format(PyExc_ValueError, "Error obtaining timezone info");
+      return -1;
```
furiel: Returning -1 is treated as valid return value at the caller. It would worth signalling the error to the caller.

modules/python/compat/compat-python.h
Kokan: What do you want to achieve with that ?

furiel: I was thinking, if at some point someone wants to include this python3/timestamp.h header, they would run into trouble. It also occurred me that you should have included it instead of picking a macro. But this macro is not available in python2, so you would still have to deal with compatibility, so we wouldn't have won anything. Just picking this line is sufficient.

I do not want to force this, just me seems a good practice to add guard in such cases.

modules/python/python-logmsg.c
Kokan: I was under the impression that `PyErr_Format` already does signal for python, based on trial actually it does not make any difference. (which makes me think the other way around). Maybe you have some extra info about it.

modules/python/compat/compat-python.h
Kokan: This is a compatibility layer, and the macro exists only in python3, so I only provided it for python2 (anybody includes python3 won't get in trouble).

So one should always include `datetime.h` if want this macro, and as well the compat (maybe it would worth to make the same structure as for other compat headers in syslog-ng).

If this macro and the one in python3 is colliding somebody screw up big time, and I want to put the pain on them instead of silently hiding the issue.

modules/python/python-logmsg.c
Kokan: Fixed.

modules/python/python-logmsg.c
Kokan: Fixed.

kira-syslogng:
Build SUCCESS

modules/python/python-logmsg.c
furiel: To throw exception from C for the user, the c callback function needs to return `NULL`. In that case the python machinery does the right thing, for example allowing user to catch the exception.

It is a little tedious work, one needs to make sure to propagate error through the whole call chain, just to finally return NULL.

Kokan: Thanks, kinda strange that it does not allow the user to catch, but raises the issue anyway. Fixed.

kira-syslogng:
Build SUCCESS

modules/python/compat/compat-python.h
furiel: Okok, I agree.

PR #2645, opened 2019-03-24
logwriter: Avoid null pointer access from suspended writer

kyeongy:
There is a chance with rare timing that a suspended writer tries to
access already freed client, which causes a SEGV.

So make sure the client is valid in log_writer_error_suspend_elapsed().

### DIFF
--- a/dbld/images/helpers/pip_packages.manifest
+++ b/dbld/images/helpers/pip_packages.manifest
@@ -6,7 +6,6 @@ ply                     [centos, debian, ubuntu]
 pylint                  [centos-7, debian, ubuntu]
 
 # pip packages for python functional tests
-mockito                 [centos, debian, ubuntu]
 pathlib2                [centos, debian, ubuntu]
 psutil                  [centos, debian, ubuntu]
 pytest                  [centos, debian, ubuntu]
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,8 +4,6 @@ pep8==1.7.1
 pylint==1.8.2
 astroid==1.6.1
 logilab-common<=0.63.0
-colorlog
-mockito
 pathlib2
 psutil
 pytest


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:

kira-syslogng:
Build SUCCESS

Kokan:
@kyeongy Did you happened to run into this null pointer dereference ?

kyeongy:
@Kokan, yes. I could not reproduce easily but happened once. Here is a coredump analysis if you like to see. By the way, I think it happened when 
1. On a telnet session ("/dev/pts/0"), display syslog to the screen.
2. Logout the telnet session, which removes "/dev/pts/0".
3. Remove config for "/dev/pts/0" and reload syslog-ng config.
4. Probably issue started happening between 2 and 3 and writer fails and keeps adding it to suspend timer. Eventually a null pointer access happened.

#0  log_proto_client_get_fd (s=0x0) at lib/logproto/logproto-client.h:125
125	  return s->transport->fd;
--------------------------------------------------------------------------------
Thread Information:
  Id   Target Id         Frame 
* 1    Thread 0x3fffb6b57000 (LWP 495) log_proto_client_get_fd (s=0x0) at lib/logproto/logproto-client.h:125
--------------------------------------------------------------------------------
Current Thread Local Variables:
No locals.
--------------------------------------------------------------------------------
Thread Backtraces:
Thread 1 (Thread 0x3fffb6b57000 (LWP 495)):
#0  log_proto_client_get_fd (s=0x0) at lib/logproto/logproto-client.h:125
#1  log_writer_error_suspend_elapsed (s=0x100ac370) at lib/logwriter.c:306
#2  0x00003fffb680019c in iv_run_timers (st=0x1001c140) at iv_timer.c:149
#3  0x00003fffb6802bf4 in iv_main () at iv_main_posix.c:107
#4  0x00003fffb6aa75f0 in main_loop_run (self=0x3fffb6b22f80 <main_loop>) at lib/mainloop.c:530
#5  0x0000000010001cdc in main (argc=1, argv=0x3fffea952768) at syslog-ng/main.c:308
--------------------------------------------------------------------------------

Kokan:
@kyeongy is it possible for you to share your *minor modifications* ? I couldn't reproduce it on current master, have you tried ?

kyeongy:
The change I made locally earlier is as following, which was to fix similar NULL pointer access and it happened again so the new patch I've just submitted.

As I explained above regarding telnet session when the issue happened, I saw this error log every 60 seconds and syslog-ng crashed after about 50 times:
```
#31030 2019-02-23T11:24:43+09:00 syslog.err hostname syslog-ng[495]: I/O error occurred while writing; fd='68', error='Input/output error (5)'
#31031 2019-02-23T11:24:43+09:00 syslog.notice hostname syslog-ng[495]: Suspending write operation because of an I/O error; fd='68', time_reopen='60'
```

Note that I only saw once and not able to reproduce again recently. So I think there are some timing issue syslog-ng can get in this bad state.

```
diff --git a/lib/logwriter.c b/lib/logwriter.c
index 2f2d8922d..2312cd1b8 100644
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -229,10 +229,10 @@ log_writer_work_finished(gpointer s)
       log_writer_broken(self, NC_WRITE_ERROR);
       if (self->proto)
         {
-          log_writer_suspend(self);
           msg_notice("Suspending write operation because of an I/O error",
                      evt_tag_int("fd", log_proto_client_get_fd(self->proto)),
                      evt_tag_int("time_reopen", self->options->time_reopen));
+          log_writer_suspend(self);
         }
       return;
     }
@@ -1450,6 +1450,9 @@ log_writer_deinit(LogPipe *s)
   if (iv_timer_registered(&self->reopen_timer))
     iv_timer_unregister(&self->reopen_timer);
 
+  if (iv_timer_registered(&self->suspend_timer))
+    iv_timer_unregister(&self->suspend_timer);
+
   ml_batched_timer_unregister(&self->suppress_timer);
   ml_batched_timer_unregister(&self->mark_timer);
 
```

MrAnno:
Hi @kyeongy,

Sorry for the late answer.
I've successfully reproduced the crash:

```
@version: 3.21

options {
    time-reopen(5);
};

log {
    source{ example-msg-generator(freq(0.1)); };
    destination {
        file("/mnt/outt");
    };
};
```

1. In my environment, `/mnt` is a small loop device. I filled the free space, creating a write error:

```
I/O error occurred while writing; fd='12', error='No space left on device (28)'
Suspending write operation because of an I/O error; fd='12', time_reopen='5'
```

2. After the suspend timer kicked in, I removed the `file()` destination from my config and reloaded syslog-ng.

As you mentioned, this bug is a timing issue (task ordering issue), not a race condition between threads. The PR you've submitted looks like a workaround for the problem, because `self->proto` should never be `NULL` when `log_writer_error_suspend_elapsed()` is scheduled.

The actual fix for the problem is to stop the timer explicitly in `log_writer_deinit()`, so a reload won't let the callback of the suspend timer be scheduled (`watches_running` is `FALSE` in this case):

```diff
diff --git a/lib/logwriter.c b/lib/logwriter.c
index bf63b480f..ea5d6846c 100644
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -1452,6 +1452,9 @@ log_writer_deinit(LogPipe *s)
   if (iv_timer_registered(&self->reopen_timer))
     iv_timer_unregister(&self->reopen_timer);

+  if (iv_timer_registered(&self->suspend_timer))
+    iv_timer_unregister(&self->suspend_timer);
+
   ml_batched_timer_unregister(&self->suppress_timer);
   ml_batched_timer_unregister(&self->mark_timer);
```

The fix actually belongs to you. :)

You mentioned that this patch did not fix the crash and the NULL pointer access happened again.
Are you sure you fully recompiled/reinstalled syslog-ng with the fix included?

My reproduction is stable and stopping the timer clearly fixed the NULL pointer dereference.
I tried to find other ordering issues, but I've found nothing.

If you agree with my deduction, could you update your branch with the timer unregistration patch, please?

Thank you very much for the PR, this is a tricky one :)

kyeongy:
@MrAnno, thanks for your tests for this issue.

We have had the "timer unregistration patch" in our build for 6 months so I'm sure the NULL pointer access happened even with your suggested fix.
Note that during a half year, we have seen this only once so it is definitely rare timing issue.

My analysis when the issue happened (when destination file is "/dev/pts/0") was as following (in case it makes sense to you):
_It hasn't been reproduced but it indicates that logging out from a telnet session removes the tty ("/dev/pts/0") and syslog-ng fails to write. Syslog-ng then intends to retry later and put the writer in a timer. When the timer expires, it finds the writer client was already freed. And accessing a NULL pointer causes SEGV._

I can include "timer unregistration patch" but still my original fix seems necessary.
Let me know your opinion.

Cheers,
Kyeong

MrAnno:
@kyeongy You are absolutely right, I've found another code path, where the suspend timer should be stopped.

My reproduction steps are still not TTY-based, but I think this triggers the same bug:
1. Make a write error, for example, fill the free space on disk. The suspend timer kicks in.
2. Make the destination file unavailable for opening: delete it, or change permissions `chmod 000 /mnt/outt`
3. Initiate a reopen event, for example: `sbin/syslog-ng-ctl reopen`

```gdb
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fc4efd2a5cd in log_proto_client_get_fd (s=0x0) at ../lib/logproto/logproto-client.h:167
167       return s->transport->fd;
[Current thread is 1 (Thread 0x7fc4ee553a80 (LWP 28500))]
```

A proposed fix for the problem can be found here:
https://github.com/MrAnno/syslog-ng/commits/logwriter-suspend

Can you please try the branch above? What do you think? I hope this is the last path where we forgot to stop the timer :)

kyeongy:
@MrAnno  That sounds good and possibly that prevents the NULL pointer access I had experienced. I'm happy with those patches you prepared and I can simply drop my patch.

Although I've looked at the code in log_writer_stop_watches() further and figured out that it has intention of stopping suspend_timer as well. It seems "suspend_timer" can be set while "watches_running" flag is FALSE, so the logic didn't work always.

So possible tidyup would be either
1. Move out "suspend_timer" from the "if (self->watches_running)" block
```
  if (self->watches_running)
    {
       ...
      self->watches_running = FALSE;
    }
  if (iv_timer_registered(&self->suspend_timer))
    iv_timer_unregister(&self->suspend_timer);
```
or
2. Set "self->watches_running" flag to TRUE when "suspend_timer is set.

or 
3. if you like to keep log_writer_stop_suspend_timer() function then go with it but remove it from log_writer_stop_watches() to avoid confusion of calling same routine twice.

MrAnno:
@kyeongy Thanks for the reply.

In my opinion, `start_watches()` and `stop_watches()` should be somewhat symmetrical, so your third suggestion might be the good choice.

`log_writer_arm_suspend_timer()` is called from `start_watches()` (through `update_watches()`), but it is used in `log_writer_suspend()` as well, breaking the symmetry.

Removing `log_writer_stop_suspend_timer()` from `stop_watches()` and calling it after each occurrence of the `stop_watches()` function call should work.

Would you like to submit this modification, or should I just do it and update your pull request?

----

Unfortunately, `reopen_timer` has the same issue, it is "stopped twice" in `log_writer_deinit()`.

kira-syslogng:
do nothing -> CI won't start)

kyeongy:
@MrAnno 
> Removing log_writer_stop_suspend_timer() from stop_watches() and calling it after each occurrence of the stop_watches() function call should work.
I'm happy with that.

> Would you like to submit this modification, or should I just do it and update your pull request?
It would be great if you update and push.
Thanks

MrAnno:
@kyeongy Sorry, I've accidentally pushed the wrong branch and the stale push closed the pull request automatically. I can't reopen it for some reason, I guess because GitHub thinks the branch is merged.

I'll open a new one containing our work.

kira-syslogng:
Build SUCCESS

PR #2647, opened 2019-03-26
cmake: disable_all_modules support

furiel:
- If `-DDISABLE_ALL_MODULES=yes` specified for cmake, all switchable modules are disabled by default. One can still override specific modules by: `-DDISABLE_ALL_MODULES=yes -DENABLE_PYTHON=yes`. The order of definitions do not matter.
- Fixed a bug, when `-DENABLE_IPV6=no` resulted in compile error in a unit test.
- Fixed a bug when json unit tests were still tried to compiled even without json support.
- Removed `-DENABLE_BASICFUNCS`, as it causes compile error in some unit tests, it is not switchable in autotools, and does not have any external dependencies.

### DIFF
--- a/cmake/module_switch.cmake
+++ b/cmake/module_switch.cmake
@@ -1,0 +1,44 @@
+#############################################################################
+# Copyright (c) 2019 Balabit
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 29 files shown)


kira-syslogng:
Build SUCCESS

cmake/module_switch.cmake
```diff
@@ -0,0 +1,42 @@
...
+# COPYING for details.
+#
+#############################################################################
+
+function (module_switch NAME HELP_STRING)
+  if (DEFINED ${NAME})
+    return()
+  endif()
+
+  if (DISABLE_ALL_MODULES)
```
Kokan: This way the `DISABLE_ALL_MODULES` is not visible on gui/tui (like `ccmake`). I would prefer to have all user option to be visible. What do you think ?

furiel: Agreed. I will do so.

kira-syslogng:
Build SUCCESS

PR #2648, opened 2019-03-26
stats: deindex pruned counters/clusters

alltilla:
Dynamic counters/clusters were freed after they became expired, but they stayed in the `counter_index`, which is used for `get` queries. When we tried to access these freed counters from the index, we got undefined behavior, possibly SEGFAULT.


kira-syslogng:
Build SUCCESS

MrAnno:
Nice catch!

Issue #141, opened 2014-06-18
patterndb parser: PROGRAM and MSG should be configurable


faxm0dem:
Currently the patterndb parser enforces the use of the following macros:
- `$PROGRAM` for the _ruleset patterns_
- `$MESSAGE` for the _rule patterns_

The latter is not entirely true, as it seems to be possible according to the _source_ to use `template()` in a `db_parser` block. It is however not officially possible, from reading the docs.

Nevertheless I have the strong feeling this is an unnecessary limitation that could be easily lifted without hacking the config (copy whatever is in `$custom_message` into `$MESSAGE` and same for `*PROGRAM`). This is particularily true when using _e.g._ _JSON_ formatted input.

```
parser p_new_pdb {
  file("/var/lib/syslog-ng/patterndb/default.xml")
  message_template("$my_message")
  program_template("$my_program")
};
```

ihrwein:
I just checked this. The `template()` parameter should be supported by dbparser and it's the same as your `message_template()`. Aliasing it is easy.

The `program_template()` doesn't exists, yet. Do you think this is dbparser specific or general enough to be supported by all parsers?

faxm0dem:
program_template seems very specific to dbparser, as others only use one template.
If one were to generalize this, it would probably make more sense to allow for `template` to be passed an array, e.g. `[$MSG, $PROGRAM, ...]`

bazsi:
I think this is a good short term feature. I originally wanted to make
db-parser n dimensional, each making it possible which nvpair it would use
for matching.

But I think we don't want that but rather add this kind of filtering into
the config.

hadret:
I'm not 100% sure whether my case apply here as well, but by the overall look of it it seems so. I'm having a structured way of `$PROGRAM` name that I need to parse in order to aggregate logs the way I want to. Logs themselves can originate from the same and/or different hosts (multiple instances per host) at the same time, so I went with the following patter for the `$PROGRAM`: `app_part1/app_part2/app_part3`. The first two parts are the same for every instance, the differentiating factor is the third part. I wanted to have the following structure in the logs tree: `/var/log/`app_part1`/`app_part2`.log -- but in order to be able to do that, I would need to be able to parse `$PROGRAM`, which is, to my understanding, currently impossible.

bazsi:
No, its not. The ruleset level pattern refers to the program name and you
can use the same pattern language there.

[...]

PR #2651, opened 2019-03-29
db-parser: support for program_template

furiel:
Fixes: https://github.com/balabit/syslog-ng/issues/141

For now, the rule selection in db-parser is based on the `PROGRAM` macro value. This patchset makes this configurable: the selector can be specified by template.

There is a `template` option already available in `db-parser`, which defines the message that is processed by the parser (similarly to any other parsers). To avoid confusion with `program-template`, this patchset adds an alias for it: `message-template`.

There is an interesting feature in db-parser: pattern matching works in `PROGRAM`. It turned out that I do not need to deal with this functionality separately, it automatically worked after my changes. While I was in doubt, I saw that was not covered by a unit test, so I added one. In the example below, you can see an example for matching in program (`num`).

Example:
In the example below, `example-msg-generator` generates a general message. The reason we have a match is because the `program-template` and `message-template` is set in `db-parser`.

syslog-ng.conf
```
@version: 3.20

log {
  rewrite { set("5" value("foo")); };
  parser { db-parser(file("sshd.xml") message-template("sshd\n") program-template("sshd ${foo}")); };
  destination { file(/dev/stdout template("${.classifier.class} || ${num}\n")); };
};
```

sshd.xml
```
<patterndb version='5'><ruleset name='sshd' id='12345678'>
<patterns>
<pattern>sshd @NUMBER:num@</pattern>
</patterns>
<rules>
  <rule id='12347598' class='sshd' provider='balabit'>
     <patterns><pattern>sshd</pattern></patterns>
  </rule>
</rules>
</ruleset>
</patterndb>
```

Notes to reviewers:
- The functionality is generalized, but I still used `program-template` as a name, although `selector` or similar might have been better. The reason I stayed with it is because the term `program` is deeply part of the thought process in the module (grep for program in the module).
- grouping-by: grouping-by works a little bit differently, and the key is not only the program, but the collection of host+program+pid is used as a pre filter, then there is the `key-template` option. This changeset should not affect that behaviour.
-  pdbtool: pdbtool uses program for two reasons. One is to match an example message against the database. `pdbtool match` already has a `-P` option, that can be used to specify the program string. That option does not support templates, but for match testing I think that is good enough as it is. The other usage is testing message against an example message. Example message can be specified inside a patterndb database, and it can be used to make the database unit test-able. Similarly to match, it might make sense to support templates there too, but I think it is good enough for its purpose.

### DIFF
--- a/modules/dbparser/dbparser-grammar.ym
+++ b/modules/dbparser/dbparser-grammar.ym
@@ -67,6 +67,7 @@ SyntheticMessage *last_message;
 %token KW_DROP_UNMATCHED
 %token KW_FILE
 %token KW_PROGRAM_TEMPLATE
+%token KW_MESSAGE_TEMPLATE
 
 %type <num> stateful_parser_inject_mode
 %type <ptr> synthetic_message
@@ -106,6 +107,7 @@ parser_db_opt
           {
              log_db_parser_set_program_template_ref(last_parser, $3);
           }
+        | KW_MESSAGE_TEMPLATE '(' template_content ')'    { log_parser_set_template(last_parser, $3); }
 	| stateful_parser_opt
         ;
 
--- a/modules/dbparser/dbparser-grammar.ym
+++ b/modules/dbparser/dbparser-grammar.ym
@@ -66,6 +66,7 @@ SyntheticMessage *last_message;
 %token KW_VALUE
 %token KW_DROP_UNMATCHED
 %token KW_FILE
+%token KW_PROGRAM_TEMPLATE
 
 %type <num> stateful_parser_inject_mode
 %type <ptr> synthetic_message
@@ -101,6 +102,10 @@ parser_db_opts
 parser_db_opt
         : KW_FILE '(' string ')'                		{ log_db_parser_set_db_file(((LogDBParser *) last_parser), $3); free($3); }
 	| KW_DROP_UNMATCHED '(' yesno ')'			{ log_db_parser_set_drop_unmatched(((LogDBParser *) last_parser), $3); };
+        | KW_PROGRAM_TEMPLATE '(' template_content ')'
+          {
+             log_db_parser_set_program_template_ref(last_parser, $3);
+          }
 	| stateful_parser_opt
         ;
 
@@ -201,4 +206,3 @@ context_scope
 /* INCLUDE_RULES */
 
 %%
-
--- a/modules/dbparser/patterndb.c
+++ b/modules/dbparser/patterndb.c
@@ -714,9 +714,7 @@ pattern_db_process_with_custom_message(PatternDB *self, LogMessage *msg, const g
   PDBLookupParams lookup;
 
   pdb_lookup_params_init(&lookup, msg);
-  lookup.message_handle = LM_V_NONE;
-  lookup.message_string = message;
-  lookup.message_len = message_len;
+  pdb_lookup_params_override_message(&lookup, message, message_len);
   return _pattern_db_process(self, &lookup, NULL);
 }
 
--- a/modules/dbparser/patterndb.c
+++ b/modules/dbparser/patterndb.c
@@ -65,6 +65,7 @@ struct _PatternDB
   GStaticRWLock lock;
   PDBRuleSet *ruleset;
   CorrellationState correllation;
+  LogTemplate *program_template;
   GHashTable *rate_limits;
   TimerWheel *timer_wheel;
   GTimeVal last_tick;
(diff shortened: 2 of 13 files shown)


kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng retest this please test branch=patterndb-program-template;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=patterndb-program-template;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=patterndb-program-template;

kira-syslogng:
Build SUCCESS

furiel:
Needed to force push because of confict with patterndb alphabetic ordering PR.

furiel:
@kira-syslogng retest this please test branch=patterndb-program-template;

kira-syslogng:
Build FAILURE

furiel:
needed to rebase the test branch due to http expedite, so @kira-syslogng retest this please test branch=patterndb-program-template;

kira-syslogng:
Build SUCCESS

modules/dbparser/dbparser-grammar.ym
```diff
@@ -101,6 +103,19 @@ parser_db_opts
 parser_db_opt
         : KW_FILE '(' string ')'                		{ log_db_parser_set_db_file(((LogDBParser *) last_parser), $3); free($3); }
 	| KW_DROP_UNMATCHED '(' yesno ')'			{ log_db_parser_set_drop_unmatched(((LogDBParser *) last_parser), $3); };
+        | KW_PROGRAM_TEMPLATE '(' template_content ')'
+          {
+             log_db_parser_set_program_template_ref(last_parser, $3);
+          }
+        | KW_MESSAGE_TEMPLATE '(' string ')'    {
+                                                  LogTemplate *template;
```
bazsi: why doesn't this use template_content?

modules/dbparser/patterndb.c
```diff
@@ -699,12 +707,19 @@ _pattern_db_process(PatternDB *self, PDBLookupParams *lookup, GArray *dbg_list)
   return process_params->rule != NULL;
 }
 
+static void
+_prepare_lookup_params(PDBLookupParams *lookup, LogMessage *msg, LogTemplate *program_template)
+{
```
bazsi: shouldn't we just introduce program_template argument to pdb_lookup_params_init()? That would eliminate the need for this wrapper function, which is difficult to name in a way that actually tells what it does.

modules/dbparser/dbparser-grammar.ym
furiel: It is a nice simplification indeed. Thanks.

modules/dbparser/patterndb.c
furiel: You are right! I'll look into it.

kira-syslogng:
Build SUCCESS

modules/dbparser/patterndb.c
furiel: Fixed.

furiel:
@kira-syslogng retest this please test branch=patterndb-program-template;

kira-syslogng:
Build SUCCESS

modules/dbparser/tests/test_patterndb.c
```diff
@@ -738,6 +738,19 @@ Test(test_pathutils, test_pdb_get_filenames, .init = test_pdb_get_filenames_setu
   g_ptr_array_free(filenames, TRUE);
 }
 
+Test(pattern_db, match_in_program)
+{
+  gchar *filename;
+  PatternDB *patterndb = _create_pattern_db(pdb_test_match_in_program, &filename);
+
+  LogMessage *msg = _construct_message("sshd 5", "almafa");
```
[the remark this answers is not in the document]
furiel: Fixed, thanks!

modules/dbparser/tests/test_patterndb.c
```diff
@@ -751,6 +751,25 @@ Test(pattern_db, match_in_program)
...
+Test(pattern_db, test_program_template)
+{
+  gchar *filename;
+  PatternDB *patterndb = _create_pattern_db(pdb_test_program_template, &filename);
+
+  LogTemplate *template = log_template_new(configuration, NULL);
+  cr_assert(log_template_compile(template, "sshd 5", NULL));
+  pattern_db_set_program_template(patterndb, template);
+
+  LogMessage *msg = _construct_message("somethingelsethatdoesnotmatch", "almafa kortefa");
```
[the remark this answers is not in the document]
furiel: Fixed, thanks!

furiel:
@kira-syslogng retest this please test branch=patterndb-program-template;

furiel:
@kira-syslogng retest this please test branch=patterndb-program-template;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=patterndb-program-template;

kira-syslogng:
Build SUCCESS

furiel:
As release is delayed for a few days, I will merge this.

PR #2652, opened 2019-03-31
db-parser: fix memory leak

furiel:
There was earlier a memory leak in pattern-db: the rules in patterndb were leaked. This was attempted to be fixed in https://github.com/balabit/syslog-ng/pull/2311. However the fix was incomplete, the last rule was still not be freed.

The problem was in the end element logic: due to the `== 0`, the pdb_rule_unref was called only in case of parse error. Hence rule was never freed at the `</rule>` end element.

This patch reverts https://github.com/balabit/syslog-ng/pull/2311, and fixes the return value problem.


kira-syslogng:
Build SUCCESS

PR #2653, opened 2019-04-01
Tele voice msg source

Kokan:
Introduces a new python source based source called `tele-voice`.
This source does not have any parameter as of now, it records voice via microphone and uses voice recognition to translate speech into messages.

With this improvement syslog-ng is capable to collects information in the server rooms and personal computer possible correlate them with audit logs.

Later this could be improved with features to specify the interval, or picking up not only natural languages.


kira-syslogng:
Build FAILURE

PR #2655, opened 2019-04-02
grouping-by: use after free, memory leak, missing init calls of filters

furiel:
This patchset fixes:
- filter expressions of grouping-by were not freed
- init method of the filter expressions were not called
- use after free when reload with a non-closed context

### DIFF
--- a/modules/dbparser/groupingby.c
+++ b/modules/dbparser/groupingby.c
@@ -48,7 +48,21 @@ typedef struct _GroupingBy
   FilterExprNode *having_condition_expr;
 } GroupingBy;
 
+typedef struct
+{
+  CorrellationState *correllation;
+  TimerWheel *timer_wheel;
+} GroupingByPersistData;
+
 static NVHandle context_id_handle = 0;
+
+static void
+_free_persist_data(GroupingByPersistData *self)
+{
+  correllation_state_free(self->correllation);
+  timer_wheel_free(self->timer_wheel);
+  g_free(self);
+}
 
 void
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build SUCCESS

modules/dbparser/groupingby.c
```diff
@@ -383,6 +397,25 @@ grouping_by_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_
   return TRUE;
 }
 
+static void
+_load_data_from_persist(GroupingBy *self, GlobalConfig *cfg)
```
MrAnno: This does not just load data from persist, it creates state when the persist is empty, so it would be the best if the name could reflect this.

What do you think of `_load_correlation_state` (TimerWheel is actually coupled with the correlation state) or `_load_grouping_by_state` or `_load_state`?

MrAnno:
db-parser() is not affected by the use-after-free bug, because the entire PatternDB instance is stored and loaded from persist.

modules/dbparser/groupingby.c
furiel: Sounds good to me. I pick `_load_correlation_state`.

kira-syslogng:
Build SUCCESS

modules/dbparser/groupingby.c
MrAnno: Thanks!

PR #2656, opened 2019-04-03
Expedite threaded flush at reload

bazsi:
This branch contains some of the preparatory steps for #2496 , splitting of the bulk of the branch for easier review and integration, asked by @furiel.

This contains the "expedited" flush implemented for the  http() destination, that will also be used by the upcoming kafka destination.

It also contains some dbld/makefile improvements. 

With that this becomes a dependency of #2496


kira-syslogng:
Build FAILURE

furiel:
@bazsi thank you for the split

bazsi:
I have now rebased to current master and also made an attempt to fix the compilation errors. (Some of
the filenames were incorrect at the point of the splitting of the branch).

modules/http/http-worker.c
```diff
@@ -438,6 +438,9 @@ _flush(LogThreadedDestWorker *s)
   if (self->super.batch_size == 0)
     return LTR_SUCCESS;
 
+  if (expedite)
+    return LTR_RETRY;
```
MrAnno: Do you want to support the "quick-reload" only in the http() destination first?

MrAnno: Is it true that this logic can not be generalized and moved to the threaded destination driver, because driver implementations might want to decide whether they want the `expedite` feature?

lib/logthrdest/tests/Makefile.am
```diff
@@ -0,0 +1,15 @@
+lib_logthrdest_tests_TESTS		= \
+	lib/logthrdest/tests/test_logthrdestdrv
+
+EXTRA_DIST += lib/logthrdest/tests/CMakeLists.txt
+
+check_PROGRAMS		+= ${lib_logthrdest_tests_TESTS}
+
+if ENABLE_TESTING
```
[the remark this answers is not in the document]
bazsi: Thanks, I didn't know about that. removed.

kira-syslogng:
Build FAILURE

modules/http/http-worker.c
bazsi: Re: comment#1

I implemented "expedite" because of kafka (where we might easily have 100k messages waiting to be sent), but I implemented it in HTTP first. It would make a lot of sense to do this for SQL and anything where "flush" can be a heavy operation.

Re: comment#2

On the generalization thing, it cannot easily be generalized, as drivers tend do differ how they handle in-transit messages. For HTTP, we are building a local buffer, communication is not started to the server. For SQL, we are generating INSERT statements and a flush is basically a COMMIT. For kafka, we are similar to SQL, we post the messages that is asynchronously sent to the kafka broker, and flush is basically waiting for the thread to finish processing.

Whether it makes sense to expedite a specific driver, or how that needs to be done, cannot easily be generalized. We could probably create a "drop-queued-messages-before-reload" kind of method, that would basically be the expedite=TRUE case. But on the other hand, with kafka, in case expedite is TRUE, we are waiting 1sec for the kafka worker threads to do a final batch of messages, then we purge everything that remains. This is not a complete drop, but still ensures that the reload will be timely.

Hope this clarifies things.

MrAnno: @bazsi  Thank you!

lib/logthrdest/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+set(LOGTHRDEST_HEADERS
+    logthrdest/logthrdestdrv.h
+    logthrdest/logthrfetcherdrv.h
+    PARENT_SCOPE)
+
+set(LOGTHRDEST_DESTS
```
MrAnno: Typo: `LOGTHRDEST_SOURCES`

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/logthrdest/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+set(LOGTHRDEST_HEADERS
+    logthrdest/logthrdestdrv.h
+    logthrdest/logthrfetcherdrv.h
+    PARENT_SCOPE)
+
+set(LOGTHRDEST_SOURCES
+    logthrdest/logthrdestdrv.c
+    logthrdest/logthrfetcherdrv.c
```
MrAnno: Oops, I did not see that `logthrfetcherdrv` was left here.
I've tested the CMake build on my laptop, this should be the last CMake-related failure.

lib/logthrdest/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+set(LOGTHRDEST_HEADERS
+    logthrdest/logthrdestdrv.h
+    logthrdest/logthrfetcherdrv.h
```
MrAnno: `logthrfetcherdrv`

lib/logthrdest/CMakeLists.txt
bazsi: thanks. fixed this one as well.

kira-syslogng:
Build FAILURE

lib/logthrdest/logthrdestdrv.c
```diff
@@ -676,6 +676,7 @@ static void
 _worker_thread(gpointer arg)
 {
   LogThreadedDestWorker *self = (LogThreadedDestWorker *) arg;
+  GlobalConfig *cfg = log_pipe_get_config(&self->owner->super.super.super);
```
furiel: After turning `cfg_is_shutting_down` to void, this line is not needed.

bazsi:
Hi,

On Mon, Apr 15, 2019 at 9:08 AM furiel <notifications@github.com> wrote:

> *@furiel* commented on this pull request.
>
> I do not have any blocking request on the code. The only strange thing is
> cfg_is_shutting_down, that does not have any connection with cfg, so I
> proposed that to turn it into void. But that also looks strange, because it
> would look a typical method of a class, just without self. I do not know
> how to resolve it nicely, so if you feel so, you can leave as it is.
>
> PS: http destination tests are failing (maybe related to expledite
> flush?). If the tests are fixed, this pr is ready to go in from my point of
> view.
>
> In lib/cfg.c
> <https://github.com/balabit/syslog-ng/pull/2656#discussion_r275225909>:
>
> > @@ -297,6 +297,13 @@ cfg_shutdown(GlobalConfig *cfg)
>    main_loop_exit(main_loop);
>  }
>
> +gboolean
> +cfg_is_shutting_down(GlobalConfig *cfg)
>
> This could be void function: cfg is not used by the implementation.
>
> In lib/logthrdest/logthrdestdrv.c
> <https://github.com/balabit/syslog-ng/pull/2656#discussion_r275226850>:
>
> > @@ -676,6 +676,7 @@ static void
>  _worker_thread(gpointer arg)
>  {
>    LogThreadedDestWorker *self = (LogThreadedDestWorker *) arg;
> +  GlobalConfig *cfg = log_pipe_get_config(&self->owner->super.super.super);
>
> After turning cfg_is_shutting_down to void, this line is not needed.
>
I know, it was intentional. It follows the pattern I introduced with
"cfg_shutdown()" which does not use cfg either. The only reason they don't
use it is that right now our mainloop can only use one GlobalConfig
instance at a time. In fact, we could be calling MainLoop functions
directly, however I feel it is more natural to use the already existing
coupling with GlobalConfig and leave the rest of the assumptions
centralized.

If we accept that it is a good idea to delegate shutdown to GlobalConfig,
it makes sense to also accept the "cfg" argument, which is basically the
same as "this" in C++.

This is my reasoning, if we had the "static" qualifier like in C++, I would
not add this as an explicit parameter, but in all cases where the current
MainLoop/GlobalConfig assumption breaks, that "cfg" argument will be needed.

Bazsi

pzoleex:
@kira-syslogng test this please test branch=pzolee-expedite-threaded-flush-at-reload;

kira-syslogng:
Build SUCCESS

bazsi:
I have now done the enum stuff.

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng test this please test branch=pzolee-expedite-threaded-flush-at-reload;

kira-syslogng:
Build SUCCESS

Issue #189, opened 2014-08-06
heartbeat support for AMQP


zstyblik:
Please,

either enable or make configurable heartbeat in mod AMQP. This should be configurable and/or enabled.

https://github.com/balabit/syslog-ng-3.5/blob/master/modules/afamqp/afamqp.c#L334
https://github.com/omniti-labs/Net--RabbitMQ/blob/master/amqp.h#L234

Thank you.

[Edited by @Kokan]

As `rabbitmq` already supports heartbeat, from syslog-ng side it only has to make it configurable. For details see `amqp_login` function.

ihrwein:
@dnsjts could you look at this?

furiel:
As a hint, here is how I created a working environment for development. With better understanding of rabbitmq, there is probably a simpler way. Please share if you know one!

1) Starting rabbitmq server from dockerhub.
```
$ docker run --name rabbitmq --net=host -ti rabbitmq
```
2) Adding `logs` exchange for vhost `"/"`. 
For some reason the default `""` exchange did not work with the example script from the rabbitmq tutorial, so I needed to create another one.

- Enable `webui`:
```
docker exec -ti rabbitmq rabbitmq-plugins enable rabbitmq_management
```
This will start a web service on http://localhost:15672

- Add exchange:
In the webui with guest:guest. On the Exchanges tab, fill the form
name: logs
type: fanout
durability: transient
(rest with default value)
Add exchange.

3) start syslog-ng with the configuration example below. Comment out `example-msg-generator()` if you do not want periodic messages.

```
@version: 3.20

log {
    source { stdin(flags(no-parse)); };
    source { example-msg-generator(); };
    destination { amqp(vhost("/")
                  exchange("logs")
                  body("hello world")
                  username(guest) password(guest)); };
};
```

4)
To read logs, one can use the python snippet provided in the [rabbitmq tutorial](https://github.com/rabbitmq/rabbitmq-tutorials/blob/master/python/receive_logs.py):

<details>

```
#!/usr/bin/env python
import pika

connection = pika.BlockingConnection(
    pika.ConnectionParameters(host='localhost'))
channel = connection.channel()

exchange="logs"

channel.exchange_declare(exchange=exchange, exchange_type='fanout')

result = channel.queue_declare('', exclusive=True)
queue_name = result.method.queue

channel.queue_bind(exchange=exchange, queue=queue_name)

print(' [*] Waiting for logs. To exit press CTRL+C')

[...]

PR #2660, opened 2019-04-05
amqp: fixing double connect

furiel:
During startup: connect was called two times in amqp destination, hence amqp destination built two connections towards the rabbitmq server. One connect was called from thread_init, the other one is called because of worker.connect virtual function:
https://github.com/balabit/syslog-ng/blob/5a03be7255736941a55ed67fb27c6a08c84dd275/modules/afamqp/afamqp.c#L662
and
https://github.com/balabit/syslog-ng/blob/5a03be7255736941a55ed67fb27c6a08c84dd275/modules/afamqp/afamqp.c#L744

Interestingly, this did not cause any particular problem on its own, even valgrind did not show any error. However, if someone wants to [add heartbeat to amqp](https://github.com/balabit/syslog-ng/issues/189) at some point, this would cause a little headache, as the first connection would be terminated due to timeout, as only the second was used for messaging, hence a false-positive error message is visible in the rabbitmq server logs.

### DIFF
--- a/modules/afamqp/afamqp.c
+++ b/modules/afamqp/afamqp.c
@@ -473,12 +473,12 @@ afamqp_dd_login(AMQPDestDriver *self)
 }
 
 static gboolean
-afamqp_dd_connect(AMQPDestDriver *self, gboolean reconnect)
+afamqp_dd_connect(AMQPDestDriver *self)
 {
   int sockfd_ret;
   amqp_rpc_reply_t ret;
 
-  if (reconnect && self->conn)
+  if (self->conn)
     {
       ret = amqp_get_rpc_reply(self->conn);
       if (ret.reply_type == AMQP_RESPONSE_NORMAL)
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

modules/afamqp/afamqp.c
```diff
@@ -645,7 +645,7 @@ afamqp_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
 {
   AMQPDestDriver *self = (AMQPDestDriver *)s;
 
-  if (!afamqp_dd_connect(self, TRUE))
+  if (!afamqp_dd_connect(self))
```
MrAnno: `afamqp_dd_connect()` has a hidden connection check and disconnect/reconnect logic, which bypasses our `time-reopen()` functionality.

This `connect()` call could be removed from `insert()`, and the connection probe could be extracted from `insert()` as well if we decreased the default value of `time-reopen` at least for threaded destinations.

This is not the scope of this PR, so :+1: .

PR #2664, opened 2019-04-08
patterndb: Add sort option to pdbtool merge

### LINKED ISSUE #294: add filename based sorting to pdbtool merge in a similar vein to what run-parts does
https://bugzilla.balabit.com/show_bug.cgi?id=211

This issue is still open and IMHO very important.
The use case I have is when merging a `pam_unix` ruleset which matches `PROGRAM==sshd` and an `openssh` ruleset matching `PROGRAM==sshd` there are unpredictable results unless both rulesets are in the same file.

alltilla:
Adds sort option to `pdbtool merge`:
`  -s, --sort                      Sort files during merge (alphabetic order, with files first, directories after)
`

By default the option is not set, and the file sequence is not guaranteed to be sorted.

Fixes #294.

### DIFF
--- a/modules/dbparser/pdb-file.c
+++ b/modules/dbparser/pdb-file.c
@@ -165,3 +165,84 @@ pdb_file_validate_in_tests(const gchar *filename, GError **error)
 {
   return _pdb_file_validate(filename, error, _get_xsddir_in_build);
 }
+
+GPtrArray *
+pdb_get_filenames(const gchar *dir_path, gboolean recursive, gchar *pattern, GError **error)
+{
+  GDir *dir;
+
+  if ((dir = g_dir_open(dir_path, 0, error)) == NULL)
+    return NULL;
+
+  GPtrArray *filenames = g_ptr_array_new_with_free_func(g_free);
+  const gchar *path;
+
+  while ((path = g_dir_read_name(dir)) != NULL)
+    {
+      gchar *full_path = g_build_filename(dir_path, path, NULL);
+      if (recursive && is_file_directory(full_path))
+        {
+          GPtrArray *recursive_filenames = pdb_get_filenames(full_path, recursive, pattern, error);
+          if (recursive_filenames == NULL)
+            {
+              g_ptr_array_free(recursive_filenames, TRUE);
+              g_ptr_array_free(filenames, TRUE);
+              g_free(full_path);
+              g_dir_close(dir);
+              return NULL;
+            }
+          for (guint i = 0; i < recursive_filenames->len; ++i)
+            g_ptr_array_add(filenames, g_ptr_array_index(recursive_filenames, i));
+          g_ptr_array_free(recursive_filenames, FALSE);
+          g_free(full_path);
+        }
+      else if (is_file_regular(full_path) && (!pattern || g_pattern_match_simple(pattern, full_path)))
+        {
+          g_ptr_array_add(filenames, full_path);
+        }
+      else
+        {
+          g_free(full_path);
+        }
+    }
+  g_dir_close(dir);
+
+  return filenames;
+}
+
+static guint
+pdbtool_get_path_depth(const gchar *path)
+{
+  gint depth = 0;
+
+  while (*path)
+    {
+      if (*path++ == '/')
+        ++depth;
+    }
+
+  return depth;
+}
+
+static int
+pdbtool_path_compare(gconstpointer a, gconstpointer b)
+{
+  const gchar *path_a = *(const gchar **)a;
+  const gchar *path_b = *(const gchar **)b;
+
+  guint a_depth = pdbtool_get_path_depth(path_a);
+  guint b_depth = pdbtool_get_path_depth(path_b);
+
+  if (a_depth > b_depth)
+    return 1;
+  else if (a_depth < b_depth)
+    return -1;
+  else
+    return strcmp(path_a, path_b);
+}
+
+void
+pdb_sort_filenames(GPtrArray *filenames)
+{
+  g_ptr_array_sort(filenames, pdbtool_path_compare);
+}
--- a/modules/dbparser/pdb-file.h
+++ b/modules/dbparser/pdb-file.h
@@ -29,5 +29,7 @@
 gint pdb_file_detect_version(const gchar *pdbfile, GError **error);
 gboolean pdb_file_validate(const gchar *filename, GError **error);
 gboolean pdb_file_validate_in_tests(const gchar *filename, GError **error);
+GPtrArray *pdb_get_filenames(const gchar *dir_path, gboolean recursive, gchar *pattern, GError **error);
+void pdb_sort_filenames(GPtrArray *filenames);
 
 #endif
--- a/modules/dbparser/pdbtool/pdbtool.c
+++ b/modules/dbparser/pdbtool/pdbtool.c
@@ -80,6 +80,7 @@ static const gchar **colors = empty_colors;
 
 static const gchar *patterndb_file = PATH_PATTERNDB_FILE;
 static gboolean color_out = FALSE;
+static gboolean sort = FALSE;
 
 typedef struct _PdbToolMergeState
 {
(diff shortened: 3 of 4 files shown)


kira-syslogng:
Build SUCCESS

lib/pathutils.c
```diff
@@ -26,6 +26,8 @@
 #include "pathutils.h"
 #include <sys/stat.h>
 #include <string.h>
+#include <stdio.h>
```
furiel: Maybe `msg_error` could be used instead of `fprintf`.

lib/pathutils.c
```diff
@@ -92,3 +94,108 @@ get_filename_extension(const gchar *filename)
...
+_add_path_to_filenames(gchar *path, gchar **filenames, size_t *pos, size_t *array_len)
+{
+  if (*pos < *array_len)
+    {
+      filenames[*pos] = path;
+      (*pos)++;
+    }
+  else
+    {
+      *array_len *= 2;
```
furiel: Is it possible to use some glib data structure? Like `GPtrArray`? Then you would not need to implement such logics.

lib/tests/test_pathutils.c
```diff
@@ -88,3 +88,89 @@ Test(test_pathutils, test_get_filename_extension)
...
+void
+test_get_filenames_teardown(void)
+{
+  const gchar *dirs[] =
+    {
+      "pathutils_get_filenames",
+      "pathutils_get_filenames/testdir",
+      "pathutils_get_filenames/testdir2"
+    };
+  const gchar *files[] =
```
furiel: I would change these to global variables, so to avoid duplication in setup.

lib/tests/test_pathutils.c
```diff
@@ -88,3 +88,89 @@ Test(test_pathutils, test_get_filename_extension)
...
+      "pathutils_get_filenames/test.file",
+      "pathutils_get_filenames/test2.file",
+      "pathutils_get_filenames/testdir/test.file",
+      "pathutils_get_filenames/testdir2/test23.file",
+      "pathutils_get_filenames/testdir2/test22.file"
+    };
+  size_t dirs_len = sizeof(dirs) / sizeof(gchar *);
+  size_t files_len = sizeof(files) / sizeof(gchar *);
+
+  for (gint i = 0; i < files_len; ++i)
```
furiel: [g_remove](https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-remove) theoretically removes both files and directories.

lib/pathutils.c
alltilla: Thanks, fixed.

lib/pathutils.c
alltilla: Thanks, it helped a lot! :)

lib/tests/test_pathutils.c
alltilla: Thanks, fixed :)

lib/tests/test_pathutils.c
alltilla: Thanks, fixed :)

kira-syslogng:
Build SUCCESS

lib/tests/test_pathutils.c
furiel: I was thinking this could be used to recursively delete files, but it seems it is not the case. Sorry for the extra round, the original version was fine.

lib/tests/test_pathutils.c
```diff
@@ -88,3 +106,47 @@ Test(test_pathutils, test_get_filename_extension)
...
+
+Test(test_pathutils, test_get_sorted_filenames, .init = test_get_filenames_setup, .fini = test_get_filenames_teardown)
+{
+  const gchar *expected[] =
+  {
+    "pathutils_get_filenames/test2.file",
+    "pathutils_get_filenames/testdir2/test22.file",
+    "pathutils_get_filenames/testdir2/test23.file"
+  };
+  guint expected_len = sizeof(expected) / sizeof(gchar *);
```
furiel: G_N_ELEMENTS could be used.

alltilla: Thanks, fixed :) :+1:

kira-syslogng:
Build SUCCESS

lib/pathutils.c
```diff
@@ -92,3 +93,87 @@ get_filename_extension(const gchar *filename)
 
   return start_ext + 1;
 }
+
+GPtrArray *
+get_filenames(const gchar *dir_path, gboolean recursive, gchar *pattern)
```
Kokan: Should the `libsyslog-ng` be extended with those ?
This pattern can be found in plugin listing, wildcard file monitor, config include. Moving this into a common place I would expect to do code reuse; at least one reason could be. In this case I would not do that.

But rather keep it in the scope of `pdbtool`, which I would prefer.

lib/tests/test_pathutils.c
```diff
@@ -88,3 +106,47 @@ Test(test_pathutils, test_get_filename_extension)
...
+void
+test_get_filenames_setup(void)
+{
+  for (gint i = 0; i < dirs_len; ++i)
+    cr_assert_not(mkdir(dirs[i], S_IRUSR | S_IWUSR | S_IXUSR) < 0, "mkdir error %s", dirs[i]);
+
+  for (gint i = 0; i < files_len; ++i)
+    {
+      int fd = open(files[i], O_CREAT | O_RDWR, 0644);
+      cr_assert_not(fd < 0, "open error %s", files[i]);
```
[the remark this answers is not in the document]
alltilla: Thanks for the heads up, I did not check that. :)

After a digging a bit, I believe we can assert there.

According to https://criterion.readthedocs.io/en/master/starter.html
"If a setup crashes, you will get a warning message, and the test will be aborted and marked as a failure. Is a teardown crashes, you will get a warning message, and the test will keep its result."

I added `cr_assert(FALSE);` as the first line of my setup function, and got the following:
```
[14:31] ~/Work/build/OSE $ lib/tests/test_pathutils
[----] ../../repos/OSE/lib/tests/test_pathutils.c:113: Assertion failed: The expression (0) is false.
[----] Warning! The test `test_pathutils::test_get_sorted_filenames` crashed during its setup or teardown.
[====] Synthesis: Tested: 6 | Passing: 5 | Failing: 1 | Crashing: 1 
```

With these, I believe asserting is allowed in setup and teardown. What do you think?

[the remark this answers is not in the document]
MrAnno: A crash is different than an assertion. AFAIK assertions inside setup and teardown are not documented. You can ask Snaipe to document the behavior, but until then, I would not use assert in setup().

lib/pathutils.c
alltilla: I agree, this will be moved.

lib/tests/test_pathutils.c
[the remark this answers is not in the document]
MrAnno: Well, adding an assertion to `setup()` resulted in a SIGSEGV in a longjmp call inside the forked test case, which is something unexpected IMHO (it might be a bug in Criterion, or just ASLR again).

I'll investigate this. `g_assert()` is safer to use here in my opinion.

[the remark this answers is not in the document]
Kokan: Does asserting if the mkdir, or create file was successful? I mean with that you are testing if mkdir or open works.
It is possible disk has issue, but debugging it is going to be nasty either way.
There is always the alternative, to create those files and store them in git, you can be sure they exists and no setup or assertion is needed.
This is not ideal either, as the paint point is that the tested code cannot function without disk.

Imho if you have to assert in setup phase, that's probably issue in your test strategy, as one test should test one thing.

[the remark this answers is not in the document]
Kokan: Does asserting if the mkdir, or create file was successful? I mean with that you are testing if mkdir or open works.
It is possible disk has issue, but debugging it is going to be nasty either way.
There is always the alternative, to create those files and store them in git, you can be sure they exists and no setup or assertion is needed.
This is not ideal either, as the paint point is that the tested code cannot function without disk.

Imho if you have to assert in setup phase, that's probably issue in your test strategy, as one test should test one thing.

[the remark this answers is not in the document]
furiel: Personally I'm on the assert side, because I think it helps the debugging if we report the problem at the point of origin. Can we discuss this IRL?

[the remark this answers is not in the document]
Kokan: sure.

lib/pathutils.c
alltilla: I moved it :)

lib/tests/test_pathutils.c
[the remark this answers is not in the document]
alltilla: I changed it to `g_assert()`.

alltilla:
@Kokan, @furiel, @MrAnno:
I moved the function to `pdb-file.c`. Also updated the asserts in the setup/teardown to `g_assert()`.
I fixed a missing error check in the recursive call, please check that also. Thanks!

alltilla:
+There is no `_sorted_` function now, instead I made a sort function. I think its cleaner to use on the call site.

kira-syslogng:
Build SUCCESS

lib/tests/test_pathutils.c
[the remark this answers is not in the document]
MrAnno: Thanks.

kira-syslogng:
Build SUCCESS

lib/tests/test_pathutils.c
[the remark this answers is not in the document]
MrAnno: In general, I agree with @Kokan. We should not assert in setup and teardown functions. They should contain initialization code that is not part of the actual test case, containing unconditional truths. If the setup/teardown logic contains something that can fail, a separate test case should be added to test that logic, and every other test case should assume that the initialization logic is correct and won't fail anymore.

This is true for unit tests, where ideally all unpredictable resources are mocked out.
For example, you could have mocked `g_dir_read_name` and other disk operations out from `pdb_get_filenames`, so you could write a unit test for the code you wrote, not for the code you wrote plus glib's disk functions and all the other unpredictable factors that a file system operation might cause.

JUnit has methods for "stating assumptions about the conditions in which a test is meaningful".
https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html
This would be perfect for your situation.
You test case is about 2 things: testing `pdb_get_filenames` and `pdb_sort_filenames`.
The test about `pdb_sort_filenames` is "not meaningful" if you can't fetch its input, but failing to fetch filenames does not mean that `pdb_sort_filenames` is broken.

I'll keep all this in mind when rewriting our worst-looking test cases.

In case you write a functional/component test, you might want to assert initialization code as well, but I would not do that in setup() and teardown(), especially because Criterion can only handle a _crash_ in those functions, not assertion failures (by design IMHO).

----

I don't want to request a change on this PR, I just wanted to share how I would like to write tests from now on. The JUnit guys have really good principles/paradigms for writing clean and beautiful tests, a few might apply to our low-level environment as well in my opinion.

PR #2666, opened 2019-04-09
nondumpable-allocator: fixing mmap error handling

furiel:
mmap returns -1 instead of NULL in case of error, because in
some (extreme) cases, NULL can be start of a mmapped area. Aligning the
code according to this.

Users of nondumpable-allocator need not be changed. Nondumpable
allocator puts metadata to the front of a page, and returns the memory
area after the metadata. Hence, NULL is never returned after a
successful allocation, so NULL can still be used for error handling.

### DIFF
--- a/lib/secret-storage/nondumpable-allocator.c
+++ b/lib/secret-storage/nondumpable-allocator.c
@@ -94,11 +94,11 @@ static gpointer
 _mmap(gsize len)
 {
   gpointer area = mmap(NULL, len, PROT_READ|PROT_WRITE, MAP_ANONYMOUS|MAP_PRIVATE, -1, 0);
-  if (!area)
+  if (area == MAP_FAILED)
     {
       logger_fatal("secret storage: cannot mmap buffer",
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

lib/secret-storage/nondumpable-allocator.c
```diff
@@ -94,11 +94,11 @@ static gpointer
 _mmap(gsize len)
 {
   gpointer area = mmap(NULL, len, PROT_READ|PROT_WRITE, MAP_ANONYMOUS|MAP_PRIVATE, -1, 0);
-  if (!area)
+  if (area == (void *)-1)
```
MrAnno: man says the return value on error is `MAP_FAILED` from `<sys/mman.h>`.
It is (void *)-1, but this is a implementation detail. Please use `MAP_FAILED` if possible.

kira-syslogng:
Build FAILURE

MrAnno:

kira-syslogng:
Build SUCCESS

PR #2669, opened 2019-04-10
Fix unit test with function pointer dereference in case of ASLR, Criterion

Kokan:
The `test_diskq_statistics` could crash under certain conditions.
If the linux ASLR is enabled, and the compiler creates executable
that is position independent(`-fPIC`).

The function pointer could became invalid, in the unit test.
As the Criterion does *fork* and *exec* before executing tests.

Additionally the parameters are created in the *parent*,
and passed down to the *child* process, which - because of ASLR -
possible stores the referenced function in a different address.

This patch does not solves the above issue, but hides this behaviour
with additionally calling the function pointer passed by
via function pointer.

Because of this, a `@PLT` trampoline function is created, which
address is used, that can be made position independent hence
valid after *fork* plus *exec* if compiled with `-fno-pie`,
which is done in our case.

There is a nice post about related topics: https://www.macieira.org/blog/2012/01/sorry-state-of-dynamic-libraries-on-linux/

### DIFF
--- a/modules/diskq/tests/test_diskq.c
+++ b/modules/diskq/tests/test_diskq.c
@@ -309,17 +309,25 @@ Test(diskq, testcase_with_threads)
 typedef struct restart_test_parameters
 {
   gchar *filename;
-  LogQueue *(*lqdisk_new_func)(DiskQueueOptions *options, const gchar *persist_name);
   gboolean reliable;
 } restart_test_parameters;
+
+static LogQueue *
+queue_new(gboolean reliable, DiskQueueOptions *options, const gchar *persist_name)
+{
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build FAILURE

Kokan:

Kokan:
@furiel sorry I've kept it as WIP as I want to write a proper description for both github PR gui and git commit, as that commit solves an actual issue, which is not communicated (the fixing) and why.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Issue #904, opened 2016-02-05
Numerical usage of yesno, validation, documentation


bkil-syslogng:
Inheritance fixes started my thought process about the following. From the grammar, I could see that you can specify yes/no values either via tokens or via a number (usually 0 or 1, but see below).
- We should document such numerical usage.
- What about hidden options like negative numbers? (I first thought that this Boolean was actually a Maybe Boolean, however I started to get sad after seeing -2 and beyond in the code to suppress inheritance in cases or something like that.)
- We should check that the given integer is in the accepted range.

bazsi:
Hi,

I dont see the value this would add to the product. The reason it supports
both numbers and yes/no tokens is that at one point in the past 18 years we
were using it inconsistently, ans it was unified this way.

Adding additional tokens is not that simple either, as the current solution
is using keywords, adding additional keywords can introduce grammar
collisions. Migrating this over to a non-keyword solution could work
though, but that's a bit more involved than merely adding a few additional
tokens.

I feel that there are more important architectural issues than validating
extremal values on the grammar <> implementation interface (like decreasing
the coupling between GlobalConfig and almost everything else).

So if this doesn't convince you, this ticket should be split, one about the
functional change, the other about validation.

bkil-syslogng:
I'm updated the description and split off #906

gaborznagy:
Some thoughts:

> - We should document such numerical usage.

Well, after dropping aliases for yesno (https://github.com/balabit/syslog-ng/pull/2719), I don't think we would like to document it.

> - What about hidden options like negative numbers? (I first thought that this Boolean was actually a Maybe Boolean, however I started to get sad after seeing -2 and beyond in the code to suppress inheritance in cases or something like that.)

With a quick grep, I have not found inheritance bypasses within version 3.22.

> - We should check that the given integer is in the accepted range.

If we don't want to support integer values in place of yesno, then there is nothing to validate.

With these in mind, I would rather close this ticket.

PR #2671, opened 2019-04-12
cfg-parser: add aliases for yesno

### LINKED ISSUE #906: Additional aliases for yesno
Consider to add some aliases for `yes`/`no` in configuration settings, for example:
- `true`/`false`,
- `enable`/`disable`,
- `enabled`/`disabled`,
- `on`/`off`.

Although the syntax is correctly documented in the guide, this terminology surprised me when I first used `syslog-ng`. Of course `yes`/`no` is also used in config files of other software as well, many (if not all) tools that I've used did provide a subset of the above aliases. Just to let you stay in touch with the real world and to keep in mind the principle of least surprise.

I had this idea after seeing undocumented numerical aliases in #904.

alltilla:
New aliases:
 * true/false
 * enable/disable
 * enabled/disabled

Fixes #906.

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -1029,9 +1029,17 @@ string
 	;
 
 yesno
-	: KW_YES				{ $$ = 1; }
-	| KW_NO					{ $$ = 0; }
-	| LL_NUMBER				{ $$ = $1; }
+  : LL_IDENTIFIER
+    {
+      gboolean value;
+      gboolean success = cfg_process_yesno($1, &value);
+      CHECK_ERROR((TRUE == success), @1, "invalid yesno value: %s. Use one of the following: yes/no, on/off, true/false, enable/disable, enabled/disabled", $1);
+      $$ = value;
+    }
+  | LL_NUMBER
+    {
+      $$ = $1;
+    }
 	;
 
 dnsmode
--- a/lib/cfg-parser.c
+++ b/lib/cfg-parser.c
@@ -168,11 +168,6 @@ static CfgLexerKeyword main_keywords[] =
   { "type",               KW_TYPE },
   { "tags",               KW_TAGS },
 
-  /* on/off switches */
-  { "yes",                KW_YES },
-  { "on",                 KW_YES },
-  { "no",                 KW_NO },
-  { "off",                KW_NO },
   { NULL, 0 }
 };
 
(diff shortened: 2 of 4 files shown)


kira-syslogng:
Build SUCCESS

alltilla:
@szemere: Good point, although the fix is easy, I am also not sure if we want this. 

I say, that this change does not remove any previous functionality, and just one sentence in the admin guide could list all the aliases, so this is not a big pain IMO. I can accept closing this PR and #906, though, if we do not want this change.

furiel:
The patch looks good to me! My only concern is an old [comment](https://github.com/balabit/syslog-ng/issues/904#issuecomment-180260148) from @bazsi about keyword collision. I don't entirely understand the comment, nor know it is still relevant, but it would worth a quick sync about it.

alltilla:
Let's discuss this IRL next week, than :) Thanks for the comments!

bazsi:
If we want to support these tokens, they should not be added via keywords. Rather the "yes" "no" rules
should accept an LL_IDENTIFIER and an explicit lookup be used for the potential aliases.

This way if someone has a destination called "enabled" will not cause keyword collision.

I have seen these in the past, the one I remember vividly is a keyword called "null" and I think SUSE using this name as a destination in their default configuration.

alltilla:
Thanks @bazsi, I will do that! :)

kira-syslogng:
Build FAILURE

lib/cfg-parser.c
```diff
@@ -452,10 +447,52 @@ cfg_process_flag(CfgFlagHandler *handlers, gpointer base, const gchar *flag_)
   return FALSE;
 }
 
+gboolean
+cfg_validate_yesno(const gchar *yesno, gboolean *value)
```
Kokan: Have you considered merging the `cfg_validate_yesno` and `cfg_process_yesno` ?

lib/cfg-grammar.y
```diff
@@ -1029,9 +1029,17 @@ string
...
 
 yesno
-	: KW_YES				{ $$ = 1; }
-	| KW_NO					{ $$ = 0; }
-	| LL_NUMBER				{ $$ = $1; }
+  : LL_IDENTIFIER
+    {
+      gboolean value;
+      gboolean success = cfg_validate_yesno($1, &value);
+      CHECK_ERROR((TRUE == success), @1, "Invalid yesno value: %s", $1);
```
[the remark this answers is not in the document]
alltilla: Good idea, I will add this. Thanks!

kira-syslogng:
Build SUCCESS

lib/cfg-parser.c
alltilla: Yep, I just did that :P

kira-syslogng:
Build SUCCESS

lbudai:
Hmmm... with this patch we just broke all the configs that uses `peer-verify(yes/no)`.
Unfortunately we don't have any tests for this case (not even in the internal testdb...).

The problem is :
`KW_PEER_VERIFY '(' yesno ')' `
`KW_PEER_VERIFY '(' string ')'` ...

and as `yesno` is an `LL_IDENTIFIER`, we got the follwing message from Bison (this was the trigger for me, I've just modified the grammar, and got this error.. but I was sure that I did not make any mistake, so I checked master and found this patch)
```
modules/afsocket/afsocket-grammar.y: warning: 1 reduce/reduce conflict [-Wconflicts-rr]
```


Maybe we should also check bison warning...

I'm going to revert this patch (it's just an old, internally reported, nice-to-have improvement, so we can ship these changes in the next release).

alltilla:
Sure thing, thanks for checking, and sorry for introducing such a bug.
I will make it work in the next release.

PR #2672, opened 2019-04-12
syslogformat: set $MSG even if the incoming message is empty

bazsi:
With RFC5424, the $MSG payload is optional for the message. In that
case the syslog-parser() bailed out without explicitly setting $MSG to
an empty string.

This works if this is the initial parsing of the message (e.g. right
within the source driver), but doesn't if we are using an explicit
syslog-parser() later on. In that case the $MSG will remain the original
value, potentially the entire raw message, if it was received via
flags(no-parse).


kira-syslogng:
Build SUCCESS

PR #2678, opened 2019-04-14
WIP: undefined warnings fixes

### LINKED ISSUE #2677: warning elimination: undefined

This issue should cover the undefined warning. I am splitting issue from the ticket #1965

     12 [-Wundef]

mehul-m-prajapati:
fixes #2677

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,7 +331,8 @@ if (ENABLE_EXTRA_WARNINGS)
     -Wpointer-arith
     -Wpointer-sign
     -Wmissing-format-attribute
-    -Wsuggest-attribute=noreturn)
+    -Wsuggest-attribute=noreturn
+    -Wundef)
 endif()
 
 add_compile_options(${IMPORTANT_WARNINGS} ${ACCEPTABLE_WARNINGS} ${EXTRA_WARNINGS})
--- a/Makefile.am
+++ b/Makefile.am
@@ -84,7 +84,8 @@ AM_CFLAGS += \
 	-Wpointer-arith \
 	-Wpointer-sign \
 	-Wmissing-format-attribute \
-	-Wsuggest-attribute=noreturn
+	-Wsuggest-attribute=noreturn \
+	-Wundef
 endif
 
 TEST_CFLAGS		= -I$(top_srcdir)/libtest -DTOP_SRCDIR=\"$(abs_top_srcdir)\" $(AM_CFLAGS) $(CRITERION_CFLAGS)
--- a/lib/compat/pio.c
+++ b/lib/compat/pio.c
@@ -24,7 +24,7 @@
 
 #include "compat/pio.h"
 
-#if !SYSLOG_NG_HAVE_PREAD || SYSLOG_NG_HAVE_BROKEN_PREAD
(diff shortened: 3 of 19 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

furiel:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

mehul-m-prajapati:
@furiel : I am not sure, how to fix these warnings for cmake. Can you guide me on this?

furiel:
In general, it would be best if you could avoid any changes from explicit `#if` to `defined`. These are not equivalent: for example if I set `-DVARIABLE=0`, `defined` would be true, but `#if` would be false. Any such identifier that comes from syslog-ng would be best if handled as explicit value, and not `defined`. You can derive from it, but then please add an explanation about the reason.
An individual identifier might fail for various reasons. For example there is a trick in autotools that we typically forget:
`AX_PREFIX_CONFIG_H(syslog-ng-config.h, "SYSLOG_NG")` adds `SYSLOG_NG_` prefix to every identifiers automatically. So for example `HAVE_GETOPT_H` is mistyped, and `SYSLOG_NG_HAVE_GETOPT_H` would probably work. 
One special case: I did not find much about `SYSLOG_NG_HAVE_BROKEN_PREAD`, only in an older version of syslog-ng. This was introduced due to a compile error in HP-UX. You can define this for zero globally for now, and we add proper detection later.
If you are stuck with any specific identifiers, feel free to ask for help. There might be other nontrivial ones.

Identifiers from compiler can be changed to `defined`. Specifically, `defined (_AIX)` or `defined (__FreeBSD__)` is reasonable.

About cmake: these are the methods that we use to add identifiers to code (that I know of): [configure_file](https://cmake.org/cmake/help/v3.2/command/configure_file.html?highlight=cmakedefine), the other one is [target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html). Please check these documentations, if they help you.
In case of cmake: it is possible that an identifier is missing because the a detection itself missing for it. In that case it would be best if you could write the detections too, or at least collect those in a separate issue, to avoid missing them later. In such cases, you can define those identifiers as zeroes.

mehul-m-prajapati:
@furiel : Thanks for the detailed explanation. I will look into it and make the changes accrodingly.

pzoleex:

pzoleex:
@mehul-m-prajapati 
From now you can trigger the test system via
"@kira-syslogng test this please;" command. You don't need to close/reopen or force push the pr (at least this is the theory, please try it)

kira-syslogng:
Build SUCCESS

mehul-m-prajapati:
@pzoleex : Thanks

furiel:
@mehul-m-prajapati I created another PR (https://github.com/balabit/syslog-ng/pull/2806) based on your work. Thank you for your effort, I carried most parts with your signoff. I am closing this in favor or #2806.

PR #2683, opened 2019-04-16
filter: Fix "!=" filter

alltilla:
The `KW_NUM_NE` and `KW_NE` keywords did not set any comparison flag for `fop_cmp_eval()`. It was handled by checking for no flag set (`self->cmp_op == 0`), which is not true in case of `KW_NUM_NE`, because with `KW_NUM_NE` the `FCMP_NUM` flag is set.

After this the `|| self->cmp_op == 0` check is not needed, but we have to make a boolean from `self->cmp_op & FCMP_LT`, which was passively done by having `|| self->cmp_op == 0` after it. I cleaned it up with double negation.

Sample config for reproduction:
```
@version: 3.20
@include "scl.conf"

filter f_even { "$(% ${R_SEC} 2)" == "0" };
filter f_odd { "$(% ${R_SEC} 2)" != "0" };

destination d_local {
	file("/tmp/ne_test.log");
};

log {
	filter(f_even);
	destination(d_local);
};

log {
	filter(f_odd);
	destination(d_local);
};

```
Erroneous logs:
```
[2019-04-16T13:36:55.776895] cmp() evaluation started; left='1', operator='!=', right='0', msg='0x22242a0'
[2019-04-16T13:36:55.776967] <<<<<< filter rule evaluation result; result='UNMATCHED - Dropping message from LogPipe', rule='f_odd', location='/home/alltilla/install/etc/syslog-ng.conf:5:15', msg='0x22242a0'
```

### DIFF
--- a/lib/filter/filter-cmp.c
+++ b/lib/filter/filter-cmp.c
@@ -78,11 +78,11 @@ fop_cmp_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg)
     }
   else if (cmp < 0)
     {
-      result = self->cmp_op & FCMP_LT || self->cmp_op == 0;
+      result = !!(self->cmp_op & FCMP_LT);
     }
   else
     {
-      result = self->cmp_op & FCMP_GT || self->cmp_op == 0;
+      result = !!(self->cmp_op & FCMP_GT);
     }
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

PR #2684, opened 2019-04-17
junos: add application adapter for Junos classification

bazsi:

### DIFF
--- a/scl/Makefile.am
+++ b/scl/Makefile.am
@@ -10,6 +10,7 @@ SCL_SUBDIRS	= \
 	graylog2	\
 	hdfs		\
 	iptables	\
+	junos		\
 	kafka		\
 	linux-audit	\
 	loadbalancer	\
--- a/scl/junos/plugin.conf
+++ b/scl/junos/plugin.conf
@@ -1,0 +1,28 @@
+#############################################################################
+# Copyright (c) 2019 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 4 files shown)


kira-syslogng:
Build SUCCESS

scl/Makefile.am
```diff
@@ -10,6 +10,7 @@ SCL_SUBDIRS	= \
 	graylog2	\
 	hdfs		\
 	iptables	\
+	junos		\
```
furiel: Please add the scl to cmake.

furiel: Please add the scl to debian packaging.

bazsi: just did that in parallel :)

scl/junos/plugin.conf
```diff
@@ -0,0 +1,28 @@
...
+
+application junos[syslog] {
+	filter { "${.SDATA.junos@2636.1.1.1.2.129.reason}" ne "" };
```
furiel: What is the purpose of this addition? 
We usually add something to the messages other than filtering. But in this case we accept the original parsing. Is it here to increase performance so that the app parser would quit earlier?

kira-syslogng:
Build SUCCESS

scl/junos/plugin.conf
bazsi: It turned out that it is very valuable to not just parse but also to identify logs. Splunk in most cases only needs this classification information and completely ignores parsing.

So I am actually thinking about doing an `app-classifier()` that will only use the filter portions of adapters and associate a `${.app.name}` value without doing the in-depth parsing, this can later be used to associate a Splunk `sourcetype`.

In a lot of scenarios syslog-ng filters are used to do this identification, but instead of doing these one-by-one, I'd like to provide a coherent framework and some samples. This can also be something that can easily be contributed back to the project.

furiel: Oh ok I forgot about that app parser sets the `.app.name` automatically. All makes sense now. Thanks for explaining.

PR #2685, opened 2019-04-17
system-source: remove unused app-parser

gaborznagy:
App-parsers usage, including using the "system-unix" topic
introduced in 286e1a64b001f12e6be2b4bf5cccab4b21c2081b.

I think removing an unused code is a good step,
especially that it could cause problems when the first system-unix topic
would be introduced.

Reason:
- app-parser creates a channel, which groups some junctions-channels
- 2 app-parsers following each other therefore means 2 channels, i.e. a log pipe sequence.
If a drop happens in the first part, then it will never reach the second part, i.e. the 2nd app-parser topic.

Conclusion is that we either remove the unused app-parser,
or fix it with an additional junction-channel/if-else.

### DIFF
--- a/modules/appmodel/app-parser-generator.c
+++ b/modules/appmodel/app-parser-generator.c
@@ -132,6 +132,12 @@ _generate_applications(AppParserGenerator *self, AppModelContext *appmodel)
 
 
 static void
+_generate_empty_frame(AppParserGenerator *self)
+{
+  g_string_append(self->block, "\nchannel { filter { tags('.app.doesnotexist'); }; flags(final); };");
+}
+
+static void
 _generate_framing(AppParserGenerator *self, AppModelContext *appmodel)
 {
   g_string_append(self->block,
@@ -139,16 +145,11 @@ _generate_framing(AppParserGenerator *self, AppModelContext *appmodel)
                   "    junction {\n");
 
   _generate_applications(self, appmodel);
-
+  _generate_empty_frame(self);
   g_string_append(self->block, "    };\n");
   g_string_append(self->block, "}");
 }
 
-static void
-_generate_empty_frame(AppParserGenerator *self)
-{
-  g_string_append(self->block, "\nchannel {}");
-}
 
 static gboolean
 _parse_auto_parse_arg(AppParserGenerator *self, CfgArgs *args, const gchar *reference)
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
It is true that we don't have our SCLs for this hook at the moment, but the whole point is to allow user extensions at this point.

I am not naive, so this is probably unused at this point, but still I think being able to classify/parse logs in the knowledge the it was originally a journal originated event makes sense.

With all that said, it makes sense to change behaviour, so that stuff not matched by systemd-unix is not dropped, like with default-network-drivers () but I wouldn't remove it

gaborznagy:
@bazsi, thanks for the clarification, but as I described in some cases it won't work for the users, when they add the first app-parser.

I've mentioned an alternative for the removal: we need if/elif branches, otherwise the 2nd app-parser will not receive messages in some cases. But it would need some testing as well, which is missing currently.

kira-syslogng:
Build SUCCESS

gaborznagy:
I forgot to mention that I would still keep this PR open as I still vote for removal.

bazsi:
hmm... this actually revealed a larger problem. what should app-parser() do if no parsers are registered at the specific topic? Right now it matches and lets the message through.

However, that's not really consistent to what it does when there are adapters registered. In this case, app-parser() drops the message if NONE of the parsers match. And this makes sense, as this is what allows us to nicely combine app-parsers() in the way default-network-drivers() does.

So I argue, that app-parser() should drop the message if there's no adapters and the surrounding configuration should handle this by still allowing the message to pass. This is done in default-network-drivers() and I also created a similar patch to system() source. (https://github.com/balabit/syslog-ng/commit/0f8b03bd146ac27d811ea0c5b497951e7b419343)

what do you think?

gaborznagy:
> when there are adapters registered. In this case, app-parser() drops the message if NONE of the parsers match. 
Sorry if I don't find it, but this is provided by the if-elif-else statements, and not by app-parser, right?
App parser creates junction-channel elements, those are not dropping without a final flag.

furiel:
My thoughts:
- this is a not documented feature of system source, so users might not know about this feature, nor know how to use it.
- system-unix is an empty topic. That means we do not know any use case that could be covered with this feature. That suggests users would not have either.
- It is not tested, so we cannot ensure quality here, although it might affect system() source that is intended to use for production.

So I am on the removal side too.
If document this feature, add at least one parser to this topic, and cover with an end to end test, then I would be fine if it remains.

gaborznagy:
After our team discussion we agreed that `system-unix` topic will be dropped, but we will change the default behaviour of app-parser, so "app-parser() drops the message if no adapters are registered".

I've picked the corresponding commit from #2698.

kira-syslogng:
Build FAILURE

gaborznagy:

kira-syslogng:
Build SUCCESS

PR #2686, opened 2019-04-18
light: test app parser applications

furiel:
This patchset tries to add test coverage for applications included into the supported app parser framework.

These are not e2e tests in the sense that we still work with hardcoded example messages. Despite that, the changeset might still have some value: adds a little guarantee that we would not break these functionalities, at least on these messages.

Bonus: light now supports app parser and syslog parser, and one can specify include files in the configuration.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/applications/test_applications_with_app_parser.py
```diff
@@ -0,0 +1,64 @@
...
+    ("syslog-raw", "<159>Dec 19 10:48:57 EST 10.203.28.21 vendor=Websense product=Security product_version=7.7.0 action=permitted severity=1 category=153 user=- src_host=10.64.134.74 src_port=62189 dst_host=mail.google.com dst_ip=74.125.224.53 dst_port=443 bytes_out=197 bytes_in=76 http_response=200 http_method=CONNECT http_content_type=- http_user_agent=Mozilla/5.0_(Windows;_U;_Windows_NT_6.1;_enUS;_rv:1.9.2.23)_Gecko/20110920_Firefox/3.6.23 http_proxy_status_code=200 reason=- disposition=1034 policy=- role=8 duration=0 url=https://mail.google.com", "${.app.name}", "websense")
+]
```
mitzkia: +1 For the parametrized tests. Some small change request to here. If possible I would not create a new variable like: `test_parameters_raw`, I would add the needed parameters inside the "@pytest.mark.parametrize()", and for each test parameter I would define separate id. 
For more information see: https://docs.pytest.org/en/latest/example/parametrize.html

tests/python_functional/functional_tests/applications/test_applications_with_app_parser.py
```diff
@@ -0,0 +1,64 @@
...
+    config.add_include("scl.conf")
+
+    generator_source = config.create_example_msg_generator(num=1, template="\"" + input_message + "\\n\"")
+    app_parser = config.create_app_parser(topic=topic)
+    file_destination = config.create_file_destination(file_name="output.log", template="\"" + template + "\\n" + "\"")
+    config.create_logpath(statements=[generator_source, app_parser, file_destination])
+
+    syslog_ng.start(config)
+    assert file_destination.read_log() == expected_value + "\n"
+
```
mitzkia: If possible I would eliminate the "\n"-s from the testcases.

tests/python_functional/functional_tests/applications/test_applications_with_app_parser.py
```diff
@@ -0,0 +1,64 @@
...
+    ("syslog", "Apr 18 12:33:50 furiel sudo[25173]:   furiel : TTY=pts/13 ; PWD=/home/furiel ; USER=root ; COMMAND=/bin/ls", "${.app.name}", "sudo")
+]
+@pytest.mark.parametrize("topic, input_message, template, expected_value", test_parameters_syslog,
+                         ids=list(map(str, range(len(test_parameters_syslog)))))
```
mitzkia: As I mentioned earlier I would add parameters into the `@pytest.mark.parametrize`, and define separate `id`

tests/python_functional/functional_tests/applications/test_applications_with_app_parser.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
```
mitzkia: I would create a `parsers/app-parser` directory instead of `applications`

tests/python_functional/functional_tests/applications/test_applications_with_app_parser.py
```diff
@@ -0,0 +1,64 @@
...
+    ("syslog", "Apr 18 12:33:50 furiel sudo[25173]:   furiel : TTY=pts/13 ; PWD=/home/furiel ; USER=root ; COMMAND=/bin/ls", "${.app.name}", "sudo")
+]
```
mitzkia: Just found that an example is missing for "scl/cim/adapter.conf" which also uses syslog topic

tests/python_functional/functional_tests/applications/test_applications_with_app_parser.py
mitzkia: Maybe it will be more readable (searchable) if the testcase can be separate into two files: topic-syslog-raw and topic-syslog

tests/python_functional/src/syslog_ng_config/renderer.py
```diff
@@ -47,6 +49,10 @@ def __render(self, re_create_config=None):
     def __render_version(self):
         self.__syslog_ng_config_content += "@version: {}\n".format(self.__syslog_ng_config["version"])
 
+    def __render_includes(self):
+        includes = self.__syslog_ng_config["includes"]
+        self.__syslog_ng_config_content += reduce(lambda x, y: "{}\n{}".format(x, "@include \"{}\"".format(y)), includes, "")
```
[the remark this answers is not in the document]
furiel: Fixed

tests/python_functional/src/syslog_ng_config/renderer.py
```diff
@@ -20,7 +20,7 @@
 # COPYING for details.
 #
 #############################################################################
-
+from functools import reduce
```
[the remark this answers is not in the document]
furiel: Fixed

tests/python_functional/functional_tests/applications/test_applications_with_app_parser.py
furiel: Fixed

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
```diff
@@ -112,3 +117,11 @@ def create_example_msg_generator(self, **options):
         generator_source.DEFAULT_MESSAGE = "-- Generated message. --"
         generator_source.options = options
         return generator_source
+
+    def create_app_parser(self, topic, **options):
+        opts = options.copy()
```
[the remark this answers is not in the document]
furiel: It works without it, but the user would not expect that this function will modifiy its input parameter. Performance-wise it does not really matter, but copying results in safer code. I would stick with this version.

tests/python_functional/functional_tests/applications/test_applications_with_app_parser.py
furiel: Added a test for it. Please check if it looks good! I could not find appropriate message, so I just reverse engieered something that would make the test pass. If you or anyone else has valid log example, please let me know and I will update the test with it.

kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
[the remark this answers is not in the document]
mitzkia: Sorry, I would add another comment here. I understand that why the `topic` option is a separate parameter here (because you wanted to help test writers), but I think it is not necessary. I would say if the test writer did not know the mandatory option of the related config object (driver/parser), he/she should write a minimal testcase with that driver (which means to call a `config.create_my_config_object()` method with empty parameters from the testcase) and syslog-ng will describe which mandatory option is missing (at this point run a half written test is not takes much time, testcase log should contains why syslog-ng can not started). Maybe this is not nice, but the config API (in long-term) can be more clean. If we follow your way we need to be consequent in all drivers/parsers/config objects (which is a lot). If you remember I started to eliminate this kind of config parameters from config methods in my previous PR (which I closed).
I understand that you wanted to help the test writers (imho: +1), but where will be the borderline?
I wanted to keep the API as flexible as can and it can be achieve only with simplifications.
At here you added a hint: test writer must define a topic option, but I can not writer such test where I want to define app-parser without any topic (and the topic="" is not the same as creating an app-parser without a topic option like: app-parser(); ).
And if we can eliminate the topic parameter from the method the copy will also did not needed anymore.

tests/python_functional/functional_tests/parsers/app-parser/test_topic_syslog.py
```diff
@@ -0,0 +1,44 @@
...
+    ("Apr 18 12:33:50 nobody sudo[25173]:   nobody : TTY=pts/13 ; PWD=/home/nobody ; USER=root ; COMMAND=/bin/ls", "${.app.name}", "sudo"),
+    (r"""<0>1 2012-03-05T15:10:34+02:00 localhost kernel 1234 - - @cim: {\"name1\":\"value1\", \"name2\":\"value2\"}""", "${.app.name}", "cim"),
+]
+@pytest.mark.parametrize("input_message, template, expected_value", test_parameters_syslog,
+                         ids=list(map(str, range(len(test_parameters_syslog)))))
```
mitzkia: Sorry, I do not know if you have already finished your change (I saw that tests are already separated in two files), but there was another comment from my side related to test_parameters_syslog/mark.parametrize/id. 
Ones more sorry if you have plan to add this changes later.

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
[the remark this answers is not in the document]
furiel: Fixed.

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/applications/test_applications_with_app_parser.py
furiel: The individual parameters for a test is very long, both in terms of line length and row length, so I think a separate variable increases readability. I tried the inlining, but I think it looks very confusing.

About the different id-s: `ids=list(map(str, range(len(test_parameters_raw))))` already generates unique ids. What is the request here? To couple the test parameters and id into a `pytest.param` class? Or the id's are not expressive enough?

I can do both: with this this stateful generator I could generate ids from the last parameter + index. 
This works for now as the last member is simple. Later we might need trimming and special character sanitization. 

```
class TestParamGenerator():
    def __init__(self):
        self.occurrences = {}
    def __call__(self, test_parameters):
        last = test_parameters[-1]
        if last in self.occurrences:
            self.occurrences[last] = str(int(self.occurrences[last]) + 1)
        else:
            self.occurrences[last] = "0"
        return pytest.param(*test_parameters, id=last+self.occurrences[last])
```

Does this sound good?

mitzkia: Please check my changes. I would change the testcase in this way.
In my changes:
- I have added `id` to the parameters, the main goal is here: to identify each testcase from outside (when the test is running our working directory will contains this `id` and the working_dir will be displayed in foreground console log, or we can also grep for these cases. With this the case identifier: `id` is very next to the testcase. with using classes we will lose this behaviour).
- I have also eliminates the "\n"-s from the test itself, I have put them inside the parameters, these "\n"-s are the parts of our input_message, template and expected_value
- I have also truncates the long lines with some indentations, I think this is fine.

```
import pytest


@pytest.mark.parametrize("input_message, template, expected_value", [
    pytest.param(
        '''"<0>1 2012-03-05T15:10:34+02:00 localhost kernel 1234 - - transport_info='UDP: [10.30.35.100]:60513->[10.30.0.35]:162', DISMAN-EVENT-MIB::sysUpTimeInstance='(277195) 0:46:11.95', SNMPv2-MIB::snmpTrapOID.0='SNMPv2-SMI::enterprises.9.9.41.2.0.1.0', SNMPv2-SMI::enterprises.9.9.41.1.2.3.1.2.0='user', SNMPv2-SMI::enterprises.9.9.41.1.2.3.1.3.0='6', SNMPv2-SMI::enterprises.9.9.41.1.2.3.1.4.0='', SNMPv2-SMI::enterprises.9.9.41.1.2.3.1.5.0='[ 2771.344837] PF: filter/input DROP IN=eth0 OUT= MAC=08:00:27:d5:33:1e:34:e6:d7:1f:04:58:08:00 SRC=10.30.0.35 DST=10.30.35.100 LEN=60 TOS=0x00 PREC=0x00 TTL=1 ID=62309 DF PROTO=TCP SPT=52988 DPT=5355 WINDOW=29200 RES=0x00 SYN URGP=0 ', SNMPv2-SMI::enterprises.9.9.41.1.2.3.1.6.0='(277195) 0:46:11.95', SNMP-COMMUNITY-MIB::snmpTrapAddress.0='127.0.0.1\n'"''',
        '''"${.app.name}\n"''',
        "iptables\n",
        id="iptables"
    ),
    pytest.param(
        """'<14>1 2019-04-12T14:29:06.576Z - RT_FLOW - RT_FLOW_SESSION_CLOSE [junos@2636.1.1.1.2.129 reason="TCP FIN" source-address="10.3.255.203" source-port="47776" destination-address="8.23.224.110" destination-port="80" connection-tag="0" service-name="junos-http" nat-source-address="10.3.136.49" nat-source-port="19162" nat-destination-address="8.23.224.110" nat-destination-port="80" nat-connection-tag="0" src-nat-rule-type="source rule" src-nat-rule-name="nat1" dst-nat-rule-type="N/A" dst-nat-rule-name="N/A" protocol-id="6" policy-name="permit_all" source-zone-name="trust" destination-zone-name="untrust" session-id-32="5" packets-from-client="6" bytes-from-client="337" packets-from-server="4" bytes-from-server="535" elapsed-time="1" application="HTTP" nested-application="UNKNOWN" username="N/A" roles="N/A" packet-incoming-interface="ge-0/0/0.0" encrypted="No" application-category="Web" application-sub-category="N/A" application-risk="4" application-characteristics="Can Leak Information;Supports File Transfer;Prone to Misuse;Known Vulnerabilities;Carrier of Malware;Capable of Tunneling;"] session closed TCP FIN: 10.3.255.203/47776->8.23.224.110/80 0x0 junos-http 10.3.136.49/19162->8.23.224.110/80 0x0 source rule nat1 N/A N/A 6 permit_all trust untrust 5 6(337) 4(535) 1 HTTP UNKNOWN N/A(N/A) ge-0/0/0.0 No Web N/A 4 Can Leak Information;Supports File Transfer;Prone to Misuse;Known Vulnerabilities;Carrier of Malware;Capable of Tunneling;\n'""",
        '''"${.app.name}\n"''',
        "junos\n",
        id="junos"
    ),
    pytest.param(
        '''"Apr 18 12:33:50 nobody sudo[25173]:   nobody : TTY=pts/13 ; PWD=/home/nobody ; USER=root ; COMMAND=/bin/ls\n"''',
        '''"${.app.name}\n"''',
        "sudo\n",
        id="sudo"
    ),
    pytest.param(
        '''"<0>1 2012-03-05T15:10:34+02:00 localhost kernel 1234 - - @cim: {'name1':'value1', 'name2':'value2'}\n"''',
        '''"${.app.name}\n"''',
        "cim\n",
        id="cim"
    ),
])
def test_application_syslog(config, syslog_ng, input_message, template, expected_value):
    config.add_include("scl.conf")

    generator_source = config.create_example_msg_generator(num=1, template=input_message)
    syslog_parser = config.create_syslog_parser(flags="syslog-protocol")
    app_parser = config.create_app_parser(topic="syslog")
    file_destination = config.create_file_destination(file_name="output.log", template=template)
    config.create_logpath(statements=[generator_source, syslog_parser, app_parser, file_destination])

    syslog_ng.start(config)
    assert file_destination.read_log() == expected_value
```

furiel: I would prefer the other version.

The suggested version introduces copy paste (the expected value is the same as the id), that my version avoids by generating exactly the same ids. (There is also a static copy paste `pytest.param`, 
but that is a minor concern for me, the logical copy paste bothers me more).
The suggested version suppresses the case when multiple tests would cover the same application. `TestParamGenerator` covers this case too by appending ordinals to the id.
`"\n"` is not advisable to be added to the messages. `"\n"` comes from the fact that we are using file destination. If we used udp destination for example, the same test would work without involving newlines. This means `\n` is a file destination specific detail, and in that sense it should not be part of the message.

tests/python_functional/functional_tests/applications/test_applications_with_app_parser.py
furiel: I pushed a patch to the top which moves the `\n`-s to the parameters, instead of adding them in the test. Please check if this is better. Now the parameters are a little bit trickier, because the new lines need to be escaped differently in the template variable and the expected variable: for example
```
([cisco message example], "${.app.name}\\n", "cisco\n")
```
Of course the same symptom existed in the previous version (the root cause is the same), just expressed in the test instead of the test parameters.

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/parsers/app-parser/test_topic_syslog.py
```diff
@@ -0,0 +1,44 @@
...
+    ("Apr 18 12:33:50 nobody sudo[25173]:   nobody : TTY=pts/13 ; PWD=/home/nobody ; USER=root ; COMMAND=/bin/ls", "${.app.name}\\n", "sudo\n"),
+    (r"""<0>1 2012-03-05T15:10:34+02:00 localhost kernel 1234 - - @cim: {\"name1\":\"value1\", \"name2\":\"value2\"}""", "${.app.name}\\n", "cim\n"),
+]
+@pytest.mark.parametrize("input_message, template, expected_value", test_parameters_syslog,
+                         ids=list(map(str, range(len(test_parameters_syslog)))))
+def test_application_syslog(config, syslog_ng, input_message, template, expected_value):
+    config.add_include("scl.conf")
+
+    generator_source = config.create_example_msg_generator(num=1, template="\"" + input_message + "\"")
+    syslog_parser = config.create_syslog_parser(flags="syslog-protocol")
```
Kokan: Combined with testing each parser, having this classification and do an end-to-end with default-network driver I think it should cover the majority, meaning this test is imho fine only testing classification; and the result of as such the `template` parameter could have been part of the test itself as fixed value. But this is more like a note, not really a change request, not even an optional one.

PR #2687, opened 2019-04-18
Version generation improvement

MrAnno:
In case when git is not installed or no tags exist, we should fall back to the VERSION file.

This commit fixes the compilation in environments where git is not available or where `git clone --depth` is used, such as Travis.

### DIFF
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -26,9 +26,8 @@ BASEDIR=$(dirname "$0")
 
 cd $BASEDIR/../
 
-if [ -d .git ]; then
-  git describe --tags --dirty --abbrev=7 | sed 's/^syslog-ng-//' | tr '-' '.' | tr -d '\n'
+if [ -d .git ] && GIT_VERSION=$(git describe --tags --dirty --abbrev=7); then
+  echo $GIT_VERSION | sed 's/^syslog-ng-//' | tr '-' '.' | tr -d '\n'
 else
   cat VERSION
 fi
-
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -29,5 +29,5 @@ cd $BASEDIR/../
 if [ -d .git ] && GIT_VERSION=$(git describe --tags --dirty --abbrev=7); then
   echo $GIT_VERSION | sed 's/^syslog-ng-//' | tr '-' '.' | tr -d '\n'
 else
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

scripts/version.sh
```diff
@@ -26,7 +26,7 @@ BASEDIR=$(dirname "$0")
 
 cd $BASEDIR/../
 
-if GIT_VERSION=$(git describe --tags --dirty --abbrev=7); then
+if GIT_VERSION=$(git describe --tags --dirty --abbrev=7 --match 'syslog-ng-*'); then
```
Kokan: I still would not include filtering based on tag name, as it should not care much about tag naming; I do not see any benefits via excluding tag names.
My assumption are the following: you must never tag master other then released stuff, otherwise it became very fast very ugly (seen git tag for every CI execution...)
If you tag a commit, you really want it to called by its name, so getting back that name like prefer to see: `best-feature-ever-1-g21ced7a` instead of `3.51-12313-g21ced7a`.

MrAnno: This is a conceptual thing, so I don't mind accepting all tags, if you prefer to do so,  but in this case I sense some contradiction:

If we accept all tags as a version, why would we cut the `syslog-ng-` or the `v` prefix off from `syslog-ng-3.21.1`/`v3.21.1`? In my understanding, we do this because we have a specific expectation on how the version should look like.

MrAnno:
> (btw not having git should not be solved as you either has .git and git or not; well most of the time)

I've added back the `.git` dircheck.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

scripts/version.sh
Kokan: Nor I agree with `syslog-ng-` prefixed tags, nor I have all the knowledge why such tagging strategy is used; should just `3.20.1` be just fine, and there were no need to removing strings etc...

I would have been fine with `syslog-ng-3.20.1` format; but did the transformation in order not to break the builds created from released commits.

The `tr '-' '.'` is truly unfortunate thing :(, maybe I should have solved it in a different way, as packaging leaked into this version script.

kira-syslogng:
Build SUCCESS

Issue #2670, opened 2019-04-11
feature request - expand apache parser


nambrosch:
Would it be possible to add support for the unofficial vcommon & vcombined formats to the apache parser?  I would like to pass logs via syslog however would be unable to pass vhosts.

apache config:
```
LogFormat "%v:%p %a %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-agent}i\"" vcombined
CustomLog "|/usr/bin/logger -S 65536 -t 'apache' -p info" vcombined
```

This will add the additional parameters (hostname and port) so a request can be identified:
```
localhost:80 1.3.4.4 - - [11/Apr/2019:14:52:52 -0400] "GET / HTTP/1.1" 403 4897 "-" "curl/7.29.0"
```

A simple regex to identify would be `^[A-Za-z0-9-_.]{4,254}:[0-9]{1,5}` (although it could be made a bit stricter).  I do not have a better way to get the vhost into syslog-ng.  I am currently using a custom script for logging instead of syslog-ng and would like to change that.

bazsi:
This is a good idea, and I think can be trivially added using an if {}
statement, if the format can be identified.

I think wr could use the colon in the first column to do that.

E.g. something like

```
if (match("^[\-a-zA-Z0-9-]+:[0-9]+ ")) {
   # combined format, csv-parser according to the combined format
   parser { csv-parser(...) };
}
else {
   #... Current format
}
```

Could you cook a pull request? The apache parser is somewhere in scl/apache
subdirectory.

Thanks

nambrosch:
Before I submit the PR, how does this look?  Three new macros are exposed, `${vhost}`, `${port}`, and `${vhostport}`.

```code
block parser apache-accesslog-parser(prefix(".apache.") template("${MESSAGE}")) {
    channel {
        # log format including vhost:port
        if {
            filter { match("^[\-a-zA-Z0-9-]+:[0-9]+ " value("MESSAGE")); };
            parser {
                csv-parser(
                    prefix(`prefix`)
                    dialect(escape-double-char)
                    flags(strip-whitespace)
                    delimiters(" ")
                    template(`template`)
                    quote-pairs('""[]')
                    columns("vhostport", "clientip", "ident", "auth",
                            "timestamp", "rawrequest", "response",
                            "bytes", "referrer", "agent"));

                csv-parser(
                    prefix(`prefix`)
                    template("${`prefix`rawrequest}")
                    delimiters(" ")
[...]

PR #2688, opened 2019-04-19
support for vhost:port in common/combined log formats

nambrosch:
This patch will add support for vhost & port in common & combined apache logs as discussed @ https://github.com/balabit/syslog-ng/issues/2670.  Here is that format -

```
foo.com:443 1.2.3.4 - - [15/Apr/2019:14:30:16 -0400] "GET /bar.html HTTP/2.0" 500 - "https://foo.com/referer.html" "Mozilla/5.0 ..."
```

Below is my use case that I'm using on a test server, using the `${.apache.vhost}` macro to create a directory then logging using combined format -

apache:
```
LogFormat "%v:%p %h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-agent}i\"" vcombined
CustomLog "|/usr/bin/logger -S 65536 -t apache -p info -d -n 127.0.0.1 -P 5141" vcombined
```

syslog-ng:
```
source s_apache_access {
  network(ip(127.0.0.1) port(5141) transport(udp) flags(syslog-protocol));
};

destination d_apache_access {
  file (
    "/var/log/httpd/${.apache.vhost}/access_log_${YEAR}${MONTH}${DAY}"
    template("${.apache.clientip} ${.apache.ident} ${.apache.auth} [${.apache.timestamp}] \"${.apache.rawrequest}\" ${.apache.response} ${.apache.bytes} \"${.apache.referrer}\" \"${.apache.agent}\"\n")
  );
};

log {
  source(s_apache_access);
  parser { apache-accesslog-parser(); };
  destination(d_apache_access);
};
```

### DIFF
--- a/scl/apache/apache.conf
+++ b/scl/apache/apache.conf
@@ -47,7 +47,6 @@ block parser apache-accesslog-parser(prefix(".apache.") template("${MESSAGE}")) 
                     delimiters(":")
                     dialect(escape-none)
                     columns("vhost", "port"));
-
             };
 
         # if traditional log format
--- a/scl/apache/apache.conf
+++ b/scl/apache/apache.conf
@@ -20,9 +20,21 @@
 #
 #############################################################################
 
-# combined & common format including vhost
+# Parse apache access.log
+#
+# Formats recognized:
+#
 # LogFormat "%v:%p %h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" vhost_combined
+#    virtualhost:443 127.0.0.1 - frank [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326 "http://www.example.com/start.html" "Mozilla/4.08 [en] (Win98; I ;Nav)"
+#
 # LogFormat "%v:%p %h %l %u %t \"%r\" %>s %b" vhost_common
+#    virtualhost:443 127.0.0.1 - frank [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326
+#
+# LogFormat "%h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-agent}i\"" combined
+#    127.0.0.1 - frank [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326 "http://www.example.com/start.html" "Mozilla/4.08 [en] (Win98; I ;Nav)"
+#
+# LogFormat "%h %l %u %t \"%r\" %>s %b" common
+#    127.0.0.1 - frank [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326
 block parser apache-accesslog-parser-vhost(prefix() template()) {
     channel {
         rewrite {
@@ -73,6 +85,8 @@ block parser apache-accesslog-parser-combined(prefix() template()) {
 };
 
 block parser apache-accesslog-parser(prefix(".apache.") template("${MESSAGE}")) {
+    # parse into a logstash-like schema
+    # https://github.com/elastic/logstash/blob/v1.4.2/patterns/grok-patterns#L90
     channel {
 
         # parser for formats including vhost:port
@@ -84,7 +98,8 @@ block parser apache-accesslog-parser(prefix(".apache.") template("${MESSAGE}")) 
             parser { apache-accesslog-parser-combined(prefix(`prefix`) template(`template`)); };
         };
 
-        # parser for all formats
+        # mungle values to match Kibana/elastic schema and common to all
+        # supported formats.
         parser {
             csv-parser(
                 prefix(`prefix`)
--- a/scl/apache/apache.conf
+++ b/scl/apache/apache.conf
@@ -20,22 +20,54 @@
 #
 #############################################################################
 block parser apache-accesslog-parser(prefix(".apache.") template("${MESSAGE}")) {
-
-
     channel {
-        parser {
-            csv-parser(
-                prefix(`prefix`)
-                dialect(escape-double-char)
-                flags(strip-whitespace)
-                delimiters(" ")
-                template(`template`)
-                quote-pairs('""[]')
-                # field names match of that of Logstash
-                columns("clientip", "ident", "auth",
-                        "timestamp", "rawrequest", "response",
-                        "bytes", "referrer", "agent"));
+        rewrite {
+            set("`template`" value("1"));
+        };
 
+        # if log line begins with vhost:port
+        if {
+            filter { match("^[A-Za-z0-9\-\._]+:[0-9]+ " value("1")); };
+            parser {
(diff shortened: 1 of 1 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

nambrosch:
i used four spaces for indenting instead of tabs, what is your convention?

bazsi:
4 spaces is much better than a tab. So this change is welcome.

kira-syslogng:
Build SUCCESS

furiel:
The patch itself looks good to me. I just have one question for clarification.

> Would it be possible to add support for the unofficial vcommon & vcombined formats to the apache parser?

Could you explain what do you mean by unofficial format? I tried to investigate, and here is what I found. According to the [documentation](https://httpd.apache.org/docs/2.4/logs.html), the official combined log format is
```
LogFormat "%h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-agent}i\"" combined
```
which is different than the introduced version.

I fired up an apache server on centos:7 in docker, and they use the official format above.
However, with ubuntu:bionic, they go with your version, except they call it vhost_combined instead of vcombined.

So what is happening here is that some distributions deviate from the official log format? And the goal here is to support those distributions too? Out of curiosity, which distribution do you use?

Btw, I do not have problem with that. I like the idea that syslog-ng should work out of the box, and support the deviations of major distributions.

scl/apache/apache.conf
```diff
@@ -20,22 +20,53 @@
...
-                        "bytes", "referrer", "agent"));
+                csv-parser(
+                    prefix(`prefix`)
+                    template("$2")
+                    delimiters(":")
+                    dialect(escape-none)
+                    columns("vhost", "port"));
+            };
 
+        # if traditional log format
```
furiel: From clean code point of view, instead of this comment, you could give a name to these parser blocks. Analogously to
```
block parser apache-accesslog-parser(prefix(".apache.") template("${MESSAGE}"))
```
you could extract these branches into `apache-access-log-traditional-part` and `apache-access-log-combined-part`. That would make the scl a little more readable. As maybe there are other distributions with different format, it would be easier to extend later on.

nambrosch: I wasn't aware that ubuntu has this LogFormat already available, thanks for pointing it out.  Fedora/CentOS does not include this format.  It might make more sense to reference the name as vhost_combined instead of vcombined.

Shifting from 1 -> 3 blocks is a good idea, I'll explore that tonight.

kira-syslogng:
Build SUCCESS

bazsi:
Hi,

This looks better indeed @furiel. Another layer of refinement occured to
me, but instead of describing it in email, here's the patch. I am sure
@nambrosch is a bit fed up with all the improvements :) can you still apply
this to your solution?

Basically I've moved the filter to apache-accesslog-parser-vhost(), which
will drop all messages that don't match the required format. I've also
fixed a bug, so that template() option to
apache-accesslog-parser-combined() actually uses that value, instead of
hard-coding $1.

I didn't test this this round, so it'd be appreciated if you could
doublecheck that it still works for you.


```
diff --git a/scl/apache/apache.conf b/scl/apache/apache.conf
index 440e64269..9046f2452 100644
--- a/scl/apache/apache.conf
+++ b/scl/apache/apache.conf
@@ -25,6 +25,10 @@
 # LogFormat "%v:%p %h %l %u %t \"%r\" %>s %b" vhost_common
 block parser apache-accesslog-parser-vhost(prefix() template()) {
     channel {
+        rewrite {
+            set("`template`" value("1"));
+        };
+        filter { match("^[A-Za-z0-9\-\._]+:[0-9]+ " value("1")); };
         parser {
             csv-parser(
                 dialect(escape-double-char)
@@ -59,7 +63,7 @@ block parser apache-accesslog-parser-combined(prefix()
template()) {
                 dialect(escape-double-char)
                 flags(strip-whitespace)
                 delimiters(" ")
-                template($1)
+                template(`template`)
                 quote-pairs('""[]')
                 columns("clientip", "ident", "auth",
                         "timestamp", "rawrequest", "response",
@@ -69,14 +73,12 @@ block parser apache-accesslog-parser-combined(prefix()
template()) {
 };

 block parser apache-accesslog-parser(prefix(".apache.")
template("${MESSAGE}")) {
+    # parse into a logstash-like schema
+    #
https://github.com/elastic/logstash/blob/v1.4.2/patterns/grok-patterns#L90
     channel {
-        rewrite {
-            set("`template`" value("1"));
-        };

         # parser for formats including vhost:port
         if {
-            filter { match("^[A-Za-z0-9\-\._]+:[0-9]+ " value("1")); };
             parser { apache-accesslog-parser-vhost(prefix(`prefix`)
template("$1")); };

         # parser for standard formats
@@ -84,7 +86,8 @@ block parser apache-accesslog-parser(prefix(".apache.")
template("${MESSAGE}"))
             parser { apache-accesslog-parser-combined(prefix(`prefix`)
template("$1")); };
         };

-        # parser for all formats
+        # mungle values to match Kibana/elastic schema and common to all
+        # supported formats.
         parser {
             csv-parser(
                 prefix(`prefix`)
```

bazsi:
Here's that last branch as a git branch, feel free to integrate it into yours.

https://github.com/bazsi/syslog-ng/pull/new/apache-access-log-parser-improvements

nambrosch:
i'm still using $1 two more times in `block parser apache-accesslog-parser` so removing ``set("`template`" value("1"));`` causes it to break.

moving the filter and changing $1 -> `template` in `block parser apache-accesslog-parser-combined` is logical though.

bazsi:
I have just added a few more patches on top of the initial patch and sent a PR to @nambrosch With those patches this has my approval and I think it would be great if we could merge them by 3.21

Alternatively, we could my branch directly, it contains both this patch and my improvements
https://github.com/bazsi/syslog-ng/tree/apache-access-log-parser-improvements

nambrosch:
@bazsi Thanks, I'm happy with those changes - confirmed it works on a fedora host and merged the PR.

kira-syslogng:
Build SUCCESS

bazsi:
Btw, I implemented a template () option for match () to make these use
cases easier to implement, but this can be merged as is, we could
eventually make use of that in this parser as well.

I hope I can publish that branch once I finished the tests.

Bazsi

nambrosch:
> Approved. It would be nice if you could get rid off the merge commit within the branch, just make sure its a set of linear patches.

Thanks.  Is it possible to remove that?  I think trying to revert the merge commit in nambrosch/syslog-ng creates an additional PR, no?

Kokan:
@nambrosch you should be fine to hard reset to @bazsi patch and force push it. (it should not create new PR)

Something like this: https://asciinema.org/a/zRtDlGZjId1VheJeiMoqP4qr9

nambrosch:
ah ha, perfect, all set.

kira-syslogng:
Build SUCCESS

PR #2691, opened 2019-04-22
Reset the timezone on config reload event.

ring-oleksii-hamov:
Currently, the behaviour of syslog-ng regarding timezone changes handling
depends on SYSLOG_NG_HAVE_LOCALTIME_R configuration parameter. If
SYSLOG_NG_HAVE_LOCALTIME_R is set to '1', the timezone is not updated
in run time and vise versa. However, timezone setting is considered as
a costly operation, so it can be not desired to perform it
frequently. One of the wice options is setting the timezone on-demand.
The code sets timezone on config reload event.

### DIFF
--- a/lib/timeutils/misc.c
+++ b/lib/timeutils/misc.c
@@ -26,6 +26,8 @@
 #include "timeutils/cache.h"
 #include "messages.h"
 
+#include <apphook.h>
+
 #include <string.h>
 
 /**
@@ -141,4 +143,23 @@ glong
 timespec_diff_nsec(struct timespec *t1, struct timespec *t2)
 {
   return (glong)((t1->tv_sec - t2->tv_sec) * 1e9) + (t1->tv_nsec - t2->tv_nsec);
+}
+
+void timeutils_setup_timezone_hook(void);
+
+static void
+timeutils_reset_timezone(gint type, gpointer context)
+{
+  tzset();
+
+  // Invalidate time cache to apply time zone as soon as possible.
+  invalidate_cached_time();
+  clean_time_cache();
+  timeutils_setup_timezone_hook();
+}
+
+void
+timeutils_setup_timezone_hook(void)
+{
+  register_application_hook(AH_CONFIG_CHANGED, timeutils_reset_timezone, NULL);
 }
--- a/syslog-ng/main.c
+++ b/syslog-ng/main.c
@@ -38,6 +38,7 @@
 #include "plugin.h"
 #include "reloc.h"
 #include "resolved-configurable-paths.h"
+#include "timeutils/misc.h"
 
 #include <sys/types.h>
 #include <stdio.h>
@@ -276,6 +277,9 @@ main(int argc, char *argv[])
    */
   g_process_start();
   app_startup();
+
+  timeutils_setup_timezone_hook();
+
   main_loop_options.server_mode = ((SYSLOG_NG_ENABLE_FORCED_SERVER_MODE) == 1 ? TRUE : FALSE);
   main_loop_init(main_loop, &main_loop_options);
   rc = main_loop_read_and_init_config(main_loop);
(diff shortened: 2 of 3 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

bazsi:
Instead of adding these calls into the reload logic directly, you should use the apphook framework to register at AH_CONFIG_CHANGED and do that from timeutils.

ring-oleksii-hamov:
@bazsi, thanks for the review. I'll fix this, however, as far as I understand, apphook framework allows only one-shot hooks. Also, a hook can't add itself while being called, which means that it has to be added from another function. If it is ok for you, I can do that.

bazsi:
there could be a global initialization function that registers the callback
the first time.

AH_CONFIG_CHANGED is invoked at every reload and the function can register
itself for the next reload.

bazsi:
@ring-oleksii-hamov do you plan on improving these patches?

ring-oleksii-hamov:
@bazsi, done.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/timeutils/misc.c
```diff
@@ -142,3 +144,23 @@ timespec_diff_nsec(struct timespec *t1, struct timespec *t2)
 {
   return (glong)((t1->tv_sec - t2->tv_sec) * 1e9) + (t1->tv_nsec - t2->tv_nsec);
 }
+
+void timeutils_setup_timezone_hook(void);
+
+static void timeutils_reset_timezone(gint type, gpointer context)
+{
```
bazsi: please break the return value into a separate line as elsewhere in the codebase.

lib/timeutils/misc.c
```diff
@@ -142,3 +144,23 @@ timespec_diff_nsec(struct timespec *t1, struct timespec *t2)
...
+  (void) type;
+  (void) context;
+
```
bazsi: please either add G_GNUC_UNUSED to the arguments, or remove these lines. I don't think the compile would warn about these.

lib/timeutils/misc.c
```diff
@@ -142,3 +144,23 @@ timespec_diff_nsec(struct timespec *t1, struct timespec *t2)
...
+  tzset();
+
+  // Invalidate time cache to apply time zone as soon as possible.
+  invalidate_cached_time();
+  clean_time_cache();
+  timeutils_setup_timezone_hook();
+}
+
+void timeutils_setup_timezone_hook(void)
+{
```
bazsi: void should go to a separate line.

syslog-ng/main.c
```diff
@@ -208,6 +208,8 @@ main(int argc, char *argv[])
   GOptionContext *ctx;
   GError *error = NULL;
 
+  timeutils_setup_timezone_hook();
+
   MainLoop *main_loop = main_loop_get_instance();
```
bazsi: please move this right after app_startup(). should be just as good, unless I am missing something. This probably works here, but "feels" a bit too early. I don't think it is important to cover our command line parsing.

ring-oleksii-hamov:
@bazsi, addressed your comments

kira-syslogng:
Build SUCCESS

Kokan:
1. Please fix the Travis build failure (Travis build is made with `--enable-extra-warnings` used during configure).
2. Follow commit message format:
```
module: short description

optional longer description

Signed-off-by: 
```

kira-syslogng:
Build SUCCESS

ring-oleksii-hamov:
@Kokan, done. Seems like now Travis failure occurred due to a non-related error:
```
curl: (6) Could not resolve host: homebrew.bintray.com
Error: Failed to download resource "protobuf"
```

m10812021:
How do I set SYSLOG_NG_HAVE_LOCALTIME_R ?

alltilla:
It is calculated in configure time. `configure` or `cmake` looks for the `localtime_r` function in `time.h`.

m10812021:
Thank you for your swift reply.

PR #2698, opened 2019-04-30
Fix system source app parser reference

bazsi:
This is an alternative for #2685 opened to facilitate discussion.


kira-syslogng:
Build SUCCESS

gaborznagy:
After our team discussion we agreed that `system-unix` topic will be dropped, but we will change the default behaviour of app-parser, so "app-parser() drops the message if no adapters are registered".

I've moved the corresponding commit to #2685.
I'm closing this PR.

PR #2700, opened 2019-04-30
Add implode and explode template functions

bazsi:
This branch adds $(explode) and $(implode) template functions. The first one turns a string separated by a specific character into a list, the other turns a list into a string gluing the pieces together with a separator.

Example:

                            value("MESSAGE", "$1 1 0 $(implode ' ' $(list-slice :-1 $(context-values $MSG)))")

### DIFF
--- a/modules/basicfuncs/tests/test_basicfuncs.c
+++ b/modules/basicfuncs/tests/test_basicfuncs.c
@@ -400,6 +400,14 @@ Test(basicfuncs, test_list_funcs)
   assert_template_format("$(list-slice :-6 foo,bar,xxx,baz,bad)", "");
 
   assert_template_format("$(list-count foo,bar,xxx, baz bad)", "5");
+
+  assert_template_format("$(explode ' ' foo bar xxx baz bad)", "foo,bar,xxx,baz,bad");
+  assert_template_format("$(explode ' ' 'foo bar xxx baz bad')", "foo,bar,xxx,baz,bad");
+  assert_template_format("$(explode ';' foo;bar;xxx;baz;bad)", "foo,bar,xxx,baz,bad");
+  assert_template_format("$(explode ';' foo;bar xxx;baz;bad)", "foo,bar,xxx,baz,bad");
+
+  assert_template_format("$(implode ' ' foo,bar,xxx,baz,bad)", "foo bar xxx baz bad");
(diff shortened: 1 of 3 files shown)


modules/basicfuncs/tests/test_basicfuncs.c
```diff
@@ -397,6 +400,14 @@ Test(basicfuncs, test_list_funcs)
   assert_template_format("$(list-slice :-6 foo,bar,xxx,baz,bad)", "");
 
   assert_template_format("$(list-count foo,bar,xxx, baz bad)", "5");
+
+  assert_template_format("$(explode ' ' foo bar xxx baz bad)", "foo,bar,xxx,baz,bad");
+  assert_template_format("$(explode ' ' 'foo bar xxx baz bad')", "foo,bar,xxx,baz,bad");
+  assert_template_format("$(explode ';' foo;bar;xxx;baz;bad)", "foo,bar,xxx,baz,bad");
+  assert_template_format("$(explode ';' foo;bar xxx;baz;bad)", "foo,bar,xxx,baz,bad");
```
Kokan: I would say this is a bug rather expected behaviour of an explode, the result should be something like this: `foo,"bar xxx",baz,bad".

bazsi: Template functions take a  list of arguments and a lot tries to interpret those in some way. I think in most cases $(explode) would only take one argument, and the question is what to do with the multiple argument case.

To concatenate strings, you can just use the template syntax:

```
$(explode ';' $SOMETHING$SOMETHINGELSE)
```

The question what to do in case it is two separate args:

```
$(explode ';' $SOMETHING $SOMETHINGELSE)
```

If you want to split those strings, I wouldn't consider the first one an element on its own right, as both should be split and the resulting list be concatenated. I think this is a useful use-case, although I don't have a concrete example at hand. The above is equivalent to to more verbose:

```
$(list-concat $(explode ';' $SOMETHING) $(explode ';' SOMETHINGELSE))
```

And like I said, this makes more sense to me when $(explode) gets multiple arguments. And do note that the space between "bar" and "xxx" is not part of the argument, rather it is a separator in the argument list.

MrAnno: This makes sense to me, other template functions also have this append-like behavior.

PR #2701, opened 2019-04-30
groupingby: add sort-key() option

bazsi:
This patch implements sorting of the elements accumulated into a context.
It is useful when entries are not received in order.

Example:
```
block parser cisco-ise-parser() {
    channel {
        parser {
            csv-parser(columns('1', '2', '3', 'MSG') flags(greedy));
            grouping-by(
                        scope("host")
                        key("$1")
                        sort-key("$3")
                        trigger("$(context-length)" eq "$2")
                        timeout(5)
                        aggregate(
                            inherit-mode(last-message)
                            value("MESSAGE", "$(if ('$(context-length)' eq '$(+ $2 1)')
                                                        '$1 1 0 $(implode \" \" $(list-slice :-1 $(context-values $MSG)))'
                                                        '$1 1 0 SYSLOG TRUNCATED $(list-slice :-1 $(context-values $3-$MSG))')")
                            tags(".cise.joined_message"))
            );
        };
        filter { tags(".cise.joined_message"); };
    };
};
```

### DIFF
--- a/lib/template/templates.c
+++ b/lib/template/templates.c
@@ -79,6 +79,10 @@ _calculate_triviality(LogTemplate *self)
 
   LogTemplateElem *e = (LogTemplateElem *) self->compiled_template->data;
 
+  /* reference to non-last element of the context, that's not trivial */
+  if (e->msg_ref > 0)
+    return FALSE;
+
   switch (e->type)
     {
     case LTE_FUNC:
--- a/lib/template/templates.c
+++ b/lib/template/templates.c
@@ -28,11 +28,90 @@
 #include "template/escaping.h"
 #include "cfg.h"
 
+gboolean
+log_template_is_trivial(LogTemplate *self)
+{
+  return self->trivial;
+}
+
+const gchar *
+log_template_get_trivial_value(LogTemplate *self, LogMessage *msg, gssize *value_len)
+{
+  g_assert(self->trivial);
+
+  LogTemplateElem *e = (LogTemplateElem *) self->compiled_template->data;
+
+  switch (e->type)
+    {
+    case LTE_MACRO:
+      if (e->text_len > 0)
+        {
+          if (value_len)
+            *value_len = e->text_len;
+          return e->text;
+        }
+      else if (e->macro == M_MESSAGE)
+        return log_msg_get_value(msg, LM_V_MESSAGE, value_len);
+      else if (e->macro == M_HOST)
+        return log_msg_get_value(msg, LM_V_HOST, value_len);
+      g_assert_not_reached();
+    case LTE_VALUE:
+      return log_msg_get_value(msg, e->value_handle, value_len);
+    default:
+      g_assert_not_reached();
+    }
+}
+
+static gboolean
+_calculate_triviality(LogTemplate *self)
+{
+  /* if we need to escape, that's not trivial */
+  if (self->escape)
+    return FALSE;
+
+  /* no compiled template */
+  if (self->compiled_template == NULL)
+    return FALSE;
+
+  /* more than one element */
+  if (self->compiled_template->next != NULL)
+    return FALSE;
+
+  LogTemplateElem *e = (LogTemplateElem *) self->compiled_template->data;
+
+  switch (e->type)
+    {
+    case LTE_FUNC:
+      /* functions are never trivial */
+      return FALSE;
+    case LTE_MACRO:
+      /* Macros are trivial if they only contain a text but not a real
+       * macro.  Empty strings are represented this way.  */
+      if (e->macro == M_NONE)
+        return TRUE;
+      if (e->text_len > 0)
+        return FALSE;
+
+      /* we have macros for MESSAGE and HOST for compatibility reasons, but
+       * they should be considered trivial */
+
+      if (e->macro == M_MESSAGE || e->macro == M_HOST)
+        return TRUE;
+      return FALSE;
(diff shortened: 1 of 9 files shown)


modules/dbparser/dbparser-grammar.ym
```diff
@@ -126,6 +127,7 @@ grouping_by_opts
 
 grouping_by_opt
 	: KW_KEY '(' template_content ')'                       { grouping_by_set_key_template(last_parser, $3); log_template_unref($3); }
+        | KW_SORT_KEY '(' string ')'				{ grouping_by_set_sort_key(last_parser, log_msg_get_value_handle($3)); free($3); }
```
MrAnno: Just one question:
`key()` receives a template, for example: `key("${MESSAGE}")`, but `sort-key()` wants a string, the name of the key: `sort-key("PROGRAM")`.

Wouldn't it be better if it was unified?
Using NVHandles is much cheaper than formatting templates, that's an advantage.

However, `key()` as a template supports multiple keys: `key("$field1 $field2")`. The same might be possible for sorting keys.

modules/dbparser/groupingby.c
```diff
@@ -259,6 +268,8 @@ grouping_by_expire_entry(TimerWheel *wheel, guint64 now, gpointer user_data)
             evt_tag_str("context-id", context->key.session_id),
             log_pipe_location_tag(&self->super.super.super));
 
+  if (self->sort_key)
+    correllation_context_sort(context, self->sort_key);
```
MrAnno: It might worth mentioning in the documentation that sorting is done when the context is about to be closed, but before evaluating `having()` and creating the synthetic message.

(I don't know if it's possible to use message indexes in filters for example, but if so, `trigger()` will see a different order than `having()`.)

modules/dbparser/dbparser-grammar.ym
bazsi: I saw the inconsistency, however this is much faster. The compare function needs to be called a lot more times than the key function, and I was thinking this should be enough.

Maybe a sort-key() and a sort-value() could work?

Or we might be able to recognize if the parameter is a template or just one value.

bazsi: I saw the inconsistency, however this is much faster. The compare function needs to be called a lot more times than the key function, and I was thinking this should be enough.

Maybe a sort-key() and a sort-value() could work?

Or we might be able to recognize if the parameter is a template or just one value.

MrAnno: Both ideas sound good to me. I guess both can be implemented later without breaking anything, so I don't want to block this PR.

modules/dbparser/groupingby.c
MrAnno: @fekete-robert

lbudai:
@bazsi : do you plan to update this PR? I mean, I think it's ready to merge as-is... but if you want to change something then we don't merge now.

fekete-robert:
Thanks, I added it to my todo list

bazsi:
I'll try to add template support to sort-key() in a way that retains the performance characteristics of the current solution, while using a consistent syntax.

e.g. sort-key('$3') instead of sort-key('3')

so don't merge this yet.

lbudai:
@bazsi: ok

modules/dbparser/dbparser-grammar.ym
bazsi: I have now implemented template support for sort-key(), in a way that "trivial" templates are
using a faster code-path.

bazsi:
I have finished supporting the template syntax in sort-key() without sacrificing performance, so this should be good to go in.

kira-syslogng:
Build FAILURE

bazsi:
This pull request **introduces 3 alerts** when merging 8b5c768bb75c659d79285c5d7affbb157ef36b52 into bbcb24f76c5e5d45ae79669990da91c5169e6449 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-1b5fbb8b7a72e0f6b92233d532915a788b113074)

**new alerts:**

---

kira-syslogng:
Build SUCCESS

lib/template/templates.c
```diff
@@ -28,11 +28,90 @@
...
+  LogTemplateElem *e = (LogTemplateElem *) self->compiled_template->data;
+
+  switch (e->type)
+    {
+    case LTE_FUNC:
+      /* functions are never trivial */
+      return FALSE;
+    case LTE_MACRO:
+      /* Macros are trivial if they only contain a text but not a real
+       * macro.  Empty strings are represented this way.  */
```
MrAnno: I don't fully understand this part. Could you explain for me what "only contain a text but not a real macro" means?

bazsi: A template is represented by a list of LogTemplateElem structs. Each one is a pair of a (literal text, dynamic reference).

Formatting works by first appending the literal text, then expanding the dynamic reference.

The reason this was combined was performance (half the items on the list than if we'd split the pairs), but there's a good chance this representation could be improved without to much performance consequences. E.g. having twice the elements but in a linear array rather than a linked list.

kira-syslogng:
Build SUCCESS

lib/template/templates.c
MrAnno: Oh I see, thank you.

PR #2704, opened 2019-05-02
Dynamic fc

lbudai:
First PR that implements dynamic window functionality.

# Dynamic window

Motivation/problem to be solved: 
 * when a network connection has a high number of clients
 * AND only a small subset of the active clients sends messages at a high rate AND
 * there is a limited amount of memory, 
then it would be nice from syslog-ng if it could somehow dynamically re-calculate window sizes (periodically during runtime).

The proposed solution is dynamic-window.

Key points during implementation:
 * 100% backward compatibility when someone don't want to use
 * default is static only
 * dynamic windowing performance should be very close to the static case
 * starvation is not an option, so a static window is always guaranteed (`iw-size`/`max-connections`)

Keep in mind that main goal of this PR is to implement the 'framework' of dynamic windowing and to show an example for a simple heuristic.
In a future PR we want to make it easier to use and also want to provide more/better heuristic.

How it works from users point of view?
* "Main" option
   * `dynamic-window-size(N)` : enable and set dynamic window
* Tuning parameters
   * `dynamic-window-stats-freq(x milliseconds)` : every x milliseconds syslog-ng check and update window stats
   * `dynamic-window-realloc-ticks(k)`           : after k window-stats-freq() dynamic window will be reallocated
   * `dynamic-window-decrease-threshold(y %)`    : when (free/used) * 100 > y then syslog-ng decrease the dynamic window
   * `dynamic-window-increase-threshold(x %)`    : when (free/used) * 100 < x then syslog-ng increase the dynamic window

Typical network source with dynamic window could be:

```
@version: 3.20
options {
 min-iw-size-per-reader(1);
 stats_level(2);
};

source s_network_dynamic {
  network(ip(0.0.0.0)
    port(61001)
    log_fetch_limit(1000)
    log_iw_size(10000)
    max_connections(10000)
    dynamic-window-size(100000)
    dynamic-window-stats-freq(0.1)
    dynamic-window-realloc-ticks(10)
    dynamic-window-decrease-threshold(50)
    dynamic-window-increase-threshold(20)
    flags(no-parse) 
  );
};
```

In this case, the static window per connection is `1` (the `min-iw-size-per-reader(1)` makes it possible to set such a small window size, in the future this option will be enabled by default when `dynamic-window-size()` is used).
When 50% of the available window is free for a connection, syslog-ng will decrease the window (but never goes under the static window size, which is 1 in this example).
When only 20% of the available window is free (or, in other words, if 80% of the available window is used) for a connection, syslog-ng will try to increase the window.

### Some UML diagrams (more details will be shared little bit later on diagrams, for reviewers):
dynamic-fc-class-dia

dynamic-fc-seq

### DIFF
--- a/lib/dynamic-window-counter.c
+++ b/lib/dynamic-window-counter.c
@@ -30,6 +30,8 @@ DynamicWindowCounter *
 dynamic_window_counter_new(void)
 {
   DynamicWindowCounter *self = g_new0(DynamicWindowCounter, 1);
+  g_atomic_counter_set(&self->ref_cnt, 1);
+
   return self;
 }
 
@@ -38,12 +40,25 @@ void dynamic_window_counter_init(DynamicWindowCounter *self)
   self->window = self->iw_size;
 }
 
-void dynamic_window_counter_free(DynamicWindowCounter *self)
+DynamicWindowCounter *
+dynamic_window_counter_ref(DynamicWindowCounter *self)
 {
-  if (!self)
-    return;
+  g_assert(!self || g_atomic_counter_get(&self->ref_cnt) > 0);
 
-  g_free(self);
+  if (self)
+    g_atomic_counter_inc(&self->ref_cnt);
+
+  return self;
+}
+
+void dynamic_window_counter_unref(DynamicWindowCounter *self)
+{
+  g_assert(!self || g_atomic_counter_get(&self->ref_cnt));
+
+  if (self && g_atomic_counter_dec_and_test(&self->ref_cnt))
+    {
+      g_free(self);
+    }
 }
 
 void dynamic_window_counter_set_iw_size(DynamicWindowCounter *self, gsize iw_size)
--- a/lib/dynamic-window-counter.c
+++ b/lib/dynamic-window-counter.c
@@ -45,3 +45,8 @@ void dynamic_window_counter_free(DynamicWindowCounter *self)
 
   g_free(self);
 }
+
+void dynamic_window_counter_set_iw_size(DynamicWindowCounter *self, gsize iw_size)
+{
+  self->iw_size = iw_size;
+}
--- a/lib/dynamic-window-counter.c
+++ b/lib/dynamic-window-counter.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2002-2019 One Identity
  * Copyright (c) 2019 Laszlo Budai <laszlo.budai@balabit.com>
+ * Copyright (c) 2019 László Várady <laszlo.varady@balabit.com>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -22,21 +23,25 @@
  *
  */
 
-#ifndef DYNAMIC_WINDOW_COUNTER_H_INCLUDED
-#define DYNAMIC_WINDOW_COUNTER_H_INCLUDED
+#include <syslog-ng.h>
+#include "dynamic-window-counter.h"
 
-#include "syslog-ng.h"
-
-typedef struct _DynamicWindowCounter DynamicWindowCounter;
-
-struct _DynamicWindowCounter
+DynamicWindowCounter *
+dynamic_window_counter_new(void)
 {
-  gsize iw_size;
-  gsize window;
-};
+  DynamicWindowCounter *self = g_new0(DynamicWindowCounter, 1);
+  return self;
+}
 
-DynamicWindowCounter *dynamic_window_counter_new(void);
-void dynamic_window_counter_init(DynamicWindowCounter *self);
-void dynamic_window_counter_free(DynamicWindowCounter *self);
+void dynamic_window_counter_init(DynamicWindowCounter *self)
+{
+  self->window = self->iw_size;
+}
 
-#endif
+void dynamic_window_counter_free(DynamicWindowCounter *self)
+{
+  if (!self)
+    return;
+
+  g_free(self);
+}
--- a/lib/dynamic-window-counter.c
+++ b/lib/dynamic-window-counter.c
@@ -27,10 +27,12 @@
 #include "dynamic-window-counter.h"
 
 DynamicWindowCounter *
-dynamic_window_counter_new(void)
+dynamic_window_counter_new(gsize iw_size)
 {
   DynamicWindowCounter *self = g_new0(DynamicWindowCounter, 1);
   g_atomic_counter_set(&self->ref_cnt, 1);
+
+  self->iw_size = iw_size;
 
   return self;
 }
@@ -59,11 +61,6 @@ void dynamic_window_counter_unref(DynamicWindowCounter *self)
     {
       g_free(self);
     }
-}
-
-void dynamic_window_counter_set_iw_size(DynamicWindowCounter *self, gsize iw_size)
-{
-  self->iw_size = iw_size;
 }
 
 gsize
--- a/lib/dynamic-window-counter.c
+++ b/lib/dynamic-window-counter.c
@@ -50,3 +50,18 @@ void dynamic_window_counter_set_iw_size(DynamicWindowCounter *self, gsize iw_siz
 {
   self->iw_size = iw_size;
 }
+
+gsize
+dynamic_window_counter_request(DynamicWindowCounter *self, gsize requested_size)
+{
+  gsize offered = MIN(self->window, requested_size);
+  self->window -= offered;
+
+  return offered;
+}
+
+void dynamic_window_counter_release(DynamicWindowCounter *self, gsize release_size)
+{
+  self->window += release_size;
+  g_assert(self->window <= self->iw_size);
+}
--- a/lib/logsource.c
+++ b/lib/logsource.c
@@ -183,9 +183,9 @@ log_source_enable_dynamic_window(LogSource *self, DynamicWindowCounter *window_c
 void
 log_source_dynamic_window_update_statistics(LogSource *self)
 {
-  dynamic_window_stat_update(&self->dynamic_window, window_size_counter_get(&self->window_size, NULL));
+  dynamic_window_stat_update(&self->dynamic_window.stat, window_size_counter_get(&self->window_size, NULL));
   msg_trace("Updating dynamic window statistic", evt_tag_int("avg window size",
-                                                             dynamic_window_stat_get_avg(&self->dynamic_window)));
+                                                             dynamic_window_stat_get_avg(&self->dynamic_window.stat)));
 }
 
 static inline void
@@ -247,14 +247,14 @@ log_source_dynamic_window_realloc(LogSource *self)
    * only incrementation is possible by destination threads */
 
   /* TODO: add abstraction for heuristics */
-  gsize free_avg = dynamic_window_stat_get_avg(&self->dynamic_window);
+  gsize free_avg = dynamic_window_stat_get_avg(&self->dynamic_window.stat);
   if (free_avg > self->full_window_size * (self->options->dynamic_window_decrease_threshold / 100.0f))
     _decrease_window(self);
 
   if (free_avg < self->full_window_size * (self->options->dynamic_window_increase_threshold / 100.0f))
     _increase_window(self);
 
-  dynamic_window_stat_reset(&self->dynamic_window);
+  dynamic_window_stat_reset(&self->dynamic_window.stat);
 }
 
 void
--- a/lib/logsource.c
+++ b/lib/logsource.c
@@ -175,9 +175,9 @@ log_source_flow_control_suspend(LogSource *self)
 }
 
 void
-log_source_enable_dynamic_window(LogSource *self, DynamicWindow *dynamic_window)
+log_source_enable_dynamic_window(LogSource *self, DynamicWindowCounter *window_ctr)
 {
-  if (self->pos_tracked && dynamic_window)
+  if (self->pos_tracked && window_ctr)
     {
       msg_warning("WARNING: dynamic window size control works only with non-position tracking sources"
                   "Falling back to static window size (dynamic won't be used).",
@@ -185,7 +185,7 @@ log_source_enable_dynamic_window(LogSource *self, DynamicWindow *dynamic_window)
       return;
     }
 
-  self->dynamic_window = dynamic_window;
+  dynamic_window_set_counter(&self->dynamic_window, window_ctr);
 }
 
 void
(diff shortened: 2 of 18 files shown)


bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging d19ec6642222f2e9c302c72719c66d935fc00b3f into 616b0ecdb5fe897913c93d7e8698b9cbb48128f9 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-7cef9170128720572a645330f534b4e61791a7f3)

**new alerts:**

**fixed alerts:**

---

kira-syslogng:
Build SUCCESS

modules/afsocket/afsocket-grammar.ym
```diff
@@ -386,6 +391,12 @@ source_afsocket_stream_params
 	: KW_KEEP_ALIVE '(' yesno ')'		{ afsocket_sd_set_keep_alive(last_driver, $3); }
 	| KW_MAX_CONNECTIONS '(' positive_integer ')'	 { afsocket_sd_set_max_connections(last_driver, $3); }
 	| KW_LISTEN_BACKLOG '(' positive_integer ')'	{ afsocket_sd_set_listen_backlog(last_driver, $3); }
+	| KW_DYNAMIC_WINDOW_SIZE '(' nonnegative_integer ')' { afsocket_sd_set_dynamic_window_size(last_driver, $3); }
+  | KW_DYNAMIC_WINDOW_STATS_FREQ '(' nonnegative_integer ')' { afsocket_sd_set_dynamic_window_stats_freq(last_driver, $3); }
+  | KW_DYNAMIC_WINDOW_STATS_FREQ '(' LL_FLOAT ')' { afsocket_sd_set_dynamic_window_stats_freq(last_driver, $3); }
+  | KW_DYNAMIC_WINDOW_REALLOC_TICKS '(' nonnegative_integer ')' { afsocket_sd_set_dynamic_window_realloc_ticks(last_driver, $3); }
+  | KW_DYNAMIC_WINDOW_INCREASE_THRESHOLD '(' nonnegative_integer ')' { last_reader_options->super.dynamic_window_increase_threshold = $3; }
```
furiel: These are measured in percents. Do we want to allow zero and 100?

modules/afsocket/afsocket-source.c
```diff
@@ -703,6 +813,9 @@ afsocket_sd_init_method(LogPipe *s)
 {
   AFSocketSourceDriver *self = (AFSocketSourceDriver *) s;
 
+  if (self->dynamic_window_ctr)
+    dynamic_window_counter_init(self->dynamic_window_ctr);
```
Kokan: In case of reload, this resets `window` value to `iw_size`, which is going to cause "overflow" when *dynamic* window is released.

lib/logsource.c
```diff
@@ -174,6 +174,95 @@ log_source_flow_control_suspend(LogSource *self)
...
+      new_full_window_size = self->full_window_size - subtrahend;
+    }
+
+  window_size_counter_sub(&self->window_size, subtrahend, NULL);
+
+  msg_trace("Decreasing dynamic window", evt_tag_int("previous_window", self->full_window_size),
+            evt_tag_int("new_window", new_full_window_size), log_pipe_location_tag(&self->super));
+
+  self->full_window_size = new_full_window_size;
+  dynamic_window_release(&self->dynamic_window, subtrahend);
```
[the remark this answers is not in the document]
MrAnno: `subtrahend` should not contain static slots, we don't go under the static window: `MAX(self->options->init_window_size, self->full_window_size / 2);`

modules/afsocket/afsocket-source.c
MrAnno: Nice catch, we've seen this too while running reload tests.

lib/logsource.c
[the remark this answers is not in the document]
Kokan: Sorry, I did the maths again and checks out.

lib/logsource.c
```diff
@@ -174,6 +174,95 @@ log_source_flow_control_suspend(LogSource *self)
...
+
+  dynamic_window_set_counter(&self->dynamic_window, window_ctr);
+}
+
+void
+log_source_dynamic_window_update_statistics(LogSource *self)
+{
+  dynamic_window_stat_update(&self->dynamic_window, window_size_counter_get(&self->window_size, NULL));
+  msg_trace("Updating dynamic window statistic", evt_tag_int("avg window size",
+                                                             dynamic_window_stat_get_avg(&self->dynamic_window)));
```
bazsi: please use underscores in the message tags, this breaks our logging fomrat.

lib/dynamic-window-counter.c
```diff
@@ -0,0 +1,67 @@
...
+
+DynamicWindowCounter *
+dynamic_window_counter_new(void)
+{
+  DynamicWindowCounter *self = g_new0(DynamicWindowCounter, 1);
+  return self;
+}
+
+void dynamic_window_counter_init(DynamicWindowCounter *self)
+{
```
bazsi: please have the return type on a separate line, I regularly grep for "^function_name" which wouldn't work here.

lib/logsource.c
```diff
@@ -174,6 +174,95 @@ log_source_flow_control_suspend(LogSource *self)
...
+   * only incrementation is possible by destination threads */
+
+  /* TODO: add abstraction for heuristics */
+  gsize free_avg = dynamic_window_stat_get_avg(&self->dynamic_window);
+  if (free_avg > self->full_window_size * (self->options->dynamic_window_decrease_threshold / 100.0f))
+    _decrease_window(self);
+
+  if (free_avg < self->full_window_size * (self->options->dynamic_window_increase_threshold / 100.0f))
+    _increase_window(self);
+
```
bazsi: I would try to avoid float arithmetic here, at least if this is a fast path, which I cannot judge at this point.

MrAnno: Fortunately, this is not a fast path, but we can get rid of the floating-point arithmetic.

bazsi: Am I right to assume that the reason we don't need locking around window allocation/release is that we always do this in the main thread? It seems our timer is in the main thread. And we either call this path if working is FALSE or we defer this call until the fetch() function finishes. In any case, we'll only do dynamic window updates from the main thread, thus we don't need any locking.

Nice.

MrAnno: Please note that the actual heuristic has problems and probably has to be extended/re-implemented.

MrAnno: > Am I right to assume that the reason we don't need locking around window allocation/release is that we always do this in the main thread?

That is correct, I really liked this too. :)

lib/logsource.c
```diff
@@ -174,6 +174,95 @@ log_source_flow_control_suspend(LogSource *self)
...
+static void
+_release_dynamic_window(LogSource *self)
+{
+  g_assert(self->ack_tracker == NULL);
+
+  gsize dynamic_part = self->full_window_size - self->options->init_window_size;
+  msg_trace("Releasing dynamic part of the window", evt_tag_int("dynamic_window_to_be_released", dynamic_part),
+            log_pipe_location_tag(&self->super));
+  self->full_window_size -= dynamic_part;
```
bazsi: could you make the log tag names consistent? so in all invocation, we have the same names for the same stuff.

sometimes it's previous_window/new_window, now it's dynamic_window_to_be_released?

Also, if I may ask, the evt_tag() functions should be on their own lines, I find them much easier to read that way. Thanks. This last one is not a blocker at all, simply my preference.

lib/dynamic-window-counter.c
```diff
@@ -0,0 +1,67 @@
...
+  if (!self)
+    return;
+
+  g_free(self);
+}
+
+void dynamic_window_counter_set_iw_size(DynamicWindowCounter *self, gsize iw_size)
+{
+  self->iw_size = iw_size;
+}
```
bazsi: can we use a different name for iw_size? this is very similar to what we have for the static case. at the very least include the qualifier "dynamic" in the name please. the configuration interface uses the term "dynamic-window-size()" and this is where that value gets stored. would be great to have that name here somewhere.

modules/afsocket/afsocket-source.h
```diff
@@ -40,8 +41,13 @@ struct _AFSocketSourceDriver
...
           window_size_initialized:1;
   struct iv_fd listen_fd;
+  struct iv_timer dynamic_window_timer;
+  gsize dynamic_window_timer_tick;
+  gfloat dynamic_window_stats_freq;
+  gint dynamic_window_realloc_ticks;
   gint fd;
   LogReaderOptions reader_options;
+  DynamicWindowCounter *dynamic_window_ctr;
   LogProtoServerFactory *proto_factory;
```
bazsi: I think it would make sense to move all related attributes here in the same struct for easy embedding into call-sites. I expect the very same could be used in wildcard-file() sources as well.

lib/dynamic-window-counter.c
MrAnno: I agree, we called this class `DynamicWindowPool` on the diagrams, we probably should use a similar name in code and find a better name for its `iw_size` field too.

modules/afsocket/afsocket-source.c
```diff
@@ -287,19 +288,19 @@ afsocket_sd_set_dynamic_window_size(LogDriver *s, gint dynamic_window_size)
 }
 
 void
-afsocket_sd_set_dynamic_window_stats_freq(LogDriver *s, gint stats_freq)
+afsocket_sd_set_dynamic_window_stats_freq(LogDriver *s, gfloat stats_freq)
 {
   AFSocketSourceDriver *self = (AFSocketSourceDriver *) s;
 
-  self->dynamic_window_stats_freq = stats_freq;
+  self->dynamic_window_stats_freq = stats_freq * 1000.0f;
```
alltilla: By `LL_FLOAT`, negative numbers can be set here, so I think we should give a stricter restriction in the grammar file, so it throws error at config load time.

```
@version: 3.20
options {
 min-iw-size-per-reader(1);
 stats_level(2);
};

source s_network_dynamic {
  network(ip(0.0.0.0)
    port(61001)
    log_fetch_limit(1000)
    log_iw_size(10000)
    max_connections(10000)
    dynamic-window-size(100000)
    dynamic-window-stats-freq(-0.1)
    dynamic-window-realloc-ticks(10)
    dynamic-window-decrease-threshold(50)
    dynamic-window-increase-threshold(20)
    flags(no-parse) 
  );
};

destination d_local {
	file("/tmp/ne_test.log");
};

log {
	source(s_network_dynamic);
	destination(d_local);
};
```

```
[2019-05-06T10:46:58.080088] Accepting connections; addr='AF_INET(0.0.0.0:61001)'
[2019-05-06T10:46:58.080254] Running application hooks; hook='1'
[2019-05-06T10:46:58.080274] Running application hooks; hook='6'
[2019-05-06T10:46:58.080312] syslog-ng starting up; version='3.20.1.350.gd19ec66'
[2019-05-06T10:46:58.080333] Running application hooks; hook='2'
[2019-05-06T10:46:58.080378] Dynamic window timer elapsed; tick='1'
[2019-05-06T10:46:58.080399] Dynamic window timer elapsed; tick='2'
[2019-05-06T10:46:58.080404] Dynamic window timer elapsed; tick='3'
[2019-05-06T10:46:58.080408] Dynamic window timer elapsed; tick='4'
[2019-05-06T10:46:58.080413] Dynamic window timer elapsed; tick='5'
[2019-05-06T10:46:58.080419] Dynamic window timer elapsed; tick='6'
[2019-05-06T10:46:58.080425] Dynamic window timer elapsed; tick='7'
[2019-05-06T10:46:58.080430] Dynamic window timer elapsed; tick='8'
[2019-05-06T10:46:58.080435] Dynamic window timer elapsed; tick='9'
[2019-05-06T10:46:58.080439] Dynamic window timer elapsed; tick='10'
[2019-05-06T10:46:58.080443] Dynamic window timer elapsed; tick='1'
[2019-05-06T10:46:58.080447] Dynamic window timer elapsed; tick='2'
[2019-05-06T10:46:58.080451] Dynamic window timer elapsed; tick='3'
[2019-05-06T10:46:58.080455] Dynamic window timer elapsed; tick='4'
[2019-05-06T10:46:58.080459] Dynamic window timer elapsed; tick='5'
[2019-05-06T10:46:58.080463] Dynamic window timer elapsed; tick='6'
[2019-05-06T10:46:58.080467] Dynamic window timer elapsed; tick='7'
[2019-05-06T10:46:58.080471] Dynamic window timer elapsed; tick='8'
[2019-05-06T10:46:58.080475] Dynamic window timer elapsed; tick='9'
[2019-05-06T10:46:58.080478] Dynamic window timer elapsed; tick='10'
```

alltilla: We could make a `positive_float` type, which accepts `positive_integer` or `LL_FLOAT` with validation. This way we do not need to write the rule in 2 lines. Same can be done with nonnegative, etc...

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging ea50f1df280eeed70ae2927512555d9d5db574a7 into bbcb24f76c5e5d45ae79669990da91c5169e6449 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-4fc9ca4453fb45443561e6b1870b60cf7bbbba30)

**new alerts:**

**fixed alerts:**

---

kira-syslogng:
Build SUCCESS

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 1de26fbd98c7660d41decd7be05cce548036f841 into f107780c30649ef02c388e53b5b17ac80787c707 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-0f18be30d121c9d02533d290791f9af89f0c46d5)

**new alerts:**

**fixed alerts:**

---

kira-syslogng:
Build SUCCESS

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 81882b13af666e0128cf0b29ea053a180629a2ee into 47b3d2443844fc609bce25cf7a4ed38735c6aaba - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-2f987332d12ccf4fb9ce06696212b82aee3e8a5a)

**new alerts:**

**fixed alerts:**

---

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lbudai:
I'm closing this in favour of #2772 .

Thanks for the review notes, we will process them (some of them have been already processed).

PR #2709, opened 2019-05-03
Add support for usec precision when parsing time

smortex:
Python added a `%f` field ("Microsecond as a decimal number") in it's strptime() implementation to parse times with subsecond precision.

Borrow this idea and implement it in syslog-ng using the same field descriptor.

**TODO**:

- [X] Add support for `%f` field to parse microseconds;
- [X] Eat extra digits (i.e. nanosecond precision timestamp, while we only handle microseconds) [see request](https://github.com/balabit/syslog-ng/pull/2709#discussion_r280968591);
- [x] Move the time parsing code to `wall_clock_time_strptime()` [see comment](https://github.com/balabit/syslog-ng/pull/2709#discussion_r280968552)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

lib/timeutils/strptime-tz.c
```diff
@@ -349,6 +351,26 @@ strptime_with_tz(const char *buf, const char *fmt, struct tm *tm, long *tm_gmtof
...
+            {
+              *tm_usec *= 10;
+            }
+
+          bp = end;
+          LEGAL_ALT(ALT_O);
+          state |= S_USEC;
+        }
+        continue;
+
```
smortex: `astyle(1)` get very confused about this new code and insists in pretending that it should be indented that way :confused: As a result, the `continue` has the same indentation level as the `case` statements…

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

bazsi:
This is actually something I had on my shortlist to implement, and in exactly the same manner (using %f to follow Python's lead here). Thank you.

There's one more problem I originally wanted to solve: since the microseconds part is optional, the dot should be parsed as the part of that piece, and I know %f does not consume the dot, but maybe we should implement a %F which does.

An example:

```
timestamp: Apr 29 13:58:40.411
date-parser(format("%b %d %H:%M:%S%F"));
```
Hmm I just noticed that %F is already taken, maybe we could use a multi-letter variant, like "%.f"

what do you think?

lib/timeutils/strptime-tz.c
```diff
@@ -208,7 +210,7 @@ first_wday_of(int yr)
  * their struct tm. This is a slightly modified NetBSD strptime() with
  * some modifications and explicit zone related parameters. */
 char *
-strptime_with_tz(const char *buf, const char *fmt, struct tm *tm, long *tm_gmtoff, const char **tm_zone)
+strptime_with_tz(const char *buf, const char *fmt, struct tm *tm, long *tm_gmtoff, const char **tm_zone, int *tm_usec)
 {
```
bazsi: I increasingly have the feeling that this should be parsing into a WallClockTime instance, instead of littering its argument list with more and more values.

The only reason I didn't do this previously was that I tried to keep it as close to the original NetBSD implementation as possible. But we have forked this function anyway and I don't expect it to sync with upstream.

If that's the case we could possibly remove some code from it as well. This is not a blocking comment.

lib/timeutils/strptime-tz.c
bazsi: just add the continue inside the braces.

lib/timeutils/tests/test_wallclocktime.c
```diff
@@ -49,10 +49,31 @@ Test(wallclocktime, test_strptime_parses_broken_down_time)
...
+
+Test(wallclocktime, test_strptime_parses_overflowed_usec)
+{
+  WallClockTime wct = WALL_CLOCK_TIME_INIT;
+  gchar *end;
+
+  end = wall_clock_time_strptime(&wct, "%b %d %Y %H:%M:%S.%f", "Jan 16 2019 18:23:12.0123456");
+  cr_assert(*end == '6');
+  cr_expect(wct.wct_usec == 12345);
+}
```
bazsi: I think in this case we should consume the numbers but ignore them.

lib/timeutils/strptime-tz.c
smortex: Yeah, I share this view: I first hacked in the area that seems to be the right one, and was happy to discover that `strptime_with_tz()` was called in a single place :smile: I did not renamed the function (`strptime_with_tz_and_usec_and_probably_other_stuff_in_the_future()`), but I makes sense to move the code at the `wall_clock_time_strptime()` level.

Maybe we can solve all the minor glitches and feedback before adding a commit that just moves the code around without changing it?

lib/timeutils/tests/test_wallclocktime.c
smortex: Makes sense, I'll work on this.

kira-syslogng:
Build SUCCESS

smortex:
I amended my commit with style fixes, and added a new one to eat extra digit in the usec field as suggested.

-----

Regarding consuming the dot `.`, I happen to be working on this PR after PuppetDB has changed the way it logs timestamps:

* `2017-02-02 00:29:16,706` (old, 5.x I guess)
* `2019-05-04T21:55:46.989+02:00` (new 6.x)

As you can see, the old format seems to use a coma `,` for separating seconds and millisecond ; while the new format use a dot `.`.  So, as far as I am concerned, leaving responsibility of matching the right symbol outside the field seems more appropriate to me, and the complexity of a modifier does not worth it.  It also match what Python does with the `%f` field.

Maybe we can keep it that way for now, and rely on users feedback to determine if it should be extended it in the future?

lib/timeutils/strptime-tz.c
bazsi: ok, makes sense.

bazsi:
I think it makes sense, and there's one more place this could be addressed: introduce support for multiple format() arguments to date-parser, trying them in order and accepting the timestamp at first success.

For example:

```
date-parser(format(
     "%b %d %H:%M:%S.%f",
     "%b %d %H:%M:%S,%f"
     "%b %d %H:%M:%S",
    ));
```

This would handle either ',' or '.' and even with the fractions omitted. I have a relevant use-case in the cisco-parser:
```
        if {
            # timestamp from Cisco Unified Call Manager, example "Jun 14 11:57:27 PM.685 UTC"
            # NOTE: drops msecs and timezone
            filter { match('^[.*]?([A-Za-z]{3} [0-9 ]\d \d{2}:\d{2}:\d{2} ((AM)|(PM)))' value('1') flags(store-matches)); };
            parser { date-parser(format('%b %d %H:%M:%S %p') template("$1")); };
        } elif {
            # timestamp without AM/PM mark, example: "Apr 29 13:58:40.411"
            # NOTE: drops msecs and timezone

            filter { match('^[.*]?([A-Za-z]{3} [0-9 ]\d \d{2}:\d{2}:\d{2})' value('1') flags(store-matches)); };
            parser { date-parser(format('%b %d %H:%M:%S') template("$1")); };
        } else {
            # timestamp with year information, example: "Apr 29 2017 13:58:40.411"
            filter { match('^[.*]?([A-Za-z]{3} [0-9 ]\d \d{4} \d{2}:\d{2}:\d{2})' value('1') flags(store-matches)); };
            parser { date-parser(format('%b %d %Y %H:%M:%S') template("$1")); };
        };
```

Would you give this a go, even in a separate PR? That would be awesome.

smortex:
@bazsi I added a few commits to move the code around.  Is it what you expected?

Regarding adding support for multiple formats to `date-parser()`, it seems a good idea, but I will not be able to work on this topic right now, so I would recommend splitting this into a separate issue / PR.

kira-syslogng:
Build SUCCESS

lib/timeutils/wallclocktime.c
```diff
@@ -22,12 +22,144 @@
...
 #include "timeutils/unixtime.h"
-#include "timeutils/strptime-tz.h"
 #include "timeutils/cache.h"
 #include "timeutils/misc.h"
 
+#define __UNCONST(a)    ((void *)(unsigned long)(const void *)(a))
+
+#define TM_YEAR_BASE 1900
+#define TM_SUNDAY     0
+#define TM_MONDAY     1
```
bazsi: Please retain the copyright/license of the original code

lib/timeutils/wallclocktime.c
```diff
@@ -38,7 +170,575 @@ wall_clock_time_unset(WallClockTime *self)
...
+            }
+          else
+            {
+              ep = find_string(bp, &i, (const char *const *)tzname, NULL, 2);
+              if (ep != NULL)
+                {
+                  wct->tm.tm_isdst = i;
+#ifdef SYSLOG_NG_HAVE_TIMEZONE
+                  wct->wct_gmtoff = -(timezone);
+#endif
```
bazsi: We always have this field in wallclocktime.

lib/timeutils/wallclocktime.c
bazsi: Also, could you move this entire chunk towards the end of the file? I feel the main functions of the wall clock time struct are more important when approaching the file, and they get somewhat deemphasized with the strptime implementation or in front.

smortex: Ooops, I forgot to move the license!

I am not sure to understand exactly what you meant in the second comment, so implemented what I understood.  Please tell me if I missed your point :wink:

lib/timeutils/wallclocktime.c
smortex: Removed

kira-syslogng:
Build SUCCESS

smortex:
@bazsi, I am back from holidays :palm_tree: 

I squashed the related commits as requested and rebased my work on top of the master branch.

kira-syslogng:
Build SUCCESS

lib/timeutils/strptime-tz.c
```diff
@@ -349,6 +351,35 @@ strptime_with_tz(const char *buf, const char *fmt, struct tm *tm, long *tm_gmtof
...
+           * If we read less than 6 digits, we need to adjust the value:
+           * "012" was parsed as 12 but is 12000 us.
+           */
+          while (digits++ < 6)
+            {
+              *tm_usec *= 10;
+            }
+
+          bp = end;
+          LEGAL_ALT(ALT_O);
```
furiel: I am unfamiliar with these `LEGAL_ALT`, `ALT_O` concepts. What is this line used for?

smortex: D'oh, good catch, it's a mistake!  According to `strftime(3)`:

```
     %E* %O*                                                                    
           POSIX locale extensions.  The sequences %Ec %EC %Ex %EX %Ey %EY %Od  
           %Oe %OH %OI %Om %OM %OS %Ou %OU %OV %Ow %OW %Oy are supposed to      
           provide alternate representations.                                   

           Additionally %OB implemented to represent alternative months names
           (used standalone, without day mentioned).
```

`%Of` does not mean anything, and I should have written `LEGAL_AT(0)` to ensure we are _not_ parsing `%Of` nor `%Ef`…  I'll amend the commit accordingly.

smortex: Done, thanks!

kira-syslogng:
Build SUCCESS

lib/timeutils/wallclocktime.c
smortex: I guess the conditional was here not to avoid setting a value in a field that may not exist, but because `timezone` is something not portable being defined as `extern long int timezone;` on a Linux system and `char *timezone(int zone, int dst);` on FreeBSD…

The build currently fails on FreeBSD with:
```
  CC       lib/timeutils/libsyslog_ng_la-wallclocktime.lo
lib/timeutils/wallclocktime.c:551:37: error: invalid argument type 'char *(*)(int, int)' to unary expression
                  wct->wct_gmtoff = -(timezone);
                                    ^~~~~~~~~~~
1 error generated.
```

Issue #2713, opened 2019-05-07
RPM should have feature flags for optional components


duckfez:
# syslog-ng

## Version of syslog-ng

```
syslog-ng 3 (3.20.1)
Config version: 3.20
Installer-Version: 3.20.1
Revision:
Compile-Date: May  7 2019 00:17:47
Module-Directory: //usr/lib64/syslog-ng
Module-Path: //usr/lib64/syslog-ng
Include-Path: /usr/share/syslog-ng/include
Available-Modules: add-contextual-data,afamqp,affile,afprog,afsocket,afstomp,afuser,appmodel,basicfuncs,cef,confgen,cryptofuncs,csvparser,date,dbparser,disk-buffer,examples,graphite,hook-commands,json-plugin,kvformat,linux-kmsg-format,map-value-pairs,pseudofile,sdjournal,snmptrapd-parser,stardate,syslogformat,system-source,tags-parser,tfgetent,xml
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: on
```

## Platform

RHEL 7.6

# Issue

I want to be able to build the syslog-ng RPM without Java or DBI dependencies.  My environment does not require these, but the RPM lacks options for enabling / disabling these components.

MrAnno:
Implemented in #2724.

PR #2714, opened 2019-05-07
RPM:  Add feature flags for Java and DBI components

duckfez:
Fix for https://github.com/balabit/syslog-ng/issues/2713

Signed-off-by: Duane Waddle <duane.waddle@defpoint.com>

### DIFF
--- a/packaging/rhel/syslog-ng.spec
+++ b/packaging/rhel/syslog-ng.spec
@@ -1,5 +1,8 @@
 %global ivykis_ver 0.36.1
 
+%define enable_java 1
+%define enable_dbi 1
+
 Name: syslog-ng
 Version: 3.21.1
 Release: 1%{?dist}
@@ -22,7 +25,9 @@ BuildRequires: glib2-devel >= 2.10.1
 BuildRequires: ivykis-devel >= %{ivykis_ver}
 BuildRequires: json-c-devel
 BuildRequires: libcap-devel
+%if 0%{?enable_dbi}
 BuildRequires: libdbi-devel
+%endif
 BuildRequires: libnet-devel
 BuildRequires: openssl-devel
 BuildRequires: pcre-devel >= 6.1
@@ -33,9 +38,13 @@ BuildRequires: systemd-devel
 BuildRequires: hiredis-devel
 BuildRequires: riemann-c-client-devel
 BuildRequires: python-devel
+
+%if 0%{?enable_java}
 BuildRequires: java-devel
 #BuildRequires: gradle    # Upstream version installed manually, discussed in: https://github.com/balabit/syslog-ng/issues/2262
 BuildRequires: syslog-ng-java-deps
+%endif
+
 BuildRequires: libcurl-devel
 BuildRequires: cyrus-sasl-devel
 BuildRequires: libmaxminddb-devel
@@ -77,7 +86,7 @@ Key features:
  * hand on messages for further processing using message queues (like
    AMQP), files or databases (like PostgreSQL or MongoDB).
 
-
+%if 0%{?enable_dbi}
 %package libdbi
 Summary: libdbi support for %{name}
 Group: Development/Libraries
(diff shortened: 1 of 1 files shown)


kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

furiel:
Thank you for the PR! I agree it is a reasonable request to be able to install syslog-ng with partial features. I am just not sure if this is the right approach.

With deb packaging, we build all available features of syslog-ng, but the individual parts are separated into different packages, like syslog-ng-core, and syslog-ng-mod-java, etc. This way, users can install the full feature set, but they can also use a subset by selectively installing packages.

The best would be if there were similar packages for rpm as in case of the debian packages. We already provide a convenient way to generate rpm (`$ dbld/rules rpm`), that will start an environment with every dependency available, and generate a monolithic package. The same command could generate individual packages too, which would support your use case.

I think the upstream should contain an rpm spec file that is best for general usage. The proposed approach is more like a specialization for an environment, and with that a wrong direction of dependency appears: the general solution depends on a specialization. So I would not merge the proposed version as it is.

duckfez:
The current-in-master RPM spec file does make the different packages already.  Or, at least the one I patched does.  (Is there a different spec file that `dbld/rules` uses?)  So in general, the ability to install specific sub-components was already there.  What I added was the ability to pick whether some sub-components are actually built at build-time.  

The whole PR began in our environment as basically, "I don't need the java stuff, and things like the Gradle build dependency is a pain in the neck so I'd rather just not even try to compile it!"

What I did is really a half-measure only dealing with the DBI and Java sub-packages.  But, I think making all of the other optional components changeable via build-time feature flags would be further in the direction you don't want to go.

furiel:
You are right. Overlook from my side. The current rpm spec file is not monolithic. Also, the current spec file already contains an `0%{?rhel}` compile switch in similar fashion. The purpose is similar: on some platforms the current spec file does not want to bother bringing dependencies. Compared to the rhel macro, your version with the per-package excludes feels cleaner. In short, I am convinced.

Let me tag @czanik for review and opinion, as he is also using these same spec files.

PS: I just tried dbld/rules rpm, and it is currently broken unfortunately. I will open an issue for it.

bazsi:
I agree with the intention too. As I see, rpm has conditional builds supported: https://rpm.org/user_doc/conditional_builds.html

May I ask why are you not using that feature?

Thanks

duckfez:
> May I ask why are you not using that feature?

@bazsi I had no idea that feature existed.  Assuming it's available in older versions of RPM, this is clearly a much better approach.  I will rework my patch to use this.  How old of an RPM-based distro do you think is reasonable to support?   My personal interests are only in Centos 7 and up, but am willing to test with an older version of RPM if you want.

Kokan:
@duckfez I think centos:6 as of now is reasonable. (at least till 2020 something)

So please test it with centos:6 as well.

bazsi:
in the meantime i worked on this on the dbld-kira-build branch, its also
opened as a pull request.

I covered Java and libdbi myself, so at this point I don't think there's
anything else left to do, apart from testing.

I am not yet finished with the branch, but centos-7 should work just fine.

duckfez:
@bazsi so should I abandon work on this then?

bazsi:
Yes, I think the changes in #2724 will resolve your problem as well. I have the following build options:

%bcond_with python3
%bcond_without python2
%bcond_without sql
%bcond_without mongodb
%bcond_without systemd
%bcond_without redis
%bcond_without riemann
%bcond_without maxminddb
%bcond_without amqp
%bcond_without java
%bcond_without kafka
%bcond_without snmpdest

by default we'd try to compile all this optional features, but you can disable them using the `--without <whatever>` option to rpmbuild.

PR #2727, opened 2019-05-14
[WIP] Travis: Move bash logic from .travis.yml into bash script

mitzkia:
The main goal is to move the bash logic (described in section "script") from .travis.yml into a separate bash file(s): Imho this is the recommended way to use bash logic in .travis.yml.

The current working mode (adding bash logic to .travis.yml) is not so debug friendly, and after found a place where one of the shell line in .travis.yml did not propagated back the error to the Travis this solution even more needed.

Expected result from this PR:
- better reading of .travis.yml
- exclude the unpropagated bash returns
- possible to run scripts from dev machine

Todo:
- [ ] Decide how to use bash commands (Travis way (with "-") or from executable bash file)
- [ ] Deal with "--keep-going"
- [ ] Cleanup bash logic from .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,34 +80,7 @@ before_script:
       "
   - ./configure $CONFIGURE_FLAGS
 script:
-  - . tests/build-log-cflags-propagation.sh;
-    if [ "$CC" = "gcc" ]; then
-      export DISTCHECK_CONFIGURE_FLAGS="$CONFIGURE_FLAGS";
-      exec_prop_check "make distcheck -j 3 V=1 --keep-going" ||
-      {
-        S=$?;
-        make V=1 distcheck;
-        find . -name test-suite.log | xargs cat;
-        return $S;
-      };
-      find . -name test-suite.log | xargs cat;
-    else
-      make --keep-going -j $(nproc);
-      S=$?;
-      if [ "$S" = "0" ]; then
-        make install
-        . scripts/get-libjvm-path.sh || return $?;
-        export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$JNI_LIBDIR";
-        make func-test V=1;
-        make pytest-self-check;
-        make pytest-check;
-      elif [ "$S" = "42" ]; then
-        return $S;
-      else
-        make V=1 install;
-        return $S;
-      fi;
-    fi
+  - tests/travis-common-script.sh
 compiler:
   - gcc
   - clang
--- a/tests/travis-common-script.sh
+++ b/tests/travis-common-script.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+set -e
+
+function main() {
+    . tests/build-log-cflags-propagation.sh;
+    if [ "$CC" = "gcc" ]; then
+      export DISTCHECK_CONFIGURE_FLAGS="$CONFIGURE_FLAGS";
+      exec_prop_check "make distcheck -j 3 V=1 --keep-going";
+      find . -name test-suite.log | xargs cat;
+    else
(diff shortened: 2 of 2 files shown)


kira-syslogng:
Build SUCCESS

tests/travis-common-script.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -e
```
furiel: I like this change (this is a bugfix), but this also breaks the current functionality. The current code depends on that the script does not exit early even in case of error: for example `S=$?`.

This should be resolved somehow.

mitzkia: I think this is not necessary a bugfix. 
I can change my opinion, but for this I need to understand why we did not want to return early if there is an error (return with not 0), imho this can be a real bug.

.travis.yml
```diff
@@ -80,34 +80,7 @@ before_script:
...
-        make pytest-self-check;
-        make pytest-check;
-      elif [ "$S" = "42" ]; then
-        return $S;
-      else
-        make V=1 install;
-        return $S;
-      fi;
-    fi
+  - tests/travis-common-script.sh
```
MrAnno: You have to export `CONFIGURE_FLAGS` before executing a child process that uses it.

tests/travis-common-script.sh
```diff
@@ -0,0 +1,29 @@
...
+        make pytest-self-check;
+        make pytest-check;
+      elif [ "$S" = "42" ]; then
+        return $S;
+      else
+        make V=1 install;
+        return $S;
+      fi;
+    fi
+}
```
MrAnno: Please do not hide this complexity, it does not help IMHO.

We should clean up this mess.
In my opinion, our build steps should be listed directly in travis.yml, in a simple and sequential way, such as:

```

gcc-job:
- . tests/build-log-cflags-propagation.sh;
  exec_prop_check "make distcheck -j 3 V=1 --keep-going";
- print_test_logs

clang-job:
- make --keep-going -j $(nproc);
- make install
- setup_libjvm
- make func-test V=1
- make pytest-self-check
- make pytest-check;
```

Utility functions (which are not important build steps), such as `print_test_logs` and `setup_libjvm` can be placed into separate bash files. This way, an "executable build documentation" is what you get, and you only hide the non-important parts.

mitzkia: Thanks for the review and a comment. My purpose with this change was the following:
if we can move the bash logic from .travis.yml into a well named bash script file we will be able to run it on our machine as well (if there are any problem during the Travis run and we need to debug something).

.travis.yml
mitzkia: +1, Thanks for the review and a comment.

mitzkia:
After discussing this topic with @MrAnno I have updated the "Todo" list of this PR. 
We have agreed in that we would like to clean the bash logic in .travis.yml. 
We can handle this in two ways:
- using the Travis way: 
  - commands remains in .travis.yml 
  - each command goes into separate lines with "-" marks
  - benefits: main commands are in one place + easier to read .travis.yml
- script files:
  - move all bash logics into separate bash executables
  - 'script' sections will contains only an execution of created bash files
  - benefits: easier to read .travis.yml + created bash files can be executed outside of travis


Other tasks:
  - eliminate 'if' statements like: gcc and lang and create separate logic for them
  - separate non important utility functions from other functions (separate bash files)

furiel:
We discussed with @mitzkia that I am taking over this task. Closing this PR in favor of https://github.com/balabit/syslog-ng/pull/2809

Issue #2466, opened 2018-12-14
segfault in 3.18.1+ latest master


soubhik-c:
# syslog-ng
   * Consistently getting a segfault with http and load-balancing in the worker. We are ingesting into elasticsearch using native http. 
   * This bug is being hit after removing batch-timeout which essentially was doing reinit to the destination frequently & leaking memory (related ticket #2462). 
   * With this change, reaping of connections went down drastically (from ~1600 to 41 times in 10 hrs) and I was hoping this race condition won't get hit.

source revision number: 
```
commit 82a7a012353143314d8482b7f249e56367a4da59 (origin/master, origin/HEAD, master)
Merge: 42789ba97 0701ca590
Author: László Várady <3130044+MrAnno@users.noreply.github.com>
Date:   Thu Dec 6 16:12:33 2018 +0100
```
  
## Version of syslog-ng

```
syslog-ng 3 (3.18.1)
Config version: 3.18
Installer-Version: 3.18.1
Revision: 
Compile-Date: Dec 12 2018 08:22:17
Module-Directory: /opt/syslog-ng-dbg/lib/syslog-ng
Module-Path: /opt/syslog-ng-dbg/lib/syslog-ng
Include-Path: /opt/syslog-ng-dbg/share/syslog-ng/include
Available-Modules: add-contextual-data,affile,afprog,afsocket,afstomp,afuser,appmodel,basicfuncs,cef,confgen,cryptofuncs,csvparser,date,dbparser,disk-buffer,examples,graphite,hook-commands,http,json-plugin,kvformat,linux-kmsg-format,map-value-pairs,mod-java,mod-python,pseudofile,sdjournal,snmptrapd-parser,stardate,syslogformat,system-source,tags-parser,tfgetent,xml
Enable-Debug: on
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: on
```

## Platform

CentOS 7 (with ubuntu-bionic inside docker container) 

## Debug bundle
[syslog.debun.syslogng-1.2018-12-12_12-34.521.tar.gz](https://github.com/balabit/syslog-ng/files/2679422/syslog.debun.syslogng-1.2018-12-12_12-34.521.tar.gz)

# Issue
The process terminates after a segmentation fault when run over a prolonged period of time ~10 hrs.

## Failure

dmesg logs the following error.
```
[...]

PR #2728, opened 2019-05-14
Elasticsearch java dest optimizations

soubhik-c:
All observation started from syslogng 3.16 and found to be accurate upto 3.20.

While scaling for 1 million msgs/sec ingestion into elasticsearch using syslog-ng elastic-v2 java destination, we found few bottlenecks when we ran experiments for several hours (upto 72 hrs of constant traffic). Our goal was to get a sustained input without any jitters while listening on multiple udp ports in syslog-ng. 

Primary issue we found in the plugin is: 
a) Too much garbage generation and inefficient json formatting using Gson. 
b) Another glaring issue was no matter how many threads we employed, things weren't scaling due to threads either spending its time in formatting log messages or waiting for http reply but both never happening at the same time. In other words, a destination thread was either formatting message or sending bulk message & formatting was costlier than sending and waiting.

Throughput was viewed in kibana discovery graph and was extremely wavy. We found ES wasn't pushed hard enough in the beginning.

We tried async httpcomponent and connection pool causing http timeout. Main issue was too many ESReject of the bulk message occurring when constant traffic flowed in for longer duration. 

Also, with some http message profiling, we noted message was getting copied multiple times in its path, firstly during formatting and then Jest api level string concatenation and lastly httpcomponent trying to push the message. Message fragmentation at the httpcomponent level wasn't getting avoided due to lack of length prefix provided and these message copy was causing too much gc pressure.

Initially we tried introducing multiple threads in the elastic-v2 plugin that will accumulate individual messages and format them into a bulk message & do send/receive. Idea was if enough threads are introduced, at the cost of higher context switches, we should be able to send continuous stream of bulk messages using http connection pool timeout settings appropriately. This didn't turned out very effective as CMS garbage collector still paused heavily the entire operation and G1GC in java-8 was showed instability on continuous load. Quickly realised upgrading to java-11 isn't easy due to jni level incompatibilities.

Second approach which is in this PR was to introduce a dedicated threadpool of finite size according to the hardware configuration. Speaking of h/w configuration, though syslogng is designed for 1 instance per host, our client had high end boxes (in aws terms xLarge with 32 cores and 168GB ram) for syslogng. Another imposition was to use docker containers. This changed the dynamics quite a bit because now multiple syslogng instances were contesting for the same h/w resources. Decoupling log formatting and bulk request send/receive showed promising results and supported our analysis of the bottleneck. Early closure of the response by reading response code '200' worked out great as it means all messages were success and doesn't really need parsing of the whole response for message ids.
Traffic overload could gracefully be handled by dropping messages and inferences from statistics was more accurate. Our usecase was to give preference to latest data over earlier minute data i.e. unlike syslogng using FIFO queue, we intend to use LIFO queue at a later time to guarantee message delivery of stale data during the lull period instead of trying to catchup with the queued messages at the very next second. For network security analytics, current second data (whatever we could capture) is more important than retrying message of last few seconds. However, eventually we need all of the data to be in the system for later analysis.

The connection pool at the plugin level increments dropped message count of the same destination statistics for now. In future, with the use of LIFO queue, guaranteed delivery can be achieved. The entire pool implementation can be disabled with concurrent-request(1) in the configuration.
 
We could achieve better stable input to ES when compared to native http performance introduced in syslogng. One primary disadvantage with native http plugin as of now is higher network payload mainly due to complete parsing of the http bulk message response from ES which is significant. That effectively slows down on the next request, consumes higher cpu and eats up network bandwidth as well as blocks ES side writer thread pool longer.

In ES, the bulk message hops around multiple servers in the current design. In the java plugin, we introduced facility to avoid such hoping without need for 'key' based shard allocation. Using key based shard routing mandates that key to be known during query time which becomes counter productive while asking adhoc queries. So, to achieve single hop bulk message processing in ES, a more viable design is to define one index per ES host and freeze the REST url's host ip and index name association. In java plugin, index name can't be embedded into url, we enabled it optionaly in this PR. This resulted in great overall stability in traffic ingestion as cross distribution of messages in ES side was all local among multiple ES instances hosted on a single machine. OTOH, syslogng could roundrobin bulk messages and keep the load shifting among ES instances and remained uneffected during index merges happening on the other end. So, ESReject bulk message from one ES server helped failing over to another without any involved changes.

This is just a brief description and if the community thinks this is good to pursue, I can provide with more details on the experiments (basically the numbers around tps/mem/cpu/network) and configuration details.

Few comments in another ticket with additional experiment details.
https://github.com/balabit/syslog-ng/issues/2466#issuecomment-456366108
https://github.com/balabit/syslog-ng/issues/2466#issuecomment-456872007

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,16 @@ dbld/rules.conf
 debian/build-tree/
 debian/*.log
 debian/autoreconf.*
+*.lo
+*grammar.c
+*grammar.h
+*grammar.y
+*.la
+*.o
+.dirstamp
+.libs
+modules/java-modules/*/gradle
+.gradle
+gradle*
+.idea
+**/out
--- a/Makefile.am
+++ b/Makefile.am
@@ -23,7 +23,8 @@
 
 SUBDIRS			=
 DIST_SUBDIRS		=
-AM_MAKEFLAGS		= --no-print-directory
+#AM_MAKEFLAGS		= --no-print-directory
+AM_MAKEFLAGS		= 
 AM_YFLAGS		= -Wno-yacc -Wno-other
 
 AM_TESTS_ENVIRONMENT	= top_srcdir="$(top_srcdir)"
--- a/modules/java-modules/build.gradle
+++ b/modules/java-modules/build.gradle
@@ -7,6 +7,7 @@ subprojects {
     apply plugin: 'idea'
 
     repositories {
+      mavenLocal()
         maven {
           url "/usr/lib/syslog-ng-java-compile-deps-mvn-repo"
         }
@@ -18,8 +19,8 @@ subprojects {
     }
 
     compileJava {
-        sourceCompatibility "1.7"
-        targetCompatibility "1.7"
+        sourceCompatibility "1.8"
+        targetCompatibility "1.8"
     }
 }
 
--- a/modules/java-modules/elastic-v2/src/main/java/io/searchbox/client/ESJestClientFactory.java
+++ b/modules/java-modules/elastic-v2/src/main/java/io/searchbox/client/ESJestClientFactory.java
@@ -1,3 +1,27 @@
+/*
+ * Copyright (c) 2016 Balabit
+ * Copyright (c) 2016 Laszlo Budai <laszlo.budai@balabit.com>
+ *
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
 package io.searchbox.client;
 
 import com.google.gson.Gson;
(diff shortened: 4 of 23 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

lbudai:
@soubhik-c: thank you for your pull request and also for the detailed description.
We are going to check your work soon.
Just some thoughts: 
 * what if we would move to the official elastic REST lib? What do you think, would it solve your issue? Or we would need to port your solution? (currently we are not planning to replace the JEST lib)
* did you try to use the native, http-driver based elasticsearch destination? If no, could you take a look at that?(after we understand your problem and your solution, we will also check, if it could be solved in the native http based dst)

soubhik-c:
@lbudai  thanks for your attention.

* I haven't tried official elastic REST lib off late but it uses httpasyncclient underneath. In my experiments, I tried to use httpasynclient directly & found under heavy load it behaved erratically and left lots of open connection to ES. Essentially traffic flow halted after sometime. Having said that, I would expect ES folks must have done a better job than me while using httpasynclient. Synch api won't solve the problem for sure. GC pressure is another concern and I am not sure how ES driver high level or low level api caters to multiple message copy. In summary, if ES async api is gc friendly as well as gives a mechanism to parsing the complete response only on Error, my thread pool changes can be avoided (assuming connection timeout of the httpclient pool bug of ES is fixed). For Jest lib, I have overridden some of its classes in this changeset to make it play nicely with plugin level consumer threads minimizing garbage to the maximum extent. This gave me a good boost in sustaining load.

* I have tried native http-driver and mentioned my observations & limitations of it in the description already.

faxm0dem:
FWIW I just updated one of my elasticsearch feeding instances from 3.9.1/elasticsearch-v2-java to 3.21.1/elastic-http, and while it works perfectly, I noticed a high increase in network traffic (almost doubled), and an even higher increase in disk/IO activity on the Elasticsearch nodes (more than doubled).

soubhik-c:
@faxm0dem Thanks for the info. I haven't tried elastic-http but tried native-http in 3.20. I too observed doubled traffic but C impl wasn't able to match performance or stability with these java changes at 500K msg/sec continuous load for 72hrs or more. I am waiting for the experts to get a hang of it and then discuss my changes in greater details. I can spare some cycles to experiment on specific aspect of the C impl from what comes out of our discussion.

pzoleex:

kira-syslogng:
Build triggered

kira-syslogng:
Build FAILURE
 
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
Closing due to inactivity.

Issue #2730, opened 2019-05-15
-d cmd option vs debug --set=on


cmichallek:
```
syslog-ng 3 (3.20.1)
Config version: 3.20
Installer-Version: 3.20.1
Revision: 3.20.1-1
Compile-Date: Feb 26 2019 15:16:58
Module-Directory: /usr/lib/syslog-ng/3.20
Module-Path: /usr/lib/syslog-ng/3.20
Include-Path: /usr/share/syslog-ng/include
Available-Modules: add-contextual-data,afamqp,affile,afmongodb,afprog,afsmtp,afsocket,afsql,afstomp,afuser,appmodel,basicfuncs,cef,confgen,cryptofuncs,csvparser,date,dbparser,disk-buffer,examples,geoip-plugin,geoip2-plugin,graphite,hook-commands,http,json-plugin,kvformat,linux-kmsg-format,map-value-pairs,mod-java,mod-python,pacctformat,pseudofile,redis,riemann,sdjournal,snmptrapd-parser,stardate,syslogformat,system-source,tags-parser,tfgetent,xml
Enable-Debug: off
```

im using syslog-ng in a Docker container and starting it with `syslog-ng -Fd`, as i need the debug output to detect problems occurring in my elasticsearch destinations.
Now i wanted to send syslog-ng's own logs to ES themself.
I noticed nothing goes there, so i did a simple test setup:

```
source s_internal {
  internal();
};
destination d_messages { file("/var/log/messages"); };

log {
  source(s_internal);
  destination(d_messages);
};

```
to my surprise nothing gets written to internal() it seems.
this seems to be related to the -d flag. When i start `syslog-ng -F` and then enable debugging manually with `syslog-ng-ctl debug --set=on`, debug logs are going to internal().

This is a bit awkward, as executing commands after the Docker Entrypoint is kinda hacky.
Is there a way to enable debugging via cmd and still get it send to internal()?

greetings

Christian

bazsi:
-d implies -e today (which stands for --stderr) that instructs syslog-ng to
log everything to stderr and to not log stuff through internal().

At the code level, we have the capability of doing both (e.g. to log to
internal() as well as stderr), but today it's not possible to configure it
this way. Patches are appreciated :)

Kokan:
@bazsi I tried to dig up why the `-e` is implied by `-d`, and found this commit: https://github.com/balabit/syslog-ng/commit/5f82516db500209df040215c39da1b5adfc4d143

[...]

PR #2731, opened 2019-05-15
Debug mode does not imply stderr

Kokan:
Now the `syslog-ng -d` does not mean `syslog-ng -de` (same goes with `-r` option).

This solves #2730


kira-syslogng:
Build SUCCESS

Issue #2488, opened 2019-01-11
python destination: optional return values (for lazies :)) - feature request


furiel:
While writing python code, I often forget about the return values...

In simple cases there is no much error handling needed. This is especially true for init call, that I typically use to prepare some local objects or store options. There is no much things that could go wrong there.

My request is, if a user omits the return value for the api methods, that should be considered as implicit success by the language binding.

Tecnically, when there is no explicit return statement, python still returns the None (singleton) object, that could be handled as a special case.
https://docs.python.org/2/c-api/none.html

That would involve the `init`, `send` and `flush` methods.

For example the snippet below should work:
```
class MyDestination(object):
    def init(self, options):
        print("init called")

    def send(self, message):
        print("send called")
```

I am marking this as good first issue.

tsunderino:
Hi, I would like to work on this issue in the next few days.

furiel:
Awesome thanks!

furiel:
Can you please work on the top of my branch? https://github.com/balabit/syslog-ng/pull/2487
I was a bit early with the feature request and the flush method is introduced in that pr, and the send method is modified too in terms of return values. That would cause significant conflicts for either of us. Sorry for the inconvenience.

tsunderino:
No problem, it got merged before I've started. 
I've found the following python methods:
- python destination: is_opened, open, flush, send, init
- python parser: init, process
- python source: init
- python fetcher: init, open

Should this work for all of these or just the ones mentioned in the request?

furiel:
`is_opened` is not a problem. As it is used as a predicate, it is hard to forget about the return value. Also if `is_opened` could be made to mean implicit success, that means there is no use case for the `open/close/is_opened` trio in that problem context, so users should omit those methods. So `is_opened` can remain as it is.

`open` method in destination would have been useful, but is `is_opened` is used as a predicate for `open` by the python binding, python destination does not check its return value. There is not much which can be done there. 

However, the `open` method in the fetcher is relevant, it is implemented correctly.

So other than `is_opened` and `open` in python destination, the others would worth to change.

MrAnno:
@tsunderino Are you working on this issue?

furiel:
We discussed this topic in a team meeting, and concluded that we would prefer the strict signature in every methods. Closing this issue.

PR #2738, opened 2019-05-17
python binding: optional return values (for lazies :))

tsunderino:
Closes: #2488

### DIFF
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -153,10 +153,19 @@ _dd_py_invoke_void_method_by_name(PythonDestDriver *self, const gchar *method_na
 }
 
 static gboolean
-_dd_py_invoke_bool_method_by_name_with_args(PythonDestDriver *self, const gchar *method_name)
+_dd_py_invoke_bool_method_or_concede(PythonDestDriver *self, const gchar *method_name)
 {
-  return _py_invoke_bool_method_by_name_with_args(self->py.instance, method_name, self->options, self->class,
-                                                  self->super.super.super.id);
+  gboolean result = FALSE;
+  PyObject *args_obj = self->options ? _py_create_arg_dict(self->options) : NULL;
+  PyObject *ret = _py_invoke_method_by_name(self->py.instance, method_name, args_obj, self->class,
+                                            self->super.super.super.id);
+  if(ret)
+    result = ret == Py_None ? TRUE : PyObject_IsTrue(ret);
+
+  Py_XDECREF(ret);
+  Py_XDECREF(args_obj);
+
+  return result;
 }
 
 static gboolean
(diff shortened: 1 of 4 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kjira-syslog-ng: ok to est

MrAnno:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
@tsunderino: thanks for the pull request.

What if someone just forget to return the result of an insert? 
What the original issue suggests (at least to me) is that we should let Python modules loose messages without any control. I know that this is true in many cases (when a Python library not propagates back a result), but after this PR our message is that we don’t care return values... and we encourage Python module developers to be ‘lazies’.

pzoleex:
I agree with lbudai, if no return value it's a way to allow developers to lose messages even accidentally

tsunderino:
@lbudai 
@pzoleex 
I agree with you in that message loss is possible if send() defaults to True, but the requested functionality might be useful for open() and init(). I think send() is the only call site where it is harmful.
If I understand the code correctly, syslog-ng drops the message if a parser returns False, so a default True for parse() can avoid message loss.

What do you think? Can I keep those changes or should I just close the full PR?

lbudai:
@tsunderino: I think only the `send` method where this approach is problematic (so don't close the PR).

furiel:
@tsunderino we discussed this topic in a team meeting, and concluded that we would prefer the strict signature in every methods. My original proposal is invalidated, that's why I am closing this PR. Despite the outcome, thank you for your contribution, we value the effort you made into the topic. We will contact you with a small swag as compensation.

tsunderino:
@furiel Thanks for the gifts, I especially love the tshirt!
I understand your point, but the parse() method is still bothering me a bit. Wouldn't it be useful to return default True in case something bad happens during parsing the message?

furiel:
Me feels returning false is better. My thought process is something like this: a filter parses a message, then based on business logic, permits or denies the message. In case the parsing part fails, that means the message does not fit even into the domain of the filter, hence the right thing is to drop the message. Letting through a message that cannot be controlled by a filter can still break things. Let's say, someone wants to write messages to a file, where the filename comes from a ${.patterndb.class}, and there is a filter that makes sure the template will be not empty. Letting through the message due to error and empty template would mean the message is still lost, because the syslog-ng cannot write to an empty filename.

tsunderino:
I see, thanks for the thorough explanation.

PR #2739, opened 2019-05-17
fix infinite loop

Kokan:
A `reopen` after a `reload` could cause infinite loop, as all of the registered `reopen` hook was linked into the list after the current; creating an infinite loop.

### DIFF
--- a/lib/apphook.c
+++ b/lib/apphook.c
@@ -82,7 +82,7 @@ register_application_hook(gint type, ApplicationHookFunc func, gpointer user_dat
       entry->func = func;
       entry->user_data = user_data;
 
-      application_hooks = g_list_append(application_hooks, entry);
+      application_hooks = g_list_prepend(application_hooks, entry);
     }
   else
     {
--- a/lib/tests/test_apphook.c
+++ b/lib/tests/test_apphook.c
@@ -102,6 +102,35 @@ Test(test_apphook, reopen_hook)
   cr_assert_eq(triggered_count, 1);
 }
 
+static void
+_re_registering_hook(gint type, gpointer user_data)
+{
+  _hook_counter(type, user_data);
+  register_application_hook(type, _re_registering_hook, user_data);
+}
+
+Test(test_apphook, hook_register_from_hook)
+{
+  gint triggered_count = 0;
+
+  register_application_hook(AH_REOPEN_FILES, _re_registering_hook, (gpointer)&triggered_count);
+
+  app_reopen_files();
+  cr_assert_eq(triggered_count, 1);
+}
+
+Test(test_apphook, hook_register_from_hook_and_other_not_triggered_hooks)
+{
+  gint triggered_count = 0;
+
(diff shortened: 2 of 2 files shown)


kira-syslogng:
Build SUCCESS

lib/apphook.c
```diff
@@ -82,7 +82,7 @@ register_application_hook(gint type, ApplicationHookFunc func, gpointer user_dat
       entry->func = func;
       entry->user_data = user_data;
 
-      application_hooks = g_list_append(application_hooks, entry);
+      application_hooks = g_list_prepend(application_hooks, entry);
```
lbudai: hmmm... I need to check code (on my computer, currently I have only my phone), but are you sure that this kind of reordering (or reverse ordering) is the best solution?

Kokan: I was thinking about:
1. "permanent" hooks, but the current hook is general enough if there is a correct option to re-register (did not want to extend interface if possible)
2. move hooks that is going to be triggered into some tmp list, and trigger them one-by-one (so two distinct list)
3. insert hooks before the current (this is a kinda variation of the `2.`)

Mainly I was thinking about `2.` and `3.`, possible the `3.` would be more readable and memory allocation/copy should not be a concern here. 

I do not mind the change of ordering, as there was no guarantee of order before; and imho should not be.

What do you think ?

lbudai: okay, thanks for detailed answer
(it would be nice to share these kind of thoughts next time :) )

furiel: > I do not mind the change of ordering, as there was no guarantee of order before; and imho should not be.

Can you explain what do you mean by "no guarantee of order"? Weren't the hooks executed in the order they were registered?

furiel:
Sorry for being slow here, but how does the error condition depend on reload? For me this is not clear from the code.

Technically, AH_REOPEN_FILES always registers itself, so I thought it should trigger the infinite loop even without reload. But apparently that works fine. However, after a reload, I can always see the infinite loop, so there is definitely something with the reload that causes trouble for reopen.

Kokan:
@furiel it is triggered if there is already a hook registered after `reopen`, because if you check the next item in the list is cached into `l_next`, and the hook function is called after. So if there is no hook after `reopen`, it saves `NULL` and terminates the loop even if a new hook is registered.

The `reload` is just special in a way that some hook got registered after the `reopen` hook.

lib/apphook.c
Kokan: Before this patch they were, but not after. As the hooks for now are inserted into the list head.

furiel:
Thanks for explanation, I understand now. In summary: there is no infinite loop if AH_REOPEN_FILES is the last element of the hooks. 
During startup, it gets to the end of the list by affile_dd_global_init. During reload, however, AH_CONFIG_CHANGED is appended to the list, so the original AH_REOPEN_FILES shifts towards the head. Being not the last element of the list, it causes the infinite loop.

kira-syslogng:
Build SUCCESS

PR #2742, opened 2019-05-21
Add support for floating point operations

### LINKED ISSUE #2734: Support for operation on floats
> We started to talk about the float issue, how to address it.
_Originally posted by @Kokan in https://github.com/balabit/syslog-ng/pull/1710#issuecomment-341640884_

Some logs I am parsing have a duration as `duration=1m1.802729771s`, I though I could capture the minutes (1) and seconds (1.802729771) in two variables, and then `$(+ $(* $(or ${minutes} 0) 60) ${seconds})`, to discover that the operators can only operate on integers.  So I did some dirty workaround, but operators that act on floats may be better (e.g. for adding 2 such duration).

I may be able to allocate some time to add support for this, but before starting anything, I would like to have some feedback because I see multiple ways to reach this goal, the most straightforward being:

* Change the implementation to always manipulate `gdouble` instead of `gint64`;
* Add a new set of operators dedicated to floating point operations (e.g. `$(+. ${a} ${b})`)

Your feedback is appreciated :wink:

smortex:
This PR is about supporting math expression using floating-point values.  The code is published for review for now, because some decisions might need to be discussed.  I plan to amend the commits to address the issues and comments raised by the syslog-ng community, and remove the `[Do not merge]` tag when I consider the changes to be ready.

Fixes #2734

### DIFF
--- a/modules/basicfuncs/numeric-funcs.c
+++ b/modules/basicfuncs/numeric-funcs.c
@@ -21,11 +21,84 @@
  *
  */
 
+#include <math.h>
+
 typedef gboolean (*AggregateFunc)(gpointer, gint64);
+
+typedef struct _Number
+{
+  enum
+  {
+    Integer,
+    Float
+  } value_type;
+  union
+  {
+    gint64 raw_integer;
+    gdouble raw_float;
+  } value_data;
+} Number;
+
+gdouble
+number_as_double(Number number)
+{
+  if (number.value_type == Float)
+    return number.value_data.raw_float;
+
+  return number.value_data.raw_integer;
+}
+
+gint64
+number_as_int(Number number)
+{
+  if (number.value_type == Integer)
+    return number.value_data.raw_integer;
+
+  return number.value_data.raw_float;
+}
+
+gboolean
+number_is_zero(Number number)
+{
+  if (number.value_type == Integer)
+    return number.value_data.raw_integer == 0;
+
+  return fabs(number.value_data.raw_float) < DBL_EPSILON;
+}
+
+gboolean
+parse_integer_or_float(const char *str, Number *number)
+{
+  if (parse_dec_number(str, &(number->value_data.raw_integer)))
+    {
+      number->value_type = Integer;
+      return TRUE;
+    }
+
+  if (parse_float(str, &(number->value_data.raw_float)))
+    {
+      number->value_type = Float;
+      return TRUE;
+    }
+
+  return FALSE;
+}
+
+void
+format_number(GString *result, Number n)
+{
+  if (n.value_type == Integer)
+    {
+      format_int64_padded(result, 0, ' ', 10, number_as_int(n));
+      return;
+    }
+
+  g_string_append_printf(result, "%.20f", number_as_double(n));
+}
 
 static gboolean
 tf_num_parse(gint argc, GString *argv[],
-             const gchar *func_name, gint64 *n, gint64 *m)
+             const gchar *func_name, Number *n, Number *m)
 {
   if (argc != 2)
     {
@@ -34,7 +107,7 @@ tf_num_parse(gint argc, GString *argv[],
       return FALSE;
     }
 
-  if (!parse_dec_number(argv[0]->str, n))
+  if (!parse_integer_or_float(argv[0]->str, n))
     {
       msg_debug("Parsing failed, template function's first argument is not a number",
                 evt_tag_str("function", func_name),
@@ -42,7 +115,7 @@ tf_num_parse(gint argc, GString *argv[],
       return FALSE;
     }
 
-  if (!parse_dec_number(argv[1]->str, m))
+  if (!parse_integer_or_float(argv[1]->str, m))
     {
       msg_debug("Parsing failed, template function's second argument is not a number",
                 evt_tag_str("function", func_name),
@@ -56,7 +129,7 @@ tf_num_parse(gint argc, GString *argv[],
 static void
 tf_num_plus(LogMessage *msg, gint argc, GString *argv[], GString *result)
 {
-  gint64 n, m;
+  Number n, m, res;
 
   if (!tf_num_parse(argc, argv, "+", &n, &m))
     {
@@ -64,7 +137,18 @@ tf_num_plus(LogMessage *msg, gint argc, GString *argv[], GString *result)
       return;
     }
 
-  format_int64_padded(result, 0, ' ', 10, n + m);
+  if (n.value_type == Integer && m.value_type == Integer)
+    {
+      res.value_type = Integer;
+      res.value_data.raw_integer = number_as_int(n) + number_as_int(m);
+    }
+  else
+    {
+      res.value_type = Float;
+      res.value_data.raw_float = number_as_double(n) + number_as_double(m);
+    }
+
+  format_number(result, res);
 }
 
 TEMPLATE_FUNCTION_SIMPLE(tf_num_plus);
@@ -72,7 +156,7 @@ TEMPLATE_FUNCTION_SIMPLE(tf_num_plus);
 static void
 tf_num_minus(LogMessage *msg, gint argc, GString *argv[], GString *result)
 {
-  gint64 n, m;
+  Number n, m, res;
 
   if (!tf_num_parse(argc, argv, "-", &n, &m))
     {
@@ -80,7 +164,18 @@ tf_num_minus(LogMessage *msg, gint argc, GString *argv[], GString *result)
       return;
     }
 
-  format_int64_padded(result, 0, ' ', 10, n - m);
+  if (n.value_type == Integer && m.value_type == Integer)
+    {
+      res.value_type = Integer;
+      res.value_data.raw_integer = number_as_int(n) - number_as_int(m);
+    }
+  else
+    {
+      res.value_type = Float;
+      res.value_data.raw_float = number_as_double(n) - number_as_double(m);
+    }
+
+  format_number(result, res);
 }
 
 TEMPLATE_FUNCTION_SIMPLE(tf_num_minus);
@@ -88,7 +183,7 @@ TEMPLATE_FUNCTION_SIMPLE(tf_num_minus);
 static void
 tf_num_multi(LogMessage *msg, gint argc, GString *argv[], GString *result)
 {
-  gint64 n, m;
+  Number n, m, res;
 
   if (!tf_num_parse(argc, argv, "*", &n, &m))
     {
@@ -96,7 +191,18 @@ tf_num_multi(LogMessage *msg, gint argc, GString *argv[], GString *result)
       return;
     }
 
-  format_int64_padded(result, 0, ' ', 10, n * m);
+  if (n.value_type == Integer && m.value_type == Integer)
+    {
+      res.value_type = Integer;
+      res.value_data.raw_integer = number_as_int(n) * number_as_int(m);
+    }
+  else
+    {
+      res.value_type = Float;
+      res.value_data.raw_float = number_as_double(n) * number_as_double(m);
+    }
+
+  format_number(result, res);
 }
 
 TEMPLATE_FUNCTION_SIMPLE(tf_num_multi);
@@ -104,15 +210,26 @@ TEMPLATE_FUNCTION_SIMPLE(tf_num_multi);
 static void
 tf_num_div(LogMessage *msg, gint argc, GString *argv[], GString *result)
 {
-  gint64 n, m;
+  Number n, m, res;
 
-  if (!tf_num_parse(argc, argv, "/", &n, &m) || !m)
+  if (!tf_num_parse(argc, argv, "/", &n, &m) || number_is_zero(m))
     {
       g_string_append_len(result, "NaN", 3);
       return;
     }
 
-  format_int64_padded(result, 0, ' ', 10, n / m);
+  if (n.value_type == Integer && m.value_type == Integer)
+    {
+      res.value_type = Integer;
+      res.value_data.raw_integer = number_as_int(n) / number_as_int(m);
+    }
+  else
+    {
+      res.value_type = Float;
+      res.value_data.raw_float = number_as_double(n) / number_as_double(m);
+    }
+
+  format_number(result, res);
 }
 
 TEMPLATE_FUNCTION_SIMPLE(tf_num_div);
@@ -120,15 +237,26 @@ TEMPLATE_FUNCTION_SIMPLE(tf_num_div);
 static void
 tf_num_mod(LogMessage *msg, gint argc, GString *argv[], GString *result)
 {
-  gint64 n, m;
+  Number n, m, res;
 
-  if (!tf_num_parse(argc, argv, "%", &n, &m) || !m)
+  if (!tf_num_parse(argc, argv, "%", &n, &m) || number_is_zero(m))
     {
       g_string_append_len(result, "NaN", 3);
       return;
     }
 
-  format_uint64_padded(result, 0, ' ', 10, n % m);
+  if (n.value_type == Integer && m.value_type == Integer)
+    {
+      res.value_type = Integer;
+      res.value_data.raw_integer = number_as_int(n) % number_as_int(m);
+    }
+  else
+    {
+      res.value_type = Float;
+      res.value_data.raw_float = fmod(number_as_double(n), number_as_double(m));
+    }
+
+  format_number(result, res);
 }
 
 TEMPLATE_FUNCTION_SIMPLE(tf_num_mod);
--- a/modules/basicfuncs/numeric-funcs.c
+++ b/modules/basicfuncs/numeric-funcs.c
@@ -261,6 +261,87 @@ tf_num_mod(LogMessage *msg, gint argc, GString *argv[], GString *result)
 
 TEMPLATE_FUNCTION_SIMPLE(tf_num_mod);
 
+static void
+tf_num_round(LogMessage *msg, gint argc, GString *argv[], GString *result)
+{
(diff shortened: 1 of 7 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

modules/basicfuncs/numeric-funcs.c
```diff
@@ -21,11 +21,66 @@
...
+
+struct value
+{
+  int value_type;
+  union
+  {
+    gint64 gint64;
+    gdouble gdouble;
+  } value_data;
+};
```
smortex: `value` is an awful name.  Any suggestion for a better name is welcome :wink:, the idea here is to be able to distinguish if we parsed and integer or a float, and store the value in an easy to manipulate format.

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
@smortex I brough up this feature to the team, and we want to discuss it. It may take some time, but I hope not later than next Wednesday.

smortex:
Sure, thanks you for taking care of it!  I started "the easy way" and I guess a lot of topics will be raised, such as:

* should `sum`, `minimum`, `maximum` & co be updated to support floats;
* should the system be reworked to minimize the impact of roundings when the result of an operation is used in another operation;
* etc.

Maintainers of the project can push commits to my branch, feel free to do so, but please send me a heads-up in this case so that I do not `git push -f` and loose your changes unexpectedly :wink:

modules/basicfuncs/numeric-funcs.c
```diff
@@ -21,11 +21,66 @@
...
+{
+  if (parse_dec_number(str, &(value->value_data.gint64)))
+    value->value_type = VALUE_TYPE_GINT64;
+  else if (parse_float(str, &(value->value_data.gdouble)))
+    value->value_type = VALUE_TYPE_GDOUBLE;
+  else
+    return FALSE;
+
+  return TRUE;
+}
```
Kokan: [optional]
```c
gboolean
parse_value(const char *str, struct value *value)
{
  if (parse_dec_number(str, &(value->value_data.gint64))) 
   {
       value->value_type = VALUE_TYPE_GINT64;
       return TRUE;
    }

 if (parse_float(str, &(value->value_data.gdouble)))
   {
       value->value_type = VALUE_TYPE_GDOUBLE;
       return TRUE;
    }

  return FALSE;
}
```

modules/basicfuncs/numeric-funcs.c
```diff
@@ -56,83 +111,204 @@ tf_num_parse(gint argc, GString *argv[],
...
-  format_int64_padded(result, 0, ' ', 10, n + m);
+  if (n.value_type == VALUE_TYPE_GINT64 && m.value_type == VALUE_TYPE_GINT64)
+    {
+      format_int64_padded(result, 0, ' ', 10, value_as_int(n) + value_as_int(m));
+    }
+  else
+    {
+      gdouble res = value_as_double(n) + value_as_double(m);
+      g_string_append_printf(result, "%f", res);
+    }
```
Kokan: Please extract this part into a function (as used in many places) like `format_number`.

modules/basicfuncs/numeric-funcs.c
```diff
@@ -56,83 +111,204 @@ tf_num_parse(gint argc, GString *argv[],
...
+  if (!parse_value(argv[0]->str, &n))
+    {
+      msg_debug("Parsing failed, template function's first argument is not a number",
+                evt_tag_str("function", "round"),
+                evt_tag_str("arg1", argv[0]->str));
+      g_string_append_len(result, "NaN", 3);
+      return;
+    }
+
+  g_string_append_printf(result, "%ld", (gint64) round(value_as_double(n)));
```
Kokan: Please use `format_int64_padded` instead of printf.

modules/basicfuncs/numeric-funcs.c
```diff
@@ -56,83 +111,204 @@ tf_num_parse(gint argc, GString *argv[],
...
+  if (!parse_value(argv[0]->str, &n))
+    {
+      msg_debug("Parsing failed, template function's first argument is not a number",
+                evt_tag_str("function", "ceil"),
+                evt_tag_str("arg1", argv[0]->str));
+      g_string_append_len(result, "NaN", 3);
+      return;
+    }
+
+  g_string_append_printf(result, "%ld", (gint64) ceil(value_as_double(n)));
```
Kokan: Please use `format_int64_padded` instead of printf.

modules/basicfuncs/numeric-funcs.c
```diff
@@ -56,83 +111,204 @@ tf_num_parse(gint argc, GString *argv[],
...
+  if (!parse_value(argv[0]->str, &n))
+    {
+      msg_debug("Parsing failed, template function's first argument is not a number",
+                evt_tag_str("function", "floor"),
+                evt_tag_str("arg1", argv[0]->str));
+      g_string_append_len(result, "NaN", 3);
+      return;
+    }
+
+  g_string_append_printf(result, "%ld", (gint64) floor(value_as_double(n)));
```
Kokan: Please use `format_int64_padded` instead of printf.

modules/basicfuncs/tests/test_basicfuncs.c
```diff
@@ -219,6 +219,19 @@ Test(basicfuncs, test_numeric_funcs)
...
+  assert_template_format("$(/ 3 2)", "1");
+  assert_template_format("$(/ 3.0 2)", "1.500000");
+  assert_template_format("$(/ 3 2.0)", "1.500000");
+  assert_template_format("$(* 1.5 2.0)", "3.000000");
+  assert_template_format("$(% 3.14 0.7)", "0.340000");
+
+  assert_template_format("$(round 2.0)", "2");
+  assert_template_format("$(round 0.5)", "1");
+  assert_template_format("$(floor 0.7)", "0");
+  assert_template_format("$(ceil 0.2)", "1");
```
Kokan: [optional]
Could you please extend the UT test set with scientific notation number parsing cases. Like `$(+ 1e-1 0)` (it already supports those).

Kokan:
I tried to summarise the result of that conversation:

A short summary of the discussion about this topic:
===

## Re/defining the arithmetic operations:

The way how operation should be defined: like the one you proposed in your PR, the one in the issue( new operations for float).
The conclusion on that was that your proposal is the solution we want to have. (Mainly because of the next point, and because it is natural on some level.)

## Impact on current configurations

With your version it is not an issue, as float was not supported before and operations with float resulted invalid/unwanted result (like *NaN*)

## Storage method/1

There is a recurring idea that nv-pairs should have more sophisticated way to store different types (other then the string representation).
This discussion was halted as it is kinda complex, and that would explode the work needed to be done now (also could be splitted as well).

## Storage method/2

Your proposal stores the float in a string representation. (this is somewhat connected to the next topic)
That introduces a precision syslog-ng could handle, which must be defined and compatible what *float*/*double* precision could do.


## Operations error propagation

A brief discussion about the errors each operation could introduce, or just the converting into string.
The conclusion on this was that syslog-ng is not a scientific tool, and as such the error propagation is a best effort basis.
Meaning if some issue raises on the matter, we can open a discussion to solve or state that won't solve such an issue;
as likely such discussion will never come up.

# What it means for this PR:

* The way you defined the operation extension is OK
* The precision/storage need some improvement

modules/basicfuncs/numeric-funcs.c
```diff
@@ -56,83 +111,204 @@ tf_num_parse(gint argc, GString *argv[],
...
 
-  format_int64_padded(result, 0, ' ', 10, n - m);
+  if (n.value_type == VALUE_TYPE_GINT64 && m.value_type == VALUE_TYPE_GINT64)
+    {
+      format_int64_padded(result, 0, ' ', 10, value_as_int(n) - value_as_int(m));
+    }
+  else
+    {
+      gdouble res = value_as_double(n) - value_as_double(m);
+      g_string_append_printf(result, "%f", res);
```
Kokan: Regarding the precision, while the `double` could hold a nice range. Saving that into that as string has impact on memory usage. It would be really nice to know the usecases it could come up.

I would suggest an arbitrary number to be 20 fractal digit.
This is based on assumption that *float* is used for time (like fraction of seconds), and the current limitation of such measurement according to wikipedia is somewhere near the `10e-17` range. [Atomic clock](https://en.wikipedia.org/wiki/Atomic_clock#Secondary_representations_of_the_second)
Note: this is a really big reduction specially for `double`, but keep in mind that each new digit is stored as string
I am happy with other values, this is more like a conversation starter.

What precision would you suggest ?

smortex: My use case is related to milliseconds.  Because of the roundings, I assumed a 6 digits precision was okayish (as you say, syslog-ng is not a scientific tool), but for sure, more precision is probably better and 20 digits seems a reasonable choice.

modules/basicfuncs/tests/test_basicfuncs.c
smortex: Sure, will use `5e-1` because, you know IEEE-754 cannot represent a lot of numbers with 20 digits precision: actual=0.10000000000000000555, expected=0.10000000000000000000 :smile:

smortex:
I added a bunch of commits to fix the reported issues (many thanks, the level of details is just perfect!).

In order to make reviewing easier, I have not touched previous commits, and will only squash them when you are okay with it.

kira-syslogng:
Build SUCCESS

smortex:
@Kokan, just to be sure, can you please confirm that your approval has value of "ok for squashing the commits and moving on to the next step"?

Kokan:
@smortex yes for the squash; but if by next step you mean merging the PR, then no. as that requires at least two approver.

smortex:
Sure! By "next step" I meant further work on the PR ;-)

smortex:
I reordered the commits, fixed and squashed the commits after rebasing on top of the master branch.  I pushed the two modified commits one by one in order to trigger two builds.  Let's see how is goes :wink:

kira-syslogng:
Build SUCCESS

smortex:
CI on Travis-CI fails on OSX because unit tests are disabled.  This seems to be caused by criterion not being installed in a previous stage of the build.  Transient failure?

Kokan:
Yes, after restarting that is usually fixed. Sometimes brew cannot install criterion :/

smortex:
Cool!  Thank you for re-triggering the build.

Do you think that I should add a way to ask `round` to accept an optional argument (defaulting to 0) for setting the precision in decimal digits:

```
(round 2.123456)   #=> "2"
(round 2.123456 3) #=> "2.123"
```

I am thinking that making it possible to have an integer or a float with 20 decimal digits is maybe a bit restrictive for some use cases (in my case, I manipulate data with 3 digits precision, so it does not make sense to have more than 3 digits in my final result)?

Kokan:
I agree on supporting the number of digit precision for `round`, and as well with the default to 0.
I would say it is fine to expand this PR until your usecase is completely covered; unless you want/could split this PR up into smaller chunks that make sense. In this case I would include the change regarding to `round`.

kira-syslogng:
Build FAILURE

smortex:
Thanks!  I added the optional precision argument to round, and reached a complete coverage of my use-case.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/basicfuncs/tests/test_basicfuncs.c
```diff
@@ -219,6 +219,23 @@ Test(basicfuncs, test_numeric_funcs)
...
+  assert_template_format("$(- -1.5 .25)", "-1.75000000000000000000");
+  assert_template_format("$(/ 3 2)", "1");
+  assert_template_format("$(/ 3.0 2)", "1.50000000000000000000");
+  assert_template_format("$(/ 3 2.0)", "1.50000000000000000000");
+  assert_template_format("$(* 1.5 2.0)", "3.00000000000000000000");
+  assert_template_format("$(% 3.14 0.7)", "0.34000000000000030198");
+
+  assert_template_format("$(+ 5e-1 0)", "0.50000000000000000000");
+
+  assert_template_format("$(round 2.0)", "2");
```
Kokan: [optional] I would extend the testcase with
```c
assert_template_format("$(round 2 -1)", "NaN");
assert_template_format("$(round 2 21)", "NaN");
assert_template_format("$(round 2 0)", "2");
assert_template_format("$(round 2 20)", "2.00000000000000000000");
```

modules/basicfuncs/numeric-funcs.c
```diff
@@ -56,83 +137,248 @@ tf_num_parse(gint argc, GString *argv[],
...
+                evt_tag_str("arg1", argv[0]->str));
+      g_string_append_len(result, "NaN", 3);
+      return;
+    }
+
+  if (argc > 1)
+    {
+      if (!parse_dec_number(argv[1]->str, &precision))
+        {
+          msg_error("Parsing failed, template function's second argument is not a number",
```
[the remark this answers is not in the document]
smortex: Oops, this was completely unintentional, the rest of the code use `msg_debug()`, I guess I copy-pasted some code too far away and did not see that the function was not the same :confused:

modules/basicfuncs/numeric-funcs.c
```diff
@@ -21,11 +21,92 @@
...
+
+  g_string_append_printf(result, "%.20f", number_as_double(n));
+}
+
+void
+format_number_with_precision(GString *result, Number n, unsigned int precision)
+{
+  char format[6];
+  snprintf(format, sizeof(format), "%%.%df", precision);
+  g_string_append_printf(result, format, number_as_double(n));
```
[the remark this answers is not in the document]
smortex: I realized this after committing my first batch of changes, and amended this old code.

modules/basicfuncs/numeric-funcs.c
```diff
@@ -87,19 +106,24 @@ parse_integer_or_float(const char *str, Number *number)
...
+      if (n.value_type == Integer)
+        {
+          format_int64_padded(result, 0, ' ', 10, number_as_int(n));
+          return;
+        }
+
+      /*
+       * This special case is required because:
+       * round(0.5)          #=> 1
+       * printf("%.0f", 0.5) #=> 0
```
smortex: I am still investigating the reason of this, and hope to simplify this code when I fully understand what is going on here.

smortex:
I added a bunch of commit that address the problems you spotted.  Thanks!

kira-syslogng:
Build FAILURE

modules/basicfuncs/numeric-funcs.c
smortex: The [POSIX specification of the 'f' format string](https://pubs.opengroup.org/onlinepubs/009695399/functions/fprintf.html) says (emphasis by myself):

> f, F
>     The double argument shall be converted to decimal notation in the style "[-]ddd.ddd", where the number of digits after the radix character is equal to the precision specification. If the precision is missing, it shall be taken as 6; if the precision is explicitly zero and no '#' flag is present, no radix character shall appear. If a radix character appears, at least one digit appears before it. **The low-order digit shall be rounded in an implementation-defined manner.**

So le'ts use `round(3)` when possible, and rely on the implementation-defined rounding otherwise.

smortex: Do you have an opinion about this?

1) Is this okay as it is currently?
2) Should we always round using %f (that is the test suite needs to be updated because 0.5 is rounded to 0)?
3) Should we always use round(3) (mutiply the number by 10^precision, do the rounding, divide by 10^precision and display this result with %f and precision digits)?

More and more I think 3 is the best way to go because of the consistency of the roundings, but 1 is ways more simple and maybe acceptable for a logging software. 2 is hybrid and inconsistent, so maybe it's the worst possibility.

What do you think?

kira-syslogng:
Build FAILURE

modules/basicfuncs/numeric-funcs.c
Kokan: I can agree on the 3rd option, as precision cannot be a conserve while it is limited to 20 digits from the 300 something `gdouble` has, in addition to the reasons you wrote.

Kokan: I left a note for the wip patch you linked.

Kokan:
@smortex please feel free to fixup the changes in the next step (if you agree with my comment on your commit), and from my side it is good to go.

kira-syslogng:
Build SUCCESS

smortex:
Thanks @Kokan I squashed the last commits into a single one.  It looks like the "default" kira build fails, but I cannot access the system.  Is there something I can fix myself?

Kokan:
@smortex please ignore the *default* kira.

bazsi:
That job is work in progress, so ignore it please. I have disabled it for
now.

Kókai Péter <notifications@github.com> ezt írta (időpont: 2019. jún. 6., Cs
23:05):

Kokan:
@kira-syslogng reteset this please
(@smortex I'll do some stuff here to make default green)

Kokan:

kira-syslogng:
Build FAILURE

Kokan:

kira-syslogng:
Build SUCCESS

Issue #2741, opened 2019-05-20
geoip2 error messages


faxm0dem:
since upgrading from syslog-ng 3.9.1 to 3.21.1, we're getting lots of geoip errors:

```
maxminddb_error; error='Unknown error code', where='lookup'
Error from call to getaddrinfo; gai_error='Name or service not known', where='lookup'
```

here's our config:

```
parser p_geoip {
    geoip2(
        "${geoip.ip}",
        prefix(
            "geoip2."
        ),
        database(
            "/opt/geoip/dat/GeoLite2-City.mmdb"
        )
    );
};
```

bazsi:
Hmm, this might be because the reverse lookup of the ip address is not
registered? The new implementation uses maxminddb and maybe that's more
chatty.

Kokan:
The `maxmindb` can only lookup ip addresses, and it does call `getaddrinfo` (which could resolve hostnames), but also specifies that it is going to get an IP so no need to do dns resolution: https://github.com/maxmind/libmaxminddb/blob/85ff16afb2d547e1a74999e50667a4e493a3a13a/src/maxminddb.c#L868

Thus the following error you get. But the error is somewhat misleading :( the `Name or service not known` part is provided by `getaddrinfo` itself.

faxm0dem:
It could be useful to know the address it's failing on ;-)

faxm0dem:
I just tested this patch and here's the output:

```
geoip2(): getaddrinfo failed; gai_error='Name or service not known', ip='[', location='/etc/syslog-ng/syslog-ng.conf:1869:5'
```

much better :-) thanks so much!

MrAnno:
@faxm0dem Have you managed to find out how you end up resolving the `[` character as an IP? Can we help you?

faxm0dem:
Yes thanks to the patch, thank you very much!

PR #2743, opened 2019-05-21
geoip2: include ip into the error message

Kokan:
This was inspired by #2741

### DIFF
--- a/modules/geoip2/geoip-parser.c
+++ b/modules/geoip2/geoip-parser.c
@@ -61,13 +61,13 @@ _mmdb_load_entry_data_list(GeoIPParser *self, const gchar *input, MMDB_entry_dat
 
   if (!result.found_entry)
     {
-      if (0 != _gai_error)
+      if (_gai_error != 0)
         msg_error("geoip2(): getaddrinfo failed",
                   evt_tag_str("gai_error", gai_strerror(_gai_error)),
                   evt_tag_str("ip", input),
                   log_pipe_location_tag(&self->super.super));
 
-      if (MMDB_SUCCESS != mmdb_error)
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

modules/geoip2/geoip-parser.c
```diff
@@ -75,7 +75,16 @@ _mmdb_load_entry_data_list(GeoIPParser *self, const gchar *input, MMDB_entry_dat
 
   if (!result.found_entry)
     {
-      mmdb_problem_to_error(_gai_error, mmdb_error, "lookup");
+      if (0 != _gai_error)
```
lbudai: Are there any reasons to use this pattern in conditional expressions?
In good old days, this kind of pattern was used to save your variable from an unwanted assignment... (eg.: `_gai_error = 0` vs. `_gai_error == 0`) but nowadays compilers detects this situation (-Wparentheses maybe).
And in Java there is a similar pattern when you write this:
```
"mystring".equalsTo(object)
```
instead of checking null pointers, etc...

In syslog-ng we usually use the other way, variable first, then literal/constant... and it's more readable (at least for me...).

lbudai: Sorry, I see that you just keep the original code... then we will discuss this pattern IRL on next handover :)

Kokan: Yeah. I stopped the refactor there, but sure I'll update accordingly.

Note: Personally I like yoda condition, but do not have hard feeling for it.

lbudai: Yoda condition... so they also gave a name for this pattern. Thanks :)

kira-syslogng:
Build SUCCESS

PR #2746, opened 2019-05-23
snmp dest leak template

Kokan:
* Fix unit test issues:
  * `cfg_new_snippet` should be used instead of `cfg_new` in UT
  * `log_pipe_init` should not be used before setter methods (as it is not done in syslog-ng normal execution either)
* Some minor refactors
* Removed tailing `NULL` from `msg_` macros (`scripts/rm-extra-msg-sentinel.sh`)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
@bazsi agree on the msg additional cleanup; but that I think would make more sense when the snmp-dst is reworked on top of threaded destination driver (as might a lot of message may disappear), and that refactor is really needed imho.

PR #2747, opened 2019-05-23
global options: support generator plugin

Kokan:
Fix
* generator plugin was not accepted in options block
* multiple semicolons were not accepted in a few place (for example in options)

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -479,7 +479,7 @@ start
 	;
 
 stmts
-        : stmt ';' stmts
+        : stmt semicolons stmts
 	|
 	;
 
@@ -863,7 +863,7 @@ template_fn
 	;
 
 template_items
-	: template_item ';' template_items
+	: template_item semicolons template_items
 	|
 	;
 
@@ -951,7 +951,7 @@ block_arg
         ;
 
 options_items
-	: options_item ';' options_items
+	: options_item semicolons options_items
 	|
 	;
 
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -814,7 +814,10 @@ log_flags_items
 /* END_RULES */
 
 options_stmt
-        : KW_OPTIONS '{' options_items '}'
+        : KW_OPTIONS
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
I can't think of a good use case, but we support plugins in context `LL_CONTEXT_OPTIONS`, so this functionality may turn out to be useful.

I validated the PR with this not-so-useful example block, but the last semicolon causes some problems:
```
block options setup-dns(use-dns()) {
	keep-hostname(no);
	use-dns(`use-dns`);
	use-fqdn(`use-dns`);
	dns-cache(`use-dns`);
};

options {
	setup-dns(use-dns(yes));
};
```

```
Error parsing global options, syntax error, unexpected ';'
12---->         setup-dns(use-dns(yes));
12---->                                ^
```

Removing the semicolon either after `dns-cache()` or after `setup-dns()` makes it work.

Kokan:
I do not think either that there is *real* usecase - which cannot be solved better.
This was triggered via the mail list: https://lists.balabit.hu/pipermail/syslog-ng/2019-May/025294.html
(where actually somebody wanted to use it)

But even if it is not useful, it is an implementation bug; as generator plugin should act the same way as any other plugin.

The issue you found I consider as a general generator plugin bug, it does only replaces the `generator-name()` part and not the whole `generator-name();`, therefore it creates two `;;`, which is not accepted in every place.

MrAnno:
I agree that the lack of the `cfg_lexer_push_context` call is a bug-like thing.
The underscore issue I mentioned is another bug, but it makes the whole functionality (that you've enabled) hard to use.

Do we want to fix the underscore issue in this PR or you prefer postponing it?

Kokan:
well I planned to solve that issue when there is proper pre-processing;I'll give it a try, though if I do not like the patches I'll postpone the fix. (will report back on conclusion)

Kokan:
I've found a way (not how initially wanted) to handle the multiple semicolon issue.

kira-syslogng:
Build SUCCESS

MrAnno:
Oh, that is tricky :)

PR #2752, opened 2019-05-26
debun: add acquire_running_syslog_config function

mehul-m-prajapati:
This function will save ```syslog-ng-ctl config --preprocessed``` output in syslog-ng.conf if syslong-ng is live,
otherwise it will store syslog-ng.conf from disk.

resolves https://github.com/mehul-m-prajapati/gsoc-debun-tool/issues/1

### DIFF
--- a/contrib/syslog-ng-debun
+++ b/contrib/syslog-ng-debun
@@ -839,6 +839,7 @@ acquire_syslog_startup_smf () {
 acquire_syslog_all () {
 	printf "\nStart Syslog-specific info collection\n"
 	acquire_syslog_config
+	acquire_running_syslog_config
 	acquire_syslog_pki_info
 	acquire_syslog_pids
 	acquire_syslog_info
@@ -868,6 +869,14 @@ fhs_set_linux () {
 
 fhs_set_unix () {
 	:
+}
+
+acquire_running_syslog_config() {
+
+	if ${syslogngctlbin} config ; then	
+		${syslogngctlbin} config -p > ${tmpdir}/syslog-ng.ctl.running.conf 2>&1
+		remove_passwords_from_file "${tmpdir}/syslog-ng.ctl.running.conf"
+	fi
 }
 
 rpm_verify () {


kira-syslogng:
Build SUCCESS

contrib/syslog-ng-debun
```diff
@@ -1664,6 +1674,7 @@ setup_env
 debun_run
 [ "$syslogfhs" = "linux" ] && fhs_set_linux
 [ "$syslogfhs" = "unix" ] && fhs_set_unix
+save_config_file
```
Kokan: Does it align with the `$privacy_mode`  mode ?

contrib/syslog-ng-debun
```diff
@@ -870,6 +870,16 @@ fhs_set_unix () {
 	:
 }
 
+save_config_file() {
```
Kokan: The `acquire_syslog_config` already saves the configuration (from disk to debun).

contrib/syslog-ng-debun
```diff
@@ -870,6 +870,16 @@ fhs_set_unix () {
...
+
+	if ${syslogngctlbin} config; then	
+		${syslogngctlbin} config -p > ${tmpdir}/syslog-ng.conf 2>&1
```
Kokan: 1. I would add some different name, so people could identify the way it was created. (for example `syslog-ng.ctl.running.conf`)
2. There is a `remove_passwords_from_file` function that would be advisable to use on any config

contrib/syslog-ng-debun
mehul-m-prajapati: ok.

contrib/syslog-ng-debun
mehul-m-prajapati: ok. thanks for a good suggestion.

contrib/syslog-ng-debun
mehul-m-prajapati: I think it will save config file in privacy mode, too.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

contrib/syslog-ng-debun
```diff
@@ -870,6 +871,14 @@ fhs_set_unix () {
 	:
 }
 
+acquire_running_syslog_config() {
+
+	if ${syslogngctlbin} config ; then
```
furiel: What is the purpose of this line? Is this how we check if syslog-ng is running?

furiel:
I am fine with this changeset as an initial (quick) implementation. I just have a little concern with it. I do not want you to address these now, I am just interested in your opinion.

Is it fine to depend on a running syslog-ng instance? Do we want to silently do less when there is no syslog-ng running?
I am thinking about that debun is usually executed after a problem occurs with syslog-ng, including crash. Service management may or may not restarts syslog-ng, but there is a chance that syslog-ng is not running.

It would be useful if you could elaborate on the use case, for now I am guessing: the goal here is not a lie in collectiong the **running** config, but to collect every files, every includes that the config is using. And preprocessed is a mean (implementation detail) to achieve this goal.
The running part is not too useful on its own, even on a single configuration file case: if you see the running config is the same as the config on disk, then you do not acquire any extra information from the running config. If those differs, then you only know that reporter changes configuration on disk (maybe experimenting with the bug), but you would not know if the other parts of the debun (core file) relates to that config running config.

That's why I think the goal here is to collect syslog-ng configuration in general. In that case, maybe it would worth rename the function accordingly, or better: move under `acquire_syslog_config`.

I would prefer a solution however that would work even when syslog-ng is stopped. Also, and this may be a personal preference, but I like to read the non-preprocessed version of the config. It is more convenient.

Kokan:
The main goal is to collect files referenced in `syslog-ng` configuration (such as `kafka` properties-file), which could of course be collected based on the currently saved config/preprocessed config.

While the preprocessed config is more accurate, it is better to use that as a source for the above task.

They are connected in a way that there is a slight possibility that `syslog-ng` is running, and the local config have been tampered with in a way it is no longer valid. Thus `preprocess-into` cannot work.
Or in minor cases you mentioned.

The `debun` should not start syslog-ng (unless explicitly asked it) as it is possible in production environment.

furiel:
Ok, thanks for explanation. Sound good to me.

contrib/syslog-ng-debun
mehul-m-prajapati: @furiel : This is just to check ```syslog-ng-ctl``` is present on the disk and it could fetch config from currently running syslog-ng process.

PR #2757, opened 2019-05-29
http: Warn if less workers used than urls

gaborznagy:
If less workers are used, then some url targets will not receive messages.
We even warn about this in documentation.


kira-syslogng:
Build FAILURE

gaborznagy:

kira-syslogng:
Build SUCCESS

bazsi:
I think we should give more guidance, what happens in the case that value is lower.

gaborznagy:
> I think we should give more guidance, what happens in the case that value is lower.

AFAIK not used urls will work as a failover-server, but I don't think we should advertise this as a feature.

kira-syslogng:
Build SUCCESS

PR #2761, opened 2019-05-31
stats: dynamic constant registration

furiel:
This patchset implements dynamic type registration in stats. Module specific `SCS_` constants disappear, and the individual modules need to register them. There are a few lib-specific `SCS_` constants. Those are still registered in `stats_init`.

Risks:
- As I saw every module specific constants are registered from the mainloop, so I was brave enough to add a `main_loop_assert_main_thread();` to the registration. The changeset expects that the registrations are happening sequentially. If it's not the case, a little more work is needed. Fingers crossed.
- The way the registered constants are passed to the core did not change, still need to assign an integer to `stats_source`. This means if some module used hard-coded values, those might not be aware of the change (compile time), and there is some possibility that those modules break silently (wrong stats type will be written, or crash).


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng test this please test branch=fix-severity-test;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng test this please test branch=fix-severity-test;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng test this please test branch=fix-severity-test;

kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng do perftest;

kira-syslogng:

mitzkia:
@kira-syslogng test this please test branch=micek_fix_severity; test=functions/other-functions/statistics/003-stats-facility-severity.zts;

lib/stats/stats-cluster.c
```diff
@@ -26,6 +26,25 @@
...
 
+GPtrArray *stats_types;
+static const gchar *module_names[];
+
+void
+stats_cluster_init(void)
+{
+  g_assert(!stats_types);
+  stats_types = g_ptr_array_new_with_free_func(g_free);
+  for (int i = 0; i < SCS_MAX; i++)
```
lbudai: Use `ARRAY_SIZE(module_names)` instead of `SCS_MAX`.

lib/stats/stats-cluster.c
```diff
@@ -28,15 +28,25 @@
...
+  g_assert(stats_register_type("global") == SCS_GLOBAL);
+  g_assert(stats_register_type("center") == SCS_CENTER);
+  g_assert(stats_register_type("host") == SCS_HOST);
+  g_assert(stats_register_type("sender") == SCS_SENDER);
+  g_assert(stats_register_type("program") == SCS_PROGRAM);
+  g_assert(stats_register_type("severity") == SCS_SEVERITY);
+  g_assert(stats_register_type("facility") == SCS_FACILITY);
+  g_assert(stats_register_type("tag") == SCS_TAG);
+  g_assert(stats_register_type("filter") == SCS_FILTER);
+  g_assert(stats_register_type("parser") == SCS_PARSER);
```
lbudai: Why are these asserts? (or, why are these asserts are missing in other cases?).

What if we would keep the array (a "core" array with these names).

kira-syslogng:
Build SUCCESS

furiel:

kira-syslogng:
Build SUCCESS

lib/stats/stats-cluster.c
furiel: When we had a separate array for the names, it had a certain risk: the list of names and the enum could diverge. This happened once with `SCS_STDIN`, and was fixed in 5a071f4eac89aace93046d06b5a0f532b9767a42. That's when we added a static assert to check if the size of the two entities are the same. Checking lengths mitigates some of the risks, but could not catch bugs coming from when the items in the of them are reordered while the other one did not follow this change up. I think this kind of bug has not yet happened, so maybe I am too paranoid here.

Bringing back the original array would result in little more complex code, at least to my taste: we need a static array, we need a dynamic array, we need to test SCS_STATIC_MAX with the static array length (I would not let this check as we had run into problem due to this). Then the static array still need to be registered, because the dynamic registration should not overlap with it. This latter problem could be resolved in other way, for example dynamic allocation would return dynamic_value + SCS_STATIC_MAX, and during counter registration, we could decide if the counter is static or dynamic by comparing the input value with SCS_STATIC_MAX, and thus branch the lookup in the static array or in the dynamic array. Me feels this version simpler.

I can drop the g_asserts though if you think we do not have much benefit from them. Then I would just g_assert the lengths.

furiel:
@lbudai reworded the last commit message to dynamic type registration instead of dynamic constant registration. Thanks for pointing out.

lib/stats/stats-cluster.c
furiel: I would like to come back to this question if we conclude that I need to bring back the static array, as in the current version this code is eliminated. Also I think I can reason why SCS_MAX is better here, but I do not want to open that topic if we do not need this code anyway.

kira-syslogng:
Build SUCCESS

lib/stats/stats-cluster.c
lbudai: okay, and can't we simply set these values during registration (I mean do we need these pre-defined SCS_xxx values)?

furiel: Good idea. I can make these availables via extern. I will do this.

furiel: Instead of extern, I created a stateful function that is responsible for getting and setting these constants. This has two benefits: these "constants" are not mutable through accidental assignment, and also I can assert if caller tries to use them before initialization.

kira-syslogng:
Build SUCCESS

bazsi:
This pull request **introduces 2 alerts** when merging bdc94e322e327a331a8826cafd3b2b7781d70094 into cf10b3c4a50d8feda50826b518428719fd974a1c - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-4ed2b19549eaa7a1334f8ccecbd51d0e26183c03)

**new alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

lib/stats/stats-cluster.c
```diff
@@ -22,10 +22,85 @@
...
+      g_assert(data < SCS_SOURCE_MASK);
+      constant_storage[constant] = data;
+      return 0;
+    default:
+      g_assert_not_reached();
+    }
+
+  return 0;
+}
+
```
bazsi: I don't see why we don't split this function into a separate set and get function? the only reason I can think of is that it contains the constant_storage as a static variable. Pushing that into the global scope (but keeping it static) would get rid off SCS_ACTION_GET/SET and make this pretty complex function two simple ones. I think that's a fair price to pay for the increase of visibility for the constant_storage variable.

modules/afamqp/afamqp.c
```diff
@@ -791,7 +791,7 @@ afamqp_dd_new(GlobalConfig *cfg)
   self->super.worker.insert = afamqp_worker_insert;
 
   self->super.format_stats_instance = afamqp_dd_format_stats_instance;
-  self->super.stats_source = SCS_AMQP;
+  self->super.stats_source = stats_register_type("amqp");
```
bazsi: it might actually make a lot of things simpler if we converted stats_source into a `const char *` and then pass the amqp string there and make the actual allocation in lib.

modules/afsocket/tests/test-transport-mapper-inet.c
```diff
@@ -31,6 +31,13 @@
...
 #include <unistd.h>
 
+guint SCS_TCP;
+guint SCS_TCP6;
+guint SCS_UDP;
+guint SCS_UDP6;
+guint SCS_NETWORK;
+guint SCS_SYSLOG;
+
 static inline void
```
bazsi: these are pretty ugly again. if we followed the same pattern (e.g. pass a const char * to the function that needs the stats identifier), we could get rid of these variables.

modules/afamqp/afamqp.c
bazsi: even better idea, make SCS_AMQP equal ot the "amqp" string. with that the code does not even have to change.

furiel: I was thinking about this version too, I even started to implement it, but there were some technical and theoretical questions that I could not reliably answer. I will give another try now.
The changeset will definitely will complex, because I need to separate the stats name and direction into separate parameters for the registration, and this will have interesting consequences. Marking the PR wip.

furiel: In this test: https://github.com/balabit/syslog-ng/blob/f3298e8d187c5b2ac9a89755d3af79595b5553b2/lib/stats/tests/test_dynamic_ctr_reg.c#L72
we link two different names (SCS_HOST, SCS_SENDER) to the same counter. Am I right that this is invalid? 19 | 27 = 27, so I think during lookup it would print "sender", but I am not sure if this is intended behavior. Do we want to support such thing? I did not see any such application apart from test environment.

If this is a thing, then the const gchar * version cannot work. More, dynamic registration of the current version cannot not work either, because we cannot register constants randomly, we should maintain some relations like SCS_HOST | SCS_SENDER = SCS_SENDER.

lib/stats/stats-cluster.c
furiel: My new version was not very good either. For now I dropped this patch. Lets address this at a later time.

lib/stats/stats-cluster.c
furiel: You are right. For now I reverted that patch.

kira-syslogng:
Build SUCCESS

PR #2768, opened 2019-06-07
Light: fix restart ordering

mitzkia:
This patch set fixes two related issues in SyslogNg() class
- fixes restart() ordering, started with stop() follows with start()
- do not allow to start same syslog process twice

- testcases are not created

PR #2723, opened 2019-05-11
[WIP] light: Refactor SyslogNgConfig API (external/internal)

mitzkia:
This patch set provides a general API for syslog-ng config drivers with the following changes:
- `working_dir` class variable has been converted into a global variable (it was also a main dependency of SyslogNgConfig classes)
- After that driver/parser/filter related wrapper functions (such as create_file_source()) could be removed from SyslogNgConfig() (they were boilerplates).
- Start using own classes (such as FileSource, FileDestination) for creating driver/parser/filter like objects in testcases
- Change functionality of SourceDriver and DestinationDriver to SourceWriter and DestinationReader, 
from now these classes are responsible only for general driver related input/output functionalities.
- Expose file related option handling into a new class: `FileBasedOptionHandler()`
- Follow the changes in functional testcases

Todos:
- find a better place to config.stringify()
- separate logpath related wrappers from SyslogNgConfig
- automatically create logpath (if needed)
- add missing functional testcase: change file-path during testrun
- add missing functional testcase: add new file based option during testrun (an example usage for OptionHandler() class)

PR #2769, opened 2019-06-07
Light: Remove working dir from SyslogNgConfig dependencies

mitzkia:
This PR is based on my #2768 PR.

The patch set of this PR was previously part of my #2723. I have just separated it from there.
I have decided to create smaller separate PRs to move forward Light related PRs faster.

Patch description:
- with initializing working_dir as a global variable we will able to eliminate it from SyslogNgConfig() class.
- with this change SyslogNgConfig() class can be used more easily in Ligth, and opens it for a refactoring

### DIFF
--- a/tests/python_functional/conftest.py
+++ b/tests/python_functional/conftest.py
@@ -97,8 +97,8 @@ def testcase_parameters(request):
 
 
 @pytest.fixture
-def config(request, testcase_parameters):
-    return SyslogNgConfig(testcase_parameters.get_working_dir(), request.getfixturevalue("version"))
+def config(request):
+    return SyslogNgConfig(request.getfixturevalue("version"))
 
 
 @pytest.fixture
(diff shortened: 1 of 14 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
What's the exact reason behind creating a global variable from `testcase_parameters.get_working_dir()`?

mitzkia:
As I wrote in my first comment for now `working_dir` is an external dependency in several classes eg: SyslogNgConfig() https://github.com/balabit/syslog-ng/blob/cf10b3c4a50d8feda50826b518428719fd974a1c/tests/python_functional/src/syslog_ng_config/syslog_ng_config.py#L39-L40

I would like to eliminate this dependency:
 * to reduce constructor parameters in related classes
 * to be able to call/use related classes easier not "depending" on visibility of working_dir 
 * to be able to refactor related classes, I will continue this with SyslogNgConfig()

MrAnno:
Okay, thanks.

If you end up adding a second global variable related to "test config arguments", please consider introducing a class/dict/object containing those arguments, then you can inject this object where those parameters are needed.

Please rebase your branch, it contains patches from other PRs.

kira-syslogng:
Build SUCCESS

PR #2772, opened 2019-06-07
Dynamic fc balanced

lbudai:
# Dynamic window

## Motivation/problem to be solved: 
 * when a network connection has a high number of clients
 * AND only a small subset of the active clients sends messages at a high rate AND
 * there is a limited amount of memory, 
then it would be nice from syslog-ng if it could somehow dynamically re-calculate window sizes (periodically during runtime).

The proposed solution is dynamic-window.

Key points during implementation:
 * 100% backward compatibility when someone don't want to use
 * default is static only
 * dynamic windowing performance should be very close to the static case
 * starvation is not an option, so a static window is always guaranteed (`iw-size`/`max-connections`)

Keep in mind that main goal of this PR is to implement the 'framework' of dynamic windowing and to solve the main problem that is memory exhaustion. In a future PR we want to design&implement an algorithm that maximizes the ACK-rate (performance) while minimizes the window size (so the goal will be to find the minimum window size for the best performance).

## Algorithm: balanced dynamic window

Goal: being as simple as possible

* Target window size:
per-connection-window-size = static-part + dynamic-window-size / active-connections
static-part is constant, static-window-size/maxconn

So the per-connection window size is proportional to the number of active connections.
 
### Events:
 * client connects
 * client disconnects
 * rebalance timer elapsed

#### connect

* When a client is connected, a balanced window size(BWS) is re-calculated (decreasing), then we try to allocate a dynamic window for the new connection(size=BWS). When rebalance is initiated, each `LogSource` have to decrease the current_window_size by BWS amount of dynamic window slots.

There are two main cases:
 * free window >= requested_size : a `LogSource` is able to release immediately all of the requested slots
 * free window < requested size: release the difference immediately(when free window is not zero...), the remaining part will be reclaimed when the messages that are using those slots are ACKed

This means that during rebalance sometimes we need to to suppress the normal flow-control-window-adjustment. Suppress means that the source has to reclaim a bunch of window-slots (instead of increasing its own window) to the common dynamic-window pool.

#### disconnect
* When a client is disconnected, it releases its dynamic window, the balanced window size (BWS) is re-calculated (increasing), and all of the active connections will get this increasement during the next rebalance event.

#### rebalance

* When the rebalance-timer elapsed, rebalance is scheduled.
When a `LogReader` has a working I/O job, we cannot do a rebalance immediately, we need to schedule the rebalance *after* the fetch.

## options
* "Main" option
   * `dynamic-window-size(N)` : enable and set dynamic window
* Tuning parameters
   * `dynamic-window-stats-freq(x milliseconds)` : every x milliseconds syslog-ng check and update window stats
   * `dynamic-window-realloc-ticks(k)`           : after k window-stats-freq() dynamic window will be reallocated

Note: these options are part of the `framework`. We need to collect some statistics, then evaluate and reallocate. Currently the realloc is just a rebalance, so it's not based on statistics, but can be used for diagnostics/debugging.

Typical network source with dynamic window could be:

```
@version: 3.20
options {
 stats_level(2);
};

source s_network_dynamic {
  network(ip(0.0.0.0)
    port(61001)
    log_fetch_limit(1000)
    log_iw_size(10000)
    max_connections(10000)
    dynamic-window-size(100000)
    dynamic-window-stats-freq(1)
    dynamic-window-realloc-ticks(10)
    flags(no-parse) 
  );
};
```

Note: When dynamic-window-size() is set, log-iw-size() defaults to a lower value: 10 * max-connections(), but this value can be decreased further to 1 per connection.

## Changes in a nutshell

* added new counter for dynamic window (dynamic window pool)
* afsocket-source
   * added a timer and options
   * added shared dynamic window pool
   * min-iw-size-per-reader(1) when dynamic window is enabled
   * calculate balanced window size when client are connected/disconnected
* logreader
   * LogReader struct moved to the header
   * schedule realloc virtual method
* logsource
   * all the window logic (increase/decrease/release/reclaim/rebalance)
   * ack-tracker
      * use a linked list instead of fixed size array

## remaining tasks:
 * make dynamic-window related counters(pool, statistics) available via stats API
 * debug/warning/etc... messages
 * process some previous review notes
 * late-ack-tracker: for static window, we should use the fixed size Ringbuffer

### DIFF
--- a/lib/dynamic-window.c
+++ b/lib/dynamic-window.c
@@ -50,6 +50,9 @@ dynamic_window_stat_reset(DynamicWindow *self)
 gsize
 dynamic_window_stat_get_avg(DynamicWindow *self)
 {
+  if (self->window_stat.n == 0)
+    return 0;
+
   return self->window_stat.sum / self->window_stat.n;
 }
 
@@ -64,4 +67,3 @@ dynamic_window_stat_get_sum(DynamicWindow *self)
 {
   return self->window_stat.sum;
 }
-
--- a/lib/dynamic-window.c
+++ b/lib/dynamic-window.c
@@ -55,7 +55,7 @@ dynamic_window_stat_get_number_of_samples(DynamicWindowStat *self)
   return self->n;
 }
 
-gsize
+guint64
 dynamic_window_stat_get_sum(DynamicWindowStat *self)
 {
   return self->sum;
--- a/lib/dynamic-window.c
+++ b/lib/dynamic-window.c
@@ -64,24 +64,24 @@ dynamic_window_stat_get_sum(DynamicWindowStat *self)
 void
 dynamic_window_set_counter(DynamicWindow *self, DynamicWindowCounter *ctr)
 {
-  self->window_ctr = ctr;
+  self->ctr = ctr;
   dynamic_window_stat_reset(&self->stat);
 }
 
 gsize
 dynamic_window_request(DynamicWindow *self, gsize size)
 {
-  if (!self->window_ctr)
+  if (!self->ctr)
     return 0;
 
-  return dynamic_window_counter_request(self->window_ctr, size);
+  return dynamic_window_counter_request(self->ctr, size);
 }
 
 void
 dynamic_window_release(DynamicWindow *self, gsize size)
 {
-  if (!self->window_ctr)
+  if (!self->ctr)
     return;
 
-  dynamic_window_counter_release(self->window_ctr, size);
+  dynamic_window_counter_release(self->ctr, size);
 }
--- a/lib/dynamic-window.c
+++ b/lib/dynamic-window.c
@@ -40,14 +40,28 @@ dynamic_window_stat_update(DynamicWindow *self, gsize window)
   self->window_stat.n++;
 }
 
-void dynamic_window_stat_reset(DynamicWindow *self)
+void
+dynamic_window_stat_reset(DynamicWindow *self)
 {
   self->window_stat.sum = 0;
   self->window_stat.n = 0;
 }
 
-gsize dynamic_window_stat_get_avg(DynamicWindow *self)
+gsize
+dynamic_window_stat_get_avg(DynamicWindow *self)
 {
   return self->window_stat.sum / self->window_stat.n;
+}
+
+gsize
+dynamic_window_stat_get_number_of_samples(DynamicWindow *self)
+{
+  return self->window_stat.n;
+}
+
+gsize
+dynamic_window_stat_get_sum(DynamicWindow *self)
+{
+  return self->window_stat.sum;
 }
 
--- a/lib/dynamic-window.c
+++ b/lib/dynamic-window.c
@@ -62,32 +62,32 @@ dynamic_window_stat_get_sum(DynamicWindowStat *self)
 }
 
 void
-dynamic_window_set_counter(DynamicWindow *self, DynamicWindowPool *ctr)
+dynamic_window_set_pool(DynamicWindow *self, DynamicWindowPool *pool)
 {
-  self->ctr = ctr;
+  self->pool = pool;
   dynamic_window_stat_reset(&self->stat);
 }
 
 gboolean
 dynamic_window_is_enabled(DynamicWindow *self)
 {
-  return !!self->ctr;
+  return !!self->pool;
 }
 
 gsize
 dynamic_window_request(DynamicWindow *self, gsize size)
 {
-  if (!self->ctr)
+  if (!self->pool)
     return 0;
 
-  return dynamic_window_pool_request(self->ctr, size);
+  return dynamic_window_pool_request(self->pool, size);
 }
 
 void
 dynamic_window_release(DynamicWindow *self, gsize size)
 {
-  if (!self->ctr)
+  if (!self->pool)
     return;
 
-  dynamic_window_pool_release(self->ctr, size);
+  dynamic_window_pool_release(self->pool, size);
 }
--- a/lib/dynamic-window.c
+++ b/lib/dynamic-window.c
@@ -67,3 +67,21 @@ dynamic_window_stat_get_sum(DynamicWindow *self)
 {
   return self->window_stat.sum;
 }
+
+gsize
+dynamic_window_request(DynamicWindow *self, gsize size)
+{
+  if (!self->window_ctr)
+    return 0;
+
+  return dynamic_window_counter_request(self->window_ctr, size);
+}
+
+void
+dynamic_window_release(DynamicWindow *self, gsize size)
+{
+  if (!self->window_ctr)
+    return;
+
+  dynamic_window_counter_release(self->window_ctr, size);
+}
--- a/lib/dynamic-window.c
+++ b/lib/dynamic-window.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2002-2019 One Identity
  * Copyright (c) 2019 Laszlo Budai <laszlo.budai@balabit.com>
+ * Copyright (c) 2019 László Várady <laszlo.varady@balabit.com>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -22,27 +23,31 @@
  *
  */
 
-#ifndef DYNAMIC_WINDOW_H_INCLUDED
-#define DYNAMIC_WINDOW_H_INCLUDED
+#include <syslog-ng.h>
+#include "dynamic-window.h"
 
-#include "syslog-ng.h"
-#include "dynamic-window-counter.h"
-
-typedef struct _DynamicWindow DynamicWindow;
-
-struct _DynamicWindow
+void
+dynamic_window_set_counter(DynamicWindow *self, DynamicWindowCounter *ctr)
 {
-  DynamicWindowCounter *window_ctr;
-  struct
-  {
-    gsize n;
-    gsize sum;
-  } window_stat;
-};
+  self->window_ctr = ctr;
+  dynamic_window_stat_reset(self);
+}
 
-void dynamic_window_set_counter(DynamicWindow *self, DynamicWindowCounter *ctr);
-void dynamic_window_stat_update(DynamicWindow *self, gsize window);
-void dynamic_window_stat_reset(DynamicWindow *self);
-gsize dynamic_window_stat_get_avg(DynamicWindow *self);
+void
+dynamic_window_stat_update(DynamicWindow *self, gsize window)
+{
+  self->window_stat.sum += window;
+  self->window_stat.n++;
+}
 
-#endif
+void dynamic_window_stat_reset(DynamicWindow *self)
+{
+  self->window_stat.sum = 0;
+  self->window_stat.n = 0;
+}
+
+gsize dynamic_window_stat_get_avg(DynamicWindow *self)
+{
+  return self->window_stat.sum / self->window_stat.n;
+}
+
--- a/lib/dynamic-window.c
+++ b/lib/dynamic-window.c
@@ -27,45 +27,45 @@
 #include "dynamic-window.h"
 
 void
+dynamic_window_stat_update(DynamicWindowStat *self, gsize value)
+{
+  self->sum += value;
+  self->n++;
+}
+
+void
+dynamic_window_stat_reset(DynamicWindowStat *self)
+{
+  self->sum = 0;
+  self->n = 0;
+}
+
+gsize
+dynamic_window_stat_get_avg(DynamicWindowStat *self)
+{
+  if (self->n == 0)
+    return 0;
+
+  return self->sum / self->n;
+}
+
+gsize
+dynamic_window_stat_get_number_of_samples(DynamicWindowStat *self)
+{
+  return self->n;
+}
+
+gsize
+dynamic_window_stat_get_sum(DynamicWindowStat *self)
+{
+  return self->sum;
+}
+
+void
 dynamic_window_set_counter(DynamicWindow *self, DynamicWindowCounter *ctr)
 {
   self->window_ctr = ctr;
-  dynamic_window_stat_reset(self);
-}
-
-void
-dynamic_window_stat_update(DynamicWindow *self, gsize window)
-{
-  self->window_stat.sum += window;
-  self->window_stat.n++;
-}
-
-void
-dynamic_window_stat_reset(DynamicWindow *self)
-{
-  self->window_stat.sum = 0;
-  self->window_stat.n = 0;
-}
-
-gsize
-dynamic_window_stat_get_avg(DynamicWindow *self)
-{
-  if (self->window_stat.n == 0)
-    return 0;
-
-  return self->window_stat.sum / self->window_stat.n;
-}
-
-gsize
-dynamic_window_stat_get_number_of_samples(DynamicWindow *self)
-{
-  return self->window_stat.n;
-}
-
-gsize
-dynamic_window_stat_get_sum(DynamicWindow *self)
-{
-  return self->window_stat.sum;
+  dynamic_window_stat_reset(&self->stat);
 }
 
 gsize
--- a/lib/dynamic-window.c
+++ b/lib/dynamic-window.c
@@ -68,6 +68,12 @@ dynamic_window_set_counter(DynamicWindow *self, DynamicWindowCounter *ctr)
   dynamic_window_stat_reset(&self->stat);
 }
 
+gboolean
+dynamic_window_is_enabled(DynamicWindow *self)
+{
+  return !!self->ctr;
+}
+
 gsize
 dynamic_window_request(DynamicWindow *self, gsize size)
 {
--- a/lib/dynamic-window.c
+++ b/lib/dynamic-window.c
@@ -62,7 +62,7 @@ dynamic_window_stat_get_sum(DynamicWindowStat *self)
 }
 
 void
-dynamic_window_set_counter(DynamicWindow *self, DynamicWindowCounter *ctr)
+dynamic_window_set_counter(DynamicWindow *self, DynamicWindowPool *ctr)
 {
   self->ctr = ctr;
   dynamic_window_stat_reset(&self->stat);
@@ -80,7 +80,7 @@ dynamic_window_request(DynamicWindow *self, gsize size)
   if (!self->ctr)
     return 0;
 
-  return dynamic_window_counter_request(self->ctr, size);
+  return dynamic_window_pool_request(self->ctr, size);
 }
 
 void
@@ -89,5 +89,5 @@ dynamic_window_release(DynamicWindow *self, gsize size)
   if (!self->ctr)
     return;
 
-  dynamic_window_counter_release(self->ctr, size);
+  dynamic_window_pool_release(self->ctr, size);
 }
--- a/lib/dynamic-window.h
+++ b/lib/dynamic-window.h
@@ -35,14 +35,14 @@ typedef struct _DynamicWindowStat DynamicWindowStat;
 struct _DynamicWindowStat
 {
   gsize n;
-  gsize sum;
+  guint64 sum;
 };
 
 void dynamic_window_stat_update(DynamicWindowStat *self, gsize value);
 void dynamic_window_stat_reset(DynamicWindowStat *self);
 gsize dynamic_window_stat_get_avg(DynamicWindowStat *self);
 gsize dynamic_window_stat_get_number_of_samples(DynamicWindowStat *self);
-gsize dynamic_window_stat_get_sum(DynamicWindowStat *self);
+guint64 dynamic_window_stat_get_sum(DynamicWindowStat *self);
 
 struct _DynamicWindow
 {
--- a/lib/dynamic-window.h
+++ b/lib/dynamic-window.h
@@ -46,7 +46,7 @@ gsize dynamic_window_stat_get_sum(DynamicWindowStat *self);
 
 struct _DynamicWindow
 {
-  DynamicWindowCounter *window_ctr;
+  DynamicWindowCounter *ctr;
   DynamicWindowStat stat;
 };
 
--- a/lib/dynamic-window.h
+++ b/lib/dynamic-window.h
@@ -46,5 +46,7 @@ void dynamic_window_stat_reset(DynamicWindow *self);
 gsize dynamic_window_stat_get_avg(DynamicWindow *self);
 gsize dynamic_window_stat_get_number_of_samples(DynamicWindow *self);
 gsize dynamic_window_stat_get_sum(DynamicWindow *self);
+gsize dynamic_window_request(DynamicWindow *self, gsize size);
+void dynamic_window_release(DynamicWindow *self, gsize size);
 
 #endif
--- a/lib/dynamic-window.h
+++ b/lib/dynamic-window.h
@@ -26,7 +26,7 @@
 #define DYNAMIC_WINDOW_H_INCLUDED
 
 #include "syslog-ng.h"
-#include "dynamic-window-counter.h"
+#include "dynamic-window-pool.h"
 
 typedef struct _DynamicWindow DynamicWindow;
 
@@ -46,11 +46,11 @@ guint64 dynamic_window_stat_get_sum(DynamicWindowStat *self);
 
 struct _DynamicWindow
 {
-  DynamicWindowCounter *ctr;
+  DynamicWindowPool *ctr;
   DynamicWindowStat stat;
 };
 
-void dynamic_window_set_counter(DynamicWindow *self, DynamicWindowCounter *ctr);
+void dynamic_window_set_counter(DynamicWindow *self, DynamicWindowPool *ctr);
 gboolean dynamic_window_is_enabled(DynamicWindow *self);
 gsize dynamic_window_request(DynamicWindow *self, gsize size);
 void dynamic_window_release(DynamicWindow *self, gsize size);
--- a/lib/dynamic-window.h
+++ b/lib/dynamic-window.h
@@ -46,11 +46,11 @@ guint64 dynamic_window_stat_get_sum(DynamicWindowStat *self);
 
 struct _DynamicWindow
 {
-  DynamicWindowPool *ctr;
+  DynamicWindowPool *pool;
   DynamicWindowStat stat;
 };
 
-void dynamic_window_set_counter(DynamicWindow *self, DynamicWindowPool *ctr);
+void dynamic_window_set_pool(DynamicWindow *self, DynamicWindowPool *pool);
 gboolean dynamic_window_is_enabled(DynamicWindow *self);
 gsize dynamic_window_request(DynamicWindow *self, gsize size);
 void dynamic_window_release(DynamicWindow *self, gsize size);
--- a/lib/dynamic-window.h
+++ b/lib/dynamic-window.h
@@ -44,5 +44,7 @@ void dynamic_window_set_counter(DynamicWindow *self, DynamicWindowCounter *ctr);
 void dynamic_window_stat_update(DynamicWindow *self, gsize window);
 void dynamic_window_stat_reset(DynamicWindow *self);
 gsize dynamic_window_stat_get_avg(DynamicWindow *self);
+gsize dynamic_window_stat_get_number_of_samples(DynamicWindow *self);
+gsize dynamic_window_stat_get_sum(DynamicWindow *self);
 
 #endif
--- a/lib/dynamic-window.h
+++ b/lib/dynamic-window.h
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2002-2019 One Identity
+ * Copyright (c) 2019 Laszlo Budai <laszlo.budai@balabit.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#ifndef DYNAMIC_WINDOW_H_INCLUDED
+#define DYNAMIC_WINDOW_H_INCLUDED
+
+#include "syslog-ng.h"
+#include "dynamic-window-counter.h"
+
+typedef struct _DynamicWindow DynamicWindow;
+
+struct _DynamicWindow
+{
+  DynamicWindowCounter window;
+  struct
+  {
+    gsize n;
+    gsize sum;
+  } window_stat;
+};
+
+#endif
--- a/lib/dynamic-window.h
+++ b/lib/dynamic-window.h
@@ -30,22 +30,27 @@
 
 typedef struct _DynamicWindow DynamicWindow;
 
+typedef struct _DynamicWindowStat DynamicWindowStat;
+
+struct _DynamicWindowStat
+{
+  gsize n;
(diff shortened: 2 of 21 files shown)


bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 07ded14340d1ccb0f595e413310a0141e5249ed3 into 9fcbe16217e6133c0aff3d2b23c92f4e7a4bf282 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-e2f182d566a2e3d9f6ba28896b0143c052ee2539)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 271b5b0239cca348df4c6fd6f11ab622854052be into 9fcbe16217e6133c0aff3d2b23c92f4e7a4bf282 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-711d8e514e59acef19693f969b63d671d7fa1d47)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

lib/logsource.c
```diff
@@ -44,10 +44,38 @@ log_source_wakeup(LogSource *self)
...
+
+  if (old > 0)
+    {
+      if (old - window_size_increment >= 0)
+        {
+          window_size_increment = 0; // all the acks goes back to the common window pool
+        }
+      else
+        {
+          reclaimed_dec = window_size_increment - old;
```
alltilla: If I understand right, the `reclaimed_dec` variable stores the number of ACKs (window_size_increments), which we give to the dynamic pool.
We run to this `else` branch, when the `old` (window size needed for the dynamic pool) is less, than the ACKs we received. In this case we give back `old` to the dynamic pool, and not `window_size_increment - old`.
Please correct me, if I understand it wrong.

alltilla: I prepared a little refactor for this code also, with the problem fixed, too.
```
static inline guint32
_take_reclaimed_window(LogSource *self, guint32 window_size_increment)
{
  guint32 remaining_window_size_increment = window_size_increment;
  gssize old = atomic_gssize_sub(&self->window_size_to_be_reclaimed, window_size_increment);
  gboolean need_to_reclaim = (old > 0);

  if (!need_to_reclaim)
    return remaining_window_size_increment;

  remaining_window_size_increment = MAX(window_size_increment - old, 0);

  if (remaining_window_size_increment)
    atomic_gssize_add(&self->pending_reclaimed, old);
  else
    atomic_gssize_add(&self->pending_reclaimed, window_size_increment);

  return remaining_window_size_increment;
}
```

modules/afsocket/afsocket-grammar.ym
```diff
@@ -386,6 +389,10 @@ source_afsocket_stream_params
 	: KW_KEEP_ALIVE '(' yesno ')'		{ afsocket_sd_set_keep_alive(last_driver, $3); }
 	| KW_MAX_CONNECTIONS '(' positive_integer ')'	 { afsocket_sd_set_max_connections(last_driver, $3); }
 	| KW_LISTEN_BACKLOG '(' positive_integer ')'	{ afsocket_sd_set_listen_backlog(last_driver, $3); }
+	| KW_DYNAMIC_WINDOW_SIZE '(' nonnegative_integer ')' { afsocket_sd_set_dynamic_window_size(last_driver, $3); }
+  | KW_DYNAMIC_WINDOW_STATS_FREQ '(' nonnegative_integer ')' { afsocket_sd_set_dynamic_window_stats_freq(last_driver, $3); }
+  | KW_DYNAMIC_WINDOW_STATS_FREQ '(' LL_FLOAT ')' { afsocket_sd_set_dynamic_window_stats_freq(last_driver, $3); }
+  | KW_DYNAMIC_WINDOW_REALLOC_TICKS '(' nonnegative_integer ')' { afsocket_sd_set_dynamic_window_realloc_ticks(last_driver, $3); }
```
alltilla: `FREQ` and `REALLOC_TICKS` should be positive, I believe. Also, `LL_FLOAT` will allow to have negative float numbers here.

lib/dynamic-window.c
```diff
@@ -0,0 +1,93 @@
...
+}
+
+guint64
+dynamic_window_stat_get_sum(DynamicWindowStat *self)
+{
+  return self->sum;
+}
+
+void
+dynamic_window_set_counter(DynamicWindow *self, DynamicWindowPool *ctr)
```
alltilla: `counter` and `ctr` should be renamed to `pool`. (1/7)

lib/dynamic-window.h
```diff
@@ -0,0 +1,58 @@
...
+
+void dynamic_window_stat_update(DynamicWindowStat *self, gsize value);
+void dynamic_window_stat_reset(DynamicWindowStat *self);
+gsize dynamic_window_stat_get_avg(DynamicWindowStat *self);
+gsize dynamic_window_stat_get_number_of_samples(DynamicWindowStat *self);
+guint64 dynamic_window_stat_get_sum(DynamicWindowStat *self);
+
+struct _DynamicWindow
+{
+  DynamicWindowPool *ctr;
```
alltilla: `ctr` should be renamed to `pool`. (2/7)

lib/logsource.c
```diff
@@ -169,6 +197,153 @@ log_source_flow_control_suspend(LogSource *self)
   window_size_counter_suspend(&self->window_size);
 }
 
+void
+log_source_enable_dynamic_window(LogSource *self, DynamicWindowPool *window_ctr)
```
alltilla: `window_ctr` should be renamed to `window_pool`. (3/7)

lib/tests/test_dynamic_window.c
```diff
@@ -0,0 +1,55 @@
...
+
+Test(dynamic_window, window_stat_reset)
+{
+  DynamicWindow win;
+  dynamic_window_stat_reset(&win.stat);
+  cr_expect_eq(win.stat.sum, 0);
+  cr_expect_eq(win.stat.n, 0);
+}
+
+Test(dynamic_window, window_stat_reset_when_counter_is_set)
```
alltilla: `counter` should be renamed to `pool`. (4/7)

modules/afsocket/afsocket-source.c
```diff
@@ -691,6 +879,40 @@ afsocket_sd_save_listener(AFSocketSourceDriver *self)
...
+  else
+    {
+      dynamic_window_pool_unref(self->dynamic_window_ctr);
+    }
+
+  self->dynamic_window_ctr = NULL;
+}
+
+static gboolean
+afsocket_sd_restore_dynamic_window_ctr(AFSocketSourceDriver *self)
```
alltilla: `ctr` should be renamed to `pool`. (5/7)

modules/afsocket/afsocket-source.c
```diff
@@ -691,6 +879,40 @@ afsocket_sd_save_listener(AFSocketSourceDriver *self)
...
+{
+  GlobalConfig *cfg = log_pipe_get_config(&self->super.super.super);
+
+  if (!self->connections_kept_alive_across_reloads)
+    return FALSE;
+
+  DynamicWindowPool *ctr = cfg_persist_config_fetch(cfg, afsocket_sd_format_dynamic_window_pool_name(self));
```
alltilla: `ctr` should be renamed to `pool`. (6/7)

modules/afsocket/afsocket-source.h
```diff
@@ -40,8 +41,14 @@ struct _AFSocketSourceDriver
...
           window_size_initialized:1;
   struct iv_fd listen_fd;
+  struct iv_timer dynamic_window_timer;
+  gsize dynamic_window_size;
+  gsize dynamic_window_timer_tick;
+  gfloat dynamic_window_stats_freq;
+  gint dynamic_window_realloc_ticks;
   gint fd;
   LogReaderOptions reader_options;
+  DynamicWindowPool *dynamic_window_ctr;
```
alltilla: `dynamic_window_ctr` should be renamed to `dynamic_window_pool`. (7/7)

alltilla:
The equation in the description is misleading:
```
Target window size:
per-connection-window-size = 1/active_connections * (static-window-size + dynamic-window-size)
```
It suggests, that the `static-window-size` is divided by the active connections, which, I think, is not true.
I would change the equation to:
```
Target window size:
per-connection-window-size = static-window-size / max_connections + dynamic-window-size / active-connections
```

lib/dynamic-window.c
alltilla: I found 7 function or variable declarations, which were not renamed to pool from counter. I marked all of the comments with (x/7), so it is easier to track them.

lib/logsource.c
```diff
@@ -169,6 +197,153 @@ log_source_flow_control_suspend(LogSource *self)
...
+  g_assert(self->ack_tracker == NULL);
+
+  gsize dynamic_part = self->full_window_size - self->options->init_window_size;
+  msg_trace("Releasing dynamic part of the window", evt_tag_int("dynamic_window_to_be_released", dynamic_part),
+            log_pipe_location_tag(&self->super));
+  self->full_window_size -= dynamic_part;
+  window_size_counter_sub(&self->window_size, dynamic_part, NULL);
+  dynamic_window_release(&self->dynamic_window, dynamic_part); //TODO: rename release to ...
+
+  dynamic_window_pool_unref(self->dynamic_window.ctr); //TODO: move to dynamic_window_release_counter()
```
alltilla: Do you plan to address these TODOs? If not, please remove the comments.

lib/logsource.c
```diff
@@ -169,6 +197,153 @@ log_source_flow_control_suspend(LogSource *self)
...
+  if (empty_window <= subtrahend)
+    {
+      remaining_sub = subtrahend - empty_window;
+      if (empty_window == 0)
+        {
+          subtrahend = 0;
+        }
+      else
+        {
+          subtrahend = empty_window - 1;
```
alltilla: Where does this `1` come from? Shouldn't this be the `static-window-size`?

MrAnno: `- 1` is here to avoid suspending the source on one more code path. The new `window_size` would be 0 without `- 1`).

The suspend-wakeup logic looks simple, but its task ordering and threading implications  are complicated enough, that's why I did not want to add another suspend call.

alltilla: Ok, thanks! :)

lib/logsource.c
alltilla: We can further simplify this by replacing
```
  if (remaining_window_size_increment)
    atomic_gssize_add(&self->pending_reclaimed, old);
  else
    atomic_gssize_add(&self->pending_reclaimed, window_size_increment);
```
with
```
  atomic_gssize_add(&self->pending_reclaimed, window_size_increment - remaining_window_size_increment);
```

kira-syslogng:
Build FAILURE

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 8d8495fa27b463c7fa92951180a618ae015fbf1c into f3298e8d187c5b2ac9a89755d3af79595b5553b2 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-326a00c846524f0cbe7c8b725c576d5ae324d319)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

kira-syslogng:
Build FAILURE

lib/logsource.c
```diff
@@ -44,10 +44,38 @@ log_source_wakeup(LogSource *self)
   msg_diagnostics("Source has been resumed", log_pipe_location_tag(&self->super));
 }
 
+static inline guint32
+_take_reclaimed_window(LogSource *self, guint32 window_size_increment)
+{
+  gssize old = atomic_gssize_sub(&self->window_size_to_be_reclaimed, window_size_increment);
```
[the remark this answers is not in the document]
lbudai: Yep... and the counter will underflow (1M EPS) cca 290000 years...
Of course this number is true only for 64bits machines, so we will fix it.

lib/logsource.c
```diff
@@ -169,6 +197,153 @@ log_source_flow_control_suspend(LogSource *self)
...
+
+  gsize old_window_size = window_size_counter_add(&self->window_size, offered_dynamic, NULL);
+  if (old_window_size == 0 && offered_dynamic != 0)
+    log_source_wakeup(self);
+}
+
+static void
+_dec_balanced(LogSource *self, gsize dec)
+{
+  gsize new_full_window_size = MAX(self->options->init_window_size, dec);
```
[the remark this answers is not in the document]
lbudai: This should be a dead code. Decrement by dec should be enough (so this subtrahend/new-full-window-size "dance" is not required, check the callsite, current-dynamic-window counter is already decremented by the init-window-size
Thanks for finding this :)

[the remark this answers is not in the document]
lbudai: This should be a dead code. Decrement by dec should be enough (so this subtrahend/new-full-window-size "dance" is not required, check the callsite, current-dynamic-window counter is already decremented by the init-window-size
Thanks for finding this :)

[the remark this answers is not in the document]
lbudai: This should be a dead code. Decrement by dec should be enough (so this subtrahend/new-full-window-size "dance" is not required, check the callsite, current-dynamic-window counter is already decremented by the init-window-size
Thanks for finding this :)

lib/logsource.c
lbudai: @altilla:  you are right, but I think that else branch is not possible : cannot get more ACKs (I need to check again with pen&paper).

lbudai: @altilla:  you are right, but I think that else branch is not possible : cannot get more ACKs (I need to check again with pen&paper).

lib/dynamic-window.c
lbudai: Thx! (sed...)

lib/late_ack_tracker.c
```diff
@@ -136,17 +136,50 @@ late_ack_tracker_track_msg(AckTracker *s, LogMessage *msg)
...
+
+  Bookmark *bookmark = &(record->bookmark);
+  bookmark->save(bookmark);
+}
+
+static void
+_ack_records_delete_range(LateAckTracker *self, GList *last, guint32 ack_range_length)
+{
+  GList *delete_head = self->ack_records.head;
+  self->ack_records.head = last->next;
```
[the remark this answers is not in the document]
lbudai: In glib, the GList data structure contains prev but GLib does not use. So I won't update.

lib/logsource.c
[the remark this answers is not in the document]
Kokan: I did not do the calculation (should known better), that makes it less worrisome; and **optional**.

lib/logsource.h
```diff
@@ -80,6 +80,7 @@ struct _LogSource
 
   void (*wakeup)(LogSource *s);
   void (*dynamic_window_realloc)(LogSource *s);
```
MrAnno: This commit should remove this method in favor of `schedule_dynamic_window_realloc`.

lib/logsource.c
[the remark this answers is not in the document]
MrAnno: I've validated atomic operations (dec, inc, reclaim), `MAX(init_window_size, ...)` is incorrect here, my suggestion is the following:
```C
static void
_dec_balanced(LogSource *self, gsize dec)
{
  gsize new_full_window_size = self->full_window_size - dec;

  gsize empty_window = window_size_counter_get(&self->window_size, NULL);
  gsize remaining_sub = 0;
  if (empty_window <= dec)
    {
      remaining_sub = dec - empty_window;
      if (empty_window == 0)
        {
          dec = 0;
        }
      else
        {
          dec = empty_window - 1;
        }
      new_full_window_size = self->full_window_size - dec;
      _reclaim_dynamic_window(self, remaining_sub);
    }

  window_size_counter_sub(&self->window_size, dec, NULL);

  msg_warning("Balance::decrease",
              log_pipe_location_tag(&self->super),
              evt_tag_printf("connection", "%p", self),
              evt_tag_int("old_full_window_size", self->full_window_size),
              evt_tag_int("new_full_window_size", new_full_window_size),
              evt_tag_int("to_be_reclaimed", remaining_sub));

  self->full_window_size = new_full_window_size;
  dynamic_window_release(&self->dynamic_window, dec);
}
```

lib/late_ack_tracker.c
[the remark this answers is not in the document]
lbudai: hmm... I was wrong, so I'm going to update it :)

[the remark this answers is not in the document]
Kokan: would it be possible to create an *internal* glib like function for this, as at this level I do not care about prev, next.

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 8942616eac5f71cf149ea6b056c2ef55b41e9c71 into b362324ca0bf14c60bc7fb4412273b6628349114 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-e0ad5abfdcf936bf7e42937a616f080f080169dd)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

lib/logsource.c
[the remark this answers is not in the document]
lbudai: @Kokan : but it's easy to fix and we still support 32bit machines ;)

lbudai:
@alltilla , @Kokan , @MrAnno 
I think I addressed all the important review notes.
I've added fixup commits, please check them, then I'd rebase the branch.

kira-syslogng:
Build FAILURE

lib/logsource.c
```diff
@@ -44,11 +45,35 @@ log_source_wakeup(LogSource *self)
...
+  gboolean need_to_reclaim = (old > 0);
+
+  if (!need_to_reclaim)
+    {
+      atomic_gssize_sub(&self->window_size_to_be_reclaimed, old);
+      return window_size_increment;
+    }
+
+  guint32 remaining_window_size_increment = MAX(window_size_increment - old, 0);
+  guint32 pending_increment = window_size_increment - remaining_window_size_increment;
```
MrAnno: Nice :)

lib/logsource.c
```diff
@@ -44,11 +45,35 @@ log_source_wakeup(LogSource *self)
...
 
+static inline guint32
+_take_reclaimed_window(LogSource *self, guint32 window_size_increment)
+{
+  gssize old = atomic_gssize_sub(&self->window_size_to_be_reclaimed, window_size_increment);
+  gboolean need_to_reclaim = (old > 0);
+
+  if (!need_to_reclaim)
+    {
+      atomic_gssize_sub(&self->window_size_to_be_reclaimed, old);
```
MrAnno: What does this mean? I though the inverse of `atomic_gssize_sub(&self->window_size_to_be_reclaimed, window_size_increment)` is needed here, something like:

`atomic_gssize_add(&self->window_size_to_be_reclaimed, window_size_increment);`

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 5501312e63d0079e23ce55e21e41cbbf4e57f43f into b362324ca0bf14c60bc7fb4412273b6628349114 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-57e59bcb98682cee5a2bea1e92bd29fbb24254d8)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

lib/logsource.c
lbudai: I think we don't need the inverse... this is about avoiding underflow.
What you suggest is to keep this value on its original value (which is already a _negative_ value, as `old` need to be less than 0). What I do is trying to keep the value 0... But I realized, that this is the wrong place where I'm doing this, so thanks for the question!

There are two competing thread:
a) main thread: timer, sets the `window_size_to_be_reclaimed` when needed _and_ `pending_reclaimed` is 0 (which means that all the requested window slots has been reclaimed)

b) a worker thread (where the ACKs are coming)
In this situation the three (sub/set/add) atomic instructions are 'racy', I mean it is possible that first we decrease the old value (`pending_reclaimed` is 0!) from the worker thread, main thread checks `pending_reclaimed`, it is 0, sets the `window_size_to_be_reclaimed`, then we _increase_ its value... 

The solution is to set it to 0 in the main thread (in the timer callback). When the `pending_reclaimed` is 0 _and_ the `window_size_to_be_reclaimed` is less than 0 _and_ we don't have to reclaim window slots, we have to set the `window_size_to_be_reclaimed` to 0. With this change we can avoid underflow...

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lbudai:

kira-syslogng:
Build FAILURE

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging fbd9b255f485342fe9bfd67fa59965bd5cea421a into b362324ca0bf14c60bc7fb4412273b6628349114 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-5dbec44aceca624e6e0923535074a5a8acd07412)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

lib/tests/test_dynamic_window.c
alltilla: This remained the same after the fixup commits. Also 3 lines below, instead of `dynamic_window_set_counter()`, the function `dynamic_window_set_pool()` should be used,

modules/afsocket/afsocket-source.c
alltilla: This remained the same after the fixup commits.

modules/afsocket/afsocket-source.h
alltilla: This remained the same after the fixup commits.

modules/afsocket/afsocket-source.c
alltilla: This remained the same after the fixup commits. Also please fix the `afsocket_sd_save_dynamic_window_ctr()` function name please (I missed that in my last review :( )

lib/logsource.c
alltilla: Kindly fix this in the header as well.

modules/afsocket/afsocket-grammar.ym
alltilla: This was not changed in the fixup commits. Was it accidental or you don't agree with my suggestion?

lib/logsource.c
```diff
@@ -311,10 +310,11 @@ _reclaim_window_instead_of_rebalance(LogSource *self)
...
       dynamic_window_release(&self->dynamic_window, total_reclaim);
-      if ((gssize)atomic_gssize_get(&self->window_size_to_be_reclaimed) > 0)
-        {
-          reclaim_in_progress = TRUE;
-        }
+      gssize to_be_reclaimed = (gssize)atomic_gssize_get(&self->window_size_to_be_reclaimed);
+      reclaim_in_progress = to_be_reclaimed > 0;
+      //to avoid underflow, we need to set a value <= 0
+      if (to_be_reclaimed < 0)
+        atomic_gssize_set(&self->window_size_to_be_reclaimed, 0);
```
MrAnno: This only reduces the possibility of an overflow, because `_take_reclaimed_window()` (the function containing the subtraction) is called for every ACK even when reclaim is not in progress, while this sets `window_size_to_be_reclaimed` to 0 only when `pending_reclaimed` is nonzero.

I think `atomic_gssize_set(&self->window_size_to_be_reclaimed, 0);` should be moved out of the `if (total_reclaim > 0)` branch.

kira-syslogng:
Build FAILURE

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 4ae86eef125613ff6d45e8076ef1caf8c58eb535 into 18b69c763bac1cac33067ad1385f6c51cc39303e - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-c165aabb85428185afcaaca44909f572cafdb8da)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

kira-syslogng:
Build FAILURE

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 6a337039d99ddb0806c72436ddb4d6fbb4b30b3e into d2d0df30e03a3047775c8dc442c110112e1eaa9b - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-0f4113a5bb30c3c048491b136ccf102729ed3adb)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

kira-syslogng:
Build SUCCESS

alltilla:
I thought about a corner case:

Let's imagine, that the user sets `dynamic-window-size(200)` and `max-connections(101)` (we can see later, that they can set it to 200, the problem can still happen).
At some point in time, there are 100 connections, so they all get 2 window each. Then another connection is established, and as we can't give every connection 2 window each, they all get 1.
So with that one new connection we dropped from using 200 windows to using only 101, this is nearly 50% drop.
So this can happen with `max-connections(200)` also, as both 100 and 101 connections can be established with that option, but it should be more likely to cause a problem with `max-connections(101)`, as with `max-connections(200)` every new connection would make the situation better, which is not possible in the other case.

We talked about this with @lbudai , and we had multiple suggestions for avoiding/soothing this situation:
 1. If we set the `dynamic-window-size()` to at least a 10 or so-ish multiple of `max-connections()` we drastically drop the impact of the problem. The higher the better.
 2. We can tweak this even better, with syncing these 2 options, by changing `dynamic-window-size()` to be a multiple of `max-connections()`. This will help us, as we can expect to have `current-connections == max-connections()` for a longer period of time, and we can optimize that state to not have unused dynamic windows.

MrAnno:
@kira-syslogng do perftest

kira-syslogng:
Build SUCCESS

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 0eea6a8348bb423094c7f39285bb369edaffe215 into d1b2f02825d050c2ded89a43bd5ebac4caea5d2a - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-0a6084f83a330424602f7b0d59b2966e02c3880c)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

alltilla:
> I thought about a corner case:
> 
> Let's imagine, that the user sets `dynamic-window-size(200)` and `max-connections(101)` (we can see later, that they can set it to 200, the problem can still happen).
> At some point in time, there are 100 connections, so they all get 2 window each. Then another connection is established, and as we can't give every connection 2 window each, they all get 1.
> So with that one new connection we dropped from using 200 windows to using only 101, this is nearly 50% drop.
> So this can happen with `max-connections(200)` also, as both 100 and 101 connections can be established with that option, but it should be more likely to cause a problem with `max-connections(101)`, as with `max-connections(200)` every new connection would make the situation better, which is not possible in the other case.
> 
> We talked about this with @lbudai , and we had multiple suggestions for avoiding/soothing this situation:
> 
> 1. If we set the `dynamic-window-size()` to at least a 10 or so-ish multiple of `max-connections()` we drastically drop the impact of the problem. The higher the better.
> 2. We can tweak this even better, with syncing these 2 options, by changing `dynamic-window-size()` to be a multiple of `max-connections()`. This will help us, as we can expect to have `current-connections == max-connections()` for a longer period of time, and we can optimize that state to not have unused dynamic windows.

A quick graph to represent these values:
https://www.desmos.com/calculator/mj2g9kwnbe
This shows `dynamic-window-size(20000)` and we can imagine `max-connections()` wherever we want.
We can see, that the drops always happen, when the `dynamic-window-size` is divisible by the number of current connections and a new connection comes. The drop is `current-connections - 1`. So as we said in the first suggestion, a relatively small `max-connections()` with magnitudes larger `dynamic-window-size()` should work well.
The second suggestion is also helpful, as we can stop accepting new connections, so the existing connections have the ideal window-size.

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging 6437ed2e7e641c81c1c2054c0c1570618f3b15ff into d1b2f02825d050c2ded89a43bd5ebac4caea5d2a - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-9141ab8f52404dc88994acf10759615fda18fe12)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

kira-syslogng:
Build SUCCESS

bazsi:
This pull request **introduces 3 alerts** and **fixes 2** when merging abcad4895316353712880f9bc5c3030512a3638d into f6aa5e6e6551c50474b831bb710dd96f5db972a5 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-3f2bd777b70c5d89162503e279a552b947363532)

**new alerts:**

**fixed alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng do perftest

kira-syslogng:

PR #2778, opened 2019-06-12
logwriter: Avoid null pointer access from suspended writer

MrAnno:
I've accidentally closed #2645 with a stale push.

Original author: @kyeongy

> There is a chance with rare timing that a suspended writer tries to
access already freed client, which causes a SEGV.

https://github.com/balabit/syslog-ng/pull/2645#issuecomment-477816858
https://github.com/balabit/syslog-ng/pull/2645#issuecomment-500114948
https://github.com/balabit/syslog-ng/pull/2645#issuecomment-500254227
https://github.com/balabit/syslog-ng/pull/2645#issuecomment-500420282

### DIFF
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -1120,7 +1120,7 @@ log_writer_flush_finalize(LogWriter *self)
   return FALSE;
 }
 
-gboolean
+static gboolean
 log_writer_write_message(LogWriter *self, LogMessage *msg, LogPathOptions *path_options, gboolean *write_error)
 {
   gboolean consumed = FALSE;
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng do perftest

kira-syslogng:

MrAnno:
@kyeongy Thanks again for catching and fixing this crash!

lib/logwriter.c
```diff
@@ -547,6 +545,8 @@ log_writer_stop_watches(LogWriter *self)
 
       self->watches_running = FALSE;
     }
+
+  log_writer_stop_suspend_timer(self);
```
Kokan: Before in the `log_writer_io_handler` did not called `log_writer_stop_suspend_timer` unconditional. Just leave the note, as every other missing place, you first introduced the `log_writer_stop_suspend_timer`.

But I think it is correct to stop the `suspend_timer` there, and even required.

PR #2780, opened 2019-06-14
geoip: remove deprecated module, geoip2 database location detection

### LINKED ISSUE #2776: remove support for legacy maxmind
# Issue

syslog-ng support legacy maxmind and current geoip versions. Because Maxmind has ended maint for legacy formats only geoip should be retained

Kokan:
The legacy geoip has been deprecated in both syslog-ng (since `3.11`) and as API as well.

Fixes #2776 

Also additionally it tries to detect the geoip2 database location, if it can find in those locations the *database* is no longer a required option. Detection was tested on ubuntu:bionic, fedora:30, archlinux.


kira-syslogng:
Build FAILURE

bazsi:
is it possible to provide a compatible $(geoip) template function with the maxminddb based functionality (e.g. by the geoip2 module?)

As I see the only difference is that geoip2 requires a path to the database, which could probably be made default.

furiel:
I am afraid not all distributions package maxminddb packages, as opposed to the legacy database. At least I could not find such package for the new database format in ubuntu xenial. That's why I originally made it a mandatory option.

In the libgeoip code there is a compiler switch for `GEOIPDATADIR` that is defaulted to `$(pkgdatadir)`, however I cannot see similar functionality in libmaxminddb. That's why I tend to think there is no intended default path in case of the new version.

furiel:
Now that geoip module name became available, did you consider a rename also? geoip2 -> geoip?

Kokan:
@furiel thanks to bring it up; I considered, and here is my reasoning not to do.

The `geoip` and `geoip2` different in configuration, and I do not see a way to provide some compat layer. As such, if we remove `geoip`, that is an easy thing. It is obsolete and have been removed.

1) I was thinking if it is possible to provide a nicer map for the users (maybe I am overthinking stuff).
As first step remove `geoip`, and promote to the users to use `geoip2` (they get the feel and deal with `geoip2`; also maintainer could decide to keep `geoip` if they want to with an easy patch)
2) When everybody got used to `geoip2` is the new `geoip`, and nobody dreams about `geoip` (LTS distros does not have it), we depricate `geoip2` (uh it sound bad I know), and redirect to `geoip`
3) Remove `geoip2` as a configuration option and long live `geoip`

But this creates a gap preparing a nice transition period for the users. (and honestly I am not so nice, I just want to avoid issues with invalid `geoip` config, or not working `geoip`)

Please give a feedback about this plan, as I could convince myself but it was not easy to do so; and maybe I did in fact overthink stuff so we could go with straight `geoip` is the new thing.

Kokan:
> I am afraid not all distributions package maxminddb packages, as opposed to the legacy database. At least I could not find such package for the new database format in ubuntu xenial. That's why I originally made it a mandatory option.
> 
> In the libgeoip code there is a compiler switch for `GEOIPDATADIR` that is defaulted to `$(pkgdatadir)`, however I cannot see similar functionality in libmaxminddb. That's why I tend to think there is no intended default path in case of the new version.

Even if we do not have *intended default path*, we could look for main distribution. I was thinking - after your summary about why it is needed - to look up some path during configure time, and burn that path into. Other possibility is to make that lookup runtime at `init` phase (that is not that costly to do). Even if those strategies fails, we can provide an error like we could not find default db, please provide one.

furiel:
Both plans sounds good to me!

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

furiel:
Looks good to me. Once the legacy test that is failing is removed, approve from my side.

Kokan:
@kira-syslogng retest this please; branch=remove-geoip

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=remove-geoip;

kira-syslogng:
Build SUCCESS

PR #2781, opened 2019-06-14
Fix escaped quote in block argument

### LINKED ISSUE #2777: block regex argument results in syntax error
# syslog-ng

## Version of syslog-ng

OSE 3.21

## Platform

RHEL 7.6

# Issue

## Failure

While trying to create a config block that expects a regex as one of its parameters, I encountered a problem, that I wasn't able to work around.
The configuration snippets posted below result in syntax errors in config parsing time, but they should work just fine.

## Configuration

```
@version: 3.21
@include "scl.conf"
block root dummy-block (
		loopback-socket-path("`syslog-ng-root`/var/run/message-loopback.sock")
		loopback-name(loopback_internal)
		feeder-source()
		return-destination()
		field-name()
		content-filter()
    )
{
		log {
			filter {
				not match( "`content-filter`" value("list_element"));
			};
		};
};

dummy-block(
	feeder-source(s_feeder)
	return-destination(d_return)
	field-name("EVENT_USERNAME")
	content-filter("([^\\\/\:\*\?\"\<\>\|]{1,15}|[^\,\~\:\!\@\#\$\%\^\&\'\.\(\)\{\}\_ ]{2,255})\\[^\"\/\\\[\]\:\;\|\=\,\+\*\?\<\>]+")
)
```

A shortened version of the config, that produces the same error:

```

@version: 3.21

block root dummy-block (
      content-filter("")
)
{
	log {
		filter {
			not match( "`content-filter`" value("list_element"));
		};
	};
};

dummy-block (
      content-filter("[^\"]")
)
```

bazsi:
This should fix #2777

### DIFF
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -390,11 +390,12 @@ word	[^ \#'"\(\)\{\}\[\]\\;\r\n\t,|\.@:]
                              _cfg_lex_extend_token_location_to_next_line(yyextra);
                            }
 
-    /* continuation line within blocks: for now an incorrect behavior to not
-     * cause functional change in this patch.  Drop the continuation line
-     * from the block value while still updating the location info.  */
+    /* continuation line within blocks: in the block header we just ignore
+     * the continuation line, whereas in string/qstring we retain them
+     * literally.  */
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

PR #2782, opened 2019-06-14
debun: collect kafka properties file

mehul-m-prajapati:
Resolves https://github.com/mehul-m-prajapati/gsoc-debun-tool/issues/6

### DIFF
--- a/contrib/syslog-ng-debun
+++ b/contrib/syslog-ng-debun
@@ -846,7 +846,7 @@ acquire_syslog_all () {
 	acquire_syslog_var
 	acquire_syslog_ldinfo
 	acquire_syslog_startup_method
-
+	acquire_syslog_config_files
 }
 
 acquire_syslog_nprv () {
@@ -855,6 +855,47 @@ acquire_syslog_nprv () {
 	acquire_syslog_info
 	acquire_syslog_ldinfo
 	acquire_syslog_startup_method
+}
+
+remove_ssl_passwords_from_file () {
+	#
+	# Double quotes are used because escaping single quotes within a single quoted string is ugly.
+	# The goal is matching substrings that look like this, to replace the password string, and preserve original formatting:
+	# ssl.keystore.password='password-string' -->  ssl.keystore.password='___PASSWORD_REMOVED___'
+	# ssl.key.password='password-string' -->  ssl.key.password='___PASSWORD_REMOVED___'
+	# ssl.truststore.password='password-string' -->  ssl.truststore.password='___PASSWORD_REMOVED___'
+	###
+	
+	${sed_equivalent_cmd} "s|ssl.keystore.password=([\"']?)[^)\"']+([\"']?)|ssl.keystore.password=\1___PASSWORD_REMOVED___\2|g" <"${1}" >"${1}.bak1" 2>/dev/null
+	${sed_equivalent_cmd} "s|ssl.key.password=([\"']?)[^)\"']+([\"']?)|ssl.key.password=\1___PASSWORD_REMOVED___\2|g" <"${1}.bak1" >"${1}.bak2" 2>/dev/null
+	${sed_equivalent_cmd} "s|ssl.truststore.password=([\"']?)[^)\"']+([\"']?)|ssl.truststore.password=\1___PASSWORD_REMOVED___\2|g" <"${1}.bak2" >"${1}.bak3" 2>/dev/null
+	mv "${1}.bak3" "${1}"	
+	rm -f "${1}".bak*
+}
+
+copy_kafka_properties_file() {
+	while read FILE; do
+		if [ -f "$FILE" ]; then
+			filename=$(echo ${FILE} | ${sed_equivalent_cmd} "s|[\/]|_|g")
+			kafka_file="syslog-ng.kafka.properties_${filename}"
+			cp "${FILE}" "${tmpdir}/${kafka_file}" && remove_ssl_passwords_from_file "${tmpdir}/${kafka_file}"
+		fi
+	done
+}
+
+acquire_kafka_properties_file () {
+	if [ -f "${tmpdir}/syslog-ng.ctl.running.conf" ]; then	
+		${sed_equivalent_cmd} "s|#.*||g" ${tmpdir}/syslog-ng.ctl.running.conf | perl -pe "s|[\r\n\f\v]||" | grep -oP "kafka([ \t\r\n\v\f]*)\((.*?)\};" | grep -oP "properties-file\((.*?)\)" | ${sed_equivalent_cmd} "s|properties-file||g" | ${sed_equivalent_cmd} "s|([\t]*)([\`\"'\(\)]?)||g" | copy_kafka_properties_file
+	else
+		find ${tmpdir}/config -type f -name "*.conf*" | \
+		while read FILE; do \
+			${sed_equivalent_cmd} "s|#.*||g" $FILE | perl -pe "s|[\r\n\f\v]||" | grep -oP "kafka([ \t\r\n\v\f]*)\((.*?)\};" | grep -oP "properties-file\((.*?)\)" | ${sed_equivalent_cmd} "s|properties-file||g" | ${sed_equivalent_cmd} "s|([\t]*)([\`\"'\(\)]?)||g" | copy_kafka_properties_file
+		done
+	fi
+}
+
+acquire_syslog_config_files () {
+	acquire_kafka_properties_file
 }
 
 fhs_set_linux () {
@@ -1680,5 +1721,6 @@ if [ -n "$privacy_mode" ]; then
 else
 	acquire_syslog_all
 fi
+
 [ -n "$debug_mode" ] && run_debug
 debun_final


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
Build SUCCESS

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

contrib/syslog-ng-debun
```diff
@@ -857,6 +857,39 @@ acquire_syslog_nprv () {
...
+	${sed_equivalent_cmd} "s|ssl.key.password([ \t\r\n\v\f]*)\(([ \t\r\n\v\f]*)([\"']?)[^)\"']+([\"']?)([ \t\r\n\v\f]*)\)|ssl.key.password\1(\2\3___PASSWORD_REMOVED___\4\5)|g" <"${1}.bak1" >"${1}.bak2" 2>/dev/null
+	${sed_equivalent_cmd} "s|ssl.truststore.password([ \t\r\n\v\f]*)\(([ \t\r\n\v\f]*)([\"']?)[^)\"']+([\"']?)([ \t\r\n\v\f]*)\)|ssl.truststore.password\1(\2\3___PASSWORD_REMOVED___\4\5)|g" <"${1}.bak2" >"${1}.bak3" 2>/dev/null
+	mv "${1}.bak3" "${1}"	
+	rm -f "${1}".bak*
+}
+
+acquire_kafka_properties_file () {
+	
+	if [ -f "${tmpdir}/syslog-ng.ctl.running.conf" ]; then
+		prop_file=$( sed -n "/kafka/,/);/ p" ${tmpdir}/syslog-ng.ctl.running.conf | grep properties-file | sed 's/.*(\(.*\))/\1/' | sed 's/"//g' )
```
jszigetvari: I would recommend using ${sed_equivalent_cmd} instead of sed, but grep will likely be okay.

contrib/syslog-ng-debun
```diff
@@ -857,6 +857,39 @@ acquire_syslog_nprv () {
...
+	else
+		find ${tmpdir}/config -name "*.conf*" | \
+		while read FILE; do \
+			prop_file=$( sed -n "/kafka/,/);/ p" $FILE | grep properties-file | sed 's/.*(\(.*\))/\1/' | sed 's/"//g' )
```
jszigetvari: Here too.

contrib/syslog-ng-debun
```diff
@@ -1679,6 +1712,8 @@ if [ -n "$privacy_mode" ]; then
 	acquire_syslog_nprv
 else
 	acquire_syslog_all
+	acquire_kafka_properties_file
```
jszigetvari: I would probably create a wrapper subroutine here for gathering similar property files, not only kafka.
Of course with your current contribution it would be the only subroutine called there, but it could be extended easily later on.

mehul-m-prajapati: Okay. Thanks for review

contrib/syslog-ng-debun
mehul-m-prajapati: @jszigetvari : I am using below command to fetch properties file from ```kafka( ... )``` config.

kira-syslogng:
Build SUCCESS

contrib/syslog-ng-debun
jszigetvari: @mehul-m-prajapati 
> If I replace 1st part sed -n "/kafka/,/);/ p" $FILE with ${sed_equivalent_cmd} , it breaks all the sequence.

What do you mean exactly? Does it result in a syntax error, or does it produce incorrect output?
I will check it in larger detail tomorrow, and will get back to you.

Kokan:
Could you please verify this solution with a following cases ?

```
kafka(
  properties-file("quoted-file-path")
  properties-file('different-type-of-quote')
);

kafka(
  properties-file(multiple-kafka)
);

# it is not on an error collecting these, would be better to ignore
#kafka(
#  properties-file(commented-section)
#);

kafka( properties-file(kafka-without-indentation) );

kafka(properties-file(kafka-without-borders));

block destination franz(properties-file("")
{
   #this actually it does not need to obtain
   #(cannot even without parsing the config file itself)
   #just a case where the content of properties-files are not valid
   kafka(
     properties-file("`properties-file`")
   );
};

kafka( #withouth properties-file
);

```

mehul-m-prajapati:
@Kokan : Thanks for these test cases. I have tested all of them and made some changes in the script.
1) The debun script will get files which are enclosed in single or double quotes.
```
kafka(
  properties-file("quoted-file-path")
  properties-file('different-type-of-quote')
);
```

2) It will ignore commented section
```
# it is not on an error collecting these, would be better to ignore
#kafka(
#  properties-file(commented-section)
#);
```

3)

contrib/syslog-ng-debun
mehul-m-prajapati: @jszigetvari : Okay. I have made some changes now.

kira-syslogng:
Build SUCCESS

alltilla:
@jszigetvari Could you kindly check, if you are happy with the changes for your requests?

jszigetvari:
If we manage to overcome the properites file naming problem (Currently the properties file can be named anything, and will make it difficult to process the bundle automatically, as we can't rely on how it will be called within the bundle), and the `sed -E` problem, then I'll approve this MR as well.

jszigetvari:
If we manage to overcome the properites file naming problem (Currently the properties file can be named anything, and will make it difficult to process the bundle automatically, as we can't rely on how it will be called within the bundle), and the `sed -E` problem, then I'll approve this MR as well.

mehul-m-prajapati:
@jszigetvari : Thanks for your suggestions. I will make the changes in my patch accordingly.

jszigetvari:
@mehul-m-prajapati You are very welcome! In the meantime I will go on holiday/PTO for over a week, and my availability will be pretty limited during that time. I will be returning on the 12th of July.

kira-syslogng:
Build SUCCESS

mehul-m-prajapati:
> If we manage to overcome the properites file naming problem (Currently the properties file can be named anything, and will make it difficult to process the bundle automatically, as we can't rely on how it will be called within the bundle), and the `sed -E` problem, then I'll approve this MR as well.

@jszigetvari : Now, I have used `sed -E` and `grep -oP` with regex to fetch kafka file. Also, I have tested this patch with `perl -p -e` as well and it works.

For file naming, I have used `syslog-ng.kafka.properties_` as prefix. Let me know if you have any better suggestion.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

jszigetvari:
@mehul-m-prajapati Okay, the things I asked for have been taken care of.
The file properties file name may be problematic later on, because if there are multiple property files with the same filename, then only the last one will be stored. This is not a particularly big problem, and I can take care of it later.

Soi here is my +1

contrib/syslog-ng-debun
```diff
@@ -1679,6 +1721,8 @@ if [ -n "$privacy_mode" ]; then
 	acquire_syslog_nprv
 else
 	acquire_syslog_all
+	acquire_all_files
```
furiel: This would be be better moved inside acquire_syslog_all function. That would leave the branching on `privacy_mode` clean.

contrib/syslog-ng-debun
```diff
@@ -857,6 +857,48 @@ acquire_syslog_nprv () {
...
+		fi
+	done
+}
+
+acquire_kafka_properties_file () {
+	if [ -f "${tmpdir}/syslog-ng.ctl.running.conf" ]; then	
+		prop_file=$(${sed_equivalent_cmd} "s|#.*||g" ${tmpdir}/syslog-ng.ctl.running.conf | perl -pe "s|[\r\n\f\v]||" | grep -oP "kafka([ \t\r\n\v\f]*)\((.*?)\};" | grep -oP "properties-file\((.*?)\)" | ${sed_equivalent_cmd} "s|properties-file||g" | ${sed_equivalent_cmd} "s|([ \t]*)([\`\"'\(\)]?)||g")
+		copy_kafka_properties_file "$prop_file"
+	else
+		find ${tmpdir}/config -name "*.conf*" | \
```
furiel: `-type f` would be handy here.

contrib/syslog-ng-debun
mehul-m-prajapati: @furiel : ok

mehul-m-prajapati:
@jszigetvari : There is one more thing which I found during testing. If there are multiple kafka properties file, and either file-path has spaces. Then, this patch removes all spaces from that path resulting in an invalid path. 
For example,
```
properties-file("/home/mehul/Desktop/kafka_prop  ")
properties-file("   /home/mehul/Desktop/kafka.prope rti es   ")
```
So, instead of removing all spaces, I should remove all trailing and initial spaces. 

My concern is, the result of `sed -E` command is stored in a variable which might have two or more file-paths. I am iterating this variable over a for loop which automatically separates the multiple file-paths considering space character as a delimiter.

Somehow, I need to find a way to get a list of paths having newline as a delimiter.

kira-syslogng:
Build SUCCESS

furiel:
> Somehow, I need to find a way to get a list of paths having newline as a delimiter.

@mehul-m-prajapati is it allowed to have spaces around the path name? I mean, can syslog-ng start and use it successfully? If not, then I think you do not need to deal with the starting and terminating spaces. That could save some complexity from your code.

The behaviour certainly depends on the implementation (which I did not check), but for example `open` does not work with spaces in front:

```
#include <stdio.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>

int main()
{
  printf("AAAA: %d\n", open("/home/furiel/tmp/test.c", 'r'));
  printf("BBBB: %d\n", open("  /home/furiel/tmp/test.c", 'r'));
}
```

results in

```
AAAA: 3
BBBB: -1

```

jszigetvari:
@mehul-m-prajapati Yes, your observation is quite right. The for loop is not that much suitable in this situation.
What I would recommend, would be to modify the copy_kafka_properties file function so that it would expect its input through STDIN, and it would basically consist of a while read FILE loop, and would process the file list through the standard input.
Its invokation would also become easier as the function would be run as the last element of a series of piped commands, thereby eliminating the necessity for the prop_file variable as well.

What do you think? I believe that this would be the easiest solution for your problem.

mehul-m-prajapati:
> > Somehow, I need to find a way to get a list of paths having newline as a delimiter.
> 
> @mehul-m-prajapati is it allowed to have spaces around the path name? I mean, can syslog-ng start and use it successfully? If not, then I think you do not need to deal with the starting and terminating spaces. That could save some complexity from your code.
> 
> The behaviour certainly depends on the implementation (which I did not check), but for example `open` does not work with spaces in front:
> 
> ```
> #include <stdio.h>
> #include <sys/types.h>
> #include <sys/stat.h>
> #include <fcntl.h>
> 
> int main()
> {
>   printf("AAAA: %d\n", open("/home/furiel/tmp/test.c", 'r'));
>   printf("BBBB: %d\n", open("  /home/furiel/tmp/test.c", 'r'));
> }
> ```
> 
> results in
> 
> ```
> AAAA: 3
> BBBB: -1
> ```

@furiel : Yes, you are right. I don't need to handle this trailing and leading spaces.

mehul-m-prajapati:
> What I would recommend, would be to modify the copy_kafka_properties file function so that it would expect its input through STDIN, and it would basically consist of a while read FILE loop, and would process the file list through the standard input.
> Its invokation would also become easier as the function would be run as the last element of a series of piped commands, thereby eliminating the necessity for the prop_file variable as well.
> 
> What do you think? I believe that this would be the easiest solution for your problem.

@jszigetvari : That is an awesome idea. I have implemented exactly as you suggested now.
Also, to deal with identical filenames at different locations, I have included an absolute path in the name of kafka file.  Please review these changes.

kira-syslogng:
Build SUCCESS

PR #2785, opened 2019-06-17
logsource: create AckTracker in log_source_init

lbudai:
I think it is better to open a dedicated PR for this small change.

### DIFF
--- a/lib/logproto/logproto-server.h
+++ b/lib/logproto/logproto-server.h
@@ -195,6 +195,7 @@ void log_proto_server_free(LogProtoServer *s);
 static inline void
 log_proto_server_set_ack_tracker(LogProtoServer *s, AckTracker *ack_tracker)
 {
+  g_assert(ack_tracker);
   s->ack_tracker = ack_tracker;
 }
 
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build FAILURE

lbudai:

kira-syslogng:
Build SUCCESS

lbudai:
@bazsi:  you are right, the motivation is that...

* short version: 
... I need two types of "ringbuffer" (or, `AckRecordContainer`) and I don't want to add a 9th parameter (`enable-dynamic-window`) to `log_source_options()`

* long version
In case of dynamic windowing we cannot use a fixed-size `Ringbuffer`, and I changed the code that uses a `list<AckRecord>` unconditionally, so even in the static case. I think it is problem as this linked list based implementation has a negative performance impact (and I think, it is possible to eliminate, or decrease the impact, but not now - due to lack of time). One of the key point was during the dynamic-window development is to guarantee that the static case won't be affected the changes, we want to be sure that we won't break anything with the new feature. Flow-control is like blood in human body: implementation is not too big(code of lines...), but it is everywhere, like a fractal, this is why we want to be very careful.
So I'm going to open another PR soon which is about refactoring the `LateAckTracker` and make it possible to create two kind of trackers: one uses a fixed size `Ringbuffer` as an `AckRecordStore`, the other will use the linked list based `AckRecordStore`.

PR #2786, opened 2019-06-18
[WIP] stats type to const gchar *

furiel:
This is a competing alternative to https://github.com/balabit/syslog-ng/pull/2761.

Module specific SCS_ constants disappear. The lookup strings are passed by modules during registration.

Review warmly welcome, but please let me do the MERGE, because this has two followup tasks that need to be scheduled properly.

Some key points for reviewers:
- As future improvement, we wanted to support users to provide their own strings. This suggests some kind of dynamic support, but this alternative (as opposed to #2761) makes this harder, because implementation needs to ensure the string to register lives through statistics (this time of writing module deinit is not supported), possibly statically allocated. This is nothing impossible to resolve, just a bit inconvenient.
- I needed to modify the `stats_cluster_hash` to imitate original behaviour. Please give it some thought that we hare having something that is equally useful.
- SCS_GROUP needed to be addressed specifically, please check last patch.

### DIFF
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -455,7 +455,7 @@ afinter_message_posted(LogMessage *msg)
 
       stats_lock();
       StatsClusterKey sc_key;
-      stats_cluster_logpipe_key_set(&sc_key, SCS_GLOBAL, "internal_queue_length", NULL );
+      stats_cluster_logpipe_key_set(&sc_key, SCS_GLOBAL, SCS_NONE, "internal_queue_length", NULL );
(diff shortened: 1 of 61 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

furiel:
In some cases we expect that the "registered" string literals reside in the same memory address. This holds quite high risk on the framework. Strcmp would have performance impact. Hence closing this alternative.

PR #2784, opened 2019-06-17
ctags: add ctags config (for excluding dbld...)

lbudai:
motivation: sometimes when I forget to remove dbld/install (always...),
then ctags parses dbld/install/{header-files} and I'm finding myself in
a sitation when I don't understand why my changes
 a) even compiles at the very first attempt (first, I'm happy, later it
becomes suspicious)
 b) don't have any effects... this could be very-very frustrating

Signed-off-by: Laszlo Budai <laszlo.budai@balabit.com>

PR #2794, opened 2019-06-21
gitignore: tags file and build directory

Kokan:
Inspired by https://github.com/balabit/syslog-ng/pull/2784 

well I use the `build` directory to build shame on me, but `ag` could exclude this directory from searching IF it is listed in `.gitignore`
the other `tags` as well
(I know I could create my own ignore file, but well..)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@ dbld/rules.conf
 debian/build-tree/
 debian/*.log
 debian/autoreconf.*
+tags
--- a/.gitignore
+++ b/.gitignore
@@ -46,4 +46,5 @@ dbld/rules.conf
 debian/build-tree/
 debian/*.log
 debian/autoreconf.*
+build/
 tags


furiel:
You can create global gitignore file. It works perfectly for me. Add something similar to you `.gitconfig.`

```
[core]
	excludesfile = /home/furiel/.gitignore_global
```

kira-syslogng:
Build FAILURE

Kokan:
@furiel yeah I know, but that's what I want to avoid, I prefer to track in the same repository what I do not want to track (keep your enemies local).

lbudai:
After .ctags is added, on today I also realized that I should had to add the .tags file to .gitignore :)
thanks

lbudai:

kira-syslogng:
Build SUCCESS

szemere:
Two other workarounds for this situation:

A)
**.gitignore** files could ignore themselves. (Unfortunately it is too late for us, to put it in the root directory, since we are already version tracking a .gitignore file there.)

```
syslog-ng$: ls -a foo/
.gitignore
bar

syslog-ng$: cat foo/.gitignore
.gitignore
bar
foobar
```
Neither `foo/bar.foo` or `foo/.gitignore` will show up in the output of a `git status` command.

B)
edit the `.git/info/exclude` file directly.


Version A is more transparent, but it only works with subdirectories. Version B works together with the already existing .gitignore file, but it smells a little bit.

PR #2798, opened 2019-06-24
affile: Add option to disable time-reap()

alltilla:
Fixing 2 problems with `time-reap()`

The first problem: `time-reap()` can only be set globally, for all destinations.
The second problem: It can not be disabled. We should be able to disable it, if we are sure, that no problematic templating is set for that destination.

Example usage:
```
@version: 3.21

options {
    time-reap(2); # global time-reap()
};

source s_input {
    example-msg-generator(freq(5));
};

destination d_fifo {
    pipe(
        "/tmp/test.fifo",
        time-reap(0)  # disable time-reap() for this destination
    );
};

log {
    source(s_input);
    destination(d_fifo);
};
```

The PR also introduces a new functionality: If the `pipe()` or `file()` destination's filepath does not have a template, the `time-reap()` is not inherited from the global options, but is disabled, by default. The user can override this by explicitly setting `time-reap()` to non-zero at that destination.

So the hierarchy is the following:
 1. if `time-reap()` is set at the destination, we will use that value
 2. if `time-reap()` is not set at the destination, and the destination does not have a template in its filepath option, we disable `time-reap()`
 3. if `time-reap()` is not set at the destination, and the destination has a template in its filepath, the global option is inherited, which is 60 by default, but can be overridden at `options { ... };`

### DIFF
--- a/modules/affile/affile-grammar.ym
+++ b/modules/affile/affile-grammar.ym
@@ -249,7 +249,7 @@ dest_affile_option
 	| KW_CREATE_DIRS '(' yesno ')'		{ affile_dd_set_create_dirs(last_driver, $3); }
 	| KW_OVERWRITE_IF_OLDER '(' nonnegative_integer ')'	{ affile_dd_set_overwrite_if_older(last_driver, $3); }
 	| KW_FSYNC '(' yesno ')'		{ affile_dd_set_fsync(last_driver, $3); }
-	| KW_TIME_REAP '(' positive_integer ')'		{ affile_dd_set_time_reap(last_driver, $3); }
+	| KW_TIME_REAP '(' nonnegative_integer ')'		{ affile_dd_set_time_reap(last_driver, $3); }
 	;
 
 dest_afpipe_params
@@ -271,7 +271,7 @@ dest_afpipe_option
 	: dest_writer_option
 	| dest_driver_option
 	| file_perm_option
-	| KW_TIME_REAP '(' positive_integer ')'		{ affile_dd_set_time_reap(last_driver, $3); }
+	| KW_TIME_REAP '(' nonnegative_integer ')'		{ affile_dd_set_time_reap(last_driver, $3); }
 	;
 
 
--- a/modules/affile/affile-grammar.ym
+++ b/modules/affile/affile-grammar.ym
@@ -69,6 +69,7 @@ LogProtoFileReaderOptions *last_log_proto_options;
 %token KW_MULTI_LINE_MODE
 %token KW_MULTI_LINE_PREFIX
 %token KW_MULTI_LINE_GARBAGE
+%token KW_TIME_REAP
 
 %token KW_WILDCARD_FILE
 %token KW_BASE_DIR
@@ -248,6 +249,7 @@ dest_affile_option
 	| KW_CREATE_DIRS '(' yesno ')'		{ affile_dd_set_create_dirs(last_driver, $3); }
 	| KW_OVERWRITE_IF_OLDER '(' nonnegative_integer ')'	{ affile_dd_set_overwrite_if_older(last_driver, $3); }
 	| KW_FSYNC '(' yesno ')'		{ affile_dd_set_fsync(last_driver, $3); }
+	| KW_TIME_REAP '(' positive_integer ')'		{ affile_dd_set_time_reap(last_driver, $3); }
 	;
 
 dest_afpipe_params
@@ -269,6 +271,7 @@ dest_afpipe_option
 	: dest_writer_option
 	| dest_driver_option
 	| file_perm_option
+	| KW_TIME_REAP '(' positive_integer ')'		{ affile_dd_set_time_reap(last_driver, $3); }
 	;
 
 
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -968,7 +968,7 @@ options_item
 	| KW_CHECK_HOSTNAME '(' yesno ')'	{ configuration->check_hostname = $3; }
 	| KW_BAD_HOSTNAME '(' string ')'	{ cfg_bad_hostname_set(configuration, $3); free($3); }
 	| KW_TIME_REOPEN '(' positive_integer ')'		{ configuration->time_reopen = $3; }
-	| KW_TIME_REAP '(' positive_integer ')'		{ configuration->time_reap = $3; }
+	| KW_TIME_REAP '(' nonnegative_integer ')'		{ configuration->time_reap = $3; }
 	| KW_TIME_SLEEP '(' nonnegative_integer ')'	{}
 	| KW_SUPPRESS '(' nonnegative_integer ')'		{ configuration->suppress = $3; }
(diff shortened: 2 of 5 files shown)


alltilla:
Whoops, I pushed the branch `time_reap_disable` to `balabit/syslog-ng` instead of my fork. Let's not forget to remove the stale branch, if the PR gets merged.

kira-syslogng:
Build FAILURE

alltilla:
KIRA fails because of an issue, which is not relevant to this PR. I will rerun it, when it is fixed.

MrAnno:

kira-syslogng:
Build FAILURE

mitzkia:

kira-syslogng:
Build FAILURE

MrAnno:

kira-syslogng:
Build SUCCESS

modules/affile/affile-grammar.ym
```diff
@@ -269,6 +271,7 @@ dest_afpipe_option
 	: dest_writer_option
 	| dest_driver_option
 	| file_perm_option
+	| KW_TIME_REAP '(' nonnegative_integer ')'		{ affile_dd_set_time_reap(last_driver, $3); }
 	;
```
bazsi: I don't think it its applicable to pipe()

MrAnno: The PR was inspired by a mailing list question. Someone had problems using `pipe()`, because `time-reap()` closed the pipe.

I think, in that case, we should just disable the time-reap functionality in case of named pipes.
`time-reap(0)` IMHO can be useful in other places: non-templated file destinations, other file-based internal modules.

alltilla: I did not want to remove existing functionality (time-reap for pipe). I am totally okay with removing it, though, that would solve the user's issue, too. Up to you.

bazsi:
Hmm I am not against making this configurable, however if its about
disabling time-reap() for cases where the destination is not using macros,
I would simply disable reaping automatically in that case.

There's code to decide if a destination is macro based or not.

Bazsi

Attila Szakacs <notifications@github.com> ezt írta (időpont: 2019. jún.
24., H 20:46):

kira-syslogng:
Build FAILURE

alltilla:

alltilla:
@bazsi:
I added a commit, which disables `time-reap()` for non-templated filename destinations by default.
The user still has the option, to turn it on, and still has the option to turn off for templated filename destinations. This applies for both `pipe()` and `file()`.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

abundantior:
Note that this change breaks the expected behavior of some installs that perform an operation on a file destination, like an in-place sed, where time-reap had relied on the global default.  It may be that using sed on a log file is bad practice, but the result is that the file destination becomes a blank file, while syslog-ng continues to log to the old file, lost in space.

alltilla:
@abundantior If I understand correctly, you have a non-templated file destination, and you call `sed` on that file. Can you specify when did you call `sed`? Before or after time-reap? Was the `sed` triggered by time-reap somehow? Are you renaming the file?
Sorry, if i misunderstood something.

abundantior:
A number of people with Asus routers running the Merlin firmware have implemented syslog-ng instead of syslogd, using it to separate the usual system log into specific log files:  one is for an ad blocker, another is for a malware and ip blocker, and so on.  these destinations are in the usual file destination format.  The malware and ip blocker hourly runs a script that operates on its specific log file with an in-place sed that summarizes the messages for the past hour.  That script takes about 10 seconds to run, summarizing messages that are written every 3 to 4 seconds, deleting them, and replacing them with a single line, and then restarting the process.

Before this change, this operated as expected.  After this change, the specific law file on the hour became and remained empty.  Using the documentation of the change, if the file destination is defined with a time-reap element, the prior pattern is restored.  If the file destination is defined to use a specific template (and specifically one restating the default message template, the prior pattern is not restored.  I confess I don't understand what a "non-templated file destination" is, but it appears our file definitions are falling into the second category, where the default time-reap is not now applied and the file remains open forever.

alltilla:
@abundantior If I understand right, your script relied on `skynet-0.log` being closed, while the script runs. It was indeterminate before, because even though `time-reap` was set by default, if the logs arrived in a high frequency, the file was not closed/reaped.

What do you think about the following idea:
 1. Change the destination to `file("/opt/var/log/skynet-0-${HOUR}.log");`. This will create a new file every hour.
 2. Your aggregating script could be modified to run at every hour and 5 minutes, operating on the previous hour's log. It could even remove the file.

abundantior:
Yes, I think the script was incorrect before as well.  Before, the file was closed after 60 seconds, and the script was completing some times, and not others, depending on incoming messages. After, using the original destination definition, the file was never closed and the script never completed.  That is when we noticed.

The problem, I think, occurs when sed -i wants to rename its temporary file back to the original file name, and can't, because syslog-ng is still holding the file open.

I think your suggestion would work because syslog-ng would never be writing to a file the script is working on.  The difficulty is the script needs to keep changing the filepath of the log file it is looking for, and the script also needs to accomodate those who use syslogd instead of syslog-ng.

So, I think our work around is to use time-reap(2) or (1) in the file definition.  Or, find a way for syslog-ng to close the file after every write.  I don't know if that is a performance issue.

Many thanks!

bazsi:
The pseudofile() destination will open and close the file for all the
writes, but it doesn't support macros in the file name.

It was originally meant to be used for manipulating /proc files.

szemere:
Hello,
 is logrotate available on that platform? Syslog-ng works with it quite well, hopefully rsyslog too. And your maintenance script should only work on a limited variety of files.

Br,
Laci

abundantior:
Yes, logrotate is on the platform and works well.  That raises an interesting question.  If my problem with `sed -i` arises when it wants to rename its temporary file back to the log file, and it fails because syslog-ng is holding the file open until the time-reap period runs, how does logrotate not have the same problem when it wants to create an new log file with the same name?  How does logrotate work on a file destination that syslog-ng is holding open?

bazsi:
Logrotate sends a sighup (or syslog-ng-ctl reopen-files) to syslog-ng after
it renames the old file.

abundantior:
>
> Yes, I think you do that with a postrotate command to restart syslog-ng.
> I meant, how does logrotate overwrite a file that syslog-ng has open?  In
> some way that sed -i can't.

bazsi:
The point is that sed needs to rename the file and send syslog-ng a signal
to reopen the file with the old name. This way logs are not lost. This is
the sequence:

* Logrotate renames the file that syslog-ng has open, syslog-ng is still
writing to this (renamed file)
* Logrotate sends syslog-ng the signal, causing it to close the file and
reopen it, from which point the old is not written to
* Logrotate can then do anything it pleases (compress etc)

Your script needs to do the rename-sighup dance a similar way.

abundantior:
I get it now.  Thanks!

PR #2802, opened 2019-06-25
Revert "packaging/debian: merge rdkafka and kafka modules"

lbudai:
This reverts commit 6f252f4b6bb171ae4ff5949e5539dedc7ba5734a.

I'm not sure that we really want to merge them... when someone installs the native mod-rdkafka, I'm sure that this user don't want to install the whole Java ecosystem (and even if this what the users want, then we have to _replace_ the previous mod-rdkafka by mod-kafka). With this patch we 'broke' current installations (mod-rdkafka has version 3.21). 

Of course I'm open to discuss it, but I have a very strong opinion on this question :)

### DIFF
--- a/packaging/debian/syslog-ng-mod-kafka.install
+++ b/packaging/debian/syslog-ng-mod-kafka.install
@@ -1,3 +1,2 @@
 usr/lib/syslog-ng/*/java-modules/kafka.jar
 usr/share/syslog-ng/include/scl/kafka/*
-usr/lib/syslog-ng/*/libkafka.so
--- a/packaging/debian/Makefile.am
+++ b/packaging/debian/Makefile.am
@@ -69,6 +69,7 @@ EXTRA_DIST		+= \
 		packaging/debian/syslog-ng-mod-java-http.install \
 		packaging/debian/syslog-ng-mod-examples.install \
 		packaging/debian/syslog-ng.systemd \
+		packaging/debian/syslog-ng-mod-rdkafka.install \
 		packaging/debian/copyright \
 		packaging/debian/MPL \
 		packaging/debian/not-installed \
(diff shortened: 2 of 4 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
ok, good point about having to install java for the rdkafka stuff. That should only be done if we abandon the java part.

btw, I feel that our debian build is a bit too modular (like add-contextual-data in a separate package). But this way, we will need to find a place for the SCL that switches between the java/native implementation.

packaging/debian/syslog-ng-mod-kafka.install
```diff
@@ -1,3 +1,2 @@
 usr/lib/syslog-ng/*/java-modules/kafka.jar
 usr/share/syslog-ng/include/scl/kafka/*
```
bazsi: the kafka scl needs to go to the core package in this case, as one would need to install both rdkafka and the java based kafka implementation to get the switching between the two implementations.

MrAnno: I don't fully understand the switching part (I've seen the `kafka-implementation` define and subst), but if `scl/kafka/plugin.conf` will be moved to the core package and the default implementation is the Java version, then a `@requires` pragma should be added for Java.

gaborznagy: To be able to have the switching between the implementations as @bazsi said, I guess the kafka SCL has to be split then, right?
We cannot put the `kafka-java` part into the core, as then the core will depend on java also.
What about putting it into `scl/scl.conf` file? It already contains `java-module-dir` definition.

gaborznagy:
AFAIK we don't want to merge this, yet.

PR #2804, opened 2019-06-26
[WIP] Light: Refactor src/syslog_ng_config/ related codebase

mitzkia:
This patch-set is mostly a refactor on src/syslog_ng_config/ codebase with some 
additional minor features.
Some feature is configuration related, others are not.
I have updated and extends functional tests to use features introduced in this patch-set.
I have tried to create small, functionally connected commits.

My goal is to give a flexible and consistent API for configuration based code.
Maybe this will help us later to move forward from this state to a second part of configuration
refactor.

### DIFF
--- a/tests/python_functional/functional_tests/destination_drivers/file_destination/test_fd_positional_option.py
+++ b/tests/python_functional/functional_tests/destination_drivers/file_destination/test_fd_positional_option.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+#############################################################################
+# Copyright (c) 2015-2019 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+import pytest
+from pathlib2 import Path
+
+import src.testcase_parameters.testcase_parameters as tc_parameters
+
+
+def test_fd_positional_option_missing(config, syslog_ng):
+    dummy_source = config.create_dummy_source()
+    file_destination = config.create_file_destination(file_name="skip")
+    config.create_logpath(statements=[dummy_source, file_destination])
+
+    with pytest.raises(Exception) as excinfo:
+        syslog_ng.start(config)
+    assert "syslog-ng can not started due to config syntax error" in str(excinfo.value)
+
+
+def test_fd_positional_option_empty(config, syslog_ng, generate_msgs_for_dummy_src_and_file_dst):
+    dummy_source = config.create_dummy_source()
+    file_destination = config.create_file_destination(file_name="")
+    config.create_logpath(statements=[dummy_source, file_destination])
+
+    input_messages, _ = generate_msgs_for_dummy_src_and_file_dst
+    dummy_source.write_log(input_messages[1])
+
+    syslog_ng.start(config)
+
+    assert syslog_ng.wait_for_console_log(message="Error opening file for writing; filename='', error='No such file or directory (2)'")
+
+
+def test_fd_positional_option_custom_dir(config, syslog_ng, generate_msgs_for_dummy_src_and_file_dst):
+    custom_dir = Path(tc_parameters.WORKING_DIR, "custom_dir")
+    custom_dir.mkdir()
+    dummy_source = config.create_dummy_source()
+    file_destination = config.create_file_destination(file_name=str(custom_dir))
+    config.create_logpath(statements=[dummy_source, file_destination])
+
+    input_messages, _ = generate_msgs_for_dummy_src_and_file_dst
+    dummy_source.write_log(input_messages[1])
+
+    syslog_ng.start(config)
+
+    assert syslog_ng.wait_for_console_log(message="Error opening file for writing")
+
+
+def test_fd_positional_option_custom_file(config, syslog_ng):
+    dummy_source = config.create_dummy_source()
+    file_destination = config.create_file_destination(file_name="output.log")
+    config.create_logpath(statements=[dummy_source, file_destination])
+
+    syslog_ng.start(config)
+
+
+def test_fd_positional_option_auto_file(config, syslog_ng):
+    dummy_source = config.create_dummy_source()
+    file_destination = config.create_file_destination()
+    config.create_logpath(statements=[dummy_source, file_destination])
+
+    syslog_ng.start(config)
--- a/tests/python_functional/functional_tests/source_drivers/file_source/test_fs_positional_option.py
+++ b/tests/python_functional/functional_tests/source_drivers/file_source/test_fs_positional_option.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python
+#############################################################################
+# Copyright (c) 2015-2019 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+import pytest
+from pathlib2 import Path
+
+import src.testcase_parameters.testcase_parameters as tc_parameters
+
+
+def test_fs_positional_option_missing(config, syslog_ng):
+    # source source_24638 { file (); };
+    file_source = config.create_file_source(file_name="skip")
+    dummy_destination = config.create_dummy_destination()
+    config.create_logpath(statements=[file_source, dummy_destination])
+
+    with pytest.raises(Exception) as excinfo:
+        syslog_ng.start(config)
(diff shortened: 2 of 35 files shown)


kira-syslogng:
Build FAILURE

folti:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

mitzkia:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
```diff
@@ -65,10 +65,10 @@ def write_config(self, config_path):
     def set_version(self, version):
         self.__syslog_ng_config["version"] = version
 
-    def add_include(self, include):
+    def set_include(self, include):
```
alltilla: I liked `add_include()`, as the function itself `append`s, not `update`s.

tests/python_functional/src/syslog_ng_config/statements/sources/file_source.py
```diff
@@ -32,8 +33,13 @@ def __init__(self, file_name=None, **options):
         self.driver_name = "file"
         self.group_type = "source"
         self.options = options
-        if file_name:
-            self.positional_option = Path(tc_parameters.WORKING_DIR, file_name)
+        if file_name is not None:
+            if file_name == "":
+                self.positional_option = "''"
+            elif file_name == "auto":
```
alltilla: So how can you create `file("auto");`?
I would rather have an `AUTO_FILENAME = -1` "constant" instead of `"auto"`, which could be handled here (note the type difference, you can still set `"-1"` as the filename, only `-1` as an integer would be handled differently).

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
mitzkia: itHere I wanted some consistency in function names. My plan was the following:
- every function which returns with some object will started with "create_" prefix (it was already ok)
- other functions should stars with an other prefix (to avoid confusion: one is return with something other is not). Started thinking with "global_options()", and I concluded that the clearer prefix is "set_"

tests/python_functional/src/syslog_ng_config/statements/sources/file_source.py
mitzkia: Thanks for the note, and an idea :) .
For negative testcases it will be good to support -1 (as an integer) also. So I would not reserve it as a special case. It is less chance to choose "auto" as a filename. So I would keep this change ("auto").

tests/python_functional/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -64,7 +64,7 @@ def __syntax_check(self):
         result = self.__syntax_only()
         if result["exit_code"] != 0:
             logger.error(result["stderr"])
-            raise Exception("syslog-ng can not started")
+            raise Exception("syslog-ng can not started due to config syntax error")
```
alltilla: Grammar mistake: I would use "cannot be started" or "cannot start".

alltilla:

kira-syslogng:
Build FAILURE

tests/python_functional/src/syslog_ng_config/statements/sources/file_source.py
```diff
@@ -28,16 +28,18 @@
 
 
 class FileSource(SourceDriver):
-    def __init__(self, file_name, **options):
+    def __init__(self, file_name=None, **options):
```
furiel: `config.create_file_source` also defaults `file_name` to `None`. This might be omitted.

tests/python_functional/src/syslog_ng_config/statements/sources/source_writer.py
```diff
@@ -25,15 +25,15 @@
 
 
 class SourceWriter(object):
-    def __init__(self, IOClass):
-        self.__IOClass = IOClass
+    def __init__(self, driver_io_cls):
+        self.__driver_io_cls = driver_io_cls
```
furiel: Optional, probably matter of taste: but I think the original was better. Camelcase hints users that it is a class type and not an instance.

tests/python_functional/functional_tests/destination_drivers/file_destination/test_fd_positional_option.py
```diff
@@ -0,0 +1,80 @@
...
+
+def test_fd_positional_option_custom_dir(config, syslog_ng, generated_bsd_messages):
+    custom_dir = Path(tc_parameters.WORKING_DIR, "custom_dir")
+    custom_dir.mkdir()
+    file_source = config.create_file_source(file_name="input.log")
+    file_destination = config.create_file_destination(file_name=str(custom_dir))
+    config.create_logpath(statements=[file_source, file_destination])
+
+    input_messages, expected_messages = generated_bsd_messages
+    file_source.write_log(input_messages[1], counter=1)
```
furiel: Counter is not needed. input_messages[0] should be used.

tests/python_functional/functional_tests/source_drivers/file_source/test_fs_positional_option.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
+#############################################################################
+# Copyright (c) 2015-2018 Balabit
```
furiel: 2019

kira-syslogng:
Build SUCCESS

mitzkia:
@furiel Thank you very much for your review notes. I will try to handle them in a following days. Maybe in some cases we need to discuss notes in IRL.

tests/python_functional/src/syslog_ng_config/statements/destinations/file_destination.py
```diff
@@ -28,13 +28,18 @@
...
+        if file_name:
+            self.positional_option = Path(tc_parameters.WORKING_DIR, file_name)
+        super(FileDestination, self).__init__(FileIO)
 
     def get_path(self):
-        return self.path
+        return self.positional_option
+
+    def set_path(self, pathname):
+        self.positional_option = Path(tc_parameters.WORKING_DIR, pathname)
```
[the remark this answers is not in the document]
mitzkia: Thanks for the note. In the previous version if I wrote _any_ new driver (also in case of any simple driver: example_message_generator) I should know about the exists of self.positional_options variable, and I should create it even if my new driver is not using it at all. I would not like to do this. In this case drivers are using the minimal set of variables. 
Unfortunately 'aliasing' (self.path = self.positional_option) will not work here, because in case of `set_path()` we should overwrite it (`self.path = Path(...)`) and after that self.positional_option will not use the new value (which we will need in Render() class ).

tests/python_functional/src/syslog_ng_config/renderer.py
```diff
@@ -86,7 +89,10 @@ def __render_statement_groups(self):
                 self.__syslog_ng_config_content += "    {} (\n".format(statement.driver_name)
 
                 # driver options
-                self.__render_positional_options(statement.positional_options)
+                if hasattr(statement, "positional_options"):
```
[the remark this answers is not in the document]
mitzkia: Thanks for the note. Related to `self.positional_options` I have an answer below.

tests/python_functional/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -98,15 +98,15 @@ def start(self, config):
         if self.__process:
             raise Exception("syslog-ng has been already started")
 
-        config.write_content(self.__instance_paths.get_config_path())
+        config.write_config(self.__instance_paths.get_config_path())
```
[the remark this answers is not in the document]
mitzkia: Thanks for the note. I have fixed this part.

tests/python_functional/src/syslog_ng_config/statements/destinations/destination_reader.py
```diff
@@ -31,17 +31,24 @@ class DestinationReader(object):
...
+
+    def construct_message_reader(self):
+        self.driver_io.wait_for_creation()
+        self.__message_reader = MessageReader(
+            self.driver_io.read, self.__driver_io_parser_cls(),
+        )
+
+    def read_logs(self, driver_io_parameter, counter):
+        self.construct_driver_io(driver_io_parameter)
+        self.construct_message_reader()
```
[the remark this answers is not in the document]
mitzkia: Thanks for the note. I will check this part.

MrAnno:
> Maybe this will help us later to move forward from this state to a second part ...

My eye just caught the word "maybe". I don't see the big picture here, so I might be mistaken, but if you're unsure about the right abstractions and APIs or you just couldn't agree with each other on these topics, then I think the best would be to write **many more** real-world tests. Network source or destination support would be really useful to expand our test collection.

Writing new or rewriting internal tests related to network sources/destinations in addition to the existing `file()` tests could settle these architectural questions more quickly in my opinion, because actual tests could lead the abstractions in the right direction. We should talk about exact and specific things (for example, "I want to rewrite this specific internal test in Light, so I need X and Y to do this elegantly). This way, you wouldn't have to think so much ahead, and you would have more specific requirements.
Building the library based on one module and really few exact use-cases (tests) is difficult and does not seem right.

MrAnno:
As I see, you're doing something similar that I've suggested, because this PR contains 13 new test cases.

I think this is a good way of moving forward, I'd event consider renaming the PR to reflect your intention:
every refactor or modification in the test framework is because you want to add a new test or make an existing test better.

mitzkia:
Thanks @MrAnno for checking this PR. I have totally agreed with you and my plan was the same.
With this PR I would like to stabilize and make more readable and logical current driver behaviors.
Because of this I have added refactors and created new (missing) testcases. 
My plans for the future (after this PR merged):
- Add (in small steps) new drivers specially: network based source and destination drivers (TCP, TCP6, UDP, syslog, unix-stream, unix-dgram,...) of course with new valuable functional testcases
- Of course new testcases are always welcome for the existing drivers: file, parsers, filters
- When we will have enough different drivers I would like to check again SyslogNgConfig related codebase to find similar logic to do a new refactor, this was that when I wrote about: "Maybe this will help us later to move forward from this state to a second part"

mitzkia:
I have added some review fixes, merge conflict fixes. 
But still not finished review fixes. I am currently working on this.

kira-syslogng:
Build SUCCESS

mitzkia:
I am still working on this PR to resolve @furiel's review comments. We are agreed in some review points IRL. I need some time to finish this.

kira-syslogng:
Build SUCCESS

MrAnno:
@mitzkia This is growing huge.

I may be wrong, but this PR looks more than a refactor.
- The 702bb397..89964592 range seems like an actual refactor. Can we merge it?
- 6db78fbd is a big one, it deserves a separate PR in my opinion.
- The 7bdd484f..cf43f665 range looks like a series of new features, new tests.

kira-syslogng:
Build SUCCESS

mitzkia:
@MrAnno thanks for your comment.
After I think over your suggestion, if it helps I can split this huge PR into 3 "smaller" parts as you suggested.

mitzkia:
For now I am closing this PR, soon I will open new smaller ones, as it was discussed earlier.

PR #2806, opened 2019-06-26
cmake, autotools: -Wundef to enable-extra-warning 

furiel:
This PR is based on https://github.com/balabit/syslog-ng/pull/2678, adding a few changes.

Please check the individual commits. One inconvenience comes from the difference from cmake and autotools: `AC_CHECK_FUNCS` only define a compiler switch if the symbol is found. However `#cmakedefine01` defines it unconditionally to 0 or 1. The resolution is to use `#cmakedefine` instead of `#cmakedefine01`. Inside the code, we need to use `#if(n)def` instead if `#if` in those cases.


kira-syslogng:
Build FAILURE

folti:

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

furiel:
Thanks @bazsi for the info! I eliminated the dlsym related code.

kira-syslogng:
Build SUCCESS

configure.ac
```diff
@@ -1853,7 +1853,7 @@ AC_DEFINE_UNQUOTED(ENABLE_SYSTEMD, `enable_value $enable_systemd`, [Enable syste
 AC_DEFINE_UNQUOTED(ENABLE_KAFKA, `enable_value $enable_kafka`, [Enable kafka support])
 AC_DEFINE_UNQUOTED(SYSTEMD_JOURNAL_MODE, `journald_mode`, [Systemd-journal support mode])
 AC_DEFINE_UNQUOTED(HAVE_INOTIFY, `enable_value $ac_cv_func_inotify_init`, [Have inotify])
-AC_DEFINE_UNQUOTED(HAVE_GETRANDOM, `enable_value $ac_cv_func_getrandom`, [Have getrandom])
+#AC_DEFINE_UNQUOTED(HAVE_GETRANDOM, `enable_value $ac_cv_func_getrandom`, [Have getrandom])
```
alltilla: Please remove the whole line if it is not needed. Also could you kindly elaborate more, why do we have to remove this?

MrAnno: It is used by the `threaded-random-generator` example plugin.

furiel: This instance of `AC_DEFINE_UNQUOTED` must be removed: getrandom is detected using AC_CHECK_FUNCS too. This line otherwise messes up the already set compiler switch. I will move its removal into a separate commit.

furiel: I could not find a good way to separate it, so I just squashed and added a note in the commit message.

MrAnno: Thanks! 👍

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2807, opened 2019-06-26
Relax commit check

MrAnno:
In my opinion, the following subjects should be accepted:

```
  Revert "cfg-parser: add aliases for yesno"
  geoip2: extend msg with ip
  Light: Remove unused Light dependencies
  autotools,cmake: add -Wundef
  $(grep): fixed possible segfault on parse errors
  elastic-v2: fixed debian packaging
  network(): added grammar skeleton and plugin registration
  version.sh: unset CDPATH
  dbld/images/centos{6,7}: add Python/pip support
```

### DIFF
--- a/tests/commits/check.sh
+++ b/tests/commits/check.sh
@@ -40,7 +40,7 @@ git log --no-merges --pretty="%H" $commit_range -- | (
 
     subject=$(git show -s --format='%s' $commit)
 
-    if ! echo "$subject" | egrep -q '^[a-z/_-]+:'; then
+    if ! echo "$subject" | egrep -q '^Revert |^[a-zA-Z0-9/(){}$,._-]+:'; then
       commit_has_valid_subject=0
     fi
 


kira-syslogng:
Build SUCCESS

tests/commits/check.sh
```diff
@@ -37,9 +37,14 @@ git log --no-merges --pretty="%H" $commit_range -- | (
   while read commit; do
     commit_has_valid_subject=1
     commit_has_signed_off_by=1
+    revert_commit=0
 
     subject=$(git show -s --format='%s' $commit)
 
+    if echo "$subject" | egrep -q '^Revert '; then
+      revert_commit=1
+    fi
```
Kokan: Do we really want the "revert" commits to be excluded ?
The revert commit does not really have any details that distinguish it from other commits (message can be altered any point).
By including the `Signed-off` in the commit message, does not really mean anything at all. (git at least does not specify, and I could not find any statement in the syslog-ng repository either, except that you must include in every commit)

Despite the above I like `Signed-Off` as you can easily give credit with it to other people.

tests/commits/check.sh
```diff
@@ -40,7 +40,7 @@ git log --no-merges --pretty="%H" $commit_range -- | (
 
     subject=$(git show -s --format='%s' $commit)
 
-    if ! echo "$subject" | egrep -q '^[a-z/_-]+:'; then
+    if ! echo "$subject" | egrep -q '^[a-zA-Z0-9/(){}$,._-]+:'; then
```
Kokan: I was kinda happy with the strict version, especially not to include upper case letters :) I see the value in the rest modification. As this is a very opinionated topic, it is not a change request.

tests/commits/check.sh
MrAnno: We have 22 revert commits in this repository, only 3 of them contain signoff lines. We can start using signoffs there too.

kira-syslogng:
Build SUCCESS

tests/commits/check.sh
Kokan: thanks.

furiel:
I have a commit like `autotools, cmake: add -Wundef ` that fails now. What is the guidance here? Should I split to two commits and have `autotools: add -Wundef` and `cmake: add -Wundef`? Or should I make up a module name, like `build: add -Wundef`?

Kokan:
@furiel according to this PR (see the examples):  `autotools,cmake: add -Wundef` (removing the space)
Personally I would prefer the space removal in this case, but sure if we find some case that fails it could be additionally included. (let this be an incremental process to improve the check script)

mitzkia:
* Sorry, but the "Light" starts with capital "L", I would not like change it to: "l"
* My other question is what would be the main rules for the commit subject. Could you please give some (different) examples?
* What is the main goal with this rule?
* I understand that it would be easier to read if the subjects are consequent.

MrAnno:
I personally prefer lowercase subjects, but I've added uppercase letters to the commit-checker because I don't think we should be that strict.

mitzkia:
@MrAnno Thanks for the clarification. I am approving this.

PR #2808, opened 2019-06-27
radix: fix grouping in pcre

### LINKED ISSUE #2793: grouping doesn't work in pcre
I stumbled on a bug when playing with regular expressions.
This has probably nothing to do with patterndb, but it was easier for me to test this way.
The issue: `(re)` grouping doesn't seem to work, but shy-grouping `(?:re)` does:

```xml
<?xml version='1.0' encoding='UTF-8'?>
  <patterndb version="4" pub_date="2019-06-20 14:43:37">
     <ruleset id="pcre grouping issue" name="grouping woes">
        <rules>
           <rule id="yay! shy grouping works too!">
              <patterns>
                 <pattern>@PCRE::(?:shy|group)@</pattern>
              </patterns>
              <examples>
                 <example>
                    <test_message program="d">group</test_message>
                 </example>
              </examples>
           </rule>
           <rule id="without grouping it works!">
              <patterns>
                 <pattern>@PCRE::foo|bar@</pattern>
              </patterns>
              <examples>
                 <example>
                    <test_message program="a">foo</test_message>
                 </example>
              </examples>
           </rule>
           <rule id="with escaped parens grouping still doesn't work">
              <patterns>
                 <pattern>@PCRE::\(brown|fox\)@</pattern>
              </patterns>
              <examples>
                 <example>
                    <test_message program="c">fox</test_message>
                 </example>
              </examples>
           </rule>
           <rule id="with grouping it doesn't work">
              <patterns>
                 <pattern>@PCRE::(baz|blah)@</pattern>
              </patterns>
              <examples>
                 <example>
                    <test_message program="b">baz</test_message>
                 </example>
              </examples>
           </rule>
        </rules>
     </ruleset>
  </patterndb>
```

```
$ pdbtool test /tmp/ts.xml  -v
Module loaded and initialized successfully; module='syslogformat'
Module loaded and initialized successfully; module='basicfuncs'
Testing message: program='b' message='baz'
 Wrong match name='.classifier.rule_id', value='', expected='with grouping it doesn't work'
Testing message: program='c' message='fox'
 Wrong match name='.classifier.rule_id', value='', expected='with escaped parens grouping still doesn't work'
Testing message: program='a' message='foo'
 Match name='.classifier.rule_id', value='without grouping it works!', expected='without grouping it works!'
Testing message: program='d' message='group'
 Match name='.classifier.rule_id', value='yay! shy grouping works too!', expected='yay! shy grouping works too!'
```

alltilla:
`pcre_exec()` did not have enough place to store the offsets in the offset_vector (`matches` variable). I copied the API usage from `logmatcher.c`, which seems to work.

https://www.pcre.org/original/doc/html/pcre_exec.html requires the offset vector's size to be a multiple of 3. `ovecsize     Number of elements in the vector (a multiple of 3)`

Note for the reviewers:
 * radix.c is abyssal for me, and I could not really prove to myself, that adding the `matches` array to `RParserPCREState` will not cause a race condition.

Fixes #2793

### DIFF
--- a/modules/dbparser/radix.c
+++ b/modules/dbparser/radix.c
@@ -138,11 +138,35 @@ r_parser_pcre(gchar *str, gint *len, const gchar *param, gpointer state, RParser
 {
   RParserPCREState *self = (RParserPCREState *) state;
   gint rc;
-  gint matches[2];
+  gint num_matches;
 
-  rc = pcre_exec(self->re, self->extra, str, strlen(str), 0, 0, matches, 2);
-  if (rc <= 0)
-    return FALSE;
+  if (pcre_fullinfo(self->re, self->extra, PCRE_INFO_CAPTURECOUNT, &num_matches) < 0)
+    g_assert_not_reached();
+  if (num_matches > RE_MAX_MATCHES)
+    num_matches = RE_MAX_MATCHES;
+
+  gsize matches_size = 3 * (num_matches + 1);
+  gint *matches = g_alloca(matches_size * sizeof(gint));
+
+  rc = pcre_exec(self->re, self->extra, str, strlen(str), 0, 0, matches, matches_size);
+
+  if (rc == PCRE_ERROR_NOMATCH)
+    {
+      return FALSE;
+    }
+
+  if (rc < 0)
+    {
+      msg_error("Error while matching regexp", evt_tag_int("error_code", rc));
+      return FALSE;
+    }
+
+  if (rc == 0)
+    {
+      msg_error("Error while storing matching substrings");
+      return FALSE;
+    }
+
   *len = matches[1] - matches[0];
   return TRUE;
 }
--- a/modules/dbparser/tests/test_radix.c
+++ b/modules/dbparser/tests/test_radix.c
@@ -1004,6 +1004,16 @@ ParameterizedTestParameters(dbparser, test_radix_search_matches)
       .key = "jjjj abcabcd foobar",
       .expected_pattern = {"regexp", "abcabc", NULL},
     },
+    {
+      .node_to_insert = {"@PCRE:regexp:(foo|bar)@", NULL},
+      .key = "foo",
+      .expected_pattern = {"regexp", "foo", NULL},
+    },
+    {
+      .node_to_insert = {"@PCRE:regexp:(?:foo|bar)@", NULL},
+      .key = "foo",
+      .expected_pattern = {"regexp", "foo", NULL},
+    },
     /* test_nlstring_matches */
     {
       .node_to_insert = {"@NLSTRING:nlstring@\n", NULL},


kira-syslogng:
Build SUCCESS

alltilla:
This might slow down `patterndb` a little, might need to rethink the fix.
(maybe it is enough to set `matches_size` to `3 * (1 + 1)`)

modules/dbparser/radix.c
```diff
@@ -138,11 +138,37 @@ r_parser_pcre(gchar *str, gint *len, const gchar *param, gpointer state, RParser
...
+  if (pcre_fullinfo(self->re, self->extra, PCRE_INFO_CAPTURECOUNT, &num_matches) < 0)
+    g_assert_not_reached();
+
+  if (num_matches > RE_MAX_MATCHES)
+    num_matches = RE_MAX_MATCHES;
+
+  gsize matches_size = 3 * (num_matches + 1);
+  gint *matches = g_alloca(matches_size * sizeof(gint));
+
+  rc = pcre_exec(self->re, self->extra, str, strlen(str), 0, 0, matches, matches_size);
```
Kokan: As far as I can see the matches are not *really* needed, but the length of matched string. Is there a way to only obtain that much ?
If not, the allocation part could be at least moved to the `r_parser_pcre_compile_state`.

Kokan: As far as I can see there is no way to only obtained the length of matched string (peeked inside the source).
Also if we do not provide - enough - space for matches, internally it is going to call for `pcre_malloc` (that user could override, but still), therefore I think it is better to allocate in the `r_parser_pcre_compile_state` as suggested before.

modules/dbparser/radix.c
```diff
@@ -138,11 +138,37 @@ r_parser_pcre(gchar *str, gint *len, const gchar *param, gpointer state, RParser
...
+  if (rc < 0)
+    {
+      switch (rc)
+        {
+        case PCRE_ERROR_NOMATCH:
+          break;
+
```
Kokan: Could you please also additionally clean up this error handling ? (kinda hard to read)

modules/dbparser/radix.c
gaborznagy: I can agree with @Kokan that calculating the matches_size can be done in compile() step, next to `pcre_compile` and `pcre_study`.
It does not seem to be a big effort either.

However, then it should be done in `log_matcher_pcre` too, right? I wouldn't like to increase the scope of this PR, but it would only makes sense.

modules/dbparser/radix.c
alltilla: Sure, done! :)

modules/dbparser/radix.c
alltilla: I moved the alloc to compile!
I will open another PR for the `log_matcher_pcre` :)

kira-syslogng:
Build FAILURE

alltilla:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/dbparser/radix.c
```diff
@@ -131,22 +131,52 @@ typedef struct _RParserPCREState
...
+_alloc_matches(RParserPCREState *self)
+{
+  gint num_matches;
+  if (pcre_fullinfo(self->re, self->extra, PCRE_INFO_CAPTURECOUNT, &num_matches) < 0)
+    g_assert_not_reached();
+  if (num_matches > RE_MAX_MATCHES)
+    num_matches = RE_MAX_MATCHES;
+
+  self->matches_size = 3 * (num_matches + 1);
+  self->matches = g_malloc(self->matches_size * sizeof(gint));
```
furiel: Me seems similar to https://github.com/balabit/syslog-ng/pull/2827#discussion_r303752604. Can `matches` be part of `RParserPCREState`?

bazsi: The radix evaluation runs (or should be running in parallel), IIRC there was a lock just because of reload but eventually I would like to get rid off that. Querying the number of matches of a compiled regexp should be a very simple operation, here's the code from pcre_fullinfo():

```
  case PCRE_INFO_CAPTURECOUNT:
  *((int *)where) = re->top_bracket;
  break;
```

So apart from the function call overhead, this is basically nothing. Querying only the count should be fine, but the matches array should be allocated on the spot (hopefully with alloca instead of malloc), as the first one is much faster.

alltilla: Thanks, I moved the allocation to be done per message, and it is done with `alloca`.

kira-syslogng:
Build FAILURE

alltilla:

kira-syslogng:
Build FAILURE

alltilla:

kira-syslogng:
Build SUCCESS

PR #2810, opened 2019-06-28
update according to clang, gcc diagnostic differences, and fix newly found issues

Kokan:
The main focus was on `cmake+clang`, sorry `autotools`.
Changes that were to replace/remove a warning was also changed in the `autotools`.
But a few option was removed in case of `clang` as `clang` does not define them, and those were not updated in `autotools` (in case of `autotools` those warnings must be turned on via `--enable-extra-warnings`).

Additionally I also turned on the extra warnings for `cmake+clang` combination.

Note: it seems clang is able to detect issues even if they are hidden via macro usage, while gcc is unable to see such errors (anything written in macros gcc just fine with it).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -337,8 +337,8 @@ if (ENABLE_EXTRA_WARNINGS)
     -Wold-style-declaration
     -Wstrict-prototypes
     -Wswitch-default
+    -Wimplicit-int
     -Wall
-    -Wmissing-parameter-type
     -Wuninitialized
     -Wunused-but-set-parameter
     -Wcast-align
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -353,6 +353,17 @@ if (ENABLE_EXTRA_WARNINGS)
     -Wsuggest-attribute=noreturn
     -Wold-style-definition
     -Wundef)
+
+if ("${CMAKE_C_COMPILER_ID}" MATCHES "Clang")
+  set(EXTRA_WARNINGS
+    ${EXTRA_WARNINGS}
+    )
+else()
+  set(EXTRA_WARNINGS
+    ${EXTRA_WARNINGS}
+    )
+endif()
+
 endif()
 
 add_compile_options(${IMPORTANT_WARNINGS} ${ACCEPTABLE_WARNINGS} ${EXTRA_WARNINGS})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -339,7 +339,6 @@ if (ENABLE_EXTRA_WARNINGS)
     -Wimplicit-int
     -Wall
     -Wuninitialized
-    -Wcast-align
     -Wdeprecated
     -Wdeprecated-declarations
     -Woverflow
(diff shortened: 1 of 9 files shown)


kira-syslogng:
Build SUCCESS

lib/tests/test_type_hints.c
```diff
@@ -214,14 +214,21 @@ ParameterizedTestParameters(type_hints, test_double_cast)
                              sizeof(string_value_pairs) / sizeof(string_value_pairs[0]));
 }
 
+static gboolean
+cr_assert_gdouble_eq(gdouble a, gdouble b)
+{
+  return fabs(a-b) < G_MINDOUBLE;
```
alltilla: Shouldn't we use `cr_assert` instead of `return`?

Kokan: omg ?!? well at least it does not fails...
thanks, sure we do.

kira-syslogng:
Build FAILURE
 
Build FAILURE

kira-syslogng:
Build FAILURE

Kokan:

kira-syslogng:
Build SUCCESS

Makefile.am
```diff
@@ -74,7 +74,6 @@ AM_CFLAGS += \
 	-Wall \
 	-Wuninitialized \
 	-Wunused-but-set-parameter \
-	-Wcast-align \
```
furiel: Do we want to drop this entirely or plan to fix this later? In the latter case, you might open a jira issue for it.

lib/tests/test_type_hints.c
```diff
@@ -214,14 +214,31 @@ ParameterizedTestParameters(type_hints, test_double_cast)
...
+  const gint is_b_inf = isinf((long double)a);
+
+  cr_assert_eq(is_a_inf, is_b_inf);
+  if (is_a_inf && is_b_inf)
+    {
+      cr_assert(TRUE);
+      return;
+    }
+
+  cr_assert(fabs(a-b) < G_MINDOUBLE);
```
furiel: Me feels G_MINDOUBLE too small, but have no better idea. Nonetheless it is equivalent change this way. Later we might want to pass this as a parameter if we run into trouble due to precision.

kira-syslogng:
Build FAILURE

CMakeLists.txt
```diff
@@ -350,9 +347,21 @@ if (ENABLE_EXTRA_WARNINGS)
     -Wpointer-arith
     -Wpointer-sign
     -Wmissing-format-attribute
-    -Wsuggest-attribute=noreturn
     -Wold-style-definition
     -Wundef)
+
+if ("${CMAKE_C_COMPILER_ID}" MATCHES "Clang")
```
alltilla: We might want to add the same to `configure.ac` with:
```
if test "x$ac_compiler_gnu" = "xyes"; then
...
```

Kokan: clang is kinda a gnu compiler, at least it pretends to be one. this condition won't help here.

alltilla:

kira-syslogng:
Build SUCCESS

PR #2816, opened 2019-07-02
redis: handle any number of parameters at `command`

### LINKED ISSUE #2813: redis `command` should accept arbitrary length of arguments not limited to four.
# syslog-ng
## Version of syslog-ng

```
syslog-ng 3 (3.22.1)
Config version: 3.22
Installer-Version: 3.22.1
Revision: 3.22.1-2
Compile-Date: Jun 21 2019 12:50:04
Module-Directory: /usr/lib/syslog-ng/3.22
Module-Path: /usr/lib/syslog-ng/3.22
Include-Path: /usr/share/syslog-ng/include
Available-Modules: hook-commands,csvparser,appmodel,pseudofile,redis,json-plugin,confgen,afsocket,kvformat,afuser,mod-python,date,disk-buffer,xml,cef,basicfuncs,tags-parser,system-source,cryptofuncs,dbparser,afprog,linux-kmsg-format,affile,syslogformat,sdjournal
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: on
```
# Issue

We want to add data to a redis stream with `XADD`, this redis command accepts an arbitrary length of fields, but the redis `command` only accepts four arguments which it too little for `XADD` and maybe even a few others. 

I also tried `command` with shoving all fields into the last argument which also does not work. 

## Configuration

```
destination d_redis {
    # I would love to use this but `command` only allows arguments
    # which for XADD is to little as we supply two key, value pairs `message_id` and `data`
    redis(
        host("redis-host")
        port(13451)
        auth("...")
        command("XADD", "instanceLogged", "*", "message_id $MESSAGE_UUID") # Here I want to add more `fields`
    );
};

```

alltilla:
There are some redis commands, which needs more that 3 fields after the command.
This PR allows the user to set any number of parameters to redis's `command()` option.
The first must be a string, the remaining ones can be templates as well.

Example usage:
```
destination d_redis {
  redis(
      host("127.0.0.1")
      port(6379)
      command("XADD", "instanceLogged", "*", "message_id", "$HOST")
  );
};
```
Fixes #2813

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -426,6 +426,7 @@ DNSCacheOptions *last_dns_cache_options;
 %type   <ptr> dest_plugin
 
 %type   <ptr> template_content
+%type   <ptr> template_content_list
 
 %type   <ptr> filter_content
 
@@ -894,6 +895,11 @@ template_content_inner
 template_content
         : { last_template = log_template_new(configuration, NULL); } template_content_inner	{ $$ = last_template; }
         ;
+
+template_content_list
+	: template_content template_content_list { $$ = g_list_prepend($2, $1); }
+	| { $$ = NULL; }
+	;
 
 /* END_RULES */
 
--- a/modules/redis/redis.c
+++ b/modules/redis/redis.c
@@ -33,6 +33,8 @@
 #include "driver.h"
 #include "plugin-types.h"
 #include "logthrdest/logthrdestdrv.h"
+#include "scratch-buffers.h"
+#include "utf8utils.h"
 
 typedef struct
 {
@@ -45,14 +47,13 @@ typedef struct
   LogTemplateOptions template_options;
 
   GString *command;
-  LogTemplate *key;
-  GString *key_str;
-  LogTemplate *param1;
-  GString *param1_str;
-  LogTemplate *param2;
-  GString *param2_str;
+  GList *arguments;
 
+  /* Worker thread */
   redisContext *c;
+  gint argc;
+  gchar **argv;
+  size_t *argvlen;
 } RedisDriver;
 
 /*
@@ -85,23 +86,22 @@ redis_dd_set_auth(LogDriver *d, const gchar *auth)
   self->auth = g_strdup(auth);
 }
 
+static void
+_template_unref(gpointer data)
+{
+  LogTemplate *template = (LogTemplate *)data;
+  log_template_unref(template);
+}
+
 void
 redis_dd_set_command_ref(LogDriver *d, const gchar *command,
-                         LogTemplate *key,
-                         LogTemplate *param1, LogTemplate *param2)
+                         GList *arguments)
 {
   RedisDriver *self = (RedisDriver *)d;
 
   g_string_assign(self->command, command);
-
-  log_template_unref(self->key);
-  self->key = key;
-
-  log_template_unref(self->param1);
-  self->param1 = param1;
-
-  log_template_unref(self->param2);
-  self->param2 = param2;
+  g_list_free_full(self->arguments, _template_unref);
+  self->arguments = arguments;
 }
 
 LogTemplateOptions *
@@ -221,6 +221,34 @@ redis_dd_disconnect(LogThreadedDestDriver *s)
   self->c = NULL;
 }
 
+static void
+_fill_argv_from_template_list(RedisDriver *self, LogMessage *msg)
+{
+  for (gint i = 1; i < self->argc; i++)
+    {
+      GString *buffer = scratch_buffers_alloc();
+      log_template_format(g_list_nth_data(self->arguments, i-1), msg, &self->template_options, LTZ_SEND,
+                          self->super.worker.instance.seq_num, NULL, buffer);
+      self->argvlen[i] = buffer->len;
+      self->argv[i] = buffer->str;
+    }
+}
+
+static GString *
+_argv_to_string(RedisDriver *self)
+{
+  GString *full_command = scratch_buffers_alloc();
+
+  full_command = g_string_append(full_command, self->argv[0]);
+  for (gint i = 1; i < self->argc; i++)
+    {
+      g_string_append(full_command, " \"");
+      append_unsafe_utf8_as_escaped_text(full_command, self->argv[i], self->argvlen[i], "\"");
+      g_string_append(full_command, "\"");
+    }
+  return full_command;
+}
+
 /*
  * Worker thread
  */
@@ -229,10 +257,6 @@ static LogThreadedResult
 redis_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
 {
   RedisDriver *self = (RedisDriver *)s;
-  redisReply *reply;
-  const char *argv[5];
-  size_t argvlen[5];
-  int argc = 2;
 
   if (!redis_dd_connect(self))
     return LTR_NOT_CONNECTED;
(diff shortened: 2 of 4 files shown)


kira-syslogng:
Build SUCCESS

modules/redis/redis.c
```diff
@@ -246,57 +252,40 @@ redis_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
...
-      argvlen[3] = self->param2_str->len;
-      argc++;
-    }
+  gint index = 1;
+  gpointer user_data[] = { self, msg, argv, argvlen, &index };
+  g_list_foreach(self->fields, _fill_argv, user_data);
 
   reply = redisCommandArgv(self->c, argc, argv, argvlen);
 
+  GString *full_command = g_string_new(argv[0]);
```
MrAnno: This is a per-message allocation but the string is only used for debug/error messages.
I'd use a scratch-buffer or a truncated buffer on the stack.

kira-syslogng:
Build FAILURE

lib/cfg-grammar.y
```diff
@@ -895,6 +897,15 @@ template_content
         : { last_template = log_template_new(configuration, NULL); } template_content_inner	{ $$ = last_template; }
         ;
 
+template_content_list
+        : template_content_list_build                     { $$ = $1; }
+        ;
+
+template_content_list_build
```
MrAnno: I think you can omit the `template_content_list_build` non-terminal symbol and just use `template_content_list`:

```
template_content_list : template_content template_content_list { $$ = prepend(); } | { $$ = NULL; }
```

modules/redis/redis.c
```diff
@@ -221,6 +207,26 @@ redis_dd_disconnect(LogThreadedDestDriver *s)
...
 }
 
+static void
+_fill_argv(gpointer data, gpointer user_data)
+{
+  RedisDriver *self = ((gpointer *)user_data)[0];
+  LogMessage *msg = ((gpointer *)user_data)[1];
+  const char **argv = ((gpointer *)user_data)[2];
+  size_t *argvlen = ((gpointer *)user_data)[3];
+  gint *index = ((gpointer *)user_data)[4];
```
MrAnno: This is a lot of user data. Since you need the index as well (and `g_list_foreach` won't give you the index) , I'd transform this into an ordinary loop.

modules/redis/redis.c
alltilla: Thanks, will do!

alltilla:
There are some filthy things here (const char ** cast...), I am marking this as WIP, and will continue the work next week. Comments are highly appreciated.

modules/redis/redis.c
alltilla: Done

lib/cfg-grammar.y
alltilla: Done thanks!

modules/redis/redis.c
alltilla: Done, thanks!

modules/redis/redis.c
```diff
@@ -246,57 +233,51 @@ redis_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
...
-  argv[1] = self->key_str->str;
-  argvlen[1] = self->key_str->len;
 
-  if (self->param1)
+  for (int i = 0; i < g_list_length(self->fields); i++)
     {
-      argv[2] = self->param1_str->str;
-      argvlen[2] = self->param1_str->len;
-      argc++;
+      GString *str = g_string_sized_new(1024);
```
MrAnno: I think this is another good scratch-buffer candidate.

modules/redis/redis.c
```diff
@@ -246,57 +233,51 @@ redis_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
...
+  argv = g_malloc(argc * sizeof(char *));
+  argvlen = g_malloc(argc * sizeof(size_t));
 
   argv[0] = self->command->str;
   argvlen[0] = self->command->len;
-  argv[1] = self->key_str->str;
-  argvlen[1] = self->key_str->len;
 
-  if (self->param1)
+  for (int i = 0; i < g_list_length(self->fields); i++)
```
MrAnno: Optional: Please consider extracting this loop (maybe with the argv{len} allocation?) into a separate function.
For example: `create_command()`, `format_command()`, etc.

modules/redis/redis.c
```diff
@@ -246,57 +233,51 @@ redis_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
...
-            evt_tag_str("param1", self->param1_str->str),
-            evt_tag_str("param2", self->param2_str->str));
+            evt_tag_str("command", full_command->str));
+
   freeReplyObject(reply);
+  for (int i = 1; i < argc; i++)
+    g_free(argv[i]);
+  g_free(argv);
+  g_free(argvlen);
+  scratch_buffers_reclaim_marked(marker);
```
MrAnno: This function has early return statements. We should not leak these objects in case of errors.

kira-syslogng:
Build FAILURE

lib/cfg-grammar.y
```diff
@@ -895,6 +896,8 @@ template_content
         : { last_template = log_template_new(configuration, NULL); } template_content_inner	{ $$ = last_template; }
         ;
 
+template_content_list : template_content template_content_list { $$ = g_list_prepend($2, $1); } | { $$ = NULL; }
+
```
MrAnno: Sorry, it was just pseudocode, I forgot to add `;`:

```
template_content_list
	:  template_content template_content_list { $$ = g_list_prepend($2, $1); }
	| { $$ = NULL; }
	;
```

alltilla:
Sorry I was in a hurry. I will fix these on Tuesday. Thanks for the comments!

kira-syslogng:
Build FAILURE

lib/cfg-grammar.y
alltilla: Thanks, fixed! :)

modules/redis/redis.c
alltilla: Nice catch, fixed :)

modules/redis/redis.c
alltilla: Done, I also moved the argv alloc to the worker thread init, and I am reusing it every time.

modules/redis/redis.c
alltilla: Thanks, done :)

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

alltilla:

kira-syslogng:
Build SUCCESS

modules/redis/redis.c
```diff
@@ -221,6 +212,40 @@ redis_dd_disconnect(LogThreadedDestDriver *s)
...
+}
+
+static GString *
+_argv_to_string(RedisDriver *self, ScratchBuffersMarker *marker)
+{
+  GString *full_command = scratch_buffers_alloc_and_mark(marker);
+
+  full_command = g_string_append(full_command, self->argv[0]);
+  for (gint i = 1; i < self->argc; i++)
+    g_string_append_printf(full_command, " \"%s\"", self->argv[i]);
```
MrAnno: Absolutely optional:

`argv[i]` is not escaped between `\"%s\"`.

You can use `lib/utf8utils` to escape them if you want to. An example from the JSON module:
```
static inline void
tf_json_append_escaped(GString *dest, const gchar *str, gssize str_len)
{
  append_unsafe_utf8_as_escaped_text(dest, str, str_len, "\"");
}
```

alltilla: Thanks, I fixed it! :)

kira-syslogng:
Build SUCCESS

modules/redis/redis.c
MrAnno: Nice, thanks! :)

modules/redis/redis.c
```diff
@@ -87,21 +88,12 @@ redis_dd_set_auth(LogDriver *d, const gchar *auth)
...
-
-  log_template_unref(self->key);
-  self->key = key;
-
-  log_template_unref(self->param1);
-  self->param1 = param1;
-
-  log_template_unref(self->param2);
-  self->param2 = param2;
+  self->fields = fields;
```
Kokan: You should free `self->fields` before setting it into a new value.

modules/redis/redis.c
```diff
@@ -341,7 +341,7 @@ redis_dd_init(LogPipe *s)
   if (!log_threaded_dest_driver_init_method(s))
     return FALSE;
 
-  if (!self->key && !self->param1)
+  if (!g_list_nth_data(self->fields, 0) && !g_list_nth_data(self->fields, 1))
```
Kokan: [optional] you could use `g_list_length(self->fields) >= 2`

modules/redis/redis.c
```diff
@@ -45,14 +47,13 @@ typedef struct
   LogTemplateOptions template_options;
 
   GString *command;
-  LogTemplate *key;
-  GString *key_str;
-  LogTemplate *param1;
-  GString *param1_str;
-  LogTemplate *param2;
-  GString *param2_str;
+  GList *fields;
```
Kokan: [optional] instead of `fields` I would prefer `arguments`.

modules/redis/redis.c
```diff
@@ -246,58 +272,32 @@ redis_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
...
-    {
-      argv[3] = self->param2_str->str;
-      argvlen[3] = self->param2_str->len;
-      argc++;
-    }
+  redisReply *reply = redisCommandArgv(self->c, self->argc, (const gchar **)self->argv, self->argvlen);
 
-  reply = redisCommandArgv(self->c, argc, argv, argvlen);
+  ScratchBuffersMarker marker;
+  GString *full_command = _argv_to_string(self, &marker);
```
Kokan: [optional] you don't have to inject a marker into this function, simply set the marker outside:

```C
ScratchBufferMarker marker;
scratch_buffers_mark(&marker);
GString *full_command = _argv_to_string(self);
...
_free_argv_templated_part(self);
scratch_buffers_reclaim_marked(marker);
```

modules/redis/redis.c
```diff
@@ -221,6 +213,44 @@ redis_dd_disconnect(LogThreadedDestDriver *s)
...
+{
+  ScratchBuffersMarker marker;
+  GString *buffer = scratch_buffers_alloc_and_mark(&marker);
+
+  for (gint i = 1; i < self->argc; i++)
+    {
+      log_template_format(g_list_nth_data(self->fields, i-1), msg, &self->template_options, LTZ_SEND,
+                          self->super.worker.instance.seq_num, NULL, buffer);
+      self->argvlen[i] = buffer->len;
+      self->argv[i] = g_strdup(buffer->str);
```
Kokan: hmm, this just invalidates the scratch buffer usage for me. does these values need to outlive the current message processing ?

modules/redis/redis.c
alltilla: Thanks fixed!

modules/redis/redis.c
alltilla: The old check was buggy, as it allowed to have no `key`, if `param1` was set, but `key` was used regardless in `redis_worker_insert()`, without checking if it existed:

```                                                                                                                                                                                               
  log_template_format(self->key, msg, &self->template_options, LTZ_SEND,                                                                                                                      
                      self->super.worker.instance.seq_num, NULL, self->key_str); 
```
So, I changed it to check if the list is empty.

modules/redis/redis.c
alltilla: Thanks, good idea, I changed it.

modules/redis/redis.c
alltilla: Thanks, I think I finally understood how scratch_markers work, and I merged the usage of the marker with `_fill_argv_from_template_list()`.

modules/redis/redis.c
alltilla: Thanks, I think I fixed it! :)

kira-syslogng:
Build SUCCESS

PR #2818, opened 2019-07-04
Timezone related rewrite rules

bazsi:
The aim of this branch is to allow manipulating the timezone information with rewrite rules, that is:

- set a timezone to a specific value (e.g. convert to a target timezone)
- fix a timezone if it was improperly parsed (e.g. parsed as CET, but we KNOW that it is EST), takes a timezone value as a template
- guess the timezone value assuming the sender is sending messages in near-real-time, e.g. the guess-timezone functionality but as a rewrite rule.

Examples are available in the individual commit messages.

This branch now also contains a couple of bugfixes and refactors, so that we can abandon our caches upon the receipt of a SIGHUP.


bazsi:
btw, some of the patches could be integrated separately, let me know if you prefer that I split the branch. But I would prefer to do that if those could actually be merged faster than the whole.

kira-syslogng:
Build FAILURE

bazsi:
This should be good to go.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
This branch now also contains a change in the copyright checker script as I couldn't declare non-balabit copyrights easily. Those are added to the beginning of the branch and I now hope that this becomes nice and green.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:
Finally green, again let me know if you want me to split the branch.

lib/timeutils/zonecache.c
```diff
@@ -0,0 +1,68 @@
...
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+#include "timeutils/zonecache.h"
+#include "timeutils/zoneinfo.h"
+#include "timeutils/zonedb.h"
+#include "tls-support.h"
+#include "../cache.h"
```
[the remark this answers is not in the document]
bazsi: we have a name collision between lib/cache.h  and lib/timeutils/cache.h and without the "../" the wrong one is used. That's the reason. I don't like it either, but this was the easiest solution.

lib/timeutils/zonecache.c
```diff
@@ -0,0 +1,68 @@
...
+
+#define tz_cache  __tls_deref(tz_cache)
+
+typedef struct _TimeZoneResolver
+{
+  CacheResolver super;
+} TimeZoneResolver;
+
+gpointer
+time_zone_resolver_resolve(CacheResolver *s, const gchar *tz)
```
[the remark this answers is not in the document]
bazsi: done.

lib/rewrite/rewrite-expr-grammar.ym
```diff
@@ -60,15 +58,13 @@ LogRewrite *last_rewrite;
...
 
-%token KW_CONDITION
 %token KW_SET_TAG
 %token KW_CLEAR_TAG
 %token KW_GROUP_SET
 %token KW_GROUP_UNSET
 %token KW_SET
 %token KW_UNSET
 %token KW_SUBST
-%token KW_VALUE
```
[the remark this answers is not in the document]
bazsi: done.

lib/timeutils/tests/test_unixtime.c
```diff
@@ -25,9 +25,13 @@
 #include "timeutils/wallclocktime.h"
 #include "timeutils/conv.h"
 #include "timeutils/zonecache.h"
+#include "timeutils/format.h"
+
 #include <criterion/criterion.h>
 #include "fake-time.h"
 
+#include <stdio.h>
```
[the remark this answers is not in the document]
bazsi: done

modules/timestamp/rewrite-guess-timezone.c
```diff
@@ -0,0 +1,78 @@
...
+            evt_tag_long("new_gmtoff", msg->timestamps[self->stamp].ut_gmtoff));
+}
+
+static LogPipe *
+_clone(LogPipe *s)
+{
+  RewriteGuessTimeZone *self = (RewriteGuessTimeZone *) s;
+  LogRewrite *cloned;
+
+  cloned = rewrite_guess_time_zone_new(s->cfg);
```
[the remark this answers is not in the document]
bazsi: done

lib/timeutils/zonecache.c
```diff
@@ -0,0 +1,68 @@
...
+
+  return &self->super;
+}
+
+TimeZoneInfo *
+cached_get_time_zone_info(const gchar *tz)
+{
+  if (!tz_cache)
+    tz_cache = cache_new(time_zone_resolver_new());
+  TimeZoneInfo *result = cache_lookup(tz_cache, tz);
```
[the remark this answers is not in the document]
bazsi: hmm... good point. we recently had a PR to invalidate the cached timezone information at reload... This is not exactly the same (the local timezone may change during startup, DST transitions do change, but pretty rarely and years into the future. Anyway I am looking into how this could be done.

bazsi:
Thanks for the comments @Kokan I've pushed fixes for some of your comments I am putting this back to "WIP" to check out the last one.

kira-syslogng:
Build SUCCESS

lib/timeutils/zonecache.c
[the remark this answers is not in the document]
bazsi: With the latest iteration, I've added this change to this branch. It did cause some complications but in the process I've also fixed a couple of related bugs, so I consider this worth it.

It did increase the patch numbers though.

kira-syslogng:
Build FAILURE

Kokan:
@bazsi the `test_zone` unit test fails:

```
==1595==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f4788b6b60a bp 0x7ffd8b8b5600 sp 0x7ffd8b8b55f0 T0)
==1595==The signal is caused by a READ memory access.
==1595==Hint: address points to the zero page.
    #0 0x7f4788b6b609 in cache_clear /home/peterkokai/src/syslog-ng/build/../lib/cache.c:50:14
    #1 0x7f4788c09320 in _validate_timeutils_cache /home/peterkokai/src/syslog-ng/build/../lib/timeutils/cache.c:99:7
    #2 0x7f4788c0934c in cached_localtime /home/peterkokai/src/syslog-ng/build/../lib/timeutils/cache.c:195:3
    #3 0x7f4788c0d9ec in get_local_timezone_ofs /home/peterkokai/src/syslog-ng/build/../lib/timeutils/misc.c:44:3
    #4 0x539c93 in assert_time_zone_offset /home/peterkokai/src/syslog-ng/build/../tests/unit/test_zone.c:117:12
    #5 0x53a493 in zone_test_time_zone_offset_impl /home/peterkokai/src/syslog-ng/build/../tests/unit/test_zone.c:142:5
    #6 0x7f4788a55e43 in criterion_internal_test_main (/usr/lib/libcriterion.so.3+0x8e43)
    #7 0x53a464 in zone_test_time_zone_offset_jmp /home/peterkokai/src/syslog-ng/build/../tests/unit/test_zone.c:131:1
    #8 0x7f4788a55308  (/usr/lib/libcriterion.so.3+0x8308)
    #9 0x7f4788a66b4c  (/usr/lib/libcriterion.so.3+0x19b4c)
    #10 0x7f478812bee2 in __libc_start_main (/usr/lib/libc.so.6+0x26ee2)
    #11 0x41e1dd in _start (/home/peterkokai/src/syslog-ng/build/tests/unit/test_zone+0x41e1dd)
```

The `cache_clear` is NULL.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
Those who only seeks the new configuration options, examples can be found in the following commit messages:
* 59c05abb448568a1a4d1ce408b7328032e3dd139 timestamp: add guess-timezone rewrite operation
* 1292ac2c309487cfb8ded7939ed25e43becefdc2 timestamp: add fix-timezone() rewrite operation

lib/timeutils/zoneinfo.c
```diff
@@ -462,15 +462,15 @@ zone_info_get_offset(ZoneInfo *self, gint64 timestamp)
 
   if (self->last_transitions_index != -1 &&
       self->last_transitions_index < (self->timecnt - 1) &&
-      self->transitions[self->last_transitions_index].time < timestamp &&
+      self->transitions[self->last_transitions_index].time <= timestamp &&
```
furiel: Could you explain this patch a bit more?

I am not familiar with this transition table feature. Can you explain what is it used for?

lib/timeutils/tests/test_unixtime.c
```diff
@@ -73,6 +75,347 @@ Test(unixtime, unix_time_fix_timezone_adjusts_timestamp_as_if_was_parsed_assumin
...
+#undef TESTCASE_START
+}
+
+Test(unixtime, unix_time_fix_timezone_with_tzinfo_to_a_zone_forwards_during_sprint_daylight_saving_hour)
+{
+  UnixTime ut = UNIX_TIME_INIT;
+  UnixTime base_ut;
+
+
+#define TESTCASE_START(base_ofs) \
```
furiel: Me seems as if this is repetitively defined in each test. Would it make sense to pull out globally?

lib/timeutils/zoneinfo.c
bazsi: This loop is processing the daylight saving periods, e.g. when dst is enabled/disabled. The current time and the table is measured in UTC

bazsi: Also, the table contains periods that are back to back, e.g. no seconds in between. And the off by one causes the right entry to be skipped.

You can check the tables using the zdump tool.

lib/timeutils/tests/test_unixtime.c
bazsi: I have added a commit that replaces the macro with a local function. I originally wanted to "highlight" each step in that testcase function and I've figured a macro with all caps name would be a good choice.

I moved that into the comment right above the function call I just introduced.

bazsi:
rebased against current master and addressed all review constraints.

kira-syslogng:
Build SUCCESS

modules/timestamp/rewrite-fix-timezone.c
```diff
@@ -0,0 +1,114 @@
...
+}
+
+static void
+_process(LogRewrite *s, LogMessage **pmsg, const LogPathOptions *path_options)
+{
+  RewriteFixTimeZone *self = (RewriteFixTimeZone *) s;
+  GString *result = scratch_buffers_alloc();
+  LogMessage *msg = *pmsg;
+
+  log_template_format(self->zone_template, msg, NULL, LTZ_LOCAL, 0, NULL, result);
```
furiel: Instead of `LTZ_LOCAL`, `self->stamp` would be more suitable.

modules/timestamp/rewrite-set-timezone.c
```diff
@@ -0,0 +1,113 @@
...
+}
+
+static void
+_process(LogRewrite *s, LogMessage **pmsg, const LogPathOptions *path_options)
+{
+  RewriteSetTimeZone *self = (RewriteSetTimeZone *) s;
+  GString *result = scratch_buffers_alloc();
+  LogMessage *msg = *pmsg;
+
+  log_template_format(self->zone_template, *pmsg, NULL, LTZ_LOCAL, 0, NULL, result);
```
furiel: Similarly as above, `LTZ_LOCAL` could be `self->stamp`.

modules/timestamp/rewrite-fix-timezone.c
```diff
@@ -0,0 +1,114 @@
...
+  gint stamp;
+};
+
+void
+rewrite_fix_time_zone_set_zone_template_ref(LogRewrite *s, LogTemplate *zone_template)
+{
+  RewriteFixTimeZone *self = (RewriteFixTimeZone *) s;
+
+  log_template_unref(self->zone_template);
+  self->zone_template = zone_template;
```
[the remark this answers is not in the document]
furiel: Ok I was wrong. This works similarly to `redis_dd_set_command_ref`. This avoids a ref, unref dance in the grammar file.

modules/timestamp/rewrite-fix-timezone.c
bazsi: That's a different beast. LTZ_LOCAL does not reference a specific timestamp (which self->stamp references) rather it selects which timezone the date macros would use: LTZ_LOCAL uses the timezone of the syslog-ng process (can be overrridden using the local-time-zone() setting in the options block)

LTZ_SEND would use the timezone setting of a destination. Since we are not even in an output context this one doesn't make sense.

Hope it is cleaner with this explanation.

furiel: Ok I understand, thanks!

bazsi:
Thanks \o/

PR #2827, opened 2019-07-11
pcre: calculate matches_size in pcre_re_compile

alltilla:
We can call `pcre_fullinfo()` only once, when we compile the expression, instead of doing these per message.

### DIFF
--- a/lib/logmatcher.c
+++ b/lib/logmatcher.c
@@ -291,7 +291,26 @@ typedef struct _LogMatcherPcreRe
   pcre *pattern;
   pcre_extra *extra;
   gint match_options;
+  gsize matches_size;
 } LogMatcherPcreRe;
+
+static gsize
+_calculate_matches_size(LogMatcherPcreRe *self, const gchar *pattern)
+{
+  gint num_matches;
+  if (pcre_fullinfo(self->pattern, self->extra, PCRE_INFO_CAPTURECOUNT, &num_matches) < 0)
+    g_assert_not_reached();
+  if (num_matches > RE_MAX_MATCHES)
+    {
+      msg_warning("WARNING: The regexp pattern has too many possible matches. "
+                "syslog-ng will only handle a limited number.",
+                evt_tag_str("pattern", pattern),
+                evt_tag_int("possible_matches", num_matches),
+                evt_tag_int("match_limit", RE_MAX_MATCHES));
+      num_matches = RE_MAX_MATCHES;
+    }
+  return (3 * (num_matches + 1));
+}
 
 static gboolean
 log_matcher_pcre_re_compile(LogMatcher *s, const gchar *re, GError **error)
@@ -354,6 +373,8 @@ log_matcher_pcre_re_compile(LogMatcher *s, const gchar *re, GError **error)
                   re, errptr);
       return FALSE;
     }
+
+  self->matches_size = _calculate_matches_size(self, re);
 
   return TRUE;
 }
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

alltilla:

kira-syslogng:
Build SUCCESS

lib/logmatcher.c
```diff
@@ -291,8 +291,22 @@ typedef struct _LogMatcherPcreRe
...
 } LogMatcherPcreRe;
 
+static void
+_alloc_matches(LogMatcherPcreRe *self)
+{
+  gint num_matches;
+  if (pcre_fullinfo(self->pattern, self->extra, PCRE_INFO_CAPTURECOUNT, &num_matches) < 0)
+    g_assert_not_reached();
+  if (num_matches > RE_MAX_MATCHES)
+    num_matches = RE_MAX_MATCHES;
```
furiel: Please add a warning that syslog-ng will only index the first `RE_MAX_MATCHES`.

lib/logmatcher.c
```diff
@@ -425,50 +441,30 @@ static gboolean
...
-    num_matches = RE_MAX_MATCHES;
-
-  matches_size = 3 * (num_matches + 1);
-  matches = g_alloca(matches_size * sizeof(gint));
-
   rc = pcre_exec(self->pattern, self->extra,
-                 value, value_len, 0, self->match_options, matches, matches_size);
-  if (rc < 0)
+                 value, value_len, 0, self->match_options, self->matches, self->matches_size);
+  if (rc == PCRE_ERROR_NOMATCH)
```
furiel: Please split the if-s based on early return when possible, instead of else if.

alltilla: Done! :)

lib/logmatcher.c
alltilla: Done, thanks! :)

kira-syslogng:
Build SUCCESS

lib/logmatcher.c
```diff
@@ -355,20 +375,23 @@ log_matcher_pcre_re_compile(LogMatcher *s, const gchar *re, GError **error)
...
 {
   gint i;
   gboolean indirect = _shall_set_values_indirectly(value_handle);
+  LogMatcherPcreRe *self = (LogMatcherPcreRe *) s;
 
   for (i = 0; i < (RE_MAX_MATCHES) && i < match_num; i++)
     {
-      gint begin_index = matches[2 * i];
-      gint end_index = matches[2 * i + 1];
+      gint begin_index = self->matches[2 * i];
```
furiel: I'm afraid matches was not part of the state of `LogMatchrePcreRe` for a reason. The `log_rewrite_subst_process` (and every filter/parser/rewrite in general) runs in source threads, but they are using the same class instance. This change introduces a race condition: the different source threads are using the same `self->matches` buffer.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/logmatcher.c
alltilla: Agreed! With the new patch, I only moved the `pcre_fullinfo()` call to `pcre_re_compile()`. Thanks!

kira-syslogng:
Build SUCCESS

lib/logmatcher.c
```diff
@@ -425,50 +446,34 @@ static gboolean
 log_matcher_pcre_re_match(LogMatcher *s, LogMessage *msg, gint value_handle, const gchar *value, gssize value_len)
 {
   LogMatcherPcreRe *self = (LogMatcherPcreRe *) s;
-  gint *matches;
-  gsize matches_size;
-  gint num_matches;
   gint rc;
+  gint *matches = g_alloca(self->matches_size * sizeof(gint));
```
Kokan: Would it be possible to do the allocation also in pcre compile phase ?

bazsi: the whole point is that we don't want to share this buffer, and alloca() allocates on the stack, so not really an expensive operation.

If you really wanted to allocate in advance, I would use per-thread allocations where the thread_id (small integer number queried via main_loop_worker_set_thread_id() ), can be used to select which thread state is to be used.

alltilla:
Closing the PR, as this is not a significant improvement performance-vise.

PR #2832, opened 2019-07-12
dbparser: fix crash when context times out in the middle of another rule

MrAnno:
`_advance_time_based_on_message()` is responsible for updating the time based on the incoming message and for triggering  `pattern_db_expire_entry()` on expired contexts.

`pattern_db_expire_entry()` (invoked by `timer_wheel_set_time`) then executes the actions of the expired context. In order to do this, it needs a `PDBProcessParams` instance as an output object, so it can, for example, buffer emitted messages.

An expired context can change the fields of `process_params`, therefore `_advance_time_based_on_message()` should not be called in the middle of another rule execution, because it can lead to crashes.

To avoid such scenarios, `_pattern_db_advance_time_and_flush_expired()` has been added, which triggers the actions of the expired contexts and flushes the output separately.

--------

Reproduction:

```
@version: 3.22
log {
	source { network(port(4444)); };

	parser {
		db-parser(file("boom.xml"));
	};

	destination {
		network("127.0.0.1" port(5555));
	};
};
```

```XML
<patterndb version='4' pub_date='2015-04-13'>
  <ruleset name='test' id='dbe6b2a5e744f9d4522ee189ae2c058898f1c337'>
    <patterns>
      <pattern>BOOM</pattern>
    </patterns>
    <rules>
      <rule provider='me' id='cb4fef7cc01d68cd3cc3fc50799defc16f9977e4' class='sequence'>
        <patterns>
            <pattern>boom @ANYSTRING:boom@</pattern>
        </patterns>
        <actions>
        </actions>
      </rule>
    </rules>
  </ruleset>
  <ruleset name='test2' id='5902f25e-c1aa-4394-8029-a71fd0360081'>
    <patterns>
      <pattern>TAP</pattern>
    </patterns>
    <rules>
      <rule provider='me' id='22a6f6b5-85c3-43ea-a46f-ad0fad248509' context-id='whatever' context-scope='process' class='system'>
        <patterns>
          <pattern>tap @ANYSTRING:tap@</pattern>
        </patterns>
        <actions>
          <action>
            <message inherit-properties='TRUE'>
              <values>
                <value name='MESSAGE'>patterndb correlation</value>
              </values>
            </message>
          </action>
        </actions>
      </rule>
    </rules>
  </ruleset>
</patterndb>

```

Note that the second rule (`22a6f6b5-85c3-43ea-a46f-ad0fad248509`) has an implicit 0-second timeout.

```sh
$ nc -lk 5555 # openbsd-netcat
$ sbin/syslog-ng -Fe
$ cat input.txt
BOOM boom tap
TAP tap bang

$ bin/loggen -dl -R input.txt --inet --rate=1000000 --active-connections=4 localhost 4444
```

After a few seconds, syslog-ng crashes. The coredump clearly shows that `process_params` changed during the execution of a rule. In most cases, the modified `process_params` contains a message that has already been freed. Accessing this message results in a crash.

@faxm0dem

### DIFF
--- a/modules/dbparser/patterndb.c
+++ b/modules/dbparser/patterndb.c
@@ -602,7 +602,6 @@ _pattern_db_process_matching_rule(PatternDB *self, PDBProcessParams *process_par
   GString *buffer = g_string_sized_new(32);
 
   g_static_rw_lock_writer_lock(&self->lock);
-  _advance_time_based_on_message(self, process_params, &msg->timestamps[LM_TS_STAMP]);
   if (rule->context.id_template)
     {
       CorrellationKey key;
@@ -674,12 +673,22 @@ _pattern_db_process_matching_rule(PatternDB *self, PDBProcessParams *process_par
 }
 
 static void
+_pattern_db_advance_time_and_flush_expired(PatternDB *self, LogMessage *msg)
+{
+  PDBProcessParams process_params = {0};
+
+  g_static_rw_lock_writer_lock(&self->lock);
+  _advance_time_based_on_message(self, &process_params, &msg->timestamps[LM_TS_STAMP]);
+  g_static_rw_lock_writer_unlock(&self->lock);
+  _flush_emitted_messages(self, &process_params);
+}
+
+static void
 _pattern_db_process_unmatching_rule(PatternDB *self, PDBProcessParams *process_params)
 {
   LogMessage *msg = process_params->msg;
 
   g_static_rw_lock_writer_lock(&self->lock);
-  _advance_time_based_on_message(self, process_params, &msg->timestamps[LM_TS_STAMP]);
   _emit_message(self, process_params, FALSE, msg);
   g_static_rw_lock_writer_unlock(&self->lock);
 }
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build FAILURE

MrAnno:

kira-syslogng:
Build SUCCESS

faxm0dem:
I've been testing this patch against 3.21.1 for 4 days (~150000000 MSG's) and have never had any crash! Before I had at least 10 segfaults per day.
Thanks so much @MrAnno for your patience and hard work on this one!

modules/dbparser/patterndb.c
```diff
@@ -673,13 +672,23 @@ _pattern_db_process_matching_rule(PatternDB *self, PDBProcessParams *process_par
...
+  g_static_rw_lock_writer_unlock(&self->lock);
+  _flush_emitted_messages(self, &process_params);
+}
+
 static void
 _pattern_db_process_unmatching_rule(PatternDB *self, PDBProcessParams *process_params)
 {
   LogMessage *msg = process_params->msg;
 
   g_static_rw_lock_writer_lock(&self->lock);
```
MrAnno: After the extraction of `_pattern_db_advance_time_and_flush_expired()`, this lock became unnecessary as `_emit_message()` operates only on `process_params` (per-thread, stack variable), and uses only a read-only variable from `self`.

PR #2833, opened 2019-07-12
[WIP] Framework to manipulate FilterExprNode filter expression tree

Kokan:
This is just a draft version, that aims to outline the big picture of such framework.

The filters can be combined with `not`, `or`, `and` and with `(...)` to form a more complex filter. Those are converted directly into a filter expression tree and evaluated every time.

Different filters (while matching the same messages) could have different performance, and it is kinda currently up to the user to choose the best option (if any).

This PR aims to provide a framework for feature developers to optimise that filter expression tree, based on prior measurements.

Example algorithm for the framework:

* Find the chains of same filters (different only in the pattern), that are connected with `or` operator.
   `message("pattern1") or message("pattern2") or ... or message("patternK");`
* Replace those filters with one that contains the combination with `|` regular expression operator (the union of those languages)
   `message("pattern1|pattern2|...|patternK");`

This UML is not updated according to the PR (to be done), but I think still helps to connect some dots.
class diagram

* The inner commits has some code, that later replaced (not every patch clean). Also the first few patch contains general bugfixes, so those could be skipped for now (as some also has alternative fix anyway).
* The new filter is currently created using the config parser, which I intend to replace.

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -272,6 +272,7 @@ lib_libsyslog_ng_la_SOURCES		= \
 	$(transport_sources)		\
 	$(logproto_sources)		\
 	$(filter_sources)		\
+	$(optimizer_sources)		\
 	$(parser_sources)		\
 	$(rewrite_sources)		\
 	$(template_sources)		\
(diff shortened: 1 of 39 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
Closes in favour of referenced PRs.

PR #2834, opened 2019-07-16
[WIP] python: initialize Python "program name" with Py_SetProgramName()

MrAnno:
"The `Py_SetProgramName()` function should be called before `Py_Initialize()` to inform the interpreter about paths to Python run-time libraries."

Setting it before `Py_Initialize()` is useful when syslog-ng is bundled with its own Python interpreter, so it finds the bundled libraries automatically.
It does not cause problems when system Python is used, because it falls back to system search paths.

### DIFF
--- a/modules/python/compat/compat-python-v2.c
+++ b/modules/python/compat/compat-python-v2.c
@@ -25,6 +25,12 @@
 #include "python-helpers.h"
 
 void
+py_set_program_name(void)
+{
+  Py_SetProgramName("syslog-ng");
+}
+
+void
 py_init_argv(void)
 {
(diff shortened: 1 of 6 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

MrAnno:
There are more explicit ways to achieve what I wanted, so I'm closing this one.
#2845 will include the final modification.

PR #2837, opened 2019-07-19
eventlog: flush escaped_buffer when full

Kokan:
This issue could be reproduced with using address sanatizer and debug level, with logs longer than the `escaped_buffer` size (128).

### DIFF
--- a/lib/eventlog/src/evtstr.c
+++ b/lib/eventlog/src/evtstr.c
@@ -117,7 +117,7 @@ evt_str_append_escape_bs(EVTSTR *es,
           escaped_buffer[escaped_buffer_length++] = unescaped[i];
         }
 
-      if (escaped_buffer_capacity < escaped_buffer_length + escaped_char_max_len)
+      if (escaped_buffer_capacity <= escaped_buffer_length + escaped_char_max_len)
         {
           if (!evt_str_append_len(es, escaped_buffer, escaped_buffer_length))
             return 0;


kira-syslogng:
Build SUCCESS

lib/eventlog/src/evtstr.c
```diff
@@ -117,7 +117,7 @@ evt_str_append_escape_bs(EVTSTR *es,
           escaped_buffer[escaped_buffer_length++] = unescaped[i];
         }
 
-      if (escaped_buffer_capacity < escaped_buffer_length + escaped_char_max_len)
+      if (escaped_buffer_capacity <= escaped_buffer_length + escaped_char_max_len)
```
MrAnno: Nice catch.
The buffer overwrite could happen because of `sprintf` writing a fifth byte to the end of the 4-byte sequence.
I think in this case, it would be cleaner to display that +1 byte explicitly in the condition or we should at least mention it in the commit message.

Kokan: sure I could mention in the comment, but would rather avoid +1ing conditions.

kira-syslogng:
Build SUCCESS

lib/eventlog/src/evtstr.c
gaborznagy: @Kokan I think I understand your point: you don't want to highlight sprintf() implementation detail (that it actually copies n+1 bytes) as the if statement would become a less clear code.

PR #2838, opened 2019-07-19
dbparser: remove unnecessary lock

MrAnno:
After the extraction of `_pattern_db_advance_time_and_flush_expired()` in f5d2335ae3aacb4aeb56cbd5cf6bdc049cb6343a, this lock became unnecessary as `_emit_message()` operates only on `process_params` (per-thread, stack variable), and uses only a readonly field from `self`.


kira-syslogng:
Build SUCCESS

PR #2840, opened 2019-07-21
syslog-ng-ctl: add start and stop internal log message command.

mehul-m-prajapati:
Hi,

This patch will add 2 new syslog-ng-ctl commands,
1. syslog-ng-ctl internal-logs --start (Start collecting internal log messages)
2. syslog-ng-ctl internal-logs --stop (Display collected internal log messages and stop further collection of messages)

Later, I will use these commands in debun to save internal log messages in a file.

resolves https://github.com/mehul-m-prajapati/gsoc-debun-tool/issues/5

Thanks,
Mehul


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/afinter.c
```diff
@@ -496,3 +501,75 @@ afinter_global_deinit(void)
     }
   current_internal_source = NULL;
 }
+
+/****************************************************************************
+ * Live collection of internal log messages.
+ ****************************************************************************/
+
+gboolean
+afinter_start_live_collection(GString *result)
```
Kokan: The `result` variable is a coupling between `syslog-ng-ctl` functionality and `afinter`; and as such should not exists. Specially because there is a nice result information already conveyed as a return value, if the `success`/`failure` is not enough to represent the result, you could extend it via an enum describing the states, and not using strings those states.

Don't get me wrong, writing such errors is fine, but the string should be created in the correct layer (`syslog-ng-ctl` related functions).

mehul-m-prajapati:
``

> Could you please explain the reason for this change?

With this change, it will collect logs at higher levels. I think it was one of the requirements of this feature.

lib/afinter.c
mehul-m-prajapati: okay. I will add enum which will describe the states.

mehul-m-prajapati:
> This is inheritly racy, you get messages and stop collection. Any number of message could arrive between the two call. First you have to stop the collection and than collect the logs. (or done in one step)

ok. I will first stop and then get the logs

kira-syslogng:
Build FAILURE

Kokan:
> ``
> 
> > Could you please explain the reason for this change?
> 
> With this change, it will collect logs at higher levels. I think it was one of the requirements of this feature.

Yes, it is. But the question how much higher should it be is still open, well it gave an answer as highest. But imho that is not good enough. So it should be up to the user to decide, it could be configured like `syslog-ng-ctl internal-logs start --debug/--warning`.
On the other hand this is kinda a possibility already via `syslog-ng-ctl`:
```
Syntax: /tmp/install/sbin/syslog-ng-ctl <command> [options]
Possible commands are:
...
    verbose              Enable/query verbose messages
    debug                Enable/query debug messages
    trace                Enable/query trace messages
...
```
Which is good enough for now, one could simply could do:
```
syslog-ng-ctl debug --set 1
syslog-ng-ctl internal-logs start
#wait 
syslog-ng-ctl stop
syslog-ng-ctl debug --set 0
```

lib/afinter.c
```diff
@@ -496,3 +501,74 @@ afinter_global_deinit(void)
...
+        }
+    }
+}
+
+void
+afinter_get_collected_messages(GString *result)
+{
+  LogMessage *msg;
+  const gchar *msg_text;
+
```
Kokan: What happens if `afinter_message_posted` is called while this function executes?

mehul-m-prajapati: @Kokan : In that case, it could create an issue when ```afinter_message_posted``` tries to push message and this function tried to get those at the same time.
May be, I should use ```g_static_mutex_lock(&internal_msg_lock);``` in ```afinter_get_collected_messages``` function

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/afinter.c
```diff
@@ -335,6 +336,10 @@ afinter_sd_init(LogPipe *s)
   if (!log_src_driver_init_method(s))
     return FALSE;
 
+  /* If live collection of internal messages was on, and if we reload
+   * configuration having internal source, then live collection should be stopped. */
+  afinter_reset_live_collection();
```
jszigetvari: Calling this unconditionally will clear the internal source's queue's content even if no "live collection" is running, and the internal source is simply used elsewhere.

lib/afinter.c
```diff
@@ -496,3 +501,77 @@ afinter_global_deinit(void)
...
+          g_free(self);
+        }
+    }
+}
+
+void
+afinter_get_collected_messages(GString *result)
+{
+  LogMessage *msg;
+  const gchar *msg_text;
```
jszigetvari: While not processed, the internal source's queue will hold the internal messages, and once read out, this single variable will hold all the messages. On busy systems, that can mean gigabytes of logs. Is there a limit for the memory consumption of these buffers? Is there any limit in regards to the number of messages they can hold?

lib/afinter.c
```diff
@@ -496,3 +501,77 @@ afinter_global_deinit(void)
...
+
+  g_static_mutex_lock(&internal_msg_lock);
+
+  while (!g_queue_is_empty(internal_msg_queue))
+    {
+      msg = g_queue_pop_head(internal_msg_queue);
+      msg_text = log_msg_get_value(msg, LM_V_MESSAGE, NULL);
```
jszigetvari: msg_text will only hold the MESSAGE part of the log message.
We would need a properly formatted log message here, with timestamps, program field, and all.
You can check how the internal logs look when they are written to STDERR when using -Fedv. That should be your model.

lib/afinter.h
```diff
@@ -28,6 +28,13 @@
 #include "driver.h"
 #include "logsource.h"
 
+enum
```
jszigetvari: You should assign a name to this enum to reference it later as a type.
I will mark later where it should be used: (*)

lib/afinter.h
```diff
@@ -28,6 +28,13 @@
...
+{
+  AF_LIVE_COLLECTION_STARTED,
```
jszigetvari: These constants should be named in a consistent manner. The AF part is not really necessary, however they should have a consistent prefix, like LC_, LIVE_COLLECTION_ or something similar.
That way the constants would become something like LIVE_COLLECTION_STARTED, LIVE_COLLECTION_ALREADY_RUNNING and LIVE_COLLECTION_INTERNAL_SRC_ALREADY_IN_USE

lib/afinter.h
```diff
@@ -43,4 +50,8 @@ LogDriver *afinter_sd_new(GlobalConfig *cfg);
 void afinter_global_init(void);
 void afinter_global_deinit(void);
 
+guint8 afinter_start_live_collection(void);
```
jszigetvari: (*) use the enum type here

lib/mainloop-control.c
```diff
@@ -277,6 +278,48 @@ control_connection_list_files(ControlConnection *cc, GString *command, gpointer
   control_connection_send_reply(cc, result);
 }
 
+static void
+control_internal_logs(ControlConnection *cc, GString *command, gpointer user_data)
+{
+  GString *result = g_string_new("");
+  gchar **arguments = g_strsplit(command->str, " ", 0);
+  guint8 ret;
```
jszigetvari: (*) use the enum type here

syslog-ng-ctl/syslog-ng-ctl.c
```diff
@@ -243,6 +243,31 @@ slng_config(int argc, char *argv[], const gchar *mode, GOptionContext *ctx)
...
+{
+  { "start", 's', 0, G_OPTION_ARG_NONE, &config_options_start, "start live collection of internal logs", NULL },
+  { "stop", 'x', 0, G_OPTION_ARG_NONE, &config_options_stop, "stop live collection of internal logs", NULL },
+  { NULL,           0,   0, G_OPTION_ARG_NONE, NULL,                         NULL,           NULL }
+};
+
+static gint
+slng_internal(int argc, char *argv[], const gchar *mode, GOptionContext *ctx)
+{
+  GString *cmd = g_string_new("INTERLOGS ");
```
jszigetvari: Why do we concatenate strings here? We could as well use constant strings, which would be much more efficient.

syslog-ng-ctl/syslog-ng-ctl.c
```diff
@@ -243,6 +243,31 @@ slng_config(int argc, char *argv[], const gchar *mode, GOptionContext *ctx)
...
+
+  if (config_options_start)
+    g_string_append(cmd, "START");
+  else if (config_options_stop)
+    g_string_append(cmd, "STOP");
+
+  gint res = _dispatch_command(cmd->str);
```
jszigetvari: If we don't specify either start or stop on the CLI, then this code will simply send the "INTERLOGS " string over the socket, which is problematic.

lib/afinter.h
MrAnno: Note: The `LC_` prefix is reserved (by `CfgTree`: `LC_CATCHALL`, `LC_FALLBACK`, etc.).
`AFINTER_`, for example, can be a good choice.

lib/afinter.c
mehul-m-prajapati: > Is there a limit for the memory consumption of these buffers? Is there any limit in regards to the number of messages they can hold?

I think there is no limit as of now. @Kokan had discussed this scenario with me during GSoC.
We agreed that it is the responsibility of the user to start/stop live collection at an appropriate time.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

syslog-ng-ctl/syslog-ng-ctl.c
mehul-m-prajapati: @jszigetvari : Yes, I will use the constant strings and remove this part

syslog-ng-ctl/syslog-ng-ctl.c
mehul-m-prajapati: @jszigetvari : With the usage of constant strings, this will be resolved.

lib/mainloop-control.c
mehul-m-prajapati: @jszigetvari : okay

lib/afinter.h
mehul-m-prajapati: @jszigetvari : ok

lib/afinter.h
mehul-m-prajapati: @MrAnno : Yes, I am now using ```AINTER_```

lib/afinter.h
mehul-m-prajapati: @jszigetvari : Okay

lib/afinter.c
mehul-m-prajapati: @jszigetvari : Now, I have put the live_collection flag check inside the ```afinter_reset_live_collection``` function.

kira-syslogng:
Build SUCCESS

lib/afinter.c
mehul-m-prajapati: @jszigetvari : Yes, I have checked debug messages with timestamps, program field etc. 

I think ```log_writer_format_log``` function formats the internal logs. 

Is there any function similar to this, which I can use to format the internal log messages?

lib/afinter.c
jszigetvari: Okay, I can accept that.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

syslog-ng-ctl/syslog-ng-ctl.c
```diff
@@ -61,6 +61,34 @@ slng_reload(int argc, char *argv[], const gchar *mode, GOptionContext *ctx)
...
+  { "start", 's', 0, G_OPTION_ARG_NONE, &config_options_start, "start live collection of internal logs", NULL },
+  { "stop", 'x', 0, G_OPTION_ARG_NONE, &config_options_stop, "stop live collection of internal logs", NULL },
+  { NULL,           0,   0, G_OPTION_ARG_NONE, NULL,                         NULL,           NULL }
+};
+
+static gint
+slng_internal(int argc, char *argv[], const gchar *mode, GOptionContext *ctx)
+{
+  GString *cmd = g_string_new("");
+
```
furiel: If neither start or stop are provided, I think we should write an error. Now the cli exits silently. Well the error code is not zero, so not entirely silently. Still, we could write something out.

syslog-ng-ctl/syslog-ng-ctl.c
```diff
@@ -61,6 +61,34 @@ slng_reload(int argc, char *argv[], const gchar *mode, GOptionContext *ctx)
...
+  if (config_options_start)
+    g_string_assign(cmd, "INTERLOGS START");
+  else if (config_options_stop)
+    g_string_assign(cmd, "INTERLOGS STOP");
+  else
+    return 1;
```
furiel: We should not allow that start and stop are provided at the same time.

lib/mainloop-control.c
```diff
@@ -409,6 +410,48 @@ export_config_graph(ControlConnection *cc, GString *command, gpointer user_data)
...
+          break;
+
+        default:
+          break;
+        }
+    }
+  else if (g_str_equal(arguments[1], "STOP"))
+    {
+      afinter_stop_live_collection();
+      afinter_get_collected_messages(result);
```
furiel: I get assertion error with the following case:
```
@version: 3.25

log {
    destination { file(/dev/stdout); };
};
```

Start syslog-ng with `./syslog-ng -Fe -f ../etc/internal.conf`.

After calling start, and then stop, syslog-ng emits: 
```
[2020-03-03T13:05:58.048621] g_queue_is_empty: assertion 'queue != NULL' failed
```

lib/afinter.c
```diff
@@ -496,3 +501,83 @@ afinter_global_deinit(void)
...
+      g_string_append_printf(result, "%s\n", msg_text);
+      stats_counter_dec(internal_queue_length);
+    }
+
+  if (!strlen(result->str))
+    g_string_assign(result, "No live messages collected");
```
furiel: The log collection does not work for me some reason. I am generating messages continuously with:
```
@version: 3.25

log {
    destination { file(/dev/stdout); };
};
```

Starting syslog-ng with -Fevdt, I can see internal logs generating continuously. However when I start the cli, wait a few seconds, then stop it, I only see this error message: No live messages collected.

furiel: My bad. If I start syslog-ng in foreground, then there will be no internal messages.

Could you please handle as error if someone wants to start the log collection while syslog-ng is running in foreground? Others can run into this as well.

lib/afinter.c
```diff
@@ -496,3 +501,83 @@ afinter_global_deinit(void)
...
+
+void
+afinter_get_collected_messages(GString *result)
+{
+  LogMessage *msg;
+  const gchar *msg_text;
+
+  g_static_mutex_lock(&internal_msg_lock);
+
+  while (!g_queue_is_empty(internal_msg_queue))
```
furiel: I steadily get deadlock in the following case:
```
@version: 3.25

log {
    destination { file(/dev/stdout); };
};
```

Start syslog-ng with `./syslog-ng -F -f ../etc/internal.conf` (! just -F, not -Fdt. With -Fdt, I did not see deadlock yet).

After start and stop, syslog-ng hangs. I crashed syslog-ng with segv (kill -segv <pid>), then checked the core dump with gdb: I see syslog-ng hangs while trying to grab a lock during pushing internal message. Moving up in the stack, the message is triggered by `g_queue_is_empty` (through a `g_log` call). The hint is `internal_msg_queue` is empty for some reason, and I think the code wants to emit the same assert that I mentioned in the other comment. So maybe the root cause and the fix is the same.

lib/afinter.c
```diff
@@ -496,3 +501,83 @@ afinter_global_deinit(void)
...
+    {
+      msg = g_queue_pop_head(internal_msg_queue);
```
furiel: This message needs to be unreffed  using `log_msg_unref`.

furiel:
@mehul-m-prajapati I think this PR is close to the finish line. Did you have time to look into the review comments? Is there anything to be clarified? Let me know if you need any help.

mehul-m-prajapati:
> @mehul-m-prajapati I think this PR is close to the finish line. Did you have time to look into the review comments? Is there anything to be clarified? Let me know if you need any help.

Hi @furiel ,

Thanks for taking time to review my PR. I will do changed in this week.

mehul-m-prajapati:
Hi @HofiOne,

I am actively working on the review changes for this PR. 
Please remove the `abandoned` tag and replace it with `work in progress.`

Thanks,
Mehul

syslog-ng-ctl/syslog-ng-ctl.c
mehul-m-prajapati: @furiel : Okay, I have added the error message for cases where neither 'start' nor 'stop' is provided.

lib/mainloop-control.c
```diff
@@ -409,6 +410,48 @@ export_config_graph(ControlConnection *cc, GString *command, gpointer user_data)
   control_connection_send_reply(cc, result);
 }
 
+static void
+control_internal_logs(ControlConnection *cc, GString *command, gpointer user_data)
+{
+  GString *result = g_string_new("");
+  gchar **arguments = g_strsplit(command->str, " ", 0);
```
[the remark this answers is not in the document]
mehul-m-prajapati: @furiel : okay, done.

lib/mainloop-control.c
mehul-m-prajapati: @furiel : I tried this scenario and did not receive the error in the new version.

lib/afinter.c
mehul-m-prajapati: @furiel : Okay, I have added a check to ensure that when syslog-ng is running in the foreground, it does not start internal logs

lib/afinter.c
mehul-m-prajapati: @furiel : okay, done.

syslog-ng-ctl/syslog-ng-ctl.c
mehul-m-prajapati: @furiel : Are you talking about the command: `syslog-ng-ctl internal-logs --start --stop` ?
Can you suggest any solution for this?

lib/afinter.c
mehul-m-prajapati: @furiel : this is fixed in the new changes.

mehul-m-prajapati:
Hi @HofiOne,

Can you please review this PR ?

Thanks,
Mehul

HofiOne:
> Can you please review this PR ?

I'll do my best, I promise, but I do have a bit too much on my plate, sorry.

lib/afinter.c
```diff
@@ -635,3 +641,107 @@ afinter_get_metrics(void)
...
+{
+  afinter_stop_live_collection();
+
+  if (is_live_collection)
+    {
+      if (internal_msg_queue)
+        _release_internal_msg_queue();
+    }
+}
+
```
HofiOne, line 687: I guees `afinter_stop_live_collection()` must also be called in somewhere in the cleanup chain e.g. in `afinter_sd_deinit()`

lib/afinter.c
```diff
@@ -635,3 +641,107 @@ afinter_get_metrics(void)
...
+/****************************************************************************
+ * Live collection of internal log messages.
+ ****************************************************************************/
+
+AFInterLive
+afinter_start_live_collection(void)
+{
+  AFInterSource *self;
+
+  if (is_live_collection)
```
HofiOne, line 654: for a better concurrency handling this whole initialization must also be protected via `internal_msg_lock` mutex

lib/afinter.c
```diff
@@ -635,3 +641,107 @@ afinter_get_metrics(void)
...
+    {
+      if (internal_msg_queue)
+        _release_internal_msg_queue();
+    }
+}
+
+void
+afinter_stop_live_collection(void)
+{
```
HofiOne, line 690: for a better concurrency handling you must protect this whole block with the internal_msg_lock mutex

lib/afinter.c
```diff
@@ -635,3 +641,107 @@ afinter_get_metrics(void)
...
+  g_string_append_printf(result, "Size of internal logs: %d", g_queue_get_length(internal_msg_queue));
+}
+
+void
+afinter_get_collected_messages(GString *result)
+{
+  LogMessage *log_msg;
+  GString *msg;
+  LogTemplate *template = NULL;
+
```
HofiOne, line 719: this one is also needs a better concurrency protection

you must be sure that e.g. the `internal_msg_queue` stat is not changing between the NULL check and the pop

HofiOne (changes requested):

lib/afinter.c
mehul-m-prajapati: @HofiOne : okay, done 👍

lib/afinter.c
mehul-m-prajapati: @HofiOne : okay, done 👍

lib/afinter.c
mehul-m-prajapati: @HofiOne : okay, done 👍

lib/afinter.c
mehul-m-prajapati: @HofiOne : okay, I added protection for the whole block.

lib/afinter.c
```diff
@@ -635,3 +643,107 @@ afinter_get_metrics(void)
...
+  g_mutex_lock(&internal_msg_lock);
```
HofiOne, line 723: my main concerns here are
- the queue can grow to a huge size here that can eat up all the memories if a huge number of messages are coming
- it also means a doubled memory usage because of the copied content in the result response string
- locking the queue here, during the whole content processing, means a lock on the sender side too, as I can see on the main thread that will stop syslog-ng message processing for this time, yes, I know, it is called only now after the log processing stopped, but as this function publicly accessible, nothing prevents it to be called from anywhere else.
I suggest making this static local only and calling it from the stop directly, internally.

HofiOne (changes requested):

lib/afinter.c
```diff
@@ -635,3 +643,107 @@ afinter_get_metrics(void)
...
+      msg = g_string_sized_new(1024);
+
+      while (!g_queue_is_empty(internal_msg_queue))
+        {
+          log_msg = g_queue_pop_head(internal_msg_queue);
+          /* format log */
+          template = log_template_new(configuration, NULL);
+          log_template_compile(template, "[$DATE] $HOST $MSGHDR$MSG\n", NULL);
+          log_template_format(template, log_msg, &DEFAULT_TEMPLATE_EVAL_OPTIONS, msg);
+          g_string_append_printf(result, "%s\n", msg->str);
```
HofiOne, line 735: i guess the `\n` is not needed here

HofiOne (changes requested):

lib/afinter.c
```diff
@@ -635,3 +643,107 @@ afinter_get_metrics(void)
...
+          g_free(self);
+        }
+    }
+  g_mutex_unlock(&internal_msg_lock);
+}
+
+void
+afinter_get_size_of_internal_logs(GString *result)
+{
+  g_string_append_printf(result, "Size of internal logs: %d", g_queue_get_length(internal_msg_queue));
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : okay, done.

syslog-ng-ctl/syslog-ng-ctl.c
```diff
@@ -63,6 +63,41 @@ slng_reload(int argc, char *argv[], const gchar *mode, GOptionContext *ctx)
...
+static gboolean config_options_start = FALSE;
+static gboolean config_options_stop = FALSE;
+static gboolean config_options_size = FALSE;
+
+static GOptionEntry internal_options[] =
+{
+  { "start", 's', 0, G_OPTION_ARG_NONE, &config_options_start, "start live collection of internal logs", NULL },
+  { "stop", 'x', 0, G_OPTION_ARG_NONE, &config_options_stop, "stop live collection of internal logs", NULL },
+  { "size", 'l', 0, G_OPTION_ARG_NONE, &config_options_size, "get size of internal logs", NULL },
+  { NULL,           0,   0, G_OPTION_ARG_NONE, NULL,                         NULL,           NULL }
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : okay, done.

syslog-ng-ctl/syslog-ng-ctl.c
```diff
@@ -126,6 +161,7 @@ static CommandDescriptor modes[] =
   { "list-files", no_options, "Print files present in config", slng_listfiles, NULL },
   { "export-config-graph", no_options, "export configuration graph", slng_export_config_graph, NULL },
   { "healthcheck", healthcheck_options, "Health check", slng_healthcheck, NULL },
+  { "internal-logs", internal_options, "Collect internal syslog-ng logs", slng_internal, NULL },
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : Are you talking about help options?
```
mehul@mehul-VirtualBox ~/mehul/syslog-ng (gsoc-internal-logs)$ sudo ./syslog-ng-ctl/syslog-ng-ctl internal-logs --help
Usage:
  syslog-ng-ctl [OPTION…] /home/mehul/mehul/syslog-ng/syslog-ng-ctl/.libs/syslog-ng-ctl

Collect internal syslog-ng logs

Help Options:
  -h, --help                 Show help options

Application Options:
  -s, --start                start live collection of internal logs
  -x, --stop                 stop live collection of internal logs
  -l, --size                 get size of internal logs
  -c, --control=<socket>     syslog-ng control socket

```

lib/mainloop-control.c
```diff
@@ -450,6 +452,61 @@ export_config_graph(ControlConnection *cc, GString *command, gpointer user_data,
...
 
+static void
+control_internal_logs(ControlConnection *cc, GString *command, gpointer user_data, gboolean *cancelled)
+{
+  GString *result = g_string_new("");
+  gchar **arguments = g_strsplit(command->str, " ", 0);
+
+  if (g_str_equal(arguments[1], "START"))
+    {
+      AFInterLive ret;
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : okay, done.

lib/afinter.c
mehul-m-prajapati: @HofiOne : okay, done.

lib/afinter.c
mehul-m-prajapati: @HofiOne : okay, good one.

syslog-ng-ctl/syslog-ng-ctl.c
```diff
@@ -63,6 +63,41 @@ slng_reload(int argc, char *argv[], const gchar *mode, GOptionContext *ctx)
...
+      fprintf(stderr, "Error: Unknown command\n");
+      return 1;
+    }
+
+  gint res = dispatch_command(cmd->str);
+  g_string_free(cmd, TRUE);
+
+  return res;
+}
+
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : okay, done.

lib/mainloop-control.c
```diff
@@ -450,6 +452,61 @@ export_config_graph(ControlConnection *cc, GString *command, gpointer user_data,
...
+          g_string_assign(result, "FAIL Error: Internal source is already configured");
+          break;
+
+        default:
+          break;
+        }
+    }
+  else if (g_str_equal(arguments[1], "STOP"))
+    {
+      afinter_stop_live_collection();
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : okay, done.

lib/mainloop-control.c
```diff
@@ -450,6 +452,61 @@ export_config_graph(ControlConnection *cc, GString *command, gpointer user_data,
...
+      afinter_get_collected_messages(result);
+    }
+  else if (g_str_equal(arguments[1], "SIZE"))
+    {
+      afinter_get_size_of_internal_logs(result);
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : okay, done.

lib/afinter.c
```diff
@@ -635,3 +643,107 @@ afinter_get_metrics(void)
...
 }
+
+/****************************************************************************
+ * Live collection of internal log messages.
+ ****************************************************************************/
+
+AFInterLive
+afinter_start_live_collection(void)
+{
+  AFInterSource *self;
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : okay, done.

kira-syslogng:
Build FAILURE

lib/mainloop-control.c
```diff
@@ -450,6 +452,68 @@ export_config_graph(ControlConnection *cc, GString *command, gpointer user_data,
   control_connection_send_reply(cc, result);
 }
 
+static void
+control_internal_logs(ControlConnection *cc, GString *command, gpointer user_data, gboolean *cancelled)
+{
+  GString *result = g_string_new("");
+  gchar **arguments = g_strsplit(command->str, " ", 0);
+  static AFInterLive ret = AFINTER_LIVE_COLLECTION_NONE;
```
HofiOne, line 460: Please do not try to double-manage the internal state of the live collection.

both `afinter_stop_live_collection` and `afinter_get_size_of_internal_logs` can give you the state back, like in the case of `afinter_start_live_collection` you did

or, alternatively, you can add a public state-getter if it is really needed (but, I guess this is not that case, as I mentioned above, both functions can signal the state back like `afinter_start_live_collection` does

HofiOne (changes requested):

lib/mainloop-control.c
mehul-m-prajapati: @HofiOne : okay, done.

lib/afinter.c
```diff
@@ -635,3 +643,117 @@ afinter_get_metrics(void)
...
+
+  current_internal_source = self;
+  is_live_collection = TRUE;
+
+  g_mutex_unlock(&internal_msg_lock);
+
+  return AFINTER_LIVE_COLLECTION_STARTED;
+}
+
+void
```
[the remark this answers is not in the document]
mehul-m-prajapati: okay, done.

syslog-ng-ctl/syslog-ng-ctl.c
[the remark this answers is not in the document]
mehul-m-prajapati: okay, done.

mehul-m-prajapati:
@HofiOne : Added relevant documentation here: https://github.com/syslog-ng/syslog-ng.github.io/pull/141

lib/afinter.c
```diff
@@ -635,3 +643,107 @@ afinter_get_metrics(void)
...
+
+AFInterLive
+afinter_start_live_collection(void)
+{
+  AFInterSource *self;
+
+  g_mutex_lock(&internal_msg_lock);
+
+  if (is_live_collection)
+    return AFINTER_LIVE_COLLECTION_RUNNING;
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : There is an unlock function call before the end of this function.

lib/afinter.c
```diff
@@ -635,3 +643,107 @@ afinter_get_metrics(void)
...
+
+  if (current_internal_source != NULL)
+    return AFINTER_INTERNAL_SRC_PRESENT;
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : Do you mean lock/unlock should be for each `if` condition here?

lib/afinter.c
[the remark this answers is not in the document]
HofiOne: but you might return from the function here, before reaching that unlock ever

lib/afinter.c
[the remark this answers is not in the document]
HofiOne: please see my comment above

lib/afinter.c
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : okay, I have put lock after both `if` conditions.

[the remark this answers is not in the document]
HofiOne: The purpose of locking here is to protect the internal objects from concurrent access (read/write), this way you leave `is_live_collection` and `current_internal_source` unprotected, and concurrent threads can modify them between entering into the `afinter_start_live_collection` and the `g_mutex_lock(&internal_msg_lock)` can take effect in line 661.

e.g. this can be a better approach

```
AFInterLive
afinter_start_live_collection(void)
{
  g_mutex_lock(&internal_msg_lock);

  if (is_live_collection)
    {
      g_mutex_unlock(&internal_msg_lock);
      return AFINTER_LIVE_COLLECTION_RUNNING;
    }
  if (current_internal_source != NULL)
    {
      g_mutex_unlock(&internal_msg_lock);
      return AFINTER_INTERNAL_SRC_PRESENT;
    }

  AFInterSource *self = g_new0(AFInterSource, 1);
  AFInterSourceOptions *options = g_new0(AFInterSourceOptions, 1);

  afinter_source_options_defaults(options);
  self->options = options;
  stats_counter_set(metrics.queue_capacity, self->options->queue_capacity);

  current_internal_source = self;
  is_live_collection = TRUE;

  g_mutex_unlock(&internal_msg_lock);

  return AFINTER_LIVE_COLLECTION_STARTED;
}
```

lib/mainloop-control.c
```diff
@@ -450,6 +452,63 @@ export_config_graph(ControlConnection *cc, GString *command, gpointer user_data,
   control_connection_send_reply(cc, result);
 }
 
+static void
+control_internal_logs(ControlConnection *cc, GString *command, gpointer user_data, gboolean *cancelled)
+{
+  GString *result = g_string_new("");
+  gchar **arguments = g_strsplit(command->str, " ", 0);
+  AFInterLive status = afinter_get_live_collection_status();
```
HofiOne, line 460: as I mentioned, in this case, signaling back to the caller a probably bad status would be better implemented in the `afinter_stop_live_collection` and the `afinter_get_size_of_internal_logs` functions

the main reason is that both of them can follow a flow like

- locking properly
- determining a possible invalid status, and 
- unclocking and returning with an invalid status signal, or
- doing their job
- unlocking

this will guarantee that the internal state will not change during the whole process, that is not the case unfortunately using this `afinter_get_live_collection_status()` function

HofiOne (changes requested):

HofiOne:
> @HofiOne : Added relevant documentation here: [syslog-ng/syslog-ng.github.io#141](https://github.com/syslog-ng/syslog-ng.github.io/pull/141)

thanks! to be able to contribute you must accept the CLA, so please do so.

Screenshot 2024-09-30 at 11 18 35

lib/afinter.c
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : okay, thanks for the code snipped.

lib/mainloop-control.c
mehul-m-prajapati: @HofiOne : okay, thanks for the insights.

mehul-m-prajapati:
> > @HofiOne : Added relevant documentation here: [syslog-ng/syslog-ng.github.io#141](https://github.com/syslog-ng/syslog-ng.github.io/pull/141)
> 
> thanks! to be able to contribute you must accept the CLA first, so please do so.
> 
> Screenshot 2024-09-30 at 11 18 35

okay done.

lib/afinter.c
```diff
@@ -635,3 +644,141 @@ afinter_get_metrics(void)
...
+          log_template_format(template, log_msg, &DEFAULT_TEMPLATE_EVAL_OPTIONS, msg);
+          g_string_append_printf(result, "%s\n", msg->str);
+          log_msg_unref(log_msg);
+          log_msg = NULL;
+          stats_counter_dec(internal_queue_length);
+        }
+
+      g_string_free(msg, TRUE);
+      internal_msg_queue = NULL;
+    }
```
HofiOne, line 728: i can't see why this would be needed, as the queue itself should/can be reused in case of a future start/stop call

also it lacks any kind of free of the queue (either via _release_internal_msg_queue or g_queue_free_full, etc.)

HofiOne (changes requested):

lib/mainloop-control.c
```diff
@@ -450,6 +452,63 @@ export_config_graph(ControlConnection *cc, GString *command, gpointer user_data,
   control_connection_send_reply(cc, result);
 }
 
+static void
+control_internal_logs(ControlConnection *cc, GString *command, gpointer user_data, gboolean *cancelled)
+{
+  GString *result = g_string_new("");
+  gchar **arguments = g_strsplit(command->str, " ", 0);
+  AFInterLive status = AFINTER_LIVE_COLLECTION_INIT;
```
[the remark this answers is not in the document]
mehul-m-prajapati: @HofiOne : okay, done.

lib/afinter.c
HofiOne: ok, but why do you release the queue? 

though it is reallocated in the `afinter_message_posted(LogMessage *msg)` at the next message posted, but in that case the counters will be re-initialized as well (that is surely wrong, as there was no reset of them anywhere)

you emptied the queue, but you can leave the container itself as it is

`afinter_global_deinit` will take care of it when it is not needed anymore, meaningless to destroy the container as I can see

mehul-m-prajapati: @HofiOne : so, should I remove `g_queue_free_full ` ?

HofiOne (changes requested):

lib/afinter.c
mehul-m-prajapati: @HofiOne : okay removed the `free` call

mehul-m-prajapati:
@HofiOne : Can you review and move forward?

HofiOne:
> @HofiOne : Can you review and move forward?

@mehul-m-prajapati, sorry, I had to stop, we had a tight schedule we must focus on now :S
Meanwhile, there have been a few interesting changes around the separated, controlled collection of some log types which might have an effect on your PR as well.
I'll keep you informed I promised, but for now, I ask for some more patience. Thank you!

mehul-m-prajapati:
@HofiOne : How are you doing?

MrAnno:
I no longer have permissions to unassign myself. @HofiOne Can you remove me from the reviewers, please?

HofiOne:
> I no longer have permissions to unassign myself. @HofiOne Can you remove me from the reviewers, please?

i do not have the proper rights either it seems, I asked Folti to manage this, sorry.

mehul-m-prajapati:
Closing PR because of low priority and lack of review after new changes

HofiOne:
This feature is implemented in https://github.com/syslog-ng/syslog-ng/pull/5238 in a more generic way.

HofiOne:
The latest enhancements supersede this solution https://github.com/syslog-ng/syslog-ng/pull/5238

mehul-m-prajapati:
okay, thanks @HofiOne

PR #2844, opened 2019-07-25
Light: fix get_short_hostname(), use keep-hostname(yes) for basic tests

MrAnno:
This PR makes Travis green again.

Travis now has multiple hostnames:
```
127.0.1.1 travis-job-0198b942-d444-4c1f-95e9-84a18577f20d travis-job-0198b942-d444-4c1f-95e9-84a18577f20d ip4-loopback trusty64
127.0.0.1 localhost nettuno travis vagrant travis-job-0198b942-d444-4c1f-95e9-84a18577f20d 
```

In Light, `get_short_hostname()` returns a different hostname than what syslog-ng resolves, making a few non-hostname-related tests fail.

In my opinion, a basic file source acceptance test and a catch-all path test should not use dynamic values, so I changed them to a static hostname and enabled `keep-hostname()`.
I have the same opinion on `LogMessage`. If it has a default value for hostname, it should not be dynamic (should not be dependent of the environment).

Of course, `get_short_hostname()` can be reimplemented to return the same name syslog-ng would resolve, but these 2 tests are not meant to test such thing.

### DIFF
--- a/tests/python_functional/functional_tests/logpath/test_flags_catch_all.py
+++ b/tests/python_functional/functional_tests/logpath/test_flags_catch_all.py
@@ -40,6 +40,7 @@ def test_flags_catch_all(config, syslog_ng, log_message, bsd_formatter):
 
     config.create_logpath(statements=[file_source, inner_logpath])
     config.create_logpath(statements=[catch_all_destination], flags="catch-all")
+    config.create_global_options(keep_hostname="yes")
 
     input_message = bsd_formatter.format_message(log_message)
     expected_message = bsd_formatter.format_message(log_message.remove_priority())
--- a/tests/python_functional/functional_tests/source_drivers/file_source/test_acceptance.py
+++ b/tests/python_functional/functional_tests/source_drivers/file_source/test_acceptance.py
@@ -22,10 +22,8 @@
 #############################################################################
 import pytest
 
-from src.common.network_operations import get_short_hostname
-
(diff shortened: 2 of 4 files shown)


kira-syslogng:
Build SUCCESS

Kokan:
while this is imho acceptable as per current test, it is still a workaround for the issue ahead of us.
also this implies that the framework/test system cannot test `keep-hostname(no)` which is default for now, so somehow I have to remember to flip a switch in new tests just because (that sucks).
if we accept this, at least there should be an open issue about the `get_short_hostname()` issue.

MrAnno:
OK, I "fixed" `get_short_hostname()`:

socket.getfqdn():
>  Return a fully qualified domain name for name. To find the fully
>  qualified name, the hostname returned by gethostbyaddr() is checked,
>  followed by aliases for the host, if available. The first name which
>  includes a period is selected. In case no fully qualified domain name
>  is available, the hostname as returned by gethostname() is returned.

Unfortunately, this is not entirely true when a local name is resolved.

For example:
`gethostname() == travis-job-1234`

/etc/hosts:
```
127.0.1.1 travis-job-1234
127.0.0.1 localhost travis-job-1234
```

In this case, `getfqdn(gethostname())` returns `localhost`, because it starts to use (prefers) 127.0.0.1 instead of 127.0.1.1.

This is not an incorrect behavior, but Light should resolve names the same way as syslog-ng does.

kira-syslogng:
Build SUCCESS

PR #2846, opened 2019-07-26
[WIP]run-id: validate if run-id written correctly to persist file

lbudai:
Why we need to validate this?
Thanks to @furiel, we know that there are some persist state related
issues across syslog-ng versions, that can lead to a weird situation,
namely, even we increase run-id, it won't be written to the persist
file.

@furiel : please check if my understanding is correct.

### DIFF
--- a/lib/run-id.c
+++ b/lib/run-id.c
@@ -38,6 +38,29 @@ typedef struct _RunIDState
   gint run_id;
 } RunIDState;
 
+static gboolean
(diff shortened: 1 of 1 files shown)


lbudai:
Notes:
 * error message could be better
 * please note that it does not fix the persist file, just failed to start syslog-ng: it's a little bit unfriendly step, so maybe it would be enough to report a warning message, or, if possible, we could fix the persist file

PR #2849, opened 2019-07-29
logmatcher: posix not supported, not even for 3.6

Kokan:
pcre/posix support in logmatcher

version | pcre | posix 
------------ | ------------- | -------------
3.6 or below | supported | default
3.14 or above | default | not supported

The removed code handled the case when the configuration was still 3.6, as it sets the default handle to `posix` instead of `pcre`, the called setter function replaces the `posix` with `pcre` as `posix` is no longer supported.

The `GlobalConfig` was passed around just to do the above `pcre`=>`posix`=>`pcre` dance.

### DIFF
--- a/lib/logmatcher.c
+++ b/lib/logmatcher.c
@@ -201,7 +201,7 @@ log_matcher_string_replace(LogMatcher *s, LogMessage *msg, gint value_handle, co
 }
 
 LogMatcher *
-log_matcher_string_new(GlobalConfig *cfg, const LogMatcherOptions *options)
+log_matcher_string_new(const LogMatcherOptions *options)
 {
   LogMatcherString *self = g_new0(LogMatcherString, 1);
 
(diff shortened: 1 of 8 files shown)


kira-syslogng:
Build SUCCESS

lib/logmatcher.c
```diff
@@ -736,16 +736,12 @@ log_matcher_options_defaults(LogMatcherOptions *options)
...
 
 void
-log_matcher_options_init(LogMatcherOptions *options, GlobalConfig *cfg)
+log_matcher_options_init(LogMatcherOptions *options)
 {
   if (!options->type)
     {
       const gchar *default_matcher = "pcre";
 
-      if (cfg_is_config_version_older(cfg, 0x0306))
```
gaborznagy: Should we warn the user the same way as if in case of explicit `posix` regex?

gaborznagy: My mistake, we do the same thing.
See `log_matcher_options_set_type`:
```
 if (strcmp(type, "posix") == 0) 
   msg_warning(...);
   type = "pcre";
```

PR #2851, opened 2019-07-31
Travis,cmake: ut verbose if failed

Kokan:
The autotools find and prints the output of test execution, in case of cmake only the result failed/passed is printed. This PR changes the behaviour, so if a test failed it act as if verbose flag were used, otherwise state passed as before.

Also contains some cleanup, and renaming.


kira-syslogng:
Build SUCCESS

PR #2855, opened 2019-08-02
Dqtool relocate diskq

### LINKED ISSUE #1851: disq location cannot be modified

lbudai:
Add `relocate` function to `dqtool`.
Fixes #1851

##### examples:
* relocate a single queue file:

```    
    bin/dqtool relocate --new_path /tmp/dq --persist var/syslog-ng.persist /tmp/syslog-ng-00000.rqf
```
    
* relocate multiple queue files

```
    bin/dqtool relocate --new_path /tmp/dq --persist var/syslog-ng.persist /tmp/syslog-ng-00000.rqf /tmp/syslog-ng-00000.qf
```
    
* relocate all queue files

```
    bin/dqtool relocate --new_path /tmp/dq --persist var/syslog-ng.persist --all
```

### DIFF
--- a/modules/diskq/dqtool.c
+++ b/modules/diskq/dqtool.c
@@ -33,6 +33,7 @@
 #include "logmsg/logmsg-serialize.h"
 #include "scratch-buffers.h"
 #include "mainloop.h"
+#include "pathutils.h"
 
 #include <stdio.h>
 #include <string.h>
@@ -64,7 +65,7 @@ static GOptionEntry relocate_options[] =
 {
   {
     "new_path", 'n', 0, G_OPTION_ARG_STRING, &new_diskq_path,
-    "New path for diskq file", "<new_path>"
+    "New path(directory) for diskq file(s)", "<new_path>"
   },
   {
     "persist", 'p', 0, G_OPTION_ARG_STRING, &persist_file_path,
@@ -193,19 +194,51 @@ dqtool_info(int argc, char *argv[])
 }
 
 static gboolean
-_relocate_validate_options(void)
+_is_read_writable(const gchar *path)
 {
-  if (!new_diskq_path)
+  return access(path, R_OK|W_OK) == F_OK;
+}
+
+static gboolean
+_validate_diskq_path(const gchar *path)
+{
+  if (!path)
     {
       fprintf(stderr, "relocate: missing mandatory option: new_path\n");
+      return FALSE;
     }
 
-  if (!persist_file_path)
+  if (!(is_file_directory(path) && _is_read_writable(path)))
+    {
+      fprintf(stderr, "relocate: new_path should be point to a readable/writable directory\n");
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
+_validate_persist_file_path(const gchar *path)
+{
+  if (!path)
     {
       fprintf(stderr, "relocate: missing mandatory option: persist\n");
+      return FALSE;
     }
 
-  return (new_diskq_path != NULL) && (persist_file_path != NULL);
+  if (!(is_file_regular(path) && _is_read_writable(path)))
+    {
+      fprintf(stderr, "relocate: persist should point to a readable/writable file\n");
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
+_relocate_validate_options(void)
+{
+  return _validate_diskq_path(new_diskq_path) && _validate_persist_file_path(persist_file_path);
 }
 
 static gint
--- a/modules/diskq/dqtool.c
+++ b/modules/diskq/dqtool.c
@@ -40,6 +40,8 @@
 #include <unistd.h>
 
 gchar *template_string;
+gchar *new_diskq_path;
+gchar *persist_file_path;
 gboolean display_version;
 gboolean debug_flag;
 gboolean verbose_flag;
@@ -55,6 +57,19 @@ static GOptionEntry cat_options[] =
 
 static GOptionEntry info_options[] =
 {
+  { NULL, 0, 0, G_OPTION_ARG_NONE, NULL, NULL }
+};
+
+static GOptionEntry relocate_options[] =
+{
+  {
+    "new_path", 'n', 0, G_OPTION_ARG_STRING, &new_diskq_path,
+    "New path for diskq file", "<new_path>"
+  },
+  {
+    "persist", 'p', 0, G_OPTION_ARG_STRING, &persist_file_path,
+    "syslog-ng persist file", "<persist>"
+  },
   { NULL, 0, 0, G_OPTION_ARG_NONE, NULL, NULL }
 };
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

jszigetvari:
Cool stuff, I like it! (Although I am not a member of the strictly viewed dev team, so I don't feel that would be entitled to approve this.)

modules/diskq/dqtool.c
```diff
@@ -241,12 +251,292 @@ _relocate_validate_options(void)
...
 
+  if (!g_threads_got_initialized)
+    {
+      g_thread_init(NULL);
+    }
+
+  main_thread_handle = get_thread_id();
+
+  PersistState *state = persist_state_new(persist_file_path);
+  persist_state_start_edit(state);
```
Kokan: Please handle if this function return with false.

modules/diskq/dqtool.c
```diff
@@ -241,12 +251,292 @@ _relocate_validate_options(void)
...
+  return TRUE;
+}
+
+static gboolean
+_mmap_content_from_fd(gint fd, gchar **content, gsize *content_length)
+{
+  if (!_get_content_length_by_fd(fd, content_length))
+    return FALSE;
+
+  *content = mmap(NULL, *content_length, PROT_READ, MAP_PRIVATE, fd, 0);
```
Kokan: Mapping a diskq file twice could be troublesome. Why did you choose to use `mmap + memcpy` instead of traditional `read + write` from a file ?

furiel: I feel some risk here. Disk queue files might be large. Can we run into trouble mmaping gigabytes of ram? (I am thinking about ulimit).

modules/diskq/dqtool.c
```diff
@@ -241,12 +251,292 @@ _relocate_validate_options(void)
...
+  gint rename_res = rename(from, to);
+
+  switch (rename_res)
+    {
+    case 0:
+      return TRUE;
+    case EXDEV:
+      return _move_file_between_different_filesystems(from, to);
+    default:
+      return FALSE;
```
Kokan: [optional] might help to write the cause of the error.

modules/diskq/dqtool.c
```diff
@@ -241,12 +251,292 @@ _relocate_validate_options(void)
...
+static gboolean
+_move_file(const gchar *from, const gchar *to)
+{
+  gint rename_res = rename(from, to);
+
+  switch (rename_res)
+    {
+    case 0:
+      return TRUE;
+    case EXDEV:
```
furiel: `EXDEV` is and errno value. `rename_res` is still -1 in that case. (switch-case on wrong variable).

modules/diskq/dqtool.c
```diff
@@ -241,12 +251,292 @@ _relocate_validate_options(void)
...
+    {
+      fprintf(stderr, "File reading error: %s\n", filename);
+      fclose(f);
+      return FALSE;
+    }
+
+  fclose(f);
+
+  idbuf[4] = '\0';
+  if (reliable && strcmp(idbuf, "SLRQ") != 0)
```
furiel: fread can return with less than 4, in that case some bytes of `idbuf` is undefined. It is better memset `idbuf` to zero.

modules/diskq/dqtool.c
```diff
@@ -241,12 +251,292 @@ _relocate_validate_options(void)
...
+  if (!is_file_regular(value))
+    return FALSE;
+
+  return _file_is_diskq(value);
+}
+
+static gchar *
+_append_filename_to_dirname(const gchar *dirname, const gchar *filename)
+{
+  if (!dirname || !filename)
```
furiel: `g_build_filename` could work?

modules/diskq/dqtool.c
```diff
@@ -241,12 +251,292 @@ _relocate_validate_options(void)
...
+      gchar *relocated_qfile = _append_filename_to_dirname(new_diskq_path, basename(qfile));
+      if (!relocated_qfile)
+        {
+          fprintf(stderr, "Invalid path. new_diskq_dir: %s, qfile: %s\n", new_diskq_path, qfile);
+          return;
+        }
+
+      if (_move_file(qfile, relocated_qfile))
+        {
+          printf("new qfile_path: %s\n", relocated_qfile);
```
furiel: As this is just a diagnostics log, it is better send to stderr.

modules/diskq/dqtool.c
```diff
@@ -177,6 +203,343 @@ dqtool_info(int argc, char *argv[])
...
+    {
+      fprintf(stderr, "Error: %s\n", strerror(errno));
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gint
+_open_fd_for_writing(const gchar *fname)
```
furiel: [Optional]: The copy related functions clutter this file a little bit. I suggest adding them to a separate c file.

modules/diskq/dqtool.c
lbudai: ...or just check if `fread` returned 4

modules/diskq/dqtool.c
lbudai: `mmap` is an easy&good alternative for fread/fwrite, but well.. if someone limiting the address space (with `ulimit -v`) then this won't work (@Kokan : but it works even when you have a small amount of memory... so if you have 3GB of RAM, you can easily load a 6GB diskq file - and use a 12GB virt. address space).

`g_file_copy`: that uses gio (and currently we don't link to gio lib)
(or, `g_file_move`)

So if you don't like the `mmap` based copy, I can re-implement it to fread/fwrite, but won't/cannot use `g_file_copy`.

modules/diskq/dqtool.c
lbudai: thx!

kira-syslogng:
Build SUCCESS

modules/diskq/dqtool.c
```diff
@@ -241,12 +251,292 @@ _relocate_validate_options(void)
...
+    {
+      fprintf(stderr, "mmap failed, reason: %s\n", strerror(errno));
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
+_mmap_copy_to_fd(gint fd, gchar *content, gsize content_length)
```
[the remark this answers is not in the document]
furiel: As discussed IRL, g_file_copy() cannot work as we do not link to GIO. Not worth the effort.

kira-syslogng:
Build SUCCESS

faxm0dem:
I'm sorry I didn't take the time to test this 🙏

PR #2775, opened 2019-06-11
WIP: Dbld kira improvements

bazsi:
This branch builds on top of #2724 and adds a couple of further improvements.

PR #2857, opened 2019-08-03
Dbld gradle caching

bazsi:
This branch has been split off from #2775 and does the following:

* cleans up gradle argument passing from Makefiles (but not in CMakeLists.txt, help appreciated there)
* enables build-caching (e.g. --build-cache option)
* enables gradle cache-reuse in dbld builds, e.g. the gradle cache dir is stored `/dbld/build/gradle-home`

This speeds up java builds to a point I found them bearable without `--disable-java`, also it speeds up CI builds if the dbld environment is still on the host (e.g. jenkins, but not travis where this is not persisted).

### DIFF
--- a/modules/java-modules/Makefile.am
+++ b/modules/java-modules/Makefile.am
@@ -2,7 +2,8 @@ if ENABLE_JAVA_MODULES
 
 
 GRADLE_PROJECT_DIR ?= $(abs_top_srcdir)/modules/java-modules
-GRADLE_PROJECT_CACHE_DIR ?= $(abs_top_builddir)/modules/java-modules/.gradle
+GRADLE_PROJECT_CACHE_DIR_DEFAULT = $(abs_top_builddir)/modules/java-modules/.gradle
+GRADLE_PROJECT_CACHE_DIR ?= $(GRADLE_PROJECT_CACHE_DIR_DEFAULT)
 GRADLE_FLAGS ?=
 GRADLE_ARGS = $(GRADLE_FLAGS) \
 	--project-dir $(GRADLE_PROJECT_DIR) \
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

modules/java-modules/Makefile.am
```diff
@@ -1,22 +1,26 @@
 if ENABLE_JAVA_MODULES
 
-export GRADLE_OPTS
 
 JAVA_MOD_DST_DIR=$(DESTDIR)/$(moduledir)/java-modules
 MOD_JARS=$(shell find $(abs_top_builddir)/modules/java-modules -name '*.jar' -not -path "$(abs_top_builddir)/modules/java-modules/.gradle/*")
 JAVA_MODULES_INSTALLED_JARS=$(shell find $(JAVA_MOD_DST_DIR) -name '*.jar')
-GRADLE_WORKDIR=$(abs_top_builddir)/modules/java-modules/.gradle
+GRADLE_USER_HOME ?= $(abs_top_builddir)/modules/java-modules/.gradle
+GRADLE_FLAGS ?= --build-cache --project-cache-dir $(GRADLE_USER_HOME)
```
furiel: Adding changes requested because `--build-cache` option breaks our installer generation.

The build on centos6 uses @czanik 's repository for installing gradle.
https://copr-be.cloud.fedoraproject.org/results/czanik/syslog-ng316epel6/epel-6-x86_64/
(Note: syslog-ng 3.16).

That installs gradle 2.6. With that version.
```
[root@d66dc9ef07a5 yum.repos.d]# /opt/gradle/bin/gradle --build-cache

Unknown command-line option '--build-cache'.
[...]
```

One option to be is to upgrade gradle in our build system.

However, there might be other distributions that do not ship recent enough gradle to support this option. Which means we potentially break building from source for other users too.

bazsi: I have now pushed the setting of --build-cache into the dbld/rules file, which ensures that a recent enough gradle is present (right now 4.10).

At the same time I've attempted to clean up the variable mess in that Makefile.am and although I still think it could be improved by a better integration with gradle, I am not proficient enough in gradle matters to do that.

In any case that latest iteration should fix the behavior with old gradle versions.

kira-syslogng:
Build SUCCESS

bazsi:
With the latest changes this should be green in travis and was in kira. I would now continue with the rest of the dbld related PRs

kira-syslogng:
Build FAILURE

bazsi:
@kira-syslogng: retest this please

kira-syslogng:
Build SUCCESS

PR #2858, opened 2019-08-03
Dbld image caching

bazsi:
This branch annotates dbld images with the commit ID they were built with and do this both
for locally generated images (e.g. ./dbld/rules image-<whatever>) and dockerhub generated ones.

It also adds a check-image-<whatever> target that rebuilds the image conditionally, e.g. in case the dbld subdirectory has changed since the local copy. This can be used to significantly improve
CI performance if the previous image is still locally available.

### DIFF
--- a/dbld/rules
+++ b/dbld/rules
@@ -141,6 +141,29 @@ pull-image: pull-image-$(DEFAULT_IMAGE)
 pull-image-%:
 	$(DOCKER) pull balabit/syslog-ng-$*
 
+cache-image-%:
+	IMAGE=$$(docker images -q balabit/syslog-ng-$*:latest | head -1);						\
+	if [ "$$IMAGE" = "" ]; then 											\
+		$(MAKE) -f $(DBLD_DIR)/rules pull-image-$*;								\
+	fi;														\
+	IMAGE=$$(docker images -q balabit/syslog-ng-$*:latest | head -1);						\
+	if [ "$$IMAGE" != "" ]; then 											\
+		image_commit=$$(docker inspect  --format '{{ index .Config.Labels "COMMIT"}}' $$IMAGE);			\
+		dbld_changes=$$( 											\
+			(git cat-file -e $${image_commit:-NO_COMMIT_LABEL_IN_DBLD_IMAGE}^{commit} 			\
+				&& git diff $${image_commit} -- dbld | wc -l						\
+				|| echo 1)										\
+		);													\
+	else														\
+		dbld_changes=1; 											\
+	fi;														\
+	if [ "$${dbld_changes}" -gt 0 ]; then									\
+		echo "$IMAGE: triggering rebuild, as dbld directory changed since the build of the image, or the cached image has no COMMIT label, or couldn't fetch from dockerhub" &&	\
+		$(MAKE) -f $(DBLD_DIR)/rules image-$*;									\
+	else 														\
+		echo "DBLD version in image matches, not initiating rebuild...";					\
+	fi
+
 exec: exec-$(DEFAULT_IMAGE)
 exec: EXEC_COMMAND=echo Specify EXEC_COMMAND to do something sensible here
 exec-%: setup
--- a/dbld/rules
+++ b/dbld/rules
@@ -134,7 +134,7 @@ shell-%: setup
 images: $(foreach image,$(IMAGES), image-$(image))
 image: image-$(DEFAULT_IMAGE)
 image-%:
-	cd $(DBLD_DIR)/images/ && $(DOCKER) build --network=host -t balabit/syslog-ng-$* -f $*.dockerfile .
+	cd $(DBLD_DIR)/images/ && $(DOCKER) build --build-arg=OS_PLATFORM=$* --build-arg=COMMIT=$$(git rev-parse --short HEAD) --network=host -t balabit/syslog-ng-$* -f $*.dockerfile .
 
 pull-images: $(foreach image,$(BUILDER_IMAGES), pull-image-$(image))
 pull-image: pull-image-$(DEFAULT_IMAGE)
--- a/dbld/images/centos-7.dockerfile
+++ b/dbld/images/centos-7.dockerfile
@@ -2,7 +2,9 @@ FROM centos:7
 LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>, Balazs Scheidler <balazs.scheidler@oneidentity.com>"
 
 ARG OS_PLATFORM
+ARG COMMIT
 ENV OS_PLATFORM ${OS_PLATFORM}
+LABEL COMMIT=${COMMIT}
 
 COPY helpers/* /helpers/
 
--- a/dbld/images/centos-7.dockerfile
+++ b/dbld/images/centos-7.dockerfile
@@ -1,6 +1,8 @@
 FROM centos:7
 LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>, Balazs Scheidler <balazs.scheidler@oneidentity.com>"
-ENV OS_PLATFORM centos-7
+
+ARG OS_PLATFORM
+ENV OS_PLATFORM ${OS_PLATFORM}
 
 COPY helpers/* /helpers/
 
--- a/dbld/images/debian-buster.dockerfile
+++ b/dbld/images/debian-buster.dockerfile
@@ -1,10 +1,12 @@
 FROM debian:buster
 LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>, Balazs Scheidler <balazs.scheidler@oneidentity.com>"
 
+ARG OS_PLATFORM
+ENV OS_PLATFORM ${OS_PLATFORM}
+
 ENV DEBIAN_FRONTEND=noninteractive
 ENV DEBCONF_NONINTERACTIVE_SEEN=true
 ENV LANG C.UTF-8
-ENV OS_PLATFORM debian-buster
 
 COPY helpers/* /helpers/
 
--- a/dbld/images/debian-buster.dockerfile
+++ b/dbld/images/debian-buster.dockerfile
@@ -2,7 +2,9 @@ FROM debian:buster
 LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>, Balazs Scheidler <balazs.scheidler@oneidentity.com>"
 
 ARG OS_PLATFORM
+ARG COMMIT
 ENV OS_PLATFORM ${OS_PLATFORM}
+LABEL COMMIT=${COMMIT}
 
 ENV DEBIAN_FRONTEND=noninteractive
 ENV DEBCONF_NONINTERACTIVE_SEEN=true
--- a/dbld/images/debian-jessie.dockerfile
+++ b/dbld/images/debian-jessie.dockerfile
@@ -1,10 +1,12 @@
 FROM debian:8
 LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>, Balazs Scheidler <balazs.scheidler@oneidentity.com>"
 
+ARG OS_PLATFORM
+ENV OS_PLATFORM ${OS_PLATFORM}
+
 ENV DEBIAN_FRONTEND=noninteractive
 ENV DEBCONF_NONINTERACTIVE_SEEN=true
 ENV LANG C.UTF-8
-ENV OS_PLATFORM debian-jessie
 
 COPY helpers/* /helpers/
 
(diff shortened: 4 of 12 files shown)


kira-syslogng:
Build SUCCESS

dbld/rules
```diff
@@ -134,12 +134,19 @@ shell-%: setup
 images: $(foreach image,$(IMAGES), image-$(image))
 image: image-$(DEFAULT_IMAGE)
 image-%:
-	cd $(DBLD_DIR)/images/ && $(DOCKER) build --network=host -t balabit/syslog-ng-$* -f $*.dockerfile .
+	cd $(DBLD_DIR)/images/ && $(DOCKER) build --build-arg=OS_PLATFORM=$* --build-arg=COMMIT=$$(git rev-parse --short HEAD) --network=host -t balabit/syslog-ng-$* -f $*.dockerfile .
+
+check-image-%:
+	@image_commit=$$(docker inspect  --format '{{ index .Config.Labels "COMMIT"}}' balabit/syslog-ng-$*);		\
+	dbld_changed=$$(git diff $${image_commit:-NO_COMMIT_LABEL_IN_DBLD_IMAGE} -- dbld | wc -l);		\
```
Kokan: Although it "handles" the case when there is no commit label, but `| wc -l` is going to result in `0` thus no change, which is kinda dangerous.
My biggest issue with it that it kinda assumes that the commit is in fact in the repository. Which can be untrue in numerous case:
* because of github workflow one fork could miss commits from upstream other forks
* for example travis does shallow copy of repositories (so it could miss commit even if it is in the repository)

In those case it is going to report no change was found.

My alternative suggestion is to use the commit hash of the `dbld` image when the image is created, and simple compare those hash values if different there was probably a change. But even if it is a false report it is better than silently ignore a possible change.

You could use: `git rev-list --abbrev=7 --abbrev-commit -1 HEAD -- dbld`

dbld/rules
```diff
@@ -136,11 +136,6 @@ image: image-$(DEFAULT_IMAGE)
 image-%:
 	cd $(DBLD_DIR)/images/ && $(DOCKER) build --network=host -t balabit/syslog-ng-$* -f $*.dockerfile .
 
-pull-images: $(foreach image,$(BUILDER_IMAGES), pull-image-$(image))
-pull-image: pull-image-$(DEFAULT_IMAGE)
```
furiel: These targets seemed useful. Is it necessary to remove them?

Maybe this is a rebase problem. For example, I miss setting --build-arg=OS_PLATFORM in this patch, on the other hand this change is not necessary.

bazsi:
This is a rebase issue, I will fix it up.

dbld/rules
bazsi: I actually prefer storing the GIT commit instead of an obscure object within the repository, as the COMMIT is usable for use-cases outside of this caching.

I don't see how the github workflow would cause missing commits. dockerhub will always build from master, so any branch forking off of master should get the commit that was used to generate the image (except if it reverted it, but still the commit would be available and the diff would be non-empty).

On the other hand shallow copies could cause problems, I agree. That could be mitigated by using a larger depth value (travis uses 50 by default but that can be customized). How is this handled by the `scripts/version.sh` script?

szemere:
Hello,
1. I have found the missing `--build-arg` in: 20e13cfcaf8c376cb261869f576f56941ba02958
2. Could you please provide some example of how to use this new target? (I get various errors - most probably because none of my images are labelled.)

bazsi:
I have fixed the rebase issue, and will continue with the remaining open comments next.

kira-syslogng:
Build FAILURE

dbld/rules
Kokan: What kind of other use-cases do you have in mind ?

If you have like the `original` and `fork_a` and `fork_b`, both `fork_a`. A docker image is built on `fork_a` with commit hash `forkAhash` and also a build made on `fork_b` via `forkBhash`. Let's also assume that both fork modifies `dbld` related functionality. Neither `fork_a` nor `fork_b` is going to have the same commit, as they are not integrated into the master.

The builder caches the `fork_a` (as it was the first), and now builds `fork_b`, which does not have `forkAhash` thus this code implies no change in the `dbld` using the old img. While in fact there was change!

(maybe I am missing some point, please correct me if so)

The `scripts/version.sh` handles in a way, that does not handle shallow copy and falls back to the `VERSION` file. (this also falls back to the same original behaviour)

kira-syslogng:
Build SUCCESS

dbld/rules
bazsi: > What kind of other use-cases do you have in mind ?

having the COMMIT available will document exactly which commit was used to generate a specific image, that can possibly be corelated back to a jenkins/travis/dockerhub build as they all save the git commit for every build. 

It also gives us a sense of time (as commits do have dates associated with them), whereas if we only annotate using the object id of the directory we lose that. The dbld images should only depend on the contents of the "dbld" directory, still I feel that associating images with git IDs is more natural.

It could possibly make sense to use the dbld object id as an image version in docker (e.g instead of "latest"), which could allow us to use multiple images in our cache. But this would make referencing images less intuitive

> 
> If you have like the `original` and `fork_a` and `fork_b`, both `fork_a`. A docker image is built on `fork_a` with commit hash `forkAhash` and also a build made on `fork_b` via `forkBhash`. Let's also assume that both fork modifies `dbld` related functionality. Neither `fork_a` nor `fork_b` is going to have the same commit, as they are not integrated into the master.
> 
> The builder caches the `fork_a` (as it was the first), and now builds `fork_b`, which does not have `forkAhash` thus this code implies no change in the `dbld` using the old img. While in fact there was change!
> 
> (maybe I am missing some point, please correct me if so)
> 

I see your point now. Let me think a bit.

> The `scripts/version.sh` handles in a way, that does not handle shallow copy and falls back to the `VERSION` file. (this also falls back to the same original behaviour)

bazsi: I misunderstood your idea about the commit ID of the last dbld change. I thought you referred to the object id of the DBLD directory. But as far as I see directories do not even have that (only files have a content id, directories don't).

kira-syslogng:
Build SUCCESS

dbld/rules
Kokan: You can ask for commit hash for a files, and for directories as well(if directory is tracked indirectly via tracking a file inside that directory).
The proposal was to query and use that has (for command see my original post).

Out of the original caching solution and the one I mentioned I would prefer mine. I am okay with different solution, it just have to detect the change in the above written cases.

bazsi: I understand that you can get a commit ID that last changed a specific directory. The question is in what way should we use it?

1) use it for the docker image ID (e.g. instead of `latest` we would have a series of docker images each tagged with the commit that last changed dbld). I have decided against this, as I think there's a lot of temporal state encapsulated in an image. Which means that one `dbld` revision is not the same as another one depending on the time the image was generated (because apt/pip/obs/etc changes). This means that if we used docker revisions, we would need to start maintaining those revisions, and regenerate them from time to time. The difficulty comes from knowing how far we need to go back maintaining these images and how to remove stale ones.

2) use it for a label within the specific image (e.g. we are using `latest`, but we'd have a DBLD_COMMIT label within the image). I don't see how this helps us over what I have in there currently. The change detection might be simpler (e.g. just compare two commit IDs), but we would need the same history anyway (to come up with our `dbld` commit id we need to use rev-list, that again needs history).

So my rationale has been for just using COMMIT
1) I don't want to maintain a series of dbld builds (e.g. latest only)
2) COMMIT is enough for change detection and it imposes the same requirements as DBLD_COMMIT would (e.g. we would need git history on the build machine that should go back to the latest image built by dockerhub).
3) I don't think dbld would change that often, but I expect dockerhub to build daily from master, and I would like to make it trivial which commit dockerhub was using to build an image.

With those in the latest revision, I think I've handled all cases where local history is not available (e.g. if the commit is not there I _always_ trigger a rebuild of the image). This is not a perfect caching decision, for that we would need the entire history, but should work in most of the cases, speeding up the builds considerably as long as the previous build image is available on the build host.

bazsi:
>     2. Could you please provide some example of how to use this new target? (I get various errors - most probably because none of my images are labelled.)

The point is to use this on jenkins slaves to conditionally rebuild the image, but only if needed. With those our `distro packages` build should be much faster. 

This is what `./dbld/rules cache-image-ubuntu-xenial` would  do:
1) check if we have any build image for ubuntu-xenial locally
2) if there's nothing, it would perform a `docker pull` from dockerhub
3) check if the local version is "suitable" for building the current source tree (the "suitability" question is what we've discussed in the previous comments).
4) if it is suitable, do nothing, if it's not rebuild it.

With those you can be sure that a dbld image is available by the time the target finishes, and in ideal conditions this would only take less than a second.

Ideal conditions in the order of "ideality":
1) local image exists and is suitable (e.g. no dbld changes)
2) docker pull gets us a good enough version, thus no need to rebuild
3) we originally had a local image, but it is not suitable, so we need to rebuild
4) we didn't have an image, so we pulled one, but is not suitable and we also need to rebuild

So worst case is that we pull AND rebuild. This could be improved by first checking that the version on dockerhub is suitable before doing the pull, but I haven't implemented that.

kira-syslogng:
Build SUCCESS

dbld/rules
Kokan: > if the commit is not there I always trigger a rebuild of the image

That's the important thing I wanted, and did not care what metadata storage achieves it (only tried to provide one that I know would work, and never forcing to use that).
I'll check the code.

szemere: First things first. I do not have a proposal yet, and I don't know a better solution. Sorry for that.
My concerns/how do I see it:

With image caching we introduced the plain old caching problem. We don't know what time we should refresh the cache.
- Currently we do not follow / mark the versions of the upstream dependencies.
- Using COMMIT instead of DBLD_COMMIT "just" increase the refreshing frequency, but still don't make any logical connection with the content of the image. (i.e.: a simple date check will do the same)

Totally agree with Bazsi's 1. point.
> I think there's a lot of temporal state encapsulated in an image. Which means that one dbld revision is not the same as another one depending on the time the image was generated (because apt/pip/obs/etc changes)

Only my conclusion differs. In theory: Two docker image should be identical if DBLD_COMMIT hasn't change, regardless of the master commit. And neither one of them is suitable to track the "temporal state" changes.

bazsi:
On Thu, Aug 22, 2019, 11:27 László Szemere <notifications@github.com> wrote:

> *@szemere* commented on this pull request.
>
> In dbld/rules
> <https://github.com/balabit/syslog-ng/pull/2858#discussion_r316580855>:
>
> > @@ -134,12 +134,19 @@ shell-%: setup
>  images: $(foreach image,$(IMAGES), image-$(image))
>  image: image-$(DEFAULT_IMAGE)
>  image-%:
> -	cd $(DBLD_DIR)/images/ && $(DOCKER) build --network=host -t balabit/syslog-ng-$* -f $*.dockerfile .
> +	cd $(DBLD_DIR)/images/ && $(DOCKER) build --build-arg=OS_PLATFORM=$* --build-arg=COMMIT=$$(git rev-parse --short HEAD) --network=host -t balabit/syslog-ng-$* -f $*.dockerfile .
> +
> +check-image-%:
> +	@image_commit=$$(docker inspect  --format '{{ index .Config.Labels "COMMIT"}}' balabit/syslog-ng-$*);		\
> +	dbld_changed=$$(git diff $${image_commit:-NO_COMMIT_LABEL_IN_DBLD_IMAGE} -- dbld | wc -l);		\
>
> First things first. I do not have a proposal yet, and I don't know a
> better solution. Sorry for that.
> My concerns/how do I see it:
>
> With image caching we introduced the plain old caching problem. We don't
> know what time we should refresh the cache.
>
>    - Currently we do not follow / mark the versions of the upstream
>    dependencies.
>    - Using COMMIT instead of DBLD_COMMIT "just" increase the refreshing
>    frequency, but still don't make any logical connection with the content of
>    the image. (i.e.: a simple date check will do the same)
>
>
>
> Totally agree with Bazsi's 1. point.
>
> I think there's a lot of temporal state encapsulated in an image. Which
> means that one dbld revision is not the same as another one depending on
> the time the image was generated (because apt/pip/obs/etc changes)
>
> Only my conclusion differs. In theory: Two docker image should be
> identical if DBLD_COMMIT hasn't change, regardless of the master commit.
> And neither one of them is suitable to track the "temporal state" changes.
>
Our current set of requirements don't need precise tracking of
dependencies, as our build products don't include those (well except the
docker image that is). As our users install the deb/rpm on their systems,
their specific dependencies may be different from what we built syslog-ng
with.

And this is not a problem as long as those dependencies keep binary
compatibility, which they do in a specific distro version.

So again, with dbld and with the packages it generates, we don't really
care the exact dependency version as long as the runtime package of those
dependencies are available and they keep binary compatibility.

This may become different with production docker images and with installers
that bundle dependencies. But I argue they should use a different logic for
building and those are in an entirely different layer on top of what the
syslog-ng OSE source tarball provides.

bazsi:
I think I have answered all open concerns, so this should be ready to go in. I have two branches building on top of this one, so it'd be great if we could expedite this.

thanks.

bazsi:
Any updates on when this could go in? I have works in progress in other branches but  dont want to  submit them until the dbld blockage clears up.

bazsi:
Thank you!

PR #2863, opened 2019-08-06
filter modifier framework

Kokan:
The #2833 is split into multiple PR, and this is one of them.
The goal is here to introduce a filter, that provides a programmable internal interface (libsyslog-ng), that allows to manipulate, interact with the `FilterExprNode` tree. Both filter pipe element, and other filtering functionality is provided via the `FitlerExprNode` tree, that nodes either a filter or a logical operator.

There are two examples for the framework usage:
* A debug-like feature that prints out the `FilterExprNode` (this does not modify the tree itself)
* Unit test that implements a dummy replace all `FilterExprNode` to provide example for manipulating the tree itself.

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -272,6 +272,7 @@ lib_libsyslog_ng_la_SOURCES		= \
 	$(transport_sources)		\
 	$(logproto_sources)		\
 	$(filter_sources)		\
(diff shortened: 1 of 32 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #2865, opened 2019-08-06
WIP: filter: combine multiple or filters via merging regex's

Kokan:
The #2833 is split into multiple PR, and this is one of them. This PR depends on #2863 and as such contains patches from #2863 , the first commit for this branch starts here: https://github.com/balabit/syslog-ng/commit/08aebed13c2ec5445ae5bdfa6aeaceca11b47525

This optimiser algorithm provides a way to combine multiple filter connected with `or` operator into one, that is assumed have better performance. (performance data is going to be provided here)
The following: `message("pattern1") or message("pattern2") or ... or message("patternK");` could be replaced with the `message("pattern1|pattern2|...|patternK");`.

- [ ] Share performance data

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -272,6 +272,7 @@ lib_libsyslog_ng_la_SOURCES		= \
 	$(transport_sources)		\
 	$(logproto_sources)		\
 	$(filter_sources)		\
+	$(optimizer_sources)		\
 	$(parser_sources)		\
 	$(rewrite_sources)		\
 	$(template_sources)		\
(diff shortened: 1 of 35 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
The performance measurement indicated that some cases this modification could even reduce performance, which is not acceptable. There are plans to improve upon, but as it requires additional work. I'll close this PR until it is reviewable again.

PR #2866, opened 2019-08-07
python test: use malloc_debug for python unit tests

Kokan:
As part of the work to push #2178 python had the most leak from all of them.
After some investigation it turned out that from syslog-ng point of view the memory management is fine, and in fact the leak is reported because of `pymalloc` (for details check out cpython valgrind guide).

Since python 3.6 it is possible to change the allocator via an environment varialbe (`PYTHONMALLOC`). As of now there is: `pymalloc`, `malloc` and those with a combination of `_debug` suffix.
The guid suggest to use `malloc` for valgrind, and there is no big surprise that it also helps with address sanitizer as well.

This PR suggest to use `malloc_debug` as that also additionally reports issues, I could find one misuse of the `GIL`.

Note: this is not going to have effect on Travis or Kira as they are using older version of python. But this still could be a good documentation, and useful locally.

### DIFF
--- a/modules/python/tests/CMakeLists.txt
+++ b/modules/python/tests/CMakeLists.txt
@@ -3,7 +3,11 @@ add_unit_test(LIBTEST CRITERION
   INCLUDES "${PYTHON_INCLUDE_DIR}" "${PYTHON_INCLUDE_DIRS}"
   DEPENDS syslogformat mod-python "${PYTHON_LIBRARIES}")
 
+set_property(TEST test_python_logmsg APPEND PROPERTY ENVIRONMENT "PYTHONMALLOC=malloc_debug")
+
 add_unit_test(LIBTEST CRITERION
   TARGET test_python_template
   INCLUDES "${PYTHON_INCLUDE_DIR}" "${PYTHON_INCLUDE_DIRS}"
   DEPENDS mod-python "${PYTHON_LIBRARIES}" syslogformat)
+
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

modules/python/tests/CMakeLists.txt
```diff
@@ -3,7 +3,11 @@ add_unit_test(LIBTEST CRITERION
...
   DEPENDS syslogformat mod-python "${PYTHON_LIBRARIES}")
 
+set_property(TEST test_python_logmsg APPEND PROPERTY ENVIRONMENT "PYTHONMALLOC=malloc_debug")
+
 add_unit_test(LIBTEST CRITERION
   TARGET test_python_template
   INCLUDES "${PYTHON_INCLUDE_DIR}" "${PYTHON_INCLUDE_DIRS}"
   DEPENDS mod-python "${PYTHON_LIBRARIES}" syslogformat)
+
+set_property(TEST test_python_template APPEND PROPERTY ENVIRONMENT "PYTHONMALLOC=malloc_debug")
```
bazsi: can you also add this to Makefile.am?

just assign PYTHONMALLOC to the same value and then "export PYTHONMALLOC" in Makefile.am

Kokan: Like this https://github.com/balabit/syslog-ng/pull/2866/commits/31d7c93da0653ac41330d4649268cc3c52906831#diff-68c196e53bce5b3146d2b805debf05dcR1 ?

kira-syslogng:
Build SUCCESS

PR #2867, opened 2019-08-07
light: syslogngcli include exit_value

Kokan:
One of my branch failed because of this exception, this just add a very minor information that might help in case of this exception.


kira-syslogng:
Build SUCCESS

PR #2868, opened 2019-08-07
few small fries leak fixed

Kokan:
The `recv_time_zone` is an actual not UT leak, but well only if configured.

### DIFF
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -602,6 +602,7 @@ cfg_free(GlobalConfig *self)
 
   if (self->bad_hostname_compiled)
     regfree(&self->bad_hostname);
+  g_free(self->recv_time_zone);
   if (self->source_mangle_callback_list)
     g_list_free(self->source_mangle_callback_list);
   g_free(self->bad_hostname_re);
--- a/lib/secret-storage/tests/test_secret_storage.c
+++ b/lib/secret-storage/tests/test_secret_storage.c
@@ -286,6 +286,7 @@ Test(secretstorage, test_rlimit)
     {
       cr_assert_not(secret_storage_store_string(key_fmt, "value"), "offending_key: %s", key_fmt);
     }
+  g_free(key_fmt);
 }
 
 static void
--- a/lib/logmsg/nvhandle-descriptors.c
+++ b/lib/logmsg/nvhandle-descriptors.c
@@ -39,6 +39,11 @@ nvhandle_desc_array_new(guint reserved_size)
 void
 nvhandle_desc_array_free(NVHandleDescArray *self)
 {
+  for (gint i=0; i < self->len; ++i)
+    {
+      NVHandleDesc *handle = (NVHandleDesc *)&self->data[i];
+      nvhandle_desc_free(handle);
+    }
   g_free(self->data);
   g_ptr_array_free(self->old_buffers, TRUE);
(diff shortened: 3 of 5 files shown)


lib/logmsg/tests/test_nvhandle_desc_array.c
```diff
@@ -28,10 +28,10 @@
 
 Test(nvhandle_desc_array, add_elements)
 {
-  NVHandleDesc desc0 = { .name = g_strdup("test0"), .flags = 0, .name_len = strlen("test0")};
-  NVHandleDesc desc1 = { .name = g_strdup("test1"), .flags = 1, .name_len = strlen("test1")};
-  NVHandleDesc desc2 = { .name = g_strdup("test2"), .flags = 2, .name_len = strlen("test2")};
-  NVHandleDesc desc3 = { .name = g_strdup("test3"), .flags = 3, .name_len = strlen("test3")};
+  NVHandleDesc desc0 = { .name = "test0", .flags = 0, .name_len = strlen("test0")};
+  NVHandleDesc desc1 = { .name = "test1", .flags = 1, .name_len = strlen("test1")};
```
furiel: In the production code:

```
  stored.flags = 0;
  stored.name_len = len;
  stored.name = g_strdup(name);
  nvhandle_desc_array_append(self->names, &stored);
```
means we strdup the entry name, similarly to the original version of the test. Doesn't this mean that the production code still leaks? Maybe g_free is missing from `nvhandle_desc_array_free`?

kira-syslogng:
Build SUCCESS

lib/secret-storage/tests/test_secret_storage.c
```diff
@@ -286,6 +286,7 @@ Test(secretstorage, test_rlimit)
     {
       cr_assert_not(secret_storage_store_string(key_fmt, "value"), "offending_key: %s", key_fmt);
     }
+  g_free(key_fmt);
```
furiel: Good catch. I am thinking maybe there is no need to have key_fmt on the heap at all, and could be allocated on the stack, removing a little complexity from the unit test.

lib/logmsg/tests/test_nvhandle_desc_array.c
Kokan: yes, you are right. i'll mark it as a wip and solve this.

Kokan: I left some explanation about the solution in the comment: https://github.com/balabit/syslog-ng/pull/2868/commits/7b5ab5077516acce86598c51eb7a97fc07808d41

kira-syslogng:
Build SUCCESS

Kokan:
@furiel thanks for the review, and sorry but style checked failed so I had to update :(

kira-syslogng:
Build SUCCESS

lib/cfg.c
```diff
@@ -602,6 +602,8 @@ cfg_free(GlobalConfig *self)
 
   if (self->bad_hostname_compiled)
     regfree(&self->bad_hostname);
+  if (self->recv_time_zone)
```
lbudai: `g_free` handles `NULL`

Kokan: done.

kira-syslogng:
Build SUCCESS

PR #2870, opened 2019-08-07
filter: small refactors

Kokan:
Spring cleanup from my disk.
This commit is somewhat questionable: https://github.com/balabit/syslog-ng/commit/06ce5030352c6732710331861bf20a7bad74cb33

### DIFF
--- a/lib/filter/filter-re.c
+++ b/lib/filter/filter-re.c
@@ -179,8 +179,10 @@ filter_match_set_template_ref(FilterExprNode *s, LogTemplate *template)
 }
 
 static gboolean
-filter_match_eval_against_program_pid_msg(FilterMatch *self, LogMessage **msgs, gint num_msg)
+filter_match_eval_against_program_pid_msg(FilterExprNode *s, LogMessage **msgs, gint num_msg)
 {
+  FilterMatch *self = (FilterMatch *) s;
+
   const gchar *pid;
   gssize pid_len;
   gchar *str;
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

lib/filter/filter-re.c
```diff
@@ -179,8 +179,10 @@ filter_match_set_template_ref(FilterExprNode *s, LogTemplate *template)
 }
 
 static gboolean
-filter_match_eval_against_program_pid_msg(FilterMatch *self, LogMessage **msgs, gint num_msg)
+filter_match_eval_against_program_pid_msg(FilterExprNode *s, LogMessage **msgs, gint num_msg)
```
furiel: For other reviewers: the motivation is these functions will be used as virtual fuction

lib/filter/filter-re.c
```diff
@@ -244,13 +244,15 @@ filter_match_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg)
   FilterMatch *self = (FilterMatch *) s;
 
   if (G_LIKELY(self->super.value_handle))
-    return filter_re_eval(s, msgs, num_msg);
+    self->super.super.eval = filter_re_eval;
```
furiel: I understand we save a dispatch function during runtime, so that's a +1. Also, this makes the `G_LIKELY` unnecessary, so that's again a +1.

On the other hand, it is harder to reason about correctness. For example we need to think that the lazy update works with multi threads. That's easy to reason though. Still, I tend to think the original is easier to maintain.

Is it possible to calculate the dispatch earlier? Say in init phase? That would be the best.

Kokan: yes, it can be.

alltilla:
Reviewed the code, and could not find any problems :) When furiel's comment gets fixed, you can expect an approve from me!

lib/filter/filter-re.c
Kokan: I did not rewrite the branch, but pushed a last commit doing just that.

kira-syslogng:
Build SUCCESS

Kokan:

kira-syslogng:
Build SUCCESS

Kokan:

kira-syslogng:
Build SUCCESS

Kokan:

kira-syslogng:
Build SUCCESS

PR #2872, opened 2019-08-09
Dqtool assign to persist

lbudai:
```
dqtool: add assign dqfile to persist file feature
    
    example:
    bin/dqtool assign -p var/syslog-ng.persist -n "afsocket_dd_qfile(stream,localhost:15554)"
    /tmp/syslog-ng-dq/syslog-ng-00000.rqf
    
    When only a filename is given for diskq, it will be appended to the current working
    dir.
    
    One bad thing: user need to figure out the correct persist name.
```

How it works?
When you know what will the persist name for the diskq file but you want to manually assign an existing queue file to your destination, then with this feature you can set the queue file manually (even when you don't have an entry for the diskq file in the persist).

How can you try it?
* Create a simple network destination, without diskbuffer.

```
destination d_net {
  network(
    "localhost" port(15554)
    transport(tcp)
  );
};
```

* starts syslog-ng, it creates the persist file

* stop syslog-ng, modify the config and set diskbuffer for the network dst

```
destination d_net {
  network(
    "localhost" port(15554)
    transport(tcp)
    disk-buffer(
      mem-buf-size(10M)
      disk-buf-size(16M)
      reliable(yes)
      dir("/tmp/syslog-ng-dq")
    )
  );
};
```
After you set the diskq file :

```
    bin/dqtool assign -p var/syslog-ng.persist -n "afsocket_dd_qfile(stream,localhost:15554)"
    /tmp/syslog-ng-dq/syslog-ng-00000.rqf
```

you can start syslog-ng and it will use the queue file you set.

Questions:
 * are you okay with adding `g_canonicalize_filename`  to `compat/glib`?
 * what do you think about this feature? (can you use it for something, @jszigetvari ? )


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

jszigetvari:
@lbudai If I understand right, this will override the existing disk buffer path in the persist file, if a matching entry exists, and will create a new one, if it doesn't exist.
Also, this will only work with a stopped syslog-ng (obviously). That's not a problem for me, I'm just saying.

All in all, I think it will be a useful feature for us. (But please be sure to include the above usage example in --help, because if the persist file doesn't yet contain any disk-queue mappings, then it will be hard to conjure up the correct persist name for the destination without help.)

lbudai:
@jszigetvari : thanks for the feedback.
could you check the last commit? (that contains some help: are you okay with that?)

kira-syslogng:
Build SUCCESS

modules/diskq/dqtool.c
```diff
@@ -526,9 +531,29 @@ _assign_validate_options(void)
...
+static void
+_assign_print_help(void)
+{
+  fprintf(stderr, "example:"
+          "   bin/dqtool assign -p var/syslog-ng.persist -n \"afsocket_dd_qfile(stream,localhost:15554)\""
+          "                    /tmp/syslog-ng-dq/syslog-ng-00000.rqf\n\n"
+          "When only a filename is given for diskq, it will be appended to the current working dir.\n"
+          "One bad thing: user need to figure out the correct persist name.\n\n");
+  fprintf(stderr, "How it works?\n"
+          "When you know what will the persist name for the diskq file but you want to manually assign"
```
jszigetvari: "When you know what the persist name for the diskq file is, and you want to manually..."

modules/diskq/dqtool.c
```diff
@@ -526,9 +531,29 @@ _assign_validate_options(void)
...
+          " an existing queue file to your destination, then with this feature you can set the queue file"
+          " manually (even when you don't have an entry for the diskq file in the persist).\n");
```
jszigetvari: (even if you don't have an entry for the disk-queue file in the persist file)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/diskq/dqtool.c
```diff
@@ -506,6 +520,50 @@ dqtool_relocate(int argc, char *argv[])
   return 0;
 }
 
+static gboolean
+_assign_validate_options(void)
+{
+  return _validate_persist_file_path(persist_file_path);
```
furiel: In the previous PR you wrote a nice `_validate_diskq_path` that could be used here too.

modules/diskq/dqtool.c
```diff
@@ -506,6 +525,70 @@ dqtool_relocate(int argc, char *argv[])
...
+}
+
+static void
+_assign_print_help(void)
+{
+  fprintf(stderr, "example:"
+          "   bin/dqtool assign -p var/syslog-ng.persist -n \"afsocket_dd_qfile(stream,localhost:15554)\""
+          "                    /tmp/syslog-ng-dq/syslog-ng-00000.rqf\n\n"
```
alltilla: The indenting seems a little odd here. It looks like this in action:
```
$ ./dqtool assign --example
example:   bin/dqtool assign -p var/syslog-ng.persist -n "afsocket_dd_qfile(stream,localhost:15554)"                    /tmp/syslog-ng-dq/syslog-ng-00000.rqf
```
Apart from this formatting inconvenience, it is an approve from me.

modules/diskq/dqtool.c
lbudai: @furiel : not really, as that function is for validating directories, not files

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/diskq/dqtool.c
furiel: Indeed, I wrote wrong function. `_file_is_diskq` is the one that can validate that a correct diskq is passed as parameter. As discussed irl, this can be useful thing to validate.

modules/diskq/dqtool.c
```diff
@@ -583,6 +583,13 @@ dqtool_assign(int argc, char *argv[])
   const gchar *diskq_file = argv[optind];
   gchar *diskq_full_path = g_canonicalize_filename(diskq_file, NULL);
 
+  gchar *old_entry = persist_state_lookup_string(state, assign_persist_name, NULL, NULL);
+  if (old_entry)
+    {
+      fprintf(stderr, "Entry overrided during the assign process. Old entry: %s\n", old_entry);
```
furiel: cosmetic: overridden

lbudai: ah, thx. fixed

kira-syslogng:
Build SUCCESS

PR #2874, opened 2019-08-10
fix leak in stats test

Kokan:


kira-syslogng:
Build SUCCESS

lib/stats/tests/test_stats_cluster.c
```diff
@@ -127,7 +133,14 @@ Test(stats_cluster, test_stats_cluster_key_equal_when_custom_tags_are_the_same)
   StatsClusterKey sc_key2;
   stats_cluster_single_key_set_with_name(&sc_key1, SCS_SOURCE | SCS_FILE, "id", "instance", "name");
   stats_cluster_single_key_set_with_name(&sc_key2, SCS_SOURCE | SCS_FILE, "id", "instance", "name");
+  StatsCluster *sc1 = stats_cluster_new(&sc_key1);
+  StatsCluster *sc2 = stats_cluster_new(&sc_key2);
+
   cr_assert(stats_cluster_key_equal(&sc_key1, &sc_key2), "%s", __FUNCTION__);
```
szemere: Is there any particular reason we can not use the `assert_stats_cluster_equals_and_free` function?

Kokan: this connection is somewhat ugly, but refactoring the whole stats was kinda out of scope for me. so I tried to keep the UT as nice as the stats code :)

Issue #2842, opened 2019-07-22
fix "standalone" config parsing


szemere:
# Issue

Syslog-ng fails to parse the config below with the following error message:

```Error parsing filter expression, filter plugin filter not found OR you may not used double quotes in your filter expression in <string>:1:1-1:9:```

# Configuration
KUDOS for @Kokan 
```
@version: 3.22

filter f1 { message("a"); };

log {
 rewrite { set("$(if message('foo') a b)" value("a")); };
 rewrite { set("$(if filter(f1) a b)" value("a")); };
};
```

<details>
<summary>Since the keyword `filter` occurs many times in the error message. Here is a version where it is replaced, to make the problem more obvious.</summary>

# Issue
```Error parsing filter expression, filter plugin xyz not found OR you may not used double quotes in your filter expression in <string>:1:1-1:9:```

# Configuration
```
@version: 3.22

filter f1 { message("a"); };

log {
 rewrite { set("$(if message('foo') a b)" value("a")); };
 rewrite { set("$(if xyz(f1) a b)" value("a")); };
};
```
</details>

# Reason
The `filter` keyword is unknown in this context because it is declared in the [cfg-parser.c](https://github.com/balabit/syslog-ng/blob/master/lib/cfg-parser.c#L53), and during the evaluation of the template parameter of the rewrite rule we do not derive the keywords from the "global scope".
Thus the `filter` keyword is not recognized by the lexer and it is handled as a simple `LL_IDENTIFIER`, which results in a parsing error eventually.

However the example mentions only the `filter` keyword, there are other affected keywords in the config. (i.e.: `type`, `flags`)

# Possible solutions
## Solution1
Repeat the necessary keywords inside the modules. (Started to implement in #2815) The implementation in that PR is not complete, which points out the issue with this approach. We have to traverse the whole path of possible config options started from global cfg-parser, and repeat ALL the occurring keywords along the path inside the module.
1. We introduce redundancy
3. There is a high chance that someone will forget to update this part of the code. (From the developer point of view, even a unit test can not raise the problem, because it only affects new features IF those features are used INSIDE an another module.)

## Solution2
Create a helper function for parsing config statements, which implicitly inherits all the keywords from the global cfg-parser. This way we never miss to include the needed keywords, and don't have to repeat them later.
This approach has some other benefits:
### We can add some safeguards to our code
[...]

PR #2876, opened 2019-08-12
Add global context to standalone parsers

szemere:
The problem is described in #2842

This patch:
- allows to parse modules with the keywords from the main_parser.
- calls missing init statements, so filters will work properly.


kira-syslogng:
Build SUCCESS

PR #2881, opened 2019-08-15
travis: commit style check use commit range

Kokan:
Moves travis specific code into `.travis.yml` and uses `${TRAVIS_COMMIT_RANGE}` variable configured by travis to specify the commits needs to be tested.

This range contains the commit of the current branch, and commits that are in the upstream (target remote/branch) has and the current branch do not have.

Note: when there is no PR opened just the branch pushed into github, this is empty (and falls back to the script default).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -135,8 +135,11 @@ matrix:
       compiler: gcc
       install:
       before_script:
+      git:
+        submodules: false
+        depth: false
       script:
-        - tests/commits/check.sh
+        - tests/commits/check.sh "${TRAVIS_COMMIT_RANGE}"
 
     - env: B=cmake-submodules-and-pytest
       compiler: gcc
--- a/tests/commits/check.sh
+++ b/tests/commits/check.sh
@@ -23,11 +23,6 @@
 
 commit_range=origin/master..HEAD
 
-if [ -n "$TRAVIS_COMMIT" ]; then
-  git fetch -q origin master:master
-  commit_range="master..$TRAVIS_COMMIT"
-fi
-
 if [ $# -gt 0 ]; then
   commit_range=$1
 fi


kira-syslogng:
Build SUCCESS

Kokan:

kira-syslogng:
Build SUCCESS

Kokan:

kira-syslogng:
Build SUCCESS

.travis.yml
```diff
@@ -136,6 +136,7 @@ matrix:
       install:
       before_script:
       script:
+        - git fetch --deepen 500 #this implies it only works with PR less than 500 commits in it
```
MrAnno: There's
```
git:
  depth: N
```
in Travis. It looks a bit nicer, what do you think?

Kokan: yeah, i just used this to test locally the fix. i'll change as i do not mind either way.

Kokan: well actually I'll use `git.depth: false` that won't have the limit for 500 commit per pr (although still hope it won't come to this ever)

MrAnno: `git.depth false` might cause timeouts, in this case, we should triple-check Travis in the busiest times.

Kokan: now it took 4-5 sec instead of 1-2 (when depth=50), why do you thing it is going to cause any issue ?

MrAnno: That's nice, I didn't know this setting can be set for only one job. I've read a few timeout issues when depth false was applied, but 4-5 secs is nothing.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
Both the current and the proposed version has pitfalls which originates from git not able (or at least I do not know a way) to tell a list of commit which connected to a changes and only the changes on that branch.
As both solution has its own issue, I'll drop the related patches. Still splitting the travis related logic from the script is a good thing, and I plan to keep that at least.

kira-syslogng:
Build SUCCESS

PR #2882, opened 2019-08-15
Version/3.23

lbudai:


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

MrAnno:
I've pushed a commit to fix `Distro packages`.

I think disabling the version check for snapshot builds is the right move, because snapshots should have the correct version, which is not 3.23.1+something in this PR, it's 3.22.1+something.

I think this because this PR is just a version bump, it's not the actual 3.23.1 release. We might want to merge more patches (quick fixes, etc.) before putting the seal on 3.23.1 (before pushing the actual tag), so I've added the `--force-bad-version` flag to debchange, which sounds misleading, but this is what we actually want (IMHO).

packaging/rhel/syslog-ng.spec
```diff
@@ -577,6 +577,9 @@ fi
 %{_datadir}/%{name}/tools/
 
 %changelog
+* Fri Aug  15 2019 Laszlo Budai <laszlo.budai@balabit.com> - 3.23.1-1
```
MrAnno: There is an unnecessary second whitespace between `Aug` and `15`.
Aug 15 should be `Thu`.

kira-syslogng:
Build SUCCESS

lbudai:
I would not block the actual release, so merge this as-is, but I disagree with using `--force-bad-version` flag.

PR #2884, opened 2019-08-16
fix heap usage after free in cfg_run_parser_with_main_context

Kokan:
The cfg_run_parser_with_main_context uses a cfg_run_parser, that frees the lexer.
Therefore pop-ing from the freed lexer is not okay.


kira-syslogng:
Build SUCCESS

PR #2886, opened 2019-08-23
Add contextual data make filters config plugin aware

bazsi:
the primary goal of this PR is this patch:

```
add-contextual-data: fix missing plugins in the filters configuration
    
It was reported by Mark Bonsack when using add-contextual-data() using the
filters() selector that the secondary configuration file misses all plugins
that the primary one has already loaded.
    
This patch makes this secondary config use the same set of plugins as the
primary one, eliminating the issue.
```

The branch also contains a plugin refactor (basically just moving code around and renaming stuff to make that uglyness somewhat nicer).

The fix relies on the concept of "subordinate" configurations, which by their definition share some things with the main config, this time the set of plugins available to it. Later, this could potentially be extended, but honestly I hope the entire filters() mode of add-contextual-data() can be eliminated completely.

PR #2887, opened 2019-08-23
avoid `@requires` loading the plugins themselves

bazsi:
The `@requires` statement should only be a declaration that the specific include file relies on
the availability of a specific plugin. We shouldn't necessarily load it as it might actually be unused.

Also, by loading it we are triggering ugly messages during startup (the kind that some plugins are registered multiple times).

This branch now properly handles the ordering between plugin discovery and the `@requires` keyword. It removes some old cruft.

### DIFF
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -368,7 +368,8 @@ cfg_set_version(GlobalConfig *self, gint version)
   if (cfg_is_config_version_older(self, 0x0300))
     {
       msg_error("ERROR: compatibility with configurations below 3.0 was dropped in " VERSION_3_13
-                ", please update your configuration accordingly");
+                ", please update your configuration accordingly",
+                cfg_format_config_version_tag(self));
       return FALSE;
     }
 
@@ -378,20 +379,23 @@ cfg_set_version(GlobalConfig *self, gint version)
                   "Please update it to use the " VERSION_CURRENT " format at your time of convenience. "
                   "To upgrade the configuration, please review the warnings about incompatible changes printed "
                   "by syslog-ng, and once completed change the @version header at the top of the configuration "
-                  "file.");
+                  "file",
+                  cfg_format_config_version_tag(self));
     }
   else if (version_convert_from_user(self->user_version) > VERSION_VALUE)
     {
       msg_warning("WARNING: Configuration file format is newer than the current version, please specify the "
                   "current version number ("  VERSION_CURRENT_VER_ONLY ") in the @version directive. "
-                  "syslog-ng will operate at its highest supported version in this mode");
+                  "syslog-ng will operate at its highest supported version in this mode",
+                  cfg_format_config_version_tag(self));
       self->user_version = VERSION_VALUE;
     }
 
   if (cfg_is_config_version_older(self, 0x0303))
     {
       msg_warning("WARNING: global: the default value of log_fifo_size() has changed to 10000 in " VERSION_3_3
-                  " to reflect log_iw_size() changes for tcp()/udp() window size changes");
+                  " to reflect log_iw_size() changes for tcp()/udp() window size changes",
+                  cfg_format_config_version_tag(self));
     }
   return TRUE;
 }
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -355,10 +355,16 @@ cfg_deinit(GlobalConfig *cfg)
   return cfg_tree_stop(&cfg->tree);
 }
 
+void
+cfg_set_version_without_validation(GlobalConfig *self, gint version)
+{
+  self->user_version = version;
+}
+
 gboolean
 cfg_set_version(GlobalConfig *self, gint version)
 {
-  self->user_version = version;
+  cfg_set_version_without_validation(self, version);
   if (cfg_is_config_version_older(self, 0x0300))
     {
       msg_error("ERROR: compatibility with configurations below 3.0 was dropped in " VERSION_3_13
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -364,6 +364,13 @@ cfg_set_version_without_validation(GlobalConfig *self, gint version)
 gboolean
 cfg_set_version(GlobalConfig *self, gint version)
 {
+  if (self->user_version != 0)
+    {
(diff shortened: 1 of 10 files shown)


kira-syslogng:
Build FAILURE

bazsi:
I have since rebased this on top of master, so it doesn't implicitly depend on #2886 and can be merged independently.

kira-syslogng:
Build SUCCESS

lib/pragma-grammar.ym
```diff
@@ -123,7 +123,7 @@ module_stmt
 requires_stmt
 	: KW_REQUIRES string
           {
-            if (!plugin_load_module(&configuration->plugin_context, $2, NULL))
+            if (!plugin_is_module_available(&configuration->plugin_context, $2))
```
MrAnno: Now, `KW_REQUIRES` depends on the initial discovery of candidate modules.

This currently appears on the lexer level, candidates are loaded after the first non-pragma token.
Are we sure this fix is correct? I mean, `@include` and `@requires` are pragma tokens.

bazsi: I would consider that plugin discovery is pretty much turned on everywhere, so we could rely on that. However you have a point that `@requires` might come in the first couple of lines, in which case plugin discovery has not run yet. I don't see this as a very important usecase, but I agree this should be handled with a bit more care.

MrAnno: > I would consider that plugin discovery is pretty much turned on everywhere, so we could rely on that.

I agree.

> However you have a point that `@requires` might come in the first couple of lines, in which case plugin discovery has not run yet.

If the first included SCL file has any `@requires` pragma, that SCL will be skipped incorrectly.
This can be fragile, and since we use `@include 'scl/*/*.conf'`, the bug depends on the file traversal order.

We should fix that before switching to the new and better `plugin_is_module_available()` check.

bazsi:
Agreed.

lib/pragma-grammar.ym
bazsi: I have now resolved this comment in the latest rebase.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lib/cfg.c
```diff
@@ -226,6 +212,20 @@ cfg_load_forced_modules(GlobalConfig *self)
...
+{
+  gboolean autoload_enabled = atoi(cfg_args_get(self->globals, "autoload-compiled-modules") ? : "1");
+
+  if (self->use_plugin_discovery &&
+      autoload_enabled)
+    {
+      _sync_plugin_module_path_with_global_define(self);
+      plugin_discover_candidate_modules(&self->plugin_context);
+    }
+  cfg_load_forced_modules(self);
```
Kokan: I would not expect a discover candidate modules to load certain modules, while pushing force module load into - cfg - internal business I could relate, not in this way. Would it be okay for you to drop the commit doing it?

bazsi: Honestly I would rather get rid off the entire "forced" modules concept. But I know the reason of its existence. And if it does exist already, it should not make it into an API that needs to be called by innocent layers above.

I agree that plugin_discovery might not be the best place to call it from, but I have added a new place where discovery would run and I was under the impression that forced modules needed to be loaded in that case too. But checking the code this does not seem to be the case, so yeah, I can drop that patch.

bazsi: The revision I've just force-pushed gets rid off these patches.

kira-syslogng:
Build FAILURE

bazsi:
@pzoleex this testcase syslog-ng-pe-testdb/functions/source-drivers/file-source-driver/process-operations/003-signal-9.zts/01.platform=ubuntu-xenial_amd64syslogger=syslogdversion=3.24product=ose (from syslog-ng-pe-testdb._functions_source-drivers_file-source-driver_process-operations_003-signal-9.zts) 

seems to be racy, as we are sending a SIGKILL to syslog-ng and expect it to make all lines to the destination. by the looks of it, file() destinations don't warrant this (as they ack messages back immediately after adding them to the file buffer). Seemingly all the pieces are there to ack messages back only in case they made it to disk, but it doesn't seem to be used.

In any case, my patch is pretty far from the problem. Thus, 

kira-syslogng:
Build SUCCESS

lib/cfg.c
```diff
@@ -226,6 +221,33 @@ cfg_load_forced_modules(GlobalConfig *self)
...
+}
+
+void
+cfg_discover_candidate_modules(GlobalConfig *self)
+{
+  if (self->use_plugin_discovery && _is_module_autoload_enabled(self))
+    {
+      if (_sync_plugin_module_path_with_global_define(self) || !self->plugin_discovery_finished)
+        plugin_discover_candidate_modules(&self->plugin_context);
+      self->plugin_discovery_finished = TRUE;
```
szemere: Hello,
 we discussed this IRL with @Kokan , and we have a proposal to eliminate the new `plugin_discovery_finished` variable. Handling global path change as a "plugin_context invalidation" event.

```
if (_sync_plugin_module_path_with_global_define(self))
  {
    _free_candidate_plugins(&self->plugin_context);
    self->plugin_context = NULL;
  }

if (self->plugin_context == NULL)
  plugin_discover_candidate_modules(&self->plugin_context);
```

note: This code should not call `_free_candidate_plugins` with a NULL value, but I would add a NULL check, like g_free has.


We also came up with an idea to move this discovery out of the lexer's scope, but we rather merge this PR, than widening the scope any further.

szemere:
I only have concerns with plugin discovery, otherwise LGTM. (If the version part were in a different PR, I would already merged it.)

lib/cfg.c
bazsi: This way we would be calling a private function from a higher layer (_free_candidate_plugins), which is not a good idea I think.

Also, plugin_context is not a pointer right now. I am not sure why that extra variable is bad, but if you think it is better, then a plugin_has_discovery_run() function could achieve something similar.

bazsi: I have implemented this in the latest incarnation.

kira-syslogng:
Build SUCCESS

PR #2890, opened 2019-08-27
add format-flat-json template function

Kokan:
The current `$(format-json a.b.c=1)` json formatting template function created a nested json:
```json
{
  "a": {
    "b": {
      "c": "1"
    }
  }
}
```
(withouth the whitespaces)

In some json parsing system cannot handle this nested-json format, as such a *flattened* json (in a sense that the keys are joined together).

With the new function `$(format-flat-json a.b.c=1)` became:
```json
{
  "a.b.c": "1"
}
```

In other functionality and options the `$(format-json)` and `$(format-flat-json)` are identical.

Question:
* Is it better to have a new function name instead of a simple switch (`$(format-json --flatten)`) ?
* Should it support other than `.` separator in the keys between levels ?

### DIFF
--- a/modules/json/format-json.c
+++ b/modules/json/format-json.c
@@ -208,13 +208,9 @@ tf_json_append_list(const gchar *name, const gchar *value, gsize value_len,
 }
 
 static gboolean
-tf_json_value(const gchar *name, const gchar *prefix,
-              TypeHint type, const gchar *value, gsize value_len,
-              gpointer *prefix_data, gpointer user_data)
+tf_json_append_with_type_hint(const gchar *name, TypeHint type, json_state_t *state, const gchar *value,
+                              const gssize value_len, const gboolean on_error)
 {
-  json_state_t *state = (json_state_t *)user_data;
-  gint on_error = state->template_options->on_error;
-
   switch (type)
     {
     case TYPE_HINT_STRING:
@@ -306,10 +302,21 @@ tf_json_value(const gchar *name, const gchar *prefix,
       break;
     }
     }
+  return FALSE;
+}
+
+static gboolean
+tf_json_value(const gchar *name, const gchar *prefix,
+              TypeHint type, const gchar *value, gsize value_len,
+              gpointer *prefix_data, gpointer user_data)
+{
+  json_state_t *state = (json_state_t *)user_data;
+
+  gboolean result = tf_json_append_with_type_hint(name, type, state, value, value_len, state->template_options->on_error);
 
   state->need_comma = TRUE;
 
-  return FALSE;
+  return result;
 }
 
 static gboolean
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
I would use a switch, as there are already switches that affect the
formatting of the output.

The reason it might be useful to support a separator other than '.' is that
dotted notation and object notation should be equivalent, but I would wait
until a usecase emerges. Until then the value pairs mechanism can be used
to replace the dot. This is slow and unintuitive, but possible.

lbudai:
@bazsi : the `format-flat-json` seems better to me, especially if we don't want to use context-dependent switches (eg.: a (currently WIP) separator switch would not have effect with the flatten switch)

bazsi:
I'd prefer if we limited the proliferation of template function names, but let's ask the documentation guys and maybe Gergely about their opinion. @fekete-robert 

Anyone using `$(format-json)` could also chime in.

@faxm0dem

modules/json/format-json.c
```diff
@@ -333,15 +333,18 @@ tf_json_call(LogTemplateFunction *self, gpointer s,
...
   TFJsonState *state = (TFJsonState *)s;
-  gint i;
-  gboolean r = TRUE;
   gsize orig_size = result->len;
 
-  for (i = 0; i < args->num_messages; i++)
-    r &= tf_json_append(result, state->vp, args->messages[i], args->opts, args->seq_num, args->tz);
+  for (gint i = 0; i < args->num_messages; i++)
+    {
+      gboolean r = tf_json_append(result, state->vp, args->messages[i], args->opts, args->seq_num, args->tz);
```
lbudai: it is possible to factor out this into a separate func:
```
static gboolean
_json_append(result, state, args)
{
...
}
```

```
for (gint i = 0; i < args->....; i++ )
  {
    if (!_json_append(result, state, args))
      return;
  }
```

modules/json/format-json.c
```diff
@@ -398,7 +398,99 @@ tf_flat_json_value(const gchar *name, const gchar *prefix,
 
   GString *full_name = _join_name(prefix, name);
 
-  tf_json_append_value(full_name->str, value, value_len, state, TRUE);
+  gint on_error = state->template_options->on_error;
+
+  switch (type)
```
lbudai: refactor this monster

faxm0dem:
I'd vote for a `--flatten` or `--flat` switch

fekete-robert:
I agree with faxm0dem, IMHO this should be handled using an option/flag/switch if possible. If not, I'd recommend $(format-json-flat), so anyone looking for such things in the docs would find the two implementation close to each other (though admittedly format-flat-json might be more intuitive)

gaborznagy:
@fekete-robert When you document this feature, please include currently existing option ''--leave-initial-dot" into documentation as well.

modules/json/format-json.c
gaborznagy: I would leave it as it is,  as the 2 _call functions use 2 different _append functions.
Yes, that's the only difference, but otherwise it would be more complex:
```
static gboolean
_json_append(TFJsonState *state, const LogTemplateInvokeArgs args, LogMessage *msg, 
             JsonAppendCb callback, GString *result)
{
....
}


  for (gint i = 0; i < args->num_messages; i++)
    {
      if(!_json_append(state, args, args->messages[i], tf_flat_json_append, result))
        return;
    }   
```

fekete-robert:
Thanks for pointing this out, I've added it to our todo list

gaborznagy:
I have not merged the PR yet as I was waiting for feedback to my comments.

What do you think @faxm0dem, @bazsi?

bazsi:
I think the code can be changed to accomodate the user interface.

lbudai:
@bazsi: I don't really understand your answer, but wanted to merge this as this solves an internal issue.
Of course, we can modify it later...until not released.

PR #2892, opened 2019-08-28
on 32bit platform diskq ftruncate could fail due to size 32/64 interface

Kokan:
On 32bit platforms glong is 32bit, therefore we explicitly casted
a signed 64-bit value to a signed 32-bit value, which caused an EINVAL error
on passing a negative value to ftruncate.

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -276,7 +276,7 @@ qdisk_get_empty_space(QDisk *self)
 
   if (wpos > bpos)
     {
-      return (qdisk_get_size(self) - wpos) +
+      return (qdisk_get_maximum_size(self) - wpos) +
              (bpos - QDISK_RESERVED_SPACE);
     }
 
@@ -1021,7 +1021,7 @@ qdisk_set_length(QDisk *self, gint64 new_value)
 }
 
 gint64
-qdisk_get_size(QDisk *self)
+qdisk_get_maximum_size(QDisk *self)
 {
   return self->options->disk_buf_size;
 }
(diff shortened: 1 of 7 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -437,7 +437,8 @@ qdisk_pop_head(QDisk *self, GString *record)
         {
           self->hdr->backlog_head = self->hdr->read_head;
 
-          if (!self->options->read_only && self->hdr->length == 0)
+          g_assert(self->hdr->backlog_len == 0);
```
MrAnno: @gaborznagy I think this deserves a separate commit with some explanation.
What do you think?

modules/diskq/qdisk.c
```diff
@@ -216,7 +224,7 @@ _truncate_file(QDisk *self, gint64 new_size)
       msg_error("Error truncating disk-queue file",
                 evt_tag_error("error"),
                 evt_tag_str("filename", self->filename),
-                evt_tag_long("expected-size", new_size),
+                evt_tag_printf("expected-size", "%ld", new_size),
```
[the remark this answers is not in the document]
MrAnno: `evt_tag_long` already uses `long long`. Why do we need this modification?

modules/diskq/qdisk.c
MrAnno: Discussed IRL, this belongs to the patch.

kira-syslogng:
Build SUCCESS

PR #2894, opened 2019-08-29
templates: change the $LOGHOST macro to honour use-fqdn()

bazsi:
This has been a dormant branch on my laptop, and I can't really remember what the trigger was to create this branch, but I thought I'd submit it as a PR and possibly find the reason and come up with a decision.

The LOGHOST macro always expanded to the FQDN of the host running
syslog-ng, whereas it is more logical to honour the use-fqdn() setting
like we do with other hostname related operations.

### DIFF
--- a/lib/template/templates.c
+++ b/lib/template/templates.c
@@ -370,6 +370,7 @@ log_template_options_init(LogTemplateOptions *options, GlobalConfig *cfg)
     options->frac_digits = cfg->template_options.frac_digits;
   if (options->on_error == -1)
     options->on_error = cfg->template_options.on_error;
+  options->use_fqdn = cfg->host_resolve_options.use_fqdn;
   options->initialized = TRUE;
 }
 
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lib/template/templates.c
```diff
@@ -395,6 +396,7 @@ log_template_options_defaults(LogTemplateOptions *options)
   options->frac_digits = -1;
   options->ts_format = -1;
   options->on_error = -1;
+  options->use_fqdn = TRUE;
```
gaborznagy: GlobalConfig's `use_fqdn` option is FALSE by default. Is this default okay then?

bazsi: good point. I've changed this to FALSE and also added tests to see if LOGHOST honours the setting.

I was thinking about an "upgrade" feature, but then decided against it, as it would spread GlobalConfig dependency to LogMessage, which I felt was not warranted.

I can still add a warning on the change for older config versions, but since I think this is actually a bugfix, I haven't done it. Let me know if you disagree.

bazsi: btw, documentation should probably need to be clarified, it is very short on this macro, saying it matches the output of the hostname call. @fekete-robert

kira-syslogng:
Build SUCCESS

Issue #2893, opened 2019-08-29
syslog-ng-3.22.1 throws assertion on stop


hbakken:
# syslog-ng

## Version of syslog-ng

```
 $ syslog-ng --version
syslog-ng 3 (3.22.1)
Config version: 3.22
Installer-Version: 3.22.1
Revision: 
Compile-Date: Aug 26 2019 13:59:02
Module-Directory: //lib/syslog-ng
Module-Path: //lib/syslog-ng
Include-Path: //share/syslog-ng/include
Available-Modules: add-contextual-data,affile,afprog,afsocket,afuser,appmodel,basicfuncs,cef,confgen,cryptofuncs,csvparser,date,dbparser,disk-buffer,examples,graphite,hook-commands,kvformat,linux-kmsg-format,map-value-pairs,pseudofile,sdjournal,snmptrapd-parser,stardate,syslogformat,system-source,tags-parser,tfgetent,xml
Enable-Debug: on
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: off
Enable-TCP-Wrapper: off
Enable-Linux-Caps: on
Enable-Systemd: on
```

## Platform

Homebrew linux distro, kernel versions vary (2.6.37, 3.4.x, 3.10.x and 4.4.x), arm.

## Debug bundle
```
$ syslog-ng-debun -r
Syslog-NG DEBUg buNdle generator
Start environment detection
Linux-type FHS detected
Unknown or unsupported Linux distribution!

Operating System Name: Linux
Unknown Distro, perhaps unsupported
Unknown Distro, perhaps unsupported
/sbin/syslog-ng-debun: line 941: semodule: command not found

Start general info collection
System's full uname: Linux hgbsvea 4.4.190+ #1 SMP PREEMPT Mon Aug 26 13:49:32 CEST 2019 aarch64 GNU/Linux
Getting network-interface information: Success
Getting network routes: Success
Getting DNS resolution-related information: Done
List all processes
Mount and disk free info collection

Start Syslog-specific info collection
Copy configs from /etc/syslog-ng
/sbin/syslog-ng-debun: line 654: cpio: command not found
/sbin/syslog-ng-debun: line 876: /usr/sbin/syslog-ng-ctl: No such file or directory
Gathering PKI information... done.
cat: can't open '/var/lib/syslog-ng/syslog-ng.pid': No such file or directory
SVpid:  SNGpid:  Chpids: 
[...]

PR #2895, opened 2019-08-29
logsource: add explicit (un)initialized state to WindowSizeCounter

lbudai:
Fixes: #2893

On 32 bit systems (or non-64 bit systems), syslog-ng could abort during
shutdown.

What was the reason of the abort?
a) in `log_source_set_options` where we set the initial window size
conditionally, the condition was false thus the `full_window_size`
remained 0
b) when `log_source_free` is called during shutdown,
 * `_release_dynamic_window` called unconditionally and
 *  a dynamic_part is calculated as full_window_size(=0) - init_window_size(=default 100),
 so dynamic_part = -100
 * window_size is decremented by dynamic_part(-100) and the
 `window_size_counter_sub` asserts on old_value >= value, and this
 assert failed, so syslog-ng aborted

So the questions are
1) why we did not set initial window size?
2) why old_value was not greater than value?

Answers:
1) the value we get from `window_size_counter_get` is the masked
value... on 64 bit systems this value is a 63 bits of `1` and it is compared to
a 32 bits of `1` but the 63 bits are truncated to 32 thanks to an explicit cast
And what if we are on a 32 bits system?
Well... the sizeof(gsize) is 4 , sizeof(gint) is also 4 on these
systems. This means that the `window_size_counter_get` returns 31 bits of
`-1`, and it is compared to 32 bits of `1` : they are obviously not
equals -> we won't set full_window_size

2) old_value is a -1, which is masked, so the actual old value is 2^31-1, while new value is a
-100, which is (2^32-100), so on a 32 bits system 31 bit negative value is
compared to a  to 32 bits negative value...

Proposed solution:
 * add a initialized state to LogSource: this is checked/(set to TRUE) only in
 `log_source_set_options`, and set to FALSE only in `log_source_init_instance`

### DIFF
--- a/lib/logsource.c
+++ b/lib/logsource.c
@@ -707,7 +707,9 @@ log_source_free(LogPipe *s)
 
   ack_tracker_free(self->ack_tracker);
   self->ack_tracker = NULL;
-  _release_dynamic_window(self);
+
+  if (G_UNLIKELY(dynamic_window_is_enabled(&self->dynamic_window)))
+    _release_dynamic_window(self);
 }
 
 void
--- a/lib/logsource.c
+++ b/lib/logsource.c
@@ -633,7 +633,20 @@ log_source_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options
                   evt_tag_printf("msg", "%p", msg));
 
   msg_set_context(NULL);
+}
 
+static void
+_initialize_window(LogSource *self, gint init_window_size)
+{
+  self->window_initialized = TRUE;
+  window_size_counter_set(&self->window_size, init_window_size);
+  self->full_window_size = init_window_size;
+}
+
+static gboolean
+_is_window_initialized(LogSource *self)
+{
+  return self->window_initialized;
 }
 
 void
@@ -645,11 +658,9 @@ log_source_set_options(LogSource *self, LogSourceOptions *options,
    * configuration and we received a SIGHUP.  This means that opened
    * connections will not have their window_size changed. */
 
-  if ((gint)window_size_counter_get(&self->window_size, NULL) == -1)
-    {
-      window_size_counter_set(&self->window_size, options->init_window_size);
-      self->full_window_size = options->init_window_size;
-    }
+  if (!_is_window_initialized(self))
+    _initialize_window(self, options->init_window_size);
+
   self->options = options;
   if (self->stats_id)
     g_free(self->stats_id);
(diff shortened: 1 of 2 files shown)


lib/logsource.c
```diff
@@ -633,7 +633,20 @@ log_source_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options
...
 
   msg_set_context(NULL);
+}
+
+static void
+_initialize_dynamic_window(LogSource *self, gint init_window_size)
+{
+  self->dynamic_window_initialized = TRUE;
+  window_size_counter_set(&self->window_size, init_window_size);
+  self->full_window_size = init_window_size;
```
MrAnno: `self->window_size` and `self->full_window_size` are used even when dynamic window is disabled. I'd rename these methods and the field to `_initialize_window`, `_is_window_initialized` and `LogSource::window_initialized`.

lib/logsource.c
```diff
@@ -680,6 +691,7 @@ log_source_init_instance(LogSource *self, GlobalConfig *cfg)
   self->super.init = log_source_init;
   self->super.deinit = log_source_deinit;
   window_size_counter_set(&self->window_size, (gsize)-1);
+  self->dynamic_window_initialized = FALSE;
```
MrAnno: Since this fields needs to be renamed to `window_initialized`, I think the field could belong to `WindowSizeCounter`.

`window_size_counter_set()` could set this flag, or `window_size_counter_initialize()` can be added instead. What do you think?

----


`window_size_counter_set(&self->window_size, (gsize)-1);` can be removed, the faulty -1 is not used anymore.

kira-syslogng:
Build SUCCESS

lib/logsource.c
lbudai: This field belongs the user, I would not add to the counter struct (and counter represents a counter that can be used in multithreaded environment, while this flag is definitely not)

lib/logsource.c
lbudai: Indeed.

lib/logsource.c
lbudai: And yes, the extra set should be removed.

lbudai: Making the `release_dynamic_window` call conditionally also a good idea, will update the PR soon. Thanks!

kira-syslogng:
Build SUCCESS

PR #2896, opened 2019-08-29
Invalidate LEGACY_MSGHDR when PID or PROGRAM is unset

MrAnno:
`log_msg_set_value()` contains the same invalidation logic for the legacy message header.
The same has to be done for `unset()` in order to be consistent.

For example, the legacy message header is used by `LogWriter` if no template is set, and unsetting the `PID` will not remove the `PID` field from the output:

```
rewrite { unset(value("PID")); };
Aug  2 11:11:11 test test[2222]: test
```

```
rewrite { set("" value("PID")); };
Aug  2 11:11:11 test test: test
```

More info: https://serverfault.com/questions/981150/disable-process-id-showing-in-syslog-ng-logs/981212

### DIFF
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -573,6 +573,9 @@ void
 log_msg_unset_value(LogMessage *self, NVHandle handle)
 {
   nv_table_unset_value(self->payload, handle);
+
+  if (handle == LM_V_PROGRAM || handle == LM_V_PID)
+    log_msg_unset_value(self, LM_V_LEGACY_MSGHDR);
 }
 
 void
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

lib/logmsg/logmsg.c
```diff
@@ -573,6 +573,9 @@ void
 log_msg_unset_value(LogMessage *self, NVHandle handle)
 {
   nv_table_unset_value(self->payload, handle);
+
+  if (handle == LM_V_PROGRAM || handle == LM_V_PID)
+    log_msg_unset_value(self, LM_V_LEGACY_MSGHDR);
```
lbudai: maybe we could this into a function, or, just factor out a predicate function (`_need_to_unset_legacy_header`).
But this is optional :)

MrAnno: Will do, thanks! :)

kira-syslogng:
Build SUCCESS

MrAnno:
Sorry, I pushed the wrong branch. Fixed.

kira-syslogng:
Build SUCCESS

PR #2755, opened 2019-05-29
WIP: transport: add support for recvmmsg()

bazsi:
I received this patch from ferenc.sipos@oneidentity.com and both a review and a performance check is due, but I've opened the PR so it does not get lost.

PR #2899, opened 2019-08-31
Add support for DESTIP/DESTPORT/PROTO macros

bazsi:
This is the reopening of #2495, which has been closed due to conflicts and inactivity. Since this came up again in a conversation, I rebased it against latest master.

Also, last time we've discussed that the proto/daddr (and therefore saddr) members should be grouped somehow within the LogMessage struct. I made an attempt at this, which is indeed nicer, however increases the size of the LogMessage struct by 8 bytes. Still we fit into 3 cachelines, so it shouldn't matter that much.

Please let me know if you like this reorganization, it is the last patch. This last piece is very fragile, so if it's not something you like, I would drop that as fast as possible, to avoid having to rebase it. IIRC it was @furiel and @lbudai  who preferred something like this, let me know your opinion.

The use-case behind the PR is as follows:
* someone has an appliance which sends out log messages via both UDP and TCP
* the format of the two are different, and he wants to capture either with the simplest possible filter
* netmask() doesn't work because the IP addresses are the same
* host() doesn't work because the hostnames are the same

That's how the idea of $PROTO was born. Then I've figured that if we have SOURCEIP and PROTO, we should have $DESTIP and $DESTPORT so we can have access the entire IP level tuple.


kira-syslogng:
Build FAILURE

bazsi:
This pull request **fixes 1 alert** when merging 277a8f675fe598ca4463e739128f5b0684242003 into a7eb15eeca138ed349fb3aca05f02bd843908b60 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-67624364636837acd695d8d6a1bd0ad796959d87)

**fixed alerts:**

* 1 for FIXME comment

---

**Warning** - Automated code review for balabit/syslog-ng will be disabled on October 1, 2019. You can avoid this by <a href="https://github.com/apps/lgtm-com/installations/new">installing the LGTM.com GitHub App</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

lib/transport/transport-udp-socket.c
```diff
@@ -0,0 +1,169 @@
...
+  struct cmsghdr *cmsg;
+
+  for (cmsg = CMSG_FIRSTHDR(msg); cmsg != NULL; cmsg = CMSG_NXTHDR(msg, cmsg))
+    {
+
+#ifdef __FreeBSD__
+      if (cmsg->cmsg_level == IPPROTO_IP && cmsg->cmsg_type == IP_RECVDSTADDR)
+        {
+          gint cmsg_header_len = (gchar *) CMSG_DATA(cmsg) - (gchar *) cmsg;
+          struct sockaddr *sa = (struct sockaddr *) CMSG_DATA(cmsg);
```
furiel: Could you extract the blocks inside the compiler switches to a separate functions? It would improve readibility.

lib/transport/transport-udp-socket.c
```diff
@@ -0,0 +1,169 @@
...
+
+          log_transport_aux_data_set_local_addr_ref(aux, g_sockaddr_new(sa, cmsg->cmsg_len - cmsg_header_len));
```
furiel: Just saw that in the manual there is a `CMSG_LEN` macro too. Could you use that here?

lib/transport/transport-udp-socket.c
```diff
@@ -0,0 +1,169 @@
...
+          struct in_pktinfo *inpkt = (struct in_pktinfo *) CMSG_DATA(cmsg);
+
+          g_assert(self->local_addr->sa.sa_family == AF_INET);
+          sin = *(struct sockaddr_in *) &self->local_addr->sa;
+          sin.sin_addr = inpkt->ipi_addr;
+          log_transport_aux_data_set_local_addr_ref(aux, g_sockaddr_new((struct sockaddr *) &sin, sizeof(sin)));
+          break;
+        }
+#endif
+      if (cmsg->cmsg_level == IPPROTO_IPV6 && cmsg->cmsg_type == IPV6_PKTINFO)
```
furiel: `SYSLOG_NG_ENABLE_IPV6` is missing here?

lib/transport/transport-udp-socket.c
```diff
@@ -0,0 +1,169 @@
...
+
+  if (domain == AF_INET)
+    {
+#ifdef __FreeBSD__
+      setsockopt(fd, IPPROTO_IP, IP_RECVDSTADDR, &on, sizeof(on));
+#else
+      setsockopt(fd, IPPROTO_IP, IP_PKTINFO, &on, sizeof(on));
+#endif
+    }
+  else if (domain == AF_INET6)
```
furiel: `SYSLOG_NG_ENABLE_IPV6` is missing here?

lib/transport/transport-udp-socket.c
```diff
@@ -0,0 +1,169 @@
...
+typedef struct _LogTransportUDP LogTransportUDP;
+struct _LogTransportUDP
+{
+  LogTransportSocket super;
+  GSockAddr *local_addr;
+};
+
+#if defined(SYSLOG_NG_HAVE_CTRLBUF_IN_MSGHDR)
+static void
+_feed_aux_from_cmsg(LogTransportUDP *self, LogTransportAuxData *aux, struct msghdr *msg)
```
furiel: I am curious if this has a performance impact? From now for every udp packet, we need to read the ancillary data too. Can this cause problem? Should we put the code behind optionality?

lib/transport/transport-factory-socket.c
```diff
@@ -24,13 +24,14 @@
...
 #include "transport/transport-factory-socket.h"
 #include "transport/transport-socket.h"
+#include "transport/transport-udp-socket.h"
 
 DEFINE_TRANSPORT_FACTORY_ID_FUN("socket", transport_factory_socket_id);
 
 static LogTransport *
 _construct_transport_dgram(const TransportFactory *s, gint fd)
 {
-  return log_transport_dgram_socket_new(fd);
```
furiel: I was thinking the new implementation serves as a drop-in replacement for the original implementation. Do we need to old code keep around, or it can be eliminated?

lib/transport/transport-udp-socket.c
bazsi: Sure, will do so.

lib/transport/transport-udp-socket.c
bazsi: Ill check why I wasn't using that. IIRC it wasn't available everywhere.

bazsi:
Thanks for the review comments, I am addressing them as soon as possible

folti:

kira-syslogng:
Build FAILURE

lib/transport/transport-udp-socket.c
bazsi: I've just pushed b22887c6f4b5efed365f192bc875e8d0d4e652b0 to address this issue.

kira-syslogng:
Build FAILURE

lib/transport/transport-udp-socket.c
bazsi: I went back to check my reasons. 

First of all CMSG_LEN() is not doing what you would expect by its name, e.g. it does not give you the size of the payload given the size of an ancilliary block, rather it expects the size of the payload and adds the overhead on top. It is basically a macro to initialize the cmsg_len member of "struct cmsghdr", which contains both the payload and the - aligned - header.

What we need at that point is given cmsg->cmsg_len, give me the length of the payload.  In most ancilliary messages, the payload size is constant so the code would be simpler by assuming the size of the struct being received in the payload. However, this might not work properly in the IP_RECVDSTADDR case, as the payload is of variable length and the size can only be calculated this way.

As a plug, CMSG_LEN is unfortunately not even portable.

To sum it up:
1) CMSG_LEN is not what we need and there's no function/macro that gives us what we need
2) CMSG_LEN is not portable

I could extract a function that does the size calculation, but since it's only used at one place, I don't consider it very important. Let me know if you think otherwise.

lib/transport/transport-udp-socket.c
bazsi: done in 383531132

lib/transport/transport-udp-socket.c
bazsi: also done in 51f0fefab38fc4b18262547b6673db6be227be2b

lib/transport/transport-udp-socket.c
bazsi: I don't think it would be a huge issue (well, we already incur a lot of penalty by allocating a GSockAddr instance for every datagram, now we do two). The ancillary data is retrieved in the same syscall and should be much smaller than the datagram itself.

What I would do if UDP performance should be improved is to embed the GSockAddr instance into LogMessage and avoid the memory allocation altogether, possibly doing a special case for ipv4/ipv6 traffic.

bazsi: let me know if you have a different opinion.

kira-syslogng:
Build FAILURE

lib/transport/transport-factory-socket.c
bazsi: This is a good point. as I see nothing really uses the base class anymore. Maybe some functionality could be merged by changing internal signatures.

Do you think I can delay that work once we merge this? I mean this is already pretty large and I guess incorporating that change into this PR would not make it smaller.

lbudai:
@bazsi : without reviewing all the changes, I'm sure that this is not exactly the same what we've talked about :)
As I remember, only the transport struct should be added, others shouldn't be touched.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
> @bazsi : without reviewing all the changes, I'm sure that this is not exactly the same what we've talked about :)
> As I remember, only the transport struct should be added, others shouldn't be touched.

Well, if we want to extract the transport fields why not group the others too? I mean not all messages are syslog messages either, so things like "SDATA" shouldn't be stored in the main struct, should they?

I am not arguing too hard about merging that change (I am not sure I like it either), I just wanted to do the renames as you suggested so we can look at it. I think extracting only the "transport" stuff would only be half the work, which I would not do by itself.

LogMessage is already full of optimizations (padding and size) and tight coupling with the queueing/acks subsystems, so I don't consider the extra GSockAddr pointer (from which we already have one) or the proto value is that much of an issue.

bazsi:
This pull request **fixes 1 alert** when merging 446726a4eddf8c4b7dc59ddc919c5b28de406e99 into 2fb02c3f7384ddf38af07599a87f60d4f71b804e - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-02c9c6a64c7768363c9811715796fb8125c57998)

**fixed alerts:**

* 1 for FIXME comment

---

**Warning** - Automated code review for balabit/syslog-ng will be disabled on October 1, 2019. You can avoid this by <a href="https://github.com/apps/lgtm-com/installations/new">installing the LGTM.com GitHub App</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

bazsi:
This pull request **fixes 1 alert** when merging 36db05f8ba0588fcd834c57a703bcf97a1d9089b into 2fb02c3f7384ddf38af07599a87f60d4f71b804e - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-1e1cc37b579bacaa75e30730b9cc7c87f7d7736a)

**fixed alerts:**

* 1 for FIXME comment

---

**Warning** - Automated code review for balabit/syslog-ng will be disabled on October 1, 2019. You can avoid this by <a href="https://github.com/apps/lgtm-com/installations/new">installing the LGTM.com GitHub App</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

kira-syslogng:
Build SUCCESS

lbudai:
@bazsi: then why not including a pointer to a transport struct instance?

(I have the feeling that you just wanted show us how bad idea what we've requested.)

bazsi:
There's no "transport instance" today, the socket addresses are per message values, so they can't reference the log transport instance at least with UDP. With TCP we are sharing the GSockAddr instances for all messages that are received on the same TCP connection.

With UDP each and every datagram may be received on a different IP address, this it can't easily be shared.

I didn't intend to play fun games whether this was a good or a bad idea. I was curious too, but honestly I don't like the end result very much either.

On the other hand I would like to avoid having to allocate GSockAddr instances for every packet received, and rather inline them into the base struct somehow (like we do with the LogQueueNode stuff for instance). That may actually hit two birds with one stone: make these fields a bit more separate and at the same time improve performance dramatically.

What do you think?

lib/transport/transport-factory-socket.c
furiel: Sure, this can be done later.

lib/transport/transport-udp-socket.c
furiel: I'm afraid I cannot tell. The syscall potentially lasts longer, and processing the result is also involved. The safest is to do some measurement. I will ask help from @pzoleex, and try to validate this.

bazsi:
I have just rebased this against current master. I've also removed the patch that renames LogMessage members.

I am working towards embedding LogTransport related fields into LogMessage, but that's going to be a separate PR.

kira-syslogng:
Build SUCCESS

lbudai:

kira-syslogng:
Build SUCCESS

lib/gsockaddr.c
```diff
@@ -84,7 +84,7 @@ g_sockaddr_new(struct sockaddr *sa, int salen)
 }
 
 GSockAddr *
-g_sockaddr_new_from_peer_fd(gint fd)
+g_sockaddr_new_from_peer_name(gint fd)
```
furiel: Unfortunately this change breaks installer generation. PE modules expects `g_sockaddr_new_from_peer_fd` to be available. I could fixup that in PE modules, but I am just not sure if the new name is better. Actually, the GSockAddr is still created from fd, and not from string. In that sense, the original name was better.

bazsi: The socket address is traditionally called "name" in the various socket APIs. See getsockname() & getpeername(). Also, the next patch in the series introduces a g_sockaddr_new_from_local_name(), which will use the same "fd".

Also, I understand the PE concerns, all I can offer is to  provide a patch against that code if you (as in One Identity) choose to publish them :)

bazsi:
I've rebased against current master and - hopefully - fixed the failing travis job.

kira-syslogng:
Build SUCCESS

lib/gsockaddr.c
lbudai: (just a note)
... where name means address and peer is an fd...
so we could just call it as `g_sockaddr_new_from_peer` (or `get_peer_gsockaddr`) and it's an internal detail that we get the peer address with `getpeername`.

So I think if we rename this function (and also rename it in the internal `ALTP` module), then I'd choose another name :)

furiel: Does `name` refers to the output parameter `addr`, instead of the input parameter `fd`, both in case of `getsockname` and `getpeername`? Me feels more like getting the name, instead of getting something from the name. `

Anyhow, for now, I just created a pe patch with the new name. Also I needed to rebase the branch to master, because one of the pe unit tests fail. In case you need to change the code, could you rebase against master. It would help me with the followup a little.

Other than these, the installer is generated, so no other blocking problem so far. I try to find help and time to do the performance test next week.

bazsi: We are constructing a GSockAddr instance based on a socket peer's name, where socket is denoted by fd. Maybe `g_sockaddr_new_from_connection_peer()` would work better? Then the other variant would be `g_sockaddr_new_from_connection_local()`, which feels a bit strange and pretty long too.

Another solution would be not using a constructor at all, rather add a function `g_socket_get_local_name()` and `g_socket_get_peer_name()` to gsocket.c

You know what, I just implemented the last paragraph and it does feel much better!

What do you think?

kira-syslogng:
Build SUCCESS

lib/transport/transport-udp-socket.c
```diff
@@ -185,7 +185,7 @@ log_transport_udp_setup_fd(LogTransportUDP *self, gint fd)
   gint domain;
   socklen_t domain_len = sizeof(domain);
 
-  self->bind_addr = g_sockaddr_new_from_local_name(fd);
+  self->bind_addr = g_socket_get_local_name(fd);
```
furiel: I got the following warning for implicit function declaration. Probably the header is missing somewhere.

```
/home/furiel/workspace/destip-destport-proto-macros/syslog-ng-core/lib/transport/transport-udp-socket.c:188:21: warning: implicit declaration of function ‘g_socket_get_local_name’ [-Wimplicit-function-declaration]
   self->bind_addr = g_socket_get_local_name(fd);
```

lib/gsockaddr.c
furiel: Looks good to me. Thank you!

lib/transport/transport-udp-socket.c
bazsi: In the meantime I've started working on containing the GSockAddr instances right within LogMessage, but that's a huge change, which I'd prefer do in a separate PR. Initial performance testing suggests that there's ~5% improvement in performance if we do that, assuming that _both_ addresses are stored, which was around 350k/sec per thread on my laptop (without the inlineing), and ~365k (with the inlining).

What this boils down to is that even storing two instances instead of one is not that much of a headache (e.g. 2-3% max), and that only applies to UDP reception speed.

To improve our UDP story, I'd do something like the recvmmsg() patches in #2755 and also do the inlining. That should give us a considerable boost.

With all this said, I think merging this should be fine for now from the performance standpoint. GSockAddr inlining, UDP performance boost could come later.

lib/transport/transport-udp-socket.c
bazsi: thanks, fixed.

lib/transport/transport-udp-socket.c
furiel: I disagree with that. I am happy to hear there are other ways to improve udp performance. But if this patch introduces a performance problem, we need to know it now. Because if there is a performance problem, someone needs to come back at some point to fix it, no matter the other performance improvements in other parts of the code.

The action point is at the testing team now. And I am really sorry that the performance takes this long, we are just before release and there is no testing capacity available. I hope we can have progress next week.

kira-syslogng:
Build SUCCESS

furiel:
@bazsi could you check why the mac build fails? Maybe an include is missing?

```
lib/transport/transport-socket.c:37:34: error: use of undeclared identifier 'SO_PROTOCOL'

  if (getsockopt(fd, SOL_SOCKET, SO_PROTOCOL, &ipproto, &ipproto_len) >= 0)
```

lib/transport/transport-udp-socket.c
furiel: @pzoleex did us the measurement, and you were right: there is no performance problem. Sorry for the delay! Marking as resolved. 

```
nightly
100k eps -> Average outgoing lines/sec: 88130 msg/sec
200k eps -> Average outgoing lines/sec: 104996 msg/sec
new code:
100k eps -> Average outgoing lines/sec: 86271 msg/sec
200k eps -> Average outgoing lines/sec: 103333 msg/sec
original code:
100k eps -> Average outgoing lines/sec: 88066 msg/sec
200k eps -> Average outgoing lines/sec: 105556 msg/sec
```
Measurement: simple udp source -> file destination on single thread. EPS measured on the output file.

furiel:
@bazsi from my point of view, if the travis is green, this can go in.

kira-syslogng:
Build SUCCESS

furiel:
@bazsi unfortunately mac build is still failing.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
@bazsi just occured to me. Is it intended that `PROTO` emits integer?

For a config I received:
```
  destination { file(/dev/stdout template("$DESTIP $DESTPORT $PROTO\n")); };
```
```
127.0.0.1 5555 6
```

Shouldn't we write "tcp" instead?

If we plan to write human readable version, then we should not expose `PROTO` macro now. Otherwise we could not ensure backward compatibility.

furiel:
Just to clarify, the currently available getent template function can do the transformation if needed (supports protocols), so it is not the end of the world if we emit code instead of name.

I just wanted to make sure this is what we want. I added my approve, so I am fine with this version too. Though, my personal preference would be a string representation instead integer.

bazsi:
We discussed this IRL however it makes sense to document that here too: the numeric representation for PROTO was intentional, as this matches how a firewall log entry encapsulates a five-tuple (srcip, dstip, proto, srcport, dstport).

And the $PROTO value is expected to be used rarely, on the first-hop syslog-ng instance.

bazsi:
Let me know if it helps to rebase this to current master, I don't want to invalidate @furiel's review now that we are so close to merging this.

bazsi:
rebased and resolved conflicts.

kira-syslogng:
Build SUCCESS

PR #2903, opened 2019-09-04
logreader,logsource: move scratch-buffer mark and reclaim into LogSource

gaborznagy:
To be able to support using simple scratch_buffers_alloc() call for end-users,
scratch-buffers should be marked and reclaimed after each message that is passed through the log path.
For this the best location is on source side, so this happens after every message  (currently mainloop-workers use an explicit_gc call, therefore reclaiming every used scratch-buffers).

The place for doing this should be in LogSource rather then LogReader, as in case the latter, we need to add scratch_buffers_reclaim to every source: internal(), systemd-journal(), python-source(), and all threaded-source drivers as well.
This shows that it's real place is in LogSource::log_source_post().

### DIFF
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -25,7 +25,6 @@
 #include "logreader.h"
 #include "mainloop-call.h"
 #include "ack-tracker/ack_tracker.h"
-#include "scratch-buffers.h"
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

PR #2905, opened 2019-09-06
afinet: fix multicast ip detection

### LINKED ISSUE #2889: syslog-ng mutlicast destination TTL stuck at 1
I am trying to set up a multicast network destination so the logs can be received by multiple hosts without having to send it out X number of times. 

I have:
destination d_multicast { udp(“239.0.0.1” port(4440) ip_ttl(8) ); };

The multicast message goes out, however,  it never reaches the destination because the ttl doesn’t get set to 8. It stays at 1. I can’t get the message off network with a ttl of 1. I’ve tried several different destination setups but none allow the ttl to go over 1.

I tried RHEL7.5 packaged version (3.5.6) and also compiled 3.7.1 and 3.22.1 and all had the same issue? Is there any version that this works in?  Conf file and capture attached. 

TTL
[conf.txt](https://github.com/balabit/syslog-ng/files/3541542/conf.txt)

alltilla:
IP multicast detection should to be done on the destination ip,
not the bind ip, as it was erroneously done before.

Fixes #2889

### DIFF
--- a/modules/afsocket/afinet-dest-failover.c
+++ b/modules/afsocket/afinet-dest-failover.c
@@ -250,6 +250,7 @@ _setup_failback_fd(AFInetDestDriverFailover *self)
   if (!transport_mapper_open_socket(self->failover_transport_mapper.transport_mapper,
                                     self->failover_transport_mapper.socket_options,
                                     self->bind_addr,
+                                    self->primary_addr,
                                     AFSOCKET_DIR_SEND,
                                     &self->fd.fd))
     {
--- a/modules/afsocket/afsocket-dest.c
+++ b/modules/afsocket/afsocket-dest.c
@@ -307,8 +307,8 @@ afsocket_dd_start_connect(AFSocketDestDriver *self)
   g_assert(self->transport_mapper->transport);
   g_assert(self->bind_addr);
 
-  if (!transport_mapper_open_socket(self->transport_mapper, self->socket_options, self->bind_addr, AFSOCKET_DIR_SEND,
-                                    &sock))
+  if (!transport_mapper_open_socket(self->transport_mapper, self->socket_options, self->bind_addr, self->dest_addr,
+                                    AFSOCKET_DIR_SEND, &sock))
     {
       return FALSE;
     }
--- a/modules/afsocket/afsocket-source.c
+++ b/modules/afsocket/afsocket-source.c
@@ -819,8 +819,8 @@ _sd_open_stream(AFSocketSourceDriver *self)
       if (!afsocket_sd_acquire_socket(self, &sock))
         return self->super.super.optional;
       if (sock == -1
-          && !transport_mapper_open_socket(self->transport_mapper, self->socket_options, self->bind_addr, AFSOCKET_DIR_RECV,
-                                           &sock))
+          && !transport_mapper_open_socket(self->transport_mapper, self->socket_options, self->bind_addr,
+                                           self->bind_addr, AFSOCKET_DIR_RECV, &sock))
         return self->super.super.optional;
     }
   self->fd = sock;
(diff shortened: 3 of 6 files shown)


kira-syslogng:
Build SUCCESS

modules/afsocket/socket-options-inet.h
```diff
@@ -36,6 +36,8 @@ typedef struct _SocketOptionsInet
   gint tcp_keepalive_intvl;
   gint tcp_keepalive_probes;
   gchar *interface_name;
+  /* calculated */
+  gboolean multicast;
```
MrAnno: From the perspective of `SocketOptionsInet`, it's an option that has to be set by the user of `SocketOptionsInet`, just like any other options. I think we shouldn't note here that the value is calculated from something unknown to this class. I'd also remove `/* user settings */`.

The best would be if this flag was removed from the struct, and we could pass it as a parameter of `socket_options_inet_setup_socket()`. However, this is not possible, because `socket_options_inet_setup_socket()` is virtual and isn't called directly in `AFInetDestDriver` (and `multicast` is meaningless at the abstract `SocketOptions` level); so I think it is the good place.

MrAnno:
As an alternative fix, have you considered passing the `dest_addr` instead of `bind_addr` to `transport_mapper_open_socket()` only on the destination side?

This way, you don't need to add this flag and move the calculation outside of the `SocketOptionsInet` class.

I mean, `bind_addr` does not really mean anything on the destination side,  `SocketOptionsInet` only uses this address to determine the socket family and whether the address is multicast.

For example:

```diff
--- a/modules/afsocket/afsocket-dest.c
+++ b/modules/afsocket/afsocket-dest.c
@@ -306,15 +306,14 @@ afsocket_dd_start_connect(AFSocketDestDriver *self)

   g_assert(self->transport_mapper->transport);
   g_assert(self->bind_addr);
+  g_assert(self->dest_addr);

-  if (!transport_mapper_open_socket(self->transport_mapper, self->socket_options, self->bind_addr, AFSOCKET_DIR_SEND,
+  if (!transport_mapper_open_socket(self->transport_mapper, self->socket_options, self->dest_addr, AFSOCKET_DIR_SEND,
                                     &sock))
     {
       return FALSE;
     }

-  g_assert(self->dest_addr);
-
   rc = g_connect(sock, self->dest_addr);
   if (rc == G_IO_STATUS_NORMAL)
     {
```

MrAnno:
As an alternative fix, have you considered passing the `dest_addr` instead of `bind_addr` to `socket_options_setup_socket()` only on the destination side?

This way, you don't need to add this `multicast` flag or move the calculation outside of the `SocketOptionsInet` class.

I mean, `bind_addr` does not really mean anything on the destination side,  `SocketOptionsInet` only uses this address to determine the socket family and whether the address is multicast.


This has its drawbacks too:
`transport_mapper_open_socket()` needs both `bind_addr` and `dest_addr` this way, because there is one difference between these 2 addresses on the destination side: the port number. You can bind a specific local port for your destination, that's why `bind_addr` is needed too.

And of course, `transport_mapper_open_socket()` is not a destination-only function, so it may look weird if it received 2 different addresses.

bazsi:
I would either move the entire multicast related setup code to a higher layer (e.g.  afsocket) or simply pass the destination address (calling it peer_addr) to the setup function.

I think that the 2nd would be both easier and nicer. There are cases in the code where the concept of `peer_addr` exists and its value can be NULL (e.g. afsocket-source and used for datagram oriented sockets).

alltilla:
@bazsi @MrAnno : Thanks for the tips, I added `peer_addr` to `transport_mapper_open_socket()`.

kira-syslogng:
Build SUCCESS

lbudai:

kira-syslogng:
Build SUCCESS

PR #2907, opened 2019-09-06
logthrsource/fetcher: crash after failed reload

furiel:
For example with diskq-source: syslog-ng crashes after a failed reload.

Example: start syslog-ng with the config below, comment out the nonexistent destination, then reload. 
After reload, syslog-ng will crash.

```
@version: 3.23

log {
  source { example-diskq-source(file("/tmp/doesnotmatter.bin")); };
#  destination(doesnotexist);
};

```

The root cause is that the worker in logthreaded source (and thus logthreaded fetcher) is created during new, but freed during deinit. During reload, syslog-ng calls deinit for all pipes. But if the reload fails, syslog-ng will reinitialize pipes: calling init. In the end, the same worker that was deallocated previously will be used by the source.

Delaying the construction of worker fixes the problem.

### DIFF
--- a/lib/logthrsource/tests/test_logthrsourcedrv.c
+++ b/lib/logthrsource/tests/test_logthrsourcedrv.c
@@ -39,10 +39,15 @@ typedef struct _TestThreadedSourceDriver
   gint num_of_messages_to_generate;
   gboolean suspended;
   gboolean exit_requested;
+  gboolean blocking_post;
 } TestThreadedSourceDriver;
 
 MainLoopOptions main_loop_options = {0};
 MainLoop *main_loop;
+
+static void _request_exit(LogThreadedSourceDriver *s);
+static void _run_simple(LogThreadedSourceDriver *s);
+static void _run_using_blocking_posts(LogThreadedSourceDriver *s);
 
 static const gchar *
 _generate_persist_name(const LogPipe *s)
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
I started to implement, and I although I got stuck on a technical problem, I can see it would extremely reduce the complexity of this patch. Also, maybe I could add the asserts back that I needed to remove for now (@MrAnno also misses those as he mentioned a few comments above). I added the wip flag back. Unfortunately, my focus will be elsewhere for a few days, but I want to come back soon.

furiel:
Unfortunately I could not get it done. The problem is as soon I would like to add extra parameters to the init function, it cannot function as pipe init function. I was thinking about if this could be workarounded somehow, but anything I could come up with leads to the same concept: LogThreadedSourceDriver should be an abstract class, and must be inherited from it properly. And architecturally the cleanest approach is what @MrAnno did by lazy constructing the worker (in practice: adding a create_worker virtual function).

The only part that is missing from @MrAnno's branch is currently there is no place to add options to ThreadedFetcher. ThreadedSourceOptions is inlined into the structure, so we cannot inherit from it.

I would like to spend a little more time on that maybe I can finish @MrAnno's work in a timely manner.

lbudai:

lbudai:
@furiel : okay, then I'd merge this as-is
( Of course, later we need to integrate @MrAnno 's branch. )

kira-syslogng:
Build SUCCESS

furiel:
@lbudai thank you! I agree. Removed the WIP flag.

lib/logthrsource/tests/test_logthrsourcedrv.c
```diff
@@ -198,8 +222,6 @@ Test(logthrsourcedrv, test_threaded_source_suspend)
   s->num_of_messages_to_generate = 5;
```
szemere: optional: Here we have an implicit `blocking_post = FALSE` , set by the `g_new0` in the `test_threaded_source_driver_init_method` function. It would be harder to misuse this test if we have an explicit function pointer.

PR #2911, opened 2019-09-09
logpipe: local_path_options use after out of scope

Kokan:
The path_options in case of hard flow control uses the address of local_path_options, that scope ends before the path_options is used in `s->queue(s, msg, path_options)`.


kira-syslogng:
Build SUCCESS

PR #2862, opened 2019-08-05
[wip] Fix persist issues

lbudai:

PR #2914, opened 2019-09-10
Fix persist versions

szemere:
Work in progress. Based on @lbudai's #2862 
The main goal of the PR is to seamlessly convert old persist entries into their newer format.
WIP because: LogProtoBufferedServerState will require additional work, unfortunately we have mixing persist file and persist entry logic.

### DIFF
--- a/lib/host-id.c
+++ b/lib/host-id.c
@@ -69,44 +69,39 @@ host_id_init(PersistState *state)
   guint8 version;
   PersistEntryHandle handle;
   HostIdState *host_id_state;
-  gboolean new_host_id_required = FALSE;
-  gboolean use_legacy_host_id = FALSE;
+  gboolean host_id_found = TRUE;
   guint32 legacy_hostid = 0;
 
 
   handle = persist_state_lookup_entry(state, HOST_ID_PERSIST_KEY, &size, &version);
-
   if (handle == 0)
     {
+      host_id_found = FALSE;
+
       handle = persist_state_alloc_entry(state, HOST_ID_PERSIST_KEY, sizeof(HostIdState));
-      use_legacy_host_id = _load_host_id_from_legacy_persist_entry(state, &legacy_hostid);
-
-      new_host_id_required = !use_legacy_host_id;
-    }
-
-  if (!handle)
-    {
-      msg_error("host-id: could not allocate persist state");
-      return FALSE;
+      if (handle == 0)
+        {
+          msg_error("host-id: could not allocate persist state");
+          return FALSE;
+        }
     }
 
   host_id_state = persist_state_map_entry(state, handle);
-  {
-    if (new_host_id_required)
-      {
-        global_host_id = _create_host_id();
-        host_id_state->host_id = global_host_id;
-      }
-    else if (use_legacy_host_id)
-      {
-        host_id_state->host_id = legacy_hostid;
-        global_host_id = legacy_hostid;
-      }
-    else
-      {
-        global_host_id = host_id_state->host_id;
-      }
-  }
+
+  if (!host_id_found)
+    {
+      if (_load_host_id_from_legacy_persist_entry(state, &legacy_hostid))
+        {
+          host_id_state->host_id = legacy_hostid;
+        }
+      else
+        {
+          host_id_state->host_id = _create_host_id();
+        }
+    }
+
+  global_host_id = host_id_state->host_id;
+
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

lbudai:
@szemere:  why is it still WIP ?
Yesterday we had a discussion and we agreed on that this should be merged. (I guess you moved on the next task :) but before moving to the other task, finish this one).

kira-syslogng:
Build SUCCESS

lib/run-id.c
```diff
@@ -48,9 +48,11 @@ run_id_init(PersistState *state)
 
   handle = persist_state_lookup_entry(state, RUN_ID_PERSIST_KEY, &size, &version);
 
-  if (handle == 0)
+  if (handle == 0 || size != sizeof(RunIDState))
```
alltilla: Can we use the value of the invalid size "run_id" for the new "run_id" entry?

lib/run-id.c
```diff
@@ -48,9 +48,11 @@ run_id_init(PersistState *state)
...
     {
-      handle = persist_state_alloc_entry(state, RUN_ID_PERSIST_KEY, sizeof(RunIDState) );
+      if (handle != 0)
+        msg_warning("run-id: persist state: invalid size, allocating a new one");
```
furiel: [Optional] It might worth emitting the actual and expected size.

lib/host-id.c
```diff
@@ -51,13 +70,18 @@ host_id_init(PersistState *state)
   PersistEntryHandle handle;
   HostIdState *host_id_state;
   gboolean new_host_id_required = FALSE;
+  gboolean use_legacy_host_id = FALSE;
```
furiel: For me it is a little hard to follow because the new code is interleaving with the original code in a complex way. There are also a lot of states here in the machinery: new_host_id_required, use_legacy_host_id, legacy_host_id, global_host_id.

I am thinking of reorganizing this function as follows:

```
gboolean
host_id_init(PersistState *state)
{
  convert_legacy_host_id_entry(...);
  [... leave everything as it is now].
}
```

Whereas `convert_legacy_host_entry`:
- lookup the legacy key. If not available, just return.
- lookup the new key. If available, just delete the old entry and return: the damage is already done, we do not want to change host_id that is currently used.
- Rename the entry.

From that we can do everything as usual: we are either having a proper host id entry or need to create a fresh one from scratch. Hence the rest of host_id_init can stay.

What do you think?

lib/logproto/logproto-buffered-server.c
```diff
@@ -415,6 +415,58 @@ log_proto_buffered_server_convert_state(LogProtoBufferedServer *self, guint8 per
...
+  state = persist_state_map_entry(persist_state, old_state_handle);
+  if (state->header.version > 1)
+    {
+      msg_error("Internal error restoring log reader state, stored data is too new",
+                evt_tag_int("version", state->header.version));
+      result = FALSE;
+    }
+
+  if (state->header.version == 0)
+    {
```
furiel: Is something missing here? As I see, if version is zero, we assign zero to the version, which is a noop.

lib/host-id.c
```diff
@@ -62,51 +62,45 @@ _load_host_id_from_legacy_persist_entry(PersistState *state, guint32 *host_id)
   return TRUE;
 }
 
-gboolean
-host_id_init(PersistState *state)
+gboolean host_id_init(PersistState *state)
```
alltilla: Please move the function name to a new line, so it can be found with `git grep ^host_id_init`.

lib/host-id.c
alltilla: I feel a dissonance in a clean code perspective with:
```
  if (new_host_id_required)
    {
      if (_load_host_id_from_legacy_persist_entry(state, &legacy_hostid))
```
If a new `host_id` is required, why are we loading it? I understand the algorithm here, just the names/order are confusing.

furiel's refactor idea seems nice to me.

alltilla: Personally, I am okay with the new version.

furiel: We discussed with @szemere and concluded that we would like to promote the happy path: no conversion is needed. In my version, the conversion would be checked each time. In summary, I'm fine with the new version.

kira-syslogng:
Build SUCCESS

PR #2917, opened 2019-09-11
java-modules: add gradle-user-home to gradle_args

alltilla:
With 283523530b29cea8091585c9f69d68734646a966 it was removed (`-g $(GRADLE_WORKDIR)` option), but it is needed to be able to disable caching.

It does not break caching, and provides the old functionality, when caching is not enabled.

### DIFF
--- a/modules/java-modules/Makefile.am
+++ b/modules/java-modules/Makefile.am
@@ -8,6 +8,7 @@ GRADLE_FLAGS ?=
 GRADLE_ARGS = $(GRADLE_FLAGS) \
 	--project-dir $(GRADLE_PROJECT_DIR) \
 	--project-cache-dir=$(GRADLE_PROJECT_CACHE_DIR) \
+	--gradle-user-home=$(GRADLE_PROJECT_CACHE_DIR) \
 	-PsyslogBuildDir=$(abs_top_builddir)/modules/java-modules \
 	-PsyslogDepsDir=$(abs_top_builddir)/modules/java/syslog-ng-core/libs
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
Hmmm I think this should be set outside of the makefile, and it is indeed set by dbld/rules. Why is this setting needed?

alltilla:
So before #2857, the downloaded .jar files were stored in `modules/java-modules/.gradle`, and they were removed with `make clean`. After the PR, `gradle-user-home` is not set, and the .jar files are cached somewhere else, so `make clean` wont remove them. We talked with @bazsi IRL about this problem, and came to the conclusion, that if the user wants to redownload the .jar files every time, `gradle-user-home` should be changed on the build environment, and not in the Makefile.

So closing the PR.

lbudai:
just to be precise: -g (--gradle-user-home) is not for 'disabling' caching... there are build environments out there (yes, OBS is such a build environment), where persistent cache is meaningless but they need a gradle home dir (and cache... ).
If possible, try to take OBS into account when touching gradle in the future.

@czanik : your builds could be also affected (the error message is totally useless, but this is the root cause behind that)

bazsi:
Then set gradle home from the packaging script. It should not be set from makefiles.

lbudai:
@bazsi : yes, it can be done (and there are at least two options for doing that... but only after you figured it out from a generic error message what happened to your build...). So what I'm saying is that there are other build systems out there, not just dbld...

bazsi:
True enough, but all build systems should use appropriate layering or else
we will have a mess.

PR #2920, opened 2019-09-15
LightPreCommitConfig: Remove pyupgrade from hooks

mitzkia:
- it was added accidentally
- until we need to support Python v2 we did not profit from it
during the linting process


kira-syslogng:
Build SUCCESS

PR #2921, opened 2019-09-15
LightRunWithStrace: Run syslog-ng behind strace

mitzkia:
For now there is only one integrated tool which can help to debug syslog-ng during the tests it is valgrind.

I wanted to debug why syslog-ng starts slowly in my environment, so I have added strace as a new external debug tool.
Example execution:
```console
python3 -m pytest --installdir=[path] [testcasepath] --run-with-strace
```
Strace logs will be saved also in current working dir started with "strace_" prefix
If this option is enabled all discovered testcases will use strace.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng/syslog_ng_executor.py
```diff
@@ -59,6 +59,24 @@ def run_process_with_valgrind(self):
...
+    def run_process_with_strace(self):
+        strace_command_args = [
+            "strace",
+            "-s",
+            "4096",
+            "-tt",
+            "-T",
+            "-ff",
+            "-o",
+            self.__instance_paths.get_strace_log_path(),
```
MrAnno: It would be better if the user could specify the parameters of `strace`. We can use a default of course, if not specified.

tests/python_functional/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -89,6 +90,8 @@ def __wait_for_start(self):
     def __start_syslog_ng(self):
         if self.__valgrind_usage:
             self.__process = self.__syslog_ng_executor.run_process_with_valgrind()
+        elif self.__strace_usage:
+            self.__process = self.__syslog_ng_executor.run_process_with_strace()
```
MrAnno: This is a lot of boilerplates, the same getters/helpers are copied everywhere as with valgrind.

After adding `strace`, I had the feeling that this should be generalized.
For example:
- replace `--run-with-valgrind`, `--run-with-strace`, `--run-with-*` with `--run-under=randombinary`
- you can use a short name, good defaults will be used in this case: `--run-under=valgrind`, `--run-under=strace`
- you can specify the command, its flags, and a marker where to insert the process, for example: `--run-under='something-that-light-doesnt-even-know -a -b -c {}'`

mitzkia: Thanks for the review notes.
How about if we could work like this way:
1, Remove: --run-with-valgrind command line option (we will broke user API, but we can added an alternative)
2, Add: --external-tool-name=[valgrind/strace/other], if user choose from valgrind or strace we could offer our "default" command line start parameters for this tools. User also can define own tool (other) for own risk
3, Add: User also can specify custom external tool parameters: --external-tool-parameters=[-o file-path -s 2222 -ff...], this will override the default start parameters for integrated tools.

Command line example ideas:
1, Run strace with "hardcoded" (default) parameters
python3 -m pytest --installdir=[path] [func-test-path] --external-tool-name=strace
2, Run strace with "custom" parameters
python3 -m pytest --installdir=[path] [func-test-path] --external-tool-name=strace --external-tool-parameters="my custom strace parameters"

And of course +1 for removing boilerplates.

MrAnno: The name `--external-tool-*` does not really tell what it will be used for. `run-under` and `run-with` were better. Splitting the parameters from the command can be confusing in my opinion.
And if Light does not specially support the "tool", you need something to specify where to insert the monitored command and its options. We should be able to run the test under anything without adding explicit support in my opinion. Not because we want to use hundreds of tools in the future, it's about avoiding to add the support for every tool manually/creating boilerplates.

mitzkia: Maybe there is a misunderstanding, so I would like to clarify things.

* About wrong option name: 
  - every option name has a short description, we can describe what it means.
* My original plan with the external tools was only the following:
  - user can choose from the "supported" list, and Light will run syslog-ng behind this tool with a predefined parameters. There was no plan to support other things (apply to other process)
* What internal "support" is mean in case of this tools:
  - Light knows where is the "working directory" for each testcases and the output of tools will be saved into this directory. Because of this 

By comparison of this:
- I am ok with defining tool name with one parameter: --external-tool="" or --run-under="" or something similar. This can remove some boilerplates.
- I am ok with defining tool parameters externally from command line. This can be acceptable because user can know other parameter options instead what we predefined internally. But in this case Light can not help to redirect the output of the tool into the current working dir. "In this case Light does not support the tool"
- I am ok with defining unknown tools. It make sense because Light can not support every tool. "In this case again Light does not support the tool"
- I am not really understand how we can use one parameter for defining tool name and it's parameters. And again I would like to support predefined parameter list for known tools.
- For now I did not see (and did not like to support) how this tools can help in debugging other than syslog-ng process. 

My plan for this PR was:
- adding support to run syslog-ng behind strace with predefined parameters

mitzkia: After a discussion with @MrAnno I have updated the code.
In this step (PR)  I have refactored the code in a way where additional tools are easy to add.
Next steps (next PR) could support parameters from command line.

kira-syslogng:
Build SUCCESS

tests/python_functional/src/conftest.py
```diff
@@ -39,7 +39,7 @@ def fake_testcase_parameters(request, tmpdir):
     tc_parameters.WORKING_DIR = tmpdir.join("workingdir")
     request.config.option.installdir = tmpdir.join("installdir")
     request.config.option.reports = tmpdir.join("reports")
-    request.config.option.runwithvalgrind = False
+    request.config.option.runundertool = False
```
alltilla: I would change the default to `""`, as we are expecting strings here.

tests/python_functional/conftest.py
```diff
@@ -41,7 +41,7 @@
 # Command line options
 def pytest_addoption(parser):
     parser.addoption("--runslow", action="store_true", default=False, help="run slow tests")
-    parser.addoption("--run-with-valgrind", action="store_true", default=False, help="Run tests behind valgrind")
+    parser.addoption("--run-under", help="Run syslog-ng under selected tool, eg: [valgrind]")
```
alltilla: > eg: [valgrind]

I would change this to: `supported tools: valgrind`, and then add `strace` as well in the next commit.

tests/python_functional/src/conftest.py
mitzkia: Thanks for the review note. I have fixed this.

tests/python_functional/conftest.py
mitzkia: Thanks for the review note. I have fixed this.

tests/python_functional/src/syslog_ng/syslog_ng_executor.py
mitzkia: As we discussed this PR only prepares the possibility of this feature.
After this PR next can resolve this feature also.

kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng/syslog_ng_paths.py
```diff
@@ -98,5 +95,12 @@ def get_syslog_ng_bin(self):
     def get_syslog_ng_ctl_bin(self):
         return self.__syslog_ng_paths["binary_file_paths"]["syslog_ng_ctl"]
 
-    def get_valgrind_log_path(self):
-        return self.__syslog_ng_paths["file_paths"]["valgrind"]
+    def register_external_tool_output_path(self, external_tool):
+        self.__syslog_ng_paths['file_paths'].update(
```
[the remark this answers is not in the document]
mitzkia: Understand, and agreed.

tests/python_functional/src/syslog_ng/syslog_ng_paths.py
```diff
@@ -98,5 +95,12 @@ def get_syslog_ng_bin(self):
...
+            {
+                external_tool: Path(self.working_dir, "{}_{}.log".format(external_tool, get_unique_id())),
```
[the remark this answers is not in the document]
mitzkia: +1 Agreed, I will fix this in this PR.

tests/python_functional/src/syslog_ng/syslog_ng_paths.py
```diff
@@ -98,5 +95,12 @@ def get_syslog_ng_bin(self):
...
+            },
+        )
+
+    def get_external_tool_output_path(self, external_tool):
```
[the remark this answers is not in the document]
mitzkia: I will think about that how can we eliminate external_tool parameter here. You are right SyslogNgPath() class knows about the registered external_tool.

tests/python_functional/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -87,8 +87,8 @@ def __wait_for_start(self):
             raise Exception("Start message not arrived")
 
     def __start_syslog_ng(self):
-        if self.__valgrind_usage:
-            self.__process = self.__syslog_ng_executor.run_process_with_valgrind()
+        if self.__external_tool:
+            self.__process = self.__syslog_ng_executor.run_process_with_external_tool(self.__external_tool)
```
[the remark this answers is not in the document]
mitzkia: Yes, I agreed with you. I will fix this in a new PR.

tests/python_functional/src/syslog_ng/syslog_ng_paths.py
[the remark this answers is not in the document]
mitzkia: Fixed

kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build SUCCESS

PR #2922, opened 2019-09-15
Fix wildcard-file() option validation

MrAnno:
This PR
- fixes the multi-line option validation when `wildcard-file()` is used (it was missing),
- adds another validation: `pad-size()` and `multi-line-mode()` are mutually exclusive options.

### DIFF
--- a/modules/affile/affile-source.c
+++ b/modules/affile/affile-source.c
@@ -82,19 +82,6 @@ affile_sd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options)
 }
 
 static gboolean
-_are_multi_line_settings_invalid(AFFileSourceDriver *self)
-{
-  LogProtoMultiLineServerOptions *multi_line_options =
-    (LogProtoMultiLineServerOptions *) &self->file_reader_options.reader_options.proto_options;
-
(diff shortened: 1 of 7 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

modules/affile/affile-source.c
```diff
@@ -104,20 +91,16 @@ affile_sd_init(LogPipe *s)
     return FALSE;
 
   file_reader_options_init(&self->file_reader_options, cfg, self->super.super.group);
+  if (!file_reader_options_validate(&self->file_reader_options))
```
Kokan: This feels more like the responsibility of the `FileReader` class, thus pushing it down the `FileReader` init function. What do you think ?

MrAnno: I agree. This is here, because `wildcard-file()` creates its readers on-the-fly, so the validation would run after `init()`.

MrAnno: We've discussed this IRL, `validate()` is not really necessary, I'll move the validation into `file_reader_options_init()` (changing its return value).

kira-syslogng:
Build SUCCESS

PR #2925, opened 2019-09-16
run-id: Fix "equality comparison with extraneous parentheses"

alltilla:


kira-syslogng:
Build SUCCESS

PR #2928, opened 2019-09-16
afsocket-dest: rename old persist entries

szemere:
In d29b98c9a261cbfd04dbe6afbb17008fa5108e7c we changed the persist name
from afsocket_dd_connection to afsocket_dd_connections (notice the s)
We also changed the hostname parameters.

There is no API to actually rename an entry in the persist file (We can
do it for runtime use only.), so we actually allocate a new entry with
the new name, and copy the content of the old one.

Question for the reviewer: Why don't we check the return value of https://github.com/balabit/syslog-ng/blob/master/modules/afsocket/afsocket-dest.c#L237

### DIFF
--- a/modules/afsocket/afsocket-dest.c
+++ b/modules/afsocket/afsocket-dest.c
@@ -246,28 +246,30 @@ _copy_persist_entry_value(PersistState *state, PersistEntryHandle from, PersistE
 }
 
 static gboolean
-_update_legacy_connection_persist_name(AFSocketDestDriver *self, PersistState *state)
+_update_legacy_connection_persist_name(AFSocketDestDriver *self)
 {
+  GlobalConfig *cfg = log_pipe_get_config(&self->super.super.super);
+
   const gchar *old_name, *new_name;
   PersistEntryHandle old_handle, new_handle;
   gsize size;
   guint8 version;
 
   new_name = afsocket_dd_format_connections_name(self);
-  new_handle = persist_state_lookup_entry(state, new_name, &size, &version);
+  new_handle = persist_state_lookup_entry(cfg->state, new_name, &size, &version);
   if (new_handle)
     return TRUE;
 
   old_name = afsocket_dd_format_legacy_connection_name(self);
-  old_handle = persist_state_lookup_entry(state, old_name, &size, &version);
+  old_handle = persist_state_lookup_entry(cfg->state, old_name, &size, &version);
   if (!old_handle)
     return TRUE;
 
-  new_handle = persist_state_alloc_entry(state, new_name, size);
+  new_handle = persist_state_alloc_entry(cfg->state, new_name, size);
   if (!new_handle)
     return FALSE;
 
-  _copy_persist_entry_value(state, old_handle, new_handle, size);
+  _copy_persist_entry_value(cfg->state, old_handle, new_handle, size);
 
   msg_debug("Connections persist name changed",
             evt_tag_str("old name", old_name),
@@ -302,9 +304,6 @@ afsocket_dd_connected(AFSocketDestDriver *self)
     return FALSE;
 
   proto = log_proto_client_factory_construct(self->proto_factory, transport, &self->writer_options.proto_options.super);
-
-  if (!_update_legacy_connection_persist_name(self, cfg->state))
-    return FALSE;
 
   log_proto_client_restart_with_state(proto, cfg->state, afsocket_dd_format_connections_name(self));
   log_writer_reopen(self->writer, proto);
@@ -657,6 +656,9 @@ afsocket_dd_init(LogPipe *s)
     {
       return FALSE;
     }
+
+  if (!_update_legacy_connection_persist_name(self))
+    return FALSE;
 
   if (!_dd_init_socket(self))
     {
--- a/modules/afsocket/afsocket-dest.c
+++ b/modules/afsocket/afsocket-dest.c
@@ -28,6 +28,8 @@
 #include "stats/stats-registry.h"
 #include "mainloop.h"
 #include "timeutils/misc.h"
+#include "hostname.h"
+#include "persist-state.h"
 
 #include <string.h>
 #include <sys/types.h>
@@ -107,6 +109,20 @@ _get_module_identifier(const AFSocketDestDriver *self)
 }
 
 static const gchar *
+_get_legacy_module_identifier(const AFSocketDestDriver *self)
+{
+  static gchar legacy_module_identifier[128];
+  const gchar *hostname = get_local_hostname_fqdn();
+
+  g_snprintf(legacy_module_identifier, sizeof(legacy_module_identifier), "%s,%s,%s",
+             (self->transport_mapper->sock_type == SOCK_STREAM) ? "stream" : "dgram",
+             afsocket_dd_get_dest_name(self),
+             hostname);
+
+  return legacy_module_identifier;
+}
+
+static const gchar *
 afsocket_dd_format_name(const LogPipe *s)
 {
   const AFSocketDestDriver *self = (const AFSocketDestDriver *)s;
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build FAILURE

modules/afsocket/afsocket-dest.c
```diff
@@ -207,6 +233,46 @@ afsocket_dd_start_reconnect_timer(AFSocketDestDriver *self)
   iv_timer_register(&self->reconnect_timer);
 }
 
+// Returns TRUE in case of success, or there is nothing to do. FALSE will indicate an error.
```
MrAnno: Please remove the comments. All of them can be expressed in code.

modules/afsocket/afsocket-dest.c
```diff
@@ -207,6 +233,46 @@ afsocket_dd_start_reconnect_timer(AFSocketDestDriver *self)
...
+  new_name = afsocket_dd_format_connections_name(self);
+  new_handle = persist_state_lookup_entry(state, new_name, &size, &new_version);
+  if (new_handle)
+    return TRUE;
+
+  // Find the old entry
+  old_name = afsocket_dd_format_legacy_connection_name(self);
+  old_handle = persist_state_lookup_entry(state, old_name, &size, &old_version);
+  if (!old_handle)
+    return TRUE;
```
MrAnno: The first 2 sections can be extracted, so you could write something like this in `afsocket_dd_construct_transport`:
```
if (_need_to_rename_legacy_persist_key(self, state) &&
    !afsocket_dd_rename_connection_persist_name(self, cfg->state))
{
  return FALSE;
}
```
or `_legacy_persist_entry_exists()` if you want to remove the legacy entry after the rename (or anything else :)).

This way, only the third section remains in `rename_connection_persist_name`, which is the real "rename" part.

modules/afsocket/afsocket-dest.c
```diff
@@ -207,6 +233,46 @@ afsocket_dd_start_reconnect_timer(AFSocketDestDriver *self)
...
+  old_value = persist_state_map_entry(state, old_handle);
+  new_value = persist_state_map_entry(state, new_handle);
+  memcpy(new_value, old_value, size);
+  persist_state_unmap_entry(state, old_handle);
+  persist_state_unmap_entry(state, new_handle);
+
+  msg_debug("Connections persist name changed",
+            evt_tag_str("old name", old_name),
+            evt_tag_str("new name", new_name));
+
```
MrAnno: Shouldn't we remove the old entry?

szemere: Not needed. The in_use bit logic will handle the old entry: https://github.com/balabit/syslog-ng/blob/bd6fac1eb892275e81562d93a33bdeb26ee5e833/lib/persist-state.c#L91

Actually the https://github.com/balabit/syslog-ng/blob/bd6fac1eb892275e81562d93a33bdeb26ee5e833/lib/persist-state.c#L760 function is just set this flag, and do not actually remove the entry.

kira-syslogng:
Build SUCCESS

modules/afsocket/afsocket-dest.c
```diff
@@ -207,6 +233,42 @@ afsocket_dd_start_reconnect_timer(AFSocketDestDriver *self)
...
 }
 
+static gboolean
+_afsocket_dd_rename_connection_persist_name(AFSocketDestDriver *self, PersistState *state)
+{
+  const gchar *old_name, *new_name;
+  PersistEntryHandle old_handle, new_handle;
+  gsize size;
+  guint8 old_version, new_version;
+  gpointer old_value, new_value;
```
MrAnno: If my previous suggestion can't be done, and it's not worth extracting anything from this method, then please declare these variables where they initialized. At least this reduces the scope of these 9 variables and makes the function a bit shorter.

modules/afsocket/afsocket-dest.c
```diff
@@ -207,6 +233,42 @@ afsocket_dd_start_reconnect_timer(AFSocketDestDriver *self)
   iv_timer_register(&self->reconnect_timer);
 }
 
+static gboolean
+_afsocket_dd_rename_connection_persist_name(AFSocketDestDriver *self, PersistState *state)
```
MrAnno: If this called unconditionally, I'd consider renaming it to something that has a slight hint of conditionality in it.
`update_legacy_connection_persist_name()`, `rename_legacy_persist_name()`, etc.

modules/afsocket/afsocket-dest.c
szemere: Thank You for the suggestion! I even extracted the actual move part to an inline function, so those variables are local now.

(I found extracting that part a little bit hard, because it will cause lots of output parameters in the extracted function.)

modules/afsocket/afsocket-dest.c
MrAnno: It's hard to implement it this way, marking as resolved.

modules/afsocket/afsocket-dest.c
MrAnno: Thanks.

bazsi:
This pull request **introduces 2 alerts** when merging 93c1f4d4c7c9ddf0c9bbe1d393e681bf3a68165d into 67fe2b0560217f718cb11a92c79671ae9380dfe2 - [view on LGTM.com](https://lgtm.com/projects/g/balabit/syslog-ng/rev/pr-46741a6fc9458454a264eb70c3bcb11c7d987b08)

**new alerts:**

* 2 for Lossy pointer cast

---

**Warning** - Automated code review for balabit/syslog-ng will be disabled on October 1, 2019. You can avoid this by <a href="https://github.com/apps/lgtm-com/installations/new">installing the LGTM.com GitHub App</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

szemere:
Force push: thanks LGTM!

kira-syslogng:
Build SUCCESS

szemere:

pzoleex:

kira-syslogng:
maybe success maybe full

kira-syslogng:
amit csak akarsz :D

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/afsocket/afsocket-dest.c
```diff
@@ -234,6 +303,9 @@ afsocket_dd_connected(AFSocketDestDriver *self)
 
   proto = log_proto_client_factory_construct(self->proto_factory, transport, &self->writer_options.proto_options.super);
 
+  if (!_update_legacy_connection_persist_name(self, cfg->state))
```
MrAnno: Can you move this out from `afsocket_dd_connected()`?
If the connection can't be established during the upgrade, we will lose the old entry (this code never runs and the entry will be deleted becaused it's unused).

kira-syslogng:
Build SUCCESS

modules/afsocket/afsocket-dest.c
szemere: Thank You for pointing it out! I put it after `afsocket_dd_setup_transport` because we need to know if it is a stream or a dgram connection.

kira-syslogng:
Build SUCCESS

modules/afsocket/afsocket-dest.c
```diff
@@ -207,6 +233,49 @@ afsocket_dd_start_reconnect_timer(AFSocketDestDriver *self)
   iv_timer_register(&self->reconnect_timer);
 }
 
+static inline void
+_copy_persist_entry_value(PersistState *state, PersistEntryHandle from, PersistEntryHandle to, gsize size)
```
alltilla: Optional: This would look nice as a PersistState API function, although a lot bigger refactors are needed to fully support the version conversion from PersistState API side...

PR #2931, opened 2019-09-17
WIP: Accept BSD timestamps with the day value not space padded (produced by Cisco ISE)

bazsi:
Cisco ISE devices sometimes don't pad the "day" value in its BSD timestamps using spaces, contrary to the documentation.

https://www.cisco.com/c/en/us/td/docs/security/ise/2-6/Cisco_ISE_Syslogs/Cisco_ISE_Syslogs/Cisco_ISE_Syslogs_chapter_00.html

this causes the timestamp parsing to fail. I am still getting an actual log sample from  @rfaircloth-splunk (lost due to slack dropping history :( but in the meantime I've implemented it.

This branch originally contained a number of improvements that can go in separately, but I've split that off to #2935, which should go in first, so setting this to WIP.

### DIFF
--- a/lib/timeutils/tests/test_scan-timestamp.c
+++ b/lib/timeutils/tests/test_scan-timestamp.c
@@ -133,6 +133,11 @@ Test(parse_timestamp, bsd_extensions)
   _expect_rfc3164_timestamp_eq("Dec  3 2019 09:10:12 ", "2019-12-03T09:10:12.000+01:00");
 }
 
+Test(parse_timestamp, bsd_timestamp_with_missing_padding_in_front_of_day)
+{
+  _expect_rfc3164_timestamp_eq("Dec 3 09:10:12", "2017-12-03T09:10:12.000+01:00");
+}
+
 Test(parse_timestamp, standard_bsd_format_year_in_the_future)
 {
   /* compared to 2017-12-13, this timestamp is from the future, so in the year 2018 */
--- a/lib/timeutils/scan-timestamp.c
+++ b/lib/timeutils/scan-timestamp.c
@@ -138,7 +138,11 @@ scan_month_abbrev(const gchar **buf, gint *left, gint *mon)
 static gboolean
 __is_bsd_rfc_3164(const guchar *src, guint32 left)
 {
-  return left >= 15 && src[3] == ' ' && src[6] == ' ' && src[9] == ':' && src[12] == ':';
+  return
+    /* two character days (e.g. DD is padded with a space or 0 */
+    (left >= 15 && src[3] == ' ' && src[6] == ' ' && src[9] == ':' && src[12] == ':') ||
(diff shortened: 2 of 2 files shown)


bazsi:
Here's an actual sample:

```
<181> Sep 8 18:11:57 ISE-1 CISE_Failed_Attempts 000231543 1 0 2019-09-08 18:11:57.232 -07:00 0003426523 5400 NOTICE Failed-Attempt: Authentication failed, ConfigVersionId=199, Device IP Address=192.168.2.2, Device Port=1645, DestinationIPAddress=192.168.3.2, DestinationPort=1645, RadiusPacketType=AccessRequest, UserName=SEP00AABBCCDDEE, Protocol=Radius, RequestLatency=12, NetworkDeviceName=mySwitch-1, User-Name=CP-7800G-SEP00AABBCCDDEE, NAS-IP-Address=192.168.2.2, NAS-Port=48000, Service-Type=Framed, Framed-MTU=1500
```

kira-syslogng:
Build SUCCESS

bazsi:
Please hold this until a second log sample can be acquired. There was a bit of concern that the pattern we used was not authentic. @rfaircloth-splunk

kira-syslogng:
Build SUCCESS

lib/timeutils/tests/test_scan-timestamp.c
```diff
@@ -293,6 +295,38 @@ Test(parse_timestamp, rfc5424_timestamps)
...
+  gint length = strlen(ts);
+  WallClockTime wct = WALL_CLOCK_TIME_INIT;
+  gint it = 1000000;
+
+  start_stopwatch();
+  for (gint i = 0; i < it; i++)
+    {
+      scan_rfc3164_timestamp(&data, &length, &wct);
+    }
+  stop_stopwatch_and_display_result(it, "RFC3164 timestamp parsing speed");
```
Kokan: What is the purpose of such "test" ? if we could call it a test without any assertion, time limitation.

kira-syslogng:
Build SUCCESS

lib/timeutils/tests/test_scan-timestamp.c
bazsi: This was addressed by excluding these performance tests during unit test invocation using a criterion environment variable. Still, if I (or anyone) runs the test manually, they get executed.

bazsi:
I've just rebased this against master and only a trivial & small patch remained in the branch, the one that enables the functional change.

@rfaircloth-splunk I'd appreciate if you could get a fresh log sample, now that the day is less than 10, e.g. can be represented by a single digit. If the sample proves genuine, we'd merge this patch.

kira-syslogng:
Build SUCCESS

bazsi:
I just received confirmation that Cisco doesn't have the padding issue, so merging this is not needed, closing the pr.

PR #2935, opened 2019-09-20
Timeutils refactor and performance improvements

bazsi:
This PR contains the refactoring/performance improvements pieces of #2931 which we can merge independently of the other, where we are waiting for a 3rd party confirmation of the log format.

This branch also contains the exclusion of performance tests in "make check", reducing the time needed to run make check by 25% on my laptop. ~~This is only done for automake, cmake help would be appreciated.~~

### DIFF
--- a/lib/str-format.c
+++ b/lib/str-format.c
@@ -295,24 +295,36 @@ format_hex_string(gpointer data, gsize data_len, gchar *result, gsize result_len
 gboolean
 scan_positive_int(const gchar **buf, gint *left, gint field_width, gint *num)
 {
+  const gchar *original_buf = *buf;
+  gint original_left = *left;
   guint32 result;
 
   result = 0;
 
-  while (*left > 0 && field_width > 0)
+  while (*left > 0 && field_width > 0 && **buf == ' ')
     {
-      if ((**buf) >= '0' && (**buf) <= '9')
-        result = result * 10 + ((**buf) - '0');
-      else if (!isspace((int) **buf))
-        return FALSE;
+      (*buf)++;
+      (*left)--;
+      field_width--;
+    }
+
+  while (*left > 0 && field_width > 0 &&
+         (**buf) >= '0' && (**buf) <= '9')
+    {
+      result = result * 10 + ((**buf) - '0');
       (*buf)++;
       (*left)--;
       field_width--;
     }
   if (field_width != 0)
-    return FALSE;
+    goto fail;
   *num = result;
   return TRUE;
+
+fail:
+  *buf = original_buf;
+  *left = original_left;
+  return FALSE;
 }
 
 gboolean
--- a/lib/str-format.c
+++ b/lib/str-format.c
@@ -293,7 +293,7 @@ format_hex_string(gpointer data, gsize data_len, gchar *result, gsize result_len
 /* parse 32 bit ints */
 
 gboolean
-scan_uint32(const gchar **buf, gint *left, gint field_width, guint32 *num)
+scan_positive_int(const gchar **buf, gint *left, gint field_width, gint *num)
 {
   guint32 result;
 
@@ -312,17 +312,6 @@ scan_uint32(const gchar **buf, gint *left, gint field_width, guint32 *num)
   if (field_width != 0)
     return FALSE;
   *num = result;
-  return TRUE;
-}
-
-gboolean
-scan_int(const gchar **buf, gint *left, gint field_width, gint *num)
-{
-  guint32 value;
-
-  if (!scan_uint32(buf, left, field_width, &value))
-    return FALSE;
-  *num = (gint) value;
   return TRUE;
 }
 
--- a/lib/tests/test_str_format.c
+++ b/lib/tests/test_str_format.c
@@ -112,6 +112,13 @@ assert_scan_positive_int_in_field_fails(const gchar *input, gint field)
   cr_assert_not(success);
 }
 
+static void
+assert_scan_positive_int_fails(const gchar *input)
+{
+  return assert_scan_positive_int_in_field_fails(input, -1);
+}
+
+
 Test(test_str_format, scan_positive_int)
 {
   assert_scan_positive_int_value_equals("1", 1);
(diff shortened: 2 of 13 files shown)


kira-syslogng:
Build SUCCESS

lib/timeutils/format.c
```diff
@@ -46,13 +46,13 @@ _append_frac_digits(glong usecs, GString *target, gint frac_digits)
     }
 }
 
-static gint
-_format_zone_info(gchar *buf, size_t buflen, glong gmtoff)
+void
+append_format_zone_info(GString *target, glong gmtoff)
```
furiel: Are you sure this is performance wise better? The original one time append is replaced with four appends, which can result in four reallocs. Also, format_uint32_padded contains a lot of function calls and parse logic.

I am asking, because the original code was way more readable.

lib/timeutils/format.c
```diff
@@ -97,7 +97,7 @@ append_format_wall_clock_time(const WallClockTime *wct, GString *target, gint ts
   switch (ts_format)
     {
     case TS_FMT_BSD:
-      g_string_append(target, month_names_abbrev[wct->wct_mon]);
+      g_string_append_len(target, month_names_abbrev[wct->wct_mon], 3);
```
furiel: Here and below you could also use the newly introduced constant names.

lib/str-format.c
```diff
@@ -303,8 +305,8 @@ scan_positive_int(const gchar **buf, gint *left, gint field_width, gint *num)
     {
       if ((**buf) >= '0' && (**buf) <= '9')
         result = result * 10 + ((**buf) - '0');
-      else if (!isspace((int) **buf))
-        return FALSE;
+      else if (!isspace((int) **buf) || result != 0)
```
furiel: Would it be readable to skip spaces first, then error in every nondigit letters? I do not understand how is this works. What is the connection between terminating spaces and result? We want to accept `" 0 "`, and only ignore `" 3 "`?

lib/timeutils/scan-timestamp.c
```diff
@@ -244,7 +244,10 @@ __is_bsd_linksys(const guchar *src, guint32 left)
 {
   /* "MMM DD HH:MM:SS YYYY " */
   return (left >= 21
-          && __is_bsd_rfc_3164(src, left)
+          && src[3] == ' '
+          && src[6] == ' '
```
furiel: Could you elaborate more what is happening here? I do not really understand the commit message. What I can see is a function was inlined and eliminated.

furiel:
@bazsi about the cmake part: I tried this and it worked for me:

In `add_tests.cmake`, there is already one environment variable specified, that you can extend:
From
```
  set_tests_properties(${ADD_UNIT_TEST_TARGET} PROPERTIES ENVIRONMENT "ASAN_OPTIONS=detect_odr_violation=0'")
```
into this
```
  set_tests_properties(${ADD_UNIT_TEST_TARGET} PROPERTIES ENVIRONMENT "ASAN_OPTIONS=detect_odr_violation=0;CRITERION_TEST_PATTERN='!(*/*performance*)'")
```

lib/tests/test_str_format.c
```diff
@@ -70,3 +70,68 @@ Test(test_str_format, hex_string_with_delimiter__two_bytes__perfect)
   cr_expect_str_eq(output, expected_output,
                    "format_hex_string_with_delimiter output does not match in case of two bytes!");
 }
+
+static void
+assert_scan_positive_int_in_field_value_equals(const gchar *input, gint field, gint expected_result)
+{
+  const gchar *buf = input;
+  gint left = strlen(buf);
+  gint result = 99;
```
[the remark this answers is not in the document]
bazsi: I wanted to make sure that result is not accidentaly correct, but it's better to initialize it using expected_result, so that's what I just did.

lib/timeutils/format.c
bazsi: the original was using snprintf() which is _slow_ (for a fastpath), that's why format_uint32_padded() and friends were born in the first place.

Also, GString instances are reallocated in exponential terms, so a single realloc will probably suffice even if the GString instance doesn't have enough space at the start of the function. But it usually would have that space, as at this call we are usually appending the timezone information at the end of the timestamp (19 characters into the string, which would have been set to 32 characters when we were at 16).

As for readability, I agree. snprintf() just doesn't cut it performance wise, at least in our fastpath. If the user sets a ISOTIME in his/her template or uses RFC5424, this can make a difference.

With all that said, I didn't do a detailed micro-benchmark of this change, but as you can see, we are already using format_uint32_padded() functions in the same file for these reasons.

And even if sprintf() has improved since this was implemented, I still would retain to the optimized implementation, simply for platform independence.

lib/timeutils/format.c
bazsi: I remember doing those, that's why I introduced the constants... Ah, I did that in macros.c, good catch.
fixed.

lib/str-format.c
bazsi: I actually tried unrolling the loop and skip initial spaces, but the code got slower in my microbenchmark and not by a bit. We can do ~300m timestamp parsing per second and with unrolling it went down to ~280m

I guess the compiler is too damn smart about optimizing this loop, and it gets worse when I use two separate loops.

My intention was to accept initial whitespace but don't accept whitespace at the end of the field, but you are right that this accepts " 0 ", which is not good. (I was using the "result != 0" condition to see if we already processed digit characters, not wanting to introduce a separate variable for that fact).

Let me see what I can do here.

lib/timeutils/scan-timestamp.c
bazsi: The function was not eliminated as it is retained for BSD log format detection. Here's my rationale in more details:

* although the BSD and LinkSys formats are the same in their initial characters, the *reason* for potentially changing either logic is separate. e.g. either can change without the other having to change.
  - I might want to improve the BSD one (which I do in a followup patch), which I explicitly don't want to get applied to the LinkSys one, because I know that linksys generates formats in a specific way.
  - by separating these the code gets longer (so DRY would apply), but at the same time the two formats are conceptually different, so it is not a good idea to reuse the logic

lib/str-format.c
bazsi: I improved it, made some measurements again and I think it got better. Thanks for pushing this.

kira-syslogng:
Build SUCCESS

lib/str-format.c
furiel: Looks nice, thank you

PR #2940, opened 2019-09-23
dbld: warn about /usr/bin/man AppArmor bug if it is biting us

bazsi:
This branch adds an explicit check to deb builds that /usr/bin/man is operational. On ubuntu a bug in apparmor may cause the man binary not to work (complaining about a missing .so file), in which case deb builds fail.

The check will display an actionable error message to the user how to fix the issue.

### DIFF
--- a/dbld/deb
+++ b/dbld/deb
@@ -22,6 +22,7 @@ function capture_debs() {
 
 cd /build
 
+validate_container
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

PR #2945, opened 2019-09-25
persist-state: add g_assert to persist_state_map_entry

szemere:
We encountered an error during development, where one of the drivers
did not save the allocated persist handle, which resulted calling the
persist_state_map_entry with a 0 value.

Calling map_entry with 0 value, will result overwriting the persist file
header.


kira-syslogng:
Build SUCCESS

PR #2950, opened 2019-09-30
cmake: fix test exclude pattern

Kokan:
The exclude pattern was incorrect in case of CTest/CMake. As CTest sets the environment variable as is, do not remove the additional quoting (like `'..'`).
As a result a lot of test were ignored in case of cmake.


furiel:
Whops sorry. All I was checked that the performance tests were disabled. :-(
Did not check the other tests are not disabled.

kira-syslogng:
Build SUCCESS

PR #2952, opened 2019-10-02
logproto-buffered-server: fix an issue with bookmark/persist handling

szemere:
LogProtoBufferedServer is designed to work without persist state if the
allocation fails. (persist-state will schedule a mainloop exit in this
case, but LPBServer will keep runing until we shut down gracefully)

The statless mode is handled internally by returning a dummy state in
case of allocation failure. There are safeguards in _get_state, and
_put_state to not actually try to "persist_state_map_enty" this value.

However in Bookmark handling we only pass the persist_handle to the
callback function, not the whole LogProtoBufferedServer, and we can not
use our safeguards to prevent 'persist_state_map_entry' that invalid
value. This patch separates that case by using a different callback
function.

### DIFF
--- a/lib/logproto/logproto-buffered-server.c
+++ b/lib/logproto/logproto-buffered-server.c
@@ -61,6 +61,12 @@ log_proto_buffered_server_put_state(LogProtoBufferedServer *self)
     persist_state_unmap_entry(self->persist_state, self->persist_handle);
 }
 
+static inline gboolean
+_log_proto_buffered_server_fallback_non_persistent(LogProtoBufferedServer *self)
+{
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

szemere:
@MrAnno , I did not manage to completely understand it either (yet!). But I totally agree with you, we should not start at all in such a case.* My long term plan is to completely remove this state1 workaround. This is just a quick fix for the issue.

*As I can see, the persist-state is initiating a shutdown in such a case, so working with this fake state is just a temporary operation.

PR #2954, opened 2019-10-02
afsql, afsocket: fix a persist handling issue in init phase

szemere:
As turned out, we can call the drivers init method multiple times
by reloading syslog-ng.

The initial idea was to spare an extra check, by just looking to a
legacy entry during persist name update.
Since we do not actually delete from the persist file (see the in_use
flag for further explanation), we can override the new entry multiple
times, loosing any change in the new value since the last move.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Issue #2924, opened 2019-09-16
implement list search functionality in syslog-ng


jszigetvari:
Currently there is no out of the box solution in syslog-ng for searching in list data structures.
Some list functions are present, but even with them it is pretty awkward to search a specific value in a list.

There already is a list-count() template function that counts the element of the list passed in. (On a sidenote: given that we don't store the list's length anywhere, list-count() has to walk through the whole list each time it is invoked.)
Similarly to list-count() it would be feasible to create a template function to implement searching in a list.
It could take the list, the regex pattern, and possibly a starting position as an argument (and perhaps flags for the type of match for the filter (regex, literal, glob, prefix, substring, etc.), and would return with the index of the list element, where the pattern matches.)
Also, a reverse search functionality would likely be useful, similarly to the forward search described above.

alltilla:
@jszigetvari 

I implemented a new template function, called `list-search()`, which returns the index of the first occurrence of a keyword in a list. It handles the start index, too.

Regarding the reverse or backwards search function: It is harder than it seems. Currently our list implementation is just a well formatted string, plus some convenient APIs which operate on them (by this, I do not mean the template functions, but the `ListScanner` class). We have a `list_scanner_scan_next()`, which in theory could have a `list_scanner_scan_prev()` pair, but it is a bit more complicated. `ListScanner` is based on the fact, that we are going forwards, and we are identifying escaped characters by the `\` character. Walking through the list backwards means going character by character backwards, which is hard because of the backslashes, where we jump backwards 2 characters. It surely can be done, but would take a bigger effort.

I would suggest an alternative for it: What if we had a `list-findall()` template function, which returns a list of indexes. The list would be monotonously increasing. Could that be useful to you? Or should I work on `list_scanner_scan_prev()`?

I did not investigate the type of match for the filter, yet.

jszigetvari:
Hi Attila,

Good work. The list-search() functionality sounds in deed promising.
For my use case at the moment, the reverse search would not be necessary, it's just a type of functionality often enccountered around the topics of searching and arrays.
I know syslog-ng's lists are not arrays, but they are conceptually similar.

The list find-all() could also come in handy, however in that case I would still be faced with the problem that I would have to work with a list which will have an unknown number of elements, which I will have to traverse individually, and syslog-ng not offer any elegant way to iterate through a list at the moment.

For the moment let's forget the use case of backwards search. We can cross that bridge at a later time, if it's more convenient for you this way.

[...]

PR #2955, opened 2019-10-02
basicfuncs: add list-search function

alltilla:
Returns with the index of the first match of `keyword` in the `list`.
Starts the search at `start_index`. 0 based. If not found, returns empty string.

New usage:
 * Literal match:
     `$(list-search <pattern> ${list})`
     or
     `$(list-search --mode literal <pattern> ${list})`
 * Prefix match: `$(list-search --mode prefix <pattern> ${list})`
 * Substring match: `$(list-search --mode substring <pattern> ${list})`
 * Glob match: `$(list-search --mode glob <pattern> ${list})`
 * Regex match: `$(list-search --mode pcre <pattern> ${list})`

Add `--start-index <index>` to change the start index.

Implements #2924

TODO:
- [x] Add starting position
- [x] ~Add backwards search~
- [x] Support regex, glob, etc...
- [x] Support prefix, substring
- [x] Cleanup commits

### DIFF
--- a/modules/basicfuncs/list-funcs.c
+++ b/modules/basicfuncs/list-funcs.c
@@ -332,7 +332,8 @@ TEMPLATE_FUNCTION_SIMPLE(tf_implode);
 typedef enum _StringMatchMode
 {
   SMM_LITERAL = 0,
-  SMM_PREFIX
+  SMM_PREFIX,
+  SMM_SUBSTRING
 } StringMatchMode;
 
 typedef struct _StringMatcher
@@ -360,6 +361,8 @@ string_matcher_match(StringMatcher *self, const char *string, gsize string_len)
       return (strcmp(string, self->pattern) == 0);
     case SMM_PREFIX:
       return (strncmp(string, self->pattern, strlen(self->pattern)) == 0);
+    case SMM_SUBSTRING:
+      return (strstr(string, self->pattern) != NULL);
     default:
       g_assert_not_reached();
     }
@@ -410,6 +413,8 @@ _list_search_mode_str_to_string_match_mode(const gchar *mode_str, StringMatchMod
     *string_match_mode = SMM_LITERAL;
   else if (strcmp(mode_str, "prefix") == 0)
     *string_match_mode = SMM_PREFIX;
+  else if (strcmp(mode_str, "substring") == 0)
+    *string_match_mode = SMM_SUBSTRING;
   else
     result = FALSE;
 
@@ -441,7 +446,7 @@ _list_search_parse_options(StringMatchMode *mode, gint *start_index, gint *argc,
     {
       g_set_error(error, LOG_TEMPLATE_ERROR, LOG_TEMPLATE_ERROR_COMPILE,
                   "$(list-search) Invalid list-search mode: %s. "
-                  "Valid modes are: literal, prefix", mode_str);
+                  "Valid modes are: literal, prefix, substring", mode_str);
       goto exit;
     }
 
--- a/modules/basicfuncs/list-funcs.c
+++ b/modules/basicfuncs/list-funcs.c
@@ -23,6 +23,7 @@
 
 #include "scanner/list-scanner/list-scanner.h"
 #include "str-repr/encode.h"
+#include "compat/pcre.h"
 
 static void
 _append_comma_between_list_elements_if_needed(GString *result, gsize initial_len)
@@ -334,7 +335,8 @@ typedef enum _StringMatchMode
   SMM_LITERAL = 0,
   SMM_PREFIX,
   SMM_SUBSTRING,
-  SMM_GLOB
+  SMM_GLOB,
+  SMM_PCRE
 } StringMatchMode;
 
 typedef struct _StringMatcher
@@ -342,6 +344,8 @@ typedef struct _StringMatcher
   StringMatchMode mode;
   gchar *pattern;
   GPatternSpec *glob;
+  pcre *pcre;
+  pcre_extra *pcre_extra;
 } StringMatcher;
 
 static gboolean
@@ -353,15 +357,66 @@ string_matcher_prepare_glob(StringMatcher *self)
 }
 
 static gboolean
+string_matcher_prepare_pcre(StringMatcher *self)
+{
+  const gchar *errptr;
+  gint erroffset;
+  gint rc;
+
+  self->pcre = pcre_compile2(self->pattern, PCRE_ANCHORED, &rc, &errptr, &erroffset, NULL);
+  if (!self->pcre)
+    {
+      msg_error("Error while compiling regular expression",
+                evt_tag_str("regular_expression", self->pattern),
+                evt_tag_str("error_at", &self->pattern[erroffset]),
+                evt_tag_int("error_offset", erroffset),
+                evt_tag_str("error_message", errptr),
+                evt_tag_int("error_code", rc));
+      return FALSE;
+    }
+  self->pcre_extra = pcre_study(self->pcre, PCRE_STUDY_JIT_COMPILE, &errptr);
+  if (errptr)
+    {
+      msg_error("Error while optimizing regular expression",
+                evt_tag_str("regular_expression", self->pattern),
+                evt_tag_str("error_message", errptr));
+      pcre_free(self->pcre);
+      if (self->pcre_extra)
+        pcre_free_study(self->pcre_extra);
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
 string_matcher_prepare(StringMatcher *self)
 {
   switch (self->mode)
     {
     case SMM_GLOB:
       return string_matcher_prepare_glob(self);
+    case SMM_PCRE:
+      return string_matcher_prepare_pcre(self);
     default:
       return TRUE;
     }
+}
+
+static gboolean
+string_matcher_match_pcre(StringMatcher *self, const char *string, gsize string_len)
+{
+  gint rc = pcre_exec(self->pcre, self->pcre_extra, string, string_len, 0, 0, NULL, 0);
+  if (rc == PCRE_ERROR_NOMATCH)
+    {
+      return FALSE;
+    }
+  if (rc < 0)
+    {
+      msg_error("Error while matching pcre", evt_tag_int("error_code", rc));
+      return FALSE;
+    }
+  return TRUE;
 }
 
 static gboolean
@@ -377,6 +432,8 @@ string_matcher_match(StringMatcher *self, const char *string, gsize string_len)
       return (strstr(string, self->pattern) != NULL);
     case SMM_GLOB:
       return (g_pattern_match_string(self->glob, string));
+    case SMM_PCRE:
+      return (string_matcher_match_pcre(self, string, string_len));
     default:
       g_assert_not_reached();
     }
@@ -400,6 +457,10 @@ string_matcher_free(StringMatcher *self)
     g_free(self->pattern);
   if (self->glob)
     g_pattern_spec_free(self->glob);
+  if (self->pcre)
+    pcre_free(self->pcre);
+  if (self->pcre_extra)
+    pcre_free_study(self->pcre_extra);
   g_free(self);
 }
 
@@ -433,6 +494,8 @@ _list_search_mode_str_to_string_match_mode(const gchar *mode_str, StringMatchMod
     *string_match_mode = SMM_SUBSTRING;
   else if (strcmp(mode_str, "glob") == 0)
     *string_match_mode = SMM_GLOB;
+  else if (strcmp(mode_str, "pcre") == 0)
+    *string_match_mode = SMM_PCRE;
   else
     result = FALSE;
 
@@ -464,7 +527,7 @@ _list_search_parse_options(StringMatchMode *mode, gint *start_index, gint *argc,
     {
       g_set_error(error, LOG_TEMPLATE_ERROR, LOG_TEMPLATE_ERROR_COMPILE,
                   "$(list-search) Invalid list-search mode: %s. "
-                  "Valid modes are: literal, prefix, substring, glob", mode_str);
+                  "Valid modes are: literal, prefix, substring, glob, pcre", mode_str);
       goto exit;
     }
 
--- a/modules/basicfuncs/list-funcs.c
+++ b/modules/basicfuncs/list-funcs.c
@@ -333,20 +333,32 @@ typedef enum _StringMatchMode
 {
   SMM_LITERAL = 0,
   SMM_PREFIX,
-  SMM_SUBSTRING
+  SMM_SUBSTRING,
+  SMM_GLOB
 } StringMatchMode;
 
 typedef struct _StringMatcher
 {
   StringMatchMode mode;
   gchar *pattern;
+  GPatternSpec *glob;
 } StringMatcher;
+
+static gboolean
+string_matcher_prepare_glob(StringMatcher *self)
+{
+  self->glob = g_pattern_spec_new(self->pattern);
+
+  return TRUE;
+}
 
 static gboolean
 string_matcher_prepare(StringMatcher *self)
 {
   switch (self->mode)
     {
+    case SMM_GLOB:
+      return string_matcher_prepare_glob(self);
     default:
       return TRUE;
     }
@@ -363,6 +375,8 @@ string_matcher_match(StringMatcher *self, const char *string, gsize string_len)
       return (strncmp(string, self->pattern, strlen(self->pattern)) == 0);
     case SMM_SUBSTRING:
       return (strstr(string, self->pattern) != NULL);
+    case SMM_GLOB:
+      return (g_pattern_match_string(self->glob, string));
     default:
       g_assert_not_reached();
     }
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/basicfuncs/list-funcs.c
```diff
@@ -24,6 +24,8 @@
 #include "scanner/list-scanner/list-scanner.h"
 #include "str-repr/encode.h"
 
+#include <pcre.h>
```
Kokan: Please use `compat/pcre.h` instead, that covers `PCRE_STUDY_JIT_COMPILE` and others.

modules/basicfuncs/list-funcs.c
```diff
@@ -70,6 +72,278 @@ tf_list_append(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+  GString *keyword;
+  GPatternSpec *glob;
+  GString *keyword_reversed;
+  pcre *regex;
+  pcre_extra *regex_extra;
+} LSPattern;
+
+enum ListSearchMode
+{
+  LSM_LITERAL = 0,
```
Kokan: It feels like you reimplement the `LogMatcher` just working with list instead of `LogMessage`.
[optional] would you take a look if the two could have the same root ?

modules/basicfuncs/tests/test_basicfuncs.c
```diff
@@ -367,6 +367,52 @@ Test(basicfuncs, test_list_funcs)
   assert_template_format("$(list-append foo,bar,baz 'xxx,')", "foo,bar,baz,\"xxx,\"");
   assert_template_format("$(list-append foo,bar,baz 'a\tb')", "foo,bar,baz,\"a\\tb\"");
 
+  assert_template_format("$(list-search)", "");
+  assert_template_format("$(list-search almafa)", "");
+  assert_template_format("$(list-search almafa 0)", "");
```
Kokan: I would consider those as incorrect call, as searching nothing sure could mean nothing, but most likely an user error this is.

modules/basicfuncs/list-funcs.c
```diff
@@ -70,6 +72,300 @@ tf_list_append(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+    {
+      msg_error("Error while compiling regular expression",
+                evt_tag_str("regular_expression", self->keyword->str),
+                evt_tag_str("error_at", &self->keyword->str[erroffset]),
+                evt_tag_int("error_offset", erroffset),
+                evt_tag_str("error_message", errptr),
+                evt_tag_int("error_code", rc));
+      return FALSE;
+    }
+  self->pcre_extra = pcre_study(self->pcre, 0, &errptr);
```
Kokan: Please use `PCRE_STUDY_JIT_COMPILE` to enable JIT whenever possible.

modules/basicfuncs/list-funcs.c
```diff
@@ -70,6 +72,300 @@ tf_list_append(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+
+  ListScanner scanner;
+  gint index = -1;
+
+  list_scanner_init(&scanner);
+  list_scanner_input_gstring_array(&scanner, argc, argv);
+
+  while (list_scanner_scan_next(&scanner))
+    {
+      if (++index < parameters->start_index)
```
Kokan: [optional] you could extract the skip part into a separate while wrapped with a function calling it like skip until start index.

modules/basicfuncs/list-funcs.c
```diff
@@ -70,6 +72,300 @@ tf_list_append(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+        continue;
+
+      if (_list_search_match(list_scanner_get_current_value_as_gstring(&scanner), parameters))
```
Kokan: I am not fun of the `get_as_type` like functions. In this case I think the current `list_scanner_get_current_value` could return with `GString`, and the users could be updated accordingly.

kira-syslogng:
Build SUCCESS

bazsi:
On Fri, Oct 4, 2019, 11:55 Kókai Péter <notifications@github.com> wrote:

> *@Kokan* commented on this pull request.
>
> In modules/basicfuncs/list-funcs.c
> <https://github.com/syslog-ng/syslog-ng/pull/2955#discussion_r331407452>:
>
> > @@ -24,6 +24,8 @@
>  #include "scanner/list-scanner/list-scanner.h"
>  #include "str-repr/encode.h"
>
> +#include <pcre.h>
>
> Please use compat/pcre.h instead, that covers PCRE_STUDY_JIT_COMPILE and
> others.
>
> In modules/basicfuncs/list-funcs.c
> <https://github.com/syslog-ng/syslog-ng/pull/2955#discussion_r331409292>:
>
> > @@ -70,6 +72,278 @@ tf_list_append(LogMessage *msg, gint argc, GString *argv[], GString *result)
>
>  TEMPLATE_FUNCTION_SIMPLE(tf_list_append);
>
> +typedef struct _LSPattern
> +{
> +  GString *keyword;
> +  GPatternSpec *glob;
> +  GString *keyword_reversed;
> +  pcre *regex;
> +  pcre_extra *regex_extra;
> +} LSPattern;
> +
> +enum ListSearchMode
> +{
> +  LSM_LITERAL = 0,
>
> It feels like you reimplement the LogMatcher just working with list
> instead of LogMessage.
> [optional] would you take a look if the two could have the same root ?
>
I am all for this, just a note: LogMatcher is not a good base for this, I
hate that code. So the new base layer should be created and then an
improved LogMatcher layer put on top.


>
> In modules/basicfuncs/tests/test_basicfuncs.c
> <https://github.com/syslog-ng/syslog-ng/pull/2955#discussion_r331411148>:
>
> > +  assert_template_format("$(list-search)", "");
> +  assert_template_format("$(list-search almafa)", "");
> +  assert_template_format("$(list-search almafa 0)", "");
>
> I would consider those as incorrect call, as searching nothing sure could
> mean nothing, but most likely an user error this is.
>
The template function infrastructure has a distinction over whether the
user passed an argument which is empty over the user didnt pass an argument
(via argc which denotes the number of arguments). I agree that missed
arguments should be reported.


kira-syslogng:
Build SUCCESS

modules/basicfuncs/list-funcs.c
alltilla: Thanks, done!

modules/basicfuncs/tests/test_basicfuncs.c
alltilla: Thanks for the heads-up! I added error messages, and removed the invalid calls from here.

modules/basicfuncs/list-funcs.c
alltilla: Thanks, I added it to the `ListScanner` API.

modules/basicfuncs/list-funcs.c
alltilla: I went the other way, and converted `match()` to handle `char *` and `gsize` instead of `GString*`.

modules/basicfuncs/list-funcs.c
alltilla: Thanks, done!

alltilla:
@Kokan, @bazsi:
Yeah, this is I believe the 3rd implementation of `pcre` now :/
Should I do the refactor in this PR, or before/after it?

kira-syslogng:
Build SUCCESS

Kokan:
Because that refactor requires to change the `LogMatcher` that has effect on many modules/functionality; and the logic can be cut off from this change request I would do it in a different PR.
The current PR is imho okay to merge, as the `list-search` implementation only effects the new code.
You could resolve the "conflict" when one of them is merged.

Short: do it "later" / in parallel.

But I am fine with whatever schedule as well.

modules/basicfuncs/list-funcs.c
alltilla: I agree. I will work on the refactor by a new PR.

modules/basicfuncs/list-funcs.c
```diff
@@ -92,13 +96,48 @@ list_search_pattern_prepare_glob(ListSearchPattern *self)
...
+
 static gboolean
 list_search_pattern_prepare(ListSearchPattern *self, ListSearchMode mode)
 {
   switch (mode)
     {
     case LSM_GLOB:
       return list_search_pattern_prepare_glob(self);
+    case LSM_PCRE:
+      return list_search_pattern_prepare_pcre(self);
```
gaborznagy: After glob and pcre is introduced I think `$(list-search ) should be a `TEMPLATE_FUNCTION` instead of a `TEMPLATE_FUNCTION_SIMPLE`.

It can have a state where things that need one-time initialization (`pcre_compile`)  can be initialized.

alltilla: Thanks, really good catch! Done :)

modules/basicfuncs/list-funcs.c
```diff
@@ -70,9 +70,188 @@ tf_list_append(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+{
+  ListSearchMode mode;
+  gint64 start_index;
+  ListSearchPattern *pattern;
+} ListSearchParameters;
+
+static void
+_step_args(gint *argc, GString **argv[])
+{
+  (*argc)--;
```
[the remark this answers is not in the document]
alltilla: Didn't have time to change this :( Can we merge it this way?

modules/basicfuncs/list-funcs.c
```diff
@@ -70,9 +70,188 @@ tf_list_append(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+  self->pattern->keyword = (*argv)[0];
+  _step_args(argc, argv);
+
+  return TRUE;
+}
+
+static gboolean
+list_search_parameters_parse_start_index(ListSearchParameters *self, gint *argc, GString **argv[])
+{
+  if (*argc < 1)
```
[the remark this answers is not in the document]
alltilla: No, not really. If we do it the `GOptionContext` way, it can be easily done.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/basicfuncs/list-funcs.c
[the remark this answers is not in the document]
alltilla: Nevermind, I did it :D

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/basicfuncs/list-funcs.c
```diff
@@ -329,25 +329,223 @@ tf_implode(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+  state->matcher = string_matcher_new(mode, pattern);
+
+  if (!string_matcher_prepare(state->matcher))
+    {
+      g_set_error(error, LOG_TEMPLATE_ERROR, LOG_TEMPLATE_ERROR_COMPILE,
+                  "$(list-search) Failed to prepare pattern: %s", pattern);
+      return FALSE;
+    }
+
+  state->args = _char_p_array_to_str_slist(argc-2, argv+2);
```
gaborznagy: I think this is not needed, and template resolving is not working.
You are saving the arguments into state, while those should be accessed in the `call` function.

What you need to do is
- derive `ListSearchState` from `TFSimpleFuncState`,
- use `tf_simple_func_prepare()` in prepare function,
- use `tf_simple_func_free_state()`
- add eval function
```
-TEMPLATE_FUNCTION(ListSearchState, tf_list_search, tf_list_search_prepare, NULL,
+TEMPLATE_FUNCTION(ListSearchState, tf_list_search, tf_list_search_prepare, tf_simple_func_eval,
```

modules/basicfuncs/list-funcs.c
```diff
@@ -329,25 +329,223 @@ tf_implode(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
 static void
 tf_list_search_call(LogTemplateFunction *self, gpointer s, const LogTemplateInvokeArgs *args, GString *result)
 {
+  ListSearchState *state = (ListSearchState *)s;
+  ListScanner scanner;
+  gint index = state->start_index;
+
+  list_scanner_init(&scanner);
+  g_slist_foreach(state->args, _load_arg, &scanner);
+  list_scanner_skip_n(&scanner, index);
```
gaborznagy: These are not necessary, and unfortunately not working.

Not necessary to load arguments as using tf simple functions prior to call, you will have evaluated template results in `args`.

Use `args->argv` as input to `list_scanner_input_string`. 
I guess only 1 list is supported as input, so a simple call to `list_scanner_input_string` would be enough and no need for `_load_args()`

modules/basicfuncs/list-funcs.c
alltilla: Really nice catch thanks! :)

modules/basicfuncs/list-funcs.c
alltilla: Thanks, fixed :)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/basicfuncs/list-funcs.c
```diff
@@ -328,3 +329,305 @@ tf_implode(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+
+typedef struct _StringMatcher
+{
+  StringMatchMode mode;
+  gchar *pattern;
+  GPatternSpec *glob;
+  pcre *pcre;
+  pcre_extra *pcre_extra;
+} StringMatcher;
+
```
bazsi: I think we should move this to lib instead of keeping them in this module. It pretty much has everything to be a module on its own right. Later we can work on merging the common pieces between this and the logmatcher module.

modules/basicfuncs/list-funcs.c
```diff
@@ -328,3 +329,305 @@ tf_implode(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+    *string_match_mode = SMM_LITERAL;
+  else if (strcmp(mode_str, "prefix") == 0)
+    *string_match_mode = SMM_PREFIX;
+  else if (strcmp(mode_str, "substring") == 0)
+    *string_match_mode = SMM_SUBSTRING;
+  else if (strcmp(mode_str, "glob") == 0)
+    *string_match_mode = SMM_GLOB;
+  else if (strcmp(mode_str, "pcre") == 0)
+    *string_match_mode = SMM_PCRE;
+  else
```
bazsi: this functionality (to map the string to the SMM_ enum), could be moved alongside the StringMatcher interface as it could be useful for other uses of the interface.

modules/basicfuncs/list-funcs.c
alltilla: Actually I already have a working-branch doing this with a bit of cleanup. I would not like to include it in this PR, as it is old enough, and was reviewed many times. The refactor will come soon :)

The refactored API looks like this currently (changes might be needed, but I think this should be in a different PR's scope):
```
typedef struct _StringMatcher StringMatcher;

gboolean string_matcher_prepare(StringMatcher *self, gpointer user_data);
gboolean string_matcher_match(StringMatcher *self, const char *string, gsize string_len);
void string_matcher_free(StringMatcher *self);

StringMatcher *string_matcher_literal_new(const gchar *pattern);
StringMatcher *string_matcher_prefix_new(const gchar *pattern);
StringMatcher *string_matcher_substring_new(const gchar *pattern);
StringMatcher *string_matcher_glob_new(const gchar *pattern);
StringMatcher *string_matcher_pcre_new(const gchar *pattern);
```

modules/basicfuncs/list-funcs.c
alltilla: Thanks for the tip, as I mentioned in the other comment, I am preparing another PR for this :)

kira-syslogng:
Build SUCCESS

bazsi:
Right now it is a convention that list handling functions concatenate
arguments as separate lists, which means you dont need to use list-append.

modules/basicfuncs/list-funcs.c
```diff
@@ -329,26 +329,198 @@ tf_implode(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+static gboolean
+_list_search_init_matcher(ListSearchState *state, StringMatchMode mode, gint argc, gchar *argv[], GError **error)
+{
+  if (argc < 2)
+    {
+      g_set_error(error, LOG_TEMPLATE_ERROR, LOG_TEMPLATE_ERROR_COMPILE,
+                  "$(list-search) Pattern is missing. Usage: $(list-search [options] <pattern> ${list})");
+      return FALSE;
+    }
+  else if (argc < 3)
```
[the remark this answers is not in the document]
alltilla: The function now accepts any number of arguments and handles them as lists concatenated.

kira-syslogng:
Build SUCCESS

PR #2961, opened 2019-10-05
scl/default-network-drivers: add max-connections

uvNikita:
Allows to passthrough max-connections parameter to the network tcp source.
The default value of 10 connections is very limiting.

### DIFF
--- a/scl/default-network-drivers/plugin.conf
+++ b/scl/default-network-drivers/plugin.conf
@@ -24,6 +24,7 @@
 block source default-network-drivers(
 	udp-port(514)
 	tcp-port(514)
+	max-connections(10)
 	rfc5424-tls-port(6514)
 	rfc5424-tcp-port(601)
 	log-msg-size(65536)
@@ -36,6 +37,7 @@ block source default-network-drivers(
 		source {
 			network(transport(tcp)
 				port(`tcp-port`)
+				max-connections(`max-connections`)
 				log-msg-size(`log-msg-size`)
 				flags(no-parse, `flags`)
 				hook-commands(`hook-commands`));
@@ -63,10 +65,12 @@ block source default-network-drivers(
 		source {
 			syslog(transport(tls) tls(`tls`)
 			       port(`rfc5424-tls-port`)
+			       max-connections(`max-connections`)
 			       flags(`flags`)
 	                       log-msg-size(`log-msg-size`));
 			syslog(transport(tcp)
 			       port(`rfc5424-tcp-port`)
+			       max-connections(`max-connections`)
 			       flags(`flags`)
 	                       log-msg-size(`log-msg-size`));
 		};


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:

MrAnno:
@uvNikita Thanks for your contribution!

Just a few notes that we should discuss during the review:
- `max-connections` is used to calculate the per-connection window size,  a value > 10 will produce a warning:
  > Window sizing for tcp sources were changed in syslog-ng 3.3, the configuration value was divided by the value of max-connections(). The result was too small, clamping to value of min_iw_size_per_reader.

   I think this is OK, it's not worth adding a `log-iw-size()` option to the default network drivers SCL.


- `max-connections` should be configurable for `rfc5424-tls` and `rfc5424-tcp` as well:
  A single common `max-connections()` option may be enough, but separate parameters such as `rfc5424-tls-max-connections()` and `rfc5424-tcp-max-connections()` sound good as well.

kira-syslogng:
Build SUCCESS

bazsi:
I would use a single limit and document that this potentially uses 4x the
connections.

uvNikita:
@MrAnno, @bazsi thanks for such a quick reply!

I added `max-connections` parameter to `rfc5424-tls` and `rfc5424-tcp` as you pointed out.

As for documenting the total number of connections, where should I put it? Just as a comment above the `default-network-drivers` definition?

scl/default-network-drivers/plugin.conf
```diff
@@ -24,6 +24,7 @@
 block source default-network-drivers(
 	udp-port(514)
 	tcp-port(514)
+	tcp-max-connections(10)
```
MrAnno: Please rename it to `max-connections`, it's not only for tcp now.

MrAnno:
> As for documenting the total number of connections, where should I put it? Just as a comment above the default-network-drivers definition?

The documentation is in a separate repo, we'll update that.
@fekete-robert

MrAnno:
@fekete-robert `default-network-drivers()` will soon have a `max-connections()` option. The SCL can use up to `3 * max-connections()` number of connections in total:

`max-connections()` for each `network(tcp)`, `syslog(tcp)`, and `syslog(tls)`.

scl/default-network-drivers/plugin.conf
uvNikita: Done :)

uvNikita:
I just realized I forgot to change the commit message to reflect the new parameter name. Fixed now.

MrAnno:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

fekete-robert:
> @fekete-robert `default-network-drivers()` will soon have a `max-connections()` option. The SCL can use up to `3 * max-connections()` number of connections in total:
> 
> `max-connections()` for each `network(tcp)`, `syslog(tcp)`, and `syslog(tls)`.

Thanks, I've added it to our 3.25 todo list

PR #2966, opened 2019-10-08
WIP: block-name: First implementation of feature #2964

### LINKED ISSUE #2964: Rotate log files based on size
## Description of the problem
When using the logrotate with common configuration, log entries can be lost between the time in which the logrotate copies the file and the truncate of the file.

## Proposed solution
I suggest to rotate the log within syslog-ng based in a file size limit configuration parameter. Additionally one parameter with the desired suffix can be added.
I try was already done in https://github.com/pld-linux/syslog-ng/blob/master/log-file-size-limit.patch

## Alternatives
Add a disk buffer and reliable properties in syslog-ng configuration for each log file, and configure in logrotate the restart of syslog-ng in the end of each rotation. This solution has the disadvantage of periodically have to restart the syslog-ng. Potentially it can be too many restarts of syslog-ng depending on number of logs, size limits, and events recorded.

## Additional context
Already have a patch ready

amdrsantos:
Rotation of log files when the file reaches a size limit.

Fixes #2964 

Signed-off-by: Alexandre Santos <alexandre.rosas.santos@gmail.com>

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -249,6 +249,8 @@ extern struct _LogRewrite *last_rewrite;
 
 %token KW_PERSIST_NAME                10302
 
+%token KW_FILE_SIZE_LIMIT             10303
+
 %token KW_READ_OLD_RECORDS            10304
 
 /* log statement options */
@@ -1006,6 +1008,7 @@ options_item
 	| KW_CUSTOM_DOMAIN '(' string ')'	{ configuration->custom_domain = g_strdup($3); free($3); }
 	| KW_FILE_TEMPLATE '(' string ')'	{ configuration->file_template_name = g_strdup($3); free($3); }
 	| KW_PROTO_TEMPLATE '(' string ')'	{ configuration->proto_template_name = g_strdup($3); free($3); }
+	| KW_FILE_SIZE_LIMIT '(' nonnegative_integer ')'  { configuration->file_size_limit = $3; }
 	| KW_RECV_TIME_ZONE '(' string ')'	{ configuration->recv_time_zone = g_strdup($3); free($3); }
 	| KW_MIN_IW_SIZE_PER_READER '(' positive_integer ')' { configuration->min_iw_size_per_reader = $3; }
 	| { last_template_options = &configuration->template_options; } template_option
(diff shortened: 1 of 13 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

szemere:
ok to test

bazsi:
The reason we didn't integrate similar functionality in the past was that the rotation is a pretty complicated topic that I would prefer to keep out of the file driver code. However I find out of the box support for this kind of use case valueable.

The way this should be solved in my mind is to introduce driver specific events

amdrsantos:
Ok. I will try to work in a plugin.

Can you point me some example in the code?

bazsi:
There's no example right now as the framework doesn't exist yet. 

The outline of the framework would be something like this:
* We will need a way to represent a hook function (function pointer plus name-value pairs as arguments)
* We will need a way to represent registrations against those hooks (like a hash table, keyed with the hook's name, containing a list of hook functions to be invoked) this would be associated with the LogDriver instance that emit the events
* we can probably use the inner-dest/inner-source plugin interfaces to instantiate the plugins that receive the events and where we can do the actual registrations against the hooks
* We will need an API for doing the registrations that can validate that a specific hook exists for a specific plugin and another API for doing a hook invocation.
* Once we have this, we only need to add hook invocations to the file destination in case of file size rollovers and another plugin to receive the event (which will also do the rotation).
* I would also extend hook commands to allow registration to any existing hook to invoke a command.

At the end I would check if:
* Acquire_queue mechanism could use the same or not.
* The current overwrite_if_older mechanism could use this or not

alltilla:
@amdrsantos 
Hi Alexandre,

Did you have any progress with the PR since your last comment, are you planning to work on it? Can we give you any help with it?

Attila

amdrsantos:
Hi Attila,

Right now I do not have the time to spend in this development.
The log rotation is implemented using the program() destination option of
syslog-ng.

The script implementing this functionality receive as parameter: the log
file name, the maximum log file size and the maximum of log files it can
get. It is running in daemon mode following the sequence:

1. Is actively listening its stdin;
2. Receives the message and writes to the correspondent log file;
3. Checks if the size of log file is greater than the specified MAX_SIZE
4. If 3. Is true,
       a. Generates a temporary logrotate configuration file for the
correspondent log file
       b.Calls logrotate with the generated logrotate configuration file
       c. Deletes the temporary logrotate configuration file
5. Go to step 1

I will try to come back to the improvement in syslog-ng as soon as I can.

Cheers,
Alex

Kokan:
@amdrsantos Closing due to your response not having time to deal with this, and @bazsi comment about different approach to reach the final goal.

bazsi:
The framework was created and incorporated into http (the signal-slot mechanism), so this should be much easier to implement now.

PR #2969, opened 2019-10-12
Few undefined behaviour issue in code

Kokan:
Compiling with `-fsanitize=undefined` and executing the UT test set reports these issues (well others as well, but most of them are related to nvtables)

- [ ] check zoneinfo specification for size limits (connected to unsigned/signed question)

### DIFF
--- a/lib/timeutils/zoneinfo.c
+++ b/lib/timeutils/zoneinfo.c
@@ -116,7 +116,7 @@ readcoded32(unsigned char **input, gint64 minv, gint64 maxv)
   *input += 4;
 
   for (i = 0, shift = 24; i < 4; ++i, shift -= 8)
-    val |= buf[i] << shift;
+    val |= (gint64)(buf[i]) << shift;
 
   if (val < minv || val > maxv)
     {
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

lib/timeutils/zoneinfo.c
```diff
@@ -116,7 +116,7 @@ readcoded32(unsigned char **input, gint64 minv, gint64 maxv)
   *input += 4;
 
   for (i = 0, shift = 24; i < 4; ++i, shift -= 8)
-    val |= buf[i] << shift;
+    val |= (gint64)(buf[i]) << shift;
```
MrAnno: Shifting a signed value can also be undefined if you reach the sign bit. Can this shift be done on an unsigned type instead?

Kokan: In this case it cannot (reading 32 bit). But using unsigned might not hurt, I'll check it.

MrAnno: Let's not block on this.  `shift` is less then 64bit; `buf[i]` is originally unsigned ( the value on the left-hand side is guaranteed to be positive), so this should be correct.

PR #2970, opened 2019-10-13
dbld: fix OS_PLATFORM value in devshell images

bazsi:
docker seems to prefer the ENV value inherited from the base image (specified
using FROM) over the value in the command line, thus even though
dbld/rules specifies "--build-arg OS_PLATFORM", its value does not
make it to the Dockerfile.

Since the devshell is special in any case (using the bionic image as its
base), override OS_PLATFORM via a hard-wired value, fixing the package
installation.

This patch also removes redundant ENV/LABEL commands from the Dockerfile, as
all of those are already inherited from the parent image.


bazsi:
Originally reported and solution suggested by @szemere

kira-syslogng:
Build FAILURE

szemere:

kira-syslogng:
Build SUCCESS

szemere:
@mitzkia , looks like we have a non deterministic test-case. I saved the logs for further analysis.

PR #2971, opened 2019-10-13
Fix compat with older systems

bazsi:
This is the generic portions of #2860 that omits the centos6 related images, but still includes the code changes that should go in regardless. I am opening this to make this easier to merge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,11 +101,12 @@ matrix:
         - make install
         - . scripts/get-libjvm-path.sh;
         - export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$JNI_LIBDIR";
-        - make func-test V=1
-        - make python-pep8 python-pylint
-        - make pytest-linters
         - make pytest-self-check
-        - make pytest-check;
+        - make python-pep8
+        - make python-pylint
+        - make pytest-linters
+        - make VERBOSE=1 func-test
+        - make pytest-check
 
     - env: B=style-check
       dist: trusty
@@ -158,9 +159,11 @@ matrix:
             -DCMAKE_INSTALL_PREFIX=$HOME/install/syslog-ng
             ..
         - make --keep-going -j $(nproc) all check install
-        - make VERBOSE=1 func-test
-        - make pytest-linters
         - make pytest-self-check
+        - make python-pep8
+        - make python-pylint
+        - make pytest-linters
+        - make VERBOSE=1 func-test
         - make pytest-check
 
     - env: B=cmake
--- a/.travis.yml
+++ b/.travis.yml
@@ -157,8 +157,7 @@ matrix:
             -DPYTHON_VERSION=2
             -DCMAKE_INSTALL_PREFIX=$HOME/install/syslog-ng
             ..
-        - make --keep-going -j $(nproc) all install
-        - ctest -j $(nproc) --output-on-failure
+        - make --keep-going -j $(nproc) all check install
         - make VERBOSE=1 func-test
         - make pytest-linters
         - make pytest-self-check
@@ -178,8 +177,7 @@ matrix:
             -DPYTHON_VERSION=2
             -DCMAKE_INSTALL_PREFIX=$HOME/install/syslog-ng
             ..
-        - make --keep-going -j $(nproc) all install
-        - ctest -j $(nproc) --output-on-failure
+        - make --keep-going -j $(nproc) all check install
 
     - env: B=check
       os: osx
--- a/lib/compat/glib.h
+++ b/lib/compat/glib.h
@@ -50,29 +50,29 @@ gchar *g_canonicalize_filename (const gchar *filename,
 #define g_atomic_pointer_add(atomic, val) \
   (G_GNUC_EXTENSION ({                                                       \
     G_STATIC_ASSERT (sizeof *(atomic) == sizeof (gpointer));                 \
-    (void) (0 ? (gpointer) *(atomic) : 0);                                   \
-    (void) (0 ? (val) ^ (val) : 0);                                          \
+    (void) (0 ? (gpointer) *(atomic) : NULL);                                \
+    (void) (0 ? (val) ^ (val) : 1);                                          \
     (gssize) __sync_fetch_and_add ((atomic), (val));                         \
   }))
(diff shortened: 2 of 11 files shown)


kira-syslogng:
Build SUCCESS

.travis.yml
```diff
@@ -157,8 +157,7 @@ matrix:
             -DPYTHON_VERSION=2
             -DCMAKE_INSTALL_PREFIX=$HOME/install/syslog-ng
             ..
-        - make --keep-going -j $(nproc) all install
-        - ctest -j $(nproc) --output-on-failure
+        - make --keep-going -j $(nproc) all check install
```
Kokan: > 1) only build test programs in case we are running tests, and to couple
   rebuilding of the test programs with running them.

I could agree that there are use cases where one do not want to compile UT, but I know only those cases where also the user do not want to execute them either. (completely disabling the tests are fine and there is a cmake way of doing that: `BUILD_TESTING`).
Please if you have other use case than that please share.
Also the link mention something like this: `This is IMHO a very nice feature, because the normal build is not bogged down by the compilation of many test programs, or worse, comes to a grinding halt due to a compilation error in one of the test programs you didn't have time to fix yet. `
Well, I hope people does not feel "bogged" about not compiling test ...

> 2) to reference the unit test phase as a DEPENDS in add_custom_target(), as
   the "test" target is special, whereas "check" is a normal target.

This is the same reason as (1).

> 3) to remove pep8/pylint from unit test execution in a subsequent patch

Those can be removed with the exact same way as in your current branch, without this one. (they are not really dependent)

The part of the patch that introduces a `make check` that triggers a build and a test execution I am somewhat okay (if there are really people who would like to use that way). But not the part that disables the `test` target built into cmake, and the default test compilation.

bazsi:
On Sun, Oct 13, 2019, 15:13 Kókai Péter <notifications@github.com> wrote:

> *@Kokan* commented on this pull request.
>
> In .travis.yml
> <https://github.com/syslog-ng/syslog-ng/pull/2971#discussion_r334277539>:
>
> > @@ -157,8 +157,7 @@ matrix:
>              -DPYTHON_VERSION=2
>              -DCMAKE_INSTALL_PREFIX=$HOME/install/syslog-ng
>              ..
> -        - make --keep-going -j $(nproc) all install
> -        - ctest -j $(nproc) --output-on-failure
> +        - make --keep-going -j $(nproc) all check install
>
>
>    1. only build test programs in case we are running tests, and to couple
>    rebuilding of the test programs with running them.
>
> I could agree that there are use cases where one do not want to compile
> UT, but I know only those cases where also the user do not want to execute
> them either. (completely disabling the tests are fine and there is a cmake
> way of doing that: BUILD_TESTING).
> Please if you have other use case than that please share.
> Also the link mention something like this: This is IMHO a very nice
> feature, because the normal build is not bogged down by the compilation of
> many test programs, or worse, comes to a grinding halt due to a compilation
> error in one of the test programs you didn't have time to fix yet.
> Well, I hope people does not feel "bogged" about not compiling test ...
>

This wasn't my my intention (although it's been a while, so I can't
remember all the specifics).

The reasons I needed a separate target was:
* primary reason: pylint and pep8 are not unit tests and their dependencies
may not be available (e.g. on centos6) while both the production code and
the unit tests can build. so I wanted to create a target that _only_ runs
unit tests and not pylint/pep8 and other linters.
* Since our make targets are proliferating, I wanted to pull ci related
targets into "simple" do all targets to delegate responsibility of what
needs to be run in the CI (e.g. ci-build, ci-test, ci-lint, ci-functest,
etc), at the end this patch was dropped from the series (as these patches
tend to go in slowly...). "make test" is a special target that cannot be
extended by new dependencies and neither can be used as a dependency of a
normal target. make check on the other hand can be.
* "make test" is not building test code, while "make check" in automake
does. I regularly use "make check" to trigger a rebuild & run tests
locally, although it is true that I rarely use cmake (except to validate
my  changes work).

I am not a cmake expert, but I don't really like that the "test" target is
special.


>    1. to reference the unit test phase as a DEPENDS in
>    add_custom_target(), as
>    the "test" target is special, whereas "check" is a normal target.
>
>  This is the same reason as (1).
>

I don't understand the comment, see my explanation above about CI related
"simple" targets that could reference other nodes.


>
>    1. to remove pep8/pylint from unit test execution in a subsequent patch
>
> Those can be removed with the exact same way as in your current branch,
> without this one. (they are not really dependent)
>
> The part of the patch that introduces a make check that triggers a build
> and a test execution I am somewhat okay (if there are really people who
> would like to use that way). But not the part that disables the test
> target built into cmake, and the default test compilation.
>
The normal "make test" works just fine. I was just excluding the building
of these test program from the normal build, so "make test" might not find
the binaries, but also note that make test may run out-of-date binaries as
it doesn't trigger a rebuild.

But I can live with them being built in the normal process, as well as in
check.

Bazsi

.travis.yml
bazsi: we've agreed that we would keep the building of test programs in "make all", while also triggering the build of test programs in "make check".

bazsi:
I have now rebased this branch, concluded on the cmake changes and found out that my previous patch to make atomic-gssize.h work on centos6 was not the right way, I have instead updated our in-tree glib compat code to match that of glib.

Hopefully this becomes green now making syslog-ng with centos-6 compatible (again).

The dbld centos6 bits can be decided in a separate PR.

lib/compat/glib.h
```diff
@@ -50,29 +50,29 @@ gchar *g_canonicalize_filename (const gchar *filename,
 #define g_atomic_pointer_add(atomic, val) \
   (G_GNUC_EXTENSION ({                                                       \
     G_STATIC_ASSERT (sizeof *(atomic) == sizeof (gpointer));                 \
-    (void) (0 ? (gpointer) *(atomic) : 0);                                   \
-    (void) (0 ? (val) ^ (val) : 0);                                          \
+    (void) (0 ? (gpointer) *(atomic) : NULL);                                \
+    (void) (0 ? (val) ^ (val) : 1);                                          \
```
lbudai: for other reviewers: 
https://bugzilla.gnome.org/show_bug.cgi?id=793399

PR #2972, opened 2019-10-16
Improve Redis driver connection handling

szemere:
UPDATE:
Dropped the batching part due to problems with message tracking in case of connection errors, but I thought it would be nice to include the rest of the refactors in the next release.
The reduced number of  `redis_dd_connect` calls still gives us a decent speed improvement. (~20k according my measurements)
- added timeout() option for amqp. This adds a timeout for the blocking socket operations.

----------

Add **batch-lines** support for the Redis driver. Redis API calls it pipelining: https://redis.io/topics/pipelining

Also:
 - Thanks for @Kokan, some performance improvement with template handling.
 - Put `_argv_to_string` behind `debug_flag`, because it has a high performance impact.
 - Removed an extra `check_connection_to_redis` call from the insert call chain. (Since it requires a network round trip, it has a high performance impact!)
- by default amqp does not use batching, to enable it, use the `batch-lines()` option.
- added `timeout()` option for amqp. This adds a timeout for the blocking socket operations.

My measurements on my laptop:
 - original driver: ~160 msg/s
 - new driver without batch mode: ~400msg/s
 - batching mode, batch_lines(2): ~2000 msg/s
 - batching mode, batch_lines(5): ~ 50 000 msg/s

note: With some tweaks with batch_lines and batch_timeout I managed to squeez out around: ~80-100k msg/s from the driver, but I think it highly depends on the infrastructure.

### DIFF
--- a/modules/redis/redis-grammar.ym
+++ b/modules/redis/redis-grammar.ym
@@ -44,6 +44,7 @@
 %token KW_REDIS
 %token KW_COMMAND
 %token KW_AUTH
+%token KW_TIMEOUT
 
 %%
 
@@ -79,6 +80,10 @@ redis_option
           {
             redis_dd_set_command_ref(last_driver, $3, $4);
             free($3);
+          }
+        | KW_TIMEOUT '(' nonnegative_integer ')'
+          {
+            redis_dd_set_timeout(last_driver, $3);
           }
         | threaded_dest_driver_option
         | { last_template_options = redis_dd_get_template_options(last_driver); } template_option
--- a/modules/redis/redis.c
+++ b/modules/redis/redis.c
@@ -144,6 +144,36 @@ redis_dd_format_persist_name(const LogPipe *s)
   return persist_name;
 }
 
+static inline void
+_trace_reply_message(redisReply *r)
+{
+  if (trace_flag)
+    {
+      if (r->elements > 0)
+        {
+          msg_trace(">>>>>> REDIS command reply begin",
+                    evt_tag_long("elements", r->elements));
+
+          for (gsize i = 0; i < r->elements; i++)
+            {
+              _trace_reply_message(r->element[i]);
+            }
+
+          msg_trace("<<<<<< REDIS command reply end");
+        }
+      else if (r->type == REDIS_REPLY_STRING || r->type == REDIS_REPLY_STATUS || r->type == REDIS_REPLY_ERROR)
+        {
+          msg_trace("REDIS command reply",
+                    evt_tag_str("str", r->str));
+        }
+      else
+        {
+          msg_trace("REDIS command unhandled reply",
+                    evt_tag_int("type", r->type));
+        }
+    }
+}
+
 static gboolean
 send_redis_command(RedisDriver *self, const char *format, ...)
 {
@@ -154,7 +184,10 @@ send_redis_command(RedisDriver *self, const char *format, ...)
 
   gboolean retval = reply && (reply->type != REDIS_REPLY_ERROR);
   if (reply)
-    freeReplyObject(reply);
+    {
+      _trace_reply_message(reply);
+      freeReplyObject(reply);
+    }
   return retval;
 }
 
@@ -195,13 +228,15 @@ redis_dd_connect(RedisDriver *self)
   if (self->auth)
     if (!authenticate_to_redis(self, self->auth))
       {
-        msg_error("REDIS: failed to authenticate");
+        msg_error("REDIS: failed to authenticate",
+                  evt_tag_str("driver", self->super.super.super.id));
         return LTR_NOT_CONNECTED;
       }
 
   if (!check_connection_to_redis(self))
     {
-      msg_error("REDIS: failed to connect");
+      msg_error("REDIS: failed to connect",
+                evt_tag_str("driver", self->super.super.super.id));
       return LTR_NOT_CONNECTED;
     }
 
--- a/modules/redis/redis.c
+++ b/modules/redis/redis.c
@@ -237,7 +237,7 @@ _fill_argv_from_template_list(RedisDriver *self, LogMessage *msg)
     }
 }
 
-static GString *
+static const gchar *
 _argv_to_string(RedisDriver *self)
 {
   GString *full_command = scratch_buffers_alloc();
@@ -249,7 +249,7 @@ _argv_to_string(RedisDriver *self)
       append_unsafe_utf8_as_escaped_text(full_command, self->argv[i], self->argvlen[i], "\"");
       g_string_append(full_command, "\"");
     }
-  return full_command;
+  return full_command->str;
 }
 
 /*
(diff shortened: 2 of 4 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
This is cool stuff. :)

modules/redis/redis.c
```diff
@@ -185,30 +185,33 @@ redis_dd_connect(RedisDriver *self)
...
                 evt_tag_str("error", self->c->errstr),
                 evt_tag_int("time_reopen", self->super.time_reopen));
-      return FALSE;
+      return LTR_NOT_CONNECTED;
     }
 
   if (self->auth)
     if (!authenticate_to_redis(self, self->auth))
       {
         msg_error("REDIS: failed to authenticate");
```
bazsi: Can you add more details to this error message?

modules/redis/redis.c
```diff
@@ -290,15 +294,82 @@ redis_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
...
+
+static LogThreadedResult
+redis_worker_insert_selector(LogThreadedDestDriver *s, LogMessage *msg)
+{
+  if (s->batch_lines > 1)
+    s->worker.insert = redis_worker_insert_batch_mode;
+  else
+    s->worker.insert = redis_worker_insert;
+
+  return s->worker.insert(s, msg);
```
bazsi: This might be executing in multiple threads, which probably wouldnt cause problems, but I would do this initialization during init()

modules/redis/redis.c
```diff
@@ -290,15 +294,82 @@ redis_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
...
 }
 
+static LogThreadedResult
+redis_worker_insert_batch_mode(LogThreadedDestDriver *s, LogMessage *msg)
+{
+  RedisDriver *self = (RedisDriver *)s;
+
+  ScratchBuffersMarker marker;
+  scratch_buffers_mark(&marker);
+
```
bazsi: This is unused.

modules/redis/redis.c
szemere: https://github.com/redis/hiredis/blob/master/read.h#L51
hiredis has a variety of return codes. (redisReply->type) But we are currently free-ing the reply object on the fly within `send_redis_command`.

Should I add general printout inside the send command, or change the interface of the `send_redis_command` function and delegate the ownership of the reply object to the calle?

szemere: Tried to handle as many return value as possible. But only added it to the trace level.

modules/redis/redis.c
szemere: Moved into the init phase, thank you. Also managed to get rid of the "selector" function. By defaulting the non batching mode.

modules/redis/redis.c
szemere: Added the missing reclaim functions. (`_fill_argv_from_template_list` is working with scratch buffers, same as the `_argv_to_string` debug function)

kira-syslogng:
Build SUCCESS

modules/redis/redis.c
```diff
@@ -154,7 +191,11 @@ send_redis_command(RedisDriver *self, const char *format, ...)
 
   gboolean retval = reply && (reply->type != REDIS_REPLY_ERROR);
   if (reply)
-    freeReplyObject(reply);
+    {
+      if (trace_flag)
+        _trace_reply_message(reply);
```
bazsi: I think you could move the conditional inside the _trace_reply_message() function.

szemere:
WIP: discussed IRL, redis lib supports socket level timeout, which can fix some rare network related blocking in the driver, and reduce the number of connect calls in non batching mode too.

kira-syslogng:
Build SUCCESS

modules/redis/redis.c
```diff
@@ -144,6 +144,43 @@ redis_dd_format_persist_name(const LogPipe *s)
...
+        }
+
+      msg_trace("<<<<<< REDIS command reply end");
+    }
+  else if (r->type == REDIS_REPLY_STRING)
+    {
+      msg_trace("REDIS command general reply",
+                evt_tag_str("str", r->str));
+    }
+  else if (r->type == REDIS_REPLY_STATUS)
```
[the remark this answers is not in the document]
szemere: Changed to be more general, so no need for lookup. Is this ok?

modules/redis/redis.c
```diff
@@ -307,6 +307,62 @@ redis_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
...
+
+static LogThreadedResult
+redis_worker_flush(LogThreadedDestDriver *s)
+{
+  RedisDriver *self = (RedisDriver *)s;
+
+  if (s->worker.instance.batch_size == 0)
+    return LTR_SUCCESS;
+
+  LogThreadedResult connection_status = redis_dd_connect(self);
```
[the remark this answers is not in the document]
szemere: Please see the latest changes. Eliminated redis_dd_connect calls from the driver.

modules/redis/redis.c
```diff
@@ -229,11 +229,21 @@ _fill_argv_from_template_list(RedisDriver *self, LogMessage *msg)
...
+      if (log_template_is_trivial(redis_command))
+        {
+          gssize size;
+          self->argv[i] = (gchar *)log_template_get_trivial_value(redis_command, msg, &size);
+          self->argvlen[i] = size;
+        }
+      else
+        {
+          GString *buffer = scratch_buffers_alloc();
+          log_template_format(redis_command, msg, &self->template_options, LTZ_SEND,
```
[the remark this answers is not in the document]
szemere: Some refactors in the topic, please share your opinion.

modules/redis/redis.c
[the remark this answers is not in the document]
furiel: Nice! Thanks

modules/redis/redis.c
```diff
@@ -307,6 +307,58 @@ redis_worker_insert(LogThreadedDestDriver *s, LogMessage *msg)
...
+
+  if (s->worker.instance.batch_size == 0)
+    return LTR_SUCCESS;
+
+  LogThreadedResult connection_status = redis_dd_connect(self);
+  if (connection_status != LTR_SUCCESS)
+    return connection_status;
+
+  void *reply;
+  if (redisGetReply(self->c, &reply) != REDIS_OK)
```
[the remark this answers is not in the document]
gaborznagy: based on the protocol, it seems that you need to get every replies:
>... All the replies can be read at the end.

 https://redis.io/topics/protocol#multiple-commands-and-pipelining

modules/redis/redis-grammar.ym
```diff
@@ -44,6 +44,7 @@
 %token KW_REDIS
 %token KW_COMMAND
 %token KW_AUTH
+%token KW_TIMEOUT
```
gaborznagy: Please add the new keyword to redis-parser.c as well, otherwise it won't work.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/redis/redis-grammar.ym
szemere: I completely missed this comment :( Thank you very much! Fixed

modules/redis/redis.c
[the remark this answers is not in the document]
szemere: @gaborznagy thank you for pointing this out, you are absolutely right. I dropped batching support temporary, and will open a new PR with it.

kira-syslogng:
Build SUCCESS

PR #2973, opened 2019-10-16
LightPytestRuntestSetupHook: Refactor pytest_runtest_setup() hook

mitzkia:
Motivation:
There was an incorrectly used pytest object in pytest_runtest_setup() hook: `item._request`
This is not a public object and it can cause a pytest error in some circumstances.


kira-syslogng:
Build SUCCESS

PR #2974, opened 2019-10-17
logmsg: add default PRI value to log_msg_init

gaborznagy:
In case it's left uninitialized, the message could seem as having a pri LOG_KERN | LOG_EMERG.

This is only interesting, if a user would like to create a log message, but without initial parsing,
thus log_msg_new() is not suitable, as that needs a format_handler, e.g. syslog.

Currently this is done by log_msg_new_empty(), which is used by some drivers, but they are not setting
PRI fields.

### DIFF
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -48,6 +48,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <syslog.h>
 
 /*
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng restest this please test branch=default_pri_changed;

kira-syslogng:
Build SUCCESS

PR #2901, opened 2019-09-02
Fix SCL block parameter substitution 

MrAnno:
After a quoted character, transitioning into `CLS_WITHIN_STRING` is not
enough, because another quote or "string end" character may follow.

Here is the state machine of `_track_string_state()`, the red lines were missing.

string_state

Fixes #2879

PR #2977, opened 2019-10-19
Fix syntax error in telegram()

MrAnno:
After fixing the SCL parameter substitution in #2901, the Telegram destination has stopped working.

The type of the `use-system-cert-store()` option should be `yes/no`, strings are not accepted.


kira-syslogng:
Build SUCCESS

PR #2982, opened 2019-10-22
[WIP] dbld: remove the automatic rm function from the shell command

szemere:
WIP: Because I just wanted to start a discussion about the topic.

pro: I use devshell for development and it is a little inconvenient that I lose my installed packages, and other changes in case my terminal frozen, or I have to reboot my computer in case of updates.

con: With this change a user has to manually delete old containers, and it might happen that he is not working in a clean environment. (CI systems do not use interactive shells, so they are not affected by this change.)

### DIFF
--- a/dbld/rules
+++ b/dbld/rules
@@ -131,8 +131,12 @@ run-%: setup
 
 shell: shell-$(DEFAULT_IMAGE)
 shell-%: setup
-	$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -ti balabit/syslog-ng-$* /source/dbld/shell
-
+	$(DOCKER) inspect $* > /dev/null 2>&1; \
+	if [ $$? -eq 0 ]; then \
+		$(DOCKER) start -ia $*; \
+	else \
+		$(DOCKER) run $(DOCKER_RUN_ARGS) -ti --name $* balabit/syslog-ng-$* /source/dbld/shell; \
+	fi
 
 images: $(foreach image,$(IMAGES), image-$(image))
 image: image-$(DEFAULT_IMAGE)


kira-syslogng:
Build SUCCESS

mitzkia:
For me it will be also handy, so +1 for the PR. 
Other option is not to re-define existing `dbld/rules shell` command (to not broke for others), but introduce a new command, like: `dbld/rules shell-permanent` (or something like this). The actual changes could apply under this new command.

bazsi:
The persistent portions are kept outside of the container. What do you want
to persist between invocations?

mitzkia:
I also found inconvenient to install my custom apt or pip packages again and again when I re-enter the shell (because of OS restart or something like this).
What kind of "persistent portions" do you mean? OS or shell configurations (files)? 
For me it is not about like that.

bazsi:
If you find yourself installing stuff into devshell regularly, I would opt
for one of these options:

1) install those into devshell via its docker image

2) create an image of your own that uses devshell as a parent, using a
regular dockerfile.

3) when you have your local changes, just do a docker commit and use that
as devshell.

The last should be pretty straightforward, even if you dont want to mess
with dockerfiles.

bazsi:
Looking at the patch and thinking a bit I am ok with this change. I didn't know start can invoke an interactive shell. 

You might want to use a name that reflects that it is a syslog-ng image we are using, right now it would just use "debian-stretch" instead of "syslog-ng-debian-stretch"

szemere:
Thank You for your comments. What do you think about @mitzkia 's idea? Instead of changing the current behavior of the command, add a new command to dbld/rules.
(note: The name comes from the other commands, I just wanted to use the already existing names. We can change it any time.)

bazsi:
I don't like proliferating targets, I think we should decide which one makes more sense and do that. The only use-case that might break is that when I rebuild/re-pull my "devshell" image (or any of the images), the changes will not be picked up when I execute shell. This is a bit unintuitive and can easily leave me in a stale container.

We could either make this update manual (e.g. the user should invoke `docker rm syslog-ng-devshell` or a similar target in dbld), or we could automate this by adding a check to the shell target.

Hmm, what do you think if "shell" checked if the "COMMIT" label of the image is different from the container and if yes, we can assume that a build/re-pull happened. In this case we could error out with a message instructing how to get rid of the stale image (maybe also a way to reset devshell to the container's version).

I think this should be doable. What do you think?

furiel:
I would also prefer avoiding adding extra targets.

The warning mentioned above would be nice too if it is easy to implement. But I am fine with this as it is now, and later the check can be added if we run to this stale image problem too often. 

I do not use dbld too often, but when I do, I need it because I need to test something platform specific that cannot (do not want) be done on my development os, typically involving extra packages. Maybe just an apt install or pip install, or sometimes compiling something from source.

I also usually call docker commit. But it would be a little more convenient if this would work out of the box.

szemere:
Closing in favor of #3038 . Thank you all for sharing your opinions!

PR #2983, opened 2019-10-25
Mark Changes

nguyenfilip:
Currently, upon start or reload of configuration, first mark message comes only
after mark_freq. So in effect any restart/reload causes next mark to be delayed.

In addition to that, when many syslog-ng daemons start at the same
time, the mark messages are going to be correlated and may cause spikes
of those MARK's every time mark_freq passes.

This change aims to do the following. When periodical marks are used, emit first
mark as soon as possible after start/reload and then schedule second mark with
random delay in interval <1, mark_freq>.

### DIFF
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -39,6 +39,7 @@
 #include "timeutils/format.h"
 #include "timeutils/misc.h"
 
+#include <time.h>
 #include <unistd.h>
 #include <assert.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

lbudai:
@kira-syslogng : ok to test

kira-syslogng:
Build FAILURE

nguyenfilip:
@lbudai do you suggest just to hide this feature behind a config option. Or is there some other extension point/programming model in syslog-ng that I use?

kira-syslogng:
Build SUCCESS

bazsi:
I am not sure we would need an explicit option if the randomized distribution of mark messages make better sense than what we are doing today.

Can you offer more background information on why you think the new behaviour is better? What problem are you solving?

Also, if we were to randomize marks, I wouldn't spread them over mark_freq, that seems to be a bit too much, spreading them over a few seconds should solve spikes just as well. Also, I am not yet convinced synchronized MARKs cause that much headache. Are you reloading tens of thousands of syslog-ng instances in a synchronized manner?

Thanks

lbudai:
@bazsi : I would not break compatibility.

I'm not questioning that the presence of a relatively huge number of simultaneous mark messages could be a problem in _some_ environments but I think randomized mark messages should not be the default.
I understand that the generation time of an individual mark message is a stochastic variable...

On the other hand, I do not see any problems with making this optional...

@nguyenfilip : yes, just add a new config option for enabling the randomized function
But at first, try to answer the questions from @bazsi .

MrAnno:
Sorry, the PR was closed accidentally (misclick). I'm reopening, because @lbudai just turned off his computer.

kira-syslogng:
Build SUCCESS

lbudai:
@nguyenfilip : any update on this?

furiel:
Closing due to inactivity. Please reopen if you can share more about the use usecase, or need help from us about the implementation.

PR #2985, opened 2019-10-28
Affile delay acks until messages written to disk

bazsi:
As discussed in #2887 due to a failing testcase the file destination was pretty optimistic on when to acknowledge a message back, even though the entire infrastructure is now in place to do it somewhat later.

This branch does exactly that, instead of acking the message in log_proto_client_post(), we do it once write() acknowledged that it wrote the message. This combined with fsync() we can really be sure that the buffer embedded in LogProtoFileWriter will not hold messages that are not yet written but are acknowledged to the source.

I think some refactorings are heavily needed here, I am just wondering what @kira-syslogng thinks about the patch.

The branch also improves testing in this corner of the codebase.

### DIFF
--- a/modules/affile/tests/Makefile.am
+++ b/modules/affile/tests/Makefile.am
@@ -9,7 +9,8 @@ modules_affile_tests_TESTS 				= \
 	modules/affile/tests/test_collection_comparator \
 	modules/affile/tests/test_file_opener \
 	modules/affile/tests/test_wildcard_file_reader \
-	modules/affile/tests/test_file_list
+	modules/affile/tests/test_file_list		\
+	modules/affile/tests/test_file_writer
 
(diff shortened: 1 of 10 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/affile/tests/Makefile.am
```diff
@@ -9,7 +9,8 @@ modules_affile_tests_TESTS 				= \
 	modules/affile/tests/test_collection_comparator \
 	modules/affile/tests/test_file_opener \
 	modules/affile/tests/test_wildcard_file_reader \
-	modules/affile/tests/test_file_list
+	modules/affile/tests/test_file_list		\
```
furiel: Please add the test file to cmake too.

bazsi: done

bazsi:
rebased and did the outstanding change request.

kira-syslogng:
Build SUCCESS

PR #2987, opened 2019-10-29
travis: test cmake + ninja combination

Kokan:
This only replaces one of the cmake + make build with cmake + ninja.


kira-syslogng:
Build SUCCESS

PR #2988, opened 2019-10-29
Light property based test fixes

mitzkia:
This PR contains 4 fixes for bugs we've found with property-based testing method (using Light).
3 of them are trivial; the 4th (stomp() patch) fixes some buffer overrun issues that can be triggered remotely with suitably crafted input.

There are 3 more, but they deserve a separate PR as they are not that trivial.


kira-syslogng:
Build SUCCESS

furiel:
Sorry for the regression. :(

Issue #2820, opened 2019-07-05
Config objects preceded by an inline destination are ignored


MrAnno:
# syslog-ng

## Version of syslog-ng

syslog-ng v3.22.1

# Issue

The following configuration will not send any logs to `d_asd`:

```
@version: 3.22
destination d_asd {
	network("127.0.0.1" port(4448));
};

log {
	source { network(port(4444)); };
	destination {
		network("127.0.0.1" port(4447) );
	};
	destination(d_asd);
	flags(flow-control);
};
```

The cfg graph built from this config is incorrect. Changing the order of the 2 destinations (listing the inline destination last) fixes the issue.

## Root cause
Inline destinations are compiled as `sequences` (`cfg_tree_compile_sequence()`), where no multiplexer is created for other objects. In case of general/non-inline destinations (`cfg_tree_compile_reference()`), the multiplexer is there:

https://github.com/balabit/syslog-ng/blob/1742b11e5cfa6544ece38aaecde96e9b423d61c5/lib/cfg-tree.c#L842-L851

Unfortunately, it seems syslog-ng creates a new multiplexer for **every** non-inline destination instead of adding a `hop` to the previous one. I don't know the reason, but we could improve on this too.

bazsi:
Hi,

The reason for multiplexers is that the next pointer for a non-inline
destination cannot be used. So the multiplexer is always inline, its list
of next hops should contain a reference to the destination group.

I think that the inline destinations being a sequence should not cause
problems.

We could discuss this via phone or slack or similar to find a good enough
solution.

László Várady <notifications@github.com> ezt írta (időpont: 2019. júl. 5.,
P 14:24):

furiel:
Is it possible that the problem depends on the types of the config elements? I checked with an inline file destination and a network destination, and it works for me:

```
$ cat ../etc/tmp.conf 
@version: 3.24

options {
  time-reopen(5);
};

destination d_network {
    network("127.0.0.1" port(4447) template("NETWORK: $MESSAGE\n"));
};

log {
	source { example-msg-generator(); };
	destination { file(/dev/stdout template("FILE: $MESSAGE\n")); };
        destination(d_network);
	flags(flow-control);
};
$ nc -k -l 4447 &
[1] 8261
$ timeout 5 ./syslog-ng -Fe -f ../etc/tmp.conf 
[2019-10-22T08:34:25.696765] syslog-ng starting up; version='3.24.1.11.g126a19f'
[...]

PR #2989, opened 2019-10-29
cfg-tree: inline network destinations disjoint the rest of the config

furiel:
Fixes: https://github.com/syslog-ng/syslog-ng/issues/2820

Any destination after an inline network destination driver does not handle messages.

The root cause of the issue was that afsocket overwrites its `pipe_next` field with its own writer. Previously, `pipe_next` was pointing to the rest of the config graph. Essentially distjointing the config after the inline network driver. The problem only happens with network driver, and only when the network was inline. Some additional info can be found in #2820.

For example:
```
@version: 3.24

options {
  time-reopen(5);
};

destination d_network {
    network("127.0.0.1" port(4448) );
};

log {
	destination { network("127.0.0.1" port(4447) ); };
        destination(d_network);
	flags(flow-control);
};
```

Explanation of the patch is here: https://github.com/syslog-ng/syslog-ng/pull/2989#issuecomment-554925749

After this change, the config graph of the previous example becomes this.

tmp

Pros:
- From now any destinations can attach their own pipes to themselves without corrupting the config graph.

### DIFF
--- a/lib/cfg-tree.c
+++ b/lib/cfg-tree.c
@@ -1049,6 +1049,31 @@ cfg_tree_compile_sequence(CfgTree *self, LogExprNode *node,
       node_properties_propagated = TRUE;
     }
 
+  if (node->content == ENC_DESTINATION)
+    {
+      /* We want to leave pipe-next available to use for
+         destinations. Config graph uses pipe-next to pass messages forward
+         in a sequence layout. But pipe-next might be overridden (for
+         example network destination with LogWriter) hence disjointing
+         the config graph.
+
+         This patch links destinations in T form, instead of single link.
+
+         * (endpoint of sequence)
+         |
+         V
+         * (multiplexer) -(next-hop)-> destination -(pipe-next*)-> logwriter
+         |
+         (pipe-next)
+         |
+         V
+         * (rest of the sequence)
+         That way destinations are free to use the pipe-next*
+      */
+
+      last_pipe = first_pipe;
+    }
+
   *outer_pipe_tail = last_pipe;
   *outer_pipe_head = first_pipe;
   return TRUE;


kira-syslogng:
Build SUCCESS

furiel:
@pzoleex executed a network destination performance test. The eps remained the same.
Thank you for the test!

Kokan:
I plan to do the review by this Friday.

kira-syslogng:
Build SUCCESS

furiel:
After updating the config generation graph generation PR, I can show now the types of the arcs. I updated the description with a new screenshot.

Also the original title of my patch was wrong. I updated the patch description too.

We do not need to add multiplexers to the graph. In fact, we do not need to add any kind of pipe element to the graph. This was an overlook of my side, originally I added an unnecessary cfg-tree dummy pipe.

The thing is, destinations are always linked through multiplexers into the config graph. This is unfortunately not obvious from the code. For referenced destinations, it is easier, one can see the log_multiplexer_new call explicitely. For inline destinations, it is true because they are linked through the junction layout, which creates a multiplexer.

And when we have the multiplexer --next-hop--> destination subpipe, all we need to do is set the last_pipe (we are on a recursive call, first_pipe, last_pipe is used to link the currently generated pipe upwards) to the multiplexer.

In summary, this is not a layout problem. We are compiling the right layout (sequence), just the destinations need to be linked differently. So the fix is in the right place. From clean code pespective, the only inconvenient thing is that it is not obvious that destinations are pulled through a multiplexer, but that is a different story.

I added an extra ascii art to the code as comment, that might help others understand and anticipate, what is happening in the code.

Kokan:
> In summary, this is not a layout problem.

I still disagree on this part, but that is a story for an other day. It won't effect the goodness of this PR.

lib/cfg-tree.c
```diff
@@ -1049,6 +1049,31 @@ cfg_tree_compile_sequence(CfgTree *self, LogExprNode *node,
...
+         * (multiplexer) -(next-hop)-> destination -(pipe-next*)-> logwriter
+         |
+         (pipe-next)
+         |
+         V
+         * (rest of the sequence)
+         That way destinations are free to use the pipe-next*
+      */
+
+      last_pipe = first_pipe;
```
MrAnno: I also think this is not in the right place (clean code; the layout is OK in my opinion), but I don't see how this could be done cleaner without rewriting/refactoring too much thing around the junction and sequence construction.

PR #2992, opened 2019-10-30
Add disable jit flag to logmatcher

### LINKED ISSUE #2986: Getting Issue of SE Linux while ceating parsers for Syslog-NG
Hi,

I have created parsers for Syslog NG and below are the snippets written in the syslog NG configuration file:

filter sh_auth_logs
{
       program("sshd");
}

Parser p_ssh_auth_logs
{
       db_parser(file("/etc/syslog-ng/patterndb.d/sshd_events.xml"));
}

source syslog_udp
{
       syslog(transport(udp) port(514));
};

template t_ssh { template("$DATE $HOST $MSGHDR$MSG\n"); template_escape(no); };

destination df_wrt0
{
        file("/var/log/syslog-ng.log"
        template(t_ssh)
       owner("root")
        group("root")
        perm(0777));

);

log
{
       source(syslog_udp);
       filter(sh_auth_logs);
       destination(df_wrt0);
       flags(final);
};

 after doing all the above changes and restarting syslog-ng, I am getting the attached error of SE Linux.

Requesting all to provide solution for the same issue.
Syslog NG Error

bazsi:
logmatcher: allow the user to disable PCRE JIT using the disable-jit flag

this should resolve #2986 at least the normal filter() case. the same in multi-line(regexp) support is not this trivial.

**EDIT**:
config examples
```
rewrite r_demo {
  subst("a+" "b" value(".extra.info") flags("disable-jit"));
};

filter f_demo {
  host("internal-.*" flags("disable-jit"));
};

```

### DIFF
--- a/lib/logmatcher.c
+++ b/lib/logmatcher.c
@@ -294,17 +294,14 @@ typedef struct _LogMatcherPcreRe
 } LogMatcherPcreRe;
 
 static gboolean
-log_matcher_pcre_re_compile(LogMatcher *s, const gchar *re, GError **error)
+_compile_pcre_regexp(LogMatcherPcreRe *self, const gchar *re, GError **error)
 {
-  LogMatcherPcreRe *self = (LogMatcherPcreRe *) s;
   gint rc;
-  const gchar *re_comp = re;
   const gchar *errptr;
   gint erroffset;
   gint flags = 0;
 
   g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
-  log_matcher_store_pattern(s, re);
 
   if (self->super.flags & LMF_ICASE)
     flags |= PCRE_CASELESS;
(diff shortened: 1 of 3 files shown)


lib/logmatcher.c
```diff
@@ -347,7 +347,9 @@ log_matcher_pcre_re_compile(LogMatcher *s, const gchar *re, GError **error)
     }
 
   /* optimize regexp */
-  self->extra = pcre_study(self->pattern, PCRE_STUDY_JIT_COMPILE, &errptr);
+  self->extra = pcre_study(self->pattern,
```
Kokan: could you please extract that condition (` (self->super.flags & LMF_DISABLE_JIT) ? 0 : PCRE_STUDY_JIT_COMPILE`) out of the function call and assign to a local `gint options` variable instead ? also `LMF_DISABLE_JIT` should only negate the proper bit in the options and not set to `0`.

kira-syslogng:
Build SUCCESS

lib/logmatcher.c
bazsi: I extracted some functions in the area, can you check if this is ok?

kira-syslogng:
Build SUCCESS

bazsi:
I already have patches in that direction.

gaborznagy:
> what I am working on is a global use-jit() option that would control the
disable-jit flag of filters as well as the jit mode for multi-line mode

Thanks, I wanted to have some thoughts about how should we solve the JIT configuration in other modules before we merge something on a drivel level, which we should support to provide backward compatibility.

PR #2995, opened 2019-11-04
fix python exception reporting when no exception happened

furiel:
Sometimes exception reporting is called even when no exception has
happened.

For example when the python class is not found: 
While searching for python class, the code simply calls
getattr, which does not generate exception, just returns null.

In those cases python destination emitted this unnecessarily.
```
[2019-11-04T09:18:44.079993] Error printing proper Python traceback for the exception, printing the error caused by print_exception() itself;
SystemError: NULL object passed to Py_BuildValue
```

This patch makes this function no-op if no exception happened.


kira-syslogng:
Build SUCCESS

furiel:
rebased to make travis green

kira-syslogng:
Build SUCCESS

PR #2998, opened 2019-11-05
python-dest: actually use return value of open method

furiel:
Fixes: https://github.com/syslog-ng/syslog-ng/issues/2513.

Although the admin guide tells users to set the return value of open properly, the code does not check that. Instead, there is an is_opened method that was used to answer:
- if the driver is open
- In some cases, if the last open was successful or not.

This architecture makes users write unnecessary complex code, since the open-ness must be stored and updated as side-effects.

For example this does not do what one would expect, because it sends message even though open returns with false.

<details>

```
@version: 3.24

log {
  destination { python(class(PythonDest)); };
};

python {
class PythonDest(object):
    def open(self):
        return False

    def send(self, message):
        print("sending message")
        return True
};

$ ./syslog-ng -Fe -f ../etc/python.conf 
[2019-11-05T09:35:54.243644] syslog-ng starting up; version='3.24.1.70.ge11a8e4'
sending message
[2019-11-05T09:35:56.766802] syslog-ng shutting down; version='3.24.1.70.ge11a8e4'
```

</details>

This patchset make use of the return value of open method, and act accordingly. This allows user to use open without is_opened.

Conceptually, one can think of `is_opened` as a ping or heartbeat or any similar functionality that detects if the underlying resource is open or not.

For compatibility reasons, if open is void, I consider it as a success return value. At a later time, we can remove the related code.

One known functional change:
- is_opened will not be called during thread_init, before the first open call.
Logically, this should be ok, because a driver should not be open before the first open. Still, this can break code that can be open before the first open.

Updated the e2e tests because of the functional change, and added a new test dedicated to is_opened.

### DIFF
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -182,7 +182,15 @@ _py_invoke_open(PythonDestDriver *self)
   if (!self->py.open)
     return TRUE;
 
-  return _dd_py_invoke_bool_function(self, self->py.open, NULL);
+  PyObject *ret;
+  gboolean result = FALSE;
+
+  ret = _py_invoke_function(self->py.open, NULL, self->class, self->super.super.super.id);
+  if (ret)
+    result = PyObject_IsTrue(ret);
+  Py_XDECREF(ret);
+
+  return result;
 }
 
 static void
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -200,6 +200,14 @@ _py_invoke_open(PythonDestDriver *self)
     }
   Py_XDECREF(ret);
 
+  if (self->py.is_opened)
+    {
+      if (!result)
+        return FALSE;
+
+      return _py_invoke_is_opened(self);
+    }
+
   return result;
 }
 
@@ -440,8 +448,7 @@ python_dd_insert(LogThreadedDestDriver *d, LogMessage *msg)
   gstate = PyGILState_Ensure();
   if (!_py_invoke_is_opened(self))
     {
-      _py_invoke_open(self);
-      if (!_py_invoke_is_opened(self))
+      if (!_py_invoke_open(self))
         {
           result = LTR_NOT_CONNECTED;
           goto exit;
@@ -465,10 +472,10 @@ python_dd_open(PythonDestDriver *self)
   PyGILState_STATE gstate;
 
   gstate = PyGILState_Ensure();
-  _py_invoke_open(self);
+  gboolean retval = _py_invoke_open(self);
 
   PyGILState_Release(gstate);
-  return TRUE; // Will be changed to the retval of open later
+  return retval;
 }
 
 static LogThreadedResult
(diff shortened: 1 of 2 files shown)


furiel:
@kira-syslogng retest this please test branch=furiel-python-connect-retval;

kira-syslogng:
Build SUCCESS

furiel:
rebased to make travis green

@kira-syslogng retest this please test branch=furiel-python-connect-retval;

kira-syslogng:
Build SUCCESS

modules/python/python-dest.c
```diff
@@ -159,7 +160,7 @@ _dd_py_invoke_bool_method_by_name_with_args(PythonDestDriver *self, const gchar
                                                   self->super.super.super.id);
 }
 
-static gboolean
+static gboolean G_GNUC_UNUSED
```
MrAnno: `_dd_py_invoke_bool_method_by_name()` is not used anymore. Can we remove it, please? This is actually boilerplate copy-pasted to every Python plugin :(.

modules/python/python-dest.c
```diff
@@ -178,7 +179,36 @@ _py_invoke_is_opened(PythonDestDriver *self)
...
 _py_invoke_open(PythonDestDriver *self)
 {
-  return _dd_py_invoke_bool_method_by_name(self, "open");
+  if (!self->py.open)
+    return TRUE;
+
+  PyObject *ret;
+  gboolean result = FALSE;
+
+  ret = _py_invoke_function(self->py.open, NULL, self->class, self->super.super.super.id);
```
MrAnno: You can use `_py_invoke_method_by_name`, so you don't have to "cache" this rarely-called method.

furiel: I tried to do so, but it broke the code. The unfortunate thing is that, if there is no `open` method, then I want to return true. However, if exception happens during `open`, I want to return with false. In both cases, `py_invoke_method_by_name` returns NULL.

Adding an extra lookup for existence of `open` method might work. Based on that, I could early return with true.

Do you think that would be better than the caching version? If so, I can try to implement it.

modules/python/python-dest.c
furiel: I left it intentionally because I felt it just happened that all users disappeared from the code. But there is an equivalent chance that in the future there will be some users for it. It is a well written code that worked us well so far.

Still, if you insist, I can remove it.

modules/python/python-dest.c
MrAnno: Oh okay, I understand it. The cache should be OK, if you prefer it this way.

MrAnno:
Commit `python-dest: prepare for using open for success/failure indication` adds `v3.25` macros to `versioning.h`, but  it doesn't bump the config version.

We could bump (only) the config version right after every release, so the config version on master would be always the upcoming version. I guess Kira is currently not ready for such a config version bump.

PR #2999, opened 2019-11-05
fix amqp compatibility problem, amqp detection in cmake, and travis arm build

furiel:
Unfortunately I do not know how the arm travis build broke. The rollback to previous version of pre-commit-hooks did not help. And I could not reproduce the problem myself. So I just made sure pre-commit-hooks is installed only in those cases when needed.

Amqp compiling also broke. Travis changed the librabbitmq-dev version from 0.8.0 to 0.7.1 (downgrade). I fixed the compatibility problem. So from now on syslog-ng can be compiled with both. See the relevant commit for details.

Amqp detection did not use pkg-config in cmake. This means it could not link to librabbitmq with custom location. This PR fixes that as well.


kira-syslogng:
Build SUCCESS

.travis.yml
```diff
@@ -98,6 +98,7 @@ matrix:
       compiler: clang
       script:
         - set -e
+        - pip install --user pre-commit-hooks
```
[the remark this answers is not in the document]
furiel: You are right. I reworded the commit.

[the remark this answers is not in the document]
Kokan: Thank you.

kira-syslogng:
Build SUCCESS

modules/afamqp/afamqp.c
```diff
@@ -448,8 +448,12 @@ afamqp_dd_socket_init(AMQPDestDriver *self)
 
             }
         }
+#ifdef SYSLOG_NG_HAVE_AMQP_SSL_SOCKET_SET_VERIFY_PEER
```
lbudai: IMHO it would be better if we would follow the same pattern as in case of OpenSSL, or GLib: hide compatibility issues behind a separated interface/file.

furiel: Fixed thanks.

kira-syslogng:
Build SUCCESS

modules/afamqp/compat/amqp-compat.c
```diff
@@ -0,0 +1,37 @@
...
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "syslog-ng.h"
+#include "compat/amqp-compat.h"
+
+#include <amqp_ssl_socket.h>
```
Kokan: The `lib/compat/compat.h` header has a nice description how to add additional compat layer, could you please follow that ? https://github.com/syslog-ng/syslog-ng/blob/5a28fa4d75b3b46d946c52cc74f66145e0437b12/lib/compat/compat.h#L44-L63

lbudai: @Kokan : I don't think we should follow that guide in case of modules...
That guide is for the `lib`. `lib` has several dependencies while modules are not.
I think `amqp-compat` is good enough.

furiel: I agree with @lbudai on this one. I would like to keep this as simple as possible. I do not say a proper organization of code is useless, especially if the complexity increases over time. I just think this is good enough now, and it is better spending the APM elsewhere.

lbudai: @Kokan : you concrete requests would be to 
 * rename `compat/amqp-compat.h` to `compat/amqp.h` 
 * include `amqp.h` in `compat/amqp.h`
 * include only `compat/amqp.h` in `afqmqp.c` 
?

Kokan: The "include only compat/amqp.h in afamqp.", and also including `syslog-ng-config.h` instead of `syslog-ng.h`, as that could cause some issue, saidly I do not remember what exactly.

furiel: syslog-ng.h was included because I needed glib.h due to gboolean. syslog-ng.h includes compat/glib.h too. syslog-ng-config.h would pull only the compiler switches.

Kokan: Because I cannot give proper reason, pleas go ahead with this PR.

lbudai: @Kokan : thx

PR #2980, opened 2019-10-21
Improve LogSource's test coverage

MrAnno:
This PR improves the test coverage of `LogSource`. Now only `_flow_control_rate_adjust()` is untested.

Unfortunately, these are not clean unit tests, there is a `TestPipe` object, for example, that emulates the destination queue/message acks. This would be eliminated if we could mock out a few things.

Because I measured the test coverage, I also removed unreachable and unused code, etc.

PR #3001, opened 2019-11-06
Revert "logsource: make log_source_wakeup() private"

MrAnno:
This reverts commit 289408b3caa83613ab1b6f2fceaca7a36ec17e0f.

The wakeup function is not completely private, it is used internally by `LogProtos`, so I should not have made it static.

Sorry about that, I was concentrating on the unit test (#2980).


kira-syslogng:
Build SUCCESS

PR #3002, opened 2019-11-07
confgen: fix crash when parameter is empty

MrAnno:
Previously, when a confgen parameter was empty, syslog-ng crashed in a null-pointer dereference.

Adding an empty parameter (e.g. `parameters()`) does not make sense, but passing `__VARARGS__` to the confgen script is a real use-case, and `__VARARGS__` can be empty.

This PR fixes SCLs, such as the `network-load-balancer()`, where `__VARARGS__` is passed to the confgen script as a parameter:

https://github.com/syslog-ng/syslog-ng/blob/a72580ee65641e8f096958605a2834a346d38128/scl/loadbalancer/plugin.conf#L23-L31

Reproduction:
```
destination d_test {
    network-load-balancer(targets(2));
};
```

### DIFF
--- a/modules/confgen/confgen-plugin.c
+++ b/modules/confgen/confgen-plugin.c
@@ -35,14 +35,21 @@
 static void
 confgen_set_args_as_env(gpointer k, gpointer v, gpointer user_data)
 {
-  gchar buf[1024];
+  if (!v)
+    {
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

Issue #2996, opened 2019-11-04
http output module handling of 400 Errors


rfaircloth-splunk:
## Description of the problem
Common misconfigurations of client or server side for the http module can result in 4XX errors. Presently these errors cause the queue contents to be dropped.

## Proposed solution
Change 4XX errors to behave as 5XX errors which can be retried

szemere:
@furiel could you help me out with this topic? (Sorry for bothering you, this was my only trace: https://github.com/syslog-ng/syslog-ng/commit/f7cee51b70)

furiel:
One thing worth mentioning is that retry still results in message loss. After 3 tries (3 by default) the message will still be dropped.

The reason I chose that heuristic was based on that 4XX errors are non-recoverable, and 5XX are temporary errors. Specially, for 404, I think it is good to drop the message.

At that time when I created that commit it did not occured to me: an interesting question is 401 and 403 (authorization kind of messages). In those cases we can do better: do a disconnect (`LTR_NOT_CONNECTED` in threaded destination terminology) instead of drop or error. With that, the messages would not be lost, just accumulate in a buffer. If the error is fixed, messages will be sent. Still, if the buffer becomes full, this will also lead to message loss. Ultimately, we cannot save the messages 100% percent, just choose how quickly to lose them (instantly, after 3 tries, or after buffer is full).

The ultimate solution would be to add different actions for each http error codes. I once started to work on this: https://github.com/furiel/syslog-ng/commits/http-onerror, but in the end I did not need the implementation, so I abandoned it.

The idea was that we have some (any reasonable default action map), and let users override the actions from configuration. There would be an on-error() option in http destination that describes the actions. Something like this `on-error(401 error-msg("unauthorized") action(drop))`

So yes, I agree that the situation can be improved.

rfaircloth-splunk:
Some 4XX errors are recoverable with intervention for example 401 and 403 are generally due to expired or invalid authentication data. It does require admin intervention but once corrected data can be forwarded presuming persist queue was used. I prefer the buffer full option as it limits the damage of human error for example rotating a password header making a typo and restarting. The other condition is ALB misconfiguration I have seen 4XX errors that will occur temporarily.

Granted if the admin of the environment is not monitoring for lack of event flow we can't save everything but I think we can mitigate better for the transient and human errors.

rfaircloth-splunk:
@furiel the on error configuration would be great to allow handling of problems like misbehaving ALBs

[...]

PR #3007, opened 2019-11-11
http destination: better default action for response codes + response-action option

furiel:
Fixes: https://github.com/syslog-ng/syslog-ng/issues/2996

This patchset intends to improve how http destination reacts to the different response codes. Prior this patch, 2XX codes meant `success`, 4XX meant `drop`. Every other error codes meant `retry`. In some cases these are not appropriate.

This patch use more refined actions. Also, it allows users to override the default behavior with a new option: `response-action`. The following actions are supported: `success`, `retry`, `drop`, `disconnect`. Retry will try to send a message only a limited times (controlled by `retries` option, default is 3), as opposed to `disconnect`, which will try indefinitely.

For example if users want to drop messages when response is 418 ("I'm a teapot") from disconnect to drop, and 404 to retry:
```
  destination {
    http(
      url("http://localhost:8080")
      response-action(418 => drop, 404 => retry)
   );
```

The current table is the following (I will edit the description as the review proceeds).

```
|------+-----------------------------------+------------|
| code | explanation                       | action     |
|------+-----------------------------------+------------|
|  100 | "Continue"                        | disconnect |
|  101 | "Switching Protocols"             | disconnect |
|  102 | "Processing"                      | retry      |
|  103 | "Early Hints"                     | retry      |
|  200 | "OK"                              | success    |
|  201 | "Created"                         | success    |
|  202 | "Accepted"                        | success    |
|  203 | "Non-Authoritative Information"   | success    |
|  204 | "No Content"                      | success    |
|  205 | "Reset Content"                   | success    |
|  206 | "Partial Content"                 | success    |
|  300 | "Multiple Choices"                | disconnect |
|  301 | "Moved Permanently"               | disconnect |
|  302 | "Found"                           | disconnect |
|  303 | "See Other"                       | disconnect |
|  304 | "Not Modified"                    | retry      |
|  307 | "Temporary Redirect"              | disconnect |
|  308 | "Permanent Redirect"              | disconnect |
|  400 | "Bad Request"                     | disconnect |
|  401 | "Unauthorized"                    | disconnect |
|  402 | "Payment Required"                | disconnect |
|  403 | "Forbidden"                       | disconnect |
|  404 | "Not Found"                       | disconnect |
|  405 | "Method Not Allowed"              | disconnect |
|  406 | "Not Acceptable"                  | disconnect |
|  407 | "Proxy Authentication Required"   | disconnect |
|  408 | "Request Timeout"                 | disconnect |
|  409 | "Conflict"                        | retry      |
|  410 | "Gone"                            | drop       |
|  411 | "Length Required"                 | disconnect |
|  412 | "Precondition Failed"             | disconnect |
|  413 | "Payload Too Large"               | disconnect |
|  414 | "URI Too Long"                    | disconnect |
|  415 | "Unsupported Media Type"          | disconnect |
|  416 | "Range Not Satisfiable"           | drop       |
|  417 | "Expectation Failed"              | disconnect |
|  418 | "I'm a teapot"                    | disconnect |
|  421 | "Misdirected Request"             | disconnect |
|  422 | "Unprocessable Entity"            | drop       |
|  423 | "Locked"                          | disconnect |
|  424 | "Failed Dependency"               | drop       |
|  425 | "Too Early"                       | drop       |
|  426 | "Upgrade Required"                | disconnect |
|  428 | "Precondition Required"           | retry      |
|  429 | "Too Many Requests"               | disconnect |
|  431 | "Request Header Fields Too Large" | disconnect |
|  451 | "Unavailable For Legal Reasons"   | drop       |
|  500 | "Internal Server Error"           | disconnect |
|  501 | "Not Implemented"                 | disconnect |
|  502 | "Bad Gateway"                     | disconnect |
|  503 | "Service Unavailable"             | disconnect |
|  504 | "Gateway Timeout"                 | retry      |
|  505 | "HTTP Version Not Supported"      | disconnect |
|  506 | "Variant Also Negotiates"         | disconnect |
|  507 | "Insufficient Storage"            | disconnect |
|  508 | "Loop Detected"                   | drop       |
|  510 | "Not Extended"                    | disconnect |
|  511 | "Network Authentication Required" | disconnect |
|------+-----------------------------------+------------|

If these values do not match: we use success for 2XX, and disconnect otherwise.
```


furiel:
I wrote short explanations why I chose each actions per code. The descriptions are from [wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes).

<details>

# 1xx Informational response
An informational response indicates that the request was received and understood. It is issued on a provisional basis while request processing continues. It alerts the client to wait for a final response. The message consists only of the status line and optional header fields, and is terminated by an empty line. As the HTTP/1.0 standard did not define any 1xx status codes, servers must not[note 1] send a 1xx response to an HTTP/1.0 compliant client except under experimental conditions.[4]

## 100 Continue
Description: The server, has received the request headers and the client should proceed to send the request body (in the case of a request for which a body needs to be sent; for example, a POST request). Sending a large request body to a server after a request has been rejected for inappropriate headers would be inefficient. To have a server check the request's headers, a client must send Expect: 100-continue as a header in its initial request and receive a 100 Continue status code in response before sending the body. If the client receives an error code such as 403 (Forbidden) or 405 (Method Not Allowed) then it shouldn't send the request's body. The response 417 Expectation Failed indicates that the request should be repeated without the Expect header as it indicates that the server doesn't support expectations (this is the case, for example, of HTTP/1.0 servers).[5]

Syslog-ng should not send continue header. In libcurl, continue header is disabled by setting Expect: header to empty, which is done by syslog-ng, this time of writing If this error code happens, that means bug in syslog-ng code, and not problem with the data. Hence we should not lose messages. Action: disconnect.

## 101 Switching Protocols
Description: The requester has asked the server to switch protocols and the server has agreed to do so.[6]

This should be initiated by syslog-ng, which is not implemented. If this error code happends that means bug in syslog-ng code, and not problem with the data. Hence we should not lose messages. Action: disconnect.

## 102 Processing (WebDAV; RFC 2518)
A WebDAV request may contain many sub-requests involving file operations, requiring a long time to complete the request. This code indicates that the server has received and is processing the request, but no response is available yet.[7] This prevents the client from timing out and assuming the request was lost.

Syslog-ng cannot handle such error code. As the server does not signal errors, the request might result in both success and error. Action: retry

## 103 Early Hints (RFC 8297)
Used to return some response headers before final HTTP message.

Syslog-ng cannot handle such error code. As the server does not signal errors, the request might result in both success and error. Action: retry

# 2xx Success
This class of status codes indicates the action requested by the client was received, understood and accepted.[2]

Success branch.

# 3xx Redirection

This class of status code indicates the client must take additional action to complete the request. Many of these status codes are used in URL redirection.[2]

A user agent may carry out the additional action with no user interaction only if the method used in the second request is GET or HEAD. A user agent may automatically redirect a request. A user agent should detect and intervene to prevent cyclical redirects.[19]

## 300 Multiple Choices
Description: Indicates multiple options for the resource from which the client may choose (via agent-driven content negotiation). For example, this code could be used to present multiple video format options, to list files with different filename extensions, or to suggest word-sense disambiguation.[20]

This is an configuration error. Users might need to add additional headers or data in the body. There is no sign that the message is invalid, so we do not want to lose it. Action: disconnect.

## 301 Moved Permanently
Description: This and all future requests should be directed to the given URI.[21]

This is a configuration error. Accept-redirects must be enabled or the uri must be updated in the configuration. Action: disconnect.

## 302 Found (Previously "Moved temporarily")
Description: Tells the client to look at (browse to) another URL. 302 has been superseded by 303 and 307. This is an example of industry practice contradicting the standard. The HTTP/1.0 specification (RFC 1945) required the client to perform a temporary redirect (the original describing phrase was "Moved Temporarily"),[22] but popular browsers implemented 302 with the functionality of a 303 See Other. Therefore, HTTP/1.1 added status codes 303 and 307 to distinguish between the two behaviours.[23] However, some Web applications and frameworks use the 302 status code as if it were the 303.[24]

This is a configuration error. Accept-redirects must be enabled or the uri must be updated in the configuration. Action: disconnect.

## 303 See Other (since HTTP/1.1)
Description: The response to the request can be found under another URI using the GET method. When received in response to a POST (or PUT/DELETE), the client should presume that the server has received the data and should issue a new GET request to the given URI.[25]

This is a configuration error. Accept-redirects must be enabled or the uri must be updated in the configuration. Action: disconnect.

## 304 Not Modified (RFC 7232)
Description: Indicates that the resource has not been modified since the version specified by the request headers If-Modified-Since or If-None-Match. In such case, there is no need to retransmit the resource since the client still has a previously-downloaded copy.[26]

This should mean server thinks this request is already served, and the result is the same as previously. This can be success or failure, we do not know at this point. We assume the worse. Action: retry.

## 305 Use Proxy (since HTTP/1.1)
Description: The requested resource is available only through a proxy, the address for which is provided in the response. For security reasons, many HTTP clients (such as Mozilla Firefox and Internet Explorer) do not obey this status code.[27]

This is a user deployment error. If this error code is received, syslog-ng should not lose message. Action: disconnect.

## 306 Switch Proxy
Description: No longer used. Originally meant "Subsequent requests should use the specified proxy."[28]

This is a user deployment error. If this error code is received, syslog-ng should not lose message. Action: disconnect.

## 307 Temporary Redirect (since HTTP/1.1)
Description: In this case, the request should be repeated with another URI; however, future requests should still use the original URI. In contrast to how 302 was historically implemented, the request method is not allowed to be changed when reissuing the original request. For example, a POST request should be repeated using another POST request.[29]

This is a configuration error. Accept-redirects must be enabled or the uri must be updated in the configuration. Action: disconnect.

## 308 Permanent Redirect (RFC 7538)
Description: The request and all future requests should be repeated using another URI. 307 and 308 parallel the behaviors of 302 and 301, but do not allow the HTTP method to change. So, for example, submitting a form to a permanently redirected resource may continue smoothly.[30]

This is a configuration error. Accept-redirects must be enabled or the uri must be updated in the configuration. Action: disconnect.

# 4xx Client errors
This class of status code is intended for situations in which the error seems to have been caused by the client. Except when responding to a HEAD request, the server should include an entity containing an explanation of the error situation, and whether it is a temporary or permanent condition. These status codes are applicable to any request method. User agents should display any included entity to the user.[31]

This can mean either configuration error or syslog-ng bug. Syslog-ng should not lose the message. Action: disconnect.

## 400 Bad Request
Description: The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, size too large, invalid request message framing, or deceptive request routing).[32]

This is a configuration error. The error is not with the message, but the way it is transferred. Syslog-ng should not lose the message. Action: disconnect.

## 401 Unauthorized (RFC 7235)
Description: Similar to 403 Forbidden, but specifically for use when authentication is required and has failed or has not yet been provided. The response must include a WWW-Authenticate header field containing a challenge applicable to the requested resource. See Basic access authentication and Digest access authentication.[33] 401 semantically means "unauthorised",[34] the user does not have valid authentication credentials for the target resource.
Description: Note: Some sites incorrectly issue HTTP 401 when an IP address is banned from the website (usually the website domain) and that specific address is refused permission to access a website.[citation needed]

This is a configuration error. The error is not with the message, but the way it is transferred. Syslog-ng should not lose the message. Action: disconnect.

## 402 Payment Required
Description: Reserved for future use. The original intention was that this code might be used as part of some form of digital cash or micropayment scheme, as proposed, for example, by GNU Taler,[35] but that has not yet happened, and this code is not usually used. Google Developers API uses this status if a particular developer has exceeded the daily limit on requests.[36] Sipgate uses this code if an account does not have sufficient funds to start a call.[37] Shopify uses this code when the store has not paid their fees and is temporarily disabled.[38] Stripe uses this code for failed payments where parameters were correct, for example blocked fraudulent payments.[39]

This is a configuration error. The error is not with the message, but the way it is transferred. Syslog-ng should not lose the message. Action: disconnect.

## 403 Forbidden
Description: The request contained valid data and was understood by the server, but the server is refusing action. This may be due to the user not having the necessary permissions for a resource or needing an account of some sort, or attempting a prohibited action (e.g. creating a duplicate record where only one is allowed). This code is also typically used if the request provided authentication via the WWW-Authenticate header field, but the server did not accept that authentication. The request should not be repeated.

This is a configuration error. The error is not with the message, but the way it is transferred. Syslog-ng should not lose the message. Action: disconnect.

## 404 Not Found
Description: The requested resource could not be found but may be available in the future. Subsequent requests by the client are permissible.

This is a configuration error. The error is not with the message. Syslog-ng should not lose the message. Action: disconnect.

## 405 Method Not Allowed
Description: A request method is not supported for the requested resource; for example, a GET request on a form that requires data to be presented via POST, or a PUT request on a read-only resource.

This is a configuration error. The error is not with the message. Syslog-ng should not lose the message. Action: disconnect.

## 406 Not Acceptable
Description: The requested resource is capable of generating only content not acceptable according to the Accept headers sent in the request.[40] See Content negotiation.

This is a deployment or configuration error. The error is not with the message. Syslog-ng should not lose the message. Action: disconnect.

## 407 Proxy Authentication Required (RFC 7235)
Description: The client must first authenticate itself with the proxy.[41]

This is a configuration error. The error is not with the message. Syslog-ng should not lose the message. Action: disconnect.

## 408 Request Timeout
Description: The server timed out waiting for the request. According to HTTP specifications: "The client did not produce a request within the time that the server was prepared to wait. The client MAY repeat the request without modifications at any later time."[42]

Syslog-ng could not finish the request for some reason within a timeout. Either a timeout should be increased or there is bug within syslog-ng. Syslog-ng should not lose message. Action: disconnect.

## 409 Conflict
Description: Indicates that the request could not be processed because of conflict in the current state of the resource, such as an edit conflict between multiple simultaneous updates.

The conflict mentioned above should be temporary. Action: retry.

## 410 Gone
Description: Indicates that the resource requested is no longer available and will not be available again. This should be used when a resource has been intentionally removed and the resource should be purged. Upon receiving a 410 status code, the client should not request the resource in the future. Clients such as search engines should remove the resource from their indices.[43] Most use cases do not require clients and search engines to purge the resource, and a "404 Not Found" may be used instead.

This indicates that syslog-ng must not resend the message in any circumstances. Action: drop.

## 411 Length Required
Description: The request did not specify the length of its content, which is required by the requested resource.[44]

This means bug in syslog-ng. Syslog-ng should not lose message. Action: disconnect.

## 412 Precondition Failed (RFC 7232)
Description: The server does not meet one of the preconditions that the requester put on the request header fields.[45][46]

This means configuration error. Syslog-ng should not lose message. Action: disconnect.

## 413 Payload Too Large (RFC 7231)
Description: The request is larger than the server is willing or able to process. Previously called "Request Entity Too Large".[47]

This means configuration error. Possibly batch-size or batch-bytes must be set lower. Syslog-ng should not lose message. Action: disconnect.

## 414 URI Too Long (RFC 7231)
Description: The URI provided was too long for the server to process. Often the result of too much data being encoded as a query-string of a GET request, in which case it should be converted to a POST request.[48] Called "Request-URI Too Long" previously.[49]

This means configuration error. Syslog-ng should not lose message. Action: disconnect.

## 415 Unsupported Media Type (RFC 7231)
Description: The request entity has a media type which the server or resource does not support. For example, the client uploads an image as image/svg+xml, but the server requires that images use a different format.[50]

This means configuration error. Syslog-ng should not lose message. Action: disconnect.

## 416 Range Not Satisfiable (RFC 7233)
Description: The client has asked for a portion of the file (byte serving), but the server cannot supply that portion. For example, if the client asked for a part of the file that lies beyond the end of the file.[51] Called "Requested Range Not Satisfiable" previously.[52]

The request does not make sense, possibly due to the message. Action: drop.

## 417 Expectation Failed
Description: The server cannot meet the requirements of the Expect request-header field.[53]

This means syslog-ng error. This time of writing, Expect header set to empty in syslog-ng. The problem is not with the message, but the way it is transfered. Syslog-ng should not lose message. Action: disconnect.

## 418 I'm a teapot (RFC 2324, RFC 7168)
Description: This code was defined in 1998 as one of the traditional IETF April Fools' jokes, in RFC 2324, Hyper Text Coffee Pot Control Protocol, and is not expected to be implemented by actual HTTP servers. The RFC specifies this code should be returned by teapots requested to brew coffee.[54] This HTTP status is used as an Easter egg in some websites, including Google.com.[55][56]

Any action should be fine. I chose disconnect.

## 421 Misdirected Request (RFC 7540)
Description: The request was directed at a server that is not able to produce a response[57] (for example because of connection reuse).[58]

This means configuration error. The error is not with the message, but the way it is transferred. Syslog-ng should not lose the message. Action: disconnect.

## 422 Unprocessable Entity (WebDAV; RFC 4918)
Description: The request was well-formed but was unable to be followed due to semantic errors.[17]

The problem is with the message. Action: drop.

## 423 Locked (WebDAV; RFC 4918)
Description: The resource that is being accessed is locked.[17]

This is a deployment problem. Syslog-ng should not lose message. Action: disconnect.

## 424 Failed Dependency (WebDAV; RFC 4918)
Description: The request failed because it depended on another request and that request failed (e.g., a PROPPATCH).[17]

The problem is with the message. Action: drop.

## 425 Too Early (RFC 8470)
Description: Indicates that the server is unwilling to risk processing a request that might be replayed.

This is a security problem. The message must not be resent in any circumstances. Action: drop.

## 426 Upgrade Required
Description: The client should switch to a different protocol such as TLS/1.0, given in the Upgrade header field.[59]

This means configuration error. The error is not with the message, but the way it is transferred. Syslog-ng should not lose the message. Action: disconnect.

## 428 Precondition Required (RFC 6585)
Description: The origin server requires the request to be conditional. Intended to prevent the 'lost update' problem, where a client GETs a resource's state, modifies it, and PUTs it back to the server, when meanwhile a third party has modified the state on the server, leading to a conflict.[60]

This problem should be a temporary problem. Action: retry.

## 429 Too Many Requests (RFC 6585)
Description: The user has sent too many requests in a given amount of time. Intended for use with rate-limiting schemes.[60]

This is a deployment problem. Syslog-ng should not lose message. Action: disconnect.

## 431 Request Header Fields Too Large (RFC 6585)
Description: The server is unwilling to process the request because either an individual header field, or all the header fields collectively, are too large.[60]

This is a deployment problem. Syslog-ng should not lose message. Action: disconnect.

## 451 Unavailable For Legal Reasons (RFC 7725)
Description: A server operator has received a legal demand to deny access to a resource or to a set of resources that includes the requested resource.[61] The code 451 was chosen as a reference to the novel Fahrenheit 451 (see the Acknowledgements in the RFC).

This is a security problem. The message must not be resent in any circumstances. Action: drop.

# 5xx Server errors

The server failed to fulfill a request.[62]

Response status codes beginning with the digit "5" indicate cases in which the server is aware that it has encountered an error or is otherwise incapable of performing the request. Except when responding to a HEAD request, the server should include an entity containing an explanation of the error situation, and indicate whether it is a temporary or permanent condition. Likewise, user agents should display any included entity to the user. These response codes are applicable to any request method.[63]

## 500 Internal Server Error
Description: A generic error message, given when an unexpected condition was encountered and no more specific message is suitable.[64]

Without further information, we do not want to lose message. Action: disconnect.

## 501 Not Implemented
Description: The server either does not recognize the request method, or it lacks the ability to fulfil the request. Usually this implies future availability (e.g., a new feature of a web-service API).[65]

This means configuration error. The error is not with the message. Syslog-ng should not lose the message. Action: disconnect.

## 502 Bad Gateway
Description: The server was acting as a gateway or proxy and received an invalid response from the upstream server.[66]

This means configuration error. The error is not with the message. Syslog-ng should not lose the message. Action: disconnect.

## 503 Service Unavailable
Description: The server cannot handle the request (because it is overloaded or down for maintenance). Generally, this is a temporary state.[67]

This means deployment error. The error is not with the message. Syslog-ng should not lose the message. Action: disconnect.

## 504 Gateway Timeout
Description: The server was acting as a gateway or proxy and did not receive a timely response from the upstream server.[68]

This means deployment error. Still the message might have arrived and processed by the upstream server. We do not know that at this point. Action: retry.

## 505 HTTP Version Not Supported
Description: The server does not support the HTTP protocol version used in the request.[69]

This means configuration error. The error is not with the message. Syslog-ng should not lose the message. Action: disconnect.

## 506 Variant Also Negotiates (RFC 2295)
Description: Transparent content negotiation for the request results in a circular reference.[70]

This is not supported by syslog-ng. The error is not with the message. Syslog-ng should not lose the message. Action: disconnect.

## 507 Insufficient Storage (WebDAV; RFC 4918)
Description: The server is unable to store the representation needed to complete the request.[17]

This means deployment error. Syslog-ng should not lose messages. Action: disconnect.

## 508 Loop Detected (WebDAV; RFC 5842)
Description: The server detected an infinite loop while processing the request (sent instead of 208 Already Reported).

The problem is with the message. Action: drop.

## 510 Not Extended (RFC 2774)
Description: Further extensions to the request are required for the server to fulfil it.[71]

This means deployment error. Syslog-ng should not lose messages. Action: disconnect.

## 511 Network Authentication Required (RFC 6585)
Description: The client needs to authenticate to gain network access. Intended for use by intercepting proxies used to control access to the network (e.g., "captive portals" used to require agreement to Terms of Service before granting full Internet access via a Wi-Fi hotspot).[60]

This means configuration problem. The problem is not with the message. Action: disconnect.
</details>

furiel:
@kira-syslogng retest this please test branch=changing-http-response-action;

modules/http/http-grammar.ym
```diff
@@ -109,6 +120,24 @@ http_options
...
 
+on_error_action
+    : KW_SUCCESS { $$ = on_error_success; }
+    | KW_RETRY { $$ = on_error_retry; }
+    | KW_DROP { $$ = on_error_drop; }
+    | KW_DISCONNECT { $$ = on_error_disconnect; }
+    ;
+
+on_error_block
+    : '(' { last_on_error_params = g_new0(OnErrorParams, 1); }
```
bazsi: I would try to avoid allocation here, as syntax errors would generate leaks.

modules/http/http-grammar.ym
```diff
@@ -109,6 +120,24 @@ http_options
...
+       positive_integer { last_on_error_params->status_code = $3; }
+       KW_ACTION '(' on_error_action ')' { last_on_error_params->action = $7; }
+      ')' {
+             CHECK_ERROR(last_on_error_params->action, @1, "action is mandatory parameter");
```
bazsi: I don't see how action would be unset here.

modules/http/http-grammar.ym
```diff
@@ -129,6 +158,7 @@ http_option
     | http_tls_option
     | KW_TLS '(' http_tls_options ')'
     | { last_template_options = http_dd_get_template_options(last_driver); } template_option
+    | KW_ON_HTTP_ERROR on_error_block
```
bazsi: I would like to have the grammar improved somewhat. The response code from the server does not necessarily indicate an error, in http terms the code is a "status" code. Maybe even using the "arrow" syntax that we introduced for name-value pairs would be useful here.

Something like this:

```
http(.... 
         handle-response(404 => drop,
                                     418 => disconnect)
```

lib/logthrdest/logthrdestdrv.c
```diff
@@ -33,6 +33,24 @@
 
 static void _init_stats_key(LogThreadedDestDriver *self, StatsClusterKey *sc_key);
 
+const gchar *
+logthreadedresult_to_str(LogThreadedResult self)
```
bazsi: can you please use underscores in the name to make this more readable?

modules/http/http-worker.c
```diff
@@ -246,51 +246,108 @@ _add_message_to_batch(HTTPDestinationWorker *self, LogMessage *msg)
...
+find_http_code_in_list(glong http_code, glong list[])
+{
+  for (gint i = 0; list[i] != -1; i++)
+    if (list[i] == http_code)
+      return TRUE;
+  return FALSE;
+}
+
+LogThreadedResult
+default_1XX(HTTPDestinationWorker *self, const gchar *url, glong http_code)
```
bazsi: shouldnt this be static?

modules/http/http-worker.c
```diff
@@ -246,51 +246,108 @@ _add_message_to_batch(HTTPDestinationWorker *self, LogMessage *msg)
     }
 }
 
+static gboolean
+find_http_code_in_list(glong http_code, glong list[])
```
bazsi: please use a _ prefix to indicate this is a private function.

modules/http/http.c
```diff
@@ -24,6 +24,12 @@
 #include "http-worker.h"
 
 /* HTTPDestinationDriver */
+void
+http_dd_set_on_error_params(LogDriver *d, OnErrorParams *on_error_params)
```
bazsi: this is not just a setter, rather it inserts an entry into the on-error hash, so the name should indicate that.

kira-syslogng:
Build FAILURE

modules/http/on-error-handler.h
```diff
@@ -0,0 +1,57 @@
...
+OnErrorResult on_error_drop(gpointer user_data);
+OnErrorResult on_error_disconnect(gpointer user_data);
+
+typedef struct
+{
+  gint status_code;
+  OnErrorAction action;
+  gpointer user_data;
+} OnErrorParams;
+
```
bazsi: just as with the grammar, I would rather use the term "status" instead of "error" in the naming of this module.

bazsi:
btw, there's one more use-case that is somewhat related and even if not addressed by this patchset, should be considered.

It would be great if the HTTP destination could switch between URLs even if the requests are successful. e.g. we would need to send a "Connection: close" header and switch to the next URL when the same worker receives the next batch. The background of this use-case is when using a load balancer, which balances connections and not requests. In those cases we would need to let the LB pick another server. Normally, we might also specify how many requests should be going to the same URL. (e.g. trigger a close after 10 requests).

This will also resolve the case where we have more workers than URLs.

bazsi:
@rfaircloth-splunk please check the default handling of status codes if it matches HEC's expectations.

modules/http/http-grammar.ym
furiel: I would like to leave some way to further expand the match logic. The match logic might depend on the headers of the exact content of the http response. In my original branch, I had an additional error-msg(string) that would allow users to further branch per status code, based on the body of the response. I did not send that part in the end because it caused significant code complexity, and was not necessary for the resolution of #2996 . But later we might want to add such thing incrementally. That's why I used the current syntax.

Still, currently there is an unused user_data passed into the action callbacks. That could be later used to transfer the header and body content. So we can say the logic that further dispatches must be called inside the callbacks. With that we can save the `=>` syntax.

Which one is preferable in your opinion?

bazsi: I don't think we should branch on anything but the status code. REST style APIs uses this code to convey information about what has happened to the request.

Headers and response contents could be useful for a more active logic, but I don't think those should be available via configuration. I can imagine plugins on top of http() that processes the response in depth (e.g. consuming a per-message status/acknowledgement from the server returned via  JSON array in the response), but I think should we want to support that we would need to implement this via a higher-order plugin: one that can potentially be written in Python or another language. The "event" mechanism that we discussed for file-rotation could facilitate something like this in the future.

modules/http/http.c
furiel: Fixed.

modules/http/on-error-handler.h
furiel: Fixed.

lib/logthrdest/logthrdestdrv.c
furiel: Fixed.

modules/http/http-worker.c
furiel: Fixed.

modules/http/http-grammar.ym
furiel: Fixed.

modules/http/http-worker.c
furiel: Fixed.

modules/http/http-grammar.ym
furiel: Changed code to not to use allocation. Renamed add to insert.

Action will be unset by the hashtable.

kira-syslogng:
Build FAILURE

modules/http/http-grammar.ym
furiel: I understand and agree. I updated the PR with the `=>` syntax.

furiel:
@kira-syslogng retest this please test branch=changing-http-response-action;

kira-syslogng:
Build FAILURE

furiel:
> It would be great if the HTTP destination could switch between URLs even if the requests are successful. e.g. we would need to send a "Connection: close" header and switch to the next URL when the same worker receives the next batch. The background of this use-case is when using a load balancer, which balances connections and not requests. In those cases we would need to let the LB pick another server. Normally, we might also specify how many requests should be going to the same URL. (e.g. trigger a close after 10 requests).

@bazsi I like the idea. Although I am not planning to implement this now, I would like to make sure the architecture is open enough so that use case could be added.

What we would need is that the `action` function needs to have access to the response headers and the body. I don't think we would have a problem with that: the currently unused `user_data` could be used to pass extra information around.

I am a little concerned about one thing: that collecting the response headers and body takes resource, that might hit the performance. If it does, is there a way we can avoid it?

What we could do is to make header and body collection optional: we do not want to affect http destinations that do not utilize this feature. Specially I am thinking that, http destination would have special options that make different things to be collectible. Like

```
http {
    on-http-response(KW_HEADERS, KW_BODY,  # <--
                     200 => success, 400 => drop);
};
```
That would turn the headers and body collection on, and http destination would pass them as an associative array/list like structure in the user data.

So in the end, I believe we could support such things.

furiel:
@kira-syslogng retest this please test branch=changing-http-response-action;

furiel:
@kira-syslogng retest this please test branch=changing-http-response-action;

kira-syslogng:
Build SUCCESS

rfaircloth-splunk:
I like the new actions in the response table draft above

modules/http/http-worker.c
```diff
@@ -419,7 +441,11 @@ _flush_on_target(HTTPDestinationWorker *self, HTTPLoadBalancerTarget *target)
   if (http_code == 401 && owner->auth_header)
     return _renew_header(owner);
 
-  return map_http_status_to_worker_status(self, target->url, http_code);
+  HttpResponseHandler *response_handler = NULL;
+  if ((response_handler = http_response_handlers_lookup(owner->response_handlers, http_code)))
+    return _map_http_result(response_handler->action(response_handler->user_data));
+
+  return default_map_http_status_to_worker_status(self, target->url, http_code);
```
lbudai: It would be even more readable if this would be factored out as a separate `map_http_result` function.
something like similar:

```
static LogThreadedResult
_try_to_custom_map_http_status_to_worker_status(HttpDestinationWorker *self, glong http_code)
{
  HttpResponseHandler *response_handler = NULL;

  if ((response_handler = http_response_handlers_lookup(owner->response_handlers, http_code)))
    return _custom_map_http_result(response_handler->action(response_handler->user_data));

  return LTR_MAX;
}
static LogThreadedResult
_map_http_status_code(HttpDestinationWorker *self, const gchar *url, glong http_code)
{
  LogThreadedResult result = _try_to_custom_map_http_status_to_worker_status(self, http_code);
  
  if (result != LTR_MAX)
      return result;
  
  return default_map_http_status_to_worker_status(self, url, http_code);
}
```

modules/http/response-handler.c
```diff
@@ -45,3 +45,44 @@ http_result_disconnect(gpointer user_data)
...
+  HttpResponseHandler *clone = g_new(HttpResponseHandler, 1);
+  *clone = *self;
+  return clone;
+}
+
+void
+http_response_handlers_insert(HttpResponseHandlers *self, HttpResponseHandler *response_handler)
+{
+  HttpResponseHandler *clone = _clone(response_handler);
+  gint *status_code = g_new(gint, 1);
```
lbudai: Shouldn't be this `glong`?

lbudai:
@furiel : nice work!

modules/http/http-worker.c
furiel: Fixed.

furiel:
@kira-syslogng retest this please test branch=changing-http-response-action;

kira-syslogng:
Build SUCCESS

pzoleex:
My notices:

`on-http-response(418 => drop, 404 => retry)`

However I like this style (418 => drop), but for me this seems to be very unusal in syslog-ng conf.

But what if I want to add more codes, like drop in case of 418, 419, 420... then 418=>drop, 419=>drop... can be hard and easy to wrong it.

wouldn't be better this form?
```
drop-on-http-response(418, 419, 420...)
retry-on-http-response(401, 402, 403...)
```

and as far as I can see in the code, you already have this list like:
`static glong drops[] = {410, 416, 422, 424, 425, 451, -1};`

so basically, instead we could use and document that this is the default and user can override it:
`drop-on-http-response(410, 416, 422, 424, 425, 451)`
and later if we want to change something for some reason, we can just change in in the grammar instead of the C code.

What do you think?

furiel:
I would like to keep the possibility open, that action could become a plugin later. So the action cannot be part of larger token.

I intended to choose correct default actions by default per status codes, so that users do not need to change too much. If they need bulk-change actions, this PR fails its purpose at some point. Still, I cannot be sure everything will work out as expected, so I do not really have anything against supporting bulk-changing the actions.
  
I created a patch that trivially extends the arrow syntax:
https://github.com/furiel/syslog-ng/commit/fa82c5fd90c963b2453704c75e7bc21dd559817f

We can write: `(404, 200) => disconnect`.

The bulk and single options can be used together too:
```
http(
   url(on-http-response((404, 200) => disconnect, 404 => success))
);
```

I did not push this one, because I would like to ask your opinion first. Specially, I would like to draw @alltilla 's attention. I don't know how much such thing can mess up the config option database extraction logic: that there is no option name before an opening bracket.

alltilla:
> I would like to keep the possibility open, that action could become a plugin later. So the action cannot be part of larger token.
> 
> I intended to choose correct default actions by default per status codes, so that users do not need to change too much. If they need bulk-change actions, this PR fails its purpose at some point. Still, I cannot be sure everything will work out as expected, so I do not really have anything against supporting bulk-changing the actions.
> 
> I created a patch that trivially extends the arrow syntax:
> [furiel@fa82c5f](https://github.com/furiel/syslog-ng/commit/fa82c5fd90c963b2453704c75e7bc21dd559817f)
> 
> We can write: `(404, 200) => disconnect`.
> 
> The bulk and single options can be used together too:
> 
> ```
> http(
>    url(on-http-response((404, 200) => disconnect, 404 => success))
> );
> ```
> 
> I did not push this one, because I would like to ask your opinion first. Specially, I would like to draw @alltilla 's attention. I don't know how much such thing can mess up the config option database extraction logic: that there is no option name before an opening bracket.

It is completely okay, we have such things already:
```
destination kafka-c(
  bootstrap-servers/kafka-bootstrap-servers(<string>)
...
  config/option(
    <empty>
    <string> <arrow> <string-or-number>
    <string> <string-or-number>
    <string>(<string-or-number>)
  )
```
Note: `<arrow>` will have a mapping to `=>` later :)

gaborznagy:
After our discussion, what about this idea?
Instead of this:
```
drop-on-http-response(418, 419, 420...)
retry-on-http-response(401, 402, 403...)
```
This:
```
on-http-response(
  drop(400, 401)
  retry(402, 403)
  ... # LL_IDENTIFIER
)
```
- We think that it will be more common to override **multiple** default actions than we think. I don't think It's because the default value is wrong, but rather users will like to override multiple response codes. It's a good question whether we would like to make this easy with an intuitive syntax, or force them to add `code => response` pairs for each difference from the default.
- I've added the `on-http-response()` context to the grammar to support having plugins, without sacrificing the unknown keyword error handling.
I mean, if you have the `LL_IDENTIFIER` on the same level as the options, and the user mistypes an option name (e.g. `user_name` instead of `user` or `user_agent`), then from this change the user faces a "plugin not found: user_name" instead of "unknown identifier"

bazsi:
> After our discussion, what about this idea?
> Instead of this:
> 
> ```
> drop-on-http-response(418, 419, 420...)
> retry-on-http-response(401, 402, 403...)
> ```
> 
> This:
> 
> ```
> on-http-response(
>   drop(400, 401)
>   retry(402, 403)
>   ... # LL_IDENTIFIER
> )
> ```
> 
>     * We think that it will be more common to override **multiple** default actions than we think. I don't think It's because the default value is wrong, but rather users will like to override multiple response codes. It's a good question whether we would like to make this easy with an intuitive syntax, or force them to add `code => response` pairs for each difference from the default.

Hmm, I think all destinations that use the HTTP transport should be wrapped as an SCL and provided to the user as a specific driver, in which case overrides will be nicely encapsulated and not make it into user's configuration. The arrow syntax is one that we have discussed at the kafka destination which as a pattern would be extended to python() and java() option blocks, in which case it will become more often.

I think the "map" concept is becoming increasingly widespread in syslog-ng, so it deserves its own syntax.


> 
>     * I've added the `on-http-response()` context to the grammar to support having plugins, without sacrificing the unknown keyword error handling.
>       I mean, if you have the `LL_IDENTIFIER` on the same level as the options, and the user mistypes an option name (e.g. `user_name` instead of `user` or `user_agent`), then from this change the user faces a "plugin not found: user_name" instead of "unknown identifier"

I don't understand this last paragraph.

modules/http/http-grammar.ym
```diff
@@ -73,10 +76,17 @@
 %token KW_WORKERS
 %token KW_ACCEPT_REDIRECTS
 %token KW_AUTH_HEADER
+%token KW_ON_HTTP_RESPONSE
```
bazsi: why don't we call this response-action() in the configuration file as well?

e.g. instead of `on-http-response()` we would just use `response-action(404 => retry)` I think that's easier to understand. I find the "on-" variant a bit "programmery".

modules/http/http-worker.c
```diff
@@ -345,6 +402,59 @@ _debug_response_info(HTTPDestinationWorker *self, HTTPLoadBalancerTarget *target
...
+{
+  HttpResult result = response_handler->action(response_handler->user_data);
+  g_assert(result < HTTP_RESULT_MAX);
+  HTTPDestinationDriver *owner = (HTTPDestinationDriver *) self->super.owner;
+  glong http_code = response_handler->status_code;
+
+  switch (result)
+    {
+    case HTTP_RESULT_SUCCESS:
+      msg_debug("http: handled by on_error",
```
bazsi: the log message could be clarified, the option is not on_error anymore.

modules/http/http-worker.c
```diff
@@ -345,6 +402,59 @@ _debug_response_info(HTTPDestinationWorker *self, HTTPLoadBalancerTarget *target
...
+                evt_tag_str("decision", "success"),
+                evt_tag_str("url", owner->url),
+                evt_tag_int("status_code", http_code),
+                evt_tag_str("driver", owner->super.super.super.id),
+                log_pipe_location_tag(&owner->super.super.super.super));
+      return LTR_SUCCESS;
+
+    case HTTP_RESULT_RETRY:
+      msg_notice("http: handled by on_error",
```
bazsi: likewise.

modules/http/http-worker.c
```diff
@@ -345,6 +402,59 @@ _debug_response_info(HTTPDestinationWorker *self, HTTPLoadBalancerTarget *target
...
+                 evt_tag_str("decision", "drop"),
+                 evt_tag_str("url", owner->url),
+                 evt_tag_int("status_code", http_code),
+                 evt_tag_str("driver", owner->super.super.super.id),
+                 log_pipe_location_tag(&owner->super.super.super.super));
+      return LTR_DROP;
+
+    case HTTP_RESULT_DISCONNECT:
+      msg_notice("http: handled by on_error",
```
bazsi: this is the last occassion.

modules/http/http-worker.c
```diff
@@ -345,6 +402,59 @@ _debug_response_info(HTTPDestinationWorker *self, HTTPLoadBalancerTarget *target
...
+  HttpResult result = response_handler->action(response_handler->user_data);
+  g_assert(result < HTTP_RESULT_MAX);
+  HTTPDestinationDriver *owner = (HTTPDestinationDriver *) self->super.owner;
+  glong http_code = response_handler->status_code;
+
+  switch (result)
+    {
+    case HTTP_RESULT_SUCCESS:
+      msg_debug("http: handled by on_error",
+                evt_tag_str("decision", "success"),
```
bazsi: decision is probably better named "action", like in the naming of the grammar rules.

modules/http/response-handler.c
```diff
@@ -0,0 +1,88 @@
...
```
bazsi: why do we need to allocate this gint? it is already embedded in the response_handler struct, so it could just be used from there and the key wouldn't need a free function.

bazsi:
Apart from the vocabulary changes I added via comments, this branch is shaping up pretty nicely.

gaborznagy:
>Hmm, I think all destinations that use the HTTP transport should be wrapped as an SCL and provided to the user as a specific driver, in which case overrides will be nicely encapsulated and not make it into user's configuration. The arrow syntax is one that we have discussed at the kafka destination which as a pattern would be extended to python() and java() option blocks, in which case it will become more often.
>
>I think the "map" concept is becoming increasingly widespread in syslog-ng, so it deserves its own syntax.

You mean, allowing lists `drop-on-http-response(400, 404)` is not needed, users shouldn't use `on-http-response()`, it's intended for SCLs?

gaborznagy:
> I've added the `on-http-response()` context to the grammar to support having plugins, without sacrificing the unknown keyword error handling.
  I mean, if you have the `LL_IDENTIFIER` on the same level as the options, and the user mistypes an option name (e.g. `user_name` instead of `user` or `user_agent`), then from this change the user faces a "plugin not found: user_name" instead of "unknown identifier"

Let me explain it.
In the future we would like to add plugins as response-handlers (at least that's what I understood from the LB use case).
To add plugins, we usually add a rule in the grammar without a keyword, but with LL_IDENTIFIERand use `plugin_find()`.
But, if we add the LL_IDENTIFIER on "top level" with other http options:
- a not existing option will matches the LL_IDENTIFIER rule -> plugin not found error. Here an unknown keyword would be better.
- I'm not an expert on this, but an LL_IDENTIFIER on top level rule would override other LL_IDENTIFIER rules (e.g. the plugin disk-queue which is on `dest_driver_option` level).

If we use a context, we can solve this, and it's nicer if multiple actions need to be overridden (see my [comment above](https://github.com/syslog-ng/syslog-ng/pull/3007#issuecomment-557099061))

bazsi:
> > I've added the `on-http-response()` context to the grammar to support having plugins, without sacrificing the unknown keyword error handling.
> > I mean, if you have the `LL_IDENTIFIER` on the same level as the options, and the user mistypes an option name (e.g. `user_name` instead of `user` or `user_agent`), then from this change the user faces a "plugin not found: user_name" instead of "unknown identifier"
> 
> Let me explain it.
> In the future we would like to add plugins as response-handlers (at least that's what I understood from the LB use case).
> To add plugins, we usually add a rule in the grammar without a keyword, but with LL_IDENTIFIERand use `plugin_find()`.
> But, if we add the LL_IDENTIFIER on "top level" with other http options:

I don't think plugins should be allowed on a per-response code basis. I would add plugins at a higher conceptual layer, e.g. to implement some kind of API specific functionality. This might include adding/processing HTTP headers or to respond to specific response codes. In those cases the plugin would be invoked at the INNER_DEST insertion point (e.g. just like the disk buffer) and not here.

At least that's my current understanding. I would love to throw out the http-auth code in lib/, that should be going into a separate top-of-http plugin.

> 
>     * a not existing option will matches the LL_IDENTIFIER rule -> plugin not found error. Here an unknown keyword would be better.
> 
>     * I'm not an expert on this, but an LL_IDENTIFIER on top level rule would override other LL_IDENTIFIER rules (e.g. the plugin disk-queue which is on `dest_driver_option` level).
> 

I am not sure I follow here, but maybe the explanation above helps you convey my thoughts. 
> If we use a context, we can solve this, and it's nicer if multiple actions need to be overridden (see my [comment above](https://github.com/syslog-ng/syslog-ng/pull/3007#issuecomment-557099061))

bazsi:
> > Hmm, I think all destinations that use the HTTP transport should be wrapped as an SCL and provided to the user as a specific driver, in which case overrides will be nicely encapsulated and not make it into user's configuration. The arrow syntax is one that we have discussed at the kafka destination which as a pattern would be extended to python() and java() option blocks, in which case it will become more often.
> > I think the "map" concept is becoming increasingly widespread in syslog-ng, so it deserves its own syntax.
> 
> You mean, allowing lists `drop-on-http-response(400, 404)` is not needed, users shouldn't use `on-http-response()`, it's intended for SCLs?

the users are free to use these options, I just wanted to press that we might not need to over-engineer the response code customization interface, as that will - hopefully - be hidden behind a nice/easy-to-use SCL.

Specifying multiple options in a single rule could still be useful though, I like @furiel's idea on that. (e.g. (404, 405) => retry) But we might wait for an actual use-case to emerge with that.

furiel:
Thank you all for your comments. Let me wrap up and react to the comments, and share what I concluded:

> However I like this style (418 => drop), but for me this seems to be very unusal in syslog-ng conf.

I agree. This is a serious problem in syslog-ng. There are already a diversity of ways we are expressing mapping in syslog-ng. We have `key => value`, we have `key=value`, we have `key value` (for example in python destination).

We are increasingly in need of config manipulation: testdb, light, property based testing, config option database. The more diverse the configuration, the more work we need to do just to keep everything together. Let alone user experience. We need to avoid strangeness as much as possible.

> We can write: `(404, 200) => disconnect`

I dropped this idea. The config language is already complex. The starting bracket breaks config option database, and we need to teach testdb+light to handle such exception.

> Maybe even using the "arrow" syntax that we introduced for name-value pairs would be useful here.

I got positive feedbacks from many of you that you like the arrow syntax. Also, it is already added to syslog-ng, so I would stay with that.

> `drop(400, 401)` or `drop-on-http-response(418, 419, 420...)`

Although these versions would be the best to support bulk change, the ordering looks very strange. Conceptually, we attach action to the event, so the mapping is logical to be expressed as code first then action style. (f(x) : domain -> range in terms of mathematics)

> all destinations that use the HTTP transport should be wrapped as an SCL and provided to the user as a specific driver, in which case overrides will be nicely encapsulated and not make it into user's configuration.

This sounds nice, but the problem still persist in scls. Since scls are almost as simple as c-style macros, and we do not have tools for proper metaprogramming: the awkwardness of the wrapped driver option will propagate to the scl.

> The arrow syntax is one that we have discussed at the kafka destination which as a pattern would be extended to python() and java() option blocks, in which case it will become more often.
> I think the "map" concept is becoming increasingly widespread in syslog-ng, so it deserves its own syntax.

I agree. Again, the problem is that our configuration language is very inconsistent. We need to standardize things. For example introduce a `mapping` lexer rule, and use that everywhere consistently.

> I don't think plugins should be allowed on a per-response code basis. I would add plugins at a higher conceptual layer, e.g. to implement some kind of API specific functionality. This might include adding/processing HTTP headers or to respond to specific response codes. In those cases the plugin would be invoked at the INNER_DEST insertion point (e.g. just like the disk buffer) and not here.

I believe we are too far behind to see how such things can be actually implemented. At this point, I would like to leave the "door" open to introduce plugins. And maybe we do not need them. I would like to avoid restricting us now, unless really necessary.

> I just wanted to press that we might not need to over-engineer the response code customization interface, as that will - hopefully - be hidden behind a nice/easy-to-use SCL.

I would really love to see that. Just too many things are holding us back now from to have proper metaprogramming. Personally, I tend to thing it is impossible to fix this using bison.

> Specifying multiple options in a single rule could still be useful though, I like @furiel's idea on that. (e.g. (404, 405) => retry) But we might wait for an actual use-case to emerge with that.

For now, let's go with the simple arrow syntax that is already visible in the branch, without bulk change support. I understand, if users ever need that, it can kind of ugly, but manageable. On the other hand, the problem more complex just to create a local fix here. If a usecase comes, and becomes urgent to address bulk change, we can decide what to do. One quick fix is to introduce `(404, 405) => retry`, just probably without parentheses. But I really hope we get the default values better this time, so users do not really want to deviate too much.

modules/http/http-grammar.ym
furiel: You are right.This came from the "era" when I intended to add many filter parameters besides the http status code, likeheader/body matching. As we concluded earlier, we would support these within the actions. So the `on-` part does not make sense anymore.

modules/http/response-handler.c
furiel: This one is a little bit tricky. When someone inserts a different value behind the same key, by default, hashtable does not replace the key. In the hashtable, the original key is maintained. As the value is overwritten, the old value is deallocated, and together with it, the embedded status code too. The next lookup will result in data loss or invalid read.

It is easy to run into, that's why I added a unit test for overwrite.

I added notes around the code for clarification and future readers.

furiel:
@kira-syslogng retest this please test branch=changing-http-response-action;

modules/http/http-worker.c
furiel: fixed

modules/http/http-worker.c
furiel: fixed

modules/http/http-worker.c
furiel: fixed

modules/http/http-worker.c
furiel: fixed

modules/http/http-grammar.ym
furiel: Fixed.

kira-syslogng:
Build SUCCESS

modules/http/response-handler.c
furiel: After thinking a little bit more about this, I might have an idea. There is a `g_hash_table_replace` that could be used, instead of `g_hash_table_insert`. According to documentation

> Inserts a new key and value into a GHashTable similar to g_hash_table_insert(). The difference is that if the key already exists in the GHashTable, it gets replaced by the new key.

And this is exactly the problem. If I use insert, the old key remains, that would be deallocated together with the value. Replace might work.

PR #3013, opened 2019-11-13
astyle: add padding around spaces

gaborznagy:
astyle's `--pad-comma` option is introduced.

http://astyle.sourceforge.net/astyle.html
> --pad-comma / -xg
Insert space padding after commas. This is not needed if pad-oper is used. Any end of line comments will remain in the original column, if possible. Note that there is no option to unpad. Once padded, they stay padded.
```
if (isFoo(a,b))
    bar(a,b);
becomes:

if (isFoo(a, b))
    bar(a, b);
```
I highly recommend using git blame's ignore whitespace option from now(`git blame -w`). :)

Note: We've just introduced this in the source code of syslog-ng PE features.


kira-syslogng:
Build SUCCESS

PR #3014, opened 2019-11-13
python-source: fix memory corruption during reload

furiel:
According to documentation:
https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_dealloc
If the type is subtypeable, then tp_free must be called. 

The invalid deallocator causes memory corruption, which eventually
leads to crash.

Reproduce: busy reload loop on this configuration:
```
@version: 3.24

log {
  source { python-fetcher(class(PythonFetcher)); };
};

python {
from syslogng import LogFetcher
class PythonFetcher(LogFetcher):
    def fetch(self):
        return LogFetcher.FETCH_NO_DATA, None
};
```

### DIFF
--- a/modules/python/python-fetcher.c
+++ b/modules/python/python-fetcher.c
@@ -551,7 +551,7 @@ static PyTypeObject py_log_fetcher_type =
   PyVarObject_HEAD_INIT(&PyType_Type, 0)
   .tp_name = "LogFetcher",
   .tp_basicsize = sizeof(PyLogFetcher),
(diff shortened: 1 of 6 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
Fixed the other occurrences too. From now there is no `PyObject_Del` called.

kira-syslogng:
Build SUCCESS

PR #3015, opened 2019-11-14
Requires with error message

bazsi:
This is a continuation of #2979 to improve error messages if an installed SCL is missing a dependency.

With these changes we are going to have this as an error message if we are using elasticsearch-http() without the http() module present:

```
[2019-11-14T21:18:09.766325] Cannot load required module; module='http', details='The http() driver is required for elasticsearch-http(), please install syslog-ng-mod-http (Debian) or syslog-ng-http (RHEL) package', location='etc/syslog-ng-elastic.conf:7:16'
Error parsing log statement, syntax error, unexpected LL_ERROR, expecting '}' in block destination elasticsearch-http() at /install/share/syslog-ng/include/scl/elasticsearch/elastic-http.conf:24:4:1-5:1:
1       
2       #Start Block block destination elasticsearch-http() at /install/share/syslog-ng/include/scl/elasticsearch/elastic-http.conf:24
3       
4-----> @requires http "The http() driver is required for elasticsearch-http(), please install syslog-ng-mod-http (Debian) or syslog-ng-http (RHEL) package"
4-----> ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
5----->   http(
6         url("http://localhost:1234/")
7         headers("Content-Type: application/x-ndjson")
8         workers(4)
9         batch_lines(100)

Included from etc/syslog-ng-elastic.conf:7:16-7:97:
2       
3       @include "scl/elasticsearch/elastic-http.conf"
4       
5       log {
6       	source { tcp(port(2000)); };
7-----> 	destination { elasticsearch-http(index("syslog") type("syslog") url("http://localhost:1234/")); };
7-----> 	              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
8       };


```

### DIFF
--- a/scl/slack/slack.conf
+++ b/scl/slack/slack.conf
@@ -20,7 +20,6 @@
 #
 #############################################################################
 
-@requires http
 @requires json-plugin
 @requires basicfuncs
 
@@ -45,6 +44,9 @@ block destination slack(
   use-system-cert-store(yes)
   ...)
 {
+
+@requires http "The slack() driver depends on the syslog-ng http module, please install the syslog-ng-mod-http (Debian & derivatives) or the syslog-ng-http (RHEL & co) package"
+
   http(
     url('`hook-url`')
     method('POST')
--- a/lib/pragma-grammar.ym
+++ b/lib/pragma-grammar.ym
@@ -104,7 +104,7 @@ define_stmt
 								  cfg_args_set(configuration->globals, $2, $3); free($2); free($3); }
 
 module_stmt
-	: KW_MODULE string { last_module_args = cfg_args_new(); } module_param module_params
+	: KW_MODULE string { last_module_args = cfg_args_new(); } module_params
           {
             gboolean success = cfg_load_module_with_args(configuration, $2, last_module_args);
             free($2);
@@ -115,11 +115,15 @@ module_stmt
             if (!success)
               PRAGMA_ERROR();
           }
-	| KW_MODULE string
-          {
-            cfg_load_module(configuration, $2);
-            free($2);
-          }
+	;
+
+module_params
+	: module_param module_params
+	|
+	;
+
+module_param
+	: LL_IDENTIFIER '(' string_or_number ')'		{ cfg_args_set(last_module_args, $1, $3); free($1); free($3); }
 	;
 
 requires_stmt
@@ -147,14 +151,6 @@ requires_stmt
           }
 	;
 
-module_params
-	: module_param module_params
-	|
-	;
-
-module_param
-	: LL_IDENTIFIER '(' string_or_number ')'		{ cfg_args_set(last_module_args, $1, $3); free($1); free($3); }
-	;
 
 /* INCLUDE_RULES */
 
--- a/lib/pragma-grammar.ym
+++ b/lib/pragma-grammar.ym
@@ -95,19 +95,16 @@ version_stmt
                 PRAGMA_ERROR();
           }
 
-include_stmt
-        : KW_INCLUDE string LL_EOL
-          {
-            CHECK_ERROR(cfg_lexer_include_file(lexer, $2), @2, "Error including %s", $2);
-            free($2);
-          }
-        ;
-
 define_stmt
-	: KW_DEFINE LL_IDENTIFIER string_or_number		{ msg_debug("Global value changed",
-                                                                            evt_tag_str("define", $2),
-                                                                            evt_tag_str("value", $3));
-								  cfg_args_set(configuration->globals, $2, $3); free($2); free($3); }
+	: KW_DEFINE LL_IDENTIFIER string_or_number
+          {
+            msg_debug("Global value changed",
+                      evt_tag_str("define", $2),
+                      evt_tag_str("value", $3));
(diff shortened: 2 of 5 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

scl/slack/slack.conf
```diff
@@ -21,8 +21,6 @@
 #############################################################################
 
 @requires http
-@requires json-plugin
-@requires basicfuncs
```
Kokan: Please do not remove this; either move them into the block like http or leave them here.

scl/slack/slack.conf
```diff
@@ -45,6 +43,9 @@ block destination slack(
   use-system-cert-store(yes)
   ...)
 {
+
+@requires http "The slack() driver depends on the syslog-ng http module, please install the syslog-ng-mod-http (Debian & derivatives) or the syslog-ng-http (RHEL & co) package"
```
Kokan: The `@requires http` left on top of the file that defeats a purpose of this, please remove that.

scl/elasticsearch/plugin.conf
```diff
@@ -20,7 +20,6 @@
 #
 #############################################################################
 
-@requires json-plugin
```
Kokan: Please do not remove this, either keep it here or push down into the block.

scl/scl.conf
```diff
@@ -25,6 +25,16 @@
 # `include-path`, then includes all SCL supplied plugins.
 #
 
-@include 'scl/*/*.conf'
-
 @define java-module-dir "`module-install-dir`/java-modules"
+
+@requires json-plugin \
```
Kokan: The `@require` initially have been done due to json plugin not available all the time. (https://github.com/syslog-ng/syslog-ng/issues/673).

I would first of all split this part, as the requires message work is fine. This part where you make json mandatory is kinda a different issue. (If we want to do that I would also make it not optional to compile json module, as in this way it is really easy to make it fail.)

bazsi:
Re: hardwiring json-plugin dependency

I think it is entirely different on what pure syslog-ng depends on and what the SCL supplied modules depend on. 

JSON is pretty much a hard dependency and has been included in the "core" packages on both Debian and RedHat packaging. With a closer check I can see that it is made part of the default package everywhere.

Debian: https://github.com/syslog-ng/syslog-ng/blob/5e4d0f2f759e80d0ad1f6c4f48a294185d5a9994/packaging/debian/syslog-ng-core.install#L31
RedHat: https://github.com/syslog-ng/syslog-ng/blob/5e4d0f2f759e80d0ad1f6c4f48a294185d5a9994/packaging/rhel/syslog-ng.spec#L472
FreeBSD: https://svnweb.freebsd.org/ports/head/sysutils/syslog-ng324/Makefile?view=markup&pathrev=519926#l33
Arch: https://www.archlinux.org/packages/extra/x86_64/syslog-ng/

I think it is ok that we enforce JSON dependency for SCL. If someone does not want JSON he can choose not to use the SCL either. And the fragmentation of SCL that is present in some of the packagings is exactly the reason why SCL is far from an ideal state and that cannot be depended upon. Thus my aim to simplify things and make sure the "default behavior" is really a default that can be depended on.

With regards to build options, json-c has been integrated as a submodule exactly because our dependency on it is so large (and has been increasing steadily). On my part it would be ok to make it a hard dependency.

scl/slack/slack.conf
bazsi: done.

scl/slack/slack.conf
bazsi: I have separated this to the last patch, which can be dropped depending on the discussion we have on making json-plugin mandatory. But I'm resolving this conversation.

scl/elasticsearch/plugin.conf
bazsi: again, this one is pending on the discussion on making json-plugin mandatory. resolving this.

scl/scl.conf
bazsi: I have now moved all of this into a separate patch that can be split off to a separate branch if needed, I just didn't want to add the detailed comment on the missing module into n+1 locations.

like I said in the other thread, at this point I would opt for making json-plugin mandatory for the SCL. IF someone doesn't need the functionality of the SCL they can forgo it simply by removing the "@include scl.conf" in their configuration.

kira-syslogng:
Build SUCCESS

bazsi:
I've just dropped the last patch of the branch as we agreed IRL. I am opening a separate PR on making JSON mandatory.

kira-syslogng:
Build SUCCESS

PR #3017, opened 2019-11-15
configure: add Python 3.8 detection

MrAnno:
On Unix, C extensions are no longer linked to libpython.

They've added a pkg-config `python-3.8-embed` module to embed Python into an application: `pkg-config python-3.8-embed --libs` includes `-lpython3.8`.

> To support both 3.8 and older, try pkg-config python-X.Y-embed --libs first and fallback to pkg-config python-X.Y --libs (without --embed) if the previous command fails.


kira-syslogng:
Build SUCCESS

PR #3018, opened 2019-11-15
create_dirs: some cleanup

### LINKED ISSUE #2635: create-dirs() option fails in source pipe()
# syslog-ng

## Version of syslog-ng

```
syslog-ng 3.8.1
Installer-Version: 3.8.1
Revision: 3.8.1-10
Module-Directory: /usr/lib/syslog-ng/3.8
Module-Path: /usr/lib/syslog-ng/3.8
Available-Modules: mod-python,pseudofile,cryptofuncs,afsql,affile,afsocket,syslogformat,geoip-plugin,linux-kmsg-format,date,json-plugin,system-source,sdjournal,confgen,cef,redis,afprog,afsmtp,kvformat,graphite,basicfuncs,csvparser,add-contextual-data,afstomp,disk-buffer,dbparser,afuser,riemann,afamqp,afmongodb
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: off
```

## Platform

Debian 9.7

# Issue

Specifying create-dirs(yes) in source pipe() is not recognized / causes a parsing failure.  Note, this is specified in a separate conf.d/local.conf file that gets loaded later.

## Failure

```
syslog-ng[46941]: Error parsing affile, syntax error, unexpected KW_CREATE_DIRS, expecting KW_NORMALIZE_HOSTNAMES or KW_USE_DNS or KW_USE_FQDN or KW_DNS_CACHE in /etc/syslog-ng/conf.d/local.conf at line 6, column 9:
syslog-ng[46941]: included from /etc/syslog-ng/syslog-ng.conf line 162, column 1
syslog-ng[46941]:         create-dirs(yes)
syslog-ng[46941]:         ^^^^^^^^^^^
```

## Steps to reproduce

Create a conf.d/local.conf that includes the following:

```
source s_pipe {
    pipe(
        "/var/local/pipe"
        flags("no-parse")
        time-zone("UTC")
        create-dirs(yes)
        dir-owner(root)
        dir-group(daemon)
        dir-perm(0755)
        owner(root)
        group(daemon)
        perm(0664)
   );
};
```
1. start syslog-ng

If you comment out the create-dirs option, it works fine.

gaborznagy:
Fixes #2635 

**Added create_dirs() option to pipe() source**:
Pipe source creates the FIFO where it will read, thus it should be able to create the path as well.

Previously file/pipe sources created the directories (due to a bug), and it was impossible to disable directory creation.

About file source: it does not need the option create-dirs(), as it will poll the path, and it shouldn't create the directories since that is the job of the destination (the one who writes into the file).

Unix-dgram() and unix-stream() dest. drivers does not need create_dirs() option: Even if they could create the directories, they couldn't connect to a not existing socket anyway.

Other changes:
- pipe()/file() source created directories by default, and it was impossible to disable it (global option did not work) -> disabled default directory creation
- added create_dirs() option to pipe() destination (the global option worked)
- add global option inheritance to `FileOpener` users: affile/afpipe.

### Questions to reviewers

- Are there risks to allow directory creation on source side, or in general?
- backward compatibility: is it okay to change an undocumented feature (default directory creation in file/pipe sources)
- backward compatibility: is it a good design to make pipe() source fail on startup if directories dont exist and create_dirs is not enabled?
Currently, unix-dgram() socket works the same way. For pipes, a follow_freq() behaviour could be added, similarly to file() source, periodically checking the existence of the given path, but this can't be done for sockets...

### DIFF
--- a/modules/affile/affile-source.c
+++ b/modules/affile/affile-source.c
@@ -178,5 +178,6 @@ affile_sd_new(gchar *filename, GlobalConfig *cfg)
 
   self->file_reader_options.restore_state = self->file_reader_options.follow_freq > 0;
   file_opener_options_defaults_dont_change_permissions(&self->file_opener_options);
+  self->file_opener_options.create_dirs = FALSE;
   return &self->super.super;
 }
--- a/modules/affile/file-opener.c
+++ b/modules/affile/file-opener.c
@@ -226,6 +226,8 @@ void
 file_opener_options_init(FileOpenerOptions *options, GlobalConfig *cfg)
 {
   file_perm_options_inherit_from(&options->file_perm_options, &cfg->file_perm_options);
+  if (options->create_dirs == -1)
+    options->create_dirs = cfg->create_dirs;
 }
 
 void
--- a/modules/affile/affile-dest.c
+++ b/modules/affile/affile-dest.c
@@ -492,8 +492,6 @@ affile_dd_init(LogPipe *s)
   if (!log_dest_driver_init_method(s))
     return FALSE;
 
-  if (self->file_opener_options.create_dirs == -1)
-    self->file_opener_options.create_dirs = cfg->create_dirs;
   if (self->time_reap == -1)
     self->time_reap = cfg->time_reap;
 
--- a/modules/affile/affile-grammar.ym
+++ b/modules/affile/affile-grammar.ym
@@ -285,10 +285,9 @@ dest_affile_option
 	| dest_driver_option
         | file_perm_option
 	| KW_OPTIONAL '(' yesno ')'		{ last_driver->optional = $3; }
-	| KW_CREATE_DIRS '(' yesno ')'		{ affile_dd_set_create_dirs(last_driver, $3); }
 	| KW_OVERWRITE_IF_OLDER '(' nonnegative_integer ')'	{ affile_dd_set_overwrite_if_older(last_driver, $3); }
 	| KW_FSYNC '(' yesno ')'		{ affile_dd_set_fsync(last_driver, $3); }
-	| KW_TIME_REAP '(' nonnegative_integer ')'		{ affile_dd_set_time_reap(last_driver, $3); }
+        | dest_affile_common_option
 	;
 
 dest_afpipe_params
@@ -310,8 +309,13 @@ dest_afpipe_option
 	: dest_writer_option
 	| dest_driver_option
 	| file_perm_option
-	| KW_TIME_REAP '(' nonnegative_integer ')'		{ affile_dd_set_time_reap(last_driver, $3); }
+        | dest_affile_common_option
 	;
+
+dest_affile_common_option
+	: KW_TIME_REAP '(' nonnegative_integer ')'		{ affile_dd_set_time_reap(last_driver, $3); }
+	| KW_CREATE_DIRS '(' yesno ')'		{ affile_dd_set_create_dirs(last_driver, $3); }
+        ;
 
 
 /* INCLUDE_RULES */
--- a/modules/affile/affile-grammar.ym
+++ b/modules/affile/affile-grammar.ym
@@ -229,6 +229,7 @@ source_afpipe_options
 source_afpipe_option
 	: KW_OPTIONAL '(' yesno ')'			{ last_driver->optional = $3; }
 	| KW_PAD_SIZE '(' nonnegative_integer ')'	{ last_log_proto_options->pad_size = $3; }
+	| KW_CREATE_DIRS '(' yesno ')'			{ pipe_sd_set_create_dirs(last_driver, $3); }
 	| multi_line_option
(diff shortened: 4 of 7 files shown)


kira-syslogng:
Build SUCCESS

modules/affile/file-opener.c
```diff
@@ -106,7 +106,7 @@ _obtain_capabilities(FileOpener *self, const gchar *name, cap_t *act_caps)
       g_process_enable_cap("cap_dac_override");
     }
 
-  if (self->options->create_dirs &&
+  if (self->options->create_dirs == TRUE &&
```
MrAnno: `create_dirs` should be checked only after calling `file_opener_options_init()`, which is responsible for getting rid of all the temporary `-1`s. I wouldn't check explicitly for `TRUE` here.

modules/affile/affile-source.c
```diff
@@ -178,5 +178,6 @@ affile_sd_new(gchar *filename, GlobalConfig *cfg)
 
   self->file_reader_options.restore_state = self->file_reader_options.follow_freq > 0;
   file_opener_options_defaults_dont_change_permissions(&self->file_opener_options);
+  self->file_opener_options.create_dirs = FALSE;
```
MrAnno: > We need to explicitly disable directory creation, as a global create_dirs() option with a value "yes" can enable it and we don't want that.

As a default, I agree; we should disable the inheritance. Creating the directory structure (not the file itself) for the monitored file is not necessarily a bad idea, if someone explicitly sets the option in the driver.

modules/affile/file-opener.c
gaborznagy: If we don't have explicit checking for `TRUE` value, then we don't have any safety-mechanism.
> After initializing *Options objects, all of these values should be replaced with the inherited value.

Yes, but this was missing and nothing protected us from it.
I know this is a general issue.

Now, `file_opener_options_init` is initializing `create_dirs`, but I would think it more safe if there would be checking/assertion where `create_dirs` is used. Do you think a type-check(TRUE or FALSE, but nothing else) assertion in `_obtain_capabilities` would be a  good idea?
After writing it down, I don't think. :)
A function should not check for type-safety ....

I'll just remove the explicit check then. What do you think? A simple ":+1" is a good sign.

gaborznagy: I've dropped the patch.

kira-syslogng:
Build SUCCESS

modules/affile/file-opener.c
MrAnno: >  Do you think a type-check(TRUE or FALSE, but nothing else) assertion in _obtain_capabilities would be a good idea?

It doesn't add anything useful in my opinion, unless you do it for all of our options (or some kind of abstraction that validates all options against `-1` automatically in `init()`).

modules/affile/affile-source.c
MrAnno: I noticed we've never had such an option for the file source. We can add it later, if someone needs it.

MrAnno:
> Are there risks to allow directory creation on source side, or in general?

If the source is responsible for creating the "transport medium" that is represented as a file on the disk (pipe, unix socket, etc.), `create-dir()` makes sense and should be added as a possible option.

> backward compatibility: is it okay to change an undocumented feature (default directory creation in file/pipe sources)

I think your PR is right behaviour, even if it breaks something undocumented. If someone needs `create-dirs()` for file sources, we can add it as a new (documented!) feature that does not inherit the global one. A source-side `create-dir()` for `file()` sounds rare and less useful.
Either way, syslog-ng should start and follow the file as before.

> backward compatibility: is it a good design to make pipe() source fail on startup if directories dont exist and create_dirs is not enabled? Currently, unix-dgram() socket works the same way. For pipes, a follow_freq() behaviour could be added, but not for sockets...

Yes. the `pipe()` source creates the pipe, so it does not make sense to poll if the user created the directory hierarchy and then create a pipe in it.
`create-dirs(yes)` should create the hierarchy and the named pipe. IMHO, syslog-ng should not start when the pipe can't be created (because of a non-existent directory and `create-dirs(no)`).

MrAnno:
Quick summary:

- `create-dirs()` is a global, per-source, and per-destination option
- `create-dirs()` makes sense in places where at least one file is be created by syslog-ng:
  - file/pipe destinations
  - unix domain socket sources
  - pipe sources

This PR
  - fixed `create-dirs()` in the `pipe()` destination (only worked using the global option);
  - fixed `create-dirs()` in the `pipe()` source (could have worked only using the global option, but due to a bug, it was always enabled);
  - disabled `create-dirs()` for file sources, because regular file sources do not create resources on disk, they poll non-existent files/directories instead.

PR #3019, opened 2019-11-18
Loggen plugin path

norberttak:
Loggen plugin path was set by compile time. This was a problem when the user wants to install loggen into other directory then the default path.
This PR solves this issue by the reloc API to determine the syslog-ng (loggen) install path by run-time

In case of none standard installation methode, the user can set the SYSLOGNG_PREFIX envvar.

### DIFF
--- a/tests/loggen/loggen.c
+++ b/tests/loggen/loggen.c
@@ -27,6 +27,7 @@
 #include "loggen_helper.h"
 #include "file_reader.h"
 #include "logline_generator.h"
+#include "reloc.h"
 
 #include <stdio.h>
 #include <sys/time.h>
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

tests/loggen/loggen.c
```diff
@@ -463,8 +469,18 @@ main(int argc, char *argv[])
   signal(SIGPIPE, signal_callback_handler);
   setup_rate_change_signals();
 
-  int plugin_num = enumerate_plugins(SYSLOG_NG_PATH_LOGGEN_PLUGIN_DIR, plugin_array, ctx);
-  DEBUG("%d plugin successfuly loaded\n", plugin_num);
+  /* if plugin path defined by environment variable then we should use it */
+  const char *plugin_path = g_getenv(ENV_SYSLOGNG_LOGGEN_PLUGINDIR);
```
Kokan: I think the `SYSLOGNG_PREFIX` serves the same purpose. Using only that should be sufficient, even if that requires the additional `/lib/syslog-ng/loggen` suffix. (keep this simple)

norberttak:
Removed the commit of ENV_SYSLOGNG_LOGGEN_PLUGINDIR. PR description updated.

kira-syslogng:
Build SUCCESS

PR #3020, opened 2019-11-19
http destination ignores global template options values

Kokan:
The template optinos default value should be the value of global template options.
In case of http destination inheriting the global options was not possible,
as the mechanism relies on:
* First calling the *log_template_options_defaults* setting external values
  indicating nobody configured those fields.
* Processing the grammar rules (potentially values are overwritten)
* Initializing the template values with *log_template_options_init*
  (either using the global value, if the initial external value left, or keeping
  the configured value)

The *log_template_options_defaults* was missing, as a result all of the template
options defaulted to zero in case of http destination driver.


kira-syslogng:
Build SUCCESS

PR #3021, opened 2019-11-19
Light: SyslogNgConfig refactor part 1

mitzkia:
This PR is a first part of a bigger SyslogNgConfig refactor

- this part contains only functionally equivalent changes

### DIFF
--- a/tests/python_functional/src/syslog_ng_config/statements/destinations/destination_reader.py
+++ b/tests/python_functional/src/syslog_ng_config/statements/destinations/destination_reader.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #############################################################################
-# Copyright (c) 2015-2018 Balabit
+# Copyright (c) 2015-2019 Balabit
 #
 # This program is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 as published
@@ -20,15 +20,28 @@
 # COPYING for details.
 #
 #############################################################################
+import logging
+
+from src.message_reader.message_reader import MessageReader
+from src.message_reader.single_line_parser import SingleLineParser
+
+logger = logging.getLogger(__name__)
 
 
-class DestinationDriver(object):
-    group_type = "destination"
+class DestinationReader(object):
+    def __init__(self, IOClass):
+        self.__IOClass = IOClass
+        self.__reader = None
 
-    def __init__(self, positional_parameters=None, options=None):
-        if positional_parameters is None:
-            positional_parameters = []
-        self.positional_parameters = positional_parameters
-        if options is None:
-            options = {}
-        self.options = options
+    def read_logs(self, path, counter):
+        if not self.__reader:
+            io = self.__IOClass(path)
+            io.wait_for_creation()
+            message_reader = MessageReader(
+                io.read, SingleLineParser(),
+            )
+            self.__reader = message_reader
+        messages = self.__reader.pop_messages(counter)
+        read_description = "Content has been read from\nresource: {}\ncontent: {}\n".format(path, messages)
+        logger.info(read_description)
+        return messages
--- a/tests/python_functional/src/syslog_ng_config/statements/destinations/file_destination.py
+++ b/tests/python_functional/src/syslog_ng_config/statements/destinations/file_destination.py
@@ -25,19 +25,21 @@ from pathlib2 import Path
 import src.testcase_parameters.testcase_parameters as tc_parameters
 from src.driver_io.file.file_io import FileIO
 from src.syslog_ng_config.statements.destinations.destination_driver import DestinationDriver
+from src.syslog_ng_config.statements.destinations.destination_reader import DestinationReader
 
 
 class FileDestination(DestinationDriver):
     def __init__(self, file_name, **options):
         self.driver_name = "file"
         self.path = Path(tc_parameters.WORKING_DIR, file_name)
-        super(FileDestination, self).__init__(FileIO, [self.path], options)
+        super(FileDestination, self).__init__([self.path], options)
+        self.destination_reader = DestinationReader(FileIO)
 
     def get_path(self):
         return self.path
 
     def read_log(self):
-        return self.dd_read_logs(self.get_path(), counter=1)[0]
+        return self.destination_reader.read_logs(self.get_path(), counter=1)[0]
 
     def read_logs(self, counter):
-        return self.dd_read_logs(self.get_path(), counter=counter)
+        return self.destination_reader.read_logs(self.get_path(), counter=counter)
--- a/tests/python_functional/src/syslog_ng_config/statements/sources/source_writer.py
+++ b/tests/python_functional/src/syslog_ng_config/statements/sources/source_writer.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #############################################################################
-# Copyright (c) 2015-2018 Balabit
+# Copyright (c) 2015-2019 Balabit
 #
 # This program is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 as published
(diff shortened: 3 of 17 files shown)


kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng_config/statements/destinations/destination_reader.py
```diff
@@ -0,0 +1,47 @@
...
+logger = logging.getLogger(__name__)
+
+
+class DestinationReader(object):
+    def __init__(self, IOClass):
+        self.__IOClass = IOClass
+        self.__reader = None
+
+    def dd_read_logs(self, path, counter):
+        if not self.__reader:
```
MrAnno: The content of this `if` branch should go to the constructor (except the `wait_for_creation` call, that belongs here).

I don't think we need lazy initialization here.

tests/python_functional/src/syslog_ng_config/statements/destinations/file_destination.py
```diff
@@ -25,19 +25,21 @@
...
+from src.syslog_ng_config.statements.destinations.destination_reader import DestinationReader
 
 
 class FileDestination(DestinationDriver):
     def __init__(self, file_name, **options):
         self.driver_name = "file"
         self.path = Path(tc_parameters.WORKING_DIR, file_name)
-        super(FileDestination, self).__init__(FileIO, [self.path], options)
+        super(FileDestination, self).__init__([self.path], options)
+        self.destination_reader = DestinationReader(FileIO)
```
MrAnno: That looks nice.

tests/python_functional/src/syslog_ng_config/statements/destinations/destination_reader.py
```diff
@@ -0,0 +1,47 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class DestinationReader(object):
+    def __init__(self, IOClass):
+        self.__IOClass = IOClass
+        self.__reader = None
+
+    def dd_read_logs(self, path, counter):
```
MrAnno: Can we remove `dd` from the name? It's a destination reader now.

tests/python_functional/src/syslog_ng_config/statements/sources/source_writer.py
```diff
@@ -0,0 +1,41 @@
...
+class SourceWriter(object):
+    def __init__(self, IOClass):
+        self.__IOClass = IOClass
+        self.__writer = None
+
+    def __construct_writer(self, path):
+        if not self.__writer:
+            self.__writer = self.__IOClass(path)
+
+    def sd_write_log(self, path, formatted_log, counter):
```
MrAnno: I'd also remove `sd` from the method name.

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
```diff
@@ -114,12 +114,8 @@ def create_inner_logpath(self, statements=None, flags=None):
         inner_logpath = self.__create_logpath_with_conversion(statements, flags)
         return inner_logpath
 
-    def create_example_msg_generator(self, **options):
-        generator_source = SourceDriver(None)
-        generator_source.driver_name = "example_msg_generator"
-        generator_source.DEFAULT_MESSAGE = "-- Generated message. --"
-        generator_source.options = options
-        return generator_source
+    def create_example_msg_generator_source(self, **options):
```
MrAnno: I still don't like these methods, it's just boilerplate, will be hard to maintain. Calling the constructor directly from tests is much better in my opinion. We'd see actual types this way and we could import them in tests.

tests/python_functional/src/syslog_ng_config/statements/destinations/file_destination.py
MrAnno: The `DestinationDriver` class just became almost empty after extracting the reader from it. Do you have plans to add new functionality to `DestinationDriver`?

kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng_config/statements/destinations/file_destination.py
mitzkia: Yes, you are right. In my next PR I have a plan to extend SourceDriver() and DestinationDriver() classes with their new scopes. 

I really do not know if that change can fit here (it could extend this PR a lot).
"Unfortunately" this classes FileSource / FileDestination / SourceDriver / DestinationDriver / SourceWriter / DestinationReader are highly depends on each other.

One other note, I closed my previous PR because of that "huge" change.

I will think over it again. What is your suggestion after this?

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
mitzkia: Hm, interesting. If I understand you correctly this would be your suggestion?

```python
from src.syslog_ng_config.statements.sources.file_source import FileSource
from src.syslog_ng_config.statements.destinations.file_destination import FileDestination

def test_acceptance(config, syslog_ng, input_log, expected_log, counter):
    file_source = FileSource(file_name="input.log")
    file_destination = FileDestination(file_name="output.log")
    config.create_logpath(statements=[file_source, file_destination])
    config.update_global_options(keep_hostname="yes")

    file_source.write_log(input_log, counter)
    syslog_ng.start(config)
    assert file_destination.read_logs(counter) == [expected_log] * counter
```

It could be work in this way, _but_ I have a plan to add some property-based tests which are highly depends on this form:
```python
def create_some_driver_name_source():
    pass

def create_some_driver_name_destination():
    pass
```

The plan was the following:
- SyslogNgConfig() class could be generate automatically driver objects (separately sources and destinations) by their names. (What else would be the scope of the SyslogNgConfig() class besides to generate driver objects?). 


If we have a plan to remove `create_...()` like functions we should replace it with a separate list of supported drivers. I mean the following:
```
class SyslogNgConfig():
     def __init__(self):
          ....
          self.supported_source_drivers = [FileSource, ExampleMessageGenerator]
          self.supported_destination_drivers = [FileDestination]
```
with this change I think the property-based tests could generate driver objects automatically from SyslogNgConfig() class.

After this what do you think?

tests/python_functional/src/syslog_ng_config/statements/destinations/file_destination.py
MrAnno: Nothing, I was just curious. This class seems to have its place, I'm sure it'll have the right functionality later :)

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
MrAnno: There are ways to get source/destination instances without writing boilerplate or maintaining lists, for example, registering your source/destination types into a registry (so you can get instances similar to the Service Locator pattern. This is an anti-pattern because it also hides dependencies, but it's slightly better than a manual list).

But OK, it is not the scope of this PR.

tests/python_functional/src/syslog_ng_config/statements/destinations/destination_reader.py
mitzkia: You are absolutely right, this part should belongs to the constructor.
Unfortunately this part of Light is a "little bit" buggy. For now you can not re-initalize `io` variable  inside DestinationReader() with a new path. Fixing this needs more refactor steps in FileDestination(), DestinationDriver(), DestinationReader() classes (of course sources are also affected). 
My next PR will fix this issue (included with syslog-ng functional tests).

MrAnno: OK thanks.

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
mitzkia: Thank you for your suggestion. I will check that patterns.

kira-syslogng:
Build SUCCESS

lbudai:
hmmm.... `SourceWriter` and `DestinationReader`? I think we should not overload `Reader` and `Writer` ; `Reader` is related to Sources while Writer related to Destinations in syslog-ng.
This naming is confusing... 
(some suggestions: LogProducer, LogConsumer).

mitzkia:
@lbudai Thanks for the comment and the suggestion. Earlier I already had this review comment, and the main reason why I use this naming is: this is "testing side", so because of this I think it is not confusing. 
In details: from testing side while we communicate with sources we use "write" method (to test any source driver we should "write" into it), and in destination case we use "read" method (same reason).
But one thing is the naming of the classes, other is the methods which they provides.
If we would like to rename classes we should think about the provided methods also.
SourceWriter for now uses -> write_log()           -> should use produce_log()?
DestinationReader for now uses -> read_log()   -> should use consume_log()?

For me the news names are not "natural".

lbudai:
@mitzkia: if you write this framework for youself, then okay, but in that case this should not be part of syslog-ng ;) 
AFAIK this framework is also for developers...

Unfortunately I didn't (and still don't) have the time for refactor these things, but trying to share some idea with you IRL.

PR #3024, opened 2019-11-20
Fix loggen plugin directory when `--with-module-dir` is used

### LINKED ISSUE #3023: dbld/rules rpm-centos-7 fails due loggen library path changes
# syslog-ng
## Version of syslog-ng
Fresh git snapshot: syslog-ng-3.24.1.176.g39ed128

## Platform
CentOS 7 platform within DBLD

# Issue
## Failure
```dbld/rules rpm-centos-7```
fails with the following error:
```
RPM build errors:
    Directory not found: /root/rpmbuild/BUILDROOT/syslog-ng-3.24.1.176.g39ed128-1+20191120T122658.el7.x86_64/usr/lib64/syslog-ng/loggen
    File not found by glob: /root/rpmbuild/BUILDROOT/syslog-ng-3.24.1.176.g39ed128-1+20191120T122658.el7.x86_64/usr/lib64/syslog-ng/loggen/libloggen*
make: *** [dbld/rules:118: rpm-centos-7] Error 1
```
When removed the related two lines from the spec file, I get:
```
    Installed (but unpackaged) file(s) found:
   /usr/lib/syslog-ng/loggen/libloggen_socket_plugin.so
   /usr/lib/syslog-ng/loggen/libloggen_ssl_plugin.so
```

Before modifying any spec files: is this an expected change?

MrAnno:
The loggen plugin dir should follow `moduledir`, which can be set, for example, with the `--with-module-dir` configure option.

#3019 changed this behavior, to make the run-time relocation of loggen work, but this way the loggen plugin directory does not depend on `moduledir` anymore.

This PR fixes the problem.
In order not to break the loggen relocation, `${moduledir}` has to be added to the relocation API, so paths containing `${moduledir}` can be resolved correctly with `get_installation_path_for()`.

Fixes #3023

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -161,6 +161,10 @@ AC_ARG_WITH(module-dir,
    [  --with-module-dir=path   Use path as the directory to install modules into],
    moduledir=$with_module_dir)
 
+AC_ARG_WITH(loggen-plugin-dir,
+   [  --with-loggen-plugin-dir=path   Use path as the directory to install loggen plugins into],
+   loggenplugindir=$with_loggen_plugin_dir)
+
 AC_ARG_WITH(module-path,
    [  --with-module-path=path   Use path as the list of ':' separated directories looked up when searching for modules],
    module_path=$with_module_path)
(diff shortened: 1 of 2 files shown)


MrAnno:
Thanks for the test!

Sorry, I'm pushing a small modification to make the fix a bit cleaner.

kira-syslogng:
Build SUCCESS

MrAnno:
Done.

kira-syslogng:
Build SUCCESS

PR #3025, opened 2019-11-22
[WIP] python-source: warning for possible persist collision in the future

furiel:
Built on top of https://github.com/syslog-ng/syslog-ng/pull/3016.

We plan to introduce persist api for python source and python fetcher. Unfortunately, currently there is no persist name allocated automatically, which makes introduction of persist api very hard. For now different python source and fetcher instances can start with the same class. If the `persist_name` enforcing would be introduced instantly, those configurations might break.

This patch adds warning if syslog-ng detects a collision. Later the warning will be turned into error, and syslog-ng will not start.

If persist name collision is detected, users can fix it either by:

1. Specify `persist-name` option in the configuration file.
2. Specify `generate_persist_name` static method in the python class (preferable if suitable)

Something similar:

``` python
class SourceOrFetcher(...):
    @staticmethod
    def generate_persist_name(options):
        return options['port']
```

The hidden collision can happen only amongst python sources and fetchers, and if and only if the same class name is used multiple times.

----

#

### DIFF
--- a/modules/python/CMakeLists.txt
+++ b/modules/python/CMakeLists.txt
@@ -11,6 +11,8 @@ set(PYTHON_SOURCES
     python-module.h
     python-config.h
     python-config.c
+    python-persist.h
+    python-persist.c
     python-helpers.h
     python-helpers.c
     python-main.h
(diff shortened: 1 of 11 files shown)


kira-syslogng:
Build SUCCESS

furiel:
We decided to add a warning to the announcement instead. Closing this issue.

PR #3026, opened 2019-11-23
Fix infinite loop when a TLS connection breaks

### LINKED ISSUE #3009: TLS issue and 100% CPU
Hi there,
I trying to use tls connection to send logs to syslog server. Using simple tcp destination all thing is work, but when I change it to tls (changing respective ports and configuration) the connection works for 10 ~ 30 minutes and then I can see log entry to my syslog-ng  process but I cannot see them going out.
After it's occurs the CPU is 100%, only restart the syslog-ng helps

Thanks

MrAnno:
syslog-ng does not call `SSL_connect()` or `SSL_do_handshake()`, so the first `SSL_read()` or `SSL_write()` call is responsible for negotiating the TLS session. Nevertheless, the initial IO `cond` should be `0`, because the `LogProto` layer should choose the initial IO direction.

In case TLS renegotiation is required, OpenSSL returns an error (`SSL_ERROR_WANT_READ` or `SSL_ERROR_WANT_WRITE`). In this case, the same read/write function must be invoked after a IO readiness notification.

With all that being said, the initial `G_IO_IN | G_IO_OUT` transport-level IO-cond is not necessary and even harmful:

If the connection is lost in a TLS destination before sending the first message, an infinite IO loop starts in the main thread. This is because ivykis reports readiness for reading, but we won't read anything as the `LogProto` instance is not in the correct state.

Detecting the breakdown of the connection was not possible because we would have to read from the socket first, and only then could we notice about the closed connection.

Fixes #3009

### DIFF
--- a/lib/logproto/logproto-client.h
+++ b/lib/logproto/logproto-client.h
@@ -148,6 +148,8 @@ log_proto_client_process_in(LogProtoClient *s)
     /*
      * In some clients, flush is used for input processing, but it should not be.
      * Fix these clients, than remove the flush call here.
+     *
+     * SSL_ERROR_WANT_READ in the TLS transport is also built upon this.
      */
     return s->flush(s);
   else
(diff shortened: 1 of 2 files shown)


lbudai:
nice ;)
thanks for investigating this!

MrAnno:
@lbudai My pleasure. :)

I've finally found the reason why we lose a single message after reconnecting to a TLS server. I'll open another PR next week.

kira-syslogng:
Build SUCCESS

PR #3027, opened 2019-11-24
WIP: Light: SyslogNgConfig refactor part 2

mitzkia:
Reasons behind the commits and the PR:
- After all we have decided to split this PR again (because I already not yet finished one part)
- The main goal behind this PR set to define (flexible) external and internal API for SyslogNgConfig and the related (drivers, helpers) classes.
- Why this PR needed:
 - Until now there were only a few supported drivers: file source, file destination, example message generator, filter, app-parser, syslog-parser
 - To support other drivers (which can be very different) we should prepare the API for those.

In details:
- After this PR what will be the scopes of config related layers (classes):

- ConfigStatement() class
    - in class hierarchy: this class is a root parent of every statement classes
    - the main scope of this class (in the future):
        - this class will manage those things which are generally related to any config statements:
            - provide an OptionHandler() for every child class (not yet added in this PR)
            - render options for config
    - this class does not exist before

- SourceDriver(), DestinationDriver(), Filter(), Parser() classes
    - in class hierarchy: this classes are child classes of ConfigStatement() class
    - the main scope of this classes (in the future):
        - everything inherited from parent (ConfigStatement()) class
        - this classes will implement functionalities which are related only this layer:
            - instantiate reader and writer classes (SourceWriter(), DestinationReader()) in constructor time
            - provide write_log() and read_log() as wrappers
    - changes:
        - SourceDriver() class will construct a SourceWriter() instance (was moved from FileSource()), (DriverIO information will be passed from child class)
        - DestinationDriver() class will construct a DestinationReader() instance (was moved from FileDestination()), (DriverIO information will be passed from child class)
        - SourceDriver() will provide a write_log() wrapper (was moved from FileSource())
        - DestinationDriver() will provide a read_log() wrapper (was moved from FileDestination())

- FileSource(), FileDestination(), ExampleMessageGenerator() classes
    - in class hierarchy: this classes are child classes of SourceDriver() and DestinationDriver() classes
    - the main scope of this classes (in the future):
        - everything inherited from parents
        - this classes will implement functionalities which are related only for each drivers:
            - setting driver default (mandatory) values
            - setting option formatters ("file_path" -> should prefixed with actual working dir)
            - administratively defines the driver name, and passes the corresponding DriverIO object to the parent
    - changes:
        - move read_log() and write_log() related wrappers from this classes to the parent
        - move DestinationReader() and SourceWriter() compositions to the parent
        - pass DriverIO class to the parent class
        - add a general OptionHandler() composition to this class

- SyslogNgConfig() class
    - in class hierarchy: this is a separate class
    - the main scope of this class:
        - provide create_... like methods for drivers, filters, parsers
        - update global options
        - link drivers, filters, parsers into top level (or inner) logpath
    - changes:
        - the `file_name` parameter was explicilty added to the create_file_source() and create_file_destination() methods.
        Internally this parameter was already taken out from the passed options dict. This change will make cleaner how we
        would like to use this parameter internally, and also addes a hint for those who would like to TAB out the internal
        name of the positional option.

- SourceWriter() and DestinationReader() classes:
    - in class hierarchy: this classes are instantiated in SourceDriver() and DestinationReader() classes in constructor time
    - the main scope of this classes:
        - instantiate specific DriverIO classes
        - provide read_log() and write_log() methods
    - changes:
        - init_driver_io() method was added to be abel to instantiate DriverIO class separately

Todo:
- [x] : Administratively to the first comment: Summarize the goal of the commits

### DIFF
--- a/tests/python_functional/src/syslog_ng_config/statements/destinations/destination_driver.py
+++ b/tests/python_functional/src/syslog_ng_config/statements/destinations/destination_driver.py
@@ -27,7 +27,7 @@ from src.syslog_ng_config.statements.destinations.destination_reader import Dest
 class DestinationDriver(ConfigStatement):
     group_type = "destination"
 
-    def __init__(self, positional_parameters=None, options=None, driver_io_cls=None):
+    def __init__(self, positional_parameters=None, options=None, driver_io_cls=None, line_parser_cls=None):
         if positional_parameters is None:
             positional_parameters = []
         self.positional_parameters = positional_parameters
@@ -36,13 +36,14 @@ class DestinationDriver(ConfigStatement):
         self.options = options
 
         self.driver_io_cls = driver_io_cls
+        self.line_parser_cls = line_parser_cls
         self.destination_reader = None
         self.init_destination_reader()
         super(DestinationDriver, self).__init__()
 
     def init_destination_reader(self):
-        if self.driver_io_cls and self.positional_parameters:
-            self.destination_reader = DestinationReader(self.driver_io_cls)
+        if self.driver_io_cls and self.line_parser_cls and self.positional_parameters:
+            self.destination_reader = DestinationReader(self.driver_io_cls, self.line_parser_cls)
             self.destination_reader.init_driver_io(self.positional_parameters[0])
 
     def read_log(self):
--- a/tests/python_functional/src/syslog_ng_config/statements/destinations/destination_driver.py
+++ b/tests/python_functional/src/syslog_ng_config/statements/destinations/destination_driver.py
@@ -20,15 +20,31 @@
 # COPYING for details.
 #
 #############################################################################
+from src.syslog_ng_config.statements.destinations.destination_reader import DestinationReader
 
 
 class DestinationDriver(object):
     group_type = "destination"
 
-    def __init__(self, positional_parameters=None, options=None):
+    def __init__(self, positional_parameters=None, options=None, driver_io_cls=None):
         if positional_parameters is None:
             positional_parameters = []
         self.positional_parameters = positional_parameters
         if options is None:
             options = {}
         self.options = options
+
+        self.driver_io_cls = driver_io_cls
+        self.destination_reader = None
+        self.init_destination_reader()
+
+    def init_destination_reader(self):
+        if self.driver_io_cls and self.positional_parameters:
+            self.destination_reader = DestinationReader(self.driver_io_cls)
+            self.destination_reader.init_driver_io(self.positional_parameters[0])
+
+    def read_log(self):
+        return self.destination_reader.read_logs(self.positional_parameters[0], counter=1)[0]
+
+    def read_logs(self, counter):
+        return self.destination_reader.read_logs(self.positional_parameters[0], counter=counter)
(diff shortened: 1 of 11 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
As you mentioned IRL, this is the last one of the refactor series that prepares Light for the new property-based tests.

I'm going to review the PR once it's ready.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

mitzkia:
After we have discussed the code with @alltilla we have found some parts where the code could be simplified or should changed.
Maybe we could discussed it in comments.

- What we have found: 
  - the ConfigStatement() class could also define and provide a setter for driver_name and group_type objects. Child classes (FileSource(), SourceDriver()) could set their driver_name, or group_type at their level.
  - it is not visible (from a class diagramm) how FileSource() and FileDestination() uses (provides) their "DriverIO" classes (they are just passed in a parameter of `super`). It will be better to make a composition at driver level.

alltilla:
> After we have discussed the code with @alltilla we have found some parts where the code could be simplified or should changed.
> Maybe we could discussed it in comments.
> 
> * What we have found:
>   
>   * the ConfigStatement() class could also define and provide a setter for driver_name and group_type objects. Child classes (FileSource(), SourceDriver()) could set their driver_name, or group_type at their level.
>   * it is not visible (from a class diagramm) how FileSource() and FileDestination() uses (provides) their "DriverIO" classes (they are just passed in a parameter of `super`). It will be better to make a composition at driver level.
>   * mark methods as private (with _) where needed

Actually, the setter might not be the best idea, but the point is, that we have some work to do on this PR still.

tests/python_functional/src/syslog_ng_config/statements/destinations/destination_driver.py
```diff
@@ -20,15 +20,40 @@
...
+
+        self.driver_io_cls = driver_io_cls
+        self.line_parser_cls = line_parser_cls
+        self.destination_reader = None
+        self.init_destination_reader()
+        super(DestinationDriver, self).__init__()
+
+    def init_destination_reader(self):
+        if self.driver_io_cls and self.line_parser_cls and self.positional_parameters:
+            self.destination_reader = DestinationReader(self.driver_io_cls, self.line_parser_cls)
```
furiel: We discussed IRL that it would worth separating the IO related logic (DriverIO) and the Config logic (Driver).

Now a driver-io is embedded into the driver. This causes some problem: their lifecycle / management differs. For example, configuration is applied during start and reload. However an IO, like http server may or may not need to be started right away. Or in a FileSource scenario, we might not want to create the file right away.

These two things might worth store and manage separately.

For example (more like a pseudocode):
```
http = HttpDestination(port=5555)

generator = generator_source(num=1)
http = HttpDestination(port=5555)
syslog_ng.config_add(generator, http)

simulator = HttpSimulator()
simutor.init(parameters)
simulator.start()

syslog_ng.start()
time.sleep(1)
syslog_ng.stop()

simulator.stop()
```

mitzkia: Yes, I have already started this implementation and it seems it will work in our cases.
But this concept will broke our existing user API. Before going ahead I would like to discuss it IRL. So that's why I put back WIP flag

mitzkia:
I am closing this PR, preferred new one.

PR #3028, opened 2019-11-25
fixing some compile problems for centos6

furiel:
We recently introduced some glib calls that only available in newer glib versions. These changes are needed so that syslog-ng could compile on centos6.


kira-syslogng:
Build SUCCESS

PR #3032, opened 2019-11-29
response-handler: leftover fixes

furiel:
This changeset contains fixes left out from https://github.com/syslog-ng/syslog-ng/pull/3007.

Two  notable changes
- The default action for 409 is changed from retry to disconnect.
As I received an offline comment from @gaborznagy, users might resolve the conflict. The problem is not with the message itself. So we should avoid dropping it.
- The error messages contained wrong url, when the response did not return with 2XX code. Actually this is not a regression from #3007, still I wanted to fix.

----

#


kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build SUCCESS

PR #3035, opened 2019-12-03
timeutils: fix month, day name parsing

Kokan:
Only the first two letter were checked in month (Jan, Feb, ...) and days (Mon, Fri) as they are sufficient to differentiate valid months, days from each  other.


kira-syslogng:
Build SUCCESS

PR #3038, opened 2019-12-06
dbld: make the shell command customizable

szemere:
There was a discussion about removing the rm parameter from the shell
target. Thank you all for sharing your opinion in the topic! (See #2982)
With this patch you can override the shell command in your rules.conf file,
so everyone can have their own version of it.

my rules.conf for example (copied from the earlier PR):
```
DOCKER_SHELL=$(DOCKER) inspect $* > /dev/null 2>&1; \
  if [ $$? -eq 0 ]; then \
    $(DOCKER) start -ia $*; \
  else \
    $(DOCKER) run $(DOCKER_RUN_ARGS) -ti --name $* balabit/syslog-ng-$* /source/dbld/shell; \
  fi
```
Question for the reviewers:
Is there any other target which could benefit from this approach?

### DIFF
--- a/dbld/rules
+++ b/dbld/rules
@@ -41,6 +41,8 @@ TARBALL=$(BUILD_DIR)/syslog-ng-$(VERSION).tar.gz
 MODE=snapshot
 CONFIGURE_OPTS=--enable-debug --enable-manpages --with-python=2 --prefix=/install $(CONFIGURE_ADD)
 
+DOCKER_SHELL=$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -ti balabit/syslog-ng-$* /source/dbld/shell
+
 -include dbld/rules.conf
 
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

alltilla:
Personally I do not use `dbld`, but I agree with the change. I do not want to merge this behind the backs of the `dbld` users.
@bazsi, @mitzkia, @furiel: Do you have any more comments about this PR?

MrAnno:
I'm considering this PR as a refactor. If someone wants to use `dbld/rules.conf` to change docker flags for some good/bad/weird reason, I don't mind :)

(Reusing dirty containers is not a good practice, but the PR does not do such a thing.)

PR #3040, opened 2019-12-10
config_database: rename to syslog-ng-cfg-db

alltilla:
The name change is advised, as we are planning to install it to
the `bin` folder later on, where the `syslog-ng` prefix is needed.


kira-syslogng:
Build SUCCESS

PR #3046, opened 2019-12-11
basicfuncs: add `$(dns-resolve-ip)` template function

### LINKED ISSUE #3011: DNS re-write rule
## Description of the problem
The performance of reverse DNS lookup significantly impacts the overall performance of a syslog-ng instance when enabled on a network source.

## Proposed solution
Provide a new re-write function based on the approach of the one added to "fix time stamp" "fix host" would allow the administrator/developer to limit which events may cause a DNS query based on other conditions.

## Alternatives
globally enable or disable reverse dns on any source

## Additional context
Main use case, enrich a message with the actual host name using reverse DNS for a subset of events based on parser or log path conditions

alltilla:
Fixes #3011

---

Resolves the given IP(v4/v6) address to a hostname.

Usage:
```
$(dns-resolve-ip [OPTIONS] ${IP})
```

Valid options are:
```
--use-fqdn=yesno
--use-dns=yesno
--dns-cache=yesno
--normalize-hostnames=yesno
```

The options inherit the global options.

**WARNING**: 
`$(dns-resolve-ip)` uses the same blocking implementation as the source-side DNS resolution.
While the IP is being resolved by the DNS server, message processing might stop even on paths where `$(dns-resolve-ip)` is not present/not used. For more information, see ["Multithreading and scaling in syslog-ng OSE"](https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.25/administration-guide/88#TOPIC-1349634)

The main use case for `$(dns-resolve-ip)` is to granulate the DNS resolution to a subset of messages. It is advised to turn off `use-dns()` from global options, and use `$(dns-resolve-ip)` in a (conditional) `rewrite` block.

---

### DIFF
--- a/modules/basicfuncs/ip-funcs.c
+++ b/modules/basicfuncs/ip-funcs.c
@@ -22,6 +22,8 @@
  */
 
 #include "gsocket.h"
+#include "host-resolve.h"
+#include "cfg-parser.h"
 
 static void
 tf_ipv4_to_int(LogMessage *msg, gint argc, GString *argv[], GString *result)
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
This is a blocking implementation. It is clear that messages are delayed when `$(dns-resolve-ip)` is used on them, but the blocking implementation has additional unexpected consequences:

Not only affected messages or sources/destinations are delayed until the resolution has a result, every worker thread will block; affecting all syslog-ng objects, the whole pipeline. This may take _seconds_.
The dns-cache can make it smoother, but the problem still exists.

If you want to keep the blocking implementation, please add a description about the consequences to the commits/PR , and the same should be reflected in the admin guide as well (as clearly as possible).

alltilla:
> This is a blocking implementation. It is clear that messages are delayed when `$(dns-resolve-ip)` is used on them, but the blocking implementation has additional unexpected consequences:
> 
> Not only affected messages or sources/destinations are delayed until the resolution has a result, every worker thread will block; affecting all syslog-ng objects, the whole pipeline. This may take _seconds_.
> The dns-cache can make it smoother, but the problem still exists.
> 
> If you want to keep the blocking implementation, please add a description about the consequences to the commits/PR , and the same should be reflected in the admin guide as well (as clearly as possible).

We should write something like the description at `use-dns()` in the Admin Guide`:
```
Description: Enable or disable DNS usage. The persist_only option attempts to resolve 
hostnames locally from file (for example from /etc/hosts). The syslog-ng OSE application 
blocks on DNS queries, so enabling DNS may lead to a Denial of Service attack. To prevent 
DoS, protect your syslog-ng network endpoint with firewall rules, and make sure that all 
hosts which may get to syslog-ng are resolvable. This option can be specified globally, and 
per-source as well. The local setting of the source overrides the global option if available.
```

bazsi:
> This is a blocking implementation. It is clear that messages are delayed when `$(dns-resolve-ip)` is used on them, but the blocking implementation has additional unexpected consequences:
> 
> Not only affected messages or sources/destinations are delayed until the resolution has a result, every worker thread will block; affecting all syslog-ng objects, the whole pipeline. This may take _seconds_.
> The dns-cache can make it smoother, but the problem still exists.
> 
> If you want to keep the blocking implementation, please add a description about the consequences to the commits/PR , and the same should be reflected in the admin guide as well (as clearly as possible).

I think would only block the destination side of things, whereas use-dns() will cause the source threads to stop. Especially if we apply this only to a subset of messages, this will actually be an improvement. As I see dns-cache is being used here.

Doing this asynchronously only improves matters if the ordering of messages is affected too (e.g. we would reorder messages as the DNS responses come in), but I don't think we are prepared for that.

So clearly, this should be documented, but I don't see this could be improved without also introducing reordering that would allow us to parallelize.

modules/basicfuncs/ip-funcs.c
```diff
@@ -40,3 +42,121 @@ tf_ipv4_to_int(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+
+static void
+tf_dns_resolve_ip_call(LogTemplateFunction *self, gpointer s, const LogTemplateInvokeArgs *args, GString *result)
+{
+  DnsResolveIpState *state = (DnsResolveIpState *)s;
+  const gchar *hostname;
+  gsize result_len;
+
+  gchar *ip = args->argv[0]->str;
+  GSockAddr *addr = g_sockaddr_inet_or_inet6_new(ip, 0);
```
bazsi: We might want to introduce "-4" and "-6" options for forcing a specific address family instead of trying to autodetect. I have the feeling that the formats permitted by inet_aton() and inet_pton() are overlapping.

also this is a convention used by various ip related utilities (try ping -6)

This will be especially useful for stuff $(dns-resolve-name) where both address families could make sense.

modules/basicfuncs/tests/test_basicfuncs.c
```diff
@@ -136,6 +146,23 @@ Test(basicfuncs, test_str_funcs)
 {
   assert_template_format("$(ipv4-to-int $SOURCEIP)", "168496141");
 
+  if (_is_connected_to_the_internet())
```
bazsi: I think we might want to put this in a separate test function, but this is only a nitpick.

MrAnno:
> I think would only block the destination side of things, whereas use-dns() will cause the source threads to stop. 

LogReader/LogWriter-based drivers don't have dedicated source/destination threads, so a complete I/O worker thread will be blocked (we have only ~ `get_processor_count()` number of workers), so this affects everything that runs in those workers (other non-related source/destinations, rewrite rules, filters, parsers). This is serious side-effect in my opinions, which is known only by us, developers; that's why I think it's important to add a warning to the documentation.

> Especially if we apply this only to a subset of messages, this will actually be an improvement.

Only if the user disables the source-side dns-lookup. `use-dns()` is enabled by default.

> Doing this asynchronously only improves matters if the ordering of messages is affected too (e.g. we would reorder messages as the DNS responses come in), but I don't think we are prepared for that.

I agree.

modules/basicfuncs/ip-funcs.c
alltilla: `inet_aton()` accepts:
 1. `a.b.c.d`
 2. `a.b.c`
 3. `a.b`
 4. `a`

_"components of the dotted address can be specified in decimal, octal (with a leading 0), or hexadecimal, with a leading 0X"_

whereas `inet_pton()` with `AF_INET6` accepts:
 1. `x:x:x:x:x:x:x:x`
 2. abbreviation of contigous 0s with: `::`
 3. `x:x:x:x:x:x:d.d.d.d`

---
So I do not really see an overlap, correct me if I am wrong.
If we can precisely support the auto-detection, then I do not see a definite reason to add `-4` and `-6` flags for this direction of the DNS resolvation. It would be nice to know the reason of the convention used by `ping` (and `ping6`) for example, I will check that!

rfaircloth-splunk:
great work!

modules/basicfuncs/tests/test_basicfuncs.c
```diff
@@ -136,6 +146,23 @@ Test(basicfuncs, test_str_funcs)
...
+    {
+      /* Note: The dns-resolve-ip tests rely on each other, by setting/using the dns-cache */
+      assert_template_format("$(dns-resolve-ip --use-dns=no --dns-cache=no 198.41.0.4)", "198.41.0.4");
```
Kokan: Regarding the tests I would consider the followings:
* stub the actual libc call resolving ip to dns (test should not rely on internet, or network in fact. even if you check if there is internet you could still get failed case due to faulty network, and God help us investigating those ... )
* Split at least dependent tests/assert into their own (as from your comment it is far from clear which tests are dependent, or all of them ?!)

modules/basicfuncs/ip-funcs.c
```diff
@@ -40,3 +42,121 @@ tf_ipv4_to_int(LogMessage *msg, gint argc, GString *argv[], GString *result)
...
+  result = TRUE;
+exit:
+  g_option_context_free(ctx);
+  return result;
+}
+
+static gboolean
+_setup_host_resolve_options(DnsResolveIpState *state, gint *argc, gchar **argv[], GError **error)
+{
+  host_resolve_options_global_defaults(&state->host_resolve_options);
```
Kokan: Have you considered defaulting to the values in global options ? Right now the default values are the default value of global options, and not the value of global options (as normally it works for many options that has global counterpart).

Also I think it would be beneficial even if the current version is kept, you do not state with exact values what is the default (yes, no) but say that the same as the default values of the global options counterpart (as they are subject to change at any time).

Kokan: Please note that in case of `use-cache` not only the cache will be used, but the hosts file as well (that's how `dns-cache` works right now: https://github.com/syslog-ng/syslog-ng/pull/1923).

In fact in global config the `use-dns(no)` and `use-cache(yes)` combination have been eliminated, as that is probably not something that the user actually desires. Do you think it would make sense to disable here that case as well ?

modules/basicfuncs/tests/test_basicfuncs.c
alltilla: I agree! Might be possible to do the same with `lib/tests/test_host_resolve.c`, which (per my understanding) does not run with `make check`, because of the network dependency.
I will move the dependent TCs as well.

modules/basicfuncs/ip-funcs.c
alltilla: Actually, following the global options makes sense, yes. I believe we can access it from `..._prepare()`, where `LogTemplate *parent` is given as an argument, which has a `GlobalConfig *cfg` field.

alltilla: Sure, the context is the same :) I will do it, thanks for the tip!

modules/basicfuncs/tests/test_basicfuncs.c
alltilla: I stubbed on a higher layer, I think it is okay to do so.
Also, now as `use-dns=no` and `use-dns-cache=yes` together are not valid options, we do not have dependent TCs.

kira-syslogng:
Build SUCCESS

modules/basicfuncs/ip-funcs.c
alltilla: We concluded, that we will get back to these options, when `$(dns-resolve-name)` is implemented.

PR #3047, opened 2019-12-13
PR: add PR template to help creating the NEWS file

alltilla:
Creating/reviewing the NEWS file takes more time than it should.

It would worth a try to come up with the NEWS file entries when the PR is reviewed, so we do not miss any PR. Also the knowledge about the PR is the fullest at that time, thus we can make the NEWS entry content quicker.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,10 @@
+
+
+
+------------------------------------------------------------------
+_Summary:_
+<!--
+This section will be present in the NEWS file.
+Remove it, if you do not want to mention this PR in the NEWS file.
+`module`: Description of the change. (#PR, #ISSUE)
+-->


kira-syslogng:
Build SUCCESS

szemere:
Optional: Some times ago we agreed on a common requirements against PR descriptions. If I remember correctly:

1. Start with a sort description about the topic.
2. Give a technical explanation, and a reason behind the PR. (Add some figures if necessary.)
3. Ask at least two questions from the reviewers, so they can test if they really understood the PR, and validate your major decisions.
 - My memories faded out a little bit, so I don't know if it was originally included, but here comes your part: Short summary for the release notes, especially if the change is user visible.

Since we never had a template before, those requirements were not at hand all the time. I think we might want to include the suggestions above if we touch this part of the repo.

alltilla:
> Optional: Some times ago we agreed on a common requirements against PR descriptions. If I remember correctly:
> 
> 1. Start with a sort description about the topic.
> 2. Give a technical explanation, and a reason behind the PR. (Add some figures if necessary.)
> 3. Ask at least two questions from the reviewers, so they can test if they really understood the PR, and validate your major decisions.
> 
> * My memories faded out a little bit, so I don't know if it was originally included, but here comes your part: Short summary for the release notes, especially if the change is user visible.
> 
> Since we never had a template before, those requirements were not at hand all the time. I think we might want to include the suggestions above if we touch this part of the repo.

I like these ideas, and would be happy to add these fields to the template if we all agree on it :)

alltilla:
We concluded, that it is more beneficial to create a `NEWS` directory, where we would put the newsfile entries (1 PR 1 file), then merge it together before release. I will create the framework for it. Hence closing this PR.

PR #2933, opened 2019-09-19
host-resolve: try getaddrinfo() with different flags

alltilla:
### First try:
`AI_NUMERICHOST`: We will check, if the host is declared as a plain IP address, so we can skip the whole hostname lookup phase.
### Second try:
`AI_V4MAPPED | AI_ADDRCONFIG`: We will try to lookup the IP from the hostname, taking into account the system's network interfaces.
  - If the system has an IPv6 interface configured, and there is an IPv6 resolution of the hostname, we use it.
  - If the system only has an IPv4 interface configured, and there is an IPv6 resolution of the hostname, we use it.
  - If the system does not have any network interfaces configured, (lo interface does not count here), we go for the third try.
### Third try:
`AI_V4MAPPED`: We try to resolve the hostname not taking into account the system's network interfaces (we can resolve the lo interface IPs).

### Else:
If we could not resolve the hostname after these steps, there is no resolution for it.

Fixes #2900

Signed-off-by: Attila Szakacs <attila.szakacs@balabit.com>

PR #3048, opened 2019-12-13
Improve name resolution in loggen

MrAnno:
This is a fix for the same problem that #2933 resolves in syslog-ng.
The PR fixes a small memory leak as well.

> When AI_ADDRCONFIG is used, then IPv4 addresses are returned in the list
pointed to by res only if the local system has at least one IPv4 address
configured,  and  IPv6 addresses  are  returned only if the local system
has at least one IPv6 address configured. The loopback address is not
considered for this case as valid as a configured address.

> This flag is useful  on,  for  example,  IPv4-only systems, to ensure
that getaddrinfo() does not return IPv6 socket addresses that would
always fail in connect(2).

Retrying an unsuccessful resolution without `AI_ADDRCONFIG` is necessary because `::1` or `127.0.0.1` might not be resolved when the `AI_ADDRCONFIG` flag is set.

`AI_V4MAPPED` has also been added to support IPv4-mapped IPv6 addresses when no matching IPv6 addresses could be found.

### DIFF
--- a/tests/loggen/loggen_helper.c
+++ b/tests/loggen/loggen_helper.c
@@ -95,9 +95,7 @@ connect_ip_socket(int sock_type, const char *target, const char *port, int use_i
   memset(&hints, 0, sizeof(hints));
   hints.ai_family = use_ipv6 ? AF_INET6 : AF_INET;
   hints.ai_socktype = sock_type;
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

PR #3050, opened 2019-12-20
rewrite: simplify grammar with using common template_content

Kokan:
The `rewrite_template_content` could be replaced with the rule in `cfg-grammar` `template_content`, despite they are not the same.
The `template_content` supports type-hinting, and with this change the rewrite rules will do so.


kira-syslogng:
Build SUCCESS

PR #3051, opened 2019-12-20
geoip2: extend with template option

Kokan:
Opposed to parsers that by default uses `MESSAGE` and that could be further specialised with `template()` options, the `geoip2` always uses `template`. That is enforced with the grammar rule always requiring a template in a form of string: `geoip2("template-format")`.

The PR includes the following changes:
* Extend the `geoip2` grammer with general `parse_opt` rules (currently only includes `template()`
* In init phase enforces the existence of `template`
* Changed UT according to always using `template()` opposed to using `MESSAGE` field, that works for any parser, but never actually used in case of `geoip2`

As a results both are possible configuration option:
```
 geoip2("$HOST");
 geoip2(template("$HOST"));

 # geoip2();  # this is still not valid
```

### DIFF
--- a/modules/geoip2/tests/test_geoip_parser.c
+++ b/modules/geoip2/tests/test_geoip_parser.c
@@ -89,6 +89,15 @@ parse_geoip_into_log_message(const gchar *template_format)
   return msg;
 }
 
+Test(geoip2, template_is_mandatory)
+{
+  LogParser *geoip2_parser = maxminddb_parser_new(configuration);
+
+  cr_assert_not(log_pipe_init(&geoip2_parser->super));
+
+  log_pipe_unref(&geoip2_parser->super);
+}
+
 Test(geoip2, test_basics)
 {
   LogMessage *msg;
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

modules/geoip2/tests/test_geoip_parser.c
```diff
@@ -89,6 +89,15 @@ parse_geoip_into_log_message(const gchar *template_format)
   return msg;
 }
 
+Test(geoip2, template_is_mandatory)
+{
+  LogParser *geoip2_parser = maxminddb_parser_new(configuration);
+
+  cr_assert_not(log_pipe_init(&geoip2_parser->super));
```
furiel: If you want to be sure, you could use message grabbing.

Kokan: With this test I want to protect against ensuring the `geoip2` only can start if there is a template, not that there is an exact error message about it. As this test could do just that, I tend to go with kiss here.

Kokan:
> It should have been implemented originally with the template option.

@furiel do you think marking `geoip2("format");` version deprecated would be okay ? I was thinking about that, but in the end for my original goal it was not required.

PR #3054, opened 2019-12-29
travis/arm: fix criterion version

Kokan:
The ARM build used the bleeding criterion, where the cmake build is broken (switched to meson).
As I do not see any CI or instruction for meson with Criterion, it is better to stick to the released version.
(This was the plan, that as soon as "master" breaks, switch to something fix.)


kira-syslogng:
Build SUCCESS

PR #3056, opened 2020-01-03
checkpoint parser: fix parsing in case of iso timestamp

### LINKED ISSUE #3055: Checkpoint parser not working?
I'm trying to get the checkpoint parser to work on my syslog-ng 3.21 server running on Centos 7.

My checkpoint log exporter is configured as format: syslog, read-mode: raw.

I have copied the plugin.conf from this repo to /etc/syslog-ng/scl/checkpoint/plugin.conf on my server.

This is the relevant block from syslog-ng.conf
```
@version:3.21
@include "scl.conf"

options {
    flush_lines (100);
    time_reopen (10);
    log_fifo_size (1000);
    chain_hostnames (off);
    use_dns (no);
    dns_cache (no);
    use_fqdn (no);
    create_dirs (yes);
    keep_hostname (yes);
    threaded (yes);
};

#############################
## FILTER DEFINITIONS
#############################

filter f_kernel     { facility(kern); };
filter f_default    { level(info..emerg) and
                        not (facility(mail)
                        or facility(authpriv)
                        or facility(cron)); };
filter f_auth       { facility(authpriv); };
filter f_mail       { facility(mail); };
filter f_emergency  { level(emerg); };
filter f_news       { facility(uucp) or
                        (facility(news)
                        and level(crit..emerg)); };
filter f_boot   { facility(local7); };
filter f_cron   { facility(cron); };

#############################
## LOG OUTPUT DEFINITIONS
#############################

log {
    source { network(ip(172.16.0.1) transport(tcp) port(1514) flags(no-parse)); };
    parser { checkpoint-parser(); };
    destination { syslog("172.16.0.2" transport("udp") port(514)); };
};


@include "/etc/syslog-ng/conf.d/*.conf"
```

scl.conf contains this:
```
@include 'scl/*/*.conf'

@define java-module-dir "`module-install-dir`/java-modules"
```

Nothing arrives at the destination. If I comment out the line parser { checkpoint-parser(); }; I get logs at destination.

Kokan:
Fixes #3055 

If the timestamp was in iso format, the `HOST, PROGRAM, PID, MSGID, MSG` values were shifted.

### DIFF
--- a/scl/checkpoint/plugin.conf
+++ b/scl/checkpoint/plugin.conf
@@ -47,15 +47,19 @@ block parser checkpoint-parser(prefix('.checkpoint.')) {
         if {
             # syslog format, we should be handling escaped values properly
             filter { message("<" type(string) flags(prefix)); };
-            parser {
-                csv-parser(columns("1", "2", "3", "HOST", "PROGRAM", "PID", "MSGID", "MSG")
-                           flags(greedy) delimiters(" ") null("-") dialect(escape-none));
-            };
             if {
-                parser { date-parser(format("%Y-%m-%dT%H:%M:%S%z",
-                                            "%Y-%m-%dT%H:%M:%S") template("$2")); };
+                parser {
+                    csv-parser(columns("1", "2", "HOST", "PROGRAM", "PID", "MSGID", "MSG")
+                               flags(greedy) delimiters(" ") null("-") dialect(escape-none));
+                    date-parser(format("%Y-%m-%dT%H:%M:%S%z",
+                                       "%Y-%m-%dT%H:%M:%S") template("$2"));
+                };
             } else {
-                parser { date-parser(format("%Y-%m-%d %H:%M:%S") template("$2 $3")); };
+                parser {
+                    csv-parser(columns("1", "2", "3", "HOST", "PROGRAM", "PID", "MSGID", "MSG")
+                               flags(greedy) delimiters(" ") null("-") dialect(escape-none));
+                    date-parser(format("%Y-%m-%d %H:%M:%S") template("$2 $3"));
+                };
             };
             parser {
                 kv-parser(prefix(`prefix`) value-separator(':') pair-separator(';'));
--- a/tests/python_functional/functional_tests/parsers/checkpoint/test_checkpoint.py
+++ b/tests/python_functional/functional_tests/parsers/checkpoint/test_checkpoint.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+#############################################################################
+# Copyright (c) 2020 Balabit
+# Copyright (c) 2020 Kokan
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
The test I introduced just do exactly that, I copied input messages from `test_topic_syslog_raw.py`, and extended with the example provided in the current issue.

Those test only checked in the `app-parser` classification works or not.

bazsi:
What about this change instead?

```
diff --git a/scl/checkpoint/plugin.conf b/scl/checkpoint/plugin.conf
index 47cdb38e7..aae15d7fd 100644
--- a/scl/checkpoint/plugin.conf
+++ b/scl/checkpoint/plugin.conf
@@ -48,13 +48,15 @@ block parser checkpoint-parser(prefix('.checkpoint.')) {
             # syslog format, we should be handling escaped values properly
             filter { message("<" type(string) flags(prefix)); };
             parser {
-                csv-parser(columns("1", "2", "3", "HOST", "PROGRAM", "PID", "MSGID", "MSG")
+                csv-parser(columns("1", "2", "HOST", "PROGRAM", "PID", "MSGID", "MSG")
                            flags(greedy) delimiters(" ") null("-") dialect(escape-none));
             };
             if {
                 parser { date-parser(format("%Y-%m-%dT%H:%M:%S%z",
                                             "%Y-%m-%dT%H:%M:%S") template("$2")); };
             } else {
+                csv-parser(columns("1", "2", "3", "HOST", "PROGRAM", "PID", "MSGID", "MSG")
+                           flags(greedy) delimiters(" ") null("-") dialect(escape-none));
                 parser { date-parser(format("%Y-%m-%d %H:%M:%S") template("$2 $3")); };
             };
             parser {

```

This will prefer the ISO date format from the performance perspective, as I sincerely hope that that is the way forward. The "old" format would reparse the input string, but will be faster for the preferred case.

Kokan:
@bazsi  I am okay to modify this version, this wasn't the first one I did.

The diff you suggested does not work, as the first csv parse unconditionally replaces the `MSG`, that breaks the non-iso timestamp case. (Just run the test included in this branch.)

A variation of that could be to do the parsing inside the if:
```
        if {
            # syslog format, we should be handling escaped values properly
            filter { message("<" type(string) flags(prefix)); };
            if {
            parser {
                csv-parser(columns("1", "2", "HOST", "PROGRAM", "PID", "MSGID", "MSG")
                           flags(greedy) delimiters(" ") null("-") dialect(escape-none));
                date-parser(format("%Y-%m-%dT%H:%M:%S%z",
                                   "%Y-%m-%dT%H:%M:%S") template("$2"));
            };
            } else {
                parser {
                    csv-parser(columns("1", "2", "3", "HOST", "PROGRAM", "PID", "MSGID", "MSG")
                               flags(greedy) delimiters(" ") null("-") dialect(escape-none));
                    date-parser(format("%Y-%m-%d %H:%M:%S") template("$2 $3"));
                };
            };
            parser {
                kv-parser(prefix(`prefix`) value-separator(':') pair-separator(';'));
            };
            flags(final);
```
Do you think this version is better than the one in the PR ?

From performance perspective I cannot tell which is better. If we count and differentiate them based on the number of modifying pipe chain, this (and yours) is better.

bazsi:
This should perform better if the first alternative matches, which is the
anticipated case.

kira-syslogng:
Build SUCCESS

PR #3058, opened 2020-01-06
snmp adjust engineid length

szemere:
Original code accepted only 10 character long hexadecimal values.
Checked the snmp library and the RFC, we should accept any hexadecimal
number between 5 to 32 characters.
To validate it, moved the checking part from the grammar into the
snmpdest_dd_set_engine_id function, and wrote tests against it.

### DIFF
--- a/modules/snmp-dest/snmpdest.c
+++ b/modules/snmp-dest/snmpdest.c
@@ -220,12 +220,49 @@ void snmpdest_dd_set_community(LogDriver *d, const gchar *community)
   self->community = g_strdup(community);
 }
 
-void snmpdest_dd_set_engine_id(LogDriver *d, const gchar *eid)
+static gboolean
+_has_hexadecimal_prefix(const gchar *s)
+{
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

modules/snmp-dest/snmpdest.c
```diff
@@ -221,12 +220,51 @@ void snmpdest_dd_set_community(LogDriver *d, const gchar *community)
...
 }
 
-void snmpdest_dd_set_engine_id(LogDriver *d, const gchar *eid)
+static gboolean
+_has_hexadecimal_prefix(const gchar *s)
+{
+  return (s[0] == '0') && (s[1] == 'x');
+}
+
+static gboolean
```
norberttak: did you check the possibility to use g_ascii_isxdigit() instead of an own implementation?

kira-syslogng:
Build SUCCESS

PR #3059, opened 2020-01-06
WIP: Unmerge reference and inlined template in configuration

Kokan:
In many places the current `template()` option handled semi-transparent the quoted and non-quoted templates.

The result in both cases are expanding the macro into `foo`. (inlined template)
```
file("/dev/stdout" template(foo)); //non-quoted
file("/dev/stdout" template("foo")); //quoted
```

The result in both cases are expanding the macro into `bar`. (referenced template)
```
template foo "bar";
file("/dev/stdout" template(foo)); //non-quoted
file("/dev/stdout" template("foo")); //quoted
```

The template in this case prefers: referenced over inline template regardless of the quoting.

The proposal is to remove the behaviour with qoute and non-qouted templates.
The qouted template `template("foo")` should always imply inlined template.
The non-qouted template `template(foo)` should always imply referenced template.

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -287,6 +287,7 @@ extern struct _LogRewrite *last_rewrite;
 %token KW_TEMPLATE                    10270
 %token KW_TEMPLATE_ESCAPE             10271
 %token KW_TEMPLATE_FUNCTION           10272
+%token KW_TEMPLATE_REF                10273
 
 %token KW_DEFAULT_FACILITY            10300
 %token KW_DEFAULT_SEVERITY            10301
@@ -1159,7 +1160,7 @@ facility_string
         ;
 
 parser_opt
-        : KW_TEMPLATE '(' LL_STRING ')'		{
+        : KW_TEMPLATE '(' string ')'		{
                                                   GError *error = NULL;
 
                                                   LogTemplate *template = cfg_tree_check_inline_template(&configuration->tree, $3, &error);
@@ -1167,7 +1168,7 @@ parser_opt
                                                   log_parser_set_template(last_parser, template);
                                                   free($3);
                                                 }
-        | KW_TEMPLATE '(' LL_IDENTIFIER ')'	{
+        | KW_TEMPLATE_REF '(' string ')'	{
                                                   LogTemplate *template = cfg_tree_lookup_template(&configuration->tree, $3);
                                                   CHECK_ERROR(template != NULL, @3, "Error compiling template");
                                                   log_parser_set_template(last_parser, template);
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
I am against using quotes for this purpose as it has traditionally been
ignored, and rather introduce a new option (eg. template-ref() ) and
deprecate the old behavior with a config upgrade message.

Unfortunately I have seen people declaring template formats and reference
them by their name, so we will need to give it some time to get rid off
those syntaxes in the wild.

Kokan:
@bazsi I am not against using `template-ref`, but I have to think about that. Additionally I am curios about others opinion.

gaborznagy:
Oops, I forgot to review this. :)

Kokan:
@bazsi 

> I am against using quotes for this purpose as it has traditionally been
> ignored, and rather introduce a new option (eg. template-ref() ) and
> deprecate the old behavior with a config upgrade message.

After yesterday discussion about this topic, the conclusion was to use `template-ref` instead of the current solution. (Just like you suggested.)

> Unfortunately I have seen people declaring template formats and reference
> them by their name, so we will need to give it some time to get rid off
> those syntaxes in the wild.

Do you mean you would eventually get ride of `template "name" "value"` ?

bazsi:
On Wed, Jan 29, 2020, 09:52 Kókai Péter <notifications@github.com> wrote:

> @bazsi <https://github.com/bazsi>
>
> I am against using quotes for this purpose as it has traditionally been
> ignored, and rather introduce a new option (eg. template-ref() ) and
> deprecate the old behavior with a config upgrade message.
>
> After yesterday discussion about this topic, the conclusion was to use
> template-ref instead of the current solution. (Just like you suggested.)
>
> Unfortunately I have seen people declaring template formats and reference
> them by their name, so we will need to give it some time to get rid off
> those syntaxes in the wild.
>
> Do you mean you would eventually get ride of template "name" "value" ?
>
> —
>

No, I meant that people are actually using `template("name")` forms in
their destination statements.

These will need to be migrated to template-ref() but we'll need to give
them time (e.g. couple of years), until the current change is propagated to
broad use.

Bazsi

kira-syslogng:
Build FAILURE

furiel:
Could @Kokan or someone summarize what was the conclusion and what steps are necessary to move forward with with this? Let me know if I need to take over this PR.

gaborznagy:

gaborznagy:
For me it seems that this PR is not WIP, as the requested `template-ref` change is done, right?

kira-syslogng:
Build FAILURE

bazsi:
We discussed this today on our weekly call, another alternative came up. Originally I thought this came up from a syntax cleanness perspective (driven by config-database), and that's why I suggested a syntax based alternative.

The new idea (as some expressed concerns about the template-ref keyword), is to simply delay the decision whether the parameter is the name of a template or a template itself until after the parsing is complete. At that point all the template statements were processed, so we would know if that's a template or not, regardless of ordering in the configuration file.

With that in mind, I don't see a very strong reason to be hasty at this point. We might simply close this until it bites us again.

PR #3060, opened 2020-01-07
[WIP] cmake: compile with openssl from nonstandard location

furiel:
While experimenting with openssl 1.1.1, I noticed the compilation with cmake does not work. I compiled openssl with nonstandard location, and used PKG_CONFIG_PATH. I received the following error:

```
tests/loggen/libloggen_helper.a(loggen_helper.c.o): In function `open_ssl_connection':
/home/furiel/workspace/test-syslogng/syslog-ng/tests/loggen/loggen_helper.c:256: undefined reference to `SSLv23_client_method'
```

The problem was that we only used the detected openssl so only during linking: (`target_link_libraries`: `OpenSSL:SSL`). The include directories of the detected openssl were not added (`include_directories`).
The reason is: in openssl1.1.1, `SSLv23_client_method` became a define for `TLS_client_method`. So if you do not use the correct include file, loggen_helper will look for `SSLv23_client_method` symbol, which is not available in 1.1.1. If the correct include is used, loggen_helper will look for `TLS_client_method` due to the rename in define, and everything works as expected.

There were other cases where `include_directories` for openssl was missing. Those occurrences are also fixed.

The second part of the of the PR is an improvement in loggen: the `loggen_helper` module was created but was not used by `loggen` target. The source files were just added originally. From now, `loggen` will link to `loggen_helper`.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -310,6 +310,8 @@ target_include_directories(syslog-ng
 target_include_directories(syslog-ng
   SYSTEM PUBLIC
     ${CORE_INCLUDE_DIRS}
+  PRIVATE
+    ${OPENSSL_INCLUDE_DIR}
 )
 
 target_link_libraries(
--- a/modules/afsocket/CMakeLists.txt
+++ b/modules/afsocket/CMakeLists.txt
@@ -55,7 +55,7 @@ find_package(systemd)
 add_library(afsocket SHARED ${AFSOCKET_SOURCES})
 target_include_directories(afsocket
   PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}
-  PRIVATE ${CMAKE_CURRENT_BINARY_DIR}
+  PRIVATE ${CMAKE_CURRENT_BINARY_DIR} ${OPENSSL_INCLUDE_DIR}
 )
 target_link_libraries(afsocket PRIVATE
 	${WRAP_LIBRARIES}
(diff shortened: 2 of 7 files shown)


kira-syslogng:
Build SUCCESS

lib/CMakeLists.txt
```diff
@@ -310,6 +310,8 @@ target_include_directories(syslog-ng
 target_include_directories(syslog-ng
   SYSTEM PUBLIC
     ${CORE_INCLUDE_DIRS}
+  PRIVATE
```
Kokan: The "missing" include directories are not a bug but a feature of cmake: https://github.com/syslog-ng/syslog-ng/pull/2007

furiel: The mentioned compile error shows that the initial assumption in #2007 was wrong. Setting `target_link_libraries` does not set `include_directories` for `Openssl:SSL`.

MrAnno: It's not an assumption, it's a design decision in modern CMake, so you can "advertise" your own build specifications; transitive dependencies, include paths, and compiler flags without creating boilerplate and error-prone constructs on the user side.

Basically you define "packages" (targets) instead of random variables, and on the user side, you only have to use `target_link_libraries()`, which has very little to do with the low-level "linking" concept. (Please read #2007, [1](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#build-specification-and-usage-requirements), [2](https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/)).

`OpenSSL::SSL` is an imported target that uses the same modern concept, and therefore `include_directories()` is clearly set:
https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSSL.cmake#L502-L503

furiel: Ok, I understand your motivation. I try to find another resolution. Adding WIP.

furiel:
@Kokan @MrAnno you were right, my understanding was incomplete. The problem lies elsewhere. I close this PR and restart from scratch.

PR #3067, opened 2020-01-10
loggen: cmake: install dev headers

lbudai:
When syslog-ng is built with automake, it ships `loggen_helper.h` and `loggen_plugin.h`.
Now, cmake also installs these headers.


kira-syslogng:
Build SUCCESS

PR #3070, opened 2020-01-13
compat/getent: fix compile error on solaris

### LINKED ISSUE #2631: make error with 3.20.1 on solaris 11.3
# syslog-ng

Hi
I am getting an error when running gmake.

error: conflicting types for (compat_generic__getprotobynumber_r)
error: conflicting types for (compat_generic__getprotobyname_r)
error: conflicting types for (compat_generic__getservbyport_r)
error: conflicting types for (compat_generic__getservbynamer_r)

modules/getent/module_getent_libtfgetent_la-tfgetent.lo error 1

/Y

alltilla:
The actual functions look like this:

```
#define getprotobynumber_r _compat_sun__getprotobynumber_r

/* From getent-sun.h */
/* From netdb.h (Solaris type) */

struct protoent *getprotobyname_r(const char *name,
                                  struct protoent *result, char *buffer,
                                  int buflen);

/* End from netdb.h */

int _compat_sun__getprotobynumber_r(int proto,
                                    struct protoent *result_buf, char *buf,
                                    size_t buflen, struct protoent **result);

/* End from getent-sun.h */
```

The signature are different in the 2 functions, and that is what we are fixing in the `compat` layer. We would like to use the `_compat_sun__getprotobynumber_r()` signature for `getprotobynumber_r()`.
`_compat_sun__getprotobynumber_r()` itself is just a wrapper:

```
int
_compat_sun__getprotobynumber_r(int proto,
                                struct protoent *result_buf, char *buf,
                                size_t buflen, struct protoent **result)
{
  *result = getprotobynumber_r(proto, result_buf, buf, buflen);
  return (*result ? NULL : errno);
}
```

We should not write `#define getprotobynumber_r _compat_sun__getprotobynumber_r` before declaring these functions, because it is intended to change the `getprotobyname_r()` at the call site, and not the declaration site.

Fixes #2631

### DIFF
--- a/lib/compat/getent.h
+++ b/lib/compat/getent.h
@@ -28,21 +28,21 @@
 
 #ifndef SYSLOG_NG_HAVE_GETPROTOBYNUMBER_R
 
+#include "getent-generic.h"
+
 #define getprotobynumber_r _compat_generic__getprotobynumber_r
 #define getprotobyname_r _compat_generic__getprotobyname_r
 #define getservbyport_r _compat_generic__getservbyport_r
 #define getservbyname_r _compat_generic__getservbyname_r
 
-#include "getent-generic.h"
-
 #elif defined(sun) || defined(__sun)
+
+#include "getent-sun.h"
 
 #define getprotobynumber_r _compat_sun__getprotobynumber_r
 #define getprotobyname_r _compat_sun__getprotobyname_r
 #define getservbyport_r _compat_sun__getservbyport_r
 #define getservbyname_r _compat_sun__getservbyname_r
-
-#include "getent-sun.h"
 
 #endif
 #endif


kira-syslogng:
Build SUCCESS

furiel:
In the description, please add something like: getent: fix compile error on solaris

lib/compat/getent.h
```diff
@@ -28,21 +28,21 @@
...
 #define getprotobynumber_r _compat_generic__getprotobynumber_r
 #define getprotobyname_r _compat_generic__getprotobyname_r
 #define getservbyport_r _compat_generic__getservbyport_r
 #define getservbyname_r _compat_generic__getservbyname_r
 
-#include "getent-generic.h"
-
 #elif defined(sun) || defined(__sun)
 
+#include "getent-sun.h"
```
furiel: Can you explain the root cause and why the reordering fixes the problem?
Please update the commit message accordingly.

alltilla: Thanks for the tip, done :)

kira-syslogng:
Build SUCCESS

furiel:
We were thinking a little bit more on the root cause, and it turns out it is somewhat tricky, so let me elaborate.

The problem is inside getent.h, we override `getprotobynumber_r` to another string with a define. Then we include getent-sun.h, that in terms, include solaris specific getent-related headers, that ship with their own `getprotobynumber_r` declaration, that differ from what we want to use in compat. If they were the same, we would not need to have compat at all. What is happening, is essentially our define edits everything, including solaris headers unintentionally, and hence the error.

Let me show an example on a smaller scale.

```C
$ cat foo.h 
void bar();
$ cat main.c
#define bar foo
#include "foo.h"
int foo() {}
int main(){}
$ 
```

During compile, this will be put into a single file, after expanding the include we will this:
```C
#define bar foo
void bar();
int foo() {}
int main(){}
```

After collapsing the define, we have:
```C
#define bar foo
void foo();
int foo() {}
int main(){}
```

The `void foo();` and `int foo();` declarations conflict.

alltilla:
The actual functions look like this:

```
#define getprotobynumber_r _compat_sun__getprotobynumber_r

/* From getent-sun.h */
/* From netdb.h (Solaris type) */

struct protoent *getprotobyname_r(const char *name,
                                  struct protoent *result, char *buffer,
                                  int buflen);

/* End from netdb.h */

int _compat_sun__getprotobynumber_r(int proto,
                                    struct protoent *result_buf, char *buf,
                                    size_t buflen, struct protoent **result);

/* End from getent-sun.h */
```

The signature are different in the 2 functions, and that is what we are fixing in the `compat` layer. We would like to use the `_compat_sun__getprotobynumber_r()` signature for `getprotobynumber_r()`.
`_compat_sun__getprotobynumber_r()` itself is just a wrapper:

```
int
_compat_sun__getprotobynumber_r(int proto,
                                struct protoent *result_buf, char *buf,
                                size_t buflen, struct protoent **result)
{
  *result = getprotobynumber_r(proto, result_buf, buf, buflen);
  return (*result ? NULL : errno);
}
```

We should not write `#define getprotobynumber_r _compat_sun__getprotobynumber_r` before declaring these functions, because it is intended to change the `getprotobyname_r()` at the call site, and not the declaration site.

alltilla:
@furiel: Updated the commit message! Thanks again :)

kira-syslogng:
Build SUCCESS

PR #3071, opened 2020-01-13
packaging: add packaging/rhel to the tarball build

bazsi:
This branch adds packaging/rhel to the tarball and also simplifies the way the debian related files are included (e.g. instead of using a list of files, simply include the entire directory).


kira-syslogng:
Build SUCCESS

PR #3072, opened 2020-01-13
doc/man: add manual page for persist-tool

bazsi:
While preparing the Debian packages for 3.25.1, it was discovered that persist-tool is without a manual page. The xml file was recovered from the docbook versions of the OSE documentation.

btw: the way manual pages are maintained/taken back from the documentation tree is currently broken. That is not addressed by this branch.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
Tested by building the manpages, and they render nicely on my machine. The links are indeed outdated, the only documented command is `cat`. `info`, `relocate`, `assign` is not documented. Those might worth addressing to in another round.

PR #3074, opened 2020-01-13
Only complain about older config version if there was a change

### LINKED ISSUE #3068: Default RPM update behavior forces  compatibility mode
## Description

The default RPM behaviour is to treat the syslog-ng.conf file as a configuration file and NOT replace it during RPM update. Although this is appropriate for a configuration file the result is that the @Version configuration item remains unchanged which forces the updated syslog-ng into compatibility mode.

## Proposed solution

For those users who wish to change the syslog-ng.conf file, but wish to keep up to date with the @Version configuration I propose that the version.conf file be included in the RPM so that it can be included into the user configured syslog-ng.conf with the outcome that compatibility mode is not forced.

## Alternatives

Devise a versioning of the configuration file that is not locked to the version of syslog-ng. The configuration version would only change when a backwards compatibility break occurs in the configuration. New configuration items can be added without a break in backwards compatibility, so it would be rare to require a configuration version change, and would be appropriate to force compatibility mode under such a condition. For example, @Version:3.9 through @Version 3.18 might all be the same meaning, and be accepted by all versions of syslog-ng 3.9 through 3.18 because there were nod breaking changes in the configuration syntax. Then starting in 3.19 the @Version would require an update and a verification that the client configuration file meets the new changed requirements.

## Additional context

A patch file for the syslog-ng.spec file has been attached.
[syslog-ng-version-conf.patch.gz](https://github.com/syslog-ng/syslog-ng/files/4050926/syslog-ng-version-conf.patch.gz)

bazsi:
Resolves #3068 

On syslog-ng upgrade, if the `@version:` number in the config file left on the older version, a warning `WARNING: Configuration file format is too old, syslog-ng is running in compatibility mode. Please update it to use the syslog-ng ...` can be seen.

This PR makes it possible that a syslog-ng upgrade does not have to be followed by a version number upgrade in the config file, except there is a change in compatibility.
In a lot of cases there is no compatibility changes between 2 versions, so this saves some unnecessary config editing.

It's important, that users should still use the installed syslog-ng version number in their configuration,
even if the `syslog-ng -V` `Config Version` reports the version which is compatible with.

### DIFF
--- a/lib/cfg.h
+++ b/lib/cfg.h
@@ -172,7 +172,7 @@ void register_source_mangle_callback(GlobalConfig *src, mangle_callback cb);
 void uregister_source_mangle_callback(GlobalConfig *src, mangle_callback cb);
 
 static inline gboolean
-cfg_is_config_version_older(GlobalConfig *cfg, gint req)
+__cfg_is_config_version_older(GlobalConfig *cfg, gint req)
 {
   if (!cfg)
     return FALSE;
@@ -180,6 +180,16 @@ cfg_is_config_version_older(GlobalConfig *cfg, gint req)
     return FALSE;
   return TRUE;
 }
+
+/* VERSION_VALUE_LAST_SEMANTIC_CHANGE needs to be bumped in versioning.h whenever
+ * we add a conditional on the config version anywhere in the codebase.  The
+ * G_STATIC_ASSERT checks that we indeed did that */
+
+#define cfg_is_config_version_older(__cfg, __req) \
+  ({ \
+    /* check that VERSION_VALUE_LAST_SEMANTIC_CHANGE is set correctly */ G_STATIC_ASSERT((__req) <= VERSION_VALUE_LAST_SEMANTIC_CHANGE); \
+    __cfg_is_config_version_older(__cfg, __req); \
+  })
 
 static inline void
 cfg_set_use_uniqid(gboolean flag)
--- a/lib/versioning.h
+++ b/lib/versioning.h
@@ -156,6 +156,13 @@
 #define VERSION_STR_CURRENT     "3.25"
 #define VERSION_PRODUCT_CURRENT VERSION_3_25
 
+/* this value points to the last syslog-ng version where we changed the
+ * meaning of any setting in the configuration file.  Basically, it is the
+ * highest value passed to any cfg_is_config_version_older() call.
+ */
+#define VERSION_VALUE_LAST_SEMANTIC_CHANGE  VERSION_VALUE_3_22
+#define VERSION_STR_LAST_SEMANTIC_CHANGE    "3.22"
+
 #define version_convert_from_user(v)  (v)
 
 #include "pe-versioning.h"
--- a/lib/versioning.h
+++ b/lib/versioning.h
@@ -121,6 +121,9 @@
 #define VERSION_3_24 "syslog-ng 3.24"
 #define VERSION_3_25 "syslog-ng 3.25"
 
+/* VERSION_VALUE_* references versions as integers to be compared against stuff like cfg->user_version */
+/* VERSION_STR_* references versions as strings to be shown to the user */
+
 #define VERSION_VALUE_3_0  0x0300
 #define VERSION_VALUE_3_1  0x0301
 #define VERSION_VALUE_3_2  0x0302
@@ -148,11 +151,10 @@
 #define VERSION_VALUE_3_24 0x0318
 #define VERSION_VALUE_3_25 0x0319
 
-
 /* config version code, in the same format as GlobalConfig->version */
-#define VERSION_VALUE   0x0319
-#define VERSION_CURRENT VERSION_3_25
-#define VERSION_CURRENT_VER_ONLY "3.25"
+#define VERSION_VALUE_CURRENT   0x0319
+#define VERSION_STR_CURRENT     "3.25"
+#define VERSION_PRODUCT_CURRENT VERSION_3_25
 
 #define version_convert_from_user(v)  (v)
 
--- a/syslog-ng/main.c
+++ b/syslog-ng/main.c
@@ -137,7 +137,7 @@ version(void)
       installer_version = g_strdup(SYSLOG_NG_VERSION);
     }
   printf(SYSLOG_NG_PACKAGE_NAME " " SYSLOG_NG_COMBINED_VERSION "\n"
-         "Config version: " VERSION_STR_CURRENT "\n"
+         "Config version: " VERSION_STR_LAST_SEMANTIC_CHANGE "\n"
          "Installer-Version: %s\n"
          "Revision: " SYSLOG_NG_SOURCE_REVISION "\n",
          installer_version);
--- a/syslog-ng/main.c
+++ b/syslog-ng/main.c
@@ -137,7 +137,7 @@ version(void)
       installer_version = g_strdup(SYSLOG_NG_VERSION);
     }
   printf(SYSLOG_NG_PACKAGE_NAME " " SYSLOG_NG_COMBINED_VERSION "\n"
-         "Config version: " VERSION_CURRENT_VER_ONLY "\n"
+         "Config version: " VERSION_STR_CURRENT "\n"
          "Installer-Version: %s\n"
          "Revision: " SYSLOG_NG_SOURCE_REVISION "\n",
          installer_version);
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -401,7 +401,7 @@ cfg_set_version(GlobalConfig *self, gint version)
       return FALSE;
     }
 
-  if (cfg_is_config_version_older(self, VERSION_VALUE))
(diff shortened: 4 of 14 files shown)


kira-syslogng:
Build FAILURE

lib/versioning.h
```diff
@@ -151,8 +151,8 @@
 
 /* config version code, in the same format as GlobalConfig->version */
 #define VERSION_VALUE   0x0319
-#define VERSION_CURRENT VERSION_3_25
```
alltilla: Is this intentionally removed?

gaborznagy: I agree with @alltilla, it's needed.
It is still used in code, that's the reason for the failed Travis jobs.
code:
```
lib/cfg.c-  if (cfg_is_config_version_older(self, VERSION_LAST_SEMANTIC_CHANGE))
lib/cfg.c-    {
lib/cfg.c-      msg_warning("WARNING: Configuration file format is too old, syslog-ng is running in compatibility mode. "
lib/cfg.c:                  "Please update it to use the " VERSION_CURRENT " format at your time of convenience. "
lib/cfg.c-                  "To upgrade the configuration, please review the warnings about incompatible changes printed "
lib/cfg.c-                  "by syslog-ng, and once completed change the @version header at the top of the configuration "
```

We should encourage to use the latest config version, even there were no semantic change between 2 versions.

lib/versioning.h
```diff
@@ -152,6 +152,7 @@
 /* config version code, in the same format as GlobalConfig->version */
 #define VERSION_VALUE   0x0319
 #define VERSION_CURRENT_VER_ONLY "3.25"
+#define VERSION_LAST_SEMANTIC_CHANGE VERSION_3_22
```
gaborznagy: This value has to be manually maintained every time we change the **config version** (i.e. new `cfg_is_config_version_older` call), which has the risk that we forget it.

Until we find out something better, please add a one-line comment above it suggesting that this has to be the same as the latest version value used in all `cfg_is_config_version_older` calls.

lib/versioning.h
czanik: No. This warning should only appear if there is a real backwards incompatible change.

One of the reasons that distros don't allow upgrade syslog-ng is the change of version number in syslog-ng.conf and the warning coming with it. If syslog-ng warns only when there is a real change, we could do version upgrades instead of trying to hand pick patches.

gaborznagy: > This warning should only appear if there is a real backwards incompatible change.

I agree, and this is what this PR solves.
My saying is, that if the user uses a feature with an old config, say 3.18, and that feature changed with version 3.22, we should say that "although it changed with version 3.22, use version 3.25 instead".

Of course from config version point of view there is no difference between 3.22 and 3.25, but I think upgrading syslog-ng itself to 3.25 and then setting the config version to 3.22 gives a false sense of an LTS version.

gaborznagy: In `syslog-ng -V` output the "Config Version" is the latest release, so I guess this supports my argument.

https://github.com/syslog-ng/syslog-ng/blob/acdd456209f5a629a1e21f1bbdfd3f5cee4ae975/syslog-ng/main.c#L139-L143

```
$ sbin/syslog-ng -V
syslog-ng 3 (3.25.1.78.gd5fa5ef)
Config version: 3.25
Installer-Version: 3.25.1.78.gd5fa5ef
Revision: 
Compile-Date: Jan 17 2020 09:36:30
```

bazsi:
removed a compile error (sorry for not noticing) and also added a G_STATIC_ASSERT() to check
if VERSION_LAST_SEMANTIC_CHANGE wasn't updated properly.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/cfg.h
```diff
@@ -181,6 +181,16 @@ cfg_is_config_version_older(GlobalConfig *cfg, gint req)
   return TRUE;
 }
 
+/* VERSION_LAST_SEMANTIC_CHANGE needs to be bumped in versioning.h whenever
+ * we add a conditional on the config version anywhere in the codebase.  The
+ * G_STATIC_ASSERT checks that we indeed did that */
+
+#define cfg_is_config_version_older(__cfg, __req) \
+  ({ \
+    /* check that VERSION_LAST_SEMANTIC_CHANGE is set correctly */ G_STATIC_ASSERT((__req) <= VERSION_LAST_SEMANTIC_CHANGE); \
```
gaborznagy: I didn't understood why is this a one liner, but then saw the make output.
Nice way of adding error message support. :)
```
../lib/cfg.h:190:68: note: in expansion of macro ‘G_STATIC_ASSERT’
     /* check that VERSION_LAST_SEMANTIC_CHANGE is set correctly */ G_STATIC_ASSERT((__req) <= VERSION_LAST_SEMANTIC_CHANGE); \
```

gaborznagy:
@bazsi Yesterday we discussed this topic and stated that we should clearly separate config version from the actual version of syslog-ng.

This means that `syslog-ng -V` output with `Config version:` could show the last semantic version, i.e. the config version.

I've shared my personal opinion on this topic before, and I would like to have some feedback on it both from you and the team (reviewers).
> ...  from config version point of view there is no difference between 3.22 and 3.25, but I think upgrading syslog-ng itself to 3.25 and then setting the config version to 3.22 gives a false sense of an LTS version.
See https://github.com/syslog-ng/syslog-ng/pull/3074#discussion_r367421608

bazsi:
I've just rebased/updated the branch to display the last semantic change in config-version. I did some cleanups in the naming of macros in versioning.h as I tended to forget which macro I needed and by the time I've checked it out and got it to the place I wanted to use it, I forgot it again.

The conventions I introduced are documented in the versioning.h file.

Also I changed the "Config-Version" in the --version output to use a dash in place of the space, to match conventions in that output. I don't think this has been parsed by tools until now.

Sidenote (but not included in this branch): I found that we are using SYSLOG_NG_SOURCE_REVISION at a number of places but it is not set in automake based builds, so we can quite possibly eliminate that now as the version number does contain the git commit id (or we should set it to the git commit id as it was originally intended). I

kira-syslogng:
Build SUCCESS

lib/versioning.h
```diff
@@ -148,11 +151,10 @@
...
 #define VERSION_VALUE_3_25 0x0319
 
-
 /* config version code, in the same format as GlobalConfig->version */
-#define VERSION_VALUE   0x0319
-#define VERSION_CURRENT VERSION_3_25
-#define VERSION_CURRENT_VER_ONLY "3.25"
+#define VERSION_VALUE_CURRENT   0x0319
+#define VERSION_STR_CURRENT     "3.25"
+#define VERSION_PRODUCT_CURRENT VERSION_3_25
```
gaborznagy: The `VERSION_VALUE` and `VERSION_STR` conventions are okay, but the `VERSION_PRODUCT_CURRENT` I don't understand.
The previous names were a bit clearer: 
- we had version number as an integer(VERSION_VALUE),
- we had version number as a string (VER_ONLY)
- we had a version number as a string including the product name (VERSION_xx).

The only usage of `VERSION_PRODUCT_CURRENT` is in a debug message, where the product name and the current version is needed.
I would remove this macro (so we don't have to bump it's version on every release) and refactor the debug message.

syslog-ng/main.c
```diff
@@ -137,7 +137,7 @@ version(void)
       installer_version = g_strdup(SYSLOG_NG_VERSION);
     }
   printf(SYSLOG_NG_PACKAGE_NAME " " SYSLOG_NG_COMBINED_VERSION "\n"
-         "Config version: " VERSION_STR_CURRENT "\n"
+         "Config-version: " VERSION_STR_LAST_SEMANTIC_CHANGE "\n"
```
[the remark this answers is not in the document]
smortex: For what it worth, some consumers of syslog-ng rely on the output of `syslog-ng --version` (e.g. the [ccin2p3/syslog_ng](https://forge.puppet.com/ccin2p3/syslog_ng) Puppet module [exports the values as a structured fact](https://github.com/ccin2p3/puppet-syslog_ng/blob/master/lib/facter/syslog_ng.rb)), so changing the label of existing entries _may_ have unexpected side effects.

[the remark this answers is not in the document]
gaborznagy: We agreed that "Config version" will be left as it was (without dashes).

lib/versioning.h
gaborznagy: As discussed this can go in as it is.
We may need to refactor this anyway, due to an issue with but that's not yet decided.

bazsi:
rebased, and changed the "config version" output to use a space instead of a dash.

kira-syslogng:
Build SUCCESS

bazsi:
Rebased and added news entry.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

gaborznagy:

gaborznagy:
@bazsi I've resolved the merge conflict and pushed it.

kira-syslogng:
Build SUCCESS

PR #3075, opened 2020-01-14
Fix file source not able to process new message after log-msg-size increase

Kokan:
The reproduction steps are the following:
* Add a simple file source (default configuration everything)
* Send a larger message than the default log-msg-size
* syslog-ng correctly split that message (and everything works as expected)
* Change log-msg-size (insert a msg size that now larger then the sent large message before)
* Reload or Restart syslog-ng
* Send the large message again

Before this patch syslog-ng cannot process the messages from this file.

The persist file in fact stores the log-msg-size, and it is not possible to change for a file source even in case of restart.
A proposed solution is simply increasing the buffer, and updating internal buffer size so it can accommodate the bigger message size.


kira-syslogng:
Build SUCCESS

PR #3078, opened 2020-01-15
Azure auth header

lbudai:
Azure auth header is a signal-slot plugin that generates authorization header for applications connecting to Azure.

Original version developed by Ferenc Sipos.

Differences from original version:
 * interface cleanups
 * code cleanups
 * use signal-slot mechanism instead of the  `HttpAuthHeader`  interface (which will be removed soon)
 * GLib HMAC functions replaced by OpenSSL HMAC implementation


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/azure-auth-header/azure-auth-header.c
```diff
@@ -0,0 +1,224 @@
...
+
+  gsize date_max_len = 64;
+  gchar date[date_max_len];
+  gsize date_len = 0;
+
+  gsize   digest_len = 0;
+  guint8 *digest = NULL;
+  gchar  *digest_str = NULL;
+
+  date_len = _get_rfc1123date(date, date_max_len);
```
furiel: We are not really interested in`date_len`, other than the `stftime` was successful or not inside `_get_rfc1123date`. The assert could be moved there: arount `stftime`.

modules/azure-auth-header/azure-auth-header.c
```diff
@@ -0,0 +1,224 @@
...
+  g_assert(date_len);
+
+  rawstr = _azure_auth_header_get_str_to_hash(self, body->len, date);
+  digest_len = _azure_auth_header_get_digest(self, &digest, rawstr);
+  digest_str = g_base64_encode(digest, digest_len);
```
furiel: Please note that in some glib versions base64 encoding is buggy. For example, we workarounded it in `tf_base64encode`. I think we cannot use `g_base64_encode`, and write similar code that in `tf_base64encode`.

modules/azure-auth-header/azure-auth-header.c
```diff
@@ -69,7 +68,7 @@ _azure_auth_header_get_str_to_hash(AzureAuthHeaderPlugin *self, glong content_le
 static gsize
 _azure_auth_header_get_digest(AzureAuthHeaderPlugin *self, guint8 **digest, const GString *input)
 {
-  gsize digest_len = g_checksum_type_get_length(self->alg);
+  gsize digest_len = g_checksum_type_get_length(G_CHECKSUM_SHA256);
```
furiel: This patch can be squashed into the original patch. I do not see much reason to keep it separate.

modules/azure-auth-header/azure-auth-header.c
```diff
@@ -54,6 +56,14 @@ _get_rfc1123date(gchar *buf, gsize buf_len)
   return len;
 }
 
+static void
+_get_rfc1123data_or_assert(gchar *date)
+{
+  gsize date_len = 0;
+  date_len = _get_rfc1123date(date, MAX_DATE_LEN);
```
furiel: With my previous comment, this function extraction/abstraction is not necessary.

modules/azure-auth-header/azure-auth-header.c
lbudai: thanks, fixed.

kira-syslogng:
Build SUCCESS

lbudai:
Note: the wip flag is just because we need to do some additional internal tests...
I'm not planning to change the code, so it is ready to review

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/azure-auth-header/azure-auth-header.c
```diff
@@ -0,0 +1,226 @@
...
+            self->secret, self->secret_len,
+            (const guchar *)input->str, input->len,
+            digest, &md_len))
+    {
+      msg_error("Failed to generate Azure Auth Header HMAC",
+                evt_tag_str("str", input->str),
+                evt_tag_int("len", input->len));
+    }
+
+  return md_len;
```
MrAnno: In case `HMAC()` fails, I think it would be better to return early with an explicit 0, because it is not specified by the documentation whether `md_len` is undefined or untouched when the operation fails.

modules/azure-auth-header/azure-auth-header.c
```diff
@@ -0,0 +1,226 @@
...
+{
+  gsize date_len = 0;
+  date_len = _get_rfc1123date(date, MAX_DATE_LEN);
+  g_assert(date_len);
+}
+
+static GString *
+_azure_auth_header_get_str_to_hash(AzureAuthHeaderPlugin *self, glong content_len, const gchar *date)
+{
+  gchar *fmt = "%s\n%ld\n%s\nx-ms-date:%s\n%s";
```
MrAnno: This should be const.

modules/azure-auth-header/azure-auth-header.c
```diff
@@ -0,0 +1,226 @@
...
+
+
+void
+azure_auth_header_secret_set_from_b64str(AzureAuthHeaderPlugin *self, const gchar *b64secret)
+{
+  g_free(self->secret);
+  self->secret = g_base64_decode(b64secret, &self->secret_len);
+
+  if (!self->secret)
+    msg_error("Failed to base64 decode Azure secret key");
```
MrAnno: Most of the options can be validated only at run-time after trying to send the first message.

This can be pre-checked (at least if its a valid base64 string). Can we make this a bool function and fail with CHECK_ERROR() in the grammar?

modules/azure-auth-header/azure-auth-header.c
```diff
@@ -0,0 +1,226 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "azure-auth-header.h"
```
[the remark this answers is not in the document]
lbudai: `driver.h` includes `syslog-ng.h` and `driver.h` is included by `azure-auth-header.d`

modules/azure-auth-header/azure-auth-header.c
lbudai: Well... it's not a secret that `HMAC` is a helper function around the `HMAX_CTX` (new, init, update, final, free). The `HMAC_Final` needs to check whether the mdlen is `NULL` (just check the doc) ... which should mean that it is set only after everything went fine. So even it is `not specified` it will work as expected.
But I can early return...

modules/azure-auth-header/azure-auth-header.c
lbudai: yes. I just did not touch some parts of the original code

modules/azure-auth-header/azure-auth-header.c
lbudai: ...or `static const`  (or just a string literal defined by a macro...)

lbudai:
@MrAnno :
Thanks for the review notes. 
I think I've fixed all the review notes (even those ones that you would think I didn't ;) )

lbudai:
@mitzkia : could you run a manual test on this branch with the well-prepared `SCL` wrapper?

If passed, then you can merge this after the final reviews.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
I've tested it manually by myself, sentinel responded with 
```
status_code='200'
```

so I think it is okay.

PR #3079, opened 2020-01-16
commit-check: support whitespaces around commas

gaborznagy:
Our commit-check script is still too strict.

This pattern failed on commit-check due to the space between elements:
'lib, modules: don't use numeric version identifiers'

From now on, even these are supported:

'lib,    modules,	: fixup'
"li b, ,modules: fixup"

I would prefer a more relaxed commit-check,
as the basic motivation for commit-check was to enforce this pattern:
`modulename: short description of the commit`

I dedicate this PR to the failed Travis job of #3074  :+1:

### DIFF
--- a/tests/commits/check.sh
+++ b/tests/commits/check.sh
@@ -35,7 +35,7 @@ git log --no-merges --pretty="%H" $commit_range -- | (
 
     subject=$(git show -s --format='%s' $commit)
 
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

lbudai:
my personal opinion: never liked these commit check scripts... 
I think sometimes these checkers could be very-very annoying.

alltilla:
This whitespace thing was annoying, I agree. Thanks for fixing it! :)

PR #2811, opened 2019-06-28
Fix TLS connection closure (source side)

MrAnno:
This pull request fixes the TLS connection closure on the source (receiver) side by sending back close_notify alerts.

RFC5425:
>  Once the transport receiver gets close_notify from the transport sender, it MUST reply with a close_notify unless it becomes aware that the connection has already been closed by the transport sender (e.g., the closure was indicated by TCP).

TODO:
- The destination (sender) side has to be fixed as well (in a separate PR):

> A transport sender MUST close the associated TLS connection if the connection is not expected to deliver any syslog messages later.  It MUST send a TLS close_notify alert before closing the connection.  A transport sender (TLS client) MAY choose to not wait for the transport receiver's close_notify alert and simply close the connection, thus generating an incomplete close on the transport receiver (TLS server) side.

Implementing this on the destination side is more complicated, we might need to introduce `log_proto_client_shutdown() -> log_transport_shutdown()` because `SSL_shutdown()` requires non-blocking write operations, so fd watches should be registered during shutdown. Currently, we destroy/shutdown transports after stopping watches and deiniting `LogWriter`. Another tricky part is `log_writer_reopen_deferred()`.

Fixes #2703

PR #3082, opened 2020-01-17
tlscontext: add workaround for a TLS 1.3 bug to prevent data loss

MrAnno:
This is a workaround for an OpenSSL TLS 1.3 bug that results in data loss when one-way protocols are used and a connection is closed by the client right after sending data.

"TLS 1.3 session tickets make it impossible to reliably implement communication patterns where the server never sends application-level data."

- https://github.com/openssl/openssl/issues/10880
- https://github.com/openssl/openssl/issues/7948

`SSL_CTX_set_num_tickets(0)` disables TLS 1.3 session tickets (no session ticket will be sent to the client after a full handshake). Tickets are used for session resumption.

An alternative workaround would be doing a bidirectional SSL shutdown (finishing #2811), but syslog-ng is usually used together with other applications that may not take the bidi shutdown step (it is not mandatory, not even in RFC5425).


MrAnno:
@kira-syslogng do perftest

kira-syslogng:
Build SUCCESS

kira-syslogng:

furiel:
I am not understanding the context entirely, just coming from the man page:

https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_num_tickets.html

> The default number of tickets is 2; the default number of tickets sent following a resumption handshake is 1 but this cannot be changed using these functions. The number of tickets following a resumption handshake can be reduced to 0 using custom session ticket callbacks (see SSL_CTX_set_session_ticket_cb(3)).

Should this mean we need to have a `SSL_CTX_set_session_ticket_cb` somewhere as well?

furiel:
Never mind, probably not necessary. If tickets disabled, resumption cannot happen, therefore it does not matter how many tickets we send after resumption.

furiel:
Do we need to take care of loggen too? Inside `open_ssl_connection`  in `loggen_helper.c`?

furiel:
Ok loggen is client only, so not affected.

PR #3083, opened 2020-01-20
http: fix curl ssl version

furiel:
Some `ssl-version` options were not accepted by http destination, even though they should have been:
- tlsv1_0
- tlsv1_1
- tlsv1_2
- tlsv1_3

The reason was that the used compiler switches did not exist. They were an enum instead.

Hence the relevant code was disabled.

The previous behaviour was that syslog-ng starts no matter what was the `ssl-version`, just emit a warning if something unexpected is detected.

This patch fixes:
- The compiler switches
- Syslog-ng will not start if invalid `ssl-version` string was added.

This latter might break some existing configuration: if users accidentally mistyped the option, syslog-ng will not start. I think this is acceptable, considering, this options specifies the minimum version of used ssl version, and thus might have security impact.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1306,6 +1306,15 @@ if test "x$enable_http" != "xno" && test "x$with_libcurl" != "xno"; then
 		AC_MSG_ERROR(libcurl not found)
 	fi
 	enable_http=$libcurl
+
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

PR #3084, opened 2020-01-20
cfg: fix cfg_is_module_available() call

alltilla:
I believe it was intended this way, but please correct me if I am wrong.

### DIFF
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -241,7 +241,7 @@ gboolean
 cfg_is_module_available(GlobalConfig *self, const gchar *module_name)
 {
   cfg_discover_candidate_modules(self);
-  if (!_is_module_autoload_enabled(self))
+  if (_is_module_autoload_enabled(self))
     return cfg_load_module(self, module_name);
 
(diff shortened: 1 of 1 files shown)


alltilla:
@bazsi Can you take look at this please? Am I seeing this wrong?

bazsi:
no. the point of cfg_is_module_available() is to find out weather a module is available _without_ loading it. This works fine if module autoloading is enabled, as that's the only case we have visibility into the availability of modules (due to the automatic discover we do in this case).

On the other hand if module auto-loading is disabled (a rare case, I am wondering if it ever happens), then the only way to determine if a module is available is by attempting to load it.

One side-effect of this is that a module can be loaded multiple times, which triggers ugly warnings in the logs, which I was trying to fix when I introduced this code.

So, yes, the code behaves as intended.

alltilla:
Sorry, then I am closing this PR.

This idea came, because we are failing to load elasticsearch-java in our nightly PE test. The root cause must be something else.

Thanks!

kira-syslogng:
Build SUCCESS

alltilla:
More likely this was the problem: #3085

PR #3085, opened 2020-01-20
elasticsearch-java: fix `requires` in scl

alltilla:
It was `java` instead of `mod-java`. Also the `requires` on the top is not needed now.

### DIFF
--- a/scl/elasticsearch/elastic-java.conf
+++ b/scl/elasticsearch/elastic-java.conf
@@ -21,7 +21,6 @@
 #############################################################################
 
 @requires json-plugin
-@requires mod-java
 
 block destination elasticsearch2(
   index()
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build started
Build FAILURE

kira-syslogng:
Build SUCCESS

gaborznagy:
@bazsi all `requires mod-java` need to be moved inside a block, right?

Just because we forgot `scl/kafka/kafka-java.conf` and `scl/hdfs/plugin.conf`, so syslog-ng-core will depend on java...

bazsi:
Hi,

On Mon, Jan 20, 2020, 16:07 Gábor Nagy <notifications@github.com> wrote:

> @bazsi <https://github.com/bazsi> all requires mod-java need to be moved
> inside a block, right?
>

It will improve error messages, yes.

> Just because we forgot scl/kafka/kafka-java.conf and scl/hdfs/plugin.conf,
> so syslog-ng-core will depend on java...
>
No, requires will only cause the include file to be skipped in case the
module is missing.

PR #3086, opened 2020-01-20
Autodetect system cert store

lbudai:
Change `use-system-cert-store` : this option was added in order to support monolithic installations that ships even libcurl...  But later, as it came out, it's not enough. 
What we really need is to do some autodetection regarding ca-file and ca-dir.
In monolithic installations that ships even libcurl we need to enable this by default in the related SCLs.

@furiel: could you check this?

### DIFF
--- a/modules/http/http-grammar.ym
+++ b/modules/http/http-grammar.ym
@@ -33,6 +33,7 @@
 #include "http.h"
 #include "http-grammar.h"
 #include "response-handler.h"
+#include "autodetect-ca-location.h"
 #include "plugin.h"
 
 #include <string.h>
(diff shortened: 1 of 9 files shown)


kira-syslogng:
Build FAILURE

modules/http/http-grammar.ym
```diff
@@ -172,10 +173,16 @@ http_tls_option
...
     | KW_SSL_VERSION '(' string ')'           { CHECK_ERROR(http_dd_set_ssl_version(last_driver, $3), @3,
                                                             "curl: unsupported SSL version: %s", $3);
                                                 free($3); }
+    | KW_USE_SYSTEM_CERT_STORE '(' yesno ')'  {
+        const gchar *ca_dir = auto_detect_ca_dir();
+        const gchar *ca_file = auto_detect_ca_file();
+        if (!ca_dir && !ca_file) YYERROR;
+        http_dd_set_ca_file(last_driver, ca_file);
+        http_dd_set_ca_dir(last_driver, ca_dir);
+      }
```
MrAnno: I think we should do this only when `use-system-cert-store()` is set to `yes`.

(In general, we should not use system certs by default, because - in my opinion - syslog-ng should not trust "everything on the web", our trust chain should be much narrower. Of course, there are exceptions, like `telegram()`, `slack()`, and #3078, but we can wrap those into an SCL block, e.g: `block destination telegram(use-system-cert-store(yes) ...)`)

lbudai: just wanted to open the pr before go home... so it's -of course- not intentional.

The default will be -of course- *no*.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

modules/http/http-grammar.ym
```diff
@@ -172,10 +173,19 @@ http_tls_option
...
-    | KW_USE_SYSTEM_CERT_STORE '(' yesno ')'  { http_dd_set_use_system_cert_store(last_driver, $3); }
     | KW_SSL_VERSION '(' string ')'           { CHECK_ERROR(http_dd_set_ssl_version(last_driver, $3), @3,
                                                             "curl: unsupported SSL version: %s", $3);
                                                 free($3); }
+    | KW_USE_SYSTEM_CERT_STORE '(' yesno ')'  {
+        if ($3)
+          {
+            const gchar *ca_dir = auto_detect_ca_dir();
+            const gchar *ca_file = auto_detect_ca_file();
+            if (!ca_dir && !ca_file) YYERROR;
```
furiel: Could you use `CHECK_ERROR`? It would be nice to give a reason about the failure.

lbudai: yep, good idea

kira-syslogng:
Build SUCCESS

lbudai:
I guess the reason of test-mongodb-config failure on macOS is the newer libmongoc (they upgraded it from 1.15 to 1.16).

PR #3094, opened 2020-01-27
travis: force python2, add --cache-dir=$PWD/cache

alltilla:
Our python codebase (tests) is based on python2, and it was assumed, that `python` will point to `python2`.

Newer distros (the arm based runner in travis) redirect `python` to `python3`, which was not our intention.

For some reason python3 does not have all the packages we need in pip, so our travis job breaks.

This patch forces `python2` for `pip`.

I also added `--cache-dir=$PWD/cache`, because the arm travis runner does not grant the default cache dir with proper capabilities.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,8 @@ addons:
       - libsnmp-dev
       - ninja-build
 install:
-      - pip install --user --cache-dir=$PWD/pip-cache -r requirements.txt
-      - pip list
+      - python2 -m pip install --user --cache-dir=$PWD/pip-cache -r requirements.txt
+      - python2 -m pip list
 before_script:
   - echo 'Europe/Budapest' | sudo tee /etc/timezone
   - sudo dpkg-reconfigure --frontend noninteractive tzdata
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -95,7 +95,7 @@ matrix:
       compiler: clang
       script:
         - set -e
-        - pip install --user pre-commit-hooks
+        - python2 -m pip install --user --cache-dir=$PWD/cache pre-commit-hooks
```
MrAnno: `--with-python=2` is already hard-coded in travis.yml, so I think it was a good idea to use `python2` explicitly for all jobs. 

Why did you add cache-dir to the `autotools-clang` and `cmake-submodules-and-pytest` jobs?

kira-syslogng:
Build FAILURE

.travis.yml
alltilla: It seemed like a non-destructive move, although it is not necessarily needed. I can remove it, if you insist.

MrAnno: Thanks. I don't, I was just curious.

kira-syslogng:
Build SUCCESS

lbudai:
@alltilla: thanks!

PR #3095, opened 2020-01-27
http: fix auto-detect

furiel:
`CHECK_ERROR` is like assert: we need to pass true. If the predicate is
false, only then will syslog-ng generate error.

The original code breaks 001-check-scl.zts, that tests telegram. This was not executed by kira, because it is not run-local test.

No need to add change note. This is a quick fix for https://github.com/syslog-ng/syslog-ng/pull/3086


kira-syslogng:
Build SUCCESS

PR #3098, opened 2020-01-28
dbld: bionic: install libmongoc from OBS repository

lbudai:
Reason: upstream package installs cmake files with wrong paths.

### DIFF
--- a/dbld/images/helpers/packages.manifest
+++ b/dbld/images/helpers/packages.manifest
@@ -37,7 +37,9 @@ pcre-devel                      [centos, fedora]
 
 # packages to run successful configure with module support
     # libbson-dev
-    libbson-dev                 [debian-stretch, debian-buster, ubuntu-xenial, ubuntu-bionic, ubuntu-cosmic]
+    libbson-dev                 [debian-stretch, debian-buster, ubuntu-xenial, ubuntu-cosmic]
+    libbson-1.0.0=1.15.0-1      [ubuntu-bionic]
+    libbson-dev=1.15.0-1        [ubuntu-bionic]
     libbson-devel               [centos-7, fedora]
 
     # librabbitmq-dev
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
Can we not work this around from our side? Eventually I would like to decrease the dependency on non standard package versions (with the exception of one bleeding edge version, to allow the use of new features in upcoming versions), but otherwise older distro versions should rather not support a feature if the distro dependency is not recent enough.

Just my $.02

bazsi:
And the bleeding edge distro should be the one that we use as a base for devshell.

lbudai:
@bazsi : I think if there is an _LTS_ version of a distro, it's not a sin when someone builds a test/build infrastructure on that instead of the less stable bleeding edge.

### The backround of this PR:
I was told that we have two failing internal tests...  I know that we have very limited resources to upgrade test environments, and so on, so I made a decision that I'll do the quickest fix I can.
(and as one of my development env. is also bionic, and no, I won't upgrade it just because of this issue, I built this package in OBS)

(then later it came out that the internal test environment is not based on bionic, but maybe on xenial... so I don't care if this PR will be merged or not, just did a quick fix to a problem (to a problem that maybe not a problem anymore...) 

@mitzkia , @furiel : I'm waiting a feedback from you as you were reporting this as a bug and I guess you can answer this question

### One more thing
There are some tools where you can't specify what is the base distro... LGTM seem to be such a tool. This note is not against what you are suggesting, it just means that if we really want to use bleeding edge distros in our build/test environments then we should be able to control them. So when there is an issue like this, we can choose between the following options:
* OBS
* upgrade
* change the base distro to a real bleeding egde distro called Arch

furiel:
My opinion is:

- I think ubuntu will not fix this in the near future, maybe ever on bionic, even though the bug was reported for them, and it was fixed in upstream. 
- mongoc works in xenial, and also fixed in upstream -> It will be fixed in newer distributions. I think it is unfortunate that a feature is not supported continuously.
- If we can support a feature on any distribution, then I think that is preferable. Users can still decide if they want to use obs or not. But I am sure there are at least a few users that are happy if they do not need to package their own dependencies, and just use obs easily. 
- OBS is well maintained, with much attention and care, with large user base. I do not see any problem that we also use it in our test infrastructure. Especially if it fixes platform instability problems for us too.

lbudai:
I had a small talk with Bazsi about DBLD, we agreed on that we should discuss this topic IRL, 
so I'm closing this PR now.

PR #3100, opened 2020-01-29
python: persist api

furiel:
This patchset exposes persist api into python code.

This is the continuation of #3016, towards the road of having proper acknowledgement support in python.

Similar to #3016, this PR neither has any usage on its own, until bookmark handling is implemented in the future.

Features:
- A `Persist` class is importable from `syslogng` module
- The object looks like a dictionary: users can store values behind keys. The difference is it is backed by an entry in the persist file, so data is persisted between restarts.
- If no persist file created yet, `defaults` constructor parameter specifies can specify values for keys.
- Handles integers and strings.
- Inheritable
- Iterable

Limitations:
- Only a limited size of data can be added at a time. The length of the key+value must be below 4K, I do not have the exact values for now. As this module is intended for bookmarking, it does not make sense to store large values anyway. To store large number of local data securely, users need to use local database or other mechanism.
The limitation comes from the persist file in c, it can be only increased 4K at a time. This limitation can be workarounded if really needed: by storing only a small portion of the value, and incrementally append the rest of the data in small chunks.

Examples:

The code below prints five numbers, each restart/reload. Continuing from the last number of the previous execution.

```python
from syslogng import Persist, LogSource
class RangeSource(LogSource):
    def init(self, options):
        self.persist = Persist("persist_name", defaults={"bookmark" : 1})
        return True
    def run(self):
        first_item = self.persist["bookmark"]
        for i in range(first_item, first_item + 5):
            print("Item: {}".format(i))
            self.persist["bookmark"] = i + 1
    def request_exit(self):
        pass
```

You can iterate through keys:
```python
persist = Persist("persist_name", defaults={"foo" : "bar"})
for i in persist:
    print(i)
```

Developer notes:
- This patch contains a first example for actually developing core in python (see `defaults` parsing in the last patch.
- Could I have reuse nvtable? That has very similar use case: relocatable hashtable.
I concluded that it would not be a good idea. NVTable contains some features that are not necessary for persist. Also some code must have been added for the realloc to work. Besides, persist structure has a little different requirements. For example there is no defragmentation support in NVTable (due to performance reasons), which is, on the other hand, important when we are working in a limited space on disk. Also, NVTable needs to be fast on adding new items. This is not necessary for persist api. In that case adding new entries is a rare operation.
- A lot of complexity comes from the fact that syslog-ng can relocate a persist entry basically anytime. This means if the hashtable is based on pointers, the relocation can become very hard.
- A new structure SerialHash is introduced to be able to resolve the challenges of this requirement.
The idea is simple. Each key+value will be stored in nodes. These nodes are stored in a new structure called SerialList: which is essentially a doubly linked list. The speciality is that the previous and next members are not pointers, but offsets from a base pointer. From that a relocation is very easy, only the base pointer needs to be updated and the data can be memcopied. This will not corrupt the data structure.
- Defragmentation: there are two node types: DATA type nodes with data, and SPACE nodes for empty available space.
If a node is deleted, it is marked as SPACE. Then we check if the previous or next node is SPACE, and in either cases we merge the space nodes into a common node.
- Adding new elements to SerialList: the first large enough SPACE is searched serially (`O(N)`). This is not a problem, because we do not expect that addition will be a common operation.
- The other operations are speeded up. `SerialHash` encapsulates a `SerialList`, and also a regular `GHashTable` for index table: if we want to find a key-value, we can quickly lookup the index in the hashtable.
- Modification: if the data is smaller or equal in length, then we can reuse the same node -> fast operation due to the index table. If the data is larger, then the data needs to be added to a new free space, and the original node is freed. Again, finding new space is linear search, but we do not expect the data to ever increase -> uncommon operation.
- The Persist entry always stores data as string, no matter if integer or string is specified. Integers are serialized trivially. To differentiate, we also store the type of the entry, so that proper type could be returned during lookup.

### DIFF
--- a/modules/python/CMakeLists.txt
+++ b/modules/python/CMakeLists.txt
@@ -11,6 +11,11 @@ set(PYTHON_SOURCES
     python-module.h
     python-config.h
     python-config.c
+    serial-hash.h
+    serial-hash-internal.h
+    serial-hash.c
+    serial-list.h
+    serial-list.c
     python-persist.h
     python-persist.c
     python-helpers.h
--- a/modules/python/Makefile.am
+++ b/modules/python/Makefile.am
@@ -26,6 +26,11 @@ modules_python_libmod_python_la_SOURCES		= \
 	modules/python/python-module.h		  \
 	modules/python/python-config.h		  \
 	modules/python/python-config.c		  \
+	modules/python/serial-hash.h		  \
+	modules/python/serial-hash-internal.h     \
+	modules/python/serial-hash.c		  \
+	modules/python/serial-list.h		  \
+	modules/python/serial-list.c		  \
 	modules/python/python-persist.h		  \
 	modules/python/python-persist.c		  \
 	modules/python/python-helpers.h		  \
--- a/modules/python/compat/compat-python-v2.c
+++ b/modules/python/compat/compat-python-v2.c
@@ -66,3 +66,9 @@ pyobject_as_int(PyObject *object)
 {
   return PyInt_AsLong(object);
 };
+
+gboolean
+py_object_is_integer(PyObject *object)
+{
+  return PyLong_Check(object) || PyInt_Check(object);
+}
--- a/modules/python/compat/compat-python-v3.c
+++ b/modules/python/compat/compat-python-v3.c
@@ -66,3 +66,9 @@ pyobject_as_int(PyObject *object)
 {
   return PyLong_AsLong(object);
 };
+
+gboolean
+py_object_is_integer(PyObject *object)
+{
+  return PyLong_Check(object);
+}
--- a/modules/python/compat/compat-python.h
+++ b/modules/python/compat/compat-python.h
@@ -47,4 +47,6 @@ void py_init_argv(void);
 PyObject *int_as_pyobject(gint num);
 
 gint pyobject_as_int(PyObject *object);
+gboolean py_object_is_integer(PyObject *object);
+
 #endif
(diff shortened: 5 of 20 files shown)


furiel:
@kira-syslogng retest this please test branch=furiel-persist-api;

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
Force pushed to resolve confict.

furiel:
@kira-syslogng retest this please test branch=furiel-persist-api;

kira-syslogng:
Build SUCCESS

lbudai:
@furiel : I've checked your code and I'd challenge the `SerialList` concept...

If I'm right, `SerialHash` and `SerialList` exist mainly because you are serializing the values linearly with one 'real' persist key (name identifies a multi-value, a list). What if you just use a persist_name as a prefix before local keys and store regular key-values?

example:

```
Persist['counter'] = 12 //counter is the 'local' key
persist_base_name_counter 12
```

In this case the representation would not require a linear space...

(I know that there are additional values of your solution, like types, iter, but I'm not sure if it worth... and I think these can be resolved in another way, so the key question currently is do we really need the complexity of `SerialList` and `SerialHash` to have the persist functionality in Python modules?)

furiel:
Yes, you are right. I wanted users to have a way to get the list of the keys. 

I was thinking about storing data under individual entries, and I agree this could be implemented with a key builder. Let's say users need to provide a `prefix`. And during listing the keys, we could iterate through all persist entries, and return the key if the prefix matches.

I liked with the single entry version is that it is more explicit with the keys: we cannot have false positive keys, nor need to check irrelevant persist entries. On the other hand, the implementation turned out to be more complex than I anticipated.

furiel:
We concluded that I rebuild the feature without the serial hashtable implementation. Instead of storing the subkeys in the same persist entry, they will be stored in the main level of the persist file. That will remove a great deal of implementational complexitiy.

I close this PR temporarily while working on this.

PR #3102, opened 2020-01-30
python: fix test_python_persist_name

furiel:
With python3, `test_python_persist_name` crashes and there is an assertion error.

- Forgot to add gilstate lock in some places.
- Different error type is raised when trying to write readonly member.

news file change not needed. This just a fixup for #3016


kira-syslogng:
Build SUCCESS

PR #3069, opened 2020-01-13
loggen, dqtool: use g_path_get_basename() instead of basename()

alltilla:
Fixes #3065

PR #3066, opened 2020-01-10
news: add news directory and script to create NEWS.md

alltilla:
How to create a newsfile entry
==============================

 1. Create a file in the `news/` directory called `<type>-<pr-num>.md`, where `<type>` is either:
     * `feature`: New functionality.
     * `bugfix`: Fix to a reported bug.
     * `developer-note`: Changes, that are only interesting to developers. (internal API change, etc...)
     * `other`: Other important, but not categorized change.
    You can calculate your PR number, by checking the most recent Issue or PR on github,
    but it could be a good practice to add the commit, containing the news entry, after you
    have opened your PR, as you know your PR number already.
 2. Fill it with the summary of the PR.
     * You can use any markdown formatting.
     * Please make sure, that there are no lines longer than 120 characters.
     * Make it 2-3 sentences at most.
     * Start the entry with the affected modules then a colon. For example: "`network-dest`: Added..."
 3. Commit it with the message: "news: <type>-<pr-num>", for example: "news: bugfix-1234"

How to create the `NEWS.md` file for the release
================================================

 1. Run `news/create-newsfile.py` (`python3` required)
 2. Fill the `Highlights` section in the newly created `NEWS.md` file.
 3. Fill the contributors part of the `Credits` section.
    The internal news file creating tool can generate it for you.

Note, the script uses the `VERSION` file, so it is advised to run the script, after the version is bumped,
or make sure, that you fill the version correctly by yourself.

PR #3107, opened 2020-01-31
missing news files

lbudai:
Question: what about example configs?
Do we want to include them in the news file?
(I've added example configs... sometimes it is useful for users...)

### DIFF
--- a/news/developer-note-3043.md
+++ b/news/developer-note-3043.md
@@ -1,0 +1,31 @@
+`signal-slot-connector`: a generic event handler interface for syslog-ng modules
+
+* The concept is simple:
+
+There is a SignalSlotConnector which stores Signal - Slot connections 
+Signal : Slot = 1 : N, so multiple slots can be assigned to the same Signal.
+When a Signal is emitted, the connected Slots are executed.
+Signals are string literals and each module can define its own signals.
+
+* Interface/protocol between signals and slots
+
+A macro (SIGNAL) can be used for defining signals as string literals:
+
+```
+SIGNAL(module_name, signal, signal_parameter_type)
+```
+
+The parameter type is for expressing a kind of contract between signals and slots.
+
+usage:
+
+```
+   #define signal_cfg_reloaded SIGNAL(config, reloaded, GlobalConfig)
+
+   the generated literal:
+
+   "config::signal_reloaded(GlobalConfig *)"
+```
+ 
+ emit passes the argument to the slots connected to the emitted signal.
+
--- a/news/developer-note-3044.md
+++ b/news/developer-note-3044.md
@@ -1,0 +1,20 @@
+`http`: define HTTP signals
+
+This changeset defines the Signal interface for HTTP - with one signal at this time.
+
+What's in the changeset?
+
+* List ADT (abstract data type for list implementations) added to lib.
+  * Interface:
+    * append
+    * foreach
+    * is_empty
+    * remove_all
+
+Implement the List ADT in http module with `struct curl_slist` for storing the headers.
+
+* HTTP signal(s):
+Currently only one signal is added, `header_request`.
+
+Note, that the license for `http-signals.h` is *LGPL* .
+
--- a/news/feature-3078.md
+++ b/news/feature-3078.md
@@ -1,0 +1,35 @@
+`azure-auth-header`: Azure auth header is a plugin that generates
+authorization header for applications connecting to Azure.
+It can be used as a building block in higher level SCLs.
+Implemented as a `signal-slot` plugin.
+
+<details>
+  <summary>Example config, click to expand!</summary>
+
+```
+@version:3.25
+@include "scl.conf"
+destination d_http {
+  http(
+    url("http://127.0.0.1:8888")
+    method("PUT")
+    user_agent("syslog-ng User Agent")
+    body("${ISODATE} ${MESSAGE}")
+    azure-auth-header(
+      workspace-id("workspace-id")
+      secret("aa1a")
+      method("POST")
+      path("/api/logs")
+      content-type("application/json")
+    )
+  );
+};
+source s_gen {
+  example-msg-generator(num(1) template("Test message\n"));
+};
+log {
+  source(s_gen);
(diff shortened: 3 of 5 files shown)


kira-syslogng:
Build SUCCESS

alltilla:
I like the higher verbosity of these entries. At some places maybe a bit much, but the approach is good IMO.

It is true, that we did not write this much in the news file before, but that can be explained with the fact, that one person did not have the time to explain all the entries this deeply.

@MrAnno, what do you think?

I am interested in others' opinion as well :)

czanik:
I love this new effort! A lot easier to read than trying to hunt together info from commit messages and from the source code :) Thank you!

MrAnno:
I like the idea. It's much easier and more fun to just read the news file, start a docker-image, copy-paste a versioned example config snippet and try the new feature.

If the news file gets too long because of this, we could use collapsible snippets:
<details>

```
@version:3.25
@include "scl.conf"
destination d_http {
  http(
    url("http://127.0.0.1:8888")
    method("PUT")
    user_agent("syslog-ng User Agent")
    body("${ISODATE} ${MESSAGE}")
    azure-auth-header(
      workspace-id("workspace-id")
      secret("aa1a")
      method("POST")
      path("/api/logs")
      content-type("application/json")
    )
  );
};
source s_gen {
  example-msg-generator(num(1) template("Test message\n"));
};
log {
  source(s_gen);
  destination(d_http);
};
```
</details>

lbudai:
@MrAnno : thanks for the hint... this expandable cfg look good (I've modified it : https://github.com/lbudai/syslog-ng/blob/lbudai-news-files-2020-jan/news/feature-3078.md )

MrAnno:
@lbudai Thanks for trying it. :)

I just noticed that I've ruined the snippet, somehow line breaks have disappeared from my comment.. Sorry.

kira-syslogng:
Build SUCCESS

alltilla:
The generated NEWS file looks pretty neat:

3.26.1
=======

## Highlights

<Fill this block manually from the blocks below>

## Features

 * `azure-auth-header`: Azure auth header is a plugin that generates
   authorization header for applications connecting to Azure.
   It can be used as a building block in higher level SCLs.
   Implemented as a `signal-slot` plugin.
   
   <details>
   
   ```
   @version:3.25
   @include "scl.conf"
   destination d_http {
     http(
       url("http://127.0.0.1:8888")
       method("PUT")
       user_agent("syslog-ng User Agent")
       body("${ISODATE} ${MESSAGE}")
       azure-auth-header(
         workspace-id("workspace-id")
         secret("aa1a")
         method("POST")
         path("/api/logs")
         content-type("application/json")
       )
     );
   };
   source s_gen {
     example-msg-generator(num(1) template("Test message\n"));
   };
   log {
     source(s_gen);
     destination(d_http);
   };
   ```
   </details> (#3078)
 * `python`: allow specifying persist name from python.
   
   From now on users can specify a persist name template from python code.
   
   ```
   @staticmethod
   def generate_persist_name(options):
       return options["file_name"]
   ```
   
   Currently, in python destination, persist name is generated from the python class name. This can cause inconvenience if the same python class is planned to be used multiple times. For example if someone writes a PythonFileDestination, then the persist name would be python(PythonFileDestination). If a user wants to track many files, that causes persist collision, that needs to be resolved manually in configuration. After this change, module developers can specify a persist name template, that can be generate persist name uniquely (file name can be in included during generation)... Hence there is no need for users to manually resolve persist name collision. Similar example works with python source and fetcher too.
   
   Persist name from config takes precedence over `generate_persist_name`.
   
   Persist name is exposed through `self.persist_name`. (#3016)

## Bugfixes

 * `loggen, dqtool`: Fixed a crash, when writing error/debug message or relocating qfile. (#3069)

## Notes to developers

 * `NEWS.md`: From now on, for every PR that we want to include in the newsfile,
   we must create the news entry with the PR itself. See `news/README.md`. (#3066)
 * `signal-slot-connector`: a generic event handler interface for syslog-ng modules
   
   * The concept is simple:
   
   There is a SignalSlotConnector which stores Signal - Slot connections 
   Signal : Slot = 1 : N, so multiple slots can be assigned to the same Signal.
   When a Signal is emitted, the connected Slots are executed.
   Signals are string literals and each module can define its own signals.
   
   * Interface/protocol between signals and slots
   
   A macro (SIGNAL) can be used for defining signals as string literals:
   
   ```
   SIGNAL(module_name, signal, signal_parameter_type)
   ```
   
   The parameter type is for expressing a kind of contract between signals and slots.
   
   usage:
   
   ```
      #define signal_cfg_reloaded SIGNAL(config, reloaded, GlobalConfig)
   
      the generated literal:
   
      "config::signal_reloaded(GlobalConfig *)"
   ```
    
    emit passes the argument to the slots connected to the emitted signal. (#3043)
 * `http`: define HTTP signals
   
   This changeset defines the Signal interface for HTTP - with one signal at this time.
   
   What's in the changeset?
   
   * List ADT (abstract data type for list implementations) added to lib.
     * Interface:
       * append
       * foreach
       * is_empty
       * remove_all
   
   Implement the List ADT in http module with `struct curl_slist` for storing the headers.
   
   * HTTP signal(s):
   Currently only one signal is added, `header_request`.
   
   Note, that the license for `http-signals.h` is *LGPL* . (#3044)

## Other changes

 * `http`: `use-system-cert-store` functionality has been changed
   
   This option was added in order to support monolithic installations
   that ships even libcurl... But later, as it came out, it's not enough.
   What we really need is to do some autodetection regarding ca-file and ca-dir.
   In monolithic installations we enable this by default in the related SCLs. (#3086)
 * `afmongodb`: remove the support of deprecated legacy configurations (#3092)

## Credits

syslog-ng is developed as a community project, and as such it relies
on volunteers, to do the work necessarily to produce syslog-ng.

Reporting bugs, testing changes, writing code or simply providing
feedback are all important contributions, so please if you are a user
of syslog-ng, contribute.

We would like to thank the following people for their contribution:

<Fill this by the internal news file creating tool>

PR #3108, opened 2020-02-01
lgtm: install autoconf-archive

lbudai:
When cmake install fails, LGTM falls back to automake and automake build
fails when autoconf-archive is missing (it's not detected
automatically).

Update: I have a reason of merge this over #3109 : currently syslog-ng cannot be built in lgtm environment with automake... and I think we should fix it (and make it possible to use as a fallback when cmake build fails).


kira-syslogng:
Build SUCCESS

PR #3109, opened 2020-02-01
[wip] Lgtm cmake disable afmongodb

lbudai:
Another way of resolving the LGTM failure issue...

 #3108 seems simpler (to me). But... I think it would be nice if we would be able to completely disable unwanted modules when we are building syslog-ng with cmake.
This PR 
 * shows a possible example of completely disable these modules... 
 * is more complex as we tell lgtm how to configure and how to build syslog-ng...
 * as we are customizing configure, we can disable not only the afmongodb module, but also the java
which could speed up an LGTM job (of course in this case JNI related C code also skipped) 

What do you think? :
 * support 'complete' disabling of modules when building with cmake
 * disable java in LGTM
 * which solution(solution... better to say workaround) do you prefer? this one or #3108

### DIFF
--- a/modules/afmongodb/CMakeLists.txt
+++ b/modules/afmongodb/CMakeLists.txt
@@ -1,3 +1,7 @@
+if (ENABLE_MONGODB STREQUAL DISABLE)
+  return()
+endif()
+
 find_package(libmongoc-1.0)
 
 module_switch(ENABLE_MONGODB "Enable mongodb destination driver" libmongoc-1.0_FOUND)
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -2,3 +2,15 @@
 # custom query to check for that
 queries:
   - exclude: cpp/potentially-dangerous-function
+
+extraction:
+  cpp:
+    configure:
+      command:
+        - mkdir build
+        - cd build
+        - cmake ../ -DIVYKIS_SOURCE=system -DJSONC_SOURCE=system -DSYSLOG_NG_ENABLE_DEBUG=ON -DENABLE_JAVA=OFF
+    index:
+      build_command:
+        - cd build
+        - make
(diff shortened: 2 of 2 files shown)


kira-syslogng:
Build SUCCESS

modules/afmongodb/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if (ENABLE_MONGODB STREQUAL DISABLE)
+  return()
+endif()
+
```
MrAnno: I think `module_switch` supports disabling modules with the standard `OFF` syntax.

Based on the geoip2 module:
```
find_package(libmongoc-1.0)
module_switch(ENABLE_MONGODB "Enable mongodb destination driver" libmongoc-1.0_FOUND)

if (NOT ENABLE_MONGODB)
  return ()
endif ()

if (NOT libmongoc-1.0_FOUND)
  message(FATAL_ERROR "The mongodb module is enabled but libmongoc not found.")
endif ()

set(AFMONGODB_HEADERS
...
```

```
$ cmake ../ -DIVYKIS_SOURCE=system -DJSONC_SOURCE=system -DSYSLOG_NG_ENABLE_DEBUG=ON -DENABLE_JAVA=OFF -DENABLE_MONGODB=OFF
```

MrAnno: Oh I see, CMake fails at `/usr/lib/x86_64-linux-gnu/cmake/libbson-1.0/libbson-1.0-config.cmake:28`. It's really unfortunate, because `find_package()` without the REQUIRED options should not really stop the execution.

lbudai: @MrAnno : exactly... and we need to return _before_ findpackage

PR #3114, opened 2020-02-03
Fix multiple-definition issues

### LINKED ISSUE #3101: syslog-ng-3.25.1 fails to build with -fno-common
# syslog-ng
## Version of syslog-ng
3.25.1

## Platform
Linux Gentoo

## Steps to reproduce
Build syslog-ng with CFLAGS="... -fno-common"

## Issue
https://gcc.gnu.org/gcc-10/changes.html:
````
GCC now defaults to -fno-common. As a result, global variable accesses are more efficient on various targets. In C, global variables with multiple tentative definitions now result in linker errors. With -fcommon such definitions are silently merged during linking. 
````

This can be simulated on gcc 9.2.0 with CFLAGS="... -fno-common" too.

Gentoo bug: https://bugs.gentoo.org/707124

### PR DESCRIPTION
**Pull request by MrAnno**
This PR removes the `-fcommon` flag from CMake and autotools builds.

The `-fcommon` option hides (merges) multiple definition issues such as the persist-tool global variable redefinition (bc3fcb4) .

Such multi-definitions are not intentional in the syslog-ng codebase.

> The -fcommon places uninitialized global variables in a common block. This allows the linker to resolve all tentative definitions of the same variable in different compilation units to the same object.

In new GCC versions, `-fno-common` is the default:

>GCC now defaults to -fno-common. As a result, global variable accesses are more efficient on various targets. In C, global variables with multiple tentative definitions now result in linker errors. With -fcommon such definitions are silently merged during linking. 

Fixes #3101

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,8 +331,7 @@ endif()
 add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} -j $$(nproc) --output-on-failure)
 
 set(IMPORTANT_WARNINGS
-  -Wshadow
-  -fcommon)
+  -Wshadow)
 
 set(ACCEPTABLE_WARNINGS
   -Wno-stack-protector
--- a/Makefile.am
+++ b/Makefile.am
@@ -54,8 +54,7 @@ AM_CPPFLAGS		= -I$(top_srcdir)/lib -I$(top_srcdir)/modules -I$(top_builddir)/lib
 
 # Important warnings
 AM_CFLAGS = \
-	-Wshadow \
-	-fcommon
+	-Wshadow
 
 # Acceptable warnings
 AM_CFLAGS += \
--- a/lib/logmsg/tests/test_log_message.c
+++ b/lib/logmsg/tests/test_log_message.c
@@ -32,6 +32,8 @@
 #include <stdlib.h>
 #include <glib/gprintf.h>
 
+MsgFormatOptions parse_options;
+
 typedef struct _LogMessageTestParams
 {
   LogMessage *message;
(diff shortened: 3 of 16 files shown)

### CODE REVIEW

**Comment by kira-syslogng** [comment]
Build FAILURE

**Comment by kira-syslogng** [comment]
Build SUCCESS

File: lib/filter/tests/test_filters_common.c [hunk 1]
```diff
@@ -37,15 +37,14 @@
...
+#include "libtest/msg_parse_lib.h"
 #include <criterion/criterion.h>
 
 #include <time.h>
 #include <string.h>
 #include <stdlib.h>
 #include <stdio.h>
 
-MsgFormatOptions parse_options;
-
```
lbudai: Wouldn't be better to keep away us from libtest?

(I mean you wouldn't get this multiple definition issue if you would not add `LIBTEST` to the `add_unit_test` calls.
So the commit message is about resolving an issue which is caused by your change within the very same patch.
I guess that the reason of having `parse_options` definition in `test_filters_common.c` is that we didn't want to depend on `libtest`. But I also think that you are know this... so maybe there is something _I_ do not know... Could you add a more detailed explanation on why are you doing this? :) )

MrAnno: The original problem was with the autotools build, where criterion and libtest are not switchable, both are linked unconditionally. I'll fix this correctly along with the failing macOS job.

MrAnno: Thanks for the note. I've removed the `parse_options` global from the msg-parse lib. I think it's better this way, because we had a lot of unit tests where `parse_options` was used without any sign of declaration, definition, and initialization.

Removing this global revealed that `init_and_load_syslogformat_module()` actually means `init_parse_options_and_load_syslogformat()`, which is a strange combination. This could be split and refactored further.

kira-syslogng:
Build SUCCESS

PR #3115, opened 2020-02-04
Add new rewrite function: set-severity()

smortex:
This is a follow-up to #2602 by @Kokan 

I rebased the code on top of master, fixed the build, and renamed the function `set-severity` as requested in https://github.com/syslog-ng/syslog-ng/pull/2602#issuecomment-471241895

A few more points where under discussion, I will add in-line notes to discuss the topics further.

This PR also fixes `cisco-parser` $PRIORITY macro parsing.

----

Usage: `set-severtity(TEMPLATE)`

Accepted values are numeric strings `[0-7]` or named values (`emerg, emergency, panic, alert, crit, critical, err, error, warning, warn, notice, info, informational, debug`).

Examples:
```
rewrite {
  set-severity("info");
  set-severity("6");
  set-severity("${.json.severity}");
};
```


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

lib/rewrite/rewrite-set-severity.c
```diff
@@ -0,0 +1,138 @@
...
+  LogRewriteSetLevel *self = (LogRewriteSetLevel *) s;
+
+  log_msg_make_writable(pmsg, path_options);
+
+  log_template_format(self->severity, *pmsg, NULL, LTZ_LOCAL, 0, NULL, result);
+
+  const gint severity = _convert_severity(result);
+  if (severity < 0)
+    {
+      msg_warning("Warning: invalid severity to set", evt_tag_str("severity", result->str), log_pipe_location_tag(&s->super));
```
smortex: @bazsi suggested changing this to a debug message.  I have mixed feelings: this message can only result of the user using set-severity in his code, and using an improper value for the function. A warning seems fine here for me because the user has done something wrong and syslog-ng will not do what he expected, but maybe I am missing something?

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/rewrite/rewrite-expr-parser.c
```diff
@@ -36,6 +36,7 @@ static CfgLexerKeyword rewrite_expr_keywords[] =
   { "subst",              KW_SUBST },
   { "set_tag",            KW_SET_TAG },
   { "clear_tag",          KW_CLEAR_TAG },
+  { "set_severity",       KW_SET_SEVERITY },
```
smortex: I am considering adding `set_level` and `set_priority` to this list as alternatives to `set_severity`: while the RFC use *severity*, the syslog-ng doc use *level* and *priority* a lot, and those are the words I was using when I found the original PR.  What do you think?

smortex:
I fixed a bunch of refactoring mistakes yesterday, but kira is still failing. May someone with access to kira logs help by providing some insights about what is going wrong?  Thanks!

Kokan:
@smortex Only one test failed, and that failure is actually test case issue. (Cisco parser tests actually asserts on wrong priority value.) I'll push the fix tomorrow, and rerun kira.

Also I plan to do a review, regarding the rename part. I'll probably have some not one backward compat.

Kokan:
@kira-syslogng retest this please; branch=kokan-github-pr-3115

kira-syslogng:
Build FAILURE

Kokan:
@kira-syslogng retest this please; branch=kokan-github-pr-3115

kira-syslogng:
Build FAILURE

Kokan:
@kira-syslogng retest this please branch=kokan-github-pr-3115;
third time must be right :)

kira-syslogng:
Build SUCCESS

lib/rewrite/rewrite-expr-parser.c
bazsi: the problem is that "priority" in RFC parlance is the combination of "severity" and "facility". The syslog implementation in libc used the term "priority" as a combination of "facility" and "level":

>        The priority argument is formed by ORing together a facility value and a level value (described below).  If no facility value is ORed into priority, then the default value set by openlog() is used, or, if there was no preceding openlog() call, a default of LOG_USER is employed.

Unfortunately, syslog-ng was not very consistent with this naming, for instance we have a filter expression called priority() that is the same as level().

I would try to push for more consistency.

So I would opt for using the term "severity" with "level" as an obsolete alias, any inconsistent use should be obsoleted too (like the priority() filter expression). And if level is obsolete, I wouldn't add set-level().

I can be convinced that this is too pushy, in that case we should be adding level() aliases everywhere and not complain about them at all. (e.g. don't obsolete that syntax either). The reason I would be ok with this idea is that a _lot_ of users already have that syntax in their configuration and changing those config files now is a lot of work we would be pushing to the world. An alias here and there is not a maintenance nightmare. If this is what we decide, we should indeed add set-level(), but not set-priority().

My $.02

lib/rewrite/rewrite-set-severity.c
bazsi: The problem is that this message would be displayed for every incoming message, generating a lot of extra noise by syslog-ng.

I would be ok with this if we used some kind of suppression mechanism (e.g. msg_warning_once_every_minute()) per message errors are usually making the problems worse.

bazsi:
I was kind of hoping to see the related changes to use the severity term in other similar cases, e.g.

 * filter expressions: level(), and priority()
 * macros $LEVEL, $LEVEL_NUM, $PRIORITY
 * source option: default-level() => default-severity(), default-priority() => deprecate

lib/rewrite/rewrite-expr-parser.c
smortex: Thanks for the details, I understand better why those words have blur meanings in my head :smile:!

As far as I am concerned, I am in favor of improving consistency and deprecating the confusing words, therefore not adding aliases.  But I guess this is a decision that is up to the project so if you decide to go the other route, I'll add the aliases.

lib/rewrite/rewrite-set-severity.c
smortex: I understand your point.  I added a commit that change the logging to the debug leve^Wseverity.

kira-syslogng:
Build FAILURE

smortex:
D'oh, I'm afraid the last change related to changing the logging severity of the message when a logging severity cannot be parsed broke your fix @Kokan :scream:

Kokan:
@smortex now it's fine, just that branch= magic for kira is not sticky. So when you pushed it executed from the master branch again. If you want to I can trigger the job for you, and potentially you could as well with the magic words.

Kokan:
@kira-syslogng retest this please branch=kokan-github-pr-3115;

kira-syslogng:
Build SUCCESS

lib/rewrite/rewrite-expr-parser.c
bazsi: @pzoleex can you help us decide here, please?

scl/cisco/plugin.conf
```diff
@@ -72,6 +72,7 @@ block parser cisco-triplet-parser(template() prefix()) {
             };
             rewrite { set("${`prefix`facility}-$1" value('`prefix`facility')); };
         };
+	rewrite { set-severity("${`prefix`severity}"); };
```
bazsi: I think this should go to cisco-parser() and not here, the triplet parser is a low level building block that should only change name-value pairs that it is supposed to extract into. It is ok to do this at the cisco-parser() level, but I don't think it would be ok here.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

smortex:
@kira-syslogng retest this please branch=kokan-github-pr-3115

scl/cisco/plugin.conf
smortex: Fixed!

smortex:
@Kokan It looks like I cannot trigger a build (I try to cast the spell by copy/pasting the magic words of your incantation without success) :-)  May I ask you to trigger the build?  Thanks!

bazsi:
@kira-syslogng retest this please branch=kokan-github-pr-3115

kira-syslogng:
Build FAILURE

Kokan:
Oh this is actually good that it failed, because the incantation was incorrect, so it used the current version from master that must fail.
The missing ingredient is the semi-colon at the end of branch=...;

@kira-syslogng retest this please branch=kokan-github-pr-3115;

kira-syslogng:
Build FAILURE

Kokan:
@kira-syslogng retest this please branch=kokan-github-pr-3115;

kira-syslogng:
Build SUCCESS

lib/rewrite/tests/test_set_severity.c
```diff
@@ -0,0 +1,136 @@
...
+
+  LogPathOptions path_options = LOG_PATH_OPTIONS_INIT;
+  LogMessage *msg = log_msg_new_empty();
+
+  LogRewrite *set_severity = log_rewrite_set_severity_new(_create_template("random-text"), cfg);
+
+  log_pipe_init(&set_severity->super);
+  log_pipe_queue(&set_severity->super, msg, &path_options);
+
+  assert_grabbed_log_contains("invalid severity to set");
```
gaborznagy: What do you think about showing the behavior in case the rewrite rule failed to set the severity (additionally to the error message) ?
Either check explicitly to the expected default value (LOG_NOTICE) or get the current severity before `log_pipe_queue` is called and then check that it isn't changed?

Example about the former:
```
+++ b/lib/rewrite/tests/test_set_severity.c
@@ -85,12 +85,14 @@ Test(set_severity, large_number)
 
   LogPathOptions path_options = LOG_PATH_OPTIONS_INIT;
+  cr_assert_eq(msg->pri & LOG_PRIMASK, LOG_NOTICE);
 
   LogRewrite *set_severity = log_rewrite_set_severity_new(_create_template("8"), cfg);
 
   log_pipe_init(&set_severity->super);
   log_pipe_queue(&set_severity->super, msg, &path_options);
 
+  cr_assert_eq(msg->pri & LOG_PRIMASK, LOG_NOTICE);
   assert_grabbed_log_contains("invalid severity to set");
```

lib/rewrite/rewrite-set-severity.c
```diff
@@ -0,0 +1,139 @@
...
+      goto error;
+    }
+
+  _set_msg_severity(*pmsg, _convert_severity(result));
```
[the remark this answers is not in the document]
smortex: Fixed!

lib/rewrite/tests/test_set_severity.c
smortex: Makes sense!

kira-syslogng:
Build FAILURE

smortex:
@kira-syslogng retest this please branch=kokan-github-pr-3115;

smortex:
I just realized that the logs I am going to parse have _CRITICAL_ as severity in the messages (Java application FWIW) and the _full_ name of some seventies where missing.  I added them because these are the word that appear in section 6.2.1 of RFC 5424, so a built-in support may be more suited than having to rely on rewrite rules?

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please branch=kokan-github-pr-3115;

kira-syslogng:
Build SUCCESS

gaborznagy:
I've added a note in the desc about the bugfix in cisco-parser, so on next release it will help the NEWS file editing.

lib/rewrite/rewrite-set-severity.c
```diff
@@ -0,0 +1,139 @@
...
+{
+  return syslog_name_lookup_level_by_name(severity_text->str);
+}
+
+static gint
+_convert_severity(GString *severity_text)
+{
+  gint severity = _convert_severity_as_number(severity_text);
+  if (severity > 0)
+    return severity;
```
bazsi: just noticed that a valid severity value is 0, so the condition should be ">=" instead of ">"

lib/rewrite/rewrite-set-severity.c
```diff
@@ -0,0 +1,139 @@
...
+
+  severity = _convert_severity_as_text(severity_text);
+  if (severity > 0)
+    return severity;
```
bazsi: likewise

lib/rewrite/rewrite-set-severity.c
```diff
@@ -0,0 +1,139 @@
...
+  LogRewrite super;
+  LogTemplate *severity;
+};
+
+static gint
+_convert_severity_as_number(GString *severity_text)
+{
+  char *endptr;
+  long int severity = strtol(severity_text->str, &endptr, 10);
+  if (0 == endptr)
```
bazsi: we have a scan_positive_int() function that is used by the date scanning routines and if I remember their reason for existance over strtol() was performance. I am not 100% sure that this is still valid, but I think we should use that function for this use-case, as this can be performance sensitive.

bazsi: **please disregard** this comment, scan_positive_int() is pretty awkward to use in this scenario, we also seem to have a parse-number module which is used by numeric template functions and which uses strtol().

Using parse_dec_number() is thus not a performance improvement, only a slight improvement in readability. I am working on adding set-facility() (someone asked for that recently) and I'll update this to use the same mechanism for parsing a number.

smortex: :+1:

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please branch=kokan-github-pr-3115;

kira-syslogng:
Build SUCCESS

lib/rewrite/rewrite-set-severity.c
```diff
@@ -0,0 +1,139 @@
...
+    return -1;
+
+  if (endptr[0] != '\0')
+    return -1;
+
+  if (severity > 7)
+    return -1;
```
MrAnno: This is tricky because the function allows negative severity values.

Fortunately,  a negative return value is handled as an error in `_convert_severity()`, so it works as expected.

lib/rewrite/rewrite-set-severity.c
```diff
@@ -0,0 +1,139 @@
...
+static gint
+_convert_severity_as_text(GString *severity_text)
+{
+  return syslog_name_lookup_level_by_name(severity_text->str);
+}
+
+static gint
+_convert_severity(GString *severity_text)
+{
+  gint severity = _convert_severity_as_number(severity_text);
```
MrAnno: Templates evaluated to an empty string will set the highest `emergency` severity by mistake.
manpage:
> If there were no digits at all,  strtol()  stores the original value of nptr in *endptr (and returns 0).

You have a check for `(endptr[0] != '\0')`, but unfortunately, the original value is also `\0` in this case.

I've added the following test case to check this behavior:
<details>
  <summary>Click to expand!</summary>

```
Test(set_severity, test_set_severity_with_template_evaluating_to_empty_string)
{
  debug_flag = 1;

  LogPathOptions path_options = LOG_PATH_OPTIONS_INIT;

  int default_severity = msg->pri & LOG_PRIMASK;

  LogRewrite *set_severity = log_rewrite_set_severity_new(_create_template("${nonexistentvalue}"), cfg);

  log_pipe_init(&set_severity->super);
  log_msg_ref(msg);
  log_pipe_queue(&set_severity->super, msg, &path_options);

  cr_assert_eq(msg->pri & LOG_PRIMASK, default_severity, "empty templates should not change the original severity");
  assert_grabbed_log_contains("invalid severity to set");

  log_msg_unref(msg);
  log_pipe_deinit(&set_severity->super);
  log_pipe_unref(&set_severity->super);
}
```
</details>

lib/rewrite/tests/test_set_severity.c
```diff
@@ -0,0 +1,139 @@
...
+  LogMessage *msg = log_msg_new_empty();
+
+  int default_severity = msg->pri & LOG_PRIMASK;
+
+  LogRewrite *set_severity = log_rewrite_set_severity_new(_create_template("8"), cfg);
+
+  log_pipe_init(&set_severity->super);
+  log_pipe_queue(&set_severity->super, msg, &path_options);
+
+  cr_assert_eq(msg->pri & LOG_PRIMASK, default_severity);
```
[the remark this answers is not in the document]
smortex: Fixed!

lib/rewrite/rewrite-set-severity.c
smortex: Good catch!  Let's compare the addresses then…

kira-syslogng:
Build FAILURE

MrAnno:
@kira-syslogng retest this please branch=kokan-github-pr-3115;

lib/rewrite/rewrite-set-severity.c
MrAnno: Thanks!

lib/rewrite/tests/test_set_severity.c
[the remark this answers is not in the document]
MrAnno: Thanks. Unfortunately, `cr_assert_eq(msg->pri & LOG_PRIMASK, default_severity);` still reads the freed value, the `log_msg_unref()` function needs to be called after the assert.

kira-syslogng:
Build SUCCESS

lib/rewrite/tests/test_set_severity.c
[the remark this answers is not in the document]
smortex: :man_facepalming: Fixing that right now…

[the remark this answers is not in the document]
smortex: I amended the commit to call `log_msg_unref(msg)` at the correct location.  Thanks!

kira-syslogng:
Build FAILURE

lib/rewrite/tests/test_set_severity.c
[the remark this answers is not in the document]
MrAnno: Thank you. :)

gaborznagy:
@kira-syslogng retest this please branch=kokan-github-pr-3115;

kira-syslogng:
Build SUCCESS

PR #3117, opened 2020-02-04
dbld: freeze last version that supports python2

furiel:
With newer versions of hy, 008-python-destination-loaders.zts
fails, because from 0.18 python3 is not supported by hy.

If syslog-ng is compiled with `--with-python=3` or
`-DPYTHON_VERSION=3`, then hy==0.18 works.

But for now we are compiling with python2, so 0.17 needs until python2
support is dropped.

No need to add this to changenote.


kira-syslogng:
Build SUCCESS

PR #3103, opened 2020-01-30
http: batching mode: format headers only after body is finished

lbudai:
Othwerwise body-dependent headers (like azure-auth-header) will compute
bad values (bad signature in case of azure-auth-header).

Signed-off-by: Laszlo Budai <laszlo.budai@outlook.com>

PR #3120, opened 2020-02-06
http: fix signal slot headers with non-batching but prefix-suffix case

furiel:
In case batching is not enabled, but prefix and suffix is used, http-worker (`_insert_single`) calculated headers before the body is finshed (body-suffix is added).

After this patch, the msg specific headers are added in `_insert_single`, and everything else during flush, unconditionally.

No need for changenote, just a small addition to #3103.

### DIFF
--- a/modules/http/http-worker.c
+++ b/modules/http/http-worker.c
@@ -235,12 +235,9 @@ _add_common_headers(HTTPDestinationWorker *self)
 }
 
 static void
-_format_request_headers(HTTPDestinationWorker *self, LogMessage *msg)
+_format_request_headers(HTTPDestinationWorker *self)
 {
   HTTPDestinationDriver *owner = (HTTPDestinationDriver *) self->super.owner;
-
-  if (msg)
-    _add_msg_specific_headers(self, msg);
 
   _add_common_headers(self);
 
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

modules/http/tests/test_http-signal_slot.c
```diff
@@ -0,0 +1,153 @@
...
+  SignalSlotConnector *ssc = driver->super.super.super.super.signal_slot_connector;
+
+  const gchar *test_msg = "test message";
+  CONNECT(ssc, signal_http_header_request, _check, test_msg);
+
+  cr_assert(log_pipe_init((LogPipe *)driver));
+
+  _generate_message(driver, test_msg);
+
+  wait_for_signal_slot_to_finish();
```
lbudai: @furiel : signal-slot mechanism is _synchronous_ in syslog-ng.

modules/http/http-worker.c
```diff
@@ -605,9 +602,7 @@ _flush(LogThreadedDestWorker *s, LogThreadedFlushMode mode)
     return LTR_RETRY;
 
   _finish_request_body(self);
-
-  if (list_is_empty(self->request_headers))
```
lbudai: why don't we need this check?

modules/http/tests/test_http-signal_slot.c
lbudai: as @furiel  pointed out: we really start another thread...
I've suggested to use `cond_wait` instead : wait in the caller, and cond_signal in the slot... 
But it's up to you @furiel  :)

furiel: You are right. cond_wait is better. I will fix this next time.

Issue #3118, opened 2020-02-04
Forward integrity and confidentiality of logs


nbsd:
## Description of the problem

Logs processed by _syslog-ng_ are neither confidential nor integrity protected. The use of these logs for the purpose of security monitoring and forensic analysis in case of an incident is rather limited, as they may have been tampered with by an attacker trying to hide his trace.

## Proposed solution

An implementation providing forward integrity and confidentiality of logs using _syslog-ng_ was recently presented at the [FOSDEM 2020](https://fosdem.org/2020/schedule/event/security_secure_logging_with_syslog_ng/). This implementation should be merged into the _syslog-ng_ source tree so that all users of syslog-ng may add forward integrity and confidentiality to their logs in case this is needed or required.

## Alternatives

N/A

## Additional context

The implementation of the secure logging functionality providing forward integrity and confidentiality was branched off _syslog-ng_ at version 3.23.1. The development repository continued to track the master branch of _syslog-ng_. Tests have shown only one possible merge conflict in the RPM spec file. This is not relevant to the core functionality of _syslog-ng_ and might be resolved manually.

furiel:
Fixed in #3121

PR #3121, opened 2020-02-07
Forward integrity and confidentiality of logs

nbsd:
Logs processed by syslog-ng are neither confidential nor integrity protected. The use of these logs for the purpose of security monitoring and forensic analysis in case of an incident is rather limited, as they may have been tampered with by an attacker trying to hide his trace.

The implementation providing forward integrity and confidentiality of logs with syslog-ng that was presented recently at the [FOSDEM 2020](https://www.fosdem.org/2020/schedule/event/security_secure_logging_with_syslog_ng) is contained in this pull request. It provides users of syslog-ng with forward integrity and confidentiality of their logs in case this is needed or required.

This pull request solves issue #3118.


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

lbudai:
@kira-syslogng: ok to test

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

MrAnno:
Sorry for the over-complicated copyright check.
With your current license of choice, the policy file should look like this:

```diff
  LGPLv2.1+_SSL,non-balabit
 lib/timeutils/zonecache.[ch]$
+lib/slog\.[ch]$
+modules/cryptofuncs/slogimport/slogimport\.c$
+modules/cryptofuncs/slogkey/slogkey\.c$
+modules/cryptofuncs/slogverify/slogverify\.c$
  LGPLv2.1+_SSL
 autogen\.sh$
 sub-configure\.sh$
@@ -74,10 +78,7 @@ modules/timestamp/rewrite-.*$
 modules/add-contextual-data/add-contextual-data-glob-selector\.[ch]$
 modules/add-contextual-data/tests/test_glob_selector\.c$
 modules/affile/tests/test_file_writer\.c$
-lib/slog\.[ch]$
-modules/cryptofuncs/slogkey/slogimport\.c$
-modules/cryptofuncs/slogkey/slogkey\.c$
-modules/cryptofuncs/slogkey/slogverify\.c$
```

The syntax is something like an "inverse .ini"

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/slog.h
```diff
@@ -24,6 +24,25 @@
 #ifndef SLOG_H_INCLUDED
 #define SLOG_H_INCLUDED 1
 
+#ifdef __APPLE__
+#include <libkern/OSByteOrder.h>
+
+#define htobe16(x) OSSwapHostToBigInt16(x)
```
Kokan: In order to hide missing implementation on platform, or different type of implementation on platforms, there is a so called compat layer in syslog-ng source code. Actually the `compat.h` has a nice description about how to use it: https://github.com/syslog-ng/syslog-ng/blob/3d271976f470832e573528a9eb9c7ca643f9fa67/lib/compat/compat.h#L25-L68

The following defines imho would be better of in a compat layer of their own.

Kokan: BTW could you please explain where exactly are we using these defines ?

nbsd: We have two problems with our code on the MacOS platform.

1. Endianess: On Linux we use le64toh to ensure correct endianess when reading and writing key files. These functions do not exist on MacOS
2. For log verification, we use the hsearch API defined in search.h. However, our implementation currently uses the GNU extensions that also do not exist on MacOS

How would we resolve this using the compatibility mechanism of syslog-ng?

Kokan: Generally the glib is used to handle platform specific differences.
1. Okay, so the disk buffer also has issue with different endianess (you could check that out). There are macros in `glib` that could be useful for you as well: https://developer.gnome.org/glib/stable/glib-Byte-Order-Macros.html
2. Would it be acceptable to replace `hsearch` API with glib hashtable ? That should resolve this issue as well.

kira-syslogng:
Build FAILURE

lib/slog.h
nbsd: Thanks for the hints. I will certainly try the glib macros for endianess. Converting to the glib hashtable API is more of an effort. I will look into it when I find time to do so.

kira-syslogng:
Build SUCCESS

lib/slog.h
nbsd: All checks now pass after I have committed the lasted changes. May I request a review from the core team now?

lbudai:
@nbsd: it's nice to see that your PR passed on CI

Could you move all the `slog` related sources to a separate module?
(It could speed up the merge process as it can be handled as a self-contained experimental module).

nbsd:
The secure logging functionality has been moved to a separate module called "secure-logging".

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

Kokan:
@nbsd could you please also move the `slog.h`/`slog.c` file from lib to the new module ?

nbsd:
This would break the secure logging command line utilities for key management and log verification. I don't see where I would put the library functions contained in slog.[ch] such that both the module and the command line utilities will build without problems and are able to resolve the library functions at runtime.

Kokan:
@nbsd If you move every functionality (`slog.[c]h` and `$(slog)`) into the current secure-logging module (`libsecure-logging.so`) in that case:
* your module has everything it needs
* the command line tools could link to the module and obtain the needed functionality

Note1: still both of those also have to link to syslog-ng.
Note2: A set of patch just doing the above with cmake: https://gist.github.com/Kokan/a4ac2103d156d44a3ea7e866b08a4474

nbsd:
Thanks for providing the patches. However, trying to apply them results in the following errors:
```
user@host~> git apply --check 0001-secure-logging-fix-cmake-compile.patch
error: git diff header lacks filename information (line 40)
```

```
user@host~> git apply --check 0002-secure-logging-move-slog-under-modules.patch
error: corrupt patch at line 89
```
I have tried the patches against the current version of the forked repository.

Kokan:
Not sure if git could help you, but with `patch` they should apply: `patch -p1 < [file]` on top of your current branch.

nbsd:
After applying the patches and also changing the Makefile.am within the module, I get the following warning:
```
*** Warning: Linking the executable modules/secure-logging/slogverify/slogverify against the loadable module
*** libsecure-logging.so is not portable!
```
However, the syslog-ng package builds and installs correctly and the secure logging commandline tools do work.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

nbsd:
The package including the secure logging functionality build correctly with gcc and clang after having created an extra module for it. However,  ```make distcheck``` fails because of the ```slog.h``` not being found. This is not the case when building the package on various test machines. Please explain the difference between the build environments that may cause this problem.

modules/secure-logging/Makefile.am
```diff
@@ -0,0 +1,30 @@
...
+
+EXTRA_DIST += modules/secure-logging/CMakeLists.txt
+
+modules_secure_logging_libsecure_logging_la_CFLAGS	=	\
+	-I$(top_srcdir)/modules/secure-logging	                \
+	-I$(top_builddir)/modules/secure-logging
+
+modules_secure_logging_libsecure_logging_la_SOURCES		=	\
+	modules/secure-logging/secure-logging.c                         \
+	modules/secure-logging/slog.c
```
Kokan: You have to include the header files as  well here, so the automake/autotools could track those files and include into the source tarball (this is the reason why the first job fails).

Kokan:
@nbsd 
Regarding the `*** Warning: Linking the executable modules/secure-logging/slogverify/slogverify against the loadable module`, this is not my realm, hope somebody could give you proper guide on that. I only know that in case of such case (`dqtool`, `pdbtool` an intermediate library is provided, and this warning is not produced).

I only could read that this might cause issue on Mac OS (but the build seems fine on that).

The `B=distcheck` build aim is to check if the generated disk tarball is sound and a build can be made from it. First it creates a tarball, and builds syslog-ng from that.

nbsd:
Thanks for the hint. The ```modules/secure-logging/Makefile.am``` was changed accordingly. Concerning the warnings when build the library, I will do some tests on my Mac and try to identifiy potential issues on the MacOS platform.

kira-syslogng:
Build SUCCESS

nbsd:
Travis CI ```B=discheck``` still fails and produces the following message:

```error: -Wshadow/-Werror did not propagate via CFLAGS in 4 cases:```

I cannot reproduce this behavior on any of the test machines I have access to. When I execute the scripts manually on these machines everything builds correctly without any error message. 

Please advise on how to rectify this problem.

File: modules/secure-logging/Makefile.am [hunk 3]
```diff
@@ -0,0 +1,31 @@
+module_LTLIBRARIES					+=	\
+	modules/secure-logging/libsecure-logging.la
+
+EXTRA_DIST += modules/secure-logging/CMakeLists.txt
+
+modules_secure_logging_libsecure_logging_la_CFLAGS	=	\
+	-I$(top_srcdir)/modules/secure-logging	                \
```
Kokan: You are missing a `$(AM_CFLAGS)` here.

nbsd: Thank you very much. I would not have found this by myself, as everything build correctly when executling the ```make distcheck``` target manually on my build system. I corrected this problem with the latest commit to the forked repository and hope that the secure logging functionality will finally be ready for review.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
@nbsd : I've rebased this against current upstream (and resolved another conflict) : https://github.com/lbudai/syslog-ng/tree/rebased-secure-logging

Please, try to reset your branch to the rebased one, or rebase this branch against current upstream.
Until that we cannot merge the PR.

lbudai:
(sorry for the close, my mouse sometimes doing strange things...)

kira-syslogng:
Build SUCCESS

SHA3-d1b34f498fc27200f25d8a8e080bc22a0b:
Ok! :thumbsup: ...@nbsd is currently traveling and will be back next week.  :smiley:

nbsd:
> @nbsd : I've rebased this against current upstream (and resolved another conflict) : https://github.com/lbudai/syslog-ng/tree/rebased-secure-logging
> 
> Please, try to reset your branch to the rebased one, or rebase this branch against current upstream.
> Until that we cannot merge the PR.

@lbudai Thanks for having prepared the integration of the secure logging functionality. I have reset my master branch accordingly.

furiel:
@nbsd I would like to try out this feature, but unfortunately got stuck. Could you extend the description with a step-by-step guide, what needs to be done? Here is what I tried, but I must be off track.

Trying only the destination part:
```
@version: 3.25

log {
    destination { file("/dev/stdout"
                  template("$(slog -k /tmp/secure-logging/masterkey
                            -m /tmp/secure-logging/cmac.pem $MSG)\n")); };
};
```

First I generated `/tmp/secure-logging/masterkey` with the provided slogkey binary:

```
./slogkey -m /tmp/secure-logging/masterkey
```

Then I need a mac file to put into the template function. What do you mean by mac file? I went with the assumption that it needs to be generated too, with `slogkey -d`.

But `slogkey -d` needs a mac address and a serial number. What do you mean by mac address and serial number? I was thinking that the intention here is probably network transport, and most probably over tls: so the mac address is a network interface mac address, and serial number is a cert serial number. Am I right? If so, are these necessary? I just want to write the output into a file.

So I generated self signed certificates:
```
openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout /tmp/secure-logging/key.pem -out /tmp/secure-logging/cert.pem
```

Then I could dump the serial:
```
openssl x509 -noout -serial -in cert.pem
```

After that, I could generate:

```
./slogkey -d /tmp/secure-logging/masterkey xx:<redacted>:xx CAE98DA3992EDF75 /tmp/secure-logging/cmac.pem
```

After starting syslog-ng, I get errors.
```
[2020-03-17T14:35:34.609659] [SLOG] ERROR: MAC computation invalid;
[2020-03-17T14:35:34.609680] [SLOG] WARNING: Aggregated MAC not found or invalid; File='/tmp/secure-logging/cmac.pem'
```

Other than that, it writes similar output as I saw in your video:
```
[2020-03-17T14:35:34.610319] syslog-ng starting up; version='3.25.1.436.g5c30873'
AAAAAAAAAAA=:Lg7YuwZuicWnixt88jYAb0RY7kWW3vMU1905zlWQd/2BGgoE80nMuSZECZghEajabjW2cw==
AQAAAAAAAAA=:Gue45YKfWf69GQuslzT/+mCUYd8w5hoMwz1uK9bPY606FrW9u2azBecJ4DV/NIxMsWsSTw==
AgAAAAAAAAA=:U7kbc9bAZbUq5T6ZlYf9cthsqk5yHYmIFlmZkh/TnJE7iokEYtR4Bj7rlvFUmWQKhLmElQ==
```

But, if I stop syslog-ng then start again. The error message disappears:
```
[2020-03-17T14:35:38.173043] SLOG: Key successfully loaded;
[2020-03-17T14:35:38.173085] [SLOG] INFO: MAC successfully loaded;
[2020-03-17T14:35:38.173095] [SLOG] $(slog) template with key and MAC file successfully initialized.;
```

Even though I did not change anything, the problem is fixed. This is a bit strange for me.

---------

For me a little confusing that the option names are not consistent. Master key is generated with `-m` but read by `-k` in `slog`. On the other hand, there is also a `-m` option for `slog`, which is also written by the template. I lost my master key file this way.

The last parameter of `slogkey -d` also an output parameter. I lost my key to that also. Maybe it would worth adding it as `-o` parameter.

There is a linking problem with the slogkey binary. With cmake, I could not make it work it does not find `libsecure-logging.so`. With autotools, I have the same problem if I start it with absolute path. On the other hand, with relative to the bin directory, it worked for me. This can be fixed later incrementally though, just a quick feedback.

kira-syslogng:
Build SUCCESS

nbsd:
@furiel You were probably confused by the fast that **MAC** in case of secure logging means **Message Authentication Code** and not _Medium Access Layer_ as in network technology. You do not need to create a MAC file. It is automatically created on startup is it doesn't exist. 
Concerning the host key creation: You can supply any two strings to the _slogkey_ command when creating the key. The ethernet address and the serial number of the host computer is what we use in our application but you are free to choose anything else. Remember to store the host key in a safe place before starting syslog-ng, as it will be overwritten when log message arrive. You will need it to verify, i.e. decrypt the log archive.

kira-syslogng:
Build SUCCESS

furiel:
I am using ubuntu xenial, and installing syslog-ng to a nonstandard location. Oddly enough, I cannot reproduce the runtime failure with autotools now, other than this warning below. I don't know what has changed. I will do a proper check and report after we merge the PR, in case the problem still persists.

```
*** Warning: Linking the executable modules/secure-logging/slogkey/slogkey against the loadable module
*** libsecure-logging.so is not portable!
```

Analogously, there is a similar case in syslog-ng, when an executable needs some functionality from a module: dqtool and diskq module. In that case, diskq provides two libraries: one loadable library for dlopen, and one static library (libsyslog-ng-disk-buffer). dqtool is linked to the static part. Maybe we need to split the `so` similarly with secure logging.

-----------

Could you double check this branch? I think something went wrong with git. There are a lot of commits that should not be visible (170 commits added?), and on the top there is a patch that me seems a revert patch. It removes every file you added.

nbsd:
@furiel I have reset my branch to one prepared by @lbudai. You will need to checkout the branch **rebased-secure-logging** to get the secure logging functionality. The reset was needed in order to prepare the merge of this PR. 
Concerning the linking errors, I get the same error when building on my Fedora system.  I already asked the core developers about the cause of this error. It seems to have something to do with the dynamic library path resolution. At least on my system, it does not affect the functionality at all. Even when I install syslog-ng into a non-standard location, everything works as expected.

e-o-blass:
@furiel 
Let me try clarifying the use of keys, MACs (message authentication codes), and so on.

First, during regular operation, each log event is encrypted and integrity protected with a key called "host key". After syslog-ng has encrypted the current log event with the host key, it also replaces the current host key with a new version. Basically, there is a key file that stores the host key, and this file is overwritten with new content. This new content is (informally) the hash of the old host key, so the new host key, which will then be used for the subsequent log event, is the hash of the old host key. 

More formally, imagine that host keys can be enumerated. When the i-th log event is encrypted, syslog-ng uses the i-th host key `$k_i$` (which is stored in the key file). Afterwards, syslog-ng computes the (i+1)-th host key `$k_{i+1}$`, by "hashing" the i-th key and writing the hash result in that same file, thereby replacing what was there before.

Ok, so the question is then: where does the first or initial key (`$k_0$`) come from? Obviously, the admin has to create it and store it in a path which you specify in your syslog-ng.conf file. For creation, you use that slogkey tool. There is, however, a small subtlety. In our application, and in possible many other scenarios, we have *many* different machines running syslog-ng. For security, all of them require their own independent `$k_0$`. To avoid the admin store and manage a gazillion of different keys, our approach is to deterministically derive all host keys from a common "master key". Only the admin knows the master key and keeps it secret. Knowledge of the master key allows you to derive all host keys, but knowledge of any number of host keys does not allow an adversary to derive other host keys or even the master key.  

So, the first step is to generate a master key with `slogkey -m` and move it to a safe place. Afterwards, for each machine, use `slogkey -d` to derive a host key from that master key. In our application, each machine running syslog-ng is uniquely identified by its ethernet address (MAC address) and some serial, but the tool allows any two input strings (see code). The output of this operation is `$k_0$` for that specific machine. Copy it to the machine, at the position specified in your config.

In addition to encrypting and integrity protection each log event, also the log as a whole is integrity protected. This guarantees, for example, that the log is complete. We store the integrity information in yet another separate file which we call a MAC file. In this context, MAC unfortunately does not mean medium access control, but message authentication code. The admin has to specify a location for the MAC file in syslog-ng.conf. However, when syslog-ng is initialized and run for the first time, there are obviously not log events logged, so this file is expected not to be there. (A warning will show up, though). As soon as log events are processed, this file will be created and updated over time. When you stop syslog-ng (or your machine crashes which happens from time to time in our application) and you restart, syslog-ng will take the contents of a potentially existing MAC file and continue working from there.

Does that make sense? :)

furiel:
Thank you for the detailed explantion. I think I understand now.

How do we proceed now? `rebased-secure-logging` branch indeed contains the patches. However this pr was created from nbsd/master. Do you reset nbsd/master to nbsd/rebase-secure-logging? Or do you close this PR, and open a new one from `rebased-secure-logging` branch?

nbsd:
@furiel  I can rebase the nbsd/rebased-secure-logging to nbsd/master if this is required for merging this PR. I am just waiting for the appropriate instructions.

nbsd:
@furiel I did reset the branch as requested. However, I now get merge conflicts that can only be resolved by someone from the core team.

kira-syslogng:
Build SUCCESS

furiel:
Sorry, I thought @lbudai already resolved it for you. I will try to find out where the communication went wrong. This should have gone much smoother.

kira-syslogng:
Build FAILURE

nbsd:
@furiel Thanks for clarifying. We should be close to the target from my perspective. I am available to perform any additional actions if required.

furiel:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
Okay, so next step is refactoring - but before that, we merge this PR.

Thanks!

PR #3122, opened 2020-02-07
replace tempfile command with mktemp

Kokan:
No changelog is needed.
The tempfile is deprecated and removed from newer version of coreutils, mktemp can replace it.


kira-syslogng:
Build SUCCESS

PR #3123, opened 2020-02-07
Python http header

lbudai:
The `python-http-header` plugin makes it possible for users to implement HTTP header plugins in Python language. It is built on top of signal-slot mechanism: currently HTTP module defines only one signal, that is `signal_http_header_request` and `python-http-header` plugin implements a python binding for this signal. This means that when the `signal_http_header_request` signal is emitted then the connected slot executes the Python code.

The Python interface is :
```
def get_headers(self, body, headers):
```

It should return a string List.
When the plugin fails, http module won't send the http request without the header items by default. 
If you want http module to trying to send the request without these headers, just disable `mark-errors-as-critical` function.

Code, or naming is not final, but it's working. Just wanted to open the PR and get some feedback (...and run CI tests).

Original code was written by Ferenc Sipos.

<details>

```

@version: 3.25

python {
from syslogng import Logger

logger = Logger()

class TestCounter():
    def __init__(self, options):
        self.header = options["header"]
        self.counter = int(options["counter"])
        logger.debug(f"TestCounter class instantiated; options={options}")

    def get_headers(self, body, headers):
        logger.debug(f"get_headers() called, received body={body}, headers={headers}")
       
        response = ["{}: {}".format(self.header, self.counter)]
        self.counter += 1
        return response

    def __del__(self):
        logger.debug("Deleting TestCounter class instance")
};

source s_network {
  network(port(5555));
};

destination d_http {
    http(
        python_http_header(
            class("TestCounter")
            options("header", "X-Test-Python-Counter")
            options("counter", 11)
            mark-errors-as-critical(no)
        )
        url("http://127.0.0.1:8888")
    );
};

log {
    source(s_network);
    destination(d_http);
    flags(flow-control);
};
```
</details>

### DIFF
--- a/modules/http/http-worker.c
+++ b/modules/http/http-worker.c
@@ -28,6 +28,18 @@
 #include "scratch-buffers.h"
 #include "http-signals.h"
 
+#define HTTP_HEADER_FORMAT_ERROR http_header_format_error_quark()
+
+static GQuark http_header_format_error_quark(void)
+{
+  return g_quark_from_static_string("http_header_format_error_quark");
+}
+
+enum HttpHeaderFormatError
+{
+  HTTP_HEADER_FORMAT_SLOT_CRITICAL_ERROR,
+  HTTP_HEADER_FORMAT_SLOT_NON_CRITICAL_ERROR
+};
 
 /* HTTPDestinationWorker */
 
@@ -186,10 +198,37 @@ _append_auth_header(List *list, HTTPDestinationDriver *owner)
 }
 
 static void
-_collect_rest_headers(HTTPDestinationWorker *self)
+_set_error_from_slot_result(const gchar *signal,
+                            HttpHeaderRequestSlotResultType result,
+                            GError **error)
+{
+  switch (result)
+    {
+    case HTTP_HEADER_REQUEST_SLOT_SUCCESS:
+      g_assert(*error == NULL);
+      break;
+    case HTTP_HEADER_REQUEST_SLOT_CRITICAL_ERROR:
+      g_set_error(error, HTTP_HEADER_FORMAT_ERROR,
+                  HTTP_HEADER_FORMAT_SLOT_CRITICAL_ERROR,
+                  "Critical error during slot execution, signal:%s",
+                  signal_http_header_request);
+      break;
+    case HTTP_HEADER_REQUEST_SLOT_PLUGIN_ERROR:
+    default:
+      g_set_error(error, HTTP_HEADER_FORMAT_ERROR,
+                  HTTP_HEADER_FORMAT_SLOT_NON_CRITICAL_ERROR,
+                  "Non-critical error during slot execution, signal:%s",
+                  signal_http_header_request);
+      break;
+    }
+}
+
+static void
+_collect_rest_headers(HTTPDestinationWorker *self, GError **error)
 {
   HttpHeaderRequestSignalData signal_data =
   {
+    .result = HTTP_HEADER_REQUEST_SLOT_SUCCESS,
     .request_headers = self->request_headers,
     .request_body = self->request_body
   };
@@ -197,8 +236,9 @@ _collect_rest_headers(HTTPDestinationWorker *self)
   HTTPDestinationDriver *owner = (HTTPDestinationDriver *) self->super.owner;
 
   EMIT(owner->super.super.super.super.signal_slot_connector, signal_http_header_request, &signal_data);
-}
 
+  _set_error_from_slot_result(signal_http_header_request, signal_data.result, error);
+}
 
 static void
 _add_msg_specific_headers(HTTPDestinationWorker *self, LogMessage *msg)
@@ -234,8 +274,8 @@ _add_common_headers(HTTPDestinationWorker *self)
     list_append(self->request_headers, l->data);
 }
 
-static void
-_format_request_headers(HTTPDestinationWorker *self)
+static gboolean
+_try_format_request_headers(HTTPDestinationWorker *self, GError **error)
 {
   HTTPDestinationDriver *owner = (HTTPDestinationDriver *) self->super.owner;
 
@@ -244,7 +284,9 @@ _format_request_headers(HTTPDestinationWorker *self)
   if (owner->auth_header)
     _append_auth_header(self->request_headers, owner);
 
-  _collect_rest_headers(self);
+  _collect_rest_headers(self, error);
+
+  return (*error == NULL);
 }
 
 static void
@@ -582,6 +624,41 @@ _flush_on_target(HTTPDestinationWorker *self, HTTPLoadBalancerTarget *target)
   return _map_http_status_code(self, target->url, http_code);
 }
 
+static gboolean
+_format_request_headers_error_is_critical(GError *error)
+{
+  return (error->code == HTTP_HEADER_FORMAT_SLOT_CRITICAL_ERROR);
+}
+
+static void
+_format_request_headers_report_error(GError *error)
+{
+  if (error->code == HTTP_HEADER_FORMAT_SLOT_CRITICAL_ERROR)
+    {
+      msg_error("Failed to format HTTP request headers.",
+                evt_tag_str("reason", error->message),
+                evt_tag_printf("action", "request disconnect"));
+    }
+  else
+    {
+      msg_warning("Failed to format HTTP request headers",
+                  evt_tag_str("reason", error->message),
+                  evt_tag_printf("action", "trying to send the request"));
+    }
+}
+
+static gboolean
+_format_request_headers_catch_error(GError **error)
+{
+  g_assert((*error)->domain == HTTP_HEADER_FORMAT_ERROR);
+
+  _format_request_headers_report_error(*error);
+  gboolean unhandled = _format_request_headers_error_is_critical(*error);
+  g_clear_error(error);
+
+  return !unhandled;
+}
+
 /* we flush the accumulated data if
  *   1) we reach batch_size,
  *   2) the message queue becomes empty
@@ -594,6 +671,7 @@ _flush(LogThreadedDestWorker *s, LogThreadedFlushMode mode)
   HTTPLoadBalancerTarget *target, *alt_target = NULL;
   LogThreadedResult retval = LTR_NOT_CONNECTED;
   gint retry_attempts = owner->load_balancer->num_targets;
+  GError *error = NULL;
 
   if (self->super.batch_size == 0)
     return LTR_SUCCESS;
@@ -602,7 +680,12 @@ _flush(LogThreadedDestWorker *s, LogThreadedFlushMode mode)
     return LTR_RETRY;
 
   _finish_request_body(self);
-  _format_request_headers(self);
+
+  if (!_try_format_request_headers(self, &error))
+    {
+      if (!_format_request_headers_catch_error(&error))
+        return LTR_NOT_CONNECTED;
+    }
 
   target = http_load_balancer_choose_target(owner->load_balancer, &self->lbc);
 
--- a/modules/python/python-grammar.ym
+++ b/modules/python/python-grammar.ym
@@ -60,6 +60,7 @@ PythonHttpHeaderPlugin *last_header;
 %token KW_CLASS
 %token KW_IMPORTS
 %token KW_LOADERS
+%token KW_MARK_ERRORS_AS_CRITICAL
 
 %%
 
@@ -248,6 +249,10 @@ python_http_header_option
           }
         | KW_LOADERS python_http_header_loaders
         | KW_OPTIONS '(' python_http_header_custom_options ')'
+        | KW_MARK_ERRORS_AS_CRITICAL '(' yesno ')'
+          {
+            python_http_header_set_mark_errors_as_critical(last_header, $3);
+          }
         ;
 
 python_http_header_custom_options
--- a/modules/python/python-grammar.ym
+++ b/modules/python/python-grammar.ym
@@ -56,7 +56,7 @@ PythonHttpHeaderPlugin *last_header;
 
 %token KW_PYTHON
 %token KW_PYTHON_FETCHER
-%token KW_PYTHON_HEADER
+%token KW_PYTHON_HTTP_HEADER
 %token KW_CLASS
 %token KW_IMPORTS
 %token KW_LOADERS
@@ -84,7 +84,7 @@ start
             last_driver = *instance = python_fetcher_new(configuration);
           }
           '(' python_fetcher_options ')'         { YYACCEPT; }
-        | LL_CONTEXT_INNER_DEST KW_PYTHON_HEADER
+        | LL_CONTEXT_INNER_DEST KW_PYTHON_HTTP_HEADER
           {
             last_header = *instance = python_http_header_new();
           }
--- a/modules/python/python-grammar.ym
+++ b/modules/python/python-grammar.ym
@@ -28,6 +28,7 @@
 #include "python-source.h"
 #include "python-fetcher.h"
 #include "python-logparser.h"
+#include "python-http-header.h"
 #include "python-main.h"
 #include "value-pairs/value-pairs.h"
 #include "logthrsource/logthrsourcedrv.h"
@@ -40,6 +41,9 @@
 #include "cfg-parser.h"
 #include "plugin.h"
 #include "syslog-names.h"
+
+PythonHttpHeaderPlugin *last_header;
+
 }
 
 %name-prefix "python_"
@@ -52,6 +56,7 @@
 
 %token KW_PYTHON
 %token KW_PYTHON_FETCHER
+%token KW_PYTHON_HEADER
 %token KW_CLASS
 %token KW_IMPORTS
 %token KW_LOADERS
@@ -65,8 +70,8 @@ start
           }
           '(' python_dd_options ')'         { YYACCEPT; }
         | LL_CONTEXT_PARSER KW_PYTHON
-	  {
-	    last_parser = *instance = python_parser_new(configuration);
+	      {
+	        last_parser = *instance = python_parser_new(configuration);
           }
           '(' python_parser_options ')' { YYACCEPT; }
         | LL_CONTEXT_SOURCE KW_PYTHON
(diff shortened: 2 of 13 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
wow, I like this a lot :)

modules/python/python-grammar.ym
```diff
@@ -52,6 +56,7 @@
 
 %token KW_PYTHON
 %token KW_PYTHON_FETCHER
+%token KW_PYTHON_HEADER
```
bazsi: this should probably be KW_PYTHON_HTTP_HEADER as this plugin type is HTTP specific.

modules/python/python-http-header.c
```diff
@@ -0,0 +1,390 @@
...
+ * COPYING for details.
+ *
+ */
+
+#include "python-module.h"
+#include "python-helpers.h"
+
+#include "lib/driver.h"
+#include "lib/str-utils.h"
+#include "lib/list-adt.h"
```
bazsi: "lib" shouldn't be part of the includes.

modules/python/python-http-header.c
```diff
@@ -0,0 +1,390 @@
...
+#include "modules/http/http-signals.h"
+
+#include "python-http-header.h"
```
bazsi: this shoud be the very first header to be included.

modules/python/python-http-header.c
```diff
@@ -0,0 +1,390 @@
...
+  Py_ssize_t len = PyList_Size(py_list);
+  for (int i = 0; i < len; i++)
+    {
+      py_str = PyList_GetItem(py_list, i); // Borrowed reference
+      if (!py_str || !PyUnicode_Check(py_str))
+        goto exit;
+
+      if (!(str = PyUnicode_AsUTF8(py_str)))
+        goto exit;
+
```
bazsi: This should probably use _py_get_string_as_string() which handles various python versions better. (e.g. not-unicode strings, which are the default on python2)

modules/python/python-http-header.c
```diff
@@ -0,0 +1,390 @@
...
+  return TRUE;
+
+exit:
+  return FALSE;
+}
+
+static void
+_py_append_str_to_pylist(gconstpointer data, gpointer user_data)
+{
+  PyObject *py_str = PyUnicode_FromString((gchar *) data);
```
bazsi: This should use: _py_string_from_string()

modules/python/python-http-header.c
```diff
@@ -0,0 +1,390 @@
...
+  if (!py_str || !PyUnicode_Check(py_str))
```
bazsi: PyUnicode_Check() is not needed here.

modules/python/python-http-header.c
```diff
@@ -0,0 +1,390 @@
...
+
+exit:
+  Py_XDECREF(py_str); // Even if append() succeeds, it doesn't steal the reference to py_str.
+}
+
+static PyObject *
+_py_convert_list_to_pylist(List *list)
+{
+  PyObject *py_list = PyList_New(0);
+  if (!py_list)
```
bazsi: I am not sure it is worth handling errors here and in similar cases. if PyList_New() fails, I think we can abort as well, so a g_assert() would simplify the code a lot.

modules/python/python-parser.c
```diff
@@ -33,6 +33,7 @@ static CfgLexerKeyword python_keywords[] =
 {
   { "python",                   KW_PYTHON  },
   { "python_fetcher",           KW_PYTHON_FETCHER },
+  { "python_header",            KW_PYTHON_HEADER },
```
bazsi: again, python_http_header()

lbudai:
@bazsi : thanks for your review notes (this pr is in verly early stage, agree with all the notes)

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/python/python-http-header.c
```diff
@@ -0,0 +1,383 @@
...
+  Py_XDECREF(py_ret);
+  Py_XDECREF(py_list);
+  Py_XDECREF(py_ret_list);
+
+  self->last_run = now;
+
+  PyGILState_Release(gstate);
+
+exit:
+  g_assert(self->last_headers);
```
furiel: This cannot be an exception. In case of error, we can get here with `self->last_headers == NULL`.

modules/python/python-http-header.c
```diff
@@ -0,0 +1,383 @@
...
+            evt_tag_str("class", self->class),
+            evt_tag_str("method", "get_headers"),
+            evt_tag_str("return_type", "Tuple"),
+            evt_tag_int("len", PyTuple_Size(py_ret)));
+
+  if (!_py_append_pylist_to_list(py_ret_list, &self->last_headers))
+    {
+      gchar buf[256];
+
+      msg_error("Converting Python List failed",
```
furiel: We can get here in some cases even without python exception. For example `PyList_check` and `PyUnicode_Check` does not generate exception. In those failures, we only get:
```
2020-02-13T13:11:07.582163] Converting Python List failed; class='TestCounter', method='get_headers', exception='None'
```
which is not enough for debugging.

modules/python/python-http-header.c
```diff
@@ -0,0 +1,383 @@
...
+  PyObject *py_str;
+
+  if (!py_list || !PyList_Check(py_list))
+    goto exit;
+
+  Py_ssize_t len = PyList_Size(py_list);
+  for (int i = 0; i < len; i++)
+    {
+      py_str = PyList_GetItem(py_list, i); // Borrowed reference
+      if (!py_str || !PyUnicode_Check(py_str))
```
furiel: This does not work with python2. There is a `_py_is_string` function that could be used.

modules/python/python-http-header.c
```diff
@@ -0,0 +1,383 @@
...
+static void
+_append_headers(PythonHttpHeaderPlugin *self, HttpHeaderRequestSignalData *data)
+{
+  PyObject *py_ret = NULL,
+            *py_list = NULL,
+             *py_args = NULL,
+              *py_ret_list = NULL;
+
+  time_t now = time(NULL);
+  if ((now - self->last_run) < self->timeout)
```
furiel: Do we need this timeout functionality in the framework? This is something that could be implemented in the user code. In a complex scenario, when some headers can be cached while others not, users cannot give universal timeout. Performance wise, I do not think a python side caching would measure significantly either.

My main problem is that I feel unconfortable assigning a timeout parameter to a `response`, because it typically means something else. This of course could be resolved by renaming too (like cache-time or similar), but first I am interested if we want to have it at all.

modules/python/python-http-header.c
```diff
@@ -0,0 +1,383 @@
...
+
+  return py_list;
+}
+
+static gboolean
+_py_attach_bindings(PythonHttpHeaderPlugin *self)
+{
+  PyObject *py_args = NULL;
+
+  self->py.class = _py_resolve_qualified_name(self->class);
```
furiel: This may be a yagni. But ...

In case of python-source and fetcher, and in java binding, it is nice to have a superclass, that users need to inherit from. Through that class, we could inject data (members, methods, constants) cleaner. For now there is no need to inject such thing, but in the future we may need.

What we might consider is to have (and force users to use) a superclass, like `SyslogNGHttpBase`, even if it is a dummy class for now (maybe put into `modules/python/pylib/syslogng/__init__.py`, and then we might avoid writing c, and implement such things in python, similarly to https://github.com/syslog-ng/syslog-ng/pull/3100/commits/5f6ac2cdcfc724e1a043845ff5f78d2510a9d5de.)

Later it would be difficult to introduce a superclass and maintain backward compatibility at the same time.

modules/python/python-http-header.c
```diff
@@ -0,0 +1,383 @@
...
+
+      goto cleanup;
+    }
+
+  py_ret = _py_invoke_function_with_args(self->py.get_headers, py_args, self->class, "_append_headers");
+  if (!py_ret || !PyArg_ParseTuple(py_ret, "iO", &self->timeout, &py_ret_list))
+    {
+      gchar buf[256];
+
+      msg_error("Invalid response returned by Python call",
```
furiel: If there is a bug in the user code (let's say uncatched exception), then this error message is emitted, and the http destination emits the response as if no problem has happened.

I think the correct behavior would be that an error in the python http header plugin would cause error in the http destination, and nothing must be emitted: do a disconnect-reconnect cycle.

This might be challenging to implement, because for now we do not have the protocol to send success/failure through the signal-slot mechanism.

modules/python/python-http-header.c
lbudai: @furiel: this is not a python-side caching... and thus it could have a real performance benefit.
The naming is not the best, and yes, there are better alternatives. But... when the one and only reason for existence of your plugin is to implement a http auth header that has an expiration time measured in hours then this caching mechanism implemented in the C side is a good thing.

modules/python/python-http-header.c
lbudai: I think that signal-slot interfaces should be light and slots should not depend on modules who emits the signals they are connected to.

The general python http header is a slot implementation and the interface between this module and http module is the signal-slot interface.

modules/python/python-http-header.c
lbudai: It is intentional to not add a generic error handling mechanism to signal-slot model.

When a signal owner needs error handling, then it should be part of the interface and each signal-slot connections should implement its own error handling strategy (so in this specific case, we can extend `HttpHeaderRequestSignalData` with some error state).

modules/python/python-http-header.c
furiel: I would also prefer performance, but I am still not convinced we could measure performance difference. A http request is done in the milliseconds range at best, due to network latency. A call to a python function is in microseconds range.

If we could do the caching in python, then we could invent very neat APIs. Also, users could use nice caching libraries (I randomly found https://pypi.org/project/expiringdict/), or we could provide our own.

For example users could write a non-caching
```
headers = {"hdr": "value"}
return headers
```
and with caching with our domain specific decorator
```
@cache_timeout(3600)
headers = {"hdr" : "value"}
return headers
```

Less c and more python would scale better if we want to add another feature in the future. For example we might want to add a cache cancellation function too. For example if the credentials file is changed on the disk, users might want to drop cache. Or in case 403 error code has arrived as a response to the http destination.

news/feature-3123.md
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, options):
+        self.header = options["header"]
+        self.counter = int(options["counter"])
+        self.timeout = 0
+        logger.debug(f"TestCounter class instantiated; options={options}")
+
+    def get_headers(self, body, headers):
+        logger.debug(f"get_headers() called, received body={body}, headers={headers}")
+       
+        response = (self.timeout, [f"{self.header}: {self.counter}"])
```
furiel: I suggest the following change (also in the PR description).

`timeout` should not be part of self, it is intended to be calculated in `get_headers`.
`response` is not appropriate here. Probably `retval` was the intention. Still I would split timeout and headers, and write something similar, instead.

```
        headers = [f"{self.header}: {self.counter}"]
        cache_time = 0
        self.counter += 1
        return (cache_time, headers)
```

The `f` syntax is very neat in python3, but it is not available in python2. Until the python2 support is dropped, I would prefer to see a snippet that could be used in any environment. For example, use format, even though it is less readable.
```
        response = (self.timeout, ["{}: {}".format(self.header, self.counter)])
```

modules/python/python-http-header.c
```diff
@@ -0,0 +1,383 @@
...
+
+static void
+_append_headers(PythonHttpHeaderPlugin *self, HttpHeaderRequestSignalData *data)
+{
+  PyObject *py_ret = NULL,
+            *py_list = NULL,
+             *py_args = NULL,
+              *py_ret_list = NULL;
+
+  time_t now = time(NULL);
```
furiel: The cache needs to be dropped in case of some events. For example after a disconnect reconnect cycle. For example if a remote returns with an autorization error, it would be bad if we cached an invalid credential for an hour.

The reload works nice by the way: after reload, we drop the cache.

kira-syslogng:
Build FAILURE

modules/python/python-http-header.c
lbudai: We touched this topic yesterday IRL but don’t know this comment is added after or before our conversation...

Yes, we could implement more and more features on top of the base concept. But... you perfectly know that this PR opened with reason. So this is not a l’art pour l’art PR where I want to create the Holy Python Http Header module, with decorators and others... 

Shortly: anyone can easily add improvements (or features) on top of it, modify in another PR.

On the performance ...

We are heavy working on Python support.
Maybe I’m wrong, but I’m expecting more Python modules added to syslog-ng than before.
So when we are talking about performance (related to this PR), we should talk about concurrency and parallelism  ... in syslog-ng there are elements in the processing pipeline that are able to run in a parallel way. When we are implementing these elements in Python, it’s not true anymore (unless you run each Python module in a separate process and add an IPC mechanism to syslog-ng ...).
Yes, I’m talking about GIL (and yes, i know about IronPython, etc... but that would be unfair to cite those implementations here: we are using CPython).
So I think when we are talking about performance we should think about E2E, real life scenarios: and when you have more  and more python modules, it’s a good pattern if you cache values to avoid holding the GIL - so it’s not only about how to speed up (...) the HTTP requests by skipping the C->Python calls but about trying to maximalize the number of Python modules running in parallel way.

I’m thinking about split the caching part from this PR as it’s just a perf. improvement...

news/feature-3123.md
lbudai: Man, it's just an example... 
The one and only intention is to show how one can implement such a Python module.
It's an interesting question whether we should support python2 in our example modules... I can do that.

modules/python/python-http-header.c
lbudai: This is related to keep the http class final question...
Until we think that the http class is a final class then all these 'magics' (signal-slot n mechanism...) are about specializing some parts of the http class... 

So the problem you pointed out is not specific to this PR... 
Of course I can resolve it by adding signals to http response codes (or just error codes) and then cache invalidation can be done in a slot.

kira-syslogng:
Build SUCCESS

modules/python/python-http-header.c
furiel: Yes I added this comment before our irl discussion. And I agree with you.

In the end, users can still disable c-side caching if they want (by returning zero), and add their own caching implementation python side. So it is fine for me to have this kind of c-side caching.

lbudai:
Let's summarize the status (my plan):
* within this PR:
  * drop the caching (timeout) 
  * extend `HttpHeaderRequestSignalData` with error propagation

* in another PR(s):
  * when a http (error) response received, emit a signal (details: TBD)
  * add caching (timeout) to python-http-header modul

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
@furiel : (customizable) error propagation is done.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

modules/python/python-http-header.c
```diff
@@ -0,0 +1,399 @@
...
+  log_driver_plugin_free_method(s);
+}
+
+PythonHttpHeaderPlugin *
+python_http_header_new(void)
+{
+  PythonHttpHeaderPlugin *self = g_new0(PythonHttpHeaderPlugin, 1);
+
+  log_driver_plugin_init_instance(&(self->super), PYTHON_HTTP_HEADER_PLUGIN);
+
```
furiel: I would prefer to have mark-errors-as-critical to be true by default, that should be set somewhere around here.

modules/python/python-http-header.c
furiel: This comment might have been lost. A gentle reminder: can you fix this as well?

modules/http/http-worker.c
```diff
@@ -186,19 +198,44 @@ _append_auth_header(List *list, HTTPDestinationDriver *owner)
...
+  switch (result)
+    {
+    case HTTP_HEADER_REQUEST_SLOT_CRITICAL_ERROR:
+      g_set_error(error, HTTP_HEADER_FORMAT_ERROR,
+                  HTTP_HEADER_FORMAT_SLOT_CRITICAL_ERROR,
+                  "Critical error during slot execution, signal:%s",
+                  signal_http_header_request);
+      break;
+    case HTTP_HEADER_REQUEST_SLOT_PLUGIN_ERROR:
+    default:
```
furiel: The success branch is missing from this switch case. As a result: we will emit warning even if the connector was successful.

modules/python/python-http-header.c
```diff
@@ -0,0 +1,397 @@
...
+      goto exit;
+    }
+
+  Py_ssize_t len = PyList_Size(py_list);
+  for (int i = 0; i < len; i++)
+    {
+      py_str = PyList_GetItem(py_list, i); // Borrowed reference
+      if (!_py_is_string(py_str))
+        {
+          msg_debug("");
```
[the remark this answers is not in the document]
lbudai: Hmmm...

modules/python/python-http-header.c
lbudai: Yes, it could be better as default.

modules/http/http-worker.c
lbudai: Ah, indeed. thanks!

lbudai:
@furiel: I think it is done (also changed commit messages/news file with the new default value).

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

modules/python/python-logger.c
```diff
@@ -67,7 +67,10 @@ PyObject *
 py_msg_debug(PyObject *obj, PyObject *args)
 {
   if (!debug_flag)
-    return Py_None;
+    {
+      Py_INCREF(Py_None);
```
alltilla: I do not really understand the ref here. Can you elaborate a bit please? Why is it not needed elsewhere?

lbudai: It fixes a bug. It might be needed elsewhere, but it could be handled in a separate PR.

PR #3126, opened 2020-02-12
Light: Add snmp destination tests

norberttak:
Add snmp destination related test cases to Light.
The test shall be able to start the snmptrapd parser (if it is not installed, the test case will skip)
The trapd output file will be parsed to check for the expected log message.

### DIFF
--- a/tests/python_functional/functional_tests/destination_drivers/snmp_destination/test_snmp_destination_v2c_cisco_ios_traps.py
+++ b/tests/python_functional/functional_tests/destination_drivers/snmp_destination/test_snmp_destination_v2c_cisco_ios_traps.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python
+#############################################################################
+# Copyright (c) 2020 One Identity
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+import pytest
+
+from src.helpers.snmptrapd.conftest import *  # noqa:F403, F401
+
+CISCO_TIMETICKS = '97881'
+CISCO_FACILITY = "SYS"
+CISCO_SEVERITY = "6"
+CISCO_MESSAGE_TYPE = "CONFIG_I"
+CISCO_MESSAGE = "Configured from console by vty1 (10.30.0.32)"
+
+OID1 = "1.3.6.1.4.1.9.9.41.1.2.3.1.2.55"
+OID2 = "1.3.6.1.4.1.9.9.41.1.2.3.1.3.55"
+OID3 = "1.3.6.1.4.1.9.9.41.1.2.3.1.4.55"
+OID4 = "1.3.6.1.4.1.9.9.41.1.2.3.1.5.55"
+OID5 = "1.3.6.1.4.1.9.9.41.1.2.3.1.6.55"
+OID6 = "1.3.6.1.6.3.1.1.4.1.0"
+TRAP_DATA = ".1.3.6.1.4.1.9.9.41.2.0.1"
+
+input_log = "<38>Feb 11 21:27:22 testhost testprogram[9999]: test message\n"
+
+expected_logs = [
+    '.{} = STRING: "{}"'.format(OID1, CISCO_FACILITY),
+    '.{} = INTEGER: {}'.format(OID2, CISCO_SEVERITY),
+    '.{} = STRING: "{}"'.format(OID3, CISCO_MESSAGE_TYPE),
+    '.{} = STRING: "{}"'.format(OID4, CISCO_MESSAGE),
+    '.{} = Timeticks: ({}) 0:16:18.81'.format(OID5, CISCO_TIMETICKS),
+    '.{} = OID: {}'.format(OID6, TRAP_DATA),
+]
+
+
+@pytest.mark.snmp
+def test_snmp_dest_v2_cisco_ios_trap(config, syslog_ng, snmptrapd):
+    # create source driver config
+    file_source = config.create_file_source(file_name="input.log")
+
+    # create destination driver config
+    snmp_objs = (
+        "'{}', 'Octetstring', '{}'".format(OID1, CISCO_FACILITY),
+        "'{}', 'Integer', '{}'".format(OID2, CISCO_SEVERITY),
+        "'{}', 'Octetstring', '{}'".format(OID3, CISCO_MESSAGE_TYPE),
+        "'{}', 'Octetstring', '{}'".format(OID4, CISCO_MESSAGE),
+        "'{}', 'Timeticks', '{}'".format(OID5, CISCO_TIMETICKS),
+    )
+
+    trap_obj = "'.{}','Objectid','{}'".format(OID6, TRAP_DATA)
+
+    snmp_destination = config.create_snmp_destination(
+        host="'127.0.0.1'",
+        port=snmptrapd.get_port(),
+        version="'v2c'",
+        snmp_obj=snmp_objs,
+        trap_obj=trap_obj,
+    )
+
+    config.create_logpath(statements=[file_source, snmp_destination])
+    config.update_global_options(keep_hostname="yes")
+    file_source.write_log(input_log)
+
+    syslog_ng.start(config)
+
+    snmptrap_logs = snmptrapd.get_logs()[2].rstrip().split("\t")
+
+    for exp_log in expected_logs:
+        assert exp_log in snmptrap_logs
--- a/tests/python_functional/src/helpers/snmptrapd/conftest.py
+++ b/tests/python_functional/src/helpers/snmptrapd/conftest.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+#############################################################################
+# Copyright (c) 2020 One Identity
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+import os
+
+import pytest
+from pathlib2 import Path
+from psutil import TimeoutExpired
+
+import src.testcase_parameters.testcase_parameters as tc_parameters
+from src.driver_io.file.file_io import FileIO
+from src.executors.process_executor import ProcessExecutor
+from src.syslog_ng_config.statements.destinations.destination_reader import DestinationReader
+
+
+class SNMPtrapd(object):
+    def __init__(self, port):
+        self.snmptrapd_proc = None
+        self.port = port
+
+        self.snmptrapd_log = Path(tc_parameters.WORKING_DIR, "snmptrapd_log")
+        self.snmptrapd_stdout_path = Path(tc_parameters.WORKING_DIR, "snmptrapd_stdout")
+        self.snmptrapd_stderr_path = Path(tc_parameters.WORKING_DIR, "snmptrapd_stderr")
+
+    def start(self):
+        if self.snmptrapd_proc is not None:
+            return
+
+        self.snmptrapd_proc = ProcessExecutor().start(
+            [
+                "snmptrapd", "-f",
+                "--disableAuthorization=yes",
+                "-C",
+                "-On",
+                "--doNotLogTraps=no",
+                self.port,
+                "-LF", "6-6", os.path.relpath(str(self.snmptrapd_log)),
+            ],
+            self.snmptrapd_stdout_path,
+            self.snmptrapd_stderr_path,
+        )
+        return self.snmptrapd_proc.is_running()
+
+    def stop(self):
+        if self.snmptrapd_proc is None:
+            return
+
+        self.snmptrapd_proc.terminate()
+        try:
+            self.snmptrapd_proc.wait(4)
+        except TimeoutExpired:
+            self.snmptrapd_proc.kill()
+
+        self.snmptrapd_proc = None
+
+    def get_port(self):
+        return self.port
+
+    def get_logs(self):
+        file_reader = DestinationReader(FileIO)
+        return file_reader.read_all_logs(self.snmptrapd_log)
+
+
+@pytest.fixture(scope="module")
+def snmptrapd(some_port):
+    server = SNMPtrapd(some_port)
+    server.start()
+    yield server
+    server.stop()
--- a/tests/python_functional/src/syslog_ng_config/statements/destinations/destination_reader.py
+++ b/tests/python_functional/src/syslog_ng_config/statements/destinations/destination_reader.py
@@ -23,6 +23,7 @@
 import logging
 
 from src.message_reader.message_reader import MessageReader
+from src.message_reader.message_reader import READ_ALL_MESSAGES
 from src.message_reader.single_line_parser import SingleLineParser
 
 logger = logging.getLogger(__name__)
@@ -45,3 +46,6 @@ class DestinationReader(object):
         read_description = "Content has been read from\nresource: {}\ncontent: {}\n".format(path, messages)
         logger.info(read_description)
         return messages
+
+    def read_all_logs(self, path):
+        return self.read_logs(path, READ_ALL_MESSAGES)
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ addons:
       - doxygen
       - libsnmp-dev
       - ninja-build
+      - snmptt
 install:
       - python2 -m pip install --user --cache-dir=$PWD/pip-cache -r requirements.txt
       - python2 -m pip list
--- a/news/developer-note-3126.md
+++ b/news/developer-note-3126.md
@@ -1,0 +1,9 @@
+`snmp test in Light`: snmp destination test case written in Light test framework
+
+snmp destination related functional tests can be run from Light framework.
+This test requires snmptrapd as an external dependency. If you don't want to
+run this test, you can use the pytest's marker discovery feature:
+
+```python -m pytest ... -m 'not snmp'``` 
+
+The test is run by syslog-ng's Travis workflow.
--- a/tests/python_functional/functional_tests/conftest.py
+++ b/tests/python_functional/functional_tests/conftest.py
@@ -55,3 +55,13 @@ def setup(request):
 
     copy_file(testcase_parameters.get_testcase_file(), testcase_parameters.get_working_dir())
     request.addfinalizer(lambda: logger.info("Report file path\n{}\n".format(calculate_report_file_path(testcase_parameters.get_working_dir()))))
+
+
+class PortAllocator():
+    CURRENT_DYNAMIC_PORT = 30000
+
+
+@pytest.fixture(scope="module")
+def some_port():
+    PortAllocator.CURRENT_DYNAMIC_PORT += 1
+    return PortAllocator.CURRENT_DYNAMIC_PORT
--- a/tests/python_functional/pytest.ini
+++ b/tests/python_functional/pytest.ini
@@ -4,3 +4,5 @@ log_cli=false
 log_level=INFO
 log_cli_format= %(asctime)s:%(msecs)d - %(filename)s:%(lineno)d->%(funcName)s - %(levelname)s - %(message)s
 log_file_format= %(asctime)s:%(msecs)d - %(filename)s:%(lineno)d->%(funcName)s - %(levelname)s - %(message)s
+markers =
+    snmp: mark test cases which are related to snmp
--- a/tests/python_functional/src/syslog_ng_config/renderer.py
+++ b/tests/python_functional/src/syslog_ng_config/renderer.py
@@ -71,6 +71,9 @@ class ConfigRenderer(object):
                 self.__syslog_ng_config_content += "        {}(\n".format(option_name)
                 self.__render_driver_options(option_value)
                 self.__syslog_ng_config_content += "        )\n"
+            elif (isinstance(option_value, tuple) or isinstance(option_value, list)):
+                for element in option_value:
+                    self.__syslog_ng_config_content += "        {}({})\n".format(option_name, element)
             else:
                 self.__syslog_ng_config_content += "        {}({})\n".format(option_name, option_value)
 
--- a/tests/python_functional/src/syslog_ng_config/statements/destinations/snmp_destination.py
+++ b/tests/python_functional/src/syslog_ng_config/statements/destinations/snmp_destination.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+#############################################################################
+# Copyright (c) 2015-2020 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+from src.syslog_ng_config.statements.destinations.destination_driver import DestinationDriver
+
+
+class SnmpDestination(DestinationDriver):
+    def __init__(self, **options):
+        self.driver_name = "snmp"
+        super(SnmpDestination, self).__init__(None, options)
--- a/tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
+++ b/tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
@@ -27,6 +27,7 @@ from src.driver_io.file.file_io import FileIO
 from src.syslog_ng_config.renderer import ConfigRenderer
 from src.syslog_ng_config.statement_group import StatementGroup
 from src.syslog_ng_config.statements.destinations.file_destination import FileDestination
+from src.syslog_ng_config.statements.destinations.snmp_destination import SnmpDestination
 from src.syslog_ng_config.statements.filters.filter import Filter
 from src.syslog_ng_config.statements.logpath.logpath import LogPath
 from src.syslog_ng_config.statements.parsers.parser import Parser
@@ -94,6 +95,9 @@ class SyslogNgConfig(object):
 
     def create_file_destination(self, **options):
         return FileDestination(**options)
+
+    def create_snmp_destination(self, **options):
+        return SnmpDestination(**options)
 
     def create_logpath(self, statements=None, flags=None):
         logpath = self.__create_logpath_with_conversion(statements, flags)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
```diff
@@ -0,0 +1,83 @@
...
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.2.55', 'Octetstring', '%s'" % CISCO_FACILITY,
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.3.55', 'Integer', '%s'" % CISCO_SEVERITY,
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.4.55', 'Octetstring', '%s'" % CISCO_MESSAGE_TYPE,
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.5.55', 'Octetstring', '%s'" % CISCO_MESSAGE,
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.6.55', 'Timeticks', '%s'" % CISCO_TIMETICKS,
+    )
+
+    trap_obj = "'.1.3.6.1.6.3.1.1.4.1.0','Objectid','.1.3.6.1.4.1.9.9.41.2.0.1'"
+
+    snmp_nonstandard_port = 2222
```
furiel: I do not really like 2222 as hardcoded port.

If you like it, please pick https://github.com/furiel/syslog-ng/commit/03632c2d1f4a7881c439023a1598687b2b4fcb69, which contains a very simple dynamic port allocator.

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
```diff
@@ -0,0 +1,83 @@
...
+CISCO_FACILITY = "SYS"
+CISCO_SEVERITY = "6"
+CISCO_MESSAGE_TYPE = "CONFIG_I"
+CISCO_MESSAGE = "Configured from console by vty1 (10.30.0.32)"
+
+input_log = "<38>Feb 11 21:27:22 testhost testprogram[9999]: test message\n"
+expected_log = 'iso.3.6.1.4.1.9.9.41.1.2.3.1.2.55 = STRING: "' + CISCO_FACILITY + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.3.55 = INTEGER: ' + CISCO_SEVERITY + '	iso.3.6.1.4.1.9.9.41.1.2.3.1.4.55 = STRING: "' + CISCO_MESSAGE_TYPE + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.5.55 = STRING: "' + CISCO_MESSAGE + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.6.55 = Timeticks: (' + CISCO_TIMETICKS + ')'
+
+
+@pytest.mark.skipif(os.system("which snmptrapd") != 0, reason="snmptrapd can't found on system")
```
furiel: Instead of this, I would prefer installing snmptrapd into the test docker images.

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
```diff
@@ -0,0 +1,83 @@
...
+        trap_obj=trap_obj,
+    )
+
+    config.create_logpath(statements=[file_source, snmp_destination])
+    config.update_global_options(keep_hostname="yes")
+    file_source.write_log(input_log, counter)
+
+    syslog_ng.start(config)
+
+    time.sleep(1)
```
furiel: Is this sleep necessary? What is the purpose?

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
```diff
@@ -0,0 +1,83 @@
...
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.3.55', 'Integer', '%s'" % CISCO_SEVERITY,
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.4.55', 'Octetstring', '%s'" % CISCO_MESSAGE_TYPE,
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.5.55', 'Octetstring', '%s'" % CISCO_MESSAGE,
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.6.55', 'Timeticks', '%s'" % CISCO_TIMETICKS,
+    )
+
+    trap_obj = "'.1.3.6.1.6.3.1.1.4.1.0','Objectid','.1.3.6.1.4.1.9.9.41.2.0.1'"
+
+    snmp_nonstandard_port = 2222
+    snmptrapd = SNMPtrapd(snmp_nonstandard_port)
```
furiel: It would be cool if `snmptrapd` would be a fixture. The startup part would start the service, teardown would stop it (startup: code before yield, teardown: code after the yield).

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
```diff
@@ -0,0 +1,83 @@
...
+CISCO_MESSAGE_TYPE = "CONFIG_I"
+CISCO_MESSAGE = "Configured from console by vty1 (10.30.0.32)"
+
+input_log = "<38>Feb 11 21:27:22 testhost testprogram[9999]: test message\n"
+expected_log = 'iso.3.6.1.4.1.9.9.41.1.2.3.1.2.55 = STRING: "' + CISCO_FACILITY + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.3.55 = INTEGER: ' + CISCO_SEVERITY + '	iso.3.6.1.4.1.9.9.41.1.2.3.1.4.55 = STRING: "' + CISCO_MESSAGE_TYPE + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.5.55 = STRING: "' + CISCO_MESSAGE + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.6.55 = Timeticks: (' + CISCO_TIMETICKS + ')'
+
+
+@pytest.mark.skipif(os.system("which snmptrapd") != 0, reason="snmptrapd can't found on system")
+@pytest.mark.parametrize(
+    "input_log, expected_log, counter", [
```
furiel: For now I do not see much benefit of this parametrization, just increases code complexity. Just optional comment, but I would remove it.

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
lbudai: And what if someone wants to run tests in a different  environment (and don't want to install snmptrapd...)?

furiel: If there is a manual exclude mechanism (blacklist style), I am fine with that. I only have problem with the automatic exlude part.

If the snmptrapd binary somehow disappears unintentionally from our test environment, then these tests are not executed silently, and we would never know it.

We followed this philosophy in https://github.com/syslog-ng/syslog-ng/pull/1878#pullrequestreview-95697739 too.

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: ok, I reused your code. unfortunately the cherry-pick doesn't work beacuse you have other changes in the mentioned commit. I simple copy your code into my change.

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: no, it isn't. removed

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: agree. I moved the snmptrapd into a fixture

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: OK, removed

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: In my mind the output of pytest which states that test case is SKIP doesn't mean it is silent.
Of course, I'm ready to remove this skip mechanism but I afraid it will case unnecessary problems for those users whom doesn't have trapd installed on the system and don't care about snmp destination at all.

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
furiel: Yes, the console output can tell which tests are skipped. The problem is that the little box in the PR-s do not show it.

Prior this change, it was enough for us to give a quick glance that all tests are green, and know that the regression passed.
After this change, besides checking that all tests are green, we need to go into the console log too, and look for the snmp tests and make sure they executed. As we cannot be sure anymore which tests are executed, and which are not.

And that's what I would like to avoid.

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python
```
mitzkia: In pytest by default testcase discovery mechanism collects "test_.*.py" files. So in our case this testcase would not be discovered if I search tests only from top dir ("functions/"). My suggestion for testcase filename: test_snmptrapd_dest_acceptance.py and for testcase name: test_snmptrapd_dest_acceptance
Lets avoid "-" characters

tests/python_functional/src/helpers/snmptrapd/snmptrapd.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python
```
mitzkia: This file contains a fixture definition for our testcases. In this case filename should be called as: "conftest.py"

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
```diff
@@ -0,0 +1,68 @@
...
+CISCO_SEVERITY = "6"
+CISCO_MESSAGE_TYPE = "CONFIG_I"
+CISCO_MESSAGE = "Configured from console by vty1 (10.30.0.32)"
+
+input_log = "<38>Feb 11 21:27:22 testhost testprogram[9999]: test message\n"
+expected_log = 'iso.3.6.1.4.1.9.9.41.1.2.3.1.2.55 = STRING: "' + CISCO_FACILITY + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.3.55 = INTEGER: ' + CISCO_SEVERITY + '	iso.3.6.1.4.1.9.9.41.1.2.3.1.4.55 = STRING: "' + CISCO_MESSAGE_TYPE + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.5.55 = STRING: "' + CISCO_MESSAGE + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.6.55 = Timeticks: (' + CISCO_TIMETICKS + ')'
+
+
+@pytest.mark.skipif(os.system("which snmptrapd") != 0, reason="snmptrapd can't found on system")
+def test_acceptance(config, syslog_ng, snmptrapd):
```
mitzkia: In order to use our snmptrapd fixture we should import the module: "from src.helpers.snmptrapd.conftest import *"

tests/python_functional/src/helpers/snmptrapd/snmptrapd.py
```diff
@@ -0,0 +1,82 @@
...
+        self.snmptrapd_stdout_path = Path(tc_parameters.WORKING_DIR, "snmptrapd_stdout")
+        self.snmptrapd_stderr_path = Path(tc_parameters.WORKING_DIR, "snmptrapd_stderr")
+
+    def start(self):
+        self.snmptrapd_proc = ProcessExecutor().start(
+            [
+                "snmptrapd", "-f",
+                "--disableAuthorization=yes",
+                self.port,
+                "-Lf", os.path.relpath(self.snmptrapd_log),
```
mitzkia: `self.snmptrapd_log` object type is PosixPath, it does not have `startswith()` method (because of python2 compatibility). It should be converted to str.
`os.path.relpath(str(self.snmptrapd_log))` will work

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+CISCO_TIMETICKS = '97881'
+CISCO_FACILITY = "SYS"
+CISCO_SEVERITY = "6"
+CISCO_MESSAGE_TYPE = "CONFIG_I"
+CISCO_MESSAGE = "Configured from console by vty1 (10.30.0.32)"
+
+input_log = "<38>Feb 11 21:27:22 testhost testprogram[9999]: test message\n"
+expected_log = 'iso.3.6.1.4.1.9.9.41.1.2.3.1.2.55 = STRING: "' + CISCO_FACILITY + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.3.55 = INTEGER: ' + CISCO_SEVERITY + '	iso.3.6.1.4.1.9.9.41.1.2.3.1.4.55 = STRING: "' + CISCO_MESSAGE_TYPE + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.5.55 = STRING: "' + CISCO_MESSAGE + '"	iso.3.6.1.4.1.9.9.41.1.2.3.1.6.55 = Timeticks: (' + CISCO_TIMETICKS + ')'
```
mitzkia: In our case the global variables "CISCO_" are not necessary, we have used them only once. I would recommend to not used them and move this static values inside the expected_log.

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
```diff
@@ -0,0 +1,68 @@
...
+
+@pytest.mark.skipif(os.system("which snmptrapd") != 0, reason="snmptrapd can't found on system")
+def test_acceptance(config, syslog_ng, snmptrapd):
+    # create source driver config
+    file_source = config.create_file_source(file_name="input.log")
+
+    # create destination driver config
+    snmp_objs = (
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.2.55', 'Octetstring', '%s'" % CISCO_FACILITY,
+        "'1.3.6.1.4.1.9.9.41.1.2.3.1.3.55', 'Integer', '%s'" % CISCO_SEVERITY,
```
mitzkia: Format string: '%s' should be replaced with '{}'.format() form

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
```diff
@@ -0,0 +1,68 @@
...
+        trap_obj=trap_obj,
+    )
+
+    config.create_logpath(statements=[file_source, snmp_destination])
+    config.update_global_options(keep_hostname="yes")
+    file_source.write_log(input_log)
+
+    syslog_ng.start(config)
+
+    assert snmptrapd.wait_for_trap(expected_log)
```
mitzkia: Testcase assertion should be on top level, without this the report of assertion will contain only function calls which does not help on debugging. Let move back the logic here (to the testcase)
```
    dst_file = DestinationReader(FileIO)
    assert expected_log in dst_file.read_all_logs(snmptrapd.snmptrapd_log)
```

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: file renamed to test_snmp_destination_v2c_cisco_ios_traps.py

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
mitzkia: I am agreed with @furiel 
It is more important to see if something starts to fail, then to add a convenient running feature to our testcases. 
But in the other hand we should somehow be able to filter this new type of testcases (which has external dependencies). Pytest has a support to mark (label) the testcases. [Control skipping testcases](https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option)
I would recommend to introduce a new external running parameter "--skip-tests-with-external-deps". We already have such functionality (with --runslow) https://github.com/syslog-ng/syslog-ng/blob/14a905bb839867afe52109191a6cb366b6a65644/tests/python_functional/conftest.py#L43

tests/python_functional/src/helpers/snmptrapd/snmptrapd.py
norberttak: done

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: done

tests/python_functional/src/helpers/snmptrapd/snmptrapd.py
norberttak: done

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: I don't like this idea as this is against the encapsulation concept. The snmptrapd_log and the handling of trapd's log file is an internal implementation detail of the helper class.

To help the debugging, I can put debug/error log messages into wait_for_trap function to report the details of the not matching patterns.

mitzkia: Ok, I have an other suggestion. What if we change the original `wait_for_trap()` to something like this:
```
    def get_trap_logs(self):
        file_reader = DestinationReader(FileIO)
        return file_reader.read_all_logs(self.snmptrapd_log)
```

and in the testcase we can call assert like this:
```
    assert expected_log in  snmptrapd.get_trap_logs()
```

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
mitzkia: I also recommend to handle calling snmptrapd binary via ProcessExecutor() class with a `try..except` block maybe a new Exception class can be introduced here `class MissingExternalDependency(Exception)`

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: sounds good :-)

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: done

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: done

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: an exception for pylint have to be added temporary

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
norberttak: I added the --skip-tests-with-external-deps. This is False by default, so it will try to run the snmp tests as long as you don't specify the command line option.
I tried to add the snmptrapd package to dbld and travis, but I need some help with this. As far as I see for travis we use only @lbudai OBS repo. I guess we have to add the snmptrapd package to the OBS first. Am I correct?

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

tests/python_functional/src/helpers/snmptrapd/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, port):
+        self.snmptrapd_proc = None
+        self.port = port
+
+        self.snmptrapd_log = Path(tc_parameters.WORKING_DIR, "snmptrapd_log")
+        self.snmptrapd_stdout_path = Path(tc_parameters.WORKING_DIR, "snmptrapd_stdout")
+        self.snmptrapd_stderr_path = Path(tc_parameters.WORKING_DIR, "snmptrapd_stderr")
+
+    def start(self):
+        self.snmptrapd_proc = ProcessExecutor().start(
```
alltilla: What do you think about adding an `assert self.snmptrapd_proc is None`, to enforce correct API usage?

tests/python_functional/src/helpers/snmptrapd/conftest.py
```diff
@@ -0,0 +1,82 @@
...
+
+import src.testcase_parameters.testcase_parameters as tc_parameters
+from src.driver_io.file.file_io import FileIO
+from src.executors.process_executor import ProcessExecutor
+from src.syslog_ng_config.statements.destinations.destination_reader import DestinationReader
+
+
+class SNMPtrapd(object):
+    def __init__(self, port):
+        self.snmptrapd_proc = None
```
alltilla: We might want to add a `which snmptrapd` check, to make sure we have the necessary binary, or send a human readable error message.

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
mitzkia: thanks

tests/python_functional/src/helpers/snmptrapd/snmptrapd.py
mitzkia: thanks

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
mitzkia: thanks

tests/python_functional/src/helpers/snmptrapd/snmptrapd.py
mitzkia: thanks

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
mitzkia: thanks

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
mitzkia: thanks

tests/python_functional/functional_tests/destination_drivers/snmp_destination/001-snmp-destination-v2c-normal-working-compared-cisco-ios-traps-default-values.py
mitzkia: thanks

tests/python_functional/conftest.py
```diff
@@ -65,6 +71,12 @@ def get_current_date():
 
 # Pytest Hooks
 def pytest_collection_modifyitems(config, items):
+    if config.getoption("--skip-tests-with-external-deps"):
```
mitzkia: Sorry about that, we have found that pytest has an internal support to run explicit markers, for example:
```console
python -m pytest ... -m 'external_dependency'
```
- so with this I would remove this '--skip-tests-with-external-deps' option, and instead of this user can use `-m 'some-marker'` option

Other thing:
Pytest can discover all markers inside tests with
```config
python -m pytest --markers
```
- so with this we can add more specific markers like 'snmp', If we will have a lot of markers user can discover it, and add them separately (customized):
```
python -m pytest ... -m 'snmp and other-module and not some-other-modul'
```
- so for now I would rename our 'external_dependency' to more concrete: 'snmp'

norberttak: done. now you can control the running of snmp tests by the -m 'snmp' parameter

tests/python_functional/src/helpers/snmptrapd/conftest.py
norberttak: that's where we were originally. now we concluded that if snmptrapd is not available we will throw an exception. The text of exception is pretty clear and I can display almost the same message.

alltilla: Yep, that's good enough :)

kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng_config/statements/destinations/destination_reader.py
```diff
@@ -45,3 +46,6 @@ def read_logs(self, path, counter):
         read_description = "Content has been read from\nresource: {}\ncontent: {}\n".format(path, messages)
         logger.info(read_description)
         return messages
+
+    def read_all_logs(self, path):
+        return " ".join(self.read_logs(path, READ_ALL_MESSAGES))
```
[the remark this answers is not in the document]
norberttak: done

tests/python_functional/src/helpers/snmptrapd/conftest.py
norberttak: I don't want to add assert into a helper class. Instead of this I added a check and return if the process is already running.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/destination_drivers/snmp_destination/test_snmp_destination_v2c_cisco_ios_traps.py
```diff
@@ -0,0 +1,80 @@
...
+CISCO_MESSAGE = "Configured from console by vty1 (10.30.0.32)"
+
+input_log = "<38>Feb 11 21:27:22 testhost testprogram[9999]: test message\n"
+
+expected_log = (
+    'iso.3.6.1.4.1.9.9.41.1.2.3.1.2.55 = STRING: "{}"\t'
+    'iso.3.6.1.4.1.9.9.41.1.2.3.1.3.55 = INTEGER: {}\t'
+    'iso.3.6.1.4.1.9.9.41.1.2.3.1.4.55 = STRING: "{}"\t'
+    'iso.3.6.1.4.1.9.9.41.1.2.3.1.5.55 = STRING: "{}"\t'
+    'iso.3.6.1.4.1.9.9.41.1.2.3.1.6.55 = Timeticks: ({}) 0:16:18.81'
```
alltilla: We do not check the `trap_obj = "'.1.3.6.1.6.3.1.1.4.1.0','Objectid','.1.3.6.1.4.1.9.9.41.2.0.1'"` here, and with that missing, the test could hide, if our OID handling is erroneous.

It really should be added, and with that final element, the `expected_log` will be the whole line, so there will be no need to call `" ".join()` in the assert below, we can check for the exact match with `assert expected_log in snmptrapd.get_logs()`.

Please add this final touch, it is not a cosmetic incompletion, but a functional one.

alltilla: Apart from this, it is an approve from me :)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/destination_drivers/snmp_destination/test_snmp_destination_v2c_cisco_ios_traps.py
```diff
@@ -0,0 +1,86 @@
...
+        trap_obj=trap_obj,
+    )
+
+    config.create_logpath(statements=[file_source, snmp_destination])
+    config.update_global_options(keep_hostname="yes")
+    file_source.write_log(input_log)
+
+    syslog_ng.start(config)
+
+    snmptrap_logs = snmptrapd.get_logs()[2].split("\t")
```
alltilla: * adding `.rstrip()` (`snmptrap_logs = snmptrapd.get_logs()[2].rstrip().split("\t")`)
 * and removing the `\n` in line 49 (`    '.{} = OID: {}'.format(OID6, TRAP_DATA),`)

will make everything green.

kira-syslogng:
Build SUCCESS

PR #3127, opened 2020-02-12
poll-events: introduce suspend_watches virtual function

gaborznagy:
update_watches is a mandatory virtual function,
which was used for update- and suspend cases too.

Usually a suspend_watches() call would be simply
a stop_watches(), but in some cases stop_watches()
and suspend_watches() differ.

Previously, poll_events_suspend_watches() called update_watches with 0.
Therefore every update_watches implementation had to be aware of this use case.

News file is not affected.

### DIFF
--- a/lib/poll-events.h
+++ b/lib/poll-events.h
@@ -37,6 +37,7 @@ struct _PollEvents
   void (*start_watches)(PollEvents *self);
   void (*stop_watches)(PollEvents *self);
   void (*update_watches)(PollEvents *self, GIOCondition cond);
+  void (*suspend_watches)(PollEvents *self);
   void (*free_fn)(PollEvents *self);
 };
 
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

gaborznagy:
Missing news file for #2963 added, removed the WIP flag.

kira-syslogng:
Build SUCCESS

lbudai:
@gaborznagy : and when suspend differs from stop, start can be still used as resume?

gaborznagy:
@lbudai yes you can resume suspended watches with **update_watches**. Not with start_watches though, in LogReader which uses poll-events it can be seen, that in start_watches, there is an assertion that watches should not run prior to call. And this should be true for poll_events too.

This was the symmetry always, I've just clearly separated update_watches() from suspend_watches() (prior the same function have been used for 2 cases, which was confusing internally).

PR #3128, opened 2020-02-13
Migrate to using severity instead of level

bazsi:
This is a followup to #3115


kira-syslogng:
Build FAILURE

lbudai:
@bazsi : don't forget to add the news file fragment
( #3066  )

bazsi:
added the news fragment, rebased to current master.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

gaborznagy:
We have a `json-parser` test where there was a SEVERITY field, which now breaks.
I guess we need to use a prefix for that json-parser(), but I'll double-check it.

bazsi:
it seems so. "SEVERITY" as a name-value pair is reserved for syslog-ng to use (all caps without dots), so that testcase is wrong in using those names.

gaborznagy:
I've added it 3.26 milestone as https://github.com/syslog-ng/syslog-ng/pull/3115 has it too.
If we will have a `$(set-severity)` template fn, we should have `$SEVERITY` macro as well and not just `$LEVEL` or `$PRIORITY`.

gaborznagy:
@kira-syslogng retest this please test branch=json-parser-prefix;

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please test branch=json-parser-prefix;

kira-syslogng:
Build SUCCESS

gaborznagy:
Kira is happy after the json-parser test case is updated (the json-parser in the config will use a prefix, so no name "collison" will occur).

:fireworks:

lib/cfg-parser.c
```diff
@@ -121,8 +121,9 @@ static CfgLexerKeyword main_keywords[] =
   { "time_sleep",         KW_TIME_SLEEP, KWS_OBSOLETE, "time_sleep() has been deprecated" },
   { "file_template",      KW_FILE_TEMPLATE },
   { "proto_template",     KW_PROTO_TEMPLATE },
-  { "default_level",      KW_DEFAULT_LEVEL },
-  { "default_priority",   KW_DEFAULT_LEVEL },
+  { "default_level",      KW_DEFAULT_SEVERITY },
```
gaborznagy: shouldn't we use `KWS_OBSOLETE` to indicate that we intend to deprecate it?

lib/filter/filter-expr-parser.c
```diff
@@ -47,8 +47,9 @@ static CfgLexerKeyword filter_expr_keywords[] =
   { "!=",                 KW_NUM_NE },
   { ">=",                 KW_NUM_GE },
   { ">",                  KW_NUM_GT },
-  { "level",              KW_LEVEL },
-  { "priority",           KW_LEVEL },
+  { "severity",           KW_SEVERITY },
+  { "level",              KW_SEVERITY },
```
gaborznagy: shouldn't we use `KWS_OBSOLETE` to indicate that we intend to deprecate level and severity?

lib/syslog-names.c
```diff
@@ -26,23 +26,23 @@
...
-  {"warning",   LEVEL_CODE(4) },
-  {"warn",      LEVEL_CODE(4) },
-  {"notice",    LEVEL_CODE(5) },
-  {"info",      LEVEL_CODE(6) },
-  {"debug",     LEVEL_CODE(7) },
+  {"emerg",     SEVERITY_CODE(0) },
+  {"panic",     SEVERITY_CODE(0) },
+  {"alert",     SEVERITY_CODE(1) },
+  {"crit",      SEVERITY_CODE(2) },
+  {"err",       SEVERITY_CODE(3) },
```
gaborznagy: conflict ahead with #3115

lib/cfg-parser.c
bazsi: My understanding about the final conclusion was that we are not actively deprecating the aliases, as that would push a lot of work to anyone using these options and maintaining the alias is not that much of a maintenance nightmare. the level() filter is pretty common in any configuration. And I would deprecate these options as a whole (e.g. level(), default-level(), $LEVEL at the same time).

lib/filter/filter-expr-parser.c
bazsi: see above.

lib/syslog-names.c
bazsi: let's make sure that goes in first.

bazsi:
> 2 questions:
> 
> The HTTP module has an `X-Syslog-Level` header. Do you want to keep the name "Level" for compatibility reasons? It's been there since 2016, but AFAIK the name of the header is not part of any standard.

Good point. But changing that would break configurations. I am not sure many people use that header, but I don't want to break their configuration if they do. But like I made this point in a comment in the code:

```
  /* NOTE: I have my doubts that these headers make sense at all.  None of
   * the HTTP collectors I know of, extract this information from the
   * headers and it makes batching several messages into the same request a
   * bit more complicated than it needs to be.  I didn't want to break
   * backward compatibility when batching was introduced, however I think
   * this should eventually be removed */
```

I think they should eventually be removed, they are not added in batched mode. I wouldn't add this removal to this PR though, as this is a non-breaking PR.

> 
> I've found 2 macros that have not been migrated to the name `severity`:
> `LOG_PRIMASK` and `LOG_PRI()`. They are borrowed from `sys/syslog.h`, but we defined our own of these in `syslog-names.h`. Should we keep those names to be consistent with `syslog(3)`?

I left it in for consistency with syslog.h, but I can change them.

MrAnno:
The consistency with `syslog.h` sounds good to me. Thanks.

gaborznagy:
After 3115 is merged, I've rebased the banrch and resolved the conflicts.

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please test branch=json-parser-prefix;

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please test branch=json-parser-prefix;

gaborznagy:
(I had to rebase the test branch as well :) )

kira-syslogng:
Build SUCCESS

PR #3129, opened 2020-02-13
Add test for persist_state_move_entry()

MrAnno:
I've used `persist_state_move_entry()` in #2845, and haven't found any test for this method.

The PR also moves the test closer to the persist-state unit: `tests/unit/test_persist_state.c` → `lib/tests/test_persist_state.c`.

The license is changed from `GPL-2.0-or-later` to `LGPL-2.1-or-later`, as it tests a unit with the same license. AFAIK (checked the copyright policy file) this was a "Balabit-only" GPL code, so I think I'm allowed to do this.

No news file entry required.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3133, opened 2020-02-19
Affile, pipe time reap

Kokan:
According to the documentation the current `time-reap()` option describes the amount of seconds to wait before closing the destination file. The time between now and the last message to save
to a specific file (it can be different in case of templated path).

The current implementation schedule a timer at every `time-reap()` seconds and checks if there are messages in the queue to save into a file. The timer and the last message has no correlation.

This PR aims to replace the current behaviour with the one in the documentation[1]. The implementation taps into an already existing functionality of `LogWriter`'s `idle_timer`. A `LogProtoClient` could indicate a timeout value within its `_prepare` method, after that timer expires (if no message is sent till that time).

The timeout option should be implemented for multiple `LogProtoClient`:
* `lib/logproto/logproto-text-client` for pipe
* `modules/affile/logproto-file-writer` for file

The `LogProtoTextClient` is a general implementation that could be used for other destinations, currently it is needed for the `pipe` implementation.

[1] https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.25/administration-guide/34#TOPIC-1349418

### DIFF
--- a/modules/affile/affile-dest.c
+++ b/modules/affile/affile-dest.c
@@ -391,6 +391,13 @@ affile_dd_set_time_reap(LogDriver *s, gint time_reap)
 {
   AFFileDestDriver *self = (AFFileDestDriver *) s;
 
+  log_proto_client_options_set_timeout(&self->writer_options.proto_options.super, time_reap);
+}
+
+static gint
+affile_dd_get_time_reap(AFFileDestDriver *self)
+{
+  return log_proto_client_options_get_timeout(&self->writer_options.proto_options.super);
 }
 
 static inline const gchar *
@@ -471,6 +478,9 @@ affile_dd_init(LogPipe *s)
   file_opener_options_init(&self->file_opener_options, cfg);
   file_opener_set_options(self->file_opener, &self->file_opener_options);
   log_writer_options_init(&self->writer_options, cfg, 0);
+
+  if (affile_dd_get_time_reap(self) == -1)
+    affile_dd_set_time_reap(&self->super.super, cfg->time_reap);
 
   if (self->filename_is_a_template)
     {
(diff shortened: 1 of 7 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
I seem to have missed the patch that converted us to use a per-file timer vs. the earlier implementation that was using a single timer. Looking at the changes, I was the one who has implemented this in 2010... Time flies :)

In any case, I was worried that we are changing the code to use a per-file timer instead of a per-destination timer (e.g. many timers instead of a single one), but since that ship has sailed a long time ago, this worry is unwarranted.

So since we are using many timers already, the LogProtoClient interface already has support for timers (which it didn't until recently) and this change makes the code easier to understand, this has my approval, even though I was about to question it. Documentation is always easier to fix than code, especially if the behaviour is otherwise ok.

modules/affile/affile-dest.c
```diff
@@ -325,10 +302,14 @@ affile_dw_free(LogPipe *s)
...
 affile_dw_notify(LogPipe *s, gint notify_code, gpointer user_data)
 {
+  AFFileDestWriter *self = (AFFileDestWriter *)s;
   switch(notify_code)
     {
     case NC_REOPEN_REQUIRED:
-      affile_dw_reopen((AFFileDestWriter *)s);
+      affile_dw_reopen(self);
+      break;
+    case NC_CLOSE:
```
MrAnno: Just one note:
`NC_CLOSE` is coming through `LogWriter`, which produces the following internal message:

```
Destination timeout has elapsed, closing connection;
Destination timed out, reaping;
```

The word `connection` was a bad choice of mine, we can't talk about connections on the `LogWriter`-level, and now we actually have an implementation that is file-based.

I'd change this to something else, for example:
`Destination timeout has elapsed, closing;`, `initiating closure`, `closing fd`, `closing writer`.

MrAnno:
@kira-syslogng do perftest

kira-syslogng:

PR #3134, opened 2020-02-19
Add python-home configure option

MrAnno:
```
--with-python3-home=path Use path as a hard-coded Python home directory that
can not be overwritten by the PYTHONHOME environment variable.
```

This can be useful when a Python interpreter is bundled with syslog-ng.
Relocation is supported, for example: `--with-python3-home='${exec_prefix}'`

### DIFF
--- a/modules/python/python-plugin.c
+++ b/modules/python/python-plugin.c
@@ -129,7 +129,7 @@ const ModuleInfo module_info =
 {
   .canonical_name = "python",
   .version = SYSLOG_NG_VERSION,
-  .description = "The python ("PYTHON_MODULE_VERSION") module provides Python scripted destination support for syslog-ng.",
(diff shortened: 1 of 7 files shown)


kira-syslogng:
Build SUCCESS

modules/python/python-plugin.c
```diff
@@ -40,6 +40,7 @@
 #include "reloc.h"
 
 #include <stdlib.h>
+#include <string.h>
```
furiel: Is this header needed?

lbudai:
@MrAnno : thanks!

modules/python/python-plugin.c
MrAnno: It's in the wrong place, I'm moving it to `compat-python-v{2,3}.c`. Thanks.

MrAnno: Oh too late, I'll fix it in another PR then :smile:

PR #3135, opened 2020-02-19
fix gethostbyname detection

szemere:
The detection of the location of the `gethostbyname` function was faulty. (except on Solaris) `gethostbyname` can be found under `libc`.

This check caused unnecessary linking against `libnsl`, now it only links if gethostbyname can be found under libnsl.

This unnecessary linking was hidden from us for several reasons:

- On Debian-based distributions libnsl is part of the libc package, so it was loaded without any problem.
- If the linker is using the `--as-needed` option (turned on by default on Debian-based distributions), than `libnsl` was removed during the final linking since nothing uses it inside Syslog-ng.


kira-syslogng:
Build SUCCESS

szemere:
@bazsi: can you review this please? I am thinking about some legacy reason behind this detection. (We found Solaris, where gethostbyname was implemented in libnsl.)
@czanik: With this change, CentOS 8 should work out of the box.

@furiel, @Kokan: thank you for all the help! It was again a good investigation, including compilers, linkers and dependency handling.

kira-syslogng:
Build SUCCESS

PR #3138, opened 2020-02-20
news: add missing newsfile entries

furiel:


kira-syslogng:
Build FAILURE

news/bugfix-3083.md
```diff
@@ -0,0 +1,4 @@
+http: fix curl ssl version
```
alltilla: I would aim for one-liners, like:
> `http`: Fixed handling of `ssl-version()` option, which was ignored before.

Ultimately, the users don't care about the implementation being with `cURL` or anything else, so I would miss that completely.

My comment is open for arguments :)

kira-syslogng:
Build FAILURE

news/bugfix-3083.md
furiel: Thanks. I used the oneliner. However, not all options were ignored. So I think it is useful to enumerate those that were ignored.

furiel:

kira-syslogng:
Build SUCCESS

PR #3104, opened 2020-01-30
[WIP] github-actions: add test-python3-support job

alltilla:
~~It will be green, after we merge #3102, and I rebase my branch.~~
~~Should be green now.~~
It is green now.

Signed-off-by: Attila Szakacs <attila.szakacs@oneidentity.com>

PR #3141, opened 2020-02-21
travis: add bionic-python3 job

alltilla:
With this job we can test `--with-python=3`. Additional extra, that it runs on `bionic`, and it seems, that the only thing that needed extra care is to use an older `java` (I installed 1.8 in this case).

There are 2 things to fix, both are disabled right now:
- [x] The platform is `bionic`, and we get some packages from [OBS](https://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_18.04/amd64/). If we use `--enable-mongodb`, we use `libmongoc-dev-1.15`, which has deprecated `mongoc_client_get_server_status()`, and we cannot build because of that.
- [x] `make func-test` currently does not work, as it wants to run with `python3`, too, but the syntax is only valid with `python2`. @mitzkia is currently working on the port (#3144)

There was a small bug, which showed itself now, when building with fresher `gcc`, the PR has he fix for it, too.

This PR is an alternative of #3104, this one uses TravisCI.

### DIFF
--- a/news/bugfix-3141.md
+++ b/news/bugfix-3141.md
@@ -1,0 +1,1 @@
+`secret-storage`: Fixed some cases, where diagnostical logs were truncated.
--- a/.travis.yml
+++ b/.travis.yml
@@ -244,6 +244,91 @@ matrix:
         - make --keep-going -j $(nproc) all install
         - ctest -j $(nproc) --output-on-failure
 
+    - env: B=bionic-python3
+      compiler: gcc
+      dist: bionic
+      sudo: required
+      addons:
+        apt:
+          sources:
+            - sourceline: 'deb http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_18.04 ./'
+              key_url: 'http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/xUbuntu_18.04/Release.key'
+          packages:
+            - autoconf-archive
+            - bison
+            - docbook-xsl
+            - flex
+            - gradle
+            - libcap-dev
+            - libdbd-sqlite3
+            - libdbi0-dev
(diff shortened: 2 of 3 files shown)


news/bugfix-3141.md
```diff
@@ -0,0 +1 @@
+`secret-storage`: Fixed some cases, where log messages were truncated.
```
alltilla: For the reviewers: Is this clear, or misinterpret-able?

kira-syslogng:
Build FAILURE

alltilla:

lbudai:
mongodb support: in the near future we should have to follow the libmongoc changes, but I think it is better to keep focus on Python3. mongodb is tested by other jobs, and this job is about python3 testing... so this should  go first, then in a separate PR we will support newer libmongoc versions.

kira-syslogng:
Build FAILURE

alltilla:

kira-syslogng:
Build SUCCESS

news/bugfix-3141.md
furiel: I would say diagnostical logs instead of log messages, just to be sure.

alltilla:
wip until #3144 gets merged. I will rebase and remove the comment `#make VERBOSE=1 func-test # micek atportolja`

news/bugfix-3141.md
alltilla: Thanks, fixed! :)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3145, opened 2020-02-24
tls: add ca-file() option

### LINKED ISSUE #3111: FR: An option to use the OS's cert bundle for validating certificates
## Description of the problem
Using TLS for transporting logs around is good practice, but the configuration for TLS leaves lots of room for improvement. If the server side has a well known, good certificate that can be validated via the OS's installed trusted certificate bundle, the option should be present to allow this.

## Proposed solution
Right now, not only do you have to manually populate certificates into the `ca-dir` - but you also have to have non-obvious naming, or use the c_rehash utility - which is not part of the standard openssl install on most distros.

An option should be added to use the system trusted bundle to validate a presented certificate.

alltilla:
With this option, users can specify a CA-file, which holds the certificates in a bundle file.
e.g.: `ca-file("/etc/pki/tls/certs/ca-bundle.crt")`

Can be used parallelly with the `ca-dir()` option.

Fixes #3111

### DIFF
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -49,6 +49,7 @@ struct _TLSContext
   gchar *pkcs12_file;
   gchar *ca_dir;
   gchar *crl_dir;
+  gchar *ca_file;
   gchar *cipher_suite;
   gchar *ecdh_curve_list;
   gchar *sni;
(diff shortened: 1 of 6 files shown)


kira-syslogng:
Build FAILURE

alltilla:

kira-syslogng:
Build SUCCESS

alltilla:
To further improve this, we can use the following: #3086

PR #3151, opened 2020-02-26
news: add developer notes for #3038 [skip-ci]

szemere:
Add missing developer notes for #3038.


szemere:
Note: we have to merge https://github.com/syslog-ng/syslog-ng/pull/3140 first. (This commit should be there, but we don't want to dismiss the current approves.)

PR #3155, opened 2020-02-28
dbparser: introduce ${path_xmllint} to look up xmllint

Kokan:
By default this path is empty, that implies using the system default path.
It provides a possiblity at build time to change the path from which xmllint is started.


kira-syslogng:
Build SUCCESS

Kokan:
I opened it from a wrong repository.

PR #3157, opened 2020-02-28
LightReportDir: Support to define report dir with any path

mitzkia:
Before this patch we can not started Light with such report dir (the given directory has different base  than current directoy): `/tmp`
```consoe
python -m pytest -lvs functional_tests/source_drivers/file_source/test_acceptance.py --installdir=/install --reports /tmp/
```

we have got such an exception:
```
        n = len(to_abs_parts)
        cf = self._flavour.casefold_parts
        if (root or drv) if n == 0 else cf(abs_parts[:n]) != cf(to_abs_parts):
            formatted = self._format_parsed_parts(to_drv, to_root, to_parts)
            raise ValueError("{0!r} does not start with {1!r}"
>                            .format(str(self), str(formatted)))
E           ValueError: '/tmp/test_acceptance_with_ten_logs_' does not start with '/source/tests/python_functional'
```

After this patch it will work correctly, this report dir will work also: `../../../tmp`

### DIFF
--- a/news/developer-note-3157.md
+++ b/news/developer-note-3157.md
@@ -1,0 +1,3 @@
+LightReports: Support to relocate reports dir other than current base dir
+This will work:
+`python -m pytest -lvs functional_tests/source_drivers/file_source/test_acceptance.py --installdir=/install --reports /tmp/`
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3164, opened 2020-03-03
python: Revert #3153

alltilla:
We decided to keep the `bytes` type data.

### DIFF
--- a/tests/functional/sngtestmod.py
+++ b/tests/functional/sngtestmod.py
@@ -39,6 +39,7 @@ class DestTest(object):
 
     def send(self, msg):
         with open('test-python.log', 'a') as f:
+            msg = dict(map(lambda entry: (entry[0], entry[1].decode()), msg.items()))
             f.write('{DATE} {HOST} {MSGHDR}{MSG}\n'.format(**msg))
 
         return True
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

tests/functional/sngtestmod.py
```diff
@@ -39,6 +39,8 @@ def is_open(self):
 
     def send(self, msg):
```
furiel: Please remove the changenote as well.

tests/functional/sngtestmod.py
```diff
@@ -39,6 +39,8 @@ def is_open(self):
...
         with open('test-python.log', 'a') as f:
+            if (isinstance(msg['MSG'], bytes)):
```
furiel: Why do we need this check?

tests/functional/sngtestmod.py
alltilla: The NEWS file is already generated and formatted, and I removed this change from it.

tests/functional/sngtestmod.py
alltilla: I wanted to branch the `python2` from the `python3`, but yes, indeed it is not needed, as both of them use bytes... My bad.

kira-syslogng:
Build SUCCESS

PR #3171, opened 2020-03-11
python: persist api

furiel:
This patchset exposes persist api into python code.

This is the continuation of #3016, towards the road of having proper acknowledgement support in python.

Similar to #3016, this PR neither has any usage on its own, until bookmark handling is implemented in the future.

Features:
- A `Persist` class is importable from `syslogng` module
- The object looks like a dictionary: users can store values behind keys. The difference is it is backed by an entry in the persist file, so data is persisted between restarts.
- If no persist file created yet, `defaults` constructor parameter specifies can specify values for keys.
- Handles integers and strings.
- Iterable

Limitations:
- This patchset does not support editing multiple keys in the same transaction. This means if the persist data is complex and the parts need to be mutated together: like (`last-blob` + offset`), partial update may happen. As a workaround, users need to serialize the complex data into a single entry, and while using/updating, they need to deserialize and re-serialize. I plan to support this use case later.
- Only a limited size of data can be added at a time. The length of the key+value must be below 4K, I do not have the exact values for now. As this module is intended for bookmarking, it does not make sense to store large values anyway. To store large number of local data securely, users need to use local database or other mechanism.
The limitation comes from the persist file in c, it can be only increased 4K at a time. This limitation can be workarounded if really needed: by storing only a small portion of the value, and incrementally append the rest of the data in small chunks.
- In python2, there are two integral types: integer and long (for large numbers). In python3, there is only one integral type: long. The current implementation uses only long. For **python2** users, this means an integer is automatically converted to long.
- Explicit key deletion is not supported. Though there is a global cleanup mechanism in persist. If a persist entry is not referred between two restarts, the entry is automatically removed.

Examples:

The code below prints five numbers, each restart/reload. Continuing from the last number of the previous execution.

```python
from syslogng import Persist, LogSource
class RangeSource(LogSource):
    def init(self, options):
        self.persist = Persist("persist_name", defaults={"bookmark" : 1})
        return True
    def run(self):
        first_item = self.persist["bookmark"]
        for i in range(first_item, first_item + 5):
            print("Item: {}".format(i))
            self.persist["bookmark"] = i + 1
    def request_exit(self):
        pass
```

You can iterate through keys:
```python
persist = Persist("persist_name", defaults={"foo" : "bar"})
for i in persist:
    print(i)
```

Developer notes:
- The Persist entry always stores data as string, no matter if integer or string is specified. Integers are serialized trivially. To differentiate, we also store the type of the entry, so that proper type could be returned during lookup.
- The subkeys are stored on the global level of the persist. Users provide a basename and a subkey. The resulting entry will be accessed using a key builder: `basename##subkey`.

### DIFF
--- a/modules/python/python-persist.c
+++ b/modules/python/python-persist.c
@@ -46,8 +46,11 @@ typedef enum
 typedef struct
 {
   guint8 type;
-  gchar data[];
+  gchar data[0];
 } Entry;
+
+/* Ensure there is no padding between type and data */
+G_STATIC_ASSERT(offsetof(Entry, data) == sizeof(guint8) + offsetof(Entry, type));
 
 static PyObject *
 entry_to_pyobject(guint8 type, gchar *value)
--- a/modules/python/python-persist.c
+++ b/modules/python/python-persist.c
@@ -35,6 +35,7 @@ typedef enum
 {
   ENTRY_TYPE_STRING,
   ENTRY_TYPE_LONG,
+  ENTRY_TYPE_BYTES,
   ENTRY_TYPE_MAX
 } EntryType;
 
@@ -53,6 +54,8 @@ entry_to_pyobject(guint8 type, gchar *value)
       return _py_string_from_string(value, -1);
     case ENTRY_TYPE_LONG:
       return PyLong_FromString(value, NULL, 10);
+    case ENTRY_TYPE_BYTES:
+      return PyBytes_FromString(value);
     default:
       g_assert_not_reached();
     }
@@ -253,11 +256,11 @@ _allocate_persist_entry(PersistState *persist_state, const gchar *key, const gch
 static gchar *
 _serialize(guint8 type, PyObject *v)
 {
-  if (type == ENTRY_TYPE_STRING)
+  switch (type)
     {
+    case ENTRY_TYPE_STRING:
       return g_strdup(_py_get_string_as_string(v));
-    }
-  else if (type == ENTRY_TYPE_LONG)
+    case ENTRY_TYPE_LONG:
     {
       PyObject *as_str = PyObject_Str(v);
       g_assert(as_str);
@@ -265,8 +268,11 @@ _serialize(guint8 type, PyObject *v)
       Py_DECREF(as_str);
       return result;
     }
-  else
-    g_assert_not_reached();
+    case ENTRY_TYPE_BYTES:
+      return g_strdup(PyBytes_AsString(v));
+    default:
+      g_assert_not_reached();
+    }
 }
 
 static gboolean
@@ -344,9 +350,11 @@ _py_persist_type_set(PyObject *o, PyObject *k, PyObject *v)
     type = ENTRY_TYPE_STRING;
   else if (py_object_is_integer(v))
     type = ENTRY_TYPE_LONG;
+  else if (PyBytes_Check(v))
+    type = ENTRY_TYPE_BYTES;
   else
     {
-      PyErr_SetString(PyExc_TypeError, "Value must be either string or integer");
+      PyErr_SetString(PyExc_TypeError, "Value must be either string, integer or bytes");
       return -1;
     }
 
--- a/modules/python/python-persist.c
+++ b/modules/python/python-persist.c
@@ -177,6 +177,132 @@ py_persist_type_members[] =
   {NULL}
 };
 
+static gchar *
+_build_key(PyPersist *self, const gchar *key)
+{
+  return g_strdup_printf("%s##%s", self->persist_name, key);
+}
+
+static gchar *
+_lookup_entry(PyPersist *self, const gchar *key)
+{
+  gchar *query_key = _build_key(self, key);
+
+  gsize size;
+  guint8 version;
+  PersistEntryHandle handle = persist_state_lookup_entry(self->persist_state, query_key, &size, &version);
+  if (!handle)
+    {
+      PyErr_Format(PyExc_KeyError, "Persist has no such key: %s", key);
+      g_free(query_key);
+      return NULL;
+    }
+
+  const gchar *data = persist_state_map_entry(self->persist_state, handle);
+  gchar *copy = g_strdup((gchar *)data);
+  persist_state_unmap_entry(self->persist_state, handle);
+
+  g_free(query_key);
+
+  return copy;
+};
+
+static PersistEntryHandle
+_allocate_persist_entry(PersistState *persist_state, const gchar *key, const gchar *value, gsize value_len)
+{
+  gsize size;
+  guint8 version;
+  PersistEntryHandle handle = persist_state_lookup_entry(persist_state, key, &size, &version);
+  if (handle && size >= value_len)
+    return handle;
+
+  return persist_state_alloc_entry(persist_state, key, value_len);
+}
+
+static gboolean
+_store_entry(PyPersist *self, const gchar *key, const gchar *value)
+{
+  gchar *query_key = _build_key(self, key);
+  gsize value_len = strlen(value);
+
+  PersistEntryHandle handle = _allocate_persist_entry(self->persist_state, query_key, value, value_len);
+  if (!handle)
+    {
+      g_free(query_key);
+      return FALSE;
+    }
+
+  gchar *data = persist_state_map_entry(self->persist_state, handle);
+  memcpy(data, value, value_len);
+  persist_state_unmap_entry(self->persist_state, handle);
+
+  g_free(query_key);
+
+  return TRUE;
+}
+
+static PyObject *
+_py_persist_type_get(PyObject *o, PyObject *key)
+{
+  PyPersist *self = (PyPersist *)o;
+
+  if (!_py_is_string(key))
+    {
+      PyErr_SetString(PyExc_TypeError, "key is not a string object");
+      return NULL;
+    }
+
+  const gchar *name = _py_get_string_as_string(key);
+  gchar *value = _lookup_entry(self, name);
+
+  if (!value)
+    {
+      PyErr_Format(PyExc_KeyError, "No such name-value pair %s", name);
+      return NULL;
+    }
+
+  PyObject *py_value =  _py_string_from_string(value, -1);
+  g_free(value);
+  return py_value;
+}
+
+static int
+_py_persist_type_set(PyObject *o, PyObject *k, PyObject *v)
+{
+  PyPersist *self = (PyPersist *)o;
+
+  if (!_py_is_string(k))
+    {
+      PyErr_SetString(PyExc_TypeError, "key is not a string object");
+      return -1;
+    }
+
+  if (!_py_is_string(v))
+    {
+      PyErr_SetString(PyExc_TypeError, "value is not a string object");
+      return -1;
+    }
+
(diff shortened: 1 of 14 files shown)


furiel:
@kira-syslogng retest this please test branch=furiel-persist-api;

kira-syslogng:
Build SUCCESS

furiel:
@kira-syslogng retest this please test branch=furiel-persist-api;

kira-syslogng:
Build SUCCESS

lbudai:
@furiel : thanks for the changes, I'm trying to review the PR soon

modules/python/python-persist.c
```diff
@@ -22,7 +22,41 @@
...
+  gchar data[];
+} Entry;
+
+static PyObject *
+entry_to_pyobject(guint8 type, gchar *value)
+{
+  switch (type)
+    {
+    case ENTRY_TYPE_STRING:
+      return _py_string_from_string(value, -1);
```
bazsi: We might need to support binary data here (bytes). 

At least I am not sure systemd for instance doesn't give us binary bookmarks.

modules/python/python-persist.c
```diff
@@ -127,3 +161,272 @@ python_format_persist_name(const LogPipe *p, const gchar *module, PythonPersistM
...
+    }
+
+  if (_py_is_string(v))
+    type = ENTRY_TYPE_STRING;
+  else if (py_object_is_integer(v))
+    type = ENTRY_TYPE_LONG;
+  else
+    {
+      PyErr_SetString(PyExc_RuntimeError, "Value must be either string or integer");
+      return -1;
```
bazsi: Maybe we could use a better exception type here. TypeError seems more appropriate.

modules/python/python-persist.c
```diff
@@ -127,3 +161,272 @@ python_format_persist_name(const LogPipe *p, const gchar *module, PythonPersistM
...
+    }
+
+  const gchar *key = _py_get_string_as_string(k);
+
+  if (!_store_entry(self, key, type, v))
+    {
+      PyErr_SetString(PyExc_RuntimeError, "value could not be stored");
```
bazsi: We are leaking the key object as store_entry doesn't seem to consume (and probably shouldn't) it in case of an error.

bazsi: Instead of runtimeerror, IOError?

I cant remember if we are using very specific exception types elsewhere in our python binding, because we might want to do that here too.

modules/python/python-persist.c
```diff
@@ -127,3 +161,272 @@ python_format_persist_name(const LogPipe *p, const gchar *module, PythonPersistM
...
+      return NULL;
+    }
+
+  const gchar *name = _py_get_string_as_string(key);
+  guint8 type;
+  gchar *value = _lookup_entry(self, name, &type);
+
+  if (!value)
+    {
+      PyErr_Format(PyExc_KeyError, "No such name-value pair %s", name);
```
bazsi: Name is leaked

modules/python/python-persist.c
```diff
@@ -127,3 +161,272 @@ python_format_persist_name(const LogPipe *p, const gchar *module, PythonPersistM
...
+      return NULL;
+    }
+
+  if (type >= ENTRY_TYPE_MAX)
+    {
+      PyErr_Format(PyExc_RuntimeError, "Unknown data type: %d", (gint)type);
```
bazsi: Also here.

modules/python/python-persist.c
```diff
@@ -127,3 +161,272 @@ python_format_persist_name(const LogPipe *p, const gchar *module, PythonPersistM
...
+static gchar *
+_serialize(guint8 type, PyObject *v)
+{
+  if (type == ENTRY_TYPE_STRING)
+    {
+      return g_strdup(_py_get_string_as_string(v));
+    }
+  else if (type == ENTRY_TYPE_LONG)
+    {
+      PyObject *as_str = PyObject_Str(v);
```
bazsi: The str() function may fail as it calls the __str__ method.

modules/python/python-persist.c
```diff
@@ -127,3 +161,272 @@ python_format_persist_name(const LogPipe *p, const gchar *module, PythonPersistM
...
 }
+
+static PyObject *
+_persist_type_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
+{
+  PyPersist *self =(PyPersist *)type->tp_alloc(type, 0);
+
+  GlobalConfig *cfg = main_loop_get_current_config(main_loop_get_instance());
+  self->persist_state = cfg->state;
+
```
bazsi: Is there a reason this is in new() and not in init()?

modules/python/python-persist.c
furiel: You are right. I added a new patch to the top. From now, we can store string, number and bytes.

modules/python/python-persist.c
furiel: Right! Changed to TypeError.

modules/python/python-persist.c
furiel: Fixed the leak, and changed to IOError.

modules/python/python-persist.c
furiel: Technically, this should not fail. At this point we are sure `v` is a number. I added a g_assert() just in case.

modules/python/python-persist.c
furiel: I moved it ot init. As tp_new is not needed anymore, I changed it to `PyType_GenericNew`.

modules/python/python-persist.c
furiel: Name is stored either in a scratchbuffers, or inside a python object. It should not be freed.

furiel:
@kira-syslogng retest this please test branch=furiel-persist-api;

furiel:
Hmm there is a crash during reload. Adding wip flag while investigating.

kira-syslogng:
Build SUCCESS

modules/python/python-persist.c
bazsi: you can derive a new class from the builtin int class and override its __str__ method. exception from _any_ python function is possible.

the addition of the assertion is good enough for me. btw: but you can also get the value out of the int/long object and format it yourself (e.g. PyLong_AsLong())

modules/python/python-persist.c
bazsi: ok

modules/python/python-persist.c
bazsi: it was the same as the previous comment, so resolving this one as well.

furiel:
@bazsi et al. I ran into a little trouble during reload. I pushed a new patch to the top, that modifies mainloop in my favor. Could you have a look? I am interested in your opinion.

For persist api to work, I need the persist state object.

Instead of properly getting the GlobalConfig instance from a logpipe_init and passing around (as every drivers do), I wanted to use a shortcut: with `main_loop_get_current_config(main_loop_get_instance())`. From that, I can have the persist state: `cfg->state` directly from the python module. 

The alternatively, I would need to create a python representation of the GlobalConfig object, and Persist class should receive it as a constructor parameter. How do users get hold of such object? I could expose GlobalConfig into PythonDestination, PythonParser, Source and Fetcher, then users could access it through self. Something like this (inside the init method of PythonDestination). `cfg = self.cfg; persist = Persist("name", cfg)`

I feel the alternative inconvenient, because users need to know of a GlobalConfigClass in python, which is not useful for them. Still need to pass around. From implementation point of view, exposing cfg also a little boilerplate. On the other hand, persist state is a singleton. It does not change during reload.

Just a minor argument, because it is theoretical for now. We were talking about plans to use persist api in http python header. This alternative would not work there, and probably through any signal-slot implementation in general. I do not know how we could pass globalconfig for those plugins.

The patch I created does not steal the persist state from current_configuration, only after the log pipes are initialized. This fixes the reload problem.

furiel:
@kira-syslogng retest this please test branch=furiel-persist-api; test=functions/source-drivers/python-source-driver;

furiel:
@kira-syslogng retest this please test branch=furiel-persist-api;

furiel:
@kira-syslogng retest this please test branch=furiel-persist-api;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=furiel-persist-api;

kira-syslogng:
Build FAILURE

furiel:
@kira-syslogng retest this please test branch=furiel-persist-api;

kira-syslogng:
Build SUCCESS

furiel:
Two changes:
- Added a dummy entry for the master key: that stores only the version. Later if we want to change the persist layout, we can rely on that.

- Instead of getting persist_state from mainloop, I am injecting persist_state directly into the `_syslogng` module in a `PyCapsule`. The injection happens with the `module_config` framework. See the patch: `python: store state in _syslogng instead of getting from mainloop`.

Removing wip flag.

furiel:
Just a note: I will edit the description with this too.

One feature is missing: editing multiple keys in the same transaction. When using persist from c, it is relatively simple: you map a single entry, change the serialized data, then unmap it. Now the persist data is broken into multiple entries, we need to find a way to wait with the transaction, until all entries are updated. I will add this one later, because I cannot predict its complexity now.

lbudai:
@furiel : nice work!

modules/python/python-persist.c
```diff
@@ -21,8 +21,49 @@
...
+  ENTRY_TYPE_STRING,
+  ENTRY_TYPE_LONG,
+  ENTRY_TYPE_BYTES,
+  ENTRY_TYPE_MAX
+} EntryType;
+
+typedef struct
+{
+  guint8 type;
+  gchar data[];
```
bazsi, line 49: I think this may not be portable to older c compilers. doing it as "gchar data[0]" on the other hand is portable and shouldn't be more difficult implementation wise.

modules/python/python-persist.c
```diff
@@ -127,3 +168,330 @@ python_format_persist_name(const LogPipe *p, const gchar *module, PythonPersistM
...
+    {
+      g_free(value);
+      g_free(query_key);
+      return FALSE;
+    }
+
+  Entry *entry = persist_state_map_entry(self->persist_state, handle);
+  entry->type = type;
+  strcpy(entry->data, value);
+  persist_state_unmap_entry(self->persist_state, handle);
```
bazsi, line 356: I think we should assert on the fact that Entry does not have padding between type and value. I don't think there would be an issue ever with the current declaration however it is also a good way to document the assumption.

e.g. g_assert(sizeof(*entry) - sizeof(entry->data) == sizeof(entry->type));

furiel:
Sure, thanks for the review. I fix these before merge.

furiel:
@kira-syslogng retest this please test branch=furiel-persist-api;

modules/python/python-persist.c
furiel: Thanks, fixed. I ended up with a `G_STATIC_ASSERT`

modules/python/python-persist.c
furiel: Thanks, fixed.

kira-syslogng:
Build SUCCESS

PR #3175, opened 2020-03-13
format-json: fix embedded zeroes, fix github actions

furiel:
Prior to 2.64.1, `g_utf8_get_char_validated()` in glib falsely identified embedded zeroes as valid utf8 characters. As a result, format json printed the embedded zeroes as `\u0000` instead of `\x00`.

The macOS ci job fails due to this change (test_format_json). The same unit test failure happens in archlinux. https://github.com/GNOME/glib/commit/1963821a57584b4674c20895e8a5adccd2d9effd

This is a breaking change as the output of format-json may change. Alternatively, I could maintain the original behavior, and make `g_utf8_get_char_validated()` the original value in the compat layer. Personally, I agree with glib here. That's why I am proposing this version.

### DIFF
--- a/lib/compat/glib.c
+++ b/lib/compat/glib.c
@@ -315,3 +315,18 @@ slng_g_hash_table_insert(GHashTable *hash_table, gpointer key, gpointer value)
   return exists;
 }
 #endif
+
+
+#if !GLIB_CHECK_VERSION(2, 64, 0)
+gunichar
+g_utf8_get_char_validated_fixed(const gchar *p, gssize max_len)
+{
+  // https://github.com/GNOME/glib/commit/1963821a57584b4674c20895e8a5adccd2d9effd
+
+#undef g_utf8_get_char_validated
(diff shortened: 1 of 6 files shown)


lib/compat/glib.c
```diff
@@ -315,3 +315,18 @@ slng_g_hash_table_insert(GHashTable *hash_table, gpointer key, gpointer value)
...
 #endif
+
+
+#if !GLIB_CHECK_VERSION(2, 64, 0)
+gunichar
+g_utf8_get_char_validated_fixed(const gchar *p, gssize max_len)
+{
+  // https://github.com/GNOME/glib/commit/1963821a57584b4674c20895e8a5adccd2d9effd
+
+#undef g_utf8_get_char_validated
```
MrAnno: Tricky.

MrAnno:

kira-syslogng:
Build SUCCESS

PR #3180, opened 2020-03-18
configure.ac: fix --with-net-snmp

patrickdepinguin:
The option '--with-net-snmp' did not work, because later checks were using
the variable '$with_netsnmp' rather than '$with_net_snmp'.
As a result, the net-snmp-config option was always searched in the PATH.

There are two possible solutions:
- make the option '--with-netsnmp'
- change the internal variables

Opt for the second option so the user-visible options remain the same.

Signed-off-by: Thomas De Schampheleire <thomas.de_schampheleire@nokia.com>

(Note: I did not add a news item as I think very few people actually use this, and it was broken for almost a year, let me know if you would prefer one)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -151,7 +151,7 @@ AC_ARG_WITH(libnet,
 AC_ARG_WITH(net-snmp,
    [  --with-net-snmp=path      use path to net-snmp-config script],
    ,
-   [with_netsnmp=""])
+   [with_net_snmp=""])
 
 AC_ARG_WITH(pidfile-dir,
    [  --with-pidfile-dir=path   Use path as the directory for storing pidfiles],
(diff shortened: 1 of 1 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

furiel:
Thank you for the fix!

I would prefer to see a changenote . Users might have not used it either because they were unaware of this option, in which case advertizing would help. Or they workarounded it, in which case they may be interested in the change.

Either way, I think it worths a simple entry in the packaging category.

patrickdepinguin:
Thanks!

Issue #3183, opened 2020-03-20
Light: snmp test fails inside syslog-ng-devshell docker image


alltilla:
# syslog-ng
## Version of syslog-ng
Compiled from source (master: 531d70bfbcf19a97a5278f324d141386d568ce7f)

## Platform
 * `docker` image: `balabit/syslog-ng-devshell:latest`
 * `docker` runs on `ubuntu-xenial` or `ubuntu-bionic`

# Issue
## Failure
`test_snmp_dest_v2_cisco_ios_trap` fails during `make pytest-check`.
```
_________________________________________________________________________________________ test_snmp_dest_v2_cisco_ios_trap __________________________________________________________________________________________

config = <src.syslog_ng_config.syslog_ng_config.SyslogNgConfig object at 0x7f4e0ab9cd50>, syslog_ng = <src.syslog_ng.syslog_ng.SyslogNg object at 0x7f4e0ab9ca50>
snmptrapd = <src.helpers.snmptrapd.conftest.SNMPtrapd object at 0x7f4e0ab9c910>

    @pytest.mark.snmp
    def test_snmp_dest_v2_cisco_ios_trap(config, syslog_ng, snmptrapd):
        # create source driver config
        file_source = config.create_file_source(file_name="input.log")
    
        # create destination driver config
        snmp_objs = (
            "'{}', 'Octetstring', '{}'".format(OID1, CISCO_FACILITY),
            "'{}', 'Integer', '{}'".format(OID2, CISCO_SEVERITY),
            "'{}', 'Octetstring', '{}'".format(OID3, CISCO_MESSAGE_TYPE),
            "'{}', 'Octetstring', '{}'".format(OID4, CISCO_MESSAGE),
            "'{}', 'Timeticks', '{}'".format(OID5, CISCO_TIMETICKS),
        )
    
        trap_obj = "'.{}','Objectid','{}'".format(OID6, TRAP_DATA)
    
        snmp_destination = config.create_snmp_destination(
            host="'127.0.0.1'",
            port=snmptrapd.get_port(),
            version="'v2c'",
            snmp_obj=snmp_objs,
            trap_obj=trap_obj,
        )
    
        config.create_logpath(statements=[file_source, snmp_destination])
        config.update_global_options(keep_hostname="yes")
        file_source.write_log(input_log)
    
        syslog_ng.start(config)
    
>       snmptrap_logs = snmptrapd.get_logs()[2].rstrip().split("\t")
[...]

PR #3185, opened 2020-03-20
Light: fix snmptrapd start

mitzkia:
Update:
This patch waits until snmptrapd log file initialized, and with this able to receive logs.

This is also a fix for:
#3183

### DIFF
--- a/tests/python_functional/src/helpers/snmptrapd/conftest.py
+++ b/tests/python_functional/src/helpers/snmptrapd/conftest.py
@@ -27,6 +27,7 @@ from pathlib2 import Path
 from psutil import TimeoutExpired
 
 import src.testcase_parameters.testcase_parameters as tc_parameters
+from src.common.blocking import wait_until_true
 from src.driver_io.file.file_io import FileIO
 from src.executors.process_executor import ProcessExecutor
 from src.syslog_ng_config.statements.destinations.destination_reader import DestinationReader
@@ -40,6 +41,12 @@ class SNMPtrapd(object):
         self.snmptrapd_log = Path(tc_parameters.WORKING_DIR, "snmptrapd_log")
         self.snmptrapd_stdout_path = Path(tc_parameters.WORKING_DIR, "snmptrapd_stdout")
         self.snmptrapd_stderr_path = Path(tc_parameters.WORKING_DIR, "snmptrapd_stderr")
+
+    def wait_for_snmptrapd_log_creation(self):
+        return self.snmptrapd_log.exists()
+
+    def wait_for_snmptrapd_startup(self):
+        return "NET-SNMP version" in self.snmptrapd_log.read_text()
 
     def start(self):
         if self.snmptrapd_proc is not None:
@@ -58,6 +65,8 @@ class SNMPtrapd(object):
             self.snmptrapd_stdout_path,
             self.snmptrapd_stderr_path,
         )
+        wait_until_true(self.wait_for_snmptrapd_log_creation)
+        wait_until_true(self.wait_for_snmptrapd_startup)
         return self.snmptrapd_proc.is_running()
 
     def stop(self):
--- a/news/developer-note-3185.md
+++ b/news/developer-note-3185.md
@@ -1,0 +1,1 @@
+LightSNMPTrapd: Wait until snmptrapd process able to write traps into output file


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

tests/python_functional/src/helpers/snmptrapd/conftest.py
```diff
@@ -44,8 +45,8 @@ def __init__(self, port):
     def start(self):
         if self.snmptrapd_proc is not None:
             return
-
-        self.snmptrapd_proc = ProcessExecutor().start(
+        process_executor = ProcessExecutor()
```
szemere:
```suggestion
        self.process_executor = ProcessExecutor()
```

tests/python_functional/src/helpers/snmptrapd/conftest.py
```diff
@@ -58,6 +59,7 @@ def start(self):
             self.snmptrapd_stdout_path,
             self.snmptrapd_stderr_path,
         )
+        wait_until_true(process_executor.is_process_status_sleeping)
         return self.snmptrapd_proc.is_running()
```
szemere:
```suggestion
```
def _is_process_sleeping(self):
   return self.process_executor.is_process_status_sleeping()
```
```

tests/python_functional/src/helpers/snmptrapd/conftest.py
```diff
@@ -58,6 +59,7 @@ def start(self):
             self.snmptrapd_stdout_path,
             self.snmptrapd_stderr_path,
         )
+        wait_until_true(process_executor.is_process_status_sleeping)
```
szemere:
```suggestion
        wait_until_true(self._is_process_sleeping)
```

bazsi:
Hmmm, process states are not meant to be used for this, a process is also
sleeping when it is preempted by another process.

bazsi:
Btw, snmptrapd supports systemd's notification mechanism, we could
implement a server for that, so we would know that snmptrapd is ready.
Alternative we could execute it with socket activation, which ensures that
the udp datagram is not lost between us sending it and snmptrapd starting
up.

mitzkia:
@bazsi Thanks for the comment. Maybe I was just lucky with checking the process state change. 
After re-checking my PR it seems does not correct, from another image somehow it still does not work (same error happens).

Do you think I should check some other state(s)? For example there is a CONN_NONE state for UDP connections, https://psutil.readthedocs.io/en/latest/#psutil.net_connections.

I like the idea to run snmptrapd via systemd (with socket activation).
I would have some questions and notices related to here.

One question from an overall view:
- It would be good to lay the foundations for the other supported services also.
Do you think our current image can "host" them?
I would say it would be clearer to run them from a separate image (test-services image).
Also because they will need a running systemd process, and our current images 
(balabit/syslog-ng-ubuntu-xenial, balabit/syslog-ng-devshell) does not run that.

One question from a specific SNMPTrapd view:
- For now 'snmptrapd' fixture will start process for every testcase modules, with this 
we will have got separate snmptrapd log files (in working dirs). With running snmptrapd service
from another docker container (or via systemd) probably we will loose this checking convenience.
Could we save this convenience somehow (one snmptrapd log per testcase, in an easily available place (if there are 2 containers))? How would you check the SNMPTrapd outgoing result?

Maybe the long-term change (systemd service based image) will be started in a new PR.

furiel:
@mitzkia ProcessExecutor already captures the stdout/stderr of the process. The fixture could just wait for data on these. Could that work? It would be simpler than systemd.

mitzkia:
@furiel You are right it could work (in this case we can drop internal redirect of snmptrapd output).
But I don't know could it be the final solution. As I understand correctly @bazsi, systemd (with socket activation) can prove us to run snmptrapd without waiting for any special state. When it starts it can surely accept input. I would wait @bazsi's answer here.

bazsi:
I wouldn't rely on systemd, rather I would implement the notification
mechanism as employed by systemd, this way snmptrapd can tell us when its
ready.

The stdout capture based idea might be somewhat simpler.

Iirc systemd keeps an fd open to the process and informs it using an
environment variable. Reading status updates should be pretty
straightforward. But I might be mistaken :)

mitzkia:
Thanks for the discussion, I have added a small code to wait until snmptrapd output log file is created, and snmptrapd has been initialized.

kira-syslogng:
Build SUCCESS

alltilla:
Validated, it fixes #3183.

kira-syslogng:
Build SUCCESS

PR #3187, opened 2020-03-21
python: fix Py_None reference counting

lbudai:
Py_RETURN_NONE is a macro that
 * increments ref counter for the Py_None object
 * return with Py_None


kira-syslogng:
Build SUCCESS

MrAnno:
Travis checks are not working, they seem to be under scheduled maintenance: https://www.traviscistatus.com/

Issue #3184, opened 2020-03-20
mongodb: cmake find_package() runs regardless -DENABLE_MONGODB=no


alltilla:
# syslog-ng
## Version of syslog-ng
Compiling from source (master: 531d70bfbcf19a97a5278f324d141386d568ce7f)

## Platform
 * `docker` image: `balabit/syslog-ng-devshell:latest`
 * `docker` runs on `ubuntu-xenial` or `ubuntu-bionic`

# Issue
## Failure
Even though I set `-DENABLE_MONGODB=no` to cmake, `find_package(libmongoc-1.0)` still runs, which is a problem on `ubuntu-bionic`, as the `libbson` (or `libmongoc`) package is bugged there.

I am getting the following error:
```
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/libbson-1.0/libbson-1.0-config.cmake:28 (message):
  File or directory /usr/lib/include/libbson-1.0 referenced by variable
  BSON_INCLUDE_DIRS does not exist !
Call Stack (most recent call first):
  /usr/lib/x86_64-linux-gnu/cmake/libbson-1.0/libbson-1.0-config.cmake:32 (set_and_check)
  /usr/lib/x86_64-linux-gnu/cmake/libmongoc-1.0/libmongoc-1.0-config.cmake:20 (find_package)
  modules/afmongodb/CMakeLists.txt:1 (find_package)


-- Configuring incomplete, errors occurred!
See also "/syslog-ng/build/CMakeFiles/CMakeOutput.log".
See also "/syslog-ng/build/CMakeFiles/CMakeError.log".
```
## Steps to reproduce
```
$ docker pull balabit/syslog-ng-devshell:latest
$ docker run -it balabit/syslog-ng-devshell:latest bash

# git clone https://github.com/syslog-ng/syslog-ng.git
# cd syslog-ng/
# mkdir build
# cd build
# cmake -DCMAKE_INSTALL_PREFIX=$HOME/install/syslog-ng -DENABLE_MONGODB=no ..
```

## Input and output logs
[cmake.log](https://github.com/syslog-ng/syslog-ng/files/4359377/cmake.log)
[CMakeOutput.log](https://github.com/syslog-ng/syslog-ng/files/4359379/CMakeOutput.log)
[CMakeError.log](https://github.com/syslog-ng/syslog-ng/files/4359380/CMakeError.log)

lbudai:
@alltilla : yes, it is a known issue :

[...]

PR #3188, opened 2020-03-21
afmongodb: cmake: do not build module when ENABLE_MONGODB=off

lbudai:
fixes: #3184

### DIFF
--- a/modules/afmongodb/CMakeLists.txt
+++ b/modules/afmongodb/CMakeLists.txt
@@ -2,7 +2,7 @@ find_package(libmongoc-1.0)
 
 module_switch(ENABLE_MONGODB "Enable mongodb destination driver" libmongoc-1.0_FOUND)
 
-if (libmongoc-1.0_FOUND)
+if (ENABLE_MONGODB)
 
   set(AFMONGODB_SOURCES
     "afmongodb.h"
--- a/modules/afmongodb/CMakeLists.txt
+++ b/modules/afmongodb/CMakeLists.txt
@@ -1,3 +1,7 @@
+if (ENABLE_MONGODB STREQUAL OFF)
+  return()
+endif()
+
 find_package(libmongoc-1.0)
 
 module_switch(ENABLE_MONGODB "Enable mongodb destination driver" libmongoc-1.0_FOUND)
--- a/news/bugfix-3188.md
+++ b/news/bugfix-3188.md
@@ -1,0 +1,3 @@
+afmongodb: do not build module when ENABLE_MONGODB=OFF
+
+fixes #3184


kira-syslogng:
Build FAILURE

alltilla:
Unfortunately, this did not fix #3184, which fails in `find_package(libmongoc-1.0)`.
I believe if `-DENABLE_MONGODB=no` is given, `find_package(libmongoc-1.0)` should not run.

Sadly, I saw the same (problematic) pattern in our other modules as well.

Thank you for the PR, anyways!

lbudai:
@altilla: I’d disagree: if you’ve installed a dependency that has a bug, remove it, or replace it from an alternative source (OBS).
I would not prepare for badly implemented find-packages.

My patch still needed: when someone don’t need the afmongodb module without this patch it cannot be done when the mongo-c-dev is installed.

MrAnno:

kira-syslogng:
Build SUCCESS

news/bugfix-3188.md
```diff
@@ -0,0 +1,3 @@
+afmongodb: do not build module when ENABLE_MONGODB=OFF
+
+fixes #3184
```
furiel: As this change does not fix #3184, please remove this line, and also the `fixes` line from the PR description. Apart from that, the change looks good to me. I agree this change is necessary, but we still need to find a way to deal with #3184

lbudai: Now, it fixes...

But I still disagree that we install a package that has a bug - as you want to disable the module that uses this package, why do we install it?
There are two possible solutions:
* install a package that contains the fix (from newest distro, if they fixed; or from OBS if no other choice)
 * do not install the package... 

Workarounding packaging bugs in this way is not the best, IMHO.

kira-syslogng:
Build SUCCESS

news/bugfix-3188.md
furiel: I agree. It would be preferable to install the patched package, especially after going through fixing it in obs. It is a half solution, because in the end, we will not test mongodb, which would be a goal for a ci job.

On the other hand, in general: if a module is disabled, it would be nice if cmake does not run detection for it, because it is unnecessary. Independently from this problem. Maybe `module_switch` might be improved that way, so that could be applied to all modules.

 We use autotools in that way. (At least in case of riemann which I quicky checked, there may be exceptions :))

```bash
if test "$enable_riemann" != "no"; then
 PKG_CHECK_MODULES(RIEMANN_CLIENT, riemann-client >= $LRC_MIN_VERSION,,riemann_found="no")
 if test "$riemann_found" = "no" && test "$enable_riemann" = "yes"; then
   AC_MSG_ERROR([Dependency for Riemann not found!])
 fi
 if test "$riemann_found" = "no"; then
   enable_riemann="no";
 else
   enable_riemann="yes";
 fi

 if test "$enable_riemann" = "yes"; then
```

With all that said, I would have been fine if we merged this PR with its previous version, find a better way to deal with mongodb, and leave the `disable->do-not-run-detection` as future improvement. But the current version is also fine, just we may look back later to do the generalization.

modules/afmongodb/CMakeLists.txt
```diff
@@ -1,8 +1,12 @@
+if (ENABLE_MONGODB STREQUAL OFF)
```
furiel: `ENABLE_MONGODB` should be evaluated as boolean. According to documentation `0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND` would equally evaluate to false.

You can just simply write
```
if (NOT ENABLE_MONGODB)
  return()
endif()
```

lbudai: This would be an incompatible change: currently when someone builds syslog-ng with cmake, it is not required to explicitly enable a module.

`(NOT ENABLE_MONGODB)` is `false` when you not define the `ENABLE_MONGODB` variable.

furiel: Oh ok I see.

PR #3193, opened 2020-03-23
Improve max-connections

szemere:
This PR contains changes, which helps the debugging process related to flow-control and connection handling.

1. Shows the number of source connections per driver in the stats output. Which can be different from the connections listed by the kernel, in case flow-control is activated.
2. Add source name and location (in the configuration) to the max connections reached error message, which makes debugging easier.

WIP:

- [x] writing developer notes
- [x] the new stats item has to be documented - in progress

### DIFF
--- a/modules/afsocket/afsocket-source.c
+++ b/modules/afsocket/afsocket-source.c
@@ -26,6 +26,7 @@
 #include "fdhelpers.h"
 #include "gsocket.h"
 #include "stats/stats-registry.h"
+#include "stats/stats-cluster-single.h"
 #include "mainloop.h"
 #include "poll-fd-events.h"
 #include "timeutils/misc.h"
@@ -53,6 +54,33 @@ typedef struct _AFSocketSourceConnection
 } AFSocketSourceConnection;
 
 static void afsocket_sd_close_connection(AFSocketSourceDriver *self, AFSocketSourceConnection *sc);
+
+static void
+_connections_count_set(AFSocketSourceDriver *self, gint value)
+{
+  self->num_connections = value;
+  stats_counter_set(self->num_connections_stats, value);
+}
+
+static gint
+_connections_count_get(AFSocketSourceDriver *self)
+{
+  return self->num_connections;
+}
+
+static void
+_connections_count_inc(AFSocketSourceDriver *self)
+{
+  self->num_connections++;
+  stats_counter_inc(self->num_connections_stats);
+}
+
+static void
+_connections_count_dec(AFSocketSourceDriver *self)
+{
+  self->num_connections--;
+  stats_counter_dec(self->num_connections_stats);
+}
 
 static gchar *
 _format_sc_name(AFSocketSourceConnection *self, gint format_type)
@@ -397,7 +425,7 @@ afsocket_sd_process_connection(AFSocketSourceDriver *self, GSockAddr *client_add
 
 #endif
 
-  if (self->num_connections >= self->max_connections)
+  if (_connections_count_get(self) >= self->max_connections)
     {
       msg_error("Number of allowed concurrent connections reached, rejecting connection",
                 evt_tag_str("client", g_sockaddr_format(client_addr, buf, sizeof(buf), GSA_FULL)),
@@ -416,7 +444,7 @@ afsocket_sd_process_connection(AFSocketSourceDriver *self, GSockAddr *client_add
       if (log_pipe_init(&conn->super))
         {
           afsocket_sd_add_connection(self, conn);
-          self->num_connections++;
+          _connections_count_inc(self);
           log_pipe_append(&conn->super, &self->super.super.super);
         }
       else
@@ -506,7 +534,7 @@ afsocket_sd_close_connection(AFSocketSourceDriver *self, AFSocketSourceConnectio
   log_pipe_deinit(&sc->super);
   self->connections = g_list_remove(self->connections, sc);
   afsocket_sd_kill_connection(sc);
-  self->num_connections--;
+  _connections_count_dec(self);
 }
 
 static void
@@ -560,10 +588,12 @@ _dynamic_window_realloc(AFSocketSourceDriver *self)
 static void
 _dynamic_window_set_balanced_window(AFSocketSourceDriver *self)
 {
-  if (self->num_connections <= 0)
+  gint number_of_connections= _connections_count_get(self);
+
+  if (number_of_connections <= 0)
     return;
 
-  gsize new_balanced_win = self->dynamic_window_pool->pool_size / self->num_connections;
+  gsize new_balanced_win = self->dynamic_window_pool->pool_size / number_of_connections;
   if (new_balanced_win == 0)
     {
       msg_info("Cannot allocate more dynamic window for new clients. From now, only static window is allocated."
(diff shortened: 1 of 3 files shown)


modules/afsocket/afsocket-source.c
```diff
@@ -397,12 +425,18 @@ afsocket_sd_process_connection(AFSocketSourceDriver *self, GSockAddr *client_add
...
 
-  if (self->num_connections >= self->max_connections)
+  if (_connections_count_get(self) >= self->max_connections)
     {
+      gchar *source_name = cfg_tree_get_rule_name(&self->super.super.super.cfg->tree, ENC_SOURCE, self->super.super.super.expr_node);
+
       msg_error("Number of allowed concurrent connections reached, rejecting connection",
                 evt_tag_str("client", g_sockaddr_format(client_addr, buf, sizeof(buf), GSA_FULL)),
                 evt_tag_str("local", g_sockaddr_format(local_addr, buf2, sizeof(buf2), GSA_FULL)),
+                evt_tag_str("source name", source_name),
```
bazsi: please don't use spaces here. that confuses our log format. the proper solution is to use an underscore here.

modules/afsocket/afsocket-source.c
```diff
@@ -54,6 +55,33 @@ typedef struct _AFSocketSourceConnection
...
+_connections_count_set(AFSocketSourceDriver *self, gint value)
+{
+  self->num_connections = value;
+  stats_counter_set(self->num_connections_stats, value);
+}
+
+static gint
+_connections_count_get(AFSocketSourceDriver *self)
+{
+  return self->num_connections;
```
bazsi: I think we could easily use stats_counter_get() here and eliminate the member in self. Or am I missing something?

szemere: I discussed it with multiple team members IRL:
- use two variables
- or use a stats variable for multiple purpose

More people preferred the later one. I am okey with both solution.

kira-syslogng:
Build FAILURE

modules/afsocket/afsocket-source.c
bazsi: and if we did, we could just eliminate these functions as all of them would only touch a single stats counter, that would probably be just fine at the call sites.

maybe I would extract the function (and only that) that decides if we can accept a new connection.

modules/afsocket/afsocket-source.c
szemere: Fixed, thank you!

kira-syslogng:
Build FAILURE

modules/afsocket/afsocket-source.c
```diff
@@ -399,10 +399,16 @@ afsocket_sd_process_connection(AFSocketSourceDriver *self, GSockAddr *client_add
 
   if (self->num_connections >= self->max_connections)
     {
+      gchar *source_name = cfg_tree_get_rule_name(&self->super.super.super.cfg->tree, ENC_SOURCE, self->super.super.super.expr_node);
+
       msg_error("Number of allowed concurrent connections reached, rejecting connection",
                 evt_tag_str("client", g_sockaddr_format(client_addr, buf, sizeof(buf), GSA_FULL)),
                 evt_tag_str("local", g_sockaddr_format(local_addr, buf2, sizeof(buf2), GSA_FULL)),
+                evt_tag_str("source_name", source_name),
```
Kokan: The location tag should define exactly which LogPipe it deals with, sure sometimes a calculation due to buffers needed (that should be improved...). Still the source name could be not better when it is `anon-`, I would suggest to print the stats name here, as that's the data you want to cross-check anyway.

szemere: Thank you for pointing this out. I discussed the topic with the support team. They prefer keeping the current version. Together with the configuration element location it seems good enough for the current purpose.

kira-syslogng:
Build SUCCESS

modules/afsocket/afsocket-source.c
MrAnno: I wouldn't build any logic in code on the exact values of stats counters.
The lifecycle/lifetime of stats clusters are different than drivers', the registration/constructor of these counters are also complex.

In this special case, we could guarantee a valid value, because the counter is set explicitly after a reload (depending on the `keep-alive()` option (`afsocket_sd_restore_kept_alive_connections`)), so this could work, but in general, I think it's not a good idea to make decisions in code based on the exact values of stats counters.

lbudai: I think if we have to improve the diagnostic capabilities of syslog-ng then we should have to provide a generic way for getting internal values without the aforementioned lifecycle problem, with a better solution than this.

I know that this work is a support related work,  so won't block the PR.

MrAnno:
@kira-syslogng test this please test branch=connection_number_stats_item;

MrAnno:
@kira-syslogng do perftest

modules/afsocket/afsocket-source.c
```diff
@@ -397,12 +425,19 @@ afsocket_sd_process_connection(AFSocketSourceDriver *self, GSockAddr *client_add
 
 #endif
 
-  if (self->num_connections >= self->max_connections)
+  if (_connections_count_get(self) >= self->max_connections)
     {
+      gchar *source_name = cfg_tree_get_rule_name(&self->super.super.super.cfg->tree, ENC_SOURCE,
```
MrAnno: Alternatively, you can use `self->super.super.group`, which is exactly the same, but you don't have to bother with the deallocation, etc.:

```
gchar *group_name = self->super.super.group;
evt_tag_str("group_name", group_name),
```

modules/afsocket/afsocket-source.c
```diff
@@ -986,7 +1048,11 @@ afsocket_sd_init_method(LogPipe *s)
         }
     }
 
-  return afsocket_sd_restore_kept_alive_connections(self) && afsocket_sd_open_listener(self);
+  gboolean init_result = afsocket_sd_restore_kept_alive_connections(self) && afsocket_sd_open_listener(self);
+  if (init_result)
+    _afsocket_sd_register_connection_counter(self);
```
MrAnno: Why do you register the counter at the end of the init method?

`afsocket_sd_restore_kept_alive_connections()` calls `_connections_count_set`, which requires a valid/registered `num_connections_stats` counter item.

If I'm not mistaken, the counter has to be registered as soon as possible to avoid such issues.

kira-syslogng:
Build SUCCESS

kira-syslogng:

modules/afsocket/afsocket-source.c
```diff
@@ -758,13 +795,13 @@ afsocket_sd_restore_kept_alive_connections(AFSocketSourceDriver *self)
       GList *p = NULL;
       self->connections = cfg_persist_config_fetch(cfg, afsocket_sd_format_connections_name(self));
 
-      self->num_connections = 0;
+      _connections_count_set(self, 0);
```
MrAnno: When `keep-alive()` is disabled, the counter does not work properly.
This is because the lifetime of the stats counter is different than the drivers'.

Reproduction: reload syslog-ng when it has active connections with `keep-alive(no)`. The counter should reset to 0 in this case.  An explicit `_connections_count_set(self, 0);` call is needed during the initialization of the driver.

kira-syslogng:
Build SUCCESS
 
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3194, opened 2020-03-24
scl: add sumologic

### LINKED ISSUE #1077: Create SCL for sumologic
Similarly to the Loggly and Logmatic, we could create an SCL for sumologic as well:

https://help.sumologic.com/Beta/Beta_-_Sources/Beta_-_Cloud_Syslog_Source#Configuring_a_Cloud_Syslog_Source_for_Syslog-NG

(on an unrelated note, we should update the scls to use the network driver instead of the outdated tcp driver)

vatsal-encs:
fixes #1077

### DIFF
--- a/scl/sumologic/sumologic.conf
+++ b/scl/sumologic/sumologic.conf
@@ -1,0 +1,86 @@
+#############################################################################
+# Copyright (c) 2020 Balabit
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+# EXAMPLES:
+#
+# With TLS encryption (make sure you trust the sumologic CA cert by putting it
+# to /etc/ssl, or create a separate CA directory):
+# For endpoint see - https://help.sumologic.com/APIs/General-API-Information/Sumo-Logic-Endpoints-and-Firewall-Security
+# 
+# Send data using syslog:
+#    log {
+#       destination {
+#                sumologic-syslog(token("USER-TOKEN-AS-PROVIDED-BY-sumologic") deployment("ENDPOINT")
+#                       tls(peer-verify(required-trusted) ca-dir('/etc/syslog-ng/ca.d'))
+#                );
+#       };
+#    };
+#
+#
+# Send JSON data:
+#    log {
+#        source { system(); };
+#        destination {
+#            sumologic-syslog(token("USER-TOKEN-AS-PROVIDED-BY-sumologic") deployment("ENDPOINT")
+#                   tls(peer-verify(required-trusted) ca-dir('/etc/syslog-ng/ca.d')) template("$(format-json --scope all-nv-pairs)"));
+#            );
+#        };
+#    };
+#
+# Send data using http:
+#    log {
+#        destination {
+#            sumologic-http(collector("UNIQUE-HTTP-COLLECTOR-CODE-AS-PROVIDED-BY-sumologic") deployment("ENDPOINT")
+#                   tls(peer-verify(yes) ca-dir('/etc/syslog-ng/ca.d')));
+#            );
+#        };
+#    };
+#
+
+block destination sumologic-syslog(token()
+                         tag("tag")
+                         deployment()
+                         port(6514)
+                         tls()
+                         template("$MSG") ...) {
+						 
+@requires json-plugin "The json-plugin is required for sumologic-syslog()"
+
+  network("syslog.collection.`deployment`.sumologic.com" port(`port`) transport(tls) tls(`tls`)
+          template("<${PRI}>1 ${ISODATE} ${HOST} ${PROGRAM:--} ${PID:--} ${MSGID:--} [`token` tag=\"`tag`\"] `template`\n")
+          template_escape(no)
+          `__VARARGS__`
+  );
+};
+
+block destination sumologic-http(
+                         tag("tag")
+                         deployment()
+                         collector()
+                         tls() ...) {
+						 
+@requires http "The http() driver is required for sumologic-http(), please install syslog-ng-mod-http (Debian) or syslog-ng-http (RHEL) package"
+
+  http(url("https://collectors.`deployment`.sumologic.com/receiver/v1/http/`collector`") tls(`tls`) method("POST")
+          `__VARARGS__`
+  );
+};
--- a/news/feature-3194.md
+++ b/news/feature-3194.md
@@ -1,0 +1,5 @@
+scl: add sumologic destinations: sumologic-syslog() and sumologic-http()
+
+See 
+[sumologic:syslog source](https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/Cloud-Syslog-Source#Configuring_a_Cloud_Syslog_Source_for_Syslog-NG)
+[sumologic:http source](https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/HTTP-Source)
--- a/scl/CMakeLists.txt
+++ b/scl/CMakeLists.txt
@@ -27,6 +27,7 @@ set(SCL_DIRS
     snmptrap
     solaris
     sudo
+    sumologic
     syslogconf
     system
     telegram
--- a/scl/Makefile.am
+++ b/scl/Makefile.am
@@ -27,6 +27,7 @@ SCL_SUBDIRS	= \
 	snmptrap	\
 	solaris	\
 	sudo		\
+	sumologic	\
 	syslogconf	\
 	system		\
 	websense	\


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

scl/sumologic/sumologic.conf
```diff
@@ -0,0 +1,79 @@
...
+@requires json-plugin
+
+block destination sumologic(token(TOKEN)
+                         tag("tag")
+                         host('syslog.collection.ca.sumologic.com')
+                         port(6514)
+                         transport(tls)
+                         template("$MSG") ...) {
+  network("`host`" port(`port`) transport(`transport`)
+          template("<${PRI}>1 ${ISODATE} ${HOST} ${PROGRAM} ${PID} ${MSGID} [`token`@41123 tag=\"`tag`\"] `template`\n")
```
bazsi: you might want to handle the case where PROGRAM/PID/MSGID is unset, you can use the default substitution feature of the template language, e.g. ${MSGID:--} would use a single dash as value when MSGID is not set.

bazsi:
ok to test

bazsi:
apart from the single comment above, this looks good to me.

kira-syslogng:
Build SUCCESS

scl/sumologic/sumologic.conf
furiel: According to this:
https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/Cloud-Syslog-Source
sumologic expects syslog protocol. Shouldn't we use `syslog` driver instead of `network` driver? Can sumologic accept logs without syslog frame?

I am a little afraid of the unicode messages too. What happens if you pass a message containing http://www.fileformat.info/info/unicode/char/010d/index.htm?
What I would expect the stream is split in the middle of the unicode character, (`0x0d`), the first part is processed by sumologic. Then sumologic would try to parse the rest of the message as if it were a new message, but that does not conform to the syslog protocol, and terminates the connection?

scl/sumologic/sumologic.conf
```diff
@@ -0,0 +1,79 @@
...
+#            sumologic(token("USER-TOKEN-AS-PROVIDED-BY-sumologic")
+#                   tag(apache)
+#                   template("$(format-json .apache.* timestamp=${ISODATE})"));
+#        };
+#    }
+#
+
+@requires json-plugin
+
+block destination sumologic(token(TOKEN)
```
alltilla: If `token` is a mandatory option, you can define it like this: `token()`, so `syslog-ng` forces the user to set it, when using the `sumologic()` destination.

scl/sumologic/sumologic.conf
```diff
@@ -0,0 +1,79 @@
...
+                         tag("tag")
+                         host('syslog.collection.ca.sumologic.com')
```
alltilla: If I understand right only the `ca` part is the variable here. If that's right, it might be more convenient to define a `deployment()` option, and assemble the full `host` like:
```
"syslog.collection.`deployment`.sumologic.com"
```

scl/sumologic/sumologic.conf
alltilla: Isn't `@41123` part of some specific token?
The [website](https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/Cloud-Syslog-Source#send-data-to%C2%A0cloud-syslog-source-with-syslog-ng) says:
> Replace the sample token, E5kTyaEcth45/DU81M236oU4vM8j1ZaqTpWgjXB6lod7cFTeq09zzMn5ErmM0O/3@41123,
with your token.

alltilla:
Thanks for opening this PR so quick :) Well done!

czanik:
Also note, that you can push logs to SumoLogic through an HTTP API: https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/HTTP-Source/Upload-Data-to-an-HTTP-Source
This might be an easier solution for those behind firewalls.

scl/sumologic/sumologic.conf
furiel: We discussed this with the team, and the reason that we use `network` for loggly, logmatic, and may be better to use here too than `syslog` is that it more difficult to add structured data to a syslog destination, than the current proposal. I am marking my comment as resolved.

furiel:
+1 for @czanik. I would also prefer http destination, if that can work. According to the linked documentation, the http collector supports batching. With the http workers+batching+load balancing, it might be more scalable than the syslog collector.

Supporting both might be an option too: `sumologic-syslog`, `sumologic-http`.

scl/sumologic/sumologic.conf
vatsal-encs: okay

scl/sumologic/sumologic.conf
vatsal-encs: Done

scl/sumologic/sumologic.conf
vatsal-encs: `The number 41123 in the token is the Sumo Private Enterprise Number (PEN). `

I think this key is fixed. So, maybe we can hard code it in the config.

vatsal-encs:
> +1 for @czanik. I would also prefer http destination, if that can work. According to the linked documentation, the http collector supports batching. With the http workers+batching+load balancing, it might be more scalable than the syslog collector.
> 
> Supporting both might be an option too: `sumologic-syslog`, `sumologic-http`.

@furiel : That is a very good suggestion. I have added both the options now (`sumologic-syslog`, `sumologic-http`)

scl/sumologic/sumologic.conf
vatsal-encs: @bazsi : okay :+1:

kira-syslogng:
Build SUCCESS

scl/sumologic/sumologic.conf
gaborznagy: Hi @vatsal-encs ,
we shouldn't hard-code the PEN into the template as that is in the generated token as well.

This results into a duplicated PEn number, see my syslog-ng debug output:
`[2020-03-27T16:35:23.431747] Outgoing message; message='<13>1 2020-03-27T16:35:23+01:00 localhost da_logger 22918 - [token...zf/58Y@41123@41123 tag="tag"] bsd-source sent something\x0a'
`

gaborznagy:
documentation note for sumologic-http configuration:

`collector()`: use the mentioned `UniqueHTTPCollectorCode` in sumologic's user guide:
> [UniqueHTTPCollectorCode] is the string that follows the last forward slash (/) in the upload URL for the HTTP source.  
https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/HTTP-Source/Upload-Data-to-an-HTTP-Source

scl/sumologic/sumologic.conf
vatsal-encs: Hi @gaborznagy,

Okay, I will remove this hardcoded PEN.

vatsal-encs:
> documentation note for sumologic-http configuration:
> 
> `collector()`: use the mentioned `UniqueHTTPCollectorCode` in sumologic's user guide:
> 
> > [UniqueHTTPCollectorCode] is the string that follows the last forward slash (/) in the upload URL for the HTTP source.
> > https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/HTTP-Source/Upload-Data-to-an-HTTP-Source

@gaborznagy : Yes, I have used `collector()` for `UniqueHTTPCollectorCode` . Let me know if you want to have different name apart from this one.

gaborznagy:
Thanks, the PEN number fixed, I've tested it!

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

scl/sumologic/sumologic.conf
```diff
@@ -0,0 +1,82 @@
...
+#    log {
+#        destination {
+#            sumologic-http(collector("UNIQUE-HTTP-COLLECTOR-CODE-AS-PROVIDED-BY-sumologic") deployment("ENDPOINT")
+#                   tls(peer-verify(yes) ca-dir('/etc/syslog-ng/ca.d')));
+#            );
+#        };
+#    };
+#
+
+@requires json-plugin
```
Kokan: This is fine, but recently there has been a huge improvement on the `@requires` usage in this PR https://github.com/syslog-ng/syslog-ng/pull/3015.

The current version, if you do not have the `json-plugin` it is going to skip both `sumologic-syslog` and `sumologic-http` despite the later could work without json just fine (as far as I can see the http version does not require json plugin).

While if you move the `@requires json-plugin` inside the `block` and add an extra error message after it, it is going to provide nicer error message (see the linekd PR) and also only exclude that `block` and not the rest of the file.

My suggestion is somthing like this:
```
block destination sumologic-syslog(... )
{
@requires json-plugn "The json-plugin is required for sumologic-syslog(), ..."
...
}
```

scl/sumologic/sumologic.conf
```diff
@@ -0,0 +1,82 @@
...
+          `__VARARGS__`
+  );
+};
+
+block destination sumologic-http(
+                         tag("tag")
+                         deployment()
+                         collector()
+                         tls() ...) {
+  http(url("https://collectors.`deployment`.sumologic.com/receiver/v1/http/`collector`") tls(`tls`) method("POST")
```
Kokan: The http plugin was not listed as dependency, but it is kinda. Please add a require for that and keep in mind previous comment about how to use `@requires`.

kira-syslogng:
Build SUCCESS

scl/sumologic/sumologic.conf
vatsal-encs: @Kokan : okay, thanks for your review.

scl/sumologic/sumologic.conf
vatsal-encs: @Kokan : ok

kira-syslogng:
Build SUCCESS

Kokan:
Note: the travis actually already finished, and it is green (checked the commit hash) but somehow it stuck in this state, but I am merging anyway.

PR #3195, opened 2020-03-25
set-tag: add test

furiel:
No changenote necessary.

`condition='match("MATCHSTRING" value("MSG"))'` is a little awkward (the value is a verbatim string).

This is the first time a driver needs positional parameter inside an option block. (If there were no `"MATCHSTRING"`, light could handle it with a dictionary though.)

 I was wondering if I could use a tuple format + small modification in the rendering logic, like
`condition = {"match" : ("MATCHSTRING", {"value" : "MSG"}}`, but tuple is already reserved for different logic for snmptrapd.

Unfortunately I do not know any simple way to fix this in the current architecture.

Edit: introduced a filter object that can be passed to condition

### DIFF
--- a/tests/python_functional/src/syslog_ng_config/renderer.py
+++ b/tests/python_functional/src/syslog_ng_config/renderer.py
@@ -20,6 +20,7 @@
 # COPYING for details.
 #
 #############################################################################
+from src.syslog_ng_config.statements.filters.filter import Filter
 
 
 def render_version(version):
@@ -73,6 +74,10 @@ def render_name_value(name, value):
     return "        {}({})\n".format(name, value)
 
 
+def render_driver(name, driver):
+    return "        {}({})\n".format(name, render_statement(driver))
+
+
 def render_driver_options(driver_options):
     config_snippet = ""
 
@@ -81,6 +86,8 @@ def render_driver_options(driver_options):
             config_snippet += render_options(option_name, option_value)
         elif (isinstance(option_value, tuple) or isinstance(option_value, list)):
             config_snippet += render_list(option_name, option_value)
+        elif isinstance(option_value, Filter):
+            config_snippet += render_driver(option_name, option_value)
         else:
             config_snippet += render_name_value(option_name, option_value)
 
@@ -89,8 +96,11 @@ def render_driver_options(driver_options):
 
 def render_statement(statement):
     config_snippet = ""
+    config_snippet += "    {} (\n".format(statement.driver_name)
     config_snippet += render_positional_options(statement.positional_parameters)
     config_snippet += render_driver_options(statement.options)
+    config_snippet += "    )"
+
     return config_snippet
 
 
@@ -105,12 +115,10 @@ def render_statement_groups(statement_groups):
 
         for statement in statement_group:
             # driver header
-            config_snippet += "    {} (\n".format(statement.driver_name)
-
             config_snippet += render_statement(statement)
 
             # driver footer
-            config_snippet += "    );\n"
+            config_snippet += ";\n"
 
             # statement footer
         config_snippet += "};\n"
--- a/tests/python_functional/src/syslog_ng_config/renderer.py
+++ b/tests/python_functional/src/syslog_ng_config/renderer.py
@@ -22,6 +22,96 @@
 #############################################################################
 
 
+def render_version(version):
+    return "@version: {}\n".format(version)
+
+
+def render_includes(includes):
+    include_lines = ['@include "{}"'.format(include) for include in includes]
+    return "\n".join(include_lines)
+
+
+def render_global_options(global_options):
+    globals_options_header = "options {\n"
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

furiel:
Thanks for the ideas. I agree, let's not introduce a bad example. I try to find out something. Adding wip.

furiel:
Pushed my new attempt. Now the test looks like:
```
match = Match(config.stringify("MATCHSTRING"), value=config.stringify("MSG"))
config.create_set_tag("SHOULDMATCH", condition=match)
```

kira-syslogng:
Build SUCCESS

furiel:
(whops, wrong buttons: accidentally deleted the branch. restored now)

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/parsers/set-tag/test_set_tag.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
```
mitzkia, line 1: I would put this testcase under: `functional_tests/rewrite/` dir

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
```diff
@@ -81,8 +82,11 @@ def create_file_source(self, **options):
     def create_example_msg_generator_source(self, **options):
         return ExampleMsgGeneratorSource(**options)
 
+    def create_set_tag(self, tag, **options):
```
mitzkia, line 85: I would rewrite this function to: `create_rewrite_set_tag()`

tests/python_functional/src/syslog_ng_config/renderer.py
```diff
@@ -20,6 +20,129 @@
...
+def render_list(name, options):
+    config_snippet = ""
+
+    for element in options:
+        config_snippet += "        {}({})\n".format(name, element)
+
+    return config_snippet
+
+
+def render_simple(name, value):
```
mitzkia, line 73: If possible, could we change the function name? Maybe some config related name would be better.

tests/python_functional/src/syslog_ng_config/renderer.py
```diff
@@ -20,6 +20,129 @@
...
+
+
+def render_positional_options(positional_parameters):
+    config_snippet = ""
+    for parameter in positional_parameters:
+        config_snippet += "        {}\n".format(str(parameter))
+    return config_snippet
+
+
+def render_dict(name, options):
```
mitzkia, line 54: If possible, could we change the function name? Maybe some config related name would be better.

tests/python_functional/src/syslog_ng_config/renderer.py
```diff
@@ -20,6 +20,129 @@
...
+
+    config_snippet = ""
+    config_snippet += "        {}(\n".format(name)
+    config_snippet += render_driver_options(options)
+    config_snippet += "        )\n"
+
+    return config_snippet
+
+
+def render_list(name, options):
```
mitzkia, line 64: If possible, could we change the function name? Maybe some config related name would be better.

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
furiel: Fixed.

tests/python_functional/src/syslog_ng_config/renderer.py
furiel: Fixed.

tests/python_functional/functional_tests/parsers/set-tag/test_set_tag.py
furiel: Fixed.

tests/python_functional/src/syslog_ng_config/renderer.py
furiel: Fixed.

tests/python_functional/src/syslog_ng_config/renderer.py
furiel: We discussed this with Micek, and for now we leave the name as it is. In the long run, we can eliminate its need. FYI: @norberttak 

The `render_list` functionality was introduced for  `snmp`, and `snmp` is it's only user. In case of `snmp`, the usual python dictionary cannot work, because the same keys may appear in the driver, but a key can appear only once in a dictionary.

I am thinking of introducing a new data structure: let's call it `MultiDict`, that can store the same key multiple times. If it has the same `items()` interface used for rendering, and returns the multi-occurring keys too, the original render function can work without introducing the `render_list`. That would be a nice dependency reversal.

kira-syslogng:
Build SUCCESS

PR #3196, opened 2020-03-25
[WIP] example-msg-generator: crash during failed reload

furiel:
`example-msg-generator` starts it's timer during init. The timer triggers a `log_source_post`, which may push the message down the graph even before the entire `cfg_init()` finishes. This means the pushed message may arrive to uninitialized pipes. In this patch, the timer is started by apphook, instead of init.

I can steadily reproduce the crash with a failed reload:
```
@version: 3.26

filter f_my_info { level(info); }; # comment this line before reload

log {
  filter { filter(f_my_info); };

};
```

The problem found while testing: https://github.com/syslog-ng/syslog-ng/pull/3176

### DIFF
--- a/lib/apphook.c
+++ b/lib/apphook.c
@@ -94,6 +94,34 @@ register_application_hook(gint type, ApplicationHookFunc func, gpointer user_dat
     }
 }
 
+void
+unregister_application_hook(gint type, ApplicationHookFunc func, gpointer user_data)
+{
+  GList *l, *l_next;
+
+  msg_debug("Unregistering hook",
+            evt_tag_int("type", type),
+            evt_tag_ptr("func", func),
+            evt_tag_ptr("user_data", user_data));
+
+  for (l = application_hooks; l; l = l_next)
+    {
+      ApplicationHookEntry *e = l->data;
+
+      if (e->type == type && e->func == func && e->user_data == user_data)
+        {
+          l_next = l->next;
+          application_hooks = g_list_remove_link(application_hooks, l);
+          g_free(e);
+          g_list_free_1(l);
+        }
+      else
+        {
+          l_next = l->next;
+        }
+    }
+}
+
 static void
 _update_current_state(gint type)
 {
--- a/lib/apphook.h
+++ b/lib/apphook.h
@@ -63,6 +63,7 @@ gboolean app_is_starting_up(void);
 gboolean app_is_shutting_down(void);
 
 void register_application_hook(gint type, ApplicationHookFunc func, gpointer user_data);
+void unregister_application_hook(gint type, ApplicationHookFunc func, gpointer user_data);
 
 void app_thread_start(void);
 void app_thread_stop(void);
--- a/lib/eventlog/src/evtlog.h
+++ b/lib/eventlog/src/evtlog.h
@@ -141,6 +141,7 @@ void evt_rec_free(EVTREC *e);
  **/
 EVTTAG *evt_tag_str(const char *tag, const char *value);
 EVTTAG *evt_tag_int(const char *tag, int value);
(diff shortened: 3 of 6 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

furiel:

kira-syslogng:
Build SUCCESS

furiel:
@bazsi good point. I was thinking that msg-generator is threaded source based. But I see, it is a simple source in the main thread. I double check my understanding.

furiel:
@bazsi here is my revised understanding. I just added fprintf-s to the code

<details>

```
msg_generator_sd_init 0x66bec0
msg_source_init 0x66e810
start_timer_immediately
filter init 0x66bd00
[2020-03-26T13:15:19.097810] syslog-ng starting up; version='3.26.1.90.gcb23feb'
send_message

[here I reloaded]

msg_generator_sd_deinit 0x66bec0
msg_source_deinit 0x66e810
stop_timer
log_pipe_deinit 0x66e810
log_pipe_deinit 0x66bec0
log_pipe_deinit 0x63efd0
log_pipe_deinit 0x66d8d0
log_pipe_deinit 0x66da10
log_pipe_deinit 0x66bd00
msg_generator_sd_init 0x678d20
msg_source_init 0x66e810
start_timer_immediately
filter init 0x684390
[2020-03-26T13:15:30.110455] Referenced filter rule not found in filter() expression; rule='f_my_info'
[2020-03-26T13:15:30.110480] Error initializing message pipeline; plugin_name='not a plugin', location='../etc/msg-generator.conf:8:11'
[2020-03-26T13:15:30.110494] Error initializing new configuration, reverting to old config;
[2020-03-26T13:15:30.110511] Configuration reload finished;
send_message
**
ERROR:/home/furiel/workspace/failed-reload/syslog-ng/lib/logpipe.h:344:log_pipe_queue: assertion failed: ((s->flags & PIF_INITIALIZED) != 0)
```

</details>

What I can make out of the printout is that:

After reload, every pipes are deinitialized az expected. The new pipes are started initializing. First the example msg generator initializes successfully. Then the filter fails to initialize do to config error.

What would you expect is that at that point the half initialized config deinitializes, and the original config takes place.

The problem is the deinitialization is not immediate, but scheduled as an ivykis task (revert_config). On the other hand, we have an active ivykis timer too, because example-msg-generator init has passed, and the deinit was not yet called.

Now the next ivykis cycle, ivykis needs to pick the timer or the task. But ivykis runs the timers before the tasks, so the timer is called earlier. The timer executes a post, the message arrives to the not initialized filter, and then crash.

Now I believe this is what @0140454 tried to explain in https://github.com/syslog-ng/syslog-ng/pull/3176#issuecomment-601194494, I was a little slow to understand. 

In summary, I think this change is still necessary. It would be nice to have a general solution though. Every driver that starts an immediate timer during init is susceptible to this bug. 

I am yet to completely understand #3176, but I think the intention is to have a more general solution: that change ensures the `revert_config` task is scheduled in front of any other tasks. But unfortunately, it does not work on timers.

bazsi:
Hmmm what if we schedule the revert task as a timer, this way any timer
scheduled would be added _after_ that item.

This is still relying on implementation details of ivykis but at least it's
at a single location. Iirc there was a reason we scheduled revert as an
async task instead of doing it immediately, but without going back to the
code I cant remember the reason.

MrAnno:
I think this can be closed now that we have a universal fix in #3176.
We don't rely on ivykis' internal task ordering with the mentioned fix.

furiel:
You are right. Closing in favor of #3176

PR #3201, opened 2020-03-27
func-test: fix crash when `dbd_dir` env variable is not set

alltilla:
python crashes at `join(paths)` in line 77, if a list element is`None` and not a `str`.


kira-syslogng:
Build SUCCESS

PR #3220, opened 2020-04-03
configure/cmake: support python configuration with "python" prefix

alltilla:
A little background of my motivation:

I am working on introducing GitHub Actions to the syslog-ng repo, where we have an option to run jobs with multiple variable options. These options get added to the jobs name during run like:
> job (foo, bar)

I would like to use two options:
```
python: python2/python3
build-tool: autotools/cmake
```

I am using the `python` option in the configure step, where if we do not allow `--with-python=python2` just `--with-python=2`, I have to change the option to:
```
python: 2/3
```
, which would name the job like 
> job (2, autotools)

, which is not so pretty as 
>job (python2, autotools)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,6 +225,8 @@ set(SYSLOG_NG_HAVE_INOTIFY "${Inotify_FOUND}")
 
 set (PYTHON_VERSION "AUTO" CACHE STRING "Version of the installed development library" )
 
+string(REGEX REPLACE "^(python|PYTHON)" "" PYTHON_VERSION ${PYTHON_VERSION})
+
 if ("${PYTHON_VERSION}" STREQUAL "AUTO")
   find_package(PythonInterp)
   find_package(PythonLibs)
--- a/configure.ac
+++ b/configure.ac
@@ -410,6 +410,9 @@ LT_INIT([dlopen disable-static])
 if test "x$with_python" = "xauto"; then
   AM_PATH_PYTHON(,,[:])
 else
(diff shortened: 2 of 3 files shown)


kira-syslogng:
Build SUCCESS

lbudai:
could you provide more details (I mean code) on where you are using these options?

alltilla:
I am not on the computer, so I can only give these links, but I think they show my intention.

Example CI run: https://github.com/alltilla/syslog-ng/actions/runs/69784772

It's workflow file: https://github.com/alltilla/syslog-ng/actions/runs/69784772/workflow

furiel:
Are there any alternatives? Is it possible to `cut` the variable in the github actions yaml file?

alltilla:
@furiel: Sure it is possible, but I don't see the drawback of letting everyone use the "python" prefix.

furiel:
We are talking about personal preferences. Not necessarily a must from my side, but here are my reasons.

For me it would be a little confusing. The `with-python` processing is already very complex. It is less about choosing between python2 or python3, but finding an appropriate pkg config file on the system. Whatever is specified with the `with-python`, it is explicitely mapped to a pc file.
For me this is a conceptual change.

I am not confortable with the trigger either. We are trying to achieve a cosmetic change in a tool that uses our build system by reshaping our build system, and not the ci tool. This is wrong direction of dependency.

alltilla:
@furiel 
Sure, I see your point. It's okay, I will drop the PR. We will know the meaning behind `job (2, autotools)`.

PR #3224, opened 2020-04-03
Stats external counter

lbudai:
external counter: when someone wants to control the ownership of the lifecycle/permission of the embedded counter, it could be a useful extension to the existing stats-counter model.

The re-registration is little bit weird, but one might hold a `StatsCounterItem` for accessing the value of an external counter.

A question regarding to the live mask: when the `use_count` reaches 0 during an unregister/untrack, live mask remain unchanged - we have unit tests that are checking this behaviour.
Why do we need this?

@bazsi : could you check this?


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3225, opened 2020-04-05
Stats external counter

lbudai:
stats: add external stats counters

* Motivation:
There are situations when someone wants to expose internal variables through stats-api.
Without this changeset, we need two distinct variables: one is for the internal state, other is for the
stats registration (internal state cannot depends on the life cycle of the StatsCounterGroup).
In other words, this changeset makes it possible to convert any `atomi_gssize` into a (stats-)queryable counter.

* Proposed solution: External stats counters

When someone registers
 * a "normal" stats counter, the registry API provides a pointer to an instance of a StatsCounterItem.
StatsCounterItems are stored in a StatsCounterGroup (as an array, "indexed" by type), StatsCounterGroup
is part of StatsCluster. Life cycle is defined by the counter group.
StatsCounterItem has-an `atomic_gssize` variable. So the `stats_counter_dec/sub/add/...` API modifies this
`atomic_gssize` variable which is stored in an array (which is part of the StatsCounterGroup).

 * an "external" stats counter, the registry API takes a pointer to an `atomic_gssize` variable.
The provided StatsCounterItem will store this pointer AND mark the StatsCounterItem instance as read-only:
this means that the value, stored by the StatsCounterItem, can only be modified by the owner of the external
`atomic_gssize` variable.

* Implementation:
StatsCounterItem is a union: it stores an `atomic_gssize` or a pointer to an `atomic_gssize` value (so a direct
or an indirect value).

Currently, it is possible to register a StatsCounterItem and get back an external counter. The other direction
is not supported: when a counter is registered as an internal, it can't be re-registered as an external (assert).
Re-register an already registered external counter with a different `atomic_gssize` pointer is not supported (assert).

When the `use_count` reaches zero, the pointer is set to NULL and `live_mask` also updated.
This means that after all the references to an external StatsCounterItem removed (untrack/unregister calls), then it is
possibble to re-register a new StatsCounterItem (normal, or external).
As in case of the original/normal stats counter the live mask is not updated in this case (`use_count` reaches zero), trying to register an external variable ends in an assert.


kira-syslogng:
Build SUCCESS

lib/stats/stats-registry.c
```diff
@@ -140,8 +140,12 @@ _register_counter(gint stats_level, const StatsClusterKey *sc_key, gint type,
   sc = _grab_cluster(stats_level, sc_key, dynamic);
   if (sc)
     {
+      StatsCounterItem *ctr = stats_cluster_get_counter(sc, type);
```
furiel, line 143: These 4 lines confuse me. What is the goal here? To support registering already registered counters? It would be cleaner if the stats counters need to be unregistered first.

`stats_cluster_get_counter` can only return tracked counters. So the `stats_cluster_track_counter` is not needed if `stats_cluster_get_counter` returns with a counter. Not necessarily wrong (due that bit-or is idempotent), but confusing. Maybe it is called at a wrong place?

lib/stats/stats-registry.c
```diff
@@ -173,6 +214,13 @@ stats_register_counter(gint stats_level, const StatsClusterKey *sc_key, gint typ
   return _register_counter(stats_level, sc_key, type, FALSE, counter);
 }
 
+StatsCluster *
+stats_register_external_counter(gint stats_level, const StatsClusterKey *sc_key, gint type,
```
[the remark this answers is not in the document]
lbudai: I don't want to hide this... and I can imagine that someone wants to store the `StatsCounterItem` itself (for passing it to other layers as a read-only 'view').

On the other hand I can imagine that someone would require a more sophisticated API - I'll address it in another PR.

lib/stats/stats-registry.c
lbudai: `track_counter` is different: it also increases the `use_count`.

"Currently, it is possible to register a `StatsCounterItem` and get back an external counter.":
These 4 lines implements this.

lib/stats/stats-counter.h
```diff
@@ -29,45 +29,56 @@
...
-  if (counter)
+  if (counter && !stats_counter_read_only(counter))
     atomic_gssize_add(&counter->value, add);
 }
 
 static inline void
 stats_counter_sub(StatsCounterItem *counter, gssize sub)
 {
-  if (counter)
+  if (counter && !stats_counter_read_only(counter))
```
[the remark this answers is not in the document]
lbudai: This is not as simple: counters can be accessed via the stats-query tool - currently only reset is available, but it is easy to extend and I think if a setter operation is accessible via an API then from that point IMHO we cannot say that using another modifier is a programming error.

kira-syslogng:
Build SUCCESS

lib/stats/stats-registry.c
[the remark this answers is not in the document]
furiel: Ok, fine for me. A small request: if a caller is not interested in the stats `StatsCounterItem`, it would be nice if they could pass `NULL` to the output parameter.. Thus making the output optional.

lib/stats/stats-registry.c
furiel: I understand now, thanks.

lib/stats/stats-registry.c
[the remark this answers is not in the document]
lbudai: Based on your request I _removed_ the output parameter.
Now you want me to add it again and enable users to set it to NULL? (I won't do that :) )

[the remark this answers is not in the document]
furiel: Oh right, I reviewed an older version. It looks neat.

lib/stats/stats-counter.h
```diff
@@ -29,45 +29,56 @@
...
+stats_counter_read_only(StatsCounterItem *counter)
+{
+  return counter->external;
+}
+
 static inline void
 stats_counter_add(StatsCounterItem *counter, gssize add)
 {
-  if (counter)
+  if (counter && !stats_counter_read_only(counter))
```
bazsi, line 52: I would use an assert here and ensure in higher layers that we are not calling counter changing functions in case it is a read-only counter.

lbudai:
@bazsi, @furiel :
 * `reset_counters` moved to `stats-control`
 * assert when trying to modify an external counter with `stats_counter_xxx` (set/inc/dec/add/sub)
 * reimplement #3193 with external stats counters

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

modules/afsocket/afsocket-source.h
```diff
@@ -52,8 +53,7 @@ struct _AFSocketSourceDriver
   LogProtoServerFactory *proto_factory;
   GSockAddr *bind_addr;
   gint max_connections;
-  gint num_connections;
-  StatsCounterItem *num_connections_stats;
+  atomic_gssize num_connections;
```
furiel, line 56: Being an external counter, this does not need to be atomic. It can be simply gssize, and increased with `++`, and decreased with `--`, as in the original code.

lbudai: @furiel : It should be `atomic_gssize` - as the underlying api requires that.

I know that the underlying api is implemented by this PR, but I'd not change that: all the stats ctr items are atomics. There is no difference between regular and external counters from this point of view, and I'd stick to that.

PR #3229, opened 2020-04-08
Limit the size of internal()'s temporary queue

MrAnno:
This PR is a proposal that adds a limit to the `internal()` source's temporary queue.

The max capacity is hard-coded in the proposal, but a config option can be added easily, if needed.

Why do I want this?
syslog-ng currently uses message queues only on the destination side. The size and behavior of those queues can be configured with the `log-fifo-size()` destination, and the `flags(flow-control), log-iw-size()` source options. syslog-ng drops new messages when flow control is not configured and the destination queue exceeds the `log-fifo-size()` limit. When flow control is configured and syslog-ng reaches `log-iw-size()` number of messages kept in memory per source, it stops the given source.

In both cases, the memory footprint has an upper limit and it can be calculated based on the mentioned options.

The `internal()` source is different. When flow-control is not configured, it operates in a similar way as I described above. In case `flags(flow-control)` is set, or when file-based destinations are used (soft-flow-control) together with `internal()`, something else happens:

The internal source has an additional temporary source-side queue in addition to the destination queues. This is because internal() messages can be created anywhere, in any worker/dedicated threads, on any log path; and those threads and paths should never be blocked/stopped just because something is congested on the `internal()` log paths. The temporary queue is there to overcome this problem, and for other performance reasons.

Unfortunately, this temporary queue has no maximum capacity, so it can consume all available memory. During the investigation of another high-memory-usage issue, it was difficult to notice this side-issue, because we properly free the queue before shutting down syslog-ng (so technically, it's not a memory leak), and the only information about the current length of the queue was a statistic counter with the name "processed", which was a bit misleading for me:
```
global;internal_queue_length;;a;processed;482363
```
I've also renamed this counter to `global;internal_queue;;a;queued` and added `global;internal_queue;;a;dropped` in this MR.

My proposal is to have an upper limit on the internal source's temporary queue and drop messages when the capacity is reached.
My MR has an obvious effect: syslog-ng will lose internal messages if flow-control is used and both destination queues and the internal queue reach their capacity. If you don't want to block all logpaths because of `internal()`, and you also want a proper memory limit, I think you can't do better.

### DIFF
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -95,6 +95,7 @@ struct _AFInterSource
 {
   LogSource super;
   gint mark_freq;
+  const AFInterSourceOptions *options;
   struct iv_event post;
   struct iv_event schedule_wakeup;
   struct iv_timer mark_timer;
@@ -313,17 +314,20 @@ afinter_source_deinit(LogPipe *s)
 }
 
 static LogSource *
-afinter_source_new(AFInterSourceDriver *owner, LogSourceOptions *options)
+afinter_source_new(AFInterSourceDriver *owner, AFInterSourceOptions *options)
 {
   AFInterSource *self = g_new0(AFInterSource, 1);
 
   log_source_init_instance(&self->super, owner->super.super.super.cfg);
-  log_source_set_options(&self->super, options, owner->super.super.id, NULL, FALSE, FALSE,
+  log_source_set_options(&self->super, &options->super, owner->super.super.id, NULL, FALSE, FALSE,
                          owner->super.super.super.expr_node);
   afinter_source_init_watches(self);
   self->super.super.init = afinter_source_init;
   self->super.super.deinit = afinter_source_deinit;
   self->super.wakeup = afinter_source_wakeup;
+
+  self->options = options;
+
   return &self->super;
 }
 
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

lbudai:
Option is required.
What about backward compatbility?
I agree that queued is better, but what if someone already monitors this? Is it possible to deprecate a stats ctr? If you think it’s not, then an alias could do the job.
(I have another idea, maybe in the near future I’m reverting this and setup a completely different solution, but until that it can be merged.)

MrAnno:
> I think we can mark a stats_counter as obsolete with config_version and send a warning message to the user on startup.

Good idea.

Another option is to inform the user if they query a deprecated counter.
However, this would be weird, because the unfiltered `syslog-ng-ctl stats` output would contain all deprecation warnings.

I think the warning based on `config_version` with a simple alias to the old name (or a deprecation warning in the NEWS file/documentation) should be plenty enough. Then we can delete the counter in the next release.

I'll add the `log-fifo-size()` option too.

MrAnno:
@kira-syslogng do perftest

kira-syslogng:
Build SUCCESS

kira-syslogng:
Kira-performance-test: Build FAILURE

MrAnno:
I've added the deprecation warning, a counter alias (using the new external counter method :smile:), and the `log-fifo-size()` option.

MrAnno:
@kira-syslogng do perftest

kira-syslogng:
Build FAILURE

MrAnno:

kira-syslogng:
Kira-performance-test: Build FAILURE

kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng do perftest

kira-syslogng:
Kira-performance-test: Build FAILURE

MrAnno:
The performance test result is irrelevant, it fails on our master too.

kira-syslogng:
Build SUCCESS

PR #3230, opened 2020-04-09
eliminate infinite loop in case of a spurious file path

szemere:
Instead of a simple boolean result, file_opener_open_fd will return
that the error was temporary or permanent. (Currently spurious paths
is the only permanent error.)
If the error is permanent we can do nothing about it, so we skip the
logwriter_reopen method too, thus there will be no further attempt to
open the file, and the log message will be skipped by the file
destination.

WIP:
- [x] perftest
- [x] news file

### DIFF
--- a/modules/affile/affile-dest.c
+++ b/modules/affile/affile-dest.c
@@ -159,12 +159,17 @@ affile_dw_reopen(AFFileDestWriter *self)
       unlink(self->filename);
     }
 
-  if (file_opener_open_fd(self->owner->file_opener, self->filename, AFFILE_DIR_WRITE, &fd))
+  FileOpenerResult open_result = file_opener_open_fd(self->owner->file_opener, self->filename, AFFILE_DIR_WRITE, &fd);
+  if (open_result == FILE_OPENER_RESULT_SUCCESS)
     {
       LogTransport *transport = file_opener_construct_transport(self->owner->file_opener, fd);
 
       proto = file_opener_construct_dst_proto(self->owner->file_opener, transport,
                                               &self->owner->writer_options.proto_options.super);
+    }
+  else if (open_result == FILE_OPENER_RESULT_ERROR_PERMANENT)
+    {
+      return FALSE;
     }
   else
     {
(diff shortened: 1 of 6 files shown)


kira-syslogng:
Build SUCCESS

modules/affile/affile-dest.c
```diff
@@ -168,6 +169,9 @@ affile_dw_reopen(AFFileDestWriter *self)
     }
   else
     {
+      if (open_result == FILEOPENER_RESULT_ERROR_PERMANENT)
```
bazsi, line 172: this should be an "else if (open_result == FILEOPENER_RESULT_ERROR_PERMANENT) {} else {}" block, e.g.
the else and if should be combined, makes the flow more readable.

modules/affile/file-opener.h
```diff
@@ -47,6 +47,12 @@ typedef struct _FileOpenerOptions
   gint create_dirs;
 } FileOpenerOptions;
 
+typedef enum {
+  FILEOPENER_RESULT_SUCCESS,
```
bazsi, line 51: what do you think about an underscore between FILE and OPENER? those are separate terms, the filenames and other identifiers follow this convetion.

modules/affile/affile-dest.c
szemere: Thank You!

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng do perftest

kira-syslogng:

kira-syslogng:
Build SUCCESS

MrAnno:
I've used the following config to test the fix:

```
@version: 3.26

options {
  time-reopen(1);
};

log {
  source { example-msg-generator(num(1) template("../../path")); };
  destination { file("/tmp/$MSG.txt"); };
};
```

news/bugfix-3230.md
```diff
@@ -0,0 +1,8 @@
+`affile`: eliminate infinite loop in case of a spurious file path
+
+If the template evaluation of a log message will result to a spurious
+path in the file destination, syslog-ng refuses to create that file.
+However the problematic log message was left in the msg queue, so
+syslogng was trying to create that file again in time-reopen periods.
```
MrAnno, line 6: syslog-ng

kira-syslogng:
Build SUCCESS

PR #3231, opened 2020-04-09
WIP: refactor affile dest

szemere:
While investigating spurious path, I refactored affile_dd_queue and the related functions. Since they do not affect each other, I putted it in a different PR.

These are **not only** equivalent changes! (i.e: writer_hash life cycle) I tried to organize commits in a way, that they are easier to review one-by-one. Where I mention "extract", I only extracted a function, no other changes intended.

Question for the reviewers:
 IMHO we can eliminate "filename_is_a_template" check on the "fast path" with a minimal code duplication, by creating a different affile_dd_queue for the `single_writer` / `writer_hash` cases.

WIP:
- [ ] perftest
- [ ] news file
- [ ] open question regarding to affile_dd_queue

### DIFF
--- a/modules/affile/affile-dest.c
+++ b/modules/affile/affile-dest.c
@@ -668,60 +668,72 @@ affile_dd_open_writer(gpointer args[])
 }
 
 static AFFileDestWriter *
-_affile_dd_create_writer(AFFileDestDriver *self, LogMessage *msg)
+_affile_dd_create_single_writer(AFFileDestDriver *self)
 {
   AFFileDestWriter *next;
   gpointer args[2] = { self, NULL };
 
-  if (!self->filename_is_a_template)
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build FAILURE

szemere:
Update:
 Ironically. it turned out  those "remove unnecessary if statements" were not unnecessary after all.
 TL;DR I am closing this PR. There are good parts in it, but I think it's not worth to keep it in this way.

Explanation:
 Checking if single_writer / writer_hash exists is protected by locks. However at the beginning - when the writer do not exists - multiple threads will schedule the main loop call for creating it, until the first one will success and set the variable under lock.
 So those if statements in the open_writer function for checking "if writer already exists" were not a logical decision as I thought, but to eliminate the rest of the scheduled main loop calls.

PR #3234, opened 2020-04-14
New panos-parser() source as SCL

spricer:
scl: new panos-parser() parser for paloalto panOS logs

Add a new scl: panos-parser() parser. It reads and automatically parses the paloalto panos logs.
You can override prefix for the created name-value pairs using the prefix() parameter.
Example:
parser p_panos-parser {
	panos-parser(
	prefix("test."));
	};

Signed-off-by: MileK <mileek@gmail.com>

### DIFF
--- a/scl/paloalto/panos.conf
+++ b/scl/paloalto/panos.conf
@@ -53,7 +53,8 @@ block parser panos-parser (prefix(".panos.") ...) {
     elif (match('CONFIG', value('`prefix`type'))) {    
       parser {
         csv-parser(
-          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","host_name","vsys","command","admin","client","result","configuration_path","sequence_number","action_flags","before_change_detail","after_change_detail","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
+          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","host_name","vsys","command","admin","client","result","configuration_path","sequence_number","action_flags","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
+          # Note if using custom logs insert "before_change_detail","after_change_detail" after "action_flags"
           prefix("`prefix`")
           delimiters(',')
         );
--- a/scl/paloalto/panos.conf
+++ b/scl/paloalto/panos.conf
@@ -107,3 +107,7 @@ block parser panos-parser (prefix(".panos.") ...) {
     };
   };
 };
+application panos[syslog] {
+    filter { message(".*,paloalto.*"); };
+    parser { panos-parser(); };
+};
--- a/scl/paloalto/panos.conf
+++ b/scl/paloalto/panos.conf
@@ -1,0 +1,108 @@
+#############################################################################
+# Copyright (c) 2020 Balabit
+# Copyright (c) 2020 MileK <mileek@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+#
+#scl/paloalto/panos.conf -- Paloalto PAN-OS parser fro syslog-ng
+#
+#<12>Apr 14 16:48:54 paloalto.test.net 1,2020/04/14 16:48:54,unknown,SYSTEM,auth,0,2020/04/14 16:48:54,,auth-fail,,0,0,general,medium,failed authentication for user \'admin\'. Reason: Invalid username/password. From: 10.0.10.55.,1718,0x0,0,0,0,0,,paloalto
+#<14>Apr 14 16:54:18 paloalto.test.net 1,2020/04/14 16:54:18,unknown,CONFIG,0,0,2020/04/14 16:54:18,10.0.10.55,,set,admin,Web,Succeeded, deviceconfig system,127,0x0,0,0,0,0,,paloalto
+
+block parser panos-parser (prefix(".panos.") ...) {
+  channel {
+    rewrite {
+      set("${LEGACY_MSGHDR}${MESSAGE}" value("MESSAGE"));
+      set("paloalto_panos" value("PROGRAM"));
+    };
+# Common fields - set dot-nv-pairs
+    parser {
+      csv-parser(
+        columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time")
+        delimiters(',')
+        prefix("`prefix`")
+      );
+    };
+# Parse logs according to "type" field
+    if (match('SYSTEM', value('`prefix`type'))) {
+      parser {
+        csv-parser(
+          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","vsys","event_id","object","future_use3","future_use4","module","severity","description","sequence_number","action_flags","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
+          prefix("`prefix`")
+          delimiters(',')
+        );
+      };
+    } 
+    elif (match('CONFIG', value('`prefix`type'))) {    
+      parser {
+        csv-parser(
+          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","host_name","vsys","command","admin","client","result","configuration_path","sequence_number","action_flags","before_change_detail","after_change_detail","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
+          prefix("`prefix`")
+          delimiters(',')
+        );
+      };
+    } 
+    elif (match('THREAT', value('`prefix`type'))) {
+      parser {
+        csv-parser(
+          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","src_ip","dest_ip","src_translated_ip","dest_translated_ip","rule","src_user","dest_user","app","vsys","src_zone","dest_zone","src_interface","dest_interface","log_forwarding_profile","future_use3","session_id","repeat_count","src_port","dest_port","src_translated_port","dest_translated_port","session_flags","transport","action","misc","threat","raw_category","severity","direction","sequence_number","action_flags","src_location","dest_location","future_use4","content_type","pcap_id","file_hash","cloud_address","url_index","user_agent","file_type","xff","referrer","sender","subject","recipient","report_id","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
+          prefix("`prefix`")
+          delimiters(',')
+        );
+      };
+    } 
+    elif (match('TRAFFIC', value('`prefix`type'))) {
+      parser {
+        csv-parser(
+          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","src_ip","dest_ip","src_translated_ip","dest_translated_ip","rule","src_user","dest_user","app","vsys","src_zone","dest_zone","src_interface","dest_interface","log_forwarding_profile","future_use3","session_id","repeat_count","src_port","dest_port","src_translated_port","dest_translated_port","session_flags","transport","action","bytes","bytes_out","bytes_in","packets","start_time","duration","http_category","future_use4","sequence_number","action_flags","src_location","dest_location","future_use5","packets_out","packets_in","session_end_reason","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
+          prefix("`prefix`")
+          delimiters(',')
+        );
+      };
+    } 
+    elif (match('HIPWATCH', value('`prefix`type'))) {
+      parser {
+        csv-parser(
+          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","src_user","vsys","host_name","os","src_ip","hip_name","hip_count","hip_type","future_use3","future_use4","sequence_number","action_flags","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
+          prefix("`prefix`")
+          delimiters(',')
+        );
+      };
+    } 
+    elif (match('CORRELATION', value('`prefix`type'))) {
+      parser {
+        csv-parser(
+          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","src_ip","src_user","vsys","category","severity","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name","vsys_id","object","object_id","evidence")
+          prefix("`prefix`")
+          delimiters(',')
+        );
+      };
+    } 
+    elif (match('USERID', value('`prefix`type'))) {
+      parser {
+        csv-parser(
+          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","vsys","src_ip","source_name","event_id","repeat_count","timeout_threshold","src_port","dest_port","source","source_type","sequence_number","action_flags","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
+          prefix("`prefix`")
+          delimiters(',')
+        );
+      };
+    };
+  };
+};
--- a/news/feature-3234.md
+++ b/news/feature-3234.md
@@ -1,0 +1,16 @@
+@include "scl.conf"
+
+log {
+  source { network(transport("udp")); };
+  parser { panos-parser(); };
+  destination {     
+	elasticsearch-http(
+	index("syslog-ng-${YEAR}-${MONTH}-${DAY}")
+	type("")
+	url("http://localhost:9200/_bulk")
+	template("$(format-json
+	--scope rfc5424 
+	--scope dot-nv-pairs --rekey .* --shift 1 --exclude *future_* --exclude *devicegroup_* 
+	--scope nv-pairs --exclude DATE --key ISODATE @timestamp=${ISODATE})")
+    );};
+};
--- a/scl/CMakeLists.txt
+++ b/scl/CMakeLists.txt
@@ -22,6 +22,7 @@ set(SCL_DIRS
     nodejs
     osquery
     pacct
+    paloalto
     rewrite
     slack
     snmptrap
--- a/scl/Makefile.am
+++ b/scl/Makefile.am
@@ -22,6 +22,7 @@ SCL_SUBDIRS	= \
 	nodejs		\
 	osquery	\
 	pacct		\
+  paloalto \
 	rewrite	\
 	slack \
 	snmptrap	\
--- a/scl/Makefile.am
+++ b/scl/Makefile.am
@@ -22,7 +22,7 @@ SCL_SUBDIRS	= \
 	nodejs		\
 	osquery	\
 	pacct		\
-  paloalto \
+	paloalto 	\
 	rewrite	\
 	slack \
 	snmptrap	\


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

szemere:
@kira-syslogng ok to test

gaborznagy:
@kira-syslogng ok to test;

kira-syslogng:
Build SUCCESS

scl/paloalto/panos.conf
```diff
@@ -0,0 +1,97 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+# Requires syslog(transport("udp")) as source
```
furiel, line 22: Can you explain how does this depend on udp?

scl/paloalto/panos.conf
```diff
@@ -0,0 +1,97 @@
+#############################################################################
+# Copyright (c) 2020 Balabit
+# created by mileek@gmail.com
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
```
furiel, line 8: So that this could be installed and packaged: please edit `CMakeLists.txt` and `Makefile.am` in the scl directory.

scl/paloalto/panos.conf
```diff
@@ -0,0 +1,97 @@
...
+        csv-parser(
+        columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","vsys","src_ip","source_name","event_id","repeat_count","timeout_threshold","src_port","dest_port","source","source_type","sequence_number","action_flags","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
+        prefix("`prefix`")
+        delimiters(',')
+      );
+    };
+  };
+  rewrite {
+    set("paloalto_panos", value("PROGRAM"));
+    set("syslog", value("SOURCE"));};
```
furiel, line 95: Is it necessary to override the source? It is a nice macro that refers back to the config file.

scl/paloalto/panos.conf
```diff
@@ -0,0 +1,97 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
```
furiel, line 21: A few conditional parses are happening in this file. Could you provide a sample input as comments here, so that we can write a test for it?

Or, if you want to try to write a test yourself, you can get some ideas from https://github.com/syslog-ng/syslog-ng/pull/3056/

scl/paloalto/panos.conf
```diff
@@ -0,0 +1,97 @@
...
+# Copyright (c) 2020 Balabit
+# created by mileek@gmail.com
+# This program is free software; you can redistribute it and/or modify it
```
furiel, line 4: The copyright check is failing here. Could you please double check it? You can try to copy from another file, and insert your name similarly as here: https://github.com/syslog-ng/syslog-ng/blob/06df1d712cf0608845eba0662e2cba7f85703af8/scl/graphite/plugin.conf#L3

The copyright script is very strict, all whitespaces, enters matter. If you get stuck, let me know, and I try to find out what went wrong.

scl/paloalto/panos.conf
```diff
@@ -0,0 +1,97 @@
...
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
```
furiel, line 6: The commit check also fails on this PR. There is a strict format that is asserted in the CI.

You can get general idea from: https://github.com/syslog-ng/syslog-ng/blob/06df1d712cf0608845eba0662e2cba7f85703af8/CONTRIBUTING.md#L88

In your case, the module part is missing. For example, the first line could look like:
```
panos-parser: new source as SCL
```

Also, this is a parser, isn't it? You could write parser instead of source in the commit message.

scl/paloalto/panos.conf
spricer: Parser has been tested only when PaloAlto sends logs as RFC3164 (legacy or BSD-syslog) . If you set it to RFC5424 (IETF) then I am not sure how it will behave. I though there should be some  warning in this regards.

scl/paloalto/panos.conf
spricer: If I do not set source I get next i my output message:
`"#anon-source0"`

scl/paloalto/panos.conf
spricer: OK

scl/paloalto/panos.conf
spricer: Currently I can only provide log for two cases:

SYSTEM message:
`<12>Apr 14 16:48:54 paloalto.test.net 1,2020/04/14 16:48:54,unknown,SYSTEM,auth,0,2020/04/14 16:48:54,,auth-fail,,0,0,general,medium,failed authentication for user \'admin\'.  Reason: Invalid username/password. From: 10.0.10.55.,1718,0x0,0,0,0,0,,paloalto`

CONFIG message:
`<14>Apr 14 16:54:18 paloalto.test.net 1,2020/04/14 16:54:18,unknown,CONFIG,0,0,2020/04/14 16:54:18,10.0.10.55,,set,admin,Web,Succeeded, deviceconfig system,127,0x0,0,0,0,0,,paloalto`

scl/paloalto/panos.conf
spricer: OK

scl/paloalto/panos.conf
spricer: Not sure what I need to edit in `Makefile.am`

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

scl/paloalto/panos.conf
```diff
@@ -0,0 +1,100 @@
...
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+#
+#scl/paloalto/panos.conf -- Paloalto PAN-OS parser fro syslog-ng
+
+block parser panos-parser (prefix(".panos.") source("syslog")...) {
```
Kokan, line 26: Just sharing an idea, act upon on this if you feel like.
Your parser now one big parser, that does what it needs to, but imho it would be nice to split it up into different section for easier reading. You have two basic section, one handling common part of the log parser, and the different emitter specific one. So you could create a `panos-common-parser`, and a `panos-application-specific` parser. The later would deal with the big block of IFs.

```
block parser panos-parser(...)
{
   panos-common-parser(...);
   panos-app-specific-parser();
}

block parser panos-app-specific-parser(...)
{
...
# Common fields 
    parser {
      csv-parser(
        columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time")
      delimiters(',')
      prefix("`prefix`")
     );
   };
...
};


block parser panos-common-parser(...)
{
# Parse logs according to panos.type field
    if (match('SYSTEM', value('.panos.type'))) {
    ...
};
```

The other idea that could help with reading and maintaining this parser would be to use application parser/appmodel.
Instead of listing all of those IFs, you could just say that you want all of the parsers that has certain topic:
```
block parser panos-parser(...)
{
   panos-common-parser(...);
   app-parser(topic(panos-application));
}
```
In this example a new topic: `panos-application`. This parser going to try all of the `panos-application` parsers.
The question is how syslog-ng could know which is a `panos-application` parser, and the answer you must specify them:

```
application panos-system[panos-application] {
  filter { match('SYSTEM', value('.panos.type'); };
  parser {
     csv-parser(
        columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time")
      delimiters(',')
      prefix("`prefix`")
     );
 };
};

application panos-config[panos-application] {
  filter { match('CONFIG', value('.panos.type'); };
  parser {
csv-parser(
columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","host_name","vsys","command","admin","client","result","configuration_path","sequence_number","action_flags","before_change_detail","after_change_detail","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
        prefix("`prefix`")
        delimiters(',')
        ); 
 };
};
...
```

With the `application some-name[topic-to-use-this-parser]` you can extend the list of parser easily.


Oh and btw, there are already topics for the `default-network-drivers` which you could check out for examples, and actually also include to that your parser probably `syslog-raw` would be a good candidate:

```
application panos[syslog-raw] {
    filter { # here you should put something that is uniq for panos logs, i do not have the knowhow for that
    };
   parser { panos-parser(); }
};

File: scl/paloalto/panos.conf (hunk 2 above)
Comment (furiel) [reply scl/paloalto/panos.conf:9]: You need to add the directory to `SCL_SUBDIRS`, similarly to the other scl-s.

That variable will be later used to fill the `EXTRA_DIST` variable, which will make sure these directories end up in the tarball.

File: scl/paloalto/panos.conf (hunk 6 above)
Comment (furiel) [reply scl/paloalto/panos.conf:7]: Unfortunately that blog is outdated. commit-check.sh is introduced one year later.

@czanik could you update the commit message example the referred blogpost? We have a hard requirement on the commit messages since then. For example: the first line might be:
```
scl: new linux-audit() source
```

File: scl/paloalto/panos.conf (hunk 1 above)
Comment (furiel) [reply scl/paloalto/panos.conf]: So this comment is wrong in the opposite way? `syslog` source referns to `RFC5424`.
It would be the `network` source that refers to `RFC3164`.

If the only requirement is the message format, just comment the RFC here, instead of specifying the source. It will help us later include your parser into the app-parser framework.

File: scl/paloalto/panos.conf (hunk 3 above)
Comment (furiel) [reply scl/paloalto/panos.conf]: That should not be a parameter either. That macro has a[ documented](https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.25/administration-guide/64#TOPIC-1349537) behavior. It should not be used for different purpose.

What is your goal here? Maybe there is another way to achieve that. Do you want to refer to the originating message format? Or just the fact that this comes from a syslog-ng instance? What will it be used for? Is it okay to set it into a custom field, like `prefix.<a good a value name>`? What would be a good name? `source` seems misleading for me, because the true source is pan-os. But it can arrive from many other sources, for example from file.

File: scl/paloalto/panos.conf (hunk 1 above)
Comment (furiel) [reply scl/paloalto/panos.conf]: Oh I did not see you modified the patchset in the meantime, and removed the comment. Marking as resolved.

File: scl/paloalto/panos.conf (hunk 4 above)
Comment (furiel) [reply scl/paloalto/panos.conf:22]: That's nice, even if not complete coverage. Please add these two as comment, [similarly](https://github.com/syslog-ng/syslog-ng/blob/06df1d712cf0608845eba0662e2cba7f85703af8/scl/cisco/plugin.conf#L33) to the other scl-s.

File: scl/paloalto/panos.conf (hunk 6 above)
Comment (czanik) [reply scl/paloalto/panos.conf:7]: Sure. I put it on my ToDo list.

File: scl/paloalto/panos.conf (hunk 2 above)
Comment (spricer) [reply scl/paloalto/panos.conf:9]: I added dir to `CMakeLists.txt` but not sure what to do edit in `Makefile.am` as you mentioned.

File: scl/paloalto/panos.conf (hunk 3 above)
Comment (spricer) [reply scl/paloalto/panos.conf]: I guess I was using idea from
[Creating your first block for the syslog-ng configuration library (SCL)](https://www.syslog-ng.com/community/b/blog/posts/creating-your-first-block-for-the-syslog-ng-configuration-library-scl)
but that block was for source and this one is for parser. Nevertheless my use is specific with inline configuration so I will remove it in this submit just to avoid confusion.

File: scl/paloalto/panos.conf (hunk 7 above)
Comment (spricer) [reply scl/paloalto/panos.conf:26]: Hi Peter,
I forgot to introduce `prefix` variable into match `value` so after introducing my value match does not work anymore.
`filter { match("SYSTEM", value("`prefix`type"));};`
 I guess since `application` does not belong to panos-parser block.
It could be that I misunderstood your post but then again I could not find documentation in regards to either `app-parser` nor `application`.

E.g.
```
block parser panos-parser (prefix(".panos.") ...) {
  channel {
    rewrite {
      set("${LEGACY_MSGHDR}${MESSAGE}" value("MESSAGE"));
      set("paloalto_panos", value("PROGRAM"));
     };
# Common fields - set dot-nv-pairs
    parser {
      csv-parser(
        columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time")
        prefix("`prefix`")
        delimiters(',')
      );
    };
  };
  app-parser(topic(panos-application));
};

# Parse logs according to "type" field
# SYSTEM
application panos-system[panos-application] {
  filter { match("SYSTEM", value("`prefix`type"));};
  parser {
     csv-parser(
        columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","vsys","event_id","object","future_use3","future_use4","module","severity","description","sequence_number","action_flags","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
      delimiters(',')
      prefix("`prefix`")
     );
 };
};
```

**Comment by kira-syslogng** [comment]
Build SUCCESS

**Comment by kira-syslogng** [comment]
Build SUCCESS

**Comment by kira-syslogng** [comment]
Build SUCCESS

File: scl/paloalto/panos.conf (hunk 2 above)
Comment (furiel) [reply scl/paloalto/panos.conf:9]: Under https://github.com/syslog-ng/syslog-ng/blob/7f21db29622c67a00e514cfa45b3d38bfa884e15/scl/Makefile.am#L24
just add `paloalto \`

**Comment by kira-syslogng** [comment]
Build SUCCESS

**Comment by spricer** [comment]
There is unsupported commit message from me in `ee1450523` which was not parsed  by expected format from `Travis-CI` during commit. I tried to change message VIA rebase reword but it fails so I am not sure what else I can do:
```
Last commands done (19 commands done):
   pick a23f4c8fc scl: new panos-parser() parser for paloalto panOS logs
   pick 61a4e8236 logpipe: remove multi-line-timeout from logpipe source
  (see more in file .git/rebase-merge/done)
Next commands to do (68 remaining commands):
   pick fae6ae1cd rewrite: added conditional set-tag rule
   pick 3ffd4db4e news: add feature-3190
  (use "git rebase --edit-todo" to view and edit)
You are currently rebasing branch 'master' on '82ed6fa5d'.
  (all conflicts fixed: run "git rebase --continue")

nothing to commit, working tree clean
The previous cherry-pick is now empty, possibly due to conflict resolution.
If you wish to commit it anyway, use:

    git commit --allow-empty

Otherwise, please use 'git cherry-pick --skip'
Could not apply 61a4e8236... logpipe: remove multi-line-timeout from logpipe source
```

**Comment by spricer** [comment]
Hi Nazar,
I restarted SOA just in case *alter_spapi* was done after patches were
deployed.

Regards,

     ╠═MileK═╣

**Comment by kira-syslogng** [comment]
Build SUCCESS

**Comment by kira-syslogng** [comment]
Build SUCCESS

**Comment by furiel** [comment]
@spricer could you please double check the `CONFIG` example? The `SYSTEM` example looks good to me.

However when I tried the `CONFIG`, I receive:

```
{
  "_panos": {
    "vsys": "",
    "version": "0",
    "type": "CONFIG",
    "serial_number": "unknown",
    "sequence_number": "127",
    "result": "Succeeded",
    "receive_time": "2020/04/14 16:54:18",
    "log_subtype": "0",
    "host_name": "10.0.10.55",
    "generated_time": "2020/04/14 16:54:18",
    "future_use1": "1",
    "devicegroup_level4": "paloalto",
    "devicegroup_level3": "",
    "devicegroup_level2": "0",
    "devicegroup_level1": "0",
    "configuration_path": "deviceconfig system",
    "command": "set",
    "client": "Web",
    "before_change_detail": "0",
    "after_change_detail": "0",
    "admin": "admin",
    "action_flags": "0x0"
  },
  "SOURCE": "#anon-source0",
  "PROGRAM": "paloalto_panos",
  "MESSAGE": "1,2020/04/14 16:54:18,unknown,CONFIG,0,0,2020/04/14 16:54:18,10.0.10.55,,set,admin,Web,Succeeded, deviceconfig system,127,0x0,0,0,0,0,,paloalto",
  "HOST_FROM": "localhost",
  "HOST": "localhost"
}
```

For example: "vsys_name","dvc_name" does not exist. On the other hand, `devicegroup_level4` seems as if it should belong to `dvc_name`? Do some columns missing in the example? Or the list of columns is wrong in the scl?

**Comment by spricer** [comment]
Parser fields were copied from splunk repo...not excuse :) for not confirming. According to [PAN-OS Config Log Fields](https://docs.paloaltonetworks.com/pan-os/8-1/pan-os-admin/monitoring/use-syslog-for-monitoring/syslog-field-descriptions/config-log-fields.html) it seems that `Before Change Detail` and `After Change Detail` are non default fields and used only in custom logs. I removed them and added as note/comment.

**Comment by kira-syslogng** [comment]
Build SUCCESS

**Comment by furiel** [comment]
Thank you for the update! With the latest version, both `SYSTEM` and config `CONFIG` looks good.

There are two topics left: news file, and the app-parser.

For the news file: Please add a news entry under the news directory: into a file called `feature-3234.md`.

It's content can be a simple:

```
scl: new panos-parser() parser for paloalto panOS logs

@include "scl.conf"

log {
  source { network(port(5555)); };
  parser { panos-parser(); };
  destination { file("/dev/stdout" template("$(format-json --scope all-nv-pairs)\n")); };
};
```
For the app-parser part, I will update the conversation above soon.

File: scl/paloalto/panos.conf (hunk 7 above)
Comment (furiel) [reply scl/paloalto/panos.conf]: We discussed about this with the team, and concluded that we do not want to increase the number of app-parser topics at this point. On the other hand, `panos-parser` would nicely fit in the `syslog` topic.

> [...] I could not find documentation in regards to either `app-parser` nor `application`.

Yes, this is a bit unfortunate, that `app-parser` is not yet documented. It is relatively recent addition to syslog-ng, a bit experimental. On the other hand, it is very convenient feature from user experience point of view, so we would like to ask you to add `panos-parser()` to the syslog topic.

All you need to do is to add

```
application panos[syslog] {
    filter { message(".*,paloalto.*"); };
    parser { panos-parser(); };
};
```

to the end of the the scl file. 

You can test it with:

```
@version: 3.26

@include "scl.conf"

log {
  source { network(port(5555)); };
  parser { app-parser(topic(syslog)); };
  destination { file("/dev/stdout" template("$(format-json --scope all-nv-pairs)\n")); };
};
```

and by sending an example message:

```
echo "<14>Apr 14 16:54:18 paloalto.test.net 1,2020/04/14 16:54:18,unknown,CONFIG,0,0,2020/04/14 16:54:18,10.0.10.55,,set,admin,Web,Succeeded, deviceconfig system,127,0x0,0,0,0,0,,paloalto" | nc localhost 5555
```

The result is the same when you explicitely added `panos-parser` to the config, with a little extra: a new name value also appeared in the message besides the usual:

```
  "_app": {
    "name": "panos"
  },
```

Comment (Kokan) [reply scl/paloalto/panos.conf]: > We discussed about this with the team, and concluded that we do not want to increase the number of app-parser topics at this point.
Why?

**Comment by kira-syslogng** [comment]
Build SUCCESS

File: scl/paloalto/panos.conf [hunk 8]
```diff
@@ -0,0 +1,113 @@
...
+          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","vsys","src_ip","source_name","event_id","repeat_count","timeout_threshold","src_port","dest_port","source","source_type","sequence_number","action_flags","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
+          prefix("`prefix`")
+          delimiters(',')
+        );
+      };
+    };
+  };
+};
+application panos[syslog] {
+    filter { message(".*,paloalto.*"); };
```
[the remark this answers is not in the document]
gaborznagy: Maybe we can find something unique to use.

As Kokan mentioned before, the parser has 2 parts: a generic one (where type is found) and a category-specific one (SYSTEM, etc.).
Maybe we can use the generic one to check the parsed "type" field against known categories (SYSTEM, CONFIG, THREAT, etc.). If it matches, we know we have a Pan-OS log.

I'll think about it.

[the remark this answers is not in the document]
bazsi: This does not need to be perfect. the panos-parser() block would drop messages if they don't match the expected format (the big if/elif/... construct does not have an else block at the end, so if none of the branches match the log would be dropped).

In addition to this, I would probably add a flags(drop-invalid) option to the initial csv-parser() and move the rewrite rules toward the end (e.g. the setting of $PROGRAM and $MESSAGE), when we already proved the format was good.

Then I would simply not use a filter in the application adapter. What this would cause is that we would try to parse the incoming log as a panos specific log which would hopefully fail already at csv-parser, or in case it passes that, we would quickly exit when checking it against the recognized values (SYSTEM, CORRELATION, etc).

A good filter would be very useful, but that's only a performance improvement. It could be as easy as matching the first couple of characters with a regexp, e.g. message("1,[0-9]+/[0-9]+/[0-9]+ "); the pluses in that regexp could be changed to the exact number of digits, using the pcre syntax I can't recall right now (IIRC you can specify repetition count in braces).

With the filter in place and the panos-parser() dropping unrecognized log messages we should be fine.

If another device logs like panos and we recognize as panos, that can possibly be a problem, but let's wait for the conflict to arise.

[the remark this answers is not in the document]
gaborznagy: Thanks Bazsi for the feedback.

>  I would probably add a flags(drop-invalid) option to the initial csv-parser() 

I don't think you can, as that initial parser matches on a line where are more columns then configured, and we don't use the "greedy" flag, which would result in dropping the message at the first step.

> move the rewrite rules toward the end (e.g. the setting of $PROGRAM and $MESSAGE), when we already proved the format was good.

That was my initial idea too, but then realized the rewrite rule initializes the MESSAGE macro for the csv-parsers. It wouldn't be needed if we can solve the syslog() parsing step.
Basically, panos messages are csv fields added onto a semi-syslog message (we have a valid hostname, but after that the csv fields are starting.)

See example message from PR:
```
<14>Apr 14 16:54:18 paloalto.test.net 1,2020/04/14 16:54:18,unknown,CONFIG,0,0,2020/04/14 16:54:18,10.0.10.55,,set,admin,Web,Succeeded, deviceconfig system,127,0x0,0,0,0,0,,paloalto
```

bazsi:
On Wed, May 6, 2020 at 1:57 PM Gábor Nagy <notifications@github.com> wrote:

> *@gaborznagy* commented on this pull request.
>
> In scl/paloalto/panos.conf
> <https://github.com/syslog-ng/syslog-ng/pull/3234#discussion_r420733609>:
>
> > +        );
> +      };
> +    }
> +    elif (match('USERID', value('`prefix`type'))) {
> +      parser {
> +        csv-parser(
> +          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","vsys","src_ip","source_name","event_id","repeat_count","timeout_threshold","src_port","dest_port","source","source_type","sequence_number","action_flags","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
> +          prefix("`prefix`")
> +          delimiters(',')
> +        );
> +      };
> +    };
> +  };
> +};
> +application panos[syslog] {
> +    filter { message(".*,paloalto.*"); };
>
> Thanks Bazsi for the feedback.
>
> I would probably add a flags(drop-invalid) option to the initial
> csv-parser()
>
> I don't think you can, as that initial parser matches on a line where are
> more columns then configured, and we don't use the "greedy" flag, which
> would result in dropping the message at the first step.
>

if we there can be more fields than configured, we should add the greedy
flag.


> move the rewrite rules toward the end (e.g. the setting of $PROGRAM and
> $MESSAGE), when we already proved the format was good.
>
> That was my initial idea too, but then realized the rewrite rule
> initializes the MESSAGE macro for the csv-parsers. It wouldn't be needed if
> we can solve the syslog() parsing step.
> Basically, panos messages are csv fields added onto a semi-syslog message
> (we have a valid hostname, but after that the csv fields are starting.)
>

you can specify the input of the csv-parser() using template(), so you
don't need to change $MSG.


-- 
Bazsi

bazsi:
just fyi, I am working on this on my private branch.

spricer:
OK Bazsi. As for rewriting log at the beginning in my case I am not overall concerned since I am using junction as entry point which sends panos device to particular parser.

furiel:
> just fyi, I am working on this on my private branch.

@bazsi I created earlier a branch with a single commit on the top of this branch that adds an e2e test for panos parser with the two currently known example messages.
https://github.com/furiel/syslog-ng/tree/panos-test
and
https://github.com/furiel/syslog-ng/commit/059c0633b16b08d2b9b499982e7c739b63b00db8

I will update the test according to your changes, no problem. Just wanted to let you know, to avoid possible double work, in case you wanted to create a test too.

scl/paloalto/panos.conf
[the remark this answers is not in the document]
bazsi: If you look at my branch superseding this one, I've added explicit filters to drop formats that do not confirm to the format. I chose to require fields up to "device_name" which is common in all of the formats and should never be zero. Newer versions extend the format with extra fields, those are also added but remained optional.

scl/paloalto/panos.conf
bazsi: I didn't see too much value in allowing users to extend the set of PanOS log types supported. If something is not, it should be added to the panos-parser() and not into local configuration files.

bazsi:
> > just fyi, I am working on this on my private branch.
> 
> @bazsi I created earlier a branch with a single commit on the top of this branch that adds an e2e test for panos parser with the two currently known example messages.
> https://github.com/furiel/syslog-ng/tree/panos-test
> and
> [furiel@059c063](https://github.com/furiel/syslog-ng/commit/059c0633b16b08d2b9b499982e7c739b63b00db8)
> 
> I will update the test according to your changes, no problem. Just wanted to let you know, to avoid possible double work, in case you wanted to create a test too.

I included your testcase (thanks!) in the branch based on this one. #3346

bazsi:
I am closing this in favour of #3346, let me know if you think I missed something.

@spricer I would appreciate testing of the new branch, or if you could I would be very gradeful for some actual log samples that could go along the parser. Thanks.

PR #3236, opened 2020-04-14
Improve func-test

szemere:
1. While investigating func-test errors - It was easier for me to see the list of failed test cases after each run.
2. Checking logstore_reader was a dead code, since we returned from the loop on first iteration.

### DIFF
--- a/tests/functional/func_test.py
+++ b/tests/functional/func_test.py
@@ -101,14 +101,13 @@ import test_python
 import test_map_value_pairs
 
 tests = (test_input_drivers, test_sql, test_file_source, test_filters, test_performance, test_python, test_map_value_pairs)
+failed_tests = []
 
 init_env()
 seed_rnd()
 
 
 verbose = False
-success = True
-rc = 0
 if len(sys.argv) > 1:
     verbose = True
 try:
@@ -123,17 +122,24 @@ try:
             if obj[:5] != 'test_':
                 continue
             test_case = getattr(test_module, obj)
+
             if type(test_module.config) is str:
-                success = run_testcase(test_module.__name__ + '.' + obj, test_module.config, verbose, test_case)
+                test_case_name = test_module.__name__ + '.' + obj
+                if not run_testcase(test_case_name, test_module.config, verbose, test_case):
+                    failed_tests.append(test_case_name)
+
             elif type(test_module.config) is dict:
-                success = True
                 for config_name in test_module.config:
                     testcase_name = "%s.%s[%s]" %(test_module.__name__, obj, config_name)
                     config = test_module.config[config_name]
-                    success = run_testcase(testcase_name, config, verbose, test_case) and success
-            if not success:
-                rc = 1
+                    if not run_testcase(testcase_name, config, verbose, test_case):
+                        failed_tests.append(testcase_name)
+
 finally:
     stop_syslogng()
 
-sys.exit(rc)
+if len(failed_tests)>0:
+    print("List of failed test cases: %s" % ','.join(failed_tests))
+    sys.exit(1)
+
+sys.exit(0)
--- a/tests/functional/messagecheck.py
+++ b/tests/functional/messagecheck.py
@@ -25,12 +25,6 @@ from log import *
 from control import *
 from messagegen import syslog_prefix
 
-def logstore_reader(fname):
-    try:
-        return os.popen("../../src/logcat %s.lgs" % fname, "r")
-    except OSError as e:
-        print_user("Error opening file: %s, %s" % (fname, str(e)))
-
 def file_reader(fname):
     try:
         return open(fname + ".log", "r")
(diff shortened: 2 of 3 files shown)


kira-syslogng:
Build SUCCESS

tests/functional/messagecheck.py
```diff
@@ -110,12 +110,13 @@ def check_file_expected(fname, messages, settle_time=1, syslog_prefix=syslog_pre
...
+    if not check_reader_expected(file_reader(fname), messages, settle_time, syslog_prefix, skip_prefix):
+        return False
+
     if logstore_store_supported:
-        iter = (file_reader, logstore_reader)
-    else:
-        iter = (file_reader,)
-    for reader in iter:
-        return check_reader_expected(reader(fname), messages, settle_time, syslog_prefix, skip_prefix)
+        return check_reader_expected(logstore_reader(fname), messages, settle_time, syslog_prefix, skip_prefix)
```
MrAnno, line 117: If we want to keep this in OSE, I think we should at least check this functional test internally.

tests/functional/func_test.py
```diff
@@ -89,6 +89,8 @@ def run_testcase(test_name, config, verbose, test_case):
     if not stop_syslogng():
       sys.exit(1)
     print_end(test_name, success)
+    if not success:
+        failed_tests.append(test_name)
```
MrAnno, line 93: This is useful, thanks.

Optional (just an opinion): I'd probably append the `failed_tests` global outside to avoid using more globals in these functions.

kira-syslogng:
Build SUCCESS

tests/functional/messagecheck.py
furiel: I tried this one. 

There are two modes:
- in tree
There is a heuristic that collects the module directories. This collects libraries only under core, so for example the license dir is not accessible, so syslog-ng cannot start:
https://github.com/syslog-ng/syslog-ng/blob/5d75d3cb5de1f15f5113dd97f86e563ee8465d1c/tests/functional/globals.py#L34
- out of tree
With out of tree (exporting `SYSLOG_NG_BINARY`) this still does not work out of the box, because `logcat` has invalid path:
https://github.com/syslog-ng/syslog-ng/blob/5d75d3cb5de1f15f5113dd97f86e563ee8465d1c/tests/functional/messagecheck.py#L30

After I hardcoded the correct path just for the sake for POC, there is still a problem. Some of the tests are prepared to be used with logstore, some of them don't. The idea is that all test_module.config strings must contain a proper logstore destination, that is just removed in case of ose at
https://github.com/syslog-ng/syslog-ng/blob/5d75d3cb5de1f15f5113dd97f86e563ee8465d1c/tests/functional/control.py#L41
On the other hand, the logstore reader expects this destination to exist. The problem is that  some tests do not provide such destinations. Those that provide do succeed, the others fail.
For example: `test_input_drivers` is ok:
https://github.com/syslog-ng/syslog-ng/blob/5d75d3cb5de1f15f5113dd97f86e563ee8465d1c/tests/functional/test_input_drivers.py#L45
but
`test_map_value_pairs` does not have logstore destination:
https://github.com/syslog-ng/syslog-ng/blob/5d75d3cb5de1f15f5113dd97f86e563ee8465d1c/tests/functional/test_map_value_pairs.py#L44

If we merge this in this current state, that will break the MI job, which executes func-test too.

MrAnno: @furiel Thank you very much!!

IMO, we shouldn't have implemented logstore tests in OSE, and knowing all the above makes me think that it's not worth keeping them here.

furiel: I agree with removing them.

kira-syslogng:
Build SUCCESS

tests/functional/messagecheck.py
```diff
@@ -109,13 +109,7 @@ def check_reader_expected(reader, messages, settle_time, syslog_prefix, skip_pre
...
     print_user("Checking contents of output files: %s" % fname)
     flush_files(settle_time)
-
-    if logstore_store_supported:
-        iter = (file_reader, logstore_reader)
-    else:
-        iter = (file_reader,)
-    for reader in iter:
-        return check_reader_expected(reader(fname), messages, settle_time, syslog_prefix, skip_prefix)
+    return check_reader_expected(file_reader(fname), messages, settle_time, syslog_prefix, skip_prefix)
```
alltilla, line 112: Can/should we remove `def logstore_reader(fname):` as well?

kira-syslogng:
Build SUCCESS

tests/functional/func_test.py
```diff
@@ -130,10 +129,15 @@ def run_testcase(test_name, config, verbose, test_case):
                 for config_name in test_module.config:
                     testcase_name = "%s.%s[%s]" %(test_module.__name__, obj, config_name)
                     config = test_module.config[config_name]
-                    success = run_testcase(testcase_name, config, verbose, test_case) and success
-            if not success:
-                rc = 1
+                    result = run_testcase(testcase_name, config, verbose, test_case)
```
gaborznagy, line 132: We have test cases with string config as well, please add failed testcases appending here as well.
```
            if type(test_module.config) is str:
                success = run_testcase(test_module.__name__ + '.' + obj, test_module.config, verbose, test_case)	                success = run_testcase(test_module.__name__ + '.' + obj, test_module.config, verbose, test_case)
            elif type(test_module.config) is dict:	            elif type(test_module.config) is dict:
                success = True	                success = True
```

kira-syslogng:
Build SUCCESS

Issue #3215, opened 2020-04-01
Add values(key1 => value1, ...) support for example-msg-generator


furiel:
`example-msg-generator` generates a message from template, which is very useful, when someone wants to build their config iteratively, with trial and error.

For now, only the message part can be specified by the `example-msg-generator`. If I want to specify the `program` field for example, I need to add an extra rewrite rule after `example-msg-generator` to set it.

It would be nice if these assignments could be specified inside the `example-msg-generator`. The list of name-value pairs could be specified with a new `values` option, using the arrow syntax. The value part of the name-value pairs need to be evaluated as templates.

For example
```
@version: 3.26
log {
  source { example-msg-generator(template("message part")
                                 num(1)
                                 values("PROGRAM" => "program-name"
                                        "current-second" => "$C_SEC"
                                        "quick-maths" => "2+2-1=$(- $(+ 2 2) 1)"
                                );
         };
  destination { file(/dev/stdout template("$(format-json --scope all-nv-pairs)\n")); };
};
```

vivinperis:
I would like to try this out. Can you please help me get started

furiel:
You can look for `LL_ARROW` in the code for examples that already use arrow syntax.
For setting things in message, you can check `log_msg_set_value_by_name` invocations. The template evaluation can be done in similar fashion as it is done for the message part in the `example-msg-generator`:
 https://github.com/syslog-ng/syslog-ng/blob/627e585d1e3c3afcc000b218612ea24d0131da51/modules/examples/sources/msg-generator/msg-generator-source.c#L108

PR #3233, opened 2020-04-11
example-message-generator: add support for values(name1 => value1, ...) syntax 

vivinperis:
fixes: https://github.com/syslog-ng/syslog-ng/issues/3215

PR #3237, opened 2020-04-15
 example-message-generator: add support for values(name1 => value1, ...) syntax where value follows template syntax

vivinperis:
fixes: #3215

### DIFF
--- a/modules/examples/sources/msg-generator/msg-generator-source.c
+++ b/modules/examples/sources/msg-generator/msg-generator-source.c
@@ -94,12 +94,26 @@ _free(LogPipe *s)
 }
 
 static void
+_add_name_value(gpointer key, gpointer value, gpointer data)
+{
+  gchar *name = (gchar *) key;
+  LogTemplate *val = (LogTemplate *) value;
+  LogMessage *msg = (LogMessage *) data;
+  GString *msg_body = g_string_sized_new(128);
+  log_template_format(val, msg, NULL, LTZ_LOCAL, 0, NULL, msg_body);
+  log_msg_set_value_by_name(msg, name, msg_body->str, msg_body->len);
+  g_string_free(msg_body, TRUE);
+
+}
+
+static void
 _send_generated_message(MsgGeneratorSource *self)
 {
   if (!log_source_free_to_send(&self->super))
     return;
 
   LogMessage *msg = log_msg_new_empty();
+  g_hash_table_foreach(self->options->name_value, _add_name_value, msg);
   log_msg_set_value(msg, LM_V_MESSAGE, "-- Generated message. --", -1);
 
   if (self->options->template)
--- a/modules/examples/sources/msg-generator/msg-generator.c
+++ b/modules/examples/sources/msg-generator/msg-generator.c
@@ -74,7 +74,7 @@ static void
 msg_generator_sd_free(LogPipe *s)
 {
   MsgGeneratorSourceDriver *self = (MsgGeneratorSourceDriver *) s;
-
+  g_hash_table_unref(self->source_options.name_value);
   msg_generator_source_options_destroy(&self->source_options);
   log_src_driver_free(s);
 }
@@ -92,9 +92,9 @@ msg_generator_sd_new(GlobalConfig *cfg)
 {
   MsgGeneratorSourceDriver *self = g_new0(MsgGeneratorSourceDriver, 1);
   log_src_driver_init_instance(&self->super, cfg);
-
   msg_generator_source_options_defaults(&self->source_options);
-
+  self->source_options.name_value = g_hash_table_new_full(g_str_hash, g_str_equal, g_free,
+                                                          (GDestroyNotify) log_template_unref);
   self->super.super.super.init = msg_generator_sd_init;
   self->super.super.super.deinit = msg_generator_sd_deinit;
   self->super.super.super.free_fn = msg_generator_sd_free;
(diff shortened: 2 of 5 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

alltilla:
@kira-syslogng ok to test;

alltilla:
related PR: #3233

kira-syslogng:
Build SUCCESS

modules/examples/sources/msg-generator/msg-generator.c
```diff
@@ -74,7 +74,8 @@ static void
 msg_generator_sd_free(LogPipe *s)
 {
   MsgGeneratorSourceDriver *self = (MsgGeneratorSourceDriver *) s;
-
+  g_hash_table_remove_all(self->source_options.name_value);
```
furiel, line 77: You do not need to remove all elements here. `g_hash_table_unref` below will do so.

modules/examples/sources/msg-generator/msg-generator.c
```diff
@@ -92,9 +93,8 @@ msg_generator_sd_new(GlobalConfig *cfg)
 {
   MsgGeneratorSourceDriver *self = g_new0(MsgGeneratorSourceDriver, 1);
   log_src_driver_init_instance(&self->super, cfg);
-
   msg_generator_source_options_defaults(&self->source_options);
-
+  self->source_options.name_value = g_hash_table_new_full(g_str_hash, g_str_equal, NULL, NULL);
```
furiel, line 97: You need to specify the destroy functions here, specifyng the destructor function for the keys and values: `g_free` and `log_template_unref`

modules/examples/sources/msg-generator/msg-generator-source.c
```diff
@@ -93,25 +94,39 @@ _free(LogPipe *s)
...
+
 static void
 _send_generated_message(MsgGeneratorSource *self)
 {
   if (!log_source_free_to_send(&self->super))
     return;
 
-  LogMessage *msg = log_msg_new_empty();
-  log_msg_set_value(msg, LM_V_MESSAGE, "-- Generated message. --", -1);
+  self->msg = log_msg_new_empty();
```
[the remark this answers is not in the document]
vivinperis: I used it as self because I am using it in another function when I'm assigning the name value pairs(Basically for g_hash_table_foreach). How else can I pass the value?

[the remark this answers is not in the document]
furiel: Instead of passing `self` to foreach, you can pass `msg` directly as `user_data`. There is no need to pass the entire source.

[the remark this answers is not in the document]
vivinperis: I had passed message earlier as user_data, But then there was memory leak when it came to hashmap value. I found this as a fix to that and it works fine

kira-syslogng:
Build SUCCESS

modules/examples/sources/msg-generator/msg-generator-source.c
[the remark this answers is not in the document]
vivinperis: Alright

kira-syslogng:
Build SUCCESS

modules/examples/sources/msg-generator/msg-generator-source.c
```diff
@@ -93,13 +93,27 @@ _free(LogPipe *s)
   log_source_free(s);
 }
 
+static void
+_add_name_value(gpointer key, gpointer value, gpointer data)
+{
+  gchar *name = (gchar *) key;
+  LogTemplate *val = (LogTemplate *) value;
+  LogMessage *msg = (LogMessage *) data;
+  GString *msg_body = g_string_sized_new(128);
```
alltilla, line 102: It would be good to use `ScratchMarkers` here.
Example: https://github.com/syslog-ng/syslog-ng/blob/master/modules/dbparser/correllation-context.c#L53
Key points are `Marker`, `alloc_and_mark()`, `reclaim_marked()`. Simple `alloc()` is not needed here, as we only need one `GString *`.

Or, even better, get only one `GString *` with `ScratchBuffers` in `send_generated_message()` and pass it to `_add_name_value()`, which will truncate it to 0 size before it uses it.

Each solution is okay for me, but `g_string_new()` is avoidable, if possible, because of performance reasons.

furiel: We were thinking about this together with @alltilla , and concluded that the original implementation is fine. The mark-reclaim api might work, but the simple scratch_buffers_alloc might not, because only `log_msg_post` and a few places are protected by garbage collection. The performance benefit does not worth the complexity. Marking as resolved.

bazsi: Although it is true that scratch buffers can save a lot of cycles for hot paths, I don't think this is one of those. msg-generator is at best a testing tool and I doubt it would generate a huge rate of messages per second and scratch buffers only play a role in those scenarios.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/examples/sources/msg-generator/msg-generator-source.c
alltilla: Sure, I agree, but as it is an "example" driver, it could show us an example, and it would be nice to see the building blocks of our real drivers. I might have the wrong idea about the examples folder.

I still agree with furiel though, as if we need to dig deep to make sure it works in this case, it is not worth the time and complexity.

alltilla:
Please open a new PR with the news entry ASAP, and link that PR in the description of this PR. Thanks :)

kira-syslogng:
Build SUCCESS

PR #3252, opened 2020-04-21
python_persist: fix test with python3

furiel:
In python3, integer and string cannot be sorted. So the python code throws exception, hence the testcase fails.

I forgot to add the test to `Makefile.am`, so in the end, the job that is responsible for checking unit tests with python3, did not execute this test. That's why we did not find this problem during code review.

Behind this, an actual coding error was hidden. `_py_is_string` returns true both for bytes and strings. So it should be executed last, when detecting the type of the value that is added to persist.

In python2, bytes and strings are the same, so no problem is there.

### DIFF
--- a/modules/python/tests/test_python_persist.c
+++ b/modules/python/tests/test_python_persist.c
@@ -180,7 +180,7 @@ persist = Persist('persist_name')\n\
 persist['integer'] = 1\n\
 persist['str'] = 'str'\n\
 persist['bytes'] = b'bytes'\n\
-assert sorted([1, 'str', b'bytes']) == sorted([persist[k] for k in persist])";
+assert set([1, 'str', b'bytes']) == set([persist[k] for k in persist])";
 
 Test(python_persist, test_python_persist_iter_returns_proper_types)
 {
--- a/modules/python/python-persist.c
+++ b/modules/python/python-persist.c
@@ -409,12 +409,12 @@ _py_persist_type_set(PyObject *o, PyObject *k, PyObject *v)
       return -1;
     }
 
-  if (_py_is_string(v))
+  if (PyBytes_Check(v))
+    type = ENTRY_TYPE_BYTES;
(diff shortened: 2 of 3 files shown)


kira-syslogng:
Build SUCCESS

modules/python/tests/test_python_persist.c
```diff
@@ -180,7 +180,7 @@ persist = Persist('persist_name')\n\
 persist['integer'] = 1\n\
 persist['str'] = 'str'\n\
 persist['bytes'] = b'bytes'\n\
-assert sorted([1, 'str', b'bytes']) == sorted([persist[k] for k in persist])";
```
alltilla, line 183: Can you give a brief explanation about the reason of changing `sorted` to `set`?
In my understanding the difference in this case is that `set` removes the duplicated elements. Is that expected?

furiel: The reason is: in python3, `sorted` does not work on lists that have both integer and string arguments. Originally, sorted was needed because the right side is a hashtable, and it can return the entries in any order. In python2, `sorted` worked. With using `set`, I can compare the two objects easily.

The new version does not detect bugs when values and keys are shuffled. We partially defend against this in another testcase:
https://github.com/syslog-ng/syslog-ng/blob/3a250f4c2dbcd5554760166ded86542459ca2acb/modules/python/tests/test_python_persist.c#L154
It also does not defend against if the left side has an unknown key, which overlaps with a known value. I was thinking this test is good enough when trading complexity, but of course, using `sorted` with a moderately complex key function: for example map to `[str(type(v)), str(v))]`, this can be resolved, if you think it would worth doing so. Or may be there are easier alternatives.

alltilla: We can control the correct inputs for this test, so this is fine, I think.

Thanks for the explanation :)

PR #3265, opened 2020-04-23
fix GError handling

szemere:
- grammar: there are two different macros for error checking (namely: `CHECK_ERROR` and `CHECK_ERROR_GERROR`), we missed using the GError version at some places
- other: add missing `g_error_free` calls.

### Question for reviewers:
- Should we replace `g_clear_error` functions to `g_error_free` in case of local variables? `g_clear_error` makes an extra `=NULL` assignment, but it seems nitpicking for me to change them.

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -887,17 +887,17 @@ template_content_inner
         {
           GError *error = NULL;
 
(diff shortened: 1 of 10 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3270, opened 2020-04-26
timeutils: fix crash in %f conversion when non-numeric character is in usec field

gaborznagy:
e.g. with ".asd123"


kira-syslogng:
Build SUCCESS

PR #3279, opened 2020-05-05
dbld: fix dockerhub automated builds

bazsi:
I mistakenly moved the build hooks in a previous PR to "./dbld", whereas it
should have remained in "./dbld/images/hooks".

Also the cwd is set by dockerhub to dbld/images which is not what we want,
so change the script to the root of the source tree and invoke docker build
from there.

### DIFF
--- a/dbld/hooks/build
+++ b/dbld/hooks/build
@@ -1,9 +1,0 @@
-#!/bin/sh
-
-set +x
-set +e
-
-OS_PLATFORM=$(basename $DOCKERFILE_PATH .dockerfile)
-
-./dbld/prepare-image-build $OS_PLATFORM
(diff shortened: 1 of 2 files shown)


bazsi:
While working on #3277 I noticed I've broken dockerhub builds, which this PR intends to fix. You can see that this was actually built on dockerhub, here:

https://hub.docker.com/repository/docker/bazsi/syslog-ng-build-test/builds

Once this is merged the dockerhub build rules will need to be updated to use "/dbld" as context, like it is configured here:

https://hub.docker.com/repository/docker/bazsi/syslog-ng-build-test/builds/edit

Since this is breaking our docker image build on dockerhub, I'd appreciate a timely merge. Thanks.

kira-syslogng:
Build SUCCESS

MrAnno:
> Once this is merged the dockerhub build rules will need to be updated to use "/dbld" as context

Done. Thanks.

PR #3281, opened 2020-05-06
Logmsg deserialization not always updated handles when changed

Kokan:
The handles in the logmsg are generated by the log_msg_registry, that object is not persisted between syslog-ng restarts, the handles are assigned in the order of request (incremented) therefore updating a handle to a new value might be needed at logmsg deserialization.

The current solution had a bug, that it only updated if the last handle was not updated. But the handles must be updated if any handle value changed in the logmsg and the current log_msg_registry.

Note: I wanted to write an UT covering the case when the last handle is not changed, but as the values are checked with a foreach that would rely on the way foreach loops the entries, which imho is not nice and may make the test pass but not really check the issue. So I decided not to. If anybody thinks otherwise let me know.

### DIFF
--- a/lib/logmsg/logmsg-serialize-fixup.c
+++ b/lib/logmsg/logmsg-serialize-fixup.c
@@ -239,7 +239,8 @@ _fixup_entry(NVHandle old_handle, NVEntry *entry, NVIndexEntry *index_entry, gpo
   if (log_msg_is_handle_sdata(new_handle))
     _fixup_sdata_handle(state, old_handle, new_handle);
 
-  state->handle_changed = (new_handle != old_handle);
+  if (!state->handle_changed)
+    state->handle_changed = (new_handle != old_handle);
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build FAILURE

Kokan:
(that's probably due to force push after just a few sec)

kira-syslogng:
Build SUCCESS

Kokan:
@MrAnno yes it does, it was my dev brain thinking it implies. (I'll update.)

kira-syslogng:
Build SUCCESS

Kokan:
The trigger was an actual bug in the field.
Depending on the slice of messages deserialization produced different values for the same log message.

Kokan:
Only updated the news file, with mentioning it has effect on messages stored in disk buffer as @MrAnno  suggested.

kira-syslogng:
Build SUCCESS

PR #3283, opened 2020-05-08
fix snapsot version generation

furiel:
https://github.com/syslog-ng/syslog-ng/pull/3272/commits/bc10b1063dc1276fe4ea818cb471cef501e9fb29 was incomplete, it needs a little more work.

The mode variable needs to be quoted within `version.sh`. Strangely `[ ! = "release" ]` evaluates to 0 (which is true in bash), while for example  `[ ! "" = "release" ]` evaluates to 1 (which is false in bash).

Also, MODE needs to be made available for the version script.

After these changes. `dbld/rules pkg-tarball` will generate snapshot tarball, and `dbld/rules MODE=release pkg-tarball` generates release tarball

### DIFF
--- a/dbld/rules
+++ b/dbld/rules
@@ -19,7 +19,8 @@ TARBALL_IMAGE ?= ubuntu-bionic
 DEFAULT_DEB_IMAGE=$(DEFAULT_IMAGE)
 DEFAULT_RPM_IMAGE=centos-7
 DOCKER=docker
-VERSION=$(shell scripts/version.sh)
+MODE ?= snapshot
+VERSION=$(shell MODE=${MODE} scripts/version.sh)
 DOCKER_RUN_ARGS=-e USER_NAME_ON_HOST=$(shell whoami)	\
         --network=host --privileged \
 	--ulimit nofile=1024:1024 \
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
There is also a `release` dbld target that sets the `release` MODE.

Why do we need `MODE` in `version.sh`? `git describe` should return the appropriate version string when you execute it on a tagged commit.

czanik:
And what is ```pkg-tarball```? The built-in help only mentions ```tarball``` and that's what I used until now:
```
Development:
  bootstrap: bootstrap source from a git clone (e.g. autogen.sh and friends)
  shell: start a shell in the specified OS
  login: start a 2nd shell alongside an already running one
  root-login: start a 2nd shell (as root) alongside an already running one
  tarball: generate a tarball in a controlled environment (e.g. docker)
  release: generate a release deb package in a controlled environment (e.g. docker)
```

furiel:
@MrAnno Unfortunately we cannot execute it on a tagged commit at that point. The release tarball needs to be created right before pushing the release button, because we need to upload it as an artifact. The release tag will be added only by the release button.

@czanik this target also prepares the debian directory for deb packaging and the rpm spec file for deb packaging.
https://github.com/syslog-ng/syslog-ng/blob/master/dbld/generate-debian-directory
https://github.com/syslog-ng/syslog-ng/blob/master/dbld/generate-rpm-specfile

kira-syslogng:
Build SUCCESS

MrAnno:
I'm not really happy about "faking" a version when the `release` mode is enabled.

GitHub creates annotated tags AFAIK, so we might want to create a real tag manually anyway before creating the tarball.

bazsi:
I think the version number is a mandatory parameter to the "release"
procedure and should be specified by the user doing the release. I would
rather keep a fair distance between git and our build infrastructure. This
doesn't mean that pulling in information from git is a problem but in an
ideal world we would be able to do a release/snapshot with only a tarball.

Why is it better to tag in a release "in advance" to specify a version
number than to supply this as an argument?

MrAnno:
@bazsi Okay, thanks for the explanation, you convinced me.

The tag would have been better because we couldn't have created a tarball that claims to be a proper release but contains uncommitted (dirty) changes or commits that are not part of the actual release based on our git history (because an annotated tag is what I considered as a release).

PR #3292, opened 2020-05-15
secret-storage: remove additional rlimit check

### LINKED ISSUE #3181: secret_storage: UT fails inside syslog-ng-devshell docker image
# syslog-ng
## Version of syslog-ng
Compiled from source (master: 531d70bfbcf19a97a5278f324d141386d568ce7f)

## Platform
 * `docker` image: `balabit/syslog-ng-devshell:latest`
 * `docker` runs on `ubuntu-xenial` or `ubuntu-bionic`

# Issue
## Failure
`test_secret_storage` unit test fails
```
====================================================
   syslog-ng 3.26.1.55.g531d70b: ./test-suite.log
====================================================

# TOTAL: 150
# PASS:  149
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: lib/secret-storage/tests/test_secret_storage
==================================================

secret storage: cannot lock buffer : reason=len: 4096, errno: 12. Maybe RLIMIT_MEMLOCK is too small?
[----] lib/secret-storage/tests/test_secret_storage.c:282: Assertion failed: offending_key: key016
[FAIL] secretstorage::test_rlimit: (0.00s)
[====] Synthesis: Tested: 19 | Passing: 18 | Failing: 1 | Crashing: 0 
FAIL lib/secret-storage/tests/test_secret_storage (exit status: 1)
```

## Steps to reproduce
```
$ docker pull balabit/syslog-ng-devshell:latest
$ docker run -it balabit/syslog-ng-devshell:latest bash

# git clone https://github.com/syslog-ng/syslog-ng.git
# cd syslog-ng/
# ./autogen.sh
# ./configure --prefix=${HOME}/install/syslog-ng --disable-all-modules
# make
# make check
# cat test-suite.log
```

The problem is probably caused by an inefficient check here:
> lib/secret-storage/tests/test_secret_storage.c:279
```
  /* root is not restricted by rlimit */
  if (geteuid() == 0)
    {
      cr_assert(secret_storage_store_string(key_fmt, "value"), "offending_key: %s", key_fmt);
      cr_assert(secret_storage_subscribe_for_key("key000", secret_checker, "value"));
    }
  else
    {
      cr_assert_not(secret_storage_store_string(key_fmt, "value"), "offending_key: %s", key_fmt);
    }
```

alltilla:
These asserts are not precise enough, being root is not enough to
determine if rlimit is restricted or not.

We need to check if we have `CAP_IPC_LOCK` capability. We had a run
with that, but we had to depend on `SYSLOG_NG_ENABLE_LINUX_CAPS`,
which is unfortunate in a test, because if we user does not have libcap
the tests are not run, also we will never notice, if the test stops
running for any reason.

In reality these asserts test a linux functionality, and to make it run
correctly it introduces too much complexity for no real usefulness.

Fixes #3181

### DIFF
--- a/lib/secret-storage/tests/test_secret_storage.c
+++ b/lib/secret-storage/tests/test_secret_storage.c
@@ -274,18 +274,6 @@ Test(secretstorage, test_rlimit)
       cr_assert(secret_storage_store_string(key_fmt, "value"), "offending_key: %s, for_limit: %d", key_fmt, for_limit);
     }
 
-  sprintf(key_fmt, "key%03d", i);
-
-  /* root is not restricted by rlimit */
-  if (geteuid() == 0)
-    {
-      cr_assert(secret_storage_store_string(key_fmt, "value"), "offending_key: %s", key_fmt);
-      cr_assert(secret_storage_subscribe_for_key("key000", secret_checker, "value"));
-    }
-  else
-    {
-      cr_assert_not(secret_storage_store_string(key_fmt, "value"), "offending_key: %s", key_fmt);
-    }
   g_free(key_fmt);
 }
 


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/secret-storage/tests/test_secret_storage.c
```diff
@@ -276,8 +296,8 @@ Test(secretstorage, test_rlimit)
 
   sprintf(key_fmt, "key%03d", i);
 
-  /* root is not restricted by rlimit */
-  if (geteuid() == 0)
+#if SYSLOG_NG_ENABLE_LINUX_CAPS
```
Kokan, line 299: As much as I understand these asserts test linux functionality, I think maybe it is time to let them go. As with this patch if you do not have or want libcap they are not existing anyway.
Otherwise if you still want to keep these asserts, please leave a space for my second patch not yet on master that forces the false branch execution. There is a usecase where custome memory management act like the false branch: https://github.com/Kokan/syslog-ng/commit/275b8eef3cdb1a27ff50d616ea057e74bf617b1e, https://github.com/Kokan/syslog-ng/commit/e352f30b4b4e60a7f677af7892e069b2c7cf40bd my patches for that.

bazsi: I agree that introducing these kind of dependencies in unit tests is not really fortunate and we will never notice if the test stops executing for any reason.

Can't we try if CAP_LOCK works simply by trying it in the test program? Or use mocking?

alltilla: I agree with both of you, I think it would be the best to remove this, if it causes this much trouble and complexity. @furiel ?

furiel: Yes. Just remove it.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3300, opened 2020-05-22
test_secure_logging: use iterate to generate unique strings

furiel:
This small patch shows an example how we can use `iterate` and `example-msg-generator` to generate unique messages.

Prior this patch, the same message was sent to secure logging multiple times. Now a sequence number is added.

### DIFF
--- a/tests/python_functional/functional_tests/template_functions/slog/test_secure_logging.py
+++ b/tests/python_functional/functional_tests/template_functions/slog/test_secure_logging.py
@@ -26,7 +26,9 @@ import pytest
 
 from src.helpers.secure_logging.conftest import *  # noqa:F403, F401
 
-example_message = "example-message"
+seqnum = "$(iterate $(+ 1 $_) 0)"
+message_base = "example-message"
+example_message = "{}: {}".format(message_base, seqnum)
 num_of_messages = 3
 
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/template_functions/slog/test_secure_logging.py
```diff
@@ -41,7 +42,8 @@ def test_secure_logging(config, syslog_ng, slog):
     syslog_ng.start(config)
 
     logs = file_destination.read_logs(num_of_messages)
-    assert not any(map(lambda x: example_message in x, logs))
+    # test for no clear text
+    assert not any(map(lambda x: "example_message" in x, logs))
```
mitzkia, line 46: The "example_message" as a string in assertion does not match with the original input string which is: "example-message". Maybe it would be better to make an own variable for input test messag, as it will be referenced in final assert also

furiel: Fixed thanks. I introduced a variable: `message_base`.

kira-syslogng:
Build SUCCESS

PR #3301, opened 2020-05-22
map: new template function

furiel:
`map` template function applies a template function to all elements of a list. It returns the new list.

See documentation fragment for example and details.

CC: @jszigetvari

Referring: https://github.com/syslog-ng/syslog-ng/issues/2926

------------------

Future plans:

Originally, I wanted to add a `filter` template function together with map. However, `filter` turned out a bit more problematic. Unfortunately, the `context` is currently not passed to the filter `eval` functions. For example if you check: https://github.com/syslog-ng/syslog-ng/blob/d7d1149c17d9604d44f68e2ddf54d881a2ad460f/lib/filter/filter-cmp.c#L72
you can see `NULL` is passed, in the place of the context. Also, `context` is currently not part of the eval signature:
https://github.com/syslog-ng/syslog-ng/blob/d7d1149c17d9604d44f68e2ddf54d881a2ad460f/lib/filter/filter-expr.c#L45
This means, we cannot use `$_` in the `if` template function. This significantly limits what we could express with a `filter` template function. To fix this, I need to modify the `eval` signature and make sure context is passed around properly. That is a more intrusive change, so I decided to do it separately from `map`.

### DIFF
--- a/doc/doc-3301.md
+++ b/doc/doc-3301.md
@@ -1,0 +1,20 @@
+# `map` template function
+
+# Short description
+
+`map` template function applies a template function to all elements of a list. It returns the new list.
+
+# Config snippet
+
+```
+log {
+  source { example-msg-generator(template("0,1,2") num(1)); };
+  destination { file("/dev/stdout" template("$(map $(+ 1 $_) $MSG)\n")); };
+};
+```
+
+# Description
+
+`map` template function has two mandatory parameters. The first parameter is a template, that will be applied to each list elements. The second parameter is a string or template, that represents the list.
+
+You can use `$_` template to refer to the list elements inside the template function. If you are chaining `map` calls, `$_` always refers to the innermost of it's enclosing `map` functions.
--- a/modules/basicfuncs/tf-map.c
+++ b/modules/basicfuncs/tf-map.c
@@ -1,0 +1,97 @@
+/*
+ * Copyright (c) 2020 Balabit
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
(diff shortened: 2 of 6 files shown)


kira-syslogng:
Build SUCCESS

modules/basicfuncs/tf-map.c
```diff
@@ -0,0 +1,109 @@
...
+  LogTemplate *template;
+} MapState;
+
+static gboolean
+tf_map_prepare(LogTemplateFunction *self, gpointer s, LogTemplate *parent, gint argc, gchar *argv[],
+               GError **error)
+{
+  MapState *state = (MapState *)s;
+
+  GOptionContext *ctx = g_option_context_new("map");
```
alltilla, line 36: What is `ctx` used for?

jszigetvari:
@furiel Thanks for the heads up. I will wait for your other MR with the filter implementation that requires updating the `eval` signature.

modules/basicfuncs/tf-map.c
furiel: I did not want to deal with shell escaping by hand, that's why I used argparser from glib. That needs a context to work. I just tried, and `g_option_context_parse` with `NULL` causes crash.

kira-syslogng:
Build FAILURE

MrAnno:

kira-syslogng:
Build FAILURE

modules/basicfuncs/tf-map.c
furiel: Hmm I confused it with something. You are right. Context is not needed here.

kira-syslogng:
Build SUCCESS

doc/doc-3301.md
```diff
@@ -0,0 +1,20 @@
...
+  source { example-msg-generator(template("0,1,2") num(1)); };
+  destination { file("/dev/stdout" template("$(map $(+ 1 $_) $MSG)\n")); };
+};
+```
+
+# Description
+
+`map` template function has two mandatory parameters. The first parameter is a template, that will be applied to each list elements. The second parameter is a string or template, that represents the list.
+
+You can use `$_` template to refer to the list element inside the template function.
```
Kokan, line 20: [optional]
As @MrAnno suggested nested `map` also works, and would worth to mention. Also `$(map $(map ... ) ..)` works, without lambda probably not useful, but a statement regarding how the `$_` binds (which map it refers to) would be nice to have.

MrAnno:
All review notes seem to be optional. @furiel please let us know whether you want to change anything. :)

furiel:
Yes, thank you for your comments! I will update the documentation and add a the embedded map as unit testcase. I will update in a few days.

doc/doc-3301.md
furiel: Fixed thanks.

kira-syslogng:
Build SUCCESS

PR #3302, opened 2020-05-22
github-actions: add devshell based CI job

alltilla:

### DIFF
--- a/.github/workflows/devshell.yml
+++ b/.github/workflows/devshell.yml
@@ -1,0 +1,174 @@
+name: CI @ devshell
+
+on: [pull_request, push]
+
+jobs:
+  general:
+    runs-on: ubuntu-18.04
+    container:
+      image: balabit/syslog-ng-devshell:latest
+      env:
+        PYTHONUSERBASE: /github/home/python_packages
+        PATH: /github/home/python_packages/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+        CONFIGURE_FLAGS:
+          --prefix=/github/home/install/syslog-ng
+          --enable-all-modules
+          --with-python=${{ matrix.python }}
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:

kira-syslogng:
Build SUCCESS

.github/workflows/devshell.yml
```diff
@@ -0,0 +1,174 @@
...
+
+    strategy:
+      matrix:
+        python: [2, 3]
+        build-tool: [autotools, cmake]
+      fail-fast: false
+
+    steps:
+      - name: Checkout syslog-ng source
+        uses: actions/checkout@v1
```
furiel, line 30: Please use `actions/checkout@v2` so that the reexecute button could work.

alltilla: Thanks!

alltilla: Wow it break the Light tests...
```
Built target pytest-self-check
Scanning dependencies of target pytest-linters
An error has occurred: FatalError: git failed. Is it installed, and are you in a Git repository directory?
Check the log at /github/home/.cache/pre-commit/pre-commit.log
```

alltilla: https://github.com/actions/checkout/issues/126

alltilla: This error happens, when the `git` running in the docker is older, than `2.18`.

`bionic` has `git` version `2.17`, focal has `2.25`. Maybe we can merge it with `v1`, and presumably we can change the `devshell` to `focal` soon, and update to `v2 `then.

Another alternative would be to install `2.18` in the `devshell`, but, I think it would be more painful.

What do you think?

kira-syslogng:
Build SUCCESS

PR #3304, opened 2020-05-25
threaded-source: fully support default-priority() and ...-facility()

alltilla:
`LogThreadedSourceDriver` provided the `MsgFormatOption` variable and the `default-priority()` and `default-facility()` options from its grammar, but it did not provide its funcionality.

IMO news entry is not needed.


kira-syslogng:
Build SUCCESS

alltilla:
@MrAnno @gaborznagy Thanks guys :)

PR #3305, opened 2020-05-25
Rdkafka fix simple

lbudai:
temporary fixes: #3080 

I think this could work until we have the final solution for the issue.

Basic concept:
Poller thread (`worker_index == 0`) calls `rd_kafka_poll` publish messages in non-blocking way, and calls `rd_kafka_poll`. Other threads calls publish in a blocking way - and they don't call `poll()` method.

When the non-blocking publish returns an error (librdkafka's internal queue is full...), `worker_insert` returns `LTR_RETRY` - and then syslog-ng waits `poll_timeout` (converted to seconds) instead of `time_reopen`. Default value for `poll_timeout` is 1 second.

### DIFF
--- a/modules/kafka/kafka-dest-worker.c
+++ b/modules/kafka/kafka-dest-worker.c
@@ -157,12 +157,26 @@ kafka_dest_worker_free(LogThreadedDestWorker *s)
   log_threaded_dest_worker_free_method(s);
 }
 
+static gboolean
+_thread_init(LogThreadedDestWorker *s)
+{
+  KafkaDestWorker *self = (KafkaDestWorker *) s;
+  if (_is_poller_thread(self))
+    {
+      KafkaDestDriver *owner = (KafkaDestDriver *) s->owner;
+      s->time_reopen = owner->poll_timeout / 1000;
+    }
+
+  return log_threaded_dest_worker_init_method(s);
+}
+
 LogThreadedDestWorker *
 kafka_dest_worker_new(LogThreadedDestDriver *o, gint worker_index)
 {
   KafkaDestWorker *self = g_new0(KafkaDestWorker, 1);
 
   log_threaded_dest_worker_init_instance(&self->super, o, worker_index);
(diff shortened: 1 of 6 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

modules/kafka/kafka-dest-worker.c
```diff
@@ -150,12 +157,23 @@ kafka_dest_worker_free(LogThreadedDestWorker *s)
...
+}
+
 LogThreadedDestWorker *
 kafka_dest_worker_new(LogThreadedDestDriver *o, gint worker_index)
 {
   KafkaDestWorker *self = g_new0(KafkaDestWorker, 1);
 
   log_threaded_dest_worker_init_instance(&self->super, o, worker_index);
+  if (_is_poller_thread(self))
+    self->super.thread_init = _poller_thread_init;
```
bazsi, line 176: I would just initialize the function pointer to a static value here and branch on the same conditional there. It doesn't have performance implication but makes the code easier to follow.

MrAnno:

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-worker.c
lbudai: It's not about performance. My intention was to express that in the other case we don't have to implement the original method.
I'll change it (and will implement another method for the non-poller case that will just call the original method - and nothing else).

lbudai: finally I just merged the two in one `_thread_init`

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lbudai:
why the wip flag? before merge, doc needed ...

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3310, opened 2020-05-29
Running patches

johnskopis:
These are the patches I have been running with for #3080

### DIFF
--- a/modules/kafka/kafka-dest-driver.c
+++ b/modules/kafka/kafka-dest-driver.c
@@ -28,6 +28,7 @@
 #include "kafka-props.h"
 #include "kafka-dest-worker.h"
 #include "timeutils/misc.h"
+#include "stats/stats-registry.h"
 
 #include <librdkafka/rdkafka.h>
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

lbudai:
IMHO my PR solved the blocking issue (-> https://github.com/syslog-ng/syslog-ng/pull/3305 )

but the use after free (and the delegation of retry mechanism) should be checked.

When librdkafka runs DR cb (delivery report callback - per message), it performs `msg_destroy`, so it has a good chance that the use after free is still unresolved. And I like the idea about delegating the retry mechanism to librdkafka.

gaborznagy:
Hi @johnskopis,

We've had some time to check the use after free issue that's mentioned in this PR.
I would like to share my experiences:

The payload that is passed over to librdkafka is the `message` string, and we don't track any reference to that string object once it's passed over to librdkafka [we stop tracking the payload here](https://github.com/syslog-ng/syslog-ng/blob/55b224296cf18d933400cce7b011f58576ed7c59/modules/kafka/kafka-dest-worker.c#L136).
The LogMessage object is an opaque user_data that is passed to the `rd_kafka_produce` function, so we can use it later in the the delivery callback, but librdkafka doesn't take over ownership over the LogMessage object. As I've checked there is no use after free regarding the LogMessage object.

On the other hand I've found some segfaults though when I was checking the error handling of kafka-c and I've found a double acknowledgement problem:
when in the delivery callback we re-add the LogMessage to syslog-ng's queue we re-enable acknowledgement, even though it's already acknowledged back to the source. Both in flow-controlled and non-flow-controlled cases this could result that we send an ack back to a non-existing source.
Is this something you've experienced as well? You've mentioned crashes of syslog-ng when the librdkafka queue fills up

I've submitted a patch about this. PR: https://github.com/syslog-ng/syslog-ng/pull/3583

In the near future, we would like to redesign this part about the message acknowledgement, as we currently ack back earlier than the delivery callback can report a successful message sending, which causes that `flow-control` is not working as intended.

We don't know how we should do it yet, and leveraging on librdkafka's retry mechanism is a good way of handling delivery issues, but if at some point librdkafka reports back a delivery failure, we have to put back the LogMessage into syslog-ng's queue, otherwise it would be lost.

bazsi:
Could we discuss this in depth, potentially IRL?

the kafka delivery model does not easily match with the syslog-ng model,
and I vaguely remember certain tradeoffs I had to make when implementing
the flow control model.

PR #3311, opened 2020-05-29
Deb/normalize packaging fix gradle criterion

lbudai:
note: I wouldn't add a news file entry for this PR.

### DIFF
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -58,6 +58,8 @@ endif
 
 ifneq ($(filter nojava,$(DEB_BUILD_PROFILES)),)
 DH_OPTIONS += -Nsyslog-ng-mod-java -Nsyslog-ng-mod-kafka -Nsyslog-ng-mod-java-common-lib -Nsyslog-ng-mod-elastic -Nsyslog-ng-mod-java-http -Nsyslog-ng-mod-hdfs
+else
+export GRADLE_USER_HOME = /tmp/gradle
 endif
 
 ifneq ($(filter sng-python2,$(DEB_BUILD_PROFILES)),)
--- a/dbld/images/ubuntu-bionic.dockerfile
+++ b/dbld/images/ubuntu-bionic.dockerfile
@@ -21,7 +21,6 @@ RUN /dbld/builddeps install_debian_build_deps
 RUN /dbld/builddeps install_pip_packages
 
 RUN /dbld/builddeps install_criterion
-RUN /dbld/builddeps install_gradle
 RUN /dbld/builddeps install_gosu amd64
 
 VOLUME /source
--- a/dbld/images/ubuntu-focal.dockerfile
+++ b/dbld/images/ubuntu-focal.dockerfile
@@ -20,7 +20,6 @@ RUN /dbld/builddeps install_debian_build_deps
 RUN /dbld/builddeps install_pip_packages
 
 RUN /dbld/builddeps install_criterion
-RUN /dbld/builddeps install_gradle
 RUN /dbld/builddeps install_gosu amd64
 
 VOLUME /source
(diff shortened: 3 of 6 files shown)


kira-syslogng:
Build SUCCESS

lbudai:
build packages please.

kira-syslogng:
Package Build FAILURE

packaging/debian/rules
```diff
@@ -40,6 +40,7 @@ endif
 UMAJOR		   = $(shell dpkg-parsechangelog | sed -n '/^Version:/s/^Version: //p' | cut -d. -f1,2)
 
 export DH_OPTIONS += -O-Bdebian/build-tree --dbg-package=syslog-ng-dbg
+export GRADLE_USER_HOME = /tmp/gradle
```
bazsi, line 43: Apart from this single change I think the rest is ok. We discussed this last time, that it should be set external to the rules file, but that is not an option in our OBS builds. I remember you said GRADLE_USER_HOME is not properly set by gradle in that environment.

It does have a default though (quoting https://docs.gradle.org/current/userguide/build_environment.html):      

```
GRADLE_USER_HOME

    Specifies the Gradle user home directory (which defaults to $USER_HOME/.gradle if not set).
```

Or we don't have a HOME directory in OBS? The reason we want this to be set in dbld is to persist downloaded .jar files accross builds, but I don't think that is important for OBS. Can you show me a failed build when this is unset? I am trying to reproduce this case locally, so I might have an answer before you post a response but I'll let you know. :)

bazsi: I think I could reproduce this, the issue being is that gradle bails out with a fatal error if the GRADLE_USER_HOME directory is not writable by the user it is running as. This is the case with dbld if GRADLE_USER_HOME is not specified and might be the case with OBS as well.

I am trying this fix now, which I think is acceptable into our debian packaging:

```
diff --git a/packaging/debian/rules b/packaging/debian/rules
index 4fb631835..496b73eb3 100755
--- a/packaging/debian/rules
+++ b/packaging/debian/rules
@@ -58,6 +58,8 @@ endif
 
 ifneq ($(filter nojava,$(DEB_BUILD_PROFILES)),)
 DH_OPTIONS += -Nsyslog-ng-mod-java -Nsyslog-ng-mod-kafka -Nsyslog-ng-mod-java-common-lib -Nsyslog-ng-mod-elastic -Nsyslog-ng-mod-java-http -Nsyslog-ng-mod-hdfs
+else
+export GRADLE_USER_HOME ?= debian/gradle-home
 endif
 
 ifneq ($(filter sng-python2,$(DEB_BUILD_PROFILES)),)

```

With this, we only specify GRADLE_USER_HOME is java is enabled, only specifies GRADLE_USER_HOME if it is not set from the outside and puts it inside the debian build directory.

I'll confirm this builds without the external GRADLE_USER_HOME setting in a minute.

bazsi: It built successfully and the gradle files were installed into debian/gradle-home, which is actually nice, as we won't add cached artifacts into the user's own directory.

So with this change I think this change can go in.

lbudai: @bazsi : thanks. I applied the patch and squashed into the original commit

kira-syslogng:
Build SUCCESS

lbudai:
build packages please.

kira-syslogng:
Package Build FAILURE

lbudai:
build packages please.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Package Build FAILURE

lbudai:
build packages please.

kira-syslogng:
Package Build FAILURE

lbudai:
> Thanks.
> 
> Only the Fedora 32 job is failing now because of the non-existent image.
> #3315 will add the actual fedora-32 support.

yes, this is why there is a dependency between these two prs. after that merged, it should be already pass... let's try it.

lbudai:
build packages please.

kira-syslogng:
Package Build SUCCESS

Packages are available at https://ci.syslog-ng.com/jenkins/job/kira-dbld-pull-requests/686/artifact/

MrAnno:
Merging this with @bazsi's approval.

PR #3313, opened 2020-06-03
dbld: remove ubuntu-eoan support

bazsi:
eoan is not an LTS and we have focal instead.


kira-syslogng:
Build FAILURE

lbudai:

kira-syslogng:
Build FAILURE

bazsi:
@lbudai this is the same problem we discussed in email, the libmongoc version we explicitly pull in in our dbld images is missing from OBS.

lbudai:
> @lbudai this is the same problem we discussed in email, the libmongoc version we explicitly pull in in our dbld images is missing from OBS.

yes... unfortunately Packages.gz does not contain libmongoc-dev package... 
There are other issues regarding to OBS (missing Release.key file for Ubuntu Focal).

lbudai:
@kira-syslogng : retest this please.

bazsi:
What if we just fixed the libmongoc related files from dockerfile abd with that we would avoid the obs dependency?

lbudai:
> What if we just fixed the libmongoc related files from dockerfile abd with that we would avoid the obs dependency?

I think it's not a big issue - OBS upgrades keys from time to time then we have to wait for upgrade the repository. Now it should work.

kira-syslogng:
Build SUCCESS

PR #3316, opened 2020-06-04
mainloop: save and exec external sigaction handler for SIGINT

alltilla:
The trigger for this PR is from Oracle's instant client, which overwrites our sigaction handler for `SIGINT`.

No news entry or doc is needed IMO.

### DIFF
--- a/lib/signal-handler.c
+++ b/lib/signal-handler.c
@@ -110,6 +110,7 @@ _need_to_save_external_sigaction_handler(gint signum)
   switch (signum)
     {
     case SIGCHLD:
+    case SIGINT:
       return TRUE;
     default:
       return FALSE;
--- a/lib/mainloop.c
+++ b/lib/mainloop.c
@@ -489,6 +489,15 @@ sig_term_handler(gpointer user_data)
 }
 
 static void
+sig_int_handler(gpointer user_data)
+{
+  MainLoop *self = (MainLoop *) user_data;
+
+  main_loop_exit_initiate(self);
+  signal_handler_exec_external_handler(SIGINT);
+}
+
+static void
 sig_child_handler(gpointer user_data)
 {
   pid_t pid;
@@ -541,7 +550,7 @@ setup_signals(MainLoop *self)
   _register_signal_handler(&self->sighup_poll, SIGHUP, sig_hup_handler, self);
   _register_signal_handler(&self->sigchild_poll, SIGCHLD, sig_child_handler, self);
   _register_signal_handler(&self->sigterm_poll, SIGTERM, sig_term_handler, self);
-  _register_signal_handler(&self->sigint_poll, SIGINT, sig_term_handler, self);
+  _register_signal_handler(&self->sigint_poll, SIGINT, sig_int_handler, self);
   _register_signal_handler(&self->sigusr1_poll, SIGUSR1, sig_usr1_handler, self);
 }
 


kira-syslogng:
Build FAILURE

lbudai:
@alltilla : 32 won't be enough for python...  Highest signal number is `python::signal.NSIG`.
which is actually defined here (Modules/signalmodule.c at cpython src):
```
#ifndef NSIG
# if defined(_NSIG)
#  define NSIG _NSIG            /* For BSD/SysV */
# elif defined(_SIGMAX)
#  define NSIG (_SIGMAX + 1)    /* For QNX */
# elif defined(SIGMAX)
#  define NSIG (SIGMAX + 1)     /* For djgpp */
# else
#  define NSIG 64               /* Use a reasonable default value */
# endif
#endif
```
NSIG is part of signal.h on linux, and the value is 65.
So if you want to use an array for signals then the correct size is 65 - and unit tests won't crash.

alltilla:
@lbudai Not a bad idea, but I believe I found a more convenient fix:
All in all we only need to care about the signals we set sigaction to, and we can be sure about the max signum value of them.
I put an early return to our `sigaction()` function, to simply call the original `sigaction()` if we does not care about that signum:
```
/* This should be as defined in the <signal.h> */
int
sigaction(int signum, const struct sigaction *act, struct sigaction *oldact)
{
  if (!_need_to_save_external_sigaction_handler(signum))
    return _call_original_sigaction(signum, act, oldact);
```
We do not operate on the arrays before this point, so it is safe now.

I added 2 `g_assert()`s for the signum being smaller than the array size, one for the `_need_to_save_external_sigaction_handler()` function and one for the public function `signal_handler_exec_external_handler()`.
I also made the array size 128 (which is more than enough), for future-proofness, but this can work with the original 32 size, too.

One question: Should we protect our other signals as wall, in advance (`SIGHUP`, `SIGTERM`, `SIGUSR1`)?

lbudai:
@alltilla : i did not suggest a solution, just wanted to share what is the root cause behind test failures - when you remove the wip, I'll review it in details ;)

alltilla:
Thanks!

wip removed, CI should pass also, but we will see :)

kira-syslogng:
Build SUCCESS

lib/signal-handler.c
```diff
@@ -23,74 +23,85 @@
...
 
 /* This should be as defined in the <signal.h> */
 int
 sigaction(int signum, const struct sigaction *act, struct sigaction *oldact)
 {
+  if (!_need_to_save_external_sigaction_handler(signum))
+    return _original_sigaction(signum, act, oldact);
 
-  if (signum == SIGCHLD)
+  /* Internal sigactions are always the first one to arrive to this function. */
```
furiel, line 93: Generally, I am fine with this patchset if it works, but I think we could be safer. Marking this comment/request as optional:

This patchset expects that `setup_signals` is the first one that registers the signal hanlders. This is not true, if a shared library modifies signal handlers in constructor functions.

According to the man page of `sigaction`, the old signal handler needs to be stored in `oldact`. For the first two calls of sigaction, this will happen through `_original_sigaction` for the same signal, but not for later calls. If we do not want to deal with this right now, it might worth adding a NULL assert for oldact.

This latter also breaks the scenario if two modules want to edit the same signal handler (additionally to `setup_signals`). The subsequent signal handlers just overwrite each other, and the original action is not propagated back by the output parameter. So the caller does not know about the original action, and cannot save and call it.

I was thinking the best would be if we did not need to override `sigaction` at all. If the signals weren't registered through ivykis, then we could just store `oldact` and call it in our signal handler. But it is registered by `iv_signal_register` currently.

Reading the ivykis manual, i found:

```
     It  is  permitted  to  register  multiple  struct  iv_signal
     objects for the same signal number.  If all such objects are
     registered with  IV_SIGNAL_FLAG_EXCLUSIVE in  ->flags unset,
     then all  objects will have their  callback functions called
     upon arrival of the signal, in no guaranteed order.
```

I am just guessing, but maybe if if we would not set this flag in `_register_signal_handler`, at least for the underlying signals, then maybe this sigaction implementation would not be necessary.

Another alternative: a bit hacky, but you can still imitate the sigaction behavior with `iv_signal_register`. You can get `oldact` with two subsequent sigaction calls: swapping out the handler with null, then back with the original value. Once you have oldact, you can properly implement the chaining without hacking `sigaction`.

MrAnno: We've discussed that in the case of the Oracle instant client, overriding `sigaction` is necessary due to a possible bug in the client library.
We should copy `const struct sigaction *act`, because it is usually stored on the stack, and should return the appropriate `oldact` in all cases.

alltilla: @furiel Thanks for the comments.
@MrAnno Thanks, done all of that.

kira-syslogng:
Build SUCCESS

PR #3320, opened 2020-06-12
Stats freq overflow

Kokan:
The current `stats-freq` option could go as high as `2^31-1` (which of course kinda unrealistic), but the opposite happens, the stats reports start to spawn right away in an infinite loop manner.

When timer is adjusted (`timespec_add_msec(&timer->expires, freq * 1000);`) the `freq * 1000` is used, where the `freq` has the type `gint`, and the function - correctly - waits a `glong`, but the multiplication happens before the parameter assignment and in the domain of `gint` that yields `gint` overflow.


kira-syslogng:
Build SUCCESS

PR #3323, opened 2020-06-16
logsource: add use-syslogng-pid() option

alltilla:
This sets the `${PID}` macro to the `syslog-ng` process' PID.
If `${PID}` was set before, this option overrides it.

Defaults to `no`.

### DIFF
--- a/lib/logsource.c
+++ b/lib/logsource.c
@@ -34,6 +34,7 @@
 #include "scratch-buffers.h"
 
 #include <string.h>
+#include <unistd.h>
 
 gboolean accurate_nanosleep = FALSE;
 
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

lib/logsource.c
```diff
@@ -574,6 +575,25 @@ log_source_override_program(LogSource *self, LogMessage *msg)
...
+#define MAX_PID_CHAR_COUNT 20 /* max PID on 64 bit systems is 2^64 - 1, which is 19 characters, +1 for terminating 0 */
+
+  static gchar pid_string[MAX_PID_CHAR_COUNT];
+
+  if (pid_string[0] == '\0')
+    {
+#ifdef _WIN32
+      g_snprintf(pid_string, MAX_PID_CHAR_COUNT, "%lu", GetCurrentProcessId());
+#else
+      g_snprintf(pid_string, MAX_PID_CHAR_COUNT, "%d", getpid());
```
MrAnno, line 590: 64bit pids can be problematic with `%d` (the type is `pid_t`), but I've never seen such a thing IRL.

lbudai: AFAIK `pid_t` is still 32 bits.

MrAnno: Yeah, it should be.
My pid_max is `4194304` with Linux 5.6.15.

PR #3330, opened 2020-06-18
light: fix MessageReader-based tests with Python 2

MrAnno:
In certain Python 2 versions, the `read()` method of buffered file objects can not be used to follow a file: after reaching EOF, subsequent calls will not read new data.

`fsync` calls have also been added to sync data to disk before reading with `READ_ALL_AVAILABLE_MESSAGES`.

Needed for #3277.


kira-syslogng:
Build SUCCESS

PR #3333, opened 2020-06-19
github-actions: cleanup ENV variables

alltilla:
The motivation behind this change:
 1. The `env:` option in the YAML cannot process embedded variables, so we had things like: `PYTHONUSERBASE: /github/home/python_packages`, which should be `${HOME}/python_packages`.
 2. The `env:` option does not handle commands in variables. We need to set the python version, in configure time, with only the python major version, like 2 or 3, so we had 2 and 3 instead of 'python2' and 'python3' in the strategy matrix. The values of the strategy matrix are shown in the workflow run, and we got jobs like: (2, autotools), where the 2 is not comprehensible for the viewer.

The solution for both problems is a separate step, which sets the environment variables. But we have a problem with that too:

Environment variables set in one step are not available in other steps, even if they are set with `export`.

"Fortunately" GitHub Actions give a solution for this:
https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

With that we can make it work, as expected.

### DIFF
--- a/.github/workflows/devshell.yml
+++ b/.github/workflows/devshell.yml
@@ -7,25 +7,35 @@ jobs:
     runs-on: ubuntu-18.04
     container:
       image: balabit/syslog-ng-devshell:latest
-      env:
-        PYTHONUSERBASE: /github/home/python_packages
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

PR #3335, opened 2020-06-23
dbld: install virtualenv from pip

alltilla:
It was implicitly installed by some other pip module, and also it was installed by apt, too. Unfortunately they tangle up, and try to use each other's libs, it is cleaner to only install it from pip explicitly.

This PR fixes the GitHub Actions run. See: https://github.com/alltilla/syslog-ng/actions/runs/144955299
After this PR is merged, and a new devshell image is generated to Docker Hub, the upcoming runs will be green.


kira-syslogng:
Build SUCCESS

PR #3338, opened 2020-06-25
Fix signal handling when an external library/plugin sets SIG_IGN

MrAnno:
Previously, setting `SIG_IGN` in a plugin/library (for example, in a Python module) resulted in a crash.

`SIG_DFL` and `SIG_IGN` are special values for `sa_handler`, they should not be treated as valid memory addresses.


kira-syslogng:
Build SUCCESS

PR #3342, opened 2020-06-30
disk queue: fix false positive corruption or possible crash during load

furiel:
Due to some cast/signedness problems, some disk queues could not be loaded, or syslog-ng could crash during startup.
In neither cases the disk queues are corrupt, and they should be loaded correctly after this patch.

<details>

```
@version: 3.27

options {
  log-msg-size(1000000);
};

log {
  source { example-msg-generator(freq(0)); };
  destination {
    network(127.0.0.1 port(5555)
      disk-buffer(
        disk-buf-size(214748)
        dir('/tmp/diskq-test')
       )
      log_fifo_size(9000000)
    );
  };
};
```

</details>

- In the example above, the overflow area will be read as unsigned 32 bit int, which will be casted to a signed 32bit int. Which will be passed to glib as `off_t` (64 bit unsigned) at

https://github.com/syslog-ng/syslog-ng/blob/4e19fa746136833934106e779f64aa9432c611a6/modules/diskq/qdisk.c#L469
A 32bit negative number turns into a huge 64 bit positive number, hence the crash. With the example above, these are the numbers:
```
-1936967296 2358000000 18446744071772584320
```

- There was another problem with the deserialization of the messages. As a validation step, we checked if an `nventry` is indeed inside the `nvtable`. There was a cast from pointer to 32 bit int, which could lead to false positive results.


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -455,6 +455,20 @@ qdisk_pop_head(QDisk *self, GString *record)
   return FALSE;
 }
 
+static ssize_t
+_pread_all(int fd, char *buf,  size_t count, off_t offset)
+{
+  int retval = pread(fd, buf, count, offset);
```
alltilla: We should use `ssize_t` here, too.

modules/diskq/qdisk.c
```diff
@@ -494,7 +493,7 @@ _load_queue(QDisk *self, GQueue *q, gint64 q_ofs, guint32 q_len, guint32 q_count
           return FALSE;
         }
       sa = serialize_string_archive_new(serialized);
-      for (i = 0; i < q_count; i++)
+      for (gsize i = 0; i < q_count; i++)
```
alltilla, line 465: Why not `guint32`?

modules/diskq/qdisk.c
```diff
@@ -509,6 +508,8 @@ _load_queue(QDisk *self, GQueue *q, gint64 q_ofs, guint32 q_len, guint32 q_count
             {
               msg_error("Error reading message from disk-queue file (maybe corrupted file) some messages will be lost",
                         evt_tag_str("filename", self->filename),
+                        evt_tag_uint("num_of_messages", q_count),
+                        evt_tag_long("invalid_index", i),
```
alltilla, line 481: `evt_tag_long` might not be the best for `gsize`, which is `unsigned long`.

modules/diskq/diskq-serializer.c
```diff
@@ -45,6 +54,96 @@ _pread_all(int fd, char *buf,  size_t count, off_t offset)
...
+_cache_free(DiskQSerializerCache *self)
+{
+  g_string_free(self->buffer, TRUE);
+  g_free(self);
+}
+
+static gboolean
+_cache_fetch(DiskQSerializerCache *self, gchar *buf, gsize buflen)
+{
+  if (self->pos + buflen > self->buffer->len)
```
alltilla, line 78: Can you move this to a separate function, like `if (!_enough_data_in_cache(self, buflen))`?

modules/diskq/diskq-serializer.c
```diff
@@ -0,0 +1,78 @@
...
+{
+  int retval = pread(fd, buf, count, offset);
+
+  if (retval <= 0)
+    return retval;
+
+  if (retval == count)
+    return count;
+
+  return retval + _pread_all(fd, buf+retval, count-retval, offset+retval);
```
alltilla, line 55: What if the `retval` is `10` and the next `pread` call returns `-1`? We return `9` in that case, which is not good.

modules/diskq/diskq-serializer.c
```diff
@@ -45,6 +54,96 @@ _pread_all(int fd, char *buf,  size_t count, off_t offset)
...
+{
+  g_string_set_size(self->buffer, self->buffer->len + buflen);
+}
+
+static gboolean
+_push_data_to_cache(MsgSerializer *self, gsize buflen)
+{
+  gsize chunk_size = MAX(buflen, CHUNK_SIZE);
+
+  if ((self->cache->pos % 16) == 0)
```
alltilla, line 132: Where does the `16` come from?

modules/diskq/diskq-serializer.c
alltilla: I think a simple while loop would be better in this case.

modules/diskq/diskq-serializer.c
```diff
@@ -0,0 +1,195 @@
...
+msg_serializer_free(SerializeArchive *s)
+{
+  MsgSerializer *self = (MsgSerializer *)s;
+  _cache_free(self->cache);
+}
+
+SerializeArchive *
+msg_serializer_new_full(int fd, gsize offset, gsize chunk_size)
+{
+  MsgSerializer *self = g_slice_new0(MsgSerializer);
```
lbudai, line 179: why are you using `g_slice`?

modules/diskq/qdisk.c
furiel: I wanted to be "future" proof by having a sufficiently large storage. But you ar  rignt, technically the type should be the same of q_count. I would like to leave this as it is.

modules/diskq/qdisk.c
furiel: The name of the function is misleading: it is actually formatted as `"%lld"`

```
EVTTAG *
evt_tag_long(const char *tag, long long value)
{
  char buf[32]; /* a 64 bit int fits into 20 characters */

  snprintf(buf, sizeof(buf), "%lld", value);
  return evt_tag_str(tag, buf);
}
```

alltilla: Shouldn't we print `unsigned long` with `%llu` or `%lu`?

modules/diskq/qdisk.c
alltilla: Sure, thanks for the explanation. :)

modules/diskq/diskq-serializer.c
furiel: I split the patchset in two. This will be addressed in the next PR.

modules/diskq/diskq-serializer.c
furiel: Fixed thanks.

modules/diskq/diskq-serializer.c
furiel: This was accidental. Originally, I was fighting with some random corruption error problem, and I was suspecting it is related to alignment. This line supposed to avoid any alignment problem by only recycling the first part of the buffer, if the relevant data starts on a 16-divisible location. Moving the data in this case is safe. Later it turned out the failure came from the nventry check, so this is not necessary.

modules/diskq/qdisk.c
```diff
@@ -455,6 +455,20 @@ qdisk_pop_head(QDisk *self, GString *record)
...
+{
+  int retval = pread(fd, buf, count, offset);
+
+  if (retval <= 0)
+    return retval;
+
+  if (retval == count)
+    return count;
+
+  return retval + _pread_all(fd, buf+retval, count-retval, offset+retval);
```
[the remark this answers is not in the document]
furiel: Fixed thanks.

modules/diskq/diskq-serializer.c
furiel: This is unfortunate thing. The free function uses this logic too.

```
void
serialize_archive_free(SerializeArchive *self)
{
  g_clear_error(&self->error);
  g_slice_free1(self->len, self);
}
```

I do not know any reason why would we want to use slices. On the other hand, if I want to change it, I need to change all of the original serializers, which is not too much code, but I do not want to open the scope with it right now.

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -455,6 +455,20 @@ qdisk_pop_head(QDisk *self, GString *record)
...
 
+static ssize_t
+_pread_all(int fd, char *buf,  size_t count, off_t offset)
+{
+  int retval = pread(fd, buf, count, offset);
+
+  if (retval <= 0)
+    return retval;
+
+  if (retval == count)
```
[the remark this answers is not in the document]
furiel: Good point thanks. In the end, the code has changed a little, but I followed this up too: the newly introduced already_read variable and offset from the signature both have size_t type.

modules/diskq/qdisk.c
furiel: Fixed thanks.

modules/diskq/qdisk.c
furiel: I changed my mind, and will use `guint32` for `i`, which fixes the original problem.

Also, I was confused with `evt_tag_long` too, that's why I introduced `evt_tag_uint`. On the other hand, every value here can be represented with `evt_tag_long`, so I dropped that patch too.

furiel:
@lbudai @alltilla thank you for your review. I updated the PR according to your comments. I changed the PR to draft, so that before the merge, we could run internal tests for it.

kira-syslogng:
Build SUCCESS

PR #3343, opened 2020-07-01
UnitTestsNose: Replace nosetest with pytest

mitzkia:
Tested with `dbld/rules` base image generation -> dbld/rules make-check is passed (tested only with automake) 
Also checked when a particular unit test is failing, assertions will be dumped into `./dbld/build/test-suite.log`


kira-syslogng:
Build SUCCESS

modules/python/pylib/Makefile.am
```diff
@@ -66,13 +66,8 @@ clean-pylib:
 
 python-checks: python-unit python-pep8 python-pylint
 
-if NOSETESTS
 python-unit:
-	@NOSETESTS@ $(PYLIB_SRCDIR)/syslogng
-else
-python-unit:
-	$(error "missing nosetests")
```
Kokan, line 74: Is it possible to keep an error message about a missing dependency here ?

modules/python/pylib/Makefile.am
```diff
@@ -66,13 +66,8 @@ clean-pylib:
 
 python-checks: python-unit python-pep8 python-pylint
 
-if NOSETESTS
 python-unit:
```
bazsi, line 69: can you please merge python-* targets with pytest-* targets? I don't see we need both, and since the prefix "python-" seems more general, I think pytest-* should just be shoved inside the python targets (and probably be promoted to the root Makefile instead of being under modules/python/pylib).

The best would be if we had very specific targets under modules/python/pylib/ and tests/python_functional/ which would be depended on by a generic target in the root Makefile.

modules/python/pylib/Makefile.am
mitzkia: Of course it was in my mind. 
I have just updated the code.

modules/python/pylib/Makefile.am
mitzkia: Thanks for the comment, I will think over it.

kira-syslogng:
Build SUCCESS

modules/python/pylib/Makefile.am
mitzkia: If I understand you correctly, something like this would be the result:
root/Makefile:
`python-checks` (as a new target): which runs all python-specified targets from `modules/python/pylib/` and from  `tests/python_functional/`

modules/python/pylib/Makefile:
`python-module-checks` (as a new target name for this specified directory): which runs current targets from `modules/python/pylib/`, maybe renaming is needed to not conflicting with targets under `tests/python_functional/ `

tests/python_functional/Makefile:
`python-functional-checks` (as a new target name for this specified directory): which runs current targets from `tests/python_functional/`, maybe with renaming to not conflict with `modules/python/pylib/` targets

If this would be a goal, I am ok with it in this PR (maybe good target names will be hard to choose).

And what about "old" func-test target from: tests/functional/Makefile.am? Should I include it also?

kira-syslogng:
Build SUCCESS

mitzkia:
There were some other discussions (ideas) with this PR:
1, merge python related make targets, I would wait for a @bazsi 's answer here after that we have discussed this topic IRL. Do we still want to merge python make targets (please check my last comment in this topic https://github.com/syslog-ng/syslog-ng/pull/3343#discussion_r449537952)?
2, Add possibility to use external pytest arguments for pytest related make targets like (pytest-check). I have created this change locally, but would add it in a separate PR.

kira-syslogng:
Build SUCCESS

alltilla:
Can you add these make targets to the CI? Thanks!

mitzkia:
@alltilla  Thanks for the notice. 
* From automake side `python-unit` target is it is already running under `make-check` with the following reference: `modules/python/pylib/test_pylib.sh`
* From Cmake side I will check how `python-unit` target is executed.
After this I would say this PR is finished from my side. I will create a new PR for merging python related unit test targets (pylib + Light unit tests)

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

packaging/debian/control
```diff
@@ -28,7 +28,7 @@ Build-Depends: debhelper (>= 10~),
                librabbitmq-dev,
                python <sng-python2>, python-dev <sng-python2>, python-setuptools <sng-python2>,
                python3, python3-dev, dh-python,
-               python3-nose, python3-ply, python3-setuptools,
+               python3-ply, python3-setuptools,
```
lbudai, line 31: It's not a replacement, it's a removal.
Shouldn't we _replace_ `python3-nose` with `python3-pytest`?

mitzkia: First of all if we really have to we can add it of course. 
But in my opinion it depends on deb builder environment (machine):
- if the deb build process requires to run `make-check` with full feature set (with all tests) yes, than this package (`python3-pytest`) will be needed (of course if the pip packages does not work -> for this there is a `requirements.txt` in root dir where pytest is also listed). 
- if the deb build process requires to run `make-check` but without full feature set (I have added a check before the pytest tests runs, if the pytest binary exists or not, the tests are run only if the pytest binary exist), than the `python3-pytest` package will not be needed.

mitzkia: We should also consider that apt packages are way back than pip packages in package versions. And in daily basis (local dev or in dbld) we are using newest pytest packages (from pip). There can be feature differents in apt vs pip packages which could hurt later.

mitzkia:
I have made to fail 1 python related unit tests, just to verify CI runs them during autotools and cmake checks (under `make-check`).

kira-syslogng:
Build FAILURE

mitzkia:

mitzkia:
@kira-syslogng build packages please

kira-syslogng:
Package Build FAILURE

kira-syslogng:
Build SUCCESS

mitzkia:
@kira-syslogng build packages please

kira-syslogng:
Package Build FAILURE

modules/python/pylib/Makefile.am
szemere: Since there was answer to these questions, and we discussed most of them IRL, I mark this thread resolved.

mitzkia:
@kira-syslogng build packages please

kira-syslogng:
Package Build FAILURE

mitzkia:
@kira-syslogng build packages please

kira-syslogng:
Package Build SUCCESS

Packages are available at https://ci.syslog-ng.com/jenkins/job/kira-dbld-pull-requests/709/artifact/

kira-syslogng:
Build SUCCESS

PR #3346, opened 2020-07-02
panos-parser() source as SCL

bazsi:
This branch continues the great work in originally in #3234 with a couple of changes/improvements:

* support for custom logs are added (e.g. the CONFIG changes with two extra columns: before_change_detail and after_change_detail columns)
* the panos-parser() now handles non paloalto logs by dropping those that don't appear to be valid. (e.g. not enough CSV columns)
* performance improvement by using fixed string matching instead of regexps
* the naming of the fields is matched against the Paloalto official naming (which is a change compared to the original submission)
* support for fields added in later versions of Paloalto were added.
* the application adapter now classifies logs a bit better.

@spricer sorry for taking so long to go through your patches and I hope the changes above are ok with you. With manual testing this seems to work ok.

### DIFF
--- a/scl/paloalto/panos.conf
+++ b/scl/paloalto/panos.conf
@@ -51,13 +51,29 @@ block parser panos-parser (prefix(".panos.") ...) {
       };
     } 
     elif (match('CONFIG', value('`prefix`type'))) {
-      parser {
-        csv-parser(
-          columns("future_use1","receive_time","serial_number","type","log_subtype","version","generated_time","host_name","vsys","command","admin","client","result","configuration_path","sequence_number","action_flags","devicegroup_level1","devicegroup_level2","devicegroup_level3","devicegroup_level4","vsys_name","dvc_name")
-          # Note if using custom logs insert "before_change_detail","after_change_detail" after "action_flags"
-          prefix("`prefix`")
-          delimiters(',')
-        );
+      if {
+          # non-custom format first
+          parser {
+            csv-parser(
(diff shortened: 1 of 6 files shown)


kira-syslogng:
Build SUCCESS

scl/paloalto/panos.conf
```diff
@@ -0,0 +1,273 @@
...
+      # fix the message in case we forward it to a syslog consumer
+      set("${LEGACY_MSGHDR}${MESSAGE}" value("MESSAGE"));
+      set("paloalto_panos" value("PROGRAM"));
+      unset(value("`prefix`tmp"));
+    };
+
+  };
+};
+application panos[syslog] {
+    filter { match("1," value("PROGRAM") type(string) flags(prefix)); };
```
furiel, line 271: According to https://docs.paloaltonetworks.com/pan-os/8-1/pan-os-admin/monitoring/use-syslog-for-monitoring/syslog-field-descriptions/config-log-fields,
the first field is
```
Serial Number (serial) | Serial number of the device that generated the log.
```
and this is where the `1,` comes from? Originally we checked for the paloalto string in the message. How are these serial numbers are assigned to the devices? Are we safe to assume all devices will have `1` as serial number? Would it be better check if `PROGRAM` contains a `,`: like `.*,` instead? That would be more permissive, and with `drop-invalid` flags in the csv parser, we would still be safe-ish against false positives.

bazsi:
I think this is a "FUTURE USE" field and not serial number, and I used it
as an observation.

scl/paloalto/panos.conf
furiel: Yes thank you! I need new glasses... It is the `FUTURE_USE`. All sounds great then.

PR #3348, opened 2020-07-03
add example destination from gitbook

furiel:
By adding the example destination from gitbook to the repo, we have better chance to keep gitbook updated.

The port is not a blind copy paste, there are some changes comparted to the gitbook version.
- The module is not under `modules/example_destination`, but under `modules/examples/destinations/example_destinations`.
- It is not a module on it's own, but it is part of a larger `libexamples` module.
- It is built as static, instead of dynamic.
- The `ModuleInfo` is missing as it is not a module on it's own. The plugin list is moved to the libexamples.
- The functionality is changed a bit: example destination will actually write to a file: the output contains the thread id of the worker.

Other changes:
- the `add_module` in cmake has a new parameter: which can tell if a module wants to build as static or dynamic library.
- a simple e2e test added for `ExampleDestination`
- You can assert messages in light from now.

After this PR is merged, I will update the gitbook too.

### DIFF
--- a/modules/examples/destinations/example_destination/example_destination.c
+++ b/modules/examples/destinations/example_destination/example_destination.c
@@ -21,6 +21,7 @@
  */
 
 #include "example_destination.h"
+#include "example_destination_worker.h"
 #include "example_destination-parser.h"
 
 #include "plugin.h"
@@ -33,12 +34,6 @@
 #include "logthrdest/logthrdestdrv.h"
 
 
-typedef struct
-{
-  LogThreadedDestDriver super;
-  gchar *filename;
-} ExampleDestination;
-
 /*
  * Configuration
  */
@@ -46,7 +41,7 @@ typedef struct
 void
 example_destination_dd_set_filename(LogDriver *d, const gchar *filename)
 {
-  ExampleDestination *self = (ExampleDestination *)d;
+  ExampleDestinationDriver *self = (ExampleDestinationDriver *)d;
 
   g_free(self->filename);
   self->filename = g_strdup(filename);
@@ -59,7 +54,7 @@ example_destination_dd_set_filename(LogDriver *d, const gchar *filename)
 static const gchar *
 _format_stats_instance(LogThreadedDestDriver *d)
 {
-  ExampleDestination *self = (ExampleDestination *)d;
+  ExampleDestinationDriver *self = (ExampleDestinationDriver *)d;
   static gchar persist_name[1024];
 
   g_snprintf(persist_name, sizeof(persist_name),
@@ -70,7 +65,7 @@ _format_stats_instance(LogThreadedDestDriver *d)
 static const gchar *
 _format_persist_name(const LogPipe *d)
 {
-  ExampleDestination *self = (ExampleDestination *)d;
+  ExampleDestinationDriver *self = (ExampleDestinationDriver *)d;
   static gchar persist_name[1024];
 
   if (d->persist_name)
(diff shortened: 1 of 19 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
I think it would be better if the example dst modules would do _something_ ... something to make it easier to understand what should go into a worker's `thread_init` method, etc...

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
I updated the code with better method implementation.

Example destination will use the filename parameter to actually write to a file on the disk. Each log line will contain the message, and the thread id of the worker.

Using the thread id is awkward, but I could not come up with a better example. I wanted to show the difference between the construct and the thread init method, by showing something that could not be initialized during construct, only during init.

Originally I wanted to use random generators seeded from the thread id. But it turned out it is difficult: the thread safe random number functions are not available on mac-os, so I needed to add detection, and a stub for those platforms where those functions are missing. You can check the code if you want at: https://github.com/furiel/syslog-ng/tree/example-destination-with-random, but in the end I dropped, because this seems unnecessarily complex for a tutorial.

I was thinking maybe I could have used random generator from openssl. The problem with additional dependencies is it makes challenging to keep the gitbook version in sync. Makefiles are very different due to the location mismatch, so any modification need to be updated and tested manually both places. The less modification needed in makefiles is the better.

Alternatively, I could have done something with the thread local storage, but that is a complex topic on it's own, and I do not have any good use case for it either. 

I could have played with ivykis events too, but we do not really modify the event loop from the worker, so this would be an antipattern.

So I just went with the simple thread id version.

I removed the wip flag.

modules/examples/destinations/example_destination/example_destination.c
```diff
@@ -93,24 +84,23 @@ _dd_init(LogPipe *d)
...
-  ExampleDestinationDriver *self = (ExampleDestinationDriver *)s;
+  /*
+     If you created resources during init,
+     you need to destroy them here.
 
-  msg_verbose("Deinitializing ExampleDestination",
-              evt_tag_str("driver", self->super.super.super.id),
-              evt_tag_str("filename", self->filename),
-              NULL);
+     self->filename is outside of the lifecycle of init-deinit (may be
```
alltilla, line 100: I know this is the common `syslog-ng` way of handling strings like this, but we could simply create a `GString` in `new()`, free it in `free()` and `truncate()` then `set()` it in `init()` or `config-set()`.

modules/examples/destinations/example_destination/example_destination.c
```diff
@@ -0,0 +1,134 @@
...
+{
+  ExampleDestinationDriver *self = (ExampleDestinationDriver *)d;
+
+  if (!log_threaded_dest_driver_init_method(d))
+    return FALSE;
+
+  if (!self->filename)
+    self->filename = g_strdup("/tmp/example-destination-output.txt");
+
+  return log_threaded_dest_driver_start_workers(d);
```
furiel, line 90: Since https://github.com/syslog-ng/syslog-ng/pull/3176 you must not start workers manually. You can just return with TRUE here.

furiel: Fixed.

modules/examples/destinations/example_destination/example_destination.c
furiel: Added a new patch to the top. Thanks.

alltilla: Thanks!

kira-syslogng:
Build SUCCESS

PR #3351, opened 2020-07-06
requirements.txt: use isort==4.3.4

lbudai:
Due to a temporary incompatibility between the latest version of isort and Pylint (details: https://github.com/PyCQA/pylint/issues/3722), we freeze isort to version 4.3.4
Once the problem got fixed, we can remove this line and use the latest version from all library.


kira-syslogng:
Build SUCCESS

PR #3352, opened 2020-07-07
reduce disk queue memory usage during load

furiel:
Originally, syslog-ng read one area into memory, and only then it started to deserialize.

This means syslog-ng needs twice as memory during load (the raw area + decoded messages of the same area), than during save. During save syslog already writes to file in small blocks. In extreme cases, the system might not have enough memory to load a disk queue that was saved on the same system.

This patch reads the disk queue as stream, reducing memory footprint.


kira-syslogng:
Build SUCCESS

PR #3355, opened 2020-07-12
scratch-buffers: fix 'global.scratch_buffers_count.queued' bug

gaborznagy:
Due to some bug, the `global.scratch_buffers_count.queued` could shown an invalid value (e.g. 2^32 value, while the actual usage was just decreased to 0).

This bug only affected the stats_counter value, not the actual memory usage (i.e. memory usage was fine before).

### DIFF
--- a/lib/tests/test_scratch_buffers.c
+++ b/lib/tests/test_scratch_buffers.c
@@ -127,33 +127,59 @@ Test(scratch_buffers, local_usage_metrics_measure_allocs)
 extern StatsCounterItem *stats_scratch_buffers_count;
 extern StatsCounterItem *stats_scratch_buffers_bytes;
 
-Test(scratch_buffers, stats_counters_are_updated)
+Test(scratch_buffers_stats, stats_counters_are_updated)
 {
   GString *str;
 
-  for (gint count = 1; count <= ITERATIONS; count++)
+  gint allocated_counter = 0;
+  for (gint i = 1; i <= ITERATIONS; i++)
     {
       str = scratch_buffers_alloc();
       cr_assert_not_null(str);
+      allocated_counter++;
 
       /* check through accessor functions */
-      cr_assert(scratch_buffers_get_local_usage_count() == count,
+      cr_assert(scratch_buffers_get_local_usage_count() == allocated_counter,
                 "get_local_usage_count() not returning proper value, value=%d, expected=%d",
-                scratch_buffers_get_local_usage_count(), count);
+                scratch_buffers_get_local_usage_count(), allocated_counter);
 
-      cr_assert(scratch_buffers_get_local_allocation_bytes() == count * DEFAULT_ALLOC_SIZE,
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

lib/tests/test_scratch_buffers.c
```diff
@@ -153,7 +155,26 @@ Test(scratch_buffers, stats_counters_are_updated)
       /* check if byte counter is updated */
       scratch_buffers_update_stats();
       cr_assert_eq(stats_counter_get(stats_scratch_buffers_bytes), count * DEFAULT_ALLOC_SIZE);
+
+      /* Check internal state is as expected */
+      cr_assert(scratch_buffers_get_local_allocation_count() == count,
```
Kokan, line 160: [optional] I would use `cr_assert_eq` if possible.

lib/tests/test_scratch_buffers.c
```diff
@@ -153,7 +155,26 @@ Test(scratch_buffers, stats_counters_are_updated)
...
+  cr_assert(scratch_buffers_get_local_usage_count() == 0,
+            "get_local_usage_count() failed to reset to 0, value=%d, expected=%d",
+            scratch_buffers_get_local_usage_count(), 0);
+  cr_assert(scratch_buffers_get_local_allocation_count() == count,
+            "Local allocation count failed to reset to 0, value=%d, expected=%d",
+            scratch_buffers_get_local_allocation_count(), count);
+
+  scratch_buffers_allocator_deinit();
+  count = 0;
+  cr_assert(stats_counter_get(stats_scratch_buffers_count) == count,
```
Kokan, line 175: You could just simply assert against `== 0` and not reset `count` to 0. While both are correct, the later (setting count to 0 and assert if equals to count) makes harder to understand what is expected here.

Upon reading the test one could remember that count represents the number of time the `scratch_buffers_alloc` is called, and checking out this assert make not much sense. Of course it can be seen from close by code, but imho when asserting on equality the expected value should be simple, by simple I mean if possible not computed by the test.

lib/tests/test_scratch_buffers.c
```diff
@@ -131,10 +131,12 @@ Test(scratch_buffers, stats_counters_are_updated)
 {
   GString *str;
 
-  for (gint count = 1; count <= ITERATIONS; count++)
+  gint count = 0;
+  while (count < ITERATIONS)
```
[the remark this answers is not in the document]
lbudai: I agree with @Kokan : replacing `for` loop with `while` and widening the scope of the loop variable are little bit weird to me.

lib/tests/test_scratch_buffers.c
gaborznagy: done.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3356, opened 2020-07-14
tlscontext: apply ecdh-curve-list for client, too

alltilla:
https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set1_curves_list.html
> SSL_CTX_set1_groups() sets the supported groups for ctx to glistlen
> groups in the array glist. The array consist of all NIDs of groups in
> preference order. For a TLS client the groups are used directly in the
> supported groups extension. For a TLS server the groups are used to
> determine the set of shared groups.
> ...
> The curve functions are synonyms for the equivalently named group
> functions and are identical in every respect.

### DIFF
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -572,10 +572,6 @@ _load_dh_fallback(TLSContext *self)
 static gboolean
 tls_context_setup_ecdh(TLSContext *self)
 {
-  /* server only */
-  if (self->mode != TM_SERVER)
-    return TRUE;
-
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
I just saw this is a bugfix. In my understanding, this patch makes it possible to restrict curves on the client-side too.
Does it fix anything else I missed?

MrAnno:
It turned out that not all curves are available by default, so this option can also be used to extend the allowed list.

Issue #3296, opened 2020-05-20
Implement journal namespaces


bjoe2k4:
## Description of the problem
In systemd v245 the concept of namespaces for the system journal was introduced. Journal namespaces are a convenient way of grouping unrelated services. (Merged pull: https://github.com/systemd/systemd/pull/14178)

According to @MrAnno on the #syslog-ng channel freenode IRC, there is currently no convenient way to get these namespaced journal messages into syslog-ng in the same sane and safe way it currently works for the default namespace via `systemd-journal()`, which calls `sd_journal_open()` with the flag `SD_JOURNAL_LOCAL_ONLY` from systemd. For namespaces a new function `sd_journal_open_namespace()` was introduced, which has all the functionality of `sd_journal_open()`, but accepts a `namespace` parameter and two additional valid flags `SD_JOURNAL_ALL_NAMESPACES` and `SD_JOURNAL_INCLUDE_DEFAULT_NAMESPACE`. 

```c
_public_ int sd_journal_open_namespace(sd_journal **ret, const char *namespace, int flags)
```

The original `sd_journal_open()` function is actually just a wrapper:

```c
_public_ int sd_journal_open(sd_journal **ret, int flags) {
        return sd_journal_open_namespace(ret, NULL, flags);
}
```

To query the journal from the console, there are now four ways.
1. just `journalctl` uses `sd_journal_open()` (actually `sd_journal_open_namespace()` with namespace NULL) and flags `SD_JOURNAL_LOCAL_ONLY`, so this is what syslog-ng currently does as well, but via the journal API.
2. `journalctl --namespace=*` uses `sd_journal_open_namespace()` with flags `SD_JOURNAL_ALL_NAMESPACES` and  `SD_JOURNAL_LOCAL_ONLY`
3. `journalctl --namespace=foo` uses `sd_journal_open_namespace()` with namespace `foo` and flags  `SD_JOURNAL_LOCAL_ONLY`
4. `journalctl --namespace=+foo` uses `sd_journal_open_namespace()` with namespace `foo` `SD_JOURNAL_INCLUDE_DEFAULT_NAMESPACE` and `SD_JOURNAL_LOCAL_ONLY`

While way 4 is probably good for debugging, it seems less useful in the syslog-ng context to me.

## Proposed solution

[...]

PR #3358, opened 2020-07-18
systemd-journal: collect all namespaces

bjoe2k4:
This pull adds the logs of all journal namespaces by default. By default only the default namespace is defined in systemd. So this should not cause any inconvenience for current users.

This works on my system, but obviously it is not complete. Systemd version must be checked, since `SD_JOURNAL_ALL_NAMESPACES` might not be defined for systemd < 245. Any help appreciated.

Closes: #3296


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build FAILURE

modules/systemd-journal/journal-reader.c
```diff
@@ -598,7 +598,7 @@ _init(LogPipe *s)
   if (!log_source_init(s))
     return FALSE;
 
-  gint res = journald_open(self->journal, SD_JOURNAL_LOCAL_ONLY);
+  gint res = journald_open(self->journal, SD_JOURNAL_LOCAL_ONLY | SD_JOURNAL_ALL_NAMESPACES);
```
Kokan, line 601: According to the man page:

```
       sd_journal_open() opens the log journal for reading. It will find all journal files automatically and interleave them automatically when reading. As first argument it takes a
       pointer to a sd_journal pointer, which, on success, will contain a journal context object. The second argument is a flags field, which may consist of the following flags ORed
       together: SD_JOURNAL_LOCAL_ONLY makes sure only journal files generated on the local machine will be opened.  SD_JOURNAL_RUNTIME_ONLY makes sure only volatile journal files will
       be opened, excluding those which are stored on persistent storage.  SD_JOURNAL_SYSTEM will cause journal files of system services and the kernel (in opposition to user session
       processes) to be opened.  SD_JOURNAL_CURRENT_USER will cause journal files of the current user to be opened. If neither SD_JOURNAL_SYSTEM nor SD_JOURNAL_CURRENT_USER are
       specified, all journal file types will be opened.

       sd_journal_open_namespace() is similar to sd_journal_open() but takes an additional namespace parameter that specifies which journal namespace to operate on. If specified as NULL
       the call is identical to sd_journal_open(). If non-NULL only data from the namespace identified by the specified parameter is accessed. This call understands two additional flags:
       if SD_JOURNAL_ALL_NAMESPACES is specified the namespace parameter is ignored and all defined namespaces are accessed simultaneously; if SD_JOURNAL_INCLUDE_DEFAULT_NAMESPACE the
       specified namespace and the default namespace are accessed but no others (this flag has no effect when namespace is passed as NULL). For details about journal namespaces see
       systemd-journald.service(8).
```

The `sd_journal_open` does not handle this flag according to the man page. I would either consider moving to the new call (when available) or get it confirmed that this is a documentation bug.

Kokan, line 601: Passing this flag should be configurable, as this could be a not desired behaviour. Making this as a default is okay.

Kokan, line 601: Resolving the case when the `SD_JOURNAL_ALL_NAMESPACES` does not exists depends on how you resolve the my first comment.
But one way you could do is similar to pcre jit flag:
https://github.com/syslog-ng/syslog-ng/blob/ffe25103987009b7afddce1220ebd6d11820b3f2/lib/compat/pcre.h#L35-L37

bjoe2k4: `sd_journal_open` is defined as:

https://github.com/systemd/systemd/blob/12d14b73732df8c3ca46383573eeea069832969f/src/journal/sd-journal.c#L1971-L1973

So it will naturally allow this flag. Nevertheless, I agree that using the new function is a better way and circumvents all documentation ambiguities.

bjoe2k4: agreed

bjoe2k4: I'm afraid it's not that easy. `SD_JOURNAL_ALL_NAMESPACES` is defined in an enum in `<systemd/sd-journal.h>`, so the preprocessor won't know if it's defined or not. For example:

```
#include <stdio.h>
#include <systemd/sd-journal.h>

int main() {

printf("%d\n", SD_JOURNAL_ALL_NAMESPACES);

#ifndef SD_JOURNAL_ALL_NAMESPACES
# define SD_JOURNAL_ALL_NAMESPACES 64
#endif

printf("%d\n", SD_JOURNAL_ALL_NAMESPACES);

return 0;

}
```
outputs
```
32
64
```
but i'd expect
```
32
32
```

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bjoe2k4:
I hope this works as desired now and will be considered for the `syslog-ng v3.29` milestone.

modules/systemd-journal/journal-reader.c
Kokan: When possible to do either runtime and build time configuration, in this project the runtime is one to go.
The suggestion was to introduce a config file option either setting it or not.

This future of systemd introduces two option, one with this flag and one with the specifying the namespace. 

```
systemd-journal( namespace('foo') );

systemd-jouranl( namespace('') ); # invalid
systemd-journal( all-namespaces( yes/ no ) );

systemd-journal( namespace('bar') all-namespaces( yes ) ); #invalid
```
In the last case it could be said that the `all-namespaces` are stronger, but imho it is better simply not accepting so user could make informed decision.

Of course it is fine, if this PR now only focuses on the `all-namespaces(yes/no)` and not implement the namespaces.

bjoe2k4: I've tried to tackle the all-namespaces option. Somewhere must be an error. No matter if I set yes or no, all namespaces are always included.

The namespace() option i won't do, because i think this requires bigger changes.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/systemd-journal/journal-reader.c
```diff
@@ -725,6 +740,7 @@ journal_reader_options_init(JournalReaderOptions *options, GlobalConfig *cfg, co
           options->prefix = g_strdup(default_prefix);
         }
     }
+  options->all_namespaces = TRUE;
```
szemere, line 743: From the point of work it is equivalent, but this looks more like a "default value" than an initialization step. Consider putting it in `journal_reader_options_defaults`

modules/systemd-journal/journal-reader.c
```diff
@@ -598,7 +598,22 @@ _init(LogPipe *s)
...
+
+#if SYSLOG_NG_HAVE_JOURNAL_NAMESPACES
+  if (self->options->all_namespaces)
+    {
+      msg_error("all_namespaces");
+      journal_flags |= SD_JOURNAL_ALL_NAMESPACES;
+    }
+  gint res = journald_open_namespace(self->journal, NULL, journal_flags);
+#else
+  if (!self->options->all_namespaces)
```
szemere, line 611: This is just a guess, but I think the negation was used because of the default value. Which is TRUE, regardless of the systemd version in place.
I would like to start a discussion about the expected behavior. To give a proper message to the user, a binary state variable is not enough for this warning. Since you also need a "not set" state.

My proposal: 
I am thinking about a more general solution, where we simply introduce an option, called "namespace" (as a string list), which will currently ONLY handle the "ALL" option.
If the option is not set at all, or systemd version is less than 2.45, than it will not have any effect.
This option later can be extended to support actual namespaces.

bjoe2k4: Your guess is correct. I wanted to issue the warning only when someone tries to do something impossible on their system. The only case i could imagine was setting all_namespaces(no) on a <v245 system. But i agree, there should be a not set state.

I've tried to implement the suggested namespace() option instead off all-namespaces().

kira-syslogng:
Build SUCCESS

modules/systemd-journal/journal-reader.c
szemere: I just wanted to start a discussion. You went straight ahead, really nice. I would wait for others opinion, but I like the direction. 

note: When squashing the commits together at the end, I think this part should be extracted into a function. So the whole "namespace string" handling logic will be encapsulated.

bjoe2k4: I'm not very opinionated regarding the implementation, but also C is not really my strength. So I'll go with whatever is proposed and sound.

I've encapsulated the journal opening in a separate function and added a file in the news/ folder. Is there any documentation which needs to be updated? At least I couldn't find anything in this git repo.

Question: Why is the `systemd-journal()` source limited to one instance? Would it make sense to allow for more than one instance for different namespaces? Just thinking out loud here. Otherwise there might to be a lot of filtering to be done when collecting all namespaces and filtering by namespace.

kira-syslogng:
Build SUCCESS

modules/systemd-journal/journal-reader.c
szemere: The extracted function looks good, and I think the news file is enough for the PR. (I notified the documentation team, to add the new option to the admin guide too.)

Limit to one instance: I think this is coming from the old "namespace less" version. There is no reason to add the same source multiple times. (It might also cause race conditions.)
With your modification I think it would be logical to allow multiple instances. However it have to be checked globally to make sure the namespaces are unique between sources, and only allow one "default" namespace source at once. (I would rather not widening the scope of this PR. It can be done in a separate one.)

modules/systemd-journal/journal-reader.c
```diff
@@ -584,6 +584,37 @@ _add_poll_events(JournalReader *self)
...
+
+  if (strcmp(self->options->namespace, "*") == 0)
+    {
+      journal_flags |= SD_JOURNAL_ALL_NAMESPACES;
+    }
+  else if (strncmp(self->options->namespace, "+", 1) == 0)
+    {
+      journal_flags |= SD_JOURNAL_INCLUDE_DEFAULT_NAMESPACE;
+      journal_namespace = self->options->namespace;
+      memmove(journal_namespace, journal_namespace+1, strlen(journal_namespace));
```
szemere, line 602: I think the length should be `strlen(journal_namespace)-1`. However I would rather not modify the "internal" data of the driver.

My suggestion: Simply do pointer arithmetics, where the helper variable `journal_namespace` points to `self->options->namespace+1` , leaving out the `memmove` operation.

szemere:
I think we are getting close to finish this PR. (Thank you again for your work!) If you have time, please squash the commits, so it will be easier to review for the second reviewer.

modules/systemd-journal/journal-reader.c
bjoe2k4: Agreed.

modules/systemd-journal/journal-reader.c
bjoe2k4: Ok. I have to admit here that C is not really my strong side...
Is it actually necessary to check for `strlen(self->options->namespace) > 1`  here?

kira-syslogng:
Build SUCCESS

modules/systemd-journal/journal-reader.c
szemere: Sorry for being ambiguous.

I mean, the original code has a buffer overflow problem this way. You read buffer amount of bytes from the buffer+1 position.
My original intention was to change:
`  memmove(journal_namespace, journal_namespace+1, strlen(journal_namespace));`
To this:
`  memmove(journal_namespace, journal_namespace+1, strlen(journal_namespace)-1);`


My current suggestion is to completely emit the `memmove` operation:
```
{
  journal_flags |= SD_JOURNAL_INCLUDE_DEFAULT_NAMESPACE;
  journal_namespace = self->options->namespace+1;
}
```

This way the local `journal_namespace` variable will point to the second character of the `self->options->namespace` string, without making any modification on it.

bjoe2k4: What if someone decides to pass `namespace("+")`? Therefore i have added 
```c
      if (strlen(self->options->namespace) > 1)
        {
          journal_namespace = self->options->namespace + 1;
        }
```
in the squashed commit. If it's not necessary I can remove it again.

szemere: Checking the length was an excellent idea!

Sorry for not answering earlier. Inspired by your check, I started to experiment with the `journalctl --namespace=` command. I tried to figure out some "special cases". Maybe there are other options which are not documented, but should be handled by us. (I mainly focused on program crashes, instabilities. i.e.: a problematic config might corrupt the cursos position.)

I compiled your code on Focal Fossa (Ubuntu 20.04) - supports systemd 2.45. Generated log entries in some custom namespaces. (So I can easily filter on them.) And checked different filter options both via the command line `journalctl --namespace` interface, and via the `sd_journal_open_namespace` API.

Here are my findings:

- `*` : works as expected.
- `*+`: no errors or crash, but gives 0 result.
- `*bar`: no errors or crash, but gives 0 result.
My conclusion: `*` can be given only alone, nothing else. Note: your current strcmp check is perfect for this. Minor note: an error message might be handy. i.e.: somebody accidentally makes a typo error.
- `+`: strange case, since the + sign originally intended for "ALSO include the default namespace", however if ONLY + sign was given, both the command line and the API will print the default namespace without any error.
- `+bar`: (bar exists): works as expected.
- `+barrr` (barr not exists): works as expected, gives the default namespace and nothing else, since there are no existing namespace to include.
- `+*` : works in the same way as ONLY the + sign was given. No error message, but only prints the default namespace.
- `barrr`: (barr not exists): no errors or crash. Gives 0 result.


One more note: Since I tested not existing namespaces too, I tested the case where I add a new namespace to journal AFTER the command line (journalctl: -f follow mode) or Syslog-ng has started. Both of them ignores the new namespace on runtime. So one must reload Syslog-ng (or the command line tool) to collect logs from a new namespace.


Based on all above: I think the current comparisons and checkings are good, and handle all special cases. (Maybe an extra warning can be helpful, to catch typo errors with the * sign.)
Only one problem left, the `journal_namespace` variable is left uninitialised, I will leave a separate comment for that, so it will be easier to track.

modules/systemd-journal/journal-reader.c
```diff
@@ -584,6 +584,39 @@ _add_poll_events(JournalReader *self)
   return TRUE;
 }
 
+static gint
+_journal_open(JournalReader *self)
+{
+#if SYSLOG_NG_HAVE_JOURNAL_NAMESPACES
+  gint journal_flags = SD_JOURNAL_LOCAL_ONLY;
+  gchar *journal_namespace;
```
szemere, line 592:
```suggestion
  gchar *journal_namespace = NULL;
```

If the `self->options->namespace == "*"` than we use journal_namespace as an argument in https://github.com/syslog-ng/syslog-ng/pull/3358/files#diff-8506b8af7604a75a9a5108a18b1477dcR613 without initialising it's value. Giving `NULL` as a namespace string is acceptable together with the `SD_JOURNAL_ALL_NAMESPACES` flag, so the code is good, only the variable is not initialised properly.

bjoe2k4: Thx, I've modified the squashed commit accordingly.

modules/systemd-journal/journal-reader.c
bjoe2k4: Thanks for the detailed analysis. I've added a warning for the case of `*bla`.

Adding a namespace on the fly is probably something we cannot cover here, but worth to mention in the docs.

Also I've noticed that whenever the `namespace()` option is changed, i am seeing the following error message:
```
syslog-ng[206612]: [2020-08-11T00:03:29.516217] Failed to step cursor; error='Success (0)'
```

kira-syslogng:
Build SUCCESS

bazsi:
On Tue, Aug 11, 2020, 00:20 Andreas Fischer <notifications@github.com>
wrote:

> *@bjoe2k4* commented on this pull request.
>
> In modules/systemd-journal/journal-reader.c
> <https://github.com/syslog-ng/syslog-ng/pull/3358#discussion_r468217982>:
>
> > +static gint
> +_journal_open(JournalReader *self)
> +{
> +#if SYSLOG_NG_HAVE_JOURNAL_NAMESPACES
> +  gint journal_flags = SD_JOURNAL_LOCAL_ONLY;
> +  gchar *journal_namespace;
> +
> +  if (strcmp(self->options->namespace, "*") == 0)
> +    {
> +      journal_flags |= SD_JOURNAL_ALL_NAMESPACES;
> +    }
> +  else if (strncmp(self->options->namespace, "+", 1) == 0)
> +    {
> +      journal_flags |= SD_JOURNAL_INCLUDE_DEFAULT_NAMESPACE;
> +      journal_namespace = self->options->namespace;
> +      memmove(journal_namespace, journal_namespace+1, strlen(journal_namespace));
>
> Thanks for the detailed analysis. I've added a warning for the case of
> *bla.
>
> Adding a namespace on the fly is probably something we cannot cover here,
> but worth to mention in the docs.
>
> Also I've noticed that whenever the namespace() option is changed, i am
> seeing the following error message:
>
> syslog-ng[206612]: [2020-08-11T00:03:29.516217] Failed to step cursor; error='Success (0)
>
>
The namespace option may need to be added to the key of the persist state
where we save the last cursor position.

That way we could possibly multiple sources with different namespaces and
this error would probably not be triggered.

Not sure about this as there are consequences, like compatibility with
older versions.

>
>

szemere:
(Following the conversation here, and mark the earlier thread as resolved)

After some trial and error, it looks like the _CURSOR position - which is returned by journald - is somehow "namespace specific". I agree with @bazsi's suggestion. Storing the namespace in the persist file will solve this problem, but it will probably cause an issue for older versions. (Reading all entries again, at the best.) So we have to figure out some "upgrade mechanism", where we convert "namespace less" entries to the new format first.

 While experimenting with the different namespaces, and cursor positions I have found one another issue with cursor handling. If Syslog-ng is stopped before reading any entry from the journal, it will store an empty cursor position into the persist file. Which is causing issues at the next start.

@bjoe2k4 I think your code is good. The cursor issue is quite unfortunate and came from earlier design choices. Please consider if you are willing to work on that part too, or I can send a pull request to you which will address that part.

bjoe2k4:
@szemere I'm relatively short on time at the moment and unfamiliar with the persistent state. I'd merge any pull on my repo.

szemere (approved):

Kokan (approved):

PR #3361, opened 2020-07-21
stats-counter: do not crash when set an external/read-only counter

lbudai:
When trying to reset a read-only stats counter with
`syslog-ng-ctl query get '*' --reset`
syslog-ng crashed before this patch.

As this situation can be triggered by users, it does not
count as a programming error, so using assert is not reasonable.

Reported-by: Peter Kokai <peter.kokai@balabit.com>


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Issue #3345, opened 2020-07-01
Feature request - signal destination


nambrosch:
## Description of the problem
I'm interested in a new native destination for signal so messages can be sent directly to signal users.

## Proposed solution
Currently a telegram destination exists but many do not use it because of privacy concerns .  Here are a few solutions I found for sending messages, signals-cli works as advertised:

- https://github.com/signalapp/libsignal-service-java
- https://gitlab.com/thefinn93/signald
- https://github.com/AsamK/signal-cli

## Alternatives
As an alternative, messages can be piped to an external command.  Not as elegant but it works.

## Additional context
Signal is a messenger available on mobile and desktop that focuses on encryption, security, and privacy.

- https://signal.org/
- https://www.maketecheasier.com/whatsapp-vs-signal-vs-telegram/

szemere:
Hello @nambrosch,
 thank you for your detailed description! Since I don't use Signal, I started with a smaller research in the topic. Let me share you my findings:

 - Signal has a strong basics on encryption. IMHO any command line tool weakens that concept.
With other words: once you open your encrypted communication channel via an unprotected (or one with a weaker protection) command line interface, you ruined that whole ecosystem. In that case you can even use Telegram's REST interface.
However, one can definitely implement such a messaging system with the help of Syslog-ng's program destination.
 - Signal protocol requires to store some generated keys in a secured manner. Syslog-ng has a "secret-storage" for runtime secrets, but it is not permanent. (I don't know yet, if they must be stored or can be regenerated on each and every run.)
Worst case scenario, the destination will require a "set up phase" at the start of Syslog-ng, like passphrase protected keys in case of TLS connections.
- Luckily there is a C library for Signal https://github.com/signalapp/libsignal-protocol-c, so it can be implemented as a native destination driver inside Syslog-ng, which will gives us much more room to play than a strict REST interface. (Providing the necessary secrets might still be a problem.)

Summarizing it:
 I think the CLI based solution can be done in a few steps, however it has HUGE security disadvantages. Developing it as a native destination driver is far from trivial and requires a lot of attention from the security aspect. Because of that, I can not give you any guarantee when it will be implemented in Syslog-ng.
 If you would like to contribute to Syslog-ng and implement one of the options above, we can assist you in the development.


Best regards,
Laci

bazsi:
The security comment on the cli interface is only true if you dont trust
the host running syslog-ng.

nambrosch:
[...]

PR #3362, opened 2020-07-22
program destination wrapper for signal-cli

szemere:
Wrapping the https://github.com/AsamK/signal-cli binary into a program destination of syslog-ng.
resolves: #3345

# The destination requires manual setup before use!
Please refer to the official signal-cli documentation for
further information about registering it as a client:
https://github.com/AsamK/signal-cli#usage

# Example usage:

```
destination d_signal_cli{
  signal-cli-dest(
    signal-cli-executable("/usr/local/bin/signal-cli")
    own-number("+36001111111")
    partner-number("+36002222222")
  );
};
```

# WARNING! Performance measurements
Syslog-ng program destination assumes that the consumer has a
"daemon" like behaviour. The program will be started by Syslog-ng
at initialization time, and it will receive new messages on it's
standard input, separated by new line characters.

This aproach is very intentional from Syslog-ng. Since starting
and freeing up a process is a very resource intensive work,
doing it on per message basis can put a system on it's kneels.

In order to make signal-cli feasible we have to write a small
"wrapper" script, which achieves this "daemon like" behaviour.
Thus, this implementation is only intended to use for "alerts"
and "notifcations". Not for implementing heavy duty message
delivery pipelines.

### DIFF
--- a/news/feature-3362.md
+++ b/news/feature-3362.md
@@ -1,0 +1,39 @@
+signal: use signal-cli via a program destination
+
+Adding Signal (https://signal.org) destintaion to Syslog-ng,
+relying on the work of https://github.com/AsamK/signal-cli
+
+# The destination requires manual setup before use!
+Please refer to official signal-cli documentation for
+further information about registering it as a client:
+https://github.com/AsamK/signal-cli#usage
+
+
+# Example usage:
+
+```
+destination d_signal_cli{
+  signal-cli-dest(
+    signal-cli-executable("/usr/local/bin/signal-cli")
+    own-number("+36001111111")
+    partner-number("+36002222222")
+  );
+};
+```
+
+
+# WARNING! Performance measurements
+Syslog-ng program destination assumes that the consumer has a
+"daemon" like behaviour. The program will be started by Syslog-ng
+at initialization time, and it will receive new messages on it's
+standard input, separated by new line characters.
+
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lbudai:
hmmm... a program destination that executes a java program per message?

Java runtime and the libraries are hidden dependencies and to be honest, I don't really like this.
What about packaging? (I mean I'm not really plan to build the dependencies even in my OBS environment....)

I think we should checking the runtime requirements... we should report a correct error message about the missing cli and how to make it work. When the java binary is not installed, user won't notice anything - the read loop will 'process' the message (and do nothing with that).

Execute a java program for each message is not something I'd merge.
What do you think about implementing it as a java dst? Or python? (Best would be to implement this as a native destination based on the C library. We have password protected ssl key files, would it be possible to use the secret storage? )

szemere:
Since this solution has a huge performance impact (as @lbudai also stated), I would rather not merge it.

Thanks to @czanik the work is not lost. He transformed it into a basic "how to guide" explaining all the caveats, while also highlighting all the performance issues with this solution. For more details, please read Peter's post: https://www.syslog-ng.com/community/b/blog/posts/sending-alerts-to-signal-messenger-from-syslog-ng

PR #3364, opened 2020-07-24
syslog parser fix crash during handling short inputs

### LINKED ISSUE #3328: Abnormal message causes Segmentation fault
# syslog-ng
## Version of syslog-ng
```
syslog-ng 3 (3.27.1)
Config version: 3.22
Installer-Version: 3.27.1
Revision:
Module-Directory: /usr/local/lib/syslog-ng
Module-Path: /home/sense/test-fuzz/syslog-ng-3.27.1/inst/usr/local/lib/syslog-ng
Include-Path: /usr/local/share/syslog-ng/include
Available-Modules: stardate,basicfuncs,add-contextual-data,xml,examples,hook-commands,disk-buffer,appmodel,azure-auth-header,confgen,map-value-pairs,afsocket,affile,tfgetent,graphite,syslogformat,cryptofuncs,kvformat,secure-logging,cef,linux-kmsg-format,csvparser,tags-parser,afprog,afsql,afuser,timestamp,system-source,pseudofile,dbparser
Enable-Debug: on
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: off
Enable-TCP-Wrapper: off
Enable-Linux-Caps: off
Enable-Systemd: off
```

## Platform
Debian 10

## Issue
The message, that contains just bytes: "4\0  " causes Segmentation fault

## Steps to reproduce
```
printf '4\0  ' | syslog-ng \
        --foreground \
        --cfgfile=$CONFIG
Segmentation fault
```

## Configuration
```
@version:3.22

source s_src {
        stdin(follow-freq(1));
};

destination d_dst {
        file("/tmp/dest.log");
};

log {
        source(s_src);
        destination(d_dst);
};
```

Kokan:
Fixes #3328.

### DIFF
--- a/modules/syslogformat/syslog-format.c
+++ b/modules/syslogformat/syslog-format.c
@@ -48,7 +48,7 @@ static struct
 } handles;
 
 static inline gboolean
-sd_step(const guchar **data, gint *left)
+_process_any_char(const guchar **data, gint *left)
 {
   if (*left < 1)
     return FALSE;
@@ -70,7 +70,7 @@ log_msg_parse_pri(LogMessage *self, const guchar **data, gint *length, guint fla
 
   if (left && src[0] == '<')
     {
-      sd_step(&src, &left);
+      _process_any_char(&src, &left);
       pri = 0;
       while (left && *src != '>')
         {
@@ -82,12 +82,12 @@ log_msg_parse_pri(LogMessage *self, const guchar **data, gint *length, guint fla
             {
               return FALSE;
             }
-          sd_step(&src, &left);
+          _process_any_char(&src, &left);
         }
       self->pri = pri;
       if (left)
         {
-          sd_step(&src, &left);
+          _process_any_char(&src, &left);
         }
     }
   /* No priority info in the buffer? Just assign a default. */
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
Looks good to me. If not too difficult, could you add a unit test for this case?

Kokan:
@furiel Do you want me to add test for the case of skipping sequence id, or crashing due to short input ? Either of those UT are not difficult.

If possible I would avoid adding a test cases that was generated with fuzzing. The reason that I could add a lot of them, and instead the idea is to add a test that does fuzzing as the input generated by fuzzer that actually causes crashes are implementation dependent. I would rather create UT based on functionality and on coverage bases.

The fuzzer test is actually on its way, I just curved this PR out of it as it was possible to do so.

Do you think that it would still worth to add such case ?

furiel:
@Kokan I agree the tests by the fuzzer are superior. Thanks!

PR #3365, opened 2020-07-28
syslog-parser fuzzing with libfuzzer

Kokan:
Note:
* As of the moment it does not have autotools support (use cmake)
* The description is not updated based on latest greatest fuzzing options (but it should be still relevant)

The libfuzzer is a code coverage based fuzzer provided by LLVM[1]. The aim of
this fuzzing target introduced to the syslogformat to mitigate ill-formed inputs
breaking the application, causing crash, buffer overflow, null pointer
dereference. In order to detect such issues the fuzzer can and should be
combined with sanitizers such as ASAN, MSAN, UBSAN[2]. By using the existing
`-DSANITIZER={}` option to specify different kind of sanitizers (multiple is
possible), the sanitizers listed with that option is going to be combined with
the fuzzer. The libfuzzer could use `corpus`-es as initial inputs, patterns or
dictionary to compose new inptus based on coverage and randomness. Due to the
random execution it is not a set of tests that should be executed for each code
change, rather a test that should be exectued continuously. The target is not
part of the default UT test set, it must be executed manually as of now.

In order to run the fuzzer one could simply call the executable and wait as long
as it finds some crash - at which point it is going to stop, or a desired amount
of time and stop the executable, as otherwise it runs indefenetly by default.

The issues found in the next patchset is found by the following specialization of
the fuzzer:
```
./modules/syslogformat/tests/test_fuzzing_syslogformat ../modules/syslogformat/tests/corpus/ -print_pcs=1 -detect_leaks=1 -print_final_stats=1 -print_corpus_stats=1 -use_value_profile=1  -max_len=256
```

The `-use_value_profile=1` option helped to find most of the issues, despite its
being currently an experimental feature.

[1] http://llvm.org/docs/LibFuzzer.html
[2] http://llvm.org/docs/LibFuzzer.html#fuzzer-usage


kira-syslogng:
Build FAILURE

modules/syslogformat/tests/corpus/pattern
```diff
@@ -0,0 +1 @@
+<45>1 2015-08-12T13:23:12+02:00 zts-win101 syslog-ng 2876 2 [meta sequenceId=\"99\"] syslog-ng Agent shutting down; version='5.5.0alpha
```
furiel, line 1: Why did you escape 99? Copy paste error?

modules/syslogformat/tests/test_fuzzing_syslogformat.c
```diff
@@ -0,0 +1,35 @@
...
+#include <iv.h>
+
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+  if (size <= 1) return 0;
+
+  GlobalConfig *cfg = cfg_new_snippet();
+  app_startup();
+  cfg_load_module(cfg, "syslogformat");
```
furiel, line 19: How does this cfg variable works? I do not understand how it is used by syslog format. Through some side effect?

modules/syslogformat/tests/test_fuzzing_syslogformat.c
```diff
@@ -0,0 +1,35 @@
...
+  msg_format_options_defaults(&parse_options);
+
+  LogMessage *msg = log_msg_new_empty();
+
+  syslog_format_handler(&parse_options, data, size, msg);
+
+  log_msg_unref(msg);
+  cfg_free(cfg);
+  app_shutdown();
+  iv_deinit();
```
furiel, line 30: Why do we need this `iv_deinit`?

modules/syslogformat/tests/corpus/pattern
furiel: I mean why the quotes are escaped around 99.

furiel:
@Kokan I tried your code and it worked nicely.

I needed to upgrade clang: xenial ships 3.8, while 6.0 is needed, but there is an apt repository, so the upgrade went smoothly too.

I was fighting a little with passing the compiler to cmake. Strangely, setting cc=clang did not work. On the other hand, setting with -DCMAKE_C_COMPILER=clang worked.
For this kind of experiments, it would be nice if the warning about the wrong compiler would be a a fatal error.

After that, everything worked well. It also found another crashing pattern almost instantly.

modules/syslogformat/tests/corpus/pattern
Kokan: yeah, just copy paste error

modules/syslogformat/tests/test_fuzzing_syslogformat.c
Kokan: It is not "used" by syslog format that's the trick. Rather by this `cfg_load_module` which is a proper way to load a module. This part probably could be replaced with the `syslog_format_init` function.

modules/syslogformat/tests/test_fuzzing_syslogformat.c
Kokan: I do not recall if it was a crash or simply memory leak. But neither is acceptable for fuzzing.
Actually it should be called by `app_shutdown`, but it is not
https://github.com/syslog-ng/syslog-ng/blob/787f588d2ac6d91a40b3289e8857e2487586efb7/lib/apphook.c#L228-L238

Yes, most of our UT (where `app_startup` is used) should call `iv_deinit` but we can get away with it, as it is not called multiple times (sandboxing the tests "helps").

Kokan:

Kokan:
(Just testing if job disabled properly.)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3368, opened 2020-07-30
Cfg version current

lbudai:
From release to release we `sed` previous version values to the current version value...
I think it could be useful if we would have a `@version: current` syntax in the config file.
This PR implements this.

### DIFF
--- a/news/feature-3368.md
+++ b/news/feature-3368.md
@@ -1,0 +1,1 @@
+config: support `@version: current`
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

news/feature-3368.md
```diff
@@ -0,0 +1 @@
+config: support `@version: current`
```
gaborznagy, line 1: just nitpicking note: when documented I would highlight that this is a keyword and not a string, so `@version: "current"` won't work.

kira-syslogng:
Build SUCCESS

bazsi:
I don't think we should document this either in NEWS nor in the documentation, see my comments in #3293

PR #3369, opened 2020-07-30
Use cfg version current

lbudai:
IMHO news file is not required - it's internal and straightforward

### DIFF
--- a/lib/tests/test_lexer.c
+++ b/lib/tests/test_lexer.c
@@ -312,13 +312,6 @@ Test(lexer, at_version_stores_config_version_in_parsed_version_in_hex_form)
   parser->lexer->ignore_pragma = FALSE;
 
   cfg_set_version_without_validation(configuration, 0);
-  _input("@version: 3.28\n\
-foo\n");
-  assert_parser_identifier("foo");
-  cr_assert_eq(configuration->user_version, 0x031c,
-               "@version parsing mismatch, value %04x expected %04x", configuration->user_version, 0x031c);
-
-  cfg_set_version_without_validation(configuration, 0);
   _input("@version: 3.1\n\
 bar\n");
   assert_parser_identifier("bar");
@@ -331,6 +324,18 @@ baz\n");
   assert_parser_identifier("baz");
   cr_assert_eq(configuration->user_version, 0x0305,
                "@version parsing mismatch, value %04x expected %04x", configuration->user_version, 0x0305);
+}
+
+Test(lexer, current_version)
+{
+  parser->lexer->ignore_pragma = FALSE;
+
+  cfg_set_version_without_validation(configuration, 0);
+  _input("@version: current\n\
+foo\n");
+  assert_parser_identifier("foo");
+  cr_assert_eq(configuration->user_version, VERSION_VALUE_CURRENT,
+               "@version parsing mismatch, value %04x expected %04x", configuration->user_version, VERSION_VALUE_CURRENT);
 }
 
 Test(lexer, test_lexer_others)
--- a/lib/tests/test_pragma.c
+++ b/lib/tests/test_pragma.c
@@ -28,9 +28,9 @@
 
 Test(pragma_parser, process_valid_major_minor)
 {
-  const guint version = process_version_string("3.28");
+  const guint version = process_version_string("3.27");
 
-  cr_assert_eq(0x031c, version);
+  cr_assert_eq(0x031b, version);
 }
 
 Test(pragma_parser, process_version_large_minor)
--- a/modules/system-source/system-expand.sh
+++ b/modules/system-source/system-expand.sh
@@ -29,7 +29,7 @@ out="$(mktemp)"
 create_config () {
         local cfg="$1"
         cat >"${cfg}" <<EOF
-@version: 3.28
+@version: current 
 @include "scl.conf"
 source s_system {
 # ----- Cut here -----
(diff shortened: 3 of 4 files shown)


kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
```diff
@@ -70,7 +70,7 @@ def set_version(self, version):
         self.__syslog_ng_config["version"] = version
 
     def get_version(self):
-        return self.__syslog_ng_config["version"]
+        return "current"
```
Kokan, line 73: I like the opportunity of setting and getting back the version value for the configuration (even if that says `current`). imho it would be better to leave this getter alone, and use the `current` where the `SyslogNgConfig` getting the version information:
https://github.com/syslog-ng/syslog-ng/blob/3d5b1e1a4b62d2f3958e93fb2ea72fb825844117/tests/python_functional/conftest.py#L103-L105

lbudai: Maybe it is better to leave it alone and not touching this part of the code...

kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
Kokan: I am fine with that as well.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

scl/syslog-ng.conf
```diff
@@ -3,7 +3,7 @@
 # single file called /var/log/messages.
 #
 
-@version: 3.28
+@version: current
```
bazsi, line 6: Please don't do this in our default configuration. It is our goal to have a proper @version in user configurations.

lib/tests/test_lexer.c
```diff
@@ -312,11 +312,11 @@ Test(lexer, at_version_stores_config_version_in_parsed_version_in_hex_form)
...
 
   cfg_set_version_without_validation(configuration, 0);
-  _input("@version: 3.28\n\
+  _input("@version: current\n\
 foo\n");
   assert_parser_identifier("foo");
-  cr_assert_eq(configuration->user_version, 0x031c,
-               "@version parsing mismatch, value %04x expected %04x", configuration->user_version, 0x031c);
+  cr_assert_eq(configuration->user_version, VERSION_VALUE_CURRENT,
+               "@version parsing mismatch, value %04x expected %04x", configuration->user_version, VERSION_VALUE_CURRENT);
```
bazsi, line 320: with this change, this test would not test what its intention was. To parse a version number in hex form. If we want, we can add a new testcase for `@version: current`

lib/tests/test_pragma.c
```diff
@@ -28,9 +28,9 @@
 
 Test(pragma_parser, process_valid_major_minor)
 {
-  const guint version = process_version_string("3.28");
+  const guint version = process_version_string("3.27");
 
-  cr_assert_eq(0x031c, version);
+  cr_assert_eq(0x031b, version);
 }
```
bazsi, line 34: this should not need to be changed every time the version number is updated, so we can leave it at whatever version, without changing it every time we do a release.

modules/python/python-example.conf
```diff
@@ -1,4 +1,4 @@
-@version: 3.28
+@version: current
```
bazsi, line 1: again, we shouldn't use `current` in configs we are using as examples. internal test usage is ok.

scl/syslog-ng.conf
lbudai: I'll modify it.
IMO users can easily change it to a specific version...
What are the situations when a user want to change `@version` ?
When we do a change which is not compatible with the previous versions - in this case we show a warning message and in a later version we will break compatibility. In this case this warning is showed as expected - user can replace the `current`  with a specific version... so it requires the very same act from the user.

lib/tests/test_lexer.c
lbudai: here is the whole context:
```
Test(lexer, at_version_stores_config_version_in_parsed_version_in_hex_form)
{
  parser->lexer->ignore_pragma = FALSE;

  cfg_set_version_without_validation(configuration, 0);
  _input("@version: 3.28\n\
foo\n");
  assert_parser_identifier("foo");
  cr_assert_eq(configuration->user_version, 0x031c,
               "@version parsing mismatch, value %04x expected %04x", configuration->user_version, 0x031c);

  cfg_set_version_without_validation(configuration, 0);
  _input("@version: 3.1\n\
bar\n");
  assert_parser_identifier("bar");
  cr_assert_eq(configuration->user_version, 0x0301,
               "@version parsing mismatch, value %04x expected %04x", configuration->user_version, 0x0301);

  cfg_set_version_without_validation(configuration, 0);
  _input("@version: 3.5\n\
baz\n");
  assert_parser_identifier("baz");
  cr_assert_eq(configuration->user_version, 0x0305,
               "@version parsing mismatch, value %04x expected %04x", configuration->user_version, 0x0305);
}

```

There is a duplication in the test for testing the `current` version.
I can modify it, but this is independent from this PR: we have a case where we want to test the current version...

lbudai:
@bazsi : I think all the questions have been addressed.

kira-syslogng:
Build SUCCESS

PR #3372, opened 2020-07-31
Implement template syntax in the topic parameter for kafka destination

vivinperis:
The librdkafka based Kafka implementation doesn't support template syntax in the topic parameter. This patch implements the same.
The flow is as follows
Case 1 - Topic name isn't a template.
Here, we check if the topic name is valid. If the topic name isn't valid, we return a run time error. Here, we don't need fallback topic even though the grammar allows for the same since its not possible for a message publish failure due to a VALID topic name to be fixed by another valid topic name.

Case 2 - Topic name is a template.
Here, we first check if the topic name is in hashmap. If it isn't present, we check  If it is valid, then we insert the valid topic name into hash map. In case topic name isn't valid, we use the fallback topic which has to be defined(as the driver would generate run time error)

The validation for topic is based on the logic provided in the official Apache Kafka repository.

### DIFF
--- a/modules/kafka/kafka-dest-driver.c
+++ b/modules/kafka/kafka-dest-driver.c
@@ -36,12 +36,21 @@
  */
 
 void
-kafka_dd_set_topic(LogDriver *d, const gchar *topic)
+kafka_dd_set_topic(LogDriver *d, LogTemplate *topic)
 {
   KafkaDestDriver *self = (KafkaDestDriver *)d;
 
-  g_free(self->topic_name);
-  self->topic_name = g_strdup(topic);
+  log_template_unref(self->topic_name);
+  self->topic_name = topic;
+}
+
+void
+kafka_dd_set_fallback_topic(LogDriver *d, const gchar *fallback_topic)
+{
+  KafkaDestDriver *self = (KafkaDestDriver *)d;
+
+  g_free(self->fallback_topic_name);
+  self->fallback_topic_name = g_strdup(fallback_topic);
 }
 
 void
@@ -111,6 +120,12 @@ kafka_dd_get_template_options(LogDriver *d)
   return &self->template_options;
 }
 
+gboolean
+kafka_dd_is_topic_name_a_template(KafkaDestDriver *self)
+{
+  return self->topic == NULL;
+}
+
 /* methods */
 
 static const gchar *
@@ -119,7 +134,7 @@ _format_stats_instance(LogThreadedDestDriver *d)
   KafkaDestDriver *self = (KafkaDestDriver *)d;
   static gchar stats_name[1024];
 
-  g_snprintf(stats_name, sizeof(stats_name), "kafka,%s", self->topic_name);
+  g_snprintf(stats_name, sizeof(stats_name), "kafka,%s", self->topic_name->template);
   return stats_name;
 }
 
@@ -132,7 +147,7 @@ _format_persist_name(const LogPipe *d)
   if (d->persist_name)
     g_snprintf(persist_name, sizeof(persist_name), "kafka.%s", d->persist_name);
   else
-    g_snprintf(persist_name, sizeof(persist_name), "kafka(%s)", self->topic_name);
+    g_snprintf(persist_name, sizeof(persist_name), "kafka(%s)", self->topic_name->template);
   return persist_name;
 }
 
@@ -142,6 +157,98 @@ _kafka_log_callback(const rd_kafka_t *rkt, int level, const char *fac, const cha
   gchar *buf = g_strdup_printf("librdkafka: %s(%d): %s", fac, level, msg);
   msg_event_send(msg_event_create(level, buf, NULL));
   g_free(buf);
+}
+
+
+static gboolean
+_contains_valid_pattern(const gchar *name)
+{
+  const gchar *p;
+  for (p = name; *p; p++)
+    {
+      if (!((*p >= 'a' && *p <= 'z') ||
+            (*p >= 'A' && *p <= 'Z') ||
+            (*p >= '0' && *p <= '9') ||
+            (*p == '_') || (*p == '-') || (*p == '.')))
+        {
+          return FALSE;
+        }
+    }
+  return TRUE;
+}
+
+static gboolean
+_validate_topic_name(KafkaDestDriver *self, const gchar *name)
+{
+  gint len = strlen(name);
+  if (len == 0)
+    {
+      msg_error("kafka: topic name is illegal, it can't be empty",
+                evt_tag_str("driver", self->super.super.super.id),
+                log_pipe_location_tag(&self->super.super.super.super));
+      return FALSE;
+    }
+
+  if ((!g_strcmp0(name, ".")) || !g_strcmp0(name, ".."))
+    {
+      msg_error("kafka: topic name cannot be . or ..", evt_tag_str("topic", name),
+                evt_tag_str("driver", self->super.super.super.id),
+                log_pipe_location_tag(&self->super.super.super.super));
+      return FALSE;
+    }
+
+  if (len > 249)
+    {
+      msg_error("kafka: topic name cannot be longer than 249 characters", evt_tag_str("topic", name),
+                evt_tag_int("length", len),
+                evt_tag_str("driver", self->super.super.super.id),
+                log_pipe_location_tag(&self->super.super.super.super));
+      return FALSE;
+    }
+
+  if (!_contains_valid_pattern(name))
+    {
+      msg_error("kafka: topic name is illegal as it contains characters other than pattern [\\-\\._a-zA-Z0-9]+",
+                evt_tag_str("topic", name),
+                evt_tag_str("driver", self->super.super.super.id),
+                log_pipe_location_tag(&self->super.super.super.super));
+      return FALSE;
+    }
+
+  return TRUE;
+}
+
+rd_kafka_topic_t *
+_construct_topic(KafkaDestDriver *self, const gchar *name)
+{
+  g_assert(self->kafka != NULL);
+  if (_validate_topic_name(self, name))
+    {
+      return rd_kafka_topic_new(self->kafka, name, NULL);
+    }
+  return NULL;
+}
+
+rd_kafka_topic_t *
+kafka_dd_query_insert_topic(KafkaDestDriver *self, const gchar *name)
+{
+  g_mutex_lock(&self->topics_lock);
+  rd_kafka_topic_t *topic = g_hash_table_lookup(self->topics, name);
+
+  if (topic)
+    {
+      g_mutex_unlock(&self->topics_lock);
+      return topic;
+    }
+
+  topic = _construct_topic(self, name);
+  if (topic)
+    {
+      g_hash_table_insert(self->topics, g_strdup(name), topic);
+    }
+
+  g_mutex_unlock(&self->topics_lock);
+  return topic;
 }
 
 static void
@@ -163,8 +270,8 @@ _kafka_delivery_report_cb(rd_kafka_t *rk,
       LogPathOptions path_options = LOG_PATH_OPTIONS_INIT;
 
       msg_debug("kafka: delivery report for message came back with an error, putting it back to our queue",
-                self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
-                evt_tag_str("topic", self->topic_name),
+                evt_tag_str("topic", self->topic_name->template),
+                evt_tag_str("fallback_topic", self->fallback_topic_name),
                 evt_tag_printf("message", "%.*s", (int) MIN(len, 128), (char *) payload),
                 evt_tag_str("error", rd_kafka_err2str(err)),
                 evt_tag_str("driver", self->super.super.super.id),
@@ -174,8 +281,8 @@ _kafka_delivery_report_cb(rd_kafka_t *rk,
   else
     {
       msg_debug("kafka: delivery report successful",
-                self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
-                evt_tag_str("topic", self->topic_name),
+                evt_tag_str("topic", self->topic_name->template),
+                evt_tag_str("fallback_topic", self->fallback_topic_name),
                 evt_tag_printf("message", "%.*s", (int) MIN(len, 128), (char *) payload),
                 evt_tag_str("error", rd_kafka_err2str(err)),
                 evt_tag_str("driver", self->super.super.super.id),
@@ -238,21 +345,12 @@ _construct_client(KafkaDestDriver *self)
   if (!client)
     {
       msg_error("kafka: error constructing the kafka connection object",
-                self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
-                evt_tag_str("topic", self->topic_name),
+                evt_tag_str("topic", self->topic_name->template),
                 evt_tag_str("error", errbuf),
                 evt_tag_str("driver", self->super.super.super.id),
                 log_pipe_location_tag(&self->super.super.super.super));
     }
   return client;
-}
-
-rd_kafka_topic_t *
-_construct_topic(KafkaDestDriver *self)
-{
-  g_assert(self->kafka != NULL);
-
-  return rd_kafka_topic_new(self->kafka, self->topic_name, NULL);
 }
 
 static LogThreadedDestWorker *
@@ -280,8 +378,9 @@ _flush_inflight_messages(KafkaDestDriver *self)
   if (outq_len > 0)
     {
       msg_notice("kafka: shutting down kafka producer, while messages are still in-flight, waiting for messages to flush",
-                 self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
-                 evt_tag_str("topic", self->topic_name),
+
+                 evt_tag_str("topic", self->topic_name->template),
+                 evt_tag_str("fallback_topic", self->fallback_topic_name),
                  evt_tag_int("outq_len", outq_len),
                  evt_tag_int("timeout", timeout),
                  evt_tag_str("driver", self->super.super.super.id),
@@ -291,8 +390,8 @@ _flush_inflight_messages(KafkaDestDriver *self)
   if (err != RD_KAFKA_RESP_ERR_NO_ERROR)
     {
       msg_error("kafka: error flushing accumulated messages during shutdown, rd_kafka_flush() returned failure, this might indicate that some in-flight messages are lost",
-                self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
-                evt_tag_str("topic", self->topic_name),
+                evt_tag_str("topic", self->topic_name->template),
+                evt_tag_str("fallback_topic", self->fallback_topic_name),
                 evt_tag_int("outq_len", rd_kafka_outq_len(self->kafka)),
                 evt_tag_str("error", rd_kafka_err2str(err)),
                 evt_tag_str("driver", self->super.super.super.id),
@@ -330,6 +429,43 @@ _purge_remaining_messages(KafkaDestDriver *self)
 }
 
 static gboolean
+_init_template_topic_name(KafkaDestDriver *self)
+{
+  msg_debug("kafka: The topic name is a template",
+            evt_tag_str("topic", self->topic_name->template),
+            evt_tag_str("driver", self->super.super.super.id),
+            log_pipe_location_tag(&self->super.super.super.super));
+
+  if (!self->fallback_topic_name)
+    {
+      msg_error("kafka: fallback_topic() required when the topic name is a template",
+                evt_tag_str("driver", self->super.super.super.id),
+                log_pipe_location_tag(&self->super.super.super.super));
+      return FALSE;
+    }
+
+  self->fallback_topic = _construct_topic(self, self->fallback_topic_name);
+
+  if (self->fallback_topic == NULL)
+    {
+      msg_error("kafka: error constructing the fallback topic object",
+                evt_tag_str("fallback_topic", self->fallback_topic_name),
+                evt_tag_str("driver", self->super.super.super.id),
+                log_pipe_location_tag(&self->super.super.super.super));
+      return FALSE;
+    }
+
+  self->topics = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, (GDestroyNotify) rd_kafka_topic_destroy);
+  return TRUE;
+}
+
+static gboolean
+_topic_name_is_a_template(KafkaDestDriver *self)
+{
+  return (strchr(self->topic_name->template, '$') != NULL);
+}
+
+static gboolean
 kafka_dd_init(LogPipe *s)
 {
   KafkaDestDriver *self = (KafkaDestDriver *)s;
@@ -345,41 +481,39 @@ kafka_dd_init(LogPipe *s)
                 log_pipe_location_tag(&self->super.super.super.super));
       return FALSE;
     }
-  if (strchr(self->topic_name, '$') != NULL)
-    {
-      self->topicname_is_a_template = TRUE;
-      self->topic_hash = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, (GDestroyNotify) rd_kafka_topic_destroy);
-      self->temp_topic_name = log_template_new(configuration, NULL);
-      log_template_compile(self->temp_topic_name, self->topic_name, NULL);
-    }
 
   self->kafka = _construct_client(self);
   if (self->kafka == NULL)
     {
       msg_error("kafka: error constructing kafka connection object, perhaps metadata.broker.list property is missing?",
-                self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
-                evt_tag_str("topic", self->topic_name),
+                evt_tag_str("topic", self->topic_name->template),
                 evt_tag_str("driver", self->super.super.super.id),
                 log_pipe_location_tag(&self->super.super.super.super));
       return FALSE;
     }
-  if(!self->topicname_is_a_template)
+
+  if (_topic_name_is_a_template(self))
     {
-      self->topic = _construct_topic(self);
+      if (!_init_template_topic_name(self))
+        {
+          return FALSE;
+        }
+    }
+
+  else
+    {
+      self->topic = _construct_topic(self, self->topic_name->template);
+
       if (self->topic == NULL)
         {
           msg_error("kafka: error constructing the kafka topic object",
-                    self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
-                    evt_tag_str("topic", self->topic_name),
+                    evt_tag_str("topic", self->topic_name->template),
                     evt_tag_str("driver", self->super.super.super.id),
                     log_pipe_location_tag(&self->super.super.super.super));
           return FALSE;
         }
     }
-  else
-    {
-      g_assert(self->kafka != NULL);
-    }
+
   if (self->message == NULL)
     {
       self->message = log_template_new(cfg, NULL);
@@ -388,8 +522,8 @@ kafka_dd_init(LogPipe *s)
 
   log_template_options_init(&self->template_options, cfg);
   msg_verbose("kafka: Kafka destination initialized",
-              self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
-              evt_tag_str("topic", self->topic_name),
+              evt_tag_str("topic", self->topic_name->template),
+              evt_tag_str("fallback_topic", self->fallback_topic_name),
               evt_tag_str("key", self->key ? self->key->template : "NULL"),
               evt_tag_str("message", self->message->template),
               evt_tag_str("driver", self->super.super.super.id),
@@ -414,18 +548,20 @@ kafka_dd_free(LogPipe *d)
   KafkaDestDriver *self = (KafkaDestDriver *)d;
 
   log_template_options_destroy(&self->template_options);
-  if(self->topic_hash)
-    g_hash_table_unref(self->topic_hash);
+  if (self->topics)
+    g_hash_table_unref(self->topics);
   log_template_unref(self->key);
   log_template_unref(self->message);
-  log_template_unref(self->temp_topic_name);
   if (self->topic)
     rd_kafka_topic_destroy(self->topic);
+  if (self->fallback_topic)
+    rd_kafka_topic_destroy(self->fallback_topic);
+  if (self->fallback_topic_name)
+    g_free(self->fallback_topic_name);
   if (self->kafka)
     rd_kafka_destroy(self->kafka);
-  if (self->topic_name)
-    g_free(self->topic_name);
-  g_mutex_free(&self->lock);
+  log_template_unref(self->topic_name);
+  g_mutex_free(&self->topics_lock);
   g_free(self->bootstrap_servers);
   kafka_property_list_free(self->config);
   log_threaded_dest_driver_free(d);
--- a/modules/kafka/kafka-dest-driver.c
+++ b/modules/kafka/kafka-dest-driver.c
@@ -163,6 +163,7 @@ _kafka_delivery_report_cb(rd_kafka_t *rk,
       LogPathOptions path_options = LOG_PATH_OPTIONS_INIT;
 
       msg_debug("kafka: delivery report for message came back with an error, putting it back to our queue",
+                self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
                 evt_tag_str("topic", self->topic_name),
                 evt_tag_printf("message", "%.*s", (int) MIN(len, 128), (char *) payload),
                 evt_tag_str("error", rd_kafka_err2str(err)),
@@ -173,6 +174,7 @@ _kafka_delivery_report_cb(rd_kafka_t *rk,
   else
     {
       msg_debug("kafka: delivery report successful",
+                self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
                 evt_tag_str("topic", self->topic_name),
                 evt_tag_printf("message", "%.*s", (int) MIN(len, 128), (char *) payload),
                 evt_tag_str("error", rd_kafka_err2str(err)),
@@ -236,6 +238,7 @@ _construct_client(KafkaDestDriver *self)
   if (!client)
     {
       msg_error("kafka: error constructing the kafka connection object",
+                self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
                 evt_tag_str("topic", self->topic_name),
                 evt_tag_str("error", errbuf),
                 evt_tag_str("driver", self->super.super.super.id),
@@ -277,6 +280,7 @@ _flush_inflight_messages(KafkaDestDriver *self)
   if (outq_len > 0)
     {
       msg_notice("kafka: shutting down kafka producer, while messages are still in-flight, waiting for messages to flush",
+                 self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
                  evt_tag_str("topic", self->topic_name),
                  evt_tag_int("outq_len", outq_len),
                  evt_tag_int("timeout", timeout),
@@ -287,6 +291,7 @@ _flush_inflight_messages(KafkaDestDriver *self)
   if (err != RD_KAFKA_RESP_ERR_NO_ERROR)
     {
       msg_error("kafka: error flushing accumulated messages during shutdown, rd_kafka_flush() returned failure, this might indicate that some in-flight messages are lost",
+                self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
                 evt_tag_str("topic", self->topic_name),
                 evt_tag_int("outq_len", rd_kafka_outq_len(self->kafka)),
                 evt_tag_str("error", rd_kafka_err2str(err)),
@@ -343,29 +348,38 @@ kafka_dd_init(LogPipe *s)
   if (strchr(self->topic_name, '$') != NULL)
     {
       self->topicname_is_a_template = TRUE;
-      self->topic_hash = g_hash_table_new(g_str_hash, g_str_equal);
+      self->topic_hash = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, (GDestroyNotify) rd_kafka_topic_destroy);
+      self->temp_topic_name = log_template_new(configuration, NULL);
+      log_template_compile(self->temp_topic_name, self->topic_name, NULL);
     }
 
   self->kafka = _construct_client(self);
   if (self->kafka == NULL)
     {
       msg_error("kafka: error constructing kafka connection object, perhaps metadata.broker.list property is missing?",
+                self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
                 evt_tag_str("topic", self->topic_name),
                 evt_tag_str("driver", self->super.super.super.id),
                 log_pipe_location_tag(&self->super.super.super.super));
       return FALSE;
     }
-
-  self->topic = _construct_topic(self);
-  if (self->topic == NULL)
+  if(!self->topicname_is_a_template)
     {
-      msg_error("kafka: error constructing the kafka topic object",
-                evt_tag_str("topic", self->topic_name),
-                evt_tag_str("driver", self->super.super.super.id),
-                log_pipe_location_tag(&self->super.super.super.super));
-      return FALSE;
+      self->topic = _construct_topic(self);
+      if (self->topic == NULL)
+        {
+          msg_error("kafka: error constructing the kafka topic object",
+                    self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
+                    evt_tag_str("topic", self->topic_name),
+                    evt_tag_str("driver", self->super.super.super.id),
+                    log_pipe_location_tag(&self->super.super.super.super));
+          return FALSE;
+        }
     }
-
+  else
+    {
+      g_assert(self->kafka != NULL);
+    }
   if (self->message == NULL)
     {
       self->message = log_template_new(cfg, NULL);
@@ -374,6 +388,7 @@ kafka_dd_init(LogPipe *s)
 
   log_template_options_init(&self->template_options, cfg);
   msg_verbose("kafka: Kafka destination initialized",
+              self->topicname_is_a_template ? evt_tag_str("template", self->topic_name):
               evt_tag_str("topic", self->topic_name),
               evt_tag_str("key", self->key ? self->key->template : "NULL"),
               evt_tag_str("message", self->message->template),
@@ -399,15 +414,18 @@ kafka_dd_free(LogPipe *d)
   KafkaDestDriver *self = (KafkaDestDriver *)d;
 
   log_template_options_destroy(&self->template_options);
-
+  if(self->topic_hash)
+    g_hash_table_unref(self->topic_hash);
   log_template_unref(self->key);
   log_template_unref(self->message);
+  log_template_unref(self->temp_topic_name);
   if (self->topic)
     rd_kafka_topic_destroy(self->topic);
   if (self->kafka)
     rd_kafka_destroy(self->kafka);
   if (self->topic_name)
     g_free(self->topic_name);
+  g_mutex_free(&self->lock);
   g_free(self->bootstrap_servers);
   kafka_property_list_free(self->config);
   log_threaded_dest_driver_free(d);
--- a/modules/kafka/kafka-dest-driver.c
+++ b/modules/kafka/kafka-dest-driver.c
@@ -125,7 +125,6 @@ kafka_dd_is_topic_name_a_template(KafkaDestDriver *self)
 {
   return self->topic == NULL;
 }
-
 /* methods */
 
 static const gchar *
@@ -177,41 +176,46 @@ _contains_valid_pattern(const gchar *name)
   return TRUE;
 }
 
-static gboolean
-_validate_topic_name(KafkaDestDriver *self, const gchar *name)
+GQuark
+topic_name_error_quark(void)
+{
+  return g_quark_from_static_string("invalid-topic-name-error-quark");
+}
+
+gboolean
+kafka_dd_validate_topic_name(const gchar *name, GError **error)
 {
   gint len = strlen(name);
+
   if (len == 0)
     {
-      msg_error("kafka: topic name is illegal, it can't be empty",
-                evt_tag_str("driver", self->super.super.super.id),
-                log_pipe_location_tag(&self->super.super.super.super));
+      g_set_error(error, TOPIC_NAME_ERROR, TOPIC_LENGTH_ZERO,
+                  "kafka: topic name is illegal, it can't be empty");
+
       return FALSE;
     }
 
   if ((!g_strcmp0(name, ".")) || !g_strcmp0(name, ".."))
     {
-      msg_error("kafka: topic name cannot be . or ..", evt_tag_str("topic", name),
-                evt_tag_str("driver", self->super.super.super.id),
-                log_pipe_location_tag(&self->super.super.super.super));
+      g_set_error(error, TOPIC_NAME_ERROR, TOPIC_DOT_TWO_DOTS,
+                  "kafka: topic name cannot be . or ..");
+
       return FALSE;
     }
 
   if (len > 249)
     {
-      msg_error("kafka: topic name cannot be longer than 249 characters", evt_tag_str("topic", name),
-                evt_tag_int("length", len),
-                evt_tag_str("driver", self->super.super.super.id),
-                log_pipe_location_tag(&self->super.super.super.super));
+      g_set_error(error, TOPIC_NAME_ERROR, TOPIC_EXCEEDS_MAX_LENGTH,
+                  "kafka: topic name cannot be longer than 249 characters");
+
       return FALSE;
     }
 
   if (!_contains_valid_pattern(name))
     {
-      msg_error("kafka: topic name is illegal as it contains characters other than pattern [\\-\\._a-zA-Z0-9]+",
-                evt_tag_str("topic", name),
-                evt_tag_str("driver", self->super.super.super.id),
-                log_pipe_location_tag(&self->super.super.super.super));
+      g_set_error(error, TOPIC_NAME_ERROR, TOPIC_INVALID_PATTERN,
+                  "kafka: topic name %s is illegal as it contains characters other than pattern [-._a-zA-Z0-9]+", name);
+
       return FALSE;
     }
 
@@ -222,32 +226,43 @@ rd_kafka_topic_t *
 _construct_topic(KafkaDestDriver *self, const gchar *name)
 {
   g_assert(self->kafka != NULL);
-  if (_validate_topic_name(self, name))
+
+  GError *error = NULL;
+
+  if (kafka_dd_validate_topic_name(name, &error))
     {
       return rd_kafka_topic_new(self->kafka, name, NULL);
     }
+
+  msg_error("Error constructing topic", evt_tag_str("topic_name", name),
+            evt_tag_str("driver", self->super.super.super.id),
+            log_pipe_location_tag(&self->super.super.super.super),
+            evt_tag_str("error message", error->message));
+  g_error_free(error);
+
   return NULL;
 }
 
 rd_kafka_topic_t *
 kafka_dd_query_insert_topic(KafkaDestDriver *self, const gchar *name)
 {
-  g_mutex_lock(&self->topics_lock);
+  g_mutex_lock(self->topics_lock);
   rd_kafka_topic_t *topic = g_hash_table_lookup(self->topics, name);
 
   if (topic)
     {
-      g_mutex_unlock(&self->topics_lock);
+      g_mutex_unlock(self->topics_lock);
       return topic;
     }
 
   topic = _construct_topic(self, name);
+
   if (topic)
     {
       g_hash_table_insert(self->topics, g_strdup(name), topic);
     }
 
-  g_mutex_unlock(&self->topics_lock);
+  g_mutex_unlock(self->topics_lock);
   return topic;
 }
 
@@ -444,9 +459,9 @@ _init_template_topic_name(KafkaDestDriver *self)
       return FALSE;
     }
 
-  self->fallback_topic = _construct_topic(self, self->fallback_topic_name);
+  rd_kafka_topic_t *fallback_topic = _construct_topic(self, self->fallback_topic_name);
 
-  if (self->fallback_topic == NULL)
+  if (fallback_topic == NULL)
     {
       msg_error("kafka: error constructing the fallback topic object",
                 evt_tag_str("fallback_topic", self->fallback_topic_name),
@@ -456,6 +471,8 @@ _init_template_topic_name(KafkaDestDriver *self)
     }
 
   self->topics = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, (GDestroyNotify) rd_kafka_topic_destroy);
+  g_hash_table_insert(self->topics, g_strdup(self->fallback_topic_name), fallback_topic);
+
   return TRUE;
 }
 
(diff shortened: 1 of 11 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

alltilla:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-driver.c
```diff
@@ -236,20 +272,69 @@ _construct_client(KafkaDestDriver *self)
...
+      return FALSE;
+    }
+  if((!g_strcmp0(name, ".")) || !g_strcmp0(name, ".."))
+    {
+      msg_error("kafka: topic name cannot be . or ..", evt_tag_str("topic", name));
+      return FALSE;
+    }
+  if(len > 249)
+    {
+      msg_error("kafka: topic name cannot be longer than 249 characters", evt_tag_str("topic", name), evt_tag_printf("length",
```
alltilla, line 316: `evt_tag_int("length", len)` can be used here.

modules/kafka/kafka-dest-driver.c
```diff
@@ -236,20 +272,69 @@ _construct_client(KafkaDestDriver *self)
...
+                "%d", len));
+      return FALSE;
+    }
+  if(!_contains_valid_pattern(name))
+    {
+      msg_error("kafka: topic name is illegal as it contains characters other than pattern [\\-_a-zA-Z0-9]+",
```
alltilla, line 322: `.` is also accepted by `_contains_valid_pattern()`, please add it to the regexp here, or remove it from the check.

modules/kafka/kafka-dest-driver.c
```diff
@@ -144,6 +153,31 @@ _kafka_log_callback(const rd_kafka_t *rkt, int level, const char *fac, const cha
...
+
+  if(topic)
+    {
+      g_mutex_unlock(&self->lock);
+      return topic;
+    }
+
+  if(kafka_dd_validate_topic_name(name))
+    {
+      topic = rd_kafka_topic_new(self->kafka, g_strdup(name), NULL);
```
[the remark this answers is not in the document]
alltilla: Maybe this would be better, so there is no code duplication of calling `kafka_dd_validate_topic_name()`, which is already done in `_construct_topic()`:
```
...
topic = _construct_topic(self, name);
if (topic)
  g_hash_table_insert(self->topic_hash_map, g_strdup(name), topic);

g_mutex_unlock(&self->lock);
return topic;
}
```

modules/kafka/kafka-dest-driver.c
```diff
@@ -277,7 +362,9 @@ _flush_inflight_messages(KafkaDestDriver *self)
   if (outq_len > 0)
     {
       msg_notice("kafka: shutting down kafka producer, while messages are still in-flight, waiting for messages to flush",
-                 evt_tag_str("topic", self->topic_name),
+
+                 evt_tag_str("topic", self->topic_name->template),
+                 self->fallback_topic_name ? evt_tag_str("fallback_topic", self->fallback_topic_name) : NULL,
```
alltilla, line 367: I think the `NULL` here stops the processing of all the `evt_tags` after. Can you check that please? Maybe `evt_tag_str("fallback_topic", self->fallback_topic_name ? : "N/A")` would be better.

Same thing with the other log message of this function.

modules/kafka/kafka-dest-driver.c
```diff
@@ -324,6 +412,32 @@ _purge_remaining_messages(KafkaDestDriver *self)
 
 }
 
+static gboolean
+_init_template_topic_name(KafkaDestDriver *self)
+{
+  self->topicname_is_a_template = TRUE;
+  msg_debug("kafka: The topic name is a template");
```
alltilla, line 419: This debug log is either not needed, or should be extended with `log_pipe_location_tag()`, the driver and the template itself.

modules/kafka/kafka-dest-driver.c
```diff
@@ -324,6 +412,32 @@ _purge_remaining_messages(KafkaDestDriver *self)
...
+
+  if(!self->fallback_topic_name)
+    {
+      msg_error("kafka: fallback_topic() required when the topic name is a template");
```
alltilla, line 423: Please add `log_pipe_location_tag()` and the driver tag.

modules/kafka/kafka-dest-driver.c
```diff
@@ -345,20 +459,33 @@ kafka_dd_init(LogPipe *s)
...
-                evt_tag_str("topic", self->topic_name),
+                evt_tag_str("topic", self->topic_name->template),
                 evt_tag_str("driver", self->super.super.super.id),
                 log_pipe_location_tag(&self->super.super.super.super));
       return FALSE;
     }
 
-  self->topic = _construct_topic(self);
-  if (self->topic == NULL)
+  if (strchr(self->topic_name->template, '$') != NULL)
```
alltilla, line 468: Please move this check to a separate function called something like `_topic_name_is_a_template(self)`.

modules/kafka/kafka-dest-driver.c
```diff
@@ -345,20 +459,33 @@ kafka_dd_init(LogPipe *s)
...
+    }
+
+  else
+    {
+      self->topic = _construct_topic(self, self->topic_name->template);
+
+      if (self->topic == NULL)
+        {
+          msg_error("kafka: error constructing the kafka topic object",
+                    self->fallback_topic_name ? evt_tag_str("fallback_topic", self->fallback_topic_name) : NULL,
```
alltilla, line 483: fallback topic is not interesting here.

modules/kafka/kafka-dest-driver.h
```diff
@@ -35,27 +35,33 @@ typedef struct
   LogTemplateOptions template_options;
   LogTemplate *key;
   LogTemplate *message;
+  LogTemplate *topic_name;
+  GHashTable *topic_hash_map;
```
alltilla, line 39: Optional: IMO a better name would be simply `topics`.

modules/kafka/kafka-dest-driver.c
```diff
@@ -324,6 +412,32 @@ _purge_remaining_messages(KafkaDestDriver *self)
 
 }
 
+static gboolean
+_init_template_topic_name(KafkaDestDriver *self)
+{
+  self->topicname_is_a_template = TRUE;
```
alltilla, line 418: I think this variable is not needed at all, and its logic could be replaced with something like:

```
gboolean
kafka_dd_is_topic_name_a_template(KafkaDestDriver *self)
{
  return self->topic == NULL;
}
```

modules/kafka/kafka-dest-worker.c
```diff
@@ -61,16 +62,38 @@ _publish_message(KafkaDestWorker *self, LogMessage *msg)
 {
   KafkaDestDriver *owner = (KafkaDestDriver *) self->super.owner;
   int block_flag = _is_poller_thread(self) ? 0 : RD_KAFKA_MSG_F_BLOCK;
+  rd_kafka_topic_t *topic;
+  //GString *topic_name_new = g_string_sized_new(128);
```
alltilla, line 66: Please remove the comment.

modules/kafka/kafka-dest-worker.c
```diff
@@ -61,16 +62,38 @@ _publish_message(KafkaDestWorker *self, LogMessage *msg)
...
+  GString *topic_name_new = scratch_buffers_alloc();
```
alltilla, line 67: Don't we need to use a `Marker` here?

alltilla, line 67: I think there could be a separate function to get the necessary topic name, instead of the `if-else`s in this function. From this code it is hard to determine what is used in the end and why.

modules/kafka/kafka-dest-driver.h
```diff
@@ -35,27 +35,33 @@ typedef struct
...
+  GMutex lock;
```
alltilla, line 40: This could be `topics_lock` or something like that, to better show its intention.

modules/kafka/kafka-dest-driver.c
```diff
@@ -333,7 +447,7 @@ kafka_dd_init(LogPipe *s)
   if (!log_threaded_dest_driver_init_method(s))
     return FALSE;
 
-  if (!self->topic_name)
+  if (!self->topic_name->template)
```
alltilla, line 450: This will cause a SEGFAULT if `topic_name` is `NULL` (`topic()` is not set).

modules/kafka/kafka-dest-worker.c
alltilla: Or another alternative would be a reusable GString in the worker, like topic_name_buffer.

modules/kafka/kafka-dest-driver.c
```diff
@@ -163,7 +197,8 @@ _kafka_delivery_report_cb(rd_kafka_t *rk,
       LogPathOptions path_options = LOG_PATH_OPTIONS_INIT;
 
       msg_debug("kafka: delivery report for message came back with an error, putting it back to our queue",
-                evt_tag_str("topic", self->topic_name),
+                evt_tag_str("topic", self->topic_name->template),
```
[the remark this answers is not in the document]
vivinperis: Can you explain me the debug_flag part?

modules/kafka/kafka-dest-driver.c
```diff
@@ -236,20 +272,69 @@ _construct_client(KafkaDestDriver *self)
...
+  return TRUE;
+}
+
+gboolean
+kafka_dd_validate_topic_name(const gchar *name)
+{
+  gint len = strlen(name);
+  if(len == 0)
+    {
+      msg_error("kafka: topic name is illegal, it can't be empty", evt_tag_str("topic", name));
```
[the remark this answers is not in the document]
vivinperis: So I need to pass on KafkaDestDriver as a parameter argument right?

modules/kafka/kafka-dest-worker.c
vivinperis: Can you clarify on the reusable GString part?

modules/kafka/kafka-dest-driver.c
vivinperis: Yes, tested it now. It is ignoring the evt_tag after NULL
@bazsi, In the earlier PR comments, you had told me about using NULL instead of "NULL" string. 
How do I do it the right way(other than keeping fallback_topic at the end)?

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-worker.c
vivinperis: Given that there are multiple variables inside the function which get updated, how do I do this?

modules/kafka/kafka-dest-driver.c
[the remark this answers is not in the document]
vivinperis: I have a small concern. In case we evaluate the template inside the delivery report call back function, can the latency affect the topic name displayed here vs the actual topic where the message was published to?

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-worker.c
alltilla: Every worker could have a `GString *topic_name_buffer`, and the topic template could be resolved to there in each `_publish_message()` call. Every data stored in the `Worker` is safe and won't be accessed concurrently in the `publish_message()` function.

alltilla: There is no need for the `is_fallback_topic` variable, you can just write a warning when it is used.
Something like this:

```
static void
_calculate_topic_name_from_template(KafkaDestWorker *self, LogMessage *msg, GString *topic_name)
{
  // Throw a warning message if fallback topic is used
  // Use g_string_assign()
  ...
}

static void
_get_literal_topic_name(KafkaDestWorker *self, LogMessage *msg, GString *topic_name)
{
  // Use g_string_assign()
  ...
}

static void
_calculate_topic_name(KafkaDestWorker *self, LogMessage *msg, GString *topic_name)
{
  if (kafka_dd_is_topic_name_a_template(owner))
    _calculate_topic_name_from_template(self, msg, topic_name);

  _get_literal_topic_name(self, msg, topic_name);
}

static gboolean
_publish_message(KafkaDestWorker *self, LogMessage *msg)
{
...
  GString *topic_name = scratch_buffers_alloc_and_mark(&marker);
  _calculate_topic_name(self, msg, topic_name);

  if (rd_kafka_produce(topic,
                       RD_KAFKA_PARTITION_UA,
                       RD_KAFKA_MSG_F_FREE | block_flag,
                       self->message->str, self->message->len,
                       self->key->len ? self->key->str : NULL, self->key->len,
                       log_msg_ref(msg)) == -1)
    {
      msg_error("kafka: failed to publish message",
                evt_tag_str("topic", topic_name->str),
                evt_tag_str("error", rd_kafka_err2str(rd_kafka_last_error())),
                evt_tag_str("driver", owner->super.super.super.id),
                log_pipe_location_tag(&owner->super.super.super.super));

      scratch_buffers_reclaim_marked(marker);
      return FALSE;
    }

  msg_debug("kafka: message published",
            evt_tag_str("topic", topic_name->str),
            evt_tag_str("key", self->key->len ? self->key->str : "NULL"),
            evt_tag_str("message", self->message->str),
            evt_tag_str("driver", owner->super.super.super.id),
            log_pipe_location_tag(&owner->super.super.super.super));

  /* we passed the allocated buffers to rdkafka, which will eventually free them */
  g_string_steal(self->message);
  scratch_buffers_reclaim_marked(marker);
  return TRUE;
}
```

If the `topic_name_buffer` is implemented, you will no longer have to pass a `GString *` to `_calculate_topic_name()`, but you can use the worker's buffer.

modules/kafka/kafka-dest-driver.c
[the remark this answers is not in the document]
alltilla: Most of the cases it won't be a problem, because the resolution is only defined by the message. If however there is a template function, which uses something else besides the message (like random generation, or system time, etc...) the resolution will not be the same, indeed.

modules/kafka/kafka-dest-driver.c
bazsi: just include it unconditionally. the value would become "(null)" or something like this if it's null.

modules/kafka/kafka-dest-worker.c
```diff
@@ -61,32 +62,57 @@ _publish_message(KafkaDestWorker *self, LogMessage *msg)
...
+  gboolean is_fallback_topic = FALSE;
 
-  if (rd_kafka_produce(owner->topic,
+  if (kafka_dd_is_topic_name_a_template(owner))
+    {
+      log_template_format(owner->topic_name, msg, &owner->template_options, LTZ_SEND, self->super.seq_num, NULL,
+                          topic_name_new);
+      topic = kafka_dd_query_insert_topic(owner, topic_name_new->str);
+
+      if (!topic) //if topic name is invalid, then topic is fallback topic which is we know is defined
```
bazsi, line 76: please include the comment inside the if, we tend to use traditional `/* */` comments

modules/kafka/kafka-dest-worker.c
```diff
@@ -61,32 +62,57 @@ _publish_message(KafkaDestWorker *self, LogMessage *msg)
...
+        {
+          topic = owner->fallback_topic;
+          is_fallback_topic = TRUE;
+        }
+    }
+
+  else //Case where topic name isn't a template. Since we have validated the topic name at driver, it is good to eliminate the need for fallback topic
```
bazsi, line 83: likewise.

modules/kafka/kafka-dest-worker.c
```diff
@@ -61,32 +62,57 @@ _publish_message(KafkaDestWorker *self, LogMessage *msg)
 {
   KafkaDestDriver *owner = (KafkaDestDriver *) self->super.owner;
   int block_flag = _is_poller_thread(self) ? 0 : RD_KAFKA_MSG_F_BLOCK;
+  rd_kafka_topic_t *topic;
+  ScratchBuffersMarker marker;
+  GString *topic_name_new = scratch_buffers_alloc_and_mark(&marker);
```
bazsi, line 67: if you are not doing lots of allocations in this (and child) functions, you can just get away by using scratch_buffers_alloc() (and not using reclaim at exit points), as the thread function in LogThreadedDestination would call the garbage collector after every call to this function:

https://github.com/syslog-ng/syslog-ng/blob/edfe167e3f76cb419c84bea09ca3ca42d6da84f4/lib/logthrdest/logthrdestdrv.c#L415-L418

modules/kafka/kafka-dest-driver.c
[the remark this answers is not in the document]
bazsi: Although I agree it would be useful, reformatting it again does not feel right in my opinion. If we wanted to include per-message information like this we would start thinking about using an msg_opaque that is not pointing directly to LogMessage, rather a structure that encapsulates both the LogMessage and any information we might need in the delivery callback (e.g. topic name).

I don't think the topic name would include anything that is not part of the formatted log message. So I would wait until this becomes an issue :)

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-worker.c
vivinperis: So basically a global variable inside worker?

vivinperis: Also, I wont be able to use scratch buffer in that case right?

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-worker.c
```diff
@@ -56,21 +58,63 @@ _format_message_and_key(KafkaDestWorker *self, LogMessage *msg)
...
+
+  return topic;
+}
+
+static rd_kafka_topic_t *
+_get_literal_topic(KafkaDestWorker *self, LogMessage *msg)
+{
+  KafkaDestDriver *owner = (KafkaDestDriver *) self->super.owner;
+
+  g_string_assign(self->topic_name_buffer, owner->topic_name->template);
```
alltilla, line 84: It is not needed to fill the `topic_name_buffer`, if the `rd_kafka_topic_t` is already created.

modules/kafka/kafka-dest-worker.c
```diff
@@ -56,21 +58,63 @@ _format_message_and_key(KafkaDestWorker *self, LogMessage *msg)
...
+  KafkaDestDriver *owner = (KafkaDestDriver *) self->super.owner;
+  log_template_format(owner->topic_name, msg, &owner->template_options, LTZ_SEND, self->super.seq_num, NULL,
+                      self->topic_name_buffer);
+  rd_kafka_topic_t *topic = kafka_dd_query_insert_topic(owner, self->topic_name_buffer->str);
+
+  if (!topic)
+    {
+      /* if topic name is invalid, then topic is fallback topic which we know is defined*/
+      topic = owner->fallback_topic;
+      g_string_assign(self->topic_name_buffer, owner->fallback_topic_name);
```
alltilla, line 73: It is not needed to fill the `topic_name_buffer`, if the fallback `rd_kafka_topic_t` is already created.

modules/kafka/kafka-dest-worker.c
```diff
@@ -56,21 +58,63 @@ _format_message_and_key(KafkaDestWorker *self, LogMessage *msg)
...
+  if (rd_kafka_produce(topic,
                        RD_KAFKA_PARTITION_UA,
                        RD_KAFKA_MSG_F_FREE | block_flag,
                        self->message->str, self->message->len,
                        self->key->len ? self->key->str : NULL, self->key->len,
                        log_msg_ref(msg)) == -1)
     {
       msg_error("kafka: failed to publish message",
-                evt_tag_str("topic", owner->topic_name),
+                evt_tag_str("topic", self->topic_name_buffer->str),
```
alltilla, line 117: There is a getter function in `rdkafka`, which returns the `topic_name` of the `rd_kafka_topic_t` object.
That can handle the templated topic, the literal topic and the fallback topic, too, in just one function call.

https://docs.confluent.io/2.0.0/clients/librdkafka/rdkafka_8h.html#a3ea8248521e63b4afb1d292fbcc584e2

modules/kafka/kafka-dest-worker.c
```diff
@@ -79,7 +123,7 @@ _publish_message(KafkaDestWorker *self, LogMessage *msg)
     }
 
   msg_debug("kafka: message published",
-            evt_tag_str("topic", owner->topic_name),
+            evt_tag_str("topic", self->topic_name_buffer->str),
```
alltilla, line 126: You can use the getter here, too.

modules/kafka/kafka-dest-driver.c
```diff
@@ -144,6 +160,99 @@ _kafka_log_callback(const rd_kafka_t *rkt, int level, const char *fac, const cha
...
+            (*p >= 'A' && *p <= 'Z') ||
+            (*p >= '0' && *p <= '9') ||
+            (*p == '_') || (*p == '-') || (*p == '.')))
+        {
+          return FALSE;
+        }
+    }
+  return TRUE;
+}
+
```
alltilla, line 180: Please remove this additional empty line.

modules/kafka/kafka-dest-driver.c
```diff
@@ -111,6 +120,13 @@ kafka_dd_get_template_options(LogDriver *d)
   return &self->template_options;
 }
 
+gboolean
+kafka_dd_is_topic_name_a_template(KafkaDestDriver *self)
+{
+  return self->topic == NULL;
+}
+
```
alltilla, line 128: Please remove this additional empty line.

modules/kafka/kafka-dest-worker.c
```diff
@@ -56,21 +58,63 @@ _format_message_and_key(KafkaDestWorker *self, LogMessage *msg)
...
+_calculate_topic_from_template(KafkaDestWorker *self, LogMessage *msg)
+{
+  KafkaDestDriver *owner = (KafkaDestDriver *) self->super.owner;
+  log_template_format(owner->topic_name, msg, &owner->template_options, LTZ_SEND, self->super.seq_num, NULL,
+                      self->topic_name_buffer);
+  rd_kafka_topic_t *topic = kafka_dd_query_insert_topic(owner, self->topic_name_buffer->str);
+
+  if (!topic)
+    {
+      /* if topic name is invalid, then topic is fallback topic which we know is defined*/
```
alltilla, line 71: It is better to make sure, we do not have any programming errors, and `g_assert()` to the fallback topic here.
That would also make this comment obsolete.

modules/kafka/kafka-grammar.ym
```diff
@@ -82,11 +83,14 @@ kafka_options
...
 kafka_option
-        : KW_TOPIC '(' string ')'
+        : KW_TOPIC '(' template_content ')'
         {
             kafka_dd_set_topic(last_driver, $3);
-            free($3);
         }
+        | KW_FALLBACK_TOPIC '(' string ')'
+        {
+            kafka_dd_set_fallback_topic(last_driver, $3);
```
alltilla, line 92: A `free()` is missing from here.

modules/kafka/kafka-dest-worker.c
```diff
@@ -56,21 +58,63 @@ _format_message_and_key(KafkaDestWorker *self, LogMessage *msg)
...
+  KafkaDestDriver *owner = (KafkaDestDriver *) self->super.owner;
+
+  if (kafka_dd_is_topic_name_a_template(owner))
+    {
+      return _calculate_topic_from_template(self, msg);
+    }
+
+  return _get_literal_topic(self, msg);
+}
+
```
alltilla, line 100: Please remove the additional empty line here.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-driver.c
```diff
@@ -324,6 +428,47 @@ _purge_remaining_messages(KafkaDestDriver *self)
...
+    }
+
+  self->topics = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, (GDestroyNotify) rd_kafka_topic_destroy);
+  return TRUE;
+}
+
+static gboolean
+_topic_name_is_a_template(KafkaDestDriver *self)
+{
+  if (strchr(self->topic_name->template, '$'))
```
alltilla, line 465: `return (strchr(self->topic_name->template, '$') != NULL);` should suffice :)

modules/kafka/kafka-dest-worker.c
```diff
@@ -24,6 +24,7 @@
  */
 #include "kafka-dest-worker.h"
 #include "kafka-dest-driver.h"
+#include "scratch-buffers.h"
```
alltilla, line 27: You don't need `scratch-buffers.h` anymore.

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-driver.c
```diff
@@ -173,7 +281,8 @@ _kafka_delivery_report_cb(rd_kafka_t *rk,
   else
     {
       msg_debug("kafka: delivery report successful",
-                evt_tag_str("topic", self->topic_name),
+                evt_tag_str("topic", self->topic_name->template),
+                evt_tag_str("fallback_topic", self->fallback_topic_name ? self->fallback_topic_name : "NULL"),
```
bazsi, line 285: evt_tag_str() handles NULLs, so the check can be avoided (everywhere).

vivinperis: Done

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

alltilla:
I reviewed the unit tests, and found quite some things that should changed.

On high level:
* There are a lot of code, that are copy-pasted from other tests, and not needed here.
* `valgrind` shows a lot of leaks. You should take care of the object you created in the test cases. (`valgrind --leak-check=full --trace-children=yes modules/kafka/tests/test_kafka_topic`)
* Global variables should generally be avoided.
* You are including `.c` files, and calling static functions, which is really mixed up. The general way to call static functions is to make them not static, and put them in an `...-internal.h` header, which is only used by the unit test.
* There are some unnecessary tests.
* The criterion framework is not used in its full potential, `cr_assert(something == FALSE)` should be `cr_assert_not(something)` and `Parametrized` tests should be used in some cases.

So we talked about unit test on gitter. The proper way to unit test any function, is to create its inputs, and assert its outputs for the different inputs.
You only need to test 2 functions: `_validate_topic_name()` and `_calculate_topic()`.

`_validate_topic_name()` needs `KafkaDestDriver *self, const gchar *name`, but `self` is only used to log errors, so it does not need to be initialized, or anything, `kafka_dd_new(configuration)` is enough, and free it with the `LogThreadedDest` API after use.

`_calculate_topic()` needs `KafkaDestWorker *self, LogMessage *msg`, `msg` is created and filled in the unit test, `self` is a bit more complicated.
Your implementation of `_calculate_topic()` does multiple things, which shows itself in the unit test:
1. `calculates the topic name string (either from the template or the fallback topic)
2. MAYBE creates an `rd_kafka_topic_t` object, and MAYBE puts it in the HashMap
3. returns an `rd_kafka_topic_t` object from either the HashMap, `self->topic` or `self->fallback_topic` (maybe fallback_topic should be in the HashMap, too)

It can be seen, that to correctly create the inputs for `_calculate_topic()`, you need to create and init `KafkaDestWorker`. Sometimes this can be okay, but in this case splitting the function to smaller parts is probably the way to go.
You cannot test this many things at once in one unit test. I believe `_calculate_topic()` needs to be refactored and renamed to clearly show its steps, and all of its steps should be tested in separate unit tests.

Unit tests are (among other things) useful to show us if a function is not following the single responsibility principle, so it is good, that you wrote tests.

Please think about these arguments, and have another run with the tests. I might be wrong somewhere, if I missed something or you do not agree with me, please say so.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -0,0 +1,182 @@
...
+
+  return cr_make_param_array(struct topic_validity_test_params, params, sizeof(params)/sizeof(params[0]));
+}
+
+ParameterizedTest(struct topic_validity_test_params *param, kafka_topic, topic_validity_tests)
+{
+  KafkaDestDriver *driver = (KafkaDestDriver *) kafka_dd_new(configuration);
+  cr_assert_eq(kafka_dd_validate_topic_name(driver, param->topic_name), param->isvalid);
+  log_pipe_unref((LogPipe *)driver);
+}
```
bazsi, line 82: this looks good (from the allocation/free standpoint).

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -0,0 +1,182 @@
...
+
+  cr_assert_str_eq(_resolve_template_topic_name(worker, msg), "fallbacktopicname");
+  cr_assert_str_eq(_resolve_template_topic_name(worker, msg), rd_kafka_topic_name(_calculate_topic(worker, msg)));
+  log_msg_unref(msg);
+
+  log_template_unref(topic_name);
+  log_threaded_dest_worker_free(&worker->super);
+  log_pipe_unref((LogPipe *)driver);
+  cfg_free(configuration);
+}
```
bazsi, line 116: I am wondering how the return value of _calculate_topic() is managed. Is it an allocated variable? Who frees that if it's allocated?

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -0,0 +1,182 @@
...
+  log_msg_set_value_by_name(msg, "MSG", "invalid name", -1);
+
+  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic_from_template(worker, msg)), "fallbackhere");
+  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic_from_template(worker, msg)),
+                   rd_kafka_topic_name(_calculate_topic(worker, msg)));
+  log_msg_unref(msg);
+
+  log_template_unref(topic_name);
+  log_threaded_dest_worker_free(&worker->super);
+  log_threaded_dest_driver_free((LogPipe *)driver);
```
bazsi, line 150: the log_threaded_dest_driver_free() is not needed here and is taken care by the subsequent log_pipe_unref()

bazsi: the worker needs to be freed though as it is not managed by the KafkaDestDriver instance.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -0,0 +1,182 @@
...
+
+  LogMessage *msg = log_msg_new_empty();
+  cr_assert_str_eq(rd_kafka_topic_name(_get_literal_topic(worker)), "topicname");
+  cr_assert_str_eq(rd_kafka_topic_name(_get_literal_topic(worker)),
+                   rd_kafka_topic_name(_calculate_topic(worker, msg)));
+
+  log_msg_unref(msg);
+  log_template_unref(topic_name);
+  log_threaded_dest_worker_free(&worker->super);
+  log_threaded_dest_driver_free((LogPipe *)driver);
```
bazsi, line 179: again not needed.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -0,0 +1,182 @@
...
+
+  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic_from_template(worker, msg)), "fallbackhere");
+  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic_from_template(worker, msg)),
+                   rd_kafka_topic_name(_calculate_topic(worker, msg)));
+  log_msg_unref(msg);
+
+  log_template_unref(topic_name);
+  log_threaded_dest_worker_free(&worker->super);
+  log_threaded_dest_driver_free((LogPipe *)driver);
+  log_pipe_deinit((LogPipe *)driver);
```
bazsi, line 151: Right now, this happens after you free the driver a line above, so that's a use-after-free. First deinit the instance and then we can go and free it.

Free should happen via log_pipe_unref()

So this should become:

```
  log_pipe_deinit(&driver->super);
  log_pipe_unref(&driver->super);
```

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -0,0 +1,182 @@
...
+  LogMessage *msg = log_msg_new_empty();
+  cr_assert_str_eq(rd_kafka_topic_name(_get_literal_topic(worker)), "topicname");
+  cr_assert_str_eq(rd_kafka_topic_name(_get_literal_topic(worker)),
+                   rd_kafka_topic_name(_calculate_topic(worker, msg)));
+
+  log_msg_unref(msg);
+  log_template_unref(topic_name);
+  log_threaded_dest_worker_free(&worker->super);
+  log_threaded_dest_driver_free((LogPipe *)driver);
+  log_pipe_deinit((LogPipe *)driver);
```
bazsi, line 180: unref is missing

kira-syslogng:
Build SUCCESS

modules/kafka/tests/test_kafka_topic.c
vivinperis: The return type of ```_calculate_topic()``` which is a topic is freed at the end when the hashmap is freed. Incase it is non template, it is free at the end when we free the topic.
```
if (self->topic)
    rd_kafka_topic_destroy(self->topic);
```
```
  if (self->topics)
    g_hash_table_unref(self->topics);
```

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-driver.c
```diff
@@ -341,24 +487,39 @@ kafka_dd_init(LogPipe *s)
...
+      if (!_init_template_topic_name(self))
+        {
+          return FALSE;
+        }
+    }
+
+  else
+    {
+      self->topic = _construct_topic(self, self->topic_name->template);
+
```
bazsi, line 514: It would be great if you could extract the else branch as well, that would make reading this if easier, e.g.

```
if (is_topic_name_a_template(...) && !_init_template_topic_name(self) ||
   _init_literal_topic_name(self))
   return FALSE;
```

modules/kafka/kafka-dest-worker-internal.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef KAFKA_TEST_INTERNAL_H_INCLUDED
+#define KAFKA_TEST_INTERNAL_H_INCLUDED
+
+#include "logthrdest/logthrdestdrv.h"
+#include <librdkafka/rdkafka.h>
+#include "kafka-dest-worker.h"
+
+
+
+#define STRING_250_LEN "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" \
```
bazsi, line 34: this should probably be moved to the test program. (this and the next define as well).

modules/kafka/kafka-dest-driver.c
bazsi: forget the if example I gave above. That should be something like this instead:

if (_is_topic_name_a_template(self))
  {
    if (!_init_template_topic_name(self))
      return FALSE;
  }
else
  {
    if (!_init_literal_topic_name(self))
      return FALSE;
  }

bazsi: Ah, I think I found the perfect version, we need an _init_topic_name() function, that encapsulates the if, and that can be simpler as we don't need to continue with the rest of the kafka_dd_init() function.

```
static gboolean
_init_topic_name(KafkaDestDriver *self)
{
  if (_is_topic_name_a_template(self))
    return _init_template_topic_name(self);
  else
    return_init_literal_topic_name(self);
}

static gboolean
kafka_dd_init(...)
{
   ...
   if (!_init_topic_name(self))
    return FALSE;
   ...
}

```

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -0,0 +1,202 @@
...
+    {STRING_249_LEN, TRUE},
+  };
+
+  return cr_make_param_array(struct topic_validity_test_params, params, sizeof(params)/sizeof(params[0]));
+}
+
+ParameterizedTest(struct topic_validity_test_params *param, kafka_topic, topic_validity_tests)
+{
+  KafkaDestDriver *driver = (KafkaDestDriver *) kafka_dd_new(configuration);
+  cr_assert_eq(kafka_dd_validate_topic_name(driver, param->topic_name), param->isvalid);
```
bazsi, line 94: just a nitpick, but you don't need to change it: if the kafka_dd_validate_topic_name() would return the error in a GError object, https://developer.gnome.org/glib/stable/glib-Error-Reporting.html, then you could move all the error messages currently inside this function to its caller. And with that, you could completely remove the "self" argument, making it
static (as C++ calls this).

With it static, this testcase would be much simpler, you wouldn't need to allocate/free an entire KafkaDestDriver instance just to test a string validation function.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -0,0 +1,202 @@
...
+
+  g_assert(kafka_dd_init((LogPipe *) driver));
+
+  KafkaDestDriver *kafka_driver = (KafkaDestDriver *) driver;
+
+  KafkaDestWorker *worker = (KafkaDestWorker *) kafka_dest_worker_new(&kafka_driver->super, 0);
+
+  LogMessage *msg = log_msg_new_empty();
+  log_msg_set_value_by_name(msg, "MSG", "valid_template_topic", -1);
+  cr_assert_str_eq(_resolve_template_topic_name(worker, msg), "valid_template_topic");
```
bazsi, line 117: maybe it's just me, but this would read much better if the function would use a KafkaDestWorker specific prefix (e.g. kafka_dest_worker_resolve_template_topic_name(), that would clearly indicate it is the function being tested. The names starting with an underscore is usually a private function not intended to be called from the outside.

The unit tests for the KafkaDestDriver are a bit clumsy, as we lack the internal layers so we could test these functions. However, the internal layering would be too much of boilerplate for not much gain.

With all this said, I think if you removed these functions to better indicate that they are part of the class being tested, this would be easier to understand.

modules/kafka/kafka-dest-driver.h
```diff
@@ -59,9 +58,12 @@ void kafka_dd_set_message_ref(LogDriver *d, LogTemplate *message);
 void kafka_dd_set_flush_timeout_on_shutdown(LogDriver *d, gint shutdown_timeout);
 void kafka_dd_set_flush_timeout_on_reload(LogDriver *d, gint reload_timeout);
 void kafka_dd_set_poll_timeout(LogDriver *d, gint poll_timeout);
+
+gboolean kafka_dd_validate_topic_name(KafkaDestDriver *self, const gchar *name);
 gboolean kafka_dd_is_topic_name_a_template(KafkaDestDriver *self);
 rd_kafka_topic_t *kafka_dd_query_insert_topic(KafkaDestDriver *self, const gchar *name);
 LogTemplateOptions *kafka_dd_get_template_options(LogDriver *d);
+gboolean kafka_dd_init(LogPipe *s);
```
alltilla, line 66: This should be in the internal header IMO. If you move it to the internal header, rename it to `kafka-internal.h`

modules/kafka/kafka-dest-worker-internal.h
```diff
@@ -0,0 +1,51 @@
...
+                       "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" \
+                       "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" \
+                       "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+
+gchar *_resolve_template_topic_name(KafkaDestWorker *self, LogMessage *msg);
+rd_kafka_topic_t *_calculate_topic_from_template(KafkaDestWorker *self, LogMessage *msg);
+rd_kafka_topic_t *_get_literal_topic(KafkaDestWorker *self);
+rd_kafka_topic_t *_calculate_topic(KafkaDestWorker *self, LogMessage *msg);
+
+#endif
```
alltilla, line 51: New line is missing.

modules/kafka/kafka-dest-worker-internal.h
```diff
@@ -0,0 +1,51 @@
...
+
+#ifndef KAFKA_TEST_INTERNAL_H_INCLUDED
+#define KAFKA_TEST_INTERNAL_H_INCLUDED
+
+#include "logthrdest/logthrdestdrv.h"
+#include <librdkafka/rdkafka.h>
+#include "kafka-dest-worker.h"
+
+
+
```
alltilla, line 33: Unnecessary empty lines.

modules/kafka/kafka-dest-worker.c
```diff
@@ -57,32 +47,43 @@ _format_message_and_key(KafkaDestWorker *self, LogMessage *msg)
                         self->super.seq_num, NULL, self->key);
 }
 
-static rd_kafka_topic_t *
-_calculate_topic_from_template(KafkaDestWorker *self, LogMessage *msg)
+gchar *
```
alltilla, line 50: Please make this const, to show, the string is not needed to be freed by the caller.

modules/kafka/kafka-dest-worker.c
```diff
@@ -57,32 +47,43 @@ _format_message_and_key(KafkaDestWorker *self, LogMessage *msg)
...
+_calculate_topic_from_template(KafkaDestWorker *self, LogMessage *msg)
+{
+  KafkaDestDriver *owner = (KafkaDestDriver *) self->super.owner;
+  rd_kafka_topic_t *topic = kafka_dd_query_insert_topic(owner, _resolve_template_topic_name(self, msg));
 
   if (!topic)
     {
-      g_assert(owner->fallback_topic != NULL);
-      topic = owner->fallback_topic;
+      g_assert_not_reached();
```
alltilla, line 72: Simply use `g_assert(topic);` instead of the whole `if`.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -22,157 +22,181 @@
...
+    {"valid-name123", TRUE},
+    {"valid_name123", TRUE},
+    {"valid.name-123", TRUE},
+    {"valid_name-123", TRUE},
+    {"This.is.a.valid.name", TRUE},
+    {"This.is-a_valid-name", TRUE},
+    {STRING_249_LEN, TRUE},
+  };
+
+  return cr_make_param_array(struct topic_validity_test_params, params, sizeof(params)/sizeof(params[0]));
```
alltilla, line 88: You can use `G_N_ELEMENTS(params)`.

modules/kafka/tests/test_kafka_topic.c
alltilla: I really like this idea.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -22,157 +22,181 @@
...
   LogMessage *msg = log_msg_new_empty();
-  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic(worker, msg)), "topicname");
+  log_msg_set_value_by_name(msg, "MSG", "valid_template_topic", -1);
+  cr_assert_str_eq(_resolve_template_topic_name(worker, msg), "valid_template_topic");
+  cr_assert_str_eq(_resolve_template_topic_name(worker, msg), rd_kafka_topic_name(_calculate_topic(worker, msg)));
+
+  log_msg_set_value_by_name(msg, "MSG", "invalid name", -1);
+
+  cr_assert_str_eq(_resolve_template_topic_name(worker, msg), "fallbacktopicname");
+  cr_assert_str_eq(_resolve_template_topic_name(worker, msg), rd_kafka_topic_name(_calculate_topic(worker, msg)));
```
alltilla, line 123: I don't see the point of having both of these asserts (same for the other asserts in this function):
```
  cr_assert_str_eq(_resolve_template_topic_name(worker, msg), "fallbacktopicname");
  cr_assert_str_eq(_resolve_template_topic_name(worker, msg), rd_kafka_topic_name(_calculate_topic(worker, msg)));
```
`_resolve_template_topic_name()` is called by `_calculate_topic()`, I don't understand the necessity of comparing them.
Which from these 2 functions do you want to test here?

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -22,157 +22,181 @@
...
+
+  g_assert(kafka_dd_init((LogPipe *) driver));
+
+  KafkaDestDriver *kafka_driver = (KafkaDestDriver *) driver;
+
+  g_assert_nonnull(g_hash_table_lookup(kafka_driver->topics, "fallbackhere"));
+
+  KafkaDestWorker *worker = (KafkaDestWorker *) kafka_dest_worker_new(&kafka_driver->super, 0);
+
+  g_assert_null(g_hash_table_lookup(kafka_driver->topics, "validtopic"));
```
alltilla, line 151: Why do you need the HashTable asserts? The HashTable IMO is an implementation detail, if someone would like to refactor it, these checks would prevent it.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -22,157 +22,181 @@
...
 
-  cr_assert_str_eq(kafka_driver->topic_name->template, "$MSG");
-  worker = (KafkaDestWorker *) kafka_dest_worker_new(&kafka_driver->super, 0);
   LogMessage *msg = log_msg_new_empty();
+  log_msg_set_value_by_name(msg, "MSG", "validtopic", -1);
+  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic_from_template(worker, msg)), "validtopic");
+  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic_from_template(worker, msg)),
```
alltilla, line 156: Again, I don't understand the point of comparing `_calculate_topic_from_template()` and `_calculate_topic()` as one calls another. Either only use `_calculate_topic_from_template()` here, or remove this and create a test which tests `_calculate_topic()` and include both the literal and templated cases there. I prefer the latter.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -22,157 +22,181 @@
...
-  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic(worker, msg)), "fallbacktopicname");
-  log_msg_set_value_by_name(msg, "MSG", "", -1);
-  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic(worker, msg)), "fallbacktopicname");
+  g_assert_null(kafka_driver->topics);
+
+  KafkaDestWorker *worker = (KafkaDestWorker *) kafka_dest_worker_new(&kafka_driver->super, 0);
+
+  LogMessage *msg = log_msg_new_empty();
+  cr_assert_str_eq(rd_kafka_topic_name(_get_literal_topic(worker)), "topicname");
+  cr_assert_str_eq(rd_kafka_topic_name(_get_literal_topic(worker)),
```
alltilla, line 194: Same as in `test_calculate_topic_from_template`.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -22,157 +22,181 @@
...
+  configuration = cfg_new_snippet();
+  LogDriver *driver = kafka_dd_new(configuration);
+
+  LogTemplate *topic_name = log_template_new(configuration, NULL);
+
+  log_template_compile(topic_name, "$MSG", NULL);
+  kafka_dd_set_topic(driver, topic_name);
+  kafka_dd_set_fallback_topic(driver, "fallbackhere");
+
+  g_assert(kafka_dd_init((LogPipe *) driver));
```
alltilla, line 143: Use `cr_assert()` instead of `g_assert()` in the test files.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -22,157 +22,181 @@
...
 
-  cr_assert_str_eq(kafka_driver->topic_name->template, "$MSG");
-  worker = (KafkaDestWorker *) kafka_dest_worker_new(&kafka_driver->super, 0);
-  LogMessage *msg = log_msg_new_empty();
+  LogTemplate *topic_name = log_template_new(configuration, NULL);
+
+  log_template_compile(topic_name, "topicname", NULL);
+  kafka_dd_set_topic(driver, topic_name);
+
+  g_assert(kafka_dd_init((LogPipe *) driver));
```
alltilla, line 184: `cr_assert()`.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -22,157 +22,181 @@
...
+
+  KafkaDestDriver *kafka_driver = (KafkaDestDriver *) driver;
 
-  log_msg_set_value_by_name(msg, "MSG", "invalid topic", -1);
-  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic(worker, msg)), "fallbacktopicname");
-  log_msg_set_value_by_name(msg, "MSG", "..", -1);
-  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic(worker, msg)), "fallbacktopicname");
-  log_msg_set_value_by_name(msg, "MSG", "", -1);
-  cr_assert_str_eq(rd_kafka_topic_name(_calculate_topic(worker, msg)), "fallbacktopicname");
+  g_assert_null(kafka_driver->topics);
```
alltilla, line 188: This is implementation detail, also you should not use `g_assert()`. Use `cr_assert_not(_topic_name_is_a_template(kafka_driver));`

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -22,157 +22,181 @@
...
-  kafka_driver->kafka = _construct_client(kafka_driver);
-  _init_template_topic_name(kafka_driver);
+  configuration = cfg_new_snippet();
+  LogDriver *driver = kafka_dd_new(configuration);
+
+  LogTemplate *topic_name = log_template_new(configuration, NULL);
+
+  log_template_compile(topic_name, "$MSG", NULL);
+  kafka_dd_set_topic(driver, topic_name);
+  kafka_dd_set_fallback_topic(driver, "fallbackhere");
```
alltilla, line 141: These 4 lines are used multiple times in this file, it might be worth to factor them out to a helper function.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -22,157 +22,181 @@
...
 {
-  _init_valid_template_topic_names();
-  kafka_driver->kafka = _construct_client(kafka_driver);
-  _init_template_topic_name(kafka_driver);
+  configuration = cfg_new_snippet();
+  LogDriver *driver = kafka_dd_new(configuration);
+
+  LogTemplate *topic_name = log_template_new(configuration, NULL);
+
+  log_template_compile(topic_name, "$MSG", NULL);
```
alltilla, line 139: While `$MSG` works, it is a bad suggestion. `$MSG` will never be a valid topic name.
Maybe some imaginary macro like `$kafka_topic` would be better.

modules/kafka/kafka-dest-worker-internal.h
```diff
@@ -0,0 +1,51 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#ifndef KAFKA_TEST_INTERNAL_H_INCLUDED
+#define KAFKA_TEST_INTERNAL_H_INCLUDED
```
alltilla, line 26: Nitpicking: We usually use the filename here, with `-` and `.` replaced by `_`, and all uppercase.

modules/kafka/tests/test_kafka_topic.c
vivinperis: It was primarily to test both the functions. Should I write a separate test for just ```_calculate_topic()``` instead?

kira-syslogng:
Build SUCCESS

modules/kafka/tests/test_kafka_topic.c
alltilla: That sounds good!

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-driver.c
```diff
@@ -223,11 +228,19 @@ _construct_topic(KafkaDestDriver *self, const gchar *name)
...
+  if (kafka_dd_validate_topic_name(name, &error))
     {
       return rd_kafka_topic_new(self->kafka, name, NULL);
     }
 
+  msg_error("Error constructing topic", evt_tag_str("topic_name", name),
+            evt_tag_str("driver", self->super.super.super.id),
+            log_pipe_location_tag(&self->super.super.super.super),
+            evt_tag_str("message", error->message));
```
alltilla, line 241: `"message"` is not the best name for it.

modules/kafka/kafka-dest-worker.c
```diff
@@ -54,7 +54,7 @@ kafka_dest_worker_resolve_template_topic_name(KafkaDestWorker *self, LogMessage
   log_template_format(owner->topic_name, msg, &owner->template_options, LTZ_SEND, self->super.seq_num, NULL,
                       self->topic_name_buffer);
 
-  if (kafka_dd_validate_topic_name(owner, self->topic_name_buffer->str))
+  if (kafka_dd_validate_topic_name(self->topic_name_buffer->str, NULL))
```
alltilla, line 57: You lost the error logs here.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -45,12 +45,19 @@
                        "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" \
                        "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
 
-struct topic_validity_test_params
+struct valid_topic_test_params
 {
   gchar *topic_name;
   gboolean isvalid;
```
alltilla, line 51: This is unnecessary now.

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -45,12 +45,19 @@
...
 };
 
+struct invalid_topic_test_params
+{
+  gchar *topic_name;
+  gboolean isvalid;
```
alltilla, line 57: This is unnecessary now.

modules/kafka/kafka-dest-worker-internal.h
alltilla: This is not fixed, don't mark it resolved.

modules/kafka/tests/test_kafka_topic.c
alltilla: This is not fixed, don't mark it resolved.

modules/kafka/kafka-dest-worker-internal.h
vivinperis: Why new line though? I will update it

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-worker-internal.h
alltilla: Try calling cat on the file with and without newline.

You might not agree with the review comments some are surely arguable, but not fixing them, neither writing your point, and still marking them as resolved is not cool...

vivinperis: Sorry, I missed this. Realized while reading the email notification. Will, go through the review comments once more to make sure I haven't missed anything.

vivinperis: I hope its fine now?

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

alltilla (approved):

bazsi (approved):
Thank you all the changes. Approved.

kira-syslogng:
Build SUCCESS

PR #3376, opened 2020-08-05
stats-query: speedup

furiel:
This  patchset speeds up `syslog-ng-ctl query get "*"` command. When the stats counters were collected into a list, they were appended instead of prepend and reverse in the end. The former is quadratic in complexity, the latter is linear. This made `query` impractical to be used with large number of counters.

PS: I experimented with `stats` too, but me seemed changing `num` changes the execution time linearly. So I believe there is no such problem with stats.

```
@version: 3.29

options {
 stats-level(3);
};

log {
  source { example-msg-generator(freq(0) num(50000)
};
```

Before:
```
$ time ./syslog-ng-ctl query get "*" | wc -l
194637

real	0m59.636s
user	0m0.003s
sys	0m0.019s
```
After:

```
$ time ./syslog-ng-ctl query get "*" | wc -l
200057

real	0m0.160s
user	0m0.011s
sys	0m0.014s
```

In the same scenario, stats is a bit slower, but not significantly.

```
$ time ./syslog-ng-ctl stats | wc -l
200058

real	0m0.356s
user	0m0.015s
sys	0m0.008s
```

### DIFF
--- a/lib/stats/stats-query.c
+++ b/lib/stats/stats-query.c
@@ -167,7 +167,7 @@ _query_counter_hash(const gchar *key_str)
       if (_is_pattern_matches_key(pattern, key))
         {
           StatsCounterItem *counter = (StatsCounterItem *) value;
-          counters = g_list_append(counters, counter);
+          counters = g_list_prepend(counters, counter);
 
           if (single_match)
             break;
@@ -176,7 +176,7 @@ _query_counter_hash(const gchar *key_str)
   g_static_mutex_unlock(&stats_query_mutex);
 
   g_pattern_spec_free(pattern);
-  return counters;
+  return g_list_reverse(counters);
 }
 
 static GList *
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

lib/stats/stats-query.c
```diff
@@ -243,7 +243,7 @@ _get_views(const gchar *filter)
   g_static_mutex_unlock(&stats_query_mutex);
 
   g_pattern_spec_free(pattern);
-  return views;
+  return g_list_reverse(views);
```
Kokan, line 246: The order is unspecified, as such reversing it is not really required here.

furiel: Nice finding! I was thinking about this, and I would like to leave this as it is.

I agree the order is unspecified: it depends on the internal representation of the hashtables containing the counters. While the ordering in unspecified, it is deterministic. Removing this would reverse the output that users got used to. This is not a strong argument, and I am fine to discuss this separately. Personally, I do not really like that the output is not ordered. But for now, I  would like to maintain the original behavior.

Kokan: I would say that even a "keeping the ordering" is not true; as it is far being deterministic. Due to multi threading and the fact that the order of inserting a key has an effect on this listing. (sure with the above constructed syntactical config it is). As such user could not get used to an order... and should not either way.

Sure the gain with the lack of list reverse is way smaller, and the PR does improve speed as is, so I could approve.

(I would rather do `g_list_shuffle` instead of ordering it, just to make sure nobody relies on any ordering, see how much appending costs... as at the end of the day as a user I could do `syslog-ng-ctl stats|sort` that costs nothing to syslog-ng.)

PR #3379, opened 2020-08-06
Ack tracker refactors

lbudai:
All the changes are equivalent changes.

The motivation behind this is 
 * to make it clear:
   * early ack tracker is actually not 'early' - it's just processes an ACK immediately and has an impact on flow control at the receiving time of the ACK
   * early ack tracker _could_ save positions... currently it's not, but it could (and my plan is it will - my next PR is about implementing this)
   * late ack tracker is waiting for a sequence of consecutive ACKs and calls the save method on the last `Bookmark`
 * to enable for each `LogProtoServer` to define its own ACK strategy - as protos fill the bookmarks, they know how to handle ACKs

Naming maybe not the best... I'm open to discuss it.
And I felt some part of the code that this refactor is a step back... I found position tracking more readable than playing with `AckTrackerType` but finally I convinced myself that this is the correct way.

Why now? We need to implement a new `AckTracker` - `BatchedAckTracker`, and that would be extremely hard to understand how does it fit into existing `AckTrackers`. And this renaming work finally ended up in this refactor.

This PR should has an internal counterpart (ALTP)... I'll open it soon.


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/logproto/logproto-buffered-server.c
```diff
@@ -970,5 +961,5 @@ log_proto_buffered_server_init(LogProtoBufferedServer *self, LogTransport *trans
   else
     self->convert = (GIConv) -1;
   self->stream_based = TRUE;
-  self->pos_tracking = options->position_tracking_enabled;
+  self-> pos_tracking = (options->ack_tracker_type != ACK_INSTANT_BOOKMARKLESS);
```
Kokan, line 964: This code does not really help to extend ack trackers, as it is going to bite.
I was thinking about the following alternatives: (in my preference order)

1) The logsource is the one who pulls the strings here, and as such it should decide on which ack tracker should be used. The logproto could keep reporting if it is position tracked or not.

2) A dummy alternative (sure this not really solves the issue just makes it easier to detect if happens)
```c
gboolean _is_position_tracked(AckTrackerType act)
{
   switch (act)
   {
     case ACK_CONSECUTIVE:
        return TRUE;
     case ACK_INSTANT_BOOKMARKLESS:
        return FALSE;
    default:
       g_assert_not_reached();
   }

  g_assert_not_reached();
  return FALSE;
}
```

3) On the other hand it can be 'argued' that actually there is no other bookmarkless ack tracker implementation that is possible. So in the end this cover everything. (no change)

lbudai: @Kokan:
1:
The main difference here from previous versions is that the direction is _inverted_ between the type of `AckTracker` and the `pos_tracking` flag : `AckTrackerType` does not depend on position tracking but position tracking depends on the type of `AckTracker`.

2:
The referred `pos_tracking` is used by protos to decide whether they need to save position or not.
And the code expresses the fact that when ack tracker type is _not_ bookmarkless, then position is saved...

3:
Bookmarkless is special and we won't have another bookmarkless variants.

MrAnno:
We need to adjust internal modules after merging this PR.

PR #3380, opened 2020-08-07
Instant ack tracker with bookmark

lbudai:
This PR implements an instant AckTracker that saves bookmarks.
Where can be used? In sources where we delegate the ACK saving to the originator and batched ack tracking is not supported - in these cases the ordering doesn't matter as it is tracked by the originator.

depends on: #3379


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3381, opened 2020-08-08
add dupnames pcre flag

ctr49:
Allow dupnames flag to be used in PCRE expressions, allowing duplicate names for named subpatterns as explained here https://www.pcre.org/original/doc/html/pcrepattern.html#SEC16

### DIFF
--- a/lib/logmatcher.c
+++ b/lib/logmatcher.c
@@ -333,6 +333,12 @@ _compile_pcre_regexp(LogMatcherPcreRe *self, const gchar *re, GError **error)
           return FALSE;
         }
     }
+  if (self->super.flags & LMF_DUPNAMES)
+    {
+      if (!PCRE_DUPNAMES)
+        msg_warning("syslog-ng was compiled against an old PCRE which doesn't support the 'dupnames' flag");
+      flags |= PCRE_DUPNAMES;
+    }
 
   /* complile the regexp */
   self->pattern = pcre_compile2(re, flags, &rc, &errptr, &erroffset, NULL);
@@ -743,6 +749,7 @@ CfgFlagHandler log_matcher_flag_handlers[] =
   { "substring",       CFH_SET, offsetof(LogMatcherOptions, flags), LMF_SUBSTRING     },
   { "prefix",          CFH_SET, offsetof(LogMatcherOptions, flags), LMF_PREFIX        },
   { "disable-jit",     CFH_SET, offsetof(LogMatcherOptions, flags), LMF_DISABLE_JIT   },
+  { "dupnames",        CFH_SET, offsetof(LogMatcherOptions, flags), LMF_DUPNAMES      },
 
   { NULL },
 };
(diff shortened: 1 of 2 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

bazsi:
Ok to test

bazsi:
A testcase would be great though. There's a unit test that tests regexps and flags so this would only become an extra check there.

kira-syslogng:
Build SUCCESS

szemere:
Hello @ctr49,
 sorry for not answering earlier. I was trying to figure out an issue with the current implementation. (The problem is probably not in your code.)
 While the flag is working, I was unable to retrieve the value of the named capture group.

Example:
```
filter f_filter1 {
  match("(?<FOOBAR>bar)|(?<FOOBAR>foo)" value(MSG) flags(store-matches, dupnames)
  );
};
```

Without the `dupnames` flag I got the proper error message:
```
Error parsing filter expression, compiling the regexp failed, error=Failed to compile PCRE expression >>>(?<FOOBAR>bar)|(?<FOOBAR>foo)<<< `two named subpatterns have the same name' at character 24 in /conf/ng.conf:22:3-22:4:
```

I send in the following messages (using the `dupnames` flag):
```
hello world
hello bar
hello foo
```

Using the following template: `template("Hello $FOOBAR\n")` The output is:
```
Hello bar
Hello 
```

- No `Hello world`, since there is no matching message.
- `Hello bar` as `(?<FOOBAR>bar)` matches, and stores the value in the `$FOOBAR` macro.
- `Hello ` as `(?<FOOBAR>foo)` matches, but the macro is not set.

Note: If I use the numbered macros:
- `$0` works as expected
- `$1` works as above
- `$2` gives the same result but reversed, with `foo`

We are retrieving the matching group names in this function: https://github.com/syslog-ng/syslog-ng/blob/master/lib/logmatcher.c#L412
The `name_table` contains both matches for both messages, however the second name in the regex (I swapped the order for a short test) always got `-1` as an index, so we skip it's value.
Currently I am reading PCRE documentation, but I was unable to figure out the reason behind it.

Best regards,
Laci

Kokan:
@szemere  the issue you found here can be reproduced on master as well. I posted a fix for that https://github.com/syslog-ng/syslog-ng/pull/3393 . Imho this can be merged as is, but UT and news file should be provided.

szemere:
Hello @ctr49 ,
 I compiled your code on top of #3393 and it works perfectly! I think we can go with the manual testing for now, and I will create a "good-first-issue" to review all the regex related test cases.
 If you provide a news file we can merge it, and ship it with 3.29.

Br,
Laci

bazsi:
#3393 has gone in, so this could as well. Code wise this seems to be approved, we'd need a NEWS file.

lbudai:
@ctr49 : thanks the PR!

I opened #3400 with news file so I'm merge this as-is.

lbudai (approved):

PR #3383, opened 2020-08-09
Fixed location reporting and long lines in config errors

bazsi:
This branch fixes a couple of issues in configuration error reporting:

1) when the configuration file has lines longer than 1024 bytes, the report could become confusing
2) the location of the error was incorrectly reported in some cases

+1 I've added the parsed value to the main grammar error message for cases where the value is a string.

All this together would probably have avoided the bug report in #3382

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -903,7 +903,7 @@ template_content_inner
           CHECK_ERROR_GERROR(log_template_compile(last_template, $3, &error), @3, error, "Error compiling template");
           free($3);
 
-          CHECK_ERROR_GERROR(log_template_set_type_hint(last_template, $1, &error), @1, error, "Error setting the template type-hint");
+          CHECK_ERROR_GERROR(log_template_set_type_hint(last_template, $1, &error), @1, error, "Error setting the template type-hint \"%s\"", $1);
           free($1);
         }
         ;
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

lib/cfg-grammar.y
```diff
@@ -903,7 +903,7 @@ template_content_inner
           CHECK_ERROR_GERROR(log_template_compile(last_template, $3, &error), @3, error, "Error compiling template");
           free($3);
 
-          CHECK_ERROR_GERROR(log_template_set_type_hint(last_template, $1, &error), @1, error, "Error setting the template type-hint");
+          CHECK_ERROR_GERROR(log_template_set_type_hint(last_template, $1, &error), @1, error, "Error setting the template type-hint \"%s\"", $1);
```
szemere, line 906: Note: I was concerned, what if we get a NULL or invalid pointer if the string parameter is not provided. Answer: It is handled earlier by the parser as a missing parameter. So we got at least an "" empty string at this point.

szemere, line 906: I like the new error messages. Checked each use case. I think only this one will give a "double" error message, since the `error` string will also contain the original parameter: https://github.com/syslog-ng/syslog-ng/blob/master/lib/type-hinting.c#L70

szemere:
I added the "user-visible-feature" flag, since we are changing the output. However I think this one do not need extra documentation in the admin guide. With a news file, it is approved from my sight.

furiel:
Looks good to me too. @bazsi  please add the news file. Then this can go in.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

PR #3387, opened 2020-08-11
Threaded ctl command runner

lbudai:
Split #3349 

First part of the PR is:
 * run stats/query commands in a separate thread

### DIFF
--- a/lib/control/control-server.c
+++ b/lib/control/control-server.c
@@ -47,6 +47,9 @@ typedef struct _ThreadedCommandRunner
     GCond tid_saved_cond;
     gboolean tid_saved;
     ThreadId tid;
+    GMutex state_lock;
+    gboolean cancelled;
+    gboolean finished;
   } real_thread;
   ControlConnectionCommand func;
   GString *response;
@@ -62,6 +65,7 @@ _thread_command_runner_new(ControlConnection *cc, GString *cmd, gpointer user_da
   self->user_data = user_data;
   g_mutex_init(&self->real_thread.tid_saved_lock);
   g_cond_init(&self->real_thread.tid_saved_cond);
+  g_mutex_init(&self->real_thread.state_lock);
   self->real_thread.tid_saved = FALSE;
   self->real_thread.tid = 0;
 
@@ -73,6 +77,7 @@ _thread_command_runner_free(ThreadedCommandRunner *self)
 {
   g_mutex_clear(&self->real_thread.tid_saved_lock);
   g_cond_clear(&self->real_thread.tid_saved_cond);
+  g_mutex_clear(&self->real_thread.state_lock);
   g_string_free(self->command, TRUE);
   g_free(self);
 }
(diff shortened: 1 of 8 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/control/control-server.c
```diff
@@ -41,11 +166,27 @@ void
...
 }
 
 void
 control_server_free(ControlServer *self)
 {
+  // it's not racy as the iv_main() is executed by this thread
+  // posted events are not executed as iv_quit() is already called (before control_server_free)
+  // but it is possible that some ThreadedCommandRunner::thread are still running
+  if (self->worker_threads)
+    g_list_free_full(self->worker_threads, _delete_thread_command_runner);
```
furiel, line 189: Unfortunately this does not cancel the long running commands. `control_server_free` is called too late: after main loop is not running. For example, syslog-ng-ctl stats is running in mainloop, so main_loop_deinit will not be called until it is finished.

You can try this:

```
@version: 3.29

options {
 stats-level(3);
};

log {
  source { example-msg-generator(freq(0) num(100000000)
};
```

Let syslog-ng to run a bit: example-msg-generator will generate a bit slowly, the longer you wait, the longer stats runs. After waiting a bit, call syslog-ng-ctl stats and syslog-ng-ctl stop at the same time. syslog-ng-ctl-stop will "hang".

lbudai: @furiel : Thanks!!
(It works perfectly with my original branch, when control server is run in a different thread - I'll fix it soon).

kira-syslogng:
Build SUCCESS

lib/control/control-server.c
```diff
@@ -26,10 +26,181 @@
...
+  ThreadedCommandRunner *runner = _thread_command_runner_new(self, command, user_data);
+  _thread_command_runner_run(runner, cmd_cb);
+}
+
+static void
+_delete_thread_command_runner(gpointer data)
+{
+  ThreadedCommandRunner *self = (ThreadedCommandRunner *) data;
+  g_assert(self->real_thread.tid_saved == TRUE);
+  gboolean has_to_free = FALSE;
```
szemere, line 175: discussed IRL: replaced by https://github.com/syslog-ng/syslog-ng/pull/3387/files#diff-c8826ab0571c7b2f08d7f66e83e5c29aR125

lib/control/control-server.c
```diff
@@ -26,10 +26,181 @@
...
+      _thread_command_runner_sync_run(self, func);
+      return;
+    }
+
+  iv_event_register(&self->response_received);
+
+  self->thread = g_thread_new(self->command->str, (GThreadFunc) _thread, self);
+  _thread_command_runner_wait_for_tid_saved(self);
+  ControlServer *server = self->connection->server;
+  server->worker_threads = g_list_append(server->worker_threads, self);
```
szemere, line 159: nitpicking: g_list_prepend would give a better performance

kira-syslogng:
Build SUCCESS

lib/control/control-server.c
lbudai: Unfortunately only my original solution works... the reason is very simple: `stats_lock`...
When generator source (or any other source executed by the main thread) produces messages, it also updates its stat counters.  When the stats mutex is locked by a `syslog-ng-ctl stats` then ivykis will wait the actual 'job' to be finished (which can happen only after the mutex is unlocked) - until that it won't process the events of the control loop...
This is why we need to execute `ControlServer` in its own thread: that won't lock the stats mutex.

furiel (approved):

szemere (approved):

PR #3389, opened 2020-08-14
Batched ack tracker

lbudai:


kira-syslogng:
Build FAILURE

lib/ack-tracker/batched_ack_tracker.c
```diff
@@ -125,6 +174,8 @@ _manage_msg_ack(AckTracker *s, LogMessage *msg, AckType ack_type)
 
   if (ack_type != AT_ABORTED)
     {
+      if (!iv_timer_registered(&self->batch_timer))
+        _start_batch_timer(self);
```
alltilla, line 178: To me this does not feel like a timeout, but a periodic forced ack. If I think about a timeout, I would expect the timer to be restarted, if a batch is acked.

lib/logsource.h
```diff
@@ -29,7 +29,6 @@
 #include "stats/stats-registry.h"
 #include "window-size-counter.h"
 #include "dynamic-window.h"
-#include "ack-tracker/ack_tracker_types.h"
```
alltilla, line 32: For some reason I can't comment on `logproto-server.h`'s affected line... The same include removal can be done there, too.

lib/ack-tracker/batched_ack_tracker.c
```diff
@@ -0,0 +1,261 @@
...
+static void
+_free(AckTracker *s)
+{
+  BatchedAckTracker *self = (BatchedAckTracker *) s;
+  g_mutex_clear(&self->acked_records_lock);
+
+  _stop_batch_timer(self);
+
+  if (self->acked_records)
+    _ack_batch(self, self->acked_records);
```
alltilla, line 207: Is it safe to ack in free? I think the ack callback might want to use some variables of the source, but the source itself is freed earlier:
```
void
log_source_free(LogPipe *s)
{
  LogSource *self = (LogSource *) s;

  g_free(self->name);
  g_free(self->stats_id);
  g_free(self->stats_instance);
  log_pipe_detach_expr_node(&self->super);
  log_pipe_free_method(s);

  ack_tracker_factory_unref(self->ack_tracker_factory);
  ack_tracker_free(self->ack_tracker);
  self->ack_tracker = NULL;

  if (G_UNLIKELY(dynamic_window_is_enabled(&self->dynamic_window)))
    _release_dynamic_window(self);
}
```

lib/ack-tracker/ack_tracker_factory.c
```diff
@@ -0,0 +1,63 @@
...
+#include "consecutive_ack_tracker.h"
+#include "ack_tracker.h"
+
+void
+ack_tracker_factory_init_instance(AckTrackerFactory *self)
+{
+  g_atomic_counter_set(&self->ref_cnt, 1);
+}
+
+AckTrackerFactory *ack_tracker_factory_ref(AckTrackerFactory *self)
```
alltilla, line 36: Could you move the function name to a new line? Thanks!

lib/ack-tracker/ack_tracker_factory.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+static inline void
+_free(AckTrackerFactory *self)
+{
+  if (self && self->free_fn)
+    self->free_fn(self);
+}
+
+void ack_tracker_factory_unref(AckTrackerFactory *self)
```
alltilla, line 55: Could you move the function name to a new line? Thanks!

lib/ack-tracker/batched_ack_tracker.c
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+  log_msg_unref(msg);
+  log_pipe_unref((LogPipe *)self->super.source);
+}
+
+static void
+_disable_bookmark_saving(AckTracker *self)
+{
+  // TODO -> ??
```
alltilla, line 195: What will be the future of this function? Is it needed?

lib/logproto/logproto-server.c
```diff
@@ -100,6 +101,33 @@ find_eom(const guchar *s, gsize n)
...
+AckTrackerFactory *
+log_proto_server_get_ack_tracker_factory(LogProtoServer *self)
+{
+  if (!self->ack_tracker_factory)
+    self->ack_tracker_factory =  _default_ack_tracker_factory();
+
+  return self->ack_tracker_factory;
+}
+
+gboolean log_proto_server_is_position_tracked(LogProtoServer *s)
```
alltilla, line 125: Could you move the function name to a new line? Thanks!

lib/ack-tracker/batched_ack_tracker.c
lbudai: It should not happen as `LogSource` is  a refcounted `LogPipe` and whenever a message is tracked by the acktracker, the refcounter is increased. So `log_source_free` is called _after_ all messages (tracked by the tracker) has been acked.

lib/ack-tracker/batched_ack_tracker.c
lbudai: I think, it is - I need some time to finalize this.

lib/ack-tracker/batched_ack_tracker.c
alltilla: In this case, what is the point of acking in `AckTracker` `free`, when we know, that it will only be called, after everything is acked?

lbudai: @alltilla : connection between `LogSource` and `AckTracker` is a little bit weird: just see how complex the creation of the `AckTracker`. One of the reasons of this complexity is come from the design of the `LogProtos`: they can tell the type of the AckTracker they will need (rltp...) - and sometimes they don't.
How are the aforementioned problems are related? We know that AckTracker has to keep LogSource alive... now.
But if we could make AckTracker part of LogSource - then it would be a different story.

And there is another reason - a more practical: I want to test `BatchedAckTracker` separately from `LogSource`  (your question is absolutely valid as there is no sign of these tests in the code...)

lib/ack-tracker/batched_ack_tracker.c
lbudai: I think, you are right... will fix it soon.

lib/ack-tracker/batched_ack_tracker.c
lbudai: This is used only by `FileReader`  (`file_reader_stop_follow_file` - which is used only by `WildcardSource`), so no, it won't be supported by `BatchedAckTracker`.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/ack-tracker/batched_ack_tracker.c
```diff
@@ -0,0 +1,299 @@
...
+  g_mutex_lock(&self->acked_records_lock);
+  {
+    batch = self->acked_records;
+    self->acked_records = NULL;
+    self->acked_records_num = 0;
+  }
+  g_mutex_unlock(&self->acked_records_lock);
+  if (batch)
+    {
+      _ack_batch(self, batch);
```
MrAnno, line 164: I think we should reregister the timer here as well.

kira-syslogng:
Build SUCCESS

lib/ack-tracker/batched_ack_tracker.c
```diff
@@ -0,0 +1,300 @@
...
+      _start_batch_timer(self);
```
MrAnno, line 165: Sorry, I've found another one:

I think the timer should be restarted even when `batch` is empty.
In case there are no messages for `batch_timeout` millisecond, the timer won't be restarted, so we'll wait for a full batch after the inactive time period.

lbudai: you are right, I was added that line at the wrong place - it should be re-registered unconditionally

I fixed it not, please check it.

kira-syslogng:
Build SUCCESS

PR #3392, opened 2020-08-17
Python source acknowledgement

MrAnno:
This PR adds acknowledgment support for Python sources.

Example:
```Python
from syslogng import LogSource
from syslogng import LogMessage
from syslogng import Logger
from syslogng import InstantAckTracker, ConsecutiveAckTracker, BatchedAckTracker

import signal
signal.signal(signal.SIGINT, signal.SIG_IGN)

import threading
import time

class TestSource(LogSource):
    def init(self, options):
        self.stop_event = threading.Event()

        self.ack_tracker = InstantAckTracker(ack_callback=self.save_bookmark)
        self.ack_tracker = ConsecutiveAckTracker(ack_callback=self.save_bookmark)
        self.ack_tracker = BatchedAckTracker(timeout=5000, batch_size=100,
                                             batched_ack_callback=self.save_bookmarks)
        return True

    def deinit(self):
        pass

    def save_bookmark(self, bookmark):
        print(bookmark)

    def save_bookmarks(self, bookmarks):
        for b in bookmarks:
            print(b)

    def run(self):
        while not self.stop_event.is_set():
            time.sleep(1)

            msg = LogMessage("test", bookmark="bookmark-data")
            # msg.set_bookmark("bookmark-data")
            self.post_message(msg)

    def request_exit(self):
        self.stop_event.set()
```

### DIFF
--- a/modules/python/python-fetcher.c
+++ b/modules/python/python-fetcher.c
@@ -179,6 +179,12 @@ _ulong_to_fetch_result(unsigned long ulong, ThreadedFetchResult *result)
     }
 }
 
+static inline AckTracker *
+_py_fetcher_get_ack_tracker(PythonFetcherDriver *self)
+{
+  return ((LogSource *) self->super.super.worker)->ack_tracker;;
+}
+
 static gboolean
 _py_fetcher_fill_bookmark(PythonFetcherDriver *self, PyLogMessage *pymsg)
 {
@@ -190,7 +196,7 @@ _py_fetcher_fill_bookmark(PythonFetcherDriver *self, PyLogMessage *pymsg)
       return FALSE;
     }
 
-  AckTracker *ack_tracker = ((LogSource *)self->super.super.worker)->ack_tracker;
+  AckTracker *ack_tracker = _py_fetcher_get_ack_tracker(self);
   Bookmark *bookmark = ack_tracker_request_bookmark(ack_tracker);
 
   PyBookmark *py_bookmark = py_bookmark_new(pymsg->bookmark_data, self->py.ack_tracker_factory->ack_callback);
--- a/modules/python/python-fetcher.c
+++ b/modules/python/python-fetcher.c
@@ -28,6 +28,8 @@
 #include "str-utils.h"
 #include "string-list.h"
 #include "python-persist.h"
+#include "python-ack-tracker.h"
+#include "python-bookmark.h"
 
 #include <structmember.h>
 
@@ -177,6 +179,27 @@ _ulong_to_fetch_result(unsigned long ulong, ThreadedFetchResult *result)
     }
 }
 
+static gboolean
+_py_fetcher_fill_bookmark(PythonFetcherDriver *self, PyLogMessage *pymsg)
+{
+  if (!self->py.ack_tracker_factory)
+    {
+      msg_error("Error in Python fetcher, bookmarks can not be used without creating an AckTracker instance (self.ack_tracker)",
+                evt_tag_str("driver", self->super.super.super.super.id),
+                evt_tag_str("class", self->class));
+      return FALSE;
+    }
+
+  AckTracker *ack_tracker = ((LogSource *)self->super.super.worker)->ack_tracker;
+  Bookmark *bookmark = ack_tracker_request_bookmark(ack_tracker);
+
+  PyBookmark *py_bookmark = py_bookmark_new(pymsg->bookmark_data, self->py.ack_tracker_factory->ack_callback);
+  py_bookmark_fill(bookmark, py_bookmark);
+  Py_XDECREF(py_bookmark);
+
+  return TRUE;
+}
+
 static ThreadedFetchResult
 _py_invoke_fetch(PythonFetcherDriver *self, LogMessage **msg)
 {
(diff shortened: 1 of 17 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

furiel:
Me feels the proposed api is more complex than it should be. It is looks very similar to the c api, probably because this way it is the easiest to hook to the core. However, there may be things we can do to make this api easier to use.

In practice, users want to have control over two things: the acknowledgement strategy (instant, consecutive or batched), and the acknowledgement function. The rest is boilerplate, that would be great if we could hide from the users.

Do you think similar api could be implemented? (Omitted a lot of things for sake of brevity, but hope it shows the intention.)
```
class TestSource(LogSource):
    def init(self, options):
        self.ack_tracker = ConsecutiveAckTracker()
        return True

    def ack(self, data):
        print(data)

    def run(self):
        while not self.stop_event.is_set():
            time.sleep(1)
            msg = LogMessage("test", ack = functools.partial(self.ack, "test-data"))
            self.post_message(msg)
```

Normally, users should not be aware that the acktracker is created through a factory pattern. The term "Factory"  makes them focus on delayed execution (which is actually happening), but they should not know about that. The acknowledgement alone is a difficult concept, we should keep it as simple as possible. If we really need to setup the `LogSource->ack_tracker_factory` during the init method, then we may still remove the "factory" term for easier learning, and resolve the `self.ack_tracker` collision somehow differently.

If we directly specify the bookmark/acknowledgement function to logmessage as constructor parameter, that eliminates the `set_bookmark` call. Bookmarks cannot really stand on their own, they are always connected to the logmessage. So separating the bookmark from the logmessage through `set_bookmark` is not really useful. The Bookmark class is not added to the syslogng module either.

I would also drop the Bookmark class/api. It is just alternative for Callable, and used for the same reason. We could use functions directly instead instead of Bookmark.
Another problem with the bookmark api is that self.data stands out a bit. In c, that is usually a `gpointer user_data`, because we do not have any other means to express generic data. But in python it does not look idiomatic.

Eliminating request_bookmark is a bit trickier. But I think if we save enough information inside the log message, we can properly initialize the c-bookmark before log_source_post.

lbudai:
@furiel : could you rephrase your example (using `BatchedAckTracker`)? 
I think Bookmarks are not the same as ACKs: in case of BatchedAckTracker you don't want to ACK all messages, but want to implement a function that receives the batch of bookmarks.

furiel:
We talked about this problem in the ose meeting, and we have a refined version. @MrAnno will write a proper explanation, but the summary is that we can provide both kind of acknowledgement functions (batched and per message) in the constructors of the ack tracker factory. This works for the per message version too, because the acknowledgement function is always the same, only the data changes. So instead of the new ack parameter in the message, we will only pass the data there.

Kokan (approved):

MrAnno:
@lbudai @furiel 

This is an example of the refined version. This gets rid of the Bookmark object and the ack tracker factories.
Requesting bookmark is not necessary anymore, the bookmark data is bound to the LogMessage instance.

```
from syslogng import LogSource
from syslogng import LogMessage
from syslogng import Logger

import signal
signal.signal(signal.SIGINT, signal.SIG_IGN)

import threading
import time

class TestSource(LogSource):
    def init(self, options):
        self.stop_event = threading.Event()
        self.create_instant_ack_tracker(ack_callback=self.save_bookmark)
        self.create_consecutive_ack_tracker(ack_callback=self.save_bookmark)
        self.create_batched_ack_tracker(timeout=10000, batch_size=100, batched_ack_callback=self.save_bookmarks)
        return True

    def deinit(self):
        pass

    def save_bookmark(self, bookmark):
        print(bookmark)

    def save_bookmarks(self, bookmarks):
        for b in bookmarks:
            print(b)

    def run(self):
        while not self.stop_event.is_set():
            time.sleep(1)

            msg = LogMessage("test", bookmark="bookmark-data")
            msg.set_bookmark("bookmark-data")
            self.post_message(msg)

    def request_exit(self):
        self.stop_event.set()
```

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

MrAnno:
@furiel I've pushed the refined version, I've ended keeping the constructors, but not using the "Factory" suffix:

```
self.ack_tracker = InstantAckTracker(ack_callback=self.save_bookmark)
```

This way, we'll be able to replace `self.ack_tracker` with a real AckTracker type before starting the Python source, if necessary.
Currently, it does not seem to be necessary.

kira-syslogng:
Build SUCCESS

lbudai (approved):

lbudai:
I think we should add a method for `ack_tracker_init` and `deinit` methods.
Where we are using a `BatchedAckedTtracker`, we want to ack partial batches before closing the client.

kira-syslogng:
Build SUCCESS

modules/python/python-fetcher.c
```diff
@@ -384,6 +386,30 @@ _py_parse_options_new(PythonFetcherDriver *self, MsgFormatOptions *parse_options
...
+      msg_error("Python source attribute ack_tracker needs to be an AckTracker subtype",
+                evt_tag_str("driver", self->super.super.super.super.id),
+                evt_tag_str("class", self->class));
+      return FALSE;
+    }
+
+  self->py.ack_tracker_factory = (PyAckTrackerFactory *) py_ack_tracker_factory;
+
+  AckTrackerFactory *ack_tracker_factory = self->py.ack_tracker_factory->ack_tracker_factory;
+  self->super.super.worker_options.ack_tracker_factory = ack_tracker_factory;
```
lbudai, line 446: `ack_tracker_factory_unref` is called in `log_threaded_source_worker_options_destroy`, so I think `ack_tracker_factory` should be referenced.

lbudai (changes requested):

modules/python/python-fetcher.c
MrAnno: Fixed, thank you.

kira-syslogng:
Build SUCCESS

Kokan (approved):

lbudai (approved):

PR #3393, opened 2020-08-19
Filter regex missing named matches

Kokan:
When the regex filters used the `store-matches` flag and it was possible to not to set all of the named matches.
The current known case for not setting is when there were an optional named match: `(?<foobar>foobar) (?<foo>foo)?(?<bar>bar)` such as `foo` here, and if the provided string did not contain the `foo` string itself. In that case either `foobar` or `bar` were not saved into the `LogMessage`.

Additionally there was a test case error, that could lead to false positive asserts.

### DIFF
--- a/lib/filter/tests/test_filters_common.c
+++ b/lib/filter/tests/test_filters_common.c
@@ -213,7 +213,9 @@ testcase_with_backref_chk(const gchar *msg,
     }
   else
     {
-      cr_assert_eq(strncmp(value_msg, value, length), 0,
+      const gint value_len = strlen(value);
+      cr_assert_eq(length, value_len);
+      cr_assert_eq(strncmp(value_msg, value, value_len), 0,
                    "Filter test failed (value chk); msg='%s', expected_value='%s', value_in_msg='%s'",
                    msg, value, value_msg);
     }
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
Apart from the comment on non-nul terminated matches this looks good to me. Good catch.

Kokan:
@Bazsi I cannot see your comment on the non null terminated part, but likely you refer to the `strncmp` where the value obtained from `LogMessage` can be not null terminated.
I could add an additional `cr_assert_eq(length, strlen(value));` that should  resolve such case, but imho that is going to fail for me either way as `value` won't be equal to `value_msg`.

bazsi:
On Wed, Aug 19, 2020 at 02:11:26AM -0700, Kókai Péter wrote:
> @Bazsi I cannot see your comment on the non null terminated part, but likely you refer to the `strncmp` where the value obtained from `LogMessage` can be not null terminated.
> I could add an additional `cr_assert_eq(length, strlen(value));` that should  resolve such case, but imho that is going to fail for me either way as `value` won't be equal to `value_msg`.
> 

You might cause a read-overflow in value_msg, in case of a failed assertion. 
Just imagine that your expected value is 'foobar\0' (e.g.  strlen(value) is
6) and the value returned by log_msg_get_value() points to "foo" and the
returned length is 3.  We don't know anything about bytes following the
match, but they might accidentally be "bar", and they are not considered
part of the name-value pair.  

In this case, your assertion passes, while it shouldn't have.

The "length" returned by log_msg_get_value() is an important part when the
value is not NUL terminated.

Such non-NUL terminated values are used by regexp matches and by db-parser()
when we are using indirect NVEntry instances (e.g.
log_msg_set_match_indirect())

-- 
Bazsi

kira-syslogng:
Build SUCCESS

szemere (approved):

bazsi (approved):

PR #3401, opened 2020-08-26
json-parser: per message internal logs should be on debug level

Kokan:


kira-syslogng:
Build SUCCESS

modules/json/json-parser.c
```diff
@@ -239,7 +239,7 @@ json_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_
   jso = json_tokener_parse_ex(tok, input, input_len);
   if (tok->err != json_tokener_success || !jso)
     {
-      msg_error("json-parser(): failed to parse JSON payload",
+      msg_debug("json-parser(): failed to parse JSON payload",
```
MrAnno, line 242: `json-parser()` drops the message when it fails. I think `msg_error()` is not a bad choice in that case.

Using `json-parser()` in `if` blocks, however, becomes problematic, because these error messages are produced for each unsuccessful attempt.

Kokan: That log message do not indicate in any way that a message is going to be dropped, as such it is not helpful. Also dropped messages are not logged, but counted as a counter. If a parser drops the message or not is not really straight forward in syslog-ng.

As such I do not see the argument to be valid, while on the other hand such per message errors are always pushed to at least debug level.

MrAnno (approved):

modules/json/json-parser.c
szemere: I think it was a very good point from @MrAnno, json-parser is used in multiple ways inside syslog-ng. I can accept @Kokan's answer. Approve  from me.

szemere (approved):

PR #3404, opened 2020-08-26
logthrsourcedrv: keep LogSource alive until ACKs can be received

lbudai:
If a Python source uses a library that sends back ACKs from Futures,
it is possible that after a reload (deinit()) ACKs are still received.


kira-syslogng:
Build SUCCESS

PR #3405, opened 2020-08-26
python: fixing printing None instead of the actual exception

Kokan:
Kudos for @alltilla finding the ordering issue.

If an exception is triggered in python code, such as with the following
configuration[1], normally a message reports in which function what kind of
exception happened. If syslog-ng was compiled with recent clang (10.0) instead of
showing the exception the default None was displayed.

I did not dig into deep enough why this only happens with clang and not with gcc,
it is a good idea - just like in case of ERRNO - to save the exception before
anything else, this patch does exactly that.

Note: also unit test fails due to this issue with clang

[1] configuration that could trigger this issue
```
python {

from syslogng import LogSource

class S(LogSource):
  def init(self, options):
    raise ValueError('exception')

  def run(self):
    pass

  def request_exit(self):
    pass

};

log {
  source {
    python(
        class(S)
    );
  };
```

Signed-off-by: Kokan <kokaipeter@gmail.com>

### DIFF
--- a/modules/python/python-debugger.c
+++ b/modules/python/python-debugger.c
@@ -117,10 +117,11 @@ python_fetch_debugger_command(void)
   if (!ret)
     {
       gchar buf[256];
+      _py_format_exception_text(buf, sizeof(buf));
 
       msg_error("Error calling debugger fetch_command",
                 evt_tag_str("function", DEBUGGER_FETCH_COMMAND),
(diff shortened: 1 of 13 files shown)


kira-syslogng:
Build SUCCESS

szemere (approved):

alltilla (approved):

PR #3407, opened 2020-08-27
Add namespace to the systemd-journal persist name

szemere:
WIP: This PR is a modification on top of #3358 (it should be merged first), however I was unable to open a pull request on: https://github.com/bjoe2k4/syslog-ng

This PR helps to resolve _CURSOR issues in the persist file in case the namespace option changes.
(According to the systemd documentation the _CURSOR identifies an entry. However we got seeking errors in case the namespace parameters changes, even if syslog-ng should seek to the same log entry again.)

This PR also prepares the code to allow multiple systemd-journal sources at the same time. Before systemd v2.45 it was illogical to add multiple journal sources (they could even conflict with each other) so it is prohibited now. However if two journal sources are configured to read from two different namespaces they can work side by side.

I would like to thank for @bjoe2k4 for all the work to introduce the namespace option to the systemd-journal source.

### DIFF
--- a/modules/systemd-journal/journal-reader.c
+++ b/modules/systemd-journal/journal-reader.c
@@ -615,6 +615,12 @@ _journal_open(JournalReader *self)
           journal_namespace = self->options->namespace;
         }
     }
+
+  if (journal_namespace)
+    {
+      g_free(self->persist_name);
+      self->persist_name = g_strdup_printf("systemd_journal(%s)", self->options->namespace);
+    }
   gint res = journald_open_namespace(self->journal, journal_namespace, journal_flags);
 #else
   gint res = journald_open(self->journal, SD_JOURNAL_LOCAL_ONLY);
--- a/modules/systemd-journal/journal-reader.c
+++ b/modules/systemd-journal/journal-reader.c
@@ -585,47 +585,61 @@ _add_poll_events(JournalReader *self)
   return TRUE;
 }
 
-static gint
-_journal_open(JournalReader *self)
-{
 #if SYSLOG_NG_HAVE_JOURNAL_NAMESPACES
-  gint journal_flags = SD_JOURNAL_LOCAL_ONLY;
-  gchar *journal_namespace = NULL;
-
-  if (strcmp(self->options->namespace, "*") == 0)
+static void
+_journal_process_namespace(gchar *namespace_option, gchar **journal_namespace, gint *journal_flags)
+{
+  if (strcmp(namespace_option, "*") == 0)
     {
-      journal_flags |= SD_JOURNAL_ALL_NAMESPACES;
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

modules/systemd-journal/journal-reader.c
```diff
@@ -585,47 +585,61 @@ _add_poll_events(JournalReader *self)
...
+  return journald_open_namespace(self->journal, journal_namespace, journal_flags);
+#else
+  return journald_open(self->journal, journal_flags);
+#endif
+}
+
+static void
+_init_persist_name(JournalReader *self)
+{
+  if (SYSLOG_NG_HAVE_JOURNAL_NAMESPACES && (strcmp(self->options->namespace, "*") != 0))
```
Kokan, line 635: Alternatives:
1. let the `*` be the default value of `self->options->namespace` (and set it to that even if there is no namespace support), as a result here `SYSLOG_NG_HAVE_JOURNAL_NAMESPACES` check is not needed.
2. the `SYSLOG_NG_HAVE_JOURNAL_NAMESPACES` should be used in preprocessor only not in condition

szemere: Thank You!

kira-syslogng:
Build SUCCESS

Kokan (approved):

kira-syslogng:
Build SUCCESS

gaborznagy (approved):

furiel (dismissed):

kira-syslogng:
Build SUCCESS

news/bugfix-3407.md
```diff
@@ -0,0 +1 @@
+systemd-journal: add namespace to the persist name
```
gaborznagy, line 1: I don't want to nitpick with news file reviews, but can we emphasize here what you wrote in PR description, please?

> ....  allow multiple systemd-journal sources at the same time. .....  two journal sources are configured to read from two different namespaces they can work side by side.

gaborznagy (approved):
With just a proposal for news file content, approve from me.

news/bugfix-3407.md
Kokan: The current version pushed forward to support multiple `systemd-journal()` source, yet not lifted the restriction. From user point of view the one instance limitation still stands. As such, it would not help to mention in the news file.

gaborznagy: Yes, you are right, I forgot the `journal_reader_initialized` global variable restriction...

https://github.com/syslog-ng/syslog-ng/blob/master/modules/systemd-journal/journal-reader.c#L630

Kokan (approved):

PR #3414, opened 2020-09-01
Stats response opt

lbudai:
on my machine it's not significantly better than the old method... (2.9GB memory usage -> 1.63 vs 1.68 s)
-> my plan is to change stats callback from `GString*` to `StringArray`  and remove `join`.

config that I used:
```@version: 3.29

options {
 stats-level(3);
};

log {
  source {
  example-msg-generator(freq(0) num(3000000)
};


```


kira-syslogng:
Build FAILURE

lib/string-array.c
```diff
@@ -0,0 +1,119 @@
...
+}
+
+GString *
+string_array_join(StringArray *self, gboolean free_elements)
+{
+  gsize len = _total_str_len(self);
+  GString *str = g_string_sized_new(len + 1);
+  string_array_foreach(self, _join, str);
+  if (free_elements)
+    self->array = g_ptr_array_remove_range(self->array, 0, string_array_len(self));
```
furiel, line 115: You could use: `g_ptr_array_set_size(self->array, 0)`.

furiel (commented):
So far looks good to me.

PR #3415, opened 2020-09-02
systemd-journal: match g_free with g_strdup

Kokan:
A libc allocation must be deallocated with libc deallocaton (example: strdup -> free),
and the same with glib (example: g_strdup -> g_free).

News file not needed.

### DIFF
--- a/modules/systemd-journal/journal-reader.c
+++ b/modules/systemd-journal/journal-reader.c
@@ -780,6 +780,58 @@ journal_reader_options_init(JournalReaderOptions *options, GlobalConfig *cfg, co
 }
 
 void
+journal_reader_options_set_default_severity(JournalReaderOptions *self, gint severity)
+{
+  if (self->default_pri == 0xFFFF)
+    self->default_pri = LOG_USER;
+  self->default_pri = (self->default_pri & ~7) | severity;
+}
+
+void
+journal_reader_options_set_default_facility(JournalReaderOptions *self, gint facility)
+{
+  if (self->default_pri == 0xFFFF)
+    self->default_pri = LOG_NOTICE;
+  self->default_pri = (self->default_pri & 7) | facility;
+}
+
+void
+journal_reader_options_set_time_zone(JournalReaderOptions *self, gchar *time_zone)
+{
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

modules/systemd-journal/systemd-journal-grammar.ym
```diff
@@ -104,15 +104,15 @@ source_systemd_journal_option
           {
             if (last_journal_reader_options->recv_time_zone)
               free(last_journal_reader_options->recv_time_zone);
-            last_journal_reader_options->recv_time_zone = strdup($3);
+            last_journal_reader_options->recv_time_zone = g_strdup($3);
```
bazsi, line 107: this should have been a setter function though. This already improves things, so I wouldn't insist, but if we touch this code, we might have added the setter function too.

Kokan: sure, will do, it was just something I saw during review and wanted it get fixed.

kira-syslogng:
Build SUCCESS

gaborznagy (approved):
Thanks for the refactors as well!

modules/systemd-journal/journal-reader.c
```diff
@@ -779,6 +779,58 @@ journal_reader_options_init(JournalReaderOptions *options, GlobalConfig *cfg, co
...
+void
+journal_reader_options_set_max_field_size(JournalReaderOptions *self, gint max_field_size)
+{
+  self->max_field_size = max_field_size;
+}
+
+void
+journal_reader_options_set_namespace(JournalReaderOptions *self, gchar *namespace)
+{
+  if (self->namespace)
```
bazsi, line 823: Nitpick: we dont need the conditional, g_free() checks if the argument is NULL.

bazsi (approved):

MrAnno (approved):

modules/systemd-journal/journal-reader.c
Kokan: agree, but the pre allocated time fitted that much for now.

PR #3416, opened 2020-09-03
afprog: manual cloexec

### LINKED ISSUE #3409: network: syslog-ng fails to bind() after config revert
# syslog-ng
## Version of syslog-ng
3.28

# Issue
## Failure
When having a program source or destination and a network destination in the config, if we reload with an invalid config, syslog-ng crashes, as it cannot init the old network source, because its address is in use.

There have been some investigation about this issue: the problem is caused by the way we start the program destination.

 1. In case of a config revert we have an `init()` -> `deinit()` -> `init()` happening really fast. In the first `init()`, we `bind()` the socket, and create the program dest with a separate pid (via `fork()` then `execl()`), which will have our FDs. Although we set the `CLOEXEC`[1] flag set on the FD, it still takes time to close the FD, when `execl()` is running in the other process.
 2. We then call `deinit()`, where we close the FD, but the other process still did not close the FD via `CLOEXEC`, as it did not get there in the `execl()` call yet.
 3. Then we call `init()`, we try to `bind()` to the socket, but is is still in use.

[1] https://www.gnu.org/software/libc/manual/html_node/Descriptor-Flags.html

The problem is reproducible with this Jancsika code:
```
#include <unistd.h>
#include <stdio.h>
#include <sys/socket.h>
#include <stdlib.h>
#include <netinet/in.h>
#include <string.h>
#include <arpa/inet.h>
#include <netinet/tcp.h>
#include <fcntl.h>
#include <assert.h>
#include <signal.h>
#include <errno.h>
#include <string.h>
#include <sys/wait.h>

#define PORT 12345

int
socket_open_and_bind()
{
  int fd = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);

  fcntl(fd, F_GETFL);
  fcntl(fd, F_SETFL, O_RDWR|O_NONBLOCK);
  fcntl(fd, F_GETFD);
  fcntl(fd, F_SETFD, FD_CLOEXEC);

  int opt = 1;
  int ka = 1;
  int ki = 60;
  int kc = 6;
  int kin = 10;
  setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, &ka, sizeof(ka));
  setsockopt(fd, SOL_TCP, TCP_KEEPIDLE, &ki, sizeof(ki));
  setsockopt(fd, SOL_TCP, TCP_KEEPCNT, &kc, sizeof(kc));
  setsockopt(fd, SOL_TCP, TCP_KEEPINTVL, &kin, sizeof(kin));
  setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &opt, sizeof(opt));

  struct sockaddr_in address;
  address.sin_family = AF_INET;
  address.sin_addr.s_addr = inet_addr("0.0.0.0");
  address.sin_port = htons(PORT);

  int rc = bind(fd, (struct sockaddr *)&address, sizeof(address));
  if (rc < 0)
    {
      printf("bind failed %d %s\n", rc, strerror(errno));
      exit(EXIT_FAILURE);
    }
  printf("bind success\n");

  listen(fd, 255);

  int oob = 1;
  setsockopt(fd, SOL_SOCKET, SO_OOBINLINE, &oob, sizeof(oob));

  return fd;
}

int
main(int argc, char const *argv[])
{
  int fd = socket_open_and_bind();

  int ctl_pipe[2];
  pipe(ctl_pipe);

  pid_t pid = fork();
  if (pid == 0)
    {
      setpgid(0, 0);
      execl("/bin/sh", "/bin/sh", "-c", "sleep 10", NULL);
      assert(0);
    }
  else
    {
      close(fd);

      pid_t pgid = getpgid(pid);
      if (pgid != -1 && pgid != getpgrp())
        killpg(pgid, SIGTERM);

      fd = socket_open_and_bind();
    }

  return 0;
}
```

To fix the issue, we should wait for the `execl()` to close the FDs when we are `init()`-ing, and we should wait for the child process to stop when we are `deinit()`-ing.

## Steps to reproduce
 1. Start `syslog-ng` with the configuration below.
 2. Uncomment the `filter` line in the config.
 3. `syslog-ng-ctl reload`.
 4. Observe the error and the crash.

## Configuration
```
@version: 3.29

source s_network {
    network(port(12345)); # keep-alive(yes/no) can be set, the problem still persists
};

destination d_program {
    program("sleep 600");
};

filter f_1 {
    filter(f_1);
};

log {
    source(s_network);
    destination(d_program);

    #filter(f_1);
};
```

## Input and output logs
```
[2020-08-28T10:16:31.840373] Module loaded and initialized successfully; module='syslogformat'
[2020-08-28T10:16:31.840506] Accepting connections; addr='AF_INET(0.0.0.0:12345)'
[2020-08-28T10:16:31.840556] Starting destination program; cmdline='sleep 600'
[2020-08-28T10:16:31.840935] Program destination started; cmdline='sleep 600', fd='15'
[2020-08-28T10:16:31.841097] Loop detected in filter rule; rule='f_1'
[2020-08-28T10:16:31.841132] Error initializing message pipeline; plugin_name='not a plugin', location='/home/alltilla/Work/install/OSE/etc/syslog-ng.conf:11:13'
[2020-08-28T10:16:31.841153] Error initializing new configuration, reverting to old config;
[2020-08-28T10:16:31.841193] Closing listener fd; fd='11'
[2020-08-28T10:16:31.841252] Sending destination program a TERM signal; cmdline='sleep 600', child_pid='24998'
[2020-08-28T10:16:31.841266] Sending TERM signal to the process group; pid='24998'
[2020-08-28T10:16:31.841292] Closing log transport fd; fd='15'
[2020-08-28T10:16:31.841426] Unable to detect fully qualified hostname for localhost, use_fqdn() will use the short hostname;
[2020-08-28T10:16:31.841589] Error binding socket; addr='AF_INET(0.0.0.0:12345)', error='Address already in use (98)'
[2020-08-28T10:16:31.841626] Error initializing message pipeline; plugin_name='network', location='/home/alltilla/Work/install/OSE/etc/syslog-ng.conf:4:5'
Quit (core dumped)
```
```
#0  0x00007ffff7baa55b in kill () at ../sysdeps/unix/syscall-template.S:78
#1  0x00007ffff7ef47cd in main_loop_reload_config_revert (
    user_data=0x7ffff7fc4800 <main_loop>) at ../lib/mainloop.c:245
#2  0x00007ffff7ef497a in main_loop_reload_config_apply (user_data=0x7ffff7fc4800 <main_loop>)
    at ../lib/mainloop.c:281
#3  0x00007ffff7ef66cf in _consume_action (action=0x555555590e60)
    at ../lib/mainloop-worker.c:240
#4  0x00007ffff7ef670b in _invoke_sync_call_actions () at ../lib/mainloop-worker.c:250
#5  0x00007ffff7ef6a68 in main_loop_worker_sync_call (
    func=0x7ffff7ef4858 <main_loop_reload_config_apply>, user_data=0x7ffff7fc4800 <main_loop>)
    at ../lib/mainloop-worker.c:396
#6  0x00007ffff7ef4c55 in main_loop_reload_config_commence (self=0x7ffff7fc4800 <main_loop>)
    at ../lib/mainloop.c:340
#7  0x00007ffff7ef7360 in control_connection_reload (cc=0x5555555a02c0, 
    command=0x555555562980, user_data=0x7ffff7fc4800 <main_loop>)
    at ../lib/mainloop-control.c:167
#8  0x00007ffff7f46181 in control_connection_io_input (s=0x5555555a02c0)
    at ../lib/control/control-server.c:197
#9  0x00007ffff7f64ee5 in iv_fd_poll_and_run (st=0x555555563f30, 
    abs=0x7ffff7fc7900 <stats_timer>) at ../../../../lib/ivykis/src/iv_fd.c:223
#10 0x00007ffff7f662fa in iv_main () at ../../../../lib/ivykis/src/iv_main_posix.c:112
#11 0x00007ffff7ef5660 in main_loop_run (self=0x7ffff7fc4800 <main_loop>)
    at ../lib/mainloop.c:658
#12 0x0000555555556d1c in main (argc=1, argv=0x7fffffffde48) at ../syslog-ng/main.c:314
```

## Possible fix
The Jancsika code works, when syncing the child and parent processes BEFORE `execl()`, and waiting for the child process to stop, when calling `killpg()` (these changes fix it in `syslog-ng`, too):
```
int
main(int argc, char const *argv[])
{
  int fd = socket_open_and_bind();

  int ctl_pipe[2];
  pipe(ctl_pipe);

  pid_t pid = fork();
  if (pid == 0)
    {
      setpgid(0, 0);

      close(ctl_pipe[0]);
      close(ctl_pipe[1]);
      printf("child in sync\n");

      execl("/bin/sh", "/bin/sh", "-c", "sleep 10", NULL);
      assert(0);
    }
  else
    {
      close(ctl_pipe[1]);
      char buff[1];
      assert(read(ctl_pipe[0], buff, sizeof(buff)) == 0);
      close(ctl_pipe[0]);
      printf("parent in sync\n");

      close(fd);

      pid_t pgid = getpgid(pid);
      if (pgid != -1 && pgid != getpgrp())
        killpg(pgid, SIGTERM);

      int status = 420;
      pid_t wp = waitpid(pgid, &status, 0);
      printf("waitpid: %d %s\n", wp, strerror(errno));
      printf("program closed %d\n", status);

      fd = socket_open_and_bind();
    }

  return 0;
}
```
However this fix is not complete, as the sync between the child and parent should be after `execl()` closes the FD. As `execl()` 
does not let us run code between the close, and changing the current process to the new, we cannot simply sync.
We can however write a 2 step `exec()`, where in the C code we `execl()` a shell script, which does the sync with the parent process, then calls `exec()` itself, with the user's program.

Kokan:
Parent waits for child to close all *fd* before `exec`.

Fixes #3409.

### DIFF
--- a/modules/afprog/afprog.c
+++ b/modules/afprog/afprog.c
@@ -33,6 +33,7 @@
 #include "logproto/logproto-text-client.h"
 #include "poll-fd-events.h"
 
+#include <sys/resource.h>
 #include <sys/types.h>
 #include <errno.h>
 #include <fcntl.h>
@@ -91,10 +92,38 @@ _exec_program(const gchar *cmdline)
   execl("/bin/sh", "/bin/sh", "-c", cmdline, NULL);
 }
 
+static void
+_close_all_fd(void)
+{
+  struct rlimit rlp;
+  if (getrlimit(RLIMIT_NOFILE, &rlp) < 0)
+    {
+      /*
+       * Failing to query max *fd*, still closing an arbitrary range might make sense,
+       * tring with some arbitrary big number. The only issue could be that it cannot
+       * close all of the needed *fd* and fail to bind to a socket (original issue).
+       */
+      rlp.rlim_max = 10000;
+    }
+
+  for (int i = rlp.rlim_max; i > 2; --i)
+    {
+      close(i);
+    }
+}
+
+static void
+_wait_for_child_closing_fds(int pipe[2])
+{
+  gchar buff[1];
+  while (read(pipe[0], buff, 1) != 0);
+}
+
 static gboolean
 afprogram_popen(AFProgramProcessInfo *process_info, GIOCondition cond, gint *fd)
 {
   int msg_pipe[2];
+  int sync_pipe[2];
 
   g_return_val_if_fail(cond == G_IO_IN || cond == G_IO_OUT, FALSE);
 
@@ -106,12 +135,24 @@ afprogram_popen(AFProgramProcessInfo *process_info, GIOCondition cond, gint *fd)
       return FALSE;
     }
 
+  if (pipe(sync_pipe) == -1)
+    {
+      msg_error("Error creating program pipe",
+                evt_tag_str("cmdline", process_info->cmdline->str),
+                evt_tag_error(EVT_TAG_OSERROR));
+      close(msg_pipe[0]);
+      close(msg_pipe[1]);
+      return FALSE;
+    }
+
   if ((process_info->pid = fork()) < 0)
     {
       msg_error("Error in fork()",
                 evt_tag_error(EVT_TAG_OSERROR));
       close(msg_pipe[0]);
       close(msg_pipe[1]);
+      close(sync_pipe[0]);
+      close(sync_pipe[1]);
       return FALSE;
     }
 
@@ -141,9 +182,12 @@ afprogram_popen(AFProgramProcessInfo *process_info, GIOCondition cond, gint *fd)
           dup2(devnull, 1);
           dup2(devnull, 2);
         }
+      dup2(sync_pipe[1], 3);
       close(devnull);
       close(msg_pipe[0]);
       close(msg_pipe[1]);
+
+      _close_all_fd();
 
       if (process_info->inherit_environment)
         _exec_program(process_info->cmdline->str);
@@ -152,6 +196,11 @@ afprogram_popen(AFProgramProcessInfo *process_info, GIOCondition cond, gint *fd)
 
       _exit(127);
     }
+
+  close(sync_pipe[1]);
+  _wait_for_child_closing_fds(sync_pipe);
+  close(sync_pipe[0]);
+
   if (cond == G_IO_IN)
     {
       *fd = msg_pipe[0];
--- a/news/bugfix-3416.md
+++ b/news/bugfix-3416.md
@@ -1,0 +1,5 @@
+afsocket: syslog-ng fails to bind() after config revert
+
+When having a program source or destination and a network destination in the
+config, if we reload with an invalid config, syslog-ng crashes, as it cannot init
+the old network source, because its address is in use.


modules/afprog/afprog.c
```diff
@@ -91,10 +91,72 @@ _exec_program(const gchar *cmdline)
...
+      gint result = write(pipe[1], sync_msg+written, sync_msg_len-written);
+      if (result < 0)
+        {
+          msg_warning("afprog(): sending sync to parent failed", evt_tag_error(EVT_TAG_OSERROR));
+          return;
+        }
+      written += result;
+    }
+  while (written < sync_msg_len);
+}
```
bazsi, line 126: this can be simplified a bit, as we don't need a message. it is enough to just close the write sides of a pipe and the reader would get a zero sized read.

so parent does a blocking read(), and once it returns indicating an EOF, that child is guaranteed to have closed the fd.
no need to read/write actual data.

modules/afprog/afprog.c
```diff
@@ -91,10 +91,72 @@ _exec_program(const gchar *cmdline)
   execl("/bin/sh", "/bin/sh", "-c", cmdline, NULL);
 }
 
+static void
+_closing_all_fd_expect(int ignore[], int ignore_len)
```
bazsi, line 95: typo, "except" instead of expect

bazsi: also, the "closing" part of the name should probably be a verb.

modules/afprog/afprog.c
```diff
@@ -91,10 +91,72 @@ _exec_program(const gchar *cmdline)
...
+{
+  for (int i = 10000; i > 5; --i)
```
bazsi, line 97: getrlimit(RLIMIT_NOFILE)

modules/afprog/afprog.c
```diff
@@ -145,13 +217,25 @@ afprogram_popen(AFProgramProcessInfo *process_info, GIOCondition cond, gint *fd)
       close(msg_pipe[0]);
       close(msg_pipe[1]);
 
+      _closing_all_fd_expect(sync_pipe, 2);
```
bazsi, line 220: if you dup2() the writer side of sync_pipe to fd 3, you can simply close all fds above 3, no need for the "ignore" support.

kira-syslogng:
Build SUCCESS

modules/afprog/afprog.c
bazsi: and if you iterate backwards (e.g. from large fds to small ones), that could even do the closing of the sync pipe.

modules/afprog/afprog.c
Kokan: thanks, I was just looking for this right now.

modules/afprog/afprog.c
Kokan: yeah, the first version was like that but it did not worked for me, there was an error and such it woun't read EOF. now fixed, and simple version also works.

alltilla:
Do we still need the `CLOEXEC` flag on the fds? In other words: can `exec()` be called from any other place, than the `program` `dest/src`? Maybe our `java` or `python` bindings do such thing.

We might be able to remove the `CLOEXEC` flag, because it is replaced by this new logic.

kira-syslogng:
Build SUCCESS

Kokan:
Both java and python languages bindings uses threads and not forking. In spite of those modules there is still fork used in syslog-ng.
1. The very first when syslog-ng stars and spawn a supervisor for itself (as needed); which is not an issue as that is created before anything else, so there is interesting open *fd*s.
2. command-hooks also uses fork (but it directly calls `system` which yields `fork`+`exec`)

Not to mention that we could have a lib that does `fork`+`exec` without us knowing (sure not likely), imho better to keep them.

alltilla (approved):
This works perfectly, thanks! :+1:

gaborznagy (commented):
Besides Bazsi's comment to use fd 3 approve from my side.

kira-syslogng:
Build SUCCESS

modules/afprog/afprog.c
```diff
@@ -106,12 +135,22 @@ afprogram_popen(AFProgramProcessInfo *process_info, GIOCondition cond, gint *fd)
       return FALSE;
     }
 
+  if (pipe(sync_pipe) == -1)
+    {
+      msg_error("Error creating program pipe",
+                evt_tag_str("cmdline", process_info->cmdline->str),
+                evt_tag_error(EVT_TAG_OSERROR));
+      return FALSE;
```
alltilla, line 143: Shouldn't we close `msg_pipe` in case of failure?

Kokan: yes, we should.

alltilla (approved):
Thanks!

kira-syslogng:
Build SUCCESS

gaborznagy (approved):
Thanks!

PR #3420, opened 2020-09-07
WIP: Light: Introducing FileReader and FileWriter classes

mitzkia:
**Important note:** This patch set is changing the current user API for the SyslogNgConfig model.

About the patch set:
- This patch set is a minor step further to refactor SyslogNgConfig model for Light.
- Before this patch there was a strong coupling between the file object itself and the file based config object. Which gives us an ambiguous usage in testcases, and makes hard to refactor the whole SyslogNgConfig model.
- The patch set contains 4 patches:
   - a minor followup change related to pytest configuration updates
   - splitting the old FileIO into Reader and Writer classes and make followup changes in the code
   - make grounding the usage of the FileReader and FileWriter in existing config model: 
     - move wrappers into their new place
     - delete old unused code
   - followup changes in functional testcases

My questions related to this PR:
- I have replaced old wrappers (like: create_file_source()) with a new one to prevent boilerplates in functional testcases (importing and starting using proper FileReader or FileWriter classes in each testcases)
- For now I am a little bit uncertain of this new wrappers like: create_file_source_and_writer(), this wrappers does two things (not one), but it is also true in most cases we will need a separate reader or writer for that particular config object which we have created. 
- Other thing is I did not like this new constructed name: create_some_driver_and_writer or create_some_driver_and_reader, they are not self-evident (so new ideas are welcome).


kira-syslogng:
Build SUCCESS

mitzkia:
I have got a new idea about the separating of read - write side from config, which I want to check (test), until than I put this PR to WIP.

PR #3425, opened 2020-09-11
Editline debugger cli

lbudai:
support BSD licensed `editline` when using debuggercli

motivation: when you have a python without the GPL licensed readline, then you can install [pyeditline](https://pypi.org/project/pyeditline/) for enabling command-completion support. This patch is for enabling completion support also for the built-in debuggercli.

The default is readline - when your python have readline module, then using editline in debuggercli is pointless.

I've tested it my environments and worked as expected.

### DIFF
--- a/modules/python/pylib/syslogng/debuggercli/__init__.py
+++ b/modules/python/pylib/syslogng/debuggercli/__init__.py
@@ -21,11 +21,31 @@
 #############################################################################
 
 from __future__ import absolute_import
-from syslogng.debuggercli.readline import setup_readline
 
+def is_readline_available():
+    try:
+        import readline
+    except ImportError:
+        return False
+    return True
+
+def is_editline_available():
+    try:
+        import editline
+    except ImportError:
+        return False
+    return True
+
+def setup_read_or_editline():
+    if is_readline_available():
+        from syslogng.debuggercli.readline import setup_readline
+        setup_readline()
+    elif is_editline_available():
+        from syslogng.debuggercli.editline import setup_editline
+        setup_editline()
 
 def fetch_command():
-    setup_readline()
+    setup_read_or_editline()
     try:
         input_function = raw_input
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/python/pylib/syslogng/debuggercli/editline.py
```diff
@@ -0,0 +1,88 @@
...
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+from __future__ import absolute_import, print_function
```
alltilla, line 24: This seems like an unnecessary import.

alltilla (approved):
Apart from that one comment, it is an approve from me. Thanks for the PR!

modules/python/pylib/syslogng/debuggercli/__init__.py
```diff
@@ -21,11 +21,31 @@
...
+    return True
+
+def is_editline_available():
+    try:
+        import editline
+    except ImportError:
+        return False
+    return True
+
+def setup_read_or_editline():
```
szemere, line 39: Just a small note: I thought it would be nice to move `__setup_performed__` to this level, to avoid code duplication. So I started to tweak this part.
If none of the setup succeeded, we only lost tab completion, and everything else works fine! Which is really good.

szemere (approved):
Other than the extra import, approve from me. Thank You!

PR #3426, opened 2020-09-11
new filter template function, and fix if inside map

furiel:
## Feature
A new `filter` template function is introduced in syslog-ng. Given a list, it executes a filter expression on each element, keeping only those that satisfy the filter expression. You can use macros, logical expressions, template functions inside the expression.

The current value is referred by `$_`, similarly to the `map` template function.

### Syntax:
```
$(filter filter-expression list)
```

### Example
Filtering even numbers:
```
log {
  source { example-msg-generator(num(1) values(INPUT => "0,1,2,3")); };
  destination {
     file("/dev/stdout"
           template("$(filter ('$(% $_ 2)' eq '0') $INPUT)\n)")
     );
  };
};
```

results in `0,2`.

## Bugfix
As `context_id`-s weren't passed when evaluating `FilterExprNodes`, `if` template function could not use `$_`: it received NULL (`""`) instead.

For example, these configurations will work after this PR:
```
log {
  source { example-msg-generator(num(1) values(INPUT => "0,1,2,3")); };
  destination {
     file("/dev/stdout"
           template("$(map $(if ('$(% $_ 2)' eq '0') 'even' 'odd') $INPUT)'\n)")
     );
  };
};
```
with result: `even,odd,even,odd`.

Referring: https://github.com/syslog-ng/syslog-ng/issues/2926

CC: @jszigetvari


kira-syslogng:
Build SUCCESS

lib/filter/filter-expr.c
```diff
@@ -42,38 +42,41 @@ filter_expr_node_init_instance(FilterExprNode *self)
...
  */
 gboolean
-filter_expr_eval_with_context(FilterExprNode *self, LogMessage **msg, gint num_msg)
+filter_expr_eval_with_context(FilterExprNode *self, LogMessage **msg, gint num_msg, const gchar *context_id)
 {
   gboolean res;
 
   g_assert(num_msg > 0);
 
+  if (self->eval_with_context)
```
Kokan, line 51: This is kinda hard to deal with, some filter going to provide `eval_with_context` and some the same old `eval` ? What does it mean for an `filter_expr_eval_with_context` not to call `eval_with_context` ?

Providing an interface to implement for filter with eval and without is okay(if there is really a need for both), but I would put an extra layer of abstraction as `FilterWithContext` and `FilterWithoutContext` (probably not with these names).

I would prefer a different direction, where the `eval` signature is changed to support `context_id`.
1. As anybody who implemented a filter, now their code is broken, as it should be. This imho fixes a bug, the only way to fix that bug is to include and handle `context_id`.
2. How the context_id handled is pushed into one place(speaking of one filter), now it is in at least two different places (where it is decided `eval_with_context` or `eval` is called).
3. The `filter_expr_eval_with_context` is mostly glue code for oop, thus imho should be as trivial as it can be.

furiel: There are two kind of filters: those that need to evaluate inner template: for example the compare filter.
And there are those that don't: for example [netmask](https://github.com/syslog-ng/syslog-ng/blob/088fc40ef803939a4fd14a07c466ae10529fb0f5/lib/filter/filter-netmask.c#L40). It actually just accesses sockaddr of the message, and works with that.

That's why I did not go with option 1. I wanted to keep that distinction.

Conceptually, these two filter categories complement each other, still we want to use them at the same places: we are expecting polymorhism on object that have different api. This is a contradiction on it's own. In our specific case, we are in luck, because a context-aware filter can evaluate a non-aware inner filter. And this is what I implemented. This also has the benefit that the api of the non-context aware filters are not broken.

I am willing to improve the patch in any way, even break the entire filter eval api. I am just not sure how to do it without adding too much boilerplate.

bazsi: Hmm... I don't know how this would play out, but what I have in mind when talking about "context" is the combination of the list of messages as well as all information required to expand a template (e.g. the argument of the log_template_format_with_context()  function).

I have a very old branch lying around (from 2015) where this struct was introduced:
```
+typedef struct _LogTemplateEvalArgs
+{
+  /* context in case of correllation */
+  LogMessage **messages;
+  gint num_messages;
+
+  /* options for recursive template evaluation, inherited from the parent */
+  const LogTemplateOptions *opts;
+  gint tz;
+  gint seq_num;
+  const gchar *context_id;
+} LogTemplateEvalArgs;
```

These fields are pretty often passed around in function arguments, especially in the implementation of template functions.

With that said, maybe "context" is not at all template specific, rather it belongs a bit more to our message model, e.g. a LogMessage might be part of a context. In a number of cases where we pass data in function arguments, we should rather pass this structure to ensure we recursively pass around information that would otherwise be required.

Because we don't have this abstraction today, we stumble into cases where we forget one thing or the other.

Looking at the struct above, I would try to remove the LogTemplate specific things in it and then try to apply it into cases where we are invoking `*_with_context()` functions (e.g. filters, template functions, templates).

bazsi: This is the branch I mentioned: https://github.com/bazsi/syslog-ng/tree/wip/template-eval-arguments-as-a-struct-instead-of-parameters

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/filter/filter-expr.c
furiel: Thank you all for the comments and the branch. I updated the PR based on them. Removing WIP flag.

kira-syslogng:
Build SUCCESS

Kokan (approved):

alltilla (approved):

PR #3428, opened 2020-09-16
internal() source running in dedicated thread

Kokan:
The `internal()` source currently runs in the main thread, while the main thread main goal is to orchestrate the work and not doing it.

### DIFF
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -187,8 +187,6 @@ afinter_source_mark(gpointer s)
   AFInterSource *self = (AFInterSource *) s;
   struct timespec nmt;
 
-  main_loop_assert_main_thread();
-
   g_static_mutex_lock(&internal_mark_target_lock);
   nmt = next_mark_target;
   g_static_mutex_unlock(&internal_mark_target_lock);
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -125,6 +125,38 @@ afinter_source_post(gpointer s)
 }
 
 static void
+afinter_source_run(gpointer s)
+{
+  AFInterSource *self = (AFInterSource *) s;
+
+  iv_init();
+
+  iv_main();
+
+  iv_deinit();
+}
+
+static void
+afinter_source_request_exit(gpointer s)
+{
+  AFInterSource *self = (AFInterSource *) s;
+
+}
+
+static gboolean
+afinter_sd_start_thread(LogPipe *s)
+{
+  AFInterSourceDriver *self = (AFInterSourceDriver *) s;
+
+  main_loop_create_worker_thread((WorkerThreadFunc) afinter_source_run,
+                                 (WorkerExitNotificationFunc) afinter_source_request_exit,
+                                 self->source, NULL);
+
+  return TRUE;
+}
+
+
+static void
 afinter_source_mark(gpointer s)
 {
   AFInterSource *self = (AFInterSource *) s;
@@ -412,6 +444,7 @@ afinter_sd_new(GlobalConfig *cfg)
   self->super.super.super.init = afinter_sd_init;
   self->super.super.super.deinit = afinter_sd_deinit;
   self->super.super.super.free_fn = afinter_sd_free;
+  self->super.super.super.on_config_inited = afinter_sd_start_thread;
 
   afinter_source_options_defaults(&self->source_options);
 
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -353,7 +353,7 @@ afinter_source_deinit(LogPipe *s)
   iv_event_unregister(&self->schedule_wakeup);
 
   afinter_source_stop_watches(self);
-  return log_source_deinit(s);
+  return log_source_deinit(&self->super.super);
 }
 
 static LogSource *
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/afinter.c
```diff
@@ -122,16 +123,71 @@ afinter_source_post(gpointer s)
       log_source_post(&self->super, msg);
     }
   afinter_source_update_watches(self);
+  main_loop_worker_invoke_batch_callbacks();
```
gaborznagy, line 126: I've ran through the patches and they look good to me, but I wonder, why do we need to call `main_loop_worker_invoke_batch_callbacks`?

Kokan: That's a great question. It is not something that I know should be there (and possible cannot explain in 100% accuracy).

This new thread is a worker thread (`main_loop_create_worker_thread`), it processes messages from source till destination queue.
The current implementation of queue fifo has a fast and slow path, depending on if it was called from a worker thread or a simple thread.
When a simple thread is used it stuck with the slow path, otherwise from worker thread it can use the fast path with per thread queue.

The fast path at some point shall merge the per thread queues, the current solution that fast path registers a call back doing that job later. A worker thread has a framework to deal with these callbacks, but have no mean of calling it automatically (as it should be executed in the worker thread when it has time, but regularly). 

If you remove this call two things going to happen:
1. An assert guarding the worker thread shutdown with checking for pending jobs.
2. The messages actually won't be processed by destination (in case of logqueue fifio) as they are waiting for the merger from per thread queue to the queue a destination works from.

This is my current understanding here.

kira-syslogng:
Build FAILURE

lib/afinter.c
Kokan: Note to this, in the end I'v disabled fast path for this source. As the fast path currently entangled with the number of io worker thread, thread id assignment. The number of thread queue set based on the number of io worker. IMHO this is not right, but a separate issue instead of opening the scope I do what every other threaded src do currently and use slow path. (https://github.com/syslog-ng/syslog-ng/pull/3428/commits/e3c016ee6fbdd0cb3096af3f5d5577b3382e90fa)

It leads to crash otherwise, as the logqueue fifo assert on its own expected number of queue and the current thread value.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lib/afinter.c
gaborznagy: Thanks for the explanation. I saw that there is only one place where `main_loop_worker_register_batch_callback` is called, and it was `logqueue-fifo.c`. I've sensed a rabbit hole, and you proved it. :)
It's an interesting topic though.

kira-syslogng:
Build SUCCESS

Kokan:
Only rebase to follow up macos ci fix on master, no code change since last push.

lib/afinter.c
```diff
@@ -124,14 +125,70 @@ afinter_source_post(gpointer s)
...
+  self->free_to_send = TRUE;
+  g_static_mutex_unlock(&internal_msg_lock);
+
+  afinter_source_start_watches(self);
+  afinter_source_update_watches(self);
+
+  iv_main();
+
+  iv_event_unregister(&self->exit);
+  iv_event_unregister(&self->post);
```
MrAnno, line 154: I have a crash because of a possible race, where an internal message triggers the post event, but the it has previously been unregistered.

I'm investigating this scenario.

MrAnno (changes requested):

lib/cfg.c
```diff
@@ -368,7 +368,9 @@ cfg_init(GlobalConfig *cfg)
   if (!cfg_tree_start(&cfg->tree))
     return FALSE;
 
-  g_assert(cfg_tree_on_inited(&cfg->tree));
+  if (!cfg_tree_on_inited(&cfg->tree))
+    return FALSE;
```
MrAnno, line 372: Please keep the assertion here. It is intentional, see https://github.com/syslog-ng/syslog-ng/pull/3176#issuecomment-638849597.

TLDR: A half-initialized pipeline turned out to be really hard to deinitialize correctly when dedicated source/destination threads are spawned (because we would have to wait for workers to stop and guarantee some internal task/timer/fdwatch ordering in ivykis during this action).

The best would be to extract these 2 lines or just add a comment.

kira-syslogng:
Build SUCCESS

lib/cfg.c
Kokan: Thanks, I add your comment, and a link to the original comment, those should be a nice point to unroll the shenanigans of this assert.

kira-syslogng:
Build SUCCESS

MrAnno (approved):
Thanks!

Kokan:
Change: the `on_config_init` was initialised in source driver init function, instead of source driver new. (Thanks for @gaborznagy noticing it.)

kira-syslogng:
Build SUCCESS

gaborznagy (approved):
Thanks for the changes!

MrAnno (approved):

gaborznagy:
LGTM has some internal issues around APT, I'm merging this anyway (previously it was green).
```
Could not connect to archive.ubuntu.com:80 (91.189.88.142), connection timed out Could not connect to archive.ubuntu.com:80 (91.189.88.152), connection timed out
```

PR #3434, opened 2020-09-23
nvtable: use atomic refcount

furiel:
nv_table_serialize can be called from destinations, for example in
LogStore. In destination (logwriter) context, this means these
counters are changed from multiple threads, so they need to be
protected. This patch turns ref_cnt to an atomic counter.

As there are no destinations that serialize messages in ose,
no changenote is necessary.

### DIFF
--- a/lib/logmsg/nvtable.c
+++ b/lib/logmsg/nvtable.c
@@ -674,7 +674,7 @@ nv_table_init(NVTable *self, gsize alloc_length, gint num_static_entries)
   self->used = 0;
   self->index_size = 0;
   self->num_static_entries = num_static_entries;
-  self->ref_cnt = 1;
+  g_atomic_counter_set(&self->ref_cnt, 1);
   self->borrowed = FALSE;
   memset(&self->static_entries[0], 0, self->num_static_entries * sizeof(self->static_entries[0]));
 }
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

Kokan (approved):

lib/logmsg/nvtable.c
```diff
@@ -718,7 +718,7 @@ nv_table_realloc(NVTable *self, NVTable **new)
   if (new_size == old_size)
     return FALSE;
 
-  if (self->ref_cnt == 1 && !self->borrowed)
+  if (g_atomic_counter_get(&self->ref_cnt) == 1 && !self->borrowed)
```
lbudai, line 721: Is it possible that more thread runs this code path concurrently?

furiel: nv_table_realloc is called when the message is modified: new value pairs are added or removed. Sources and destinations do not share logmessage/nvtable, the rest of the logpipes are protected by logmessage write protection and COW. So this line is safe.

Borrowed is only changed during `nv_table_realloc`, and `nv_table_clone`. `nv_table_clone` is also called only when the message is changed. So borrowed does not need to be protected against concurrent access.

lbudai: okay, thanks.

lbudai (approved):

bazsi:
Hmm, did you run a performance test on this change?

furiel:
@kira-syslogng do perftest

Issue #3418, opened 2020-09-07
systemd-journal compile problem on CentOS 7


czanik:
Both when using ```dbld``` and when using ```Copr``` to build the latest git master snapshot for CentOS 7, it fails:

```
  CC       modules/systemd-journal/modules_systemd_journal_libsdjournal_la-systemd-journal.lo
  CC       modules/systemd-journal/modules_systemd_journal_libsdjournal_la-systemd-journal-grammar.lo
  CC       modules/systemd-journal/modules_systemd_journal_libsdjournal_la-systemd-journal-parser.lo
  CC       modules/systemd-journal/modules_systemd_journal_libsdjournal_la-systemd-journal-plugin.lo
  CC       modules/systemd-journal/modules_systemd_journal_libsdjournal_la-journal-reader.lo
  CC       modules/systemd-journal/modules_systemd_journal_libsdjournal_la-journald-subsystem.lo
modules/systemd-journal/journal-reader.c: In function '_journal_process_namespace':
modules/systemd-journal/journal-reader.c:594:25: error: 'SD_JOURNAL_ALL_NAMESPACES' undeclared (first use in this function)
       *journal_flags |= SD_JOURNAL_ALL_NAMESPACES;
                         ^
modules/systemd-journal/journal-reader.c:594:25: note: each undeclared identifier is reported only once for each function it appears in
modules/systemd-journal/journal-reader.c:602:25: error: 'SD_JOURNAL_INCLUDE_DEFAULT_NAMESPACE' undeclared (first use in this function)
       *journal_flags |= SD_JOURNAL_INCLUDE_DEFAULT_NAMESPACE;
                         ^
make[2]: *** [modules/systemd-journal/modules_systemd_journal_libsdjournal_la-journal-reader.lo] Error 1
make[2]: *** Waiting for unfinished jobs....
modules/systemd-journal/journald-subsystem.c: In function 'journald_open_namespace':
modules/systemd-journal/journald-subsystem.c:200:3: warning: implicit declaration of function 'sd_journal_open_namespace' [-Wimplicit-function-declaration]
   return sd_journal_open_namespace(&self->journal, namespace, flags);
   ^
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

You can check the whole build process at: https://download.copr.fedorainfracloud.org/results/czanik/syslog-ng-githead/epel-7-x86_64/01651082-syslog-ng/builder-live.log.gz

[...]

PR #3435, opened 2020-09-24
Fix journald namespaces detection on centos-7

gaborznagy:
With older pkg-config versions (0.27.1 on centos-7) PKG_CONFIG_CHECK_MODULES() macro doesn't check the library's version on multiple calls (if they have the same prefix).
It seems to be that the macro is idempotent.

With version 0.29.1 it is allowed to call the macro multiple times with the same prefix.

See config.log on centos-7 with pkg-config  version 0.27.1, on the second call it simply returns true.
```
configure:18300: checking for libsystemd
configure:18307: $PKG_CONFIG --exists --print-errors "libsystemd >= ${LIBSYSTEMD_MIN_VERSION}"
configure:18310: $? = 0
configure:18324: $PKG_CONFIG --exists --print-errors "libsystemd >= ${LIBSYSTEMD_MIN_VERSION}"
configure:18327: $? = 0
configure:18365: result: yes
configure:18372: checking for libsystemd
configure:18437: result: yes
```

Alternatively, this could be solved without additional PKG_CHECK_MODULES() calls on libsystemd,
with calling $PKG_CONFIG --modversion directly if we detected libsystemd in the first place.

Fixes: #3418

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1407,7 +1407,7 @@ dnl ***************************************************************************
 PKG_CHECK_MODULES(libsystemd, libsystemd >= ${LIBSYSTEMD_MIN_VERSION},
                   have_libsystemd="yes", have_libsystemd="no")
 
-PKG_CHECK_MODULES(libsystemd, libsystemd >= ${LIBSYSTEMD_WITH_JOURNAL_NAMESPACES_MIN_VERSION},
+PKG_CHECK_MODULES(libsystemd_namespaces, libsystemd >= ${LIBSYSTEMD_WITH_JOURNAL_NAMESPACES_MIN_VERSION},
                   journal_namespaces="yes", journal_namespaces="no")
 
 dnl ***************************************************************************
--- a/dbld/packages.manifest
+++ b/dbld/packages.manifest
@@ -91,7 +91,7 @@ python-devel                [centos, fedora]
 python3-devel               [fedora]
 python3-dev                 [ubuntu-focal]
 python-pip                      [centos, fedora, debian-stretch, debian-buster, ubuntu-trusty, ubuntu-xenial, ubuntu-bionic]
-python3-pip                     [centos, debian, ubuntu]
+python3-pip                     [debian, ubuntu]
 
 # libmongoc and libbson packages on various platforms
 # Because we are using fixed version of libmongoc on Bionic, we need to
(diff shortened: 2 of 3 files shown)


gaborznagy:

kira-syslogng:
Package Build FAILURE

kira-syslogng:
Build SUCCESS

gaborznagy:
building systemd-journal works fine, but now python/editline fails :cry: :cry: :cry:

```
+ /usr/lib/rpm/check-buildroot
+ /usr/lib/rpm/redhat/brp-compress
+ /usr/lib/rpm/redhat/brp-strip-static-archive /usr/bin/strip
+ /usr/lib/rpm/brp-python-bytecompile /usr/bin/python 1
Compiling /home/jenkins/rpmbuild/BUILDROOT/syslog-ng-3.29.1.77.g7f0f773-1+20200924T103720.x86_64/usr/lib64/syslog-ng/python/syslogng/debuggercli/editline.py ...
  File "/usr/lib64/syslog-ng/python/syslogng/debuggercli/editline.py", line 51
    def __init__(self, subeditor, completer, namespace: dict = None):
                                                      ^
SyntaxError: invalid syntax



RPM build errors:
error: Bad exit status from /var/tmp/rpm-tmp.v8caJp (%install)
    Bad exit status from /var/tmp/rpm-tmp.v8caJp (%install)
dbld/rules:125: recipe for target 'package-centos-7' failed
make: *** [package-centos-7] Error 1
```

I'll try to find the reason.

Kokan:
@gaborznagy Type hint is 3.5 feature: https://docs.python.org/3/library/typing.html
But during package build (rpm) tries to compile python to byte code not respecting python version and uses OS default: python2.

This failure is not due to your change, so I think after review this is okay to merge, and the above error should be fixed in a new PR.

gaborznagy:
Thanks for the tip, I didn't dig into it yet. I've suspected some python2-python3 issue.

I'll push the fix here, if it's not merged until I found the best resolution to the problem.

gaborznagy:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Package Build SUCCESS

Packages are available at https://ci.syslog-ng.com/jenkins/job/kira-dbld-pull-requests/711/artifact/

gaborznagy:
I've fixed the rpm building error, by simply removing the type-hints from debuggercli/editline.py, as they are python3 only (no [`__future__`](https://docs.python.org/2/library/__future__.html) for that.

I didn't want this solution, as pyeditline is only available on python3 and it would have been nicer to create some conditionals to only put editline.py into the tarball on python3, but this was faster and since only a partial of the code had type hints, it didn't seem a huge backward step.

Kokan (approved):

alltilla (approved):
Tested with `dbld/rules release`, it works now :)

PR #3436, opened 2020-09-25
nv_table_serialize: do not nv_table_ref

furiel:
This patch reverts: https://github.com/syslog-ng/syslog-ng/pull/3434.

As it turns out `nv_table_ref` is not necessary during `nv_table_serialize`: we have a better resolution of the problem in #3434 

Changenote is not needed. This is just a fixup for #3434

### DIFF
--- a/lib/logmsg/logmsg-serialize.c
+++ b/lib/logmsg/logmsg-serialize.c
@@ -45,6 +45,15 @@ _setup_ts_processed(UnixTime *timestamps, const UnixTime *processed)
     timestamps[LM_TS_PROCESSED] = timestamps[LM_TS_RECVD];
 }
 
+static void
+nv_table_serialize_with_compaction(LogMessageSerializationState *state, NVTable *old)
+{
+  NVTable *payload;
+  payload = nv_table_compact(old);
+  nv_table_serialize(state, payload);
+  nv_table_unref(payload);
+};
+
 static gboolean
 _serialize_message(LogMessageSerializationState *state)
 {
@@ -70,15 +79,10 @@ _serialize_message(LogMessageSerializationState *state)
   serialize_write_uint8(sa, msg->alloc_sdata);
   serialize_write_uint32_array(sa, (guint32 *) msg->sdata, msg->num_sdata);
 
-  NVTable *payload;
-
   if (state->flags & LMSF_COMPACTION)
-    payload = nv_table_compact(msg->payload);
+    nv_table_serialize_with_compaction(state, msg->payload);
   else
-    payload = nv_table_ref(msg->payload);
-
-  nv_table_serialize(state, payload);
-  nv_table_unref(payload);
+    nv_table_serialize(state, msg->payload);
   return TRUE;
 }
 
--- a/lib/logmsg/nvtable-serialize-legacy.c
+++ b/lib/logmsg/nvtable-serialize-legacy.c
@@ -316,7 +316,7 @@ nv_table_deserialize_22(SerializeArchive *sa)
   if(!res)
     return NULL;
 
-  g_atomic_counter_set(&res->ref_cnt, 1);
+  res->ref_cnt = 1;
   res->borrowed = FALSE;
 
   if (!_deserialize_struct_22(sa, res))
@@ -453,7 +453,7 @@ nv_table_deserialize_legacy(SerializeArchive *sa)
     return NULL;
 
   res->borrowed = FALSE;
-  g_atomic_counter_set(&res->ref_cnt, 1);
+  res->ref_cnt = 1;
 
   if (!_deserialize_blob_v22(sa, res, nv_table_get_top(res), swap_bytes))
     {
--- a/lib/logmsg/nvtable-serialize.c
+++ b/lib/logmsg/nvtable-serialize.c
@@ -150,7 +150,7 @@ _read_header(SerializeArchive *sa, NVTable **nvtable)
     goto error;
 
   res->borrowed = FALSE;
-  g_atomic_counter_set(&res->ref_cnt, 1);
+  res->ref_cnt = 1;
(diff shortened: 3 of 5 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
I don't understand the reasoning behind this pr, can you pls elaborate?

Thanks

furiel:
As I understand, the nvpairs in the hashtable are located at the back, at the the [(`self->size`-`self->used`), (self->size)] interval. In the previous PR, I increased the header part, however, size and used remained the same. If `self->used` is large enough, the deserialized data overlap with the header part, corrupting the nvtable. To counter that, I need to allocate a larger memory area, so that I can safely deserialize the nvpairs to the end.

furiel:
Forgot to mention this only causes problem if you want to deserialize a message that was serialized with the old struct. Serializing with the extended struct, and deserializing does not cause problem. The bug is about backward compatibility.

bazsi:
I don't see how you increased the header part....

```
static void
_write_struct(SerializeArchive *sa, NVTable *self)
{
  serialize_write_uint32(sa, self->size);
  serialize_write_uint32(sa, self->used);
  serialize_write_uint16(sa, self->index_size);
  serialize_write_uint8(sa, self->num_static_entries);
  serialize_write_uint32_array(sa, self->static_entries, self->num_static_entries);
  serialize_write_uint32_array(sa, (guint32 *) nv_table_get_index(self), self->index_size * 2);
}

```

This function has not changed in the last PR.

furiel:
> I don't see how you increased the header part....

Yes this was what confused me too. `_write_struct` already skipped `ref_cnt` and `borrowed`, so I was thinking there will be no need for conversion.

The trick is the start of the nvpairs is not calculated by `offsetof`, as one might expect. It is directly calculated from `used` and `self`.

```
static inline gchar *
nv_table_get_top(NVTable *self)
{
  return NV_TABLE_ADDR(self, self->size);
}

static inline gchar *
nv_table_get_bottom(NVTable *self)
{
  return nv_table_get_top(self) - self->used;
}
```

Deserialization starts by allocating `size` sized area, and casts it to `NVTable *` pointer. After reading back the headers, it will copy nvpairs: 

```
static inline gboolean
_read_payload(SerializeArchive *sa, NVTable *res)
{
  return serialize_read_blob(sa, NV_TABLE_ADDR(res, res->size - res->used), res->used);
}
```
which fit for the original `NVTable`, but I added new bytes to the front in the new one. So there can be overlap.

Something like this:

```
    <--- header --------> <----nv pairs ------->

   |---------------------|----------------------|     OLD NVTABLE
    size               size-used               size
    used

                        || (serialize)
                        \/

   (serialized data: size = size, used = used)

                        || deserialize:
                        || malloc: size
                        || fill the headers
                        || copy nvdata between [(size-used), size]
                        \/

                          new end of headers (+ 4 byte)
                               \ /
   |---------------------|------|---------------|     NEW NVTABLE
                    size - used               size
             (original end of headers)

   <----new header structure---->
                        |overlap|
                         <----- nv pairs ------->
```

MrAnno:
`GAtomicCounter` is not a fixed-size field, it has a `gint` inside. I think it's not portable if `self->size` is serialized and its value contains the size of the ref count.

bazsi:
I think it might make sense to fix this now if we are changing the header
version anyway. We could fix this by encapsulating the header into a fixed
size (e.g. by using a union in the NVTable struct) and check that the sum
of fields is fits into the fixed size array.

lbudai:
> `GAtomicCounter` is not a fixed-size field, it has a `gint` inside. I think it's not portable if `self->size` is serialized and its value contains the size of the ref count.
> 
> Can we extract all runtime-only fields into a separate struct that embeds the one that is serializable? I guess some additional calculation will be needed then to get the size of the runtime struct, or just calculate everything based on the starting address of the embedded struct.



Currently this could be done within a reasonable time frame (this problem is blocking some internal release process). On the other hand I don't think this would be the final solution - but at least revertable. If we cannot find a better solution within a timebox , I'd suggest to open a github issue for the problem (with the whole context) and merge this PR (with the changes suggested by @MrAnno) .

kira-syslogng:
Build SUCCESS

furiel:
It turned out `nv_table_ref` is not needed at all, so we can have a better fix for the bug. I updated the PR, and the description.

lib/logmsg/logmsg-serialize.c
```diff
@@ -45,6 +45,15 @@ _setup_ts_processed(UnixTime *timestamps, const UnixTime *processed)
     timestamps[LM_TS_PROCESSED] = timestamps[LM_TS_RECVD];
 }
 
+static void
+nv_table_serialize_with_compaction(LogMessageSerializationState *state, NVTable *old)
+{
+  NVTable *payload;
+  payload = nv_table_compact(old);
+  nv_table_serialize(state, payload);
+  nv_table_unref(payload);
```
Kokan, line 54: This on the other hand means possible leaking the payload when compact is turned on.
e.g.: two thread sees `ref_cnt=2`, does unref and `ref_cnt` became 1 so no free is "needed".

Despite this I would merge it as is, as fixes a crash compared to a leak (that is present in the prev release anyway, and the feature triggering it is turned off by default).

Kokan: Oh sorry my bad, payload here is the new `NVTable`, local variable so no issue here.

Kokan (approved):

lib/logmsg/logmsg-serialize.c
lbudai: maybe we should rename `nv_table_compact` to `nv_table_clone_compact` ...

lbudai (approved):

PR #3437, opened 2020-09-25
Proxy protocol

lbudai:
Add [Proxy protocol](http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) support to syslog-ng.
Originally implemented by Ferenc Sipos.

Changes:
 * source driver support removed
 * TLS support added
 * parser reimplemented

There are some known issues, so it's a draft PR.

<details>

```
@version: 3.29

source s_tcp_pp {
    network(
        port(7777)
#        transport("proxied-tcp")
        transport("proxied-tls")
        tls(
             key-file("/openssl/certs/certs/server/server.rsa")
             cert-file("openssl/certs/certs/server/server.crt")
             ca-dir("/openssl/certs/certs/CA")
#             peer-verify("optional-untrusted")
             peer-verify("required-trusted")
         )
    );
};

destination d_file {
    file("/var/log/pp.log" template("$(format-json --scope nv-pairs)\n"));
};

log {
    source(s_tcp_pp);
    destination(d_file);
};

```

</details>


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/transport/transport-aux-data.h
```diff
@@ -31,7 +31,7 @@ typedef struct _LogTransportAuxData
   GSockAddr *peer_addr;
   GSockAddr *local_addr;
   gint proto;
-  gchar data[1024];
+  gchar data[1536];
```
MrAnno, line 34: Based on how aux data is stored, `proxied-tcp` can increase aux-data with the following amount:

```
PROXIED_SRCIP - max 64 bytes
PROXIED_DSTIP - max 64 bytes
PROXIED_SRCPORT - max 8 bytes
PROXIED_DSTPORT - max 8 bytes
PROXIED_IP_VERSION - max 8 bytes
```

Storage size: `name_len + 1 + value_len + 1 + 1`

```
13 + 1 + 64 + 1 + 1 +
13 + 1 + 64 + 1 + 1 +
15 + 1 + 8 + 1 + 1 +
15 + 1 + 8 + 1 + 1 +
18 + 1 + 8 + 1 + 1 =

241
```

👍🏻

lib/logproto/logproto-proxied-text-server.c
```diff
@@ -0,0 +1,156 @@
...
+}
+
+static void
+_log_proto_proxied_text_server_free(LogProtoServer *s)
+{
+    LogProtoProxiedTextServer *self = (LogProtoProxiedTextServer *) s;
+
+    msg_debug("Freeing PROXY protocol source driver", evt_tag_printf("driver", "%p", self));
+
+    g_free((gpointer) self->info);
```
MrAnno: The cast is not necessary.

lib/logproto/logproto-proxied-text-server.c
```diff
@@ -0,0 +1,156 @@
...
+    return FALSE;
+}
+
+static LogProtoStatus
+_log_proto_proxied_text_server_fetch(LogProtoServer *s, const guchar **msg, gsize *msg_len, gboolean *may_read,
+                                     LogTransportAuxData *aux, Bookmark *bookmark)
+{
+    LogProtoProxiedTextServer *self = (LogProtoProxiedTextServer *) s;
+
+    LogProtoStatus status = self->fetch_from_buffer(&self->super.super.super, msg, msg_len, may_read, aux, bookmark);
```
MrAnno: I don't see how this works.
This method replaces `log_proto_buffered_server_fetch`, but does not call that method, so the input buffer will never be filled:

```
self->super.super.super.fetch = _log_proto_proxied_text_server_fetch;
```

----

Update: Oh I see, `fetch_from_buffer` is actually a pointer to `log_proto_buffered_server_fetch`, not to `log_proto_text_server_fetch_from_buffer` as I thought.

Can we publish `log_proto_buffered_server_fetch()` and call it directly, so there will be no confusion (or just rename this cached pointer to `fetch`)?

lib/logproto/logproto-proxied-text-server.c
```diff
@@ -0,0 +1,156 @@
...
+
+static gboolean
+_log_proto_proxied_text_server_handshake_in_progress(LogProtoServer *s)
+{
+    LogProtoProxiedTextServer *self = (LogProtoProxiedTextServer *) s;
+
+    // Run handshake() only once
+    if (!self->handshake_done)
+    {
+        self->handshake_done = TRUE;
```
MrAnno: I think the flag should be set in the handshake method. This method becomes a simple getter this way.

lib/logproto/logproto-proxied-text-server.c
```diff
@@ -68,7 +68,11 @@ _log_proto_proxied_text_server_handshake(LogProtoServer *s)
   status = self->fetch_from_buffer(&self->super.super.super, &msg, &msg_len, &may_read, NULL, NULL);
 
   if (status != LPS_SUCCESS)
-    return status;
+    {
+      if (status == LPS_AGAIN)
+        self->handshake_done = FALSE;
```
MrAnno, line 96: I may be mistaken, but my understanding is that if we set `handshake_done` to TRUE only if we are really done with the handshake (we've extracted a full line that contains the proxy information or `PROXY UNKNOWN`), then `LPS_AGAIN` would not be needed here.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

lib/logproto/logproto-proxied-text-server.c
```diff
@@ -0,0 +1,156 @@
...
+#define PROXY_UNKNOWN_HDR "PROXY UNKNOWN"
+
+static gboolean
+_log_proto_proxied_text_server_parse_header(LogProtoProxiedTextServer *self, const guchar *msg, gsize msg_len)
+{
+    GString *input = g_string_new_len(msg, msg_len);
+    int matches = 0;
+    int compares = -1;
+
+    matches = sscanf(input->str, PROXY_TCP_HDR, &self->info->ip_version, self->info->src_ip, self->info->dst_ip,
```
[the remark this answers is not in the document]
lbudai: check the final version, parsing is modified

lib/logproto/logproto-proxied-text-server.c
lbudai: indeed

lib/logproto/logproto-proxied-text-server.c
lbudai: done (it is more readable, thanks)

kira-syslogng:
Build SUCCESS

lib/logproto/logproto-proxied-text-server.c
lbudai: it's not that simple - the method is called even when a TLS handshake is in progress, when this is the case we cannot set handhsake_done to TRUE

kira-syslogng:
Build SUCCESS

lib/logproto/logproto-proxied-text-server.c
lbudai: one additional note: handshake_done is TRUE when we've received a string that should be parsed as a header
it is possible that the handshake will fail - in that case handshake_done is set to TRUE (and connection will be closed...)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/logproto/logproto-proxied-text-server.c
```diff
@@ -0,0 +1,290 @@
...
+  GString *params_str = g_string_new_len((const gchar *)msg, msg_len);
+  gboolean result = FALSE;
+  msg_debug("PROXY header params", evt_tag_str("params", (const gchar *)msg));
+
+  gchar **params = strsplit(params_str->str, ' ', 5);
+  gint params_n = g_strv_length(params);
+  if (params_n < 4)
+    goto ret;
+
+  strncpy(self->info->src_ip, params[0], IP_BUF_SIZE - 1);
```
MrAnno, line 116: `strncpy` will not null-terminate `src_ip` in case `params[0]` is longer than `IP_BUF_SIZE - 1`, but the `info` struct is created with `g_new0`, so there will be an implicit \0 at the end of each buffer, which is perfect. 👍🏻

lib/logproto/logproto-proxied-text-server.c
```diff
@@ -0,0 +1,290 @@
...
+  gboolean may_read = TRUE;
+  gboolean parsable;
+  LogProtoStatus status;
+
+  LogProtoProxiedTextServer *self = (LogProtoProxiedTextServer *) s;
+
+  // Fetch a line from the transport layer
+  status = log_proto_buffered_server_fetch(&self->super.super.super, &msg, &msg_len, &may_read, NULL, NULL);
+
+  self->handshake_done = (status == LPS_SUCCESS);
```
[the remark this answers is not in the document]
lbudai: Yes, I'd like to keep that.
I know that this part of the PR is far from being finished and any improvements that move things forward are warmly welcomed, thanks!

lib/logproto/logproto-proxied-text-server.c
```diff
@@ -0,0 +1,290 @@
...
+  return _check_header(msg, msg_len, PROXY_HDR_UNKNOWN, header_len);
+}
+
+static gboolean
+_parse_unknown_header(LogProtoProxiedTextServer *self, const guchar *msg, gsize msg_len)
+{
+  if (msg_len == 0)
+    return TRUE;
+
+  GString *params_str = g_string_new_len((const gchar *)msg, msg_len);
```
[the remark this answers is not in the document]
lbudai: yes, and this allocation is not needed... will fix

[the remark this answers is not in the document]
lbudai: diff: 
```
-  GString *params_str = g_string_new_len((const gchar *)msg, msg_len);
-  msg_warning("PROXY UNKNOWN header contains unexpected paramters", evt_tag_str("parameters", params_str->str));
+  msg_warning("PROXY UNKNOWN header contains unexpected paramters",
+              evt_tag_printf("parameters", "%.*s", (gint) msg_len, msg));
+

```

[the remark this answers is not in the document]
MrAnno: Thanks!

lib/logproto/logproto-proxied-text-server.c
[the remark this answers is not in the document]
MrAnno: Thanks!

MrAnno (approved):

kira-syslogng:
Build SUCCESS

alltilla (approved):

PR #3440, opened 2020-09-28
autotools: support sanitizers

Kokan:
This is an attempt to popularise sanitiser as default option for development.
As of now only cmake supported sanitiser, as a developper using autotools might had harder time to configure and so set as a "default" configure option for themself.

The option is `--with-sanitizer=address/thread/etc`.

This option not a drop in replacement for valgrind, both valgrind and sanitizer cover leaks and memory violations but they won't cover almost the same subset of issues  (but neither of them a subset of the other).

The benefit of using sanitiser opposed to valgrind that sanitizer built into the binary and such it is always on. No need for extra step to check for leaks as well as it is done every time syslog-ng & tests are started.
 
The current downside is that current syslog-ng violates the odr(one definition rulel) defining the same global `module_info` in the modules (which is not likely to change) as the option should be disabled with setting the following environment varialbe `ASAN_OPTIONS="detect_odr_violation=0"` (Note: this is "documented" in the makefile as assigned to tests).

The `LOG_COMPILER` option is needed to overcome a Criterion bug, that it won't report a test failed based on the executed testcase exit status. The same solution already exists in CMake with a difference that CMake by defualt supports regex pattern to fail a test. In the long run it should be removed once the Criterion bug is fixed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -26,7 +26,8 @@ DIST_SUBDIRS		=
 AM_MAKEFLAGS		= --no-print-directory
 AM_YFLAGS		= -Wno-yacc -Wno-other -Werror=conflicts-sr -Werror=conflicts-rr
 
-AM_TESTS_ENVIRONMENT	= top_srcdir="$(top_srcdir)" CRITERION_TEST_PATTERN='!(*/*performance*)'
+AM_TESTS_ENVIRONMENT	= top_srcdir="$(top_srcdir)" CRITERION_TEST_PATTERN='!(*/*performance*)' ASAN_OPTIONS="detect_odr_violation=0"
+LOG_COMPILER            = $(top_srcdir)/scripts/test-grep.sh
 
 ACLOCAL_AMFLAGS		= -I m4 --install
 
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

alltilla:
I'm okay with the `LOG_COMPILER` workaround.
Ran the sanitizer, it produced 14 UT failures. I might try to fix some of them in the near future :)

alltilla (approved):

Kokan:
That would be great, please before you start to work on them check out my branch (from time to time I have fixes there not yet upstreamed): https://github.com/Kokan/syslog-ng/commits/address-sanatize the branch also has commits that are likely not correct, but keep tracks of work in progress findings.; or just ping me if I have something for a test you would like to fix.

gaborznagy (approved):
Nice one! Thanks for adding this to autotools too!

For reference, the `LOG_COMPILER` documentation:
https://www.gnu.org/software/automake/manual/html_node/Parallel-Test-Harness.html#Parallel-Test-Harness

PR #3441, opened 2020-09-29
Light: add loggen support

norberttak:
Add the possibility to use loggen from Light scripts.

Typical use:
loggen.start("127.0.0.1", 6666, tcp=True, inet=True, rate=1)


kira-syslogng:
Build SUCCESS

alltilla:
I would rather have something like:
```
loggen.start("127.0.0.1", 6666, tcp=True, inet=True, rate=1)
```

kira-syslogng:
Build SUCCESS

tests/python_functional/src/helpers/loggen/conftest.py
```diff
@@ -0,0 +1,172 @@
...
+from src.syslog_ng.syslog_ng_paths import SyslogNgPaths
+
+
+class Loggen(object):
+
+    def __init__(self, testcase_parameters):
+        self.loggen_proc = None
+        self.target = None
+        self.port = None
+        self.parameters = None
```
alltilla, line 39: I think `target`, `port` and `parameters` should be in the `start()` function scope, and they are not needed in `self`

tests/python_functional/src/helpers/loggen/conftest.py
```diff
@@ -0,0 +1,172 @@
...
+        self.loggen_stdout_path = Path(tc_parameters.WORKING_DIR, "loggen_stdout")
+        self.loggen_stderr_path = Path(tc_parameters.WORKING_DIR, "loggen_stderr")
+        self.instance_path = SyslogNgPaths(testcase_parameters).set_syslog_ng_paths("server")
```
alltilla, line 42: Is `.set_syslog_ng_paths("server")` needed?

tests/python_functional/src/helpers/loggen/conftest.py
```diff
@@ -0,0 +1,172 @@
...
+        try:
+            self.loggen_proc.wait(4)
+        except TimeoutExpired:
+            self.loggen_proc.kill()
+
+        self.loggen_proc = None
+
+
+@pytest.fixture
+def loggen(testcase_parameters):
```
alltilla, line 86: As a fixture, the TC (and possibly different TCs running in parallel) can only start 1 `Loggen` at once. We might not want to use it as a fixture.

tests/python_functional/src/helpers/loggen/conftest.py
norberttak: do you know a better way to obtain the path of syslog-ng instance? That's the only one way I find but maybe I missed something here.

tests/python_functional/src/helpers/loggen/conftest.py
norberttak: in the current form we can start only one instance. it won't be a big work to make it possible to start multi instances. for this we have to prepare the stop function to stop a process with the given process id.

please note: the stop function is not called from test case scripts directly so the process ids can be stored internally in the loggen/conftest.py file. the stop shall iterate over all the internally stored process ids.

I'll do it later

tests/python_functional/src/helpers/loggen/conftest.py
norberttak: removed from the self structure

norberttak:
> I would rather have something like:
> 
> ```
> loggen.start("127.0.0.1", 6666, tcp=True, inet=True, rate=1)
> ```

tests/python_functional/src/helpers/loggen/conftest.py
alltilla: This might work:
```
import src.testcase_parameters.testcase_parameters as tc_parameters
...
self.loggen_bin_path = tc_parameters.INSTANCE_PATH.get_loggen_bin()
```
At least, this is how we do it at `syslog-ng-ctl`: https://github.com/syslog-ng/syslog-ng/blob/master/tests/python_functional/src/syslog_ng_ctl/driver_stats_handler.py#L37

kira-syslogng:
Build SUCCESS

tests/python_functional/src/helpers/loggen/conftest.py
```diff
@@ -0,0 +1,160 @@
...
+        if udp is True:
+            start_parameters.append("--dgram")
+
+        if use_ssl is True:
+            start_parameters.append("--use-ssl")
+
+        if dont_parse is True:
+            start_parameters.append("--dont-parse")
+
+        if read_file is True:
```
furiel, line 67: This should be `is not None`, similarly to the one below.

kira-syslogng:
Build SUCCESS

tests/python_functional/src/helpers/loggen/conftest.py
alltilla: Or the users can import `Loggen` and create multiple objects from it. It is perfect in this form. Thanks!

tests/python_functional/src/helpers/loggen/conftest.py
alltilla: @furiel: I think we don't want to add the flag, if `read_file=False` is set. `is not None` is true for `False` I believe.

tests/python_functional/src/helpers/loggen/conftest.py
```diff
@@ -0,0 +1,160 @@
...
+            idle_connections, ipv6, debug, number, csv, quiet,
+        )
+
+        self.loggen_proc = ProcessExecutor().start(
+            [self.loggen_bin_path] + self.parameters + [target, port],
+            self.loggen_stdout_path,
+            self.loggen_stderr_path,
+        )
+
+        return self.loggen_proc
```
alltilla, line 141: What happens, when the `loggen` run is finished by its own? Can we run it again? I think we cannot, because `self.loggen_proc` is not set to None.

norberttak: with the latest commits it is OK. you can call the start again and it will start another instance of loggen.

kira-syslogng:
Build SUCCESS

alltilla (approved):
Thanks!

kira-syslogng:
Build SUCCESS

tests/python_functional/src/helpers/loggen/conftest.py
```diff
@@ -0,0 +1,187 @@
...
+
+import src.testcase_parameters.testcase_parameters as tc_parameters
+from src.executors.process_executor import ProcessExecutor
+
+
+class Loggen(object):
+
+    instanceIndex = -1
+    @staticmethod
+    def get_new_instance_index():
```
alltilla, line 35: This should start with `_` to make it "private".

tests/python_functional/src/helpers/loggen/conftest.py
```diff
@@ -0,0 +1,187 @@
...
+        end_pattern = ", "
+        if end_pattern not in content:
+            return 0
+        index_end = content.find(end_pattern, index_start)
+
+        return int(content[index_start:index_end])
+
+
+@pytest.fixture
+def loggen(testcase_parameters):
```
alltilla, line 184: If the fixture is here, the testcases does not get auto-completion for `loggen`. I moved it to `src/conftest.py` and it seemed to work for me.

tests/python_functional/src/helpers/loggen/conftest.py
```diff
@@ -0,0 +1,187 @@
...
+            self.loggen_proc.kill()
+
+        self.loggen_proc = None
+
+    def get_sent_message_count(self):
+        if not self.loggen_stderr_path.exists():
+            return 0
+
+        # loggen puts the count= messages to the stderr
+        f = open(self.loggen_stderr_path, "r")
```
alltilla, line 166: `str(self.loggen_stderr_path)` was needed for me.

tests/python_functional/src/helpers/loggen/conftest.py
Kokan: You could actually make it private with  double `_`.
(https://www.tutorialsteacher.com/python/private-and-protected-access-modifiers-in-python)

tests/python_functional/src/helpers/loggen/conftest.py
furiel: `read_file` should not be boolean, it is a string (filename).

```
file-reader
  -R, --read-file=<filename>                 Read log messages from file
```

tests/python_functional/src/helpers/loggen/conftest.py
furiel: A little fun with the `count` generator, if you want to:

```
import itertools
class Loggen(object):
    index = itertools.count()
    get_new_index = lambda : next(Loggen.index)

print(Loggen.get_new_index())
print(Loggen.get_new_index())
```

You can add `_` in the front too for personal preference.

tests/python_functional/src/helpers/loggen/conftest.py
alltilla: Yes, you are right! My bad.

tests/python_functional/src/helpers/loggen/conftest.py
```diff
@@ -0,0 +1,187 @@
...
+    @staticmethod
+    def get_new_instance_index():
+        Loggen.instanceIndex += 1
+        return Loggen.instanceIndex
+
+    def __init__(self):
+        self.loggen_proc = None
+        self.loggen_bin_path = tc_parameters.INSTANCE_PATH.get_loggen_bin()
+
+    def _decode_start_parameters(
```
szemere, line 43: Missing parameter:

`-s, --size=<size>                          Specify the size of the syslog message`

kira-syslogng:
Build SUCCESS

tests/python_functional/src/helpers/loggen/conftest.py
alltilla: Resolved from my side. Thanks! :)

tests/python_functional/src/helpers/loggen/conftest.py
alltilla: @norberttak I think this is still missing.

alltilla (approved):

kira-syslogng:
Build SUCCESS

szemere:
Yes!

PR #3444, opened 2020-10-01
logwriter: destination timeout to verbose from notice

furiel:
After this change, these two logs will occur together:

```
Destination timeout has elapsed, closing connection; fd='12'
Destination timed out, reaping; template='/dev/stdout', filename='/dev/stdout'
```

Originally, only the first one was visibe (without verbose
flag). Withouth the context of the second, this can cause confusion.


kira-syslogng:
Build SUCCESS

Kokan (approved):

MrAnno (approved):

PR #3448, opened 2020-10-05
Light framework - use plugins

szemere:
Pytest imports all the conftest.py files all along the path of your test case. However we use a different folder (src) for our "common functionalities", like loggen, snmptrapd, etc...

While we still use the conftest.py files to implement our helpers, they are not included automatically. Leaving us with two possible solution (we use both of them currently):
1. Put our fixtures into the main conftest.py in the root folder of the unit test framework
2. Import the needed helpers in the testcases one-by-one.

note: `pytest --fixtures`, or `pytest --trace-config` will show the list of available fixtures. (Including the ones in the src folder.) However not all of them will be loaded during testcase execution. More on that here: https://github.com/pytest-dev/pytest/issues/3039

`pytest_plugins` allows us to include plugins globally:
- making the individual import statements futile
- cleans up the global conftest.py file
- our helpers can have a proper filenames, with clean internal structure

note: unfortunately our helpers are not "plugins" in the strict sense of the world, but they can be imported in the framework. (They can not be distributed as a full featured 3rd party plugins, without the setup.py file and other meta information.)


kira-syslogng:
Build SUCCESS

mitzkia:
From my point of view it is a good initiative. Thanks. 
Just one question, there are several other fixtures (syslog-ng, syslog-ng-ctl, config, log_message ...) in main conftest.py, what is your plan with them? I would place them in their right place also (you have already moved the bsd_format fixture).

mitzkia:
New module names: Logger.py, SecureLogging.py, SNMPTestParams.py, SNMPtrapd.py should be in lowercase, underscore are allowed. Thanks

szemere:
> From my point of view it is a good initiative. Thanks.
> Just one question, there are several other fixtures (syslog-ng, syslog-ng-ctl, config, log_message ...) in main conftest.py, what is your plan with them? I would place them in their right place also (as you have already moved the bsd_format fixture).

Thank You for the question. I want to convert them too. I just wanted to get the first review notes, before flipping everything upside down.
(Also my main focus is currently on proxy proto.)

szemere:
We had a long discussion about the project structure. We concluded to not use `pytest_plugins`, rather include our helpers in the "main" conftest.py, and declare the globally available fixtures in one place.
I am closing this PR in favour to #3458, eliminating the last user of the pytest_plugins variable.

Thank you all for your help.

PR #3449, opened 2020-10-05
Small fixes in the light framework

szemere:
- Change read to readline in FileIO
  431bc3583c25842c6c4155230dfeb52787191926 fixed the Console part, but the same is needed for FileIO.
- Change the syntax of PortAllocator
  Small hack in the pytest framework, port allocator is now a function, so it can be called multiple times if multiple port numbers needed in a testcase
- Missing parameter in Loggen
- Use Loggen as a plugin (we would like to change all of the helpers to this, so no need to import them individually)


kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/conftest.py
```diff
@@ -74,7 +74,11 @@ class PortAllocator():
...
 
 
-@pytest.fixture(scope="module")
-def some_port():
+def _get_next_port():
     PortAllocator.CURRENT_DYNAMIC_PORT += 1
     return PortAllocator.CURRENT_DYNAMIC_PORT
+
+
+@pytest.fixture(scope="session")
```
furiel, line 82: Do we need fixtures in the end? If `_get_next_port` would be public, we could just call it.

szemere: I think the public function will work in this situation. @mitzkia ?

(My approach was from visibility. With scope="session" we guarantee the lifetime of the inside Class in the pytest way.)

tests/python_functional/functional_tests/conftest.py
```diff
@@ -74,7 +74,11 @@ class PortAllocator():
     CURRENT_DYNAMIC_PORT = 30000
 
 
-@pytest.fixture(scope="module")
-def some_port():
+def _get_next_port():
```
norberttak, line 77: if your intention was to mark _get_next_port() as a private function, pls use a double underscore in the name.
__get_next_port

tests/python_functional/functional_tests/conftest.py
norberttak: Without the fixture we should have instantiated the class first which would be very inconvenient. I don't like that way.
Or we don't need this whole class at all. It could be a simple (I mean not a class member) function.

mitzkia: In the current code state if we remove this fixture (port_allocator) we should not instantiate the class (if you mean PortAllocator).
The difference should be something like this on a caller side:

```diff
+from functional_tests.conftest import get_next_port
...
-@pytest.fixture
-def snmptrapd(port_allocator):
-    server = SNMPtrapd(port_allocator())
+@pytest.fixture
+def snmptrapd():
+    server = SNMPtrapd(get_next_port())
```

From my point of view it is almost equivalent change.
From one side fixtures gives us a little flexibility about their life-cycles and usage (do not require import) , but in the other hand they (fixtures) can be very quickly multiplies, and "everything" will be a fixture, which is not so good. 
For now I would say we can keep this fixture, but later we should think them over (what can be a fixture and what not).

mitzkia:
Please use lowercase for the new modules, like Loggen.py -> loggen.py

tests/python_functional/functional_tests/conftest.py
furiel: If you want this function to be private, then instead of adding underscores,  you could make it a local function inside the `port_allocator`:

```python
@pytest.fixture(scope="session")
def port_allocator():
    def get_next_port():
        PortAllocator.CURRENT_DYNAMIC_PORT += 1
        return PortAllocator.CURRENT_DYNAMIC_PORT

    return get_next_port
```
(note that I did not underscore inside the `port_allocator` either)

What is the reason behind  `_get_next_port` to be private?

In general, I do not really like the effort we are putting into making things private by adding underscores. We are doing this way more often than it is reasonable. One downside is that it is harder to write unit test for such code. I do not find start with underscores aesthetically pleasing either. Sometimes, they can also get in the way: when I really want to call these functions, it is frustrating to sed the names. This would worth a discussion with the whole team. I thnk there are better ways for achieve information hiding.

norberttak (approved):

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

mitzkia (approved):

norberttak (approved):

PR #3453, opened 2020-10-12
date-parser: extend %z to support time zones other then local and gmt

Kokan:
Our date parser is based on the NetBSD `strptime` implementation, that differs from the linux implementation.
The version of `strptime` used from NetBSD supported local and GMT timezone with `%z` and the `%Z` supported the other timezones. This was changed in linux and also in later version of NetBSD.

In our documentation the newer NetBSD behaviour was written, this patch updates to code accordingly.

### DIFF
--- a/lib/timeutils/wallclocktime.c
+++ b/lib/timeutils/wallclocktime.c
@@ -245,7 +245,6 @@ static int first_wday_of(int yr);
 #define HAVE_HOUR(s)    (s & S_HOUR)
 #define HAVE_USEC(s)    (s & S_USEC)
 
-static char gmt[] = { "GMT" };
 static char utc[] = { "UTC" };
 /* RFC-822/RFC-2822 */
 static const char *const nast[5] =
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
I think this drops support for recognizing the name of the timezone that is currently set on the current host running syslog-ng.

e.g. the tzname global variable that is set by libc.

Try setting CET as the current timezone and send a message using "CET" as the timezone name.

I am not sure this is that useful, but Ciscos tend to send timezone names instead of offsets and at the very least those would be recognized as long as the router is in the same timezone. In contrast to this we should attempt to recognize any timezone names, that we could extract from tzinfo files, as we do with allowed timezone offsets (e.g. _is_gmtoff_valid() function in unixtime.c)

kira-syslogng:
Build SUCCESS

MrAnno (approved):

MrAnno:
Nice!

bazsi:
@rfaircloth-splunk Once we discussed syslog-ng's date-parser() "%Z" format parsing timezone names. This branch improves that with the US based timezone names, still not global, but stuff like EST/EDT would be recognized with this branch integrated. Do you also see non-US timezones in the wild? I guess you do, I just wanted to gauge if it makes sense to support those as well.

bazsi (approved):
Could still be improved with all the global timezone names, but still a useful step.

Kokan:
There is already two approve and green CI, so I am merging this.

PR #3455, opened 2020-10-15
Tarball image to debian testing

### LINKED ISSUE #3461: "dbld/rules tarball" fails due to bison
When trying to run ```dbld/rules tarball``` it fails due to bison:
```
configure: WARNING: bison is found, but your bison version 3.0.4 is not recent enough, at least 3.4.2 is required
configure: error: syslog-ng requires bison 3.4.2 or later. Your source tree seems to be from git, which doesn't have the bison generated files (like cfg-grammar.c). Please install/upgrade bison or use a distribution tarball.
make: *** [dbld/rules:101: tarball-ubuntu-bionic] Error 1
```
I even tried to re-build the container image, but it did not help. I guess the problem is related to: https://github.com/syslog-ng/syslog-ng/commit/4b9d6f332964c20ad99ddeacb4c30014fcbe042a

bazsi:
This builds on top of #2526, which should be merged before this one.

Fixes #3461


gaborznagy:

kira-syslogng:
Package Build FAILURE

kira-syslogng:
Build SUCCESS

gaborznagy:
```
make  distdir-am
 (cd misc && make  top_distdir=../../../syslog-ng-3.29.1.138.g9aba409 distdir=../../../syslog-ng-3.29.1.138.g9aba409/lib/ivykis/misc \
make  distdir-am
test -n ":" \
|| find "../../syslog-ng-3.29.1.138.g9aba409/lib/ivykis" -type d ! -perm -755 \
	-exec chmod u+rwx,go+rx {} \; -o \
  ! -type d ! -perm -444 -links 1 -exec chmod a+r {} \; -o \
  ! -type d ! -perm -400 -exec chmod a+r {} \; -o \
  ! -type d ! -perm -444 -exec /bin/bash /source/lib/ivykis/install-sh -c -m a+r {} {} \; \
|| chmod -R a+r "../../syslog-ng-3.29.1.138.g9aba409/lib/ivykis"
 (cd lib/jsonc && make --no-print-directory top_distdir=../../syslog-ng-3.29.1.138.g9aba409 distdir=../../syslog-ng-3.29.1.138.g9aba409/lib/jsonc \
make  distdir-am
:
test -d "../../syslog-ng-3.29.1.138.g9aba409/lib/jsonc" || mkdir "../../syslog-ng-3.29.1.138.g9aba409/lib/jsonc"
 (cd tests && make  top_distdir=../../../syslog-ng-3.29.1.138.g9aba409 distdir=../../../syslog-ng-3.29.1.138.g9aba409/lib/jsonc/tests \
make  distdir-am
make  \
  top_distdir="../../syslog-ng-3.29.1.138.g9aba409" distdir="../../syslog-ng-3.29.1.138.g9aba409/lib/jsonc" \
  dist-hook
test -d "../../syslog-ng-3.29.1.138.g9aba409/lib/jsonc/doc" || mkdir "../../syslog-ng-3.29.1.138.g9aba409/lib/jsonc/doc"
chmod -R u+w "../../syslog-ng-3.29.1.138.g9aba409/lib/jsonc/doc"
cd "../../syslog-ng-3.29.1.138.g9aba409/lib/jsonc" && doxygen
/bin/bash: doxygen: command not found
make[5]: *** [Makefile:1148: dist-hook] Error 127
make[4]: *** [Makefile:786: distdir-am] Error 2
make[3]: *** [Makefile:779: distdir] Error 2
make[2]: *** [Makefile:25105: distdir-am] Error 1
make[1]: *** [Makefile:25099: distdir] Error 2
make: *** [Makefile:25205: dist] Error 2
dbld/rules:101: recipe for target 'tarball-debian-testing' failed
```

tarball generation fails since `doxygen` is missing. I'm reviewing the `dbld/packages.manifest` file and send a PR to you @bazsi soon!

gaborznagy:
The PR: https://github.com/bazsi/syslog-ng/pull/8

bazsi:
@kira-syslogng build packages please.

bazsi:

kira-syslogng:
Package Build FAILURE

kira-syslogng:
Package Build FAILURE

gaborznagy:
@kira build-packages please;

kira-syslogng:
Package Build FAILURE

gaborznagy:
@bazsi I've built locally the `tarball` image and tested tarball creation. Since it worked, I've pushed the `balabit/syslog-ng-tarball` image to [DockerHub](https://hub.docker.com/repository/docker/balabit/syslog-ng-tarball).
Without this, the build packages job doesn't work, as it want's to pull a non-existing image (there is no fallback mechanism in dbld to build the image)

I don't know if we want to publish/host this new `balabit/syslog-ng-tarball` image, or want to refactor dbld to build tarball image if it's missing.

The tarball step works now, however we have a lot of package build issues.

gaborznagy:
Even ubuntu-focal packaging is broken, I have no what could be the reason:
```
java                        : yes
  java modules                : yes
  afsnmp module               : yes
  native bindings             : auto
make[1]: Leaving directory '/build/syslog-ng-3.29.1.141.gb82fd29'
   dh_auto_build
	cd debian/build-tree && make -j16
make[1]: Entering directory '/build/syslog-ng-3.29.1.141.gb82fd29/debian/build-tree'
make[1]: *** No rule to make target '../../modules/java/src/main/java/org/syslog_ng/LogMessage.java', needed by 'modules/java/syslog-ng-core/classes/org/syslog_ng/LogMessage.class'.  Stop.
make[1]: Leaving directory '/build/syslog-ng-3.29.1.141.gb82fd29/debian/build-tree'
dh_auto_build: error: cd debian/build-tree && make -j16 returned exit code 2
make: *** [debian/rules:231: build] Error 25
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
dbld/rules:126: recipe for target 'package-ubuntu-focal' failed
make: *** [package-ubuntu-focal] Error 2
```

alltilla:
@gaborznagy This is due to too fresh java, isn't it?

gaborznagy:
> @gaborznagy This is due to too fresh java, isn't it?

No, we still have the same 1.8 java everywhere (focal, devshell too).
I've tried this on a xenial machine too locally (i.e not with dbld) with an updated bison (3.6.4) and syslog-ng does not compile with java 1.8 on it...

dbld/rules
```diff
@@ -8,13 +8,14 @@ BUILDER_IMAGES=		\
...
-	kira
+	ubuntu-bionic	\
+	ubuntu-focal	\
+	kira		\
+	tarball
 
 IMAGES=${BUILDER_IMAGES} devshell
 DEFAULT_IMAGE ?= ubuntu-bionic
-TARBALL_IMAGE ?= ubuntu-bionic
+TARBALL_IMAGE ?= tarball
```
gaborznagy, line 18: This topic come up in #3461, but why do we allow to build the tarball on other platforms, then the TARBALL_IMAGE?
I mean the `dbld/rules tarball-<platform>` command.
`		$(DOCKER) run $(DOCKER_RUN_ARGS) --rm -i balabit/syslog-ng-$* /dbld/tarball; \	`

Shouldn't we limit that tarball can be run only our tarball image?

bazsi:
build packages please.

kira-syslogng:
Package Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:

There's one issue left: we are defining the bison version in SYSLOG_NG_BISON_MAJOR/MINOR based on what we have on the system instead of what we used to compile the grammars with. So in case we have a new bison in tarball generation and the one
in the build environment is older, we might still run into a compilation issue around YYEMPTY.

The solution is to propagate the bison version we used during make dist to the compilation phase, but I am not sure yet how.

With all this said, this should still fail on old distros complaining with YYEMPTY missing, but only because the workarond fails
to kick in.

```
#if SYSLOG_NG_BISON_MAJOR == 3 && SYSLOG_NG_BISON_MINOR <= 5

/* old bison versions (e.g. 3.5.1 and below) when using api.prefix, didn't
 * generate the PRAGMA_EMPTY enum, but had YYEMPTY, so let's define
 * PRAGMA_EMPTY to YYEMPTY, so the grammar itself is able to use the new
 * style */
#define PRAGMA_EMPTY YYEMPTY
#endif

```

kira-syslogng:
Package Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:
I think I have solved the bison versioning issue with automake but I would need help to do the same with cmake. This is the key part of the solution:

```
.y.c:
	$(AM_V_YACC)$(am__skipyacc) $(SHELL) $(YLWRAP) $< y.tab.c $@ y.tab.h $*.h y.output $*.output -- $(YACCCOMPILE)
	$(AM_V_GEN) sed -i -e 's/^#define YYBISON_VERSION "\([0-9]\+\)\.\([0-9]\+\)\..*$$/&\n#define YYBISON_VERSION_MAJOR \1\n#define YYBISON_VERSION_MINOR \2/' $@
```

e.g. we run a sed script on the generated source. With this in place, we can only do a release with automake.

bazsi:
It's not just the dist tarball that we can't build with automake. When running bison from cmake the result is not compiling due to the missing macros.
:( 

this one really needs a solution. With the current state of master, we can't produce working tarballs, with the current state in this PR that already works but cmake is broken.

@lbudai @MrAnno

kira-syslogng:
Build SUCCESS

bazsi:

Kokan:
I would really love to avoid changing the generated file, it is kinda risky.
An alternative idea (not tested) would be to keep the current MAJOR, MINOR parsing, and inject those values not into the `sylsog-ng.h` but rather directly into the grammar files. That could be done for example with the `merge-grammar.py`.

When there is a generated grammar file, it would contain the versions that it was generated with.
When it needs to recompile the grammar file, it would put the version of bison on the system in hand.

Additonally that should resolve both the cmake and autotools issue as well without even touching them. (less difference to maintain)

@bazsi what do you think ?

kira-syslogng:
Package Build FAILURE

alltilla:
Pardon my simplicity, but isn't it enough to just put it behind `#ifndef`?
What are the scenarios where the macro is not defined because of some other reason than the bison version being old?

kira-syslogng:
Build FAILURE

bazsi:
YYSTYPE is a typedef not a macro in most cases, but I'll double-check.

alltilla:
I think we are interested in `PRAGMA_EMPTY`, but yeah, it is an enum...

`lib/pragma-grammar.h`:
```
#ifndef PRAGMA_TOKENTYPE
# define PRAGMA_TOKENTYPE
  enum pragma_tokentype
  {
    PRAGMA_EMPTY = -2,
    PRAGMA_EOF = 0,                /* "end of file"  */
```

bazsi:
One issue is that cfg-grammar.y is not run through merge-grammar.py, which
might not be using the YYEMPTY today but might in the future.

Kokan:
But still imho it would be better to include the computed major/minor version into the yacc files instead of replacing the generated file as that could easily break as the version string might change. Doing an extra step for cfg-grammar.y might not be that intrusive as using unexpected sed in build process.

bazsi:
> But still imho it would be better to include the computed major/minor version into the yacc files instead of replacing the generated file as that could easily break as the version string might change. Doing an extra step for cfg-grammar.y might not be that intrusive as using unexpected sed in build process.

Sorry, but I don't see how "that could easily break as the version string might change".

kira-syslogng:
Build SUCCESS

gaborznagy:
@bazsi another idea I was unable to explain my idea:

We agreed, that we shouldn't override the bison versions in configure time, and should use the bison version when the grammars are created (at tarball creation time -> `make dist`).
How about that we create a new output file instead of using the `syslog-ng-config.h`? 
We can add an additional step to `make dist` (or the grammar generating make rule) to create a `bison-version.h` file and use that in the grammar files.
What do you think?

gaborznagy:

kira-syslogng:
Package Build SUCCESS

Packages are available at https://ci.syslog-ng.com/jenkins/job/kira-dbld-pull-requests/721/artifact/

kira-syslogng:
Build SUCCESS

dbld/rules
bazsi: The target we are promoting for building tarballs is `./dbld/rules tarball` as per the help screen of the rules file, where the name of the platform to use is implied. The qualified name `tarball-<OS>` is useful as an internal target, not to mention the fact that `tarball` is just using the same conventions all of our other targets.

All in all, I don't think it hurts, users should not use it directly, but if they do that's up to them.

bazsi:
> @bazsi another idea I was unable to explain my idea:
> 
> We agreed, that we shouldn't override the bison versions in configure time, and should use the bison version when the grammars are created (at tarball creation time -> `make dist`).
> How about that we create a new output file instead of using the `syslog-ng-config.h`?
> We can add an additional step to `make dist` (or the grammar generating make rule) to create a `bison-version.h` file and use that in the grammar files.
> What do you think?

I think this would bring other problems. When would we generate syslog-ng-bison.h? 

1) At configure time? Not good, as we would overwrite this file on the build host where we are trying to use the released tarball.
 
2) at "dist" time? If you extract a release tarball on a separate computer and try to build it, you are fine as long as you DON'T want to regenerate the grammar, which is something that you might want. e.g. when you apply patches to the grammar. The debian packaging does this.

3) In the .y.c rule? This is the place where I added the sed commands (e.g. at the point where we generate .c file from .y). This is executed for each of the .y/.ym files around, so we would generate bison-version.h multiple times, which one would we store? what if we only re-generated a subset of the grammars and not all of them?

Frankly I don't see why this would be any better than inserting the defines into the .c file, as I've implemented it today. The version of bison we use to generate the grammar is encoded in its output as it should be. The bison version is not a property of the .y/.ym, it is the property of the bison _output_, so that's the proper place to save this information. I've also opened a ticket for bison to add this information in the future: https://github.com/akimd/bison/issues/55

kira-syslogng:
Build SUCCESS

bazsi:
> > But still imho it would be better to include the computed major/minor version into the yacc files instead of replacing the generated file as that could easily break as the version string might change. Doing an extra step for cfg-grammar.y might not be that intrusive as using unexpected sed in build process.
 
I have now changed the patch to use the bison version detected in configure instead of relying on the .c output.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:
Once we conclude on the sed/non-sed argument this should be ready to go in and fix the issue w/ tarball generation. https://github.com/syslog-ng/syslog-ng/issues/3461

Kokan:
> Frankly I don't see why this would be any better than inserting the defines into the .c file, as I've implemented it today. The version of bison we use to generate the grammar is encoded in its output as it should be. The bison version is not a property of the .y/.ym, it is the property of the bison output, so that's the proper place to save this information. I've also opened a ticket for bison to add this information in the future: akimd/bison#55

My thinking followed as such, there is a compat layer created in the grammar files by us due to bison incompatibilities. But that compat layer has a wrong dependency, as that depends on the fact it receives a proper bison version; that promise is broken.

This PR surely fixes that promise, as my suggestion.

I agree that this should be a property of the generated files, but it isn't currently. That should be fixed with a compatibility layer, and not modifying the generated files (surely if feasible only, but in this case it is)

I only even bothered to propose an alternative, because your solution fixes the promise on a level I would never expect.

Regarding the bison versioning format change, sure it was just an imaginary scenario (that is kinda unlikely, but not impossible) where the version format is changed in a way the current regex won't match. The main point is I would rather avoid rewriting any generated code if possible.

So my issue is not about using sed, rather how and where it is used.
Also please note that - even thou I do not like this version for reasons explained above - this was a suggestion how to resolve differently and not a change request. This cannot be merged in versus this shouldn't be merged.


## cmake

Moving on, about cmake part of your proposal.

CMake provides a function/macro free of charge (`bison_target`) that you can just parameterise to generate the grammar files.
This macro is part of the cmake source code, so if you have cmake you have that macro.

In order to attach the extra echo and sed command to the `bison_target`, there are some possibilities that I am aware of:

1. Change the cmake implementation of `bison_target` in our local version.
For this one should commit the https://github.com/Kitware/CMake/blob/d749d4681414c7590ac06e98697e2fed324239d2/Modules/FindBISON.cmake#L269-L274 in our `cmake/Modules` directory (as this directory is used before global cmake directories).
After that a simple change in the above code is required:
```diff
       add_custom_command(OUTPUT ${BISON_TARGET_outputs}
         COMMAND ${BISON_EXECUTABLE} ${BISON_TARGET_cmdopt} -o ${BisonOutput} ${_BisonInput}
+        COMMAND -i -e "1i\\$${nl}#define SYSLOG_NG_BISON_MAJOR @bison_version_major@$${nl}1i\\$${nl}#define SYSLOG_NG_BISON_MINOR @bison_version_minor@"  ${BisonOutput}
         VERBATIM
         DEPENDS ${_BisonInput}
         COMMENT "[BISON][${Name}] Building parser with bison ${BISON_VERSION}"
         WORKING_DIRECTORY ${_BISON_WORKING_DIRECTORY})
```

2. Wrap the current `bison_target`
This can be done with either using the fact that if you define `bison_target` again, you can call the previous one with `_bison_target` name. (this is really not recommended by efficient cmake[1])
Simply create a new name for it, and replace usage (simple enough).

The wrapper should generate two files, the first is generated by bison and the second one is by use (with sedding stuff).

```cmake
macro(BISON_GEN Name BisonInput BisonOutput)
   bison_target(${Name} ${BisonInput} ${BisonOutput}.tmp)

   add_custom_command(${BisonOutput}
                      DEPENDS ${BisonOutput}.tmp
                      COMMAND sed ... ${BisonOutput}.tmp > ${BisonOutput})
endmacro(BISON_GEN)
```

I would suggest the latest one. If you want to I can do either version for you.


[1] https://crascit.com/2018/09/14/do-not-redefine-cmake-commands/

bazsi:
Thanks a lot for your suggestions. Some additional comments below.

> My thinking followed as such, there is a compat layer created in the grammar files by us due to bison incompatibilities. But that compat layer has a wrong dependency, as that depends on the fact it receives a proper bison version; that promise is broken.
> 
> This PR surely fixes that promise, as my suggestion.
> 
> I agree that this should be a property of the generated files, but it isn't currently. That should be fixed with a compatibility layer, and not modifying the generated files (surely if feasible only, but in this case it is)
> 
> I only even bothered to propose an alternative, because your solution fixes the promise on a level I would never expect.

This is interesting. I find that this is the best spot to fix it and I don't understand why it is not so. bison generates a file with a deficiency, it is also breaking an established API. I understand why YYEMPTY is renamed to <PREFIX>_EMPTY, but this is actually a breaking change. bison documents which variables are renamed because of api.prefix and YYEMPTY is not listed:


> The renamed symbols include ‘yyparse’, ‘yylex’, ‘yyerror’, ‘yynerrs’,
> ‘yylval’, ‘yylloc’, ‘yychar’ and ‘yydebug’.  If you use a push parser,
> ‘yypush_parse’, ‘yypull_parse’, ‘yypstate’, ‘yypstate_new’ and
> ‘yypstate_delete’ will also be renamed.  The renamed macros include
> ‘YYSTYPE’, ‘YYLTYPE’, and ‘YYDEBUG’, which is treated specifically —
> more about this below.
> 

What this means is that we have a broken bison, we could consider either 3.4-3.5 broken (because that's when api.prefix became mandatory and those versions didn't rename YYEMPTY yet), or we could also consider 3.6 and subsequent versions broken (because they break their API). I didn't open an issue in bison to fix this, as the damage is already done. If they reverted the YYEMPTY change, it would just cause even more pain for us down the line.

To fix this up, we need the version number of bison to work with. I hated the fact that I had to add this to the configure script in the previous iteration. I even considered to use whatever macros bison publishes in my original workaround instead of having to go back to the configure script to export the version number. At the end I concluded on using configure and #2526 got merged. 

This is where we are today: the solution in the configure phase is br*ken, as I missed the fact that the configure script is run even in we have generated versions of the grammar in the tarball.

Now, I was thinking: should my bison workaround be pushed to even deeper and completely unrelated parts of our build system (e.g. merge-grammar.py), or fix it exactly where the problem is: where we exec bison. I find it the right spot, work around the problem exactly where it happens.

> 
> Regarding the bison versioning format change, sure it was just an imaginary scenario (that is kinda unlikely, but not impossible) where the version format is changed in a way the current regex won't match. The main point is I would rather avoid rewriting any generated code if possible.

The version number detection is not dependant on the generated .c file anymore, I've reverted to using the values detected by configure. I find it a bit uglier/more complex, but I've agreed that a regex against a generated piece of code is not hassle-free in the long term. With all this said, I'd point out  that the configure script detects bison version using a very similar regexp.

> 
> So my issue is not about using sed, rather how and where it is used.
> Also please note that - even thou I do not like this version for reasons explained above - this was a suggestion how to resolve differently and not a change request. This cannot be merged in versus this shouldn't be merged.

I think I understand your arguments. I still disagree with them though.

> ## cmake
> 
> Moving on, about cmake part of your proposal.
> 
> CMake provides a function/macro free of charge (`bison_target`) that you can just parameterise to generate the grammar files.
> This macro is part of the cmake source code, so if you have cmake you have that macro.
> 
> In order to attach the extra echo and sed command to the `bison_target`, there are some possibilities that I am aware of:
> 
>     1. Change the cmake implementation of `bison_target` in our local version.
>        For this one should commit the https://github.com/Kitware/CMake/blob/d749d4681414c7590ac06e98697e2fed324239d2/Modules/FindBISON.cmake#L269-L274 in our `cmake/Modules` directory (as this directory is used before global cmake directories).
>        After that a simple change in the above code is required:
> 
> 
> ```diff
>        add_custom_command(OUTPUT ${BISON_TARGET_outputs}
>          COMMAND ${BISON_EXECUTABLE} ${BISON_TARGET_cmdopt} -o ${BisonOutput} ${_BisonInput}
> +        COMMAND -i -e "1i\\$${nl}#define SYSLOG_NG_BISON_MAJOR @bison_version_major@$${nl}1i\\$${nl}#define SYSLOG_NG_BISON_MINOR @bison_version_minor@"  ${BisonOutput}
>          VERBATIM
>          DEPENDS ${_BisonInput}
>          COMMENT "[BISON][${Name}] Building parser with bison ${BISON_VERSION}"
>          WORKING_DIRECTORY ${_BISON_WORKING_DIRECTORY})
> ```
> 
>     1. Wrap the current `bison_target`
>        This can be done with either using the fact that if you define `bison_target` again, you can call the previous one with `_bison_target` name. (this is really not recommended by efficient cmake[1])
>        Simply create a new name for it, and replace usage (simple enough).
> 
> 
> The wrapper should generate two files, the first is generated by bison and the second one is by use (with sedding stuff).
> 
> ```cmake
> macro(BISON_GEN Name BisonInput BisonOutput)
>    bison_target(${Name} ${BisonInput} ${BisonOutput}.tmp)
> 
>    add_custom_command(${BisonOutput}
>                       DEPENDS ${BisonOutput}.tmp
>                       COMMAND sed ... ${BisonOutput}.tmp > ${BisonOutput})
> endmacro(BISON_GEN)
> ```
> 
> I would suggest the latest one. If you want to I can do either version for you.
> 
> [1] https://crascit.com/2018/09/14/do-not-redefine-cmake-commands/

Right now the code builds with cmake, and I am not sure we need to do this, but we might want to. Currently cmake does not allow building a dist tarball. As long as this is the case, the entire problem does not exist there. We never ever generate .c files from .y/.ym files that would then be distributed to other users running a different bison version. And as long as this is the case, detecting the bison version number in the configure phase is enough, and that's what my patch does currently. Should we ever want to create dist tarballs in cmake, we would need to add a similar solution though. But by that time, we hopefully can rely on bison v3.6+ everywhere and the workaround can be removed. Or https://github.com/akimd/bison/issues/55 gets fixed and we can standardize on that version in our `image-tarball`, which means that the entire workaround can be removed.

So, shall we improve the cmake part as a part of this PR or we can wait until something described above triggers it?

alltilla:
@bazsi 
I created a commit which implements the sed logic to cmake. It was checked IRL by @Kokan.
https://github.com/bazsi/syslog-ng/pull/9

kira-syslogng:
Build SUCCESS

lbudai (dismissed):

Kokan (approved):

Kokan:
We should wait for the result of the internal build, that is done by @gaborznagy with the merge.

czanik:
Creating a tarball worked, just as building a CentOS 7 packages using dbld. On the other hand building openSUSE packages failed:
```
[   24s] /bin/sh /home/abuild/rpmbuild/BUILD/syslog-ng-3.29.1.196.g986eeb2/install-sh -d modules/xml/
[   24s] /usr/bin/python3 ./lib/merge-grammar.py modules/xml/xml-grammar.ym > modules/xml/xml-grammar.y
[   24s] /bin/sh ./ylwrap lib/rewrite/rewrite-expr-grammar.y y.tab.c lib/rewrite/rewrite-expr-grammar.c y.tab.h lib/rewrite/rewrite-expr-grammar.h y.output lib/rewrite/rewrite-expr-grammar.output -- echo Required bison not found && false -Wno-yacc -Wno-other -Werror=conflicts-sr -Werror=conflicts-rr -d
[   24s] Required bison not found /home/abuild/rpmbuild/BUILD/syslog-ng-3.29.1.196.g986eeb2/lib/rewrite/rewrite-expr-grammar.y
[   24s] make: *** [Makefile:27619: lib/rewrite/rewrite-expr-grammar.c] Error 1
[   24s] make: *** Waiting for unfinished jobs....
```

czanik:
OK. I don't know what ```dbld``` does differently, but when I tried to build real-world packages using OBS and Copr, they failed on older platforms, including CentOS 7:

Copr: https://copr.fedorainfracloud.org/coprs/czanik/syslog-ng-githead/build/1722494/

OBS: https://build.opensuse.org/package/show/home:czanik:syslog-ng-githead/syslog-ng

gaborznagy:

kira-syslogng:
Package Build FAILURE

gaborznagy:
1 dbld packaging job failed: on fedora-32 python is not found. 
I think this is not related to this PR.
```
[91m+ install_pip_packages
+ case "${OS_PLATFORM}" in
+ python_executables=python
+ for python_executable in ${python_executables}
+ case "${OS_PLATFORM}" in
+ python -m pip install --upgrade pip
/dbld/builddeps: line 197: python: command not found
```

alltilla:
fedora-32 failed in Distro packages:
```
[fedora-32] Step 14/20 : RUN /dbld/builddeps install_pip_packages
[fedora-32]  ---> Running in b34ff567b137
[fedora-32] [91m+ install_pip_packages
[fedora-32] + case "${OS_PLATFORM}" in
[fedora-32] + python_executables=python
[fedora-32] + for python_executable in ${python_executables}
[fedora-32] + case "${OS_PLATFORM}" in
[fedora-32] + python -m pip install --upgrade pip
[fedora-32] /dbld/builddeps: line 197: python: command not found
[fedora-32] [0mThe command '/bin/sh -c /dbld/builddeps install_pip_packages' returned a non-zero code: 127
[fedora-32] /mnt/jenkins/workspace/kira-dbld-pull-requests/fedora-32/dbld/rules:195: recipe for target 'image-fedora-32' failed
[fedora-32] make[1]: *** [image-fedora-32] Error 127
[fedora-32] dbld/rules:204: recipe for target 'cache-image-fedora-32' failed
[fedora-32] make: *** [cache-image-fedora-32] Error 2
[Pipeline] [fedora-32] }
[Pipeline] [fedora-32] // stage
[Pipeline] [fedora-32] }
[fedora-32] Failed in branch fedora-32
```

gaborznagy:
I look into the fedora issue, otherwise ready for merge (internal build was successful).

gaborznagy:
@czanik The trick is that the dbld package generation targets (`deb`/`rpm` or `package-<os>`) build from a tarball, which was previously generated on a suitable platform (i.e. where bison version is >= 3.4.2).

In Copr and OBS builds the bison dependency is the distribution shipped version.
Based on the bison versions in the different distributions (see https://repology.org/project/bison/versions), I see why it's passed on OpenSuse Tumbleweed and fedora-32 and fedora-33.

You need to use an updated version of bison on those platforms where the build failed.

gaborznagy:
The fedora package build issue's root cause is a missing package: `python-unversioned-command`.
This package provides the `/usr/bin/python` command (symlink to python3) which we depend on explicitly in `dbld/builddeps`:
https://github.com/syslog-ng/syslog-ng/blob/059c1154ab87833bfb2598600c20ede4f09a8aaf/dbld/builddeps#L174-L178

Last docker build was 5 days ago and then the package was installed at that time(see logs, note the "weak dependency" part)
https://hub.docker.com/repository/docker/balabit/syslog-ng-fedora-32/builds/8ad3ec0d-59a2-4f58-b85f-54e2f03925eb
```
Installing weak dependencies:
libxcrypt-compat x86_64 4.4.17-1.fc32 updates 97 k
perl-IO-Socket-IP noarch 0.39-441.fc32 fedora 42 k
perl-IO-Socket-SSL noarch 2.068-1.fc32 fedora 217 k
perl-Mozilla-CA noarch 20200520-1.fc32 updates 12 k
python-unversioned-command noarch 3.8.6-1.fc32 updates 13 k
```

We can either install the missing package, or stop depending on `/usr/bin/python` in `dbld/builddeps` and use python3 simply.
I wonder why doesn't the unversioned-command is installed, as Fedora states it clearly:
https://fedoraproject.org/wiki/Changes/Python_means_Python3
>make python3 (instead of python2) recommend python-unversioned-command
And checking the package itself, it should work, 
```
# rpm -q --recommends python3
/usr/bin/python
python3-pip
python3-setuptools
```

gaborznagy:
@bazsi I've created a PR to fix the fedora-32 image build issue

bazsi:
@czanikp: The problem with rpm based platforms is worked around in this
patch that I've included in this patch:

https://github.com/syslog-ng/syslog-ng/pull/3455/commits/775c1c823e2b7708890ba7bdb72fdfd6fcb862ae

The issue is that the contents of merge-grammar.py is modified in %setup,
causing its timestamp to get newer than the .ym files. This will result in
the rerun of bison even though those files are already included in the
tarball.

Include the workaround or better yet, let's work on merging your rpm
packaging files into the source tree. :)

czanik:
Build problems for older SUSE and Red Hat releases are now resolved thanks to @gaborznagy, who pointed me at this spec file change: https://github.com/syslog-ng/syslog-ng/pull/3455/commits/775c1c823e2b7708890ba7bdb72fdfd6fcb862ae

Please include this change in the developer section of release notes!

bazsi:
You mean the requirement on bison 3.4.2+? That should be documented, but
shouldn't be needed just for packaging unless you also want to be able to
change the grammar files.

alltilla:
@czanik 
Thanks for the heads up, the news file was missing in #2526 
I created a PR for it: https://github.com/syslog-ng/syslog-ng/pull/3470

gaborznagy:
@bazsi I've asked Kira to do a build-packages on my branch (see https://github.com/bazsi/syslog-ng/pull/10), to see the fix in action.
It passes, so when the PR is merged, this PR will go green and ready for merge.
https://ci.syslog-ng.com/jenkins/job/kira-dbld-pull-requests/723/

kira-syslogng:
Build SUCCESS

gaborznagy:

kira-syslogng:
Package Build SUCCESS

Packages are available at https://ci.syslog-ng.com/jenkins/job/kira-dbld-pull-requests/724/artifact/

alltilla:
About the macOS job fail:
There is a new bug in brew introduced 12 hours ago. https://github.com/Homebrew/homebrew-core/issues/63619 
I don't think it should block the merging of this PR, we can solve it after.

alltilla (approved):

Kokan (approved):

gaborznagy (approved):

PR #3462, opened 2020-10-22
Loggen: command line parameters for proxy protocol header

norberttak:
This PR adds three new command-line option to the loggen to allow the users to define the
source IP, destination IP, and port parameters used in the proxy protocol header.

### DIFF
--- a/tests/loggen/loggen.c
+++ b/tests/loggen/loggen.c
@@ -57,7 +57,6 @@ static PluginOption global_plugin_option =
   .target = NULL,
   .port = NULL,
   .rate = 1000,
-  .proxied = FALSE,
 };
 
 static char *sdata_value = NULL;
@@ -71,6 +70,7 @@ static int read_from_file = 0;
 static gint64 raw_message_length = 0;
 static gint64 *thread_stat_count = NULL;
 static gint64 *thread_stat_count_last = NULL;
+static gboolean proxied = FALSE;
 
 static GMutex *message_counter_lock = NULL;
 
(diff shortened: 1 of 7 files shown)


kira-syslogng:
Build SUCCESS

alltilla:
Thanks for the PR!

What was the reason behind moving the proxy header generation from `socket_plugin.c` to `loggen.c`? Was it to make it work with ssl?
If so, than I think we should use `generate_proxy_header()` in `ssl_plugin.c` instead. I think it is not an accident that we have `generate_proxy_header()` in `loggen_helpers.c`, so any plugin can use it.

norberttak:
Thanks @alltilla for your review comment. Proxy protocol is no more than (from loggen point of view) generate a simple text message at the beginning of the connection. To generate the log lines (or read from a file) we already have a separated part which is the generate_message() function used by every plugin.
Neither socket nor SSL plugin needs to know about proxy protocol. This is a higher-level function.

In the original form of the code if somebody call the SSL plugin with the proxied option, it will silently ignore the option which is not a good solution. To put the same code to all plugin is also not what I really want to do.

alltilla:
I see.

Just thinking out loud:
We might have plugins which does not support proxy protocol (_\*cough\* ALTP_). With the common solution they will be able to produce proxy protocol messages, which we cannot handle in `syslog-ng`. But after all, it might not be a problem, maybe we will have that implemented in `syslog-ng` some time.

Thanks for the explanation, I am convinced. :)

tests/loggen/loggen.c
```diff
@@ -81,7 +84,10 @@ static GOptionEntry loggen_options[] =
   { "interval", 'I', 0, G_OPTION_ARG_INT, &global_plugin_option.interval, "Number of seconds to run the test for", "<sec>" },
   { "permanent", 'T', 0, G_OPTION_ARG_NONE, &global_plugin_option.permanent, "Send logs without time limit", NULL},
   { "syslog-proto", 'P', 0, G_OPTION_ARG_NONE, &syslog_proto, "Use the new syslog-protocol message format (see also framing)", NULL },
-  { "proxied", 'H', 0, G_OPTION_ARG_NONE, &global_plugin_option.proxied, "Generate PROXY protocol v1 header", NULL },
+  { "proxied", 'H', 0, G_OPTION_ARG_NONE, &proxied, "Generate PROXY protocol v1 header", NULL },
+  { "proxy-src-ip", 0, 0, G_OPTION_ARG_STRING, &proxy_src_ip, "Source IP for the PROXY protocol v1 header", "<ip address>" },
+  { "proxy-dst-ip", 0, 0, G_OPTION_ARG_STRING, &proxy_dst_ip, "Destination IP for the PROXY protocol v1 header", "<ip address>" },
+  { "proxy-port", 0, 0, G_OPTION_ARG_STRING, &proxy_port, "Port for the PROXY protocol v1 header", "<port>" },
```
alltilla, line 90: I know it wasn't generated in the original code, but can we have both source and destination port?

kira-syslogng:
Build SUCCESS

alltilla (approved):

Kokan (approved):

PR #3473, opened 2020-11-04
usertty: error message on open error

gaborznagy:
With this PR, on 1 message there could be multiple error messages (as many as logged in users are for the configured username), this could fill the internal log pretty much.
Nevertheless, I think usertty() destination is mainly intended for alarms and critical log messages, thus logging this much internal  messages is okay.

I don't think a news entry is needed (not a bugfix, and not really something that should be announced).

### DIFF
--- a/modules/afuser/afuser.c
+++ b/modules/afuser/afuser.c
@@ -93,6 +93,16 @@ _utmp_entry_matches(UtmpEntry *ut, GString *username)
   return FALSE;
 }
 
+static gchar *
+_get_utmp_username(UtmpEntry *ut)
+{
+#ifdef SYSLOG_NG_HAVE_MODERN_UTMP
+  return ut->ut_user;
+#else
+  return ut->ut_name;
+#endif
+}
+
 G_LOCK_DEFINE_STATIC(utmp_lock);
 
 static void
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

Kokan (commented):
lgtm, please fix style-check job

modules/afuser/afuser.c
```diff
@@ -134,23 +144,24 @@ afuser_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options)
...
                     evt_tag_str("line", line));
           fd = open(line, O_NOCTTY | O_APPEND | O_WRONLY | O_NONBLOCK);
-          if (fd != -1)
+          if (fd == -1)
+            {
+              msg_error("Opening tty device failed",
+                        evt_tag_str("user", _get_utmp_username(ut)),
+                        evt_tag_str("line", line),
+                        evt_tag_error("errno"));
+              continue;
```
bazsi, line 156: we might similarly want to disable the usertty() destination in this case, don't we?

gaborznagy: I've added the disabling part too. Thanks for the suggestion!

kira-syslogng:
Build SUCCESS

Kokan (commented):
After adding disabling of usertty, it might be good to include into a news file.

MrAnno (dismissed):

gaborznagy:
> After adding disabling of usertty, it might be good to include into a news file.

Thanks @Kokan , I've added one to the bugfix category.
It's not an actual bugfix, but it helps to the investigate issues with usertty() destination.

kira-syslogng:
Build SUCCESS

Kokan (approved):

MrAnno (approved):

PR #3474, opened 2020-11-04
option truncating logwriter based destination msg payload size

Kokan:
The PR contains the following changes:
* A new option for `LogWriter` based destinations for truncating formatted message length `tuncate_size`, with default not truncating any messages.
* Change the default truncate_size only for udp network destination to 65507.
* New counters indicating truncation:
  ```
   dst.network;d_local#0;udp,127.0.0.1:1111;a;truncated_count;1
   dst.network;d_local#0;udp,127.0.0.1:1111;a;truncated_bytes;1
  ```
* Fix logwriter unit test (`LogWriterOptions` was "hand" initialized)

### DIFF
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -25,6 +25,7 @@
 #include "logwriter.h"
 #include "messages.h"
 #include "stats/stats-registry.h"
+#include "stats/stats-cluster-single.h"
 #include "hostname.h"
 #include "host-resolve.h"
 #include "seqnum.h"
@@ -69,6 +70,11 @@ struct _LogWriter
   StatsCounterItem *suppressed_messages;
   StatsCounterItem *processed_messages;
   StatsCounterItem *written_messages;
+  struct
+  {
+    StatsCounterItem *count;
+    StatsCounterItem *bytes;
+  } truncated;
   LogPipe *control;
   LogWriterOptions *options;
   LogMessage *last_msg;
@@ -1096,7 +1102,12 @@ log_writer_format_log(LogWriter *self, LogMessage *lm, GString *result)
 
   if (self->options->truncate_size != -1 && result->len > self->options->truncate_size)
     {
+      const gint truncated_bytes = result->len - self->options->truncate_size;
+
       g_string_truncate(result, self->options->truncate_size);
+
+      stats_counter_inc(self->truncated.count);
+      stats_counter_add(self->truncated.bytes, truncated_bytes);
     }
 }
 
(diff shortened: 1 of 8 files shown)


kira-syslogng:
Build SUCCESS

bazsi (approved):
This looks good to me.

news/feature-3474.md
```diff
@@ -0,0 +1,5 @@
+logwriter based destinations: new truncate_size option introduced to truncate an output message to a specified max size. default value is -1 (disabled).
```
MrAnno, line 1: I think we should just list those destinations, `logwriter` does not say much for an end-user:

`file, network, program destinations: new truncate-size() option`. 

Can we use dashes in option names, please?

news/bugfix-3474.md
```diff
@@ -0,0 +1 @@
+network/udp: message was lost (not sent) if it was too large, and a time reopen amount of time needed to expire to send the next message lowering the thoughtput.
```
MrAnno, line 1: Now, it's truncated at ~64k.

modules/afsocket/afinet-dest.c
```diff
@@ -50,6 +50,8 @@
 #  define _GNU_SOURCE 1
 #endif
 
+static const gint MAX_UDP_PAYLOAD_SIZE = 65000;
```
MrAnno, line 53: max datagram is `65535 - (IP + UDP header)`, not considering jumbograms 👍🏻

MrAnno (approved):

news/feature-3474.md
Kokan: Sure.

lib/logwriter.c
```diff
@@ -1093,6 +1093,11 @@ log_writer_format_log(LogWriter *self, LogMessage *lm, GString *result)
         }
 
     }
+
+  if (self->options->truncate_size != -1 && result->len > self->options->truncate_size)
+    {
+      g_string_truncate(result, self->options->truncate_size);
```
gaborznagy, line 1099: I wonder if we could help the user with a warning message.
Our default `log-msg-size()` is 64k, so this problem can happen when the `log-msg-size()` is increased or when using `template()`.

I know emitting warning/debug messages on each message can lower the performance, and spam the internal() log...
I'm just thinking out loud.

lbudai: maybe a stat counter could be added for tracking number of truncated messages (and maybe another counter could be useful: truncated bytes - but I think the first ctr would be more than enough)

modules/afsocket/afinet-dest.c
Kokan: If the user comes close to this number they should just adjust it acknowledging that they know what they are doing. Is it okay if we leave it just that or you want to set exactly the limit ?

lib/logwriter.c
bazsi: I would prefer the counter.

modules/afsocket/afinet-dest.c
bazsi: TIL: I didn't know that ipv6 supported datagrams above 64k (up to 4GiB-1, a bit crazy)

In any case UDP on top of ipv6 would only support up to 64k as the UDP header contains a 16 bit value of the same.

I would personally either default to 1024 (which would probably be unexpected, but that's the limit in the RFC), or the exact value. 65000 is difficult to reason about.

Kokan: 65507 it is.

lib/logwriter.c
Kokan: The following counters were added (for all logwriter based destination)
```
dst.network;d_local#0;udp,127.0.0.1:1111;a;truncated_count;1
dst.network;d_local#0;udp,127.0.0.1:1111;a;truncated_bytes;1
```

kira-syslogng:
Build SUCCESS

modules/afsocket/afinet-dest.c
bazsi: Thanks.

bazsi (approved):

lbudai (approved):
Thanks the counters :)

PR #3476, opened 2020-11-05
Dbld: devshell on tarball

gaborznagy:
Today, when the topic of the failing devshell:distcheck github actions job [see log](https://github.com/syslog-ng/syslog-ng/runs/1352540423?check_suite_focus=true) came up,  we've agreed that devshell should be moved to a rolling release, precisely debian-testing, as the tarball generation.

Question: do we need perf packages in a docker container? Isn't `perf` dependent on the kernel, which in case of docker is the host's kernel?


gaborznagy:

kira-syslogng:
Package Build SUCCESS

Packages are available at https://ci.syslog-ng.com/jenkins/job/kira-dbld-pull-requests/725/artifact/

gaborznagy:
devshell/distcheck job will not be green until it's not merged. It would be good to see it green before, but at the moment I don't have any idea besides a temporary docker build...

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

dbld/rules
```diff
@@ -189,7 +189,7 @@ shell-%: setup
 images: $(foreach image,$(IMAGES), image-$(image))
 image: image-$(DEFAULT_IMAGE)
 image-kira: image-ubuntu-focal
-image-devshell: image-ubuntu-focal
+image-devshell: image-tarball
```
lbudai, line 192: I think we could add `doxygen` and `docbook-xsl` to [devshell] in `package.manifest`.

The question is what is `distcheck`... if we think that this is something related to devshell (IMHO this is the case), then aforementioned packages should be part of devshell.
Or we can create a ci-shell, based on devshell - but I think devshell would be better.

After we solve this distcheck problem: what is the difference between tarball and devshell images?

gaborznagy: We discussed this with @bazsi and thought that basing `devshell`  on a rolling release would have advantages (should be same distro on which we do release tarball, which is debian testing).

Those packages should be part of devshell, and `distcheck` should run on `devshell` image, but they are installed on `tarball` images (which is an abstraction of debian testing).

`devshell` image doesn't have to be kept minimal contrary to `tarball`.

gaborznagy:
marking as WIP until we don't see a green devshell distcheck job (currently the PR uses my docker image and not the official)

dbld/rules
lbudai: we should run `make distcheck`  where we run `make dist`

kira-syslogng:
Package Build SUCCESS

Packages are available at https://ci.syslog-ng.com/jenkins/job/kira-dbld-pull-requests/727/artifact/

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

gaborznagy:

kira-syslogng:
Build SUCCESS

bazsi:
I have found further issues in make dist in case Python is missing, this fixed it for me:

```
diff --git a/modules/python/Makefile.am b/modules/python/Makefile.am
index ddf69a1d3..687dbfa7a 100644
--- a/modules/python/Makefile.am
+++ b/modules/python/Makefile.am
@@ -79,7 +79,6 @@ modules_python_libmod_python_la_DEPENDENCIES= \
 
 modules/python mod-python: modules/python/libmod-python.la
 
-include modules/python/tests/Makefile.am
 else
 modules/python mod-python:
 endif
@@ -100,3 +99,5 @@ EXTRA_DIST                                   += \
        modules/python/CMakeLists.txt
 
 .PHONY: modules/python mod-python
+
+include modules/python/tests/Makefile.am
diff --git a/modules/python/pylib/Makefile.am b/modules/python/pylib/Makefile.am
index 5ce539f95..c2a38f8af 100644
--- a/modules/python/pylib/Makefile.am
+++ b/modules/python/pylib/Makefile.am
@@ -1,5 +1,3 @@
-if ENABLE_PYTHON
-
 EXTRA_DIST += \
        modules/python/pylib/syslogng/__init__.py                               \
        modules/python/pylib/syslogng/debuggercli/__init__.py                   \
@@ -41,6 +39,9 @@ EXTRA_DIST += \
        modules/python/pylib/test_pylib.sh      \
        modules/python/pylib/CMakeLists.txt
 
+if ENABLE_PYTHON
+
+
 PYLIB_PATH = modules/python/pylib
 PYLIB_BUILDDIR = $(abs_builddir)/$(PYLIB_PATH)
 PYLIB_SRCDIR = $(top_srcdir)/modules/python/pylib
```

bazsi:
This branch should revert the installation of doxygen et al from the devshell image.

kira-syslogng:
Build SUCCESS

gaborznagy:
travis job got duplicated (notice the `push` job beside the `pr`) as I've accidentally pushed my branch onto upstream, and deleting the branch didn't help to remove the job from the PR.
So, to make the travis/push job green, I had to push the branch to upstream.

kira-syslogng:
Package Build SUCCESS

Packages are available at https://ci.syslog-ng.com/jenkins/job/kira-dbld-pull-requests/729/artifact/

kira-syslogng:
Build FAILURE

gaborznagy:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

gaborznagy:
Tested devshell CI job with re-built dbld devshell image: https://github.com/syslog-ng/syslog-ng/actions/runs/415927152

kira-syslogng:
Build SUCCESS

dbld/builddeps
```diff
@@ -222,7 +222,11 @@ function enable_dbgsyms {
 }
 
 function install_perf {
```
szemere, line 224: I do not know the original intention behind `install_perf` (It looks like it wanted to handle package name differences between debian based distributions.), but currently we use the `packages.manifest` file for that.

My suggestion is to remove the `install_perf` function completely. (currently only the devshell.dockerfile calls it)

szemere (dismissed):
My `install_perf` comment is optional. We can merge this without that change.

dbld/builddeps
bazsi: The perf tools depend on the kernel versions available in those distributions, and I would rather not maintain them in packages.manifest, as the set of kernel versions supported in a distro might change, as distros regularly add extra kernels.

dbld/rules
bazsi: Agreed. This means that we need to move distcheck from this workflow to somewhere else. Right now it runs on devshell, but with the latest fixes that should work too.

However moving distcheck to somewhere else should be a separate PR.

modules/python/tests/test_python_template.c
```diff
@@ -53,7 +53,9 @@ _py_init_interpreter(void)
   }
   PyGILState_Release(gstate);
 
+#if PY_MAJOR_VERSION == 2 || (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION < 7)
   PyEval_InitThreads();
+#endif
```
bazsi, line 58: This might be extracted into the python compat code so we wouldn't have these ugly ifdefs.

bazsi (dismissed):
The Python compat comment would be nice to have, but I am ok with merging this as is.

dbld/builddeps
gaborznagy: As far as I see, `install_perf` was introduced with a157ec41493a147988049634c127538ec29b8feb in #1950.
I still don't know how could you use perf, as your host kernel might be totally different...

@szemere 
One does not simply move the content of `install_perf` into packages.manifest, as we have a logic here to select the correct package name and not a package name. Run this command on a debian/ubuntu distro: `apt-cache search --names-only "linux-tools.*generic"`.

@bazsi Can you clarify this, please? You introduced `perf` in `devshell` and I don't know someone else who is using it.

modules/python/tests/test_python_template.c
gaborznagy: Good point, I was just quickly throwing stuff together to see a green light on github actions side. :)

gaborznagy:
I'll do the python part quickly, please stand by.

modules/python/tests/test_python_template.c
gaborznagy: Done. Thanks for pointing out the compat layer. :)

kira-syslogng:
Build SUCCESS

gaborznagy:
Python compat layer is done, I've answered the `perf` comment, but I don't know how to resolve it.
I've fixed it in a way to have `perf` tool on debian, but I don't know if we should include it in a docker image.

dbld/builddeps
gaborznagy: Ooops, I didn't saw @bazsi's comment before writing mine.
From that I conclude that we should still add `perf` to devshell.

dbld/rules
gaborznagy: I've added this as a to-do: https://github.com/syslog-ng/syslog-ng/issues/3517

szemere (approved):

alltilla (approved):

PR #3484, opened 2020-11-14
WIP: github: use the new $GITHUB_ENV file instead of ::set-env

bazsi:

PR #3487, opened 2020-11-17
github-actions: use new env API

alltilla:
Based upon #3484. I have opened a new PR, because this problem just become critical, as `set-env` is removed, and nearly all of our CI jobs fail.

### DIFF
--- a/.github/workflows/devshell.yml
+++ b/.github/workflows/devshell.yml
@@ -16,16 +16,12 @@ jobs:
       fail-fast: false
 
     steps:
+      - name: Checkout syslog-ng source
+        uses: actions/checkout@v2
+
       - name: Set ENV variables
         run: |
-          gh_export() {
-            while [ "x$1" != "x" ]; do
-              echo "$1<<__EOF__" >> $GITHUB_ENV
-              eval echo \$$1     >> $GITHUB_ENV
-              echo "__EOF__"     >> $GITHUB_ENV
-              shift;
-            done
-          }
+          . .github/workflows/gh_export.sh
 
           PYTHONVERSION="`printf ${{ matrix.python }} | tail -c 1`"
           PYTHONUSERBASE="${HOME}/python_packages"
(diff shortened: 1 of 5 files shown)


alltilla:
distcheck fails:
```make[4]: *** No rule to make target 'modules/python/pylib/test_pylib.sh'.  Stop.```
Is this a new problem?

**Comment by kira-syslogng** [comment]
Build SUCCESS

File: .github/workflows/devshell.yml [hunk 1]
```diff
@@ -144,7 +144,7 @@ jobs:
         run: ./autogen.sh
 
       - name: configure
-        run: ./configure --disable-all-modules
+        run: ./configure ${DISTCHECK_CONFIGURE_FLAGS}
```
gaborznagy: This patch actually is needed. We discussed this with @bazsi that we should test `make distcheck` with a tarball that is built in a minimal environment.
`make distcheck` will use a different set of configure flags through `DISTCHECK_CONFIGURE_FLAGS`
He even found an issue around Python, IIRC.
We can discuss this IRL.

gaborznagy:
> distcheck fails:
> `make[4]: *** No rule to make target 'modules/python/pylib/test_pylib.sh'. Stop.`
> Is this a new problem?

I guess that's what @bazsi found when there is no python when the tarball is created: https://github.com/syslog-ng/syslog-ng/pull/3484/commits/07c4103a0292b7ba7cab70ce60d83443a1854075

kira-syslogng:
Build SUCCESS

.github/workflows/devshell.yml
alltilla: Oh, I see. That is another kind if change, tho, not related to the github actions env API.

bazsi: yup. I've just found a few other cases. These might go into a separate PR though.

gaborznagy (dismissed):

gaborznagy (dismissed):

alltilla:
I moved the `gh_export()` function to a separate file, which just needs to be sourced. If the CI jobs became green, I would consider the PR final.

kira-syslogng:
Build SUCCESS

gaborznagy (approved):

lbudai (approved):

bazsi:
I like the way you could share code with the steps (e.g. gh_export.sh). I was looking for a means to do this, but couldn't find a solution. Thanks.

PR #3490, opened 2020-11-18
dbld: bump versioning.h in prepare-release

alltilla:
Note: it is enough to merge this PR after releasing 3.30.1, if we manually do these bumps this time for 3.30.1.

### DIFF
--- a/dbld/prepare-release
+++ b/dbld/prepare-release
@@ -6,8 +6,12 @@ cd /source
 VERSION=$1
 
 OLD_VERSION=`cat VERSION`
-OLD_MAJOR=$(echo $OLD_VERSION | cut -d. -f1-2)
-NEW_MAJOR=$(echo $VERSION | cut -d. -f1-2)
+OLD_MAJOR=$(echo $OLD_VERSION | cut -d. -f1)
+OLD_MINOR=$(echo $OLD_VERSION | cut -d. -f2)
+OLD_PATCH=$(echo $OLD_VERSION | cut -d. -f3)
+NEW_MAJOR=$(echo $VERSION | cut -d. -f1)
+NEW_MINOR=$(echo $VERSION | cut -d. -f2)
+NEW_PATCH=$(echo $VERSION | cut -d. -f3)
 USER_EMAIL=$(git config --get user.email || echo "noemail@set.any.where")
 USER_NAME=$(git config --get user.name || echo "unknown user")
 VERSION_FILE_LIST="README.md doc/man/*.xml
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

dbld/prepare-release
```diff
@@ -6,8 +6,12 @@ cd /source
 VERSION=$1
 
 OLD_VERSION=`cat VERSION`
-OLD_MAJOR=$(echo $OLD_VERSION | cut -d. -f1-2)
-NEW_MAJOR=$(echo $VERSION | cut -d. -f1-2)
+OLD_MAJOR=$(echo $OLD_VERSION | cut -d. -f1)
+OLD_MINOR=$(echo $OLD_VERSION | cut -d. -f2)
+OLD_PATCH=$(echo $OLD_VERSION | cut -d. -f3)
```
szemere, line 11: I like the symmetry in the code (also extracting PATCH from the version number), but unfortunately PATCH is never used. Could you give an example for that?

szemere (approved):

dbld/prepare-release
alltilla: Yes, I extracted `PATCH` for the sake of completeness. I did not want to split the `VERSION` to 3 parts, then only provide 2 of them with variables.

alltilla:
The logic here needs to be validated in the `dbld/rules release` step. I will do that this week.

alltilla:
Ready for review.

kira-syslogng:
Build SUCCESS

gaborznagy:
I've just noticed that `dh-systemd` is missing during docker image creation, which causes that [tarball](https://github.com/syslog-ng/syslog-ng/pull/3490/checks?check_run_id=1537112749) github actions job breaks.

Strangely, I don't find dh-systemd in testing(bullseye), but it's in Sid though.
https://packages.debian.org/search?searchon=names&keywords=dh-systemd

alltilla:
@gaborznagy It is interesting. I think it is not related to this PR, but it definitely needs a fix, regardless.

gaborznagy:
> @gaborznagy It is interesting. I think it is not related to this PR, but it definitely needs a fix, regardless.

`dh-systemd` seems to be a transitional package in `debian buster` [already](https://packages.debian.org/buster/dh-systemd), which means we can delete it.
I've checked `debhelper` package and it seems to contain the necessary binaries:
```
/# cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
root@minke:/# 


root@minke:/# dpkg -L dh-systemd
/.
/usr
/usr/share
/usr/share/doc
/usr/share/doc/dh-systemd
/usr/share/doc/dh-systemd/changelog.gz
/usr/share/doc/dh-systemd/copyright


root@minke:/# dpkg -L debhelper | grep systemd
/usr/bin/dh_installsystemd
/usr/bin/dh_installsystemduser
/usr/bin/dh_systemd_enable
/usr/bin/dh_systemd_start

```

gaborznagy:
I'll forgot to say, I'll open a PR once I've verified it works.

bazsi (dismissed):
I like that we are moving the validation logic into a separate script, outside of the makefile.

szemere (approved):

alltilla:
Rebased to master, to obtain the fix for "Binary packages".

kira-syslogng:
Build SUCCESS

szemere (approved):

bazsi (approved):

gaborznagy (approved):

PR #3493, opened 2020-11-22
GitHub actions use minimal configure flags for distcheck

bazsi:
This PR contains the following set of changes:
  - a minor refactor of the recent gh_export.sh addition, renaming gh_export.sh to gh-tools.sh and also extracting another github related helper function (gh_path)
  - the addition of --enable-manpages-install, along with --enable-manpages that allows all manpage related use-cases to operate (see the commit for more details
  - Minimizing the configure run for distcheck, so we can actually validate that all files make it into the dist even if no optional features are enabled.

This PR is in favour of #3484 which I am closing in a minute.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

.github/workflows/devshell.yml
```diff
@@ -143,7 +143,7 @@ jobs:
         run: ./autogen.sh
 
       - name: configure
-        run: ./configure ${DISTCHECK_CONFIGURE_FLAGS}
+        run: ./configure --disable-all-modules --enable-manpages
```
gaborznagy, line 146: Shouldn't `DISTCHECK_CONFIGURE_FLAGS` use the new `--enable-manpages-install` instead of `--enable-manpages`?

bazsi: No, distcheck needs the manual pages and git does not have them. So in order to successfully generate "make dist", we will need --enable-manpages.

alltilla (dismissed):

.github/workflows/devshell.yml
gaborznagy: Yes, you are right. `DISTCHECK_CONFIGURE_FLAGS` is used when the tarball is created, not just when make tries to build the package from the tarball.
https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html

.github/workflows/gh-tools.sh
```diff
@@ -7,3 +7,8 @@ gh_export()
         shift;
     done
 }
+
+gh_path()
+{
+    echo "$1" >> $GITHUB_PATH
```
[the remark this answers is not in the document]
bazsi: Looking at the context, right now the environment setup is a dedicated step, does not run commands at all. I am not completely against it though, but I would wait for a concrete use, where this would make sense.

kira-syslogng:
Build SUCCESS

alltilla (approved):

szemere (approved):

PR #3495, opened 2020-11-27
Version: 9.99.1

kira-syslogng:
Tarball is available at: https://github.com/syslog-ng/syslog-ng/actions/runs/386667146


kira-syslogng:
Build SUCCESS

PR #3502, opened 2020-11-28
dbld/rules: remove deb/rpm build from the "release" target

bazsi:
We don't publish .deb/.rpm as a part of release (only a tarball), so
remove them to make builds faster.


kira-syslogng:
Build SUCCESS

gaborznagy (approved):
This is okay to be removed, since package building will be checked during release anyway once #3464 is merged, as that adds package building CI jobs to every pull request, hence to the version bump PR too.

alltilla (approved):
Merge after https://github.com/syslog-ng/syslog-ng/pull/3464

PR #3504, opened 2020-12-01
glib-compat: fix g_cond_wait_until() with older glib versions (< 2.32) 

MrAnno:
When compiling syslog-ng with old glib versions (< 2.32), our compactibility implementation of `g_cond_wait_until()` is used.
The implementation of the compat call had a bug, which could cause indefinite waiting when `loggen` tries to connect to a TCP/TLS host.
https://github.com/syslog-ng/syslog-ng/blob/f7812c4a1d2fd9e4402f25e85f0cf5940500d50f/tests/loggen/socket_plugin/socket_plugin.c#L199

----

The deprecated `g_cond_timed_wait()` call operates with system wall-clock time, passing `GTimeVal` based on system monotonic time is not valid.

We've found this issue with `tests/functional/test_performance.py`. This test failed due to undefined, but very short timeouts.

A code snippet to reproduce the problem:
https://gist.github.com/MrAnno/6b4ec783a91975177a1028bd357fddbc

Docs:
- https://developer.gnome.org/glib/stable/glib-Threads.html#g-cond-wait-until
- https://developer.gnome.org/glib/stable/glib-Threads.html#g-cond-timed-wait

### DIFF
--- a/lib/compat/glib.c
+++ b/lib/compat/glib.c
@@ -344,8 +344,12 @@ g_thread_new(const gchar *name, GThreadFunc func, gpointer data)
 gboolean
 g_cond_wait_until (GCond *cond, GMutex *mutex, gint64 end_time)
 {
-  glong diff_in_sec = (end_time - g_get_monotonic_time())/G_TIME_SPAN_SECOND;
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

alltilla (approved):

furiel (approved):

PR #3507, opened 2020-12-03
autotools: fix shadowed test failures

gaborznagy:
The LOG_COMPILER wrapper script is used to workaround a criterion bug, which prevents to detect if the test case exit status shows an error.
This was introduced when autotools sanitizer support added in #3440 .

However, the test-grep.sh wrapper script shadows the exit status of the test case causing false positive results, so we need to propagate this back to Make.
CMake were not affected by this, I've double-checked it (see it on [my fork](https://github.com/gaborznagy/syslog-ng/actions/runs/398144520). CMake tests had failed as expected, while autotools shows false positive tests).

### Question:
Initially, I was about to disable the custom test wrapper `LOG_COMPILER` in case no sanitizer is used.
Later I've changed my mind. I wanted to make the test evaluation as close as possible in both cases (with or without sanitizer).
Should we still disable LOG_COMPILER usage in case sanitizer is not used?

autotools LOG_COMPILER documentation for reference:
https://www.gnu.org/software/automake/manual/html_node/Parallel-Test-Harness.html#Parallel-Test-Harness


kira-syslogng:
Build SUCCESS

scripts/test-grep.sh
```diff
@@ -22,13 +22,21 @@
...
+exit_status=$?
 
 if egrep -q 'ERROR: (LeakSanitizer|AddressSanitizer)' $1.result; then
    echo "SAN report detected"
    exit 1
 fi
 
+# Put error message into test-suite.log too, 
+# on passed cases, output is not redirected into test-suite.log
+cat $1.result
```
Kokan, line 35: This should be before the `if egrep -q ...`, as the script could exit, before the cat.

gaborznagy: Good point. I've fixed it.

Kokan (dismissed):

kira-syslogng:
Build SUCCESS

MrAnno (approved):
Oh...
Thank you for fixing this.

> Should we still disable LOG_COMPILER usage in case sanitizer is not used?

I don't think we'll ever touch this script again, but I'd prefer NOT using it in non-sanitizer tests.
Do you want to change that, or should we just merge the PR?

MrAnno (approved):

gaborznagy:
The three of us talked a lot about this and had good reasons whether to use the `test-grep.sh` by default or not.
Personally, I would have kept it to make sure the script is not going to break, but I admit, if `test-grep.sh` is needed only for sanitizer-enabled use cases (because of Criterion), then use it only in that cases.

I had to test this script manually with a modified test having a  `cr_assert(FALSE)` statement to see if it can report failing tests (each time with and without sanitizer). So if anything is still broken, it would be sad to repeat the whole testing process.
With this only the sanitizer use case left when test-grep.sh should be manually tested (or some tests should be added for the script.)

Nevertheless, we should discuss if sanitizers should be used in CI.

kira-syslogng:
Build SUCCESS

Kokan (approved):

PR #3511, opened 2020-12-08
Light: Refactor File() and FileIO() related classes

mitzkia:
This is a bigger refactor in internal File() and FileIO() API.
Also removes MessageReader() and DestinationReader() classes.

Internal reasons:
- before this patch-set FileIO() API was wrongly used in every places in
Light where files was read or written
- to make File() API clearer it was moved to 'common' directory, and it has been
extended with general File releated APIs
- FileIO() API has been reduced to simple read() and write() which is
used from File() with composition
- This patch-set fixes the relationship between particular Light classes and the
File() or FileIO() classes

Only syslog-ng driver related classes (where syslog-ng driver writes the output file) should use FileIO(), such as:
 - FileDestination(),
 - ExampleDestination()
Every other class should use the native File() class.

We have decided to drop MessageReader() and SingleLineParser() classes
because they were unnecessarily complex.
Finally we have also dropped DestinationReader() class.


kira-syslogng:
Build SUCCESS

gaborznagy:
I plan to take a look into this on this Friday.

gaborznagy (commented):
I haven't finished yet, I'll finish the review next week. I still have to go through the `MessageReader` patches.

Looking through the patches in general, I think `File` and `FileIO` classes cleared up and their interface become much cleaner in this PR.

tests/python_functional/src/common/blocking.py
```diff
@@ -23,21 +23,21 @@
 import time
 
 DEFAULT_TIMEOUT = 20  # 20 sec
-POLL_FREQ = 0.001  # 1 millisecond
+POLL_FREQ = 0.001  # 10 millisecond
```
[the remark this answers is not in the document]
mitzkia: Yes, it is should be 1 millisecond. I will fix it.

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
```diff
@@ -65,7 +65,10 @@ def write_config(self, config_path):
         else:
             rendered_config = ConfigRenderer(self.__syslog_ng_config).get_rendered_config()
         logger.info("Generated syslog-ng config\n{}\n".format(rendered_config))
-        FileIO(config_path).rewrite(rendered_config)
+        f = File(config_path)
```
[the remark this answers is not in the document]
mitzkia: You are absolutely right, we can use the create_file() here instead of write this 4 lines (create + write + close file).
There is one minor problem, in create_file() there is a check if the written file is already exist or not, and the function raises if the file exist. But we in syslog_ng_config.py file need to overwrite the original file (before syslog-ng reload, config could changed), so if we start using create_file() here, we should extend it with a new parameter: 'overwrite=False|True', or we should create a new function which will overwrite the existing file: overwrite_file(file, content). 
What do you think, which solution would be better?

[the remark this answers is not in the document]
gaborznagy: IMHO neither of them. My idea was bad.
As far as I understand `common.operations` is functioning as a wrapper in case some python2/python3 compatibility and as a module for high-level helper functions.

I would remove `create_file` from `common.operations` as nobody uses it currently, and also because I don't see any reason why `create_file` couldn't be put into the `File` class.
Nevertheless, don't put `create_file` into `File` either, please. 

I would fancy if `File` could be used within a `with` statement:
```
with File(config_path, "w+") as config_file:
    config_file.write(rendered_config)
```
This is optional to do in this PR, I'm okay with the current form, we can enhance the `File` class when we see the need for it, but please remove `create_file` from `common.operations`

tests/python_functional/src/common/file.py
```diff
@@ -22,33 +22,84 @@
...
-            logger.debug("File has been created:\n{}".format(self.__file_path))
+            logger.debug("File has been created: {}".format(self.path))
         else:
-            logger.debug("File not created:\n{}".format(self.__file_path))
+            raise Exception("File was not created in time: {}".format(self.path))
         return file_created
 
     def open(self, mode):
-        self.__opened_file = open_file(self.__file_path, mode)
+        self.__opened_file = open_file(self.path, mode)
```
gaborznagy, line 53: Can we move `open_file` here from `common.operations`? It could be a local helper function.
As I see, `File` class usage is cleared up in Light: everyone uses the `File` class and there's no other user of `open_file`

tests/python_functional/src/driver_io/file/file_io.py
```diff
@@ -26,20 +26,16 @@
...
     def __init__(self, file_path):
         super(FileIO, self).__init__(file_path)
-        self.__readable_file = None
+        self.__readable_file = File(file_path)
         self.__writeable_file = File(file_path)
 
-    def read(self):
-        if not self.__readable_file:
-            self.__readable_file = self.open(mode="r")
+    def read(self, counter):
```
gaborznagy, line 31: Do we need a mandatory counter for read operation?
I see one use case where a mandatory `counter` in `FileIO::read` cannot be used: when a test with a file() destination wants to assert on the number of lines in a file (or wants to read the entire file).

In `functional_tests/logpath/test_multiple_flags.py` a pattern can be seen that the test reads exactly the same amount of lines from the beginning of the file that was previously written into it. But how about a misconfigured filter() in syslog-ng's config which leads that the destination file has more lines in it?
I would like to do the same which is used in SNMP related test cases, see below.

In `SNMPtrapd` (helpers/snmptrapd/conftest.py), `get_log` does a raw readline() without any wait_for_lines methods. The test asserts on if the expected log is in the read data.
See `functional_tests/destination_drivers/snmp_destination/v2c/test_snmp_destination_v2c_custom_community.py`:
```
    assert expected_traps == snmptrapd.get_traps(len(expected_traps))
    assert test_community in snmptrapd.get_log()
```
There's also a `get_traps` method which is similar to `FileIO:read` in a sens that the user must provide the expected number of results that is expected.
Of course `SNMPtrapd` uses `File` class, therefore it has flexibility.
(I'm not suggesting that SNMPtrapd should be using FileIO instead)

My first question can be answered if the purpose of `FileIO` class is answered.
I think `FileIO` provides a higher-level interface to its users, who needs more than `File` class's raw I/O operations (however File just got some wait_for_lines methods).
If that is the case `FileIO:read()` either should be renamed to `wait_for_lines` or it should do raw I/O operations as its name suggests.

tests/python_functional/functional_tests/destination_drivers/snmp_destination/general/test_snmp_destination_acceptance.py
```diff
@@ -39,6 +39,6 @@ def test_snmp_dest_acceptance(config, syslog_ng, snmptrapd, snmp_test_params):
 
     syslog_ng.start(config)
 
-    received_traps = snmptrapd.get_traps()
-    assert snmp_test_params.get_expected_basic_trap() == received_traps
-    assert any(snmp_test_params.get_default_community() in line for line in snmptrapd.get_raw_traps())
+    expected_traps = snmp_test_params.get_expected_basic_trap()
+    assert expected_traps == snmptrapd.get_traps(len(expected_traps))
+    # assert any(snmp_test_params.get_default_community() in line for line in snmptrapd.get_raw_traps())
```
gaborznagy, line 44: is this a leftover?

mitzkia: Yes, this should be deleted. I will fix it.

tests/python_functional/src/driver_io/file/file_io.py
mitzkia: Thanks for this review comment, interesting question.
I am agreed with you in that for now we can not achieve such use-case for file destinations where we want to assert on that how many lines are written by syslog-ng (with the help of FileDestination class).
If I understand you correctly you did not pointing that this is a loosed feature from the past (this case was not covered also before this PR). Just for the note before this PR there was a function DestinationReader::read_all_logs which was used only by SNMPtrapd class.
Mainly this PR wanted to keep the original features for the related syslog-ng drivers: FileDestination and SNMPtrapd destination. If this is not applied please notice it.

The purpose of FileIO class is the following:
- it provides only IO related functions for file based syslog-ng drivers such: reading/writing files
- this class should not be used from non FileBased drivers such: SyslogNg ConsoleLogReader, or SNMPtrapd and so on
- this class should be used only in such syslog-ng destination or source drivers where syslog-ng reads from or writes to files
- from my point of view this class can have several read/write methods, what file based drivers can dictate, and in this case I also can imagine such a method: read_all() which would returns with the entire file content.


The purpose of File class is the following:
- this class provides such general functions which are related to native File objects not specially to syslog-ng drivers, where syslog-ng drivers are not affected
- can be used in a following places where syslog-ng drivers are not in context: ConsoleLogReader (reading sterr of syslog-ng), or SNMPtrapd (reading output of snmptrapd tool), SyslogNgConfig (writing syslog-ng config file)
- I think it is easier to think about this class such a general "helper" class for native file based objects

All in all I would extend later FileIO with read_all() method (as you suggest). For now without this read_all() SyslogNgCtl class can help to ensure if syslog-ng writes the correct number of lines, example:
```
assert snmp_destination.get_stats() == {'written': expected_message_counter, 'processed': expected_message_counter, 'dropped': 0, 'queued': 0}
```

Do you agreed with this concept, or was it helpfull?

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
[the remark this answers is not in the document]
mitzkia: I am ok with removing `create_file()` from `common.operations` and replace it `with` statement (as you suggested),  but if there will be another place where we should create a file with some content I would put this method to the `File()` class (but would we remember for this?).
What do you think?

tests/python_functional/src/common/file.py
mitzkia: Sorry, if I misunderstood you suggest that to move: `common.operations::open_file()` to `common.file::open_file()` as a local function? If this is the idea, I am ok with that, but I would move all the other general file related functions also such: copy_file(), copy_shared_file(), ...

What do you think? Do you agreed with that?

gaborznagy: Yes please, that's what I tried to say. :)
Of course, other functions can remain in `common.operations`.

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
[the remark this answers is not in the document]
gaborznagy: Yes, we should remove `create_file` from `common.operations`, no one uses it.
Thanks if you like the idea of the `with` statement. I would totally accept if you would like to do it in another PR.

About the question of future use cases of `create_file` method: I don't think it's needed, especially after having the `with` statement support, but that's just my opinion (although this PR doesn't use `create_file` in any place either). 
Users can see the interface of the `File` class (e.g. the `open`, `close` methods) so we would remember it.

tests/python_functional/src/driver_io/file/file_io.py
gaborznagy: Thanks for discussing this with me thoroughly!

As you ask, one functionality did change with this PR: `FileIO:read()` methods definition has a mandatory `counter` argument. I'm not against this as you've stated the reasons behind it. 
```diff
-    def read(self):
-        if not self.__readable_file:
-            self.__readable_file = self.open(mode="r")
+    def read(self, counter):
```
As you mentioned, `FileIO` class can be extended with a `read_all` method when/if needed. It would simply call the underlying `File:read()` method.
Thanks, this is resolved by me.

mitzkia:
@gaborznagy Thanks for the review comments. 
I have fixed all findings.
- `src.common.blocking`: 1 ms typo (the appended comment was a typo)
- `test_snmp_destination_acceptance.py`: commented assert was removed
- `src.syslog_ng_config.syslog_ng_config`: using `with` statement for file creation and content writing
- `src.common.operations`: `create_file()` method was removed (it was unused)
- `src.common.operations`: open, copy, delete file methods was moved to `src.common.file`

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/destination_drivers/snmp_destination/general/test_snmp_destination_acceptance.py
mitzkia: I have fixed this.

tests/python_functional/src/common/file.py
mitzkia: I have fixed this.

tests/python_functional/src/common/blocking.py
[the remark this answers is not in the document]
mitzkia: I have fixed this.

tests/python_functional/src/syslog_ng_config/syslog_ng_config.py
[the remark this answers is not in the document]
mitzkia: I have fixed this.

tests/python_functional/src/driver_io/file/file_io.py
mitzkia: You are right, this part of FileIO API has changed.
But from the testcase side it "remains" same logic:

This is how before this PR file destination read has working:
```
file_destination:read_log() -> file_destination:read_logs(counter) -> destination_reader:read_logs(counter) -> message_reader:pop_messages(counter)
```

After this PR file destination read will work in this way:
```
file_destination:read_log() -> file_destination:read_logs(counter) -> fileio:read(counter)
```

mitzkia: And about the `destination_reader:read_all_logs`, yes it has been removed, but in one of the next PR-s it will be added (as I count)

kira-syslogng:
Build SUCCESS

mitzkia:
Due to the yesterdays Light merges, this PR has broken and I should add new minor "features" top of the original PR.

- `common:file` got new method: `wait_for_any_content()` - this will be needed for the next new element - squashed to 0dbd026e4c58a418ba9e53350b422654a967186a
- `driver_io:file:file_io` got new method: `read_all()` - this will be needed for reading entire file (for example example destination output file - when the source is our internal source) - squashed to dbaea43a052fc4da031f7a1d1fed84ece6c61f6c
- `syslog_ng_config:statements:destinations:file_destination` got new method: `read_all_logs()` - just for the symmetric - squashed to fe5ed4a5ace12205f0be87aede7185348364485b
- `syslog_ng_config:statements:destinations:example_destination` got new method `read_all_logs()` - the whole changeset was because of this. In the new internal source testcase Light should be able to read from the output file of the example_destination driver. But when it generates it does not mean it has a content in it. So we should wait with read (read_all_logs in this case) until there will be any content in the file. Without this the testcase will be flaky, and it can be happen Light can not read anything from the file because the content is not there yet. - squashed ca078ce79faa36996d6be2ac655835ba68bc75af

kira-syslogng:
Build SUCCESS

gaborznagy (dismissed):

kira-syslogng:
Build SUCCESS

mitzkia:
After discussing the latest changes. we decided to drop the `wait_for_any_content()` and `read_all()` related functions (which I added with my latest changeset), because we already have a satisfying function called `common:file:wait_for_lines(self, lines)`. With the help of this function we can read files until an expected message appears in it. This is what we needed in case of internal source + example destination testcase. 
I have exposed this function as a `file_io:read_until_lines()`. Example destination and file destination classes will wrap this function as: `read_until_logs(self, logs)`, where `logs` is a list of messages what we expect in a file

pzoleex:

kira-syslogng:
Build triggered

kira-syslogng:
Build started
Build FAILURE

pzoleex:

kira-syslogng:
Build SUCCESS

mitzkia:
@kira-syslogng test this please test branch=micek_fix_collided_port_allocation;

kira-syslogng:
Build FAILURE

gaborznagy:

mitzkia:
@kira-syslogng test this please test branch=micek_remove_get_random_port_allocator;

kira-syslogng:
Build FAILURE

mitzkia:
@kira-syslogng test this please test branch=micek_remove_get_random_port_allocator;

kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng_config/statements/destinations/file_destination.py
```diff
@@ -25,21 +25,23 @@
...
     def get_path(self):
         return self.path
 
     def read_log(self):
-        return self.destination_reader.read_logs(self.get_path(), counter=1)[0]
+        return self.read_logs(1)[0]
 
     def read_logs(self, counter):
-        return self.destination_reader.read_logs(self.get_path(), counter=counter)
+        return self.io.read_number_of_lines(counter)
```
MrAnno, line 44: `MessageReader` meant to be an abstraction that could have been able to read/buffer data from a stream-like medium and extract messages from it based on a configurable separator. What about multi-line and framed messages now?

tests/python_functional/src/common/file.py
```diff
@@ -22,33 +22,84 @@
...
+
+    def readline(self):
+        if not self.is_opened():
+            raise Exception("File was not opened before trying to read from it.")
+        return self.__opened_file.readline()
+
+    def read(self):
+        content = ""
+        while True:
+            buffer = self.readline()
```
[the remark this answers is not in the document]
mitzkia: Thanks for the review comment. 
I have checked this and it works with read() also. So I will fix this to use read().
As I remember there was an other place where we should use readline() instead of read() similar to your previous finding in
https://github.com/syslog-ng/syslog-ng/commit/431bc3583c25842c6c4155230dfeb52787191926

tests/python_functional/src/syslog_ng_config/statements/destinations/file_destination.py
mitzkia: Thanks, for the very good questions. 
Not yet think about this, but I can imagine two options how we can handle multi-line case for file drivers:
1, strict way: in this case we will implement such an API which matches with syslog-ng settings: prefix, garbage and so on. This could work, but for the first thought it can be overhead.
2, loose way: in this case we just read everything from the destination side into one string object and then we can assert for regexp matching in it.

In case of framed messages the low level IO class (which is not yet exist) should handle the working mode.

Please correct me if I misunderstood something.

MrAnno: Message reader was a good abstraction that can handle everything you mentioned in one place, in a configurable manner.
The implementation was unnecessarily complex, but the interface could have been really good if it had been used in the right place.

Anyway, you can add this or something similar back when you need it while implementing multi-line/framed message support. (The regexp trickery in each multiline test is not a clean way of writing tests in my opinion.)

mitzkia: Yes, I would like to add this part later, when it is actual by testcases. I will fix soon your other review comment.

kira-syslogng:
Build SUCCESS

MrAnno (approved):

kira-syslogng:
Build SUCCESS

mitzkia:
@MrAnno Thanks for the approve. I needed to update the branch with one last commit, where I updated the file list for the tarball.
https://github.com/syslog-ng/syslog-ng/pull/3511/commits/38b350df83f16373b5db05f2006a0f7df6e28d7a

MrAnno (approved):

gaborznagy (approved):
I really like this PR, thanks!
The `FileIO` and `File` usage cleanup is great (that `File` should be used by tools e.g. `SyslogNgConfig` and ConsoleReader)!

In the near-future, we should consider @MrAnno's comment about abstraction:
https://github.com/syslog-ng/syslog-ng/pull/3511#discussion_r595102197
He is right: the current interface of `File` and `FileDestination` is heavily based on lines.

PR #3513, opened 2020-12-10
affile/wildcard: fix crash when reader replaced

Kokan:
reproduction of the crash:

config:
```
@version: 7.0

log {

source {
  wildcard_file(
    base-dir("/tmp/input/")
    filename-pattern("*")
  );
};

destination {
  file("/tmp/output.txt");
};

flags(flow-control);

};
```
create a non empty file in /tmp/input/

1. start syslog-ng with the above configuration
2. start an infinite loop with renaming the file created in /tmp/input/
   (for example while true ; mv /tmp/input/2 /tmp/input1 && mv
/tmp/input/1 /tmp/input2 ; end)
The crash might occure right away, or after a few minutes.

Signed-off-by: Kokan <kokaipeter@gmail.com>

### DIFF
--- a/modules/affile/wildcard-source.c
+++ b/modules/affile/wildcard-source.c
@@ -77,11 +77,20 @@ _remove_file_reader(FileReader *reader, gpointer user_data)
     }
   log_pipe_unref(&reader->super);
 
-  gchar *full_path = pending_file_list_pop(self->waiting_list);
-  if (full_path)
+  for (GList *pending_iterator = pending_file_list_begin(self->waiting_list);
+       pending_iterator != pending_file_list_end(self->waiting_list);
+       pending_iterator = pending_file_list_next(pending_iterator))
     {
+      gchar *full_path = pending_iterator->data;
+      if (g_hash_table_lookup_extended(self->file_readers, full_path, NULL, NULL))
+        {
+          continue;
+        }
+      pending_file_list_steal(self->waiting_list, pending_iterator);
       _create_file_reader(self, full_path);
+      g_list_free_1(pending_iterator);
       g_free(full_path);
+      break;
     }
 }
 
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -81,8 +81,7 @@ log_reader_set_options(LogReader *s, LogPipe *control, LogReaderOptions *options
   log_source_set_ack_tracker_factory(&self->super, ack_tracker_factory_ref(factory));
 
   log_pipe_unref(self->control);
-  log_pipe_ref(control);
-  self->control = control;
+  self->control = log_pipe_ref(control);
 
   self->options = options;
   log_proto_server_set_options(self->proto, &self->options->proto_options.super);
--- a/modules/affile/file-list.c
+++ b/modules/affile/file-list.c
@@ -56,11 +56,14 @@ void pending_file_list_add(PendingFileList *self, const gchar *value)
 
 gchar *pending_file_list_pop(PendingFileList *self)
 {
-  gchar *data = g_queue_pop_head(self->priority_queue);
-  if (data)
-    {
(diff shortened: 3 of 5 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

alltilla:
Can you add a news entry please? Thanks!

I am checking the macOS problem in the meantime. It is not caused by this PR.

modules/affile/wildcard-source.c
```diff
@@ -77,11 +77,17 @@ _remove_file_reader(FileReader *reader, gpointer user_data)
...
 
-  gchar *full_path = pending_file_list_pop(self->waiting_list);
-  if (full_path)
+  gchar *full_path = NULL;
+  while ((full_path = pending_file_list_pop(self->waiting_list)))
     {
+      if (g_hash_table_lookup_extended(self->file_readers, full_path, NULL, NULL))
+        {
+          g_free(full_path);
+          continue;
```
MrAnno, line 86: Is it OK to remove (pop) the path from the waiting list when we skip creating a reader from it?
I mean the old and new file with the same name can be different.
We might read an old deleted/renamed file with its FD kept open, but we do want to schedule the new one after we reach EOF in the old one.

kira-syslogng:
Build SUCCESS

modules/affile/wildcard-source.c
Kokan: you are right, i've added a commit to it, please check.

modules/affile/wildcard-source.c
```diff
@@ -77,11 +77,24 @@ _remove_file_reader(FileReader *reader, gpointer user_data)
     }
   log_pipe_unref(&reader->super);
 
-  gchar *full_path = pending_file_list_pop(self->waiting_list);
-  if (full_path)
+  gchar *full_path = NULL;
+  GQueue *q = g_queue_new();
```
MrAnno, line 81: This has to be freed with `g_queue_free`.

modules/affile/wildcard-source.c
```diff
@@ -77,11 +77,24 @@ _remove_file_reader(FileReader *reader, gpointer user_data)
...
+  while ((full_path = pending_file_list_pop(self->waiting_list)))
     {
+      if (g_hash_table_lookup_extended(self->file_readers, full_path, NULL, NULL))
+        {
+          g_queue_push_tail(q, full_path);
+          g_free(full_path);
```
MrAnno, line 87: GQueue does not copy elements, the queue will have dangling pointers.

modules/affile/wildcard-source.c
MrAnno: Just a thought (completely optional):
I was thinking of a cleaner implementation, based on `g_queue_peek_nth()` or `g_queue_peek_head_link()`, or with a simple GList reimplementation, but `pending_file_list` would require an iterator interface with cancellation.

For example:
```
full_path = pending_file_find_and_steal(self->waiting_list, FindFunc);
```
or
```
while ((full_path_glist = pending_file_list_next(self->waiting_list)))
{
  if (...)
    pending_file_list_unlink(full_path_glist)
}
```

modules/affile/wildcard-source.c
MrAnno: Thanks!

kira-syslogng:
Build SUCCESS

szemere (approved):
I think all of @MrAnno's comments were addressed. Approve from my side. (macOS@CI is still pending)

MrAnno (approved):
Nice. Thank you very much for adding the iterator. 👍🏻 

(The failing macOS job is fixed on master, I think we can ignore that in this PR.)

PR #3516, opened 2020-12-11
dbld: remove obsolete package dh-systemd

gaborznagy:
It's removed from debian testing due to being a transitional package. [1]
Debhelper package is already used, so ne need to add a new package to dbld.

It can be removed from some ubuntu images as well. [2]

[1] https://packages.debian.org/search?searchon=names&keywords=dh-systemd
[2] https://packages.ubuntu.com/search?suite=all&searchon=names&keywords=dh-systemd


gaborznagy:
This was found in #3490

MrAnno (approved):

kira-syslogng:
Build SUCCESS

bazsi (approved):

PR #3519, opened 2020-12-17
Fixing formatting SDATA with enterpriseID

Kokan:
There are two formats for SD-ID names:
* Names that do not contain an at-sign ("@", ABNF %d64) are reserved
* Anyone can define additional SD-IDs using names in the format
  name@<private enterprise number>, e.g., "ourSDID@32473".

SDATA is stored as nvpair as string, and no structural information
saved. The way it currently works is looking up the last dot(.), and
claim everything before that as SD-ID. Resulting in invalid sdata, when
the PARAM-NAME has dot(.) in it, which is valid.

The following configuration results in invalide sdata on the
destination:
```
@Version: 3.30

log {
 source      { example-msg-generator(num(1)); };
 rewrite     { set("112" value(".SDATA.test@1.2.3.log.fac")); };
 destination { syslog("127.0.0.1" port(1111)); };
};
```

The fix depends on the fact that SD-ID has either contains at(@) or not,
when it contains at(@), it must followed by enterpriseId(number)
optionally with sub-id (number also).
As such the first non dot(.) or non digit shall be part of PARAM-NAME.

TODO: include news file.


kira-syslogng:
Build FAILURE

lib/logmsg/logmsg.c
```diff
@@ -923,6 +923,21 @@ log_msg_sdata_append_escaped(GString *result, const gchar *sstr, gssize len)
...
+first_non_dot_or_digit_char(const gchar *str, gint str_len)
+{
+  const gchar *pos = str;
+
+  while (str_len > 0 && pos && (*pos == '.' || isdigit(*pos)))
+    {
+      ++pos;
+      --str_len;
+    }
+  if (str_len == 0) return NULL;
```
bazsi, line 936: using strspn(str, "0123456789.") might be faster (and is a standard part of libc, so no need to open code it.

this is an optional remark.

bazsi (dismissed):
Thank you for breaking up the changes into such small commits, it made reviewing this - otherwise pretty involved - change pretty easy.

Approved, with a small comment. You also have a few typos in the commit messages, that could be fixed, if there's another iteration coming up.

lib/logmsg/logmsg.c
Kokan: Sure, thanks for the note. The PR needs update either way, so I'll do it.

kira-syslogng:
Build SUCCESS

gaborznagy:
Sorry, CI devshell/general jobs are broken. The fix was just submitted: https://github.com/syslog-ng/syslog-ng/pull/3520

lib/logmsg/logmsg.c
```diff
@@ -923,6 +923,41 @@ log_msg_sdata_append_escaped(GString *result, const gchar *sstr, gssize len)
...
+        }
+
+      return sd_id_sep;
+    }
+
+  /* The SD-ID has two format, one with enterpriseId:
+   * SD-ID : <key>@<enterprise-id>[.<sub-id>]*.<sd-name> = <sd-value>
+   * Both <enterprise-id> and <sub-id> must be a number.
+   */
+  const gchar *at = memchr(sdata_name + 7, '@', sdata_name_len - 7);
```
szemere, line 950: **minor** comment: there is a constant for ".SDATA." length: https://github.com/syslog-ng/syslog-ng/blob/2d594f664a5d8c01081c96c966ac3d05ef8675d4/lib/logmsg/logmsg.c#L209

szemere (dismissed):

lib/logmsg/logmsg.c
Kokan: fixed.

kira-syslogng:
Build SUCCESS

szemere (dismissed):

MrAnno (commented):
I also tried a few corner cases, they all seem to produce a reasonable output, except the first one, where a character is missing, but that's an incorrect input anyway:

```
Test(sdata_format, test_corner)
{
  log_msg_set_value_by_name(msg, ".SDATA.a@123", "33", -1); // [a@12] 3 is missing

  log_msg_set_value_by_name(msg, ".SDATA.a@awoeifjwoijf", "33", -1);
  log_msg_set_value_by_name(msg, ".SDATA.a@a", "33", -1);
  log_msg_set_value_by_name(msg, ".SDATA.b@", "33", -1);
  log_msg_set_value_by_name(msg, ".SDATA.@", "33", -1);

  log_msg_format_sdata(msg, result, 0);

  cr_assert_str_eq(result->str, "[a@12][a awoeifjwoijf=\"33\" a=\"33\"][b][]");
}
```

MrAnno:
In general, a dot is expected after the enterprise ID. If that's missing, the last character will be discarded:
```
log_msg_set_value_by_name(msg, ".SDATA.a@123abc", "33", -1); // [a@12 abc="33"] last character is missing
```

kira-syslogng:
Build SUCCESS

MrAnno (approved):

lib/logmsg/logmsg.c
```diff
@@ -968,30 +998,24 @@ log_msg_append_format_sdata(const LogMessage *self, GString *result,  guint32 se
       sdata_elem = sdata_name + 7;
       sd_id_len = (handle_flags >> 8);
 
-      if (sd_id_len)
+      dot = find_sd_id_in_key(sd_id_len, sdata_elem, sdata_name, sdata_name_len);
```
bazsi, line 1001: I just checked, and nv_registry handle_flags would _NEVER_ contain the length of the SD ID, so sd_id_len is _always_ zero.

As it seems I've removed its setting in e85b77daf7adefc20186784f538ac642f414fb6b. Originally the intention has been to store the length of the SD ID, so that our joining of the name-value pair wouldn't cause a change in the SDATA format.

I am not sure if the removal was the right move, but definitely, we would also need to remove the code conditional on sd_id_len != 0. 

That trick might be used to "cache" our decision on the separation of the SD-ID and the SD-PARAM, so this code could be faster. But instead of doing this in syslogformat (where I've removed from), it should be added to the SDATA slow path, e.g. the point where an SDATA nvpair is registered and get applied to any kind of SDATA nvpair, be them parsed from an actual RFC5424 message, or set in a different place (E.g. json-parser()).

PR #3520, opened 2020-12-18
Fix devshell CI

gaborznagy:
After merging #3476 some devshell-based jobs failed (they were badly tested, and since github actions uses a previously built balabit/syslog-ng-devshell image CI did not detect the issue either).

As it's discussed in #3476 java modules are not tested in devshell, they are tested with the [binary packages](https://github.com/syslog-ng/syslog-ng/actions?query=workflow:%22Binary+packages%22) CI job.

This PR also fixes a python 3.9 deprecation warning about `base64.decodestring().

### DIFF
--- a/tests/functional/func_test.py
+++ b/tests/functional/func_test.py
@@ -45,15 +45,8 @@ def init_env():
         if e.errno != errno.EEXIST:
             raise
 
-
-def seed_rnd():
-    # Some platforms lack kernel supported random numbers, on those we have to explicitly seed the RNG.
-    # using a fixed RND input does not matter as we are not protecting real data in this test program.
-    try:
-        import base64
-        import socket
-        import ssl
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

Kokan (approved):

tests/functional/func_test.py
```diff
@@ -53,7 +53,7 @@ def seed_rnd():
         import base64
         import socket
         import ssl
-        rnd = base64.decodestring("""k/zFvqjGWdhStmhfOeTNtTs89P8soknF1J9kSQrz8hKdrjIutqTXMfIqCNUb7DXrMykMW+wKd1Pg
+        rnd = base64.decodebytes("""k/zFvqjGWdhStmhfOeTNtTs89P8soknF1J9kSQrz8hKdrjIutqTXMfIqCNUb7DXrMykMW+wKd1Pg
```
MrAnno: `decodebytes` is not available in Python 2, but we have CI jobs for Python 2 builds/functests.

MrAnno (approved):

bazsi:
There's a shld target that can rebuild an image of necessary, so we won't run into these stale image problems

PR #3523, opened 2020-12-26
cmake: small build related fixes

furiel:
I ran into a small trouble while compiling syslog-ng using cmake on osx. With these patches applied, I could start syslog-ng with an example-msg-generator and stdout destination.


kira-syslogng:
Build SUCCESS

bazsi:
The amqp build fails in devshell when compiling with cake, because it doesn't test for a deprecated function call.

The same works in automake.

bazsi (approved):
This looks good to me, except for the broken build in devshell

furiel:
@bazsi thank for the analysis. I will look into the devshell failure.

kira-syslogng:
Build SUCCESS

MrAnno (approved):
Yay!

CMakeLists.txt
```diff
@@ -79,6 +79,8 @@ include(CheckCSourceCompiles)
 
 add_definitions(-D_GNU_SOURCE=1)
 add_definitions(-D_LARGEFILE64_SOURCE=1)
+add_definitions(-D__APPLE_USE_RFC_3542)
```
MrAnno, line 82: We might add a conditional around this macro next time, so it won't appear on irrelevant platforms.

gaborznagy (approved):

PR #3526, opened 2021-01-06
host-resolve: invalidate ivykis "now" after blocking calls to DNS

bazsi:

### DIFF
--- a/modules/afsocket/afinet-dest.c
+++ b/modules/afsocket/afinet-dest.c
@@ -362,6 +362,7 @@ afinet_dd_setup_addresses(AFSocketDestDriver *s)
   if (!_setup_dest_addr(self))
     return FALSE;
 
+  iv_invalidate_now();
   return TRUE;
 }
 
--- a/lib/host-resolve.c
+++ b/lib/host-resolve.c
@@ -28,6 +28,8 @@
 #include "tls-support.h"
 #include "compat/socket.h"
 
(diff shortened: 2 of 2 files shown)


kira-syslogng:
Build SUCCESS

modules/afsocket/afinet-dest.c
```diff
@@ -362,6 +362,7 @@ afinet_dd_setup_addresses(AFSocketDestDriver *s)
   if (!_setup_dest_addr(self))
     return FALSE;
 
+  iv_invalidate_now();
```
MrAnno, line 365: I think this is not necessary, as `_setup_dest_addr()` invalidates `now` through `resolve_hostname_to_sockaddr()`.

MrAnno (approved):
I pushed a small change onto this branch (`#include <iv.h>`) to make everything green.
This fixes #3514.

kira-syslogng:
Build SUCCESS

gaborznagy (approved):
I agree with @MrAnno's comment that iv_invalide_now() is not needed in afinet-dest.c, otherwise looks good to me.

@bazsi please consider that comment, but it can be merged if you like to keep it this way.

MrAnno:
My comment is not really important, the extra `iv_invalidate_now()` in afinet-dest won't have side-effects.

I'll send a newsfile entry.

PR #3528, opened 2021-01-08
 afinet: don't disable checksum calculation with spoof-source'd UDP6

### LINKED ISSUE #3525: Invalid checksum of 0 in UDP6 packet when using spoof_source(yes)
Hello,

spoof_source(yes) works fine in IPv4. In IPv6, it does correctly spoof the source, but creates an invalid checksum 0 for the packet, and so it ends up being dropped by the receiver as invalid. If spoof_source(no) is used instead, then the IPv6 checksum is calculated correctly, and so the packet is correctly processed by the receiver.

I suspect the issue might be that this "temporary workaround" is no longer working:

https://github.com/syslog-ng/syslog-ng/blob/1a7cc761b6c61c4abd2bcfaea4c6af3be6d862cc/modules/afsocket/afinet-dest.c#L552

Very likely that workaround is not needed anymore because it was added 8 years ago.

alltilla:
The problem, we worked around, has been fixed in libnet 1.1.6, which is out in every major distro. The workaround can be safely removed.

Fixes #3525.
~NOTE: Internal testing is still in progress!~

### DIFF
--- a/modules/afsocket/afinet-dest.c
+++ b/modules/afsocket/afinet-dest.c
@@ -429,7 +429,7 @@ afinet_dd_init(LogPipe *s)
           if (!self->lnet_ctx)
             {
               msg_error("Error initializing raw socket, spoof-source support disabled",
-                        evt_tag_str("error", NULL));
+                        evt_tag_str("error", error));
             }
         }
     }
--- a/modules/afsocket/afinet-dest.c
+++ b/modules/afsocket/afinet-dest.c
@@ -549,11 +549,6 @@ afinet_dd_construct_ipv6_packet(AFInetDestDriver *self, LogMessage *msg, GString
   if (udp == -1)
     return FALSE;
 
-  /* There seems to be a bug in libnet 1.1.2 that is triggered when
-   * checksumming UDP6 packets. This is a workaround below. */
-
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

MrAnno (dismissed):

Kokan:
Adding min version with 1.1.6 might help people not bump into having older version and silently causing issues.

alltilla:
@Kokan Totally valid thing to do. Unfortunately it is pretty hard to do in case of libnet.
 1. `libnet` does net have a `.pc` file, so `pkg-config` cannot help.
 2. `libnet` itself only has full string version definition in its header (`#define LIBNET_VERSION "1.1.6"`), so comparison is a pain...

The only solution I see is to do the painful comparison using `strtok()` and `atoi()`, not sure if it is worth it.

Kokan:
Thanks for checking out possibilities. I hoped that at least `libnet-config` provides version number, well as I see upstream at least going to provide .`pc` file, so in future we may have a version number...
At least please include the min version change in the news entry, so in the release news file it can be highlighted.

Otherwise lgtm from me.

kira-syslogng:
Build SUCCESS

Kokan (approved):

bazsi (approved):

PR #3529, opened 2021-01-08
dbld: improve cache-image rule

gaborznagy:
Dependencies of dbld docker images come from:
  - manifest files, e.g. dbld/packages.manifest
  - packaging files (see dbld/builddeps install_debian_build_deps() and install_rpm_build_deps())

We should check if a dependency is changed in the packaging files.
However rpm spec file and debian control file contain other packaging instructions besides build dependencies.
This means that e.g. when we are releasing, the PR that is opened by "version bump" github actions will re-build the docker images just because of the version bump itself.

@bazsi What do you think of this change?

### DIFF
--- a/dbld/rules
+++ b/dbld/rules
@@ -222,6 +222,7 @@ pull-image-%:
 cache-image-%:
 	@IMAGE=balabit/syslog-ng-$*:latest;										\
 	IMAGE_ID=$$(docker images -q $$IMAGE | head -1);								\
+	WATCHED_FILES="dbld packaging/rhel/syslog-ng.spec packaging/debian/control";					\
 	if [ "$$IMAGE_ID" = "" ]; then 											\
 		$(DBLD_RULES) pull-image-$*;										\
 	fi;														\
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
I generally like the change and I am not very concerned about rebuilding docker images in a version bump, that's a rare PR anyway.

If we want to solve that, we might extract the logic to check docker related changes into a script (e.g. dbld/validate-docker-image) and in that script:
* check if dbld changed at all
* validate that the changes in debian/control and syslog-ng.spec affect Build-Depends or BuildRequires lines (e.g. grep into the diff output and check if these lines were changed).

And only rebuild the image if validate-docker-image indicates this. The validate script would probably need the git history and the commit ID of the docker image and the commit ID of the current HEAD.

gaborznagy:
> I generally like the change and I am not very concerned about rebuilding docker images in a version bump, that's a rare PR anyway.
> 
> If we want to solve that, we might extract the logic to check docker related changes into a script (e.g. dbld/validate-docker-image) and in that script:
> 
> * check if dbld changed at all
> * validate that the changes in debian/control and syslog-ng.spec affect Build-Depends or BuildRequires lines (e.g. grep into the diff output and check if these lines were changed).
> 
> And only rebuild the image if validate-docker-image indicates this. The validate script would probably need the git history and the commit ID of the docker image and the commit ID of the current HEAD.

Thanks for the feedback @bazsi. I agree with your more sophisticated idea for this, nevertheless I would merge this in it's current form if the team agrees.

bazsi (approved):

Kokan (approved):

PR #3532, opened 2021-01-11
host-resolve: make iv_now invalidation conditional

MrAnno:
#3526 introduced coupling between the host-resolve unit and ivykis.

`iv_now` invalidation is needed because DNS lookup can be a long-running blocking call, and the ivykis time cache has to be invalidated "after any operation that takes a significant amount of wall clock time".
(Similar coupling exists in several syslog-ng units, ivykis operates on global and thread-local states.)

This commit makes `iv_now` invalidation conditional. This avoids crashes when host-resolve functions are called from non-ivykis threads. (An internal module encountered such a scenario.)

No news file needed.

### DIFF
--- a/lib/host-resolve.c
+++ b/lib/host-resolve.c
@@ -228,6 +228,13 @@ resolve_hostname_to_sockaddr_using_gethostbyname(GSockAddr **addr, gint family, 
 }
 #endif
 
+static inline void
+_attempt_to_invalidate_iv_now(void)
+{
+  if (iv_inited())
+    iv_invalidate_now();
+}
+
 gboolean
 resolve_hostname_to_sockaddr(GSockAddr **addr, gint family, const gchar *name)
 {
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
I didn't like introducing that coupling, however I thought the entire code base depends on ivykis, so it should be fine.

The conditional could be extracted into a function though or even pushed to upstream ivykis as a separate function.

But I am fine with this as is.

bazsi (dismissed):

MrAnno:
Good idea, I'll try to open a PR for ivykis.  Until then, I'm going to just extract this within syslog-ng.

kira-syslogng:
Build SUCCESS

gaborznagy (approved):

Kokan (approved):

PR #3536, opened 2021-01-15
scl/fortigate: add parser for fortigate format

bazsi:
This depends on #3538 that allows me to parse a stray <PRI> value without going into further details in the message. 

Otherwise it contains fortigate-parser() and an application adapter for app-parser() to parse the same automatically.

PR #3538, opened 2021-01-16
add "no-header" flag to syslog-parser()

bazsi:
This patch was originally part of #3536 but is split off to a separate PR, as it is a feature of its own right. #3536 would depend on this.

The new header can be used either at a source:

```
source s_udp { 
    udp(flags(no-header)); 
};
```

Or with syslog-parser()
```
parser p_syslog {
    syslog-parser(flags(no-header));
};
```

The no-header flag means that the syslog header is not present (or does not adhere to the conventions/RFCs), so the entire message (except from the PRI field) is shoved into $MSG.

It is somewhat similar to flags(no-parse), except that no-parse would skip the PRI field too.

### DIFF
--- a/modules/syslogformat/syslog-format.c
+++ b/modules/syslogformat/syslog-format.c
@@ -722,33 +722,12 @@ error:
   return ret;
 }
 
-
-/**
- * log_msg_parse_legacy:
- * @self: LogMessage instance to store parsed information into
- * @data: message
- * @length: length of the message pointed to by @data
- * @flags: value affecting how the message is parsed (bits from LP_*)
- *
- * Parse an RFC3164 formatted log message and store the parsed information
- * in @self. Parsing is affected by the bits set @flags argument.
- **/
 static gboolean
-log_msg_parse_legacy(const MsgFormatOptions *parse_options,
-                     const guchar *data, gint length,
-                     LogMessage *self, gint *position)
+log_msg_parse_legacy_header(LogMessage *self, const guchar **data, gint *length, const MsgFormatOptions *parse_options)
 {
-  const guchar *src;
-  gint left;
+  const guchar *src = *data;
+  gint left = *length;
   GTimeVal now;
-
-  src = (const guchar *) data;
-  left = length;
-
-  if (!log_msg_parse_pri(self, &src, &left, parse_options->flags, parse_options->default_pri))
-    {
-      goto error;
-    }
 
   log_msg_parse_cisco_sequence_id(self, &src, &left);
   log_msg_parse_skip_chars(self, &src, &left, " ", -1);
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build FAILURE

MrAnno:

modules/syslogformat/syslog-format.c
```diff
@@ -833,7 +833,8 @@ log_msg_parse_legacy(const MsgFormatOptions *parse_options,
       goto error;
     }
 
-  log_msg_parse_legacy_header(self, &src, &left, parse_options);
+  if ((parse_options->flags & LP_NO_HEADER) == 0)
```
MrAnno, line 836: I think it's worth mentioning that `no-header` works only in `RFC3164`-mode, which means
- when `network(), tcp(), udp()` are used
- when `syslog()` or `flags(syslog-protocol)` is used (`RFC5424`), but the incoming message does not contain correct `pri` and `version` fields, so we fall back to `RFC3164`.

In other words, `flags(no-header)` does not work with properly formatted `RFC5424` messages.
@bazsi Should we address that? I can not really think of a good use-case, but the name `no-header` suggests that it should work in both cases.

MrAnno (approved):

kira-syslogng:
Build SUCCESS

bazsi:
On Sat, Jan 16, 2021, 14:34 László Várady <notifications@github.com> wrote:

> *@MrAnno* approved this pull request.
>
> In modules/syslogformat/syslog-format.c
> <https://github.com/syslog-ng/syslog-ng/pull/3538#discussion_r558884197>:
>
> > @@ -833,7 +833,8 @@ log_msg_parse_legacy(const MsgFormatOptions *parse_options,
>        goto error;
>      }
>
> -  log_msg_parse_legacy_header(self, &src, &left, parse_options);
> +  if ((parse_options->flags & LP_NO_HEADER) == 0)
>
> I think it's worth mentioning that no-header works only in RFC3164-mode,
> which means
>
>    - when network(), tcp(), udp() are used
>    - when syslog() or flags(syslog-protocol) is used (RFC5424), but the
>    incoming message does not contain correct pri and version fields, so
>    we fall back to RFC3164.
>
> In other words, flags(no-header) does not work with properly formatted
> RFC5424 messages.
> @bazsi <https://github.com/bazsi> Should we address that? I can not
> really think of a good use-case, but the name no-header suggests that it
> should work in both cases.
>

I can't think of a use case either, rfc5424 should be parsed fully and I
wouldn't want to water it up.

Having only the pri field and the rest something random, is pretty common,
and so far all we could do was to do flags(no-parse) and then manually
parse the pri field. Which didn't happen in cisco-parser() either.

Also, back in the day when rfc5424 was specified the general conclusion was
that the single most kept part of rfc3164 was the pri field, and that's why
I have added this into the syslog-parser() and not have come up with a new
one.

I was thinking of naming before coming up with this one, but if you look at
the rest of the flags, they are quite similar (no-hostname for instance).

The big structural issue with these flags is that they are part of the
MsgFormat interface whereas we do parse more than just syslog there. But
this applies all the other flags too, and I don't really know how we could
resolve that.

lbudai (approved):

PR #3546, opened 2021-01-21
add set-pri() as a rewrite operation

### LINKED ISSUE #3541: Parse only PRI
## Description of the problem
In certain cases where the source is more invalid than valid and the syslog parser would rightfully mangle the event I don't call the syslog-ng parser instead I parse using patterndb or other features of syslog-ng.

For example
<134>1 1111611101486.978292274 CLL_FWL flows src=xxx.xxx.xxx.xxx dst=xxx.xxx.xxx.xxx protocol=tcp sport=xxxx dport=xxxx pattern: 1 all

## Proposed solution
Provide a parser or re-write function that would take "134" as a parameter set PRI, SEVERITY, and FACILITY

## Alternatives
Use Python to calculate facility and severity then call set-severity and set-facility re-write but this is more steps and python is slow for this use

bazsi:
This branch adds a new rewrite operation, set-pri() that allows one to set the pri value of the message (e.g. facility * 8 + severity) based on a template string.

The template string needs to expand to a number between in the interval [0, 1023] inclusive, that permits up to 127 facilities.

The format accepted by set-pri() is a simple number, e.g. if you extract this value from a syslog header, the brackets "<" and ">" needs to be removed first.

The format may be extended in the future if use-cases arise, I was thinking of allowing the brackets to be present, or the use of the dot separated, string based format, such as "mail.info".

Fixes #3541

### DIFF
--- a/lib/rewrite/rewrite-set-pri.c
+++ b/lib/rewrite/rewrite-set-pri.c
@@ -1,0 +1,115 @@
+/*
+ * Copyright (c) 2021 Balazs Scheidler <bazsi77@gmail.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "rewrite-set-pri.h"
+#include "template/templates.h"
+#include "syslog-names.h"
+#include "scratch-buffers.h"
+
+#include <stdlib.h>
+
+typedef struct _LogRewriteSetPri LogRewriteSetPri;
+
+struct _LogRewriteSetPri
+{
+  LogRewrite super;
+  LogTemplate *pri;
+};
+
+static gint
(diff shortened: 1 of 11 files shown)


mbonsack:
Allowing the brackets to be present would be helpful, but not critical.  Thanks!

kira-syslogng:
Build SUCCESS

lib/rewrite/rewrite-set-pri.c
```diff
@@ -0,0 +1,115 @@
...
+  LogRewrite super;
+  LogTemplate *pri;
+};
+
+static gint
+_convert_pri(GString *pri_text)
+{
+  char *endptr;
+  long int pri = strtol(pri_text->str, &endptr, 10);
+  if (0 == endptr)
```
alltilla, line 44: Should be `NULL`, right?

lib/rewrite/tests/test_set_pri.c
```diff
@@ -0,0 +1,138 @@
...
+Test(set_pri, numeric)
+{
+  _perform_set_pri(_create_template("7"), msg);
+  cr_assert(_msg_pri_equals(msg, LOG_KERN | LOG_DEBUG));
+
+  _perform_set_pri(_create_template("189"), msg);
+  cr_assert(_msg_pri_equals(msg, LOG_LOCAL7 | LOG_NOTICE));
+
+  _perform_set_pri(_create_template("$PRI"), msg);
+  cr_assert(_msg_pri_equals(msg, 189));
```
alltilla, line 84: Removing `_perform_set_pri(_create_template("$PRI"), msg);` still produces green tests, so this test could hide problems, where the template resolution does not affect the previous `PRI` value.

I would rather test the template resolution with a different macro and different value, like:
```
  log_msg_set_value_by_name(msg, "MSG", "137", -1);
  _perform_set_pri(_create_template("$MSG"), msg);
  cr_assert(_msg_pri_equals(msg, 137));
```

alltilla (commented):
I have found these 2 minor problems, otherwise it looks good to me!

lib/rewrite/rewrite-set-pri.c
bazsi: fixed.

lib/rewrite/tests/test_set_pri.c
bazsi: fixed this as well, and added a few more testcases to cover more invalid cases.

kira-syslogng:
Build SUCCESS

alltilla (approved):

MrAnno (approved):

PR #3550, opened 2021-01-26
diskq: create the path specified by dir() if not exists

Kokan:
No configuration change needed, or option to disable. The directory is created with the same permission as the syslog-ng is running with.

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -777,6 +777,14 @@ _update_header_with_default_values(QDisk *self)
   self->hdr->backlog_len = 0;
 }
 
+static gboolean
+_create_path(const gchar *filename)
+{
+  FilePermOptions fpermoptions;
+  file_perm_options_defaults(&fpermoptions);
+  return file_perm_options_create_containing_directory(&fpermoptions, filename);
+}
+
 gboolean
 qdisk_start(QDisk *self, const gchar *filename, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
 {
@@ -810,6 +818,13 @@ qdisk_start(QDisk *self, const gchar *filename, GQueue *qout, GQueue *qbacklog, 
         {
           new_file = TRUE;
         }
+    }
+  if (!_create_path(filename))
+    {
+      msg_error("Error creating dir for disk-queue file",
+                evt_tag_str("filename", filename),
+                evt_tag_error("error"));
+      return FALSE;
     }
 
   self->filename = g_strdup(filename);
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -816,6 +816,16 @@ qdisk_start(QDisk *self, const gchar *filename, GQueue *qout, GQueue *qbacklog,
   /* assumes self is zero initialized */
   openflags = self->options->read_only ? (O_RDONLY | O_LARGEFILE) : (O_RDWR | O_LARGEFILE | (new_file ? O_CREAT : 0));
 
+  FilePermOptions fpermoptions;
```
gaborznagy, line 819: I would move this code part before the `openflags = ...` line.

gaborznagy (approved):

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -777,6 +777,14 @@ _update_header_with_default_values(QDisk *self)
   self->hdr->backlog_len = 0;
 }
 
+static gboolean
+_create_path(const gchar *filename)
+{
+  FilePermOptions fpermoptions;
+  file_perm_options_defaults(&fpermoptions);
+  return file_perm_options_create_containing_directory(&fpermoptions, filename);
```
MrAnno, line 785: We have global `dir-owner()`, `dir-group()`, and `dir-perm()` options.

Shouldn't we inherit them (in case they are set) with
`file_perm_options_inherit_from(&fpermoptions, &cfg->file_perm_options)`?

MrAnno: `create-dirs()` is not taken into account either, If I understand it correctly, we just want disk-buffer directory creation to work without complicating things with these options.

MrAnno (approved):

modules/diskq/qdisk.c
Kokan: > `create-dirs`

Creating the path if not exists is something that make sense almost always (maybe there are some strange use-cases where it is not), and as such doing it unconditionally is fine.

> We have global dir-owner(), dir-group(), and dir-perm() options.

I was considering adding support for owner/group/perm. Just inheriting those globals is no good, in that case we should provide a way to set it locally. Also the disk queue file itself should inherit it as well (as it does not currently).
At this point I was thinking that addition is nice, but should be a new patchset and not in strict scope. That can be done always later.

Kokan:
wip till internal testing is done.

Kokan:
Test passed, it is okay to merge.

PR #3553, opened 2021-01-31
Fix crash in dateparser %z parsing

bazsi:
@rfaircloth-splunk encountered crashes since 3.30 was released. After looking at the core it turned out that there's a quite severe
race that we triggered with the latest changes in strptime() %z format.

If someone is using "%z" in his config and multiple threads are used (and running due to load), we were crashing every few seconds because of a race on the "tzname" global variable.

I've initially included a number of not-closely related patches within this PR, but I've now split them off to a separate #3561. This now concentrates on the fix and only the fix.

### DIFF
--- a/lib/mainloop.c
+++ b/lib/mainloop.c
@@ -269,9 +269,18 @@ main_loop_reload_config_apply(gpointer user_data)
       is_reloading_scheduled = FALSE;
       return;
     }
+
   self->old_config->persist = persist_config_new();
   cfg_deinit(self->old_config);
   cfg_persist_config_move(self->old_config, self->new_config);
+
+  /* The threads have stopped, deinit methods were called, but
+   * self->current_configuration still points to the old config.  We either
+   * go to the new config is cfg_init() is successful (just below) or revert
+   * to the old one if it's not.
+   * */
+
+  app_config_stopped();
 
   self->last_config_reload_successful = cfg_init(self->new_config);
   if (!self->last_config_reload_successful)
@@ -609,6 +618,7 @@ main_loop_read_and_init_config(MainLoop *self)
       return 0;
     }
 
+  app_config_stopped();
   if (!main_loop_initialize_state(self->current_configuration, resolvedConfigurablePaths.persist_file))
     {
       return 2;
--- a/lib/timeutils/wallclocktime.c
+++ b/lib/timeutils/wallclocktime.c
@@ -276,6 +276,7 @@ wall_clock_time_strptime(WallClockTime *wct, const gchar *format, const gchar *i
   int alt_format, i, split_year = 0, neg = 0, state = 0,
                      day_offset = -1, week_offset = 0, offs, mandatory;
   const char *new_fmt;
+  const char *const *system_tznames;
 
   bp = (const unsigned char *)input;
 
@@ -670,14 +671,13 @@ recurse:
                   bp = ep;
                   continue;
                 }
-              ep = find_string(bp, &i, (const char *const *)tzname, NULL, 2);
+              system_tznames = cached_get_system_tznames();
+              ep = find_string(bp, &i, system_tznames, NULL, 2);
               if (ep != NULL)
                 {
                   wct->tm.tm_isdst = i;
-#ifdef SYSLOG_NG_HAVE_TIMEZONE
-                  wct->wct_gmtoff = -(timezone);
-#endif
-                  wct->wct_zone = tzname[i];
+                  wct->wct_gmtoff = -cached_get_system_tzofs();
+                  wct->wct_zone = system_tznames[i];
                   bp = ep;
                   continue;
                 }
--- a/lib/timeutils/wallclocktime.c
+++ b/lib/timeutils/wallclocktime.c
@@ -676,7 +676,7 @@ recurse:
               if (ep != NULL)
                 {
                   wct->tm.tm_isdst = i;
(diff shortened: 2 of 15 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

MrAnno:
(spot-instance issue)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/mainloop.c
```diff
@@ -317,6 +317,7 @@ main_loop_reload_config_prepare(MainLoop *self, GError **error)
 
   service_management_publish_status("Reloading configuration");
 
+  app_config_stopped();
```
Kokan, line 320: At this point threads still can be running.

lib/timeutils/wallclocktime.c
```diff
@@ -670,13 +671,12 @@ wall_clock_time_strptime(WallClockTime *wct, const gchar *format, const gchar *i
...
+              system_tznames = cached_get_system_tznames();
+              ep = find_string(bp, &i, system_tznames, NULL, 2);
               if (ep != NULL)
                 {
                   wct->tm.tm_isdst = i;
-#ifdef SYSLOG_NG_HAVE_TIMEZONE
-                  wct->wct_gmtoff = -(timezone);
-#endif
+                  wct->wct_gmtoff = -cached_get_system_tzofs();
                   wct->wct_zone = tzname[i];
```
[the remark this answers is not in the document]
bazsi: Absolutely right, thanks for pointing this out.

lib/mainloop.c
bazsi: True as well, I remember trying to find the right spot, but at the end I may have inserted it incorrectly.

mitzkia:
I would like to create a testcase for this issue in Light. I already have several tries without any success.
I have tested with native date-paser() where the input was only %z equivalent values like: PST, PDT, GMT
I have also tested with apache-accesslog-parser() where the input was real apache logs (from the SCL file).
In every case I have used example-msg-generator-source() with very low freq option, and a high num option which generates messages very fast, of course the template was the input message.
In destination side I have used example_destination.
Inside the config I have put in one logpath: 10 example-msg-generator-source(), 10 date-parser() in other case 10 apache-accesslog-parser(), and 10 example_destinations.
And the whole config contains 10 logpaths from the above example.
After syslog-ng has started I can see 100% CPU load for a while (~15 sec.), but syslog-ng did not crash for me.

Could you please help me in a config creation or in message input when syslog-ng could crash?

Kokan:
@mitzkia Does your trial contained multiple `example-msg-generator-source() ` ?

mitzkia:
@kokan , Yes, as I mentioned my config contains 10 logpaths from the following example:
```
example-msg-generator1()
...
example-msg-generator10()
date_parser1()
...
date_parser10()
exmaple_destination1()
...
exmaple_destination10()
```

Kokan:
@mitzkia I tried with
```
@version: 3.30
@include "scl.conf"

log {
        source {
network( port(1111));
};
        parser { date-parser(format("%z")); };
};
```

and
```
/tmp/install/bin/loggen --inet --rate=1000000 --size=512 --interval=30 --active-connections=10 --syslog-proto --read-file=/tmp/loggen_input_file-glib-slow.txt --loop-reading --dont-parse
localhost 1111
```

The file containing:
```
CET
+05
M

```

That is an instant crash.

mitzkia:
Thanks, I will try your example

mitzkia:
@Kokan Thanks for the config example again. For now I can reproduce the crash locally without Light, but the crash happens usually for me after 30 sec. 
I will work on Light integration.

bazsi:
I am not sure it's worth to create a test case for this bug. I mean it is
important to confirm that we indeed fixed it, but it is pretty unlikely
that the very same race condition would resurface in the future.

mitzkia:
@bazsi Thanks for the comment. 
Finally I also realized, that this issue (which is a race condition) can not be reproduced easily and with exact steps (specially in my environment).
For me the crash happens in various time (with the same manual steps): immediately, after 5 sec, after 30 sec

So sadly, but I would skip to write testcase for this issue.

mitzkia:
If it helps I could verified the fix from testing side. 
Yesterday I wrote a flaky testcase for date-parser() which can reproduce the crash in Light (7 times out of 10). 
For now I run this testcase 25 times passed on this branch.

lib/timeutils/wallclocktime.c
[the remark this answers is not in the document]
bazsi: fixed.

lib/mainloop.c
bazsi: I am almost positive I identified a different spot where this needs to go, but somehow messed up at the end. Thanks a lot for finding this. I've now fixed it.

bazsi:
While fixing one of @Kokan's review notes, I've identified another bug in the strptime() implementation (namely the daylight saving hour for the system zones were ignored when calculating gmtoff). I've added a fix and unit test for that too (in a separate patch).

I think I've resolved all of the open issues now.

kira-syslogng:
Build SUCCESS

Kokan (approved):

alltilla (approved):

PR #3554, opened 2021-01-31
Dbld shell improvements

bazsi:
I finally had enough of my gdb not saving its history file, so here's a couple of patches that allows me to do that.

### DIFF
--- a/dbld/shell
+++ b/dbld/shell
@@ -14,4 +14,4 @@ if [ "$OS_PLATFORM" != "devshell" ]; then
 	echo -e "NOTE: the 'devshell' dbld image is more developer friendly, you might want to use that"
 fi
 cd /source
-HISTFILE=/build/.bash_history exec bash
+HOME=/build exec bash
--- a/dbld/shell
+++ b/dbld/shell
@@ -14,4 +14,4 @@ if [ "$OS_PLATFORM" != "devshell" ]; then
 	echo -e "NOTE: the 'devshell' dbld image is more developer friendly, you might want to use that"
 fi
 cd /source
-HOME=/build exec bash
+debian_chroot="dbld" HOME=/build exec bash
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build FAILURE

dbld/shell
```diff
@@ -14,4 +14,4 @@ if [ "$OS_PLATFORM" != "devshell" ]; then
 	echo -e "NOTE: the 'devshell' dbld image is more developer friendly, you might want to use that"
 fi
 cd /source
-HISTFILE=/build/.bash_history exec bash
\ No newline at end of file
+debian_chroot="dbld" HOME=/build exec bash
```
szemere, line 17: Minor note, otherwise LGTM!

It won't let me comment on line 13 ``˙if [ "$OS_PLATFORM" != "devshell" ]; then˙``

AFAIK we do not support other platforms as "shell". (Note: We DO support other platforms as build platforms.)
Based on this, either line 13 is not necessary, or debian_chroot is not "wide enough" for other platforms.

szemere (approved):

dbld/shell
bazsi: I regularly invoke shell on other platforms to try platform specific stuff. So we don't "support" that with all the usability stuff that we have in devshell or the ability to regenerate automake files, but it works fine nevertheless and I think its support at this level is important.

But the prompt is not that important, as I mostly use devshell.

mitzkia:
From GitHub Actions Centos-7 build is Fixed in #3556

mitzkia:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

gaborznagy (approved):
For me it's okay that the prompt change only works on debian/ubuntu platforms.

szemere (approved):

mitzkia (approved):

PR #3555, opened 2021-01-31
wallclocktime: fix non-mandatory parsing of timezone name

bazsi:
Since #3453 %Z was converted to allow a missing timezone specifier (porting over https://github.com/NetBSD/src/commit/1ab766532027329c2d69f6c89c0b9b5907553d15)

However, one case was missing: when a non-numeric timezone was specified but we couldn't find a match in nast/nadt/tzname. In that case we errored out, while NetBSD accepted it if the timezone was not mandatory.

I've added unit test coverage for both %z and %Z with this branch and identified another bug (but in this case it is an actual bug in NetBSD too), if the timezone was only using two digits (in the form "+05"), that should be interpreted in HOURs and not minutes. I've fixed that issue too.

@Kokan please let me know if you think this is not right. Thanks

### DIFF
--- a/lib/timeutils/tests/test_wallclocktime.c
+++ b/lib/timeutils/tests/test_wallclocktime.c
@@ -84,10 +84,24 @@ Test(wallclocktime, test_strptime_usec_parse_finds_character)
   cr_expect(wct.wct_usec == 0);
 }
 
-Test(wallclocktime, test_strptime_parses_rfc822_timezone)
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
removed "WIP", this should be good to go in.

kira-syslogng:
Build SUCCESS

Kokan (approved):

MrAnno (approved):

PR #3557, opened 2021-02-02
python: fix LogMessage.keys() listing non-existenting keys and duplicates

MrAnno:
Previously, the `keys()` method of `LogMessage` returned non-existent keys and duplicates that were part of the name-value registry.

The name-value registry contains all the possible names that are referenced in our codebase or in the syslog-ng configuration.

Reproducer code:
```
rewrite r_neverused { set("${neverused}" value("nonexistent")); };

log {
    source { stdin(); };
    parser { python(class(Test)); };
    destination{ file("/dev/stdout"); };
};

python {
class Test(object):
    def parse(self, log_message):
        for akey in log_message.keys():
            print(akey)
        return True
};
```

The above Python snippet prints "neverused" and "nonexistent" as if they were valid name-value pairs of the message, as well as duplicates, because the preceding `log_msg_values_foreach()` call added all keys including regex matches and builtin values.

This commit fixes the issue by filtering the nv-registry and allowing only macros.

Reported by @faxm0dem (thank you)

### DIFF
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -215,10 +215,8 @@ _collect_nvpair_names_from_logmsg(NVHandle handle, const gchar *name, const gcha
 }
 
 static gboolean
-_is_macro_name_visible_to_user(const gchar *name)
+_is_macro_name_visible_to_user(const gchar *name, NVHandle handle)
 {
-  NVHandle handle = log_msg_get_value_handle(name);
-
   return log_msg_is_handle_macro(handle) && !_is_key_blacklisted(name);
 }
 
@@ -226,9 +224,10 @@ static void
 _collect_macro_names(gpointer key, gpointer value, gpointer user_data)
 {
   const gchar *name = (const gchar *)key;
+  NVHandle handle = GPOINTER_TO_UINT(value);
   PyObject *list = (PyObject *)user_data;
 
-  if (_is_macro_name_visible_to_user(name))
+  if (_is_macro_name_visible_to_user(name, handle))
     {
       PyObject *py_name = PyBytes_FromString(name);
       PyList_Append(list, py_name);
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -215,7 +215,7 @@ _collect_nvpair_names_from_logmsg(NVHandle handle, const gchar *name, const gcha
 }
 
 static gboolean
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

modules/python/python-logmsg.c
```diff
@@ -214,26 +214,16 @@ _collect_nvpair_names_from_logmsg(NVHandle handle, const gchar *name, const gcha
...
+  NVHandle handle = log_msg_get_value_handle(name);
   gssize value_len;
 
   return (!_is_key_blacklisted(name) &&
-          (!_is_key_assigned_to_match_handle(name) ||
-           (_is_key_assigned_to_match_handle(name) &&
+          (log_msg_is_handle_macro(handle) ||
+           log_msg_is_handle_builtin(handle) ||
+           (log_msg_is_handle_match(handle) &&
             log_msg_get_value_by_name(logmsg, name, &value_len) != NULL
```
bazsi, line 227: We have an log_msg_is_set() function that allows empty values

MrAnno: Thank you. I'll use that too.

bazsi:
It might be enough on its own, well except macros.

kira-syslogng:
Build FAILURE

bazsi:
I think we should exclude unset name value pairs

bazsi:
Ahh, maybe the logmsg foreach function skips those already?

kira-syslogng:
Build FAILURE

MrAnno:
Yeah, I'll check what happens with unset values. I'm also having problems with `log_msg_set_match()`, still investigating that one.

MrAnno:
Unset and regexp-match keys are working as expected.

MrAnno:

kira-syslogng:
Build SUCCESS

alltilla (dismissed):
Thanks for this, nice catch and fix!

@mitzkia Can we write some Light tests? :) The `python` parser seems to have quite simple grammar, and we can write the `python` class to a separate file, and copy it next to the config.

mitzkia:
@alltilla I am agreed with you, of course we can support python parser in Light. There are already some parser examples in Light codebase.

MrAnno:
I've added a test case in the meantime.

kira-syslogng:
Build SUCCESS

alltilla (approved):

szemere (approved):

Issue #3548, opened 2021-01-24
syslog-ng 3.30 crashes on reload (triggered by the new threaded internal() source)


bazsi:
# syslog-ng
## Version of syslog-ng

3.30, 3.29 shouldn't be affected, but only due to the internal() reorganization that was merged for 3.30

## Platform

RHEL 8

# Issue
## Failure

syslog-ng crashes during reload while processing internal messages. It has to do with a race that the new threaded internal() source and the on_config_init method LogThreadedDestDriver introduces.

The problem is that the internal driver may start sending messages right after init() finishes (which is OK), but LogThreadedDestDriver has not finished initializing yet, as the DestWriter instances are initialized in destination threads.

What happens is that the internal source starts emitting messages to an uninitialized LogThreadedDestDriver state. The backtrace shows that we call log_threaded_dest_driver_queue() but the LogQueue instance within the writer is NULL at the point of the crash (but non-NULL in the core file, due to the nature of the race).

**Workaround:** disable internal messages or send them to stderr.

To solve this issue, we shouldn't defer any initialization to on_config_inited() that would prevent the the queue() method to fail. That is the only thing we can do in on_config_inited() is the startup of the threads, writer instances and its associated queues should be put into place during init.

Once we have both writers and their associated queues in place during init(), the delivery from the internal() source should be processed properly, the crash doesn't occur and by the time the threads start up, they will find work to do in their queues.

Also, the comments in afinter.c are pretty stale since it was reorganized to be a thread based source.

```

static gboolean
_start_worker_thread(LogThreadedDestDriver *self)
{
  gint worker_index = self->workers_started;

////////////////////////// THIS SECTION SHOULD HAPPEN DURING init()
  LogThreadedDestWorker *dw = _construct_worker(self, worker_index);

  msg_debug("Starting dedicated worker thread",
            evt_tag_int("worker_index", worker_index),
            evt_tag_str("driver", self->super.super.id),
            log_expr_node_location_tag(self->super.super.super.expr_node));
  g_assert(self->workers[worker_index] == NULL);
  self->workers[worker_index] = dw;
  self->workers_started++;

////////////////////////// UP TO THIS POINT

  main_loop_create_worker_thread(_worker_thread,
                                 _request_worker_exit,
                                 dw, &self->worker_options);
  _wait_for_startup_finished(dw);
  return !dw->startup_failure;
}

...

/* This method is only used when a LogThreadedDestDriver is directly used
[...]

PR #3559, opened 2021-02-03
LogThreadedDestDriver move worker (and queue) construction into the init phase

szemere:
Solves #3548 

Based on @bazsi 's instructions: moved every construction into the LogThreadedDestDriver init phase, so from now on *on_config_inited* only starts the constructed workers.

KUDO's for @MrAnno for all the help and second opinions.

### DIFF
--- a/lib/logthrdest/logthrdestdrv.c
+++ b/lib/logthrdest/logthrdestdrv.c
@@ -1084,12 +1084,6 @@ log_threaded_dest_driver_start_workers(LogPipe *s)
   return startup_success;
 }
 
-static gboolean
-log_threaded_dest_driver_init(LogPipe *s)
-{
-  return log_threaded_dest_driver_init_method(s);
-}
-
 gboolean
 log_threaded_dest_driver_deinit_method(LogPipe *s)
 {
@@ -1132,7 +1126,7 @@ log_threaded_dest_driver_init_instance(LogThreadedDestDriver *self, GlobalConfig
 
   self->worker_options.is_output_thread = TRUE;
 
-  self->super.super.super.init = log_threaded_dest_driver_init;
+  self->super.super.super.init = log_threaded_dest_driver_init_method;
   self->super.super.super.deinit = log_threaded_dest_driver_deinit_method;
   self->super.super.super.queue = log_threaded_dest_driver_queue;
   self->super.super.super.free_fn = log_threaded_dest_driver_free;
--- a/lib/logthrdest/logthrdestdrv.c
+++ b/lib/logthrdest/logthrdestdrv.c
@@ -930,24 +930,18 @@ _request_worker_exit(gpointer s)
 }
 
 static gboolean
-_start_worker_thread(LogThreadedDestDriver *self)
+_start_worker_thread(LogThreadedDestWorker *self)
 {
-  gint worker_index = self->workers_started;
-  LogThreadedDestWorker *dw = _construct_worker(self, worker_index);
-
   msg_debug("Starting dedicated worker thread",
-            evt_tag_int("worker_index", worker_index),
-            evt_tag_str("driver", self->super.super.id),
-            log_expr_node_location_tag(self->super.super.super.expr_node));
-  g_assert(self->workers[worker_index] == NULL);
-  self->workers[worker_index] = dw;
-  self->workers_started++;
+            evt_tag_int("worker_index", self->worker_index),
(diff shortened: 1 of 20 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

szemere:
Discussed with @mitzkia. cherry-pick'd the Light test framework changes.

szemere:
Rebased on latest master to include python pip fixes.

kira-syslogng:
Build FAILURE

lib/logthrdest/logthrdestdrv.c
```diff
@@ -1080,9 +1065,22 @@ log_threaded_dest_driver_start_workers(LogPipe *s)
...
+                evt_tag_str("driver", self->super.super.id),
+                log_expr_node_location_tag(self->super.super.super.expr_node));
+
+      main_loop_create_worker_thread(_worker_thread,
+                                     _request_worker_exit,
+                                     dw, &self->worker_options);
+
+      _wait_for_startup_finished(dw);
+
+      if (dw->startup_failure)
```
MrAnno, line 1089: Just an opinion, but the original extracted method was cleaner to me.

`_start_worker_thread` could have been refactored even further, because `log_threaded_dest_driver_start_workers()` is a destination driver method, which iterates on its workers and calls their start method.
The details of how a worker is started belong to the worker, so `_start_worker_thread()` could have been a worker method (`LogThreadedDestWorker` as `self`).

MrAnno:
Nice, LGTM 👍🏻 

It seems not only the Python destination, but the Java binding has to be fixed as well (Kira). I guess the same persist_name generation problem happened there as well.

kira-syslogng:
Build SUCCESS

MrAnno (approved):
I had just one optional note about a refactor step.
👍🏻

szemere:
> I had just one optional note about a refactor step.
> 👍🏻

Sorry, I already fixed it locally, just working on something else. I totally agree with your note, I will push it shortly.

kira-syslogng:
Build SUCCESS

MrAnno (dismissed):

alltilla (dismissed):

MrAnno:
Waiting for internal tests.

kira-syslogng:
Build SUCCESS

MrAnno:
I've checked the initialization of all of our threaded destinations:
AMQP, MongoDB, SMTP, SNMP, SQL, stomp, example, HTTP, Java, Kafka, Python, Redis, Riemann

They all seem to be OK now. (I've sent you a PR with the missing parts: szemere/syslog-ng#9)

kira-syslogng:
Build SUCCESS

MrAnno (approved):

modules/afamqp/afamqp.c
```diff
@@ -718,16 +718,16 @@ afamqp_dd_init(LogPipe *s)
...
   if (self->auth_method == AMQP_SASL_METHOD_PLAIN && (!self->user || !self->password))
     {
       msg_error("Error initializing AMQP destination: username and password MUST be set!",
                 evt_tag_str("driver", self->super.super.super.id));
       return FALSE;
     }
 
+  if (!log_threaded_dest_driver_init_method(s))
+    return FALSE;
+
```
lbudai, line 730: hmmm... this part of the changeset is not so readable...
There is a hidden dependency - log_threaded_dest_driver_init_method depends on persist name (actually acquire queue method depends on that).
And currently only the relocation of this codechunk expresses this dependency.

It can goes in, but later we should express this dependency in a more formal manner.

szemere: We discussed it in details with @MrAnno. This is the most bothering part for us too.

What do you think about the proposal to extract the queue initialization method from the init(), making the dependency more explicit? (That way we can also move most of the init methods back to their original place.)

lbudai: What if you extend the ctor with persist_name?
Or, create a `log_threaded_dst_driver_init_with_queue()` method?

szemere:
Resolved Light fw conflict.

kira-syslogng:
Build SUCCESS

MrAnno (approved):

mitzkia:
Internal test has passed.

alltilla (approved):

PR #3561, opened 2021-02-04
Apphook add "REPEAT" run mode for hooks

bazsi:
This PR is split off from #3553 which adds a new run-mode for apphooks, namely the ability to be called any number of times
instead of just once. This simplifies a number of locations where the hook needs to re-register itself.

It also contains fixes to minor bugs, where this re-registration was missing.

Right now it depends on #3553, as it converts that call-site to the new API too. If we drop that patch (ie. the last patch) from the series, this can be merged even prior to that.

### DIFF
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -579,7 +579,7 @@ afinter_register_posted_hook(gint hook_type, gpointer user_data)
 void
 afinter_global_init(void)
 {
-  register_application_hook(AH_CONFIG_CHANGED, afinter_register_posted_hook, NULL);
+  register_application_hook(AH_CONFIG_CHANGED, afinter_register_posted_hook, NULL, AHM_RUN_ONCE);
 }
 
 void
--- a/lib/apphook.c
+++ b/lib/apphook.c
@@ -53,6 +53,7 @@
 typedef struct _ApplicationHookEntry
 {
   gint type;
+  ApplicationHookRunMode run_mode;
   ApplicationHookFunc func;
   gpointer user_data;
 } ApplicationHookEntry;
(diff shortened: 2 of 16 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

MrAnno:
@bazsi Could you resolve the conflict, please?

Should we also add a `developer-note` to the news file about `AHM_RUN_REPEAT`?

bazsi:
rebased, added news.

kira-syslogng:
Build SUCCESS

Kokan (approved):

MrAnno:
@bazsi We've merged #3575, which contains a really nice fix for a bug in the `AH_REOPEN_FILES` affile hook, but this caused another conflict in this PR.

I'm pushing into this branch to resolve the conflict.

The diff is the following:
```diff
-  register_application_hook(AH_REOPEN_FILES, affile_dd_register_reopen_hook, NULL, AHM_RUN_REPEAT);
+  static gboolean initialized = FALSE;
+
+  if (!initialized)
+    {
+      register_application_hook(AH_REOPEN_FILES, affile_dd_register_reopen_hook, NULL, AHM_RUN_REPEAT);
+      initialized = TRUE;
+    }
```

At first glance, this may seem contradictory, but I think it's not: we want the reopen hook to be AHM_RUN_REPEAT, so it will be called every time a reopen is requested.  `static gboolean initialized` is there to prevent this hook to be run multiple times for each reopen event:

Quoting from #3575:
> affile_module_init calls affile_dd_global_init to register an application hook.
> If we reload configuration 1000 times, there will be 1000 entries in applications_hooks.

MrAnno (approved):
Approve.

Kokan (approved):

kira-syslogng:
Build SUCCESS

PR #3563, opened 2021-02-06
Light: Extend error handlings in SyslogNgCli

mitzkia:
This PR adds some missing error handlings to stop() and reload() methods.

### DIFF
--- a/tests/python_functional/src/syslog_ng/syslog_ng_cli.py
+++ b/tests/python_functional/src/syslog_ng/syslog_ng_cli.py
@@ -73,7 +73,7 @@ class SyslogNgCli(object):
         def is_alive(s):
             if not s.is_process_running():
                 self.__process = None
-                raise Exception("syslog-ng is not running")
+                self.__error_handling("syslog-ng is not running")
             return s.__syslog_ng_ctl.is_control_socket_alive()
         return wait_until_true(is_alive, self)
 
--- a/tests/python_functional/src/syslog_ng/syslog_ng_cli.py
+++ b/tests/python_functional/src/syslog_ng/syslog_ng_cli.py
@@ -109,9 +109,11 @@ class SyslogNgCli(object):
         config.write_config(self.__instance_paths.get_config_path())
 
         # effective reload
-        self.__syslog_ng_ctl.reload()
+        result = self.__syslog_ng_ctl.reload()
 
         # wait for reload and check reload result
+        if result["exit_code"] != 0:
+            self.__error_handling("Control socket fails to reload syslog-ng")
         if not self.__wait_for_control_socket_alive():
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

tests/python_functional/src/syslog_ng/syslog_ng_cli.py
```diff
@@ -131,6 +131,7 @@ def stop(self, unexpected_messages=None):
             # wait for stop and check stop result
             if result["exit_code"] != 0:
                 self.__error_handling()
+                raise Exception("Control socket fails to stop syslog-ng")
```
Kokan, line 137: [optional] Do we have use-case not to throw an exception when `self.__error_handling()` is called ? also vice-versa, if not I'd merge them.

Kokan (dismissed):

MrAnno (dismissed):

tests/python_functional/src/syslog_ng/syslog_ng_cli.py
mitzkia: Thanks for question. Yes it is possible. 
My answer is about why we can not merge `raise Exception` with the other `raise Exception` inside the `__handle_core_file()`
Imagine a case when an exit code is not null (syslog-ng can not be stopped with control socket, something went wrong)
Maybe syslog-ng crashes during the stop, but the environment where the test runs, did not create core files (wrong, missing settings for core file pattern or so; core file settings is inherited from the original host in case of docker containers). In this case `error_handling()` will run, but core files will not founded in `__handle_core_file()`. The `error_handling()` function will not raises, in this case the only way to knowing that something went wrong is an other (second) raise outside of `__handle_core_file()`.

While I am answering for this question, I realized a following solution: 

```
    if result["exit_code"] != 0:
        self.__error_handling("Control socket fails to stop syslog-ng")

def __error_handling(self, error_message):
    self.__console_log_reader.dump_stderr()
    self.__handle_core_file()
    raise Exception(error_message)

```

What is your opinion about this second solution?

Kokan: I am not sure that I follow on merging other exceptions. But my suggestion was exactly to point out the 'second' solution.

mitzkia: Thanks, I am agreed with this solution. 
Ok, I will change this PR to this soon.

mitzkia: I have fixed this part. PR has updated.

kira-syslogng:
Build SUCCESS

Kokan (approved):

MrAnno (approved):

PR #3564, opened 2021-02-06
Dbld: Remove CentOS-6 as a supported platform

mitzkia:
As the autobuild for CentOS-6 based syslog-ng image as already stopped, I am removing it from DBLD also.


kira-syslogng:
Build SUCCESS

bazsi (approved):
Looks good to me

MrAnno (approved):

mitzkia:
Not yet understand why 'test_python.test_python' functional testcase failing in one case constantly.
I believe it is not related to this PR.
I am interested could we grab 'test-python.log' as an artifact for 'make func-test' jobs? Without it it is hard to understand why this testcase fails.

mitzkia:
My other idea (totally not related to this PR): 
- it would be really handfull if when I locally use devshell as my local dev enviroment I would be able to call (maybe with some new make targets or template scripts) any CI (github actions) tasks locally, to check the errors on my host itself.
- In my current example, I would like to run build for devshell with the following combinations: python3 + autotools + clang 

I know this is an other topic, but I am just interesting is this an issue only for me, or is it a general problem for everyone?
(For me it is the best if I can reproduce something on my localhost without any problem)

alltilla:
It is red on master, too: https://github.com/syslog-ng/syslog-ng/runs/1840862212

bazsi:
> My other idea (totally not related to this PR):
> 
>     * it would be really handfull when I locally use devshell as my local dev enviroment I would be able to call (maybe with some new make targets or template scripts) any CI (github actions) tasks locally, to check the errors on my host itself.
> 
>     * In my current example, I would like to run build for devshell with the following combinations: python3 + autotools + clang
> 
> 
> I know this is an other topic, but I am just interesting is this an issue only for me, or is it a general problem for everyone?
> (For me it is the best if I can reproduce something on my localhost without any problem)

For this to work, we would need to read the github actions workflow files, which does not seem trivial to me (albeit could be done with a yaml parser). But if you look at the workflow, it's quite possible you can manually recreate the specific environment as most of the logic is actually in the makefiles and not in the yaml.

If you set the proper environment variables just like in the workflow, you should get the same results:
```
          PYTHONVERSION="`printf ${{ matrix.python }} | tail -c 1`"
          PYTHONUSERBASE="${HOME}/python_packages"
          CC="${{ matrix.cc }}"
          CONFIGURE_FLAGS="
            --prefix=${HOME}/install/syslog-ng
            --enable-all-modules
            --disable-java
            --disable-java-modules
            --with-python=${PYTHONVERSION}
            `[ $CC = clang ] && echo '--enable-force-gnu99' || true`
          "
          CMAKE_FLAGS="
            -DCMAKE_C_FLAGS=-Werror
            -DCMAKE_INSTALL_PREFIX=${HOME}/install/syslog-ng
            -DPYTHON_VERSION=${PYTHONVERSION}
          "
          gh_export PYTHONVERSION PYTHONUSERBASE CC CONFIGURE_FLAGS CMAKE_FLAGS
          gh_path "${PYTHONUSERBASE}"

      - name: autogen.sh
        if: matrix.build-tool == 'autotools'
        run: ./autogen.sh

      - name: configure
        if: matrix.build-tool == 'autotools'
        run: |
          mkdir build
          cd build
          ../configure ${CONFIGURE_FLAGS}

      - name: cmake
        if: matrix.build-tool == 'cmake'
        run: |
          mkdir build
          cd build
          cmake ${CMAKE_FLAGS} ..

      - name: make
        working-directory: ./build
        run: make V=1 -j $(nproc)

      - name: make check
        working-directory: ./build
        run: make V=1 check

      - name: make install
        working-directory: ./build
        run: make install
```

Alternatively, we could extract even the bulk of the steps into a separate script, place that into .github/workflows and then have a single step in the workflow file, so that you can execute the said script yourself.

Btw, we might even rely a bit more on dbld when running the devshell workflow: the devshell workflow reuses the same docker image, but in a different setup. (in dbld we are using /install as prefix, the devshell uses ~/install), dbld bootstrap uses CONFIGURE_OPTS instead of CONFIGURE_FLAGS, etc.

By merging these two, the interactive use-cases could easily be improved by moving more of the logic from the workflow files to dbld.

mitzkia:
@bazsi  Thanks for your suggestion. 
I am also found the workflow file for GitHub actions, but it is a little bit problematic to create an environment every time for every different use-case. This is why I came up with the idea when dbld could helps us to create such environments.
I would be happy if we can reach your last step (this would be the step what I imagine): 
"By merging these two, the interactive use-cases could easily be improved by moving more of the logic from the workflow files to dbld."

alltilla:
The failed CI job is not caused by this PR. That will be fixed in #3568.

PR #3565, opened 2021-02-06
Syslog parser drop invalid

bazsi:
This PR contains a combination of a long-needed msg-format related refactor (moving common code out of format modules into the msg-format layer) and the ability for syslog-parser() to drop incorrectly formatted messages (right now it is only RFC5424 that can indicate invalid messages, RFC3164 consumes everything).

The motivation behind this feature is that more and more rfc5424 style messages are actually incorrectly formatted, while still containing the initial "1 " version indicator right after the priority field. Instead of making the parser less strict, the drop-invalid() setting allows the easy identification of these messages, so that an alternative parser can be applied, like this:

```
@version: 3.30

log {
        source { tcp(port(2000) flags(no-parse)); };

        if {
                parser { syslog-parser(drop-invalid(yes) flags(syslog-protocol)); };
        } else {
                # drop RFC5424 indicator as it was not right
                rewrite { subst("^(<[0-9]+>)1 ", "$1"); };
                parser { syslog-parser(drop-invalid(no)); };
        };
        destination { file("log"); };
};
```

This was a long requested feature by the https://github.com/splunk/splunk-connect-for-syslog team @rfaircloth-splunk

### DIFF
--- a/lib/msg-format.c
+++ b/lib/msg-format.c
@@ -28,6 +28,14 @@
 #include "plugin-types.h"
 #include "find-crlf.h"
 
+static gsize
+_rstripped_message_length(const guchar *data, gsize length)
+{
+  while (length > 0 && (data[length - 1] == '\n' || data[length - 1] == '\0'))
+    length--;
+  return length;
+}
+
 void
 msg_format_inject_parse_error(LogMessage *msg, const guchar *data, gsize length, gint problem_position)
 {
@@ -83,6 +91,20 @@ msg_format_postprocess_message(MsgFormatOptions *options, const guchar *data, gs
     msg->flags |= LF_UTF8;
 }
 
+static void
+msg_format_process_message(MsgFormatOptions *options, const guchar *data, gsize length, LogMessage *msg)
+{
+  if ((options->flags & LP_NOPARSE) == 0)
+    {
+      options->format_handler->parse(options, data, length, msg);
+    }
+  else
+    {
+      log_msg_set_value(msg, LM_V_MESSAGE, (gchar *) data, _rstripped_message_length(data, length));
+      msg->pri = options->default_pri;
+    }
+}
+
 void
 msg_format_parse(MsgFormatOptions *options, const guchar *data, gsize length, LogMessage *msg)
 {
@@ -94,7 +116,7 @@ msg_format_parse(MsgFormatOptions *options, const guchar *data, gsize length, Lo
 
   msg_trace("Initial message parsing follows");
   msg_format_preprocess_message(options, data, length, msg);
-  options->format_handler->parse(options, data, length, msg);
+  msg_format_process_message(options, data, length, msg);
   msg_format_postprocess_message(options, data, length, msg);
 }
 
--- a/lib/msg-format.c
+++ b/lib/msg-format.c
@@ -50,7 +50,12 @@ msg_format_inject_parse_error(LogMessage *msg, const guchar *data, gsize length,
 }
 
 static void
-msg_format_postprocess_message(MsgFormatOptions *options, LogMessage *msg)
+msg_format_preprocess_message(MsgFormatOptions *options, const guchar *data, gsize length, LogMessage *msg)
+{
+}
+
+static void
+msg_format_postprocess_message(MsgFormatOptions *options, const guchar *data, gsize length, LogMessage *msg)
 {
   if (options->flags & LP_NO_PARSE_DATE)
     {
@@ -88,8 +93,9 @@ msg_format_parse(MsgFormatOptions *options, const guchar *data, gsize length, Lo
     }
 
   msg_trace("Initial message parsing follows");
+  msg_format_preprocess_message(options, data, length, msg);
   options->format_handler->parse(options, data, length, msg);
-  msg_format_postprocess_message(options, msg);
+  msg_format_postprocess_message(options, data, length, msg);
 }
 
 void
--- a/lib/msg-format.c
+++ b/lib/msg-format.c
@@ -27,6 +27,7 @@
 #include "plugin.h"
 #include "plugin-types.h"
 #include "find-crlf.h"
+#include "scratch-buffers.h"
 
 static gsize
 _rstripped_message_length(const guchar *data, gsize length)
@@ -39,20 +40,23 @@ _rstripped_message_length(const guchar *data, gsize length)
 static void
 msg_format_inject_parse_error(LogMessage *msg, const guchar *data, gsize length, gint problem_position)
 {
-  gchar buf[2048];
+  GString *buf = scratch_buffers_alloc();
 
   log_msg_clear(msg);
 
   msg->timestamps[LM_TS_STAMP] = msg->timestamps[LM_TS_RECVD];
   log_msg_set_value(msg, LM_V_HOST, "", 0);
 
-  g_snprintf(buf, sizeof(buf), "Error processing log message: %.*s>@<%.*s", (gint) problem_position-1,
-             data, (gint) (length-problem_position+1), data+problem_position-1);
+  if (problem_position > 0)
+    g_string_printf(buf, "Error processing log message: %.*s>@<%.*s", (gint) problem_position-1,
+                    data, (gint) (length-problem_position+1), data+problem_position-1);
+  else
+    g_string_printf(buf, "Error processing log message: %.*s", (gint) length, data);
 
-  log_msg_set_value(msg, LM_V_MESSAGE, buf, -1);
+  log_msg_set_value(msg, LM_V_MESSAGE, buf->str, buf->len);
   log_msg_set_value(msg, LM_V_PROGRAM, "syslog-ng", 9);
-  g_snprintf(buf, sizeof(buf), "%d", (int) getpid());
-  log_msg_set_value(msg, LM_V_PID, buf, -1);
+  g_string_printf(buf, "%d", (int) getpid());
+  log_msg_set_value(msg, LM_V_PID, buf->str, buf->len);
 
   msg->pri = LOG_SYSLOG | LOG_ERR;
 }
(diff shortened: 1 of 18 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
dropping "WIP", this should be good enough for review. The changes are not complex, however there's a series of them. I'd recommend going patch-by-patch.

The motiviation behind all the changes, the feature itself is trivial, once all the preparation is there.

kira-syslogng:
Build FAILURE

mitzkia:

kira-syslogng:
Build FAILURE

rfaircloth-splunk:
looking forward to having this in the toolbox

lib/msg-format.c
```diff
@@ -94,7 +116,7 @@ msg_format_parse(MsgFormatOptions *options, const guchar *data, gsize length, Lo
 
   msg_trace("Initial message parsing follows");
   msg_format_preprocess_message(options, data, length, msg);
-  options->format_handler->parse(options, data, length, msg);
+  msg_format_process_message(options, data, length, msg);
```
gaborznagy: In case of `no-parse` flag, we still call `msg_format_postprocess_message`, and thus setting additional flags from config (e.g. assume-utf8, local).
I'm okay with this, but previously neither `syslog-format`, nor `linux-kmsg` set these flags.

lib/msg-format.c
```diff
@@ -124,12 +124,25 @@ msg_format_parse(MsgFormatOptions *options, LogMessage *msg, const guchar *data,
...
 
-  if (!msg_format_process_message(options, msg, data, length, &problem_position))
-    msg_format_inject_parse_error(msg, data, _rstripped_message_length(data, length), problem_position);
+  if (!msg_format_parse_conditional(options, msg, data, length, &problem_position))
+    {
+      msg_format_inject_parse_error(msg, data, _rstripped_message_length(data, length), problem_position);
 
-  msg_format_postprocess_message(options, msg, data, length);
+      /* the injected error message needs to be postprocessed too */
+      msg_format_postprocess_message(options, msg, data, length);
```
gaborznagy, line 144: Why do we need to post-process messages that are failed to parse?
Isn't processing `no-multi-line` could interfere with the `>@<` mark from the error-injecting function?

I'm okay that `msg_format_parse_conditional` does post-processing, as the message will be passed over for further processing in the log path.

lib/msg-format.c
```diff
@@ -72,6 +72,8 @@ msg_format_postprocess_message(MsgFormatOptions *options, LogMessage *msg)
           p++;
         }
     }
+  if (options->flags & LP_LOCAL)
+    msg->flags |= LF_LOCAL;
```
gaborznagy, line 93: I would move both `local` and `assume-utf8` flags to preprocess.
Reasons:
- syslog format handler in legacy-hander uses the `LF_LOCAL` flag
- regardles of `no-parse` flag, the post-processing is done unconditionally, so it doesn't matter if it's in pre- or post-processing.

modules/syslogformat/syslog-parser.c
```diff
@@ -36,8 +44,16 @@ syslog_parser_process(LogParser *s, LogMessage **pmsg, const LogPathOptions *pat
             evt_tag_str ("input", input),
             evt_tag_printf("msg", "%p", *pmsg));
 
-  msg_format_parse(&self->parse_options, msg, (guchar *) input, input_len);
-  return TRUE;
+  if (self->drop_invalid)
```
gaborznagy, line 47: Wouldn't improve things a `G_UNLIKELY` here? Just like we do in `LogParser` with template option

gaborznagy (commented):
I only have some questions/comments.

The new feature looks good to me!

lib/msg-format.c
gaborznagy: I think this is the reason that kira fails, as one test with flag `kernel` fails.

lib/msg-format.c
```diff
@@ -58,6 +59,19 @@ msg_format_postprocess_message(MsgFormatOptions *options, LogMessage *msg)
                              time_zone_info_get_offset(options->recv_time_zone_info,
                                                        msg->timestamps[LM_TS_RECVD].ut_sec));
     }
+  if (G_UNLIKELY(options->flags & LP_NO_MULTI_LINE))
```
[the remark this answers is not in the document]
bazsi: I think it makes better sense that we apply flags to both parsed/unparsed/error cases. If we asked syslog-ng not to give us newlines in messages, that's possibly because we  have something in our processing pipeline that doesn't handle that.

If in rare cases we don't perform as asked, then that's not that intuitive.

So I think it's better to fix this behavior.

lib/msg-format.c
```diff
@@ -120,7 +121,12 @@ msg_format_parse(MsgFormatOptions *options, const guchar *data, gsize length, Lo
 
   msg_trace("Initial message parsing follows");
   msg_format_preprocess_message(options, data, length, msg);
-  msg_format_process_message(options, data, length, msg);
+
+  gsize problem_position = 0;
```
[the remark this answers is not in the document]
bazsi: I'll fix this.

lib/msg-format.c
bazsi: The behavioral change was intentional, and again it is the "should have been this way" case.

The utf8 validation/sanitization is not yet centralized (but probably should be), I didn't touch that because of the performance implications.

I could easily centralize assume-utf8, so I did that. And the message can be utf-8 even if it wasn't parsed.

The local flag was set even previously depending on the LP_LOCAL flag for the source, which is done by the file() source.

lib/msg-format.c
bazsi: The same reason as above, failed parsing results in a message too.

lib/msg-format.c
bazsi: Can you elaborate that?

modules/syslogformat/syslog-parser.c
bazsi: I can add this.

bazsi: Hmm.. I have second thoughts on this. A config may easily use drop_invalid mode exclusively, in which case G_UNLIKELY is not true.

In the case of the "template" option, that branch is slow anyway, so the G_UNLIKELY is the least of the concerns anyway.

So I would just leave it to the compiler.

lib/msg-format.c
bazsi: hmmm. I have now pushed a WIP patch on top, which clears all the flags in log_msg_clear(). I still have to give some more thoughts on the behavior, but that means that the error message becomes completely uncoupled with the original one.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please branch=syslog-parser-update-wrong-format-test;

modules/syslogformat/syslog-parser.c
gaborznagy: I see, so this wouldn't make too much sense then.

kira-syslogng:
Build SUCCESS

gaborznagy:
I haven't checked why does the `test_secure_logging` UT fail only with cmake.
I've updated a test that was affected by the injected error message format, so kira is green now.

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please branch=syslog-parser-update-wrong-format-test;

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please branch=syslog-parser-update-wrong-format-test;

kira-syslogng:
Build FAILURE

gaborznagy:
(oops, forgot to rebase the test branch, I will restart the kira test)

gaborznagy:
@kira-syslogng retest this please branch=syslog-parser-update-wrong-format-test;

kira-syslogng:
Build SUCCESS

bazsi:
@kira-syslogng retest this please branch=syslog-parser-update-wrong-format-test;

kira-syslogng:
Build SUCCESS

bazsi:
@kira-syslogng retest this please branch=syslog-parser-update-wrong-format-test;

kira-syslogng:
Build SUCCESS

bazsi:
@kira-syslogng retest this please branch=syslog-parser-update-wrong-format-test;

kira-syslogng:
Build SUCCESS

Kokan (approved):

Kokan:
Also if possible please write a news entry under `news/` directory. But this does not block the merge, if somebody merges this as is, they can write one and open a PR.

bazsi:
added news file + rebased

kira-syslogng:
Build FAILURE

bazsi:
@kira-syslogng retest this please branch=syslog-parser-update-wrong-format-test;

kira-syslogng:
Build SUCCESS

Kokan (approved):

gaborznagy (approved):

gaborznagy:
I've created an internal doc update ticket about the new option and the changed behaviour of msg-format flags (i.e. from now on if "no-parse" flag is used additional flags e.g. "no-multi-line" can be used as well.)

gaborznagy:
@bazsi I wonder: did we intentionally not document the injected error message?
This is the only thing in the documentation: 
> If syslog-ng OSE cannot parse a message, it results in an error

I would mention the details of the error message (rewritten program and message fields, modified facility and severity values).

bazsi:
There was no intention, it just happened this way.

Documenting the error handling might indeed be useful, as well as the
drop-invalid option in the context of syslog-parser ().

gaborznagy:
Thanks, I've submitted a ticket to the doc team.

PR #3568, opened 2021-02-10
python: do not double free seqnum and template_options

alltilla:
When these two objects are passed to `PyObject_SetAttrString()`, their reference counter is incremented. In the next line, we drop our own reference to them with `Py_DECREF()`, but still keep track of them to free it later in `_py_free_bindings()`.

It is not needed, the python lib will take care of the reference, that was added in `PyObject_SetAttrString().

Because we do not need these objects in our code, it is the easiest, to not store them in our struct.

### DIFF
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -55,8 +55,6 @@ typedef struct
     PyObject *open;
     PyObject *send;
     PyObject *flush;
-    PyObject *log_template_options;
-    PyObject *seqnum;
     PyObject *generate_persist_name;
     GPtrArray *_refs_to_clean;
   } py;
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

modules/python/python-dest.c
```diff
@@ -405,8 +405,6 @@ _py_init_bindings(PythonDestDriver *self)
   g_ptr_array_add(self->py._refs_to_clean, self->py.open);
   g_ptr_array_add(self->py._refs_to_clean, self->py.flush);
   g_ptr_array_add(self->py._refs_to_clean, self->py.send);
-  g_ptr_array_add(self->py._refs_to_clean, self->py.log_template_options);
-  g_ptr_array_add(self->py._refs_to_clean, self->py.seqnum);
```
MrAnno, line 409: In case we store these object in `PythonDestDriver`, we should not decrement our reference counter after setatttr, because the reference added by `PyObject_SetAttrString()` can easily be decremented on the Python side by assigning a new value to that attribute. `pyseqnum` and `pylog_template_options` would point to invalid memory in that case.

I checked and we don't really use `self->py.log_template_options` and `self->py.seqnum` after the setattr call, so I think we could just remove these refs from the C struct.

MrAnno (commented):
Nice catch!

modules/python/python-dest.c
alltilla: Yes-yes, good point!

Let's just make them local variables then!

alltilla: Fixed!

MrAnno (approved):

kira-syslogng:
Build SUCCESS

alltilla:
Updated: added `-DCMAKE_BUILD_TYPE=Debug` to `CMAKE_FLAGS`, to have debug symbols.

Test run with a manually generated corefile: https://github.com/alltilla/syslog-ng/actions/runs/554768804

kira-syslogng:
Build SUCCESS

MrAnno (approved):

gaborznagy (approved):

PR #3569, opened 2021-02-11
scl: add support for "@cee" prefixed JSON payload

bazsi:
```
    This format was originally specified by Mitre in its CEE specification,
    later adopted by the Lumberjack project run by Fedora/RedHat.
    
    https://cee.mitre.org/
    https://www.phoronix.com/scan.php?page=news_item&px=MTA2NDU
    
    The fedorahosted site is not completely gone.
    
    Now both projects are defunct, but there seems to be some adoption, for
    instance the JSON logging format of sudo does support emitting messages
    using this format.
    
    https://www.syslog-ng.com/community/b/blog/posts/parsing-sudo-json-logs-building-a-syslog-ng-configuration
    
    
    rsyslog supports this as well:
    https://www.rsyslog.com/json-elasticsearch/
    
```


kira-syslogng:
Build SUCCESS

MrAnno (approved):
LGTM.

A question:
I've noticed that the `syslog` topic is embedded into `default-network-drivers()` (which is great), but it is also used in the `system()` source.

Is that intentional that we add Mitre CEE (CTS) and Lumberjack (and junos, panos, etc.) parsing to the system source?

Update:
> the JSON logging format of sudo does support emitting messages using this format.

Sorry, I just saw this. I think that answers the CEE part of the question.

bazsi:
On Sun, Feb 14, 2021, 17:35 László Várady <notifications@github.com> wrote:

> *@MrAnno* approved this pull request.
>
> LGTM.
>
> A question:
> I've noticed that the syslog topic is embedded into
> default-network-drivers() (which is great), but it is also used in the
> system() source.
>
> Is that intentional that we add Mitre CEE (CTS) and Lumberjack (and junos,
> panos, etc.) parsing to the system source?
>

It is intentional, since that's also "syslog", and it allows us to push
parsing to the edge, so the central only needs to parse ewmm.

>

Kokan (approved):

PR #3572, opened 2021-02-16
Change PERSIST_FILE_WATERMARK size

szemere:
Persist file uses a "pre-allocation" mechanism to circumvent some side effects of a running out of space scenario.

More on that in PR #1785, especially in 45a4157.

Thus PERSIST_FILE_WATERMARK must be bigger than any entry we want to store in the persist file. (Including the extra headers and flags.)

One of our internal module has an 8k buffer. Considering the extra space for the headers, the current value is: 8k+256 bytes.

---
Note: I have two alternative follow up plans. Currently work in progress, but I do not want to block this fix: 
A) https://github.com/szemere/syslog-ng/tree/persist_changes_wip Create a static assert on struct sizes. 
B) https://github.com/szemere/syslog-ng/tree/persist_changes_wip2 Eliminate the pre-allocation logic. This way we can always check on the real alloc size, so there is no need for heuristics.

### DIFF
--- a/lib/persist-state.c
+++ b/lib/persist-state.c
@@ -37,7 +37,7 @@
 
 #define PERSIST_FILE_INITIAL_SIZE 16384
 #define PERSIST_STATE_KEY_BLOCK_SIZE 4096
-#define PERSIST_FILE_WATERMARK 8448
+#define PERSIST_FILE_MAX_ENTRY_SIZE 8448
 
 /*
  * The syslog-ng persistent state is a set of name-value pairs,
@@ -209,7 +209,7 @@ _commit_store(PersistState *self)
 static gboolean
 _check_watermark(PersistState *self)
 {
-  return (self->current_ofs + PERSIST_FILE_WATERMARK < self->current_size);
+  return (self->current_ofs + PERSIST_FILE_MAX_ENTRY_SIZE < self->current_size);
 }
 
 static inline gboolean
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

MrAnno (approved):
https://github.com/szemere/syslog-ng/tree/persist_changes_wip2 seems to be a good idea, it would be a nice improvement.

lib/persist-state.c
```diff
@@ -37,7 +37,7 @@
 
 #define PERSIST_FILE_INITIAL_SIZE 16384
 #define PERSIST_STATE_KEY_BLOCK_SIZE 4096
-#define PERSIST_FILE_WATERMARK 4096
+#define PERSIST_FILE_WATERMARK 8448
```
bazsi, line 40: this should be aligned on a 4k boundary.

bazsi: let's add an assert into the persist file allocation function that would limit the allocation to sizes that we can actually guarantee.

bazsi (dismissed):

lib/persist-state.c
MrAnno: Is the 4k page-boundary necessary? `PERSIST_FILE_WATERMARK` seems to be just a free-space threshold, syslog-ng always allocates `PERSIST_FILE_INITIAL_SIZE=16384` chunks, and `_grow_store()` makes sure it's page-aligned.

MrAnno: Is the 4k page-boundary necessary? `PERSIST_FILE_WATERMARK` seems to be just a free-space threshold, syslog-ng always allocates `PERSIST_FILE_INITIAL_SIZE=16384` chunks, and `_grow_store()` makes sure it's page-aligned with `getpagesize()` (I think that might return 8 and 16k as well).

bazsi: Ok, that's ok this way then.

MrAnno: @bazsi Thank you.

kira-syslogng:
Build SUCCESS

MrAnno (approved):

bazsi (approved):

PR #3575, opened 2021-02-17
Fix improper initialization in affile and LogWriter

### LINKED ISSUE #3574: Heap consumption increased unexpectedly after reloading several times
# syslog-ng
## Version of syslog-ng
3.29

## Platform
Linux

# Issue
## Failure
Memory 

## Steps to reproduce
1. Use the configuration in next section
2. Reload syslog-ng 1024 times
3. Massif result shows that heap consumption increased

## Configuration
```
@version: 3.29

source src {
        internal();
};

destination dest {
        file("/dev/null");
};

log {
        source(src);
        destination(dest);
};
```

image

0140454:
When reloading configuration, syslog-ng performs the following operations
1. Load modules
2. Initialize modules (calling `%s_module_init`)
3. ...
4. Create log pipes
5. Initialize log pipes

There are two issues:

---

`affile_module_init` calls `affile_dd_global_init` to register an application hook.

However, it doesn't use any flag to avoid registering hook repeatedly.
Therefore, if we reload configuration 1000 times, there will be 1000 entries in `applications_hooks`.

This can be observed by executing the following commands in gdb
```
set $hook=application_hooks
while ($hook != 0)
  p *((ApplicationHookEntry*)($hook->data))
  set $hook=$hook->next
end
```

---

`affile-dest` creates LogWriter if necessary ([code](https://github.com/syslog-ng/syslog-ng/blob/master/modules/affile/affile-dest.c#L194)), and initialize it every time ([code](https://github.com/syslog-ng/syslog-ng/blob/master/modules/affile/affile-dest.c#L205)).

Currently, `log_writer_init` uses `log_pipe_add_info` to add `writer` to `LogWriter::info`. If we reload 1000 times, `affile_dw_init` will be performed 1000 times, `log_writer_new` and `log_writer_init` will be called 1 time and 1000 times repectively.

---

The memory allocated for hook and LogWriter::info entry will not be freed until syslog-ng exits.
Therefore, massif shows like the following

image

Apply this patchset, massif shows like the following

image_fixed

---

Fix #3574

### DIFF
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -1452,8 +1452,6 @@ log_writer_init(LogPipe *s)
       log_writer_postpone_mark_timer(self);
     }
 
-  log_pipe_add_info(s, "writer");
-
   return TRUE;
 }
 
@@ -1715,6 +1713,8 @@ log_writer_new(guint32 flags, GlobalConfig *cfg)
   g_static_mutex_init(&self->suppress_lock);
   g_static_mutex_init(&self->pending_proto_lock);
   self->pending_proto_cond = g_cond_new();
+
+  log_pipe_add_info((LogPipe *)self, "writer");
 
   return self;
 }
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -1714,7 +1714,7 @@ log_writer_new(guint32 flags, GlobalConfig *cfg)
   g_static_mutex_init(&self->pending_proto_lock);
   self->pending_proto_cond = g_cond_new();
 
-  log_pipe_add_info((LogPipe *)self, "writer");
+  log_pipe_add_info(&self->super, "writer");
 
   return self;
 }
--- a/news/bugfix-3574.md
+++ b/news/bugfix-3574.md
@@ -1,1 +1,1 @@
-`affile`: Fix improper initialization in affile and LogWriter
+`affile`: Fix improper initialization in affile and LogWriter to avoid memory leak when reloading
--- a/news/bugfix-3574.md
+++ b/news/bugfix-3574.md
@@ -1,0 +1,1 @@
+`affile`: Fix improper initialization in affile and LogWriter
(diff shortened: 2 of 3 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

szemere:
ok to test

Kokan (dismissed):
Thanks, superb PR!

news/bugfix-3574.md
```diff
@@ -0,0 +1 @@
+`affile`: Fix improper initialization in affile and LogWriter
```
bazsi, line 1: I would mention memory leaks here along with the severity of the leaks, so that a user can understand how much her/his installation might be impacted.

lib/logwriter.c
```diff
@@ -1716,6 +1714,8 @@ log_writer_new(guint32 flags, GlobalConfig *cfg)
   g_static_mutex_init(&self->pending_proto_lock);
   self->pending_proto_cond = g_cond_new();
 
+  log_pipe_add_info((LogPipe *)self, "writer");
+
```
bazsi, line 1718: using &self->super as the argument keeps it typesafe, it's usually better than the cast.

bazsi (dismissed):
I've filed two nitpicks, but otherwise it's good enough for me to be merged.

news/bugfix-3574.md
0140454: OK, news file updated.

lib/logwriter.c
0140454: Thanks for this tips. I have changed to `&self->super`.

bazsi (approved):

Kokan (approved):

PR #3582, opened 2021-02-19
get-github-actions-status: add new script

alltilla:
This will be used to poll the status of the GitHub Actions jobs in the Jenkins dashboard.

### DIFF
--- a/contrib/get-github-actions-status.py
+++ b/contrib/get-github-actions-status.py
@@ -1,0 +1,53 @@
+#!/bin/env python3
+
+import json
+import sys
+import urllib.parse
+import urllib.request
+from time import sleep
+
+
+def get_workflow_runs(workflow):
+    url = "https://api.github.com/repos/syslog-ng/syslog-ng/actions/workflows/{}.yml/runs?branch=master".format(workflow)
+    request = urllib.request.Request(url, None, {})
+    try:
+        with urllib.request.urlopen(request) as response:
+            binary_response = response.read()
+    except urllib.error.HTTPError:
+        print('"' + workflow + '" is not a valid workflow. See .github/workflows/')
+        sys.exit(1)
+
+    runs = json.loads(binary_response.decode())["workflow_runs"]
+    if len(runs) == 0:
+        print('No runs have been executed of "' + workflow + '" yet.')
+        sys.exit(1)
+
+    return runs
+
+
+def main():
+    workflow = sys.argv[1]
+    while True:
+        runs = get_workflow_runs(workflow)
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

contrib/get-github-actions-status.py
```diff
@@ -0,0 +1,44 @@
...
+    with urllib.request.urlopen(request) as response:
+        binary_response = response.read()
+
+    return json.loads(binary_response.decode())["workflow_runs"]
+
+
+def main():
+    workflow = sys.argv[1]
+    while True:
+        runs = filter(lambda run: run["name"] == workflow, get_all_runs())
```
gaborznagy, line 22: You can save the usage of `filter()` if you query the specific github actions:
https://docs.github.com/en/rest/reference/actions#list-workflow-runs
e.g. `https://api.github.com/repos/syslog-ng/syslog-ng/actions/workflows/devshell.yml/runs`

contrib/get-github-actions-status.py
```diff
@@ -0,0 +1,44 @@
...
+        latest_run = max(runs, key=lambda run: run["updated_at"])
```
gaborznagy, line 23: minor improvement idea for input validation: if the `workflow` is not found in the list of github workflows, then the call to `max()` returns a ValueError. 

`get_all_runs` could have some error handling or an input validation after the `filter()`

gaborznagy (approved):
I have minor suggestions, looks good to me!

Thanks!

contrib/get-github-actions-status.py
szemere: note: I tried to simplify it further by using the "per_page" attribute in the query. My tests shows that it will return the latest run, which will simplify the code, and reduce the amount of transferred data. However there is no mention of any ordering in the API documentation :( So no luck there.

Otherwise LGTM!

contrib/get-github-actions-status.py
alltilla: Thanks, this is much better! :)

contrib/get-github-actions-status.py
alltilla: @gaborznagy I added the necessary checks, which happened to be different after changing the API, but I think they serve the same purpose, as you mentioned.

@szemere Yes, I also had the same observation as you, but no definite thing in the documentation. I prefer to keep the `max` logic.

gaborznagy (approved):

kira-syslogng:
Build SUCCESS

szemere (approved):

PR #3583, opened 2021-02-22
kafka-c: fix double ack on re-added messages

gaborznagy:
Delivery_report_callback re-adds the message to the head of the queue for resending.
(with more workers, the message is possibly re-added to a different worker thread's queue than the original worker who tried to sent out the message, as we always poll from worker#0 and re-add the message to worker#0's queue).

The problem was, that the message was re-added to the queue with acknowledgement re-enabled.
This can cause trouble in both flow-controlled and non flow-controlled cases, as even with a flow-controlled log path, the message is already acknowledged since kafka-c driver's insert method returned with SUCCESS, and delivery failure only detected in the delivery_report callback.

Acknowledgement in kafka-c has to be revised, so we can acknowledge at the right time. Currently, flow-control doesn't work as it is intended, the messages are ack-ed back before the delivery callback could give us feedback on the message sending.
This will be done in a later step.

### DIFF
--- a/modules/kafka/kafka-dest-driver.c
+++ b/modules/kafka/kafka-dest-driver.c
@@ -282,7 +282,7 @@ _kafka_delivery_report_cb(rd_kafka_t *rk,
     {
       LogThreadedDestWorker *worker = (LogThreadedDestWorker *) self->super.workers[0];
       LogQueue *queue = worker->queue;
-      LogPathOptions path_options = LOG_PATH_OPTIONS_INIT;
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

szemere (approved):

bazsi (approved):
Nice catch.

lbudai (approved):

PR #3585, opened 2021-02-24
afuser: support time-reopen() 

alltilla:
Support `time-reopen()` option. `syslog-ng` will wait `time-reopen()` seconds (instead of the previously hard-coded 600), when it could not connect to the usertty terminal, then try to connect again.

Add per driver `time-reopen()` option, for drivers, which were previously supporting it, but was not able to set it per driver.

### DIFF
--- a/lib/logthrdest/logthrdestdrv.c
+++ b/lib/logthrdest/logthrdestdrv.c
@@ -69,6 +69,14 @@ log_threaded_dest_driver_set_batch_timeout(LogDriver *s, gint batch_timeout)
   self->batch_timeout = batch_timeout;
 }
 
+void
+log_threaded_dest_driver_set_time_reopen(LogDriver *s, time_t time_reopen)
+{
+  LogThreadedDestDriver *self = (LogThreadedDestDriver *) s;
+
+  self->time_reopen = time_reopen;
+}
+
 /* this should be used in combination with LTR_EXPLICIT_ACK_MGMT to actually confirm message delivery. */
 void
 log_threaded_dest_worker_ack_messages(LogThreadedDestWorker *self, gint batch_size)
--- a/lib/logthrdest/logthrdestdrv.c
+++ b/lib/logthrdest/logthrdestdrv.c
@@ -1061,7 +1061,7 @@ log_threaded_dest_driver_init_method(LogPipe *s)
 
   self->under_termination = FALSE;
 
-  if (cfg && self->time_reopen == -1)
+  if (self->time_reopen == -1)
     self->time_reopen = cfg->time_reopen;
 
   self->shared_seq_num = GPOINTER_TO_INT(cfg_persist_config_fetch(cfg,
--- a/modules/afuser/afuser.c
+++ b/modules/afuser/afuser.c
@@ -36,6 +36,7 @@ typedef struct _AFUserDestDriver
   LogDestDriver super;
   GString *username;
   time_t disable_until;
+  time_t time_reopen;
 } AFUserDestDriver;
 
 #ifdef SYSLOG_NG_HAVE_UTMPX_H
@@ -105,6 +106,14 @@ _get_utmp_username(UtmpEntry *ut)
 
 G_LOCK_DEFINE_STATIC(utmp_lock);
 
+void
+afuser_dd_set_time_reopen(LogDriver *s, time_t time_reopen)
+{
+  AFUserDestDriver *self = (AFUserDestDriver *) s;
+
+  self->time_reopen = time_reopen;
+}
+
 static void
 afuser_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options)
 {
@@ -149,18 +158,20 @@ afuser_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options)
           fd = open(line, O_NOCTTY | O_APPEND | O_WRONLY | O_NONBLOCK);
           if (fd == -1)
             {
-              msg_error("Opening tty device failed, disabling usertty() for 10 minutes",
+              msg_error("Opening tty device failed, disabling usertty() for time-reopen seconds",
                         evt_tag_str("user", _get_utmp_username(ut)),
                         evt_tag_str("line", line),
+                        evt_tag_int("time_reopen", self->time_reopen),
                         evt_tag_error("errno"));
-              self->disable_until = now + 600;
+              self->disable_until = now + self->time_reopen;
               continue;
             }
           if (write(fd, buf, strlen(buf)) < 0 && errno == EAGAIN)
             {
-              msg_notice("Writing to the user terminal has blocked for writing, disabling for 10 minutes",
-                         evt_tag_str("user", self->username->str));
-              self->disable_until = now + 600;
+              msg_notice("Writing to the user terminal has blocked for writing, disabling for time-reopen seconds",
+                         evt_tag_str("user", self->username->str),
+                         evt_tag_int("time_reopen", self->time_reopen));
+              self->disable_until = now + self->time_reopen;
             }
           close(fd);
         }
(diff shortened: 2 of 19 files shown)


kira-syslogng:
Build SUCCESS

modules/afuser/afuser.c
```diff
@@ -189,5 +192,7 @@ afuser_dd_new(gchar *user, GlobalConfig *cfg)
   self->super.super.super.queue = afuser_dd_queue;
   self->super.super.super.free_fn = afuser_dd_free;
   self->username = g_string_new(user);
+  self->time_reopen = cfg ? cfg->time_reopen : 60;
```
szemere, line 195: Small typo here. Otherwise LGTM!

modules/afuser/afuser.c
```diff
@@ -149,18 +150,20 @@ afuser_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options)
...
           if (fd == -1)
             {
-              msg_error("Opening tty device failed, disabling usertty() for 10 minutes",
+              msg_error("Opening tty device failed, disabling usertty() for time-reopen seconds",
                         evt_tag_str("user", _get_utmp_username(ut)),
                         evt_tag_str("line", line),
+                        evt_tag_int("time_reopen", self->time_reopen),
                         evt_tag_error("errno"));
-              self->disable_until = now + 600;
+              self->disable_until = now + self->time_reopen;
```
Kokan, line 158: questions:
1. why the default 600 is changed ?
2. if you plan to override the `time-reopen()` from global if configured, please allow local settings as well, so user could change per afuser destination.

alltilla: 1. There are some problems with `time-reopen()` in general. When `GlobalConfig` is created its `time_reopen` is set to 60, and every driver will get that value. So actually it does not matter what value I put here, it will always come from the `GlobalConfig`.
2. There are other drivers, which support `time-reopen()`, but none of them support it on the driver level, only from the global `options{}` block. So I decided to not support it here, either.

So in general, I tried to follow the current `time-reopen()` logic. But your questions are valid, and I think that we should rework `time-reopen()` a bit:
1. Set `GlobalConfig`'s `time_reopen` to -1, by default.
2. Follow up this change in each driver; if they were using the `GlobalConfig` default, set their default to 60.
3. Add `KW_TIME_REOPEN` to driver level options, where the driver supports it.
4. Run tests / add new tests to check regression.

It will be a bit bigger patch, but I think this rework was due for some time. What do you think?

modules/afuser/afuser.c
alltilla: Thanks for the review, I addressed the problems with the default value in another comment: https://github.com/syslog-ng/syslog-ng/pull/3585#discussion_r582613561

alltilla:
I'm expecting KIRA to fail, because this probably needs some internal changes. Please do not merge, until we have prepared the internal changes.

kira-syslogng:
Build SUCCESS

alltilla:
Well this is awkward, because KIRA did not fail.. I will check if I am wrong, or we are missing some `YFLAGS` in the internal build.

lib/logthrdest/logthrdestdrv.c
```diff
@@ -1061,7 +1061,7 @@ log_threaded_dest_driver_init_method(LogPipe *s)
 
   self->under_termination = FALSE;
 
-  if (cfg && self->time_reopen == -1)
+  if (self->time_reopen == -1)
```
Kokan, line 1072: Do not remove that check. You may know that is the case, still the compiler does not.

The logic is the following the `cfg` might be `NULL` or not. I have no clue, nor can I find out in compile time so I assume both. The next step is, well you dereference a pointer that could be `NULL`, that is undefined behavior. At that point compiler are free to do what its heart desires, and they do if they smell opportunity of performance gain. (likely this does what you would expect, which is worse because let you think it is okay).

Kokan (changes requested):

lib/logthrdest/logthrdestdrv.c
alltilla: This is new to me.

I am just thinking, why isn't this a problem in every one of our variables in every one of our functions? Like the `self` here, or we can find a lot of examples. We do not check for every variable to not be `NULL` in the beginning of our functions.

Why is this case different, and how can someone detect it?

kira-syslogng:
Build FAILURE

lib/logthrdest/logthrdestdrv.c
alltilla: We discussed this IRL, and added an assert to the getter.

kira-syslogng:
Build SUCCESS

Kokan (approved):

alltilla:
Internal MR !317 is prepared. This PR is now ready to review and merge.

amdrsantos:
@alltilla Do you know when this fix will be released? Which release and date?

alltilla:
@amdrsantos

We will release 3.32.1 approximately 2 months from now, if everything goes well. This code will most probably be shipped with that.

From where do you obtain your syslog-ng package? Is it okay for you to use an unreleased package? As you can see it from the "Binary packages" CI run, we generate packages for every PR, and anyone can do that by themselves also with the help of `dbld`.

If you need an installer sooner, I would suggest to wait until this PR gets its 2nd approve, and gets merged, then you can generate a package from the master branch.

lib/logthrdest/logthrdestdrv.c
bazsi: btw, I think the compiler behavior is the other way around, e.g. if it does see an unconditional dereference, it would assume the pointer is NOT null and thus drop the NULL checks, as explained here: https://lwn.net/Articles/342330/

I don't see how the code with the original change would cause suprising code to be generated.

bazsi (approved):
There's one note I was thinking about, namely the fact that we are peeking into LogWriterOptions in unrelated code, with expressions like "owner->writer_options->time_reopen". With that we implement an assumption that a driver is actually using LogWriter to implement its functionality.

With this we couple with an implementation detail in unrelated code. 

This does not necessarily increase coupling today, as we might have the same assumption in other parts of the same code. Also, it might be a safe assumption, as I can't see why we would want to change this implementation in the future.

I just wanted to mention it. Approve on my part, just consider the above. Thanks.

amdrsantos:
@alltilla Thanks, I will next available release

amdrsantos:
Hey @alltilla, I already saw that syslog-ng 3.32.1 was released.
Any idea when it will be available for Debian10 at http://download.opensuse.org/repositories/home:/laszlo_budai:/syslog-ng/Debian_10/ ?
Thanks in advance.
Alex

alltilla:
Hi @amdrsantos,

It will probably be done in 1-2 weeks, we will send an email to the syslog-ng mail list, when it is ready.

Attila

amdrsantos:
@alltilla do you know if Debian 10 (buster) packaging of syslog-ng 3.32.1 will be released shortly?

alltilla:
Hi @amdrsantos 

You might have seen on the mail list, that the OBS packages are discontinued. I would advise to use `dbld` to generate the packages manually.

I generated one for you with the following commands, but you can do it manually, too:
```
wget https://github.com/syslog-ng/syslog-ng/releases/download/syslog-ng-3.32.1/syslog-ng-3.32.1.tar.gz
tar -xzvf syslog-ng-3.32.1.tar.gz
cd syslog-ng-3.32.1
./dbld/rules package-debian-buster
cd dbld/build/
zip syslog-ng-3.32.1-debian-buster-packages.zip debian-buster/*
```

We will work on providing packages in the future somewhere, but we have not decided the details yet.

[syslog-ng-3.32.1-debian-buster-packages.zip](https://github.com/syslog-ng/syslog-ng/files/6574979/syslog-ng-3.32.1-debian-buster-packages.zip)

Cheers,
Attila

amdrsantos:
@alltilla Is it possible to use dbld to generate the packages for arm64?
If yes, do you know how?
Thanks in advance,
Alex

bazsi:
If you are running on arm64 it should work out of the box. Cross compiling
should also be possible but is not supported atm.

PR #3588, opened 2021-02-25
dbld: install patched bison into the tarball image

bazsi:


kira-syslogng:
Build SUCCESS

bazsi:
This is a workaround for Github.com/skins/bison/issues/74 that we discussed IRL the other day. With this we can at least release a 3.31 until a better fix becomes available.

Kokan (approved):

gaborznagy (approved):
Approve from my side.

I've checked 2 things: 
- we install bison from debian package repository as well (`install_debian_build_deps`), and it's installed under `/usr/bin`. I was not sure, but of course installing our patched bison under `/usr/local` is safe as it's preceeding `/usr/bin` in `PATH` env variable.
- `disable-nls` configure option is okay to be used, as "internationalization"/"native language support" is not needed (doc about nsl on [GNU bison page](https://www.gnu.org/software/bison/manual/html_node/Enabling-I18n.html))

PR #3589, opened 2021-02-25
kafka: fix missing log_msg_unref in case of failure

gaborznagy:
We increase the ref-count when we publish the message to the kafka broker, and normally decrease it in the delivery callback.
However if the publishing fails, the delivery callback is never called, though the message re-publishing will be re-tried again increasing the ref-count of the same message.

After certain number of retries an integer overflow can happen to the ref-counter, causing an abort on the next call of log_msg_ref.


kira-syslogng:
Build SUCCESS

Kokan (approved):

MrAnno (approved):

bazsi (approved):
Nice.

PR #3595, opened 2021-03-04
dbld/release: create dbld/{install,release,build} dir before docker does

Kokan:
dbld/rules in some cases may start docker before calling mkdir for the
above listed directories. This may lead to permission issues, as docker
runs as root and creates those directories with root, but the user
calling dbld/rules may not be root user, thus not able to utilize those
directories.

A possible solution would be to make sure in every execution path the
directories are created before docker run. That is full of trial and
error.

This alternative solution adds those directories to git, and as such
make sures the files exists before even the user able to run dbld/rules.

No news file, neither updating NEWS.md.

Tested in my own fork: https://github.com/Kokan/syslog-ng/runs/2030802765?check_suite_focus=true

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,9 +39,6 @@ m4/ax*.m4
 .deps
 /lib/cfg-grammar.h
 /test-driver
-dbld/build/
-dbld/release/
-dbld/install/
 dbld/rules.conf
 dbld/extra-files/
 debian/build-tree/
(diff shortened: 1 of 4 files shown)


alltilla (approved):

MrAnno (approved):

kira-syslogng:
Build SUCCESS

PR #3603, opened 2021-03-12
github-actions: remove python2 tests

alltilla:
As `python2` became EOL and as every syslog-ng packager switched to `python3` (except some old platforms in OBS) we no longer feel the need to support `python2`.

We do not plan to remove the `python2` related code yet, because it would take a lot of effort, but we can easily remove the `python2` related tests from the CI. This means with future changes we might break `python2`, and we will not see it, but we do not plan to support it anyways.

### DIFF
--- a/.github/workflows/devshell.yml
+++ b/.github/workflows/devshell.yml
@@ -15,7 +15,6 @@ jobs:
 
     strategy:
       matrix:
-        python: [python2, python3]
         build-tool: [autotools, cmake]
         cc: [gcc, clang]
       fail-fast: false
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

Kokan (commented):
Please write a news entry, as with this a direct python2 support also dropped. Even if python works for a while.

Kokan (approved):

kira-syslogng:
Build SUCCESS

MrAnno (approved):

PR #3611, opened 2021-03-16
tlscontext: eliminate TVM_NONE when using peer-verify(none)

szemere:
TVM_NONE was only used in one place, and it was used in the same
way as "optional-untrusted". So there were no need for it.
However I left the TVM_NONE as the original 0th element of the enum.
Hopefully it will reveal hidden uninitialised issues.

Note: original intention was to modify the behavior of the `tls_session_verify` function, so it will skip certificate validation in case of `peer-verify(none)`.
https://github.com/syslog-ng/syslog-ng/blob/de4997facb5a66c394f625d8562b4923c15425f3/lib/tlscontext.c#L186

### DIFF
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -424,9 +424,6 @@ tls_context_setup_verify_mode(TLSContext *self)
 
   switch (self->verify_mode)
     {
-    case TVM_NONE:
-      verify_mode = SSL_VERIFY_NONE;
-      break;
     case TVM_OPTIONAL | TVM_UNTRUSTED:
       verify_mode = SSL_VERIFY_NONE;
       break;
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build FAILURE

lib/tlscontext.c
```diff
@@ -911,7 +908,7 @@ gboolean
 tls_context_set_verify_mode_by_name(TLSContext *self, const gchar *mode_str)
 {
   if (strcasecmp(mode_str, "none") == 0)
-    self->verify_mode = TVM_NONE;
+    self->verify_mode = TVM_UNTRUSTED;
```
gaborznagy, line 911: I think you'll need `TVM_OPTIONAL | TVM_UNTRUSTED` as `TVM_UNTRUSTED` alone won't match due to the bitwise OR in the switch-case in `tls_context_setup_verify_mode`.

Alternatively, just move the `strcmp( .., "none")` line to the line.
```
else if (strcasecmp(mode_str, "optional-untrusted") == 0 || strcasecmp(mode_str, "optional_untrusted") == 0)
```

szemere: Thank You for your comment. I agree with you and I will move it to the proposed place. (Small note: the switch statement also come to my mind, but the code will not reach it.)

kira-syslogng:
Build SUCCESS

gaborznagy:
Can we remove the WIP flag?

lib/tlscontext.c
```diff
@@ -424,9 +424,6 @@ tls_context_setup_verify_mode(TLSContext *self)
 
   switch (self->verify_mode)
     {
-    case TVM_NONE:
```
gaborznagy, line 427: I've checked if there was any reason in the past to handle "none" and "optional-trusted" in separate cases, but I haven't found anything.
The original [commit](https://github.com/syslog-ng/syslog-ng/commit/0a287bd61240503a970e6431d4ebfb65d42b469d) set `SSL_VERIFY_NONE` in both cases, so we should merge the 2 options into the same case.

gaborznagy (approved):

Kokan (approved):

Kokan:
I did not merge, as I wish the https://github.com/syslog-ng/syslog-ng/pull/3616 and release process could be done before any new content added to master.

lib/tlscontext.c
szemere: Thank You @gaborznagy. I successfully overseen this part twice.

szemere:
Thank You for the review comments!
Internal tests passed, removing the WIP flag.

PR #3618, opened 2021-03-18
transport-tls: handle missing errno in case of unexpected EOF

szemere:
(Triggered by the failover mechanism.) Syslog-ng reported a very strange error
message if the other end closed the connection before sending any data first:

> I/O error occurred while reading; fd='13', **error='Success (0)'**

The root cause of the problem is that `SSL_read` handles the read()==0
as a syscall error, and such as expect it to set an errno number.

This is a known bug in OpenSSL. Unfortunately the fix got reverted because
it broke too many legacy code. Details can be found here: https://github.com/openssl/openssl/pull/11400

### DIFF
--- a/lib/transport/transport-tls.c
+++ b/lib/transport/transport-tls.c
@@ -135,8 +135,11 @@ log_transport_tls_read_method(LogTransport *s, gpointer buf, gsize buflen, LogTr
               rc = (log_transport_tls_send_shutdown(self) >= 0) ? 0 : -1;
               break;
             case SSL_ERROR_SYSCALL:
-              rc = -1;
-              /* errno is set accordingly */
+              // https://github.com/openssl/openssl/pull/11400
+              // There is a known bug in OpenSSL where it reports SSL_ERROR_SYSCALL without setting
+              // the proper errno value. The mentioned PR were reverted because lot of legacy code
+              // were broken by the fix. OpenSSL 3.0.0 will contain it.
+              rc = (errno == 0) ? 0 : -1;
               break;
             default:
               goto tls_error;
@@ -191,6 +194,16 @@ log_transport_tls_write_method(LogTransport *s, const gpointer buf, gsize buflen
           break;
         case SSL_ERROR_SYSCALL:
           /* errno is set accordingly */
+
+          // https://github.com/openssl/openssl/pull/11400
+          // There is a known bug in OpenSSL where it reports SSL_ERROR_SYSCALL without setting
+          // the proper errno value. The mentioned PR were reverted because lot of legacy code
+          // were broken by the fix. OpenSSL 3.0.0 will contain it.
+          if (errno == 0)
+            {
+              rc = -1;
+              errno = ECONNRESET;
+            }
           break;
         default:
           goto tls_error;


kira-syslogng:
Build SUCCESS

MrAnno:
Also: openssl/openssl#11400

>     Revert the change of EOF detection while reading in libssl to avoid
>     regressions in applications depending on the current way of reporting
>     the EOF. As the existing method is not fully accurate the change to
>     reporting the EOF via SSL_ERROR_SSL is kept on the current development
>     branch and will be present in the 3.0 release.

lib/transport/transport-tls.c
```diff
@@ -135,8 +135,11 @@ log_transport_tls_read_method(LogTransport *s, gpointer buf, gsize buflen, LogTr
               rc = (log_transport_tls_send_shutdown(self) >= 0) ? 0 : -1;
               break;
             case SSL_ERROR_SYSCALL:
-              rc = -1;
-              /* errno is set accordingly */
+              // https://github.com/openssl/openssl/pull/10907
```
MrAnno, line 138: I think referencing #11400 here would be more direct, because that PR contains the reasoning about the revert and proof that SSL_ERROR_SYSCALL+errno=0 indicates an unexpected EOF.

lib/transport/transport-tls.c
```diff
@@ -135,8 +135,11 @@ log_transport_tls_read_method(LogTransport *s, gpointer buf, gsize buflen, LogTr
...
+              // There is a known bug in OpenSSL where it reports SSL_ERROR_SYSCALL without setting
+              // the proper errno value. The mentioned PR were reverted because lot of legacy code
+              // were broken by the fix. OpenSSL 3.0.0 will contain it.
+              rc = (errno == 0) ? 0 : -1;
```
MrAnno, line 142: I think we should fix the same in `log_transport_tls_write_method()` as well.

lib/transport/transport-tls.c
szemere: Thank You! Also updated the PR description.

lib/transport/transport-tls.c
szemere: According to my knowledge write() could not return 0. (Thus not trigger the openssl problem.) However I really like the simmetry behind it, and I am still convincible to implement this change.

kira-syslogng:
Build SUCCESS

lib/transport/transport-tls.c
MrAnno: I think `SSL_ERROR_SYSCALL` with `errno` value of 0 is possible when calling `SSL_write()` as well.

You are right, the interface of `log_transport_write()` is based on the `write()` syscall, which does not return 0 on EOF.
In the above scenario, we could return -1 and set errno ECONNRESET or EPIPE instead.

What do you think?

kira-syslogng:
Build SUCCESS

lib/transport/transport-tls.c
```diff
@@ -190,7 +193,16 @@ log_transport_tls_write_method(LogTransport *s, const gpointer buf, gsize buflen
...
+          // https://github.com/openssl/openssl/pull/11400
+          // There is a known bug in OpenSSL where it reports SSL_ERROR_SYSCALL without setting
+          // the proper errno value. The mentioned PR were reverted because lot of legacy code
+          // were broken by the fix. OpenSSL 3.0.0 will contain it.
+          if (errno == 0)
+            {
+              rc = -1;
+              errno = ECONNRESET;
+            }
+          rc = -1;
```
MrAnno, line 205: Note: I would keep the original comment as well (`/* errno is set accordingly */`), as it is the intended behavior, which can't easily be expressed by code.

MrAnno (approved):

MrAnno (dismissed):

Kokan (approved):

Kokan (approved):

kira-syslogng:
Build SUCCESS

MrAnno (approved):
@szemere wants to run some internal TLS tests too before merging this PR.

Approve.

gaborznagy (approved):

szemere:
Internal tests are passed. KUDO for @mitzkia for the detailed analysis.

PR #3626, opened 2021-03-25
test_filters_regexp: skip a negative test in case of old libpcre

gaborznagy:
PCRE introduced a constraint in pcre 8.34 to named subpatterns: names must start with a non-digit character. [1]
This means that the following pattern is invalid: `(?P<1name>foo)`.

For the current documentation, see: http://pcre.org/original/doc/html/pcrepattern.html#SEC16

[1] https://vcs.pcre.org/pcre?view=revision&sortby=rev&revision=1394

### DIFF
--- a/lib/filter/tests/test_filters_regexp.c
+++ b/lib/filter/tests/test_filters_regexp.c
@@ -34,6 +34,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


szemere (approved):

kira-syslogng:
Build SUCCESS

Kokan (approved):

PR #3629, opened 2021-03-29
cfg: stop if there is no source configured

Kokan:
If there is no source configured in syslog-ng, there is not much it can do. Instead starting up and doing nothing, syslog-ng should not accept configuration that has no source.

This config has a source `s`, but that is not connected to any logpath, thus nothing is going to really happen.
```
@version: 3.30

source s { file("/tmp/foo"); };

log {
  destination { file("/tmp/bar"); }
};
```

Doing a detection on empty logpath (not nested ones) could be also done, and warning the user in those cases.
A related issue: https://github.com/syslog-ng/syslog-ng/issues/1161 (but this PR does not everything covered by that issue).
News file is missing.

### DIFF
--- a/lib/cfg-tree.c
+++ b/lib/cfg-tree.c
@@ -1334,6 +1334,54 @@ cfg_tree_check_inline_template(CfgTree *self, const gchar *template_or_name, GEr
 }
 
 gboolean
+_count_sources(LogExprNode *node, gpointer user_data)
+{
+  gint *counter = (gint *)user_data;
+
+  if (node->content == ENC_SOURCE)
+    ++(*counter);
+
+  return TRUE;
+}
+
+static void
+log_expr_node_foreach(LogExprNode *node, gboolean (*func)(LogExprNode *, gpointer user_data), gpointer user_data)
+{
+  g_assert(func != NULL);
+  g_assert(node != NULL);
+
+  if (!func(node, user_data))
+    return;
+
+  if (node->children)
+    log_expr_node_foreach(node->children, func, user_data);
+
+  if (node->next)
+    log_expr_node_foreach(node->next, func, user_data);
+}
+
+static void
+cfg_tree_foreach(CfgTree *self, gboolean (*func)(LogExprNode *, gpointer user_data), gpointer user_data)
+{
+  for (gint i = 0; i < self->rules->len; i++)
+    {
+      LogExprNode *rule = (LogExprNode *) g_ptr_array_index(self->rules, i);
+
+      log_expr_node_foreach(rule, func, user_data);
+    }
+}
+
+static gboolean
+cfg_tree_check_empty_pipelines(CfgTree *self)
+{
+  gint source_counter = 0;
+
+  cfg_tree_foreach(self, _count_sources, &source_counter);
+
+  return (source_counter > 0);
+}
+
+gboolean
 cfg_tree_compile(CfgTree *self)
 {
   gint i;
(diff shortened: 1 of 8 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/cfg-tree.c
```diff
@@ -1333,6 +1333,50 @@ cfg_tree_check_inline_template(CfgTree *self, const gchar *template_or_name, GEr
...
+  for (gint i = 0; i < self->rules->len; i++)
+    {
+      LogExprNode *rule = (LogExprNode *) g_ptr_array_index(self->rules, i);
+
+      cfg_tree_foreach(rule, _count_sources, &source_counter);
+      if (source_counter > 0)
+        return TRUE;
+    }
+
+  return (source_counter > 0);
```
bazsi, line 1377: This is always false as we are returning early within the loop.

lib/cfg-tree.c
```diff
@@ -1359,6 +1403,13 @@ cfg_tree_compile(CfgTree *self)
           return FALSE;
         }
     }
+
+  if (self->check_zero_source_count && !cfg_tree_check_empty_pipelines(self))
+    {
+      msg_error("There are no source configured in any of the logpaths, there is nothing for syslog-ng to do");
```
bazsi, line 1409: There **is** no source configured

Or there are no **sources**

lib/cfg-tree.c
```diff
@@ -1333,6 +1333,50 @@ cfg_tree_check_inline_template(CfgTree *self, const gchar *template_or_name, GEr
...
+  gint *counter = (gint *)user_data;
+
+  if (node->content == ENC_SOURCE)
+    ++(*counter);
+
+  return TRUE;
+}
+
+static void
+cfg_tree_foreach(LogExprNode *node, gboolean (*func)(LogExprNode *, gpointer user_data), gpointer user_data)
```
bazsi, line 1348: This should probably be a method for LogExprNode

lib/cfg-tree.c
```diff
@@ -1333,6 +1333,50 @@ cfg_tree_check_inline_template(CfgTree *self, const gchar *template_or_name, GEr
...
+  if (node->next)
+    cfg_tree_foreach(node->next, func, user_data);
+}
+
+static gboolean
+cfg_tree_check_empty_pipelines(CfgTree *self)
+{
+  gint source_counter = 0;
+
+  for (gint i = 0; i < self->rules->len; i++)
```
bazsi, line 1368: And the for loop itself is cfg_tree_foreach(), while the remaining logic is the check_empty_pipelines() function.

lib/cfg-tree.c
```diff
@@ -1359,6 +1403,13 @@ cfg_tree_compile(CfgTree *self)
           return FALSE;
         }
     }
+
+  if (self->check_zero_source_count && !cfg_tree_check_empty_pipelines(self))
```
bazsi, line 1407: I think the validation step could be a separate function that the main program calls and the test programs don't. This way test programs don't need to opt out of validation by setting a boolean to false.

I would envision similar methods, like cfg_tree_optimize() that could remove no-op logpipe instances from the graph.

bazsi (commented):
I think there are some things yet to improve, based on my comments I added to the patches.

Kokan:
Thanks for the review. This code is far from solid, hence the draft.

The reason why I included the option, that I was thinking maybe we should make it configurable just like `allow-config-dups`.
I was thinking when an empty config or source less config would make sense.

bazsi:
> Thanks for the review. This code is far from solid, hence the draft.
> 
> The reason why I included the option, that I was thinking maybe we should make it configurable just like `allow-config-dups`.
> I was thinking when an empty config or source less config would make sense.

I don't see one myself. If one comes up, we can come up with the option :)

lib/cfg-tree.c
Kokan: I'll check this.

kira-syslogng:
Build FAILURE

Kokan:
I am not sure when I'll pick it up, and it just occupies space here. Anybody who want to, feel free to use the code without asking.

Issue #3469, opened 2020-10-27
loggen should attempt reconnect when destination connections are lost


haidangwa:
## Description of the problem

We tested with an AnyCast configuration between two geographically distant locations, and when we shutdown all the network interfaces at one location, we expected all the loggen messages to be sent automatically to the second site. We confirmed that Anycast was detecting the downed interfaces and resolving all traffic to the second site; however, loggen failed to send data as soon as all services in site 1 was down. If we re-run loggen while the interfaces in site1 were still down, the traffic connected to interfaces in site2 successfully, as we originally expected.

Target destination was over UDP 514 (standard UDP syslog) and TCP 514. The result was the same whether we tested with UDP or TCP.

## Additional context

This was the loggen output. The lines showing `error` occurred as soon as we downed one of 2 interfaces in site1. When we downed the final of 2 interfaces of site1, loggen just stopped sending and never reconnected -- this is demonstrated by the rate of messages sent dropping off at the end of the output.

```
loggen -i -r 1000 -I 60 -P --read-file /tmp/vanhalen -l --active-connections 20 -F oursyslog.com 514 -d
count=36983, rate = 19177.47 msg/sec
count=46583, rate = 19195.74 msg/sec
count=56173, rate = 19176.09 msg/sec
count=65774, rate = 19197.47 msg/sec
count=75384, rate = 19216.93 msg/sec
count=84976, rate = 19180.05 msg/sec
count=94591, rate = 19225.73 msg/sec
count=104188, rate = 19190.24 msg/sec
count=113790, rate = 19200.39 msg/sec
count=123390, rate = 19196.08 msg/sec
count=132990, rate = 19195.97 msg/sec
count=142591, rate = 19198.24 msg/sec
count=152197, rate = 19207.97 msg/sec
count=161803, rate = 19207.35 msg/sec
count=171401, rate = 19192.12 msg/sec
count=181008, rate = 19209.20 msg/sec
count=190606, rate = 19191.82 msg/sec
count=200206, rate = 19196.08 msg/sec
count=209806, rate = 19196.24 msg/sec
count=219414, rate = 19212.20 msg/sec
count=229022, rate = 19212.20 msg/sec
error [socket_plugin.c:active_thread_func:393] error sending buffer on 35 (rc=-1)
error [socket_plugin.c:active_thread_func:393] error sending buffer on 30 (rc=-1)
error [socket_plugin.c:active_thread_func:393] error sending buffer on 29 (rc=-1)
[...]

PR #3630, opened 2021-03-30
Fix issue #3469: add reconnect opt in loggen

LittleFish33:
## Purpose / Description

Fix [#3469](https://github.com/syslog-ng/syslog-ng/issues/3469) loggen should attempt reconnect when destination connections are lost

When destination connections are lost, loggen just stopped sending and never reconnected.

## Approach

Add a new opt in loggen_options,

```
{ "reconnect", 0, 0, G_OPTION_ARG_NONE, &global_plugin_option.reconnect, "Attempt to reconnect when destination connections are lost", NULL},
```

and if encounter connection_error and option->reconnect is true, try to reconnect.

### DIFF
--- a/tests/loggen/loggen.c
+++ b/tests/loggen/loggen.c
@@ -57,6 +57,7 @@ static PluginOption global_plugin_option =
   .target = NULL,
   .port = NULL,
   .rate = 1000,
+  .reconnect = 0,
 };
 
 static char *sdata_value = NULL;
@@ -99,6 +100,7 @@ static GOptionEntry loggen_options[] =
   { "number", 'n', 0, G_OPTION_ARG_INT, &global_plugin_option.number_of_messages, "Number of messages to generate", "<number>" },
   { "quiet", 'Q', 0, G_OPTION_ARG_NONE, &quiet, "Don't print the msg/sec data", NULL },
   { "debug", 0, 0, G_OPTION_ARG_NONE, &debug, "Enable loggen debug messages", NULL },
+  { "reconnect", 0, 0, G_OPTION_ARG_NONE, &global_plugin_option.reconnect, "Attempt to reconnect when destination connections are lost", NULL},
   { NULL }
 };
 
--- a/tests/loggen/loggen_plugin.h
+++ b/tests/loggen/loggen_plugin.h
@@ -45,6 +45,7 @@ typedef struct _plugin_option
   const char *target; /* command line argument */
   const char *port;
   int  rate;
+  int reconnect;
 } PluginOption;
 
 typedef struct _thread_data
--- a/tests/loggen/socket_plugin/socket_plugin.c
+++ b/tests/loggen/socket_plugin/socket_plugin.c
@@ -390,10 +390,13 @@ active_thread_func(gpointer user_data)
 
       connection_error = send_msg(fd, message, str_len);
 
-      thread_context->sent_messages++;
-      thread_context->buckets--;
+      if(!connection_error)
+        {
+          thread_context->sent_messages++;
+          thread_context->buckets--;
+        }
 
-      if(connection_error && option->reconnect)
+      if(connection_error && option->reconnect && thread_run)
         {
           shutdown(fd, SHUT_RDWR);
           close(fd);
--- a/tests/loggen/socket_plugin/socket_plugin.c
+++ b/tests/loggen/socket_plugin/socket_plugin.c
@@ -392,6 +392,35 @@ active_thread_func(gpointer user_data)
 
       thread_context->sent_messages++;
       thread_context->buckets--;
+
+      if(connection_error && option->reconnect)
+        {
+          shutdown(fd, SHUT_RDWR);
+          close(fd);
+
+          ERROR("destination connection %s:%s (%p) is lost, try to reconnect\n", option->target, option->port, g_thread_self());
+          if (unix_socket_x)
+            fd = connect_unix_domain_socket(sock_type, option->target);
+          else
+            fd = connect_ip_socket(sock_type, option->target, option->port, option->use_ipv6);
+
+          while(fd < 0 && !thread_check_exit_criteria(thread_context))
+            {
+              ERROR("can not reconnect to %s:%s (%p), try again after %d sec\n", option->target, option->port, g_thread_self(), 1);
+              g_usleep(1e6);
+
+              if (unix_socket_x)
+                fd = connect_unix_domain_socket(sock_type, option->target);
+              else
+                fd = connect_ip_socket(sock_type, option->target, option->port, option->use_ipv6);
+            }
+
+          if(fd > 0)
+            {
+              DEBUG("(%d) reconnected to server on socket %d (%p)\n", thread_context->index, fd, g_thread_self());
+              connection_error = FALSE;
+            }
+        }
     }
   DEBUG("thread (%s,%p) finished\n", socket_loggen_plugin_info.name, g_thread_self());
 
--- a/tests/loggen/ssl_plugin/ssl_plugin.c
+++ b/tests/loggen/ssl_plugin/ssl_plugin.c
@@ -382,6 +382,32 @@ active_thread_func(gpointer user_data)
 
       thread_context->sent_messages++;
       thread_context->buckets--;
+
+      if(connection_error && option->reconnect)
+        {
+          close_ssl_connection(ssl);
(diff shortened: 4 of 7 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

Kokan (commented):
It would be nice if you could write a functional test for loggen new reconnect option. You can find functional tests under `tests/python_functional/`. If you need help for that feel free to ask here.

LittleFish33:
@Kokan 
Hi, I have read the code in `tests/python_functional/` and tried to write a functional test. 

First, I have revised the code of `sent_message` and `buckets`:
```c++
if(!connection_error)
{
    thread_context->sent_messages++;
    thread_context->buckets--;
}
```

And then I have added a functional_test called `test_pp_with_reconnenct`:
```python
NUMBER_OF_MESSAGES = 5

def test_pp_with_reconnenct(config, port_allocator, syslog_ng, loggen):
    network_source = config.create_network_source(ip="localhost", port=port_allocator())
    file_destination = config.create_file_destination(file_name="output.log")
    config.create_logpath(statements=[network_source, file_destination])

    syslog_ng.start(config)

    loggen.start(network_source.options["ip"], network_source.options["port"], inet=True, rate=1, 
                active_connections=1, reconnect=True, interval=30, number=NUMBER_OF_MESSAGES)
    
    syslog_ng.stop()
    time.sleep(1)
    syslog_ng.start(config)

    time.sleep(6)

    logs = file_destination.read_logs(NUMBER_OF_MESSAGES)

    assert len(logs) == NUMBER_OF_MESSAGES
```

If the reconnect is False, `len(logs)` equals to 1, as we expected.
But if the reconnect is True, `len(logs)` always equals to `NUMBER_OF_MESSAGES-1`. I have tried to change the rate of loggen, but the results are still `NUMBER_OF_MESSAGES-1`.

To address this, I checked the report generated by `make pytest-check`. I found that in the `syslog_ng_server_stderr` file, the summary of content is
```
initialization
receive message
initialization
receive message x 3
```
But in the `loggen_stderr_6` file, the content is:
```
successfully send message x 2
error sending buffer
successfully send message x 3
```

The last message before the server is shut down is always not received by `syslog-ng`, but `loggen` doesn't know about it.

I think this is because calling the `send` function of socket does not mean that data is immediately transferred to the target host. Instead, it is responsible for writing data to the output buffer, and the data is sent from the output buffer to the target host by the underlying communication protocols. Therefore, when `loggen` found out that the `syslog-ng` is stopped, the message in the buffer will be discarded, result in `len(logs)=NUMBER_OF_MESSAGES-1`.

I'm not sure if there's a problem with my understanding, any help will be appreciated.

> It would be nice if you could write a functional test for loggen new reconnect option. You can find functional tests under `tests/python_functional/`. If you need help for that feel free to ask here.

Kokan:
Sorry for the long delay.
No, I think you got it right regarding the buffers. The fact that the other side closed the socket could be detected via reading it or using the `SIGPIPE`. In the end message loss are a possibility, there are application level logics for it to prevent. So I think it is goodd as is.

kira-syslogng:
Build SUCCESS

LittleFish33:
Thanks, I have updated the commits

Kokan:
Please remove the merge commit from your branch: https://github.com/syslog-ng/syslog-ng/pull/3630/commits/1b2851d0639c8660bd4057f371d85580ea954725 (if you want to update your branch use `git rebase origin/master` instead).
Please rewrite the original commit with the fixups you did in the second one.
So the first should contain changes in loggen, and the second the changes regarding new testcase and test framework.

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/source_drivers/network_source/proxyprotocol/test_pp_with_reconnect.py
```diff
@@ -0,0 +1,49 @@
...
+
+    syslog_ng.start(config)
+
+    loggen.start(
+        network_source.options["ip"], network_source.options["port"], inet=True,
+        rate=1, active_connections=1, reconnect=True, interval=30, number=NUMBER_OF_MESSAGES,
+    )
+
+    syslog_ng.stop()
+    time.sleep(1)
```
mitzkia, line 41: Please remove this time.sleep(1) and time.sleep(5) expressions from the testcase.
Testcase will pass without them also. Unfortunately It will highly slow down the full functional testcase run.

tests/python_functional/functional_tests/source_drivers/network_source/proxyprotocol/test_pp_with_reconnect.py
```diff
@@ -0,0 +1,49 @@
...
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+import time
+
+NUMBER_OF_MESSAGES = 5
```
mitzkia, line 25: Please reduce the this NUMBER_OF_MESSAGES counter as low as can it be.
As this value is higher the full functional testcase run will be slower.

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/source_drivers/network_source/proxyprotocol/test_pp_with_reconnect.py
mitzkia: Thanks for the fix.

kira-syslogng:
Build SUCCESS

Kokan (dismissed):

tests/python_functional/functional_tests/source_drivers/network_source/proxyprotocol/test_pp_with_reconnect.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+
+    syslog_ng.stop()
+    syslog_ng.start(config)
+
+    logs = file_destination.read_logs(counter=NUMBER_OF_MESSAGES - 1)
+
+    # the send function of socket only write data to the output buffer, if the remote host is lost
+    # the message in the buffer will be discarded, result in len(logs)=NUMBER_OF_MESSAGES-1
+    assert len(logs) == NUMBER_OF_MESSAGES - 1
```
alltilla, line 46: I really appreciate, that you have written a `Light` test for this! Thanks!

However, this test does not use proxy-protocol, so this folder is not right for it. Thinking about it, it tests a `loggen` command, so technically a self-test should be written for it, but there is no place for that right now.
I think the best would be to move this to `functional_tests/source_drivers/network_source/` simply, then, when we make the self-tests for the `helper` classes, we can move it there, probably without using `syslog-ng`, not sure yet.

@mitzkia What do you think?

mitzkia: Yes I also want to thank for updating Light (this is how we called framework under: `tests/python_functional/`) and adding new functional testcase.
Thanks for the comment, partially I am agreed with you. 
Basically this is a new loggen functional testcase for reconnect argument.
I would place it under `functional_tests/tools/loggen/test_loggen_with_reconnect.py`

And just for the note, at the end please add new testcase file to the list of `tests/python_functional/functional_tests/Makefile.am`

LittleFish33: Thanks for your comments, I would move it to `functional_tests/tools/loggen/test_loggen_with_reconnect.py`.

alltilla (approved):

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/source_drivers/network_source/proxyprotocol/test_pp_with_reconnect.py
alltilla: Thanks!

Kokan (approved):

haidangwa:
Would it be possible to add official documentation for this feature option?

PR #3631, opened 2021-03-30
WIP: logthreaded_dest_driver_init_method: add persist_name as a parameter

szemere:
Continuation of #3559.

We moved the log_threaded_dest_driver_init_method in each ancesstor
to a place where it is safe to call the generate_persist_name virtual
function. But the code failed to express the reason behind it.

This PR adds the persist_name as a parameter of the public function
to make that dependency between those methods explicit.

To avoid any mismatch between the provided parameter and the result of
later generate_persist_name function calls, the LogThreadedDestDriver
stores the provided value.

Note: To preserve the LogPipe interface, it is still possible to call
the original init_method function. (This is the case when an ancestor
doesn't override it.) In that case the LogThreadedDestDriver will
generate the name internally using the generate_persist_name function.

### DIFF
--- a/lib/logthrdest/logthrdestdrv.c
+++ b/lib/logthrdest/logthrdestdrv.c
@@ -107,19 +107,15 @@ log_threaded_dest_worker_rewind_messages(LogThreadedDestWorker *self, gint batch
 static gchar *
 _format_queue_persist_name(LogThreadedDestWorker *self)
 {
-  LogPipe *owner = &self->owner->super.super.super;
-
   if (self->worker_index == 0)
     {
       /* the first worker uses the legacy persist name, e.g.  to be able to
        * recover the queue previously used.  */
-      return g_strdup(log_pipe_get_persist_name(owner));
+      return g_strdup(log_pipe_get_persist_name((LogPipe *)self->owner));
     }
   else
     {
-      return g_strdup_printf("%s.%d.queue",
-                             log_pipe_get_persist_name(owner),
-                             self->worker_index);
+      return g_strdup_printf("%s.%d.queue", log_pipe_get_persist_name((LogPipe *)self->owner), self->worker_index);
     }
 }
 
@@ -772,9 +768,9 @@ ok:
 static gboolean
 _acquire_worker_queue(LogThreadedDestWorker *self)
 {
-  gchar *persist_name = _format_queue_persist_name(self);
-  self->queue = log_dest_driver_acquire_queue(&self->owner->super, persist_name);
-  g_free(persist_name);
+  gchar *queue_persist_name = _format_queue_persist_name(self);
+  self->queue = log_dest_driver_acquire_queue(&self->owner->super, queue_persist_name);
+  g_free(queue_persist_name);
 
   if (!self->queue)
     return FALSE;
@@ -1031,12 +1027,10 @@ _unregister_stats(LogThreadedDestDriver *self)
 static gchar *
 _format_seqnum_persist_name(LogThreadedDestDriver *self)
 {
-  static gchar persist_name[256];
+  static gchar seqnum_persist_name[256];
 
-  g_snprintf(persist_name, sizeof(persist_name), "%s.seqnum",
-             self->super.super.super.generate_persist_name((const LogPipe *)self));
-
-  return persist_name;
+  g_snprintf(seqnum_persist_name, sizeof(seqnum_persist_name), "%s.seqnum", log_pipe_get_persist_name((LogPipe *)self));
+  return seqnum_persist_name;
 }
 
 static gboolean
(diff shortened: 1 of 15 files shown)


kira-syslogng:
Build SUCCESS

lib/logthrdest/tests/test_logthrdestdrv.c
```diff
@@ -771,6 +775,25 @@ Test(logthrdestdrv, test_explicit_ack_accept)
   cr_assert(dd->super.shared_seq_num == 11, "%d", dd->super.shared_seq_num);
 }
 
+Test(logthrdestdrv, test_if_generate_persist_name_called_via_the_logpipe_interface)
+{
+  cr_assert(dd->generate_persist_name_call_counter == 1,
```
Kokan, line 780: The test should not limit a code to call generate persist name only once. As I think the important thing is that it is called at least once. The fact that it is cached, and only called once is an implementation detail.
On the other hand, I do not see a test that checks regarding the persist-name actually correctly set.

lib/logthrdest/logthrdestdrv.c
```diff
@@ -107,19 +107,15 @@ log_threaded_dest_worker_rewind_messages(LogThreadedDestWorker *self, gint batch
...
 _format_queue_persist_name(LogThreadedDestWorker *self)
 {
-  LogPipe *owner = &self->owner->super.super.super;
-
   if (self->worker_index == 0)
     {
       /* the first worker uses the legacy persist name, e.g.  to be able to
        * recover the queue previously used.  */
-      return g_strdup(log_pipe_get_persist_name(owner));
+      return g_strdup(self->owner->persist_name);
```
Kokan, line 114: Not particularity this line, but in general this code violates the contract of `log_pipe_get_persist_name` gives me the persist-name. It would be really hard for me to remember that I no longer can use that in case of `LogThreadedDest...`.

Instead, I would keep the new init as you did.
Add a new general `generate_persist_name` to `LogThreadedDestDriver`
```c
static const gchar *
_persist_name(const LogPipe *s)
{
  LogThreadedDestDriver *self = (LogThreadedDestDriver*)s;
  g_assert(self->persist_name);
  return self->persist_name;
}
```
That simply always returns with the cache value (it is cached in the new init anyway), and simply do not override the `generate_persist_name` in the classes inherited from this one. With that here you do not need to change a thing.

szemere: Thank You for the review! Based on your observations, I asked myself again: Why am I doing this?

> I try to force the generate_persist_name virtual function in the inherited classes.

To be more precise: I try to force the case, when it is safe to call it. (If we call it before certain init steps, it could end up in a null pointer dereference.)

My observations:

- Providing a LogThreadedDestDriver implementation of the generate_persist_name function do not prevent overriding it in the inherited classes. So any safeguard in that function will be futile.
- Overriding the generate_persist_name function is not a bad thing. I can still safely call the log_pipe_get_persist_name with it. Which is much more better than my current implementation.
- I only started to store the provided value to avoid any accidental mismatch between the init/deinit phases. i.e. the drivers usually generate the persist name from connection details (IP, Port, etc..), which might change during the init step.
  

It might be possible to force this stability without breaking the log_pipe_get_persist_name contract. proposal:
 - Keep the new init method, and store the provided value.
 - Do not access the stored value in the code directly, keep the log_pipe_get_persist_name function. (Which implicitly means, we will continue to use the generate_persist_name functions of the inherited classes.)
 - Compare the output of the log_pipe_get_persist_name function with the stored value in the key stages (init_method, on_config_inited, deinit) as a safeguard around the current/future implementations. That will pop out any programming error with the first test runs.

szemere: Discussed with @Kokan IRL. As it turned out my description was not clear enough. Since the earlier code had the highlighted problems, there were no reason keeping it. I updated the pull request with the proposal, and marking it as WIP while we discuss it.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/logthrdest/logthrdestdrv.c
```diff
@@ -1058,12 +1052,23 @@ _create_workers(LogThreadedDestDriver *self)
...
+{
+  // Make sure that generate_persist_name virtual function still returns
+  // the same value as provided during the init phase. Because several
+  // functionality of the LogThreadedDestDriver depends on it.
+  g_assert_cmpstr(self->persist_name_for_assertion, ==, log_pipe_get_persist_name((LogPipe *)self));
+}
+
 gboolean
-log_threaded_dest_driver_init_method(LogPipe *s)
+log_threaded_dest_driver_init_method(LogPipe *s, const gchar *persist_name)
```
bazsi, line 1065: honestly, I don't really like that we are changing the signature of this function, this is just the implementation of the log_pipe_init() interface, which a derived function needs to call, so in a way, it is the equivalent of calling the super()'s same-named method in Python. I can see that you have added a wrapper to make this compatible with log_pipe_init()

I understand that you want to be explicit about the dependency, but if that's the case, then it might make sense to make log_pipe_init of LogThreadedDestDriver non-overridable (as if you declared it private in C++) and create a new layer for the initialization of threaded dest drivers, which derived classes can hook into.

How that new layer (method) could be named, I have no clue at this point, but then it's only a naming issue :)

bazsi (changes requested):

alltilla:
I can think about this problem from another perspective.

A `generate_persist_name()` implementation can depend on values, that are initialized in a class's init method.
I, personally, never thought about checking those variables before using them, because naturally I thought, that those are inited at the point, where this function is called. If I understand right, the original problem had the same presumption.

It seems like, that making a dependency between `init()` and `generate_persist_name()` is complex. What if we communicate the developer, that the `generate_persist_name()` function might be called with uninitialized variables?
Make `generate_persist_name()` return a `bool` additionally to the string. Rewrite every `generate_persist_name()` function with this in mind, check every variable which is used, and return with `FALSE`, if something is not looking right. Then, instead of calling `self->super.generate_persist_name()`, make a `log_pipe_generate_persist_name()` function, which checks, if the function pointer is set, then it calls the implementation, and abort (because it is a coding error), if it fails, or return the persist string, if it succeeded.

If we do this, the developer of a new driver will look at other implementations and naturally think about `generate_persist_name()` as a function, which needs pre-checks, and which can fail. If they see, that during a test the `log_pipe_generate_persist_name()` aborted, they will have the chance to see the backtrace, and check why were not the variables initialized at that point. In a situation like the original problem, it will be seen, that it was called from `log_threaded_dest_driver_init_method()`, which was called in the class's `init()` before the necessary variables were initialized.

I know that this is not a compile time restriction, but it can reliably show us the problem, and the problem (most probalby) can be quickly fixed.

What do you think?

PR #3634, opened 2021-04-01
new parser: smart tagger

Kokan:
Parsing messages is a daily struggle when the standards are not followed. It is hard to follow up on every new source in order to provide something that makes sense. Also some part of the messages are hard to make rules to parse.

This PR introduces a parser that is based on AI text processing, it tries to parse the message and create tags with an AI.
Example configuration:
```
parser { ai-tagging( api_key("MANDATORY") ); }
```
As the new `ai-tagging` uses https://deepai.org/, the user should obtain their own api key, which you can do by registering via the dashboard.  https://deepai.org/dashboard/profile .

The parser by default checks the `$MESSAGE` macro and fills the tags into the `.ai.tags` nv entry.

### DIFF
--- a/scl/CMakeLists.txt
+++ b/scl/CMakeLists.txt
@@ -1,5 +1,6 @@
 set(SCL_DIRS
     apache
+    ai_tagging
     cee
     cim
     checkpoint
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

PR #3644, opened 2021-04-14
cfg-lex: lookup already defined arguments in block argument definitions

lbudai:
notes:
 *  it might not the best solution  - but short and solves my problem... it's just a proposal that can be discussed
 *  nothing depends on this PR

example:
```
block destination test_http (
  body-parts("host=$HOST severity=$LEVEL")
  body("$(format-json --scope none `body-parts`)")
  ...)
{
          http(
            method("POST")
            url("http://localhost:8888")
            body(`body`)
            `__VARARGS__`
            );
};
```

Motivation:

when we have an existing SCL, like this:

```
block destination test_http (
  body-parts("host=$HOST severity=$LEVEL")
  ...)
{
          http(
            method("POST")
            url("http://localhost:8888")
            body("$(format-json --scope none `body-parts`)")
            `__VARARGS__`
            );
};
```

and `body-parts` cannot express what we really need, we have to create
a new SCL. With resolving already defined block arguments we can avoid unnecessary
copy-pastes and we can also keep backward compatibility...
On the other hand - this works only for already defined arguments. So
it won't work if the order of `body-parts` and `body` are replaced.

### DIFF
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -40,7 +40,7 @@ yy_input_run_backtick_substitution(CfgLexer *self, gchar *buf, gsize buf_size, g
   GError *error = NULL;
   CFG_LTYPE *cur_lloc = &self->include_stack[self->include_depth].lloc;
 
-  res = cfg_lexer_subst_args_in_input(self->cfg->globals, NULL, NULL, buf, -1, len, &error);
+  res = cfg_lexer_subst_args_in_input(self->cfg->globals, last_block_args ?: NULL, NULL, buf, -1, len, &error);
   if (!res)
     {
       msg_error("Error performing backtick substitution in configuration file",


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
I've created a config to test this out, where in the `test` rewrite block the `E()` means expected, that is a string I think should be computed. The case where you do not change the `A()` (or in your example the `body()`), but want to change the `sub-A()` (your example the `body-parts()`) does not effect the `test` rewrite block, as it is already substituted during config (see the preprocessed result of the config).

```syslog-ng
@version: 3.30

block rewrite test (
  E()
  sub-A("sub-A")
  A("[`sub-A`]")
  ...)
{
  set("`A`" value("A") `__VARARGS__`);
  set("`E`" value("E"));
};

source s { example-msg-generator(num(1)); };
destination d { file("/dev/stdout" template("A=${A} should equal to E=${E}\n")); };

log { source(s); rewrite { test(E("[sub-A]")); }; destination(d); }; # ok
log { source(s); rewrite { test(E("B") A("B")); }; destination(d); }; # ok
log { source(s); rewrite { test(E("[sub-B]") sub-A("sub-B")); }; destination(d); }; # nok
log { source(s); rewrite { test(E("C") sub-A("sub-C") A("C")); }; destination(d); }; # ok
log { source(s); rewrite { test(E("D") A("D") sub-A("sub-D")); }; destination(d); }; # ok
```

lbudai:
@Kokan : so in shortly the problem is that a block arg is substituted too early...

lbudai:
I have a 'half-working' branch, but well... that would be too much change (handling double backticks, etc...), it simply doesn't worth it...

lbudai:
Finally, I found a different solution - it's still a short patch and hopefully not too ugly... 
The idea is the following:
* postpone backtick substitution for block arg defs
* identify postponed cases and do the final backtick substitution

I used the config below (this is a sligthly modified version of the cfg attached by @Kokan, I've added double backticks).

```
@version: 3.30

block rewrite test (
  E()
  sub-A("sub-A")
  A("[`sub-A` ``sub-A`` ``sub-X``]")
  ...)
{
  set("`A`" value("A") `__VARARGS__`);
  set("`E`" value("E"));
};

source s { example-msg-generator(num(1)); };
destination d { file("/dev/stdout" template("A=${A} should equal to E=${E}\n")); };

log { source(s); rewrite { test(E("[sub-A ``sub-A`` ``sub-X``]")); }; destination(d); }; # ok
log { source(s); rewrite { test(E("B") A("B")); }; destination(d); }; # ok
log { source(s); rewrite { test(E("[sub-B ``sub-A`` ``sub-X``]") sub-A("sub-B")); }; destination(d); }; # ok
log { source(s); rewrite { test(E("C") sub-A("sub-C") A("C")); }; destination(d); }; # ok
log { source(s); rewrite { test(E("D") A("D") sub-A("sub-D")); }; destination(d); }; # ok
```

PR #3649, opened 2021-04-15
configure.ac: add option --enable-thread-tls to manage thread ssl sup…

yifan19:
…port

The thread local storage caused arm-gcc broken while compiling                          │
syslog-ng with option '-g -O'.                                                          │
...                                                                                     │
dnscache.s: Assembler messages:                                                         │
dnscache.s:100: Error: invalid operands (.text and *UND* sections) for `-'              │
...                                                                                     │
                                                                                        │
Add option --enable-thread-tls to manage the including of thread
local storage, so we could explicitly disable it.

Signed-off-by: Hongxu Jia <hongxu.jia@windriver.com>

change default to 'yes'

Signed-off-by: Yi Fan Yu <yifan.yu@windriver.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -210,6 +210,8 @@ AC_ARG_WITH(sanitizer,
               [  --with-sanitizer=[address/undefined/etc...]
                                          Enables compiler sanitizer supports (default: no)]
               ,,with_sanitizer="no")
+AC_ARG_ENABLE(thread-tls,
+              [  --enable-thread-tls        Enable Thread Local Storage support (default: yes)],,enable_thread_tls="yes")
 
 AC_ARG_ENABLE(dynamic-linking,
               [  --enable-dynamic-linking        Link everything dynamically.],,enable_dynamic_linking="auto")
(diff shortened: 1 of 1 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

yifan19:
I can't reproduce the reported failure for arm32bit (arm7), this local patch was part of meta-oe
(yocto project) for around 2 years, I 'll drop it.

Maintainers, do you think this is any use?

Kokan:
If there is a user of such option I think okay to keep, but as you stated it works (so no user). I do not see much value of keeping it.
(Also as that is a detection, probably that should be improved if fails to work as intended in some cases.)

yifan19:
> If there is a user of such option I think okay to keep, but as you stated it works (so no user). I do not see much value of keeping it.
> (Also as that is a detection, probably that should be improved if fails to work as intended in some cases.)

ok... closing it for now.

PR #3655, opened 2021-04-20
qdisk: modify debug messages during loading diskqueue

szemere:
Since this PR is only a cosmetic change mainly focusing on working with the `dqtool info` command. I am marking it as a Draft, and I am open for second opinions.

---

Sometimes GitHub mixes the lines for me in the diff. What happened here:
 - On info level
 Print the filename, and the number of messages in the file.
 - On debug level:
 Pint the filename, and the raw header positions / counters, so one can calculate the necessary information.

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -567,6 +567,22 @@ _load_non_reliable_queues(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *q
 
 #define _clear(obj) memset(&obj, 0, sizeof(obj));
 
+static gint64
+_number_of_messages(QDisk *self)
+{
+  if (self->options->reliable)
+    {
+      return self->hdr->length + self->hdr->backlog_len;
+    }
+  else
+    {
+      return self->hdr->length +
+             self->hdr->qbacklog_pos.count +
+             self->hdr->qout_pos.count +
+             self->hdr->qoverflow_pos.count;
+    }
+}
+
 static void
 _reset_queue_pointers(QDisk *self)
 {
@@ -618,10 +634,16 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
 
       msg_info("Disk-buffer state loaded",
                evt_tag_str("filename", self->filename),
-               evt_tag_long("qout_length", self->hdr->qout_pos.count),
-               evt_tag_long("qbacklog_length", self->hdr->qbacklog_pos.count),
-               evt_tag_long("qoverflow_length", self->hdr->qoverflow_pos.count),
-               evt_tag_long("qdisk_length", self->hdr->length));
+               evt_tag_long("number_of_messages", _number_of_messages(self)));
+
+      msg_debug("Disk-buffer internal state",
+                evt_tag_str("filename", self->filename),
+                evt_tag_long("qout_length", self->hdr->qout_pos.count),
+                evt_tag_long("qbacklog_length", self->hdr->qbacklog_pos.count),
+                evt_tag_long("qoverflow_length", self->hdr->qoverflow_pos.count),
+                evt_tag_long("qdisk_length", self->hdr->length),
+                evt_tag_long("read_head", self->hdr->read_head),
+                evt_tag_long("write_head", self->hdr->write_head));
 
       _reset_queue_pointers(self);
     }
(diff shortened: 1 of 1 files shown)


modules/diskq/qdisk.c
```diff
@@ -567,6 +567,22 @@ _load_non_reliable_queues(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *q
 
 #define _clear(obj) memset(&obj, 0, sizeof(obj));
 
+static gint64
+_number_of_messages(QDisk *self)
```
Kokan, line 571: This should be kept as an implementation details of each type of diskq. (virtual function)

Kokan (commented):
I have only comment on implementation, the log change looks ok to me.

kira-syslogng:
Build FAILURE

modules/diskq/qdisk.c
gaborznagy: Just an informal comment before any virtual function is implemented (I still need to check the entire disk-buffer code): there is a `qdisk_get_length` function which only deals with `QDisk->hdr->length` member.
I would check that one before.

I just wonder if we have write-, read- and backlog_head why can't the `hdr->length` show the actual number of messages... This is a question to myself though...

szemere: To use the already existing `LogQueueDisk->get_length` virtual function, we should bring this debug message back into `LogQueueDisk->load_queue`. Which would be far more superior than double implementing the get length logic :(

Unfortunately this change: https://github.com/syslog-ng/syslog-ng/commit/a6ed296b0b63fcd8a456c9f13c3a5e8aac07b4ce widens the scope of such a refactor. Should I move that fix too, or leave the current version?

My thinking: where should the fix be? In the `qdisk.c`, because it is file handling related. Or in the `logqueue-disk-non-reliable.c`, because it is related to handling non reliable diskq's.

modules/diskq/qdisk.c
```diff
@@ -618,10 +634,16 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
 
       msg_info("Disk-buffer state loaded",
                evt_tag_str("filename", self->filename),
-               evt_tag_long("qout_length", self->hdr->qout_pos.count),
-               evt_tag_long("qbacklog_length", self->hdr->qbacklog_pos.count),
-               evt_tag_long("qoverflow_length", self->hdr->qoverflow_pos.count),
-               evt_tag_long("qdisk_length", self->hdr->length));
+               evt_tag_long("number_of_messges", _number_of_messages(self)));
+
+      msg_debug("Disk-buffer state loaded",
```
bazsi, line 639: could you use a different "message" than the one on msg_info() printed just before this one?

modules/diskq/qdisk.c
szemere: I do not recommend moving the `_reset_queue_pointers` function call from it's current location, because:

- We have to change the publicity of the `_reset_queue_pointers` function or the `QDiskFileHeader` struct.
- I would rather keep the details of header handling inside qdisk.c and do not spread this implementation detail: https://github.com/syslog-ng/syslog-ng/blob/master/modules/diskq/qdisk.c#L862

modules/diskq/qdisk.c
szemere: Suggestion:

info: "Diisk-buffer state loaded"
debug: "Disk-buffer internal state"

kira-syslogng:
Build FAILURE

gaborznagy:
Thanks for fixing my findings. Once Kira is green, this can go in.

szemere:
@kira-syslogng test this please test branch=modify_diskq_messages;

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -618,10 +634,16 @@ _load_state(QDisk *self, GQueue *qout, GQueue *qbacklog, GQueue *qoverflow)
...
                evt_tag_str("filename", self->filename),
-               evt_tag_long("qout_length", self->hdr->qout_pos.count),
-               evt_tag_long("qbacklog_length", self->hdr->qbacklog_pos.count),
-               evt_tag_long("qoverflow_length", self->hdr->qoverflow_pos.count),
-               evt_tag_long("qdisk_length", self->hdr->length));
+               evt_tag_long("number_of_messages", _number_of_messages(self)));
+
+      msg_debug("Disk-buffer internal state",
+                evt_tag_str("filename", self->filename),
+                evt_tag_long("#qout_length", self->hdr->qout_pos.count),
```
mitzkia, line 641: I would remove the # prefix from all cases. This can confuse users when they found this debug logs.
I understand that this prefix could mean that the value is something which is countable and not a digital unit.

If hard to make a different between which value is a counter or digital unit I would prefer a new naming convention in syslog-ng's debug messages. Of course it would not be a good choice if we know that users are already used this fields as a metric name in their scripts.

szemere:
@kira-syslogng test this please test branch=modify_diskq_messages;

Rebased on master

Kokan (approved):

alltilla (approved):

szemere:
@kira-syslogng test this please test branch=modify_diskq_messages;

szemere:
@kira-syslogng test this please test branch=modify_diskq_messages;

kira-syslogng:
Build SUCCESS

Kokan (approved):

alltilla (approved):

PR #3656, opened 2021-04-20
Postpone backtick substitution for block args

lbudai:
### Motivation:

When we have an existing SCL, like this:
    
```
block destination test_http (
    body-parts("host=$HOST severity=$LEVEL")
      ...)
    {
              http(
                method("POST")
                url("http://localhost:8888")
                body("$(format-json --scope none `body-parts`)")
                `__VARARGS__`
                );
    };

```    
 and `body-parts` cannot express what we really need, we have to add a 
 a new SCL block. With resolving already defined block arguments we can avoid unnecessary
 copy-pastes and we can also keep backward compatibility.

With this patch we can write this:

```
    block destination test_http (
      body-parts("host=$HOST severity=$LEVEL")
      body("$(format-json --scope none `body-parts`)")
      ...)
    {
              http(
                method("POST")
                url("http://localhost:8888")
                body(`body`)
                `__VARARGS__`
                );
    };

```

(I had a previous PR, but that did not work as expected. Finally, I found a different solution - it's still a short patch and hopefully not too ugly... )

### Idea

* postpone backtick substitution for block arg defs
* identify postponed cases and do the final backtick substitution

I used the config below (this is a sligthly modified version of the cfg attached by @Kokan).

```
@version: 3.30

block rewrite test (
  E()
  sub-A("sub-A")
  A("[`sub-A` ``sub-A`` ``sub-X``]")
  ...)
{
  set("`A`" value("A") `__VARARGS__`);
  set("`E`" value("E"));
};

source s { example-msg-generator(num(1)); };
destination d { file("/dev/stdout" template("A=${A} should equal to E=${E}\n")); };

log { source(s); rewrite { test(E("[sub-A ``sub-A`` ``sub-X``]")); }; destination(d); }; # ok
log { source(s); rewrite { test(E("B") A("B")); }; destination(d); }; # ok
log { source(s); rewrite { test(E("[sub-B ``sub-A`` ``sub-X``]") sub-A("sub-B")); }; destination(d); }; # nok
log { source(s); rewrite { test(E("C") sub-A("sub-C") A("C")); }; destination(d); }; # ok
log { source(s); rewrite { test(E("D") A("D") sub-A("sub-D")); }; destination(d); }; # ok

```

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -851,7 +851,7 @@ block_stmt
         : KW_BLOCK
           { cfg_lexer_push_context(lexer, LL_CONTEXT_BLOCK_DEF, block_def_keywords, "block definition"); }
           LL_IDENTIFIER LL_IDENTIFIER
-          '(' { last_block_args = cfg_args_new(); } block_definition ')'
+          '(' { cfg_lexer_postpone_backtick_substition(lexer); last_block_args = cfg_args_new(); } block_definition ')'
           { cfg_lexer_push_context(lexer, LL_CONTEXT_BLOCK_CONTENT, NULL, "block content"); }
           LL_BLOCK
           {
@@ -871,6 +871,7 @@ block_stmt
             free($4);
             free($10);
             last_block_args = NULL;
+            cfg_lexer_enable_immediate_backtick_substition(lexer);
           }
         ;
 
@@ -892,7 +893,11 @@ block_arg
           LL_BLOCK
           {
             cfg_lexer_pop_context(lexer);
-            cfg_args_set(last_block_args, $1, $3); free($1); free($3);
+            if ($3 && cfg_lexer_is_backtick_substitution_postponed(lexer) && strchr($3, '`'))
+              cfg_args_set(last_block_args, $3, $3);
+            cfg_args_set(last_block_args, $1, $3);
+            free($1);
+            free($3);
           }
         ;
 
--- a/lib/cfg-lex.l
+++ b/lib/cfg-lex.l
@@ -36,6 +36,10 @@
 static gboolean
 yy_input_run_backtick_substitution(CfgLexer *self, gchar *buf, gsize buf_size, gsize *len)
 {
+  if (cfg_lexer_is_backtick_substitution_postponed(self))
+    {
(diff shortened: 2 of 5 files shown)


kira-syslogng:
Build SUCCESS

lib/cfg-grammar.y
```diff
@@ -871,6 +871,7 @@ block_stmt
             free($4);
             free($10);
             last_block_args = NULL;
+            cfg_lexer_enable_immediate_backtick_substition(lexer);
```
bazsi, line 874: do we also need to skip backtick substitution in the body of the block or just in arguments? this seems to apply it for the body too.

lib/cfg-grammar.y
```diff
@@ -892,7 +893,11 @@ block_arg
           LL_BLOCK
           {
             cfg_lexer_pop_context(lexer);
-            cfg_args_set(last_block_args, $1, $3); free($1); free($3);
+            if ($3 && cfg_lexer_is_backtick_substitution_postponed(lexer) && strchr($3, '`'))
```
bazsi, line 896: As it seems, here backtick substitution is always delayed. why do we need the check?

also the strchr() should probably be wrapped into a function that describes its intent.

lib/cfg-grammar.y
```diff
@@ -892,7 +893,11 @@ block_arg
...
+              cfg_args_set(last_block_args, $3, $3);
```
bazsi, line 897: I don't understand the first cfg_args_set call. It sets the arg name $3 with the same value as the key? That's the body of the argument, so in this case:
```
block foo(arg1(alma) arg2(`arg1`)) {

}
```
The first arg doesn't have a backtick, so we'd store "arg1"="alma", the second arg does have a backtick, so we would store "`arg1`"="`arg1`" and arg2="`arg1`"

What's the point in storing the backticked version?

Can you show an example? Thanks

lbudai: backtick in a key can only happens if the backtick substitution is postponed - so the `cfg_lexers_subst_invoke`  can check if a value is stored in the block_args as name - if it is stored, then backtick subst is postponed

PR #3658, opened 2021-04-27
Fix config parsing for kafka-c/config

Kokan:
Global keywords could conflict with librdkafka's own configuration keywords. An example of this is the `retries`. This PR fixes that, as in destination/kafka-c/config context temporarly remove the keywords (using mechanism of `CFG_KEYWORD_STOP`).
```
@Version: 3.30

log {
  destination {
    kafka-c(
      topic("test-topic")
      bootstrap-servers("localhost:9092")
      config(
          retries => 1
      )
    );
  };
};
```

### DIFF
--- a/modules/kafka/kafka-grammar.ym
+++ b/modules/kafka/kafka-grammar.ym
@@ -52,6 +52,7 @@
 /* INCLUDE_DECLS */
 
 %type <ptr> kafka_properties
+%type <ptr> kafka_properties_list
 %type <ptr> kafka_property
 
 %token KW_KAFKA
@@ -123,7 +124,20 @@ kafka_option
         ;
 
 kafka_properties
-        : kafka_property kafka_properties			{ $$ = g_list_prepend($2, $1); }
+	:
+	{
+		static CfgLexerKeyword empty_keywords[] = {{ CFG_KEYWORD_STOP }};
+		cfg_lexer_push_context(lexer, LL_CONTEXT_CONFIG, empty_keywords, "kafka_config_block");
+	}
+	kafka_properties_list
+	{
+		cfg_lexer_pop_context(lexer);
+		$$ = $2;
+	}
+	;
+
+kafka_properties_list
+        : kafka_property kafka_properties_list			{ $$ = g_list_prepend($2, $1); }
         |							{ $$ = NULL; }
         ;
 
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

gaborznagy (dismissed):
Looks good to me.

Only a minor thing, that's optional: as we reuse the same mechanism that `block()` uses, we should reconsider creating a wrapper function in cfg-lexer for better readability: e.g. `cfg_lexer_disable_keyword_processing(gchar *desc)` instead of push_context.

bazsi:
An alrenrative idea is to use a dedicated context, instead of a overloading the BLOCK_DEF context.

Kokan:
For now, I've only added a new context `LL_CONTEXT_CONFIG`, and stopped here. This should be good enough for fixing the issue. I have  a branch that needs some more work to extract the `KW_CONFIG` block as a general solution, and use it in other places.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

gaborznagy (approved):

modules/kafka/kafka-grammar.ym
```diff
@@ -123,7 +124,20 @@ kafka_option
...
-        : kafka_property kafka_properties			{ $$ = g_list_prepend($2, $1); }
+	:
+	{
+		static CfgLexerKeyword empty_keywords[] = {{ CFG_KEYWORD_STOP }};
+		cfg_lexer_push_context(lexer, LL_CONTEXT_CONFIG, empty_keywords, "kafka_config_block");
+	}
+	kafka_properties_list
+	{
+		cfg_lexer_pop_context(lexer);
+		$$ = $2;
```
MrAnno, line 135: `$2` does not seem to be a valid index here.

news/bugfix-3658.md
```diff
@@ -0,0 +1 @@
+kafka: the config() block couldn't contain option that is a already a keyword in syslog-ng (example: retries)
```
MrAnno, line 1: typo: `a`

modules/kafka/kafka-grammar.ym
MrAnno: Sorry, I was wrong, `kafka_properties_list` is going to be `$2` here due to the first block, in which we create a context.

gaborznagy (approved):

MrAnno (approved):
Thank you.

kira-syslogng:
Build SUCCESS

PR #3661, opened 2021-04-28
Fix compilation issues on OpenBSD

millert:
This PR fixes the following issues that prevent building syslog-ng on OpenBSD:
 * OpenBSD uses a different flavor of reentrant getent functions
 * OpenBSD doesn't support IPv4-mapped IPv6 addresses (they are considered harmful)
 * OpenBSD uses IP_RECVDSTADDR for UDP dest-IP like FreeBSD does
 * tlscontext.c uses AF_INET and AF_INET6 but doesn't include sys/socket.h where they are defined.

### DIFF
--- a/lib/compat/getent-openbsd.c
+++ b/lib/compat/getent-openbsd.c
@@ -1,0 +1,113 @@
+/*
+ * Copyright (c) 2017 Balabit
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "compat/getent-openbsd.h"
+
+#ifdef __OpenBSD__
+#include <errno.h>
+#include <stddef.h>
+
+int
+_compat_openbsd__getprotobynumber_r(int proto,
+                                    struct protoent *result_buf, char *buf,
+                                    size_t buflen, struct protoent **result)
+{
+  struct protoent_data *data_buf = (struct protoent_data *)buf;
+
+  if (buflen < sizeof(*data_buf))
+    {
+      *result = NULL;
+      return ERANGE;
+    }
+  if (getprotobynumber_r(proto, result_buf, data_buf) == -1)
+    {
+      *result = NULL;
+      return errno;
+    }
+  *result = result_buf;
+  return 0;
+}
+
+int
+_compat_openbsd__getprotobyname_r(const char *name,
+                                  struct protoent *result_buf, char *buf,
+                                  size_t buflen, struct protoent **result)
+{
+  struct protoent_data *data_buf = (struct protoent_data *)buf;
+
+  if (buflen < sizeof(*data_buf))
+    {
+      *result = NULL;
+      return ERANGE;
+    }
+  if (getprotobyname_r(name, result_buf, data_buf) == -1)
+    {
+      *result = NULL;
+      return errno;
+    }
+  *result = result_buf;
+  return 0;
+}
+
+int
+_compat_openbsd__getservbyport_r(int port, const char *proto,
+                                 struct servent *result_buf, char *buf,
+                                 size_t buflen, struct servent **result)
+{
+  struct servent_data *data_buf = (struct servent_data *)buf;
+
+  if (buflen < sizeof(*data_buf))
+    {
+      *result = NULL;
+      return ERANGE;
+    }
+  if (getservbyport_r(port, proto, result_buf, data_buf) == -1)
+    {
+      *result = NULL;
+      return errno;
+    }
+  *result = result_buf;
+  return 0;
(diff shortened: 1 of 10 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

lib/compat/getent-openbsd.c
```diff
@@ -0,0 +1,92 @@
...
+                                    struct protoent *result_buf, char *buf,
+                                    size_t buflen, struct protoent **result)
+{
+  struct protoent_data *data_buf = (struct protoent_data *)buf;
+  int ret = 0;
+
+  if (buflen < sizeof(*data_buf))
+    ret = ERANGE;
+  else if (getprotobynumber_r(proto, result_buf, data_buf) == -1)
+    ret = errno;
```
Kokan, line 40: Usually in syslog-ng the curly bracket used with condition in order to improve readability (while there are exception from this, like early return cases). The other thing that I would ask you to do is exactly early returns.

My proposal would be the following:
```c
struct protoent_data *data_buf = (struct protoent_data *)buf;

if (buflen < sizeof(*data_buf)) 
   return ERANGE;

if (getprotobynumber_r(proto, result_buf, data_buf) == -1)
  return errno;
  
  *result = result_buf;
  return 0;
```

It has a behavior change as well, in case of an error `result` buffer is not touched, and that should be fine. The caller should not check the `result` buffer in case of an error as its content is undefined most of the time.
(This of course applies for the other places, just the comment is the same.)

kira-syslogng:
Build SUCCESS

lib/host-resolve.c
```diff
@@ -36,6 +36,10 @@
 #include <string.h>
 #include <resolv.h>
 
+#ifndef AI_V4MAPPED
```
Kokan, line 39: I would move this to a "global" level, as in other places in the future we might relay on this constant and that would prevent breaking openbsd again.
By global I mean the `syslog-ng-config.h` which is created by configure and cmake. Also that way it is detected

Kokan (commented):
Thanks for the patches.

lib/compat/getent-openbsd.c
millert: The Linux API requires that result be set to NULL on error and the code in syslog-ng which calls it relies on this behavior.  That is why I was careful to always clear it on failure.

lib/host-resolve.c
millert: Do you mean adding a configure/cmake test and then something like this in `syslog-ng-config.h`?
```
#ifndef SYSLOG_NG_HAVE_AI_V4MAPPED
#define AI_V4MAPPED 0
#endif
```

lib/compat/getent-openbsd.c
millert: I pushed an update to this commit that adds braces and uses early return.

kira-syslogng:
Build SUCCESS

lib/host-resolve.c
millert: Unfortunately, I don't see a way to accomplish that in `syslog-ng-config.h` itself using autoconf since any define will be prefixed with `SYSLOG_NG_`.  This is the case even when using `AH_BOTTOM`

Kokan: Yes, and that needs to be remapped. But I am fine with this version. The plan is to include this change into the next release.

Kokan (approved):

MrAnno (commented):
Thank you for this pull request!

One thing I noticed is that `getent-openbsd.c` is not built, the following lines seems to be missing from our build files:

```diff
diff --git a/lib/compat/CMakeLists.txt b/lib/compat/CMakeLists.txt
index 92330189d..2e2511c65 100644
--- a/lib/compat/CMakeLists.txt
+++ b/lib/compat/CMakeLists.txt
@@ -11,6 +11,7 @@ set(COMPAT_HEADERS
     compat/pcre.h
     compat/getent.h
     compat/getent-sun.h
+    compat/getent-openbsd.h
     compat/getent-generic.h
     compat/un.h
     PARENT_SCOPE)
@@ -27,6 +28,7 @@ set(COMPAT_SOURCES
     compat/time.c
     compat/openssl_support.c
     compat/getent-sun.c
+    compat/getent-openbsd.c
     compat/getent-generic.c
     PARENT_SCOPE)

diff --git a/lib/compat/Makefile.am b/lib/compat/Makefile.am
index d537b3914..968e154fe 100644
--- a/lib/compat/Makefile.am
+++ b/lib/compat/Makefile.am
@@ -17,6 +17,7 @@ compatinclude_HEADERS         =       \
        lib/compat/pcre.h               \
        lib/compat/getent.h             \
        lib/compat/getent-sun.h         \
+       lib/compat/getent-openbsd.h \
        lib/compat/getent-generic.h     \
        lib/compat/un.h

@@ -32,6 +33,7 @@ compat_sources                        =       \
        lib/compat/time.c               \
        lib/compat/openssl_support.c    \
        lib/compat/getent-sun.c         \
+       lib/compat/getent-openbsd.c \
        lib/compat/getent-generic.c

 include lib/compat/tests/Makefile.am
```

MrAnno:
I was able to compile this PR with the above Makefile patch on OpenBSD 6.9 (bison 3.7.6 compiled from source, gmake).

`NULL` was undeclared, but including stddef.h fixed it:

```
../lib/compat/getent-openbsd.c: In function '_compat_openbsd__getprotobynumber_r':
../lib/compat/getent-openbsd.c:38: error: 'NULL' undeclared (first use in this function)
```

```diff
diff --git a/lib/compat/getent-openbsd.c b/lib/compat/getent-openbsd.c
index 1b1af9eaf..ca36fb07b 100644
--- a/lib/compat/getent-openbsd.c
+++ b/lib/compat/getent-openbsd.c
@@ -25,6 +25,7 @@

 #ifdef __OpenBSD__
 #include <errno.h>
+#include <stddef.h>

 int
 _compat_openbsd__getprotobynumber_r(int proto,
```

millert:
@MrAnno Thanks, I came up with an identical fix for the compilation error.

kira-syslogng:
Build SUCCESS

Kokan (approved):

MrAnno (approved):
Thank you.

PR #3647, opened 2021-04-15
*.py: s/python/python3/ (exclude tests)

yifan19:
As stated by https://github.com/syslog-ng/syslog-ng/pull/3603
python2 is EOL.

Fix all shebangs calling python instead of python3
except the tests.

(correcting lib/merge-grammar.py)
Signed-off-by: Joe Slater <joe.slater@windriver.com>
(adding the rest)2a1a44fade7e73872361568bd2dad235d083dcf1
Signed-off-by: Yi Fan Yu <yifan.yu@windriver.com>

PR #3663, opened 2021-05-05
packaging/rhel: remove python3 path fixup in syslog-ng.spec

gaborznagy:
After PR #3647 lib/merge-grammar.py has python3 in it's shebang instead of "python",
there is no need for the sed command anymore in the .spec file (python2 is not supported).

Also, the sed expression currently creates an invalid python reference, see result:
`#!/usr/bin/python33`

Packaging CI jobs doesn't hit this issue, as they are creating packages from tarball
(as intended due to bison v3.7.6 which is not yet available on many platforms).
This bug prevented releasing on rpm platforms where the grammar files are changed.

This PR is not critical from OSE release point of view.

### DIFF
--- a/packaging/rhel/syslog-ng.spec
+++ b/packaging/rhel/syslog-ng.spec
@@ -274,10 +274,6 @@ ryslog is not on the system.
 # fix perl path
 %{__sed} -i 's|^#!/usr/local/bin/perl|#!%{__perl}|' contrib/relogger.pl
 
-# fix Python path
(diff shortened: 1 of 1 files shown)


gaborznagy:
@czanik Can you check this PR, please?

kira-syslogng:
Build SUCCESS

MrAnno (approved):

czanik:
:+1: tested the change both on RHEL and openSUSE, and both work fine.

czanik (approved):
Works fine on the platforms I tested.

PR #3670, opened 2021-05-10
Logreader ignore aux data flag

bazsi:
syslog-ng automatically collects information such as PID, UNIX UID/GID and similar process related information on the sender application when local logs are delivered via the unix-dgram() or unix-stream() drivers. This information becomes part of each log message, under the ".unix" prefix.

In extreme logging loads, the auxiliary data processing path might become a bottleneck, causing syslog-ng to consume more CPU than the application generating the logs (it was a Zorp instance at -v8, which is known to produce huge amount of logs).

This branch introduces a new flag, "ignore-aux-data" that allows disabling the collecting of UNIX related aux data.

### DIFF
--- a/modules/afsocket/compat-unix-creds.c
+++ b/modules/afsocket/compat-unix-creds.c
@@ -21,29 +21,16 @@
  *
  */
 
-#ifndef _SYSLOG_NG_UNIX_CREDENTIALS_H
-#define _SYSLOG_NG_UNIX_CREDENTIALS_H
+#include "compat-unix-creds.h"
 
-#include <sys/types.h>
-#include <sys/socket.h>
-
-#include "syslog-ng.h"
-
-#if defined(__linux__)
-# if SYSLOG_NG_HAVE_STRUCT_UCRED && SYSLOG_NG_HAVE_CTRLBUF_IN_MSGHDR
-# define CRED_PASS_SUPPORTED
-# define cred_t struct ucred
-# define cred_get(c,x) (c->x)
-# endif
-#elif defined(__FreeBSD__)
-# if SYSLOG_NG_HAVE_STRUCT_CMSGCRED && SYSLOG_NG_HAVE_CTRLBUF_IN_MSGHDR
-#  define CRED_PASS_SUPPORTED
-#  define SCM_CREDENTIALS SCM_CREDS
-#  define cred_t struct cmsgcred
-#  define cred_get(c,x) (c->cmcred_##x)
-# endif
+void
+setsockopt_so_passcred(gint fd, gint onoff)
+{
+#ifdef LOCAL_CREDS
+  setsockopt(fd, 0, LOCAL_CREDS, &onoff, sizeof(onoff));
+#else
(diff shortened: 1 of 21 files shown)


bazsi:
I have now rebased and improved this pull request according to our latest IRL discussion. Now it comprises of the following
changes:
  * I factored out the SO_PEERCRED related code from TransportMapperUnix to a newly created SocketOptionsUnix (we only had SocketOptionsInet so far), this is where it should have been implemented in the first place
  * it adds so-peercred() option, quite akin to the TCP keep alive/ttl/etc options
  * I deprecated pass-unix-credentials() options as having two set of options to control the same is confusing and pass-unix-credentials() had a global, none of the other socket-options have. We still support the keyword, but emit a warning during startup to avoid its use.
  * it has a potential memory leak fix
  * a version bump, because I needed it for the keyword deprecation.

With these, I think this can be merged by 3.35.

kira-syslogng:
Build FAILURE

bazsi:
I have added a NEWS entry. Hopefully this suffices for this to be merged.

Thanks

kira-syslogng:
Build SUCCESS

modules/afsocket/socket-options-unix.h
```diff
@@ -0,0 +1,43 @@
...
+  SocketOptions super;
+  gint so_passcred;
+} SocketOptionsUnix;
+
+static inline void
+socket_options_unix_set_so_passcred(SocketOptions *s, gint so_passcred)
+{
+  SocketOptionsUnix *self = (SocketOptionsUnix *) s;
+
+  self->so_passcred = TRUE;
```
OverOrion, line 38: I might have misunderstood something, but shouldn't it be `self->so_passcred = so_passcred`?

bazsi: good catch.

bazsi: fixed this in the latest rebase round.

modules/afsocket/compat-unix-creds.c
```diff
@@ -21,24 +21,16 @@
...
  */
 
-#include "syslog-ng.h"
-#include "unix-credentials.h"
-
+#include "compat-unix-creds.h"
 
 void
-socket_set_pass_credentials(gint fd)
+setsockopt_so_passcred(gint fd, gint onoff)
```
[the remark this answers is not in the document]
bazsi: I've decided against this as this is only used by the afsocket module, but indeed this could be moved there and should if there's any other user.

kira-syslogng:
Build SUCCESS

Kokan (approved):

bazsi:
@Kokan I've noticed some dead code in one of the tests, which I have removed with a new commit. Could you please re-approve?

@alltilla @OverOrion could you please review this so it can get in?

thanks

kira-syslogng:
Build SUCCESS

alltilla (approved):

kkovaacs:
Thanks a lot, guys!

PR #3679, opened 2021-05-17
java(): fix debug logging of Java-based destinations

MrAnno:
This PR fixes Java debug logging.
Previously log4j's log level was left on its default value, even when syslog-ng was started in debug/trace mode.

Now the log4j log level is set to syslog-ng's internal log level, so they'll be in sync.

Note:
Currently, the log level is not updated when debug/trace is enabled or disabled at runtime (using syslog-ng-ctl).

----

The class names and their responsibilities can be confusing here:
- `InternalMessageSender` - this is a native Java "proxy" class that can be used to send messages through the `internal()` source.
- `SyslogNgInternalLogger` - this is a log4j log appender implementation, which uses `InternalMessageSender` - it extends `org.apache.log4j.AppenderSkeleton`.

### DIFF
--- a/modules/java-modules/kafka/src/main/java/org/syslog_ng/KafkaDestination.java
+++ b/modules/java-modules/kafka/src/main/java/org/syslog_ng/KafkaDestination.java
@@ -46,8 +46,11 @@ public class KafkaDestination extends StructuredLogDestination {
 
   public KafkaDestination(long handle) {
     super(handle);
+
+    Logger.getLogger("org.apache.kafka").setLevel(SyslogNgInternalLogger.getLevel());
     logger = Logger.getRootLogger();
     SyslogNgInternalLogger.register(logger);
+
     options = new KafkaDestinationOptions(this);
     properties = new KafkaDestinationProperties(options);
   }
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

modules/java-modules/kafka/src/main/java/org/syslog_ng/KafkaDestination.java
```diff
@@ -46,8 +46,11 @@
 
   public KafkaDestination(long handle) {
     super(handle);
+
+    Logger.getLogger("org.apache.kafka").setLevel(SyslogNgInternalLogger.getLevel());
```
Kokan, line 50: Why do we need to set this logger ? Does this mean that it branches off too early from root logger and not having the proper log level, or that root logger log level is not set according to syslog-ng level ?
Should we set log level for other java based destination's logger?

Kokan:
Also please write a news entry for the new java method (`getLevel`).

modules/java-modules/kafka/src/main/java/org/syslog_ng/KafkaDestination.java
MrAnno: The root log level has been set to the syslog-ng internal's log level in this PR.

> Should we set log level for other java based destination's logger?

I think it's not necessary because we set the root log level and all other child loggers inherit the root (their parent's) level.

Kafka seems to be different, I haven't checked their implementation, but they might set their own log level because ours is not taken into account. I would consider this commit as a workaround.

MrAnno:
`getLevel()` is `public` only because of the workaround in the Kafka destination. I don't think we should add this to the NEWS file, not even to the documentation.

modules/java-modules/kafka/src/main/java/org/syslog_ng/KafkaDestination.java
Kokan: Thanks for the explanation.

Kokan (approved):

szemere (approved):
Sorry for the long delay. I reviewed the code earlier, but I also wanted to test how it works locally. Good Job! Thank You!

PR #3681, opened 2021-05-21
Kafka transactional producer

Kokan:
Uses the transaction api for kafka. 

Note1: currently it reuses the `sync_send` option, but that is not final.
Note2: the delivery report should be informal in case of transactional api, but it also removes the not delivered message push back mechanism without it. The reasoning that message loss cannot be prevented with pushing back not delivered messages that (due reduces the chance), but the user could do that with transactional api. (which is probably much slower)


kira-syslogng:
Build FAILURE

gaborznagy:
> Uses the transaction api for kafka.
> 
> Note1: currently it reuses the `sync_send` option, but that is not final.
> Note2: the delivery report should be informal in case of transactional api, but it also removes the not delivered message push back mechanism without it. The reasoning that message loss cannot be prevented with pushing back not delivered messages that (due reduces the chance), but the user could do that with transactional api. (which is probably much slower)

Don't forget to mention that librdkafka has it's own retry mechanism which is highly configurable, so even though it seems rough that we will dismiss failed delivery attempt, in reality the situation is better.

modules/kafka/kafka-dest-driver.c
```diff
@@ -310,8 +305,8 @@ _kafka_delivery_report_cb(rd_kafka_t *rk,
                 evt_tag_str("error", rd_kafka_err2str(err)),
                 evt_tag_str("driver", self->super.super.super.id),
                 log_pipe_location_tag(&self->super.super.super.super));
-      log_msg_unref(msg);
     }
+  log_msg_unref(msg);
   log_threaded_dest_worker_wakeup_when_suspended((LogThreadedDestWorker *) self->super.workers[0]);
```
gaborznagy, line 310: just a note, I'll have to check it too:
I think wakeup here is not needed: the actual ack is done right after `insert`, I think wakeup here was meant for the error-handling case.

modules/kafka/kafka-dest-driver.c
```diff
@@ -512,6 +516,23 @@ _init_topic_name(KafkaDestDriver *self)
 }
 
 
+static gboolean
+_transaction_init(KafkaDestDriver *self)
+{
+  rd_kafka_error_t *error = rd_kafka_init_transactions(self->kafka, -1);
```
gaborznagy, line 522: I've found this in the documentation of librdkafka:
> Upon successful return from this function the application has to perform at least one of the following operations within transaction.timeout.ms to avoid timing out the transaction on the broker: ...

I guess this means that if there's no message at startup for transaction.timeout.ms (default 60 seconds), we will have failure on the first message producing.

gaborznagy (commented):
Looks good to me at first glance, I'll do a thorough review next week!

modules/kafka/kafka-dest-driver.c
Kokan: Not really, but actually passing the whole logmsg probably not needed here at all.

gaborznagy: I agree, we can remove the LogMessage passing.

About the `log_threaded_dest_worker_wakeup_when_suspended`: I've double checked: this wakeup is not necessary anymore. It's purpose is to avoid the worker thread getting stuck (i.e. being idle while there are messages in its queue: when there is no new message, but we pushed back a failed-to-deliver message from the delivery_callback). However sometimes it didn't work: `log_threaded_dest_worker_wakeup_when_suspended` needed that the worker's state is `suspended = TRUE`, which only happens if the previous publish call in worker#0 failed.

Also, during shutdown the last flush attempt from deinit can cause a segmentation fault: as `log_threaded_dest_worker_wakeup_when_suspended` is being called from the main thread, which calls the `lib/logthrdest/logthrdestdrv.c:_perform_work` function, and that's leading that the main thread tries to access a different ivykis main loop's timers and objects (especially that the worker thread is already stopped, and the worker thread's ivykis main loop is already deinitialized).

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-driver.c
gaborznagy: With this version I don't see the segfault anymore.

Also, on master branch, I've removed only `log_threaded_dest_worker_wakeup_when_suspended` for a test run to see if it fixes the crash and I didn't see any crash at all.

modules/kafka/kafka-dest-driver.c
gaborznagy: I've checked the previous version of this PR and although the documentation states that if no produce/commit/abort happens in `transaction.timeout.ms` time the transaction will fail, I didn't experience that.

Nevertheless, I agree with the new way of initializing the transaction.

modules/kafka/kafka-dest-driver.c
```diff
@@ -273,25 +285,19 @@ _kafka_delivery_report_cb(rd_kafka_t *rk,
...
   /* we already ACKed back this message to syslog-ng, it was kept in
    * librdkafka queues so far but successfully delivered, let's unref it */
 
   if (err != RD_KAFKA_RESP_ERR_NO_ERROR)
     {
-      LogThreadedDestWorker *worker = (LogThreadedDestWorker *) self->super.workers[0];
-      LogQueue *queue = worker->queue;
-      LogPathOptions path_options = LOG_PATH_OPTIONS_INIT_NOACK;
-
       msg_debug("kafka: delivery report for message came back with an error, putting it back to our queue",
```
gaborznagy, line 294: minor stuff, but still: Can you reword the debug message too, please? :)
It states that we re-add the message to the queue of syslog-ng.

gaborznagy (commented):
Looks good to me.

modules/kafka/kafka-dest-driver.c
Kokan: Sure!

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-worker.c
```diff
@@ -102,6 +102,89 @@ kafka_dest_worker_calculate_topic(KafkaDestWorker *self, LogMessage *msg)
...
+#ifdef SYSLOG_NG_HAVE_RD_KAFKA_INIT_TRANSACTIONS
+  KafkaDestDriver *owner = (KafkaDestDriver *) self->super.owner;
+
+  if (!owner->transaction_commit)
+    return TRUE;
+
+  if (owner->transaction_inited)
+    return TRUE;
+
+  rd_kafka_error_t *error = rd_kafka_init_transactions(owner->kafka, -1);
```
gaborznagy, line 117: Just a point for improvement: 
I didn't consider the `timeout` parameters of `rd_kafka_init_transactions` and `rd_kafka_commit_transaction` as it seemed fine to block on these calls. However, during shutdown/reload this could take a long time (it's configurable though e.g. `transaction.timeout.ms` defaults to 60 seconds).

We have `flush_timeout_on_shutdown` and `flush_timeout_on_reload` options. We could reuse them even if the option names are tied to the `rd_kafka_flush` call.

gaborznagy:
Please add a news entry.

modules/kafka/kafka-dest-driver.c
```diff
@@ -281,25 +281,16 @@ _kafka_delivery_report_cb(rd_kafka_t *rk,
...
-   * librdkafka queues so far but successfully delivered, let's unref it */
 
   if (err != RD_KAFKA_RESP_ERR_NO_ERROR)
     {
-      LogThreadedDestWorker *worker = (LogThreadedDestWorker *) self->super.workers[0];
-      LogQueue *queue = worker->queue;
-      LogPathOptions path_options = LOG_PATH_OPTIONS_INIT_NOACK;
-
-      msg_debug("kafka: delivery report for message came back with an error, putting it back to our queue",
+      msg_debug("kafka: delivery report for message came back with an error, message is lost",
```
gaborznagy, line 291: I've recently found that some error/warning message still indicate that on failure to send a message it will be readded to syslog-ng's queue:
https://github.com/syslog-ng/syslog-ng/blob/3e62aa98fafc1507f7fe8344b850e7e608e92c02/modules/kafka/kafka-dest-driver.c#L423-L426

We should change this to `... the remaining entries will be purged and lost`

dbld/images/devshell.dockerfile
```diff
@@ -8,3 +8,4 @@ RUN /dbld/builddeps install_perf
 RUN /dbld/builddeps install_apt_packages
 RUN /dbld/builddeps install_pip2
 RUN /dbld/builddeps install_pip_packages
+RUN /dbld/builddeps install_bison_from_source
```
gaborznagy, line 11: I missed this one: why do we need this?
We inherit bison from `balabit/syslog-ng-tarball` image:
https://github.com/syslog-ng/syslog-ng/blob/master/dbld/images/tarball.dockerfile#L1-L10

```
$ docker run -it --rm balabit/syslog-ng-devshell bash
root@c1c04bd96a0b:/# 
root@c1c04bd96a0b:/# 
root@c1c04bd96a0b:/# bison --version
bison (GNU Bison) 3.7.6
Written by Robert Corbett and Richard Stallman.

Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

```

Kokan: We don't need this. It would have been its own commit explaining why we need this if it was on purpose. I'll remove.

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-driver.c
Kokan: This should be fixed, also a commit close by.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/kafka/kafka-dest-driver.c
```diff
@@ -606,16 +661,9 @@ kafka_dd_free(LogPipe *d)
...
   log_template_options_destroy(&self->template_options);
-  if (self->topics)
-    g_hash_table_unref(self->topics);
+  _destroy_kafka(&self->super.super.super);
   log_template_unref(self->key);
   log_template_unref(self->message);
-  if (self->topic)
-    rd_kafka_topic_destroy(self->topic);
-  if (self->fallback_topic_name)
-    g_free(self->fallback_topic_name);
```
Barkodcz, line 616: fallback_topic_name should be free up, right?

Kokan: fixed.

kira-syslogng:
Build SUCCESS

Barkodcz (dismissed):

modules/kafka/kafka-dest-worker.c
```diff
@@ -340,6 +351,26 @@ _thread_init(LogThreadedDestWorker *s)
...
+static gboolean
+kafka_dest_worker_connect(LogThreadedDestWorker *s)
+{
+  KafkaDestWorker *self = (KafkaDestWorker *)s;
+  KafkaDestDriver *owner = (KafkaDestDriver *) self->super.owner;
+
+  static gboolean first_init = FALSE;
+  if (!first_init)
+    {
+      first_init = TRUE;
```
gaborznagy, line 363: This branch expects that kafka_dd_init already initialized the librdkafka client.
Can we do an assert here at first_init or an explicit check on owner->kafka object ?

gaborznagy (commented):
Only minor comments about rename and assertion.

szemere (approved):

modules/kafka/kafka-dest-worker.c
```diff
@@ -102,6 +102,32 @@ kafka_dest_worker_calculate_topic(KafkaDestWorker *self, LogMessage *msg)
   return kafka_dest_worker_get_literal_topic(self);
 }
 
+static LogThreadedResult
+_handle_error(KafkaDestWorker *self, rd_kafka_error_t *error)
```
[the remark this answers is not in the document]
Kokan: Sure.

modules/kafka/kafka-dest-worker.c
```diff
@@ -272,21 +272,23 @@ static LogThreadedResult
 kafka_dest_worker_insert(LogThreadedDestWorker *s, LogMessage *msg)
 {
   KafkaDestWorker *self = (KafkaDestWorker *)s;
+  LogThreadedResult result;
 
-  _drain_responses(self);
```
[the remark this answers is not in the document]
Kokan: No, it wasn't. I'd keep this drain function call in place.

modules/kafka/kafka-dest-worker.c
```diff
@@ -340,6 +351,26 @@ _thread_init(LogThreadedDestWorker *s)
   return log_threaded_dest_worker_init_method(s);
 }
 
+static gboolean
+kafka_dest_worker_connect(LogThreadedDestWorker *s)
+{
+  KafkaDestWorker *self = (KafkaDestWorker *)s;
+  KafkaDestDriver *owner = (KafkaDestDriver *) self->super.owner;
```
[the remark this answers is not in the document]
Kokan: I do not see an expression that I could use in that assert. (Asserting on either running in `worker#0` or `sync-send(yes)` would not do justice.) Using this in case of multiple worker would require additional consideration.

kira-syslogng:
Build SUCCESS

Barkodcz (approved):

modules/kafka/kafka-dest-worker.c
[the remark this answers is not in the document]
gaborznagy: Okay, thanks for checking.

gaborznagy (approved):

Issue #3551, opened 2021-01-26
High disk utilisation after network destination reconnect with reliable disk buffer


L4rS6:
# syslog-ng
## Version of syslog-ng
```
syslog-ng 3 (3.30.1)
Config version: 3.29
Installer-Version: 3.30.1
Revision:
Compile-Date: Nov 22 2020 00:16:57
Module-Directory: /usr/local/lib/syslog-ng
Module-Path: /usr/local/lib/syslog-ng
Include-Path: /usr/local/share/syslog-ng/include
Available-Modules: appmodel,json-plugin,stardate,affile,system-source,disk-buffer,timestamp,http,map-value-pairs,examples,confgen,afstomp,kvformat,afuser,csvparser,afprog,basicfuncs,xml,afsocket,hook-commands,tfgetent,secure-logging,syslogformat,add-contextual-data,dbparser,cef,pseudofile,graphite,linux-kmsg-format,cryptofuncs,tags-parser,azure-auth-header
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: off
Enable-TCP-Wrapper: off
Enable-Linux-Caps: off
Enable-Systemd: off
```

## Platform
`FreeBSD 11.4`

# Issue
After network reconnect which uses a reliable disk buffer syslog-ng utilise the disk nearly to 100%. This problem can be fixed by disabling the disk buffer, restarting syslog-ng and reenabling the disk buffer.

## Steps to reproduce

1. Start logging: `while true; do logger -p kern.crit test; sleep 0.001; done`
2. Check disk utilisation: `gstat` (Disk has nearly no utilisation)
3. Drop TCP connection to logging host: `tcpdrop 10.0.0.1 5555 10.0.0.2 514`
4. Wait until syslog-ng reconnects to remote logging host (default 1 minute)
5. Check disk utilisation: `gstat` (Disk has nearly 100% utilisation)
6. Issue can be fixed by: Remove disk buffer part in configuration --> `service syslog-ng restart` --> Readd disk buffer part in configuration --> `service syslog-ng restart`

## Configuration
```
@version:3.30

options {
  create_dirs(yes);
  keep_hostname(yes);
  threaded(yes);
  log_fifo_size(32000);
  frac_digits (3);
};

source s_local {
  system();
  internal();
 };

destination d_network {
  syslog(
    "logging.myhost"
    port(514)
    disk-buffer(
[...]

PR #3689, opened 2021-05-27
qdisk: don't truncate if it would make a small size difference

alltilla:
TODO:
- [x] Run performance test (No noticeable speed-up)
- [x] Check KIRA failure
- [x] Add `disk-buffer()` block to global options
- [x] rename to `truncate-size-ratio()`

---

`ftruncate()` is an expensive call.

In the current form, in case of a reliable disk-buffer when the destination acks the message, the truncate is usually called.

Let's look at the situation, where the destination can keep up with the speed of the source:
   1. Incoming message, it gets written to the disk buffer, the write head moves to `RESERVED_SPACE` + `msg_size`.
   2. We read the message from the disk-buffer, and send it to the destination, the read head moves to `RESERVED_SPACE` + `msg_size`.
   3. The destination acks the message, the backlog_head moves to `RESERVED_SPACE` + `msg_size`.
   4. At the end of the ack, we see, that the disk buffer is "empty", meaning all heads are at the same position, so we truncate the file to `RESERVED_SPACE`, and move all 3 heads there.
   5. A new message arrives, and we go to step 1.

In this scenario we call `ftruncate()` for every incoming message.
I have made a test code (a separate PoC code, **not** in syslog-ng), which measures the `fwrite`, `fread`, `ftruncate` iteration's time taken, and I observed, that more than half of the real word time was taken by `ftruncate()`:
```
Writing, reading 1048576 * 1024 bytes messages, truncating the file to 4096 bytes per message.
   [1/10] DONE in 7.856170541 seconds, from which ftruncate() took 4.734487726 seconds.
   [2/10] DONE in 7.773038602 seconds, from which ftruncate() took 4.678402817 seconds.
   [3/10] DONE in 7.780979727 seconds, from which ftruncate() took 4.664749331 seconds.
   [4/10] DONE in 7.805741840 seconds, from which ftruncate() took 4.675652459 seconds.
   [5/10] DONE in 7.745974862 seconds, from which ftruncate() took 4.642292874 seconds.
   [6/10] DONE in 7.800890915 seconds, from which ftruncate() took 4.690607455 seconds.
   [7/10] DONE in 7.774325797 seconds, from which ftruncate() took 4.693368755 seconds.
   [8/10] DONE in 7.806343633 seconds, from which ftruncate() took 4.683172069 seconds.
   [9/10] DONE in 7.896176455 seconds, from which ftruncate() took 4.757315806 seconds.
   [10/10] DONE in 7.868837540 seconds, from which ftruncate() took 4.715273111 seconds.
```
I've done the measurements on an SSD on my personal Ubuntu Focal machine.

To optimize the calls of `ftruncate()`, in this commit I have introduced a new logic: we calculate the possible size reduction in each `_truncate_file()` call, and we do not call `ftruncate()` if it is less, than 1 percent of the maximal disk-buffer size.

My measurements changed drastically when implementing the same logic in the test code:
```
Writing, reading 1048576 * 1024 bytes messages, truncating the file to 4096 bytes if the size reduction reaches 1% of max size.
   [1/10] DONE in 2.255469215 seconds, from which ftruncate() took 0.107140776 seconds.
   [2/10] DONE in 2.226006879 seconds, from which ftruncate() took 0.109389325 seconds.
   [3/10] DONE in 2.239258063 seconds, from which ftruncate() took 0.110315009 seconds.
   [4/10] DONE in 2.229625672 seconds, from which ftruncate() took 0.110243967 seconds.
   [5/10] DONE in 2.255923025 seconds, from which ftruncate() took 0.108881297 seconds.
   [6/10] DONE in 2.223326513 seconds, from which ftruncate() took 0.107711562 seconds.
   [7/10] DONE in 2.251854651 seconds, from which ftruncate() took 0.134951278 seconds.
   [8/10] DONE in 2.235311433 seconds, from which ftruncate() took 0.126703875 seconds.
   [9/10] DONE in 2.228932757 seconds, from which ftruncate() took 0.109558716 seconds.
   [10/10] DONE in 2.259198891 seconds, from which ftruncate() took 0.109496328 seconds.
```

This logic can cause the disk-buffer file to be maximum 1% in size of the maximal size, with having only already processed messages in it. I believe it is not a big tradeoff, when seeing how much time it can save.

I have done measurements with other thresholds:
   1. 50% of max size: took the same time as for 1% (~0.1 seconds)
   2. 0.1%, 0.01%, 0.001%: took nearly the same time as for 1% (~0.1 seconds)
   3. 0.0001%, 0.00001%: took as much as read and write (~4 seconds)

What I observed is that it gets slow, when truncating nearly per msg. Then we reach a point (approximately at every 500th message), where no matter if we truncate at 1% or 50%, it takes the same time.

Thinking about a general config, where the disk buffer is 1GB, its 1% is 10MB, so if we receive smaller then 20KB sized messages, we have the optimal truncate time. Or for a 100MB disk-buffer it is 2KB.
Max 2KB message size and 100MB disk-buffer are common values, so I think the 1% of the max size threshold works most of the time.

Related issue, but does not get fixed by this PR: #3551

### DIFF
--- a/modules/diskq/diskq-config.c
+++ b/modules/diskq/diskq-config.c
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2021 One Identity
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "diskq-config.h"
+#include "cfg.h"
+
+#define MODULE_CONFIG_KEY "disk-buffer"
+
+static void
+disk_queue_config_free(ModuleConfig *s)
+{
+  module_config_free_method(s);
+}
+
+static void
+_set_default_values(DiskQueueConfig *self)
+{
+  self->truncate_size_ratio = 0.01;
+}
+
+DiskQueueConfig *
+disk_queue_config_new(void)
+{
+  DiskQueueConfig *self = g_new0(DiskQueueConfig, 1);
+
+  self->super.free_fn = disk_queue_config_free;
+  _set_default_values(self);
+  return self;
+}
+
+DiskQueueConfig *
+disk_queue_config_get(GlobalConfig *cfg)
+{
+  DiskQueueConfig *dqc = g_hash_table_lookup(cfg->module_config, MODULE_CONFIG_KEY);
+  if (!dqc)
+    {
+      dqc = disk_queue_config_new();
+      g_hash_table_insert(cfg->module_config, g_strdup(MODULE_CONFIG_KEY), dqc);
+    }
+  return dqc;
+}
+
+void
+disk_queue_config_set_truncate_size_ratio(GlobalConfig *cfg, gdouble truncate_size_ratio)
+{
+  DiskQueueConfig *dqc = disk_queue_config_get(cfg);
+  dqc->truncate_size_ratio = truncate_size_ratio;
+}
+
+gdouble
+disk_queue_config_get_truncate_size_ratio(GlobalConfig *cfg)
+{
+  DiskQueueConfig *dqc = disk_queue_config_get(cfg);
+  return dqc->truncate_size_ratio;
+}
--- a/modules/diskq/diskq-config.h
+++ b/modules/diskq/diskq-config.h
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2021 One Identity
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#ifndef DISKQ_CONFIG_H_INCLUDED
+#define DISKQ_CONFIG_H_INCLUDED
+
+#include "module-config.h"
+
+typedef struct _DiskQueueConfig
+{
+  ModuleConfig super;
+  gdouble truncate_size_ratio;
+} DiskQueueConfig;
+
+DiskQueueConfig *disk_queue_config_get(GlobalConfig *cfg);
+void disk_queue_config_set_truncate_size_ratio(GlobalConfig *cfg, gdouble truncate_size_ratio);
+gdouble disk_queue_config_get_truncate_size_ratio(GlobalConfig *cfg);
+
+#endif
+
--- a/modules/diskq/diskq-grammar.ym
+++ b/modules/diskq/diskq-grammar.ym
@@ -39,6 +39,7 @@
 #include "messages.h"
 #include "diskq.h"
 #include "diskq-options.h"
+#include "diskq-config.h"
 
 #include <string.h>
 
@@ -74,6 +75,7 @@ DiskQueueOptions *last_options;
 
 start
 	: LL_CONTEXT_INNER_DEST dest_diskq { YYACCEPT; }
+	| LL_CONTEXT_OPTIONS KW_DISK_BUFFER '(' diskq_global_options ')' { YYACCEPT; }
 	;
 
 dest_diskq
@@ -100,6 +102,15 @@ dest_diskq_option
         | KW_QOUT_SIZE '(' nonnegative_integer ')'       { disk_queue_options_qout_size_set(last_options, $3); }
         | KW_DIR '(' string ')'                          { disk_queue_options_set_dir(last_options, $3); free($3); }
         | KW_TRUNCATE_SIZE_RATIO '(' float_between_0_and_1 ')' { disk_queue_options_set_truncate_size_ratio(last_options, $3); }
+        ;
+
+diskq_global_options
+        : diskq_global_option diskq_global_options
+        |
+        ;
+
+diskq_global_option
+        : KW_TRUNCATE_SIZE_RATIO '(' float_between_0_and_1 ')' { disk_queue_config_set_truncate_size_ratio(configuration, $3); }
         ;
 
 float_between_0_and_1
--- a/modules/diskq/diskq-grammar.ym
+++ b/modules/diskq/diskq-grammar.ym
@@ -22,6 +22,8 @@
  *
  */
 
+%type <fnum> float_between_0_and_1
+
 %code top {
 #pragma GCC diagnostic ignored "-Wswitch-default"
 
@@ -65,6 +67,7 @@ DiskQueueOptions *last_options;
 %token KW_MEM_BUF_SIZE
 %token KW_QOUT_SIZE
 %token KW_DIR
+%token KW_TRUNCATE_SIZE_RATIO
 
 
 %%
@@ -96,6 +99,11 @@ dest_diskq_option
         | KW_DISK_BUF_SIZE '(' nonnegative_integer64 ')' { disk_queue_options_disk_buf_size_set(last_options, $3); }
         | KW_QOUT_SIZE '(' nonnegative_integer ')'       { disk_queue_options_qout_size_set(last_options, $3); }
         | KW_DIR '(' string ')'                          { disk_queue_options_set_dir(last_options, $3); free($3); }
+        | KW_TRUNCATE_SIZE_RATIO '(' float_between_0_and_1 ')' { disk_queue_options_set_truncate_size_ratio(last_options, $3); }
+        ;
+
+float_between_0_and_1
+        : nonnegative_float { CHECK_ERROR(($1 <= 1), @1, "It cannot be bigger, than 1"); }
         ;
 
 /* INCLUDE_RULES */
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -47,8 +47,6 @@
 
 #define PATH_QDISK              PATH_LOCALSTATEDIR
 
-#define TRUNCATE_SIZE_RATIO 0.01
-
 typedef union _QDiskFileHeader
 {
   struct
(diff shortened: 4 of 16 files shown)


kira-syslogng:
Build SUCCESS

Kokan:
This looks promising. The following is written without checking the code, so take these comments with that in mind.
As my understanding all of these measurements were takon on BSD with ZFS. Can you confirm that ?
I would really love to see the same measurements on linux. Kinda curios if the improvements have the same impact on that platform as well.

* The truncation is always an issue from user side (why disk buffer is not initial size without any message). This seems to resurface those questions. As such, it would be nice if dqtool info could state if truncate is not happened as not reached the truncation threshold.
* Have you tinckered with the idea of not truncating at all (this surfaced I think one of the handover) ? Just on the level of performance, if that has significant benefits compared to your threshold solution.
* I would do a little more investigation (measurement) on the different threshold values with different messages size (large messages etc).
* An option should be made to configure this (at least on the level of enable/disable), or threshold value itself. (The option to configure is due to this is a heuristics based on some measurements.)

alltilla:
After talking about this with @Kokan IRL, we have the some thoughts about the thresholding logic:

As these measurements were done outside syslog-ng, we cannot be sure, that it would speed up anything at all inside syslog-ng. A performance measurement is due, which will probably done next week or the week after that.

The threshold logic itself introduces a complexity, which might not worth the benefits, it would be better to aim for a simpler solution.
An alternative solution would be to make a config option, which turns on or off the truncation everytime for the specified disk-buffer (or every disk-buffer if it is set from the global options). Of course by default this would be set to match the previous behavior.

If we turn off the truncation, we get the following behavior: disk-buffers start as minimal, and can become maximal in size at one point, from where it will not shrink ever.
A small sidenote: as the maximal disk-buffer sizes are always calculatable from the config (we do not create disk-buffers on the fly) it would be a misconfiguration if the user's disk could not take all the disk-buffer's maximal size at once.
We can introduce an additional behavior with the option: we can pre-alloc the full disk-buffer at start, if the option is set, which would have another benefit, as the user could test if the configured disk-buffers' maximal size can fit on their disk.

This new option would introduce less complexity to the code and the disk-buffer logic, it could be a validation tool of the configured disk-buffer sizes, and it would have the same, even a bit more performance improvement as the thresholding logic has. The only downside of it would be the big disk-buffer sizes, which are not used 100%, but that should not be a problem, as we would not break the old behavior by default, and the maximal disk-buffers should fit on the disk anyways.

@Kokan Do you have anything to correct or add to these?

alltilla:
After talking about this topic with the team we concluded, that we can turn on the truncate-threshold logic by default with 0.01, and make it configurable with a new option, so the old behavior is achievable, and complete truncation disabling is achievable, too.

kira-syslogng:
Build FAILURE

alltilla:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

MrAnno:
Kira: `022-test-disk-buffer-partial-message-destination-stopped`
This test case expects a certain minimum disk buffer size, which changed with this PR.

I think we should adjust that size expectation.

modules/diskq/diskq-grammar.ym
```diff
@@ -96,6 +101,20 @@ dest_diskq_option
...
+        : diskq_global_option diskq_global_options
+        |
+        ;
+
+diskq_global_option
+        : KW_TRUNCATE_SIZE_RATIO '(' float_between_0_and_1 ')' { disk_queue_config_set_truncate_size_ratio(configuration, $3); }
+        ;
+
+float_between_0_and_1
+        : nonnegative_float { CHECK_ERROR(($1 <= 1), @1, "It cannot be bigger, than 1"); }
```
MrAnno, line 117: Shouldn't we set the result of the rule here? `$$ = `

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -257,6 +258,123 @@ Test(diskq_truncate, test_diskq_truncate_on_push)
...
+
+  cr_assert_eq(qdisk->hdr->read_head, QDISK_RESERVED_SPACE, "Read head was not reset!");
+  cr_assert_eq(qdisk->hdr->write_head, QDISK_RESERVED_SPACE, "Write head was not reset!");
+  cr_assert_eq(qdisk->hdr->backlog_head, QDISK_RESERVED_SPACE, "Backlog head was not reset!");
+}
+
+Test(diskq_truncate, test_diskq_truncate_size_ratio_default)
+{
+  const gint truncate_size_ratio = (gint)(1100000 * 0.01);
+  const gint below_threshold_message_number = truncate_size_ratio / 134;
```
szemere, line 274: As far as I can see, these "magic numbers" are the consequences of `feed_some_messages` which is generating `log_msg_new_empty` messages. And our code is highly rely on the current size of those messages.

Thus, I am not expecting to refactor the whole unit test system to use "sizeof" functions. However I highly recommend to name every constant before it's use. Like you did with the `truncate_size_ratio`.

i.e.: `const gint log_msg_size_dependent_divider = 134;`

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -216,7 +216,8 @@ Test(diskq_truncate, test_diskq_truncate_on_push)
   GString *filename = g_string_new("test_dq_truncate_on_write.rqf");
 
   DiskQueueOptions options;
-  q = _create_reliable_diskqueue(filename->str, &options);
+  q = _create_reliable_diskqueue(filename->str, &options, FALSE);
+  options.truncate_size_ratio = 0.01;
```
gaborznagy, line 220: Is it safe to set a Qdisk option after load-queue?

gaborznagy: By the way, why set it if it's the same as the default value?

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -257,6 +258,123 @@ Test(diskq_truncate, test_diskq_truncate_on_push)
...
+  QDisk *qdisk = ((LogQueueDisk *)q)->qdisk;
+
+  cr_assert_eq(qdisk->hdr->read_head, QDISK_RESERVED_SPACE, "Read head was not reset!");
+  cr_assert_eq(qdisk->hdr->write_head, QDISK_RESERVED_SPACE, "Write head was not reset!");
+  cr_assert_eq(qdisk->hdr->backlog_head, QDISK_RESERVED_SPACE, "Backlog head was not reset!");
+}
+
+Test(diskq_truncate, test_diskq_truncate_size_ratio_default)
+{
+  const gint truncate_size_ratio = (gint)(1100000 * 0.01);
```
gaborznagy, line 273: minor, it's a bit confusing that this is the same name as the option itself: truncate_size_ratio.
The units of the option is percentage, while this constant is in bytes, like disk-buf-size.

please rename it.

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -257,6 +258,123 @@ Test(diskq_truncate, test_diskq_truncate_on_push)
...
+  const gint below_threshold_message_number = truncate_size_ratio / 134;
+  const gint above_threshold_message_number = below_threshold_message_number + 1;
+
+  LogQueue *q;
+  GString *filename = g_string_new("test_dq_truncate_size_ratio_default.rqf");
+
+  DiskQueueOptions options;
+  q = _create_reliable_diskqueue(filename->str, &options, TRUE);
+  options.truncate_size_ratio = 0.01;
```
gaborznagy, line 282: If this is testing the default value, the don't set it please.

Otherwise I see that it's easier from test side if you just set it, rather then get the default value after you've initialized the disk-buffer.

gaborznagy (commented):
Besides some minor stuff, looks good to me! I like the refactors and the disk-buffer context in global config !!

gaborznagy:
One more question, sorry:
Shouldn't we add a compatibility mode with config_version?

modules/diskq/diskq-grammar.ym
alltilla: From https://www.gnu.org/software/bison/manual/html_node/Actions.html:

If you don’t specify an action for a rule, Bison supplies a default: $$ = $1. Thus, the value of the first symbol in the rule becomes the value of the whole rule. Of course, the default action is valid only if the two data types match. There is no meaningful default action for an empty rule; every empty rule must have an explicit action unless the rule’s value does not matter.

So I think it is implicitly set.

MrAnno:
I think a `cfg_is_config_version_older()` check with the older behavior might be a good idea, especially because it turned out this has no performance gain under normal circumstances.

@alltilla What do you think?

This check needs `VERSION_VALUE_3_33`. We can add the check in the version bump PR, in case we need this and don't want to complicate this PR further with version-related stuff.

modules/diskq/qdisk.c
```diff
@@ -203,6 +205,14 @@ qdisk_is_space_avail(QDisk *self, gint at_least)
 static void
 _truncate_file(QDisk *self, off_t expected_size)
 {
+  off_t truncate_threshold = (off_t)(qdisk_get_maximum_size(self) * TRUNCATE_SIZE_RATIO);
+  off_t possible_size_reduction = self->file_size - expected_size;
+
+  if (possible_size_reduction >= 0 && possible_size_reduction < truncate_threshold)
```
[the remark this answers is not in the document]
alltilla: Actually that condition provided that we call `ftruncate()` if we want to extend the file with it.

Though, I agree that it is hard to think through in the current form, so I made small helper functions for the conditions. I hope it will make it more readable.

modules/diskq/qdisk.c
```diff
@@ -203,6 +205,14 @@ qdisk_is_space_avail(QDisk *self, gint at_least)
 static void
 _truncate_file(QDisk *self, off_t expected_size)
 {
+  off_t truncate_threshold = (off_t)(qdisk_get_maximum_size(self) * TRUNCATE_SIZE_RATIO);
```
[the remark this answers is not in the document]
alltilla: Thanks, fixed.

modules/diskq/tests/test_diskq_truncate.c
alltilla: Thanks, fixed.

modules/diskq/tests/test_diskq_truncate.c
alltilla: I changed the order of the setting, and obtained the default value from `DiskQueueConfig`, so these should be fixed now.

modules/diskq/diskq-config.h
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (c) 2021 One Identity
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
```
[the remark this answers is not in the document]
alltilla: Thanks, fixed.

modules/diskq/diskq-config.c
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (c) 2021 One Identity
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
```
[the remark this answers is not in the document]
alltilla: Thanks, fixed.

modules/diskq/tests/test_diskq_truncate.c
alltilla: Thanks, fixed.

modules/diskq/tests/test_diskq_truncate.c
alltilla: I changed the order of the setting, and obtained the default value from `DiskQueueConfig`, so these should be fixed now.

alltilla:
@gaborznagy @MrAnno
I added a commit, which handles the old config version with a warning, and the old truncate behavior.

modules/diskq/diskq-grammar.ym
MrAnno: Thanks for the clarification.

kira-syslogng:
Build SUCCESS

alltilla:
@kira-syslogng test this please test branch=022-use-dqtool;

szemere (approved):
Thank You for all the changes! Waiting for kira.

Barkodcz (approved):
Thank you!

MrAnno:
@gaborznagy Please merge this PR after you're done with your second review round.

kira-syslogng:
Build SUCCESS

gaborznagy:
Thanks for fixing my findings.

PR #3698, opened 2021-06-08
Kafka: remove obsolete options and add some fix up around configuration

gaborznagy:
- `properties-file` option is deprecated
- remove currently deprecated options (that was introduced for an easy transition from the Java-based kafka driver)
- making sure that `bootstrap-servers` option is mandatory
- marking some librdkafka properties as "protected", which means that we don't allow to configure librdkafka properties that have an explicit option in the syslog-ng configuration. For example the `bootstrap-servers()` config option can be overriden by the `bootstrap.servers` and `metadata.broker.list` librdkafka properties (which can be set in the `config()` option or the now deprecated `properties-file()`).
To prevent any misconfiguration, we don't allow to set these properties.
The currently disabled properties:
  - "bootstrap.servers" (would override `bootstrap-servers()` config)
  - "metadata.broker.list (would override `bootstrap-servers()` config)


kira-syslogng:
Build SUCCESS

modules/kafka/tests/test_kafka_topic.c
```diff
@@ -78,6 +78,12 @@ _init_topic_names(LogDriver *driver, const gchar *topic, const gchar *fallback_t
   kafka_dd_set_fallback_topic(driver, fallback_topic);
 }
 
+static void
+_set_bootstrap_server(LogDriver *driver, const gchar *bootstrap_server)
```
Kokan, line 82: This wrapper does not really provide any value. Simply inlining its content imho better.

modules/kafka/tests/test_kafka_config.c
```diff
@@ -0,0 +1,163 @@
...
+  log_pipe_unref(&driver->super);
+}
+
+Test(kafka_config, property_setting_works)
+{
+  LogDriver *driver = kafka_dd_new(configuration);
+  _setup_topic(driver, "default-test-topic");
+  _setup_bootstrap_server(driver, "test-host:9092");
+  _setup_kafka_property(driver, "retries", "1000");
+  _setup_kafka_property(driver, "queue.buffering.max.messages", "65535");
```
Kokan, line 137: Could you also please check with `topic.partitioner`?

modules/kafka/tests/test_kafka_topic.c
gaborznagy: Ok.

modules/kafka/tests/test_kafka_config.c
gaborznagy: Good idea to test if we can override a property that is set by default in KafkaDestDriver.

gaborznagy:
I think I created some inconsistency in the terminology.
I keep saying `protected properties`, however what's protected is the config option.

Also, I could add the syslog-ng config option name too in the warning message, so the user can see what's conflicting with what.
E.g.
```
kafka: protected properties cannot be overriden; config='bootstrap.servers'
```
would change to
```
kafka: protected properties cannot be overriden; config='bootstrap.servers, option=bootstrap-servers()'
```
Maybe it's confusing for the user to differentiate `config` and `option` ....

modules/kafka/kafka-grammar.ym
```diff
@@ -63,9 +63,7 @@
 %token KW_FLUSH_TIMEOUT_ON_SHUTDOWN
 %token KW_FLUSH_TIMEOUT_ON_RELOAD
 %token KW_POLL_TIMEOUT
-%token KW_CLIENT_LIB_DIR
 %token KW_BOOTSTRAP_SERVERS
-%token KW_SYNC_SEND
```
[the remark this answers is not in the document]
Kokan: Thanks.

kira-syslogng:
Build SUCCESS

Kokan (approved):

gaborznagy:
> kafka: protected config properties cannot be overriden; name='bootstrap.servers'

We agreed that the warning message is okay in this form:
```
kafka: protected config properties cannot be overridden; name='bootstrap.servers'
kafka: protected config properties cannot be overridden; name='metadata.broker.list'
```

Kokan (approved):

kira-syslogng:
Build SUCCESS

Barkodcz (approved):

Kokan (approved):

PR #3700, opened 2021-06-09
Fix crash when switching between disk-based and memory queues

MrAnno:
When a disk-buffer is removed from the configuration, and syslog-ng is reloaded, syslog-ng starts acquiring a memory queue instead, but first it fetches and tries to use the previous queue from cfg-persist. The type of the old queue is not validated, so we end up using a different type of queue than what the user configured.

This is not a disaster, this behavior could have been documented, but we also crash when trying to use the old queue, because it was not initialized with the queue's own initialization method (`DiskQ::acquire_queue()`).

This PR fixes the crash by safely closing the disk-buffer queue, and allocating a memory queue instead (exactly what the user wants; and unsent messages remain available in the disk buffer file).

Note: switching from a memory queue to a disk-based queue works the same way, the PR fixes a crash with the diskq-to-memqueue direction.

### DIFF
--- a/lib/driver.c
+++ b/lib/driver.c
@@ -272,6 +272,12 @@ log_dest_driver_acquire_memory_queue(LogDestDriver *self, const gchar *persist_n
   if (persist_name)
     queue = cfg_persist_config_fetch(cfg, persist_name);
 
+  if (queue && !log_queue_has_type(queue, log_queue_fifo_get_type()))
+    {
+      log_queue_unref(queue);
+      queue = NULL;
+    }
+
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

MrAnno:

kira-syslogng:
Build SUCCESS

alltilla (approved):

gaborznagy (approved):

PR #3702, opened 2021-06-12
GSoC 2021 Project: Add regexp-parser()

LittleFish33:
[GSoC2021-Idea-&-Project-list#add-regexp-parser](https://github.com/syslog-ng/syslog-ng/wiki/GSoC2021-Idea-&-Project-list#add-regexp-parser)

This PR introduces a parser that can parse messages with regular expressions.

### Declaration:
```
parser parser_name {
  regexp-parser(
    patterns()
    prefix()
    flags()
  );
};
```
### patterns()
regular expression patterns that you want to find a match, regexp-parser supports multiple patterns
### prefix()
Insert a prefix before the name part of the parsed name-value pairs to help further processing
### flags()
Synopsis: disable-jit, dupnames, global, ignore-case, newline, store-matches, unicode, utf8

### Example
```
parser {
  regexp-parser(
    prefix(".regexp.") 
    patterns("(?<DN>foo)", "(?<DN>ball)")
    flags(ignore-case)
    template("${MESSAGE}")
  );
};
```


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

bazsi:
@kira-syslogng  ok to test

kira-syslogng:
Build SUCCESS

lib/logmatcher.h
```diff
@@ -66,6 +66,7 @@ struct _LogMatcher
   gint ref_cnt;
   gint flags;
   gchar *pattern;
+  gchar *prefix;
```
bazsi, line 69: This should go in the PCRE specific matcher class, that is defined in logmatcher.c and is called LogMatcherPcreRe

lib/logmatcher.c
```diff
@@ -209,6 +209,7 @@ log_matcher_string_new(const LogMatcherOptions *options)
   self->super.compile = log_matcher_string_compile;
   self->super.match = log_matcher_string_match;
   self->super.replace = log_matcher_string_replace;
+  self->super.prefix = NULL;
```
bazsi, line 212: this is not needed if https://github.com/syslog-ng/syslog-ng/pull/3702/files#r650710436 is fixed.

lib/logmatcher.c
```diff
@@ -279,6 +280,7 @@ log_matcher_glob_new(const LogMatcherOptions *options)
   self->super.match = log_matcher_glob_match;
   self->super.replace = NULL;
   self->super.free_fn = log_matcher_glob_free;
+  self->super.prefix = NULL;
```
bazsi, line 283: likewise.

modules/Makefile.am
```diff
@@ -47,6 +47,7 @@ include modules/systemd-journal/Makefile.am
 include modules/system-source/Makefile.am
 include modules/tagsparser/Makefile.am
 include modules/xml/Makefile.am
+include modules/regexp/Makefile.am
```
bazsi, line 50: maybe we should rename the directory to regexp-parser, regexps are used in a lot of ways within syslog-ng and not all of them would concentrate in the "modules/regexp" directory (neither now or long term), so let's try to use a more specific directory.

lib/logmatcher.c
```diff
@@ -445,7 +447,17 @@ log_matcher_pcre_re_feed_named_substrings(LogMatcher *s, LogMessage *msg, int *m
...
+            {
+              GString *formatted_key = g_string_new(self->super.prefix);
+              g_string_append(formatted_key, tabptr + 2);
+              log_msg_set_value_by_name(msg, formatted_key->str, value + begin_index, end_index - begin_index);
+              g_string_free(formatted_key, TRUE);
+            }
+          else
+            {
+              log_msg_set_value_by_name(msg, tabptr + 2, value + begin_index, end_index - begin_index);
+            }
```
bazsi, line 460: This can be a pretty hot path so we need to be aware of (a) performance implications. Also, this code is becoming a bit too complex, so (b) extracting the functionality into a separate function seems to be a good move.

Hints:
(a) g_string_new() is pretty slow, it allocates both a GString header and then the payload, which is potentially reallocated right at the next line. So we have 2-3 memory allocations plus a free operation in a path that is executed multiple times for each message. Message rates can be pretty high with syslog-ng, reaching 100s of thousands per second, although regexps would slow those down to a 10s of thousands. In any case, this is potentially executing 100k per sec. There's an API on top of GStrings that will give you cheap/free allocation of GString instances, look at ```scratch-buffers.h``` and try to find a user of that code for examples.

(b) you could move this part in its entirety into a local/static function, which would make this function a lot easier to read.

modules/regexp/regexp-parser-grammar.ym
```diff
@@ -0,0 +1,87 @@
...
+
+parser_regexp_opts
+	: parser_regexp_opt parser_regexp_opts
+	|
+	;
+
+parser_regexp_opt
+	: { last_matcher_options = regexp_parser_get_matcher_options(last_parser); } matcher_option
+	| KW_PREFIX '(' string ')'		{ regexp_parser_set_prefix(last_parser, $3); free($3); }
+    | KW_PATTERN '(' string ')'		{ regexp_parser_set_pattern(last_parser, $3); free($3); }
```
bazsi, line 81: we should support multiple patterns, just like date-parser() does that with multiple potential formats. If any of the pattern matches we should accept (match) the message.

If none of them matches we should just return "unmatched", just as date-parser() does this when none of the formats match.

modules/regexp/regexp-parser.c
```diff
@@ -0,0 +1,175 @@
...
+  return formatted_key->str;
+}
+
+static const gchar *
+_get_formatted_key(RegexpParser *self, const gchar *key, GString *formatted_key)
+{
+  if (!self->prefix)
+    return key;
+  return _get_formatted_key_with_prefix(self, key, formatted_key);
+}
```
bazsi, line 110: these functions do not seem to be used.

lib/logmatcher.h
bazsi, line 69: I can't see how this prefix gets initialized.

LittleFish33: Thanks! But I have a question. If I define the `prefix` in the `LogMatcherPcreRe` class, I can't set the prefix in the `regexp-parser.c` because the definition of `LogMatcherPcreRe` is in the `logmatcher.c`. I can think of two ways to solve this problem:

1. move the definition of `LogMatcherPcreRe` from `logmatcher.c` to `logmatcher.h` (but may be I should also move the definitions of `LogMatcherGlob` and `LogMatcherString` to  `logmatcher.h` for consistency)
2. declare a new function called `log_matcher_pcre_set_prefix` in `logmatcher.h`, which can set the prefix of a LogMatcherPcreRe object

bazsi:
On Mon, Jun 14, 2021 at 5:13 PM LittleFish ***@***.***> wrote:

> *@LittleFish33* commented on this pull request.
>
> In lib/logmatcher.h
> <https://github.com/syslog-ng/syslog-ng/pull/3702#discussion_r651039461>:
>
> > @@ -66,6 +66,7 @@ struct _LogMatcher
>    gint ref_cnt;
>    gint flags;
>    gchar *pattern;
> +  gchar *prefix;
>
> Thanks! But I have a question. If I define the prefix in the
> LogMatcherPcreRe class, I can't set the prefix in the regexp-parser.c
> because the definition of LogMatcherPcreRe is in the logmatcher.c. I can
> think of two ways to solve this problem:
>
>    1. move the definition of LogMatcherPcreRe from logmatcher.c to
>    logmatcher.h (but may be I should also move the definitions of
>    LogMatcherGlob and LogMatcherString to logmatcher.h for consistency)
>    2. declare a new function called log_matcher_pcre_set_prefix in
>    logmatcher.h, which can set the prefix of a LogMatcherPcreRe object
>
>
>
the 2nd looks good. we usually do a cast in the impementation function in
these cases and take a pointer to the parent class, as it is simpler in the
grammar, which uses a single variable to represent any LogMatcher instance.

-- 
Bazsi

kira-syslogng:
Build SUCCESS

lib/logmatcher.c
```diff
@@ -435,6 +458,7 @@ log_matcher_pcre_re_feed_named_substrings(LogMatcher *s, LogMessage *msg, int *m
       /* Now we can scan the table and, for each entry, print the number, the name,
          and the substring itself.
        */
+      GString *formatted_name = scratch_buffers_alloc();
```
bazsi, line 461: this allocation can move inside the log_matcher_pcre_re_feed_value_by_name() function, so you don't have to pass it as an argument.

lib/logmatcher.c
```diff
@@ -414,6 +417,26 @@ log_matcher_pcre_re_feed_backrefs(LogMatcher *s, LogMessage *msg, gint value_han
...
 }
 
+static inline void
+log_matcher_pcre_re_feed_value_by_name(LogMatcherPcreRe *s, LogMessage *msg, GString *formatted_name, gchar *tabptr,
+                                       const gchar *value, gint begin_index, gint end_index)
+{
+  if(s->prefix != NULL)
+    {
+      if (formatted_name->len > 0)
+        g_string_truncate(formatted_name, s->prefix_len);
```
bazsi, line 427: formatted_name can be a newly allocated GString instance, so truncating may not produce the required prefix. you will need to assign a new value every time.

bazsi: I just recognized that there's a similar pattern in kv-parser(), which seems to be buggy as well. The scratch-buffers allocator does not guarantee that we get the same GString instance every time (though it may happen). Can you fix that part too?

modules/regexp-parser/regexp-parser.c
```diff
@@ -0,0 +1,170 @@
...
+{
+  RegexpParser *self = (RegexpParser *) s;
+
+  log_matcher_options_init(&self->matcher_options);
+  for (GList *item = self->patterns; item; item = item->next)
+    {
+      self->matchers = g_list_prepend(self->matchers, log_matcher_new(&self->matcher_options));
+      log_matcher_pcre_set_prefix((LogMatcher *)self->matchers->data, self->prefix);
+
+      if(!log_matcher_compile((LogMatcher *)self->matchers->data, item->data, NULL))
```
bazsi, line 81: do we handle correctly the case where the 3rd pattern fails to compile? any memory leaks?

modules/regexp-parser/regexp-parser.c
```diff
@@ -0,0 +1,170 @@
...
+            evt_tag_printf("msg", "%p", *pmsg));
+
+  gboolean result = FALSE;
+  for (GList *item = self->matchers; item; item = item->next)
+    {
+      msg_trace("regexp-parser message processing for",
+                evt_tag_str ("input", input),
+                evt_tag_str ("pattern", ((LogMatcher *)item->data)->pattern));
+
+      if (log_matcher_match((LogMatcher *)item->data, *pmsg, LM_V_NONE, input, input_len))
```
bazsi, line 108: although it's a bit of a layering violation, we may actually know that we are matching against LM_V_MESSAGE (in case the template() option is unset). And with that knowledge the LogMatcher instance can store indirect matches instead of full copies.

So instead of LM_V_NONE, you could use a conditional to see if ```self->super.template``` is set, and use LM_V_NONE if it is, and LM_V_MESSAGE if it's not.

This would result in the use of "indirect references" within the LogMessage, meaning that slices of the MESSAGE payload would be used as values. This might compress the size of the serialized log message and also reduce CPU usage as the new name-value pair is stored.

modules/regexp-parser/regexp-parser.c
```diff
@@ -0,0 +1,170 @@
...
+
+  for (GList *item = self->matchers; item; item = item->next)
+    {
+      log_matcher_unref((LogMatcher *)item->data);
+
+    }
+  if (self->matchers)
+    {
+      g_list_free(self->matchers);
+    }
```
bazsi, line 132: g_list_free_full() allows you to pass a GDestroyNotify callback.

bazsi (commented):
This is good progress. A few more nitpicks added as comments. I was especially happy about the unit test.

Can you perhaps add a Python based functional test too? (tests/python_functional) it is a pytest based test suite. It's not a complete test suite yet, but it is the preferred place for functional tests.

You should find examples in that directory and @mitzkia can also help.

bazsi:
@LittleFish33 any plans on when you start working on my comments? thanks.

LittleFish33:
@bazsi Sorry for the long delay, I will continue to work on this tomorrow

lib/logmatcher.c
LittleFish33: > formatted_name can be a newly allocated GString instance, so truncating may not produce the required prefix. you will need to assign a new value every time.

Hi, bazsi, I have a question. In the `scratch_buffers_alloc` function, the allocated buffer will be truncated to 0.
```c++
// lib/scratch-buffers.c, lines 127-128
GString *buffer = g_ptr_array_index(scratch_buffers, scratch_buffers_used);
g_string_truncate(buffer, 0);
```
In this case, if formatted_name is a newly allocated GString instance, it will go to the else case and call `g_string_assign(formatted_name, s->prefix);`.

modules/regexp-parser/regexp-parser.c
LittleFish33: @bazsi I'm a little confused about this problem. Does the memory leak here mean that the space allocated by `log_matcher_new` has not been released? But if any pattern fails to compile, the program will terminal directly and print the error on the console.

lib/logmatcher.c
bazsi: you are right. I missed the loop on the regexp values, with scratch_buffers_alloc() before the loop it all makes sense now.

sorry.

modules/regexp-parser/regexp-parser.c
bazsi: I just wanted to confirm that if the loop exits early (because of log_matcher_compile() returning failure), no memory leaks remain.

It is true that the exiting process would free all allocations, but we should ensure that all internal objects do that on their own without relying on the process to exit.

In any case, just make sure that log_matcher_free() is properly called for the partially populated list too.

kira-syslogng:
Build SUCCESS

LittleFish33:
@bazsi I have updated the commit and added a Python based functional test. Thanks for your comments.

bazsi:
I'll check it out in the coming days.

bazsi:
Can you pls rebase to resolve the conflict? Thanks

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

LittleFish33:
I have resolved the conflict, but some checks were not successful. It shows that `Kira-starter — Build FAILURE`. But I just resolved the conflict by replacing
```
modules/(afamqp|affile|afmongodb|afprog|afsmtp|afsocket|afsql|afstomp|afstreams|afuser|azure-auth-header|basicfuncs|cef|confgen|cryptofuncs|csvparser|timestamp|diskq|dbparser|geoip2|graphite|json|kvformat|linux-kmsg-format|pacctformat|pseudofile|python|redis|riemann|syslogformat|systemd-journal|getent|system-source|stardate|snmptrapd-parser|xml|openbsd|examples|kafka|afsnmp|mqtt|[^/]*$)
```
with
```
modules/(afamqp|affile|afmongodb|afprog|afsmtp|afsocket|afsql|afstomp|afstreams|afuser|azure-auth-header|basicfuncs|cef|confgen|cryptofuncs|csvparser|timestamp|diskq|dbparser|geoip2|graphite|json|kvformat|linux-kmsg-format|pacctformat|pseudofile|python|redis|riemann|syslogformat|systemd-journal|getent|system-source|stardate|snmptrapd-parser|xml|openbsd|examples|kafka|afsnmp|mqtt|regexp-parser|[^/]*$)
```
, and the last check was successful. And it seems that the console output of `Kira-starter` doesn't explain why it failed.
Any help will be appreciated. Thank you.

Kokan:
@LittleFish33 I've checked, the Kira failure is not due to your changes. Actually there are some cases on master that's incorrect. You cannot and shouldn't do anything regarding that, once fixed we'll notify you.

Kokan:

kira-syslogng:
Build SUCCESS

mitzkia:

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
```diff
@@ -0,0 +1,56 @@
...
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+import pytest
+
+
+test_parameters_raw = [
+    ("foo", '''""''', '''"(?<DN>foo)"''', "", "${DN}", True, "foo"),
```
mitzkia, line 28: First of all thanks for the Light testcases.
My comment for different testcases: 
- Can you please add a more basic testcase, when 'patterns' is a simple pattern match without any reference saving: \<DN\>. In this case you can use a custom prefix value to store the result.
- Can you please give a better name for DN. If I am right it is just custom macro name
- Could you please give a testcase for this statement also? "use multiple regular expressions, these should be processed in order, the first successfully matching one causes the processing to stop"
- Could you please add a testcase for the following: "Error parsing regexp-parser, compiling the regexp failed", in this case syslog-ng should not started

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
```diff
@@ -0,0 +1,56 @@
...
+    ("abc", '''""''', '''"(?<DN>Abc)"''', "ignore-case", "${DN}", True, "abc"),
+]
+
+
+@pytest.mark.parametrize(
+    "input_message, prefix, patterns, flags, template, expected_result, expected_value", test_parameters_raw,
+    ids=list(map(str, range(len(test_parameters_raw)))),
+)
+def test_regexp_parser(config, syslog_ng, input_message, prefix, patterns, template, flags, expected_result, expected_value):
+    config.add_include("scl.conf")
```
mitzkia, line 43: This testcase does not require scl.conf

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
LittleFish33: Thank you for your comments, I will work on this soon.

modules/regexp-parser/regexp-parser.c
```diff
@@ -0,0 +1,178 @@
...
+
+  log_matcher_options_init(&self->matcher_options);
+  gboolean result = TRUE;
+
+  for (GList *item = self->patterns; item; item = item->next)
+    {
+      self->matchers = g_list_prepend(self->matchers, log_matcher_new(&self->matcher_options));
+      log_matcher_pcre_set_prefix((LogMatcher *)self->matchers->data, self->prefix);
+
+      if(!log_matcher_compile((LogMatcher *)self->matchers->data, item->data, NULL))
```
bazsi, line 83: Please handle the error in a more user-friendly way, I got no reason why the compilation of my regexp failed:

Error parsing regexp-parser, compiling the regexp failed in etc/syslog-ng-re.conf:5:11-5:61:
1       @version: 3.32
2       
3       log { 
4       	source { tcp(port(2000)); };
5-----> 	parser { regexp-parser(patterns("(?foo)a+") prefix(".re.")); };
5-----> 	         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
6       	destination { file("/tmp/log" template("$(format-json *)\n")); };
7       };

bazsi:
While testing the regexp-parser() plugin, I've encountered that it applies all regexps to the input string, regardless if they matched or not, the original GSoC specification contained this point: 

> use multiple regular expressions, these should be processed in order, the first successfully matching one causes the  processing to stop

I was using this config:
```
@version: 3.32

log { 
	source { tcp(port(2000)); };
	parser { regexp-parser(patterns("(?<foo>)a+", "(?<bar>)b+") prefix(".re.")); };
	destination { file("/tmp/log" template("$(format-json *)\n")); };
};
```
Albeit applying all of them could make sense in some cases, the general idea is that you specify multiple regexps to support multiple incarnations of the _same_ meaning, potentially in some variations. In this case, it is expected that you list the alternative regexps to parse the same thing. And in this case it would make sense to quickly break out of the matching loop the first time it matches, expecting the patterns to be ordered by their likelyhood of being encountered.

Please either break out at the first match or add an option to short-circuit at the first matching regexp and make it the default behavior.

thanks

bazsi:
I have resolved most of the open questions, a few still remained. I manually tested the functionality and it works great. I have one issue pointed out in one of my latest comments, but it's very close to become merge-able.

@alltilla could you also make a review so we have two reviews to get it in?

Could you fix the remaining issues?

thanks

modules/regexp-parser/regexp-parser-grammar.ym
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright (c) 2021 Balabit
```
alltilla, line 2: Please change "Balabit" to "One Identity" here, and in every file you have created.

modules/regexp-parser/regexp-parser.c
```diff
@@ -0,0 +1,178 @@
...
+}
+
+static void
+regexp_parser_free(LogPipe *s)
+{
+  RegexpParser *self = (RegexpParser *) s;
+
+  g_free(self->prefix);
+  string_list_free(self->patterns);
+  log_parser_free_method(s);
```
alltilla, line 139: Please make this the last call in this function.

alltilla (commented):
I agree with Bazsi about the early return in `process()`. Please also add unit tests that make sure we do not store matches after the first matching pattern.

---

Please add a news file entry. See `news/README.md`.

---

Please make the commit message(s) more verbose.
Tell about the new module, and show an example config snippet.

---

Also it would have been better from a reviewer's and maintainer's perspective, if this code change was split into some smaller commits:
 * Adding `prefix` to `LogMatcher`.
 * Adding the skeleton of the new module `regexp-parser`.
 * Adding the implementation of `regexp-parser`.
 * Adding unit tests.
 * Adding functional tests.
 * Adding `mod-regexp-parser` to the packages.
 * Adding news file.

Please do so.

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
```diff
@@ -0,0 +1,56 @@
...
+    ("foo", '''".reg."''', '''"(?<DN>foo)"''', "", "${DN}", True, ""),
+    ("foo", '''".reg."''', '''"(?<DN>foo)|(?<DN>bar)"''', "dupnames", "${.reg.DN}", True, "foo"),
+    ("abc", '''""''', '''"(?<DN>Abc)"''', "", "${DN}", False, ""),
+    ("abc", '''""''', '''"(?<DN>Abc)"''', "ignore-case", "${DN}", True, "abc"),
+]
+
+
+@pytest.mark.parametrize(
+    "input_message, prefix, patterns, flags, template, expected_result, expected_value", test_parameters_raw,
+    ids=list(map(str, range(len(test_parameters_raw)))),
```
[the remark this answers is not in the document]
LittleFish33: Thanks, I have added `ids` for each test case in the latest commits.

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
LittleFish33: I have updated the commits with the following changes:
* A new test case has been added, whose `id` is `match_without_reference_saving`
* I change the reference name to `<key>`
* A new test case has been added, whose `id` is `match_multiple_regular_expressions`
* A new test case has been added, whose `id` is `regular_expression_compile_error`
* A new test case has been added, whose `id` is `check_MSG_macro`
* A new test case has been added, whose `id` is `unmatch_case`. I have made it available to use 'DriverStatsHandler' in Parser class and extended the API in 'DriverStatsHandler' for parsers. The details can be found in commits [a16d34](https://github.com/syslog-ng/syslog-ng/pull/3702/commits/a16d34f3e3fef3840c040bab565834b3fad34925) and [4b2b90](https://github.com/syslog-ng/syslog-ng/pull/3702/commits/4b2b90a2548537a690cd0dba78e06b6bb7634334)

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
LittleFish33: `config.add_include("scl.conf")` has been removed.

modules/regexp-parser/regexp-parser.c
LittleFish33: I have updated the `regexp-parser-grammar.ym` file and the `regexp_parser_compile` function. In the latest commits, syslog-ng will output the error with the failure reason:
```
Error parsing regexp-parser, compiling the regexp failed, error=Failed to compile PCRE expression >>>(?foo)a+<<< `unrecognized character after (? or (?-' at character 2 in /etc/syslog-ng.conf:57:11-57:61:
55      log { 
56      	source(s_file);
57----> 	parser { regexp-parser(patterns("(?foo)a+") prefix(".re.")); };
57----> 	         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
58      	destination { file("/tmp/log" template("$(format-json *)\n")); };
59      };
```

lib/logmatcher.c
```diff
@@ -291,6 +292,8 @@ typedef struct _LogMatcherPcreRe
   pcre *pattern;
   pcre_extra *extra;
   gint match_options;
+  gchar *prefix;
+  gint prefix_len;
```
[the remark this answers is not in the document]
LittleFish33: I have changed the `prefix` and `prefix_len` to `nv_prefix` and `nv_prefix_len`.

modules/regexp-parser/regexp-parser-grammar.ym
LittleFish33: Related changes have been udpated.

modules/regexp-parser/regexp-parser.c
LittleFish33: Related changes have been udpated.

LittleFish33:
> I agree with Bazsi about the early return in `process()`. Please also add unit tests that make sure we do not store matches after the first matching.

I have updated the code and added a new test case for this change. The test case `id` is `match_multiple_regular_expressions`.

bazsi:
Thanks for the updates. I will check them out. Hopefully this is the last
iteration.

alltilla (dismissed):
I only have 1 optional review note. Approve.

kira-syslogng:
Build SUCCESS

modules/regexp-parser/regexp-parser.c
```diff
@@ -0,0 +1,175 @@
...
+  for (GList *item = self->matchers; item; item = item->next)
+    {
+      msg_trace("regexp-parser message processing for",
+                evt_tag_str ("input", input),
+                evt_tag_str ("pattern", ((LogMatcher *)item->data)->pattern));
+
+      if ((G_LIKELY(!self->super.template)
+           && log_matcher_match((LogMatcher *)item->data, *pmsg, LM_V_MESSAGE, input, input_len)) ||
+          (!G_LIKELY(!self->super.template)
+           && log_matcher_match((LogMatcher *)item->data, *pmsg, LM_V_NONE, input, input_len)))
```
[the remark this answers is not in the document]
LittleFish33: Thank you for your comments. I've updated the code as suggested.

alltilla (approved):
Thanks!

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
mitzkia: Thanks for the changes, I will check it soon.

mitzkia: Thanks for the changes again, they are very good. I have only a few minor review comments.

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
```diff
@@ -0,0 +1,67 @@
...
+@pytest.mark.parametrize(
+    "input_message, prefix, patterns, flags, template, compile_result, expected_result, expected_value", test_parameters_raw,
+    ids=[
+        "match_literally", "match_regular_expression", "match_without_reference_saving",
+        "match_with_prefix", "match_with_prefix_and_wrong_template", "match_with_dupnames", "unmatch_case",
+        "match_ignore_case_flag", "match_multiple_regular_expressions", "regular_expression_compile_error",
+        "check_MSG_macro",
+    ],
+)
+def test_regexp_parser(config, syslog_ng, syslog_ng_ctl, input_message, prefix, patterns, flags, template, compile_result, expected_result, expected_value):
```
mitzkia, line 51: Thanks for the changes, they are very good.
My comment here: syslog_ng_ctl fixture is not used in the testcase, it can be deleted from testcase parameter list

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
```diff
@@ -0,0 +1,67 @@
...
+    config.update_global_options(stats_level=1)
+    generator_source = config.create_example_msg_generator_source(num=1, template=config.stringify(input_message))
+    regexp_parser = config.create_regexp_parser(prefix=prefix, patterns=patterns, flags=flags)
+
+    file_destination = config.create_file_destination(file_name="output.log", template=config.stringify(template + "\n"))
+    config.create_logpath(statements=[generator_source, regexp_parser, file_destination])
+
+    if compile_result is True:
```
mitzkia, line 59: It is enough to use: 'if compile_result:'

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
```diff
@@ -0,0 +1,67 @@
...
+        syslog_ng.start(config)
+        if expected_result:
+            assert file_destination.read_log().strip() == expected_value
```
mitzkia, line 62: Maybe we can add 'assert regexp_parser.get_query() == {'discarded': 0}' line after 'assert file_destination.read_log().strip() == expected_value'

mitzkia (commented):
Thanks for the changes. 
Other than my minor review comments I am approving functional testcases for Light

kira-syslogng:
Build SUCCESS

tests/python_functional/functional_tests/parsers/regexp-parser/test_regexp_parser.py
LittleFish33: Thanks for your comments, I've updated the code.

mitzkia:
Thanks for the functional testcases.
From my point of view functional Light testcases are approved.

bazsi (approved):
After doing the review iterations and fixing the notes (thank you!) I am happy with this PR, so approve from me.

alltilla (approved):

PR #3705, opened 2021-06-17
dbld: remove OBS dependencies from the ubuntu-bionic image

MrAnno:
As the unofficial OBS repository is discontinued, I checked what depended on it in our CI and dbld.

- Ubuntu Bionic has librdkafka <1.0.0, so we added librdkafka >1.0 from OBS because that is our min. required version and we wanted a full feature set in our devshell image. The devshell image is now based on Debian Testing, therefore this workaround is not needed anymore. Note: The kafka-c destination is no longer built/packaged on Ubuntu Bionic.

- libmongoc had a CMake-only build issue in v1.9.2, thus we used v1.15.0 instead. The devshell image is now based on Debian Testing, so this is not needed anymore. This has no impact on the MongoDB destination's packaging.

The ubuntu-bionic was the last user of `add_obs_repo()` in dbld, so it has been removed as well.

### DIFF
--- a/dbld/build.manifest
+++ b/dbld/build.manifest
@@ -32,9 +32,9 @@ debian-buster		python2,nojava,nokafka
 debian-testing		python3,nojava
 debian-sid		python3,nojava
 
-# on ubuntu, we fetch java from OBS and start using Python3 at focal onwards.
+# on ubuntu, we start using Python3 at focal onwards.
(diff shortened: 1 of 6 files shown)


kira-syslogng:
Build SUCCESS

Kokan (approved):

gaborznagy (approved):

PR #3706, opened 2021-06-17
kafka: minimum version is changed to 1.1.0 from 1.0.0

gaborznagy:
We use the rd_kafka_conf() function in unit tests, which is only available since 1.1.0 release [1].
I was unable to find a better alternative to rd_kafka_conf(), and I don't think that we should turn off the unit tests on older librdkafka versions.

I think the version bump will not have a great impact on users, as librdkafka 1.0.X versions only packaged on 2 distributions. [2]
On these 2 distributions (Alpine Linux 3.10 and Funto 1.4), the packaged syslog-ng's version doesn't have kafka-c (3.19.1 and 3.20.1, while kafka is added in 3.21.1), so syslog-ng packaging is not going to be broken on these distros.

Every other distribution is either older than our current 1.0.0 min version (e.g. Debian Buster is 0.11.6), or greater than 1.1.0 (e.g. Ubuntu Focal LTS ships 1.2.1 version).

[1] https://github.com/edenhill/librdkafka/releases/tag/v1.1.0
[2] https://repology.org/project/librdkafka/versions


kira-syslogng:
Build SUCCESS

Kokan (approved):

MrAnno (approved):

PR #3707, opened 2021-06-17
Stats split response

lbudai:
## Motivation

When there are a lots of stats counters, getting these counters via `syslog-ng-ctl 
stats` is a challenge.
Before #3387 it was not even possible to stop (gracefully) syslog-ng when such a request was in progress.
Now, we can stop syslog-ng - but at the price of calling a `pthread_cancel` which can sometimes cause problems (-> and in some circumstances syslog-ng still cannot be stopped... when stats-mutex is locked when cancel is received... we could play with setting pthread cancel type to `PTHREAD_CANCEL_DEFERRED` and disable/enable cancel state - but that s*cks).

The aim of this PR is to solve two issues:
  * get rid of `pthread_cancel` by implementing 'cancellable' stats queries
  * decrease the required memory for serving a syslog-ng-ctl request (and send back partial results to the syslog-ng-ctl)

Before this patchset we collected all the responses into one giant GString object -> until all the counters are not visited we didn't send anything back to the `syslog-ng-ctl`.
  
*  notes
     * this work can be continued with improving `syslog-ng-ctl query`.
     * this PR is a kind of PoC - just wanted to share the problem and my code (as a kind of farewell PR :) )
 
 ## What's in the PR?

### batches

 * stats-csv interface changed: instead of returning a GString object that contains all the counters, I've added a callback which is called for all CSV record (->for all counter). 
 * functionality is reimplemented with the new interface (on a CSV record callback basis)
 * new methods are added for supporting batches : one is for adding a batch to the current 'session', other is for closing the batch (session) by adding a '.\n' to the end
 * syslog-ng-ctl also modified to handle response 'streams' (before this PR, response was appended to a GString object until '\n.\n' received, now it prints responses at reception time)
 * as `syslog-ng-ctl stats` handler is run in a separete thread, some synchronization is required:
    * batches are inserted into a thread safe GQueue
    * an iv_event is posted when a batch is sent from the producer thread (-> direct manipulation of ivykis objects from a different thread is not allowed )

### eliminate `pthread_cancel`

 * store cancelled state (could be an atomic variable, but I think it's okay to use a non-atomic variable) at ControlServer object
 * implement cancallable stats queries (-> g_hash_table_foreach replaced with an iterator-loop which can checks the `cancelled` state and break the loop when needed
 * thread id saving also removed (thread id needed by `thread_cancel`)
 
Some parts are implemented at night, so be careful with the review ;)
_Also note that I'm not sure if I'll have the time for fixing review notes - in this case you can close my PR, or if you find it useful, you can freely modify to make it merge-ready._

### DIFF
--- a/lib/control/control-server-unix.c
+++ b/lib/control/control-server-unix.c
@@ -75,6 +75,13 @@ control_connection_unix_stop_watches(ControlConnection *s)
   iv_fd_unregister(&self->control_io);
 }
 
+static gboolean
+_pending_output(ControlConnection *s)
+{
+  return (s->output_buffer && s->output_buffer->len > s->pos) ||
+         !g_queue_is_empty(s->response_batches);
+}
+
 static void
 control_connection_unix_update_watches(ControlConnection *s)
 {
@@ -85,7 +92,7 @@ control_connection_unix_update_watches(ControlConnection *s)
       iv_fd_set_handler_out(&self->control_io, NULL);
       iv_fd_set_handler_in(&self->control_io, NULL);
     }
-  else if (s->output_buffer->len > s->pos)
+  else if (_pending_output(s))
     {
       iv_fd_set_handler_out(&self->control_io, s->handle_output);
       iv_fd_set_handler_in(&self->control_io, NULL);
--- a/lib/control/control-server-unix.c
+++ b/lib/control/control-server-unix.c
@@ -78,8 +78,11 @@ control_connection_unix_stop_watches(ControlConnection *s)
 static gboolean
 _pending_output(ControlConnection *s)
 {
-  return (s->output_buffer && s->output_buffer->len > s->pos) ||
-         !g_queue_is_empty(s->response_batches);
+  g_mutex_lock(s->response_batches_lock);
+  gboolean result = (s->output_buffer && s->output_buffer->len > s->pos) ||
+                    !g_queue_is_empty(s->response_batches);
+  g_mutex_unlock(s->response_batches_lock);
+  return result;
 }
 
 static void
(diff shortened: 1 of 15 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

MrAnno:
Hi,

Sorry for the huge delay, and thank you for opening this PR.

This looks really good in my opinion.
Getting rid of `thread_cancel()` calls with cancellable queries sounds amazing.

The speed of `query get` and `stats csv` has been improved drastically in the past, but reducing its memory footprint can be useful as well. For example, when a lot of macroed/wildcard sources/destinations are used, the output can be huge (I don't think we could ever reach the gigabyte limit, but I've seen ~100MB once). I've also noticed that the allocator in newer glibc versions behaves really "lazily" nowadays, so a short-term peak in syslog-ng's memory usage can be seen in a `ps` output for days.

I'm going to review the PR thoroughly in the next 1-2 weeks.
We will resolve the conflicts in `control-server` and `syslog-ng-ctl` that we caused, fix everything we find, and hopefully, the PR will be merged after the v3.35 release.

lbudai:
> 
> 
> Hi,
> 
> Sorry for the huge delay, and thank you for opening this PR.
> 
> This looks really good in my opinion. Getting rid of `thread_cancel()` calls with cancellable queries sounds amazing.
> 
> The speed of `query get` and `stats csv` has been improved drastically in the past, but reducing its memory footprint can be useful as well. For example, when a lot of macroed/wildcard sources/destinations are used, the output can be huge (I don't think we could ever reach the gigabyte limit, but I've seen ~100MB once). I've also noticed that the allocator in newer glibc versions behaves really "lazily" nowadays, so a short-term peak in syslog-ng's memory usage can be seen in a `ps` output for days.
> 
> I'm going to review the PR thoroughly in the next 1-2 weeks. We will resolve the conflicts in `control-server` and `syslog-ng-ctl` that we caused, fix everything we find, and hopefully, the PR will be merged after the v3.35 release.

Hi!

if I'll have the time, I'll try to resolve the conflicts on this weekend - just for checking what did you change :)
I guess you have enough task to do, so it's really up to you when you merge the PR.

thanks,
L.

bazsi:
closing in favour of #3844

PR #3710, opened 2021-06-23
Added macOS System-Source

yashmathne:
The system-source() source driver for syslog-ng automatically identifies the system and then links the appropriate source to read the system and kernel logs from. 
On macOS, since macOS version 10.8, there is no separate kernel log file, instead both system logs and kernel logs get logged onto a file named system.log. ( `/var/log/system.log` ). 
I have modified the `system-source.c` file to add a file driver to this file using a function already defined for creating files drivers for the different systems in accordance with the system-source() [documentation](https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.30/administration-guide/27#TOPIC-1594950).

The application logs are excluded since application logs are highly dependent on what applications you run and shouldn't be in the standard system source.


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

Kokan:
@kira-syslogng ok to test

MrAnno:
Hi,

Reading this regular text file (`/var/log/system.log`) can be a good workaround for macOS, but the original intention with our `system()` source is to replace the given system's syslog implementation and receive messages directly, or in case that's not possible (for example with systemd-journal), just read system logs in a native manner.

In this specific case, `/var/log/system.log` is a filtered textual _output_ of Apple's syslogd.

AFAIK macOS uses its own structured/binary logging mechanism called ASL, which is a replacement for the standard `syslog()` interface. They seem to provide read/query/search methods as well:
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/asl.3.html
They seem to deprecate this API too in favor of the higher level `OSLog` facility:
https://developer.apple.com/documentation/oslog

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

MrAnno:
Closing in favor of #3908.

Thanks again for opening this PR. I'll test the patch on the latest macOS version and hopefully, it can be part of syslog-ng v3.36.1.

HofiOne:
Native macOS system log source added in https://github.com/syslog-ng/syslog-ng/pull/4423

Issue #3709, opened 2021-06-21
python 3.10 not detected


czanik:
While trying to compile the latest syslog-ng git master snapshot on ```Fedora Rawhide```, detection of Python failed. It seems, that they upgraded to ```Python 3.10```, and ./configure cannot detect it:
```
checking for libsystemd >= 245... yes
checking for 3.10-embed... no
checking for 3.10... no
configure: error: Could not find the requested Python development libraries
```
```sudo``` resolved a similar problem recently

Barkodcz:
Hi,

created a bugfix for this issue. Until then can be solve the issue by using "--with-python=python-3.10" instead of "--with-python=3.10". After the PR is merged can be used agian "--with-python=3.10".

MrAnno:
Fixed by #3713.

PR #3713, opened 2021-06-25
bugfix/python 3.10 not detected FIX

Barkodcz:
The problem was the new python 3.10 isn't compatible with previous "python finding script", because it check is only 1 digit or "<digit>.<digit>" and I added a new condition to check "<digit>.<digit><digit>".

This bug can be easily caused. Only need a python 3.10 version and run the configure with this flag: "--with-python=3.10" it will not find the 3.10, but with the bugfix it will found.

Other possible solution:
- using if-else instead of switch-case, because in the if statement can be put regex expression.
- using the configure "if-else", because there can put regex expression also. 

issue: #3709

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1455,6 +1455,9 @@ if test "x$enable_python" != "xno"; then
       [[0-9]].[[0-9]])
           with_python="python-${with_python}"
           ;;
+      [[0-9]].[[0-9][0-9]])
+          with_python="python-${with_python}"
+          ;;
     esac
 
     PKG_CHECK_MODULES(PYTHON, $with_python-embed, python_found="yes", [
--- a/news/packaging-3713.md
+++ b/news/packaging-3713.md
@@ -1,0 +1,1 @@
+configure: now supporting python with two digit minor version


kira-syslogng:
Build SUCCESS

news/bugfix-3713.md
```diff
@@ -0,0 +1,10 @@
+The problem was the new python 3.10 isn't compatible with previous "python finding script", because it check is only 1 digit or "<digit>.<digit>" and I added a new condition to check "<digit>.<digit><digit>".
+
+This bug can be easily caused. Only need a python 3.10 version and run the configure with this flag: "--with-python=3.10" it will not find the 3.10, but with the bugfix it will found.
+
+Other possible solution:
+
+- using if-else instead of switch-case, because in the if statement can be put regex expression.
+- using the configure "if-else", because there can put regex expression also.
+
+issue: #3709
```
Kokan, line 10: This si going to be part of the news file (an example: https://github.com/syslog-ng/syslog-ng/releases/tag/syslog-ng-3.32.1), if you check out that the description are brief (specially in case of simple bugfix). This description is nice for the PR description, but too detailed for a news entry.

From user point of view the implementation details are not interesting (still they can uncover it as there is a reference to the original PR), I would rather include the fact that configure cannot accept python 3.10 version, and now it does.

gaborznagy (commented):
Looks good to me. 
There is a failing macos CI job. 
I think it's some intermittent Homebrew issue.

Kokan (approved):

kira-syslogng:
Build SUCCESS

alltilla (approved):

gaborznagy (approved):

kira-syslogng:
Build SUCCESS

czanik:
Thanks! Tested by compiling a recent git master snapshot on Fedora Rawhide, and it works fine.

PR #3715, opened 2021-06-30
Fix dbld devshell dbgsyms

Barkodcz:
When devshell rebased on syslog-ng-tarball image (which is based on debian-testing), the enable_dbgsyms improved to be able to install debug symbols on debian too.
However a recent change On Debian Testing (Release renamed from "testing" to "11", see lsb_release output) broke the devshell image build.

In this fix a replaced one of the env. variable (OS_PLATFORM) and introduces 2 more env. variable (OS_DISTRIBUTION && RELEASE_CODE_NAME).
With these change in builddeps can set precisely things. (For example: which release need to install debugging symbol)

OS_PLATFORM env. replaced by IMAGE_PLATFORM
OS_DISTRIBUTION env. store the docker system type
RELEASE_CODE_NAME env. store the the docker system code name

Related PR: https://github.com/syslog-ng/syslog-ng/pull/3712

Other possible solution is:
create a bash script which can set these env. variable (OS_DIS... RELEASE..) by name of the docker file. But I think is more "elegant" the current solution, because it didn't create a new file and only has a little code duplication (but I think if we have more than 20 docker file should consider to create a mapping script - which can set env. variable by image platform)


kira-syslogng:
Build SUCCESS

dbld/builddeps
```diff
@@ -91,22 +91,20 @@ function download_target() {
 }
 
 function filter_packages_by_platform {
-    FILENAME=$1
```
bazsi, line 94: It's common practice to name positional arguments via assignments so that we don't need to remember which argument is which. So I would leave FILENAME here.

dbld/builddeps
```diff
@@ -16,18 +16,18 @@ set -x
...
+    case "${OS_DISTRIBUTION}" in
+        centos)
             $YUM_INSTALL wget yum-utils
             ;;
-        debian-*|ubuntu-*)
+        debian|ubuntu)
             apt-get update
             $APT_INSTALL wget gnupg2
-            if [ "${OS_PLATFORM}" = "ubuntu-xenial" ]; then
+            if [ "${IMAGE_PLATFORM}" = "ubuntu-xenial" ]; then
```
bazsi, line 26: This should be dependent on $RELEASE_CODE_NAME instead, although we currently only have one xenial based image.

dbld/builddeps
```diff
@@ -154,22 +152,22 @@ function install_yum_packages {
...
-    case "${OS_PLATFORM}" in
+    case "${IMAGE_PLATFORM}" in
         fedora-*)
             python_executables="python"
             ;;
-        ubuntu-focal|tarball)
+        ubuntu-focal|tarball|devshell)
             python_executables="python3"
             ;;
-        devshell|kira)
```
bazsi, line 164: This drops python2 from de shell, which I would keep as long as we might have any sensible need for it.

This may be ok to do now, but why keep it in Kira then?

bazsi: Also old versions of debian use python 2 iirc.

dbld/images/debian-stretch.dockerfile
```diff
@@ -1,9 +1,11 @@
 FROM debian:9
 LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>, Balazs Scheidler <balazs.scheidler@oneidentity.com>"
+ENV OS_DISTRIBUTION=debian
+ENV RELEASE_CODE_NAME=9
```
gaborznagy, line 4: This will not work IMHO, code name shouldn't be the release number. Debian 9 code name is "stretch"

dbld/images/centos-7.dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM centos:7
 LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>, Balazs Scheidler <balazs.scheidler@oneidentity.com>"
+ENV OS_DISTRIBUTION=centos
+ENV RELEASE_CODE_NAME=7
```
gaborznagy, line 4: Seems nitpicking, but without context the name `RELEASE_CODE_NAME` doesn't really suggesting what should be the content.
And it's really tied to `OS_DISTRIBUTION` (e.g. it's release 7 of centos distro), so we should name it to emphasize this.
Also, we already have syslog-ng release related variables (see `RELEASE_DIR`), this would really mix things up.

What do you think about changing it to `OS_RELEASE_CODE_NAME` or `DISTRIBUTION_CODE_NAME` or OS_DISTRIBUTION_CODE_NAME`?

kira-syslogng:
Build FAILURE

dbld/builddeps
Barkodcz: The reason why we drops python2 from devshell is when I tried run dbld/rules script it throw error, because can't use python2 for install pip, (and also I discussed with the guys, and if I understand everything correctly, devshell shouldn't use the pip at all). 
The Kira need to be keep on python2, because some test function is using only python2

dbld/images/debian-stretch.dockerfile
Barkodcz: Thanks, I fixed

dbld/images/centos-7.dockerfile
Barkodcz: At the beginning of the development I thought how to name it, and the reason of why I named to `RELEASE_CODE_NAME` is I checked the official debian site and they are referring to release code.
But your suggestion is better :) so I changed to `OS_DISTRIBUTION_CODE_NAME`

dbld/builddeps
Barkodcz: ohh okey, I put it back the FILENAME

dbld/builddeps
Barkodcz: okey, I took your advice.

kira-syslogng:
Build SUCCESS

dbld/images/ubuntu-xenial.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:xenial
```
MrAnno, line 1: Why do we use codename tags in Docker?

dbld/builddeps
MrAnno: Can we clarify this, please?
Why can't you use python2 in devshell? Kira needs it anyway, so if you encountered a issue, it should be fixed in the Kira image too.

MrAnno:
I have only 2 questions, otherwise 👍🏻 .

dbld/images/kira.dockerfile
```diff
@@ -1,7 +1,10 @@
 FROM balabit/syslog-ng-ubuntu-focal
 LABEL maintainer="Andras Mitzki <andras.mitzki@balabit.com>, Laszlo Szemere <laszlo.szemere@balabit.com>, Balazs Scheidler <balazs.scheidler@oneidentity.com>"
 
-ENV OS_PLATFORM kira
+ARG IMAGE_PLATFORM
+ARG COMMIT
+ENV IMAGE_PLATFORM kira
```
alltilla, line 6: Why don't we use `${IMAGE_PLATFORM}` ? Here, and other places, too.
Or alternatively we don't need to use `ARG`s anywhere, and burn in the values to every dockerfile like we do it for kira.

dbld/images/ubuntu-xenial.dockerfile
MrAnno: Discussed IRL.

dbld/builddeps
Barkodcz: I re-tried, if devshell still not working with python2, but now it is working with python2. So devshell now using python2 also (as before)

kira-syslogng:
Build SUCCESS

dbld/builddeps
MrAnno: That'd good news. Thanks.

MrAnno (dismissed):
Please remove the test commit when you're ready.

dbld/images/kira.dockerfile
gaborznagy: I've checked and there is no reason to manually override `IMAGE_PLATFORM` in case of `kira` or `tarball` or `devshell`. We can use the environment variable like in other platforms:

```
dbld/images/ubuntu-bionic.dockerfile:ENV IMAGE_PLATFORM ${IMAGE_PLATFORM}
```

gaborznagy (commented):
Looks good to me.

When Attila's question is resolved, we can merge this.

alltilla (dismissed):

kira-syslogng:
Build SUCCESS

MrAnno (approved):
Thank you.

kira-syslogng:
Build SUCCESS

alltilla (approved):

dbld/images/kira.dockerfile
Barkodcz: We discussed IRL, with the same name can't be set (because of inheritance), but If ARG variable name is different of ENV variable can be set (current version)

PR #3717, opened 2021-07-04
scl: add discord() destination

MrAnno:
This PR adds a `discord()` SCL destination to syslog-ng.

Based on @czanik's [blog post](https://www.syslog-ng.com/community/b/blog/posts/first-steps-of-sending-alerts-to-discord-and-others-from-syslog-ng-http-and-apprise).

Example usage:

```
destination {
  discord(url("https://discord.com/api/webhooks/x/y"));
};
```

Options:

```
discord(
  url("https://discord.com/api/webhooks/x/y")
  avatar-url("https://example.domain/any_image.png")
  username("$HOST-bot") # Custom bot name, accepts macros
  tts(true) # Text-to-Speech message
  template("${MSG:-[empty message]}") # Message to send, can't be empty
);
```

Setup guide: https://support.discord.com/hc/en-us/articles/228383668-Intro-to-Webhooks

image

----

API docs:
https://discord.com/developers/docs/resources/webhook#execute-webhook

The default throttle value is set to 5:
https://discord.com/developers/docs/topics/rate-limits#global-rate-limit

We also need a limit for the maximum message length, it's currently 2000 characters.

----

This SCL uses Discord [webhooks](https://discord.com/developers/docs/resources/webhook), instead of the Discord "bot" API.

Webhooks are easier to set up, and they also support embedded hyperlinks in messages, and multiple "embed" objects within a single message:

> Webhooks are a low-effort way to post messages to channels in Discord.
> They do not require a bot user or authentication to use.

Webhook-based bots can not read messages, but we obviously don't need that.

If the regular "bot" API becomes a better choice in the future, we can extend this SCL and add a `token()` option in a backward-compatible manner, but this requires a "preprocessor if" statement in our config language, because the base URL will be different when using this API.
Another option is just to add a new SCL (`discord-bot(token() ...)`), when needed.

### DIFF
--- a/modules/basicfuncs/tests/test_basicfuncs.c
+++ b/modules/basicfuncs/tests/test_basicfuncs.c
@@ -202,6 +202,10 @@ Test(basicfuncs, test_str_funcs)
   assert_template_format("$(substr $HOST -1)", "p");
   assert_template_format("$(substr $HOST -2 1)", "r");
 
+  assert_template_format("$(substr 'ssstring-shorter-than-the-specified-length' 2 1400)",
+                         "string-shorter-than-the-specified-length");
+
+
   assert_template_format("$(strip ${APP.STRIP1})", "value");
   assert_template_format("$(strip ${APP.STRIP2})", "value");
   assert_template_format("$(strip ${APP.STRIP3})", "value");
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan (approved):

Barkodcz (approved):

PR #3726, opened 2021-07-07
diskq: fix invalid read after disk_buf_size

alltilla:
We do not check beforehand, whether we can fit a new message to the disk-buffer, we decide to wrap the `write_head` after we wrote over the `disk-buf-size` limit.

It is possible to be just before the `disk-buf-size` with the `write_head` and write a huge message, then wrap the `write_head` to the beginning. This will change `self->file_size` to be a large value.

Let's say we set `truncate-size-ratio(1)`, so we never truncate and never change the `file_size` from that large value.

Imagine, that we handled every message, then wrote a lot of messages to the disk-buffer again, so we are just before the `disk-buf-size` limit.
We write 2 small messages, one will end just after `disk-buf-size` but still before the huge `self->file_size`, the next one will start at the beginning, and end a bit after that.
This is because we decide to wrap the `write_head` based on the position related to disk-buf-size (see `_is_qdisk_overwritten()`).

When we are reading from the disk-buffer we currently decide to wrap the `read_head` based on whether it is after the `self->file_size`.
We read the small message that ends just after `disk-buf-size`, and we decide not to wrap, and try to read one more message there, which is the middle of an old message.

This PR matches the wrap logic between the write and read head, so they both measure their position relative to `disk-buf-size`.

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -238,7 +238,7 @@ _possible_size_reduction_reaches_truncate_threshold(QDisk *self, gint64 expected
 }
 
 static void
-_truncate_file(QDisk *self, gint64 expected_size)
+_maybe_truncate_file(QDisk *self, gint64 expected_size)
 {
   if (_ftruncate_would_reduce_file(self, expected_size) &&
       !_possible_size_reduction_reaches_truncate_threshold(self, expected_size) &&
@@ -293,7 +293,7 @@ _truncate_file_to_minimal(QDisk *self)
 {
   if (qdisk_is_file_empty(self))
     {
-      _truncate_file(self, QDISK_RESERVED_SPACE);
+      _maybe_truncate_file(self, QDISK_RESERVED_SPACE);
       return;
     }
 
@@ -301,7 +301,7 @@ _truncate_file_to_minimal(QDisk *self)
   if (file_end_offset <= QDISK_RESERVED_SPACE)
     return;
 
-  _truncate_file(self, file_end_offset);
+  _maybe_truncate_file(self, file_end_offset);
 }
 
 
@@ -389,7 +389,7 @@ qdisk_push_tail(QDisk *self, GString *record)
         {
           msg_debug("Unused area ahead of write_head, truncate queue file",
                     evt_tag_long("new size",  self->hdr->write_head));
-          _truncate_file(self, self->hdr->write_head);
+          _maybe_truncate_file(self, self->hdr->write_head);
         }
       else
         {
@@ -1114,7 +1114,7 @@ qdisk_reset_file_if_empty(QDisk *self)
   self->hdr->write_head = QDISK_RESERVED_SPACE;
   self->hdr->backlog_head = QDISK_RESERVED_SPACE;
 
-  _truncate_file(self, QDISK_RESERVED_SPACE);
+  _maybe_truncate_file(self, QDISK_RESERVED_SPACE);
 }
 
 DiskQueueOptions *
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -192,6 +192,14 @@ qdisk_is_space_avail(QDisk *self, gint at_least)
 {
   /* sizeof(guint32): record_length is a 4 bytes long value which is stored before each serialized LogMessage */
   gint64 msg_len = at_least + sizeof(guint32);
+  /* write follows read (e.g. we are appending to the file) OR
+   * there's enough space between write and read.
+   *
+   * If write follows read we need to check two things:
+   *   - either we are below the maximum limit (GINT64_FROM_BE(self->hdr->write_head) < self->options->disk_buf_size)
+   *   - or we can wrap around (GINT64_FROM_BE(self->hdr->read_head) != QDISK_RESERVED_SPACE)
+   * If neither of the above is true, the buffer is full.
+   */
   return (
            (_is_backlog_head_prevent_write_head(self)) &&
            (_is_write_head_less_than_max_size(self) || _is_able_to_reset_write_head_to_beginning_of_qdisk(self))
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

alltilla:
@kira-syslogng do stresstest

MrAnno:
Can you explain the exact difference between `useful_file_size` and `file_size` and when they have to be updated?
Maybe we could find a more "intention-revealing" variable name that way.

alltilla:
file_size is the physical disk used by the disk-queue file. useful_file_size is the physical disk usage of the mmapped region and the log messages which are waiting to be processed (read or acked). Sorry if I am cryptic.

kira-syslogng:

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -380,6 +380,83 @@ Test(diskq_truncate, test_diskq_truncate_size_ratio_1)
...
+  send_some_messages(q, some_messages_to_let_write_head_wrap);
+  log_queue_ack_backlog(q, some_messages_to_let_write_head_wrap);
+
+  // 7. feed 2 to wrap write_head and to add one message to the beginning
+  feed_some_messages(q, 1);
+  cr_assert(qdisk->hdr->write_head == QDISK_RESERVED_SPACE, "write_head should have wrapped");
+  feed_some_messages(q, 1);
+
+  // 8. process all messages
+  send_some_messages(q, just_under_max_size_message_number - some_messages_to_let_write_head_wrap + 2);
```
Barkodcz, line 447: [optional]

Instead of using number 2 (as a "magic number") creating a variable, that counting unsent && unacknowledged message.

MrAnno:
It seems that `useful_file_size` is not recovered after restarting syslog-ng, so the fix may be incomplete.

I think the best would be if we could fix this issue without adding more "state".
If we could calculate `useful_file_size` at restart, then it's actually an unnecessary field, otherwise, we should persist this field as well.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

MrAnno:
It turned out that the additional "state" is required, and it also has to be persisted (has to be added to the disk buffer header).

This is because the truncation logic was not only used to reduce file size, but for 2 special operations as well:

- to remove persisted memory-buffers (qout, overflow backlog) from the disk buffer after a restart (non-reliable qdisk)
- to remove an overflowing disk buffer section when we write past `disk-buf-size()` (we allow a single message at the end of the disk-buffer to exceed the disk-buf-size limit)

Since we need to add a new field to the header due to the new truncation logic, the diskq version needs to be bumped too.

alltilla:
@kira-syslogng do stresstest

alltilla:
@kira-syslogng do stresstest

kira-syslogng:
Build FAILURE

alltilla:
KIRA failed only because of an unrelated problem (multi-line-timeout).

kira-syslogng:

MrAnno:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/diskq/tests/test_diskq_truncate.c
alltilla: Thanks, fixed!

alltilla:
@kira-syslogng do stresstest

modules/diskq/qdisk.c
```diff
@@ -954,25 +981,16 @@ qdisk_start(QDisk *self, const gchar *filename, GQueue *qout, GQueue *qbacklog,
       if (self->hdr->version == 0)
         _update_header_with_default_values(self);
 
-      if ((self->hdr->big_endian && G_BYTE_ORDER == G_LITTLE_ENDIAN) ||
-          (!self->hdr->big_endian && G_BYTE_ORDER == G_BIG_ENDIAN))
+      if (!_is_header_version_current(self))
```
gaborznagy, line 1005: minor: this patch belongs to the next commit

kira-syslogng:

kira-syslogng:
Build SUCCESS

gaborznagy (commented):
I've just glimpsed into the PR to see what's changing.

alltilla:
@kira-syslogng do stresstest

kira-syslogng:
Build SUCCESS

kira-syslogng:

MrAnno (approved):

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -380,6 +388,134 @@ Test(diskq_truncate, test_diskq_truncate_size_ratio_1)
...
+  send_some_messages(q, some_messages_to_let_write_head_wrap);
+  log_queue_ack_backlog(q, some_messages_to_let_write_head_wrap);
+  unprocessed_messages_in_buffer -= some_messages_to_let_write_head_wrap;
+
+  // 3. feed additional messages, so write_head wraps
+  feed_some_messages(q, 300);
+  unprocessed_messages_in_buffer += 300;
+  cr_assert(qdisk_get_writer_head(qdisk) < qdisk_get_reader_head(qdisk), "write_head should have wrapped");
+
+  // Restart
```
szemere, line 504: I think it is a really good "side effect" of the new truncation feature that we started to test on the number of messages instead of pure file size. (They were technically equal before, but it was semantically incorrect.)

However the current implementation (and I think a unit test can rely on that detail) depends on the fact that we truncate away the non reliable queue's (qout, qbacklog, qoverflow), even if the truncation is completely disabled.

As a result: my proposal is that we should check on the actual file size after load, prior starting to send out any messages.

MrAnno: IMHO the current implementation that truncates memory-buffers away should be fixed eventually.
dqtool open the disk queue in read-only mode, so there were bugs there too.

szemere (approved):
Thank You!

PR #3729, opened 2021-07-12
diskq: fix invalid read after disk_buf_size

alltilla:
We do not check beforehand, whether we can fit a new message to the disk-buffer, we decide to wrap the `write_head` after we wrote over the `disk-buf-size` limit.

It is possible to be just before the `disk-buf-size` with the `write_head` and write a huge message, then wrap the `write_head` to the beginning. This will change `self->file_size` to be a large value.

Let's say we set `truncate-size-ratio(1)`, so we never truncate and never change the `file_size` from that large value.

Imagine, that we handled every message, then wrote a lot of messages to the disk-buffer again, so we are just before the `disk-buf-size` limit.
We write 2 small messages, one will end just after `disk-buf-size` but still before the huge `self->file_size`, the next one will start at the beginning, and end a bit after that.
This is because we decide to wrap the `write_head` based on the position related to disk-buf-size (see `_is_qdisk_overwritten()`).

When we are reading from the disk-buffer we currently decide to wrap the `read_head` based on whether it is after the `self->file_size`.
We read the small message that ends just after `disk-buf-size`, and we decide not to wrap, and try to read one more message there, which is the middle of an old message.

This commit matches the wrap logic between the write and read head, so they both measure their position relative to `disk-buf-size`.

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -67,6 +67,8 @@ typedef union _QDiskFileHeader
     QDiskQueuePosition qoverflow_pos;
     gint64 backlog_head;
     gint64 backlog_len;
+
+    gboolean use_old_read_head_wrap_condition;
   };
   gchar _pad2[QDISK_RESERVED_SPACE];
 } QDiskFileHeader;
@@ -78,7 +80,6 @@ struct _QDisk
   gint fd;
   gint64 file_size;
   QDiskFileHeader *hdr;
-  gboolean use_old_read_head_wrap_condition;
   DiskQueueOptions *options;
 };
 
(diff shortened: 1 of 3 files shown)


alltilla:
@kira-syslogng do stresstest

kira-syslogng:
Build FAILURE

kira-syslogng:

kira-syslogng:
Build FAILURE

alltilla:
@kira-syslogng do stresstest

kira-syslogng:
Build SUCCESS

kira-syslogng:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

alltilla:
Closing in favor of #3726.

PR #3732, opened 2021-07-13
Add workers() option to the Redis destination

OverOrion:
This PR adds `workers()` support for the `redis()` destination driver.

This increases the throughput of the Redis destination driver.

PR #3741, opened 2021-07-23
log-threaded-dest: add grammar rules to opt-in for workers() and batch-...() options

alltilla:
Not every `LogThreadedDest` implementation utilizes `batch-lines()` and `batch-timeout()`.
Similarly not every `LogThreadedDest` implementation utilizes `workers()`.

The batch-related options had the problem, where any descendant driver inherited them, even if they did not use them.
The `workers()` option had the problem, where every descendant driver needed to re-implement it in order to use it.

This PR fixes both.

---

**Note for the reviewers:**

This PR is not a simple refactor, I needed to check which threaded drivers use the batch-related options.

When reviewing, please make sure to check that I have found every driver which uses them, and that I have not put it to drivers, which do not use them.

Thanks!

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -1194,14 +1194,17 @@ dest_driver_option
         | driver_option
         ;
 
+threaded_dest_driver_batch_option
+        : KW_BATCH_LINES '(' nonnegative_integer ')' { log_threaded_dest_driver_set_batch_lines(last_driver, $3); }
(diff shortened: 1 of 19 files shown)


alltilla:
Conflicts with #3732. We need to rebase the PR which will be merged later.

kira-syslogng:
Build SUCCESS

Kokan (approved):

Barkodcz (approved):
Thank you!

PR #3746, opened 2021-08-02
logqueue-disk: serialize outside of the lock in the source threads

alltilla:
It is common to have multiple source threads.
When we use disk-buffer, the message's serialization and writing it to the disk happen in the source thread.

Currently, the serialization happens while the source thread is holding the `LogQueue`'s lock. Also it is inbetween two accesses of the `QDisk`.
If we just simply move the serialization outside the lock, we are not guaranteed to have the same `QDisk` state, as we were having beforehand, so it is better to move the whole serialization between `QDisk` is ever accessed and the lock is obtained.

This can yield significant performance improvement, because while the source threads are doing their slow serialization, the destination thread can process messages from the `QDisk`. In my local measurements, this change introduced a 40-50% performance gain compared to syslog-ng-3.32.1.

No news file needed, IMO.

### DIFF
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -239,10 +239,13 @@ exit:
 }
 
 static void
-_rewind_backlog (LogQueueDisk *s, guint rewind_count)
+_rewind_backlog(LogQueue *s, guint rewind_count)
 {
   guint i;
-  LogQueueDiskNonReliable *self = (LogQueueDiskNonReliable *) s;
+  LogQueueDiskNonReliable *self = (LogQueueDiskNonReliable *)s;
+
+  g_static_mutex_lock(&s->lock);
+
   rewind_count = MIN(rewind_count, _get_message_number_in_queue(self->qbacklog));
 
   for (i = 0; i < rewind_count; i++)
@@ -253,9 +256,17 @@ _rewind_backlog (LogQueueDisk *s, guint rewind_count)
       g_queue_push_head (self->qout, ptr_opt);
       g_queue_push_head (self->qout, ptr_msg);
 
-      log_queue_queued_messages_inc(&self->super.super);
-      log_queue_memory_usage_add(&self->super.super, log_msg_get_size((LogMessage *)ptr_msg));
+      log_queue_queued_messages_inc(s);
+      log_queue_memory_usage_add(s, log_msg_get_size((LogMessage *)ptr_msg));
     }
+
+  g_static_mutex_unlock(&s->lock);
+}
+
+static void
+_rewind_backlog_all(LogQueue *s)
+{
+  _rewind_backlog(s, -1);
 }
 
 static LogMessage *
@@ -422,7 +433,8 @@ _set_virtual_functions (LogQueueDisk *self)
 {
   self->super.get_length = _get_length;
   self->super.ack_backlog = _ack_backlog;
-  self->rewind_backlog = _rewind_backlog;
+  self->super.rewind_backlog = _rewind_backlog;
+  self->super.rewind_backlog_all = _rewind_backlog_all;
   self->pop_head = _pop_head;
   self->super.push_head = _push_head;
   self->push_tail = _push_tail;
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -333,10 +333,25 @@ _push_head(LogQueue *s, LogMessage *msg, const LogPathOptions *path_options)
   g_static_mutex_unlock(&s->lock);
 }
 
-static gboolean
-_push_tail (LogQueueDisk *s, LogMessage *msg, LogPathOptions *local_options, const LogPathOptions *path_options)
+static void
+_drop_msg(LogQueueDiskNonReliable *self, LogMessage *msg, const LogPathOptions *path_options)
 {
-  LogQueueDiskNonReliable *self = (LogQueueDiskNonReliable *) s;
+  stats_counter_inc(self->super.super.dropped_messages);
+
+  if (path_options->flow_control_requested)
+    log_msg_ack(msg, path_options, AT_SUSPENDED);
+  else
+    log_msg_drop(msg, path_options, AT_PROCESSED);
+}
+
+static void
+_push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *path_options)
+{
+  LogQueueDiskNonReliable *self = (LogQueueDiskNonReliable *)s;
+
+  g_static_mutex_lock(&s->lock);
+
+  LogPathOptions local_options = *path_options;
 
   if (HAS_SPACE_IN_QUEUE(self->qout) && qdisk_get_length (self->super.qdisk) == 0)
     {
@@ -347,33 +362,43 @@ _push_tail (LogQueueDisk *s, LogMessage *msg, LogPathOptions *local_options, con
       g_queue_push_tail (self->qout, LOG_PATH_OPTIONS_FOR_BACKLOG);
       log_msg_ref (msg);
 
-      log_queue_memory_usage_add(&self->super.super, log_msg_get_size(msg));
+      log_queue_memory_usage_add(s, log_msg_get_size(msg));
     }
   else
     {
-      if (self->qoverflow->length != 0 || !log_queue_disk_write_message(s, msg))
+      if (self->qoverflow->length != 0 || !log_queue_disk_write_message(&self->super, msg))
         {
           if (HAS_SPACE_IN_QUEUE(self->qoverflow))
             {
               g_queue_push_tail (self->qoverflow, msg);
               g_queue_push_tail (self->qoverflow, LOG_PATH_OPTIONS_TO_POINTER (path_options));
               log_msg_ref (msg);
-              local_options->ack_needed = FALSE;
-              log_queue_memory_usage_add(&self->super.super, log_msg_get_size(msg));
+              local_options.ack_needed = FALSE;
+              log_queue_memory_usage_add(s, log_msg_get_size(msg));
             }
           else
             {
               msg_debug ("Destination queue full, dropping message",
                          evt_tag_str  ("filename", qdisk_get_filename (self->super.qdisk)),
-                         evt_tag_long ("queue_len", _get_length(&s->super)),
+                         evt_tag_long ("queue_len", log_queue_get_length(s)),
                          evt_tag_int  ("mem_buf_length", self->qoverflow_size),
                          evt_tag_long ("disk_buf_size", qdisk_get_maximum_size (self->super.qdisk)),
-                         evt_tag_str  ("persist_name", self->super.super.persist_name));
-              return FALSE;
+                         evt_tag_str  ("persist_name", s->persist_name));
+              _drop_msg(self, msg, path_options);
+              g_static_mutex_unlock(&s->lock);
+              return;
             }
         }
     }
-  return TRUE;
+
+  log_queue_push_notify(s);
+  log_queue_queued_messages_inc(s);
+  log_msg_ack(msg, &local_options, AT_PROCESSED);
+  log_msg_unref(msg);
+
+  g_static_mutex_unlock(&s->lock);
+
+  return;
 }
 
 static void
@@ -446,7 +471,7 @@ _set_virtual_functions (LogQueueDisk *self)
   self->super.rewind_backlog_all = _rewind_backlog_all;
   self->super.pop_head = _pop_head;
   self->super.push_head = _push_head;
-  self->push_tail = _push_tail;
+  self->super.push_tail = _push_tail;
   self->start = _start;
   self->super.free_fn = _free;
   self->load_queue = _load_queue;
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -25,6 +25,7 @@
 #include "logpipe.h"
 #include "messages.h"
 #include "syslog-ng.h"
+#include "scratch-buffers.h"
 
 #define ITEM_NUMBER_PER_MESSAGE 2
 
@@ -146,6 +147,30 @@ _has_movable_message(LogQueueDiskNonReliable *self)
              || qdisk_is_space_avail (self->super.qdisk, 4096));
 }
 
+static gboolean
+_write_message_to_disk(LogQueueDisk *self, LogMessage *msg)
+{
+  gboolean consumed = FALSE;
+
+  if (qdisk_started(self->qdisk) && qdisk_is_space_avail(self->qdisk, 64))
+    {
+      ScratchBuffersMarker marker;
+      GString *write_serialized = scratch_buffers_alloc_and_mark(&marker);
+
+      if (!qdisk_serialize_msg(self->qdisk, msg, write_serialized))
+        {
+          scratch_buffers_reclaim_marked(marker);
+          return FALSE;
+        }
+
+      consumed = qdisk_push_tail(self->qdisk, write_serialized);
+
+      scratch_buffers_reclaim_marked(marker);
+    }
+
+  return consumed;
+}
+
 static void
 _move_messages_from_overflow(LogQueueDiskNonReliable *self)
 {
@@ -166,7 +191,7 @@ _move_messages_from_overflow(LogQueueDiskNonReliable *self)
         }
       else
         {
-          if (log_queue_disk_write_message(&self->super, msg))
+          if (_write_message_to_disk(&self->super, msg))
             {
               log_queue_memory_usage_sub(&self->super.super, log_msg_get_size(msg));
             }
@@ -366,7 +391,7 @@ _push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *path_options)
     }
   else
     {
-      if (self->qoverflow->length != 0 || !log_queue_disk_write_message(&self->super, msg))
+      if (self->qoverflow->length != 0 || !_write_message_to_disk(&self->super, msg))
         {
           if (HAS_SPACE_IN_QUEUE(self->qoverflow))
             {
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -270,20 +270,22 @@ _rewind_backlog_all(LogQueue *s)
 }
 
 static LogMessage *
-_pop_head (LogQueueDisk *s, LogPathOptions *path_options)
+_pop_head(LogQueue *s, LogPathOptions *path_options)
 {
-  LogQueueDiskNonReliable *self = (LogQueueDiskNonReliable *) s;
+  LogQueueDiskNonReliable *self = (LogQueueDiskNonReliable *)s;
   LogMessage *msg = NULL;
+
+  g_static_mutex_lock(&s->lock);
 
   if (self->qout->length > 0)
     {
       msg = g_queue_pop_head (self->qout);
       POINTER_TO_LOG_PATH_OPTIONS (g_queue_pop_head (self->qout), path_options);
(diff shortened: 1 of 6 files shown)


MrAnno:
> This can yield significant performance improvement

Could you share the measurement, please? As far as I remember, it was a 60% improvement in general when the reliable disk-buffer was used.

The non-reliable disk-buffer has some qout/qoverflow move logic that slows us down, we're investigating that one separately.

> No news file needed, IMO.

I think we could add a NEWS entry about the overall performance improvement in the last pull request.

----

I'm sharing two flame graphs, which shows the ratios between the actual disk writes and the serialization/deserialization logic:

image
image

kira-syslogng:
Build SUCCESS

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -162,7 +163,19 @@ _move_messages_from_overflow(LogQueueDiskNonReliable *self)
         }
       else
         {
-          if (log_queue_disk_write_message(&self->super, msg))
+          ScratchBuffersMarker marker;
+          GString *serialized = scratch_buffers_alloc_and_mark(&marker);
```
Barkodcz, line 167: I think it would better, if you "allocate" the GString outside of the while loop and here just truncate the GString.
If can't put from overflow into qout, but can put into disk, then it needs "allocate" a GString.

alltilla:
> > This can yield significant performance improvement
> 
> Could you share the measurement, please? As far as I remember, it was a 60% improvement in general when the reliable disk-buffer was used.
> 
> The non-reliable disk-buffer has some qout/qoverflow move logic that slows us down, we're investigating that one separately.
> 
> > No news file needed, IMO.
> 
> I think we could add a NEWS entry about the overall performance improvement in the last pull request.
> 
> I'm sharing two flame graphs, which show the ratios between the actual disk writes and the serialization/deserialization logic:
> 
> image
> image

Sure, let me measure everything again tomorrow. Thanks!

modules/diskq/qdisk.c
```diff
@@ -569,6 +555,73 @@ qdisk_remove_head(QDisk *self)
...
+{
+  gboolean result = TRUE;
+  SerializeArchive *sa = serialize_string_archive_new(serialized);
+  *msg = log_msg_new_empty();
+
+  if (!log_msg_deserialize(*msg, sa))
+    {
+      msg_error("Can't read correct message from disk-queue file",
+                evt_tag_str("filename", qdisk_get_filename(self)),
+                evt_tag_long("read_position", qdisk_get_reader_head(self)));
```
alltilla, line 615: read_position might not be the best to be used here.

modules/diskq/logqueue-disk-non-reliable.c
alltilla: Thanks, fixed!

modules/diskq/qdisk.c
alltilla: Removed.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -569,6 +569,23 @@ qdisk_remove_head(QDisk *self)
...
 
+gboolean
+qdisk_serialize_msg(QDisk *self, LogMessage *msg, GString *serialized)
+{
+  gboolean result = TRUE;
+  SerializeArchive *sa = serialize_string_archive_new(serialized);
+
+  if (!log_msg_serialize(msg, sa, self->options->compaction ? LMSF_COMPACTION : 0))
+    {
+      g_string_truncate(serialized, 0);
```
Kokan: [optional] truncating the serialized method not really needed, by default I would expect that if a function returns with an error, it leaves the `serialized` in an unspecified (but valid) state[1]. So I need to for example free it, but cannot assume much about the content.
As an alternative to the unspecified, that it does not change the `serialized` string in case of failure. This is good, when you want to process multiple messages and batch into one string.

That's the reasoning why I would not do truncate here.

[1] One could say that always truncating is one of the unspecified and valid state, which is true. But not having code to reach that always nice.

modules/diskq/qdisk.c
```diff
@@ -342,31 +342,17 @@ qdisk_push_tail(QDisk *self, GString *record)
       self->hdr->write_head = QDISK_RESERVED_SPACE;
     }
 
-  if (!qdisk_is_space_avail(self, record->len))
+  if (!qdisk_is_space_avail(self, record->len - sizeof(guint32)))
```
Kokan, line 345: The responsibility of adding the length of the string was removed from `qdisk_push_tail`, so it should be removed from `qdisk_is_space_avail` as well, I mean it is hard to argue why I have to subtract that amount, and for sure remembering it is unlikely.

modules/diskq/qdisk.c
```diff
@@ -569,18 +555,47 @@ qdisk_remove_head(QDisk *self)
...
   if (!log_msg_serialize(msg, sa, self->options->compaction ? LMSF_COMPACTION : 0))
     {
-      g_string_truncate(serialized, 0);
       result = FALSE;
+      goto exit;
     }
 
+  if (!_overwrite_with_real_record_length(serialized))
+    {
+      msg_error("Error serializing empty message for the disk-queue file");
```
Kokan, line 590: Please provide identifiers for investigating later on (like diskq filename).

modules/diskq/qdisk.c
Kokan: What is the issue using `read_position` here ?

alltilla: It is not guaranteed, that the `deserialize` function will be called on a message, that was read from the `read_head`.

Kokan: what if you move the error message from here, where you can be sure about that ? as it seems like useful information to have, and error reporting separation from logic itself not so bad idea otherwise as well.

alltilla: I added another error message to the caller site. I think having 2 errors is not a big problem, this problem should rarely occur.

modules/diskq/logqueue-disk.c
```diff
@@ -86,20 +84,23 @@ _push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *path_options)
       goto exit;
     }
 
+  g_static_mutex_lock(&self->super.lock);
```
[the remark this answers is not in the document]
alltilla: Good insight, I moved the 3 mentioned function calls outside the lock.

modules/diskq/qdisk.c
alltilla: I removed the truncate from here, and removed the `log_msg_unref()` from the deserialization, too.

modules/diskq/qdisk.c
alltilla: Good point, I made your suggested adjustment.

modules/diskq/qdisk.c
alltilla: I added the filename to the error message, and also modified the code, to report errors from other failure points of this function.

kira-syslogng:
Build SUCCESS

alltilla:
Added the hint for needing serialized message.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:
Build SUCCESS

kira-syslogng:

modules/diskq/logqueue-disk.h
```diff
@@ -59,6 +59,6 @@ void log_queue_disk_restart_corrupted(LogQueueDisk *self);
 
 
 LogMessage *log_queue_disk_read_message(LogQueueDisk *self, LogPathOptions *path_options);
-gboolean log_queue_disk_write_message(LogQueueDisk *self, LogMessage *msg);
+gboolean log_queue_disk_write_message(LogQueueDisk *self, GString *serialized);
```
MrAnno, line 63: The symmetry is broken here with this change. The name of the function might have to be adjusted as well.

modules/diskq/logqueue-disk.h
```diff
@@ -36,8 +36,8 @@ struct _LogQueueDisk
   LogQueue super;
   QDisk *qdisk;         /* disk based queue */
   gint64 (*get_length)(LogQueueDisk *s);
-  gboolean (*push_tail)(LogQueueDisk *s, LogMessage *msg, LogPathOptions *local_options,
-                        const LogPathOptions *path_options);
+  gboolean (*push_tail)(LogQueueDisk *s, LogMessage *msg, GString *serialized,
+                        LogPathOptions *local_options, const LogPathOptions *path_options);
```
MrAnno, line 40: This looks bad.

modules/diskq/logqueue-disk.c
```diff
@@ -293,7 +293,7 @@ log_queue_disk_read_message(LogQueueDisk *self, LogPathOptions *path_options)
 gboolean
 log_queue_disk_write_message(LogQueueDisk *self, GString *serialized)
 {
-  if (qdisk_started(self->qdisk) && qdisk_is_space_avail(self->qdisk, 64))
+  if (qdisk_started(self->qdisk) && qdisk_is_space_avail(self->qdisk, serialized->len))
```
MrAnno, line 315: That looks much better :)

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -346,6 +382,20 @@ _push_tail (LogQueueDisk *s, LogMessage *msg, LogPathOptions *local_options, con
   return TRUE;
 }
 
+static gboolean
+_hint_need_serialized(LogQueueDisk *s)
```
MrAnno, line 386: The hint looks okay in my opinion, it's not intrusive and well separated.

MrAnno:
All of my negative review notes are based on the ways of how the serialization logic was moved out from the lock.

The previous interface looks like this:
image

I think it's really not okay to change the signature of `push_tail()` in `LogQueueDisk`, because the existence of those function pointers in the struct is just some inheritance trickery that makes it possible to lock the whole data structure in the top-level abstract class easily. 
In reality, those function pointers should have the exact same signature as `LogQueue` has.

Why don't we move the locks where they actually belong?
I think we could remove lock/unlock calls from `LogQueueDisk`, move them to where they are actually needed/used (`LogQueueDiskReliable` and `LogQueueDiskNonReliable`).

We may also get rid of the 2 protected methods of `LogQueueDisk`: `write_message()` and `read_message()`.

This way, message serialization can be done a separate unlocked step in the non-reliable and reliable disk buffer, and you could also implement the non-reliable serialization hint without polluting `LogQueueDisk` with something that only makes sense in the non-reliable disk-buffer implementation.

alltilla:
@MrAnno Thanks, I implemented your suggestion. The old implementation is available at: https://github.com/alltilla/syslog-ng/commits/diskq-perf-serialize-outside-lock-orig if for some reason it will be needed.

kira-syslogng:
Build SUCCESS

alltilla:
@kira-syslogng do stresstest

kira-syslogng:

modules/diskq/qdisk.h
```diff
@@ -82,4 +82,6 @@ const gchar *qdisk_get_filename(QDisk *self);
 
 guint64 qdisk_skip_record(QDisk *self, guint64 position);
 
+gboolean qdisk_serialize_msg(QDisk *self, LogMessage *msg, GString *serialized);
```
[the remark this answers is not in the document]
alltilla: I agree, but I would like to fix it in a separate PR.

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -347,33 +362,43 @@ _push_tail (LogQueueDisk *s, LogMessage *msg, LogPathOptions *local_options, con
...
               msg_debug ("Destination queue full, dropping message",
                          evt_tag_str  ("filename", qdisk_get_filename (self->super.qdisk)),
-                         evt_tag_long ("queue_len", _get_length(&s->super)),
+                         evt_tag_long ("queue_len", _get_length(s)),
                          evt_tag_int  ("mem_buf_length", self->qoverflow_size),
                          evt_tag_long ("disk_buf_size", qdisk_get_maximum_size (self->super.qdisk)),
-                         evt_tag_str  ("persist_name", self->super.super.persist_name));
-              return FALSE;
+                         evt_tag_str  ("persist_name", s->persist_name));
+              _drop_msg(self, msg, path_options);
```
MrAnno, line 457: I'm really happy that what we say in the debug message is actually happening in the next line. :)

modules/diskq/logqueue-disk-reliable.c
```diff
@@ -296,11 +312,16 @@ _push_tail(LogQueueDisk *s, LogMessage *msg, LogPathOptions *local_options, cons
...
-      local_options->ack_needed = FALSE;
+      log_queue_memory_usage_add(s, log_msg_get_size(msg));
+      local_options.ack_needed = FALSE;
     }
 
-  return TRUE;
+  log_queue_push_notify(s);
+  log_queue_queued_messages_inc(s);
+  log_msg_ack(msg, &local_options, AT_PROCESSED);
+  log_msg_unref(msg);
```
MrAnno, line 338: Just a note: Now that this has been moved here, it can be seen that `log_msg_ack()` should not always be called, that's why we have the `ack_needed = FALSE;` trickery a few lines above.

modules/diskq/logqueue-disk.c
```diff
@@ -210,10 +210,10 @@ log_queue_disk_restart_corrupted(LogQueueDisk *self)
...
 void
-log_queue_disk_init_instance(LogQueueDisk *self, const gchar *persist_name)
+log_queue_disk_init_instance(LogQueueDisk *self, DiskQueueOptions *options, const gchar *persist_name)
 {
   log_queue_init_instance(&self->super, persist_name);
-  self->qdisk = qdisk_new();
-
   self->super.type = log_queue_disk_type;
+
+  self->qdisk = qdisk_new(options, "SLQF");
```
MrAnno, line 192: This won't work, `SLRQ` vs `SLQF` was not a typo.

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -152,7 +152,7 @@ _write_message_to_disk(LogQueueDisk *self, LogMessage *msg)
 {
   gboolean consumed = FALSE;
 
-  if (qdisk_started(self->qdisk) && qdisk_is_space_avail(self->qdisk, 64))
```
MrAnno: Question:
This with the hard-coded value `64` can be considered as a hint, so we can skip message serialization and its cost when the disk-buffer has no "minimum" space at all.

Can't we add this to your new hint function?

MrAnno (commented):
I think the refactor was a really good direction. Now that locks are closer to the actual implementation, we already see additional places where those locks are not necessary.

A few lines of code duplication was added as a result of the refactor/inheritance fix, but further refactor rounds will be much easier upon this work than on the original code IMHO.

I think only one of my notes should be fixed in this PR (the accidental change of the disk-buffer file's magic numbers).

modules/diskq/logqueue-disk.c
```diff
@@ -271,5 +266,4 @@ log_queue_disk_init_instance(LogQueueDisk *self, const gchar *persist_name)
   self->super.type = log_queue_disk_type;
   self->super.push_tail = _push_tail;
   self->super.pop_head = _pop_head;
-  self->super.free_fn = _free;
```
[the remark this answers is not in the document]
MrAnno: Hm, it actually does not matter, `LogQueueDisk` have become a real abstract class with your refactor.

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -275,38 +315,113 @@ _pop_head (LogQueueDisk *s, LogPathOptions *path_options)
...
+  stats_counter_inc(self->super.super.dropped_messages);
+
+  if (path_options->flow_control_requested)
+    log_msg_ack(msg, path_options, AT_SUSPENDED);
+  else
+    log_msg_drop(msg, path_options, AT_PROCESSED);
+}
+
+static inline gboolean
+_need_serialized(LogQueueDiskNonReliable *self)
```
MrAnno, line 366:
```suggestion
/* _is_msg_serialization_needed_hint() must be called without holding the queue's lock.
 * This can only be used _as a hint_ for performance considerations, because as soon as the lock
 * is released, there will be no guarantee that the result of this function remain correct. */
static inline gboolean
_is_msg_serialization_needed_hint(LogQueueDiskNonReliable *self)
```

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -275,38 +315,113 @@ _pop_head (LogQueueDisk *s, LogPathOptions *path_options)
...
+
+static void
+_push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *path_options)
+{
+  LogQueueDiskNonReliable *self = (LogQueueDiskNonReliable *)s;
+
+  ScratchBuffersMarker marker;
+  GString *serialized_msg = NULL;
+
+  if (_need_serialized(self))
```
MrAnno, line 408:
```suggestion
  if (_is_msg_serialization_needed_hint(self))
```

modules/diskq/qdisk.c
```diff
@@ -569,12 +555,66 @@ qdisk_remove_head(QDisk *self)
...
+                evt_tag_str("error", error),
+                evt_tag_str("filename", qdisk_get_filename(self)));
+    }
+
+  serialize_archive_free(sa);
+  return error == NULL;
+}
+
+gboolean
+qdisk_deserialize_msg(QDisk *self, GString *serialized, LogMessage **msg)
```
Kokan, line 608: I would expect this function to unref the msg in case of an error.

Kokan (commented):
Does this version has the same performance figures as the previous proposed change ?
Have you measured that only the refactor part had any performance difference ?

modules/diskq/logqueue-disk-non-reliable.c
alltilla: Thanks, added!

modules/diskq/logqueue-disk-non-reliable.c
alltilla: Thanks, added!

modules/diskq/logqueue-disk-non-reliable.c
alltilla: Thanks, added!

modules/diskq/qdisk.c
alltilla: Thanks, fixed!

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -335,7 +339,7 @@ _push_tail (LogQueueDisk *s, LogMessage *msg, LogPathOptions *local_options, con
             {
               msg_debug ("Destination queue full, dropping message",
                          evt_tag_str  ("filename", qdisk_get_filename (self->super.qdisk)),
-                         evt_tag_long ("queue_len", _get_length(s)),
+                         evt_tag_long ("queue_len", _get_length(&s->super)),
```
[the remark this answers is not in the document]
alltilla: Thanks, fixed!

modules/diskq/logqueue-disk.c
alltilla: Good catch, thanks! Fixed!

alltilla:
@Kokan

> Does this version has the same performance figures as the previous proposed change ?

Yes.

> Have you measured that only the refactor part had any performance difference ?

No, I can do so, if you would like me to.

kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

MrAnno (approved):
Thank you

Kokan:
> No, I can do so, if you would like me to.

Not needed, I was just curios if we have the data.

Kokan (approved):

PR #3747, opened 2021-08-02
MQTT destination improvements

MrAnno:
This PR adds the following capabilities to the `mqtt()` destination:
- TLS and WebSocket Secure support (`ssl://`, `wss://`)
   ```
    mqtt(
      address("ssl://localhost:8883")
      topic("syslog/$HOST")
      fallback-topic("syslog/fallback")

      tls(
        ca-file("/path/to/ca.crt")
        key-file("/path/to/client.key")
        cert-file("/path/to/client.crt")
        peer-verify(yes)
        ssl-version("tlsv1_2")
        cipher-suite("ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305")
      )
    );
   ```
- Authentication with username and password
   ```
   mqtt(
     address("tcp://localhost:1883")
     topic("syslog/messages")
     username("user")
     password("passwd")
   );
   ```
- HTTP/HTTPS proxy support for WebSocket
   ```
   mqtt(
     address("wss://localhost:8883")
     topic("syslog/messages")
     http-proxy("https://localhost:4445")
     tls(...)
   );
   ```

The PR also improves the error reporting and debug logging of the MQTT module.

WARNING: The `ssl-version()` TLS option will not work if libpaho-mqtt is compiled with OpenSSL >= 1.1.
This is a bug in the library, I'm going to report this upstream.

@czanik


kira-syslogng:
Build SUCCESS

modules/mqtt/destination/mqtt-worker.c
```diff
@@ -191,9 +191,11 @@ _connect(LogThreadedDestWorker *s)
   gint rc;
 
   MQTTClient_connectOptions conn_opts = MQTTClient_connectOptions_initializer;
-
   conn_opts.keepAliveInterval = owner->keepalive;
   conn_opts.cleansession = FALSE;
+  conn_opts.username = owner->username;
+  conn_opts.password = owner->password;
```
alltilla, line 224: Is it okay to use these unconditionally?
For example can it cause a problem if only password is given?

If yes, then it would be better to notify the user about their incomplete config at syslog-ng init time, rather than at connect time.

modules/mqtt/destination/mqtt-destination.c
```diff
@@ -120,6 +124,87 @@ void mqtt_dd_set_password(LogDriver *d, const gchar *password)
...
+
+  self->peer_verify = verify;
+}
+
+void
+mqtt_dd_use_system_cert_store(LogDriver *d, gboolean use_system_cert_store)
+{
+  MQTTDestinationDriver *self = (MQTTDestinationDriver *) d;
+
+  /* TODO: auto_detect_ca_file() from the HTTP module */
```
alltilla, line 204: Are you planning to implement this in the current PR?

modules/mqtt/destination/mqtt-worker.c
```diff
@@ -202,6 +220,9 @@ _connect(LogThreadedDestWorker *s)
       conn_opts.httpsProxy = owner->http_proxy;
     }
 
+  MQTTClient_SSLOptions ssl_opts = _create_ssl_options(self);
+  conn_opts.ssl = &ssl_opts;
```
alltilla, line 233: Can this conflict with username/password? Also the same question is relevant here about setting these unconditionally / pre-checking the necessary values init time.

modules/mqtt/destination/mqtt-worker.c
```diff
@@ -182,6 +183,13 @@ _insert(LogThreadedDestWorker *s, LogMessage *msg)
   */
 }
 
+static gint
+_log_ssl_errors(const gchar *str, gsize len, gpointer u)
+{
+  msg_error("MQTT TLS error", evt_tag_printf("line", "%.*s", (gint) len, str));
```
alltilla, line 189: Can we use `MQTTDestinationDriver *self` here to give more info about the affected driver?

modules/mqtt/destination/mqtt-worker.c
MrAnno: NULL/0-values are considered as unset fields in this struct.
This is documented (it is not just an implementation detail):

https://www.eclipse.org/paho/files/mqttdoc/MQTTClient/html/struct_m_q_t_t_client__connect_options.html

modules/mqtt/destination/mqtt-destination.c
MrAnno: The current "implementation" actually works well with syslog-ng OSE, so I would do this later.

modules/mqtt/destination/mqtt-worker.c
MrAnno: User/pass authentication works well with TLS. Actually, it is the preferred way, otherwise, the password would be sent as plaintext.

(Using mutual authentication in TLS makes this redundant, of course.)

modules/mqtt/destination/mqtt-worker.c
MrAnno: Good idea, thanks.

kira-syslogng:
Build SUCCESS

configure.ac
```diff
@@ -1317,15 +1317,15 @@ if test "x$enable_mqtt" = "xyes" || test "x$enable_mqtt" = "xauto"; then
 		CPPFLAGS_SAVE="$CPPFLAGS"
 		LDFLAGS_SAVE="$LDFLAGS"
 		CPPFLAGS="$CPPFLAGS -I$with_libpaho_mqtt/include"
-		LDFLAGS="$LDFLAGS -L$with_libpaho_mqtt -lpaho-mqtt3c"
+		LDFLAGS="$LDFLAGS -L$with_libpaho_mqtt -lpaho-mqtt3cs"
```
gaborznagy, line 1320: FYI: On my computer, with local-built paho-mqtt lib this caused that mqtt lib was not found. Turned out, SSL support is disabled by default in paho-mqtt. :)

gaborznagy (approved):

configure.ac
MrAnno: Debian, Ubuntu, Fedora, and Arch AUR enables TLS, I checked them.

gaborznagy: I've tested it too on Debian Testing :+1.
This was just a small note, in case someone else builds paho-mqtt locally with default flags.

PR #3748, opened 2021-08-03
Handle redis reply error

Kokan:
If redis reply returned with error, we did not detect it. Simple reproduction is to configure a command that does not exists, and syslog-ng reports is as written message.


kira-syslogng:
Build SUCCESS

gaborznagy (approved):

MrAnno (approved):

PR #3750, opened 2021-08-09
disk-buffer(): fix a rare memory leak

MrAnno:
log_queue_disk_drop_message() is responsible for dropping a message
when the disk-buffer reaches its full capacity.

In case flow-control is enabled, the source will stop before the queue
becomes full, so no message loss is possible, but there are special/rare
cases when the user configures mem-buf-length() or mem-buf-size()
incorrectly, and the message has to be dropped.

log_queue_disk_drop_message() must force-suspend the source in the above
case, but we had already lost one single message by then.

log_msg_drop() not only acks the message with the specified ack type,
it also calls log_msg_unref() on the message, which should be done in
both cases to avoid leaking the memory.

### DIFF
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -351,17 +351,6 @@ _push_head(LogQueue *s, LogMessage *msg, const LogPathOptions *path_options)
   g_static_mutex_unlock(&s->lock);
 }
 
-static void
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

alltilla (approved):
Nice catch!

szemere (approved):

Issue #3701, opened 2021-06-10
netbsd tcp() source problem


czanik:
Working on NetBSD 9.2 I started to update syslog-ng in pkgsrc from version 3.17.2 to 3.32.1. Collecting local logs and over UDP seems to work, but the tcp() source does not. Interestingly enough it shows up in syslog-ng's debug output on startup:
```
[2021-06-10T09:41:16.528278] Module loaded and initialized successfully; module='affile'
[2021-06-10T09:41:16.528457] Running application hooks; hook='6'
[2021-06-10T09:41:16.529528] Accepting connections; addr='AF_INET(172.16.167.136:514)'
[2021-06-10T09:41:16.530396] Running application hooks; hook='1'
[2021-06-10T09:41:16.530493] Running application hooks; hook='7'
```
But it does not show up in the output of netstat and when I try to send logs or just simply telnet to the port, connection is refused:
```
localhost# telnet 172.16.167.136 514
Trying 172.16.167.136...
telnet: Unable to connect to remote host
localhost# 
```
I used the following two patches to get syslog-ng working on NetBSD, updated from pkgsrc:
[patch-lib_signal-handler.txt](https://github.com/syslog-ng/syslog-ng/files/6629362/patch-lib_signal-handler.txt)
[patch-modules_system-source_system-source.txt](https://github.com/syslog-ng/syslog-ng/files/6629363/patch-modules_system-source_system-source.txt)
Without these syslog-ng crashes on startup and system() cannot find local log sources.

alltilla:
signal-handler related issues: #3391 #3445

MrAnno:
We agreed to do the following:
1. We'll merge `patch-modules_system-source_system-source.txt`, so it no longer have to be maintained separately.
2. For the signal-handler issue, we'll check whether we can come up a more portable `sigaction` trickery.
3. @OverOrion is investigating the `network()/tcp()` source issue (`listen()` does not seem to work).

OverOrion:
With the help of the team I was able to identify the root cause: 
It turned out that the `network` source had never worked on NetBSD but the issue manifested only after 4 hours.
Syslog-ng set the `TCP_KEEPALIVE` values on a `listening` socket which causes NetBSD to stop accepting connections after a certain amount of time. On all other platforms this setting did not cause any issues as these values are inherited by the clients.
With syslog-ng-3.29.1 the default values for `TCP_KEEPALIVE` had been changed in #3357. 

We are working on a fix.

[...]

PR #3751, opened 2021-08-10
afsocket: move TCP keepalive options setting after `accept()` call

OverOrion:
This should improve POSIX (and by extension NetBSD) compatibility as
the POSIX standard does not guarantee the inheritance of options between listening and peer sockets.
I moved the `socket_options_inet_setup_tcp_keepalive_timers()` function
after `accept()`, and created a new virtual function (`setup_peer_socket()`) in `SocketOptions`
which is redefinied in `SocketOptionsInet`.

Solves #3701

Signed-off-by: Parrag Szilárd <szilard.parrag@gmail.com>

### DIFF
--- a/modules/afsocket/socket-options-inet.c
+++ b/modules/afsocket/socket-options-inet.c
@@ -136,7 +136,6 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
   if (!socket_options_setup_socket_method(s, fd, addr, dir))
     return FALSE;
 
-  socket_options_inet_setup_tcp_keepalive_timers(self, fd);
 
   if (self->interface_name)
     {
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
I think it's worth mentioning that the actual bug is coming from the fact that setting TCP keepalive and SO_KEEPALIVE options on the listening socket is not portable. Linux guarantees that those options will be inherited by the peer sockets, but it's not a standard.

NetBSD behaved really weirdly when we did that: it closed the listening socket after the specified TCP keepalive timeout and rendered the entire `network()` source inoperable. (Let's say the behavior was undefined because we set TCP keepalive timers on the listening socket, which is not really allowed on NetBSD. FreeBSD, however, does the same as Linux).

MrAnno (commented):
LGTM.

There are some problems with the current `SocketOptions` hierarchy as it sets TCP options for UDP sockets as well, etc.

We probably need to move a few other options to the new `setup_peer_socket()` method as well, but I think it's a good start.

Please write a bugfix news entry for this PR.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

MrAnno (dismissed):

modules/afsocket/socket-options-inet.c
```diff
@@ -233,6 +232,17 @@ socket_options_inet_setup_socket(SocketOptions *s, gint fd, GSockAddr *addr, AFS
   return TRUE;
 }
 
+static gboolean
+socket_options_inet_setup_peer_socket(SocketOptions *s, gint fd, GSockAddr *addr)
+{
+  SocketOptionsInet *self = (SocketOptionsInet *) s;
+
+  socket_options_setup_peer_socket_method(s, fd, addr);
```
Kokan, line 240: Please do error handling based on this method return  value.

kira-syslogng:
Build SUCCESS

Kokan (approved):

MrAnno (approved):

PR #3752, opened 2021-08-10
qdisk: fix read and write head guessing

alltilla:
We need the next tail's position and the head's position info in the reliable disk-buffer, because it is matching its memory queue's messages to the qdisk's messages by their position in the qdisk.

Currently we guess these positions on the caller side, based on the `write_head`'s and `read_head`'s position.
This guess is not always right, because for example `push_tail()` can change the `write_head` before writing to the disk.

I think accessing any of the three QDisk::`..._head`s publicly should not be allowed, because it is an implementation detail, but I did not want to refactor the whole code around that. So I only added the 2 function, which I think would be the correct to have, and used them in the respective parts of the code.

---

In the news entry, I mention that we fixed a rare bug where the reliable disk-buffer did not ever resume after triggering flow-control. I did not observe this behavior anywhere, I just found this bug in the code, and thought about what possible unexpected behavior it can cause.

### DIFF
--- a/modules/diskq/logqueue-disk-reliable.c
+++ b/modules/diskq/logqueue-disk-reliable.c
@@ -86,7 +86,7 @@ _ack_backlog(LogQueue *s, gint num_msg_to_ack)
   for (i = 0; i < num_msg_to_ack; i++)
     {
       gint64 pos;
-      if (qdisk_get_backlog_head (self->super.qdisk) == qdisk_get_reader_head (self->super.qdisk))
+      if (qdisk_get_backlog_head(self->super.qdisk) == qdisk_get_head_position(self->super.qdisk))
         {
           goto exit_reliable;
         }
@@ -212,7 +212,7 @@ _pop_head(LogQueue *s, LogPathOptions *path_options)
     {
       gint64 *temppos = g_queue_pop_head (self->qreliable);
       gint64 pos = *temppos;
-      if (pos == qdisk_get_reader_head (self->super.qdisk))
+      if (pos == qdisk_get_head_position(self->super.qdisk))
         {
           msg = g_queue_pop_head (self->qreliable);
           log_queue_memory_usage_sub(s, log_msg_get_size(msg));
@@ -255,7 +255,7 @@ _pop_head(LogQueue *s, LogPathOptions *path_options)
         }
       else
         {
-          qdisk_set_backlog_head(self->super.qdisk, qdisk_get_reader_head(self->super.qdisk));
+          qdisk_set_backlog_head(self->super.qdisk, qdisk_get_head_position(self->super.qdisk));
         }
       log_queue_queued_messages_dec(s);
     }
@@ -286,7 +286,7 @@ _push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *path_options)
 
   LogPathOptions local_options = *path_options;
 
-  gint64 last_wpos = qdisk_get_writer_head (self->super.qdisk);
+  gint64 message_position = qdisk_get_next_tail_position(self->super.qdisk);
   if (!qdisk_push_tail(self->super.qdisk, serialized_msg))
     {
       /* we were not able to store the msg, warn */
@@ -311,7 +311,7 @@ _push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *path_options)
        * the message is written but into the overflow area
        */
       gint64 *temppos = g_malloc (sizeof(gint64));
-      *temppos = last_wpos;
+      *temppos = message_position;
       g_queue_push_tail (self->qreliable, temppos);
       g_queue_push_tail (self->qreliable, msg);
       g_queue_push_tail (self->qreliable, LOG_PATH_OPTIONS_TO_POINTER(path_options));
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -461,12 +461,6 @@ _try_reading_record_length(QDisk *self, guint32 *record_length)
 {
   guint32 read_record_length;
   gssize bytes_read = _read_record_length_from_disk(self, &read_record_length);
-  if (bytes_read == 0)
-    {
-      /* hmm, we are either at EOF or at hdr->qout_ofs, we need to wrap */
-      self->hdr->read_head = QDISK_RESERVED_SPACE;
-      bytes_read = _read_record_length_from_disk(self, &read_record_length);
-    }
 
   if (!_is_record_length_valid(self, bytes_read, read_record_length))
     return FALSE;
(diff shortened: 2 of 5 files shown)


alltilla:
@kira-syslogng do stresstest

kira-syslogng:
Build SUCCESS

kira-syslogng:

alltilla:
Rebased to master.

kira-syslogng:
Build SUCCESS

modules/diskq/logqueue-disk-reliable.c
```diff
@@ -86,7 +86,7 @@ _ack_backlog(LogQueue *s, gint num_msg_to_ack)
   for (i = 0; i < num_msg_to_ack; i++)
     {
       gint64 pos;
-      if (qdisk_get_backlog_head (self->super.qdisk) == qdisk_get_reader_head (self->super.qdisk))
+      if (qdisk_get_backlog_head(self->super.qdisk) == qdisk_get_head_position(self->super.qdisk))
```
gaborznagy, line 89: `qdisk_get_reader_head` and the new function `qdisk_get_head_position` looks the same. Can you just rename the old function, please? It will be easier later to follow thorugh the git history.

alltilla: Actually I do not think, that `qdisk_get_reader_head()` should even be a thing. I would not rename it, but maybe I can remove it at this point, and change all its other users.

modules/diskq/qdisk.h
```diff
@@ -54,6 +54,8 @@ gint64 qdisk_get_empty_space(QDisk *self);
 gboolean qdisk_push_tail(QDisk *self, GString *record);
 gboolean qdisk_pop_head(QDisk *self, GString *record);
 gboolean qdisk_remove_head(QDisk *self);
+gint64 qdisk_get_next_tail_position(QDisk *self);
+gint64 qdisk_get_head_position(QDisk *self);
```
MrAnno, line 58: I think publishing positions for anything other than tests is not a good idea, but currently we do that to track things outside, in memory-only queue segments.

It looks better and a little bit simpler with these new methods, but ideally we should not need these at all in my opinion (that was the reason why the bug was actually introduced, because we tried to be clever about the internals of `QDisk`, but something was changed inside, therefore something else broke outside).

We won't have the time to refactor/fix everything this time, but I think we should mention the above in the commit message, so we won't forget it completely.

modules/diskq/qdisk.c
```diff
@@ -461,12 +461,6 @@ _try_reading_record_length(QDisk *self, guint32 *record_length)
 {
   guint32 read_record_length;
   gssize bytes_read = _read_record_length_from_disk(self, &read_record_length);
-  if (bytes_read == 0)
-    {
-      /* hmm, we are either at EOF or at hdr->qout_ofs, we need to wrap */
-      self->hdr->read_head = QDISK_RESERVED_SPACE;
-      bytes_read = _read_record_length_from_disk(self, &read_record_length);
```
MrAnno, line 468: I don't understand the reasoning in the commit message. You mentioned 62152b1, but that was an equivalent transformation.

Are you sure this is dead code? I mean, for example, we don't update positions if either `_try_reading_record_length()` or `_read_record_from_disk()` fails.

MrAnno: We discussed this one IRL, it is dead code.
In case any of the mentioned calls fail, that means the disk-buffer is corrupt, and we'll move the current disk-buffer and create a new one.

gaborznagy: On error cases, the disk-buffer file is replaced by a new disk-buffer file as it seems to be corrupted.

gaborznagy: On the other hand, could a truncate by `qdisk_push_tail` cause that read_head is at end of file?

alltilla: @gaborznagy 

> On the other hand, could a truncate by qdisk_push_tail cause that read_head is at end of file?

That cannot happen, as we only truncate in `qdisk_push_tail()` when `write_head > read_head`.

modules/diskq/qdisk.h
alltilla: Thanks, I agree. I added some comments about it in the commit msg:
```
    qdisk: add and use get_next_tail_position() and get_head_position()                         

    We need these info in the reliable disk-buffer, because it is matching                      
    its memory queue's message to the qdisk's messages by their position                        
    in the qdisk.                                                                               

    Currently we guess these positions on the caller side, based on the                         
    write_head's and read_head's position.      

    This guess is not always right, because for example push_tail() can
    change the write_head before writing to the disk.                                           

    ------------------------------------------------------------------------                    

    This commit introduces a new interface around QDisk, where we do not                        
    make the internal read-, write- and backlog-head public, but give                           
    functions closer to a queue implementation (head and tail).                                 

    This is still not the ideal interface. Ideally QDisk should not give                        
    any info about the records position on the disk, as it is still an                          
    implementation detail, somehow the callor site should keep track of the
    messages' location, with an index array, or something like that.

    Unfortunately we cannot refactor that much within the scope of this PR,                     
    that will be done later.                                                                    

    ------------------------------------------------------------------------                    

    The other problem is that we still have the 2 not so perfect interface
    in QDisk simultaneously. The reason behind not removing the old one is
    that we do not have a good place for get_backlog_head() in the new                          
    interface.                                                                                  

    Though, if we say, that QDisk has a backlog-like logic inside                               
    - which it has - then the moving of the backlog_head should be done
    in QDisk, not with public getter and setter functions.                                      

    If the backlog_head modification happens solely in QDisk, we will not                       
    need any public function to do so, so we can completely remove the old                      
    interface and use the new one.                                                              

    We think, that the two mentioned problem can be fixed in one go, and we                     
    plan to do it later.

```

kira-syslogng:
Build SUCCESS

MrAnno (approved):

modules/diskq/qdisk.c
gaborznagy: We discussed and checked this: we don't see it possible that `read_head` would be at EOF position:
- during `push_tail` we can truncate to `write_head` if `read_head` follows `write_head` (read_head < write_head and not read_head <= write_head !!!). In this case even if it's truncated, read_head would have at least one message until EOF
- `qdisk_pop_head` will not be called on an empty disk-buffer: On LogQueueDisk level pop_head is not called if the queue is empty. This is important as it can happen that `qdisk_pop_head` doesn't set `read_head` to the file's beginning even if the queue is empty due to `truncate-size-ratio`.

However, if a read at EOF would still happen, the record_length reading will got a "short read" error, and the disk-buffer will be marked corrupted and a new one would be started.

modules/diskq/logqueue-disk-reliable.c
gaborznagy: As discussed in https://github.com/syslog-ng/syslog-ng/pull/3752#discussion_r688478400 2 interfaces will co-exist fow now.

gaborznagy (approved):

PR #3763, opened 2021-08-26
Compile and upload binary packages (nightly and release-draft)

szemere:
This is the first step to create a repository from the build artifacts.
    Key points of this commit:

    - GitHub Actions - currently - do not support includes or other forms
      of reuse of the Workflows. As a result, we build packages in three
      different locations: packages.yml, nightly.yml, draft-release.yml.

      To make it easier to validate the content of draft-release.yml, the
      packages step in nightly.yml is almost identical. (Only the upload
      path differs.)

    - Currently, our official release artifact is the release tarball.
      We want to ensure that it is possible to create binary packages
      ONLY from the tarball. To strengthen this, the packages jobs do
      not use our git source tree.

      Note: By not using our source tree, it is not possible to use the
      dbld's cache-image-<image-name> functionality. There is a plan to
      store information inside the tarball for this. Until than, the only
      solution is to rebuild the builder image for each run.

For the reviewers:
The AZURE_SAS_TOKEN secret is already added to the syslog-ng/syslo-ng repository.


OverOrion (dismissed):

.github/workflows/packages.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           path: dbld/build/*.tar.gz
 
   packages:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
Kokan, line 33: There are some `ubuntu-18.04` in devshell, I think those can be changed for the same reason.

dbld/Makefile.am
```diff
@@ -31,6 +31,7 @@ EXTRA_DIST += \
 	dbld/images/fedora-33.dockerfile \
 	dbld/images/debian-stretch.dockerfile \
 	dbld/images/debian-buster.dockerfile \
+	dbld/images/debian-sid.dockerfile \
```
Kokan, line 34: The `debian-testing.dockerfile` also missing.

modules/afmongodb/afmongodb.c
```diff
@@ -144,7 +144,7 @@ _format_persist_name(const LogPipe *s)
 static inline void
 _uri_set_default_server_selection_timeout(mongoc_uri_t *uri_obj)
 {
-#if SYSLOG_NG_HAVE_DECL_MONGOC_URI_SET_OPTION_AS_INT32
+#if SYSLOG_NG_HAVE_DECL_MONGOC_URI_SET_OPTION_AS_INT32 && SYSLOG_NG_HAVE_DECL_MONGOC_URI_SERVERSELECTIONTIMEOUTMS
```
Kokan, line 147: Do we have to check for both `MONGOC_URI_SERVERSELECTIONTIMEOUTMS` and `mongoc_uri_set_option_as_int32` ? Maybe it is guaranteed that if `MONGOC_URI_SERVERSELECTIONTIMEOUTMS` the `mongoc_uri_set_option_as_int32` also exists ?

.github/workflows/draft-release.yml
```diff
@@ -85,14 +85,19 @@ jobs:
...
+        - [ 'debian-sid',       'deb' ]
+        - [ 'ubuntu-xenial',    'deb' ]
+        - [ 'ubuntu-bionic',    'deb' ]
+        - [ 'ubuntu-focal',     'deb' ]
+        - [ 'centos-7',         'rpm' ]
+        - [ 'fedora-33',        'rpm' ]
       fail-fast: false
+    env:
+      DISTRIBUTION: ${{ matrix.dbld-builder-images[0] }}
+      PACKAGE_MANAGER: ${{ matrix.dbld-builder-images[1] }}
```
Kokan, line 100: If a package is a `deb` is should not mean that the upload is needed. As I only see use case for uploading or not, this should represent just that and not package manager.

This could be resolved either having yet an other job that only uploads (and the distro list would contains the distroes that needs uploading) or you could just make the 2nd parameter like upload: yes/no.

.github/workflows/nightly-packages.yml
```diff
@@ -0,0 +1,80 @@
+name: Nightly packages
+
+on:
+  workflow_dispatch:
+  schedule:
```
gaborznagy, line 5: We have a `cron` trigger in `packages` workflow as well.

How about adding an additonal step in `packages` about the package upload instead of copying the entire packages workflow?

We could protect the package upload by checking if the event source is `cron` and the branch is master.

configure.ac
```diff
@@ -1268,6 +1268,7 @@ CPPFLAGS_SAVE="$CPPFLAGS"
 CPPFLAGS="$CPPFLAGS $LIBMONGO_CFLAGS"
 
 AC_CHECK_DECLS([mongoc_uri_set_option_as_int32], [], [], [[#include <mongoc.h>]])
+AC_CHECK_DECLS([MONGOC_URI_SERVERSELECTIONTIMEOUTMS], [], [], [[#include <mongoc.h>]])
```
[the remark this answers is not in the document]
MrAnno: Thanks

.github/workflows/draft-release.yml
MrAnno: I've added an "upload" flag for each matrix element.
Laci will refactor this part even further when he returns from vacation.

https://github.com/MrAnno/syslog-ng/actions/runs/1210275712

MrAnno (approved):
I think the `version-bump` action should also generate packages in order to validate its output.(Uploading those packages is not necessary, of course.)

We can address this and other optional review notes in a separate PR.
IMHO we should merge this before starting the v3.34.1 release, so we can test the new `draft-release` packages job.

kira-syslogng:
Build SUCCESS

Kokan (approved):

alltilla (approved):

MrAnno:
https://github.com/syslog-ng/syslog-ng/actions/runs/1212653848

.github/workflows/draft-release.yml
```diff
@@ -47,6 +47,12 @@ jobs:
         run: |
           ./dbld/rules release VERSION=`cat VERSION`
 
+      - name: Store release tarball as artifact
```
alltilla, line 50: The error handling can be improved a bit.
The problems are:
 * The comment to the Version Bump PR is sent prematurely.
 * The draft release is created before trying to upload the packages to the APT repo. The upload could fail and we are left with a draft release which will have to be removed manually.

I propose something like this:
 * `create-release-tarball` job, which uploads the tarball as an artifact
 * `packages` job, which needs the `create-release-tarball` job, and generates and uploads deb packages to the APT repo
 * `draft-release` job, which needs the `create-release-tarball` and `packages` jobs, and it creates the draft release t o Github
 * `comment-workflow-status` job, which needs the `packages` and `draft-release` jobs, and it comments on the Version Bump PR, merging the 2 results.

PR #3771, opened 2021-09-01
Dbld/images/Kira: Add gdb tool to kira image

mitzkia:
- with help of gdb tool testdb can extract a backtrace from core files

Signed-off-by: Andras Mitzki <andras.mitzki@quest.com>

### DIFF
--- a/dbld/packages.manifest
+++ b/dbld/packages.manifest
@@ -117,6 +117,7 @@ libmongoc-dev               [debian, ubuntu-xenial]
 # Tools required to run @kira-syslogng, our testbot.
 #############################################################################
 
+gdb                             [kira]
 faketime                        [kira]
 libdbd-mysql                    [kira]
 lsb-release                     [kira]


kira-syslogng:
Build SUCCESS

OverOrion (approved):

dbld/packages.manifest
```diff
@@ -117,6 +117,7 @@ libmongoc-dev               [debian, ubuntu-xenial]
 # Tools required to run @kira-syslogng, our testbot.
 #############################################################################
 
+gdb                             [kira]
```
gaborznagy, line 120: I was not aware that we have the same packages multiple times in packages.manifest: both `kira` and `devshell` have `netcat`, `lsof`, etc.

See https://github.com/syslog-ng/syslog-ng/pull/3771/files#diff-3e3755d0abaea8947f179974dcaba4899f63862914d0e060d43c78e84b500a16R144-R147

If this is intentional, then looks good to me.

mitzkia: Yes, it was intentional due to recently we have got lot of crashes inside Kira run, where testdb prints:
"(127, '/bin/sh: 1: gdb: not found\n')".
With help of gdb testdb can extract backtrace from core files.

mitzkia: Other packages are not relevant for now

gaborznagy (approved):

PR #3780, opened 2021-09-13
dbld: change default python case to python3

OverOrion:
As new distributions use `python3` by default it makes sense to explicitly state older platforms which use `python2` instead of the other way around, so it is not necessary to add that new platform to the `python3` case.
Signed-off-by: Parrag Szilárd <szilard.parrag@gmail.com>

### DIFF
--- a/dbld/builddeps
+++ b/dbld/builddeps
@@ -154,17 +154,17 @@ function install_yum_packages {
 
 function install_pip_packages {
     case "${IMAGE_PLATFORM}" in
-        fedora-*)
-            python_executables="python"
+        ubuntu-xenial|ubuntu-bionic)
+            python_executables="python2"
             ;;
-        ubuntu-focal|tarball)
-            python_executables="python3"
+        debian-stretch|debian-buster)
+            python_executables="python2"
             ;;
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

dbld/builddeps
```diff
@@ -157,14 +157,17 @@ function install_pip_packages {
         fedora-*)
```
szemere, line 157: checked: fedora-33 and fedora-34 both have "python3". So we do not need the "python" alias anymore. (We can leave it to the default branch.)

kira-syslogng:
Build SUCCESS

alltilla:
Could you show the successful image build runs for our dbld images?
If I understand correctly, running the nightly-packages job on your fork on this feature branch should show us, if there was a problem.

kira-syslogng:
Build SUCCESS

OverOrion:
[Succesful image build run](https://github.com/OverOrion/syslog-ng/actions/runs/1240969001)

szemere (approved):

Kokan (approved):

PR #3784, opened 2021-09-13
build: support build with bison 3.8

alltilla:
Starting with bison 3.8, bison generates the function declarations we have already done in `CFG_PARSER_DECLARE_LEXER_BINDING`. https://github.com/akimd/bison/blob/master/NEWS#L19

The only difference between the generated one, and our macro is a `const` qualifier before `CFG_LTYPE *yylloc`.

In order to work with both 3.8 and older bison versions, we should keep our manual function definitions in the code, but it should match with the one, bison generates.

Green macOS CI job should validate this PR: https://formulae.brew.sh/formula/bison

### DIFF
--- a/lib/cfg-parser.c
+++ b/lib/cfg-parser.c
@@ -233,7 +233,7 @@ _print_underline(const gchar *line, gint whitespace_before, gint number_of_caret
 }
 
 static void
-_print_underlined_source_block(CFG_LTYPE *yylloc, gchar **lines, gint error_index)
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

OverOrion (approved):
LGTM, thank you! :)

bazsi (approved):

MrAnno (approved):
👍🏻

Issue #3445, opened 2020-10-02
Support shared library load order randomization


gaborznagy:
Originally it was reported in #3391 that on HardenedBSD/OPNsense platform syslog-ng can crash in a call to `sigaction`.

We later investigated that our sigaction wrapper mechanism in [lib/signal-handler.c](https://github.com/syslog-ng/syslog-ng/blob/master/lib/signal-handler.c) conflicts with the shared object load order randomization feature of HardenedBSD [1].
Precisely the `dlsym(RTLD_NEXT, sigaction)` call depends on the load ordering of the shared libraries: if libsyslog-ng is loaded after libc, then `sigaction` will not be found resulting in a segmentation fault.

As explained in https://github.com/syslog-ng/syslog-ng/issues/3391#issuecomment-686969707
>  https://www.freebsd.org/cgi/man.cgi?query=dlsym 
> If dlsym() is called with the special handle RTLD_NEXT, then the search
for the symbol is limited to the shared objects which were loaded after
the one issuing the call to dlsym(). Thus, if the function is called
from the main program, all the shared libraries are searched. If it is
called from a shared library, all subsequent shared libraries are
searched.

We came to the conclusion that we should include the signal-handler code in the main binary, so `sigaction` in libc will be found despite the load order of shared objects.

[1] https://hardenedbsd.org/article/shawn-webb/2014-12-14/shared-object-load-order-randomization

PR #3786, opened 2021-09-16
Portability fixes

0-wiz-0:


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

MrAnno (dismissed):
Hi,

AFAIK these are the remaining patches from the NetBSD syslog-ng packaging.

The `sigaction` dlsym trickery that we use in `signal-handler.c` has additional issues due to its unportable nature, for example, it does not work with shared object load order randomization (#3445). We are planning to reimplement this part in a more portable manner. Until then, I think this `__NetBSD__` ifdef would not hurt :)

0-wiz-0:
@MrAnno Yes, I took those from the pkgsrc package.
There is one additional patch but I don't know how much of that you'd want to take -- please check it out:

https://github.com/NetBSD/pkgsrc/blob/trunk/sysutils/syslog-ng/patches/patch-ac

kira-syslogng:
Build SUCCESS

MrAnno:
Oh, I did not know about this script (it seems to be a nice syslogd to syslog-ng config converter).
The `contrib` folder is not actively maintained, but I think we could add NetBSD support to this script as well, if you use that. 
However, we need to check the `@UNAME@` part.

@alltilla What do you think?

lib/signal-handler.c
```diff
@@ -84,12 +84,16 @@ signal_handler_exec_external_handler(gint signum)
 static int
 _original_sigaction(int signum, const struct sigaction *act, struct sigaction *oldact)
 {
+#ifdef __NetBSD__
+  __libc_sigaction14(signum, act, oldact);
```
MrAnno, line 88: The original sigaction result should be returned.

```suggestion
  return __libc_sigaction14(signum, act, oldact);
```

0-wiz-0:
On NetBSD, uname is in `/usr/bin/uname`. Could we remove the use of the path component in the script, since both /bin and /usr/bin are usually in the user's paths?
I've left out this part for now but added the uncontroversial (I hope) parts.

kira-syslogng:
Build SUCCESS

MrAnno:
@0-wiz-0 Thank you for addressing everything.

I'm not an `awk` wizard, but calling `uname` without specifying the full path worked for me as well. I've seen `awk` scripts doing that.

One last thing: unfortunately, there is a failing commit check in our CI.
- We prefer signing all of our commits using git's sign-off feature. You can do that using the `--signoff` flag of the `git commit` command, or if you want to add your sign-off line to an existing commit, `git rebase` has the same flag.
- Commit `Avoid unportable test(1) operator. ` does not specify the "unit"/"module" where the change was made. For example, the following would be accepted by the commit-check job:
  - `build: avoid unportable test(1) operator` or
  - `autotools: avoid unportable test(1) operator`, etc.

0-wiz-0:
Thanks for the explanations, @MrAnno ! I think I've addressed all of these in the last push.

MrAnno (approved):
👍🏻

kira-syslogng:
Build SUCCESS

Kokan (approved):

PR #3792, opened 2021-09-24
Add TLS keylog capability

### LINKED ISSUE #3592: add possibility to save/log TLS symmetric keys in case of network() and syslog() sources and destinations
## Description of the problem
When troubleshooting TLS problems it would come in handy if we could unpack the TLS encryption to find out what traffic is actually traversing the network.
This is currently possible in the http* and derived destinations, but not in the traditional network() and syslog() sources and destinations.

## Proposed solution
Achieving this goal would be technically feasible as OpenSSL offers an interface for doing this:
https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_keylog_callback.html
The exact questions of how this could be implemented and how this possibility should be made available for the user will need to be decided on. Maybe a tls() block option?

OverOrion:
This PR adds a new TLS block option: `keylog-file()` which will be used to export TLS secrets between clients/servers. This should aid debugging sessions as the specified file can be loaded into Wireshark and then Wireshark is able to decrypt the traffic.

Fixes #3592

### DIFF
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -31,6 +31,7 @@
 #include <arpa/inet.h>
 #include <unistd.h>
 #include <stdio.h>
+#include <glib/gstdio.h>
 #include <openssl/x509_vfy.h>
 #include <openssl/x509v3.h>
 #include <openssl/err.h>
@@ -45,6 +46,12 @@ struct _TLSContext
   TLSMode mode;
   gint verify_mode;
   gchar *key_file;
+  struct
+  {
+    gchar *keylog_file_path;
+    FILE *keylog_file;
+    GMutex keylog_file_lock;
+  };
   gchar *cert_file;
   gchar *dhparam_file;
   gchar *pkcs12_file;
@@ -335,6 +342,46 @@ _set_sni_in_client_mode(TLSSession *self)
   return FALSE;
 }
 
+void
+_write_line_to_keylog_file(const char *file_path, const char *line, FILE *keylog_file, GMutex *mutex)
+{
+  if(!keylog_file)
+    return;
+
+
+  g_mutex_lock(mutex);
+  gint ret_val = fprintf(keylog_file, "%s\n", line);
+  if (ret_val != strlen(line)+1)
+    msg_error("Couldn't write to TLS keylogfile", evt_tag_errno("error", ret_val));
+
+  fflush(keylog_file);
+  g_mutex_unlock(mutex);
+}
+
+static inline void
+_dump_tls_keylog(const SSL *ssl, const char *line)
+{
+  if(!ssl)
+    return;
+
+  TLSSession *self = SSL_get_app_data(ssl);
+  _write_line_to_keylog_file(self->ctx->keylog_file_path, line, self->ctx->keylog_file, &self->ctx->keylog_file_lock);
+}
+
+static gboolean
+tls_session_keylog_setup_file(TLSSession *self, const char *keylog_file_path)
+{
+  self->ctx->keylog_file = g_fopen(keylog_file_path, "a");
+  if(!self->ctx->keylog_file)
+    {
+      msg_error("Error opening keylog-file",
+                evt_tag_str(EVT_TAG_FILENAME, keylog_file_path),
+                evt_tag_error(EVT_TAG_OSERROR));
+      return FALSE;
+    }
+  return TRUE;
+}
+
 static TLSSession *
 tls_session_new(SSL *ssl, TLSContext *ctx)
 {
@@ -353,6 +400,15 @@ tls_session_new(SSL *ssl, TLSContext *ctx)
       tls_context_unref(self->ctx);
       g_free(self);
       return NULL;
+    }
+  if(ctx->keylog_file_path)
+    {
+      if(!tls_session_keylog_setup_file(self, self->ctx->keylog_file_path))
+        {
+          tls_context_unref(self->ctx);
+          g_free(self);
+          return NULL;
+        }
     }
 
   return self;
@@ -837,6 +893,11 @@ _tls_context_free(TLSContext *self)
   g_free(self->cipher_suite);
   g_free(self->ecdh_curve_list);
   g_free(self->sni);
+  g_free(self->keylog_file_path);
+
+  if(self->keylog_file)
+    fclose(self->keylog_file);
+
   g_free(self);
 }
 
@@ -997,6 +1058,21 @@ tls_context_set_key_file(TLSContext *self, const gchar *key_file)
   self->key_file = g_strdup(key_file);
   SSL_CTX_set_default_passwd_cb(self->ssl_ctx, _pem_passwd_callback);
   SSL_CTX_set_default_passwd_cb_userdata(self->ssl_ctx, self->key_file);
+}
+
+gboolean
+tls_context_set_keylog_file(TLSContext *self, gchar *keylog_file_path)
+{
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L
+  g_free(self->keylog_file_path);
+  msg_warning_once("WARNING: TLS keylog file has been set up, it should only be used during debugging sessions.",
+                   evt_tag_str("keylog-file", keylog_file_path));
+  self->keylog_file_path = g_strdup(keylog_file_path);
+  SSL_CTX_set_keylog_callback(self->ssl_ctx, _dump_tls_keylog);
+  return TRUE;
+#else
+  return FALSE;
+#endif
 }
 
 void
--- a/lib/tlscontext.h
+++ b/lib/tlscontext.h
@@ -118,6 +118,7 @@ gint tls_context_get_verify_mode(const TLSContext *self);
 void tls_context_set_verify_mode(TLSContext *self, gint verify_mode);
 void tls_context_set_key_file(TLSContext *self, const gchar *key_file);
 void tls_context_set_cert_file(TLSContext *self, const gchar *cert_file);
+gboolean tls_context_set_keylog_file(TLSContext *self, gchar *keylog_file_path);
 void tls_context_set_pkcs12_file(TLSContext *self, const gchar *pkcs12_file);
 void tls_context_set_ca_dir(TLSContext *self, const gchar *ca_dir);
 void tls_context_set_crl_dir(TLSContext *self, const gchar *crl_dir);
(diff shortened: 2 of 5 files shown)


lib/tlscontext.c
```diff
@@ -697,6 +675,50 @@ tls_context_load_key_and_cert(TLSContext *self)
   return TLS_CONTEXT_OK;
 }
 
+static void _write_line_to_keylog_file_mutex(const char *file_path, const char *line)
+{
+  FILE *keylog_file;
+  gint ret_val;
+  static GStaticMutex mutex;
```
MrAnno, line 682: The lock is global this way, which will block all TLS key writes, even if they are in separate keylog files.


I think this should go to `TLSContext` as a field, for example: `keylog_file_lock`.
This would be safe, because every source/destination driver has a separate keylog file, and in case of sources, all TLS connections from the same driver shares a single `TLSContext` instance.

lib/tlscontext.c
```diff
@@ -697,6 +675,50 @@ tls_context_load_key_and_cert(TLSContext *self)
...
+  g_static_mutex_lock(&mutex);
+  keylog_file = fopen(file_path, "a");
+  if(keylog_file)
+    {
```
MrAnno, line 686: Please report fopen() errors. You can also early-return in case of an error.

lib/tlscontext.h
```diff
@@ -29,6 +29,8 @@
 #include "atomic.h"
 #include <openssl/ssl.h>
 
+#define _TLS_KEYLOG_INDEX 0
```
MrAnno, line 32: We should keep this in the C file (it's an implementation detail)

lib/tlscontext.c
```diff
@@ -697,6 +675,50 @@ tls_context_load_key_and_cert(TLSContext *self)
   return TLS_CONTEXT_OK;
 }
 
+static void _write_line_to_keylog_file_mutex(const char *file_path, const char *line)
```
MrAnno, line 678: I think it's not necessary to state that this function locks the keylog_file mutex.
I'd convert this to a `TLSContext` method, and just simply lock `self->keylog_file_lock`.

lib/tlscontext.c
```diff
@@ -999,6 +1026,15 @@ tls_context_set_key_file(TLSContext *self, const gchar *key_file)
   SSL_CTX_set_default_passwd_cb_userdata(self->ssl_ctx, self->key_file);
 }
 
+void
+tls_context_set_keylog_file(TLSContext *self, const gchar *keylog_file)
+{
+  g_free(self->keylog_file);
+  msg_warning("WARNING: TLS keylog file has been set up, it should only be used during debugging sessions, NOT in production environments",
```
MrAnno, line 1033: Let's use `msg_warning_once()`.

lib/tlscontext.c
```diff
@@ -697,6 +675,50 @@ tls_context_load_key_and_cert(TLSContext *self)
...
+{
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L
+  {
+    SSL_CTX_set_ex_data(self->ssl_ctx, _TLS_KEYLOG_INDEX, self->keylog_file);
+    SSL_CTX_set_keylog_callback(self->ssl_ctx, _dump_tls_keylog);
+  }
+#else
+  msg_error("SSL_CTX_set_keylog_callback function is not available in your OpenSSL version, please upgrade it to at least 1.1.1",
+            evt_tag_str("Your version", OPENSSL_VERSION_NUMBER));
+#endif
```
MrAnno, line 719: Ifdefed compatibility implementations should go into `lib/compat/openssl_support`.

Since this is not an alternative implementation, just error reporting, I'd report this error sooner.
Why don't we add a `boolean` return value to `tls_context_set_keylog_file()`? That way, we could report the missing functionality as a config parsing error with `CHECK_ERROR()`.

lib/tlscontext.c
```diff
@@ -697,6 +675,50 @@ tls_context_load_key_and_cert(TLSContext *self)
   return TLS_CONTEXT_OK;
 }
 
+static void _write_line_to_keylog_file_mutex(const char *file_path, const char *line)
+{
+  FILE *keylog_file;
+  gint ret_val;
+  static GStaticMutex mutex;
+  g_static_mutex_lock(&mutex);
+  keylog_file = fopen(file_path, "a");
```
MrAnno, line 684: Opening and closing the file for every line might be problematic.
We could add `keylog_file` to `TLSContext`, next to the keylog file's lock.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

lib/tlscontext.c
```diff
@@ -61,6 +64,10 @@ struct _TLSContext
   gchar *location;
 };
 
+#define _TLS_KEYLOG_INDEX 0
+#define _TLS_KEYLOG_FILE 1
+#define _TLS_KEYLOG_FILE_LOCK 2
```
MrAnno, line 69: It turned out that `SSL_get_ex_data()` is not really necessary, because we already use `SSL_get_app_data()`.
This way, we don't need to worry about OpenSSL versions either.

For example:
```C
TLSSession *self = SSL_get_app_data(ssl);
self->ctx->keylog_file
self->ctx->keylog_file_path
self->ctx->keylog_file_lock
```

lib/tlscontext.c
```diff
@@ -335,6 +342,57 @@ _set_sni_in_client_mode(TLSSession *self)
...
+      FILE *file = SSL_get_ex_data(ssl, _TLS_KEYLOG_FILE);
+      setbuf(file, NULL);
+      _write_line_to_keylog_file(file_path, line, file, file_lock);
+    }
+}
+
+static gboolean
+tls_session_keylog_setup_file(TLSSession *self, TLSContext *ctx)
+{
+  self->ctx->keylog_file = fopen(ctx->keylog_file_path, "a");
```
MrAnno, line 382: After IRL discussion, we agreed that opening/closing the file inside the callback may be more expedient, as it is not a frequent operation in syslog-ng. We could also call `fflush()`, before closing the file, so a real-time packet capture with Wireshark is more likely to work.

MrAnno (commented):
LGTM, I have 2 additional comments.

It seems the last commit was squashed incorrectly. Could you fix the history, please? :)

kira-syslogng:
Build SUCCESS

MrAnno:
Please add a news entry.

kira-syslogng:
Build SUCCESS

lib/tlscontext.c
```diff
@@ -45,6 +45,9 @@ struct _TLSContext
   TLSMode mode;
   gint verify_mode;
   gchar *key_file;
+  gchar *keylog_file_path;
```
Kokan, line 48: optional: these could be captured in a struct as they are connected values.

lib/tlscontext.c
```diff
@@ -335,6 +338,52 @@ _set_sni_in_client_mode(TLSSession *self)
...
+      if (ret_val != strlen(line))
+        msg_error("Couldn't write to TLS keylogfile", evt_tag_errno("error", ret_val));
+
+      fflush(keylog_file);
+    }
+  else
+    {
+      msg_error("Error opening keylog-file",
+                evt_tag_str(EVT_TAG_FILENAME, file_path),
+                evt_tag_error(EVT_TAG_OSERROR));
```
Kokan, line 358: This error message is misleading as there is no file opening here at all.
At this point the file is either opened or not, it would be possible to try open if we were not successful in the past.
I would vote not to try open the file here, and in that case there is no need to log anything or do anything at all.

Also I think the check for `keylog_file` can be done outside of the mutex, as it is only changed once when the tls context created.
```c
void
_write_line_to_keylog_file(onst char *file_path, const char *line, FILE *keylog_file, GMutex *mutex)
{
  if (!keylog_file)
      return;

  gint ret_val;
  g_mutex_lock(mutex);
  ...
}
```

lib/tlscontext.c
```diff
@@ -335,6 +338,52 @@ _set_sni_in_client_mode(TLSSession *self)
...
+    }
+  g_mutex_unlock(mutex);
+}
+
+static inline void
+_dump_tls_keylog(const SSL *ssl, const char *line)
+{
+  if(ssl != NULL)
```
Kokan, line 366: optional: use early returns

lib/tlscontext.c
```diff
@@ -335,6 +338,52 @@ _set_sni_in_client_mode(TLSSession *self)
...
 
+void
+_write_line_to_keylog_file(const char *file_path, const char *line, FILE *keylog_file, GMutex *mutex)
+{
+  gint ret_val;
+  g_mutex_lock(mutex);
+  if(keylog_file)
+    {
+      ret_val = fprintf(keylog_file, "%s\n", line);
+      if (ret_val != strlen(line))
```
Kokan, line 349: The `printf` above writes the `line` plus a newline to the file, as such the `ret_val` should be `strlen(line)+1`.

lib/tlscontext.c
```diff
@@ -335,6 +338,52 @@ _set_sni_in_client_mode(TLSSession *self)
...
+    {
+      TLSSession *self = SSL_get_app_data(ssl);
+      _write_line_to_keylog_file(self->ctx->keylog_file_path, line, self->ctx->keylog_file, &self->ctx->keylog_file_lock);
+    }
+}
+
+static gboolean
+tls_session_keylog_setup_file(TLSSession *self, const char *keylog_file_path)
+{
+  self->ctx->keylog_file = fopen(keylog_file_path, "a");
```
Kokan, line 376: Please use glib alternatives (`fopen` -> `g_fopen`).

lib/tlscontext.c
```diff
@@ -999,6 +1062,21 @@ tls_context_set_key_file(TLSContext *self, const gchar *key_file)
   SSL_CTX_set_default_passwd_cb_userdata(self->ssl_ctx, self->key_file);
 }
 
+gboolean
+tls_context_set_keylog_file(TLSContext *self, gchar *keylog_file_path)
+{
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L
+  g_free(self->keylog_file_path);
+  msg_warning_once("WARNING: TLS keylog file has been set up, it should only be used during debugging sessions, NOT in production environments",
```
Kokan, line 1070: Could we skip the `NOT in production environments` part ?
Reasoning, I agree this is not a feature that should be turned on all the time. But it must be turned on in production environment if we want to do a debug session there, and the message would yield otherwise.

kira-syslogng:
Build SUCCESS

MrAnno (approved):
👍🏻 
(Sorry for taking so long)

MrAnno:
I have validated that this works in a real-time Wireshark session using 2 drivers: a `network()` source, and a `syslog()` destination.

Kokan (approved):

PR #3793, opened 2021-09-27
actions: use GitHub docker registry in CI@devshell action

szemere:
WIP: Depends on https://github.com/syslog-ng/syslog-ng/pull/3782

Signed-off-by: Laszlo Szemere <laszlo.szemere@oneidentity.com>

PR #3798, opened 2021-09-29
[wip] github-actions: use cache-image in devshell workflow

alltilla:
Depends on #3782 and #3793


kira-syslogng:
Build FAILURE

.github/workflows/devshell.yml
```diff
@@ -1,16 +1,77 @@
...
+    env:
+      CACHED_IMAGE_NAME: ghcr.io/syslog-ng/dbld-devshell:latest
+
+    steps:
+      - name: Checkout syslog-ng source
+        uses: actions/checkout@v2
+
+      - name: Determine cached image ID
+        run: |
+          ./dbld/rules pull-image-devshell || true # remove the || true
```
gaborznagy, line 25: is this a left-in TODO?

gaborznagy (commented):
please rebase the PR.

.github/workflows/devshell.yml
```diff
@@ -1,16 +1,77 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Upload the temporary docker image
+        if: env.USE_CACHED_IMAGE == 'false'
+        run: |
+          docker push ${{ steps.determine-actual-image.outputs.docker-image }}
+
```
bazsi, line 67: Could we push this logic into dbld/rules itself?

I mean:
1) we already have a pull-image invoked from cache-image, and we have a separate step here. I'd love to see a single cache-image here and do anything to produce an image (potentially form cache) within dbld/rules
2) we seem to do a lot of back and forth deciding whether we are using the pulled or the rebuilt image, which is then used to decide if we need to push the temporary image. 
  - What if dbld/rules cache-image would be able to return information that we need here, without having to do a prefetch
  - also what if cache-image would be able to push to the registry, if a specific make variable is set (and if we did a rebuild)

I hope I make sense here, my point is if we push this entire logic to dbld/rules, and add a means to communicate information in a structured form, then we might be able to use the same cache-image-* target in all similar cases, avoiding the need to copy steps between github workflows files.

What do you think?

PR #3801, opened 2021-09-30
actions: transform container registry name to lowercase

MrAnno:
syslog-ng forks with uppercase letters in their names had problems building dbld images:

> invalid reference format: repository name must be lowercase

https://github.community/t/additional-function-s-lowercase-uppercase/140632

### DIFF
--- a/.github/workflows/dbld-images.yml
+++ b/.github/workflows/dbld-images.yml
@@ -22,12 +22,17 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
-    env:
-      CONTAINER_REGISTRY: "ghcr.io/${{ github.repository_owner }}"
-  
     steps:
       - name: Checkout source code
         uses: actions/checkout@v2
+
+      - name: Set container registry
+        run: |
+          CONTAINER_REGISTRY="ghcr.io/${{ github.repository_owner }}"
+          CONTAINER_REGISTRY="$(echo "$CONTAINER_REGISTRY" | tr '[:upper:]' '[:lower:]')"
+
+          . .github/workflows/gh-tools.sh
+          gh_export CONTAINER_REGISTRY
 
       - name: Build the images
         run: dbld/rules images -j $(nproc)


.github/workflows/dbld-images.yml
```diff
@@ -22,13 +22,18 @@ jobs:
...
-      CONTAINER_REGISTRY: "ghcr.io/${{ github.repository_owner }}"
-  
     steps:
       - name: Checkout source code
         uses: actions/checkout@v2
 
+      - name: Set container registry
+        run: |
+          CONTAINER_REGISTRY="ghcr.io/${{ github.repository_owner }}"
+          CONTAINER_REGISTRY="$(echo "$CONTAINER_REGISTRY" | awk '{print tolower($0)}')"
```
bazsi, line 32: ${CONTAINER_REGISTRY,,} is a bash expression that converts the value to lowercase. no need for external processes or awk.

```
bazsi@bzorp:~/src/syslog-ng/syslog-ng$ value=CamelCase
bazsi@bzorp:~/src/syslog-ng/syslog-ng$ echo ${value^^}
CAMELCASE
bazsi@bzorp:~/src/syslog-ng/syslog-ng$ echo ${value,,}
camelcase

```

bazsi: NOTE: this is bash related, so dash does not support it.

MrAnno: I was planning to use the bash >=4.0 syntax you mentioned, but I thought POSIX compatibility would be better for us.

I can switch to this syntax, if it is preferred :)

kira-syslogng:
Build SUCCESS

.github/workflows/dbld-images.yml
bazsi: no need. I can see now that even old bash does not support it. I hate awk code within shell scripts, but this is still ok. I would probably have used tr '[A-Z]' '[a-z]' once I recognized the bash version constraints.

but like I said, this is fine.

kira-syslogng:
Build SUCCESS

Kokan (approved):

OverOrion (approved):

PR #3814, opened 2021-10-14
Add rewrite rule to rename a macro

faxm0dem:
```
rewrite { rename(source('old_name'), destination('new_name')); };
```

### DIFF
--- a/scl/rewrite/rename.conf
+++ b/scl/rewrite/rename.conf
@@ -1,0 +1,41 @@
+@version: 3.19
+#############################################################################
+# Copyright (c) 2013 Balabit
+# Copyright (c) 2013 Márton Illés <marci@balabit.com>
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
@faxm0dem Thanks for opening this.

This PR is a result of an IRC conversation with @faxm0dem.
He used regexp filters with store-matches (the same applies to `regexp-parser()` as well), where named capture groups can not contain dots (due to a PCRE limitation), so when a simple one-depth prefix is not enough, values need to be renamed manually.
But in general, `rename()` might be useful in other use-cases.

IMHO `rename()` is a nice rewrite rule, maybe we should consider implementing it natively, but I personally like the SCL too. :)

MrAnno:
These are my thoughts on the native version:

- Performance (but I don't see that it could be significant)
- Debugging the native version may be a bit easier (preprocess-into, debug/trace messages, etc.)
- Another, but really really weak reason would be the fact that a native `rename()` could work without including `scl.conf`.

But again, I do like this SCL, so I think we could make it merge ready, if no one implements the native `rename()` until v3.35.
(We can replace the SCL with the native implementation whenever we want.)

faxm0dem:
ok, so I'll move forward and implement the corner cases using `if`s or `channels`

bazsi:
I just wanted to mention, that I have another use-case for renaming name-value pairs, which I would prefer to do natively, namely I would like to be able to "shift" the name-value pairs that store matches (e.g. $1, $2, etc), similarly how bash can do that.

With that we could have limited array support for processing JSON arrays, which I am experimenting with in #3768

MrAnno:
Wow, it would be really cool :)

bazsi:
I've just merged #3841 which solves this via native code, also handling all the corner cases correctly. Closing this in favour of that one. Thanks @faxm0dem for suggesting/working on this.

PR #3821, opened 2021-10-23
[WIP] issue-template: use YAML format

alltilla:
GitHub released a beta feature to make more configurable issue templates: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

What is useful to us:
  * Questions are not removable.
  * There are mandatory fields to be filled by the user.

The features needed by us seem to work correctly, so I think we can use this feature, while still in beta, but we don't need to hurry.

---

Other notable changes:
  * I did some minor rewording in the "Bug report" template.
  * I added links to our mailing list and Gitter.

---

You can see the PR in action here: https://github.com/alltilla/syslog-ng/issues/new/choose

---

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -236,9 +236,6 @@ TEST_EXTENSIONS = .sh
 
 VALGRIND_SUPPRESSIONS_FILES = $(top_srcdir)/tests/valgrind/unit-test-leak.supp
 EXTRA_DIST += syslog-ng.supp	\
-	.github/ISSUE_TEMPLATE/10-bug-report.md	\
-	.github/ISSUE_TEMPLATE/20-documentation-issue.md	\
-	.github/ISSUE_TEMPLATE/30-feature-request.md	\
 	GPL.txt	\
 	LGPL.txt	\
 	CMakeLists.txt


.github/ISSUE_TEMPLATE/10-bug-report.yml
```diff
@@ -0,0 +1,81 @@
...
+      required: true
+  - type: input
+    attributes:
+      label: Platform
+      description: Name and version of OS
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Debug bundle
```
gaborznagy, line 27: We rarely used this to resolve bug reports yet.

I would move this to the bottom, or just leave out. We can always ask for a debun bundle.

.github/ISSUE_TEMPLATE/10-bug-report.yml
```diff
@@ -0,0 +1,81 @@
...
+        $ syslog-ng-debun -r
+        ```
+  - type: markdown
+    attributes:
+      value: "# Issue"
+  - type: textarea
+    attributes:
+      label: Failure
+      description: |
+        Backtrace, error messages and detailed description of the failure.
```
gaborznagy, line 54: I would put "steps to reproduce" and "input and output logs" in this description too than having it in separate text fields.

I've only mentioned these to make the issue reporting page smaller.
It's just an idea.

PR #3824, opened 2021-10-30
dbld: install criterion from release tarball on debian-testing

alltilla:
libcriterion-dev was removed from debian-testing.
Until it gets fixed we can install criterion from their release tarball.

! Revert this, when libcriterion-dev gets fixed on debian-testing !

https://tracker.debian.org/pkg/criterion


kira-syslogng:
Build SUCCESS

Kokan (approved):

gaborznagy (approved):

PR #3825, opened 2021-11-02
Generate packages for ubuntu-hirsute and ubuntu-impish

MrAnno:


kira-syslogng:
Build SUCCESS

dbld/images/ubuntu-hirsute.dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY images/fake-sudo.sh /usr/bin/sudo
+COPY images/entrypoint.sh /
+COPY . /dbld/
+
+RUN /dbld/builddeps install_dbld_dependencies
+RUN /dbld/builddeps install_apt_packages
+RUN /dbld/builddeps install_debian_build_deps
+RUN /dbld/builddeps install_pip_packages
+
+RUN /dbld/builddeps install_criterion
```
alltilla, line 24: Hirsute and Impish have the `libcriterion-dev` package, I think we can/should use that.
https://launchpad.net/ubuntu/+source/criterion

alltilla:
Perhaps we should merge #3824 beforehand, so we can see the green CI run under this PR.

dbld/images/ubuntu-hirsute.dockerfile
MrAnno: Thanks.

MrAnno:
I just noticed that January 2022 is just 2 months from now. Should I drop Hirsute?

gaborznagy:
> I just noticed that January 2022 is just 2 months from now. Should I drop Hirsute?

I guess so, for the same reason we don't want to add Centos 8.

kira-syslogng:
Build SUCCESS

Kokan (approved):

PR #3836, opened 2021-11-10
Version: 3.35.1

kira-syslogng:
Tarball is available at: https://github.com/syslog-ng/syslog-ng/actions/runs/1443669436

TODO:
- [x] #3770
- [x] #3809


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

packaging/debian/changelog
```diff
@@ -1,3 +1,9 @@
+syslog-ng (3.35.1-1) unstable; urgency=medium
```
MrAnno, line 1: I think we should change the `distribution` to `stable` as we use that name in the APT repo as well.

bazsi:
Debian usually builds stuff in unstable and then those binaries become the
new stable. So it doesn't have to match. But I think it's better to keep it
this way so whenever the Debian stuff is synced, this won't be a difference.

MrAnno:
Thanks, @bazsi.

Kokan (approved):

kira-syslogng:
Build SUCCESS

MrAnno (approved):

PR #3841, opened 2021-11-16
Rewrite rename

Kokan:
This is a native implementation of https://github.com/syslog-ng/syslog-ng/pull/3814. The interface has been tweaked a little with using positional options.

```
rewrite { rename("old_name" "new_name"); };
```

In the following cases the `rename()` does not changes the message:
* If `source()` equals to `destination()`
* If `source()` entry not present in the message

### DIFF
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -510,6 +510,21 @@ _value_invalidates_legacy_header(NVHandle handle)
 }
 
 void
+log_msg_rename_value(LogMessage *self, NVHandle from, NVHandle to)
+{
+  if (from == to)
+    return;
+
+  gssize value_len = 0;
+  const gchar *value = log_msg_get_value_if_set(self, from, &value_len);
+  if (!value)
+    return;
+
+  log_msg_set_value(self, to, value, value_len);
+  log_msg_unset_value(self, from);
+}
+
+void
 log_msg_set_value_with_type(LogMessage *self, NVHandle handle, const gchar *value, gssize value_len, NVType type)
 {
   const gchar *name;
--- a/lib/rewrite/rewrite-rename.c
+++ b/lib/rewrite/rewrite-rename.c
@@ -33,22 +33,6 @@ struct _LogRewriteRename
   NVHandle destination_handle;
 };
 
-void
-log_rewrite_rename_set_source(LogRewrite *s, const gchar *source)
-{
-  LogRewriteRename *self = (LogRewriteRename *) s;
-
-  self->source_handle = log_msg_get_value_handle(source);
-}
-
-void
-log_rewrite_rename_set_destination(LogRewrite *s, const gchar *destination)
-{
-  LogRewriteRename *self = (LogRewriteRename *) s;
-
-  self->destination_handle = log_msg_get_value_handle(destination);
-}
-
 static void
 log_rewrite_rename_process(LogRewrite *s, LogMessage **pmsg, const LogPathOptions *path_options)
 {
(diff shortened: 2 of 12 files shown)


MrAnno:
+1 for the positional version :)

A general confusion already exists when using name-value pairs, due to the `value()` syntax, which we sometimes call `key()`, but the `template()` syntax is there too, which accepts macros.

I'm not sure which version would be more intuitive for our users, but for me, `rename("old_name" "new_name");` is the one.

kira-syslogng:
Build SUCCESS

lib/rewrite/rewrite-rename.c
```diff
@@ -0,0 +1,116 @@
...
+  log_msg_make_writable(pmsg, path_options);
+
+  gssize value_len = 0;
+  const gchar *value = log_msg_get_value_if_set(*pmsg, self->source_handle, &value_len);
+  if (!value)
+    return;
+
+  log_msg_set_value(*pmsg, self->destination_handle, value, value_len);
+  log_msg_unset_value(*pmsg, self->source_handle);
+}
```
bazsi, line 69: can we perhaps handle the special case where we only need to change the NVTable entry, instead of copying the data?

lib/rewrite/rewrite-rename.c
```diff
@@ -0,0 +1,116 @@
...
+
+  log_msg_make_writable(pmsg, path_options);
+
+  gssize value_len = 0;
+  const gchar *value = log_msg_get_value_if_set(*pmsg, self->source_handle, &value_len);
+  if (!value)
+    return;
+
+  log_msg_set_value(*pmsg, self->destination_handle, value, value_len);
+  log_msg_unset_value(*pmsg, self->source_handle);
```
bazsi, line 68: Would be great if this was a log_msg_rename_value() function, so we could use the same functionality in a different rewrite rule.

lib/rewrite/rewrite-rename.c
bazsi: btw, this might be more complex than warranted for this rename() use-case, so as long as we delegate this functionality to a lower level function where this might be added later, this is completely optional.

lib/rewrite/rewrite-rename.c
Kokan: I've added a rename function.

lib/rewrite/rewrite-rename.c
Kokan: I would go here with a simple rename function on log message level, and as you said later this can be improved. Likely with introducing shift.

lib/logmsg/logmsg.c
```diff
@@ -509,6 +509,21 @@ _value_invalidates_legacy_header(NVHandle handle)
   return handle == LM_V_PROGRAM || handle == LM_V_PID;
 }
 
+void
+log_msg_rename(LogMessage *self, NVHandle from, NVHandle to)
```
bazsi, line 513: log_msg_rename_value() perhaps?

Kokan: Sure.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

news/feature-3838.md
```diff
@@ -0,0 +1,8 @@
+add new rewrite: rename
+
+```
+rewrite {
+  rename( source("renamed-from") destination("renamed-to") );
+};
+```
```
bazsi, line 7: the arguments have become positional, so the example needs to be adjusted.

kira-syslogng:
Build SUCCESS

bazsi (approved):

bazsi (dismissed):

Kokan:
Resolved conflict.

kira-syslogng:
Build SUCCESS

MrAnno (approved):
Cool :)

bazsi (approved):

PR #3831, opened 2021-11-06
WIP: user visible parts of type support

bazsi:
This branch builds on top of PR #3768 (already merged) and #3850) and adds type support throughout syslog-ng in order to improve JSON support and enhance the power of syslog-ng template functions.

This PR might need to be split eventually, but I include all user-visible typing changes here:

This contains the following chunks:
 * support typing in value-pairs
 * support type propagation in template expressions
 * type hinting becomes a type-cast, internally supported in LogTemplate
 * support typing for ${value} references in template expressions
 * support typing for template functions, any template function can return type in addition to value
 * improve type support in json-parser() and in $(format-json).

Still todo:
 * typing support with macros
 * type support in Python bindings

JSON related notes:
With this branch, json-parser() can now fully consume a JSON payload and convert it into syslog-ng name-value pairs, the following way:
  * simple types (string, int32, int64, boolean, null) turn into syslog-ng name-value pairs
  * null is turned into a null() type
  * arrays are turned into syslog-ng lists, as long as elements are simple types. If the array contains other arrays or objects, the entire array is turned into a literal() value that contains the JSON representation of the array. This makes it possible to reproduce the JSON in $(format-json), lists are turned into simple arrays, literals are simply copied verbatim.
  * objects are recursed into, so the rules above apply with keys concatenated.

Changes:
  * null values are now supported
  * top-level arrays were previously not processed and resulted in an array, now we can turn them into top-level matches (e.g. $0 becomes the first element, $1 the 2nd and so on)

This is an incompatible change:
  * arrays were converted into a not-really-usable a series of name-value pairs with keys: key[0], key[1], key[2], this was NOT documented behavior and we were not able to reproduce the original JSON array based on these keys.

Here's the config I was testing with.

```
@version: 3.34

log {
	source { tcp(port(2000) flags(no-parse)); };
	parser { json-parser(prefix("stg.")); };
#	rewrite { set(int("$PID") value("PID")); };
	destination { file("/tmp/log" template("$(format-json stg.*)\n")); };

}
```

PR #3842, opened 2021-11-17
RFC: JSON avoid using trailing zeroes when representing double values

bazsi:
This is a spinoff from #3831 and currently breaks kira. Let's get back to this when #3831 is merged.


kira-syslogng:
Build FAILURE

bazsi:
I am closing this, as this is not important and we have too many open things.

PR #3843, opened 2021-11-19
cisco-parser: allow leading dot in timestamps

gaborznagy:
We've received an example log from the community, in which the timestamp had a leading dot.
I've slightly modified the hostname and the message parts
`<180>782431: machine1: .Nov 18 21:03:22.631 GMT: %CDP-4-NATIVE_VLAN_MISMATCH: Native VLAN mismatch discovered on TenGigabitEthernet.`

I haven't found an official documentation about the leading dot yet, only cisco community posts, e.g.:
https://community.cisco.com/t5/other-network-architecture/logging-shows-an-asterisk-quot-jan-13-11-05-40-quot/td-p/564707
https://community.cisco.com/t5/other-security-subjects/log-entries-with-a-dot-before-timestamp/td-p/590136

The leading asterisk is better documented in multiple system guides:
https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-1/system_monitoring/command/reference/sysmon_cr41crs_chapter4.html#wp255058350
> If the system clock has not been set, the date and time are preceded by an asterisk (*), which indicates that the date and time have not been set and should be verified.

Or here: https://www.cisco.com/c/en/us/td/docs/routers/access/wireless/software/guide/SysMsgLogging.html#wp1054710

### DIFF
--- a/news/bugfix-3843.md
+++ b/news/bugfix-3843.md
@@ -1,0 +1,1 @@
+cisco-parser(): allow a leading dot in the timestamp (not synced clocks)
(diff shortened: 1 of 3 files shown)


kira-syslogng:
Build SUCCESS

MrAnno (approved):
Thanks for the thorough explanation :)

bazsi (approved):

PR #3844, opened 2021-11-20
Stats split response (reworked version by Bazsi)

bazsi:
This is a continuation of #3707 originally by @lbudai which I have rebased against current master. Here's the original description of the PR:

## Motivation

When there are a lots of stats counters, getting these counters via `syslog-ng-ctl 
stats` is a challenge.
Before #3387 it was not even possible to stop (gracefully) syslog-ng when such a request was in progress.
Now, we can stop syslog-ng - but at the price of calling a `pthread_cancel` which can sometimes cause problems (-> and in some circumstances syslog-ng still cannot be stopped... when stats-mutex is locked when cancel is received... we could play with setting pthread cancel type to `PTHREAD_CANCEL_DEFERRED` and disable/enable cancel state - but that s*cks).

The aim of this PR is to solve two issues:
  * get rid of `pthread_cancel` by implementing 'cancellable' stats queries
  * decrease the required memory for serving a syslog-ng-ctl request (and send back partial results to the syslog-ng-ctl)

Before this patchset we collected all the responses into one giant GString object -> until all the counters are not visited we didn't send anything back to the `syslog-ng-ctl`.
  
*  notes
     * this work can be continued with improving `syslog-ng-ctl query`.
     * this PR is a kind of PoC - just wanted to share the problem and my code (as a kind of farewell PR :) )
 
 ## What's in the PR?

### batches

 * stats-csv interface changed: instead of returning a GString object that contains all the counters, I've added a callback which is called for all CSV record (->for all counter). 
 * functionality is reimplemented with the new interface (on a CSV record callback basis)
 * new methods are added for supporting batches : one is for adding a batch to the current 'session', other is for closing the batch (session) by adding a '.\n' to the end
 * syslog-ng-ctl also modified to handle response 'streams' (before this PR, response was appended to a GString object until '\n.\n' received, now it prints responses at reception time)
 * as `syslog-ng-ctl stats` handler is run in a separete thread, some synchronization is required:
    * batches are inserted into a thread safe GQueue
    * an iv_event is posted when a batch is sent from the producer thread (-> direct manipulation of ivykis objects from a different thread is not allowed )

### eliminate `pthread_cancel`

 * store cancelled state (could be an atomic variable, but I think it's okay to use a non-atomic variable) at ControlServer object
 * implement cancallable stats queries (-> g_hash_table_foreach replaced with an iterator-loop which can checks the `cancelled` state and break the loop when needed
 * thread id saving also removed (thread id needed by `thread_cancel`)
 
Some parts are implemented at night, so be careful with the review ;)
_Also note that I'm not sure if I'll have the time for fixing review notes - in this case you can close my PR, or if you find it useful, you can freely modify to make it merge-ready._


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

syslog-ng-ctl/control-client-unix.c
```diff
@@ -86,40 +86,43 @@ control_client_send_command(ControlClient *self, const gchar *cmd)
...
-
+      GString *reply = g_string_new_len(buff, len);
       if (reply->str[reply->len - 1] == '\n' &&
           reply->str[reply->len - 2] == '.' &&
           reply->str[reply->len - 3] == '\n')
         {
-          g_string_truncate(reply, reply->len - 3);
-          break;
+          g_string_truncate(reply, reply->len - 2);
+          received_all = TRUE;
```
Kokan, line 117: The detection of all data received relied on having `\n.\n` in the buffer, but `read` may not be so kind and gave it back in parts yielding missing this condition.

bazsi: @lbudai do you want to address this request or shall I?

lbudai: @bazsi :  of course, do all the required changes

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

syslog-ng-ctl/control-client-unix.c
bazsi: with the latest push, I've changed the client code to use stdio, thus the end-of-response detection has become trivial.

bazsi: @Kokan thanks for catching this, it would have been very easy to oversee this one.

bazsi:
I have one issue in my notes, the debug messages include the entire "command" string which could contain sensitive data, which should be masked. Otherwise architecturally this is in a good shape now.

kira-syslogng:
Build SUCCESS

bazsi:
Hmm, there are still some outstanding issues, that I don't yet understand as I didn't spend time to decipher them. @lbudai if you have hints, those would be useful:

 * there's a finished/cancel dance in ControlCommandThread and we sometimes do a join and sometimes we don't
 * the we have "cancelled" attribute for both ControlServer and ControlCommandThread wouldn't one of them is enough?
 * I guess we can lose "real_thread" struct within ControlCommandThread, right?

I think that's all.

lbudai:
> Hmm, there are still some outstanding issues, that I don't yet understand as I didn't spend time to decipher them. @lbudai if you have hints, those would be useful:
> 
>     * there's a finished/cancel dance in ControlCommandThread and we sometimes do a join and sometimes we don't
> 
>     * the we have "cancelled" attribute for both ControlServer and ControlCommandThread wouldn't one of them is enough?
> 
>     * I guess we can lose "real_thread" struct within ControlCommandThread, right?
> 
> 
> I think that's all.

@bazsi : will check soon

bazsi:
To answer my own questions:
* I have removed the redundancy in "cancelled"
* I have removed the "real_thread" wrapper

and I kept the finished/cancelled dance, which does the following:
* if the thread is still running at the time we try to cancel it (finished == FALSE), we do an explicit g_thread_join()
* if the thread has exited by that time (e.g. finished == TRUE), we don't do a g_thread_join(). The end of the thread might still be executing (e.g. the lines after unlocking state_lock that protects the finished variable), but I assume that can finish even without the mainloop running.

I have a parallel branch (#3879) where I introduced MainLoopThreadedWorker and I am starting to think that it is very similar to what we are achieving with ControlCommandThread in this branch. I don't want to stall either PRs now but a subsequent branch could merge these two into one class/implementation.

With that said, I think I've addressed all open concerns I had with the code originally introduced by @lbudai, so this should be ready to go in, pending review.

kira-syslogng:
Build SUCCESS

Kokan (approved):

MrAnno (approved):

PR #3845, opened 2021-11-22
qdisk: fix a race condition crash in serialization

Barkodcz:
This PR does 3 things:
  * Removes a layering violation in `qdisk.c` regarding message serialization.
  * Fixes the crash mentioned in the title of the PR.
  * Force disables compaction when saving the memory queues of non-reliable `disk-buffer` for safety reasons.

More details can be seen in the commit messages.

Signed-off-by: Balázs Barkó <Balazs.Barko@oneidentity.com>

### DIFF
--- a/modules/diskq/logqueue-disk.c
+++ b/modules/diskq/logqueue-disk.c
@@ -205,3 +205,32 @@ log_queue_disk_init_instance(LogQueueDisk *self, DiskQueueOptions *options, cons
 
   self->qdisk = qdisk_new(options, qdisk_file_id);
 }
+
+static gboolean
+_serialize_msg(SerializeArchive *sa, gpointer user_data)
+{
+  LogQueueDisk *self = ((gpointer *) user_data)[0];
+  LogMessage *msg = ((gpointer *) user_data)[1];
+
+  DiskQueueOptions *options = qdisk_get_options(self->qdisk);
+
+  return log_msg_serialize(msg, sa, options->compaction ? LMSF_COMPACTION : 0);
+}
+
+gboolean
+log_queue_disk_serialize_msg(LogQueueDisk *self, LogMessage *msg, GString *serialized)
+{
+  gpointer user_data[] = { self, msg };
+  GError *error = NULL;
+
+  if (!qdisk_serialize(serialized, _serialize_msg, user_data, &error))
+    {
+      msg_error("Error serializing message for the disk-queue file",
+                evt_tag_str("error", error->message),
+                evt_tag_str("persist-name", self->super.persist_name));
+      g_error_free(error);
+      return FALSE;
+    }
+
+  return TRUE;
+}
--- a/modules/diskq/logqueue-disk.c
+++ b/modules/diskq/logqueue-disk.c
@@ -108,7 +108,7 @@ _pop_disk(LogQueueDisk *self, LogMessage **msg)
       return FALSE;
     }
 
-  if (!qdisk_deserialize_msg(self->qdisk, read_serialized, msg))
+  if (!log_queue_disk_deserialize_msg(self, read_serialized, msg))
     {
       msg_error("Cannot read correct message from disk-queue file",
                 evt_tag_str("filename", qdisk_get_filename(self->qdisk)),
@@ -231,6 +231,36 @@ log_queue_disk_serialize_msg(LogQueueDisk *self, LogMessage *msg, GString *seria
       g_error_free(error);
       return FALSE;
     }
+
+  return TRUE;
+}
+
+static gboolean
+_deserialize_msg(SerializeArchive *sa, gpointer user_data)
+{
+  LogMessage *msg = user_data;
+
+  return log_msg_deserialize(msg, sa);
+}
+
+gboolean
+log_queue_disk_deserialize_msg(LogQueueDisk *self, GString *serialized, LogMessage **msg)
+{
+  LogMessage *local_msg = log_msg_new_empty();
+  gpointer user_data = local_msg;
+  GError *error = NULL;
+
+  if (!qdisk_deserialize(serialized, _deserialize_msg, user_data, &error))
+    {
+      msg_error("Error deserializing message from the disk-queue file",
+                evt_tag_str("error", error->message),
+                evt_tag_str("persist-name", self->super.persist_name));
+      log_msg_unref(local_msg);
+      g_error_free(error);
+      return FALSE;
+    }
+
+  *msg = local_msg;
 
   return TRUE;
 }
--- a/modules/diskq/logqueue-disk.c
+++ b/modules/diskq/logqueue-disk.c
@@ -203,6 +203,8 @@ log_queue_disk_init_instance(LogQueueDisk *self, DiskQueueOptions *options, cons
   log_queue_init_instance(&self->super, persist_name);
   self->super.type = log_queue_disk_type;
 
+  self->compaction = options->compaction;
+
   self->qdisk = qdisk_new(options, qdisk_file_id);
 }
 
@@ -212,9 +214,7 @@ _serialize_msg(SerializeArchive *sa, gpointer user_data)
   LogQueueDisk *self = ((gpointer *) user_data)[0];
   LogMessage *msg = ((gpointer *) user_data)[1];
 
-  DiskQueueOptions *options = qdisk_get_options(self->qdisk);
-
-  return log_msg_serialize(msg, sa, options->compaction ? LMSF_COMPACTION : 0);
+  return log_msg_serialize(msg, sa, self->compaction ? LMSF_COMPACTION : 0);
 }
 
 gboolean
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -84,6 +84,20 @@ struct _QDisk
   DiskQueueOptions *options;
 };
 
+#define QDISK_ERROR qdisk_error_quark()
+#define QDISK_ERROR_SERIALIZE 0
+
+GQuark
+qdisk_error_quark(void)
+{
+  static GQuark q = 0;
+
+  if (q == 0)
+    q = g_quark_from_static_string("qdisk-error");
+
+  return q;
+}
+
 static gboolean
 pwrite_strict(gint fd, const void *buf, size_t count, off_t offset)
 {
@@ -577,40 +591,32 @@ _overwrite_with_real_record_length(GString *serialized)
 }
 
 gboolean
-qdisk_serialize_msg(QDisk *self, LogMessage *msg, GString *serialized)
+qdisk_serialize(GString *serialized, QDiskSerializeFunc serialize_func, gpointer user_data, GError **error)
 {
-  gchar *error = NULL;
   SerializeArchive *sa = serialize_string_archive_new(serialized);
 
   /* Leave space for the real record_length for later */
   if (!serialize_write_uint32(sa, 0))
     {
-      error = "cannot write record length";
+      g_set_error(error, QDISK_ERROR, QDISK_ERROR_SERIALIZE, "failed to write record length");
       goto exit;
     }
 
-  if (!log_msg_serialize(msg, sa, self->options->compaction ? LMSF_COMPACTION : 0))
+  if (!serialize_func(sa, user_data))
     {
-      error = "cannot serialize LogMessage";
+      g_set_error(error, QDISK_ERROR, QDISK_ERROR_SERIALIZE, "failed to serialize data");
       goto exit;
     }
 
   if (!_overwrite_with_real_record_length(serialized))
     {
-      error = "message is empty";
+      g_set_error(error, QDISK_ERROR, QDISK_ERROR_SERIALIZE, "serializable data is empty");
       goto exit;
     }
 
 exit:
-  if (error)
-    {
-      msg_error("Error serializing message for the disk-queue file",
-                evt_tag_str("error", error),
-                evt_tag_str("filename", qdisk_get_filename(self)));
-    }
-
   serialize_archive_free(sa);
-  return error == NULL;
+  return *error == NULL;
 }
 
 gboolean
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -86,6 +86,7 @@ struct _QDisk
 
 #define QDISK_ERROR qdisk_error_quark()
 #define QDISK_ERROR_SERIALIZE 0
+#define QDISK_ERROR_DESERIALIZE 1
 
 GQuark
 qdisk_error_quark(void)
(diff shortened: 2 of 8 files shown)


kira-syslogng:
Build SUCCESS

modules/diskq/qdisk.c
```diff
@@ -881,8 +882,14 @@ _save_queue(QDisk *self, GQueue *q, QDiskQueuePosition *q_pos)
   return success;
 }
 
+/*
+  TODO: this function should be in the LogqueueDisk class
+        because _save_queue function is serializing
+        and qdisk class should not know "how to serialize"
```
MrAnno, line 866: What do you think about something like this?

> TODO: The non-reliable memory queue saving logic within this method does not belong to QDisk.
> That should be moved into LogQueueDiskNonReliable, that way LogQueueDiskSerializeMsgFunc could be removed too.

modules/diskq/qdisk.c
```diff
@@ -955,8 +962,13 @@ _upgrade_header(QDisk *self)
   self->hdr->version = QDISK_HDR_VERSION_CURRENT;
 }
 
+/*
+  TODO: this function should be in LogqueueDisk class (at least part of it)
+        because `qdisk_save_state` function is using serialization
+*/
```
MrAnno, line 946: > TODO: remove LogQueueDiskSerializeMsgFunc after fixing the layering violation in qdisk_save_state().

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -76,12 +76,12 @@ _save_diskqueue(LogQueue *q)
...
 
 static gsize
-_calculate_serialized_empty_message_size(QDisk *qdisk)
+_calculate_serialized_empty_message_size(LogQueueDisk *queue_disk)
 {
   LogMessage *msg = log_msg_new_empty();
   GString *serialized_msg = g_string_new(NULL);
 
-  cr_assert(qdisk_serialize_msg(qdisk, msg, serialized_msg));
+  cr_assert(log_queue_disk_serialize_msg(queue_disk, msg, serialized_msg));
```
MrAnno, line 84: This commit probably belongs to the first one, where `qdisk_serialize_msg()` is removed.
(We prefer creating commits that at least can be compiled.)

modules/diskq/logqueue-disk.c
```diff
@@ -289,3 +289,22 @@ log_queue_disk_serialize_msg(LogQueueDisk *self, LogMessage *msg, GString *seria
...
+
+gboolean
+log_queue_disk_deserialize_msg(LogQueueDisk *self, GString *serialized, LogMessage **msg)
+{
+  SerializeArchive *sa = serialize_string_archive_new(serialized);
+  LogMessage *local_msg = log_msg_new_empty();
+
+  if (!log_msg_deserialize(local_msg, sa))
+    {
+      log_msg_unref(local_msg);
```
MrAnno, line 301: The error message `Error deserializing message from the disk-queue file` is missing from here.

MrAnno (commented):
Nice fix (and refactor)! :)

alltilla:
Maybe we can push this refactor a bit further, making even cleaner code:

**1:**
IMO the business logic of framing a `GString` should be in `QDisk`.

I would imagine the `request_serializer()` and `_overwrite_with_real_record_length()` (with a bit more generic name) in the `QDisk` code. The user of these functions would be `LogQueueDisk`, and it would call `log_msg_serialize()` between the two functions.

**2:**
IMO the `compaction` option has no business being in `QDisk`, heck, even `DiskQueueOptions` should not be in `QDisk`.

I checked the grammar and there we prepare a `DiskQueueOptions` instance, which gets passed to `log_queue_disk_init_instance()`, which passes it to `QDisk`. `QDisk` should not know about `DiskQueue`/`LogQueueDisk` related options. We have layering violations unfortunately, as we extensively use the `reliable` field in `QDisk`, and we use the `compaction` field there, too.

So IMO there should be a separation in `log_queue_disk_init_instance()`: it receives `DiskQueueOptions`, and then it should store anything from it which is necessary for `LogQueueDisk` (`reliable` (not in this PR) and `compaction`), then it should create something like `QDiskOptions`, which does not have those values, and should pass it to `QDisk`.

@MrAnno, @Barkodcz What do you think?

modules/diskq/logqueue-disk.c
```diff
@@ -203,4 +225,86 @@ log_queue_disk_init_instance(LogQueueDisk *self, DiskQueueOptions *options, cons
...
+
+  if ((error = _overwrite_with_real_record_length(serialized)) != NULL)
+    goto exit;
+
+exit:
+  if (error)
+    {
+      msg_error("Error serializing message for the disk-queue file",
+                evt_tag_str("error", error),
+                evt_tag_str("filename", log_queue_disk_get_filename(&self->super)));
```
Kokan, line 286: The `log_queue_disk_get_filename` still access qdisk without locking (not asking for lock here), that yields the same segfault as this PR aims to solve.

modules/diskq/logqueue-disk.c
```diff
@@ -194,6 +194,28 @@ log_queue_disk_restart_corrupted(LogQueueDisk *self)
...
+
+static void
+_set_serializer(LogQueueDisk *self)
+{
+  DiskQueueOptions *options = qdisk_get_options(self->qdisk);
+
+  if (options->compaction)
+    self->serialize_msg = _serialize_msg_with_compaction;
+  else
+    self->serialize_msg = _serialize_msg_without_compaction;
```
Kokan, line 217: Currently there is only the compaction flag exists here, which makes this a simple condition and two function, but possible adding more flags would make it less appealing to branch out due to these flags. (In addition this replaces a simple int argument passing with pointer dereference.)

Barkodcz:
> Maybe we can push this refactor a bit further, making even cleaner code:
> 
> **1:** IMO the business logic of framing a `GString` should be in `QDisk`.
> 
> I would imagine the `request_serializer()` and `_overwrite_with_real_record_length()` (with a bit more generic name) in the `QDisk` code. The user of these functions would be `LogQueueDisk`, and it would call `log_msg_serialize()` between the two functions.
> 
> **2:** IMO the `compaction` option has no business being in `QDisk`, heck, even `DiskQueueOptions` should not be in `QDisk`.
> 
> I checked the grammar and there we prepare a `DiskQueueOptions` instance, which gets passed to `log_queue_disk_init_instance()`, which passes it to `QDisk`. `QDisk` should not know about `DiskQueue`/`LogQueueDisk` related options. We have layering violations unfortunately, as we extensively use the `reliable` field in `QDisk`, and we use the `compaction` field there, too.
> 
> So IMO there should be a separation in `log_queue_disk_init_instance()`: it receives `DiskQueueOptions`, and then it should store anything from it which is necessary for `LogQueueDisk` (`reliable` (not in this PR) and `compaction`), then it should create something like `QDiskOptions`, which does not have those values, and should pass it to `QDisk`.
> 
> @MrAnno, @Barkodcz What do you think?

1. Yes, I will do this!
2. Yes, it would be great that refactor, but I would do that in a separate PR. This PR is just a bug fix, not a refactor PR (I know it has some refactoring, but it was necessary for the bugfix)
If you are okay with that I will do it (or somebody else) in a separate PR

modules/diskq/logqueue-disk.c
Barkodcz: The serialize_msg function needs a LogMessaget and a SerializeArchive parameter (and nothing else) for the save_queue function to get the serialize function. If the save_queue function does not need the serialize function, then can get the flag as a parameter.

So in the future can be an int parameter, but right now can not

kira-syslogng:
Build SUCCESS

MrAnno:
@alltilla Both ideas sound good to me. :)

modules/diskq/logqueue-disk.c
MrAnno: I was also curious why we replaced a simple condition with a pointer dereference. I thought it was a virtual method for `LogQueueDisk` that can be overridden, but it turned out that we have another layering problem in `QDisk` (`save_queue()`), which we could fix later. This function pointer temporarily keeps `save_queue()` in a functioning state without requiring a `LogQueueDisk` or a `DiskQueueOptions` instance with multiple LogQueueDisk calls inside QDisk.

MrAnno:
Please add a news entry.

kira-syslogng:
Build SUCCESS

Kokan (approved):

modules/diskq/qdisk.h
```diff
@@ -36,6 +36,7 @@
  * which has side effects.
  */
 #define POINTER_TO_LOG_PATH_OPTIONS(ptr, lpo) (lpo)->ack_needed = (GPOINTER_TO_INT(ptr) & ~0x80000000)
+typedef gboolean (*LogQueueDiskSerializeMsgFunc)(LogMessage *msg, SerializeArchive *sa);
```
MrAnno, line 39: This belongs to `LogQueueDisk`.

modules/diskq/qdisk.c
```diff
@@ -1355,10 +1297,65 @@ qdisk_free(QDisk *self)
...
+  return TRUE;
+}
+
 QDisk *
 qdisk_new(DiskQueueOptions *options, const gchar *file_id)
 {
   QDisk *self = g_new0(QDisk, 1);
   qdisk_init_instance(self, options, file_id);
+  self->start_serialize = _start_serialize;
+  self->end_serialize = _end_serialize;
```
MrAnno, line 1359: There is no reason for these 2 to be virtual. They are set here, they are not meant to be overridden.

modules/diskq/qdisk.c
```diff
@@ -1355,10 +1297,65 @@ qdisk_free(QDisk *self)
   g_free(self);
 }
 
+gboolean
+qdisk_start_serialize(QDisk *self, gpointer *args)
```
MrAnno, line 1301: Why do we use gpointers args here instead of the actual parameters?

modules/diskq/qdisk.c
```diff
@@ -1355,10 +1297,65 @@ qdisk_free(QDisk *self)
...
+{
+  if (self->start_serialize)
+    return self->start_serialize(args);
+
+  return FALSE;
+}
+
+gboolean
+qdisk_end_serialize(QDisk *self, gpointer *args)
```
MrAnno, line 1310: The same question applies.

modules/diskq/qdisk.c
Barkodcz: In the future, if we want to create different methods of saving data to file (qdisk), then we have the options to create. And that's why the args argumentum.

But if we agree to we want only this method, then I will remove the virtual function (and the args parameter)

MrAnno: It's `QDisk` responsibility to define how to store a binary record. Even if there was a different method,  that would be a private implementation detail (but I don't think another method could be expected in the foreseeable future).

modules/diskq/logqueue-disk.c
```diff
@@ -203,4 +225,61 @@ log_queue_disk_init_instance(LogQueueDisk *self, DiskQueueOptions *options, cons
...
+{
+  gpointer args[] = {sa, error};
+  return qdisk_start_serialize(self->qdisk, args);
+}
+
+static gboolean
+_serialize_msg(LogQueueDisk *self, LogMessage *msg, SerializeArchive **sa)
+{
+  if (self->serialize_msg)
+    return self->serialize_msg(msg, *sa);
```
bazsi, line 242: if we passed self to serialize_msg at this point (as we usually do for virtual methods), the _set_serializer() would become moot and unnecessary, making the flow a bit easier to follow, IMHO.

That way the implementation of serialize_msg could either use the compaction flag or not depending on options->compaction. Am I missing something?

modules/diskq/logqueue-disk.c
```diff
@@ -203,4 +225,61 @@ log_queue_disk_init_instance(LogQueueDisk *self, DiskQueueOptions *options, cons
...
+}
+
+gboolean
+log_queue_disk_serialize_msg(LogQueueDisk *self, LogMessage *msg, GString *serialized)
+{
+  GError *error = NULL;
+  SerializeArchive *sa = serialize_string_archive_new(serialized);
+
+  /* Leave space for the real record_length for later */
+  if (!_request_serializer(self, sa, &error))
```
bazsi, line 261: What does "_request" mean in this context? Can you suggest a better name? By reading it I don't really have a clue what I should expect there.

bazsi: looking at the actual code, I can see that this is what emits a record length, e.g. it implements the framing used within the on-disk format. With that in mind this should be something like _prepare_frame. Also, I don't really see whether we need to delegate this back to the qdisk layer. If we do, it could simple prepend the length prior to calling this function that way we could straighten the execution flow. For me this is a bit ziggy-zaggy now.

I would check two alternatives:
1) define which layer has responsibility over what: qdisk is about the disk file format, LogQueueDisk is about the logical representation + the LogQueue glue. Where do we want to delegate the fact that the content in the file is framed. qdisk seems a logical choice, but I don't know whether LogQueueDisk has already been infested by this knowledge or not.

2) if the answer is that we want to delegate framing to qdisk, I would probably implement the framing even before calling into this function. The sole responsibility of this function is to produce a string representation of a LogMessage and qdisk should take it from there. _start_serialize and _end_serialize should not be managed from right within this function.

That's my $.02

modules/diskq/qdisk.c
bazsi: If we remove the responsibility of managing this in log_queue_disk_serialize() function and just do this where this was called, we would still have the options for changing the on-disk format if we ever have to. But I don't anticipate that at this point either.

bazsi (dismissed):
I didn't go through the entire patchset but I've already made a couple of notes on the commits here and there, which should be addressed before this goes in.

alltilla:
@kira-syslogng Do stresstest please

alltilla:
@bazsi @MrAnno @Kokan 

Me and @Barkodcz have agreed that I will finish this PR.

I had a bit different idea of fixing the issue, so some of your review comments became obsolete.
If you have the time, kindly re-review this PR.

Thanks!

kira-syslogng:
Build SUCCESS

kira-syslogng:

MrAnno:
> If you have the time, kindly re-review this PR.

I'll re-review it by the end of Thursday.

Kokan (approved):

MrAnno (approved):
LGTM.

MrAnno:
We are waiting for an internal test.

MrAnno:
@bazsi Attila made some changes, he fixed this issue from a slightly different angle than the original proposal. The layering violation has been resolved elegantly in my opinion.

We'd like to add this fix to one of our PE releases. Do you mind, if we remove your "change request" so we can integrate this into master?

modules/diskq/logqueue-disk.c
MrAnno: `self->serialize_msg` has been removed from this PR.

modules/diskq/logqueue-disk.c
MrAnno: I believe the ziggy-zaggy way of executing the serialization has been fixed.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

modules/diskq/logqueue-disk.c
MrAnno: I'm currently writing proper unit tests for `QDisk`, and it can be seen clearly there that QDisk should be responsible for its framing format. Unfortunately,  its `push_tail()` method allows unframed data, but `pop_head()` expects the framed format, which is a problem.

I will clarify these problems in a separate PR.

PR #3850, opened 2021-11-24
support typing in the LogMessage layer (not-user-visible)

bazsi:
This branch lays the foundations for using types within syslog-ng and is a continuation of #3768. This branch has been split off from #3831 as that has become pretty sizable. Also #3857 was extracted partially from this branch, so that needs to go in first. 

In summary:

  * #3768 is already merged
  * #3857 the elimination of testutils, is next up,
  * #3850 is this branch, can go in once testutils is merged
  * #3831 comes after, probably further splits expected. That's where I work now.

An increased unit test coverage for the already merged #3768 is here.

This branch only has a limited user-visibility bits, most of it is just groundwork for the rest the followup PR contains the user-visible changes. It was split off in the hope of making it easier to review/merge.

Here's the list of changes that are covered in this branch in separate sequences of commits:
  * complete the type coverage at the LogMessage layer (general type enum, addition of new functions to allow set/get type information), conversion of db-parser() types to the new typing structure, integration of type-hinting into the new typing framework
  * improved test_logmsg_serialize unit test for more coverage and some cleanups
  * db-parser(): convert to the new typing framework so that newly parsed messages would use the new type values
  * set(): make set type aware, so the type of the expression is propagated, this only works with an explicit type hint as templates don't support types yet

next steps:
  * propagate NVEntry->type in value-pairs() expressions, that can be overridden by type-hints (coming in #3831)
  * support types in json-parser() (coming in #3831)
  * support types in macros (no PR yet, only some local code)
  * support types in simple template expressions, so that trivial template expressions yield correct type information
  * support types in the Python binding

A few features that have become possible (but are not part of this PR):
 * incoming JSON payloads can be reproduced fully (my understanding that this is now complete, so any JSON comes in can be reproduced with $(format-json), thus _any_ change within any JSON payload becomes possible from within syslog-ng using name-value pairs.
 * JSON arrays can now be represented, manipulated and reproduced at the output
 * simple JSON types (such as int64, double, boolean, etc) are reproduced at the output by default, without the need to use type-hints
 * type aware destinations (mongodb for instance) automatically benefit from the new types

Future features that we could implement:
  * type aware comparisons (e.g. the famous ne vs != operator problem)
  * further type aware consumers (mongodb for instance)
  * type aware python bindings


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:

kira-syslogng:
Build SUCCESS

bazsi:
As discussed IRL, I have removed the value-pairs change from this PR, namely the one that exposes types to value-pairs based functionality, e.g. this should be completely invisible to users and safe to merge without compatibility concerns. The value pairs patch is moved to the subsequent JSON PR (e.g. #3831)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:
I've extracted the testutils removal and some test improvements into #3857, which needs to go in first, then this one.

lib/logmsg/logmsg-serialize-fixup.c
```diff
@@ -28,6 +28,56 @@
 
 #include <stdlib.h>
 
+enum
```
Kokan, line 31: I would really like to have a comment or a commit message that explains this a little more.

My understanding that the `LEGACY_DBPARSER_TYPE_` captures the current state of the `RPT_`. Meaning if `RPT_IP` is represented as 8, the `LEGACY_DBPARSER_TYPE_IP` should as well. The plan is to not change the `LEGACY_DBPARSER_TYPE_` ever, while the `RPT_` can be changed (order/extend). Is that correct, or I completely misunderstood this code?

bazsi: We capture these values once and db-parser() would never produce them again, at least as type values. 

If you look at the db-parser() changes in the same PR, it will start to emit a  LM_VT_XXX instead of RPT_*. This way RPT_* would simply capture the index of the parser, not even sure if that's being used anywhere.

.... looked at it, parser_type is only ever used in debug messages. So once this is merged the coupling between types and the RPT_* enums is discontinued.

bazsi: I have added a comment to clarify this.

bazsi:
Albeit the title says this is not user visible, there's one minor feature that is user visible but is not an incompatible change, set() now supports a typed expression, so if one uses a type-hint that type-hint would be used:

set(int64($PID)) works.

If you want I can  pull that last patch into the followup PR, I left it there so that the feature can be tested end-to-end. This might be considered however depending on how we want to introduce the incompatible changes, as for compatibility I'd rather choose to ignore the stored type field and force it to LM_VT_STRING. If that will happen,  this set() use-case will stop working.

Let me know which one you'd prefer.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

MrAnno:
Sorry for the extreme delay. I will finish the review of this PR over the weekend.

kira-syslogng:
Build SUCCESS

lib/logmsg/logmsg.h
```diff
@@ -142,6 +142,34 @@ enum
   __UNUSED_LF_LEGACY_MSGHDR    = 0x00020000,
 };
 
+typedef NVType LogMessageValueType;
```
MrAnno, line 145: NVType is also typedefed in nvtable.h to a different type: guint8.
Typedef redefinitions produce compiler errors even without the `-pedantic` flag in older gcc versions, so I think the redef should be eliminated.

MrAnno (commented):
LGTM. (I have not validated the old syslog-ng serialized binary arrays.)

We should do some follow-ups internally, and run a regression test there too before merging.

Notes:
- If I'm not mistaken, templates (even trivial ones) "lose" their type annotations.
- `log_msg_set_value_indirect_with_type()` now exists, but it isn't used yet.

MrAnno:
> Let me know which one you'd prefer.

I like the `set()` type-hint implementation in this PR. 👍🏻

lib/logmsg/logmsg.h
bazsi: It is not NVType that gets defined here, it is LogMessageValueType that becomes an alias for NVType.

I choose this, so that nvtable does not need to care about
the specifics and semantic meaning of the type field. 
LogMessage should be responsible for translation.

MrAnno: Sorry again. I don't know what I thought originally when I wrote this...

MrAnno (approved):

Kokan (approved):

PR #3851, opened 2021-11-26
reset queued stats if disk queue corrupted and restarted

Kokan:
The disk buffer is re-created when it detects it became corrupt, but the queued message statistics is not set accordingly.

### DIFF
--- a/modules/diskq/logqueue-disk-non-reliable.c
+++ b/modules/diskq/logqueue-disk-non-reliable.c
@@ -169,7 +169,7 @@ _move_messages_from_overflow(LogQueueDiskNonReliable *self)
     }
 }
 
-static void
+static gboolean
 _move_messages_from_disk_to_qout(LogQueueDiskNonReliable *self)
 {
   do
@@ -181,26 +181,32 @@ _move_messages_from_disk_to_qout(LogQueueDiskNonReliable *self)
       LogMessage *msg = log_queue_disk_read_message(&self->super, &path_options);
 
       if (!msg)
-        break;
+        return FALSE;
 
       g_queue_push_tail(self->qout, msg);
       g_queue_push_tail(self->qout, LOG_PATH_OPTIONS_TO_POINTER(&path_options));
       log_queue_memory_usage_add(&self->super.super, log_msg_get_size(msg));
     }
   while (HAS_SPACE_IN_QUEUE(self->qout));
+
+  return TRUE;
 }
 
-static inline void
+static inline gboolean
 _maybe_move_messages_among_queue_segments(LogQueueDiskNonReliable *self)
 {
+  gboolean ret = TRUE;
+
   if (qdisk_is_read_only(self->super.qdisk))
-    return;
+    return TRUE;
 
   if (self->qout->length == 0 && self->qout_size > 0)
-    _move_messages_from_disk_to_qout(self);
+    ret = _move_messages_from_disk_to_qout(self);
 
   if (self->qoverflow->length > 0)
     _move_messages_from_overflow(self);
+
+  return ret;
 }
 
 /* runs only in the output thread */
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build FAILURE

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -312,14 +319,18 @@ _pop_head(LogQueue *s, LogPathOptions *path_options)
     }
 
 success:
-  _maybe_move_messages_among_queue_segments(self);
+  if (!_maybe_move_messages_among_queue_segments(self))
+    {
+      stats_update = FALSE;
```
MrAnno, line 324: I don't understand this commit. Could you please explain it?

My understanding is that `_pop_head()` should decrement the queued messages counter whenever a message is returned from this function.

AFAIK `_maybe_move_messages_among_queue_segments()` is a transparent side effect, and it should not change the way we calculate queued messages.

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -181,26 +181,32 @@ _move_messages_from_disk_to_qout(LogQueueDiskNonReliable *self)
       LogMessage *msg = log_queue_disk_read_message(&self->super, &path_options);
 
       if (!msg)
-        break;
+        return FALSE;
```
MrAnno, line 184: This will eventually be FALSE when we can't fully saturate `qout`.

MrAnno: Found it, sorry:
```
if (qdisk_get_length(self->super.qdisk) <= 0)
```

modules/diskq/logqueue-disk-non-reliable.c
MrAnno: I get it now: the move operation can really fail, hence the conditional stats update.

MrAnno (dismissed):
👍🏻

Kokan:
@kira-syslogng please retest this

Kokan:

kira-syslogng:
Build SUCCESS

mitzkia:

kira-syslogng:
Build SUCCESS

MrAnno:
Oh, I forgot about the NEWS file entry. Could you please add one?

Kokan:
Sure, you are on fire with news entries ;)

kira-syslogng:
Build FAILURE

MrAnno:
Sorry, I had no intention of being intrusive 😄

MrAnno (approved):

gaborznagy:
I've checked if `memory_usage` has to be reset as well, but as I see, it will be adjusted during `restart` by `LogQueueDisk::start`.
Can you confirm, if I'm right, please?

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -312,14 +319,18 @@ _pop_head(LogQueue *s, LogPathOptions *path_options)
...
+    }
 
   g_mutex_unlock(&s->lock);
 
   if (s->use_backlog)
     _push_tail_qbacklog(self, msg, path_options);
 
-  log_queue_queued_messages_dec(s);
+  if (stats_update)
+    log_queue_queued_messages_dec(s);
```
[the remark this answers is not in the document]
gaborznagy: Of course, then this would be done under locking, which wanted to avoid.
it's just an idea.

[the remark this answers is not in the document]
Kokan: I did consider this variation, this reduces the stress in the code here. I wanted to avoid that, due to it is easy to overlook the fact in `_maybe_move_messages_among_queue_segments` fatal stuff can happen. In general I would incline to do as you suggested, but here I think keeping the tension is helpful fur future debugging/review as this code raises the attention of the reader.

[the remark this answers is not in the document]
gaborznagy: Thanks for the explanation. Good point about visibility, it did got my attention the first time.
I see your point, let's keep it this way then.

gaborznagy:

gaborznagy (approved):
Waiting for a green Kira before merge.

kira-syslogng:
Build SUCCESS

PR #3860, opened 2021-12-16
Github Actions: continue to build dbld images even if one platform fails

gaborznagy:
Sometimes an image build fails, but most of the time due to some infrastructure failure.

We discussed that we should build the other images and also upload them regularly, not just in case of special events (e.g. dbld folder's content is changed).

I've tested image build failure (in a dummy way though), see some workflow runs on my fork:
- [with synchronized output](https://github.com/gaborznagy/syslog-ng/actions/runs/1587504844)
- [with image upload](https://github.com/gaborznagy/syslog-ng/runs/4498754819) (by explicitly enabling image upload)
- [ordinary run](https://github.com/gaborznagy/syslog-ng/actions/runs/1587371301) (no image upload due to running on fork)

### DIFF
--- a/.github/workflows/dbld-images.yml
+++ b/.github/workflows/dbld-images.yml
@@ -35,9 +35,10 @@ jobs:
           gh_export CONTAINER_REGISTRY
 
       - name: Build the images
-        run: dbld/rules images -j $(nproc)
+        run: dbld/rules images -j $(nproc) --keep-going
 
       - name: Should we upload the images?
+        if: always()
         run: |
           if [ "${{ github.event.inputs.upload_images }}" = "true" ] || \
              ( \
@@ -55,7 +56,7 @@ jobs:
           gh_export UPLOAD_IMAGES_INTERNAL
 
       - name: Log in to the Container registry
-        if: env.UPLOAD_IMAGES_INTERNAL == 'true'
+        if: always() && env.UPLOAD_IMAGES_INTERNAL == 'true'
         uses: docker/login-action@v1
         with:
           registry: ghcr.io
@@ -63,7 +64,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Push the images
-        if: env.UPLOAD_IMAGES_INTERNAL == 'true'
+        if: always() && env.UPLOAD_IMAGES_INTERNAL == 'true'
         run: |
           IMAGES=$(docker image ls --filter "reference=${{ env.CONTAINER_REGISTRY }}/dbld-*:latest" --format '{{.Repository}}:{{.Tag}}')
           for IMAGE in $IMAGES; do
--- a/.github/workflows/dbld-images.yml
+++ b/.github/workflows/dbld-images.yml
@@ -44,7 +44,7 @@ jobs:
              ( \
               [ "${{ github.repository_owner }}" = "syslog-ng" ] && \
               [ "${{ github.ref }}" = "refs/heads/master" ] && \
-              [ "${{ github.event_name }}" = "push" ] \
+              [[ "${{ github.event_name }}" =~ ^(push|workflow_dispatch|schedule)$ ]] \
             )
           then
             UPLOAD_IMAGES_INTERNAL="true"
--- a/.github/workflows/dbld-images.yml
+++ b/.github/workflows/dbld-images.yml
@@ -60,8 +60,4 @@ jobs:
       - name: Push the images
         if: always() && env.UPLOAD_IMAGES_INTERNAL == 'true'
         run: |
-          IMAGES=$(docker image ls --filter "reference=${{ env.CONTAINER_REGISTRY }}/dbld-*:latest" --format '{{.Repository}}:{{.Tag}}')
-          for IMAGE in $IMAGES; do
-            echo "Pushing image: $IMAGE"
-            docker push $IMAGE
-          done
+          dbld/rules push-images -j $(nproc) --keep-going --output-sync
--- a/.github/workflows/dbld-images.yml
+++ b/.github/workflows/dbld-images.yml
@@ -13,7 +13,7 @@ on:
     - cron: '00 03 * * *'
   workflow_dispatch:
     inputs:
-      upload_images:
+      testing_image_upload:
         description: Should we upload the images into GitHub Packages? (true/false)
         required: false
         default: "false"
@@ -40,7 +40,7 @@ jobs:
       - name: Should we upload the images?
         if: always()
         run: |
-          if [ "${{ github.event.inputs.upload_images }}" = "true" ] || \
+          if [ "${{ github.event.inputs.testing_image_upload }}" = "true" ] || \
              ( \
               [ "${{ github.repository_owner }}" = "syslog-ng" ] && \
               [ "${{ github.ref }}" = "refs/heads/master" ] && \
--- a/.github/workflows/dbld-images.yml
+++ b/.github/workflows/dbld-images.yml
@@ -29,7 +29,7 @@ jobs:
           gh_export CONTAINER_REGISTRY
 
       - name: Build the images
-        run: dbld/rules images -j $(nproc) --keep-going
+        run: dbld/rules images -j $(nproc) --keep-going --output-sync
 
       - name: Should we upload the images?
         if: always()
--- a/.github/workflows/dbld-images.yml
+++ b/.github/workflows/dbld-images.yml
@@ -3,12 +3,6 @@ name: Compile dbld-images
 on:
   pull_request:
   push:
-    paths:
-      - .github/workflows/dbld-images.yml
-      # List is from dbld/rules cache-image-% "WATCHED_FILES"
-      - "dbld/**"
-      - "packaging/rhel/syslog-ng.spec"
-      - "packaging/debian/control"
   schedule:
     - cron: '00 03 * * *'
   workflow_dispatch:
--- a/dbld/rules
+++ b/dbld/rules
@@ -217,6 +217,11 @@ image-%:
 	$(DBLD_DIR)/prepare-image-build $* && \
         $(DOCKER) build $(DOCKER_BUILD_ARGS) --build-arg=ARG_IMAGE_PLATFORM=$* --build-arg=COMMIT=$$($(GIT) rev-parse --short HEAD || echo "") --build-arg=CONTAINER_REGISTRY=${CONTAINER_REGISTRY} --network=host -t ${CONTAINER_REGISTRY}/dbld-$* -f $(DBLD_DIR)/images/$*.dockerfile $(DBLD_DIR)
 
+push-images: $(foreach image,$(IMAGES), push-image-$(image))
+push-image-%:
+	@echo "Pushing image: $(image)"
+	$(DOCKER) push ${CONTAINER_REGISTRY}/dbld-$*
+
 pull-images: $(foreach image,$(BUILDER_IMAGES), pull-image-$(image))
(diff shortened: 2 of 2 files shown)


kira-syslogng:
Build SUCCESS

.github/workflows/dbld-images.yml
```diff
@@ -35,15 +29,15 @@ jobs:
...
       - name: Should we upload the images?
+        if: always()
         run: |
-          if [ "${{ github.event.inputs.upload_images }}" = "true" ] || \
+          if [ "${{ github.event.inputs.testing_image_upload }}" = "true" ] || \
              ( \
               [ "${{ github.repository_owner }}" = "syslog-ng" ] && \
-              [ "${{ github.ref }}" = "refs/heads/master" ] && \
-              [ "${{ github.event_name }}" = "push" ] \
+              [ "${{ github.ref }}" = "refs/heads/master" ] \
```
alltilla, line 40: I understand that there is filtering in the `on:` keyword above, so we do not have to check them here, but I am afraid that someone adds a new `on:` trigger, without thinking through the effects of it in this step. Maybe we should be stricter here. What do you think?

.github/workflows/dbld-images.yml
```diff
@@ -3,17 +3,11 @@ name: Compile dbld-images
 on:
   pull_request:
   push:
-    paths:
-      - .github/workflows/dbld-images.yml
-      # List is from dbld/rules cache-image-% "WATCHED_FILES"
-      - "dbld/**"
-      - "packaging/rhel/syslog-ng.spec"
-      - "packaging/debian/control"
```
alltilla, line 11: For some reason I remembered that these paths were added to the `pull_request` trigger. Wouldn't they makes sense there even after your changes?

alltilla (commented):
Nice changes, thanks! :)

I only have 2 minor comments, both are optional.

.github/workflows/dbld-images.yml
gaborznagy: I see your point, but I wouldn't add that extra protection: right now every trigger that is listed under the `on` keyword is allowed to upload images.
I would only start explicitly checking for the event type if we introduce events that are not allowed to upload images.

alltilla: Why is `pull_request` allowed to upload images? It will upload to the fork's registry, and the images will never be used later, so IMO it is unnecessary.

gaborznagy: Well, it's not true what I said: we run the github actions on pull requests and it's not allowed to upload images from a pull request.
My bad, sorry.

I'll add an explicit check for event types: `workflow_dispatch`, `push` and `schedule` will be allowed to upload images.
What do you think?

.github/workflows/dbld-images.yml
gaborznagy: I've checked the git history and I didn't find anything about pull requests and the above paths.

We could add those lines under the `pull_request` event type, if we would like to save some workflow run on PRs (i.e. only check dbld image build if dbld or packaging is changed).
As far as I understand GitHub pricing, we don't need to save because of that.

Another reason why we could limit this workflow to certain pull requests only is to avoid a CI failure due to an updated build-dependency. However since we don't use the newly built docker image for checking if syslog-ng can be compiled on it, I don't think that would be useful.

.github/workflows/dbld-images.yml
alltilla: I like that idea :)

.github/workflows/dbld-images.yml
alltilla: I am worried about the build time, but that does not seem to be a problem right now, so It is fine by me, if it stays like this.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

.github/workflows/dbld-images.yml
gaborznagy: I've added a check for the github event type using bash pattern matching with regex (`=~`).
I've tested the logic with a modified yaml file in my fork:
https://github.com/gaborznagy/syslog-ng/runs/4584715528?check_suite_focus=true

.github/workflows/dbld-images.yml
```diff
@@ -55,15 +50,15 @@ jobs:
           gh_export UPLOAD_IMAGES_INTERNAL
 
       - name: Log in to the Container registry
-        if: env.UPLOAD_IMAGES_INTERNAL == 'true'
+        if: always() && env.UPLOAD_IMAGES_INTERNAL == 'true'
```
bazsi, line 53: why do we need the ```always()``` portion of the conditional?

.github/workflows/dbld-images.yml
```diff
@@ -35,15 +29,16 @@ jobs:
...
       - name: Should we upload the images?
+        if: always()
         run: |
-          if [ "${{ github.event.inputs.upload_images }}" = "true" ] || \
+          if [ "${{ github.event.inputs.testing_image_upload }}" = "true" ] || \
              ( \
               [ "${{ github.repository_owner }}" = "syslog-ng" ] && \
               [ "${{ github.ref }}" = "refs/heads/master" ] && \
-              [ "${{ github.event_name }}" = "push" ] \
+              ( [[ "${{ github.event_name }}" =~ ^(push|workflow_dispatch|schedule) ]] ) \
```
bazsi, line 41: can you explain why we need to enclose this into parenthesis and use [[ ]] instead of single ones? Or if we need that why not convert the rest of the conditional too?

thanks (my bash fu is getting old I guess)

.github/workflows/dbld-images.yml
```diff
@@ -55,15 +50,15 @@ jobs:
...
+        if: always() && env.UPLOAD_IMAGES_INTERNAL == 'true'
         uses: docker/login-action@v1
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Push the images
-        if: env.UPLOAD_IMAGES_INTERNAL == 'true'
+        if: always() && env.UPLOAD_IMAGES_INTERNAL == 'true'
```
bazsi, line 61: likewise, why do we need always() here?

.github/workflows/dbld-images.yml
```diff
@@ -55,15 +50,15 @@ jobs:
...
         run: |
           IMAGES=$(docker image ls --filter "reference=${{ env.CONTAINER_REGISTRY }}/dbld-*:latest" --format '{{.Repository}}:{{.Tag}}')
```
bazsi, line 63: couldn't we push this functionality into dbld/rules ? I mean we already have a ./dbld/rules pull-image-<whatever>, we could have a push-image or even a push-images target so the github action would become thinner and we would gain the ability to push images easily even when we are working interactively in dbld shell.

bazsi (dismissed):
I had a few questions that are more to fulfill my curiosity than nitpicks. Also I have a more concrete request regarding dbld/rules push-image but that's completely optional too.

approved on my side.

.github/workflows/dbld-images.yml
gaborznagy: I agree it looks weird, but this the GitHub actions way to allow a following step to run if the previous fails.
https://docs.github.com/en/actions/learn-github-actions/expressions#job-status-check-functions

As the documentation states, there is an implicit `success()` condition at each step:
> A default status check of success() is applied unless you include one of these functions.

.github/workflows/dbld-images.yml
gaborznagy: I've only met this recently, but inside bash's test operator `[[` you can use a regex pattern matching.
I only wanted to avoid a long logical expression checking against 3 types of github event types, so I've simplified it to one regex pattern.

https://www.gnu.org/software/bash/manual/bash.html#Conditional-Constructs

> An additional binary operator, ‘=~’, is available, with the same precedence as ‘==’ and ‘!=’. When it is used, the string to the right of the operator is considered a POSIX extended regular expression and matched accordingly (using the POSIX regcomp and regexec interfaces usually described in regex(3)).

.github/workflows/dbld-images.yml
gaborznagy: I'ts the same: https://github.com/syslog-ng/syslog-ng/pull/3860#discussion_r772980844

.github/workflows/dbld-images.yml
gaborznagy: That sounds like a good idea. I'll add it soon.

gaborznagy:
> I had a few questions that are more to fulfill my curiosity than nitpicks. Also I have a more concrete request regarding dbld/rules push-image but that's completely optional too.
> 
> approved on my side.

Thank you for the review Bazsi. \o/
I would like to do the `push-images` part, just like we have pull-images.

kira-syslogng:
Build SUCCESS

gaborznagy:
@bazsi I've added `dbld/rules push-images` target.

Checked image pushing on my fork:
https://github.com/gaborznagy/syslog-ng/actions/runs/1606505388

kira-syslogng:
Build SUCCESS

.github/workflows/dbld-images.yml
```diff
@@ -55,18 +50,14 @@ jobs:
...
       - name: Push the images
-        if: env.UPLOAD_IMAGES_INTERNAL == 'true'
+        if: always() && env.UPLOAD_IMAGES_INTERNAL == 'true'
         run: |
-          IMAGES=$(docker image ls --filter "reference=${{ env.CONTAINER_REGISTRY }}/dbld-*:latest" --format '{{.Repository}}:{{.Tag}}')
-          for IMAGE in $IMAGES; do
-            echo "Pushing image: $IMAGE"
-            docker push $IMAGE
-          done
+          dbld/rules push-images -j $(nproc) --output-sync
```
bazsi, line 63: previously this only pushed built images, while now it pushes all images ./dbld/rules knows about. what happens if a build does not finish successfully?

It looks good to me otherwise.

.github/workflows/dbld-images.yml
bazsi: I understand [[]] now but what about the parenthesis?

.github/workflows/dbld-images.yml
gaborznagy: Yes, good point, I was just adding this quickly in a hurry.

How about a cheap solution: adding `--keep-going` to dbld/rules?
It will make the upload step failed, but I don't know if that's an issue since the image build already failed too.

I don't think `push-images` should do a filtering for existing images internally, or using the previous `docker image ls ` filtering either.

.github/workflows/dbld-images.yml
gaborznagy: that's probably unnecessary at this point: I was trying to solve this in a different way first, which needed the parentheses to group the same way it's already used.
I'll remove it.

kira-syslogng:
Build SUCCESS

.github/workflows/dbld-images.yml
bazsi: I like the --keep-going option. that would upload those that succeeded and the job would still fail, but I think that is fair.

bazsi (dismissed):

gaborznagy:
@bazsi: Thanks for the approve! I've added one last improvement to the PR: a regex end character (`$`) is added to the regex pattern that checked the triggering event in github actions.

alltilla (approved):

kira-syslogng:
Build SUCCESS

bazsi (approved):

PR #3863, opened 2021-12-23
Fix cloning negated filters 

MrAnno:
When cloning negated filter expressions, the negation field was left out.

This was a regression in v3.35.

Config to reproduce the issue:

```
filter f_f2 {
  not (program("a") and message("b"));
};

filter f_f3 {
  filter(f_f2)
};

log {
  source { stdin(); };
  filter(f_f3);
};
```

Prior to this fix, `f_f2` and `f_f3` produced non-identical results.

Note: this PR is about fixing the actual issue. I'm preparing another branch for some refactor around filter expressions and for another filter-related bugfix.

Reported by @hbakken on IRC.

### DIFF
--- a/lib/filter/filter-expr.c
+++ b/lib/filter/filter-expr.c
@@ -98,10 +98,10 @@ filter_expr_unref(FilterExprNode *self)
 FilterExprNode *
 filter_expr_clone(FilterExprNode *self)
 {
-  if (self->clone)
-    {
-      return self->clone(self);
-    }
+  if (!self->clone)
+    return filter_expr_ref(self);
 
-  return filter_expr_ref(self);
+  FilterExprNode *cloned = self->clone(self);
+  cloned->comp = self->comp;
+  return cloned;
 }
--- a/news/bugfix-3863.md
+++ b/news/bugfix-3863.md
@@ -1,0 +1,5 @@
+filters: fix `not` operator in filter expressions (regression in v3.35.1)
+
+Reusing a filter that contains the `not` operator more than once, or
+referencing a complex expression containing `not` might have caused invalid results
+in the previous syslog-ng version (v3.35.1).  This has been fixed.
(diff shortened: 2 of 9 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi (dismissed):
Good catch and nice test coverage.

MrAnno:
I've rebased this PR against master and added a minor modification to the Light test.

kira-syslogng:
Build SUCCESS

news/bugfix-3863.md
```diff
@@ -0,0 +1,5 @@
+filters: fix `not` operator in filter expressions
```
bazsi, line 1: I think we should include which version was buggy.

bazsi (dismissed):

news/bugfix-3863.md
MrAnno: I've mentioned the affected version.

kira-syslogng:
Build SUCCESS

mitzkia:
Light improvement and the new testcase is approved from my side.

lib/filter/filter-expr.c
```diff
@@ -98,10 +98,10 @@ filter_expr_unref(FilterExprNode *self)
...
-      return self->clone(self);
-    }
+  if (!self->clone)
+    return filter_expr_ref(self);
 
-  return filter_expr_ref(self);
-}
\ No newline at end of file
+  FilterExprNode *cloned = self->clone(self);
+  cloned->comp = self->comp;
```
alltilla, line 105: Shouldn't we also copy the `modify` field?

alltilla (approved):
Nice catch!

I only had 1 question, which is not strictly related to the bug you are fixing. Apart from that it is an approve from me.

lib/filter/filter-expr.c
MrAnno: We should, I agree. "Modify" is currently set by the init method of FilterRE and FilterOp, so I think their "cloning" works correctly.

I'd like to do a refactor round in a separate PR, which will hopefully make filter-clone-related things a bit cleaner.
I've also found a possible cloning bug, which affects the rate-limit/throttle filter only.

Please let me know if you think the `modify` field is not working properly, in that case we should fix that too ASAP.

OverOrion (approved):

PR #3864, opened 2021-12-28
lgtm: minor fixes

MrAnno:
LGTM reports gboolean bit fields as ambiguously signed declarations.

This is because `gboolean` is typedefed to `gint`, and it is implementation specific whether an int is signed or unsigned.

Since `gboolean` is meant to represent a TRUE/FALSE value, this is not a real issue.

I've added a [modified CodeQL query](https://github.com/github/codeql/blob/main/cpp/ql/src/Likely%20Bugs/AmbiguouslySignedBitField.ql) that accepts gboolean bit fields.

### DIFF
--- a/modules/basicfuncs/str-funcs.c
+++ b/modules/basicfuncs/str-funcs.c
@@ -206,7 +206,8 @@ TEMPLATE_FUNCTION_SIMPLE(tf_strip);
 typedef struct _TFSanitizeState
 {
   TFSimpleFuncState super;
-  gint ctrl_chars:1, replacement:8;
+  gboolean ctrl_chars;
+  gchar replacement;
   gchar *invalid_chars;
 } TFSanitizeState;
 
--- a/modules/dbparser/radix.c
+++ b/modules/dbparser/radix.c
@@ -523,17 +523,26 @@ r_parser_ip(gchar *str, gint *len, const gchar *param, gpointer state, RParserMa
   return r_parser_ipv4(str, len, param, state, match) || r_parser_ipv6(str, len, param, state, match);
 }
 
+static inline void
+_scan_digits(gchar *str, gint *len)
+{
+  while (g_ascii_isdigit(str[*len]))
+    (*len)++;
+}
+
 gboolean
 r_parser_float(gchar *str, gint *len, const gchar *param, gpointer state, RParserMatch *match)
 {
-  gboolean dot = FALSE;
-
   *len = 0;
   if (str[*len] == '-')
     (*len)++;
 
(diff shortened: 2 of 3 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
Upstream enquiry:
https://github.com/github/codeql/issues/7491

MrAnno:
I'm marking this as "draft", it seems upstream is interested in this change. I'm opening a PR there too.

kira-syslogng:
Build SUCCESS

MrAnno:
github/codeql#7493 has been merged, I'm removing the now irrelevant part from this PR.

modules/basicfuncs/str-funcs.c
```diff
@@ -206,7 +206,8 @@ TEMPLATE_FUNCTION_SIMPLE(tf_strip);
 typedef struct _TFSanitizeState
 {
   TFSimpleFuncState super;
-  gint ctrl_chars:1, replacement:8;
+  gboolean ctrl_chars:1;
```
bazsi, line 209: I think we can drop using bitfields here.

bazsi (approved):

modules/basicfuncs/str-funcs.c
MrAnno: I've replaced the bitfield. Thanks.

kira-syslogng:
Build SUCCESS

modules/dbparser/radix.c
```diff
@@ -532,8 +532,13 @@ r_parser_float(gchar *str, gint *len, const gchar *param, gpointer state, RParse
...
 
-  while (g_ascii_isdigit(str[*len]) || (!dot && str[*len] == '.' && (dot = TRUE)))
-    (*len)++;
+  while (g_ascii_isdigit(str[*len]) || (!dot && str[*len] == '.'))
+    {
+      if (str[*len] == '.')
+        dot = TRUE;
+
+      (*len)++;
+    }
```
Kokan, line 541: If we are at readability, I would suggest something like this:
```c
void read_digists(gchar *str, gint *len)
{
   while (str[*len] && g_ascii_isdigit(str[*len])) (*len)++;
}
```

```c
read_digits(str, len);
if (str[*len} && str[*len] == '.')
{
    read_digits(str, len);
}
```

Note: this also fixes that `.` are not valid number, but not solves `.12` (the later can be argued to be "valid").

MrAnno: Done, thanks.

kira-syslogng:
Build SUCCESS

Kokan (approved):

OverOrion (approved):
LGTM!

PR #3866, opened 2021-12-29
throttle-filter: extract it to separate module

MrAnno:
This PR extracts the `throttle()` filter to a separate plugin, which has now become our first filter plugin.

Suggested by @bazsi: https://github.com/syslog-ng/syslog-ng/pull/3781#issuecomment-995576182

Note: `throttle()` conflicts with the throttle keyword, so I renamed the plugin to `rate-limit()`, but added back `throttle()` in a tricky (and terrible) way for compatibility.

TODO:
- [x] #3863 contains a rate-limit() bugfix, it should be reapplied here too


MrAnno:
Almost all of our modules are licensed as [GPL-2.0-or-later](https://spdx.org/licenses/GPL-2.0-or-later.html).

The `throttle()` filter's license is currently [LGPL-2.1-or-later](https://spdx.org/licenses/LGPL-2.1-or-later.html), but changing it to the usual GPL-2.0-or-later would benefit the open source community in my opinion.

@sinkingpoint Could we have your permission to re-release your contribution under GPL-2.0-or-later?

sinkingpoint:
Go for it :)

MrAnno:
@czanik The `throttle()`/`rate-limit()` builtin filter is becoming a separate plugin. I've modified the rpm spec file to include this plugin in the "core" syslog-ng package. Could you please do the same with all of your spec files after this PR gets merged?

This would be important because the mentioned filter was introduced as a builtin in v3.35, and we don't want anyone to loose that functionality in the next release.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

czanik:
@MrAnno Thanks for the heads up!

lib/filter/filter-expr-grammar.ym
```diff
@@ -144,9 +140,24 @@ filter_simple_expr
...
 	| filter_plugin
 	| filter_comparison
-	| filter_throttle   { $$ = last_filter_expr; }
+	| filter_rate_limit_compat
 	;
 
+filter_rate_limit_compat
+  : KW_THROTTLE
+      {
+        Plugin *p = cfg_find_plugin(configuration, LL_CONTEXT_FILTER, "rate-limit");
```
bazsi, line 149: why do we rename the plugin to rate-limit()?

which use is the "official" one? shouldn't we deprecate the other?

lib/filter/filter-expr-grammar.ym
```diff
@@ -144,9 +140,24 @@ filter_simple_expr
 	| filter_re					{ $$ = last_filter_expr; }
 	| filter_plugin
 	| filter_comparison
-	| filter_throttle   { $$ = last_filter_expr; }
+	| filter_rate_limit_compat
 	;
 
+filter_rate_limit_compat
```
bazsi, line 146: this rule could be merged into the filter_plugin to avoid code duplication, e.g. create a non-terminal that is either LL_IDENTIFIER or KW_THROTTLE and use that in the filter_plugin rule.

bazsi (commented):
Thanks for working on this. I had a few notes that I would love to discuss before this is merged.

lib/filter/filter-expr-grammar.ym
MrAnno: The plugin's name conflicts with an existing keyword. I renamed it to show that such a name conflict is not acceptable when implementing plugins.

I think `rate-limit()` should be the official one, hence the news entry about the rename.
The KW_THROTTLE trick is only there for compatibility.

lib/filter/filter-expr-grammar.ym
MrAnno: When KW_THROTTLE is found in this context, the name of the plugin we are looking for has to be changed.
But I think I can do that without code duplication, let me try it.

MrAnno: Fixed.

kira-syslogng:
Build SUCCESS

lib/filter/filter-expr-grammar.ym
bazsi: To use a plugin name that matches the name of a core provided keyword does cause complications I agree. In the long term, this is probably best assured by keeping new features as separate plugins.

For the current case we need a workaround anyway (which you have implemented). With the workaround in place we can choose whether we want to reuse the keyword (and keep the name we accepted for the previous version of syslog-ng) or we don't (and we communicate the change). We will need a workaround in both cases.

I am OK both ways.

If we pursue the name-change, however, we should also rename the module directory, the .so files produced and any leftover traces of the old name.

lib/filter/filter-expr-grammar.ym
```diff
@@ -168,6 +162,16 @@ filter_plugin
           }
 	;
 
+filter_identifier
+  : LL_IDENTIFIER
+  | KW_THROTTLE
+      {
+        msg_warning_once("WARNING: throttle() is deprecated, the filter has been renamed to rate-limit()",
```
bazsi, line 169: WARNING: the throttle() filter has been renamed to rate-limit() in syslog-ng VERSION, please update your configuration to use the name rate-limit() instead of throttle()

lib/filter/filter-expr-grammar.ym
MrAnno: Oh sorry, I forgot to rename the module, .so, functions, etc.

If we agree with the name change, I'll rename those as well.

@Kokan @alltilla @bazsi  What do you suggest? My train of thought was to rename the module due to the name conflict to make it clear this is not something that we usually do. But I needed to add a workaround to be backward-compatible, so the whole idea might sound silly :) Should I drop the name change?

bazsi: I think whether we should rename the module should come from a user perspective and not a technical one. 

Which one is easier to discover: find it in the documentation or via google search?

If they are equivalent, we should stick with throttle() as that's the one we already merged and have available. One less change to poke users with. 

If rate-limit() is much better, we can and should do the rename now, keeping the compatibility for those who already evaluated the feature.

@sinkingpoint came up with the original name and we accepted it. He may have a perspective, he may have had pattern in mind in similar software and that's how he came up with the name.

With that said:
* rsyslog seems to call a similar feature "rate-limiting": https://rsyslog.adiscon.narkive.com/nfFmcc1b/is-there-a-mechanism-in-to-throttle-the-messages-being-forwarded-to-the-repositories
* nifi: ControlRate, albeit the "throttle" verb is also used in the documentation: https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-standard-nar/1.12.0/org.apache.nifi.processors.standard.ControlRate/index.html
* logstash: ratelimit, again throttle is used as a verb in the description, https://packetrevolt.com/2018/02/a-simple-logstash-ratelimiter/

Also, syslog-ng uses the throttle() option to "blindly" throttle the output at the specific messages/seconds. This filter is a lot smarter.

With that I would vote for the change, unless @sinkingpoint has another argument.

MrAnno: `rate-limit()` sounds a bit "smarter" to me too, imho it would suit such a template-based filter :)

MrAnno: A side-note: In the REST  world and in other web APIs, throttle and rate-limit are sightly different methods, but the key aspects are the same as the ones @bazsi described. Throttling is enforced by the server, it usually affects the whole API service. It's basically a global "stupid" thing. Rate-limit is "smarter", they can be defined differently on each resource of the API, and the input itself can be taken into account.

TLDR: the difference is roughly the following: "key-level rate-limiting" vs" global-API-level rate-limiting"

Kokan:
The https://github.com/syslog-ng/syslog-ng/pull/3815 was merged with light testcases using old throttle name, please update those as well.

MrAnno:
I've addressed all review notes and completed the rename of throttle() to rate-limit().

kira-syslogng:
Build SUCCESS

bazsi (approved):

gaborznagy (approved):

PR #3874, opened 2022-01-15
Fix a rare disk-buffer corruption issue

MrAnno:
A disk-buffer is considered empty when backlog_head (and read_head) points to write_head. The same could have been done when the buffer is full, but currently we don't accept that, because we can't use `length` to distinguish between these two states.

Due to the above, when a disk-buffer is filled completely, write_head will point to the end of the disk-buffer file as backlog_head prevents the wrapping logic to do its job.

In that specific case, if the disk-buffer then gets emptied, the read and the backlog heads will point to the end of the buffer as well. An additional write will wrap the write head and allow the read head to read into an invalid position, which ends up detected as an invalid disk-buffer file.

This has been fixed by adding a preliminary check to each disk-buffer operation (qdisk_skip_record() is used for moving backlog_head).

Note: we should probably allow "read_head == backlog_head == write_head" situations when the buffer is full in the future in order to be completely consistent with the size of the buffer (if we don't allow that, we can fill a 400 bytes long disk-buffer with two 200-byte messages, but we can NOT pop one of them and then push it back, because we require a single byte empty space between backlog_head and write_head).


MrAnno:
@kira-syslogng do strestest

kira-syslogng:
Build SUCCESS

kira-syslogng:

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

bazsi (commented):
This looks very good at first sight. I can see a Todo comment and the last commit might be a fixup nőtt yet squashed, but I am ready to approve.

4am? That was a long night.

MrAnno:
@bazsi Thank you for the review. I'll address the todo comment next week.

It was more of a "morning" for me, I woke up probably because I felt I was too close to fixing this :)

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

OverOrion (dismissed):
LGTM!

modules/diskq/tests/test_qdisk.c
```diff
@@ -164,6 +164,41 @@ Test(qdisk, qdisk_basic_push_pop)
...
+                "There should not be more free space than the previously popped message size");
+
+  record_len -= 100;
+  push_dummy_record(qdisk, record_len);
+  /* 1 byte of empty space (between backlog and write head) is reserved */
+  cr_assert(qdisk_is_space_avail(qdisk, 100 - 1));
+
+  qdisk_stop(qdisk);
+  g_string_free(data, TRUE);
+  cleanup_qdisk(filename, qdisk);
```
alltilla, line 199: In case of a failure the file does not get cleaned up, and it messes with the next test run. Can you add this to a `.fini` function?

modules/diskq/qdisk.c
```diff
@@ -213,6 +213,13 @@ _is_free_space_between_write_head_and_backlog_head(QDisk *self, gint msg_len)
   return self->hdr->write_head + msg_len < self->hdr->backlog_head;
 }
 
+static inline gboolean
+_is_free_space_at_the_beginning_of_qdisk(QDisk *self, gint msg_len)
```
alltilla, line 211: [optional]

The function name feels weird a bit, can you consider `_is_there_free_space_at_the_beginning_of_qdisk()`?

modules/diskq/qdisk.c
```diff
@@ -230,7 +237,11 @@ qdisk_is_space_avail(QDisk *self, gint at_least)
       if (_is_write_head_less_than_max_size(self))
         return TRUE;
 
-      return _is_able_to_reset_write_head_to_beginning_of_qdisk(self);
+      /* exact size-check is needed as we have unread/unacked data after the write head
+       * that is being reset
+       */
+      return _is_able_to_reset_write_head_to_beginning_of_qdisk(self)
+             && _is_free_space_at_the_beginning_of_qdisk(self, at_least);
```
alltilla, line 238: I think `return _is_free_space_at_the_beginning_of_qdisk(self, at_least);` is enough.

My (overcomplicated) proof:

Let A be `_is_able_to_reset_write_head_to_beginning_of_qdisk()` and B be `_is_free_space_at_the_beginning_of_qdisk()`.

Possible cases are:
  * A: true, B: true -> true
  * A: true, B: false -> false
  * A: false, B: true -> false
  * A: false B: false -> false

When checking the implementation of A and B, if we presume that msg_size is 0 or larger then there cannot be a case where A is false and B is true.

If we remove `A: false, B: true -> false` from the cases above, we get this:
  * A: true, B: true -> true
  * A: true, B: false -> false
  * A: false B: false -> false

From this we can see, that the result is true if and only if B is true. So it is enough to only check B.

alltilla (approved):
I really like the changes! :)
I left some comments, none of them are strictly necessary to fix, but please consider them.
If you decide not to change code, let me know and I will merge the PR!

modules/diskq/tests/test_qdisk.c
```diff
@@ -0,0 +1,180 @@
...
+  scratch_buffers_reclaim_marked(marker);
+  return TRUE;
+}
+
+gboolean
+push_dummy_record(QDisk *qdisk, guint record_size)
+{
+  /* TODO: fix interface:
+   * framing should be the responsibility of push_tail() because pop_head() expects framing
+   */
```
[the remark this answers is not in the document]
MrAnno: Let's measure the performance after simplifying the interface.

modules/diskq/qdisk.c
```diff
@@ -189,7 +189,7 @@ _is_qdisk_overwritten(QDisk *self)
 }
 
 static inline gboolean
-_is_backlog_head_prevent_write_head(QDisk *self)
+_is_backlog_head_precedes_write_head(QDisk *self)
```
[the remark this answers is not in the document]
MrAnno: Oh no, that's bad... Sorry, I'm going to fix it ASAP.

modules/diskq/tests/test_qdisk.c
```diff
@@ -0,0 +1,180 @@
...
+}
+
+gboolean
+reliable_pop_record_without_backlog(QDisk *qdisk, GString *record)
+{
+  if (!qdisk_pop_head(qdisk, record))
+    return FALSE;
+
+  qdisk_set_backlog_head(qdisk, qdisk_get_head_position(qdisk));
+  return TRUE;
```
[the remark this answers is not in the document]
MrAnno: I'm planning to simplify the interface, or if that's not possible due to performance reasons, I'd like to add abstractions that show how the API should be used correctly. Since I'm planning to do this in the near future, I'd leave the test as-is in this PR.

modules/diskq/tests/test_qdisk.c
MrAnno: I think it was a deliberate decision not to do this in failing tests so that we can collect them as artifacts when a flaky test fails in our CI.

modules/diskq/qdisk.c
MrAnno: This follows the name of another function: `_is_free_space_between_write_head_and_backlog_head()`. I can rename both.

modules/diskq/qdisk.c
MrAnno: You are absolutely right. The only reason I would rather leave this condition in is that it might help the learning/understanding process of how qdisk works.

(If I'm lucky, I'll even be able to remove all of these pre-checks from the code and replace all position-correction logic with a simple modulo operation in the next PR.)

modules/diskq/tests/test_qdisk.c
```diff
@@ -0,0 +1,180 @@
...
+create_qdisk(const gchar *filename, TestDiskQType dq_type, gint64 disk_buf_size)
+{
+  DiskQueueOptions *opts = construct_diskq_options(TDISKQ_RELIABLE, disk_buf_size);
+  QDisk *qdisk = qdisk_new(opts, "TEST");
+
+  return qdisk;
+}
+
+static void
+cleanup_qdisk(const gchar *filename, QDisk *qdisk)
```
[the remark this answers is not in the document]
MrAnno: I could not, because `qdisk_stop()` deallocates the filename before calling the cleanup function.

I decided not to add `qdisk_stop()` to `cleanup_qdisk()` so that we can play with restarts inside tests.

OverOrion (approved):

modules/diskq/tests/test_qdisk.c
alltilla: Ah, makes sense, thanks :)

modules/diskq/qdisk.c
alltilla: Cool, thanks for thinking it through:)

alltilla (approved):

kira-syslogng:
Build SUCCESS

PR #3879, opened 2022-01-17
Fix worker startup race with exit notifications

bazsi:
This is a fix for #3867 which has become way more complicated than I originally anticipated. The race described in the issue happens because the thread of internal() driver is requested to exit while the iv_event used for this purpose as not yet registered at that point.

Delving into the problem further it became apparent that the same bug exists in LogThreadedSourceDriver, but not in LogThreadedDestDriver, which implemented proper startup synchronization.

The solution in this pull request first extracts the MainLoopThreadedWorker layer into its own module (earlier this part of MainLoopWorker) and makes this a proper class. Then it ports thread related functionality from LogThreadedDestDriver up to a point where they are generalized for all of these use-cases.

The fix for LogThreadedSourceDriver/FetcherDriver were a bit more complicated than the other two, simply because LogThreadedSourceDriver made an attempt at "hiding" its worker class (LogThreadedSourceWorker) in the implementation module. Although the final solution should be to match the design of LogThreadedSourceDriver with LogThreadedDestDriver, I stopped at a point where the race could be fixed.

I consulted @MrAnno during the implementation, so his review is appreciated.


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
the CI detected a memory corruption issue, the MainLoopThreadedWorker instance is freed by the embedding object causing a heap corruption :(

I'll probably have to make that object an allocated one, unless I have a better idea. Regardless of the bug this can already be reviewed.

kudos to @kira-syslogng for finding the issue :)

bazsi:
hopefully the heap corruption is fixed.

kira-syslogng:
Build SUCCESS

MrAnno:
@bazsi  Thank you for fixing the race. I really like the new `mainloop-threaded-worker` unit, those thread-specific things in `mainloop-worker` have been bothering me for a long time.

I'll review this PR by next Monday.

MrAnno:
I've found an old patchset from 2019, where I made the worker of the threaded source driver public: https://github.com/MrAnno/syslog-ng/commits/thrsource-worker

This branch seems to fix some kind of a crash too. I'm going to recreate those patches on the current master and open a PR from it so that we could match the design of the threaded source with the threaded destination.

lib/mainloop-threaded-worker.h
```diff
@@ -37,12 +37,17 @@ struct _MainLoopThreadedWorker
 
   gpointer data;
   MainLoopWorkerType worker_type;
+  GMutex lock;
+  GCond started_up;
+  gboolean startup_finished;
+  gboolean startup_result;
```
MrAnno, line 40: Can we somehow group these fields together to show the bool+cond+mutex is there for startup synchronization?
If that's inconvenient, I would just rename `lock` to `startup_lock` to indicate this is for the startup condition, not for protecting the whole class's resources.

lib/mainloop-threaded-worker.c
```diff
@@ -37,6 +37,25 @@ main_loop_threaded_worker_signal_startup_finished(MainLoopThreadedWorker *self,
   g_mutex_unlock(&self->lock);
 }
 
+static gboolean
+_thread_init(MainLoopThreadedWorker *self)
+{
+  main_loop_worker_thread_start(self->worker_type);
+
+  gboolean result = self->thread_init(self);
```
MrAnno, line 45: I think we should allow unimplemented `thread_init`/`thread_deinit()` methods.
(request_exit can be implemented with atomic variables, or the resource needed for this operation might be something that was allocated in the main thread at syslog-ng startup)

lib/logthrsource/logthrsourcedrv.h
```diff
@@ -74,6 +71,9 @@ struct _LogThreadedSourceDriver
   LogThreadedSourceWorker *worker;
 
   const gchar *(*format_stats_instance)(LogThreadedSourceDriver *self);
+  LogThreadedSourceWorkerRunFunc run;
+  LogThreadedSourceWorkerRequestExitFunc request_exit;
+  LogThreadedSourceWorkerWakeupFunc wakeup;
```
MrAnno: I think these belong to the worker.

I'm okay with keeping them here for this PR, but I'd like to sort these out in a follow-up PR, like this:
https://github.com/MrAnno/syslog-ng/commits/thrsource-worker

lib/afinter.c
```diff
@@ -362,6 +358,10 @@ afinter_source_new(AFInterSourceDriver *owner, AFInterSourceOptions *options)
   log_source_init_instance(&self->super, owner->super.super.super.cfg);
   log_source_set_options(&self->super, &options->super, owner->super.super.id, NULL, FALSE,
                          owner->super.super.super.expr_node);
+  main_loop_threaded_worker_init_instance(&self->thread, MLW_THREADED_INPUT_WORKER, self);
+  self->thread.run = afinter_source_run;
+  self->thread.request_exit = afinter_source_thread_request_exit;
```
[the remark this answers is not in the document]
MrAnno: This has been addressed by the last commit.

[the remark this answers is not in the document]
MrAnno: @bazsi After fixing this, I think the `MainLoopThreadedWorker` interface started to miss something: a method which can check whether the thread is exited. (like a "busy" `join()` or "peekable" `is_finished` method).

`main_loop_threaded_worker_clear()` should only be called after we made sure the thread is finished, that is why I miss such a method.

lib/logthrsource/logthrsourcedrv.h
bazsi: I agree, however in that case the "self" method should point to a worker instance which it didn't in this case. And it would have opened the scope of these patches a lot.

In a way these are similar to the compat methods in LogThreadedDestDriver which also reside in the driver class and take a pointer to the driver as self.

lib/afinter.c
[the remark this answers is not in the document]
bazsi: > This seems to imitate multiple inheritance, which is new in syslog-ng.
> 
> ...

If it's not clear to everyone, I agreed with @MrAnno so much that I actually changed from the "multiple-inheritance-like" stuff to a more usual "use-of" pattern.

lib/mainloop-threaded-worker.h
bazsi: What if I was using the same lock for exit synchronization? I wouldn't dedicate a separate lock for that. I could rename started_up to startup_cond or even do this:

```
struct
{
  GCond cond;
  gboolean finished;
  gboolean result;
} startup;
```

But I don't think "lock" belongs there. Lock granularity comes up for locks & data structures that are used at a high frequency but in this case a single lock for the current/possibly future uses in this data structure should suffice. I agree
that documenting the locking strategy is a good idea, but taking it into a direction where _all_ synchronization needs to use a separate, specific lock is not right.

I'll implement the struct then you can respond and we discuss :)

lib/afinter.c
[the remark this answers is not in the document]
bazsi: I've made an attempt to add the "join()" functionality, however I found that:

a) it is not really needed (as mainloop already ensures that the thread exits by that t ime)
b) there's not always a good call site for doing main_loop_threaded_worker_join() (for instance, LogThreadedDestWorker, the only place we could call is right where we call _clear())

With those in mind, I've let the idea of a main_loop_threaded_worker_join() method go. But at the same time, I've added
a g_thread_join() right within clear() along with a nice comment to explains it. This also fixes a reference counting issue
of the GThread instance returned by g_thread_new().

With that I consider this conversation resolved.

alltilla (commented):
Thanks for the PR, the refactored code became a lot more easier to understand!

I have finished the review and did not find any additional comments. I am running some additional internal tests on threaded source based drivers, if they pass, I will approve this PR.

lib/afinter.c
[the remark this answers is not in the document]
MrAnno: Sounds perfect to me. Thank you.

lib/mainloop-threaded-worker.h
MrAnno: My idea was based on Rust mutexes, where a lock encapsulates the protected resource/data and only allows access to those resources through RAII guards. This explicitly binds the protected entity to the lock, but I agree, it would be overkill here.

Thank you for talking this through with me.

kira-syslogng:
Build SUCCESS

lib/logthrsource/logthrsourcedrv.h
MrAnno: @bazsi mentioned that it might be worth keeping this interface for "simplified" threaded sources even after we allow multiple workers in threaded sources. I'm marking this comment as resolved, as I'll try refactoring/implementing the remaining bits in a separate PR.

kira-syslogng:
Build FAILURE

MrAnno (approved):
Thank you.

I just saw the last patch about resolving a deadlock caused by the new "just-for-safety" `join()` call. I didn't think it would cause such interesting complications.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

alltilla (approved):

PR #3884, opened 2022-01-22
versioning: add versioning framework for the syslog-ng 4.0 transition (4.0)

bazsi:
The plan to release syslog-ng 4.0 would be as follows:

  * we are accumulating changes destined for 4.0 by merging them into
    subsequent syslog-ng 3.x releases, but working in a compatible mode so
    that no changes would be visible to 3.x users.

  * Usually we report changes to behavior using warnings at
    startup and we do this release-by-release. With 4.0, we expect a broader
    set of changes that we would like to present to users as a whole. For
    this reason we will suppress these warnings for changes we plan for
    4.0 at least until the user explicitly indicates interest in the
    experimental features.

  * the user can start using the new features as follows:

    - by using @version: 3.255 in a config, the warnings about changes get
      enabled (while continuing to work in 3.x compatibility mode)

    - by using @version: 4.0 in a config, the changes are enabled and
      syslog-ng would start working in the new mode we plan for 4.0.

    - in a way all new features/changes would operate behind a feature flip,
      which the user can control using the @version line in a
      configuration.

  * the features provided by the 4.0 mode are experimental and this mode is
    not intended for production use.

  * once the 4.0 features/changes are all baked, we release syslog-ng 4.0,
    which would make the new features/behaviours official. The v3.255 mode
    of operation would be disabled and the feature flip removed. Essentially
    the only change would be that upgrade warnings would always get enabled.

### DIFF
--- a/lib/cfg.h
+++ b/lib/cfg.h
@@ -184,13 +184,43 @@ __cfg_is_config_version_older(GlobalConfig *cfg, gint req)
 
 /* VERSION_VALUE_LAST_SEMANTIC_CHANGE needs to be bumped in versioning.h whenever
  * we add a conditional on the config version anywhere in the codebase.  The
- * G_STATIC_ASSERT checks that we indeed did that */
+ * G_STATIC_ASSERT checks that we indeed did that.
+ *
+ * We also allow merging the features for the upcoming major version.
+ */
 
 #define cfg_is_config_version_older(__cfg, __req) \
   ({ \
-    /* check that VERSION_VALUE_LAST_SEMANTIC_CHANGE is set correctly */ G_STATIC_ASSERT((__req) <= VERSION_VALUE_LAST_SEMANTIC_CHANGE); \
+    /* check that VERSION_VALUE_LAST_SEMANTIC_CHANGE is set correctly */ G_STATIC_ASSERT((__req) <= VERSION_VALUE_LAST_SEMANTIC_CHANGE || (__req) == VERSION_VALUE_NEXT_MAJOR); \
     __cfg_is_config_version_older(__cfg, __req); \
   })
+
+/* This function returns TRUE if a version based feature flip is enabled.
+ * This will enable or disable feature related upgrade warnings.
+ *
+ * This was initially introduced for "typing" support, where the following
+ * is implemented:
+ *
+ * As long as we are using a 3.x version number (less than 3.255), the upgrade
+ * warnings are suppressed and compatibility mode is enabled.  Once @version
+ * is something equal or larger than 3.255, we still enable compatibility but
+ * warnings will NOT be suppressed.  This is controlled with
+ * FEATURE_TYPING_MIN_VERSION.
+ *
+ * Once we release 4.0, FEATURE_TYPING_MIN_VERSION needs to be set to zero in
+ * versioning.h With that all upgrade notices start to appear when syslog-ng
+ * finds a 3.x configuration.
+ */
+#define __cfg_is_feature_enabled(cfg, min_version) \
+  ({ \
+    /* the flip-over for min_version reached, set min_version to 0 */ G_STATIC_ASSERT(VERSION_VALUE_CURRENT < ((min_version) - 1));       \
+    version_convert_from_user(cfg->user_version) >= (min_version) - 1;  \
+  })
+
+#define cfg_is_feature_enabled(cfg, topic)        __cfg_is_feature_enabled(cfg, FEATURE_ ## topic ## _MIN_VERSION)
+#define cfg_is_typing_feature_enabled(cfg)        cfg_is_feature_enabled(cfg, TYPING)
+
+#define cfg_is_experimental_feature_enabled(cfg)  (cfg_is_typing_feature_enabled(cfg))
 
 static inline void
 cfg_set_use_uniqid(gboolean flag)
--- a/lib/tests/test_lexer.c
+++ b/lib/tests/test_lexer.c
@@ -312,6 +312,7 @@ Test(lexer, at_version_stores_config_version_in_parsed_version_in_hex_form)
 {
   parser->lexer->ignore_pragma = FALSE;
 
+  start_grabbing_messages();
   cfg_set_version_without_validation(configuration, 0);
   _input("@version: 3.1\n\
 bar\n");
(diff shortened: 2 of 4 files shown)


kira-syslogng:
Build SUCCESS

bazsi:
The dbld images fail to build as "dwarves" (a tool I use in devshell from time to time) has become missing from debian-testing. It should resurface tomorrow, so I am not removing it from the build manifest.

MrAnno (approved):
An example of how the 4.0-versioning API can be used:

https://github.com/syslog-ng/syslog-ng/blob/77ace16b4a2976905c933909a4c46de087083522/modules/json/format-json.c#L86-L101

kira-syslogng:
Build SUCCESS

MrAnno (approved):

lib/cfg.h
```diff
@@ -188,10 +188,36 @@ __cfg_is_config_version_older(GlobalConfig *cfg, gint req)
...
+ */
+static inline gboolean
+cfg_is_config_to_be_migrated_to_post_4_0(GlobalConfig *cfg)
+{
+  /* this assert should fail once we release 4.0 */
+  G_STATIC_ASSERT(VERSION_VALUE_CURRENT < VERSION_VALUE_3_LAST);
+
+  if (!cfg)
+    return FALSE;
+  return version_convert_from_user(cfg->user_version) >= VERSION_VALUE_3_LAST;
```
Kokan, line 218: The `VERSION_VALUE_CURRENT` can contain higher value then 3.x or even 4.x, and yet it should not work in 4.0 mode.

bazsi: `VERSION_VALUE_3_LAST` is set to 3.99 which I assume will give us enough 3.x versions until we release 4.0

When we do release 4.0 this entire function and all its call sites should be removed.

The static assert in this function should ensure that when we set `VERSION_VALUE_CURRENT` to 4.0 compilation would fail, until at least the assert is removed, but the comment right next to the assert explains that the entire function should be removed at that point.

Does this clarify the intent here? Am I documenting this clearly enough? Or am I missing your point?

Kokan: You are completely missing my point. This code would break our internal building (due to assert), and without the assert it would make our version work in 4.0 mode. It is because you assume everything above 3.99 is 4.0. But in fact it should not. (See https://github.com/balabit-deps/syslog-ng-7.0-core/commit/4aa1246970375664e4eab35bd899a0266f02ed47 )

bazsi: I am not following the PE versioning, but that was the point of version_convert_from_user(cfg->user_version), e.g. to map the users version number to an OSE version.

7.0 could be mapped to VERSION_VALUE_CURRENT

MrAnno:
(Just for the record: I approved the PR to indicate that I like the idea of how we'd like to transition into v4. We're still discussing the details and its internal implications.)

bazsi:
I have now implemented version number based feature flips. I would appreciate help in checking if this would work your internal builds of PE.

Also, if version_from_user() is not used in versioning.h, can we eliminate that indirection? Also, a lot of defines in versioning.h is not actually used in the codebase, I'd probably eliminate that too.

MrAnno:
Thanks, @bazsi.

I'll check the internal parts by Friday.

kira-syslogng:
Build SUCCESS

MrAnno:
@bazsi I think I got a bit confused at the "OSE handover" discussion,  the current versioning PR makes it possible to map internal versions easily, for example:
https://gist.github.com/MrAnno/26bf5e3ee626790c6834dbcb0dfabc2e

It seems a bit overcomplicated because we need to redefine what VERSION_VALUE_4_0 and VERSION_VALUE_3_LAST mean anyway (making next_major and typing_min_version a bit unnecessary), but it works in my opinion. I'll test it in practice tomorrow.

bazsi:
> @bazsi I think I got a bit confused at the "OSE handover" discussion, the current versioning PR makes it possible to map internal versions, for example: https://gist.github.com/MrAnno/26bf5e3ee626790c6834dbcb0dfabc2e
> 
> It seems a bit overcomplicated because we need to redefine what VERSION_VALUE_4_0 and VERSION_VALUE_3_LAST mean anyway (making next_major a bit unnecessary), but it works in my opinion. I'll test it in practice tomorrow.

I convinced myself to simplify things a bit, so we won't need VERSION_VALUE_3_LAST and use the (flip_over_version - 1). With that you will only need to redefine VERSION_VALUE_4_0 and and FEATURE_TYPING_MIN_VERSION both to 0x0800

lib/tests/test_lexer.c
```diff
@@ -339,6 +345,38 @@ foo\n");
...
+  cfg_set_version_without_validation(configuration, 0);
+
+  gint flip_version = FEATURE_TYPING_MIN_VERSION - 1;
+  g_snprintf(buf, sizeof(buf), "@version: %d.%d\nbar\n",
+             (flip_version & 0xFF00) >> 8,
+             flip_version & 0xFF);
+  _input(buf);
+  assert_parser_identifier("bar");
+  cr_assert(configuration->user_version == flip_version);
+  assert_grabbed_log_contains("experimental behaviors of the future syslog-ng 4.0");
```
MrAnno, line 364: Please use `VERSION_4_0` instead of hard-coded "syslog-ng 4.0".

lib/tests/test_lexer.c
```diff
@@ -339,6 +345,38 @@ foo\n");
...
+  cr_assert(cfg_is_config_version_older(configuration, VERSION_VALUE_4_0));
+  cr_assert(cfg_is_typing_feature_enabled(configuration));
+
+  reset_grabbed_messages();
+
+  cfg_set_version_without_validation(configuration, 0);
+  _input("@version: 4.0\n\
+bar\n");
+  assert_parser_identifier("bar");
+  assert_grabbed_log_contains("experimental behaviors of the future syslog-ng 4.0");
```
MrAnno, line 375: Please use `VERSION_4_0` instead of hard-coded "syslog-ng 4.0".

lib/tests/test_lexer.c
```diff
@@ -339,6 +345,38 @@ foo\n");
...
+  cr_assert(configuration->user_version == flip_version);
+  assert_grabbed_log_contains("experimental behaviors of the future syslog-ng 4.0");
+
+  cr_assert(cfg_is_config_version_older(configuration, VERSION_VALUE_4_0));
+  cr_assert(cfg_is_typing_feature_enabled(configuration));
+
+  reset_grabbed_messages();
+
+  cfg_set_version_without_validation(configuration, 0);
+  _input("@version: 4.0\n\
```
MrAnno, line 372: This one is also problematic when remapping versions.
Adding a new field for "4.0" (`VERSION_STR_4_0`) to versioning.h sounds like overkill, but can we, for example, just add a macro to the top of this test, so we can patch it internally?

MrAnno:
Internal tests: apart from the above minor things, everything else seems to be OK.

bazsi:
I've resolved all comments with an update to the branch.

kira-syslogng:
Build SUCCESS

MrAnno (approved):
👍🏻 

@bazsi Thank you very much for doing this in a way that helps our internal builds.

Kokan (approved):

Issue #1073, opened 2016-05-31
Enhancement: json-parser and arrays


faxm0dem:
Currently, `json-parser` will parse an array into key-values with an index in the key: `{"foo":["bar","baz"]}` → `{"foo[0]": "bar","foo[1]":"baz"}`. IMHO as long as there is no array support in syslog-ng, it would be saner and more consistent (see `$TAGS`)  to yield a coma-separated list instead: `{"foo":"bar,baz"}`

faxm0dem:
I realize the proposed change would break nested objects like `{"foo":[{"a":"b"},"c"]}`.
I guess there is no good solution as long as syslog-ng doesn't support arrays internally.

faxm0dem:
one way to handle this would be to have `format-json` do the symmetric operation, *i.e.* `{"foo[0]": "bar","foo[1]":"baz"}` → `{"foo":["bar","baz"]}` what do you think?

bazsi:
I am working on list support by using comma separation like in the first
example.

That would make it pretty easy to integrate arrays as a concept. I'll try
to publish that branch in the coming days.

erempel:
While you are in this cde, can you please fix issue #999 for unicode log lines.

du291:
Any progress on this? Are there plans to support JSON arrays (somehow) in syslog-ng?

nzlosh:
Will this issue add support for the _null_ and _object_ data types too?

pradipkumarr:
Is this functionality added, I am facing the same issue while parsing the array using JSON parser, it is like 
`data.minions[0]', value='minion-1', data.minions[1]', value='minion-2', data.minions[2]', value='win-minion108'`
the arraylist is unpredictable, it would be good if it can be like this:
`data.minions, value='minion-1, minion-2, win-minion108'`

gaborznagy:
@pradipkumarr Thank you for your interest!
Arrays as comma-separated list is not supported currently in json-parser (it works as an indexed, separate keys, see description).

I cannot give you any deadline when we will pick up this feature request, but we always welcome and support any community pull request with high priority.

gaborznagy:
Array support is added in #3885. See details in b7364f240296bc14513356099a3d44438df08ede.

Consider the following input:
```
{    
    "array": [ 1, 2, 3 ],    
}
```
An example output with config `json-parser(prefix(".json.")) and $(format-json .json.*)` output BEFORE #3885:
```
{    
    "_json": { 
[...]

PR #3885, opened 2022-01-22
JSON type support (4.0)

bazsi:
This branch has been split off from #3831 and contains the first change on the road to 4.0.

It introduces typing support in json-parser() so that $(format-json) would reproduce the parsed JSON with proper types.

The branch has the following noteworthy changes:
  * value-pairs based destinations and template functions become type aware, e.g. they will get the type information associated with a name-value pairs. The destination/template function may ignore this information, but if it uses it (like $(format-json)), typing would just work (tm)
  * Most uses of value pairs ignore types (e.g. just uses the string representation), in other cases this might generate a compatibility concern (more on that later).
  * embedded JSON arrays are converted to syslog-ng lists
  * top-level JSON arrays are converted to syslog-ng matches ($1, $2, $3, ...)

Compatibility
  * in syslog-ng 3.x mode, value-pairs would default to casting all values to strings (unless there's an explicit type-hint already).
  * in 3.x mode a warning is produced in mongodb() and $(format-json) that informs the user about the change. To suppress the warning, one can explicitly request casting (using --cast/--no-cast).
  * in 4.x mode, the automatic casting is disabled, everything becomes typed
  * The SCL contains some $(format-json) based templates: $(format-ewmm) for instance. I've made these to use explicit casting where I could. Where I couldn't, the same 4.x feature flip would trigger the change to the typed behaviour (--auto-cast) but without triggering the warning.
  * unfortunately template functions are not in a good position to report WHERE in the configuration you can find the problematic template. This should be improved before 4.0 is released.

These call sites should be checked for compatibility/upgrade warnings
- [x] groupset: set operations generally don't need compatibility warnings, as we are doing those on the "get" side, e.g. output, template functions, etc. Nevertheless there's a patch queued to propagate types in groupset(), which is not yet the case in this branch.
- [x] amqp: this is currently ignoring types altogether
- [x] stomp: this is currently ignoring types altogether
- [x] cef: this is currently ignoring types altogether
- [x] graphite: this is currently ignoring types altogether
- [x] welf: this is currently ignoring types altogether
- [x] map-value-pairs: again, provides set functionality. There's a patch to make it type aware.
- [x] python: this is currently ignoring types altogether
- [x] riemann: this is currently ignoring types for value-pairs output, as all attributes supplied this way are strings in riemann. If this were not the case, a warning would be needed (and will be in templates).
- [x] json: needs a warning, DONE!
- [x] mongodb: needs a warning, DONE!

### DIFF
--- a/lib/value-pairs/value-pairs.c
+++ b/lib/value-pairs/value-pairs.c
@@ -896,6 +896,16 @@ value_pairs_new(GlobalConfig *cfg)
   vp->patterns = g_ptr_array_new();
   vp->transforms = g_ptr_array_new();
 
+  if (cfg_is_config_version_older(cfg, VERSION_VALUE_4_0))
+    {
+      /* we don't have an upgrade warning here, as it could only be a very
+       * generic message, not helping our users too much.  I've added this
+       * kind of warning to affected modules instead, like $(format-json)
+       * and mongo-db() */
+
+      vp->cast_to_strings = TRUE;
+    }
+
   return vp;
 }
 
--- a/lib/value-pairs/value-pairs.c
+++ b/lib/value-pairs/value-pairs.c
@@ -287,6 +287,10 @@ vp_msg_nvpairs_foreach(NVHandle handle, const gchar *name,
   sb = scratch_buffers_alloc();
 
   g_string_append_len(sb, value, value_len);
+
+  if (vp->cast_to_strings)
+    type = LM_VT_STRING;
+
   vp_results_insert(results, vp_transform_apply(vp, name), type, sb);
 
   return FALSE;
@@ -395,6 +399,9 @@ vp_merge_builtins(ValuePairs *vp, VPResults *results, LogMessage *msg, LogTempla
         {
           continue;
         }
+
+      if (vp->cast_to_strings)
+        type = LM_VT_STRING;
 
       vp_results_insert(results, vp_transform_apply(vp, spec->name), type, sb);
     }
@@ -854,6 +861,27 @@ value_pairs_add_transforms(ValuePairs *vp, ValuePairsTransformSet *vpts)
 {
   g_ptr_array_add(vp->transforms, vpts);
   vp_update_builtin_list_of_values(vp);
+}
+
+void
+value_pairs_set_cast_to_strings(ValuePairs *vp, gboolean enable)
+{
+  vp->cast_to_strings = enable;
+  vp->explicit_cast_to_strings = TRUE;
+}
+
+void
+value_pairs_set_auto_cast(ValuePairs *vp)
+{
+  /* cast is based on @version, in 3.x mode we use strings, in 4.x we use
+   * types but we won't get the warning */
+  vp->explicit_cast_to_strings = TRUE;
+}
+
+gboolean
+value_pairs_is_cast_to_strings_explicit(ValuePairs *vp)
+{
+  return vp->explicit_cast_to_strings;
 }
 
 ValuePairs *
--- a/lib/value-pairs/value-pairs.c
+++ b/lib/value-pairs/value-pairs.c
@@ -855,7 +855,7 @@ value_pairs_add_transforms(ValuePairs *vp, ValuePairsTransformSet *vpts)
 }
 
 ValuePairs *
-value_pairs_new(void)
+value_pairs_new(GlobalConfig *cfg)
 {
   ValuePairs *vp;
 
@@ -872,7 +872,7 @@ value_pairs_new(void)
 ValuePairs *
 value_pairs_new_default(GlobalConfig *cfg)
 {
-  ValuePairs *vp = value_pairs_new();
+  ValuePairs *vp = value_pairs_new(cfg);
 
   value_pairs_add_scope(vp, "selected-macros");
   value_pairs_add_scope(vp, "nv-pairs");
--- a/lib/value-pairs/value-pairs.c
+++ b/lib/value-pairs/value-pairs.c
@@ -287,7 +287,7 @@ vp_msg_nvpairs_foreach(NVHandle handle, const gchar *name,
   sb = scratch_buffers_alloc();
 
   g_string_append_len(sb, value, value_len);
-  vp_results_insert(results, vp_transform_apply(vp, name), LM_VT_STRING, sb);
+  vp_results_insert(results, vp_transform_apply(vp, name), type, sb);
 
   return FALSE;
 }
@@ -368,6 +368,7 @@ vp_merge_builtins(ValuePairs *vp, VPResults *results, LogMessage *msg, LogTempla
   for (i = 0; i < vp->builtins->len; i++)
     {
       ValuePairSpec *spec = (ValuePairSpec *) g_ptr_array_index(vp->builtins, i);
+      LogMessageValueType type;
 
       sb = scratch_buffers_alloc();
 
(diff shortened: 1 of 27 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/json/json-parser.c
```diff
@@ -144,25 +176,7 @@ json_parser_process_single(struct json_object *jso,
...
-       * (e.g.  "NULL" as a value), however if we encounter a null at a
-       * place where we might also have a sub-object, and we explicitly
-       * reference a field within that sub-object, we would get an empty
-       * string in that case too (due to the key being unknown).  All in
-       * all, the base case is I think to use an empty string as value, and
-       * LM_VT_NULL as type.  Type aware consumers would ignore the string
-       * anyway.
-       */
-
       parsed = TRUE;
```
MrAnno: In e23ddc690751da33f13fa19edf54a7436865b10a, the extraction seems to be incomplete as this line remained here.

modules/json/json-parser.c
```diff
@@ -160,24 +161,33 @@ json_parser_process_single(struct json_object *jso,
...
+            {
+              /* unknown type, encode the entire array as JSON */
+              g_string_assign(value, json_object_to_json_string_ext(jso, JSON_C_TO_STRING_PLAIN));
+              type = LM_VT_JSON;
+              break;
+            }
         }
-      break;
-    }
+
```
MrAnno, line 209: This resolves #1073 :)

I could not find this documented anywhere, I think we can change this without providing a compat layer.

modules/json/json-parser.c
```diff
@@ -251,6 +251,33 @@ json_parser_process_object(struct json_object *jso,
...
+      LogMessageValueType element_type;
+
+      if (json_parser_extract_string_from_simple_json_object(el, element_value, &element_type))
+        {
+          log_msg_set_match_with_type(msg, i + 1, element_value->str, element_value->len, element_type);
+        }
+      else
+        {
+          /* unknown type, encode the entire value as JSON */
+          log_msg_set_match_with_type(msg, i + 1, json_object_to_json_string_ext(el, JSON_C_TO_STRING_PLAIN), -1, LM_VT_JSON);
```
MrAnno, line 276: Do we want to start the index from 1 instead of 0?

modules/json/json-parser.c
```diff
@@ -265,6 +292,11 @@ json_parser_extract(JSONParser *self, struct json_object *jso, LogMessage *msg)
       json_parser_process_object(jso, self->prefix, msg);
       return TRUE;
     }
+  if (json_object_is_type(jso, json_type_array))
+    {
+      json_parser_process_array(jso, self->prefix, msg);
```
MrAnno, line 297: Adding top-level JSON array support in a way that we parse them into `$1, $2, $3` sounds good, but this implementation suggests top-level arrays are not really supported out-of-the-box, as matches are special nv-pairs that, for example, are not transmitted over network when using a basic $(format-json) template. `$1, $2, $3` sounds like we want to support this only in SCL blocks where these can be processed further.

I'm just thinking, but wouldn't it be better to do the same with the top-level arrays as we do with embedded arrays (creating a $(list))? We could use `prefix()` as the name of the list.
If `prefix` is not set, we could still fall back to `$1, $2...` or just use a special hard-coded name for the list.

modules/json/format-json.c
```diff
@@ -83,6 +83,23 @@ tf_json_prepare(LogTemplateFunction *self, gpointer s, LogTemplate *parent,
       value_pairs_add_transforms(state->vp, vpts);
     }
 
+  if (cfg_is_config_version_older(parent->cfg, VERSION_VALUE_4_0) &&
```
MrAnno, line 86: Shouldn't we use `FEATURE_TYPING_MIN_VERSION` here?

modules/json/json-parser.c
MrAnno: This is actually really exciting because $(format-json) as well supports lists `LM_VT_LIST`, so we produce almost identical results when applying json-parser and then format-json (except for embedded objects).

modules/json/json-parser.c
bazsi: thanks, fixed this up.

modules/json/json-parser.c
bazsi: It even works with embedded objects, as in that case json-parser() recognizes there are non-trivial-typed elements in the array, so we fall back to using LM_VT_JSON for the entire array, which will be inserted into the JSON verbatim.

Indeed it does resolve #1073.

modules/json/json-parser.c
bazsi: I agree that just putting these into matches is not much, however mapping this list to $prefix does not really help much either.

My solution was to implement `$*` as a macro in a followup patch that produces a list from the numbered matches, which makes it much easier to process this further. I don't think the array form is to be considered a final representation of the message, thus it shouldn't be added as a "real" name-value pair.

This is the patch that implements `$*`: https://github.com/syslog-ng/syslog-ng/pull/3831/commits/2131b60f60bef4294be3539bde6685286071320b

We could discuss this IRL perhaps.

modules/json/json-parser.c
bazsi: I wouldn't $0 is always the entire string when doing regexps. With bash arguments $0 is the command itself and $1 is the first argument. $* does not contain $0.

modules/json/json-parser.c
MrAnno: Thank you, it would be great to discuss this. 
We had a few thoughts on this topic at the weekly OSE meeting.

lib/value-pairs/value-pairs.c
```diff
@@ -896,6 +896,16 @@ value_pairs_new(GlobalConfig *cfg)
   vp->patterns = g_ptr_array_new();
   vp->transforms = g_ptr_array_new();
 
+  if (cfg_is_config_version_older(cfg, VERSION_VALUE_4_0))
```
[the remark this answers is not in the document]
bazsi: We might do that just to show the intent here, but my understanding is that we don't have to. VERSION_VALUE_4_0 would be mapped to 0x0800 in the case of PE.

But you may be right this would be just make things more difficult to understand.

modules/json/format-json.c
bazsi: the same applies as with the value-pairs case and I guess the afmongodb case is again the same.

lib/value-pairs/value-pairs.c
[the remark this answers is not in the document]
MrAnno: Yeah, it works both ways. I was just wondering, but I'm fine with `VERSION_VALUE_4_0` as well. :)

kira-syslogng:
Build SUCCESS

gaborznagy:
@bazsi I was starting to ignore `centos-7` image builds as they more and more tend to fail due to network issues.
Now `centos-7` build constantly fail, but it seems that `https://copr.fedoraproject.org`'s certificate expired.

modules/json/json-parser.c
MrAnno: We've discussed this IRL on the weekly OSE meeting, it was a long and complicated topic with lots of pros and cons.
We agreed to go with the "$1, $2, ..." method with top-level arrays for now because there were multiple issues with the out-of-the box support "nameless", "top-level" entities in syslog-ng (format-json, value-pairs, etc.).

This special case does not break anything in v3.xx, only adds experimental support for top-level arrays. We can change the details of it until 4.0.

MrAnno (approved):

gaborznagy (approved):

PR #3886, opened 2022-01-23
Propagate types over set operations (4.0)

bazsi:
This branch adds support for the propagation of typing in map-value-pairs() and groupset() operations and depends on #3885.

These are relatively simple changes that only depend on value-pairs handing over type information to them. There's not much compatibility concern here, as value-pairs does the right  thing (e.g. casts everything to strings in 3.x mode).

Even if this wasn't the case, all these changes do is *set* new name-value pairs and if they wouldn't become strings, that type information should be ignored in 3.x mode anyway.

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -1392,7 +1392,7 @@ matcher_flags
 value_pair_option
 	: KW_VALUE_PAIRS
           {
-            last_value_pairs = value_pairs_new();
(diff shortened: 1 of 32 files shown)


kira-syslogng:
Build FAILURE

bazsi:
Closing this as this one unfortunately depends on some pieces later in my mega-branch, I'll reopen this later.

kira-syslogng:
Build SUCCESS

PR #3892, opened 2022-01-27
logproto: fix invalid buffer handling when encoding() is used

MrAnno:
`log_proto_text_server_split_buffer()` moves data within `self->data` for which we maintain additional pointers: `buffer_start`.

This pointer needs to be updated after the `memmove()` call, otherwise it could point to partially overwritten data in case the `memmove()` operation overlaps.

A few lines below my fix, `buffer_start` is passed to a decoder function called `log_proto_text_server_get_raw_size_of_buffer()`, which could fail and report incorrect positions.

This could result in message duplication, partial message loss, but it only occurs when non-fixed length or less known fixed-length encodings are used (encodings that are not listed in `fixed_encodings`).

### DIFF
--- a/lib/logproto/logproto-buffered-server.c
+++ b/lib/logproto/logproto-buffered-server.c
@@ -594,6 +594,116 @@ log_proto_buffered_server_read_data_method(LogProtoBufferedServer *self, guchar 
   return log_transport_read(self->super.transport, buf, len, aux);
 }
 
+static void
+log_proto_buffered_server_maybe_realloc_reverse_buffer(LogProtoBufferedServer *self, gsize buffer_length)
+{
+  if (self->reverse_buffer_len >= buffer_length)
+    return;
+
+  /* we free and malloc, since we never need the data still in reverse buffer */
+  g_free(self->reverse_buffer);
+  self->reverse_buffer_len = buffer_length;
+  self->reverse_buffer = g_malloc(buffer_length);
+}
+
+/*
+ * returns the number of bytes that represent the UTF8 encoding buffer
+ * in the original encoding that the user specified.
+ *
+ * NOTE: this is slow, but we only call this for the remainder of our
+ * buffer (e.g. the partial line at the end of our last chunk of read
+ * data). Also, this is only invoked if the file uses an encoding.
+ */
+static gsize
+log_proto_buffered_server_get_raw_size_of_buffer(LogProtoBufferedServer *self, const guchar *buffer, gsize buffer_len)
+{
+  gchar *out;
+  const guchar *in;
+  gsize avail_out, avail_in;
+  gint ret;
+
+  if (self->reverse_convert == ((GIConv) -1) && !self->convert_scale)
+    {
+      /* try to speed up raw size calculation by recognizing the most
+       * prominent character encodings and in the case the encoding
+       * uses fixed size characters set that in self->convert_scale,
+       * which in turn will speed up the reversal of the UTF8 buffer
+       * size to raw buffer sizes.
+       */
+      self->convert_scale = log_proto_get_char_size_for_fixed_encoding(self->super.options->encoding);
+      if (self->convert_scale == 0)
+        {
+          /* this encoding is not known, do the conversion for real :( */
+          self->reverse_convert = g_iconv_open(self->super.options->encoding, "utf-8");
+        }
+    }
+
+  if (self->convert_scale)
+    return g_utf8_strlen((gchar *) buffer, buffer_len) * self->convert_scale;
+
+
+  /* Multiplied by 6, because 1 character can be maximum 6 bytes in UTF-8 encoding */
+  log_proto_buffered_server_maybe_realloc_reverse_buffer(self, buffer_len * 6);
+
+  avail_out = self->reverse_buffer_len;
+  out = self->reverse_buffer;
+
+  avail_in = buffer_len;
+  in = buffer;
+
+  ret = g_iconv(self->reverse_convert, (gchar **) &in, &avail_in, &out, &avail_out);
+  if (ret == (gsize) -1)
+    {
+      /* oops, we cannot reverse that we ourselves converted to UTF-8,
+       * this is simply impossible, but never say never */
+      msg_error("Internal error, couldn't reverse the internal UTF8 string to the original encoding",
+                evt_tag_printf("buffer", "%.*s", (gint) buffer_len, buffer));
+      return 0;
+    }
+  else
+    {
+      return self->reverse_buffer_len - avail_out;
+    }
+}
+
+void
+log_proto_buffered_server_split_buffer(LogProtoBufferedServer *self, LogProtoBufferedServerState *state,
+                                       const guchar *buffer_start, gsize buffer_bytes)
+{
+  gsize raw_split_size;
+
+  /* buffer is not full, but no EOL is present, move partial line
+   * to the beginning of the buffer to make space for new data.
+   */
+
+  memmove(self->buffer, buffer_start, buffer_bytes);
+  state->pending_buffer_pos = 0;
+  state->pending_buffer_end = buffer_bytes;
+  buffer_start = self->buffer;
+
+  if (G_UNLIKELY(self->pos_tracking))
+    {
+      /* NOTE: we modify the current file position _after_ updating
+         buffer_pos, since if we crash right here, at least we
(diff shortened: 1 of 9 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:
Build SUCCESS

kira-syslogng:

MrAnno:
@Kokan I need to validate my theory, but this might be one of the root causes of the Big5 encoding issue that we've discussed IRL.

news/bugfix-3892.md
```diff
@@ -0,0 +1,5 @@
+`file()` source: fixed invalid buffer handling when `encoding()` is used
+
+A bug has been fixed that - under rare circumstances - could cause message
+duplication, partial message loss, even crashes when non-fixed length or less
```
bazsi, line 4: do we have security implications here?

MrAnno: I'm still investigating the crash part of it, but AFAIK this should have no security implications, because we always stay within the buffer's boundaries, we just mess up the positions within those boundaries.

MrAnno: The crash I reproduced was a bug in my tests. I'm removing the crash part from the news entry.

lib/logproto/logproto-buffered-server.c
```diff
@@ -624,7 +624,7 @@ log_proto_buffered_server_fetch_from_buffer(LogProtoBufferedServer *self, const
       goto exit;
     }
 
-  success = self->fetch_from_buffer(self, buffer_start, buffer_bytes, msg, msg_len);
+  success = self->fetch_from_buffer(self, &buffer_start, buffer_bytes, msg, msg_len);
```
bazsi, line 627: why do we need to propagate the changes to buffer_start back here? This is a local variable which is ignored after fetch_from_buffer() returns. I am still reading the code to see what I am missing here.

lib/logproto/logproto-text-server.c
```diff
@@ -199,7 +200,7 @@ log_proto_text_server_split_buffer(LogProtoTextServer *self, LogProtoBufferedSer
 
 
       if (self->super.super.options->encoding)
-        raw_split_size = log_proto_text_server_get_raw_size_of_buffer(self, buffer_start, buffer_bytes);
+        raw_split_size = log_proto_text_server_get_raw_size_of_buffer(self, *buffer_start, buffer_bytes);
```
bazsi, line 203: If I understand things right, then this is the only problem that we are fixing. log_proto_text_server_get_raw_size_of_buffer() needs the new buffer_start (e.g. the beginning of self->super.buffer) instead of the buffer we got as parameter. Is that right?

If so, then I don't think we need to change the argument to be a pointer to a pointer, the fix is simply using self->super.buffer as the 2nd argument to log_proto_text_server_get_raw_size_of_buffer()

bazsi (commented):
The bug itself is a good catch, must have been very difficult to reproduce and fix it. However I don't think we need to change the argument list of fetch_from_buffer().

lib/logproto/logproto-text-server.c
MrAnno: Yes, that is the only problematic misuse currently. If we don't change the signature, I'll have to add a comment to buffered server and to text server to indicate that the local variable should not be used below the problematic function call.

lib/logproto/logproto-buffered-server.c
MrAnno: I've removed this part.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:
Build SUCCESS

kira-syslogng:

lib/logproto/logproto-text-server.c
```diff
@@ -384,7 +218,7 @@ log_proto_text_server_fetch_from_buffer(LogProtoBufferedServer *s, const guchar
         }
       else
         {
-          log_proto_text_server_split_buffer(self, state, buffer_start, buffer_bytes);
+          log_proto_buffered_server_split_buffer(&self->super, state, buffer_start, buffer_bytes);
           goto exit;
```
bazsi, line 222: Thanks, the last changes look good.

One more idea: what if we pushed the the invocation had 3 different returns from "fetch_from_buffer"
* We have a line (current TRUE value)
* we don't have a line but the buffer is full and cannot be split
* We don't have a line but can split.
* 
Now as I think of it the last two cases don't even need separate returns. The decision whether we can split can be decided at the buffered-server level.

So maybe you could copy the remainder of the text_server_fetch_from_buf() to the call site.

Just an idea.

MrAnno: I'll try that, thanks.

One thing I noticed is that we currently split the buffer too often: for example, when a message is received from N partial read operations, each attempt will trigger a `memmove()` operation. I believe we only need to split our buffer when we're about to reach the end of it.

If I'm right about this, I could easily remove the split logic from the text server completely (moving it to the buffered server), and we may also increase performance a bit.

MrAnno: Done.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng do stresstest

lib/logproto/logproto-text-server.c
MrAnno: Hmm, I might be wrong about the performance impact of my last commit, because I've introduced more `read()` and `iconv()` calls on smaller chunks of data to save a few `memmove()`s. I'm dropping that commit.

kira-syslogng:
Build SUCCESS

kira-syslogng:

MrAnno:
@kira-syslogng do stresstest

lib/logproto/logproto-text-server.c
```diff
@@ -381,30 +203,37 @@ log_proto_text_server_fetch_from_buffer(LogProtoBufferedServer *s, const guchar
...
-          goto exit;
-        }
+      log_proto_text_server_yield_whole_buffer_as_message(self, state, buffer_start, buffer_bytes, msg, msg_len);
+      goto success;
     }
 
+  return FALSE;
+
+success:
   log_proto_text_server_remove_trailing_newline(msg, msg_len);
```
bazsi, line 224: looks great so far.

if you pushed this call back to the caller that would still simplify this function by removing all the goto statements here.

MrAnno: I'd leave it as is if that's okay with you. I prefer `_fetch_msg_from_buffer()` to produce a "cleaned-up" message without the new lines.

kira-syslogng:
Build SUCCESS

lib/logproto/logproto-text-server.c
bazsi: Sure.

MrAnno: Thank you.

kira-syslogng:

bazsi (approved):

Kokan (approved):

MrAnno:
We're running some additional tests before merge.

lib/logproto/tests/test-text-server.c
```diff
@@ -589,3 +591,33 @@ Test(log_proto, test_log_proto_text_server_io_eagain)
...
 }
+
+Test(log_proto, buffer_split_with_encoding_and_position_tracking)
+{
+  GString *data = g_string_new("Lorem ipsum\xe2\x98\x83lor sit amet, consectetur adipiscing elit\n");
+  GString *data_smaller = g_string_new("muspi merol\n");
+  gsize bytes_should_not_fit = 5;
+
+  proto_server_options.max_msg_size = data->len + data_smaller->len - bytes_should_not_fit;
+  proto_server_options.max_buffer_size = proto_server_options.max_buffer_size;
```
gaborznagy, line 602: I think this is a typo in the initialization: `max_buffer_size = max_buffer_size` is a self-assignment.

MrAnno: Nice catch, thanks. I'll fix it.

kira-syslogng:
Build SUCCESS

gaborznagy (approved):

bazsi (approved):

PR #3893, opened 2022-01-28
Accept iso timestamp with a space instead of t

bazsi:
This patch implements accepting ISO8601 timestamps with a space instead of a 'T' in the middle, as this format happens a lot in the wild.

For instance ZScaler produces this syslog header:
```
<134>2022-01-27 10:32:35      reason=Allowed... # message continues
```

The patch causes syslog-ng to accept this timestamp both in RFC5424 and RFC3164 mode. I've decided this way as even if we permitted the incorrect formatting only in RFC3164 mode, I expect the same to happen in 5424 implementations too. And I know it's against the spec, but this seems to be an uphill battle.

### DIFF
--- a/lib/timeutils/scan-timestamp.c
+++ b/lib/timeutils/scan-timestamp.c
@@ -193,7 +193,7 @@ __is_iso_stamp(const gchar *stamp, gint length)
   return (length >= 19
           && stamp[4] == '-'
           && stamp[7] == '-'
-          && stamp[10] == 'T'
+          && (stamp[10] == 'T' || stamp[10] == ' ')
           && stamp[13] == ':'
           && stamp[16] == ':'
          );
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

MrAnno:
LGTM.
The macOS job failed on one of the unit tests:
```
[----] tests/unit/test_msgparse.c:163: Assertion failed: Unexpected timestamp, value=1162083600, expected=1162080000, msg=<7>2006-10-29 02:00:00
```

Please add a news entry.

bazsi:
> ```
> [----] tests/unit/test_msgparse.c:163: Assertion failed: Unexpected timestamp, value=1162083600, expected=1162080000, msg=<7>2006-10-29 02:00:00
> ```

Probably MacOS has a different timezone and that test does not mock the timezone setting. I think the other testcase covers the parsing piece, so I've removed this failing testcase from test_msgparse. We still have one but that contains timezone
information and thus we validate that timestamps without 'T" are parsed end-to-end via the normal parser.

That test_msgparse unit test is a dire need of a refactor...

> 
> Please add a news entry.

I was about to. :)

MrAnno:
> That test_msgparse unit test is a dire need of a refactor...

Yeah, that one is terrible. Sorry about that, I was lazy when I ported this test to Criterion :(

MrAnno (approved):

kira-syslogng:
Build SUCCESS

alltilla (approved):

PR #3895, opened 2022-01-29
test_matcher: avoid using complete syslog messages as test input

bazsi:
Since we are only testing logmatcher, we don't need to use the entire syslog message as input in these testcases. Make
things simpler by using log_msg_set_value() instead of running it through syslogformat to parse it.

Originally part of my typing work, but completely independent, so by opening a separate PR I hope we can merge it sooner.

### DIFF
--- a/tests/unit/test_matcher.c
+++ b/tests/unit/test_matcher.c
@@ -31,8 +31,6 @@
 #include <stdlib.h>
 #include <string.h>
 
-MsgFormatOptions parse_options;
-
 static LogMessage *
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build SUCCESS

MrAnno (approved):

MrAnno:
This PR simplifies a unit test. As the change can be considered trivial and only a single test file was changed, I think we can make an exception and merge it with 1 approval.

MrAnno (approved):

PR #3896, opened 2022-01-29
remove module-path and autoload-compiled-modules global defines (4.0)

bazsi:
This branch introduces a significant simplification of the startup process by eliminating a couple of global `@defines` in the configuration. At the same time it lays the groundwork for a more elaborate expression language in filter comparisons.

Let's go one by one:
  * in a way this is part of the 4.0 patches, however can be merged earlier, as it is based on master and not on any other part of the typing branch
  * this branch gets rid of two defines: `module-path` (replaced by a similarly named command line option), and also `autoload-compiled-modules` (replaced by --no-module-discovery command line option)
  * by getting rid of these, we can simplify startup a lot. Previously even the lexer was involved in deciding when to start plugin discovery.
  * also, this enables the use of a new token LL_PLUGIN for plugin insertion points, instead of the generic LL_IDENTIFIER that we used. This allowed me to greatly extend the filter comparison grammar.

Compatibility concerns:
  * this patch introduces an incompatible change, but I have never seen the use of `module-path` or `autoload-compiled-modules`. I even tried to google them for 10 minutes.
  * These options are documented though.

Other than these concerns, My opinion is that this could be pulled in to a regular 3.x release as well.

### DIFF
--- a/lib/block-ref-grammar.ym
+++ b/lib/block-ref-grammar.ym
@@ -55,15 +55,7 @@ block_args
         ;
 
 block_arg
-        : LL_IDENTIFIER
-          {
-            cfg_lexer_push_context(lexer, LL_CONTEXT_BLOCK_ARG, NULL, "block argument");
-          }
-          LL_BLOCK
-          {
-            cfg_lexer_pop_context(lexer);
-            cfg_args_set(*result, $1, $3); free($1); free($3);
-          }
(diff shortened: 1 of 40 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

Kokan:
This looks good to me, I want to do some internal testing and follow ups. When it is done I'll approve.

Kokan (approved):

alltilla (approved):

PR #3898, opened 2022-01-31
dbld: use separate dirs in extra-files for each platform

alltilla:
When we are building multiple images simultaneously, it is possible that `dbld/extra-files` is modified by multiple image build tasks.

Although we only copy files there, in rare scenarios it can cause a race condition.


MrAnno (approved):

kira-syslogng:
Build SUCCESS

OverOrion (approved):

PR #3899, opened 2022-01-31
Create MQTT tests in light

Barkodcz:
This PR introduced the MQTT test in light.
To test MQTT we need also need to configure the broker. This option is enabled with MQTTConfig class.

In the future, I  plan to create more tests. But for now, I only make 1 source test and 1 destination test.

Also, have some TODO comments in the code, I plan to code that piece of code when I needed that function (for example TLS)

There is a commit (commit message: `ci: SHOULD BE REMOVE!!! ...`) this commit should remove before the merge. It needs because I get some new dependencies.


kira-syslogng:
Build SUCCESS

tests/light/src/syslog_ng_config/statements/sources/mqtt_source.py
```diff
@@ -0,0 +1,42 @@
...
+# COPYING for details.
+#
+#############################################################################
+import paho.mqtt.publish as publish
+
+from src.syslog_ng_config.statements.sources.source_driver import SourceDriver
+
+
+class MQTTSource(SourceDriver):
+    def __init__(self, mqtt_config, **options):
```
mitzkia, line 29: I need some time to think over this class too

tests/light/src/driver_io/mqtt_broker/mqtt_broker.py
```diff
@@ -0,0 +1,67 @@
...
+
+        f = File(path)
+        with f.open('w+') as config_file:
+            config_file.write(self.config.generateMQTTConfig())
+
+        print(path.resolve())
+        return path.resolve()
+
+    def start(self):
+        mqtt_broker_stdout_path = Path(tc_parameters.WORKING_DIR, "mqtt_broker_stdout")
```
mitzkia, line 48: at the end of the `start()` please return with the started process object (you can find examples in other places)

tests/light/src/driver_io/mqtt_broker/mqtt_config.py
```diff
@@ -0,0 +1,175 @@
...
+#
+#############################################################################
+from pathlib2 import Path
+
+import src.testcase_parameters.testcase_parameters as tc_parameters
+from src.executors.command_executor import CommandExecutor
+
+
+class MQTTConfig():
+
```
mitzkia, line 30: I need some time to think over this class

tests/light/src/syslog_ng_config/statements/destinations/mqtt_destination.py
```diff
@@ -0,0 +1,52 @@
...
+# COPYING for details.
+#
+#############################################################################
+import paho.mqtt.subscribe as subscribe
+
+from src.syslog_ng_config.statements.destinations.destination_driver import DestinationDriver
+
+
+class MQTTDestination(DestinationDriver):
+    def __init__(self, mqtt_config, **options):
```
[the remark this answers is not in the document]
mitzkia: My idea is to separate syslog-ng related config options and Mqtt broker related config options. 
It would be good if we can follow an IO class separation pattern of this logic, as it  used in `FileIO()` or `NetworkIO()`
What do you think about that?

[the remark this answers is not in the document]
Barkodcz: I understand your point!
While I developed this feature, I felt this isn't the best solution (1 object response for 2 things - that break the SOLID). The reason to do that, is I don't want to make two separate objects to configure the MQTT. But I think it would be greater to make two separate objects. 
I'm trying to get any good idea to "defend" the current solutions, but I don't find any :( (probably because this isn't the best solution) 
I will check the idea that you described. As soon as I can, I will process every review comment

mitzkia:
In next week we will concluded how we want to long term support 3rd party dependencies in Light, in our case MQTT service.
After that discussion I will share the results in this PR (of course I can help in the further implementations if you needed).

MrAnno:
I far as I can see this has been stuck for half a year.
Does anyone have plans to move this PR forward?

mitzkia:
I would push this PR forward. First of all we need to discuss how we can long term support 3rd party dependencies in Light . 
@Kokan came up with this question before. Peter could you please explain your thoughts in this topic?

MrAnno:
@mitzkia Let's discuss this IRL :)

PR #3900, opened 2022-01-31
Fix SIGSEGV in the "CONFIG" command over syslog-ng-ctl

bazsi:
Fix a SIGSEGV triggered by an incorrectly formatted "CONFIG"
command, received on the syslog-ng control socket.  The only known
implementation of the control protocol is syslog-ng-ctl itself, which always
sends a correct command, but anyone with access to the UNIX domain socket
`syslog-ng.ctl` (root only by default) can trigger a crash.

### DIFF
--- a/lib/control/tests/test_control_cmds.c
+++ b/lib/control/tests/test_control_cmds.c
@@ -117,7 +117,7 @@ Test(control_cmds, test_log)
   const gchar *response;
 
   _run_command("LOG", &response);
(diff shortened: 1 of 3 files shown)


MrAnno (dismissed):
👍🏻

Kokan:
Hmm, it seems that some UT assert on the error message that was changed.

Kokan (approved):

kira-syslogng:
Build SUCCESS

MrAnno (approved):

PR #3902, opened 2022-02-04
package-indexer: add script for indexing packages

alltilla:
This script can be used to index the `deb` packages generated by `dbld`.

With this PR we only add the code and some static analysis workflows in GitHub Actions, but not the actual indexing workflow.


kira-syslogng:
Build SUCCESS

MrAnno (approved):

gaborznagy:
@alltilla: can you add the packaging/package-indexer dir to the dist tarball, please?

Sorry, I just spotted this now.

alltilla:
Sure I can, but do we want it in the tarball, though?

MrAnno:
The GitHub actions jobs are not in the tarball either.
I thought this was intentional. My reasoning is that these are not useful inside the syslog-ng tarball.

GitHub Actions configs are used for package generation and testing in a dev environment that requires git anyway.
I think the same applies to the package indexer code: it can be used to publish syslog-ng packages in an APT repository, which is specific to our use-case with ose-repo.syslog-ng.com (azure/DO backend, nightly/stable, etc.).

gaborznagy:
> The GitHub actions jobs are not in the tarball either. I thought this was intentional. My reasoning is that these are not useful inside the syslog-ng tarball.
> 
> GitHub Actions configs are used for package generation and testing in a dev environment that requires git anyway. I think the same applies to the package indexer code: it can be used to publish syslog-ng packages in an APT repository, but it is specific to our use-case with ose-repo.syslog-ng.com (azure/DO backend, nightly/stable, etc.).

Thank you for both of you for the reasoning. I'm convinced.
To be honest, I just thought that we should include scripts by default, I didn't think more into it, my bad.

gaborznagy (approved):

PR #3903, opened 2022-02-07
WIP: GitHubActions: Reusable tests for APT packages

mitzkia:
This PR is about to add some basic GitHubActions checks for APT repo packages.

The solution uses a reusable workflow concept. Which means there is a "trigger" job which can start the main workflow with different inputs. With this solution we can reduce the boilerplate codes in workflows.
Our inputs are: 
- "pkg_type" which refers for the installer type: nightly or stable
- "revision_name" which refers for the installer revision: snapshot or stable

Current checks on installed packages:
- is the installed syslog-ng revision correct related to actual package type
- are the installed syslog-ng package versions correct related to installed revision
- can syslog-ng starts with the default config

One minor question is remaining to finish this PR, when should the jobs triggered?
For now I set it to "push" event, but it is ineffective. I suggest to use "schedule" event, but when it should be started?

Example successful run:
https://github.com/mitzkia/syslog-ng/actions/runs/1806191556

Example failed run 1:
https://github.com/mitzkia/syslog-ng/actions/runs/1807409122

Example failed run 2:
https://github.com/mitzkia/syslog-ng/actions/runs/1807428615

### DIFF
--- a/.github/workflows/test-apt-packages.yml
+++ b/.github/workflows/test-apt-packages.yml
@@ -1,0 +1,78 @@
+name: Run smoke tests on APT packages
+
+on:
+  workflow_call:
+    inputs:
+      pkg_type:
+        required: true
+        type: string
+
+jobs:
+  TestPackagesWith:
+    strategy:
+      matrix:
+        distro:
+          - "debian:stretch"
+          - "debian:buster"
+          - "debian:bullseye"
+          - "debian:testing"
+          - "debian:sid"
+          - "ubuntu:xenial"
+          - "ubuntu:bionic"
+          - "ubuntu:focal"
+          - "ubuntu:impish"
+      fail-fast: false
+    runs-on: ubuntu-latest
+    container: ${{ matrix.distro }}
+    steps:
+      - name: Install prerequisites for adding 3rd party repository
+        run: |
+          apt-get update -qq
+          apt-get install --yes wget gnupg2 ca-certificates apt-transport-https
+
+      - name: Add nightly OSE repository
+        if: inputs.pkg_type == 'nightly'
+        run: |
+          wget -qO - https://ose-repo.syslog-ng.com/apt/syslog-ng-ose-pub.asc | gpg --dearmor > /usr/share/keyrings/ose-repo-archive-keyring.gpg
+          echo "deb [trusted=yes] https://ose-repo.syslog-ng.com/apt/ ${{ inputs.pkg_type }} $(echo ${{ matrix.distro }} | sed 's/:/-/g')" | tee --append /etc/apt/sources.list.d/syslog-ng-ose.list
+
+      - name: Add stable OSE repository
+        if: inputs.pkg_type == 'stable'
+        run: |
+          wget -qO - https://ose-repo.syslog-ng.com/apt/syslog-ng-ose-pub.asc | gpg --dearmor > /usr/share/keyrings/ose-repo-archive-keyring.gpg
+          echo "deb [signed-by=/usr/share/keyrings/ose-repo-archive-keyring.gpg] https://ose-repo.syslog-ng.com/apt/ ${{ inputs.pkg_type }} $(echo ${{ matrix.distro }} | sed 's/:/-/g')" | tee --append /etc/apt/sources.list.d/syslog-ng-ose.list
+
+      - name: Install the last but one syslog-ng OSE package
+        run: |
+          apt-get update -qq
+          DEBIAN_FRONTEND=noninteractive apt-get install --yes syslog-ng=$(apt-cache madison syslog-ng | sed -n 2p | awk -F"|" '{print $2}' | sed 's/ //g')
+
+      - name: Upgrade to the latest syslog-ng OSE package
+        run: |
+          DEBIAN_FRONTEND=noninteractive apt-get install --yes syslog-ng
+
+      - name: Get syslog-ng revision
+        run: |
+          syslog-ng -V
+          echo "::set-output name=REVISION::$(syslog-ng -V | grep Revision | cut -d" " -f2)"
+        id: syslog_ng_revision
+
+      - name: Check revision
+        run: |
+          if [ ${{ inputs.pkg_type }} = "nightly" ]; then
+            echo "${{ steps.syslog_ng_revision.outputs.REVISION }}" | egrep "^[0-9]{1}.[0-9]{1,2}.[0-9]{1}.[0-9]{3}.[a-z0-9]{8}-snapshot\+[0-9]{8}T[0-9]{6}$";
+          elif [ ${{ inputs.pkg_type }} = "stable" ]; then
+            echo "${{ steps.syslog_ng_revision.outputs.REVISION }}" | egrep "^[0-9]{1}.[0-9]{1,2}.[0-9]{1}-[0-9]{1}$"
+          fi
+
+      - name: Check if installed package version matches with install revision
+        run: |
+          echo "Installed revision value: ${{ steps.syslog_ng_revision.outputs.REVISION }}"
+          dpkg-query --show | grep syslog-ng
+          dpkg-query --show | grep syslog-ng | while read installed_syslog_ng_package ; do echo $installed_syslog_ng_package | grep ${{ steps.syslog_ng_revision.outputs.REVISION }} ; done
+
+      - name: Check if syslog-ng can start with default config
+        run: |
+          nohup syslog-ng -Fe &
+          sleep 5
+          syslog-ng-ctl stop
--- a/.github/workflows/trigger-tests-on-nightly-packages.yml
+++ b/.github/workflows/trigger-tests-on-nightly-packages.yml
@@ -1,0 +1,14 @@
+name: Trigger tests for nightly APT packages
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '00 3 * * *'
+
+
+jobs:
+  test_nightly_packages:
+    if: (github.event_name == 'schedule' && github.repository_owner == 'syslog-ng') || (github.event_name != 'schedule')
+    uses: ./.github/workflows/test-apt-packages.yml
+    with:
+      pkg_type: "nightly"
--- a/.github/workflows/trigger-tests-on-stable-packages.yml
+++ b/.github/workflows/trigger-tests-on-stable-packages.yml
@@ -1,0 +1,14 @@
+name: Trigger tests for stable APT packages
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - 'syslog-ng-*'
+
+
+jobs:
+  test_stable_packages:
+    uses: ./.github/workflows/test-apt-packages.yml
+    with:
+      pkg_type: "stable"


kira-syslogng:
Build SUCCESS

mitzkia:
Maybe we can also test an upgrade from previous version in both cases: nightly and stable release.
Should I add this test too?

gaborznagy:
The reusable workflow looks really cool! :)

About when to run this github action: we should discuss this topic on Thursday with the team.
The fastest idea would be to have a cron-scheduled workflow for the nightly packages, and have a possibility to run the workflow manually, but we could find out better solutions too (webhooks, etc.).

Upgrade test sounds good too: does this mean that first it installs the previous nightly/stable version with an explicit version and then does an upgrade with `apt-get upgrade`?

.github/workflows/tests-on-packages.yml
```diff
@@ -0,0 +1,65 @@
...
+          - "ubuntu:impish"
+      fail-fast: false
+    runs-on: ubuntu-latest
+    container: ${{ matrix.distro }}
+    steps:
+      - name: Add ${{ inputs.pkg_type }} OSE repository and install syslog-ng (for Debian based OS)
+        run: |
+          apt-get update -qq
+          apt-get install -y wget gnupg2 ca-certificates apt-transport-https
+          wget -qO - https://ose-repo.syslog-ng.com/apt/syslog-ng-ose-pub.asc | apt-key add -
```
gaborznagy, line 35: `apt-key` is deprecated since Ubuntu 20.10 ( and in Debian Bullseye too I guess) and will be removed:

https://askubuntu.com/questions/1328806/how-to-solve-apt-key-deprecated

This is how we add the OSE APT repo in the syslog-ng` docker image:
```
wget -qO - https://ose-repo.syslog-ng.com/apt/syslog-ng-ose-pub.asc | gpg --dearmor > /usr/share/keyrings/ose-repo-archive-keyring.gpg && \    
  echo "deb [signed-by=/usr/share/keyrings/ose-repo-archive-keyring.gpg] https://ose-repo.syslog-ng.com/apt/ stable debian-testing" | tee --append /etc/apt/sources.list.d/syslog-ng-ose.list    
```

I don't know if this is supported on older Ubuntus and Debians as well.
We should probably change the `README.md` file about the APT repo too.

.github/workflows/tests-on-packages.yml
```diff
@@ -0,0 +1,65 @@
...
+    inputs:
+      pkg_type:
+        required: true
+        type: string
+      revision_name:
+        required: true
+        type: string
+
+jobs:
+  TestPackagesWithLight:
```
mitzkia, line 14: Just a reminder for myself: the name of the job should be simply: TestPackages or something like this.
Just for the note in my first version I have also added to execute Light for the installed syslog-ng (but it was removed, there was some complexity)

.github/workflows/trigger-tests-on-packages.yml
```diff
@@ -0,0 +1,17 @@
+name: Trigger tests for APT packages
+
+on:
+  push:
```
MrAnno, line 4: Let's discuss when and how should these checks run.

I believe `stable` should be started by a release tag trigger, for example:
https://github.com/syslog-ng/syslog-ng/blob/master/.github/workflows/publish-docker-image.yml#L7

But the best would be if we could also test the production docker image itself (before publishing it).

`nightly` can be scheduled, just as you mentioned.

kira-syslogng:
Build FAILURE

.github/workflows/tests-on-packages.yml
mitzkia: Thanks for the comment, I have  fixed this part.

.github/workflows/tests-on-packages.yml
mitzkia: fixed

.github/workflows/trigger-tests-on-packages.yml
mitzkia: Thanks for the review comment. I have fixed the triggering time on new jobs. Let see I they runs correctly.
I am agreed to test our production image. I will think about it,

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

.github/workflows/trigger-tests-on-nightly-packages.yml
```diff
@@ -0,0 +1,12 @@
+name: Trigger tests for nightly APT packages
+
+on:
+  schedule:
+    - cron: '00 3 * * *'
+
+jobs:
+  test_nightly_packages:
+    uses: ./.github/workflows/test-packages.yml
```
MrAnno, line 9: Note: we should review all of our "on schedule" workflows and consider adding `if: github.repository_owner == 'syslog-ng'` in order not to run them in forks.

.github/workflows/trigger-tests-on-stable-packages.yml
```diff
@@ -0,0 +1,13 @@
+name: Trigger tests for stable APT packages
+
+on:
+  push:
+    tags:
+      - 'syslog-ng-*'
```
MrAnno, line 6: Testing the already released version can be useful, but we should probably create a "pre-release"  section next to "stable" and "nightly" so that we can test the release candidate APT packages before the actual release.

kira-syslogng:
Build SUCCESS

.github/workflows/test-apt-packages.yml
```diff
@@ -0,0 +1,82 @@
...
+        if: inputs.revision_name == 'snapshot'
+        run: |
+          echo ${{ steps.syslog_ng_revision.outputs.REVISION }} | grep ${{ inputs.revision_name }}
+
+      - name: Check if revision is stable revision
+        if: inputs.revision_name == 'stable'
+        run: |
+          echo ${{ steps.syslog_ng_revision.outputs.REVISION }} | egrep "^[0-9]{1}.[0-9]{1,2}.[0-9]{1}-[0-9]{1}$"
+
+      - name: Check if installed package version match with install revision
```
OverOrion, line 73:
```suggestion
      - name: Check if installed package version matches with install revision
```

.github/workflows/test-apt-packages.yml
```diff
@@ -0,0 +1,82 @@
+name: Run tests on APT packages
```
OverOrion, line 1: [OPTIONAL]

I think a more accurate name (basic smoke / startup test?) might be better if we are not planning to extend it with more tests.

OverOrion:
Otherwise looks good to me, a really handy feature! :)

.github/workflows/test-packages.yml
```diff
@@ -0,0 +1,82 @@
+name: Run tests on APT packages
```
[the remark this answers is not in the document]
mitzkia: Thanks for the comment. 
Done

.github/workflows/test-apt-packages.yml
mitzkia: Thanks for the review note.
Fixed.

.github/workflows/trigger-tests-on-nightly-packages.yml
mitzkia: Thanks for the review comment, agreed with this.
Fixed

.github/workflows/trigger-tests-on-stable-packages.yml
mitzkia: I am agreed with this too.

kira-syslogng:
Build SUCCESS

OverOrion (dismissed):

MrAnno (commented):
LGTM.
@gaborznagy indicated that he wants to review this, so I've only validated my previous review notes.

Thanks for fixing them.

.github/workflows/trigger-tests-on-stable-packages.yml
MrAnno: I think we won't have the capacity to implement this for this release.
@mitzkia What do you think about adding a line for allowing this job to be run manually, so we can start it after uploading the new packages but before announcing the actual release?

mitzkia: I am also agreed with this method.

kira-syslogng:
Build SUCCESS

.github/workflows/test-apt-packages.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Get syslog-ng revision
+        run: |
+          syslog-ng -V
+          echo "::set-output name=REVISION::$(syslog-ng -V | grep Revision | cut -d" " -f2)"
+        id: syslog_ng_revision
+
+      - name: Check if revision is snapshot revision
+        if: inputs.revision_name == 'snapshot'
+        run: |
+          echo ${{ steps.syslog_ng_revision.outputs.REVISION }} | grep ${{ inputs.revision_name }}
```
gaborznagy, line 66: Do we need this input field? We can use the same `inputs.pkg_type` variable here too. `nightly` would mean check for `"snapshot"`.

Alternatively we can keep the`revision_name` input variable, and use just one revision-checking step, where the input itself is the pattern to check for: e.g. `snapshot` or `^[0-9]{1}.[0-9]{1,2}.[0-9]{1}-[0-9]{1}$`.

mitzkia: Thanks, I like your idea.
What if we drop the "revision_name" input variable and also we will check the revision in one step.
Something like this:
```
      - name: Check revision
        if: inputs.pkg_type == 'nightly'
        run: |
          echo ${{ steps.syslog_ng_revision.outputs.REVISION }} | egrep "^[0-9]{1}.[0-9]{1,2}.[0-9]{1}.[0-9]{3}.[a-z0-9]{8}-snapshot+[0-9]{8}T[0-9]{6}$"
        if: inputs.pkg_type == 'stable'
        run: |
          echo ${{ steps.syslog_ng_revision.outputs.REVISION }} | egrep "^[0-9]{1}.[0-9]{1,2}.[0-9]{1}-[0-9]{1}$"
```

mitzkia: My first idea was to link logically the different (snapshot and stable) revision checks to the steps as it is linked in case of installing different packages.
But this new revision_name input field has almost a redundant meaning with the pkg_type value (which represents how we used them in real).
It would be clearer if they has the same usage in real: in apt source and in installer revision.

gaborznagy: Removing the unnecessary `revision_name` input variable sounds good to me.
I'm not sure that your first example would work with multiple `run` and `if` statements, but you can merge the `run` statements, and you can use an `if` statement on bash level, i.e. inside the `run` statement:
```
run: |
  if [[ ${{ inputs.pkg_type }} == "nightly" ]]; then
    echo "${{ steps.syslog_ng_revision.outputs.REVISION }}" | egrep "^[0-9]{1}.[0-9]{1,2}.[0-9]{1}.[0-9]{3}.[a-z0-9]{8}-snapshot+[0-9]{8}T[0-9]{6}$";
  fi
  elif [[ ${{ inputs.pkg_type }} == "stable" ]]; then
    echo "${{ steps.syslog_ng_revision.outputs.REVISION }}" | egrep "^[0-9]{1}.[0-9]{1,2}.[0-9]{1}-[0-9]{1}$"
  fi
```

.github/workflows/test-apt-packages.yml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: Check if revision is stable revision
+        if: inputs.revision_name == 'stable'
+        run: |
+          echo ${{ steps.syslog_ng_revision.outputs.REVISION }} | egrep "^[0-9]{1}.[0-9]{1,2}.[0-9]{1}-[0-9]{1}$"
+
+      - name: Check if installed package version matches with install revision
+        run: |
+          echo "Installed revision value: ${{ steps.syslog_ng_revision.outputs.REVISION }}"
```
gaborznagy, line 75: Can we add a `dpkg -l | grep syslog-ng` command too for debugging purposes here, please?

The actual checking with `dpkg -l` doesn't show the list of installed packages if the validation fails:
https://github.com/mitzkia/syslog-ng/runs/5095931848?check_suite_focus=true
(Thumbs up that you tried out failing cases of version validation)

.github/workflows/test-apt-packages.yml
```diff
@@ -0,0 +1,82 @@
...
+          dpkg -l | grep syslog-ng | while read installed_syslog_ng_package ; do echo $installed_syslog_ng_package | grep ${{ steps.syslog_ng_revision.outputs.REVISION }} ; done
```
gaborznagy, line 76: optional as the current validation method works too:

You can check the version more strictly.
`dpkg-query --show | grep syslog-ng` would have a limited output of package\tversion e.g. `syslog-ng-core  3.35.1-1`, which you check against by extracting the version number from that output.

gaborznagy (commented):
I'll only have these minor comments.

kira-syslogng:
Build SUCCESS

.github/workflows/trigger-tests-on-nightly-packages.yml
```diff
@@ -0,0 +1,11 @@
+name: Trigger tests for nightly APT packages
+
+on:
+  push:
```
alltilla, line 4: Shouldn't this be a cron schedule? Am i missing something?

.github/workflows/trigger-tests-on-stable-packages.yml
```diff
@@ -0,0 +1,11 @@
+name: Trigger tests for stable APT packages
```
alltilla, line 1: I think this should be only started by a `workflow_dispatch` trigger. We can even do that in the `.github/workflows/test-apt-packages.yml` file itself, like:
```
on:
  workflow_call:
    inputs:
      pkg_type:
        required: true
        type: string

  workflow_dispatch:
    inputs:
      pkg_type:
        description: package-type
        type: choice
        options:
          - nightly
          - stable
```

See https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/

It could be possible to schedule the nightly run there, if we add a `DEFAULT_PKG_TYPE: nightly` env variable and use it, but I am not sure if it is the correct way to go.

.github/workflows/trigger-tests-on-nightly-packages.yml
gaborznagy: It's under testing I guess (I did the same when testing). I've requested a minor change.

.github/workflows/trigger-tests-on-stable-packages.yml
gaborznagy: @alltilla FYI, this was the last list of triggers.
(AFAIK it is planned in the future to check a pre-release package)

```
on:
  workflow_dispatch:
  push:
    tags:
      - 'syslog-ng-*'
```

BTW I like the `choice` input method. :)

.github/workflows/trigger-tests-on-nightly-packages.yml
mitzkia: Sorry about it, I have tried some new changes, and I would like to know if it works or not -> because of this I have changed it temporary to push.

.github/workflows/trigger-tests-on-stable-packages.yml
alltilla: Thanks for the clarification :)

.github/workflows/test-apt-packages.yml
mitzkia: Thanks for the note, agreed with it. Fixed.

.github/workflows/test-apt-packages.yml
mitzkia: I have tried it in several way, but the internal bash did not like it.
So finally I have put back the original state

.github/workflows/trigger-tests-on-nightly-packages.yml
MrAnno: Sorry, I was not precise enough, something like this will do the trick:
`if: (github.event_name == 'schedule' && github.repository_owner == 'syslog-ng') || (github.event_name != 'schedule')`

Basically, we
- should allow starting all of our scheduled jobs manually, even in forks
- should not run those jobs on schedule in forks (we have 400 forks, fortunately only ~5% of them is updated to run our workflows)

kira-syslogng:
Build SUCCESS

mitzkia:
Thanks for the review comments, I have fixed them.
Could you please check the latest version?

gaborznagy (approved):
Thanks for the changes!! :)

kira-syslogng:
Build SUCCESS

PR #3905, opened 2022-02-07
Provides CRI log file parser for CRI-O/K8s

### LINKED ISSUE #3904: Parser for CRI-O logs
## Description of the problem
The syslog-ng might be used as DaemonSet on K8s deployments with providing the way to process log files of CRI format.
https://github.com/kubernetes/kubernetes/issues/44976

## Proposed solution
Provide parser as SCL.

## Alternatives
n-a

## Additional context
CRI define tagging for final lines and partial log entries for multiline messages.

peterviskup:
Implements #3904 feature request
Does not support multiline processing ATM

Signed-off-by Peter Viskup

Resolves #3904

### DIFF
--- a/scl/cri/cri.conf
+++ b/scl/cri/cri.conf
@@ -18,9 +18,13 @@
 
 @define crio-log-regexp "([a-z0-9-]+)_([a-z0-9-]+)_([a-z0-9]+)-([a-z0-9]{64}).log"
 
-block rewrite extract-container-id (value("MESSAGE")){
+block rewrite extract-container-metadata (value("MESSAGE") prefix('.cri.'){
     # make the container-id of 12 characters long as usual in cli
     subst("`crio-log-regexp`", "$(substr $4 0 12)", value("`value`"), flags(global, store-matches), type(pcre));
+    set("$1", value("`prefix`pod.name"));
+    set("$2", value("`prefix`pod.namespace"));
+    set("$3", value("`prefix`container.name"));
+    groupunset(values("0 1 2 3 4"));
 };
 
 block parser cri-parser(prefix('.cri.')) {
@@ -38,10 +42,7 @@ block parser cri-parser(prefix('.cri.')) {
         #filter { not match("" value("`prefix`COMMAND") type(string)); };
         rewrite {
             set("$(basename \"${FILE_NAME}\")", value("`prefix`container.id"));
-            extract-container-id(value("`prefix`container.id"));
-            set("$1", value("`prefix`pod.name"));
-            set("$2", value("`prefix`pod.namespace"));
-            set("$3", value("`prefix`container.name"));
+            extract-container-metadata(value("`prefix`container.id"));
         };
     };
 };
--- a/scl/cri/cri.conf
+++ b/scl/cri/cri.conf
@@ -16,7 +16,8 @@
 # Example log:
 # 2022-02-04T18:14:43.219493781+01:00 stdout F Starting up on port 80
 
-@define crio-log-regexp "([a-z0-9-]+)_([a-z0-9-]+)_([a-z0-9]+)-([a-z0-9]{64}).log"
+#                        pod.name    _pod.namespace _container -container.id
+@define crio-log-regexp "([a-z0-9-]+)_([a-z0-9-]+)_([a-z0-9-]+)-([a-z0-9]{64}).log"
 
 block rewrite extract-container-metadata (value("MESSAGE") prefix('.cri.'){
     # make the container-id of 12 characters long as usual in cli
--- a/scl/cri/cri.conf
+++ b/scl/cri/cri.conf
@@ -1,0 +1,51 @@
+#########################################################
+# Parser for CRI log format
+# ISO_TIME(nano) SP STREAM SP TAG SP MSG
+# https://github.com/kubernetes/kubernetes/issues/44976
+#========================================================
+# Retrieves information about K8S meta from log filename:
+#  - container ID
+#  - container Name
+#  - POD NameSpace
+#  - POD Name
+# Use .cri prefix
+# Set .cri.time as S_ISODATE with use of date-parser()
+#
+# Note: does not support multiline, line = message
+#########################################################
+# Example log:
+# 2022-02-04T18:14:43.219493781+01:00 stdout F Starting up on port 80
+
+@define crio-log-regexp "([a-z0-9-]+)_([a-z0-9-]+)_([a-z0-9]+)-([a-z0-9]{64}).log"
+
+block rewrite extract-container-id (value("MESSAGE")){
+    # make the container-id of 12 characters long as usual in cli
+    subst("`crio-log-regexp`", "$(substr $4 0 12)", value("`value`"), flags(global, store-matches), type(pcre));
+};
+
+block parser cri-parser(prefix('.cri.')) {
+    channel {
+        parser {
+            csv-parser(
+                columns("`prefix`time", "`prefix`stream", "`prefix`flag", "`prefix`message")
+                delimiters(" ")
+                flags(greedy)
+            );
+            # set S_ISODATE to value of ${.cri.time}
+            date-parser(format("%FT%H:%M:%S.%f%Z") template("${.cri.time}"));
+        };
+	# drop everything which does not match (to exclude logs)
+        #filter { not match("" value("`prefix`COMMAND") type(string)); };
+        rewrite {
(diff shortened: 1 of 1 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

scl/cri/cri.conf
```diff
@@ -0,0 +1,51 @@
...
+            extract-container-id(value("`prefix`container.id"));
+            set("$1", value("`prefix`pod.name"));
+            set("$2", value("`prefix`pod.namespace"));
+            set("$3", value("`prefix`container.name"));
+        };
+    };
+};
+
+application cri[syslog] {
+    parser { cri-parser(); };
```
bazsi, line 50: Can you add a filter that pre-classifies this log format?
For performance reason it's much better to find a filter expression that exclude irrelevant logs

scl/cri/cri.conf
```diff
@@ -0,0 +1,51 @@
...
+block rewrite extract-container-id (value("MESSAGE")){
+    # make the container-id of 12 characters long as usual in cli
+    subst("`crio-log-regexp`", "$(substr $4 0 12)", value("`value`"), flags(global, store-matches), type(pcre));
+};
+
+block parser cri-parser(prefix('.cri.')) {
+    channel {
+        parser {
+            csv-parser(
+                columns("`prefix`time", "`prefix`stream", "`prefix`flag", "`prefix`message")
```
bazsi, line 30: These are all part of the $MESSAGE macro? What is in the syslog header in this case?

Could you show an example?

scl/cri/cri.conf
```diff
@@ -0,0 +1,51 @@
...
+                delimiters(" ")
+                flags(greedy)
+            );
+            # set S_ISODATE to value of ${.cri.time}
+            date-parser(format("%FT%H:%M:%S.%f%Z") template("${.cri.time}"));
+        };
+	# drop everything which does not match (to exclude logs)
+        #filter { not match("" value("`prefix`COMMAND") type(string)); };
+        rewrite {
+            set("$(basename \"${FILE_NAME}\")", value("`prefix`container.id"));
```
bazsi, line 40: This is only applied to a file source?

Maybe in that case the "syslog" topic is not the right one.

Apart from a sample log line can you post a complete config?

Or at least explain how these messages get to syslog-ng? Thanks

scl/cri/cri.conf
```diff
@@ -0,0 +1,51 @@
...
+            extract-container-id(value("`prefix`container.id"));
+            set("$1", value("`prefix`pod.name"));
+            set("$2", value("`prefix`pod.namespace"));
+            set("$3", value("`prefix`container.name"));
```
bazsi, line 44: Those set() operations could be moved to the block. It's better when the $1, $2 etc are used locally, where they are defined

scl/cri/cri.conf
peterviskup: Yes. It's file source.
Example source config:
https://github.com/peterviskup/syslog-ng-container/blob/main/syslog-ng/conf.d/10-source-containers.conf
parser call
https://github.com/peterviskup/syslog-ng-container/blob/main/syslog-ng/conf.d/20-store-containers.conf
and template for syslog forward
https://github.com/peterviskup/syslog-ng-container/blob/main/syslog-ng/conf.d/09-templates.conf#L5
and Kubectl DeamonSet for syslog-ng with /var/log/containers and /var/log/pods paths available
https://github.com/peterviskup/syslog-ng-container/blob/main/kubernetes/syslog-ng-daemonset.yaml

scl/cri/cri.conf
peterviskup: It's file source with no-parse flag set.
https://github.com/peterviskup/syslog-ng-container/blob/main/syslog-ng/conf.d/10-source-containers.conf#L7

scl/cri/cri.conf
peterviskup: The only meaningful filter can be something like:
"FILE_NAME=/var/log/containers/"
other filters for message processing might be set in syslog-ng configuration. Some K8s namespace or POD might be out of interest.

Following are all variables available prior parsing (file source with no-parse flag set):
FILE_NAME=/var/log/containers/coredns-84c56f7bfb-rjb99_kube-system_coredns-5fa6c8559123e406fcfa49bf60a01908498d34480d811720d145299a166e8b04.log
HOST=s_containers@syslog-xp8lx
HOST_FROM=syslog-xp8lx
MESSAGE="2022-02-03T10:03:52.545549629+01:00 stdout F [WARNING] No files matching import glob pattern: /etc/coredns/custom/*.server"
SOURCE=s_containers

mitzkia:
Could you please provide some cri-parser() testcases in Light?
See examples for other parsers here:  [parser testcases for Light](https://github.com/syslog-ng/syslog-ng/tree/master/tests/light/functional_tests/parsers)

scl/cri/cri.conf
bazsi: Oh, I didn't  know your syslog-ng-container initiative but it is really exciting.

Normally, each "topic" in app-parser identifies a means how log is entered syslog-ng, and reading application specific logfiles fits the bill of "means-of-getting-logs". And I want to make this as out-of-the-box as possible. The topic makes it possible to add application adapters, e.g. configuration snippets that classify and parse application specific originating from a specific topic.

The default-network-drivers() source encapsulates the best practice how someone should consume syslog in a number of transports. This driver defines "syslog" and "syslog-raw" topics. "syslog-raw" as a topic can be used for non-standard logs, Cisco for instance breaks syslog in a number of interesting ways. So the application adapter for cisco logs taps into syslog-raw, where it gets a chance to fix it prior to parsing. "syslog" on the other hand is for log data that is successfully  parsed.

With that in mind, I I would use a separate topic for logs coming from /var/log/container, especially if they are in a specific format (which seemingly is).

And the source to encapsulate your file-source could be a source block of its own:

```
block source local-containers() {
  wildcard-file(
    base-dir("/var/log/containers")
    filename-pattern("*.log")
    recursive(no)
    follow-freq(1)
    flags(no-parse)
  );
  channel {
      parser { app-parser(topic(containers-log)); };
  };
};
```

Then the user is able to use your snippet this way:

```
source s_whatever {
    system();
    default-network-drivers();
    local-containers();
};
```

This way the user has freedom to define her sources, while the entire logic of fetching the logs/classifying/parsing them is
encapsulated in the local-containers() block and the application adapters "subscribing" to the "containers-log" topic.

Of course the naming is probably not right and could be improved above, also some of the parameters might need to
be extracted so that the user can override it, but I hope you get the idea.

This might not match exactly what you are trying to achieve. Can we perhaps jump to our gitter.im channel for a chat? I would like to understand your use-case and see if the current app-parser() model applies entirely or it needs to be adapted. Thanks.

peterviskup: this seems way better and does make sense
of course, we can have a chat

scl/cri/cri.conf
peterviskup: moved in https://github.com/peterviskup/syslog-ng/commit/bc7c82bc1701d2ac19419b0b00f96d501b26dfa3

kira-syslogng:
Build SUCCESS

peterviskup:
Definitely, will have a look on it on Monday.

Peter

Dňa pi 25. 3. 2022, 16:50 Gábor Nagy ***@***.***> napísal(a):

bazsi:
@peterviskup I finally had some time to do some reading in k8s to understand more about /var/log/containers, cri etc. Hefty reading, long time due.

I was wondering why we are not trying to fetch from /var/log/pods, seems to be the new direction. Also /var/log/pods has started to contain the same information about pods than files under /var/log/containers since kubernetes 1.14. 
https://github.com/kubernetes/kubernetes/pull/74441

I've started to change your SCL snippet to include support for both /var/log/containers and /var/log/pods. I also added decoding of the log lines and support for multi-line logs. (although I don't know if I handle embedded newlines correctly, now as I think of it I may not)

It'd be possible to add support for the JSON format also used by k8s.

Here's the patch: https://github.com/bazsi/syslog-ng/tree/cri-parser-ala-bazsi

What do you think?

bazsi:
I have now opened #3990 which contains my patches on top of yours.

bazsi:
I am closing this in favour of #3990, thanks for your contributions so far. Any feedback on the followup PR would be appreciated.

PR #3912, opened 2022-02-10
filter/test_filters_fop_cmp: refactor testcases for easier modification

bazsi:
This patch refactors this test program to make it easier to modify in upcoming changes.

### DIFF
--- a/lib/filter/tests/test_filters_fop_cmp.c
+++ b/lib/filter/tests/test_filters_fop_cmp.c
@@ -237,6 +237,21 @@ Test(filter, test_string_ordering)
   cr_assert(evaluate("11", KW_GT, "10"));
 }
 
+Test(filter, test_string_ordering_with_non_numbers)
+{
+  cr_assert(evaluate("alma", KW_LT, "korte"));
+  cr_assert(evaluate("alma", KW_LE, "korte"));
+  cr_assert_not(evaluate("alma", KW_EQ, "korte"));
+  cr_assert_not(evaluate("alma", KW_GE, "korte"));
+  cr_assert_not(evaluate("alma", KW_GT, "korte"));
+
+  cr_assert_not(evaluate("korte", KW_LT, "alma"));
+  cr_assert_not(evaluate("korte", KW_LE, "alma"));
+  cr_assert_not(evaluate("korte", KW_EQ, "alma"));
+  cr_assert(evaluate("korte", KW_GE, "alma"));
+  cr_assert(evaluate("korte", KW_GT, "alma"));
+}
+
 static void
 setup(void)
 {


kira-syslogng:
Build SUCCESS

MrAnno (commented):
LGTM.
Shouldn't we keep a few real string comparisons (like the original "alma, korte") in the test, to show those are meant to be used mainly with alphabetical characters?

lib/filter/tests/CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ set(TEST_FILTERS_NETMASK6_SOURCE
 add_unit_test(CRITERION TARGET test_filters_facility SOURCES ${TEST_FILTERS_FACILITY_SOURCE} DEPENDS syslogformat)
 add_unit_test(CRITERION TARGET test_filters_level_new SOURCES ${TEST_FILTERS_LEVEL_NEW_SOURCE} DEPENDS syslogformat)
 add_unit_test(LIBTEST CRITERION TARGET test_filters_regexp SOURCES ${TEST_FILTERS_REGEXP_SOURCE} DEPENDS syslogformat)
-add_unit_test(CRITERION TARGET test_filters_fop_cmp SOURCES ${TEST_FILTERS_FOP_CMP_SOURCE} DEPENDS syslogformat)
+add_unit_test(LIBTEST CRITERION TARGET test_filters_fop_cmp SOURCES ${TEST_FILTERS_FOP_CMP_SOURCE} DEPENDS syslogformat)
```
Kokan, line 46: With this change, the test should not depend on syslogformat.

lib/filter/tests/test_filters_fop_cmp.c
```diff
@@ -21,72 +22,234 @@
...
+  cr_assert(evaluate("10", KW_LE, "11"));
+  cr_assert_not(evaluate("10", KW_EQ, "11"));
+  cr_assert_not(evaluate("10", KW_GE, "11"));
+  cr_assert_not(evaluate("10", KW_GT, "11"));
+
+  cr_assert_not(evaluate("11", KW_LT, "10"));
+  cr_assert_not(evaluate("11", KW_LE, "10"));
+  cr_assert_not(evaluate("11", KW_EQ, "10"));
+  cr_assert(evaluate("11", KW_GE, "10"));
+  cr_assert(evaluate("11", KW_GT, "10"));
```
MrAnno, line 237: I think it would also be worth adding a test case that documents the weirdness of string comparison applied on numeric characters, for example:

```
"11" is KW_GT than "1000"
```

lib/filter/tests/CMakeLists.txt
bazsi: done, in the Makefile as well.

lib/filter/tests/test_filters_fop_cmp.c
bazsi: I think this is included, with "7" and "10" which is the same problem.

bazsi:
I have addressed add review comments I think, with followup commits so it's easy to review it incrementally.

Kokan (approved):

lib/filter/tests/test_filters_fop_cmp.c
MrAnno: Sorry, I did not see them.

MrAnno (approved):
Thank you. LGTM.

kira-syslogng:
Build SUCCESS

PR #3913, opened 2022-02-11
Allow embedded nuls

### LINKED ISSUE #3755: In what case does syslog-ng cannot parse source message
I have seen this case where syslog-ng cannot parse source message like if device is sending logs to syslog-ng then syslog-ng parse only half of the message i think the reason might be dot `.` present in the message.

bazsi:
This patch adds a new transport "text-with-nuls" that can be used to accept log messages that have embedded NULs in them.

It is recommended to use flags(no-multi-line) on the source in this case, so that NULs are replaced with a space upon
reception. Otherwise funny things could happen.

This sample source works for me;

```
	source { network(port(2000) transport(text-with-nuls) flags(no-multi-line)); };
```

Resolves #3755

### DIFF
--- a/lib/eventlog/src/evtlog.h
+++ b/lib/eventlog/src/evtlog.h
@@ -140,6 +140,7 @@ void evt_rec_free(EVTREC *e);
  * Return value: 0 to indicate failure and 1 for success
  **/
 EVTTAG *evt_tag_str(const char *tag, const char *value);
+EVTTAG *evt_tag_mem(const char *tag, const void *value, size_t len);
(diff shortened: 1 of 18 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi:

kira-syslogng:
Build SUCCESS

MrAnno (approved):

OverOrion (approved):

PR #3916, opened 2022-02-12
WIP: LogThreadedSourceDriver batching support AND increase the maximum number of threads to 256

bazsi:
This branch implements batching support in LogThreadedSourceDriver, making it
possible to perform better in LogQueueFifo, due to the ability to use per-thread input queues.

At the same time this branch simplifies thread_id allocation and removes the previously hard-coded limit of 64 worker threads and bumps the current maximum to 256. 

@MrAnno we've discussed these changes IRL a few weeks ago.
@czanik I think you already tried syslog-ng on a box that had more than 64 cores, would be great to test it on more.

### DIFF
--- a/lib/logthrsource/logthrsourcedrv.c
+++ b/lib/logthrsource/logthrsourcedrv.c
@@ -311,12 +311,35 @@ _apply_default_priority_and_facility(LogThreadedSourceDriver *self, LogMessage *
   msg->pri = parse_options->default_pri;
 }
 
+/*
+ * Call this every some messages so consumers that accumulate multiple
+ * messages (LogQueueFifo for instance) can finish accumulation and go on
+ * processing a batch.
+ *
+ * Basically this calls main_loop_worker_invoke_batch_callbacks(), which is
+ * done by the minaloop-io-worker layer whenever we go back to the main
+ * loop.  Whether this is done automatically by LogThreadedSourceDriver is
+ * controlled by the auto_close_batches member, in which case we do this
+ * every message.
+ *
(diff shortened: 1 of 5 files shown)


kira-syslogng:
Build FAILURE

MrAnno:
A bit unrelated, but right after the previous threaded source modification, the threaded fetcher and threaded source unit tests have started crashing unpredictably, but more and more frequently. We should check it.

bazsi:
There's still an issue with this branch, namely thread based sources will be able to allocate thread_ids, but their number is not calculated in log_queue_max_threads, which means that the per-thread state of LogQueueFifo is not properly allocated.

In that case this assert would fail in logqueue-fifo.c:

```
  g_assert(thread_id < 0 || log_queue_max_threads > thread_id);
```

One way to solve that is to _always_ allocate thread states up to MAIN_LOOP_MAX_WORKER_THREADS. With this patch
that value is increased to 256, the per-thread state for LogQueueFifo is 56 bytes, which would be 16kB per LogQueue instance, e.g. this much per destination, most of which would be unused.

The stuff I am working on needs a similarly sized per-thread state, 80 bytes.

This memory use is not *that* bad, but still. If we were to increase MAX_WORKER_THREADS further, embedded people would hate us :)

bazsi:
> A bit unrelated, but right after the previous threaded source modification, the threaded fetcher and threaded source unit tests have started crashing unpredictably, but more and more frequently. We should check it.

can you show such an failure?

bazsi:
I am going to open this as a part of a separate PR.

PR #3928, opened 2022-02-24
Logthrdestdrv dont count non local messages in seq num

bazsi:
logthrdestdrv: don't set sequenceId for non-local messages
    
    The SDATA meta.sequenceId is the concept behind our SEQNUM macro, e.g.  it
    is a destination specific sequence number of local messages being sent.
    
    But this sequenceId only needs to be assigned to messages where we are the
    originator, as described here:
    
        7.3.1.  sequenceId
    
           The "sequenceId" parameter tracks the sequence in which the
           originator submits messages to the syslog transport for sending.  It
           is an integer that MUST be set to 1 when the syslog function is
           started and MUST be increased with every message up to a maximum
           value of 2147483647.  If that value is reached, the next message MUST
           be sent with a sequenceId of 1.
    
    NOTE the word "originator" in the first sentence.  Non-local messages don't
    have a locally generated SEQNUM unless the original message (coming from its
    originator) had one.
    
    There are two ways sequenceId are associated with incoming messages: the
    structured data meta.sequenceId, or in the case of Cisco messages the
    sequence number prepended to each messages.
    
    Now, LogThreadedDestDriver based plugins don't ever implement RFC5424, but
    $SEQNUM as macro is available even in these cases, and up to this patch, we
    generated a sequenceId for all outgoing messages, local and non-local alike.
    
    Our $SDATA macro then automatically patches in the sequence number if one is
    specified by the driver which was the case.  This meant that if you are
    using the $SDATA macro in the context of http() for instance, your SDATA
    will magically include an extra meta.sequenceId element, which is not what
    you want.  This behavior is fine when _we_ are the originator of said
    message (as in that case, syslog-ng itself is free to add such an meta
    element), but can be confusing and to be avoided when we are forwarding
    incoming logs.
    
    This patch changes LogThreadedDestDriver behavior to match that of LogWriter
    instances.

### DIFF
--- a/news/other-3928.md
+++ b/news/other-3928.md
@@ -1,0 +1,6 @@
+`java()/python() destinations`: the $SEQNUM macro (and "seqnum" attribute in
+Python) was erroneously for both local and non-local logs, while it should
+have had a value only in case of local logs to match RFC5424 behavior
+(section 7.3.1).  This bug is now fixed, but that means that all non-local
+logs will have $SEQNUM set to zero from this version on, e.g.  the $SEQNUM
+macro would expand to an string, to match the syslog() driver behaviour.
--- a/lib/logthrdest/logthrdestdrv.h
+++ b/lib/logthrdest/logthrdestdrv.h
@@ -191,10 +191,15 @@ log_threaded_dest_worker_disconnect(LogThreadedDestWorker *self)
 static inline LogThreadedResult
 log_threaded_dest_worker_insert(LogThreadedDestWorker *self, LogMessage *msg)
 {
-  if (self->owner->num_workers > 1)
-    self->seq_num = step_sequence_number_atomic(&self->owner->shared_seq_num);
+  if (msg->flags & LF_LOCAL)
+    {
+      if (self->owner->num_workers > 1)
+        self->seq_num = step_sequence_number_atomic(&self->owner->shared_seq_num);
+      else
+        self->seq_num = step_sequence_number(&self->owner->shared_seq_num);
+    }
   else
-    self->seq_num = step_sequence_number(&self->owner->shared_seq_num);
+    self->seq_num = 0;
   return self->insert(self, msg);
 }
 
--- a/lib/logthrdest/tests/test_logthrdestdrv.c
+++ b/lib/logthrdest/tests/test_logthrdestdrv.c
@@ -30,7 +30,6 @@
 #include "mainloop-worker.h"
 #include "apphook.h"
 
-
 typedef struct TestThreadedDestDriver
 {
   LogThreadedDestDriver super;
@@ -109,7 +108,7 @@ _spin_for_counter_value(StatsCounterItem *counter, gssize expected_value)
 }
 
 static void
-_generate_messages(TestThreadedDestDriver *dd, gint n)
+_generate_messages(TestThreadedDestDriver *dd, gint n, gboolean local)
 {
   LogMessage *msg;
   LogPathOptions path_options = LOG_PATH_OPTIONS_INIT_NOACK;
@@ -121,6 +120,10 @@ _generate_messages(TestThreadedDestDriver *dd, gint n)
 
       g_snprintf(buf, sizeof(buf), "%d", i);
       log_msg_set_value(msg, LM_V_PID, buf, -1);
+      if (local)
+        msg->flags |= LF_LOCAL;
+      else
+        msg->flags &= ~LF_LOCAL;
 
       log_pipe_queue(&dd->super.super.super.super, msg, &path_options);
     }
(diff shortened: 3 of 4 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
added tests and newsfile. 

@rfaircloth-splunk this should fix your "[meta sequenceId]" addition to $SDATA.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

MrAnno (commented):
LGTM.

This bugfix affects the Python and Java destinations as well.
I think it would be worth adding a `other-3928.md` news entry for notifying Python/Java "users"  that we previously always set seqnum by mistake, and that they should prepare for 0 values from now on.

MrAnno:
@kira-syslogng test this please test branch=logthrdestdrv-dont-count-non-local-messages-in-seq-num;

kira-syslogng:
Build FAILURE

MrAnno:
@kira-syslogng test this please test branch=logthrdestdrv-dont-count-non-local-messages-in-seq-num;

kira-syslogng:
Build FAILURE

MrAnno:
@kira-syslogng test this please test branch=logthrdestdrv-dont-count-non-local-messages-in-seq-num;

kira-syslogng:
Build SUCCESS

bazsi:
> This bugfix affects the Python and Java destinations as well. I think it would be worth adding a `other-3928.md` (or a developer note) news entry for notifying Python/Java "users" that we previously always set seqnum by mistake, and that they should prepare for empty values from now on.

I have added this note.

news/other-3928.md
```diff
@@ -0,0 +1,6 @@
+`java()/python() destinations`: the $SEQNUM macro (and "seqnum" attribute in
+Python) was erroneously for both local and non-local logs, while it should
+have had a value only in case of local logs to match RFC5424 behavior
+(section 7.3.1).  This bug is now fixed, but that means that all non-local
+logs will have $SEQNUM set to zero from this version on, e.g.  the $SEQNUM
+macro would expand to an string, to match the syslog() driver behaviour.
```
MrAnno, line 6: > would expand to an string

The word "empty" seems to be missing (we can fix the typo when generating the news file).

MrAnno (approved):
Thank you.

kira-syslogng:
Build FAILURE

MrAnno:
@kira-syslogng test this please test branch=logthrdestdrv-dont-count-non-local-messages-in-seq-num;

kira-syslogng:
Build SUCCESS

MrAnno:
@kira-syslogng test this please test branch=logthrdestdrv-dont-count-non-local-messages-in-seq-num;

(We've made a test stricter)

kira-syslogng:
Build SUCCESS

alltilla:
I think more destinations are affected, other than python and java:
  * amqp
  * mongodb
  * smtp
  * sql
  * stomp
  * http
  * kafka
  * mqtt
  * redis
  * riemann

Most of them use the `seq_num` field through `LogTemplateEvalOptions`, but there is one, which uses it for another reason:
https://github.com/syslog-ng/syslog-ng/blob/master/modules/afstomp/afstomp.c#L297
```
  if (self->ack_needed)
    {
      g_snprintf(seq_num, sizeof(seq_num), "%i", self->super.worker.instance.seq_num);
      stomp_frame_add_header(&frame, "receipt", seq_num);
    };
```
If I understand correctly, it is used as a unique ID for message acknowledgement. With the change the uniqueness breaks. I do not know how the server behaves if this happens.

Useful documentations:
  * https://stomp.github.io/stomp-specification-1.2.html#Header_receipt
  * https://stomp.github.io/stomp-specification-1.2.html#RECEIPT

bazsi:
> I think more destinations are affected, other than python and java:
> 
> * amqp
> * mongodb
> * smtp
> * sql
> * stomp
> * http
> * kafka
> * mqtt
> * redis
> * riemann
> 
> Most of them use the `seq_num` field through `LogTemplateEvalOptions`, but there is one, which uses it for another reason: https://github.com/syslog-ng/syslog-ng/blob/master/modules/afstomp/afstomp.c#L297
> 
> ```
>   if (self->ack_needed)
>     {
>       g_snprintf(seq_num, sizeof(seq_num), "%i", self->super.worker.instance.seq_num);
>       stomp_frame_add_header(&frame, "receipt", seq_num);
>     };
> ```
> 
> If I understand correctly, it is used as a unique ID for message acknowledgement. With the change the uniqueness breaks. I do not know how the server behaves if this happens.
> 
> Useful documentations:
> 
> * https://stomp.github.io/stomp-specification-1.2.html#Header_receipt
> * https://stomp.github.io/stomp-specification-1.2.html#RECEIPT

This is a single use identifier that is waited on right after testing. Does not need to be unique as long as we don't  have multiple messages in flight. And right now we don't.

alltilla:
@bazsi 

Thanks for the explanation.

I opened a PR to clarify this: #3958

alltilla (approved):

PR #3932, opened 2022-03-02
github-actions: do concurrent image building (dbld-build-images)

gaborznagy:
It scales better and much easier to debug what had failed.

Depending images (e.g. `tarball` -> `debian-testing`) still not resolved.

Also changed the cases when the github actions is triggered: in cases of PR/push events the github action will only start if
the content of `dbld`, the `packaging/(rhel|debian)` directories or the github actions yaml file itself change.

Example run, where `centos-7` failed: https://github.com/gaborznagy/syslog-ng/actions/runs/1924046749

### DIFF
--- a/.github/workflows/dbld-images.yml
+++ b/.github/workflows/dbld-images.yml
@@ -16,6 +16,24 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        image:
+          - centos-7
+          - fedora-33
+          - debian-bullseye
+          - debian-buster
+          - debian-sid
+          - debian-stretch
+          - debian-testing
+          - ubuntu-bionic
+          - ubuntu-focal
+          - ubuntu-impish
+          - ubuntu-xenial
+          - devshell
+          - kira
+          - tarball
+      fail-fast: false
     steps:
       - name: Checkout source code
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build SUCCESS

Kokan (dismissed):

.github/workflows/dbld-images.yml
```diff
@@ -29,7 +47,7 @@ jobs:
           gh_export CONTAINER_REGISTRY
 
       - name: Build the images
-        run: dbld/rules images -j $(nproc) --keep-going --output-sync
+        run: dbld/rules image-${{ matrix.image }}
```
mitzkia, line 50: Is not missing the original parameter list: "-j $(nproc) --keep-going --output-sync"

MrAnno:
This increases the number of PR check jobs to 32, which is a lot. Can we also add a filter to start these jobs on push/PRs only when something changes in dbld?

We can run the it scheduled without any conditions.

.github/workflows/dbld-images.yml
gaborznagy: Thanks @Kokan for the explanation, I agree, I removed them as they were unnecessary.

mitzkia: Thanks for the explanation.

kira-syslogng:
Build SUCCESS

MrAnno (approved):

PR #3934, opened 2022-03-03
cisco-parser: support ISO 8601 timestamp

gaborznagy:
We've received a new Cisco log format with ISO 8601 timestamp format
```
<166>2022-02-16T15:31:53Z na-zy-int-fp1140-p02 : %FTD-6-305012: Teardown dynamic TCP translation from FOO-WAN_IN:10.44.60.80/59877 to FOO-OUTSIDE:6.7.8.9/59877 duration 0:01:01
```

Originally reported on the mail list:
https://lists.balabit.hu/pipermail/syslog-ng/2022-February/026383.html
https://lists.balabit.hu/pipermail/syslog-ng/2022-March/026398.html

**TODO**:
 * [x] solve the seqid removing vs ISO timestamp format issue
 
 ### This depends on #3948


gaborznagy:
@bazsi I would like to know your opinion on this.

bazsi:
From both the patch and the documentation of Firepower (https://www.cisco.com/c/en/us/td/docs/security/firepower/Syslogs/b_fptd_syslog_guide/about.html)  the timestamp in this case omits the NTP synchronization indicator (e.g. '.' or '*')  so the timestamp parsing part seems OK. It might make sense to look into improving the '.' vs '*' parsing a bit (the builtin syslogformat module parses that and extracts its value into .SDATA.timeQuality.isSynced).

Using the current method (e.g. two matches in sequence) is somewhat slower for Firepower logs, which could be improved by separating out the timeQuality parsing.

With that said however, I don't think this new Firepower format would  dominate Cisco logs, so for now it's fine.

On the sequence number part however,that needs to stay there in some form. Normally the sequence number is terminated using ": " while the ISO timestamp does not have that space right after the colon. So I am thinking this should work:

subst("^([0-9]+: )?", "", value('1'));

I am trying this in a moment.

bazsi:
On Mon, Mar 7, 2022, 20:42 John ***@***.***> wrote:

> I've applied this patch, but I'm not seeing anything filtered on my end.
> What's the best way to test this out on my live data? Should I do something
> like:
>
> tail raw.log | syslog-ng -f /tmp/log/test.conf -v -d -t -e -F
>
> Of do I need to push the data via 'nc' instead to get some testing going?
>

Yup, something like

$ cat raw.log | nc -q0 localhost <port>

John

bazsi:
The easiest to debug this ATM is to enable tracing and sending a single message and look at the result.

If you post the log I can help diagnosing it.

gaborznagy:
Thanks @bazsi for the seqnum part suggestion. I was just quickly commented out that part, but I knew we needed that.
I've checked it too and it works for me too!

I'll update the PR, so we should see that the `devshell` jobs pass.

gaborznagy:
@bazsi @l8gravely  I was using the example messages from the mailing list:
https://lists.balabit.hu/pipermail/syslog-ng/2022-February/026383.html

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

bazsi (approved):

gaborznagy:
I'm setting this PR as WIP, as I would like to investigate the subst() capture group problem, see https://github.com/syslog-ng/syslog-ng/pull/3934/commits/8abdb963dcf39c2893e2b7d4ee6ee6a52a39dc59

The seqno removal subst() rewrite rule can add a null-terminated string, which can cause problems.
The reasons are not entirely clear, but the capture groups in a regex pattern sets the `$1` name-value pair, this overwriting the entire input.
Since we have a reference to the nvtable element after `log_msg_get_value()` our input parameter will be overriden.
E.g.:
```
[2022-04-04T12:31:10.151216] Setting value; name='1', value='782431: machine1: .Nov 18 21:03:22.631 GMT', type='string', msg='0x7f9a28014750'

...

[2022-04-04T12:31:10.151410] >>>>>> rewrite rule evaluation begin; rule='#anon-rewrite1', location='block parser cisco-parser() at /install/syslog-ng-OSE/share/syslog-ng/include/scl/cisco/plugin.conf:87:21:13', msg='0x7f9a28015c90'
[2022-04-04T12:31:10.151441] Setting value; name='0', value='782431: ', type='string', msg='0x7f9a28015c90'
[2022-04-04T12:31:10.151451] Setting value; name='1', value='782431: ', type='string', msg='0x7f9a28015c90'
[2022-04-04T12:31:10.151466] Performing subst() rewrite; rule='#anon-rewrite1', value='1', input='782431: ', type='pcre', pattern='^([0-9]+: )?', replacement='', location='block parser cisco-parser() at /install/syslog-ng-OSE/share/syslog-ng/include/scl/cisco/plugin.conf:87:21:13'
[2022-04-04T12:31:10.151477] Setting value; name='1', value='.achine1: .Nov 18 21:03:22.631 GMT', type='string', msg='0x7f9a28015c90'

The output:
2022-11-18T21:03:22+01:00 - {"_cisco":{"severity":"4","mnemonic":"NATIVE_VLAN_MISMATCH","facility":"CDP"},"PROGRAM":"","MSG":"%CDP-4-NATIVE_VLAN_MISMATCH: Native VLAN mismatch discovered on TenGigabitEthernet.","HOST":"\\x00achine1"}
```

The expected behaviour would be this:
```
(gdb) p log_msg_get_value_by_name(msg, "1", 0)
$2 = (const gchar *) 0x7ffff0016502 "782431: machine1: .Nov 18 21:03:22.631 GMT"
(gdb) p value
$3 = (const gchar *) 0x7ffff0016502 "782431: machine1: .Nov 18 21:03:22.631 GMT"
(gdb) info locals
self = 0x555555616ac0
msg = 0x7ffff0015c90
nvtable = 0x7ffff0016280
value = 0x7ffff0016502 "782431: machine1: .Nov 18 21:03:22.631 GMT"
new_value = 0x7ffff0016b50 "machine1: .Nov 18 21:03:22.631 GMT"
length = 42
new_length = 34
```

gaborznagy:
I wouldn't use matches/temporary variables (e.g. `$1`) as input parameters, but then in the end where should be allowed to use temporary variables in cisco-parser?

kira-syslogng:
Build SUCCESS

gaborznagy:

kira-syslogng:
Build FAILURE

bazsi:
Once we rebase this against #3948 it should be green.

kira-syslogng:
Build SUCCESS

gaborznagy:
I've did a tiny refactor (match() filter -> regexp-parser() replacement), and rebased the branch on the latest #3948 .

gaborznagy:
The PR is finalized and ready for review. It depends on #3948, so it cannot be merged before #3948.

kira-syslogng:
Build FAILURE

alltilla:
Wrote some E2E tests to check if we correctly parse all the necessary fields:
https://github.com/gaborznagy/syslog-ng/pull/10

It seems like that with the new pattern, the `${HOST}` field is not getting parsed.
```
>       assert output_json["HOST"] == expected_kv_pairs["HOST"]                                                                                                                                                      
E       AssertionError: assert 'alltilla-Precision-5560' == 'na-zy-int-fp1140-p02'                                                                                                                                   
E         - na-zy-int-fp1140-p02                                                                                                                                                                                     
E         + alltilla-Precision-5560
```

I think we should parse it as we do with all the other formats.
What do you think?

gaborznagy:
> Wrote some E2E tests to check if we correctly parse all the necessary fields: [gaborznagy#10](https://github.com/gaborznagy/syslog-ng/pull/10)
> 
> It seems like that with the new pattern, the `${HOST}` field is not getting parsed.
> 
> ```
> >       assert output_json["HOST"] == expected_kv_pairs["HOST"]                                                                                                                                                      
> E       AssertionError: assert 'alltilla-Precision-5560' == 'na-zy-int-fp1140-p02'                                                                                                                                   
> E         - na-zy-int-fp1140-p02                                                                                                                                                                                     
> E         + alltilla-Precision-5560
> ```
> 
> I think we should parse it as we do with all the other formats. What do you think?

I agree, I'll check if I can extend the heuristic that parses the HOSt macro.

gaborznagy:
@kira-syslogng retest this please branch=cisco-parser-pr-3934;

kira-syslogng:
Build SUCCESS

gaborznagy:
About the Kira failure: when I've replaced the `match()` filters with `regexp-parser()`in the datetime-parser section, one test failed because it checked the "discard" stats counter of the cisco-parser.
The "discard" counter was set because of the missing (optional) timestamp field caused that the regexp-parser() match failed.

I don't think the "discarded" counter can be used of a parser which is implemented purely as an SCL.

kira-syslogng:
Build FAILURE

gaborznagy:
@kira-syslogng retest this please branch=cisco-parser-pr-3934;

kira-syslogng:
Build SUCCESS

MrAnno (approved):

MrAnno:
@alltilla @OverOrion Could you check this PR, please?
It depends on #3948, but that one will most likely be part of v3.37, so we could release Cisco iso-timestamp support as well.

alltilla (approved):

gaborznagy:
@kira-syslogng retest this please branch=cisco-parser-pr-3934;

gaborznagy:
I've rebased this onto master. Please don't forget the Kira test modifications after this is merged.

alltilla (approved):

MrAnno (approved):

PR #3937, opened 2022-03-04
Make public helm registry for installing syslog-ng

jmcorallo:
Following: https://helm.sh/docs/howto/chart_releaser_action/

To install syslog-ng from a public helm registry:
```
helm repo add jmcorallo https://jmcorallo.github.io/syslog-ng
helm repo update
helm install my-syslog jmcorallo/syslog-ng
```

After this is merged you can replace jmcorallo for syslog-ng


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

kira-syslogng:
Build SUCCESS

Kokan:
Notes for review:
* `gh-pages` branch must exists and that should be the source branch for github pages
* A release is automatically created (https://github.com/jmcorallo/syslog-ng/releases)

MrAnno:
Thank you for your contribution.

Currently, I'm not really familiar with Helm charts. The original chart was contributed to us, we placed it to the `contrib` to indicate that this is not fully monitored/supported/maintained by us. I'm planning to get to know it a bit better to answer a few basic questions like:

- if we moved out the charts from `contrib`, what basic steps/automation would need to be done to keep it stable
- is it a good idea to produce Helm chart releases in the main repository? For example, creating GitHub releases other than our main syslog-ng releases would be confusing in my current understanding.

jmcorallo:
Thanks for the feedback @MrAnno

> is it a good idea to produce Helm chart releases in the main repository? For example, creating GitHub releases other than our main syslog-ng releases would be confusing in my current understanding.

Good point, actually a separate repo would make more sense. Other organizations do it like this, see for example:
- https://github.com/prometheus-community/helm-charts
- https://github.com/aws/eks-charts
- https://github.com/bitnami/charts

arvindkonar:
do we know when we will have an official help repo created like https://syslog-ng.github.io/syslog-ng. Currently, we've used https://jmcorallo.github.io/syslog-ng and that works but official one would be better..

bazsi:
This PR has been overlooked over a year, sorry for that.

The reason this had such a rough path into the git repo is that this repose has been traditionally been the source code which was used to generate Linux packages.

Those contributing here (including myself) didn't know enough about helm charts to make decisions about it, thus even the original one has been in contrib.

There's however a project called AxoSyslog that aims to be the "cloud native distribution" for syslog-ng. Here's the announcement:

https://axoflow.com/cloud-ready-syslog-ng-images/

And here's the repository for the container
https://github.com/axoflow/axosyslog-docker

And the helm chart:
https://github.com/axoflow/axosyslog-charts

I do see some momentum building around that project, transforming the documentation to be cloud first and also adding typical cloud related features such as Prometheus like metrics.

I think it would make sense to move helm charts to a better home and AxoSyslog could be that home.

What do you think?

PS: I am both the original founder of syslog-ng and Axoflow.

jmcorallo:
@bazsi thanks for coming back to this PR. I agree, it makes sense to move it to it's own repo

PR #3941, opened 2022-03-09
logmsg: add "type" field to set_value()'s trace logs

alltilla:
News entry is not needed IMO.

Question for the reviewers:
  * The new field in the log message hints type features on the source side, which is a 4.0 feature. I think it is not a problem, but if we want, we can show it only if 4.0 is enabled. What do you think?

### DIFF
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -203,6 +203,28 @@ const gchar *builtin_value_names[] =
   NULL,
 };
 
+const gchar *
+log_msg_value_type_to_str(LogMessageValueType self)
+{
+  g_assert(self <= LM_VT_NONE);
+
+  static const gchar *as_str[] =
+  {
+    [LM_VT_STRING] = "string",
+    [LM_VT_JSON] = "json",
+    [LM_VT_BOOLEAN] = "boolean",
+    [LM_VT_INT32] = "int32",
+    [LM_VT_INT64] = "int64",
+    [LM_VT_DOUBLE] = "double",
+    [LM_VT_DATETIME] = "datetime",
+    [LM_VT_LIST] = "list",
+    [LM_VT_NULL] = "null",
+    [LM_VT_NONE] = "none",
+  };
+
+  return as_str[self];
+}
+
 static void
 __free_macro_value(void *val)
 {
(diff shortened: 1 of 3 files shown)


lib/logmsg/logmsg.c
```diff
@@ -203,6 +203,25 @@ const gchar *builtin_value_names[] =
...
+    [LM_VT_JSON] = "JSON",
+    [LM_VT_BOOLEAN] = "BOOLEAN",
+    [LM_VT_INT32] = "INT32",
+    [LM_VT_INT64] = "INT64",
+    [LM_VT_DOUBLE] = "DOUBLE",
+    [LM_VT_DATETIME] = "DATETIME",
+    [LM_VT_LIST] = "LIST",
+    [LM_VT_NULL] = "NULL",
+    [LM_VT_NONE] = "NONE",
+  };
```
bazsi, line 220: Could you add an assert () to ensure that we don't access that array out of bounds?

Maybe a static assert would be sufficient.

Also I'd prefer lowercase names for the types, that way we would match type hints.

And one last comment, completely optional: we have the reverse mapping in the type hint module, maybe we could refactor both cases into one set of functions/array.

alltilla: Sure, I am looking into your points!

kira-syslogng:
Build SUCCESS

lib/logmsg/logmsg.c
alltilla: Merging the two functions to use the same array was turning to be rather unpleasant, but I was able to add a `log_msg_value_type_from_str()` function for the sake of symmetry. If you are not fond of it, we can drop the last commit. :)

alltilla:
@bazsi 
IMO, the new logs would look nice in https://syslog-ng-future.blog/syslog-ng-4-theme-typing/

kira-syslogng:
Build SUCCESS

Kokan (dismissed):

lib/logmsg/logmsg.c
bazsi: This looks good to me. Thanks

bazsi (dismissed):

alltilla:
Rebased to master.

kira-syslogng:
Build SUCCESS

Kokan (approved):

Kokan:
Also to the question, it is not like there were no type system before but it was always string, and in some situation syslogng tried to interpret those strings as other things (int, list). Before 4.0 it does not expose something not existent rather something not relevant. So it is fine to write out type before 4.0.

bazsi (approved):

PR #3948, opened 2022-03-12
Matches as an array (4.0)

bazsi:
This builds on top of #3888 and implements $* as a list and other match related improvements.

These are the new features:
  * $* is a new macro that generates a list from $1, $2, $3 ... name-value pairs
  * the number of such name value pairs are tracked internally. Any new match operation (e.g. regexp matching) would reset the entire match array, while previously "stale" values could remain, if the new match operation would generate less matches than the previous one
  * added set-matches() and unset-matches() rewrite operations, the first expecting a list, while the second just resets the matches array to be empty.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

bazsi:

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

gaborznagy:
Just a note from our discussion: investigate the possibility of unsetting the stale/obsolete matches from earlier (i.e. when we have fewer matches in a subsequent matching than before).

bazsi:
> Just a note from our discussion: investigate the possibility of unsetting the stale/obsolete matches from earlier (i.e. when we have fewer matches in a subsequent matching than before).

It seems that I have implemented this already and then forgot about it. It is this patch: c67006b63d05d60530f1a435750b423281935ffe

This means that if we set non-consecutive match values then anything between the old range and the new one are unset explicitly.

I've now noticed that there's a function to reset the number of matches: log_msg_clear_matches() so I've committed changes to
use that function instead of open-coding the same.

kira-syslogng:
Build SUCCESS

lib/rewrite/rewrite-set-matches.c
```diff
@@ -0,0 +1,128 @@
...
+  log_template_format_value_and_type(self->value_template, *pmsg, &options, result, &type);
+
+  log_msg_make_writable(pmsg, path_options);
+
+  ListScanner scanner;
+
+  list_scanner_init(&scanner);
+  list_scanner_input_string(&scanner, result->str, result->len);
+  log_msg_clear_matches(*pmsg);
+  log_msg_unset_match(*pmsg, 0);
```
gaborznagy, line 67: This is just a question, but why do you explicitly unset `$0` here, while `unset-matches()` only does a `log_msg_clear_matches`?

`$*` doesn't care about `$0`, so I guess this is up to if we allow/expect that the individual matches (`$1`, `$2`, etc.) can be used separately.

I would use an actual unset than a simple clear, so the name-value pairs will be unset (empty).

gaborznagy (commented):
I've checked it, and it looks good to me!

I have only some minor remarks about commit history, about a unit test, and some general remarks.

I don't have a better, more general alternative to the term "matches" in `set-matches` or `unset-matches`.
I'm not sure of their general nature either: users of `$1` or `set-matches()` should be careful of regex capture groups that could overwrite their explicitly set matches.

In Bash, `$*`, `$1`, etc. all refer to positional parameters.
In Perl it **was** related to [multiline matching in regular expressions](https://perldoc.perl.org/perlvar#$*), but not since Perl 5.10.

gaborznagy:
I've sent a PR about the news entries.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

tests/unit/test_matcher.c
```diff
@@ -264,3 +264,45 @@ Test(matcher, pcre812_incompatibility, .description = "tests a pcre 8.12 incompa
...
+
+  cr_assert_eq(msg->num_matches, 2);
+
+  log_matcher_compile(m, "^(ki)(wi).*", NULL);
+  value = log_msg_get_value(msg, LM_V_MESSAGE, &value_len);
+  result = log_matcher_match(m, msg, LM_V_MESSAGE, value, value_len);
+  cr_assert(result);
+
+  assert_log_message_value(msg, LM_V_MESSAGE, msg_payload);
+  assert_log_message_match_value(msg, 0, value);
```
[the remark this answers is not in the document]
bazsi: good point. I have now added another step in the same unit test. The test was about LogMatcher "setting" the num_matches value, but you are right that
increasing and decreasing the number of matches should be tested distinctively.

The next push would contain a patch for this.

lib/logmsg/logmsg.c
```diff
@@ -1990,6 +2002,7 @@ log_msg_registry_init(void)
 
       g_snprintf(buf, sizeof(buf), "%d", i);
       match_handles[i] = nv_registry_alloc_handle(logmsg_registry, buf);
+      nv_registry_set_handle_flags(logmsg_registry, match_handles[i], (i << 8) + LM_VF_MATCH);
```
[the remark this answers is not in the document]
bazsi: good catch. moved bits between patches and I even reordered patches and I've missed this.

Again fixed in the next iteration.

lib/rewrite/rewrite-set-matches.c
bazsi: Good point.

I've removed this and added test coverage that log_msg_clear_matches() clears $0 too.

json-parser() and set-matches() sets the match variables starting from $1.. and $0 should be empty, but this is achieved by the log_msg_clear_matches() call (that sets num_matches to 0) and then eventually setting $1, which at
this point clears $0 lazily.

In the previous iteration $0 wouldn't have been unset and that's why it was present explicitly. But now with the explicit unsetting of matches, this is resolved.

bazsi:
Thanks @gaborznagy for the good feedback. I've now addressed all your review notes and pushed out another iteration. There seems to be some style-check related issue in the CI about a missing Python command. Otherwise things seem to be green (unless I am breaking something with the last force-push :)

Reviewing this one would be appreciated.

kira-syslogng:
Build SUCCESS

tests/unit/test_matcher.c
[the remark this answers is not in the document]
gaborznagy: Thanks!

gaborznagy (dismissed):
Thanks for resolving my comments.

gaborznagy:
@bazsi this needs one last rebase to fix the [Light style-check issue](https://github.com/syslog-ng/syslog-ng/runs/6210869837?check_suite_focus=true) in Github actions.

modules/json/json-parser.c
```diff
@@ -258,8 +258,7 @@ json_parser_process_array(struct json_object *jso,
 {
   gint i;
 
-  msg->num_matches = 0;
-  log_msg_unset_match(msg, 0);
+  log_msg_clear_matches(msg);
```
OverOrion, line 261:
```
log {
  source(s_src);
  rewrite {
    set("foo bar" value('1'));
    unset-matches();
  };
  destination { file("/dev/stdout" template("$(format-json --scope nv-pairs --key 1)\n")); };
};
```

With this patch we can still query the $1 value after the `unset-matches();` is it an intended behaviour?
I would argue that after `unset-matches()` is called every $n should be empty (""), or at least that is what I expect.

bazsi: Good catch. The log_msg_values_foreach() did not filter out out-of-range matches. I'm pushing a fix for this, however I am starting to feel that it'd be much better to explicitly unset all matches explitictly, rather than trying to track the number of matches and then checking their number in all the code paths that query name-value pairs.

Might even be faster doing it only once, instead of at every log_msg_values_foreach() iteration.

bazsi: This is the fix for this: e234146b1ae44af0cdfff06e65917d40fa14c275

kira-syslogng:
Build SUCCESS

modules/json/json-parser.c
gaborznagy: I don't want to be nitpicking especially since I've approved this PR once, but this fix in https://github.com/syslog-ng/syslog-ng/commit/e234146b1ae44af0cdfff06e65917d40fa14c275 only fixes the case when a cleared match is referenced through value-pairs, e.g. through `format-json`.

If the match (e.g. `$1`) is used later as an input in a parser or rewrite rule, then it would still hold the previously (cleared) value.
Maybe this scenario is rather a programming error, e.g. the writer of an SCL block should be aware of this.
A dummy example SCL:
```
parser example-parser() {
....
  set("foo bar" value('1'));
  unset-matches();
  syslog-parser(template('$1'));
};
```

TLDR: how about using `log_msg_unset_match()` instead of `log_msg_clear_matches()`?

kira-syslogng:
Build SUCCESS

modules/json/json-parser.c
bazsi: In the last iteration, I have changed the approach of resetting values, so the branch now contains a patch that changes things - again.

With that said, this case has worked correctly before, '$1' was unset because of this branch in log_msg_get_match():

```
+const gchar *
+log_msg_get_match_if_set_with_type(const LogMessage *self, gint index_, gssize *value_len,
+                                   LogMessageValueType *type)
+{
+  g_assert(index_ >= 0 && index_ < LOGMSG_MAX_MATCHES);
+
+  if (index_ >= self->num_matches)
+    return NULL;
+
+  return nv_table_get_value_if_set(self->payload, match_handles[index_], value_len, type);
+}
+
```

e.g. if a match number was outside the range of num_matches we always considered it to be unset (NOTE the NULL return). This worked correctly even before. The value-pairs code path had a bug as @OverOrion noticed correctly, as it didn't use log_msg_get_match(), rather it iterated over all name-value pairs using log_msg_values_foreach(), which did not filter out $1 in this case. That's what my last patch fixed.

With all this iterations however, at the end I've decided to go back to the original concept and call log_msg_unset() all matches whenever the num_matches would change. This eliminates the need to handle matches separately from other name-value pairs in the query APIs. 

This is this patch that reverses most of the above.

commit 855e0c4174ebbba5db05d9395a2d33efb6b1cc4f
Author: Balazs Scheidler <bazsi77@gmail.com>
Date:   Tue May 3 13:48:12 2022 +0200

    logmsg: unset matches proactively

With that we are failing the cisco-parser() testcase in light, so I am now going to find that issue.

kira-syslogng:
Build SUCCESS

bazsi:
Cleaned up the branch and rebased against current master.

kira-syslogng:
Build SUCCESS

OverOrion:
Thanks @bazsi, I've checked the current state of this PR and it LGTM, only a rebase is needed (because #3972 has been merged).
I've tested it with mentioned config and the values are no longer set after `unset-values()`.

Just a small note (more like a question):
The newly created `lib/rewrite/tests/test_set_matches.c` file contains this line:
`* Copyright (c) 2019 Kokan <kokaipeter@gmail.com>`
I did not find a reason for it, was there any?

bazsi:
I've rebased against current master and resolved the conflicts. The copyright header was incorrect as I've copied the header from a different test program. Fixes those as well.

Thanks for the review. Hope it is good to go in at this point.

kira-syslogng:
Build SUCCESS

bazsi:
fixed CI failure

kira-syslogng:
Build SUCCESS

gaborznagy (commented):
Sorry Bazsi for the late reply, I didn't finish with this on Friday.

I can still reproduce the issue, and I think the problem is elsewhere:
The problem is around `log_matcher_replace` and not `log_matcher_match` (.e.g `subst` rewrite rule).

Even though the input value is now protected in `log_matcher_pcre_re_feed_backrefs` where matches (e.g. $1) are set from capturing groups, there is another place where this protection is needed: where the actual replace happens (and non-matching parts are appended).
Currently, it uses a reference to `$1` which was just overwritten by the capture group in the matching pattern.

This is the affected part:
https://github.com/syslog-ng/syslog-ng/blob/2021180380f6c27d16746e70f8c7a89d36625604/lib/logmatcher.c#L658-L667

I've drafted a unit test in `test_matcher.c`, I'll send you a PR in a few minutes.

I've also a minimal config that can reproduce the issue:
```
log {
  source { network(ip("localhost") port(1234) flags("no-parse")); };
  parser { csv-parser(delimiters(' ') columns('1' '2' '3') flags('greedy')); };
  rewrite { subst('^(\w)' '' value('1') type('pcre')); };
  destination {
    file("/dev/stdout"
         template("$DATE $HOST $PROGRAM:$MSG - $(format-json testmatch --scope dot-nv-pairs $0=$0 $1=$1 MATCHES=list($*) MATCH_NUM=$(list-count $*) )\n"));
  };
};
```

Send e.g. the `foo bar baz` input:
output:
```
May 16 12:32:16 localhost :foo bar baz - {"MATCH_NUM":"1","MATCHES":["\\x00o"],"$1":"","$0":"f"}
```
expected (checked by using $MSG as value() in the `subst()` rewrite rule):
```
May 16 12:34:13 localhost :oo bar baz - {"MATCH_NUM":"1","MATCHES":["f"],"$1":"f","$0":"f"}
```

gaborznagy:
I've opened https://github.com/bazsi/syslog-ng/pull/15 as a draft. It just demonstrates the issue, feel free to change it, or you can close it as well.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

tests/unit/test_matcher.c
```diff
@@ -337,23 +339,22 @@ Test(matcher, test_replace_works_on_matches_input_with_store_matches_too)
...
-  GlobalConfig *cfg = cfg_new_snippet();
-  replace_template = log_template_new(cfg, NULL);
+  replace_template = log_template_new(configuration, NULL);
   cr_assert(log_template_compile(replace_template, replace_pattern, NULL));
 
   NVHandle input_handle = log_msg_get_value_handle("1");
   const gchar *input = log_msg_get_value(msg, input_handle, &value_len);
 
+  NVTable *payload = nv_table_ref(msg->payload);
   const gchar *result = log_matcher_replace(m, msg, input_handle, input, value_len,
```
gaborznagy, line 349: Thanks for the fixup! I'm sorry, but I forgot to free the result string from `log_matcher_replace`...

tests/unit/test_matcher.c
```diff
@@ -337,23 +339,22 @@ Test(matcher, test_replace_works_on_matches_input_with_store_matches_too)
...
   cr_assert_arr_eq(result, expected_result, strlen(expected_result),
-                   "replace failed; result: %s (length %ld), expected: %s (length %d)",
+                   "replace failed; result: %s (length %ld), expected: %s (length %ld)",
                    result, result_len, expected_result, strlen(expected_result));
 
-  scratch_buffers_reclaim_allocations();
   log_template_unref(replace_template);
   log_matcher_unref(m);
   log_msg_unref(msg);
-  cfg_free(cfg);
```
gaborznagy: shouldn't this move to the teardown() function as well?

lib/logmatcher.c
```diff
@@ -675,7 +675,7 @@ log_matcher_pcre_re_replace(LogMatcher *s, LogMessage *msg, gint value_handle, c
   if (new_value)
     {
       /* append the last literal */
-      g_string_append_len(new_value, &value[last_offset], value_len - last_offset);
+      g_string_append_len(new_value, &result.source_value[last_offset], value_len - last_offset);
```
gaborznagy, line 678: I really like this fix! 

One minor thing: can you replace `value_len` with `result.source_value_len` in this function, please so it will be consistent?

lib/logmatcher.c
```diff
@@ -662,7 +662,7 @@ log_matcher_pcre_re_replace(LogMatcher *s, LogMessage *msg, gint value_handle, c
           if (!new_value)
             new_value = g_string_sized_new(value_len);
```
gaborznagy, line 663: Can you replace `value_len` with `result.source_value_len` here too, please?

gaborznagy (commented):
Thanks for the latest fix about the overwritten input values! I've checked this branch against my cisco-parser PR #3934 and it works! \o/

I've checked most of the patches.
I still need to review the LogTemplate-related patches and the LogMatcher wrapper refactors, nevertheless I think this can go in soon!

tests/unit/test_matcher.c
bazsi: done.

tests/unit/test_matcher.c
bazsi: fixed.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

news/feature-3948.md
```diff
@@ -0,0 +1,10 @@
+`set-matches()` and `unset-matches()`: these new rewrite operations allow
+the setting of match variables ($1, $2, ...) in a single operation, based
+on a syslog-ng list expression.
+
+Example:
+  # set $1, $2 and $3 respectively
+  set-matches("foo,bar,baz")
+
+  # likewise, but using a list function
+  set-matches("$(explode ':' 'foo:bar:baz')
```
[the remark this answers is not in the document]
bazsi: fixed in the last iteration.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
I have found yet another case that clobbers the borrowed input of LogMatcher instances, which caused  the change in
#3934 to fail. Added some more unit tests and a fix too.

kira-syslogng:
Build SUCCESS

OverOrion (approved):
LGTM!

Also thank your for not only addressing the review notes, but writing additional tests for those as well!

gaborznagy (approved):
Approved. This was a very long time and lot of efforts for all of us, but I think it really worth it.
Sorry for taking it so long. :)

Otherwise I would ask to squash some patches (especially that some feature took opposite directions), but it would require a lot of effort again from all of us.
The patches are self-contained (not fixup patches), so we can merge it in this state.

PR #3958, opened 2022-03-23
afstomp: always use "0" in the receipt header

alltilla:
After 3d124fb the uniqueness of the `seq_num` field is not guaranteed.
In theory this could cause a problem in the STOMP communication, but with the current implementation it works correctly.

With this commit I intend to emphasize that we do not need a unique ID in the receipt header by using the same one each time. I also added a comment which explains why it is not a problem currently, and how it could cause a problem in the future.

------------------------------------------------------------------------

Explanation:

We check the server's response before sending a new frame to it.
Because of this, we do not need a unique receipt header.

This changes if multiple workers and/or batching support is introduced.
Make sure to use a unique receipt-id if one of the above gets implemented.

------------------------------------------------------------------------

Useful links:

https://stomp.github.io/stomp-specification-1.2.html#Header_receipt
https://stomp.github.io/stomp-specification-1.2.html#RECEIPT

------------------------------------------------------------------------

No news entry needed IMO.

### DIFF
--- a/modules/afstomp/afstomp.c
+++ b/modules/afstomp/afstomp.c
@@ -279,7 +279,6 @@ afstomp_worker_publish(STOMPDestDriver *self, LogMessage *msg)
   GString *body = NULL;
   stomp_frame frame;
   stomp_frame recv_frame;
-  gchar seq_num[16];
 
   if (!self->conn)
     {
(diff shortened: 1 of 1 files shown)


MrAnno (approved):

kira-syslogng:
Build SUCCESS

modules/afstomp/afstomp.c
```diff
@@ -296,8 +295,14 @@ afstomp_worker_publish(STOMPDestDriver *self, LogMessage *msg)
...
+       * We check the server's response before sending a new frame to it.
+       * Because of this, we do not need a unique receipt header.
+       *
+       * This changes if multiple workers and/or batching support is introduced.
+       * Make sure to use a unique receipt-id if one of the above gets implemented.
+       */
+      stomp_frame_add_header(&frame, "receipt", "0");
     };
 
   LogTemplateEvalOptions options = {&self->template_options, LTZ_SEND, self->super.worker.instance.seq_num, NULL, LM_VT_STRING};
```
gaborznagy, line 308: Should we set `seq_num` in the `LogTemplateEvalOptions` to 0 too so `$SEQNUM` macro will be 0 consistently?

alltilla: I would not do that, because maybe someone would need the actual `$SEQNUM` in the body of the message.

IMO the fact that previously `seq_num` was a unique value, and that we needed a unique value for each message in the `receipt` header, was only a coincidence. Both `seq_num` and `receipt` header have their own uses, `seq_num` is a message metadata, `receipt` is used for acknowledgement in the STOMP protocol.

With this commit I only wanted to remove the coincidental connection between the two.

gaborznagy (approved):

PR #3962, opened 2022-03-24
Support template expressions in set-tag()

### LINKED ISSUE #3956: Set Tags does not evaluate macros
## Description of the problem
I realize I am using tags slightly outside of the original design. In my use case I am sending tags to the next destination and would like to pass some dynamic tags. For example 
```c
rewrite { set-tag('wireport:${DESTPORT}'); };
```

bazsi:
Fixes #3956 and adds both light and unit test coverage.

Test config:
```
@version: 3.36

rewrite foo {
	set-tag("dyn::$PROGRAM");
};

log {
	source { tcp(port(2000)); };
	rewrite(foo);
	rewrite { set-tag("dyn::$HOST"); };
	rewrite(foo);
};
```

The result of the above config is that we get trace messages on the tags being set properly.

### DIFF
--- a/lib/rewrite/rewrite-expr-grammar.ym
+++ b/lib/rewrite/rewrite-expr-grammar.ym
@@ -136,16 +136,16 @@ rewrite_expr
             last_rewrite = log_rewrite_rename_new(configuration, from, to);
           }
           rewrite_rename_opts ')'              { $$ = last_rewrite; }
-	| KW_SET_TAG '(' string 
+	| KW_SET_TAG '(' template_content
 	  { 
 	    last_rewrite = log_rewrite_set_tag_new($3, TRUE, configuration);
-            free($3);
+            log_template_unref($3);
           }
 	  rewrite_settag_opts ')'             	{ $$ = last_rewrite; }
-	| KW_CLEAR_TAG '(' string
+	| KW_CLEAR_TAG '(' template_content
           {
             last_rewrite = log_rewrite_set_tag_new($3, FALSE, configuration);
-            free($3);
+            log_template_unref($3);
           }
           rewrite_settag_opts ')'               { $$ = last_rewrite; }
         | KW_GROUP_SET '(' template_content
--- a/lib/rewrite/rewrite-set-tag.c
+++ b/lib/rewrite/rewrite-set-tag.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2002-2013 Balabit
  * Copyright (c) 1998-2013 Balázs Scheidler
+ * Copyright (c) 2022 Balazs Scheidler <bazsi77@gmail.com>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -23,6 +24,8 @@
  */
 
 #include "rewrite-set-tag.h"
+#include "template/templates.h"
+#include "scratch-buffers.h"
 
 /* LogRewriteSetTag
  *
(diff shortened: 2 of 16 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/rewrite/rewrite-set-tag.c
```diff
@@ -35,45 +38,106 @@ struct _LogRewriteSetTag
...
-  cloned->value = self->value;
   return &cloned->super.super;
 }
 
+static void
+_free(LogPipe *s)
+{
+  LogRewriteSetTag *self = (LogRewriteSetTag *) s;
+
+  log_template_unref(self->tag_template);
```
alltilla, line 112: This `unref()` is does not have its `ref()` counterpart in `new()` if `log_template_is_literal_string(tag_template)` is true.

I had another comment in the grammar file about a `ref()`-`unref()` inconsistency. I am starting to get a feeling that we should consistently `ref()` the template in `new()`, even if we will not use it later.

lib/rewrite/rewrite-set-tag.c
```diff
@@ -35,45 +38,106 @@ struct _LogRewriteSetTag
   LogRewrite super;
   LogTagId tag_id;
   gboolean value;
+  LogTemplate *tag_template;
+  LogTemplateOptions template_options;
 };
 
+LogTemplateOptions *
+log_rewrite_set_tag_get_template_options(LogRewrite *s)
```
alltilla, line 46: This is missing from the header.

lib/rewrite/rewrite-expr-grammar.ym
```diff
@@ -120,16 +120,18 @@ rewrite_expr
...
+            log_template_unref($3);
           }
 	  rewrite_settag_opts ')'             	{ $$ = last_rewrite; }
-	| KW_CLEAR_TAG '(' string
+	| KW_CLEAR_TAG '(' template_content
           {
             last_rewrite = log_rewrite_set_tag_new($3, FALSE, configuration);
-            free($3);
+            last_template_options = log_rewrite_set_get_template_options(last_rewrite);
+            log_template_unref($3);
```
alltilla, line 134: If the template is a literal string, as we do not `ref()` it in `new()`, it will be `free()`d here, making `last_template_options` invalid.

lib/rewrite/rewrite-expr-grammar.ym
```diff
@@ -120,16 +120,18 @@ rewrite_expr
             last_rewrite = log_rewrite_rename_new(configuration, from, to);
           }
           rewrite_rename_opts ')'              { $$ = last_rewrite; }
-	| KW_SET_TAG '(' string 
+	| KW_SET_TAG '(' template_content
 	  { 
 	    last_rewrite = log_rewrite_set_tag_new($3, TRUE, configuration);
-            free($3);
+            last_template_options = log_rewrite_set_get_template_options(last_rewrite);
```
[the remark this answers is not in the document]
bazsi: Good catch.

lib/rewrite/rewrite-set-tag.c
bazsi: If tag_template is literal, we won't even save the template, it will be null, which unref will ignore.

lib/rewrite/rewrite-set-tag.c
bazsi: The original typo caused this not to be noticed. Thanks.

lib/rewrite/rewrite-expr-grammar.ym
bazsi: It won't invalidate that. Template options is embedded in the rewrite rule.

lib/rewrite/rewrite-set-tag.c
bazsi: I've decided to drop the ability to set template options from set-tag(), as this feature is only supported in set() and not
in any other rewrite operation (e.g. set-facility, set-pri, etc).

On the other hand I've found an issue in set(), which I've fixed in this branch as a followup.

lib/rewrite/rewrite-set-tag.c
bazsi: This is also resolved, as I've removed template_options from set-tag().

kira-syslogng:
Build FAILURE

lib/rewrite/rewrite-expr-grammar.ym
alltilla: Ah, I see. I mixed up `template` and `template_options`.

bazsi:
Thanks @talien for the unit test on groupset() you wrote in 2014. 

It saved me this time, as I almost broke groupset() with a seemingly innocent change. There are lots of things to dislike about our test suite, but we are much better off with them. :)

bazsi:
Does anyone have an idea why github actions did not run on this PR?

alltilla:
No idea. o.O

bazsi:
When I opened a second pull request with the same content, this started too. I still don't get it. The branch does not touch the .github directory at all.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build FAILURE

gaborznagy:
@bazsi **kira** is failing due to the [fix for a git vulnerability CVE-2022-24765
](https://github.blog/2022-04-12-git-security-vulnerability-announced/)

We are working on a proper solution to adjust the permissions kira uses, but if nothing else remains we will just add the `safe.directory` config as a workaround.

bazsi:

kira-syslogng:
Build SUCCESS

lib/template/templates.c
```diff
@@ -319,6 +319,24 @@ log_template_options_init(LogTemplateOptions *options, GlobalConfig *cfg)
...
 
+void
+log_template_options_clone(LogTemplateOptions *source, LogTemplateOptions *dest)
+{
+  dest->ts_format = source->ts_format;
+  for (gint i = 0; i < LTZ_MAX; i++)
+    {
+      if (source->time_zone[i])
+        dest->time_zone[i] = g_strdup(source->time_zone[i]);
+    }
```
OverOrion, line 330: Shouldn't we clone `time_zone_info` too?

bazsi: No, this has been a while ago, but I remember that this clone function clones the options structure in an "uninitialized state", so these timezone settings would be used to instantiate the TimeZoneInfo class, meaning that those would be unshared, but I don't feel that's a problem in this case.

kira-syslogng:
Build SUCCESS

OverOrion (approved):
LGTM!

kira-syslogng:
Build SUCCESS

MrAnno (approved):
LGTM.

OverOrion (approved):
LGTM!

PR #3966, opened 2022-03-28
Pipeline processing on separate thread

bazsi:
This is a continuation of #3969 where sources with dedicated source threads (e.g. LogThreadedSourceDriver or AFInternalSourceDriver) get capable of doing source batching.

This branch introduces LogScheduler, a helper class to perform thread-switch in the course of a LogMessage processing
in the pipeline. This way the reading of the source messages can happen in a thread separate from where it is filtered/processed (and also separate from destination processing).

This increased syslog-ng performance for single-connection inputs to these levels:

average rate = 627821.45 msg/sec, count=6280000, time=10.0028, (average) msg size=256, bandwidth=156955.36 kB/sec

Here's the NEWS entry, with examples how to use it and more description.

`parallelize()` support for pipelines: syslog-ng has traditionally performed
processing of log messages arriving from a single connection sequentially.
This was done to ensure message ordering as well as most efficient use of
CPU on a per message basis.  This mode of operation is performing well as
long as we have a relatively large number of parallel connections, in which
case syslog-ng would use all the CPU cores available in the system.

In case only a small number of connections deliver a large number of
messages, this behaviour may become a bottleneck.

With the new parallelization feature, syslog-ng gained the ability to
re-partition a stream of incoming messages into a set of partitions, each of
which is to be processed by multiple threads in parallel.  This does away
with ordering guarantees and adds an extra per-message overhead. In exchange
it will be able to scale the incoming load to all CPUs in the system, even
if coming from a single, chatty sender.

To enable this mode of execution, use the new parallelize() element in your
log path:

```
log {
	source {
		tcp(
			port(2000)
			log-iw-size(10M) max-connections(10) log-fetch-limit(100000)
		);
	};
	parallelize(partitions(4));

	# from this part on, messages are processed in parallel even if
	# messages are originally coming from a single connection

	parser { ... };
	destination { ... };
};
```

The config above will take all messages emitted by the tcp() source and push
the work to 4 parallel threads of execution, regardless of how many
connections were in use to deliver the stream of messages to the tcp()
driver.

parallelize() uses round-robin to allocate messages to partitions by default.
You can however retain ordering for a subset of messages with the
partition-key() option.

You can use partition-key() to specify a message template. Messages that
expand to the same value are guaranteed to be mapped to the same partition.

For example:

```
log {
	source {
		tcp(
			port(2000)
			log-iw-size(10M) max-connections(10) log-fetch-limit(100000)
		);
	};
	parallelize(partitions(4) partition-key("$HOST"));

	# from this part on, messages are processed in parallel if their
	# $HOST value differs. Messages with the same $HOST will be mapped
	# to the same partition and are processed sequentially.

	parser { ... };
	destination { ... };
};
```


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

lib/logscheduler.c
```diff
@@ -0,0 +1,429 @@
...
+
+          LogMessage *msg = log_msg_ref(node->msg);
+
+          LogPathOptions path_options = LOG_PATH_OPTIONS_INIT;
+          path_options.ack_needed = node->ack_needed;
+          path_options.flow_control_requested = node->flow_control_requested;
+
+          log_msg_free_queue_node(node);
+
+          log_msg_refcache_start_consumer(msg, &path_options);
```
MrAnno, line 98: Why do we call the consumer version of the refcache here?

lib/logscheduler.c
```diff
@@ -0,0 +1,429 @@
...
+
+static gpointer
+_flush_batch(gpointer s)
+{
+  LogScheduler *self = (LogScheduler *) s;
+
+  gint thread_index = main_loop_worker_get_thread_index();
+  g_assert(thread_index >= 0);
+
+  LogSchedulerThreadState *thread_state = &self->thread_states[thread_index];
```
MrAnno, line 227: `thread_index` sometimes can be higher than `sizeofarray(thread_states)`.

lib/logscheduler.c
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+  LogSchedulerThreadState *thread_state = &self->thread_states[thread_index];
+  _queue_thread(self, thread_state, msg, path_options);
+}
+
+LogScheduler *
+log_scheduler_new(LogSchedulerOptions *options, LogPipe *front_pipe)
+{
+  gint max_threads = main_loop_worker_get_max_number_of_threads();
```
MrAnno, line 337: This includes destination threads as well.

Shouldn't we allocate thread state objects only for input threads?

MrAnno (commented):
I have (hopefully) thoroughly reviewed this PR and have not found anything weird around thread boundaries.

During the tests, however, I found something quite scary:
Flow-control and queuing do not seem to work correctly in certain high-performance cases.
This turned out to be independent of this PR, it even has traces in syslog-ng v3.14. I am investigating this separately.

However, I could not reach the ~600000 single-connection EPS with `parallelize()`.
My tests showed around 300000 EPS both with and without parallelize. I could only get a 2x multiplier with a lot of json-parser().

Could you please share the configuration you used for testing?

MrAnno:
Note: This PR depends on #4476, where a race condition may still exist.

lib/logscheduler-pipe.c
```diff
@@ -0,0 +1,88 @@
...
+_init(LogPipe *s)
+{
+  LogSchedulerPipe *self = (LogSchedulerPipe *) s;
+  GlobalConfig *cfg = log_pipe_get_config(s);
+
+  if (!log_scheduler_options_init(&self->scheduler_options, cfg))
+    return FALSE;
+
+  if (!self->scheduler)
+    self->scheduler = log_scheduler_new(&self->scheduler_options, self->super.pipe_next);
```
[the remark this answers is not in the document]
bazsi: This is called from log_pipe_init(), which is _after_ pre_init. Or am I missing something?

lib/logscheduler.c
bazsi: The LogScheduler is not a producer of log messages, so we can't use that. While there can only be one producer, there can be multiple consumers.

maybe the naming for consumer mode is not perfect. it applies to anything that takes over messages from another thread for processing.

lib/logscheduler.c
```diff
@@ -0,0 +1,429 @@
...
+
+void
+_partition_clear(LogSchedulerPartition *partition)
+{
+  g_mutex_clear(&partition->batches_lock);
+}
+
+/* LogSchedulerThreadState */
+
+static glong
```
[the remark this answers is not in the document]
bazsi: I changed everything to guint in the call chain.

lib/logscheduler.c
```diff
@@ -0,0 +1,429 @@
...
+{
+  if (!self->options->partition_key)
+    {
+      gint partition_index = thread_state->last_partition;
+      thread_state->last_partition = (thread_state->last_partition + 1) % self->options->num_partitions;
+      return partition_index;
+    }
+  else
+    {
+      return _get_template_hash(self->options->partition_key, msg) % self->options->num_partitions;
```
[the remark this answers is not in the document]
bazsi: done.

lib/logscheduler.c
bazsi: how? the _flush_batch() callback is associated with a specific thread in _queue_thread(), which in turn is called by log_scheduler_push() which validates thread_index against self->num_threads (and against 0).

lib/logscheduler.c
bazsi: I am starting to think that we should, yeah. I got rid off using separate ID space for destinations not long ago.

Our per-thread state is 48 bytes at the moment, so with 64 threads in total, it's 3k of memory per LogScheduler.

I don't think this is a pressing issue for now. We also have the InputQueue for LogQueueFifo, which is 64 bytes and is also a per-thread struct.

If we were to run on 256+ threads we might want to do something  with this.

bazsi:
I dropped loggen patches from this PR.

lib/logscheduler-pipe.c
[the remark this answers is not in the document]
MrAnno: No, I just wanted to note that both `log_scheduler_new()` and `log_queue_fifo_new()` have some hidden requirement: they can only be called after pre_init.

lib/logscheduler.c
MrAnno: Yeah, I was wrong. Thank you.

I just saw that we use `main_loop_worker_get_thread_index()` without validating against `self->num_threads`.

MrAnno (approved):

bazsi:
\o/

PR #3969, opened 2022-03-30
Add scalable FIFO support to LogThreadedSource based sources

bazsi:
This is a spin-off from #3966 which merges the thread ID space for worker threads and associates a thread_id even with
threads spawned by LogThreadedSourceDriver. This will enable the scalable input processing in LogQueueFifo and makes a few things simpler.

It also increases the number of supported threads to 256 which is becoming a simple compile-time constant, instead of a constraint imposed by the underlying data structure (the previous limit of 64 being the number of bits in a guint64).

While #3966 still depends on an ivykis change (https://github.com/buytenh/ivykis/issues/24) but I don't want to stall the commits that make sense even without the ivykis dependency.

### DIFF
--- a/lib/logqueue-fifo.c
+++ b/lib/logqueue-fifo.c
@@ -659,9 +659,9 @@ LogQueue *
 log_queue_fifo_new(gint log_fifo_size, const gchar *persist_name)
 {
   LogQueueFifo *self;
-  gint i;
 
-  self = g_malloc0(sizeof(LogQueueFifo) + log_queue_max_threads * sizeof(self->input_queues[0]));
+  gint max_threads = main_loop_worker_get_max_number_of_threads();
+  self = g_malloc0(sizeof(LogQueueFifo) + max_threads * sizeof(self->input_queues[0]));
 
   log_queue_init_instance(&self->super, persist_name);
   self->super.type = log_queue_fifo_type;
@@ -678,8 +678,8 @@ log_queue_fifo_new(gint log_fifo_size, const gchar *persist_name)
 
   self->super.free_fn = log_queue_fifo_free;
 
-  self->num_input_queues = log_queue_max_threads;
-  for (i = 0; i < num_input_queues; i++)
+  self->num_input_queues = max_threads;
+  for (gint i = 0; i < self->num_input_queues; i++)
     {
       INIT_IV_LIST_HEAD(&self->input_queues[i].items);
       worker_batch_callback_init(&self->input_queues[i].cb);
--- a/lib/logqueue-fifo.c
+++ b/lib/logqueue-fifo.c
@@ -101,6 +101,7 @@ typedef struct _LogQueueFifo
   /* legacy: flow-controlled messages are included in the log_fifo_size limit */
   gboolean use_legacy_fifo_size;
 
+  gint num_input_queues;
   InputQueue input_queues[0];
 } LogQueueFifo;
 
@@ -152,7 +153,7 @@ log_queue_fifo_is_empty_racy(LogQueue *s)
   else
     {
       gint i;
-      for (i = 0; i < log_queue_max_threads && !has_message_in_queue; i++)
+      for (i = 0; i < self->num_input_queues && !has_message_in_queue; i++)
         {
           has_message_in_queue |= self->input_queues[i].finish_cb_registered;
         }
@@ -343,7 +344,11 @@ log_queue_fifo_push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *pat
 
   thread_id = main_loop_worker_get_thread_id();
 
-  g_assert(thread_id < 0 || log_queue_max_threads > thread_id);
+  /* if this thread has an ID than the number of input queues we have (due
+   * to a config change), handle the load via the slow path */
+
+  if (thread_id >= self->num_input_queues)
+    thread_id = -1;
 
   /* NOTE: we don't use high-water marks for now, as log_fetch_limit
    * limits the number of items placed on the per-thread input queue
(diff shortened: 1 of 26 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
Rebased against current master.

kira-syslogng:
Build FAILURE

bazsi:

mitzkia:

kira-syslogng:
Build FAILURE

mitzkia:

lib/logqueue-fifo.c
```diff
@@ -673,7 +678,8 @@ log_queue_fifo_new(gint log_fifo_size, const gchar *persist_name)
 
   self->super.free_fn = log_queue_fifo_free;
 
-  for (i = 0; i < log_queue_max_threads; i++)
+  self->num_input_queues = log_queue_max_threads;
+  for (i = 0; i < num_input_queues; i++)
```
MrAnno: This variable is undeclared.

lib/logqueue-fifo.c
```diff
@@ -343,7 +344,11 @@ log_queue_fifo_push_tail(LogQueue *s, LogMessage *msg, const LogPathOptions *pat
 
   thread_id = main_loop_worker_get_thread_id();
 
-  g_assert(thread_id < 0 || log_queue_max_threads > thread_id);
+  /* if this thread has an ID than the number of input queues we have (due
+   * to a config change), handle the load via the slow path */
+
+  if (thread_id >= self->num_input_queues)
+    thread_id = 0;
```
MrAnno, line 351: [change-request] I believe this should be `-1` in order to trigger the slow path (locked per-message push).
Pushing from any non-thread0 thread to thread0's queue would not be safe as far as I see.

lib/logqueue-fifo.c
```diff
@@ -678,8 +678,8 @@ log_queue_fifo_new(gint log_fifo_size, const gchar *persist_name)
 
   self->super.free_fn = log_queue_fifo_free;
 
-  self->num_input_queues = log_queue_max_threads;
-  for (i = 0; i < num_input_queues; i++)
+  self->num_input_queues = max_threads;
+  for (gint i = 0; i < self->num_input_queues; i++)
```
MrAnno, line 682: The undeclared variable I previously mentioned is fixed in this commit.

MrAnno (changes requested):
I've marked my important note with `[change-request]`.

lib/logqueue-fifo.c
bazsi: You are right. That should be -1.

There's a -1 dance scattered everywhere in the code, sometimes the "id" is already a zero-based value, sometimes is 1 based.

I am thinking about naming one an "id" (the 1 based one), and the other "index" (the zero based one), so it's easy to understand which one is which.

lib/mainloop-worker.c
```diff
@@ -63,33 +63,47 @@ static struct iv_task main_loop_workers_reenable_jobs_task;
 /* thread ID allocation */
 static GMutex main_loop_workers_idmap_lock;
 
-static guint64 main_loop_workers_idmap;
+#define MAIN_LOOP_IDMAP_BITS_PER_ROW    (sizeof(guint64)*8)
+#define MAIN_LOOP_IDMAP_ROWS            (MAIN_LOOP_MAX_WORKER_THREADS / MAIN_LOOP_IDMAP_BITS_PER_ROW)
+
+static guint64 main_loop_workers_idmap[MAIN_LOOP_IDMAP_ROWS + 1];
```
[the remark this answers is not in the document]
bazsi: I am not sure. It's been a while. Maybe related to the same "+1" dance we have with IDs.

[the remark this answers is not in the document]
bazsi: I think this was related to the case where the number of threads is not divisable by 64, in which case we would need a "final" element in this array, not completely used.

As I see I added a G_STATIC_ASSERT to validate that our MAX_THREADS value is indeed dividable by 64.

So this is not needed.

bazsi:
I have just added two fixup commits that should address the review notes above. I also made an attempt to clarify thread_id vs. thread_index.

I did not fix the ordering issue with the undeclared variable though.

kira-syslogng:
Build FAILURE

lib/mainloop-worker.c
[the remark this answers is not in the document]
MrAnno: Thanks.

MrAnno:
LGTM.

MrAnno:
A rebase+squash will fix the CI and we can hopefully merge it.
@HofiOne will be available next week, we'll fix Kira then.

bazsi:
rebased and squashed.

kira-syslogng:
Build FAILURE

MrAnno:
There is still a fixup commit in the history.

It seems the commit check is broken, it shows success incorrectly. I'm fixing it.

bazsi:
Sorry, I needed to move the bank port patch earlier as autosquash failed at
first. Let me do that again.

kira-syslogng:
Build FAILURE

MrAnno (approved):

MrAnno:
The macOS runner seems to have been updated and now it contains a broken net-snmp pc file. I'm working on a fix.

Issue #3967, opened 2022-03-29
Trace /Debug usability improvements


rfaircloth-splunk:
## Description of the problem
Troubleshooting parser errors in the field can be difficult. The following basic enhancements would help make trace and debug work easier.

* Use the uniqueid of the message in process on the debug/trace statement
* Allow "modular" trace and debug flags at startup rather than "d or t" "source=d logpath=d dest=d" this can avoid expensive (volumeus) output for destination or logpath when only source side parsers are in question

PR #3972, opened 2022-04-03
add rcptid to message related trace messages

bazsi:
The new log_msg_evt_tags() macro adds the rcptid value to the trace messages
making it easier to identify which msg we are manipulating at the moment.

In order to have a non-zero value for rcptid, one needs to enable
use-rcptid() in the configuration.

Fixes portions of #3967

### DIFF
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -504,6 +504,11 @@ cfg_new(gint version)
 
   self->use_uniqid = FALSE;
 
+  /* The rcptid part of use_uniqid() is used in trace messages, enable it
+   * explicitly.  We don't care about performance if trace is enabled.  */
+  if (trace_flag)
+    self->use_uniqid = TRUE;
+
   stats_options_defaults(&self->stats_options);
 
   self->min_iw_size_per_reader = 100;
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -454,7 +454,7 @@ log_reader_handle_line(LogReader *self, const guchar *line, gint length, LogTran
             evt_tag_printf("input", "%.*s", length, line),
             evt_tag_msg_reference(m));
 
-  msg_format_parse(&self->options->parse_options, m, line, length);
+  msg_format_parse_into(&self->options->parse_options, m, line, length);
 
   _log_reader_insert_msg_length_stats(self, length);
   if (aux)
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -449,7 +449,7 @@ log_reader_handle_line(LogReader *self, const guchar *line, gint length, LogTran
 {
   LogMessage *m;
 
-  m = log_msg_new_empty();
+  m = msg_format_construct_message(&self->options->parse_options, line, length);
   msg_debug("Incoming log entry",
             evt_tag_printf("input", "%.*s", length, line),
             evt_tag_msg_reference(m));
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -449,11 +449,12 @@ log_reader_handle_line(LogReader *self, const guchar *line, gint length, LogTran
 {
   LogMessage *m;
 
+  m = log_msg_new_empty();
   msg_debug("Incoming log entry",
-            evt_tag_printf("line", "%.*s", length, line));
-  /* use the current time to get the time zone offset */
-  m = log_msg_new((gchar *) line, length,
-                  &self->options->parse_options);
+            evt_tag_printf("input", "%.*s", length, line),
+            evt_tag_msg_reference(m));
+
+  msg_format_parse(&self->options->parse_options, m, line, length);
 
   _log_reader_insert_msg_length_stats(self, length);
   if (aux)
--- a/lib/filter/filter-call.c
+++ b/lib/filter/filter-call.c
@@ -52,7 +52,7 @@ filter_call_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg, LogTemplate
 
   msg_trace("filter() evaluation started",
             evt_tag_str("called-rule", self->rule),
-            evt_tag_printf("msg", "%p", msgs[num_msg - 1]));
+            evt_tag_msg_reference(msgs[num_msg - 1]));
 
   return res ^ s->comp;
 }
--- a/lib/filter/filter-cmp.c
+++ b/lib/filter/filter-cmp.c
@@ -93,7 +93,7 @@ fop_cmp_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg, LogTemplateEval
             evt_tag_str("left", left_buf->str),
             evt_tag_str("operator", self->super.type),
             evt_tag_str("right", right_buf->str),
-            evt_tag_printf("msg", "%p", msgs[num_msg - 1]));
+            evt_tag_msg_reference(msgs[num_msg - 1]));
 
   scratch_buffers_reclaim_marked(marker);
   return result ^ s->comp;
--- a/lib/filter/filter-in-list.c
+++ b/lib/filter/filter-in-list.c
@@ -52,7 +52,7 @@ filter_in_list_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg, LogTempl
   gboolean result = (g_tree_lookup(self->tree, value) != NULL);
   msg_trace("in-list() evaluation started",
             evt_tag_str("value", value),
-            evt_tag_printf("msg", "%p", msg));
+            evt_tag_msg_reference(msg));
 
   return result ^ s->comp;
 }
--- a/lib/filter/filter-netmask.c
+++ b/lib/filter/filter-netmask.c
@@ -67,7 +67,7 @@ filter_netmask_eval(FilterExprNode *s, LogMessage **msgs, gint num_msg, LogTempl
             evt_tag_inaddr("msg_address", addr),
             evt_tag_inaddr("address", &self->address),
             evt_tag_inaddr("netmask", &self->netmask),
-            evt_tag_printf("msg", "%p", msg));
+            evt_tag_msg_reference(msg));
   return res ^ s->comp;
 }
 
--- a/lib/filter/filter-netmask6.c
+++ b/lib/filter/filter-netmask6.c
@@ -132,7 +132,7 @@ _eval(FilterExprNode *s, LogMessage **msgs, gint num_msg, LogTemplateEvalOptions
             evt_tag_inaddr6("msg_address", address),
             evt_tag_inaddr6("address", &self->address),
             evt_tag_int("prefix", self->prefix),
-            evt_tag_printf("msg", "%p", msg));
+            evt_tag_msg_reference(msg));
   return result ^ s->comp;
 }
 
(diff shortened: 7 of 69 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

gaborznagy:
I'm not entirely convinced that adding `rcptid` to trace messages would help much by itself.
Trace messages are useful when only one message is sent to syslog-ng to debug the route a specific message takes.
For one message, an additional `rcptid` field to the trace message is redundant next to the address of the LogMessage address.

Of course, when a copy-on-write takes place that address is changed, so this is the scenario where `rcptid` would be useful, but then should we use the address of the LogMessage too?

And wouldn't be more convenient if trace level would turn on `rcptid` too ? (only from a user's point of view, as on code level they are very distant).

What do you think @rfaircloth-splunk, @bazsi?

gaborznagy:
Just a note about what we discussed:
- we could enable `rcptid` by default, if it doesn't have significant negative effects on performance. We'll do a perf test on this.
- we agreed that `rcptid` in the trace messages is better than the address of the LogMessage object.

Kokan:
Just out of curiosity did some testing:

```
use_rcptid(no)
center;;received;a;processed;32729670
center;;received;a;processed;32791101
center;;received;a;processed;33424028
center;;received;a;processed;32887275
center;;received;a;processed;31181524
; (32729670+32791101+33424028+32887275+31181524)/5
        32602719.6

use_rcptid(yes)
center;;received;a;processed;26653406
center;;received;a;processed;27627752
center;;received;a;processed;29596720
center;;received;a;processed;29943992
center;;received;a;processed;30723320
; (26653406+27627752+29596720+29943992+30723320)/5
        28909038

; 28909038/32602719.6
        ~0.88670633476846514363
```

The configuration:
```
@version: 3.35

options {
#use_rcptid(yes);
};

source s {
    example_msg_generator (
        freq(0.00001)
        template("test message")
    );
    example_msg_generator (
        freq(0.00001)
        template("test message#2")
    );
    example_msg_generator (
        freq(0.00001)
        template("test message#2")
    );
};

log {
    source(s);
destination { file("/dev/null"); };
};
```

Running: `syslog-ng -Fe & ; sleep 60 ; syslog-ng-ctl stats|grep 'center;;received' >> nouniqid ; syslog-ng-ctl stop`

Based on this, it does not look like promising.

kira-syslogng:
Build SUCCESS

mitzkia:
@kira-syslogng do stresstest this please test branch=micek_perf_with_rcptid

mitzkia:
@kira-syslogng do stresstest branch=micek_perf_with_rcptid

kira-syslogng:
Build SUCCESS

kira-syslogng:

mitzkia:
I have also run an internal perf test for enabled rcptid and it shows a ~30 percent decrease in our fastest combination. 
- it was a 5 minutes run
- config
```bash
options {
    keep_hostname(yes);
    keep_timestamp(no);
    use_rcptid(yes);
    stats_level(2);
    use_dns(no);
};
source s_network_d7e6b375b62e415f8924fd4d883e0e80 {
network(
    ip(0.0.0.0)
    log_fetch_limit(1000)
    log_iw_size(10000)
    max_connections(10)
    port(61001)
);
};
destination d_file_91b6a8918dc742e98e8f447485517f5f {
file(
    "/tmp/output.$HOST" 
    flush_lines(1000)
    log_fifo_size(20000)
);
};
log {
source(s_network_d7e6b375b62e415f8924fd4d883e0e80);
destination(d_file_91b6a8918dc742e98e8f447485517f5f);
flags(flow-control);
};
```
- loggen was used with the following parameters:
```
bin/loggen --active-connections=10 --dont-parse --inet --interval=300 --loop-reading --quiet --rate=1000000 --read-file=/tmp/loggen_input_file.txt --size=512 [ip] [port]
```

bazsi:
The original idea of rcptid was to be able to reorder the messages to its
reception order, so the reason it is monotonous was this.

I am not sure anyone uses it for this purpose though. Such a number is
useful in cases where you have multiple parallel paths to storage.

Before changing semantics, I'd probably revisit the locking strategy within
rcptid, because I am sure it could be improved.

But I'd merge this as is, maybe with the addition of -t command line option
enabling rcptid as well.

ryanfaircloth:
This PR is exciting. One comment early on in the PR was trace is really only useful for a single message. I think that is a very legacy way of thinking. Using trace on console this is true if I was to get a json blog with a the rcpti field into Splunk, Humio or other observability tools I could much more easily troubleshoot a running problem

bazsi:
I have now added a commit to enable rcptid generation when --trace is passed as a command line option.

Also, I went back and added rcptid to the very first debug message we emit about a new incoming log message.

We can't enable rcptid now due to performance reasons, I'll revisit that in the future.

But with all of this, this can be merged in my opinion.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

mitzkia:

kira-syslogng:
Build SUCCESS

gaborznagy (changes requested):
@bazsi, the local ivykis is still updated in this PR.

Sorry for putting request for change flag on it, but I would like to avoid others accidentally merge this.

Otherwise looks good to me.

bazsi:
Sorry. git seems to add submodules to patches automatically during rebase, if I have it updated locally.

I now removed it explicitly.

gaborznagy (approved):
Thanks, approve!

kira-syslogng:
Build SUCCESS

bazsi:
I have done one last rebase in which I've renamed the macro generating the tags to match conventions, as these functions
tend to be called evt_tag_*()

The trigger was that I was adding another set of these tags and it felt somewhat alien, I did some grepping:

lib/gprocess.c:evt_tag_cap_t(const char *tag, cap_t cap)
lib/value-pairs/evttag.c:evt_tag_value_pairs(const char *key, ValuePairs *vp, LogMessage *msg, LogTemplateEvalOptions *options)

We have some inconsistent names still:

bazsi@bzorp:~/src/syslog-ng/syslog-ng$ git grep ^EVTTAG *.h
lib/cfg-lexer.h:EVTTAG *cfg_lexer_format_location_tag(CfgLexer *self, CFG_LTYPE *yylloc);
lib/cfg-tree.h:EVTTAG *log_expr_node_location_tag(LogExprNode *self);
lib/eventlog/src/evtlog.h:EVTTAG *evt_tag_str(const char *tag, const char *value);
lib/eventlog/src/evtlog.h:EVTTAG *evt_tag_mem(const char *tag, const void *value, size_t len);
lib/eventlog/src/evtlog.h:EVTTAG *evt_tag_int(const char *tag, int value);
lib/eventlog/src/evtlog.h:EVTTAG *evt_tag_long(const char *tag, long long value);
lib/eventlog/src/evtlog.h:EVTTAG *evt_tag_errno(const char *tag, int err);
lib/eventlog/src/evtlog.h:EVTTAG *evt_tag_printf(const char *tag, const char *format, ...) G_GNUC_PRINTF(2, 3);
lib/eventlog/src/evtlog.h:EVTTAG *evt_tag_inaddr(const char *tag, const struct in_addr *addr);
lib/eventlog/src/evtlog.h:EVTTAG *evt_tag_inaddr6(const char *tag, const struct in6_addr *addr);
lib/logpipe.h:EVTTAG *log_pipe_location_tag(LogPipe *pipe);
lib/tlscontext.h:EVTTAG *tls_context_format_tls_error_tag(TLSContext *self);
lib/tlscontext.h:EVTTAG *tls_context_format_location_tag(TLSContext *self);
lib/value-pairs/evttag.h:EVTTAG *evt_tag_value_pairs(const char *key, ValuePairs *vp, LogMessage *msg, LogTemplateEvalOptions *options);

Nevertheless, I like the one with evt_tag prefix better, so I quickly renamed the change in this patchset.

There are no other changes since @gaborznagy approved it, so please reapprove.

lib/cfg.c
```diff
@@ -504,6 +504,11 @@ cfg_new(gint version)
 
   self->use_uniqid = FALSE;
 
+  /* The rcptid part of use_uniqid() is used in trace messages, enable it
+   * explicitly.  We don't care about performance if trace is enabled.  */
+  if (trace_flag)
+    self->use_uniqid = TRUE;
```
MrAnno, line 510: The trace flag can be enabled at runtime using syslog-ng-ctl as well. In that specific case, rcptid will not work.

lib/logreader.c
```diff
@@ -449,11 +449,12 @@ log_reader_handle_line(LogReader *self, const guchar *line, gint length, LogTran
...
+  m = log_msg_new_empty();
   msg_debug("Incoming log entry",
-            evt_tag_printf("line", "%.*s", length, line));
-  /* use the current time to get the time zone offset */
-  m = log_msg_new((gchar *) line, length,
-                  &self->options->parse_options);
+            evt_tag_printf("line", "%.*s", length, line),
+            evt_tag_msg_reference(m));
+
+  msg_format_parse(&self->options->parse_options, m, line, length);
```
MrAnno, line 457: Using `log_msg_new_empty()` + `msg_format_parse()` instead of the original `log_msg_new()` might have some minor performance related side effects (for example, it allocates the payload a bit differently).

Is this change intentional? I've found it in a commit that is meant to introduce rcptid reporting only (`logreader: report rcptid in the "Incoming log entry" message`).

lib/cfg.c
bazsi: I know about that, but I felt that the more common use-case is to start syslog-ng interactively in foreground mode. And rcptid can only initialized at startup at the moment. Not saying that it couldn't be done, but I was fine with this restriction.

lib/logreader.c
bazsi: The change is needed, as it's difficult to report msg related fields in a log message if the message does not exist yet :)

But good point on the allocation difference, I didn't notice that (I should have though). I am fixing that in a moment.

bazsi: I have just pushed another round that fixes the allocation issue you pointed out rightly.

It also fixes a circular dependency between msg-format.{c,h} and logmsg.{c,h} by moving the log_msg_new() functionality to become a method of MsgFormat. It also decreases test program dependency on log message parsing (and thus the syslogformat module) by using lower-level LogMessage functions. The latter is not complete but does away with a few easy-to-get-rid of cases.

bazsi:
@gaborznagy I am afraid the new additions to this branch as described in https://github.com/syslog-ng/syslog-ng/pull/3972#discussion_r865779334 warrant a new  review.

This branch is now a lot more than the rcptid tags in trace messages. The changes should be trivial but are quite sizable. Let me know if you want to split this.

kira-syslogng:
Build SUCCESS

lib/logreader.c
MrAnno: Thank you. 👍🏻

MrAnno (approved):
Thank you for the refactor commits, they look really good.

MrAnno (approved):

kira-syslogng:
Build SUCCESS

gaborznagy (approved):
Sorry for kept you waiting, I've checked the internal build as discussed.
Approve.

PR #3974, opened 2022-04-04
mainloop: add main_loop_get_pending_new_config()

MrAnno:
This method can be used to obtain the pending configuration we wish to switch to during reload.

One of our internal modules does some questionable things at an early stage of the reload process. Those things need to acquire the pending configuration.

This logic could be delayed easily by an apphook (`AH_CONFIG_CHANGED`), but unfortunately, it is currently not possible as we register source mangle callbacks there, and those need to be registered before cfg_init() to avoid data races with threaded sources.

### DIFF
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -805,6 +805,12 @@ void register_source_mangle_callback(GlobalConfig *src, mangle_callback cb)
   src->source_mangle_callback_list = g_list_append(src->source_mangle_callback_list, cb);
 }
 
+gboolean
+is_source_mangle_callback_registered(GlobalConfig *src, mangle_callback cb)
+{
+  return !!g_list_find(src->source_mangle_callback_list, cb);
+}
+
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
Without checking the details fyi: In one of my parallel pull requests I introduced PRE_CONFIG_INIT hook and also a PRE_PRE_CONFIG_INIT those could be useful for this.

bazsi:
It's in #3969

Kokan (approved):

MrAnno:
Thank you for the note. PRE_CONFIG_INIT is a good candidate but it currently has the same problem as all the other hooks.
I'll look into fixing this in a cleaner way later.

alltilla (approved):

MrAnno:
Merging it with 2 approvals as we need it to fix a release blocker issue in an internal module.

PR #3990, opened 2022-04-24
Support Kubernetes container logs

bazsi:
This patch reworks Peter Viskup's original cri-parser() to complete support for container logging in Kubernetes clusters.
    
This defines a new source, kubernetes() that follows the files under /var/log/containers or /var/log/pods and extracts both k8s metadata and decodes the multi-line format as generated by kubelet and the underlying container runtime.
    
This only supports the CRI log format (e.g.  the tabulated text based format and not JSON), support for the JSON version should be pretty straight-forward.
    
The k8s messages are decoded as:
    
      MESSAGE -- decoded message, potentially multi-line
      .k8s.*  -- POD/container meta information
      HOST    -- the name of the node (e.g. $LOGHOST)
      PROGRAM -- namespace/pod/container
      PID     -- pod_uuid
    
This way both syslog and name-value aware destinations can easily consume this format.

This is a sample config file:

```
@version: 3.36

log {
    source { kubernetes(); };
    destination { file("foo.log"); };
};
```

There are lower level components in the patch, such as "kubernetes-file-parser()" which does the heavy lifting.


kira-syslogng:
Build SUCCESS

peterviskup:
Looks very promising. I would not be able to write it this way.

Possible limitation is that the 
> PROGRAM -- namespace/pod/container

might quickly reach the limit of 48 characters for APP-NAME in rfc5424.
Would be good to mention it in the documentation or provide the way to let the parser know which 'segments' to use in that field. For simple deployments the 'container' might be sufficient, for others maybe namespace/container or just container and make the namespace part of HOST. Most of the deployments are one-pod:one-container, but more advanced might be one-pod:more-containers.

The other thing I thought about within the original implementation was the
> HOST    -- the name of the node (e.g. $LOGHOST)

could be the K8S cluster name, but that needs to be get from shell or defined as an configuration macro.

bazsi:
On Mon, Apr 25, 2022, 22:10 peterviskup ***@***.***> wrote:

> Looks very promising. I would not be able to write it this way.
>
> Possible limitation is that the
>
> PROGRAM -- namespace/pod/container
>
> might quickly reach the limit of 48 characters for APP-NAME in rfc5424.
>
Good point.

Would be good to mention it in the documentation or provide the way to let
> the parser know which 'segments' to use in that field. For simple
> deployments the 'container' might be sufficient, for others maybe
> namespace/container or just container and make the namespace part of HOST.
> Most of the deployments are one-pod:one-container, but more advanced might
> be one-pod:more-containers.
>
I am thinking of putting all this into structured data somehow. I am a bit
reluctant to duplicate information in .k8s.* and the same under .sdata.k8s
though.

The first set makes sense if we are sending data to something like elastic
or splunk which consume JSON. The second set only makes sense if we are
sending syslog.

Let me think of that a bit. I consider the first to be the canonical
representation, the 2nd should just be a re-mapping of the first.

The other thing I thought about within the original implementation was the
>
> HOST -- the name of the node (e.g. $LOGHOST)
>
> could be the K8S cluster name, but that needs to be get from shell or
> defined as an configuration macro.
>

Do we ever need the node name at all? Syslog-ng can execute external
commands using the confgen plugin, however a selinux and friends usually
complain about those.

bazsi:
I have now removed the WIP pieces of this PR and tried to concentrate on the Kubernetes SCL, pushing out grouping-by() improvements into a separate PR (coming up in a few minutes).

@peterviskup I've modified the syslog-style name-value pairs a bit to avoid hitting the 48 character limit that you pointed out:

  $HOST -- cluster-name/namespace/pod-name
  $PROGRAM -- container-name
  $PID -- pod_uid/container_id + log stream name (stdout or stderr)

The cluster-name piece of the above is a parameter to the kubernetes() source, I haven't found a way to query this information at runtime from kubelet or anywhere. Do you have a idea for a solution here? Thanks.

kira-syslogng:
Build SUCCESS

bazsi:
@peterviskup I have included a copyright header in the kubernetes.conf file and attributed the copyright to you as well. Please, if you agree, acknowledge it explicitly that you are ok with this attribution. If you do, I'd add an explicit Signed-off-by line with your name/email in my followup commit.

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
Closing this PR.

PR #3995, opened 2022-04-29
disable caps API

gaborznagy:
Some internal tools reuse code parts from syslog-ng modules, which uses capabilities.
For some of these tools this is not needed, so let's allow them to disable it, similarly to `--no-caps` option of syslog-ng.

### DIFF
--- a/lib/gprocess.c
+++ b/lib/gprocess.c
@@ -1572,6 +1572,13 @@ g_process_process_mode_arg(const gchar *option_name G_GNUC_UNUSED, const gchar *
   return TRUE;
 }
 
+void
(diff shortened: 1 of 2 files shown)


bazsi (approved):
The return type of the new function should be on a separate line but otherwise looks ok.

gaborznagy:
> The return type of the new function should be on a separate line but otherwise looks ok.

Ohh, I was in a hurry, sorry. Fixed it. :)

kira-syslogng:
Build SUCCESS

OverOrion (approved):
LGTM!

MrAnno (approved):

PR #3997, opened 2022-04-29
logreader: fix use-after-free bug in non-threaded mode

bazsi:
This is an alternative fix for #3985. 

fix `threaded(no)` related crash: if threaded mode is disabled for asynchronous sources and destinations (all syslog-like drivers such as tcp/udp/syslog/network qualify), a use-after-free condition can happen due to a reference counting bug in the non-threaded code path.  

The `threaded(yes)` setting has been the default since 3.6.1 so if you are using a more recent version, you are most probably unaffected.  If you are using `threaded(no)` a use-after-free condition happens as the connection closes.  The problem is more likely to surface on 32 bit platforms due to pointer sizes and struct layouts where this causes a NULL pointer ereference.

### DIFF
--- a/lib/logreader.c
+++ b/lib/logreader.c
@@ -572,8 +572,10 @@ log_reader_io_handle_in(gpointer s)
        */
       if (!main_loop_worker_job_quit())
         {
+          log_pipe_ref(&self->super.super);
           log_reader_work_perform(s, G_IO_IN);
           log_reader_work_finished(s);
+          log_pipe_unref(&self->super.super);
         }
     }
 }
--- a/lib/logwriter.c
+++ b/lib/logwriter.c
@@ -279,8 +279,10 @@ log_writer_io_handler(gpointer s, GIOCondition cond)
 
       if (!main_loop_worker_job_quit())
         {
+          log_pipe_ref(&self->super);
           log_writer_work_perform(s, cond);
           log_writer_work_finished(s);
+          log_pipe_unref(&self->super);
         }
     }
 }
--- a/modules/systemd-journal/journal-reader.c
+++ b/modules/systemd-journal/journal-reader.c
@@ -555,8 +555,10 @@ _io_process_input(gpointer s)
     {
       if (!main_loop_worker_job_quit())
         {
+          log_pipe_ref(&self->super.super);
           _work_perform(s, G_IO_IN);
           _work_finished(s);
+          log_pipe_unref(&self->super.super);
         }
     }
 }
--- a/news/bugfix-3997.md
+++ b/news/bugfix-3997.md
@@ -1,0 +1,9 @@
+fix `threaded(no)` related crash: if threaded mode is disabled for
+asynchronous sources and destinations (all syslog-like drivers such as
+tcp/udp/syslog/network qualify), a use-after-free condition can happen due
+to a reference counting bug in the non-threaded code path.  The
+`threaded(yes)` setting has been the default since 3.6.1 so if you are using
+a more recent version, you are most probably unaffected.  If you are using
+`threaded(no)` a use-after-free condition happens as the connection closes.
+The problem is more likely to surface on 32 bit platforms due to pointer
+sizes and struct layouts where this causes a NULL pointer dereference.


kira-syslogng:
Build SUCCESS

bazsi:
This is pretty critical to go in as fast as possible. Since it's been in the code for years, I wouldn't do an emergency release, but it should come the sooner the better.

Kudos to @sparlane for the original, detailed report and the help in fixing the issue.

kira-syslogng:
Build SUCCESS

MrAnno (approved):
Oh, nice catch.

This seems to be the non-threaded version of `self->io_job.engage` and `self->io_job.release` from #2434.

MrAnno:
Note: In the original PR, the backtrace seems to show a direct call of `log_reader_work_finished()` from the ivykis poller, which might suggest that the original issue happened in threaded mode. `log_reader_io_handle_in()` can be optimized away I guess, but we should make sure. 

```
#1  log_reader_work_finished (s=0x100bde80) at lib/logreader.c:402
        self = 0x100bde80
#2  0x0fc65b98 in iv_fd_poll_and_run (st=st@entry=0x100270f0, abs=<optimized out>) at iv_fd.c:219
```

@sparlane Could you share your configuration?

sparlane:
Hi @MrAnno,

Build configuration (from a different target than crashes, but it will be the same except for the target/host options):
`./configure --target=x86_64-unknown-linux-gnu --host=x86_64-unknown-linux-gnu --build=x86_64-linux-gnu --prefix=/usr --config-cache --enable-shared --disable-static --sysconfdir=/etc --localstatedir=/var --prefix=/usr --enable-dependency-tracking--enable-dynamic-linking --with-pidfile-dir=/var/run --disable-amqp --enable-ipv6 --with-ivykis=system --disable-json --disable-mongodb --disable-python --disable-stomp --disable-systemd`

syslog-ng.conf:
```
@version:3.21
#
# Generic Syslog-ng configuration
#

#
# include the configs in /etc/syslog-conf.d/
#
@include "syslog-conf.d"
```
/etc/syslog-conf.d/Main:
```
## First configuration file loaded
options {
	threaded(no);
	perm(0644);
	keep_hostname(yes);
	use_uniqid(yes);

	# Flow control
	#
	# Write messages as soon as they arrive and limit buffering for when messages cannot be written.
	flush_lines(0);
	log_fifo_size(100);

	# Statistics
	#
	# Collect extensive statistics, but don't log messages about them.
	stats_freq(0);
	stats_level(3);
	stats_lifetime(2592000);

	# DNS
	use_fqdn(yes);
	use_dns(persist_only);
	dns_cache_hosts(/etc/hosts);
};

#
# global sources
#
source s_src {
	unix-dgram("/run/systemd/journal/syslog");
	file("/proc/kmsg");
	pipe("/tmp/vcs_log" flags(expect-hostname));
	pipe("/tmp/atmf_log" flags(expect-hostname));
	pipe("/tmp/offload_log" flags(expect-hostname));
	internal();
};

#
# global templates
#
template t_loghost {
    template("<${PRI}>${FULLDATE} ${HOST} ${MSGHDR}${MSG}\n");
    template_escape(no);
};

template t_loghost_iso {
    template("<${PRI}>${ISODATE} ${HOST} ${MSGHDR}${MSG}\n");
    template_escape(no);
};

template t_remote_logfile {
	template("<${PRI}> ${ISODATE} ${HOST} ${MSGHDR}${MSG}\n");
};

template t_logfile {
	template("${FULLDATE} ${FACILITY}.${LEVEL} ${HOST} ${MSGHDR}${MSG}\n");
};

#
# global filters
#
filter f_all_levels {
	level(emerg,alert,crit,err,warning,notice,info,debug);
};
filter f_all {
	filter(f_permanent) or filter(f_console) or filter(f_buffered) or facility(local6);
};

#
# trace debug
#
template t_trace "${FULLDATE} ${HOST} ${MSGHDR}${MSG}\n";
filter f_trace { facility(local7); };
destination d_trace { file("/var/log/trace" template(t_trace) flags(no-multi-line)); };
log { source(s_src); filter(f_trace); destination(d_trace); };
```

So yes, we have threaded(no) in our config, I missed this when looking earlier because this part of our config is stored away from other parts of our syslog config in our git repo

MrAnno:
Thank you for the confirmation.

bazsi:
In the original patch in #2434 we had this ref/unref calls in LogWriter and I was arguing it was not needed.

I was wrong :(

Also, the same problem seems to exists in JournalReader although I think that can't be closed this way. I'll check nevertheless.

bazsi:
added the ref/unref calls in JournalReader. I didn't find any other usage of MainLoopIOJob

kira-syslogng:
Build SUCCESS

OverOrion (approved):

PR #4000, opened 2022-05-02
Add new TLS options sigalgs() and client-sigalgs() 

### LINKED ISSUE #3993: Support SSL_CONF_cmd() to configure advanced client tls options
## Description of the problem
For TLS configuration, unable to specify the signatureAlgorithms/clientSignatureAlgorithms.

## Proposed solution
Instead of using SSL_CTX_set_cipher_list()/SSL_CTX_set_ciphersuites the TLS configuration should use SSL_CONF_cmd

## Alternatives
NA

## Additional context
For certain compliance standards, say NIAP we need to limit Client-Hello's cipher-suites, signatureAlgs, clientSigntureAlgs, elliptic-curves and perform OCSP validation. Such features are not configurable via commands being currently in use (SSL_CTX_set_cipher_list()/SSL_CTX_set_ciphersuites)

See discussion in https://github.com/syslog-ng/syslog-ng/discussions/3983

MrAnno:
This PR adds 2 new TLS options, which make it possible to restrict which signature/hash pairs can be used in digital signatures.

This is a TLS extension specified by RFC5246:
https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.1.4.1

sigalgs:
> This sets the supported signature algorithms for TLSv1.2 and TLSv1.3.
> For clients this value is used directly for the supported signature
> algorithms extension. For servers it is used to determine which
> signature algorithms to support.

client-sigalgs:
> This sets the supported signature algorithms associated with client
> authentication for TLSv1.2 and TLSv1.3. For servers the value is used
> in the signature_algorithms field of a CertificateRequest message.
> For clients it is used to determine which signature algorithm to use
> with the client certificate.

Example config:

```
source {
	network(port(4444) transport(tls)
		tls(
			pkcs12-file("/home/anno/Work/syslog-ng/tls/localhost.p12")
			peer-verify(yes)
			client-sigalgs("RSA-PSS+SHA256:ed25519")
		)
	);
};

destination {
	network("localhost" port(4444) transport(tls)
		tls(
			pkcs12-file("/home/anno/Work/syslog-ng/tls/localhost.p12")
			peer-verify(yes)
			sigalgs("RSA-PSS+SHA256:ed25519")
		)
	);
};
```

Partially resolves #3993

### DIFF
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -60,6 +60,8 @@ struct _TLSContext
   gchar *ca_file;
   gchar *cipher_suite;
   gchar *tls13_cipher_suite;
+  gchar *sigalgs;
+  gchar *client_sigalgs;
   gchar *ecdh_curve_list;
   gchar *sni;
   SSL_CTX *ssl_ctx;
@@ -663,6 +665,18 @@ tls_context_setup_cipher_suite(TLSContext *self)
   return TRUE;
 }
 
+static gboolean
+tls_context_setup_sigalgs(TLSContext *self)
+{
+  if (self->sigalgs && !SSL_CTX_set1_sigalgs_list(self->ssl_ctx, self->sigalgs))
+    return FALSE;
+
+  if (self->client_sigalgs && !SSL_CTX_set1_client_sigalgs_list(self->ssl_ctx, self->client_sigalgs))
+    return FALSE;
+
+  return TRUE;
+}
+
 static PKCS12 *
 _load_pkcs12_file(TLSContext *self, const gchar *pkcs12_file)
 {
@@ -826,6 +840,9 @@ tls_context_setup_context(TLSContext *self)
   if (!tls_context_setup_cipher_suite(self))
     goto error;
 
+  if (!tls_context_setup_sigalgs(self))
+    goto error;
+
   return TLS_CONTEXT_SETUP_OK;
 
 error:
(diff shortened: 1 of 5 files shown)


bazsi (dismissed):

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

lib/tlscontext.c
```diff
@@ -60,6 +60,8 @@ struct _TLSContext
   gchar *ca_file;
   gchar *cipher_suite;
   gchar *tls13_cipher_suite;
+  gchar *sigalgs;
```
gaborznagy, line 63: These should be freed ( in `_tls_context_free()`).

MrAnno: Fixed, thanks.

kira-syslogng:
Build SUCCESS

gaborznagy (approved):

OverOrion (approved):
LGTM, tested:
```
Signature Hash Algorithms (2 algorithms)
    Signature Algorithm: rsa_pss_rsae_sha256 (0x0804)
    Signature Algorithm: ed25519 (0x0807)
```

mohitvaid:
@MrAnno @gaborznagy is this change applicable for MQTT TLS as well?

MrAnno:
Unfortunately, the MQTT source/destination uses a library (`paho.mqtt.c`), which does not support restricting sigalgs.
The best way to add support to the MQTT driver for this is asking them to either make it possible to freely configure an OpenSSL context or just add support for sigalgs restriction as we did.

https://github.com/eclipse/paho.mqtt.c

mohitvaid:
How about supporting the ecdh-curve-list? I will follow up with paho separately.

MrAnno:
The same is true for `ecdh-curve-list()`: it is supported in standard network drivers, but not in MQTT/HTTP.

Supporting it in the HTTP destination is possible, let us know if you need it by opening a feature request .

PR #4009, opened 2022-05-19
fix: Warning elimination for Wbad-function-cast flag

shikharvashistha:
Greetings of the day,

This pull adds the elimination of warning mentioned in https://github.com/syslog-ng/syslog-ng/issues/3255, kindly review the pull and let me know in case of changes if any.

Best Regards
@shikharvashistha


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

shikharvashistha:
As I've installed bison in the same directory, therefore `make style-format` was unable to open some bison header files for formatting. 

Fixed the issue 👍

lib/logmsg/tags.c
```diff
@@ -64,12 +64,15 @@ log_tags_get_by_name(const gchar *name)
...
+  gpointer key;
 
   g_assert(log_tags_hash != NULL);
 
   g_mutex_lock(&log_tags_lock);
 
-  id = GPOINTER_TO_UINT(g_hash_table_lookup(log_tags_hash, name)) - 1;
+  key = g_hash_table_lookup(log_tags_hash, name);
+  id = GPOINTER_TO_UINT(key) - 1;
+
```
bazsi, line 75: Using declarations right at the place of use would make this clearer.

E.g.

```
gpointer key = ...;
guint id = ...;
```

lib/logthrdest/logthrdestdrv.c
```diff
@@ -1108,8 +1109,9 @@ log_threaded_dest_driver_init_method(LogPipe *s)
   if (self->time_reopen == -1)
     self->time_reopen = cfg->time_reopen;
 
-  self->shared_seq_num = GPOINTER_TO_INT(cfg_persist_config_fetch(cfg,
-                                         _format_seqnum_persist_name(self)));
+  value = cfg_persist_config_fetch(cfg, _format_seqnum_persist_name(self));
```
bazsi, line 1112: Again, make ```value``` inline declared. Maybe a better name would make it even better. (E.g. persisted_value)

lib/stats/aggregator/stats-change-per-second.c
```diff
@@ -246,12 +246,13 @@ static void
...
 {
   gsize diff = _get_count(self) - _get_last_count(logic);
+  gdouble elapsed_time_since_last;
   _set_last_count(logic, _get_count(self));
 
   if (!_is_less_then_duration(self, logic, now))
     {
-      gsize elapsed_time_since_last = (gsize)_calc_sec_between_time(&self->last_add_time, now);
-      diff -= _get_average(logic) * elapsed_time_since_last;
+      elapsed_time_since_last = _calc_sec_between_time(&self->last_add_time, now);
```
bazsi, line 254: As far as I understand all invocations of _calc_sec_between_time() cast the return value to gsize.

It's a local, inline function. Why don't we change it to return gsize instead?

Honestly the choice to represent time values as gsize is also strange.

Even an int would be better. A time_t would be the best probably..hmm why does

(A follow-up question why difftime() returns double if its arguments are time_ts is also beyond me
 But it's not the fault of this or)

lib/template/macros.c
```diff
@@ -719,10 +719,13 @@ log_macro_lookup(const gchar *macro, gint len)
 {
   gchar buf[256];
   gint macro_id;
+  gpointer p;
 
   g_assert(macro_hash);
   g_strlcpy(buf, macro, MIN(sizeof(buf), len+1));
-  macro_id = GPOINTER_TO_INT(g_hash_table_lookup(macro_hash, buf));
+  p = g_hash_table_lookup(macro_hash, buf);
```
bazsi, line 726: Inline, better name.

modules/afsocket/afsocket-source.c
```diff
@@ -843,13 +843,14 @@ _sd_open_stream(AFSocketSourceDriver *self)
...
   gint sock = -1;
+  gpointer fd;
   if (self->connections_kept_alive_across_reloads)
     {
       /* NOTE: this assumes that fd 0 will never be used for listening fds,
        * main.c opens fd 0 so this assumption can hold */
-      sock = GPOINTER_TO_UINT(
-               cfg_persist_config_fetch(cfg, afsocket_sd_format_listener_name(self))) -
-             1;
+      fd = cfg_persist_config_fetch(cfg, afsocket_sd_format_listener_name(self));
```
bazsi, line 851: Using fd as name and declaring it a pointer is misleading

modules/diskq/tests/test_diskq_truncate.c
```diff
@@ -96,8 +96,12 @@ _calculate_full_disk_message_num(LogQueueDisk *queue_disk)
    * Assumes that only empty messages are stored.
    */
   QDisk *qdisk = queue_disk->qdisk;
+  gdouble num_messages;
+
   gsize msg_size = _calculate_serialized_empty_message_size(queue_disk);
-  return (gint64) ceil((qdisk->options->disk_buf_size - QDISK_RESERVED_SPACE) / (double) msg_size);
+  num_messages = ceil((qdisk->options->disk_buf_size - QDISK_RESERVED_SPACE) / (double) msg_size);
+
+  return (gint64)num_messages;
```
bazsi, line 104: llrint() ?

bazsi (changes requested):
Well this warning is not necessarily the most successful GCC warning I have ever seen. In any case if the work is done we should merge this.

I had some comments against changes in your PR.

lib/logmsg/tags.c
shikharvashistha: Fixed 👍

lib/logthrdest/logthrdestdrv.c
shikharvashistha: fixed 👍

lib/template/macros.c
shikharvashistha: fixed 👍

modules/afsocket/afsocket-source.c
shikharvashistha: renamed 👍

modules/diskq/tests/test_diskq_truncate.c
shikharvashistha: Updated 👍

lib/stats/aggregator/stats-change-per-second.c
shikharvashistha: After going through few articles/blogs(attaching for reference), I found that the time_t is measured in seconds, and difftime() is equivalent to arithmetic subtraction, but C standards allows fractional units for time_t therefore, it's preffered to use double as the return type. Though I've made the changes as desired.

- [1](https://en.cppreference.com/w/c/chrono/difftime)
- [2](https://en.cppreference.com/w/c/chrono/time_t)
- [3](https://stackoverflow.com/questions/34457931/why-does-difftime-return-a-double)

lib/stats/aggregator/stats-change-per-second.c
```diff
@@ -239,7 +240,7 @@ _is_less_then_duration(StatsAggregatorCPS *self, CPSLogic *logic, time_t *now)
   if (logic->duration == -1)
     return TRUE;
 
-  return _calc_sec_between_time(&self->init_time, now) <= logic->duration;
+  return (gsize)_calc_sec_between_time(&self->init_time, now) <= logic->duration;
```
bazsi, line 243: You still seem to have a couple of gsize casts here and there.

shikharvashistha: Updated 👍

bazsi (approved):
This looks good to me, approved.

modules/diskq/qdisk.h
```diff
@@ -35,7 +35,12 @@
  * uses this breaks, as it passes the result of a g_queue_pop_head call,
  * which has side effects.
  */
-#define POINTER_TO_LOG_PATH_OPTIONS(ptr, lpo) (lpo)->ack_needed = (GPOINTER_TO_INT(ptr) & ~0x80000000)
+#define POINTER_TO_LOG_PATH_OPTIONS(ptr, lpo) \
+({ \
+  gpointer p = ptr; \
+  LogPathOptions *opts = lpo; \
```
Kokan, line 41: Why do we need `opts` variable here ?

shikharvashistha: Updated accordingly

bazsi (approved):
still approved.

lib/stats/aggregator/stats-change-per-second.c
bazsi: Sorry, I've assumed that difftime() was our own function. I understand that time_t can also be an alias to double, but syslog-ng's codebase assumes that it is an integer value. If we were to port to one of those platforms where time_t is a double, we would probably have trouble compiling.

But for now, I don't think we want to change this assumption, so this is fine how it is now.

MrAnno (approved):

PR #4012, opened 2022-05-19
OpenSSL V3 (3.0) support

OverOrion:
This PR adds support for warning-free compilation with OpenSSL 3.0.
Kudos to @MrAnno

A few notes:
- MD4 hash template function will be deprecated  (see 67ae1c3 for details)
- Ubuntu 22.04 ships with OpenSSL 3.0 by default if you want to check the warning-free compilation without compiling OpenSSL for yourself
- Please do not merge it yet, I still want to run some (internal) tests

I was hesitant to move the modifications regarding `cmac` function to the compat layer, my reasoning:
-  It is in the `secure-log` module meaning that it is acceptable to directly interface with the OpenSSL API

If it should be in the compat layer, then please let me know and I will move it accordingly.

### DIFF
--- a/lib/compat/openssl_support.c
+++ b/lib/compat/openssl_support.c
@@ -188,6 +188,9 @@ openssl_ctx_setup_ecdh(SSL_CTX *ctx)
 gboolean
 openssl_ctx_setup_dh(SSL_CTX *ctx)
 {
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+  return SSL_CTX_set_dh_auto(ctx, 1);
+#else
   DH *dh = DH_new();
   if (!dh)
     return 0;
@@ -214,8 +217,10 @@ openssl_ctx_setup_dh(SSL_CTX *ctx)
 
   DH_free(dh);
   return ctx_dh_success;
+#endif
 }
 
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
 static long
 _is_dh_valid(DH *dh)
 {
@@ -234,6 +239,7 @@ _is_dh_valid(DH *dh)
 
   return !error_flag_is_set;
 }
+#endif
 
 gboolean
 openssl_ctx_load_dh_from_file(SSL_CTX *ctx, const gchar *dhparam_file)
@@ -242,6 +248,21 @@ openssl_ctx_load_dh_from_file(SSL_CTX *ctx, const gchar *dhparam_file)
   if (!bio)
     return 0;
 
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+  EVP_PKEY *dh_params = PEM_read_bio_Parameters(bio, NULL);
+  BIO_free(bio);
+
+  if (!dh_params)
+    return 0;
+
+  int ctx_dh_success = SSL_CTX_set0_tmp_dh_pkey(ctx, dh_params);
+
+  if (!ctx_dh_success)
+    EVP_PKEY_free(dh_params);
+
+  return ctx_dh_success;
+
+#else
   DH *dh = PEM_read_bio_DHparams(bio, NULL, NULL, NULL);
   BIO_free(bio);
 
@@ -255,6 +276,7 @@ openssl_ctx_load_dh_from_file(SSL_CTX *ctx, const gchar *dhparam_file)
 
   DH_free(dh);
   return ctx_dh_success;
+#endif
 }
 
 #if !SYSLOG_NG_HAVE_DECL_DH_SET0_PQG
--- a/lib/compat/openssl_support.c
+++ b/lib/compat/openssl_support.c
@@ -162,7 +162,8 @@ openssl_init(void)
 #endif
 }
 
-void openssl_ctx_setup_ecdh(SSL_CTX *ctx)
+void
+openssl_ctx_setup_ecdh(SSL_CTX *ctx)
 {
 #if OPENSSL_VERSION_NUMBER >= 0x10100000L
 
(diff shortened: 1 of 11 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

modules/secure-logging/slog.c
```diff
@@ -420,15 +424,34 @@ gchar *convertToBase64(unsigned char *input, gsize len)
  */
 void cmac(unsigned char *key, const void *input, gsize length, unsigned char *out, gsize *outlen)
 {
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
```
alltilla, line 427: Reflecting on
> It is in the secure-log module meaning that it is acceptable to directly interface with the OpenSSL API

Generally it is allowed to use OpenSSL API directly from any module. Just like before `tlscontext: move setup methods of DH-params to openssl-compat`, the `lib/tlscontext.c` file used them. Most of the functions in `lib/compat/openssl_support.c` are redefinitions of original OpenSSL functions for the sake of version compatibility, but sometimes we put "helper" functions there, just like you did in the mentioned commit.

I think what you did in that commit is really similar to moving out this function.

You might ask: what do we gain by moving these helper functions to the compat layer? I think it is better to see the OpenSSL version preprocessor `#if`s in one separate file, than them being everywhere in the codebase. So I would prefer moving these functions there.

modules/secure-logging/slog.c
```diff
@@ -420,15 +424,34 @@ gchar *convertToBase64(unsigned char *input, gsize len)
...
 
   CMAC_Init(ctx, key, KEY_LENGTH, EVP_aes_256_cbc(), NULL);
   CMAC_Update(ctx, input, length);
 
-  size_t outsize;
-  CMAC_Final(ctx, out, &outsize);
-  *outlen = outsize;
+  size_t out_len;
+  CMAC_Final(ctx, out, &out_len);
+  *outlen = out_len;
```
alltilla, line 452: [nitpicking, no need to change anything]

Reformatting lines of code, which we did not modify could go to separate commits. Same applies for removing dead code (`sLogDecrypt()` in the header file).

kira-syslogng:
Build SUCCESS

MrAnno:
OpenSSL 3.0 is already available in debian-sid. Hopefully, it will reach debian-testing soon.

lib/compat/openssl_support.c
```diff
@@ -184,6 +185,78 @@ void openssl_ctx_setup_ecdh(SSL_CTX *ctx)
 #endif
 }
 
+long
```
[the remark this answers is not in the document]
MrAnno: Yes, gbool would be better for methods that are not strict OpenSSL compat functions.

modules/secure-logging/slog.c
```diff
@@ -420,15 +424,34 @@ gchar *convertToBase64(unsigned char *input, gsize len)
...
+    OSSL_PARAM_utf8_string("cipher", "aes-256-cbc", 0),
+    OSSL_PARAM_END,
+  };
+
+  EVP_MAC_CTX *ctx = EVP_MAC_CTX_new(mac);
+
+  EVP_MAC_init(ctx, key, KEY_LENGTH, params);
+  EVP_MAC_update(ctx, input, length);
+  size_t out_len;
+  EVP_MAC_final(ctx, out, &out_len, 99999);
```
MrAnno: This was probably an experiment, please fix the 4. parameter.

modules/secure-logging/slog.c
MrAnno: cmac() is not general enough in its current form (and is not well-audited) to be moved under compat/openssl.

We have to be really careful what we put there. IMHO the openssl_support unit should contain trivial copy-pated compatibility functions, other non-trivial helpers, but we should not place custom cryptographic primitives there unless they are general enough to be used anywhere and they are audited properly.

modules/secure-logging/slog.c
MrAnno: Because crypto-related code is modified here, and we almost missed a bug in the few lines above, I think it would be best to split this commit to make it as atomic and clean as possible.

lib/compat/openssl_support.c
```diff
@@ -333,3 +338,38 @@ openssl_ctx_setup_session_tickets(SSL_CTX *ctx)
   SSL_CTX_set_num_tickets(ctx, 0);
 #endif
 }
+
+void
+openssl_cmac(unsigned char *key, size_t key_size, const void *input, gsize length, unsigned char *out, gsize *outlen)
```
MrAnno, line 343: I think we should not do this in its current form. See my comment above.

modules/secure-logging/slog.c
OverOrion: Thank you, I completely missed it when I started working again on this branch

OverOrion:
Thank you @alltilla, @MrAnno for your notes.
Based on them:
- `cmac` stays in its original place
- the dummy 4th parameter in  `EVP_MAC_final` is now the correct value*
- `SYSLOG_NG_HAVE_DECL_DIGEST_MD4` macro is defined as `1`

*Because of propagation it was necessary to change the `cmac` and its callers's API that's why there are so many commits.
The change is introduced in 23fc6ee21c851e49094d14a1c3f3da5811a3880f and the commits after it are follow-ups for each function.
Finally, to eliminate the array calculating logic, the `GET_BUF_SIZE` is introduced.

Kindly check them. Thanks in advance!

kira-syslogng:
Build SUCCESS

MrAnno:
@OverOrion I might have said something confusing in my comment about atomicity.
Could you please squash the commits that follow the new cmac parameter up?
They are hard to revert/bisect/review this way (they don't compile, which is one of our requirements).

modules/secure-logging/slog.c
```diff
@@ -493,19 +493,20 @@ void PRF(unsigned char *key, unsigned char *originalInput, guint64 inputLength,
 
   // Make sure that temporary buffer can hold at least outputLength bytes, rounded up to a multiple of AES_BLOCKSIZE
   unsigned char buf[outputLength+AES_BLOCKSIZE];
+  gsize buf_capacity = sizeof(buf) / sizeof(buf[0]);
   // Prepare plaintext
   for (int i=0; i<outputLength/AES_BLOCKSIZE; i++)
     {
       gsize outlen;
-      cmac(key, input, AES_BLOCKSIZE, &buf[i*AES_BLOCKSIZE], &outlen);
+      cmac(key, input, AES_BLOCKSIZE, &buf[i*AES_BLOCKSIZE], &outlen, buf_capacity);
```
MrAnno, line 501:
```diff
- cmac(key, input, AES_BLOCKSIZE, &buf[i*AES_BLOCKSIZE], &outlen, buf_capacity);
+ cmac(key, input, AES_BLOCKSIZE, buf + (i*AES_BLOCKSIZE), &outlen, G_N_ELEMENTS(buf) - (i*AES_BLOCKSIZE));
```

modules/afsql/afsql.c
```diff
@@ -384,15 +384,15 @@ afsql_dd_create_index(AFSqlDestDriver *self, const gchar *table, const gchar *co
...
       if ((strlen(table) + strlen(column)) > 25)
         {
 
-          guchar hash[MD5_DIGEST_LENGTH];
+          guchar hash[SHA_DIGEST_LENGTH];
           gchar hash_str[31];
           gchar *cat = g_strjoin("_", table, column, NULL);
 
-          MD5((guchar *)cat, strlen(cat), hash);
+          SHA1((guchar *)cat, strlen(cat), hash);
```
MrAnno, line 395: We've discussed this IRL and we decided to keep the old MD5 index to be fully compatible with older versions.

A non-deprecated API can be used: EVP_Digest.

kira-syslogng:
Build SUCCESS

MrAnno (dismissed):

modules/cryptofuncs/cryptofuncs.c
```diff
@@ -94,7 +94,12 @@ tf_hash_prepare(LogTemplateFunction *self, gpointer s, LogTemplate *parent, gint
       return FALSE;
     }
   state->length = length;
-  md = EVP_get_digestbyname(strcmp(argv[0], "hash") == 0 ? "sha256" : argv[0]);
+  const char *digest_name = strcmp(argv[0], "hash") == 0 ? "sha256" : argv[0];
+  if (strcmp(digest_name, "md4") == 0)
+    {
+      msg_warning_once("WARNING: MD4 hash template function is deprecated and will be removed in future versions");
```
gaborznagy, line 100: Sorry, I didn't review the whole branch, I just have this one comment:

I think we should send a deprecation warning if syslog-ng compiled with Openssl 3 too (and nost just in case with Openssl 1.1):
We could add a dummy template function (.e.g a TEMPLATE_FUNCTION_SIMPLE for md4 in case of Openssl 3.
There is also an example how to do error reporting in case of deprecation:

https://github.com/syslog-ng/syslog-ng/blob/f69ec0a98b34f43820df3157a95d8f52b6b50c63/modules/cryptofuncs/cryptofuncs.c#L97-L102

Thanks @MrAnno to confirm my idea.

MrAnno (approved):

alltilla (approved):

modules/cryptofuncs/cryptofuncs.c
gaborznagy: I've checked the updated version, looks good to me, thanks!

kira-syslogng:
Build SUCCESS

MrAnno (approved):

alltilla (approved):

MrAnno:
Please do not merge until @OverOrion verifies the latest version of this PR with OpenSSL 3.0.
(We currently don't have a CI job for it, so it is going to be done manually.)

PR #4015, opened 2022-05-22
Kubernetes log source

bazsi:
This is an alternative PR inspired by #3905 but rewritten completely. It adds support for a kubernetes() source driver that tails container logs managed by kubelet and extracts both kubernetes metadata and a log content.

```
# Components defined in this SCL:
#
#   kubernetes() source          -- source driver that tails /var/log/containers or /var/log/pods
#                                   (depending on the base-dir() parameter)
#                                   and extracts the container logs using the parsers below.
#
#   kubernetes-file-parser()     -- extract pod log message from log file
#                                   along with kubernetes metadata.
#
#   kubernetes-metadata-parser() -- extract pod related metadata from
#                                   $FILE_NAME
#
#
# Kubernetes meta information is automatically extracted from
# $FILE_NAME, such as:
#
#  - ${.k8s.container_id} or ${.k8s.pod_uuid}
#  - ${.k8s.container_name}
#  - ${.k8s.pod_name}
#  - ${.k8s.namespace_name}
#
# Messages that are written to the log file as multiple chunks are
# automatically decoded, so it becomes a single log entry.
```

### DIFF
--- a/scl/kubernetes/kubernetes.conf
+++ b/scl/kubernetes/kubernetes.conf
@@ -1,0 +1,136 @@
+#############################################################################
+# Copyright (c) 2022 Balazs Scheidler <bazsi77@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+#
+# Integration for Kubernetes container log files as described here
+#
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

scl/kubernetes/kubernetes.conf
```diff
@@ -0,0 +1,134 @@
...
+        rewrite {
+	    set("`cluster-name`/${`prefix`namespace_name}/${`prefix`pod_name}" value("HOST"));
+            set("${`prefix`container_name}" value("PROGRAM"));
+            set("$(if ('${`prefix`pod_uuid}' ne '') ${`prefix`pod_uuid} ${`prefix`container_id})/${`prefix`stream}" value("PID"));
+            clear-tag('.tmp.k8s');
+        };
+    };
+};
+
+block source kubernetes(prefix('.k8s.' base-dir("/var/log/containers") cluster-name('k8s'))) {
```
gaborznagy, line 123: prefix() option misses a closing braces:
`prefix('.k8s.'`

OverOrion:
I've tried to check it, however with `source(kubernetes());`  syslog-ng will report a syntax error:
```
Error parsing within source, syntax error, unexpected ')', expecting LL_IDENTIFIER or LL_STRING in block source kubernetes() at ...
```
~~I think there might be small problem but I can't see it now~~
Kudos to @gaborznagy for finding it :100: :D

bazsi:
Fff.. sorry for that and thanks for pointing it out.

bazsi:
I have fixed the issue pointed out earlier and tested the kubernetes() source end-to-end with an actual kubernetes installation with both /var/log/containers and /var/log/pods as sources and both worked.

kira-syslogng:
Build SUCCESS

MrAnno (approved):

PR #4017, opened 2022-05-22
Rename type cast to datetime int

bazsi:
This is a minor refactor to change a function name that was difficult to understand for me while working on typing. This patch is independent of everything so should be relatively straightforward to merge.


kira-syslogng:
Build SUCCESS

MrAnno (approved):

OverOrion (approved):
LGTM!

PR #4018, opened 2022-05-22
basicfuncs: fix comparisons in test_basicfuncs testcases

bazsi:
There were two issues with these:
  * the $APP.VALUE wasn't using braces, so it expanded to the value
    of ${APP} and ".value" concatenated
  * it used numeric comparisons, meaning it compared two
    non-numbers numerically, which is always true (0 == 0)


kira-syslogng:
Build SUCCESS

MrAnno (approved):

Issue #4020, opened 2022-05-24
$if() does not work.


Roffild:
# syslog-ng
## Version of syslog-ng
syslog-ng 3 (3.30.1)
Config version: 3.29
Installer-Version: 3.30.1

## Platform
alpine 3.15

## Config
```
@version: 3.30
log {
source {
   channel {
      source {
         internal();
      };
      rewrite {
         set('LOG990', value('PGSEVERITY'));
         set('$(if ("${PGSEVERITY}" == "ERROR" or "${PGSEVERITY}" == "WARNING") "error" "info")', value("PGTEST"));

         set-severity('$(if (
         "${PGSEVERITY}" == "PANIC"
         or "${PGSEVERITY}" == "FATAL" or
         "${PGSEVERITY}" == "ERROR" or
         "${PGSEVERITY}" == "WARNING") "error" "info")');
      };
   };
};
destination {
   file("/home/neo/logs/${PROGRAM}/${DAY}-${PROGRAM}-${SEVERITY}.log",
        template("------------------${ISODATE}\nPGS=${PGSEVERITY} == ${PGTEST}\n"),
        create-dirs(yes),
        flush-lines(1)
   );
};
};
```
## Output
```
PGS=LOG990 == error 
```
Should be "info".

bazsi:
Hi,

You will need to use the "eq" operator instead of "==" to get string based
comparison.

Actually I am working on a PR right now that will modify this to be more
intuitive, by using type information to decide how to compare two things,
but that's not yet merged.

Here is my blog post about the topic:
https://syslog-ng-future.blog/syslog-ng-4-theme-typing/

And here's the last part of my pull request, most of it has already gone in.

https://github.com/syslog-ng/syslog-ng/pull/3831

bazsi:
@Roffild did this solve your problem?

Roffild:
Replacing `==` with `eq` solved the problem.

PR #4016, opened 2022-05-22
dbparser: propagate type information in <value> tags (4.0)

bazsi:
This completes db-parser() support for typing. Unfortunately it depends on #3948 while I thought it could go in independently. So I am rebasing this against that for now.

PR #4022, opened 2022-05-25
Type aware comparison operators (4.0)

bazsi:
This branch contains the last iteration of typing support, namely type aware comparisons. At this point, it depends all the previous portions of my typing tree, some of which are not yet merged. Namely:
  * #3948 
  * #4016

Once those are merged, this PR should become just a dozen patches, until that point it seems huge.

The intent of this change is to make the usual comparison operators (<, <=, ==, >=, > and !=) type aware, instead of interpreting and forcing their arguments to be numeric. What this means in practice:
  * we are using the internal type information associated to name-value pairs, macros and literals
  * if the values to compare are strings or are JSON objects (list, json), they are compared as strings
  * otherwise (e.g. if any of the arguments is numeric) both expressions are converted to numeric form and compared as such.

The conversion to numeric is modelled after JavaScript comparisons:
  * json and list values are converted to 1 (just as JavaScript)
  * null values are converted to 0
  * boolean values are converted 0 or 1 for false and true respectively
  * datetime values are coverted to msecs since epoch (but this is up for discussion)
  * strings/int32/int64/double values are parsed using our parse_generic_number() function that should handle integers/floating point numbers and NaN values

Once converted the two arguments are compared numerically.

The rationale behind this behavior:
  * the intuitive comparison operators (e.g. >= and not ge) are often applied to strings (for instance #4020) and confusion happens if the outcome is not as expected
  * javascript comparison operators are widely understood and even though they are not necessarily the best suited for a general purpose language (see [the good and the bad of javascript](https://www.amazon.com/JavaScript-Good-Parts-Douglas-Crockford/dp/0596517742)), but our context matches quite well to how JavaScript was conceived as an extension language for Web browsers, where most values were strings (just as with syslog-ng)
  * if we apply "=="/"!=" to two strings (the most common case) the right thing would happen. we would compare them as strings.
  * if we compare a name-value pair to a literal number, the right thing would happen (as the literal number is assumed to be  a value with a numeric type)
  * if we compare a name-value pair to another name-value pair, typing would decide how we'd compare them. This would be a change from old behavior, where these operators applied numeric conversion, whereas it could easily become string based. I plan an "===" and a "!==" operator (just as JavaScript) to ensure mismatching types are recognized

Compatibility:
  * config versions below 4.0 continue to use the old behaviour, ">=" and co would apply numeric comparisons, "ge" would apply string based.
  * once config version is set to 3.255, we start displaying warnings where we detect potential changes, which are:
    * quoted numeric values in the configuration language, the quotes should be removed.
    * comparisons against non-literal expressions can change (e.g. when two name-value pairs are compared). we only warn if the user has not added an explicit type-cast to these expressions. 
  * once we flip the version 4.0, everything works with the new behaviour.


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

alltilla:
I have reviewed the PR up until the commit: `filter-cmp: implement type-aware comparison of values`.
Maybe I misunderstand something, but I think this implementation does not mimic JavaScript's behavior.

E.g. lists in JS:
```js
console.log(Number([1, 2, 3]))
console.log([1, 2, 3] == [1, 2, 3])
console.log([1, 2, 3] == [4, 5, 6])
console.log([1, 2, 3] == 1)

===========================================================================

NaN
false
false
false
```

Or JSON objects in js:
```js
console.log(Number({"a" : 1, "b": [1, 2, 3]}))
console.log({"a" : 1, "b": [1, 2, 3]} == {"a" : 1, "b": [1, 2, 3]})
console.log({"a" : 1, "b": [1, 2, 3]} == {"x" : 10, "y": [10, 20, 30]})
console.log({"a" : 1, "b": [1, 2, 3]} == 1)

===========================================================================

NaN
false
false
false
```

`NaN` is not equal with `NaN` if compared with `==`:
```js
console.log(NaN == NaN)

===========================================================================

false
```

alltilla:
Maybe we could split off the generic number parts of this PR. I would definitely approve that change, and it would help out in a lot of other generic number dependent PRs.

bazsi:
> I have reviewed the PR up until the commit: `filter-cmp: implement type-aware comparison of values`. Maybe I misunderstand something, but I think this implementation does not mimic JavaScript's behavior.
> 
> E.g. lists in JS:
> 
> ```js
> console.log(Number([1, 2, 3]))
> console.log([1, 2, 3] == [1, 2, 3])
> console.log([1, 2, 3] == [4, 5, 6])
> console.log([1, 2, 3] == 1)
> 
> ===========================================================================
> 
> NaN
> false
> false
> false
> ```

I think this matches syslog-ng behavior.

(syslog-ng) p "$(if (list('[1,2,3]') == list('[1,2,3]')) true false)"
true
(syslog-ng) p "$(if (list('[1,2,3]') == list('[4,5,6]')) true false)"
false
(syslog-ng) p "$(if (list('[1,2,3]') == 1) true false)"
true

E.g. as long as both sides of the expression are lists, lists are compared as lists (internally, as strings though)

we currently don't have a Number() cast at the moment, it's done internally only, but when comparing lists to numeric values, it's considered '1'.

bazsi:
> 
> I think this matches syslog-ng behavior.

I have just noticed that lists in JS are never equal. Sorry, let me check this again.

bazsi:
hmm... I concentrated so much on JavaScript behavior when the types are NOT the same that I missed that if they are the same, the behaviour is not useful at all.

I think syslog-ng behaves as it should, however it does not match JavaScript behaviour in these cases, which makes documentation more difficult.

Can we discuss this in today's team call?

alltilla:
> hmm... I concentrated so much on JavaScript behavior when the types are NOT the same that I missed that if they are the same, the behaviour is not useful at all.
> 
> I think syslog-ng behaves as it should, however it does not match JavaScript behaviour in these cases, which makes documentation more difficult.
> 
> Can we discuss this in today's team call?

Sure, let's discuss it today :)

kira-syslogng:
Build FAILURE

lib/cfg-grammar.y
```diff
@@ -814,10 +814,23 @@ template_content_inner
           CHECK_ERROR_GERROR(log_template_set_type_hint(last_template, $1, &error), @1, error, "Error setting the template type-hint \"%s\"", $1);
           free($1);
         }
+        | LL_NUMBER
```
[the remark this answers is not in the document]
bazsi: I think we could even support null as a sole token, however before adding these, I'd want to implement support for further literals as we extend the syntax for filter expressions.

For the record, this is what I came up with:

```
diff --git a/lib/cfg-grammar.y b/lib/cfg-grammar.y
index 8a80adb6c..3b1b03272 100644
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -814,6 +814,14 @@ template_content_inner
           CHECK_ERROR_GERROR(log_template_set_type_hint(last_template, $1, &error), @1, error, "Error setting the template type-hint \"%s\"", $1);
           free($1);
         }
+        | LL_IDENTIFIER
+        {
+          GError *error = NULL;
+
+          CHECK_ERROR_GERROR(log_template_compile(last_template, "", &error), @0, error, "Error compiling empty template");
+          CHECK_ERROR_GERROR(log_template_set_type_hint(last_template, $1, &error), @1, error, "Error setting the template type-hint \"%s\"", $1);
+          free($1);
+        }
         | LL_NUMBER
         {
           gchar decimal[32];
```

This could work also by adding an opening/closing bracket, but I am afraid using LL_IDENTIFIER here will generate conflicts down the line. I'd probably need to add "null" as a keyword, which might cause issues with existing configurations where I've seen "null" being used unquoted as a destination name.

All in all, I like this suggestion, but let get this merged and add these down the line.

OverOrion (commented):
This looks promising, I had a few notes, kindly check them.

The last commit is missing the `Signed-off-by:`  line , which is why the CI fails.

lib/CMakeLists.txt
```diff
@@ -318,6 +320,7 @@ target_include_directories(syslog-ng
 target_link_libraries(
     syslog-ng
   PUBLIC
+    m
```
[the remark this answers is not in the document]
gaborznagy: IMHO this is to link libmath (libm). In autotools Makefiles.am, you can the `-lm` flag.
Generic Numbers use `math.h`.

lib/tests/Makefile.am
```diff
@@ -3,6 +3,7 @@ lib_tests_TESTS		= \
 	lib/tests/test_lexer_block	\
 	lib/tests/test_cfg_tree		\
 	lib/tests/test_parse_number	\
+	lib/tests/test_generic_number	\
```
gaborznagy, line 6: test_generic_number is missing from lib/tests/CMakeLists.txt

lib/filter/filter-cmp.c
```diff
@@ -25,74 +25,206 @@
...
+  return result;
+}
+
+
+/* NOTE: this function mimics JavaScript when converting a value to a
+ * number.  As described here:
+ *
+ *   Primitive types: https://javascript.info/type-conversions
+ *   Objects: https://javascript.info/object-toprimitive
+ */
```
alltilla, line 99: If I remember correctly we deviate in some cases from JS. Shouldn't we remove this comment?

alltilla (dismissed):
Thanks for the changes, awesome feature!

I made a chart for the `==` comparison, and checked the extremal values manually, they work as expected. Unit tests cover these cases as well.

Approve! :)
Screenshot from 2022-08-20 11-59-24

lib/filter/filter-cmp.c
MrAnno: I think it's worth keeping it and completing the note that the implementation follows JS but with slight modifications.

lib/generic-number.c
```diff
@@ -0,0 +1,131 @@
...
+    return (gdouble) number->value.raw_int64;
+  g_assert_not_reached();
+}
+
+void
+gn_set_double(GenericNumber *number, gdouble value, gint precision)
+{
+  number->type = GN_DOUBLE;
+  number->value.raw_double = value;
+  number->precision = precision > 0 ? precision : 20;
```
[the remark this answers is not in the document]
bazsi: well, it was 20 in numeric-funcs.c, that's why. I think the reasoning behind is that 20 is roughly the maximum number of significant digits a double would store.

lib/generic-number.c
```diff
@@ -0,0 +1,131 @@
...
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ */
+#include "generic-number.h"
+#include <math.h>
+
+gdouble
+gn_as_double(const GenericNumber *number)
+{
+  if (number->type == GN_DOUBLE)
```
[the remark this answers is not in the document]
bazsi: I would prefer not to touch this at this stage, but I'd be fine converting this to a switch. I don't expect much more type values in the future though.

lib/generic-number.c
```diff
@@ -0,0 +1,131 @@
...
+  number->precision = 0;
+}
+
+gboolean
+gn_is_zero(const GenericNumber *number)
+{
+  if (number->type == GN_INT64)
+    return number->value.raw_int64 == 0;
+
+  return fabs(number->value.raw_double) < DBL_EPSILON;
```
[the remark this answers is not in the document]
bazsi: You are right. I have now pushed a patch that generates an abort() in this case (both here and gn_compare().

lib/generic-number.c
```diff
@@ -0,0 +1,131 @@
...
+}
+
+void
+gn_set_nan(GenericNumber *number)
+{
+  number->type = GN_NAN;
```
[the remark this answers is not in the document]
bazsi: with the new aborts in place I think we can leave this as is.

lib/generic-number.c
```diff
@@ -0,0 +1,131 @@
...
+  if (fabs(l - r) < DBL_EPSILON)
+    return 0;
+  else if (l < r)
+    return -1;
+
+  return 1;
+}
+
+gint
+gn_compare(const GenericNumber *left, const GenericNumber *right)
```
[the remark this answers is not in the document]
bazsi: I've slightly changed this so that if either left or right is NaN we fail with a g_assert_not_reached().

We will need to check in #4024 that this is a fair assumption even with that user in sight.

lib/tests/Makefile.am
bazsi: added, thanks.

lib/filter/filter-cmp.c
bazsi: I have improved the comments both here and on top of _evaluate_typed()

bazsi:
I have now pushed another iteration, this time squashing commits closer to how they should have looked like initially.

I am resolving comments I have answered, please do check your comments and unresolve if you don't agree.

Thanks for all the review and feedback.

bazsi:
pushed a commit to fix the clang warning.

bazsi:
I have one last doubt as indicated on the wonder truth table prepared by @alltilla

truth-table-type-aware
.

In JavaScript Number(null) is indeed zero, but `null` is not automatically converted to a number when doing  a numeric comparison

image

We don't exactly have the equivalent of the Number() function, as int64/int32 casts would not accept "null" as a literal string.

MrAnno:
I just realized this means we don't have exact null checks. A null-value that behaves the "usual way" would be useful IMHO, it could represent unset values too in comparisons.

bazsi:
> I just realized this means we don't have exact null checks. A null-value that behaves the "usual way" would be useful IMHO, it could represent unset values too in comparisons.

Yup, that's why I have my doubts. We could simulate it using the `===` and `!==` operators though. I think I am sending a followup branch with just this change. I'd merge this, it has been open for too long. We are not releasing 4.0 anyway.

bazsi:
I have added a last commit to fix the null cases:

commit 3ab47a3e1b67edcba2f562431655a51e059d903b (HEAD -> type-aware-comparison-operators, origin/type-aware-comparison-operators)
Author: Balazs Scheidler <bazsi77@gmail.com>
Date:   Wed Aug 24 11:55:03 2022 +0200

    filter-cmp: change the truth table when comparing against null
    
    null values are only converted to a numeric value (e.g. 0) if it's compared
    with less than or greater than operators. When checking for equality
    or non-equality, we check if the target value is null too.
    
    Equality and non-equality:
    
    null == null is true
    null != null is false
    <non-null> == null is false
    <non-null> != null is true
    
    Numeric comparisons (null converts to 0 in these cases):
    
    null < 1 is true
    null > -1 is true
    
bazsi:
Let me know if you want me to split the last bit.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/filter/tests/test_filters_fop_cmp.c
```diff
@@ -311,11 +311,33 @@ Test(filter, test_type_aware_comparison_strings_are_compared_as_strings_if_types
...
 {
   cfg_set_version_without_validation(configuration, VERSION_VALUE_4_0);
+  cr_assert(evaluate("null('') == null('')"));
+  cr_assert_not(evaluate("null('') != null('')"));
   cr_assert(evaluate("'$nullvalue' == null('')"));
-  cr_assert(evaluate("'$nullvalue' == 0"));
+  cr_assert_not(evaluate("'$nullvalue' != null('')"));
+  cr_assert(evaluate("string('') != null('')"));
+  cr_assert(evaluate("int64('0') != null('')"));
+  cr_assert(evaluate("double('0.0') != null('')"));
```
gaborznagy, line 323: This could work if the patch supporint numbers in templates with type-hint applied from #4025 :
```
diff --git a/lib/cfg-grammar.y b/lib/cfg-grammar.y
index 8a80adb6c..30f912587 100644
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -804,7 +804,7 @@ template_content_inner
           CHECK_ERROR_GERROR(log_template_compile(last_template, $1, &error), @1, error, "Error compiling template");
           free($1);
         }
-        | LL_IDENTIFIER '(' string ')'
+        | LL_IDENTIFIER '(' string_or_number ')'
         {
           GError *error = NULL;
```

bazsi: This hunk is already in a followup branch (the one to support === and !== operators).

MrAnno (approved):

lib/filter/tests/test_filters_fop_cmp.c
gaborznagy: This is optional, not important. :) Let's keep the string type in the test as You've fixed it:

`cr_assert(evaluate("double('0.0') != null('')"));`

bazsi: I don't understand exactly what you mean here, sorry. Can you elaborate?

gaborznagy (approved):

MrAnno:
@bazsi Thank you for your patience :)

bazsi:
\o/

Type aware operators are IN! The rest is just a piece of cake. Thanks for all the good feedbacks and your help to get this in a shape where it could be merged.

alltilla:
Hurray! :)

lib/filter/tests/test_filters_fop_cmp.c
gaborznagy: It's probably nothing, but this is what I meant:
The grammar doesn't support numbers in a template after a type-hint, so you've changed the template in the unit tests to fix the failing unit tests.

```diff
-  cr_assert(evaluate("int64(0) != null('')"));
-  cr_assert(evaluate("double(0.0) != null('')"));
+  cr_assert(evaluate("int64('0') != null('')"));
+  cr_assert(evaluate("double('0.0') != null('')"));
```

I've seen that the template's grammar rule is changed in #4025 and numbers will be supported in a type-hinted template, therefore I've suggested using it in this PR too.

PR #4025, opened 2022-05-25
Implement === and !== operators (4.0)

bazsi:
This branch implements === and !== operators to verify type AND value at the same time, similarly to the same-named JavaScript
operators.

Question: is it a good idea to have both int32 and int64 types? These operators would differentiate between int64(32) and int32(32), which is probably not what we want.

### DIFF
--- a/lib/filter/filter-expr-grammar.ym
+++ b/lib/filter/filter-expr-grammar.ym
@@ -118,6 +118,7 @@ _translate_number_literals(CfgLexer *lexer, gint compare_mode, LogTemplate *expr
 %left   KW_NOT
 %left   KW_STR_LT KW_STR_LE KW_STR_EQ KW_STR_NE KW_STR_GE KW_STR_GT
 %left   KW_TA_LT KW_TA_LE KW_TA_EQ KW_TA_NE KW_TA_GE KW_TA_GT
+%left   KW_TAV_EQ KW_TAV_NE
 
 %type	<node> filter_expr
 %type	<node> filter_simple_expr
@@ -251,6 +252,8 @@ operator
 	| KW_STR_NE         { $$ = FCMP_STRING_BASED | FCMP_LT | FCMP_GT; }
 	| KW_STR_GE         { $$ = FCMP_STRING_BASED | FCMP_EQ | FCMP_GT; }
 	| KW_STR_GT         { $$ = FCMP_STRING_BASED | FCMP_GT; }
+	| KW_TAV_EQ	    { $$ = FCMP_TYPE_AND_VALUE_BASED | FCMP_EQ; }
+	| KW_TAV_NE         { $$ = FCMP_TYPE_AND_VALUE_BASED | FCMP_LT | FCMP_GT; }
 	;
 
 filter_re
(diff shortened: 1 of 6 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

bazsi:
handle number conversions...

lib/filter/filter-expr-grammar.ym
```diff
@@ -118,6 +118,7 @@ _translate_number_literals(CfgLexer *lexer, gint compare_mode, LogTemplate *expr
 %left   KW_NOT
 %left   KW_STR_LT KW_STR_LE KW_STR_EQ KW_STR_NE KW_STR_GE KW_STR_GT
 %left   KW_TA_LT KW_TA_LE KW_TA_EQ KW_TA_NE KW_TA_GE KW_TA_GT
+%left   KW_TAV_EQ KW_TAV_NE
```
OverOrion, line 121: If `KW_TAV_EQ` stands for type aware comparison, then it should be `KW_TAW_EQ`

OverOrion (commented):
I've checked the last commit (as the preceding commits are in #4022). This looks good, I only have a minor remark.

bazsi:
It denotes type-and-value. Type aware is using "TA" as qualifier.

lib/filter/tests/test_filters_fop_cmp.c
```diff
@@ -357,6 +357,19 @@ Test(filter, test_type_aware_comparison_objects_are_compared_as_strings)
...
+{
+  cfg_set_version_without_validation(configuration, VERSION_VALUE_4_0);
+  cr_assert(evaluate("$strvalue === $strvalue"));
+  cr_assert(evaluate("$strvalue === string"));
+  cr_assert(evaluate("string(64) === string(64)"));
+  cr_assert_not(evaluate("string(64) !== string(64)"));
+  cr_assert_not(evaluate("string(64) === int64(64)"));
+  cr_assert(evaluate("string(64) !== int64(64)"));
+  cr_assert(evaluate("int32(64) !== int64(64)"));
+  cr_assert_not(evaluate("int32(64) === int64(64)"));
```
gaborznagy, line 370: [optional]

The essential parts are here, but I would add some negative cases too, where the types are matching, but the values are not equal:
```
  cr_assert_not(evaluate("foo === bar"));
  cr_assert_not(evaluate("int64(123) === int64(256)"));
  cr_assert_not(evaluate("123 === 456"));
 }
```

gaborznagy (approved):
This looks good to me. When #4022 is merged, this can go in as well.

BTW, there is no news file in either PRs yet. We should not forget these in the 3.38 release.

lib/filter/tests/test_filters_fop_cmp.c
bazsi: I have added these, thanks.

MrAnno:
We're rebasing this one more time + fixing a stylecheck issue so that it could go in.

MrAnno (approved):

PR #4030, opened 2022-05-31
Correct logrotate not reloading syslog-ng

ninjabenji:
The existing logritate config uses a postrotate command of:
  invoke-rc.d syslog-ng reload > /dev/null

This fails for several reasons.

A working equivalent is:
  syslog-ng-ctl reload

This commit simply replaces the existing /etc/logrotate.d/syslog-ng file
with an exact copy that has the failing command swapped out for the
working one.

The consequence of the failing command is that if cron is started,
logrotate will be called and will rotate the /var/logs/messages log
written by syslog-ng but syslog-ng will continue writing to the old log
because it fails to be reloaded.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,6 +15,7 @@ RUN apt-get update -qq && apt-get install -y \
     && rm -rf /var/lib/apt/lists/*
 
 ADD syslog-ng.conf /etc/syslog-ng/syslog-ng.conf
+ADD logrotate_syslog-ng /etc/logrotate.d/syslog-ng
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

alltilla:

Kokan:
The correction should be done at packing level, the docker image created by the dockerfile uses the syslog-ng debian packages (which we build). The file to correct: `packaging/debian/syslog-ng-core.syslog-ng.logrotate`.

PR #4031, opened 2022-05-31
Correct logrotate not reloading syslog-ng

ninjabenji:
The existing logrotate config uses a postrotate command of:
  invoke-rc.d syslog-ng reload > /dev/null

This fails for several reasons.

A working equivalent is:
  syslog-ng-ctl reload > /dev/null

This commit simply swaps out the failing command for a working one

The consequence of the failing command is that if cron is started,
logrotate will be called and will rotate the /var/logs/messages log
written by syslog-ng but syslog-ng will continue writing to the old log
because it fails to be reloaded.

### DIFF
--- a/packaging/debian/syslog-ng-core.syslog-ng.logrotate
+++ b/packaging/debian/syslog-ng-core.syslog-ng.logrotate
@@ -7,7 +7,7 @@
 	delaycompress
 	compress
 	postrotate
-		invoke-rc.d syslog-ng reload > /dev/null
+		syslog-ng-ctl reload > /dev/null
 	endscript
 }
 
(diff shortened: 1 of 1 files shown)


alltilla:

kira-syslogng:
Build SUCCESS

ninjabenji:
@MrAnno Commit msg format updated as requested and signoff added. Hopefully thats all good now. Thanks for the clear instructions for what was required (Im kinda new at pull requests).

MrAnno (approved):

kira-syslogng:
Build SUCCESS

Kokan (approved):

Kokan:
Nice catch, thanks for the fix.

PR #4041, opened 2022-06-10
readme: remove log4j statement & mention APT nightly release

alltilla:


README.md
```diff
@@ -164,22 +164,28 @@ and Ubuntu releases from our APT repository.
...
     ```
     echo "deb https://ose-repo.syslog-ng.com/apt/ stable ubuntu-focal" | sudo tee -a /etc/apt/sources.list.d/syslog-ng-ose.list
     ```
 
 3. Run `apt update`
 
+#### Nightly release
+
+Nightly packages are built and released from the git `master` branch everyday.
+Replace `stable` with `nightly` in step 2 to use the nightly APT repository.
```
MrAnno, line 186: What do you think about this phrasing: "Use `nightly` instead of `stable` in step 2 to ..."

Maybe we could add a one-liner example command.
Another idea: we could indicate that the nightly is for trying out new features/bugfixes, not for production (but I can agree that the name is quite obvious :))

alltilla: Sure, the rephrasing sounds nice, and I can add a one-liner.

alltilla: I am preparing an announcement to Discussions. Maybe we can add the 2 pharagraphs about the pros and cons to the readme, too.

---

From now on nightly builds are available in the `nightly` APT repository.

Newly added features and bugfixes can be obtained quickly with these packages, even before the next stable release.

However, as the name suggests, these packages are not guaranteed to be stable. Between two stable releases, by the nature of continous development, there is no guarantee for any feature to work, so be aware when using the nightly builds.

For instructions on how to install nightly syslog-ng builds on Debian/Ubuntu distributions, see the [README](https://github.com/syslog-ng/syslog-ng/blob/master/README.md#nightly-release).

kira-syslogng:
Build SUCCESS

README.md
MrAnno: I like it. The pro-cons sentences could go to the "Discussions", I would keep the README a bit shorter, for example:


> Nightly packages are built from the git `master` branch every day.
> Nightly builds are considered unstable, we recommend using them only for testing purposes (obtaining new features and bugfixes).

or 

> Nightly builds can be used for testing purposes (obtaining new features and bugfixes) at the risk of breakage.

MrAnno (approved):

Kokan (approved):

kira-syslogng:
Build SUCCESS

PR #4044, opened 2022-06-14
Fix crash in case of kept alive connections

gaborznagy:
We've found a crash when after reload a half-initialized dest-driver (e.g. hostname resolving fails) with kept-alive connections receives a connections close event from the remote.

### DIFF
--- a/modules/afsocket/afsocket-dest.c
+++ b/modules/afsocket/afsocket-dest.c
@@ -558,6 +558,7 @@ afsocket_dd_setup_writer(AFSocketDestDriver *self)
           log_writer_reopen(self->writer, proto);
         }
     }
+  self->connection_initialized = TRUE;
   return TRUE;
 }
 
@@ -567,7 +568,6 @@ afsocket_dd_setup_connection(AFSocketDestDriver *self)
   if (!log_writer_opened(self->writer))
     _dd_reconnect_with_current_addresses(self);
 
-  self->connection_initialized = TRUE;
   return TRUE;
 }
 
--- a/modules/afsocket/afsocket-dest.c
+++ b/modules/afsocket/afsocket-dest.c
@@ -38,6 +38,7 @@
 typedef struct _ReloadStoreItem
 {
   LogProtoClientFactory *proto_factory;
+  GSockAddr *dest_addr;
   LogWriter *writer;
 } ReloadStoreItem;
 
@@ -47,6 +48,7 @@ _reload_store_item_new(AFSocketDestDriver *afsocket_dd)
   ReloadStoreItem *item = g_new(ReloadStoreItem, 1);
   item->proto_factory = afsocket_dd->proto_factory;
   item->writer = afsocket_dd->writer;
+  item->dest_addr = g_sockaddr_ref(afsocket_dd->dest_addr);
   return item;
 }
 
@@ -59,6 +61,7 @@ _reload_store_item_free(ReloadStoreItem *self)
   if (self->writer)
     log_pipe_unref((LogPipe *) self->writer);
 
+  g_sockaddr_unref(self->dest_addr);
   g_free(self);
 }
 
@@ -483,7 +486,7 @@ afsocket_dd_setup_addresses_method(AFSocketDestDriver *self)
 }
 
 static gboolean
-_afsocket_dd_try_to_restore_writer(AFSocketDestDriver *self)
+_afsocket_dd_try_to_restore_connection_state(AFSocketDestDriver *self)
 {
   /* If we are reinitializing an old config, an existing writer may be present */
   if (self->writer)
@@ -501,6 +504,7 @@ _afsocket_dd_try_to_restore_writer(AFSocketDestDriver *self)
   if (_is_protocol_compatible_with_writer_after_reload(self, item))
     self->writer = _reload_store_item_release_writer(item);
 
+  self->dest_addr = g_sockaddr_ref(item->dest_addr);
   _reload_store_item_free(item);
   return TRUE;
 }
@@ -520,7 +524,7 @@ afsocket_dd_construct_writer_method(AFSocketDestDriver *self)
 static gboolean
 afsocket_dd_setup_writer(AFSocketDestDriver *self)
 {
-  gboolean kept_alive_connection = _afsocket_dd_try_to_restore_writer(self);
+  gboolean kept_alive_connection = _afsocket_dd_try_to_restore_connection_state(self);
 
   if (!self->writer)
     {
--- a/modules/afsocket/afsocket-dest.c
+++ b/modules/afsocket/afsocket-dest.c
@@ -482,12 +482,12 @@ afsocket_dd_setup_addresses_method(AFSocketDestDriver *self)
   return TRUE;
 }
 
-static void
+static gboolean
 _afsocket_dd_try_to_restore_writer(AFSocketDestDriver *self)
 {
   /* If we are reinitializing an old config, an existing writer may be present */
   if (self->writer)
-    return;
+    return TRUE;
 
   ReloadStoreItem *item = cfg_persist_config_fetch(
                             log_pipe_get_config(&self->super.super.super),
@@ -496,12 +496,13 @@ _afsocket_dd_try_to_restore_writer(AFSocketDestDriver *self)
   /* We don't have an item stored in the reload cache, which means */
   /* it is the first time when we try to initialize the writer */
   if (!item)
-    return;
+    return FALSE;
 
   if (_is_protocol_compatible_with_writer_after_reload(self, item))
     self->writer = _reload_store_item_release_writer(item);
 
   _reload_store_item_free(item);
+  return TRUE;
 }
 
 LogWriter *
@@ -519,7 +520,7 @@ afsocket_dd_construct_writer_method(AFSocketDestDriver *self)
 static gboolean
 afsocket_dd_setup_writer(AFSocketDestDriver *self)
 {
-  _afsocket_dd_try_to_restore_writer(self);
+  gboolean kept_alive_connection = _afsocket_dd_try_to_restore_writer(self);
 
   if (!self->writer)
     {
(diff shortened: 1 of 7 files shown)


kira-syslogng:
Build FAILURE

modules/afsocket/afsocket-dest.c
```diff
@@ -428,6 +428,7 @@ afsocket_dd_try_connect(AFSocketDestDriver *self)
       afsocket_dd_start_reconnect_timer(self);
       return;
     }
+  g_assert(!iv_timer_registered(&self->reconnect_timer));
   self->reconnect_timer.handler = (void (*)(void *)) afsocket_dd_reconnect;
```
bazsi, line 442: I don't think it is good practice to change the handler of a timer during execution.

I'd either keep state and react accordingly in the same function _OR_ use a separate timer for the two cases.

modules/afsocket/afsocket-dest.c
```diff
@@ -323,6 +323,7 @@ _afsocket_dd_connection_in_progress(AFSocketDestDriver *self)
           msg_error("Syslog connection failed",
                     evt_tag_int("fd", self->fd),
                     evt_tag_str("server", g_sockaddr_format(self->dest_addr, buf, sizeof(buf), GSA_FULL)),
+                    evt_tag_str("group_name", self->super.super.group),
```
bazsi, line 326: why do we add the "group_name" here? It's not a usual practice to add it in similar messages. If we want to make it easier to locate this driver, maybe simply add the location in the config?

modules/afsocket/afsocket-dest.c
```diff
@@ -423,7 +429,11 @@ afsocket_dd_try_connect(AFSocketDestDriver *self)
 {
   if ((!afsocket_dd_setup_addresses(self)) || !afsocket_dd_setup_connection(self))
     {
+      gchar buf[MAX_SOCKADDR_STRING];
+      char *server_address = self->dest_addr ? g_sockaddr_format(self->dest_addr, buf, sizeof(buf), GSA_FULL) : "";
```
bazsi, line 433: I think we asserted on self->dest_addr being non-NULL, why check it here?

modules/afsocket/afsocket-dest.c
```diff
@@ -696,10 +696,13 @@ afsocket_dd_notify(LogPipe *s, gint notify_code, gpointer user_data)
     case NC_WRITE_ERROR:
       log_writer_reopen(self->writer, NULL);
 
+      char *server_address = self->dest_addr ? g_sockaddr_format(self->dest_addr, buf, sizeof(buf), GSA_FULL) : "";
       msg_notice((notify_code == NC_CLOSE) ? "Syslog connection closed" : "Syslog connection broken",
                  evt_tag_int("fd", self->fd),
-                 evt_tag_str("server", g_sockaddr_format(self->dest_addr, buf, sizeof(buf), GSA_FULL)),
+                 evt_tag_str("server", server_address),
```
bazsi, line 714: Do I get it right, that this is a kept-alive connection that we could resolve when the connection was initially established, but then when the reload happens we cannot resolve it anymore? Is that the scenario we are addressing here?

Maybe we could report the unresolvable name in this case, instead of an empty string.

But if this is the case we might need to be fully operational while self->dest_addr is NULL. How come our assertion does not trigger?

bazsi: btw: we seem to have a wrapper struct that we store in PersistConfig, so we could as well save dest_addr there.

it'd be strange that we don't know our own dest address. Or we could as well do a getpeername() call and restore the destination address that way.

modules/afsocket/afsocket-dest.c
```diff
@@ -552,8 +553,19 @@ afsocket_dd_setup_writer(AFSocketDestDriver *self)
...
+  if (kept_alive_connection)
+    {
+      LogProtoClient *proto = log_writer_steal_proto(self->writer);
+
+      if (proto)
+        {
+          self->fd = log_proto_client_get_fd(proto);
+          self->connection_initialized = TRUE;
+          log_writer_reopen(self->writer, proto);
+        }
```
bazsi, line 567: I am not sure this is easier to follow, as this puts an extra burden on the LogWriter's owner. 

We have similar reload logic in afprog for instance.

bazsi (commented):
I had a few notes and at this point I'd love to see alternative solutions to be evaluated.

I don't really like that the LogWriter's owner needs to steal and reopen the proto, the previous assumption on reload was that we simply save the LogWriter instance and we are done.

On a related note (not strictly associated with this PR, but this is where I noticed this): I hate the names "ReloadStoreItem" (what about AFProgramReloadState, or since we are calling the object that encapsulates this state PersistConfig, let's call these also AFProgramPersistState, or whatever). The afsocket specific naming of the same structure is even worse: ReloadStoreItem. I'd have trouble describing what this function does: _reload_store_item_release_writer()

MrAnno:
Considering kept-alive connections or any other objects that we store in the "reload store", we face a problem that does not seem to be solved in general: those objects usually have a shared state with their previous owner or they have a state that is derived from the owner's state (address, proto, transport, fd), and those need to be kept in sync with the new owner too (ideally).

To be consistent with these injected "dependencies", we have a lot of setters/reinitializers in LogReader and LogWriter, such as `reopen(self, proto)`, `set_options(not-only-options, control_pipe)`, `set_queue()`, `close_proto()`, and now `steal_proto()`.

Getting rid of that derived/shared state would be a good fix, but I think that is not possible with the current abstractions.

Either we don't store objects with derived state during reload, or if we do, we need to make that state queryable and restorable. As I can see, the latter is done in this PR, but instead of just querying, we steal and reinject it.

I think LogWriter secretly reopening itself is not a good practice, because LogWriter communicates with its owner (for example, through `log_pipe_notify()`), and that expects a few things to be in sync with the destination driver, which was not done properly due to LogWriter hiding the need for that.

----

In case we don't want to pollute LogReader/LogWriter further with reload-related reinitialization logic, I think we could try fixing this issue in a different way: let's store in the "reload store" something else that does not require reinitialization at all; the object that actually represents what we want to keep alive: LogProto.

I'm not sure this can be done easily, but it would keep the connection alive in a more explicit way, and a new LogWriter instance could be created instead of reopening externally or reopening secretly in its init method.

modules/afsocket/afsocket-dest.c
gaborznagy: Since the assertion that reconnect timer is not running fails in some cases (thanks @kira-syslogng for finding it), I'll refactor the `reconnect_timer`'s handler changing one too.

modules/afsocket/afsocket-dest.c
gaborznagy: I'll get rid of these as we will have the dest address stored across reload.

gaborznagy:
Thanks @bazsi and @MrAnno for your comments!

I'll rework this PR to save the destination address too in case of kept alive connections.

About the new LogWriter steal interface and the responsibility to start the LogWriter: IMHO LogWriter should be started by its owner. We have plenty of places where we use log_writer_reopen(), and LogWriter expects a state of the AFSocketDestriver when it is running, which was missing.
I think LogReader too always has to be opened by its owner.

LogWriter's code is much simpler if we don't hide such complexities inside it.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

gaborznagy:

kira-syslogng:
Build FAILURE

modules/afsocket/afsocket-dest.c
gaborznagy: I've removed the handler-changing during execution. It's part of a massive refactor.

modules/afsocket/afsocket-dest.c
gaborznagy: I've kept `dest_addr` across reloads, so no need of this anymore.

modules/afsocket/afsocket-dest.c
gaborznagy: I've saved the `dest_addr` across reloads. It was tricky to reuse it (AFInetDestDriver now has to check for an open connection in setup_address to avoid blindly dropping it).

I haven't replaced the LogWriter with LogProto in the "reload_item" (i.e. the object in CfgPersist, which we would like to reuse after reload) yet.

I would like to check if it's possible, as it **could remove the need for the new `log_writer_steal_proto()` API**.
(p.s. I still think the asymmetrical `log_writer_init` is not okay, but that's independent of that).

gaborznagy:
I've reached the point where this PR is review-ready again.
@bazsi Can you check the changes, please? I've done a huge refactor in afsocket around the initial connection establishing and the reconnect logic to reduce complexity, and added the destination address to the.

I think we still need to discuss the asymmetrical behaviour of log_writer_init.
The new `log_writer_steal_proto` could be removed if we rather store the LogProto in the CfgPersist (a todo for me).

Some things are left to do, but in the meantime, I would like to have a re-review of the changes:
- you mentioned to rename the ReloadStoreItem,
- save a LogProto instead of a whole LogWriter across reload
- a minor renaming TODO
- check test coverage, add additional test cases

Since a lot of cases for possible regression are not covered by test cases (e.g. I broke the program destination in case of kept-alive) I would be hesitant at this point to merge this, thus I'm leaving the WIP flag on.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

modules/afsocket/afsocket-dest.c
```diff
@@ -558,6 +558,7 @@ afsocket_dd_setup_writer(AFSocketDestDriver *self)
           log_writer_reopen(self->writer, proto);
         }
     }
+  self->connection_initialized = TRUE;
```
MrAnno, line 528: This is used to decide whether we want to keep the connection open during reload.

Since the feature is about keeping the connection open (not the writer, or proto), the original place seemed more correct.

MrAnno (commented):
Nice work!

I have one comment about checking kept-alive connections.

modules/afsocket/afsocket-dest.c
MrAnno: I might be wrong here, but keeping a broken connection does not seem right.

(The queue won't be lost (acquire_queue() uses `cfg_persist` too.)

MrAnno: Yep, I've misunderstood the original code. Even though a broken connection might not be worth keeping, we did that anyway.

MrAnno (approved):

modules/afsocket/afsocket-dest.c
gaborznagy: Sorry for not answering sooner.

Even though you are right that the feature is about keeping the connection across reloads, and now it looks like it's coupled to the LogWriter's initialization only, IMHO this was the case already in the initial commit 006b3d2ac056d29c97a86a177fbb6728a4960661.

You are right that we should rework this kept-alive-across-reload feature to save the writer (thus the connection) only if the connection is established. However, maybe there are other states from the LogWriter that should be saved too even in case of a broken connection, e.g. the `seq_num`.

I'm not planning to do it yet. (Just for transparent communication :) )

PR #4046, opened 2022-06-14
tests/unit: move all of tests/unit to lib/tests

bazsi:
Originally tests/unit was kept to store test programs that were yet to
be using criterion. The migration happened, but the test programs were not
incrementally moved to their new home.

So this patch moves everything at once.

There was even one test program that existed in both locations:
test_pathutils existed both in lib/tests and tests/unit with different
contents.


bazsi:
The reason that this remained separate are probably because of the GPL->LGPL transition.

bazsi:
The license transition would either need an audit on the changes of these programs OR simply set tests/copyright/policy so that lib/tests/ may contain some GPL licensed programs.

Either could work, the first being a bit more tedious. What do you think?

kira-syslogng:
Build SUCCESS

MrAnno:
> Either could work, the first being a bit more tedious. What do you think?

I think this would be a good time to rephrase [COPYING](https://github.com/syslog-ng/syslog-ng/blob/master/COPYING) a bit to reflect what we currently do.

We originally had GPL modules+SCLs, and an LGPL core lib, but that has changed since in favor of open-source principles:
There are LGPL modules (app-parser, basicfuncs, disk-buffer partially, http signals, java, python, secure logging), and GPL modules but with LGPL tests, and LGPL units with GPL tests. I also remember having LGPL SCL blocks.

I think we could state in COPYING that our intention is to use LGPL on codes that we consider reusable, everything else (mainly modules and tests) is GPL, and each unit is marked accordingly in its header file.

bazsi:
I have now fixed the copyright/policy problems, added the ability to add comments to that file and also made an attempt
to clarify the language of COPYING.

MrAnno (dismissed):
LGTM.

Thanks for the readability changes and for the license clarification :)

gaborznagy:
I've found only one small thing, I've opened a PR for it: https://github.com/bazsi/syslog-ng/pull/16
It's a fixup commit, so a rebase is needed anyway.

Looks good to me though!

gaborznagy (approved):

gaborznagy:
@bazsi I've rather pushed the small fix about the redundant `test_pathutils` in the makefiles. :)

MrAnno (approved):

PR #4052, opened 2022-06-21
Use generic-number in riemann (4.0)

bazsi:
This PR changes the riemann driver to use GenericNumber and its associated number parsing functions, instead of using  using strtoll() without error handling.

This is not really a functional change, rather it is a refactor, but has originally been done in the scope of the 4.0 typing work.

### DIFF
--- a/modules/riemann/riemann-worker.c
+++ b/modules/riemann/riemann-worker.c
@@ -25,6 +25,8 @@
 #include "riemann.h"
 #include "riemann-worker.h"
 #include "scratch-buffers.h"
+#include "generic-number.h"
+#include "parse-number.h"
 
 #include <riemann/simple.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


bazsi:
no news file needed.

kira-syslogng:
Build SUCCESS

OverOrion (commented):
I have checked the last commit, and it looks good to me.

I just want to mention that the old behavior was that `g_scii_strtoll()` returned with `0` if it was unable to parse the string as number, meanwhile the new one only calls the setter if it was parsed successfully.

If #4022 gets merged, then it can get merged too (after a rebase, if needed).

kira-syslogng:
Build FAILURE

gaborznagy (approved):

MrAnno (approved):

PR #3295, opened 2020-05-18
afsql: added support to type-hint feature for the values

vivinperis:
fixes : #2429

PR #4053, opened 2022-06-21
Afsql typing support (4.0)

bazsi:
This pull requests improves @vivinperis 's PR #3295 so that it integrates with the latest typing changes.

This only received limited testing at this point.

### DIFF
--- a/modules/afsql/afsql-grammar.ym
+++ b/modules/afsql/afsql-grammar.ym
@@ -130,7 +130,7 @@ dest_afsql_values
 
 dest_afsql_values_build
         : template_content dest_afsql_values_build	{ $$ = g_list_append($2, log_template_ref($1)); log_template_unref($1); }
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build SUCCESS

kira-syslogng:
Build SUCCESS

OverOrion:
Nice catch @mitzkia!

---
I've fiddled around with it a bit and it seems that the trigger to the crash is the `default` keyword in `values`. The old `_free_valid_str_pointers` (in `string_list_free`) had a guards against it:
https://github.com/syslog-ng/syslog-ng/blob/master/lib/string-list.c#L85-L97

I think that a custom `_free_valid_sql_values` function with the same guard could be the solution.

bazsi:
I have pushed a fix (and a test to exercise it)

OverOrion (approved):
LGTM, thank you!

PR #4056, opened 2022-06-24
disk-buffer: add prealloc() option, and change defaults in 4.0

alltilla:
TODO:
- [ ] include the change in the 4.0 news file PR.


kira-syslogng:
Build SUCCESS

MrAnno:
Can we also pre-allocate the full disk-buffer size in case truncation is completely disabled?

alltilla:
Yes we can, I added that, too.

modules/diskq/diskq.c
```diff
@@ -175,6 +175,8 @@ _attach(LogDriverPlugin *s, LogDriver *d)
     self->options.qout_size = 1000;
   if (self->options.truncate_size_ratio < 0)
     _set_default_truncate_size_ratio(self, cfg);
+  if (!cfg_is_config_version_older(cfg, VERSION_VALUE_4_0))
+    self->options.prealloc = TRUE;
```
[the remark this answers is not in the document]
alltilla: Done!

modules/diskq/qdisk.c
```diff
@@ -1167,6 +1167,11 @@ qdisk_start(QDisk *self, const gchar *filename, GQueue *qout, GQueue *qbacklog,
 
   if (new_file)
     {
+      if (self->options->prealloc)
+        {
+          _maybe_truncate_file(self, qdisk_get_maximum_size(self));
```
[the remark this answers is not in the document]
alltilla: Thanks, fixed :)

modules/diskq/qdisk.c
```diff
@@ -1173,6 +1173,11 @@ qdisk_start(QDisk *self, const gchar *filename, GQueue *qout, GQueue *qbacklog,
 
   if (new_file)
     {
+      if (self->options->prealloc)
+        {
+          _truncate_file(self, qdisk_get_maximum_size(self));
```
Kokan, line 1178: The `ftruncate` does not pre allocate space, or it does not have to. It can actually just pretend that there is a file that size and actually can later a write can fail due to not enough space on the disk. (sparse file). There is a `posix_fallocate` method that with a correct arguments actually does write the zeros to the disk. (which probably much slower)

A short demo based on this gist: https://gist.github.com/jpauli/8afec7c4fc2b38f8ff27

```bash
/tmp $ df -h /tmp/
Filesystem      Size  Used Avail Use% Mounted on
tmpfs            16G  388M   16G   3% /tmp
/tmp $ ./gistfile1 1 1073741824 #1gb
now opening /tmp/__allocate_chunk
Using ftruncate
Done !, st_blocks=0
/tmp $ df -h /tmp/
Filesystem      Size  Used Avail Use% Mounted on
tmpfs            16G  388M   16G   3% /tmp
/tmp $ ls -hali /tmp/__allocate_chunk
24116 -rw-r--r-- 1 kokan kokan 1.0G Jul  1 11:33 /tmp/__allocate_chunk
/tmp $ rm /tmp/__allocate_chunk
/tmp $ ./gistfile1 2 1073741824 #1gb, fallocate
now opening /tmp/__allocate_chunk
Using fallocate
Done !, st_blocks=2097152
/tmp $ df -h /tmp/
Filesystem      Size  Used Avail Use% Mounted on
tmpfs            16G  1.4G   15G   9% /tmp
/tmp $ ls -hali /tmp/__allocate_chunk
24117 -rw-r--r-- 1 kokan kokan 1.0G Jul  1 11:34 /tmp/__allocate_chunk
/tmp $
```

modules/diskq/diskq.c
```diff
@@ -178,6 +178,13 @@ _attach(LogDriverPlugin *s, LogDriver *d)
   if (self->options.prealloc < 0)
     self->options.prealloc = disk_queue_config_get_prealloc(cfg);
 
+  if (self->options.truncate_size_ratio < 1 && self->options.prealloc)
+    {
+      msg_error("Cannot disable disk-buffer truncation while preallocation is enabled. "
```
Kokan, line 183: [optional] Would it be possible to obtain which logpipe this is connected for easier diagnostics ?

alltilla: Fixed

modules/diskq/qdisk.c
alltilla: Thanks, fixed!

alltilla:
It seems like macOS does not have `posix_fallocate()` :(

alltilla:
Added a hand-written preallocation for platform, which do not have `posix_fallocate()`.

MrAnno:
I'm just wondering, how long it might take to allocate 100-1000 gigabytes of storage with `posix_fallocate` on an average HDD.
Hopefully, this fallocate version won't take minutes, but the compat version actually writes the disk, which could take a lot of time at startup.

alltilla:
We can decide to not support preallocation at all on platforms, where we cannot guarantee it to be relatively fast.

Kokan:
Please also note that `posix_fallocate` where it implemented, it does not mean it is fast/efficient, as if `fallocate` is not supported it falls back to a code that writes the zeroes (just like your compat solution)(see notes for manpage). It is not only platform dependent, on linux (and possible other unix) it is also file system dependent (like fat32 does not support `fallocate`).  

If we want to avoid slow operation of pre-allocation we should check if `fallocate` present in compile time, and also check runtime the result of `fallocate` and both case fall back not to pre-allocate (I would even say not to truncate the file size.) So it would work like pre-allocation turned off.

alltilla:
On a HDD with ext4, the preallocation takes milliseconds.
On a HDD with NTFS it is really slow (10 GiB took ~5 minutes), and cannot be cancelled, due to my dumb implementation.

alltilla:
`prealloc()`: We will make the default `no`, there will be no `auto` for now.

alltilla:
Updated as we discussed. This is now ready for review.

modules/diskq/diskq.c
```diff
@@ -180,6 +180,14 @@ _attach(LogDriverPlugin *s, LogDriver *d)
...
```
MrAnno, line 185: I think the error message is incorrect, we should state the opposite here:
"Truncation can not be enabled together with preallocation. Please disable truncation if you want to use preallocation."

I think we should disable truncation for the user automatically in case they don't specify the truncate size ratio explicitly (that's probably the 90% of our use cases) but they enabled preallocation . We could show this error in case they explicitly enabled both features.

alltilla: Nice catch!

Fixed the warning and the commit message in the original commit.

Implementing the automatic disable was more complex, so I have done it in new commits. It is complex, because with the current implementation we cannot know if the values are set explicitly in the global options block. It is also complex, because we already have 3 different "default" `truncate-size-ratio()` values (set in 2 different places), absed on the config version. Anyways, I have implemented it, but we can drop it if you do not like it.

MrAnno (dismissed):
Jeez, this was a tough one. Thank you very much for implementing that complicated option validation.

modules/diskq/diskq.c
MrAnno: Wow. Thank you :)

MrAnno:
May I merge this? (We can edit the 4.0 NEWS separately)

alltilla:
@MrAnno I just saw, that I forgot adding `log_pipe_location_tag()` to the new 2 warnings. I would like to rebase this PR, add the tag to the warning logs, then I would be happy, if you could approve and merge it :)

MrAnno:
@kira-syslogng do stresstest

MrAnno (approved):

MrAnno:
(Stress test is out of free disk space)

kira-syslogng:

PR #4064, opened 2022-06-29
afuser: Fix more utmp buffer overruns

dkl:
Hello!

Previously commit 816d2eca2c already fixed an issue with a potential buffer overrun in afuser's utmp buffer handling, but it looks like there are two more issues:
* The `line` buffer was not getting null-terminated if `strncpy()` hit the requested limit of 32 chars without encountering a null terminator in `ut_line`, causing the `open()` and debug/error messages afterwards to potentially overrun the `line` buffer. This can only happen with tty device names >= 32 chars (which is rare, I know :)).
* The `ut_user/ut_name` field was used in debug/error messages as if it was null-terminated (`passed to evt_tag_str()`), but it may also be non-null-terminated, just like `ut_line`.

The two patches should fix this. I've been testing them here both with master and with syslog-ng 3.31.2, on an embedded Linux system.

For testing I used a hackish approach:
* Added some users with names with lengths 31, 32, 33. The user name with 33 chars is truncated to 32 chars in ut->ut_user, as expected. Both cases with 32 chars are not null-terminated.
* Simulated unique tty device names for each user by overwriting `ut_line` in `afuser_dd_queue()`: lengths 31, 32 and 32 (since 33 is not possible to store in ut_line). Both cases with 32 chars are not null-terminated.
* Created symlinks for the fake tty names to a real tty.

The user & tty names with lengths 31 work as before. The ones with length 32 now work as expected too (no more failing to open(), no more junk or worse in error & debug messages). Users > 32 chars may work, since `_utmp_entry_matches()` allows partial matches, but it may also match incorrectly (if there are multiple users or `usertty()` instances with names with common prefixes). The same issue exists with tty names > 32 chars - they're also truncated by utmp, so either `open()` will fail now, or it may open the wrong tty, if there is one that matches the truncated name.

### DIFF
--- a/modules/afuser/afuser.c
+++ b/modules/afuser/afuser.c
@@ -140,7 +140,7 @@ afuser_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options)
     {
       if (_utmp_entry_matches(ut, self->username))
         {
-          gchar line[128];
+          gchar line[5 + G_N_ELEMENTS(ut->ut_line) + 1]; /* /dev/ prefix, line device name, null terminator */
           gchar *p = line;
           int fd;
 
@@ -149,9 +149,11 @@ afuser_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options)
               strcpy(line, "/dev/");
               p = line + 5;
             }
-          else
-            line[0] = 0;
-          strncpy(p, ut->ut_line, MIN(sizeof(ut->ut_line), sizeof(line) - (p - line)));
+
+          /* ut_line may not be null-terminated. Copy it and ensure it's null-terminated. */
+          strncpy(p, ut->ut_line, G_N_ELEMENTS(ut->ut_line));
+          p[G_N_ELEMENTS(ut->ut_line)] = 0;
+
           msg_debug("Posting message to user terminal",
                     evt_tag_str("user", _get_utmp_username(ut)),
                     evt_tag_str("line", line));
--- a/modules/afuser/afuser.c
+++ b/modules/afuser/afuser.c
@@ -84,24 +84,15 @@ _utmp_entry_matches(UtmpEntry *ut, GString *username)
   if (strcmp(username->str, "*") == 0)
     return TRUE;
 
+  /* ut_user/ut_name may not be null-terminated */
 #ifdef SYSLOG_NG_HAVE_MODERN_UTMP
-  if (strncmp(username->str, ut->ut_user, sizeof(ut->ut_user)) == 0)
+  if (strncmp(username->str, ut->ut_user, G_N_ELEMENTS(ut->ut_user)) == 0)
 #else
-  if (strncmp(username->str, ut->ut_name, sizeof(ut->ut_name)) == 0)
+  if (strncmp(username->str, ut->ut_name, G_N_ELEMENTS(ut->ut_name)) == 0)
 #endif
     return TRUE;
 
   return FALSE;
-}
-
-static gchar *
-_get_utmp_username(UtmpEntry *ut)
-{
-#ifdef SYSLOG_NG_HAVE_MODERN_UTMP
-  return ut->ut_user;
-#else
-  return ut->ut_name;
-#endif
 }
 
 G_LOCK_DEFINE_STATIC(utmp_lock);
@@ -112,6 +103,17 @@ afuser_dd_set_time_reopen(LogDriver *s, time_t time_reopen)
   AFUserDestDriver *self = (AFUserDestDriver *) s;
 
   self->time_reopen = time_reopen;
+}
+
+static EVTTAG *
+_evt_tag_utmp_username(UtmpEntry *ut)
+{
+  /* ut_user/ut_name may not be null-terminated */
+#ifdef SYSLOG_NG_HAVE_MODERN_UTMP
+  return evt_tag_mem("user", ut->ut_user, G_N_ELEMENTS(ut->ut_user));
+#else
+  return evt_tag_mem("user", ut->ut_name, G_N_ELEMENTS(ut->ut_name));
+#endif
 }
 
 static void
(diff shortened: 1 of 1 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

modules/afuser/afuser.c
```diff
@@ -154,14 +145,25 @@ afuser_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options)
...
 
+          /* ut_user/ut_name may not be null-terminated. Copy it and add null-terminator manually. */
+#ifdef SYSLOG_NG_HAVE_MODERN_UTMP
+          gchar user[sizeof(ut->ut_user) + 1]; /* user name, null terminator */
+          strncpy(user, ut->ut_user, sizeof(ut->ut_user));
+          user[sizeof(ut->ut_user)] = 0;
+#else
+          gchar user[sizeof(ut->ut_name) + 1]; /* user name, null terminator */
+          strncpy(user, ut->ut_name, sizeof(ut->ut_name));
+          user[sizeof(ut->ut_name)] = 0;
```
Kokan, line 156: Instead of using `evt_tag_str` you could simply use `evt_tag_printf`, which work just like a printf, where you can specify the length of the string (`evt_tag_printf("user", "%.*s", string, string_length)`).

Alternative idea is to create a length restricted version of `evt_tag_str` something like `evt_tag_str_len`, as I see  a few possible candidates for that:

```
lib/logmsg/logmsg.h:522:  return evt_tag_printf(name, "%.*s", (gint) value_len, value);
lib/logproto/logproto-buffered-server.c:140:                            evt_tag_printf("buffer", "%.*s", (gint) state->pending_buffer_end, self->buffer));
lib/logproto/logproto-buffered-server.c:660:                evt_tag_printf("buffer", "%.*s", (gint) buffer_len, buffer));
lib/logproto/logproto-framed-server.c:163:                    evt_tag_printf("header", "%.*s", (gint) (i - self->buffer_pos), &self->buffer[self->buffer_pos]));
lib/logproto/logproto-proxied-text-server.c:97:              evt_tag_printf("parameters", "%.*s", (gint) msg_len, msg));
lib/logproto/logproto-proxied-text-server.c:278:  msg_debug("PROXY protocol header received", evt_tag_printf("line", "%.*s",
lib/logreader.c:454:            evt_tag_printf("input", "%.*s", length, line),
lib/rewrite/rewrite-subst.c:71:                evt_tag_printf("input", "%.*s", (gint) length, value),
lib/rewrite/rewrite-subst.c:83:                evt_tag_printf("input", "%.*s", (gint) length, value),
lib/logmatcher.c:243:                      evt_tag_printf("value", "%.*s", (gint) value_len, value));
lib/logmatcher.c:252:                  evt_tag_printf("value", "%.*s", (gint) value_len, value));
modules/afsmtp/afsmtp.c:361:                 evt_tag_printf("message", "%.*s", buflen, buf));
modules/afsmtp/afsmtp.c:366:                evt_tag_printf("message", "%.*s", buflen, buf));
modules/afsmtp/afsmtp.c:371:                evt_tag_printf("data", "%.*s", buflen, buf));
modules/confgen/confgen-plugin.c:122:            evt_tag_printf("block", "%.*s", (gint) result->len, result->str));
modules/kafka/kafka-dest-driver.c:298:                evt_tag_printf("message", "%.*s", (int) MIN(len, 128), (char *) payload),
modules/kafka/kafka-dest-driver.c:308:                evt_tag_printf("message", "%.*s", (int) MIN(len, 128), (char *) payload),
modules/mqtt/destination/mqtt-destination.c:115:  msg_error("MQTT TLS error", evt_tag_printf("line", "%.*s", (gint) len, str),
modules/mqtt/source/mqtt-source.c:101:  msg_error("MQTT TLS error", evt_tag_printf("line", "%.*s", (gint) len, str),
modules/pseudofile/pseudofile.c:97:  return evt_tag_printf("message", "%.*s%s",
```

Kokan (commented):
Thanks for the pull requests, I have a few notes that should be addressed. Please ask if you have any question.

modules/afuser/afuser.c
```diff
@@ -140,7 +140,7 @@ afuser_dd_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_options)
     {
       if (_utmp_entry_matches(ut, self->username))
         {
-          gchar line[128];
+          gchar line[5 + sizeof(ut->ut_line) + 1]; /* /dev/ prefix, line device name, null terminator */
```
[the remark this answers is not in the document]
dkl: As far as I know `5 + sizeof(ut->ut_line) + 1` is a constant expression, so no VLA (compiling with `-Wvla` doesn't show anything for this), but keeping the `128` would certainly be possible aswell.

modules/afuser/afuser.c
dkl: Ah yea, interesting, I somehow missed `evt_tag_printf()`, but that solves the problem without requiring the copy. How about a wrapper like this:
```
static EVTTAG *
_evt_tag_utmp_username(UtmpEntry *ut)
{
  /* ut_user/ut_name may not be null-terminated */
#ifdef SYSLOG_NG_HAVE_MODERN_UTMP
  return evt_tag_printf("user", "%.*s", (int)sizeof(ut->ut_user), ut->ut_user);
#else
  return evt_tag_printf("user", "%.*s", (int)sizeof(ut->ut_name), ut->ut_name);
#endif
}
```

modules/afuser/afuser.c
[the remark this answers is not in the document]
Kokan: My bad, I did not checked `ut_line` and assumed dynamic size.

modules/afuser/afuser.c
dkl: Regarding `evt_tag_str_len`: Looks like there already is `evt_tag_mem` which basically does exactly that (luckily it prints the memory as ASCII string, not hex dump or such).

Kokan: Thanks for the update, it is fine to do the whole project replace. You can just use `refactor:` or `all: ` (see 5ce893d36c17ad6a78d159c1ff710a971619940b) , sure this is not the *module* but the regex really just checks for a structure `something: something more`.

Kokan (approved):
Thanks for the fix.

OverOrion (approved):
LGTM, thank you very much for these nice and clean fixes!

PR #4067, opened 2022-07-03
add-contextual-data: explicitly free scratch buffers we use while loading

bazsi:
We are using ScratchBuffers while loading the CSV input for
add-contextual-data(), and we do this in the main thread. These
buffers are not returned to the system once we start up in the
anticipation of needing more of these buffers in the future.

### DIFF
--- a/modules/add-contextual-data/context-info-db.c
+++ b/modules/add-contextual-data/context-info-db.c
@@ -23,6 +23,8 @@
 #include "context-info-db.h"
 #include "atomic.h"
 #include "messages.h"
+#include "scratch-buffers.h"
+
 #include <string.h>
 #include <stdio.h>
 #include <sys/types.h>
@@ -320,10 +322,15 @@ context_info_db_import(ContextInfoDB *self, FILE *fp, const gchar *filename,
 
   while (_get_line_without_eol(&line_buf, &line_buf_len, fp))
     {
+      ScratchBuffersMarker marker;
+
       lineno++;
       if (line_buf_len == 0)
         continue;
+
+      scratch_buffers_mark(&marker);
       next_record = contextual_data_record_scanner_get_next(scanner, line_buf, filename, lineno);
+      scratch_buffers_reclaim_marked(marker);
       if (!next_record)
         {
           context_info_db_purge(self);


MrAnno:
`ScratchBuffers` are meant to be used anywhere without requiring explicit deallocation.

For that reason, can we add an additional GC call to the for loop of  `cfg_tree_start()` so that we reuse buffers in each LogPipe's `init()`?

bazsi:
I just started to think if there's a better place to put these calls. Unfortunately putting it around log_pipe_init() wouldn't solve the waste of scratch buffers. Although it might decrease the burden just enough so that we can avoid this coupling.

MrAnno:
I'm actually happy with the fix, I was just wondering if we should add another GC call into ` cfg_tree_start()` so that at least different pipes won't increase the scratch count even more.

`ScratchBuffers` always seemed to be a bit intrusive, but I think that's just the cost of a not-so-complicated GC :)

OverOrion (commented):
(Sorry for taking this long, I had some local trouble regarding my environment, but then I was able to manually test it)

I've generated two CSV files with 10k lines and on `master` stats is
`global;scratch_buffers_bytes;;a;queued;1250251776`

With this PR it goes **drastically** goes down:
`global;scratch_buffers_bytes;;a;queued;32768`

I still want to verify some things, but this looks really promising, thank you!

OverOrion:
How about this?

```diff

diff --git a/lib/scanner/csv-scanner/csv-scanner.c b/lib/scanner/csv-scanner/csv-scanner.c
index 0adc1464f..3c497d321 100644
--- a/lib/scanner/csv-scanner/csv-scanner.c
+++ b/lib/scanner/csv-scanner/csv-scanner.c
@@ -416,6 +416,7 @@ csv_scanner_init(CSVScanner *scanner, CSVScannerOptions *options, const gchar *i
   memset(scanner, 0, sizeof(*scanner));
   scanner->state = CSV_STATE_INITIAL;
   scanner->src = input;
+  scratch_buffers_mark(&scanner->marker);
   scanner->current_value = scratch_buffers_alloc();
   scanner->current_column = NULL;
   scanner->options = options;
@@ -424,6 +425,7 @@ csv_scanner_init(CSVScanner *scanner, CSVScannerOptions *options, const gchar *i
 void
 csv_scanner_deinit(CSVScanner *self)
 {
+  scratch_buffers_reclaim_marked(self->marker);
 }
 
 const gchar *
diff --git a/lib/scanner/csv-scanner/csv-scanner.h b/lib/scanner/csv-scanner/csv-scanner.h
index 1e1cf573c..a5292b9ef 100644
--- a/lib/scanner/csv-scanner/csv-scanner.h
+++ b/lib/scanner/csv-scanner/csv-scanner.h
@@ -27,6 +27,7 @@
 #define CSVSCANNER_H_INCLUDED
 
 #include "syslog-ng.h"
+#include "scratch-buffers.h"
 
 typedef enum
 {
@@ -79,6 +80,7 @@ typedef struct
   const gchar *src;
   GString *current_value;
   gchar current_quote;
+  ScratchBuffersMarker marker;
 } CSVScanner;
 
 const gchar *csv_scanner_get_current_name(CSVScanner *pstate);



```

With this the fact that the `CSVScanner` uses `scratchbuffers` is not leaked out and if anyone else uses `CSVScanner` then they do not need to worry about its `scratchbuffers`.

bazsi:
I was thinking about this, but this might also free buffers unrelated to
the scanner instance.

I think it is better to delegate garbage collection toward the "owner of
the thread" e.g. closer to the global state, and not scatter it around to
objects which may not know everything about specific buffers.

By placing the free operation where I've put them I think I can prove that
the use of buffers does not cross garbage collection. If we put it to the
scanner, it all depends on how the scanner is applied. There might be cases
where a user of a scanner makes independent use of scratch buffers on its
own, and the scanner would free those too.

OverOrion (approved):

gaborznagy:
Sorry for being late, I'm going to finish this in the following days.
I just have to understand some more things around scratch_buffers before I can approve this.

Generally, I agree with your reasoning @bazsi, I just really don't like that CSVScanner's implementation leaks out because of the scratch-buffer GC's limitiation.

bazsi:
A garbage collector is always a global resource/concern. Unless it is
automatic (which scratch buffers is not), it always becomes coupled with
various code paths that are seemingly unrelated. Sometimes you do invoke
the garbage collector in dynamic languagues explicitly, just to indicate
that we expect a lot of garbage to exist.

The good thing in these calls is that they don't really impact
functionality, they might just do nothing. These are usually added as an
"as needed" basis.

As long as you can prove that no reference survives the garbage collection
barrier we are fine. If the need for the barrier disappears, we continue to
be fine.

If this becomes too burdensome, we can always improve the garbage
collector, but I believe that will not be needed, ever.

gaborznagy (approved):

PR #4081, opened 2022-07-16
lib/tlscontext.c: fix libressl build

ffontaine:
Fix the following build failure with libressl raised since version 3.37.1 and https://github.com/syslog-ng/syslog-ng/commit/c7ba645918480da06b9e065c62ebd630312b0c7f:

```
/nvmedata/autobuild/instance-5/output-1/host/lib/gcc/nios2-buildroot-linux-gnu/10.3.0/../../../../nios2-buildroot-linux-gnu/bin/ld: ./lib/.libs/libsyslog-ng.so: undefined reference to `SSL_CTX_set1_client_sigalgs_list'
/nvmedata/autobuild/instance-5/output-1/host/lib/gcc/nios2-buildroot-linux-gnu/10.3.0/../../../../nios2-buildroot-linux-gnu/bin/ld: ./lib/.libs/libsyslog-ng.so: undefined reference to `SSL_CTX_set1_sigalgs_list'
```

Fixes:
 - http://autobuild.buildroot.org/results/84cf9f6a0f6de831a685b73c42eab868952c8b15

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

MrAnno:
Hi @ffontaine,

We don't support LibreSSL explicitly, but we could build our sigalg feature conditionally based on a function-check, for example:
https://github.com/syslog-ng/syslog-ng/blob/76a312818528af211c8014c42729f0767816ecf2/lib/tlscontext.c#L1085-L1095

In that case, we need to propagate the error back to our configuration language, so that we won't just silently ignore the option, but report an error:
https://github.com/syslog-ng/syslog-ng/blob/f5b61f50d9161409f8e3403bc41ec5fd4c3b2bbc/modules/afsocket/afsocket-grammar.ym#L901-L906

ffontaine:
I updated the PR as requested.

MrAnno:
@ffontaine Thank you. Looks good to me.

In case the sigalgs and client-sigalgs config options are not available due to building against libressl (or any other reasons), the user should be notified instead of ignoring their config options silently. That part (the `afsocket-grammar.ym` part above) would be important to implement in my opinion.

Let me know if you're interested in implementing it.
If not, I'll be happy to add the remaining part to your pull request so that we can merge it.

ffontaine:
Thanks for your proposal, I'll let you implement the remaining part.

MrAnno (approved):

alltilla (approved):
Awesome, thanks for this!

PR #4084, opened 2022-07-19
buffered-server: skip invalid characters 

### LINKED ISSUE #4083: Daemon stops reading from network source following charset decoding error
# syslog-ng
## Version of syslog-ng

[root@syslog-ng ~]# syslog-ng --version
syslog-ng 3 (3.37.1)
Config version: 3.35
Installer-Version: 3.37.1
Revision: 
Compile-Date: Jun  8 2022 07:08:52
Module-Directory: /usr/lib64/syslog-ng
Module-Path: /usr/lib64/syslog-ng
Include-Path: /usr/share/syslog-ng/include
Available-Modules: add-contextual-data,affile,afprog,afsocket,afstomp,afuser,appmodel,basicfuncs,cef,confgen,cryptofuncs,csvparser,dbparser,disk-buffer,examples,graphite,hook-commands,json-plugin,kvformat,linux-kmsg-format,map-value-pairs,pacctformat,pseudofile,rate-limit-filter,regexp-parser,sdjournal,stardate,syslogformat,system-source,tags-parser,tfgetent,timestamp,xml,kafka
Enable-Debug: off
Enable-GProf: off
Enable-Memtrace: off
Enable-IPv6: on
Enable-Spoof-Source: on
Enable-TCP-Wrapper: on
Enable-Linux-Caps: on
Enable-Systemd: on
## Platform

Linux syslog-ng 3.10.0-862.el7.x86_64 #1 SMP Fri Apr 20 16:44:24 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
## Debug bundle

# Issue
Daemon stops reading from network source following charset decoding error
## Failure
[2022-07-16T17:05:56.175081] Invalid byte sequence or other error while converting input, skipping character; encoding='GB18030', char='0x97

## Configuration

@version: 3.37
@include "scl.conf"
options {
       flush-lines(10000);
       log-fetch-limit(10000);
       log-fifo-size(20480);
       log-msg-size(204800);
       use-dns(no);
       dns-cache(no);
       create_dirs(yes);
       threaded(yes);
};

source s_network {
        network(
        ip("0.0.0.0")
        port(514)
        transport("udp")
        keep-hostname(no)
    );
};
source s_network_gb {
        network(
        ip("0.0.0.0")
        port(515)
        transport("udp")
        encoding("GB18030")
        keep-hostname(no)
    );
};

destination d_kafka {
  channel {
    rewrite {
      set("${HOST}"    value(".eventv1.host"));
      set("1"          value(".eventv1.@version"));
      set("${R_ISODATE}" value(".eventv1.@timestamp") condition("${.eventv1.@timestamp}" eq ""));
      set("${PROGRAM}" value(".eventv1.program")    condition("${.eventv1.program}" eq ""));
      set("${PID}" value(".eventv1.pid")    condition("${.eventv1.pid}" eq ""));
      set("${LEVEL_NUM}" value(".eventv1.level")    condition("${.eventv1.level}" eq ""));
      set("${MESSAGE}" value(".eventv1.message")    condition("${.eventv1.message}" eq ""));
      set("generic"    value(".eventv1.type")       condition("${.eventv1.type}" eq ""));
    };
    destination {
      kafka-c(config(queue.buffering.max.ms("1000"))
            bootstrap-servers("10.76.230.155:39092,10.76.230.156:39092,10.76.230.157:39092")
            topic("syslog")
            fallback_topic(syslog)
            message("$(format-json --key .eventv1.* --rekey .eventv1.* --shift 9)"));
    };
  };
};

log {
  source(s_network);
  source(s_network_gb);
  destination(d_kafka);
};

alltilla:
### For the reviewers
I think skipping one byte might cause a misalignment in case of wide characters, which could possibly produce jumbled output.
Would it be better to stop processing and return the successfully converted string instead?

---

Fixes #4083

### DIFF
--- a/lib/logproto/logproto-buffered-server.c
+++ b/lib/logproto/logproto-buffered-server.c
@@ -142,10 +142,19 @@ log_proto_buffered_server_convert_from_raw(LogProtoBufferedServer *self, const g
                 }
               break;
             case EILSEQ:
-            default:
-              msg_notice("Invalid byte sequence or other error while converting input, skipping character",
+              msg_notice("Invalid byte sequence, skipping character",
                          evt_tag_str("encoding", self->super.options->encoding),
                          evt_tag_printf("char", "0x%02x", *(guchar *) raw_buffer));
+              raw_buffer++;
+              avail_in--;
+
+              state->pending_buffer_end = state->buffer_size - avail_out;
+              break;
+            default:
+              msg_error("Unknown error while converting input",
+                        evt_tag_error("errno"),
+                        evt_tag_str("encoding", self->super.options->encoding),
+                        evt_tag_printf("char", "0x%02x", *(guchar *) raw_buffer));
               goto error;
             }
         }
--- a/lib/logproto/tests/test-text-server.c
+++ b/lib/logproto/tests/test-text-server.c
@@ -221,6 +221,22 @@ Test(log_proto, test_log_proto_text_server_partial_chars_before_eof)
   log_proto_server_free(proto);
 }
 
+Test(log_proto, test_log_proto_text_server_invalid_char_with_encoding)
+{
+  LogProtoServer *proto;
+
+  log_proto_server_options_set_encoding(&proto_server_options, "GB18030");
+  proto = construct_test_proto(
+            log_transport_mock_stream_new(
+              "foo" "\x80" "bar" "\x80" "baz", -1,
+              LTM_EOF));
+
+  cr_assert(log_proto_server_validate_options(proto),
+            "validate_options() returned failure but it should have succeeded");
+  assert_proto_server_fetch(proto, "foobarbaz", -1);
+  log_proto_server_free(proto);
+}
+
 Test(log_proto, test_log_proto_text_server_not_fixed_encoding)
 {
   LogProtoServer *proto;
--- a/news/bugfix-4084.md
+++ b/news/bugfix-4084.md
@@ -1,0 +1,2 @@
+socket based sources: Fixed a bug, where syslog-ng halted the input instead of skipping a character
+in case of a character conversion error.


bazsi:
I remember commenting on this, but I can't find my review notes, maybe I forgot to submit it.

In any case, I'd love to see some kind of replacement string in the input, instead of just silently dropping the offending characters.

E.g. \xFF or something similar.

alltilla:
> I remember commenting on this, but I can't find my review notes, maybe I forgot to submit it.
> 
> In any case, I'd love to see some kind of replacement string in the input, instead of just silently dropping the offending characters.
> 
> E.g. \xFF or something similar.

I think you meant this comment: https://github.com/syslog-ng/syslog-ng/issues/4083#issuecomment-1197914279

I will add a replacement character, sure :) I am on vacation right now, so I will not get to implement it for 1-2 weeks, so if the PR is needed to be merged, anyone can feel free to push to my branch.

alltilla:
@bazsi Added the '/xFF' character.

I think this PR is ready for review.

lib/logproto/logproto-buffered-server.c
```diff
@@ -142,10 +142,21 @@ log_proto_buffered_server_convert_from_raw(LogProtoBufferedServer *self, const g
...
-            default:
-              msg_notice("Invalid byte sequence or other error while converting input, skipping character",
+              msg_notice("Invalid byte sequence, skipping character",
                          evt_tag_str("encoding", self->super.options->encoding),
                          evt_tag_printf("char", "0x%02x", *(guchar *) raw_buffer));
+              raw_buffer++;
+              avail_in--;
+
+              state->pending_buffer_end = state->buffer_size - avail_out;
+              *((gchar *)(self->buffer + state->pending_buffer_end)) = '\xFF';
```
MrAnno, line 152: 0xFF will cause problems everywhere on the pipeline, I think we should just skip the invalid character.

If we don't want to silently drop invalid byte sequences, I suggest returning additional information to LogReader about the incomplete conversion so that it can inject a tag, name-value pair, or a valid marker to the affected `LogMessage`.

alltilla: Reverted back to the original solution. Feel free to push to this PR any time :)

MrAnno (commented):
LGTM.
Can you add a logproto test for this case?

alltilla:
@MrAnno Yep, added.

MrAnno (approved):
Thank you.

lib/logproto/logproto-buffered-server.c
bazsi: Sorry, I didn't mean to add an 0xFF character, rather a sequence of characters injected that contains the original offeding character in hexadecimal form.

PR #4092, opened 2022-07-29
Light: Dropping working dir concept

mitzkia:
Behind the working dir the main concept was the following:
 - we are running Light from the following base dir: tests/light/
 - in setup phase for every testcase working dir has been calculated: tests/light/reports/date/testcase_name/
 - during the testrun Light will collect/write each testcase related files into the working dir
 - the code uses the path reference for these files with their absoulute or relative paths

The problem with working dir concept:
 - basically with the working dir we reference for the files with their absolute path
 - in some cases this path can not be longer than a certain character length (~160 chars for sockets), which makes
 a hard limit for reports dir and testcase file names also 
 - because of this relative working dir path will be introduced 
 -  because there are so many files related with testcases the concept has been leaked in various
    places of the code which was hard to manage and adds some complexity

The original concept will be replaced with the following:
  - in setup phase of each testcase Light will calculate actual working dir as previously
  - after the working dir has been calculated Light will change (chdir) the actual 'running dir' to the
    working dir (this is new)
  - with this we can reference for all testcase related files as it is placed in the actual dir, so
    working dir and relative working dir can be dropped

### DIFF
--- a/tests/light/functional_tests/conftest.py
+++ b/tests/light/functional_tests/conftest.py
@@ -26,7 +26,6 @@ import os
 import pytest
 from pathlib2 import Path
 
-import src.testcase_parameters.testcase_parameters as tc_parameters
 from src.common.file import copy_file
 from src.common.pytest_operations import calculate_testcase_name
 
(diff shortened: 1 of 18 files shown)


alltilla (approved):
Nice job, approve! :)

I feel like the `TestcaseParameters` class can still be refactored, because it has some globally useful variables, like `INSTANCE_PATH` and `shared_dir`, `install_dir`, etc... and some TC specific variables, like `testcase_name` and `testcase_file`. It has 2 types of accessing variables, one is with a global variable (`INSTANCE_PATH`) the other is through the instantiated object itself. So I feel like there is a bit of mixup here still, but that could be fixed in a separate PR, this PR does a lot of wonders on its own :)

MrAnno:
LGTM. Please squash the last commit before merging.

mitzkia:
@alltilla and @4103  Thanks for the review. 
@alltilla Also thanks for the additional notice about the `TestcaseParameters` class. I am agreed with you, maybe we can refactor that class for better understanding and for easier usage in a new PR.

mitzkia:
I have  squashed my last commit, and rebased the PR against master.

MrAnno (approved):

PR #4093, opened 2022-08-02
add support for custom separators in json-parser() and $(format-json)

bazsi:
Unfortunately, sometimes the input JSON contains keys that have dots in them (e.g. kubernetes labels). In those cases, using
dot notation to represent the input data is ambiguous.

E.g. these JSONs parse to the same representation in syslog-ng:

```
{"foo.bar.baz": "value"}
and
{"foo":{"bar":{"baz": "value"}}}
```

This would not necessarily cause an issue, however, when we $(format-json) this, we would always get the latter even if the input is the former.

The only solution I could come up with was to add support for custom separators, e.g. instead of using dots to represent "levels" of the input, use a special character, which will not clash with other embedded dots.

Here's an example configuration:

```
@version: 3.37

log {
        source { tcp(port(2000) flags(no-parse)); };

        parser { json-parser(key-delimiter("~") prefix("json.")); };

        destination { file("/tmp/foo" template("$(format-json --key-delimiter ~ json.* --shift-levels 1)")); };
};
```

This configuration would use '~' instead of dot to represent the structure.

### DIFF
--- a/lib/value-pairs/cmdline.c
+++ b/lib/value-pairs/cmdline.c
@@ -364,17 +364,13 @@ vp_cmdline_parse_cast(const gchar *option_name, const gchar *value,
   return TRUE;
 }
 
-ValuePairs *
-value_pairs_new_from_cmdline (GlobalConfig *cfg,
-                              gint *argc, gchar ***argv,
-                              gboolean ignore_unknown_options,
-                              GError **error)
+static gboolean
+value_pairs_parse_command_line(ValuePairs *vp,
+                               gint *argc, gchar ***argv,
+                               GOptionGroup *custom_options,
+                               GError **error)
 {
-  ValuePairs *vp;
   GOptionContext *ctx;
-
-  vp = value_pairs_new(cfg);
-
   GOptionEntry vp_options[] =
   {
     {
@@ -443,29 +439,45 @@ value_pairs_new_from_cmdline (GlobalConfig *cfg,
     },
     { NULL }
   };
-  GOptionGroup *og;
   gpointer user_data_args[4];
   gboolean success;
 
-  user_data_args[0] = cfg;
+  user_data_args[0] = vp->cfg;
   user_data_args[1] = vp;
   user_data_args[2] = NULL;
   user_data_args[3] = NULL;
 
   ctx = g_option_context_new("value-pairs");
-  og = g_option_group_new(NULL, NULL, NULL, user_data_args, NULL);
+  GOptionGroup *og = g_option_group_new("value-pairs", "", "", user_data_args, NULL);
   g_option_group_add_entries(og, vp_options);
+
+  /* only the main group gets to process G_OPTION_REMAINING options, so
+   * vp_options is the main one */
+
   g_option_context_set_main_group(ctx, og);
-  g_option_context_set_ignore_unknown_options(ctx, ignore_unknown_options);
+  if (custom_options)
+    g_option_context_add_group(ctx, custom_options);
 
   success = g_option_context_parse(ctx, argc, argv, error);
   vp_cmdline_parse_rekey_finish(user_data_args);
   g_option_context_free(ctx);
+  return success;
+}
 
-  if (!success)
+ValuePairs *
+value_pairs_new_from_cmdline(GlobalConfig *cfg,
+                             gint *argc, gchar ***argv,
+                             GOptionGroup *custom_options,
+                             GError **error)
+{
+  ValuePairs *vp;
+
+  vp = value_pairs_new(cfg);
+
+  if (!value_pairs_parse_command_line(vp, argc, argv, custom_options, error))
     {
       value_pairs_unref(vp);
-      vp = NULL;
+      return NULL;
     }
 
   return vp;
--- a/modules/json/format-json.c
+++ b/modules/json/format-json.c
@@ -37,7 +37,27 @@ typedef struct _TFJsonState
 {
   TFSimpleFuncState super;
   ValuePairs *vp;
+  gchar key_delimiter;
 } TFJsonState;
+
+static gboolean
+_parse_key_delimiter(const gchar *option_name,
+                     const gchar *value,
+                     gpointer data,
+                     GError **error)
+{
+  TFJsonState *state = (TFJsonState *) data;
+
+  if (strlen(value) > 1 || strlen(value) == 0)
+    {
+      g_set_error(error, G_OPTION_ERROR, G_OPTION_ERROR_UNKNOWN_OPTION,
+                  "$(format-json) --key-delimiter only accepts a single character, found: '%s'", value);
+      return FALSE;
+    }
+  /* take off the first character */
+  state->key_delimiter = value[0];
+  return TRUE;
+}
 
 static gboolean
 tf_json_prepare(LogTemplateFunction *self, gpointer s, LogTemplate *parent,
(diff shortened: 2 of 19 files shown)


MrAnno (approved):
LGTM.

It's hard to see at first, but this feature is really interesting and important, as many use-cases require extending/modifying JSON content, but dots may appear in JSON keys, which is also our way to express nesting, so it could alter the original structure. (I think it's worth adding a news entry.)

modules/json/format-json.c
```diff
@@ -51,19 +70,20 @@ tf_json_prepare(LogTemplateFunction *self, gpointer s, LogTemplate *parent,
   GOptionEntry format_json_options[] =
   {
     { "leave-initial-dot", 0, G_OPTION_FLAG_REVERSE, G_OPTION_ARG_NONE, &transform_initial_dot, NULL, NULL },
+    { "key-delimiter", 0, 0, G_OPTION_ARG_CALLBACK, _parse_key_delimiter, NULL, NULL },
```
alltilla, line 73: Can you add some unit tests please?

E.g.:
```c
Test(format_json, test_format_json_with_key_delimiter)
{
  assert_template_format("$(format-json --key-delimiter \"\t\" \".foo\t.b.a.r.\"=\"baz\")",
                         "{\".foo\":{\".b.a.r.\":\"baz\"}}");
  assert_template_format("$(format-json --key-delimiter \".\" \".foo.bar\"=\"baz\")",
                         "{\".foo\":{\"bar\":\"baz\"}}");
  assert_template_format("$(format-json \".foo.bar\"=\"baz\")",
                         "{\"_foo\":{\"bar\":\"baz\"}}");
  assert_template_format("$(format-json --key-delimiter=\"\" \".foo.bar\"=\"baz\")",
                         "{\"_foo\":{\"bar\":\"baz\"}}");
}

```

alltilla (commented):
Nice feature! :)

modules/json/format-json.c
bazsi: I planned to you were just too fast with the review.  :)

lib/value-pairs/cmdline.c
```diff
@@ -443,29 +439,45 @@ value_pairs_new_from_cmdline (GlobalConfig *cfg,
...
   ctx = g_option_context_new("value-pairs");
-  og = g_option_group_new(NULL, NULL, NULL, user_data_args, NULL);
+  GOptionGroup *og = g_option_group_new("value-pairs", "", "", user_data_args, NULL);
   g_option_group_add_entries(og, vp_options);
+
+  /* only the main group gets to process G_OPTION_REMAINING options, so
+   * vp_options is the main one */
+
   g_option_context_set_main_group(ctx, og);
-  g_option_context_set_ignore_unknown_options(ctx, ignore_unknown_options);
```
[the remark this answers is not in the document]
alltilla: I see, thanks :)

modules/json/format-json.c
```diff
@@ -51,19 +70,20 @@ tf_json_prepare(LogTemplateFunction *self, gpointer s, LogTemplate *parent,
...
+  GOptionGroup *og = g_option_group_new("format-json", "", "", state, NULL);
   g_option_group_add_entries(og, format_json_options);
 
-  state->vp = value_pairs_new_from_cmdline (parent->cfg, &argc, &argv, og, error);
+  state->vp = value_pairs_new_from_cmdline(parent->cfg, &argc, &argv, og, error);
   if (!state->vp)
     return FALSE;
 
-  if (transform_initial_dot)
+  if (transform_initial_dot && state->key_delimiter == 0)
```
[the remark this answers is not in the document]
bazsi: Fixed this up, so that --key-delimiter '.' behaves as if key-delimiter was omitted.

modules/json/format-json.c
bazsi: Added these as unit tests. Thanks for writing them.

I've changed some of these as I have changed --key-delimiter behaviour.

bazsi:
I have simplified $(format-flat-json) implementation as the last patch. I can move it to a separate PR if you want me to.

bazsi:
there's something strange going on with the cmake/clang build. I tried to reproduce the crashes of the test suite but without success. I also tried to get a backtrace using the core dumps that github actions produce but without the binaries I wasn't able to get any.

Help appreciated.

alltilla:
I believe cmake/clang faile on master, too, so it is not related to this PR.

If you would like to investigate it, though, you can reproduce each step locally, as the CI runs in devshell.

bazsi:
Here's the crash dump from the unit test. This seems to be a criterion/clang incompatibility issue:

@MrAnno do you have an idea? This seems to be triggered even if I revert my entire branch.

Core was generated by `boxfort-worker'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  __strlen_evex () at ../sysdeps/x86_64/multiarch/strlen-evex.S:77
77	../sysdeps/x86_64/multiarch/strlen-evex.S: No such file or directory.
[Current thread is 1 (Thread 0x7fc34269cac0 (LWP 9362))]
(gdb) bt
#0  __strlen_evex () at ../sysdeps/x86_64/multiarch/strlen-evex.S:77
#1  0x0000559efe0b53c4 in assert_if_tokenizer_concatenated_result_not_match (
    tokenizer=0x559efe0b5260 <__test_strtok_r>, 
    delim=0x558fdb6670ff <error: Cannot access memory at address 0x558fdb6670ff>, 
    input=0x558fdb66713f <error: Cannot access memory at address 0x558fdb66713f>, 
    expected=0x558fdb66712c <error: Cannot access memory at address 0x558fdb66712c>)
    at /__w/syslog-ng/syslog-ng/lib/compat/tests/test_strtok_r.c:68
#2  0x0000559efe0b583f in strtok_with_literals_impl (param=0x7fc34267a288)
    at /__w/syslog-ng/syslog-ng/lib/compat/tests/test_strtok_r.c:124
#3  0x00007fc3438a55e5 in criterion_internal_test_main ()
   from /usr/lib/x86_64-linux-gnu/libcriterion.so.3
#4  0x0000559efe0b5805 in strtok_with_literals_jmp ()
    at /__w/syslog-ng/syslog-ng/lib/compat/tests/test_strtok_r.c:122
#5  0x00007fc3438a4a3e in ?? () from /usr/lib/x86_64-linux-gnu/libcriterion.so.3
#6  0x00007fc3438b1075 in ?? () from /usr/lib/x86_64-linux-gnu/libcriterion.so.3
#7  0x00007fc342ec281d in __libc_start_main (main=0x7fc34389ca70 <main>, argc=1, 
    argv=0x7fffccbb6e08, init=<optimized out>, fini=<optimized out>, 
    rtld_fini=<optimized out>, stack_end=0x7fffccbb6df8) at ../csu/libc-start.c:332
#8  0x0000559efe0b519a in _start ()
(gdb) frame 2
#2  0x0000559efe0b583f in strtok_with_literals_impl (param=0x7fc34267a288)
    at /__w/syslog-ng/syslog-ng/lib/compat/tests/test_strtok_r.c:124
124	/__w/syslog-ng/syslog-ng/lib/compat/tests/test_strtok_r.c: No such file or directory.
(gdb) p param
$1 = (struct strtok_params *) 0x7fc34267a288
(gdb) p *param
$2 = {delim = 0x0, input = 0x0, expected = 0x0}

bazsi:
To add to the fun: if I run the binary that I got from github in my local tree, it runs correctly.

MrAnno:
@bazsi Thanks for looking into it.

This is probably very similar to Snaipe/Criterion#208. I will look for a proper root cause and a fix soon.
Probably a new clang version broke it, it has nothing to do with this PR.

bazsi:
I think this is the issue (from cmake configure logs)

-- Performing Test NO_PIE_AVAILABLE
-- Performing Test NO_PIE_AVAILABLE - Failed

bazsi:
And that's caused by this:
Building C object CMakeFiles/cmTC_db14d.dir/src.c.o
/usr/lib/ccache/clang -DNO_PIE_AVAILABLE  -Werror  -fPIE   -no-pie -MD -MT CMakeFiles/cmTC_db14d.dir/src.c.o -MF CMakeFiles/cmTC_db14d.dir/src.c.o.d -o CMakeFiles/cmTC_db14d.dir/src.c.o -c /build/cmake/CMakeFiles/CMakeTmp/src.c
clang: error: argument unused during compilation: '-no-pie' [-Werror,-Wunused-command-line-argument]

bazsi:
Once I enable ASLR on my local host, I could reproduce the issue locally. I added a workaround as the last patch.

Let's hope it works.

MrAnno:
My suggested fix would be #4095, which removes the manual no-pie checks completely.
If reviewers don't like that, I'll use your fix instead.

Can we remove the CMake fix from this PR, so it can go in separately?

bazsi:
Yes, sure. I am going to rebase my branch on top of yours.

alltilla (approved):

MrAnno (approved):

PR #4096, opened 2022-08-04
.github/workflows/devshell: collect test binaries along with corefiles

bazsi:
This change would collect test binaries along with core files, should the test fail.

This allows me to analyse core files from test programs and avoids the roundtrip of reproducing the test issue locally.

### DIFF
--- a/.github/workflows/devshell.yml
+++ b/.github/workflows/devshell.yml
@@ -80,7 +80,7 @@ jobs:
       - name: make check
         id: make_check
         working-directory: ./build
-        run: make V=1 check || (cat test-suite.log && false)
+        run: make V=1 check || (mkdir -p ${COREFILES_DIR} && find . -executable -a -type f | tar -cf ${COREFILES_DIR}/test-binaries.tar  --files-from=- && cat test-suite.log && false)
 
       - name: make install
         working-directory: ./build


MrAnno (approved):

.github/workflows/devshell.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       - name: make check
         id: make_check
         working-directory: ./build
-        run: make V=1 check || (cat test-suite.log && false)
+        run: make V=1 check || (mkdir -p ${COREFILES_DIR} && find . -executable -a -type f | tar -cf ${COREFILES_DIR}/test-binaries.tar  --files-from=- && cat test-suite.log && false)
```
Kokan, line 83: I think saving the whole build directory could be better.

alltilla, line 83: [optional]

I would rather see a separate artifact gathering and uploading step, which runs on any kind of failure, not just on the `make check` step, but it is up to you.

bazsi: You might be right, but it wasn't missed before and would produce a much larger tarball.

The core files produced by test executables are unusable, with this we stand a slight chance :) but feel free to improve it further.

bazsi: That would be a useful improvement, I agree. But I'd merge this as is at this point.

PR #4111, opened 2022-08-22
github-actions: add correct docker privileges for `distcheck` step

### LINKED ISSUE #4106: distcheck fails in GitHub Actions
This issue is for sharing information about a distcheck failure in one of our GitHub Actions jobs.

https://github.com/syslog-ng/syslog-ng/actions/runs/2878419876

test_patterndb fails only during distcheck, `g_spawn_command_line_sync()` returns a strange error:

```
[----] ../../modules/dbparser/tests/test_patterndb.c:67: Assertion Failed
[----]   
[----]   Error validating patterndb, error=Failed to close file descriptor for child process (Operation not permitted)
[----]   
[FAIL] pattern_db::match_in_program: (0.00s)
[----] ../../modules/dbparser/tests/test_patterndb.c:67: Assertion Failed
[----]   
[----]   Error validating patterndb, error=Failed to close file descriptor for child process (Operation not permitted)
[----]   
...
```

OverOrion:
`g_spawn_command_line_sync()` will make a call to `clone()`, which is not available by default.

By default an unprivileged container runs with the default `seccomp` policy, which has an allowlist for syscalls:
https://github.com/moby/moby/blob/master/profiles/seccomp/default.json#L59

The reason why `g_spawn_command_line_sync()` was unable to run succesfully was its call to `clone()`:

sample program
```C
#include <glib.h>

int main () {

  char *stderr_buff = NULL;
  char *stdout_buff = NULL;

  int exit_status;
  g_spawn_command_line_sync("ls -l", &stdout_buff, &stderr_buff, &exit_status, NULL);
}
```
Compile and check it with `strace`
``` shell
gcc `pkg-config --cflags glib-2.0` main.c -o spawn `pkg-config --libs glib-2.0` && strace -o spawn_strace_output.txt ./spawn
```

Looked for syscalls **NOT** in the [allowlist](https://github.com/moby/moby/blob/master/profiles/seccomp/default.json#L59
):
``` shell
$ cat strace_output.txt | grep -v -E "accept|accept4|...|writev" | cut -f 1 -d '(' | sort -u
clone
```

Docker documentation regarding seccomp:
https://docs.docker.com/engine/security/seccomp/#run-without-the-default-seccomp-profile

Fixes #4106

### DIFF
--- a/.github/workflows/devshell.yml
+++ b/.github/workflows/devshell.yml
@@ -147,6 +147,7 @@ jobs:
     runs-on: ubuntu-18.04
     container:
       image: ghcr.io/syslog-ng/dbld-devshell:latest
+      options: --security-opt seccomp=unconfined --ulimit core=-1
 
     steps:
       - name: Checkout syslog-ng source


MrAnno (approved):

MrAnno:
This is needed to release v3.38, so I'm merging it regardless of my question.

PR #4113, opened 2022-08-24
get rid off separate int32/int64 types

bazsi:
We currently have two types for integers in syslog-ng. int32 and int64.

Because type hints are expected to be used more often in comparisons, separating these two can easily become a source for confusion and bugs.

List of arguments for this change:
  * For instance, int32(16) !== int64(16) since the types are different.
  * We need separate switch cases in drivers to support these two. More often than not, these cases are exactly the same.
  * JavaScript or JSON does not have separate types for 32/64 bit ints.
  * should we want to support BIGINTs we would need a separate type

We were using this type information in just one single driver: MongoDB.
  * mongodb uses a binary encoding of JSON (BSON), which has a separate type for int32 and int64
  * semantically BSON is equivalent to a JSON, which means that the actual encoding for the integer value does not matter to MongoDB, 
  * I resolved this issue by automatically choosing a BSON integer type, based on the integer value we have. If our value to encode is between MININT32 and MAXINT32, then we use 32 bits. Otherwise we will use 64.

This branch remains compatible even with pre-4.0 versions with typing support. This means that even if someone experimented with 4.0 features and typing and have some messages stored in a disk queue, which has LM_VT_INT32 values in it, they will get updated to LM_VT_INTEGER automatically.

Otherwise this patch is just renaming LM_VT_INT32/LM_VT_INT64 to the new LM_VT_INTEGER enum.


OverOrion (approved):
LGTM, sorry for the delay!

bazsi:
\o/

thanks for reviewing and merging this.

PR #4116, opened 2022-08-26
github-actions: build and publish nightly production docker images

alltilla:
`stable` tags will work like they did before.

`nightly` tags will look something like these:
  * `nightly-3.37.1.191.gedca61e-20220826`
  * `nightly-latest`

This is not the actual APT version name, but that would be really hard to get and I think it is close enough.

Example image: https://hub.docker.com/r/alltilla/syslog-ng-prod-alltilla/tags


kira-syslogng:
do nothing -> CI won't start)

MrAnno:

axoflow-dev (approved):
This is awesome. Thanks

PR #4117, opened 2022-08-26
github-actions: build and publish nightly docker image

alltilla:
`stable` tags will work like they did before.
`nightly` tags will look something like these:
  * nightly-3.37.1.191.gedca61e-20220826
  * nightly-latest

This is not the actual APT version name, but that would be really hard to get and I think it is close enough.


bazsi (dismissed):
This looks awesome. Thanks

.github/workflows/publish-docker-image.yml
```diff
@@ -26,15 +36,34 @@ jobs:
...
 
+      - name: Compose Docker image tags
+        id: tags
+        run: |
+          if [[ '${{ inputs.pkg-type }}' = 'stable' ]]; then
+            TAGS='${{ steps.stable-tags.outputs.tag }}'
+          elif [[ '${{ inputs.pkg-type }}' = 'nightly' ]]; then
+            VERSION="$(scripts/version.sh)"
+            DATE="$(date +%Y%m%d)"
+            TAGS="${DOCKER_IMAGE_NAME}:nightly-${VERSION}-${DATE},${DOCKER_IMAGE_NAME}:nightly-latest"
```
MrAnno, line 54: Can we use just `nightly` instead of `nightly-latest`?
I think the phrase `nightly` indicates clearly that it is created each night, `latest` may just cause confusion.

date-tagging nightly releases can be useful, but in that case, we should probably add a cleanup mechanism as well.
For example, can you keep the latest 14 nightly releases and remove the older ones?
(Or alternatively, we can start with just the undated `nightly` tag, and add a N-day history later)

alltilla: I quickly checked the settings on DockerHub, and we do not have the right plan to auto remove old tags, so sure, let's just keep it `nightly` for now, it still brings a lot of benefits.

alltilla: Fixed.

MrAnno: Awesome, thanks :)

news/other-4117.md
```diff
@@ -0,0 +1 @@
+docker image: Nightly production docker images are now available as `balabit/syslog-ng:nightly`
```
MrAnno, line 1: This is good news, I'll add it to our nightly announcement if you don't mind.

alltilla: Sure, thanks :)

.github/workflows/nightly-release.yml
```diff
@@ -57,3 +57,12 @@ jobs:
     uses: ./.github/workflows/test-apt-packages.yml
     with:
       pkg_type: nightly
+
+  publish-docker-image:
+    needs: test-packages
+    uses: ./.github/workflows/publish-docker-image.yml
```
MrAnno, line 63: FYI CDN caches are not invalidated atomically in `index-packages.yml`, so we can expect apt installation issues during the image build.

I'm not sure how big of a problem it is, but we'll see in action.

MrAnno (approved):

PR #4123, opened 2022-08-30
lib: `_free_value()` fix

OverOrion:
As `_map_header_of_entry_from_handle()` may fail, the return value
should be checked.

It could cause a crash, if the the handle or the entry was corrupted.

### DIFF
--- a/lib/persist-state.c
+++ b/lib/persist-state.c
@@ -305,7 +305,8 @@ _free_value(PersistState *self, PersistEntryHandle handle)
       PersistValueHeader *header;
 
       header = _map_header_of_entry_from_handle(self, handle);
-      header->in_use = 0;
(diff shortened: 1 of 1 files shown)


kira-syslogng:
Build FAILURE

OverOrion:

bazsi:
Is there an actual occurrence of this bug? How was this found? Otherwise this looks good to me.

bazsi (approved):

OverOrion:
We've had a crash report, and we found it during the investigation.

(The reason why I opened it as draft was because I wanted to find code paths like this (e.g, the return value of a function that may fail is used without checks), but some of them are not that straightforward to validate, so I will merge this PR, and open a separate one for those.)

PR #4132, opened 2022-09-13
Potential null pointer fixes / checks

OverOrion:
In #4123 the fix was to check the return value of a failable function.
With the help of [Infer](https://github.com/facebook/infer) I was able to find similar cases and fix them.


kira-syslogng:
Build FAILURE

lib/logwriter.c
```diff
@@ -1201,7 +1199,7 @@ log_writer_write_message(LogWriter *self, LogMessage *msg, LogPathOptions *path_
             {
               if (!consumed)
                 {
-                  g_free(self->line_buffer->str);
+                  g_string_free(self->line_buffer, TRUE);
                   log_writer_realloc_line_buffer(self);
```
bazsi, line 1203: this doubles the number of allocations as it will allocate both the GString struct and its "str" member. The aim was to optimize away this case.

A better alternative would be to use scratch-buffers for this purpose, or maybe wrap this ugly trick into a nicer looking function (aka g_string_steal).

Also, this seems to be a use-after-free, you are freeing self->line_buffer and then calling g_string_set_size() on it.

modules/affile/affile-dest.c
```diff
@@ -479,7 +479,6 @@ affile_dd_reuse_writer(gpointer key, gpointer value, gpointer user_data)
   affile_dw_set_owner(writer, self);
   if (!log_pipe_init(&writer->super))
     {
-      affile_dw_set_owner(writer, NULL);
```
bazsi, line 482: I think removing this would cause a leak through the circular reference through the AFFileDestWriter->owner member.

bazsi (changes requested):
I've filed a few comments that should change before this can be merged.

lib/logreader.c
```diff
@@ -389,7 +389,7 @@ log_reader_work_finished(void *s)
       self->notify_code = 0;
       log_pipe_notify(self->control, notify_code, self);
     }
-  if (self->super.super.flags & PIF_INITIALIZED)
+  if ((self->super.super.flags & PIF_INITIALIZED) && self->proto)
```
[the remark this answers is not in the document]
MrAnno: We concluded that `self->proto` being NULL here is allowed, as a pending_close can NULL it out a few lines above.

modules/affile/affile-dest.c
MrAnno: I agree, `affile_dw_set_owner()` is necessary here, but its implementation needs to be fixed in order not to dereference NULL pointers. This is a real issue, we've reproduced the crash.

`affile_dw_unset_owner()` would be even cleaner.

MrAnno:
@OverOrion Do you have plans to finish this PR?

lib/logreader.c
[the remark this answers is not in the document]
OverOrion: This was a rather naive approach, I blindly copied what I had found in `LogWriter`, thanks for catching it!

lib/control/control-command-thread.c
```diff
@@ -91,6 +91,7 @@ control_command_thread_run(ControlCommandThread *self)
   self->thread_finished.cookie = control_command_thread_ref(self);
   iv_event_register(&self->thread_finished);
 
+  g_assert(self->connection);
```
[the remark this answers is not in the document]
OverOrion: Indeed, it was a false positive, thanks once again for the catch!

lib/logwriter.c
OverOrion: I gave it a second thought and refactored the function to take a `ownership_transferred` boolean which decides whether the old buffer should be freed or not. I think it helps with readability as the freeing would happen inside the function instead of the call side.

I am more than happy to revert it, because it does not contribute much to the spirit of the PR either.

modules/affile/affile-dest.c
OverOrion: Thanks once again, `affile_dw_set_owner` got fixed, it should take care of the possible `NULL` owner now.

MrAnno:
Thanks for fixing these. I've reproduced one of the crashes, will review the PR tomorrow.

lib/logwriter.c
```diff
@@ -1203,17 +1206,15 @@ log_writer_write_message(LogWriter *self, LogMessage *msg, LogPathOptions *path_
 
       self->partial_write = (status == LPS_PARTIAL);
 
-      if (consumed)
-        log_writer_realloc_line_buffer(self);
+      log_writer_realloc_line_buffer(self, consumed);
```
MrAnno, line 1209: This does not look like an equivalent transformation to me.
What's the idea behind this patch?

modules/affile/affile-dest.c
```diff
@@ -292,24 +292,28 @@ static void
 affile_dw_set_owner(AFFileDestWriter *self, AFFileDestDriver *owner)
 {
   GlobalConfig *cfg = log_pipe_get_config(&owner->super.super.super);
+  log_pipe_set_config(&self->super, cfg);
```
MrAnno, line 295: This does not fix the NULL pointer deref issue as `log_pipe_get_config` dereferences `owner`.

modules/affile/affile-dest.c
```diff
@@ -316,6 +316,14 @@ affile_dw_set_owner(AFFileDestWriter *self, AFFileDestDriver *owner)
                            self->filename);
 }
 
+static void
+affile_dw_unset_owner(AFFileDestWriter *self)
+{
+  if (self->owner)
+    log_pipe_unref(&self->owner->super.super.super);
+  self->owner = NULL;
```
MrAnno, line 324: Probably we should do something with `self->super.expr_node` too.

MrAnno (changes requested):

lib/logwriter.c
OverOrion: The idea was that the `freeing` was outside the other buffer manipulations and I wanted them to be close to each other (on the call site -> inside the function).

The `consumed` boolean is responsible for deciding whether a specific buffer has been passed yet or not. If it has been passed, then it should not be freed, because the reference is needed (there is something that refers to it).
If it has not been passed (nothing else refers to it), then the old buffer must be freed.

Basically I moved the `g_free` call into the function body.

modules/affile/affile-dest.c
OverOrion: You are right, I messed it up during the rebase :facepalm:  .

modules/affile/affile-dest.c
OverOrion: Upon a closer look it seems that the `expr_node` setting is done without freeing the old one in other places as well.

Should I fix it in this PR or in a separate one?

PR #4133, opened 2022-09-14
db-parser prefix() support

### LINKED ISSUE #4130: Improve patterndb parser support prefix()
## Description of the problem
When using the patterndb parser I would like to prefix the fields extracted 

## Proposed solution
Add "prefix()" to this parser similar to other parsers like csv

bazsi:
This intends to fix #4130

@ryanfaircloth the original idea behind db-parser() was to be the "global" parser for any kind of data. For that reason, I kind of expected all name-value pairs to be specified right within the pattern database. It kind of failed in that mission, today I view db-parser() as a useful alternative to regexps, because it is MUCH faster. But in this role you are right, prefix() would make sense.

These are just a few preparatory patches, but I wanted to submit them so 1) you see there's progress, 2) I can get feedback via the automatic tests.

In the meanwhile, could you please describe your use-case how you'd apply db-parser() to it? It would be useful to learn from. Thanks.

### DIFF
--- a/modules/dbparser/dbparser-grammar.ym
+++ b/modules/dbparser/dbparser-grammar.ym
@@ -165,6 +165,7 @@ grouping_by_opt
             CHECK_ERROR_WITHOUT_MESSAGE(cfg_parser_parse(&filter_expr_parser, lexer, (gpointer *) &filter_expr, NULL), @1);
             grouping_by_set_trigger_condition(last_parser, filter_expr);
           } ')'
+	| KW_PREFIX '(' string ')'				{ grouping_by_set_prefix(last_parser, $3); free($3); };
 	| stateful_parser_opt
 	;
 
--- a/modules/dbparser/dbparser-grammar.ym
+++ b/modules/dbparser/dbparser-grammar.ym
@@ -67,6 +67,7 @@ SyntheticMessage *last_message;
 %token KW_PROGRAM_TEMPLATE
 %token KW_MESSAGE_TEMPLATE
 %token KW_SORT_KEY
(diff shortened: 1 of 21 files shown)


ryanfaircloth:
I'm using it in a way that is close to the original idea. When I extract fields from the message I use a prefix so I can more easily map all of the extracted fields into a destination format ".fields.*" is great when I don't know what the fields will be but I need to map them to json in a destination format

bazsi:
This is now pretty much complete. Tests are running locally and now in GitHub CI.  Let's see if it turns green.

ryanfaircloth:
Looks green

ryanfaircloth:
@bazsi can this be merged?

bazsi:
removed WIP, rebased and added news file. I am trying to get some reviews so this can be merged.

In the meantime can you perhaps test this if it works for you? The CI produces a set of binaries, so you don't have to compile it yourself. Thanks

ryanfaircloth:
@bazsi sure how do I get bits from a PR?

bazsi:
The current run is here

https://github.com/syslog-ng/syslog-ng/actions/runs/3142427727

It is built by the "packages" github action for every PR. You will find the
packages among the artifacts, for every distro built.

ryanfaircloth:
Strange the objects are not downloadable must be a github issue I can try again in a bit

ryanfaircloth:
@bazsi I was able to download the images my current dev machine is now an apple with M1 I was unable to use the rpm the binary RPMs are x86_64

bazsi:
rebased against current master

PR #4137, opened 2022-09-16
http driver compression options

### LINKED ISSUE #3923: HTTP destination driver support gzip compression
## Description of the problem
Users facing metered egress charges can benefit from gzip compressed posts this is a feature of libcurl used by syslog-ng but needs to be exposed in the destination driver to enable.

therandomstring:
Added compression ability to http destination driver using partially built-in libcurl-capabilities (for accepting compressed headers) and own compression algorithm for outward compression using zlib.
Currently supported compression types are gzip and deflate.

resolves #3923


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

bazsi:
Wow. @ryanfaircloth I think you were interested in something like this.

modules/http/http.c
```diff
@@ -261,6 +263,66 @@ http_dd_set_ssl_version(LogDriver *d, const gchar *value)
...
+  if(_http_dd_curl_compression_string_match(type, CURL_COMPRESSION_DEFLATE)) return TRUE;
+#endif
+#if defined(CURL_COMPRESSION_IMPLEMENTED_BROTLI) && defined(BROTLI_AVAILABLE)
+  if(_http_dd_curl_compression_string_match(type, CURL_COMPRESSION_BROTLI)) return TRUE;
+#endif
+#if defined(CURL_COMPRESSION_IMPLEMENTED_ZSTD) && defined(ZSTD_AVAILABLE)
+  if(_http_dd_curl_compression_string_match(type, CURL_COMPRESSION_ZSTD)) return TRUE;
+#endif
+  return FALSE;
+}
```
alltilla, line 289: Seeing that `http-curl-compression.{c,h}` gives both the enum and the string representation, I think the responsibility of check/match/convert between them should be there.

A similar pattern can be seen with `_LogMessageValueType`, `log_msg_value_type_to_str()` and `log_msg_value_type_from_str()` in `lib/logmsg/logmsg.{c,h}`.

modules/http/http-curl-compression.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+#ifndef SYSLOG_NG_HTTP_CURL_COMPRESSION_H
+#define SYSLOG_NG_HTTP_CURL_COMPRESSION_H
+
+/* Uncomment upon implementation */
+#define CURL_COMPRESSION_IMPLEMENTED_GZIP
+#define CURL_COMPRESSION_IMPLEMENTED_DEFLATE
+//#define CURL_COMPRESSION_IMPLEMENTED_BROTLI
+//#define CURL_COMPRESSION_IMPLEMENTED_ZSTD
```
alltilla, line 31: This looks really weird to me.

Why do we make the code more complicated with these flags and also introduce dead code (in a later commit), when we could just implement what we can now and add the rest later? I we wanted to add something that reminds us that there are functionalities left to implement, why no just put 2 lines of TODO comments before `CurlCompressionTypes` and never even introduce their enums, string representations, empty compression functions, etc?

modules/http/http.c
```diff
@@ -261,6 +263,66 @@ http_dd_set_ssl_version(LogDriver *d, const gchar *value)
...
+_http_dd_curl_compression_string_match(const gchar *string, gint curl_compression_index)
+{
+  return (strcmp(string, curl_compression_types[curl_compression_index]) == 0);
+}
+
+gboolean
+_http_dd_check_curl_compression(const gchar *type)
+{
+  if(_http_dd_curl_compression_string_match(type, CURL_COMPRESSION_UNCOMPRESSED)) return TRUE;
+#if defined(CURL_COMPRESSION_IMPLEMENTED_GZIP) && defined(ZLIB_AVAILABLE)
```
alltilla: [optional]

I am reviewing this PR, and going commit-by-commit.

The `..._AVAILABLE` flags were used here prematurely, they will be introduced in a later commit, then in an even later commit another alternative gets introduced and gets used in this function.

A more comprehensible way of building up the git history would be to write the logic of setting the correct flag first (`SYSLOG_NG_HAVE_ZLIB`), then use it in the necessary places.

In the end you have arrived to a correct code, but the way of getting there could be improved in the git history.

modules/http/http-curl-compression.c
```diff
@@ -28,7 +28,63 @@ gint8 CURL_COMPRESSION_DEFAULT = CURL_COMPRESSION_UNCOMPRESSED;
...
+}
+static inline gboolean _raise_compression_status(GString *compression_dest, int algorithm_exit)
 {
   return FALSE;
+}
+
+int _gzip_string(GString *compressed, const GString *message);
+int _deflate_string(GString *compressed, const GString *message);
+int _brotli_string(GString *compressed, const GString *message);
+int _zstd_string(GString *compressed, const GString *message);
```
alltilla, line 85: [optional]

It would be cleaner to implement these functions here, so these lines can be removed.

modules/http/http-curl-compression.c
```diff
@@ -28,7 +28,63 @@ gint8 CURL_COMPRESSION_DEFAULT = CURL_COMPRESSION_UNCOMPRESSED;
...
+
+gboolean http_dd_compress_string(GString *compression_destination, const GString *message, const gint compression)
+{
+  int err_compr;
```
alltilla, line 91: If I understand correctly this will be basically an enum with extra steps. It would be better to introduce the result codes with properly named type. See `LogThreadedResult` in `lib/logthrdest/logthrdestdrv.h`.

alltilla (commented):
This is a really nice feature, good job!

Unfortunately, I did not have the time to finish my review, but I will get back to it later.
In the meantime I think it is worth to share some of my findings, so you have the possibility to start working on them.

modules/http/http-worker.c
```diff
@@ -512,7 +515,21 @@ _curl_perform_request(HTTPDestinationWorker *self, HTTPLoadBalancerTarget *targe
...
+          curl_easy_setopt(self->curl, CURLOPT_POSTFIELDSIZE, self->request_body_compressed->len);
+        }
+      else
+        {
+          curl_easy_setopt(self->curl, CURLOPT_POSTFIELDS, self->request_body->str);
+        }
+
+    }
+  else
+    curl_easy_setopt(self->curl, CURLOPT_POSTFIELDS, self->request_body->str);
```
alltilla, line 532: If the corresponding `if` has curly braces, please add them to the `else` branch, too.

alltilla, line 532: This could go to a separate function.

modules/http/http-worker.c
```diff
@@ -512,7 +515,21 @@ _curl_perform_request(HTTPDestinationWorker *self, HTTPLoadBalancerTarget *targe
...
+    {
+      if(http_dd_compress_string(self->request_body_compressed, self->request_body, owner->message_compression))
+        {
+          curl_easy_setopt(self->curl, CURLOPT_POSTFIELDS, self->request_body_compressed->str);
+          curl_easy_setopt(self->curl, CURLOPT_POSTFIELDSIZE, self->request_body_compressed->len);
+        }
+      else
+        {
+          curl_easy_setopt(self->curl, CURLOPT_POSTFIELDS, self->request_body->str);
+        }
```
alltilla, line 528: [optional]

Should we raise a warning maybe, when we fail to compress?

modules/http/http-worker.c
```diff
@@ -760,6 +777,7 @@ _init(LogThreadedDestWorker *s)
   HTTPDestinationDriver *owner = (HTTPDestinationDriver *) self->super.owner;
 
   self->request_body = g_string_sized_new(32768);
+  self->request_body_compressed = g_string_sized_new(32768);
```
alltilla, line 780: [optional]

We waste memory here, if compression is not set. Can we skip this in that case?

modules/http/http-parser.c
```diff
@@ -64,6 +65,8 @@ static CfgLexerKeyword http_keywords[] =
   { "body_prefix",      KW_BODY_PREFIX },
   { "body_suffix",      KW_BODY_SUFFIX },
   { "delimiter",        KW_DELIMITER },
+  { "accept_encoding",  KW_ACCEPT_ENCODING },
+  { "content_compression",    KW_MESSAGE_COMPRESSION },
```
alltilla, line 69: Shouldn't this be `"message_compression"`?

modules/http/http.h
```diff
@@ -87,6 +88,8 @@ void http_dd_set_cipher_suite(LogDriver *d, const gchar *ciphers);
 gboolean http_dd_set_tls13_cipher_suite(LogDriver *d, const gchar *tls13_ciphers);
 void http_dd_set_proxy(LogDriver *d, const gchar *proxy);
 gboolean http_dd_set_ssl_version(LogDriver *d, const gchar *value);
+void http_dd_set_accept_encoding(LogDriver *d, const gchar *accept);
```
alltilla, line 91: [nitpick]

`const gchar *encoding` might explain the value a bit better here.

modules/http/Makefile.am
```diff
@@ -18,7 +18,9 @@ modules_http_libhttp_la_SOURCES = \
   modules/http/http-parser.c        \
   modules/http/http-parser.h        \
   modules/http/http-plugin.c        \
-  modules/http/http-signals.h
+  modules/http/http-signals.h       \
+  modules/http/http-curl-compression.c \
+  modules/http/http-curl-compression.h
```
alltilla, line 23: These belong to an earlier commit: `http: Added compression files and skeleton`

news/feature-4137.md
```diff
@@ -0,0 +1,14 @@
+`http()`: Added compression ability for use with metered egress/ingress
+
+The new features can be accessed with the 
+ * `accept-encoding()` for inward http messages (currently not used by syslog-ng). The available options are `identity` (for no compression), `gzip` or `deflate` between quotation marks. If you want the driver to accept multiple compression types, you can list them separated by commas, or write `all`, if you want to enable all available compression types.
```
alltilla, line 4: Is this really not used by `syslog-ng`?
If yes, then why do we introduce a feature which does not get used by it?

It cannot be tested, it might break accidentally in the future, and confuses the users with the option.

modules/http/http-curl-compression.c
```diff
@@ -87,4 +98,62 @@ int _zstd_string(GString *compressed, const GString *message)
 {
   //not yet implemented
   g_assert_not_reached();
+}
+
+int _deflate_type_compression(GString *compressed, const GString *message, const gint deflate_algorithm_type)
```
alltilla, line 134: Please add unit tests.

modules/http/http-curl-compression.c
```diff
@@ -0,0 +1,34 @@
+/*
```
alltilla, line 1: I wonder why are these in the `modules/http/` directory. I do no see any HTTP or CURL specific in the final file, it simply implements different compresion strategies. Shouldn't it go under `lib/` instead? Maybe we could use it in other places, too.

modules/http/http-curl-compression.h
```diff
@@ -0,0 +1,51 @@
...
+  CURL_COMPRESSION_BROTLI = 3,
+  CURL_COMPRESSION_ZSTD = 4
+};
+extern gint8 CURL_COMPRESSION_TYPES_LEN;
+
+extern gint8 CURL_COMPRESSION_DEFAULT;
+extern gchar *CURL_COMPRESSION_LITERAL_ALL;
+
+extern gchar *curl_compression_types[];
+gboolean http_dd_compress_string(GString *compression_destination, const GString *message, const gint compression);
```
alltilla, line 49: Passing the `compression` in is not very OOP-style. I could imagine a general `Compression` interface, which would be implemented by `GZipCompression`, `DeflateCompression`, `BrotliCompression`, etc...
The user could call `brotli_compression_compress()` for example.

news/feature-4137.md
MrAnno: accept-encoding() reduces network traffic by making the server compress its response.
It has been tested, we just don't parse/use the response itself currently.

modules/http/http-curl-compression.c
MrAnno: This unit depends on zlib, and will eventually depend on additional compression libraries too.
I think it's best to keep them under modules/http until we find another user.

Note that `zlib` was just a transitive dependency of ours. With this PR, it became a direct dep of the HTTP module (`libsyslog-ng' would be too much and unnecessary at the moment).

modules/http/http-worker.c
MrAnno: Let's make this note optional as it's a matter of taste.

(Mandatory code style things should be agreed upon and included in our astyle check in my opinion.)

news/feature-4137.md
alltilla: Oh, okay, thanks for the clarification. It might be me, but I did not understand this sentence as you explained. Maybe we could rephrase this a bit, but it is optional.

modules/http/http-curl-compression.c
alltilla: Good point, okay, let's keep it here.

However, I still think that we should not "burn in" that it is http/curl specific, if it is not, maybe we could rename the file and the functions, so we can see that it is general.

modules/http/http-worker.c
alltilla: Agreed, sorry.

modules/http/http-worker.c
therandomstring: `http_dd_compress_string()` does raise a warning @ line 109 in `http-curl-compress.c`

modules/http/http-parser.c
therandomstring: I named it content compression because it might be less misunderstandable to the end user this way. To make the code more uniform, I would suggest renaming the keyword and associated variables instead to reflect "content compression".

therandomstring: Done

modules/http/http.h
therandomstring: [antinitpick]

You are right.

therandomstring: Done

modules/http/http-worker.c
alltilla: That is true, however we do not notify the user, that we are falling back to non-compression send. I think it would be worth to do that, because from the current message we do not know whether it gets dropped, or tried again or something else.

modules/http/http-parser.c
alltilla: Thank you. Would it be possible also to rename the setter and the field, too? `http_dd_set_message_compression()`, `gint8 message_compression` Thanks!

modules/http/http-curl-compression.h
therandomstring: Done

kira-syslogng:
Build FAILURE

modules/http/Makefile.am
therandomstring: Originally wanted to name this commit "Added automake support". Will fix in next rebase.

modules/http/http-curl-compression.h
therandomstring: Thait is true, however it fits with C conventions, and does not violate separation of concerns IMHO. If you insist, I could refactor, but I'd rather not.

modules/http/http-curl-compression.c
therandomstring: Could you ellaborate, please?

modules/http/http-curl-compression.c
```diff
@@ -21,13 +21,24 @@
  *
  */
 
+#define _DEFLATE_WBITS_DEFLATE MAX_WBITS
+#define _DEFLATE_WBITS_GZIP MAX_WBITS + 16
```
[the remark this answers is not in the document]
therandomstring: ZLIB magick™... I will change the commit message on the next rebase

modules/http/http-curl-compression.h
alltilla: We are trying to live up to 21st century C standards, doing everything we can in OOP style.
Clean coding and maintainability are really important to the project, and if we can implement the same functionality with more modern philosophy, I think we should.

It is not up to me to decide where we draw the line, though. @MrAnno @bazsi What do you think?

modules/http/http-curl-compression.c
alltilla: You need a multi-value result code.

Your current implementation uses `int` as its type. With this the signatures of the functions do not show the real result type.
A cleaner implementation would define the result type as an enum, which makes the signatures more on point.

There are good examples of the enum style implementation in the code (`LogThreadedResult`, `_LogMessageValueType`). Please consider implementing this like them.

---

Again, sure this works, but we try to conform to cleaner code than old C standards does.

modules/http/http-curl-compression.c
```diff
@@ -87,4 +98,62 @@ int _zstd_string(GString *compressed, const GString *message)
...
+  z_stream _compress_stream = {0};
+  gint err;
+
+  _compress_stream.data_type = Z_TEXT;
+
+  _compress_stream.next_in = (guchar *)message->str;
+  _compress_stream.avail_in = message->len;
+  _compress_stream.total_in = _compress_stream.avail_in;
+
+  g_string_set_size(compressed, (guint)(_compress_stream.avail_in * 1.1) + 22);
```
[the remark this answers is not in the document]
therandomstring: Code change done, will add commit message

modules/http/http-curl-compression.c
```diff
@@ -155,5 +186,19 @@ int _deflate_type_compression(GString *compressed, const GString *message, const
...
+      return _COMPRESSION_ERR_BUFFER;
+    case Z_MEM_ERROR:
+      return _COMPRESSION_ERR_MEMORY;
+    case Z_STREAM_ERROR:
+      return _COMPRESSION_ERR_STREAM;
+    case Z_DATA_ERROR:
+      return _COMPRESSION_ERR_DATA;
+    default:
+      return _COMPRESSION_ERR_UNSPECIFIED;
+    }
```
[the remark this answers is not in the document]
therandomstring: Done

modules/http/http-worker.c
therandomstring: `_handle_compression_error` does write it is sending uncompressed data, as seen in `_compression_error_message' (both in `http-curl-compression.c`), however, this does violate SoC, but that is an other matter.
I will split the warning about the compression error and sending uncompressed data, but only becouse the SoC violation.

therandomstring: Done

modules/http/http-worker.c
therandomstring: The tradeoffs of wasting memory versus the speedloss of extra conditional executions. If you think the first is a bigger problem than the latter, I will add the `if` clause.

modules/http/http-parser.c
therandomstring: I found it more logical to use the "inner name" `message` to the programmer, as in this context it makes more sense to me, but to name it `content` for the end user, as it is less ambiguous. Is such practice allowed, or is it discouraged?

modules/http/http-worker.c
alltilla: It would be great if we had such optimized performance, where an extra `if` in `init()` slowed us down significantly :) However 32K unused memory per worker is a bit more significant IMO.

modules/http/http-curl-compression.c
therandomstring: You are referring to `_gzip_string()` and `_brotli_string()`, right? In those cases it might make sense to return with an enum value, which would enable changing from macros to an enum (`_COMPRESSION_...`).
In the case of `http_dd_compress_string` I wanted to hide the true error case, as that is handled internally, adhering to SoC.

modules/http/http-parser.c
alltilla: I think it is preference.

It is allowed, but you would not see many examples of it, because we usually get confused, when we are debugging a `message-compression()` option, and we do not see a `message_compression` field in the driver. (Talking about personal experience, but I might be the only one with this problem :) )

Feel free to ignore this review comment. You can leave this as is.

modules/http/http-curl-compression.c
```diff
@@ -87,4 +98,62 @@ int _zstd_string(GString *compressed, const GString *message)
...
+      _wbits = _DEFLATE_WBITS_DEFLATE;
+      break;
+    case DEFLATE_TYPE_GZIP:
+      _wbits = _DEFLATE_WBITS_GZIP;
+      break;
+    default:
+      g_assert_not_reached();
+    }
+
+  deflateInit2(&_compress_stream, Z_DEFAULT_COMPRESSION, Z_DEFLATED, _wbits, MAX_MEM_LEVEL, Z_DEFAULT_STRATEGY);
```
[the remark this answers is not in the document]
therandomstring: Done

modules/http/http-worker.c
therandomstring: Makes sense. I'm on it

modules/http/http-parser.c
therandomstring: Valid concern. I will think about it.

modules/http/http-worker.c
therandomstring: Done

modules/http/http-curl-compression.c
therandomstring: I had to start working on an other syslog-ng-related project, but these are coming SOON™. End-to-end tests are also in the pipeline. In the meantime, we've done manual testing with @MrAnno.

news/feature-4137.md
therandomstring: @alltilla do you have any suggestions on a clearer phrasing? I can't think of any better right now, but if you have suggestions, I'm open to ideas.

kira-syslogng:
Build FAILURE

modules/http/http-curl-compression.c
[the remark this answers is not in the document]
therandomstring: Done

modules/http/http-curl-compression.c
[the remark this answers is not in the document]
therandomstring: Done

modules/http/Makefile.am
therandomstring: Done

news/feature-4137.md
alltilla: What do you think about:

* `accept-encoding()` for requesting HTTP response compression from the server. ...

therandomstring: I'll reword it a little bit, but it's a bit clearer tan my original.

therandomstring: If I'm here, I'll rephrase some other things to be in a tad bit more correct English.

therandomstring: Done

modules/http/http-curl-compression.c
```diff
@@ -87,4 +98,62 @@ int _zstd_string(GString *compressed, const GString *message)
...
+    {
+    case DEFLATE_TYPE_DEFLATE:
+      _wbits = _DEFLATE_WBITS_DEFLATE;
+      break;
+    case DEFLATE_TYPE_GZIP:
+      _wbits = _DEFLATE_WBITS_GZIP;
+      break;
+    default:
+      g_assert_not_reached();
+    }
```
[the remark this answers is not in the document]
therandomstring: I think a single relatively short and unambiguous switch-case does not warrant it's own function, since it is only used once, does not hinder readability and does not obfuscate controll-flow. I, however, can be convinced of the of the opposite :)

modules/http/http-curl-compression.c
```diff
@@ -87,4 +98,62 @@ int _zstd_string(GString *compressed, const GString *message)
...
+      if (err != Z_OK && err != Z_STREAM_END)
+        break;
+      if (err == Z_STREAM_END)
+        {
+          err = Z_OK;
+          deflateEnd(&_compress_stream);
+          g_string_set_size(compressed, compressed->len - _compress_stream.avail_out);
+          break;
+        }
+    }
```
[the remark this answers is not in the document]
therandomstring: It is personal preference, but I do not like to `break` and `continue` in the same loop, as it in my opinion makes the controll flow harder to follow. Two conditions in an `if` clause is a perfectly good tradeoff in my book for not having to use `break` and `continue` together. Now, if it was four or five conditions in a clause, that would be an altogether diffrerent beast...

modules/http/http-curl-compression.c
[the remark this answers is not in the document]
alltilla: This function does 4 things as far as I can see:
  1. prepares the compression stream
  2. gets the compression related wbit size
  3. compresses the data
  4. converts result code

When someone will read this code in the future, they will not necessary be interested in all of the impementation details, maybe they are just interested about the high-level steps of this conversion (see the 4 points above). Maybe they will be interested in the prepare step, or the wbit size choosing after they check this function, and if those are moved to separate functions, the reader can jump to it and get into the details of only those things. This is generally endorsed by clean code principles.

In an ideal case this function would only call 4 other functions, one for each point I mentioned. We do not always conform to this of course, and the preference of the developer matters.

In this case I would split this function like I mentioned, this is my feeling/preference, but you can leave it as is if you want.

---

Probably an [optional] tag and this explanation should have been there in my original comment, sorry for that. :/

modules/http/http-curl-compression.c
alltilla: Are you sure the last 2 comments were intended to go here? This thread is about file/function renames, not compression implementations.

modules/http/http-curl-compression.c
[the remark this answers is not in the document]
therandomstring: No problem :). I might refactor it to make easier to understand

modules/http/http-curl-compression.c
therandomstring: You are right. I don't know how they got here :sweat_smile:

modules/http/http-curl-compression.c
therandomstring: Done

modules/http/http-curl-compression.h
therandomstring: Done.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

modules/http/http-curl-compression.c
therandomstring: The OO refactor removed them completely.

modules/http/http-curl-compression.c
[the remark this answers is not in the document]
therandomstring: Done

modules/http/http-curl-compression.c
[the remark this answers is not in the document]
therandomstring: Done

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

MrAnno:
A few HTTP-related e2e tests have failed since the last refactor round:
https://ci.syslog-ng.com/jenkins/job/kira-test-compile-automake-run-testdb/10153/#showFailuresLink

We could take a look at them tomorrow.

MrAnno:

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

modules/http/http.c
alltilla: There are still responsibilites and inner logic leaking from the `compression` module, like exposing and using `curl_compression_types` and calling check manually on the user side.

With your current implementation the users of the compression module need to know inner knowledge about the enum value matching to the position in the `curl_compression_types` array. They can also forget to check if the compression is supported with the user's build, this leaves places for error. I think all of these responsibilities should be in the compression module.

I think everything can be done with only two functions in the compression module:
```
gboolean compression_type_from_str(const gchar *in_string, CurlCompressionType *out_compression_type);
const gchar *compression_type_to_str(CurlCompressionType compression_type);
```

`compression_type_from_str()` can return `FALSE`, if the input string is invalid or the corresponding compression type is not supported with the user's build.
`compression_type_to_str()` is a simple C-style mapping from `enum` to `const char*`.

For an example implementation, see: https://github.com/syslog-ng/syslog-ng/blob/master/lib/logmsg/logmsg.c#L233

modules/http/http-curl-compression.c
alltilla: Please use the enum type on the user side, too:
```
-  gint8 message_compression;
+  CurlCompressionType message_compression;
```

For this, you need to typedef it:
```
-enum CurlCompressionTypes
+typedef enum CurlCompressionType_
 {
   CURL_COMPRESSION_UNCOMPRESSED = 0,
   CURL_COMPRESSION_GZIP = 1,
   CURL_COMPRESSION_DEFLATE = 2
-};
+} CurlCompressionType;
```

modules/http/http-curl-compression.c
alltilla: @therandomstring I still think the compression module is not HTTP or CURL specific. Would you kindly rename the files?

modules/http/http-curl-compression.c
```diff
@@ -0,0 +1,243 @@
...
+
+#define _DEFLATE_WBITS_DEFLATE MAX_WBITS
+#define _DEFLATE_WBITS_GZIP MAX_WBITS + 16
+
+#include "http-curl-compression.h"
+#include "messages.h"
+#include <zlib.h>
+
+gint8 CURL_COMPRESSION_TYPES_LEN = 3;
+gint8 CURL_COMPRESSION_DEFAULT = CURL_COMPRESSION_UNCOMPRESSED;
```
alltilla, line 32: These two are unused.

modules/http/http-curl-compression.h
```diff
@@ -0,0 +1,85 @@
...
+
+typedef enum
+{
+  _COMPRESSION_OK,
+  _COMPRESSION_ERR_BUFFER,
+  _COMPRESSION_ERR_DATA,
+  _COMPRESSION_ERR_STREAM,
+  _COMPRESSION_ERR_MEMORY,
+  _COMPRESSION_ERR_UNSPECIFIED
+} _CompressionUnifiedErrorCode;
```
alltilla, line 52: I think this could go in the C file only, there is no need to make these public.

modules/http/http-curl-compression.h
```diff
@@ -0,0 +1,85 @@
...
+
+typedef struct Compressor Compressor;
+struct Compressor
+{
+  gboolean (*compress) (Compressor *, GString *, const GString *);
+  void (*free_fn) (Compressor *self);
+};
```
alltilla, line 59: The fields of `Compressor`, `GzipCompressor` and `DeflateCompressor` could be made private (moved to the C file), as we have public accessor functions for everything we need publicly.

modules/http/http.c
```diff
@@ -261,6 +263,37 @@ http_dd_set_ssl_version(LogDriver *d, const gchar *value)
...
+    {
+      self->message_compression = CURL_COMPRESSION_UNCOMPRESSED;
+      return;
+    }
+  else if(http_dd_curl_compression_string_match(encoding, CURL_COMPRESSION_GZIP))
+    self->message_compression = CURL_COMPRESSION_GZIP;
+  else if(http_dd_curl_compression_string_match(encoding, CURL_COMPRESSION_DEFLATE))
+    self->message_compression = CURL_COMPRESSION_DEFLATE;
+  gchar *buffer = g_strdup_printf("Content-Encoding: %s", curl_compression_types[self->message_compression]);
+  self->headers= g_list_append(self->headers,  buffer);
```
alltilla, line 294: This should not be appended in the setter, as technically the user could set multiple compression in the config, where the last should be activated only. I think we could move this to the `init()` fn.

modules/http/http.h
```diff
@@ -93,5 +96,7 @@ void http_dd_set_body_suffix(LogDriver *d, const gchar *body_suffix);
 void http_dd_set_delimiter(LogDriver *d, const gchar *delimiter);
 void http_dd_insert_response_handler(LogDriver *d, HttpResponseHandler *response_handler);
 LogTemplateOptions *http_dd_get_template_options(LogDriver *d);
+void http_dd_set_accept_encoding(LogDriver *d, const gchar *encoding);
+void http_dd_set_message_compression(LogDriver *d, const gchar *encoding);
```
alltilla, line 100: Copy-paste error in the input parameter (encoding -> compression)

alltilla:
@therandomstring I have done another review round, kindly consider my comments. Thanks :)

modules/http/http-curl-compression.h
therandomstring: Done

modules/http/http.c
therandomstring: Done

modules/http/http-curl-compression.c
therandomstring: `CURL_COMPRESSION_DEFAULT` was meant to be used as a fallback, but somehow got forgotten. Thank you for bringing this to my attention.

modules/http/http-curl-compression.h
therandomstring: Done

modules/http/http-curl-compression.c
therandomstring: Should be done.

modules/http/http-curl-compression.c
therandomstring: Should be done

kira-syslogng:
Build FAILURE

modules/http/http.h
therandomstring: cURL terminology uses `accept_encoding` to refer to HTTP response compression. Since compressions are a subset  of encodings, I think the naming is correct here. If you insist on changing the argument name, I would also suggest changing the name of `http_dd_set_accept_encoding()` to make the naming more homogeneous.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

modules/http/http-worker.c
```diff
@@ -520,8 +522,23 @@ _curl_perform_request(HTTPDestinationWorker *self, HTTPLoadBalancerTarget *targe
             evt_tag_str("url", target->url));
 
   curl_easy_setopt(self->curl, CURLOPT_URL, target->url);
+  if (owner->message_compression != CURL_COMPRESSION_UNCOMPRESSED)
+    {
+      if(compressor_compress(self->compressor, self->request_body_compressed, self->request_body))
```
alltilla, line 527: [optional]

I think we should add a `g_string_truncate(self->request_body_compressed, 0)` to `_reinit_request_body()`.
The `request_body` is also truncated there, and these 2 variables should change together.

modules/http/tests/test_http-compression.c
```diff
@@ -0,0 +1,265 @@
...
+
+static void
+_init_worker(gchar *compression_type)
+{
+  http_dd_set_message_compression((LogDriver *) driver, compression_type);
+  worker = (HTTPDestinationWorker *) driver->super.workers[0];
+  log_threaded_dest_worker_init((LogThreadedDestWorker *) worker);
+}
+
+/* FIXME: Admittedly a bit of a hack-job.
```
alltilla, line 139: So I played around with this test. To really make this work, we should mock `curl_easy_perform()`, and use the drivers queue, etc... I think it is not worth to test this feature this way, because mimicking the real behavior is complex and could change in the future. At the end of the day this file only tests that the compress fn is called where it needs to be called, and IMO this is overkill. Either we should write proper E2E tests with light or zbs, or just test this manually E2E.

alltilla:
Thanks for the changes, this looks good!

modules/http/compression.c
```diff
@@ -0,0 +1,268 @@
...
+{
+  DEFLATE_TYPE_DEFLATE,
+  DEFLATE_TYPE_GZIP
+};
+
+const gchar *_compression_error_message = "Failed due to %s error.";
+static inline void _handle_compression_error(GString *compression_dest, gchar *error_description)
+{
+  msg_error("compression", evt_tag_printf("error", _compression_error_message, error_description));
+  g_string_free(compression_dest, TRUE);
```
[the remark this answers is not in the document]
therandomstring: No idea, how I missed this one. Of course it is meant to be emptied, not freed.

modules/http/tests/test_http-compression.c
therandomstring: As you remember, we did some manual E2E testing with @MrAnno way back when. A zts test is in the pipeline, but is not part of this PR.

modules/http/tests/test_compression.c
```diff
@@ -0,0 +1,158 @@
...
+  0x42, 0xd6, 0x12, 0xb0, 0xd6, 0xfa, 0x4f, 0x28, 0x03, 0x05, 0xce, 0xb8, 0x2b,
+  0x39, 0xfa, 0x36, 0x84, 0x41, 0x33, 0x8b, 0x98, 0x07, 0xde, 0x5f, 0x2c, 0xc3,
+  0x25, 0x36, 0xc6, 0x64, 0x60, 0xcc, 0x24, 0x9c, 0x73, 0x1c, 0x43, 0x0b, 0xf9,
+  0xde, 0xc8, 0x14, 0x63, 0x9a, 0x16, 0xe9, 0x9b, 0xe2, 0x26, 0x95, 0x47, 0x39,
+  0xea, 0xa0, 0x9d, 0x1b, 0x76, 0x9e, 0xca, 0x4a, 0x28, 0xb2, 0x64, 0xee, 0x6e,
+  0xb3, 0xba, 0x6d, 0xd0, 0x06, 0xa4, 0x89, 0x63, 0xfd, 0x71, 0x8d, 0x76, 0xfc,
+  0x00, 0xbd, 0xc5, 0xb2, 0x98, 0xbd, 0x01, 0x00, 0x00
+};
+
+guint test_message_gzipped_length = 282;
```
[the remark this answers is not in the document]
therandomstring: I agree that it is more elegant that way.

modules/http/tests/test_compression.c
```diff
@@ -0,0 +1,158 @@
...
+// Needed because Criterion cr_assert_str_eq does not like random non-terminating \0s in a string.
+static inline void
+test_compression_results(GString *compression_result, const guint8 *expected_byte_array, const uint expected_size)
+{
+  cr_assert_eq(compression_result->len, expected_size);
+  for(int i = 0; i < compression_result->len; i++)
+    {
+      cr_assert_eq((guint8) compression_result->str[i], expected_byte_array[i]);
+    }
+}
```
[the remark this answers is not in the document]
therandomstring: This was easier to debug. If you insist, I could change it to `memcmp()`.

modules/http/tests/test_compression.c
```diff
@@ -0,0 +1,158 @@
...
+Test(compression, compressor_gzip_compression)
+{
+  replace_gzip_header_os_id(test_message_gzipped_bytes);
+  compressor = gzip_compressor_new();
+  cr_assert_not_null(compressor);
+  result = g_string_new("");
+  compressor_compress(compressor, result, input);
+  test_compression_results(result, test_message_gzipped_bytes, test_message_gzipped_length);
+  compressor_free(compressor);
+  g_string_free(result, 0);
```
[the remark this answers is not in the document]
therandomstring: No idea, why I wrote `0` instead of  `TRUE`...

modules/http/tests/test_http-compression.c
therandomstring: Modifying `_reinit_request_body()` breaks this test on a fundamental level, so it will deleted.

modules/http/compression.c
```diff
@@ -0,0 +1,268 @@
...
+
+#include "compression.h"
+#include "messages.h"
+#include <zlib.h>
+
+gint8 CURL_COMPRESSION_DEFAULT = CURL_COMPRESSION_UNCOMPRESSED;
+gchar *CURL_COMPRESSION_LITERAL_ALL = "all";
+gchar *curl_compression_types[] = {"identity", "gzip", "deflate"};
+
+gboolean http_dd_curl_compression_string_match(const gchar *string, gint curl_compression_index)
```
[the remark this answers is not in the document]
therandomstring: Done

modules/http/tests/test_compression.c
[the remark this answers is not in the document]
therandomstring: Done

modules/http/tests/test_compression.c
```diff
@@ -0,0 +1,158 @@
...
+
+TestSuite(compression, .init = setup, .fini = teardown);
+
+Test(compression, compressor_gzip_compression)
+{
+  replace_gzip_header_os_id(test_message_gzipped_bytes);
+  compressor = gzip_compressor_new();
+  cr_assert_not_null(compressor);
+  result = g_string_new("");
+  compressor_compress(compressor, result, input);
```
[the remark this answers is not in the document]
therandomstring: Added an assert at both occurrence

modules/http/tests/test_compression.c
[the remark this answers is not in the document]
alltilla: It's okay like this. Thanks!

modules/http/http-worker.c
alltilla: Thanks!

modules/http/tests/test_compression.c
```diff
@@ -0,0 +1,158 @@
...
+  test_compression_results(result, test_message_gzipped_bytes, test_message_gzipped_length);
+  compressor_free(compressor);
+  g_string_free(result, 0);
+}
+
+Test(compression, compressor_deflate_compression)
```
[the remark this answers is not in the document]
alltilla: Thanks!

modules/http/http.c
```diff
@@ -261,6 +263,37 @@ http_dd_set_ssl_version(LogDriver *d, const gchar *value)
   return TRUE;
 }
 
+void
+http_dd_set_accept_encoding(LogDriver *d, const gchar *encoding)
+{
+  HTTPDestinationDriver *self = (HTTPDestinationDriver *) d;
+
+  if(strcmp(encoding, CURL_COMPRESSION_LITERAL_ALL) == 0) self->accept_encoding = "";
+  else self->accept_encoding = g_strdup(encoding);
```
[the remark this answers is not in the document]
alltilla: Thanks!

alltilla (approved):
Amazing feature, good job!

PR #4145, opened 2022-09-20
GSoC'22: Log File Rotation

### LINKED ISSUE #2964: Rotate log files based on size

shikharvashistha:
## Description

This pull request is intended to introduce a log file rotation mechanism for the syslog-ng application which will solve the current issue of log files becoming very large and thus affecting the performance of the application.

## Overview of changes

The following changes have been made in order to implement the solution:

* LogProtoFileWriter class have been modified to connect to the existing SignalSlotConnector class in order to receive notifications via `log_proto_file_writer_new()` function when a log file is opened and will carry filename and reopener pointer as user_data.

* A new slot have been added to the SignalSlotConnector class which will receive notifications when a log file is opened and will call a new function `log_proto_file_writer_reopen()` in order to check if the opened file needs to be rotated and all the necessary logic is applied to rotate the log file for which the file-rotation functionality was built.

### Whole configuration file

```
# Default syslog-ng.conf file which collects all local logs into a
# single file called /var/log/messages.
#

@version: 3.37
@include "scl.conf"

destination d_file {
    file("/tmp/temp.log", file-rotation(size(0KB), interval("daily"), dateformat("-%Y-%m-%d"), rotate(2)));
};

source my_tcp {
    stdin();
};


log {
source(my_tcp);
destination(d_file);
};
```
### Necessary changes

```
destination d_file {
    file("/tmp/temp.log", file-rotation(size(0KB), interval("daily"), dateformat("-%Y-%m-%d"), rotate(2)));
};
```

## Alternatives/Problem faced

1. lseek vs fstat call. Can refer the discussion [here](https://github.com/MrAnno/syslog-ng/pull/18#discussion_r971037562)

2. Attaching a reopner function to the LogProtoFileWriter class vs using a global flag to signal file-rotation plugin to rotate the file. 

## Discussions

*  [Pull Request - 1 ](https://github.com/MrAnno/syslog-ng/pull/17)

* [Pull Request - 2](https://github.com/MrAnno/syslog-ng/pull/18)

Resolves #2964


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

kira-syslogng:

modules/affile/logproto-file-writer.c
```diff
@@ -32,6 +33,9 @@
 typedef struct _LogProtoFileWriter
 {
   LogProtoClient super;
+  SignalSlotConnector *signal_slot_connector;
```
bazsi, line 36: this should probably named to resemble the functionality that is associated with the slot, e.g. file_rotation_signal or something like that.

modules/affile/affile-dest.c
```diff
@@ -162,8 +169,10 @@ affile_dw_reopen(AFFileDestWriter *self)
 
       LogTransport *transport = file_opener_construct_transport(self->owner->file_opener, fd);
 
+      FileReopener reopener = { .request_reopen = _file_reopen_cb, .cookie = self};
       proto = file_opener_construct_dst_proto(self->owner->file_opener, transport,
-                                              &self->owner->writer_options.proto_options.super);
+                                              &self->owner->writer_options.proto_options.super, self->owner->super.super.super.signal_slot_connector, self->filename,
```
bazsi, line 174: the slot is probably better placed in the FileOpener constructor so that if an implementation of construct_dst_proto needs it it can find it in self and we wouldn't need to change the prototype here.

I guess this only applies to regular files and not so much to /proc/kmsg or named pipes.

modules/affile/file-signals.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "signal-slot-connector/signal-slot-connector.h"
+
+typedef struct _FileFlushSignalData FileFlushSignalData;
+typedef struct _FileReopener FileReopener;
+
+struct _FileFlushSignalData
+{
+  const gchar *filename;
+  FileReopener *reopener;
```
bazsi, line 35: I think we could allow the file opener plugin to indicate that a reopen is needed via a return value, so we wouldn't have to push the FileReopener instance down to the guts of the rotation plugin.

This way the reopen would  either be delegated to the LogProtoFileWriter or even better back to the driver via the use of the log_pipe_notify() mechanism.

shikharvashistha: Do you mean that we can have a returning flag of some sort for the purpose of reopening the file ?

modules/affile/logproto-file-writer.c
```diff
@@ -218,7 +234,8 @@ log_proto_file_writer_prepare(LogProtoClient *s, gint *fd, GIOCondition *cond, g
 }
 
 LogProtoClient *
-log_proto_file_writer_new(LogTransport *transport, const LogProtoClientOptions *options, gint flush_lines, gint fsync_)
+log_proto_file_writer_new(LogTransport *transport, const LogProtoClientOptions *options, gint flush_lines, gint fsync_,
+                          SignalSlotConnector *connector, const gchar *filename, FileReopener reopener)
```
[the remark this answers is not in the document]
shikharvashistha: Okay sure

modules/affile/logproto-file-writer.c
shikharvashistha: I've updated the name

shalgf:
Dear shikharvashistha:

Thank you for your feature Log File Rotation, it is very helpful for me.
I get your submittion and compile, found that there is not file-rotation in Makefile.in,
and there is not files file-rotation-grammar.y, file-rotation-grammar.c and file-rotation-grammar.h,
Will you please tell me how can I generate them?

Thanks
shalgf

tuxillo:
What happened to this ?

mehul-m-prajapati:
@HofiOne @bazsi : I can work on the remaining part of this feature. Do you want to have this log rotation ?

czanik:
Thanks @mehul-m-prajapati for your offer. However, we discussed this question internally, and we do not want to continue this effort. If we ever get to this, it will be a different implementation.

PR #4147, opened 2022-09-22
Add Docker image for arm64, update actions

lucacome:
Signed-off-by: Luca Comellini <luca.com@gmail.com>

Adds Docker images for arm64

### DIFF
--- a/.github/workflows/publish-docker-image.yml
+++ b/.github/workflows/publish-docker-image.yml
@@ -25,10 +25,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout syslog-ng source
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Log in to Docker Hub
-        uses: docker/login-action@v1
+        uses: docker/login-action@v2
         with:
           username: ${{ secrets.dockerhub-username }}
           password: ${{ secrets.dockerhub-password }}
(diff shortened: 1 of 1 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

MrAnno:
Hi @lucacome,

Unfortunately, our Docker images are based on the [APT repository](https://github.com/syslog-ng/syslog-ng#adding-the-apt-repository) we provide, which currently does not support arm64.
(Publishing aarch64 packages would be the first step of supporting arm64 Docker images.)

MrAnno:
@lucacome Those are official Debian packages lagging behind a few versions, not ours.

lucacome:
yeah @MrAnno but that's what the Dockerfile in the project seems to use...

MrAnno:
The purpose of our Dockerfile (and Docker images) would be to always use the latest version of syslog-ng, which is coming from `https://ose-repo.syslog-ng.com/apt/`, an APT repo that is built and hosted by us, syslog-ng developers.

That repository currently does not support aarch64, so with your change, the build of the image falls back using the base image's official repo: `deb.debian.org`. I consider this as an unwanted side-effect because the official Debian repository contains older versions of syslog-ng, and the way they package syslog-ng is slightly different than ours.

lucacome:
ah that makes sense, thanks for the explanation!

MrAnno:
Let's reopen this PR once we have ARM64 packages.

MrAnno:
FYI cloud-ready syslog-ng images are available by Axoflow (arm64 too):
https://github.com/axoflow/syslog-ng-docker/pkgs/container/syslog-ng

Issue #4155, opened 2022-09-27
add-context bug with type hints


ryanfaircloth:
# syslog-ng
## Version of syslog-ng
3.38.1 with beta features enabled

```txt
[2022-09-27T14:11:46.579838] add-contextual-data(): the failing line is; input='"tcp_13",name,"Daytime (RFC 867)"', filename='/etc/syslog-ng/core/lib/parsers/p_well_known_ports/p_well_known_ports.csv:14'
```

it seems use of () in the value field is problematic

bazsi:
You are right. As a workaround try enclosing the value into string() that should satisfy the parser and force it to the right type.

I am not sure if this should become the final solution, as this is a slight incompatibility with older versions. I think I'll replace that error with a warning and a recommendation to add the explicit type cast to string.

I think that would work. Alternatively I could introduce a 4 column format but a very similar issue exists in db-parser() where I extended the syntax the same way and there's no separate column (although I could add the type information as an XML attribute in that case).

ryanfaircloth:
Could we assume string for values that are quoted ?

bazsi:
Unfortunately that quotation is processed by the CSV parser.
add-contextual-data() does not know if a column was escaped or not.

ryanfaircloth:
The work around seems to work, just removing () from the string also works

bazsi:
I ended up adding warnings that get emitted in case you are using an old config version and template expressions that would conflict with the new syntax. This is a sample log message:


```
WARNING: the value field in add-contextual-data() CSV files has been changed to support typing from syslog-ng 4.0. You are using an older config version and your CSV file contains an unrecognized type-cast, probably a parenthesis in the value field. This will be interpreted in the `type(value)' format in future versions. Please add an explicit string() cast as shown in the 'fixed-value' tag of this log message or remove the parenthesis; config-version='3.38', selector='bzorp/mail', name='pid', value='foobar(1224)', fixed-value='string(foobar(1224))'
```

PR #4158, opened 2022-09-29
Type hinting compat fixes

bazsi:
This branch adds a couple of new warnings where the new type-hinting functionality causes incompatibilities.

These are db-parser(), grouping-by() and add-contextual-data() where the template strings can now contain type-hints, which were not supported before.

These cause an incompatibility if the previous templates did contain a parenthesis, as encountered by @ryanfaircloth in issue #4155

### DIFF
--- a/modules/add-contextual-data/contextual-data-record-scanner.c
+++ b/modules/add-contextual-data/contextual-data-record-scanner.c
@@ -94,6 +94,10 @@ _fetch_value(ContextualDataRecordScanner *self, ContextualDataRecord *record)
 
   record->value = log_template_new(self->cfg, NULL);
 
+
+  GError *error = NULL;
+  gboolean success;
+
   if (cfg_is_config_version_older(self->cfg, VERSION_VALUE_3_21) &&
       strchr(value_template, '$') != NULL)
     {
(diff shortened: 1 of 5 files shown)


modules/add-contextual-data/contextual-data-record-scanner.c
```diff
@@ -103,26 +107,58 @@ _fetch_value(ContextualDataRecordScanner *self, ContextualDataRecord *record)
...
+          if (!success)
+            {
+              log_template_set_type_hint(record->value, "string", NULL);
+              msg_warning("WARNING: the value field in add-contextual-data() CSV files has been changed "
+                          "to support typing from " FEATURE_TYPING_VERSION ". You are using an older config "
+                          "version and your CSV file contains an unrecognized type-cast, probably a "
+                          "parenthesis in the value field. This will be interpreted in the `type(value)' "
+                          "format in future versions. Please add an "
+                          "explicit string() cast as shown in the 'fixed-value' tag of this log message "
+                          "or remove the parenthesis",
```
alltilla, line 131: Shouldn't we write something about the fact that although we throw a warning,we continue with `string` type hint?

modules/dbparser/synthetic-message.c
```diff
@@ -90,17 +90,51 @@ synthetic_message_add_value_template_string(SyntheticMessage *self, GlobalConfig
...
+              log_template_set_type_hint(value_template, "string", NULL);
+
+              msg_warning("WARNING: the template specified in value()/<value> options for your grouping-by() or "
+                          "db-parser() configuration has been changed to support typing from "
+                          "" FEATURE_TYPING_VERSION ". You are using an older config "
+                          "version and your template contains an unrecognized type-cast, probably a "
+                          "parenthesis in the value field. This will be interpreted in the `type(value)' "
+                          "format in future versions. Please add an "
+                          "explicit string() cast as shown in the 'fixed-value' tag of this log message "
+                          "or remove the parenthesis",
```
alltilla, line 114: Same here.

alltilla (commented):
If I understand correctly, this PR does not just add warnings, but falls back to the old behavior in the problematic cases. Should we mention that in the PR description and/or news entry?

modules/add-contextual-data/contextual-data-record-scanner.c
bazsi: I added this to the warning.

modules/dbparser/synthetic-message.c
bazsi: likewise

MrAnno (approved):

alltilla (approved):

modules/add-contextual-data/contextual-data-record-scanner.c
alltilla: Thanks!

Issue #4136, opened 2022-09-15
Issue using S_ISODATE in python


ryanfaircloth:
# syslog-ng
## Version of syslog-ng
3.38.1

## Platform
linux (docker)
## Steps to reproduce
using the following global
```
```
and having parsed a ts correct with msec
the TS returned in the following python does not include ms
```python
log_message['S_ISODATE'].decode("utf-8")
```
```
"timestamp": "2022-09-15T21:56:38+00:00",
```

Using a json template 
```
template t_json_kafka {
    template('$(format-json --scope rfc5424 --exclude DATE --exclude LEGACY_MSGHDR --key RAWMSG --key TAGS --key MSGID --key SEQNUM --key S_ISODATE --key R_ISODATE --key S_UNIXTIME --key R_UNIXTIME --key .metadata.* --key .SDATA.* --key fields.*)');
};

```

Output
```json
{
	"fields": {
		"message": "Notification: 3/17/2022 18:35:12 system-login-change severity-level:minor host-name:\"XYZ_vManage_East\" system-ip:1.1.1.3 user-name:\"mn2c\" user-id:2227"
	},
	"TAGS": [
		"wireformat:rfc",
		"wireformat:rfc3164",
		"source_identified",
		"msgtype:unstructured",
		"msg:unparsed",
		".app.app-cisco-cisco_viptela",
		".app.core-proto-3164-cisco_syslog1",
		".source.s_net_default"
	],
	"S_UNIXTIME": "1663432932.000000",
	"S_ISODATE": "2022-09-17T16:42:12.000000+00:00",
	"R_UNIXTIME": "1663432932.749567",
	"R_ISODATE": "2022-09-17T16:42:12.749567+00:00",
	"RAWMSG": "<73> Sep 17 16:42:12 SYSMGR[919]: %Viptela-desktop-53-sysmgrd-6-INFO-1400002: Notification: 3/17/2022 18:35:12 system-login-change severity-level:minor host-name:\"XYZ_vManage_East\" system-ip:1.1.1.3 user-name:\"mn2c\" user-id:2227",
	"PROGRAM": "SYSMGR",
	"PRIORITY": "alert",
	"PID": "919",
[...]

PR #4161, opened 2022-10-03
Python: make config accessible globally

bazsi:
While working on the resolution of #4136 I recognized that Python code is associated to a specific configuration. Whenever we
reload the config, the Python "instance" is reloaded as well.

Sometimes the Python code intends to do something that is related to the current config, make it available in the main module
as _syslogng_main.__config__

This is actually a PyCapsule instance holding a reference to GlobalConfig.

The current change simplifies the initialization of PyLogMessage->cast_to_strings, but can similarly be used to pass the GlobalConfig instance to log_template_new(), so it can actually resolve template functions.

### DIFF
--- a/modules/python/python-logtemplate-options.c
+++ b/modules/python/python-logtemplate-options.c
@@ -46,6 +46,24 @@ py_log_template_options_new(LogTemplateOptions *template_options)
   return (PyObject *) self;
 }
 
+PyObject *
+py_log_template_options_pynew(PyTypeObject *type, PyObject *args, PyObject *kwds)
+{
+  PyLogTemplateOptions *self = PyObject_New(PyLogTemplateOptions, &py_log_template_options_type);
+
+  if (!self)
+    return NULL;
+
+  if (!PyArg_ParseTuple(args, ""))
+    return NULL;
+
+  GlobalConfig *cfg = python_get_associated_config();
+  memset(&self->template_options, 0, sizeof(self->template_options));
+  log_template_options_defaults(&self->template_options);
+  log_template_options_init(&self->template_options, cfg);
+
+  return (PyObject *) self;
+}
 
 PyTypeObject py_log_template_options_type =
 {
@@ -55,6 +73,7 @@ PyTypeObject py_log_template_options_type =
   .tp_dealloc = (destructor) PyObject_Del,
   .tp_flags = Py_TPFLAGS_DEFAULT,
   .tp_doc = "LogTemplateOptions class encapsulating a syslog-ng LogTemplateOptions",
+  .tp_new = py_log_template_options_pynew,
   0,
 };
 
@@ -62,4 +81,5 @@ void
 py_log_template_options_global_init(void)
 {
   PyType_Ready(&py_log_template_options_type);
+  PyModule_AddObject(PyImport_AddModule("_syslogng"), "LogTemplateOptions", (PyObject *) &py_log_template_options_type);
 }
--- a/modules/python/python-logtemplate-options.c
+++ b/modules/python/python-logtemplate-options.c
@@ -31,14 +31,13 @@ py_is_log_template_options(PyObject *obj)
 }
 
 PyObject *
-py_log_template_options_new(LogTemplateOptions *template_options)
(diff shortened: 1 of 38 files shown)


kira-syslogng:
Build FAILURE

bazsi:
Hi,

There are already cases where we'd need the GlobalConfig instance.

* The LogMessage wrapper to determine the version number for compat purposes
* LogTemplate to be able to resolve template functions (currently it
crashes)
* PersistState

And resolving macros on logmessages which would need at least
cfg->template_options to use frac-digits () for instance.

alltilla:
Very cool!

I can see that you are depending in my python PR with multiple PRs now, I am sorry to keep you waiting. Hopefully I can finish it this week and these can move forward.

kira-syslogng:
Build FAILURE

modules/python/python-main.c
```diff
@@ -84,6 +84,9 @@ _py_construct_main_module(void)
       Py_XDECREF(builtins_module);
     }
 
+  /* there's a circular reference between GlobalConfig -> PythonConfig -> main_module -> main_module.__config__ -> GlobalConfig */
+  PyDict_SetItemString(module_dict, "__config__", PyCapsule_New(pc->cfg, "_syslogng_main.__config__", NULL));
```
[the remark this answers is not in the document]
bazsi: There are already cases where we'd need the GlobalConfig instance.

* The LogMessage wrapper to determine the version number for compat purposes
* LogTemplate to be able to resolve template functions (currently it crashes)
* PersistState

And resolving macros on logmessages which would need at least cfg->template_options to use frac-digits () for instance.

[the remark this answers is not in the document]
MrAnno: I guess we don't have plans to publish the GlobalConfig's fields then, just the instance for C-Python binding use.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
I don't think this PR needs a NEWS entry, so it lacks one. It fixes a crash when using template functions from Python, but since has never been reported, I guess noone really  uses that case.

I would announce the whole batch when merging #4165 that contains a lot of user-visible changes.

kira-syslogng:
Build FAILURE

bazsi:
I just found the bug found by @kira-syslogng causing syslog-ng to crash when a template_options argument is missing from both
the LogTemplate constructor and the parse() function

BTW, before the black list remove patches (merged just recently), kira was trying to resolve the STAMP macro which was blacklisted. now it gets resolved properly.

modules/python/python-logtemplate.c
```diff
@@ -24,34 +24,28 @@
...
 #include "python-types.h"
+#include "python-main.h"
 #include "scratch-buffers.h"
 #include "messages.h"
 
 PyTypeObject py_log_template_type;
 PyObject *PyExc_LogTemplate;
 
-void
-py_log_template_free(PyLogTemplate *self)
```
MrAnno, line 34: This patch contains multiple refactor rounds as well, making it a bit harder to review.

modules/python/python-logtemplate-options.c
```diff
@@ -46,6 +46,24 @@ py_log_template_options_new(LogTemplateOptions *template_options)
   return (PyObject *) self;
 }
 
+PyObject *
+py_log_template_options_pynew(PyTypeObject *type, PyObject *args, PyObject *kwds)
+{
+  PyLogTemplateOptions *self = PyObject_New(PyLogTemplateOptions, &py_log_template_options_type);
```
MrAnno: We should call tp_alloc here:

> The [tp_new](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_new) function should call subtype->tp_alloc(subtype, nitems) to allocate space for the object, and then do only as much further initialization as is absolutely necessary.

MrAnno (approved):

modules/python/python-logtemplate-options.c
MrAnno: Removed in a commit later.

PR #4162, opened 2022-10-04
Python: add support for datetime() type

bazsi:
This PR adds support for handling name-value pairs that have datetime() as type, such name-value pairs would be propagated
to Python as a datetime.datetime instance. 

When setting a name-value pair, if the value is a datetime.datetime instance, it would similarly be converted to a value with syslog-ng's datetime as type.

Example config:

```
@version: 4.0

options {
        ts-format(iso);
};

python {

def foo(msg):
        return msg["dtvalue"].isoformat()

};


log {
        source { tcp(port(2000)); };
        rewrite { set(datetime("1664885832.123") value("dtvalue")); };
        destination { file("foobar" template("$(python foo)\n")); };
};
```

This is how you set the timestamp:
```
        msg.set_timestamp(datetime.datetime(1977, 8, 14, 5, 30))
```

This is how you set a name-value pair with datetime type:

```
        msg['dtvalue'] = datetime.datetime(1977, 8, 14, 5, 30)

```

At the moment timezones other than the local time are not supported for name-value pairs, but that's an easy limitation to lift.

### DIFF
--- a/modules/python/python-value-pairs.c
+++ b/modules/python/python-value-pairs.c
@@ -42,9 +42,8 @@ python_worker_vp_add_one(const gchar *name,
     {
       gchar buf[256];
 
-      _py_format_exception_text(buf, sizeof(buf));
       msg_error("python-value-pairs: error converting a name-value pair to a Python object",
-                evt_tag_str("exception", buf));
+                evt_tag_str("exception", _py_format_exception_text(buf, sizeof(buf))));
       _py_finish_exception_handling();
 
       return type_cast_drop_helper(template_options->on_error, value, log_msg_value_type_to_str(type));
--- a/modules/python/python-value-pairs.c
+++ b/modules/python/python-value-pairs.c
@@ -36,16 +36,17 @@ python_worker_vp_add_one(const gchar *name,
 {
   const LogTemplateOptions *template_options = (const LogTemplateOptions *)((gpointer *)user_data)[0];
   PyObject *dict = (PyObject *)((gpointer *)user_data)[1];
-  gboolean fallback = template_options->on_error & ON_ERROR_FALLBACK_TO_STRING;
 
   PyObject *obj = py_obj_from_log_msg_value(value, value_len, type);
-  if (!obj && fallback)
-    {
-      obj = py_obj_from_log_msg_value(value, value_len, LM_VT_STRING);
-    }
-
   if (!obj)
     {
+      gchar buf[256];
+
+      _py_format_exception_text(buf, sizeof(buf));
+      msg_error("python-value-pairs: error converting a name-value pair to a Python object",
+                evt_tag_str("exception", buf));
+      _py_finish_exception_handling();
+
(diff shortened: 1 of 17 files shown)


bazsi:
@ryanfaircloth this contains a couple of bits that could resolve your Python time related woes (e.g. the one where you are trying to resolve S_ISODATE) I am also on that specific issue.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
although this could have a separate NEWS entry, the python typing work didn't have one and I think we should have a single NEWS entry about typing support in Python.

kira-syslogng:
Build FAILURE

modules/python/python-value-pairs.c
```diff
@@ -36,16 +36,16 @@ python_worker_vp_add_one(const gchar *name,
 {
   const LogTemplateOptions *template_options = (const LogTemplateOptions *)((gpointer *)user_data)[0];
   PyObject *dict = (PyObject *)((gpointer *)user_data)[1];
-  gboolean fallback = template_options->on_error & ON_ERROR_FALLBACK_TO_STRING;
 
   PyObject *obj = py_obj_from_log_msg_value(value, value_len, type);
-  if (!obj && fallback)
-    {
-      obj = py_obj_from_log_msg_value(value, value_len, LM_VT_STRING);
-    }
```
MrAnno, line 45: I've found this in an error message improvement commit.
It is probably here because of the `type_cast_error` fallback.

MrAnno (commented):
I have 2 minor comments related to commit history.

I've also started an internal test run on this PR.
LGTM otherwise.

MrAnno:
(Since our review capacity is the current bottleneck, I think we need to optimize for it, which is why I think the above notes should be fixed.)

MrAnno:
Internal tests have passed.

modules/python/python-value-pairs.c
bazsi: in the meantime that _log_py_object_conversion_error_with_invalid_type() was removed from master and I didn't fix it correctly during rebase.

I am running a compile test on each patch in the series, hopefully that resolves it.

bazsi:
Every patch compiles on its own with the latest iteration. I still have one unit test that fails in the middle and then gets fixed, namely `test_python_persist_name`. This happens because embedding the _py_format_exception_text() into the argument list of the
logging function means that ordering is changed between these two functions:

const gchar *_py_get_callable_name(PyObject *callable, gchar *buf, gsize buf_len);
const gchar *_py_format_exception_text(gchar *buf, gsize buf_len);

and _py_get_callable_name() did not save/restore the exception state originally (was a bug even then, it just wasn't apparent up to now). The last patch fixes this issue, so the end of the branch executes tests correctly and is not ordering dependant anymore.

MrAnno (approved):
Thank you.

PR #4165, opened 2022-10-06
Python support for installing python modules

bazsi:
This pull request builds on all pending Python PRs already opened and allows a Python based module to become part of the syslog-ng source tree.

Unfortunately this has become a "monster PR", let me break down the changes here:
  * LogDestination and LogParser base classes were introduced and now the C glue code validates that the Python counterparts are actually derived from these classes (previously we just expected the signatures of various functions to be right). This is subject to compatibility, so the validation only happens in `@version: 4.0`, anything earlier will work as before
  * instead of "hiding' the internals of the various Python base classes (LogDestination, LogSource, LogFetcher, LogParser), I published those in the syslogng module along with extensive docstring based documentation. The syslogng module also publishes fakes of most classes in order to be able to run unit tests outside of syslog-ng itself.
  * publishing various enums from C to Python were unified, LogFetcher and LogDestination uses enums to indicate status, these are now published as class attributes, but their values come from an enum exported by the C glue code
  * the Python side can now use the official "logging" module to emit log messages, they will be submitted to the internal() driver for processing, just like any other log messages
  * there was no official directory to put Python modules up to this point. Now we are adding /etc/syslog-ng/python to the PYTHONPATH automatically, so if someone has local Python code, it is possible to put it somewhere
  * the python-modules directory was created to host Python based syslog-ng elements. This directory would host both the low-level interface code (e.g. the syslogng module) and higher-level code (e.g. any source/destination/parser/...) elements
  * bugfix: fix a crash where a Persist instance is created before the configuration was initialized. Instead of a crash we are now raising an exception (RuntimeError)
  * bugfix: there was a potential deadlock in the ack mechanism

In a subsequent PR I also added virtualenv support, which is kind of infrastructure, but this PR was already too large those bits are in #4173


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

modules/python/pylib/Makefile.am
```diff
@@ -45,27 +45,27 @@ if ENABLE_PYTHON
...
-SETUPPY_MANIFEST = $(PYLIB_BUILDDIR)/install-manifest.txt
+PYLIB_MANIFEST = $(PYLIB_BUILDDIR)/install-manifest.txt
 PYTHON_ROOT = $(if $(DESTDIR),$(DESTDIR),/)
 
 INSTALL_EXEC_HOOKS += install-pylib
 UNINSTALL_HOOKS += uninstall-pylib
 CLEAN_HOOKS += clean-pylib
 
-PYSETUP_OPTIONS ?= --root="$(PYTHON_ROOT)" --install-lib="$(python_moduledir)"
+PYLIB_SETUP_OPTIONS ?= --root="$(PYTHON_ROOT)" --install-lib="$(python_moduledir)"
```
MrAnno, line 55: In case we want to install 3rd party dependencies too into our own python folder, the following options may come in handy to workaround a few pip bugs around custom install dirs (they sometimes mix up system dirs with custom dirs when checking which deps to install):

```
--ignore-installed --cache-dir "$(PYLIB_BUILDDIR)/cache" --build "$(PYLIB_BUILDDIR)/build"
```

modules/python/pylib/setup.py
```diff
@@ -21,7 +21,7 @@
 #
 #############################################################################
 
-from distutils.core import setup
+from setuptools import setup
```
MrAnno, line 24: Note: We could list dependencies here under `install_requires=[]`.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

modules/python/pylib/Makefile.am
bazsi: hmm.. this variable is now unused as it seems. when I migrated to using pip to install packages. I am not sure when you did this review. 

at the moment I have --upgrade so it overwrites the existing installation (which --ignore-installed is probably achieving)

customizing build/cache directories might make sense, let me add those... 

Hmm. as it seems neither build or cache are supported by pip, so I guess those options only apply to setup.py

I am not sure they are needed now, though. I was reading packaging manuals for hours to come up with the pip based solution as it seems to be the way forward now. Using setup.py to install packages directly is not supported anymore.

modules/python/pylib/setup.py
bazsi: I know and the current solution does that. It's somewhat ugly though (as pip insists on installing my package and its dependencies in the same directory), but look at the current solution and let me know how you like it. Let's discuss if a discussion is needed.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

modules/python/pylib/Makefile.am
MrAnno: `--ignore-installed` was necessary for our internal modules to avoid a few known pip install prefix bugs.
I'll find and reference those issues soon.

kira-syslogng:
Build FAILURE

modules/python/pylib/Makefile.am
bazsi: nevermind, I am still fighting with setuptools/pip and related problems. I previously wanted to migrate to using pip (due to setuptools claiming our use-case was deprecated) but I am now thinking about adopting poetry instead. That seems to be much better. 

But instead of  rolling that into this PR (already too big), I now try to stick to setup.py only, as it worked for us until now, I am not trying to find the replacement features in pip (if they exist at all).

I am also thinking about merging modules/python/pylib and modules/python-modules to install a single "syslogng" package (both the code exporting core features, and debuggercli and any official python modules).

But let me first try to make the current state work, with setup.py and then see if the merger of these modules would help.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
@kira-syslogng rebuild this please;

bazsi:
@kira-syslogng restest this please;

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

MrAnno:
Can we split this branch into smaller PRs for an easier review?

I see 3 main topics here:
- general Python binding improvements
- 2 bugfixes
- support for writing Python modules

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

modules/python/pylib/Makefile.am
bazsi: I think with the virtualenv management stuff, this is resolved.

MrAnno (changes requested):
I'm requesting a change so that my comments won't disappear ("collapse") due to the large number of git pushes.

----

Can we split this branch into smaller PRs for an easier review?

I see 4 main topics here:
- general Python binding improvements
- 2 bugfixes
- support for writing Python modules
- virtualenv support seems to be in this PR too

#4185 has been self-merged without review, which implies that the usage of virtualenv can no longer be questioned during review.

alltilla:
I will review this PR this week. Sorry for the delay!

bazsi:
sorry for iterating so much on this branch, I was not aware that you already started reviewing it and that I was making that more difficult for you.

> I'm requesting a change so that my comments won't disappear ("collapse") due to the large number of git pushes.
> 
> Can we split this branch into smaller PRs for an easier review?

I am doing that now. This will, again, generate rebases possibly on this branch too. But I am opening smaller batches.

> 
> I see 4 main topics here:
> 
>     * general Python binding improvements
> 
>     * 2 bugfixes
> 
>     * support for writing Python modules
> 
>     * virtualenv support seems to be in this PR too
> 
> 
> #4185 has been self-merged without review, which implies that the usage of virtualenv can no longer be questioned during review.

That branch was only adding the python3-venv package to the dbld images on debian platform. It has not changed anything code-wise. If we come up with a solution that is not virtualenv based, we can certainly drop that package.

I merged it as otherwise I needed to add "DROPTHIS" patches on top of my branch and I eagerly wanted to see a green state in the CI which I've fought a lot for.

Anyway, will open those PRs this morning.

MrAnno:
> sorry for iterating so much on this branch, I was not aware that you already started reviewing it and that I was making that more difficult for you.

It's not a problem at all. :) I mentioned frequent pushes only to explain why I clicked on "change request" and why I repeated an earlier comment that spun out of sight. We can use the "draft mode" when you don't prefer premature review notes. (Kira is currently an internal E2E test tool running on pull requests only, but its functionality is very useful for validation purposes as well :))

> I am doing that now. This will, again, generate rebases possibly on this branch too. But I am opening smaller batches.

Thank you. I will review them soon.

> That branch was only adding the python3-venv package to the dbld images on the debian platform. It has not changed anything code-wise. If we come up with a solution that is not virtualenv based, we can certainly drop that package.

I completely understand. My only concern was the way the team looked at the master branch: it has to be always stable, but on top of that, it only contained validated "final" decisions. I know there is no such thing as "final", but we considered it a big mistake if it turned out the work had to be reverted for some reason.
This is obviously not the case here, we are going to add virtualenv support. :) I just wanted to share my thoughts.

bazsi:
All of this PR was broken down to smaller PRs, now opened on github. I'd close this one, unless there's any review notes that'd be important to keep. let me know.

MrAnno:
Thank you. I think we can close this one.

PR #4173, opened 2022-10-15
Kubernetes metadata enrichment

bazsi:
This work was originally done by @alltilla to implement a parser that associates kubernetes metadata with log messages. I turned this into a fully fledged Python module as implemented by #4165

### DIFF
--- a/modules/python-modules/syslogng/modules/kubernetes/__init__.py
+++ b/modules/python-modules/syslogng/modules/kubernetes/__init__.py
@@ -61,6 +61,9 @@ class KubernetesAPIEnrichment(LogParser):
         for label_name, label_value in pod.metadata.labels.items():
             pod_metadata[self.add_prefix("labels.{}".format(label_name))] = label_value
 
+        for annotation_name, annotation_value in pod.metadata.annotations.items():
+            pod_metadata[self.add_prefix("annotations.{}".format(annotation_name))] = annotation_value
+
         pod_metadata[self.add_prefix("namespace_name")] = namespace_name
         pod_metadata[self.add_prefix("pod_name")] = pod_name
 
--- a/modules/python-modules/syslogng/modules/kubernetes/__init__.py
+++ b/modules/python-modules/syslogng/modules/kubernetes/__init__.py
@@ -1,0 +1,101 @@
+#############################################################################
+# Copyright (c) 2022 Attila Szakacs <szakacs.attila96@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 as published
+# by the Free Software Foundation, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+from syslogng import Logger, LogParser
+import importlib
+
+
+class KubernetesAPIEnrichment(LogParser):
+    logger = Logger()
+
+    def init(self, options):
+        try:
+            config = importlib.import_module("kubernetes.config")
+            client = importlib.import_module("kubernetes.client")
+        except ModuleNotFoundError:
+            logger.warning("Unable to import kubernetes libraries for Python, no data enrichment is performed")
+            self.has_kubernetes_module = False
+            return True
+
+        self.has_kubernetes_module = True
+
+        self.__metadata = {}
+        self.__prefix = options["prefix"]
+
+
+        if options["in_pod"] == "yes":
+            config.load_incluster_config()
+        else:
+            config.load_kube_config()
+
+        self.__client_api = client.CoreV1Api()
+
+        return True
+
+    def add_prefix(self, string):
+        return "{}{}".format(self.__prefix, string)
+
+    def __get_cached_pod_metadata(self, namespace_name, pod_name):
+        self.logger.debug("Trying to find cached metadata for pod {}".format(pod_name))
+        return self.__metadata[namespace_name][pod_name]
+
+    def __query_pod_metadata_from_kubernetes_api_server(self, namespace_name, pod_name):
+        self.logger.debug("Trying to query metadata for pod {} from the Kubernetes API".format(pod_name))
+        pod = self.__client_api.list_namespaced_pod(namespace_name, field_selector="metadata.name=={}".format(pod_name)).items[0]
+        container_status = pod.status.container_statuses[0]
+
+        namespace = self.__metadata.setdefault(namespace_name, {})
+        pod_metadata = namespace.setdefault(pod_name, {})
+
+        pod_metadata[self.add_prefix("pod_uuid")] = pod.metadata.uid
+
+        for label_name, label_value in pod.metadata.labels.items():
+            pod_metadata[self.add_prefix("labels.{}".format(label_name))] = label_value
+
+        pod_metadata[self.add_prefix("namespace_name")] = namespace_name
+        pod_metadata[self.add_prefix("pod_name")] = pod_name
+
+        pod_metadata[self.add_prefix("container_name")] = container_status.name
+        pod_metadata[self.add_prefix("container_image")] = container_status.image
+        pod_metadata[self.add_prefix("container_hash")] = container_status.image_id.replace("docker-pullable://", "", 1)
+        pod_metadata[self.add_prefix("docker_id")] = container_status.container_id.replace("docker://", "", 1)
+
+        return pod_metadata
+
+    def get_pod_metadata(self, namespace_name, pod_name):
+        try:
+            return self.__get_cached_pod_metadata(namespace_name, pod_name)
+        except KeyError:
+            return self.__query_pod_metadata_from_kubernetes_api_server(namespace_name, pod_name)
+
+    def parse(self, msg):
+        if not self.has_kubernetes_module:
+            return True
+
+        namespace_name = msg[self.add_prefix("namespace_name")].decode()
+        pod_name = msg[self.add_prefix("pod_name")].decode()
+
+        pod_metadata = self.get_pod_metadata(namespace_name, pod_name)
+        for name, value in pod_metadata.items():
+            if msg[name] == b"":
+                msg[name] = value
+
+        return True
--- a/modules/python-modules/syslogng/modules/kubernetes/__init__.py
+++ b/modules/python-modules/syslogng/modules/kubernetes/__init__.py
@@ -45,29 +45,34 @@ class KubernetesAPIEnrichment(LogParser):
         return self.__metadata[namespace_name][pod_name]
 
     def __query_pod_metadata_from_kubernetes_api_server(self, namespace_name, pod_name):
-        self.logger.debug("Trying to query metadata for pod {} from the Kubernetes API".format(pod_name))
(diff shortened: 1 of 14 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi:
Style check fails because of #4215

kira-syslogng:
Build FAILURE

MrAnno:

modules/python-modules/syslogng/modules/kubernetes/scl/kubernetes.conf
```diff
@@ -78,6 +78,12 @@ block parser kubernetes-metadata-parser (prefix() template("$FILE_NAME")) {
                 template(`template`)
                 prefix(`prefix`)
             );
+            python(
+                class("syslogng.modules.kubernetes.KubernetesAPIEnrichment")
+                options(
+                    "prefix" => "`prefix`"
```
MrAnno, line 84: We should publish the `in_pod` option too, I needed that in my deployment.

modules/python-modules/syslogng/modules/kubernetes/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+        namespace = self.__metadata.setdefault(namespace_name, {})
+        pod_metadata = namespace.setdefault(pod_name, {})
+
+        pod_metadata[self.add_prefix("pod_uuid")] = pod.metadata.uid
+
+        for label_name, label_value in pod.metadata.labels.items():
+            pod_metadata[self.add_prefix("labels.{}".format(label_name))] = label_value
+
+        for annotation_name, annotation_value in pod.metadata.annotations.items():
+            pod_metadata[self.add_prefix("annotations.{}".format(annotation_name))] = annotation_value
```
MrAnno, line 65: [request-change-reason]: We should handle metadata optionality:

```
File "/home/anno/dev/syslog-ng/install/lib/syslog-ng/python/syslogng/modules/kubernetes/__init__.py", line 64, in __query_pod_metadata_from_kubernetes_api_server
    for annotation_name, annotation_value in pod.metadata.annotations.items():
AttributeError: 'NoneType' object has no attribute 'items'
```

modules/python-modules/syslogng/modules/kubernetes/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+        self.logger.debug("Trying to query metadata for pod {} from the Kubernetes API".format(pod_name))
+        pod = self.__client_api.list_namespaced_pod(namespace_name, field_selector="metadata.name=={}".format(pod_name)).items[0]
+        container_status = pod.status.container_statuses[0]
+
+        namespace = self.__metadata.setdefault(namespace_name, {})
+        pod_metadata = namespace.setdefault(pod_name, {})
+
+        pod_metadata[self.add_prefix("pod_uuid")] = pod.metadata.uid
+
+        for label_name, label_value in pod.metadata.labels.items():
```
MrAnno, line 61: This might be optional (see my other comment below).

modules/python-modules/syslogng/modules/kubernetes/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+            pod_metadata[self.add_prefix("labels.{}".format(label_name))] = label_value
+
+        for annotation_name, annotation_value in pod.metadata.annotations.items():
+            pod_metadata[self.add_prefix("annotations.{}".format(annotation_name))] = annotation_value
+
+        pod_metadata[self.add_prefix("namespace_name")] = namespace_name
+        pod_metadata[self.add_prefix("pod_name")] = pod_name
```
MrAnno, line 68: `namespace_name`, `pod_name`, and `container_name` are set by the SCL as well. Should we remove it?

MrAnno (changes requested):
I've tested this PR on my K3s deployment. It has started working after some minor modifications.

I marked my important review notes with `[request-change-reason]`.

----

> ./configure --enable-python-maintainer-mode to enable a couple of Makefile targets that only need to be used in case someone is actively working on Python-implemented modules.

I could not find this change in any of the open PRs.

> added support for Python virtualenvs to host Python dependencies for our Python implemented plugins

I think this belongs to #4197.

> drop Python2 from Debian packaging

That was in #4196.

modules/python-modules/syslogng/modules/kubernetes/__init__.py
MrAnno: I think we could be more careful with all the other fields, as an exception could lead to message loss.

bazsi:
> 



> I've tested this PR on my K3s deployment. It has started working after some minor modifications.
> 
> I marked my important review notes with `[request-change-reason]`.

cool, thanks for the review. will respond to those one-by-one.

> 
> > ./configure --enable-python-maintainer-mode to enable a couple of Makefile targets that only need to be used in case someone is actively working on Python-implemented modules.
> 
> I could not find this change in any of the open PRs.

This does not exist anymore, I killed it in favour of the build-time and production-time virtualenvs and their automatic management by syslog-ng. So no wonder you couldn't find it. The PR description is outdated, but I've removed it completely from the history.


> 
> > added support for Python virtualenvs to host Python dependencies for our Python implemented plugins
> 
> I think this belongs to #4197.

And that was merged as a part  of that.

> 
> > drop Python2 from Debian packaging
> 
> That was in #4196.

yup. sorry for not updating the description here.

packaging/debian/syslog-ng-mod-python.install
```diff
@@ -1,6 +1,7 @@
 usr/lib/syslog-ng/*/libmod-python.so
 usr/lib/syslog-ng/*/python/requirements.txt
 usr/lib/syslog-ng/*/python/syslogng/*.py
-usr/lib/syslog-ng/*/python/syslogng/debuggercli/*.py
+usr/lib/syslog-ng/*/python/syslogng/debuggercli/*
+usr/lib/syslog-ng/*/python/syslogng/modules/kubernetes/*
```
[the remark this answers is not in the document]
bazsi: True, however this is more of a distribution concern (which have their own packaging files anyway), as we manage Python dependencies ourselves simply by bootstrapping a virtualenv. And once PyPI is available for that, we can easily pull in any Python library that any of our modules would possibly need.

This is worth of a discussion, I decided to use the same package to host all Python based modules, as any other way, we would need to:
  1) either create a virtualenv management framework which installs mod-XXX specific python packages as the module is installed
  2) try to use Debian/Ubuntu packages for the Python modules, which would limit us to Python packages/versions available in Debian/Ubuntu.

These are roughly the reasons I opted against using separate packages, but we can discuss.

modules/python-modules/syslogng/modules/kubernetes/scl/kubernetes.conf
```diff
@@ -68,82 +68,6 @@
 # filename format under /var/log/pods
 @define kubernetes-log-pods-regexp '.var.log.pods.(?<namespace_name>[^_]+)_(?<pod_name>[^_]+)_(?<pod_uuid>[a-z0-9-]*).(?<container_name>.+)/[^.]+.log'
 
-python {
```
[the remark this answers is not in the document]
bazsi: scl.conf not being overwritten is an explicit functionality that was done back when SCL was introduced. scl.conf has been assumed to be user-modified file.

In hindsight, this was probably not a good idea. This is in this commit: 287993339599deac0442e26355c600b5aee63583

The only rationale why this was made is found in scl.conf:

# This file is placed into /etc/syslog-ng in order to make it trivial to
# include in user written syslog-ng.conf files.  It sets up 'scl-root' and
# `include-path`, then includes all SCL supplied plugins.

I don't think fixing either of the above problems (kubernetes.conf being removed or scl.conf not being overwritten) should be part of this PR. There are a lot of installed files that behave this way, for instance our versioned libsyslog-ng.so or renamed modules that leave the previous versions of those files in the install directory.

The proper way to handle this situation if you are compiling from source is to "make uninstall" in the old version first and then "make install". That should do the right thing.

With any kind of packaging, neither is an issue.

I would probably move scl.conf to /usr/share/syslog-ng/include/scl and drop it from being installed in /etc/syslog-ng. We could do this with 4.0 and I can prepare a PR for that.

With that said, what could be done to improve this situation is to move the `@include` to a separate SCL include file that is included by the default scl.conf as well. I am doing that in a followup.

modules/python-modules/syslogng/modules/kubernetes/scl/kubernetes.conf
bazsi: good point. I am thinking about we can actually detect this value, instead of having to make this explicit. That would be a lot better.

I am thinking about trying to load the config file and falling back to the incluster config if the first one raises an exception.

@alltilla what do you think? I am trying in a moment if this would work.

bazsi: I am pushing another iteration where this is implemented, it seems there's even this function in the kubernetes python libs.

modules/python-modules/syslogng/modules/kubernetes/__init__.py
bazsi: I have added error handling with the latest iteration.

modules/python-modules/syslogng/modules/kubernetes/__init__.py
bazsi: likewise

modules/python-modules/syslogng/modules/kubernetes/__init__.py
bazsi: done.

modules/python-modules/syslogng/modules/kubernetes/scl/kubernetes.conf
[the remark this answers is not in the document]
bazsi: done.

bazsi:
I tried to attempt all review comments with the latest push. Thanks for the review.

modules/python-modules/syslogng/modules/kubernetes/scl/kubernetes.conf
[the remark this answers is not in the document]
MrAnno: Thank you.

packaging/debian/syslog-ng-mod-python.install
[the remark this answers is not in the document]
MrAnno: I understand. Thanks for the explanation and the reminder.

MrAnno:
> yup. sorry for not updating the description here.

I removed the irrelevant parts from the description.


Thanks for addressing the notes. I'll check the changes and retest the PR tomorrow.

scl/python/python-modules.conf
```diff
@@ -0,0 +1,24 @@
...
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+@define python-module-dir "`module-install-dir`/python/syslogng/modules"
+
+@include '`python-module-dir`/*/scl/*.conf'
```
MrAnno, line 24: An automake+cmake change is needed to install this.

MrAnno:
LGTM otherwise.

scl/python/python-modules.conf
bazsi: sorry, I missed that. Added now.

MrAnno (approved):
👍

PR #4174, opened 2022-10-15
Light: Remove Python 2 compatibility

mitzkia:
The main goal of this PR:
- From the early life of Light there was an endeavor to support Python 2 along with Python 3 (which I intended as an additional value)
- Python 2 has been EOL-ed
- In this PR I would like to drop the Python 2 support from Light to focusing on pure Python 3 usage

Additional tasks with this PR:
- [x] Fixup git commit descriptions
- [x] Light related make targets (pytest-check, pytest-self-check) should Pass
- [x] Check a full GitHub Actions pipeline with a new devshell image (related pip package was deleted)

### DIFF
--- a/dbld/pip_packages.manifest
+++ b/dbld/pip_packages.manifest
@@ -8,7 +8,6 @@ hyrule                  [devshell, kira]                                       [
 requests                [kira]                                                 [python2, python3]
 
 # pip packages for python functional tests
-pathlib2                [centos, fedora, debian, ubuntu, devshell]             [python3]
 psutil                  [centos, fedora, debian, ubuntu, devshell]             [python3]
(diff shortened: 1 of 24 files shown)


kira-syslogng:
Build FAILURE

mitzkia:

mitzkia:
Maybe I am wrong, but as I can see new devshell image could built with the changes:
https://github.com/syslog-ng/syslog-ng/actions/runs/3258534477/jobs/5350698749
Could someone else confirm this? 

If this is true than I would say:
- the new (devshell) image could built with the package changes
- Light could run with the changes in a pure Python 3 environment

So the changes could work.

bazsi (approved):
I reviewed the changes in detail, approved.

It seems that we only need Python2 for the kira image now.

MrAnno (approved):

PR #4181, opened 2022-10-31
Light: Closing open fds on testcase teardown

### LINKED ISSUE #4178: Light: leaking open files and open sockets after each testcase
Adding the following lines to the beginning of: `tests/light/functional_tests/conftest.py:setup`

```python
 @pytest.fixture(autouse=True)
 def setup(request):
+
+    import psutil
+    print("Open fds on testcase start: %s\n" % psutil.Process().open_files())
+    print("Open sockets on testcase start: %s\n" % psutil.Process().connections())
+
     testcase_parameters = request.getfixturevalue("testcase_parameters")
 
     copy_file(testcase_parameters.get_testcase_file(), Path.cwd())
```

and starting Light with the following command:
```shell
$ python3 -m pytest -lvvvvs --installdir=/install functional_tests/destination_drivers/network_destination/test_network_destination_transport.py -o log_cli=true
```

We can notice that the number of open fds and sockets are increasing after each testcase:
```
Open fds on testcase start: [popenfile(path='/source/tests/light/reports/2022-10-27-04-20-07-517434/test_network_destination_transport_default_/syslog_ng_server_stdout', fd=9, position=0, mode='a', flags=558081), popenfile(path='/source/tests/light/reports/2022-10-27-04-20-07-517434/test_network_destination_transport_default_/syslog_ng_server_stderr', fd=10, position=186810, mode='a', flags=558081), popenfile(path='/source/tests/light/reports/2022-10-27-04-20-07-517434/test_network_destination_transport_default_/syslog_ng_server_stderr', fd=11, position=186810, mode='r', flags=557056), popenfile(path='/source/tests/light/reports/2022-10-27-04-20-07-517434/test_network_destination_transport_tcp_/syslog_ng_server_stdout', fd=14, position=0, mode='a', flags=558081), popenfile(path='/source/tests/light/reports/2022-10-27-04-20-07-517434/test_network_destination_transport_tcp_/syslog_ng_server_stderr', fd=15, position=186815, mode='a', flags=558081), popenfile(path='/source/tests/light/reports/2022-10-27-04-20-07-517434/test_network_destination_transport_tcp_/syslog_ng_server_stderr', fd=16, position=186815, mode='r', flags=557056), popenfile(path='/source/tests/light/reports/2022-10-27-04-20-07-517434/test_network_destination_transport_udp_/testcase.reportlog', fd=17, position=0, mode='w', flags=557057)]

Open sockets on testcase start: [pconn(fd=13, family=<AddressFamily.AF_INET: 2>, type=<SocketKind.SOCK_STREAM: 1>, laddr=addr(ip='127.0.0.1', port=30001), raddr=addr(ip='127.0.0.1', port=39837), status='CLOSE_WAIT'), pconn(fd=5, family=<AddressFamily.AF_INET: 2>, type=<SocketKind.SOCK_STREAM: 1>, laddr=addr(ip='127.0.0.1', port=30002), raddr=(), status='LISTEN'), pconn(fd=8, family=<AddressFamily.AF_INET: 2>, type=<SocketKind.SOCK_STREAM: 1>, laddr=addr(ip='127.0.0.1', port=30001), raddr=(), status='LISTEN'), pconn(fd=18, family=<AddressFamily.AF_INET: 2>, type=<SocketKind.SOCK_STREAM: 1>, laddr=addr(ip='127.0.0.1', port=30002), raddr=addr(ip='127.0.0.1', port=33067), status='CLOSE_WAIT')]
```

We already have the mechanism to close this resources:
https://github.com/syslog-ng/syslog-ng/blob/e775521ac5da4d6e7a0d581bf2612478ae1ce35c/tests/light/src/common/file.py#L59
and
https://github.com/syslog-ng/syslog-ng/blob/e775521ac5da4d6e7a0d581bf2612478ae1ce35c/tests/light/src/driver_io/network/network_io.py#L48

but as you can see they are in the `atexit` "time"
The definition of `atexit`:
> The [atexit](https://docs.python.org/3/library/atexit.html#module-atexit) module defines functions to register and unregister cleanup functions. Functions thus registered are automatically executed upon normal interpreter termination.

The `atexit` was introduced in Light because of this:
```
commit 3d08a978b7e88091f30acf4958b80e547c8387e7
Author: Andras Mitzki <andras.mitzki@quest.com>
Date:   Fri May 7 11:47:34 2021 +0200

    LightFile: Removing __del__
    
    - replacing __del__ with atexit expression, due to if there is an AssertionError
    exception python will exit and __del__() will not run, this mechanism was used also
    in NetworkIO()
```

The resolution for this would be simply to move the registered functions from 'atexit' to '__del__', but in this case we will loose the original goal why we added 'atexit' to Light.
I think the bigger issue is when we leaks resources after each testcase, of course we should also handle the case of 'atexit'. I have already checked what happens when we are using both 'atexit' and '__del__' too. In this case we should add '__del__' to much more place. Why I would like to handle this now? I am working on a new PR where I will easily increase the internal testcases for drivers (of course only in small steps :) ). And during this I got an error from the kernel "too many open files". 
I am interesting what is your opinion about this "issue".
Thanks

mitzkia:
Fixes #4178

### DIFF
--- a/tests/light/conftest.py
+++ b/tests/light/conftest.py
@@ -102,8 +102,8 @@ def testcase_parameters(request):
 
 
 @pytest.fixture
-def config(request):
-    return SyslogNgConfig(request.getfixturevalue("version"))
+def config(request, teardown):
+    return SyslogNgConfig(request.getfixturevalue("version"), teardown)
 
 
 @pytest.fixture
--- a/tests/light/conftest.py
+++ b/tests/light/conftest.py
@@ -107,10 +107,10 @@ def config(request, teardown):
 
 
 @pytest.fixture
-def syslog_ng(request, testcase_parameters):
+def syslog_ng(request, testcase_parameters, teardown):
     tc_parameters.INSTANCE_PATH = SyslogNgPaths(testcase_parameters).set_syslog_ng_paths("server")
-    syslog_ng = SyslogNg(tc_parameters.INSTANCE_PATH, testcase_parameters)
-    request.addfinalizer(lambda: syslog_ng.stop())
+    syslog_ng = SyslogNg(tc_parameters.INSTANCE_PATH, testcase_parameters, teardown)
+    teardown.register(syslog_ng.stop)
     return syslog_ng
 
 
--- a/tests/light/conftest.py
+++ b/tests/light/conftest.py
@@ -114,6 +114,24 @@ def syslog_ng(request, testcase_parameters):
     return syslog_ng
 
 
+class TeardownRegistry:
+    teardown_callbacks = []
+
+    def register(self, teardown_callback):
+        TeardownRegistry.teardown_callbacks.append(teardown_callback)
+
+    def execute_teardown_callbacks(self):
+        for teardown_callback in TeardownRegistry.teardown_callbacks:
+            teardown_callback()
+
+
+@pytest.fixture(scope="function")
+def teardown():
+    teardown_registry = TeardownRegistry()
+    yield teardown_registry
+    teardown_registry.execute_teardown_callbacks()
+
+
 @pytest.fixture
 def syslog_ng_ctl(syslog_ng):
     return SyslogNgCtl(syslog_ng.instance_paths)
--- a/tests/light/src/driver_io/network/network_io.py
+++ b/tests/light/src/driver_io/network/network_io.py
@@ -48,9 +48,7 @@ class NetworkIO():
         loggen_input_file_path = Path("loggen_input_{}.txt".format(get_unique_id()))
 
         loggen_input_file = File(loggen_input_file_path)
-        loggen_input_file.open(mode="w")
-        loggen_input_file.write(content)
-        loggen_input_file.close()
+        loggen_input_file.write_content_and_close(content)
 
         Loggen().start(self.__ip, self.__port, read_file=str(loggen_input_file_path), dont_parse=True, permanent=True, rate=rate, **self.__transport.value)
 
--- a/tests/light/src/driver_io/network/network_io.py
+++ b/tests/light/src/driver_io/network/network_io.py
@@ -20,7 +20,6 @@
 # COPYING for details.
 #
 #############################################################################
-import atexit
 import socket
 from enum import Enum
 from enum import IntEnum
@@ -44,8 +43,6 @@ class NetworkIO():
         self.__ip_proto_version = NetworkIO.IPProtoVersion.V4 if ip_proto_version is None else ip_proto_version
         self.__server = None
         self.__message_reader = None
-
-        atexit.register(self.stop_listener)
 
     def write(self, content, rate=None):
         loggen_input_file_path = Path("loggen_input_{}.txt".format(get_unique_id()))
--- a/tests/light/src/syslog_ng/syslog_ng.py
+++ b/tests/light/src/syslog_ng/syslog_ng.py
@@ -28,7 +28,7 @@ class SyslogNg(object):
     def __init__(self, instance_paths, testcase_parameters, teardown):
         self.instance_paths = instance_paths
         self.teardown = teardown
-        self.__syslog_ng_cli = SyslogNgCli(instance_paths, testcase_parameters)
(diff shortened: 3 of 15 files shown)


tests/light/conftest.py
```diff
@@ -114,6 +114,24 @@ def syslog_ng(request, testcase_parameters):
     return syslog_ng
 
 
+class FdCloseCallback:
```
bazsi, line 117: I would use a bit more generic name here. This is actually a testcase teardown function, isn't it?

Is it limited to doing the closure of FDs? In my understanding this could be any cleanup function, closing an FD being a prime example.

tests/light/src/driver_io/network/network_io.py
```diff
@@ -45,8 +44,6 @@ def __init__(self, ip, port, transport, ip_proto_version=None):
         self.__server = None
         self.__message_reader = None
 
-        atexit.register(self.stop_listener)
```
bazsi, line 48: If at all possible I'd squash this commit with the following one, as you are removing one mechanism in place for another and I'd argue that this replacement was more apparent if it was in the same patch.

tests/light/src/syslog_ng/syslog_ng.py
```diff
@@ -25,7 +25,7 @@
 
 
 class SyslogNg(object):
-    def __init__(self, instance_paths, testcase_parameters):
+    def __init__(self, instance_paths, testcase_parameters, fd_close_callback):
```
bazsi, line 28: this variable is not stored anywhere.

bazsi (commented):
I like this direction in general, but I think the teardown functionality could have a more generic name, so the code would be easier to understand.

tests/light/src/syslog_ng/syslog_ng.py
mitzkia: Thank you very much for the review comment. 
Yes, sorry, it was stored in an other commit. With editing commits I will fix this too.

tests/light/src/driver_io/network/network_io.py
mitzkia: Thank you very much for the review comment. 
Agreed and I will fix it.

tests/light/conftest.py
mitzkia: Thanks for the naming suggestion. 

Yes, I would not pass `request` as a parameter because of the following:
 - it always needs some clarification about the naming: what is it, where does it came from and so on
 - and it's functionality is too broad: it could do a lot of things
 - and with the `request` for the new pytest users it is hard to understand how the teardown works for the testcases (`yield` or `request.addfinalizer`)

I think it is better to use a parameter with the name of `teardown` itself instead of `request` or what I currently uses in the PR `fd_close_callback`. Currently I have checked that the `teardown` name is not a reserved keyword so we could use it as a fixture name.

mitzkia:
Currently I am still working on this PR, fixing review comments, and using new `teardown` fixture generally. 
With my latest push, I just added the rename from `FdCloseCallback` to `TeardownRegistry` and  from `fd_close_callback` fixture to `teardown`.

tests/light/src/syslog_ng/syslog_ng.py
mitzkia: In my last push I have fixed this

tests/light/src/driver_io/network/network_io.py
mitzkia: In my last push I have fixed this

tests/light/conftest.py
mitzkia: In my last push I have added the new rename concept.

mitzkia:
I have finished the review comment fixes.

kira-syslogng:
Build FAILURE

mitzkia:

bazsi (approved):
this looks good, thank you.

bazsi:
I have approved this, please go ahead and merge this, if it's ok for you.

mitzkia:
Thank you very much.

PR #4184, opened 2022-10-31
.travis.yml: remove Travis configuration

bazsi:
Extracting from a larger Python refactor, this is not really used, so removing as dead code.


MrAnno (approved):
Travis is waiting for a replacement for its ARM build: #3847

Since it hasn't been running for months, we can consider it dead.

PR #4185, opened 2022-10-31
dbld: add python3-venv package on Debian and derivatives

bazsi:
I am adding virtualenv support in #4165 and I need python3-venv for that. This needs to go in first as I can't use the just updated
dbld image in "CI @ devshell"


bazsi:
news file is not needed.

PR #4188, opened 2022-11-02
Python log dest and parser base classes

bazsi:
This is the first chunk of #4165 to add base classes for Python based LogDestination and LogParser classes.

### DIFF
--- a/modules/python/python-helpers.c
+++ b/modules/python/python-helpers.c
@@ -376,6 +376,24 @@ _py_perform_imports(GList *imports)
   g_list_foreach(imports, _foreach_import, NULL);
 }
 
+const gchar *
+_py_object_repr(PyObject *s, gchar *buf, gsize buflen)
+{
+  const gchar *res;
+
+  PyObject *r = PyObject_Repr(s);
+  if (!r)
+    {
+      _py_finish_exception_handling();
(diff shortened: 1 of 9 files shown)


modules/python/python-helpers.c
```diff
@@ -376,6 +376,18 @@ _py_perform_imports(GList *imports)
   g_list_foreach(imports, _foreach_import, NULL);
 }
 
+const gchar *
+_py_object_repr(PyObject *s, gchar *buf, gsize buflen)
+{
+  const gchar *res;
+
+  PyObject *r = PyObject_Repr(s);
+  if (py_bytes_or_string_to_string(r, &res))
```
MrAnno, line 385: `PyObject_Repr()` might fail. Shouldn't we add a null-check here? `py_bytes_or_string_to_string()` is not null-safe (`PyBytes_CheckExact`, `PyUnicode_Check`, etc.)

MrAnno (commented):
I've left one minor note.

LGTM otherwise.

modules/python/python-helpers.c
bazsi: nice catch. thanks. fixed this up in the next iteration.

MrAnno (approved):

PR #4189, opened 2022-11-02
Python log message improvements

bazsi:
This is the 2nd chunk of #4165 clarifying log output of the Python module and fixing up an ugly logging issue in one of our unit tests. 

NOTE: This builds on top of #4188 but could be merged separately by rebasing to master first. I didn't want to base it on master, as that would break my larger PR or I would have to merge PRs into my mega-PR.

### DIFF
--- a/modules/python/python-dest.c
+++ b/modules/python/python-dest.c
@@ -208,7 +208,7 @@ _py_invoke_open(PythonDestDriver *self)
     {
       if (ret == Py_None)
         {
-          msg_warning_once("Since " VERSION_3_25 ", the return value of open method in python destination "
+          msg_warning_once("python-dest: Since " VERSION_3_25 ", the return value of the open() method "
                            "is used as success/failure indicator. Please use return True or return False explicitly",
                            evt_tag_str("class", self->class));
           result = TRUE;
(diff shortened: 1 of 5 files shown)


modules/python/python-dest.c
```diff
@@ -208,8 +208,9 @@ _py_invoke_open(PythonDestDriver *self)
     {
       if (ret == Py_None)
         {
-          msg_warning_once("Since " VERSION_3_25 ", the return value of open method in python destination "
-                           "is used as success/failure indicator. Please use return True or return False explicitly",
+          msg_warning_once("python-dest: Since " VERSION_3_25 ", the return value of the open() method "
+                           "is used as success/failure indicator. Please use return True or return False explicitly "
+                           "instead of returning None",
```
MrAnno, line 213: This warning message is also shown when the user code does not return anything.

modules/python/python-source.c
```diff
@@ -221,25 +221,26 @@ _py_resolve_class(PythonSourceDriver *self)
...
+                evt_tag_str("class-repr", _py_object_repr(self->py.class, buf, sizeof(buf))),
                 evt_tag_str("exception", _py_format_exception_text(buf, sizeof(buf))));
       _py_finish_exception_handling();
       return FALSE;
     }
 
   if (!_py_is_log_source(self->py.instance))
     {
-      msg_error("Error initializing Python source, class is not a subclass of LogSource",
+      msg_error("python-source: Error instantiating Python driver calss, class is not a subclass of LogSource",
```
MrAnno, line 240: typo: calss

MrAnno (commented):
I've found just a typo, LGTM otherwise.

modules/python/python-source.c
bazsi: fixed.

modules/python/python-dest.c
bazsi: if the user code does not have a return statement, that is the same as returning None in my book. but you are right,
users may be inexperienced Python coders, so omitting that extra bit of information.

MrAnno (approved):

PR #4192, opened 2022-11-02
Python support for python implemented log functionality in tree

bazsi:
This is chunk 5 from #4165 

This implements the framework to host Python modules that are part of the source tree.  This was originally implemented as a separate directory (e.g. we had modules/python/pylib and modules/python-modules) and then  merged as I worked toward a workable environment where I can both easily deploy and develop Python modules. There are remains of this in the commit history, first I was working on "just" making it possible to deploy an example plugin, developed in Python, then later the two separate directory structures were merged into one.


modules/python-modules/Makefile.am
```diff
@@ -0,0 +1,125 @@
...
+
+#
+# The syslogng package is installed into ${prefix}/lib/syslog-ng, while its
+# requirements all installed to the normal Python site-packages directory.
+#
+# Therefore we need to separate the installation of both.
+#
+pymodules-install: pymodules-install-requirements
+	(cd $(PYMODULES_SRCDIR) && $(PYTHON) setup.py \
+		build --build-base="$(PYMODULES_BUILDDIR)/build" \
```
alltilla, line 78: Shouldn't `build` be during build (`make`) instead of install (`make install`)? Building/installing with cmake does that and it seems cleaner.

alltilla:
[optional]

Not sure why, but `__pycache__` folders are present in the install dir:
```
$ find -name __pycache__
./lib/syslog-ng/python/syslogng/__pycache__
./lib/syslog-ng/python/syslogng/debuggercli/__pycache__
./lib/syslog-ng/python/syslogng/modules/example/__pycache__
```
They don't seem to appear in the final `mod-python.deb` file, except the one from `debuggercli`:
```
$ find -name __pycache__
./3.38/python/syslogng/debuggercli/__pycache__
```

We might not want to fix this right now, but I though I mention.

alltilla:
The PR seems to do its job :) Nice improvement, thank you for implementing it!

I am planning to read the README.md thoroughly, then my review will be finished.

modules/python-modules/Makefile.am
bazsi: We don't really need the build step, and install does that automatically. The reason it is there explicitly is that we want to change the default build directory and this is the only way to do that.

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+    def init(self, options):
+        self.ack_tracker = BatchedAckTracker(timeout=500, batch_size=100,
+                                             batched_ack_callback=self.messages_acked)
+        return True
+
+    def messages_acked(self, acked_message_bookmarks):
+        pass
+```
+
+BatchedAckTracker would call your callback every now and then, as specified
```
alltilla, line 479: [optional]

BatchedAckTracker is souce code, so it could be backticked.
```suggestion
`BatchedAckTracker` would call your callback every now and then, as specified
```

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+by the `timeout` argument in milliseconds. `batch_size` specifies the number of
+outstanding messages at a time. With this interface it's quite easy to
+perform acknowledgements back to the source interface where per-message
+acknowledgements are needed (e.g. Google PubSub).
+
+### Creating a syslog-ng a LogSource based Source plugin
```
alltilla, line 485: This does not seem to be gramatically correct.
```suggestion
### Creating a syslog-ng LogSource based Source plugin
```

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+operation in a sequential manner: you fetch a batch of messages, feed the
+syslog-ng pipeline with those messages and then repeat.
+
+LogSource is more low-level but allows the use of an asynchronous framework
+(asyncio, etc) to perform message fetching along multiple threads of
+execution.
+
+This sample uses asyncio to generate two independent sequences of messages,
+one is generated every second, the other is every 1.5 seconds, running
+concurrently via an asyncio event loop.
```
alltilla, line 498: [optional]

LogFetcher, asyncio and LogSource are source codes, so they could be backticked.
```suggestion
While `LogFetcher` gives us a convinient interface for fetching messages from
backend services via blocking APIs, it is limited to performing the fetching
operation in a sequential manner: you fetch a batch of messages, feed the
syslog-ng pipeline with those messages and then repeat.

`LogSource` is more low-level but allows the use of an asynchronous framework
(`asyncio`, etc) to perform message fetching along multiple threads of
execution.

This sample uses `asyncio` to generate two independent sequences of messages,
one is generated every second, the other is every 1.5 seconds, running
concurrently via an `asyncio` event loop.
```

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+            await asyncio.sleep(1.5)
+            self.post_message(LogMessage("message2"))
+
+    def request_exit(self):
+        self.cancelled = True
+
+```
+
+Acknowledgement mechanisms (ConsecutiveAckTracker, BatchedAckTracker) can be
+used similarly to how it was described at LogFetcher.
```
alltilla, line 534: [optional]

ConsecutiveAckTracker, BatchedAckTracker and LogFetcher are source codes, so they could be backticked.
```suggestion
Acknowledgement mechanisms (`ConsecutiveAckTracker`, `BatchedAckTracker`) can be
used similarly to how it was described at `LogFetcher`.
```

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+
+## Making it more native
```
alltilla, line 536: In the beginning of this readme, you have mentioned that python should be used, where there is no native solution to solve an issue. To me this title (`Making it more native`) suggests something like delegating business logic to C code then calling them in the python code, maybe I am strange. I would suggest to rephrase this a bit.
```suggestion
## Making it more native config-wise
```

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+PYTHONPATH (normally /etc/syslog-ng/python), with that in mind
+add the following code to `/etc/syslog-ng/python/mytemplate.py`:
+
+
+```python
+
+def template_function(msg):
+    return b"Hello World from Python! Original message: " + msg['MSGHDR'] + msg['MESSAGE']
+
+
```
alltilla, line 72: [optional]

There are unnecessary newlines.
```suggestion
def template_function(msg):
    return b"Hello World from Python! Original message: " + msg['MSGHDR'] + msg['MESSAGE']
```

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+        self.cancelled = True
+
+```
+
+Acknowledgement mechanisms (ConsecutiveAckTracker, BatchedAckTracker) can be
+used similarly to how it was described at LogFetcher.
+
+## Making it more native
+
+All the examples above used some form of python() driver to be actually
```
alltilla, line 538: [optional]

python() can be considered as source code, so it could be backticked.
```suggestion
All the examples above used some form of `python()` driver to be actually
```

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+syntax to users of your code that is more convinient to use.
+
+```
+block destination my-destination(option1(value)
+                                 option2(value)) {
+    python(class(mydestination.MyDestination)
+                 options(option1 => `option1`,
+                         option2 => `option2`));
+
+
```
alltilla, line 570: [optional]

These newlines are unnecessary.
```suggestion
```

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+functionality.
+
+You should add this wrapper to your Python module in an "scl" subdirectory
+as a file with a .conf extension. syslog-ng will automatically include these
+files along the rest of the SCL.
+
+## Adding the code to syslog-ng
+
+To add Python based modules to syslog-ng, create a Python package (e.g.  a
+directory with __init__.py and anything that file references).  Add these
```
alltilla, line 585: The double underscore made this bold. We should backtick it maybe.
```suggestion
directory with `__init__.py` and anything that file references).  Add these
```

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+file.
+
+If you would like to use pip/requirements.txt to deploy dependencies, you
+can choose to: invoke pip at "make install" time so that syslog-ng's private
+Python directory would contain all the dependencies that you require.
+
+## Adding Python code to syslog-ng deb package
+
+To add your module to the syslog-ng deb package, create a new file in
+packaging/debian/ with a name like `syslog-ng-mod-<yourmodule>.install`.
```
alltilla, line 614: [optional]

Other paths in this readme have been backticked, we could backtick this also.
```suggestion
`packaging/debian/` with a name like `syslog-ng-mod-<yourmodule>.install`.
```

alltilla (commented):
This readme is pure gold, we really needed an up-to-date guide for the python binding. Also it has some very well written low level syslog-ng behavior documentation, too. Thanks for this!

Sorry, I have had a lot of formatting comments, mostly missing backticks. `(Maybe I have an unhealthy attraction to backticks :D)` Most of the comments are optional, so feel free to ignore them.

To make "fixing" the comments easier, I have prepared code change suggestions for each comment, so if everything works, you can accept them, add them to a batch and commit them together.

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+
+Once all required methods are implemented, you can use this using the
+"python" destination in the syslog-ng configuration language.
+
+```
+
+destination whatever {
+    python(class(mydestination.MyDestination));
+};
+
```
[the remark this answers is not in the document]
alltilla: The suggestions around the empty newlines are strange it seems like GitHub ignores the removal of the trailing newlines. If you agree to remove the trailing newlines, please make sure that commiting the suggested change does it properly.

modules/python-modules/README.md
```diff
@@ -0,0 +1,648 @@
...
+```
+
+There's a more complete example destination in the python_example()
+destination plugin, that is located in the directory
+`modules/python-modules/example/` within the source tree or the same files
+installed under `${exec_prefix}/syslog-ng/python/syslogng/modules` in a
+production deployment.
```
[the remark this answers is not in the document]
bazsi: I have this in syslog-ng-mod-examples:

```
$ cat packaging/debian/syslog-ng-mod-examples.install
usr/lib/syslog-ng/*/libexamples.so
usr/lib/syslog-ng/*/python/syslogng/modules/example/*.py
usr/lib/syslog-ng/*/python/syslogng/modules/example/scl/*
```

so it should definitely contain these files.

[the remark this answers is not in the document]
alltilla: Hmm really it is there, how did I miss it?

alltilla (approved):
Approve!

The commits-check fails, because the last commit does not have a `Signed-off` line. Once it gets fixed, I will reapprove :)

bazsi:
> [optional]
> 
> Not sure why, but `__pycache__` folders are present in the install dir:
> 
> ```
> $ find -name __pycache__
> ./lib/syslog-ng/python/syslogng/__pycache__
> ./lib/syslog-ng/python/syslogng/debuggercli/__pycache__
> ./lib/syslog-ng/python/syslogng/modules/example/__pycache__
> ```
> 
> They don't seem to appear in the final `mod-python.deb` file, except the one from `debuggercli`:
> 
> ```
> $ find -name __pycache__
> ./3.38/python/syslogng/debuggercli/__pycache__
> ```
> 
> We might not want to fix this right now, but I though I mention.

I am not sure what we want to do with __pycache__ directories. Python packages don't have it, so we probably shouldn't either. I just pushed a patch that should exclude it from our mod-python package.

alltilla (approved):

PR #4202, opened 2022-11-03
modules/timestamp: add $(format-date) template function

bazsi:
$(format-date [options] format-string [timestamp])

$(format-date) takes a timestamp in the DATETIME representation and formats it according to an strftime() format string.  The DATETIME representation in syslog-ng is a UNIX timestamp formatted as a decimal number, with an optional fractional part, where the seconds and the fraction of seconds are separated by a dot.

If the timestamp argument is missing, the timestamp of the message is used.

Options:
  --time-zone <TZstring> -- override timezone of the original timestamp


bazsi:
Note to self, this should probably be $(format-date) to match date-parser()

kira-syslogng:
Build FAILURE

MrAnno (approved):

PR #4206, opened 2022-11-10
Don't recommend logrotate in Debian packaging

bazsi:
This PR moves logrotate from being Recommended to Suggested, which causes our docker image to be a lot leaner, as logrotate pulls in a lot of dependencies.


MrAnno:
AFAIK this won't have any negative side-effects on our APT repository, `apt upgrade` will not remove logrotate when updating from previous versions.

MrAnno:
Related thread:
balabit/syslog-ng-docker#52

We concluded that `logrotate` is out of scope of our docker image.

MrAnno (approved):

PR #4207, opened 2022-11-10
ci: upgrade macOS OpenSSL to v3.0

MrAnno:
This is used as a dependency in the macOS CI job.


MrAnno:
Closing as it turned out that the `brew` version of `curl` still depends on OpenSSL 1.1, which is the dependency of one of our plugins.

PR #4209, opened 2022-11-11
#3993 support ssl conf cmd

### LINKED ISSUE #3785: Syslog is not sending ca names
Hello

One of our customers recognized that the syslog-ng server is not sending ca names during the TLS handshake on the source connector.

Here is our configuration:
```
source customer_int_tls_source {
        syslog(ip(x.x.x.x)
                port(7003)
                transport("tls")
                tls(
                        key-file("/etc/syslog-ng/cert.d/server_ext.key")
                        cert-file("/etc/syslog-ng/cert.d/server_ext.pem")
                        ca-dir("/etc/syslog-ng/ca.d/customer_int")
                        peer-verify(required-trusted)
                        ssl-options(no-sslv2, no-sslv3, no-tlsv1, no-tlsv11)
                        trusted-keys(
                                "SHA1:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00"
                        )

                )
        );
};

```

And here are the list of our ca-dir:

```
[root@server customer_int]# ls -al /etc/syslog-ng/ca.d/customer_int/
total 8
drwxr-xr-x. 2 root root  168 Sep 14 13:53 .
drwxr-xr-x. 9 root root 4096 Jul 22 09:38 ..
lrwxrwxrwx. 1 root root   46 Feb 10  2021 22b7e9b1.0 -> /etc/syslog-ng/ca.d/RootCA.pem
lrwxrwxrwx. 1 root root   44 Jul  7 13:12 35448bde.0 -> /etc/syslog-ng/ca.d/IntermediateCA.pem
```

Howewer openssl s_client tells the following:
```
---
No client certificate CA names sent
Requested Signature Algorithms: ECDSA+SHA256:ECDSA+SHA384:ECDSA+SHA512:Ed25519:Ed448:RSA-PSS+SHA256:RSA-PSS+SHA256:RSA-PSS+SHA384:RSA-PSS+SHA384:RSA-PSS+SHA512:RSA-PSS+SHA512:RSA+SHA256:RSA+SHA384:RSA+SHA512:ECDSA+SHA224:RSA+SHA224:ECDSA+SHA1:RSA+SHA1
Shared Requested Signature Algorithms: ECDSA+SHA256:ECDSA+SHA384:ECDSA+SHA512:Ed25519:Ed448:RSA-PSS+SHA256:RSA-PSS+SHA256:RSA-PSS+SHA384:RSA-PSS+SHA384:RSA-PSS+SHA512:RSA-PSS+SHA512:RSA+SHA256:RSA+SHA384:RSA+SHA512
Peer signing digest: SHA256
Peer signature type: RSA-PSS
Server Temp Key: X25519, 253 bits
---
SSL handshake has read 6439 bytes and written 429 bytes
Verification: OK
---
```

Is there an option for enabling sending ca names? The problem is that the library of the customer does not send a client certificate if there is no matching CA sent from the server.

Thank you very much
willcey

### LINKED ISSUE #3993: Support SSL_CONF_cmd() to configure advanced client tls options

HofiOne:
SSL_CONF_cmd is added to support a common interface for setting the so many various SLL_CTX and SSL options
that can be set earlier via separated openssl function calls (e.g. SSL_CTX_set_options(), SSL_CTX_set_min_proto_version(), SSL_CTX_set_cipher_list(), etc.)

IMPORTANT:
The newly introduced `openssl-conf-cmds` always has the highest priority, its content parsed last,
so it will override any other options can be found in the `tls()` section, does not matter if they
appear before or after `openssl-conf-cmds`.

Regarding to the other `tls()` options.
As described in the SSL_CONF_cmd documentation, the order of operations is significant and executed in
top-down order, basically it means if there are multiple occurrences of setting the same option then the 'last wins'.
This is also true for options that can be set multiple ways (e.g. used cipher suites and/or protocols).
For further info please see: https://www.openssl.org/docs/man1.1.1/man3/SSL_CONF_cmd.html

About the implementation.
It would have been much simpler and cleaner to apply `conf-cmds()` lines immediately during the parsing in afsocket-grammar.y file but (mainly) because of the above (ordering) facts I stayed with the same initialization style that already can be seen e.g. for `trusted_keys()`
Also for simplicity and effectiveness I stayed with the usage of GList (even though it adds a tiny restriction in the afsocket-grammar.y on how the parsed elements must be pushed onto the bison stack about that I've added a comment in the file)

Tests are made manually in two ways.
- I've set up a server/client syslog-ng configuration both with simple and mutual authentication using TLS then set up multiple ciphers setting in many, various ways via `conf-cmds()` (matching, none matching, none existing cipher pairs) to test if the client could establish a TLS connection to the server and send log messages via TLS.
- Using the server instance from the above configuration I've also set up many different kinds of TLS setting via `conf-cmds()` and run sslscan (https://github.com/rbsec/sslscan) then checked/compared the output against my `conf-cmds()` settings.

Fixes #3993
Resolves #3785

### DIFF
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -876,6 +876,7 @@ tls_context_setup_cmd_context(TLSContext *self)
                       tls_context_format_location_tag(self));
           result = result && (add_result >= 1);
         }
+      result = result && (SSL_CONF_CTX_finish(ssl_conf_ctx) == 1);
     }
 
   SSL_CONF_CTX_free(ssl_conf_ctx);
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -503,6 +503,7 @@ tls_session_set_trusted_fingerprints(TLSContext *self, GList *fingerprints)
 {
   g_assert(fingerprints);
 
+  g_list_foreach(self->trusted_fingerprint_list, (GFunc) g_free, NULL);
   self->trusted_fingerprint_list = fingerprints;
 }
 
@@ -511,6 +512,7 @@ tls_session_set_trusted_dn(TLSContext *self, GList *dn)
 {
   g_assert(dn);
 
+  g_list_foreach(self->trusted_dn_list, (GFunc) g_free, NULL);
   self->trusted_dn_list = dn;
 }
 
--- a/lib/tlscontext.c
+++ b/lib/tlscontext.c
@@ -66,6 +66,7 @@ struct _TLSContext
   gboolean ocsp_stapling_verify;
 
   SSL_CTX *ssl_ctx;
+  GList *conf_cmds_list;
   GList *trusted_fingerprint_list;
   GList *trusted_dn_list;
   gint ssl_options;
@@ -848,6 +849,39 @@ tls_context_setup_sigalgs(TLSContext *self)
   return TRUE;
 }
 
+static gboolean
+tls_context_setup_cmd_context(TLSContext *self)
+{
+  SSL_CONF_CTX *ssl_conf_ctx = SSL_CONF_CTX_new();
+  const int ctx_flags = SSL_CONF_FLAG_FILE | SSL_CONF_FLAG_CLIENT | SSL_CONF_FLAG_SERVER |
+                        SSL_CONF_FLAG_CERTIFICATE | SSL_CONF_FLAG_SHOW_ERRORS;
+  int set_flags = SSL_CONF_CTX_set_flags(ssl_conf_ctx, ctx_flags);
+
+  gboolean result = (set_flags == ctx_flags);
+  if (result)
+    {
+      SSL_CONF_CTX_set_ssl_ctx(ssl_conf_ctx, self->ssl_ctx);
+
+      for (GList *next_pair = self->conf_cmds_list; next_pair; next_pair = next_pair->next)
+        {
+          gchar *cmd = next_pair->data;
+          next_pair = next_pair->next;
+          gchar *value = next_pair->data;
+
+          int add_result = SSL_CONF_cmd(ssl_conf_ctx, cmd, value);
+          if (add_result < 1)
+            msg_error("Error setting up SSL conf context, invalid cmd or value in config list",
+                      evt_tag_str("cmd", cmd),
+                      evt_tag_str("value", value),
+                      tls_context_format_location_tag(self));
+          result = result && (add_result >= 1);
+        }
+    }
+
+  SSL_CONF_CTX_free(ssl_conf_ctx);
+  return result;
+}
+
 static PKCS12 *
 _load_pkcs12_file(TLSContext *self, const gchar *pkcs12_file)
 {
@@ -1008,11 +1042,7 @@ tls_context_setup_context(TLSContext *self)
 
   tls_context_setup_ssl_options(self);
   if (!tls_context_setup_ecdh(self))
-    {
-      SSL_CTX_free(self->ssl_ctx);
-      self->ssl_ctx = NULL;
-      return TLS_CONTEXT_SETUP_ERROR;
-    }
+    goto error_no_print;
 
   if (!tls_context_setup_dh(self))
     goto error;
@@ -1023,10 +1053,14 @@ tls_context_setup_context(TLSContext *self)
   if (!tls_context_setup_sigalgs(self))
     goto error;
 
+  if (!tls_context_setup_cmd_context(self))
+    goto error;
+
   return TLS_CONTEXT_SETUP_OK;
 
 error:
   _print_and_clear_tls_session_error(self);
+error_no_print:
   if (self->ssl_ctx)
     {
       SSL_CTX_free(self->ssl_ctx);
@@ -1096,6 +1130,7 @@ _tls_context_free(TLSContext *self)
 {
   g_free(self->location);
   SSL_CTX_free(self->ssl_ctx);
+  g_list_foreach(self->conf_cmds_list, (GFunc) g_free, NULL);
   g_list_foreach(self->trusted_fingerprint_list, (GFunc) g_free, NULL);
   g_list_foreach(self->trusted_dn_list, (GFunc) g_free, NULL);
   g_free(self->key_file);
@@ -1375,6 +1410,14 @@ tls_context_set_client_sigalgs(TLSContext *self, const gchar *sigalgs, GError **
               "Setting client sigalgs is not supported with the OpenSSL version syslog-ng was compiled with");
   return FALSE;
 #endif
+}
+
+gboolean
+tls_context_set_conf_cmds(TLSContext *self, GList *cmds, GError **error)
+{
+  g_list_foreach(self->conf_cmds_list, (GFunc) g_free, NULL);
+  self->conf_cmds_list = cmds;
+  return TRUE;
 }
 
 void
--- a/modules/afsocket/afsocket-parser.c
+++ b/modules/afsocket/afsocket-parser.c
@@ -66,6 +66,7 @@ static CfgLexerKeyword afsocket_keywords[] =
   { "sni",                KW_SNI },
   { "allow_compress",     KW_ALLOW_COMPRESS },
   { "ocsp_stapling_verify", KW_OCSP_STAPLING_VERIFY },
+  { "openssl_conf_cmds",  KW_CONF_CMDS},
 
   { "localip",            KW_LOCALIP },
   { "ip",                 KW_IP },
--- a/news/feature-4209.md
+++ b/news/feature-4209.md
@@ -1,0 +1,50 @@
+`network`/`syslog`/`tls` context options: SSL_CONF_cmd support
+
+SSL_CONF_cmd TLS configuration support for `network()` and `syslog()` driver has been added.
+
+OpenSSL offers an alternative, software-independent configuration mechanism through the SSL_CONF_cmd
+interface to support a common solution for setting the so many various SSL_CTX and SSL options that
+can be set earlier via multiple, separated openssl function calls only.
+This update implements that similar to the mod_ssl in Apache.
+
+IMPORTANT:
+The newly introduced `openssl-conf-cmds` always has the highest priority, its content parsed last,
+so it will override any other options can be found in the `tls()` section, does not matter if they
+appear before or after `openssl-conf-cmds`.
+
+Regarding to the other `tls()` options.
+As described in the SSL_CONF_cmd documentation, the order of operations is significant and executed in
+top-down order, basically it means if there are multiple occurrences of setting the same option then the 'last wins'.
+This is also true for options that can be set multiple ways (e.g. used cipher suites and/or protocols).
+
+Example config:
+```
+source source_name {
+    network (
+        ip(0.0.0.0)
+        port(6666)
+        transport("tls")
+        tls(
+            ca-dir("/etc/ca.d")
+            key-file("/etc/cert.d/serverkey.pem")
+            cert-file("/etc/cert.d/servercert.pem")
+            peer-verify(yes)
+
+            openssl-conf-cmds(
+                # For system wide available cipher suites use: /usr/bin/openssl ciphers -v
+                # For formatting rules see: https://www.openssl.org/docs/man1.1.1/man3/SSL_CONF_cmd.html
+                # For quick and dirty testing try: https://github.com/rbsec/sslscan
+                #
+                "CipherString" => "ECDHE-RSA-AES128-SHA",                                   # TLSv1.2 and bellow
+                "CipherSuites" => "TLS_CHACHA20_POLY1305_SHA256:TLS_AES_256_GCM_SHA384",    # TLSv1.3+ (OpenSSl 1.1.1+)
+
+                "Options" => "PrioritizeChaCha",
+                "Protocol" => "-ALL,TLSv1.3",
+            )
+        )
+    );
+};
+
+
+
+```
--- a/lib/tlscontext.h
+++ b/lib/tlscontext.h
@@ -137,6 +137,7 @@ void tls_context_set_cipher_suite(TLSContext *self, const gchar *cipher_suite);
 gboolean tls_context_set_tls13_cipher_suite(TLSContext *self, const gchar *tls13_cipher_suite, GError **error);
 gboolean tls_context_set_sigalgs(TLSContext *self, const gchar *sigalgs, GError **error);
 gboolean tls_context_set_client_sigalgs(TLSContext *self, const gchar *sigalgs, GError **error);
+gboolean tls_context_set_conf_cmds(TLSContext *self, GList *cmds, GError **error);
 void tls_context_set_ecdh_curve_list(TLSContext *self, const gchar *ecdh_curve_list);
 void tls_context_set_dhparam_file(TLSContext *self, const gchar *dhparam_file);
 void tls_context_set_sni(TLSContext *self, const gchar *sni);
--- a/modules/afsocket/afsocket-grammar.ym
+++ b/modules/afsocket/afsocket-grammar.ym
@@ -197,6 +197,7 @@ systemd_syslog_grammar_set_source_driver(SystemDSyslogSourceDriver *sd)
 %token KW_ALLOW_COMPRESS
 %token KW_KEYLOG_FILE
 %token KW_OCSP_STAPLING_VERIFY
+%token KW_CONF_CMDS
 
 /* INCLUDE_DECLS */
 
@@ -236,6 +237,9 @@ systemd_syslog_grammar_set_source_driver(SystemDSyslogSourceDriver *sd)
 
 %type <ptr> dest_failover_options
 %type <ptr> dest_failover_modes_options
+
+%type <ptr> tls_conf_cmds
+%type <ptr> tls_conf_cmd_list_build
 
 %%
 
@@ -890,9 +894,37 @@ tls_option
           {
              transport_mapper_inet_set_allow_compress(last_transport_mapper, $3);
           }
+	| KW_CONF_CMDS '(' tls_conf_cmds ')'
+		{
+			GError *error = NULL;
+			CHECK_ERROR_GERROR(tls_context_set_conf_cmds(last_tls_context, $3, &error), @3, error, "Error setting conf-cmds()");
+		}
         | KW_ENDIF {
 }
         ;
+
+tls_conf_cmds
+	: tls_conf_cmd_list_build
+		{
+			$$ = $1;
+		}
+	;
+
+tls_conf_cmd_list_build
+	: string LL_ARROW string tls_conf_cmd_list_build
+		{
+			/* Revers order is important here as this is inside of a bison right recursion */
+			$$ = g_list_prepend($4, g_strdup($3));
+			$$ = g_list_prepend($$, g_strdup($1));
+
+			free($1);
+			free($3);
+		}
+	|
+		{
+			$$ = NULL;
+		}
+  ;
 
 tls_cipher_suites
         : tls_cipher_suite tls_cipher_suites


cmake/openssl_functions.cmake
```diff
@@ -44,7 +44,8 @@ function (openssl_set_defines)
     SSL_CTX_set_num_tickets
     SSL_CTX_set_ciphersuites
     SSL_CTX_set1_sigalgs_list
-    SSL_CTX_set1_client_sigalgs_list)
+    SSL_CTX_set1_client_sigalgs_list
+    SSL_CONF_cmd)
```
bazsi, line 48: I think we could require libssl 1.0.2 where this function is already available, so we wouldn't need to check specifically for this function and we won't need to make it conditional in the codebase.

Right now, configure.ac lists 0.9.8 as minimum, and cmake does not seem to declare any version dependency, but I think it's a mistake to build syslog-ng against anything below 1.0.2 (which is centos-7's version). Honestly I feel more of requiring 1.1.0 but centos-7 is still an important platform for us.

lib/tlscontext.c
```diff
@@ -66,6 +66,10 @@ struct _TLSContext
   gboolean ocsp_stapling_verify;
 
   SSL_CTX *ssl_ctx;
+#if SYSLOG_NG_HAVE_DECL_SSL_CONF_CMD
+  SSL_CONF_CTX *ssl_conf_ctx;
+  GList *conf_cmds_list;
+#endif
```
bazsi, line 72: let's avoid making this conditional.

lib/tlscontext.c
```diff
@@ -848,6 +854,50 @@ tls_context_setup_sigalgs(TLSContext *self)
...
+
+static gboolean
+tls_context_setup_cmd_context(TLSContext *self)
+{
+#if SYSLOG_NG_HAVE_DECL_SSL_CONF_CMD
+  self->ssl_conf_ctx = SSL_CONF_CTX_new();
+
+  const int ctx_flags = SSL_CONF_FLAG_FILE | SSL_CONF_FLAG_CLIENT | SSL_CONF_FLAG_SERVER | SSL_CONF_FLAG_CERTIFICATE | SSL_CONF_FLAG_SHOW_ERRORS;
+  int set_flags = SSL_CONF_CTX_set_flags(self->ssl_conf_ctx, ctx_flags);
+  g_assert(set_flags == ctx_flags && "Could not set SSL CONF CTX flags");
```
bazsi, line 892: do we need this assert here? shouldn't we handle as errors instead of an abort() ?

lib/tlscontext.c
```diff
@@ -848,6 +854,50 @@ tls_context_setup_sigalgs(TLSContext *self)
   return TRUE;
 }
 
+#if SYSLOG_NG_HAVE_DECL_SSL_CONF_CMD
+static gboolean
+_apply_cmd_context(TLSContext *self)
+{
+  gboolean result = TRUE;
+
+  if (self->conf_cmds_list) {
```
bazsi: do we need this if? I think the for should handle the case where the head of the list is NULL.

lib/tlscontext.c
```diff
@@ -848,6 +854,50 @@ tls_context_setup_sigalgs(TLSContext *self)
...
+        gchar *cmd = next_pair->data;
+        next_pair = next_pair->next;
+        gchar *value = next_pair->data;
+
+        int add_result = SSL_CONF_cmd(self->ssl_conf_ctx, cmd, value);
+        if (add_result < 1)
+          msg_error("Error setting up SSL conf context, invalid cmd or value in config list",
+                    evt_tag_str("cmd", cmd),
+                    evt_tag_str("value", value),
+                    tls_context_format_location_tag(self));
```
bazsi, line 875: IIRC you specified SSL_CONF_FLAG_SHOW_ERRORS as well. Will I receive both errors (e.g. the ones from openssl and this one too?) do we need both?

lib/tlscontext.c
```diff
@@ -1092,10 +1140,23 @@ tls_context_new(TLSMode mode, const gchar *location)
...
-_tls_context_free(TLSContext *self)
+_tls_ssl_context_free(TLSContext *self)
 {
-  g_free(self->location);
   SSL_CTX_free(self->ssl_ctx);
+  self->ssl_ctx = NULL;
+#if SYSLOG_NG_HAVE_DECL_SSL_CONF_CMD
+  SSL_CONF_CTX_free(self->ssl_conf_ctx);
+  self->ssl_conf_ctx = NULL;
+  g_list_foreach(self->conf_cmds_list, (GFunc)g_free, NULL);
```
bazsi, line 1151: the cmd list should probably be freed in tls_context_free() and not here. we only want to free it once (a single TLSContext instance can create a new SSL instance multiple times.

news/feature-4209.md
```diff
@@ -0,0 +1,43 @@
+`network`/`syslog`/`tls` context options: SSL_CONF_cmd support
+
+SSL_CONF_cmd TLS configuration support for `network()` and `syslog()` driver has been added.
+
+SSL_CONF_cmd is added to support a common interface for setting the so many various SLL_CTX and SSL options
```
bazsi, line 5: SLL typo

news/feature-4209.md
```diff
@@ -0,0 +1,43 @@
+`network`/`syslog`/`tls` context options: SSL_CONF_cmd support
+
+SSL_CONF_cmd TLS configuration support for `network()` and `syslog()` driver has been added.
```
bazsi, line 3: I think the description here is a bit too technical and tied into OpenSSL function calls. It would make sense to explain that openssl offers an alternative, software-independent configuration mechanism through the SSL_CONF_cmd interface and this implements that. We could also mention that in apache mod_ssl a similar mechanism exists.

bazsi (commented):
It's looking good. I would try to lose the conditional compilation piece and I had a few smaller nitpicks. Pretty close to being approved on my part, but I'd love to see at least some of those comments fixed (or argued against).

bazsi:
I don't understand why this PR was closed. This seems to be genuinely useful. Are you planning to reopen it? Thanks

MrAnno:
The branch was deleted accidentally.

@HofiOne Here is the closed PR.
If you push your work back to the original branch (`#3993-Support-SSL_CONF_cmd`), we can reopen this PR (review comments will not be lost).

lib/tlscontext.c
HofiOne: > It's looking good. I would try to lose the conditional compilation piece and I had a few smaller nitpicks. Pretty close to being approved on my part, but I'd love to see at least some of those comments fixed (or argued against).

Sorry for the late answer, somehow I totally missed the notifications about the review comments, doing my best to answer within a day, thanks for the patience!

HofiOne:
> 



> It's looking good. I would try to lose the conditional compilation piece and I had a few smaller nitpicks. Pretty close to being approved on my part, but I'd love to see at least some of those comments fixed (or argued against).

Sorry for the late answer, somehow I totally missed the notifications about the review comments, doing my best to answer within a day, thanks for the patience!

lib/tlscontext.c
HofiOne: I'll remove all the conditionals, we had a round around this with MrAnno, just wanted to see your opinions too.

cmake/openssl_functions.cmake
HofiOne: MrAnno had a very close opinion on this, maybe it is time to set libssl 1.0.2 as the bare minimum?!

lib/tlscontext.c
HofiOne: I just followed the practice I found e.g. in tls_context_setup_ocsp_stapling, but you are right, it could be an error message too now (as this setup is not in the constructor anymore like was during my earlier versions)

lib/tlscontext.c
HofiOne: This call is part of _tls_ssl_context_free() that
- was born to have a common internal cleanup function (instead of the earlier code repetition) 
- is called basically 
  - from _tls_context_free() as you wrote, and
  - when we do initialization error cleanup (but those are exceptional cases) 

Also, the ssl_conf_ctx is bound to the ssl_conf_ctx (via SSL_CONF_CTX_set_ssl_ctx()) so I think they should be treated as one and handled/freed together.

lib/tlscontext.c
HofiOne: SSL_CONF_FLAG_SHOW_ERRORS is not giving enough details here, unfortunately.
This message is a must-have here I guess as it explains what key-value pair is the problematic one, unfortunately, the parser cannot give proper hint about the exact location of the problem at this point, we are too deep inside the call stack and the location info is lost somewhere during the error handling flow.

The error looks like this now:

     [2022-11-22T00:42:47.034299] Error setting up SSL conf context, invalid cmd or value in config list; cmd='InvalidKey', value='ECDHE-RSA-AES128-SHA', location='/etc/syslog-ng.conf:19:3'
     [2022-11-22T00:42:47.034334] Error setting up TLS session context; tls_error='error:1414E182:SSL routines:SSL_CONF_cmd:unknown cmd name', location='/etc/syslog-ng.conf:19:3'
     [2022-11-22T00:42:47.034346] Error initializing message pipeline; plugin_name='network', location='/etc/syslog-ng.conf:15:5'

lib/tlscontext.c
HofiOne: definitely, it's a leftover again, thanks for pointing that out!

cmake/openssl_functions.cmake
bazsi: Yes, please do so, and eliminate any conditional code (not just the ones related to this PR but the existing ones too) in the process.

lib/tlscontext.c
bazsi: cool

lib/tlscontext.c
bazsi: ok, thanks.

lib/tlscontext.c
bazsi: I think the command list itself is really a property of TLSContext. There's a method tls_context_set_conf_cmds() that sets them and I think they should remain available until TLSContext itself is freed.

It is TLSContext's responsibility  to set up an SSL_CTX based on settings stored in its members, by invoking tls_context_setup_context()

Although tls_context_setup_context() implementation seems to assume that it can only be invoked once, I still think 
that anything that is needed to setup SSL_CTX should remain to be available. 

We are not freeing cert_file or ca_file either, do we? 

I think we should review the layering here (I also have a parallel PR that touches on this subject, see #4221):
  * TLSContext is an object that stores all tls() related settings,
  * based on those settings it is able to initialize an SSL_CTX
  * based on SSL_CTX it can produce SSL instances

The initialization of SSL_CTX is decoupled from the constructor simply it is not the right time to report errors in most cases.

With that in mind, all settings (parsed from the config) should be stored in TLSContext until the life-time of that object and not freed, especially not partially.

I view tls_context_setup_context() similar to what init() does in the rest of the codebase (we have the init-style conventions in a lot of places in the codebase), init() should not free stuff that could be needed for a repeated init() call.

With all that said, the question is what should be the lifecycle for the SSL_CTX and SSL_CONF_CTX objects. SSL_CTX is currently initialized in the constructor, this would probably be better situated in the setup_context() call. 

With SSL_CONF_CTX, I view that as a temporary, which is _ONLY_ needed for initialization and not any later than that. 
  * so my conclusion would be to make SSL_CONF_CTX just a local variable in tls_context_setup_cmd_context()
  * and maybe move the creation of ssl_ctx to the setup_context() call (and free it and only it if there's an error)

This means that TLSContext->ssl_ctx being NULL clearly indicates an error: either that the setup was unsuccessful or that TLSContext wasn't even initialized.

Does this make sense?

HofiOne: I see your point, but I guess such a refactor should go to a separate run, I guess the ongoing #4221 is a good candidate, can we move it over there?

cmake/openssl_functions.cmake
HofiOne: I've found the following probably related conditionals

    #cmakedefine01 SYSLOG_NG_HAVE_DECL_EVP_MD_CTX_RESET
    #cmakedefine01 SYSLOG_NG_HAVE_DECL_ASN1_STRING_GET0_DATA
    #cmakedefine01 SYSLOG_NG_HAVE_DECL_SSL_CTX_GET0_PARAM
    #cmakedefine01 SYSLOG_NG_HAVE_DECL_SSL_CTX_SET_CIPHERSUITES
    #cmakedefine01 SYSLOG_NG_HAVE_DECL_SSL_CTX_SET1_SIGALGS_LIST
    #cmakedefine01 SYSLOG_NG_HAVE_DECL_SSL_CTX_SET1_CLIENT_SIGALGS_LIST
    #cmakedefine01 SYSLOG_NG_HAVE_DECL_SSL_CTX_SET_NUM_TICKETS
    #cmakedefine01 SYSLOG_NG_HAVE_DECL_X509_STORE_CTX_GET0_CERT
    #cmakedefine01 SYSLOG_NG_HAVE_DECL_X509_GET_EXTENSION_FLAGS
    #cmakedefine01 SYSLOG_NG_HAVE_DECL_DH_SET0_PQG
    #cmakedefine01 SYSLOG_NG_HAVE_DECL_BN_GET_RFC3526_PRIME_2048

Unfortunately, in some cases, I cannot find in the documentation which version the given functionality is introduced.

Do you mean everywhere these can be found?

bazsi: I am fine with just bumping the requirement to 1.0.2 and not checking for SSL_CONF_cmd, but assume it's there.

The existing ones can eliminated later or now, as you prefer.

lib/tlscontext.c
bazsi: ok

cmake/openssl_functions.cmake
HofiOne: I've already removed the new conditional introduced with my changes and I prefer to add that refactoring to a separate run instead, thank you!

lib/tlscontext.c
HofiOne: It must be my fault, sorry, but after reading over it once again, I can see now, we are not on the same page either ;)

What I see now (and it is the reason I implemented this way)

1. As you mentioned it is not clear how the ssl_ctx bound ssl_conf_ctx should be handled. In my interpretation, once it is bound to the ssl_ctx it is part of it and must be handled and freed together with it.
2. This is the main reason.
There's nothing in the code that prevents a given syslog-ng.conf TLS section item (same is true for other sections and items as well) to appear repeatedly.
This can be treated as a feature or a bug, I think it is a useful feature.
According to this, we have to be prepared for this case, this is why I've added 
- missing resource freeing code (more might exist, needs further investigation) in some functions that are called at the end of parsing of a given section item (e.g.  tls_session_set_trusted_dn)
- prepared setup of ssl_conf_ctx to be able to handle repeated setup calls
3. 
 ```
/* freeing of ssl_ctx and setting it to NULL should be done at the error label in tls_context_setup_context(), this should be just one location and not more. */
  SSL_CTX_free(self->ssl_ctx);
  self->ssl_ctx = NULL;
```
Exactly. It should have been, but it has not been.
In the original version, you can see there were 2 different kinds of error handling, one that cleared up the ssl_ctx without calling `_print_and_clear_tls_session_error(self);` and one with it (via goto error;)
That is why _tls_ssl_context_free() was born 
- to keep the original logic (not to call `_print_and_clear_tls_session_error(self);` when it is not needed), and
- to replace the otherwise repeated code
              SSL_CTX_free(self->ssl_ctx);
              self->ssl_ctx = NULL;
- to create a function that could handle the management of the ssl_ctx bound ssl_conf_ctx as well

Hope this makes sense.

Comment (bazsi) [reply lib/tlscontext.c:1140]: My assumption was that the SSL_CONF_ctx is not needed once the commands have been executed, so you can free that right after that. It is only bound to a specific ssl_ctx as it is making changes to it due to the commands.

Comment (HofiOne) [reply lib/tlscontext.c:1140]: the actual implementation handles nicely the possibly repeated config entries, I think we should keep this behavior

Comment (MrAnno) [reply lib/tlscontext.c:1140]: @HofiOne 

> In my interpretation, once it is bound to the ssl_ctx it is part of it and must be handled and freed together with it.

My understanding is that this statement is not true.
`SSL_CONF_CTX` can be considered as a configurator object that calls SSL and SSL_CTX API calls based on strings you provide using `SSL_CONF_cmd()`.
After you send in the last config command, the `SSL_CONF_CTX` can be freed.

This is documented behavior, even if somewhat implicitly stated.

An example of this "local" usage:
https://github.com/nginx/nginx/blob/765ef1098d169622ada67eb4eccad1d45d47a61c/src/event/ngx_event_openssl.c#L1575-L1620

File: lib/tlscontext.c [hunk 9]
```diff
@@ -846,6 +852,43 @@ tls_context_setup_sigalgs(TLSContext *self)
...
+      gchar *value = next_pair->data;
+
+      int add_result = SSL_CONF_cmd(self->ssl_conf_ctx, cmd, value);
+      if (add_result < 1)
+        msg_error("Error setting up SSL conf context, invalid cmd or value in config list",
+                  evt_tag_str("cmd", cmd),
+                  evt_tag_str("value", value),
+                  tls_context_format_location_tag(self));
+      result = result && (add_result >= 1);
+    }
```
MrAnno, line 873: It is undocumented, but we also need a `SSL_CONF_CTX_finish()` call to fill CA lists.

modules/afsocket/afsocket-parser.c
```diff
@@ -66,6 +66,7 @@ static CfgLexerKeyword afsocket_keywords[] =
   { "sni",                KW_SNI },
   { "allow_compress",     KW_ALLOW_COMPRESS },
   { "ocsp_stapling_verify", KW_OCSP_STAPLING_VERIFY },
+  { "conf_cmds",          KW_CONF_CMDS},
```
MrAnno, line 69: Can we call this feature `openssl-conf-cmd()`?
It would clarify the possible future direction when we introduce an alternative TLS library, which does not support OpenSSL commands.

MrAnno (commented):
Nice. 👍🏻 

Please don't forget to document and mention in the NEWS entry that this new option has higher precedence over all the other general TLS options.

lib/tlscontext.c
HofiOne: I Hope fixed all the discussed topics now.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

bazsi (approved):
This looks great. I would squash the review fixes into the base commits so the history is leaner once this is merged, but it's up to you.

MrAnno:
LGTM too.

Your last commit contains multiple fixes.
As @bazsi suggested it, I think they should be squashed to where it originally should have been introduced, except for the `SSL_CONF_CTX_finish` modification.
I think that one deserves a separate commit,  or at least a detailed commit message in the original commit.

I think we should mention that `SSL_CONF_CTX_finish()` is undocumented, but instead of referring to `SSL_CONF_FLAG_REQUIRE_PRIVATE`, we could document that we need this call to make `CANames` work properly (the functionality that we need in #3785).

HofiOne:
> LGTM too.
> 
> Your last 2 commits contain multiple fixes. As @bazsi suggested it, I think they should be squashed to where it originally should have been introduced, except for the `SSL_CONF_CTX_finish` modification. I think that one deserves a separate commit, or at least a detailed commit message in the original commit.
> 
> I think we should mention that `SSL_CONF_CTX_finish()` is undocumented, but instead of referring to `SSL_CONF_FLAG_REQUIRE_PRIVATE`, we could document that we need this call to make `RequestCAFile` work properly (the functionality that we need for #3785).

actually, I kept those changes based on our earlier conversion, to see why final those changes happened.
in this current case, I would keep separate commits if you do not mind.

regarding the SSL_CONF_CTX_finish(), the commit already mentioned this, would you like to see it even more detailed?

MrAnno:
> Regarding the SSL_CONF_CTX_finish(), the commit already mentioned this, would you like to see it even more detailed?

The commit message mentions `SSL_CONF_FLAG_REQUIRE_PRIVATE` and private keys, which we don't use.
I think we should mention the part that is completely undocumented: SSL_CONF_CTX_finish() is required to set CA names (certificate_authorities extension) according to the `RequestCAFile` option.

HofiOne:
> > Regarding the SSL_CONF_CTX_finish(), the commit already mentioned this, would you like to see it even more detailed?
> 
> The commit message mentions `SSL_CONF_FLAG_REQUIRE_PRIVATE` and private keys, which we don't use. I think we should mention the part that is completely undocumented: SSL_CONF_CTX_finish() is required to set CA names (certificate_authorities extension) according to the `RequestCAFile` option.

Added a bit more detailed commit message

HofiOne:
Squashed as requested

MrAnno (approved):
This is a very useful feature.

Thanks for it.

PR #4175, opened 2022-10-18
hypr audit source

bazsi:
This pull request merges in @danelder 's work on the Hypr audit trail API into the syslog-ng source tree. It changes the user interface and implementation significantly, but changes were documented in the README.md file in the source tree.

This PR contains the following items:
  * hypr-audit-trail() & hypr-app-audit-trail() source: see the README.md file in python-modules/syslogng/modules/hypr directory

The testing for hypr relies on a mocked HTTP backend, which is obviously not very thorough but does validate that the code more-or-less functions. I am unable to try the hypr source with an actual API endpoint as I lack the credentials.

Depends on #4455, #4459

PR #4214, opened 2022-11-15
Python based confgen (4.0)

bazsi:
This PR is split off from #4175 and implements Python based confgen. The original confgen module allows someone to write an external script/program to generate parts of the syslog-ng configuration.

This PR brings this into the syslog-ng process by allowing you to execute a Python function instead of the external script. An example how this can be used is in #4175 or this example:

```
@version: 4.0

python {
from syslogng import register_config_generator

def generate_foobar(args):
	print(args)
	return "tcp(port(2000))"

register_config_generator("source", "foobar", generate_foobar)

};

log {
	source { foobar(this(is) a(value)); };
	destination { file("logfile"); };
};

```

Please notice that the source declaration in the log statements towards the end of the example uses the foobar() source. That same foobar source is a registered in the Python block at the top of the file:

```
register_config_generator("source", "foobar", generate_foobar)
```

This tells syslog-ng that whenever it encounters the "foobar" token in a "source" context, it calls the function specified. That function receives the set of arguments in its "args" parameter (a dict) and can emit whatever config should be inserted in place of the foobar() reference. This is the same infrastructure that blocks use, it's just using program code to generate the resulting output.

Use cases:
- configuration that depends on something external (e.g. environment variable or some file content)
- a complex config that is easier to generate (e.g. load-balancing configs or app-parser())

### DIFF
--- a/modules/python-modules/Makefile.am
+++ b/modules/python-modules/Makefile.am
@@ -1,5 +1,6 @@
 EXTRA_DIST += \
 	modules/python-modules/syslogng/__init__.py				\
+	modules/python-modules/syslogng/confgen.py				\
 	modules/python-modules/syslogng/dest.py					\
 	modules/python-modules/syslogng/logger.py				\
 	modules/python-modules/syslogng/message.py				\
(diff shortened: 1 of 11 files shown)


alltilla (approved):
Nice one!

Please don't forget to add a news entry to the 4.0 MR. Thanks :)

PR #4221, opened 2022-11-18
Tls refactor tls context into smaller modules

bazsi:
As a preparation for #4219 I wanted to make the tlscontext module a bit more manageable. This PR moves tlscontext.c to lib/transport and splits it into 3 different modules. The interdependencies are not cleaned up completely between those modules, but I consider this a lot more readable already.

Couple of notes:
- I was considering a separate lib/tls subdirectory to hold TLSContext and friends, however it's only lib/transport that actually uses this, so I was satisfied to move it there from the really crowded lib directory
- TLSContext needed to be published to the header (while it was in the implementation module so far). The reason is that TLSSession still uses it. I could have moved it to a tls-internals.h header as the users of this API don't need the details of the struct. At the end I was lazy to do that, considering that we often publish these structs for similar reasons. We have setters/getters for all its fields, so accessing the struct members directly is off-limits.
- apart from moving the code, I tried not to change anything, apart from making it compile in the new setting. I touched the keylog functions to avoid an unnecessary indirection, but it's fairly trivial.

#4209 will probably heavily conflict with this, and the easier path forward would be to merge this first and then #4209 but I can accomodate any order as we see fit.


bazsi:
no news file is needed.

bazsi:
rebased against latest master (so the SSL_CONF_cmds conflicts are now resolved)

kira-syslogng:
Build FAILURE

lib/transport/tls-context.c
```diff
@@ -1003,6 +461,9 @@ tls_context_setup_context(TLSContext *self)
   if (!self->ssl_ctx)
     goto error;
 
+  if (!_setup_keylog_file(self))
```
MrAnno, line 464: Hmm, something is weird here.
`_setup_keylog_file()` was called for each new sessions (connections), which opened a new file for appending, but it seems to be the same file.

Moving this here changes the time and frequency of opening this file.
The time does not seem to matter as this option is used for debugging.
The "frequency change" seems correct to me, so this probably fixed an fd leak too, right?

MrAnno (commented):
I have a question to validate if I understood the situation correctly.
LGTM otherwise.

Thank you for doing this refactor.

lib/transport/tls-context.c
bazsi: It seems so. I misread one part of the code and I thought this was guarding against reopens, but now as I look at it again, this is not a guard, just checks whether the file name is set up.

```
  if(ctx->keylog_file_path)
    {
      if(!tls_session_keylog_setup_file(self, self->ctx->keylog_file_path))
        {
          tls_context_unref(self->ctx);
          g_free(self);
          return NULL;
        }
    }
```

So you are right, there has been an fd leak, but I didn't notice it.

But I think the new behaviour is a lot easier to follow.

We will open the file just once and will set the appropriate callbacks through SSL_CTX (and not SSL)

All connections will use the same file, in append mode, just as before.
I also tested this functionality manually.

bazsi:
rebased against latest master.

kira-syslogng:
Build FAILURE

lib/transport/tls-context.c
MrAnno: Thanks.

MrAnno (approved):

MrAnno:
Kira's failure is unrelated.

PR #4228, opened 2022-11-28
Apache access log fix for non http requests

bazsi:
Added NEWS file, test coverage et al.

This contains a new CSV escaping mode, unit + light tests and a couple of fixes to csv-parser()


smortex (dismissed):
Successfully tested here! Looks great!

PR #4229, opened 2022-11-28
Enhance make, cmake builds

HofiOne:
- Fixed some automake and cmake issues happened building syslog-ng on macOS (Silicon/Ventura 13.1)
- Added a module enabled state summary to configure end

image

### DIFF
--- a/cmake/print_config_summary.cmake
+++ b/cmake/print_config_summary.cmake
@@ -1,0 +1,56 @@
+# ############################################################################
+# Copyright (c) 2022 One Identity
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+# ############################################################################
+
+# Lists actual module definition results
+# Behaves similar to the cmake -L option but filters out the relevant information and adds some highlighting
+#
+function(print_config_summary)
+  get_cmake_property(_variableNames VARIABLES)
+  list(SORT _variableNames)
+
+  message(STATUS "-----------------------------")
+  message(STATUS "syslog-ng Open Source Edition ${SYSLOG_NG_VERSION} configured")
+  message(STATUS "-----------------------------")
+
+  string(ASCII 27 Esc)
+  set(Red "${Esc}[31m")
+  set(Green "${Esc}[32m")
+
+  foreach(_variableName ${_variableNames})
+    string(REGEX MATCH "^ENABLE_" _match "${_variableName}")
+
+    if(_match)
+      string(LENGTH "${_variableName}" _variableNameLen)
+      math(EXPR _variableNameLen "32 - ${_variableNameLen}")
+      string(REPEAT " " ${_variableNameLen} _spaces)
+      string(REGEX MATCH "^[Oo][Nn]" _on "${${_variableName}}")
+
+      if(_on)
+        message("${_variableName}${_spaces}${Green}On")
+      else()
+        message("${_variableName}${_spaces}${Red}Off")
+      endif()
+    endif()
+  endforeach()
+  
+endfunction ()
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -1151,6 +1151,9 @@ fi
 
 if test "$enable_afsnmp" = "yes"; then
   NETSNMP_CFLAGS="`$NETSNMP_CONFIG --cflags`"
+  # net-snmp-config --cflags could return bogus, invalid paths on arm/macOS Ventura, dirty, quick cleanup till net-snmp-config is not fixed
+  # take care, autotools require double [[]] here!!!
+  NETSNMP_CFLAGS=${NETSNMP_CFLAGS/[[ \t]]-[[0-9]]/}
```
MrAnno, line 1156: I think it's not worth adding such a nasty workaround to our configure script, because one can install net-snmp from brew as a workaround (only the system-provided net-snmp is bogus).

HofiOne: indead, will revert back this and add the environment modification requiremnts for net-snmp to the macos build documentation

cmake/Modules/FindLIBNET.cmake
```diff
@@ -41,27 +41,43 @@ FIND_PROGRAM(LIBNET_CONFIG libnet-config)
...
+  EXEC_PROGRAM(${LIBNET_CONFIG} ARGS --libs OUTPUT_VARIABLE _LIBNET_LIBRARIES)
   string(REGEX REPLACE "[\r\n]" " " _LIBNET_CFLAGS "${_LIBNET_CFLAGS}")
-  set (LIBNET_LIBRARIES ${_LIBNET_LIBRARIES} CACHE STRING "The libraries needed for LIBNET")
-  set (LIBNET_CFLAGS ${_LIBNET_CFLAGS} CACHE STRING "The compiler switches needed for LIBNET")
-  set (LIBNET_FOUND TRUE CACHE BOOL "LibNet is found")
+  string(REGEX REPLACE "[\r\n]" " " _LIBNET_LIBRARIES "${_LIBNET_LIBRARIES}")
 
-# this is due to libnet-config provides old fashined defines, which triggers warning on newer systems
-# for details see: https://github.com/libnet/libnet/pull/71
+  string(REPLACE "-I" "" _LIBNET_INCLUDES "${_LIBNET_CFLAGS}")
```
MrAnno, line 57: I think we should not modify the cflags queried from libnet-config, unless we absolutely need to.
CMake accepts `-I` as well.

cmake/dump_module_states.cmake
```diff
@@ -0,0 +1,58 @@
+# ############################################################################
+# Copyright (c) 2020 One Identity
```
MrAnno, line 2: *2022

cmake/dump_module_states.cmake
```diff
@@ -0,0 +1,58 @@
...
+
+# Lists actual module definition results
+# Behaves similar to the cmake -L option but filters out the relevant information and add some highlighting
+#
+function(dump_module_states)
+  get_cmake_property(_variableNames VARIABLES)
+  list(SORT _variableNames)
+
+  message(STATUS "---------------------------------")
+  message(STATUS "Switchable modules enabled state:")
```
MrAnno, line 32: I think we can't be sure these are modules only.

I'd just mimic a similar summary we have in autotools.

```
---------------------------------
syslog-ng Open Source Edition 3.38.1.260.geaac09a configured
---------------------------------

ENABLE_* On/Off

File: cmake/print_config_summary.cmake [hunk 5]
```diff
@@ -0,0 +1,58 @@
...
+
+# Lists actual module definition results
+# Behaves similar to the cmake -L option but filters out the relevant information and adds some highlighting
+#
+function(print_config_summary)
+  get_cmake_property(_variableNames VARIABLES)
+  list(SORT _variableNames)
+
+  message(STATUS "-----------------------------")
+  message(STATUS "syslog-ng Open Source Edition ${PKG_CONFIG_PACKAGE_VERSION} configured")
```
MrAnno, line 32: One last minor thing:

Can we use SYSLOG_NG_VERSION here? pkg-config is a helper tool for compilation, it does not really make sense using one of its variables here ([PACKAGE_VERSION](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Initializing-configure.html) in autotools is actually the project's version).

cmake/print_config_summary.cmake
```diff
@@ -0,0 +1,58 @@
...
+
+      if(_on)
+        message("${_variableName}${_spaces}${Green}On")
+      else()
+        message("${_variableName}${_spaces}${Red}Off")
+      endif()
+    endif()
+  endforeach()
+
+# message(STATUS "")
```
MrAnno, line 56: This was accidentally forgotten here.

cmake/print_config_summary.cmake
HofiOne: upps, searched for the autotools [PACKAGE_VERSION](https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Initializing-configure.html) and when I found a reference in the cmake files I blindly used the first one :S , shame on me

cmake/print_config_summary.cmake
HofiOne: thanks!
seems I can practice interactive rebase again :D

HofiOne: ok, a simple amend did the trick this time, maybe next time :)
hope, now it would be fine!

cmake/print_config_summary.cmake
MrAnno: Thank you.

MrAnno (approved):
A color reset (`[m`) might be missing, but it seems to be okay even without it.

PR #4231, opened 2022-12-03
GitHub Actions: fix (some) deprecation warnings

OverOrion:
Currently there are some warnings in the CI (see https://github.com/syslog-ng/syslog-ng/actions/runs/3608033376 for example).

- [X]  Nodejs 12 deprecation --> simple version bumps should take care of it
- [X]  [GitHub Actions: Deprecating save-state and set-output commands ](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) --> I updated the `set-output` as suggested by the official blog post.
- [ ] [CI @ devshell: .github#L1](https://github.com/syslog-ng/syslog-ng/commit/2debcbe11bc66e70444cf6b0db108d30f472fb18#annotation_6685971219):
> The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002

This PR addresses the first two (as they are straightforward to fix) and I will happily fix the `ubuntu-18.04` environment deprecation as well, just tell me which version it should be bumped to.

No news file needed.

### DIFF
--- a/.github/workflows/stable-release.yml
+++ b/.github/workflows/stable-release.yml
@@ -1,5 +1,8 @@
-name: Stable release
+# Please note that the syslog-ng git repository is not available in this workflow.
+# It means that certain convenience/helper functions are not available (e.g. gh_output).
+# This is intentional as syslog-ng will be acquired from the tarball.
 
+name: Stable release
 
 on:
   workflow_dispatch:
--- a/.github/workflows/stable-release.yml
+++ b/.github/workflows/stable-release.yml
@@ -42,7 +42,7 @@ jobs:
             false
           fi
 
-          echo "::set-output name=DRAFT_RELEASE_RUN_ID::${DRAFT_RELEASE_RUN_ID}"
+          echo "{DRAFT_RELEASE_RUN_ID}=${DRAFT_RELEASE_RUN_ID}" >> $GITHUB_OUTPUT
 
   index-packages:
     needs: find-draft-release-run
--- a/.github/workflows/test-apt-packages.yml
+++ b/.github/workflows/test-apt-packages.yml
@@ -1,5 +1,8 @@
-name: Run smoke tests on APT packages
+# Please note that the syslog-ng git repository is not available in this workflow.
+# It means that certain convenience/helper functions are not available (e.g. gh_output).
+# This is intentional as syslog-ng will be acquired from the APT repository.
 
+name: Run smoke tests on APT packages
 on:
   workflow_call:
     inputs:
--- a/.github/workflows/test-apt-packages.yml
+++ b/.github/workflows/test-apt-packages.yml
@@ -54,7 +54,7 @@ jobs:
       - name: Get syslog-ng revision
         run: |
           syslog-ng -V
-          echo "::set-output name=REVISION::$(syslog-ng -V | grep Revision | cut -d" " -f2)"
+          echo "{REVISION}={$(syslog-ng -V | grep Revision | cut -d" " -f2)}" >> $GITHUB_OUTPUT
         id: syslog_ng_revision
 
       - name: Check revision
--- a/.github/workflows/comment-on-version-bump-pr.yml
+++ b/.github/workflows/comment-on-version-bump-pr.yml
@@ -21,7 +21,7 @@ jobs:
       COMMIT_URL: https://github.com/${{ github.repository_owner }}/syslog-ng/commit/${{ github.sha }}
     steps:
       - name: Checkout syslog-ng source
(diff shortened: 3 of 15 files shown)


kira-syslogng:
do nothing -> CI won't start)

MrAnno:
@kira-syslogng ok to test

.github/workflows/stable-release.yml
```diff
@@ -42,7 +42,7 @@ jobs:
             false
           fi
 
-          echo "::set-output name=DRAFT_RELEASE_RUN_ID::${DRAFT_RELEASE_RUN_ID}"
+          echo "{DRAFT_RELEASE_RUN_ID}=${DRAFT_RELEASE_RUN_ID}" >> $GITHUB_OUTPUT
```
bazsi, line 45: We have a gh_output function in our shell script library that could be used here.

.github/workflows/test-apt-packages.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - name: Get syslog-ng revision
         run: |
           syslog-ng -V
-          echo "::set-output name=REVISION::$(syslog-ng -V | grep Revision | cut -d" " -f2)"
+          echo "{REVISION}={$(syslog-ng -V | grep Revision | cut -d" " -f2)}" >> $GITHUB_OUTPUT
```
bazsi, line 49: Likewise

bazsi (commented):
This looks good, but would be great to migrate to gh_output if possible, as indicated in the comments.

.github/workflows/stable-release.yml
OverOrion: Thank you, I switched to it in 517978d.

.github/workflows/test-apt-packages.yml
OverOrion: Thank you, I switched to it in 517978d.

.github/workflows/stable-release.yml
alltilla: Both places where `set-output` is used is done like that because we do not checkout the `syslog-ng` git repo. Both jobs are acquiring `syslog-ng` from other sources, like the `stable-release` job through the tarball and the `test-apt-packages` job through `apt`. We intentionally refused to checkout the git repo to make sure that no source code is necessary for those tasks.

I think these should be kept this way, but even if we are okay with sourcing the script, we should add a checkout first.

Also, please test the modified jobs in your fork, as some of them are not executed under PRs. We could have broken `stable-release` and `test-apt-packages` if we were not careful.

bazsi (dismissed):

.github/workflows/stable-release.yml
bazsi: I didn't know this, sorry. Makes sense, but maybe a comment would be great to have in those workflows so this information is recorded.

OverOrion: Thanks for the heads up @alltilla, will do test runs and then get back to it :+1:

I could also add a comment as well about the intention, let me know if I should do so.

alltilla: Thank you! Yes, if you could do that I would be really happy :)

OverOrion: I tried running the modified jobs in my fork, but [create-packages / debian-stretch](https://github.com/OverOrion/syslog-ng/actions/runs/3725510797/jobs/6318380310#logs) failed. I checked out master and currently it fails as well, which is why I submitted a new issues for that: #4246.

bazsi:
Apart from the last commit, this looks good to me. Thanks. Rebase with the last commit removed and I'd merge this.

On the Debian stretch issue, I've just opened a new PR. This one doesn't cause that breakage so this can go in.

.github/workflows/stable-release.yml
bazsi: This one got resolved by me dropping Debian stretch support.

bazsi:
I have removed the last patch and now waiting for the CI to finish. With that I'd like to merge this, the warnings are really bothering me :)

bazsi:
I would do the Ubuntu image switch in a different PR and merge the rest.

bazsi (approved):

bazsi:
@HofiOne now as 4.0.0 has been released, could you please merge the @kira-syslogng tests that fix the test results from Kira? I am merging this even though @kira-syslogng was red, as this one does not touch code at all, only github changes are included here. Thanks

Issue #4230, opened 2022-12-02
'make check fails to link: error: only libraries may -dlpreopen a convenience library: './modules/python/libmod-python.la'


yurivict:
# syslog-ng
## Version of syslog-ng
3.38.1

## Platform
FreeBSD 13.1

## Debug bundle

# Issue
## Failure
```
/bin/sh ./libtool  --tag=CC   --mode=link cc -DTOP_SRCDIR=\"/disk-samsung/freebsd-ports/sysutils/syslog-ng/work/syslog-ng-3.38.1\" -Wshadow -Wno-stack-protector -Wno-unused-parameter -Wno-variadic-macros   -I/usr/local/include -DBXF_STATIC_LIB=1   -I./modules/python -O2 -g -pthread -Wno-initializer-overrides -O2 -pipe  -I/usr/local/include/python3.9 -fstack-protector-strong -fno-strict-aliasing   -fstack-protector-strong  -no-install -no-pie -o modules/python/tests/test_python_logmsg modules/python/tests/test_python_logmsg-test_python_logmsg.o ./libtest/libsyslog-ng-test.a -L/usr/local/lib -lcriterion  ./lib/libsyslog-ng.la -lm -lrt -L/usr/local/lib -Wl,--export-dynamic -lgmodule-2.0 -lglib-2.0 -lgthread-2.0 -pthread -lglib-2.0 -lintl  ./lib/eventlog/src/libevtlog.la ./lib/secret-storage/libsecret-storage.la   -L/usr/local/lib -lpcre  -lregex -ldl -dlpreopen ./modules/python/libmod-python.la -dlpreopen ./modules/syslogformat/libsyslogformat.la  
libtool:   error: only libraries may -dlpreopen a convenience library: './modules/python/libmod-python.la'
gmake[4]: *** [Makefile:15135: modules/python/tests/test_python_logmsg] Error 1
gmake[3]: *** [Makefile:26732: check-am] Error 2
gmake[2]: *** [Makefile:24985: check-recursive] Error 1
gmake[1]: *** [Makefile:26735: check] Error 2
```

## Steps to reproduce
'make check'

clang-14
FreeBSD 13.1

bazsi:
This is strange as the rest of those test programs have used those linker options forever. There's only a slight ordering difference, can you check if this improves the situation?

```
diff --git a/modules/python/tests/Makefile.am b/modules/python/tests/Makefile.am
index e02e93e98..0c65d4986 100644
[...]

PR #4232, opened 2022-12-05
Freebsd unit test fixes

bazsi:
No NEWS file is needed, especially until the fixes in this branch are confirmed.

@yurivict I couldn't reproduce the test_python_logmsg breakage on my local FreeBSD box (e.g. as you reported it in #4230), but while testing that I've found an unrelated issue that I've fixed in this branch. 

Can you please check if it fixes both issues in your environment? Thanks.

### DIFF
--- a/lib/timeutils/cache.c
+++ b/lib/timeutils/cache.c
@@ -138,9 +138,9 @@ _capture_timezone_state_from_variables(void)
    * static, as we are either just starting up or being reloaded.
    **/
 
+  global_state.timezone = _get_system_tzofs();
   global_state.tzname[0] = g_strdup(tzname[0]);
   global_state.tzname[1] = g_strdup(tzname[1]);
-  global_state.timezone = _get_system_tzofs();
 }
 
 /* NOTE: copy the contents of the global cache to thread local variables */
--- a/modules/python/tests/Makefile.am
+++ b/modules/python/tests/Makefile.am
@@ -16,8 +16,7 @@ modules_python_tests_TESTS = \
 modules_python_tests_test_python_logmsg_CFLAGS = $(TEST_CFLAGS) $(PYTHON_CFLAGS) -I$(top_srcdir)/modules/python
 modules_python_tests_test_python_logmsg_LDADD = $(TEST_LDADD) \
 	-dlpreopen $(top_builddir)/modules/python/libmod-python.la \
-	$(PREOPEN_SYSLOGFORMAT) \
-	$(PYTHON_LIBS)
+	$(PYTHON_LIBS) $(PREOPEN_SYSLOGFORMAT)
 
 modules_python_tests_test_python_template_CFLAGS = $(TEST_CFLAGS) $(PYTHON_CFLAGS) \
 	-I$(top_srcdir)/modules/python -I$(top_srcdir)/modules/syslogformat
--- a/modules/python/tests/test_python_logmsg.c
+++ b/modules/python/tests/test_python_logmsg.c
@@ -95,6 +95,7 @@ _construct_py_parse_options(void)
 void
 setup(void)
 {
+  setenv("TZ", "UTC", TRUE);
   app_startup();
 
   init_parse_options_and_load_syslogformat(&parse_options);


bazsi:
@czanik can you please check these unit test problems on FreeBSD? I've tried to reproduce the issue but my FreeBSD skills are somewhat limited. Thanks

czanik:
@bazsi ```make check``` is not part of the FreeBSD ports Makefile, and I never compiled syslog-ng without ports. Now updating my build machine and trying to figure out what @yurivict did. It seems that he did a mixture of using ports and compiling by hand.

czanik:
@bazsi @yurivict  OK. Here is what did. In sysutils/syslog-ng first I did a ```make check-orphans```, the command I use to follow up with packaging list changes. Once it finished I changed to work/syslog-ng-3.38.1 and did a ```gmake check```. It failed, saying that unit tests are disabled. It turned out that ```criterion``` is now packaged for FreeBSD, so I installed it and repeated the process:
```
FAIL: modules/timestamp/tests/test_date
```
So, with 3.38.1 release a single test failed. As far as I can see, your current patch is related to time.

It seems that if I want to integrate ```make check``` support to the port, I'll have to edit the package list as well. But first I need to figure out if it can be made optional.

Before changing anything in the port, I check your fix on an up-to-date git snapshot.

bazsi:
Yeah, my change in this pr should fix test_date

czanik:
OK, test_date is fixed. However, there is now another fail:
```
FAIL: modules/python/tests/test_python_logmsg
```
Could it be related to --disable-python-modules, which is necessary not to look for Kubernetes Python modules?

bazsi:
Can you run this separately, outside of make check? And paste the output
here.

Also, it might be needed to log into that environment if possible.

czanik:
```
root@fb131:~/syslog-ng/work/syslog-ng-3.38.1.358.g3093f4e # ./modules/python/tests/test_python_logmsg
[----] modules/python/tests/test_python_logmsg.c:210: Assertion Failed
[----]   
[----]   The expression (as strings) (value) == (params->expected_log_msg_value) is false: actual=`1664887692.123` expected=`1664894892.123`.
[----]   
[FAIL] python_log_message::test_python_logmessage_set_value: (0.00s)
[====] Synthesis: Tested: 18 | Passing: 17 | Failing: 1 | Crashing: 0 
```

czanik:
Oops, wrong button... Reopening.

bazsi:
Thanks. This seems to be time related too. There are roughly 2 hours
between those timestamps, which is the amount between CET and GMT.

bazsi:
This should fix this:

```
diff --git a/modules/python/tests/test_python_logmsg.c b/modules/python/tests/test_python_logmsg.c
index 4f632faa3..dfc4b96a2 100644
--- a/modules/python/tests/test_python_logmsg.c
+++ b/modules/python/tests/test_python_logmsg.c
@@ -95,6 +95,7 @@ _construct_py_parse_options(void)
 void
 setup(void)
 {
+  setenv("TZ", "UTC", TRUE);
   app_startup();
 
   init_parse_options_and_load_syslogformat(&parse_options);

```

bazsi:
The crashes sometimes are expected, when the tests exercise paths that
should abort, but I'll check those.

For me a threaded testcase hung when I last tried it. the one that tests
wakeups. Isn't a test program running in the background when this is hung?

czanik:
```
root   2056   0.0  1.8 31460 18296  0  I+   11:49    0:00.65 gmake --no-print-directory check-TESTS
root   2063   0.0  0.3 13596  2960  0  I+   11:50    0:00.00 /bin/sh -c set +e; bases='libtest/tests/test_mock_transport.log lib/ack-tracker/tests/test_consecutive_ack_record_container.log lib/ack-tracker/tests/test_instant_ack_tracker.l
root   2069   0.0  1.7 26340 17316  0  I+   11:50    0:00.60 gmake --no-print-directory test-suite.log TEST_LOGS=libtest/tests/test_mock_transport.log lib/ack-tracker/tests/test_consecutive_ack_record_container.log lib/ack-tracker/tests/
root   2262   0.0  0.3 13596  2892  0  I+   11:52    0:00.00 /bin/sh ./test-driver ./lib/logproto/tests/test_logproto
root   2269   0.0  1.2 33232 12520  0  S    11:52    0:00.11 ./lib/logproto/tests/test_logproto
root  61855   0.0  1.4 26340 14152  0  I+   11:48    0:00.43 gmake check
root  61857   0.0  1.4 26340 14004  0  I+   11:48    0:00.45 gmake --no-print-directory check-recursive
root  61859   0.0  0.3 13596  2864  0  I+   11:48    0:00.00 /bin/sh -c fail=; \\\nif (target_option=k; case ${target_option-} in ?) ;; *) echo "am__make_running_with_option: internal error: invalid" "target option '${target_option-}' sp
root  61864   0.0  1.7 26340 17108  0  I+   11:48    0:00.52 gmake --no-print-directory check-am

```

czanik:
Well. Interesting. I had a couple of meetings, and I left ```gmake check``` running. It turned out that it not hung, only it's extremely slow. Normally running all tests finish in my vm in less than a minute. Now about 50 tests took more than an hour. Neither the host, nor the vm has a high load.

czanik:
It took a couple of hours, but both failing tests are green now.

MrAnno:
@czanik Is it always `./lib/logproto/tests/test_logproto` the one that goes into an interruptible sleep for a long time?

czanik:
@MrAnno  That's the first one. But after that many others as well. But just a week earlier everything was fast. So, I would not be surprised if everything worked again quickly after the next port upgrade.

czanik:
@bazsi @MrAnno Did a ports update on my FreeBSD build machine. Python and criterion dependencies got updated. Tests suddenly finish in less than a minute in my VM. With this patch all tests are green. :+1:

czanik (approved):
All tests are now green on FreeBSD

MrAnno (approved):

MrAnno:
I think we can merge this for v4.0.
I'm changing the milestone. Let me know if I overlooked something.

PR #4238, opened 2022-12-08
riemann: report flush errors with msg_error() instead of trace

bazsi:

### DIFF
--- a/modules/riemann/riemann-worker.c
+++ b/modules/riemann/riemann-worker.c
@@ -351,22 +351,29 @@ riemann_worker_flush(LogThreadedDestWorker *s, LogThreadedFlushMode mode)
       return LTR_ERROR;
     }
 
-  msg_trace("riemann: flushing messages to Riemann server",
-            evt_tag_str("server", owner->server),
-            evt_tag_int("port", owner->port),
-            evt_tag_int("batch_size", self->event.n),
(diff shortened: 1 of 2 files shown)


smortex (approved):
While the issue we try to diagnose does not reach this code, it still makes sense to log accordingly to the error status. LGTM!

modules/riemann/riemann-worker.c
```diff
@@ -348,25 +348,41 @@ riemann_worker_flush(LogThreadedDestWorker *s, LogThreadedFlushMode mode)
                                                  MAX(1, owner->super.batch_lines));
   if (!r)
     {
+      msg_error("riemann: error calling riemann_communicate()",
+                evt_tag_str("server", owner->server),
+                evt_tag_int("port", owner->port),
+                evt_tag_int("batch_size", self->event.n),
+                evt_tag_int("ok", r->ok),
+                evt_tag_str("error", r->error),
```
smortex, line 356: In this codepath, `r` is `NULL`, so we can't access `ok` / `error` here.

bazsi: thanks for the note, I've fixed this.

bazsi:
added news file, ready for review.

alltilla (approved):

PR #4248, opened 2022-12-19
dbld: remove support for Debian stretch

bazsi:
Debian stretch has Python 3.5, which limits the versions of PyPI packages as required by our requirements.txt and also causes a compatibility issue with our Python plugin.

Albeit we could support Python 3.5 in our Python plugin, this is the oldest version of Python that we have in our range of supported distros and the support period for Debian stretch is over, originally ending in June 2022.

So the easiest solution for us is to stop supporting it.


MrAnno:
This means we are dropping debian-stretch support from our APT repository as well.
In that case, please remove the debian-stretch line from the README too.

bazsi:
Will do so. Is this otherwise ok?

PR #4252, opened 2022-12-20
packaging/debian: add a syslog-ng-scl package

bazsi:
To mimic how syslog-ng is packaged on Debian, supply an "scl" package to contain all our SCL files.

This resolves an issue with updating from Debian installed syslog-ng to our own packages that caused problems in 4.0.0

PR #4253, opened 2022-12-21
.github/workflows/test-apt-packages: test against Debian's official packages

bazsi:
This is an emergency PR to help resolve our 4.0.0 release related issue. It adds coverage for smoke testing .deb packages against Debian's official packages, not just against our own, previous version.

The PR does not touch code, except a .github workflow and should cause the nightly to fail (which it should have previously as well).

Once nightly fails, I can resolve that using the packaging fix I have in #4252


kira-syslogng:
Build FAILURE

bazsi:
Merging without review (after a discussion with @alltilla) this as there's a stall in our CI workflows.

PR #4262, opened 2022-12-28
Fix mainloop call crash on freebsd

### LINKED ISSUE #4258: Occasional crash in main_loop_call_handler()
# syslog-ng
## Version of syslog-ng
3.38.1

## Platform
FreeBSD 14.0-CURRENT amd64

# Description

We are seeing syslog-ng crash occasionally (a couple dozen times per day across our fleet of thousands of systems), with symptoms that always look like this:

```
Program terminated with signal SIGABRT, Aborted.
Sent by thr_kill() from pid 700 and user 0.
#0  thr_kill () at thr_kill.S:4
4       thr_kill.S: No such file or directory.
[Current thread is 1 (LWP 100285)]
(gdb) bt
#0  thr_kill () at thr_kill.S:4
#1  0x00001e7cd28fb2c4 in __raise (s=s@entry=0x6) at /usr/src/lib/libc/gen/raise.c:52
#2  0x00001e7cd29ac199 in abort () at /usr/src/lib/libc/stdlib/abort.c:67
#3  0x00001e7ccdf838f5 in g_thread_abort (status=<optimized out>, function=0x1e7ccde6fdba "pthread_mutex_lock") at ../glib/gthread-posix.c:88
#4  0x00001e7ccdf838b4 in g_mutex_lock (mutex=mutex@entry=0x1e7cf21a0268) at ../glib/gthread-posix.c:225
#5  0x00001e7ccb3b0456 in main_loop_call_handler (user_data=<optimized out>) at lib/mainloop-call.c:115
#6  0x00001e7ccb3ff07e in __iv_event_run_pending_events (_st=<optimized out>) at iv_event.c:56
#7  0x00001e7ccb403db8 in iv_fd_kqueue_poll (st=<optimized out>, active=0x1e7cc9009190, abs=0x1e7cd81cc7b8) at iv_fd_kqueue.c:228
#8  0x00001e7ccb401182 in iv_fd_poll_and_run (st=st@entry=0x1e7cd813a000, abs=0x1e7cd81cc7b8) at iv_fd.c:206
#9  0x00001e7ccb402143 in iv_main () at iv_main_posix.c:112
#10 0x00001e7ccb3afda4 in main_loop_run (self=0x1e7ccb4168b8 <main_loop>) at lib/mainloop.c:677
#11 0x00001e74a87a994b in main (argc=<optimized out>, argv=0x1e7cc9009310) at syslog-ng/main.c:312
(gdb) frame 5
#5  0x00001e7ccb3b0456 in main_loop_call_handler (user_data=<optimized out>) at lib/mainloop-call.c:115
115     lib/mainloop-call.c: No such file or directory.
(gdb) p *site
$1 = {
  list = {
    next = 0x1e7cf21a0228,
    prev = 0x1e7cf21a0228
  },
  func = 0x1e7ccb3b23d0 <ml_batched_timer_perform_update>,
  user_data = 0x1e7cd82637c0,
  result = 0x0,
  pending = 0x1,
  wait = 0x0,
  cond = {
    p = 0x1e7cf4276008,
    i = {0x0, 0x0}
  },
  lock = {
    p = 0x1e7cf4276000,
    i = {0xf4276000, 0x1e7c}
  }
}

(gdb) p *(pthread_mutex_t)site->lock.p
$2 = {
  m_lock = {
    m_owner = 0x2,
    m_flags = 0x0,
    m_ceilings = {0x1, 0x0},
    m_rb_lnk = 0x762f3b3000435455,
    m_spare = {0x202020, 0x742f676f}
  },
  m_flags = 0x435455,
  m_count = 0x6c2e736e,
  m_spinloops = 0x202020,
  m_yieldloops = 0x67736d3b,
  m_ps = 0x435455,
  m_qe = {
    tqe_next = 0x7473640a00202020,
    tqe_prev = 0x6f643b6500435455
  },
  m_pqe = {
    tqe_next = 0x6f63646100202020,
    tqe_prev = 0x676f6c7400435455
  },
  m_rb_prev = 0x2f72617600202020
}
```

On FreeBSD, the mutex m_owner field being 2 means that the mutex has been destroyed, which is why the pthread_mutex_lock() call in g_mutex_lock() has failed.  Note that many other fields of this pthread_mutex_t structure contain ASCII strings, which suggests that the structure was probably freed and the memory reallocated for some other purpose. 

This mutex structure is attached to the per-thread TLS variable "call_info" in lib/mainloop-call.c, and it is allocated when a worker thread is created and freed when the worker thread is shutting down.  This TLS call_info structure is used by main_loop_call() to have the main thread of the process call a function and optionally have the worker thread wait for the main thread to complete this work.  In all of the crashes that we see, the function to be called in the main thread is ml_batched_timer_perform_update, and "wait" is false when this function is the one that is queued.  main_loop_call() has logic to wait for the main thread to be done with the current thread's call_info before using the call_info again, however, there is no corresponding logic to wait for the main thread to finish with a thread's call_info before the call_info is deinited in main_loop_call_thread_deinit() when a worker thread exits.

So I suspect that what is happening is that a worker thread is queuing an async main_loop_call() to ml_batched_timer_perform_update and then goes on to do other things, but then before the queued call is processed by the main thread, the worker thread shuts down, destroying the mutex in main_loop_call_thread_deinit().  When the main thread finishes the call to the queued function, it returns to main_loop_call_handler() and tries to re-lock the now-destroyed mutex.

If this scenario can really happen then a fix could look like:
```
diff -urp lib/mainloop-call.c.orig lib/mainloop-call.c
--- lib/mainloop-call.c.orig	2022-08-28 04:50:04.000000000 -0700
+++ lib/mainloop-call.c	2022-12-14 15:17:12.802711000 -0800
@@ -134,6 +134,13 @@ main_loop_call_thread_init(void)
 void
 main_loop_call_thread_deinit(void)
 {
+  g_mutex_lock(&main_task_lock);
+  if (&call_info.pending) {
+    call_info.wait = TRUE;
+    while (&call_info.pending)
+      g_cond_wait(&call_info.cond, &main_task_lock);
+  }
+  g_mutex_unlock(&main_task_lock);
   g_cond_clear(&call_info.cond);
   g_mutex_clear(&call_info.lock);
 }
```

bazsi:
The patches in this branch fixes #4258. 

Thanks a lot for @chucksilvers for the diagnosis and the proposed fix. At the end I wasn't using the proposed fix as it was, but I did included a Reported-By header. This fix would probably not exist today if it wasn't your hard work to reproduce and diagnose the issue.

### DIFF
--- a/lib/mainloop-call.c
+++ b/lib/mainloop-call.c
@@ -56,12 +56,9 @@ static GMutex main_task_lock;
 static struct iv_list_head main_task_queue = IV_LIST_HEAD_INIT(main_task_queue);
 static struct iv_event main_task_posted;
 
-gpointer
-main_loop_call(MainLoopTaskFunc func, gpointer user_data, gboolean wait)
+static void
+main_loop_wait_for_pending_call_to_finish(void)
 {
-  if (main_loop_is_main_thread())
-    return func(user_data);
-
   g_mutex_lock(&main_task_lock);
 
   /* check if a previous call is being executed */
@@ -79,6 +76,16 @@ main_loop_call(MainLoopTaskFunc func, gpointer user_data, gboolean wait)
     {
       g_mutex_unlock(&call_info.lock);
     }
+  g_mutex_unlock(&main_task_lock);
+}
+
+gpointer
+main_loop_call(MainLoopTaskFunc func, gpointer user_data, gboolean wait)
+{
+  if (main_loop_is_main_thread())
+    return func(user_data);
+
+  main_loop_wait_for_pending_call_to_finish();
 
   /* call_info.lock is no longer needed, since we're the only ones using call_info now */
   INIT_IV_LIST_HEAD(&call_info.list);
@@ -86,6 +93,7 @@ main_loop_call(MainLoopTaskFunc func, gpointer user_data, gboolean wait)
   call_info.func = func;
   call_info.user_data = user_data;
   call_info.wait = wait;
+  g_mutex_lock(&main_task_lock);
   iv_list_add(&call_info.list, &main_task_queue);
   iv_event_post(&main_task_posted);
   if (wait)
--- a/lib/mainloop-call.c
+++ b/lib/mainloop-call.c
@@ -142,6 +142,8 @@ main_loop_call_thread_init(void)
 void
 main_loop_call_thread_deinit(void)
 {
+  main_loop_wait_for_pending_call_to_finish();
+
   g_cond_clear(&call_info.cond);
   g_mutex_clear(&call_info.lock);
 }


kira-syslogng:
Build FAILURE

bazsi:
Btw I reproduced the issue by inserting a sleep(15) into a mainloop_call handler, to trigger a worker exit while the call is executing.

The one I used was the ml_batched_timer()'s update function.

MrAnno (approved):

PR #4270, opened 2023-01-05
ci: add missing safedir setup for commit and copyright check

MrAnno:
Sorry, I misread these two.

### DIFF
--- a/.github/workflows/publish-docker-image.yml
+++ b/.github/workflows/publish-docker-image.yml
@@ -45,7 +45,7 @@ jobs:
         id: tags
         run: |
           if [[ '${{ inputs.pkg-type }}' = 'stable' ]]; then
-            TAGS='${{ steps.stable-tags.outputs.tag }}'
+            TAGS='${{ steps.stable-tags.outputs.tags }}'
           elif [[ '${{ inputs.pkg-type }}' = 'nightly' ]]; then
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build FAILURE

.github/workflows/publish-docker-image.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         id: tags
         run: |
           if [[ '${{ inputs.pkg-type }}' = 'stable' ]]; then
-            TAGS='${{ steps.stable-tags.outputs.tag }}'
+            TAGS='${{ steps.stable-tags.outputs.tags }}'
```
bazsi, line 48: Ah, I think we will definitely need some kind of mechanism for testing the "stable" pipeline before doing the actual release. I don't see how else we would be catching these kind of errors.

bazsi (approved):

MrAnno:
This fixes false-negative CI commit style check results. @HofiOne @bazsi @therandomstring Feel free to merge it.

.github/workflows/publish-docker-image.yml
MrAnno: Yeah, we previously tested these jobs manually after each change.
We could do some release CI tests using a test APT repo/a test docker namespace, etc.

PR #4276, opened 2023-01-09
Secret-storage no unlock fix

therandomstring:
A small bugfix concerning proper shutdown.
the bug concerns the mutex guarding the initialisation of secret-storage, where the mutex has been locked, but no unlock is present.

### DIFF
--- a/lib/secret-storage/secret-storage.c
+++ b/lib/secret-storage/secret-storage.c
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2023 One Identity LLC
  * Copyright (c) 2018 Balabit
  *
  * This library is free software; you can redistribute it and/or
@@ -79,6 +80,8 @@ void
 secret_storage_deinit(void)
 {
   g_assert(!secret_manager_uninitialized);
+  g_atomic_int_inc(&secret_manager_uninitialized);
+  g_assert(secret_manager_uninitialized == 1);
   g_hash_table_destroy(secret_manager);
   secret_manager = NULL;
 }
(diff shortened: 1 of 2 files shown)


MrAnno (commented):
What do you intend to fix with this commit?
Mutexes and atomic counters have different purposes, guarantees and costs.
Using an atomic counter here is plenty enough in my opinion, and serves its purpose.

lib/gprocess.c
```diff
@@ -1543,6 +1544,7 @@ g_process_finish(void)
   */
 #endif
 
+  free(process_opts.argv_orig);
```
bazsi, line 1547: There's a comment right above this new line that explains why we don't want to free that line. I am more than happy to prove that this comment is already stale and we can actually perform this free operation, but please remove the comment along with it.

therandomstring:
@MrAnno this fixes an issue while testing with libFuzzer, it hanging on the second run, as it tries to lock the mutex/counter not unlocked at the end of the previous test.
As per your request, I will revert back to an atomic counter.

lib/gprocess.c
therandomstring: Said comment is the reason, why I only freed this variable, and not `**environ`. However the fact that somehow I placed the free outside the if clause is proof enough that I should have drank at least two more cups of coffee yesterday.
I would be happy to investigate this leak further, but given the circumstances, I am not confident enough to still include this in this merge request, as the other fix is instrumental to the fuzzing project, so I do not want to delay that one.
Sorry for the blunder.

kira-syslogng:
Build FAILURE

MrAnno (approved):

PR #4280, opened 2023-01-12
tlscontext: add ignore-unexpected-eof option for TLS

MrAnno:
> Some TLS implementations do not send the mandatory close_notify alert on
> shutdown. If the application tries to wait for the close_notify alert but
> the peer closes the connection without sending it, an error is generated.

Error message:
> error:0A000126:SSL routines::unexpected eof while reading

This option disables above error messages but enables the possibility of undetected truncation attacks.

Do not use this option unless you are actively hit by this error message, AND you use an acknowledged protocol type that prevents/detects truncation, AND you have no control over fixing the other side that does not send close_notify as it should.

In any other cases, update/fix the other party that does not send close_notify properly.

### DIFF
--- a/lib/transport/tls-context.c
+++ b/lib/transport/tls-context.c
@@ -219,6 +219,13 @@ tls_context_setup_ssl_options(TLSContext *self)
       if(self->ssl_options & TSO_NOTLSv13)
         ssl_options |= SSL_OP_NO_TLSv1_3;
 #endif
+
+#ifdef SSL_OP_IGNORE_UNEXPECTED_EOF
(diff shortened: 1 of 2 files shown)


MrAnno:
NEWS file is missing intentionally.

kira-syslogng:
Build FAILURE

bazsi (approved):
LGTM, but the mac build fails for some, probably unrelated, reason.

kira-syslogng:
Build FAILURE

MrAnno:
@kira-syslogng test this please test branch=OSE-4.0-testdb-fix;

MrAnno:

bazsi (approved):

PR #4281, opened 2023-01-15
add sdata-parser() & accept unquoted RFC5424 SD-PARAM-VALUE

bazsi:
This allows someone to incrementally parse a malformed RFC5424 message, where the SDATA portion is not at the right spot. The SDATA format is relatively complex, so extracting it as a separate component makes it easier to process this kind of data.

This is WIP, because:
- [x]  no specific tests
- [x] I want to make "prefix" customizable and I am not finished with that yet
- [x] NEWS file

### DIFF
--- a/modules/syslogformat/syslog-format.c
+++ b/modules/syslogformat/syslog-format.c
@@ -693,6 +693,23 @@ _syslog_format_parse_sd(LogMessage *msg, const guchar **data, gint *length, cons
                   else
                     goto error;
                 }
+              else if (left)
+                {
+                  pos = 0;
+
+                  while (left && (*src != ' ' && *src != ']'))
+                    {
+                      if (pos < sizeof(sd_param_value) - 1)
+                        {
+                          sd_param_value[pos] = *src;
+                          pos++;
+                        }
+                      _skip_char(&src, &left);
+                    }
+                  sd_param_value[pos] = 0;
+                  sd_param_value_len = pos;
+
+                }
               else
                 {
                   goto error;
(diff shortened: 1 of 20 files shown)


kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

modules/syslogformat/syslog-format.c
```diff
@@ -720,6 +720,10 @@ log_msg_parse_sd(LogMessage *self, const guchar **data, gint *length, const MsgF
         }
       while (left && open_sd != 0);
     }
+  else
+    {
+      goto error;
+    }
```
MrAnno, line 744: This will break the parsing of RFC 5424 messages that have no SDATA fields and forgot to mark it with a NILVALUE (`-`).

This would not be a valid RFC 5424, so I guess it is okay to fall back to legacy parsing mode.

MrAnno (approved):

modules/syslogformat/syslog-format.c
bazsi: you are right, this might have been intentional. I removing this from this PR.

bazsi: removed the functional change.

kira-syslogng:
Build FAILURE

modules/syslogformat/syslog-format.c
bazsi: I have now readded the stricter check when being invoked from sdata-parser() but at the same time relaxed it when doing SDATA parsing from RFC5424. This will match historical behaviour, as well as be strict when invoking it from sdata-parser() itself.

Also it makes the unit tests pass again. Let me know what you think.

MrAnno: It is perfect. Thank you.

MrAnno (approved):

PR #4282, opened 2023-01-16
Fix worker warning

MrAnno:
Reproduction steps:
`--worker-threads=1` and a network source with an active connection


HofiOne (approved):
Fine for me

PR #4288, opened 2023-01-24
ci: fix multiline output issue with tags

MrAnno:
`gh_output` can not export multi-line output, and it is not trivial to do that safely:
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings

(see the warning about script injection).

### DIFF
--- a/.github/workflows/publish-docker-image.yml
+++ b/.github/workflows/publish-docker-image.yml
@@ -40,6 +40,7 @@ jobs:
         with:
           images: ${{ env.DOCKER_IMAGE_NAME }}
           tags: type=match,pattern=syslog-ng-(.*),group=1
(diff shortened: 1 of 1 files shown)


bazsi:
Is the intent to use ',' instead of newlines to avoid multi-line output in TAGS? I mean this is only possible if the tag in question contains multiple matches, right? But even then we have "group=1" in the match option, so as far as I can see it can only have one value, can't it?

MrAnno:
`tags: type=match,pattern=syslog-ng-(.*),group=1` generates 2 tags: one with the version number, and it outputs the "latest" tag as well.

bazsi (approved):

PR #4293, opened 2023-01-27
Preliminary gdb support to light

bazsi:
This patch allows you to run syslog-ng under gdb as it is being exercised by the light test framework.

```
    To use gdb to launch syslog-ng, you will need the following (in the
    container if you are running this in dbld)
      * xterm
      * access to your X server (-v /tmp/.X11-unix/:/tmp/.X11-unix -e DISPLAY=$$DISPLAY to docker)
    
    You can use it by supplying --run-under=gdb to light.
    
    syslog-ng will start up under gdb while the tests execute, gdb will
    be opened in an xterm window.
```

To test, you will need to ```./dbld/rules root-login``` into your container and apt-get install xterm. You also need a couple of options specified to docker as it runs your container, I added those into my ```./dbld/rules.conf```

### DIFF
--- a/tests/light/src/syslog_ng/syslog_ng_executor.py
+++ b/tests/light/src/syslog_ng/syslog_ng_executor.py
@@ -20,6 +20,8 @@
 # COPYING for details.
 #
 #############################################################################
+import shlex
+
 from src.common.random_id import get_unique_id
 from src.executors.command_executor import CommandExecutor
 from src.executors.process_executor import ProcessExecutor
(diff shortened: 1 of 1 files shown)


bazsi:
This proved very useful, again. So I rebased the patch and documented how to use it.

Removing "draft" state.

bazsi:
I think this should be fine being merged, I am using this occassionally and since this is only changing light, this should be safe to merge even if not perfect.

MrAnno (approved):

Issue #4287, opened 2023-01-24
Redeclaration errors with older gcc versions


MrAnno:
#4245 introduced some "redeclaration as different kind of symbol" errors with older gcc versions.

I'm thinking on a fix of something like this:

```
void
load_journald_mock(void)
{
  sd_journal_open = &mock_sd_journal_open;
  ...
}
```

But we also need non-pointer function declarations for the `SYSLOG_NG_SYSTEMD_JOURNAL_MODE == SYSLOG_NG_JOURNALD_SYSTEM` case.

```
 CC       modules/systemd-journal/tests/test_systemd_journal-test-source.o
/syslog-ng-core/modules/systemd-journal/tests/journald-mock.c:138:1: error: ‘sd_journal_open’ redeclared as different kind of symbol
 sd_journal_open(sd_journal **s, int flags)
 ^
In file included from /syslog-ng-core/modules/systemd-journal/tests/journald-mock.h:27:0,
                 from /syslog-ng-core/modules/systemd-journal/tests/journald-mock.c:24:
/syslog-ng-core/modules/systemd-journal/journald-subsystem.h:68:14: note: previous declaration of ‘sd_journal_open’ was here
 extern int (*sd_journal_open)(sd_journal **ret, int flags);
              ^
/syslog-ng-core/modules/systemd-journal/tests/journald-mock.c:153:1: error: ‘sd_journal_close’ redeclared as different kind of symbol
 sd_journal_close(sd_journal *s)
 ^
In file included from /syslog-ng-core/modules/systemd-journal/tests/journald-mock.h:27:0,
                 from /syslog-ng-core/modules/systemd-journal/tests/journald-mock.c:24:
/syslog-ng-core/modules/systemd-journal/journald-subsystem.h:72:15: note: previous declaration of ‘sd_journal_close’ was here
 extern void (*sd_journal_close)(sd_journal *j);
               ^
/syslog-ng-core/modules/systemd-journal/tests/journald-mock.c:166:1: error: ‘sd_journal_seek_head’ redeclared as different kind of symbol
 sd_journal_seek_head(sd_journal *s)
 ^
In file included from /syslog-ng-core/modules/systemd-journal/tests/journald-mock.h:27:0,
                 from /syslog-ng-core/modules/systemd-journal/tests/journald-mock.c:24:
/syslog-ng-core/modules/systemd-journal/journald-subsystem.h:73:14: note: previous declaration of ‘sd_journal_seek_head’ was here
[...]

PR #4297, opened 2023-01-31
Redeclaration-errors-with-older-gcc-versions-4287

HofiOne:
This PR fixes the broken mock part of journald integration tests when --with-systemd-journal=optional configure option is used

Also added  some builds genereated leftover to the .gitignore

fixes: #4287

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,5 @@ TAGS
 GPATH
 GRTAGS
 GTAGS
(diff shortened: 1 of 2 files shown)


bazsi (approved):
This looks great. Thanks @HofiOne

bazsi:
The devshell CI issue should be fixed in #4298

PR #4303, opened 2023-02-01
Apache rawrequest escaping binary characters

### LINKED ISSUE #4274: Regression in `rawrequest` extraction with Apache SCL
# syslog-ng
## Version of syslog-ng
4.0.1

## Platform
Debian

## Debug bundle
n/a

# Issue
## Failure
When an Apache HTTP VirtualHost receive an HTTPS connection, it returns an 400 response code and logs something like this:

```
main:80 152.89.196.211 "-" - [23/Dec/2022:19:04:19 +0100] "\x16\x03\x01" 400 226 "-" "-"
```

The raw request `\x16\x03\x01` is extracted by apache SCL and available as `prefix.rawrequest`.

We updated from syslog-ng-3.38.1-1 to syslog-ng-4.0.1-1 on 2022-12-23. I just realize today that we have some surprising values in our rawrequests:

* On syslog-ng-3.38.1-1, rawrequest was `\x16\x03\x01`. Result of searching for "\\x16\\x03\\x01" in OpenSearch Dashboards:
  Result of searching for "\\x16\\x03\\x01" in OpenSearch Dashboards
* On syslog-ng-4.0.1-1, rawrequest is `x16x03x01`. Result of searching for "x16x03x01" in OpenSearch Dashboards:
  Result of searching for "x16x03x01" in OpenSearch Dashboards

The `\` have been stripped-out and shouldn't. This looks like a regression in syslog-ng-4.0.

This is likely related to #4228 (but I have not confirmed it, I am writing this issue so it does not get lost and might start to have a look when the weekend is over).

## Steps to reproduce
Setup an Apache HTTP server and try to access it using HTTPS.  Parse the log with apache SCL.

## Configuration

```
source s_apache_access {
    wildcard-file(
        base_dir(
            /var/log/apache2
        ),
        filename_pattern(
            *access.log
        ),
        flags(
            no-parse
        ),
        program_override(
            "apache"
        )
    );
};

parser p_apache {
    apache-accesslog-parser(
        prefix(
            "apache."
        )
    );
};

log {
    source(s_apache_access);
    source(s_apache_access_ssl);
    parser(p_apache);
    parser(p_apache_geoip);
    rewrite(r_apache_geoip);
    rewrite(r_apache_bytes);
    destination(d_riemann);
};

```

n/a

bazsi:
This branch fixes #4274 by:

1) making sure that apache style escaping in access.logs are parsed properly, even if it contains escaped binary characters
2) on output uses \xXX escaping style for binary characters, instead of \uXXXX, which shouldn't be needed for control characters anyway.

This might still be a change compared to our 3.38 behaviour, I guess in that case these binary characters were considered a single field and then re-escaped as it was sent out in JSON to ElasticSearch.

So it was probably sent to Elastic as: ```"\\x16\\x03\\x01"``` Note the double backslashes.

This patch will make that use binary, e.g. elastic would receive the original: ```"\x16\x03\x01"``` string, which it would de-escape probably. So this is still a change compared to 3.38 but I think this behaviour is better than our previous one, as we are not able to process the binary characters properly.

### DIFF
--- a/lib/scanner/csv-scanner/csv-scanner.c
+++ b/lib/scanner/csv-scanner/csv-scanner.c
@@ -176,6 +176,30 @@ _parse_left_whitespace(CSVScanner *self)
   _skip_whitespace(&self->src);
 }
 
+static gint
+_decode_xdigit(gchar xdigit)
+{
+  if (xdigit >= '0' && xdigit <= '9')
+    return xdigit - '0';
+  if (xdigit >= 'a' && xdigit <= 'f')
+    return xdigit - 'a' + 10;
+  if (xdigit >= 'A' && xdigit <= 'F')
+    return xdigit - 'A' + 10;
+  return -1;
+}
+
+static gint
+_decode_xbyte(gchar xdigit1, gchar xdigit2)
+{
+  gint nibble_hi, nibble_lo;
+
+  nibble_hi = _decode_xdigit(xdigit1);
+  nibble_lo = _decode_xdigit(xdigit2);
+  if (nibble_hi < 0 || nibble_lo < 0)
+    return -1;
+  return (nibble_hi << 4) + nibble_lo;
+}
+
 static void
 _parse_character_with_quotation(CSVScanner *self)
 {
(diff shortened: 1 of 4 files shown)


kira-syslogng:
Build FAILURE

MrAnno (approved):
We'll need a news entry for this.

PR #4304, opened 2023-02-02
systemd-journal: fix --with-systemd-journal=optional mode

HofiOne:
Further modifications in the lookup of sd_id128_XXX symbols (that can be presented in separate modules on some OSes e.g. Centos)

### DIFF
--- a/modules/systemd-journal/journald-subsystem.c
+++ b/modules/systemd-journal/journald-subsystem.c
@@ -22,6 +22,8 @@
  */
 
 #include "journald-subsystem.h"
+#include "messages.h"
+
 #include <glib.h>
 #include <gmodule.h>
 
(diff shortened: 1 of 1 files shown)


modules/systemd-journal/journald-subsystem.c
```diff
@@ -57,20 +57,25 @@ int (*sd_id128_get_boot)(sd_id128_t *ret);
 
 #define JOURNAL_LIBRARY_NAME "libsystemd-journal.so.0"
 #define SYSTEMD_LIBRARY_NAME "libsystemd.so.0"
+#define ID128_LIBRARY_NAME   "libsystemd-id128.so"
 
 const char *journald_libraries[] = {JOURNAL_LIBRARY_NAME, SYSTEMD_LIBRARY_NAME, NULL};
+const char *journald_ext_libraries[] = {ID128_LIBRARY_NAME, JOURNAL_LIBRARY_NAME, SYSTEMD_LIBRARY_NAME, NULL};
 
 static GModule *journald_module;
+static GModule *journald_ext_module;
```
bazsi, line 66: we could probably move these to local variables in load_journald_subsystem() and then pass them along
using arguments to _load_journald_subsystem()

bazsi (approved):
I am fine with this as is, I added a small improvement that would have made reviewing this easier but that's optional too.

With that change, this unit would be drastically more readable than it has been a month ago :)

Thanks

modules/systemd-journal/journald-subsystem.c
HofiOne: yeah, it came to my mind as well, the only reason I left it this way is that once we might want to add a really proper cleanup then we might want to call unload_journald_subsystem() and that would not be that nice to pass these to that function too

HofiOne:
@bazsi Sorry, I've made a typo in libsystemd-id128.so.0, corrected, and added some logging of the lookup process and possible errors. Hope it's fine now.

HofiOne:
Centos tests are passed, seems we are fine now

HofiOne:

HofiOne:
@bazsi Could you please take a look at this one more (and hopefully the last time) again?
Thanks!

bazsi (approved):

PR #4306, opened 2023-02-02
fixed to possible use ip source as table name sql, and add timestamps…

143981:
… for events.

### DIFF
--- a/modules/afsql/afsql.c
+++ b/modules/afsql/afsql.c
@@ -557,7 +557,7 @@ _table_create(AFSqlDestDriver *self, const gchar *table)
       return FALSE;
     }
 
-  g_string_printf(query_string, "CREATE TABLE %s (", table);
+  g_string_printf(query_string, "CREATE TABLE `%s` (`time` timestamp NOT NULL DEFAULT current_timestamp(), ", table);
   for (i = 0; i < self->fields_len; i++)
     {
       g_string_append_printf(query_string, "%s %s", self->fields[i].name, self->fields[i].type);
@@ -872,7 +872,7 @@ afsql_dd_build_insert_command(AFSqlDestDriver *self, LogMessage *msg, GString *t
   GString *value = g_string_sized_new(512);
   gint i, j;
 
-  g_string_printf(insert_command, "INSERT INTO %s (", table->str);
+  g_string_printf(insert_command, "INSERT INTO `%s` (", table->str);
 
   for (i = 0; i < self->fields_len; i++)
     {
--- a/modules/afsql/afsql.c
+++ b/modules/afsql/afsql.c
@@ -557,7 +557,7 @@ _table_create(AFSqlDestDriver *self, const gchar *table)
       return FALSE;
     }
 
-  g_string_printf(query_string, "CREATE TABLE `%s` (`time` timestamp NOT NULL DEFAULT current_timestamp(), ", table);
+  g_string_printf(query_string, "CREATE TABLE `%s` (", table);
   for (i = 0; i < self->fields_len; i++)
     {
       g_string_append_printf(query_string, "%s %s", self->fields[i].name, self->fields[i].type);
--- a/modules/afsql/afsql.c
+++ b/modules/afsql/afsql.c
@@ -406,15 +406,15 @@ afsql_dd_create_index(AFSqlDestDriver *self, const gchar *table, const gchar *co
 
           format_hex_string(hash, md_len, hash_str, sizeof(hash_str));
           hash_str[0] = 'i';
-          g_string_printf(query_string, "CREATE INDEX %s ON %s (%s)",
+          g_string_printf(query_string, "CREATE INDEX %s ON `%s` (%s)",
                           hash_str, table, column);
         }
       else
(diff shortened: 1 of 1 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

bazsi:
Ok to test

modules/afsql/afsql.c
```diff
@@ -557,7 +557,7 @@ _table_create(AFSqlDestDriver *self, const gchar *table)
       return FALSE;
     }
 
-  g_string_printf(query_string, "CREATE TABLE %s (", table);
+  g_string_printf(query_string, "CREATE TABLE `%s` (`time` timestamp NOT NULL DEFAULT current_timestamp(), ", table);
```
bazsi, line 560: I think this could be added via configuration. You can specify default values to a field.

bazsi (commented):
Apart from the time field which can be added as config this looks good.

alltilla:
I have a vague memory that the quoting might differ in some SQL server kinds (looking at you oracle -_-). If that is the case, we need to do this in a more sophisticated way. There are E2E tests which do not run on these PRs, and they do test the different SQL destinations. They have to be started manually. @HofiOne Sorry for flooding you, but could you kindly help with this also, when you have the time? Thanks! <3

HofiOne:
> I have a vague memory that the quoting might differ in some SQL server kinds (looking at you oracle -_-). If that is the case, we need to do this in a more sophisticated way. There are E2E tests which do not run on these PRs, and they do test the different SQL destinations. They have to be started manually. @HofiOne Sorry for flooding you, but could you kindly help with this also, when you have the time? Thanks! <3

yepp, of course, the main problem now is that we can E2E test only specific branches as the master is still fails, after the last missing fix is merged (#4304) tests will run automatically and we will see the result immediately
the same is true if you wish to pre-test (before merging) these, I can run the tests based on these branches, but first we need a base that will not break the test earlier, far before the test reaches these cases

HofiOne:
@alltilla  Though builds are fine now we have a bunch of issues accessing any 3rd party service, mainly because of outdated credentials, expired licenses, tokens, and accounts, DBO is working on it, I'll keep you updated as there's any progress on this.

alltilla:
> @alltilla Though builds are fine now we have a bunch of issues accessing any 3rd party service, mainly because of outdated credentials, expired licenses, tokens, and accounts, DBO is working on it, I'll keep you updated as there's any progress on this.

Thank you, I really appreciate that! :)

bazsi:
According to this: https://stackoverflow.com/questions/10573922/what-does-the-sql-standard-say-about-usage-of-backtick

using backticks is a MySQLism. Unfortunately libdbi does not offer such quoting functionality.

if you could extract the quoting of identifiers into a separate function, which does it via backticks if the driver is mysql and nothing otherwise, I think this could be merged. Otherwise, this would just break all the other drivers.

143981:
maybe it would be easier so that it could be written in the config? with 

> table("`${HOST}`")

with this config - table name is empty...

bazsi:
The backtick character is used by the syslog-ng language as well. I am not
sure theres a way to escape it at the moment.

bazsi:
> The backtick character is used by the syslog-ng language as well. I am not sure theres a way to escape it at the moment.

I've just checked, a double backtick would include a single one.

So this should work:

```
table("``${HOST}``")
```

143981:
> > The backtick character is used by the syslog-ng language as well. I am not sure theres a way to escape it at the moment.
> 
> I've just checked, a double backtick would include a single one.
> 
> So this should work:
> 
> ```
> table("``${HOST}``")
> ```

` version for the right syntax to use near \'.8.164_ `
`query='CREATE TABLE _10.16.8.164_ (time datet....`

no, it do not work..

bazsi:
> > > The backtick character is used by the syslog-ng language as well. I am not sure theres a way to escape it at the moment.
> > 
> > 
> > I've just checked, a double backtick would include a single one.
> > So this should work:
> > ```
> > table("``${HOST}``")
> > ```
> 
> `version for the right syntax to use near \'.8.164_` `query='CREATE TABLE _10.16.8.164_ (time datet....`
> 
> no, it do not work..

the backticks do reach the SQL destination, but then the SQL destination removes invalid characters from the name of
the table and therefore does not allow the backtick character.

Everything is invalid that is not in this set:

```
static gboolean _sql_identifier_is_valid_char(gchar c)
{
  return ((c == '.') ||
          (c == '_') ||
          (c >= '0' && c <= '9') ||
          (g_ascii_tolower(c) >= 'a' && g_ascii_tolower(c) <= 'z'));
}
```

So the only way to do this is via explicit escaping from the SQL module, as you implemented it. Or by replacing the '.' characters with something else before reaching the SQL destination. (e.g. subst() right before hitting the sql() destination)

bazsi:
Are you willing to add the backtick style escaping when formatting the table name?

If you extract the escaping to a function that checks if we are running on MySQL and only performs it there, I'd be willing to merge that.

HofiOne:
Implemented in https://github.com/syslog-ng/syslog-ng/pull/4460

PR #4307, opened 2023-02-03
github-actions/macos: temporarily disable amqp

alltilla:
rabbitmq-c 0.12.0 broke the pc file version, which breaks our minimum version check. Once it is fixed, we should reenable amqp.

https://github.com/alanxz/rabbitmq-c/issues/755

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -44,6 +44,10 @@ jobs:
             --with-python=3
           "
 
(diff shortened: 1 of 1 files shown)


alltilla:
@HofiOne This should fix the red macOS CI. :)

HofiOne (approved):
Upps, I was blind, it was not the MongoDB of course :)
Thanks!

PR #4308, opened 2023-02-03
diskq: `disk-buf-size()` related fixes: `dqtool cat`, `example-diskq-source()`, changing `disk-buf-size()`

alltilla:
There is a bug in `dqtool` which was caused by 2 things:
  1. We changed the condition of the cursor wrapping from file size to `disk-buf-size()`.
  2. We do not store the `disk-buf-size()` in the disk-queue header.

There is no way to tell where we should wrap without knowing the `disk-buf-size()` before loading the disk-queue.

With this patch set we now store the initial `disk-buf-size()` in the header and use that for the wrap condition.

Also we can now detect whether the user tries to change the `disk-buf-size()` of an existing disk-queue file. For now we will reject that change and go with the original `disk-buf-size()`, which is a bugfix itself, as before this patch set we did not realize that with a config change and might have tried to overread/overwrite the disk-queue file.

We can implement the `disk-buf-size()` change in the future in a safe way, however it is not the scope of this patch set.

### DIFF
--- a/modules/diskq/diskq.c
+++ b/modules/diskq/diskq.c
@@ -227,14 +227,7 @@ _attach(LogDriverPlugin *s, LogDriver *d)
   LogDestDriver *dd = (LogDestDriver *) d;
   GlobalConfig *cfg = log_pipe_get_config(&d->super);
 
-  if (self->options.disk_buf_size == -1)
-    {
-      msg_error("The required 'disk_buf_size()' parameter of diskq module has not been set.",
-                log_pipe_location_tag(&dd->super.super));
-      return FALSE;
-    }
-
-  if (self->options.disk_buf_size < MIN_DISK_BUF_SIZE && self->options.disk_buf_size != 0)
+  if (self->options.disk_buf_size < MIN_DISK_BUF_SIZE && self->options.disk_buf_size > 0)
     {
       msg_warning("The value of 'disk_buf_size()' is too low, setting to the smallest acceptable value",
                   evt_tag_long("min_space", MIN_DISK_BUF_SIZE),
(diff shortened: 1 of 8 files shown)


alltilla:
@HofiOne If I remember correctly, there is a disk-buffer file upgrade E2E test in testdb, which stores what syslog-ng versions bumped the disk-buffer header version. I think we could/should run that test on this PR and extend the list when a new PE release comes out. Could you kindly help with this?

kira-syslogng:
Build FAILURE

alltilla:
Kira fails with 2 disk-buffer related tests. They seem to be relevant to this change. I am checking those.

bazsi:
I like the patches, those look good.

I did some testing, "dqtool info' displays:

```
$ /install/bin/dqtool info syslog-ng-00000.qf 
Disk-buffer state loaded; filename='syslog-ng-00000.qf', number_of_messages='3436'
WARNING: disk-buf-size() has changed since the last syslog-ng run. syslog-ng currently does not support changing the disk-buf-size() of existing disk-queues. Continuing with the old one; filename='syslog-ng-00000.qf', active_old_disk_buf_size='1048576', ignored_new_disk_buf_size='0'
```

The queue file was created with the current syslog-ng version, so it should have the size in the header. Cat seems to work fine.

alltilla:
@bazsi Nice catch! Fixed.

HofiOne:
>

HofiOne:
> So good news, the disk-buffer E2E test case is running in Kira, and it caught an error, which I managed to fix, now it is green. Still, when we have time, we should add some new variants to it as a new disk-buffer header version was introduced.

Sorry for the accidental Close :S

@alltilla do we still need the nightly random platform test for this?

alltilla:
@Hofi No problem :) If I remember correctly every disk-buffer testdb tests are run in KIRA, too, so there is no need to run PE testdb tests. Stress ( == perf) testing would be nice, but I can see that it cannot run currently :(

bazsi (approved):

PR #4319, opened 2023-02-10
add value() parameter to date-parser()

bazsi:
This PR improves 3 things to deliver more complete support for processing non-standard timestamps.

1) type-hinting: improve the format we can use for DATETIME values by allowing usec granularity (instead of msec) and adding timezone offset information.

2) date-parser now allows extracting timestamps into a name-value pair (in addition to storing the extracted time in the message timestamps)

3) $(format-date) now uses the improved datetime format, so the timezone offset is properly handled when formatting the timestamp.

### DIFF
--- a/lib/logmsg/type-hinting.c
+++ b/lib/logmsg/type-hinting.c
@@ -139,48 +139,57 @@ type_cast_to_double(const gchar *value, gdouble *out, GError **error)
   return success;
 }
 
+static gboolean
+_parse_fixed_point_timestamp_in_nsec(const gchar *value, gchar **endptr, gint64 *sec, gint64 *nsec)
+{
+  *sec = (gint64) strtoll(value, endptr, 10);
+
+  if (**endptr == 0)
+    {
+      *nsec = 0;
+      return TRUE;
+    }
+
+  if (**endptr == '.')
+    {
+      const gchar *nsec_start = (*endptr) + 1;
+
+      *nsec = (gint64) strtoll(nsec_start, endptr, 10);
+      gint nsec_length = (*endptr) - nsec_start;
+      if (nsec_length == 0)
+        return FALSE;
+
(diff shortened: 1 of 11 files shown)


lib/logmsg/type-hinting.c
```diff
@@ -166,22 +168,28 @@ _parse_fixed_point_timestamp_in_nsec(const gchar *value, gchar **endptr, gint64
...
+  gint tzofs = -1;
 
   if (!_parse_fixed_point_timestamp_in_nsec(value, &endptr, &sec, &nsec))
     goto error;
 
-  if (*endptr != 0)
+  const guchar *tz_start = (guchar *) endptr;
+  gint tz_length = strlen(endptr);
+  if (*tz_start != 0 &&
+      !scan_iso_timezone(&tz_start, &tz_length, &tzofs))
```
alltilla, line 187: This adds an unnecessary `strlen()` call, when timezone info is not present. I think we should check `*tz_start != 0` first.

lib/logmsg/type-hinting.c
```diff
@@ -150,6 +151,7 @@ _parse_fixed_point_timestamp_in_nsec(const gchar *value, gchar **endptr, gint64
       return TRUE;
     }
 
+  *nsec = 0;
```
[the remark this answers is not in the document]
bazsi: good point. This was changed from one commit to the next, but eventually this become unneeded, so I've removed this.

lib/logmsg/type-hinting.c
bazsi: reorganized, the strlen() is only done if *tz_start != 0

thanks.

alltilla (approved):

alltilla:
The macOS fail is not related to this PR, it fails on master, too.

PR #4320, opened 2023-02-12
Fix the Python based frontend for the interactive debugger

bazsi:
This branch fixes a couple of bugs that rendered the interactive debugger inoperable. The problems were introduced when we did Python type support.


alltilla (approved):
Sorry for breaking this :(

PR #4321, opened 2023-02-12
logmsg: remove LF_SIMPLE_HOSTNAME

bazsi:
This flag was unused, even if the functionality has merit, it should have been passed to LogSource through options and not via LogMessage flags.

The flag itself will not survive clones, so not really useful anyway.

@HofiOne this one is probably used by one of your internal modules. An easier alternative to achieve the same is to set LogSourceOptions->chain_hostnames to FALSE. No need to use a LogMessage flag for this purpose.

No NEWS needed.

PS: I am trying to add another LogMessage flag in an alternative branch, that's when I encountered this one.

### DIFF
--- a/lib/logmsg/logmsg.h
+++ b/lib/logmsg/logmsg.h
@@ -123,8 +123,6 @@ enum
   LF_STATE_OWN_SDATA   = 0x0100,
   LF_STATE_OWN_MASK    = 0x01F0,
 
-  /* In the log header the hostname shall be printed individually (no group name, no chain hosts)*/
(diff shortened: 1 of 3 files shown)


HofiOne:
> @HofiOne this one is probably used by one of your internal modules. An easier alternative to achieve the same is to set LogSourceOptions->chain_hostnames to FALSE. No need to use a LogMessage flag for this purpose.

Thanks for signaling this, I'll take a look at later on!

MrAnno (approved):

PR #4324, opened 2023-02-13
cfg-lexer refactor and cleanup

bazsi:
This branch is a preparation for a major overhaul of our "debugger" functionality. As a prep to that I cleaned up the CfgLexer code
and extracted CfgIncludeLevel to be more of a separate class, instead of just being a data structure manipulated in a complicated
way.

The outcome of the branch is that cfg_lexer_start_next_include() became a few lines of code.

No NEWS entry is needed.

### DIFF
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -291,6 +291,14 @@ cfg_lexer_alloc_include_level(CfgLexer *self)
   return &self->include_stack[self->include_depth];
 }
 
+void
+cfg_lexer_drop_include_level(CfgLexer *self, CfgIncludeLevel *level)
+{
+  g_assert(&self->include_stack[self->include_depth] == level);
+  cfg_lexer_clear_include_level(self, level);
+  self->include_depth--;
+}
+
 gboolean
 cfg_lexer_start_next_include(CfgLexer *self)
 {
@@ -427,7 +435,8 @@ cfg_lexer_include_file_simple(CfgLexer *self, const gchar *filename)
                     evt_tag_str("filename", filename),
                     evt_tag_str("error", error->message));
           g_error_free(error);
-          goto drop_level;
+          cfg_lexer_drop_include_level(self, level);
+          return FALSE;
         }
       while ((entry = g_dir_read_name(dir)))
         {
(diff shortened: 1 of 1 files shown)


bazsi:
I just did a patch-by-patch "make check" test. I did some edits, rerunning again.

lib/cfg-lexer.c
```diff
@@ -450,29 +450,35 @@ cfg_lexer_start_next_include(CfgLexer *self)
...
+      if ((level->include_type == CFGI_BUFFER) ||
+          (level->include_type == CFGI_FILE && !level->file.files))
+        level_finished = TRUE;
+
       cfg_lexer_include_level_close_buffer(self, level);
     }
 
-  if ((level->include_type == CFGI_BUFFER && buffer_processed) ||
-      (level->include_type == CFGI_FILE && !level->file.files))
+  if (level_finished)
```
alltilla, line 461: Before this commit, the `if`'s condition was true with `!level->yybuf` and `level->include_type == CFGI_FILE && !level->file.files`, now with `!level->yybuf` we always go to the else branch. Is this intended? Is this a problem?

alltilla (commented):
I only had one minor comment, otherwise LGTM.

lib/cfg-lexer.c
bazsi: A CfgIncludeLevel struct describes an include operation that can involve including a

* Single file
* buffer
* Or a list of files (e.g. wildcard or directory include).

This function is called at these points:
* When the include operation is not processed yet, we are just going ahead and doing the first include. In this case yybuf is NULL
* When we just finished an include operation and we need to switch to the next file (yybuf non NULL and the files list has items to be processed
* When we finished an include operation and there's nothing more in the files list.

The if branch here decides whether we are finished with the current include level. But this can only happen if yybuf is non NULL, otherwise we are yet to process the very first include.

So I think the old, relatively complex conditions can be simplified, and that's what I did here.

alltilla: Thanks for the explanation :)

alltilla (approved):

PR #4325, opened 2023-02-13
New metric system (Prometheus data format)

MrAnno:
This PR adds a new metric naming scheme (labeling capability) to syslog-ng and a querying tool to produce Prometheus-like metric exports.

```
$ sbin/syslog-ng-ctl stats prometheus

syslogng_events_allocated_bytes 18446744073709550560
syslogng_filtered_events_total{id="#anon-filter0",result="matched"} 0
syslogng_filtered_events_total{id="#anon-filter0",result="not_matched"} 2
syslogng_filtered_events_total{id="ff",result="matched"} 0
syslogng_filtered_events_total{id="ff",result="not_matched"} 2
syslogng_input_events_total{id="#anon-source0#0",driver_instance="-",result="processed"} 0
syslogng_input_events_total{id="s_network#1",result="processed"} 2
syslogng_internal_source{result="dropped"} 0
syslogng_internal_source{result="queued"} 0
syslogng_output_events_total{id="d_dest#0",driver_instance="tcp,127.0.0.1:5555",result="delivered"} 0
syslogng_output_events_total{id="d_dest#0",driver_instance="tcp,127.0.0.1:5555",result="dropped"} 0
syslogng_output_events_total{id="d_dest#0",driver_instance="tcp,127.0.0.1:5555",result="queued"} 0
syslogng_output_events_total{id="d_dest#1",driver_instance="http,https://localhost:8080",result="delivered"} 0
syslogng_output_events_total{id="d_dest#1",driver_instance="http,https://localhost:8080",result="dropped"} 0
syslogng_output_events_total{id="d_dest#1",driver_instance="http,https://localhost:8080",result="queued"} 0
syslogng_parsed_events_total{id="#anon-parser0",result="discarded"} 0
syslogng_parsed_events_total{id="#anon-parser0",result="processed"} 0
syslogng_scratch_buffers_bytes 0
syslogng_scratch_buffers_total 3
syslogng_tagged_events_total{id=".source.#anon-source0",result="processed"} 0
syslogng_tagged_events_total{id=".source.s_network",result="processed"} 2

$ sbin/syslog-ng-ctl stats prometheus --with-legacy-metrics

syslogng_src_facility_processed{stat_instance="18"} 0
syslogng_dst_network_memory_usage{id="d_dest#0",stat_instance="tcp,127.0.0.1:5555"} 0
syslogng_src_host_processed{id="s_network#1",stat_instance="AnnoW"} 2
syslogng_global_sdata_updates_processed 0
syslogng_global_msg_clones_processed 0
syslogng_global_internal_queue_length_processed 0
syslogng_global_payload_reallocs_processed 4
...
```

syslog-ng stats labels are similar to the labels in the Prometheus data model:
https://prometheus.io/docs/concepts/data_model/

The (id, labels) pair identifies a new-style key. The order of the labels
should not matter, and a label with an empty value should be equivalent to
a label non-existing, but the current implementation is rudimentary and
does not take these considerations into account.

The old (component, instance , id) keying is marked as `legacy`; any of the legacy functions should not be used when introducing new metrics.

Questions:
- when should we remove `stats-log`?

Related PR: #4318


lib/stats/tests/test_stats_cluster.c
```diff
@@ -163,6 +163,61 @@ Test(stats_cluster, test_stats_cluster_key_equal_when_custom_tags_are_the_same)
...
+  stats_cluster_logpipe_key_set(&sc_key2, "id", labels2, G_N_ELEMENTS(labels2));
+  sc2 = stats_cluster_new(&sc_key2);
+
+  cr_assert_not(stats_cluster_key_equal(&sc_key1, &sc_key2), "%s", __FUNCTION__);
+
+  stats_cluster_free(sc2);
+  stats_cluster_free(sc1);
+}
+
+Test(stats_cluster, test_stats_cluster_key)
```
alltilla, line 195: I was thinking about the comparisons. If I understand correctly there are these variables:
 1. Keys cannot be NULLs, so they can be:
     a. same (non-NULL, non-NULL)
     b. different (non-NULL, non-NULL)
 2. Labels can be NULLs, so they can be:
     a. same (NULL, NULL),
     b. same (non-NULL, non-NULL)
     c. different (NULL, non-NULL)
     d. different (non-NULL, non-NULL)

From these, we can create 2x4=8 test scenarios. I have made a table for them with examples (comparing A and B):
| 1. keys      | key-A | key-B | 2. labels                         | label-A | label-B | result |
| ------------ | ----- | ----- | --------------------------------- | ------- | ------- | ------ |
| a. same      | key-1 | key-1 | a. same (NULL, NULL)              | NULL    | NULL    | ==     |
| a. same      | key-1 | key-1 | b. same (non-NULL, non-NULL)      | label-1 | label-1 | ==     |
| a. same      | key-1 | key-1 | c. different (NULL, non-NULL)     | NULL    | label-1 | !=     |
| a. same      | key-1 | key-1 | d. different (non-NULL, non-NULL) | label-1 | label-2 | !=     |
| b. different | key-1 | key-2 | a. same (NULL, NULL)              | NULL    | NULL    | !=     |
| b. different | key-1 | key-2 | b. same (non-NULL, non-NULL)      | label-1 | label-1 | !=     |
| b. different | key-1 | key-2 | c. different (NULL, non-NULL)     | NULL    | label-1 | !=     |
| b. different | key-1 | key-2 | d. different (non-NULL, non-NULL) | label-1 | label-2 | !=     |

Some of them are not tested here, and I think they should be.

lib/stats/tests/test_stats_cluster.c
```diff
@@ -163,6 +163,61 @@ Test(stats_cluster, test_stats_cluster_key_equal_when_custom_tags_are_the_same)
   stats_cluster_free(sc2);
 }
 
+Test(stats_cluster, test_stats_cluster_key_equal_with_labels)
```
alltilla, line 166: I don't see much difference between this test case and `test_stats_cluster_key`, other than creating `StatsCluster`s here.

I think they could be merged, and we could create `StatsCluster`s in the merged TC.

Also I think we should check the keys stored in the `StatsCluster`s, if we are creating them.

alltilla (commented):
Reviewed until `stats: add labeling capability`, inclusive. It looks good, I only had UT related comments.

I will check the last 2 commits soon.

kira-syslogng:
Build FAILURE

lib/stats/stats-cluster.c
```diff
@@ -78,11 +78,69 @@ stats_cluster_deinit(void)
...
+
 StatsClusterKey *
 stats_cluster_key_clone(StatsClusterKey *dst, const StatsClusterKey *src)
 {
-  dst->legacy.component = src->legacy.component;
   dst->id = g_strdup(src->id ? : "");
+  dst->labels = stats_cluster_key_labels_clone(src->labels, src->labels_len);
+  dst->labels_len = src->labels_len;
+  dst->legacy.component = src->legacy.component;
   dst->legacy.instance = g_strdup(src->legacy.instance ? : "");
```
alltilla, line 146: This `strdup()` bothers me. It does not produce the same value after clone (`NULL` vs `""`).

Previously this was not an issue, because in `stats_cluster_single_key_set()`, if the user has given `NULL` it was replaced to `""`, which means that `src->instance` and `src->id` was never `NULL`, making this check unnecessary. 

However, now there are cases, where `legacy.instance` and later `legacy.id` are `NULL`, so this `strdup()` produces a different result, which can mess up the `equal()` function, or at least this has to be accounted for in the `equal()` function.

lib/stats/stats-cluster.c
```diff
@@ -261,22 +289,29 @@ stats_counter_group_init_equals(const StatsCounterGroupInit *self, const StatsCo
...
 {
-  return key1->legacy.component == key2->legacy.component
-         && strcmp(key1->id, key2->id) == 0
-         && strcmp(key1->legacy.instance, key2->legacy.instance) == 0
-         && stats_counter_group_init_equals(&key1->counter_group_init, &key2->counter_group_init)
-         && stats_cluster_key_labels_equal(key1->labels, key1->labels_len, key2->labels, key2->labels_len);
+  if (stats_cluster_key_is_legacy(key1) != stats_cluster_key_is_legacy(key2))
+    return FALSE;
+
+  if (stats_cluster_key_is_legacy(key1) && stats_cluster_key_is_legacy(key2))
```
alltilla, line 295: [optional]

At this point we know, that the two `is_legacy()` are the same, so it is enough to check one of them.

lib/stats/stats-cluster.c
```diff
@@ -261,22 +289,29 @@ stats_counter_group_init_equals(const StatsCounterGroupInit *self, const StatsCo
...
+    {
+      return key1->legacy.component == key2->legacy.component
+             && strcmp(key1->legacy.id, key2->legacy.id) == 0
+             && strcmp(key1->legacy.instance, key2->legacy.instance) == 0
+             && stats_counter_group_init_equals(&key1->counter_group_init, &key2->counter_group_init);
+    }
+
+  return strcmp(key1->name, key2->name) == 0
+         && stats_cluster_key_labels_equal(key1->labels, key1->labels_len, key2->labels, key2->labels_len)
+         && stats_counter_group_init_equals(&key1->counter_group_init, &key2->counter_group_init);
```
alltilla, line 305: This function raised some questions for me.

It seems like the implementation only accounts for cleanly new-style and cleanly legacy-style keys. By this I mean if both of them are legacy, only the legacy parts are compared, if they are both new-style, only the new-style values are checked.

If I understand correctly there is a third case, which is new-style with legacy alias, where all of the fields should be checked.

I wonder what should happen, if two keys have the same new-style values, but different legacy aliases or vice versa. Should we even allow that? If not, how should we implement that, so it does not get really slow?

If I understand correctly, currently if there are two keys with different new-style values and the same alias, they get the same cluster with the new-style values of the first one to register. I don't think this is okay.

MrAnno: I think it will be easier to remove the legacy code this way. We should not use different new-style keys with the same legacy alias.

lib/stats/stats-cluster.c
alltilla: I think the current implementation works. There are 2 scenarios:
 1. legacy alias was not set
     * the original key has `NULL` as `legacy.instance`
     * the cloned key gets `""` as `legacy.instance`
     * both `legacy.set` is 0
     * because of the previous point the `hash()` and `equal()` functions do not compare the legacy values, so it works
 2. legacy alias was set
     * original key has a non-NULL `legacy.instance`, because `_legacy_set()` forces it to be non-NULL.
     * the cloned key will copy the `legacy.instance`, everything works as expected

In other words the legacy values are only compared if `legacy.set` is set to 1, and when it is set to 1 (`_legacy_set()`) the problematic values are always non-NULL.

lib/stats/stats-cluster.c
alltilla: Agreed.

lib/logthrdest/logthrdestdrv.c
```diff
@@ -966,13 +967,26 @@ log_threaded_dest_driver_queue(LogPipe *s, LogMessage *msg,
...
+  enum { labels_len = 1 };
+  static StatsClusterLabel labels[labels_len];
+
+  labels[0] = stats_cluster_label("id", self->super.super.id ? : "");
+  stats_cluster_logpipe_key_set(sc_key, "output_events_total", labels, labels_len);
+  stats_cluster_logpipe_key_add_legacy_alias(sc_key, self->stats_source | SCS_DESTINATION,
+                                             self->super.super.id,
+                                             self->format_stats_instance(self));
+}
+
```
alltilla, line 989: `_init_stats_key()` now provides the new-style "output_event_total" counter. It is passed to `log_queue_register_stats_counters()`:
```c
static void
_register_worker_stats(LogThreadedDestWorker *self)
{
  StatsClusterKey sc_key;

  stats_lock();
  _init_stats_key(self->owner, &sc_key);
  log_queue_register_stats_counters(self->queue, 0, &sc_key);
  stats_unlock();
}
```
It registers the common counters and also adds place for each queue implementation to register their own counters:
```c
void
log_queue_register_stats_counters(LogQueue *self, gint stats_level, const StatsClusterKey *sc_key)
{
  _register_common_counters(self, stats_level, sc_key);

  if (self->register_stats_counters)
    self->register_stats_counters(self, stats_level, sc_key);
}
```
Although it is not used by any of our queue implementation, but if it was, I think we should not always register the custom counters under the new `"output_event_total"` cluster.

I think we should pass both the new and the legacy keys, so the queue implementation can decide which one to use, or we should remove the possibility of registering custom counters altogether.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

lib/stats/tests/test_stats_cluster.c
MrAnno: Done, thanks for the table.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

MrAnno:
@kira-syslogng do stresstest

MrAnno:
@HofiOne Can you help us fix the stress test?

15:53:49 # An exception occured:
15:53:49 #   Type: IOError
15:53:49 #   Message: [Errno 28] No space left on device

kira-syslogng:

folti:
@kira-syslogng do stresstest

lib/scratch-buffers.c
```diff
@@ -299,9 +299,11 @@ scratch_buffers_register_stats(void)
   StatsClusterKey sc_key;
 
   stats_lock();
-  stats_cluster_logpipe_key_legacy_set(&sc_key, SCS_GLOBAL, "scratch_buffers_count", NULL);
+  stats_cluster_logpipe_key_set(&sc_key, "scratch_buffers_count", NULL, 0);
+  stats_cluster_logpipe_key_add_legacy_alias(&sc_key, SCS_GLOBAL, "scratch_buffers_count", NULL);
   stats_register_counter(0, &sc_key, SC_TYPE_QUEUED, &stats_scratch_buffers_count);
-  stats_cluster_logpipe_key_legacy_set(&sc_key, SCS_GLOBAL, "scratch_buffers_bytes", NULL);
+  stats_cluster_logpipe_key_set(&sc_key, "scratch_buffers_bytes", NULL, 0);
+  stats_cluster_logpipe_key_add_legacy_alias(&sc_key, SCS_GLOBAL, "scratch_buffers_bytes", NULL);
```
alltilla, line 306: [optional]

`queued` seems odd here.

Adding `stats_cluster_single_key_add_legacy_alias_with_name()` and converting this to a single counter would probably solve this issue (see my poc patch), but as this is an experimental feature and scratch buffers counters are not top prio ATM, I think it is okay like this for now.

[scratch-buffers.diff.txt](https://github.com/syslog-ng/syslog-ng/files/10783456/scratch-buffers.diff.txt)

lib/afinter.c
```diff
@@ -582,7 +582,8 @@ afinter_message_posted(LogMessage *msg)
 
       stats_lock();
       StatsClusterKey sc_key;
-      stats_cluster_logpipe_key_legacy_set(&sc_key, SCS_GLOBAL, "internal_source", NULL );
+      stats_cluster_logpipe_key_set(&sc_key, "internal_source", NULL, 0);
+      stats_cluster_logpipe_key_add_legacy_alias(&sc_key, SCS_GLOBAL, "internal_source", NULL );
```
alltilla, line 586: [optional]

It would be nice, if the internal source counters could be `internal_events_total` with a sensible sum. We would need a `written` or something analogous counter to make it work, and as this is an experimental feature and internal source is not top prio ATM, I think we can keep it like this for now.

kira-syslogng:

folti:
@kira-syslogng do stresstest

kira-syslogng:

folti:
@kira-syslogng do stresstest

kira-syslogng:

lib/logwriter.c
```diff
@@ -1444,8 +1444,14 @@ _register_counters(LogWriter *self)
...
-  stats_cluster_logpipe_key_legacy_set(&sc_key, self->options->stats_source | SCS_DESTINATION, self->stats_id,
-                                       self->stats_instance);
+  StatsClusterLabel labels[] =
+  {
+    stats_cluster_label("id", self->stats_id),
+    stats_cluster_label("driver_instance", self->stats_instance),
+  };
+  stats_cluster_logpipe_key_set(&sc_key, "output_events_total", labels, G_N_ELEMENTS(labels));
+  stats_cluster_logpipe_key_add_legacy_alias(&sc_key, self->options->stats_source | SCS_DESTINATION, self->stats_id,
+                                             self->stats_instance);
```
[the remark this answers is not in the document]
alltilla: Actually `processed` does not get increased when we restart with a disk-buffer, but `queued` does. I think this makes it more appealing to use `dropped + queued + written`.

alltilla:
@folti Thank you, the stress test can now run. The env is working, it fails because of syslog-ng related reasons.

@MrAnno Both failures are caused by the `python` destination:
```
    [2023-02-20T13:32:58.083703] python: private virtualenv is not initialized, use the `syslog-ng-update-virtualenv' script to initialize it or make sure all required Python dependencies are available in the system Python installation; path='/mnt/jenkins/workspace/kira-performance-test-different-destinations/syslog-ng-ose/install/var/python-venv'
    [2023-02-20T13:32:58.094302] WARNING: With use-dns(no), dns-cache() will be forced to 'no' too!;
    [2023-02-20T13:32:58.094599] Accepting connections; addr='AF_INET(0.0.0.0:61001)'
    [2023-02-20T13:32:58.094655] python-dest: Error initializing Python destination, class is not a subclass of LogDestination; driver='d_python_313ad2b2840e402f91ab7eff68f2052e#0', class='MyPythonDestination', class-repr='<class \'_syslogng_main.MyPythonDestination\'>'
    [2023-02-20T13:32:58.094661] Error initializing message pipeline; plugin_name='python', location='/root/var/testdb_working_dir0/etc/syslog-ng.conf.local:16:5'
```

So they are not related to this PR. I think we can ignore them for this PR and they should be fixed separately.

HofiOne:
> @folti Thank you, the stress test can now run. The env is working, it fails because of syslog-ng related reasons.
> 
> @MrAnno Both failures are caused by the `python` destination:
> 
> ```
>     [2023-02-20T13:32:58.083703] python: private virtualenv is not initialized, use the `syslog-ng-update-virtualenv' script to initialize it or make sure all required Python dependencies are available in the system Python installation; path='/mnt/jenkins/workspace/kira-performance-test-different-destinations/syslog-ng-ose/install/var/python-venv'
>     [2023-02-20T13:32:58.094302] WARNING: With use-dns(no), dns-cache() will be forced to 'no' too!;
>     [2023-02-20T13:32:58.094599] Accepting connections; addr='AF_INET(0.0.0.0:61001)'
>     [2023-02-20T13:32:58.094655] python-dest: Error initializing Python destination, class is not a subclass of LogDestination; driver='d_python_313ad2b2840e402f91ab7eff68f2052e#0', class='MyPythonDestination', class-repr='<class \'_syslogng_main.MyPythonDestination\'>'
>     [2023-02-20T13:32:58.094661] Error initializing message pipeline; plugin_name='python', location='/root/var/testdb_working_dir0/etc/syslog-ng.conf.local:16:5'
> ```
> 
> So they are not related to this PR. I think we can ignore them for this PR and they should be fixed separately.

I'm trying to fix the Python part and will drop you a line if is's done.

HofiOne:
@kira-syslogng do stresstest

kira-syslogng:

HofiOne:
@kira-syslogng do stresstest

kira-syslogng:

HofiOne:
The python issues related to the usage of the wrong python test base-classes are fixed now, but we still have some other seems to be also a python, and older issue.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

HofiOne:
@kira-syslogng do stresstest

HofiOne:
Seems the stress test is fine now, I've started a final re-test, but is should be fine now.

kira-syslogng:

lib/logwriter.c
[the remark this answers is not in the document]
MrAnno: Good point. I've removed the processed counter from the new metric output.

lib/scratch-buffers.c
MrAnno: Thank you very much for the patch.

lib/logthrdest/logthrdestdrv.c
MrAnno: I removed this abstract function. From the LogQueue's perspective, this could be a useful interface, but I don't think we will use it anytime soon.

lib/afinter.c
alltilla: We can do this later.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

lib/logthrdest/logthrdestdrv.c
```diff
@@ -1056,9 +1056,12 @@ _register_stats(LogThreadedDestDriver *self)
 
     _init_stats_key(self, &sc_key);
     stats_register_counter(0, &sc_key, SC_TYPE_DROPPED, &self->dropped_messages);
-    stats_register_counter(0, &sc_key, SC_TYPE_PROCESSED, &self->processed_messages);
     stats_register_counter(0, &sc_key, SC_TYPE_WRITTEN, &self->written_messages);
 
+    stats_cluster_single_key_legacy_set_with_name(&sc_key, self->stats_source | SCS_DESTINATION, self->super.super.id,
```
alltilla, line 1061: This results in a `StatsClusterKey`, which has a logpipe new-style value and a single legacy value, which is really odd for me, and don't know how should/would behave. Also with this we have two counters, which share the new-style values, but differ in the legacy values, which we do not support. Could you kindly create a separate single legacy key for this, like in `logwriter.c`?

lib/stats/stats-cluster-logpipe.c
```diff
@@ -51,7 +51,12 @@ _counter_group_logpipe_get_type_label(StatsCounterGroup *self, gint type, StatsC
   if (type >= SC_TYPE_MAX)
     return FALSE;
 
-  *label = stats_cluster_label("result", tag_names[type]);
+  const gchar *type_name = tag_names[type];
+
+  if (type == SC_TYPE_WRITTEN)
+    type_name = "delivered";
+
+  *label = stats_cluster_label("result", type_name);
```
alltilla, line 58: [optional]

I think it would be better to grab this from the other side, and rename `SC_TYPE_WRITTEN` and its corresponding format string, driver level `StatsCounterItem` names, etc, as this implementation suggests, that the legacy naming is the real one and the new-style has a little workaround for it. In other words, this implementation makes it harder to remove the legacy stats.

lib/logthrdest/logthrdestdrv.c
MrAnno: All key setter functions sets the full key by zero-initializing it. `stats_cluster_*_key_add_*()` functions can be used to extend a previous set call.

alltilla: Oh yes, this is not an legacy alias adding call, which does not zero init. Thanks for the explanation.

lib/stats/stats-cluster-logpipe.c
MrAnno: You're right. I'd like to fix this separately. This PR became huge.

alltilla: Sure!

alltilla (dismissed):

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

alltilla (approved):

PR #4329, opened 2023-02-15
Improve debugger source display

bazsi:
This PR builds on the cfg-lexer refactor and will improve the debugger source display dramatically as we become able to
track code that is inside blocks properly.

It also adds the ability to generate syslog-ng configuration using external tools and mark generated code with "@line" to refer to the original source file, in a similar vein to how C/C++ tracks header files, etc.

### DIFF
--- a/lib/cfg-lexer.c
+++ b/lib/cfg-lexer.c
@@ -160,7 +160,7 @@ _find_closest_file_inclusion(CfgLexer *self, CFG_LTYPE *yylloc)
 }
 
 const gchar *
-cfg_lexer_format_location(CfgLexer *self, CFG_LTYPE *yylloc, gchar *buf, gsize buf_len)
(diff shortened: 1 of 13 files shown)


alltilla:
Rebased to master.

lib/cfg-lexer.c
```diff
@@ -329,7 +306,7 @@ cfg_lexer_init_include_level_buffer(CfgLexer *self, CfgIncludeLevel *level,
                                     gsize length)
 {
   level->include_type = CFGI_BUFFER;
-  level->name = g_strdup(name);
+  level->lloc.name = g_intern_string(name);
```
MrAnno, line 320: I don't fully understand this patch.

Why do we use intern strings? It populates a global hash table. Do we have some kind of a lifetime problem here?

MrAnno (approved):

PR #4330, opened 2023-02-15
Preliminary tracing support in debugger

bazsi:
This is more a proof of concept that improves the interactive debugger of syslog-ng (-i). There are actually two changes in here:

1) the debugger will only stop at user-specified LogPipe instances (and not all of them, which is cumbersome to trace through)
2) a new "trace" command that will display tracing/timing information on how a message traverses the configuration
3) a new "info pipe" command

These three could actually be split. The first two patches implement the 1), the two subsequent patch implement 2) and the new info pipe is the last patch. I just didn't want to bomb you would a lot of PRs. But I can split this easily.

NEWS is still missing, let me see if it's green.

### DIFF
--- a/lib/debugger/debugger.c
+++ b/lib/debugger/debugger.c
@@ -128,6 +128,7 @@ _cmd_help(Debugger *self, gint argc, gchar *argv[])
 {
   printf("syslog-ng interactive console, the following commands are available\n\n"
          "  help, h, or ?            Display this help\n"
+         "  info                     Display information about the current execution state\n"
          "  continue or c            Continue until the next breakpoint\n"
          "  trace                    Display timing information as the message traverses the config\n"
          "  print, p                 Print the current log message\n"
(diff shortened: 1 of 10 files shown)


lib/debugger/debugger.c
```diff
@@ -350,6 +361,22 @@ debugger_stop_at_breakpoint(Debugger *self, LogPipe *pipe_, LogMessage *msg)
...
+{
+  struct timespec ts, *prev_ts = &self->last_trace_event;
+  gchar buf[1024];
+
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+
+  long diff = prev_ts->tv_sec == 0 ? 0 : timespec_diff_nsec(&ts, prev_ts);
+  printf("[%"G_GINT64_FORMAT".%09"G_GINT64_FORMAT" +%ld] Tracing %s\n",
+         ts.tv_sec, ts.tv_nsec, diff,
+         log_expr_node_format_location(pipe_->expr_node, buf, sizeof(buf)));
```
alltilla, line 404: [optional]

I think seeing the source line, like with breakpoints, would be great here as well:
```
  printf("[%"G_GINT64_FORMAT".%09"G_GINT64_FORMAT" +%ld] Tracing %s\n",
         ts.tv_sec, ts.tv_nsec, diff,
         log_expr_node_format_location(pipe_->expr_node, buf, sizeof(buf)));
  _display_source_line(pipe_->expr_node);
```

lib/debugger/debugger.c
```diff
@@ -201,6 +202,32 @@ _cmd_quit(Debugger *self, gint argc, gchar *argv[])
...
+  printf("LogPipe %p at %s\n", pipe, log_expr_node_format_location(pipe->expr_node, buf, sizeof(buf)));
+  _display_source_line(pipe->expr_node);
+
+  return TRUE;
+}
+
+static gboolean
+_cmd_info(Debugger *self, gint argc, gchar *argv[])
+{
+  if (argc && strcmp(argv[1], "pipe") == 0)
```
alltilla, line 219: `argc` should be checked to be 2. The current implementation segfaults.

bazsi: nice catch. thanks.

lib/debugger/debugger.c
bazsi: this is just a preliminary stuff, I intend to improve this part a lot later (like random sampling, and backtraces, etc).

for now this is not very useful, because it displays timing information for a single message, this can be affected by jitter a lot, so not very reliable.

alltilla (approved):

PR #4333, opened 2023-02-16
Fix memory leak in timeutils/cache (#4332)

meeusr:
Each time a worker thread terminates, the 2 allocated tznames are leaked. 
Solution is to explicitly deinit the cache from the stop function of the thread.

### DIFF
--- a/lib/timeutils/cache.h
+++ b/lib/timeutils/cache.h
@@ -38,6 +38,7 @@ time_t cached_mktime(struct tm *tm);
 void cached_localtime(time_t *when, struct tm *tm);
 void cached_gmtime(time_t *when, struct tm *tm);
 
+void timeutils_cache_deinit();
 
 static inline void
 cached_localtime_wct(time_t *when, WallClockTime *wct)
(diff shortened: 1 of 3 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

bazsi:
Ok to test.

lib/timeutils/cache.h
```diff
@@ -38,6 +38,7 @@ time_t cached_mktime(struct tm *tm);
 void cached_localtime(time_t *when, struct tm *tm);
 void cached_gmtime(time_t *when, struct tm *tm);
 
+void timeutils_cache_deinit();
```
bazsi, line 41: If you add void as the argument list, the ci issue gets fixed.

bazsi:
Closing this PR in favour of the new one #4334

Would have been enough to force push this branch. But as I can see the other pr fixes the problem I have indicated here.

PR #4337, opened 2023-02-20
msg-stats: add stats() block & add syslog-stats() global stats() option

alltilla:
### Global `stats()` options block

We have a significant number of stats related global options. I added another one, so I felt like it is time to group these options together.

This change breaks the behavior of the `old stats()` global option, which was an alias for `stats-freq()`, but it has been deprecated for more than a decade now, I think it is safe to remove it now.

The old `stats-...()` options are still supported for compatibility reasons, but new options will be introduced inside the `stats()` block.

---

### `stats(syslog-stats())`

It changes the behavior of counting messages based on different syslog-proto fields, like `SEVERITY`, `FACILITY`, `HOST`, etc...

Possible values are:
  * `yes` => force enable
  * `no` => force disable
  * `auto` => let `stats-level()` decide (old behavior)


kira-syslogng:
Build FAILURE

alltilla:

MrAnno (approved):

PR #4339, opened 2023-02-23
logmpx: fix matched logic

alltilla:
We can have a multiplexer with both `pipe_next` and `next_hops`.

The `path_options`'s `matched` field must only be set to false if and only if every `next_hop` and also the `pipe_next` failed to deliver. If at least one `next_hop` or at least the `pipe_next` delivered it, it must not be set to false.

With the old logic, if the multiplexer had a `pipe_next`, the `next_hops`' delivery result was ignored completely, which is not the correct behavior.

### DIFF
--- a/lib/logmpx.c
+++ b/lib/logmpx.c
@@ -113,25 +113,12 @@ log_multiplexer_queue(LogPipe *s, LogMessage *msg, const LogPathOptions *path_op
         }
     }
 
-  /* NOTE: non of our multiplexed destinations delivered this message, let's
-   * propagate this result.  But only if we don't have a "next".  If we do,
-   * that would be responsible for doing the same, for instance if it is a
-   * filter.
-   *
-   * In case where this matters most (e.g.  the multiplexer attached to the
-   * source LogPipe), "next" will always be NULL.  I am not sure if there's
-   * ever a case, where a LogMpx has both "next" set, and have branches as
-   * well.
-   *
-   * If there's such a case, then from a conceptual point of view, this Mpx
-   * instance should transfer the responsibility for setting "matched" to
-   * the next pipeline element, which is what we do here.
-   */
+  log_pipe_forward_msg(s, msg, &local_options);
+  if (matched)
+    delivered = TRUE;
 
-  if (!s->pipe_next && !delivered && path_options->matched)
+  if (!delivered && path_options->matched)
     *path_options->matched = FALSE;
-
-  log_pipe_forward_msg(s, msg, path_options);
 }
 
 static void


alltilla:
Let me show you a minimal config, which reproduces the behavior, which I do not agree with.

We support the following syntax:
```
@version: 4.0
@include "scl.conf"

log {
    source { file("/dev/stdin"); };
    destination {file ("/dev/null"); };

    log {
        filter { 1 == 2; };
    };
};
```

This generates the following tree:
tmp

`/dev/null` is connected to the next hop, while `/dev/stdout` is connected to the `pipe_next`.

In this case the message gets delivered to `/dev/null`, so the mpx should propagate `matched = TRUE`. However, as the mpx has a `pipe_next`, it completely forgets its internal `delivered` bool, and passes the `path_options` to the direction of `/dev/stdout`, which does not match, because of the always false filter, and sets `matched = FALSE`, so the mpx propagates `matched = FALSE`.

If we remove the internal log path, we get this:
 ```
@version: 4.0
@include "scl.conf"

log {
    source { file("/dev/stdin"); };
    destination {file ("/dev/null"); };
};
```
tmp2

Similarly, `/dev/null` is connected to a `next_hop`, and it delivers the message, so the mpx does not set `matced = FALSE`, it keeps it `TRUE`. Here we do not have a `pipe_next` at all, so the non-existent `pipe_next` does not set `matched = FALSE`, and the mpx propagates `match = TRUE`.

I think adding an extra internal log path, which does not match, should not hide the fact that the message was delivered to `/dev/null`, so I think this is a bug.

**Comment by bazsi** [comment]
I agree this is a bug, but I think #4058 is a better solution. I am not sure I understand the entire problem but a pretty similar issue
was reported by @faxm0dem couple of months ago and I started to fix it in #4058 but I abandoned it a bit.

It took me a while to understand what I have done 6 months ago (NOTE to self, never ever abandon a cfg-tree related change before it is finished), but I have now rebased and cleaned up the branch and hopefully all the tests are now green (including the one I added in that branch that intends to exercise the different cases of conditional routing)

With that patch and your configuration I get this LogPipe structure.

q

With that patch we section junctions  as if they were a single element in a sequence, if the junction as a whole drops the message (on all its branches). 

I'd love to come back with a testcase from your example config to see that fix addresses your problem too, but I think it does.

**Comment by bazsi** [comment]
I have now improved #4058 a bit further and gave it a lot of thought, how we should handle mid-point destinations vs. filtering.

This is what I have came up with:
* a log path consists of filtering and non-filtering elements
* a filtering element may drop a message (e.g. filter {} or parser {}) while non-filtering elements do not do that (e.g. rewrite or destination)
* a log path is considered "matching" if all filtering elements match, regardless of their position in the sequence
* a more complex form of filtering (e.g. if/else or junctions) are also filtering elements, e.g. if they don't match as a whole, they will cause the path to be "unmatched" just as if they were a simple filter expression.
* destinations do not alter the path "matched" state, even if it drops the message, this means that "dropping" all messages as a part of a destination's functionality, will not cause the path to be unmatched.

This means that something like this:
```
log {
    source(whatever);
    destination(dst1);
    filter { 1 eq 0; };
    destination(dst2);
};
```

is not considered a matching path, as we have a filter in the middle that drops messages. As a side effect of trying this path, dst1 will receive a message while dst2 will not.

On the other hand this is a matching log path, even though the side effects are the same:

```
log {
    source(whatever);
    destination(dst1);
    destination{
      channel { 
         filter { 1 eq 0; };
         destination(dst2);
      };
    };
};
```

This is matching as the "drop" will happen as a part of a deliver to a destination and that result is ignored on the top-level log path.

This might take some more discussions, possibly even IRL.

alltilla:
Closing in favor of #4058.

PR #4344, opened 2023-02-24
cfg: add named logpaths with counters

alltilla:
E.g.:
```
log top-level {
    source(s_local);

    log inner-1 {
        filter(f_inner_1);
        destination(d_local_1);
    };

    log inner-2 {
        filter(f_inner_2);
        destination(d_local_2);
    };
};
```

Each named log path counts its ingress and egress messages:
```
syslogng_log_path_ingress{id="top-level"} 114
syslogng_log_path_ingress{id="inner-1"} 114
syslogng_log_path_ingress{id="inner-2"} 114
syslogng_log_path_egress{id="top-level"} 103
syslogng_log_path_egress{id="inner-1"} 62
syslogng_log_path_egress{id="inner-2"} 41
```

Note that the egress statistics only count the messages which have been have not been filtered out from the related
log path, it does care about whether there are any destinations in it or that any destination delivers or drops the message.

Depends on #4325 and #4339.

### DIFF
--- a/lib/cfg-grammar.y
+++ b/lib/cfg-grammar.y
@@ -403,6 +403,7 @@
 
 %type	<cptr> string
 %type	<cptr> string_or_number
+%type	<cptr> optional_string
 %type	<cptr> normalized_flag
 %type   <ptr> string_list
 %type   <ptr> string_list_build
@@ -508,7 +509,15 @@ rewrite_stmt
           }
 
 log_stmt
-        : KW_LOG _log_context_push '{' log_content '}' _log_context_pop             { $$ = $4;}
+        : KW_LOG optional_string _log_context_push '{' log_content '}' _log_context_pop
+          {
+            if ($2)
+              {
+                log_expr_node_set_name($5, $2);
+                free($2);
+              }
+            $$ = $5;
+          }
 	;
 
 
@@ -696,8 +705,24 @@ log_forks
         ;
 
 log_fork
-        : KW_LOG '{' log_content '}'            { $$ = $3; }
-        | KW_CHANNEL '{' log_content '}'        { $$ = $3; }
+        : KW_LOG optional_string '{' log_content '}'
+          {
+            if ($2)
+              {
+                log_expr_node_set_name($4, $2);
+                free($2);
+              }
+            $$ = $4;
+          }
+        | KW_CHANNEL optional_string '{' log_content '}'
+          {
+            if ($2)
+              {
+                log_expr_node_set_name($4, $2);
+                free($2);
+              }
+            $$ = $4;
+          }
         ;
 
 log_conditional
@@ -1061,6 +1086,11 @@ string_or_number
         : string                                { $$ = $1; }
         | LL_NUMBER                             { $$ = strdup(lexer->token_text->str); }
         | LL_FLOAT                              { $$ = strdup(lexer->token_text->str); }
+        ;
+
+optional_string
+        : string                                { $$ = $1; }
+        |                                       { $$ = NULL; }
         ;
 
 path
--- a/lib/cfg-tree.c
+++ b/lib/cfg-tree.c
@@ -897,6 +897,16 @@ cfg_tree_propagate_expr_node_properties_to_pipe(LogExprNode *node, LogPipe *pipe
     pipe->expr_node = node;
 }
 
+static gboolean
+_is_log_path_name_unique(CfgTree *self, const LogExprNode *node)
+{
+  if (g_hash_table_contains(self->log_path_names, node->name))
+    return FALSE;
+
+  g_hash_table_insert(self->log_path_names, g_strdup(node->name), NULL);
+  return TRUE;
+}
+
 /**
  * cfg_tree_compile_sequence:
  *
@@ -942,6 +952,14 @@ cfg_tree_compile_sequence(CfgTree *self, LogExprNode *node,
       /* the catch-all resolution code clears this flag */
 
       msg_error("Error in configuration, catch-all flag can only be specified for top-level log statements");
+      goto error;
+    }
+
+  if (node->content == ENC_PIPE && node->name && !_is_log_path_name_unique(self, node))
+    {
+      msg_error("Error in configuration, duplicate log path definition",
+                evt_tag_str("log_path_name", node->name),
+                log_expr_node_location_tag(node));
       goto error;
     }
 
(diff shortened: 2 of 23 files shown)


alltilla:
Updated the E2E tests according to what we agreed on.

kira-syslogng:
Build FAILURE

lib/metrics-pipe.c
```diff
@@ -0,0 +1,123 @@
...
+  MetricsPipe *self = (MetricsPipe *) s;
+
+  StatsClusterKey ingress_sc_key;
+  StatsClusterKey egress_sc_key;
+  _init_stats_keys(self, &ingress_sc_key, &egress_sc_key);
+
+  stats_lock();
+  {
+    stats_register_counter(0, &ingress_sc_key, SC_TYPE_SINGLE_VALUE, &self->ingress_counter);
+    stats_register_counter(0, &egress_sc_key, SC_TYPE_SINGLE_VALUE, &self->egress_counter);
```
MrAnno, line 71: Shouldn't we put these metrics to stats-level 1 at least?

Naming a log path means we probably want to get those metrics, but when the configuration is generated, it might be worth allowing the user to disable them without removing the names.

Moreover, we may start using named log paths for several things in the future, so metrics may not be the only intention.

MrAnno (commented):
Nice!

I would like to look into the commit `[cfg-tree: inject metrics-pipes to named log paths](https://github.com/syslog-ng/syslog-ng/pull/4344/commits/e572f54b4eaf06b82d2fe8a9a94c5cbac6bbb88d)` more closely, I'll get back with an approve or another review note in a few hours.

kira-syslogng:

MrAnno:
@HofiOne Kira is broken due to an internal Java change. Can you check it, please?

lib/cfg-grammar.y
```diff
@@ -508,7 +509,15 @@ rewrite_stmt
           }
 
 log_stmt
-        : KW_LOG _log_context_push '{' log_content '}' _log_context_pop             { $$ = $4;}
+        : KW_LOG optional_string '{' _log_context_push log_content _log_context_pop '}'
```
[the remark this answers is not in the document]
alltilla: Oh this is a leftover from a previous iteration. Fixed.

lib/metrics-pipe.c
alltilla: I agree, changed it to `STATS_LEVEL1`.

kira-syslogng:
Build FAILURE

lib/metrics-pipe.c
MrAnno: Thanks.

HofiOne:
> @HofiOne Kira is broken due to an internal Java change. Can you check it, please?

done, sorry for the inconvenience

MrAnno:

MrAnno:
@kira-syslogng do stresstest

@HofiOne Thanks

kira-syslogng:

lib/cfg-grammar.y
```diff
@@ -508,7 +509,15 @@ rewrite_stmt
           }
 
 log_stmt
-        : KW_LOG _log_context_push '{' log_content '}' _log_context_pop             { $$ = $4;}
+        : KW_LOG optional_string _log_context_push '{' log_content '}' _log_context_pop
+          {
+            if ($2)
+              {
+                log_expr_node_set_name($5, $2);
+                g_free($2);
```
bazsi, line 517: This should be free() and not g_free()

And I know don't tell me

lib/metrics-pipe.c
```diff
@@ -0,0 +1,123 @@
...
+  log_pipe_init_instance(&self->super, cfg);
+  self->super.queue = _queue;
+  self->super.init = _init;
+  self->super.deinit = _deinit;
+  self->super.free_fn = _free;
+
+  self->log_path_name = g_strdup(log_path_name);
+
+  log_pipe_add_info(&self->super, "metrics-pipe");
+  log_pipe_add_info(&self->super, self->log_path_name);
```
bazsi, line 120: We should start specifying some kind of "type" qualifier to info, so that we could use the same mechanism for the debugger, but in a way that allows programmatic use.

Like use a key-value pair here.

bazsi: This is more like a where should we go next kind of comment, not to be addressed here.

lib/cfg-grammar.y
alltilla: Thanks, fixed!

lib/cfg-grammar.y
```diff
@@ -696,8 +705,24 @@ log_forks
         ;
 
 log_fork
-        : KW_LOG '{' log_content '}'            { $$ = $3; }
-        | KW_CHANNEL '{' log_content '}'        { $$ = $3; }
+        : KW_LOG optional_string '{' log_content '}'
+          {
+            if ($2)
+              {
+                log_expr_node_set_name($4, $2);
```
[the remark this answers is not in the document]
alltilla: I agree, fixed.

lib/metrics-pipe.c
alltilla: I agree, we could add more detailed info for the `LogPipe`s. Let's gather the use cases and make the `LogPipe` info functionality smarter.

lib/cfg-tree.c
```diff
@@ -897,6 +897,12 @@ cfg_tree_propagate_expr_node_properties_to_pipe(LogExprNode *node, LogPipe *pipe
     pipe->expr_node = node;
 }
 
+static gboolean
+_is_log_path_name_unique(CfgTree *self, const LogExprNode *node)
+{
+  return g_hash_table_insert(self->log_path_names, g_strdup(node->name), NULL);
```
MrAnno, line 913: > Starting from GLib 2.40, this function returns a boolean value
2.32 is our min version currently :(

MrAnno:
This is my last comment, I promise. LGTM otherwise.

lib/cfg-tree.c
alltilla: Wow, what a nice catch, thanks! :)

I have added a `contains()` call before it, which is available from 2.32: https://developer-old.gnome.org/glib/2.32/glib-Hash-Tables.html#g-hash-table-contains

MrAnno: Thanks.

MrAnno (approved):

PR #4350, opened 2023-02-28
Fix quick internal tests

HofiOne:
These changes were needed to 

- fix the quick internal tests, as the following issues are found
             - there were many hardcode JVM paths 
             - the automated detection and setup of JVM related environment was broken, as based on hardcoded values and wrong assumptions
             these issues could be found both in the internal test codes and in the dldb scripts (and therefore the docker builds)
             so, I've added
             - a bit more generic detection method
             - some tiny extensions to the already existing JVM env setup
             - a bit more sophisticated env setup to the kira/zts code and test scripts

- prepare support for none amd64 platform docker image usage
           the already existing DOCKER_BUILD_ARGS can be used to pass additional arguments to docker builds (e.g. a different than the default platform), so I just added this arg to the build command line where it was missing, and tried to add support of usage of an already, externally (pre-)defined value too

### DIFF
--- a/dbld/builddeps
+++ b/dbld/builddeps
@@ -81,8 +81,41 @@ function install_gosu() {
     chmod +x /usr/local/bin/gosu
 }
 
+# NOTE: libjvm.so is not anymore in the java lib dir itself, a better choice could be to use libnio.so to find the real lib dir
+#       for backward compatibility the libjvm.so is kept as the default search base
+function get_jvm_paths() {
+    LIB_FILE=libjvm.so
+    if [ $# -gt 0 ]; then
+        LIB_FILE=$1
+    fi
+    find / -name "${LIB_FILE}" 2>/dev/null | sed "s@/${LIB_FILE}@@g"
+}
+
+function set_jvm_envs() {
+	if [ "${JAVA_HOME}" == "" ]; then
+		JAVA_LIBRARY_PATH=$(get_jvm_paths $@)
+
+		if [ "${JAVA_LIBRARY_PATH}" != "" ]; then
+			OLD_LIB_PATH=${LD_LIBRARY_PATH}
+			if [ "${OLD_LIB_PATH}" != "" ]; then
+				export LD_LIBRARY_PATH=${JAVA_LIBRARY_PATH}:${OLD_LIB_PATH}
+			else
+				export LD_LIBRARY_PATH=${JAVA_LIBRARY_PATH}
+			fi
+
+			for i in {1..4}; do
+				if [ "$(basename $(dirname \"${JAVA_LIBRARY_PATH}\"))" == "jvm" ]; then
+					break
+				fi
+				JAVA_LIBRARY_PATH=$(dirname "${JAVA_LIBRARY_PATH}")
+            done
+            export JAVA_HOME=${JAVA_LIBRARY_PATH}
+		fi
+	fi
+}
+
 function set_jvm_paths() {
-    find / -name 'libjvm.so' | sed 's@/libjvm.so@@g' | tee --append /etc/ld.so.conf.d/openjdk-libjvm.conf
+    get_jvm_paths $@ | tee --append /etc/ld.so.conf.d/openjdk-libjvm.conf
     ldconfig
 }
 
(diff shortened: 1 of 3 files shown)


dbld/rules
```diff
@@ -39,7 +39,6 @@ DOCKER_RUN_ARGS=-e USER_NAME_ON_HOST=$(shell whoami)	\
 	-e GRADLE_PROJECT_CACHE_DIR=/build/gradle-cache \
 	-e GRADLE_FLAGS=--build-cache \
 	$(if $(wildcard ${HOME}/.gitconfig),-v ${HOME}/.gitconfig:/build/.gitconfig)
-DOCKER_BUILD_ARGS=
```
bazsi, line 42: I wouldn't remove this from here, as we are using it in our image-% rule below this point. you can still override this value, or maybe we could use ?= to only assign the empty value if it's unset.

dbld/images/hooks/build
```diff
@@ -10,4 +10,4 @@ DBLD_DIR="./${BUILD_PATH}"
 
 ARG_IMAGE_PLATFORM=$(basename $DOCKERFILE_PATH .dockerfile)
 ${DBLD_DIR}/prepare-image-build $ARG_IMAGE_PLATFORM
-docker build --build-arg=COMMIT=$(git rev-parse --short HEAD) --build-arg=ARG_IMAGE_PLATFORM=${ARG_IMAGE_PLATFORM} -t $IMAGE_NAME -f ${DBLD_DIR}/$DOCKERFILE_PATH ${DBLD_DIR}
+docker build ${DOCKER_BUILD_ARGS} --build-arg=COMMIT=$(git rev-parse --short HEAD) --platform=linux/amd64 --build-arg=ARG_IMAGE_PLATFORM=${ARG_IMAGE_PLATFORM} -t $IMAGE_NAME -f ${DBLD_DIR}/$DOCKERFILE_PATH ${DBLD_DIR}
```
bazsi, line 13: this build command was only used on dockerhub and we now build our stuff using github, so I guess the dockerhub based build environment could even be removed.

Can you elaborate why this is needed?

bazsi: also my docker build tool does not have --platform. I am probably using an old version.

The question again arises, where are we using the hooks/build script

dbld/builddeps
```diff
@@ -92,9 +125,9 @@ function install_gradle {
     mkdir -p /opt/gradle
     unzip -d /opt/gradle /tmp/gradle.zip
     rm -rf /tmp/gradle.zip
+    rm -rf /usr/bin/gradle
     ln -s /opt/gradle/gradle-${GRADLE_VERSION}/bin/gradle /usr/bin/gradle
-    find / -name 'libjvm.so' | sed 's@/libjvm.so@@g' | tee --append /etc/ld.so.conf.d/openjdk-libjvm.conf
-    ldconfig
+    set_jvm_paths $@
```
bazsi, line 130: I don't see any image that would use the argument to set_jvm_paths, could that be removed?

dbld/builddeps
```diff
@@ -81,8 +81,41 @@ function install_gosu() {
...
+#       for backward compatibility the libjvm.so is kept as the default search base
+function get_jvm_paths() {
+    LIB_FILE=libjvm.so
+    if [ $# -gt 0 ]; then
+        LIB_FILE=$1
+    fi
+    find / -name "${LIB_FILE}" 2>/dev/null | sed "s@/${LIB_FILE}@@g"
+}
+
+function set_jvm_envs() {
```
bazsi, line 94: this function is introduced here, but I don't see a user. where is this used from?

bazsi (commented):
thanks for the PR. I am probably missing at least some of the context. Can you please elaborate?

dbld/rules
HofiOne: the removal was intentional, having that over there will empty an already, externally pre-defined value, so you will not be able to pass it through the chain

dbld/builddeps
HofiOne: This is the main function that solves the internal build issues, as the environment manipulation partially happens in these scripts I just added the needed extensions next to the already existing ones

bazsi: I am not against adding this per-se. But it would be a lot better to have visibility into what uses this. Do you think you could publish the docker file that uses "dbld/builddeps"? or you are using it in a different context?

I am sorry but I don't really have a recollection of the internal tools, but adding this function here, without seeing why this is needed would be a fragile situation.

At least some comments on top would be useful here.

PR #4353, opened 2023-03-03
cmake: Reset fg color settings in Cmake config summary

HofiOne:
Trying to reset fg settings in config summary per line basis as it is not always happening in every terminal after the new line is opened

Signed-off-by: Hofi [hofione@gmail.com](mailto:hofione@gmail.com)


bazsi (approved):
This looks good to me. I didn't try it myself, but as I can imagine it improving things. And it seems to have run properly.

Just out of curiosity which terminal didn't work?

HofiOne:
> 

I guess i was wrong again, it is not a problem with the terminal, instead it is just an issue of how the redirected std output is captured, in the terminal some settings are reset automatically (e.g. the FG/BG color) in case a newline sequence received, but it is happening internally in the terminal implementation and not by added a new control sequence(s) to the std out stream, therefore when the captured log is presented there will be only sequences that set those setting but never there will be a reset sequence, so the output will be like you can see e.g. [here](https://ci.syslog-ng.com/jenkins/job/kira-test-compile-cmake-run-testdb/11005/consoleFull) 
this way now the reset is added forcibly to the output stream too and the colors will always bee correct (if the log viewer supports those control characters)

PR #4354, opened 2023-03-04
`python`: add support for typed custom options

alltilla:
This applies for `python` source, `python-fetcher` source, `python` destination, `python` parser and `python-http-header` inner destination.

Example config:
```
python(
  class("TestClass")
  options(
    "string_option" => "example_string"
    "bool_option" => yes  # becomes True in the python code
    "integer_option" => 123456789
    "double_option" => 123.456789
    "string_list_option" => ["string1", "string2", "string3"]
  )
);
```


bazsi:
This sounds like a great idea. Thanks for creating this

modules/python/python-types.c
```diff
@@ -34,15 +34,23 @@
 PyObject *
 py_bytes_from_string(const char *value, gssize len)
 {
+  PyGILState_STATE gstate = PyGILState_Ensure();
+
```
bazsi, line 38: I am a bit concerned if this has a runtime overhead for the normal case, e.g. during production, when the gil is already taken. If this is non trivial then I think we should create wrappers instead.

modules/python/python-grammar.ym
```diff
@@ -135,13 +135,7 @@ python_dd_custom_options
...
 
 python_dd_custom_option
-        : string string_or_number
-        {
-          python_dd_set_option(last_driver, $1, $2);
-          free($1);
-          free($2);
-        }
-	| string LL_ARROW string_or_number
+        : string optional_arrow string_or_number
```
bazsi: This also removes the actual setting of options, I guess it is readded by a subsequent commit. Would be nicer with not removing that.

modules/python/python-dest.c
```diff
@@ -86,11 +86,12 @@ python_dd_set_class(LogDriver *d, gchar *class_name)
 }
 
 void
-python_dd_set_option(LogDriver *d, gchar *key, gchar *value)
+python_dd_add_option(LogDriver *d, const gchar *name, PyObject *value)
```
bazsi, line 89: Can't we encapsulate the hashtable and related functions into a PythonOptions (or similarly named) structure which the python components would just contain?

tests/light/functional_tests/source_drivers/python_fetcher/test_python_fetcher_custom_options.py
```diff
@@ -0,0 +1,60 @@
...
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+#
+# As an additional exemption you are allowed to compile & link against the
+# OpenSSL libraries as published by the OpenSSL project. See the file
+# COPYING for details.
+#
+#############################################################################
+
+
+def test_python_fetcher_custom_options(config, syslog_ng):
```
bazsi, line 25: We wouldn't need to test all python components for all types, if we just merged the implementation into one. Then a simple happy path that the option stuff is working would be enough in my opinion.

Test execution time is a scarce resource too.

modules/python/python-types.c
alltilla: This got solved while introducing the `PythonOption` class.

modules/python/python-grammar.ym
alltilla: Actually it does not, just one of the two setter rule become unused. However, this was solved differently with the `PythonOptions` class, so it is kinda "fixed".

modules/python/python-dest.c
alltilla: My first implementation was like you have mentioned, but half way through it, I felt like it had too much boilerplate. However, I finished that implementation now and it is okay, I guess. Now we have 2 implementations, so we can decide :)

tests/light/functional_tests/source_drivers/python_fetcher/test_python_fetcher_custom_options.py
alltilla: I would really like to test all of these, as even with the new `PythonOptions` class there are still some code duplication. Also the python confgen has a completely different implementation.

I added python confgen tests, and merged all of these together to one light TC, which runs for 0.22 seconds, so I guess it is good enough.

alltilla:
@bazsi I have rebased this to master and fixed your comments. I feel like this implementation is good enough on the boilerplate front, but we can still carve from it if you find it necessary. Could you kindly take a look at it? Thanks!

alltilla:
Also added template support.

alltilla:
> Also added template support. Not sure how to differentiate it from the simple string type in the config. I chose `LogTemplate()`, which is the name of the python class, but we could use `template()` or something similar.
> 
> I like the idea, that we set the right handside with a python-like syntax, like with [] for the string list and here with LogTemplate. The only exception is yes and no, but we could make it, so only here, True and False is allowed additionally to yes and no, so we can write confgens and scls easily with yes and no, and we can also provide python like syntax with True and False.

Implemented that in a separate commit. We can drop it, if we want to do this in a different way.

modules/python/python-types.c
```diff
@@ -124,6 +124,29 @@ py_list_from_list(const gchar *list, gssize list_len)
...
+      return NULL;
+    }
+
+  for (const GList *elem = string_list; elem; elem = elem->next)
+    {
+      const gchar *string = (const gchar *) elem->data;
+      PyObject *py_string = py_string_from_string(string, -1);
+      if (!py_string || PyList_Append(obj, py_string) != 0)
+        {
+          Py_DECREF(obj);
```
bazsi, line 142: if PyList_Append() fails we leak py_string

modules/python/python-types.c
```diff
@@ -345,6 +368,31 @@ py_list_to_list(PyObject *obj, GString *list)
...
+
+  *string_list = NULL;
+
+  for (Py_ssize_t i = 0; i < PyList_GET_SIZE(obj); i++)
+    {
+      PyObject *element = PyList_GET_ITEM(obj, i);
+
+      const gchar *element_as_str;
+      if (!py_bytes_or_string_to_string(element, &element_as_str))
+        return FALSE;
```
bazsi, line 388: we should probably clean up the partially built string_list in this case.

modules/python/python-grammar.ym
```diff
@@ -330,6 +334,55 @@ python_parser_loaders
...
+optional_arrow
+          : LL_ARROW
+          |
+          ;
+
+python_option
+          : string optional_arrow string
+            {
+              PythonOption *option = python_option_string_new($1, $3);
+              python_options_add_option(last_python_options, option);
```
bazsi, line 265: this rule could return the PythonOption instance in $$ so the reference of this rule could perform python_options_add_option(), avoiding some of the duplication here.

bazsi: Also, python_options_add_option() will clone its argument, so "option" itself is leaked here.

I think we could get away with consuming the reference passed as argument, so no need for the clone. Maybe instead of clone we would only need refcounting?

bazsi (commented):
I had some smaller issues posted as review comments, otherwise this looks good. The LogTemplate functionality is not present the PR but is mentioned in the summary.

alltilla:
> The LogTemplate functionality is not present the PR but is mentioned in the summary.

D'oh, that's what I get for developing this on 2 computers... I will dig up the LogTemplate functionality and force push this PR.

modules/python/python-types.c
alltilla: Thanks, fixed!

modules/python/python-types.c
alltilla: Thanks, fixed!

modules/python/python-grammar.ym
alltilla: I think I will just get rid of the cloning as a whole. It is only used in the grammar, and it was only done to keep the ownerships "cleaner", but I don't think it served its purpose well.

modules/python/python-options.c
```diff
@@ -268,6 +269,54 @@ python_option_string_list_new(const gchar *name, const GList *value)
...
+} PythonOptionTemplate;
+
+static PyObject *
+_template_create_value_py_object(const PythonOption *s)
+{
+  PythonOptionTemplate *self = (PythonOptionTemplate *) s;
+  PyObject *template_str = py_string_from_string(self->value, -1);
+  PyObject *args = PyTuple_Pack(1, template_str);
+  PyObject *py_template = PyObject_Call((PyObject *) &py_log_template_type, args, NULL);
+  Py_DECREF(template_str);
```
bazsi, line 287: we might need to handle exceptions as all of these functions can return NULL in case of failure.

alltilla: Nice catch! I have handled exceptions by logging an error, but still continuing to build the rest of the `dict()`.

modules/python/python-grammar.ym
alltilla: Done.

alltilla:
Rebased to master, resolved the merge conflict.

modules/python/python-options.c
```diff
@@ -233,6 +234,51 @@ python_option_string_list_new(const gchar *name, const GList *value)
...
+_template_create_value_py_object(const PythonOption *s)
+{
+  PythonOptionTemplate *self = (PythonOptionTemplate *) s;
+
+  PyObject *template_str = py_string_from_string(self->value, -1);
+  if (!template_str)
+    return NULL;
+
+  PyObject *args = PyTuple_Pack(1, template_str);
+  PyObject *py_template = PyObject_Call((PyObject *) &py_log_template_type, args, NULL);
```
MrAnno, line 255: Why don't we just call `py_log_template_new()` here?

news/feature-4354.md
```diff
@@ -0,0 +1,19 @@
...
+`python` parser and `python-http-header` inner destination.
+
+Example config:
+```
+python(
+  class("TestClass")
+  options(
+    "string_option" => "example_string"
+    "bool_option" => True  # supported values are: True, False, yes, no
+    "integer_option" => 123456789
```
MrAnno, line 13: This PR introduces a little breaking change.
In the case of numbers and previously unquoted strings (true/false, yes/no), the Python-side representation changes.

If it would be difficult to keep the original behavior for old config versions, then I think we should mention it in the NEWS file.

modules/python/python-options.c
```diff
@@ -0,0 +1,336 @@
...
+    for (GList *elem = self->options; elem; elem = elem->next)
+      {
+        const PythonOption *option = (const PythonOption *) elem->data;
+        const gchar *name = python_option_get_name(option);
+        PyObject *value = python_option_create_value_py_object(option);
+
+        if (!value)
+          continue;
+
+        PyDict_SetItemString(py_dict, name, value);
```
MrAnno, line 319: This and the constructor call may fail. I don't know how much this actually happens. Probably in OOM cases.

MrAnno (commented):
LGTM.

modules/python/python-options.c
MrAnno: Oops. Because it is not implemented, only the declaration exists.

news/feature-4354.md
alltilla: Nice catch! I have added a runtime warning and a **Breaking change** sentence to the news file entry.

modules/python/python-options.c
alltilla: Thanks, I have handled those errors. In case of ctor fail, we can only return NULL, in case of set failure I have added an error message and continued.

news/feature-4354.md
MrAnno: Thanks.

PR #4362, opened 2023-03-09
syslog-ng health metrics and `syslog-ng-ctl healthcheck`

MrAnno:
This PR introduces a new syslog-ng-ctl command, which can be used to query a healthcheck status from syslog-ng.
Currently, 2 basic health values are reported:
- `mainloop_io_worker_roundtrip_latency_nanoseconds`: mainloop->io-worker-job->mainloop roundtrip - a basic measure for syslog-ng 
- `io_worker_latency_nanoseconds`: io-worker-job start latency

`syslog-ng-ctl healthcheck --timeout <seconds>` can be specified to use it as a boolean healthy/unhealthy check.

The 2 health checks are published as periodically updated metrics as well.
The frequency of these checks can be configured with the `stats(healthcheck-freq())` option. The default is 5 minutes.

### DIFF
--- a/lib/healthcheck/healthcheck-control.c
+++ b/lib/healthcheck/healthcheck-control.c
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2023 László Várady <laszlo.varady93@gmail.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#include "healthcheck-control.h"
+#include "healthcheck.h"
+
+#include "control/control.h"
+#include "control/control-connection.h"
+#include "control/control-commands.h"
+
+static void
+control_connection_healthcheck(ControlConnection *cc, GString *command, gpointer user_data, gboolean *cancelled)
+{
+
+}
+
+void
+healthcheck_register_control_commands(void)
+{
+  control_register_command("HEALTHCHECK", control_connection_healthcheck, NULL, FALSE);
+}
--- a/lib/healthcheck/healthcheck-control.c
+++ b/lib/healthcheck/healthcheck-control.c
@@ -33,15 +33,19 @@ static void
 _send_healthcheck_reply(HealthCheckResult result, gpointer c)
 {
   ControlConnection *cc = (ControlConnection *) c;
+  gchar double_buf[G_ASCII_DTOSTR_BUF_SIZE];
+
+  gdouble io_worker_latency = result.io_worker_latency / 1e9;
+  gdouble mainloop_io_worker_roundtrip_latency = result.mainloop_io_worker_roundtrip_latency / 1e9;
 
   GString *reply = g_string_new("OK ");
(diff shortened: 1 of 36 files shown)


lib/healthcheck/healthcheck-control.c
```diff
@@ -27,11 +27,39 @@
...
+
+static void
+_send_healthcheck_reply(HealthCheckResult result, gpointer c)
+{
+  ControlConnection *cc = (ControlConnection *) c;
+
+  GString *reply = g_string_new("OK ");
+
+  g_string_append_printf(reply, PROMETHEUS_METRIC_PREFIX
+                         "io_worker_latency_nanoseconds %" G_GUINT64_FORMAT"\n",
```
alltilla, line 40: prometheus [recommends](https://prometheus.io/docs/practices/naming/) to use base units, e.g. **seconds**:
> ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). See below for a list of base units.

alltilla (commented):
Nice feature :)

LGTM (except 1 minor thing) so far.

lib/healthcheck/healthcheck-control.c
MrAnno: Yeah. Unfortunately, representing this as a `double` is not precise enough (it showed 0 in most cases), so I decided to use nanoseconds anyway.

MrAnno: I don't think it would make sense as Prometheus will store it the same way as we would, losing precision. But let me know if you disagree.

MrAnno:
@kira-syslogng do stresstest

lib/healthcheck/healthcheck-control.c
MrAnno: We've talked about this IRL. I'll double-check the precision issue and try implementing "unit" support for stats.

lib/healthcheck/healthcheck-stats.c
```diff
@@ -0,0 +1,164 @@
...
+
+  stats_lock();
+  stats_unregister_counter(&sc_key_io_worker_latency, SC_TYPE_SINGLE_VALUE, &self->io_worker_latency);
+  stats_unregister_counter(&sc_key_mainloop_iow_rt_latency, SC_TYPE_SINGLE_VALUE,
+                           &self->mainloop_io_worker_roundtrip_latency);
+  stats_unlock();
+}
+
+void
+healthcheck_stats_init(HealthCheckStatsOptions *options)
```
alltilla, line 120: I think it would be worth to skip the init and deinit if the config is on an insufficient stats level.

alltilla: Or it could be done in timer_start(), like with freq. :)

alltilla: Or it could be done in timer_start(), like with freq. :)

MrAnno: Done. Thanks.

lib/healthcheck/healthcheck-control.c
MrAnno: I've added unit of measure support.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

alltilla (approved):
Nice feature!

I have had 2 optional comments, kindly consider them.

lib/stats/stats-prometheus.c
```diff
@@ -66,6 +67,54 @@ stats_format_prometheus_sanitize_name(const gchar *name)
   return sanitized_name->str;
 }
 
+gchar *
+stats_format_prometheus_format_value(const StatsClusterKey *key, const StatsCounterItem *counter)
+{
+  GString *value = scratch_buffers_alloc();
+
+  gsize stored_value = stats_counter_get((StatsCounterItem *) counter);
```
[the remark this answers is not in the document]
MrAnno: Fixing constness would be deeper here, as we'd need to go 3 layers down to Glib (where a cast is required).
I think it's not worth doing this in C, I'd leave this as-is or remove const completely.

lib/stats/stats-cluster.h
```diff
@@ -46,6 +46,23 @@ enum
   SCS_SOURCE_MASK    = 0xff
 };
 
+typedef enum _StatsClusterUnit
+{
+  SCU_NONE,
```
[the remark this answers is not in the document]
MrAnno: Done

MrAnno:
@kira-syslogng do stresstest

lib/stats/stats-cluster.h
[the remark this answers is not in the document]
alltilla: Thanks!

alltilla (approved):

PR #4335, opened 2023-02-18
qdisk: refactor qdisk_start()

alltilla:
This PR refactors the monster `qdisk_start()` function. There are some minor edge case fixes, but no functional change is expected.

No NEWS file needed.

Signed-off-by: Attila Szakacs <szakacs.attila96@gmail.com>

PR #4364, opened 2023-03-09
disk-buffer: refactor LogQueueDisk::start(), restart(), load_queue(), save_queue()

alltilla:
Depends on #4335

There are some corrupted restart related fixes, too.


modules/diskq/logqueue-disk-reliable.c
```diff
@@ -395,9 +396,9 @@ _free(LogQueue *s)
 {
   LogQueueDiskReliable *self = (LogQueueDiskReliable *)s;
 
-  _empty_queue(self->qreliable);
-  _empty_queue(self->qbacklog);
-  _empty_queue(self->qout);
+  _empty_queue(self, self->qreliable);
+  _empty_queue(self, self->qbacklog);
+  _empty_queue(self, self->qout);
   g_queue_free(self->qreliable);
```
bazsi, line 402: hmm can we have elements in those queues once stop() is called? I mean stop() does free these elements, doesn't it?

Maybe an assert() would be a better choice (e.g. asserting that stop was called or asserting that the queues are empty)

bazsi (commented):
Like the previous PR this looks awesome. Since this code is extremely critical, it would be important to apply as much testing to this as possible.

modules/diskq/logqueue-disk-reliable.c
alltilla: You are right, in theory we should not have any messages in the memory queues at this point.

kira-syslogng:
Build FAILURE

kira-syslogng:
Build FAILURE

alltilla:
@kira-syslogng do stresstest

kira-syslogng:

alltilla:
I had pushed some new commits to fix the memory usage counting. Everything is connected with everything in diskq... :/

alltilla:
@kira-syslogng do stresstest

modules/diskq/logqueue-disk-reliable.c
```diff
@@ -276,6 +281,9 @@ _pop_head(LogQueue *s, LogPathOptions *path_options)
 
   log_queue_queued_messages_dec(s);
 
+  if (qdisk_corrupt)
+    log_queue_disk_restart_corrupted(&self->super);
```
MrAnno, line 286: Could you explain this fix in detail, please?

What about threading? As I see, this is running in the same thread as `log_queue_disk_read_message()`.

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -224,6 +224,7 @@ _ack_backlog(LogQueue *s, gint num_msg_to_ack)
         return;
       msg = g_queue_pop_head(self->qbacklog);
       POINTER_TO_LOG_PATH_OPTIONS(g_queue_pop_head(self->qbacklog), &path_options);
+      log_queue_memory_usage_sub(&self->super.super, log_msg_get_size(msg));
```
MrAnno, line 227: When rewinding, we move messages from qbacklog to qout.
My understanding is that we should not do anything with memory_usage counters as qout_pop will decrease that for us.

modules/diskq/logqueue-disk-reliable.c
```diff
@@ -125,6 +125,7 @@ _ack_backlog(LogQueue *s, gint num_msg_to_ack)
               gint64 position;
               _pop_from_memory_queue_head(self->qbacklog, &position, &msg, &path_options);
 
+              log_queue_memory_usage_sub(s, log_msg_get_size(msg));
```
MrAnno, line 129: My previous comment must be true for qbacklog as well.

modules/diskq/qdisk.c
```diff
@@ -978,18 +978,17 @@ _save_queue(QDisk *self, GQueue *q, QDiskQueuePosition *q_pos)
...
        * saving them to disk, we ack them, they are restored as
        * non-flow-controlled entries later, but then we've saved them to
        * disk anyway. */
 
-      POINTER_TO_LOG_PATH_OPTIONS(g_queue_pop_head(q), &path_options);
+      msg = g_queue_peek_nth(q, i);
+      POINTER_TO_LOG_PATH_OPTIONS(g_queue_peek_nth(q, i+1), &path_options);
       log_msg_serialize(msg, sa, 0);
-      log_msg_ack(msg, &path_options, AT_PROCESSED);
-      log_msg_unref(msg);
```
MrAnno, line 992: I think this patch needs to be explained carefully.

According to my previous comments, I think that queues should not always be emptied when the disk-buffer is stopped (corrupt cases). Was this a double-ack/unref issue prior to your changes?

MrAnno (changes requested):

modules/diskq/logqueue-disk-reliable.c
MrAnno: About queue emptying: I think we should try sending these messages out before replacing the queue (to avoid message loss).

modules/diskq/logqueue-disk.c
```diff
@@ -82,7 +82,6 @@ log_queue_disk_get_filename(LogQueue *s)
 void
 log_queue_disk_free_method(LogQueueDisk *self)
 {
-  qdisk_stop(self->qdisk);
```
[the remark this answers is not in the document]
alltilla: I think I did the same, but with the naming start() - stop() vs load_queue() - save_queue().

modules/diskq/logqueue-disk-non-reliable.c
alltilla: I agree that if we are moving between memory queues, we should not modify the value of the memory usage counter.
But here, we are in `ack_backlog()`, not in `rewind_backlog()`, where messages are removed from the `qbacklog`, and not pushed to another one, so I think we should decrease the counter.

modules/diskq/logqueue-disk.c
```diff
@@ -182,9 +180,6 @@ _restart_diskq(LogQueueDisk *self)
     }
   g_free(new_file);
 
-  if (self->restart)
-    self->restart(self, options);
```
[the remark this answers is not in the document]
alltilla: I agree. Updated the commit msg a bit:
```
    logqueue-disk: remove restart virtual function

    With both implementation it only sets values which are already set
    in the QDisk, as the queue type and options do not change during
    a corrupted reload.
```

modules/diskq/logqueue-disk-reliable.c
alltilla: It would be nice to send them out, but I think it is a complex task.

As you know, in case of reliable disk buffer, the memory queue messages are stored with positions, and we only pop a message from them, if we popped its duplicate from the disk, based on its position.
If we start a new disk-buffer file, but keep the messages in the memory queues, this would rarely be true, and if it were, it would be a problem, as the message in the memory queue and the message in the disk are on the same position, but they are different.
So we would need another memory only message queue for the messages left behind from the previous disk-buffer file.

We can do this, but it was never done before, so it is not a regression, and I would not prioritize it high.

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -521,13 +535,19 @@ _save_queue(LogQueueDisk *s, gboolean *persistent)
...
     {
       *persistent = TRUE;
-      return TRUE;
+      result = TRUE;
     }
 
-  return FALSE;
+  _empty_queue(self->qoverflow);
+  _empty_queue(self->qout);
+  _empty_queue(self->qbacklog);
```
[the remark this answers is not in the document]
alltilla: I agree, we can easily do this with non-reliable queues, and my commits introduced a regression in this sense, so I will implement this.

modules/diskq/logqueue-disk.c
```diff
@@ -191,7 +191,10 @@ static void
 _restart_diskq(LogQueueDisk *self)
 {
   gchar *filename = g_strdup(qdisk_get_filename(self->qdisk));
-  qdisk_stop(self->qdisk);
+
+  gboolean persistent;
+  if (self->save_queue)
+    self->save_queue(self, &persistent);
```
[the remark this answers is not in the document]
alltilla: If I understand correctly this resonates with your other requested change comment. If I implement keeping the messages in the memory queues in case of a corrupted reload, I think this should resolve.

modules/diskq/qdisk.c
alltilla: Prior to my commits, the messages were freed in `_save_queue()` -> `qdisk_save_state()` -> `_save_queue()`, then there were unnecessary frees in `_free()` -> `_free_queue()`, and no one decreased the memory usage counter.

With my commits the messages are only freed in `_stop()` -> `empty_queue()`, where the memory usage counter is decreased.

I will add a verbose commit message, because it is hard to follow my changes through different commits.

alltilla:
```
qdisk: do not consume messages in save_state()

Prior to my this PR, the messages were freed in
_save_queue() -> qdisk_save_state() -> _save_queue(), then there were
unnecessary frees in _free() -> _free_queue(), and no one decreased the
memory usage counter.

With the previous commits I have moved the freeing to _stop(), and the
memory usage is decreased there as well, but the qdisk_save_state()
branch frees it before we try to do it in _stop(), so there are no
messages there to free, and we cannot correctly decrease the memory
usage counter.

This commit changes this, so we no longer free the messages in
qdisk_save_state(), so they are only freed in _stop() -> empty_queue(),
where the memory usage counter is decreased.
```

alltilla:
I added one commit as a fixup, and modified some commit messages.

modules/diskq/logqueue-disk.c
```diff
@@ -165,13 +165,36 @@ log_queue_disk_drop_message(LogQueueDisk *self, LogMessage *msg, const LogPathOp
...
+}
+
 static void
 _restart_diskq(LogQueueDisk *self)
 {
   gchar *filename = g_strdup(qdisk_get_filename(self->qdisk));
   qdisk_stop(self->qdisk);
 
-  gchar *new_file = g_strdup_printf("%s.corrupted", filename);
+  gchar *new_file = _get_next_corrupted_filename(filename);
```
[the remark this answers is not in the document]
alltilla: Nice catch, fixed.

modules/diskq/logqueue-disk-reliable.c
MrAnno: Right, okay.

modules/diskq/qdisk.c
MrAnno: Thanks.

modules/diskq/logqueue-disk-non-reliable.c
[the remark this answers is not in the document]
alltilla: I have added the logic to keep the memory queues if we restart corrupted.

MrAnno:
LGTM.

kira-syslogng:
Build FAILURE

modules/diskq/logqueue-disk-non-reliable.c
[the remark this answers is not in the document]
MrAnno: Thanks.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

MrAnno (approved):

PR #4366, opened 2023-03-09
logpipe: fix crash for synthetic messages in log_path_options_pop_jun…

bazsi:
This is a critical fix for a bug reported on gitter

https://matrix.to/#/!JQkTKQPSHHDdacODVM:gitter.im/$tSlHLhOvj4xbOQ6jn11Ut3WA8M4ANZBunGmAT5BfEqI?via=gitter.im&via=matrix.org&via=anno.io

This brings up a separate issue, namely what happens to messages dropped due to correlation (e.g. grouping-by() in aggregate-only). I think those messages should need to be considered "matching" even though we are dropping them as we consume them in our correlation context. But that's not a crasher, so let's do that separately.

If this seems to be a good fix, we probably need to do a 4.1.1, unfortunately.

### DIFF
--- a/lib/logpipe.h
+++ b/lib/logpipe.h
@@ -231,14 +231,25 @@ log_path_options_push_junction(LogPathOptions *local_path_options, gboolean *mat
 static inline void
 log_path_options_pop_conditional(LogPathOptions *local_path_options)
 {
-  local_path_options->matched = local_path_options->parent->matched;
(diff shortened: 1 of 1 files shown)


smortex (approved):
I confirm this PR fix the issue I was facing!

:100:

alltilla (approved):

MrAnno (approved):

PR #4367, opened 2023-03-10
unit-tests: add 5 min timeout for all tests which did not have any

alltilla:
I faced another stuck unit test in our CI run, and it was a pain to determine which one was it, because there is no "test started" message.

I think it would be a good habit to set a timeout for all of our existing and new unit tests. 5 minutes should be plenty enough, maybe we can even make it 1 minute. There is no reason to make the CI run for 1 hour if a unit test gets stuck.

### DIFF
--- a/lib/ack-tracker/tests/test_batched_ack_tracker.c
+++ b/lib/ack-tracker/tests/test_batched_ack_tracker.c
@@ -140,7 +140,7 @@ _teardown(void)
   cfg_free(cfg);
 }
 
-TestSuite(batched_ack_tracker, .init = _setup, .fini = _teardown);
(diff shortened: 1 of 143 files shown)


alltilla:
There are some unit tests, which do not define `TestSuite`. I don't really want to create one for them right now, it would take too much time. They won't get the 5 min timeout.

alltilla:
Wow distcheck is hanging on this PR too, this is rude... :D And probably I didn't fix the problem :(

MrAnno:
If we need a 5-minute timeout for all of our tests, I think we should do that in a different layer: not (only) in unit tests, and somewhere central, where it can be forced properly without repetition.

PR #4370, opened 2023-03-10
fix "matched" logic in case of correlating parsers

bazsi:
As a followup to #4366 and #4058, another edge case was identified where syslog-ng evaluation handles matched messages incorrectly.
    
If we are using a correlation while processing the message, the incoming messages are consumed into a context and then the individual messages get dropped, at least when using inject-mode(aggregate-only).
    
In these cases we should not consider dropping of these messages to be unmatching, as dropping happens as a part of correlation.
    
This patch adds a light based testcase to exercise this problem and includes a fix, so that messages consumed into corrlation in `inject-mode(aggregate-only)` mode will not cause those messages to traverse any alternative paths.

### DIFF
--- a/lib/parser/parser-expr.c
+++ b/lib/parser/parser-expr.c
@@ -78,8 +78,8 @@ log_parser_process_message(LogParser *self, LogMessage **pmsg, const LogPathOpti
   return success;
 }
 
-static void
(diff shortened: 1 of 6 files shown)


MrAnno:
LGTM.

Please write a news entry.

bazsi:
Added a news entryr.

MrAnno (approved):

Issue #4212, opened 2022-11-14
Native macOS system() source support to syslog-ng


HofiOne:
## Description of the problem
Add macOS native syslog collection support to the system() driver via appropriate macOS APIs, replace the current file source file("/var/log/system.log" follow-freq(1))

## Proposed solution
Depending on the syslog-ng backward-supported macOS versions, prior to macOS 10.12 we can use ASL syslog API, above that the OSLog framework can be used

## Additional context
https://yash6866.gitbook.io/syslog-macos-testing/modules/system-source-1

PR #4372, opened 2023-03-15
darwinosl:  Native macOS system() source module #4212 

HofiOne:
This is now just a draft, so there are some todos in the source as well, needs additional test cases, and also should add here
- full details of the modifications
- detailed how to use
- news file
(till that please see #4212 for an explanation)


kira-syslogng:
Build FAILURE

HofiOne:
Found some severe API bugs leading to a high memory watermark, and it probably needs some deep refactoring to implement a workaround (if any)

PR #4378, opened 2023-03-21
stats: add StatsClusterKeyBuilder

alltilla:
This will be useful for places, where multiple layers add labels on the call chain.

No news file needed.

Related PR #4356.


kira-syslogng:
Build FAILURE

MrAnno:

lib/stats/stats-cluster-key-builder.c
```diff
@@ -0,0 +1,182 @@
...
+    return g_strdup_printf("%s%s", self->name, self->name_suffix);
+
+  return g_strdup(self->name);
+}
+
+void
+stats_cluster_key_builder_build_single(StatsClusterKeyBuilder *self, StatsClusterKey *sc_key)
+{
+  gchar *name = _format_name(self);
+  g_array_sort(self->labels, (GCompareFunc) _labels_sort);
```
MrAnno, line 146: Why do we need to sort this?

lib/stats/stats-cluster-key-builder.h
```diff
@@ -0,0 +1,46 @@
...
+
+void stats_cluster_key_builder_set_name(StatsClusterKeyBuilder *self, const gchar *name);
+void stats_cluster_key_builder_set_name_suffix(StatsClusterKeyBuilder *self, const gchar *name_suffix);
+void stats_cluster_key_builder_add_label(StatsClusterKeyBuilder *self, const StatsClusterLabel label);
+void stats_cluster_key_builder_set_unit(StatsClusterKeyBuilder *self, StatsClusterUnit unit);
+void stats_cluster_key_builder_set_legacy_alias(StatsClusterKeyBuilder *self, guint16 component, const gchar *id,
+                                                const gchar *instance);
+void stats_cluster_key_builder_set_legacy_alias_name(StatsClusterKeyBuilder *self, const gchar *name);
+
+void stats_cluster_key_builder_build_single(StatsClusterKeyBuilder *self, StatsClusterKey *sc_key);
```
MrAnno, line 43: Suggestion:
Why don't we separate cheap and more expensive key creations?

Let's say that stack-based key allocation is cheap, and the KeyBuilder should not be used in such cases (the builder is heap-allocated, it can be considered more expensive: you wouldn't use it for dynamic counters).

Since we don't really have heap-allocated keys yet, we could say that the more expensive method should be used through the KeyBuilder.

Based on the above, we could change the signature of a few functions:
```
StatsClusterKey *stats_cluster_key_builder_build_single(StatsClusterKeyBuilder *self);
StatsClusterKey *stats_cluster_key_builder_build_logpipe(StatsClusterKeyBuilder *self);
```
And then we could:
```diff
+ stats_cluster_key_new() // package-private method (only the builder should use it)
-stats_cluster_key_cloned_free()
+stats_cluster_key_free() // package-private method (only the builder should use it)
```

lib/stats/stats-cluster-key-builder.c
```diff
@@ -0,0 +1,182 @@
...
+                                                            self->legacy.instance, self->legacy.name);
+      else
+        stats_cluster_single_key_add_legacy_alias(&temp_key, self->legacy.component, self->legacy.id,
+                                                  self->legacy.instance);
+    }
+  stats_cluster_single_key_add_unit(&temp_key, self->unit);
+  stats_cluster_key_clone(sc_key, &temp_key);
+
+  g_free(name);
+  return;
```
MrAnno, line 163: This is unneeded.

MrAnno (commented):
I absolutely love this. 👍🏻 👍🏻 👍🏻 👍🏻

lib/stats/stats-cluster-key-builder.c
```diff
@@ -0,0 +1,182 @@
...
+stats_cluster_key_builder_build_logpipe(StatsClusterKeyBuilder *self, StatsClusterKey *sc_key)
+{
+  gchar *name = _format_name(self);
+  g_array_sort(self->labels, (GCompareFunc) _labels_sort);
+
+  StatsClusterKey temp_key;
+  stats_cluster_logpipe_key_set(&temp_key, name, (StatsClusterLabel *) self->labels->data, self->labels->len);
+  if (self->legacy.set)
+    {
+      stats_cluster_logpipe_key_add_legacy_alias(&temp_key, self->legacy.component, self->legacy.id, self->legacy.instance);
```
MrAnno, line 176: Should we add a `!self->legacy.name` assert?

lib/stats/stats-cluster-key-builder.c
alltilla: Fixed :)

lib/stats/stats-cluster-key-builder.h
alltilla: This is a really good idea.

I think `free()` must be used by the caller, to free the created key.

And if I understand correctly `new()` would be a `g_new0()` call only, which can be done in the `build()` functions, too, so we do not expose something, which is not meant to be used.

If these are all true, then this task reduces to a function rename from `stats_cluster_key_cloned_free()` to `stats_cluster_key_free()`, and allocating the key inside `build()` and returning that.

lib/stats/stats-cluster-key-builder.c
alltilla: I was checking why the unit test passes like this, and had a mini heart attack, because the legacy name is not checked in `stats_cluster_key_equal()`, and on the first glance, I think it should be. I need a bit more time to think this through, let's discuss it tomorrow :)

MrAnno: The name is for single keys, group_equals use this correctly.
I think we should just add an assert for the logpipe builder.

lib/stats/stats-cluster-key-builder.h
alltilla: Oh `stats_cluster_key_free()` needs to free `self`, too, so we need both, I think.

alltilla: Implemented this, but we can iterate on it further.

MrAnno: LGTM.

PR #4380, opened 2023-03-21
qdisk: preallocate qdisk file after creating header

alltilla:
Related comment: https://github.com/syslog-ng/syslog-ng/pull/4335#discussion_r1135321974

No news file needed.

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -294,14 +294,7 @@ _maybe_truncate_file(QDisk *self, gint64 expected_size)
             evt_tag_int("fd", self->fd));
 }
 
-#if SYSLOG_NG_HAVE_POSIX_FALLOCATE
-static gint
-_posix_preallocate(QDisk *self, gint64 size)
-{
-  return posix_fallocate(self->fd, 0, (off_t) size);
-}
-
-#else
+#if !SYSLOG_NG_HAVE_POSIX_FALLOCATE
 static gint
 _compat_preallocate(QDisk *self, gint64 size)
 {
@@ -338,7 +331,7 @@ _preallocate_qdisk_file(QDisk *self, off_t size)
   gint result;
 
 #if SYSLOG_NG_HAVE_POSIX_FALLOCATE
-  result = _posix_preallocate(self, size);
+  result = posix_fallocate(self->fd, 0, size);
 #else
   result = _compat_preallocate(self, size);
 #endif
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -331,9 +331,9 @@ _preallocate_qdisk_file(QDisk *self, off_t size)
   gint result;
 
 #if SYSLOG_NG_HAVE_POSIX_FALLOCATE
-  result = posix_fallocate(self->fd, 0, size);
+  result = posix_fallocate(self->fd, QDISK_RESERVED_SPACE, size - QDISK_RESERVED_SPACE);
 #else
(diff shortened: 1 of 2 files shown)


modules/diskq/qdisk.c
```diff
@@ -320,12 +320,7 @@ _compat_preallocate(QDisk *self, gint64 size)
   gint64 buf_write_iterations = size / buf_size;
   gint64 additional_write_size = size - buf_write_iterations * buf_size;
 
-  gchar buf[buf_size];
-  for (gint i = 0; i < buf_size; i++)
-    {
-      buf[i] = '\0';
-    }
-  off_t pos = 0;
```
MrAnno, line 328: `pos` has been accidentally removed from here.

modules/diskq/qdisk.c
```diff
@@ -343,8 +341,7 @@ _compat_preallocate(QDisk *self, gint64 size)
       return FALSE;
     }
 
-  self->cached_file_size = pos + additional_write_size;
-  g_assert(self->cached_file_size == size);
+  g_assert(pos + additional_write_size == len + offset);
```
MrAnno, line 318: `len + offset` are undeclared in this commit.

modules/diskq/qdisk.c
```diff
@@ -296,22 +296,23 @@ _maybe_truncate_file(QDisk *self, gint64 expected_size)
 
 #if !SYSLOG_NG_HAVE_POSIX_FALLOCATE
 static gint
-_compat_preallocate(QDisk *self, gint64 size)
+_compat_preallocate(int fd, off_t offset, off_t len)
 {
   const size_t buf_size = QDISK_RESERVED_SPACE;
-  gint64 buf_write_iterations = size / buf_size;
-  gint64 additional_write_size = size - buf_write_iterations * buf_size;
+  gint64 buf_write_iterations = (len - offset) / buf_size;
```
MrAnno, line 302: `len` is not necessarily greater than `offset`.

I think this calculation is incorrect.

> The function posix_fallocate() ensures that disk space is
       allocated for the file referred to by the file descriptor fd for
       the bytes in the range starting at offset and continuing for len
       bytes.

Based on this, we should write exactly `len` bytes after `offset`.

MrAnno:
I've just found this (a bit unrelated, sorry):

`$ man 3p mmap`
> If  the size of the mapped file changes after the call to mmap() as a result of some other operation on the mapped file, the effect of references to portions of the mapped region that correspond to added or removed portions of the file is unspecified.

I think we should reorder a few lines in `_create_header()`.

modules/diskq/qdisk.c
alltilla: Thanks, fixed.

modules/diskq/qdisk.c
alltilla: Yes, I messed up the commit order. Fixed. :)

modules/diskq/qdisk.c
alltilla: Thanks, fixed.

alltilla:
> I've just found this (a bit unrelated, sorry):
> 
> `$ man 3p mmap`
> 
> > If  the size of the mapped file changes after the call to mmap() as a result of some other operation on the mapped file, the effect of references to portions of the mapped region that correspond to added or removed portions of the file is unspecified.
> 
> I think we should reorder a few lines in `_create_header()`.

Nice catch, added a new commit, which implements this.

MrAnno (approved):

kira-syslogng:

PR #4381, opened 2023-03-21
qdisk: fix filename race

alltilla:
This PR fixes a rare race condition when calculating disk-buffer filename:

As two threads or two syslog-ng instances can run `_next_filename()` simultaneously, and they can get the same filename.

To mitigate this I have introduced an flock, which solves the issue between two syslog-ng instances, but does not solve it for threads in the same syslog-ng instance (man 2 flock):

> If  a  process  uses  open(2)  (or similar) to obtain more than one file descriptor for the same
> file, these file descriptors are treated independently by flock().  An attempt to lock the  file
> using one of these file descriptors may be denied by a lock that the calling process has already
> placed via another file descriptor.

To solve this, I have introduced a static lock, too.

Depends on #4380.

### DIFF
--- a/modules/diskq/qdisk.c
+++ b/modules/diskq/qdisk.c
@@ -37,6 +37,7 @@
 #include <errno.h>
 #include <unistd.h>
 #include <sys/types.h>
+#include <sys/file.h>
 
 /* MADV_RANDOM not defined on legacy Linux systems. Could be removed in the
  * future, when support for Glibc 2.1.X drops.*/
@@ -49,6 +50,10 @@
 #define PATH_QDISK              PATH_LOCALSTATEDIR
 
 #define QDISK_HDR_VERSION_CURRENT 3
+
+#define DIRLOCK_FILENAME "syslog-ng-disk-buffer.dirlock"
+
+static GMutex filename_lock;
 
 typedef union _QDiskFileHeader
 {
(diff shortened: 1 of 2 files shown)


modules/diskq/qdisk.c
```diff
@@ -210,6 +210,11 @@ _next_filename(QDisk *self, gchar *filename, gsize filename_len)
       return FALSE;
     }
 
+  gint fd = -1;
+  if (!_create_file(filename, &fd))
```
MrAnno, line 273: `_create_file()` probably should not open the created file.

modules/diskq/qdisk.c
```diff
@@ -182,6 +187,55 @@ _create_file(const gchar *filename, gint *fd)
...
+    {
+      msg_error("Failed to open disk-buffer flock file",
+                evt_tag_str("filename", flock_file_path),
+                evt_tag_error("error"));
+      g_mutex_unlock(&filename_lock);
+      g_free(flock_file_path);
+      return FALSE;
+    }
+
+  if (flock(*fd, LOCK_EX) < 0)
```
MrAnno, line 217: `flock` is not completely portable, but it seems to be working on our supported platforms.

modules/diskq/qdisk.c
alltilla: Yes, you are right. Fixed :)

modules/diskq/qdisk.c
alltilla: Thanks for checking!

modules/diskq/qdisk.c
```diff
@@ -181,6 +186,55 @@ _create_file(const gchar *filename)
   return TRUE;
 }
 
+static gboolean
+_grab_filename_locks(const gchar *dir, gint *fd)
+{
+  gchar *flock_file_path = g_strdup_printf("%s/%s", dir, FLOCK_FILENAME);
```
MrAnno, line 192: I think using `g_build_path(G_DIR_SEPARATOR_S, ...` would be a bit better (portable separator and handles a few corner cases when building paths).

MrAnno (commented):
LGTM.

modules/diskq/qdisk.c
```diff
@@ -50,6 +51,10 @@
 
 #define QDISK_HDR_VERSION_CURRENT 3
 
+#define FLOCK_FILENAME "disk-buffer.flock"
```
MrAnno, line 54: Can we add `syslog-ng` to the name?
The disk-buffer dir should be private and dedicated for syslog-ng, but sometimes it's just `/var`.

For example:
`syslog-ng-disk-buffer.flock`
`syslog-ng-disk-buffer.dirlock`
`syslog-ng-disk-buf-dir.lock`
`syslog-ng.diskqdirlock`

modules/diskq/qdisk.c
alltilla: Thanks, fixed!

modules/diskq/qdisk.c
alltilla: Yes, added it :)

alltilla:
@kira-syslogng do stresstest

MrAnno (approved):

PR #4385, opened 2023-03-23
logqueue: fix queued_messages_reset() 

alltilla:
The old implementation erroneously set the shared counter to the queue's new queued counter value, which is a problem, when there are multiple queues connected to the same driver.

### DIFF
--- a/news/bugfix-4385.md
+++ b/news/bugfix-4385.md
@@ -1,0 +1,1 @@
+`disk-buffer`: Fixed queued messages stats counting, when a disk-buffer became corrupted.


MrAnno (approved):

alltilla:
Sorry, I had to force push, because the new UT left a file behind, which was a problem for distcheck.

alltilla:
@kira-syslogng do stresstest

kira-syslogng:

MrAnno (approved):
Nice catch.

PR #4386, opened 2023-03-23
Update Dockerfile

wojtekxtx:
Make dockerfile upto modern standards

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,24 +4,13 @@ LABEL org.opencontainers.image.source="https://github.com/syslog-ng/syslog-ng"
 
 ARG PKG_TYPE=stable
 
-RUN apt-get update -qq && apt-get install -y \
-    wget \
-    ca-certificates \
-    gnupg2 \
-    && rm -rf /var/lib/apt/lists/*
-
-RUN wget -qO - https://ose-repo.syslog-ng.com/apt/syslog-ng-ose-pub.asc | gpg --dearmor > /usr/share/keyrings/ose-repo-archive-keyring.gpg && \
-  echo "deb [signed-by=/usr/share/keyrings/ose-repo-archive-keyring.gpg] https://ose-repo.syslog-ng.com/apt/ ${PKG_TYPE} debian-testing" | tee --append /etc/apt/sources.list.d/syslog-ng-ose.list
-
-RUN apt-get update -qq && apt-get install -y \
-    libdbd-mysql libdbd-pgsql libdbd-sqlite3 syslog-ng libjemalloc2 \
-    && rm -rf /var/lib/apt/lists/*
+RUN apt update -qq && apt install -y wget ca-certificates gnupg2 libdbd-mysql libdbd-pgsql libdbd-sqlite3 syslog-ng libjemalloc2 && rm -rf /var/lib/apt/lists/*&& rm -rf /var/lib/apt/lists/* && \
+wget -qO - https://ose-repo.syslog-ng.com/apt/syslog-ng-ose-pub.asc | gpg --dearmor > /usr/share/keyrings/ose-repo-archive-keyring.gpg && \
+echo "deb [signed-by=/usr/share/keyrings/ose-repo-archive-keyring.gpg] https://ose-repo.syslog-ng.com/apt/ ${PKG_TYPE} debian-testing" | tee --append /etc/apt/sources.list.d/syslog-ng-ose.list
 
 ADD syslog-ng.conf /etc/syslog-ng/syslog-ng.conf
 
-EXPOSE 514/udp
-EXPOSE 601/tcp
-EXPOSE 6514/tcp
+EXPOSE 514/udp 601/tcp 6514/tcp
 
 HEALTHCHECK --interval=2m --timeout=5s --start-period=30s CMD /usr/sbin/syslog-ng-ctl healthcheck --timeout 5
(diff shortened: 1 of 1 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

docker/Dockerfile
```diff
@@ -4,24 +4,13 @@ LABEL org.opencontainers.image.source="https://github.com/syslog-ng/syslog-ng"
...
-
-RUN wget -qO - https://ose-repo.syslog-ng.com/apt/syslog-ng-ose-pub.asc | gpg --dearmor > /usr/share/keyrings/ose-repo-archive-keyring.gpg && \
-  echo "deb [signed-by=/usr/share/keyrings/ose-repo-archive-keyring.gpg] https://ose-repo.syslog-ng.com/apt/ ${PKG_TYPE} debian-testing" | tee --append /etc/apt/sources.list.d/syslog-ng-ose.list
-
-RUN apt-get update -qq && apt-get install -y \
-    libdbd-mysql libdbd-pgsql libdbd-sqlite3 syslog-ng libjemalloc2 \
-    && rm -rf /var/lib/apt/lists/*
+RUN apt update -qq && apt install -y wget ca-certificates gnupg2 libdbd-mysql libdbd-pgsql libdbd-sqlite3 syslog-ng libjemalloc2 && rm -rf /var/lib/apt/lists/*&& rm -rf /var/lib/apt/lists/* && \
+wget -qO - https://ose-repo.syslog-ng.com/apt/syslog-ng-ose-pub.asc | gpg --dearmor > /usr/share/keyrings/ose-repo-archive-keyring.gpg && \
+echo "deb [signed-by=/usr/share/keyrings/ose-repo-archive-keyring.gpg] https://ose-repo.syslog-ng.com/apt/ ${PKG_TYPE} debian-testing" | tee --append /etc/apt/sources.list.d/syslog-ng-ose.list
```
MrAnno, line 9: Please keep those line breaks. Readability is also important.

MrAnno:
@kira-syslogng ok to test

docker/Dockerfile
wojtekxtx: Readability is not changed here; its even improved slightly.One can always set line-wrap to the same effect. So I dont understand your objection(s).

docker/Dockerfile
```diff
@@ -4,24 +4,13 @@ LABEL org.opencontainers.image.source="https://github.com/syslog-ng/syslog-ng"
...
-    gnupg2 \
-    && rm -rf /var/lib/apt/lists/*
-
-RUN wget -qO - https://ose-repo.syslog-ng.com/apt/syslog-ng-ose-pub.asc | gpg --dearmor > /usr/share/keyrings/ose-repo-archive-keyring.gpg && \
-  echo "deb [signed-by=/usr/share/keyrings/ose-repo-archive-keyring.gpg] https://ose-repo.syslog-ng.com/apt/ ${PKG_TYPE} debian-testing" | tee --append /etc/apt/sources.list.d/syslog-ng-ose.list
-
-RUN apt-get update -qq && apt-get install -y \
-    libdbd-mysql libdbd-pgsql libdbd-sqlite3 syslog-ng libjemalloc2 \
-    && rm -rf /var/lib/apt/lists/*
+RUN apt update -qq && apt install -y wget ca-certificates gnupg2 libdbd-mysql libdbd-pgsql libdbd-sqlite3 syslog-ng libjemalloc2 && rm -rf /var/lib/apt/lists/*&& rm -rf /var/lib/apt/lists/* && \
```
MrAnno, line 7: These 2 apt install commands can not be fully merged as `syslog-ng` needs to come from `ose-repo.syslog-ng.com`.

`wget`, `ca-certificates`, and `gnupg2` are needed to add our repository to `apt`, and only then we can install syslog-ng.

MrAnno (changes requested):

docker/Dockerfile
MrAnno: Okay. Since this is only a matter of taste, I'm marking my comment as resolved.

docker/Dockerfile
wojtekxtx: > wget, ca-certificates, and gnupg2 are needed to add our repository to apt, and only then we can install syslog-ng.

@MrAnno top be completely honest, all above mentioned packets are included in every Debian derrativative by default. So there is no need to install them. Hence this line could be safely deleted.

MrAnno: > @MrAnno top be completely honest, all above mentioned packets are included in every Debian derrativative by default.

I don't think that's true for the `debian:testing` Docker image, but you can easily check it.

> And also: apt uses built in function to add repo keys. This function does not use wget per se. It uses function thaty is similar in functionality.

Feel free to fix it, if you don't like the current implementation.

MrAnno: So to clarify:
The current version of this pull request breaks our Docker image as syslog-ng is installed _before_ adding `ose-repo.syslog-ng.com` to `/etc/apt/sources.list.d`.

The intention here is to install syslog-ng from our own APT repository, where the latest version is available.

bazsi: @wojtekxtx would you like to continue working on this PR? The combination of the three EXPOSE commands into one makes sense, but the rest breaks our build process. If you don't, I might just merge the expose related changes and then close this PR.

bazsi:
Closing because of inactivity.

PR #4391, opened 2023-03-23
python-modules: raise NotImplementedError in mandatory abstract methods

MrAnno:


modules/python-modules/syslogng/dest.py
alltilla, line 1: [optional]

Not related to this PR, but we missed to show these two functions: `generate_persist_name()` and `flush()`.

modules/python-modules/syslogng/source.py
alltilla, line 1: [optional]

Not related to this PR, but we missed to show these `LogSource` functions: `deinit()`, `suspend()`, `wakeup()`, `generate_persist_name()`.

Also these `LogFetcher` functions: `deinit()`, `request_exit()`, `generate_persist_name()`.

modules/python-modules/syslogng/dest.py
```diff
@@ -103,4 +103,4 @@ def send(self, msg):
         Returns:
             int: one value from the LogDestinationResult enum
         """
-        return self.SUCCESS
+        raise NotImplementedError
```
alltilla, line 106: [optional]

While we are here, we could descend from `ABC`, to show these error in init time:
```diff
--- a/modules/python-modules/syslogng/dest.py
+++ b/modules/python-modules/syslogng/dest.py
@@ -20,6 +20,7 @@
 # COPYING for details.
 #
 #############################################################################
+from abc import ABC, abstractmethod
 try:
     from _syslogng import LogDestination, LogDestinationResult

@@ -39,7 +40,7 @@ except ImportError:
         NOT_CONNECTED = auto(),


-class LogDestination(LogDestination):
+class LogDestination(ABC, LogDestination):
     DROP = LogDestinationResult.DROP
     ERROR = LogDestinationResult.ERROR
     SUCCESS = LogDestinationResult.SUCCESS
@@ -91,6 +92,7 @@ class LogDestination(LogDestination):
         """
         pass

+    @abstractmethod
     def send(self, msg):
         """Send a message to the target service
```

alltilla (dismissed):
[optional]

`template.py` is missing its base class implementation. :(


---

I only had optional, nice to have comments, LGTM.

modules/python-modules/syslogng/dest.py
MrAnno: I would rather avoid multiple inheritance. I had a quick look at metaclasses too...

modules/python-modules/syslogng/dest.py
MrAnno: Adding `generate_persist_name()` is currently not simple as the existence of this method disables the C-side persist name generation.

`flush()` would have some performance penalty.

modules/python-modules/syslogng/source.py
MrAnno: `suspend(), wakeup(), generate_persist_name()` would alter default functionality.

I'm adding `deinit()`, thank you.

alltilla (approved):

modules/python-modules/syslogng/source.py
alltilla: Oh, I did not know this. Thanks for the heads-up and adding `deinit()` and `request_exit()`.

PR #4395, opened 2023-03-25
Extract cfg-persist into separate module

bazsi:
As a preparation for improving metric labeling I started this refactorm which can be submitted separately.

### DIFF
--- a/lib/cfg.c
+++ b/lib/cfg.c
@@ -52,46 +52,6 @@
 #include <stdlib.h>
 #include <iv_work.h>
 
-/* PersistConfig */
-
-struct _PersistConfig
-{
-  GHashTable *keys;
-};
-
-typedef struct _PersistConfigEntry
-{
-  gpointer value;
-  GDestroyNotify destroy;
-} PersistConfigEntry;
-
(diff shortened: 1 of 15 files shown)


lib/cfg.c
```diff
@@ -764,56 +724,21 @@ void
...
+      if (destroy)
+        destroy(value);
       return;
     }
-  else if (destroy && value)
-    {
-      destroy(value);
-    }
-  return;
+  persist_config_add(cfg->persist, name, value, destroy, force);
```
MrAnno: I think this changes functionality.
We should call `persist_config_add()` only if we have both persist and a value.
We we shouldn't store NULL values or when both value and persist is NULL.

bazsi: I don't think we would ever attempt to store a null value, but if we ever do it doesn't matter much as it fetch will return null in both cases. But you are right I'm fixing this

bazsi: I have now added to drop NULL explicitly in the cfg.c layer so this should be fixed.

MrAnno (approved):

PR #4396, opened 2023-03-27
json-parser: fix parsing non-string arrays 

### LINKED ISSUE #4387: json-parser parses array of integers to list of strings
# syslog-ng
## Version of syslog-ng
4.1.1

# Issue
## Failure
`json-parser()` works in a way that if it can parse the JSON data to a compatible syslog-ng key-value type, it should do that, otherwise it should save it as a json-literal.

A syslog-ng list is always a list of strings. An array of integers cannot be stored as a syslog-ng list, so it should be stored as a json literal. syslog-ng instead parses it to a list of strings.

## Steps to reproduce
Use the config below.

## Configuration
```
@version: 4.1
@include "scl.conf"

log {
  source {
    example-msg-generator(
      num(1)
      template('{"x": [1, 2, 3, 4]}}')
    );
  };

  parser {
    json-parser(prefix(".json."));
  };

  destination {
    file (
      "/dev/stdout"
      template("$(format-json .json.*)\n")
    );
  };
};
```

## Input and output logs
```
[2023-03-23T13:36:42.617606] json-parser message processing started; input='{"x": [1, 2, 3, 4]}}', prefix='.json.', marker='(null)', msg='0x55ca088353a0', rcptid='3'
[2023-03-23T13:36:42.617648] Setting value; name='.json.x', value='1,2,3,4', type='list', msg='0x55ca088353a0', rcptid='3'
...
{"_json":{"x":["1","2","3","4"]}}
```

alltilla:
Fixes #4387.

### DIFF
--- a/modules/json/json-parser.c
+++ b/modules/json/json-parser.c
@@ -205,14 +205,12 @@ json_parser_extract_values_from_complex_json_object(JSONParser *self,
       for (gint i = 0; i < json_object_array_length(jso); i++)
         {
           struct json_object *el = json_object_array_get_idx(jso, i);
-          GString *element_value = scratch_buffers_alloc();
-          LogMessageValueType element_type;
-
-          if (json_parser_extract_string_from_simple_json_object(self, el, element_value, &element_type))
+          if (json_object_get_type(el) == json_type_string)
             {
+              const gchar *element_value = json_object_get_string(el);
               if (i != 0)
                 g_string_append_c(value, ',');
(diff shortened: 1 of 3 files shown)


bazsi (approved):
I was using the string based arrays as we don't have simple accessors for JSON content at the moment. (e.g. a $(jq) template function...)

But you are right, if the content of the array is non-string based, we would be changing the types of the elements as we reproduce the output JSON.

So this shuold be done this way instead, it's just we should do something with the JSON accessor functionality.

MrAnno (approved):

PR #4401, opened 2023-03-28
timeutils: fix strptime(%z) when no Daylight Saving Time is specified

MrAnno:
From libc docs:
> If Daylight Saving Time is never used, tzname[1] is the empty string.

Empty strings should not be compared against the input of strptime().


kira-syslogng:
Build FAILURE

alltilla (dismissed):

PR #4399, opened 2023-03-27
disk-buffer: add dir metrics

alltilla:
Example metrics:
```
syslogng_disk_buffer_dir_available_space_bytes{dir="/var/syslog-ng"} 870109413376
```

Signed-off-by: Attila Szakacs <attila.szakacs@axoflow.com>

PR #4402, opened 2023-03-28
disk-buffer: add metrics for abandoned disk-buffer files

### LINKED ISSUE #4400: another SLES 12 post build check fail
Yet another SLES 12 post build check fail.

```
[   73s] I: Program is using implicit definitions of special functions.
[   73s]    these functions need to use their correct prototypes to allow
[   73s]    the lightweight buffer overflow checking to work.
[   73s]      - Implicit memory/string functions need #include <string.h>.
[   73s]      - Implicit *printf functions need #include <stdio.h>.
[   73s]      - Implicit *printf functions need #include <stdio.h>.
[   73s]      - Implicit *read* functions need #include <unistd.h>.
[   73s]      - Implicit *recv* functions need #include <sys/socket.h>.
[   73s] E: syslog-ng implicit-fortify-decl lib/stats/stats-cluster-key-builder.c:157
```
The good news is, that both RHEL 7 and SLES 12 will be end of life in 2024 :-)

alltilla:
Availability is the same as the `disk_queue_dir_available_bytes` metric.

Example metrics:
```
syslogng_disk_queue_capacity_bytes{abandoned="true",path="/var/syslog-ng/syslog-ng-00000.rqf",reliable="true"} 104853504
syslogng_disk_queue_disk_allocated_bytes{abandoned="true",path="/var/syslog-ng/syslog-ng-00000.rqf",reliable="true"} 273408
syslogng_disk_queue_disk_usage_bytes{abandoned="true",path="/var/syslog-ng/syslog-ng-00000.rqf",reliable="true"} 269312
syslogng_disk_queue_events{abandoned="true",path="/var/syslog-ng/syslog-ng-00000.rqf",reliable="true"} 860
```

Depends on #4399 and #4356.

Fixes #4400

### DIFF
--- a/lib/stats/stats-registry.c
+++ b/lib/stats/stats-registry.c
@@ -356,6 +356,31 @@ stats_get_cluster(const StatsClusterKey *sc_key)
 }
 
 gboolean
+stats_remove_cluster(const StatsClusterKey *sc_key)
+{
+  g_assert(stats_locked);
+  StatsCluster *sc;
+
+  sc = g_hash_table_lookup(stats_cluster_container.dynamic_clusters, sc_key);
+  if (sc)
+    {
+      if (stats_cluster_is_orphaned(sc))
+        return g_hash_table_remove(stats_cluster_container.dynamic_clusters, sc_key);
+      return FALSE;
+    }
+
+  sc = g_hash_table_lookup(stats_cluster_container.static_clusters, sc_key);
+  if (sc)
+    {
+      if (stats_cluster_is_orphaned(sc))
+        return g_hash_table_remove(stats_cluster_container.static_clusters, sc_key);
+      return FALSE;
+    }
+
+  return FALSE;
+}
+
+gboolean
 stats_contains_counter(const StatsClusterKey *sc_key, gint type)
 {
   g_assert(stats_locked);
--- a/modules/diskq/diskq-global-metrics.c
+++ b/modules/diskq/diskq-global-metrics.c
@@ -202,22 +202,6 @@ diskq_global_metrics_init(void)
 }
 
 void
-diskq_global_metrics_watch_dir(const gchar *dir)
-{
-  DiskQGlobalMetrics *self = &diskq_global_metrics;
-
-  g_mutex_lock(&self->lock);
-  {
-    if (!g_hash_table_contains(self->dirs, dir))
-      {
-        GHashTable *tracked_files = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
-        g_hash_table_insert(self->dirs, g_strdup(dir), tracked_files);
-      }
-  }
-  g_mutex_unlock(&self->lock);
-}
-
-void
 diskq_global_metrics_file_acquired(const gchar *abs_filename)
 {
   DiskQGlobalMetrics *self = &diskq_global_metrics;
@@ -228,7 +212,11 @@ diskq_global_metrics_file_acquired(const gchar *abs_filename)
   g_mutex_lock(&self->lock);
   {
     GHashTable *tracked_files = g_hash_table_lookup(self->dirs, dir);
-    g_assert(tracked_files);
+    if (!tracked_files)
+      {
+        tracked_files = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
+        g_hash_table_insert(self->dirs, g_strdup(dir), tracked_files);
+      }
 
     _track_acquired_file(tracked_files, filename);
   }
--- a/modules/diskq/diskq-global-metrics.c
+++ b/modules/diskq/diskq-global-metrics.c
@@ -63,6 +63,24 @@ _dir_watch_timer_start(DiskQGlobalMetrics *self)
   iv_timer_register(&self->dir_watch_timer);
 }
 
+static void
+_track_acquired_file(GHashTable *tracked_files, const gchar *filename)
+{
+  g_hash_table_insert(tracked_files, g_strdup(filename), GINT_TO_POINTER((gint) TRUE));
+}
+
+static void
+_track_released_file(GHashTable *tracked_files, const gchar *filename)
+{
+  g_hash_table_insert(tracked_files, g_strdup(filename), GINT_TO_POINTER((gint) FALSE));
+}
+
+static gboolean
+_is_file_tracked(GHashTable *tracked_files, const gchar *filename)
+{
+  return g_hash_table_contains(tracked_files, filename);
+}
+
 static gboolean
 _get_available_space_mib_in_dir(const gchar *dir, gint64 *available_space_mib)
 {
@@ -128,7 +146,7 @@ _new(gint type, gpointer c)
   self->dir_watch_timer.handler = _update_all_dir_metrics;
   self->dir_watch_timer.cookie = self;
 
-  self->dirs = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
+  self->dirs = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, (GDestroyNotify) g_hash_table_destroy);
 }
 
 static void
@@ -191,7 +209,52 @@ diskq_global_metrics_watch_dir(const gchar *dir)
   g_mutex_lock(&self->lock);
   {
     if (!g_hash_table_contains(self->dirs, dir))
-      g_hash_table_insert(self->dirs, g_strdup(dir), NULL);
+      {
+        GHashTable *tracked_files = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
+        g_hash_table_insert(self->dirs, g_strdup(dir), tracked_files);
+      }
   }
   g_mutex_unlock(&self->lock);
+}
+
+void
+diskq_global_metrics_file_acquired(const gchar *abs_filename)
+{
+  DiskQGlobalMetrics *self = &diskq_global_metrics;
+
+  gchar *dir = g_path_get_dirname(abs_filename);
+  gchar *filename = g_path_get_basename(abs_filename);
+
+  g_mutex_lock(&self->lock);
+  {
+    GHashTable *tracked_files = g_hash_table_lookup(self->dirs, dir);
+    g_assert(tracked_files);
+
+    _track_acquired_file(tracked_files, filename);
+  }
+  g_mutex_unlock(&self->lock);
+
+  g_free(filename);
+  g_free(dir);
+}
+
+void
+diskq_global_metrics_file_released(const gchar *abs_filename)
+{
+  DiskQGlobalMetrics *self = &diskq_global_metrics;
+
+  gchar *dir = g_path_get_dirname(abs_filename);
+  gchar *filename = g_path_get_basename(abs_filename);
+
+  g_mutex_lock(&self->lock);
+  {
+    GHashTable *tracked_files = g_hash_table_lookup(self->dirs, dir);
+    g_assert(tracked_files);
+
+    _track_released_file(tracked_files, filename);
+  }
+  g_mutex_unlock(&self->lock);
+
+  g_free(filename);
+  g_free(dir);
 }
--- a/modules/diskq/diskq-global-metrics.c
+++ b/modules/diskq/diskq-global-metrics.c
@@ -318,6 +318,42 @@ _update_all_dir_metrics(gpointer s)
 }
 
 static void
+_unset_dir_metrics(const gchar *dir)
+{
+  StatsClusterKey available_bytes_sc_key;
+  _init_dir_sc_keys(&available_bytes_sc_key, dir);
+
+  stats_lock();
+  {
+    stats_remove_cluster(&available_bytes_sc_key);
+  }
+  stats_unlock();
+}
+
+static void
+_unset_abandoned_disk_buffer_file_metrics_foreach_fn(gpointer key, gpointer value, gpointer user_data)
+{
+  const gchar *filename = (const gchar *) key;
+  gboolean acquired = (gboolean) GPOINTER_TO_INT(value);
+  const gchar *dir = (const gchar *) user_data;
+
+  if (acquired)
+    return;
+
+  _unset_abandoned_disk_buffer_file_metrics(dir, filename);
+}
+
+static void
+_unset_all_metrics_in_dir(gpointer key, gpointer value, gpointer user_data)
+{
+  gchar *dir = (gchar *) key;
+  GHashTable *tracked_files = (GHashTable *) value;
+
+  _unset_dir_metrics(dir);
+  g_hash_table_foreach(tracked_files, _unset_abandoned_disk_buffer_file_metrics_foreach_fn, dir);
+}
+
+static void
 _new(gint type, gpointer c)
 {
   DiskQGlobalMetrics *self = &diskq_global_metrics;
@@ -354,6 +390,7 @@ _deinit(gint type, gpointer c)
 
   g_mutex_lock(&self->lock);
   {
+    g_hash_table_foreach(self->dirs, _unset_all_metrics_in_dir, NULL);
     g_hash_table_remove_all(self->dirs);
   }
   g_mutex_unlock(&self->lock);
--- a/modules/diskq/diskq-global-metrics.c
+++ b/modules/diskq/diskq-global-metrics.c
@@ -29,9 +29,14 @@
 #include "stats/stats-registry.h"
 #include "stats/stats-cluster-single.h"
 #include "timeutils/misc.h"
+#include "qdisk.h"
 
 #include <iv.h>
+#include <dirent.h>
+#include <sys/stat.h>
 #include <sys/statvfs.h>
+#include <sys/types.h>
+#include <unistd.h>
 
 typedef struct DiskQGlobalMetrics_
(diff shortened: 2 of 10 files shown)


kira-syslogng:
Build FAILURE

alltilla:
@kira-syslogng do stresstest

kira-syslogng:

modules/diskq/diskq-global-metrics.c
```diff
@@ -169,6 +173,66 @@ _init_abandoned_disk_buffer_sc_keys(StatsClusterKey *queued_sc_key, StatsCluster
...
+  options.read_only = TRUE;
+
+  LogQueue *queue;
+  if (reliable)
+    queue = log_queue_disk_reliable_new(&options, abs_filename, NULL, STATS_LEVEL0, NULL, NULL);
+  else
+    queue = log_queue_disk_non_reliable_new(&options, abs_filename, NULL, STATS_LEVEL0, NULL, NULL);
+  LogQueueDisk *queue_disk = (LogQueueDisk *) queue;
+
+  if (!log_queue_disk_start(queue))
```
MrAnno, line 197: Questions:
I'm wondering how expensive this call can be.

Is it okay to enable this by default? If so, is it OK to repeat it every 5 minutes?
A dir might have thousands of abandoned disk-buffer files (or not-abandoned, or just regular files).

This feature implies that 2 syslog-ng instances can not use the same disk-buffer directory, otherwise they will report each others' buffers as abandoned.
I think this is good behavior, but considering the above, I think it is unnecessary to do periodic abandoned buffer checks, because no one will change those values between an acquire/release cycle.

TLDR: I think we should remove the periodic abandoned buffer metrics update from the timer callback, and call it only when a new directory is added by `acquire`. This would keep this potentially expensive call at startup time.
What about buffers that were abandoned during runtime? I think it's not possible unless they are corrupted, but we should not try to open corrupted buffer files in the main syslog-ng process. Reloading might abandon buffers, but that will be detected in `acquire`.

MrAnno (commented):
Nice feature.

For the record: `diskq-global-metrics` considers buffers abandoned when they are not used in the currently running syslog-ng instance. This is a bit different than what we do in our persist file: we abandon buffers only when 2 consecutive syslog-ng restart does not acquire them.

I think this is not a real problem, I just wanted to mention it.

modules/diskq/logqueue-disk-non-reliable.c
```diff
@@ -169,6 +170,7 @@ _move_messages_from_disk_to_qout(LogQueueDiskNonReliable *self)
       g_queue_push_tail(self->qout, msg);
       g_queue_push_tail(self->qout, LOG_PATH_OPTIONS_TO_POINTER(&path_options));
       log_queue_memory_usage_add(&self->super.super, log_msg_get_size(msg));
+      log_queue_disk_update_disk_related_counters(&self->super);
```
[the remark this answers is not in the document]
alltilla: Yes, It will when I rebase, sorry for that.

modules/diskq/diskq-global-metrics.c
```diff
@@ -103,6 +105,32 @@ _update_dir_available_space_metrics(const gchar *dir)
   stats_unlock();
 }
 
+static gboolean
+_is_disk_buffer_file(const gchar *filename)
+{
+  const gsize min_len = sizeof("syslog-ng-00000.qf") - 1;
+  const gsize prefix_len = sizeof("syslog-ng-") - 1;
+  const gsize extension_start = sizeof("syslog-ng-00000") - 1;
```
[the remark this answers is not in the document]
alltilla: It is just an optimization, to calculate these compile time. I can change it to strlen, would you like me to do so for the sake of readability?

modules/diskq/diskq-global-metrics.c
alltilla: I agree. The only case where this module does not get notified about a new disk-buffer file is corruption. The code currently tries to open the corrupted file and if it succeeds it creates metrics for it.
Two things:
  1. We can call the `diskq_global_metrics_file_released()` fn in the corrupted restart fn, so we don't need the poll at all.
  2. As you mentioned, we might not want to create metrics for the corrupted files. But I would ask you what is the reasoning behind it? I did not see a problem with it.

modules/diskq/diskq-global-metrics.c
```diff
@@ -197,7 +197,63 @@ diskq_global_metrics_watch_dir(const gchar *dir)
...
+  gchar *filename = g_path_get_basename(abs_filename);
+
+  g_mutex_lock(&self->lock);
+  {
+    GHashTable *acquired_files = g_hash_table_lookup(self->dirs, dir);
+    if (!acquired_files)
+      {
+        g_mutex_unlock(&self->lock);
+        g_free(filename);
+        goto exit;
```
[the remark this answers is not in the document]
MrAnno: If you don't want to force-watch, a `g_assert_not_reached()` might save us some future trouble.

modules/diskq/diskq-global-metrics.c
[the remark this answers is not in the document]
MrAnno: No, I was just curious. Thanks.

modules/diskq/diskq-global-metrics.c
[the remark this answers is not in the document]
alltilla: I started with that and the unit tests crashed on that, because they don't call watch, but call acquire_queue. I can add a watch to the unit tests, too, but it would seem strange IMO.

modules/diskq/diskq-global-metrics.c
MrAnno: I'm afraid a corrupted disk-buffer file may crash the syslog-ng instance. Also, corrupt files may have corrupt headers, therefore invalid entries.

I know... Every crash possibility should be fixed.
But if we could avoid doing all of this at runtime (moving them to startup time only), I think it would be much safer and cheaper. A metric, after all, should not affect performance or stability.

I'd vote for 2), but if you insist, 1) is good too.

modules/diskq/diskq-global-metrics.c
```diff
@@ -105,6 +113,205 @@ _update_dir_metrics(gpointer key, gpointer value, gpointer user_data)
...
+    return FALSE;
+
+  for (gint i = prefix_len; i < extension_start; i++)
+    {
+      if (filename[i] < '0' || filename[i] > '9')
+        return FALSE;
+    }
+
+  if (strncmp(&filename[extension_start], ".qf", 3) != 0 && strncmp(&filename[extension_start], ".rqf", 4) != 0)
+    return FALSE;
```
MrAnno, line 137: Sorry, another note because I did not notice:

I think we could be a bit stricter by not allowing any suffixes.
For example, if we don't want to check corrupt files, we should not allow a suffix, otherwise we should look for `corrupted`.

modules/diskq/diskq-global-metrics.c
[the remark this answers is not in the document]
MrAnno: From a clean-code perspective, if the test does not call watch() but only calls acquire(), that's an invalid usage of the API and the test should probably catch that.

it's either we do a force-watch, or we don't allow calling this method without a previous watch() call. With the current code, calling `diskq_global_metrics_file_acquired()` without a previous watch() call will misfunction.

modules/diskq/diskq-global-metrics.c
MrAnno: Ahh, corrupted buffers are also numbered... Okay... It's not worth all the trouble.

modules/diskq/diskq-global-metrics.c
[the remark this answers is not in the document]
alltilla: I agree, fixed.

modules/diskq/diskq-global-metrics.c
```diff
@@ -103,6 +105,32 @@ _update_dir_available_space_metrics(const gchar *dir)
   stats_unlock();
 }
 
+static gboolean
+_is_disk_buffer_file(const gchar *filename)
```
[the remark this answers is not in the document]
alltilla: Moved to `qdisk.c`.

modules/diskq/diskq-global-metrics.c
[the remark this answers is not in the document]
alltilla: In the end I have changed it, as the new code would have been a lot more complicated.

modules/diskq/diskq-global-metrics.c
```diff
@@ -103,6 +105,32 @@ _update_dir_available_space_metrics(const gchar *dir)
...
+  gsize len = strlen(filename);
+  if (len < min_len)
+    return FALSE;
+
+  if (strncmp(filename, "syslog-ng-", prefix_len) != 0)
+    return FALSE;
+
+  for (gint i = prefix_len; i < extension_start; i++)
+    {
+      if (filename[i] < '0' || filename[i] > '9')
```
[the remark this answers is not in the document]
alltilla: Good idea, I changed it.

modules/diskq/diskq-global-metrics.c
```diff
@@ -131,12 +132,93 @@ _is_disk_buffer_file(const gchar *filename)
   return TRUE;
 }
 
+static gboolean
+_is_reliable(const gchar *filename)
+{
+  const gsize extension_start = sizeof("syslog-ng-00000") - 1;
```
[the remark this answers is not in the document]
alltilla: As this function moved to `qdisk.c` I have modified this to be able to return with error. I have added this check, too.

modules/diskq/diskq-global-metrics.c
alltilla: I have added a check to skip the corrupted files.

modules/diskq/diskq-global-metrics.c
alltilla: Yes :/ I have kept it like this.

alltilla:
Rebased to master and fixed all the comments I think. I have not tested it manually yet, but I will do that tomorrow.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

modules/diskq/diskq-global-metrics.c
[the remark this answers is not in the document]
MrAnno: Thanks

alltilla:
Tested manually, the behavior looks good to me.

lib/stats/stats-registry.c
```diff
@@ -355,6 +355,15 @@ stats_get_cluster(const StatsClusterKey *sc_key)
...
 
+gboolean
+stats_remove_cluster(const StatsClusterKey *sc_key)
+{
+  g_assert(stats_locked);
+
+  return g_hash_table_remove(stats_cluster_container.dynamic_clusters, sc_key) ||
+         g_hash_table_remove(stats_cluster_container.static_clusters, sc_key);
+}
+
```
alltilla, line 366: @MrAnno Should the stats API only allow removing if the cluster is abandoned?

alltilla: Orphaned* ...

alltilla: I will fix this. Is it okay to do in a separate PR, if you find this one good except this?

modules/diskq/diskq-global-metrics.h
```diff
@@ -27,5 +27,7 @@
 
 void diskq_global_metrics_init(void);
```
MrAnno, line 28: I've checked and thanks to `disk_buffer_module_init()`, global metrics are not initialized when no disk-buffer is used in the configuration. 👍🏻

MrAnno:
I've reproduced a crash:
```
**
ERROR:../modules/diskq/diskq-global-metrics.c:373:diskq_global_metrics_file_acquired: assertion failed: (acquired_files)
Bail out! ERROR:../modules/diskq/diskq-global-metrics.c:373:diskq_global_metrics_file_acquired: assertion failed: (acquired_files)
fish: Job 1, 'sbin/syslog-ng -Fedv' terminated by signal SIGABRT (Abort)
```

Reproduction steps:
-  start syslog-ng with `network(127.0.0.1 port(4444) disk-buffer(dir("/tmp/asd") disk-buf-size(10M)));`
- change dir and reload:  `network(127.0.0.1 port(4444) disk-buffer(dir("/tmp/asd2") disk-buf-size(10M)));`

MrAnno:
The issue persist even after restarts. I was wrong about the persist vs. global metrics diff:
If we open a disk-buffer file in a directory that is recorded in the persist file but not in the configuration, we'll need a force-watch.

modules/diskq/diskq-global-metrics.c
```diff
@@ -63,6 +72,156 @@ _dir_watch_timer_start(DiskQGlobalMetrics *self)
...
+  options.read_only = TRUE;
+
+  LogQueue *queue;
+  if (reliable)
+    queue = log_queue_disk_reliable_new(&options, abs_filename, NULL, STATS_LEVEL0, NULL, NULL);
+  else
+    queue = log_queue_disk_non_reliable_new(&options, abs_filename, NULL, STATS_LEVEL0, NULL, NULL);
+  LogQueueDisk *queue_disk = (LogQueueDisk *) queue;
+
+  if (!log_queue_disk_start(queue))
```
MrAnno, line 116: This has some `internal()` side effects on the info level:
```
[2023-04-05T11:57:37.812272] Disk-buffer state loaded; filename='/tmp/asd/syslog-ng-00abandoned000.qf', number_of_messages='0'
```

I don't think it's a problem. Please just click "resolve" if you agree.

modules/diskq/diskq-global-metrics.c
MrAnno: Thank you.

lib/stats/stats-registry.c
MrAnno: As you see fit.

MrAnno:
I've found one (hopefully) last thing:

Now that we know that we need to watch (or force-watch) directories coming NOT from the config, but from persisted disk-buffer locations, the following two cases became more valid:
- a directory is removed
  `disk-buffer: Failed to list files in dir; dir='/tmp/asd2', error='No such file or directory (2)'`
  `disk-buffer: Failed to get filesystem info; dir='/tmp/asd2', error='No such file or directory (2)'`
- a file is removed: nothing happens but we don't remove abandoned counters (on reload I think we should)

And a more interesting case; when a new directory is added to the config, but because of the persist file, we still use the old location:
```
[2023-04-05T12:21:41.035721] The disk buffer directory has changed in the configuration, but the disk queue file cannot be moved; qfile='/tmp/asd/syslog-ng-00001.qf', dir='/tmp/asd2'
[2023-04-05T12:21:41.036112] disk-buffer: Failed to get filesystem info; dir='/tmp/asd2', error='No such file or directory (2)'
```

lib/stats/stats-registry.c
alltilla: Fixed.

alltilla:
@kira-syslogng do stresstest

alltilla:
I have loosened an assert in `QDisk`, as I am a bit paranoid that if the file gets emptied by something external in a bad split second, we could fail that assert.

kira-syslogng:

alltilla:
@kira-syslogng do stresstest

kira-syslogng:

MrAnno:
I'm investigating a SIGABRT in disk-buffer when config initialization is not successful (I'm not sure this is related to this PR).

modules/diskq/qdisk.c
```diff
@@ -278,6 +290,35 @@ qdisk_get_next_filename(const gchar *dir, gboolean reliable)
...
+{
+  const gsize min_len = MIN(strlen(QDISK_FILENAME_REL_EXAMPLE), strlen(QDISK_FILENAME_NON_REL_EXAMPLE));
+  const gsize prefix_len = strlen(QDISK_FILENAME_PREFIX);
+  const gsize extension_start = strlen(QDISK_FILENAME_PREFIX QDISK_FILENAME_IDX_EXAMPLE);
+
+  gsize len = strlen(filename);
+  if (len < min_len)
+    return FALSE;
+
+  if (strncmp(filename, "syslog-ng-", prefix_len) != 0)
```
MrAnno, line 304: Opt: `QDISK_FILENAME_PREFIX`

lib/stats/stats-registry.c
```diff
@@ -355,6 +355,31 @@ stats_get_cluster(const StatsClusterKey *sc_key)
...
+  sc = g_hash_table_lookup(stats_cluster_container.static_clusters, sc_key);
+  if (sc)
+    {
+      if (stats_cluster_is_orphaned(sc))
+        return g_hash_table_remove(stats_cluster_container.static_clusters, sc_key);
+      return FALSE;
+    }
+
+  return FALSE;
```
MrAnno, line 380: `return TRUE`

MrAnno (commented):
LGTM.

lib/stats/stats-registry.c
MrAnno: my bad

alltilla: For reference, this function mimics the return value of https://docs.gtk.org/glib/type_func.HashTable.remove.html#return-value

Issue #4403, opened 2023-03-28
patterndb inconsistent type hints for different parts of xml file 


erempel:
It seems that the patterndb-6.xsd requires that in the <values> section that the integer type hints be specified with the types of int32 or int64 while in the in the <examples> <test_values> the type hing must use the type of integer.

         <rule id="FLARE-3698" class="zmwatch" provider="UVic">
            <patterns>
               <pattern>INF ['zma -m @NUMBER:camera@' starting at @ESTRING:time:,@ pid = @NUMBER:pid@]</pattern>
            </patterns>
            <values>
               <value name="PATTERNID" type="int64">3698</value>
            </values>
            <examples>
               <example>
                  <test_message program="zmwatch">INF ['zma -m 13' starting at 14/11/16 21:28:13, pid = 27714]</test_message>
                  <test_values>
                     <test_value name="time" type="string">14/11/16 21:28:13</test_value>
                     <test_value name="pid" type="integer">27714</test_value>
                     <test_value name="camera" type="integer">13</test_value>
                  </test_values>
               </example>
            </examples>
         </rule>

bazsi:
Hmm those should each be aliases for each other.

erempel:
If I try to use integer type hint in the values section I get the error

[xerr@pangolin bin]$ /bin/pdbtool test --validate /tmp/patterns.xml
Error parsing pattern database file; filename='/tmp/patterns.xml', error='/tmp/patterns.xml:14:60: Error compiling value template, rule=FLARE-3161, name=PATTERNID, value=3161, error=Unknown type specified in type hinting: integer'

bazsi:
I have found the issue, sorry for broking this. Submitting a PR in a moment.

PR #4405, opened 2023-03-29
Fix type validation in "pdbtool test"

bazsi:
`pdbtool test`: fix two type validation bugs:

  1) When `pdbtool test` validates the type information associated with a name-value
     pair, it was using string comparisons, which didn't take type aliases
     into account. This is now fixed, so that "int", "integer" or "int64"
     can all be used to mean the same type.

  2) When type information is missing from a `<test_value/>` tag, don't
     validate it against "string", rather accept any extracted type.

In addition to these fixes, a new alias "integer" was added to mean the same
as "int", simply because syslog-ng was erroneously using this term when
reporting type information in its own messages.

### DIFF
--- a/lib/logmsg/logmsg.c
+++ b/lib/logmsg/logmsg.c
@@ -241,7 +241,7 @@ log_msg_value_type_to_str(LogMessageValueType self)
     [LM_VT_JSON] = "json",
     [LM_VT_BOOLEAN] = "boolean",
     [__COMPAT_LM_VT_INT32] = "compat-int32",
-    [LM_VT_INTEGER] = "integer",
+    [LM_VT_INTEGER] = "int",
     [LM_VT_DOUBLE] = "double",
     [LM_VT_DATETIME] = "datetime",
     [LM_VT_LIST] = "list",
@@ -261,7 +261,8 @@ log_msg_value_type_from_str(const gchar *in_str, LogMessageValueType *out_type)
     *out_type = LM_VT_JSON;
   else if (strcmp(in_str, "boolean") == 0)
     *out_type = LM_VT_BOOLEAN;
-  else if (strcmp(in_str, "int32") == 0 || strcmp(in_str, "int") == 0 || strcmp(in_str, "int64") == 0)
+  else if (strcmp(in_str, "int32") == 0 || strcmp(in_str, "int") == 0 || strcmp(in_str, "int64") == 0
+           || strcmp(in_str, "integer") == 0)
     *out_type = LM_VT_INTEGER;
   else if (strcmp(in_str, "double") == 0 || strcmp(in_str, "float") == 0)
     *out_type = LM_VT_DOUBLE;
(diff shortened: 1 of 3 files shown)


smortex:
It seems that 8c2ebe34ee6774a3e68af4057aec5099fbcbd3e5 incorrectly says "Fixes: #4403" while the fix seems to be in f19dec039b6eeb79d785497b6a8fc84f382c07d1.

Regarding 8c2ebe34ee6774a3e68af4057aec5099fbcbd3e5, I _feel_ like I used this "feature" to detect examples I had forgotten to update while migrating, so ignoring the data type would mean I would not have catch them maybe.  This looks odd, but maybe I am wrong.  Just wanted to raise this concern in case it is legit :smile:

bazsi:
> It seems that [8c2ebe3](https://github.com/syslog-ng/syslog-ng/commit/8c2ebe34ee6774a3e68af4057aec5099fbcbd3e5) incorrectly says "Fixes: #4403" while the fix seems to be in [f19dec0](https://github.com/syslog-ng/syslog-ng/commit/f19dec039b6eeb79d785497b6a8fc84f382c07d1).

fixed, thanks.

> 
> Regarding [8c2ebe3](https://github.com/syslog-ng/syslog-ng/commit/8c2ebe34ee6774a3e68af4057aec5099fbcbd3e5), I _feel_ like I used this "feature" to detect examples I had forgotten to update while migrating, so ignoring the data type would mean I would not have catch them maybe. This looks odd, but maybe I am wrong. Just wanted to raise this concern in case it is legit 


Thanks for the feedback. Good point.

My original has never been to make the type field required here, you might not types in your db-parser() using configuration after all, unless you are actually generating JSON or something that uses types.

What if I change the output somewhat in case no type information was supplied? so you could grep for this in the output.

Like this:
```
$  pdbtool test -v faxmodem.pdb 
Module loaded and initialized successfully; module='syslogformat'
Module loaded and initialized successfully; module='basicfuncs'
Testing message: program='zmwatch' message='INF ['zma -m 13' starting at 14/11/16 21:28:13, pid = 27714]'
 Match name='.classifier.rule_id', value='FLARE-3698', type='string', expected='FLARE-3698'
 Match name='time', value='14/11/16 21:28:13', type='string', expected='14/11/16 21:28:13'
 Match name='pid', value='27714', type='int', expected='27714'
 Match name='camera', value='13', type='int[unchecked]', expected='13'
```

note the `[unchecked]` qualifier in the output for the `camera` field. I am adding this to the branch. What do you think?

smortex:
I don't tell that the old behavior is right and the new wrong, I just wanted to emphasis the fact that I *used* the old behavior and it might have some value.

A warning message as you describe is very acceptable I think, and if everyone is okay with this I think it is great :100: !

Thank you!

alltilla (approved):

alltilla:

PR #4406, opened 2023-03-30
CMake:  Silenced some warnings

HofiOne:
- eliminated package/lib/source detection for modules that were explicitly disabled already (some modules already care about this, but these were not)
- handling now the QUIET option in libfind_process


bazsi (commented):
I didn't see where we would be using the quiet option. I don't know much about cmake, so it might be my ignorance.

The rest looks good

MrAnno (approved):

PR #4407, opened 2023-03-30
next-pr-id: account for discussions, too

alltilla:
Discussions share their IDs with issues and PRs.

I could not find any easy to use REST API for discussions, so I just brute force check whether a possible ID is occupied for a discussion.

### DIFF
--- a/news/next-pr-id.py
+++ b/news/next-pr-id.py
@@ -21,19 +21,43 @@
 #
 #############################################################################
 from urllib.request import urlopen
+from urllib.error import HTTPError
 from json import loads
 from argparse import ArgumentParser
 
 
-def get_next_pr_id():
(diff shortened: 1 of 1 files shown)


news/next-pr-id.py
```diff
@@ -21,19 +21,43 @@
...
+
+
+def is_a_discussion(id):
+    DISCUSSIONS_URL = "https://github.com/syslog-ng/syslog-ng/discussions/"
+
+    try:
+        urlopen(DISCUSSIONS_URL + str(id))
+        return True
+    except HTTPError as e:
+        if e.getcode() == 404:
```
MrAnno, line 46: This will report false information when GitHub is down (non-2xx statuses).

PR #4410, opened 2023-04-02
`python()`: new `LogMessage` methods and bugfixes

MrAnno:
`python()`: new LogMessage methods for querying as string and with default values

- `get(key[, default])`
  Return the value for `key` if `key` exists, else `default`. If `default` is
  not given, it defaults to `None`, so that this method never raises a
  `KeyError`.

- `get_as_str(key, default=None, encoding='utf-8', errors='strict', repr='internal')`:
  Return the string value for `key` if `key` exists, else `default`.
  If `default` is not given, it defaults to `None`, so that this method never
  raises a `KeyError`.

  The string value is decoded using the codec registered for `encoding`.
  `errors` may be given to set the desired error handling scheme.

  Note that currently `repr='internal'` is the only available representation.
  We may implement another more Pythonic representation in the future, so please
  specify the `repr` argument explicitly if you want to avoid future
  representation changes in your code.

----

`python()`: fix LogMessage subscript not raising KeyError on non-existent keys

When message fields were queried (`msg["key"]`) and the given key did not exist,
`None` or an empty string was returned (depending on the version of the config).

Neither was correct, now a KeyError occurs in such cases.

----

`python()`: fix exception handling when LogMessage value conversion fails

----

`$(python)`: fix template function prefix being overwritten when using datetime types

### DIFF
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -66,7 +66,7 @@ _py_log_message_subscript(PyObject *o, PyObject *key)
       return NULL;
     }
 
-  if (py_msg->cast_to_strings)
+  if (py_msg->cast_to_bytes)
     type = LM_VT_STRING;
 
   APPEND_ZERO(value, value, value_len);
@@ -101,7 +101,7 @@ _py_log_message_ass_subscript(PyObject *o, PyObject *key, PyObject *value)
   if (!value)
     return -1;
 
-  if (py_msg->cast_to_strings && !is_py_obj_bytes_or_string_type(value))
+  if (py_msg->cast_to_bytes && !is_py_obj_bytes_or_string_type(value))
     {
       PyErr_Format(PyExc_ValueError,
                    "str or unicode object expected as log message values, got type %s (key %s). "
@@ -149,9 +149,9 @@ py_log_message_new(LogMessage *msg, GlobalConfig *cfg)
   self->bookmark_data = NULL;
 
   if (cfg_is_config_version_older(cfg, VERSION_VALUE_4_0))
-    self->cast_to_strings = TRUE;
+    self->cast_to_bytes = TRUE;
   else
-    self->cast_to_strings = FALSE;
+    self->cast_to_bytes = FALSE;
   return (PyObject *) self;
 }
 
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -44,6 +44,25 @@ py_is_log_message(PyObject *obj)
   return PyType_IsSubtype(Py_TYPE(obj), &py_log_message_type);
 }
 
+static inline PyObject *
+_get_value(PyLogMessage *self, const gchar *name, gboolean cast_to_bytes)
+{
+  NVHandle handle = log_msg_get_value_handle(name);
+  gssize value_len = 0;
+  LogMessageValueType type;
+  const gchar *value = log_msg_get_value_with_type(self->msg, handle, &value_len, &type);
+
+  if (!value)
+    return NULL;
+
+  if (cast_to_bytes)
+    type = LM_VT_STRING;
+
+  APPEND_ZERO(value, value, value_len);
+
+  return py_obj_from_log_msg_value(value, value_len, type);
+}
+
 static PyObject *
 _py_log_message_subscript(PyObject *o, PyObject *key)
 {
@@ -54,24 +73,15 @@ _py_log_message_subscript(PyObject *o, PyObject *key)
       return NULL;
     }
 
-  NVHandle handle = log_msg_get_value_handle(name);
-  PyLogMessage *py_msg = (PyLogMessage *)o;
-  gssize value_len = 0;
-  LogMessageValueType type;
-  const gchar *value = log_msg_get_value_with_type(py_msg->msg, handle, &value_len, &type);
-
+  PyLogMessage *py_msg = (PyLogMessage *) o;
+  PyObject *value = _get_value(py_msg, name, py_msg->cast_to_bytes);
   if (!value)
     {
       PyErr_Format(PyExc_KeyError, "No such name-value pair %s", name);
       return NULL;
     }
 
-  if (py_msg->cast_to_bytes)
-    type = LM_VT_STRING;
-
-  APPEND_ZERO(value, value, value_len);
-
-  return py_obj_from_log_msg_value(value, value_len, type);
+  return value;
 }
 
 static int
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -104,10 +104,10 @@ _py_log_message_ass_subscript(PyObject *o, PyObject *key, PyObject *value)
   if (py_msg->cast_to_bytes && !is_py_obj_bytes_or_string_type(value))
     {
       PyErr_Format(PyExc_ValueError,
-                   "str or unicode object expected as log message values, got type %s (key %s). "
+                   "str or bytes object expected as log message values, got type %s (key %s). "
                    "Earlier syslog-ng accepted any type, implicitly converting it to a string. "
                    "Later syslog-ng (at least 4.0) will store the value with the correct type. "
-                   "With this version please convert it explicitly to a string using str()",
+                   "With this version please convert it explicitly to string/bytes",
                    value->ob_type->tp_name, name);
       return -1;
     }
--- a/modules/python/python-logmsg.c
+++ b/modules/python/python-logmsg.c
@@ -45,8 +45,9 @@ py_is_log_message(PyObject *obj)
 }
 
 static inline PyObject *
-_get_value(PyLogMessage *self, const gchar *name, gboolean cast_to_bytes)
+_get_value(PyLogMessage *self, const gchar *name, gboolean cast_to_bytes, gboolean *error)
 {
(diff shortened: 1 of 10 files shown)


MrAnno:
Note: I've fixed the `KeyError` issue in a backward-compatible manner. When using a `3.x` config version, we still produce empty strings with non-existent keys.

modules/python/python-logmsg.c
```diff
@@ -75,8 +75,9 @@ _py_log_message_subscript(PyObject *o, PyObject *key)
 
   PyLogMessage *py_msg = (PyLogMessage *) o;
   PyObject *value = _get_value(py_msg, name, py_msg->cast_to_bytes);
-  if (!value)
+  if (value == Py_None)
     {
+      Py_XDECREF(value);
```
bazsi, line 87: _get_value() can return NULL as well

modules/python/python-logmsg.c
```diff
@@ -53,14 +53,21 @@ _get_value(PyLogMessage *self, const gchar *name, gboolean cast_to_bytes)
   const gchar *value = log_msg_get_value_with_type(self->msg, handle, &value_len, &type);
 
   if (!value)
-    return NULL;
+    Py_RETURN_NONE;
```
bazsi, line 56: Ok from this patch on, I guess get_value will not return NULL without an exception being set.

alltilla:
I can review this tomorrow, but feel free to merge this without me.

modules/python/python-logmsg.c
MrAnno: Yes, I had to fix that as well.

modules/python/python-logmsg.c
MrAnno: I've refactored get_value() not to return NULL unless it sets an exception.

MrAnno:
Let's wait until @kira-syslogng reappears.

modules/python/python-logmsg.c
```diff
@@ -250,6 +250,31 @@ _logmessage_get_keys_method(PyLogMessage *self)
...
+    return NULL;
+
+  if (value != Py_None)
+    return value;
+
+  if (!default_value)
+    return value; /* returns None with the proper refcount */
+
+  Py_XINCREF(default_value);
+  return default_value;
```
alltilla, line 276: `Py_None` is a valid value for an NV pair of type `LM_VT_NULL`. I think we shouldn't fallback to `default_value` in this case.

This is the case for python `dict`s as well:
```python
>>> d = {"a": None}
>>> print(d["a"])
None
>>> print(d.get("a"))
None
>>> print(d.get("a", "fallback"))
None
>>> print(d["b"])
  File "<stdin>", line 1, in <module>
KeyError: 'b'
>>> print(d.get("b"))
None
>>> print(d.get("b", "fallback"))
fallback
```

modules/python/python-logmsg.c
```diff
@@ -82,8 +82,9 @@ _py_log_message_subscript(PyObject *o, PyObject *key)
 
   PyLogMessage *py_msg = (PyLogMessage *) o;
   PyObject *value = _get_value(py_msg, name, py_msg->cast_to_bytes);
-  if (!value)
+  if (value == Py_None)
```
alltilla, line 85: Same as my previous comment: `Py_None` is a valid value for `LM_VT_NULL`. I think we should only throw a warning if `value` is `NULL`, which means we do not have an NV pair with such a `name` or we have some other failure.

kira-syslogng:
Build FAILURE

modules/python/python-tf.c
```diff
@@ -119,10 +121,14 @@ _py_convert_return_value_to_result(PythonTfState *state, const gchar *function_n
                 evt_tag_str("exception", _py_format_exception_text(buf, sizeof(buf))));
       _py_finish_exception_handling();
 
+      scratch_buffers_reclaim_marked(marker);
       Py_DECREF(ret);
       return FALSE;
     }
 
+  g_string_append(result, value->str);
```
alltilla, line 129: Although the naming was misleading, I liked that the `python-types` functions appended to the `GString`s, exactly because of these use cases, where we now need to do copies, which could have been optimized with the appending logic.

As an alternative I would propose to:
  1. Rename the functions which append, to clearly communicate the behavior.
  2. Fix the datetime conversion to append.
  3. Modify the unit tests, to start with a non-empty `GString`, so the appending behavior is unit tested.

alltilla (commented):
I haven't finished reviewing, but sharing my comments so far.

MrAnno:

modules/python/python-logmsg.c
alltilla: Throw an exception*, sorry.

modules/python/python-logmsg.c
```diff
@@ -276,6 +276,45 @@ py_log_message_get(PyLogMessage *self, PyObject *args, PyObject *kwrds)
...
+          PyErr_Format(PyExc_TypeError, "default is not a string object");
+          return NULL;
+        }
+
+      Py_XINCREF(default_value);
+      return default_value;
+    }
+
+  APPEND_ZERO(value, value, value_len);
+  return PyUnicode_Decode(value, value_len, encoding, errors);
```
alltilla, line 315: If I understand correctly this function was originally imagined to solve the problem, that `str(b'some-bytes-data')` does not produce such elegant outputs as any other `str()` call for any other types.

I think that `PyUnicode_Decode()` should only be called when the type is `LM_VT_STRING`, for every other type we should return the same output as `str()` would, e.g.:
```python
>>> str(True)
'True'
>>> str(None)
'None'
>>> str(datetime.datetime(2022, 10, 4, 14, 48, 12, 123000))
'2022-10-04 14:48:12.123000'
```

Maybe the list could be a difference also, because we want to get rid of bytes there, too:
```python
>>> str([b'a', b'b', b'c'])
"[b'a', b'b', b'c']"
```
(`"['a', 'b', 'c']"` might look better)

I think the user is not interested in our internal representation of the values, like how we call true and false in our NV pairs, or how we store syslog-ng-like lists, but they would want to see an `str()` output for the non-bytes fields, and a unicode representation of bytes fields.

The `str()` call can be achieved easily with: https://docs.python.org/3/c-api/object.html#c.PyObject_Str

modules/python/tests/test_python_logmsg.c
```diff
@@ -256,6 +256,36 @@ Test(python_log_message, test_python_logmessage_get)
...
+  LogMessage *msg = log_msg_new_empty();
+
+  PyGILState_STATE gstate;
+  gstate = PyGILState_Ensure();
+  {
+    cfg_set_version_without_validation(configuration, VERSION_VALUE_4_0);
+    PyObject *msg_object = py_log_message_new(msg, configuration);
+    PyDict_SetItemString(_python_main_dict, "test_msg", msg_object);
+
+    log_msg_set_value_by_name_with_type(msg, "a", "14", -1, LM_VT_INTEGER);
```
alltilla, line 270: It would be nice to see cases for all of our different kinds of value types in the `get()` and `get_as_str()` unit tests.

alltilla (commented):
I have finished the review. Please consider my comments.

modules/python/python-logmsg.c
MrAnno: I'll take a look at this. My impression was that the LogMessage API does not make difference between NULL and unset values. There are signs of it, I was wrong.

MrAnno: I've fixed this. The code looks even better now.

modules/python/python-logmsg.c
MrAnno: Fixed. I had to use `log_msg_get_value_if_set_with_type()`, because `log_msg_get_value_with_type` returns a NULL-type for unset values.

modules/python/python-logmsg.c
MrAnno: So to clarify, this function does not convert a LogMessage value to a Python type and then try to decode it as a string. That would be unnecessarily expensive.

I get the LogMessage value, which is a `gchar*` (note that is can not be assumed that it is always UTF-8) and convert it to a Python string with PyUnicode_Decode() (the user must know the encoding).

I think the current behavior is correct.

modules/python/tests/test_python_logmsg.c
MrAnno: I'm adding some more test cases.

modules/python/python-logmsg.c
MrAnno: We've talked about this IRL and agreed to add a `repr='internal'` parameter to allow future implementation for a more Pythonic str method.

modules/python/python-logmsg.c
```diff
@@ -294,10 +294,11 @@ py_log_message_get_as_str(PyLogMessage *self, PyObject *args, PyObject *kwrds)
   PyObject *default_value = NULL;
   const gchar *encoding = "utf-8";
   const gchar *errors = "strict";
+  const gchar *repr = "internal";
```
alltilla, line 297: Validation is needed here, but as we agreed IRL, it can go in the followup PR.

alltilla (approved):

alltilla:
Very nice feature and bugfixes, thanks for this!

PR #4411, opened 2023-04-03
TCP connection metrics and minor improvements

MrAnno:
New metrics

`network()`, `syslog()`: TCP connection metrics

```
syslogng_socket_connections{id="tcp_src#0",driver_instance="afsocket_sd.(stream,AF_INET(0.0.0.0:5555))",direction="input"} 3
syslogng_socket_max_connections{id="tcp_src#0",driver_instance="afsocket_sd.(stream,AF_INET(0.0.0.0:5555))",direction="input"} 10
syslogng_socket_rejected_connections_total{id="tcp_src#0",driver_instance="afsocket_sd.(stream,AF_INET(0.0.0.0:5555))",direction="input"} 96
```

`internal()`: `internal_events_queue_capacity` metric

`syslog-ng-ctl healthcheck`: new healthcheck value `syslogng_internal_events_queue_usage_ratio`

The following Prometheus metrics have been renamed:

`log_path_{in,e}gress` -> `route_{in,e}gress_total`
`internal_source` -> `internal_events_total`

The `internal_queue_length` stats counter has been removed.
It was deprecated since syslog-ng 3.29.

### DIFF
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -25,6 +25,8 @@
 #include "afinter.h"
 #include "logreader.h"
 #include "stats/stats-registry.h"
+#include "stats/stats-cluster-logpipe.h"
+#include "stats/stats-cluster-single.h"
 #include "messages.h"
 #include "apphook.h"
 #include "mainloop-threaded-worker.h"
@@ -348,7 +350,7 @@ afinter_source_init(LogPipe *s)
 
   g_mutex_lock(&internal_msg_lock);
   current_internal_source = self;
-  metrics.queue_capacity = self->options->queue_capacity;
+  stats_counter_set(metrics.queue_capacity, self->options->queue_capacity);
   g_mutex_unlock(&internal_msg_lock);
 
   return TRUE;
@@ -589,7 +591,12 @@ afinter_message_posted(LogMessage *msg)
       stats_register_counter(0, &sc_key, SC_TYPE_QUEUED, &metrics.queued);
       stats_register_counter(0, &sc_key, SC_TYPE_DROPPED, &metrics.dropped);
       stats_register_counter(0, &sc_key, SC_TYPE_PROCESSED, &metrics.processed);
+
+      stats_cluster_single_key_set(&sc_key, "internal_events_queue_capacity", NULL, 0);
+      stats_register_counter(0, &sc_key, SC_TYPE_SINGLE_VALUE, &metrics.queue_capacity);
       stats_unlock();
+
+      stats_counter_set(metrics.queue_capacity, current_internal_source->options->queue_capacity);
 
       _register_obsolete_stats_alias(metrics.queued);
     }
@@ -636,7 +643,11 @@ afinter_global_deinit(void)
       stats_unregister_counter(&sc_key, SC_TYPE_QUEUED, &metrics.queued);
       stats_unregister_counter(&sc_key, SC_TYPE_DROPPED, &metrics.dropped);
       stats_unregister_counter(&sc_key, SC_TYPE_PROCESSED, &metrics.processed);
+
+      stats_cluster_single_key_set(&sc_key, "internal_events_queue_capacity", NULL, 0);
+      stats_unregister_counter(&sc_key, SC_TYPE_SINGLE_VALUE, &metrics.queue_capacity);
       stats_unlock();
+
       g_queue_free_full(internal_msg_queue, (GDestroyNotify)log_msg_unref);
       internal_msg_queue = NULL;
     }
--- a/lib/afinter.c
+++ b/lib/afinter.c
@@ -425,8 +425,8 @@ afinter_sd_init(LogPipe *s)
 
   if (cfg_is_config_version_older(cfg, VERSION_VALUE_3_29))
     {
-      msg_warning_once("WARNING: The internal_queue_length stat counter has been renamed to internal_source.queued. "
-                       "The old name will be removed in future versions", cfg_format_config_version_tag(cfg));
+      msg_warning_once("WARNING: The internal_queue_length stat counter has been renamed to internal_source.queued",
+                       cfg_format_config_version_tag(cfg));
     }
 
   if (!log_src_driver_init_method(s))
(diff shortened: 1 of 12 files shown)


MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

lib/healthcheck/healthcheck-control.c
```diff
@@ -28,6 +28,27 @@
...
+  stats_lock();
+  StatsCounterItem *internal_queued = stats_get_counter(&sc_key, SC_TYPE_QUEUED);
+  stats_unlock();
+
+  if (!internal_queued)
+    return;
+
+  g_string_append_printf(reply, PROMETHEUS_METRIC_PREFIX
+                         "internal_queue_events %"G_GSIZE_FORMAT"\n", stats_counter_get(internal_queued));
+}
```
alltilla, line 51: It is a bit confusing to see the value of this counter in the output, but it not being calculated into the healthcheck's result.

Can we expose some functions from `afinter`, like `get_queue_length()` and `is_queue_filled()` or something like that, and include that in the health calculation?

---

Also, the latency related metrics are exactly the same in the `stats prometheus` output and `healthcheck` output, maybe we should do the same for this, too.

modules/afsocket/afsocket-source.c
```diff
@@ -1216,8 +1224,8 @@ afsocket_sd_register_stats(AFSocketSourceDriver *self)
   StatsClusterLabel labels[] =
   {
     stats_cluster_label("id", self->super.super.id),
-    stats_cluster_label("direction", "input"),
     stats_cluster_label("driver_instance", afsocket_sd_format_name(&self->super.super.super)),
+    stats_cluster_label("direction", "input"),
```
alltilla, line 1236: [question]

Could you kindly explain why you have changed the order of the labels?

To the stats API the same labels in different orders are different labels. This means that the call site has to make sure that it sets up the labels in the right order. Although you have changed the order on both the registration and unregistration site, it is not clear to me why we would ever deviate from setting them up in alphabetical order for hard coded labels. If we do not have a globally accepted convention for the order of hard-coded labels, we could easily make mistakes, IMO. Prometheus will sort the labels in the end either way, if I remember correctly.

lib/healthcheck/healthcheck-control.c
MrAnno: The healthcheck unit is for doing real calculations. Since there is nothing to "check", it is here in the report.

The stats prometheus output contains processed/queued/dropped counters. Only the queue length is interesthing in a healthcheck report in my opinion, so I changed the name.

modules/afsocket/afsocket-source.c
MrAnno: It looked better (and we haven't released this metric yet).
We already have non-alphabetically-sorted counters, as I always preferred listing `id` first.

I think mistakes could be avoided if we always used the key builder for shared counters, which does sort labels.
For local-only uses, it is relatively easy to keep the same order by extracting a function that creates the key for registration and deregistration.

lib/healthcheck/healthcheck-control.c
alltilla: I though that if the queue is stuck (queue_length == queue_limit), it means syslog-ng is not healthy.

MrAnno: The internal queue's capacity is currently not exposed in stats Prometheus or here.
We can add a metric to both locations, but that would be a separate key, not a label.

I think syslog-ng is unhealthy if queue_length>E for more than a few seconds, where E is near 0 depending on the cost of the internal source's pipeline (destination queues should hold those messages, not the source).

I was thinking about showing the dropped counter instead, but I dismissed the idea as healthcheck should be about the current state of syslog-ng, not past states.

What do you think?

alltilla: I agree that if queue_length is more than 0 for more then a few seconds, it means syslog-ng is unhealthy. And I think it is hard to measure right now, as we only do one measurement.

But I would also say that if queue_length reaches queue_limit (which is much larger than the E you have mentioned) for even a split second, it also means that syslog-ng is unhealthy. This is a lot easier to decide, as we only need one measurement.

(IMO this is similar to cooking meat, where you can be sure that it is not raw if it has X temperature for multiple minutes/hours or if it has a higher Y temperature for a second. :D)

If I were to implement this, I would expose the two functions from afinter like I mentioned (fortunately it is in `lib/`), and I would not operate on its stats counters, but use the two functions to create and print the metric of its queue status and to decide its health.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

lib/healthcheck/healthcheck-control.c
```diff
@@ -28,6 +28,25 @@
...
+
+static inline void
+_append_internal_src_queue_metrics(AFInterMetrics metrics, GString *reply)
+{
+  gchar double_buf[G_ASCII_DTOSTR_BUF_SIZE];
+
+  gdouble queued = (gint) stats_counter_get(metrics.queued);
+  gsize queue_capacity = stats_counter_get(metrics.queue_capacity);
+
+  gdouble internal_queue_usage = 0;
```
alltilla, line 41: I think this should be 1, so if the queue_capacity is 0, we report full.

MrAnno: Right, I'm fixing this.

lib/afinter.c
```diff
@@ -423,12 +423,6 @@ afinter_sd_init(LogPipe *s)
   AFInterSourceDriver *self = (AFInterSourceDriver *) s;
   GlobalConfig *cfg = log_pipe_get_config(s);
 
-  if (cfg_is_config_version_older(cfg, VERSION_VALUE_3_29))
-    {
-      msg_warning_once("WARNING: The internal_queue_length stat counter has been renamed to internal_source.queued. "
-                       "The old name will be removed in future versions", cfg_format_config_version_tag(cfg));
-    }
-
```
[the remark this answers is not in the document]
MrAnno: I'll keep the warning.

kira-syslogng:
Build FAILURE

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:
Build started
Build FAILURE

PR #4412, opened 2023-04-03
tls: set client CA list based on the ca-dir() and ca-file() options

alltilla:
Filling the `certificate_authorities` field of the certificate request, to help clients choose certs to use.

From https://www.ietf.org/rfc/rfc5246.txt 7.4.4:
>   certificate_authorities
>      A list of the distinguished names [X501] of acceptable
>      certificate_authorities, represented in DER-encoded format.  These
>      distinguished names may specify a desired distinguished name for a
>      root CA or for a subordinate CA; thus, this message can be used to
>      describe known roots as well as a desired authorization space.  If
>      the certificate_authorities list is empty, then the client MAY
>      send any certificate of the appropriate ClientCertificateType,
>      unless there is some external arrangement to the contrary.

Example:
```
source s_network {
  network(
    port(12345)
    transport(tls)
    tls(
      ca-dir("/etc/ssl/certs/")
      ca-file("/home/alltilla/repos/syslog-ng/build/install/ssl/ca.crt")
      key-file("/home/alltilla/repos/syslog-ng/build/install/ssl/server.key")
      cert-file("/home/alltilla/repos/syslog-ng/build/install/ssl/server.crt")
      peer-verify(required-trusted)
      ssl-options(no_tlsv13)
    )
  );
};
```

Screenshot from 2023-04-03 18-01-43


alltilla:
I know that this is yet another case, where sometimes the function is a NOOP, but it would be strange to see the preprocessor macros about the SSL functions being further away from the place they are actually called.

lib/transport/tls-context.c
```diff
@@ -498,6 +532,9 @@ tls_context_setup_context(TLSContext *self)
   if (_is_file_accessible(self, self->crl_dir) && !SSL_CTX_load_verify_locations(self->ssl_ctx, NULL, self->crl_dir))
     goto error;
 
+  if (!_set_client_ca_list(self))
```
MrAnno, line 535: This is server only (`TM_SERVER `).

lib/transport/tls-context.c
```diff
@@ -460,6 +460,40 @@ tls_context_load_key_and_cert(TLSContext *self)
...
+    {
+      sk_X509_NAME_free(ca_list);
+      return FALSE;
+    }
+
+  if (_is_file_accessible(self, self->ca_file) && !SSL_add_file_cert_subjects_to_stack(ca_list, self->ca_file))
+    {
+      sk_X509_NAME_free(ca_list);
+      return FALSE;
+    }
```
MrAnno, line 486: Can we add the CAs that we parser from a PKCS #12 archive when `pkcs12-file()` is used?

lib/transport/tls-context.c
alltilla: According to https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_client_CA_list.html

>The "client CA list" functions below only have an effect when called on the server side.

This is why I have not checked which mode are we in, but I can add that explicitly as well. Would you like me to?

lib/transport/tls-context.c
alltilla: They are already added: https://github.com/syslog-ng/syslog-ng/pull/4412/files#diff-8f4f2584202391a2f25b8211042a90744d54bbcedc03da6a60eeadbe1f2cd4d3R383

I believe they are kept as I am getting, copying, appending, then setting the CA list. However, I did not have the time to check it myself with pkcs12. I will do that tomorrow.

MrAnno: Oh, I did not remember that. Thanks.

lib/transport/tls-context.c
MrAnno: I don't have a preference, I just remembered that we already have a few `TM_SERVER` checks in the code.

lib/transport/tls-context.c
alltilla: Tested the `pkcs12-file()` with `ca-dir()`, and they work together well.
image

alltilla: Updated the news entry :)

lib/transport/tls-context.c
alltilla: Added the mode check just for good measure.

alltilla:
I have observed, that in case of failure the `X509_NAME`s were leaking. Added a fix for it.

lib/transport/tls-context.c
MrAnno: Cool. Thank you.

MrAnno (approved):

PR #4415, opened 2023-04-06
format-json: fix RFC8259 number violation

MrAnno:
RFC8259 does not allow '+' sign and leading zeros in numbers, but these and leading spaces are allowed by type_cast_to_number().

Because syslog-ng should always store the original representation (normalizing the stored value would break other functionality), normalization/sanitization/etc. should be done in formatter functions.

### DIFF
--- a/modules/json/format-json.c
+++ b/modules/json/format-json.c
@@ -32,6 +32,7 @@
 #include "utf8utils.h"
 #include "scanner/list-scanner/list-scanner.h"
 #include "scratch-buffers.h"
+#include "str-format.h"
 
 typedef struct _TFJsonState
 {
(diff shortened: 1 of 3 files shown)


modules/json/format-json.c
```diff
@@ -234,6 +234,24 @@ tf_json_append_list(const gchar *name, const gchar *value, gsize value_len,
...
+  gchar double_buf[G_ASCII_DTOSTR_BUF_SIZE];
+  g_ascii_dtostr(double_buf, G_N_ELEMENTS(double_buf), d);
+  tf_json_append_value(name, double_buf, -1, state, FALSE);
+}
+
+static inline void
+tf_json_append_int(const gchar *name, gint64 i, json_state_t *state)
+{
+  gchar int_buf[32];
+  g_snprintf(int_buf, G_N_ELEMENTS(int_buf), "%"G_GINT64_FORMAT, i);
```
bazsi, line 250: could we use something other than sprintf here? I am not sure how much integer fields we can expect, but sprintf is really slow.

MrAnno: This is an interesting post about itoa() implementations:
https://jk-jeon.github.io/posts/2022/02/jeaiii-algorithm/

Since itoa is not portable, we could use our own, which does not have to be that complex.
The naive or the 2-digit-per-div method would be probably enough for us.

bazsi:
There's an implementation in str-format.c or is it str-utils.c?

MrAnno:
Thanks. I've updated the PR with the `format_int64_padded()` call.

MrAnno:
@kira-syslogng do stresstest

modules/json/format-json.c
```diff
@@ -414,17 +438,22 @@ tf_flat_json_call(LogTemplateFunction *self, gpointer s,
...
+
   *type = LM_VT_JSON;
   for (gint i = 0; i < args->num_messages; i++)
     {
       gboolean r = tf_flat_json_append(state, result, args->messages[i], args->options);
       if (!r && (args->options->opts->on_error & ON_ERROR_DROP_MESSAGE))
         {
           g_string_set_size(result, orig_size);
-          return;
+          goto exit;
```
bazsi, line 451: this changes behaviour, although supporting multiple messages in a $(format-json) is probably a bad idea.

This loop is iterating over the message context, formatting multiple JSON documents if there are multiple messages. This was probably a bad idea which we might be able to get rid off. But I would not change the behaviour with the 'goto exit;' here in neither locations.

bazsi (changes requested):
there's a change in behaviour with context-based template expansion, which I think we should address in a separate PR with potentially a larger scope (e.g. getting rid off that loop entirely).

Otherwise this looks good to me.

modules/json/format-json.c
MrAnno: Sorry, I don't understand your comment.
What did I change with the `goto exit`?
I used it only for reclaiming scratch buffers, which I think is correct even for multiple messages. Everything else should be the same.

bazsi:
Sorry. Missed the removed return.

Probably the lack of sleep.

bazsi (approved):

PR #4418, opened 2023-04-11
mainloop: deinit() configuration when init() fails on startup

MrAnno:
This does the same that syslog-ng does during reload: a partial (failed) initialization should be followed by a deinit() call.

It makes sure that the pipeline is deinitialized before exiting; for example, that disk-buffers are persisted and closed.


MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

alltilla (approved):

PR #4420, opened 2023-04-11
Config-related metrics

MrAnno:
### New configuration-related metrics

```
syslogng_last_config_reload_timestamp_seconds 1681309903
syslogng_last_successful_config_reload_timestamp_seconds 1681309758
syslogng_last_config_file_modification_timestamp_seconds 1681309877
```

The main config file is monitored by `CfgMonitor` with inotify.
If inotify fails or is not available, the config is polled every 15 minutes.

----

### syslog-ng configuration identifier

A new syslog-ng configuration keyword has been added, which allows specifying a config identifier. For example:
```
@config-id: cfg-20230404-13-g02b0850fc
```

This keyword can be used for config identification in managed environments, where syslog-ng instances and their
configuration are deployed/generated automatically.

`syslog-ng-ctl config --id` can be used to query the active configuration ID and the SHA256 hash of the full
"preprocessed" syslog-ng configuration. For example:

```
$ syslog-ng-ctl config --id
cfg-20230404-13-g02b0850fc (08ddecfa52a3443b29d5d5aa3e5114e48dd465e195598062da9f5fc5a45d8a83)
```

### DIFF
--- a/news/feature-4420-2.md
+++ b/news/feature-4420-2.md
@@ -1,0 +1,17 @@
+syslog-ng configuration identifier
+
+A new syslog-ng configuration keyword has been added, which allows specifying a config identifier. For example:
+```
+@config-id: cfg-20230404-13-g02b0850fc
+```
+
+This keyword can be used for config identification in managed environments, where syslog-ng instances and their
+configuration are deployed/generated automatically.
+
+`syslog-ng-ctl config --id` can be used to query the active configuration ID and the SHA256 hash of the full
+"preprocessed" syslog-ng configuration. For example:
+
+```
+$ syslog-ng-ctl config --id
+cfg-20230404-13-g02b0850fc (08ddecfa52a3443b29d5d5aa3e5114e48dd465e195598062da9f5fc5a45d8a83)
+```
--- a/news/feature-4420.md
+++ b/news/feature-4420.md
@@ -1,0 +1,7 @@
+New configuration-related metrics:
+
+```
+syslogng_last_config_reload_timestamp_seconds 1681309903
+syslogng_last_successful_config_reload_timestamp_seconds 1681309758
+syslogng_last_config_file_modification_timestamp_seconds 1681309877
+```


lib/pragma-grammar.ym
```diff
@@ -161,6 +163,12 @@ module_param
 	: LL_IDENTIFIER '(' string_or_number ')'		{ cfg_args_set(last_module_args, $1, $3); free($1); free($3); }
 	;
 
+config_id_stmt
+  : KW_CONFIG_ID ':' positive_integer64
+    {
+      configuration->user_config_id = $3;
```
bazsi, line 169: Don't we want to allow something like a git commit id? Or other version number like value?

lib/cfg.h
```diff
@@ -61,6 +64,11 @@ struct _GlobalConfig
   /* hex-encoded syslog-ng major/minor, e.g. 0x0201 is syslog-ng 2.1 format */
   gint user_version;
 
+  /* config identifier specified by the user */
+  gint64 user_config_id;
+
+  guint8 config_hash[CONFIG_HASH_LENGTH];
```
bazsi, line 70: I would probably make this an allocated value so we don't need to include openssl into cfg.h

lib/pragma-grammar.ym
bazsi: I can see that Prometheus does not support non-numeric values, so I understand this is numeric too, if we want to expose this value as a metric.

But do we need to export this as a metric?  I understand that we want to see if there was an issue with config reload, but we already have metrics for successful reloads and reloads in general, and if the two differs it means that the config wasn't successfully reloaded. 

We could potentially expose the timestamp of the config file, that should indicate that a reload didn't even start.

If the management service in question needs the config id, it can get it in other ways (e.g. running syslog-ng-ctl config --id)

One more idea is to include the config id as a label to the successful reload metric, although that would mean we would profilerate metrics, as we would have a separate one for each reload operation.

bazsi:
This is a cool feature. The only question I have was about using git commit ids in place of `@config-id`, if we resolve that, this could go in, I think.

lib/pragma-grammar.ym
MrAnno: We could allow strings, but in that case we can't publish this as a metric.

MrAnno: Good idea. I'll remove the config-id from metrics (allowing strings), and publish the main configuration's modification timestamp as a metric.

(Adding the config-id as a label does not sound like the "Prometheus way to go", label cardinality highly affects the performance of Prometheus, and the metric itself would be even more difficult to query/understand.)

lib/cfg.h
MrAnno: Fixed, thanks.

lib/pragma-grammar.ym
bazsi: Agreed. Although I think that having a few reloads per day shouldn't be that much of a problem and Prometheus would eventually garbage collect old counters that don't receive updates. But I am also on an opinion that if the management system is in a position to trigger a reload, it can also query the current config version.

During reconsiliation, the comparison between desired state vs. current state would likely use this logic.

MrAnno:
@kira-syslogng do stresstest

MrAnno:
I've updated the PR description according to my changes.

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

alltilla (approved):

PR #4423, opened 2023-04-12
Native mac os system source 4212

HofiOne:
`system()`, `darwin-oslog()`, `darwin-oslog-stream()` sources: Added 2 darwin OSLog based sources `darwin-oslog()` and `darwin-oslog-stream()`.

'darwin-oslog()' replaced the earlier file source based solution with a native OSLog framework based one, and is automatically used in the `system()` source on darwin platform if the **darwinosl** plugin is presented.

This plugin available only on macOS 10.15 Catalina and above, the first version that has the OSLog API.

### darwin-oslog

This is a native OSLog Framework based source to read logs from the local store of the unified logging system on darwin OSes.
For more info, see https://developer.apple.com/documentation/oslog?language=objc

The following parameters can be used for customization:

- filter-predicate
    - string value, which can be used to filter the log messages natively
    - default value: `(eventType == 'logEvent' || eventType == 'lossEvent' || eventType == 'stateEvent' || eventType == 'userActionEvent') && (logType != 'debug')`
    - for more details, see
        - `man log`
        - https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Predicates/Articles/pSyntax.html
- go-reverse
    - boolean value, setting to **yes** will provide a reverse-ordered log list (from latest to oldest)
    - default value: **no**
- do-not-use-bookmark
    - boolean value, setting to **yes** will prevent syslog-ng from continuing to feed the logs from the last remembered position after a (re-)start, which means, depending on the other settings, the feed will always start from the end/beginning of the available log list
    - default value: **no**, which means syslog-ng will attempt to continue feeding from the last remembered log position after a (re-)start
- max-bookmark-distance
    - integer value, maximum distance in seconds that far an earlier bookmark can point backward, e.g. if syslog-ng was stopped for 10 minutes and max-bookmark-distance is set to 60 then syslog-ng will start feeding the logs only from the last 60 seconds at startup, 9 minutes of logs 'will be lost'
    - default value: 0, which means no limit
- read-old-records
    - boolean value, controls if syslog-ng should start reading logs from the oldest available at first start (or if no bookmark can be found)
    - default value: yes
- fetch-delay
    - integer value, controls how much time syslog-ng should wait between reading/sending log messages, this is a fraction of a second, where wait_time = 1 second / n,  so, e.g. n=1 means that only about 1 log will be read and sent in each second, and n=1 000 000 means only 1 microsecond (the allowed minimum value now!) will be the delay between read/write attempts
    (Use with care, though lower delay time can increase log feed performance, at the same time could lead to a heavy system load!)
    - default value: 10 000
- fetch-retry-delay
    - integer value, controls how many seconds syslog-ng will wait before a repeated attempt to read/send once it's out of available logs
    - default value: 1
- log-fetch-limit
    - **Warning**: _This option is now disabled due to an OSLog API bug (https://openradar.appspot.com/radar?id=5597032077066240), once it's fixed it will be enabled again_
    - integer value, that limits the number of logs syslog-ng will send in one run
    - default value: 0, which means no limit

NOTE: the persistent OSLog store is not infinite, depending on your system setting usually, it keeps about 7 days of logs on disk, so it could happen that the above options cannot operate the way you expect, e.g. if syslog-ng was stopped for about more then a week it could happen that will not be able to restart from the last saved bookmark position (as that might not be presented in the persistent log anymore)

### darwin-oslog-stream

This is a wrapper around the OS command line "log stream" command that can provide a live log stream feed.
Unlike in the case of `darwin-oslog()` the live stream can contain non-persistent log events too, so take care, there might be a huge number of log events every second that could put an unusual load on the device running syslog-ng with this source.
Unfortunately, there's no public API to get the same programmatically, so this one is implemented using a program() source.

Possible parameters:

- params
    - a string that can contain all the possible params the macOS `log` tool can accept, please see, `log --help stream` for full reference, and `man log` for more details.

      IMPORTANT: the parameter --style is used internally (defaults to `ndjson`), so it cannot be overridden, please use other sysylog-ng features (templates, rewrite rules, etc.) for final output formatting
    - default value: `--type log --type trace --level info --level debug`, you can use \``def-osl-stream-params`\` for referencing it if you wish to keep the defaults when you add your own


czanik:
Tested both drivers on my MacBook, including some of the parameters, and they worked as expected. My favorite parameter was: read-old-records(no), as this way syslog-ng did not try to read millions of logs from mid March :-)

kira-syslogng:
Build FAILURE

HofiOne:
@kira-syslogng do stresstest

kira-syslogng:

HofiOne:

HofiOne:

modules/darwinosl/darwinosl-grammar.ym
```diff
@@ -0,0 +1,105 @@
...
+/* INCLUDE_DECLS */
+
+%token KW_DARWIN_OSL
+%token KW_FILTER_PREDICATE
+%token KW_GO_REVERSE
+%token KW_DO_NOT_USE_BOOKMARK
+%token KW_MAX_BOOKMARK_DISTANCE
+%token KW_LOG_FETCH_DELAY
+%token KW_LOG_FETCH_RETRY_DELAY
+
```
bazsi, line 59: I would lose the "log-" prefix from these keywords, I know that log-iw-size(), log-fifo-size() at all use it, but I think low-level, driver specific options should not.

modules/darwinosl/darwinosl-source.m
```diff
@@ -0,0 +1,559 @@
...
+  LogSource *worker = (LogSource *) self->super.worker;
+  Bookmark *bookmark = ack_tracker_request_bookmark(worker->ack_tracker);
+  darwinosl_source_persist_fill_bookmark(self->log_source_persist, bookmark, self->log_source_position);
+  msg_trace("darwinosl: Bookmark created",
+            evt_tag_printf("position", "%.6f", self->log_source_position.log_position),
+            evt_tag_printf("last message hash", "%u", self->log_source_position.last_msg_hash),
+            evt_tag_printf("filter predicate hash", "%u", self->log_source_position.last_used_filter_predicate_hash));
+
+  LogMessage *msg;
+  gsize msg_len = _log_message_from_string(log_string, self->options.format_options, &msg);
```
bazsi, line 281: I was wondering if we need the round-trip to the LogMessage becoming an RFC5424 formatted string and then reparsed into a LogMessage.

modules/darwinosl/darwinosl-source.m
```diff
@@ -0,0 +1,559 @@
...
+                evt_tag_str("bookmark", [OSLogSource.RFC3339DateFormatter stringFromDateWithMicroseconds:
+                                                                          nextLogEntry.date].UTF8String));
+    }
+  else
+    {
+      if (self->log_source_position.last_msg_hash)
+        msg_info("darwinosl: Could not restore last bookmark (filter might be changed or max_bookmark_distance took effect?)",
+                 evt_tag_str("bookmark", [OSLogSource.RFC3339DateFormatter stringFromDateWithMicroseconds:
+                                                                           [NSDate dateWithTimeIntervalSince1970:self->log_source_position.log_position]].UTF8String),
+                 evt_tag_str("new start position", [OSLogSource.RFC3339DateFormatter stringFromDateWithMicroseconds:
```
bazsi, line 149: please don't use spaces in the tag name. it needs to be parseable relatively easily and spaces break that format.

modules/darwinosl/darwinosl-source.m
```diff
@@ -0,0 +1,559 @@
...
+                         filterString:(self->options.filter_predicate ? [NSString stringWithUTF8String:self->options.filter_predicate] : nil)
+                         options:(self->options.go_reverse ? OSLogEnumeratorReverse : 0)] == nil)
+    {
+      [self->osLogSource closeAll];
+      return FALSE;
+    }
+
+  if (positionRestored)
+    {
+      OSLogEntry *nextLogEntry = [self->osLogSource fetchNextEntry];
```
bazsi, line 223: We are fetching a log entry after restoring our position, but is that  log entry delivered at the end? My assumption from reading the code is that fetchNextEntry() would move the current position, so by the time _fetch() is called, we would skip this one, unless we make sure that it gets delivered somehow.

But I don't see that here.

modules/darwinosl/darwinosl-source.m
```diff
@@ -0,0 +1,559 @@
...
+}
+
+static void
+_request_exit(LogThreadedSourceDriver *s)
+{
+  DarwinOSLogSourceDriver *self = (DarwinOSLogSourceDriver *) s;
+  g_atomic_counter_set(&self->exit_requested, TRUE);
+}
+
+/* TODO: Using some iv_timer or similar instead ?! */
```
bazsi, line 300: this is a threaded source, so we don't need to use ivykis, a blocking call is fine.

modules/darwinosl/darwinosl-source.m
```diff
@@ -0,0 +1,559 @@
...
+          {
+            @autoreleasepool
+            {
+              LogThreadedFetchResult res = _fetch(s);
+              if (res.result == THREADED_FETCH_SUCCESS)
+                _send(s, res.msg);
+              else
+                break;
+            }
+            _sleep(self, 1.0 / self->options.fetch_delay);
```
bazsi, line 347: could we avoid floating point calculations here? We tried to avoid those in fast paths.
Maybe by precalculate it or used fixed mode arithmetic

scl/darwinosl/plugin.conf
```diff
@@ -0,0 +1,79 @@
...
+
+		parser {
+	    json-parser ();
+			date-parser(
+				format("%Y-%m-%d %T.%f%z")
+				template("$timestamp")
+			);
+	    add-contextual-data(
+				selector("$messageType"),
+				database("../share/syslog-ng/include/scl/darwinosl/darwinosl-metadata-db.csv"),
```
bazsi, line 60: relative path will not work when installed. There's a define called scl-root that you can use between backticks
to reference the SCL directory.

scl/darwinosl/plugin.conf
```diff
@@ -0,0 +1,79 @@
...
+				default-selector("unknown"),
+				prefix(".priorities.")
+			);
+		};
+
+		rewrite {
+			subst(
+					" ",
+					"\\0x20",
```
bazsi, line 69: I think we should use underscore here, we should only change $PROGRAM and not the processImagePath field.

scl/darwinosl/plugin.conf
```diff
@@ -0,0 +1,79 @@
...
+					no-parse
+					no-multi-line
+					dont-store-legacy-msghdr
+					assume-utf8
+				)
+			);
+		};
+
+		parser {
+	    json-parser ();
```
bazsi, line 53: let's put these name value pairs under a prefix, e.g. ".darwinoslog." or something like that. We usually make these prefixes configurable, but at the same time use a sensible default that lets the user associate where a key is coming from.

scl/darwinosl/plugin.conf
```diff
@@ -0,0 +1,79 @@
...
+													"
+
+block source darwin-oslog(...) {
+@requires darwinosl "The darwinosl() source driver depends on the darwinosl module"
+	darwinosl(
+		filter-predicate("`def-osl-predicate`")
+		`__VARARGS__`
+		flags(syslog-protocol dont-store-legacy-msghdr assume-utf8)
+	);
+};
```
bazsi, line 36: I think you could use set defaults from the C code, I would not complicate it by using a block as well.

scl/darwinosl/plugin.conf
```diff
@@ -0,0 +1,79 @@
...
+	    json-parser ();
+			date-parser(
+				format("%Y-%m-%d %T.%f%z")
+				template("$timestamp")
+			);
+	    add-contextual-data(
+				selector("$messageType"),
+				database("../share/syslog-ng/include/scl/darwinosl/darwinosl-metadata-db.csv"),
+				default-selector("unknown"),
+				prefix(".priorities.")
```
bazsi, line 62: this should probably be mapped into the same prefix as the rest of the JSON fields.

modules/darwinosl/darwinosl-source.m
```diff
@@ -0,0 +1,559 @@
...
+  self->osLogSource = nil;
+
+  darwinosl_sd_options_destroy(&self->options);
+  darwinosl_source_persist_free(self->log_source_persist);
+
+  if (self->persist_name)
+    g_free((gpointer)self->persist_name);
+  if (self->stat_persist_name)
+    g_free((gpointer)self->stat_persist_name);
+
```
bazsi, line 437: g_free handles NULLs. why do we need the cast?

modules/darwinosl/darwinosl-source.m
```diff
@@ -60,12 +62,73 @@ void darwinosl_sd_options_defaults(DarwinOSLogSourceOptions *self,
...
+static void
+_unregister_aggregated_stats(DarwinOSLogSourceDriver *self)
+{
+  stats_aggregator_lock();
+
+  stats_unregister_aggregator_maximum(&self->max_message_size);
+  stats_unregister_aggregator_average(&self->average_messages_size);
+  stats_unregister_aggregator_cps(&self->CPS);
+
```
bazsi, line 128: I hate that we need to add this here. Can't we move these stats registrations to the LogThreadedSourceDriver level?

scl/darwinosl/plugin.conf
HofiOne: In the first version I also wanted to use underscore, but escaping it could signal for any later possible parsing attempt what the original value was.
Does it make sense?

scl/darwinosl/plugin.conf
HofiOne: strange as I was very curious if it worked or not, tested multiple times and worked fine for me even on the final installation paths

scl/darwinosl/plugin.conf
HofiOne: I wanted to expose these here to
- give a properly visible overview, hint about the default to the users
- prevent overwriting of flags(syslog-protocol dont-store-legacy-msghdr assume-utf8) defaults

modules/darwinosl/darwinosl-grammar.ym
HofiOne: log-iw-size, log-fifo-size is the smaller issue, my problem here is that we already have `log-fetch-limit` that is the reason i used the `log-fetch`  (not simply the `log-` prefix)

modules/darwinosl/darwinosl-source.m
HofiOne: The name `fetchNextEntry` is confusing a bit here unfortunately (I just kept it to follow the original Apple API name convention) :S

At the first time invocation the underlying NSEnumerator will fetch the first element from the collection that in this case is the one that the call of _open_osl (more precisely inside that the `entriesEnumeratorWithOptions:position:predicate:error` function) positioned the enumerator to (or `nil` if no such item anymore) that is the one we saved as the bookmark.
So basically it positions the enumerator "before" the bookmarked items pos.

This is why the _check_restored_postion in the following if() could work as well.

modules/darwinosl/darwinosl-source.m
HofiOne: I guess the other option would be to compose the LogMessage field-by-field manually?

I'm not sure if I could compose it more optimally, but of course, I can give it a try ;)

kira-syslogng:
Build FAILURE

czanik:
It would be fantastic if it could be merged and made available in the next release. Now it comes up in discussions even here in Hungary, not just at US events.

HofiOne:

modules/darwinosl/darwinosl-grammar.ym
```diff
@@ -0,0 +1,105 @@
...
+%type   <ptr> source_darwinosl_params
+
+%%
+
+start
+  : LL_CONTEXT_SOURCE source_darwinosl { YYACCEPT; }
+  ;
+
+source_darwinosl
+	: KW_DARWIN_OSL
```
kovgeri01, line 70: There are some tabs inside this file, at lines: 70,75,78,79,80

modules/darwinosl/darwinosl-grammar.ym
```diff
@@ -0,0 +1,105 @@
...
+      darwinosl_sd_set_filter_predicate(last_driver, $3);
+      free($3);
+    }
+  | KW_GO_REVERSE '(' yesno ')'	{ darwinosl_sd_set_go_reverse(last_driver, $3); }
+  | KW_DO_NOT_USE_BOOKMARK '(' yesno ')'	{ darwinosl_sd_set_do_not_use_bookmark(last_driver, $3); }
+  | KW_MAX_BOOKMARK_DISTANCE '(' nonnegative_integer ')'	{ darwinosl_sd_set_max_bookmark_distance(last_driver, $3); }
+  | KW_FETCH_DELAY '(' positive_integer ')'	{ darwinosl_sd_set_log_fetch_delay(last_driver, $3); }
+  | KW_FETCH_RETRY_DELAY '(' nonnegative_integer ')'	{ darwinosl_sd_set_log_fetch_retry_delay(last_driver, $3); }
+  /*
+    TODO: This cannot be used due to an actual macOS API bug now, as this would lead to unclosed enumerator threads with
```
kovgeri01, line 95: Can you please put some kind of reference to this issue s it can be checked if it is resolved, when someone stumbles upon this TODO?

modules/darwinosl/darwinosl-source-oslog.m
```diff
@@ -0,0 +1,201 @@
...
+
+  if (@available(macOS 12.0, *))
+    self.logStore = [OSLogStore storeWithScope:OSLogStoreSystem error:&error];
+  else
+    self.logStore = [OSLogStore localStoreAndReturnError:&error];
+
+  if (self.logStore == nil)
+    msg_error("darwinosl: Error getting handle to OSLog store",
+              evt_tag_long("error code", error.code),
+              evt_tag_str("error", error.localizedDescription.UTF8String));
```
kovgeri01, line 85: as Bazsi said it, do not put spaces in tag names
```suggestion
              evt_tag_long("error_code", error.code),
```

modules/darwinosl/darwinosl-source-oslog.m
```diff
@@ -0,0 +1,201 @@
...
+      self.filterPredicate = filterPredicate;
+      self.logPosition = [self.logStore positionWithDate:startDate];
+
+      self.enumerator = [self.logStore entriesEnumeratorWithOptions:self.enumeratorOptions
+                                       position:self.logPosition
+                                       predicate:self.filterPredicate
+                                       error:&error];
+      if (self.enumerator == nil)
+        msg_error("darwinosl: Could not get OSLog enumerator",
+                  evt_tag_long("error code", error.code),
```
kovgeri01, line 110: as Bazsi said it, do not put spaces in tag names
```suggestion
                  evt_tag_long("error_code", error.code),
```

modules/darwinosl/darwinosl-source-oslog.m
```diff
@@ -0,0 +1,201 @@
...
+
+  if (self.logStore == nil)
+    msg_error("darwinosl: Error getting handle to OSLog store",
+              evt_tag_long("error code", error.code),
+              evt_tag_str("error", error.localizedDescription.UTF8String));
+
+  return self.logStore;
+}
+
+- (OSLogEnumerator *) openEnumeratorWithDate:(NSDate *)startDate
```
kovgeri01, line 90: Can you rename this to either `openEnumeratorWithDateAndFilter` or just simply `openEnumerator`, I personally prefer the latter as based on the codeboth is optional, the startDate can be nil (for example if `self->options.do_not_use_bookmark == TRUE` ), and filterString also can be nil as well.

modules/darwinosl/darwinosl-source-oslog.m
```diff
@@ -0,0 +1,201 @@
...
+  filterString:(NSString *)filterString
+  options:(OSLogEnumeratorOptions)options
+{
+  g_assert(self.enumerator == nil);
+  @try
+    {
+      NSPredicate *filterPredicate = (filterString.length > 0 ? [NSPredicate predicateWithFormat:filterString] : nil);
```
kovgeri01, line 97: I might be wrong on this, but looks like that there are some cases when filterString is nil when. Will ObjectiveC survive a `filterString.length` when `filterString==nil`?

kovgeri01: Same in the catch block as well at  line 117

kovgeri01: nvm I checked it it is fine

modules/darwinosl/darwinosl-grammar.ym
HofiOne: i can't unfortunately, it is a bug reported by me, and Apple bug reports are not public (even though i could add an open radar item, and probably link that here)

HofiOne: hehe, it is already added, as it can be seen in the PR description :D
https://openradar.appspot.com/radar?id=5597032077066240
adding here too

kira-syslogng:
Build started
Build FAILURE

kira-syslogng:
Build started
Build FAILURE

scl/darwinosl/plugin.conf
```diff
@@ -0,0 +1,81 @@
...
+block source darwin-oslog(...) {
+@requires darwinosl "The darwinosl() source driver depends on the darwinosl module"
+	darwinosl(
+		filter-predicate("`def-osl-predicate`")
+		`__VARARGS__`
+		flags(syslog-protocol dont-store-legacy-msghdr assume-utf8)
+	);
+};
+
+block source darwin-oslog-stream(params("--style ndjson --type log --type trace --level info --level debug") user-params("")) {
```
kovgeri01, line 38: I an trying to think that it might be better to have these as a boolean parameter with true and false like:

```suggestion
block source darwin-oslog-stream(
  collect-type-log(true)
  collect-type-trace(true)
  collect-type-activities(false)
  collect-level-info(true)
  collect-level-debug(true)
  collect-level-default(false)
  predicate("")
  process("")
){
............
         program(
             "/usr/bin/log stream --style ndjson $(if ("`collect-type-log`" == "true") "--type log " "") $(if ("`collect-type-trace`" == true) "--type trace " "") $(if ("`collect-type-activities`" == "true") "--type activities " "") $(if ("`collect-level-info`" == "true") "--level info " "") $(if ("`collect-level-debug`" == "true") "--level debug " "") $(if ("`collect-level-default`" == "true") "--level default " "") $(if ("`predicate`" ne "") "--predicate `predicate` " "") $(if ("`process`" ne "") "--predicate `process` " "")"
             flags(
             ......
         )
}
```

A solution like this will make sure that the user will not intentionally set something really bad for themselfs or unset the ` --style ndjson` flag by accident. (I might not get it eight at first but it should work based on the code at https://github.com/syslog-ng/syslog-ng/blob/master/scl/splunk/splunk.conf#L119C22-L119C91)

If the conf file can't do this than it might be worth to do itusing a generator like how https://github.com/syslog-ng/syslog-ng/blob/master/scl/loadbalancer/plugin.conf works

Also this will make it easier to do the documentation for the driver as right now the news.md does not contain the options that can be set (`params` and `user-params` are not referenced in the news file )

czanik:
OK. Tested again. Works perfectly when compiled with ```CC=clang```. Build fails if compiled using GCC (it worked earlier, but broke recently):

```
  CC       modules/metrics-probe/libmetrics_probe_la-label-template.lo
  CCLD     modules/metrics-probe/libmetrics-probe.la
ld: warning: -dylib_file is deprecated. Use -F or -L to control where indirect dylibs are found
  CC       modules/darwinosl/libdarwinosl_la-darwinosl-plugin.lo
  OBJC     modules/darwinosl/libdarwinosl_la-darwinosl-source.lo
libtool: compile: unable to infer tagged configuration
libtool:   error: specify a tag with '--tag'
make[2]: *** [modules/darwinosl/libdarwinosl_la-darwinosl-source.lo] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2

```

@HofiOne will fix it.

@bazsi @alltilla I'd like to have the MacOS source in 4.6.0 as it is a regularly asked feature, and I pointed people to the PR for well over half a year now...

HofiOne:
> OK. Tested again. Works perfectly when compiled with `CC=clang`. Build fails if compiled using GCC (it worked earlier, but broke recently):
> 
> ```
>   CC       modules/metrics-probe/libmetrics_probe_la-label-template.lo
>   CCLD     modules/metrics-probe/libmetrics-probe.la
> ld: warning: -dylib_file is deprecated. Use -F or -L to control where indirect dylibs are found
>   CC       modules/darwinosl/libdarwinosl_la-darwinosl-plugin.lo
>   OBJC     modules/darwinosl/libdarwinosl_la-darwinosl-source.lo
> libtool: compile: unable to infer tagged configuration
> libtool:   error: specify a tag with '--tag'
> make[2]: *** [modules/darwinosl/libdarwinosl_la-darwinosl-source.lo] Error 1
> make[1]: *** [all-recursive] Error 1
> make: *** [all] Error 2
> ```
> 
> @HofiOne will fix it.
> 
> @bazsi @alltilla I'd like to have the MacOS source in 4.6.0 as it is a regularly asked feature, and I pointed people to the PR for well over half a year now...

Fix is pushed, and seems to be fine now, all green

@kovgeri01  re-started the review and we are working on this actively, hopefully, we can get it integrated within a few days

scl/darwinosl/plugin.conf
HofiOne: these 2 here as a kind of an example, and a predefined, default filtered log stream
it is meaningless to modelling all the possible params the user can use here as a filter predicate can be a very-very complex expression
instead, we provide them the filter-predicate that via thay can filter the stream and the static log database as thay want, freely
the forcing of the -type log --type trace --level info --level debug params are intentional, we'd like to have a useful set of retrieved log entries
if it does not for someone thay can even override them, add filter predicate, or completly re-define the stream if needed

HofiOne: All the concerns you described are valid, but this is not intended to be a foolproof all in one, supertool
What you described is again would be a lot of effort to implement correctly, so as i wrote above, i would let it to the user now, in a next round we might make an enhancement on this.

modules/darwinosl/darwinosl-source.m
HofiOne: just for the record, i just searched for a single type of tag (evt_tag_long),  and found a bunch of tag names with spaces, anyway, renamed them

modules/darwinosl/darwinosl-source.m
HofiOne: As i can recall i used a LogMessage because 
- it helps keep everything in one container, and allows an easier passing for further processing
- it can parse nicely the string out of the box and gives access to each msg elements when needed
- this fetch is conforming to  _LogThreadedFetcherDriver::LogThreadedFetchResult (*fetch)(LogThreadedFetcherDriver *self), and there was a version when this was part of a real _super.fetch implementation

modules/darwinosl/darwinosl-source-oslog.m
HofiOne: This is part of the ObjectiveC naming convention, last tag in the function name is referring to the first function parameter, pls see the entriesEnumeratorWithOptions: function as a reference bellow

modules/darwinosl/darwinosl-source-oslog.m
HofiOne: yepp, in ObjectiveC the nil is handled safely, automatically (strange, and i did not like it at first, but this is how it works)

news/feature-4423.md
```diff
@@ -0,0 +1,60 @@
...
+
+NOTE: the persistent OSLog store is not infinite, depending on your system setting usually, it keeps about 7 days of logs on disk, so it could happen that the above options cannot operate the way you expect, e.g. if syslog-ng was stopped for about more then a week it could happen that will not be able to restart from the last saved bookmark position (as that might not be presented in the persistent log anymore)
+
+### darwin-oslog-stream
+
+This is a wrapper around the OS command line "log stream" command that can provide a live log stream feed.
+Unlike in the case of `darwin-oslog()` the live stream can contain non-persistent log events too, so take care, there might be a huge number of log events every second that could put an unusual load on the device running syslog-ng with this source.
+Unfortunately, there's no public API to get the same programmatically, so this one is implemented using a program() source.
+
+For possible parameters please see, `log --help stream` for full reference, and `man log` for more details.
```
kovgeri01, line 52: The parameter type currently looks confusing can you do it like:
```suggestion
Possible parameters:
- `params`: consists of the default parameters for the `log stream` command. The default value is: `--type log --type trace --level info --level debug`
- `user-params`: consists of additional parameters for the `log stream` command if the `params` parameters are needed.

For example the following config:
`darwin-oslog-stream(params("--level info") user-params("--type log"))`
Will result in a call of `log stream --level info --type log`

*Note*: that the `--style` parameter should not be used as it will break the parsing of the messages
```

I am not sure but please check it if the `--timeout` `--source` options and the `--color` can break the parsing, as these might break the format that syslog-ng is parsing, and add them to the note part of the

HofiOne: i messed up the again the param descriptions, i will fix them

--style cannot be overriden at all now as it is the last parameter so it will be used by the log tool, i guess that prt is fine as it is

modules/darwinosl/darwinosl-source.m
```diff
@@ -0,0 +1,575 @@
...
+  return &self->super.super.super;
+}
+
+/* Options */
+
+gboolean
+darwinosl_sd_set_filter_predicate(LogDriver *s, const gchar *filter_predicate_str)
+{
+  DarwinOSLogSourceDriver *self = (DarwinOSLogSourceDriver *)s;
```
kovgeri01, line 487: Check the predicate syntax by try to initialize it or something.

Currently if the syntax is invalid for example `filter-predicate("adassdadad")` than the exception will not be catched and the ObjC exception will kill the process, but if the predicate is only invalid in terms of semantics like `filter-predicate("adassdadad == true")` syslog-ng wil warn you about it like `[2024-01-10T16:36:00.574676] darwinosl: Could not get OSLog enumerator; error_code='-1', error='issue with predicate: no such field: adassdadad'` but it will continue executing.

By checking the predicate's format here will catch this issue, and maybe we can even show where the error is inside the configuration file for better user experience when having >1 darwing-oslog sources

news/feature-4423.md
HofiOne: removed the user_params (i did not like it anyway), i hope this one is fine now

modules/darwinosl/darwinosl-source.m
HofiOne: Seems there is no chance to catch the NSInvalidArgumentException, so this will still always crash if the predicte string has certain parsing issues. I concider this an Apple bug, trying to get a workaround of this later.

kovgeri01 (approved):
I created the issue #4786 that stands for adding the generic metrics (or some of the metrics) here to the LogThreadedSourceDriver as it would be potentially quite huge of a change impacting multiple modules.As for the code and the documentation it is fine by me. (Also it looks like that based on our discussions that the ObjC exception only escapes the catch clause when compiled in debug mode, and youve made it so that an actual error will be shown inside the configuration file if the predicate has invalid syntax)

With this I approve youre PR hopefully @bazsi @alltilla it is fine by you

PR #4426, opened 2023-04-13
TLS custom certificate validation plugin

Ricfilipe:
As discussed in #4219, currently there is no OCSP/CRL validation for remote peer certificates.
In this PR, a hook is added for a custom certificate validation for remote certificate. The hook is only called when the remote certificate is being verified. The hook will call the plugin with SSL_CONTEXT which allows the custom validation.

A plugin example, called tls-test-validation,  was added that verifies if a certificate has an expected identifier.

```
destination test {
    syslog("172.16.17.4" transport("tls") ip-protocol(4) port(6514) keep-alive(no)
        suppress(5)
        tls(
            ca-dir("/path/to/ca/")
            peer-verify(yes)
        )
        tls-test-validation(identity("test"))
        disk-buffer(
            mem-buf-length(2000)
            disk-buf-size(8388608)
            dir("/tmp")
        )
    );
```

Let me know if any changes are needed

### DIFF
--- a/modules/examples/inner-destinations/tls-test-validation/tls-test-validation.h
+++ b/modules/examples/inner-destinations/tls-test-validation/tls-test-validation.h
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2020 One Identity
+ * Copyright (c) 2020 Laszlo Budai <laszlo.budai@outlook.com>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 as published
+ * by the Free Software Foundation, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ * As an additional exemption you are allowed to compile & link against the
+ * OpenSSL libraries as published by the OpenSSL project. See the file
+ * COPYING for details.
+ *
+ */
+
+#ifndef TLS_TEST_VALIDATION_H_INCLUDED
+#define TLS_TEST_VALIDATION_H_INCLUDED
+
+#include "driver.h"
+
+typedef struct _TlsTestValidationPlugin TlsTestValidationPlugin;
+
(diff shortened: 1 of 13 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

Ricfilipe:
@MrAnno, @bazsi  FYI

MrAnno:
@kira-syslogng ok to test

modules/examples/inner-destinations/tls-test-validation/tls-test-validation.h
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2020 One Identity
+ * Copyright (c) 2020 Laszlo Budai <laszlo.budai@outlook.com>
```
MrAnno, line 3: You can use your own name in the newly introduces files.
Please also update the year.

MrAnno (commented):
This is excellent work. Thank you for working on this PR.

I only had a minor note.

bazsi:
Thanks for creating this.

I like the signal infrastructure, looks good. I was wondering if the "tls-test-validation" plugin could somehow be evolved into a more evolved set of features and moved out of examples. What has come up were stuff like calling an external script to perform validation or even do an OCSP call. Anyway, an example is good first step.

On a related note, I can see it performs a `tls_verify_certificate_name()`, but that is already validated by the afinet driver too, with a different value (host_name). Is there an actual use-case with doing this validation twice, or it's just meant to be used as an example?

Ricfilipe:
The "tls-test-validation" is just meant to be used as an example and demonstrate how to access the X509_STORE_CTX and validate a certificate.

I am planning to create a custom plugin that will communicate with an external certificate validator through a socket based on these [changes ](https://github.com/syslog-ng/syslog-ng/compare/master...amdrsantos:syslog-ng:tls-custom-certificate-validation). 

Since the use case is pretty specific, I was not planning to add the plugin to this PR and instead added the example.

MrAnno:
Thank you.

Your contribution is valid without the OI copyright.
To make it pass to the copyright  check, please add the following:

```diff
diff --git a/tests/copyright/policy b/tests/copyright/policy
index 3da6f2a03..567616b8d 100644
--- a/tests/copyright/policy
+++ b/tests/copyright/policy
@@ -166,6 +166,7 @@ modules/diskq/tests/test_logqueue_disk\.c$
 modules/metrics-probe/metrics-probe(|-parser|-plugin|-grammar)\.(c|h|ym)$
 modules/metrics-probe/label-template\.(c|h)$
 modules/metrics-probe/tests/test_metrics_probe\.c$
+modules/examples/inner-destinations/tls-test-validation
 scl/fortigate/.*\.conf$
 scl/cee/.*\.conf$
 scl/mariadb/.*\.conf$
```

MrAnno (approved):

PR #4428, opened 2023-04-14
cmake: Added missing handling of ENABLE_TESTING option in cmake builds

HofiOne:
CMake builds did not fully respect the ENABLE_TESTING option

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -305,30 +305,30 @@ set_target_properties(check-copyright PROPERTIES
   ADDITIONAL_MAKE_CLEAN_FILES
   "copyright-run.log;copyright-err.log")
 
-# The inclusion of CTest triggers enable_testing()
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build FAILURE

HofiOne:
> CMake has its own option called `BUILD_TESTING`: https://cmake.org/cmake/help/latest/module/CTest.html
> 
> I think we should use that one instead of specifying our own.

Actually, this is not a new one, I just used the one that the automake version is using.
Do we really want to use a different option than the automake introduced one?

kira-syslogng:
Build FAILURE

HofiOne:
> I think yes, we should use the option that the build system provides if there is one.

as you wish

MrAnno (approved):

HofiOne:
thanks, https://github.com/syslog-ng/syslog-ng/pull/4474

PR #4432, opened 2023-04-18
lib:  make persist_state_alloc_string() reuse the same entry if possible

### LINKED ISSUE #3337: make persist_state_alloc_string() reuse the same entry if possible
If a module needs to store a string inside the persist file, there are two nice helper functions:

```
gchar *persist_state_lookup_string(PersistState *self, const gchar *key, gsize *length, guint8 *version);
void persist_state_alloc_string(PersistState *self, const gchar *persist_name, const gchar *value, gssize len);
```

These work well in general with a small inconvenience: `persist_state_alloc_string()` always allocates a new persist entry and invalidates the old entry. Invalidation is not removal, just setting the `in_use` flag to false. This means the persist file grows after each call on the disk.

It would be a nice optimalizaton that if the string to store would fit into the original place, then `persist_state_alloc_string()` should reuse it.

The change would require a few lines of addition around:
https://github.com/syslog-ng/syslog-ng/blob/a66a82dd6fe04113bd48d1b05c6737550de57869/lib/persist-state.c#L899
And of course a new unit test. We can test if the same place reused by comparing the original and new `PersistEntryHandle`-s

Marking as good first issue.

alexb271:
lib:  make persist_state_alloc_string() reuse the same entry if possible
Resolves #3337

### DIFF
--- a/lib/persist-state.c
+++ b/lib/persist-state.c
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2023 One Identity LLC.
  * Copyright (c) 2002-2013 Balabit
  * Copyright (c) 1998-2012 Balázs Scheidler
  *
@@ -889,26 +890,34 @@ void
 persist_state_alloc_string(PersistState *self, const gchar *persist_name, const gchar *value, gssize len)
 {
   PersistEntryHandle handle;
-  SerializeArchive *sa;
+  SerializeArchive *serializer;
   GString *buf;
   gboolean success;
   gpointer block;
+
+  gsize lookup_size;
+  guint8 lookup_result_version;
 
   if (len < 0)
     len = strlen(value);
 
   buf = g_string_sized_new(len + 5);
-  sa = serialize_string_archive_new(buf);
-
-  success = serialize_write_cstring(sa, value, len);
+  serializer = serialize_string_archive_new(buf);
+  success = serialize_write_cstring(serializer, value, len);
   g_assert(success == TRUE);
+  serialize_archive_free(serializer);
 
-  serialize_archive_free(sa);
+  handle = persist_state_lookup_entry(self, persist_name, &lookup_size, &lookup_result_version);
 
-  handle = persist_state_alloc_entry(self, persist_name, buf->len);
+  if (!handle || lookup_size < buf->len)
+    {
+      handle = persist_state_alloc_entry(self, persist_name, buf->len);
+    }
+
   block = persist_state_map_entry(self, handle);
   memcpy(block, buf->str, buf->len);
   persist_state_unmap_entry(self, handle);
+
   g_string_free(buf, TRUE);
 }
 
--- a/lib/tests/test_persist_state.c
+++ b/lib/tests/test_persist_state.c
@@ -1,4 +1,5 @@
 /*
+ * Copyright (c) 2023 One Identity LLC.
  * Copyright (c) 2010-2016 Balabit
  * Copyright (c) 2010-2014 Balázs Scheidler
  * Copyright (c) 2014 Viktor Tusa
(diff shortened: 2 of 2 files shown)


news/feature-4432.md
```diff
@@ -0,0 +1,4 @@
+Make `persist_state_alloc_string` reuse the same entry if possible.
+
+The `persist_state_alloc_string` function has been modified to first call `persist_state_lookup_entry` and check if an allocation with the same name already exists. If it does and also the allocation is large enough to hold the target message then instead of creating a new allocation the existing one is used.
```
bazsi, line 3: Does this change have a user visible consequence? If so that should be documented in the news file. If no, the we might not need a news entry.

alexb271: No, this change has no user visible consequence.

lib/persist-state.c
```diff
@@ -889,26 +890,43 @@ void
...
+  SerializeArchive *serializer;
   GString *buf;
+  gsize buflen;
   gboolean success;
   gpointer block;
 
+  PersistEntryHandle lookup_handle;
+  gsize lookup_size;
+  guint8 lookup_result_version;
+  gpointer lookup_block;
```
MrAnno, line 902: Feel free to declare variables where you assign values to them first. It might be a matter of preference/taste, I just wanted to let you know with this comment that these old-style declarations at the top of functions are not requirements.

lib/persist-state.c
```diff
@@ -889,26 +890,43 @@ void
...
+      memcpy(lookup_block, buf->str, buf->len);
+      persist_state_unmap_entry(self, lookup_handle);
+    }
+  else
+    {
+      handle = persist_state_alloc_entry(self, persist_name, buf->len);
+      block = persist_state_map_entry(self, handle);
+      memcpy(block, buf->str, buf->len);
+      persist_state_unmap_entry(self, handle);
+    }
```
MrAnno, line 928: Optional:

You can reduce a bit of duplication by making only the allocation conditional, for example:

```
PersistEntryHandle handle = persist_state_lookup_entry(...);
if (!handle || lookup_size < buflen)
  handle = persist_state_alloc_entry(self, persist_name, buf->len);

gpointer block = persist_state_map_entry(self, handle);
memcpy(...);
persist_state_unmap_entry(...);
```

lib/tests/test_persist_state.c
```diff
@@ -206,6 +207,25 @@ Test(persist_state, test_persist_state_remove_entry)
...
+  persist_state_alloc_string(state, "test", "test_longer", -1);
+  persist_state_lookup_entry(state, "test", &size_1, &version);
+
+  persist_state_alloc_string(state, "test", "test_short", -1);
+  persist_state_lookup_entry(state, "test", &size_2, &version);
+
+  cr_assert_eq(size_1, size_2, "allocation versions do not match");
+
+  cancel_and_destroy_persist_state(state);
+}
```
MrAnno, line 227: Thanks for writing a test.

Can we also add a test case, where test "reuse" scenarios?
You could look up the entries with the non-string API and compare their `PersistEntryHandle` to check whether they are identical.

MrAnno (commented):
Nice work.

All of my comments are optional. Let me know if you want to work on any of them, otherwise, I'd merge the PR.

MrAnno:
(Adding `Fixes #3337`/`Resolves #3337` to the description of the PR will autoclose the issue once the PR merged.)

lib/persist-state.c
MrAnno: I think this would worth fixing though.

lib/persist-state.c
alexb271: Thank you for letting me know! For now I'll leave it as is to avoid two different styles within a single function.

lib/tests/test_persist_state.c
alexb271: Good idea, I'll see what I can do.

lib/persist-state.c
```diff
@@ -889,26 +890,36 @@ void
...
+  buf = g_string_sized_new(buflen);
+  serializer = serialize_string_archive_new(buf);
+  success = serialize_write_cstring(serializer, value, len);
   g_assert(success == TRUE);
+  serialize_archive_free(serializer);
 
-  serialize_archive_free(sa);
+  handle = persist_state_lookup_entry(self, persist_name, &lookup_size, &lookup_result_version);
+
+  if (!handle || lookup_size < buflen)
```
MrAnno, line 914: We should use `buf->len` instead of `buflen` here, as it's `serialize_write_cstring()` responsibility to append the `GString` buffer, which may end up being larger than our pre-calculated assumption.

MrAnno (commented):
Thank you.

I have one last comment. Sorry, I didn't notice earlier.

MrAnno:
Thanks.

PR #4435, opened 2023-04-19
`syslog-ng`: add `--config-id` command line option

MrAnno:
Similarly to `--syntax-only`, this command line option parses the configuration
and then prints its ID before exiting.

It can be used to query the ID of the current configuration persisted on
disk.

Note:
`--syntax-only` could have been used for the same functionality.
In that case, `--config-id` would just be an alias.


alltilla (approved):
Cool!

PR #4436, opened 2023-04-20
cmake: Fixed some file permissions during installation

HofiOne:

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -5,7 +5,7 @@ set(sysconfdir ${SYSLOG_NG_PATH_SYSCONFDIR})
 set(localstatedir ${SYSLOG_NG_PATH_LOCALSTATEDIR})
 
 configure_file (${CMAKE_CURRENT_SOURCE_DIR}/update-patterndb.in ${CMAKE_CURRENT_BINARY_DIR}/update-patterndb)
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/update-patterndb DESTINATION bin)
+install(PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/update-patterndb DESTINATION bin)
 install(DIRECTORY DESTINATION etc/patterndb.d)
 
 set(python_venvdir ${SYSLOG_NG_PYTHON_VENV_DIR})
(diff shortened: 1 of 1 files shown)


MrAnno (commented):
~Doesn't using `install(SCRIPT)` instead of `install(FILES)` fix your permission issues?~

Sorry, I misread the docs.
What permissions do these files have in the source directory in the build environment, where the install() command produces wrong permissions?

MrAnno:
`configure_file()` defaults to `USE_SOURCE_PERMISSIONS`, so the exec perms should be there.

What permissions does `syslog-ng-update-virtualenv.in` have in your source dir?
Did you mount the source into a VM or emulator?

HofiOne:
> `configure_file()` defaults to `USE_SOURCE_PERMISSIONS`, so the exec perms should be there.
> 
> What permissions does `syslog-ng-update-virtualenv.in` have in your source dir? Did you mount the source into a VM or emulator?
> 
> `x` is kept in my environment.
> 
> ```
> $ ls -al scripts/
> -rwx------ 1 anno users 2992 márc  29 13.55 syslog-ng-update-virtualenv.in*
> ```
> 
> ```
> $ ls -al _install/bin/
> ---x------ 1 anno users   3059 ápr   20 12.49 syslog-ng-update-virtualenv
> ```

full native build on macOS ARM macBook and x85 mac mini , no venv included here

probably this is a `configure_file()` bug of cmake, as it had no effect in various tried scenarios, also as `install()` was already used, therefore I stayed with `install()` that works nicely on the tried boxes



original result without this fix 

syslog-ng-update-virtualenv 
---x------

update-patterndb
-rw-r--r-



after the fix

syslog-ng-update-virtualenv 
-rwxr-xr-x

update-patterndb
-rwxr-xr-x


so, 
- update-patterndb had no execute bit set at all
- syslog-ng-update-virtualenv  had no other bit set than execute for the owner

MrAnno (changes requested):
Please use `install(PROGRAMS)` without any additional permission parameters, instead of working around file permissions.

> The PROGRAMS form is identical to the FILES form except that the default permissions for the installed file also include OWNER_EXECUTE, GROUP_EXECUTE, and WORLD_EXECUTE.

MrAnno (approved):
Thanks.

PR #4438, opened 2023-04-20
afprog: Fixed possible freezing on some OSes

HofiOne:
There were 2 issues in the `_close_all_fd` implementation

- iterating through the file handler numbers used an `int` though the init value is an `rlim_t` (that can be an `unsigned long long`) that can have an `RLIM_INFINITY` value, so an overflow could happen, therefore no file handlers were closed at all that led to a freeze in `_wait_for_child_closing_fds`
- as the `rlp.rlim_max` can be `RLIM_INFINITY` that is an extremely large number in that case, calling close() for such a huge number of handlers could take an extremely long time.

### DIFF
--- a/modules/afprog/afprog.c
+++ b/modules/afprog/afprog.c
@@ -95,21 +95,28 @@ _exec_program(const gchar *cmdline)
 static void
 _close_all_fd(void)
 {
+  const rlim_t min_range = 10000;
   struct rlimit rlp;
   if (getrlimit(RLIMIT_NOFILE, &rlp) < 0)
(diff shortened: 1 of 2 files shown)


MrAnno (approved):
Thank you for fixing this.

MrAnno:
Could you write a bugfix news entry for this, please? :)

HofiOne:
> Could you write a bugfix news entry for this, please? :)

upss, sorry, done.

MrAnno (approved):

PR #4441, opened 2023-04-24
Retrieve packet timestamp from the kernel

bazsi:
This PR adds support for SO_TIMESTAMPNS getsockopt to retrieve an accurate timestamp from the kernel to retain packet ordering.

### DIFF
--- a/lib/transport/transport-socket.c
+++ b/lib/transport/transport-socket.c
@@ -86,13 +86,40 @@ _determine_proto(gint fd, gint address_family)
   return result;
 }
 
+void
+log_transport_socket_parse_cmsg_method(LogTransportSocket *s, struct cmsghdr *cmsg, LogTransportAuxData *aux)
+{
+  return;
+}
+
 static void
 log_transport_socket_init_instance(LogTransportSocket *self, gint fd)
 {
   log_transport_init_instance(&self->super, fd);
   self->address_family = _determine_address_family(fd);
   self->proto = _determine_proto(fd, self->address_family);
+  self->parse_cmsg = log_transport_socket_parse_cmsg_method;
 }
+
+#if defined(SYSLOG_NG_HAVE_CTRLBUF_IN_MSGHDR)
+
+static void
+_parse_cmsg_to_aux(LogTransportSocket *self, struct msghdr *msg, LogTransportAuxData *aux)
+{
+  struct cmsghdr *cmsg;
+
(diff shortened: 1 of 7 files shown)


bazsi:
Add the same to unix-dgram. Maybe to stream oriented transports as well (TCP, unix-stream)

MrAnno:
What type of packet order retainment does this provide?

My understanding is that in addition to getting a more accurate reception timestamp, this may fix some local kernel-user space reordering, for example, when SO_REUSEPORT is used and there are multiple consumers.

lib/transport/transport-socket.c
```diff
@@ -167,7 +161,14 @@ log_transport_dgram_socket_read_method(LogTransport *s, gpointer buf, gsize bufl
...
   while (rc == -1 && errno == EINTR);
-  return _extract_from_msghdr_method(self, rc, &msg, aux);
+  rc = _extract_from_msghdr_method(self, rc, &msg, aux);
+
+  if (rc == 0)
+    {
+      /* DGRAM sockets should never return EOF, they just need to be read again */
+      rc = -1;
+      errno = EAGAIN;
+    }
```
alltilla: [note]

This change removed the `return` statement of this function. It gets added back in a subsequent commit, though.

lib/transport/transport-socket.c
```diff
@@ -124,13 +124,7 @@ _parse_cmsg_to_aux(LogTransportSocket *self, struct msghdr *msg, LogTransportAux
...
 _extract_from_msghdr_method(LogTransportSocket *self, gint rc, struct msghdr *msg, LogTransportAuxData *aux)
 {
-  if (rc == 0)
-    {
-      /* DGRAM sockets should never return EOF, they just need to be read again */
-      rc = -1;
-      errno = EAGAIN;
-    }
-  else if (rc > 0)
+  if (rc > 0)
```
alltilla, line 146: The signature of this function suggests, that it might change the `rc`, which is not the case. I think it would be cleaner if we moved the `rc > 0` to the call site, make this a `void` function, and remove `rc` from its arguments.

alltilla (commented):
I have reviewed this except the last commit. Sharing my comments so far.

MrAnno:
@kira-syslogng do stresstest

MrAnno:
@kira-syslogng do stresstest

kira-syslogng:

Issue #4439, opened 2023-04-23
high memory usage of add-contextual-data


gpfei:
# syslog-ng
## Version of syslog-ng

4.1.1

## Platform

OpenEuler 22.04 LTS (similar with CentOS 9)

## Steps to reproduce

generate a large csv file, for example, about 1M lines and 100M bytes.

## Configuration

```
parser p_foo {
    add-contextual-data(
        selector("$someField"),
        database("/etc/syslog-ng/conf.d/large-db.csv"),
    );
};
```

## command

`LD_PRELOAD=/usr/lib64/libjemalloc.so.2 syslog-ng -F -p /var/run/syslogd.pid`

## Input and output logs (if possible)

`RES` value of `syslog-ng` in `top` output is about 1.3 GB, which is far more than expected.

Thanks.

bazsi:
I've just opened a PR to decrease the memory use somewhat in #4444 still not perfect, but seems much better for me.

bazsi:
This should have been improved by the merge of #4444. Closing.

PR #4444, opened 2023-04-25
Fix memory use of add contextual data

bazsi:
The PR significantly decreases the memory use for add-contextual-data() (e.g. 40%) as reported in #4439 

I'd appreciate some testing with your original CSV database, so testing would absolutely be welcome. 

Note that add-contextual-data()'s data representation is not really optimized for memory usage, I could easily slash it almost half, but anything more would seems more complicated. 

db-parser() is difficult to use but it's performance/memory use is difficult to beat. If you are interested I could help you on Axoflow's discord server for a more interactive chat, https://discord.gg/E65kP9aZGm

Depending on how you deploy syslog-ng, you could use the PR specific binaries produced by github or compile from source, etc.

### DIFF
--- a/lib/template/templates.c
+++ b/lib/template/templates.c
@@ -211,6 +211,13 @@ log_template_compile(LogTemplate *self, const gchar *template, GError **error)
   return result;
 }
 
+void
+log_template_forget_template_string(LogTemplate *self)
+{
+  g_free(self->template);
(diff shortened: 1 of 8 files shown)


kira-syslogng:
Build FAILURE

MrAnno (approved):

PR #4447, opened 2023-04-26
kubernetes: add `input_events_total` and `input_event_bytes_total` metrics

alltilla:
Also added the `increment()` option to `metrics-probe()`.


modules/metrics-probe/metrics-probe.c
```diff
@@ -143,6 +153,38 @@ _lookup_stats_counter(MetricsProbe *self, LogMessage *msg)
...
+_calculate_increment(MetricsProbe *self, LogMessage *msg)
+{
+  if (!self->increment_template)
+    return 1;
+
+  ScratchBuffersMarker marker;
+  GString *increment_buffer = scratch_buffers_alloc_and_mark(&marker);
+
+  const gchar *increment_str = _format_increment_template(self, msg, increment_buffer);
+  gint64 increment = strtol(increment_str, NULL, 10);
```
MrAnno, line 181: Shouldn't we use `strtoll` for `gint64/gssize`?

modules/metrics-probe/metrics-probe.c
```diff
@@ -153,7 +195,8 @@ _process(LogParser *s, LogMessage **pmsg, const LogPathOptions *path_options, co
             evt_tag_msg_reference(*pmsg));
 
   StatsCounterItem *counter = _lookup_stats_counter(self, *pmsg);
-  stats_counter_inc(counter);
+  gint64 increment = _calculate_increment(self, *pmsg);
```
MrAnno, line 198: `gint64` and `gssize` are not always the same.
I think we should use  `gssize` here, which is the base type of stats counters.

MrAnno (commented):
Thank you. :)

modules/metrics-probe/metrics-probe.c
alltilla: `gssize` + `strtol()` would be the best right?

MrAnno: I think gssize and strtoll would be the best, and we could let it overflow on 32-bit.

(Or scanf() with the gssize formatter, but that's expensive.)

MrAnno (approved):
Thank you.

PR #4450, opened 2023-04-27
Update the sample Debian config to not throw warning

smortex:
When installing the Debian package, the installed sample configuration
file include some legacy settings that cause a warning:

> WARNING: Your configuration file uses an obsoleted keyword, please update your configuration; keyword='stats_freq', change='Use the stats() block. E.g. stats(freq(1));', location='/etc/syslog-ng/syslog-ng.conf:10:4'

Adjust the sample configuration to use the updated syntax instead.


kira-syslogng:
do nothing -> CI won't start)

bazsi (approved):

bazsi:
Ok to test.

bazsi (approved):

PR #4452, opened 2023-04-28
Value pairs upper and lower transforms

bazsi:
This is an attempt to add support for lower/upper transformations in value-pairs expressions as discussed in #4422 

This is how to use it:

```file("logfile" template("$(format-json * --lower)\n")); ```

Or, a slightly different example:

```file("logfile" template("$(format-json * foo.bar=value --rekey foo.* --upper)\n"))```


alltilla (approved):
LGTM! Fixed the style-check with a force push.

PR #4459, opened 2023-05-05
python: fix reference counting of generator function 

alltilla:
We are parsing the function pointer a couple lines above with "O".

> O (object) [PyObject *]
> Store a Python object (without any conversion) in a C object pointer. The C program thus receives the actual object that was passed. The object’s reference count is not increased. The pointer stored is not NULL.

Its reference count is not increased with the parsing, however we decrease its reference in `python_config_generator_free()`.
This causes a problem, when we are reloading and re-registering the same function.

### DIFF
--- a/modules/python/python-confgen.c
+++ b/modules/python/python-confgen.c
@@ -121,6 +121,7 @@ py_register_config_generator(PyObject *self, PyObject *args, PyObject *kwargs)
       return NULL;
     }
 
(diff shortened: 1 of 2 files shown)


kira-syslogng:
Build FAILURE

alltilla:

MrAnno (approved):
This must have been hard to debug. Nice catch.

PR #4463, opened 2023-05-08
examples: fix memory leak

alexb271:
This patch fixes a memory leak in case the destination worker fails in the _dw_insert method.


MrAnno (approved):

MrAnno:
Thanks.

PR #4465, opened 2023-05-09
configure: add --with-python-venv-dir option

MrAnno:
This can be useful in containerized environments, where the venv is built into the image, but users want to mount an empty statedir.


alltilla (approved):

alltilla:
Cool!

Maybe we could mention this in the packaging section of the news file.

PR #4467, opened 2023-05-10
python-modules: add register_jinja_config_generator convenience fn

alltilla:
Example usage:
```
python {
import syslogng

syslogng.register_jinja_config_generator(context="destination", name="my_dest", template=r"""

file({{ args.path }});

""")
};

destination d_my_dest{
  my-dest(path("/dev/stdout"));
};
```

TODO: make it possible to load `.jinja` file.

### DIFF
--- a/modules/python-modules/syslogng/confgen.py
+++ b/modules/python-modules/syslogng/confgen.py
@@ -20,10 +20,22 @@
 # COPYING for details.
 #
 #############################################################################
-# pylint: disable=unused-import
 
 try:
     from _syslogng import register_config_generator
 except ImportError:
     def register_config_generator(context, name, config_generator):
+        pass
+
+try:
+    import jinja2
+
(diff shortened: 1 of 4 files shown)


modules/python-modules/syslogng/confgen.py
```diff
@@ -20,10 +20,22 @@
...
+    import jinja2
+
+    def register_jinja_config_generator(context, name, template):
+        def jinja_confgen_fn(args):
+            jinja2_env = jinja2.Environment()
+            jinja2_template = jinja2_env.from_string(template)
+            return jinja2_template.render(args=args)
+        register_config_generator(context, name, jinja_confgen_fn)
+except ImportError:
+    def register_jinja_config_generator(context, name, template):
```
bazsi, line 40: I think this should raise an exception if called as jinja2 might be unavailable if the user is not using a virtualenv.

PR #4469, opened 2023-05-10
Fix syslog-ng-cfg-db.py

smortex:
I just read about `./contrib/config_option_database/syslog-ng-cfg-db.py` which I was not aware of and tried to use it on the master `branch` to have some insights on the new parsers of syslog-ng 4.2.0 but encountered errors (and this makes me think we can improve the vim syntax maintenance using this).

Once the data has been generated and cached, it seems not to be refreshed, so you have to `git clean -fdx` between tests to trigger the errors.

While the first commit is a quick workaround due to the fact the script tries to parse C code in a too quick way, the second makes me more skeptical, so Cc @bazsi who may have not added this code on purpose and may want to raise a point.

### DIFF
--- a/contrib/config_option_database/utils/OptionParser.py
+++ b/contrib/config_option_database/utils/OptionParser.py
@@ -100,7 +100,7 @@ def _get_resolve_db():
     root_dir = Path(__file__).parents[3]
     if not resolve_db:
         resolve_db = {}
-        struct_regex = re.compile(r'CfgLexerKeyword[^;]*')
+        struct_regex = re.compile(r'CfgLexerKeyword.*?};')
         entry_regex = re.compile(r'{[^{}]+,[^{}]+}')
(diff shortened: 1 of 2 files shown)


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

alltilla:
Wow, cool to see someone use this :D

I have written this code a long time ago and it came out rather un-clean. I can imagine not being much fun to investigate and fix a bug in it. I have a branch which refactors this code as a whole and you inspired me to pick it up again.

Will check this PR tomorrow, thanks for the contribution!

alltilla:
@kira-syslogng ok to test

alltilla:
Btw, if I femember correctly -r or -R drops the cache and rebuilds the db.

PR #4476, opened 2023-05-14
Loggen performance improvements

bazsi:
This is split off from #3966, I've discovered that loggen  was only capable of doing  ~250k msg/sec per thread, which is not enough for the latest round of scalability improvements in syslog-ng.

### DIFF
--- a/tests/loggen/logline_generator.c
+++ b/tests/loggen/logline_generator.c
@@ -23,6 +23,7 @@
 
 #include "logline_generator.h"
 #include "loggen_helper.h"
+#include "loggen_plugin.h"
 
 #include <stdio.h>
 #include <string.h>
@@ -97,7 +98,12 @@ prepare_log_line_template(int syslog_proto, int framing, int message_length, cha
 }
 
 int
-generate_log_line(char *buffer, int buffer_length, int syslog_proto, int thread_id, unsigned long seq)
+generate_log_line(ThreadData *thread_context,
+                  char *buffer, int buffer_length,
+                  int syslog_proto,
+                  int thread_id,
+                  unsigned long rate,
+                  unsigned long seq)
 {
   if (!buffer)
     {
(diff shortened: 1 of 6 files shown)


tests/loggen/logline_generator.c
```diff
@@ -110,16 +111,29 @@ generate_log_line(char *buffer, int buffer_length, int syslog_proto, int thread_
...
   struct timeval now;
-  gettimeofday(&now, NULL);
+  static struct timeval ts_formatted = {0};
   struct tm tm;
-  char stamp[32];
-  localtime_r(&now.tv_sec, &tm);
-  int len = strftime(stamp, sizeof(stamp), "%Y-%m-%dT%H:%M:%S", &tm);
-  memcpy(&buffer[pos_timestamp2], stamp, len);
+  static char stamp[32] = {0};
+  static int len;
```
MrAnno, line 117: These globals are written by multiple threads, the race can produce invalid timestamps.

We could use TLS here, or just cache the timestamp on the caller side.

bazsi: Good catch. I was concentrating so much on my single-threaded case...

Anyway, I just used a mutex here, generating that timestamp should not happen very frequently now, a mutex is just fine.

tests/loggen/logline_generator.c
```diff
@@ -110,16 +113,31 @@ generate_log_line(char *buffer, int buffer_length, int syslog_proto, int thread_
...
+          len = strftime(stamp, sizeof(stamp), "%Y-%m-%dT%H:%M:%S", &tm);
+          ts_formatted = now;
+
+          if (syslog_proto)
+            format_timezone_offset_with_colon(stamp, sizeof(stamp), &tm);
+        }
+      g_mutex_unlock(&generator_lock);
+    }
 
+  memcpy(&buffer[pos_timestamp2], stamp, len);
```
MrAnno, line 140: I think the `memcpy()` call should be protected as well (read-write race).

Since now this will become expensive, I think caching on the caller side or TLS would be suitable.

bazsi: fixed.

bazsi:
rebased against master, now we are using thread specific variables for our timestamp cache.

MrAnno (approved):

PR #4477, opened 2023-05-15
format-json: fix adding unnecessary comma when failing to cast 

alltilla:


bazsi (approved):
LGTM

PR #4478, opened 2023-05-15
correlation: make grouping-by clone ids unique

### LINKED ISSUE #4443: `grouping-by()`, `group-lines()`: persist name is dependent on the number of times it gets cloned
# syslog-ng
## Version of syslog-ng
4.1.1

# Issue
`group-lines()` and `grouping-by()` filters' persist names depend on how many times they are getting cloned.

When cloning such a parser, the `clone_id` of the cloned instance gets the one from `self`,  and `self`'s `clone_id` is incremented. This is done so the original instance can be cloned multiple times without 2 of its clones having the same `clone_id`. https://github.com/syslog-ng/syslog-ng/blob/syslog-ng-4.1.1/modules/correlation/groupingby.c#L539

Let's say we have a config, where the parser instance gets cloned 2 times. It has `clone_id=2` in its persist name and stores its values in the persist with that name. We add another reference/user of the parser in the config, and reload. Now it gets `clone_id=3`, and it won't find its persist entry from before reload and starts a new, however the new clone gets an already filled persist entry.

We should fix this issue, while still guaranteeing that no 2 clones get the same `clone_id`.

alexb271:
Resolves #4443


modules/correlation/groupingby.c
```diff
@@ -192,6 +193,12 @@ _init(LogPipe *s)
   return grouping_parser_init_method(s);
 }
 
+static int next_clone_id()
+{
+  static guint id = 0;
```
alltilla, line 198: I don't think this will work.

This variable is shared between all `grouping-by()` instances, which is not good because the `clone_id` changes if a `grouping-by()` driver gets configured/unconfigured.

E.g.: `A` gets 0, `B` gets 1, `C` gets 2. We remove `A` and `B` from the config, so `C` will get 0 instead of 2.

alltilla (commented):
Please also fix `group-lines()`. Thanks!

alltilla:
I'm starting to feel like an incremental `clone_id` will never provide us the uniqueness we need. Maybe we should think about replacing it with something unique to each clone if that is possible.

alexb271:
I've changed the implementation to a clone_id that is incremented only among instances with a common ancestor. Perhaps that could work?

modules/correlation/groupingby.c
```diff
@@ -198,6 +203,11 @@ _clone(LogPipe *s)
   GroupingBy *self = (GroupingBy *) s;
   GroupingBy *cloned;
 
+  if (self->next_id == NULL)
+    {
+      self->next_id = g_atomic_rc_box_new0(guint);
```
alltilla, line 208: Our GLib min version is 2.32: https://github.com/syslog-ng/syslog-ng/blob/master/configure.ac#L43
`atomic_rc_box` is available from 2.58: https://docs.gtk.org/glib/func.atomic_rc_box_new.html

We should try to find something that does not require raising our GLib min version.

As far as I remember cloning happens during config initialization in the main thread so we don't need to account for thread safety.

alltilla (commented):
I like this solution :)

However while reviewing your previous one I realized that maybe incremental clone-ids will never solve our issue. I think the user can modify the config so that the new clone happens in the middle of the cloning chain. E.g.:

Let's say the config produces the following clone-id pairing:
A: 1, B: 2, C: 3

The user adds another reference of this parser (D), but it is placed in such a place in the config, where it get's cloned after B. We get the following ids:
A: 1, B: 2, D: 3, C: 4.

Or the user deletes the parser from B and C, and inserts it to D. We get the following ids:
A: 1, D: 2.

This means the clone-id is still dependent on other uses of the parser in the config.

If it is possible we should find a unique identifier of where the cloning happens, so we don't run into this problem. Not sure what it should be or if it is even possible. We should investigate what information is available from the context, where the cloning happens.

If we cannot find such info, I think your implementation is better than the one on master, so we could merge it, but it is still not complete, unfortunately.

modules/correlation/groupingby.c
MrAnno: `lib/atomic.h` and `lib/atomic-gssize.h` can be used, if needed; but I also believe that this does not have to be thread safe as we clone in the main thread during initialization.

alexb271: > Our GLib min version is 2.32: https://github.com/syslog-ng/syslog-ng/blob/master/configure.ac#L43 `atomic_rc_box` is available from 2.58: https://docs.gtk.org/glib/func.atomic_rc_box_new.html
> 
> We should try to find something that does not require raising our GLib min version.
> 
> As far as I remember cloning happens during config initialization in the main thread so we don't need to account for thread safety.

I have changed the implementation to a simple self-made one that does not rely on glib and also is not tgread safe.

If there's a better way to do this please let me know.

alexb271:
> I like this solution :)
> 
> However while reviewing your previous one I realized that maybe incremental clone-ids will never solve our issue. I think the user can modify the config so that the new clone happens in the middle of the cloning chain. E.g.:
> 
> Let's say the config produces the following clone-id pairing: A: 1, B: 2, C: 3
> 
> The user adds another reference of this parser (D), but it is placed in such a place in the config, where it get's cloned after B. We get the following ids: A: 1, B: 2, D: 3, C: 4.
> 
> Or the user deletes the parser from B and C, and inserts it to D. We get the following ids: A: 1, D: 2.
> 
> This means the clone-id is still dependent on other uses of the parser in the config.
> 
> If it is possible we should find a unique identifier of where the cloning happens, so we don't run into this problem. Not sure what it should be or if it is even possible. We should investigate what information is available from the context, where the cloning happens.
> 
> If we cannot find such info, I think your implementation is better than the one on master, so we could merge it, but it is still not complete, unfortunately.

Thanks!

I think a fully non-incremental solution could be difficult to implement. We cannot rely on the filter's order of occurance or line number in the config, therefore its ID should then be based on a semantic understanding of the config itself.

One shortcut would be to store (or hash) the configuration above the group-by() call, but then irrelevant changes in other sections would also invalidate the persist state.

A full solution would be to understand and store (or hash) the group-by() call's relevant context in the config file, such as the log path it is declared in, or any other possible context.

In my opinion that would be a larger project that might best be handled in a separate enhancement issue.

modules/correlation/groupingby.c
```diff
@@ -198,6 +203,11 @@ _clone(LogPipe *s)
   GroupingBy *self = (GroupingBy *) s;
   GroupingBy *cloned;
 
+  if (self->id_counter == NULL)
+    {
+      self->id_counter = id_counter_new();
+    }
```
alltilla, line 209: I think the lifecycle of the `id_counter` is more complex than necessary. Let me share some deeper level syslog-ng knowledge.

Managing the lifecycle of a driver's field is unfortunately pretty complex in syslog-ng. The fields can be allocated in `new()`, `init()`, `set_...()`, `clone()`, or sometimes the config parser allocates them for us (`string_list`, ...). They are usually freed in `deinit()` or `free()`. The calling order of these is also complex, a driver can be constructed by the config parser, it can be cloned, syslog-ng can reload, where a new set of drivers are initialized alongside the old ones, syslog-ng can have a failed reload, where the old config's drivers are reinitialized, etc... A driver must be working properly in case of `new()` -> `init()` -> `deinit()` -> `init()` -> `deinit()`, which happens in case of a failed reload. The general rule of thumb is that if it is possible to allocate something in `new()`, do it there, and deallocate it in `free()`, if allocating is only possible in `init()`, do it there and deallocate it in `deinit()`. It is complicated (although possible) to think through the lifecycle if we do not follow this rule of thumb.

I would suggest to allocate the `id_counter` in `new()` (also get the ID there) and `unref()` it in `free()`. In the `clone()` method you can `unref()` the newly created `id_counter` and replace it with `self`'s with a `ref()` (also getting a new ID for the clone there).

modules/correlation/id-counter.h
```diff
@@ -0,0 +1,39 @@
...
+typedef struct _IdCounter
+{
+  guint id;
+  guint ref_count;
+} IdCounter;
+
+IdCounter *id_counter_new(void);
+guint id_counter_get_next_id(IdCounter *self);
+IdCounter *id_counter_acquire(IdCounter *self);
+void id_counter_release(IdCounter *self);
```
alltilla, line 37: `acquire()` and `release()` is usually called `ref()` and `unref()` in syslog-ng:

<details>

```
$ git grep -o '^[^ ]*_ref(' | code -

lib/ack-tracker/ack_tracker_factory.c:ack_tracker_factory_ref(
lib/cfg-args.c:cfg_args_ref(
lib/cfg-block-generator.c:cfg_block_generator_ref(
lib/cfg-tree.c:log_expr_node_ref(
lib/control/control-command-thread.c:control_command_thread_ref(
lib/control/control-connection.c:control_connection_ref(
lib/dynamic-window-pool.c:dynamic_window_pool_ref(
lib/eventlog/src/evtctx.c:evt_ctx_ref(
lib/eventlog/src/evtrec.c:evt_rec_ref(
lib/filter/filter-expr.c:filter_expr_ref(
lib/filter/filter-re.c:filter_match_set_template_ref(
lib/gsockaddr.c:g_sockaddr_ref(
lib/healthcheck/healthcheck.c:healthcheck_ref(
lib/logmatcher.c:log_matcher_ref(
lib/logmsg/logmsg.c:log_msg_set_saddr_ref(
lib/logmsg/logmsg.c:log_msg_set_daddr_ref(
lib/logmsg/logmsg.c:log_msg_update_ack_and_ref(
lib/logmsg/logmsg.c:log_msg_ref(
lib/logmsg/nvtable.c:nv_table_ref(
lib/logpipe.c:log_pipe_ref(
lib/logqueue.h:log_queue_ref(
lib/logscheduler.c:log_scheduler_options_set_partition_key_ref(
lib/multi-line/multi-line-pattern.c:multi_line_pattern_ref(
lib/stats/stats-registry.c:_assert_when_internal_or_stores_different_ref(
lib/template/compiler.c:parse_msg_ref(
lib/template/templates.c:log_template_ref(
lib/transport/tls-context.c:tls_context_ref(
lib/transport/tls-verifier.c:tls_verifier_ref(
lib/transport/transport-aux-data.h:log_transport_aux_data_set_peer_addr_ref(
lib/transport/transport-aux-data.h:log_transport_aux_data_set_local_addr_ref(
lib/value-pairs/value-pairs.c:value_pairs_ref(
modules/add-contextual-data/context-info-db.c:context_info_db_ref(
modules/correlation/correlation-context.c:correlation_context_ref(
modules/correlation/correlation.c:correlation_state_ref(
modules/correlation/dbparser.c:log_db_parser_set_program_template_ref(
modules/correlation/pdb-program.c:pdb_program_ref(
modules/correlation/pdb-rule.c:pdb_rule_ref(
modules/java/native/java_machine.c:java_machine_ref(
modules/json/dot-notation.c:_compile_dot_notation_array_ref(
modules/json/dot-notation.c:_compile_dot_notation_member_ref(
modules/kafka/kafka-dest-driver.c:kafka_dd_set_key_ref(
modules/kafka/kafka-dest-driver.c:kafka_dd_set_message_ref(
modules/mqtt/destination/mqtt-destination.c:mqtt_dd_set_message_template_ref(
modules/python/python-options.c:python_option_ref(
modules/redis/redis.c:redis_dd_set_command_ref(
modules/timestamp/rewrite-fix-timezone.c:rewrite_fix_time_zone_set_zone_template_ref(
modules/timestamp/rewrite-set-timezone.c:rewrite_set_time_zone_set_zone_template_ref(
```

</details>

Can you please follow this trend here?

modules/correlation/id-counter.c
```diff
@@ -0,0 +1,58 @@
...
+
+IdCounter *id_counter_new(void)
+{
+  IdCounter *self = g_new(IdCounter, 1);
+  self->id = 0;
+  self->ref_count = 1;
+  return self;
+}
+
+guint id_counter_get_next_id(IdCounter *self)
```
alltilla, line 33: Unfortunately we cannot enforce this by our style checker because of a limitation of `astyle`, we tend to separate the return type and the function name to separate lines in the `.c` files. The reason is that we can find implementations a lot easier with grep if we do an anchored regexp, e.g.:
```
$ git grep "^log_threaded_dest_driver_init_instance"
lib/logthrdest/logthrdestdrv.c:log_threaded_dest_driver_init_instance(LogThreadedDestDriver *self, GlobalConfig *cfg)
```

Could you please do this for the new functions in this file? Thanks!

modules/correlation/id-counter.c
```diff
@@ -0,0 +1,58 @@
...
+void id_counter_release(IdCounter *self)
+{
+  if (self == NULL) return;
+
+  self->ref_count -= 1;
+
+  if (self->ref_count == 0)
+    {
+      g_free(self);
+      self = NULL;
```
alltilla, line 56: `self = NULL` only changes the local value, it does not affect the call site, so it is a no operation here.

modules/correlation/groupingby.c
```diff
@@ -30,7 +32,8 @@ typedef struct _GroupingBy
   FilterExprNode *where_condition_expr;
   FilterExprNode *having_condition_expr;
   gchar *prefix;
-  gint clone_id;
+  guint clone_id;
```
alltilla, line 35: `clone_id` is initialized to 0 because of `g_new0()`. Fortunately a newly created `IdCounter` returns 1 for the first `get_next_id()` call. This works, but the knowledge is in two different parts of the code. I would suggest to always set `clone_id` with the help of `id_counter`. This needs `id_counter` to be allocated in `new()` (there is another argument in my other comment to move it there) and we need to make sure that `IdCounter` returns 0 for the `get_next_id()` (initing its `id` field to -1 would suffice).

modules/correlation/id-counter.h
```diff
@@ -0,0 +1,39 @@
...
+#ifndef ID_COUNTER_H_INCLUDED
+#define ID_COUNTER_H_INCLUDED
+
+#include "glib.h"
+
+typedef struct _IdCounter
+{
+  guint id;
+  guint ref_count;
+} IdCounter;
```
alltilla, line 32: [optional]

These are technically private fields, so the struct definition should be in the `.c` file.

modules/correlation/id-counter.c
alexb271: Good catch, thank you.

modules/correlation/groupingby.c
alexb271: I think a newly created `IdCounter` will return 0 on the first call to `get_next_id()` because the private member `id` is initialized to 0 and and the `return self->id++` line in `get_next_id()` will only increment the value after returning it.

However, I have made the changes to allocate `IdCounter` in `new()` as you have suggested.

modules/correlation/id-counter.h
alexb271: Thanks for the suggestion, I have moved the struct definition to the `.c` file.

alltilla:
Thanks! :)

PR #4479, opened 2023-05-16
Replace GTimeVal by struct timespec

ylse:
GTimeVal in glib-2.0 is being deprecated since 2.62 because it is not
year-2038-safe. Replace it by struct timespec.

Signed-off-by: Ylse <27207371+ylse@users.noreply.github.com>


kira-syslogng:
do nothing -> CI won't start)

kira-syslogng:
do nothing -> CI won't start)

alltilla:

bazsi:
Can you order commits so that each would compile? that might mean that a lot of these needs to be squashed into a single commit (e.g. cached_g_current_time() prototype change will cause most of these changes to go into a single commit)

This would make it easier to bisect later, also it makes it more difficult to review, as we have non-compiling patches.

As a followup to this idea, it'd be great if we could increase UnixTimeStamp granularity to nanoseconds, at the moment it's microseconds only.

lib/logqueue.c
```diff
@@ -185,13 +185,13 @@ log_queue_check_items(LogQueue *self, gint *timeout, LogQueuePushNotifyFunc para
     {
       gint64 diff;
       gint new_buckets;
-      GTimeVal now;
+      struct timespec now;
 
-      g_get_current_time(&now);
+      timespec_get(&now, TIME_UTC);
```
bazsi, line 190: I would use clock_gettime() if possible. Honestly I didn't even know about timespec_get() and was looking for its implementation when I had to realize this is standard too. But seems to be more recent than clock_gettime() being part of C11, whereas clock_gettime() is from the 90s.

Also, I find it strange that I need to specify a timezone to an epoch based timestamp.

bazsi:
I am working on a somewhat different implementation of the same in #4741, so closing this instead of that.

PR #4483, opened 2023-05-16
afmongodb: add bulk operations support 2 mongo db destination

HofiOne:
As a final step for https://github.com/syslog-ng/syslog-ng/issues/3454 bulk insert is added via the following options

- `bulk`  (**yes**/no)  turns on [bulk insert ](http://mongoc.org/libmongoc/current/bulk.html)usage, `no` forces the old behavior (each log is inserted one by one into the MongoDB) 
- `bulk_unordered` (yes/**no**)  turns on/off [unordered MongoDB bulk operations](http://mongoc.org/libmongoc/current/bulk.html#unordered-bulk-write-operations)
- `bulk_unacknowledged` (yes/**no**)  turns on/off [unacknowledged MongoDB bulk operations](http://mongoc.org/libmongoc/current/bulk.html#unacknowledged-bulk-writes)


modules/afmongodb/afmongodb-worker.c
```diff
@@ -397,8 +470,17 @@ _worker_insert(LogThreadedDestWorker *s, LogMessage *msg)
...
+  if (owner->mongoc_opt_use_bulk)
+    {
+      if (self->bulk_op == NULL)
+        self->bulk_op = mongoc_collection_create_bulk_operation_with_opts(self->coll_obj, self->bson_opts);
+
+      mongoc_bulk_operation_insert(self->bulk_op, (const bson_t *)self->bson);
+      return LTR_QUEUED;
+    }
+  else
+    success = mongoc_collection_insert(self->coll_obj, MONGOC_INSERT_NONE,
```
MrAnno, line 482: We should also pass write-concern options to `mongoc_collection_insert()` to support acked/unacked/wmajority in non-bulk mode as well.

modules/afmongodb/afmongodb-private.h
```diff
@@ -44,6 +44,10 @@ typedef struct _MongoDBDestDriver
   const gchar *const_db;
   mongoc_uri_t *uri_obj;
   mongoc_client_pool_t *pool;
+
+  gboolean mongoc_opt_use_bulk;
+  gboolean mongoc_opt_unordered;
+  gboolean mongoc_opt_unacknowledged;
```
MrAnno, line 50: This struct is all about Mongo, please consider removing the `mongoc_opt` prefix.
Using the same name you introduced in the grammar would be the best in my opinion, for example:

`use_bulk, bulk_unordered, write_concern`

modules/afmongodb/afmongodb-worker.c
```diff
@@ -29,15 +29,62 @@
...
+      self->bson_opts = g_new0(bson_t, 1);
+      bson_t def_opts = BSON_INITIALIZER;
+      *self->bson_opts = def_opts;
+
+      if (owner->mongoc_opt_unordered)
+        BSON_APPEND_BOOL (self->bson_opts, "ordered", false);
+
+      if (owner->mongoc_opt_unacknowledged)
+        {
+          self->write_concern = mongoc_write_concern_new();
```
MrAnno, line 64: `write_concern` is not a bulk-related option. Let's construct it separately from bulk options.

modules/afmongodb/afmongodb-worker.c
```diff
@@ -146,13 +199,9 @@ _worker_connect(LogThreadedDestWorker *s)
...
     }
 
-
   if (!_check_server_status(self, read_prefs))
     {
-      mongoc_collection_destroy(self->coll_obj);
-      self->coll_obj = NULL;
-      mongoc_client_pool_push(owner->pool, self->client);
-      self->client = NULL;
+      _worker_disconnect(s);
```
MrAnno, line 204: This commit has a few refactor steps as well, like this one.
Please consider creating a separate refactor commit for them next time. :)

modules/afmongodb/afmongodb-worker.c
```diff
@@ -353,6 +402,30 @@ _vp_process_value(const gchar *name, const gchar *prefix, LogMessageValueType ty
   return FALSE;
 }
 
+static LogThreadedResult
+_worker_batch_flush(LogThreadedDestWorker *s, LogThreadedFlushMode expedite)
+{
+  MongoDBDestWorker *self = (MongoDBDestWorker *) s;
+
+  bson_error_t error;
+  int result = mongoc_bulk_operation_execute(self->bulk_op, NULL, &error);
```
MrAnno, line 411: `reply` is not optional:

https://mongoc.org/libmongoc/current/mongoc_bulk_operation_execute.html

modules/afmongodb/afmongodb-worker.c
```diff
@@ -65,6 +112,12 @@ _switch_collection(MongoDBDestWorker *self, const gchar *collection)
   if (!self->client)
     return FALSE;
 
+  if (self->bulk_op)
+    {
+      mongoc_bulk_operation_destroy(self->bulk_op);
+      self->bulk_op = NULL;
+    }
```
MrAnno, line 119: Maybe I misunderstood something, but `_switch_collection()` seems to be broken with bulk sending.

We are currently switching collections before each insertion, but bulk sending requires a `flush()` call in order to send out messages. With the current code, we'd lose the entire batch during collection switch  (when using a templated `collection()`).  We either should trigger a flush manually before switching collection (it can be done in insert()), or we should not support templates while operating in bulk.

I'd go with the first option and document that bulk sending is only efficient if the user sets a constant collection or a template that does not use too many collections within a reasonable time range.

MrAnno (commented):
This is a long-awaited improvement. Thanks for working on it, it is really cool! 👍🏻

modules/afmongodb/afmongodb-worker.c
HofiOne: actually it is ;)
the documentation says nothing about optionallity, it say, is always initialized, even upon failure
if this parameter is not presented than it will not be used, allocated, there's nothing to free via bson_destroy()
it is working nicely this way, i did not see a crash or leak at all

HofiOne: double checked meanwhile in the mongo-c-driver source
reply safely can be NULL, it is always guarded by `_mongoc_bson_init_if_set (reply);`

modules/afmongodb/afmongodb-worker.c
HofiOne: good point, missed that
done

modules/afmongodb/afmongodb-parser.c
```diff
@@ -33,6 +33,9 @@ static CfgLexerKeyword afmongodb_keywords[] =
   { "mongodb", KW_MONGODB },
   { "uri", KW_URI },
   { "collection", KW_COLLECTION },
+  { "bulk", KW_BULK },
+  { "bulk_unordered", KW_BULK_UNORDERED },
+  { "bulk_unacknowledged", KW_BULK_UNACKNOWLEDGED },
```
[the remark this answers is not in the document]
HofiOne: going with `write-concern` and trying to add support of the 4th `n` param as well

modules/afmongodb/afmongodb-worker.c
```diff
@@ -29,15 +29,62 @@
...
+      bson_t def_opts = BSON_INITIALIZER;
+      *self->bson_opts = def_opts;
+
+      if (owner->mongoc_opt_unordered)
+        BSON_APPEND_BOOL (self->bson_opts, "ordered", false);
+
+      if (owner->mongoc_opt_unacknowledged)
+        {
+          self->write_concern = mongoc_write_concern_new();
+          mongoc_write_concern_set_w(self->write_concern, 0);
```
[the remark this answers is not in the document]
HofiOne: missed the define for these, thanks

modules/afmongodb/afmongodb-worker.c
MrAnno: 2 rows below:
```
error: An *optional* location for a [bson_error_t] or NULL.
```

What is not documented is subject to change. I would go the safe route.

modules/afmongodb/afmongodb-worker.c
HofiOne: actually it is not ONLY bulk related, but I see your point, will separate them

modules/afmongodb/afmongodb-worker.c
HofiOne: usually, I do, but  _worker_disconnec had changed too, therefore I did not want to add a version that reflects the changes duplicated again here than replace the whole block separately

modules/afmongodb/afmongodb-worker.c
HofiOne: yeep, I saw that too, but that documentation is outdated, and full of wrong information, that is why I started to check the source code itself for everything I had doubts about
as you wish, but it is a unnecessary overhead here both in speed and allocations

modules/afmongodb/afmongodb-worker.c
HofiOne: upps, i did not notice the switch in the insert, i just saw the one in connect
will try to fix this

modules/afmongodb/afmongodb-worker.c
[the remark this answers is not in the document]
MrAnno: Write concern numeric values are still hard-coded, they are in the grammar now.

We can't really include mongoc in the grammar, as we should be able to generate the grammar (the tarball) on machines where mongoc is not available.

modules/afmongodb/afmongodb-worker.c
```diff
@@ -29,15 +29,79 @@
...
+      self->bson_opts = g_new0(bson_t, 1);
+      bson_t def_opts = BSON_INITIALIZER;
+      *self->bson_opts = def_opts;
+
+      if (FALSE == BSON_APPEND_BOOL(self->bson_opts, "ordered", false == owner->bulk_unordered))
+        msg_error("Error seting bulk option",
+                  evt_tag_str("option", "ordered"),
+                  evt_tag_str("driver", owner->super.super.super.id));
+
+      if (self->write_concern)
```
MrAnno, line 50: `write_concern` is never null, this should probably be a g_assert().

modules/afmongodb/afmongodb-worker.c
```diff
@@ -29,15 +29,79 @@
...
+static LogThreadedResult _do_bulk_flush(MongoDBDestWorker *self);
+
+static void
+_compose_bulk_op_options(MongoDBDestWorker *self)
+{
+  MongoDBDestDriver *owner = (MongoDBDestDriver *) self->super.owner;
+
+  if (owner->use_bulk)
+    {
+      self->bson_opts = g_new0(bson_t, 1);
```
MrAnno, line 41: Optional:
`bson_new()` and `bson_destroy()` may be easier as we could avoid the ` bson_destroy(self->bson_opts); g_free(self->bson_opts);` 2 step free.

modules/afmongodb/afmongodb-worker.c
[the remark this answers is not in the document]
HofiOne: sorry, something is wrong here, probably you see the old version somehow, this part is totally rewritten, could you please double check

modules/afmongodb/afmongodb-worker.c
```diff
@@ -388,17 +482,38 @@ _worker_insert(LogThreadedDestWorker *s, LogMessage *msg)
             evt_tag_value_pairs("message", owner->vp, msg, &options),
             evt_tag_str("driver", owner->super.super.super.id));
 
-
   if (!owner->collection_is_literal_string)
     {
+      g_string_assign(self->last_collection, self->collection->str);
```
MrAnno, line 487: last_collection is not really useful in the MongoDB worker struct, as it is local information needed for a very short time.

`scratch_buffers_alloc_and_mark()` or `scratch_buffers_alloc()` can be used to achieve the same performance.

modules/afmongodb/afmongodb-worker.c
```diff
@@ -388,17 +482,38 @@ _worker_insert(LogThreadedDestWorker *s, LogMessage *msg)
...
+      if (self->bulk_op == NULL)
+        {
+          msg_error("Failed to create MongoDB bulk operation",
+                    evt_tag_int("time_reopen", self->super.time_reopen),
+                    evt_tag_str("driver", owner->super.super.super.id));
+          return LTR_ERROR;
+        }
+
+      mongoc_bulk_operation_insert(self->bulk_op, (const bson_t *)self->bson);
+      return LTR_QUEUED;
```
MrAnno, line 512: Optional: I'd extract this to a function so that `_worker_insert()` could look like this:

```
if (owner->use_bulk)
   _build_insert();
else
  mongoc_collection_insert();

```

news/feature-4483.md
```diff
@@ -0,0 +1,5 @@
+`afmongodb`: Bulk MongoDB insert is added via the following options
+
+- `bulk`  (**yes**/no)  turns on [bulk insert ](http://mongoc.org/libmongoc/current/bulk.html)usage, `no` forces the old behavior (each log is inserted one by one into the MongoDB)
```
MrAnno, line 3: Does bulk sending have any known side effects?

(I'm asking this because we're changing the default value here.)

modules/afmongodb/afmongodb-worker.c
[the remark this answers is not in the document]
MrAnno: The latest version still uses hard-coded values in the grammar file, but I think we can't do much about it if we don't want to map their values manually.

MrAnno (commented):
LGTM otherwise.

modules/afmongodb/afmongodb-worker.c
```diff
@@ -29,15 +29,79 @@
...
+{
+  MongoDBDestDriver *owner = (MongoDBDestDriver *) self->super.owner;
+
+  if (owner->use_bulk)
+    {
+      self->bson_opts = g_new0(bson_t, 1);
+      bson_t def_opts = BSON_INITIALIZER;
+      *self->bson_opts = def_opts;
+
+      if (FALSE == BSON_APPEND_BOOL(self->bson_opts, "ordered", false == owner->bulk_unordered))
```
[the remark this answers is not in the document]
HofiOne: because it is hard to read :) 
but, yepp, it is already used in the same unit elswhere, better to be consistent, will replace it

modules/afmongodb/afmongodb-worker.c
```diff
@@ -29,15 +29,79 @@
...
+        msg_error("Error seting bulk option",
+                  evt_tag_str("option", "ordered"),
+                  evt_tag_str("driver", owner->super.super.super.id));
+
+      if (self->write_concern)
+        if (FALSE == mongoc_write_concern_append(self->write_concern, self->bson_opts))
+          msg_error("Error seting bulk option",
```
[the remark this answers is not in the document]
HofiOne: good old kopipeszt, thanks!

modules/afmongodb/afmongodb-worker.c
HofiOne: _compose_write_concern will crash already if it is cannot be allocated, removing instead

modules/afmongodb/afmongodb-worker.c
HofiOne: good point

news/feature-4483.md
HofiOne: good question, probably we could stay with the old behavior by default.
i guess i switch back to the old one, just to be sure not breaking anything

modules/afmongodb/afmongodb-worker.c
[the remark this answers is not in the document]
HofiOne: oh, I see, in the grammar, definitly, but as you said not to much we can do if we nated to keep the grammar afmongodb instalation independent

MrAnno:
Thanks. LGTM.

I couldn't find any side effects of the bulk send feature (I tested your previous version), so we can enable it by default if you want it too.

Please squash your review commits so that we can merge the PR.

HofiOne:
Done.

MrAnno (approved):

Issue #4482, opened 2023-05-16
python: multiple `options()` do not merge


alltilla:
# syslog-ng
## Version of syslog-ng
4.2.0

# Issue
#4354 introduced a change which is not backward compatible.
Before it the content of multiple `options()` blocks were merged together, after it, the later one rewrites the earlier one.

## Steps to reproduce
Run with the config below.

## Configuration
```
@version: 4.2

python {
import syslogng

def assert_options(options):
    assert options["a"] == "a"
    assert options["b"] == "b"
    assert options["c"] == "c"
    assert options["d"] == "d"


class TestSource(syslogng.LogSource):
  def init(self, options):
    assert_options(options)
    return True
};

source s_python {
  python(
    class("TestSource")
    options(
      "a" => "a"
      "b" => "b"
    )
    options(
      "c" => "c"
      "d" => "d"
    )
  );
};

destination d_local {
  file("/dev/stdout");
};

log {
  source(s_python);
  destination(d_local);
};
```

## Input and output logs
```
$ ./install/sbin/syslog-ng -Fe
[2023-05-16T12:42:45.382482] Exception while calling a Python function; caller='s_python#0', class='TestSource', function='init', exception='KeyError: \'a\''
Traceback (most recent call last):
  File "/home/alltilla/repos/syslog-ng/build/install/etc/syslog-ng.conf{python-global-code:3}", line 13, in init
    assert_options(options)
  File "/home/alltilla/repos/syslog-ng/build/install/etc/syslog-ng.conf{python-global-code:3}", line 5, in assert_options
    assert options["a"] == "a"
KeyError: 'a'
[2023-05-16T12:42:45.382578] python-source: Error initializing Python driver object, init() returned FALSE; driver='s_python#0', class='TestSource'
[2023-05-16T12:42:45.382586] Error initializing message pipeline; plugin_name='python', location='/home/alltilla/repos/syslog-ng/build/install/etc/syslog-ng.conf:20:3'
```

PR #4487, opened 2023-05-17
python: multiple `options` do not merge

HofiOne:
Fix for #4482 

### DIFF
--- a/modules/python/tests/test_python_options.c
+++ b/modules/python/tests/test_python_options.c
@@ -164,7 +164,8 @@ Test(python_options, test_python_option_template)
   python_option_free(option);
 }
 
-Test(python_options, test_python_options)
+static PythonOptions *
+_generate_options_test_options(void)
 {
   PythonOptions *options = python_options_new();
   PythonOption *option;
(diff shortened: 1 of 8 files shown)


alltilla (approved):
Nice!

---
[optional]

Can you add a simple unit test for this?

HofiOne:
> [optional]
> 
> Can you add a simple unit test for this?

Done.

modules/python/tests/test_python_options.c
```diff
@@ -219,6 +217,30 @@ Test(python_options, test_python_options)
...
+
+  gchar *script2 = g_strdup_printf("%s%s", script, script);
+  PythonOptions *options2 = _generate_options_test_options();
+  python_options_add_options(options2, options);
+  python_options_release(options);
+
+  _run_py_options_test(options2, script2);
+
+  g_free(script2);
+  python_options_free(options2);
```
alltilla, line 245: Sorry if I am missing something, but I do not how this test case works.

We test an `Options` object with every option types, then we are merging it with the same values and running the test two more times on it?

I think we should test `add_options()` with options that are not already present in the `Options` object. Also I do not understand the reason behind `gchar *script2 = g_strdup_printf("%s%s", script, script);`, it should be enough to run the script once.

HofiOne: oh, I see, did not take into account that these are named params
will fix it

alltilla: Ah, I see why you did that :) Yes they are named. Thanks for changing it!

HofiOne: hope it's fine now

PR #4488, opened 2023-05-18
logscale: raise workers to 20

alltilla:


bazsi:
This is a good stop gap measure, but I would want to decouple batches from workers in the longer term.

bazsi (approved):

PR #4489, opened 2023-05-20
Keep python core while disabling python modules

### LINKED ISSUE #4486: --disable-python-modules kills Python support
While trying to split Python modules from Python support in FreeBSD packaging, I ran into a problem. If I use the ```--disable-python-modules``` configure option, then the complete ```/usr/local/lib/syslog-ng/python/``` directory is missing. When I try to start or syntax check a previously working syslog-ng configuration with an in-line Python code, I get the following error:

```
root@fb132:~/freebsd/syslog-ng4-devel # syslog-ng -s
[2023-05-17T14:25:22.588072] python: the current set of requirements installed in our virtualenv seems to be out of date, use the `syslog-ng-update-virtualenv' script to upgrade Python dependencies; path='/var/db/python-venv'
[2023-05-17T14:25:22.617683] Error evaluating global Python block; exception='ModuleNotFoundError: No module named \'syslogng\''
  File "/usr/local/etc/syslog-ng.conf{python-global-code:186}", line 10, in <module>
    from syslogng import LogParser
ModuleNotFoundError: No module named 'syslogng'
Error parsing python, Error processing global python block in /usr/local/etc/syslog-ng.conf:186:1-186:7:
181     # *.*                            /var/log/ppp.log
182     #
183     log { source(src); filter(f_ppp); destination(ppp); };
184     
185     
186---> python {
186---> ^^^^^^
187     
188     """
189     very simple syslog-ng Python parser example
190     resolves IP to hostname
191     value pair names are hard-coded
```

It is on FreeBSD 13.2 and syslog-ng 4.2.0 release.

My expectation was, that only ```/usr/local/lib/syslog-ng/python/syslogng/modules``` will be missing with the above option.

If I install syslog-ng with --enable-python-modules, and remove this directory by hand, then the above syslog-ng configuration starts without an error. The warning message I mentioned in https://github.com/syslog-ng/syslog-ng/issues/4471 is still there if I remove ```/var/db/python-venv```, but syslog-ng does not fail to start anymore:

```
root@fb132:/var/db # syslog-ng -s
[2023-05-17T14:41:09.099197] python: private virtualenv is not initialized, use the `syslog-ng-update-virtualenv' script to initialize it or make sure all required Python dependencies are available in the system Python installation; path='/var/db/python-venv'
```

bazsi:
This should fix #4486 by installing the core syslog-ng Python modules even if --disable-python-modules is used.

@czanik can you please test if this fixes your issue?

### DIFF
--- a/Mk/python-venv.am
+++ b/Mk/python-venv.am
@@ -1,7 +1,7 @@
 
 PYTHON_VENV_TOUCHFILE=$(abs_builddir)/.python-venv-built
 
-$(PYTHON_VENV_TOUCHFILE): Makefile
+$(PYTHON_VENV_TOUCHFILE): dev-requirements.txt
 	if [ "$(with_python_packages)" = "venv" ]; then \
 	    $(top_srcdir)/scripts/build-python-venv.sh "$(PYTHON)" "$(PYTHON_VENV_DIR)" "$(top_srcdir)"; \
 	fi
--- a/configure.ac
+++ b/configure.ac
@@ -350,7 +350,7 @@ AC_ARG_ENABLE(python,
               ,,enable_python="auto")
 
 AC_ARG_ENABLE(python-modules,
-              [  --disable-python-modules  Disable Python based syslog-ng modules]
+              [  --disable-python-modules  Disable Python based syslog-ng addons while installing the core Python modules]
               ,,enable_python_modules="auto")
 
 AC_ARG_WITH(python,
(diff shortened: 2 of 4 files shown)


czanik:
@bazsi thanks! Tested on FreeBSD and works as expected.

In my RPM packages I also split off Python-based modules. There I cannot use --disable-python-modules, as I want to package Python-based modules, just in a separate sub-package. I'll double check that I split the files along these lines. As far as I can see, all I have to do is to make sure that ```lib/syslog-ng/python/syslogng-1.0-py3.9.egg-info/requires.txt``` goes to the ```syslog-ng-python-modules``` package.

alltilla (approved):

